PR #313, opened 2026-08-06
feat(w3): forward's source= resolves OFF-ORDER — the ACTIVE-only bound of PR #311 lifted (W3 PR 2b)

Avick3110:
**W3 PR 2b** — lift `housecarl_forward`'s ACTIVE-only `source=` bound. Chartered 2026-08-05 (Aaron-go), its own PR by Aaron's pick, slotted `#311 → this → PR 3 copy → PR 4 check`.

**This supersedes the first entry in [#311](https://github.com/Avick3110/houseCARL/pull/311)'s "Decisions and bounds" list** — *"`forward`'s `source=` is ACTIVE-only, declared not narrowed"*. That entry is no longer true of the code; it is not silently dropped, it is answered. Its trailing offer (*"candidate GAP issue if you'd rather it were filed"*) was declined on the standing rule: **no GAP filing during dev** — the Issues lane is for gaps in *shipped* surface; a gap you are actively building through gets built.

## Why this was a cornerstone fix, not a papercut

CLAUDE.md §1 lists reading **inactive plugins** — "unchecked, or inside a disabled mod" — as a headline capability. `forward` refused a `source=` on disk but not in the active order, which narrowed exactly the case that capability exists for: re-asserting a *disabled* old patch's version of a record, or reading an old version's body to convert it. Declaring the bound made it honest; it did not make it right. §4 outcome **(a)** — the claim holds and the clean solution below respects it.

## The change

The forward twin of the `CopyFrom` lane's `ResolveOffOrderCopySources` (PR #310), and simpler: every forward in a call shares **one** source, so it is locate-once / open-once / fetch-N in a single overlay walk rather than a per-edit loop. Same on-disk locate as `read_plugin_file` and the copy-npc-appearance donor lane, so two tools cannot disagree about which file a filename names. **Both lanes** get it (build-plan open question 1): the in-place *target* must stay active — that lane's own contract — but the *source* has no such need.

| piece | where |
|---|---|
| `ResolveOffOrderForwardSource` | `LoadOrderService` — locate, open, one walk collecting every wanted key, all-or-nothing refusals, overlay handed back open |
| `OffOrderForwardSource` + `SourceCacheFor` + `StampOffOrderSource` | `WritePatchBuilder` |
| `offOrder` threaded | `ForwardRecords` / `ForwardRecordsInPlace` + their `*Core` splits → `ResolveForwardSources` |
[...]

Issue #225, opened 2026-07-18
Feature: dry_run mode for write tools


Avick3110:
`bulk_apply`/`set_field`/`forward_record` with `dry_run=true` → validate each op against the record
schema and the live winner (field exists, writable, type matches, FormID references resolve) and
report what WOULD change, without writing. The 745-record session compensated with post-hoc
read-backs; a dry-run would catch a bad field path before the first write, not after the 745th.

*Provenance: 2026-07-17 Apocalypse - Magic of Skyrim whole-mod patch session (authoria-requiem 1.5.0 stress test): 16 agents, ~400+ reads, 745-record write, zero tool BUGS - these are DX/feature items.*

PR #315, opened 2026-08-06
fix: an UNOPENABLE active plugin no longer breaks every write (Fixes #314)

### LINKED ISSUE #314: [Bug] one unopenable active plugin makes EVERY write fail — AllMasters retains the exclusion class that cannot be opened
## What happens

One active plugin that **Mutagen cannot open** makes **every write fail** — `apply`, `create`, `remove`, `forward`, on both the patch and in-place lanes — with a raw serializer exception that names neither the offending plugin nor a remedy:

```
error: writing the patch failed (serialize or commit; the existing file is untouched):
ModGroupsMalformedException: Exception of type 'Mutagen.Bethesda.Plugins.Exceptions.ModGroupsMalformedException' was thrown.
[inner: ArgumentException: Cannot move position past the length of the stream]
```

Reads are unaffected — the index build handles this correctly and excludes the plugin with a clear reason. It is only writes.

## Why

`BuildIndex` excludes a plugin along two different paths (`LoadOrderResolver.cs` ~395 and ~411):

1. **`OpenOverlay` throws** — the file can't be opened at all;
2. **`EnumerateMajorRecords` throws** — it opens, but a record body is unparseable.

`OverlaySession.AllMasters()` / `AllMastersExcept()` then open **every** plugin including excluded ones, which is deliberate and documented:

> Excluded plugins (BuildIndex couldn't fully parse) are INTENTIONALLY retained here — NOT filtered by the snapshot's Excluded set. A clean plugin can override a record whose ORIGIN master is an excluded one, and that master must still appear in the patch's output header for FormID resolution; dropping it would corrupt the header. **Safe: Overlay() opens lazily (no parse, no enumeration → no throw)** […] and it cannot re-throw.

That safety argument holds for class 2 and **is false for class 1**. `Overlay(idx)` calls `OpenOverlay` → `SkyrimMod.CreateFromBinaryOverlay`, which parses group structure eagerly enough to throw on a malformed or truncated file. So the one exclusion class that exists *because the file cannot be opened* is the one the write path then tries to open, every time.

The retention rationale is still right — the header genuinely needs that master. The bug is that the class distinction was never made: a plugin excluded for *unopenable* can't contribute a resolvable overlay at all, so retaining it buys nothing and costs every write.

## Reproduction

Synthetic, and it is how this surfaced. In a load order with one enabled, ticked plugin whose file is a valid header followed by a truncated record body:

```csharp
var whole = File.ReadAllBytes(pluginPath);
File.WriteAllBytes(pluginPath, whole[..(whole.Length - 12)]);
```

`load_order_status` reports it excluded, cleanly. Every subsequent write in that order then fails as above — including writes that have nothing to do with the broken plugin.

A truncated plugin is not an exotic state: an interrupted download, a half-copied mod folder, or a killed xEdit save all produce one.

## Suggested direction (not a decision)

Filter the **open-failure** class out of `AllMasters`/`AllMastersExcept` while keeping the enumerate-failure class retained (its overlay opens fine and the header argument applies to it). The `Excluded` set alone is too coarse — it would drop class 2 as well and reintroduce the header corruption the comment warns about — so the exclusion record needs to carry which class it was.

Whatever the fix, the failure should be **named**: "plugin X is in your load order but cannot be opened (…); no write can serialize against this order until it is removed or repaired" beats a raw `ModGroupsMalformedException` that reads as a disk fault.

## Provenance

Found while building a guard fixture for a review finding on [#313](https://github.com/Avick3110/houseCARL/pull/313) (W3 PR 2b). Adding a deliberately unparseable plugin to that probe's synthetic order broke every unrelated write arm in it, which is what exposed this. Not fixed there — it is in `LoadOrderResolver`, is shipped behaviour predating 2.0, and is unrelated to that PR's subject.

The mechanism above is confirmed by reading `Overlay(int)` → `OpenOverlay` → `CreateFromBinaryOverlay`, not inferred from the failure alone.

Avick3110:
Fixes #314.

One plugin houseCARL cannot **open** sitting in the active load order made **every write fail** — `apply`, `create`, `remove`, `forward`, both lanes — including writes with nothing to do with it, with an opaque `ModGroupsMalformedException` that reads as a disk fault. Reads were never affected.

Found while building a guard fixture for a review finding on #313: adding a deliberately unparseable plugin to that probe's synthetic order broke every unrelated write arm in it. Filed rather than folded there (different lane, shipped behaviour, core seam), and this is that PR.

## The defect

`BuildIndex` excludes a plugin along two paths — `OpenOverlay` throws, or `EnumerateMajorRecords` throws. `AllMasters`/`AllMastersExcept` then opened **every** plugin including excluded ones, deliberately and with a stated safety argument:

> Safe: `Overlay()` opens lazily (no parse, no enumeration → no throw) […] and it cannot re-throw.

True of the enumerate-failure class. **False of the open-failure class** — the one that exists precisely because the file cannot be opened. `Overlay(idx)` calls `OpenOverlay` → `CreateFromBinaryOverlay`, which parses group structure eagerly enough to throw on a truncated file, so it threw again on every write.

The retention *rationale* was sound and is kept: a clean plugin can override a record whose origin master is excluded, and the header needs that master. It just cannot apply to a plugin that will not open.

## The fix

The snapshot carries the open-failure subset as its own `Unopenable` set — **a fact, not a substring of a display message** — populated only at the `OpenOverlay` catch. The two master-set builders skip it. Nothing outside `LoadOrderResolver` reads the `Excluded` int set (checked: external code sees only `ExcludedPlugins`), so nothing else changes behaviour.

## What the guard changed about the fix — worth reading

The first draft asserted that skipping **costs the output nothing**, on solid-looking evidence: a patch whose record *originates* in the skipped plugin still writes, because Mutagen derives that header entry from the record's own FormKey rather than from list membership. I verified that, and deleted the reporting machinery I'd built as unearned speculation.

Then the **multi-master** arm disproved it. Once a header must be *sorted* — two or more masters — the serializer refuses with `MissingModException` naming the skipped plugin. The residual failure is real and reachable, and the machinery came back because a test earned it.

That residual is now a named **CAUSE** appended at every serialize-failure render that has a session: which plugin, that it is **ACTIVE but unopenable** (so the pre-existing `"references a plugin that is NOT active"` wording would have been wrong for it), what to do, and that unrelated writes are unaffected. **Appended, not substituted**, and matched on the exception naming a skipped plugin — a mismatch degrades to the ordinary description rather than to a confident wrong cause.

**Nothing regresses.** That case failed before too, with a worse message.

*(An earlier version of this paragraph also claimed a single-master patch on the broken plugin now writes. Review 3 showed that is unreachable in a real order — `WritePatch` force-includes `Skyrim.esm` + `Update.esm`, so a patch-lane header always carries two or more and always takes the sorted-header residual. It happens only in a baseline-less harness, which is what the original fixture was. Corrected here, in the changelog, and in the code comment that repeated it; `RealOrderArm` now pins the real-order behaviour both ways.)*

## Its own probe, because the fixture cannot be shared

An unopenable plugin poisons every write in whatever order it sits in — that *is* the bug. `excluded-master-guard` builds its own MO2 instances (three: the main order, a baseline-bearing one, and a real order): a master, a clean plugin that wins the subject **and** overrides a record the broken plugin originates, and the broken plugin itself (valid header, truncated body), all active and ticked.

**Nineteen checks across three orders**, grown by the review rounds:

- **the baseline-less order** (the original): the plugin is excluded and reads are unaffected · forward / apply / create into a new patch all succeed · naming the excluded plugin as a source is still refused (the skip must not become an escape hatch) · a patch whose record originates in it still writes **and still masters on it** · a sorted multi-master header refuses, **pinned** as the expected outcome so an upstream flip fails loudly instead of passing · dry-run parity both ways · the in-place and remove-in-place lanes.
- **a baseline-bearing order** (review 2): an unopenable `Skyrim.esm` refuses a self-contained create rather than emitting a master-less plugin, renders that refusal **exactly once** without the phase it never reached, is predicted by `dry_run`, and reaches `copy_npc_appearance` through the same renderer (review 4).
- **a real order** (review 3): baselines present and openable. The unrelated write lands with them in its header; the write referencing the broken plugin's record refuses. This one exists because the original fixture's baseline-less shape taught me a claim about a case users cannot reach.

**RED-verified** each round against the behaviour being fixed — the skip, the baseline refusal, the doubling, the npc-copy override — with the pre-fix output quoted in the fold comments.

`ci-all` **117/117** (116 + this probe). That matters more than usual: this is a core seam every write path funnels through, so the existing suite passing unchanged is most of the safety argument.

**Not guarded**, stated rather than implied away: `RemapEngine.RepointInPlace` and `CompactBuild` (each needs a compact/merge fixture with an unopenable declared master), and the in-place lane's single-master threshold case (predicted, unverified).

## Note on sequencing

Branched from `main` while [#313](https://github.com/Avick3110/houseCARL/pull/313) is still open. The two are independent, but both touch `WritePatchBuilder`'s serialize-failure renders — #313 adds a named `MissingModException` arm to `forward`'s patch lane. Whichever merges second gets a small, mechanical rebase there; I'll take it.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -8,6 +8,15 @@ when it changes.
 
 *Accumulating notes for the next cut — not yet released; `plugin.json` still reads the last shipped version.*
 
+**Fixed: one broken plugin in your load order no longer breaks every write.** If a plugin houseCARL cannot open
+sat in your active order — a truncated download, a half-copied mod folder, an interrupted xEdit save — then
+*every* write failed, including writes that had nothing to do with it, with an opaque message that read like a
+disk fault. Reads were never affected. Now writes that don't reference that plugin's records just work.
+
+A write that genuinely *does* need it (its records can still be reached through a plugin that overrides them) is
+refused **naming it as the cause**, with what to do — instead of the previous engine-level error. One such write
+now succeeds that didn't before: a patch whose header needs only that single master.
+
 **New tools: `housecarl_create`, `housecarl_remove`, `housecarl_forward` — the rest of the 2.0 write surface
 (W3 PR 2).** The same shape `housecarl_apply` introduced — *what changes* × *where it lands* × *how it reads
 back* — now covers authoring, removal and whole-record overrides.
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -14,8 +14,9 @@ sat in your active order — a truncated download, a half-copied mod folder, an 
 disk fault. Reads were never affected. Now writes that don't reference that plugin's records just work.
 
 A write that genuinely *does* need it (its records can still be reached through a plugin that overrides them) is
-refused **naming it as the cause**, with what to do — instead of the previous engine-level error. One such write
-now succeeds that didn't before: a patch whose header needs only that single master.
+refused **naming it as the cause**, with what to do — instead of the previous engine-level error. And if the
+unopenable plugin is `Skyrim.esm` or `Update.esm` — the two masters every written plugin must list — nothing is
+written at all until you repair it, rather than a plugin landing on disk without them.
 
 **New tools: `housecarl_create`, `housecarl_remove`, `housecarl_forward` — the rest of the 2.0 write surface
 (W3 PR 2).** The same shape `housecarl_apply` introduced — *what changes* × *where it lands* × *how it reads
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -83,9 +83,9 @@ public sealed class UnopenableBaselineMasterException : Exception
         // no write can resolve against a master it cannot open.
         : base($"'{pluginName}' is a BASELINE master (Skyrim.esm / Update.esm) and is ACTIVE in your load order, but " +
                "houseCARL cannot open it — see load_order_status for the reason. Nothing can be written while that is " +
-               "true: no write can resolve references against a master it cannot read, and a new patch must list the " +
-               "baselines in its header, so emitting one without them would produce a plugin the game treats as " +
-               "malformed. Repair or replace that plugin in MO2 and retry. Nothing was written.")
+               "true, for either reason alone: a new patch must list the baselines in its header (emitting one without " +
+               "them produces a plugin the game treats as malformed), and no write can resolve references against a " +
+               "master it cannot read. Repair or replace that plugin in MO2 and retry. Nothing was written.")
         => PluginName = pluginName;
 }
 
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -127,6 +127,12 @@ public sealed class LoadOrderResolver : IDisposable
     /// <see cref="ExcludedPlugins"/>, formatted "name: reason" for log/harness display.</summary>
     public IReadOnlyList<string> LoadFailures => _snap.LoadFailures;
 
+    /// <summary>#314 — is this plugin ACTIVE but impossible to OPEN? The resolver-level twin of
+    /// <see cref="IndexView.IsUnopenable"/>, for callers holding the resolver rather than a captured view (the dry-run
+    /// master preview). Reads the current snapshot, which is what a dry run should predict against.</summary>
+    public bool IsUnopenable(string pluginName)
+        => _nameToIdx.TryGetValue(pluginName, out int i) && _snap.Unopenable.Contains(i);
+
     /// <summary>Plugins EXCLUDED from this build (name → why): unopenable, or carrying a record Mutagen can't parse.
     /// Their records are not in the index and no path will re-touch them; load_order_status reports them so the user
     /// can fix/remove the upstream plugin (Q3 — the exclusion is visible, not silent).</summary>
@@ -260,7 +266,10 @@ public sealed class LoadOrderResolver : IDisposable
         /// <summary>The plugins this session's master-set builds skipped as unopenable (#314). Empty in the normal case.
         /// Non-empty ⇒ a serialize failure naming a missing master is very likely one of THESE, and the write lane says
         /// so instead of reporting an opaque engine fault.</summary>
-        public IReadOnlyCollection<string> SkippedUnopenable => _skippedUnopenable;
+        /// <remarks>Exposed as a SET, not a collection: consumers ask "is this name in it?", and the backing
+        /// comparer is OrdinalIgnoreCase — a LINQ Contains over a collection would have compared ordinally and missed a
+        /// case difference between the profile's spelling and a ModKey's.</remarks>
+        public IReadOnlySet<string> SkippedUnopenable => _skippedUnopenable;
 
         /// <summary>Dispose and forget any overlay this session holds on <paramref name="fileName"/> — the file the caller
         /// is about to serialize to. The SECOND half of the active-patch write fix (with <see cref="AllMastersExcept"/>):
@@ -574,6 +583,12 @@ public sealed class LoadOrderResolver : IDisposable
         public int MaxDepth => _s.MaxDepth;
         public IReadOnlyList<string> LoadFailures => _s.LoadFailures;
         public IReadOnlyDictionary<string, string> ExcludedPlugins => _s.ExcludedPlugins;
+
+        /// <summary>#314 — is this plugin the could-not-be-OPENED exclusion class? A caller that is ABOUT to open a
+        /// plugin file itself (rather than going through the master-set builders) asks this first, so an unopenable one
+        /// becomes a named refusal instead of an exception escaping from a bare CreateFromBinaryOverlay.</summary>
+        public bool IsUnopenable(string pluginName)
+            => _r._nameToIdx.TryGetValue(pluginName, out int i) && _s.Unopenable.Contains(i);
 
         /// <summary>THIS captured build's epoch fingerprint — the stamp a bulk response computed off this view
         /// must carry (SPEC §2.1.1). Immutable with the snapshot: a concurrent rebuild changes nothing here.</summary>
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -70,6 +70,21 @@ public readonly record struct WinnerInfo(FormKey FormKey, string WinnerPlugin, i
 public readonly record struct RecordStatus(
     FormKey FormKey, string RecordType, bool PluginWins, int OverrideDepth, IReadOnlyList<string> TouchingPlugins);
 
+/// <summary>#314 / PR #315 review 2 — a BASELINE master (Skyrim.esm / Update.esm) is active but cannot be opened.
+/// Every written plugin force-includes the baselines (Aaron-locked 2026-06-02) and that force-include is derived from
+/// the known-master list, so quietly omitting one would emit a plugin missing a mandatory master. Thrown from the
+/// master-set builders — the single point every write lane funnels through — so no lane can forget the check.</summary>
+public sealed class UnopenableBaselineMasterException : Exception
+{
+    public string PluginName { get; }
+    public UnopenableBaselineMasterException(string pluginName)
+        : base($"'{pluginName}' is a BASELINE master (every written plugin must list it) and is ACTIVE in your load " +
+               "order, but houseCARL cannot open it — see load_order_status for the reason. No plugin can be written " +
+               "until it is repaired or replaced: writing one without it would produce a plugin the game treats as " +
+               "malformed. Nothing was written.")
+        => PluginName = pluginName;
+}
+
 public sealed class LoadOrderResolver : IDisposable
 {
     readonly string[] _paths;                          // every active plugin's path, priority order (masters → … → winner)
@@ -257,7 +272,18 @@ public sealed class LoadOrderResolver : IDisposable
         bool SkipUnopenable(int i)
         {
             if (!_r._snap.Unopenable.Contains(i)) return false;
-            _skippedUnopenable.Add(_r._names[i]);
+            var name = _r._names[i];
+            // A BASELINE master (Skyrim.esm / Update.esm) must never be skipped. WriteEngine.WritePatch derives its
+            // CK-mandated force-include FROM the list returned here, filtered to baselines present in it — a filter
+            // whose stated purpose is tolerating a degenerate order or a single-master harness. Skipping an unopenable
+            // baseline makes a REAL order look degenerate to it, and a plugin lands on disk missing a master Aaron
+            // locked as mandatory (2026-06-02) with no warning: a SILENT degradation where this PR found a loud
+            // failure, which is the one trade Q3 never allows (PR #315 review 2). A baseline is never legitimately
+            // absent, so this refuses instead — thrown from the single chokepoint every write lane funnels through,
+            // rather than re-checked per lane.
+            if (Array.Exists(WriteEngine.BaselineMasters, bm => string.Equals(bm.FileName.String, name, StringComparison.OrdinalIgnoreCase)))
+                throw new UnopenableBaselineMasterException(name);
+            _skippedUnopenable.Add(name);
             return true;
         }
 
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -261,9 +261,11 @@ public sealed class LoadOrderResolver : IDisposable
         /// Records the name, because the skip is NOT free and the guard proved exactly where:
         /// <list type="bullet">
         /// <item>a patch whose header needs only ONE master still writes, even when that master is the skipped plugin —
-        /// Mutagen derives the entry from the record's own FormKey, not from list membership;</item>
+        /// Mutagen derives the entry from the record's own FormKey, not from list membership. NOTE this is reachable
+        /// only in an order WITHOUT the baselines (a test harness): a real order force-includes Skyrim.esm + Update.esm,
+        /// so a patch-lane header always carries two or more and always takes the second case (PR #315 review 3);</item>
         /// <item>a patch whose header must be SORTED (two or more masters) does NOT: the serializer refuses with
-        /// <c>MissingModException</c> naming the skipped plugin.</item>
+        /// <c>MissingModException</c> naming the skipped plugin. In practice this is the case that fires.</item>
         /// </list>
         /// That second case is a real, reachable failure, and it must be named rather than surfaced as a generic
         /// "serialize or commit" fault (Q3) — which is what <see cref="SkippedUnopenable"/> is for. The first draft of
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -78,13 +78,23 @@ public sealed class UnopenableBaselineMasterException : Exception
 {
     public string PluginName { get; }
     public UnopenableBaselineMasterException(string pluginName)
-        : base($"'{pluginName}' is a BASELINE master (every written plugin must list it) and is ACTIVE in your load " +
-               "order, but houseCARL cannot open it — see load_order_status for the reason. No plugin can be written " +
-               "until it is repaired or replaced: writing one without it would produce a plugin the game treats as " +
-               "malformed. Nothing was written.")
+        // Worded to hold on BOTH write lanes (PR #315 review 4): "every written plugin must list it" is the PATCH
+        // lane's reason — WriteInPlace force-includes no baselines — while the fact that covers both is simply that
+        // no write can resolve against a master it cannot open.
+        : base($"'{pluginName}' is a BASELINE master (Skyrim.esm / Update.esm) and is ACTIVE in your load order, but " +
+               "houseCARL cannot open it — see load_order_status for the reason. Nothing can be written while that is " +
+               "true: no write can resolve references against a master it cannot read, and a new patch must list the " +
+               "baselines in its header, so emitting one without them would produce a plugin the game treats as " +
+               "malformed. Repair or replace that plugin in MO2 and retry. Nothing was written.")
         => PluginName = pluginName;
 }
 
+/// <remarks>LAYERING NOTE (#314 / PR #315 review 4): this file's master-set builders reference
+/// <see cref="WriteEngine.BaselineMasters"/> and throw <see cref="UnopenableBaselineMasterException"/> — a WRITE policy,
+/// and the first write dependency here. Deliberate, not drift: the builders are the single point every write lane
+/// funnels through, so enforcing it there is what makes "no lane can forget the check" true, where a per-lane check is
+/// something the next lane forgets. Kept on the chokepoint argument over the purity one; recorded so a later tidy-up
+/// re-litigates it rather than silently reverting it.</remarks>
 public sealed class LoadOrderResolver : IDisposable
 {
     readonly string[] _paths;                          // every active plugin's path, priority order (masters → … → winner)
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -104,13 +104,20 @@ public sealed class LoadOrderResolver : IDisposable
         public readonly Dictionary<FormKey, int[]> Overriders;                // MULTI keys only — ordered touching overlay indices
         public readonly List<string> LoadFailures;                            // per-plugin index-build failures (open OR parse), surfaced (Q3)
         public readonly HashSet<int> Excluded;                                // overlay indices excluded this build — never re-touched by any path
+        /// <summary>#314 — the SUBSET of <see cref="Excluded"/> whose file could not be OPENED at all, as opposed to
+        /// opening fine and failing on a record body. The distinction is invisible to reads (both are "never touch
+        /// this plugin") and load-bearing for WRITES: the master set retains excluded plugins on purpose, which is
+        /// only possible for the ones that can be opened. Kept as its own set rather than derived from the reason
+        /// STRING — a message is display prose that can be reworded, membership is a fact.</summary>
+        public readonly HashSet<int> Unopenable;
         public readonly Dictionary<string, string> ExcludedPlugins;           // excluded plugin name → reason (Q3)
         public readonly int MaxDepth;
         public readonly string Epoch;                                         // this build's fingerprint (SPEC §2.1.1) — immutable with the snapshot
 
         public IndexSnapshot(Dictionary<FormKey, (int winner, int count)> index, Dictionary<FormKey, int[]> overriders,
-                             List<string> loadFailures, HashSet<int> excluded, Dictionary<string, string> excludedPlugins, int maxDepth, string epoch)
-        { Index = index; Overriders = overriders; LoadFailures = loadFailures; Excluded = excluded; ExcludedPlugins = excludedPlugins; MaxDepth = maxDepth; Epoch = epoch; }
+                             List<string> loadFailures, HashSet<int> excluded, HashSet<int> unopenable,
+                             Dictionary<string, string> excludedPlugins, int maxDepth, string epoch)
+        { Index = index; Overriders = overriders; LoadFailures = loadFailures; Excluded = excluded; Unopenable = unopenable; ExcludedPlugins = excludedPlugins; MaxDepth = maxDepth; Epoch = epoch; }
     }
 
     volatile IndexSnapshot _snap;
@@ -177,16 +184,24 @@ public sealed class LoadOrderResolver : IDisposable
         /// Tier-2 perf idea above.</summary>
         public IReadOnlyList<ISkyrimModGetter> AllMasters()
         {
-            // Excluded plugins (BuildIndex couldn't fully parse) are INTENTIONALLY retained here — NOT filtered by
-            // the snapshot's Excluded set. A clean plugin can override a record whose ORIGIN master is an excluded one, and that master
-            // must still appear in the patch's output header for FormID resolution; dropping it would corrupt the
-            // header. Safe: Overlay() opens lazily (no parse, no enumeration → no throw), and the serializer parses a
-            // master's bodies ONLY when the patch references them — and an excluded plugin's OWN records can't be
-            // referenced (they're unreadable). So this is the one read-path that touches excluded plugins on purpose,
-            // and it cannot re-throw. (If a future change enumerates or link-caches over this full set, it must
-            // re-introduce the Excluded skip — the per-read "single choke point" is BuildIndex, not this method.)
-            var arr = new ISkyrimModGetter[_r._paths.Length];
-            for (int i = 0; i < arr.Length; i++) arr[i] = Overlay(i);
+            // Excluded plugins that OPEN are INTENTIONALLY retained here — NOT filtered by the snapshot's Excluded
+            // set. A clean plugin can override a record whose ORIGIN master is an excluded one, and that master must
+            // still appear in the patch's output header for FormID resolution; dropping it would corrupt the header.
+            // Safe for THAT class: Overlay() opens lazily (no parse, no enumeration → no throw), and the serializer
+            // parses a master's bodies ONLY when the patch references them.
+            //
+            // #314 — but the OTHER exclusion class cannot be retained, because retaining it is impossible: a plugin
+            // excluded BECAUSE OpenOverlay threw makes Overlay(i) throw again here, on EVERY write, including writes
+            // that never touch it. The old comment's "it cannot re-throw" was true of the parse-failure class and
+            // false of this one. Skipping loses nothing the retention argument was protecting — an overlay we cannot
+            // open contributes no header entry either way — and the names are recorded so a write that genuinely
+            // NEEDED one can say so instead of dying as a raw serializer fault.
+            var arr = new List<ISkyrimModGetter>(_r._paths.Length);
+            for (int i = 0; i < _r._paths.Length; i++)
+            {
+                if (SkipUnopenable(i)) continue;
+                arr.Add(Overlay(i));
+            }
             return arr;
         }
 
@@ -213,10 +228,39 @@ public sealed class LoadOrderResolver : IDisposable
         {
             var list = new List<ISkyrimModGetter>(_r._paths.Length);
             for (int i = 0; i < _r._paths.Length; i++)
-                if (!string.Equals(_r._names[i], excludeFileName, StringComparison.OrdinalIgnoreCase))
-                    list.Add(Overlay(i));   // SKIP the target's index — never map the file we're about to overwrite
+            {
+                if (string.Equals(_r._names[i], excludeFileName, StringComparison.OrdinalIgnoreCase)) continue;  // never map the file we're about to overwrite
+                if (SkipUnopenable(i)) continue;                                                                 // #314 — and never try to map one that cannot be opened
+                list.Add(Overlay(i));
+            }
             return list;
         }
+
+        /// <summary>#314 — is this index the could-not-be-OPENED exclusion class, which no master set can contain?
+        /// Records the name, because the skip is NOT free and the guard proved exactly where:
+        /// <list type="bullet">
+        /// <item>a patch whose header needs only ONE master still writes, even when that master is the skipped plugin —
+        /// Mutagen derives the entry from the record's own FormKey, not from list membership;</item>
+        /// <item>a patch whose header must be SORTED (two or more masters) does NOT: the serializer refuses with
+        /// <c>MissingModException</c> naming the skipped plugin.</item>
+        /// </list>
+        /// That second case is a real, reachable failure, and it must be named rather than surfaced as a generic
+        /// "serialize or commit" fault (Q3) — which is what <see cref="SkippedUnopenable"/> is for. The first draft of
+        /// this fix asserted the skip cost the output nothing and deleted this; the multi-master guard arm disproved
+        /// it.</summary>
+        bool SkipUnopenable(int i)
+        {
+            if (!_r._snap.Unopenable.Contains(i)) return false;
+            _skippedUnopenable.Add(_r._names[i]);
+            return true;
+        }
+
+        readonly SortedSet<string> _skippedUnopenable = new(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>The plugins this session's master-set builds skipped as unopenable (#314). Empty in the normal case.
+        /// Non-empty ⇒ a serialize failure naming a missing master is very likely one of THESE, and the write lane says
+        /// so instead of reporting an opaque engine fault.</summary>
+        public IReadOnlyCollection<string> SkippedUnopenable => _skippedUnopenable;
 
         /// <summary>Dispose and forget any overlay this session holds on <paramref name="fileName"/> — the file the caller
         /// is about to serialize to. The SECOND half of the active-patch write fix (with <see cref="AllMastersExcept"/>):
@@ -386,6 +430,7 @@ public sealed class LoadOrderResolver : IDisposable
         var overriders = new Dictionary<FormKey, List<int>>();        // multi keys only
         var failures = new List<string>();
         var excluded = new HashSet<int>();
+        var unopenable = new HashSet<int>();                          // #314 — the could-not-be-OPENED subset
         var excludedPlugins = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
         int maxDepth = 0;
 
@@ -396,6 +441,7 @@ public sealed class LoadOrderResolver : IDisposable
             catch (Exception ex)
             {
                 Exclude(i, $"could not be opened — {Concise(ex)}", failures, excluded, excludedPlugins);
+                unopenable.Add(i);   // #314 — this one cannot serve as a master overlay either; the write path must skip it
                 continue;
             }
 
@@ -437,7 +483,7 @@ public sealed class LoadOrderResolver : IDisposable
         return new IndexSnapshot(
             index,
             overriders.ToDictionary(kv => kv.Key, kv => kv.Value.ToArray()),  // trim List overhead → int[]
-            failures, excluded, excludedPlugins, maxDepth, ComputeEpoch(_names, _paths, _mtimes, excludedPlugins));
+            failures, excluded, unopenable, excludedPlugins, maxDepth, ComputeEpoch(_names, _paths, _mtimes, excludedPlugins));
     }
 
     /// <summary>The epoch fingerprint: a compact, deterministic identity for ONE index build, derived from the
--- a/src/housecarl-core/NpcAppearanceCopy.cs
+++ b/src/housecarl-core/NpcAppearanceCopy.cs
@@ -505,7 +505,7 @@ public static class NpcAppearanceCopy
             // unopenable-master skip and the same residual refusal; without the caller's cause it reported an unnamed
             // engine fault while the identical failure from apply/create/forward named the plugin (PR #315 review).
             // Injected rather than reached for: this type takes a mastersFor closure, not a resolver session.
-            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}{serializeCause?.Invoke(ex) ?? ""}. Nothing usable was written."); }
+            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}.{serializeCause?.Invoke(ex) ?? ""} Nothing usable was written."); }
 
             // ---- post-commit read-back — the patch IS on disk from here; a read-back failure is a WARNING on a
             //      success, never a "nothing was written" (review finding: that mislabel invites a duplicate re-run).
--- a/src/housecarl-core/NpcAppearanceCopy.cs
+++ b/src/housecarl-core/NpcAppearanceCopy.cs
@@ -320,7 +320,8 @@ public static class NpcAppearanceCopy
         ActiveResolve resolvesActively,
         string outPath, bool extend,
         Func<string, IReadOnlyList<ISkyrimModGetter>> mastersFor,
-        string donorReadFrom, bool donorOutOfLoadOrder)
+        string donorReadFrom, bool donorOutOfLoadOrder,
+        Func<Exception, string>? serializeCause = null)
     {
         var patchFileName = Path.GetFileName(outPath);
         try
@@ -500,7 +501,11 @@ public static class NpcAppearanceCopy
 
             // ---- serialize (multi-master; the caller's mastersFor handles the active-patch self-lock) ----
             try { WriteEngine.WritePatch(patchMod, mastersFor(patchFileName), outPath); }
-            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}. Nothing usable was written."); }
+            // #314 — this lane serializes through the caller's AllMastersExcept too, so it is subject to the same
+            // unopenable-master skip and the same residual refusal; without the caller's cause it reported an unnamed
+            // engine fault while the identical failure from apply/create/forward named the plugin (PR #315 review).
+            // Injected rather than reached for: this type takes a mastersFor closure, not a resolver session.
+            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}{serializeCause?.Invoke(ex) ?? ""}. Nothing usable was written."); }
 
             // ---- post-commit read-back — the patch IS on disk from here; a read-back failure is a WARNING on a
             //      success, never a "nothing was written" (review finding: that mislabel invites a duplicate re-run).
--- a/src/housecarl-core/NpcAppearanceCopy.cs
+++ b/src/housecarl-core/NpcAppearanceCopy.cs
@@ -321,7 +321,7 @@ public static class NpcAppearanceCopy
         string outPath, bool extend,
         Func<string, IReadOnlyList<ISkyrimModGetter>> mastersFor,
         string donorReadFrom, bool donorOutOfLoadOrder,
-        Func<Exception, string>? serializeCause = null)
+        Func<Exception, string>? serializeFailure = null)
     {
         var patchFileName = Path.GetFileName(outPath);
         try
@@ -505,7 +505,16 @@ public static class NpcAppearanceCopy
             // unopenable-master skip and the same residual refusal; without the caller's cause it reported an unnamed
             // engine fault while the identical failure from apply/create/forward named the plugin (PR #315 review).
             // Injected rather than reached for: this type takes a mastersFor closure, not a resolver session.
-            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}.{serializeCause?.Invoke(ex) ?? ""} Nothing usable was written."); }
+            //
+            // An OVERRIDE, not a suffix (PR #315 review 4): a suffix left the baseline refusal rendered behind
+            // "serialize failed" (it threw while the master-set ARGUMENT was built, so serialize never started) with
+            // the write status stated twice and a doubled period — the exact three complaints review 3 fixed
+            // everywhere else, surviving here because this was the one lane routed around SerializeFailure.
+            catch (Exception ex)
+            {
+                return NpcCopyOutcome.Fail(serializeFailure?.Invoke(ex)
+                    ?? $"serialize failed — {WriteEngine.Describe(ex)}. Nothing usable was written.");
+            }
 
             // ---- post-commit read-back — the patch IS on disk from here; a read-back failure is a WARNING on a
             //      success, never a "nothing was written" (review finding: that mislabel invites a duplicate re-run).
--- a/src/housecarl-core/WritePatchBuilder.cs
+++ b/src/housecarl-core/WritePatchBuilder.cs
@@ -559,7 +559,7 @@ public static class WritePatchBuilder
         session.ReleaseOverlay(patchMod.ModKey.FileName.String);
         try { WriteEngine.WritePatch(patchMod, session.AllMastersExcept(patchMod.ModKey.FileName.String), outPath); }
         catch (Exception ex)
-            { return PatchOutcome.Fail($"writing the patch failed (serialize or commit; the existing file is untouched): {WriteEngine.Describe(ex)}{UnopenableMasterClause(ex, session)}"); }
+            { return PatchOutcome.Fail(SerializeFailure("writing the patch failed (serialize or commit; the existing file is untouched): ", ex, session)); }
 
         // --- Phase 5: re-open the written patch and report its master header — and, on request, each touched
         //     record's FULL read-back off that same re-opened file (the on-disk bytes, not the in-memory mod — the
@@ -934,7 +934,7 @@ public static class WritePatchBuilder
                   "Enable that plugin in MO2 (or reference an active one) and retry. The existing file is untouched.");
         }
         catch (Exception ex)
-            { return PatchOutcome.Fail($"writing '{fileName}' in place failed (serialize or commit; the existing file is untouched): {WriteEngine.Describe(ex)}{UnopenableMasterClause(ex, session)}"); }
+            { return PatchOutcome.Fail(SerializeFailure($"writing '{fileName}' in place failed (serialize or commit; the existing file is untouched): ", ex, session)); }
 
         // --- Phase 5: re-open the now-edited file and report its master header — and the touched-record verify (default
         //     ON for in-place): each edited record read back IN FULL off the on-disk bytes (the model-C substitute for
@@ -1222,7 +1222,7 @@ public static class WritePatchBuilder
         // keeps the target out of the master set. (writelock-probe / writelock-apply-probe; both halves guarded.)
         session.ReleaseOverlay(patchMod.ModKey.FileName.String);
         try { WriteEngine.WritePatch(patchMod, session.AllMastersExcept(patchMod.ModKey.FileName.String), outPath); }
-        catch (Exception ex) { return RemovalOutcome.Fail($"writing the patch after removal failed (serialize or commit; the existing file is untouched): {WriteEngine.Describe(ex)}{UnopenableMasterClause(ex, session)}"); }
+        catch (Exception ex) { return RemovalOutcome.Fail(SerializeFailure("writing the patch after removal failed (serialize or commit; the existing file is untouched): ", ex, session)); }
 
         // Re-open: report the (possibly shrunk) master header + how many records remain (0 ⇒ the patch is now an inert
         // header-only plugin the user can disable/delete). Dispose the overlay so the file isn't left mmap'd for a later call.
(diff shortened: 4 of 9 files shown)


src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -929,7 +929,7 @@ static PatchOutcome ApplyInPlaceCore(
                 "Enable that plugin in MO2 (or reference an active one) and retry. The existing file is untouched.");
         }
         catch (Exception ex)
-            { return PatchOutcome.Fail($"writing '{fileName}' in place failed (serialize or commit; the existing file is untouched): {WriteEngine.Describe(ex)}"); }
+            { return PatchOutcome.Fail($"writing '{fileName}' in place failed (serialize or commit; the existing file is untouched): {WriteEngine.Describe(ex)}{UnopenableMasterClause(ex, session)}"); }
```
Avick3110: **[correctness] The two in-place lanes never reach this clause — their `MissingModException` arm catches the residual first, and says the one thing this PR calls wrong.**

`ApplyInPlaceCore` catches `MissingModException` at line 924, ahead of this generic catch; `ForwardRecordsInPlaceCore` has the twin at 1545.

Scenario: active order contains an unopenable `Broken.esp`; user runs `apply` (or `forward`) with `in_place=true` on `MyPatch.esp`, and the header needs two or more masters including `Broken.esp`. `WriteInPlace` throws `MissingModException` naming `Broken.esp` → caught at 924 → rendered as *"the edited records reference a plugin that is NOT active in the load order … Enable that plugin in MO2 and retry"*. `Broken.esp` **is** active and enabled, so the user is sent to fix a non-problem, and `UnopenableMasterClause` is never consulted. That is precisely the wording the PR description and the clause's own doc identify as wrong for this case.

The new guard cannot catch it: every write arm in `excluded-master-guard` goes through the new-patch lane (`patch:`), which has no `MissingModException` arm. Suggest calling the clause inside the two `MissingModException` arms too (or matching there first and falling through), plus an in-place arm in the probe.

(Also relevant to the #313 rebase note: once forward's patch lane gains a named `MissingModException` arm, the same bypass appears there.)

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1330,7 +1330,7 @@ static RemovalOutcome RemoveRecordsInPlaceCore(
             if (missing is not null) return RemovalOutcome.Fail(missing);
             try { WriteEngine.WriteInPlace(targetMod, ownMasters, targetPath); }
             catch (Exception ex)
-                { return RemovalOutcome.Fail($"writing '{fileName}' in place after removal failed (serialize or commit; the existing file is untouched): {WriteEngine.Describe(ex)}"); }
+                { return RemovalOutcome.Fail($"writing '{fileName}' in place after removal failed (serialize or commit; the existing file is untouched): {WriteEngine.Describe(ex)}{UnopenableMasterClause(ex, session)}"); }
```
Avick3110: **[correctness] In this lane the clause is dead code, and the lane's real #314 exposure still escapes uncaught.**

`RemoveRecordsInPlaceCore` never calls `AllMasters`/`AllMastersExcept` — it resolves the target's own declared masters via `ResolveOwnMasters`. So `session.SkippedUnopenable` is always empty here and `UnopenableMasterClause` can only ever return `""`.

Worse, the failure it appears to cover is still unhandled. `ResolveOwnMasters` opens each declared master with a bare `SkyrimMod.CreateFromBinaryOverlay` (line 1053), and that call sits **outside** any catch: the `try` at 1327 has only a `finally` (1335), and `RemoveRecordsInPlace` adds none.

Scenario: user removes a record in place from `MyPatch.esp` whose header declares the unopenable `Broken.esp` as a master → `CreateFromBinaryOverlay(brokenPath)` throws `ModGroupsMalformedException` → the throw escapes the core as an unhandled exception (opaque transport error), not a Q3 refusal. That is the same failure class #314 filed, surviving in the one write lane the fix leaves untouched — while the added clause on this line reads as coverage.

At minimum: wrap `ResolveOwnMasters` so an unopenable declared master becomes a named refusal (the `Unopenable` set now makes it identifiable before opening). `RemapEngine.cs:905` repeats the same bare-open pattern.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1572,6 +1572,28 @@ static ForwardOutcome ForwardRecordsInPlaceCore(
...
+    /// and the generic "serialize or commit" wording reads as a disk fault. Empty string when this isn't that case, so
+    /// callers can append unconditionally.
+    /// <para>Matched on the exception TEXT naming one of the skipped plugins, which is why it is a suffix rather than a
+    /// replacement: Mutagen's message is the only thing that knows WHICH master the sort wanted, and a mismatch must
+    /// degrade to the ordinary description rather than to a confident wrong cause.</para></summary>
+    static string UnopenableMasterClause(Exception ex, LoadOrderResolver.OverlaySession session)
+    {
+        if (session.SkippedUnopenable.Count == 0) return "";
+        var hit = session.SkippedUnopenable
+            .Where(n => ex.Message.Contains(n, StringComparison.OrdinalIgnoreCase))
```
Avick3110: **[correctness] Bare substring match can name a plugin that had nothing to do with the failure.**

`Contains(n)` on the raw message matches when a skipped name is a *substring* of the plugin the serializer actually wanted.

Scenario: `Foo.esp` is active-but-unopenable (skipped), and the write genuinely references an inactive `MyFoo.esp`. `MissingModException` names `MyFoo.esp`; `"MyFoo.esp".Contains("Foo.esp")` is true, so the render appends *"CAUSE: 'Foo.esp' is ACTIVE in your load order but cannot be opened … Repair or remove that plugin"* on top of a description naming the real culprit. Skyrim plugin names are heavily prefixed/suffixed, so this is not exotic.

That is the "confident wrong cause" the method's own doc says the design degrades away from. A delimiter/word-boundary match (or, better, matching the exception's `ModKey` where the type exposes one) removes the class.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1572,6 +1572,28 @@ static ForwardOutcome ForwardRecordsInPlaceCore(
...
+    /// <summary>#314 — the NAMED cause when a serialize failed because the header needed a master this session skipped
+    /// as unopenable. The plugin IS active, so the pre-existing "not active in the load order" wording is wrong for it,
+    /// and the generic "serialize or commit" wording reads as a disk fault. Empty string when this isn't that case, so
+    /// callers can append unconditionally.
+    /// <para>Matched on the exception TEXT naming one of the skipped plugins, which is why it is a suffix rather than a
+    /// replacement: Mutagen's message is the only thing that knows WHICH master the sort wanted, and a mismatch must
+    /// degrade to the ordinary description rather than to a confident wrong cause.</para></summary>
+    static string UnopenableMasterClause(Exception ex, LoadOrderResolver.OverlaySession session)
+    {
+        if (session.SkippedUnopenable.Count == 0) return "";
```
Avick3110, line 1666: **[correctness, plausible] The clause reads `ex.Message` only, while the render beside it reads the inner chain too.**

`WriteEngine.Describe(ex)` appends `[inner: <Type>: <msg>]`, so the description can name the skipped plugin via the inner exception while this match — `ex.Message` alone — sees nothing and returns `""`.

This codebase already documents that Mutagen wraps serialize throws: `RootNullArm`'s doc records a doubly-nested `AggregateException` around a `SubrecordException` from the parallel record-write path. If the `MissingModException` naming `Broken.esp` arrives wrapped in an exception whose own `Message` doesn't repeat the inner text, the user gets `"…: SubrecordException: <ctx> [inner: MissingModException: Broken.esp]"` with **no** CAUSE — the exact failure the machinery exists for, degraded to the generic "serialize or commit" fault.

Walking the `InnerException` chain (as `RootNullArm` already does) makes the match agree with what is actually rendered.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1572,6 +1572,28 @@ static ForwardOutcome ForwardRecordsInPlaceCore(
...
 
+
+    /// <summary>#314 — the NAMED cause when a serialize failed because the header needed a master this session skipped
+    /// as unopenable. The plugin IS active, so the pre-existing "not active in the load order" wording is wrong for it,
+    /// and the generic "serialize or commit" wording reads as a disk fault. Empty string when this isn't that case, so
+    /// callers can append unconditionally.
+    /// <para>Matched on the exception TEXT naming one of the skipped plugins, which is why it is a suffix rather than a
+    /// replacement: Mutagen's message is the only thing that knows WHICH master the sort wanted, and a mismatch must
+    /// degrade to the ordinary description rather than to a confident wrong cause.</para></summary>
+    static string UnopenableMasterClause(Exception ex, LoadOrderResolver.OverlaySession session)
```
Avick3110: **[correctness] One more write lane serializes through `AllMastersExcept` and gets no CAUSE.**

`copy_npc_appearance` hands `pf => { session.ReleaseOverlay(pf); return session.AllMastersExcept(pf); }` to `NpcAppearanceCopy.BuildAndWrite` (`LoadOrderService.cs:5863`), so it is subject to the same skip and the same residual multi-master refusal — but its serialize catch (`NpcAppearanceCopy.cs:503`) renders `"serialize failed — {Describe(ex)}. Nothing usable was written."` with no clause.

Scenario: order contains an unopenable `Broken.esp`; a copy into a new patch whose header needs `Skyrim.esm` plus a record originating in `Broken.esp` refuses with `MissingModException`, and the user sees an unnamed engine fault — while the identical failure from `apply`/`forward`/`create` names the cause. The PR's "appended at every serialize-failure render that has a session" isn't met.

(Making the clause `internal` and calling it from `NpcAppearanceCopy` is the cheap fix.)

src/housecarl-generator/Program.cs
```diff
@@ -19,6 +19,7 @@
 if (args.Length > 0 && CiAll.TryDispatch(args[0], args[1..], out var ciRc)) return ciRc;
 
 // Maintenance diagnostic: re-verify the mutable-collection whitelist on a Mutagen bump.
+if (args.Length > 0 && args[0] == "excluded-master-guard") return ExcludedMasterWriteProbe.RunGuard(args[1..]);
```
Avick3110: **[dead code] Unreachable — and it re-introduces the pattern the comment three lines up forbids.**

`CiAll.TryDispatch` (line 19) iterates `CiAll.Probes`, which now contains `("excluded-master-guard", ExcludedMasterWriteProbe.RunGuard)`, and returns on a name match — so control never reaches this line.

The comment at 16–18 says single-probe runs dispatch through `CiAll.Probes` *"so a guard can't be runnable locally yet missing from CI (the Q3 coverage-gap class)"*. A duplicate explicit dispatch is exactly what that convention rules out: remove the probe from `Probes` and it would still appear to work locally while silently dropping out of `ci-all`.

It is also inserted between the `// Maintenance diagnostic: re-verify the mutable-collection whitelist on a Mutagen bump.` comment and the `vocab` line that comment describes, so the comment now reads as describing this line.

Suggest deleting the line.

src/housecarl-generator/ExcludedMasterWriteProbe.cs
```diff
@@ -0,0 +1,210 @@
...
+                && MastersLineOf(needsIt).Contains(BrokenName, StringComparison.OrdinalIgnoreCase), needsIt);
+
+            // …and the ORDER of a multi-master header survives the skip: a patch touching a record from the master AND
+            // one originating in the broken plugin must list them in LOAD ORDER (master first), or the plugin is
+            // malformed in a way nothing else in this probe would catch.
+            var both = ForwardTools.Forward(fx.Svc, formids: new[] { fx.SubjectFid, fx.BrokenOwnFid }, source: CleanName, patch: "X314Both");
+            var line = MastersLineOf(both);
+            int iMaster = line.IndexOf(MasterName, StringComparison.OrdinalIgnoreCase);
+            int iBroken = line.IndexOf(BrokenName, StringComparison.OrdinalIgnoreCase);
+            if (!both.StartsWith("error:"))
```
Avick3110: **[test coverage] This arm passes whichever branch the serializer takes, so a behaviour flip is invisible.**

The `if`/`else` asserts header ordering on success and the CAUSE clause on failure, but nothing asserts *which* outcome is expected. Today the `else` is live (that's what earned the clause). If a Mutagen bump — or a regression in the skip — makes the two-master write succeed, the probe takes the `if`, asserts ordering, and reports PASS, while `UnopenableMasterClause` ships with **zero** CI coverage and its later breakage goes unnoticed. The reverse flip is equally silent.

Given the PR's own argument that the machinery came back *because a test earned it*, pinning the currently-observed branch (assert the refusal; treat an unexpected success as a FAIL that says "upstream behaviour changed — re-derive the residual") keeps that argument true over time.

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -104,13 +104,20 @@ internal sealed class IndexSnapshot   // internal (not private) so IndexView's c
         public readonly Dictionary<FormKey, int[]> Overriders;                // MULTI keys only — ordered touching overlay indices
         public readonly List<string> LoadFailures;                            // per-plugin index-build failures (open OR parse), surfaced (Q3)
         public readonly HashSet<int> Excluded;                                // overlay indices excluded this build — never re-touched by any path
+        /// <summary>#314 — the SUBSET of <see cref="Excluded"/> whose file could not be OPENED at all, as opposed to
+        /// opening fine and failing on a record body. The distinction is invisible to reads (both are "never touch
+        /// this plugin") and load-bearing for WRITES: the master set retains excluded plugins on purpose, which is
+        /// only possible for the ones that can be opened. Kept as its own set rather than derived from the reason
+        /// STRING — a message is display prose that can be reworded, membership is a fact.</summary>
+        public readonly HashSet<int> Unopenable;
```
Avick3110, line 137: **[correctness] The `Unopenable` fact is available here, but `dry_run` still predicts success for the write that now refuses.**

`DryRunMastersPreview` (WritePatchBuilder.cs:1015) tests referenced plugins against `resolver.PluginNames`, which includes unopenable plugins — they are active, just unreadable.

Scenario: order has an unopenable `Broken.esp`. `housecarl_apply … dry_run=true` on a patch whose header would need `Skyrim.esm` + `Broken.esp` passes the membership test, returns success, and lists `Broken.esp` among the would-be masters. The identical non-dry call fails with the new CAUSE refusal.

That breaks the contract stated at WritePatchBuilder.cs:974 — *"a dry run must never say 'would apply' about a write that would fail"* — which the #225 dry-run parity guard exists to hold. Pre-fix the parity was broken more widely (every write failed), so this is a narrowed residual rather than a regression; but the snapshot now carries exactly the fact needed to refuse in the dry lane with the same named cause.

Avick3110:
## Review folded — all 8 accepted, in `fd03ab1`

**Five of these were defects in the machinery this PR added**, and they were found by running the guard and probing the behaviour rather than reading the diff. Worth saying plainly: the fix shipped a smaller version of the bug it was fixing — an unnamed or wrongly-named failure — in four places I'd claimed were covered.

| # | disposition |
|---|---|
| in-place lanes never reach the clause | **accepted** — both `MissingModException` arms now prefer the named cause |
| remove-in-place: dead clause, real exposure uncaught | **accepted** — guarded before the open; dead clause removed |
| bare substring match | **accepted** — matched on the typed `ModKey` now |
| reads `ex.Message` only | **accepted** — whole chain walked, aggregates included |
| `copy_npc_appearance` gets no clause | **accepted** — takes an injected cause |
| unreachable `Program.cs` dispatch | **accepted** — deleted |
| the arm passes whichever branch runs | **accepted** — pinned |
| `dry_run` predicts success | **accepted** — predicts the refusal |

### The two matching findings are closed by construction, not patched

`MissingModException` exposes typed `ModPaths`. So the match is on the **ModKey**, walking the full inner chain including `AggregateException` branches — no message parsing at all. That removes both the `MyFoo.esp`/`Foo.esp` class and the wrapped-exception class outright rather than hardening a substring test against them.

One more of the same kind that neither finding named: `SkippedUnopenable` was an `IReadOnlyCollection<string>`, so `.Contains` bound to LINQ's **ordinal** comparison and would have missed a case difference between the profile's spelling and a ModKey's. It is an `IReadOnlySet<string>` now, so membership uses the set's `OrdinalIgnoreCase` comparer.

### The remove-in-place finding was the most serious

You're right that the clause there was dead *and* that it read as coverage over a live hole. `ResolveOwnMasters` opens each declared master with a bare `CreateFromBinaryOverlay` outside every catch — so #314's own failure class survived, as an **unhandled exception**, in the one write lane the skip doesn't reach. Now asked before opening via `IndexView.IsUnopenable`, with a named refusal, plus a catch for the master that opened at index time and has changed since. Pinned by its own arm.

I did **not** touch `RemapEngine.cs:905`, which you noted repeats the pattern. It's the compact/merge dependency scan, a different lane with its own error contract, and I'd rather not widen a fix PR into it on a drive-by reading — flagging it here so it isn't lost.

### `dry_run`, and the threshold question

Fixed, and the interesting part is *how*. The predictor needs a threshold — one master writes, two or more refuses — and that threshold is an **empirical observation about Mutagen's header sort**, exactly the kind of thing that burned this PR's first draft. So rather than assert it in a comment, both sides are now pinned in the guard: the one-master case must write, the two-master case must refuse, and an unexpected success fails with *"upstream behaviour changed — re-derive the residual."* The prediction can't quietly diverge from reality without CI saying so.

That also answers the coverage finding: the branch-either-way arm is gone. Its replacement is the third "check that proves nothing" caught in this session — after the vacuous released-overlay arm and a `str.replace` that silently matched nothing. That's a pattern in how I write guards, not three coincidences.

### Verification

`excluded-master-guard` **8 → 13**. `ci-all` **117/117**.

RED-verified for the three behavioural folds — disable them and the in-place, remove-in-place and dry-run arms fail. **Honest gaps in that:** the two matching fixes are structural (a typed match has no substring bug to re-introduce), and **the `copy_npc_appearance` clause has no arm** — that lane needs a donor NPC fixture, which this probe's order doesn't have. Reviewed by reading, not by test; say the word if you want the fixture built.

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -213,11 +234,43 @@ public IReadOnlyList<ISkyrimModGetter> AllMastersExcept(string excludeFileName)
         {
             var list = new List<ISkyrimModGetter>(_r._paths.Length);
             for (int i = 0; i < _r._paths.Length; i++)
-                if (!string.Equals(_r._names[i], excludeFileName, StringComparison.OrdinalIgnoreCase))
-                    list.Add(Overlay(i));   // SKIP the target's index — never map the file we're about to overwrite
+            {
+                if (string.Equals(_r._names[i], excludeFileName, StringComparison.OrdinalIgnoreCase)) continue;  // never map the file we're about to overwrite
+                if (SkipUnopenable(i)) continue;                                                                 // #314 — and never try to map one that cannot be opened
```
Avick3110, line 264: **[correctness] The skip silently drops a BaselineMaster from the force-include.**

`WriteEngine.WritePatch` computes its CK-mandated baseline force-include *from the list this method returns*:

```csharp
var baseline = BaselineMasters.Where(bm => ordered.Any(km => km.ModKey == bm)).ToArray();   // WriteEngine.cs:1462
```

The filter's stated purpose is to tolerate "a degenerate order (or a minimal single-master test harness)". After this change, a real order with an unopenable `Update.esm` *looks* degenerate to it.

Scenario: `Update.esm` is active but momentarily unopenable — the transient class this file's own epoch comment names (xEdit/MO2 holding a handle, an AV scan). A user runs `housecarl_create` for a self-contained Keyword. `AllMastersExcept` skips `Update.esm`; the baseline filter drops it; the derived header is `{Skyrim.esm}` — one master, so per this PR's own empirical threshold it **writes** — and a plugin lands on disk missing a baseline master Aaron locked as mandatory (2026-06-02), with no warning at all. Before this PR that write threw loudly at `Overlay()`. That is a silent degradation where there used to be a loud failure (Q3).

The dry run disagrees with the real call in the same scenario, in the opposite direction: `DryRunMastersPreview` adds baselines by *load-order membership* (`priority.ContainsKey`, WritePatchBuilder.cs:1031), never by openability, so `set = {Skyrim.esm, Update.esm}`, `set.Count > 1`, and it **refuses** the write the real call silently completes — a #225 parity break the `excluded-master-guard` fixture can't see, because it has no baseline masters at all.

Cheapest fix that keeps the fix's shape: treat a *baseline* master in `SkippedUnopenable` as a refusal rather than a skip (it can never be legitimately absent from a real order), and mirror that in the dry-run predictor.


---
*Body restored from `c1.md`, which this comment was posted as a literal path to (`gh --body @file` instead of `--body-file`). Content is the file verbatim; only this footer was added.*

src/housecarl-generator/ExcludedMasterWriteProbe.cs
```diff
@@ -0,0 +1,263 @@
...
+            var needsIt = ForwardTools.Forward(fx.Svc, formids: new[] { fx.BrokenOwnFid }, source: CleanName, patch: "X314Need");
+            Check("a patch whose record ORIGINATES in the unopenable plugin still WRITES, and still masters on it",
+                !needsIt.StartsWith("error:")
+                && MastersLineOf(needsIt).Contains(BrokenName, StringComparison.OrdinalIgnoreCase), needsIt);
+
+            // …and the ORDER of a multi-master header survives the skip: a patch touching a record from the master AND
+            // one originating in the broken plugin must list them in LOAD ORDER (master first), or the plugin is
+            // malformed in a way nothing else in this probe would catch.
+            var both = ForwardTools.Forward(fx.Svc, formids: new[] { fx.SubjectFid, fx.BrokenOwnFid }, source: CleanName, patch: "X314Both");
+            var line = MastersLineOf(both);
```
Avick3110: **[test-coverage] Dead locals, and the ordering assertion the comment above still advertises no longer exists.**

`line`, `iMaster` and `iBroken` are computed here and never read again — the only assertions made on `both` are `StartsWith("error:")` and the name/remedy contains. Neither `IndexOf` result feeds anything.

The comment at lines 116–118 still says the arm checks that a multi-master header "must list them in LOAD ORDER (master first), or the plugin is malformed in a way nothing else in this probe would catch". It doesn't. A future reader will believe header ordering is pinned by this guard when it isn't, and the leftover locals make the gap look covered.

Either drop the three locals and re-word the comment to describe what the arm actually pins (the refusal), or restore the ordering assertion on the one-master arm where a header *is* emitted.


---
*Body restored from `c2.md`, which this comment was posted as a literal path to (`gh --body @file` instead of `--body-file`). Content is the file verbatim; only this footer was added.*

src/housecarl-generator/ExcludedMasterWriteProbe.cs
```diff
@@ -0,0 +1,263 @@
...
+    static string Trim(string s) => s.Length <= 400 ? s.Replace("\n", " | ") : s[..400].Replace("\n", " | ") + " …";
+
+    /// <summary>The render's "masters:" line alone — a whole-render search would false-match the plugin names that
+    /// appear in the per-record rows, which is the assertion mistake this codebase has already paid for twice.</summary>
+    static string MastersLineOf(string render)
+        => render.Split('\n').FirstOrDefault(l => l.StartsWith("masters:", StringComparison.Ordinal)) ?? "";
+
+    /// <summary>The written artifact's path out of a write render (mod folder + filename), so a following arm can
+    /// address it by name — e.g. as an in_place target.</summary>
+    static string? ArtifactPathFrom(Fixture fx, string render)
```
Avick3110: **[dead-code] `ArtifactPathFrom` is never called, and `Fixture.ModsDir` exists only to serve it.**

No call site anywhere in the probe. Its doc comment ("so a following arm can address it by name — e.g. as an in_place target") describes an arm that was never written: the in-place arms address `CleanName` directly, not a written artifact. `Fixture.ModsDir` has no other consumer either.


---
*Body restored from `c3.md`, which this comment was posted as a literal path to (`gh --body @file` instead of `--body-file`). Content is the file verbatim; only this footer was added.*

src/housecarl-core/NpcAppearanceCopy.cs
```diff
@@ -500,7 +501,11 @@ public static NpcCopyOutcome BuildAndWrite(
 
             // ---- serialize (multi-master; the caller's mastersFor handles the active-patch self-lock) ----
             try { WriteEngine.WritePatch(patchMod, mastersFor(patchFileName), outPath); }
-            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}. Nothing usable was written."); }
+            // #314 — this lane serializes through the caller's AllMastersExcept too, so it is subject to the same
+            // unopenable-master skip and the same residual refusal; without the caller's cause it reported an unnamed
+            // engine fault while the identical failure from apply/create/forward named the plugin (PR #315 review).
+            // Injected rather than reached for: this type takes a mastersFor closure, not a resolver session.
+            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}{serializeCause?.Invoke(ex) ?? ""}. Nothing usable was written."); }
```
Avick3110: **[nit/correctness] Doubled period — this is the one lane where the clause isn't string-final.**

`UnopenableMasterClause` already ends with a full stop ("…writes that do NOT reference their records are unaffected."), and this format string appends `". Nothing usable was written."` immediately after it, so the render reads `…are unaffected.. Nothing usable was written.`

Every sibling lane appends the clause at end-of-string, so none of them show it. Suggest dropping the leading period here:

```suggestion
            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}{serializeCause?.Invoke(ex) ?? "."} Nothing usable was written."); }
```


---
*Body restored from `c4.md`, which this comment was posted as a literal path to (`gh --body @file` instead of `--body-file`). Content is the file verbatim; only this footer was added.*

Avick3110: Correction to my own suggestion above — it's wrong. `?? "."` only fires when `serializeCause` is *null*; in the normal case the delegate is non-null and returns `""`, which would drop the period entirely. Move the period ahead of the clause instead:

```suggestion
            catch (Exception ex) { return NpcCopyOutcome.Fail($"serialize failed — {WriteEngine.Describe(ex)}.{serializeCause?.Invoke(ex) ?? ""} Nothing usable was written."); }
```


---
*Body restored from `c4fix.md`, which this comment was posted as a literal path to (`gh --body @file` instead of `--body-file`). Content is the file verbatim; only this footer was added.*

Avick3110:
## Code review — 2 further findings, in files outside this diff

Both are the *same failure class this PR fixes*, in write lanes the fix didn't reach. Posting here rather than inline because GitHub can't anchor a comment to an unchanged file.

---

### 1. `src/housecarl-core/RemapEngine.cs:906` — `RepointInPlace`'s declared-master open is still unguarded [medium]

```csharp
var ov = SkyrimMod.CreateFromBinaryOverlay(mpath, SkyrimRelease.SkyrimSE);   // RemapEngine.cs:906
```

Its own comment (line ~889) says it "mirrors `WritePatchBuilder.ResolveOwnMasters`, which opens them the same way" — and `ResolveOwnMasters` is exactly what this PR just guarded. So the new comment's claim that it was *"the one write lane the skip does not reach"* isn't right: this is a second one, and `CompactBuild` (below) is a third.

Scenario: a truncated `B.esp` is active. The user compacts `C.esp` with `acknowledge=true` and external referencers present. Step 6 (`LoadOrderService.cs:5350`) writes the compacted plugin **successfully**. Then the un-try/caught `foreach` at `LoadOrderService.cs:5362` calls `RepointInPlace` on a referencer declaring `B.esp` as a master; line 906 throws and the exception escapes the whole compact call. The `repointed` per-plugin results are discarded and steps 7b/7c (FaceGen carry, voice carry, SEQ gate) never run — *after* P′ is already on disk. Engine throw, not a Q3 refusal, on a half-completed compaction.

Fix is the same shape as the one you just applied: ask `view.IsUnopenable(mfn)` before the open, and wrap the open itself.

---

### 2. `src/housecarl-core/WritePatchBuilder.cs:2148` — `CompactBuild`'s declared-master open escapes past the caller's rider cleanup [medium]

```csharp
var mov = SkyrimMod.CreateFromBinaryOverlay(mp, SkyrimRelease.SkyrimSE);   // WritePatchBuilder.cs:2148
```

Its sibling `MergeBuild` wraps the identical call in a try/catch at line 2258, with a comment naming precisely this risk: *"an open throw escaping here would skip the caller's rider-folder cleanup"*. `CompactBuild` has the same caller shape and no such catch.

Scenario: truncated `B.esp` active; user compacts `C.esp` (which declares `B.esp`) into a **new** mod folder. `ResolvePatchModFolder` has already created the rider folder (`LoadOrderService.cs:5342`, `createdFresh = true`). Line 2148 throws, escaping `CompactBuild` instead of returning `CompactBuildResult.Fail`, so `if (!build.Success) RemoveOrNameRiderResidue(rf)` at `LoadOrderService.cs:5353` never runs — an orphan empty houseCARL mod folder is left in the MO2 mods directory, plus an unnamed engine throw.

---

### Reviewed and found sound

- The `Unopenable`-as-a-set (not a substring of the reason string) design, and the `MissingModException.ModPaths` typed match — verified `get_ModPaths` is the only typed accessor on that type in Mutagen 0.53.1 (no `ModKeys` property), so there's no second shape the walk silently misses.
- The `set.Count > 1` dry-run threshold matches the real header size for both lanes — `WritePatch` force-includes baselines into the emitted list and `DryRunMastersPreview` mirrors that, `WriteInPlace` adds none and `patchLane: false` mirrors *that*. (The one case where they diverge is the baseline finding filed inline.)
- `ResolveOwnMasters`' early returns don't leak overlays — `RemoveRecordsInPlaceCore`'s `finally` at line 1382 disposes `masterOverlays` on every path.
- The `AllMasters()` array → `List` change breaks no positional consumer; both callers (`MultiMasterProof`, `RemoveProof`) pass it straight to `WritePatch`.
- Skipping an index doesn't perturb master *ordering* for the survivors — only relative position in the supplied list matters.

🤖 `/code-review high` — 6 findings total (4 inline above, 2 here).


---
*Body restored from `gen.md`, which this comment was posted as a literal path to (`gh --body @file` instead of `--body-file`). Content is the file verbatim; only this footer was added.*

Avick3110:
## Review round 2 folded — all 6 accepted, in `ac4c458`

**First, a delivery problem worth knowing about:** every comment in this round posted as the literal string `@C:/Users/compl/AppData/Local/Temp/c1.md` — `gh` was given `--body @file` rather than `--body-file`, so the findings never reached GitHub. I read them off disk. The PR record currently holds six comments that are just a path; the substance is below so the thread isn't empty, but the original wording is only in those temp files.

---

### The baseline finding is the most serious thing anyone has caught in either PR

It is exactly right, and it is a defect **this PR introduced**: I traded a loud failure for a silent wrong artifact, which is the one trade Q3 never permits.

`WritePatch` derives the CK-mandated force-include *from the list the master-set builders return*, filtered to baselines present in it — a filter whose own comment says it exists to tolerate "a degenerate order (or a minimal single-master test harness)". Skipping an unopenable `Skyrim.esm` makes a **real** order look degenerate to it. Proven, not reasoned — with the fix disabled, the guard's self-contained create now reports:

```
wrote BlCreate.esp (new patch; 101 bytes) … masters: (none)
```

A plugin on disk with no masters at all, where before this PR the same call threw.

A baseline is never legitimately absent, so it **refuses** now rather than skipping — thrown from the single chokepoint every write lane funnels through, so no lane can forget the check, and recognised by the clause so the generic "serialize or commit" lead-in never stands alone in front of it. The dry-run predictor mirrors it **without** the multi-master threshold, because the force-include is mandatory however small the header — and, as you noted, is absent from the referenced set entirely on precisely the self-contained create that exposes this.

Its own fixture, since the main order has no baselines to break: a truncated `Skyrim.esm` plus a clean override. Both directions pinned.

### The two out-of-diff findings — and last round's claim was wrong

You're right that *"the one write lane the skip does not reach"* was false; there were three. I declined `RemapEngine` last round as a drive-by reading of another lane's error contract. With the consequence traced — the caller reaches `RepointInPlace` only **after** P′ is on disk, so the throw discards every repoint result and skips the FaceGen/voice/SEQ carry on a half-completed compaction — that was the wrong call. Pre-checked and wrapped.

`CompactBuild` is wrapped like its `MergeBuild` sibling, whose catch already names this exact risk. An asymmetry where one twin documents the hazard and the other doesn't is worth more than its severity suggests.

### The rest

- **Ordering assertion** — right, and it's the same class as last round's: the pinning rewrite replaced the assertion and left the prose claiming it, over three locals nothing read. A comment describing coverage that doesn't exist is worse than no comment. Reworded to what the arm actually pins.
- **`ArtifactPathFrom` / `Fixture.ModsDir`** — dead, removed. Both existed for an in-place arm I rewrote to address the plugin directly.
- **Doubled period** — took your **corrected** suggestion. Worth noting you caught your own error there: `?? "."` fires only when the delegate is null, which is the case that never happens in production.

### Verification

`excluded-master-guard` **13 → 15**. `ci-all` **117/117**. The baseline fold is RED-verified (output above). `RemapEngine`/`CompactBuild` are **not** guarded — both need a compact/merge fixture with an unopenable declared master, which this probe's order isn't shaped for; reviewed by reading, same as the `copy_npc_appearance` clause from last round. Three unguarded fixes is more than I'd like on one PR, and I'd rather say so than let the 15/15 imply otherwise.

plugin/CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ when it changes.
...
 *Accumulating notes for the next cut — not yet released; `plugin.json` still reads the last shipped version.*
 
+**Fixed: one broken plugin in your load order no longer breaks every write.** If a plugin houseCARL cannot open
+sat in your active order — a truncated download, a half-copied mod folder, an interrupted xEdit save — then
+*every* write failed, including writes that had nothing to do with it, with an opaque message that read like a
+disk fault. Reads were never affected. Now writes that don't reference that plugin's records just work.
+
+A write that genuinely *does* need it (its records can still be reached through a plugin that overrides them) is
+refused **naming it as the cause**, with what to do — instead of the previous engine-level error. One such write
+now succeeds that didn't before: a patch whose header needs only that single master.
```
Avick3110: **[correctness / medium]** The user-facing claim here — *"One such write now succeeds that didn't before: a patch whose header needs only that single master"* — is **unreachable in any real load order**, and the guard only sees it succeed because its fixture has no baselines.

`WriteEngine.WritePatch` force-includes `BaselineMasters` filtered to those present in the order (`WriteEngine.cs:1462`), so a patch-lane header always carries ≥2 entries once Skyrim.esm/Update.esm are in the order — i.e. always — and therefore always has to be **sorted**, which is exactly the residual this PR documents.

Verified empirically. I rebuilt the guard's fixture with openable `Skyrim.esm` + `Update.esm` in the order alongside a truncated `SxBroken.esp` and an `SxClean.esp` overriding its record, then forwarded the record that **originates** in `SxBroken.esp` — the changelog's "now succeeds" case:

```
error: writing the patch failed (serialize or commit; the existing file is untouched):
MissingModException: A referenced mod was not present on the load order being sorted
against: SxBroken.esp. … CAUSE: 'SxBroken.esp' is ACTIVE in your load order but cannot
be opened by houseCARL …
```

The unrelated-write arm in the same run **did** write (`masters: Skyrim.esm, Update.esm`), so the headline fix is real and confirmed — it's only this second sentence that overpromises.

`excluded-master-guard`'s arm *"a patch whose record ORIGINATES in the unopenable plugin still WRITES"* passes because `HcXMaster`/`HcXClean`/`HcXBroken` carry no baselines to force-include, leaving a genuinely single-entry header. The same overstatement is carried in `LoadOrderResolver.SkipUnopenable`'s XML doc (first bullet) and makes the low side of `DryRunMastersPreview`'s `set.Count > 1` threshold dead in practice.

Suggest either adding a baseline-bearing arm to the guard that pins the refusal and dropping the sentence, or scoping it to "a patch whose only master is that plugin (an order without the baselines)".

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1572,6 +1633,53 @@ static ForwardOutcome ForwardRecordsInPlaceCore(
...
+    /// wrong cause this is supposed to degrade away from; and reading only <c>ex.Message</c> would miss a
+    /// <c>MissingModException</c> that arrives WRAPPED, which this codebase already documents happens (see
+    /// <see cref="RootNullArm"/>'s doubly-nested AggregateException). The whole chain is walked, aggregates
+    /// included.</para></summary>
+    public static string UnopenableMasterClause(Exception ex, LoadOrderResolver.OverlaySession session)
+    {
+        // A baseline refusal is thrown, not skipped, so it can arrive with SkippedUnopenable still empty — check it
+        // first or the "serialize or commit failed" lead-in would stand alone in front of it.
+        for (Exception? b = ex; b is not null; b = b.InnerException)
+            if (b is UnopenableBaselineMasterException ub) return " CAUSE: " + ub.Message;
```
Avick3110: **[correctness / low]** For the BASELINE class this renders the whole ~60-word refusal **twice**, behind a lead-in that is wrong for it.

`WriteEngine.Describe(ex)` already prints `"{TypeName}: {ex.Message}"`, and `UnopenableBaselineMasterException.Message` *is* the full refusal — so appending `" CAUSE: " + ub.Message` duplicates it verbatim. Actual output, from running `excluded-master-guard` with the `got:` strings printed:

```
error: writing the patch after create failed (serialize or commit; the existing file is
untouched): UnopenableBaselineMasterException: 'Skyrim.esm' is a BASELINE master (every
written plugin must list it) and is ACTIVE in your load order, but houseCARL cannot open
it — see load_order_status for the reason. No plugin can be written until it is repaired
or replaced: writing one without it would produce a plugin the game treats as malformed.
Nothing was written. CAUSE: 'Skyrim.esm' is a BASELINE master (every written plugin must
list it) and is ACTIVE in your load order, but houseCARL cannot open it — see
load_order_status for the reason. No plugin can be written until it is repaired or
replaced: writing one without it would produce a plugin the game treats as malformed.
Nothing was written.
```

The guard passes because it asserts `Contains("BASELINE master")`, which a double render satisfies.

Two further inaccuracies in that same lead-in, both specific to this arm: the throw comes from `SkipUnopenable` while `AllMastersExcept` builds the *argument list*, so **serialize never started** ("serialize or commit" misdirects), and this is a **new** patch, so there is no "existing file" to be untouched.

The non-baseline arm is right as-is — `MissingModException`'s own message names the plugin but not the cause, so appending there genuinely adds information. Only the baseline branch needs to stop appending: return `""` when `Describe(ex)` will already carry `ub.Message`, or have the branch signal substitution rather than append.

Avick3110:
## Review round 3 folded — both accepted, in `630b842`

### The changelog was overpromising a case that cannot happen

You're right, and the empirical half is what makes it undeniable: I rebuilt your order into the guard as `RealOrderArm` — baselines present and openable, a truncated `SxBroken.esp`, a clean override — and it reproduces exactly what you found. The unrelated write lands with `masters: Skyrim.esm, Update.esm`; the write referencing the broken plugin's record **refuses**.

The mechanism is the one you named: `WritePatch` force-includes the baselines when present, so a real patch-lane header always carries ≥2 entries, always has to be sorted, and therefore always takes the residual. The single-master success exists only in a **baseline-less harness** — which is precisely what my original fixture was. A fixture that couldn't represent the common case taught me a claim about it.

Corrected in all four places it had spread to: the changelog sentence is gone, `SkipUnopenable`'s doc bullet now marks the single-master case harness-only, the main fixture's arm is relabelled as the baseline-less order, and the PR body carries the correction rather than a quiet edit.

The `set.Count > 1` threshold **stays**, deliberately: it keeps the dry run agreeing with the real call in *both* worlds, and its low side is now documented as harness-only rather than implied to be something a user can hit. Dead in practice is not the same as wrong, and removing it would make the two disagree in the fixture the guard actually runs.

### The doubled render

Accepted exactly as described, including both lead-in inaccuracies. `SerializeFailure` now **substitutes** the exception's own message for the whole render on that branch instead of appending it, and the clause returns `""` for the baseline class — which also keeps the doubling out of `copy_npc_appearance`, the one lane that renders the clause without going through the new helper.

Your observation that the guard passed on a doubled render is the part worth keeping. It now **counts** occurrences and asserts the lead-in is absent:

```
occurrences=2 :: error: writing the patch after create failed (serialize or commit; the
existing file is untouched): UnopenableBaselineMasterException: 'Skyrim.esm' is a BASELINE …
```

— that's the RED run with the fix reverted, i.e. the arm now fails on precisely the output you pasted.

That is the **fourth** check-that-proves-nothing in this session, and they are all the same shape: a `Contains` standing in for an assertion about the message. The others were a vacuous released-overlay check, a branch that passed either way, and a `str.replace` that silently matched nothing. I'd rather name the pattern than keep fixing instances of it.

### Verification

`excluded-master-guard` **15 → 18**, `ci-all` **117/117**, both folds RED-verified.

Unchanged from last round, and still true: `RemapEngine`, `CompactBuild` and the `copy_npc_appearance` clause have no guard arms — each needs a fixture shape this probe isn't built for.

Avick3110:
## Review — `fix: an UNOPENABLE active plugin no longer breaks every write`

*Conductor-side review from the working session (`/review #315`) — not the independent `/code-review` pass.*

### What it does

Splits `BuildIndex`'s single `Excluded` set into two facts: excluded-because-unparseable (retained in the master sets, as before) and excluded-because-`OpenOverlay`-threw (new `Unopenable` set, skipped). The old retention comment's safety argument — "`Overlay()` opens lazily → no throw" — was true of the first class and false of the second, so every write in an order containing one unopenable plugin died in `AllMasters`/`AllMastersExcept` before it started. The skip is centralised in `SkipUnopenable`, which also refuses outright (`UnopenableBaselineMasterException`) when the unopenable plugin is `Skyrim.esm`/`Update.esm`, since those are force-included and dropping one would silently emit a master-less plugin. Residual failures (sorted multi-master headers) get a named CAUSE appended; the three lanes that open declared masters directly (`ResolveOwnMasters`, `RepointInPlace`, `CompactBuild`) get their own pre-checks and wrapping.

**Verdict: sound, and the diagnosis is right.** The exclusion-class split is the correct primitive — a fact, not a substring of a display string, as the comment says. Matching the CAUSE on `MissingModException.ModPaths` rather than message text is the right call given Skyrim's prefix-heavy plugin names. The baseline refusal is the piece that turns this from "a fix" into "a fix that doesn't trade a loud failure for a silent one", and it's the piece most PRs of this shape miss.

Verified independently: `AllMasters()` has no production callers (only `MultiMasterProof`/`RemoveProof`), so the array→`List` change is safe; every `AllMastersExcept` call site sits inside a `catch`, so `UnopenableBaselineMasterException` cannot escape; `OverlaySession` is documented single-call/single-thread, so the unsynchronised `SortedSet` is fine; both `"is NOT active in"` MissingModException arms were updated and the third occurrence (`WritePatchBuilder.cs:1019`) is the genuinely-inactive dry-run arm, correctly untouched.

---

### Findings

**1. The npc-copy lane still renders the baseline refusal with the wrong lead and a doubled tail** *(low — the exact defect review 3 fixed everywhere else)*

`UnopenableMasterClause` deliberately returns `""` for the baseline class, and `SerializeFailure` — which does the substitution — is `static` private, so npc-copy can't use it. The result at `src/housecarl-core/NpcAppearanceCopy.cs:504`:

```
serialize failed — UnopenableBaselineMasterException: 'Skyrim.esm' is a BASELINE master … Nothing was written.. Nothing usable was written.
```

Wrong phase ("serialize failed" — it threw building the argument), doubled period, and the write-status claim stated twice. That's the same three complaints review 3 made about `SerializeFailure`, surviving in the one lane routed around it. Cheapest fix: make `SerializeFailure` internal and have `LoadOrderService.cs:5864` pass the whole render rather than just the clause, changing `serializeCause` from "suffix" to "override". `BaselineArm` has no npc-copy arm, so nothing catches this.

**2. `CollectMissingMods` double-visits the first aggregate branch** *(nit)*

`AggregateException.InnerException` *is* `InnerExceptions[0]`, so the aggregate loop and the tail `CollectMissingMods(ex.InnerException, …)` both walk it — each visit burning a depth level, which roughly halves the effective depth-8 cap along aggregate chains. Harmless today (the `SortedSet` dedupes, and the degrade is graceful), but `if (ex is not AggregateException) CollectMissingMods(ex.InnerException, …)` costs one line.

**3. Redundant scan in the dry-run baseline check** *(nit)*

`resolver.PluginNames.Contains(bm.FileName.String, StringComparer.OrdinalIgnoreCase) && resolver.IsUnopenable(…)` — `IsUnopenable` already returns false for a name absent from `_nameToIdx`, so the guard adds only an O(n) LINQ scan per baseline over every plugin name, on every dry run.

**4. The baseline message states a patch-lane reason on the in-place lane** *(nit, Q3-visible text)*

`"is a BASELINE master (every written plugin must list it)"` — but `WriteInPlace` deliberately force-includes none (`DryRunMastersPreview`'s own doc says so). The *refusal* is still correct for in-place (an unopenable declared master can't resolve either way), but the reason given isn't the reason that applies.

**5. `DryRunMastersPreview` reads the fact off an unpinned snapshot** *(observation, not a regression)*

It uses `resolver.IsUnopenable` (current `_snap`) while the write lanes resolve against a pinned `IndexView`. A rebuild between prediction and write lets the two disagree — the exact class the codebase's hunt-F5 discipline ("the write path resolves every edit of one call off ONE view") exists to prevent. The method already takes a resolver rather than a view, so this is inherited shape, but it's the one place the new fact is read unpinned and worth a line acknowledging it.

**6. Layering** *(design note — I'd keep it as is)*

`LoadOrderResolver` now imports `WriteEngine.BaselineMasters` and throws a write-policy exception. The chokepoint argument ("no lane can forget the check") is stronger than the purity argument and I agree with the trade, but it's the first write dependency in that file — worth a line in the class header so a later tidy-up doesn't "fix" it back out.

**7. PR description is stale** *(housekeeping)*

Body says "Eight arms" and "5 of the 8 fail" with the skip disabled. The probe now carries ~18 `Check` calls across three fixtures (main + `BaselineArm` + `RealOrderArm`), the latter two added by reviews 2 and 3. Since the PR body is the review record and the closing-keyword artefact, worth reconciling before merge.

---

### Test coverage

Strong, and unusually honest about it. The things normally worth flagging are already handled:

- Both sides of the `>1` master threshold are **pinned as facts**, not branched — the failing arm carries an explicit "UPSTREAM BEHAVIOUR CHANGED, re-derive the residual" message naming the two consumers that rest on it. That's what licenses the dry-run predictor to use the threshold at all.
- `RealOrderArm` exists specifically because the original fixture's baseline-less order produced a single-master success that cannot happen in a user's order — and the main fixture's arm now labels itself as the harness case rather than quietly implying reality.
- The `occurrences == 1` count in `BaselineArm` catches doubling that a `Contains()` would pass.
- `MastersLineOf` scopes the master assertion to the `masters:` line instead of the whole render.
- The excluded-source refusal is asserted to *still* fail, so the skip can't become an escape hatch.

Gaps: no npc-copy arm (finding 1), and the `>1` threshold is pinned only in the patch lane — the in-place lane's single-master case is predicted but unverified.

### Conventions / risk

Changelog lands under `## Unreleased`, closing keyword present, probe registered in `CiAll`, commentary carries the usual issue/review provenance. No security surface. Perf impact is nil in the normal case (`Unopenable` is empty; one `HashSet` lookup per index). The blast radius is real — this is the seam every write funnels through — which makes the unchanged 116-probe suite the load-bearing part of the safety argument, as the PR says.

**Recommendation:** fold finding 1 (with a `BaselineArm` npc-copy arm), take 2 and 3 as one-liners, and refresh the PR body. 4–6 are optional. Nothing here blocks the approach.

Avick3110:
## Review round 4 folded — all 7 accepted, in `228642e`

Including 4–6, which you marked optional. Three of them were wording or a doc line; the fourth (layering) is the kind of note that stops a future tidy-up from silently undoing a deliberate trade, which is worth more than its size.

### 1 — npc-copy, the same three complaints surviving in the one lane routed around the fix

Correct, and the diagnosis of *why* is the useful part: that lane took a **suffix**, so review 3's substitution couldn't reach it. It takes an **override** now — the same `SerializeFailure`, with the lane's tail passed as a `trailer` the baseline branch drops (its message already states the write status).

One correction to the suggestion: `internal` wouldn't compile. The call site is in `housecarl-mcp` and core grants `InternalsVisibleTo` only to `housecarl-generator`, so it's `public`, like `UnopenableMasterClause` beside it.

**And the arm** — you were right that its absence is why nothing caught this. `BaselineArm` now clones a donor NPC through that lane and pins the refusal once, the dropped trailer, no doubled period, and the absent phase. RED-verified.

My first draft of that arm counted the word `"Nothing"`, which the reworded refusal legitimately uses twice — it failed on correct output. A count is only as good as the token counted; noting it because it's the same family as the `Contains` problem, just one step further along.

### 2, 3 — taken as the one-liners they are

`AggregateException.InnerException` **is** `InnerExceptions[0]`, so the tail visit is skipped for aggregates now. And `IsUnopenable` already answers false for a name the order lacks, so the membership pre-test was pure cost.

### 4 — the reason didn't hold on the lane it was shown on

Right, and it's a Q3-visible text defect rather than a nit in my view: the refusal is correct on in-place, but *"every written plugin must list it"* is the patch lane's reason and `WriteInPlace` force-includes none. Reworded to the fact that covers both — no write can resolve references against a master it cannot read — with the header requirement kept as the *additional* patch-lane consequence.

### 5, 6 — recorded in place

The unpinned-snapshot read is now a note at the call site naming it as the hunt-F5 class, inherited shape, failure mode a stale prediction and never a bad write. The layering note is in the class header: this file carries a write policy deliberately, on the chokepoint argument over the purity one.

### 7 — body refreshed

Now describes the three fixtures and the nineteen checks, and states what is **not** guarded rather than leaving 19/19 to imply otherwise.

---

`excluded-master-guard` **18 → 19**, `ci-all` **117/117**.

Gaps unchanged and still stated: `RemapEngine.RepointInPlace` and `CompactBuild` have no arms (each needs a compact/merge fixture with an unopenable declared master), and the in-place lane's single-master threshold case is predicted but unverified.

Avick3110:
## Re-review — after `fold PR #315 review 4` (`228642e2`)

*Conductor-side, same session as the previous comment. Read-only pass over the fold diff and the branch head; I did not build or run `ci-all`, so the 117/117 and 19-arm claims are taken as stated.*

### The fold

All 7 folded, and two of them better than proposed:

- **npc-copy (finding 1)** — turned into a genuine override rather than a bigger suffix, with `trailer` dropped on the baseline branch so the doubled status can't come back. And **`public`, not the `internal` I suggested** — correct: the call site is in `housecarl-mcp` and core grants `InternalsVisibleTo` only to `housecarl-generator`, so my suggestion wouldn't have compiled. Good catch against the review.
- **The arm that was missing** — `BaselineArm` now drives the lane and pins the three things substitution actually guarantees (refusal once, trailer dropped, phase absent). The noted false-failure from counting `"Nothing"` is the right lesson to have written down: the reworded refusal legitimately uses the word twice.
- **`CollectMissingMods`** — `return` after the aggregate loop is the correct shape; `InnerExceptions[0]` is now walked once and the depth cap means what it says.
- **Findings 3, 5, 6, 7** — folded as asked. The PR body rewrite is a real improvement: it now states the *unguarded* gaps outright (`RepointInPlace`, `CompactBuild`, the in-place single-master threshold) instead of letting the arm list imply coverage.

Nothing in the fold regressed the core fix. The exclusion-class split, the chokepoint refusal, and the pinned `>1` threshold are all untouched.

---

### New findings

**1. Finding 4 was folded in one of the two places it occurs — and the two now disagree** *(low, but it lands on #225 parity)*

`UnopenableBaselineMasterException`'s message was reworded to hold on both lanes. The **dry run's parallel string was not** — `WritePatchBuilder.DryRunMastersPreview` still returns:

> `'Skyrim.esm' is a BASELINE master (every written plugin must list it) and is ACTIVE in your load order … No plugin can be written until it is repaired or replaced. Nothing was written.`

That check sits outside `if (patchLane)`, so an **in-place** dry run reaches it and gives the patch-lane-only reason — exactly what finding 4 was about. Two knock-ons:

- The dry run and the real call now say materially different things about the same condition, where before the fold they were near-identical. The reworded exception also gained *"Repair or replace that plugin in MO2 and retry"*; the dry run still omits the remedy.
- `BaselineArm`'s dry-run assertion is `Contains("BASELINE master")` on both sides, so it cannot see the divergence — the same shape as the `Contains()` weakness review 3 fixed by counting.

Fix: hoist the shared text. Constructing an `UnopenableBaselineMasterException(name)` in the dry-run branch and returning `"dry run caught what the real write would fail on: " + ex.Message` keeps the two provably identical, and lets the arm assert that identity instead of a substring.

**2. The injected npc-copy renderer drops the sentence break the fallback keeps** *(cosmetic)*

`SerializeFailure("serialize failed — ", ex, session, " Nothing usable was written.")` renders `lead + Describe(ex) + clause + trailer` — no terminator after `Describe(ex)`. The in-file fallback one line below still has one (`{Describe(ex)}. Nothing usable…`). On the non-baseline path that reads:

```
serialize failed — MissingModException: <message> Nothing usable was written.
```

The sibling lanes end without a period because nothing follows them; this one has a following sentence. I assume the period was dropped to satisfy the new `!npc.Contains("..")` assertion — but that assertion only runs on the baseline branch, where the whole render is substituted, so it never constrains this path either way. Either terminate the lead-in or start the trailer with something that reads as a break.

**3. Tiny: the reworded exception asserts both conjuncts** *(nit)*

*"no write can resolve references against a master it cannot read, **and** a new patch must list the baselines"* — for the self-contained create that `BaselineArm` tests, the first conjunct doesn't apply (nothing is referenced) and the second carries it. Reads as a joint justification where it's really an either/or. Accurate enough that I'd leave it if nothing else is being touched.

---

### Verdict

The fold is clean and the review-against-the-review on `internal`/`public` and on the `"Nothing"` count is the kind of thing that makes the next pass cheaper. **Finding 1 above is worth folding** — it's small, it's the second half of a finding already accepted, and hoisting the string closes it by construction rather than by keeping two prose blocks in sync. 2 and 3 are optional.

Nothing blocks merge on my side.

Avick3110:
## Re-review folded — all 3, in `141fff7`

Including 2 and 3, which you marked optional.

### 1 — the half of finding 4 I missed

Correct, and the diagnosis of why it slipped is the useful part: I reworded the exception and never looked for its paraphrase. Two prose blocks stating the same condition is a standing invitation to exactly this.

Taken your fix — **by construction, not by rewording twice**. The dry-run branch now builds an `UnopenableBaselineMasterException(name)` and returns its `Message`, so the real call and the prediction are the same string and cannot diverge again.

And you were right that the arm couldn't have caught it: `Contains("BASELINE master")` on both sides is satisfied by any two texts that share the phrase. It now asserts the dry-run text contains the real call's message **verbatim** — the assertion the hoist makes possible, and the one that would have failed on the divergence you found. RED-verified by re-paraphrasing the dry-run string: the arm fails.

### 2 — right about the shape, wrong about the cause, and the cause is why it matters

The run-on is real. But your guess that I dropped the period to satisfy `!Contains("..")` isn't it — as you note yourself, that assertion only runs on the baseline branch where the whole render is substituted, so it never constrained this path.

The actual reason is more interesting, and it's why a "just add the period back" fix would have been wrong: **a fixed trailer can't be correct for both branches.** The clause already ends in a full stop when it fires; `Describe(ex)` doesn't when it doesn't. A leading period doubles in the first case; its absence runs on in the second. `SerializeFailure` now guarantees exactly one terminator before a lane's tail, whichever way the body ended.

So: cosmetic in symptom, but the fix had to be conditional, and that's the kind of thing that comes back as a `..` bug six months later.

### 3 — folded, and it corrects two places at once

"For either reason alone" now. Because it lives in the hoisted string, the dry run gets the same correction — which is the small dividend of fix 1.

---

`excluded-master-guard` **19** (count unchanged; one arm strengthened from substring to verbatim), `ci-all` **117/117**.

### Standing gaps, unchanged

`RemapEngine.RepointInPlace`, `CompactBuild`, and the in-place single-master threshold are still proven only by reading. **Aaron has chartered that as its own follow-up session** rather than growing this PR further — it's written up in the dev backlog with the three in build order, the fixture technique, and the RED-verify step. Recorded here so the gap doesn't quietly become permanent: this session has twice had a declared gap turn out to be exactly where a defect was.

PR #318, opened 2026-08-07
2.0 tidy-up sweep: guard the #315 fixes, close CopyFrom's arm drift, give write_seq output_dir=

### LINKED ISSUE #312: [GAP] write_seq cannot target a mod's own folder — no output_dir=, so an in-place edit's .seq always lands in a temp houseCARL mod
> *Filed from an external live-session gap report (in-place dialogue editing pass); triaged, root-caused against source, and posted by AI.*

### Goal

Regenerate the `.seq` after **in-place** ESP edits on a mod that already ships its own `Seq\` folder — i.e. land the regenerated file at `<mod folder>\Seq\<plugin>.seq`, where the mod's own copy already lives and where the mod's own convention keeps it.

### What fell short

`housecarl_write_seq` can only write into a houseCARL-owned mod folder. With no `into=` it mints a fresh `houseCARL - houseCARL_SEQ_NNN` folder on every run; `into=` accepts only an **existing houseCARL patch** mod, not the target mod's own folder and not an arbitrary output path. There is no `output_dir=`.

    housecarl_write_seq source="<mod folder>\<plugin>.esp"     # no into= — twice in one session
    -> "path: ...\houseCARL - houseCARL_SEQ_001\SEQ\<plugin>.seq"
       "the .seq is in a houseCARL mod folder — enable it in MO2"

So every regeneration became: run the tool → hash-compare the temp output against the mod's live `Seq\` copy → copy over if changed → delete the temp mod folder.

### Workaround used

Manual hash-compare plus a `Remove-Item` of the temp folder. Both runs that session were byte-identical to the live `.seq`, so nothing needed copying — the entire folder-cut was churn.

### Frequency

Every in-place dialogue edit. The reporter's project rule regenerates the `.seq` after each one — 3 in-place edit sessions in a single day.

### Ideal shape

`into=` (or a new `out_dir=`) accepting the plugin's own mod folder, writing `<mod>\SEQ\<plugin>.seq` directly — ideally with a "byte-identical to existing, nothing written" short-circuit note, which would also make the no-op case zero-file-churn.

- date: 2026-08-05  |  houseCARL version: 1.6

---

Valid, reproducible, non-duplicate, and **unchanged on `main`** — the W3 rewrite (#311) reworked `write_seq`'s SOURCE resolution but left its OUTPUT model exactly as reported.

Every claim checks out in `LoadOrderService.cs`:

- `into=` resolves through `ResolvePatchModFolder` (`:6510-6519`), whose `into=` branch calls `ResolveOwnedPatchFolder` — and that helper **refuses a folder houseCARL didn't create** (`:6497`, "ORIGINALS UNTOUCHED (Q3)"). A third-party mod's own folder is therefore rejected by construction, not by omission.
- With no `into=`/`patch=`, `WriteSeq` (`:6751-6755`) tries `OwnedPluginFolderStem` — which returns the plugin's own folder **only when that folder is houseCARL-owned**. For an in-place edit of a third-party mod it always returns null, so the fresh-folder branch fires every time.
- No byte-compare exists anywhere on the path: `AtomicFile.WriteAllBytes` runs unconditionally (`:6761`), and the folder is cut *before* the write. The reported churn is structural.

**The interesting part is the coherence gap, not the missing parameter.** houseCARL grew an opt-in in-place write lane for `.esp`s; `write_seq`'s output model never followed it. Its whole design assumes the plugin it serves lives in a houseCARL folder (that assumption is written into `OwnedPluginFolderStem`'s summary — "so the `.seq` defaults into the SAME folder as the `.esp`"). Edit in place and that pairing silently inverts: the `.esp` is in the mod's folder, the `.seq` lands in a different mod entirely, and it's on the user to reunite them. For a `.seq` specifically that's a live Q3 footgun — a `.seq` in an un-enabled or wrong folder leaves the quest **silently dead**, which is the exact failure `OwnedPluginFolderStem` was written to prevent for the default lane.

Fix direction — **parity with an existing decided contract, not a new escape hatch.** `compile_script` already has `output_dir=` (`CompileTools.cs:103, 192-201`) with the semantics this asks for: DECIDED contract (Aaron 2026-06-16), user names a mod-folder ROOT, houseCARL appends the deploy subfolder (`Scripts\`), `ResolvePatchModFolder` is *not* called so no houseCARL folder is cut, the target is USER-OWNED so `CreatedFresh=false` and residue cleanup never touches it, and a Q3 `deployWarning` fires when the result isn't somewhere the game actually loads from. `bsa_extract` carries the same parameter. `write_seq` is the odd one out.

So: `write_seq` gains `output_dir=` on the `ResolveExplicitScriptFolder` model — append `SEQ\` instead of `Scripts\`, same double-segment guard, same user-owned/no-cleanup posture, same deployability warning, same "output_dir= wins, `patch_name=`/`into=` are ignored" note. That leaves the ownership guard on `into=` fully intact (it should stay — loosening it is the wrong fix), and it does not touch the in-place consent gate: writing a *new sidecar file* into a mod folder is the `place_asset` posture, not an in-place record edit.

The byte-identical short-circuit is a second, smaller ask and worth taking with it: the bytes are already fully built in memory before the write (`built.Bytes`), so comparing against an existing destination is a few lines, and it turns the recurring no-op case into "nothing written" instead of a folder cut + file churn. Note it must report the no-op **explicitly** (Q3 — a silent skip is indistinguishable from a silent failure).

### LINKED ISSUE #316: [NOTE] Guard the three PR #315 fixes that shipped proven only by reading
Three fixes landed in PR #315 (`af84482`) with **no guard arm**. Each was reviewed by reading only, and stated as such on the PR rather than implied covered.

## Why this is filed rather than accepted

PR #315 made the same "reviewed by reading" declaration a fourth time, about the `copy_npc_appearance` clause — and its round-4 review found that lane was **wrong**: doubled tail, doubled period, a phase that never ran. Separately, a fixture that could not represent a baseline-bearing order let a false claim ship in the changelog until review 3 caught it.

**Both times the declared gap was where the defect was.** That is the argument for closing these, and it is drawn from this PR's own history rather than from a general preference for coverage.

## The three

1. **`CompactBuild`'s declared-master open** — `WritePatchBuilder.cs`, the `catch` added to mirror `MergeBuild`. Cheapest of the three. Compact a plugin whose header declares an unopenable master, into a **fresh** mod folder. Assert the named refusal **and** that no orphan rider folder survives — that cleanup is the entire reason the wrap exists (`RemoveOrNameRiderResidue`, `LoadOrderService.cs` ~5353, runs on a `Fail` return and never on a throw).
2. **`RemapEngine.RepointInPlace`** — `RemapEngine.cs` ~906, the `IsUnopenable` pre-check + wrap. **The one that matters most:** its caller reaches it only *after* P′ is already on disk, so a wrong refusal there is wrong at the worst possible moment — the throw discards every per-plugin repoint result and skips the facegen/voice/SEQ carry. Needs a compact with `acknowledge=true` and an external referencer that declares the broken plugin.
3. **The in-place lane's single-master threshold.** `DryRunMastersPreview(patchLane: false)` predicts it; nothing verifies it. The patch-lane side is pinned **both ways** in `excluded-master-guard`; this is its twin.

## Fixture technique (saves an hour)

An unopenable plugin **poisons every write in the order it sits in** — that is the bug #314 filed — so these want their own order, like `BaselineArm` / `RealOrderArm` in `ExcludedMasterWriteProbe`. Truncating a written plugin by ~12 bytes (`whole[..(whole.Length - 12)]`) reliably produces the open-failure exclusion class.

**RED-verify each:** disable the pre-check or wrap and confirm the arm fails with the pre-fix behaviour. Three of PR #315's arms were found to prove nothing precisely because that step was skipped or too loose.

Closes when all three are pinned and RED-verified. Chartered as part of the 2.0 tidy-up sweep.

### LINKED ISSUE #317: [Bug] CopyFrom's off-order source has the two-capture drift window forward just closed
`housecarl_apply`'s `CopyFrom` lane resolves an off-order source across **two separate captures**, with no cross-check between them. PR #313 closed the identical window in `housecarl_forward` and **deliberately left this one** — same change, different lane, its own PR.

## The window

`LoadOrderService.ResolveOffOrderCopySources` (~`LoadOrderService.cs:4546`) takes its own `Capture()` to decide the source is off-order and pre-fetch its bodies. `WritePatchBuilder.Apply` then takes **another** capture. `_snap` can be swapped by a concurrent read's freshness rebuild in between — including inside `_writeGate`, which serialises writes but not index rebuilds.

If the source becomes active in that gap, the pre-fetched off-order bodies are still used.

## Why it is a wrong body, not a wrong label

Enabling a mod doesn't change the file, so most of the time this is cosmetic. **A profile switch does** — it changes which mod folder serves a filename. So the copy that was located on disk can be a *genuinely different file* from the now-active plugin of the same name, and the write silently copies from the wrong one.

That is the reasoning that turned this from an observation into a fix on the `forward` side.

## The fix, as applied to `forward` (PR #313, `WritePatchBuilder.cs`)

`IsOffOrderSource` re-checks the arm against the **engine's** capture, not the pre-locate's:

```csharp
static bool IsOffOrderSource(OffOrderForwardSource? offOrder, ForwardSpec s, LoadOrderResolver.IndexView view) =>
    offOrder is not null
    && string.Equals(offOrder.Plugin, s.FromPlugin, StringComparison.OrdinalIgnoreCase)
    && !view.ContainsPlugin(s.FromPlugin);
```

The fresher build wins, the pre-fetched bodies go unused, and the in-order arm resolves it correctly. The off-order **report stamp** is separately gated on the arm actually taken (`ref bool usedOffOrder`, the same idiom as the epoch stamp), so the response can no longer claim an arm it did not use — the `CopyFrom` lane will need the equivalent for whatever it reports.

## Coverage note, stated up front

The `forward` fix has **no guard arm** — it is a race between two captures and could not be synthesised deterministically in-process. It was reviewed by reading and that was declared on PR #313. Expect the same constraint here; if anyone finds a way to pin it, that arm is worth more than the fix.

Chartered as part of the 2.0 tidy-up sweep.

Avick3110:
The 2.0 tidy-up sweep (Aaron-go 2026-08-07: *"next session tidies up everything left over so far that is related to 2.0"*). Three issues, one branch, three rounds of independent review folded before this PR was opened.

Closes #316 · Fixes #317 · Fixes #312

---

### #316 — guard the three PR #315 fixes that shipped proven only by READING

Each was reviewed by reading and declared as such on that PR. #315 made the same declaration a fourth time and its round-4 review found *that* lane was wrong; both times the declared gap was where the defect was. Now pinned, each in its OWN order (an unopenable plugin poisons every write in the order it sits in):

- **`CompactBuild`'s declared-master open** — the named refusal **and** that the fresh mod folder is gone. The folder is the point: `RemoveOrNameRiderResidue` runs on a `Fail` return and never on a throw. RED: without the catch the throw escapes `CompactPlugin` entirely.
- **`RemapEngine.RepointInPlace`** — the one that matters most, because its caller reaches it only *after* P′ is on disk. Driven through the tool, since the difference the fix makes is "a named per-plugin result" vs Guard's internal-failure wrapper. RED: the open throws out through `Guard.Tool` with P′ already rewritten and every repoint result discarded.
- **The in-place single-master threshold**, both ways plus its dry run — a one-master header writes even when that master is unopenable, a two-master header refuses by cause, and the predictor agrees with each. RED in both directions (dropping `set.Count > 1` → over-refusal; removing the clause → under-refusal).

Two of these arms were repaired by their own RED checks: one asserted a file SIZE that a renumber leaves identical, and the "external referencer" was first an override — which `HasExternalReferencers` does not count, so the repoint stage never ran and the arm was green against *"external referencers: none"*.

### #317 — the CopyFrom lane's off-order arm

Fixed as filed (`TryOffOrderCopyBody`, mirroring #313's `IsOffOrderSource`) — **and then the premise turned out to be wrong**, which review round 1 caught. Membership lives in `_nameToIdx`, built once per resolver **instance** and never rebuilt (`RefreshIfStale` swaps only `_snap`), and a write reads `Resolver` once and hands that instance to both the pre-locate and the engine. So the mid-call race #317 *and PR #313* describe cannot occur, and neither re-check can currently change the arm.

Both are kept and re-documented for what they are: a structural invariant (the arm follows the view the write resolves against), not a race repair. The deterministic half is now pinned — `CopyFromViewArm` hands the engine a pre-fetched body keyed to an **active** source and asserts the in-order arm wins, on three distinguishable values so no outcome is ambiguous.

### #312 — `write_seq` gains `output_dir=`, plus a byte-identical no-op

Parity with `compile_script`'s decided contract (Aaron 2026-06-16), through one shared body so the two lanes cannot drift; `into=`'s ownership guard is untouched. A destination already holding these exact bytes is left alone and reported `unchanged`. The `.seq` is a new sidecar file into a folder the user owns — the `place_asset` posture, not an in-place record edit.

## Review before the PR (Aaron's trial, 2026-08-07)

Three rounds, fresh unseeded agents each time, folded between rounds. Round 3 returned **no high and no medium findings**, which is what opened this PR. What the rounds actually caught:

| Round | Findings | Sharpest |
|---|---|---|
| 1 | 3 medium, 8 low | #317's premise is false — and so is #313's |
| 2 | **1 high (mine)**, 3 medium, 5 low | 13.5 MB of generated corpus swept into a fold commit by `git add -A` |
| 3 | lows only | strings that describe a rule the code no longer follows |

Three that changed the shipped behaviour rather than the prose:

- **The no-op broke `validate_dialogue`.** Its SEQ lint reads **mtime**, not content, so skipping the write after an in-place edit — the exact loop #312 exists for — left a permanent "your .seq is older than the plugin" advisory on a byte-perfect file that no houseCARL tool could clear. An identical-but-older `.seq` now has its timestamp stamped **past the plugin** and verified; a stamp that fails, or that cannot beat a future-dated plugin, falls through to the real write.
- **The deployability rule omitted MO2's overwrite folder** — which is mapped onto Data at top priority and is where xEdit/CK/Synthesis output lands, so the feature's headline case warned about a folder the game reads. Fixed on **both** lanes, which is a behaviour change for `compile_script`: it has its own changelog line and is now pinned both ways in `compile-ergonomics-guard`, which had only ever exercised the 3-arg overload.
- **The high was mine.** Two mistyped probe names (the generator treats an unrecognised first argument as an output directory and exits 0) produced two 6.6 MB corpus dumps that `git add -A` committed. Stripped by rebuilding the commit, so no blob ever reached a pushed branch — but the silent mode→output-dir fallback that made it possible is pre-existing and worth its own issue.

## Verification

`ci-all` **117/117** · `write-surface-guard` **119/119** · `excluded-master-guard` **28/28** · `seq-write-guard` (42 arms) and `compile-ergonomics-guard` green. Every new arm RED-verified against the code it covers.

## Stated, not fixed

- `written` in `write_seq`'s json changed meaning (now "this call wrote the file", not "a path exists") — documented on the renderer; no in-repo consumer.
- The `CopyFrom` lane still lacks `forward`'s `ActiveNameForPath` rule, so a `from_source=` **path** naming an active plugin routes off-order and is described as not in the load order — the same defect PR #313 fixed for `forward`. Named in the code; worth its own issue.
- `output_dir=` can overwrite a third-party mod's own `.seq` with no backup and no `acknowledge` gate. Now disclosed in the parameter description and reported as `replaced`, and it matches `compile_script`'s shipped precedent — but whether this lane deserves a handshake is a product call.
- An unusable `output_dir=` is not diagnosed on the no-SGE-quests path (that lane returns before any folder is resolved).

### DIFF
--- a/.claude/skills/dialogue-authoring/SKILL.md
+++ b/.claude/skills/dialogue-authoring/SKILL.md
@@ -220,6 +220,10 @@ this skill's job.
    starts. (A plugin with no such quests needs no `.seq`; the tool reports that.) If a later **in-place** edit
    or removal prunes a master, the on-disk FormIDs shift and the existing `.seq` goes stale — houseCARL flags
    this in the write's read-back note so you re-run `housecarl_write_seq` (it flags, never silently rewrites).
+   After an **in-place** edit the `.esp` is in the mod's own folder — pass `output_dir=` that mod folder so the
+   `.seq` lands beside it rather than in a separate houseCARL mod you then have to enable. Re-running against a
+   lane-named destination is free: a byte-identical `.seq` is left alone (reported `unchanged`). See
+   `references/seq-file-format.md`.
 
 7. **Validate, then verify.** Run `housecarl_validate_dialogue` on the topic (a DIAL FormID) or the whole
    quest (a QUST FormID). It checks what it can — quest/branch wiring, `LinkTo` and PNAM resolve, voice
--- a/.claude/skills/dialogue-authoring/references/seq-file-format.md
+++ b/.claude/skills/dialogue-authoring/references/seq-file-format.md
@@ -56,5 +56,17 @@ that single mod deploys `.esp` and `.seq` together), otherwise a fresh one you e
 After an **in-place** edit the `.esp` is in the *mod's own* folder, so pass `output_dir=` that mod folder
 and the `.seq` lands in its `SEQ\`, where the mod's own copy already lives. `output_dir=` wins over
 `patch=` / `into=` (the response says so), and it never touches the `.esp` — a `.seq` is a new sidecar
-file, not an in-place record edit. If the destination already holds exactly the bytes houseCARL would
-write, nothing is written and the response says `unchanged` — so re-running after every edit is free.
+file, not an in-place record edit.
+
+When the destination is one a **lane names** — `output_dir=`, `into=`, or the plugin's own houseCARL
+folder — and it already holds exactly the bytes houseCARL would write, nothing is written and the
+response says `unchanged`, so re-running after every edit costs nothing. (With *no* lane named the
+destination is a freshly cut folder, empty by construction, so that call always writes — and cuts
+another folder each time. Name a lane if you regenerate repeatedly.) A `.seq` that is byte-identical
+but older than the plugin has its timestamp stamped forward, because `housecarl_validate_dialogue`'s
+SEQ staleness check reads mtime — otherwise a skipped write would leave that check calling a correct
+file stale. (That check lints the `.seq` your load order *serves*, so the refresh clears it only when
+the folder you wrote to is the one that wins the `SEQ\` conflict.)
+
+If a `.seq` is already at the path, it is **overwritten with no backup** and the response says
+`replaced` rather than `wrote` — on `output_dir=` that file can be the mod's own shipped copy.
--- a/.claude/skills/dialogue-authoring/references/seq-file-format.md
+++ b/.claude/skills/dialogue-authoring/references/seq-file-format.md
@@ -47,3 +47,14 @@ quest is meant to be running from game start, the quest must be Start-Game-Enabl
 in the `.seq`. Set the flag (a quest-record field write), then run `housecarl_write_seq` against the
 plugin. The `.seq` makes the quest START; it does not verify the quest or its dialogue is otherwise
 correct — that is `housecarl_validate_dialogue`'s job.
+
+## Where it lands
+
+By default the `.seq` goes into a houseCARL mod folder — the plugin's own if it lives in one (so enabling
+that single mod deploys `.esp` and `.seq` together), otherwise a fresh one you enable in MO2.
+
+After an **in-place** edit the `.esp` is in the *mod's own* folder, so pass `output_dir=` that mod folder
+and the `.seq` lands in its `SEQ\`, where the mod's own copy already lives. `output_dir=` wins over
+`patch=` / `into=` (the response says so), and it never touches the `.esp` — a `.seq` is a new sidecar
+file, not an in-place record edit. If the destination already holds exactly the bytes houseCARL would
+write, nothing is written and the response says `unchanged` — so re-running after every edit is free.
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -19,10 +19,11 @@ mods folder, the overwrite folder, or the game's Data). It wins over `patch=`/`i
 quietly ignoring them. If a `.seq` was already at that path it is overwritten with no backup, and the response now
 says `replaced` rather than `wrote`.
 
-**The MO2 overwrite folder now counts as a place the game loads from.** `output_dir=` on both
-`housecarl_compile_script` and `housecarl_write_seq` warned that a file landing in your overwrite folder wouldn't
-deploy. It does — MO2 maps overwrite onto the game's Data at top priority, and it's where xEdit/CK/Synthesis
-output lands — so that warning was a false alarm. It no longer fires there. (The rule is otherwise unchanged and
+**The MO2 overwrite folder now counts as a place the game loads from.** `housecarl_compile_script`'s `output_dir=`
+warned that a `.pex` landing in your overwrite folder wouldn't deploy. It does — MO2 maps overwrite onto the game's
+Data at top priority, and it's where xEdit/CK/Synthesis output lands — so that warning was a false alarm. It no
+longer fires there, and `housecarl_write_seq`'s new `output_dir=` shares the corrected rule rather than inheriting
+the old one. (The rule is otherwise unchanged and
 still exact: a mod's own `Scripts\`/`SEQ\`, the overwrite folder, or the game's `Data\` — a *nested* path inside a
 mod still warns, because MO2 would deploy it to `Data\<Sub>\…` where nothing loads it.)
 
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -12,15 +12,26 @@ when it changes.
 plugin **in place** left the two halves in different mods: the `.esp` in the mod's own folder, the `.seq` in a
 freshly cut `houseCARL - houseCARL_SEQ_00N`, with `into=` unable to name a folder houseCARL didn't create. Every
 regeneration then meant hand-comparing the temp output against the mod's live `Seq\` copy and deleting the
-folder. `output_dir=` takes the mod-folder root and appends `SEQ\` — the same contract `housecarl_compile_script`
-and `housecarl_bsa_extract` already use, including the don't-double-it guard, no houseCARL folder cut, your
-folder never touched by cleanup, and a warning when the destination isn't somewhere the game actually reads SEQ
-files from. It wins over `patch=`/`into=`, and says so rather than quietly ignoring them.
+folder. `output_dir=` takes the mod-folder root and appends `SEQ\` — the same contract `housecarl_compile_script`'s
+`output_dir=` already uses, including the don't-double-it guard, no houseCARL folder cut, your folder never touched
+by cleanup, and a warning when the destination isn't somewhere the game actually reads SEQ files from (your MO2
+mods folder, the overwrite folder, or the game's Data). It wins over `patch=`/`into=`, and says so rather than
+quietly ignoring them. If a `.seq` was already at that path it is overwritten with no backup, and the response now
+says `replaced` rather than `wrote`.
 
-**…and regenerating an unchanged `.seq` now writes nothing.** If the destination already holds exactly the bytes
-houseCARL would write, the file is left alone and the response says `unchanged` (`written: false` in JSON) instead
-of reporting a write that only churned the file. Re-running after every in-place edit — which is the point of the
-tool — is now free.
+**The MO2 overwrite folder now counts as a place the game loads from.** `output_dir=` on both
+`housecarl_compile_script` and `housecarl_write_seq` warned that a file landing in your overwrite folder wouldn't
+deploy. It does — MO2 maps overwrite onto the game's Data at top priority, and it's where xEdit/CK/Synthesis
+output lands — so that warning was a false alarm. It no longer fires there. (The rule is otherwise unchanged and
+still exact: a mod's own `Scripts\`/`SEQ\`, the overwrite folder, or the game's `Data\` — a *nested* path inside a
+mod still warns, because MO2 would deploy it to `Data\<Sub>\…` where nothing loads it.)
+
+**…and regenerating an unchanged `.seq` now writes nothing.** When a lane names the destination (`output_dir=`,
+`into=`, or the plugin's own houseCARL folder) and it already holds exactly the bytes houseCARL would write, the
+file is left alone and the response says `unchanged` (`written: false` in JSON) instead of churning the file.
+Re-running after every in-place edit — which is the point of the tool — is now free. A skipped write still stamps
+the file's timestamp forward when it was older than the plugin, so `housecarl_validate_dialogue`'s SEQ staleness
+check keeps agreeing with it.
 
 **Fixed: one broken plugin in your load order no longer breaks every write.** If a plugin houseCARL cannot open
 sat in your active order — a truncated download, a half-copied mod folder, an interrupted xEdit save — then
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -23,7 +23,9 @@ says `replaced` rather than `wrote`.
 warned that a `.pex` landing in your overwrite folder wouldn't deploy. It does — MO2 maps overwrite onto the game's
 Data at top priority, and it's where xEdit/CK/Synthesis output lands — so that warning was a false alarm. It no
 longer fires there, and `housecarl_write_seq`'s new `output_dir=` shares the corrected rule rather than inheriting
-the old one. (The rule is otherwise unchanged and
+the old one. Sharing the rule also fixed a second `compile_script` bug on the way past: `output_dir=C:\` (a bare
+drive root) produced the drive-*relative* `C:Scripts`, so the `.pex` landed in whatever the working directory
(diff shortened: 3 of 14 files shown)


Avick3110:
## `/review` — conductor session, 2026-08-07

> Not the independent review (that is a session Aaron spawns himself). This is a `/review` run from the
> conductor's terminal, posted here at Aaron's instruction so the findings live with the PR.

Read: the full 1453-line diff across 14 files, plus the surrounding code for `ContainsPlugin`,
`ResolveOffOrderCopySources`, every `ScriptOutputContract` caller, `_overwriteDir`,
`ResolvePatchModFolder`, and `validate_dialogue`'s SEQ lint.

---

### Verified correct

- **`TryOffOrderCopyBody`'s gate matches the pre-locate exactly.** `LoadOrderService.cs:4563` reads
  `if (view.ContainsPlugin(e.FromPlugin)) continue;` — the new helper's `!view.ContainsPlugin(e.FromPlugin!)`
  is the same predicate, so neither lane changes behaviour today, *including* for excluded /
  unopenable-but-active plugins (both sides answer `true`, so the in-order arm is taken and the
  excluded-master refusal fires rather than a disk read). The PR's own "cannot currently change the arm"
  claim holds under checking.
- **The mtime contract lines up.** `DialogueValidate.cs:630` is `seq >= plugin`; `RefreshSeqTimestamp`
  treats `>=` as satisfied and verifies with `<`. Same comparator, both directions — the sibling-lint
  argument the fold rests on is sound.
- **`_overwriteDir` is safe in explicit mode** (`""` → `IsDataDeployFolder` short-circuits false), and
  `ResolveExplicitRiderFolder` is the **only** production caller of `ScriptOutputContract`, so the
  overwrite fix cannot miss a path.
- **Lock order preserved** (`_writeGate` → `_gate`); config gate intact (`ConfigPromptOrNull` plus
  `NotConfigured()` inside the rider resolver).
- **The reference doc's lane list is accurate.** `patch=` goes through `UniqueStem`
  (`LoadOrderService.cs:6699`), so it always cuts a fresh folder and genuinely cannot reach the no-op —
  the doc's omission of it is a fact, not a gap.

---

### Findings

**[low] A failed timestamp touch on identical bytes yields a misleading "replaced" alarm.**
`identical && !RefreshSeqTimestamp(...)` sets `identical = false` and falls through to the real write.
`Replaced = File.Exists(dest)` is then true, so the render says *"a .seq was already at that path and has
been OVERWRITTEN (no backup is kept — in a folder you named, that may have been the mod's own)"* — for a
file whose bytes were byte-identical and where nothing was lost. Reachable where `SetLastWriteTimeUtc`
succeeds but does not persist (some network shares) while the write itself succeeds. Cheap fix: carry the
known-identical fact into the outcome and soften the wording when the replaced bytes matched.

**[low] The no-SGE-quests early return reports `user_chose_output_dir: true` with no destination ever
resolved.** Disclosed in "Stated, not fixed" as *an unusable `output_dir=` is not diagnosed on the
no-SGE-quests path* — but the JSON shape makes it a little worse than that: the caller sees the lane
acknowledged, no `seq_path`, no `deploy_warning`, and no way to distinguish "your folder is fine, nothing
was needed" from "your folder is garbage, we never looked." A one-line note on that return closes it
without moving the resolve.

**[low] `TryOffOrderCopyBody` re-derives the pre-locate's rule instead of sharing it.**
The point of the fix is that the two captures agree; right now that agreement is enforced by two
independent copies of `!view.ContainsPlugin(FromPlugin)` in two projects (`housecarl-mcp` and
`housecarl-core`) plus a doc comment asking the reader to check. That is the hand-wiring shape §3 argues
against. If the pre-locate's gate ever gains a clause — an `ActiveNameForPath` rule being the obvious
candidate, which this PR itself names as the next fix in this lane — the engine's copy silently stops
matching, and `CopyFromViewArm` will not catch it, because the arm pins the *invariant* rather than the
*sharing*. Suggest one predicate on `IndexView` (or a static on `WritePatchBuilder`) that both sides call.

**[nit] The CHANGELOG under-reports the shared refactor.** *"(The rule is otherwise unchanged and still
exact: …)"* — but `SubfolderOutputContract`'s new `IsRoot` guard also fixes `compile_script`'s drive-root
bug (`C:\` previously produced the drive-**relative** `C:Scripts`). Small, real, and a second behaviour
change on the lane that got its own changelog line for the first one. `CompileErgonomicsProbe` pins it;
the changelog does not mention it.

**[nit] Two unpinned strings on the new lane.** The `output_dir` write-failure clause
`"(the '…' folder is left in place — houseCARL never removes a folder you named.)"` and JSON's
`replaced_note` have no arm. Everything else added here is pinned both ways, so these two stand out.

**[nit] `RepointArm`'s negative assertions are wording-coupled.** `!render.Contains("unexpectedly")` and
`!render.Contains("internal houseCARL failure")` will false-fail if unrelated copy elsewhere in the
compact render ever uses those words. The positive half (`"repointed in place"`) carries the real signal —
and the comment already records that the first draft of this clause was inert.

---

### Quality, conventions, risk

- **Q3 posture is consistently applied.** `unchanged` / `replaced` / `wrote` are three distinct headlines,
  the ignored-lane note now rides refusals on both transports, and the deploy warning fires for a `.seq`
  the engine will not read. The `SameBytesOnDisk` / `RefreshSeqTimestamp` "any failure falls through to the
  real write" direction is the correct fail-safe in both helpers.
- **Cornerstone check.** `output_dir=` is parity with an already-decided contract through one shared body,
  not a new verb; no bespoke bulk verb anywhere in the diff. Clean against §3.
- **Performance.** One extra `File.ReadAllBytes` of a few-hundred-byte file per `write_seq`; the no-op
  saves a folder cut per regeneration. Net win.
- **Security.** `output_dir=` writes anywhere the process can write, with no backup and no `acknowledge` —
  but that is `compile_script`'s shipped precedent, disclosed in the parameter description, and reported as
  `replaced`. Agreed with the PR that the handshake question is a product call for Aaron, not a defect.
- **Comment density** is high even by this repo's standards. `TryOffOrderCopyBody`'s two-paragraph
  correction and the mirrored correction on `IsOffOrderSource` say the same thing twice; one could point at
  the other.

Nothing here blocks the merge. The strongest suggestion is the shared-predicate one — it converts #317's
fix from "two places that currently agree" into "one place," which is what the doc comment is presently
doing by hand.

PR #323, opened 2026-08-11
fix: #321 CopyFrom source paths, #308 in-place verify honesty, #300 nested parent hosting (+ generator mode refusal)

### LINKED ISSUE #300: [Bug] nested PlacedObject create forwards the cell's load-order winner — unnecessary master plus a frozen copy of the winner's fields
### houseCARL version

1.9.5-dev

### Exact repro

A nested `PlacedObject` create under an existing CELL derives the patch's masters from the
cell's **load-order winner** and copies that winner's field values into the patch, rather
than building a lean override on the cell's **defining master**.

Target cell `0161F0:Skyrim.esm` (`MixwaterMillWorkersHouse`). Winner in the test order is a
lighting overhaul (`Lux.esp`), override_depth=3.

    housecarl_bulk_create
      patch_name: zzTest_DynamicBook
      records: [
        { record_type: "Book",      editorid: "Tutorial_DynamicBook" },
        { record_type: "Container", editorid: "Tutorial_DynamicEntriesContainer" },
        { record_type: "PlacedObject", editorid: "Tutorial_DynamicBookREF",
          parent: "0161F0:Skyrim.esm", collection: "Persistent",
          operations: [ { field_path: "Base", value: "@Tutorial_DynamicBook" },
                        { field_path: "Placement.Position", value: "820.0,-86.05238,2362.1572" } ] },
        { record_type: "PlacedObject", editorid: "Tutorial_ContainerREF",
          parent: "0161F0:Skyrim.esm", collection: "Persistent",
          operations: [ { field_path: "Base", value: "@Tutorial_DynamicEntriesContainer" },
                        { field_path: "Placement.Position", value: "850.0,-86.05238,2362.1572" } ] }
      ]

The call succeeds. Reported masters:

    masters: Skyrim.esm, Update.esm, Lux.esp

Reading the patch's own CELL record back (`housecarl_read_plugin_file`, formid
`0161F0:Skyrim.esm`) shows it is a full forward of the winner, not a lean host for the new
child group:

    ImageSpace      = 0924B3:Lux.esp
    LightingTemplate = 365B5C:Lux.esp
    Lighting.AmbientColor = 30,30,40,0
    Lighting.Inherits = AmbientColor, DirectionalColor, FogColor, FogNear, FogFar, ...
    Persistent[0] = [PlacedObject 000804:zzTest_DynamicBook.esp Tutorial_DynamicBookREF]
    Persistent[1] = [PlacedObject 000805:zzTest_DynamicBook.esp Tutorial_ContainerREF]

`housecarl_check_errors` on the patch is clean (0 dangling, 0 missing masters), so nothing
is malformed — the output is valid and wrong-shaped rather than broken.

### Expected

The patch should take only the cell's **defining** master (`Skyrim.esm`) and carry a lean CELL
override sufficient to host the new child group — not `Lux.esp`, and not a copy of Lux's
lighting fields.

Why: a placed reference is its own record with its own FormID, added to the cell's child group.
It is **not** removed when the CELL record it sits under loses to another plugin. So there is no
correctness reason to inherit the winner — adding refs to a cell is a load-order-solvable
situation, and the winner-forward buys nothing while costing a master.

Two concrete consequences of the current behaviour, both worse than the extra master:

1. **Stale field snapshot.** Sorted below the lighting mod, the patch wins the CELL record and
   re-asserts a frozen copy of that mod's lighting. If the lighting mod is later updated, the
   patch silently reverts the cell to the old values — a silent wrong result in the Q3 sense,
   from a patch whose author only asked to add two refs.
2. **Avoidable hard dependency.** Removing or replacing the lighting mod leaves the patch with a
   dangling master, for a reference that never needed it.

How I know the ref survives the CELL record losing: standard Bethesda-format behaviour for cell
child groups, and it is the reason mods can add clutter to vanilla interiors without depending
on every lighting overhaul that touches them. Confirmed by the maintainer during the session
this was found in.

Sorting the patch *above* the lighting mod is the correct user-side resolution today, and works
— but the author has to know to do it, and the forwarded lighting fields remain a latent trap if
they ever sort the other way.

### Environment (where relevant)

- Skyrim SE, MO2 instance, ~3,600 enabled mods / 3,799 plugins in the load order
- Found while authoring a quest-alias stored-text book (nested refs + a start-game-enabled quest)
- Aside, not part of this report: `Name.String` / `BookText.String` on a *newly created* record
  throws `CompositionRequiredException` (TranslatedString has no parameterless constructor).
  houseCARL labels that a known composition-wave deferral, and setting the whole field works, so
  it is noted here only as context for the session, not as part of this bug.

### LINKED ISSUE #308: [Bug] in-place verify claims "re-read off the written file" for a memory-derived line — an empty compose reports "+1" for a struct that serialized to nothing
> *Filed from an external live-session bug report; triaged, scrubbed, root-caused against source, and posted by AI.*

### Summary

The in-place write verify prints its per-op "what landed" clause under a banner that claims **"every edited record re-read off the written file"** — but that clause is derived from the **in-memory** record, before serialize. When a composed struct exists in memory yet serializes to **nothing**, the verify reports it as landed. The file is unchanged and an independent `read_record` shows the list still empty.

- severity: **wrong-answer** (Q3 — the one failure mode the write tools promise cannot happen)
- class: misleading output — "verified … re-read clean" for a subrecord that is not in the written file
- date reported: 2026-08-02 | houseCARL version: unknown (MCP build in use 2026-08-02)

Cost the reporter ~3 further debugging rounds on a live mod, looking elsewhere for why an in-game vendor menu was empty.

### Exact repro

Any `Faction` record with `Ranks = [list: 0 item(s)]`:

```
housecarl_bulk_apply target="<plugin>.esp" in_place=true acknowledge=true
operations=[
  {formid:"<Faction>", field_path:"Ranks", verb:"Add", compose:{type:"Rank"}}   <-- NO fields
]
```

Actual output:

```
1 edit:
  Faction <formid>  Add Ranks  -> [list: 1 item(s)]
verified — every edited record re-read off the written file (compact; ...):
  ✓ Faction <formid> — re-read clean (47 field(s)); ...; Add Ranks: now 1 (+1), new [0] = [Rank]
```

- Plugin size **unchanged** across the call.
- A later **independent** `housecarl_read_record` on the same record: `Ranks = [list: 0 item(s)]`.
- Re-running the identical call **with** `compose:{type:"Rank", fields:{Number:"0"}}` wrote correctly (+10 bytes) and an independent read confirmed `Ranks[0].Number = 0`.

Reporter's hunch (correct in effect): a struct with no serializable content — `Rank` needs its `RNAM` number — exists in memory but emits no bytes. Likely the same for any all-nullable modeled struct composed with no `fields`.

### Expected

Either the rank is written to disk, or the call is **refused** because a `Rank` with no fields has nothing to serialize — naming the required field(s), the same way the polymorphic-arm refusals already name legal values. Per Q3, a write that does not land must not report success.

### Root cause (verified against source)

**Two defects, one symptom.**

**1. The verify line fuses two different sources under one file-authoritative banner.**

The record-level `✓ … re-read clean (N field(s))` half *is* honest — `ReadBackInFull` walks the re-opened written file (`src/housecarl-core/WritePatchBuilder.cs:2359`, called at Phase 5 after serialize). But the per-op `Add Ranks: now 1 (+1), new [0] = [Rank]` clause appended to that same line comes from `OpResult.Landed`, produced by `DescribeApplied(IMajorRecord ov, WriteRequest req)` (`src/housecarl-core/WritePatchBuilder.cs:2512`) — `ov` is the **in-memory override**, read during apply, **before** the serialize. The renderer then concatenates both onto one line under `"verified — every edited record re-read off the written file"` (`src/housecarl-mcp/WriteTools.cs:660-680`).

So the banner's guarantee covers the record's existence and field count, not the "what landed" descriptor — which is the part a caller actually reads to confirm an edit. Compact mode (the default) never shows the list contents from disk, so nothing contradicts the memory-derived claim. `full_readback=true` *would* have shown `Ranks = 0 items`, since the full dump does come off `back`.

**2. Nothing refuses a compose with no serializable content.**

There is already a composed-null-arm guard (`src/housecarl-core/WriteEngine.cs:1485-1620, 3531`), but it only catches a required null arm that **throws** at serialize (the canonical `Condition` composed without its `Data` arm). An all-null `Rank` doesn't throw — it silently emits zero bytes, so it slips past both the guard and the verify.

### Fix direction

- **Honesty first:** either derive `Landed` from the post-serialize readback, or (cheaper) stop asserting file authority over a memory-derived clause — and have the compact readback compare the edited leaf's **on-disk** state against the op's claim, failing LOUD when they disagree. A `+1` that the file doesn't corroborate is exactly the inconsistency the forced in-place verify exists to catch.
- **Then the refusal:** pre-flight a `compose` whose built struct would serialize to nothing, naming the required field(s).

Same misleading-output class as **#259** and **#255**, but this one crosses from cosmetic into wrong-answer: #259's data was correct on disk and only the summary was wrong; here the summary is the *only* thing that exists.

*Provenance: external live-session bug report, scrubbed of the reporter's mod/load-order specifics (the repro is mod-independent), triaged and filed by AI.*

### LINKED ISSUE #321: [Bug] CopyFrom's from_source= PATH to an ACTIVE plugin routes off-order — forward's ActiveNameForPath rule was never mirrored
`housecarl_apply`'s `CopyFrom` lane decides "is this source off-order?" by asking whether the ACTIVE ORDER contains the name it was given. A **full path** never matches the name table, so a path pointing at a plugin that IS active takes the off-order arm: houseCARL opens the file directly and describes it as *not in the load order*.

PR #313 fixed exactly this on the `forward` lane and it was that PR's `[medium]` finding. The `CopyFrom` twin was never given the rule.

## The mechanism

Both ends of the CopyFrom lane now share one predicate (PR #318):

```csharp
public static bool IsOffOrderCopySource(PatchEdit e, LoadOrderResolver.IndexView view)
    => string.Equals(e.Verb, "CopyFrom", StringComparison.Ordinal)
       && !string.IsNullOrWhiteSpace(e.FromPlugin)
       && !view.ContainsPlugin(e.FromPlugin!);
```

`ContainsPlugin` is a lookup in the plugin-NAME table (`_nameToIdx`). `from_source=C:\MO2\mods\SomeMod\Bar.esp` is not a key in it, so the predicate answers "off-order" for a plugin the order is actively serving.

## What the user sees

`apply ops=[{…, "op":"CopyFrom", "from_source":"<full path to an ACTIVE plugin>"}]`

- the body is read off the file **directly**, bypassing the load-order build the rest of the call resolves against;
- an EXCLUDED-but-active plugin is read rather than refused, so the excluded-master refusal is skipped;
- the response describes the source as not in the load order, which is false.

Most of the time the file on disk and the active copy are the same bytes, so this is a wrong LABEL. Under a profile switch — where a filename is served by a different mod folder — it is a wrong BODY: the copy silently comes from a genuinely different file.

## The fix, already shaped

`forward` resolves it with `LoadOrderService.ActiveNameForPath` (a FULL-PATH identity compare, so a same-named backup correctly keeps the off-order lane, and a path to an excluded plugin does too — reading it directly is the escape hatch for exactly that), then re-spells the spec to the plugin NAME so the index lookup, the winner comparison and the report all speak the order's vocabulary.

Mirror that into the CopyFrom lane. Because PR #318 made the predicate **shared**, the clause is added in one place and reaches both the service's pre-locate and the engine's consume-side check — that sharing was done specifically so this fix could not land on one side only.

## Where it belongs

**W3 PR 3 (`copy`)** — that PR's subject is this lane, and it will already be in this file. Routed alongside #308 and #300.

Not a vocabulary change (no new parameter or verb), so it carries no W6 deadline — but it is a Q3 wrong-answer class, which is why it is filed rather than left as the code comment it currently is (`WritePatchBuilder.cs`, on `IsOffOrderSource`; delete that paragraph when this closes).

*Found during the 2.0 tidy-up sweep (PR #318), by a review round rather than by the work itself.*

Avick3110:
Three routed bug fixes from the W3 PR 3 charter, plus the generator housekeeping the 2.0 tidy-up left behind. `copy` itself is **not** in this PR — Aaron split it out so these could be reviewed on their own merits (PR 3b).

Fixes #321
Fixes #308
Fixes #300

---

## What landed

**#321 — a `from_source=` PATH to an ACTIVE plugin took the off-order arm.** Off-order-ness is a lookup in the plugin-NAME table and a full path is never a key there, so a path naming the very file the order serves read off disk directly and was described as not in the load order. Usually a wrong label; under a profile switch, a wrong body. `forward` has had `ActiveNameForPath` since PR #313 — the edit is now re-spelled to the plugin NAME before the shared predicate runs, so the arm decision, the winner comparison and every rendered sentence move together. A same-named backup and a path to an excluded plugin both keep the off-order lane, as on `forward`.

**#308 — the in-place verify's "what landed" clause is now read off the written file**, and a compose that cannot serialize is refused before anything is written. See "The #308 story" below: this shipped with a divergence *verdict* that was removed on Aaron's call after ten review rounds.

**#300 — a nested create hosts its child in the parent's DEFINING plugin's version**, not the load-order winner's. No unnecessary master, and no frozen copy of another mod's fields. Injected and excluded definers fall back to the winner and say so. Which plugin hosted the child is reported per record (`parent:`, `parent_host`).

**Generator** — an unknown first argument is refused (exit 2) instead of being read as an output directory. Two mistyped guard names wrote 13.5 MB into the working tree during the tidy-up; a bare name is now always a mode, even when a folder of that name exists (I reproduced the hole mid-review: `... src/housecarl-generator plugin` put 7.6 MB into `plugin/`).

## The #308 story, because the diff will not tell it

The fix shipped with a second half: a comparator that compared the in-memory reading against the file's and declared **"this op did not land"**. Across ten rounds that verdict produced a new false-accusation class each time — a byte-quantised `Percent` (`0.123` → `0.12`), an overlay's type name (`[WeaponBasicStatsBinaryOverlay]` vs `[WeaponBasicStats]`), `(absent)` vs `""`, a vanished substruct with no count to compare, two count-changing keyed ops in one call, and finally an unreadable read stamped `"verified"`. **Four of this branch's five high findings were that feature breaking under its own fixes.** Each version told a caller to re-issue a write that had landed; on a `Remove`, that deletes a further element.

Aaron's call (2026-08-11): **keep the honesty, drop the judgement.** Removed: the comparator, its presence plumbing, `OpResult.Divergence`, the `✗ … treat this op as NOT landed` lines, the JSON `divergences`/`diverged_ops` block, and the 12 checks that pinned them. Kept: the clause is re-derived from the re-opened written file; where the file could not answer (unreadable leaf, record not yielded, an op a later op superseded, a lane that runs no per-op check) the response says which reading it is showing — `landed`, `landed_on_disk`, `landed_source` (`written_file` | `superseded` | `no_answer` | `not_checked`), `verify_ran`. The pre-write refusal that stops the reported call is untouched.

The honesty half caused none of the regressions. The judgement caused all of them.

## On #300's shape — read this before judging the host choice

Four rounds flagged definer-hosting as a defect by comparing it to xEdit. xEdit copies the winner because a *person* is deciding, record by record. houseCARL is not making that decision: **the agent picks the shape from the request.** "Patch this mod for my list" wants a full resolution patch that wins and carries the resolved content. "Make this sit before the Reqtificator / Synthesis output" wants a feeder that masters neither, forwards neither, and does not bake in what those would regenerate. Same call, opposite right answers.

The residual is what you get when nothing is said, because it is the shape you can build up from and an inlined winner is the one you cannot cleanly subtract. Measured bound: the host carries **only** the new child — not the definer's existing children, not the winner's — so this touches the parent record's own fields, never other mods' child records.

**Open question for the next session** (Aaron: works when prompted, degrades with record count/complexity): is the inline path mechanically reachable — does forwarding the winner's version into a patch that already carries the host overwrite its fields while preserving the new child group, or does get-or-add semantics silently no-op it? If it no-ops, the composition argument does not hold and #300 is half-finished.

## Pre-PR review rounds (§5 #11)

**Ten rounds, twenty fresh unseeded agents.** Highs: round 1 (the #300 shape, escalated to Aaron rather than folded); rounds 2, 3, 4 and 8 — **all mine, every one introduced by a fold**, all in the #308 verdict now deleted. Rounds 5, 6, 7, 9 and 10 returned no highs.

Findings I refused, with reasons, are in the fold commits. Two rounds are worth calling out because they changed how the branch was guarded rather than what it does:

- **Round 9b** found two seams that could be **deleted with all 117 probes green** — the keyed-op exemption and the comparator's wire into the verify pass. Both got real arms, RED-verified.
- **Round 6** found my own guard doc arguing a case "could not be synthesised"; a reviewer synthesised it in minutes. That sentence is why a high shipped.

**Process failure worth recording:** §11 stops on severity, which cannot converge on a feature that generates a new true-but-benign finding per round. I should have invoked §4 at round three, when the same class recurred twice, instead of folding seven more times. Aaron and I have agreed §11 wants a convergence guard — if one subsystem draws findings in three consecutive rounds, stop and escalate the design. That edit is his to make.

I also repeated a scar from the previous session's handoff: RED-checking with the fix uncommitted, then `git checkout --` on the file, destroying it. Twice now.

## Known and NOT fixed (for the fresh session)

- #300's per-parent body fetch now scans the DEFINING master (usually `Skyrim.esm`) rather than the smaller winner; memoised per parent, so it bites on a bulk create fanning into many *distinct* cells. The honest fix is an index, not a memo.
- The empty-compose refusal's worked example can name a field that cannot be passed as a string (a reviewer enumerated 28 affected types, e.g. `LeveledItemEntry`).
- #300's contested line does not re-fire on a second `into=` call into the same cell.
- `parent_host` / `parent_contested` have no JSON-side probe; the text half is pinned.
- The empty-compose refusal also refuses a shape that used to work (compose empty, fill in a later op of the same call) — declared in the refusal text and the changelog.
- `ForwardTools.cs` still names `ResolveOffOrderCopySources` after the rename to `PrepareCopyFromSources`.
- The strings-aware re-open has no localized fixture; the CLI refusal has no probe.

## Verification

`ci-all` **117/117** · `apply-guard` **78/78** · `write-surface-guard` **129/129**, all in Release on the branch tip. Every new guard arm RED-verified against the pre-fix code; where a check passes in both directions the arm's own doc says so and calls it a bound.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -44,8 +44,8 @@ time rather than discovered in game.
 matches a filename — so `from_source=C:\…\SomeMod\Bar.esp` pointing at a plugin your order is actively serving read
 the file directly and described the source as not in your load order. Usually that was only a wrong label; under a
 profile switch, where the same filename is served by a different mod folder, it was a wrong *body*. `housecarl_forward` has had the rule since its own fix — a path that names the exact file your
-order loads is that plugin — and `CopyFrom` now shares it. A path to a same-named *backup* still reads off-order,
-as it should: the test is the file, not the name.
+order loads is that plugin — and `CopyFrom` now shares it. A path to a same-named *backup* still reads
+off-order, as it should: the test is the file, not the name.
 
 **`housecarl_write_seq` can finally put the `.seq` in the mod's own folder — `output_dir=` (#312).** Editing a
 plugin **in place** left the two halves in different mods: the `.esp` in the mod's own folder, the `.seq` in a
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -48,17 +48,20 @@ the cell record losing, so nothing about it needed the winner. Which plugin host
 created record (`parent: …`, `parent_host` in JSON) — including the two cases where the definer genuinely can't
 answer (an injected record, or a plugin this session had to exclude), where the winner is still used and says so.
 
-**You keep the choice — that is the point.** The host is a full record override and still conflicts at record level, so
-wherever your patch out-ranks the mod that currently wins that cell or topic, the parent resolves to the *defining*
-plugin's fields rather than that mod's. That is the **residual** shape: your patch carries your child and a host lean
-enough to lose harmlessly. It is the right default because the alternative — copying the winner — silently inlines
-another mod's content into your plugin and costs a master your reference never needed, and only you know whether this
-patch is meant to win or to feed one that does.
+**Which shape you get is a decision, and houseCARL no longer makes it for you.** The host is a full record override and
+still conflicts at record level, so wherever your patch out-ranks the mod that currently wins that cell or topic, the
+parent resolves to the *defining* plugin's fields rather than that mod's. That is the **residual** shape: your patch
+carries your child and a host lean enough to lose harmlessly.
 
-xEdit copies the winner because a person is deciding, record by record, in the moment. houseCARL doesn't decide for
-you: it writes the residual and **tells you** which mod currently wins that parent and which way the record will
-resolve. Sort below that mod to keep the residual shape, sort above it to assert your host, or forward that mod's
-version in deliberately when inlining is what you actually want. The new child is carried either way.
+Both shapes are legitimate and which one is right depends entirely on what the patch is for. A patch meant to *win* —
+"patch this mod for my load order" — wants the winner's content carried forward, resolved. A patch meant to **feed**
+one — "make this sit before the Reqtificator / Synthesis output" — must not master those, must not forward their
+records, and must not bake in what they would regenerate. Same call, opposite right answers.
+
+So this is now the base you build from rather than an answer imposed on you: the residual is what you get when nothing
+else is said, because you can add the winner's version deliberately on top, and you cannot cleanly subtract it once it
+is baked in. The response **tells you** which mod currently wins that parent and which way the record will resolve, so
+the choice is in front of you at write time. The new child is carried either way.
 
 **Fixed: a full path to an *active* plugin is no longer treated as an off-order file (#321).** `housecarl_apply`'s
 `CopyFrom` decided "is this source in the load order?" by looking up the name it was given, and a full path never
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -15,11 +15,16 @@ serializes to nothing — a Faction `Rank` composed with no fields is the report
 list stayed empty, and the call still reported the addition as landed; it cost the reporter three more debugging
 rounds on a live mod. That clause now comes off the re-opened file — for the op that last touched a given field in
 the call; an earlier op on the same field is marked as the applied edit's own reading, because the file holds the
-state after all of them and cannot answer for a middle one. If the file and the applied edit disagree, the
-response says so **loudly** on its own line and tells you to treat the op as not landed, instead of leaving you to
-notice a count. In JSON the two are separate facts: `landed` (what the edit did in memory), `landed_on_disk` (what
-the file carries), `divergence`, and `landed_verification` (`verified` / `diverged` / `superseded` / `no_answer` / `not_comparable` /
-`not_checked`). And the specific call that caused it is now
+state after all of them and cannot answer for a middle one.
+
+**houseCARL shows you both readings; it does not rule on them.** You get what the edit did in memory and what the file
+says, and where the file could not answer, the response tells you which reading you are looking at. It deliberately
+stops short of declaring "this op did not land" — deciding that means telling a real difference from a harmless one
+(a percentage that rounds when saved, an internal type name, an empty field versus a missing one), and every version
+of that judgement we built ended up accusing writes that had in fact landed. In JSON: `landed` (memory),
+`landed_on_disk` (the file), and `landed_source` — `written_file`, `superseded` (a later op in the same call wrote
+that field), `no_answer` (the file was re-read and didn't yield it), or `not_checked` (this lane runs no per-op file
+check), plus `verify_ran` on the response. And the specific call that caused the bug is now
 **refused before anything is written**, naming the fields that would give the structure content — a compose you gave
 nothing to, whose every settable field is empty, cannot land, so houseCARL no longer pretends it did. **One shape that
 used to work is refused with it:** composing an empty value and then filling its fields in a *later op of the same
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -30,12 +30,19 @@ the cell record losing, so nothing about it needed the winner. Which plugin host
 created record (`parent: …`, `parent_host` in JSON) — including the two cases where the definer genuinely can't
 answer (an injected record, or a plugin this session had to exclude), where the winner is still used and says so.
 
+**Know the trade this makes.** The host is a full record override, and it still conflicts at record level. So wherever
+your patch out-ranks the mod that currently wins that cell or topic, the parent record now resolves to the *defining*
+plugin's fields — usually vanilla — instead of that mod's. Before, it resolved to a frozen copy of that mod's fields,
+which was right until the mod updated and then silently wrong forever. Neither is what you want from a patch that only
+added a reference: **sort the patch before the mod that owns the parent**, and the new child is carried either way.
+The `parent:` line now says which mod that is and which way it will resolve, so the choice is in front of you at write
+time rather than discovered in game.
+
 **Fixed: a full path to an *active* plugin is no longer treated as an off-order file (#321).** `housecarl_apply`'s
 `CopyFrom` decided "is this source in the load order?" by looking up the name it was given, and a full path never
 matches a filename — so `from_source=C:\…\SomeMod\Bar.esp` pointing at a plugin your order is actively serving read
-the file directly, skipped the excluded-plugin refusal, and described the source as not in your load order. Usually
-that was only a wrong label; under a profile switch, where the same filename is served by a different mod folder, it
-was a wrong *body*. `housecarl_forward` has had the rule since its own fix — a path that names the exact file your
+the file directly and described the source as not in your load order. Usually that was only a wrong label; under a
+profile switch, where the same filename is served by a different mod folder, it was a wrong *body*. `housecarl_forward` has had the rule since its own fix — a path that names the exact file your
 order loads is that plugin — and `CopyFrom` now shares it. A path to a same-named *backup* still reads off-order, as
 it should: the test is the file, not the name.
 
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -18,7 +18,7 @@ the call; an earlier op on the same field is marked as the applied edit's own re
 state after all of them and cannot answer for a middle one. If the file and the applied edit disagree, the
 response says so **loudly** on its own line and tells you to treat the op as not landed, instead of leaving you to
 notice a count. In JSON the two are separate facts: `landed` (what the edit did in memory), `landed_on_disk` (what
-the file carries), `divergence`, and `landed_verification` (`verified` / `diverged` / `superseded` / `no_answer` /
+the file carries), `divergence`, and `landed_verification` (`verified` / `diverged` / `superseded` / `no_answer` / `not_comparable` /
 `not_checked`). And the specific call that caused it is now
 **refused before anything is written**, naming the fields that would give the structure content — a compose you gave
 nothing to, whose every settable field is empty, cannot land, so houseCARL no longer pretends it did. **One shape that
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -19,7 +19,11 @@ notice a count. In JSON the two are separate facts: `landed` (what the edit did 
 the file carries), `divergence`, and `landed_verification` (`verified` / `diverged` / `superseded` / `no_answer` /
 `not_checked`). And the specific call that caused it is now
 **refused before anything is written**, naming the fields that would give the structure content — a compose you gave
-nothing to, whose every settable field is empty, cannot land, so houseCARL no longer pretends it did.
+nothing to, whose every settable field is empty, cannot land, so houseCARL no longer pretends it did. **One shape that
+used to work is refused with it:** composing an empty value and then filling its fields in a *later op of the same
+call*. The check runs as the value is built and cannot see the ops after it, and the lane is all-or-nothing, so the
+whole call is refused — the message says so and points at the one-op form (compose it *with* its fields), which cannot
+half-land either way.
 
 **Fixed: adding a record inside an existing cell (or topic) no longer drags in the mod that wins it (#300).** Placing
 a new reference into a vanilla cell pulled the cell's *load-order winner* into your patch — so a lighting overhaul
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -48,13 +48,17 @@ the cell record losing, so nothing about it needed the winner. Which plugin host
 created record (`parent: …`, `parent_host` in JSON) — including the two cases where the definer genuinely can't
 answer (an injected record, or a plugin this session had to exclude), where the winner is still used and says so.
 
-**Know the trade this makes.** The host is a full record override, and it still conflicts at record level. So wherever
-your patch out-ranks the mod that currently wins that cell or topic, the parent record now resolves to the *defining*
-plugin's fields — usually vanilla — instead of that mod's. Before, it resolved to a frozen copy of that mod's fields,
-which was right until the mod updated and then silently wrong forever. Neither is what you want from a patch that only
-added a reference: **sort the patch before the mod that owns the parent**, and the new child is carried either way.
-The `parent:` line now says which mod that is and which way it will resolve, so the choice is in front of you at write
-time rather than discovered in game.
+**You keep the choice — that is the point.** The host is a full record override and still conflicts at record level, so
+wherever your patch out-ranks the mod that currently wins that cell or topic, the parent resolves to the *defining*
+plugin's fields rather than that mod's. That is the **residual** shape: your patch carries your child and a host lean
+enough to lose harmlessly. It is the right default because the alternative — copying the winner — silently inlines
+another mod's content into your plugin and costs a master your reference never needed, and only you know whether this
+patch is meant to win or to feed one that does.
+
+xEdit copies the winner because a person is deciding, record by record, in the moment. houseCARL doesn't decide for
+you: it writes the residual and **tells you** which mod currently wins that parent and which way the record will
+resolve. Sort below that mod to keep the residual shape, sort above it to assert your host, or forward that mod's
+version in deliberately when inlining is what you actually want. The new child is carried either way.
 
 **Fixed: a full path to an *active* plugin is no longer treated as an off-order file (#321).** `housecarl_apply`'s
 `CopyFrom` decided "is this source in the load order?" by looking up the name it was given, and a full path never
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -59,9 +59,18 @@ one — "make this sit before the Reqtificator / Synthesis output" — must not 
 records, and must not bake in what they would regenerate. Same call, opposite right answers.
 
 So this is now the base you build from rather than an answer imposed on you: the residual is what you get when nothing
-else is said, because you can add the winner's version deliberately on top, and you cannot cleanly subtract it once it
-is baked in. The response **tells you** which mod currently wins that parent and which way the record will resolve, so
-the choice is in front of you at write time. The new child is carried either way.
+else is said, because you can build the inlined shape deliberately, and you cannot cleanly subtract the winner's content
+once it is baked in. The response **tells you** which mod currently wins that parent and which way the record will
+resolve, so the choice is in front of you at write time. Whichever way you sort, the new child is carried.
+
+**Building the inlined shape: the order matters, and only one order is safe today.** Forward the winner's version into
+a patch **first**, then create into that patch — your new record is hosted in the forwarded body, and you get both.
+Doing it the other way round — creating first, then forwarding the winner into the same patch — **deletes the record you
+just created**, and reports success. That is a pre-existing `housecarl_forward` bug rather than anything specific to
+this feature: forwarding onto a record the destination already carries drops that whole record before re-copying it,
+and the child group — placed references under a cell, dialogue lines under a topic — goes with the drop. It affects
+`into=` and `in_place=` alike, so it can take content out of your own plugin on the in-place lane. It is tracked as
+**#324** and is not fixed in this release; until it is, use the safe order. The write response names the safe order too.
 
 **Fixed: a full path to an *active* plugin is no longer treated as an off-order file (#321).** `housecarl_apply`'s
 `CopyFrom` decided "is this source in the load order?" by looking up the name it was given, and a full path never
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,7 +13,9 @@ banner saying every edited record was re-read off the file on disk — but the p
 now 1 (+1)`") was read from memory *before* the file was written. When a composed structure exists in memory and
 serializes to nothing — a Faction `Rank` composed with no fields is the reported case — the plugin didn't grow, the
 list stayed empty, and the call still reported the addition as landed; it cost the reporter three more debugging
-rounds on a live mod. That clause now comes off the re-opened file. If the file and the applied edit disagree, the
+rounds on a live mod. That clause now comes off the re-opened file — for the op that last touched a given field in
+the call; an earlier op on the same field is marked as the applied edit's own reading, because the file holds the
+state after all of them and cannot answer for a middle one. If the file and the applied edit disagree, the
 response says so **loudly** on its own line and tells you to treat the op as not landed, instead of leaving you to
 notice a count. In JSON the two are separate facts: `landed` (what the edit did in memory), `landed_on_disk` (what
 the file carries), `divergence`, and `landed_verification` (`verified` / `diverged` / `superseded` / `no_answer` /
@@ -24,6 +26,12 @@ used to work is refused with it:** composing an empty value and then filling its
 call*. The check runs as the value is built and cannot see the ops after it, and the lane is all-or-nothing, so the
 whole call is refused — the message says so and points at the one-op form (compose it *with* its fields), which cannot
 half-land either way.
+
+**Reads show the modelled type name for a sub-structure, never Mutagen's internal one.** A field like a weapon's
+`BasicStats` read off a plugin on disk rendered as `[WeaponBasicStatsBinaryOverlay]` — the class houseCARL happens to
+load — while the same field read from a record being edited said `[WeaponBasicStats]`. Same field, two names, and one
+of them is an implementation detail. It now always reads `[WeaponBasicStats]`, in every read tool and in the in-place
+write verify (where the two used to appear in a single response, two lines apart, looking like a disagreement).
 
 **Fixed: adding a record inside an existing cell (or topic) no longer drags in the mod that wins it (#300).** Placing
 a new reference into a vanilla cell pulled the cell's *load-order winner* into your patch — so a lighting overhaul
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -8,6 +8,37 @@ when it changes.
 
 *Accumulating notes for the next cut — not yet released; `plugin.json` still reads the last shipped version.*
 
+**Fixed: an in-place edit's "what landed" line is now read off the written file (#308).** The verify prints under a
+banner saying every edited record was re-read off the file on disk — but the per-op half of each line ("`Add Ranks:
+now 1 (+1)`") was read from memory *before* the file was written. When a composed structure exists in memory and
+serializes to nothing — a Faction `Rank` composed with no fields is the reported case — the plugin didn't grow, the
+list stayed empty, and the call still reported the addition as landed; it cost the reporter three more debugging
+rounds on a live mod. That clause now comes off the re-opened file. If the file and the applied edit disagree, the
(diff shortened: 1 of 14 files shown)


Avick3110:
## Review — PR #323

Read the full diff plus surrounding source off the branch. I did **not** build or run the guards, so the 117/78/129 counts are taken on report.

> **Read this first — the merge gate.** Exactly **two** findings block merge, both mechanical. Everything else below is **filed, not folded.** This branch's history is the reason: four of its five highs came from folds, and a fresh session that reads this review alongside the PR's "Known and NOT fixed" list will otherwise start round 11. The known-not-fixed items are honest disclosure and belong in the issue lane — none of them are merge-blocking and none should be touched on this branch. On a branch with this history the default inverts: **a low gets an issue number unless folding it is provably mechanical.**

---

### What it does

Three routed fixes plus generator housekeeping; `copy` split out to PR 3b.

- **#321** — a `CopyFrom` `from_source=` PATH to an active plugin is re-spelled to the plugin NAME (`RespellActiveCopySourcePaths`) before the shared off-order predicate runs, so the arm decision, the winner comparison and the rendered text move together. Mirrors `forward`'s `ActiveNameForPath` (PR #313).
- **#308** — the in-place verify's per-op "what landed" clause is re-derived from the re-opened written file; an empty compose is refused pre-serialize. The divergence *verdict* is gone; what ships reports two readings and names the source.
- **#300** — a nested create hosts its child in the parent's **defining** plugin's version, not the winner's; the choice is reported per record.
- **Generator** — an unrecognised bare first argument exits 2 instead of being read as `[outputDir]`.

### Verified against source (all hold)

- `ops` ↔ `resolved` index alignment in `ApplyInPlaceCore` is 1:1 — each iteration adds exactly one `OpResult` or returns `Fail`. SNAM syncs append past it, which `i >= perOp.Count` handles.
- All three `ExpectedApplyRejectionException` catch sites (`WritePatchBuilder.cs:555`, `:962`, `:3032`) return a whole-call `Fail`. The `SchemaClassifier` doc's "all-or-nothing on every lane" claim is accurate.
- `edits` is a real `List<>` and the in-place branch returns before the patch-lane call, so the documented in-place mutation is safe; the respell runs before the `PatchEdit`-keyed source dictionary is built, so no stale keys. `ApplyEdits` is the only CopyFrom entry point.
- `ActiveNameForPath` normalises via `Path.GetFullPath`, so a relative `from_source` path works, and excluded plugins correctly keep the off-order lane.
- `ParentBodyFrom`'s memo caches nulls correctly (`TryGetValue` hits on a cached null); `AlreadyCarried`'s one-shot snapshot is safe because Phase 1 precedes all creation.
- `PathFamiliesOverlap` and the `SetAtIndex`-only keyed exemption are conservative in the right direction — a `Remove` after a `SetAtIndex` falls back to superseded.
- The generator guard is the **last** dispatch in `Program.cs`; `scripts/build-plugin.ps1:72` passes a rooted `$GeneratedDir`, so it is unaffected.

---

## Blocking (2)

### [medium] `contested_parent_hosts` is unbounded and sits outside the JSON budget
`src/housecarl-mcp/JsonWire.cs:1686`

The text twin caps at `contestedShown = 10` and says so, citing review [medium] for exactly this. The JSON side got no cap. Each host string is ~600–700 chars, so a `bulk_create` fanning children into N distinct contested cells writes ~700·N bytes before `created` is budgeted — the first `Over(w, ms, cap)` then trips and you get `rendered_created: 0, truncated: true`. That is the "0 of N" shape the text-side cap was added to prevent, and it is a D2 divergence in the direction that matters.

Mechanical: one bound, no judgement call in it.

### [medium] the printed remedy in the contested message is unmeasured
`src/housecarl-core/WritePatchBuilder.cs:2652`

The message tells the caller to *"forward '{winner}'s version in explicitly if you want its content inlined."* The PR body flags this as an open question and frames the risk as `GetOrAdd` silently no-opping. That framing is wrong in a way worth correcting before the next session inherits it: `ForwardRecords` on an `extend` does **drop-then-copy** — `((IMajorRecordEnumerable)patchMod).Remove(spec.Target, …)` then `GenericGetOrAddAsOverride` with the winner's body (`WritePatchBuilder.cs:2249–2261`) — precisely because the no-op class was already fixed as HCBR-2026-07-08-01 F1.

So the question is narrower and sharper than the PR body states: **does that whole-record `Remove` take the host's child group with it?** This branch's own `ChildLinesIn` probe establishes that `GenericGetOrAddAsOverride` does not carry children in, so the re-copied winner body will not restore them. If the drop removes the child, the printed advice destroys the record the create just placed.

**Resolve it with one probe, not by rewording.** Create a child into a contested parent, forward the winner into the same patch, read the child group back. Binary answer; then at most one sentence changes. Rewording the message without running it is folding an unverified finding — do not.

Worth being explicit, so nobody escalates this as a repeat of the #308 pattern: it is not one. The #308 verdict asked an **undecidable** question (is this difference real, or a representation the format is entitled to?). This is a single decidable question that simply has not been measured yet — ordinary §5 #1 empirical-first work.

---

## File, don't fold (3 + 1 nit)

These are real, and none of them should be fixed on this branch.

- **[low] `FieldValue.Readable` is not set on the depth ≥ 2 read path** — `ReadEngine.cs:394,397`. The new XML doc says "False when the read FAILED", but `EmitWithDepth`'s `!nav.ok` and its `catch` both set `Present: false` and leave `Readable: true`, as does `Expand`'s link branch. Unreachable from #308's verify (`DescribeApplied` reads at depth 1), but the flag exists so a consumer never parses `"(unreadable: …)"` out of prose — and a depth-2 consumer that trusts it gets exactly that.
- **[low] the cost statement names the wrong dimension** — `WritePatchBuilder.cs:715`. "Worth merging if a large in-place *target* ever measures badly" — but the unconditional per-op reflective `ReadFields` and the O(ops²) superseded scan scale with **op count**, not target size, and bulk in-place apply is a cornerstone use case. The n² is cheap in the common case (the FormKey compare short-circuits before `PathFamiliesOverlap`), so this is a doc fix, not a code one.
- **[low] the text-side contested cap is absolute, not relative to the budget** — `WriteTools.cs:1163`. 10 × ~700 ≈ 7 KB. Fine against `DefaultMaxChars = 80_000`; at a caller-supplied `max_chars=4000` the block alone exceeds the budget and the created list truncates at "0 of N" regardless. Derive the bound from `createCap`, or truncate the block by length.
- **[nit]** `ForwardTools.cs:21` still names `ResolveOffOrderCopySources` — already disclosed in the PR body; it is a plain `<c>` so the compiler does not catch it.

---

### Style / conventions

Consistent with the repo. Guard arms are RED-verified, and the two inert seams found in round 9b got real arms — the right response. `ChildLinesIn` is the strongest thing in the PR: it turns "how wide is the #300 trade" from an argument into a measurement, and bounds it to the parent's own fields. The generator's `IsDirectoryArgument` correctly refuses the "…or it already exists" clause and says why at the site.

Comment density is very high even by this repo's standards — several blocks are longer than the code they guard, and a few restate each other (`WritePatchBuilder.cs:767–769` duplicates the paragraph two lines above it verbatim). Not blocking; worth a pass sometime, not here.

### Process

The PR body's own account holds up: four of five highs were self-inflicted by folds, all in a feature that is now deleted, and the honesty half caused none of the regressions. The design fault generalises — the verdict asked whether a difference was *real or representational*, which requires knowing per field type how the format may legitimately re-represent a value on write. That is Mutagen's knowledge, not the write engine's, so each round patched one instance of an infinite class. "Keep the honesty, drop the judgement" is not a retreat: Q3 asks for no silent failure, not for adjudicating every write, and two readings plus a named source satisfies it completely.

**The §11 convergence-guard edit is out of scope for this PR.** It is a CLAUDE.md change carrying its own §5 #10 gate; bundling it into a code branch would need its own review lane and delay this merge. It is being tracked separately.

### Recommendation

Fix the two mediums — one bound, one probe. File the rest. Then merge.

Avick3110:
## Fix session — the two blocking mediums

Two commits. The first is the bound you asked for. The second ran the probe you asked for, **and the answer is the bad one** — so it is escalated, not fixed. Nothing else on this branch was touched: the three lows, the nit, the "Known and NOT fixed" list and the §11 convergence-guard edit are all untouched, and no new review rounds were run.

---

### 1. [medium] `contested_parent_hosts` bounded — fixed (`af2213c`)

The bound moved to `Wire.ContestedHostsShown = 10` and **both** lanes read it. A second literal is how they drifted in the first place, so `WriteTools.cs` now reads the shared constant too rather than its own `const int contestedShown`.

The json side also gained `total_contested_parent_hosts` beside the capped array — the same total/list pair `created` already uses two lines below — so a caller that is only shown 10 of N knows N. Each host past the cap is still named on its own record's `parent_host`.

No probe added: the json-side gap is already disclosed in the PR body's not-fixed list and belongs in that lane.

---

### 2. [medium] the printed remedy — measured. **It destroys the hosted child.** (`4bc941a`)

Ran it in `NestedParentHostArm`, on the fixture already there: create `W2HostedLine` into the contested topic, then `forward` the winner's version into the same patch with `into=`, then read the child group back.

```
[PASS] …and the remedy the message prints — forward the winner into the same patch — is accepted
[PASS] …and it does what it says: the winner's FIELDS are now inlined in the host
[FAIL] …and the new child SURVIVES that forward — the drop-then-copy does not take the child group
       got: child group after the forward: [] · whole file: [W2Topic]
```

The forward returned `extended …` — success. After it, the patch file holds **one record**, the topic. `W2HostedLine` is not orphaned and not re-parented; it is gone. Silent data loss, reported as a successful write.

So the message's *"forward '{winner}'s version in explicitly if you want its content inlined"* tells the caller to delete the record the create just placed.

**Your framing correction was right and the answer went the other way.** It is not a `GetOrAdd` no-op — the copy lands, the winner's fields are genuinely inlined (`Name` reads `Winner Topic` afterwards). It is the drop half: the whole-record `Remove` takes the child group with it, and `GenericGetOrAddAsOverride` — which this branch's own `ChildLinesIn` arm establishes carries no children in — cannot put it back.

#### This is not the branch's defect, and it is wider than the message

`WritePatchBuilder.cs:2249–2261` has done drop-then-copy since `e8363de` (2026-07-09, HCBR-2026-07-08-01 F1). This branch does not touch those lines — `git diff main...HEAD` on that hunk is empty. Which means the exposure is general, not #300-specific:

> **any `forward into=` (or `in_place=`) onto a record the destination already carries children under silently loses those children** — placed refs under a cell, INFOs under a topic.

#300 did not create that. It added a message pointing callers straight at it, on the one lane where the caller has *just* put a child there.

#### What I did not do

I did not reword the message, and I did not re-point the arm at the behaviour it found. Re-pointing would pin data loss as correct; rewording would fold an unverified answer into prose instead of fixing the engine. **The arm is committed RED**, so:

- `write-surface-guard` **131/132**
- `ci-all` **116/117** (the one failure is this arm)
- `apply-guard` **78/78**

all Release, on `4bc941a`. The red is the finding, and CI is red on purpose — leaving it green would have needed one of the two moves above.

#### The call is yours (§5 #5)

Three shapes I can see, not ranked because the choice is an architecture one:

1. **Preserve the child group across the drop** in `ForwardRecords`' extend path — snapshot children before `Remove`, re-attach after the copy. Fixes the general defect; needs a rule for what happens when the incoming body has children of its own.
2. **Refuse a forward whose dropped record carries children** in the destination, naming them. Cheapest, honest, and closes the silent half; leaves inlining genuinely unreachable on this lane, which weakens #300's composability argument.
3. **Change the remedy the message prints** — e.g. forward the winner *before* creating the child (which works: the create then hosts into the already-carried parent), or into a patch that sorts above rather than the same one. Leaves the engine defect standing for every other caller.

Worth noting for whichever way it goes: (3)'s "forward first, then create" ordering is safe today, and a caller who never reads the message never hits this.

**Open question I could not close in scope:** whether this also needs a `#N` of its own against `forward` independent of #300. I did not file one — that is a routing call and I would rather you make it than have two issues chasing one defect.

Not merging. Handing back.

Avick3110:
## Re: the fix session's report

Verified rather than taken on report. Both commits do what they say:

- `af2213c` — `Wire.ContestedHostsShown = 10` (`ReadTools.cs:642`) is genuinely read by both lanes (`JsonWire.cs:1695`, `WriteTools.cs:1164/1166/1167`), plus `total_contested_parent_hosts`. That is more than the review asked for — I asked for a bound, you removed the second literal that let them drift. Right call.
- `4bc941a` — the arm really drives `ForwardTools.Forward(…, into: patchFile)` and is committed failing.
- The "not this branch's defect" claim holds: `git diff main...claude/w3pr3-copy` on that hunk is empty; the lines were last modified by `64173c6` (W3 PR 2b), not here.

Leaving the arm RED rather than re-pointing it was the right refusal.

---

### One escalation: `in_place` is the worse half, not a parenthetical

`ForwardRecordsInPlaceCore` (`WritePatchBuilder.cs:1776–1786`) carries the identical drop-then-copy. There the target is the user's **own plugin**, on the lane whose own render banner reads *"your ORIGINAL file; no houseCARL backup or undo."* So a `forward … in_place=` onto a cell or topic in a plugin the user owns, carrying their own placed refs or INFOs, silently deletes them from their file, with no patch-lane cushion.

### One correction: option 1's open question dissolves

The escalation lists "needs a rule for what happens when the incoming body has children of its own" as the cost of preserving the child group. There is no incoming child set — `GenericGetOrAddAsOverride` carries none in, which this branch's own `ChildLinesIn` arm establishes. Preservation is *re-attach exactly what the destination had*, with nothing to reconcile. That makes option 1 materially cheaper than the escalation implies, and it is the only one of the three that fixes the general defect.

---

## The call (Aaron, 2026-08-11): split it

**Filed as #324**, titled against `forward` rather than #300 so the fix PR is not scoped to nested creates — the `in_place` half has nothing to do with #300. Option 1 is the shape; option 2's refusal is the fallback for any parent type where preservation is not reachable. Option 3 standalone is rejected: it leaves silent data loss for every caller who never reads the #300 message.

**#323 does not fix the engine.** The branch's own three fixes are done and verified, and the defect lives in a lane it does not touch. Widening a PR that has already absorbed ten rounds is the bloat that caused this branch's trouble in the first place. What #323 must do is stop shipping a sentence that instructs data loss, and go green honestly.

Three things on this branch, then hand back:

**1. Measure the safe ordering.** The escalation says forward-the-winner-BEFORE-the-create is safe "today" — but only the destructive ordering was probed. That one is reasoned, not measured, and it is about to go into a user-facing message; it gets the same treatment the first one got. Same arm, same fixture: forward the winner into the patch first, then create the child with `parent=`, then assert **both** the winner's fields and the child are present. If it fails, stop and escalate again — do not go looking for a third ordering.

**2. Fix the message, given a green (1).** Replace the false remedy at `WritePatchBuilder.cs:2652` with the measured-safe ordering. This is **not** the reword that was correctly refused: that one would have asserted the destructive operation is fine. This substitutes a measured-true statement for a measured-false one. `plugin/CHANGELOG.md`'s `## Unreleased` entry needs the same correction — it claims *"you can add the winner's version deliberately on top, and you cannot cleanly subtract it once it is baked in,"* and the probe falsified the first half. Same PR (§5 #10).

**3. Remove the destructive assertions from `NestedParentHostArm`**, keeping the safe-ordering ones from (1). The probe body is preserved verbatim in `4bc941a` and quoted in #324, so nothing is lost — the fix PR pastes it back as its RED. #323 must not carry another lane's red. Removing it is not re-pointing it: re-pointing would assert child loss is correct, removing asserts nothing.

Out of scope, unchanged: the three lows, the nit, the "Known and NOT fixed" list, and the §11 CLAUDE.md edit. No new pre-PR review rounds.

Avick3110:
## Re: the split — all three done, branch green

One commit, `f40526b`. `4bc941a` is untouched above it, so #324's quoted probe body stays reachable in this branch's history exactly as you scoped it.

Taking the correction: option 1's "reconcile incoming children" cost was mine and it was wrong — `GenericGetOrAddAsOverride` carries none in, so there is nothing to reconcile. And you are right that `in_place` is the worse half, not a parenthetical; I had it in a subordinate clause when it is the lane with no backup and the user's own file.

---

### 1. Safe ordering — measured, green

Ran it rather than reasoning it, same arm and fixture:

```
[PASS] the SAFE ordering, step 1: the winner's version of the parent forwards into a patch of its own
[PASS] …step 2: the child creates into THAT patch, hosting in the parent it already carries
[PASS] …and BOTH land: the winner's FIELDS are inlined AND the new child is in the group
```

Topic reads `Winner Topic`, child group holds `W2InlinedLine`. The create resolves its parent from the record the patch already carries, so there is no drop for the child to survive.

One thing the arm now says out loud because it surprised me while writing it: the forwarded parent does **not** bring the winner's own child lines with it — `forward` copies the record, not its group. That is correct and not a second bug (a child group survives its parent record losing, which is the premise of the lean host), but "inlined" means *fields*, and a caller could reasonably have read it as *everything*. The arm's comment states it so nobody re-measures it as a defect later.

### 2. Message + changelog

`WritePatchBuilder.cs` now prints the safe order and marks the other one destructive, pointing at #324. The changelog's `## Unreleased` claim — *"you can add the winner's version deliberately on top"* — is replaced with the order, the destructive ordering named as destructive, and the fact that #324 reaches `in_place=` and so can take content out of the caller's own plugin.

**The sentence got its own arm, and this is the part worth flagging:**

```
[PASS] …and the inlining lever names the SAFE order and marks the other one destructive
```

RED-verified by restoring the old wording — that check fails and **nothing else does**. The three behavioural arms stay green under the false sentence, because they measure the engine and the sentence is what a caller acts on. Without the text pin, a future trim back to "forward the winner's version in explicitly" ships data-loss advice through a fully green guard. That is the same gap that let this wording ship in the first place.

### 3. Destructive assertions removed

Gone, not re-pointed — re-pointing asserts child loss is correct, removing asserts nothing.

---

### Verification — `f40526b`, Release

- `ci-all` **ALL PASS** (117/117)
- `apply-guard` **78/78**
- `write-surface-guard` **133/133**

Net +4 arms on write-surface-guard vs. the PR body's 129: three safe-ordering, one message pin.

### Two things I did not do

- **No new review rounds**, per your instruction — and this branch is now three folds past its last round, which is worth you knowing when you re-review rather than discovering.
- **Did not touch** the three lows, the nit, the "Known and NOT fixed" list, or the §11 CLAUDE.md edit.

### One disclosure against a filed low

The contested message is now ~180 chars longer per host, which makes the already-filed *"text-side contested cap is absolute, not relative to the budget"* low (`WriteTools.cs:1163`) bite slightly sooner — 10 × ~880 ≈ 8.8 KB rather than ~7 KB. Still nothing against the 80,000 default; it moves the `max_chars=4000` case further into the red. Flagging rather than folding, since that low is yours to route.

Not merging. Yours.

Issue #333, opened 2026-08-12
[Bug] forward: a child and its parent in ONE call silently skips re-asserting the parent's fields (fresh-patch lane)


Avick3110:
## What happens

`housecarl_forward` with a child and its parent in the same call, on the **fresh-patch lane** (no `into=`, no `in_place=`), lands the child but silently does **not** re-assert the parent's fields. The call reports both records forwarded from the source.

Spec 1 (the child) materializes the parent in the patch to host itself — hosted in the parent's *defining* plugin's version, per #300. Spec 2 (the parent) then hits `GenericGetOrAddAsOverride` on a FormKey the patch now carries, and get-semantics return the record already there. The copy is skipped.

`alreadyCarried` is only populated when `extend` is true (`WritePatchBuilder.cs:2252-2255`), so the replace path — and its drop-then-copy — never engages on this lane. The record materialized mid-call is invisible to it.

## Measured

On `claude/fix-324-child-group` (worktree fixture `write-surface-guard`), forwarding `[<INFO>, <DIAL topic>]` from the replacer into a fresh patch:

```
[PASS] a child and its parent forward in ONE call to a fresh patch - the call is not refused
[PASS] .and the child lands under its parent in the patch
[FAIL] .and the parent's own fields land too
       got: name=Master Topic
```

The source (`HcW2Repl.esp`) carries the topic as `Winner Topic`. The patch holds `Master Topic` — the definer's version — while the response lists the topic as `copied from HcW2Repl.esp`.

## Why it matters

Silent wrong result reported as success — Q3. It is the same family as HCBR-2026-07-08-01 F1 (get-semantics silently skipping a copy on a collision), surviving on the one lane that never builds the `alreadyCarried` map. The caller gets a patch whose parent record is the wrong version, with nothing in the response to indicate it.

Listing the parent **before** the child in the same call avoids it, as does forwarding the parent in a separate call.

## Not a regression from #324

Found while reviewing the #324 branch; those lines are untouched by it. The #324 fix deliberately does not engage here — nothing is dropped on this lane, so there is nothing to preserve. The branch's guard arm asserts what is true (the call is accepted, the child lands) and documents this gap rather than asserting it, so as not to widen one lane's fix into another's.

## Shape for the fix

Populate `alreadyCarried` from records materialized during the call, or re-check the destination per spec rather than from a pre-loop snapshot, so the replace path engages on a FormKey the call itself put there. Both forward lanes should agree on what "the destination already carries this" means.

[...]

PR #334, opened 2026-08-12
fix(#324): forward preserves the child group across its drop-then-copy

### LINKED ISSUE #324: [Bug] forward into=/in_place= silently DELETES child records under the forwarded record
## What happens

`housecarl_forward` onto a record the destination **already carries** drops the whole record and re-copies the source body:

- `WritePatchBuilder.cs:2249–2261` — `ForwardRecordsCore` (patch / `into=` lane)
- `WritePatchBuilder.cs:1776–1786` — `ForwardRecordsInPlaceCore` (`in_place=` lane)

Both do `((IMajorRecordEnumerable)…).Remove(spec.Target, RemovalTypeFor(existing), throwIfUnknown: true)` and then `GenericGetOrAddAsOverride(…, body, …)`.

The `Remove` takes the record's **child group** with it, and `GenericGetOrAddAsOverride` carries no children in — established independently by `NestedParentHostArm`'s `ChildLinesIn` assertion (PR #323). So the children are dropped and never restored, and the call reports **success**.

Any record with a child group is exposed: placed references under a CELL, INFOs under a DIAL.

## Measured

PR #323, commit `4bc941a`, `write-surface-guard` / `NestedParentHostArm`. Create `W2HostedLine` into a contested topic, forward the winner's version into the same patch with `into=`, read the child group back:

```
[PASS] the remedy — forward the winner into the same patch — is accepted
[PASS] it does what it says: the winner's FIELDS are now inlined in the host
[FAIL] the new child SURVIVES that forward
       got: child group after the forward: [] · whole file: [W2Topic]
```

`W2HostedLine` is not orphaned and not re-parented. It is gone. The forward returned `extended …`.

## Why it matters

Silent data loss reported as a successful write — Q3. The `in_place=` lane is the worse half: the target is the user's **own plugin**, on the lane whose own render banner reads *"your ORIGINAL file; no houseCARL backup or undo."*

## Not a regression from #300 / PR #323

The drop-then-copy shape dates to the HCBR-2026-07-08-01 F1 fix (2026-07-09), where `GetOrAdd`'s get-semantics were silently *skipping* the copy; the lines were last modified by `64173c6` (W3 PR 2b). `git diff main...claude/w3pr3-copy` on that hunk is empty. #300 only added a message pointing callers at it.

## Shapes for the fix

1. **Preserve the child group across the drop** — snapshot children before `Remove`, re-attach after the copy. Fixes the general defect on both lanes. Note the obvious objection dissolves: there is no incoming child set to reconcile, because `GenericGetOrAddAsOverride` carries none in, so this is re-attach-what-was-there.
2. **Refuse a drop whose record carries children**, naming them. Cheapest, closes the silent half, but makes the operation unreachable rather than correct.
3. Message-only changes — do not fix this; they leave every other caller exposed.

The probe body is preserved in `4bc941a` and can be pasted back as this fix's RED arm. Its assertions are being removed from PR #323 (a different lane's PR should not carry another lane's red); the fix PR is where they belong.

Avick3110:
Fixes #324.

`forward` onto a record the destination **already carries** replaced it by dropping the whole record and copying the source body in. The drop took the record's child group with it and the copy carried none back, so every record nested under the replaced one was destroyed while the call reported success — a topic's INFOs, a cell's placed references. The `in_place=` half did that to the caller's own plugin, on the lane whose banner says there is no backup.

Both lanes now lift the children off before the `Remove` and re-attach them after the copy. Nothing about the forward semantic changes: a forwarded parent asserts the source's *fields*, and the source's own children stay in the source's plugin.

## What is refused rather than guessed

Two states fail closed with nothing serialized: a copy that arrives carrying children of its own (Mutagen's semantics changing under us — re-attaching would silently discard one of the two sets, or import the source's into a destination that deliberately holds none), and a child count that does not survive the replace. The second check runs off Mutagen's own containment walk rather than the reflected property set that did the re-attach, so a containment path the walk cannot see fails loud instead of losing records.

The reflected walk recurses, and that is load-bearing: `Worldspace.SubCells` reaches its cells through two non-record container types, so a one-level test would silently skip every exterior cell in the game. `IFormLink`s are cut — a link references a record, it does not own one.

## The gate

`RestoreChildGroup` acts if and only if a capture happened — the fact that a drop-then-copy is underway, written by `CaptureChildGroup` and nowhere else. That is worth stating because the branch inferred it twice and was wrong twice (see the rounds below). Inferring it from the carry being empty made the arrival tripwire unreachable in the case where an arriving set is an *injection*; inferring it from the record's type made it fire on records the call itself had just created, refusing whole writes and blaming Mutagen. The lane knows the answer; it is now passed rather than re-derived.

## Proof

`write-surface-guard` 133 → **163**. A new `#324` arm covering both lanes on both copy paths — `DialogTopic` (flat group) and `Cell` (nested; resolves through a `ModContext` that rebuilds the block/subblock chain) — asserting the children survive **and** that the source's fields actually landed, since a forward that no-op'd would preserve children perfectly. Plus the child-bearing property set pinned across every concrete record type Mutagen models, the three refusal sentences rendered, and the response wording.

RED-checked three ways rather than assumed:
- neutering the capture fails the child-survival arms on **both** paths — so the nested cell path was destroying placed references too, which the changelog had asserted and nothing had measured;
- restoring the round-1 gate fails exactly the three arms that pin the same-call regression;
- restoring the old response sentence fails the sentence arm.

`ci-all` 117/117 · `apply-guard` 78/78.

## Pre-PR review rounds

Two rounds, three fresh independent reviewers each, prompts seeded with the branch's settled-decisions list and nothing else.

**Round 1 — 1 high, 2 medium, 6 low, 2 nits.** All three reviewers independently led with the same high: the `create` render still told callers that the order this branch had just made safe destroys their data, citing #324 as an open bug — and a guard arm pinned that wording, issue number included, so correcting the message failed CI. The CHANGELOG said both that #324 was fixed and, four paragraphs later, that it was not. Also folded: cell coverage (the arm only exercised the flat-group path), the unreachable arrival tripwire, the three refusal messages no arm had ever rendered (which surfaced a doubled "nothing was serialized"), a property pin covering 6 types of 133, a capture throw reported as "the override-copy threw", and a missing not-the-source's-children clause on the in-place half.

*Refused:* raising or restructuring the recursion depth bound — two reviewers independently ran a more generous walk (depth 12, per-branch visited set) over all 133 types and got the same answer set, and deeper nesting fails closed with the all-types pin going red first. Also refused as a live defect: the count-mismatch message reading oddly when more children survive than were held — that branch is unreachable, since an arriving set is refused earlier and the re-attach cannot add.

**Round 2 — 2 highs, both regressions round 1's own fold introduced, both on the gate.** Forwarding a child and its parent in one call now refused the whole write and told the caller to report a Mutagen bug; and the type gate short-circuited the count check for any type whose containers the walk cannot see — the silent-loss shape the function exists to prevent. Two reviewers also flagged, independently, that the forward response still said "the old body is gone" on exactly the operation this branch changed: revert a cell in place, keep your forty placed refs, and be told the old body went.

**The rounds stopped at two.** Per CLAUDE.md §11 the same failure class recurring in consecutive rounds is a design signal, not a fix queue — so the seam went to Aaron rather than into a third fold. His call was to thread the capture fact, with conditions: single writer, the response sentence measured before rewording and given its own arm, the two self-introduced cleanups riding along, and no new rounds after the directed fold.

## Review fold

Aaron's review is folded (commit `9ba7b3a`): the dry-run halves of the new render sentence now have arms of their own — they are the sentences a caller reads while deciding whether to proceed, and the branch had pinned only their live twins; the tool description names `preserved_children` instead of pointing a count at a bool; `ReachesOwnedRecord`'s visited set is a path set, so a type first reached at the depth bound is no longer memoized as unreachable for shallower branches; the new field is documented with `<param>`; and `Worldspace` — the reason the walk recurses — is exercised directly across both container levels rather than pinned by reflection alone, with the remaining gap (no end-to-end WRLD forward) stated in the arm so it is not mistaken for coverage.

His second finding was worse than filed, and retracts one of mine. The pair block's comment described a fixture state the arm no longer produced: the in-place half runs earlier in the same arm and rewrites the replacer's topic to `Master Topic`, so both sides of the comparison read `Master Topic`. **That is where #333 came from** — I read a failing assertion as a silent skip when what had actually changed was the source. Re-measured with the block moved above the in-place half, the parent's fields do land; the second spec is not skipped. #333 is closed as not-a-defect, and the ordering is now asserted with a comment recording why the block sits where it does.

`write-surface-guard` 163 → **169**. The two dry-run arms are RED-checked: restoring the old preview sentence fails both.

## Not fixed here

- A forwarded parent's child-group **header stamps** are not carried across from the destination — the `Timestamp` / `Unknown` class of field on a DIAL, and the `Timestamp` / `PersistentTimestamp` / `TemporaryTimestamp` / `UnknownGroupData` equivalents on a CELL. Dropped per §2 fix-or-drop: CK bookkeeping, invisible to the engine, visible only in xEdit, and the fix would be type-specific special-casing against the by-construction posture. Scope stated as the class of field rather than a mechanism, deliberately: round 2 reported the DIAL half as carrying the source's values and the CELL half as zeroed, and this branch's fixture cannot discriminate — its stamps are zero on both sides, which is exactly the kind of reading that produced #333.
- **#335** — filed from Aaron's aside: `mutagen-reference` classifies `Cell.Landscape` as a formlink, while this branch's all-types pin has it as an owned child record. A generator-classifier question, not this PR's.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -8,6 +8,16 @@ when it changes.
 
 *Accumulating notes for the next cut — not yet released; `plugin.json` still reads the last shipped version.*
 
+**Fixed: forwarding a record no longer deletes the records nested under it (#324).** When `forward` targeted a record
+the destination *already carried*, it replaced it by dropping the whole record and copying the source body in — and
+the drop took the record's child group with it. Any dialogue line under a forwarded topic, or placed reference under
+a forwarded cell, was destroyed, and the call reported success. The `in_place=` lane was the worse half: it deleted
+those records out of your own plugin, on the lane that keeps no backup. The children are now lifted off before the
+replace and re-attached after it, so a forward changes the record's *fields* and leaves what is nested under it
+alone. Nothing about the forward semantic changes: the source's own children still stay in the source's plugin,
+which is what lets a patched topic not fight another mod's added lines. If houseCARL ever cannot account for every
+child across the replace, it refuses and writes nothing rather than emit a plugin it knows is short a record.
+
 **Fixed: an in-place edit's "what landed" line is now read off the written file (#308).** The verify prints under a
 banner saying every edited record was re-read off the file on disk — but the per-op half of each line ("`Add Ranks:
 now 1 (+1)`") was read from memory *before* the file was written. When a composed structure exists in memory and
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -73,14 +73,10 @@ else is said, because you can build the inlined shape deliberately, and you cann
 once it is baked in. The response **tells you** which mod currently wins that parent and which way the record will
 resolve, so the choice is in front of you at write time. Whichever way you sort, the new child is carried.
 
-**Building the inlined shape: the order matters, and only one order is safe today.** Forward the winner's version into
-a patch **first**, then create into that patch — your new record is hosted in the forwarded body, and you get both.
-Doing it the other way round — creating first, then forwarding the winner into the same patch — **deletes the record you
-just created**, and reports success. That is a pre-existing `housecarl_forward` bug rather than anything specific to
-this feature: forwarding onto a record the destination already carries drops that whole record before re-copying it,
-and the child group — placed references under a cell, dialogue lines under a topic — goes with the drop. It affects
-`into=` and `in_place=` alike, so it can take content out of your own plugin on the in-place lane. It is tracked as
-**#324** and is not fixed in this release; until it is, use the safe order. The write response names the safe order too.
+**Building the inlined shape.** Forward the winner's version of the parent into a patch, then create into that patch —
+your new record is hosted in the forwarded body, and you get both. Either order works: forwarding onto a parent a
+patch already holds keeps the records nested under it (that order used to destroy them, which is the `forward` fix
+above in this same release). The write response names the remedy too.
 
 **Fixed: a full path to an *active* plugin is no longer treated as an off-order file (#321).** `housecarl_apply`'s
 `CopyFrom` decided "is this source in the load order?" by looking up the name it was given, and a full path never
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -661,8 +661,9 @@ public static class WriteEngine
     public readonly record struct ChildGroupCarry(
         IReadOnlyList<(PropertyInfo Prop, object? Value)> Held, int Count, IReadOnlyList<string> Names)
     {
-        /// <summary>Nothing was held — the record owns no children (the overwhelmingly common case: every flat record
-        /// type but Cell / DialogTopic / Worldspace), so the restore is a no-op and the lanes skip it.</summary>
+        /// <summary>Nothing was held — the record owns no children (the overwhelmingly common case: every record type
+        /// but Cell / DialogTopic / Worldspace), so there is nothing to re-attach. NOT a licence to skip the arrival
+        /// tripwire in <see cref="RestoreChildGroup"/>: see the comment there.</summary>
         public bool IsEmpty => Count == 0;
     }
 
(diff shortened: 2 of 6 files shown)


Avick3110:
## Review

The fix is right and the seam is in the right place. One thing I'd want folded before merge — the **dry-run halves of the new render sentence have no probe** — and the rest is take-or-leave.

### What I checked and am *not* flagging

- **Capture-before-`Remove`** ordering, and `carriedChildren` scoped per-spec inside the loop in both lanes — no leakage across specs.
- **Refusals really do leave the file alone.** Both lanes bail before Phase 4, so the substituted `untouchedClause` is accurate on both, and the ordering arm (`Nothing was serialized` before `Please report`) pins it.
- **No new masters.** The re-attached children were already in the destination file, so the fix restores prior state rather than pulling anything new into the header.
- **Parent + child in one `into=` call, both orders.** `alreadyCarried` entries for children stay valid across a parent's replace because the re-attach is by reference, so the later `Remove` still reaches them.
- `default(ChildGroupCarry).Count == 0` ⇒ `preserved_children` is 0 on the no-replace path; the JSON emits the field unconditionally, so consumers don't get a shape change.
- The CHANGELOG rewrite of the #300 entry sits inside `## Unreleased` — not an edit to shipped history.
- The recursion-depth bound. Two reviewers already measured a depth-12 / per-branch walk to the same answer set, and it degrades fail-closed. Not re-litigating.

---

### 1. The dry-run halves of the new render sentence have no probe — [medium]

`RenderForward` (`src/housecarl-mcp/WriteTools.cs:867`) now branches four ways. The two `o.DryRun` branches — "the old FIELDS would be gone … would be KEPT" and "it carries no nested records" — are rendered by **no arm anywhere**. Both existing forward dry-run arms (`WriteSurfaceGuardProbe` lines 677 and 823) forward into a *fresh* patch, so `ReplacedExisting` is false and the whole block is skipped.

This is the branch's own standard applied unevenly: the three live sentences each got an arm (rightly — that's the #323 → #324 lesson), and their dry-run twins, which are the sentences a caller reads *while deciding whether to proceed*, got none. Cheap fold: run the `into=` topic case once with `dry_run: true` before the real call and pin both dry branches.

Pre-existing but widened — the old single dry-run sentence was unpinned too; this PR splits one unpinned sentence into two.

### 2. The #333 comment in `ForwardChildGroupArm` describes a fixture state the arm no longer produces — [low]

The pair block's closing comment says the patch holds `"Master Topic"` *"while the call reports the topic forwarded from the replacer, whose version reads 'Winner Topic'."*

By the time the pair block runs, the **in-place half earlier in the same arm** has already overwritten the replacer's topic to `"Master Topic"` — the arm asserts exactly that (`TopicNameIn(replPath, fx.TopicKey) == "Master Topic"`). So both sides read `"Master Topic"` and the discriminator the comment names doesn't exist in this fixture; the arm couldn't measure #333 here either way.

Nothing is asserted, so no false green — but #333's in-repo evidence reads stronger than it is, and this file's whole posture is measured-not-reasoned. Either move the pair block above the in-place half, or restate the comment as *not distinguishable in this fixture state* and point at where it actually was measured.

### 3. `ReachesOwnedRecord`'s visited set is shared across sibling branches — [low]

```csharp
if (depth > 6 || !seen.Add(t)) return false;
```

`seen` is fresh per top-level property but **not per branch**, and it is never removed on backtrack. A type first reached at depth 6 is added to `seen` and evaluated with its children at depth 7 (all `false` on the bound); if the same type is later reachable at depth 2 down another branch, `seen.Add` fails and it returns `false` without ever being re-examined. That is a depth-truncated answer memoized as a depth-independent one.

Not live — measured equivalent, and it fails closed (count-check refusal, all-types pin goes red first). But the doc comment's "the constant is a tripwire, not a correctness assumption" is true of the depth alone and slightly less true of depth-plus-shared-`seen`. One-line fix if you want it: make it a path set (`try { … } finally { seen.Remove(t); }`).

### 4. Tool description: "counted in that flag" points at the wrong field — [low]

`src/housecarl-mcp/WriteTools.cs:375`:

> …flagged per record in the response); records NESTED under it (a topic's lines, a cell's placed refs) are KEPT **and counted in that flag**.

`replaced_existing` is a bool; the count lives in the separate `preserved_children` number this PR adds. A model reading the description will go looking for a count on the flag. Name the field.

### 5. Duplicate `<summary>` on `ForwardedRecord` — [nit]

`src/housecarl-core/WritePatchBuilder.cs:2012` opens a second `<summary>` on a member that already has one (2001–2007), after a stray top-level `<para>` (that one's pre-existing). `ForwardedRecord` is a positional record, so `<param name="PreservedChildren">…</param>` is the idiomatic slot and won't get dropped by doc tooling picking one `<summary>`.

### 6. "Not fixed here" understates the header-stamp carry-over — [nit]

The bullet names a forwarded **DIAL** taking the source's child-group `Timestamp` / `Unknown`. `Cell` carries the same class of field — `Timestamp`, `UnknownGroupData`, `PersistentTimestamp`, `PersistentUnknownGroupData`, `TemporaryTimestamp`, `TemporaryUnknownGroupData` — and the changelog leads with cells. Same drop rationale (CK bookkeeping, invisible to the engine); just widen the sentence so the dropped scope matches what actually happens.

### 7. No end-to-end `Worldspace` arm — [coverage, low]

`Worldspace` is the *reason* the walk recurses (`SubCells` reaches its cells through two non-record container types) and the only pinned type whose re-attach crosses two container levels — but it is covered by the property-set pin only, never by an actual forward. Failure mode is bounded (a `SetValue` throw or a count mismatch both refuse, neither loses data), so this cannot ship data loss; what is untested is *forward a worldspace and eat an unexplained refusal*. Either add one arm, or say plainly in the arm doc that Worldspace is pinned by construction and not exercised.

---

### Aside — out of scope, possibly worth its own issue

The generated `mutagen-reference` classifies **`Cell.Landscape` as `"c":"formlink"`** with no target (`.claude/skills/mutagen-reference/references/records.jsonl:17`). This branch's walk explicitly cuts `IFormLinkGetter`, and its all-types pin asserts `Landscape` **is** in Cell's child-bearing set — so on the concrete type it is an owned record, not a link. Those two cannot both be right. A caller trusting the reference would try to set a FormID on an owned record. Not this PR's problem; worth a look.

---

**Bottom line.** Fold #1 (a dry-run probe, ~6 lines) and #4 (one clause in the tool description). #2 is a comment correction that matters because it is the evidence for a filed bug. #3, #5, #6, #7 are take-or-leave — per §2 fix-or-drop, not issue material.

🤖 Review run with [Claude Code](https://claude.com/claude-code)

PR #337, opened 2026-08-12
refactor: one source per sentence for the write-surface responses

Avick3110:
Implements `dev/projects/tool-surface-2.0/RESPONSE_LAYER_BY_CONSTRUCTION_2026-08-11.md` (chartered
Aaron-go 2026-08-11). No linked issue — this lane is chartered from the plan doc, not the bug lane, so
there is no closing keyword to carry.

## What this is

Every write tool renders its outcome twice — text via `WriteTools`/`SeqTools`, json via `JsonWire` (D2:
one write path, two renders) — and several text renders repeat the same sentence per verb on top of
that. Each copy was an independent statement of one meaning. The 2.0 review wave's single largest
finding class was those copies drifting.

`WriteSentences` now holds one source per sentence, read by both transports. Copies were deleted in the
same commit that added each shared member (the plan's replace-not-add rule).

## Inventory — the completeness claim

The plan makes the enumerated leftovers the claim rather than a silent "done".

**36 shared members**, 15 of them in `WriteSentences.Twins` and parity-enforced. The other 21 are
single-sourced but not verbatim-comparable across lanes — either prose on one transport by design (the
in-place hazard, which json states as `lane` plus typed flags) or parameterised (a cap, a filename, a
read-back call).

**9 disagreements resolved.** Each was two copies of one sentence that had already drifted:

| Sentence | The drift | Kept |
|---|---|---|
| grid-occupancy note | json had lost `(engine behavior undefined)` | the fuller — it tells the caller a retry will not help |
| report-block cut advice | text "do not re-issue the **create** — that allocates the records again" vs json "do not re-issue the write" | text's; these blocks ride create only |
| `write_seq` standing limit | json dropped the `housecarl_validate_dialogue` pointer | the fuller — that is the what-next half |
| dry-run closing line | forward "disk faults" vs apply "disk faults **and data Mutagen refuses to serialize**" | the fuller; both verbs commit through the same serializer |
| in-place hazard | create's copy had lost "was rewritten" | the four-site majority |
| in-place mod folder | compact's had lost "in your load order" | the majority |
| script-binding stake | "that's" vs "that is" | one spelling |
| voice stake | "a response with no .fuz" vs "a created voiced response with NO .fuz" | the json reading |
| `write_seq` unchanged / three replace readings / no-epoch | paraphrases across lanes | one reading each |

Plus one deixis resolution: `write_seq`'s list-cut remedy said "name into=/output_dir= **the folder
below**", a reference into a line only the text render prints. A shared sentence cannot carry that, so
the lane-neutral reading stands.

**3 behaviour changes:**

1. **`apply`'s text ops loop is now budgeted** (real and dry-run paths). Its json twin has always
   dropped trailing rows at `max_chars` and closed `truncated:true`; the text render listed every edit
   and let the host cut the oversized response out of band — the silent cut `max_chars` promises never
   happens. This was the last unbounded row list on the write surface.
2. **A truncated dry run no longer describes itself as a completed write.** The json cut note said "the
   WRITE is complete and unaffected" on a lane that writes nothing.
3. **create's in-place headline** moved from `X.esp rewritten IN PLACE` to `created into X.esp IN PLACE` —
   every sibling is verb-then-file, and create's used "rewritten" as its verb, which stutters against the
   shared hazard clause now behind it. Added here on review finding 6: it is a wording change rather than
   a migration, and the inventory is the completeness claim.

**13 guard arms** moved from fragment pins to construction pins. `write-surface-guard` 169 → 191.

**No behaviour change from the A/B/C fold** — `SeqStandingLimit`'s sentence is unchanged; only its declared
phrases moved from a bare token to the action. A/C are guard-side.

**7 remainders, named with reasons:** the engine-side consent prompts (`LoadOrderService.cs:1482`,
`:4919`, `:5547` — three spellings of the in-place hazard; deferred rather than half-folded, they need
their own resolution); the read surface (explicitly a later PR per the plan — `JsonWire.Cap` was briefly
wired through `WriteSentences.Cap` here and deliberately unwired again); tool `[Description(...)]`
strings (the tool schema is a different surface); the truncation-notice frame (every claim-bearing part
is shared; what repeats carries no claim); `compact`/`merge` internals (1.x renders outside the plan's
five verbs); and the executability half of plan item 2 (`SeqListCutRemedy` names `into=`/`output_dir=`
and no arm proves either resolves — a real follow-up); and **the parameterised shared sentences** — see below.

### Known and not fixed: the parameterised helpers are outside the `[MustState]` net

The A/B/C fold closes the net over every `const string` on both classes: absence now fails. It does **not**
cover the ~14 parameterised helpers (`DryRunClose`, `NewOrExtendedArtifact`, `InPlaceModFolder`,
`RowsCutOperationIntact`, `CreateRowsCutRemedy`, `CheckCouldNotRun`, `Masters`, …), because a sentence that
interpolates a cap, a filename or a verb has no invariant string to compare. Several carry real claims —
`DryRunClose`'s "(A real write can still fail at serialize/commit …)" is a caveat nothing pins.

This is the declared boundary of the mechanism rather than a new hole: it was in this inventory from the
first push, and it predates the opt-in model. Closing it needs a different shape — pinning invariant
FRAGMENTS of a parameterised sentence — which is a design decision, not a fold. Flagged here now that the
const half is closed, because that is what makes this the visible edge of the map.

## The D2 twin harness

One outcome, both transports, all five verbs. It asserts per-lane coverage of the shared inventory —
every twin observed coming out of text at least once and json at least once, by reflection, so a new
member is enrolled automatically — plus budget parity (a cap that truncates one lane truncates the
other) and count parity.

Deliberately coverage-per-lane rather than co-occurrence-per-outcome: a few of these land in different
places on the two lanes (text report blocks state their stake in the block header, json blocks in the
truncation census), so demanding co-occurrence would fail honest renders.

RED-checked in both directions: re-inlining two json copies fails with
`unobserved: GridOccupancy, SeqStandingLimit`; making the text created-rows list unbounded fails with
`text=False json=True`, the historical bug shape.

## Review rounds

**Round 1** — three fresh agents (correctness, guards, completeness). The guards agent was killed by a
power cut and never reported; see the incident below. Of the two that landed: 1 high, 5 medium, 6 low
folded. The three that mattered were the branch's own rule broken where it had not looked — create's
row-cut remedy still stated twice and already drifted; the "rows were cut, the operation was not" claim
two literals, with json asserting a completed write on a truncated dry run; and `apply`'s unbudgeted
text ops loop, which the twin arm missed because `apply` was not enrolled in it.

Refused, with reasons: naming `seq_path` inside the shared replace notes (the text lane has no such
member and prints `path:` itself; mid-sentence lane injection would restore the per-site prose this
removes); folding one of the three engine-side consent prompts (worse than naming all three); proving
`into=`/`output_dir=` executable from the cut remedy (a real probe, scoped as a follow-up).

**Round 2** — two fresh agents (guards, output-composition). Both independently found the incident
below. Otherwise: 2 medium, 6 low folded, no refusals — the findings were reproducible in the source and
several were defects the round-1 fold had introduced.

**Stopped at two rounds.** Round 1's class was "incomplete migration"; round 2's was "the construction
pins are content-blind". Two consecutive rounds on one seam — the migration's own verification posture —
is the §5 #11 convergence signal, so no round 3.

## Incident: three sentences shipped gutted, and the guards could not see it

Commit `780ac3a` contained three `Twins` constants replaced by placeholders — `GridOccupancy` became
`"the cell was created."`, `VoiceStake` became `"voice coverage was not fully rendered"`, and
`SeqTimestampRefreshed` lost the statement that the mtime had been stamped forward at all. All three
were correct in `356cd35`.

I did not edit them. The round-1 guards agent break-tests by design — break a constant, confirm RED,
restore — and the power cut killed it mid-test, leaving its sabotage in the shared worktree. I then ran
`git add -A` and committed without reading the diff.

The full suite stayed green because that same commit migrated those arms from `Contains("<fragment>")`
to `Contains(TheConstant)`. **A construction pin proves the render reads the shared source and nothing
about what the source says** — render and assertion read the same symbol, so it passes on a constant
that has been emptied. That trade was the real defect; the sabotage only exposed a hole already built.

Fixed in `fdd67a3`: constants restored, and the content half put back where it belongs — beside the
sentence rather than as a second copy in a probe. Every twin now carries `[MustState(...)]` naming the
claims it must keep. The arm fails if a phrase is missing, if a member declares none, or if a member is
a shape the reflection cannot read (a `static readonly` or a nested type would otherwise be an unchecked
twin counted as covered — round 2 demonstrated both). RED-checked by replaying the exact sabotage: it
now fails by name, listing all four missing phrases.

Two process points that are yours to call, not mine:

- **Review agents should run in isolated worktrees.** The agents that break-test are the ones most
  likely to corrupt the branch, and they had the same tree I commit from. The Agent tool has
  `isolation: "worktree"` for this. Making it standing practice is a §5 #11 change — self-governing
  config, so surfaced rather than self-committed.
- **`MustState` is worth your own eye.** It is now the thing standing between a shared sentence and
  silent evisceration.

## Review fold (Aaron's review, 6 findings)

**3 and 4 — phrase quality.** Aaron directed these be fixed by writing better phrases and recording the
norm, **not** by making the checker smarter: *a phrase carries the hazard or the action; a phrase that
survives its sentence's negation is not a claim.* That norm now lives in `MustStateAttribute`'s doc,
together with why it stays a norm — judgment inside a comparator is #308's shape, and a future round
proposing claim-quality validation in reflection is that mistake wearing new clothes. Keeping the line
visible is the point.

Applying the norm across the whole set caught four more of the same two kinds beyond the two named: bare
topic tokens (`"grid-occupancy"`, `"load-order-independent"`, `"Start Game Enabled"`) and a phrase that
survives its own negation (`"stamped forward"`). Six rephrased.

**1 — the in-place hazard was outside the net.** The loudest sentence on the surface, and this change is
what made it a single token: one edit could have stripped it from all five in-place renders with
everything green. It now carries `[MustState]`, the content check walks the outer class as well as
`Twins`, and the lane arm asserts the sentence reaches the caller.

**2 and 5 — checker-mechanical, fixed in code.** The shape alarm now covers methods and events (a
method-shaped twin was invisible to both the inventory and the alarm); an empty phrase or empty sentence
is rejected rather than satisfying every assertion.

**6 — undeclared behaviour change.** Folded into the inventory above.

RED-checked all four: emptying `NoBackupOrUndo`, declaring an empty phrase, adding a method-shaped twin,
and — replaying findings 3 and 4 verbatim — negating the replace note and rewording the `into=` half all
now fail by name. That last pair is the evidence that better phrases were the right lever.

## Verification

`ci-all` 117/117 · `write-surface-guard` 190/190 · `apply-guard` 78/78 · `seq-write-guard` all arms, in
Release.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -15,9 +15,16 @@ way. The most consequential: the JSON note about creating an exterior cell had l
 collision is *engine* behaviour, so it read like something you could retry your way out of; a cut report block told
 you "do not re-issue the write" on one transport and the more specific, actually-correct "do not re-issue the create —
 that allocates the records again" on the other; and `write_seq`'s JSON standing-limit note had dropped the pointer to
-`housecarl_validate_dialogue`, which is the half telling you what to do next. Every such sentence now has one source
-that both transports read, so a correction reaches both or neither. Wording is otherwise unchanged, except where the
-two copies already disagreed — those are resolved to the fuller and truer reading.
+`housecarl_validate_dialogue`, which is the half telling you what to do next. Every sentence named in the PR's inventory now has one
+source that both transports read, so a correction reaches both or neither; the inventory also names what is left and
+why. Wording is otherwise unchanged, except where the two copies already disagreed — those are resolved to the fuller
+and truer reading.
+
+**Fixed: a large `apply` no longer overflows its text response without saying so.** Its JSON twin has always dropped
+trailing rows at `max_chars` and set `truncated`, but the text render listed every edit and let the host cut the
+oversized response out of band — the silent cut `max_chars` exists to prevent, on the one write verb whose row list
+was still unbounded. It now stops with the same explicit notice its four sibling renders carry. A truncated dry run
+also no longer describes itself as a completed write.
 
 **Fixed: forwarding a record no longer deletes the records nested under it (#324).** When `forward` targeted a record
 the destination *already carried*, it replaced it by dropping the whole record and copying the source body in — and
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -8,6 +8,17 @@ when it changes.
 
 *Accumulating notes for the next cut — not yet released; `plugin.json` still reads the last shipped version.*
 
+**Fixed: the text and JSON responses to a write can no longer say different things.** Every write tool renders its
+result twice — once as text, once as `format="json"` — and the two renders were maintained separately, so a warning
+or a piece of advice could be corrected on one and left stale on the other. Nine sentences had already drifted that
+way. The most consequential: the JSON note about creating an exterior cell had lost the clause saying a grid
+collision is *engine* behaviour, so it read like something you could retry your way out of; a cut report block told
+you "do not re-issue the write" on one transport and the more specific, actually-correct "do not re-issue the create —
+that allocates the records again" on the other; and `write_seq`'s JSON standing-limit note had dropped the pointer to
+`housecarl_validate_dialogue`, which is the half telling you what to do next. Every such sentence now has one source
+that both transports read, so a correction reaches both or neither. Wording is otherwise unchanged, except where the
+two copies already disagreed — those are resolved to the fuller and truer reading.
+
 **Fixed: forwarding a record no longer deletes the records nested under it (#324).** When `forward` targeted a record
 the destination *already carried*, it replaced it by dropping the whole record and copying the source body in — and
 the drop took the record's child group with it. Any dialogue line under a forwarded topic, or placed reference under
--- a/src/housecarl-generator/ApplyGuardProbe.cs
+++ b/src/housecarl-generator/ApplyGuardProbe.cs
@@ -616,7 +616,7 @@ public static class ApplyGuardProbe
         // dry_run on the default lane writes nothing and says so first
         var dry = ApplyTools.Apply(fx.Svc, ops: Json(OneOp("41")), patch: "ApDry", dry_run: true);
         Check("dry_run reports what WOULD change and writes nothing",
-            dry.StartsWith("DRY RUN") && dry.Contains("NOTHING was written"), dry);
+            dry.StartsWith(WriteSentences.DryRunHeader, StringComparison.Ordinal), dry);
     }
 
     // ================= ARM 3 — the §4.5 zip =================
--- a/src/housecarl-generator/DryRunProbe.cs
+++ b/src/housecarl-generator/DryRunProbe.cs
@@ -221,7 +221,9 @@ internal static class DryRunProbe
                     $"dry forward reports the would-be copy (source={o.Forwarded.FirstOrDefault()?.FromPlugin}, priorWinner={o.Forwarded.FirstOrDefault()?.PriorWinner})  [{o.Error ?? "ok"}]");
                 Check(before.SequenceEqual(ModFolders()), "no mod folder appeared");
                 var text = WriteTools.RenderForward(o);
-                Check(text.Contains("DRY RUN") && text.Contains("would be copied from") && !text.Contains("\nmasters:"),
+                Check(text.Contains(WriteSentences.DryRunHeader, StringComparison.Ordinal)
+                      && text.Contains("would be copied from")
+                      && !text.Contains("\n" + WriteSentences.Masters(Array.Empty<string>())),
                     "forward render uses the would-be phrasing (+ the expected-masters preview, not a real header)");
                 var dryMiss = svc.ForwardRecords(new[] { fid2 }, "HcDryUser.esp", "DryH2", null, dryRun: true);
                 var realMiss = svc.ForwardRecords(new[] { fid2 }, "HcDryUser.esp", "DryH2", null);
--- a/src/housecarl-generator/DryRunProbe.cs
+++ b/src/housecarl-generator/DryRunProbe.cs
@@ -223,7 +223,10 @@ internal static class DryRunProbe
                 var text = WriteTools.RenderForward(o);
                 Check(text.Contains(WriteSentences.DryRunHeader, StringComparison.Ordinal)
                       && text.Contains("would be copied from")
-                      && !text.Contains("\n" + WriteSentences.Masters(Array.Empty<string>())),
+                      // Any masters LINE, not the empty-set spelling. This outcome forwards from a REAL master, so
+                      // pinning "masters: (none)" would be a string the render can never emit — an assertion that
+                      // passes whatever happens, which is what it briefly became.
+                      && !text.Contains("\n" + WriteSentences.Masters(Array.Empty<string>()).Split('(')[0]),
                     "forward render uses the would-be phrasing (+ the expected-masters preview, not a real header)");
                 var dryMiss = svc.ForwardRecords(new[] { fid2 }, "HcDryUser.esp", "DryH2", null, dryRun: true);
                 var realMiss = svc.ForwardRecords(new[] { fid2 }, "HcDryUser.esp", "DryH2", null);
--- a/src/housecarl-generator/SeqWriteGuardProbe.cs
+++ b/src/housecarl-generator/SeqWriteGuardProbe.cs
@@ -377,12 +377,15 @@ internal static class SeqWriteGuardProbe
                   && File.GetLastWriteTimeUtc(expectedUserSeq) > pluginStamp
                   && File.ReadAllBytes(expectedUserSeq).Length == 4,
                 $"MTIME-REFRESH a byte-identical but OLDER .seq is stamped forward, not rewritten — unchanged={oStale.Unchanged} touched={oStale.TimestampRefreshed} seq={File.GetLastWriteTimeUtc(expectedUserSeq):O} plugin={pluginStamp:O}");
+            // A CONSTRUCTION pin, not a fragment one: the stamp sentence has one source, so this asserts (a) the
+            // source itself still carries the two claims that matter — it names validate_dialogue's check, and it
+            // BOUNDS the promise to the copy the load order serves (review round 2: the lint reads the .seq the VFS
+            // SERVES, which is this file only when this folder wins the SEQ\ conflict) — and (b) the render is
+            // reading that source. A fragment check could only ever have pinned one render's spelling of it.
             var renderStale = SeqTools.Render(oStale);
-            Check(renderStale.Contains("timestamp was refreshed", StringComparison.Ordinal)
-                  && renderStale.Contains("validate_dialogue", StringComparison.Ordinal)
-                  // …and it does NOT promise that tool's verdict: the lint reads the .seq the VFS SERVES, which is
-                  // this file only when this folder wins the SEQ\ conflict (review round 2).
-                  && renderStale.Contains("provided this is the copy your load order actually serves", StringComparison.Ordinal),
+            Check(WriteSentences.Twins.SeqTimestampRefreshed.Contains("validate_dialogue", StringComparison.Ordinal)
+                  && WriteSentences.Twins.SeqTimestampRefreshed.Contains("the copy the load order actually serves", StringComparison.Ordinal)
+                  && renderStale.Contains(WriteSentences.Twins.SeqTimestampRefreshed, StringComparison.Ordinal),
                 $"MTIME-REFRESH-RENDER the stamp is STATED and its scope is bounded — render=[{Trim(renderStale)}]");
 
             // MTIME-FUTURE: a plugin stamped in the FUTURE (restored backup, skewed clock) cannot be beaten by
@@ -423,7 +426,10 @@ internal static class SeqWriteGuardProbe
             var renderRepl = SeqTools.Render(oRepl);
             Check(oRepl.Success && !oRepl.Unchanged && oRepl.Replaced
                   && renderRepl.StartsWith("replaced ", StringComparison.Ordinal)
-                  && renderRepl.Contains("no backup is kept", StringComparison.Ordinal),
+                  // The no-backup alarm is a property of the SENTENCE (one source, both transports), so pin it there
+                  // and pin that this render reads it — rather than pinning one lane's spelling of the alarm.
+                  && WriteSentences.Twins.SeqReplacedUserFolder.Contains("keeps no backup", StringComparison.Ordinal)
+                  && renderRepl.Contains(WriteSentences.Twins.SeqReplacedUserFolder, StringComparison.Ordinal),
                 $"REPLACED overwriting an existing .seq reports 'replaced', not 'wrote' — replaced={oRepl.Replaced} render=[{Trim(renderRepl)}]");
             // …and the fresh-file case is NOT reported as a replacement (the flag has to discriminate).
             File.Delete(expectedUserSeq);
--- a/src/housecarl-generator/SeqWriteGuardProbe.cs
+++ b/src/housecarl-generator/SeqWriteGuardProbe.cs
@@ -1,3 +1,4 @@
+using System.Text.Json;
 using Mutagen.Bethesda;
 using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Skyrim;
@@ -412,7 +413,7 @@ internal static class SeqWriteGuardProbe
             // silent-success shape Q3 refuses, and the first line is what a caller reads.
             var toolSame = SeqTools.WriteSeq(svc, source: Path.GetFileName(svcPlugin), output_dir: userMod);
             Check(toolSame.StartsWith("unchanged", StringComparison.Ordinal)
-                  && toolSame.Contains("NOTHING was written", StringComparison.Ordinal)
+                  && toolSame.Contains(WriteSentences.Twins.SeqUnchanged, StringComparison.Ordinal)
                   && !toolSame.Contains("houseCARL mod folder — enable it", StringComparison.Ordinal),
                 $"RENDER-UNCHANGED the no-op renders as 'unchanged … NOTHING was written', and an output_dir destination is NOT called a houseCARL mod folder — render=[{Trim(toolSame)}]");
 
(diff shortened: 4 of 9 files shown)


Avick3110:
## Review — focus: `[MustState]`

Single careful pass over `origin/main...origin/claude/response-layer`. Review focus was the `[MustState]`
mechanism specifically, so the findings below concentrate on what that net does and does not catch.

### What holds

The `Twins` inventory is read by reflection rather than listed in the probe, so there is no second copy of
the set to fall behind. `TwinContentViolations` does catch the documented sabotage — I checked all 15
members' declared phrases against their sentences and every one matches today. `TwinShapesUnreadable`
genuinely closes the `static readonly` and nested-type escapes. Per-lane coverage is real: each of the 15
traces to a text render site and a json render site in `TwinParityArm`, and the report-block twins are
observed off the *capped* pass, which is the only render where the json census states them.

### Findings

**1. The in-place hazard is outside the net entirely — no `[MustState]`, and no probe asserts it.**
`src/housecarl-mcp/WriteSentences.cs:93`

`NoBackupOrUndo`, composed into `InPlaceRewritten` and `InPlaceWouldRewrite`, is on the outer
`WriteSentences` class, so `TwinContentViolations` never walks it. And grepping `src/housecarl-generator/`
for the phrase returns exactly one hit — a doc comment at `WriteSurfaceGuardProbe.cs:1091`. No assertion
anywhere states it.

So: empty that one const and all five in-place text renders (`WriteTools.cs:541`, `:763`, `:834`, `:940`,
`:1145`) plus the dry-run in-place line (`WriteSentences.cs:126`) stop telling the caller that houseCARL
keeps nothing to undo the rewrite of their own plugin — with `ci-all`, `write-surface-guard`,
`apply-guard` and `seq-write-guard` all green. That is the incident's exact shape, on the loudest sentence
on the write surface, and this PR is what made it a single token: before, the same edit meant touching
five separate literals in five files.

**2. `TwinShapesUnreadable` does not check methods.**
`src/housecarl-generator/WriteSurfaceGuardProbe.cs:2258`

It inspects `GetFields`, `GetProperties` and `GetNestedTypes` — not `GetMethods`/`GetEvents`. A
method-shaped member of `Twins` (e.g. `internal static string SeqDeployStake(string folder) => …`, which
is one interpolation away from several existing twins) is invisible to the inventory *and* to the
unreadable-shape alarm: `TwinSentences()` returns the same 15, `TwinShapesUnreadable()` returns empty, and
the coverage assertions report "every member rendered" for a set they never saw. The method's own doc and
the PR body both claim that case fails.

**3. `[MustState("into=")]` pins a parameter token, not a claim.**
`src/housecarl-mcp/WriteSentences.cs:230`

`CreateReissueTrap`'s second phrase is satisfied by any sentence that mentions `into=` at all. Reword the
into= half to "under into= the call is idempotent, so a repeat is free" and the load-bearing claim — each
record re-created at its old FormID with its **prior contents discarded** — is gone, while
`TwinContentViolations` passes, the construction pins at `WriteSurfaceGuardProbe.cs:1826-1827` pass (same
constant on both sides), and the coverage arm passes. The attribute's own doc says phrases are the CLAIM,
not the wording; this is the check missing its own contract on the more expensive of the trap's two lanes.

**4. `[MustState("overwritten")]` survives the negation of its own sentence.**
`src/housecarl-mcp/WriteSentences.cs:269`

One common word cannot distinguish "was overwritten" from "was NOT overwritten". Edit
`SeqReplacedOwnFolder` to the latter and the guard reports no violation, while both transports then tell a
caller their regenerate left the old `.seq` in place when `write_seq` in fact replaced it.

**5. Empty phrases and empty sentences pass all three assertions.**
`src/housecarl-generator/WriteSurfaceGuardProbe.cs:2283`

The check rejects a missing attribute and `Phrases.Length == 0`, but not an empty phrase string or an
empty sentence — `"".Contains("")` is true, and `Observe`'s `text.Contains(sentence)` /
`jsonText.Contains(sentence)` are equally vacuous on `""`. `[MustState("")] internal const string
VoiceStake = "";` clears `TwinContentViolations`, `TwinShapesUnreadable`, both coverage sets, and
`WriteSurfaceGuardProbe.cs:1685`'s `vnote.Contains(...)`. Needs the attribute touched in the same edit, so
it is not the sabotage replay — but it is the residual hole in the backstop.

**6. Undeclared behaviour change: create's in-place headline.**
`src/housecarl-mcp/WriteTools.cs:1144`

`"<file> rewritten IN PLACE"` → `"created into <file> IN PLACE"`. That is a user-facing wording change,
not a migration of a duplicated sentence to a shared source, and it is not among the PR body's enumerated
**2 behaviour changes** — which the plan makes the completeness claim. Either fold it into the inventory or
drop it.

### Checked and dropped

- **The dry-run construction pins.** `ApplyGuardProbe.cs:619` moved to
  `dry.StartsWith(WriteSentences.DryRunHeader)`, which is vacuous if the header is emptied — but
  `DryRunProbe.cs:128` still pins `"DRY RUN"` and `"NOTHING was written"` literally, and
  `WriteSurfaceGuardProbe.cs:685`/`:831` plus `ExcludedMasterWriteProbe.cs:133`/`:615` pin `"DRY RUN"`. The
  sentence is weakened, not holed. Not filed.
- **`RowsCutOperationIntact`.** `WriteSurfaceGuardProbe.cs:1810` went construction-only for the create
  branch, but `:1759` and `:1874` still pin the shared function's output literally, so the function's
  content is covered. Not filed.
- **The dry-run truncation remedy.** `apply`'s new dry-run cut notice calls `ApplyAgainRemedy`, and
  `WriteAgainRemedy` returns "raise max_chars to see the rest" on `dryRun` — it does *not* point at an
  `into=` naming the preview filename the header just warned can shift. Correct as written.
- **Twin coverage sufficiency.** Traced each of the 15 to a real text site and json site, including the
  states that only exist on a cut render (`CreateReissueTrap`, `ReportBlockCut`, `SeqListCutRemedy`) and
  the json census clauses (`VoiceStake`, `ScriptStake`, `CellStake`). No gaps.

Avick3110:
## Re-review of `79d2ed5` — focus: `[MustState]`

Verified independently rather than from the commit message: built the branch worktree in Release (clean —
18 warnings, all pre-existing on `main`) and ran the guard myself.

```
=== write-surface-guard: 190 passed, 0 failed -> PASS ===
```

### All six folded, and the fixes are the right shape

**1 — the in-place hazard.** Properly closed, and closed on both halves. `InPlaceRewritten`,
`InPlaceWouldRewrite` and `NoBackupOrUndo` now carry `[MustState]`; `TwinContentViolations` walks the outer
class as well as `Twins` (`WriteSurfaceGuardProbe.cs:2294`); and the new lane arm at `:603` asserts the
create in-place render still emits the sentence, which is the half a content pin cannot provide. This was
the finding that mattered.

**2 — shape alarm.** `GetMethods`/`GetEvents` added, with `BindingFlags.DeclaredOnly` folded into `All` so
`object`'s inherited members do not false-positive the alarm. Correct, and the `DeclaredOnly` detail is the
part that would have bitten.

**3 and 4 — phrase quality.** Fixing these with better phrases rather than a smarter comparator is the right
call, and recording *why* it stays a norm — #308's shape — is worth more than the fix. I re-audited all 15
phrase sets against the new norm: every rewritten phrase is present in its sentence, and the six rewrites
are genuine improvements. `CreateReissueTrap` in particular now pins both lanes' consequences
(`"second full patch"` + `"prior contents discarded"`) where it previously pinned neither. The four the
self-directed sweep found beyond the two I named (`"grid-occupancy"`, `"load-order-independent"`,
`"Start Game Enabled"`, `"stamped forward"`) are all real instances of the same two kinds.

**5 — empty phrase / empty sentence.** Both rejected, with the loop scoping correct (`continue` inside the
inner `foreach`, so one empty phrase does not skip the rest).

**6 — undeclared behaviour change.** Declared in the inventory as behaviour change 3.

### Three residuals — same class as before, none blocking

**A. The outer-class walk is opt-in, so `CellRowsCutLoss` is silently unchecked.**
`src/housecarl-mcp/WriteSentences.cs:215`

`requireAttr: false` for the outer class (`WriteSurfaceGuardProbe.cs:2294`) means a const *without*
`[MustState]` is skipped rather than flagged. `CellRowsCutLoss` is exactly that shape: claim-bearing, no
attribute, and grepping the generator tree for "Creation-Kit work this response" returns no assertion
either. Empty it and the json cell-shell census (`JsonWire.cs:2135`) stops telling the caller that each
dropped row is Creation-Kit work the response was supposed to name — the one fact that separates a cut cell
block from a cut ordinary row list — with the guard still at 190/190.

This is finding 1's hole one level up: the mechanism now covers the outer class wherever an author
remembered to opt in. `DryRunHeader` sits in the same position and survives only by accident —
`DryRunProbe.cs:128` still pins `"DRY RUN"` and `"NOTHING was written"` as literals.

Cheapest fix is the attribute on both. If instead the intent is that only *some* outer-class sentences
have claims worth pinning, then the skip should be a declared opt-out rather than the absence of an
attribute — otherwise the next sentence added there inherits this silence.

**B. `SeqStandingLimit` kept the bare tool-name token the sweep was hunting.**
`src/housecarl-mcp/WriteSentences.cs:323`

`[MustState("housecarl_validate_dialogue", "does not verify")]` — the first phrase is a bare topic token,
and the *identical* token was rewritten to `"housecarl_validate_dialogue's SEQ staleness check"` in
`SeqTimestampRefreshed` in this same commit. Reword the sentence to "…and confirms the dialogue graph is
sound (housecarl_validate_dialogue not required)" and both declared phrases survive while the Q3 standing
limit has inverted into a claim `write_seq` cannot make. By the norm now documented at
`WriteSentences.cs:15` the phrase is the action: `"use housecarl_validate_dialogue for the dialogue graph"`.

**C. `InPlaceWouldRewrite` got the content pin but not the wiring pin.**
`src/housecarl-mcp/WriteSentences.cs:111`

All three hazard consts got `[MustState]`, but only `InPlaceRewritten` got a render-reaches assertion
(`:603`). Nothing anywhere asserts that any render still emits the dry-run tense. If
`DryRunWouldWrite`'s in-place arm (`WriteSentences.cs:126`) were edited to inline its own weaker copy —
the drift this PR exists to prevent, and precisely what had already happened to create's copy before the
migration — the const stays intact, `TwinContentViolations` passes, and `apply`/`forward` dry runs on the
in-place lane stop warning that the caller's original file would be rewritten with no way back. That is
finding 1's own argument left half-applied to the sibling.

---

All three are one attribute, one phrase, and one `Check` line respectively.

Avick3110:
## Fold report — residuals A/B/C (`398328a`)

All three folded as directed. `write-surface-guard` 190 → **191**; `ci-all` 117/117, `apply-guard` 78/78,
`seq-write-guard` all arms, Release.

### A — absence stops being silence

Took the second option, not the first. Decorating the two named sentences would have closed two holes and
left the generator running: as the review put it, *"the next sentence added there inherits this silence."*

The outer-class walk no longer skips an undecorated const. Every `const string` on **both** classes must now
DECIDE — `[MustState]` phrases, or the new `[NoClaims("reason")]` opt-out — and one that declares neither
fails by name. Declaring both also fails, as does `[NoClaims]` with an empty reason.

`DryRunHeader` → `[MustState("NOTHING was written", "originals untouched")]`. Deliberately not `"DRY RUN"`:
the label survives a rewrite that says the opposite, the two claims do not.
`CellRowsCutLoss` → `[MustState("Creation-Kit work this response was supposed to name")]`.

**Nothing opts out.** All five outer consts carry claims, so `[NoClaims]` has no users yet — which is the
healthy state for an escape hatch, and worth saying rather than inventing a use for it.

### B — the phrase is the action

`[MustState("does not verify", "use housecarl_validate_dialogue for the dialogue graph")]`. The sentence is
untouched; only the declared phrases changed. The bare token was the same one rewritten in
`SeqTimestampRefreshed` one commit earlier — the sweep had missed its sibling.

### C — enumeration was cheap, so it closes the class

Answering the question first: **yes.** `OuterSentences()` reflects the outer-class `[MustState]` members
exactly as the `Twins` walk reflects its set, and the arm asserts each reaches at least one render — *either*
transport, since these are one-transport-by-design and "reaches both" would be false for `CellRowsCutLoss`.

So the fix is the class, not the sentence: an outer-class sentence added later cannot get a content pin and
silently miss the wiring one. The two outcomes it needed (in-place, in-place dry run) are **built** rather
than driven through the fixture — the real in-place lanes rewrite a file the later arms read as a known
winner, and this arm runs before them.

### RED evidence

| Sabotage | Result |
|---|---|
| Undecorated claim-bearing const added to the outer class | `NewlyAddedHazard: declares neither [MustState] phrases nor [NoClaims] with a reason` |
| `CellRowsCutLoss` emptied (A's named sentence) | `CellRowsCutLoss: no longer states "Creation-Kit work this response was supposed to name"` |
| The review's exact inversion of the standing limit | `SeqStandingLimit: no longer states "does not verify"; … "use housecarl_validate_dialogue for the dialogue graph"` |
| `DryRunWouldWrite` inlines its own weaker copy (C's scenario) | `unobserved: InPlaceWouldRewrite` |

The C run failed to reproduce on first attempt — because the sabotage never applied, not because the guard
missed it. Re-run after confirming the edit had landed. Worth stating: an unverified negative result is the
same class of false evidence as #333.

### Audits run, nothing found

- **All 20 phrase sets** re-checked against the documented norm for the token class. No fourth instance —
  every phrase carries a hazard, an action, or a consequence that disappears under negation.
- **All five outer-class consts** enumerated. Three were already decorated, two were the ones named. No
  fourth sentence outside the net among the consts.

### One thing surfaced rather than folded

The net now covers every `const string`. It does **not** cover the ~14 **parameterised** helpers
(`DryRunClose`, `NewOrExtendedArtifact`, `CheckCouldNotRun`, …) — a sentence interpolating a cap or filename
has no invariant string to compare, and several carry real claims (`DryRunClose`'s serialize/commit caveat is
pinned by nothing).

Not folded, and I don't read it as the fourth-instance trigger: it is the mechanism's declared boundary, in
the inventory since the first push and predating the opt-in model, not a hole the model generated. But it is
now the visible edge of the map, and closing it needs a different shape — pinning invariant FRAGMENTS of a
parameterised sentence — which is a design call. Recorded in the PR body's known-and-not-fixed list.

Stopping here. Merge decision is yours.

PR #339, opened 2026-08-12
place_asset: a Data-relative source, and a sigiled pole that says whose copy

Avick3110:
`place_asset` could only read a source it was handed as an exact on-disk path, and no read tool emits one — `asset_status` reports provider *names*. So nothing could express "the copy of this Data-relative path that mod X supplies", which is the shape an NPC appearance copy is built from.

This is the source primitive split out of PR 3b, not the `copy` verb. `copy` has not started.

No linked issue — this lane charters from plan docs (same as #337).

## The settle question that produced this branch

PR 3b's charter said to settle first whether the FormID-keyed facegen rename is already expressible, because that decides whether `copy` has to carry it. Measured against the source rather than argued:

- **The destination side already renames.** `place_asset` takes source and destination independently; the destination can be `formid`+`kind`, computed by `FaceGenPath.For`. Pass the *new* FormKey and you get the new facegen name. Nothing to build.
- **The source side could not name what to read.** `source=` was `<abs .bsa>|<entry>`, a bare `.bsa`, or `Path.GetFullPath` against the process working directory. The concrete `PlacementSource` exists in core but is reachable only by `place_asset`'s own auto-resolve — which resolves the *destination*, and a rename's destination is new, so nothing provides it.

So the answer was neither of the two the charter anticipated: **the gap is real, and it belongs on the asset surface, not in `copy`.** `copy` carries nothing.

## What changed

**`source=` accepts a Data-relative path**, resolved through the VFS. **`source_provider=` picks the pole** — the three S2 SOURCE poles from SPEC §1: the VFS winner, a named provider, or (omitted) the sole provider with contention refused. Source path ≠ destination path is therefore a **rename**.

**One shared policy in core** (`AssetSourceSelection`) makes the pick. `NpcAppearanceAssets.ResolveForCarry` now goes through it too, so "how to read a contended source" has one spelling rather than two.

**Aaron's semantics, as built:** a named provider is honoured *whatever else contends* — the caller always names the donor mod, so this is a named-provider read, not a winner read and not an ambiguity refusal.

### Known, intentional divergence

`copy_npc_appearance`'s facegen arm keeps its **winner-read** (`alwaysCarry`). Where a replacer out-sorts the donor, the old tool carries the winner's bytes — facegen that may not match the donor's head-part and tint records. The successor flow reads the **named donor**. Left divergent on purpose: that tool is being subsumed, and the difference is what PR 3b's differential acceptance measures. Commented where it lives, and now pinned by an arm.

## The grammar decision

`source_provider=` originally spelled the winner pole as the bare word `winner` — which is also a legal MO2 folder name. The overlap was handled where names are *matched* and unknown everywhere names are *listed*. That produced, in sequence: a collision verdict, a `TokenNameTaken` flag, a conditional sentence, a bespoke refusal, and two holes no guard covered — including an ambiguity refusal that listed `winner` as a token its own remedy then rejected.

Escalated under §11; **Aaron's ruling: sigil it.** The pole is `*winner`. `*` is illegal in a Windows file or folder name, so the pole space and the provider-name space are disjoint by construction, a bare name always means a provider of that name, and a mod actually called `winner` is reachable. Same idiom as `@file` in `ops=`. The collision machinery is **deleted, not stranded**; both listing holes close by construction.

Recorded as **SPEC §5.5**, a §4-gated post-freeze amendment entry (Aaron-go, dated) rather than a rewrite of frozen text: pole tokens are sigiled where the co-resident name space can legally contain the bare token. S1's `source` / `where_source` / `fields_source` stay bare — plugin names are extension-mandatory — and that exemption's ground is **pinned**, not asserted: `read-plugin-file-guard` proves an extensionless spelling refuses by name and never fuzzy-matches.

## Review rounds (§11)

**Two rounds, five agents, each in its own worktree, prompts seeded with the settled-decisions list only.** Stopped at two on the convergence rule, escalated, folded on the ruling; no rounds after the directed fold.

**Round 1 — 3 agents.** 3 HIGH, 3 MEDIUM, 2 LOW folded.
- The provider list rendered `ModA (loose)` while the pole matched bare `ModA`, so the refusal's own remedy refused and the next refusal called the caller's verbatim copy a typo. Found independently by all three.
- `source_provider` was **never exercised through either tool entrypoint** — deleting the mapping left the suite green.
- The rename-provenance arm **passed with the thing it tested removed** (the relative path is a substring of the provider's absolute path, so `Contains` held by construction).
- `IsPathRooted` made `\meshes\…` a Data-relative destination and an on-disk source in the same call.
- Quoted sources classified with quotes stripped, consumed with them intact.

**Round 2 — 2 agents.** Nothing folded; escalated instead.
- The round-1 fix (`winnerTokenFree`) shipped **with no arm in either direction** — hard-coded `true` or `false`, all guards stayed green.
- The round-1 round-trip arm **could not see an apostrophe**: single quotes are not a boundary, and `JK's Skyrim` dissolves them mid-name.
- The `.bsa` extension test pre-empted the volume test, so a genuine Data-relative `meshes\thing.bsa` still resolved against the process working directory.
- The both-slots refusal promised `source_provider=` was "fine here" for combinations the service refuses.

**Refused, with reasons** — the refusal rate is the triage signal, so both are named:
1. Arming the `AssetRootHint` / `ReadIncomplete` clauses on the source-absent message. They duplicate an already-armed pattern one lane over, it's a low, and lows here are fix-or-drop rather than a count to grow.
2. Half of the collision-remedy finding: took the in-tool route it was right that I omitted, kept the on-disk fallback rather than accepting the premise that it's useless — a caller with MO2 open can act on it. (Moot now; the sentence is deleted.)

Two defects I caught on myself mid-fold rather than folding blind: a collision remedy that told the caller to pass `source_provider=` the mod's own name — which *is* the collision — and a helper that guessed the patch folder from the naming convention, so the wire arms failed for a reason unrelated to their claim.

## Guards

`ci-all` **117/117** · `write-surface-guard` **191/191** · `place-asset-guard` arms I, K, L · `copy-npc-appearance-guard` arm 4 · `read-plugin-file-guard` arm 7c.

Probe fixtures are named **hostilely on purpose** — `JK's Skyrim` and `winner` — because friendly names cannot exercise what the provider list promises.

**RED-verified, each sabotage confirmed applied before its result was trusted, each restored:** un-sigiling the pole token · single quotes in the name list · extension tested before the qualified path · the pole-spelling tail removed · the both-slots relative-`.bsa` test removed · the rename prefix dropped · `source_provider` dropped at the wire · `IsPathRooted` restored · the carry pole swapped to sole-provider.

The `read-plugin-file-guard` arm carries a same-name-with-extension control, which caught the arm passing for the wrong reason once — everything errored because the corpus hadn't been generated.

## Outside this branch

`nif_inspect` / `nif_set`'s `mod=` has the same print/accept mismatch on `main` — it matches a bare provider name while its not-found message lists the annotated form. Not touched here. `[JsonPropertyName]` pinning for the *sibling* fields (`source`, `asset_path`) is the same pre-existing class; the branch's own new field is pinned in arm L. Both are Issues-lane items awaiting your go — say the word and I'll file them.

### DIFF
--- a/src/housecarl-generator/WriteSurfaceGuardProbe.cs
+++ b/src/housecarl-generator/WriteSurfaceGuardProbe.cs
@@ -2222,9 +2222,11 @@ public static class WriteSurfaceGuardProbe
     }
 
     /// <summary>Drive the REAL place service to each of its source-selection refusals and return the rendered
-    /// outcomes. Its own throwaway MO2 instance rather than the shared fixture: the refusals need an asset path THREE
-    /// mods contend for — one of them named "winner" — and growing the shared fixture to carry that would move a tree
-    /// eleven other arms assert against (#333's lesson) for the sake of four sentences.</summary>
+    /// outcomes. The fixture is TWO mods contending for one asset path, plus one on-disk source and one both-slots
+    /// spec — enough to reach every sentence below. (It briefly carried a third mod named "winner", for a collision
+    /// refusal that no longer exists: the pole token is sigiled, so no provider name can collide with it.) Its own
+    /// throwaway MO2 instance rather than the shared fixture, because growing that tree would move something eleven
+    /// other arms assert against — #333's lesson — for the sake of a few sentences.</summary>
     static List<string> PlaceSourceRefusalRenders(string root)
     {
         const string rel = @"meshes\hcw2\twin.nif";
--- a/src/housecarl-generator/WriteSurfaceGuardProbe.cs
+++ b/src/housecarl-generator/WriteSurfaceGuardProbe.cs
@@ -99,7 +99,7 @@ public static class WriteSurfaceGuardProbe
             // records from W2TwinFwd.esp. Those patches are never ticked into plugins.txt, so they stay out of the
             // active order and cannot move a later arm's winner; an arm added after this one should still know the
             // folders exist.
-            TwinParityArm(fx);
+            TwinParityArm(fx, root);
             // #324's arm also writes into the replacer IN PLACE (its second half is that lane), so it sits with the
             // in-place arms below rather than among the read-the-fixture ones. It rewrites the replacer TWICE and
             // changes two of its records: the TOPIC (W2Topic → "Master Topic") and the CELL (W2Cell → "Master Cell",
@@ -2047,7 +2047,7 @@ public static class WriteSurfaceGuardProbe
     /// on each transport actually reading it. Two further parity assertions ride per outcome: the BUDGET (a cap
     /// that truncates one transport truncates the other) and the COUNTS (the total each lane states is the same
     /// number).</para></summary>
-    static void TwinParityArm(Fixture fx)
+    static void TwinParityArm(Fixture fx, string root)
     {
         Console.WriteLine("── ARM 12: D2 TWIN PARITY — one outcome, both transports, one source per sentence ──");
 
@@ -2199,6 +2199,13 @@ public static class WriteSurfaceGuardProbe
             cap => JsonWire.RenderSeqOutcome(seqStates[2], cap),
             "quest_count", $"{quests.Count} start-game-enabled quests");
 
+        // ---- place_asset's source refusals ------------------------------------------------------------
+        // place_asset renders on ONE transport, so these sentences' claim is "reaches a render", not "reaches
+        // both" — and they are observed off outcomes the REAL service produced. A probe that built the
+        // PlaceResult itself would be handing this arm the very sentence it claims to have found, which is the
+        // circular shape the whole reach check exists to catch.
+        foreach (var render in PlaceSourceRefusalRenders(root)) Observe(render, "{}");
+
         // ---- the coverage assertion — what stops this arm being theatre ------------------------------
         var missingText = twins.Select(t => t.Name).Where(n => !seenText.Contains(n)).ToList();
         var missingJson = twins.Select(t => t.Name).Where(n => !seenJson.Contains(n)).ToList();
@@ -2212,6 +2219,47 @@ public static class WriteSurfaceGuardProbe
         var missingOuter = outer.Select(t => t.Name).Where(n => !seenOuter.Contains(n)).ToList();
         Check("every [MustState] sentence on WriteSentences itself reaches a render (the wiring half a content pin cannot provide)",
             missingOuter.Count == 0, missingOuter.Count == 0 ? null : "unobserved: " + string.Join(", ", missingOuter));
+    }
+
+    /// <summary>Drive the REAL place service to each of its source-selection refusals and return the rendered
+    /// outcomes. Its own throwaway MO2 instance rather than the shared fixture: the refusals need an asset path THREE
+    /// mods contend for — one of them named "winner" — and growing the shared fixture to carry that would move a tree
+    /// eleven other arms assert against (#333's lesson) for the sake of four sentences.</summary>
+    static List<string> PlaceSourceRefusalRenders(string root)
+    {
+        const string rel = @"meshes\hcw2\twin.nif";
+        var inst = Path.Combine(root, "place-sentences");
+        var mods = Path.Combine(inst, "mods");
+        var prof = Path.Combine(inst, "profiles", "Default");
+        foreach (var d in new[] { mods, prof, Path.Combine(inst, "game", "Data") }) Directory.CreateDirectory(d);
+        File.WriteAllText(Path.Combine(inst, "ModOrganizer.ini"),
+            "[General]\r\ngameName=Skyrim Special Edition\r\nselected_profile=@ByteArray(Default)\r\ngamePath=@ByteArray("
+            + Path.Combine(inst, "game").Replace(@"\", @"\\") + ")\r\n");
+
+        // Three providers, one of them carrying the reserved token's name — enough for every refusal below.
+        var providers = new[] { "W2AssetA", "W2AssetB", AssetSourceChoice.WinnerToken };
+        foreach (var m in providers)
+        {
+            var dir = Path.Combine(mods, m, "meshes", "hcw2");
+            Directory.CreateDirectory(dir);
+            File.WriteAllBytes(Path.Combine(dir, "twin.nif"), new byte[] { 1, 2, 3 });
+        }
+        File.WriteAllText(Path.Combine(mods, providers[0], "Dummy.esp"), "x");
+        File.WriteAllText(Path.Combine(prof, "loadorder.txt"), "# header\r\nDummy.esp\r\n");
+        File.WriteAllText(Path.Combine(prof, "plugins.txt"), "*Dummy.esp\r\n");
+        File.WriteAllText(Path.Combine(prof, "modlist.txt"),
+            "# header\r\n" + string.Join("\r\n", providers.Select(p => "+" + p)) + "\r\n");
+        File.WriteAllText(Path.Combine(prof, "Skyrim.ini"), "[Archive]\r\nsResourceArchiveList=\r\n");
+
+        using var svc = LoadOrderService.WithInstance(inst, 0, new UserConfigStore(Path.Combine(root, "place-sentences.user.json")));
+        string Render(PlaceRequest req) => PlaceWire.Render(svc.PlaceAssets(new[] { req }, null, null));
+        return new List<string>
+        {
+            Render(new PlaceRequest(rel, null, null)),                                          // contended, no pole
+            Render(new PlaceRequest(rel, rel, "W2NoSuchMod")),                                  // named, absent
+            Render(new PlaceRequest(rel, rel, AssetSourceChoice.WinnerToken)),                  // token vs the mod
+            Render(new PlaceRequest(rel, Path.Combine(root, "w2-ondisk.nif"), providers[0])),   // pole vs on-disk source
+        };
     }
 
     /// <summary>The budget half of the twin harness: one outcome, one cap, both transports. Asserts (a) a cap tight
--- a/src/housecarl-generator/WriteSurfaceGuardProbe.cs
+++ b/src/housecarl-generator/WriteSurfaceGuardProbe.cs
@@ -2236,8 +2236,8 @@ public static class WriteSurfaceGuardProbe
             "[General]\r\ngameName=Skyrim Special Edition\r\nselected_profile=@ByteArray(Default)\r\ngamePath=@ByteArray("
             + Path.Combine(inst, "game").Replace(@"\", @"\\") + ")\r\n");
 
-        // Three providers, one of them carrying the reserved token's name — enough for every refusal below.
-        var providers = new[] { "W2AssetA", "W2AssetB", AssetSourceChoice.WinnerToken };
+        // Two contending providers — enough for every refusal below now that no name can collide with the pole.
+        var providers = new[] { "W2AssetA", "W2AssetB" };
         foreach (var m in providers)
         {
             var dir = Path.Combine(mods, m, "meshes", "hcw2");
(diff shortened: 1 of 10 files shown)


src/housecarl-mcp/PlaceAssetTools.cs
```diff
@@ -28,10 +28,14 @@ public static class PlaceAssetTools
          "housecarl_asset_status (which reports which copy currently wins). Give the DESTINATION as asset_path (a " +
          "Data-relative path); OR, for an NPC's generated FaceGen file, as formid (the NPC's FormID 'XXXXXX:Plugin.esp') " +
          "+ kind ('mesh' = the head .nif, 'tint' = the face .dds), which houseCARL computes the path for. Give the SOURCE " +
-         "(the copy to place) as source= a loose file path, or '<archive.bsa path>|<entry inside>', or just a '.bsa' path " +
-         "(the entry is taken to be the destination — a quick way to pull ONE file out of a BSA as a loose override). If " +
-         "source is OMITTED, houseCARL auto-resolves: it uses the SOLE provider in your load order, and REFUSES (telling " +
-         "you the candidates) if more than one provides it — it will not guess which is correct. The write is crash-atomic; " +
+         "(the copy to place) as source= a DATA-RELATIVE path resolved through the VFS — with source_provider= naming " +
+         "whose copy (a mod folder / BSA filename as housecarl_asset_status renders it, or 'winner') — or as a full loose " +
```
Avick3110: The tool description still advertises the **pre-sigil** grammar: `or 'winner'` and "a mod folder / BSA filename **as `housecarl_asset_status` renders it**". Both contradict what this branch actually accepts, and the parameter description 20 lines below.

Concrete failure: an MCP client reads this description first and calls `source_provider="winner"`. `AssetSourceChoice.Parse` only matches `*winner`, so it falls through to the **Named** pole for a provider called `winner`. On a load order that contains a mod folder named `winner` — the exact case the sigil was introduced to make safe — the call **succeeds and silently places that mod's bytes** instead of the VFS winner's. On every other load order it is a refusal for a form the tool told the caller to use.

The second clause is wrong in the other direction: `asset_status` renders `SomeMod (loose)`, which the `source_provider` param description explicitly says must **not** be passed.

src/housecarl-mcp/PlaceAssetTools.cs
```diff
@@ -234,6 +249,9 @@ public sealed record PlaceAssetSpec
     [JsonPropertyName("asset_path"), Description("A Data-relative destination path (e.g. 'meshes/actors/...'), instead of formid. Provide this OR formid.")]
     public string? AssetPath { get; init; }
 
-    [JsonPropertyName("source"), Description("The correct copy to place: a loose file path, '<archive.bsa>|<entry>', or a '.bsa' path. Omit to auto-resolve the sole VFS provider. With formid and no kind, an explicit source must be a bare '.bsa' path.")]
+    [JsonPropertyName("source"), Description("The copy to place: a Data-relative path (resolved through the VFS; different from the destination = a rename), a full loose file path, '<archive.bsa>|<entry>', or a '.bsa' path. Omit to resolve the destination path through the VFS. With formid and no kind, an explicit source must be a bare '.bsa' path.")]
```
Avick3110: Stale against this branch's own change: `MapSpec` now requires `Path.IsPathFullyQualified(srcProbe)` for the both-slots expansion, and the refusal was reworded to "must be a **FULL** '.bsa' path" — but this field description still says **bare**.

A client that follows it sends `{formid: "00A1B2:Skyrim.esm", source: "MyMod - Textures.bsa"}` to `housecarl_bulk_place_asset`, `srcOkForBoth` is false, and the **whole call** is refused by a message contradicting the description it just followed.

src/housecarl-mcp/PlaceAssetTools.cs
```diff
@@ -60,10 +66,12 @@ public static string PlaceAsset(
...
+         "{ formid?, kind?, asset_path?, source?, source_provider? }: give EITHER asset_path (any Data-relative path) OR formid (an NPC " +
          "FormID — omit kind to place BOTH the FaceGen mesh and the tint; or set kind='mesh'/'tint' for just one). source " +
-         "is the copy to place (a loose file path, '<archive.bsa>|<entry>', or a '.bsa' path); omit it to auto-resolve the " +
-         "sole VFS provider (an ambiguous or absent source becomes a per-asset error — the rest still place). When you give " +
+         "is the copy to place (a Data-relative path resolved through the VFS — source_provider names whose copy — a full " +
+         "loose file path, '<archive.bsa>|<entry>', or a '.bsa' path); omit it to resolve the destination path instead " +
+         "(an ambiguous or absent source becomes a per-asset error — the rest still place). A per-asset source that " +
+         "differs from its destination is a RENAME. When you give " +
          "a FormID with no kind (placing both files), an explicit source must be a '.bsa' path (each slot's entry is " +
```
Avick3110: Same class as the `source` field description: the both-slots expansion now requires a **fully-qualified** `.bsa` path (`Path.IsPathFullyQualified` in `MapSpec`), but this still says only "a '.bsa' path". A caller passing a relative `.bsa` with formid and no kind loses the whole batch with nothing here to warn them the path had to name a volume.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8445,9 +8504,12 @@ public static NifSetResult OkInPlace(string rel, NifProvider edited, IReadOnlyLi
 
 /// <summary>One asset to PLACE (housecarl_place_asset / bulk). <see cref="AssetPath"/> is the resolved Data-relative
 /// DESTINATION (the tool computes it from a FormID+slot for FaceGen, or takes a raw path). <see cref="Source"/> is the
-/// correct copy to place — a loose file path, "&lt;archive.bsa&gt;|&lt;entry&gt;", or a ".bsa" path (entry := AssetPath);
-/// null/blank ⇒ auto-resolve (use the sole VFS provider; &gt;1 ambiguous and 0 absent are per-asset refusals, Q3).</summary>
-public sealed record PlaceRequest(string AssetPath, string? Source);
+/// copy to place — a Data-relative path resolved through the VFS, a full loose file path,
+/// "&lt;archive.bsa&gt;|&lt;entry&gt;", or a ".bsa" path (entry := AssetPath); null/blank ⇒ the VFS lane pointed at the
+/// destination path. <see cref="SourceProvider"/> picks the pole for a VFS source (a provider name, or "winner");
```
Avick3110: The pole is spelled `*winner` everywhere else on the branch; this doc — the canonical contract for the record — still carries the deleted bare spelling. An in-process caller writing `new PlaceRequest(rel, rel, "winner")` from this summary gets a **Named** read, which is silently indistinguishable from the winner pole right up until a mod folder is actually called `winner`.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1574,28 +1607,27 @@ PlaceResult PlaceOne(PlaceRequest req, AssetResolver.AssetView view, string outD
...
-                    + $"Pass source= one of: {string.Join("; ", res.Sources.Select(SourceHint))}.",
-                    winner);
-            var (b, desc, err) = ReadResolvedSource(res.Sources[0]);
+            var (b, desc, err) = ReadResolvedSource(pick.Source!);
             if (err is not null) return PlaceResult.Fail(rel, err, winner);
-            bytes = b!; sourceDesc = desc!;
+            bytes = b!;
+            // A source read from a DIFFERENT path than the destination is a rename, and the render has to say so —
+            // "placed from ModX" alone would hide that the bytes are another file's.
+            sourceDesc = sourceNamed ? $"{srcRel} — {desc}" : desc!;
```
Avick3110: The rename prefix is added whenever an explicit Data-relative `source=` was given, without comparing `srcRel` to `rel` — so a source that **equals** the destination renders as a rename.

That matters because arm I8 pins the *absence* of this prefix as the "not a rename" signal, and I4 pins its presence as the rename claim. The plain named-provider read `asset_path=meshes\a.nif source=meshes\a.nif source_provider=ModB` — the shape arms I1/I2/I3/I7/I9 themselves use — renders as `meshes\a.nif — 'ModB' (loose)`, i.e. "one file's bytes landed under another file's name", which is not what happened.

```suggestion
            sourceDesc = sourceNamed && !string.Equals(srcRel, rel, StringComparison.OrdinalIgnoreCase) ? $"{srcRel} — {desc}" : desc!;
```

src/housecarl-generator/WriteSurfaceGuardProbe.cs
```diff
@@ -2214,6 +2221,55 @@ void Observe(string text, string json)
             missingOuter.Count == 0, missingOuter.Count == 0 ? null : "unobserved: " + string.Join(", ", missingOuter));
     }
 
+    /// <summary>Drive the REAL place service to each of its source-selection refusals and return the rendered
+    /// outcomes. Its own throwaway MO2 instance rather than the shared fixture: the refusals need an asset path THREE
```
Avick3110: The summary describes a fixture the body doesn't build: it says the refusals need "an asset path **THREE** mods contend for — one of them **named `winner`**", while the code below creates exactly two providers, `W2AssetA` and `W2AssetB` (correctly — the sigil removed the need for the `winner` fixture). Worth updating so a later session extending this arm doesn't assert against a provider that isn't there.

Avick3110:
All six folded in `dd38cec`. Per finding:

**1 — token spelling, closed by construction.** Both `Description` attributes now build from `AssetSourceChoice.WinnerToken` by const concatenation, so the wire spelling and the parse site that matches it cannot drift apart — a future respelling changes one const and both descriptions follow. The token already lived beside `Parse`; this makes the descriptions read it rather than repeat it. Its two other defects went at the same time: the bare `'winner'` is gone, and so is **"as `housecarl_asset_status` renders it"** — you're right that it contradicted the parameter description one line down. `asset_status` renders `Name (kind)`; the parameter takes the name alone. It now says to pass the name without the annotation, and the parameter description says the same thing the same way.

**2 + 3 — both bulk descriptions.** The bulk tool description and `PlaceAssetSpec.Source` both now say a both-slots `source=` must be a **fully-qualified** `.bsa` path, with the reason stated (a relative `.bsa` is a Data-relative asset path, and one path cannot serve two slots). That matches `MapSpec`'s `Path.IsPathFullyQualified` check and the refusal's own wording. `MapSpec`'s XML summary carried the same stale "bare '.bsa'" and is fixed too.

**4 — `PlaceRequest` XML doc.** Rewritten to the sigiled grammar: a provider name on its own, or the sigiled token, with `<see cref="AssetSourceChoice.WinnerToken"/>` naming the const rather than spelling it. Also corrected to "fully-qualified" for the loose and `.bsa` forms, and it now states that the same path is *not* a rename.

**5 — rename prefix suppressed when source equals destination.** It keyed on whether a source was *named*, so `place meshes\x.nif from ModB` — the natural spelling of a provider-scoped placement — rendered as a rename and told the caller a file had moved when none had. Now keyed on the paths differing, compared through `SameAssetPath`, which is the VFS's own key: both paths are already through `ValidateRelPath` (separators and any leading separator folded), so the remaining axis is case, and the asset layer's lookups are ordinal-case-insensitive. A raw compare would report a rename between one file and itself.

New arm covers plain, upper-case, and forward-slash spellings of the destination. **RED-verified both ways**, each sabotage confirmed applied before its result was trusted and each restored:
- `SameAssetPath` → `StringComparison.Ordinal`: **1 fail**, the case variant only — the separator variant still passes because `ValidateRelPath` folds separators, which is exactly the claim the case-fold half makes.
- suppression removed (back to keying on `sourceNamed`): **3 fails**, all spellings.

I4's true-rename assertion is unchanged and still passing, so the suppression didn't buy itself a pass by weakening the arm next door.

**6 — `PlaceSourceRefusalRenders` summary.** Rewritten to describe the fixture the code actually builds: two contending providers, one on-disk source, one both-slots spec. The third mod named `winner` and the collision refusal it fed are both gone — the sigil removed the need for them.

**The sweep you asked for, before the commit.** Grepped the branch diff's touched surfaces for surviving pre-sigil and pre-rooted-ness wording. Two more found and fixed in the same commit:
- a probe comment still quoting the old *"must be a bare '.bsa' path"* refusal;
- `PlaceOne`'s shape comment still describing an on-disk source as *"a rooted path"*, which is precisely the distinction that fold changed.

Everything else the sweep surfaced is correct as it stands, and I'd rather name it than leave it looking unswept: S1's bare `where_source=winner` is the SPEC §5.5 exemption (plugin names are extension-mandatory), the older `CHANGELOG` entries are history and shouldn't be rewritten, and the comments that do mention the bare token mention it to explain why it changed.

`ci-all` 117/117. Full diff read before staging; staged by path.

PR #346, opened 2026-08-15
copy: the closure-copy primitive, and copy_npc_appearance as skill data

Avick3110:
`copy_npc_appearance` was a job-shaped verb: one tool that knew what a head part is, which four fields make up a face, that a race must not be walked into, and how to rename a FaceGen pair. This replaces it with a generic closure-copy primitive plus a skill that hands it the appearance knowledge as data — the composition cornerstone applied to the verb that most obviously violated it.

No linked issue — this lane charters from plan docs (same as #337, #339).

## What ships

**`housecarl_copy`** — copy a record together with its declared link closure into a patch under new FormIDs. What gets copied is decided by a rule (internalize a link defined in the plugin(s) you are copying away from, or one that does not resolve in the active order; keep everything else) rather than by a list. Where the walk starts is the caller's (`seed_paths=`), and which record classes it must not enter is the caller's (`exclude_types=`, two severities).

**The ordered source universe** (SPEC §3.1 amendment) — `from_source=` is a list of §4.2 poles tried first-hit-wins, and the readback names which source produced each record. A length-1 list is exactly today's grammar, structurally: there is no separate single-pole code path.

**`npc-appearance-copy`** — the skill. Seed set, Race exclusion, the inline bundle that rides `housecarl_apply`'s copy zip, the EditorID-preservation reason, and the FaceGen carry from the *named donor*.

## Acceptance

The behavior inventory (`dev/projects/tool-surface-2.0/PR3B_BEHAVIOR_INVENTORY_2026-08-14.md`) dispositions all 106 ancestor behaviors. `copy-differential` runs the comparison over constructed MO2 instances, driving both sides through the tool layer and comparing the artifacts on disk.

| Fixture | Result |
|---|---|
| 1 — donor-defined NPC, active mod | clone **117 projected lines identical**, attach **118 identical** |
| 2 — disabled donor via the ordered source list | **117 identical**, records only (R1) |
| 3 — contested vanilla FaceGen path | **the recorded divergence: ancestor 37B (VFS winner), successor 1B (named donor)** |

Fixture 3's contest is constructed *and asserted* in the harness — two providers, asserted to hold different bytes, loose asserted to out-sort the BSA — which is what makes its stop rule decidable. Both halves of that rule STOP the run rather than report a pass.

**Aaron's live-install pass is the empirical half of fixture 3 and is not run here.**

**Stated gap:** fixture 2's asset half is not compared (R1 rules the disabled-donor asset lane out of this branch). It belongs to the F1 PR, which depends on `copy_npc_appearance` surviving until then — the removal point is Aaron's.

## Review rounds — three, the cap

**Round 1** (3 fresh reviewers, isolated worktrees, §12-seeded): **26 findings, 5 high.** All three converged on one class — the generic surface promised any link-bearing field path, the implementation covered two shapes, and everything outside them failed silently. **Not folded; escalated under §4**, because which way it resolved was a scope call. Refused 1/26 (a latent pass-1/pass-2 nullability divergence with no constructible failure).

**Fold 1** (Aaron-go, shape ruling (a)): unsupported shapes refuse by name with the route; unset/empty assigns; kept links split into scope-vs-exclusion; R2's from-arm rendered; the method-form sentences moved to consts so `[MustState]` reaches them; inventory I1 built; duplicate `exclude_types` and blank list elements refused by name/index. The Stop-exclusion arm was a disjunction whose first half was unconditionally true — `continue → break` left the walk internalizing nothing with all four arms passing; rewritten positively. **All six deletions the test reviewer had measured as green now go RED**, each verified with the source restored byte-identical.

**Round 2**: **23 findings.** Round 1's class recurred, found independently by two of three reviewers at four sites. **§11's class-stop fired — not folded, escalated again.** Plus two new highs (an active named source never bound, so `standalone` could be false; `Type:stop` unserializable against an off-order source).

**Fold 2** (Aaron-go on all three): `ClosureWalk.ClassifySeed` is now the single shape judgement, reading the **declared** type — Mutagen declares these list properties non-nullable while the overlay returns `null`, so only the model can answer what a field is. Every site consumes it. Every named `from_source=` arm binds (`winner` exempt). `Type:stop` off-order refuses up front with both remedies. Cleanup is `finally`-shaped and a nested-group target is a typed refusal. Two of my own errors reverted: the fixture-3 stop-rule over-claim, and a round-1 fold that round 2 instrumented at zero hits — dead code, a fold without a test.

**Round 3** (the cap): findings below, **not folded** — there is no round 4 to check a fold, which is the pattern this cap exists to stop.

Refusal rate across the branch: 1 of 26 in round 1; rounds 2 and 3 were escalated/listed rather than folded, so triage was per-ruling. Every finding folded in rounds 1 and 2 was reproduced in the source first, and every sabotage was restored byte-identical with a residue grep.

**What the three rounds cost and bought, plainly:** 9 reviewers, ~70 findings, two §4 escalations. Four defects were introduced *by the folds themselves* and caught by the next round — which is the argument for the cap, and the reason round 3 is listed rather than folded.

## Round 3's open findings — for your review

**HIGH, and two reviewers found it independently.** A donor defined in a **base-game master** produces:

```
CLONED 01A51A:Dawnguard.esm -> 000800:F3New.esp
masters: Dawnguard.esm
standalone: the source is NOT a master of this patch.
```

`bound` excludes base masters by design (correct — copying a vanilla record must not internalize vanilla), and `SourceAmongMasters` is computed over that same set, so the claim is asserted from a set deliberately emptied. **The ancestor had a third render arm for exactly this** (`NpcCopyTools.cs:85` — *"an appearance transplant, not a standalone-ization"*), which the inventory dispositions as **F24 / P**. It was never built — the same shape as I1 before fold 1 caught it. It lands on the skill's own headline eval query ("a standalone follower … I don't want it to need Dawnguard's records"), and `SKILL.md` tells the caller to treat that line as their verification.

**This also names a structural limit of my differential:** `Project()` compares plugin bytes and masters, never the two tools' **response text**. A render-level divergence is invisible to it by construction.

Other round-3 findings, all measured, none folded:

- **MEDIUM** — `Type:stop` in the *clone* lane: the strip removes every bound link including the one `stop` asked to keep, so the response simultaneously claims the link was kept, that the patch is standalone, and that the link was stripped.
- **MEDIUM** — the off-order `Type:stop` refusal (fold 2) is over-broad: in the clone lane the link is always stripped, so no master is needed and a copy that would have written cleanly is refused.
- **MEDIUM** — the strip reports "1 reference(s)" when a whole property was nulled; for `VirtualMachineAdapter` that is every script on the clone, and only the offending link is named.
- **MEDIUM** — the source-miss remedy names a plugin to add, but the description says the typical cause is a dangling link, for which no such plugin exists; the actual remedies (`stop`/`refuse`) go unmentioned.
- **MEDIUM-LOW** — `AttachSeedFields` reuses `UnsupportedSeedShape` for three non-shape failures, so the render appends the shape route to refusals that are not about shape.
- **MEDIUM-LOW** — the `NoSeeds` refusal does not name the templated-donor cause that R4 attributes to it.
- **LOW** — a walk that reaches the `from` record itself internalizes it twice, leaving a stray duplicate; overlay disposal in `BuildSourceChain` is not exception-safe; `SourceArm.Spelling` is not verbatim for a path naming an active plugin; two remedy sentences point at supersets the shape gate refuses.

### And three findings against my own tests — including one I had already fixed once

- **HIGH — R5's arm is not load-bearing, measured.** Arm 7c asserts that naming an enabled plugin binds it, but the fixture's named source is `Shadow.esp` and `from` *lives in* `Shadow.esp`, so `bound.Add(sourceKey.ModKey)` binds it regardless of the R5 loop. Deleting that loop leaves `ci-all` **ALL PASS**. So R5 — a ruling made two hours ago — ships effectively unpinned, and the guard table below would be overstating it if I let it stand. **This is the same blind spot as cut 1** ("the single file arm's ModKey already equals `from`'s"), which I found, documented in a comment, and then walked into again in the very arm written to close it. The fix shape is known: seed from `Src.esp` with `from_source=['Src.esp','Shadow.esp']` and a record defined by Shadow.
- **MEDIUM** — R3's assign half is pinned to disk for the *link* shape but only to the outcome object for the *list* shape; making an empty source list report `cleared` while leaving the target's keywords intact stays green.
- **MEDIUM-LOW** — a comment I wrote in `WriteSurfaceGuardProbe` claims both new refusals are reachable end to end. True of `StopOffOrder` (proven RED), false of the nested-target one, which is only covered by a constructed outcome.
- **LOW** — a vacuous clause in the node-cap arm (`Key != default` on a `FormKey?` only re-asserts that a refusal exists); `ClosureWalkProbe`/`ClosureCopyProbe` still lack the top-level `try`/`catch` the branch's own guard-note argues for; `DiffPatches` compares set-wise, so duplicate projected lines would not show (reasoned, not measured).

## Guards

| Guard | Arms | Covers |
|---|---|---|
| `source-chain-guard` | 39 | order is the semantics · per-hit arm naming · a miss names every arm · fault ≠ miss · the no-sigil exemption |
| `closure-walk-guard` | 37+ | generic expansion · seed shape as declared type · exclusions at both severities · caps with the full chain · cycles vs diamonds |
| `closure-copy-guard` | 33 | fresh keys · scratch-mod remap scoping · nullability by model not method · required-link refusal · leak scoping |
| `copy-service-guard` | 40+ | the wired operation with the source DISABLED · two-arm chains with distinct ModKeys · the missing-master disjunct · E4 · both `Type:stop` lanes · R3's assign half (link shape only — see above) · I1 |
| `copy-parser-guard` | 30 | the wire: exclusion severities told apart by RESULT, destinations, FormID refusals and their order, `seed_paths` required and trimmed, the `from_source` default |

`write-surface-guard` 191/191. **ci-all 122/122.**

## The skill, audited against `standards/HOUSECARL_SKILL_AUTHORING.md`

19/19 on §8's reviewer-binding checklist. Description 346 chars (200–400 band), body 158 lines, trigger fan-out **recall 9/10, specificity 10/10** — both §6.3 gates passed, the one miss recorded rather than smoothed. **§6.6 is a stated partial**, not a claimed pass: it applies to procedural-write skills and asks for ≥8 outcome cases; what exists is the differential's 3 fixtures. The standard's own threshold for §6.6 is marked *Aaron-decision pending*.

Binding sub-rule adopted from this branch's own failure: **every skill step that names a tool's output gets RUN before it ships.** Three claims in the first draft were text nobody had executed and two were false — including a readback section the tool did not emit. The differential now prints the successor's Step-1 readback verbatim on every run.

## Separate commit, your go

`b6e0091` updates **CLAUDE.md §7** (13 → 14 skills). It is the operating manual, so §5 #10 puts it behind an explicit go rather than the conductor's judgement — it is split out so you can take or drop it independently.


Avick3110:
## Round 3's findings, folded — `7379b48`

**Your independent review and live-install pass are not on this PR** (all three comment streams are empty as of this fold), so the triage set here is round 3's open list alone. Nothing below pre-empts your review; the landing still waits on your go.

`ci-all` **122/122**. `copy-differential` **clean**, fixture 3 still diverging (ancestor 37B / successor 1B).

### Triage — 13 findings in, 11 folded, 2 refused (15%)

| # | Finding | Disposition |
|---|---|---|
| 1 | **HIGH** — base-master donor gets a false `standalone` claim (F24) | **Folded**, RED-checked |
| 2 | **HIGH** — R5's guard arm cannot fail | **Folded**, RED-checked |
| 3 | **MED** — `Type:stop` in the clone lane claims kept + standalone + stripped | **Folded**, RED-checked |
| 4 | **MED** — the off-order `Type:stop` refusal is over-broad | **Folded**, RED-checked |
| 5 | **MED** — strip reports "1 reference(s)" when a whole property was nulled | **Folded** |
| 6 | **MED** — source-miss remedy names no route for the dangling-link cause | **Folded** |
| 7 | **MED-LOW** — `AttachSeedFields` reuses `UnsupportedSeedShape` for three non-shape failures | **Folded** |
| 8 | **MED-LOW** — `NoSeeds` does not name the templated-donor cause | **Folded** |
| 9 | **LOW** — a walk reaching `from` internalizes it twice | **Folded** |
| 10 | **LOW** — `BuildSourceChain`'s overlay disposal is not exception-safe | **Folded** |
| 11 | **LOW** — `SourceArm.Spelling` not verbatim for a path naming an active plugin | **Refused** — see below |
| 12 | **LOW** — two remedy sentences point at supersets the shape gate refuses | **Refused** — see below |
| 13 | My own test gaps (5 items) | **Folded**, all five |

### What I refused, and why

**#11 — `SourceArm.Spelling` not verbatim for an active-plugin path.** I could not reproduce a caller-visible defect. Arm 6 already measures that a full path naming an active plugin resolves as the `ActiveOrder` arm and is not described as off-order; the spelling that reaches the readback is the one the caller passed. The finding describes an internal normalization, not a sentence that misleads. Folding it would have been changing code to satisfy a claim I could not make fail.

**#12 — remedy sentences pointing at supersets.** Reading the two sentences against `ClassifySeed`, the routes they name (`housecarl_apply`'s `bundle=`/`assignments=` zip) genuinely do accept the shapes the gate refuses — that is the whole point of R3's ruling (a). The "superset" is the route being wider than the refusal, which is correct for a route. Refused as a misreading of the ruling.

The refusal rate is 2 of 13. Both refusals were reproduced against source first and both are stated with what I checked, not just asserted.

### The two HIGHs

**F24.** `LoadOrderService` computed `baseMasters.Contains(sourceKey.ModKey)` and threw it away; it now keeps it, and `CopyTools.Render` has three arms instead of two — the self-mastered **alarm still goes first and unconditionally** (a bound plugin among the masters is worth shouting about whatever the source was), then the base-game note, then the standalone claim. The sentence is a `[MustState]` const, rendered by `WriteSurfaceGuardProbe` and driven end to end by a new `copy-service-guard` arm using a real `Dawnguard.esm` — Mutagen's actual implicit set, not a stand-in name, so the branch under test is the shipping one. Deleting the render arm takes both arms RED.

**R5.** The fixture now seeds `from` from `Src.esp` while the arm under test names `Shadow.esp`, reaching a head part `Shadow.esp` defines — and `Shadow.esp` is active, so the scope predicate's second disjunct cannot expand it either. Being named is the only route in. **Deleting the R5 loop now takes four arms RED** (measured: `ShadowBrow` not internalized, `Shadow.esp` among the masters); it left `ci-all` ALL PASS before.

Note the honest limit: `SourceAmongMasters` still passes under that sabotage, because with the loop gone `Shadow.esp` is not in `bound` and the claim is computed over the emptied set. The masters-list arm is what catches it — which is the same shape as F24, and worth seeing twice.

### `Type:stop`, scoped per lane

Findings 3 and 4 are one decision, and I took it as directed: **the refusal follows the link that survives into the artifact.**

- **Attach lane** — nothing strips, every pruned link survives (on the target unmapped, or on an internalized record). The up-front refusal is unchanged and arm 7a still drives it.
- **Clone lane** — the strip runs, so the check moved *after* it and asks the artifact directly: is there a link here naming a plugin the game does not load? That is the real serialization question, and unlike the walk-level proxy it can tell a boundary the strip removed from one it never touched — a boundary reached from an *internalized* record is not on the clone, so the strip never sees it, and that one still refuses.

The same measurement fixes finding 3: only boundaries the artifact still carries are reported as kept, so no response claims kept + stripped + standalone at once. All three behaviours were sabotaged independently and each has its own RED.

### The structural blind spot, closed

Your PR body called it out and so did I: `Project()` compares plugin bytes and the masters list, never the two tools' response text, which is exactly how both base-master findings survived three rounds while every fixture reported NO DIFF. Fixture 3's donor lives in `Dawnguard.esm` — it *is* the F24 case — so the differential now compares the two renders directly and asserts both call it a transplant.

### Contract prose swept

Every changed behaviour: the tool `Description` (the `Type:stop` claim was unconditional and is now lane-split; the strip and base-master sentences added), `SKILL.md` (the disabled-donor `Race:stop` bullet, and Verification step 1 — which told the caller to treat the standalone line as their check, and would have had them read a correct transplant note as a failure), the `CHANGELOG` `## Unreleased` entry, and the XML docs on every type touched.

### Still open, unchanged by this fold

- **Your live-install pass is fixture 3's second half and has not run.**
- Fixture 2's asset half is still uncompared (R1), and `copy_npc_appearance` survives this merge — F1's deferred comparison depends on it.
- `b6e0091` (CLAUDE.md §7, 13 → 14 skills) is still split out for you to take or drop.

**No new review rounds were run** — the cap is spent and §11 forbids them after a directed fold.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4872,6 +4872,306 @@ static void RespellActiveCopySourcePaths(LoadOrderResolver.IndexView view, IList
...
+                // copied away from), never an implicit base master — copying a vanilla-defined record must not
+                // classify vanilla as "the source" and wholesale-internalize it.
+                var baseMasters = Mutagen.Bethesda.Plugins.Implicits.Get(Mutagen.Bethesda.GameRelease.SkyrimSE).BaseMasters;
+                var bound = new HashSet<ModKey>();
+                // Kept as a FACT rather than discarded with the branch: the record's own plugin being a base master
+                // is what makes "standalone" the wrong frame for this copy, and the render needs to say so. It used
+                // to be tested here and thrown away, so the response asserted the standalone claim from a `bound`
+                // set base masters are deliberately excluded from — an assertion off a deliberately emptied set
+                // (inventory F24, review round 3).
+                var sourceIsBaseGame = baseMasters.Contains(sourceKey.ModKey);
```
Avick3110: **HIGH — the base-master render arm is keyed on `from`'s defining plugin, not on whether anything is bound.**

`sourceIsBaseGame = baseMasters.Contains(sourceKey.ModKey)` answers "where is this FormID defined", but the sentence it selects (`CopySourceBaseGame`, chosen over `CopyStandalone` at `CopyTools.cs:194`) claims *nothing is being copied away from*. With an ordered `from_source=` those are no longer the same question.

Concrete case — and it is the skill's own documented flow (`SKILL.md`, §"Reading the donor from an override"): a vanilla NPC overhauled by a mod.

```
from        = '0013BBF:Skyrim.esm'
from_source = ['TheOverhaul.esp', 'Skyrim.esm']
new_editorid= 'MyFollower'
```

`sourceKey.ModKey` is `Skyrim.esm` → `sourceIsBaseGame` **true**. But `bound = {TheOverhaul.esp}` is non-empty (the R5 loop just below binds it), so the walk internalizes TheOverhaul's head parts and `StripBoundLinks` strips its factions/outfits off the clone. The response then reads:

```
note: the source is defined in a base-game master (always loaded) — nothing is being "removed", so links
to it are kept and mastered normally; this copy is an appearance transplant, not a standalone-ization.
...
STRIPPED 3 reference(s) that still pointed into the source:
```

— "links to it are kept and mastered normally" printed above the strip list that removed them, and the standalone claim the copy actually earned never made.

`CopyServiceProbe` arm 7g only exercises `from_source=['Dawnguard.esm']`, where `bound` is empty, so the mixed case is unpinned. The condition the render wants is "is anything bound", which is exactly `bound.Count == 0`.

src/housecarl-core/ClosureCopy.cs
```diff
@@ -0,0 +1,739 @@
...
+        }
+
+        // The CLONE lane's half of the off-order check, asked of the ARTIFACT now that the strip has run. This is
+        // the direct question — "is there a link here naming a plugin the game does not load, which the writer
+        // therefore cannot master" — rather than the walk-level proxy, which could not tell a boundary the strip
+        // removed from one it never touched (a boundary reached from an INTERNALIZED record is not on the clone,
+        // so the strip never sees it, and that one does still have to be mastered).
+        // (FormKey is a struct, so FirstOrDefault yields FormKey.Null when nothing matches — and CollectLinks never
+        // stores a null key, so the IsNull test is exactly "no match", not a second condition on a real one.)
+        var liveLinks = cloneLane ? CollectLinks(patch) : null;
```
Avick3110: **MEDIUM — the clone lane's off-order check measures the whole patch, and misattributes what it finds.**

`CollectLinks(patch)` walks *every* record in the mod, so in `into=` extend mode it also sees records a previous call put there. A link on one of those trips the check, and the refusal is rendered as `CopyStopOffOrderRoute`:

> the walk pruned a record in 'X'. That plugin is not in your active load order, and 'Type:stop' KEEPS the link … Either use 'Type:refuse' for that type

Scenario: a caller extends `MyFollower.esp` with a second clone (`into='MyFollower.esp'`, `new_editorid='Second'`) after disabling a mod the patch's earlier records reference. The copy refuses, and the remedy names an `exclude_types` argument the caller never passed for a prune the walk never performed — the same "a true sentence about the wrong cause" class the exclusion-leak marker was added to fix a few lines up.

Two shapes of fix, either works: scope the measurement to what this call added (the clone plus the bodies behind `copy.Copied`), or keep the artifact-level question but only render it as `StopOffOrder` when the offending key is one `walk.Kept` marks `Excluded`, with a distinct kind otherwise.

src/housecarl-mcp/CopyTools.cs
```diff
@@ -0,0 +1,286 @@
...
+                       "copied. Remove the empty entry rather than leaving it for houseCARL to guess at.";
+        var seeds = rawSeeds.Select(s => s.Trim()).ToList();
+        if (seeds.Count == 0)
+            return "error: seed_paths is required — name the link-bearing field(s) the walk starts from " +
+                   "(e.g. seed_paths=['HeadParts','WornArmor']). Without them there is nothing to copy.";
+
+        var exclusions = new List<WalkExclusion>();
+        foreach (var raw in exclude_types ?? Array.Empty<string>())
+        {
+            if (string.IsNullOrWhiteSpace(raw)) continue;
```
Avick3110: **LOW — blank `exclude_types` elements are silently dropped, against this block's own stated rule.**

`seed_paths[i]` (lines 113–116) and `from_source[i]` (lines 144–147) both refuse a blank element by index, with a comment arguing that dropping one "silently changed the caller's list". This loop does exactly that.

`exclude_types=['  ']` — a templating slip that yields one empty entry — runs with **no exclusions at all**: `Race:refuse` never fires, the walk enters the Race and pulls the skeleton and sibling races, and the caller gets either a node-cap refusal or a copied subtree they believed they had excluded. Nothing in the response says an entry was dropped.

A blank here has no index to shift, so it can just refuse by content rather than by index — but silently continuing is the one thing the neighbouring loops rule out.

src/housecarl-core/ClosureWalk.cs
```diff
@@ -0,0 +1,405 @@
...
+            var label = $"{typeName} {key} ({hit.Body.EditorID ?? "<no editorid>"})";
+            if (hit.Body is not IFormLinkContainerGetter flc) continue;
+            foreach (var link in flc.EnumerateFormLinks())
+            {
+                var target = link.FormKey;
+                if (target.IsNull) continue;
+                // A repeat link is either a CYCLE (the target is on this node's own pull chain) or an ordinary
+                // re-convergence (a diamond). Only the first is a fact worth reporting; conflating them would
+                // bury real cycles under every shared texture set.
+                if (seen.Contains(target))
```
Avick3110: **LOW — a mutual reference between two siblings is classified as a diamond, so a genuine cycle goes unreported.**

Ancestry is measured against the BFS *tree* (`parent`), which is not the same as "reaches itself in the graph".

Trace: seed A links to X and Y; X links to Y, and Y links back to X.

1. A dequeued → `parent[X] = parent[Y] = A`, both enqueued.
2. X dequeued (`seen = {A, X}`). Its link to Y: Y is queued but not yet dequeued, so `seen.Contains(Y)` is **false** — `parent[Y]` is not overwritten and Y is re-enqueued.
3. Y dequeued (`seen = {A, X, Y}`). Its link to X: `seen.Contains(X)` holds, but `IsAncestor(X, Y)` walks `Y -> A` and never reaches X.

So the X↔Y cycle takes the silent `continue` and never appears in `WalkResult.Cycles`. The header comment's contract — "a record that reaches itself is a fact about the data worth reporting" — holds for chains through the tree and not for mutual references between siblings. Report-only, no effect on the artifact.

src/housecarl-core/ClosureCopy.cs
```diff
@@ -0,0 +1,739 @@
...
+            masters, sourceAmong, sourceIsBaseGame, bytes, warning);
+    }
+
+    /// <summary>Every distinct link key the built patch carries. Used to ask the ARTIFACT what survived rather than
+    /// inferring it from the walk — the two disagree in the clone lane, where the strip runs between them.</summary>
+    static HashSet<FormKey> CollectLinks(SkyrimMod patch)
+    {
+        var live = new HashSet<FormKey>();
+        foreach (var rec in patch.EnumerateMajorRecords())
+            foreach (var l in ((IFormLinkContainerGetter)rec).EnumerateFormLinks())
```
Avick3110: **LOW — hard cast here and in `FindBoundLeak` (line 701) contradicts the walk's own defensive check.**

`ClosureWalk.Run:380` treats this exact cast as fallible:

```csharp
if (hit.Body is not IFormLinkContainerGetter flc) continue;
```

while `CollectLinks` and `FindBoundLeak` assert it. Both cannot be right. If the walk's guard is warranted, a record that fails it makes `CollectLinks` throw `InvalidCastException` out of `BuildAndWrite` after the internalize — which `Guard.Tool` renders as "an internal houseCARL failure … not bad input", the same opaque dead end this branch already closed twice this cycle (duplicate `exclude_types`, nested target shapes).

Either drop the check in `ClosureWalk.Run` as dead, or make these two skip the same way — but the two files should not disagree about whether the cast can fail.

Avick3110:
## Your review, folded — `b7302ee`

**5 findings in, 5 folded, 0 refused.** Each cites a mechanism I reproduced in the source before touching it; none was a misreading to push back on. Stating that explicitly because a 100% fold rate is the shape that should draw suspicion — what follows is what I checked in each case.

`ci-all` **122/122**. `copy-differential` **clean**, fixture 3 still diverging.

### 1. HIGH — the base-master arm was keyed on the wrong question

You're right, and it's a regression against F24's own text. The inventory says *"`DonorIsBaseGame` when the donor-bound set is **empty**"* (row F24) — I implemented "where is `from` defined", which is only the same question when `from_source=` is degenerate.

Reproduced your case exactly. Now keyed on `bound.Count == 0`, computed **after** the R5 loop. That can only hold when the source is base-game anyway (a non-base source adds its own ModKey), so the sentence stays true where it fires and the mixed case falls through to the standalone/alarm path it earned.

Your case is now `copy-service-guard` arm **7h**: `Shadow.esp` overrides the `Dawnguard.esm` NPC and points its appearance at a Shadow-defined head part, so `from` is a base-game FormID while `bound = {Shadow.esp}`. Asserts the bound set is non-empty, the override's head part internalizes, the response does **not** say transplant, and it does take the standalone/alarm path. The transplant case stays as arm 7g.

RED both ways: restoring the old keying takes 7h's three arms red; removing the render arm takes 7g's two red.

### 2. MEDIUM — off-order check keeps its reach, loses its misattribution

Kept the artifact-level measurement per your steer. Split the render by cause — and it needed **three**, not two, which is a deviation from your directive worth flagging rather than burying:

| Cause | Refusal | Remedy it names |
|---|---|---|
| key is a boundary an exclusion pruned | `StopOffOrder` (unchanged) | `Type:refuse`, or enable the mod |
| carrier is a record **this call copied** | `CopiedOffOrderLink` (new) | name the field in `seed_paths=`, or enable the mod |
| carrier was **already in the patch** | `PatchOffOrderLink` (new) | enable the mod, or write to a new patch |

The third case is the one your directive described. The middle one is reachable and would otherwise have been told "this call did not create it" about a record it had just created: `wideNpc.HeadTexture` points into off-order `Ghost.esp`, `HeadTexture` isn't seeded, so the walk never reaches it and the whole-record duplicate carries it across. Two kinds would have made a false innocence claim there, so I built three. Say the word if you'd rather they collapse.

Arm 7i drives `CopiedOffOrderLink` end to end and asserts the remedy names `seed_paths=` and **not** `'Type:refuse'`. Collapsing the kinds back takes it red.

**`PatchOffOrderLink`'s wire is not driven, and I'm not claiming it is.** `into=` only accepts a patch houseCARL wrote, and a first call that left an off-order link in one would itself be refused by this very check — so the case isn't constructible end to end. Its sentence is pinned at the render level in `write-surface-guard`, and both the probe comment and this note record that the wire is untested. Your "scope the measurement to what this call added" alternative would have made it unreachable by construction instead; I took the other branch because the check catching pre-existing breakage is worth keeping.

### 3. LOW — blank `exclude_types`

Folded. Refuses by content, since a blank has no index to shift. Worth noting the arm that existed asserted the *opposite* — "blank exclusion entries are dropped" — so the bug was pinned in place by a test. Inverted, plus a second arm for the case that actually bites: a blank **beside** a real exclusion, which silently applied one fewer than the caller passed.

### 4. LOW — sibling mutual references

Your trace is right and I walked it through the code. `parent` is the BFS tree; `IsAncestor` asks a tree question about a graph fact.

Replaced with a post-walk DFS over a recorded edge set — three-colour, an edge into an on-stack node is a back edge, and the stack slice is the cycle. The node cap bounds it. Your X↔Y trace is the fixture (arm 3b). Both directions: the new arm reports the sibling cycle, and the pre-existing arm still asserts **exactly one** cycle on a fixture carrying two diamonds, so the pass didn't buy recall with false positives.

`WalkCycle.Path` now means the loop itself rather than the chain from the seed; its doc comment and the file header both say so.

### 5. LOW — the cast

Folded in both `ScanPatchLinks` (the renamed `CollectLinks`) and `FindBoundLeak`. Took your first option — make them skip like the walk — rather than deleting the walk's guard, since a record with no link container trivially has no links to leak. No arm: it's a robustness alignment with no constructible fixture in a synthetic instance, which is why it's listed here rather than claimed as pinned.

### Prose swept

The tool `Description` (the transplant sentence now says *nothing being copied away from*, not *base-game master*; the `Type:stop` paragraph names all three off-order causes; `exclude_types` says blanks are refused), `SKILL.md` Verification step 1 — which needed the most care, because it now tells you the overhaul flow **should** show standalone and that a transplant note there means something bound went missing from the report — and the `CHANGELOG` entry.

### The differential's fixture-3 assertion

You flagged this and it's the right thing to have flagged. It did **not** need updating, and I checked rather than assuming: fixture 3 passes no `from_source=`, so the universe is `['winner']`, and `winner` is exempt from binding — the whole load order isn't a plugin to copy away from. Bound is empty, so it is still genuinely the transplant case under the corrected condition. The assertion stands unweakened, and the harness now carries a comment saying why, pointing at arm 7h for the case that does differ.

### Standing

- **Your live-install pass is still the open gate** — fixture 3's empirical half.
- Not merging on this directive; holding for your explicit go.
- `b6e0091` (CLAUDE.md §7, 13 → 14) still yours to take or drop.

Avick3110:
## Empirical half deferred — merging

Aaron's go. The live-install pass that is fixture 3's second half is **deferred to the chartered pre-release empirical wave** (advisor-ruled), tracked as **#347**.

**The merge is not the lock — the release gate is.** Nothing about this branch landing on `main` asserts that evidence exists. #347 states what the test is (contested vanilla NPC on the real order; carried FaceGen bytes must be the named donor's, not the VFS winner's; the transplant/standalone claim checked against the actual masters list), why it was deferred (the dev build is kept off the production install), and cross-references the **F1 deferred asset differential** as the same class of debt against this same PR.

Both need `copy_npc_appearance` alive to run, which is why it survives on `main` by ruling rather than being deleted here.

`b6e0091` (CLAUDE.md §7, 13 → 14 skills) is **taken** and merges with the branch.

Landing now: `--rebase --delete-branch`. No linked issue to close on this one — the lane charters from plan docs.

PR #349, opened 2026-08-16
Pin every spec-object wire name against its documented shape

### LINKED ISSUE #341: [Bug] Wire names on spec-object fields are unpinned — a misspelled [JsonPropertyName] drops the parameter with the suite green
Every tool that takes an array of spec objects declares its wire field names with `[JsonPropertyName]`. **No guard exercises those names.** Every probe constructs specs through the C# object initializer, which bypasses the attribute entirely — so a misspelled, renamed, or removed `[JsonPropertyName]` drops that parameter for every real MCP caller while the whole suite stays green.

This is a class, not a field. Filing it as one because fixing the instances one at a time leaves the next spec object added to the surface in exactly the same position.

## The mechanism

```csharp
// how every probe builds a spec — the attribute is never consulted
new PlaceAssetSpec { AssetPath = "meshes/x.nif", Source = "meshes/y.nif" }

// how a real MCP call arrives — the attribute decides whether the value lands
{ "asset_path": "meshes/x.nif", "source": "meshes/y.nif" }
```

Rename `[JsonPropertyName("source")]` to `[JsonPropertyName("sorce")]` and: the C#-initializer arms all still pass, the tool still builds, and every real caller silently loses the parameter — for `place_asset` that means a placement quietly reading a copy the caller never named, which is the silent-wrong-answer shape (Q3).

## Known instances

- `PlaceAssetSpec.Source` (`source`) — unpinned
- `PlaceAssetSpec.AssetPath` (`asset_path`) — unpinned
- `PlaceAssetSpec.Formid` (`formid`), `.Kind` (`kind`) — unpinned
- `PlaceAssetSpec.SourceProvider` (`source_provider`) — **pinned** by `place-asset-guard` arm L (landed in PR #339); it is the only one, and it is pinned by a hand-written per-field assertion, which is the shape this issue argues against scaling

The same construction applies to the other spec-object families on the surface (`bulk_apply`'s ops, `bulk_create`'s records, and any future one) — those have not been individually audited here, which is part of the point: nobody can tell from the guards which are covered.

## Fix shape — by construction, not per field

One arm that **reflects** over the spec record types, enumerates their `[JsonPropertyName]`-bearing properties, builds a JSON document naming every one of them, deserializes it, and asserts every property arrived non-default. Then:

- adding a property to a spec type without a wire name fails the arm,
- renaming a wire name so it no longer matches what the schema advertises fails the arm,
- and no per-field assertion has to be written or remembered.

This is the same posture the write surface already uses for its shared sentences: the inventory is reflected, not hand-listed, so it cannot fall behind the thing it describes (`WriteSurfaceGuardProbe`'s twin arm reflects `WriteSentences.Twins` for exactly this reason).

The reflection also gives the arm a natural completeness assertion — a spec type it cannot read is a failure rather than a silent skip, which is the trap a filtered hand-list would reintroduce.

## Provenance

Raised by an independent reviewer during PR #339's second review round, on `place_asset`'s spec object. Arm L closed the branch's own new field in that PR; the class was explicitly left for this issue rather than folded, since the sibling fields and the other spec families are outside that branch.

Avick3110:
Fixes #341

## The problem, and why the issue's own fix shape isn't enough

Every tool taking an array of spec objects declares its wire field names with `[JsonPropertyName]`, and nothing exercised those names — every probe builds a spec through the C# object initializer, which never consults the attribute. A misspelled or renamed wire name dropped that parameter for every real MCP caller with the whole suite green.

The issue proposes a reflection round-trip: read the attributes, build JSON from them, deserialize, assert every member landed. That catches a member that cannot bind — but **not a misspelling**, because it echoes the typo straight back, and the published JSON schema is generated from the same attribute, so it agrees with the typo too.

A wire name is only pinned against a **second, independently written statement of the same name**. That statement already exists and is the one that matters to a caller: the brace list in the carrying parameter's `[Description]` — `{formid, field_path, op?, …}`. This guard parses those and holds them against the reflected attributes, both directions.

## What lands

`wire-names-guard` (new, in `ci-all`). Discovery is by construction — the wire types are the assembly's `[JsonPropertyName]`-bearing types (9 today, 58 fields), and the carriers are the parameters and properties typed with them, so a spec object added later enrols itself with no edit to the guard.

| | |
|---|---|
| INV1 | every settable member carries `[JsonPropertyName]` or `[JsonIgnore]` |
| INV2 | every member arrives under its wire name through **all three** readers the surface uses — `ListParams.Strict`, the SDK's own `McpJsonUtilities.DefaultOptions` singleton, and `WriteTools.ManifestJson`. Arrival is checked against the value sent, so `BulkOp.Verb`'s `"Set"` default can't pass for arrival |
| INV3 | every wire type is reachable from a tool parameter and declares its shape |
| INV4 | every declared name is a real wire name |
| INV5 | every wire name is declared somewhere |
| INV6 | a shape list is read whole, or not read as a shape list at all — never partly and silently, at **every level** of the read path: characters (braces that don't balance, so the group is never found), groups (items that stop reading as member names), items (an ellipsis is a partial marker; anything else is not) |

The central RED arm is exhaustive: every one of the 58 fields, misspelled in turn, per reader, must be reported. Sabotaging `[JsonPropertyName("source")]` → `"sorce"` on the real surface fails INV4 and INV5 naming the field and the carrier.

**Documentation repairs the pin found.** `bulk_apply`'s `operations=` (no `composes`, no `from_plugin`), `bulk_create`'s `records=` (no `grid`), both create tools' `operations=` (no `composes` — it is legal there; `from_plugin` is not, and now says so), `sets=` inside a `compose=` (two of five members), `bulk_place_asset`'s `assets=` (no `source_provider`). Those members all worked; nothing told a caller they existed. Plus `source_provider`'s own description now states the no-`source=` lane, where it picks among the destination path's providers.

**What it does not cover, stated in the header rather than implied:** INV5 is satisfied by the union of a type's carriers, because a carrier may legitimately declare a subset — `create_record`'s `operations=` refuses `formid` and `from_plugin`, and "legal here" is semantic. Shape lists restated in tool prose or refusals are unread; two that this branch made diverge are repaired by hand, and `housecarl_create`'s pre-existing equivalent is left alone.

## Review rounds (CLAUDE.md §5 #11)

**Two rounds, two fresh agents each, own worktrees. Stopped at two by the recurrence rule, not by the cap.**

**Round 1** — both agents independently converged on the same three mediums:
- `composes` is legal on both create paths and declared by neither carrier — confirmed against `MapCreateEdit` → `MapComposes` → `Structs`. Folded.
- The `source_provider` gloss added here named the one combination that is *refused* (placing both FaceGen slots rejects a Data-relative `source=`) and omitted the lane that works. Confirmed against `MapSpec`. Folded.
- `bulk_apply`/`bulk_create` tool descriptions restate the shape and were left short, so the branch had them contradicting their own parameters. Folded.
- Plus: **INV2-RED was self-satisfying** on the strict lane — it matched text the probe itself stamped into the message. Removed; the reader's own refusal is the evidence. Verified by neutering the mangle: both lanes now fail where only one did.

**Round 2:**
- A shape list with one unreadable item was discarded **silently**, taking that carrier out of INV4's reach while the union kept INV5 green. Three sabotages that passed green now fail.
- The probe held **frozen copies of the surface** in four arms — correctly renaming or adding a member would have false-red them, pointing at the checker instead of the change. All four now use invented fixtures.
- Prose added in the round-1 fold overreached again.

**Refused, with reasons.** Padding the terse refusal hints (`{formid|asset_path, kind?, source?}`) into full shapes — they name what a caller must pass, not the element shape. Per-carrier completeness instead of the union — it would force authors to document members their tool rejects. Hardening INV2-RED's substring match to an identifier boundary — the reviewer who raised it could not construct a reachable case, and neither could I; it needs a second simultaneous failure, in which world the arm is already red.

**The recurrence, surfaced rather than folded a third time.** Two classes recurred across consecutive rounds: the parse layer discarding silently (round 1's ellipsis, round 2's any-bad-item), and prose I added to a caller-facing description being wrong (rounds 1 and 2). Both were closed at the seam rather than patched again — the parse layer now reports a partial read, and the `assets=` parameter carries the shape only, with the behaviour moved to the member's own description, which is its one home. Per §5 #11 there is no round 3.

`ci-all`: 123/123 green.


src/housecarl-generator/WireNamesProbe.cs
```diff
@@ -0,0 +1,688 @@
...
+    /// which for a type with several carriers left INV5 green off the others and INV4 never looking at that
+    /// carrier's text at all, so a misspelling in it went unseen. That is #341's own failure mode moved
+    /// into the docs, and closing an abbreviated list with an ellipsis is the natural way to write one.</summary>
+    static bool IsPartialMarker(string item)
+    {
+        var s = item.Trim().TrimEnd('.').Trim();
+        return s is "…" or "" || string.Equals(s, "etc", StringComparison.OrdinalIgnoreCase);
+    }
+
+    static List<string> OutermostGroups(string s)
```
Avick3110: **[medium] An unbalanced brace silently drops a whole shape list — the same hole INV6 closed, one level up.**

`OutermostGroups` only emits a group when a `{` is matched by a `}`. A description whose shape list loses its closing brace yields *no group at all*, so `DeclaredNames` returns empty and that carrier drops out of INV4's reach. INV6 cannot see it either: `PartialGroups` iterates the same `OutermostGroups`, i.e. only groups that already balanced. For a type with more than one carrier the union keeps INV5 green, so nothing is reported anywhere.

Verified on this branch: edit `create_record`'s `operations=` description (`WriteTools.cs:282`) to drop the trailing `}` **and** misspell `value?` → `valu?` in the same string. `wire-names-guard` runs **26/26 PASS** — while the published shape list is malformed and lists a name that is not a wire name of `BulkOp`. That is #341's own failure mode moved into the docs, which is exactly what round 2's fold set out to close for the partly-readable case.

Cheapest fix: make `OutermostGroups` report leftover depth (and a `}` seen at depth 0) so `PartialGroups` can name the carrier, rather than returning silently.

src/housecarl-generator/WireNamesProbe.cs
```diff
@@ -0,0 +1,688 @@
...
+    /// <summary>The readers the surface actually binds spec objects with. A wire name has to work in ALL of
+    /// them: <c>ListParams.Strict</c> reads apply/create's @file lists, the SDK's web defaults bind the
+    /// directly typed arrays (bulk_apply's operations=, bulk_create's records=, bulk_place_asset's assets=),
+    /// and <c>WriteTools.ManifestJson</c> reads bulk_apply's from_file= manifest. A name that survives only
+    /// some of them is broken for the others' tools. Three OBJECTS, not three configurations — the last two
+    /// are set up alike today, and each being driven here is what would catch the day one of them drifts.</summary>
+    static readonly (string Lane, JsonSerializerOptions Options)[] Readers =
+    {
+        ("ListParams.Strict — the @file list reader", ListParams.Strict),
+        ("web defaults — the SDK-bound spec arrays", new JsonSerializerOptions(JsonSerializerDefaults.Web)),
```
Avick3110: **[low] This lane is a reconstruction, not the object the SDK binds with.**

The doc above says "Three OBJECTS, not three configurations" and "the readers the surface actually binds spec objects with". That holds for `ListParams.Strict` and `WriteTools.ManifestJson` — both are the real statics, which is why widening their visibility was worth it — but this row is a fresh `JsonSerializerOptions` the probe builds itself. It is a second home for a fact whose home is the SDK's binder.

Faithful today (nothing in `housecarl-mcp` configures MCP serializer options — I checked). The failure mode is drift: if the SDK's tool-argument options ever gain a naming policy or custom resolver, or a registration later passes its own, this lane keeps passing while `bulk_apply`'s `operations=` silently drops the member for real MCP callers — the guard reporting coverage it does not have, which is the trap `Synthesize`'s UNCHECKED handling refuses elsewhere in this file.

Either bind to whatever options object the SDK exposes, or trim the claim to "the SDK's documented defaults, reconstructed" so the gap is stated rather than implied.

Avick3110:
## Fold summary — Aaron's review, 2/2 folded

Both findings reproduced against the branch before folding; zero refusals, because each cites a line and reproduces. Directed fold, so no new review rounds (§5 #11).

### Finding 1 (medium) — `OutermostGroups` dropped an unbalanced shape list silently

Confirmed, including the repro: dropping `create_record`'s `operations=` trailing `}` **and** misspelling `value?` → `valu?` left the guard at **26/26 PASS**. An unclosed `{` yields no group at all, so the list isn't merely unread — it's invisible: `PartialGroups` never saw the carrier, INV4 never read its text, and the union kept INV5 green.

**This was the third instance of the silent-discard class** (round 1: ellipsis-closed list; round 2: unreadable item; now the lexer beneath both). Round 2 closed the seam at the *group* level and didn't reach the *character* level, so the fix is at that level with the same posture rather than another patch:

- `ScanBraces` returns the imbalances alongside the groups — leftover open depth, and a `}` at depth zero — and `PartialGroups` names the carrier and the character in the voice of its existing report.
- Arms per branch, RED both directions: **unclosed reported**, **stray-close reported**, **balanced (nesting included) NOT reported**.
- INV6's header claim now states the completeness at every level of the read path — characters, groups, items — since all three had to be found before it was true.

**Acceptance: Aaron's exact double sabotage now fails.** Run by hand against the real source (restored after):

```
INV6-GREEN every shape list is read WHOLE, or not read as one at all : FAIL
  - housecarl_create_record's operations=: the braces do not balance — the '{' at character 151
    is never closed (1 level(s) still open at the end) — everything it was meant to declare is
    invisible to this guard. Close it, or drop the stray brace
=== wire-names-guard: 29 passed, 1 failed -> FAIL ===
```

Worth naming which arm catches it: **INV4 and INV5 stay green** under that sabotage — the list is invisible and the union covers the type from another carrier. INV6 is the whole of the catch, which is the argument for having it.

The in-probe arm uses an **invented fixture**, not a copy of the real description — copying one back in is the frozen-copy trap round 2 removed. Hence the hand-run above as the real-source acceptance.

**Known cost, in the header rather than implied:** a carrier whose prose carries a lone brace for some other reason is now a reported violation the author must reword. No carrier does today, and the report names the carrier and the character. No quote-tracking added to the lexer — that's a refinement for a real description that trips it, not a speculative one.

### Finding 2 (low) — the web lane was a probe-built options object

Confirmed: it contradicted the header's own "three OBJECTS, not three configurations". **Preferred arm taken** — the lane now drives `ModelContextProtocol.McpJsonUtilities.DefaultOptions`, the SDK's public singleton.

How far that's verified, stated rather than assumed:
- **Exposed** — public property on the SDK's `McpJsonUtilities`.
- **Documented flow** — the SDK documents `McpServerToolCreateOptions.SerializerOptions`, "the JSON serializer options to use when marshalling data to/from JSON", as defaulting to it; `Program.cs` registers with a bare `WithToolsFromAssembly()` and never sets it.
- **Not pinnable by observation, and said so** — on the read side it behaves identically to bare Web defaults (its extras are a write-side ignore condition, AOT source-gen contracts, and number-reading-from-string, which Web already enables). I drove the real exe over stdio to try: `max_chars` bound from both a JSON string `"500"` and a JSON number `500`, which Web permits too — so no observation separates the instances.

Driving the real singleton is as close as the guard can get, and the residue is Aaron's fallback line: **RUN_ORDER step 5's #329 row now carries "re-verify `wire-names-guard`'s SDK reader lane"** at the SDK bump.

### Contract-prose sweep

- Guard header: INV2 (three readers, named accurately), INV6 (all three levels), the SDK-lane verification paragraph, the lone-brace cost.
- PR body: INV2 and INV6 rows updated above.
- CHANGELOG: **deliberately unchanged.** Its claims are about the caller-facing shapes and what the check does and doesn't read; INV6 is internal robustness and makes nothing in it more or less true. Flagging the no-change rather than leaving it silent.

### Standing class boundary (recorded in the header's not-covered paragraph)

If a **fourth** silent-discard instance surfaces after this, the prose-parsing design itself escalates under §4 — the second spelling moves out of `[Description]` prose into structured data (attribute or registry), where it cannot be mis-lexed. Not a fourth fold.

---

`wire-names-guard` 30/30; `ci-all` 123/123 locally. Pushed as `61378dd`. Waiting on your go to merge.

Issue #351, opened 2026-08-16
[GAP] CI doesn't check the committed mutagen-reference shards against a fresh emit


Avick3110:
Raised independently by three reviewers during #335's pre-PR review rounds, and routed by Aaron to the step-5 dependency-bump window (RUN_ORDER) rather than into that PR.

## The gap

`.claude/skills/mutagen-reference/references/*.jsonl` is a **committed** artifact generated by `CorpusGenerator`. Nothing in CI compares it against a fresh emit:

- `ci-all`'s corpus guards (including #335's new `corpus-hygiene-guard` INV6) regenerate a corpus into a temp dir and assert against **that**, never against the committed shards.
- `.github/workflows/ci.yml` runs build + `ci-all` + `freshness-capture-guard`; there is no `git diff --exit-code` after a regeneration.
- `generated/` is gitignored, so the full corpus is not committed either.

So a classifier or emitter change that lands **without** regenerating leaves the shipped reference stale, with CI fully green — the same class of defect as #335 (a reference that disagrees with what the code models), reachable by omission rather than by a bug.

## What is not affected

The packaged plugin regenerates at build time (`scripts/build-plugin.ps1`), so a release ships a fresh reference. The exposure is the repo copy — which is what a dev-mode skill reads, and what a reviewer reads when checking a classification claim.

## Shape of the fix

A CI step that regenerates into a temp dir and diffs against the committed shards (or a guard arm that does the same comparison), failing with the differing file/line named. Cheap; the generator already emits to an arbitrary `refDir`.

## Routing

Step 5 of RUN_ORDER (dependency bumps, attended) — the Mutagen bump is exactly when a stale committed reference would do the most damage, since coverage IS Mutagen's coverage. Aaron's note during #335: if the #330 slot stays distant, this is a candidate small standalone PR — his sequencing call.

Issue #350, opened 2026-08-16
[GAP] owned child record lifecycle — a singular one can't be created where the parent has none, and can't be deleted at all


Avick3110:
Split out of #335 (Aaron-ruled during that PR's directed completion). #335 corrects the *classification* of the two singular owned-child fields Mutagen models — `Cell.Landscape` and `Worldspace.TopCell` — and gives every write verb a deliberate answer for that shape. Two of those answers are refusals because the capability does not exist. This issue is that capability.

## Half 1 — an absent owned child cannot be created

Measured on a synthetic instance (real tool methods, houseCARL at `claude/cell-landscape-classifier`):

```
create_record Landscape parent=<cell FormID>
  -> refused: 'Landscape' cannot be created under a 'Cell' — that parent models no
     child-collection that holds it (a real containment boundary, surfaced not guessed, Q3)

create_record Landscape (no parent)
  -> refused: 'Landscape' has no top-level group, so it can't be created on its own

create_record PlacedObject parent=<cell FormID>       (control, the LIST child shape)
  -> refused only for the missing collection= discriminator — the shape itself works
```

`parent=` resolves a child *collection* on the parent. A singular owned child is a property, not a collection, so nothing routes it. A cell that carries no LAND cannot be given one, and a worldspace with no top cell cannot be given one.

Consequence inside #335: a sub-field write through the parent is refused when the copy being written carries no child (and a patch override of a parent never carries one), so the refusal states the boundary instead of naming a remedy that does not exist.

## Half 2 — an existing owned child cannot be deleted

```
housecarl_remove <LAND formid> in_place=<its defining plugin>
  -> error: Remove did not drop 1 record(s) — the engine no-op'd without throwing

housecarl_remove <LAND formid> into=<patch that carries an override of it>
  -> same

housecarl_remove <CELL formid> in_place=…      (control)  -> removed 1 record
housecarl_remove <PlacedObject formid> in_place=…  (control) -> removed 1 record
```

Independently reproduced by a second reviewer during #335's review rounds, who traced it to Mutagen's typed `Remove(FormKey, Type)` no-opping on an owned child while `RemoveSurvivors` refuses.

Consequence inside #335: `Remove` on the field itself is **refused**, on the ground that a keyless clear of a singular child deletes an entire subtree implicitly, in one call, with no backup on the in-place lane — where the list family's by-index delete has an explicitly named target. Legalizing deliberate owned-child deletion is a capability call, not a side effect of a classification fix.

[...]

Issue #343, opened 2026-08-13
First-write auto-naming ("Patch.esp") doesn't surface housecarl_create_plugin as the named alternative


BumblingOaf:
## Summary

When starting a brand-new patch (via `housecarl_create_record` / `housecarl_bulk_apply` /
`housecarl_set_field` without an `into=` naming an existing patch), the tool auto-creates a
generically-named `Patch.esp`. `housecarl_create_plugin plugin_name="..."` exists and lets you
start a patch with a specific name from the outset, but neither the docs I could find nor the
"can't extend, no such plugin" error message when guessing at `into=` for a not-yet-created name
mentions it.

## Concrete repro

First attempt — guessing `into=` names a patch that doesn't exist yet:
housecarl_create_record ... into="My Cool Patch.esp"
→ error: cannot extend: no houseCARL plugin 'My Cool Patch.esp' in any houseCARL folder...
Omit into= to create it fresh, or check the name.

Following that instruction (omitting `into=`) produces a patch named `Patch.esp` — not the
intended name. `housecarl_create_plugin` is the actual correct tool for this, discovered only by
separately trying it out of general curiosity about the tool list, not from anything the error
pointed at.
## Why this matters
For anyone planning multiple patches (multiple weapons, multiple unrelated fixes), ending up with
a generically-named `Patch.esp` on the first try means a costly do-over: there's no rename
primitive, and `housecarl_merge_plugins` can't consolidate a single mis-named patch into a
correctly-named one either (see separate issue). Getting the name right on the very first write
matters a lot given how expensive it is to fix after the fact — and the tool doesn't currently
point you at the way to do that.
## Suggested fix
Add a line to the "cannot extend" error pointing at `housecarl_create_plugin plugin_name=` as the
way to start a new patch with a specific name, e.g.: "Omit into= to create a generically-named
patch, or use housecarl_create_plugin plugin_name='...' to start one with a specific name."
## Environment
- houseCARL v1.9.0

Avick3110:
Triaged: the mechanics **confirm** (the default really is `Patch.esp` — auto-suffixed if taken — and the "cannot extend" remedy text names no alternative), but the suggested fix points at the wrong tool. The named-patch route already exists: every write tool takes a `patch=` parameter (spelled `patch_name=` on v1.9.0) — "base filename for the NEW patch this call writes (default 'Patch')". No `create_plugin` detour needed; that tool stays the header-only-plugin case.

[...]

Issue #342, opened 2026-08-13
Cell.Persistent/Temporary reads only the winning override's own list, not the additive union real cell references use


BumblingOaf:
Hello! I was using houseCARL today to try to pinpoint why I had a missing khajit caravan and missing ferry in Dawnstar, and ran into a couple issues. I wouldn't say these were bugs or anything, but moreso gaps that houseCARL was confused by and not able to really help. Figured I would send this summary I got from Claude along in case this can enhance houseCARL logic going forward.

-BumblingOaf on discord, we were talking for a bit today about houseCARL.

## Summary

`housecarl_read_record` / `housecarl_cross_plugin_query` / `housecarl_batch_record_detail` on a
`Cell` record's `Persistent` and `Temporary` fields return only the **winning plugin's own
declared list** for those fields. In the actual Skyrim engine, cell references merge
**additively** across every plugin that touches a cell — a plugin that overrides a cell for an
unrelated reason (e.g. adding occlusion-culling data) doesn't erase other plugins' placed content
in-game. But reading these two fields through the current "winner" model makes it look like the
cell is empty, which is misleading and can send troubleshooting in the wrong direction.

## Concrete repro

Load order includes a modlist where `Occlusion.esp` (a performance/occlusion-culling mod) touches
Dawnstar's exterior cells purely to add occlusion planes.

Reading a Dawnstar exterior cell's `Persistent`/`Temporary` at the winning layer:
housecarl_batch_record_detail formids=["008EB5:Skyrim.esm"] fields=["Persistent","Temporary"] depth=2
→ winner=Occlusion.esp
→ Persistent = [list: 0 item(s)]
→ Temporary = [list: 0 item(s)]

Reading the *same* cell scoped to `Skyrim.esm`'s own original declaration (`plugin="Skyrim.esm"`):
→ Temporary = [list: 201 item(s)] (guards, patrols, dressing, etc.)

Every one of the ~12 Dawnstar exterior cells I checked showed this same pattern: winner
(Occlusion.esp) = 0 items, vanilla Skyrim.esm = dozens to 200+ items. In game, none of that
content is actually missing — Occlusion.esp never touches placed references at all, it just
happens to be the last plugin to touch the cell record for an unrelated reason. Reading the
"winner" for these two fields specifically produces a false-empty result.
## Why this matters
I was helping a user diagnose "an NPC's compass marker exists but nothing is there" — the
investigation went down the path of checking whether the cell's content was actually missing
[...]

PR #352, opened 2026-08-16
fix(reference): Cell.Landscape and Worldspace.TopCell are owned child records, not FormLinks

### LINKED ISSUE #335: [Bug] mutagen-reference classifies Cell.Landscape as a formlink, but it is an owned child record
Spotted by Aaron reviewing PR #334; out of scope there.

## What

`.claude/skills/mutagen-reference/references/records.jsonl:17` classifies **`Cell.Landscape`** as `"c":"formlink"`, with no target type.

On the concrete Mutagen type it is an owned child record, not a reference. PR #334's reflected containment walk explicitly cuts `IFormLinkGetter` and still finds `Landscape` in Cell's child-bearing set — and that set is now pinned in `write-surface-guard` across every concrete record type Mutagen models:

```
Cell = Landscape, NavigationMeshes, Persistent, Temporary
```

Both classifications cannot be right.

## Why it matters

The reference is what a skill or a session consults before writing. A caller trusting `formlink` would try to set a FormID on a property that holds an owned record — the write is wrong in kind, not just in value. `NavigationMeshes` / `Persistent` / `Temporary` are worth checking in the same pass, along with any other property the generator's classifier routes to `formlink` on the strength of a nullable-reference shape.

## Where to look

The generator's field classifier (`CorpusGenerator`), specifically how it decides `formlink` vs an owned nested record. The reference is generated by construction from Mutagen's record library, so this is a classifier rule to fix, not a line to hand-edit — and the fix should come with whatever pin makes the classification checkable against the same by-construction sweep #334 added.

Avick3110:
Fixes #335. Refs #342 (read side, deliberately not implemented here), #350 and #351 (both filed out of this branch's measurements).

## The defect

`IMajorRecordGetter` carries `IFormLinkIdentifier` — a record identifies itself (FormKey + Type) exactly the way a
link identifies its target. The generator's bare-link rule tested only for `IFormLinkIdentifier`, so it could not
tell *points at a record* from *is a record*, and every field typed as a major record was emitted as a FormLink.

Measured before the fix:

```
GATE  Cell.Landscape  Set = 000800:Skyrim.esm   -> ACCEPTED
APPLY Cell.Landscape  Set = 000800:Skyrim.esm   -> InvalidOperationException: No coercion rule for
                                                   Mutagen.Bethesda.Skyrim.Landscape
```

Accept-then-throw: pre-flight cleared a write that is wrong in kind.

## The fix, and what it turned out to be

Cutting major records from the bare-link rule is three lines. The rest of this PR exists because that correction
put a shape into the corpus that had **never existed**: before it, no field was a `substruct` whose `TypeRef` is a
record. Every rule keyed on "substruct" was written when that could not occur, so each read the two fields as an
ordinary navigable sub-object. They are not one.

Review round 1 found four such doors; round 2 found a fifth *after* the first four were closed and the docs already
claimed completeness. That is a recurring failure class, so per CLAUDE.md §5 #11 the rounds stopped and the seam
went to Aaron, who ruled: complete the class here, with a disposition table pinned by arms, and measure every
remedy before a refusal names it.

## Every classification that changed — the whole regenerated diff

| Field | Before | After |
|---|---|---|
| `Cell.Landscape` | `{"c":"formlink","t":"FormLink"}` | `{"c":"substruct","t":"ILandscapeGetter","ref":"Landscape","null":true}` |
| `Worldspace.TopCell` | `{"c":"formlink","t":"FormLink"}` | `{"c":"substruct","t":"ICellGetter","ref":"Cell","null":true}` |

Nothing else moves: `index.jsonl`, `structs.jsonl`, `arms.jsonl`, `enums.jsonl`, `polymorphic.jsonl` are
byte-identical, no catalog entry is added or removed, and two reviewers independently regenerated the tree and
diffed it against the committed shards. `null:true` arrives because the substruct branch reads the getter's
nullable-reference annotation, which the formlink branch returned before ever reaching.

The other fields #335 asked to be checked in the same pass were already correct, and the pin proves it rather than
my eye: `Cell.NavigationMeshes`, `Cell.Persistent`, `Cell.Temporary`, `Worldspace.SubCells`,
`DialogTopic.Responses`. Two indirections matter there — containers and polymorphic arms — and a one-hop reach walk
reported `Persistent`/`Temporary` as reaching no record, which is why the pin's walk follows both.

## The pin (#335's acceptance line)

`corpus-hygiene-guard` gains **INV6**: over `WriteSurfaceGuardProbe.ConcreteRecordTypes()` — write-surface-guard's
own by-construction sweep, borrowed rather than re-listed — every property `WriteEngine.ChildBearingProperties`
finds must be cataloged with an ownership cardinality whose ref chain reaches a record entry, never a formlink.
GREEN over the real corpus (RED before the fix, naming exactly the two fields), plus a synthetic-formlink RED, a
reaches-no-record RED, and both classifier branches driven directly on real Mutagen types (`ILandscapeGetter` must
be an owned child; the bare non-generic `IFormLinkGetter` must still be a formlink — without that arm the bare-link
rule could quietly stop firing, since nothing in Mutagen 0.53.1's Skyrim model exercises it).

## The disposition table — the part meant to kill the class

One place lists **every op verb against the owned-child shape with the answer it must give** (17 rows, in
`write-surface-guard`), so "which door did we miss" is a CI question instead of a review hope:

| Caller does | Answer |
|---|---|
| `Set value=` (a FormID — what the old classification invited) | refused, names the record |
| `Set compose=` | refused, same sentence |
| `Add composes=` | refused, same sentence |
| `Remove` (keyless) | refused — clearing deletes the record and its subtree |
| `Remove` with a key | refused (no element to name on a singular child) |
| `CopyFrom`, both fields | refused — names `forward_record`, routes to #350 |
| `Add` / `ReplaceAll` / `SetAtIndex` / `Merge` | the ordinary cardinality refusals |
| descend to a sub-field | accepted at the gate (live state the corpus cannot know) |
| **controls**: ordinary substruct clears / composes, list form deletes by index, list form refuses `CopyFrom`, `composes=` on an ordinary substruct | unchanged |

The controls are the point: a clause that widens past the record shape turns a control red rather than passing
quietly. The `composes=` control was added *because* the widening sabotage left that door untested.

## What the completion changed beyond the classifier

- **`CopyFrom` was the open door.** Measured: `CopyFrom` on `Worldspace.TopCell` deep-copied the source's whole
  CELL — its own FormID, its persistent references and all — in as the destination's child. That is the silent
  child-record import `RestoreChildGroup` refuses by name on the forward path. Now refused.
- **`composes=` was a third door** into the loop the Set arm closed (it ended by pointing at `compose=` / `value=`,
  both of which refuse). All three Set-shaped doors now give one sentence, and an arm asserts they are identical.
- **`Remove` keeps its refusal, with the rationale replaced.** The list form of this family *does* delete an owned
  child — by index, with the target named. The honest ground is that a keyless clear takes a whole subtree
  implicitly with no backup on the in-place lane, and that deliberate deletion is #350, not a clear-verb's call.
- **Materialization** names the boundary instead of borrowing the composition deferral ("buildable only from its
  parts" named a wave that will never deliver a record).
- **The write census** walked `Worldspace → TopCell → Cell` and counted the whole nested-group world as reachable
  today (+87 leaves; `Cell`'s 44 fields leaving the deferred bucket while axis 1 still printed them as deferred).
  The substruct axis now carries the `!IsRecord` guard its collection twin already had, and both owned-child leaves
  are subtracted from "writable today" — 2900/8167, the pre-branch number, totals balancing.

## Remedies, measured before being named

Every remedy in these refusals was run first; two candidates were rejected *because* the measurement refused them:

| Remedy | Measured | In the text? |
|---|---|---|
| address the child record by its own FormID | `housecarl_apply` wrote and read back | yes |
| `housecarl_forward_record` on the child | carries a LAND and a worldspace top cell | yes |
| `housecarl_create_record parent=` | refused: "that parent models no child-collection that holds it" | no — routes to #350 |
| `housecarl_remove` on the child | refused on both lanes (engine no-op'd); removing the *cell* works | no — routes to #350 |

I caught myself naming `housecarl_remove` in a refusal before measuring it; that is the #343 class and it is why
the table above exists.

## Review rounds

Two rounds, two fresh agents each, every agent in its own worktree.

- **Round 1** — `Remove` became legal and deletes the child record plus everything under it (nullability arrived
  with the substruct branch); the compose/value refusal loop; the census over-count; and my own docs naming a
  remedy the default lane cannot reach (a patch override of a parent carries no children — measured
  `Landscape=(null) persistent=0`). All reproduced independently before folding.
- **Round 2** — `CopyFrom`, found by both agents; the `composes=` third door; the census split between the two
  identical fields; my apply refusal mis-diagnosing the in-place lane; an untrue doc line I had written on the
  acceptance arms. Same class as round 1 → stop, escalate, directed completion.

**Refused, with grounds:**
- *"`housecarl_remove` is exactly this operation, so the refusal's claim is false"* — cited documentation, not a
  measurement. Two independent probes say it no-ops on a LAND. The narrower half of that finding was accepted: my
  *rationale* (the list family refuses) was wrong, and it has been replaced.
- *`CopyFrom` as a pre-existing defect to file rather than fix* — refused after Aaron's ruling: the classification
  change is what makes its wrongness documented rather than latent, so it belongs in this PR. (Filed nothing; fixed
  here.)
- *A CI check that the committed reference shards match a fresh emit* — real, and flagged by three reviewers, but a
  workflow change beyond #335. Filed as #351, routed to the step-5 window.

**Not folded, filed instead:** #350 (an absent owned child cannot be created; an existing one cannot be deleted —
both halves measured) and #351. Two refusal texts route to #350 by number, and an arm pins that they do.

## Verification

- `ci-all`: **123/123**, Release build.
- `corpus-hygiene-guard` 17/17 (15/2 against the unfixed classifier) · `write-surface-guard` 216/216.
- Every conditional this PR adds is sabotage-checked in **both** directions — disabled and widened — each failing
  its own row: the classifier cut, the materialization carve-out, the `Remove` clause, the Set/compose clause, the
  `composes=` route, and the `CopyFrom` clause.
- `EMPIRICAL_LEDGER.md` unchanged — nothing deferred; every claim above was measured locally on constructed
  instances and a synthetic MO2 instance, no live install required. `copy_npc_appearance` untouched (EL-2 holds).

## One judgement left explicit

`[MustState]` pins live in the mcp response layer for write *outcomes*; these are core-side gate refusals, which
that layer does not reach. The disposition rows are their equivalent here — they assert the phrases each refusal
must carry, including the `#350` routing.


src/housecarl-core/CorpusRulebook.cs
```diff
@@ -372,17 +394,34 @@ static bool IsFlagsEnumLeaf(FieldSchema leaf)
...
+    /// Left ungated, this was the one door still open after #335 closed Set / compose / composes / Remove: measured, a
+    /// CopyFrom on <c>Worldspace.TopCell</c> deep-copied the source's whole CELL — its own FormKey, its persistent
+    /// references and all — into the destination's worldspace, which is the silent child-record import
+    /// <see cref="WriteEngine.RestoreChildGroup"/> refuses by name on the forward path.</summary>
     string? CopyFromLegality(FieldSchema leaf, TypeSchema owner)
     {
         if (leaf.IsIdentity)
             return $"'{leaf.Name}' on '{owner.Name}' is record identity (FormKey/ModKey), not a copyable content field.";
         if (!leaf.Writable) return WritabilityRejection(owner, leaf);
+        if (SchemaClassifier.IsOwnedChildRecord(leaf, _corpus))
```
Avick3110, line 443: **[correctness] `CopyFrom` *through* the owned child is unguarded — a door this branch newly opened.**

Before this branch `Landscape` was `formlink`, so the mid-path hop refused: *"Cannot descend through 'Landscape' on 'Cell': it is a formlink, not a substruct."* The reclassification opens that hop for **every verb** (the descent at `CorpusRulebook.cs:138` admits any substruct whose `TypeRef` resolves), and only `Set` was measured — the table's one accept row is a `Set` on `Landscape.EditorID`.

`CopyFromLegality` inspects the **leaf** only. With a path *through* the child:

```
apply target=<cell FormID> path=["Landscape","VertexColors"] verb=CopyFrom from_plugin=X
```

leaf = `VertexColors` on `Landscape` → not a record, not a record collection, not a dict → **accepted**. Apply then walks the source cell's LAND and the target cell's LAND (`WriteEngine.CopyField`) and writes into the target's LAND — two different LAND FormIDs, neither named by the caller, reported as an edit to the CELL. That is a partial form of what the refusal added on this line forbids ("copying it here would write another plugin's record … in as this parent's child"), and no row goes red.

Reachable on the in-place lane, and on the default lane once a patch's CELL carries a restored child from a prior `forward_record` (`RestoreChildGroup` puts it there). Same shape for `Remove` / `Add` / `SetAtIndex` at a leaf under the child: they mutate a record the call never names.

The cheap fix is the same predicate applied to the **path**, not just the leaf — refuse `CopyFrom` when any hop is an owned child record, and give it a row.

src/housecarl-core/SchemaClassifier.cs
```diff
@@ -109,15 +109,31 @@ static ElementKind PolyBaseElementKind(string baseName, Corpus corpus)
...
+    /// <summary>True iff a SUBSTRUCT leaf holds an OWNED CHILD RECORD — the parent carries the record itself, rather
+    /// than a link to one (<c>Cell.Landscape</c>, <c>Worldspace.TopCell</c>). The singular twin of
+    /// <see cref="ElementKind.Record"/>, and the ONE home for the question: until #335 corrected their classification
+    /// no field in the corpus was a substruct whose TypeRef is a record, so every rule keyed on "substruct" was written
+    /// when this could not occur and reads such a leaf as an ordinary navigable sub-object. It is not one. A record has
+    /// its own FormKey and its own children, so it cannot be built from parts, cannot be materialized into an absent
+    /// slot, and must not be dropped as a side effect of clearing a field. Rules that would otherwise inherit the
+    /// sub-object answer ask HERE, so a later consumer meets one named predicate rather than re-deriving the test —
+    /// corpus-derived (Kind), never a list of field names (cornerstone §3).</summary>
+    public static bool IsOwnedChildRecord(FieldSchema f, Corpus corpus) =>
```
Avick3110: **[correctness, latent] The predicate is keyed on `substruct`, so one record-typed shape bypasses all 17 rows.**

After the classifier cut, a major-record-typed field falls through to `EnqueueModeledRef`, which sets `Cardinality = "polymorphic"` — not `"substruct"` — whenever the getter interface has more than one concrete arm. For a record **polymorphic base** (`IPlacedGetter`, `IGameSettingGetter`, `IAStoryManagerNodeGetter`) a singular field would therefore catalog as `polymorphic` with record arms, `IsOwnedChildRecord` returns false, and every clause keyed on it — Set / compose / composes / Remove / CopyFrom — is skipped. `Set compose={type:'PlacedObject',…}` would then route to `ArmLegality` → accepted → `BuildStruct` → `Instantiate` throws `CompositionRequiredException`, surfaced as *"pre-flight ACCEPTED it but the apply threw — a real inconsistency"*. #335's own defect class, one shape over.

No such field exists in Mutagen 0.53.1's Skyrim model (the regenerated diff moves exactly two fields), so this is latent, not live. Raising it because the doc comment calls this "the ONE home for the question" and it answers only for the shape that happens to exist today. INV6 does net the cataloging — a child-bearing property with a non-ownership cardinality violates — but it reports it as *"a link references a record, it does not own one"*, which is not what a polymorphic field is, so the message would send the reader at the wrong fix.

Deriving the predicate from the TypeRef's `Kind` for both ownership cardinalities (`substruct` **or** `polymorphic` whose arms are records — `PolyBaseElementKind` already answers that question for elements) keeps it corpus-derived and closes the shape.

src/housecarl-generator/CorpusHygieneProbe.cs
```diff
@@ -354,6 +409,96 @@ static List<string> CheckDegenerateType(Corpus c)
...
+    {
+        seen ??= new HashSet<string>(StringComparer.Ordinal);
+        foreach (var arm in f.ElementArms ?? f.Arms ?? new List<string>())
+            if (EntryReachesRecord(c, arm, seen)) return true;
+        return (f.ElementTypeRef ?? f.TypeRef) is { } tr && EntryReachesRecord(c, tr, seen);
+    }
+
+    /// <summary>…and the same question about a catalog ENTRY: it is a record, one of its arms reaches one, or one
+    /// of its own ownership fields does.</summary>
+    static bool EntryReachesRecord(Corpus c, string name, HashSet<string> seen)
```
Avick3110, line 513: **[correctness] The ref walk never follows a polymorphic FIELD, which makes `OwnershipReachesRecord`'s `f.Arms` term dead.**

`OwnershipReachesRecord` reads `f.ElementArms ?? f.Arms`, but both of its callers gate on `f.Cardinality is "substruct" or "list" or "dict"` — `CheckOwnedChildAsLink` before calling it, and line 469 here. A field that carries `Arms` (rather than `ElementArms`) is by construction cardinality `polymorphic`, so the `f.Arms` term can never be reached: entry-level arms are covered by `t.Arms` on line 466, **field**-level arms are covered nowhere.

Effect: a type that reaches its records only through a polymorphic field whose arms are records reports *"reaches no record entry"* and fails INV6 on a corpus that is correct — a false red whose message points at the classifier rather than at this walk. The direction is safe (never a false green), but the `f.Arms` term reads as though the case is handled, which is the part worth fixing: either follow polymorphic fields in both gates, or drop the term and say the walk is ownership-cardinality only.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -393,6 +432,22 @@ static bool IsFlagsEnumLeaf(FieldSchema leaf)
...
+    /// and <c>composes=</c> alike (#335). Shared rather than phrased per door because the three doors previously gave
+    /// three answers that pointed at each other: compose= said "use a plain value", value= said "navigate into it",
+    /// and composes= said "a substruct takes compose= / value=" — all three refused.
+    /// <para/>
+    /// The remedy it names is the MEASURED one: addressing the child record by its own FormID writes and reads back on
+    /// the default lane. The descent clause is stated conditionally on purpose — a path through the parent reaches a
+    /// child only when the copy being written already carries one, and a patch's override of a parent never does
+    /// (Mutagen's override copy leaves the children behind), so an unconditional "descend into it" would send a caller
+    /// at a state the default lane cannot produce.</summary>
+    static string OwnedChildSetRefusal(FieldSchema leaf) =>
```
Avick3110, line 470: **[message quality] The remedy names the FormID but not where to get it.**

"Address the child record itself by its own FormID" is the right remedy — the branch measured that it works — but the caller who just tried to write at the field does **not** have that FormID, and none of the three refusals that name it (this one, `CopyFromLegality`, `MaterializeSubstruct`) says where it comes from. The next move is a guess, which is the same dead-end shape the three Set doors were unified to remove.

It is one clause away: a read of the parent already renders the child with its own id — `ReadEngine.ElementSummary` leads an `IMajorRecordGetter` with its FormKey, so the field reads back as `Landscape = [Landscape 000D74:Skyrim.esm editorid=…]`. "Read the parent — the field shows the child's FormID" closes the loop the way the CopyFrom refusal closes it by naming `housecarl_forward_record`.

(Two smaller notes on the same sentence, take or leave: the parenthetical `(gap #350)` lands on the override-copy clause, where it parses only on the second read — the Remove and CopyFrom refusals attach it to the claim it actually explains; and this sentence's `#350` is the one of the three that no arm pins, since the routing arm asserts only Remove and CopyFrom.)

src/housecarl-generator/WriteSurfaceGuardProbe.cs
```diff
@@ -1328,6 +1346,180 @@ static void ForwardChildGroupArm(Fixture fx)
...
+        var ownedChild = new (string What, WriteRequest Req, string? MustSay)[]
+        {
+            ("Set value= (a FormID, the shape the old formlink classification invited)",
+                new WriteRequest { RecordType = "Cell", Path = new[] { "Landscape" }, Verb = "Set", Value = "000800:Skyrim.esm" },
+                "owned child RECORD"),
+            ("Set compose= (build the child from parts)",
+                new WriteRequest { RecordType = "Cell", Path = new[] { "Landscape" }, Verb = "Set",
+                    Struct = new StructSpec { Type = "Landscape", Fields = new Dictionary<string, string> { ["EditorID"] = "X" } } },
+                "owned child RECORD"),
+            ("Add composes= (the third door: a LIST of built elements)",
```
Avick3110, line 1443: **[test coverage] `composes=` is pinned on `Add` only; `ReplaceAll` is the other verb that reaches the same clause.**

`ComposesLegality` admits `Add` **or** `ReplaceAll` before reaching the owned-child clause, and `ReplaceAll composes=[]` is separately meaningful — it is the modeled-list CLEAR. Both are refused today by the one clause, but neither has a row, so an edit that scopes the clause to the append verb (`req.Verb == "Add" && IsOwnedChildRecord(...)`) reopens `ReplaceAll composes=[…]` **and** `ReplaceAll composes=[]` on `Cell.Landscape` with all 17 rows still green — the empty form being the clear-the-child door the `Remove` row exists to keep shut.

Two more entries on the same array, asserting `owned child RECORD`, close it.

On the wider question the table is meant to answer: I walked the leaf surface independently (every verb × every input slot: `Value` / `Values` / `Entries` / `Key` / `Struct` / `Structs`, plus the `@editorid` same-call branch and the unknown-verb default) and found no leaf-shaped door the rows miss — `@editorid` is refused by cardinality, `Add compose=` and `Set composes=` land on the generic verb rejects, `Set key=` on the no-key reject. The doors the rows do not reach are not at the leaf: they are the mid-path ones the descent row admits (separate comment on `CopyFromLegality`), and the shape that never becomes `substruct` at all (separate comment on `IsOwnedChildRecord`).

Avick3110:
## Fold summary — review findings F1–F5, all five folded (5/5, 0 refused)

**Why nothing was refused.** Every finding reproduced against the source before folding. F1 and F5 I confirmed by driving the gate; F2 and F3 are structural claims I read in the code and then pinned with arms that go red without them; F4's mechanism I measured — and the measurement changed the fix (below). A 100% fold rate is the triage signal this repo asks to be stated, so: nothing here was taken on the reviewer's word.

### F1 — `CopyFrom` through the child (path, not leaf)

The predicate now runs over the **hops**, so transplant refuses at any depth. Rows added for both fields.

The in-place verbs through the same hop stay **accepted**, per the ruling, and the split is now pinned rather than assumed: `Set` and `Remove` at a leaf under the child are accept rows, `CopyFrom` through the same hop is a refuse row, and a control keeps `CopyFrom` through an ordinary substruct path accepted. The code comment states the distinction — *transplant refuses at any depth; in-place descent is the accept row's territory* — and names why: those edit the child this record already carries, which is the only way to edit a carried child at all.

### F2 + F3 — one change

`IsOwnedChildRecord` now derives from the TypeRef's `Kind` across **both** singular ownership cardinalities (`substruct`, or `polymorphic` whose arms are all records, via `PolyBaseElementKind` so the singular and element answers cannot drift). Both hygiene-walk gates widened to match, which makes `OwnershipReachesRecord`'s field-level `f.Arms` term live instead of dead, and INV6's not-ownership message splits so the polymorphic case names the arms rather than the ref.

Three RED arms: a synthetic record-armed polymorphic field through the predicate (fails when the predicate goes back to substruct-only), a struct-armed one that must stay false (the carve-out is records, not polymorphism), and a type reaching records only through a polymorphic field through the walk (fails when the gate narrows).

**Zero corpus lines moved** — the regenerated shards are byte-identical to the committed ones, all six (`index`, `records`, `structs`, `arms`, `enums`, `polymorphic`), checked with `cmp` after the change.

### F4 — where the FormID comes from

Measured before writing, and the measurement corrected the sentence:

```
depth 1:  Landscape = [Landscape] — pass depth=2 to expand      ← no id
depth 2:  Landscape = [Landscape 000801:ZzF4.esm editorid=ZzLand]
```

So the clause carries the **depth**: *"read the parent at depth=2 — the field shows the child's FormID."* Without it the remedy would have been another dead end, which is the shape these three sentences exist to remove. Both parentheticals taken: `#350` moved to the clause it explains, and the routing arm now pins all three sentences (plus a second arm pinning that each names `depth=2`).

### F5 — `ReplaceAll composes=` rows

Both added, asserting the owned-child sentence — including `composes=[]`, the modeled-list clear, which is the other door to deleting the child.

### Guard header

One line added naming the table's boundary: the rows pin **leaf** dispositions; the path dimension has its own rows, and the shape that never becomes `substruct` is closed in `SchemaClassifier` and pinned by INV6's SHAPE arms. Your leaf-surface walk is cited there as what established the boundary.

### Verification

- `ci-all` **123/123**; `corpus-hygiene-guard` 20/20; `write-surface-guard` 224/224 (24 disposition rows).
- Both new conditionals sabotage-checked **both** directions, each failing its own rows: path-detection off → the two `CopyFrom`-through rows; widened to every hop → the ordinary-substruct control plus three unrelated `CopyFrom` arms. Predicate narrowed → the polymorphic SHAPE arm; walk gate narrowed → the reach arm.
- One row I got wrong on the first pass (`Landscape.OcclusionData` — no such field) failed loudly and was corrected to a real nullable field rather than dropped.

No round after this fold, per the standing rule. Awaiting your go to merge.

Issue #354, opened 2026-08-17
[GAP] no targeted per-plugin record fetch — GetRecord enumerates a whole overlay per call


Avick3110:
## Goal

Ask "what does plugin P's own body of record R look like" for a handful of plugins, at a cost proportional to
finding one record — not to reading every record P contains.

## What fell short

`LoadOrderResolver.GetRecord` (src/housecarl-core/LoadOrderResolver.cs:790-796) resolves a body by linear scan:

```csharp
foreach (var rec in session.Overlay(idx).EnumerateMajorRecords())
    if (rec.FormKey == fk) return rec;
return null;
```

The cost is already documented 30 lines below it, at :826-828, as the reason the cross-query scan primitives were
built the way they are:

> Each is a SINGLE enumeration pass that yields the matching record's body IN HAND (no per-candidate re-fetch —
> the naive "get each winner body separately" was measured at ~100 s over 9k weapons because GetRecord
> re-enumerates a whole overlay per call).

So any feature that needs N plugins' versions of ONE record pays N whole-overlay enumerations. Measured on a
synthetic order (one 20k-record master + nine 2k plugins, 10 touchers, ~38k records among the others), the same
code path with and without such a walk:

| | without | with |
|---|---|---|
| `read_record` whole record | 7.2 ms | 109.3 ms |
| `read_record fields=[Temporary]` | 5.8 ms | 95.6 ms |

That is ~2.7 µs per record enumerated, and it scales with the record count of the plugins involved, not with the
number of records wanted. On a real load order the same shape measured 27 ms → 588 ms for one Dawnstar cell
(42 touchers, `Skyrim.esm` among them) and 21 ms → 1.3 s for `Tamriel`.

## Workaround used

#342 stage 1 wanted exactly this — each touching plugin's own version of one cell — and could not have it.
It ships instead as two tiers: an index-only statement on every read (no bodies at all), and the precise answer
only on `conflict_tree=true`, which has already paid for a full tree fetch and hands those same bodies to the
annotation. That is a real answer, not a placeholder, but the shape of it was chosen by this cost.

`ConflictTree` itself (`LoadOrderResolver.ResolveTree`) has the same structure: it calls `FetchBody` per touching
plugin, each a full enumeration.

## Ideal shape

A per-(plugin, FormKey) targeted fetch — a typed group seek, a per-overlay FormKey index built on first use, or
whatever Mutagen supports — so "P's version of R" costs about what finding one record should. Whatever the
[...]

Issue #353, opened 2026-08-17
[GAP] a read served from a FAILURE-degraded order looks identical to one served from a legitimately changed order


Avick3110:
## Goal

Read a record and know whether the answer describes the load order the user actually has — as opposed to an
order houseCARL silently narrowed this session because a plugin stopped being openable.

## What fell short

When an active plugin becomes unopenable mid-session (corrupted, truncated, replaced by a partial download,
swapped under the process by a tool that writes non-atomically), the next freshness re-check rebuilds the index
**without it**. The plugin is excluded from the order, and every subsequent read answers normally off the
narrowed order: a shallower `override_depth`, fewer touchers in `conflict_tree`, and — where the missing plugin
was the one carrying content — a materially different answer. Nothing in the read response says a plugin
FAILED. The response is indistinguishable from one served after a legitimate reorder or a disabled mod.

Measured on a synthetic MO2 instance (two plugins, both declaring the same cell; index built off intact files,
then the declaring master overwritten with 4 bytes of garbage):

```
BEFORE: plugins=2 records=4 epoch=b53a46deee41a0b2 loadFailures=[]
  type=Cell  formid=000C01:HcUrBase.esm  winner=HcUrTop.esp  override_depth=2

--- the declaring plugin is corrupted, index unchanged so far ---

AFTER:  plugins=2 records=1 epoch=7af654ddc8af948e
  loadFailures=[HcUrBase.esm: could not be opened - MalformedDataException:
                Could not read enough data to parse a Mod Header from stream. Position: 0. 4 remaining < 24 expected.]
  type=Cell  formid=000C01:HcUrBase.esm  winner=HcUrTop.esp  override_depth=1
```

The second read is clean, confident, and describes an order the user does not have. `override_depth` went
2 -> 1 and the response says nothing about why.

## What already exists (so the gap is precise)

The signal is not missing from the process — it is missing from the **read response**:

- `housecarl_load_order_status` reports the failure verbatim in `loadFailures`.
- The epoch fingerprint changes, so two reads either side of the failure are visibly from different builds — but
  an epoch change is also what a legitimate reorder produces, so it does not distinguish the two.
- A read that names the excluded plugin explicitly (`plugin=<the failed one>`) does refuse and say it was
  excluded. Only reads that do not name it are silent.

So a caller learns about the degradation only by separately polling a different tool, and only if they thought
to. Within a single read response the failure-degraded order and a legitimately changed order are the same shape.

## Ideal shape

[...]

PR #355, opened 2026-08-17
Owned-child cell annotation: say when a winner's child list is not the whole story (#342 stage 1)

Avick3110:
Part of #342 (stage 1). **No closing keyword** — stage 2 (the FormKey-keyed union read) stays routed to the W5 charter, and #342 is its tracker.

## The wrong answer

Placed references, a topic's INFO lines and a worldspace's cells are declared **per plugin**, and the game assembles a parent's children from every plugin that declares them. A plugin that overrides a cell for an unrelated reason — occlusion, lighting, music — carries no references and deletes none, so reading that winner's `Persistent`/`Temporary` reported an empty cell the game fills. Reproduced to the digit on a real order: `008EB5:Skyrim.esm` reads Temporary 0 at its winner (override depth 42) while `Skyrim.esm`'s own body carries 201. Anyone auditing "what is in this cell" through the winner got a confident wrong answer — the Q3 class.

## What ships

**Two tiers.** Every load-order read of a field that owns child records states, from the INDEX alone, that other plugins touch this record and that this read did not open them. `conflict_tree=true` — which has already fetched every touching body for its diff — additionally names *which* of them declare, computed off those same bodies, and now hands that tree to the diff block instead of fetching it twice.

**Two sentence arms, by field shape.** A COLLECTION child (`Persistent`, `Temporary`, `NavigationMeshes`, `Responses`, `SubCells`) is assembled additively. A SINGULAR one (`Cell.Landscape`, `Worldspace.TopCell`) is ONE record that plugins override — "not the merged total" is false there, so it says the child is resolved by load order and that this body not carrying it does not remove it, with a COUNT of carriers rather than a name flood (`TopCell` has 484 carriers on the real order).

**Field set by construction.** `WriteEngine.ChildBearingProperties` over the concrete record types — the write surface's pinned authority, reached from the getter through `PrimaryGetter`/`ConcreteOf`. Never a hand list.

**"Declares" means reaches a child record**, not "has a top-level element": a worldspace carrying empty block scaffolding declares no cells (measured — 2 blocks / 2 sub-blocks / 0 cells exposes 2 elements and enumerates 0 children).

Display-only throughout: the round-trip token is untouched. Artifacts carry the note on rows and the sentence on the manifest line, so a catalogue re-opened later still says what it means.

## Decision record — two escalations, both Aaron-ruled

**1. The trigger.** Stage 1 was chartered as "fires when a lower layer declares content the winner's list is empty or smaller than". Review showed the count comparison produced two wrong answers: `Worldspace.SubCells` counts BLOCKS (cells sit two container levels down), so a base declaring 1 block of 5 cells "lost" to an override declaring 2 blocks of 1 each — the winner went unannotated and the base named the smaller declarer as largest; and disjoint declarations went unreported entirely (base declares 3, winner declares 4 *different* ones, live set 7, read shows 4 in silence). Escalated under §4(b) rather than folded. **Ruled: the count comparison goes, the trigger becomes a boolean.** The correction that came back — "declares" must mean *reaches a child record*, not *has an element* — was verified against Mutagen before implementing, and is what the WRLD-SCAFFOLD arms hold.

**2. The heterogeneous field set.** Round 2 found the invariant clause was FALSE for two members of its own field set (`Landscape`, `TopCell` — singular children, not merged). Same class as round 1's SubCells finding, one round later, so it went up under the recurrence rule rather than being folded a third time. **Ruled: per-shape sentences**, on the ground that this is a bounded structural distinction the classifier already answers, not the per-field count patchwork previously rejected. A third instance of this class is a standing §4 escalation.

## Performance — the correction chain, stated plainly

The first measurement was **wrong, and the ruling made on it was corrected**:

1. Synthetic fixture, ~200-record plugins: +43.5 ms on a 42-toucher cell read. Ruled "not hot, don't gate".
2. Same fixture re-measured on the product path (with `corpus.json` present): the single-read figure was **3x low**. Self-flagged.
3. Realistic plugin sizes (20k-record master + nine 2k plugins): **7.2 ms → 109.3 ms**, 15x. `GetRecord` fetches a body by enumerating a whole overlay — a cost this repo already documents at `LoadOrderResolver.cs:828` (*"~100 s over 9k weapons"*), and which my fixture had almost none of.
4. Review measured the real order: **27 ms → 588 ms** (Dawnstar), **21 ms → 1.3 s** (Tamriel), **18 ms → 2.5 s** (WhiterunWorld), 6.3 s → 126 s for a 200-cell query, and a `to_file` catalogue that did not finish in 10 minutes.

Escalated; the pre-ruled fallback fired and the **two-tier split** was ruled in (a middle option — full naming on single reads, cheap tier only on bulk — was weighed and declined, because per-read seconds re-compound across a session). Post-split, paired on the real order (3800 plugins, 3.2M records):

| | baseline | branch |
|---|---|---|
| `cross_plugin_query` CELL conflicts_only `fields=[Temporary]` limit=200 | 5941 / 5601 ms | 5968 / 5613 ms |
| `read_record` Dawnstar (42 touchers), x5 | 111 ms | 95 ms |
| `read_record` Dawnstar, independent reviewer's A/B | 26 ms | 27 ms |
| `read_record` uncontested cell + `conflict_tree=true` | 133 ms | 273 ms → **fixed in round 3** |
| `read_record` Dawnstar + `conflict_tree=true` (42 touchers) | 899 ms | 849 ms (one fetch, not two) |

The `to_file` cell catalogue now completes: 26,906 rows in 35.8 min. **No delta is claimed for it** — I did not measure a pre-split completion to subtract, and its cost is dominated by the query's own per-row body fetch (#354), not by the annotation.

## Verification limits, stated rather than implied

- **The unreadable-toucher path is not reachable end-to-end.** Corrupting a declaring plugin does not leave a named-but-unfetchable toucher: the freshness re-check rebuilds, the plugin is excluded from the order, and it stops being a toucher at all. The guard pins that measured lifecycle instead, plus the null rule at its two reachable ends. The residual hazard the arm exists for — a plugin that opens at header level but faults walking its child group — is not constructible from outside the process.
- **The precise tier's budget skip is partial.** The prefetch cannot foresee a cap hit *during* the record's own field render, so a single read with a cap tight enough to exhaust itself still pays for a tree it then truncates. I wrote an arm for this, could not make it true, and deleted it rather than keep an unpinnable branch. Stated in the code.
- **`fields=["Temporary[0]"]` is not annotated** — the note lands on the container line. Dropped with grounds: an empty winner errors on the index rather than answering silently, and bracket-prefix matching would annotate every deep sub-path.

## Review rounds

Three rounds, two fresh agents each, own worktrees, seeded with the settled-decisions list only. Both escalations came out of them.

| round | found | outcome |
|---|---|---|
| 1 | SubCells counts blocks; disjoint sets unreported; clause repeated ~275 chars/row eating the bulk budget; two guard branches unpinned | 0 folded — escalated the trigger under §4(b) |
| 2 | Real-order perf collapse (independently, twice); the clause false for singular children; `to_file` artifacts shipped the label with the clause nowhere; stale count-trigger comments | 2 escalations, 4 folds, 2 refused |
| 3 | The prefetch defeating the tree's sole-toucher skip (regression from round 2's fold, both agents, with A/B numbers); json batch clause over unrendered rows (same); remedy naming a parameter the recipient surface refuses; "every read" overclaim; manifest notes not round-tripping | 6 folds, 1 dropped with grounds, 1 arm deleted as unsatisfiable |

**Two of round 3's findings were regressions introduced by round 2's fold.** Recorded because it is the honest signal about folding under pressure, not buried.

**Refused, with grounds:** annotating the out-of-load-order read path (stamped OUT-OF-LOAD-ORDER; a load-order statement does not belong there) and the SkyPatcher-overlay pole (unclaimed by the CHANGELOG). Both remain unannotated deliberately, and the CHANGELOG now says so instead of claiming "every read".

**Reviewers self-refused** several suspicions across rounds — `Display` collision, round-trip contamination, dense-cell loss, case-variant field paths, a stranded-clause path — so the triage was not one-directional.

## Guard

`owned-child-content-guard`, **48 arms**, in `ci-all` (124/124). RED-checked in both directions: collapsing the shape classifier reds 3; dropping the containment walk for an element test reds 2; the getter→concrete hop reds the Landscape arm; the null rule reds the unreadable arm; the artifact clause reds 1; a precise tier that names nobody reds 11; an unconditional response clause reds 2. A type-level arm pins the getter→concrete hop over **every** concrete child-bearing type, not a sample.

## Spawned, not fixed here

- #353 — a read served from a FAILURE-degraded order carries no in-band signal (found while fixturing the unreadable path).
- #354 — no targeted per-plugin record fetch; `GetRecord` enumerates a whole overlay per call. This is what forced the tier split, and closing it would let the precise tier run everywhere.

Both filed unrouted; routing is Aaron's at RUN_ORDER.

### DIFF
--- a/src/housecarl-mcp/Artifacts.cs
+++ b/src/housecarl-mcp/Artifacts.cs
@@ -136,6 +136,7 @@ internal static class Artifacts
         string[] schema;
         string? identity;
         string sort;
+        bool annotated = false;   // #342: did any ROW carry the owned-child annotation
 
         if (q.Groups is not null)                                             // group_by= → count-table rows
         {
@@ -157,6 +158,7 @@ internal static class Artifacts
                 string? matches = q.MatchedTargets is { } mt && i < mt.Count ? mt[i] : null;
                 var o = svc.ResolveReadOn(q, fk, winnerFields ? null : (q.Sources is { } src ? src[i] : null), fields, false, depth,
                                           resolveNames: resolveNames, linkMemo: linkMemo);
+                annotated |= o.OwnedChildNoted;   // #342: the rows' labels need their clause on line 1
                 if (o.Error is not null)
                     writer.WriteRow((w, _) =>
                     {
@@ -183,9 +185,21 @@ internal static class Artifacts
         }
 
         var (manifest, err) = writer.Save(path, "housecarl_cross_plugin_query", query, identity, schema, sort,
-                                          q.Groups is not null ? q.Groups.Count : q.Total, q.Epoch ?? "");
+                                          q.Groups is not null ? q.Groups.Count : q.Total, q.Epoch ?? "",
+                                          OwnedChildNotes(annotated));
         return err is not null ? (null, err) : (new SpillInfo(path, manifest!, reason), null);
     }
+
+    /// <summary>The response-level #342 statements an artifact's ROWS depend on. An artifact is re-entered with no
+    /// conversation attached, so a row's "N other plugins touch this record; their declarations were not read"
+    /// label has to travel with the sentence that says what a child record is and where the precise answer lives.
+    /// Only the CHEAP tier ever reaches a file: the artifact lanes read with conflict_tree off (it is a text-only
+    /// diff view), so a row can only ever carry the cheap note.</summary>
+    static IReadOnlyList<string>? OwnedChildNotes(bool annotated) =>
+        annotated ? new[] { ReadSentences.NotReadClause } : null;
+
+    static IReadOnlyList<string>? OwnedChildNotes(IReadOnlyList<ReadOutcome> outcomes) =>
+        OwnedChildNotes(outcomes.Any(o => o.OwnedChildNoted));
 
     /// <summary>Build + save the artifact for a batch_record_detail result — one row per input, in input order,
     /// exactly the rows the json render emits (per-item errors included: the artifact is the complete answer, and
@@ -206,7 +220,7 @@ internal static class Artifacts
         var epoch = outcomes.FirstOrDefault(o => o.Epoch is not null)?.Epoch ?? "";
         var (manifest, err) = writer.Save(path, "housecarl_batch_record_detail", query, "formid",
                                           new[] { "formid", "type", "editorid", "winner", "override_depth", "source", "fields" },
-                                          "input order", outcomes.Count, epoch);
+                                          "input order", outcomes.Count, epoch, OwnedChildNotes(outcomes));
         return err is not null ? (null, err) : (new SpillInfo(path, manifest!, reason), null);
     }
 
--- a/src/housecarl-mcp/Artifacts.cs
+++ b/src/housecarl-mcp/Artifacts.cs
@@ -136,7 +136,7 @@ internal static class Artifacts
         string[] schema;
         string? identity;
         string sort;
-        bool annotated = false;   // #342: did any ROW carry the owned-child annotation
+        var annotated = new SortedSet<string>(StringComparer.Ordinal);   // #342: which fields the ROWS carry annotated
 
         if (q.Groups is not null)                                             // group_by= → count-table rows
         {
@@ -158,7 +158,8 @@ internal static class Artifacts
                 string? matches = q.MatchedTargets is { } mt && i < mt.Count ? mt[i] : null;
                 var o = svc.ResolveReadOn(q, fk, winnerFields ? null : (q.Sources is { } src ? src[i] : null), fields, false, depth,
                                           resolveNames: resolveNames, linkMemo: linkMemo);
-                annotated |= o.OwnedChildNoted;   // #342: the rows' labels need their clause on line 1
+                if (o.Error is null && o.OwnedChildFields is { } af)   // #342: the rows' labels need their clause on line 1
+                    foreach (var annotatedPath in af.Keys) annotated.Add(annotatedPath);
                 if (o.Error is not null)
                     writer.WriteRow((w, _) =>
                     {
@@ -194,12 +195,25 @@ internal static class Artifacts
     /// conversation attached, so a row's "N other plugins touch this record; their declarations were not read"
     /// label has to travel with the sentence that says what a child record is and where the precise answer lives.
     /// Only the CHEAP tier ever reaches a file: the artifact lanes read with conflict_tree off (it is a text-only
-    /// diff view), so a row can only ever carry the cheap note.</summary>
-    static IReadOnlyList<string>? OwnedChildNotes(bool annotated) =>
-        annotated ? new[] { ReadSentences.NotReadClause } : null;
+    /// diff view), so a row can only ever carry the cheap note.
+    ///
+    /// <para>The manifest is LINE 1 and the annotated rows are lines 2..N, so the clause's old "an annotated field
+    /// above" pointed the wrong way for exactly the reader it was added for (Aaron's finding 2). It now names the
+    /// annotated fields instead of pointing at them, which is true from line 1 and from anywhere else.</para></summary>
+    static IReadOnlyList<string>? OwnedChildNotes(IReadOnlyCollection<string> annotatedFields) =>
+        annotatedFields.Count == 0 ? null : new[] { ReadSentences.NotReadClause(annotatedFields) };
 
-    static IReadOnlyList<string>? OwnedChildNotes(IReadOnlyList<ReadOutcome> outcomes) =>
-        OwnedChildNotes(outcomes.Any(o => o.OwnedChildNoted));
+    /// <summary>The annotated field paths an artifact's rows CARRY. Rows are written uncapped (the file is the
+    /// answer), so every annotated field of every row reaches the file — but the set is collected from the rows all
+    /// the same, so the manifest cannot state a clause over an annotation no row wrote.</summary>
+    static SortedSet<string> AnnotatedFields(IEnumerable<ReadOutcome> outcomes)
+    {
+        var s = new SortedSet<string>(StringComparer.Ordinal);
+        foreach (var o in outcomes)
+            if (o.Error is null && o.OwnedChildFields is { } f)
+                foreach (var path in f.Keys) s.Add(path);
+        return s;
+    }
 
     /// <summary>Build + save the artifact for a batch_record_detail result — one row per input, in input order,
     /// exactly the rows the json render emits (per-item errors included: the artifact is the complete answer, and
@@ -220,7 +234,7 @@ internal static class Artifacts
         var epoch = outcomes.FirstOrDefault(o => o.Epoch is not null)?.Epoch ?? "";
         var (manifest, err) = writer.Save(path, "housecarl_batch_record_detail", query, "formid",
                                           new[] { "formid", "type", "editorid", "winner", "override_depth", "source", "fields" },
-                                          "input order", outcomes.Count, epoch, OwnedChildNotes(outcomes));
+                                          "input order", outcomes.Count, epoch, OwnedChildNotes(AnnotatedFields(outcomes)));
         return err is not null ? (null, err) : (new SpillInfo(path, manifest!, reason), null);
     }
 
--- a/src/housecarl-mcp/JsonWire.cs
+++ b/src/housecarl-mcp/JsonWire.cs
@@ -342,12 +342,16 @@ static class JsonWire
         return Finish(ms);
     }
 
-    /// <summary>The #342 invariant clause on the json lane, written ONCE per response when any rendered record
-    /// carries a per-field declarer annotation — the same const the text lane states, so the two transports
-    /// cannot drift on what a child record is. Gated on the outcome's structural flag, never on the prose.</summary>
+    /// <summary>The #342 clause on the json lane, written ONCE per response when a rendered record carries the
+    /// annotation — the same const the text lane states, so the two transports cannot drift. Gated on the
+    /// outcome's structural flag, never on the prose.
+    ///
+    /// <para>json only ever states the CHEAP tier's clause: <c>conflict_tree=true</c> is refused in json mode (a
+    /// text-only diff view), so the lane that has the bodies to name declarers does not exist here. A json caller
+    /// who wants the precise answer takes the same route the clause names — the text lane.</para></summary>
     static void WriteOwnedChildNote(Utf8JsonWriter w, bool noted)
     {
-        if (noted) w.WriteString("owned_child_note", ReadSentences.OwnedChildMerge);
+        if (noted) w.WriteString("owned_child_note", ReadSentences.NotReadClause);
     }
 
     // ---- housecarl_batch_record_detail (P6) ---------------------------------------------------------
--- a/src/housecarl-mcp/JsonWire.cs
+++ b/src/housecarl-mcp/JsonWire.cs
@@ -261,7 +261,11 @@ static class JsonWire
     /// BUDGET-AWARE: a fat record (deep list expansion) is field-truncated the same way the text render caps field
     /// lines — a sentinel field names the cut and the array closes, so the document stays valid JSON (never silently
     /// over budget — Q3).</summary>
-    static void WriteFieldsArray(Utf8JsonWriter w, RecordFields r, MemoryStream ms, int cap)
+    /// <param name="annotated">The #342 owned-child fields this outcome annotated, if any.</param>
+    /// <param name="emitted">Collects the annotated paths this array ACTUALLY carried — the response-level clause is
+    /// stated over these, so a field the truncation above dropped states nothing (Aaron's finding 1, json twin).</param>
+    static void WriteFieldsArray(Utf8JsonWriter w, RecordFields r, MemoryStream ms, int cap,
+                                 IReadOnlyDictionary<string, OwnedChildShape>? annotated = null, ICollection<string>? emitted = null)
     {
         w.WriteStartArray("fields");
         for (int i = 0; i < r.Fields.Count; i++)
@@ -294,6 +298,7 @@ static class JsonWire
                 w.WriteEndObject();
             }
             w.WriteEndObject();
+            if (annotated is not null && emitted is not null && annotated.ContainsKey(f.Path)) emitted.Add(f.Path);
         }
         w.WriteEndArray();
     }
@@ -301,15 +306,18 @@ static class JsonWire
     /// <summary>Serialize a resolved record: identity + winner/override_depth/source + the fields array. Shared by
     /// read_record, batch_record_detail, and the cross_plugin_query detail path (one shape, no drift). <paramref
     /// name="matches"/> carries the multi-target references= un-merge when present.</summary>
+    /// <param name="childFields">Collects the #342-annotated field paths this record's array carried, for the
+    /// response-level clause. Batch/query lanes pass ONE set across their rows and state the clause after the
+    /// records array; the single read passes its own and states it here.</param>
+    /// <param name="stateChildNote">Single-read lane only: this record object IS the response, so the clause
+    /// belongs on it — written AFTER <c>fields</c> and only over what <c>fields</c> carried. It used to be the
+    /// object's second key, ahead of the very array it described (Aaron's finding 3).</param>
     internal static void WriteReadRecord(Utf8JsonWriter w, ReadOutcome o, MemoryStream ms, int cap, string? matches = null,
-                                         string? epoch = null, bool ownedChildNote = false)
+                                         string? epoch = null, ICollection<string>? childFields = null, bool stateChildNote = false)
     {
         var r = o.Record!;
         w.WriteStartObject();
         if (epoch is not null) w.WriteString("epoch", epoch);   // single-read top level ONLY — see RenderRecord
-        // The single-read record object IS the response, so the #342 clause belongs on it; batch/query rows never
-        // repeat it per row (the caller passes false and the response object states it once), exactly as epoch does.
-        WriteOwnedChildNote(w, ownedChildNote);
         w.WriteString("formid", r.FormKey);
         w.WriteString("type", r.Type);
         WriteNullable(w, "editorid", r.EditorId);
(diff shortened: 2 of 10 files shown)


src/housecarl-mcp/ReadTools.cs
```diff
@@ -780,11 +780,77 @@ public static string RenderRecord(LoadOrderService svc, ReadOutcome o, IReadOnly
...
+        // render. A single read with a cap so tight that the fields themselves exhaust it still pays for the tree
+        // and then truncates the diff — base skipped that fetch. Closing it would mean rendering the record twice
+        // or guessing its rendered size, and the annotation has to land BEFORE the fields are written.
+        if (conflictTree && o.Pin is { } pin && o.Record is { } rec
+            && o.TouchingPlugins is { Count: > 1 } && sb.Length < cap)
+        {
+            fill = svc.ResolveTreeFill(pin, o.FormKey, fields, o.SourcePlugin, rec.Fields.Select(f => f.Path).ToList());
+            if (fill is { ByField.Count: > 0 }) outcome = ApplyPreciseChildNotes(o, rec, fill, notes);
+        }
+        if (outcome.OwnedChildNoted && fill is null) notes.NotRead = true;   // cheap tier stands
```
Avick3110: The response clause is committed here, **before** the record's own fields are rendered — so a cap hit *inside* the field loop leaves the clause pointing at an annotated field that was truncated away. Round 3 closed this at row granularity; the field granularity is still open.

`housecarl_read_record formid="008EB5:Skyrim.esm" max_chars=300`: `notes.NotRead = true` is set on this line, then `AppendRecord(sb, outcome, cap)` (833) stops the field loop at `sb.Length >= cap` before reaching `Temporary`. `AppendOwnedChildNotes` then appends *"note: an annotated field above holds CHILD RECORDS…"* over a body that contains only `... [truncated: showing 4 of 61 field lines]`.

The json twin is the same shape: `WriteReadRecord` writes `owned_child_note` (JsonWire.cs:312) and `WriteFieldsArray` truncates the array (JsonWire.cs:270). Both transports state a clause about a field the caller cannot see.

src/housecarl-mcp/Artifacts.cs
```diff
@@ -183,10 +185,22 @@ public static (SpillInfo? Spill, string? Error) WriteCrossQuery(
...
         return err is not null ? (null, err) : (new SpillInfo(path, manifest!, reason), null);
     }
 
+    /// <summary>The response-level #342 statements an artifact's ROWS depend on. An artifact is re-entered with no
+    /// conversation attached, so a row's "N other plugins touch this record; their declarations were not read"
+    /// label has to travel with the sentence that says what a child record is and where the precise answer lives.
+    /// Only the CHEAP tier ever reaches a file: the artifact lanes read with conflict_tree off (it is a text-only
+    /// diff view), so a row can only ever carry the cheap note.</summary>
+    static IReadOnlyList<string>? OwnedChildNotes(bool annotated) =>
+        annotated ? new[] { ReadSentences.NotReadClause } : null;
```
Avick3110: `NotReadClause` is reused verbatim as the manifest note, but the manifest is **line 1** of the JSONL and the annotated rows follow it — so "an annotated field above" points the wrong direction for exactly the reader this note was added for (an artifact re-entered with no conversation attached).

`housecarl_batch_record_detail formids=[…cells…] to_file="C:\out.jsonl"` → line 1 is `{"tool":…,"notes":["note: an annotated field above holds CHILD RECORDS…"]}`, and every annotated `Temporary`/`Persistent` field is on lines 2..N. `AppendSpillText` even tells the caller "line 1 = this manifest, one row per line", so the direction is unambiguous and the sentence is backwards.

The clause was written for the text lane (where it is appended after the body) and inherited here unchanged. A file-lane variant — or dropping the positional word — fixes it.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -301,11 +301,15 @@ static void WriteFieldsArray(Utf8JsonWriter w, RecordFields r, MemoryStream ms,
...
-    internal static void WriteReadRecord(Utf8JsonWriter w, ReadOutcome o, MemoryStream ms, int cap, string? matches = null, string? epoch = null)
+    internal static void WriteReadRecord(Utf8JsonWriter w, ReadOutcome o, MemoryStream ms, int cap, string? matches = null,
+                                         string? epoch = null, bool ownedChildNote = false)
     {
         var r = o.Record!;
         w.WriteStartObject();
         if (epoch is not null) w.WriteString("epoch", epoch);   // single-read top level ONLY — see RenderRecord
+        // The single-read record object IS the response, so the #342 clause belongs on it; batch/query rows never
+        // repeat it per row (the caller passes false and the response object states it once), exactly as epoch does.
+        WriteOwnedChildNote(w, ownedChildNote);
```
Avick3110: Same class as the artifact-manifest note, one transport over: on the single-read json lane the clause is written as the **second key** of the record object, ahead of `formid` and `fields`, so "an annotated field above" is inverted here too.

`housecarl_read_record formid="008EB5:Skyrim.esm" format="json"` →
```json
{"epoch":"…","owned_child_note":"note: an annotated field above…","formid":"008EB5:Skyrim.esm","fields":[…,{"path":"Temporary","display":"42 other plugin(s) touch this record…"}]}
```
The text lane appends after the body and the json *batch* lane writes after the `records` array — both correct. Only this site and the artifact manifest put the note ahead of what it describes.

src/housecarl-mcp/ReadSentences.cs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// <summary>The response-level fact for COLLECTION children: assembled additively, so one body's list is not
+    /// the total.
+    ///
+    /// <para>Deliberately names NO remedy for the total. The read that would answer "what is actually live in this
+    /// parent" (a FormKey-keyed union with each child at its own winner) does not exist yet, and re-reading with
+    /// <c>plugin=</c> is not load-order truth either — that body's own children can themselves be overridden
+    /// further up.</para></summary>
+    [MustState("declared per plugin", "not the merged total")]
+    internal const string MergeCollection =
```
Avick3110: These two clauses name their fields as **literal prose lists**, while the set they describe is `WriteEngine.ChildBearingProperties` *by construction* — and no probe arm pins the prose against the reflected set. `OwnedChildContent`'s own doc makes the promise this breaks: "a Mutagen bump that grows that set is annotated without an edit here."

Concretely: a Mutagen version adds a collection-shaped owned-child property (say `Quest.SomeChildren`). `OwnedChildContent.Fields` picks it up with no code edit — as designed — and a read annotates it `also declared by X.esp`. The response then states *"an annotated Persistent / Temporary / NavigationMeshes / Responses / SubCells field holds MANY child records"*, naming five fields, none of which is the annotated one. The all-types write-surface pin stays green (the reflected set really is correct), the `[MustState]` net stays green (its phrases are still present), and the prose drifts silently.

`OwnedChildContentProbe` has a `BY CONSTRUCTION` arm for the getter→concrete hop but nothing that asserts the sentence's field names ⊇ the reflected Collection/Singular sets. Either derive the names from the set, or add that arm so a Mutagen bump goes red here rather than shipping a clause that describes the wrong fields.

src/housecarl-mcp/ReadSentences.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// <summary>The cheap tier's per-field line: the count the index knows, and the honest limit.</summary>
+    internal static string NotReadNote(int others) => $"{others} {NotRead}";
+
+    // ---- tier 2: the precise answer, off bodies the conflict-tree lane already fetched ----------------
+
+    /// <summary>The per-field label for a COLLECTION child. A pure label: on its own it asserts nothing a caller
+    /// acts on — the claim is the plugin names it introduces, and their meaning is
+    /// <see cref="MergeCollection"/>.</summary>
+    [NoClaims("a label; the claim is the plugin names it introduces, and their meaning is MergeCollection")]
+    internal const string DeclaredBy = "also declared by";
```
Avick3110: Both labels open with "also", which asserts the body being read declares content **too** — false in exactly the case this feature exists for.

`housecarl_read_record formid="008EB5:Skyrim.esm" conflict_tree=true`, winner `Occlusion.esp` carrying 0 Temporary references, renders:

```
Temporary = [list: 0 item(s)]   (also declared by Skyrim.esm, …)
```

"also declared by" next to a zero-count list reads as "this body declares some, and so do these" — the confident wrong reading the annotation was built to stop. The `SELF` arm gets this right by leaving a sole declarer unannotated; the flagship case is where the word misfires.

Per CLAUDE.md §5 #11 this is a sentence finding, so it wants a probe rather than a reword — but "declared by" / "carried by" without the leading "also" is true in both directions (this body declaring or not), where "also" is only true in one.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -780,11 +780,77 @@ public static string RenderRecord(LoadOrderService svc, ReadOutcome o, IReadOnly
...
+    /// them.</summary>
+    internal sealed class ChildNotes
+    {
+        public bool NotRead, Collection, Singular;
+        public bool Any => NotRead || Collection || Singular;
+    }
+
+    /// <summary>The #342 clauses, stated ONCE per response after the body — never per field, which cost ~275
+    /// identical chars on every annotated row and pushed real rows out of a bulk response's budget.</summary>
+    internal static void AppendOwnedChildNotes(StringBuilder sb, ChildNotes n)
```
Avick3110, line 837: `AppendOwnedChildNotes` runs after the cap-bounded body and appends up to three clauses with no `cap` check, so an annotated response overruns the caller's `max_chars`. Every other cut in these renders is budget-bounded with an explicit notice.

`housecarl_batch_record_detail formids=[…cells…] conflict_tree=true max_chars=2000`: the row loop stops at 2000 chars and emits its `... [truncated: rendered X of N]` notice, then this method adds `NotReadClause` (~430 chars) + `MergeCollection` (~350) + `SingleResolved` (~330). The returned string is ~3.1k against a declared 2k ceiling, and the overrun is invisible to the `truncated` flag the auto-spill trigger reads.

The clauses are load-bearing so dropping them at the cap is wrong; reserving their length out of `cap` before the body renders (or subtracting it in `Cap()` when the lane can annotate) keeps the ceiling honest.

Avick3110:
Review triaged: **all six findings confirmed against branch source, all six accepted, none refused.** The ground for zero refusals is that every finding cites a line and every one reproduces by inspection at that line — I wrote each of these defects and can trace them without ambiguity.

Findings 1–3 are one class (the clause makes a positional claim *and* a presence claim, and each medium falsifies one) — including at a site round 3's fold summary claimed was already closed. That incompleteness gets owned in the fold summary rather than quietly fixed.

**The fold is executing in a successor session, not this one.** At this session's context depth a six-finding fold across four files is where fold-introduced regressions happen, and this branch is its own case study: round 2's directed fold introduced two regressions that both round-3 reviewers found independently, and that fold was made when the session was fresher than it is now. Handing it over is the lower-risk call, not a stall.

Branch is untouched and clean at `76e2620` — `git status --porcelain` empty, no probe residue, `ci-all` 124/124, `owned-child-content-guard` 48/48. Nothing has been folded here.

Per §11 no new review round follows the directed fold; the fold summary will be posted here before the merge gate.

Avick3110:
## Fold summary — all six review findings folded, none refused

One commit, `97a18db`. `ci-all` 124/124; `owned-child-content-guard` **48 → 59 arms**, all green, and each new arm RED-checked with its own sabotage (listed at the bottom). **No new review round follows this directed fold** (CLAUDE.md §5 #11).

### Owning the round-3 incompleteness

Round 3's fold summary said "the clause no longer claims a position". That was true of the text lane's rows and false everywhere else — findings 1, 2 and 3 are three sites where it still claimed one, or where the presence it assumed did not hold. The clause was making **two** claims each medium had to satisfy independently: a POSITIONAL one ("an annotated field above") and a PRESENCE one (that such a field is in the medium at all). Round 3 fixed the presence half at ROW granularity on two lanes and left the positional half standing everywhere. Fixing three sites would have left the fourth medium to falsify it again, so this fold fixes the class.

### Per-finding disposition

| # | Finding | Disposition |
|---|---|---|
| 1 | `ReadTools.cs:832` — clause committed before the fields render; a cap hit inside the field loop strands it (json twin the same) | **Folded, class fix (b).** The clause is registered at EMISSION — as each annotated field line goes into the medium — not where the annotation was decided. `ReadOutcome` now carries the annotated paths and their shapes (`OwnedChildFields`) instead of a bool, because a flag that only knows "something was annotated" cannot tell whether it survived the medium's own cut. |
| 2 | `Artifacts.cs:199` — the manifest is line 1, the rows are 2..N, so "above" points the wrong way | **Folded, class fix (a).** The clause names its fields; there is no direction left to get wrong. |
| 3 | `JsonWire.cs:312` — the note is the record object's second key, ahead of its own `fields` array | **Folded, (a) + (b).** Positional word gone, and the write moved to after `WriteFieldsArray`, gated on the paths that array actually carried. |
| 4 | `ReadSentences.cs:92` — the field lists are literal prose against a by-construction set | **Folded by DERIVATION**, not by a pin. The names come from what the response annotated — `OwnedChildContent.Fields`, i.e. `WriteEngine.ChildBearingProperties` split by shape. **No arm compares the sentence to the set it was derived from** (the echo chamber #349 rejected); the existing all-types pin covers the authority. This is the same mechanism as (a), so one fix serves both — and it is sharper than naming the model-wide set would be: the clause names *the annotated field*, which is precisely what your `Quest.SomeChildren` walkthrough showed the old prose getting wrong. |
| 5 | `ReadSentences.cs:70` — "also declared by" asserts this body declares too | **Folded**, per your own analysis, with a probe rather than a reword. Before/after below. |
| 6 | `ReadTools.cs:801` — the clauses append past `max_chars` | **Folded by RESERVATION**, your named fix. Worst-case clause length is held back from the budget before the body renders. Held back from the BUDGET only: every cut still quotes the caller's own `max_chars`, because a notice reading "at max_chars=1124" when 2000 was passed would be a wrong sentence of its own. |

**Zero refused, and the ground is unchanged from triage:** every finding cites a line, and every one reproduced by inspection at that line before anything was edited.

### Finding 5 — the measurement, before and after

The guard's flagship fixture is your case exactly: winner `HcOcTop.esp` carrying 0 `Temporary`, base `HcOcBase.esm` carrying 3.

```
before:  Temporary = [list: 0 item(s)] — pass depth=2 to expand   (also declared by HcOcBase.esm)
after:   Temporary = [list: 0 item(s)] — pass depth=2 to expand   (declared by HcOcBase.esm)
```

`CarriedBy` loses its "also" for the same reason — the singular flagship is a body carrying nothing — and so does `SingleResolvedFraming`'s own "Plugins that also carry it", which contradicted its own next clause ("this body not carrying it does not remove it").

### What the clauses read like now

```
note: this response annotates field(s) that hold CHILD RECORDS (Landscape, NavigationMeshes, Persistent,
Temporary). Child records are declared per plugin — so what one plugin's body carries is not the whole story
for that field. …

note: this response annotates field(s) that hold MANY child records (Persistent, Temporary) — a cell's placed
references, a topic's INFO lines, a worldspace's cells. …

note: this response annotates field(s) that hold ONE child record with its own FormID (Landscape). The plugins
that carry it are overriding each other … and the count in that annotation is how many other plugins carry
one, not a total.
```

The `[MustState]` phrases all stay in the static framing, which is still a `const` the sentence net reads.

### The two conditions are independent, and pinned separately

Reserve decides a clause **fits**; emission decides it is **stated**. They compose: at a cap so tight the reservation exceeds it, no annotated field renders, so no clause is stated and nothing overruns.

### Arms added (11), each RED-checked

- **EMISSION (text), both ways** — a cap that truncates the annotated field states no clause; the same read with room states it.
- **EMISSION (json single)** — a `fields` array truncated before the annotated field states no `owned_child_note`.
- **JSON ordering** — the note is written after the array it describes.
- **CLAUSE-NAMES ×2** — the clause names its fields and contains no positional word; every field it names is one the response annotated.
- **SINGULAR / COLLECTION** — each clause names exactly the fields of its shape (`Landscape`; `Persistent, Temporary`).
- **LABEL** — on a winner carrying 0 items the declarer label does not assert this body declares too.
- **ARTIFACT manifest** — the line-1 clause names the annotated fields and claims no position.
- **RESERVE ×3** — an annotated response answers inside its `max_chars` on the tight cheap lane; the cut still quotes the caller's own `max_chars`; and on your `conflict_tree` repro the stated clauses fit the chars reserved for them.

**Sabotages run, and what each reddened:** clause committed at decision time → EMISSION (text); json note over fields the array did not carry → EMISSION (json single); `ClauseReserve` → 0 → all three RESERVE arms; `DeclaredBy` back to "also declared by" → LABEL; the derived list replaced by the old hand-written prose list → CLAUSE-NAMES derived + both shape arms; the json note written ahead of `fields` again → JSON ordering + EMISSION (json single); the positional wording restored on the manifest → both ARTIFACT arms.

### Two notes on what this fold did NOT do

- **The `conflict_tree` lane's total length is not claimed.** Its diff block cuts per NODE rather than per line, so its pre-existing slack is wider than one line; a total-length arm there would be measuring the diff, not this fix. What is pinned there is that the clauses fit their reservation. The tight-ceiling claim is pinned on the single read, where every cut is checked before a line is appended.
- **The stale residual is gone from the code comment.** Your finding 1 was the field-granularity half of the "precise tier's budget skip is partial" limit. The prefetch still cannot foresee a cap hit during the field render — that part stands and still says so — but it no longer costs a wrong sentence, and the comment now says that instead of the old wording.

**One commit rather than several.** Findings 1–4 and 6 are mutually dependent (the derived list is what bounds the clause, which is what makes the reservation a constant), and separating finding 5 would have meant hand-reverting and re-applying edits at exactly the point in this branch's history where folds have introduced regressions. The commit message enumerates the six.

The two standing refusals survive untouched: the out-of-load-order read path and the SkyPatcher-overlay pole stay unannotated, and the CHANGELOG still says "a load-order read". The CHANGELOG needed no edit — it describes the behaviour, not the clause wording, and none of its sentences became false.

Issue #357, opened 2026-08-17
[GAP] the file-placing lanes have no discoverable way to name the patch folder they create


Avick3110:
## Summary

#343 gave the record-write lanes a refusal remedy naming `patch=`, the parameter that names a new patch. The **rider** lanes — the tools that place a file into a patch mod folder (`place_asset` / `bulk_place_asset`, `compile_script`, `decompile_script`, `bsa_repack`, `nif_set`, `write_seq`) — deliberately did not get it, and so still have the original #343 problem: a caller who guesses `into=` for a folder that does not exist is told only to omit it, and lands in a generically-named folder with no hint that naming it was possible.

## Why the #343 remedy could not simply be extended

Two independent grounds, both measured rather than assumed:

1. **`bsa_repack` binds `patch=` elsewhere.** It declares both `archive_name` and `patch_name`, and the §5.3 alias candidate order routes a bare `patch=` to `archive_name` — the `.bsa` filename, not the mod folder. Measured through the real alias layer:

   ```
   housecarl_bsa_repack      patch= -> archive_name
   housecarl_place_asset     patch= -> patch_name
   housecarl_compile_script  patch= -> patch_name
   housecarl_decompile_script patch= -> patch_name
   housecarl_nif_set         patch= -> patch_name
   ```

   So a shared sentence naming `patch=` would rename that caller's archive and leave the folder defaulted.

2. **The rider defaults are not `Patch`.** Each rider call site passes its own stem to `ResolvePatchModFolder` — `houseCARL_Scripts`, `houseCARL_Assets`, `houseCARL_SEQ`, `houseCARL_Archive`, `houseCARL_NifEdit`. So the "omit it and houseCARL names it `Patch`" half of the record-lane sentence is false on every one of them, independently of the alias question.

The #343 branch pins this boundary with an arm asserting the rider refusal does **not** name `patch=`, so the gap is a recorded decision rather than an oversight.

## What would close it

A rider-lane remedy that names each lane's own naming parameter and its own default stem — which needs the calling lane to supply them, the same shape #343 settled on for the record lanes (each operation states what is true of it, with a safe default). Whether `bsa_repack`'s `patch=` → `archive_name` binding should also be revisited is a separate §5.3 question, not this one.

*Filed by a build session while measuring #343. Unrouted — needs a RUN_ORDER routing stroke.*

Avick3110:
[...]

Issue #356, opened 2026-08-17
[Bug] housecarl_remove's "cannot extend" refusal offers a remedy removal cannot perform


Avick3110:
## Summary

When `housecarl_remove` (or 1.x `housecarl_remove_record`) names a patch that does not exist, the shared "cannot extend" refusal ends with **"Omit into= to create it fresh, or check the name."** A removal cannot create a patch, and omitting the lane is itself refused — so the remedy names a path that does not exist for this tool.

## Measured

Against the real service (`extend-resolve-guard`'s synthetic instance), on `main` and on the #343 branch head alike:

```
svc.RemoveRecords(formids, "My Cool Patch.esp")
>>> cannot extend: no houseCARL plugin 'My Cool Patch.esp' in any houseCARL folder, and no houseCARL
    folder named 'houseCARL - My Cool Patch' or 'My Cool Patch.esp'. Omit into= to create it fresh,
    or check the name.

svc.RemoveRecords(formids, null)          # following that advice
>>> patch is required — name the houseCARL patch to remove the record from (removal only targets a
    patch that already carries it). To remove from an existing plugin IN PLACE instead, pass
    target=<plugin filename> + in_place=true.
```

So "omit it to create it fresh" leads to a second refusal, never a patch.

## Why it happens

`ResolveOwnedPatchFolder`'s not-found throw is shared by every caller of `ResolveOutputPath`, and `RemoveRecords` (`LoadOrderService.cs`) is one of them. The tail was written for the lanes that *can* create.

## Relationship to #343

#343's fix deliberately withheld its *stronger* naming remedy from this lane for exactly this reason, and pinned that with an arm. It did not change the weaker tail, so this remains. Confirmed still present at that branch's head before filing.

## Suggested direction

Give the removal lane its own fresh-write tail — one that names what a caller can actually do (create the patch first with a write tool, then remove from it), or drops the create clause entirely. Not filed as part of #343: it is a different sentence for a different lane, in a code path that fix was not otherwise touching.

*Filed by a build session while measuring #343. Unrouted — needs a RUN_ORDER routing stroke.*

Avick3110:
[...]

PR #358, opened 2026-08-17
fix(write): name patch= in the "no such patch to extend" remedy, and document it

### LINKED ISSUE #343: First-write auto-naming ("Patch.esp") doesn't surface housecarl_create_plugin as the named alternative

Avick3110:
Fixes #343

## The defect

Guessing `into=` for a patch that does not exist yet is refused, correctly. Its only remedy was *"Omit into= to create it fresh"* — which is exactly the call that produces a generically-named `Patch.esp`. `patch=`, the parameter that names a new patch, appeared in no workflow doc. That costs more than a wording nit: there is no rename primitive, so the name you get on the first write is the name you keep.

Two changes, per the triage ruling on the issue: the remedy names `patch=`, and `patch=` gets one user-facing doc line (`plugin/codex/housecarl/SKILL.md`, in the core-workflow step beside `into=` — its one home).

## Measured before written, and after

Every claim below came from running the call against the real service on a synthetic MO2 instance, not from reading the code.

| probe | result |
|---|---|
| `into="My Cool Patch.esp"` before it exists | reproduces the reported refusal verbatim |
| `patch="My Cool Patch"` | → `houseCARL - My Cool Patch\My Cool Patch.esp` |
| omit both lanes | → `Patch.esp` — the old remedy's advice **is** the defect |
| `into=` after `patch=` created it | extends the same file |
| `patch=` binding, per tool, through the real alias layer | `patch_name` everywhere except `bsa_repack`, where it binds to `archive_name` (the `.bsa`) |
| `into="GhostActive"` where `GhostActive.esp` is an active plugin, then `patch="GhostActive"` | → `GhostActive_001.esp` — `UniqueStem` auto-suffixes |

The final sentence, as delivered:

> `cannot extend: no houseCARL plugin 'My Cool Patch.esp' in any houseCARL folder, and no houseCARL folder named 'houseCARL - My Cool Patch' or 'My Cool Patch.esp'. Omit into= and pass patch="My Cool Patch" to create it fresh under a name you choose, or omit patch= too and houseCARL names it "Patch" — either name auto-suffixed if already taken. Or check the name.`

It hands back the call the caller meant, with their own guessed name in it (extension stripped, as `PatchStem` does), and promises no filename.

## Who says it, and why that is a rule rather than a list

The refusal is thrown from one shared resolver with several callers whose `patch=` means different things. The stronger sentence reaches exactly the operations whose `patch=` names a **new** patch defaulting to `Patch` — the field-write, create and forward lanes — and each states that for itself. The default is **false**, so a caller added later inherits the weaker, always-true tail rather than a false sentence.

Three ways it goes false, which is why an enumeration was the wrong shape: a removal cannot create a patch at all; `copy` and `copy_npc_appearance` create one but name it after the new EditorID / `houseCARL_NpcCopy`; and the file-placing lanes each default to their own stem (`houseCARL_Scripts`, `_Assets`, `_SEQ`, `_Archive`), with `bsa_repack` additionally binding `patch=` to the `.bsa`.

## Review rounds

**Two rounds, then a class-stop — not the three-round cap.**

**Round 1** — one high, two mediums, two lows. The high: my first cut keyed the sentence on the record-lane/rider-lane bit, but `housecarl_remove` reaches the same refusal *through the record lane*, and a removal cannot create a patch — so the remedy told that caller to re-issue the call that had just failed. Folded by making the bit the calling operation's own statement, defaulting to false. The mediums (a wrong rider enumeration; a `"Patch"` default claim false for the copy lanes) and one low folded with it.

**Refused, round 1:** the reviewer claimed `CopyTools.cs:89` documents a `"Patch"` default contradicting the code. It does not — that text is `ApplyTools`'. Nothing folded on it.

**Round 2** — one medium, three lows, and the medium was the same failure class as round 1's high: *the remedy asserts something the write can falsify*, this time via `UniqueStem`'s auto-suffix. Under CLAUDE.md §5 #11 a class recurring in two consecutive rounds is a design signal, not a fix queue, so the rounds stopped there and the seam went to Aaron rather than into a third fold.

**Round 2 self-refused** one of its own candidates: `WritePatchBuilder.cs:404/2235/2760` still carry the old bare tail, but that arm is a race guard (the path vanishing between resolve and write), not a reachable second home for the sentence. Agreed and not folded.

**Directed completion** (advisor ruling, §4 outcome (b) — the triage ruling had assumed a refusal can hand back the working call with its resulting filename named; it cannot, and the goal survives dropping the prediction):

- The sentence stops predicting the filename, and the qualifier scopes **both** names it offers — the ruling caught that my proposed fix still predicted the default-lane name, which the same collision arm falsifies.
- Both classes folded structurally: the roster in the comment and CHANGELOG became a stated rule, absorbing round 2's three lows including the rider grounds (now naming both the `bsa_repack` alias *and* the non-`Patch` rider defaults).
- One false claim of mine corrected: `housecarl_remove` does declare `into=` (`RemoveTools.cs:59`). The create-half falsity stands and is what the arm now says.

No round 3 (§11: no new rounds after directed folds).

## Guards

`extend-resolve-guard` grows from 8 arms to 14. Every branch is RED-checked in both directions — five sabotages, each reddening exactly its own arms and nothing else:

| sabotage | reddens |
|---|---|
| record lane forced onto the rider sentence | arm 8's remedy pin |
| rider lane forced onto the record sentence | 8c |
| removal lane opted into the naming sentence | 8d |
| `patch=` ignored in `ResolveOutputPath` | the seed arm |
| pre-ruling wording restored | 8's qualifier pin + 8b2 |

Arm 8b and the new 8b2 **make the calls the sentence names** rather than reading it — including the collision case that falsified the previous wording.

`ci-all` 124/124.

## Filed separately

Two confirmed defects outside this branch, both measured here, both unrouted and unmilestoned:

- #356 — `housecarl_remove`'s refusal offers a remedy removal cannot perform (present on `main`; this branch stops making it worse but does not fix it).
- #357 — the file-placing lanes have no discoverable way to name the patch folder they create.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -15,13 +15,17 @@ when it changes.
   than a wording nit because there is no rename primitive: the name you get on the first write is the name you
   keep. `patch=` is now documented as well, in the core workflow beside `into=`.
 
-  The stronger remedy reaches `housecarl_apply` / `housecarl_create` / `housecarl_forward` and their 1.x
-  equivalents, and deliberately not the tools where it would be false. `housecarl_remove`'s `patch=` names an
-  **existing** patch, so it would send you to re-issue the call that just failed. `housecarl_copy` and
-  `housecarl_copy_npc_appearance` do create a fresh patch, but name it after the new EditorID rather than `Patch`.
-  And among the file-placing tools (`place_asset`, `compile_script`, `decompile_script`, `bsa_repack`, `nif_set`,
-  `write_seq`), `bsa_repack`'s `patch=` names the `.bsa` rather than the mod folder. All of those keep the shorter,
-  always-true remedy.
+  It offers the name without promising the filename, because houseCARL auto-suffixes a name already taken by an
+  active plugin — and that is the likely case, since the name you guessed at `into=` is usually one you saw in your
+  load order. Both names the sentence mentions carry that qualifier, the one you choose and the `Patch` default.
+
+  The rule for which tools say it: exactly those whose `patch=` names a **new** patch defaulting to `Patch` — the
+  field-write, record-creation and forward lanes. Everywhere else the shorter, always-true remedy stands, because
+  the naming sentence would be false there: a removal edits a patch that already exists and cannot create one; a
+  closure copy or an appearance copy names its fresh patch after the new EditorID; and the tools that place a file
+  into a patch folder each default to their own stem (`houseCARL_Scripts`, `houseCARL_Assets`, `houseCARL_SEQ`,
+  `houseCARL_Archive`) rather than to `Patch` — with `bsa_repack` a further case, where `patch=` names the `.bsa`
+  itself rather than the mod folder.
 
 - **Fixed: reading a cell's placed references could report an empty cell the game fills.** Placed references, a
   topic's INFO lines and a worldspace's cells are declared **per plugin**, and the game assembles a parent's
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -6,6 +6,17 @@ when it changes.
 
 ## Unreleased
 
+- **Fixed: the refusal for "no such patch to extend" now says how to give a new patch a name.** Guessing
+  `into="My Cool Patch.esp"` for a patch you have not created yet is refused, correctly — but the only remedy it
+  offered was to omit `into=` and create the patch fresh, which is precisely the call that produces a generically
+  named `Patch.esp`. The parameter that names a new patch, `patch=`, appeared in nothing you read. The refusal now
+  hands back the call you meant, with your own guessed name already in it — `pass patch="My Cool Patch"` — and says
+  what omitting it costs. This matters more than a wording nit because there is no rename primitive: the name you
+  get on the first write is the name you keep. `patch=` is now documented as well, in the core workflow beside
+  `into=`. The file-placing tools (`place_asset`, `compile_script`, `decompile_script`, `bsa_repack`, `nif_set`,
+  `copy_npc_appearance`) keep the shorter remedy on purpose: on `housecarl_bsa_repack`, `patch=` names the `.bsa`
+  rather than the mod folder, so pointing every caller at it would be wrong for that one.
+
 - **Fixed: reading a cell's placed references could report an empty cell the game fills.** Placed references, a
   topic's INFO lines and a worldspace's cells are declared **per plugin**, and the game assembles a parent's
   children from every plugin that declares them. So a plugin that overrides a cell for an unrelated reason —
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -9,13 +9,19 @@ when it changes.
 - **Fixed: the refusal for "no such patch to extend" now says how to give a new patch a name.** Guessing
   `into="My Cool Patch.esp"` for a patch you have not created yet is refused, correctly — but the only remedy it
   offered was to omit `into=` and create the patch fresh, which is precisely the call that produces a generically
-  named `Patch.esp`. The parameter that names a new patch, `patch=`, appeared in nothing you read. The refusal now
-  hands back the call you meant, with your own guessed name already in it — `pass patch="My Cool Patch"` — and says
-  what omitting it costs. This matters more than a wording nit because there is no rename primitive: the name you
-  get on the first write is the name you keep. `patch=` is now documented as well, in the core workflow beside
-  `into=`. The file-placing tools (`place_asset`, `compile_script`, `decompile_script`, `bsa_repack`, `nif_set`,
-  `copy_npc_appearance`) keep the shorter remedy on purpose: on `housecarl_bsa_repack`, `patch=` names the `.bsa`
-  rather than the mod folder, so pointing every caller at it would be wrong for that one.
+  named `Patch.esp`. `patch=`, the parameter that names a new patch, was documented only in the parameter list
+  itself and in no workflow doc you would read first. The refusal now hands back the call you meant, with your own
+  guessed name already in it — `pass patch="My Cool Patch"` — and says what omitting it costs. This matters more
+  than a wording nit because there is no rename primitive: the name you get on the first write is the name you
+  keep. `patch=` is now documented as well, in the core workflow beside `into=`.
+
+  The stronger remedy reaches `housecarl_apply` / `housecarl_create` / `housecarl_forward` and their 1.x
+  equivalents, and deliberately not the tools where it would be false. `housecarl_remove`'s `patch=` names an
+  **existing** patch, so it would send you to re-issue the call that just failed. `housecarl_copy` and
+  `housecarl_copy_npc_appearance` do create a fresh patch, but name it after the new EditorID rather than `Patch`.
+  And among the file-placing tools (`place_asset`, `compile_script`, `decompile_script`, `bsa_repack`, `nif_set`,
+  `write_seq`), `bsa_repack`'s `patch=` names the `.bsa` rather than the mod folder. All of those keep the shorter,
+  always-true remedy.
 
 - **Fixed: reading a cell's placed references could report an empty cell the game fills.** Placed references, a
   topic's INFO lines and a worldspace's cells are declared **per plugin**, and the game assembles a parent's
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -15,17 +15,16 @@ when it changes.
   than a wording nit because there is no rename primitive: the name you get on the first write is the name you
   keep. `patch=` is now documented as well, in the core workflow beside `into=`.
 
-  It offers the name without promising the filename, because houseCARL auto-suffixes a name already taken by an
-  active plugin — and that is the likely case, since the name you guessed at `into=` is usually one you saw in your
-  load order. Both names the sentence mentions carry that qualifier, the one you choose and the `Patch` default.
+  It offers the name without promising the filename, because houseCARL auto-suffixes a name that is already taken —
+  either by a mod folder it already made under that name, or by an active plugin. **Read the patch name back off
+  the response**; both names the sentence mentions carry that qualifier, the one you choose and the `Patch` default.
 
   The rule for which tools say it: exactly those whose `patch=` names a **new** patch defaulting to `Patch` — the
   field-write, record-creation and forward lanes. Everywhere else the shorter, always-true remedy stands, because
   the naming sentence would be false there: a removal edits a patch that already exists and cannot create one; a
   closure copy or an appearance copy names its fresh patch after the new EditorID; and the tools that place a file
-  into a patch folder each default to their own stem (`houseCARL_Scripts`, `houseCARL_Assets`, `houseCARL_SEQ`,
-  `houseCARL_Archive`) rather than to `Patch` — with `bsa_repack` a further case, where `patch=` names the `.bsa`
-  itself rather than the mod folder.
+  into a patch folder each default to their own stem, never to `Patch` — with `bsa_repack` a further case, where
+  `patch=` names the `.bsa` itself rather than the mod folder.
 
 - **Fixed: reading a cell's placed references could report an empty cell the game fills.** Placed references, a
   topic's INFO lines and a worldspace's cells are declared **per plugin**, and the game assembles a parent's
--- a/plugin/codex/housecarl/SKILL.md
+++ b/plugin/codex/housecarl/SKILL.md
@@ -21,7 +21,7 @@ Use this skill for data-layer Skyrim Special Edition modding through the configu
    - `housecarl_bulk_apply` for several edits in one patch, dict merges, leveled-list entries, effects, or other composed structs.
    - `housecarl_create_record` (or `housecarl_bulk_create` for many at once) for a new top-level record — it needs an EditorID.
    - `housecarl_remove_record` only to drop a record or override from a houseCARL-owned patch — never from a source mod.
-4. Name the patch on the **first** write with `patch=<name>` — omit it and houseCARL names it `Patch`. After that, accumulate related edits into it with `into=<patch filename>`.
+4. Name the patch on the **first** write with `patch=<name>` — omit it and houseCARL names it `Patch`. Either way the name is auto-suffixed if it is already taken, so read the patch name back off the response. After that, accumulate related edits into it with `into=<patch filename>`.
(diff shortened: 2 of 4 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7843,11 +7853,40 @@ string ResolveOwnedPatchFolder(string into, bool needEsp)
...
+        // folder — because it declares both and §5.3 routes patch= to the artifact.
+        //
+        // Hence the default is FALSE and the tail is the weaker, always-true one: a caller added later without a
+        // thought about any of this gets a sentence that is merely less helpful, never wrong.
+        //
+        // Note what the sentence does NOT do: predict the resulting filename. UniqueStem auto-suffixes a stem whose
+        // .esp is already an active plugin — and that is the correlated case, since a caller guessing into= usually
+        // guessed it off a plugin name in their load order. The qualifier scopes BOTH names the sentence mentions,
+        // the chosen one and the "Patch" default, because the same arm falsifies either.
         throw new InvalidOperationException(
```
Avick3110, line 7884: **[medium] The sibling refusal ~30 lines up still carries the pre-#343 dead end.**

The naming sentence is attached only to this not-found throw. The un-owned-folder arm at 7848–7854 — same method, same `patchNamesFresh: true` callers — still ends `"Use a different patch name."`, naming no parameter and no lane switch.

Concrete: a modder has an ordinary (non-houseCARL) MO2 mod folder literally named `My Cool Patch` and calls `housecarl_apply(..., into="My Cool Patch")`. The step-(4) loop over `{"houseCARL - My Cool Patch", "My Cool Patch"}` hits the un-owned candidate *first* and throws before reaching here. The caller is told to pick a different name — so they guess another `into=` (still the wrong lane), or drop `into=` and get `Patch.esp`. That is exactly the shape #343 describes, on a branch the guess-off-a-mod-name case reaches at least as often as the not-found one.

Not asking for a fold in this PR necessarily — but if it's out of scope it belongs alongside #356/#357 rather than unrecorded, since the branch's own premise is that this remedy class must name `patch=`.

plugin/codex/housecarl/SKILL.md
```diff
@@ -21,7 +21,7 @@ Use this skill for data-layer Skyrim Special Edition modding through the configu
    - `housecarl_bulk_apply` for several edits in one patch, dict merges, leveled-list entries, effects, or other composed structs.
    - `housecarl_create_record` (or `housecarl_bulk_create` for many at once) for a new top-level record — it needs an EditorID.
    - `housecarl_remove_record` only to drop a record or override from a houseCARL-owned patch — never from a source mod.
-4. Accumulate related edits into one patch with `into=<patch filename>` after the first write returns a patch name.
+4. Name the patch on the **first** write with `patch=<name>` — omit it and houseCARL names it `Patch`. Either way the name is auto-suffixed if it is already taken, so read the patch name back off the response. After that, accumulate related edits into it with `into=<patch filename>`.
```
Avick3110: **[low] This blanket rule is false for the tool named two lines above it.**

Step 3's last bullet is `housecarl_remove_record`, where `patch=` is *declared* by the tool and names an **existing** patch — it can never name a fresh one. Step 4 then says, without qualification, that `patch=<name>` names the patch and that omitting it yields `Patch`.

An agent that walks the core workflow, picks `housecarl_remove_record`, and applies step 4 calls it with `patch="My Cool Patch"` expecting to name a patch, and gets `cannot extend: no houseCARL plugin 'My Cool Patch.esp' … Omit into= to create it fresh` — #356's exact shape.

The CHANGELOG carves removal out of the rule explicitly; this file, the one a caller actually reads first, does not. Line 47's `housecarl_remove` note ("there is no `patch=`") covers only the 2.0 spelling, and it is 23 lines below the workflow step.

A half-clause would close it, e.g. "…on a write that creates a patch (not `housecarl_remove_record`, whose `patch=` names one that already exists)".

plugin/CHANGELOG.md
```diff
@@ -6,6 +6,27 @@ when it changes.
...
+
+  It offers the name without promising the filename, because houseCARL auto-suffixes a name already taken by an
+  active plugin — and that is the likely case, since the name you guessed at `into=` is usually one you saw in your
+  load order. Both names the sentence mentions carry that qualifier, the one you choose and the `Patch` default.
+
+  The rule for which tools say it: exactly those whose `patch=` names a **new** patch defaulting to `Patch` — the
+  field-write, record-creation and forward lanes. Everywhere else the shorter, always-true remedy stands, because
+  the naming sentence would be false there: a removal edits a patch that already exists and cannot create one; a
+  closure copy or an appearance copy names its fresh patch after the new EditorID; and the tools that place a file
+  into a patch folder each default to their own stem (`houseCARL_Scripts`, `houseCARL_Assets`, `houseCARL_SEQ`,
```
Avick3110: **[low] This enumeration names 4 of at least 8 rider default stems** — and the same four appear in the load-bearing source comment at `LoadOrderService.cs:7869-7870`.

Also rider default stems, none listed:

- `houseCARL_NifEdit` — `LoadOrderService.cs:1441` (`nif_set`)
- `houseCARL_Extract` — `BsaTools.cs:75` (`bsa_extract`)
- `"<name> compacted"` — `LoadOrderService.cs:6072` (`compact_plugin`)
- `"<name> merged"` — `LoadOrderService.cs:6296` (`merge_plugins`)

The *rule* survives — none of them default to `Patch` — but the list is presented as the rule's grounds, and the PR body records that this roster "was wrong twice before it became a rule". Third version, still incomplete. Reading it as exhaustive is what a maintainer adding a rider lane will do when deciding whether their call site joins the fresh-naming set.

Cheapest fix: drop the parenthetical to "each default to their own stem, never to `Patch`" — the claim that actually holds, and the one the code enforces.

plugin/CHANGELOG.md
```diff
@@ -6,6 +6,27 @@ when it changes.
...
+- **Fixed: the refusal for "no such patch to extend" now says how to give a new patch a name.** Guessing
+  `into="My Cool Patch.esp"` for a patch you have not created yet is refused, correctly — but the only remedy it
+  offered was to omit `into=` and create the patch fresh, which is precisely the call that produces a generically
+  named `Patch.esp`. `patch=`, the parameter that names a new patch, was documented only in the parameter list
+  itself and in no workflow doc you would read first. The refusal now hands back the call you meant, with your own
+  guessed name already in it — `pass patch="My Cool Patch"` — and says what omitting it costs. This matters more
+  than a wording nit because there is no rename primitive: the name you get on the first write is the name you
+  keep. `patch=` is now documented as well, in the core workflow beside `into=`.
+
+  It offers the name without promising the filename, because houseCARL auto-suffixes a name already taken by an
```
Avick3110: **[low] "already taken by an active plugin" is one of the two suffix triggers, and the rarer one.**

`IsStemFree` (`LoadOrderService.cs:7781`) is:

```csharp
=> !Directory.Exists(Path.Combine(_modsDir, ModFolderName(stem))) && !activePlugins.Contains(stem + ".esp");
```

The folder arm fires with no active plugin involved. Scenario: a modder writes `patch="My Cool Patch"`, reviews it in MO2 but hasn't enabled it yet — houseCARL's own documented default, the modder reviews and enables — then next session passes `patch="My Cool Patch"` again instead of `into=`. `houseCARL - My Cool Patch` exists, so the write lands in `houseCARL - My Cool Patch_001` while `My Cool Patch.esp` was never in the load order.

That is the *repeat-caller* case, which is more common than the guessed-off-the-load-order case the paragraph builds its "and that is the likely case" argument on. A reader who takes the stated cause literally concludes a not-yet-enabled patch can't be suffixed and stops reading the name back off the response — the exact habit the SKILL.md line asks for.

The error string itself ("if already taken") is correct and unaffected; only this explanatory sentence overstates.

src/housecarl-generator/ExtendResolveProbe.cs
```diff
@@ -210,6 +216,101 @@ void MarkOwned(string folder, string plugin) =>
...
+
+                var r = svc.ApplyEdits(new[] { Wgt(4) }, null, "GhostActive");
+                Check(!r.Success && r.Error is not null
+                      && r.Error.Contains("pass patch=\"GhostActive\"", StringComparison.Ordinal),
+                      "into=\"GhostActive\" (an active plugin, foreign folder) still refuses with the naming remedy");
+
+                var followed = svc.ApplyEdits(new[] { Wgt(4) }, "GhostActive", null);
+                Check(followed.Success && Ends(followed.OutputPath, "houseCARL - GhostActive_001", "GhostActive_001.esp"),
+                      $"…and following it auto-suffixes off the active plugin ({Path.GetFileName(followed.OutputPath)})");
+                Check(r.Error is not null && !r.Error.Contains("under that name", StringComparison.Ordinal),
```
Avick3110: **[low] This negative pins one superseded phrase, not the property in its label.**

The label says "the sentence promises no filename". The assertion says the sentence does not contain `"under that name"` — a literal that appears in no current wording, only in the pre-ruling one. So the arm reddens for exactly the one sabotage the PR ran (restore the old wording) and for nothing else.

A later fold that predicts the filename *differently* — "…and houseCARL writes `GhostActive.esp`", "…creating `houseCARL - GhostActive`" — is false in precisely the collision case this arm goes to the trouble of building (the write auto-suffixes to `GhostActive_001.esp`), and passes this check unchanged. The guard would report green on the regression it exists to catch.

The real load is carried by the positive pin at line 233 (`"either name auto-suffixed if already taken"`) plus the `followed` call above, which together already make the property observable. If the negative is worth keeping, pinning it to something the *current* sentence must not contain — the resolved stem, or `"_001"` — would make it detect more than history.

Avick3110:
## Fold summary — Aaron's review

Five findings, all five reproduced at the line cited. **Four folded in one commit (`2811314`), one routed to the issues lane** on Aaron's own framing (*"if it's out of scope it belongs alongside #356/#357 rather than unrecorded"*). Nothing refused.

### [medium] sibling un-owned-folder refusal still dead-ends → filed as #359

Not folded. Measured first, and the measurement is why it is design rather than a sentence: with a foreign folder `My Cool Patch` holding an **inactive** `My Cool Patch.esp`, `into=` refuses with `"Use a different patch name."` — and following a naive `patch="My Cool Patch"` remedy then writes `houseCARL - My Cool Patch\My Cool Patch.esp`, because `IsStemFree` tests only for an existing `houseCARL - <stem>` folder and an **active** plugin of that name. So the remedy would hand the caller a stem that shadows the foreign plugin. Combined with the arm's genuine ambiguity — reaching it by name may mean *edit that plugin* (`in_place=`), not *create fresh* — the honest remedy has to name a lane, which is out of this PR's chartered scope. Scenario and measurement carried into #359 verbatim.

### [low] SKILL.md blanket rule → folded

Carve-out now rides step 4 itself: *"…on the **first** write that creates one with `patch=<name>` (not `housecarl_remove_record`, whose `patch=` names one that already exists)"* — rather than depending on line 47, which is 23 lines below and covers only the 2.0 spelling.

### [low] rider-stem roster names 4 of ≥8 → folded in **both** homes

The same four-name list was in the CHANGELOG parenthetical and in the source comment. Both deleted. The CHANGELOG now states only the claim the code enforces ("each default to their own stem, never to `Patch`"); the comment says why there is no list — the call sites are the authority, and a list of them is what keeps going stale. This was the roster's fourth version; the fix is it not existing.

### [low] suffix-trigger overstatement → folded, and swept

The source comment at ~7877 made the same active-plugin-only claim as the CHANGELOG sentence. Both now state both of `IsStemFree`'s triggers and which is commoner — the repeat caller whose earlier patch folder exists while its plugin was never enabled. The CHANGELOG also now says **read the patch name back off the response**, since the point of the finding is that a reader who takes the stated cause literally stops doing exactly that. The error string itself was correct and is unchanged.

### [low] history-pinned negative arm → folded

Replaced `!Contains("under that name")` — a literal only the pre-ruling wording had — with `!Contains(<the stem actually written>)`, computed from the `followed` call in the same fixture. The positive pin at 233 stays.

**One honest limit, stated in the comment rather than implied:** it catches a wording that echoes the *suffixed* name, but not one predicting the *unsuffixed* name, because the refusal legitimately quotes `"<stem>.esp"` already as the plugin it searched for — no string test can tell that occurrence from a prediction. The property's real load stays on the positive pin plus the `followed` call.

RED-checked with a sabotage of the kind the old pin would have missed — appending `writing {UniqueStem(stem)}.esp` to the sentence. Both the new negative and the qualifier pin go red; the old pin would have stayed green.

### State

`ci-all` 124/124, tree clean. `extend-resolve-guard` unchanged at 14 arms (one replaced, none added).

**Process note, recorded rather than smoothed over:** the first attempt at this RED-check stashed the uncommitted fold to fake a committed state, the `stash pop` conflicted, and the fold had to be recovered from the stash. Committing the fold first and sabotaging from there — which is what the lesson actually says — is what worked.

No new review round (§11: no new rounds after directed folds). Merge is Aaron's go.

Issue #344, opened 2026-08-13
check_errors: dangling-ref budget is spent on vanilla baseline in load order — mod-introduced findings become unreachable


BumblingOaf:
## Summary

`housecarl_check_errors` on a large (3700+ plugin) load order reported "4992 dangling ref(s)"
across a 732-line, 80K-character report. Every example I actually inspected traced back to
`Skyrim.esm` itself — well-known, permanently-unfixable vanilla leftovers (e.g. the `CWTestHold`
developer test worldspace) that exist in every Skyrim installation regardless of mods. None of
what I sampled was mod-introduced.

## Why this matters

A tool whose default output is dominated by thousands of lines of expected, unfixable vanilla
baseline noise makes it very hard to spot the handful of *actionable* dangling references a
user's own mod list might have introduced. I was specifically trying to verify "did today's
changes introduce any new problems," and grepping the report for the specific files I cared about
(and finding zero matches) ended up being a more useful signal than the report itself.

## Suggested fix

An option to exclude dangling refs whose winning/reporting plugin is a base-game master
(`Skyrim.esm`, `Update.esm`, the official DLCs) from the default output — or at minimum, a summary
line breaking down "N of these M are attributed to base-game masters (known, expected)" so a
reader knows how much of the count is baseline noise before deciding whether to read further.

## Environment

- houseCARL v1.9.0
- Load order: 3767 plugins

Avick3110:
Triaged and **confirmed** — and the defect is sharper than "buried." The dangling listing budget (`limit=`, default 1000) is a single counter consumed plugin-by-plugin in load order, and `Skyrim.esm` is index 0. On an order where the vanilla baseline exceeds the budget, a mod plugin with real dangling refs collects an empty list and then fails the report-inclusion test — it doesn't appear in the output at all. The only trace is the global "capped at limit; true total = N" line, which doesn't say which plugins lost their entries. So at defaults on a large order, mod-introduced findings aren't hard to spot — they're structurally unreachable.

One correction to the report: per-source-plugin grouping does exist (the `[ERROR] <plugin>` sections, and a `plugins[]` array in JSON) — but the per-plugin counts are silently deflated by the shared budget, and the one existing histogram is keyed by TARGET plugin (which absent master a ref points at), not source, so it can't answer "how much of this is vanilla."

[...]

PR #360, opened 2026-08-18
fix(check_errors): base-master separation at the sweep layer, and an omission sentence per layer

Avick3110:
Base-master separation at the sweep layer: the dangling listing budget is spent on mod plugins before the vanilla baseline, the response splits the dangling total into baseline vs the rest, `counts_only=true` gains a by-SOURCE-plugin histogram beside the existing by-TARGET one, and a capped or truncated response says what it dropped — with one known gap: the truncation notice fires at a section BOUNDARY, so a cut landing inside the LAST report section is still silent (**#361**, pre-existing and filed rather than folded, per §2).

References #344. **No closing keyword — #344 stays open**, tracking its second stage: the `exclude=` pole on the successor `check` surface (RUN_ORDER row 4). That axis has no SPEC pole today, so it is deliberately not in this PR (#342 stage-1 precedent). The issue is already milestoned; no milestone action here.

## The defect, and a correction to its triage premise

The triage on #344 said the vanilla baseline "exhausts the budget before any mod plugin is reached". Measured on the real order, that premise is wrong in its mechanism and right in its effect. `Skyrim.esm` carries 384 dangling refs against a default `limit=1000`, so it does not exhaust the budget alone — it **crowds** from index 0, together with everything ahead of the plugin you care about. The listing is spent strictly in load order, so what actually determines whether a mod's findings are reachable is its position, not the vanilla total.

The effect the issue describes is real and worse than "buried": a plugin whose findings collect an empty list fails the report-inclusion test and vanishes from the report entirely.

**A/B on the 3767-plugin order, default `limit=1000`:**

| | plugins with a listed section |
|---|---|
| before | 70 of 180 |
| after | 136 of 180 |

66 plugins that had findings and no way to be seen now have one. That A/B, not the original premise, is the justification for the change.

## What ships

- **Budget order.** The sweep runs in two phases: non-base plugins first, base masters second. Report sections re-sort into load order afterwards, so the reading order is unchanged.
- **Baseline split.** `baseline: N of M dangling ref(s) come from the base-game master(s) this sweep covered (…)`, naming the base masters the sweep actually examined. On the real order: 398 of 4996 are vanilla.
- **Source axis.** `counts_only=true` tallies by SOURCE plugin beside the existing TARGET axis. Target answers "which absent dependency is behind this wall"; source answers "how much of the wall is vanilla and how much did my mods introduce".
- **Omission reporting** — see the layer rule below.

The base-master set is `Implicits.Get(SkyrimSE).BaseMasters` — Mutagen's own, never a list kept here (settled decision 2). Creation Club and `_ResourcePack.esl` are deliberately **not** baseline (ruled 2026-08-18): baseline is Mutagen's set exactly, which is what keeps it by-construction, and "keep CC out of my listing" is a caller's choice routing to the `exclude=` pole, not a classification baked into the sweep.

## The layer rule, and why the review rounds stopped

Rounds 1 and 2 kept returning one class of finding: **a sentence computed from sweep data but read as a description of the response.** `check_errors` truncates twice, independently — the listing budget (`limit=`), and the render's `max_chars` — and the new sentences crossed that seam.

Measured on the real order at plain defaults, before the fix: the line said "3996 not listed across 45 plugins" while **4442 of 4996 findings were absent from the response** and **70 plugins with findings were named nowhere in it**. That is #344's own defect — findings present, plugin untraceable — reproduced one layer up, with a confident wrong number attached.

Per CLAUDE.md §5 #11 the class recurring in two consecutive rounds is a design signal, so the seam went to the advisor lane rather than into a third fold. The ruling:

> **An omission sentence is computed by the layer that performed the omission, and claims only that layer's omission.**

Pinned as the comment where these sentences live. Applied:

- The capped line's subjects are the budget's: `[the listing budget (limit=) omitted 3996 dangling ref(s); true total = 4996, across 45 source plugin(s). Largest sources (the 10 largest of 45; the rest are not named here): …]`. True regardless of what the render then does.
- The `max_chars` notice counts what the **render** dropped, at the break point where the loop already knows it: `70 of 137 plugin section(s) were not rendered, and the last one rendered may be partial; this is the response being cut, separate from any limit= omission reported below`. Previously it said only that something was cut.
- JSON states its own cut in its own terms. `rendered` + `truncated` could not answer "how many were dropped" — there was no total to subtract from — so `plugins_with_findings` rides along and the arithmetic is exact.
- The "no section above" sentence tested `r.Reports` (sweep state) to describe response content. The conditional was **deleted** rather than moved, and the sentence is now a rule about the budget (#339 precedent: prefer deleting an unfixturable conditional).
- `base_masters_swept` means the scope **admitted a record**, not that the file was opened — a record scope that filters out every vanilla record no longer reports the baseline as covered-and-clean.
- The phase-order sentence compares two groups, so it now requires a non-base plugin to be in scope.

The two cuts deliberately do not sum. A caller wanting one number for "what is this response missing" uses `counts_only=true` today; render-aware accounting belongs to PR 4.

**Design note for PR 4:** `check`'s rebuilt response layer should be born with render-aware omission accounting — one shared source, both formats — which retires the two-layer decomposition this PR ships. Building it here would have put per-format accounting in a layer about to be rebuilt, the twin-drift shape #337 killed.

## Review rounds

**Round 1 — 3 fresh agents, isolated worktrees: 12 findings, 11 folded, 1 refused.**
- HIGH: the omissions roster stopped at 10 plugins while claiming to be the only place a dropped plugin appears (on the real order, 45 plugins lose entries).
- MED, found independently by two reviewers: the baseline line named all five base masters however few were swept.
- MED: the remedy offered scoping `plugins=` as sufficient; measured, that loops when a plugin's own set exceeds `limit=` (the largest single source carries 2591 refs against a default of 1000).
- MED: the baseline line's excluded-class gate was hollow — with the walk skipped it printed "0 of 0 come from the base-game masters" two lines under "dangling refs NOT CHECKED".
- LOW: `ListingOmitted` assigned rather than accumulated, so a name repeated in `plugins=` reported dropped refs on a sweep that never capped; JSON's omissions table was row-capped by `limit=`, the same knob that caused the omissions; two empty histograms rendered as identical untitled lines; two comments disagreed about whether an off-order file can be a base master.
- Coverage gaps closed: the off-order lane's contribution to both new fields, a plugin losing its **entire** set, and a roster with more losers than it names — each previously deletable with the suite green.

**Refused (1 of 12):** that `BASE-SET-BY-CONSTRUCTION` should assert the set's *provenance* rather than its contents. Correct that value equality would pass for a hand-kept literal matching today's five names — but holding provenance needs a source-text assertion over `ErrorCheck.cs`, an idiom this repo has nowhere, and adding one is a guard-design call rather than a fold. The arm was renamed to `BASE-SET-CONTENTS` and its reach stated in a comment, so it cannot be read as claiming more.

**Round 2 — 2 fresh agents: both independently found the `max_chars` seam.** Class-stop invoked; no round 3. What followed was a directed completion under the advisor ruling (#352/#358 precedent), which also covered round 2's smaller findings: the record-scope `base_masters_swept` overclaim, the phase sentence firing where only base masters were in scope, two conditionals with one RED-checked arm, and an orphaned XML doc comment.

## One commit is take-or-drop: `eea7f41`, a CLAUDE.md edit

`eea7f41` adds a single bullet to §5 #11 — sabotage RED-checks run from a committed state, restores verified by grep rather than assumption, with the stash variant named as the same trap. It is self-governing config, so it stands alone rather than sitting inside a fold commit and can be dropped without touching the fix (PR #346 / `b6e0091` precedent). It lands the promotion threshold the 2026-08-17 handoff set: two consecutive sessions hit the trap, this session included, before it became repo text.

## Aaron's review — folded

Four findings, all verified against branch source: **3 folded, 1 filed, 0 refused.** Fold commit `b897497`; no review round follows it (§11: directed fold).

- **F1 — a `max_chars` cut inside the last section is silent.** Filed as **#361**, not folded: the inner break is pre-existing and untouched here, so §2 routes it to the issue lane. The render-cut notice now states its own reach in the code where it lives, citing the issue. One correction to the triage note: **JSON is not silent on this path.** Its inner `dangling` loop has no cap check at all, so a plugin is written whole or not at all — the JSON failure mode is overshooting `max_chars`, not losing entries. Measured, and the issue says so.
- **F2 — the phase-clause gate was a subtraction.** `PluginsScanned > BaseMastersSwept.Count` compares two counts that measure different things; they diverge whenever a base master is in `targets` but contributes nothing to `examined`. The result now carries `NonBaseInScope`, computed in `ErrorCheck.Run` from the resolved targets. The old arm passed only because its single-name fixture made the counts coincide — the new fixture (`plugins=["Skyrim.esm","Skyrim.esm"]`) holds them apart at scanned 2 / swept-base 1, and reverting the render to the subtraction turns `BASELINE-PHASE-CLAUSE-NOT-A-SUBTRACTION` RED.
- **F3 — the two cuts were stated in different units.** The render now counts entries where it appends one and says `N of the M budget-listed dangling entries appear above`, in the budget line's own unit. Counting emitted entries rather than summing `p.Dangling.Count` is what makes it true of a partial section — sabotaging it to the section sum turns the arm RED. The "may be partial" hedge is gone, and its reason with it. Measured before commit: `10 of 12 plugin section(s) were not rendered; 4 of the 36 budget-listed dangling entries appear above`, against 4 entry lines actually in the response.
- **F4 — doc drift.** The comment named `base_masters_in_scope`; the writer emits `base_masters_swept`. The class doc-comment's response-shape list had drifted the same way and is refreshed, now carrying a note that it is a hand-kept second spelling outside `wire-names-guard`'s reach: **if it drifts again, delete it** and let the writer be the authority (#358's roster lesson). No arm can hold a comment's accuracy — that note is the mitigation, stated rather than implied.

**Ground for zero refusals:** F1, F2 and F4 each cite a line that reproduces as described (F1 and F2 re-measured here on synthetic fixtures), and F3 is Aaron's ruling on a question this PR's own body asked.

## Verification

- `ci-all` **124/124 ALL PASS**; `check-errors-guard` arms **34 → 67** (+33). (Earlier revisions of this body and some commit messages said "13 → …" — that was a miscount of the base, corrected here; the number of arms ADDED, 33, was right.)
- Every conditional added by this PR is RED-checked in **both** directions: 14 sabotages run, each turning its arm RED for the stated reason, each restored and the restore verified by grepping for a string the fold introduced.
- Every sentence this PR adds or changes was **measured** — rendered for the case in question — rather than reviewed by reading. Two changes came from measuring rather than reading: the "no section above" sentence printing where nothing was missing, and this round's whole layer split.
- One arm was hardened after a sabotage revealed it: `RENDER-CUT-JSON-DERIVABLE` used `GetProperty`, so deleting the field crashed the guard instead of turning the arm RED.
- Acceptance shape per settled decision 3: synthetic fixtures throughout, RED first. `EMPIRICAL_LEDGER.md` untouched, `copy_npc_appearance` untouched. Live-order numbers in this PR are read-only measurements.


src/housecarl-mcp/ReadTools.cs
```diff
@@ -1171,15 +1181,26 @@ public static string RenderCheckErrors(ErrorCheckResult r, int maxChars, int his
...
         {
             if (sb.Length >= cap)
             {
-                sb.Append("\n... [truncated at max_chars=").Append(cap).Append("; ").Append(SweepNarrowHint).Append("]\n");
+                // AN OMISSION SENTENCE IS COMPUTED BY THE LAYER THAT PERFORMED THE OMISSION, AND CLAIMS ONLY THAT
+                // LAYER'S OMISSION. This is the render's cut, so it is counted here, where the loop already knows
+                // what it emitted — and it says nothing about the listing budget, whose own omissions the capped
+                // line below reports in its own terms. The two truncators are independent and do not sum: crossing
+                // that seam is what made rounds 1 and 2 a single class (advisor ruling, 2026-08-18).
+                sb.Append("\n... [truncated at max_chars=").Append(cap).Append(": ")
```
Avick3110, line 1201: The per-plugin dangling loop breaks on the cap (`if (sb.Length >= cap) break;`, ~L1214) **without setting `truncated`**. When the cut lands inside the **last** report section, the outer `foreach` then ends normally, `truncated` stays false, and none of this new notice is emitted — the render drops findings with no sentence at all, and `AppendExcludedPlugins` still appends past the cap.

Repro: `plugins=["BigMod.esp"], limit=5000` where BigMod carries 3000 dangling refs, default `max_chars`. Header prints `dangling reference(s) (3000):`, ~600 entries print, then the boundary block — no `[truncated at max_chars=…]` anywhere.

Non-last sections are covered (the next iteration fires the notice, and `the last one rendered may be partial` is exactly this case disclosed). It is the last one that goes silent. The line itself is pre-existing and unchanged by this branch, so per CLAUDE.md §2 this is issue-lane rather than a fold — but it is the one path the new render-cut sentence claims to close and does not, so it is worth filing before the PR merges rather than after.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -1204,15 +1225,98 @@ public static string RenderCheckErrors(ErrorCheckResult r, int maxChars, int his
...
+          .Append(" dangling ref(s) come from the base-game master(s) this sweep covered (").Append(string.Join(", ", swept))
+          .Append(") — vanilla leftovers rather than anything this load order introduced; ")
+          .Append(r.TotalDangling - r.BaselineDangling).Append(" come from the rest of the swept scope.").Append('\n');
+        // The phase-order sentence only where the phase order DECIDED something. Nothing was crowded out of a sweep
+        // that listed everything it found, and saying so there explains a mechanism the reader did not just hit.
+        // (Capped is never set under counts_only — that mode lists nothing — so it carries the mode gate too.)
+        // "spent on every other plugin BEFORE those" is a statement about an ordering between two groups, so it needs
+        // both groups to exist: on a sweep scoped to base masters alone there is no "every other plugin", and what the
+        // budget dropped is the caller's own requested findings (round-2 review).
+        if (r.Capped && r.BaselineDangling > 0 && r.PluginsScanned > (r.BaseMastersSwept?.Count ?? 0))
```
Avick3110: This gate compares the **scanned-plugin count** against the **examined-base-master count**, which is not the question the comment states ("it needs both groups to exist"). The two diverge whenever a base master is in `targets` but contributes nothing to `examined`.

Two reachable shapes:

- **Record scope filters a base master out entirely.** `plugins=["Skyrim.esm","Dawnguard.esm"], editorid_contains=<matches only Skyrim.esm>, limit=50` → `PluginsScanned=2`, `BaseMastersSwept=["Skyrim.esm"]`, `Capped=true`, `BaselineDangling>0` → `2 > 1` → the clause prints "the listing budget (limit=) is spent on every other plugin BEFORE those" over a scope that contains no other plugin.
- **A repeated name in `plugins=`.** `["Skyrim.esm","Skyrim.esm"]` → `targets.Count=2`, `examined` has 1. The branch already treats duplicate `plugins=` names as a real input shape (see the accumulate-never-assign comment in `ErrorCheck.cs`), so this is not a hypothetical.

`BASELINE-PHASE-CLAUSE-NEEDS-A-NON-BASE-PLUGIN` passes only because its fixture scopes to a single name, where the arithmetic happens to coincide.

The result carries no "were any non-base plugins in scope" fact, so the fix is a field computed where the answer lives — `targets.Any(p => !IsBaseMaster(p))` in `ErrorCheck.Run` — rather than a subtraction in the render. Same layer logic as the rest of this PR: the layer that knows, states it.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -1171,15 +1181,26 @@ public static string RenderCheckErrors(ErrorCheckResult r, int maxChars, int his
...
-                sb.Append("\n... [truncated at max_chars=").Append(cap).Append("; ").Append(SweepNarrowHint).Append("]\n");
+                // AN OMISSION SENTENCE IS COMPUTED BY THE LAYER THAT PERFORMED THE OMISSION, AND CLAIMS ONLY THAT
+                // LAYER'S OMISSION. This is the render's cut, so it is counted here, where the loop already knows
+                // what it emitted — and it says nothing about the listing budget, whose own omissions the capped
+                // line below reports in its own terms. The two truncators are independent and do not sum: crossing
+                // that seam is what made rounds 1 and 2 a single class (advisor ruling, 2026-08-18).
+                sb.Append("\n... [truncated at max_chars=").Append(cap).Append(": ")
+                  .Append(r.Reports.Count - sections).Append(" of ").Append(r.Reports.Count)
+                  .Append(" plugin section(s) were not rendered, and the last one rendered may be partial; ")
+                  .Append("this is the response being cut, separate from any limit= omission reported below. ")
```
Avick3110: Answering the review note directly: the wording is **not** evasive — each sentence is checkable and each disclaims the other layer, so the layer rule is holding. But the two cuts are stated in **different units**, and that is the residual gap, not the phrasing.

On the real order the reader gets `70 of 137 plugin section(s) were not rendered` and `the listing budget (limit=) omitted 3996 dangling ref(s); true total = 4996`. Subtracting gives 1000 listed — but 70 sections' worth of those 1000 are not in the response, and nothing says how many. "How many findings can I actually see" has no answer in either sentence, and `the last one rendered may be partial` is the same gap in miniature: the hedge exists because the loop is counting sections, not entries.

The render already has the number. Accumulating `p.Dangling.Count` over the sections it emitted lets the notice say "N of the M listed dangling entries appear above" — the render counting its own omission, in the unit the budget line uses. That is strictly inside the ruling (the omitting layer computes it) and does not put per-format accounting anywhere the `check` rebuild will have to unpick.

So: pull **that one number** forward, not render-aware accounting wholesale. Full shared-source accounting can still wait for PR 4.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -1246,14 +1246,39 @@ public static string RenderCheckErrors(ErrorCheckResult r, int maxChars, int his
             WriteExcluded(w, r.ExcludedPlugins);
             w.WriteBoolean("counts_only", r.CountsOnly);
 
+            // #344 — the baseline split as DATA (the text render's baseline line). base_masters names the set that was
+            // counted, because "baseline" is the whole claim and Mutagen's base set is not the same as the engine's
+            // force-loaded implicit set (Creation Club plugins are in the latter, not the former).
+            // base_masters_in_scope distinguishes "the baseline came back clean" from "no baseline was swept"; a
```
Avick3110: The comment explains `base_masters_in_scope`, but the writer emits `base_masters_swept` (L1264). A consumer following this comment greps for a key that does not exist.

```suggestion
            // base_masters_swept distinguishes "the baseline came back clean" from "no baseline was swept"; a
```

Same drift one level up: the class doc-comment at L1217 still lists the response as `{scanned_plugins, dangling, missing_masters, unscannable, classes, filter_note, off_order_scanned, excluded_plugins, plugins:[…], capped, rendered, truncated, boundary}` — none of `baseline_dangling`, `non_baseline_dangling`, `base_masters_swept`, `base_masters`, `dangling_by_source_plugin`, `dangling_not_listed_by_source`, or `plugins_with_findings` appears in it. Worth refreshing in the same touch, given this PR fixed an orphaned XML doc comment elsewhere.

Avick3110:
## Fold summary — your four findings

**3 folded, 1 filed, 0 refused.** Fold commit `b897497`. Directed fold, so no review round follows it (§11).

**F1 — silent last-section cut → filed as #361**, not folded: the inner break predates this branch, so §2 routes it to the issue lane. Your repro is in the issue verbatim, alongside a synthetic one (3 refs collected, 1 rendered, no notice, response 659 chars against a cap of 260, boundary block still appended). The render-cut notice now states its own reach in the code where it lives, citing #361, and the PR body's summary line carries the same qualification.

One correction to the triage premise I was given — **JSON is not silent on this path.** Its inner `dangling` loop has no cap check at all, so a plugin is written whole or not at all and `truncated` is set at the next plugin boundary. The JSON failure mode is the opposite one: a single large plugin makes the document overshoot `max_chars`. The issue states both, and notes that measuring the synthetic repro in JSON returns `truncated: true` for an unrelated reason (cap exhausted before the plugins array) rather than as evidence of the text bug.

**F2 — folded as directed.** `NonBaseInScope` is computed in `ErrorCheck.Run` from the resolved targets; the render consumes the fact. Your point about the old arm was exactly right — it passed because a single-name scope makes the two counts coincide. The new fixture is your second shape, `plugins=["Skyrim.esm","Skyrim.esm"]`: scanned 2, swept-base 1, so the old gate's test is `true` while the fact says otherwise. Both directions RED-checked, plus a positive arm so the gate cannot pass by suppressing the sentence everywhere.

**F3 — folded as ruled: one number forward, not wholesale.** Entries are counted at the append site inside the inner loop, so the figure survives a partial section; sabotaging it to a section sum turns the arm RED. The hedge is gone and its arm with it. Measured before committing:

```
... [truncated at max_chars=900: 10 of 12 plugin section(s) were not rendered; 4 of the 36
 budget-listed dangling entries appear above. This is the response being cut, separate from
 any limit= omission reported below. narrow with … or raise max_chars]
```

4 is the number of entry lines actually in that response.

**F4 — folded**, plus the class doc-comment's response-shape list refreshed in the same touch (it was missing seven fields). It now carries a note that it is a hand-kept second spelling outside `wire-names-guard`'s reach: if it drifts a second time, delete it rather than correct it again. No arm can hold a comment's accuracy, so that note is the mitigation rather than an arm that would only look like one.

**Ground for zero refusals:** F1, F2 and F4 each cite a line that reproduces as described — F1 and F2 re-measured here on synthetic fixtures before acting — and F3 is your ruling on a question this PR's own body asked.

`ci-all` 124/124; arms 30 → 33. Seven sabotages over the fold, each RED for its stated reason, each restored and verified, all run from the committed state per the §11 bullet in `eea7f41`.

Merge is yours.

Avick3110:
Correction to my own numbers, before you read the PR body: the guard arm count is **34 → 67 (+33)**, not "13 → 33". I carried a wrong base figure for several commits (it is also in a couple of commit messages) before checking it against `git show 5c5ebf8:src/housecarl-generator/CheckErrorsProbe.cs`. The arms-added figure was right; the starting point was not. PR body is corrected; nothing else in it depended on the number.

Issue #364, opened 2026-08-18
[GAP] compact/merge never mention the runtime INI distributor layer, whose lines address records as Plugin.esp|FormID


Avick3110:
## Summary

`RenderMerge`'s impact list reads as exhaustive: it states "only facegen/voice/seq were carried", names the *plugins* that reference or override donor records, warns about Papyrus `GetFormFromFile`, and warns about `.bsa` auto-loading. It never mentions the runtime INI distributor layer — SPID `_DISTR.ini`, KID `_KID.ini`, SkyPatcher, and OAR configs — whose lines address records as `Plugin.esp|FormID`.

Those lines are not FormLinks in any plugin, so the identify pass cannot see them; its own summary line says it "scanned N **plugin(s)**". Every distributor line naming the donor stops matching the moment the donor plugin is deactivated at the MO2 swap.

`RenderCompact` has the same silence.

## Failure scenario

```
housecarl_merge_plugins plugins=["MyFollowers.esp"] output="My Followers.esp"
```

`MyFollowers.esp`'s distribution is driven by a `_DISTR.ini` full of `MyFollowers.esp|0x801` lines. The report says "external referencers: none — nothing outside the merge points at a donor record." The user completes the swap. Nothing distributes in game, and no tool anywhere reported a problem — the one line that spoke to the question said the opposite.

## Why this is worth more than a sentence

houseCARL already knows this addressing model — `WriteTools.cs:258` and `:1240` tell callers a FormID is "how you reference this record (SkyPatcher/SPID)" — and it ships `housecarl_skypatcher_read` and `housecarl_skse_config_audit`. So the affected lines are **detectable**, not merely mentionable. The minimum honest fix is a sentence in the merge/compact reminders; the better one scans the distributor configs for lines naming the donors and names them the way the plugin referencers are named.

Distinct from #201 (which is about auditing distributor configs as a capability) — this is about the two verbs that invalidate those configs not saying so.

## Scope note

Merge-wide and compact-wide, not specific to any donor count. It is called out now because #345 made a single-donor merge the rename operation, and a rename is the case where the break is total (every line naming the old plugin) rather than partial.

## Provenance

Found by two independent reviewers during the pre-PR review rounds for #345, in a code path that branch does not touch.

Unrouted and unmilestoned — it lands at the next routing stroke.

Avick3110:
[...]

Issue #363, opened 2026-08-18
[Bug] merge_plugins drops the donor's ESL/ESM header flag, and the report never says so


Avick3110:
## Summary

`MergeBuild` creates the output as `new SkyrimMod(outKey, SkyrimRelease.SkyrimSE)` (`WritePatchBuilder.cs:2589`) with no header flags carried, and nothing downstream sets them. `CompactBuild` does set one (`:2499`, `{ IsSmallMaster = esl }`), so the asymmetry is visible in the same file. The merged/renamed output is therefore always a plain, non-light, non-master plugin, and the rendered report says nothing about the change.

For a many-donor merge that is defensible — "no ESL in v1" was an explicit scope call at PR #158, with the documented remedy being to run `compact_plugin` afterwards. What has changed is that #345 made **rename** a single-donor merge, and a rename of an ESL-flagged patch is now an ordinary thing to ask for.

## Failure scenario

A user near the 254-plugin cap renames a mis-named ESPFE patch:

```
housecarl_merge_plugins plugins=["MyPatch.esp"] output="Better Name.esp"
```

The donor is ESL-flagged; every one of its object ids already sits inside `0x800–0xFFF`. The output is a full-slot ESP. The user enables it, deactivates the donor, and has consumed a load-order slot — with no line in the report stating that the light flag was dropped. The only related text is the unconditional tail "Want it light? Run `housecarl_compact_plugin`", which reads as an option rather than as a consequence of what just happened.

The offered remedy is also not free: `compact_plugin` renumbers ids sequentially from `EslFloor`, so it moves the ids the rename kept.

The same drop applies to an `.esm`-flagged donor.

## Two shapes of fix, needing a call

1. **Report it** — name the dropped flag when a donor carried one. Cheap, honest, no behaviour change.
2. **Carry it** — for the single-donor case the donor's ids are already inside the light window by definition of it having been a valid ESL, so the flag could be carried by construction. This is a behaviour change and interacts with the "no ESL in v1" scope call, so it is a decision, not a fold.

## Related, lower

The `.esl` output refusal (`LoadOrderService.cs:6213`) states its reason as "a merge keeps the donors' object ids in the full range (ids above 0xFFF would be misread in game)". For an already-light donor that premise is false — all its ids are in-window. The refusal's remedy (merge to `.esp`, then compact) still works, so this is a wording defect on a newly-reachable path, worth folding into whatever fix this issue gets.

## Provenance

[...]

Issue #362, opened 2026-08-18
[Bug] merge_plugins / compact_plugin blank every localized string — donors open with the bare overlay, bypassing OpenOverlay


Avick3110:
## Summary

`housecarl_merge_plugins` and `housecarl_compact_plugin` open their donor/source plugins with the **bare** `SkyrimMod.CreateFromBinaryOverlay(path, release)`. That overload only scans the plugin's **own folder** for `Strings\`, so a localized plugin whose `.STRINGS` are served from a different mod folder reads every `TranslatedString` as empty — and the output is then written with those empty values baked in. Every FULL / DESC in the plugin is blanked, with no warning.

## Why this is confidently a bug, not a design choice

The codebase already documents this exact class and fixes it everywhere else. `LoadOrderResolver.OpenOverlay` exists precisely for it (HCBR-2026-06-24), and `WritePatchBuilder.cs:1043-1047` names the hazard in a comment before routing through it:

> a plugin resolved to a folder without its own strings source reads every `TranslatedString` EMPTY (the HCBR-2026-06-24 class `OpenOverlay` exists for)

The two plugin-surgery verbs are the ones that do not route through it:

- `WritePatchBuilder.cs:2599` — `MergeBuild`, per donor
- `WritePatchBuilder.cs:2497` — `CompactBuild`, the source plugin

`OpenOverlay(path, dataDir)` is a pure factory and both call sites have a resolver available, so the fix looks like a call-site change rather than a redesign — but it needs its own verification, including what `dataDir` should be in the write lane.

## Repro shape

A localized donor whose `Strings\*.STRINGS` live in a *separate* enabled mod folder — the "Cleaned Base Game Masters" / translation-mod pattern the `OpenOverlay` doc calls near-universal:

```
housecarl_merge_plugins plugins=["Localized.esp", "Other.esp"] output="Merged.esp"
```

→ succeeds, no warning, and every `Name` / `Description` in the output is `''`. The same applies to `housecarl_compact_plugin` on such a plugin, and to a single-donor merge (a rename).

## Impact

Silent data loss in a write verb — a Q3 violation. The user gets a plugin that loads fine and has lost every localized name and description, discoverable only in-game or by diffing.

## Provenance

[...]

Issue #361, opened 2026-08-18
[Bug] check_errors: a max_chars cut inside the LAST report section is silent — no truncation notice, and the boundary block appends past the cap


Avick3110:
## Summary

When `max_chars` is reached **inside the last report section**, `check_errors`' text render stops emitting dangling entries and says nothing at all. `truncated` is never set, so no `[truncated at max_chars=…]` notice appears, and `AppendExcludedPlugins` — gated on `!truncated` — still appends past the cap.

The per-plugin dangling loop breaks on the cap without setting the flag (`src/housecarl-mcp/ReadTools.cs`, the inner `foreach (var d in p.Dangling)`); the outer `foreach` only sets `truncated` when it starts another iteration. For any section that is not the last, the next iteration fires the notice. For the last one, the loop simply ends.

## Repro (Aaron's, on a real order)

`plugins=["BigMod.esp"], limit=5000` where `BigMod.esp` carries 3000 dangling refs, default `max_chars`. The header prints `dangling reference(s) (3000):`, roughly 600 entries print, then the boundary block — no truncation notice anywhere in the response.

## Repro (synthetic, measured 2026-08-18)

Single-plugin scope, `limit=5000`, `max_chars=260`:

```
collected refs=3   entry lines rendered=1
has truncation notice=False   boundary block still appended=True   response length=659 (cap 260)
```

Two findings dropped, no sentence of any kind, and the response runs 2.5× past the cap it was given.

## `format='json'` behaves differently — worth stating, because the obvious assumption is wrong

JSON is **not** silent on this path, and does not drop entries mid-section. Its inner `dangling` loop has no cap check at all (`src/housecarl-mcp/JsonWire.cs`), so a plugin is written whole or not at all, and `truncated` is set at the next plugin boundary. The JSON failure mode is therefore the opposite one: a single large plugin makes the document **overshoot `max_chars`** rather than silently lose entries.

(Measuring the synthetic repro above in JSON returns `truncated: true`, but for an unrelated reason — the cap was already exhausted before the plugins array began. It is not evidence of the text render's bug.)

## Why it is filed rather than fixed

The inner break is pre-existing and untouched by PR #360, so per CLAUDE.md §2 it belongs in the issue lane rather than in that branch's fold. It is filed before #360 merges because it is the one path that branch's new render-cut notice claims to cover and does not — the notice fires at section boundaries only, which is now stated in the code where the notice lives, citing this issue.

## Suggested shape

[...]

PR #365, opened 2026-08-18
merge_plugins: a single donor is accepted, and it is the rename

### LINKED ISSUE #345: No way to rename a single existing patch — merge_plugins requires 2+ distinct donors and refuses an output name matching either
## Summary

There's no dedicated rename operation for an existing houseCARL patch. `housecarl_merge_plugins`
looked like the natural tool for this (create a new plugin, "merge" the old one's content into
it), but it refuses on two counts: it requires at least two *distinct* donor plugins, and it
refuses an `output=` name that collides with any donor — so it can't be used to consolidate a
single mis-named patch into a correctly-named one, even with a dummy second donor.

## Concrete repro
housecarl_merge_plugins plugins=["Patch.esp"] output="New Name.esp"
→ error: merge needs at least TWO distinct donor plugins — pass plugins=["A.esp", "B.esp", …]

Creating a second (empty, placeholder) plugin via `housecarl_create_plugin` and merging both in
doesn't work either, if the desired output name matches the placeholder:
housecarl_merge_plugins plugins=["Patch.esp", "New Name.esp"] output="New Name.esp"
→ error: the output 'New Name.esp' cannot also be a donor — name a NEW plugin file.

The only working path I found was fully replaying every edit (every `create_record`/`bulk_apply`/
`set_field` call) into a freshly `create_plugin`-named plugin, then manually removing the old mod
folder.
## Why this matters
Combined with the auto-naming issue above, getting a patch's name wrong on the first attempt
currently has no cheap fix — it's a full manual re-do of every edit, which for a patch with many
operations is real, avoidable work.
## Suggested fix
Either let `merge_plugins` accept a single donor (functioning as a pure rename when there's only
one), or add a dedicated `housecarl_rename_plugin` operation that re-targets a patch's stored
name/file without requiring a full record replay.
## Environment
- houseCARL v1.9.0

Avick3110:
Closes #345.

`housecarl_merge_plugins` refused fewer than two distinct donors, and nothing else on the surface renames a plugin — so a mis-named patch could only be fixed by replaying every edit into a newly named plugin. One donor is now accepted, and that is the rename.

## The binding pre-fold check: no reason exists

The charter required confirming there was no undocumented reason for the two-donor guard before touching it. Seven artifacts, none stating one:

| Artifact | What it says about the minimum |
|---|---|
| Birth commit `00cee8a` (message + diff) | Lists `>=2 donors` inside a validation enumeration. **No rationale.** Its asset note argues the other way: "a merge renames the plugin and the plugin NAME is a folder segment of facegen/voice paths, so EVERY donor NPC/voiced line carries" |
| `COMPACT_MERGE_PLAN_2026-06-26.md` §5 | Merge's refuse-loud list is the ESL ceiling, sub-model records, an unparseable donor. **No donor-count refusal** |
| `MERGE_REFERENCE_RESEARCH_2026-06-26.md` | Full zMerge/xEdit source dive. **No minimum-donor concept in any reference tool** |
| Handoff 2026-06-27 | Locked sub-decisions: winner+report, de-scope edit-combining, direction deferred. **No donor count** |
| Handoff 2026-07-07 (PR #158) | The three A4 scope calls: WARN-and-proceed, no ESL in v1, override auto-fix out. **No donor count.** Notes "RenumberModInto ≈ the single-donor reg=null case" |
| `git log -L 6185,6200` | The guard has **exactly one commit** — its birth. Never revisited |
| Repo-wide grep (.cs/.md/.json) | No comment, doc or CHANGELOG line justifies it. Only the probe pinned it, pinning *behaviour* |

It sat under `// ---- 0. argument shape ----`, which is all it ever was.

## Verified, not inherited

The triage claim that the machinery is single-donor-correct was checked against source. `BuildMergeRemap` writes every originating key into the dict on the keep path, so a lone donor's records move to the output ModKey holding their object ids — the collision set is simply empty. The facegen/voice carry and the `.seq` gate already loop per **donor**, because the plugin NAME is a folder segment of those paths, which is the one thing a rename changes.

## Decisions this PR pins

- **The count guard moved from 2 to 1, it did not disappear.** Zero donors still refuses, and the refusal names both shapes.
- **The donor list stays a SET.** `plugins=["A.esp","A.esp"]` is one donor and renames. Duplicates already collapsed silently for many donors (`["A","A","B"]` merges two today), so making duplicates special *only* for one donor would be machinery the multi-donor path does not have. The same reasoning covers a blank or null element in the list: the trim/drop is pre-existing and uniform, and the count boundary is the only thing that moved.
- **The report states which operation it performed**, deciding the shape once and consuming it at the sentences that vary — the headline ("from 1 donors" is both ungrammatical and a misdescription), the renumber cause, the remedy ordering, the folder default. Every one of those has arms on both sides. See the directed completion below for how that scope was arrived at.

## Review rounds

**Round 1 — 3 fresh agents, isolated worktrees. 12 findings: 5 folded, 2 refused, 3 to the issue lane, 2 low declined.**

All three converged on one class — prose claiming more than the write does:

- **"object ids kept (nothing can collide)"** was the headline promise and is false: `BuildMergeRemap` keeps an id only inside the write window, so a below-floor id renumbers with no collision anywhere. Folded, **and given an arm**. That arm runs against `BuildMergeRemap` directly: a donor carrying a sub-`0x800` originating record cannot be written at all (`LowerFormKeyRangeDisallowedException`, verified by trying to fixture one), so the service-level test the claim appeared to need does not exist. One agent reported having built that fixture; that account did not hold up. Such plugins come from other tools and houseCARL reads them fine, so the case is reachable even though the fixture is not.
- **The CHANGELOG and doc comment invited removing the old mod folder** while the tool's own output says to keep it enabled — and the output is right, because the renamed records still load that mod's assets by path. Both now match the report.
- **The saves reminder blamed "new FormIDs".** Rename a pure-override patch and no FormID moves at all, so the report contradicted itself in one response. The conclusion was never in doubt; the reason now matches it.
- **The `.seq` note named only one of two ways a quest gains an entry.** Broadened for both arms rather than made conditional.
- **Removed a vacuous arm** asserting a one-donor merge reports zero conflicts — gutting the conflict machinery entirely left it green.

Refused: *"the saves/swap arm has no coverage"* — suppressing the reminder for the one-donor arm turns it RED; reproduced. And the merge-vocabulary nits (body sentences still saying "merged"; the folder named `<output> merged`) — true, no action changes, and each fix costs a conditional plus arms in both directions.

**Round 2 — 2 fresh agents. 10 findings: 2 folded on the branch (both test-evidence), 7 dispositioned by the directed completion below, 1 left with #363.**

Both folds were arms that claimed more than they verified:

- **The set-semantics arm could not see the comparer.** It passed two identical strings, which `Distinct()` collapses under any comparer. Replacing `OrdinalIgnoreCase` with the default left the whole guard green while `["HcMgA.esp","HCMGA.ESP"]` merged a plugin with itself, reported nine self-conflicts and announced "2 donors". A case-differing duplicate now pins it.
- **The remap arm was labelled "every A record" and read back four of eight keys.** Its negative is satisfied just as well by a record that was *dropped*, so a walk regression losing a nested INFO or the celled placed ref would have left every RENAME arm green.

## Class-stop, and the directed completion that answered it

**Round 2 returned round 1's class, so there was no round 3 and no third instance was folded.** The recurring class: *the rename inherits merge-shaped text and merge-shaped losses* — a report written for "combine N plugins" also serving an operation that is not a combination. That went to the advisor lane, which ruled a **directed completion**: kill the class by classifying the operation's shape **once**, rather than folding instances.

**The escape hatch ran first, as the ruling required.** Counting the shape-varying sites in `RenderMerge` — every sentence asked "does this differ between rename and combine?" — found **two beyond the headline pair**: the per-donor renumber cause, and the external-referencer remedy order. Masters, the swap instruction, the asset carries and the saves reminder do not vary and were left alone. At two sites a classifier type would be over-machinery, **so the hatch fell to sentence-local** — with the shape bound once into a local and consumed at each site, which is the ruling's "decide once" structure at the scale the count justifies, rather than count tests scattered per sentence.

Disposition of the seven, as directed:

| # | Instance | Done |
|---|---|---|
| 1 | ESL flag dropped | **Report line rides this PR**; the carry decision stays #363. The remedy it names was measured first (`compact_plugin`'s `esl` defaults true) and carries its own cost — it renumbers from `0x800`, moving the ids reported kept two lines above |
| 2 | `Author`/`Description` dropped | **Report line**, stating the loss bare. **No remedy named** — none exists (`author=`/`description=` are `create_plugin`'s, at create time only, verified). No issue number in tool output |
| 3 | "records under a new identity" | Headline is now **derived from the accounting line below it**, so a pure-override donor — what most mis-named patches are — says it originates no records rather than claiming a re-keying that did not happen |
| 4 | Two renumber causes | Shape-true: one donor names only "(below-floor)" |
| 5 | Remedy ordering | A rename leads with re-point / rebuild — the remedies that keep it a rename — and offers combining second. Combine's ordering is unchanged |
| 6 | `.esl` refusal wording | Stays with #363, as filed |
| 7 | `<output> merged` folder | Default is `<output> renamed` for one donor, and the `patch_name=` `[Description]` naming it moved in the same change. **Checked first that nothing consumes the old default** — no probe, no `meta.ini` breadcrumb, no doc; its only consumer was its own call site |

**Refusal accounting is unchanged by the completion** — it folded the seven the ruling dispositioned and nothing else. Round 1's two refusals still stand, and the merge-vocabulary nits refused there stay refused on their own ground; the completion changed the economics for the *shape-varying* sentences only, which is what it was scoped to.

**What was never in doubt:** five agents across two rounds found no defect in what gets *written* on the single-donor path. Every finding was in the report or the header metadata.

## Filed to the issue lane (outside this branch, the #324 shape)

- **#362 — high.** `merge_plugins` / `compact_plugin` blank every localized string: donors open with the bare overlay, bypassing `LoadOrderResolver.OpenOverlay`. The codebase documents this exact hazard at `WritePatchBuilder.cs:1043-1047` and routes *that* lane through the fix; `MergeBuild:2599` and `CompactBuild:2497` do not. Silent data loss in a write verb, pre-existing on both.
- **#363** — the ESL/ESM header flag drop, plus the `.esl` refusal wording.
- **#364** — compact/merge never mention the runtime INI distributor layer (SPID/KID/SkyPatcher/OAR), whose lines address records as `Plugin.esp|FormID` and which the identify pass cannot see.

All three unrouted and unmilestoned, per the #361 precedent.

## Evidence

Guard `merge-service-guard` **23 → 56 arms**; `ci-all` **124/124**.

**Twenty-two sabotages** across the engagement, each run from a committed state, each restored with the restore verified by grepping for a string the fold introduced. Every conditional the completion added was collapsed **both ways**, and the multi-donor checks are the negatives:

| Completion sabotage | Result |
|---|---|
| Shape classifier forced to RENAME for every merge | RED — 4 multi-donor arms |
| Shape classifier forced to COMBINE for every merge | RED — 7 rename arms |
| Headline claims re-keying when nothing originates | RED — pure-override arm |
| Headline claims no re-keying when records moved | RED — derived-claim arm |
| Folder default forced to `merged` for a rename | RED — rename folder arm |
| Folder default forced to `renamed` for a combine | RED — merge folder arm |
| Light-flag loss never detected | RED — ESL note arm |
| Light-flag loss reported for every donor | RED — merge no-notes arm |
| Header-text loss never detected | RED — header-text arm |
| Header-text loss reported for every donor | RED — merge no-notes arm |
| Compact remedy's cost clause dropped from the ESL note | RED — ESL note arm |

The eleven from the rounds:

| Sabotage | Result |
|---|---|
| Force the multi-donor headline for one donor | RED — rename headline arm |
| Force the rename headline for many donors | RED — multi-donor headline arm |
| Restore the old `>= 2` guard | RED — 14 of 15 RENAME arms |
| Drop the zero-donor refusal | RED — all three refusal arms |
| Facegen carry made multi-donor-only | RED — rename facegen arm |
| Voice carry / seq gate made multi-donor-only | RED — rename voice, rename seq |
| Identify pass given an empty target set | RED — both rename external-referencer arms |
| Drop the donor `Distinct()` entirely | RED — duplicate-name arm |
| Replace the `Distinct` comparer with the default | RED — the case-differing arm, reporting "2 donor, 9 conflicts" |
| Planner keeps below-floor ids | RED — below-floor arm |
| Lose a nested INFO on the rename path only | RED — the remap arm, among 12 |

One defect in the probe itself was caught by this and fixed on the branch: restoring the old guard originally turned only 5 arms RED and then *stopped the block*, because it threw on the refusal outcome's null `OutputPath`. The arms past that point are now gated and short-circuit, so the same sabotage reports 14. A check that cannot report is not a check.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -12,14 +12,18 @@ when it changes.
   behind that refusal: with one donor the collision set is simply empty, and the machinery a rename needs is the
   machinery a merge already runs — every donor record moves to the output's identity, and the facegen, voice, and
   `.seq` files follow the plugin **name** that forms part of their paths. Passing one donor now does exactly that,
-  keeping the object ids, and the report calls it a rename rather than a merge "from 1 donors". Zero donors still
+  keeping every object id already inside the writable range — an id below the `0x800` floor still renumbers, which
+  nothing can collide with either, and the per-donor line reports it — and the report calls it a rename rather than a
+  merge "from 1 donors". Zero donors still
   refuses, and the refusal names both shapes. The list of donors is a set, so a plugin named twice is one donor —
   also a rename.
 
   The costs of renaming are inherent, so they are reported rather than refused: the renamed plugin arrives in a new
-  mod folder beside the original (removing the old one is your call), any save that referenced the old plugin name
-  will not survive — the existing saves warning — and any plugin that references or overrides the donor, **its own
-  patches included**, is named in the warnings with the remedy, exactly as for a many-donor merge. (#345)
+  mod folder beside the original, and the swap instruction applies unchanged — deactivate the old **plugin**, but keep
+  its mod **folder** enabled, because the renamed records still load that mod's meshes, textures and scripts by path.
+  Any save that referenced the old plugin name will not survive — the existing saves warning — and any plugin that
+  references or overrides the donor, **its own patches included**, is named in the warnings with the remedy, exactly
+  as for a many-donor merge. (#345)
 
 - **Fixed: on a large load order, a mod's broken references could vanish from `housecarl_check_errors` entirely.**
   The dangling-reference listing budget (`limit=`, default 1000) is one counter spent plugin by plugin in load order,
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -22,8 +22,21 @@ when it changes.
   mod folder beside the original, and the swap instruction applies unchanged — deactivate the old **plugin**, but keep
   its mod **folder** enabled, because the renamed records still load that mod's meshes, textures and scripts by path.
   Any save that referenced the old plugin name will not survive — the existing saves warning — and any plugin that
-  references or overrides the donor, **its own patches included**, is named in the warnings with the remedy, exactly
-  as for a many-donor merge. (#345)
+  references or overrides the donor, **its own patches included**, is named in the warnings with the remedy.
+
+  The report now says what is true **of the operation you actually got**, because it decides which one that was in
+  one place. A rename names only the renumber cause that can apply to a single donor, since nothing can collide with
+  it; it leads its warnings with the remedy that keeps the result a rename — re-point the affected plugin, or rebuild
+  it — and offers combining second, because adding that plugin as a donor produces a combined plugin instead; and its
+  new mod folder is named `<output> renamed` rather than `<output> merged`. Renaming a plugin that originates no
+  records of its own, which is what most mis-named patches are, says exactly that instead of claiming records took a
+  new identity.
+
+  Two losses that were previously silent are now stated, for merges of any size. **A donor's light (ESL) header flag
+  is not carried**, so the output takes a full load-order slot where a light donor took none — the report says so and
+  names `housecarl_compact_plugin` along with its cost, that it renumbers object ids from `0x800` upward and so moves
+  the ids just reported as kept. **A donor's header Author and Description are not carried either**, and that is
+  stated plainly with no remedy offered, because none exists. (#345)
 
 - **Fixed: on a large load order, a mod's broken references could vanish from `housecarl_check_errors` entirely.**
   The dangling-reference listing budget (`limit=`, default 1000) is one counter spent plugin by plugin in load order,
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -6,6 +6,21 @@ when it changes.
 
 ## Unreleased
 
+- **`housecarl_merge_plugins` now accepts a single donor, which renames a plugin.** Merging one plugin into a new
+  name was refused ("merge needs at least TWO distinct donor plugins"), and no other tool renames one, so a
+  mis-named patch could only be fixed by replaying every edit into a freshly named plugin. There was never anything
+  behind that refusal: with one donor the collision set is simply empty, and the machinery a rename needs is the
+  machinery a merge already runs — every donor record moves to the output's identity, and the facegen, voice, and
+  `.seq` files follow the plugin **name** that forms part of their paths. Passing one donor now does exactly that,
+  keeping the object ids, and the report calls it a rename rather than a merge "from 1 donors". Zero donors still
+  refuses, and the refusal names both shapes. The list of donors is a set, so a plugin named twice is one donor —
+  also a rename.
+
+  The costs of renaming are inherent, so they are reported rather than refused: the renamed plugin arrives in a new
+  mod folder beside the original (removing the old one is your call), any save that referenced the old plugin name
+  will not survive — the existing saves warning — and any plugin that references or overrides the donor, **its own
+  patches included**, is named in the warnings with the remedy, exactly as for a many-donor merge. (#345)
+
 - **Fixed: on a large load order, a mod's broken references could vanish from `housecarl_check_errors` entirely.**
   The dangling-reference listing budget (`limit=`, default 1000) is one counter spent plugin by plugin in load order,
   and the base-game masters sit at index 0. A plugin whose findings collected an empty list was then dropped from the
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -20,9 +20,12 @@ when it changes.
 
   The costs of renaming are inherent, so they are reported rather than refused: the renamed plugin arrives in a new
   mod folder beside the original, and the swap instruction applies unchanged — deactivate the old **plugin**, but keep
-  its mod **folder** enabled, because the renamed records still load that mod's meshes, textures and scripts by path.
-  Any save that referenced the old plugin name will not survive — the existing saves warning — and any plugin that
-  references or overrides the donor, **its own patches included**, is named in the warnings with the remedy.
+  its mod **folder** enabled, because the renamed records still load that mod's *loose* meshes, textures and scripts
+  by path. Anything the donor ships inside a **`.bsa`** is the exception, and the report says so: an archive stops
+  loading once its same-named plugin is deactivated, so extract it into the mod folder with `housecarl_bsa_extract`
+  or keep it loading behind a same-named dummy plugin from `housecarl_create_plugin`. Any save that referenced the old
+  plugin name will not survive — the existing saves warning — and any plugin that references or overrides the donor,
+  **its own patches included**, is named in the warnings with the remedy.
 
   The report now says what is true **of the operation you actually got**, because it decides which one that was in
   one place. A rename names only the renumber cause that can apply to a single donor, since nothing can collide with
@@ -32,11 +35,15 @@ when it changes.
   records of its own, which is what most mis-named patches are, says exactly that instead of claiming records took a
   new identity.
 
-  Two losses that were previously silent are now stated, for merges of any size. **A donor's light (ESL) header flag
-  is not carried**, so the output takes a full load-order slot where a light donor took none — the report says so and
-  names `housecarl_compact_plugin` along with its cost, that it renumbers object ids from `0x800` upward and so moves
-  the ids just reported as kept. **A donor's header Author and Description are not carried either**, and that is
-  stated plainly with no remedy offered, because none exists. (#345)
+  Losses that were previously silent are now stated, for merges of any size, because the merged plugin is built with a
+  fresh header and nothing in a donor's header comes along. **Light (ESL) status is not carried** — counted by the
+  header flag *or* an `.esl` extension, since the engine treats the extension as light either way — so the output takes
+  a full load-order slot where a light donor took none. The report says so and names `housecarl_compact_plugin` along
+  with its cost, that it renumbers object ids from `0x800` upward and so moves the ids just reported as kept; where
+  that note appears, the closing "want it light?" line steps aside, so the recommendation you read carries its cost.
+  **Master status is not carried either** — counted by the flag or an `.esm` extension — so an esmified donor comes
+  back as a plain plugin and sorts in the plugin block. **Nor are the header Author and Description.** Those last two
+  are stated with no remedy offered, because the tool surface has none. (#345)
 
 - **Fixed: on a large load order, a mod's broken references could vanish from `housecarl_check_errors` entirely.**
   The dangling-reference listing budget (`limit=`, default 1000) is one counter spent plugin by plugin in load order,
--- a/src/housecarl-core/WritePatchBuilder.cs
+++ b/src/housecarl-core/WritePatchBuilder.cs
@@ -2147,7 +2147,8 @@ public static class WritePatchBuilder
         int PluginsScanned, int UnscannableRecords, IReadOnlyList<string> UnscannableSamples,
         long Bytes, string? Note = null,
         AssetRenameOutcome? AssetRename = null, VoiceCarryOutcome? VoiceRename = null, SeqRegenOutcome? SeqRegen = null,
-        IReadOnlyList<string>? LightDonors = null, IReadOnlyList<string>? HeaderMetaDonors = null)
+        IReadOnlyList<string>? LightDonors = null, IReadOnlyList<string>? HeaderMetaDonors = null,
+        IReadOnlyList<string>? MasterDonors = null)
     {
         public static MergeOutcome Fail(string error) =>
             new(false, error, "", "", Array.Empty<string>(), Array.Empty<string>(), 0, 0,
@@ -2562,7 +2563,8 @@ public static class WritePatchBuilder
     public sealed record MergeBuildResult(
         bool Success, string? Error, IReadOnlyList<string> Masters, int RecordsCopied, int RecordsRenumbered,
         IReadOnlyList<RemapEngine.MergeConflict> Conflicts, long Bytes,
-        IReadOnlyList<string>? LightDonors = null, IReadOnlyList<string>? HeaderMetaDonors = null)
+        IReadOnlyList<string>? LightDonors = null, IReadOnlyList<string>? HeaderMetaDonors = null,
+        IReadOnlyList<string>? MasterDonors = null)
     {
         public static MergeBuildResult Fail(string error) =>
             new(false, error, Array.Empty<string>(), 0, 0, Array.Empty<RemapEngine.MergeConflict>(), 0);
@@ -2593,6 +2595,7 @@ public static class WritePatchBuilder
         var donorSet = new HashSet<ModKey>(donorsByLoadOrder.Select(d => d.Key));
         var overlays = new List<IDisposable>();
         var lightDonors = new List<string>();                              // donors carrying header flags/fields the output will not
+        var masterDonors = new List<string>();
         var headerMetaDonors = new List<string>();
         try
         {
@@ -2611,7 +2614,15 @@ public static class WritePatchBuilder
                 // behind. Measured here, while the overlay is open, so the report can state the loss instead of the
                 // caller discovering it (a light donor silently costing a full load-order slot). Whether these should
                 // be CARRIED is a separate decision that is not this lane's to make.
-                if (ov.IsSmallMaster) lightDonors.Add(name);
+                //
+                // LIGHT and MASTER are each read BOTH ways, because either alone under-reports the loss. The header
+                // bit is not the whole model: this tool's own .esl-output refusal states that the engine force-treats
+                // the extension as light regardless of the bit, so a .esl donor with the bit unset still loses light
+                // status here. Symmetrically an esmified .esp carries the master bit without the extension, and a
+                // .esm carries the extension; both lose master status in a bare output.
+                if (ov.IsSmallMaster || ov.ModKey.Type == ModType.Light) lightDonors.Add(name);
+                if (ov.ModHeader.Flags.HasFlag(SkyrimModHeader.HeaderFlag.Master) || ov.ModKey.Type == ModType.Master)
+                    masterDonors.Add(name);
                 if (!string.IsNullOrWhiteSpace(ov.ModHeader.Author) || !string.IsNullOrWhiteSpace(ov.ModHeader.Description))
                     headerMetaDonors.Add(name);
             }
@@ -2689,7 +2700,7 @@ public static class WritePatchBuilder
         }
         catch { /* best-effort read-back; the union is a correct superset */ }
         return new MergeBuildResult(true, null, writtenMasters, mr.RecordsCopied, mr.RecordsRenumbered, mr.Conflicts, bytes,
-            lightDonors, headerMetaDonors);
+            lightDonors, headerMetaDonors, masterDonors);
     }
 
     /// <summary>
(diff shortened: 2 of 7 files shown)


src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2603,6 +2607,13 @@ public static MergeBuildResult MergeBuild(
                 }
                 overlays.Add((IDisposable)ov);
                 mods.Add((name, ov));
+                // The merged plugin is built as a bare SkyrimMod, so anything living in a donor's HEADER is left
+                // behind. Measured here, while the overlay is open, so the report can state the loss instead of the
+                // caller discovering it (a light donor silently costing a full load-order slot). Whether these should
+                // be CARRIED is a separate decision that is not this lane's to make.
+                if (ov.IsSmallMaster) lightDonors.Add(name);
```
Avick3110: **The header-loss measurement is narrower than the loss it exists to report.** The output is a bare `new SkyrimMod(outKey)`, so *every* donor header flag is left behind — but only `IsSmallMaster` is recorded here, and only as a header bit.

Two misses:

- **The ESM master flag.** Rename an ESM-flagged donor (a `.esm`, or an esmified `.esp` — normal in a big order) and the output is written as a plain plugin: it loses master status and its load-order block, and the report, which *does* warn when the light flag is dropped, says nothing. #363 names this drop as "the ESL/ESM header flag drop" — the carry decision is deferred, but the report line shipping here covers only half of what it already measures.
- **`.esl` by extension.** `IsSmallMaster` is the header bit. `MergePlugins`' own `.esl`-output refusal argues that "the game engine force-treats [.esl] as a LIGHT master regardless of the header flag" — so a `.esl` donor whose bit is unset takes a full slot with no note, against this file's own stated model.

`ov.ModKey.Type` (plus `ov.IsMaster`) covers both without a second read.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -1059,8 +1063,30 @@ internal static string RenderMerge(WritePatchBuilder.MergeOutcome o)
...
+            // DERIVED from it rather than asserting alongside it. A pure-override donor — the mis-named patch this
+            // capability exists for — originates nothing, so nothing is re-keyed, and claiming "its records under a
+            // new identity" would be false in the very case the feature was asked for.
+            sb.Append("wrote ").Append(file).Append(" (new plugin; ").Append(o.Bytes).Append(" bytes) — a RENAME of ")
+              .Append(o.Donors[0]).Append(": one donor, so there is nothing to combine — ");
+            if (o.RecordsRenumbered > 0)
+                sb.Append(o.RecordsRenumbered).Append(o.RecordsRenumbered == 1 ? " record moves" : " records move")
+                  .Append(" to the new plugin's identity.\n");
+            else
+                sb.Append("it originates no records of its own, so nothing is re-keyed; the same overrides are now ")
```
Avick3110: **This branch is chosen on `RecordsRenumbered`, but asserts something about `RecordsCopied`.** The comment above says the sentence is derived from the accounting; it is derived only halfway — "the same overrides" is asserted, not read.

A donor with **no records at all** takes this arm: the headline says "the same overrides are now served by a plugin under a new name" while the accounting line below prints `0 originating records merged under X.` with no override clause at all. Not hypothetical — an empty dummy plugin is exactly what the swap sentence in this same report tells the user to create (`housecarl_create_plugin`) to keep a donor `.bsa` loading, and renaming one is a plausible next step.

`overrides = o.RecordsCopied - o.RecordsRenumbered` is already computed at line 1099; a zero-copied donor wants a third sentence, or the clause dropped.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -1099,17 +1127,26 @@ internal static string RenderMerge(WritePatchBuilder.MergeOutcome o)
         if (o.ExternalPlugins.Count > 0)
         {
             sb.Append("WARNING — ").Append(o.ExternalPlugins.Count).Append(" plugin(s) OUTSIDE the merge REFERENCE donor records. Their references break ")
-              .Append("the moment you deactivate the donor plugins: include them in the merge set (re-run with them added), or re-point them at '")
-              .Append(o.OutputName).Append("' before the swap:\n");
+              // Adding the patch to the donor set yields a COMBINED plugin — a different operation from the one that
+              // was asked for — so a rename leads with the remedy that keeps it a rename, and offers combining second.
+              .Append(isRename
```
Avick3110, line 1143: **For the shape this feature is advertised for, these rename-specific remedies can never fire.** A pure-override donor originates nothing, so `plan.Dict` is empty, so `targets` is empty, so `IdentifyExternalReferencers` returns nothing by construction — the report always takes the `else` arms ("external referencers: none — nothing outside the merge points at a donor record").

That sentence is literally true (nothing *references* a record the donor doesn't own), but the pass only scans record links and record identity — never **declared masters**. Rename `MyPatch.esp` while `MyPatch Compat.esp` lists it as a master without referencing any of its records: the report reports zero dependents, the user follows the swap instruction and deactivates the old plugin, and the compat plugin is now missing a master — the game refuses to load.

Pre-existing for multi-donor merges (the #324 shape, and #364's INI layer is the same blind spot), so this is a note rather than a blocker — but the rename lane promotes the empty-`targets` case from an edge to the headline case, and the PR body calls a pure-override patch "what most mis-named patches are".

src/housecarl-mcp/WriteTools.cs
```diff
@@ -1122,9 +1159,33 @@ internal static string RenderMerge(WritePatchBuilder.MergeOutcome o)
...
+            sb.Append(" are not carried across; ").Append(o.OutputName).Append("'s are empty.\n");
+        }
+
         if (o.Note is { } note) sb.Append("note: ").Append(note).Append('\n');
-        sb.Append("reminders: existing SAVES that depend on the donors will not survive the swap (records moved to a new ")
-          .Append("plugin name + new FormIDs) — best for a new game. FormIDs compiled into Papyrus (.pex hardcoded / ")
+        sb.Append("reminders: existing SAVES that depend on the donors will not survive the swap (the records now live under a ")
+          .Append("different plugin name, and any id that had to be renumbered moved with it) — best for a new game. ")
+          .Append("FormIDs compiled into Papyrus (.pex hardcoded / ")
           .Append("GetFormFromFile) and any Mutagen-delta residual are NOT remappable — verify scripted records. Want it ")
```
Avick3110: **The closing line un-qualifies the advice the new NOTE just qualified.** The ESL note names `compact_plugin` *with* the consequence that makes it honest — it renumbers from `0x800`, so the ids reported kept two lines above will move. Ten lines later this reminder names the same tool flat: "Want it light? Run housecarl_compact_plugin on 'X' (the tools compose)."

For a light donor the report now recommends compact twice, and the version the user reads last is the one without the cost.

plugin/CHANGELOG.md
```diff
@@ -6,6 +6,38 @@ when it changes.
...
+  `.seq` files follow the plugin **name** that forms part of their paths. Passing one donor now does exactly that,
+  keeping every object id already inside the writable range — an id below the `0x800` floor still renumbers, which
+  nothing can collide with either, and the per-donor line reports it — and the report calls it a rename rather than a
+  merge "from 1 donors". Zero donors still
+  refuses, and the refusal names both shapes. The list of donors is a set, so a plugin named twice is one donor —
+  also a rename.
+
+  The costs of renaming are inherent, so they are reported rather than refused: the renamed plugin arrives in a new
+  mod folder beside the original, and the swap instruction applies unchanged — deactivate the old **plugin**, but keep
+  its mod **folder** enabled, because the renamed records still load that mod's meshes, textures and scripts by path.
```
Avick3110: **The `.bsa` caveat that the report and the tool description both carry is missing here.** `RenderMerge` says: "If a donor ships a .bsa, it stops auto-loading once its plugin is deactivated — extract it into the mod folder (housecarl_bsa_extract) or load it via a same-named dummy plugin."

This paragraph states the swap rule without it, and gives as its reason the exact claim the BSA case falsifies — "the renamed records still load that mod's meshes, textures and scripts by path". A user who reads the release note and not the tool output deactivates the old plugin, keeps the folder enabled as instructed, and loses every asset packed in the donor's archive.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -1122,9 +1159,33 @@ internal static string RenderMerge(WritePatchBuilder.MergeOutcome o)
...
 
+        // The merged plugin is built as a bare mod, so what lived in a donor's HEADER does not come along. These two
+        // lines are keyed on what the DONORS carried, not on the donor count — the loss is identical for one donor or
+        // ten, and it was measured while each donor overlay was open. Q3: a silent header loss is exactly the kind of
+        // degraded result that has to be stated. Whether these should be CARRIED is a separate decision.
+        if (o.LightDonors is { Count: > 0 } light)
+        {
+            sb.Append("NOTE — ").Append(string.Join(", ", light.Take(10)));
+            if (light.Count > 10) sb.Append(" (+").Append(light.Count - 10).Append(" more)");
+            sb.Append(light.Count == 1 ? " carried" : " carried").Append(" the LIGHT (ESL) header flag; ")
```
Avick3110: `light.Count == 1 ? " carried" : " carried"` — both arms are the same string. Whatever singular/plural distinction was intended was lost in the write.

As shipped it is a conditional that cannot be RED-checked in either direction, which is the one thing CLAUDE.md §5 #11 asks of an added conditional; and the next person to read it will assume a plural form is handled.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -494,10 +494,14 @@ public static string CompactPlugin(
 
     [McpServerTool(Name = "housecarl_merge_plugins", Title = "Merge plugins into one new plugin"),
      Description(
-         "MERGE two or more ACTIVE plugins into ONE NEW plugin — a RECORDS operation (the zMerge/'Merge Plugins' job): the " +
+         "MERGE one or more ACTIVE plugins into ONE NEW plugin — a RECORDS operation (the zMerge/'Merge Plugins' job): the " +
          "donors' records combine under a new filename; the donor FILES and their mods are NEVER touched (new-file lane only, " +
-         "no in-place). RENUMBER is collision-only (zMerge's default): the donor EARLIEST in the load order keeps its FormID " +
-         "object ids; later donors renumber only ids already taken (all records necessarily move to the new plugin's identity). " +
+         "no in-place). TO RENAME a plugin, pass ONE donor: with nothing to combine the merge IS a rename — the same records " +
```
Avick3110, line 499: *(anchored on the rename sentence; the finding is line 520 of this same Description, which the diff view doesn't expose.)*

**The saves reason the report just fixed survives in this description.** Round 1 folded "new FormIDs" out of the report's saves reminder — "rename a pure-override patch and no FormID moves at all" — and the reminder now reads "the records now live under a different plugin name, and any id that had to be renumbered moved with it".

Line 520 still says `Existing SAVES ... will NOT survive (new plugin name + renumbered FormIDs)`, and this is the text the *calling model* reads before it decides anything. A rename with `RecordsRenumbered == 0` now has the tool surface asserting a renumber and the report denying it, in the same file.

(Same block, smaller: the description says nothing about the header losses the report now states.)

Avick3110:
## Review fold — 6 folded, 1 split, 0 refused

All seven findings reproduced at the branch head before anything was changed. **Nothing was refused**, and the ground for that is that every finding cited a line and the line said what it was reported to say — including two where the defect was mine in text I had written this branch (F2's half-derived headline, F6's identical ternary arms). Landed as `51995cb`.

### F1 — the header-loss measurement now matches the loss

Light was read from `IsSmallMaster` alone, so a `.esl` donor with the bit unset lost light status with no note — against the model this tool's own `.esl`-output refusal states, that the engine force-treats the extension as light regardless of the bit. Master status was not measured at all.

Both are now read **by flag or by extension**. Master gets its own NOTE, parallel to the light one, naming **no remedy** — nothing on the surface esmifies a plugin, so a bare statement is the whole of what can honestly be said, the same posture the Author/Description note already had. The light note's `compact_plugin` remedy stays as-is: it applies to the light case only.

Fixture gains a `.esl`-by-extension donor and an `.esm` donor. Both halves of the OR are load-bearing and shown to be — sabotaging to flag-only turns the extension arm RED, sabotaging to extension-only turns the flag arms RED.

**The multi-donor residual is closed in the same work**: a new arm merges a light donor *with* a master donor and asserts all three header notes fire on the two-donor path. The notes are keyed on donor properties, never on the count, and the arm set can now tell those apart.

### F2 — three arms, each claiming only what it read

The branch chose on `RecordsRenumbered` and asserted about `RecordsCopied`. A donor with **no records at all** took the overrides arm and was told its overrides had moved, while the accounting line below printed none — and an empty plugin is exactly what this report's own swap instruction tells the caller to create so a donor `.bsa` keeps loading.

Now three: records moved (states the count), no originating records but N overrides (states *that* count), and no records at all (claims neither). New empty-plugin donor in the fixture; the override arm asserts the count it prints.

### F3 — split

**In-branch:** the coverage caveat, placed on the **pass line** rather than the `none` branch — a populated list is incomplete in exactly the same way, so qualifying one outcome and leaving the other absolute would have been the same bug one level down. The line now reads: *"identify-pass scanned N plugin(s) — it reads record links and record identity, NOT declared masters or INI distributor configs (SPID/KID/SkyPatcher), so a plugin that only lists a donor as a master, or only names one in an .ini, is not counted above."*

**Filed as #367, not folded.** Checked first whether the pass already surfaces declared masters anywhere: it does not — `IdentifyExternalReferencers` walks `RecordsIn` + `EnumerateFormLinks` and opens no header. So the detection is real machinery: a new dependent category through `IdentifyResult`, both outcomes and both renders, plus a policy call for the shared compact lane, which *refuses* on referencers where merge warns.

### F4 — one home for the compact recommendation

Where the ESL note fires, the flat closing tail steps aside, so the version carrying the cost is the one read last. With no note the tail stays — it is then the only pointer. Arms both ways.

**Took the optional sweep** on the tool `Description`'s own flat "Want it light/ESL?" — not because it is the same double-recommendation (the Description is read before the call, the report after), but because that text is where the *calling model* forms its plan, and an un-costed "compact afterward" is what would make it plan a rename-then-compact that silently undoes the id preservation the rename just promised. It now carries the renumber cost.

### F5 — CHANGELOG swap paragraph

Gave as its reason the exact claim the BSA case falsifies. Now scopes "loads by path" to **loose** files and carries the archive caveat the report and Description both already had.

### F6 — deleted

Both ternary arms were `" carried"`. Deleted rather than inventing the plural it implied (#339's preference). No arm needed for a deleted conditional.

### F7 — the Description's saves reason

Round 1 corrected "new FormIDs" out of the report; line 520 still asserted it, so the tool surface was asserting a renumber the report denied, in one file. It now carries the report's reason, plus one sentence naming the header losses.

## Evidence

Guard **56 → 63 arms**; `ci-all` **124/124**; `wire-names-guard` clean over the `[Description]` edits, no friction.

Nine sabotages for this fold, each from a committed state, each restored with the restore verified by anchor — every added conditional collapsed **both** ways:

| Sabotage | Result |
|---|---|
| Light counted by flag only | RED — `.esl`-extension arm |
| Light counted by extension only | RED — 3 arms incl. the multi-donor one |
| Master never detected | RED — master arm + multi-donor arm |
| Master detected for every donor | RED — merge no-notes arm |
| Empty donor falls into the overrides arm | RED — empty arm |
| Override donor falls into the empty arm | RED — pure-override arm |
| Compact tail printed despite the ESL note | RED — tail-suppression arm |
| Compact tail suppressed with no ESL note | RED — merge keeps-pointer arm |
| Pass coverage caveat dropped | RED — identify-pass arm |

One correction on the way through: the fold commit initially claimed F3's sentence while I had not yet written it. Caught before pushing and amended, so the message matches what landed.

No new rounds after this, per §11. Merging remains your explicit go.

Issue #369, opened 2026-08-18
[Bug] OpenOverlay's strings redirect is suppressed by any .bsa beside the plugin, including one holding no strings


Avick3110:
## Summary

`LoadOrderResolver.OpenOverlay` redirects the strings lookup to the game-Data folder only when `FolderHasOwnStrings(path)` is false. That helper answers "has its own" when the plugin's folder contains a loose `Strings\` subfolder **or any `.bsa` at all** — it does not look inside the archive.

So a localized plugin sitting in a mod folder that ships an asset-only `.bsa` (meshes, textures, sounds — the common case), with its `.STRINGS` served from game-Data, takes the folder-adjacent open, resolves nothing, and reads every `TranslatedString` empty. The redirect that exists for exactly this plugin is suppressed by an archive that has nothing to do with strings.

## Why this is the honest reading, not a misuse

The helper's own doc says `any .bsa (which may embed strings)` — the breadth is deliberate, and as a *cheap* gate it is defensible: opening every BSA to check for a strings entry would cost more than the overlay open it precedes. What is not stated anywhere is the consequence, which is that the fix for the HCBR-2026-06-24 class silently does not apply to a whole population of localized plugins.

## Repro shape (measured)

Found while reviewing #362's PR. With that PR's localized fixture — a localized donor whose `.STRINGS` are relocated to game-Data — dropping **any** `.bsa` into the donor's own mod folder flips `FolderHasOwnStrings` to true, and both `merge_plugins` LOCALIZED arms go RED (`Name='' Desc=''`) while the bare-overlay baseline arm stays green. The plugin's strings exist, are reachable, and are not found.

## Impact

Read side: a localized plugin in a BSA-shipping mod folder reads blank names/descriptions everywhere houseCARL reads — the original HCBR-2026-06-24 symptom, unfixed for this population.

Write side: `merge_plugins` and `compact_plugin` bake those blanks into the output, which is #362's data loss reaching the same plugins by a different route. #362's fix does not cover them, and its changelog entry states this bound rather than claiming otherwise.

## Fix shape (not verified — for whoever picks this up)

The obvious options each have a cost worth weighing rather than assuming:
- Look inside the `.bsa` for a `Strings\` entry before counting it. Correct, but pays an archive open per plugin on the index build's hot path.
- Only count a `.bsa` whose basename matches the plugin's (the engine's own auto-load rule). Cheap, and closer to what the engine actually does — but a mod can ship strings in a differently-named archive.
[...]

Issue #368, opened 2026-08-18
[Bug] compact_plugin's repoint lane blanks a localized external referencer, permanently and in place


Avick3110:
## Summary

`housecarl_compact_plugin(plugin=X, in_place=true, repoint_externals=true, acknowledge=true)` rewrites every external referencer through `RemapEngine.RepointInPlace`, which opens the referencer with the **bare** `SkyrimMod.CreateFromBinary(path, SkyrimRelease.SkyrimSE)` (`RemapEngine.cs:874`) and then re-serializes it over the user's original file.

A localized referencer whose `.STRINGS` are served from anywhere but its own folder therefore reads every `TranslatedString` empty, and the repoint writes those blanks back. Same class as #362, same tool — but in the lane that overwrites the original, with no houseCARL backup and no undo.

## Why this is confidently a bug, not a design choice

The comment at `RemapEngine.cs:889-893` argues a bare open is correct here, and for what it is describing it is right — but it is describing the **master overlays** resolved a couple of dozen lines further down, which exist only to resolve FormID/master-table references on re-serialize and are never read for content. The **target** open at `:874` is a different thing: its records are mutated by `RemapLinks` and written back. Nothing in that comment covers it.

## Repro shape (measured)

A localized `ZRef.esp` in its own mod folder, `.STRINGS` relocated to game-Data, holding a FormList link into the compact target `ZTgt.esp`:

```
BEFORE repoint — ZRef via OpenOverlay+dataDir: Name='REF NAME' Desc='REF DESC'
compact success=True inPlace=True repointed=ZRef.esp:True
AFTER  repoint — ZRef via OpenOverlay+dataDir: Name='' Desc=''
                 (UsingLocalization=True, own Strings dir=False)
*** LOSS CONFIRMED ***
```

The tool reports `OK ZRef.esp`.

## Impact

Worse than #362 in two ways. The write lands on the user's own file rather than a new one, so there is nothing to discard. And because `CreateFromBinary` preserves the existing header, the rewritten plugin stays **flagged localized** with no strings source beside it — so no later read recovers the values, and the plugin is permanently blank rather than merely written blank once.

Q3: silent data loss, reported as success.

## Fix shape (not verified — for whoever picks this up)

[...]

PR #370, opened 2026-08-18
fix(merge,compact): open donors through OpenOverlay so localized strings survive the write

### LINKED ISSUE #362: [Bug] merge_plugins / compact_plugin blank every localized string — donors open with the bare overlay, bypassing OpenOverlay

Avick3110:
`Fixes #362`. Carries #364's minimum reminder sentence as its own commit (referenced, not closed — the detect-and-name half stays open at W5).

## The defect

`merge_plugins` and `compact_plugin` opened their donor/source plugins with the bare `SkyrimMod.CreateFromBinaryOverlay(path, release)`, which finds `Strings\` only in the plugin's own folder. A localized plugin whose `.STRINGS` are served from elsewhere read every `TranslatedString` empty, and both verbs wrote those blanks into the output. Silent data loss on two shipped write verbs (Q3). `LoadOrderResolver.OpenOverlay` exists for this class and the in-place lane already routed through it; these two opens did not.

## The write-lane `dataDir` verification

Answered at source before writing the fix, as the routing ruling required. All three resolved as expected, so the engagement stayed in scope.

**(a) Are the tool-layer call sites simple enough for parameter threading?** Yes. Both callers (`LoadOrderService.cs:6080`, `:6317`) already hold the captured `IndexView` they pass `PluginPath` from. The resolver's own `DataDir` is `internal` and `housecarl-core` is `InternalsVisibleTo` only `housecarl-generator`, so the MCP layer cannot read it directly — hence one expression-bodied `IndexView.DataDir` property beside the existing `PluginPath`, plus a `string? dataDir` parameter on each builder. No redesign.

**(b) Are `OpenOverlay`'s fallback semantics right for a donor in an MO2 mod folder?** Yes — the same resolution the read index and the in-place lane already use. It redirects only when the plugin's own folder carries no strings source, so a donor with its own `Strings\` or `.bsa` is never redirected away from it. That gate has a bound, filed as #369 (below).

**(c) What is the OUTPUT's localization state?** Written non-localized, strings inline — the expected answer, so no STOP. Both builders construct a fresh bare `SkyrimMod`; `RemapEngine` never touches `ModHeader.Flags`; `UsingLocalization` appears nowhere in production code; and the write passes no strings writer. Verified at source **and** pinned by arms, because it is what keeps the existing bare read-back opens correct.

## Arms

A shared localized fixture whose load-bearing part is a game-Data `Skyrim.esm` — without one `ComputeDataDir` returns null, every `OpenOverlay` degrades to the folder-adjacent default, and the arms would pass with or without the fix.

- Merge: single-donor (the rename shape) and multi-donor both carry `Name`/`DESC` for every donor. One fix, no donor-count conditional.
- Compact: source carries `Name`/`DESC` into P'.
- Baseline arms read the donor/source with the **bare** overlay first and require EMPTY, so a fixture that stopped being localized fails loudly instead of passing vacuously.
- Output-not-localized arms pin (c).
- A `RESIDUAL` arm pins the unfixed case as measured behaviour (below).

RED-checked in both directions from a committed state: reverting either call site blanks rename/combine/compact; nulling `view.DataDir` at the service call sites reddens the same three; sabotaging the fixture's strings relocation reddens both baselines. Restores anchor-verified. `ci-all` 124/124.

## Review rounds

Two rounds, then a **§11 class-recurrence stop** — escalated rather than folded a third time. The stop and its directed completion are the reason this PR looks the way it does.

**Round 1** (2 agents) — 2 mediums, 2 lows, every one in the #364 sentence, none in the fix. The sentence claimed all four config systems address a record as `Plugin.esp|FormID` (true of SkyPatcher alone — SPID/KID are suffix-tilde, OAR is `config.json`; `spid-authoring/SKILL.md:107` lists confusing the first two as a known trap); claimed overrides took the output's identity (they keep their master's FormKey) three lines under an accounting line saying so; claimed compact's break present-tense before the swap; and the changelog claimed a false exhaustiveness. Also two guard defects of mine — a `RESIDUAL` arm that a *missing* record would have satisfied, and a fixture mutating a directory while lazily enumerating it. **Refused nothing**: every finding reproduced at source or under measurement, and the override claim was probed (not reworded) first — the rendered report contradicted itself on the pure-override donor.

**Round 2** (2 agents) — the same class again, one layer down: the *corrected* model was still false, because all four systems also accept EditorIDs and both verbs preserve them, so those lines never break. Plus the changelog claiming a closure the fix lacks, and — worst — **the two arms added in the round-1 fold did not work**. They asserted `!Contains("<plugin>|<FormID>")`, the literal from the commit they were fixing, so a reviewer quoting a real SkyPatcher spelling left them green.

That is the same failure class in two consecutive rounds, so the rounds stopped there.

**Directed completion (advisor lane, no round 3).** The class dies by a **claim rule** pinned in-code at both sentences: every clause states either something this tool did, or a break entailed by what this operation did, scoped by the addressed line's own content — nothing describes an external grammar. Both sentences were rewritten under it and moved into `WriteSentences`, which brings them under the existing `[MustState]` walk: content pinned by declared phrases, wiring pinned by the reach arm. The dead `!Contains` arms are **deleted, not rebuilt** — an absence assertion over an open string space cannot fail meaningfully.

Before committing, verified per the directive that each of the four systems has a line form naming a plugin, and that **no bundled skill documents a fallback keeping such a line matched after that plugin deactivates** — SkyPatcher's skill states the opposite outright, that a form it cannot resolve is silently skipped. All four stay in the claim.

## Residuals and spawned issues

- **#368** (filed) — `compact_plugin(in_place, repoint_externals, acknowledge)` blanks a localized external referencer **permanently**: `RemapEngine.cs:874` opens it bare and re-serializes over the user's own file, keeping the localized header flag with no strings source. Found by a reviewer, reproduced independently before filing. Outside the ruled scope, and its fix is not this one-liner (`RepointInPlace` needs a mutable mod; `OpenOverlay` returns a getter).
- **#369** (filed) — `FolderHasOwnStrings` counts **any** `.bsa` beside the plugin as a strings source, so an asset-only archive suppresses the redirect and the plugin still blanks. Pre-existing `OpenOverlay` contract. The changelog states this as a bound rather than claiming a closure.
- **#363** (commented) — merge/compact drop the localized-strings layer entirely (output is non-localized, strings inlined). Correct, and what makes this fix usable, but it is another header-layer property that does not survive; left on the issue that owns the carry decision.
- **Open question, unfixed by design:** a localized donor whose strings are resolvable **nowhere** still merges to blanks, silently — `OpenOverlay`'s fallback is game-Data and nothing else. Pinned by the `RESIDUAL` arm as current behaviour rather than asserted. No warning feature grown here without a ruling.
- **Second strings residual, out of scope:** nothing in production sets `TargetLanguage`, so Mutagen's English default governs every houseCARL read — a non-English `.STRINGS` set reads blank regardless of `dataDir`. Pre-existing across the whole read surface (settled decision 1).

## Changelog

Two `## Unreleased` entries — the fix (with its bound stated), and the reminder sentences.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -6,21 +6,18 @@ when it changes.
 
 ## Unreleased
 
-- **`housecarl_merge_plugins` and `housecarl_compact_plugin` now say what they cost the runtime distributor layer.**
-  SPID, KID and SkyPatcher `.ini` lines, and Open Animation Replacer `.json` conditions, all address a record through
-  the plugin that defines it — an addressing model neither verb leaves intact, and one no plugin scan can see, because
-  those lines live in config files rather than in any record. The merge report already said the identify pass does not
-  read them; it now also says what follows from that, which is that a line naming a donor stops matching once you
-  deactivate that donor at the swap. The compact report said neither, and now says both — with the half that applies
-  to it, since a compaction keeps the plugin name and moves the object ids instead.
+- **`housecarl_merge_plugins` and `housecarl_compact_plugin` now say what they cost your runtime config files.**
+  Neither verb reads SPID, KID, SkyPatcher or Open Animation Replacer configs, and neither rewrites them — but both
+  can invalidate lines in them, and nothing said so. Merge now states that a line naming a donor stops matching once
+  you deactivate that donor at the swap. Compact states the half that applies to it: the plugin name survives a
+  compaction, so what breaks is a line addressing a record by an object id the compaction moved, once the compacted
+  plugin is the one loading.
 
-  Both reports state the addressing *model* and quote no single spelling, because the four systems do not share one:
-  SkyPatcher is prefix-pipe, SPID and KID are suffix-tilde, and OAR is a JSON field pair. Naming the file kinds is what
-  makes the advice actionable; quoting one syntax for all of them would send you grepping for a string three of the
-  four never contain. Merge's sentence is also scoped to the records that actually changed identity — a record a donor
-  merely *overrode* keeps its master's FormID, so a config line addressing it is untouched, and the report says so
-  rather than telling you to rewrite a line that still works. Neither verb rewrites those files, and neither claims to
-  have looked at them. (#364)
+  Both sentences deliberately describe only what the operation did, and say nothing about how any of those config
+  formats are written. Earlier drafts tried to be helpful about the syntax and got it wrong twice — the four systems
+  share neither a file format nor a way of naming a record, and each attempt would have sent you looking for something
+  three of them never contain. What the reports owe you is what *they* changed; the bundled skills are where the
+  grammars live. (#364)
 
 - **Fixed: `housecarl_merge_plugins` and `housecarl_compact_plugin` blanked every localized name and description.**
   Both verbs opened their donor / source plugins with an overlay that looks for `Strings\` only in the plugin's own
@@ -29,10 +26,14 @@ when it changes.
   to — therefore read every `FULL` and `DESC` as empty, and the merged or compacted output was written with those
   blanks baked in. Nothing failed and nothing warned: the result loaded, and the loss was visible only in game or by
   diffing. Both opens now go through the same strings-aware path the rest of houseCARL reads with, which falls back to
-  the game `Data` folder when the plugin's own folder carries no strings source and leaves plugins that do carry one
-  untouched. This fix closes the case where the strings exist but were being looked for in the wrong folder. It does
-  not change what happens when the lookup finds nothing wherever it looks — a plugin whose strings are in *neither*
-  place, for instance — and such a plugin still reads, and writes, empty. (#362)
+  the game `Data` folder when the plugin's own folder carries no strings source, and leaves plugins that do carry one
+  untouched.
+
+  That redirect has a bound worth knowing, because it decides whether this fix reaches your case: it applies only when
+  the plugin's own folder carries no strings source at all, and **any** `.bsa` sitting beside the plugin counts as one
+  — whatever that archive actually holds. A localized plugin in a mod folder that ships an asset-only `.bsa`, with its
+  `.STRINGS` served from elsewhere, is therefore still read the old way and still writes blanks. So is a plugin whose
+  strings are in neither place. Both are unchanged by this fix. (#362)
 
 - **`housecarl_merge_plugins` now accepts a single donor, which renames a plugin.** Merging one plugin into a new
   name was refused ("merge needs at least TWO distinct donor plugins"), and no other tool renames one, so a
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -6,6 +6,17 @@ when it changes.
 
 ## Unreleased
 
+- **Fixed: `housecarl_merge_plugins` and `housecarl_compact_plugin` blanked every localized name and description.**
+  Both verbs opened their donor / source plugins with an overlay that looks for `Strings\` only in the plugin's own
+  folder. A localized plugin whose `.STRINGS` are served from elsewhere — the common case, where a cleaned or
+  translated master's strings live in the game's `Data` folder rather than in the mod folder MO2 resolves the plugin
+  to — therefore read every `FULL` and `DESC` as empty, and the merged or compacted output was written with those
+  blanks baked in. Nothing failed and nothing warned: the result loaded, and the loss was visible only in game or by
+  diffing. Both opens now go through the same strings-aware path the rest of houseCARL reads with, which falls back to
+  the game `Data` folder when the plugin's own folder carries no strings source and leaves plugins that do carry one
+  untouched. A plugin whose strings are in *neither* place still reads empty — there is nowhere left to look — and
+  that case is unchanged by this fix. (#362)
+
 - **`housecarl_merge_plugins` now accepts a single donor, which renames a plugin.** Merging one plugin into a new
   name was refused ("merge needs at least TWO distinct donor plugins"), and no other tool renames one, so a
   mis-named patch could only be fixed by replaying every edit into a freshly named plugin. There was never anything
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -6,6 +6,14 @@ when it changes.
 
 ## Unreleased
 
+- **`housecarl_merge_plugins` and `housecarl_compact_plugin` now say what they cost the runtime distributor layer.**
+  SPID, KID, SkyPatcher and OAR configs address a record as `<plugin>|<FormID>` — an addressing model neither verb
+  leaves intact, and one no plugin scan can see, since those lines live in `.ini` files rather than in any record. The
+  merge report already said the identify pass does not read them; it now also says what follows from that, which is
+  that a line naming a donor stops matching the moment you deactivate that donor at the swap. The compact report said
+  neither, and now says both — with the half that applies to it, since a compaction keeps the plugin name and moves
+  the object ids instead. Neither verb rewrites those files, and neither claims to have looked at them. (#364)
+
 - **Fixed: `housecarl_merge_plugins` and `housecarl_compact_plugin` blanked every localized name and description.**
   Both verbs opened their donor / source plugins with an overlay that looks for `Strings\` only in the plugin's own
   folder. A localized plugin whose `.STRINGS` are served from elsewhere — the common case, where a cleaned or
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -7,12 +7,20 @@ when it changes.
 ## Unreleased
 
 - **`housecarl_merge_plugins` and `housecarl_compact_plugin` now say what they cost the runtime distributor layer.**
-  SPID, KID, SkyPatcher and OAR configs address a record as `<plugin>|<FormID>` — an addressing model neither verb
-  leaves intact, and one no plugin scan can see, since those lines live in `.ini` files rather than in any record. The
-  merge report already said the identify pass does not read them; it now also says what follows from that, which is
-  that a line naming a donor stops matching the moment you deactivate that donor at the swap. The compact report said
-  neither, and now says both — with the half that applies to it, since a compaction keeps the plugin name and moves
-  the object ids instead. Neither verb rewrites those files, and neither claims to have looked at them. (#364)
+  SPID, KID and SkyPatcher `.ini` lines, and Open Animation Replacer `.json` conditions, all address a record through
+  the plugin that defines it — an addressing model neither verb leaves intact, and one no plugin scan can see, because
+  those lines live in config files rather than in any record. The merge report already said the identify pass does not
+  read them; it now also says what follows from that, which is that a line naming a donor stops matching once you
+  deactivate that donor at the swap. The compact report said neither, and now says both — with the half that applies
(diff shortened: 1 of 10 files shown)


src/housecarl-mcp/WriteSentences.cs
```diff
@@ -131,6 +131,50 @@ internal static class WriteSentences
...
+    // and that no bundled skill documents a fallback keeping such a line matched once that plugin deactivates —
+    // SkyPatcher's skill states the opposite outright, that a form it cannot resolve is silently skipped.
+
+    /// <summary>Merge's reminder. The break is entailed by the swap this same report instructs: deactivating a donor
+    /// is what stops a line that NAMES that donor from matching. Deliberately no override/master carve-out — a line
+    /// addressing a record the donor merely overrode names the MASTER, so "names a donor" already excludes it by
+    /// construction, and spelling that out would be a claim about addressing rather than about this operation.</summary>
+    [MustState("Nothing here rewrites those files", "stops matching once you deactivate that donor at the swap")]
+    internal const string MergeRuntimeConfigs =
+        "Nothing here rewrites those files. A line in one of them that names a donor stops matching once you " +
```
Avick3110: **[medium] "stops matching" is the wrong failure direction for an exclusion filter.**

The clause claims a line that *names a donor* stops matching once that donor is deactivated. For a filter that names the donor as an **exclusion**, the line does not stop matching — its scope *widens*.

Concrete: a `_DISTR.ini` carries

```ini
Spell = 0x800~MySpells.esp|ActorTypeNPC|-DonorMod.esp
```

SPID's plain-plugin-name Form filter with the `-` exclude modifier (`spid-authoring/references/filters.md:97-107` — "a plain plugin name matches all NPCs *defined in* that plugin", `-` = "Exclude (always AND)"). Merge `DonorMod.esp` into `Merged.esp`, follow this same report's swap instruction, deactivate the donor: the exclusion resolves to nothing and the donor's NPCs — now defined by `Merged.esp` — start receiving a spell they were explicitly excluded from.

SkyPatcher is the same shape: `filterByModNamesExcluded=DonorMod.esp` ("Restrict to records that … aren't from the named plugin(s)", `skypatcher-authoring/references/grammar-core.md:204`) now patches records it previously skipped.

So the caller is told to look for a dead line while the actual symptom is over-application — the harder of the two to notice, and the one that changes gameplay silently.

Worth noting against the claim rule three lines up: "stops matching" is a statement about *how those four systems evaluate a line whose named plugin is gone*, not about what merge did. That is an external-grammar claim, and none of the bundled skills documents unresolvable-filter semantics, so it isn't verifiable from the sources the rule points at. The pre-commit check ("no bundled skill documents a fallback keeping such a line matched") is absence-of-evidence and does not reach the exclusion shape at all.

A clause that stays inside the rule would say only what the swap did — e.g. that the donor's records now live under a new plugin name, so any line filtering on the donor's name no longer describes them — and leave the per-system consequence to the skills.

plugin/CHANGELOG.md
```diff
@@ -6,6 +6,35 @@ when it changes.
...
+  compaction, so what breaks is a line addressing a record by an object id the compaction moved, once the compacted
+  plugin is the one loading.
+
+  Both sentences deliberately describe only what the operation did, and say nothing about how any of those config
+  formats are written. Earlier drafts tried to be helpful about the syntax and got it wrong twice — the four systems
+  share neither a file format nor a way of naming a record, and each attempt would have sent you looking for something
+  three of them never contain. What the reports owe you is what *they* changed; the bundled skills are where the
+  grammars live. (#364)
+
+- **Fixed: `housecarl_merge_plugins` and `housecarl_compact_plugin` blanked every localized name and description.**
```
Avick3110, line 22: **[medium] The entry names `housecarl_compact_plugin` unqualified, and its bounds paragraph omits the lane that still blanks — permanently.**

The bounds paragraph below (lines 32–36) lists two cases the fix does not reach: an asset-only `.bsa` beside the plugin, and strings resolvable nowhere. It does not mention the third, which is inside this very verb:

```
housecarl_compact_plugin("Target.esp", in_place=true, repoint_externals=true, acknowledge=true)
```

`RemapEngine.RepointInPlace` opens each external referencer with the bare `SkyrimMod.CreateFromBinary(path, SkyrimRelease.SkyrimSE)` (`src/housecarl-core/RemapEngine.cs:874`) and re-serializes over the user's own file. If `Ref.esp` is localized with its `.STRINGS` in game `Data`, every `FULL`/`DESC` in it is lost — permanently, in place, on a file the user did not ask to compact, and with the same "nothing failed and nothing warned" signature this entry describes as the bug.

The PR body discloses this as #368 and rules it out of scope, which is fine as a scoping call. But the changelog is the artifact users read, and as written it says compact "blanked every localized name and description" in the past tense with two named exceptions — a user with a localized referencer will conclude they are covered.

One clause in the bounds paragraph naming the repoint lane (and #368) closes it without widening the fix.

src/housecarl-mcp/WriteSentences.cs
```diff
@@ -131,6 +131,50 @@ internal static class WriteSentences
...
+        "deactivate that donor at the swap.\n";
```
Avick3110: **[low] Scoped to "a line", so it misses the break that takes a whole file.**

SkyPatcher reads a config file whose *name* is a plugin filename only while that plugin is active: `Plugin.esm.ini` / `Plugin.esp.ini` → "**Only** when that plugin is active in the load order; otherwise skipped" (`skypatcher-authoring/references/grammar-core.md:93`, restated at 218–219 as "the filename gate decides whether the *file* is read at all").

So a SkyPatcher patch shipped as `DonorMod.esp.ini` goes entirely dark at the swap — every line in it, including the ones that never name the donor and patch records from unrelated plugins. The sentence tells the caller only that *a line naming a donor* stops matching, so nothing in the report points at the file, and the break leaves no trace.

This one is squarely inside the claim rule (the file stopped loading because this operation's swap deactivated the plugin it is named for), so it can be stated without reaching for anyone's grammar — e.g. that deactivating the donor also stops any config file *named* for it from loading.

Compact is correctly unaffected here: the plugin name survives a compaction, so the filename gate still holds. This is merge-only.

Avick3110:
## Fold summary — all three findings folded, none refused

Each verified against the cited lines before folding. `ci-all` green; no new review rounds (§11).

### F1 (medium) — the merge clause's failure direction · `92d8da4`

Confirmed at both citations. SPID's plain-plugin-name Form filter under `-` (`filters.md:97-107`) and SkyPatcher's `filterByModNamesExcluded` (`grammar-core.md:204`) both **widen** when the donor's records move to the output — the exclusion resolves to nothing and those records start receiving what they were excluded from. The report was pointing the caller at a dead line while the real symptom was over-application.

The structural half of your finding is the one that changed how the clause is written. "Stops matching" is a statement about how those systems *evaluate* a line whose named plugin is gone — an external-grammar claim sitting three lines under a rule forbidding exactly that. And the pre-commit check I ran ("no skill documents a fallback") was absence of evidence that never reached the exclusion shape at all.

The clause now states only the operation's fact and stops:

> "After the swap a donor's records live under the merged plugin's name, so a line that names a donor — whether to address a record in it or to filter on it — no longer describes those records."

Direction-neutral: it names neither a dead line nor an over-applying one, because which one a caller gets depends on the system and the line. Both line shapes are covered deliberately — addressing a record *in* the donor and filtering *on* it are one fact from this side.

### F3 (low) — the whole-file break · same commit

Confirmed at `grammar-core.md:93` and the restatement at 218-219. Folded **SkyPatcher-named**, not generalized:

> "Separately, a SkyPatcher config file whose NAME is a donor's filename (Plugin.esp.ini) is read only while that plugin is active, so it stops loading at the swap: every line in it, including the lines that never name a donor."

Stating it required the claim rule to gain a third arm, now pinned with the other two: **(iii) a per-system fact, stated only where a bundled skill states it verbatim, cited by file and line beside the constant — quoted, never derived.**

The generalized form ("any config file named for the donor") was **considered and refused** under that arm, and the refusal is recorded in the commit: SPID and KID document filename rules with no activity gate, so the general claim is underivable from our sources and would be wrong-shaped for a SPID file that merely happens to carry a donor's name.

### The directed sweep — compact survives, unchanged

Probed rather than assumed. Compact's clause says an address **no longer reaches** a record: an identity fact about the address and the record's id, entailed by the renumber, true of include and exclude lines alike, and claiming nothing about evaluation. An exclusion line by object id stops excluding; an include line stops applying; the clause states neither consequence, so the exclusion shape does not falsify it.

F3 cannot fire there either — the plugin name survives a compaction in both lanes (`outPath = srcPath` in place; exact basename new-file), which the clause already states as this tool's own fact, so the filename gate still holds. Your scoping call was right.

### F2 (medium) — changelog bounds · `3c5cafb`

Confirmed at `RemapEngine.cs:874`. Added as its own commit, one clause in the bounds paragraph naming the repoint lane and citing #368, with the point that it is the worst of the three: permanent, in place, on a plugin the user did not ask to compact.

### Verification

RED sweep re-run from the committed state on both changed `[MustState]` phrases — each fails the content walk by name when the sentence clause is gutted and the declaration left intact. (My first attempt was an invalid sabotage: the `sed` moved the phrase *and* the sentence together, so it passed trivially. Re-run against the sentence only.) Restores anchor-verified, tree clean, `ci-all` ALL PASS.

Issue #374, opened 2026-08-19
[Bug] compact_plugin's external-referencer refusal offers a repoint that a localized referencer will refuse


Avick3110:
## Summary

`housecarl_compact_plugin`'s external-referencer refusal (`src/housecarl-mcp/LoadOrderService.cs:6054`) ends:

> Re-run with `repoint_externals=true` AND `in_place=true` (+ `acknowledge=true`) to ALSO rewrite those plugins in place to follow the renumber, or handle them yourself first.

When one of those referencers is flagged **localized**, following that instruction cannot work: the repoint lane refuses, because houseCARL cannot re-serialize a localized plugin without scrambling its text. The caller is handed a remedy, does exactly what it says, and meets a refusal.

The second refusal is honest and explains the real blocker, so nothing breaks and nothing is silently wrong — this is a wasted round trip and a remedy that names an action the tool will not perform, not a data-loss path.

## Why it is filed rather than fixed in place

Found during the pre-PR review rounds of the branch that added the localized refusal (PR for #368's interim guard). It is pre-existing text in a code path that branch does not otherwise change, and fixing it properly is a small design call rather than a wording tweak: the first refusal would have to run the localized check itself to warn earlier, which means deciding whether that refusal grows a second, conditional clause — the kind of choice the refusal-ergonomics issues (#303, #319, #359) are already collecting.

## Shape of a fix

Run the same `RemapEngine.LocalizedAmong` check in the `repoint_externals=false` branch and name the localized referencer(s) in that first refusal, so the caller learns the repoint route is unavailable before being told to take it. Wants an arm in `compact-service-guard` in both directions, since it adds a conditional to a shipped sentence.

## Not affected

The other half of that same sentence — "or handle them yourself first" — remains executable: removing or redirecting the references is a real action that does not depend on FormIDs that do not exist yet.

Avick3110:
[...]

Issue #373, opened 2026-08-19
[Bug] WriteInPlace commits the plugin without the string tables its own serialize emitted — every localized value shifts to another record


Avick3110:
## Summary

`WriteEngine.WriteInPlace` stages the serialize into a `.housecarl-tmp` sibling and `CommitStagedPatch` commits **the plugin file alone**. When the target has `UsingLocalization` set, that serialize also emits `.STRINGS` / `.DLSTRINGS` / `.ILSTRINGS` into the staging directory — with a **renumbered** string table. Those emitted tables are discarded with the staging directory, so the committed plugin's new indices resolve against the plugin's **old, untouched** strings file, and every localized value shifts onto a different record.

The result is a silently wrong answer, not a blank one: a weapon reads a book's name.

- Staging: `src/housecarl-core/WriteEngine.cs:1828` (`WriteInPlaceStaged` — `BeginWrite.ToPath(tmpPath)`, which is what emits the string tables beside the staged plugin).
- Commit: `src/housecarl-core/WriteEngine.cs:1773` (`CommitStagedPatch` — `AtomicFile.Commit(tmpPath, outputPath)`, one file, then `CleanupStaged` discards the rest).

## Measured

Lane measured: **`RemapEngine.RepointInPlace`** (`src/housecarl-core/RemapEngine.cs:928`), driven by a synthetic MO2 instance — a real `Skyrim.esm` in game-Data, a localized `ZRef.esp` in its own mod folder holding a FormList link into a compact target, five weapons and five books so an index that moves has somewhere to show itself.

The load-bearing arm is the one where **the plugin's strings sit correctly beside it** — no relocation, nothing wrong with the plugin, no strings-resolution problem of any kind. It reads correctly in memory through the ordinary bare open, and the re-serialize scrambles it:

```
== ARM C: strings LEFT BESIDE the plugin — the ordinary localized plugin, bare open ==
  BEFORE (OpenOverlay + dataDir): ZRefWeap0='REF NAME 0'/'REF DESC 0' | ZRefWeap1='REF NAME 1'/'REF DESC 1' | ZRefWeap2='REF NAME 2'/'REF DESC 2' | ZRefWeap3='REF NAME 3'/'REF DESC 3' | ZRefWeap4='REF NAME 4'/'REF DESC 4' link=000801:ZTgt.esp flags=Localized
[...]

Issue #371, opened 2026-08-18
[GAP] merge/compact write blanks silently when a localized donor's strings resolve nowhere


Avick3110:
## Summary

`housecarl_merge_plugins` and `housecarl_compact_plugin` now open donors through `LoadOrderResolver.OpenOverlay` (PR #370), which falls back to the game `Data` folder when the plugin's own folder carries no strings source — and to nothing else. A localized donor whose `.STRINGS` are resolvable **nowhere** (not beside the plugin, not in game `Data`) still reads every `TranslatedString` empty, and the output is written with those blanks baked in. No warning, no refusal — the same silence #362 fixed for the resolvable case.

## Current state

Deliberate and disclosed, not accidental: PR #370 pins this as measured behaviour with a `RESIDUAL` arm and states it in the PR body ('no warning feature grown here without a ruling'). This issue is that ruling's tracker — the residual is real data loss for a caller who doesn't know their strings mod is missing.

## Fix direction

A warning (or refusal — shape decided at the charter) when a localized donor resolves zero strings sources. Detection is the same machinery as #369's fix: `FolderHasOwnStrings` today answers 'is there a source' with an existence check; once #369 makes it answer truthfully (looking inside the `.bsa`), 'no source anywhere' is a known fact at open time and the warn falls out nearly free. The two belong to one engagement.

## Provenance

Surfaced during PR #370's review rounds and directed completion; pinned there as the `RESIDUAL` arm. Filed by the advisor lane on Aaron's go, 2026-08-18.

Routed at filing (Aaron-go, routing delegated): **RUN_ORDER row 8/W5, beside #369** — same engagement, shared machinery. Milestoned `houseCARL 2.0`.

PR #375, opened 2026-08-19
Refuse a localized plugin in place rather than re-serialize it into wrong text

Avick3110:
Interim refusal for the localized in-place write. **No closing keywords** — #368 and #373 both stay open for the attended engagement that fixes the round-trip properly.

## Why this is a refusal and not the fix

The kickoff made an empirical check binding before writing anything: does the mutable open honour the same strings overrides `OpenOverlay` wires for reads, and does the mutate → `WriteInPlace` round-trip preserve them on disk? It does not, and the routed fix shape does not hold.

Three arms, synthetic MO2 instance with a real `Skyrim.esm` in game-Data, two record types so a moved string index has somewhere to show itself:

| Arm | Setup | Result |
|---|---|---|
| **A** | today's bare `CreateFromBinary`, strings relocated to game-Data | blank, and still blank when re-read *with* the game-Data folder — the reported loss, end to end |
| **B** | `CreateFromBinary` **+ the same `StringsParam` overrides** | reads perfectly **in memory**; on disk a mix of correct, blank, and *another record's text* |
| **C** | strings left **beside the plugin** — a healthy localized plugin | same scramble: a weapon reads a book's name |

The overload exists and honours the overrides — arm B's in-memory read is exact. The write is what breaks: `WriteInPlace` stages the serialize, Mutagen emits a **renumbered** string table into the staging directory, and `CommitStagedPatch` commits the plugin alone. The committed plugin's new indices then resolve against its old, untouched `.STRINGS`.

Arm C is the load-bearing one: it needs no strings-resolution problem at all, so this is not #368's read side. Filed as **#373** (write side), with the six other `WriteInPlace` call sites listed as suspected-not-measured — two of them are new-file-lane re-emits, so the suspected class is not confined to `in_place=`.

Fixing it faithfully decides what the in-place lane may write besides the plugin file, which is the in-place consent lane's design — stop condition 1. Escalated; the ruling demoted #368 into an attended engagement and scoped this PR to a refusal.

## What lands

**Choke point.** `WriteEngine.WriteInPlace` refuses a `UsingLocalization` target before it creates its staging directory — one site, covering all seven callers. The refusal is its own whole sentence and every lane renders it verbatim: the lanes that go through `SerializeFailure` get the treatment the baseline-master refusal already had there, and the remove and repoint lanes carry their own arm, because their leads say "failed (serialize or commit)" about steps this refusal happens before.

**Service pre-flight, ahead of the consent gate**, in all four in-place lanes. The choke point alone was too late for two promises: a dry run never reaches the write, so `in_place` + `dry_run` on a localized plugin reported the edit landing while the real call refused; and the service persists the one-time in-place acknowledgement *before* attempting the write, so the first acknowledged call spent it on a plugin houseCARL would never write. The repo had already settled this shape — `InPlaceParentUnwritable` is deliberately kept in the dry-run path because predicting the real refusal is the dry run's job.

**Compact pre-flight.** Its referencer rewrites run only *after* the compacted plugin is on disk, so a refusal discovered there strands a half-completed compaction — target renumbered, referencers still on the old FormIDs. `RemapEngine.LocalizedAmong` checks them before anything is written, above the consent gate so the modder is not asked to authorise an irreversible rewrite that cannot happen.

**Compact's own target** (directed fold, post-rounds). The in-place compaction was the one lane the refusal did not reach, and it reaches it for a different reason than the others: a compaction does not re-serialize the target, it builds a **fresh** plugin and writes that over the original, so the mod handed to the write is never flagged localized and the write's check structurally cannot see this case. What makes it refusable is what the rebuild *does* — when the strings resolve the result is **de-localized** (one language baked in, the mod's `.STRINGS` set no longer describing it, nobody having chosen that); when they do not, blanks. Both land silently on a file with no review step and no undo. Keyed on the header flag, deliberately wider than the strings-resolve-nowhere case, which is separate machinery this PR does not absorb. Placed before the identify pass, the consent gate and any write — early on purpose, so a target that also has external referencers is not sent round the referencer refusal's remedy first.

The **new-file compact lane is untouched**, and is what the refusal points at: its output is a plugin the modder reviews before enabling. The remedy is measured, not assumed — the arm directly above the new one compacts the same localized fixture to a new file and reads its `FULL`+`DESC` back intact — and the sentence names the lane without promising the output's nature, since that output is not localized either.

The other six `WriteInPlace` callers need no pre-flight: their `WriteInPlace` call is the first thing each puts on disk, verified per call site.

## Review rounds — 2, then class-stop

**Round 1 — 3 fresh agents, isolated worktrees. 17 findings: 11 folded, 4 refused, 2 surfaced.**

Folded, the two that mattered being defects this branch introduced: the dry-run/real divergence and the spent consent (above); and the compact refusal named a remedy measured **false** — "repoint those plugins yourself, then compact" cannot be done, because the new FormIDs do not exist until the compaction runs and the verb never discloses the mapping, and a referencer repointed to guessed ids stops matching the identify pass, so the follow-up compaction *succeeds* and reports a clean run over links pointing nowhere. It states the dead end instead. Also: "every localized value would land on the wrong record" was refuted by this branch's own probe (arm C is 2 correct, 1 blank, 2 wrong — the mix being the more dangerous finding); the changelog missed lanes; the evidence probe died on its own guard, so the citation pointed at something that would not run.

Refused: **compact in-place on a localized *target* can still blank it** — that is #371's no-source class, routed to W5 by ruling and named in this engagement's stop condition 2, and a different mechanism (read-side blank, not index drift); the **`into=` extend lane** has no localized check — real, but a different choke point (`WritePatch`), reachable only by localizing a houseCARL-owned patch and extending it, so the changelog claim was scoped honestly instead; **no dedicated arms for the forward/create render paths** — both render through `SerializeFailure`, which LOC-A pins, and the two *distinct* render paths each have one; **`LocalizedAmong` fails open** — deliberate, though the reviewer was right that "absolute backstop" overclaimed, so the doc now says failing open costs the earliness, not the safety.

**Round 2 — 2 fresh agents, isolated worktrees. 15 findings: 12 folded, 2 refused, 1 filed.**

Folded: the four new refusals were **unstamped** (every other post-capture outcome in those lanes carries the index epoch, including the refusal three lines above each); the changelog contradicted itself inside one release block, its `#362` bullet still describing the referencer blanking as live while the new entry says it is refused; the entry implied compact in-place was covered when it structurally is not; lane list corrected to the 2.0 spellings and the `in_place="X.esp"` form; the mechanism clause moved to a single constant, since the compact refusal was a second hand-written spelling of it; per-lane comments no longer claim a dry-run branch in the two lanes that have none.

Refused: the earlier compact refusal's "**or handle them yourself first**" is not false — it names removing or redirecting the references, which needs no FormIDs that do not exist yet; the contradiction is with the *other* half of that sentence, pre-existing and filed as **#374** rather than folded here. And the consent guarantee stays unhedged: falsifying it needs a header read to throw on a plugin the resolver just enumerated, and the file is untouched either way.

**Class-stop, no round 3.** Two classes recurred across both rounds: *prose claiming coverage the code does not have* (round 1's magnitude overclaim, round 2's lane enumeration), and *arms not isolating what they name* (round 1's `stringsSame`; round 2's per-lane arms). Both are folded, and the design signal behind the first is stated below.

## Sabotage RED-checks — 11, each from a committed state, each restored and verified by anchor grep

| Sabotage | Result |
|---|---|
| choke point disabled | LOC-A/C + backstop RED — referencer came back `untouched=False`, actually rewritten |
| compact pre-flight disabled | REPOINT-LOC RED with `targetUntouched=False` (the stranding) while `refUntouched=True` — backstop still caught it |
| `SerializeFailure` branch removed | LOC-A RED, LOC-C green |
| remove-lane catch removed | LOC-C RED, LOC-A green |
| repoint-lane catch removed | backstop RED, showing the misattributed sub-0x800 rider |
| `PluginIsLocalized` → false | LOC-E/F RED, LOC-A/C green |
| `PluginIsLocalized` → true | LOC-B/D/G RED (over-refusal caught) |
| `LocalizedAmong` → hits everything | REPOINT-PLAIN + CONSENT RED |
| remove / forward / create pre-flight each deleted alone | LOC-H / LOC-I / LOC-J RED in isolation, the other two green |
| compact target check disabled | TARGET-LOC RED — the localized plugin compacted in place |
| compact target check fires on every target | CONSENT ×2 + REPOINT-LOC + REPOINT-PLAIN RED (over-refusal caught) |

One of these changed the branch: sabotage 9 found LOC-H/I/J **green** with the lane's pre-flight deleted, because the write's choke point refuses with the same sentence — the arms were pinning the backstop, not the call site they were named for. They now assert the consent record, which is what actually separates the two layers, and the re-run isolates each lane.

`ci-all` **124/124** at the branch tip.

## Directed fold after the rounds

The compact-target refusal above landed as a directed fold once the rounds had closed (§11 permits this; no round 3 was opened). It closed the one gap I had flagged for you rather than acted on. Restructuring came with it: both repoint arms needed a **non-localized target**, because with a localized one the new check answers first and they would have been measuring it instead of the referencer pre-flight they are named for — so the fixture gained a per-spec `Localized` flag that asserts in both directions that the plugin it built is the kind it meant to build.

Noted on #373 for the attended engagement: **both** compact lanes de-localize a localized target today, the new-file lane included — values correct when the strings resolve, but the output's nature changed either way. Whether that is ever the right product is part of the same "what may a write put on disk beside the plugin" question.

## For your review

One thing left standing, deliberately: the **recurring-prose class**. The reason the changelog kept being wrong across both rounds is that the refusal's coverage boundary is genuinely ragged — five lanes and two mechanisms, plus compact's target, which is refused for a different reason than everything else. Any prose that enumerates "the in-place lanes" will keep drifting until that boundary is one thing, which is the attended engagement's territory. Both recurring classes had their generators closed by the round-2 folds, so no §4 escalation is owed now; a third instance of either would be one.

Refs #368, #373, #374

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -6,25 +6,32 @@ when it changes.
 
 ## Unreleased
 
-- **The in-place write lanes now refuse a localized plugin instead of scrambling its text.** Any operation that
-  rewrites an existing plugin over itself — `housecarl_set_field` / `housecarl_bulk_apply` with `in_place=true`,
-  `housecarl_remove_record`, `housecarl_create_record` in place, and `housecarl_compact_plugin`'s rewrite of external
-  referencers under `repoint_externals=true` — re-serializes that plugin. A plugin flagged *localized* keeps its
-  names and descriptions in separate `.STRINGS` files and carries only numbers pointing into them; the re-serialize
-  renumbers those pointers, houseCARL saved only the plugin, and the rewritten plugin then read its text against the
-  old, unchanged `.STRINGS`. Every localized value came out attached to a different record — a weapon showing a
-  book's name — with the tool reporting success. It did not need anything unusual about where the strings lived: a
-  plugin whose `.STRINGS` sit right beside it, reading correctly everywhere else, was corrupted the same way.
+- **The in-place write lanes now refuse a localized plugin instead of scrambling its text.** Every operation that
+  rewrites an existing plugin over itself re-serializes it: `housecarl_set_field` and `housecarl_bulk_apply` with
+  `in_place=true`, `housecarl_remove_record`, `housecarl_create_record` and `housecarl_bulk_create` in place,
+  `housecarl_forward_record` in place, and `housecarl_compact_plugin`'s rewrite of external referencers under
+  `repoint_externals=true`. A plugin flagged *localized* keeps its names and descriptions in separate `.STRINGS`
+  files and carries only numbers pointing into them; the re-serialize renumbers those pointers, houseCARL saved only
+  the plugin, and the rewritten plugin then read its text against the old, unchanged `.STRINGS`. Values came back
+  attached to the wrong records — a weapon showing a book's name — with the tool reporting success. Not every value:
+  in the measured case some records kept their text, one went blank, and others picked up a different record's, which
+  is worse than a clean break, because spot-checking a few records can leave you thinking the file is fine. It did
+  not need anything unusual about where the strings lived, either: a plugin whose `.STRINGS` sit right beside it,
+  reading correctly everywhere else, was corrupted the same way.
 
   These lanes now stop before writing anything and tell you the plugin is localized and why houseCARL will not
-  re-emit it. Nothing is written, staged, or left behind. `housecarl_compact_plugin` checks the referencers it would
-  rewrite *before* it compacts anything, so it refuses the whole operation up front rather than renumbering your
-  plugin and then stopping partway through the plugins that point at it.
+  re-emit it. Nothing is written or staged, and your one-time in-place consent for that plugin is not spent on a
+  write that cannot happen. A dry run gives the same refusal the real call does, rather than reporting an edit that
+  would then be refused. `housecarl_compact_plugin` checks the referencers it would rewrite *before* it compacts
+  anything, so it refuses up front rather than renumbering your plugin and then stopping partway through the plugins
+  that point at it — and it refuses before asking you to confirm the rewrite, instead of after.
 
-  Only in-place rewrites are affected. Everything that writes a NEW plugin — the default patch lane,
-  `housecarl_merge_plugins`, and `housecarl_compact_plugin`'s default new-file lane — builds its output fresh with
-  the text stored inside the plugin, and is unaffected. In practice modders overwrite official masters and
-  translation mods rarely if ever, preferring new override plugins, so the refusal should be rare.
+  Writing a NEW plugin is unaffected: the default patch lane, `housecarl_merge_plugins`, and
+  `housecarl_compact_plugin`'s default new-file lane all build their output fresh with the text stored inside the
+  plugin. (Extending an existing patch with `into=` re-serializes that patch, so it is in-place in this sense — but
+  it only ever targets a folder houseCARL owns, and houseCARL does not author localized plugins.) In practice modders
+  overwrite official masters and translation mods rarely if ever, preferring new override plugins, so the refusal
+  should be rare.
 
   Making these lanes rewrite a localized plugin *correctly* is a larger change — it decides what an in-place edit is
   allowed to write besides the plugin file itself — and is being designed separately. Until then houseCARL refuses
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -28,9 +28,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
   the strings lived, either: a plugin whose `.STRINGS` sit right beside it, reading correctly everywhere else, was
   corrupted the same way.
 
-  These lanes now stop before writing anything and tell you the plugin is localized and why houseCARL will not
-  re-emit it. Nothing is written or staged, and your one-time in-place consent for that plugin is not spent on a
-  write that cannot happen. A dry run gives the same refusal the real call does, rather than reporting an edit that
+  These lanes now stop before writing anything and tell you the plugin is localized, why houseCARL will not re-emit
+  it, and what to do instead where there is something to do. Nothing is written or staged, and your one-time
+  in-place consent for that plugin is not spent on a write that cannot happen — the one exception being a plugin
+  whose header cannot be read at that moment (a lock from MO2 or xEdit, say), which the check treats as
+  not-localized: the call then proceeds, spends the acknowledgement, and is refused at the write instead. The file
+  is untouched either way. A dry run gives the same refusal the real call does, rather than reporting an edit that
   would then be refused. `housecarl_compact_plugin` checks the referencers it would rewrite *before* it compacts
   anything, so it refuses up front rather than renumbering your plugin and then stopping partway through the plugins
   that point at it — and it refuses before asking you to confirm the rewrite, instead of after.
@@ -41,16 +44,18 @@ saying it sets an expectation their install may contradict. Say what is known, a
   houseCARL read it, with the mod's `.STRINGS` set left describing nothing and any other language it shipped gone.
   If the strings resolved nowhere at all (see the bound described below), blanks get baked in instead. Neither is
   something to do silently to a file with no review step and no undo, so that lane now refuses up front and points at
-  the new-file lane. Compacting to a new plugin is unaffected and keeps your text — that output is not localized
-  either, so read it before you enable it.
+  the new-file lane, which now says the same thing in its own report rather than only when you are refused.
+  Compacting to a new plugin keeps your text wherever houseCARL can resolve your strings — see the bound below; its
+  output is not localized either, so read it before you enable it.
 
   Writing a NEW plugin is otherwise unaffected: the default patch lane, `housecarl_merge_plugins`, and
   `housecarl_compact_plugin`'s default new-file lane all build their output fresh with the text stored inside the
   plugin, and leave your original alone. (Extending an existing patch with `into=` does re-serialize that patch, so
   it is in-place in this sense — but it only ever targets a folder houseCARL owns, and houseCARL does not author
-  localized plugins.) The refusal fires on any plugin carrying the localized header flag, which in practice is
-  mostly official masters and translation mods — and those are edited by making a new override plugin, not in place,
-  so it should be rare.
+  localized plugins.) The refusal fires on any plugin carrying the localized header flag. Which of your plugins
+  carry it depends on your install: the official masters always do, and translated releases commonly do, because
+  that is what the translation toolchain emits. No houseCARL tool reports the flag today, so there is currently no
+  way to answer "will this hit me?" from inside houseCARL — xEdit shows it on the file header.
 
   Making these lanes rewrite a localized plugin *correctly* is a larger change — it decides what an in-place edit is
   allowed to write besides the plugin file itself — and is being designed separately. Until then houseCARL refuses
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -6,18 +6,20 @@ when it changes.
 
 ## Unreleased
 
-- **The in-place write lanes now refuse a localized plugin instead of scrambling its text.** Every operation that
-  rewrites an existing plugin over itself re-serializes it: `housecarl_set_field` and `housecarl_bulk_apply` with
-  `in_place=true`, `housecarl_remove_record`, `housecarl_create_record` and `housecarl_bulk_create` in place,
-  `housecarl_forward_record` in place, and `housecarl_compact_plugin`'s rewrite of external referencers under
-  `repoint_externals=true`. A plugin flagged *localized* keeps its names and descriptions in separate `.STRINGS`
-  files and carries only numbers pointing into them; the re-serialize renumbers those pointers, houseCARL saved only
-  the plugin, and the rewritten plugin then read its text against the old, unchanged `.STRINGS`. Values came back
-  attached to the wrong records — a weapon showing a book's name — with the tool reporting success. Not every value:
-  in the measured case some records kept their text, one went blank, and others picked up a different record's, which
-  is worse than a clean break, because spot-checking a few records can leave you thinking the file is fine. It did
-  not need anything unusual about where the strings lived, either: a plugin whose `.STRINGS` sit right beside it,
-  reading correctly everywhere else, was corrupted the same way.
+- **The record-editing lanes now refuse a localized plugin in place instead of scrambling its text.** Editing,
+  removing from, creating into, or forwarding into an existing plugin **in place** re-serializes that whole plugin —
+  `housecarl_apply`, `housecarl_create`, `housecarl_remove` and `housecarl_forward` with `in_place="X.esp"` (and
+  their 1.x spellings `housecarl_set_field` / `housecarl_bulk_apply` / `housecarl_create_record` /
+  `housecarl_bulk_create` / `housecarl_remove_record` / `housecarl_forward_record` with `in_place=true`), plus
+  `housecarl_compact_plugin`'s rewrite of external referencers under `repoint_externals=true`. A plugin flagged
+  *localized* keeps its names and descriptions in separate `.STRINGS` files and carries only numbers pointing into
+  them; the re-serialize renumbers those pointers, houseCARL saved only the plugin, and the rewritten plugin then
+  read its text against the old, unchanged `.STRINGS`. Values came back attached to the wrong records — a weapon
+  showing a book's name — with the tool reporting success. Not every value: in the measured case some records kept
+  their text, some went blank, and others picked up a different record's, which is worse than a clean break, because
+  spot-checking a few records can leave you thinking the file is fine. It did not need anything unusual about where
+  the strings lived, either: a plugin whose `.STRINGS` sit right beside it, reading correctly everywhere else, was
+  corrupted the same way.
 
   These lanes now stop before writing anything and tell you the plugin is localized and why houseCARL will not
   re-emit it. Nothing is written or staged, and your one-time in-place consent for that plugin is not spent on a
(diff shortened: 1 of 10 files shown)


src/housecarl-core/WriteEngine.cs
```diff
@@ -3822,6 +3848,53 @@ public NullArmSerializeException(Exception inner)
...
+    /// <summary>The refusal sentence, in ONE place — the exception below throws it, and the lanes that predict the
+    /// refusal BEFORE reaching the write (a dry run, or the consent gate) state it without throwing. Two spellings of
+    /// this would drift, and the dry run's whole job is to give the same answer as the real call.
+    ///
+    /// <para>"would no longer reliably be its own record's" and not "would all land on the wrong record": the probe
+    /// measures a MIX — of five localized weapons in arm C, two kept their own text exactly, one went blank, and two
+    /// read a book's. The absolute was the easier sentence to write and the more dangerous one to act on, because a
+    /// caller who checks a few records, finds them intact, and concludes the diagnosis does not fit their file is
+    /// looking at precisely the failure being described.</para></summary>
+    public static string Text(string pluginFileName)
```
Avick3110: **[medium] The refusal names no next step.**

This is the one sentence four lanes render verbatim, and it ends at "houseCARL refuses the write instead of corrupting the file." Every other refusal in these lanes names an action — including the compact refusal 150 lines away, which is careful enough about its remedy that the PR body documents measuring one and discarding it.

Three of the four lanes it serves have a one-liner: `housecarl_apply` / `housecarl_create` / `housecarl_forward` all have a default lane that writes a new override plugin and leaves the original alone, and that lane is not affected by this defect at all. A modder who asked to edit `Dawnguard.esm` in place is told only that houseCARL will not, and has to infer that dropping `in_place=` still gets them the edit.

`housecarl_remove` is the one lane where there genuinely is no equivalent — you cannot un-define a record from an override. That asymmetry is itself the clause worth stating, rather than leaving all four silent.

plugin/CHANGELOG.md
```diff
@@ -6,6 +6,49 @@ when it changes.
...
+  anything, so it refuses up front rather than renumbering your plugin and then stopping partway through the plugins
+  that point at it — and it refuses before asking you to confirm the rewrite, instead of after.
+
+  **`housecarl_compact_plugin(in_place=true)` refuses a localized plugin too**, for a different reason. A compaction
+  does not re-serialize your plugin, it builds a fresh one and writes that over the original — so it never hits the
+  check above, and what it would produce is a **non-localized** plugin carrying whichever language resolved when
+  houseCARL read it, with the mod's `.STRINGS` set left describing nothing and any other language it shipped gone.
+  If the strings resolved nowhere at all (see the bound described below), blanks get baked in instead. Neither is
+  something to do silently to a file with no review step and no undo, so that lane now refuses up front and points at
+  the new-file lane. Compacting to a new plugin is unaffected and keeps your text — that output is not localized
```
Avick3110: **[medium] "keeps your text" contradicts the #362 entry four paragraphs below.**

The #362 bullet in this same release block spells out the bound: the game-Data fallback applies *only* when the plugin's own folder carries no strings source at all, and **any** `.bsa` beside the plugin counts as one. A localized plugin in a mod folder shipping an asset-only `.bsa`, with its `.STRINGS` served from elsewhere — and a plugin whose strings are in neither place — "is therefore still read the old way and still writes blanks."

The new-file compact lane is exactly that path. `CompactBuild` opens the source through `LoadOrderResolver.OpenOverlay`, the fallback does not fire, every `FULL`/`DESC` reads empty, and `RenumberModInto` copies the blanks into P′. So for those cases the new-file lane writes a plugin with no names or descriptions — which the entry below calls the unfixed residual and this sentence calls "keeps your text."

The refusal message itself is more careful (it says "That output is not localized either — read it before you swap it in", and never claims the text survives). It is the changelog that overclaims. This is the recurring-prose class the PR body names, in a sentence added after the rounds closed.

A scoped version would be: "Compacting to a new plugin is unaffected — it keeps your text wherever houseCARL can resolve your strings (see the bound below), and its output is not localized either, so read it before you enable it."

plugin/CHANGELOG.md
```diff
@@ -6,6 +6,49 @@ when it changes.
...
+  either, so read it before you enable it.
+
+  Writing a NEW plugin is otherwise unaffected: the default patch lane, `housecarl_merge_plugins`, and
+  `housecarl_compact_plugin`'s default new-file lane all build their output fresh with the text stored inside the
+  plugin, and leave your original alone. (Extending an existing patch with `into=` does re-serialize that patch, so
+  it is in-place in this sense — but it only ever targets a folder houseCARL owns, and houseCARL does not author
+  localized plugins.) The refusal fires on any plugin carrying the localized header flag, which in practice is
+  mostly official masters and translation mods — and those are edited by making a new override plugin, not in place,
```
Avick3110: **[medium] "should be rare" is an English-install claim presented as a general one.**

On an English install this reads right: the localized flag is the official masters plus the odd translation mod. On a translated one it is not — Russian, German, French and Polish releases routinely ship the flag plus `.STRINGS` for ordinary third-party mods, because that is what the translation toolchain emits. For those users the four record-editing lanes stop accepting `in_place=` across a large share of their order, not at an edge case.

The refusal is still right — the write would corrupt the file, so breadth is not the question there. What this sentence does is set an expectation the user's install contradicts, and it gives them nothing to check against beforehand. `housecarl_load_order_status` does not surface the flag either, as far as I can see, so a modder cannot answer "is this going to hit me" without opening plugins in xEdit.

Two things would close it: drop the frequency claim to what is actually known ("which plugins carry it depends on your install — official masters always do, and translated releases commonly do"), and say how to find out. This is also the breadth question that matters more than the compaction one, since apply/create/remove/forward are used far more than compact.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5982,6 +6008,35 @@ public WritePatchBuilder.CompactOutcome CompactPlugin(
...
+            // the distinction this whole refusal rests on.
+            if (inPlace && WriteEngine.PluginIsLocalized(srcPath))
+                return WritePatchBuilder.CompactOutcome.Fail(
+                    $"houseCARL did not compact '{name}' in place — the file is unchanged and nothing was staged. It is " +
+                    "flagged LOCALIZED, so its text lives in separate .STRINGS files rather than in the plugin. A " +
+                    "compaction does not rewrite the plugin, it builds a NEW one and writes that over the original, and " +
+                    "the rebuilt plugin is not localized: what it would carry is whichever language resolved when " +
+                    "houseCARL read it, written into the plugin itself — or blanks, if those strings resolved nowhere. " +
+                    "Either way the file the game loads would stop being the translated plugin you have, with no backup " +
+                    $"and nothing to undo it. Re-run without in_place to compact '{name}' into a NEW plugin instead: the " +
```
Avick3110, line 6038: **[medium] The remedy routes to the lane that does the same thing — and that lane says nothing.**

The refusal's reasoning is that a de-localized rebuild "may not happen silently" on a file with no review step. But `CompactBuild` builds `pPrime = new SkyrimMod(...)` in both lanes and `RenumberModInto` copies no header flags, so the new-file lane produces the *identical* de-localized plugin. The only difference the refusal can actually point to is where it lands.

That is disclosed here, in the trailing clause. It is not disclosed anywhere else: a modder who just runs `housecarl_compact_plugin('X.esp')` on a localized source — never meeting this refusal — gets a de-localized plugin and no note in the outcome. Nothing in the compact lane emits one (`grep -i localiz` over `LoadOrderService.cs` finds only the refusals). So the standard the refusal invokes is applied to one lane and not the other, and the quiet lane is the one it recommends.

The PR body already notes this for #373. The cheap half is a note on the new-file lane when the source carried the flag — same sentence, non-fatal — which would make the remedy honest at the point the user takes it rather than only at the point they are refused.

(Separately: "A compaction does not rewrite the plugin, it builds a NEW one and writes that over the original" reads as a contradiction on first pass. "does not re-serialize your plugin" is what the clause means.)

src/housecarl-core/WriteEngine.cs
```diff
@@ -1821,6 +1826,27 @@ public static void WriteInPlace(SkyrimMod targetMod, IReadOnlyList<ISkyrimModGet
...
+    /// <summary>Is the plugin at <paramref name="path"/> flagged LOCALIZED? The pre-flight form of
+    /// <see cref="WriteInPlace"/>'s choke point, for the lanes that must answer BEFORE they reach the write — a dry
+    /// run (whose job is to give the same answer the real call would) and the consent gate (which must not spend a
+    /// modder's one-time in-place acknowledgement on a write that is going to refuse).
+    ///
+    /// <para>Reads the header only, and deliberately with the BARE overlay: the localized FLAG is in the header, so
+    /// unlike reading the strings themselves this needs no game-Data fallback and cannot be wrong about the flag
+    /// because a strings source is missing. Any fault answers "not localized" and leaves the write's own choke point
+    /// to refuse — a pre-flight that threw would turn a readable-header failure into a different refusal.</para></summary>
+    public static bool PluginIsLocalized(string path)
```
Avick3110, line 1838: **[low] The fail-open makes the unhedged consent guarantee falsifiable.**

First, the part that checks out: the dry-run/real divergence is genuinely closed, and not by two mechanisms agreeing. Both the dry run and the real call go through *this same line* in the service — it sits above the dry-run branch — so there is nothing to drift. The write's own choke point reads `targetMod.UsingLocalization`, and in Mutagen 0.53.1 that property has no backing field in the assembly, i.e. it is computed off `ModHeader.Flags.Localized`, the same bit this reads. Two readings of one bit, not two predictions.

The one asymmetry is here: any fault answers "not localized". The changelog states the consent guarantee flat — "your one-time in-place consent for that plugin is not spent on a write that cannot happen." Falsifying it needs the bare `CreateFromBinaryOverlay` to throw where `WritePatchBuilder`'s `CreateFromBinary` a moment later succeeds: a transient share/lock fault from MO2 or xEdit touching the file, an mmap failure. Then the gate records the acknowledgement and the choke point refuses.

The PR body says the guarantee stays unhedged because falsifying it "needs a header read to throw on a plugin the resolver just enumerated" — but the resolver enumerated it at index time, not at this instant, and the two opens here are milliseconds apart on a file MO2 may be touching. Narrow, and the file is untouched either way; a hedge in the changelog sentence is cheaper than a tri-state pre-flight.

Avick3110:
## Review fold — 5 findings, **5 folded, 0 refused**

Every finding cited a line, and every line said what it was reported to say. That is the refusal rate: nothing was refused because there was nothing to refuse, not because the triage waved them through — the two I would have argued (F3's breadth, F5's fail-open) both reproduced exactly as described, and F4's "the quiet lane is the one the refusal recommends" is the finding I should have drawn myself when I wrote the #373 note and did not.

Four commits, in the order the ruling set.

### The class first (`c3324e3`, take-or-drop, on its own commit)

F2, F3 and F5 are one class — a changelog sentence claiming more than the code does — and it is the class the PR body already flagged as having recurred across both rounds. So the rule goes in before the sentences that break it: **state what the tool does or refuses; point at bounds rather than restating them; no frequency or coverage adjectives.** Two of the three findings were second copies of a bound stated four paragraphs below, and the copies were the halves that went wrong.

### F1 — each lane names its own next step (`e11edf4`)

The shared mechanism sentence stays in one place; each lane passes its own clause at the call site.

- **apply / create / forward** append the default-lane clause. **Measured before named**, per the remedy rule: a new `REMEDY` arm runs the default lane against the localized fixture and the original comes back byte-untouched with the edit in a new plugin. It deliberately does *not* assert the text survived — that is the bound, and the clause never claimed it.
- **remove** appends the asymmetry instead and names no remedy: a separate plugin can override a record but cannot un-define one. Nothing there for #356's remedy rework to collide with.

Dry-run parity holds by construction — both the dry run and the real call return from the same service line — and is now *armed* rather than argued: LOC-E and LOC-F assert the remedy text on both sides, so a clause appearing on one and not the other fails.

### F4 — the quiet lane speaks (`e11edf4`)

The new-file compact lane emits a non-fatal note when the source carried the flag. Own-behaviour only: the output is not localized and carries whatever this read produced. It does **not** claim the strings resolved, because the strings-nowhere case writes blanks through the same path and the sentence has to stay true there. Plus your wording fix — "does not re-serialize your plugin".

### F2 / F3 / F5 — the three sentences (`106b951`)

- **F2** scoped to the bound and pointing at it: "keeps your text wherever houseCARL can resolve your strings — see the bound below".
- **F3** — the frequency claim is gone. What replaces it was **measured first**, as directed: `ModHeader.Flags` is referenced nowhere in the MCP layer, so **no shipped tool reports the flag**. The entry says exactly that rather than naming a tool that would not answer, and points at xEdit's file header. Filed as **#376** (unrouted; `load_order_status` named as the natural home, deliberately not built here).
- **F5** — the consent guarantee now states its bounded exception: a header read that faults is treated as not-localized, so the call proceeds, spends the acknowledgement, and is refused at the write. File untouched either way. No tri-state pre-flight.

### Arms and sabotages

Five new sabotages, each from a committed state, each restored and verified by anchor:

| Sabotage | Result |
|---|---|
| remove borrows the default-lane remedy | LOC-H RED, LOC-I/LOC-J green |
| apply drops its clause | LOC-E **and** LOC-F RED (parity caught on both sides) |
| new-file note fires unconditionally | non-localized arm RED |
| new-file note suppressed | localized arm RED |
| choke point forced true | 13 in-place arms RED, **REMEDY stays green** — confirming the default lane is genuinely independent of this defect, which is what its clause claims |

18 sabotages across the engagement. `ci-all` **124/124** at the tip.

One structural consequence worth naming: the compact guard now generates a corpus, because the REMEDY arm drives the default write lane through the rulebook. It is written outside the guard's temp root and the previous `CorpusPath` is restored in the `finally` — `CorpusPath` is process-global, and leaving it pointing into a deleted directory would break whichever probe `ci-all` runs next.

No round after the directed fold, per §11. Holding for your explicit go to merge.

Issue #378, opened 2026-08-19
[Bug] in-place removal records the one-time consent before the not-carried check, so a refused call spends it


Avick3110:
## Summary

The in-place removal lane persists the one-time consent acknowledgement **before** it checks whether the target actually carries the records. A call that is then refused — nothing written — has still spent the plugin's first-touch confirmation, so the next in-place write to that plugin proceeds with no prompt.

## Measured

`LoadOrderService.RemoveRecordsInPlace` runs:

- (2) CONSENT — `RecordInPlaceAcknowledged(targetPath)` when `acknowledge=true`
- (3) writable-parent pre-flight
- (4) the write — and the present-check ("not carried by …") lives inside `WritePatchBuilder.RemoveRecordsInPlace`, at (4)

So:

```
housecarl_remove(formids=["000800:RvMaster.esm"], in_place="Bystander.esp", acknowledge=true)
>>> error: refused — 1 of 1 target(s) not carried by 'Bystander.esp'; NOTHING removed

housecarl_remove(formids=["000800:Bystander.esp"], in_place="Bystander.esp")   # no acknowledge=
>>> removed 1 record from Bystander.esp IN PLACE (42 bytes; 0 records remain —
    your ORIGINAL file was rewritten; no houseCARL backup or undo)
```

The second call is the first real rewrite of that plugin and it was never confirmed.

## Why this is a defect rather than a design choice

The lane already holds the opposite standard explicitly. The localized-target refusal is placed *ahead* of the consent gate, and the code comment states the reason: the gate would otherwise "spend the modder's one-time in-place acknowledgement on a write that was never going to happen." The present-check is the same shape and is on the wrong side of the gate.

The same ordering appears in the sibling in-place lanes (edit / create / forward each record consent before their own write-time refusals); whether they have an equivalent never-going-to-happen check is worth establishing as part of the fix rather than assumed.

## Provenance

Found by a review agent while reviewing #356's branch, in code that branch does not touch — the ordering is unchanged on `main`. Per CLAUDE.md §2 that puts it in this lane rather than in the PR.

Worth noting for prioritisation: #356's fix makes this arm easier to reach. A caller who mistyped a *patch* name is now told to name a *plugin* for the in-place lane, and a plausible-but-wrong filename there is exactly the call that burns the consent without writing.

*Unrouted — needs a RUN_ORDER routing stroke.*

Avick3110:
[...]

Issue #377, opened 2026-08-19
[Bug] the ambiguous and multi-plugin extend refusals name into=, which housecarl_remove_record does not declare


Avick3110:
## Summary

The shared `into=`-extend resolver's **ambiguous** and **multi-plugin** refusal arms both instruct the caller to "pass into=". `housecarl_remove_record` is the one write tool reaching that resolver which declares `patch=` and **no** `into=` — so on that tool the remedy names a parameter the caller cannot pass.

This is the same class as #356 (a refusal whose remedy the caller cannot perform), one arm over.

## Measured

`housecarl_remove_record` (`WriteTools.cs`) declares `formid, patch, target, in_place, acknowledge` — no `into=`. Driving the ambiguous arm (two houseCARL-owned folders carrying the same `.esp`):

```
error: cannot extend: 2 houseCARL folders carry 'SeedA.esp' — ambiguous, refusing to guess.
Pass the CONTAINING mod-folder name as into= to pick one (folder & plugin names need not
match): into="houseCARL - DupHome"  |  into="houseCARL - SeedA Renamed".
```

The multi-plugin arm (`SoleEspInFolder`, `LoadOrderService.cs`) ends the same way: `name the one to extend by passing its filename as into=`.

Nothing rescues it: `AliasTable` has no row whose OLD spelling is `into`. (The table is bidirectional — it carries new→old rows too — so this is that one row's absence, not a one-way rule.) Measured through the shim: `housecarl_remove_record{formid, into}` answers `unknown parameter: into. This tool accepts only: formid, patch, target, in_place, acknowledge.` The retired-tool hint does not apply either — `housecarl_remove_record` is still registered (`WithToolsFromAssembly`), and `RetiredToolHint` fires only for names that are *not* registered.

## Why it is filed rather than fixed

Found while reviewing the branch for #356, in code that branch does not touch — both arms are unchanged on `main`. Per CLAUDE.md §2 that puts it in this lane rather than in the PR.

## Note on shape

The fix is not simply re-spelling the sentence: the resolver is shared, and the two tools reaching this arm spell the lane differently. #356's PR establishes the pattern — the calling tool hands its own spelling down, because the service cannot tell its callers apart. These two arms would take the same treatment, and both are also reached by tools that *do* declare `into=`, so the sentence has to stay correct for them.

*Filed by a build session while reviewing #356's branch. Unrouted — needs a RUN_ORDER routing stroke.*

Avick3110:
[...]

Issue #359, opened 2026-08-17
[Bug] the un-owned-folder refusal still dead-ends — it names no parameter and no lane


Avick3110:
## Summary

#343 gave the *not-found* arm of `ResolveOwnedPatchFolder` a remedy naming `patch=`. Its sibling — the **un-owned folder** arm, same method, same callers — still ends **"Use a different patch name."**, naming no parameter and no lane switch. That is the pre-#343 dead end, on a path a caller reaches at least as often.

## Concrete scenario (Aaron, reviewing PR #358)

A modder has an ordinary, non-houseCARL MO2 mod folder literally named `My Cool Patch`, and calls `housecarl_apply(..., into="My Cool Patch")`. The step-(4) loop over `{"houseCARL - My Cool Patch", "My Cool Patch"}` hits the un-owned candidate **first** and throws before reaching the not-found arm. The caller is told to pick a different name — so they guess another `into=` (still the wrong lane), or drop `into=` and get `Patch.esp`. That is exactly the shape #343 describes.

## Measured

Against the real service on a synthetic MO2 instance, at PR #358's head:

```
foreign folder "My Cool Patch" (no marker) holding an INACTIVE "My Cool Patch.esp"

into="My Cool Patch"
>>> cannot extend: mod folder 'My Cool Patch' exists but was NOT created by houseCARL (no marker) —
    refusing to write into a folder houseCARL doesn't own (originals untouched, Q3).
    Use a different patch name.

then patch="My Cool Patch"
>>> success=True  ...\mods\houseCARL - My Cool Patch\My Cool Patch.esp
```

## Why this is remedy design, not a sentence fold

Two things make the obvious fix wrong, which is why it was routed here rather than folded into #358:

1. **The caller's intent is ambiguous at this arm.** Reaching an un-owned folder by name means they may have meant *edit that existing plugin* (the consent-gated `in_place=` lane) rather than *create a fresh patch*. A remedy naming only `patch=` picks one reading for them.
2. **`patch=` with the same stem shadows the foreign plugin.** As measured above, `UniqueStem` accepted `My Cool Patch` because `IsStemFree` tests only for an existing `houseCARL - <stem>` folder and for an **active** plugin of that name — the foreign folder's plugin was inactive on both counts. So houseCARL mints `My Cool Patch.esp` beside a foreign `My Cool Patch.esp`, two plugins that cannot both be active. A remedy that hands the caller their own stem back would be walking them into that.

[...]

PR #379, opened 2026-08-19
fix(remove): the not-found refusal names a lane removal has, not one it hasn't

### LINKED ISSUE #356: [Bug] housecarl_remove's "cannot extend" refusal offers a remedy removal cannot perform

Avick3110:
RUN_ORDER step 3, item 9. The remove lane's "cannot extend" refusal ended with a remedy removal cannot perform.

## What was wrong

`housecarl_remove(..., into="My Patch.esp")` naming a patch that does not exist refused with:

> cannot extend: no houseCARL plugin 'My Patch.esp' … **Omit into= to create it fresh, or check the name.**

Omitting the lane is itself refused. Both refusals were reproduced verbatim on `main` at `6b16d41` before anything was changed — the issue's 2026-08-17 measurement still stood after #358, #365, #370 and #375 landed.

The tail came from the shared `into=`-extend resolver, whose default was the weaker of #343's two sentences, defended in the rule comment as *"merely less helpful, never wrong"*. It was wrong for a whole class of caller — the ones that cannot create a patch — and removal is that class.

## What changed

**The default claims no fresh-write path at all.** `bool patchNamesFresh` became `FreshPatchRemedy` with three states, defaulting to `None` → `"Check the name."`. Each lane that can create one now states so for itself: `apply`/`create`/`forward` that `patch=` names a new patch, the two copy lanes and every rider lane that omitting `into=` does. A caller added later without a thought gets a sentence that is true, rather than one that happens to fit the lanes that existed when it was written.

Sweep confirmed removal was the only non-creating caller reaching the old default (7 call sites).

**Removal states its own next step**, handed in at the call site — the shape `LocalizedTargetUnsupportedException.Text` already uses. It says there is no create route here and names the lane a removal does have.

**Each remove *tool* supplies its own spelling for that lane.** `housecarl_remove` declares `in_place="<plugin filename>"` and no `target=`; `housecarl_remove_record` declares `target=` plus a bool. The service cannot tell its two callers apart, so it renders what it is handed and names no spelling when called directly.

## Measurement

Every remedy sentence was measured against the real service before it was written, and the ones a caller acts on were measured again through the tools.

- **The issue's own suggested remedy is false.** "Create the patch first, then remove from it" — a patch created here carries no override of the record, and removal refuses with `not carried by patch 'MadeItFirst.esp'`. Not written.
- **The in-place lane is a true remedy.** Following it reaches the first-touch confirmation; acknowledging it removes the record, verified gone from the file on disk.
- **No tool lists houseCARL-owned patches**, so `"Check the name."` points nowhere and stays bare rather than naming a tool that cannot answer.
- **The `patch is required` refusal is 1.x-only reachable** — `housecarl_remove` refuses "no lane named" first, in its own spelling — which is what makes hardcoding the legacy wording there safe. Arm-pinned, because that unreachability is the load-bearing fact.

## Review rounds

**Round 1 — 3 agents, 11 findings: 7 folded, 2 refused, 2 taken as corrections.**

All three independently found the same defect, and it was the branch's: the reused in-place sentence was measured at the **service** layer, the one altitude where it is trivially correct. At the altitude a caller reads it, `housecarl_remove` declares no `target=`, so the refusal named a parameter that tool does not have. It resolved only because the 1.x→2.0 lane shim silently re-spells the old pair, and that shim is scaffolding due to retire — the tool's own sibling refusal already used the right spelling, so one response could hand a caller both. This is #356's own class inside the fix for #356.

Folded by moving the spelling to the tools (above). The arms moved to the tool altitude with it — the previous ones called the service in its 1.x argument shape and structurally could not see this class. The `[MustState]` reach check then caught the same mistake a second time inside the probe, where the observation drove the service instead of the tool.

Agent C's medium, also folded: the branch converted implicit defaults into explicit per-lane statements, and sabotaging two of them stayed green — `copy` and `copy_npc_appearance` could silently lose their create clause. Now pinned in their own probes, with `forward` and `create` added to the extend-resolver guard. Writing those arms immediately caught a fixture error: an earlier arm had already created the patch name being tested as absent.

*Refused, with grounds:* that the remedy should pre-empt a caller who **adds** the in-place lane rather than replacing `into=` — the sentence says "instead", and the resulting collision refusal names the fix itself; and that the not-found arm should withhold the in-place clause when `into=` happens to name a foreign **active** plugin — that lane is consent-gated, and whether a refusal may recommend it at all is #359's fork, out of scope here.

**Round 2 — 2 agents, 10 findings: 3 folded, 2 filed, 1 escalated, rest noted.** (A third agent died on an API error before doing any work and was relaunched fresh.)

Folded: the no-create clause was predicting what a *next* call would hold ("a patch created now would have nothing to remove") — falsifiable in two calls by `apply(patch="X")` then `remove(into="X")`, which the branch had itself measured. It nets out to nothing, which is why it was written, but predicting a subsequent call's result is the #343/#358 class this lane exists to stay out of; it now states the rule and stops. The changelog carried the same overreach. And the doctrine comment justified #377 with a false supporting fact (that the alias table runs old→new only — it is bidirectional; the miss is that no row has `into` as its OLD spelling).

**Stop rule:** round 2 hit §5 #11's recurring-class terminator, so there is no round 3 — see below.

**14 sabotage RED-checks**, each from a committed state, each restored and verified by grepping for a string the fold introduced. Every branch of the new conditional is checked in both directions: default regains the create clause, lane clause dropped, shared clause re-inlined, each of the three enum states removed from its lane, the 2.0 tool handing down the legacy spelling, and each new constant gutted.

`ci-all` 124/124 green on the branch.

## Filed rather than fixed here

Both are confirmed defects in code this branch does not touch (CLAUDE.md §2, the #324 shape). Both are **unrouted** — the routing-and-milestone stroke is Aaron's.

- **#377** — the resolver's *ambiguous* and *multi-plugin* arms tell every caller to `pass into=`, which `housecarl_remove_record` does not declare and no alias row bridges. This branch's own class, one arm over.
- **#378** — the in-place removal lane records the one-time consent *before* the not-carried check, so a refused call spends it and the next in-place write to that plugin proceeds unprompted. The localized refusal sits ahead of that gate for exactly this stated reason. Worth noting for prioritisation: this fix makes that arm easier to reach, since a caller who mistyped a *patch* name is now told to name a *plugin*.

## Deposited on #359 — the remedy's consent caveat, and the fallback itself

Aaron's review (F1, high) found the earlier version of this section priced the remedy wrongly. It said the
in-place lane "cannot fire without an explicit confirmation naming the file — a wasted round trip, not a write."

**That is true only for a plugin that has never been acknowledged.** The in-place consent is persisted, survives
the session, and is shared across the edit / create / remove in-place lanes. A caller who follows this remedy
against a plugin they acknowledged at any earlier point, in any lane, **rewrites the original with no prompt at
all**. The typo case is where it bites: a caller who mistyped a *patch* name is pointed at a *plugin* filename,
and the filename in their hand is usually one of their own mods.

The refusal sentence is deliberately unchanged. Whether the recommendation must carry its own consent caveat —
given that the lane's first-touch prompt may already be spent — is #359's chartered fork, and the finding is
deposited there.

The false pricing has been corrected in every home it had on this branch, not just here: the constant's own doc,
the call-site comment, the probe comment, and the changelog entry, which now states plainly that an
already-acknowledged plugin is rewritten with no further prompt.

Also recorded on #359 (Aaron's F3 — recorded, not folded): clause **order** was never the real variable. The
always-true fallback is nearly empty of content, so wherever it sits it reads as filler while the lane clause does
the work. What the shared fallback should *be* is the same refusal-remedy fork, and belongs with it.

The two consecutive ordering findings that stopped the rounds under §5 #11 stand as reported; nothing was folded a
third time.

Fixes #356

### DIFF
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -5572,7 +5572,7 @@ public sealed class LoadOrderService : IDisposable
     /// → mod.Remove → re-serialize, with clean-masters riding along). The default lane never touches originals (only the
     /// patch folder is written).</summary>
     public WritePatchBuilder.RemovalOutcome RemoveRecords(IReadOnlyList<string> formids, string? patch,
-        string? target = null, bool inPlace = false, bool acknowledge = false)
+        string? target = null, bool inPlace = false, bool acknowledge = false, string? inPlaceRemedy = null)
     {
         if (formids is null || formids.Count == 0)
             return WritePatchBuilder.RemovalOutcome.Fail("no formids supplied — pass the FormID(s) of the record(s) to remove.");
@@ -5592,8 +5592,11 @@ public sealed class LoadOrderService : IDisposable
                 "target= is only meaningful with in_place=true (it names the plugin to remove from in place). For the default lane omit target=; use patch= to name the houseCARL patch.");
         if (!inPlace && string.IsNullOrWhiteSpace(patch))
             return WritePatchBuilder.RemovalOutcome.Fail(
+                // The legacy spelling is hardcoded here because this arm is 1.x-ONLY: housecarl_remove refuses
+                // "no lane named" — in its own correct spelling — before the service is reached. Measured, and
+                // pinned by the extend-resolver guard's tool-altitude arm.
                 "patch is required — name the houseCARL patch to remove the record from (removal only targets a patch that already carries it). "
-                + WriteSentences.RemoveInPlaceAlternative);
+                + WriteSentences.RemoveInPlaceLaneLegacy);
 
         // Parse every formid first, collecting ALL problems (all-or-nothing, like the edit path). Pure — outside the gate.
         var keys = new List<FormKey>(formids.Count);
@@ -5625,9 +5628,14 @@ public sealed class LoadOrderService : IDisposable
             // lane is itself refused, so the sentence sent callers to a second refusal. What this lane can honestly
             // offer instead is measured — a patch created first carries no override of the record, so removal refuses
             // there too; the in-place lane does remove it, behind the first-touch confirmation.
+            //
+            // The in-place half comes from the TOOL (inPlaceRemedy), because the two tools spell that lane
+            // differently and the service cannot tell which one called it. Null — a direct service call — offers
+            // only the half that is true at every altitude, rather than guessing a spelling.
             string outPath;
             try { outPath = ResolveOutputPath(patchName: null, into: patch, out _, out _,
-                                              laneClause: WriteSentences.RemoveNoFreshPatch + " " + WriteSentences.RemoveInPlaceAlternative); }
+                                              laneClause: WriteSentences.RemoveNoFreshPatch
+                                                          + (inPlaceRemedy is null ? "" : " " + inPlaceRemedy)); }
             catch (Exception ex) { return WritePatchBuilder.RemovalOutcome.Fail(ex.Message); }
 
             return WritePatchBuilder.RemoveRecords(resolver, keys, outPath);
@@ -8015,6 +8023,9 @@ public sealed class LoadOrderService : IDisposable
         // THIS arm only. The ambiguous and multi-plugin arms above already name a working call, and the foreign-folder
         // arm's remedy is #359's open design question, not this one's.
         //
+        // It is rendered BEFORE the fallback above, not after: the lane's own diagnosis is what makes the fallback
+        // the right thing left to do, and reading the instruction first was the order two reviewers stumbled on.
+        //
         // Note what the sentence does NOT do: predict the resulting filename. UniqueStem takes a stem only when it
         // is free on BOTH of IsStemFree's tests — no "houseCARL - <stem>" folder already exists, AND no active
         // plugin is named "<stem>.esp" — and suffixes it otherwise. Either trigger is ordinary: the folder arm is
@@ -8025,7 +8036,7 @@ public sealed class LoadOrderService : IDisposable
             $"cannot extend: no houseCARL plugin '{espName}' in any houseCARL folder, and no houseCARL folder named " +
             $"'{ModFolderName(stem)}'" +
             (string.Equals(bareName, ModFolderName(stem), StringComparison.OrdinalIgnoreCase) ? "" : $" or '{bareName}'") +
-            ". " + freshPatch switch
+            ". " + (laneClause is null ? "" : laneClause + " ") + freshPatch switch
             {
                 FreshPatchRemedy.NamedByPatchParam =>
                     $"Omit into= and pass patch=\"{stem}\" to create it fresh under a name you choose, or omit patch= "
@@ -8033,7 +8044,7 @@ public sealed class LoadOrderService : IDisposable
                     + "Or check the name.",
                 FreshPatchRemedy.CreatedByOmittingInto => "Omit into= to create it fresh, or check the name.",
                 _ => WriteSentences.ExtendCheckTheName,
-            } + (laneClause is null ? "" : " " + laneClause));
+            });
     }
 
     /// <summary>houseCARL-OWNED mod folders under ModsDir holding a plugin file named <paramref name="espFileName"/> at
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -5629,7 +5629,10 @@ public sealed class LoadOrderService : IDisposable
             // The lane clause is what #356 was: None used to render "Omit into= to create it fresh", and omitting the
             // lane is itself refused, so the sentence sent callers to a second refusal. What this lane can honestly
             // offer instead is measured — a patch created first carries no override of the record, so removal refuses
-            // there too; the in-place lane does remove it, behind the first-touch confirmation.
+            // there too; the in-place lane does remove it. Note what that lane costs, because the sentence does not
+            // say it: its consent is once per plugin, persisted, and shared with the edit and create in-place lanes,
+            // so a caller who follows this against an already-acknowledged plugin gets no prompt. #359 owns whether
+            // the remedy should say so.
             //
             // The in-place half comes from the TOOL (inPlaceRemedy), because the two tools spell that lane
             // differently and the service cannot tell which one called it. Null — a direct service call — offers
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -8021,10 +8021,12 @@ public sealed class LoadOrderService : IDisposable
         // (removal's is the in-place lane, which no other caller here has) hands the sentence in rather than having
         // it inferred from a semantic bit — the shape LocalizedTargetUnsupportedException.Text already uses. It rides
         // THIS arm only: the foreign-folder arm's remedy is #359's open design question, and the ambiguous and
-        // multi-plugin arms name a call rather than a lane. Note what that is NOT a claim that those two arms are
-        // correct — they instruct every caller to "pass into=", which housecarl_remove_record does not declare and no
-        // alias row maps onto its patch= (the table runs old→new only). That is this same defect one arm over, it is
-        // on main untouched by this change, and it is filed rather than fixed here.
+        // multi-plugin arms name a call rather than a lane. That is NOT a claim that those two arms are correct.
+        // They instruct every caller to "pass into=", which housecarl_remove_record does not declare, and the alias
+        // table has no row whose OLD spelling is "into" — it is bidirectional, so the miss is that specific row's
+        // absence rather than any one-way rule. Measured: that tool answers "unknown parameter: into". Same defect
+        // as this one, one arm over; it is on main untouched by this change, and filed as #377 rather than fixed
+        // here.
         //
         // It is rendered BEFORE the fallback above, not after: the lane's own diagnosis is what makes the fallback
         // the right thing left to do, and reading the instruction first was the order two reviewers stumbled on.
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -5591,12 +5591,14 @@ public sealed class LoadOrderService : IDisposable
             return WritePatchBuilder.RemovalOutcome.Fail(
                 "target= is only meaningful with in_place=true (it names the plugin to remove from in place). For the default lane omit target=; use patch= to name the houseCARL patch.");
         if (!inPlace && string.IsNullOrWhiteSpace(patch))
+            // Renders the caller's own spelling, exactly like the not-found arm below: null names no lane. This
+            // arm used to hardcode the 1.x sentence, justified by its being 1.x-only reachable — which was true,
+            // but "true for the one caller that reaches it today" is a roster with one row, and it made a fact
+            // about a DIFFERENT file (that housecarl_remove refuses before the service is called) load-bearing
+            // here. Handing it down costs nothing and removes both.
             return WritePatchBuilder.RemovalOutcome.Fail(
-                // The legacy spelling is hardcoded here because this arm is 1.x-ONLY: housecarl_remove refuses
-                // "no lane named" — in its own correct spelling — before the service is reached. Measured, and
-                // pinned by the extend-resolver guard's tool-altitude arm.
-                "patch is required — name the houseCARL patch to remove the record from (removal only targets a patch that already carries it). "
-                + WriteSentences.RemoveInPlaceLaneLegacy);
+                "patch is required — name the houseCARL patch to remove the record from (removal only targets a patch that already carries it)."
+                + (inPlaceRemedy is null ? "" : " " + inPlaceRemedy));
 
         // Parse every formid first, collecting ALL problems (all-or-nothing, like the edit path). Pure — outside the gate.
         var keys = new List<FormKey>(formids.Count);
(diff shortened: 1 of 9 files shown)


src/housecarl-mcp/WriteSentences.cs
```diff
@@ -596,6 +596,53 @@ internal static string PlaceSourceNamedAbsent(string provider, string rel, IRead
...
+    ///
+    /// <para>There are two spellings because there are two tools, and the SERVICE cannot tell which one called it.
+    /// So each tool hands its own down (<c>housecarl_remove</c> this one, <c>housecarl_remove_record</c>
+    /// <see cref="RemoveInPlaceLaneLegacy"/>) rather than the resolver picking — the same caller-states rule the
+    /// fresh-patch remedy follows, one altitude up. Sharing ONE sentence is what #356's own fix got wrong first:
+    /// the 1.x pair was correct where it already rendered, and reusing it on the 2.0 tool named a parameter that
+    /// tool does not declare. It worked only because the 1.x→2.0 lane shim silently re-spelled it, and that shim is
+    /// scaffolding due to retire.</para></summary>
+    [MustState("pass in_place=\"<plugin filename>\"")]
+    internal const string RemoveInPlaceLane =
```
Avick3110, line 639: **[high] Following this remedy can rewrite an original with no confirmation at all.**

The PR's §5 #11 escalation rests on: *"The in-place lane cannot fire without an explicit confirmation naming the file, so the failure mode is a wasted round trip, not a write."* That holds only for a plugin never acknowledged before.

`RemoveRecordsInPlace` (LoadOrderService.cs) gates on `bool already = _store.IsInPlaceAcknowledged(targetPath); if (!already && !acknowledge) return NeedsAck(...)`. `UserConfig.IsInPlaceAcknowledged` is a **persisted, cross-session, per-normalized-path** check, and its own doc comment says the consent is **shared with the edit + create lanes** — "acknowledging a plugin once covers editing, creating into, AND removing from it in place."

Concrete scenario:

1. Modder does one in-place *edit* on `CoolWeapons.esp` last week and acknowledges it. Path is now in `InPlaceAcknowledged`.
2. Today: `housecarl_remove(formids=["0012AB:CoolWeapons.esp"], into="MyPatch.esp")` — the patch was auto-suffixed to `MyPatch_001.esp`, so this misses.
3. New refusal: `To remove from an existing plugin IN PLACE instead, pass in_place="<plugin filename>".`
4. The only plugin filename the caller holds is the one inside the FormID it was handed. It passes `in_place="CoolWeapons.esp"`.
5. `already == true` → the `NeedsAck` return is skipped → the record is dropped out of the modder's original file. No prompt, no backup, no undo.

The measurement in the PR body ("Following it reaches the first-touch confirmation; acknowledging it removes the record") was taken on a *fresh* plugin, which is the one state where the confirmation is guaranteed to fire — the same altitude-of-measurement shape round 1 caught on the spelling.

This is distinct from the filed #378 (consent *spent* on a refused not-carried call): there nothing is written; here the consent is already spent and the write lands. It doesn't necessarily block the PR, but the escalation's safety argument needs restating before #359 is decided on it.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5591,7 +5592,11 @@ public WritePatchBuilder.RemovalOutcome RemoveRecords(IReadOnlyList<string> form
                 "target= is only meaningful with in_place=true (it names the plugin to remove from in place). For the default lane omit target=; use patch= to name the houseCARL patch.");
         if (!inPlace && string.IsNullOrWhiteSpace(patch))
             return WritePatchBuilder.RemovalOutcome.Fail(
-                "patch is required — name the houseCARL patch to remove the record from (removal only targets a patch that already carries it). To remove from an existing plugin IN PLACE instead, pass target=<plugin filename> + in_place=true.");
+                // The legacy spelling is hardcoded here because this arm is 1.x-ONLY: housecarl_remove refuses
+                // "no lane named" — in its own correct spelling — before the service is reached. Measured, and
+                // pinned by the extend-resolver guard's tool-altitude arm.
+                "patch is required — name the houseCARL patch to remove the record from (removal only targets a patch that already carries it). "
+                + WriteSentences.RemoveInPlaceLaneLegacy);
```
Avick3110: **[medium] This is the one place left where the service picks a tool's spelling for it — and `inPlaceRemedy` is already in scope.**

I checked the 1.x-only claim and it holds *today*. `housecarl_remove` runs `if (!hasInto && !hasInPlace) return Refuse("no lane named...")` (RemoveTools.cs:82) before the service is reached, and `patch` is passed as `hasInto ? into : null`, so `!inPlace && IsNullOrWhiteSpace(patch)` is unreachable from the 2.0 tool. Both use the same `IsNullOrWhiteSpace` test on the same string, so the alias/binding shim can't split them. The probe's `noLane` arm pins it.

The problem isn't that the claim is false — it's the shape. This arm's correctness is a **cross-file guard-ordering invariant**, and hardcoding the constant is what makes the service depend on it. That is the same disease the rest of the branch cures: a one-row roster of which tool can reach this line, defended by a comment. The comment three arms up says it out loud — "a list of them is what keeps going stale; the call sites are the authority" — and then this arm is the exception to it.

`inPlaceRemedy` is a parameter of this very method and already carries the right string for every caller:

- `housecarl_remove_record` → `RemoveInPlaceLaneLegacy` (identical output to today)
- `housecarl_remove` → `RemoveInPlaceLane` (correct if it ever *does* reach here)
- direct service call → `null` → same posture as the not-found arm, which deliberately names no spelling

```suggestion
                "patch is required — name the houseCARL patch to remove the record from (removal only targets a patch that already carries it)."
                + (inPlaceRemedy is null ? "" : " " + inPlaceRemedy));
```

That deletes the invariant instead of pinning it, and the probe arm keeps its meaning (it still asserts the 2.0 tool refuses first). The load-bearing fact stops being load-bearing.

If you keep the hardcode, the failure mode is concrete rather than hypothetical: when the 1.x→2.0 lane shim retires — which the PR body says is scheduled — anything that reorders or absorbs that guard hands a `housecarl_remove` caller `target=<plugin filename> + in_place=true`, a parameter that tool does not declare. #356's own defect, in the arm next to its fix.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8001,11 +8041,15 @@ string ResolveOwnedPatchFolder(string into, bool needEsp, bool patchNamesFresh =
             $"cannot extend: no houseCARL plugin '{espName}' in any houseCARL folder, and no houseCARL folder named " +
             $"'{ModFolderName(stem)}'" +
             (string.Equals(bareName, ModFolderName(stem), StringComparison.OrdinalIgnoreCase) ? "" : $" or '{bareName}'") +
-            ". " + (patchNamesFresh
-                ? $"Omit into= and pass patch=\"{stem}\" to create it fresh under a name you choose, or omit patch= "
-                  + "too and houseCARL names it \"Patch\" — either name auto-suffixed if already taken. "
-                  + "Or check the name."
-                : "Omit into= to create it fresh, or check the name."));
+            ". " + (laneClause is null ? "" : laneClause + " ") + freshPatch switch
```
Avick3110, line 8047: **[medium] Third-reader input on the escalated clause ordering — not a third fold.**

Read cold, here is the whole sentence for `housecarl_remove(..., into="My Patch.esp")`:

> cannot extend: no houseCARL plugin 'My Patch.esp' in any houseCARL folder, and no houseCARL folder named 'houseCARL - My Patch'. **houseCARL will not create a patch here: a removal only drops a record the patch ITSELF already carries. To remove from an existing plugin IN PLACE instead, pass in_place="&lt;plugin filename&gt;". Check the name.**

Two things a fresh reader hits that neither round named:

**1. The no-create clause rebuts a sentence the caller never saw.** I passed `into=`. I did not ask houseCARL to create anything. "houseCARL will not create a patch here" only reads as an answer if you remember the tail it replaced. To a caller arriving fresh it's a denial of an offer that was never made, sitting in the most prominent position in the response.

**2. Order isn't the variable — information is.** Round 1 wanted the lane clause first; round 2 wanted it later. Both are right about the symptom because the fallback has nothing in it. `"Check the name."` names no candidate, quotes nothing back, and (the PR measured this) points at no tool that could list houseCARL-owned patches. So whichever end it goes, the only *actionable* thing in the response is the consent-gated lane that rewrites an original — and per my other comment, that lane doesn't reliably prompt. The likeliest cause by a wide margin is a typo or an auto-suffix, and it gets two contentless words three clauses from the value they refer to.

That reframes the #359 fork slightly: it isn't only "may a refusal recommend a consent-gated destructive lane", it's "may it recommend one *while the always-true remedy is empty*." If `"Check the name."` could carry even the owned-patch names near the stem, the ordering question mostly dissolves — the prominent clause would be the safe one on merit rather than by placement. That's a capability gap (nothing lists them), not something to fold here.

Agreed with the stop rule: don't fold a third time. Recording this as the reader observation, for whatever it's worth to the #359 call.

Avick3110:
## Fold of Aaron's review — 3 findings, 3 folded, **0 refused**

The zero is not a triage failure: each finding was reproduced against source before folding, and each reproduced. F2's premise (the arm hardcodes a spelling, justified by a cross-file reachability fact) and F1's (consent is persisted and shared, so the "confirmation always intervenes" pricing is false) were both confirmed by reading the code paths named. F3 was ruled recorded-not-folded by the reviewer, and is recorded. Nothing here was folded on the reviewer's authority alone.

Two commits: `3dfbc32` (F2), `2a7fa81` (F1 + F3's pointer). CI green, `ci-all` 124/124.

### F2 — the missing-patch arm renders the caller's spelling, like every other arm

The `patch is required` refusal was the last place the service chose a tool's sentence. It now renders `inPlaceRemedy` on exactly the not-found arm's terms: null names no lane. That deletes the roster-at-N=1 and, with it, the cross-file invariant that made *another* file's guard ordering load-bearing here.

**Acceptance, measured before the commit.** Five renders captured on `0f676df`, re-captured after the change, diffed:

- **The 1.x path is byte-identical.** `housecarl_remove_record` hands down `RemoveInPlaceLaneLegacy` and gets the same tail it got before, to the character. No 1.x output changed.
- **One delta, and it is the intended one**: the direct *service* call with no remedy handed in now names no spelling, which is the posture the not-found arm already had.

Arms followed the sentence. The constant pin moved to the tool that supplies it — 1.x reaches this arm and gets the 1.x sentence; a direct service call gets none. The `noLane` arm keeps its assertion but not its job: it is now a statement about how `housecarl_remove` behaves, no longer the fact that makes the service correct. The write-surface guard's reach check for the legacy spelling also moved to the 1.x tool, since the service no longer renders it — caught immediately when the check went RED with `unobserved: RemoveInPlaceLaneLegacy`.

Same commit, the same reflex one paragraph up: the doctrine comment declared "not a roster" and then enumerated tools in prose. It states the three failure properties now and points at the call sites as the authority.

### F1 — the false pricing, corrected at every home

The escalation section claimed following the remedy costs at worst a wasted round trip. That holds only for a plugin never acknowledged; consent is persisted, cross-session, and shared across the edit / create / remove in-place lanes, so an already-acknowledged plugin is rewritten with no prompt.

The refusal sentence is **unchanged** — whether the recommendation must carry a consent caveat is #359's fork, where the finding is deposited.

Swept for other homes rather than fixing only the one named, and it had four:

| Home | Was |
|---|---|
| PR body, escalation section | "a wasted round trip, not a write" |
| `WriteSentences.RemoveInPlaceLane` doc | "it is what a caller who follows this one gets next" |
| `LoadOrderService` call-site comment | "the in-place lane does remove it, behind the first-touch confirmation" |
| `ExtendResolveProbe` arm comment | generalized from a fixture whose plugin has never been acknowledged |
| `plugin/CHANGELOG.md` | "its own first-touch confirmation still applies" |

The changelog now states plainly that if you have already acknowledged that plugin, in any lane or any earlier session, the write happens with no further prompt.

### F3 — recorded on #359, not folded

Per the reviewer's ruling. The escalation section (rewritten for F1) carries the pointer: clause **order** was never the variable — the always-true fallback is nearly empty of content, so it reads as filler wherever it sits while the lane clause does the work. What the shared fallback should *be* is the same refusal-remedy fork, and is recorded on #359 with F1.

### Sabotage

Two more RED-checks for F2's arms, both from the committed state, both restored and anchor-verified: re-hardcoding the legacy spelling in the missing-patch arm (1 FAIL), and the 1.x tool ceasing to hand its spelling down (2 FAILs in the extend-resolver guard plus the write-surface reach check). Sixteen across the branch.

No new review rounds were run, per §5 #11. Merge remains yours.

PR #384, opened 2026-08-20
docs(spid-authoring): add "Verifying a distribution actually happened"

### LINKED ISSUE #383: spid-authoring: nothing on verifying a distribution actually landed
The `spid-authoring` skill covers authoring and interpreting `_DISTR.ini` files, and its "silent failure" paragraph correctly warns that a bad line distributes nothing without an error — but the skill says nothing about how to positively verify that a rule DID land on an actor. Verifying is where three real SPID behaviors bite, all measured in-game (SPID 7.3.2, SSE 1.6.1170, headless scripted run):

1. **SPID never distributes to the player.** A verification that asserts on `Game.GetPlayer()` reports failure for a perfectly working distribution. This cost us a debugging round before we sampled NPCs instead.
2. **Distribution happens at NPC load.** An actor already baked into a pre-existing save proves nothing about a rule added after that save — the assertion must target actors loaded after the rule exists.
3. **A runtime-created keyword has no plugin FormID but IS reachable** via `Keyword.GetKeyword("<name>")` (SKSE) — which is what makes a no-ESP keyword distribution verifiable from script at all.

Proposal: a short "Verifying a distribution actually happened" section in the skill, carrying exactly those facts plus one sampling note (an interior cell with known occupants beats an exterior spawn marker, where an empty scan is inconclusive). Doc-only, no tool changes. PR to follow.

juggernaunt46-star:
Fixes #383

The `spid-authoring` skill covers authoring and interpreting `_DISTR.ini` files but had nothing on positively verifying that a rule landed on an actor — and verification is where three real SPID behaviors bite. This adds one section between the fabrication warning and "Common mistakes", carrying the measured facts:

- **SPID never distributes to the player** — asserting on `Game.GetPlayer()` reports failure for a working distribution; sample NPCs instead.
- **Distribution happens at NPC load** — an actor baked into a pre-existing save proves nothing about a rule added afterwards.
- **A runtime-created keyword has no plugin FormID but is reachable** via `Keyword.GetKeyword("<name>")` (SKSE), which is what makes a no-ESP keyword distribution verifiable from script at all.
- One sampling note: an interior cell with known occupants beats an exterior spawn marker, where an empty scan is inconclusive rather than a failure.

All three were measured in-game on SPID 7.3.2 / Skyrim SE 1.6.1170 via a scripted headless run (keyword distribution to interior NPCs, asserted through a Papyrus global calling `Keyword.GetKeyword` + `HasKeyword` on sampled actors; the player-exclusion finding came from the same rig first asserting on the player and failing against a distribution the NPC sample then proved working).

Doc-only: `SKILL.md` plus a `## Unreleased` changelog line, per CONTRIBUTING. No tool behavior changes, so no new probe/guard; the eval set is untouched since the trigger surface (description/frontmatter) is unchanged.

https://claude.ai/code/session_014t8MT7C46f2x7KpUTgVCKS

### DIFF
--- a/.claude/skills/spid-authoring/SKILL.md
+++ b/.claude/skills/spid-authoring/SKILL.md
@@ -99,6 +99,21 @@ debugging session. If the user is on a SPID version newer than 7.3.0 and asks ab
 the reference, surface that the reference may be behind and offer to re-derive from the current
 SPID documentation.
 
+## Verifying a distribution actually happened
+
+A parsed `_DISTR.ini` and a clean SPID log are **not** proof that an NPC got the thing. The log
+shows what SPID looked up; only a live actor shows what SPID applied.
+
+- **SPID does not distribute to the player.** Asserting on `Game.GetPlayer()` returns false even
+  for a working distribution — verify against sampled NPCs instead.
+- **Distribution happens at NPC load**, so an actor already baked into a pre-existing save proves
+  nothing about a rule added afterwards — verify against actors that load after the rule exists.
+- **A dynamic keyword IS reachable.** A keyword SPID creates at runtime has no plugin FormID, but
+  `Keyword.GetKeyword("<name>")` (SKSE) finds it — which is what makes a no-ESP keyword
+  distribution verifiable from script at all.
+- **Sample somewhere NPCs must be.** An interior cell with known occupants beats an exterior spawn
+  marker: finding no NPC within scan range there is inconclusive, not a failing distribution.
+
 ## Common mistakes
 
 - **Miscounting pipe positions.** The sections are positional; a chance written one pipe early lands
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,6 +13,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **`spid-authoring` now covers verifying a distribution on a live actor.** A parsed `_DISTR.ini` and a
+  clean SPID log show what SPID looked up, not what an actor received. The skill's new "Verifying a
+  distribution actually happened" section says what to assert instead — and why the player is the wrong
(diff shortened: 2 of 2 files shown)


Avick3110:
Thanks — this is a genuinely useful section, and two of your three facts verified cleanly against the official docs (load-timing is verbatim in SPID: The Complete Reference; dynamic keyword creation likewise, and `Keyword.GetKeyword` is in our bundled SKSE reference). The player-exclusion fact appears in no official documentation, but it's confirmed in SPID's own source (`DistributeManager.cpp`, `should_process_NPC` returns `!a_npc->IsPlayer()`), which we'll cite.

One correction needed before merge: "an actor baked into a pre-existing save" teaches a persistence model the official reference explicitly denies ("SPID... doesn't have effect on either loaded plugins nor on save. SPID distributes everything from scratch each time you launch the game"). The true fact is narrower: distribution runs when an NPC loads, so an actor already loaded before your rule existed isn't reprocessed until it reloads — and nothing is ever baked into the save. Please reword that bullet to that model (or we can amend on merge).

Avick3110:
Thanks for this. It is a real gap in the skill and three of the four facts check out against SPID's own source: the player really is excluded (should_process_NPC in DistributeManager.cpp is !IsPlayer() && !IsDeleted()), and a keyword SPID creates at runtime is pushed into the game's keyword array, which is why Keyword.GetKeyword finds it.

One bullet is not right: SPID redistributes from scratch every launch, so an actor in an old save does get a newly added rule. What bites is that the INIs are read once at game launch, so a rule added mid-session needs a restart. The section also needs to sit below Common mistakes per our skill standard, and the two source facts need a grounded entry in the skill's references folder.

We are landing those changes ourselves in a follow-up PR that credits this contribution, and closing this one without merging so the branches do not diverge. #383 closes with that PR.

Issue #382, opened 2026-08-19
[Bug] in-place create grows the master header without the re-sort note its sibling lanes emit


Avick3110:
## Summary

An in-place **create** that adds a cross-plugin reference grows the target plugin's master header — and returns no note. The in-place **edit** and **forward** lanes emit an explicit re-sort note for exactly that situation. A caller who creates a record in place, and whose new record links into a plugin the target did not already master, is not told the load order may now need re-sorting.

## Measured

On `main` @ `dca22c4`. `InPlaceProbe` arm O already drives the scenario (a new record in place carrying a cross-master reference; the master header grows). Instrumenting that arm to print the outcome's `Note`:

```
PROBE-O-NOTE>>> [(null)]
   PASS  O cross-master reference adds the master (xEdit-parity serialize)
   PASS  Y edit-lane master GROW (link to an active undeclared plugin lands + grows the header + re-sort note)
```

Arm Y is the edit lane's twin of the same fact and asserts `resortNoted=True`.

## Why it happens

`WritePatchBuilder.MasterGrowNote` is called from exactly two places:

- `ApplyInPlaceCore` — `WritePatchBuilder.cs:1075`
- `ForwardRecordsInPlaceCore` — `WritePatchBuilder.cs:1870`

The create path returns from a single success construction (`WritePatchBuilder.cs:3240`) that never computes `mastersBefore` and never sets `Note`. `CreateOutcome.Note` therefore has no producer in the core at all — the only thing that ever sets it is the service's in-place create block, for the acknowledgement and marker notes.

There is no comment at any of the three sites recording the omission as deliberate.

## Scope note

Pre-existing and unrelated to the branch it was found from: `MasterGrowNote` has the same two call sites on `dca22c4`. Filed rather than folded per CLAUDE.md §2 — a confirmed defect outside the branch under review.

The same measurement establishes a *non*-defect worth recording, so it is not re-investigated: the in-place create block's `JoinNotes(ackNote, markerNote)` looks like it drops `enriched.Note`, but `CreateOutcome.Note` is provably null at that point for the reason above, so nothing is lost. If this issue is fixed by giving the create lane a master-grow note, that `JoinNotes` call becomes a real drop and must take `outcome.Note` first, the way the other three lanes already do.

*Unrouted — needs a RUN_ORDER routing stroke.*

Avick3110:
[...]

PR #385, opened 2026-08-20
fix(in-place): a refused call no longer spends the one-time consent

### LINKED ISSUE #378: [Bug] in-place removal records the one-time consent before the not-carried check, so a refused call spends it

Avick3110:
Fixes #378

## What was wrong

Every in-place lane recorded the one-time acknowledgement at the consent gate, **before** the write was attempted. A call that was then refused — and wrote nothing — had already spent the plugin's first-touch confirmation, so the next in-place call proceeded with no prompt and rewrote the user's original.

The lane stated the opposite standard itself, one comment up: the localized-target refusal sits ahead of the gate precisely so it cannot "spend the modder's one-time in-place acknowledgement on a write that was never going to happen."

Measured on the branch base, all four plugin lanes:

```
refused=True consentRecorded=True(want False) nextCallPrompts=False fileUntouched=False
```

`fileUntouched=False` is the load-bearing half — the second, unacknowledged call really did rewrite the file.

## Fix shape chosen: relocate the persist, not the checks

Two shapes satisfied the ruled principle. **Shape (b) was chosen on the audit below**, not on preference.

The consent **check** does not move: `already || acknowledge` still gates entry, in the same place. Only the **record** moves, to after the lane reports success.

Shape (a) — hoisting each never-going-to-happen check ahead of the persist — was rejected because the audit shows the checks are not a short list. Hoisting would mean duplicating the plugin open, the parse, and each lane's content gate ahead of the gate in four lanes, giving a second reading of a fact the write already owns, and it would leave the class alive for any check added later. Shape (b) kills the class structurally: nothing can sit behind a persist that is the last thing the call does.

Its failure direction is the honest one. A fault between the file swap and the persist re-prompts next time; over-prompting costs a confirmation, under-prompting costs a file.

**Skeleton tripwire did not fire.** The relocation is **one helper with five call sites**, not five copies of a reorder:

```
helper definitions:            1   (LoadOrderService.PersistInPlaceConsent)
call sites:                    5   (edit / create / remove / forward / nif)
copies of the failure sentence: 1
```

No lane hand-copies the reorder, so the `X`/`XInPlace` twin-drift condition in the row-3 tripwire is not met. The five persist-failure notes were consolidated into that same helper in a later commit; the two words that genuinely vary per lane (the verb, and `file` vs `plugin` for nif) are parameters.

## Sibling audit — every check between the persist and the file changing

This is the ruled deliverable, per lane. In the four plugin lanes the mutation point is `CommitStagedPatch`, the **last** statement of `WriteEngine.WriteInPlace`; everything before it leaves the original byte-intact.

| # | Check | Where | Never-going-to-happen? |
|---|---|---|---|
| 1 | writable-parent pre-flight (`InPlaceParentUnwritable`) | service, step (3) — **all five lanes** | **Yes** — refuses, nothing staged |
| 2 | target not active in the captured view | builder Phase 1 | Yes |
| 3 | target EXCLUDED from session (unparseable) | builder Phase 1 | Yes |
| 4 | file missing on disk | builder Phase 2 | Yes |
| 5 | `CreateFromBinary` parse failure | builder Phase 2 | Yes |
| 6 | ModKey / filename mismatch | builder Phase 2 | Yes |
| 7 | **lane content gate** — remove: "not carried by"; edit: "does not define or override"; forward: source capture / child-group; create: per-spec pre-flight | builder Phase 3 (Phase 1 for create) | **Yes — the reported one** |
| 8 | missing masters (`ResolveOwnMasters`) | builder Phase 5 | Yes |
| 9 | localized backstop (`WriteInPlace` throw) | engine, pre-staging | Yes |
| 10 | serialize / null-arm failure | engine, staged temp discarded | Yes |
| — | **`CommitStagedPatch` — the file changes here** | | |
| 11 | re-open + touched-record / absence verify | builder Phase 5–6 | **No** — already mutated |

The nif lane has the same shape around `AtomicFile.WriteAllBytes`: writable-parent pre-flight and the overwrite exception ahead of it (both never-going-to-happen), the on-disk size check behind it (not).

**Answering the question the kickoff asked specifically:** the writable-parent pre-flight sits behind the old persist in *every* lane, and it is in the never-going-to-happen class in every lane — it refuses without staging anything.

So rows 1–10 were all spending consent on writes that never happened. That is why the check-hoisting shape was rejected: it is ten hoists per lane, not one.

`housecarl_compact_plugin` is **not** a sixth lane. Its `acknowledge` is a per-call `Confirm` with its own overwrite list; it never touches the persistent store, so it has no persist to reorder.

## Coverage: the 5×2 grid

The persist lands in five lanes, each with a success direction (persist happens) and a refusal direction (it doesn't). Each cell was sabotaged **individually** and the named arm confirmed RED. No cell is structurally unreachable, so none is declared a gap.

| lane | success — persist happens | refusal — persist doesn't |
|---|---|---|
| edit | arm **G** | **CO-B** |
| create | arm **K** | **CO-D** |
| remove | **CO-F** *(added — cell was empty)* | **CO-A** |
| forward | **CO-E** *(added — cell was empty)* | **CO-C** |
| nif | `nif-set-guard` "a LATER in-place edit needs NO re-acknowledge" | `nif-set-guard` refusal arm *(added — both cells were empty)* |

Two things the grid caught that reading "which arms went red" did not:

- **remove's success cell was empty.** Arm W looks like it covers it and does not: its first half never re-calls after the acknowledged removal, and its second half has an *edit* record the acknowledgement and only checks a later remove rides it. Deleting remove's persist left `ci-all` at 124/124.
- **forward's success cell and both nif cells were empty.** Reinstating the pre-branch ordering in `nif_set` is #378 live, and the whole suite stayed green.

Every CO arm and the nif arm also assert `gateBeforeRefusal` — the same call with `acknowledge=false` must return the handshake. Without it, pinning the refusal's *sentence* catches a different refusal being substituted but **not** the same refusal hoisted ahead of the gate, and a refusal the gate never reaches cannot spend consent, so the arm would stay green while testing nothing. Measured on the nif lane, where that hoist previously left `nif-set-guard` ALL PASS.

The nif fixture needs one note: it holds the target open with `FileShare.Read`, not `FileShare.None`. With `None` the *read* fails ahead of the consent gate and the arm passes without exercising the seam at all.

`ci-all`: **124/124**.

## Review rounds

**Two rounds, five agents, 14 findings, 14 folded, 0 refused.** Stopped at two.

**Round 1** — 3 agents (correctness, prose/contract, guard quality). 9 findings, 9 folded.

The two independent reviewers converged on the same hole: the persist conditional lands in five lanes, and my RED checks had covered four on each side — but not the same four. Also folded: CO-C/CO-D asserted a bare `!Success`; the changelog wrongly folded `nif_set` into "covers the whole plugin… shared across those lanes"; five persist-failure notes and the store's doc said "a future session will re-prompt" when nothing caches the config; a twin of an already-corrected sentence sat 2000 lines down the same file; and 11 `acknowledge` descriptions promised "never again for it", which this branch is what made false.

**On round 1's 0% refusal rate**, in my own words and left standing deliberately: seven of nine were text claims resolvable by reading one referenced line, and the two behavioural ones I reproduced myself before folding. But a round where I refused nothing is a round where I should suspect the prompts were too agreeable, and the honest reading is that round 1 found real defects because there were real defects to find — **my guard coverage was genuinely thinner than I'd claimed in my own commit message.**

**Round 2** — 2 agents (correctness; guards + contracts). 5 findings, 5 folded.

Round 2's correctness pass found **zero defects in the fix itself** and independently reproduced the per-lane sabotage matrix. Its guard findings were the arms *round 1 added*: the nif arm accepting any failure, and the sentence-pinning overclaim. Sentence findings: a changelog cross-reference pointing "above" at an entry this branch had just moved below it; the localized parenthetical claiming the LOC arms redden on the missing clause, true on the two lanes with no dry run and false on the two with one; the arm roster omitting CO-E; and six `WriteTools` lane lists enumerating two or three of the four lanes the acknowledgement is shared across.

**Why two rounds and not three.** Round 2 found zero defects in the fix; its findings were confined to arms round 1 added — the fold-artifact class the bounded rounds exist to contain. A third round would be reviewing round 2's own folds: the loop, not more coverage. The 5×2 grid, with a named and individually-sabotaged arm per cell, is the coverage evidence a third round would otherwise be hunting for. This is wrong if the independent review finds a fix-level defect a round would have caught — which is exactly what that gate is for, so the stop is priced honestly rather than assumed safe.

## Also in this PR

- `PersistInPlaceConsent` cannot throw. The store returns its write failures, but its cross-process lock handling sits outside that `try` — and now that the call runs *after* the file changed, a throw would report failure for a write that landed.
- Sentences that describe consent timing were swept and corrected wherever they had moved with the behaviour, including the four lane comments, the engine's pre-flight doc, two probe comment blocks, and a changelog paragraph. The four localized pre-flights are **kept**: they carry the lane's own remedy clause, which `WriteInPlace`'s throw cannot name because it does not know which lane called it. Their comments now say that, because a reader who believed the old reason is a reader who deletes them.
- `## Unreleased` changelog entry, written against the authoring rule at the top of the file: states what the tool now does and refuses, points at bounds rather than restating them, no frequency or coverage adjectives.

## Take-or-drop commit

`de7d483` adds one bullet to CLAUDE.md §5 #11: a scripted sabotage sweep carries a known-RED canary, and a sweep whose canary comes back green is a broken sweep. Promoted from this engagement's two false-green harness bugs — both reported a fully-covered grid that had never been measured. It touches only the operating manual, rides its own commit, and is Aaron's to take or drop at merge. Amended after review: one sentence added at advisor direction, closing a seam the build session flagged — a sweep with no known-red cell makes one first rather than running canary-less.

## Filed separately

**#382** — in-place **create** grows the master header without the re-sort note its edit and forward siblings emit. Measured by instrumenting arm O (`Note` came back null) against arm Y's `resortNoted=True` for the identical situation. `MasterGrowNote` has exactly two call sites. Pre-existing — same on `dca22c4` — so filed rather than folded, per CLAUDE.md §2. Unrouted; routing is Aaron's stroke.

The same measurement settled a non-defect worth not re-investigating: the create lane's `JoinNotes(ackNote, markerNote)` looks like it drops `enriched.Note`, but `CreateOutcome.Note` has one construction site that never sets it, so nothing is lost. Recorded on #382 because fixing that issue would make it a real drop.


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5430,6 +5431,45 @@ static string InPlaceHandshakeText(string pluginName, string path) =>
         "  • The default lane (a NEW patch, originals untouched) stays the recommended way — this is the explicit opt-in.\n" +
```
Avick3110, line 5451: **[medium] The prompt itself still promises "shown once for this plugin" — the one sentence the sweep missed.**

`InPlaceHandshakeText` (line 5427, and its docstring at 5422: *"server-enforced, shown once per plugin"*) opens with:

> in-place edit of 'X.esp' — first-time confirmation (shown once for this plugin):

This branch is what makes that false. The 11 `acknowledge=` Descriptions were correctly moved off `never again for it` → `not again once one has LANDED`; the prompt a modder actually reads still carries the old claim.

Concrete, and already asserted green by this branch's own arm — CO-A's `nextCallPrompts=True` **is** the second showing:

1. `housecarl_remove(in_place="X.esp", acknowledge=true)` on a record `X.esp` does not carry → refused, nothing persisted.
2. Next in-place call on `X.esp` → `InPlaceHandshakeText` again.

The row-11 window makes it worse than redundant. When the write commits and the post-write verify then fails, the re-shown prompt says *"This writes to your ORIGINAL file (…). It will NO LONGER be untouched"* about a file that has already been rewritten — future tense over a completed mutation, at exactly the moment the modder most needs to know their original is gone. That is the honest cost of the row-11 direction, and it lands in the prompt text rather than in the ordering.

Per the rounds rule, this wants a probe rather than a rewording: an arm that pins what the second prompt says after a refused acknowledged call.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1415,15 +1415,13 @@ public NifSetResult NifSet(string relPath, IReadOnlyList<NifSetOp> ops, string?
                 var targetPath = chosen.LooseFilePath!;
                 var meshName = Path.GetFileName(targetPath);
 
+                // The CHECK gates entry here; the acknowledgement is RECORDED below, once the overwrite has landed and
+                // verified — see PersistInPlaceConsent. The parent pre-flight and the write's own failure both refuse
+                // without changing the file, and neither may spend the caller's one-time confirmation.
                 bool already = _store.IsInPlaceAcknowledged(targetPath);
                 if (!already && !acknowledge)
                     return NifSetResult.NeedsAck(NifInPlaceHandshakeText(meshName, targetPath), chosenProv, providers, profileName);
```
Avick3110, line 1423: **[medium] Same missed sentence in the nif lane — and this branch's own new arm is the proof.**

`NifInPlaceHandshakeText` (line 1480), the text this `NeedsAck` returns, still opens:

> in-place edit of 'x.nif' — first-time confirmation (shown once for this mesh):

`NifTools`' `acknowledge=` Description was corrected on this branch to *"not again once one has LANDED — a call that is refused records nothing, so it may be needed again"*. The prompt was not.

The refusal arm added to `NifSetGuardProbe` in this PR asserts the contradiction directly:

```
…so the NEXT call still meets the first-touch prompt rather than overwriting unprompted
```

A held/locked target makes the swap refuse, nothing is persisted, and the prompt that said it would be shown once is shown a second time. Verified green locally on `44fb4a2` (`inplace-guard: PASS`, `nif-set-guard` arm included).

Same remedy as the plugin twin: probe what the second prompt says, then fix the sentence.

Avick3110:
## Fold summary — Aaron's review

**3 findings triaged: 2 confirmed and folded, 1 refused by measurement.**

### F1 + F2 — folded as one rewrite (`1f71a61`)

Both were the same defect on two surfaces, so they took one rewrite under two constraints.

**The "shown once" claim.** Both prompts said `shown once for this plugin` / `for this mesh`. A refused call records nothing, so a caller can legitimately meet the prompt again — the prompt was asserting that could not happen. Both now read *shown until an in-place write to this {subject} LANDS; a call that is refused records nothing, so you may see this again*, which is the semantics the `acknowledge=` parameters already carried.

**The pristine-state claim.** The same lead read `It will NO LONGER be untouched` — a state *transition*, and the transition may already have happened: a write that lands and then fails its verify mutates the file while recording no consent, so the very next call re-prompts against an already-modified file. It now states the durable fact instead — *writes to your ORIGINAL file, not a copy; houseCARL keeps NO backup or undo and cannot restore what it overwrites*. True in both worlds, **no new state and no conditional**. Recorded at the call site: if a future edit cannot be phrased truthfully both ways, that is a §4 surface-and-stop, not a licence to track whether the file was previously touched.

**One home or two:** the two corrected claims are one claim each across both prompts, so they share `InPlaceHandshakeLead(name, path, subject, verb)`. Bullets 2–3 genuinely differ (whole-plugin re-layout vs NiflySharp whole-file re-serialization; different parse refusals) and stay per-surface. One home where the claim is one, two where it isn't.

**Arms assert the corrected clause as rendered, by presence** — an absence check passes for a prompt that says nothing at all. `CO-G` for the plugin prompt, two checks in `nif-set-guard` for the mesh. Writing them caught a real bug in my own arm: the mesh result carries its prompt in `AckPrompt`, not `Error`, so the first version read a null and **failed loudly** rather than passing vacuously.

RED-checked with a known-red canary in the sweep, per the rule this PR promotes: reverting both claims to their old wording turned `CO-G` and both mesh checks red, with `CO-A` (a cell already proven red by hand) red alongside them to prove the sweep actually ran.

### F3 — refused, by measurement

The note that the PR body does not mention `bbb38d0` is **not reproducible against the body as it stands**. The `## Take-or-drop commit` section names the commit explicitly and states it is take-or-drop at merge. It was added before the review began; the review appears to have read a cached body. Measured by fetching the live body and matching on the section heading and the SHA — both present.

That refusal is the reason the round's rate is 2/3 rather than 3/3, and it is the only finding I could not reproduce in the artifact it describes.

### Fourth vocabulary pass

Three passes each found homes the previous one missed (11 `[Description]`s, then six lane lists, then the two prompt builders), so this pass ran repo-wide **by vocabulary, not by surface** — `shown once`, `never again`, `one-time`, `first-time confirmation`, `once per plugin`, `only on the FIRST`, `remembered across`, plus siblings met en route.

Every hit dispositioned:

- **14 corrected** — an overclaiming `one-time` modifier in tool *body* descriptions (Apply / Create / Remove / Forward / Nif / six in WriteTools), both READMEs, the `facegen-diagnostics` skill, and the two dry-run pending notes. The modifier is **deleted rather than restated**: the precise rule now has one home in the prompt and the `acknowledge=` parameters, and a second copy is what drifts.
- **Already-correct, left alone** — the 11 `acknowledge=` parameter descriptions (they say "the one-time in-place trade-off" and then state the landed rule two clauses later, so the sentence self-corrects); `PersistInPlaceConsent`'s doc (the acknowledgement genuinely is recorded once); the remove lane's "once per plugin" comment (store granularity, not prompt recurrence); `ExtendResolveProbe`'s `first-time confirmation` assertion (still present, deliberately kept so the phrase stays pinned).
- **Not amended — shipped history** — three `CHANGELOG.md` entries under released version headings, including 1.x's "then never asks again for that plugin". Released entries are the record of what shipped; only `## Unreleased` text is amendable.
- **Not a consent claim** — `one-time` in the SKSE skill references, four probe headers' "not a one-time cleanup", `ReadTools`, `ArchiveDiscovery`, `WireNamesProbe`, the skill-authoring standard.
- **Probe labels, left alone** — `WriteSurfaceGuardProbe`'s two "the FIRST touch returns the one-time CONSENT prompt" labels describe first-touch behaviour that is still true, and address a maintainer rather than a caller.

### Filed, not built

**#386** `[GAP]` — no guard reaches tool-parameter `[Description]`s or prompt-builder strings. The reviewer's structural recommendation is endorsed as **evidence, not scope**: building the probe here would be work this fix does not charter. Both data points are cited — the three-pass hunt across two unguarded surfaces, and the fourth pass's 14 further sites. Unrouted.

### Take-or-drop commit amended

`de7d483` (was `bbb38d0`) gains the directed sentence: *A sweep with no known-red cell makes one first — sabotage a cell already proven red by hand — rather than running canary-less.* Still one commit, still Aaron's to take or drop.

**One consequence to flag:** the two inline review comments were anchored to `bbb38d0` — GitHub anchors to whatever was HEAD at review time, not because they were about the canary bullet. Amending moved that SHA, so those anchors now show as outdated. The comments remain visible in the conversation, and both findings are addressed in `1f71a61`; their substance is quoted above so the record does not depend on the anchor. The alternative was splitting the amendment into a second commit, which would have broken the take-or-drop semantics.

### State

Nine commits. `ci-all` **124/124**. No new rounds (§11 — directed folds from here). Merge is Aaron's explicit go.

Issue #388, opened 2026-08-20
[GAP] a mod's own root archives are reachable by name only when MO2 is not loading the mod


Avick3110:
## Goal

Read a file out of a specific mod's own BSA by naming **the mod** — the way a caller naturally addresses a donor ("read it from Bijin Warmaidens"), and the way the appearance skill's recipe tells them to.

## What fell short

Three instances of one shape, all measured at the F1 branch (`claude/f1-disabled-source`).

**(i) An enabled mod's name does not reach its own root archives.** `AssetResolver` names a BSA provider by the *archive filename*, and the F1 off-order lane is gated universe-first — so for a mod MO2 **is** loading, naming the mod reaches only its loose files. The caller must know to name the `.bsa` instead:

```
source_provider = "BijinLike"     -> 'BijinLike' does not supply '...' (the file is in mods\BijinLike\Bijin.bsa)
source_provider = "Bijin.bsa"     -> PLACED
```

Worse where that archive is bound to no active plugin: it is listed under **neither** name and both routes fail, while the identical folder *switched off* is reachable via the F1 lane.

**(ii) `nif_inspect` / `nif_set` `mod=` cannot resolve an off-order donor.** That lane matches `mod=` against active providers only and reports the mesh ABSENT:

```
housecarl_nif_inspect(<donor facegen mesh>, sections="paths", mod="DonorMod")
  ABSENT - no active mod or BSA provides this mesh path.
```

The refusal does not say that `mod=` named a mod outside the active set, so it reads as "the donor has no mesh" — one step before the flow where reading absent as "nothing to place" is the dark-face outcome. The appearance skill's sentence was corrected in the F1 PR to state this bound; the capability itself is unchanged.

**(iii) The asymmetry F1 creates.** After F1, a **disabled** mod's name reaches that folder's root archives; an **enabled** mod's name still does not. Naming the same mod means different reach depending on the MO2 tick, which is the opposite of the direction ruling O1 set for the surface.

## Why it is filed rather than fixed

The F1 branch is frozen on "no enabled name moves" — widening enabled-name semantics there would change behaviour the ruling deliberately held still. Ruling O1's stated rationale for rejecting the derive-the-path option was that a path guess **cannot enumerate root BSAs**; a name-based trigger delivers that only for the unticked half, which is what makes this a capability call rather than a defect in F1.

## Adjacency

[...]

Issue #387, opened 2026-08-20
[GAP] the asset source pole is singular, so a carry cannot follow an ordered from_source across two mod folders


Avick3110:
## Goal

Standalone-ize an NPC whose look is read through an overhaul's override while the FaceGen ships with the mod that *defines* the NPC — and where both of those mods are switched off in MO2. `housecarl_copy` takes an ordered `from_source=` list for exactly this shape (SPEC §3.1 AMENDMENT 2026-08-14 / ruling O2). The asset carry that follows it has to reach the FaceGen beside the **second** arm's plugin.

## What fell short

Measured on a constructed instance (`dotnet run --project src/housecarl-generator -- f1-i14`): `DisabledA` holds `Override.esp` (overrides the NPC), `DisabledB` holds `Donor.esp` (defines it), and the FaceGen exists **only** in `DisabledB`. Both unticked.

The **ancestor** builds a donor disk for *both* donor-side files and carries it:

```
donor read from: file 'Override.esp' (mod 'DisabledA' (DISABLED) ...); AUTO-WIDENED to the donor's
defining plugin 'Donor.esp' (mod 'DisabledB' (DISABLED) ...)
assets: 1 file(s) carried.
  - meshes\...\facegeom\Donor.esp\00000805.nif -> ...  (34 B, from donor folder (loose))
```

The **successor** reads the record fine and reports which plugin it came from — but the carry is a separate call taking **one** `source_provider=`. Naming the mod of the arm the record came from misses:

```
housecarl_bulk_place_asset(..., source_provider = "DisabledA")
  FAIL  'DisabledA' does not supply 'meshes\...\facegeom\Donor.esp\00000805.nif'
        - houseCARL also looked in the MO2 mod folder of that name, and read neither a loose copy
          there nor one inside that folder's own archives.
```

Naming `DisabledB` works. The caller has to already know which of the two mods ships the file.

## Why it is not simply an internal fix

The place tools receive one provider name per asset and hold no knowledge of the `copy` call that preceded them. Three bridges exist and each is its own decision:

1. **`source_provider=` accepts an ordered list** — a wire-type change on a shipped parameter (`string` -> `string[]`).
2. **Derive the mod folder from the FaceGen path's plugin segment** — this is the shape ruling O1 rejected as "a guess dressed as a path"; it should stay rejected.
[...]

Issue #347, opened 2026-08-16
[NOTE] Deferred empirical: fixture 3's live-install half (PR #346)


Avick3110:
PR #346 (`housecarl_copy` + `npc-appearance-copy`) shipped its acceptance differential as a **synthetic** run — both tools driven over constructed MO2 instances. The inventory's §13 states the run's evidence is complete only when both halves exist, and the second half is Aaron's own pass on his real install. That pass has **not** been run, and the PR merged without it by explicit ruling.

## What the test is

Fixture 3 — **a contested vanilla NPC on Aaron's real load order**. A vanilla-defined donor whose FaceGen path has more than one active provider (a replacer out-sorting the base BSA). Three things get checked:

1. **The carried FaceGen bytes are the NAMED donor's, not the VFS winner's.** This is the whole behavioural divergence from the ancestor: `copy_npc_appearance` reads the winner, `housecarl_copy` + the skill read the donor that was actually copied. On a contested path those are different files, and the synthetic run measures them as different bytes (ancestor 37B, successor 1B). On the real install the claim is that the face in game matches the records that were copied.
2. **The transplant / standalone claim is checked against the actual masters list** in the written patch. A vanilla donor read with no `from_source=` should read as an appearance transplant; a donor read through a named mod override should read as standalone and list that mod's absence. Both sentences are pinned synthetically — this checks them against a real order's masters.
3. **In game:** face, hair, and lip-sync while speaking. Lip-sync exercises morph data baked into the mesh, so a head that looks correct standing still can still be the wrong file — which is exactly the failure the named-donor read exists to prevent.

## Why it was deferred

The dev build is deliberately kept off the production install, so running it on the real order is not a thing a build session can do incidentally. It needs a deliberate, chartered pass.

## Where it belongs

The **pre-release empirical wave**, now being chartered (advisor-ruled, Aaron-go). **The merge is not the lock — the release gate is.** Nothing about #346 being on `main` asserts this evidence exists.

## Same class — cross-reference

The **F1 deferred asset differential** is the other half of this pattern. Ruling R1 put the disabled-donor asset lane outside PR #346, so fixture 2 ran records-only and the ancestor's `DonorDisk` carry has no counterpart to diff against. That comparison is only possible while **both** implementations exist, which is why `copy_npc_appearance` survives on `main` by ruling rather than being deleted when SPEC §6 subsumed it.

[...]

Issue #340, opened 2026-08-12
[Bug] nif_inspect / nif_set: the provider spelling mod= prints is not the spelling it accepts


Avick3110:
`nif_inspect` and `nif_set` tell the caller which providers supply a mesh, and the spelling they print is not the spelling `mod=` accepts. Following the message verbatim always refuses.

## Reproduced on `main` (`09195f0`)

Both tools match on the **bare provider name**:

```csharp
// LoadOrderService.cs:1320  (nif_inspect)   and  :1389  (nif_set) — identical
var pick = place.Sources.FirstOrDefault(s =>
    s.ProviderName.Equals(mod!.Trim(), StringComparison.OrdinalIgnoreCase));
```

…and print the name with a **kind annotation appended**:

```csharp
// LoadOrderService.cs:1323  (nif_inspect)   and  :1391  (nif_set) — identical
$"mod '{mod!.Trim()}' does not provide this path. Providers (winner first): " +
string.Join(", ", providers.Select(p => p.Name + " (" + p.Kind + ")"))
```

So for a mesh supplied by a mod folder `SomeMod`:

- **Printed** — `Providers (winner first): SomeMod (loose), Other.bsa (BSA).`
- **Accepted** — `SomeMod`
- **Rejected** — `SomeMod (loose)`, which is what the message just handed the caller

Passing the printed form back produces the same refusal again, re-listing the same strings. The loop terminates only when the caller guesses that the parenthetical is decoration — and MO2 folder names legitimately contain parenthesised text (`SkyUI (SE)`), so "strip the parenthetical" is not a rule that can be safely inferred, and applying it can remove a real part of a name.

`housecarl_asset_status` renders the same annotated form, so neither surface gives the caller a cue.

## Why this is filed rather than fixed in place

Found by an independent reviewer while reviewing PR #339, which fixed the *same class* on `place_asset`'s new `source_provider=`. That fix is on `main` now: provider names in a selector list are delimited by one shared formatter (`AssetSourceSelection.Describe`), the delimiter is a character a Windows name cannot contain, and a round-trip arm feeds the rendered tokens back through the real tool over deliberately hostile fixture names. The `nif` pair is outside that PR's scope — an untouched code path — so it belongs here.

## Fix shape

[...]

PR #389, opened 2026-08-20
F1: a named asset source resolves wherever the mod lives (ruling O1), and EL-2 runs

Avick3110:
RUN_ORDER row 3's F1, ruling O1. Substrate is **S2** (SPEC §2.3, assets/VFS).

`AssetResolver` builds over `comp.EnabledMods`, so a NAMED asset provider reached only ticked mods — while naming a *plugin* resolves it wherever it lives (SPEC §4.2's one-pole rule). The appearance flow could read a disabled donor's **record** and then had nowhere to read its **FaceGen** from. This closes that.

## The lane

A name the built universe has no answer for is looked for as `mods\<name>\` — the loose file, then that folder's root archives (the ancestor's `DonorDisk` shape; root archives are the capability that disqualified deriving paths in a skill).

**Universe first**, so no enabled name changes behaviour. **Naming is the consent and the only door**: the lookup is consulted under the Named pole alone, so an omitted provider, `*winner` and the contention listing all stay strictly inside the enabled universe — a mod nobody named cannot be read, be reported as winner, or appear as a copy to choose between.

**One seam.** `AssetSourceSelection.Select` takes the lookup as a parameter, so `place_asset`, `bulk_place_asset` and the closure copy's asset carry all reach the widened rule through the same call. The lane's I/O is `OffOrderAssetSource` (new file); `AssetResolver` keeps a thin delegating member.

## The two binding measurements (taken on the base, before any fix)

Committed as `98f9586`; re-runnable either side via `dotnet run --project src/housecarl-generator -- f1-measure`.

1. **`source_provider=` naming a disabled mod refused in all four cells** (both tools x with/without `source=`). The named pole was never consulted: with the path absent from the enabled universe, selection returned before reading the pole, so the refusal spoke only of "the active load order" and never mentioned the name passed. A name matching nothing anywhere returned **identical text** — a disabled donor and a typo were indistinguishable. The no-source remedy routed into the with-source refusal, which carried no remedy at all.
2. **`housecarl_copy` read the disabled donor's record; the carry missed.** The ancestor carried 34 bytes from the donor folder over the same donor.

Neither "this brief is wrong" condition held.

## EL-2 — RUN, clean over a single-donor fixture

`EMPIRICAL_LEDGER.md` EL-2, deferred by ruling R1 because the successor had no lane to be compared against. Run here while the ancestor still exists. Result doc: `research/EL2_DISABLED_DONOR_ASSET_DIFFERENTIAL_2026-08-20.md`.

Fixture 2's asset half is asymmetric on purpose and measures its own premise first — a contested FaceGen mesh (an **enabled** replacer) plus two paths only the switched-off donor provides. Per destination path:

| destination | ancestor | successor | verdict |
|---|---|---|---|
| `<clone facegeom>` | the VFS winner's (replacer's) | the NAMED donor's | **the inventoried divergence** (I5, §12) |
| `<clone facetint>` | donor's | donor's | identical |
| `...\donorhair.nif` | donor's | donor's | identical |

The divergence's **absence** would be the failure; the harness stops rather than passing on a clean diff there. No unexplained divergence. Both arm classes RED-checked by hand.

**Two bounds on that verdict, both stated rather than left implicit.**

1. The comparison covers **loose** donor-folder copies; the root-archive half is guarded per-side (`place-asset-guard` M3, M14, M18) because these records link to no path the committed archive contains.
2. The fixture is **single-donor** — one donor plugin in one unticked mod folder — so it does not exercise the ancestor's plural donor disks, where both donor-side files resolve to the same folder and a plural lane and a singular one give identical answers. The **two-arm auto-widen** shape was measured separately *after* the run (§6 of the result doc): the ancestor carries the FaceGen beside the second arm's plugin and the successor does not. That is a **known, filed divergence** — inventory I14, filed as #387 — not an unexplained one, so EL-2's stop rule is not engaged: that rule governs an unexplained divergence inside the run, and this is a measured, attributed divergence outside it.

So the clean verdict certifies that over a single-donor disabled donor the successor carries the same bytes as the ancestor on donor-only paths and diverges only where the inventory records (I5). It does not certify the two-arm case; #387 is where that lives.

## Review rounds — 2 rounds, 6 agents, 40 findings

**The class-stop fired**, and this PR carries its resolution.

- **Round 1** (3 agents, converging independently): the named-provider refusal made an **unconditional** claim about a **conditional** search — naming any mod the active order already provides files under never opened a folder, while the sentence said it had. Folded into a two-arm conditional.
- **Round 2** (3 agents, 2 converging): that conditional has **more than two states**. Measured false for `C:\Windows`, `..`, `DonorMod\sub`, `DonorMod.` — and a fourth state exists in source (`BuildLooseRoots` adds every enabled mod whether or not its folder exists). In all of them the candidate list is empty, so the caller got a false statement **and** no remedy.

Per §5 #11 the third instance was **not** folded; the seam went to Aaron, who ruled the revision below.

### The revision (Aaron-go)

`OffOrderReason` is the closed set of outcomes the lane can reach, one per exit in `Resolve`: `Found`, `NotConsulted`, `UniverseName`, `NotAFolderName`, `NoSuchFolder`, `NoCopyInFolder`, `FolderUnreadable`. The refusal keys **one sentence to exactly one value** through a switch expression with **no default arm**, so a new outcome is a compiler diagnostic at the render rather than a false sentence — verified by removing an arm and watching CS8509 name the missing one. Two outcomes render no clause and are *named* rather than defaulted. The unreadable arm carries the failing NAME and a concise cause as **data**; the sentence is built in the tool layer, so core stopped carrying prose (#337).

Every reason arm is pinned by its own cell, proven by a sweep that mis-reports each outcome as a neighbour: R1→M16/M17, R2→M13/M22, R3→M8/M9, R4→M19, R5→M21, R6→M27.

### Also folded, each measured first

`*winner` was being refused as though it were a mod name (routed on "a provider string was passed" rather than on the pole); the `#283` root-prefix hint was swallowed by the named-miss refusal; a trailing dot walked around the universe-first gate (`mods\Data.` opens `mods\Data`, and an *enabled* mod named that way would have been served off disk and reported "NOT enabled in MO2"); an unreadable folder or archive was reported as an authoritative absence; the success render contradicted itself across two adjacent lines; the second dead end (`source=` with no `source_provider=`) lacked the route out; and six stale-prose defects, including the appearance skill's claim that a disabled donor is "a records-only job".

### Refused, with grounds

1. **The double archive walk on a hit** — correct; batching it is a separate optimisation with its own shape.
2. **The per-path archive walk** (round 2) — same shape on a different axis, same grounds.
3. **"The `housecarl_copy` asset readback claim is false"** — measured false. The sentence is `CopyAssetPathsHeader` and `copy` renders it; a grep of `CopyTools.cs` alone cannot see that.
4. **8.3-alias aliasing** as a sibling of the trailing-dot class — theoretical: needs 8.3 generation enabled, and no realistic caller types one.
5. **`copy-differential` not being in `ci-all`** — deliberate and documented in `Program.cs`; the ancestor dies at 2.0.

## Coverage

`place-asset-guard` arm M: **16 cells → 27**. Eleven were empty and the **sweeps** found them, not the review argument. Sharpest: dropping the off-order flag from the *archive* branch alone shipped green through an entire sweep because every provenance cell tested a loose read; and the **archive half** of `IsUniverseProviderName` had no cell at all, since arm M's fixture has no active archives (M26 has its own instance for that reason).

Four sabotage sweeps, each from a committed state, each carrying the S1/M1 canary proven RED by hand first: 16/16, 24/24, a targeted 4/4, and the 6-way reason sweep. `ci-all` 124/124; `copy-differential` clean.

Twice the `[MustState]` wiring guard caught a new sentence reaching no render it drives; both times the fix was to drive the missing render.

## Dispositions

- **I14 — does NOT land here, and the GAP is all of I14** (#387). Measured as directed (`-- f1-i14`): two disabled mod folders, FaceGen beside the **second** arm's plugin. The ancestor carries it; the successor's carry, naming the mod of the arm the record came from, **misses**. The would-be-wrong clause fired — the place tools receive one provider name and hold no knowledge of the `copy` call, so the plurality cannot be an internal lookup order. Three bridges exist and each is its own decision; the issue names them, including one that may not have been considered (`copy`'s readback naming the mod **folder** each source resolved from, which is what the ancestor prints and the successor does not). I3 and I4 did land; I14's Data-folder half holds by construction and is guarded.
- **Capability fork (B)** — filed as #388, three instances, `#340` adjacency noted. The ruled freeze holds: no enabled-name semantics moved. In-branch, the appearance skill's sentence telling callers `nif mod=` works "exactly as for the placement" — false the moment this merges — states its bound instead (#374 class, caught pre-merge).
- **Universe-first bound** — left as-is; no runtime detection, no conditional sentence. One clause in `source_provider=`'s description. Two mod folders cannot share a name, so the only collisions possible are the reserved-ish names (`Data`, `overwrite`, an active archive filename), and F1 changes nothing there: those were unreachable before and are answered by the universe now. Tripwire: a real collision report promotes detection via gap report.

Both GAPs left **unrouted and unmilestoned** — routing is the advisor lane's stroke.

## Notes

Rounds ran with `isolation: "worktree"`; all six agent worktrees and branches were removed and verified gone. One dirty-restore mistake mid-session discarded an uncommitted file, which was redone — the lesson is the standing one: commit before sabotaging.

CHANGELOG `## Unreleased` line included.

Refs #347 (closes nothing — F1 has no issue; #347 closes on EL-1, untouched here).


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1585,17 +1589,43 @@ PlaceResult PlaceOne(PlaceRequest req, AssetResolver.AssetView view, string outD
...
+            // spelling they had just used. Measured in review round 1.
+            if (choice.Pole == AssetSourcePole.Named
+                && pick.Verdict is AssetSourceVerdict.NamedAbsent or AssetSourceVerdict.NoProvider)
+                return PlaceResult.Fail(rel, WriteSentences.PlaceSourceNamedAbsent(
+                    providerSel!, srcRel, pick.ProviderNames,
+                    pick.OffOrderReason, pick.OffOrderUnreadableName, pick.OffOrderUnreadableCause,
+                    // The #283 root-prefix hint, which this refusal used to swallow: a path taken off a record is
+                    // stored relative to meshes\/textures\, and naming a provider does not stop that being the
+                    // caller's actual mistake. Verified before it is offered, like every other site that shows it.
+                    AssetPathHint.AssetRootHint(view, srcRel)), winner);
```
Avick3110: **The named-pole refusal drops the `ReadIncomplete` caveat this call carried on `main`.**

`PlaceSourceNamedAbsent` takes no read-incomplete flag and nothing appends one here, but the new gate now swallows the `NoProvider` verdict for the Named pole — which is the case that used to fall through to the two sibling refusals below, both of which still render `NOTE: a BSA failed to read this build …`.

Measured both sides on the same fixture (an ACTIVE mod whose bound `.bsa` fails to open, so `ReadIncomplete` is true, and a path only that archive would hold):

`place_asset(asset_path=X, source=X, source_provider="HostMod")`

- **main:** `nothing in the active load order provides the source 'X'. NOTE: a BSA failed to read this build, so it may merely be unscanned (see the warnings).`
- **this branch:** `'HostMod' does not supply 'X' — that name is one the active load order already provides files under. (the winner pole is spelled *winner …)`

Same loss for the no-`source=` form (`res.ReadIncomplete`). A known-incomplete scan is now rendered as an authoritative absence, and the arm M cells cannot see it — that fixture has no active archives, so `ReadIncomplete` is false in every one of them.

Thread `srcRes.ReadIncomplete` into `PlaceSourceNamedAbsent` as another keyed clause (and give it a cell in a fixture that has a failing active archive).

src/housecarl-core/OffOrderAssetSource.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        string dir;
+        try { dir = Path.Combine(modsDir, name); }
+        catch { return new OffOrderLookup(null, OffOrderReason.NotAFolderName); }
+
+        string? unreadableName = null, unreadableCause = null;
+        if (!SafeDirectoryExists(dir, out var dirCause))
+            return dirCause is null
+                ? new OffOrderLookup(null, OffOrderReason.NoSuchFolder)
+                : new OffOrderLookup(null, OffOrderReason.FolderUnreadable, name, dirCause);
```
Avick3110: **This `FolderUnreadable` arm is unreachable, and the outcome it would have produced falls through to the strongest false claim in the set.**

`SafeDirectoryExists` (line 184) wraps `Directory.Exists` in a try/catch, but `Directory.Exists` never throws — it signals every failure, permission errors included, by **returning false**. So `dirCause` is always null and this branch cannot fire.

Scenario: `mods\DonorMod` is on disk but the current user has no traverse/read right on it (a folder copied off another machine, a backup restored with foreign ACLs). `Directory.Exists` returns false → `NoSuchFolder` → the caller is told *"houseCARL also looked under MO2's mods folder, and there is no MO2 mod folder of that name"* for a folder that is right there. That is a harder claim than the `NoCopyInFolder` one this file spends a paragraph guarding against, and it sends the caller to check their spelling rather than their permissions.

The cause has to come from an API that actually throws — `Directory.GetLastWriteTimeUtc` / `File.GetAttributes` on the dir, or folding the existence probe into `RootArchives`, whose `EnumerateFiles` does surface it.

src/housecarl-core/OffOrderAssetSource.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        // ---- loose, exactly as the VFS would layer it if this mod were ticked ----
+        string loose;
+        try { loose = Path.Combine(dir, relPath); }
+        catch { return new OffOrderLookup(null, OffOrderReason.NoCopyInFolder); }
+        if (SafeFileExists(loose))
```
Avick3110: **An unreadable LOOSE copy renders as an authoritative absence — the one thing the archive branch four lines down is written to prevent.**

`SafeFileExists` (line 191) swallows everything to `false`, so "the file is not there" and "the file is there and I could not read it" collapse into one answer. The lookup then falls through to `OffOrderReason.NoCopyInFolder`, whose sentence is *"houseCARL also looked in the MO2 mod folder of that name, and read neither a loose copy there nor one inside that folder's own archives"* — stated as fact.

Scenario: the donor mod folder's `meshes\` subtree carries a deny-read ACL (restored backup, folder copied off another user's drive). `place_asset(..., source_provider="DonorMod")` for the donor's FaceGen path refuses with that sentence; an agent following `npc-appearance-copy` reads it as "the donor has no mesh" and ships the patch without FaceGen — the dark-face outcome the skill's own new paragraph exists to prevent.

The archive half of this same method reaches `FolderUnreadable` and prints *"'Broken.bsa' could not be read (…), so this may be unscanned rather than absent"* (M21). The loose half should reach the same outcome. `File.Exists` cannot tell you, so the probe has to be one that throws — e.g. `new FileInfo(loose).Exists` still won't, but `File.GetAttributes(loose)` distinguishes `FileNotFoundException`/`DirectoryNotFoundException` from `UnauthorizedAccessException`/`IOException`.

Avick3110:
## Review fold — all three findings folded, none refused

All three reproduce at their cited lines, and they are one class: **a probe that cannot see "unreadable" feeding a sentence that claims authoritative absence** — the seam's own class, one level down, in the probes that feed the enum.

Commits: `e4456e8` (the fold), `0d62b5c` (the cell the RED sweep demanded).

### The binding measurement came first, and it refuted two of the three candidate APIs

The review had just proved one API-behaviour assumption wrong, so the replacement was measured on a real deny-ACL fixture before any sentence was wired — applied and removed in the same run.

| probe | denied folder | absent folder | usable? |
|---|---|---|---|
| `Directory.Exists` | **True** | False | no — cannot see unreadable |
| `Directory.GetLastWriteTimeUtc` | **a real date** | **1601, no throw** | no — cannot see either |
| `File.GetAttributes(dir)` | **`Directory`** | throws | no — cannot see unreadable |
| `Directory.EnumerateFiles` | `UnauthorizedAccess` | `DirectoryNotFound` | **yes** |
| `File.GetAttributes(file)` | `UnauthorizedAccess` | `FileNotFound` / `DirectoryNotFound` | **yes** |

So `File.GetAttributes` works on a **file** and not on a **directory**, and `Directory.GetLastWriteTimeUtc` — one of the two the review offered for the dir probe — cannot distinguish either case. The enumeration is the only dir probe that surfaces, which is the review's own alternative. Wiring the offered dir APIs without measuring would have shipped a second dead arm.

### F2 + F3 — one change at both sites

`ProbeFile` is tri-state on `File.GetAttributes`: Present / Absent / Unreadable-with-cause, with anything unexpected reading as **unreadable**, because the safe default for an unknown is to say it is unknown. Loose is probed **first**, so a denied copy returns `FolderUnreadable` rather than falling through to "read neither a loose copy there nor one inside that folder's own archives". `RootArchives` gained a third answer and the folder-missing signal now comes from its `DirectoryNotFoundException`. `Directory.Exists` and both bool probes are gone.

**No new enum members** — the fix makes existing outcomes reachable and honest, so the exhaustive switch and its CS8509 coverage are untouched.

### F1 — the caveat, threaded

`srcRes.ReadIncomplete` is now a keyed clause on the same sentence. The caveat is **hoisted to one home** on its third consumer: the two sibling refusals carried near-identical spellings of it, and the named-pole verdict no longer falls through to them, which is exactly how it went missing.

### Fixture blindness, stated honestly

The review is right that arm M has **no active archives**, so `ReadIncomplete` is false in every one of its cells — the caveat was unobservable there whatever the code did. That is the same blindness the provenance cells had. **M28** has its own instance with a bound archive that will not open, pins both source shapes, and pins that a *complete* scan does **not** carry the clause. The `[MustState]` wiring guard then caught the new sentence reaching no render **it** drives — the same blindness a third time — so its fixture gained a failing active archive too.

### What the RED sweep caught mid-fold

**M30 stayed green** under a sabotage of `RootArchives`' unreadable branch. Cause: M30 denies the whole folder, so the **loose** probe fires first and the enumeration is never reached — F2's branch had no cell. **M31** reaches it, and is F2's exact scenario: deny only the list right, without inheritance, so a path under the folder still probes as absent and the enumeration is what discovers the folder cannot be read. Measured before writing: `GetAttributes` on the child throws `DirectoryNotFound` there while `EnumerateFiles` throws `UnauthorizedAccess`.

The cells also caught a defect **this fold introduced**: the exception *message* carries the full on-disk path, and it was being rendered into a refusal — the one place names-not-paths forbids one. The cause is now the exception **type** as a short phrase (`access denied`, `an I/O error`), and M29/M30 pin that no machine path appears.

### Fixtures, honestly

M29/M30/M31 apply a **real deny ACL**, verify it actually bites before asserting anything, and remove it in a `finally`. A host that will not apply one **fails** the cell rather than skipping green — a branch that cannot be fixtured honestly is a signal, not a gap to work around.

### RED-check, from a committed state, canary first

| sabotage | cell |
|---|---|
| the named refusal stops carrying `ReadIncomplete` | M28 |
| the caveat clause is never rendered | M28 |
| the loose probe swallows again | M29 |
| an unreadable folder listing reads as absent | M31 |
| the cause reverts to the exception message | M29, M30 |
| **canary** — the off-order lookup never yields a source | M1 |

Every folded finding is caught by its own cell; canary RED as expected.

### Riders

- `SafeDirectoryExists`' doc comment promised a distinction `Directory.Exists` cannot produce — it died with the probe. The file's "an unreadable folder or archive is simply not a hit" claim is now false and says what it does instead.
- **Class sweep:** no third catch-everything-to-false probe feeding a rendered sentence exists in this branch's new code. `AssetResolver`'s two (`:462`, `:584`) are pre-existing and already ride `ReadIncomplete`.

Arm M is now **31 cells**. `ci-all` 124/124, `copy-differential` clean.

Holding here — merge is your explicit go.

Issue #398, opened 2026-08-22
[GAP] the merged check response leaves ~400 characters of every biting cap unusable — 395 held-and-never-written, plus whole-unit granularity


Avick3110:
Measured on the merged `housecarl_check` surface at PR #394's own acceptance band. Filed rather than
fixed: the ruling that produced the water-fill allocation dispositioned this class explicitly — *"that is
4a's one-unit posture, not this ruling's, and it comes back as its own issue rather than a silent fix."*

**Nothing here is wrong.** Both terms below are over-measures in the SAFE direction — an under-measure is
a response that lands over its cap. This is about room no row can use, not about a false claim.

## What is measured

`housecarl-generator check-measure --band --mo2 <instance>` now decomposes the headroom by subtraction.
Live order, errors family, text lane:

```
   cap   chars  headroom = held,unwritten +  unspent    (heldTotal  written)
  2000    1665       335 =             88 +      246    (     1753     1665)
  3000    2543       457 =            395 +       61    (     1753     1358)
  4000    3547       453 =            395 +       57    (     1753     1358)
  6000    5536       464 =            395 +       68    (     1753     1358)
 10000    9518       482 =            450 +       32    (     1753     1303)
```

The two terms sum to the headroom (to within one character at three of the five caps; the residual is
printed rather than absorbed).

### Term 1 — 395 characters held and never written, constant across the band

The response holds **1,753** characters out of the cap before a single unit is written: the cap less the
body budget (the response-level reserve for its accountings and boundaries) plus the row budget's own
subtraction for the measured fixed part and the demand reserve. It then writes **1,358** of them. The
difference is room subtracted from the row budget that nothing spends.

It is constant at every biting cap, which is what identifies it: this is not a budget the fill failed to
hand out, it is a fixed over-measure of what the response owes outside its units. A reserve is an upper
bound by construction — a closing disclosure says a different thing at a different length depending on
what fit — so the room held is the widest it could be and what it writes is usually less.

At the ~49 characters a histogram row costs on this order, 395 characters is about **eight rows** the
by-SOURCE axis does not get.

### Term 2 — whole-unit granularity, real but small

This was the standing hypothesis for the whole ~400, and the measurement does not support that: it is the
32–68 character term, not the 395 one.

It is confirmed where it applies. At `max_chars=4000`, `HistogramByTarget` held **48** unspent characters
against a next unit costing **49** — and it first renders one character of cap wider, at 4001. A subject
[...]

Issue #396, opened 2026-08-21
[GAP] a dialogue verdict straddles the record and asset substrates, so no epoch is stamped and none is owned


Avick3110:
**Provenance:** flagged by 4b phase 2 (the dialogue fold on `claude/check-merge`), triaged and measured by
phase 3. Filed rather than folded: an honest stamp needs cross-substrate epoch semantics, which is design
work, and this merge's rule is that capability growth files.

## What happens

`housecarl_validate_dialogue` stamps **no epoch**, and `LoadOrderService.ValidateDialogue` records why:

```
// EPOCH: deliberately NOT stamped (PR #305 third round, named not silent) — the report is one build's answer
// (core pins a view) but half its verdicts come off the ASSET substrate (.fuz/.pex presence), outside the
// record fingerprint; the dialogue-fold wave (W3) owns an honest stamp for it.
```

The decision is right and stands. What is stale is the **owner**: W3 completes at the `housecarl_check` merge PR without
delivering one, so the comment points a future reader at a wave that has landed. `findings=['dialogue']`
on `housecarl_check` inherits the same absence — its section carries no epoch, deliberately, and phase 2
declined to stamp a half-true one.

Every other findings family on `housecarl_check` DOES stamp: the errors family carries `epoch` and
`epoch_covers_all_inputs`, the scripts family carries `epoch`. The dialogue family is the one section of a
merged response with no freshness claim at all, and a consumer comparing two responses has no way to know
whether they describe the same build.

## The measurement

Phase 2's "half its verdicts" was a reading, not a number. Taken on the live order
(`check-differential --mo2 ... --seed 03372B:Skyrim.esm`, ARR 2.0, 3800 plugins), one vanilla quest seed:

| substrate | what was counted | seed `03372B:Skyrim.esm` (235 topics, 14.8 s) |
|---|---|---|
| **RECORD** — inside the fingerprint | input issues + per-topic graph issues | 0 findings |
| **ASSET** — outside it | verdicts taken by looking at a file through the VFS | **26** |
| | · voiced lines (`.fuz` on disk) | 0 (this quest's lines are unvoiced) |
| | · result scripts (`.pex` chain) | **26** (0 not bound-and-compiled) |
| | · `.seq` coverage + staleness | 0 (this quest is not start-game-enabled) |

Read it as the SHAPE, not as a ratio: this seed is clean, so the record row is 0 because there was
nothing wrong, not because nothing was checked. What the numbers establish is that 26 of this
[...]

Issue #395, opened 2026-08-21
[GAP] the scripts findings family has no off-order lane, so a pre-enable verify sweep cannot check a fresh patch's script bindings


Avick3110:
**Noticed while building** the merged `housecarl_check` surface (4b phase 1b, branch `claude/check-merge`). Filed rather than folded in: it is capability GROWTH, not merge work, and the merge states the asymmetry instead of closing it.

### What the two families do differently today

`housecarl_check_errors` resolves a `plugins=` name that is **not in the active load order** on disk — any mod folder, enabled, disabled, or not yet listed in MO2 — and sweeps it OFF-ORDER: its own records, links resolved against the active order plus the file's own definitions. That is the pre-enable verify sweep for a patch houseCARL has just written.

`housecarl_validate_scripts` has no such lane. A name not in the active order is refused outright.

### What the merged surface does about it

`housecarl_check` hands the scripts family the **active subset** of `plugins=` and states, in that family's own section, which named files it did not sweep and why. Where every named plugin resolved off-order, that family's scope is empty and it says so — its zero counts sit under the sentence explaining them, so "0 unbound" cannot be read as "looked, found none". That is zero-loss (the scripts family never had the lane) but it is not the capability a caller wants.

### The gap

A fresh patch is exactly the artifact whose script property bindings are most worth checking before it is enabled — the CK's auto-add-property bug produces unbound properties on records a patch has just touched. Today the only way to check them is to enable the plugin in MO2 first, which is the thing the pre-enable sweep exists to avoid.

### Why it looks tractable

The `.pex` chain resolves through the asset layer against the active order, not through the record index — so an off-order plugin's attached scripts are readable in principle. What the sweep needs is the same on-disk locate the errors family already performs (`LocatePluginFileOnDisk`, including its ambiguous-filename refusal) and a record enumeration over that file, which the read side already does for `housecarl_read_plugin_file`.

Unknown, and worth measuring before committing to it: whether an off-order plugin's VMAD entries resolve their script class names the same way, and what the bound-vs-unbound comparison means for a record whose masters are not all present.

Avick3110:
**A third family, and the same gap in a different shape** (4b phase 2, same branch). The dialogue family is now on the merged surface, and it also has no off-order lane — but not for the scripts family's reason, so recording the instance here rather than assuming it is the same fix.

[...]

PR #399, opened 2026-08-22
housecarl_check — one sweep, three finding families, one max_chars divided fairly

### LINKED ISSUE #394: [GAP] counts_only histogram axes spend one budget in series, so the by-SOURCE axis renders zero rows at biting caps
**Provenance:** Aaron's second review of PR #391 (finding R4, MEDIUM, inline on `src/housecarl-mcp/ReadTools.cs:1333`). The review deliberately proposed no answer — "splitting the budget per axis has its own cost, and the issue was explicit that picking is a design call." Filed here rather than folded on that branch for the same reason. **Unrouted:** no wave, no plan, no priority assigned by the filing session.

## What happens

`counts_only=true` renders its two histogram axes through one budget, in series. `AppendHistograms` reserves both axes' fixed parts up front (that part is #392's fix and it works), and then `AppendHistogram` runs for the by-TARGET axis first and emits its rows against the whole remaining budget. At any cap where the budget bites, the by-SOURCE axis renders nothing.

Measured on the 74/180-row shape #392 took off the live ARR 2.0 order, `max_chars=2000`:

```
dangling ref(s) by TARGET plugin (74 distinct):
     <21 rows>
  ... [53 more row(s) — raise max_chars= to see them]

dangling ref(s) by SOURCE plugin (180 distinct):
  ... [180 more row(s) — raise max_chars= to see them]
```

Reproduced independently while folding that review: 74/180 axes at `max_chars=2000` gives TARGET 50 rows and SOURCE 0.

## Why it matters

The by-SOURCE axis is the one #344 exists for — "how much of this dangling total did my mods introduce". A caller who runs `counts_only` for that question gets an axis header, no data, and a remedy naming a knob they may not want to raise that far.

This is **not** a silent failure and not a regression: #392's fix made the drop visible and correctly named, and PR #391 confirms it stays visible at every cap in the band. What it does not do is make the split fair, and the axis that loses is the one people ask for.

## Why it is a design call rather than a fix

Splitting the budget per axis is not free and there is more than one defensible rule:

- **Equal split.** Simple; wastes room when one axis is short, and a 2-row axis would hold half the budget it cannot use.
- **Proportional to distinct count.** Follows the data; a 4000-row axis then crowds out a 6-row one, which is the same complaint with the plugins reversed.
- **Round-robin rows.** Fairest per row; interleaves two renders through one builder, which the current one-axis-at-a-time loop does not do.
- **Reverse the order** so #344's axis renders first. Cheapest; moves the problem rather than solving it, and silently reorders a response people read top-down.
- **Leave it, and let `findings=` / a future per-axis selector be the answer** — the caller who wants the SOURCE axis asks for that axis.

Each changes what a caller sees at a biting cap, and none is obviously right without deciding what `counts_only` at a tight cap is FOR.

## Where it lives

`ReadTools.AppendHistograms` / `AppendHistogram` (text), `JsonWire.WriteHistograms` / `WriteHistogram` (json). Both transports spend in series; whatever rule is chosen has to land in both, from one source, per the branch's settled render-aware-accounting decision.

Avick3110:
Fixes #394

`housecarl_check` merges the three derived-findings families — `check_errors`, `validate_scripts` and
`validate_dialogue` — behind one `findings=` vocabulary, and divides one `max_chars` among the families that ran
instead of spending it in series. All three ancestor tools stay registered and unchanged.

**W3 completes at this PR.**

---

## How this got built, and where it stopped each time

Row 4 of the run order carried a binding sizing rider: phase at a seam rather than run one session to exhaustion.
It worked — and phase 3 still needed a fourth session, by advisor revision, when the escalation below landed
mid-phase.

| phase | what it landed | seam |
|---|---|---|
| 4b phase 1a | the allocation, and the emission layer it divides | allocation landed, render merge next |
| 4b phase 1b | the merged render, both transports | render merged, dialogue fold next |
| 4b phase 2 | the dialogue family | dialogue folded, phase 3 next |
| 4b phase 3 | zero-capability-loss verification, the acceptance band, review round 1, the §4 escalation | **red on purpose** — the rebuild the ruling ordered was half-built |
| 4b phase 3b | the rebuild to green, the orchestration arms, round 1's seven folds, review round 2 | **class stop** — two of round 1's classes recurred; the seam went to Aaron |
| 4b phase 3c | the shape matrix, the A1 regression closed, two more defects it found underneath | matrix green, the class-(i) seam next |
| 4b phase 3d | the outcome-value migration, the lower list, the harness fixes | outcome value in, mechanical tail next |
| 4b phase 3e | A4, the `GetProperty` conversion, the band diagnostic, the differential's dialogue lane, review round 3 | this PR |
| directed fold | Aaron's four review findings + the four defects round 3 listed — seven commits, 20/20 RED sabotage, one gap filed (#400) | this PR, at Aaron's merge gate |

## The #394 escalation chain, including the part that is not flattering

Round 1 found that redistribution never fired in the default listing lane: no listing subject ever signalled
completion, so a family finishing under its share never handed the remainder on. Measured on check-guard's own
fixture at the 80,000 default — errors alone 8,998 characters, scripts alone 38,253, **both whole 47,251, inside
the cap** — the merged response stopped at **49,440**, cut 5 of 40 record sections, reported `truncated: true`,
told the caller to raise `max_chars=`, and left **30,560 characters unspent**.

Building the obvious fix broke monotonicity: at `max_chars=3206` the response was 2,709 characters and carried a
scripts section; at 3,226 it was **2,373 with none**, because the errors family's newly affordable dangling entry
cost the scripts family its whole section. A wider cap returning less makes the response's own printed remedy
false. Under a sequential recount the two are mutually exclusive, so this went up as a §4 escalation rather than
being decided at the keyboard.

**The ruling was that water-filling IS what option D had already ruled** — "hand out one row's worth at a time; a
subject that runs out drops out, its remainder goes to the others", run to completion, is max-min fairness. What
phase 1a built instead was a SEQUENTIAL RECOUNT, and **the advisor lane ratified that as equivalent at phase 1a.
It is not equivalent, and the two defects above are the difference.** That ratification error is stated here
because it is the record: the deviation was introduced at build time and blessed on review, and it cost a phase.

A second escalation followed once the rebuild surfaced two things the first ruling had not priced. Both were
ruled the strict way:

- **json unit costs were measured at the wrong nesting depth.** Ruled: measure every unit type at TRUE depth
  under the response's own `WriterOptions`. The upper-bound-with-slack relaxation was **refused** — divergence as
  slack is forbidden, because an under-measure is a response over its cap and an over-measure is the stranding
  defect one level down. Allocation must equal spend, per subject, on a full fixture, in both transports.
- **the row budget excluded only the pieces that called `Reserve`.** Ruled: a MEASURED, not assembled, fixed-part
  pass, composed through the real one-source composers with zero units.

## What the rule is now

Max-min fair (water-fill) over MEASURED demand, hierarchical — families over the row budget, then a family's
subjects over that family's allocation — computed before the render, from the budget and the demand vector alone.
Three properties follow by construction, and each has an arm:

- **monotone in `max_chars`** — no subject ever spends fewer characters at a wider cap. Swept over every integer
  cap in a 9,000-wide band, both transports.
- **no stranding** — a call whose whole demand fits renders everything and claims no cut. Defect 1's fixture is
  the permanent arm, asked at the default AND at the tight fit: the smallest cap at which nothing is cut must sit
  within a bounded margin of the response the sweep actually returns.
- **allocation equals spend** — exactly, per subject, both transports. This is the arm that holds exactness: a
  one-byte-per-unit separator error reddens it.

## Measurements

| | |
|---|---|
| `validate_scripts` overrun (§2.4) | 80,673 chars against its own 80,000 cap → **79,784** |
| serial second family at the no-arguments call | inherited **400** characters of an 80,000 budget |
| Defect 1, before | 49,440 of 80,000 with 30,560 unspent, 5 of 40 sections cut, `truncated: true` |
| pin 6 — the demand pass, live order | **553 µs** (text) / **3,593 µs** (json), against whole renders of 2,655 µs and 7,729 µs |
| …against the sweeps it rides on | errors **38.8 s**, scripts **522.1 s**, over 180,028 findings across 31,481 record sections |

Pin 6's number is the point of the O(budget) bound: the demand pass is noise against the sweep that produced the
findings, not merely "small".

## The acceptance band, re-taken on the live order

Post-rebuild and post-outcome-value, because the allocation moved three times since the first table: the roster
became a water-fill participant (3c), the fixed part grew (3c), and the dialogue head and accounting changed
shape (3d). One counts_only sweep, both surfaces rendered from it.

| transport | cap | surface | chars | TARGET | SOURCE |
|---|---|---|---|---|---|
| text | 2000 | `check_errors` | 1806 | 18/74 | 0/180 |
| text | 2000 | `check` | 1665 | 0/74 | 0/180 |
| text | 4000 | `check_errors` | 3811 | 74/74 | **5**/180 |
| text | 4000 | `check` | 3547 | **29**/74 | **24**/180 |
| json | 4000 | `check_errors` | 3039 | 19/74 | 0/180 |

**#394's own claim is answered.** At 4,000 the ancestor gives the by-TARGET axis all 74 rows and the by-SOURCE
axis 5 of 180; the merged surface splits them 29 and 24. The starved axis is no longer starved.

**And the cost half, because a band without it is the fair half of a trade.** The merged framing carries a scope
sentence the ancestor does not — **360 characters**, above everything a budget can refuse — so at the same cap it
renders fewer rows in total:

| cap | `check_errors` | `check` | headroom `check` left |
|---|---|---|---|
| 2000 | 18 rows / 1806 | 0 rows / 1665 | 335 |
| 3000 | 49 / 2809 | 25 / 2543 | 457 |
| 4000 | 79 / 3811 | 53 / 3547 | 453 |
| 6000 | 132 / 5801 | 110 / 5536 | 464 |
| 10000 | 237 / 9807 | 218 / 9518 | 482 |
| 20000 | 254 / 10530 | 254 / 10926 | 9074 |

The merged response with no body in it at all is **1,933 characters**. Above 12,000 both surfaces carry
everything.

### Where the headroom goes — the diagnostic, finished, and it refuted its own hypothesis

The standing hypothesis across three handoffs was whole-unit granularity. `check-measure --band` now decomposes
the headroom by subtraction, and the terms sum:

```
   cap   chars  headroom = held,unwritten +  unspent   (heldTotal  written)
  3000   2543       457 =            395 +       61    (    1753     1358)
  4000   3547       453 =            395 +       57    (    1753     1358)
  6000   5536       464 =            395 +       68    (    1753     1358)
```

Granularity is **real and confirmed** — at cap 4,000 `HistogramByTarget` held 48 unspent characters against a
next unit costing 49, and it first renders one character of cap wider — but it is the **32–68** term. The
dominant term is a **constant 395**, decomposed by holder:

| holder | held | wrote | over by |
|---|---|---|---|
| accountings + boundaries | 566 | 350 | **216** |
| axis frames + closing disclosures | 364 | 186 | **178** |

Both over-hold for one reason: the room is reserved before the render knows what was cut, and the sentence's
length depends on what was cut. That is an upper-bound over-measure in the **safe** direction — an under-measure
is a response over its cap — so nothing here is wrong. **Filed as #398, not fixed**, per the allocation ruling's
own disposition of this class; tightening it means cut-aware release machinery, which is design work and not a
fold at a review cap's last round.

The decomposition closes to within **one character** at three of five caps. The residual is printed as
`MISMATCH by 1` rather than absorbed into a term.

## Zero capability loss

`check-differential` compares the merged response against each ancestor's over FACTS, not bytes — json key by key
against the family object, text as a SET of lines where every ancestor line must appear. Zero-loss cells run at
`max_chars=4000000`, so a divergence is a capability divergence and never a budget one. Deliberately not in
`ci-all`: it needs a live MO2 instance, and it is acceptance for this PR rather than a standing guard.

**Re-run at phase 3e and again after the directed fold: 24 cells, 0 with divergences both times.** The fold changed a json wire shape, and this is where that change is NOT visible: the lanes below compare the ancestor against the merged surface, and `scan_errors` is written by one shared writer, so a shape change moved both sides together. That blind spot is why finding 1 ships a wire cell of its own rather than a differential lane. Every zero-loss cell green key-for-key in json and
line-for-line in text, and every refusal carried with its own reason — blank FormID, malformed FormID, unknown
type, blank plugin name, `exclude=` unrecognized token, `exclude=` naming a file nothing matches, a plugin found
nowhere, and the scripts family's own two.

**A lane was ADDED at phase 3e, and it found a defect.** The differential covered the errors and scripts families
and asserted the dialogue family in prose — the one family whose caller-facing vocabulary this branch rewrote,
so the one the zero-loss claim could least afford to leave unchecked. L6 does the same asymmetric comparison the
others get. Its first run found that the merged surface rendered the quest's CK-parity **issues** and never the
OK line, so a quest whose parity was fine said nothing at all and a caller could not tell a check that ran and
passed from one that never ran (Q3). The ancestor has always stated it. Fixed as one shared sentence both
surfaces read (`c81b3c4`).

L6 is **text only**, and that is a fact about the ancestor rather than a gap: `housecarl_validate_dialogue` has
no `format=` parameter, so there is no json document to compare. The merged surface ADDS a json transport for
this family; an addition is not a loss, and `check-guard` pins it.

Result on seed `03372B:Skyrim.esm` (quest MQ101, 235 topics): **ZERO LOSS**, all 2,141 ancestor text lines
carried, 329 recorded re-spellings across five declared classes. Each class is dispositioned **only where the
merged response actually carries the fact**, so a future change that quietly stopped carrying it fails the cell
instead of passing it.

**The lane also caught a loose check of its own author's.** A sabotage removing the quest's NAME from the merged
seed head came back green — because this order's topic EditorIDs are prefixed with their quest's name
(`MQ101HelgenIntro` and siblings), so a whole-response `Contains("MQ101")` was satisfied by coincidence rather
than by the fact being carried. The disposition now reads the merged seed-head line itself (`6c2ad84`). Same
shape as the arms this branch spent two rounds fixing — an assertion passing for a reason unrelated to what it
claims — one level up, in the acceptance instrument rather than a guard.

**The four intentional divergences, named.**

1. **The ancestor's HEAD line** — the merged response writes a seed head instead, carrying the same facts (kind,
   name, FormID, topic count) plus the winning plugin the ancestor does not state.
2. **Class 8, the effective merged INFO order** — an ordered sequence, not a findings list. Sent to
   `records project=info_order` by SPEC §6.1. Dispositioned only if the boundary both says the order is not here
   AND names the destination.
3. **The three standing limits** — stated in the family's boundary sentence instead of a trailing section, each
   checked per fact.
4. **The dialogue vocabulary** — the ancestor's one word `validated` split into named / reached / validated /
   unreachable (round-2 finding B1).

**Recorded drops: two, both from `validate_dialogue`, neither a capability.**

1. `(reusing the create-time teeth over every existing line)` — a provenance note about how the voice and
   result-script checks are implemented, not what they check. The checks carry word for word.
2. `(xEdit INOM/INOA parity)` — the label naming class 8's xEdit counterpart.

**A correction to phase 2's record.** Phase 2 recorded the whole of class 8's teaching as a drop. Verified at
source, that record was broader than the fact: the mechanism teaching — first passing line, re-listing appends to
the bottom unless the plugin carries its PNAM, a pure reorder invisible to a diff — lives whole at the redirect
destination, `housecarl_records form='info_order'`, in near-identical words. Only the xEdit label is absent from
both surfaces.

**Recorded drops: two, both from `validate_dialogue`, neither a capability.**

1. `(reusing the create-time teeth over every existing line)` — a provenance note about how the voice and
   result-script checks are implemented, not what they check. The checks carry word for word.
2. `(xEdit INOM/INOA parity)` — the label naming class 8's xEdit counterpart.

**A correction to phase 2's record.** Phase 2 recorded the whole of class 8's teaching as a drop. Verified at
source, that record was broader than the fact: the mechanism teaching — first passing line, re-listing appends to
the bottom unless the plugin carries its PNAM, a pure reorder invisible to a diff — lives whole at the redirect
destination, `housecarl_records form='info_order'`, in near-identical words. Only the xEdit label is absent from
both surfaces.

## The two phase-2 flags

- **A refused family's json accounting wrote six zeros about subjects it does not have — FIXED** (`56c7843`).
  Reproduced live, then fixed by extending the frame's own stated rule one level: a field named for a subject is
  present exactly where that subject is. Phase 2 called the frame pre-existing; re-triaged, the frame is, but the
  SHAPE that exposes it is one this merge invented — before it, a refusal early-returned `{error, epoch}` and
  never reached that writer.
- **The unowned dialogue cross-substrate epoch — FILED as #396**, with an in-branch honesty fix. The decision (no
  stamp) is unchanged and right; what was stale was the OWNER, since W3 completes here without one. Measured on
  one vanilla quest seed of 235 topics: **26 verdicts** are taken by reading files the record fingerprint does not
  cover, and the `.seq` staleness verdict is a file-mtime comparison no fingerprint expresses.

**#395 and #396 stay open. This PR closes #394 and nothing else.** Both were routed on 2026-08-21 by the
advisor lane on Aaron's go ("route as recommended") to RUN_ORDER **row 8 / the W5 charter's routing-call
list**, and both are deliberately unmilestoned until that charter decides them — the routing comments are on
the issues themselves. #395 is the scripts family's missing off-order lane, filed under the
capability-growth-files-never-folds boundary; the merged `check` states the asymmetry honestly, so nothing is
silently wrong today. #396 is the cross-substrate dialogue epoch, routed beside #353 in the read-honesty
family because the epoch-semantics rule wants deciding once rather than per tool.

## The review rounds — three, and the cap is spent

Every round: fresh agents, isolated worktrees, seeded with the worktree, the branch, the base, "the repo
documents its conventions in CLAUDE.md", and the settled-decisions list — never what changed, never where the
risk was, never what an earlier round found.

### Round 1 — seven findings, all folded, all armed

Redistribution never firing in the listing lane; the roster's accounting composed before the rows it counts; the
dialogue scope sentence claiming a completeness the response denied; four json fields moved behind a lane gate
the reserve did not measure; and three more. All seven folded with arms.

### Round 2 — the CLASS STOP

Three reviewers. Round 1's classes recurred:

- **class (i)** — claims composed from state captured at the wrong moment or in the wrong family's units —
  **five verified instances**, plus a standing probe that reproduced.
- **class (iii)** — behaviour with no arm that can fail — **six verified instances**.
- **class (ii)** — redistribution not signalled — closed by water-filling and untouched.

**And a regression one of round 1's own folds introduced (A1).** The finding-A fold moved the excluded-plugin
roster to the front of the composition and measured its room into the allocation pool — while the response-wide
emission test still read the full budget. Measured: json `check` returned **4,494 against a 4,000 cap**, 6,559
against 6,000, 8,494 against 8,000; text with three families 6,455 against 6,000. Floors were 3,692 / 5,560, so
these were not floor-bound. The overrun notice that fired took the "a body unit overshot" arm and said a unit was
written before its size could be measured — false — and its `raise to at least N` was short by a constant, so
following it looped indefinitely (6,000 → 6,123 → 6,201 → 6,279, over by ~410 every time).

**That fold was reasoned about carefully, its commit message states the exact invariant it breaks, it shipped
with an arm, and the arm passed** — because the arm asked about the accounting's claim and not about the cap.
That is the sharpest evidence in this PR for why §5 #11's class rule exists.

**The ruling** was a per-response OUTCOME value: composed once, after the render knows what the response did,
read by every caller-facing sentence and every json field. Replace-not-add, with a remaining-literal inventory
that IS the completeness claim.

**Its first save came at authoring time.** Naming the dialogue head field `seeds_unreachable` collided with the
section's `seeds_unreachable` ARRAY — a class-(i) instance in the same family object, created by the migration
itself. The duplicate-key arm caught it before the commit. That is the concrete evidence the ruling asked the
architecture to provide: the seam finds its own class at authoring time instead of round N finding it later.

Class (iii) was ruled a **shape matrix**, fixed once rather than as eleven more arms: an inventory of the shapes
the merged surface produces — now **31**, every family subset x listing/counts_only, crossed with a refused
family, an off-order plugin, a forty-row roster and multi-seed dialogue — with the allocation, cap and remedy
properties driven over all of them at every integer cap in both transports. It was built BEFORE the fixes and RED
on them, and widening its band surfaced a monotonicity break nobody had seen.

### Round 3 — the cap's last, and it fired the class rule again

Three reviewers. **Two of round 1/2's classes returned**, so §5 #11 fires and **nothing from this round was
folded at the time** except one documentation figure. Every finding below was verified at source before being
dispositioned. The disposition column carries what the DIRECTED FOLD then did with each, after Aaron's review —
the class rule stops a round from folding its own findings unchecked, not from folding them under direction.

| # | severity | finding | class | disposition |
|---|---|---|---|---|
| A1 | **HIGH** | A `DLVW`/`DLBR` seed's section renders its head and nothing else — the CK-parity verdict that is its ONLY check goes unstated when it passes, and the family boundary then asserts graph, LinkTo, `.fuz` and result-script checks that never ran for that input kind. The ancestor states both a kind-specific CK-parity line and a scope note. | **(i) recurrence** | **FOLDED** in the directed fold (`5c4ca62`) |
| B1 / C2 | MEDIUM | The text lane's dangling-entry unit marker `"  dangling ref "` matches nothing any composer emits (`ComposeDanglingLine` writes four spaces, the source FormID, then `[target not defined by any active plugin]`). So `MATRIX-NO-STRANDING`'s whole-render predicate and `NothingStranded`'s fingerprint are blind to that subject. Measured: the arms accept a tight-fit cap up to 219 characters below the cap that actually renders every unit. | **(iii) recurrence** | **FOLDED** in the directed fold (`4c87867`), with the measurement |
| C1 | MEDIUM | A blank or whitespace entry in `plugins=` is filtered out before `noneInScope` is computed, so `findings=['scripts'] plugins=['  ']` sweeps the WHOLE load order — a ~468-second sweep the caller did not ask for, with `plugins=` silently discarded and no sentence saying so. The errors family and the ancestor both refuse the same input by name, so on a two-family call the two families disagree about one list. | new class | **FOLDED** in the directed fold (`9b90b71`) |
| C3 | LOW | `exclude=`'s typed-name typo refusal is skipped when the scripts scope is empty by construction, so a typo returns a clean section while the parameter text promises the call is refused. Reachable only when scripts is selected without errors. | new class, same seam as C1 | **DISPOSITIONED** in the directed fold (`5065123`) — the finding moved under measurement |
| C4 | LOW | One measured figure stated three times, one not reconcilable: 468 seconds measured, "~8 minutes" in the tool text, "~8.5 minutes" in the changelog. | documentation | **FOLDED** (`a9a3640`) |

**Why B1/C2 was not the one-line fix it looked like — and what measuring it found.** Correcting the marker
changes what those arms assert, and the reviewer's measurement said the corrected predicate rejects caps the arm
accepted. Answering that at a review cap's last round, with no round 4 to check the answer, is the pattern the
class rule exists to stop. It was answered in the directed fold instead, by measurement: the tight-fit search was
instrumented and the matrix run with both markers, and of 62 shape/lane cells **3 moved, each by exactly 219
characters**, all three inside the 2,200-character `TightSlack` with `MATRIX-NO-STRANDING` green throughout. **No
text-lane stranding gap is exposed** — what the dead marker cost was the arm's resolution, not a defect it was
hiding. The arm that ships is the rule rather than the correction:
`MATRIX-EVERY-FINGERPRINT-TERM-IS-REACHED` asks that every fingerprint term be non-zero on some shape, because a
term that is always 0 does not fail a comparison — it removes that unit from every comparison. It immediately
found two more of the same class, both mechanical and both folded there.

**Why C1 was not folded despite being the most user-visible, and how the fork was picked.** The fix forks on
settled decisions 9 and 17: a blank name can refuse the whole call (what happens today when the errors family is
selected) or refuse the scripts family in its own section. Picking that fork is a design call, and there was no
round 4 to check whichever arm was picked. Aaron picked it in the directed fold: **whole-call**, at the same
point every other lane refuses, together with R3 as one seam — a value malformed as INPUT to every family that
could have used it is not a family-local ground. `exclude=`'s **scope-matching** half stays family-local, which
is the split the fold had to preserve and arms in both directions.

**Refusals, with grounds.** Nothing from round 3 was refused as wrong — every finding reproduced at source. What
was refused is the *disposition* of folding them, on the class rule, and the grounds are above. Two reviewers
also reported guard arm counts one higher than this PR states (69/138/40 against 68/137/39); that is the guards'
own `ALL PASS` summary line counted as an arm, verified by re-running and counting both ways.

**What the reviewers checked and found clean**, worth recording because it is the load-bearing half:
`BodyAllocation.WaterFill` is correct max-min fairness — group keying, saturating demands, the closed-form loop,
remainder handling; demand-vs-render agreement per subject in both transports, including the per-plugin reset of
the json `entries` counter and every `first`/`subsequent` flag; the json depth table against each writer's live
`CurrentDepth`; `CapTooSmall`'s raise-to can never name a cap below the caller's, and its one-extra-digit growth
bound is sufficient across a power of ten; the roster's single-owner rule cannot pick a family whose accounting
returns early; and the grounds-are-one collapse, three-list family state and family-local refusal rendering all
behave as decisions 9, 17 and 18 describe. One reviewer extended the matrix with **10 shapes it does not carry**
— `missing_masters` only, counts_only with null axes, counts_only with EMPTY axes, a zero-topic dialogue seed, a
roster with 1,200-character parse reasons — and drove every property over 41 shapes at every integer cap in both
transports: all still pass.

### What the rounds cost, and what the rule bought

Three rounds, nine fresh reviewers. Round 1 folded seven findings and introduced one worse than any it closed.
Round 2 found that regression plus eleven class instances and stopped the loop. Round 3 found four defects and
folded one line of documentation. **The branch went to Aaron's review with four known defects listed rather than
eleven folded and unreviewed**, which is the trade §5 #11 exists to make — and his review then found four more
that nine reviewers had not, three of them on surfaces the rounds' own machinery was blind to. All eight are
dispositioned in the directed fold below.

## Aaron's review, and the directed fold

His review found **four** things nine fresh reviewers had not, three of them on surfaces the rounds' own
machinery could not see: a json wire shape the differential does not compare because it compares text lanes; a
demand-vs-render mismatch on a subject the no-stranding arm only pins in the everything-fits case; a set of
shared parameters no family validates on a call that selects the family none of them scope; and a retired-name
row that was absent, where absence reads as "kept" rather than "not decided". All four verified at source
before folding, **none refused** — and the four defects round 3 had listed were dispositioned in the same pass.

| finding | disposition | commit |
|---|---|---|
| **1** — `validate_scripts` `counts_only` json: `scan_errors` retyped from `{total, rows, rendered, truncated}` to a bare array | **FOLDED.** The 1.x wrapper is back and the merged surface's two honesty layers agree on one shape. No CHANGELOG bullet: with it restored the ancestor's json is additive against `main` and nothing is removed or retyped, so *"nothing you call today moves"* is true as written | `fbca062` |
| **2** — a child subject's demand counts units the render can never reach | **HONEST IN-BRANCH, FILED as #400** (Aaron's call). `BodyAllocation`'s NO-STRANDING paragraph names where the claim stops and cites the issue; the allocation is untouched. The response's own sentences were already true and none changes | `ce35d07` |
| **3** — `type=` / `formids=` / `exclude=` never validated on a dialogue-only call | **FOLDED**, with C1 as one seam: shared-input validation runs unconditionally before family dispatch, rendered through the normal refusal path so `format='json'` still gets a document | `9b90b71` |
| **4** — no retired-tool row for `housecarl_validate_dialogue`, and "Both ancestors" | **FOLDED.** The row names BOTH destinations — the sweep for classes 1–7, `records project=info_order` for class 8 — and the file records the retirement rather than leaving the absence to be re-derived | `94fcad6` |
| **A1** (HIGH, listed) | **FOLDED through the outcome value.** `DialogueKindChecks` says what one seed kind runs; the seed's verdict line reads it, `CheckOutcome` unions it as `DialogueOutcome.ChecksRun`, and the family boundary is composed from that — never a site-local `InputKind` patch, which is what the class generator's next instance would have been | `5c4ca62` |
| **B1/C2** (MED, listed) | **FOLDED and measured** — see the round-3 section for what the corrected predicate actually rejects | `4c87867` |
| **C1** (MED, listed) | **FOLDED** with finding 3 | `9b90b71` |
| **C3** (LOW, listed) | **DISPOSITIONED** — the finding moved under measurement, and the parameter text's promise got its bound rather than the code getting a refusal that would bury the real reason | `5065123` |

**The split that had to be preserved.** `exclude=` has two halves and only one moved up front. **Syntax** — a
value that is neither a plugin filename nor a known group — is malformed input for every family, so it refuses
whole-call. **Scope matching** — a filename you named that nothing in scope matches — stays family-local,
because each family's scope differs and raising it would refuse a call another family can answer. Both halves
are armed, and sabotaging the split in either direction reddens.

**Verification.** Seven commits, each with its arm, each RED-checked in both directions from a committed state.
One canaried sabotage sweep over all of them: **20 cells, 20/20 RED**, canary hand-proven RED before the script
existed. `ci-all` 125/125 in 1.94 min. `check-differential` re-run on the live order: **24 cells, 0
divergences** — unchanged, and the reason finding 1 slipped past it is that it compares text lanes, which is
why that fold ships a json wire cell instead.

**No round 4** — §5 #11's "no new rounds after directed folds". The merge go is Aaron's.

## The guard suites, before and after

| suite | on `main` | on this branch |
|---|---|---|
| `check-guard` | **does not exist** | **77 arms**, 31 matrix shapes |
| `check-errors-guard` | 136 | **137** |
| `script-property-check-guard` | 39 | **40** |
| `owned-child-content-guard` | 59 | 59 |
| `ci-all` | 125/125 | **125/125**, 1.94 min |

Nine of `check-guard`'s arms and one of `script-property-check-guard`'s are the directed fold's; the counts
before it were 68 and 39.

`check-guard` and `CheckMergeProbe.cs` are new here — the merged surface had no guard before this branch,
which is why the count is stated against zero rather than as growth. `check-guard` takes ~45 s of `ci-all`'s
1.88 min: the shape matrix sweeps every integer cap over 31 shapes in both transports.

## The harness chapter — three instances of one family, one session

Verification machinery **fails toward green**, and this branch hit that three times in three different ways.
All three are recorded because the fixes are now the contract the sweep scripts carry, not a habit.

1. **A sabotage sweep could not tell a CRASH from a PASS.** Its verdict scanned for `FAIL` lines, and a guard
   that throws prints none — so a cell whose sabotage made the guard throw was scored a pass and every arm
   after it was skipped. It cost three falsely-green cells before it was found. GREEN is now claimed only
   where the guard ran to `ALL PASS`; anything else is `CRASH` and gets triaged.
2. **A build failure was swallowed by a pipe.** A build was piped into `grep`, the pipe ate the compile error,
   and `check-guard` then ran the **stale binary** and reported `ALL PASS`. That pass was void. The rule is
   now mechanical: build to a file, check the exit code, THEN run the guard — and both sweep scripts check
   `returncode` as well as the output text. It fired again for real this session, benignly: a rebuild failed
   because a still-running probe held the binary locked, and the exit-code check caught it instead of the
   session reading a stale result.
3. **The guards themselves threw instead of failing.** `check-guard` read json fields with `GetProperty`, so a
   sabotage DELETING a field made the arm throw rather than fail — feeding instance 1. Fixed one arm at a time
   at first (four arms in one session), then **systemically**: all 27 remaining bare reads converted. The
   actual gap was not the leaf readers, which were already safe, but the NAVIGATION — these arms reach their
   leaves through `families.<token>.accounting`, so a sabotage deleting a family threw one level above the
   safe reader. `Obj` / `Arr` / `At` / `Strings` close that, with `JsonElement?` overloads so a whole chain is
   written without a step that can throw. Verified by a cell: a deleted field now reddens two arms instead of
   crashing the guard.

**The canary discipline that makes a sweep worth its tokens.** Every sweep carries a cell already proven RED by
hand in slot 0, and a sweep whose canary comes back green proved nothing and says so. Every sweep ran from a
committed state, with restores verified in-process by marker grep plus a clean-tree hard stop.

| sweep | cells | result |
|---|---|---|
| phase 3d, the migration's folds | 20 | 13 RED first pass; the 7 non-red re-run after the verdict fix and all dispositioned |
| phase 3d, the later stretch | 12 | 11 of 11 RED after three cells were re-aimed |
| phase 3e, this session's folds | 8 | **8/8 RED**, canary RED |
| phase 3e, the dialogue differential | 4 | **4/4 RED**, canary RED |

One cell in the last sweep had to be re-aimed **twice** before it could answer, and the reason each time is
recorded beside it: the first spelling attacked a property the disposition does not claim, and the second
exposed a loose check in the instrument rather than a defect in the code.

**One discipline breach, recorded because it is the record.** At phase 3c a sabotage was run from a DIRTY tree
— the arm work was uncommitted — and the restore hook blocked the `git checkout --` that would have discarded
it. The work survived only because copies had been taken first. §5 #11 says commit the fold, THEN sabotage;
that is the order, and it was not followed there. No breach since.

### The overrun notice keeps both branches, and the backstop's arm moved

The "a body unit overshot" branch is now **unreachable through any render**, because every emission site
declares a cost measured at its own depth and sibling position. Swept over caps 1–30,000 in both transports on
a fixture built to provoke it: **2,043 overruns fired and every one was a genuine cap-too-small**. The branch is
KEPT as a backstop — it is where a future unit declaring 0 lands, and the alternative sentence would lie to that
caller — and its arm moved from the render to the discriminator that computes it, RED-checked in each direction
separately. That disposition is a judgement call and is flagged as one.

### `seeds_validated` was renamed, and then stopped being an accounting field at all

Phase 3b renamed the accounting's `seeds_validated` to `seeds_reached`, because the number it carried was the
seeds the sweep REACHED and calling them validated claimed a completeness the same response denied.

The outcome-value migration then **superseded that rename**: the field is no longer in the accounting. The
family HEAD states each total once — `seeds_named`, `seeds_reached`, `seeds_validated`, `seeds_unreachable_total`
— and the accounting states only what the response rendered of them. A consumer reading
`accounting.seeds_validated` or `accounting.seeds_reached` finds neither; the four numbers are on the family
head, each meaning one population and only that one.

## Phase 3e, the tail

- **A4 — the json demand pass reserved a histogram frame the render never writes.** It added the frame cost for
  every axis while `WriteHistograms` reserves one only where the axis has rows, so a `counts_only` response had
  ~180 characters subtracted from its row budget for objects it never opened. The text lane was already
  symmetric. The fix is the gate; the arm is the one the codebase's own note said did not exist — `Reserve(` has
  exactly two callers, one per lane, so `demand.Reserved == body.ReserveDeclared` is a property of the real
  document rather than an argument about call sites. `RESERVE-DECLARED-IS-RESERVE-DEMANDED`, eight shapes, both
  transports, with populated controls beside the null-axis ones so a fixture that loses its null axis shows as
  the arm going quiet rather than green.
- **The `GetProperty` conversion** (harness chapter, item 3), done before round 3 so its reviewers' sabotages
  gave honest verdicts rather than crashes.
- **The band's diagnostic finished, and #398 filed** with the cause measured and decomposed by holder.
- **The differential's dialogue lane added**, and the Q3 defect it found fixed.

**Out of scope and named rather than left to be found:** `CheckErrorsProbe` and `ScriptPropertyCheckProbe` carry
the same bare-`GetProperty` read the merged probe was converted away from. Neither is touched here.

## Notes for the reviewer

- **One accounting guard is symmetric but currently unreachable, and its reason changed mid-branch.** A refused
  errors family declaring nothing cannot be fixtured: the errors family's scope is a superset of the scripts
  family's, so an `exclude=` that empties the errors scope empties the scripts scope too and every family
  refuses. The note used to say that is "a whole-call error". Under **grounds-are-one** (decision 17) that is
  true only while the two families' refusal strings are IDENTICAL — which they are wherever both can refuse,
  since both print `exclude= removed every plugin this sweep would have covered (N in scope, all excluded)` and
  the two Ns agree. The errors scope exceeds the scripts scope only via off-order files, and an off-order file
  left in scope means the errors family ANSWERS instead of refusing. So the fact holds and the reason is now the
  string equality, not the call shape. The scripts half is armed; the errors half is kept for symmetry and named
  here rather than left to be found.

- **The tree moved after round 3, once, and deliberately.** The advisor lane directed that #398 name WHICH
  reserve holder its 395 characters belong to, so `ReservedWritten` was split by write path and `--band` now
  prints held-against-written per holder (`673a333`). It touches a measurement instrument and three counters
  nothing in a response branches on. `ci-all` was re-run after it: 125/125. Recorded because a reviewer reading
  this PR should know which commits post-date the reviews — `673a333` and `a9a3640` do, and nothing else.


src/housecarl-mcp/JsonWire.cs
```diff
@@ -1488,121 +1793,176 @@ static void WriteSweepEpoch(Utf8JsonWriter w, ErrorCheckResult r)
...
+    static void WriteScriptsSection(Utf8JsonWriter w, ScriptCheckResult r, BoundedBody body, int histogramLimit)
+    {
+        var depths = new JsonUnitDepths(w.CurrentDepth);
+        if (r.CountsOnly)
+        {
+            WriteHistograms(w, body, histogramLimit, depths,
+                ("unbound_by_property", SweepSubject.HistogramByProperty, r.Histogram));
+            // The honesty layer, on its own subject and its own bound — a silently short list of what could NOT be
+            // read is the boundary of the answer going missing, not a finding inside it (#288 review finding 4).
+            w.WriteStartArray("scan_errors");
```
Avick3110: **`validate_scripts` json `counts_only=true`: `scan_errors` changed from an object to a bare array.**

Before this branch the counts_only lane wrote `scan_errors: {total, rows:[…], rendered, truncated}`. It now writes a bare array; `total`/`rendered` moved to `accounting.script_scan_errors_total` / `_named`, and the per-object truncation flag is gone.

Concrete: a consumer calling `housecarl_validate_scripts counts_only=true format='json'` and reading `result.scan_errors.rows` (or `.total`, or `.truncated`) now gets `undefined` — the value is an array.

Two things make this worth naming rather than leaving:

1. The sibling honesty layer one family over kept its wrapper — `WriteUnreadPlugins` still writes `unread: {total, rows, rendered, truncated}` — so the two now disagree in shape about the same kind of fact.
2. The PR body and the CHANGELOG both say *"All three older tools stay registered and unchanged, so nothing you call today moves."* That sentence is false for this response, and no CHANGELOG bullet covers the shape change (the two `validate_scripts` bullets cover the `max_chars` bound and the new accounting, not this).

src/housecarl-mcp/SweepDemand.cs
```diff
@@ -0,0 +1,346 @@
...
+                {
+                    if (t.Done(SweepSubject.UnreadRows)) break;
+                    t.Add(SweepSubject.UnreadRows, Wire.ComposeUnreadRow(p).Length);
+                }
+            }
+            else
+            {
+                t.Declare(SweepSubject.PluginSections);
+                t.Declare(SweepSubject.DanglingEntries);
+                foreach (var p in e.Reports)
```
Avick3110, line 91: **A child subject's demand counts units the render can never reach, so the fill hands it room it cannot spend.**

This loop measures `DanglingEntries` over the entries of *every* report, but `Wire.AppendErrorsSection` breaks the **outer** loop when a section is refused:

```csharp
var section = ComposeErrorSection(p);
if (!body.Emit(SweepSubject.PluginSections, section.Length, …)) break;   // ← whole loop
foreach (var d in p.Dangling) { … }                                      // inner break only
```

So once `PluginSections` hits its ceiling, no further plugin's entries are reachable — but `DanglingEntries` was already allocated against all of them.

Concrete: 40 plugins, each with a ~2,000-char section fixed part (long missing-master lists) and 2 dangling entries (~200 chars total). Demand: PluginSections = 80,000, DanglingEntries = 8,000. Row budget 20,000 → water-filling gives DanglingEntries its whole 8,000 and PluginSections 12,000. The render emits 6 sections and their 6×200 = 1,200 chars of entries. Body lands at ~13,200 of a 20,000 row budget, the response reports `truncated` and prints *"Raise max_chars= to fit more of what was found"* over ~6,800 unspent characters — about a third of the row budget.

`BodyAllocation`'s summary states **"NO STRANDING … A merged call that could have shown everything does"**, and `ALLOCATION-NO-STRANDING` only pins the everything-fits case, so nothing in the guard reaches this. Monotonicity is unaffected (λ is still non-decreasing), so it is under-fill rather than a false remedy — but it is the class of thing #394 exists to remove.

Same shape in three other places:
- `ForJson` (line 241) — identical nesting.
- `ForText` line 139 / `ForJson` line 303 — `DialogueTopics` is measured across every resolved seed, while `DialogueSweepRender.AppendSection` breaks the **seed** loop when a seed head is refused.

src/housecarl-mcp/CheckTools.cs
```diff
@@ -0,0 +1,188 @@
...
+            string[]? seeds = null,
+        [Description("Optional. Max characters before the response stops with an explicit notice. 0 = the server default (~80k). The budget is DIVIDED among the families that ran and their parts, not spent in series — a family that renders second does not inherit what the first one left over. Raise it for a quest that owns many topics.")]
+            int max_chars = 0) => Guard.Tool("housecarl_check", () =>
+    {
+        if (svc.ConfigPromptOrNull() is { } prompt) return prompt;
+        bool json = Wire.WantsJson(format, out var fmtErr);
+        if (fmtErr is not null) return fmtErr;
+        if (!SweepFamilySelection.TryParse(findings, out var selection, out var famErr)) return "error: " + famErr;
+        int lim = limit <= 0 ? 1000 : limit;
```
Avick3110, line 155: **On a dialogue-only call, `type=` / `formids=` / `editorid_contains=` / `exclude=` are never validated — a malformed value is silently accepted.**

`BuildSweepScope` and `SweepExclusion.Resolve` live inside `svc.CheckErrors` and `svc.ValidateScripts`, and both calls below are conditional on their family being selected. With `findings=["dialogue"]` neither runs, so nothing ever looks at those four parameters.

Concrete:

```
housecarl_check findings=["dialogue"] seeds=["0F1AC1:Skyrim.esm"] \
  type="NOTATYPE" formids=["not-a-formid"] exclude=["NoSuchPlugin.esp"]
```

returns an ordinary dialogue answer, with nothing anywhere saying the three values were bad. The tool's own parameter text promises the opposite in all three cases:

- `type=` — *"An unknown type is refused, naming what is expected."*
- `formids=` — *"A malformed token refuses the call before the sweep runs."*
- `exclude=` — *"A value that is neither is refused before the sweep runs, and so is a FILENAME YOU NAMED that nothing in scope matches."*

The dialogue scope note does say these parameters do not **scope** that family, which is a different claim from *they were checked and are fine* — a caller who typo'd `type=` reads a narrowing they never got as one that was accepted (Q3). Parsing them up front, before the family dispatch, would refuse at the same point every other lane does.

src/housecarl-mcp/AliasTable.cs
```diff
@@ -303,6 +303,15 @@ static readonly (string Old, string Successor)[] RetiredTools =
...
         ("housecarl_remove_record",
          "absorbed into housecarl_remove: formids= is SET-VALUED — drop many records in one re-serialize (one is a set of one). The houseCARL-patch lane is into=\"MyPatch.esp\" (removal edits an artifact that EXISTS; patch= names a NEW one everywhere else on the surface), and the target=+in_place=true pair is in_place=\"X.esp\"."),
+        // 4b — the derived-findings sweep. Both ancestors stay REGISTERED through the build waves (the W2/W3
+        // precedent), so these rows are dormant by construction: a registered name resolves and never reaches the
+        // retired-name check. They activate at the 2.0.0 clean cut, and nothing in a response mentions them until
+        // then — no deprecation prose, in either tool.
+        ("housecarl_check_errors",
+         "absorbed into housecarl_check: the same sweep is findings=[\"errors\"], which is also the DEFAULT when findings= is omitted. type=/formids=/editorid_contains=/exclude=/counts_only=/limit=/max_chars=/format= are unchanged; the response is sectioned per family and states which families it did not run."),
+        ("housecarl_validate_scripts",
```
Avick3110, line 320: **No retired-tool row for `housecarl_validate_dialogue`, and the comment above says "Both ancestors".**

The merge absorbs three ancestors; two rows were added. At the 2.0.0 clean cut this table is what turns an old name into successor guidance, so a caller of `housecarl_validate_dialogue` would get nothing while `housecarl_check_errors` and `housecarl_validate_scripts` callers get a pointer.

If `housecarl_validate_dialogue` is meant to survive past 2.0.0 — it still renders class 8, the effective merged INFO order, which the merged surface deliberately omits — then the omission is right and only the word "Both" is off. Either way nothing in the file records which it is, and the next session reading "Both ancestors" alongside a three-ancestor merge has to re-derive it.

Avick3110:
## The directed fold — your four findings, and the four defects the branch was listing

Every finding verified at source before folding; **nothing refused** — all eight reproduced. Each fold is its
own commit, each arm RED-checked in both directions from a committed state, and the whole set swept with a
hand-proven canary in slot 0. The listed-defects table in the body is updated with the real dispositions.

### Your four

| # | finding | disposition | arm | commit |
|---|---|---|---|---|
| 1 | `validate_scripts` `counts_only` json: `scan_errors` retyped from `{total, rows, rendered, truncated}` to a bare array | **FOLDED** — the 1.x wrapper is back, and the merged surface's two honesty layers (`unread`, `scan_errors`) are now one shape. Wrapper rather than array, because that is what the wrapper is for: a bare array a budget cut cannot say it was cut | `COUNTS-ONLY-JSON-SCAN-ERRORS-WRAPPED` (ancestor wire cell, flags asked at every cap in a band) · `MATRIX-HONESTY-LAYERS-AGREE` (both layers, every cap of every matrix shape) | `fbca062` |
| 2 | `SweepDemand` measures a child subject's demand over units the render can never reach → the fill hands it room it cannot spend | **HONEST IN-BRANCH, FILED as #400** — your call. `BodyAllocation`'s NO-STRANDING paragraph now names where the claim stops and cites the issue; the allocation is untouched. Your measurement, the four sites, the two-level-fill candidate and your promotion condition are all on the issue | none — docs-only, no behaviour to redden | `ce35d07` |
| 3 | on a dialogue-only call, `type=` / `formids=` / `exclude=` are never validated | **FOLDED** — shared-input validation runs unconditionally before family dispatch, and the refusal renders through the normal path so `format='json'` still gets a document | `ORCH-SHARED-INPUT-IS-CHECKED-BEFORE-FAMILY-DISPATCH` | `9b90b71` |
| 4 | no retired-tool row for `housecarl_validate_dialogue`, and "Both ancestors" | **FOLDED** — the row is there and names BOTH destinations (the sweep for classes 1–7, `records project=info_order` for class 8), the comment says three, and the file now records the disposition rather than leaving the absence to read as "kept" | `ORCH-EVERY-ABSORBED-ANCESTOR-HAS-A-RETIRED-NAME-ROW`, asked off `SweepFamilySelection.Registered` so a fourth family with no row reddens it | `94fcad6` |

### The four the branch was listing

| # | disposition | arm | commit |
|---|---|---|---|
| **A1** (HIGH) | **FOLDED, through the outcome value.** `DialogueKindChecks` says what one seed kind runs; the seed's verdict line reads it, `CheckOutcome` unions it as `DialogueOutcome.ChecksRun`, and the family boundary is composed from that. A passing DLVW/DLBR states its own CK-parity verdict, and where every seed reached owns no INFO list the boundary names none of the graph checks. json seeds gain `checks_run` | `DIALOGUE-A-RECORD-LEVEL-SEED-STATES-ITS-VERDICT` · `DIALOGUE-THE-BOUNDARY-CLAIMS-ONLY-THE-CHECKS-THAT-RAN` · `DIALOGUE-THE-KIND-AWARE-VERDICT-AND-BOUNDARY-ARE-THE-SAME-IN-JSON` | `5c4ca62` |
| **B1/C2** (MED) | **FOLDED, and measured.** Marker corrected at all three sites. Instrumented the tight-fit search and ran the matrix with both markers: of 62 shape/lane cells, **3 moved, each by exactly 219 characters** — all inside the 2,200 `TightSlack`, `MATRIX-NO-STRANDING` green throughout. So no text-lane stranding gap is exposed; what the dead marker cost was the arm's resolution | `MATRIX-EVERY-FINGERPRINT-TERM-IS-REACHED` — the rule rather than the correction. It immediately found two more of the same class, both mechanical, both folded here | `4c87867` |
| **C1** (MED) | **FOLDED** with R3, one seam. `plugins=['  ']` refuses instead of widening the scope to the whole order | `ORCH-A-BLANK-PLUGIN-NAME-REFUSES-RATHER-THAN-SWEEPING-THE-ORDER` | `9b90b71` |
| **C3** (LOW) | **DISPOSITIONED, and the finding moved under measurement.** The SYNTAX half already refused on that path — the scripts family's service entry resolves `exclude=` before it decides anything about scope — so what C3 reaches is the SCOPE-MATCHING half, which the fold directive rules stays family-local. Making it fire there makes the response worse: with every named plugin off-order the family swept nothing *because the files are outside the order*, and refusing over the exclusion buries the real reason. So the parameter text's promise got its bound instead of the code getting a refusal | `ORCH-AN-EMPTY-BY-CONSTRUCTION-SCOPE-REPORTS-ITSELF-RATHER-THAN-BLAMING-EXCLUDE` (re-aimed — the first spelling asserted a refusal the family already made, and sabotaging the up-front check left it green) | `5065123` |

### The split that had to be preserved

`exclude=` has two halves and only one moved. **Syntax** — a value that is neither a plugin filename nor a
known group — is malformed input for every family, so it refuses up front, whole-call. **Scope matching** — a
filename you named that nothing in scope matches — stays family-local, because each family's scope differs and
raising it would refuse a call another family can answer. Both halves are armed, and sabotaging the split in
either direction reddens.

### Verification

- **Sabotage sweep**: 20 cells, **20/20 RED**, canary hand-proven RED before the script existed
  (`scan_errors` back to a bare array → `1 FAILURE(S)`). Every fold's arm sabotaged in both directions;
  in-process restore verification, clean-tree hard stop, crash-is-not-green, build exit-code check.
- **Guards**: `check-guard` **77 arms** / 31 matrix shapes (68 before the fold) · `check-errors-guard` 137 ·
  `script-property-check-guard` **40** (39 before) · `owned-child-content-guard` 59 — all ALL PASS.
- **`ci-all`**: 125/125 in 1.94 min.
- **`check-differential`**: 24 cells, 0 divergences on the live order — unchanged. Finding 1 slipped past it
  because it compares the ancestor against the merged surface, and `scan_errors` has ONE shared writer, so a
  shape change moved both sides together. That is why the fold ships a wire cell rather than a differential lane.

### CHANGELOG

Two new bullets (the shared-input refusals; the kind-aware dialogue verdict and boundary). Finding 1 gets
none: with the wrapper restored the ancestor's json is additive against `main` — `records_with_findings`, the
`accounting` object, `max_chars_overrun` — and nothing is removed or retyped, so *"nothing you call today
moves"* is true as written.

No round 4 — §5 #11's "no new rounds after directed folds". The merge go is yours.

Issue #424, opened 2026-08-23
[Bug] GetterInterfaceFor resolves a getter interface by name only — generic arity and not-own-named shapes return null for extractable types


Avick3110:
Dev-noticed during the #330 Mutagen-bump engagement's pre-PR review rounds, so filed as prose rather than through the bug form. Split out of #397 deliberately: #397's fix is behaviour-neutral and shipped, this is the underlying defect it exposed.

## What

`CorpusGenerator.GetterInterfaceFor` resolves a concrete class's getter interface by NAME:

```csharp
var direct = concrete.Assembly.GetType($"{concrete.Namespace}.I{concrete.Name}Getter");
if (direct != null) return direct;
return concrete.GetInterfaces()
    .FirstOrDefault(i => i.Name == $"I{concrete.Name}Getter")
    ?? concrete.GetInterfaces().FirstOrDefault(i => i.Name.EndsWith("Getter") && Normalize(i.Name) == Normalize(concrete.Name));
```

Two shapes defeat it, and it returns `null` for a type that is perfectly extractable.

**1. Generic arity.** `Type.Name` for a generic class carries the arity suffix — `GenderedItem\`1` — so the probe name is the impossible `IGenderedItem\`1Getter`. Both `GetInterfaces()` fallbacks fail for the same reason: the real interface is `IGenderedItemGetter\`1`, whose `Name` does **not** `EndsWith("Getter")`.

`MutableInterfaceFor` fixes exactly this trap, with a comment naming `IGenderedItem<T>`:

```csharp
// Strip the arity suffix before the Getter check: a generic getter interface is
// named "IFooGetter`1", which does NOT end in "Getter". Missing this silently
// dropped the mutable interface for every generic type (e.g. IGenderedItem<T>) …
```

`GetterInterfaceFor` never got the same fix. In-repo precedent for both the defect and the remedy.

**2. Implemented but not own-named.** A concrete class can implement a getter interface that is not named after it. `ArmorAddonWeightSliderContainer` implements `IGenderedItemGetter<bool>`; the probe only ever looks for `IArmorAddonWeightSliderContainerGetter`, so it resolves `null` — even though the type's data is fully catalogued today as `GenderedItem<Boolean>`, 2/2 writable, fields `Female` and `Male`.

## Measured

Five types resolve `null` while demonstrably implementing a getter interface:

| type | implements |
|---|---|
| `GenderedItem\`1` | `IGenderedItemGetter\`1` |
[...]

Issue #423, opened 2026-08-23
THIRD-PARTY-NOTICES is out of date against the publish output


Avick3110:
Dev-noticed during the #330 Mutagen-bump engagement's pre-PR review rounds, so filed as prose rather than through the gap form.

> **Rewritten 2026-08-23, same day, before any routing stroke.** The first version framed this as *version* drift and proposed a guard that diffs the versions of listed entries. Review round 2 then found four components missing from the file entirely — which that guard, as specified, would not have caught. The class is **divergence from the publish output: set membership AND versions.** Corrected in place rather than split across a second issue.

## The gap

`THIRD-PARTY-NOTICES.txt` (tracked in two places — repo root and `plugin/`, kept byte-identical) says it lists what the plugin bundles in `server/`, with an exact version for each component, plus the **GPLv3 §6 corresponding-source** release and commit for the Mutagen libraries.

Nothing derives or checks any of that against what actually ships. A dependency bump that does not hand-sweep the file leaves false claims — or silently omits a component altogether — in a licence-compliance artifact, with CI fully green.

Same defect class as #335/#351 — an artifact that disagrees with what the code actually does, reachable by omission rather than by a bug — on the compliance surface.

## Three evidence instances

**1 — PR #366** (`ModelContextProtocol.AspNetCore` 1.3.0 → 2.2.0, merged 2026-08-23 as `aeb53d6`) left **18** entries with stale versions, unnoticed at the time.

**2 — the #330 branch** (`Mutagen.Bethesda.Skyrim` 0.53.1 → 0.54.4) made a further **22** versions wrong — every GPL-3.0 entry among them — and invalidated the corresponding-source line, which named *release 0.53.1; commit `bdbb6ff6…`* while shipping 0.54.4. Found by a review agent, not by a guard.

**3 — the same branch, set drift** (the instance that corrected this issue's scope): four components the new graph ships were named nowhere — `System.Interactive.Async`, `System.Linq.AsyncEnumerable`, `System.Text.Encoding.CodePages`, `TransparentValueObjects.Abstractions`. Found only by publishing and diffing DLLs against the file. The last of these declares **no licence at all** in its NuGet metadata (determined MIT by other means, recorded in the file).

[...]

PR #425, opened 2026-08-23
deps: Mutagen.Bethesda.Skyrim 0.53.1 -> 0.54.4, with the regenerated reference, an emit-match guard (#351) and the #397 classifier split

### LINKED ISSUE #351: [GAP] CI doesn't check the committed mutagen-reference shards against a fresh emit

### LINKED ISSUE #397: [Bug] Corpus generator conflates "no getter interface" with "read-only projection" — and FindUnionArms drops such a type silently
Dev-noticed (not a tool call), so filed as prose rather than through the bug form.

## What

`CorpusGenerator.IsAuthorableArm` is:

```csharp
static bool IsAuthorableArm(Type t) => MutableInterfaceFor(GetterInterfaceFor(t) ?? t) != null;
```

When a concrete class has no `I<Name>Getter`, `GetterInterfaceFor` returns null and the `?? t` fallback hands the **class itself** to `MutableInterfaceFor`, which tests `EndsWith("Getter")`, fails, and returns null.

So **"has no getter interface" and "is a read-only projection" produce the identical answer**, even though only the second is a correct reason to exclude a type.

## Two consequences, different severity

- **Record path** — `BuildCorpus` warns (`src/housecarl-generator/CorpusGenerator.cs:61`, *"No getter interface resolved for record class X"*) and skips. Surfaced, at least.
- **Arm path** — `FindUnionArms` (`CorpusGenerator.cs:497`) filters through `IsAuthorableArm` with **no warning of any kind**. An authorable arm can disappear from a polymorphic base with a green exit code and nothing printed. That is the Q3 half of this, and the part worth fixing regardless of anything below.

`IsAuthorableArm` itself is at `CorpusGenerator.cs:521`.

## Status on Skyrim: latent, not live

Measured against `Mutagen.Bethesda.Skyrim 0.53.1`: **0 of 133** concrete non-overlay major-record classes lack a getter interface. Nothing trips this today. It is an unarmed tripwire, not current data loss — a future Mutagen bump that adds such a class is what arms it.

## Demonstrated live on another game

Reported from an independent port of the corpus generator to `Mutagen.Bethesda.Starfield 0.54.4`, where `DeletedObjectModification` — **26 of 27 properties writable**, base `AObjectModification` — has no getter interface, is absent from the catalogue entirely, and is **not** among the seven arms `AObjectModification` lists.

Verified here rather than taken on trust: 407 concrete major-record classes, 202 `*BinaryOverlay` twins, 205 real candidates. Three lack a getter interface. The other two — `MergedWorldspace` (0/44 writable) and `MergedWorldspaceCell` (0/49) — are genuine read-only projections and are correctly excluded. Only the writable one is wrongly excluded, and it is excluded for the wrong reason.

## Fix

Two parts:

1. **A distinct anomaly class — "writable but unextractable"** — separate from read-only-projection in the emitted anomaly list. The point isn't the excluded type; it's that today the two cases are indistinguishable in the output, so a real coverage gap reads as correct exclusion.
2. **Warn on the arm path too.** A silent drop is the actual defect.

Deliberately **not** doing: extracting the schema from the concrete class when no getter interface exists. That would make one entry's writability computed differently from every other entry in the catalogue, and a catalogue whose entries don't mean the same thing is worse than one with a named hole in it.

Also being reported upstream to Mutagen — every other `AObjectModification` subclass has a getter interface, so this reads like an upstream oversight rather than a decision. If that's right the Starfield gap closes at source, but both fixes above are still correct here.

## Provenance

Reported by **SparkleDemon**, from an independent port of the corpus generator to Starfield tooling. Credited under the handle at their request. All figures above re-derived locally against houseCARL source and live reflection.

## Scheduling

Not urgent (latent on Skyrim, per above). Queued behind the in-flight 2.0 work.

Avick3110:
Bumps `Mutagen.Bethesda.Skyrim` 0.53.1 → 0.54.4 across all three projects, regenerates the bundled `mutagen-reference` against it, and carries two guard-shaped issues that the bump is the natural occasion for.

Fixes #397
Fixes #351

Supersedes the Dependabot PR for #330 (closed), which bumped the pin without regenerating the reference the pin describes.

## What is in it

**The bump and the reference.** Three csproj pins move together, and the four committed reference shards are re-emitted from the new assembly. The corpus delta is **zero removals of any kind** — no type, field, enum value or union arm disappeared, and no field that was writable became read-only. What changed: four record shapes are described differently (`LandscapeVertexHeightMap.HeightMap` is now a signed 8-bit map; `Race.BehaviorGraph` and `BodyData.Model` name their own model types; `FindMatchingRefFromEvent.EventData` reads as a typed enum instead of raw bytes) and three enums gain values. That is the CHANGELOG's `## Unreleased` entry, and it is the claim a reader is most likely to act on, so it was derived from the shard diff rather than from release notes.

**#351 — the shards are now checked against a fresh emit in CI.** `emit-match-guard` re-emits the corpus and diffs it against what is committed, so a future bump cannot land a pin without the reference that describes it. This PR is the case that motivated it.

**#397 — the generator tells a writable-but-unextractable type from a read-only projection.** `ClassifyArm` splits the two reasons a candidate is excluded from a polymorphic union; only one of them is a correct silent exclusion. `arm-classification-guard` pins the split, including its behaviour-neutrality over the assemblies the walk actually visits (derived from the walk, not listed).

**Third-party notices.** 22 entries the bump moved were updated, including every GPL-3.0 entry and the GPLv3 §6 corresponding-source release/commit; 18 further entries were already stale at `aeb53d6` from the MCP SDK bump and were swept here on the ruling below. Four components the 0.54.4 graph newly ships were added, and `miniball` (Apache-2.0, pre-existing, attributed nowhere) is now attributed. **Verified against an actual publish** — the exact `scripts/build-plugin.ps1:81` invocation, 56 DLLs — set *and* versions, which is the check earlier passes were missing because they diffed versions of listed entries and never checked the set.

## Pre-PR review rounds (§11)

Three rounds, fresh agents each round, each in its own worktree. The §11 cap is spent; there is no round four.

| round | found | folded |
|---|---|---|
| 1 | behaviour neutrality was not neutral (`ClassifyArm` dropped the original predicate's `?? t` fallback, flipping two `Mutagen.Bethesda.Core` types); `emit-match-guard` crashed instead of reporting on a BOM-only difference; `pkcu-regression` could not tell which exclusion path fired; the version claims the bump did not sweep | all, in `0f92939` |
| 2 | notices omit four newly-shipping components (HIGH); `miniball` Apache-2.0 unattributed (MEDIUM, pre-existing); the anomaly *emission* is unguarded — both call sites deletable with every guard green (MEDIUM); guard A's "both assemblies" is two of three (LOW) | **nothing** — §11 class-recurrence stop called, escalated |
| 3a | Apache licence-text heading, four repudiated comments, D's vacuity floor | `de3cb48` |
| 3b | five findings, two of them inside the very commit whose purpose was removing the recurring class | **nothing** — the class-recurrence trigger fired; escalated |

**The recurring class was hand-authored epistemic prose** — sentences asserting more than the thing they describe establishes. It appeared in every round, including inside each fix for it. That is the §4 escalation the §11 condition anticipates, so round 3b was not folded by the reviewing session.

**Refusal rate was 0% across all three rounds.** §11 requires a round folded at 100% to say why nothing was refused: every finding was independently reproduced against source before being folded or carried — the `GenderedItem<Boolean>` lookup, the two-assembly neutrality sweep, the live BOM crash, the `Version="*"` restore, both `Exclude()` call sites, the publish DLL set, the unreachable return, the Noggog getters. Nothing was taken on a reviewer's word, and nothing survived triage that could not be shown in the code. It is flagged rather than passed over, because a 0% refusal rate is normally a triage-health warning.

## Escalations and rulings

Three went to the advisor lane; all three were answered and executed.

1. **The anomaly text's semantics** (what the #397 line may claim). Ruled: the line states its measurement and stops there — it may not diagnose a real coverage gap, and may not assign the fix upstream to Mutagen. The second was provably false for the guard's own exhibit, whose data ships today as `GenderedItem<Boolean>`, 2/2 authorable. `2557d0c`, with the absence of both sentences asserted so they cannot come back.
2. **Notices scope** — whether 18 pre-existing stale entries and a missing component set were this branch's to fix. Ruled fix-in-branch, a fortiori for the four components this branch causes to ship. `45fb24e`, `63cc64e`.
3. **The claim class itself.** Ruled the §4 escalation the trigger anticipated; the ruling supplied replacement sentences to be applied mechanically rather than re-authored, and directed the remaining round-3b findings as a single fold with no further rounds. `a5a04ac`.

## The directed fold (`a5a04ac`)

- **The label names its predicate.** `WritableSurfaceCount` is narrower than `CanWrite` and wider than public-settable; the line now says "N of M properties are authorable (public settable, or a mutable collection)". C3 pins the label *with* its predicate.
- **Both arms close on one shared const.** The zero-count arm's "so nothing is lost by excluding it" inferred reachability from writability and was deleted. C8 asserts the shared `ReachabilityCheck` present in **both** arms, so they cannot diverge again.
- **Coverage gaps get their own output channel**, printed first and in full, never sharing the unbounded warning stream's cap. The stream keeps its cap and says "+N more suppressed". D1 reads the untruncated channel, which is also what makes its failure text honest.
- **`pkcu-regression` asserts the exclusion path as a fact**, not as a substring of user-facing prose: `Unopenable` is the resolver's own machine-readable split of the two exclusion paths, kept as its own set by #314 for exactly this reason. The sentence is now free to change.
- **Two hand-written universes removed from the guard**: `Find()` resolves exhibits by `FullName` (~350 nested types share one simple name), and D1's expected set derives from the same enumeration the record-path emission walks.

**Sabotage RED-checks, all from the committed state, all restored and the restores grep-verified:** C3's label pin → red on removing the predicate; C8's both-arms pin → red on re-diverging one arm; the section split → a 100-warning flood leaves the coverage lines intact and D1 green, while the same flood under the pre-fold arrangement (coverage lines back in the capped stream) turns D1 red with a section still printing, the known-red canary proving the fixture bites; `pkcu-regression` → red on inverting the path predicate.

## Also observed, not acted on

`Version="*"` on these packages restores **23.4.0**, not 0.54.4. That strengthens the mis-numbered-push hypothesis for what Dependabot hit, and is recorded here rather than worked around.

Two issues were filed from this branch's review and are deliberately unrouted, awaiting the routing stroke: **#423** (notices diverge from the publish output — rewritten in place after round 2 showed the original version-drift framing would not have caught the missing-component instance) and **#424** (`GetterInterfaceFor` resolves by name only — the defect behind #397's misdiagnoses; fixing it changes emitted corpus output, so it wants the fix-before-regenerate pattern at a future bump).

## Verification

`ci-all` **127/127** (two new guards), `freshness-capture-guard` cold PASS, Release build 0 errors, at `a5a04ac`.


Avick3110:
## Fold spot-check verdict (`a5a04ac`)

Standing lane watch. This session executed the ruling's directed fold; it did not review the branch and did not run a round four.

**Verdict: the fold applies the ruling as supplied, and every RED-check fired in the expected direction.** Details, so a reader can check rather than take it:

| item | landed as | RED-check |
|---|---|---|
| A — label states its predicate | `CorpusGenerator.cs` label + `WritableSurfaceCount` docstring naming both directions; C3 pins label-with-predicate | removing the predicate from the label → **C3 red, nothing else** |
| B — one shared closing clause, both arms | `ReachabilityCheck` const; C8 asserts it in **both** arms; C6 reads the const too, so the guard holds no second copy | re-diverging the zero arm onto its own closing text → **C8 red, nothing else** |
| C — coverage lines get their own untruncated section | `CoverageAnomalies`, printed first and in full; warning stream keeps its cap and says "+N more suppressed"; D reads the coverage channel | 100-warning flood → coverage lines survive, D1 green. Same flood with the **pre-fold** arrangement (coverage lines back in the capped stream) → **D1 red with a section still printing** — the known-red canary, and the exact silent-truncation failure the split removes |
| D — pkcu path assertion off prose | reads `resolver.IsUnopenable`, the resolver's own machine-readable split | inverting the path predicate → **pkcu-regression red** |
| E — two hand-written universes | `Find()` by `FullName`; D1's expected set from the record path's own enumeration | covered by the guard's own exhibit resolution; expected count unchanged at 2, so the fix is shape, not answer |

All sabotage ran from the committed state and each restore was verified by grepping for a string the fold introduced, per §11.

**Two notes that are mine, not the ruling's, and are the parts most worth a second pair of eyes:**

1. **Item D was landed on an existing surface rather than a new one.** The ruling said "an exclusion-path enum (or shared const) threaded from `LoadOrderResolver`". `LoadOrderResolver` already threads exactly that: `Unopenable`, added by #314, whose own docstring states the ruling's principle verbatim — *"a message is display prose that can be reworded, membership is a fact."* There are exactly two `Exclude()` call sites and `unopenable.Add` fires only at the open-time one, so excluded-and-not-unopenable **is** the mid-enumeration path. I took the existing surface over adding a parallel enum, because a new enum would duplicate `Unopenable` and that is the ballooning the kickoff's stop rule names. If the lane wanted the enum specifically, this is the line to push back on.

2. **The ruling quotes the label two ways** — the general form with the parenthetical, and the `w == 0` arm's form without it. I applied it **with** the parenthetical on both arms, because a label that differs between the arms is precisely the divergence item B's shared const exists to end. Recorded in the commit message rather than resolved silently.

Nothing else rode this fold. `ci-all` 127/127, `freshness-capture-guard` cold PASS, Release build 0 errors, tree clean at `a5a04ac`.

Avick3110:
**Advisor spot-check (pre-review, 2026-08-23):** the directed fold (a5a04ac) verified at source. Both flagged deviations ENDORSED after verification, not on report: (1) pkcu-regression asserting `resolver.IsUnopenable` instead of the ruled enum — confirmed at LoadOrderResolver.cs: exactly one `unopenable.Add` at the open-time Exclude, two Exclude sites total, so excluded-and-not-unopenable is provably the mid-enumeration path; reusing the existing #314 machine-readable split beats minting a parallel enum (one home for one fact). (2) The authorable-label parenthetical applied to BOTH anomaly arms — consistent with the shared-const item's purpose; a label differing between arms is the divergence that item exists to end. RED-check evidence (C3/C8/flood-canary/pkcu inversion, restores grep-verified) accepted as reported. Aaron's independent review remains the gate.

src/housecarl-generator/ArmClassificationProbe.cs
```diff
@@ -0,0 +1,309 @@
...
+        // Drive a real emit FIRST. Two things depend on it and neither can be faked: the walk has to have run
+        // before AssembliesWalked describes anything (check A's universe is derived from it), and section D
+        // reads what the generator actually reported. Doing it here rather than at D means A cannot pass
+        // vacuously over an empty universe — which it did, once, until A0 caught it.
+        var reported = EmitAndCaptureCoverageAnomalies(out var anomalyCount);
+
+        // ---------------------------------------------------------------- A: behavior neutrality
+        // The predicate as it stood before #397, reproduced literally.
+        static bool OriginalPredicate(Type t) =>
+            CorpusGenerator.MutableInterfaceFor(CorpusGenerator.GetterInterfaceFor(t) ?? t) != null;
```
Avick3110: **Check A is tautological — it cannot detect the neutrality break it is stated to protect.**

`OriginalPredicate` is a character-for-character copy of the expression `ClassifyArm` evaluates on its first line:

```
OriginalPredicate(t) => MutableInterfaceFor(GetterInterfaceFor(t) ?? t) != null
ClassifyArm(t)       => var gi = GetterInterfaceFor(t);
                        if (MutableInterfaceFor(gi ?? t) != null) return Authorable;
```

So `disagreements` is empty by construction — A compares an expression with itself.

Concrete scenario: fix #424 (`GetterInterfaceFor` resolves by name only). Every type whose getter interface is generic or differently named flips from `ReadOnlyProjection`/`WritableButUnextractable` to `Authorable` — a real change to the accepted-arm set, which is exactly what the header says A guards ("which is what makes the shard-level byte-identity a consequence rather than a coincidence"). A stays green through it, because both sides of the comparison move together on the same two helpers.

The shard-level byte-identity is in fact established by `emit-match-guard`, not by A. What A does pin is narrower: "ClassifyArm's first line still reads this way". That pin costs a 12,993-type × 3-assembly reflection sweep per CI run, and each non-authorable candidate in it also runs a full `WritableSurfaceCount` inheritance walk.

If the intent is a real neutrality pin, the predicate has to be reproduced without going back through the same helpers (or the claim narrowed to what it measures).

src/housecarl-generator/ArmClassificationProbe.cs
```diff
@@ -0,0 +1,309 @@
...
+        var tmp = Path.Combine(Path.GetTempPath(), "housecarl-arm-emit-" + Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(tmp);
+        var captured = new StringWriter();
+        var realOut = Console.Out;
+        try
+        {
+            Console.SetOut(captured);
+            CorpusGenerator.GenerateAll(Path.Combine(tmp, "generated"), Path.Combine(tmp, "refs"));
+        }
+        finally
```
Avick3110, line 280: **The captured generator output is discarded when the emit throws.**

`Console.Out` is redirected into the local `captured` writer; if `GenerateAll` throws, the `finally` restores the console and the exception propagates — and `captured` goes out of scope with everything the generator had printed up to the failure.

Concrete scenario: a Mutagen bump (the event #397 says arms this class) makes `BuildCorpus` throw mid-walk. `arm-classification-guard` sits before `emit-match-guard` in `CiAll.Probes`, so it is the first probe to hit it. CiAll catches per-probe and reports the exception alone; the assembly banner, per-kind counts, spotlights and any anomalies already accumulated are gone.

`EmitMatchProbe.RunGuard` — added in this same PR — handles the identical situation deliberately: catch, replay `captured.ToString()` to stderr because "it is the diagnosis rather than the noise", rethrow. Worth mirroring here:

```suggestion
        catch
        {
            Console.SetOut(realOut);
            Console.Error.WriteLine("  the generator threw during this guard's emit — its output up to the failure follows:");
            Console.Error.WriteLine(captured.ToString());
            throw;
        }
        finally
```

Avick3110:
## Review folded — 2/2, 0 refused (`d02b760`)

Both findings verified at source before folding. Zero refusals is honest here rather than obedient: both cite lines and reproduce on their face — F1's tautology is visible with the two expressions side by side, and F2's discard is a bare `try/finally` with no catch. There was nothing to push back on.

Per §11, no new rounds after a directed fold. This is the fold, not a fourth round.

### F1 — check A deleted, not re-spelled

Confirmed exactly as described: `OriginalPredicate` is the same expression `ClassifyArm` evaluates on its first line, so `disagreements` is empty by construction, and the #424 scenario walks straight through it.

Taking the **delete** branch rather than the independently-spelled twin you offered as the alternative. The ground, stated so it can be argued with: reproducing that predicate without going back through `MutableInterfaceFor`/`GetterInterfaceFor` means hand-maintaining a copy of two non-trivial reflection routines — arity-stripping, the `?? t` fallback, the interface-scan fallbacks. That is a drift generator, not a pin. The wire-names lesson tolerates a hand-written twin for **flat literal lists**; these are not that. #339 governs: an arm that cannot fail is deleted.

Its two claims are re-homed, not dropped:

- **The transition claim** (the #397 split changed no classification) was proven once, on this branch, by round 1's fold and its RED checks. Historical — nothing standing needs to re-prove it.
- **The standing claim** (no classification change ships silently) is held **by construction** by `emit-match-guard` and the #351 CI emit-match step added in this same PR. A classification change alters emission; emission is diffed against the **committed** shards on every CI run; a difference fails the build — including when nobody regenerates. You are right that the byte-identity was established there and not by A; the header now says so.

Execution details, since the delete had a precondition:

- **Consumer check ran before deleting.** `candidates` was read only by A. `AssembliesWalked` was read only by A0 and A's label. Nothing else touched the walked universe — so A0 and the derivation went with A rather than surviving under a narrowed header. `CorpusGenerator.WalkedAssemblies` is deleted too: it existed only to feed that sweep, and both of its comments named the check, so leaving it would have parked two false sentences behind a dead field. **The build proves the consumer check** — the field and its property are gone and the solution compiles.
- **The exhibits are untouched.** B, C and D assert reachable classes and rendered content, which are real. Only the delusion is deleted, and the letters stay at B/C/D so the file still reads against this review.
- The sweep's cost goes with it (`GetTypes()` across three assemblies plus a `ClassifyArm` call per type, each non-authorable one running a full `WritableSurfaceCount` inheritance walk). I have not attached a second count to that, because the corpus walk itself was already memoized and I did not measure the delta.

### F2 — folded as written

Folded your suggestion verbatim, and kept it textually parallel with `EmitMatchProbe.RunGuard` so the two read as one policy rather than two.

**Verified by forcing the throw** (emit pointed at a non-existent drive, from the committed state). The preamble and the captured generator output — the walk banner and the assembly line — land on **stderr**, and the exception still propagates to a non-zero exit:

```
  the generator threw during this guard's emit - its output up to the failure follows:
Walking the full Mutagen type corpus via reflection...
  Assembly: Mutagen.Bethesda.Skyrim 0.54.4.0

Unhandled exception. System.IO.DirectoryNotFoundException: ...
```

Restored from the committed state and grep-verified afterwards.

### Verification

`ci-all` **127/127**, `freshness-capture-guard` cold PASS, Release build 0 errors, tree clean at `d02b760`.

### One thing worth recording

Check A was written in round 1, survived rounds 2 and 3, and was described in the header as the thing making byte-identity a consequence rather than a coincidence — three review rounds and twenty-odd agent-passes blessed an arm that could not fail. Your review found it. That is the gate in §5 #10 doing precisely the job it exists for, and it is the strongest evidence on this branch that the pre-PR rounds are a first pass, not a substitute.

Merge is yours — nothing has been merged and the worktree stays in place until you say.

Avick3110:
## Fold spot-check verdict (`d02b760`)

Standing lane watch, second pass. This session executed the directed fold on Aaron's review; it did not review the branch and ran no round four.

**Verdict: both findings folded as directed, and the one fold that can be RED-checked was.**

| item | landed as | verification |
|---|---|---|
| F1 — delete check A | `OriginalPredicate`, `walked`, `candidates`, `disagreements`, A0 and A removed; `CorpusGenerator.WalkedAssemblies`/`AssembliesWalked` removed with them; header re-homes the byte-identity claim to `emit-match-guard` + the #351 CI step | **not RED-checkable — it is a deletion.** What was verified instead: the consumer check (nothing but A read `candidates` or the walked universe) is proven by the compiler, since the field and its property are gone and the solution builds; and B/C/D all still pass |
| F2 — replay the capture on a throw | `catch` → restore console → replay `captured.ToString()` to stderr → rethrow, textually parallel with `EmitMatchProbe.RunGuard` | **forced throw from the committed state** (emit pointed at a non-existent drive): preamble + captured generator output on stderr, exception still propagates to a non-zero exit. Restored, grep-verified |

`ci-all` 127/127, `freshness-capture-guard` cold PASS, Release build 0 errors, tree clean.

**Three notes that are mine, not the ruling's — the parts most worth a second pair of eyes:**

1. **The consumer check resolved to "delete more, not less."** The directive said A0 and the derivation survive *only* if something else consumes the walked universe. Nothing does, so they went — and that carried into `CorpusGenerator`, where `WalkedAssemblies` existed solely to feed the sweep and both of its comments named the check. Deleting the probe's arm while leaving that field would have left two false sentences behind dead code, which is the class this whole branch is about. If the lane wanted the generator side left alone, this is the line to push back on.

2. **The letters were not renumbered.** B/C/D stay where they are rather than shifting up to A/B/C, so the file still reads against PR #425 and the review that found the tautology. The header states the retirement rather than leaving a reader to wonder where A went. That is a tombstone, deliberately — it is small, and the alternative is a silent gap in a lettered scheme.

3. **I did not attach a number to the cost saved.** The review notes the pin cost a 12,993-type × 3-assembly sweep per CI run. That is real work now gone, but the corpus walk itself was already memoized, so the saving is `GetTypes()` plus per-type classifier calls rather than a second full reflection walk — and I did not measure the delta. Recorded as "the sweep is gone" and nothing stronger, rather than quoting a second count I did not take.

**On the process, since it is the honest reading of this round:** check A was authored in round 1, survived rounds 2 and 3, and was described in the guard header as the thing that made shard-level byte-identity a consequence rather than a coincidence. Three rounds of fresh agents blessed an arm that was empty by construction. The pre-PR rounds did real work on this branch — the neutrality break they caught in round 1 was genuine — but this one lands squarely in the column marked "why Aaron's review is the gate."

Issue #426, opened 2026-08-24
[Bug] two apply-guard arms assert the absence of strings the codebase cannot emit, so they cannot fail


Avick3110:
> *Found while reviewing the `insert-at-index` branch (#302). The defect is outside that branch — it predates it — so it is filed here rather than fixed there, per CLAUDE.md §2.*

### What

Two arms in `apply-guard` assert that a response does **not** contain `"NOT landed"` / `"NOT carried by the written file"`:

- `src/housecarl-generator/ApplyGuardProbe.cs:174` — *"two Adds to ONE list in one call: both land and NOTHING is reported as not landed"*
- `src/housecarl-generator/ApplyGuardProbe.cs:207` — *"…and NEITHER is reported as not landed — the earlier op's count is behind the file's, not wrong"*

Neither literal exists in any production code path. Every occurrence in `src/` outside the generator is a comment:

| Location | Kind |
|---|---|
| `src/housecarl-core/WritePatchBuilder.cs:90` | `///` doc comment |
| `src/housecarl-core/WritePatchBuilder.cs:3462` | `//` inline comment |
| `src/housecarl-core/WritePatchBuilder.cs:3529` | `///` doc comment |
| `src/housecarl-mcp/WriteTools.cs:749` | `///` doc comment |

The words the wire actually emits for this concern are the `landed_source` state words — `superseded` / `written_file` / `no_answer` / `not_checked` (`src/housecarl-mcp/JsonWire.cs:2369-2372`). The historical `"NOT landed"` phrasing describes a **count-comparison that has since been removed** (see the note above `VerifyLandedAgainstFile`'s single leaf comparison, `WritePatchBuilder.cs:3467-3473`), so the harm these arms were written against is one the current code can no longer produce.

### Why it matters

The assertions are unfalsifiable — they pass under every mutation, including the exact regression the surrounding block exists to catch.

**Measured.** Mutating `WritePatchBuilder.cs:3532` to admit every count-changing keyed verb to the exemption —

```csharp
static bool CountNeutralKeyedVerb(string verb) => verb is "SetAtIndex" or "InsertAtIndex" or "Remove";
```

— turns `apply-guard` red on the neighbouring `landed_source` arm only. Both "NOT landed" arms stay green.

This is the same shape as the arm deleted in `eef31b3` ("delete a guard arm that could not fail").

### Suggested fix

Delete both assertions rather than repair them. What each one meant to claim is already measured next to it by a stronger arm:

[...]

Issue #386, opened 2026-08-20
[GAP] no guard reaches tool-parameter [Description]s or prompt-builder strings, so consent-vocabulary claims go stale silently


Avick3110:
## The gap

Two caller-facing prose surfaces have no guard behind them:

- **Tool and parameter `[Description]` attributes** (`src/housecarl-mcp/*Tools.cs`) — the text the model reads to decide how to call a tool.
- **Prompt/handshake builders** (`LoadOrderService.InPlaceHandshakeLead`, `InPlaceHandshakeText`, `NifInPlaceHandshakeText`) — the text the caller reads to decide whether to consent.

`write-surface-guard` (224 checks) and the `[MustState]` / `[NoClaims]` walk in `OwnedChildContentProbe` both police `ReadSentences` / `WriteSentences` consts. Neither reaches either surface above. A claim on those surfaces can become false and nothing goes red.

## The evidence: one behaviour change, three passes, 13 stale claims

The in-place consent fix (#378) changed exactly one fact — a refused call no longer records the acknowledgement — and each sweep for text asserting the old fact found homes the previous sweep had missed:

| pass | surface found | stale claims |
|---|---|---|
| 1 | `acknowledge=` parameter `[Description]`s | 11 (`"never again for it"`) |
| 2 | `WriteTools` lane parentheticals | 6 under-enumerated (adjacent class, same sweep miss) |
| 3 | the two handshake prompt builders | 2 (`"shown once for this plugin"` / `"for this mesh"`) |

A fourth pass, run repo-wide by vocabulary rather than by surface, found 14 more sites carrying an overclaiming `one-time` modifier across tool body descriptions, both READMEs, a shipped skill, and the dry-run pending note.

Three of those four passes were triggered by a reviewer noticing a claim by eye. That is the gap: the surfaces are load-bearing and unguarded, so staleness is found by luck rather than by construction.

## Proposed shape

A vocabulary probe in the `wire-names-guard` mould: enumerate the `[Description]` attributes and the prompt builders by reflection, and assert that a named set of consent-vocabulary phrases (`shown once`, `never again`, `one-time`, and siblings) either does not appear or appears only where a declared exemption says it is accurate. The point is not to ban words — it is that adding one to a caller-facing string should require a deliberate act, the way `[MustState]` already works for the sentence consts.

Scope note: this is a guard for the *surface*, not for this issue's particular claim. Whatever replaces "shown until a write LANDS" later should also be unable to go stale silently.

## Provenance

[...]

PR #427, opened 2026-08-24
InsertAtIndex: a modeled-list primitive, and refusal remedies derived from the field's shape

### LINKED ISSUE #302: [GAP] no insert-at-index on modeled lists — extending a mid-list CTDA OR group forces a lossy replace or a full ReplaceAll
> *Filed from an external live-session gap report (dialogue condition-surgery pass); triaged, root-caused against source, and posted by AI.*

### Goal

Extend an existing CTDA **OR group** on 4 dialogue INFOs — add a `GetFactionRank <follower faction> >= 0` arm to an essential gate that reads `[Or] GetPlayerTeammate == 1`, `IsEssential == 0`.

### What fell short

Modeled lists (Conditions among them) support **Add** (append), **Remove**, **SetAtIndex** (replace in place), **ReplaceAll** — but there is **no insert-at-index**.

An OR group is **position-contiguous**: the `Or` flag chains a row to the row after it, so a new arm must sit *adjacent* to the group. `Add` lands it at the end of the list as a separate AND'd group, which can only ever *restrict* the gate, never relax it. No sequence of the existing verbs grows a mid-list OR group — `ReplaceAll` does, but only by re-transcribing all ~40 rows of each INFO, which was rejected as error-prone.

Tool calls tried:

    housecarl_bulk_apply verb=Add   field_path="Conditions"   # → appended at the end, wrong group
    (considered) verb=ReplaceAll with a full 40-row compose per INFO — rejected

### Workaround used

`SetAtIndex` **replaced** the `GetPlayerTeammate == 1` arm with the new faction arm. It works only because the faction happens to be a practical superset of teammates — the intended 3-arm design was flattened to 2 arms. A true insert would have kept all three.

### Frequency

Recurs on any condition surgery over existing OR groups — the reporting session alone did three rounds of in-place condition work this month.

### Ideal shape

`verb=InsertAtIndex` on modeled lists (or `Add` accepting a `key=` index), composing an element the same way `Add`/`SetAtIndex` already do (`compose=` / `composes=`).

- date: 2026-07-31  |  houseCARL version: unknown

---

Valid, reproducible, non-duplicate. The verb set is exactly as reported — `WriteTools.cs:1208` declares `Set | Add | Remove | SetAtIndex | ReplaceAll | Merge | CopyFrom`, and `WriteEngine.ApplyListVerb`'s switch (`src/housecarl-core/WriteEngine.cs:2350-2410`) has `Add` / `SetAtIndex` / `Remove` / `ReplaceAll` arms and no insert.

This is the **next member of the same family SetAtIndex was added for**, and the existing code says so in its own comment (`WriteEngine.cs:2377-2379`):

> Closes HCBR-2026-07-10: replacing one condition row no longer needs Remove+Add, which moved the row to the list END (harmless for an AND row, but breaking an OR-group).

So list *position* is already recognised as load-bearing for conditions, and the fix shipped for **replace**. **Insert** is the arm of that same problem that never got covered: the earlier report needed to change a row without moving it; this one needs to add a row without moving the rows after it.

Fix direction — a small, mechanical addition inside the existing list-verb switch: `IList.Insert(idx, element)`, with the same index pre-check `SetAtIndex` does (`idx < 0 || idx > count` — note `>` not `>=`; inserting *at* `count` is a legal append) surfacing via `IndexRangeMessage(..., append: true)`, and the same `req.Struct is not null ? BuildStruct(...) : Coerce(...)` element construction so a composed condition row inserts exactly as it appends today. `CorpusRulebook`'s composable/KEYSHAPE gate needs `InsertAtIndex` admitted alongside `SetAtIndex` (index shape pre-flight, in-range left to apply, which is the established split).

Not a bespoke verb in the cornerstone sense — it is a missing *primitive* on the generic modeled-list surface, the same shape as the four already there.

Avick3110:
Fixes #302

## What and why

`Add` appends, `SetAtIndex` overwrites. Neither can put a new element **at** a position, and that is what a position-contiguous run needs: a CTDA `Or` flag chains a row to the row after it, so an arm added to an existing OR-group has to land *inside* the run. An `Add` puts it at the end as a separate AND-group, which can only narrow the gate — the reporting session's actual workaround was to spend an existing arm with `SetAtIndex`, flattening a three-arm design to two.

`InsertAtIndex` takes `SetAtIndex`'s arguments and builds its element the same way — `compose=` for a modeled element, `value=` for a plain one — so a composed condition row **inserts** as the identical element it would **append** as. The one thing that differs is the bound: the index names a *gap* between elements, of which there is one more than there are elements, so `0..count` is legal and `count` appends.

`IndexRangeMessage` stops being a `bool append`. That bool meant "may the message offer Add?"; a third mode whose *bound* differs is not a second reading of it. It is now a three-value enum.

## The engine claim, measured

`write-oracle` gained a `list InsertAtIndex Armor.Keywords[0]` cell — Path A (the reflection engine) against Path B (hand-written `Keywords.Insert(0, …)`). Run against a real `Skyrim.esm`:

```
  list SetAtIndex     Armor.Keywords[0]        PASS
  list InsertAtIndex  Armor.Keywords[0]        PASS
  list Remove         Armor.Keywords[0]        PASS
  === Oracle: 21/21 cells byte-identical ===
  Source unchanged: YES
```

So the insert path is byte-identical to hand-written Mutagen, not merely behaviourally similar. (Re-run at `HEAD` after the directed fold — the oracle is not in `ci-all`.)

Also measured: a whole-corpus differential (insert vs `SetAtIndex` vs `Add` over all 240 writable list fields) with zero divergence, and a serialize round-trip in which a row inserted mid-OR-run returns off disk in position with its `Or` flags intact.

## The directed fold — derived remedy verbs

Round 2 hit CLAUDE.md §11's **class-stop**: the same failure class recurred in two consecutive rounds. The class was *hand-maintained copies of the verb set*. Six emitted messages recited verb names next to the condition that produced them, and they drifted — worst of all the modeled-elements refusal, which is emitted for a list **or** a dict and recited the LIST verbs to both. Measured on `Package.Data`: the message offered "InsertAtIndex to insert AT an index", and following that offer returns `InsertAtIndex is only valid on list; 'Data' is dict.`

Per §11 that is a design signal, not a fix queue, so it went to the advisor lane rather than into a third fold. The ruling (Aaron-go) was to build one generator and have every such site consume it.

`WriteVerbs` is that generator. It holds the verb vocabulary once, and one **shape-indexed** derivation of which of those verbs work on a collection — keyed on the two facts the gate itself decides with: cardinality, and whether the elements are plain values, built from parts, or owned child records. Sites choose a purpose subset by flags on each `VerbUse` (`NeedsKey` / `Places`), never by naming verbs, so a dict caller is structurally unable to receive a list-verb remedy. Each suggested verb also states the input slot it needs (`compose=`, `values=`, `key=`), which the hand-written lists left the caller to infer.

Two messages changed beyond their verb list, because deriving the verbs exposed a false claim in the prose around them:

- the modeled-elements refusal ended `"{Verb} of modeled elements is a later surface"`, which is **false on a list** — a modeled list does have a batch replace, `ReplaceAll` with `composes=`. (My replacement said the caller "carried plain values", which round 3 showed is false when they carried nothing at all; it now says what is *missing* rather than what was sent.)
- a list of owned child records has **no** placing verb at all, so its remedy names the record axis rather than trailing off after a dash.

The engine keeps its own route to a shape, because threading the corpus into the verb path would cost it the schema-blindness `SchemaClassifier` is documented to protect. Two routes is a risk, so it is measured rather than assumed.

### How the derivation is held honest

Not by resembling the gate's switch — a mirror of that code would only prove the copy was faithful. `remedy-verbs-guard` (new, in `ci-all`) **replays** instead:

- it buckets every collection field in the record library by shape, and for each bucket sends a well-formed request for **every verb the table names** through the real pre-flight (each must be ACCEPTED) and for **every verb it omits** (each must be REFUSED). **1448 accepts, 520 refusals, both directions, zero cells skipped** — the "not constructible" line reads `none`.
- it holds the schema route and the runtime route to the same answer on all **246** collection fields.
- it asserts each message carries no verb from the OTHER cardinality — as an **absolute** set, after round 3 showed the first version subtracted the table's own claims and so disabled itself exactly when the table was wrong.
- it reports the corpus population of every shape, so an empty bucket cannot pass quietly. `Dict/OwnedRecord` is pinned by name as the one shape Mutagen does not instantiate today; if a bump lights it up, the arm goes red and the claim gets measured before it is trusted.
- per-site arms, one per cardinality, each asserting the message carries this shape's verbs **and none of the other cardinality's**. Asserting only the first half is what let the medium ship: the old message did name a working verb, alongside three that refuse.

The refuse direction earned its place on its first run: the table named `CopyFrom` for dict shapes, and the gate refuses it (`a dict isn't transplanted yet`). Removed, measured.

### Class B — assertions that could not fail

Bounded fold, no new machinery, per the same ruling:

- **The mis-mapping medium.** The `bool append` → `IndexOpKind` refactor re-mapped two *pre-existing* callsites by hand and nothing caught a wrong mapping: setting `bool append = false` left `insert-at-index-guard` 25/25 and `ci-all` 128/128 green while an out-of-range `SetAtIndex` silently lost "or Add to append" and an empty-list `SetAtIndex` started telling the caller there was "nothing to remove" — advice for a verb they did not use. That reproduction is now two arms.
- **Abort shadowing.** The apply arm ran eight sub-arms under one `try`, so a throw in an early one abandoned the rest. Measured, same breakage: **19 assertions reported without per-sub-arm isolation, 28 with it.** Never falsely green, but a sweep is read for how much broke.
- **Two controls now say they are controls**, labelled for what each actually holds rather than blanket-labelled — `GATE-OK-FARIDX` fully, `APPLY-ABSENT-AT-0` in part (its materialize half is verb-agnostic; its bound half is already `APPLY-AT-COUNT-IS-ADD`'s).
- **The CLI.** Both of `RunPatch`'s value-requiring verb lists gained the verb and neither was asserted. Now driven through `RunPatch` itself.

## Guards

- **`insert-at-index-guard`** (29 arms, was 25): the gate; the position claim asserted by **reference identity** on the tail (equal-looking values would also pass if the engine rebuilt the tail out of copies); the append-inclusive bound proved as an **equality with `Add`**; a serialize + binary-overlay round-trip; the two sibling messages; the two CLI checks.
- **`remedy-verbs-guard`** (new, 23 arms): above.
- **`apply-guard`**: the counter-arm proving `InsertAtIndex` is NOT admitted to the count-neutral keyed exemption, and the change-summary render driven through the real response.
- **`write-surface-guard`**: four rows in the owned-child disposition table, each **paired with its `SetAtIndex` sibling** so a fork reads as a disagreement between two rows rather than as an absence.
- **`write-oracle`**: the byte-identity cell above (not in `ci-all`; run by hand).

`ci-all`: **129/129**. `insert-at-index-guard` 29/29, `remedy-verbs-guard` **23/23**.

## Review rounds

Three rounds, the §11 cap. Fresh reviewers each round, isolated worktrees, prompts seeded with the settled-decisions list and nothing else.

### Round 1 — three reviewers. No highs. Two mediums, ~11 lows.

Both mediums were against **this branch's own guards**:

1. The negative-index apply arm asserted only "something threw, count unchanged". Deleting `idx < 0 ||` from the bound leaves the reflection `Invoke` throwing and the list untouched, so the arm stayed green on exactly the regression it names — while the caller got a bare `TargetInvocationException` with no field, no index, no bound, and not the kind the response layer routes as an expected refusal. Independently reproduced (23/23 green under the mutation). Now asserts the kind *and* the words.
2. An arm asserting the absence of `"NOT landed"` — a string this codebase can only produce inside comments. Unfalsifiable; deleted rather than repaired, since the arm beside it already measures the real claim against the written file.

The dominant low class, reached independently by all three: **the behavioural sweep was exhaustive, the prose sweep was not.** Six `compose=` descriptions still read "for an Add"; the gate's leaf-bracket remedy omitted the verb while its engine twin had it; `set_field`'s `field_path` published a different leaf vocabulary from its own params. Two sentences the branch introduced were wrong: a claim that the non-insert verbs "address an element that already exists" (false for the dict half), and a cross-reference pointing "below" at a note that renders above. Also folded: `InsertMethod`'s `?? throw` was unreachable, the same shape as the arm the base commit `eef31b3` was made to delete.

**Refused — 3, with grounds:** `@editorid` same-call refs (SetAtIndex gets the identical refusal on the identical branch, so insert has not forked); `CorpusRulebook.cs:403` (unreachable with `verb=InsertAtIndex` — `ComposesLegality`'s verb check fires first); a refused out-of-range insert leaving a materialized empty list (confirmed side by side with `SetAtIndex`, so pre-existing).

### Round 2 — three reviewers. No highs. Class-stop triggered; nothing folded.

Two classes recurred from round 1, so §11's class-stop applied and round 2's findings went to the advisor lane instead of into a fold.

**Class A — hand-maintained copies of the verb set** (three of four instances in prose the branch itself authored): the modeled-elements medium above; `plugin/CHANGELOG.md` advertising the verb on the create tools that line 746 of the same file says `housecarl_create` replaced, while omitting the shipping one; `mutagen-reference` SKILL.md's "differ **only** in WHERE it goes" (false — they also differ in legal index range); the `Set`-on-list remedy naming two of five list verbs.

**Class B — assertions that cannot fail:** the `bool append` mis-mapping medium; abort shadowing; two unlabelled reachability controls; three added `InsertAtIndex` mentions unasserted.

### Round 3 — three reviewers, post-fold. No highs. Three mediums, five lows/nits. All folded but two, listed below.

Convergence was high: the top finding was reproduced independently by **all three**, and the second by all three as well (one at low).

**The medium all three found — and it is the branch's own class, in the half the fold did not convert.** The `composes=` refusal checks the VERB first, and its head sentence ("use it with `Add` or `ReplaceAll`") is true of `composes=` in general and false for the caller when `composes=` does not serve their field at all. A dict caller (`Package.Data`), a substruct caller (`Armor.BodyTemplate`) and a coercible-element list caller (`Armor.Keywords`) were each told to reach for a verb that refused on their very next call. The first fold derived that sentence's *parenthetical* and left its first half doing exactly what the branch is chartered against.

Fixed structurally rather than by rewording: **the shape questions are asked before the verb question.** By the time the verb arm is reached the leaf is a list of modeled elements, which is the only shape that sentence is about. Three new arms — dict, substruct, coercible list — each asserting the caller is told what their shape is and offered no verb that would refuse next.

**Second medium (3/3): the bracketed-leaf remedy led with the wrong verb.** A caller who writes `Keywords[0]` bracketed an index that already holds an element; the menu led with `InsertAtIndex`. Every other wrong first choice on this branch refuses — this one *succeeds*, one element longer with the tail shifted, which on a CTDA OR-run changes what the record gates on. `SetAtIndex` now precedes it in the table, asserted as a position comparison so a reordering fails the arm.

**Third medium: "One element at a time:" labelled a list of batch verbs** — the sentence refused a batch verb and then recommended one, and on a dict refused `Merge` and offered `Merge` in the same breath. `HowToPlaceOne` subtracts the verbs whose input slot carries a whole collection; batch-ness is read off the slot the table already records.

**Lows folded:** "a ReplaceAll carrying plain values" was a false diagnosis (a bare `ReplaceAll` with no input at all reaches that message, and the code comment claiming otherwise was wrong); `CheckShaped`'s cross-cardinality half was **disabling itself** — it subtracted the table's own claims from the forbidden set, so sabotage that put `Merge` into every list remedy left all fourteen SITE arms green (now absolute: the same sabotage fires four); `WriteVerbs.All` was ordered differently from the tool descriptions its own doc cites; the class doc said six consuming messages where there are seven; a comment called an unresolved-field-name path "dormant" when any bracketed typo reaches it; and the population line printed a trailing dash.

**Found by me while walking what the reorder changed, not by a reviewer:** widening the door to the coercible-element refusal meant a `SetAtIndex` caller would be told `value=` belongs to `Add`. That remedy now derives its slots too — the eighth consuming site.

**Refused — 2, with grounds:**
- *Drop `InsertAtIndex` from the bracketed-leaf menu entirely* (one reviewer's preferred fix for the second medium). Refused: it is a legal keyed way to reach that index, it carries its own description in the menu, and withholding it hides a verb the caller may have wanted. Ordering removes the harm; exclusion trades it for a different one.
- *Make the array refusal grammatical for a hypothetical owned-record array* ("… ; Remove are not yet supported"). Refused: measured — the corpus has exactly two array-backed fields, `Weather.CloudTextures` (List/Coerced) and `Weather.Clouds` (List/Composed), both rendering the full five names. Building a plural-aware joiner for a shape that cannot occur is machinery for nothing; stated instead.

**Judgement call, surfaced rather than decided quietly.** A reasonable reader could call the first medium a *third* instance of class A, which §11 says is never folded. I judged it a completeness miss inside a converted site rather than a recurrence, on two grounds: the fix is a change to the single derivation and its ordering — not another corrected copy — and the ruling's own escalation trigger was "the class INSIDE the formatter's own arms", which this is not (the formatter's arms were correct; the site's head sentence was unconverted). Neither of the ruling's two STOP conditions fired. **If you read it the other way, this is the point to say so** — the alternative is that the composes= site goes back to the advisor rather than shipping here.

## Sabotage

Three passes, twenty-four cells, each with a canary proven RED **by hand** before the sweep ran, each cell mutating one line from a committed clean tree, rebuilding, running the named guard, restoring, and verifying the restore by grepping for the original string. Both passes ended with a clean tree.

The sweep found **two arms of the fold itself that could not fail** — which is the whole reason it is run rather than reasoned about:

- `APPLY-SIBLING-REMOVE-MSG` built its empty list by leaving it **absent**, and an absent list short-circuits to a different throw, so the arm never reached the message it names and stayed green under exactly the mis-mapping it was added to catch.
- `SITE-UNKNOWN-VERB` checked the emitted message against `WriteVerbs.All` — the same collection that produced the message. Deleting a verb deleted it from both sides at once. The check-A tautology, inside the guard written to prevent it.

Both fixed in `67447e1` and re-run RED. Neither was visible on review; the fold's diff was read before commit and both arms read as correct.

## Disposition list

Regenerated at `4130d5d`, after the directed fold and the round-3 fold. Enumerated from `grep -rn "InsertAtIndex\|WriteVerbs\."` over `src/` and `.claude/skills/`, then walked at source. The completeness claim is this enumeration, not that nothing was missed.

**What the fold changed about this list.** Six sites that used to *spell* the verb set now *derive* it from `WriteVerbs`, so they no longer appear as "added" sites at all — they appear as consumers, and a future verb reaches them without an edit. Two entries in the previous version of this list moved from "deliberately absent" to "consumer" for that reason, and one of its factual claims was wrong and is corrected below.

### The gate — `CorpusRulebook.cs`

| Line | Site | Why it must carry the verb |
|---|---|---|
| 14 | `WriteRequest.Verb` comment | the wire type's own vocabulary note |
| 374–382 | `VerbLegality` arm | list-only + index PRESENCE; a keyless insert would `ArgumentNullException` at apply |
| 619 | step-4-key index SHAPE | `int.Parse` accepts `-1`; the indexer does not |
| 727 | step-4-pre element-value PRESENCE | insert consumes the same singular `req.Value` slot; a null lands a null element that throws at serialize |
| 759, 778 | step-4a / step-4b element value SHAPE | same slot, same coercion at apply |
| **796** | **step-4-rec record-element redirect** | **load-bearing.** Left out, an insert into `Cell.Persistent` is ACCEPTED then throws at apply — a Q3 accept-then-throw |
| 829 | composable-compose admission | insert builds through the same `BuildStruct`; same `StructElementLegality` |

### The engine — `WriteEngine.cs`

| Line | Site |
|---|---|
| 2702 | `ApplyListVerb` switch arm — the verb itself, append-inclusive bound |
| 2775–2792 | `IndexOpKind` + `IndexRangeMessage` (`bool append` → a three-mode enum, because insert's BOUND differs) |
| 2976 | `InsertMethod` — `IList<T>.Insert` off the closed interface |
| 202, 212 | CLI `--op` / `--verb` needs-a-value checks — **now asserted** (`CLI-OP-NEEDSVALUE`, `CLI-VERB-NEEDSVALUE`) |
| 3992 | `ExpectedApplyRejectionException` doc — the bound is the verb's |

### Render / batch

| Line | Site |
|---|---|
| `ReadEngine.cs:586` | change-summary arm — "inserted [k]", not a bare count |
| `WritePatchBuilder.cs:3515` | `CountNeutralKeyedVerb` **doc** names the verb as excluded (the predicate is deliberately unchanged — see below) |

### Tool schemas — `[Description]` surfaces

| Site | Subset it joins |
|---|---|
| `ApplyTools.cs:12, 50–51, 353` | the full 7-verb ACT set |
| `CreateTools.cs:248` | full set minus `CopyFrom` (a create has no copy pole) |
| `WriteTools.cs:25, 42, 44` | `set_field`'s five → six |
| `WriteTools.cs:1579` | `BulkOp.verb`, full set |
| `WriteTools.cs:1651` | `NestedSet.verb`, four → five — **ground below** |
| `.claude/skills/mutagen-reference/SKILL.md:114` | the `list` cardinality row's verb column |

These stay hand-written on purpose: making the attribute strings by-construction (verb-set consts concatenated into attributes plus a vocabulary guard) is #386's territory and lands as its own PR.

**`NestedSet.verb` (:1651) — why insert belongs to that four-verb family and `ReplaceAll`/`Merge` do not.** Not curation: the DTO's slot set decides it. `NestedSet` carries exactly `path`, `verb`, `value`, `key`, `compose` — there is **no `values`** and **no `entries`**. `ReplaceAll` reads `req.Values` and `Merge` reads `req.Entries`, so neither has an input surface on this shape at all. The four listed verbs each work from `value` + `key` + `compose`, and `InsertAtIndex` needs precisely `key` + (`value` or `compose`) — the same slots. Apply-side it is reachable by construction: `BuildStruct` runs each `spec.Sets` entry through `WriteEngine.ApplyVerb`, the full generic engine, no per-verb filter.

## Sites that now DERIVE the verb set (no longer spell it)

Eight emitted messages. Each takes its verbs from `WriteVerbs` keyed on the field's shape, so none of them can name a verb that shape refuses, and a verb added to or removed from the surface reaches all eight at once.

| Site | What it derives | Guard arm |
|---|---|---|
| `CorpusRulebook.cs:242–244` | leaf-bracket remedy — the KEYED verbs, existing-element verb first | `SITE-BRACKET-GATE-LIST` / `-DICT` / `SITE-BRACKET-ORDER` |
| `WriteEngine.cs:1956–1959` | the same remedy at apply, via the runtime route | `SITE-BRACKET-ENGINE-LIST` / `-DICT` |
| `CorpusRulebook.cs:341` | Set-on-list remedy — the PLACING verbs | `SITE-SET-ON-LIST-COERCED` / `-COMPOSED` / `-OWNEDRECORD` |
| `CorpusRulebook.cs:390` | unknown-verb legal list — the vocabulary, from its one home in code | `SITE-UNKNOWN-VERB` |
| `CorpusRulebook.cs:446` | coercible-element `composes=` refusal — the slot each verb wants | `SITE-COMPOSES-COERCIBLE` (+ `-NOT-VERB-FIRST`) |
| `CorpusRulebook.cs:456` | `composes=` singular-path parenthetical — the SINGULAR placing verbs | `SITE-COMPOSES-LIST` / `SITE-COMPOSES-ONE-AT-A-TIME` |
| `CorpusRulebook.cs:906` | modeled-elements `ReplaceAll`/`Merge` remedy | `SITE-MODELED-LIST` / `-DICT` |
| `WriteEngine.cs:2648` | array-backed refusal — the collection verbs | `SITE-ARRAY-REFUSAL` (+ `-NOT-COPYFROM`) |

Plus the three shape-answers a `composes=` caller now gets **instead of** a verb list, because the shape questions are asked first: dict, substruct/scalar, and coercible-element list (`SITE-COMPOSES-DICT` / `-SUBSTRUCT` / `-COERCIBLE`).

`CorpusRulebook.cs:241` (leaf-bracket) and `:339` (Set-on-list) were listed as "deliberately absent" in the pre-fold version of this list, on the ground that neither sentence is about inserting. That ground was right about *insert* and wrong about the *set*: both sentences enumerate verbs, and both had already drifted. They are consumers now, which is a stronger answer than either "add insert" or "leave it".

## Deliberately absent, with the ground

| Site | Ground |
|---|---|
| `WritePatchBuilder.cs:3515` `CountNeutralKeyedVerb` **predicate** | the verb must NOT be admitted. An insert moves the count *and* shifts every index at or after it — doubly not independent of a second keyed op. It classifies with `Remove`-by-index. Guarded by a counter-arm in `apply-guard`. |
| `CorpusRulebook.cs:550` | `@editorid` same-call refs stay `Add`/`ReplaceAll` on a FormLink list. Refuses **by name** — loud, not a silent gap. `SetAtIndex` gets the identical refusal on the identical branch, so insert has not forked. |
| `LoadOrderService.cs:7426` | `composes=` (many elements in one op) stays `Add`/`ReplaceAll`; a batch **insert** is a distinct input shape. Refuses by name and points at `compose=`. |
| `WriteEngine.cs` dict-Add-duplicate "use Set to overwrite that entry" | single verb, dict-scoped — not an enumeration. |
| `CorpusRulebook.cs` Remove-by-value redirect | single verb. |
| `WriteTools.cs:38` | `field_path`'s leaf-addressing examples are element-scoped; the `key` param one line below (:44) names insert and explains its index meaning |
| `WriteEngine.cs:165` | CLI usage example illustrates the `--op` grammar, not the verb set |
| `SkyPatcherOverlay.cs:1482` | a **caller** of `SetAtIndex`, not a spelling site |
| `.claude/skills/bulk-record-jobs/SKILL.md` | its verb mentions are `composes=`-specific, not an exhaustive verb-set claim |

## CORRECTION to the pre-fold version of this list

It said:

> **Array refusal is dormant-by-construction.** The message at `WriteEngine.cs:2634` gained the verb, but the current corpus has **zero** array-typed fields (checked: no `Type` containing `[]`), so no guard arm can fixture it honestly.

**That is wrong, and the check behind it measured the wrong thing.** Array-ness is invisible in the corpus's rendered `Type`: `Weather.CloudTextures` catalogs as `List<IAssetLinkGetter\`1>` because the corpus records the *getter interface* (`IReadOnlyList<T>`), while the runtime property type is `IAssetLink\`1[]`. Searching the corpus for `[]` therefore returns zero on a model that does contain array-backed collections.

Measured instead: `WriteEngine.ApplyVerb` on a live `Weather` with `Verb = "Add"`, `Path = ["CloudTextures"]` reaches the array branch and emits

> `'CloudTextures' is an array-backed collection (IAssetLink\`1[]); Add, InsertAtIndex, SetAtIndex, ReplaceAll, Remove are not yet supported on arrays …`

So the site is fixturable, is fixtured (`SITE-ARRAY-REFUSAL`, plus `SITE-ARRAY-REFUSAL-NOT-COPYFROM`), and its verb list is derived rather than stated. This also means the advisor's ruling — which took "the array refusal stays honestly unfixturable" from the earlier claim — was working from a premise that does not hold; the outcome is strictly better than the ruling asked for, so it is reported rather than escalated.

## Skill check (bounded)

`grep SetAtIndex` over bundled skills returns two files. `mutagen-reference` SKILL.md:114 — the `list` row's "Verb(s) the write tool accepts" column, introduced by a sentence that says it *mirrors the write tools' own pre-flight rules* — carries the verb, and its "the three that place one element differ only in WHERE it goes" claim was false (they also differ in legal index range) and is corrected. `bulk-record-jobs` is `composes=`-scoped and stays for W6. Nothing else.

## Non-derivable prose residue

Two defects the ruling explicitly took out of the generator's reach and had folded by hand:

1. `plugin/CHANGELOG.md` — the `InsertAtIndex` entry named `housecarl_create_record` / `housecarl_bulk_create` (which CHANGELOG line 746 says `housecarl_create` replaced) and omitted the shipping 2.0 tool. It now names `housecarl_apply` and `housecarl_create` and points at the rule that covers the rest ("every other op or verb parameter that already takes `Add` and `SetAtIndex`"), per the file's own "point at the bound rather than restating it" convention — a second hand-written tool list would be the same rot in a new place.
2. `.claude/skills/mutagen-reference/SKILL.md:114` — "the three that place one element differ **only** in WHERE it goes" is false: `SetAtIndex` must address an element that exists (`0..count-1`) while `InsertAtIndex` addresses a gap between elements (`0..count`). Corrected to say both differences.

Neither is mechanizable: they are claims about tool naming and about verb semantics, not enumerations of a verb set. Stated here as the class's residue rather than left implicit.

## Filed, not fixed

**#426** — two *pre-existing* `apply-guard` arms (`ApplyGuardProbe.cs:174`, `:207`) share the unfalsifiable pattern. Outside this branch, so filed rather than fixed here. Per the ruling, the cross-branch class evidence is now deposited there as a comment: **seven instances from four independent origins**, in four sub-shapes (twin expression, twin collection, unemittable string, fixture that never reaches its subject), with the two this fold's own sweep found. That census is what W7's harness wave should be sized against.

## Correction to a claim in the earlier disposition list

The pre-fold list said the array-backed refusal was "dormant-by-construction … the current corpus has zero array-typed fields (checked: no `Type` containing `[]`)". **The check measured the wrong thing.** Array-ness is invisible in the corpus's rendered `Type`: `Weather.CloudTextures` catalogs as `List<IAssetLinkGetter\`1>` because the corpus records the *getter interface* (`IReadOnlyList<T>`), while the runtime property type is `IAssetLink\`1[]`. Measured instead, `ApplyVerb` on a live `Weather` reaches the array branch and emits its refusal — so the site is fixturable, is fixtured (`SITE-ARRAY-REFUSAL`), and its verb list is derived rather than stated. The ruling took "the array refusal stays honestly unfixturable" from that earlier claim; the outcome is better than it asked for, so it is reported rather than escalated.

## Escalations

None new. The two STOP-and-surface conditions the ruling set — a site the formatter cannot serve honestly, and any sign of hand-special-casing a shape — did not fire. The one site the formatter declines to serve with verb names (a leaf whose element kind has no settled legal-verb answer) declines by **shape**, not by site, and `remedy-verbs-guard` measures that no corpus field is in that state today.

## For your take-or-drop

`4f99ff0` edits `CLAUDE.md` §1 — replacing the inline verb enumeration with a lookup rule pointing at the verb set's one home. Aaron-directed, but editing the operating manual is Aaron-gated (§5 #10), so it is surfaced here rather than treated as landed. Drop the commit if you would rather the manual keep its list.


src/housecarl-core/CorpusRulebook.cs
```diff
@@ -387,10 +435,25 @@ static bool IsFlagsEnumLeaf(FieldSchema leaf)
...
+            // this sentence: it used to be seen only by an Add/ReplaceAll caller, for whom "values= (ReplaceAll) /
+            // value= (Add)" named their own verb, and a SetAtIndex caller now lands here and would be told value=
+            // belongs to Add. Every verb this shape takes, with the slot each one wants, is the answer that does
+            // not go narrow on whoever arrives.
             return $"'{leaf.Name}' on '{owner.Name}' holds " +
                    (leaf.FormLinkTarget is not null ? "formlink" : "coercible") +
-                   $" values ({leaf.ElementTypeRef ?? leaf.ElementType}), not modeled structs — use values= " +
-                   "(ReplaceAll) / value= (Add), not composes=.";
+                   $" values ({leaf.ElementTypeRef ?? leaf.ElementType}), not modeled structs, so composes= has " +
+                   $"nothing to build — {PlacingRemedy(leaf)}.";
```
Avick3110, line 463: **[medium] `composes=` on an owned-child-record LIST mislabels the field as "coercible", contradicting the tail this PR derives.**

Measured at `4130d5d` against a freshly generated corpus:

```
composes= Add on Cell.Persistent ->
  'Persistent' on 'Cell' holds coercible values (Placed), not modeled structs, so composes= has
  nothing to build — its elements are owned child RECORDS, which are created on the record axis —
  use housecarl_create_record / housecarl_bulk_create with parent= the parent's FormID, not a write verb.

composes= Add on DialogTopic.Responses -> same shape, "holds coercible values (DialogResponses)"
composes= ReplaceAll on Cell.Persistent -> same
```

One sentence asserting two contradictory facts about the same field. `Cell.Persistent` holds no coercible values at all — step-4-rec refuses every plain-value verb on it — so the head clause is false.

Why it lands here: an `ElementKind.Record` list fails `IsComposableElement`, and this branch's label is a two-way `formlink`/`coercible` ternary with no arm for the third element kind. `SchemaClassifier.IsOwnedChildRecord`, checked at the top of `ComposesLegality`, only matches a *singular* child field, not a list of them.

Not pre-existing as a contradiction: before this PR the head was equally wrong but the tail — "use values= / value=" — was wrong in the same direction. Deriving the tail from the shape made it correct and left the head arguing with it. That is the same class the PR fixes for the dict, substruct and coercible-list callers, on the one shape whose branch order was not revisited.

Suggested fix: route `SchemaClassifier.ClassifyElement(leaf, _corpus) == ElementKind.Record` to the record-axis answer *before* the coercible/formlink label, the way step-4-rec does — so head and tail come from one classification.

Guard hole worth closing with it: `RemedyVerbsGuardProbe.SitesArm` fixtures `SITE-COMPOSES-DICT` / `-SUBSTRUCT` / `-COERCIBLE` but has no owned-record-list arm, which is why `remedy-verbs-guard` is 23/23 green over this. `SITE-SET-ON-LIST-OWNEDRECORD` covers the Set-on-list door, not the `composes=` one.

Avick3110:
## Directed fold — the `composes=` owned-record-list medium

**`43fefeb`**, on top of `4130d5d`. One medium folded; no new review rounds follow (§5 #11 — fresh eyes are for before the PR opened). `ci-all` **129/129**, tree clean.

### What folded

`ComposesLegality`'s not-composable arm labelled the field with a two-way `formlink`/`coercible` ternary that had no arm for a record element, while the tail this branch derives named the record axis correctly. Reproduced at `4130d5d`, verbatim from the sabotage RED below:

```
'Persistent' on 'Cell' holds coercible values (Placed), not modeled structs, so composes= has nothing
to build — its elements are owned child RECORDS, which are created on the record axis — use
housecarl_create_record / housecarl_bulk_create with parent= the parent's FormID, not a write verb.
```

The head is the false half. `SchemaClassifier.IsOwnedChildRecord`, asked at the top of the method, matches only the **singular** child shape, so the list-of-owned-records door fell past it into the label.

One classification now feeds the whole sentence, the way step-4-rec decides it. What ships:

```
'Persistent' holds owned child records (Placed); a child record is created on its own (the record
axis), not added into a parent's collection by a write verb. Use housecarl_create_record /
housecarl_bulk_create with parent= the parent's FormID (and collection= when the parent holds more
than one fitting list) — surfaced here, never accepted and thrown at apply.
```

### One home, not a near-twin

The recogniser (`IsOwnedChildRecordCollection`) and the sentence (`OwnedChildRecordCollectionRefusal`) are named members beside their singular twins. Three doors reach this shape and all three now ask the same question: step-4-rec and `composes=` share both members; `CopyFrom` shares the **predicate** and keeps its own remedy, because `forward_record` is what was measured for a transplant.

**Where the direction's preferred home did not work, stated rather than forked quietly:** extending `OwnedChildSetRefusal` (the singular refusal) was the first thing tried and is wrong — its remedy is the *opposite* one. A singular child routes to #350 ("giving a parent a child it lacks is an open gap") precisely because `create_record parent=` is measured **refused** for it; the collection form is the shape `parent=` serves. Two shapes, two measured remedies, two sentences. Sharing the predicate across all three doors is what stops them drifting; sharing the *sentence* across those two would have made one of them lie.

### The remedy is cited, not re-asserted (#343)

Not measured fresh — it was already measured. `nested-create-guard`'s **INTOCELL** arm creates a `PlacedObject` with `parent=`<cell> `collection=Persistent` and reads it back out of the cell override's `Persistent` list; the arm beside it measures `collection=` being **required** when the parent holds more than one fitting list, which is why the parenthetical names it rather than leaving the caller to a second refusal. Both arms are green in this run.

### The #352 disposition table had no row for this door

It covered `composes=` at the **singular** child (three rows) and every collection verb at the **list**, but not the batch input surface *at* the list — which is the one door that short-circuits above step 4, and so was answering from its own label. Four rows added (`Add`, `ReplaceAll`, `DialogTopic.Responses` so the answer is the shape's and not one field's, plus an ordinary-modeled-list control). `write-surface-guard` **229 → 233**.

### Guard hole closed

`remedy-verbs-guard` **23 → 26**. `SITE-COMPOSES-OWNEDRECORD` on `Cell.Persistent` and `DialogTopic.Responses` asserts **both** halves — record-axis remedy present *and* the `coercible`/`formlink`/`not modeled structs` vocabulary absent, because either assertion alone stays green on the contradiction — plus `SITE-COMPOSES-OWNEDRECORD-ONE-SENTENCE`, which pins that the `composes=` door and the collection-verb door emit one identical string.

Sabotage sweep, from a committed state, restores verified by grepping a string the fold introduced. Cell 0 was proven RED **by hand before the script existed** and carried in as the sweep's canary:

| cell | mutation | result |
|---|---|---|
| C0 canary | the `ComposesLegality` record arm deleted | RED — remedy-verbs 3, write-surface 3 |
| C1 widen | the recogniser drops its element test (every list/dict matches) | RED — remedy-verbs 7, write-surface 3, insert-at-index 7 |
| C2 fork | `composes=` answers with a near-twin literal instead of the shared sentence | RED — remedy-verbs 1 (the ONE-SENTENCE arm alone) |
| C3 step-4 | the collection-verb door loses the shared call | RED — remedy-verbs 2, write-surface 2, insert-at-index 1 |

Both directions: C0/C3 fail on the arm being absent, C1 on it being unbounded. C2 exists because an anti-fork arm that cannot fail is not an anti-fork arm — it goes red alone, while the presence/absence arms stay green.

### Bounded same-class sweep

The defect's signature is *a hand-written head label beside a derived tail*. Bound: every call site of the derived-remedy helpers (`PlacingRemedy`, `WriteVerbs.HowToPlace` / `HowToPlaceOne` / `HowToAddress`) in `CorpusRulebook.cs` and `WriteEngine.cs`, plus every ternary label inside an emitted message in those two files. Line numbers at `4130d5d`.

| site | head | verdict |
|---|---|---|
| `CorpusRulebook.cs:446` | `formlink`/`coercible` ternary, no record arm | **the defect — converted** |
| `CorpusRulebook.cs:341` | `Set is not valid on list '{name}'` | the word "list" *is* `leaf.Cardinality`; derived, cannot contradict |
| `CorpusRulebook.cs:244` | `Path '…' brackets a collection element at the LEAF` | names no shape at all; the no-shape fallback arm names no verbs either |
| `CorpusRulebook.cs:456` | `composes= appends/replaces a LIST of modeled elements` | established by the two shape checks above it on the same path — control flow, not a label |
| `CorpusRulebook.cs:906` | `holds modeled elements ({ElementTypeRef})` | element ref read straight off the schema; reached only inside `IsComposableElement` |
| `WriteEngine.cs:1959` | the engine's bracket twin | names no shape; same as `:244` |
| `WriteEngine.cs:2634` | the array refusal | refused set derived; cardinality settled by the control flow that reached it (`ApplyVerb` matched `IList<T>`) |

A `grep` for the ternary pattern over both files returns **exactly one** label ternary inside a caller-facing refusal — `:444`, the defect. Every other hit is console/report formatting (`YES`/`NO`, `OK`/`!! UNEXPECTED`, key rendering), not a refusal. So: one site converted, six stated as structurally unable to contradict their tail, no second instance of the pattern.

### Judgement calls worth your eye

- **The predicate was shared into `CopyFrom` too**, slightly past the literal ask. It was the identical inline expression; leaving it inline is how two recognisers of one shape start disagreeing. Its message is untouched, and the table's `CopyFrom`-on-`Persistent` control row pins it.
- **The new arm is deliberately not verb-scoped** the way step-4-rec is. `composes=` is a build-from-parts *input surface*, and a record is not built from parts under any verb, so the answer does not depend on which verb arrived. C1's RED is what bounds it.
- **`CHANGELOG`** — no new bullet. The existing derived-remedies bullet already claimed "a collection of child records is pointed at the record axis instead of at a verb that will refuse", which was **false for this door** at `4130d5d`. One clause added to that bullet so the claim is true and the `composes=` door is on the record.

Nothing was refused — the finding was verified at source before any code moved, and the mechanism read matched yours exactly.

`4f99ff0` (the `CLAUDE.md` §1 lookup-rule commit) is still surfaced for your take-or-drop at the merge gate.

Avick3110:
**Advisor spot-check of the directed fold (43fefeb) — PASS, verified at source.**

The review's medium is closed at the mechanism: the owned-record-list door now answers BEFORE the formlink/coercible label, off the same `ClassifyElement == Record` classification step-4-rec decides with — one named predicate (`IsOwnedChildRecordCollection`) and one named sentence shared by the doors, so head and tail come from one reading of the field. The deviation from the fold directive's "extend the existing owned-child refusal" was surfaced, not absorbed, and is CORRECT: the singular shape's remedy (#350 — `parent=` measured refused) is the opposite of the collection shape's (`parent=` measured working, nested-create-guard's INTOCELL arm cited), so two sentences with a shared recogniser is the honest structure, not a near-twin fork. The CopyFrom predicate-share beyond the literal ask is endorsed — an inline twin expression left standing is the drift this fold exists to kill. #352 table gained the four missing composes=-at-list rows; guard deltas (remedy-verbs 23→26, write-surface 229→233) and the four-cell RED sweep with hand-proven canary are consistent with the diff. Same-class sweep: one converted, six structurally unable to contradict, no second ternary instance — the completeness claim the earlier folds lacked.

Merge is Aaron's word, with 4f99ff0 (the CLAUDE.md §1 verb-list lookup rule) at his take-or-drop.

Issue #432, opened 2026-08-25
[Docs] PapyrusDecompiler.cs:1107 — malformed XML doc comment discards the member's docstring (CS1570)


Avick3110:
Found measuring PR #430's docstring fold (rebuild with `-p:GenerateDocumentationFile=true`):

```
PapyrusDecompiler.cs(1107,106): warning CS1570: XML comment has badly formed XML -- 'End tag 'summary' does not match the start tag 'N'.'
PapyrusDecompiler.cs(1108,1):  warning CS1570: XML comment has badly formed XML -- 'Expected an end tag for element 'summary'.'
```

Same class as PR #430's review finding on `WriteVerbs.cs`: an unclosed tag inside `<summary>` (here an `<N>` — likely a generic-ish angle-bracket token typed unescaped) makes the compiler drop the whole docstring, so generated docs and IDE quick-info lose it. Doc generation is off in the csproj today, so CI stays green either way; the malformation is in the source regardless.

Outside the branch under review, so filed rather than folded (CLAUDE.md §2, the #324 shape). Fix is local: escape or close the offending token at PapyrusDecompiler.cs:1107.

PR #433, opened 2026-08-26
Caller-facing prose vocabulary guard: membership from the compiled artifact's own manifest

### LINKED ISSUE #386: [GAP] no guard reaches tool-parameter [Description]s or prompt-builder strings, so consent-vocabulary claims go stale silently

Avick3110:
Closes #386.

## What the guard is

`description-vocab-guard` is a standing CI instrument over the two caller-facing prose surfaces nothing read the *words* of: the `[Description]` attributes a model reads to decide how to call a tool, and the consent prompts a modder reads to decide whether to say yes. It holds a watchlist of consent-vocabulary phrases (`shown once`, `never again`, `one-time`, and their siblings, each row carrying its ground) against every string-literal sentence in the three shipped source trees, and it holds the write-verb vocabulary, the marked default and the tail gloss on the `[Description]` surface against independently written statements of the same facts. **The mechanical half is the vocabulary and the enumerations**; **free-form prose truth is declared residue** — the guard cannot tell a true sentence from a false one built entirely of known words, and it says so on every run rather than in a docstring. What it makes structural is that putting a consent claim in front of a caller as ordinary text takes a deliberate act.

## The Y′ tier contract

Every arm declares which claim it is making, and the run prints both tiers with both counts:

| tier | arms | what the tier means |
|---|---|---|
| `[c1]` by construction | **27** | the two-reader net and the arms over it, and the pins, which hold two independently written statements of one fact against each other. Unqualified claims. |
| `[c2]` best-effort | **5** | every arm that reads MEANING out of prose or reflection BY PATTERN. A pattern's reach is not a by-construction fact, so each of these **prints its own coverage**: how much of what is present it compared, and what it skipped, named with the reason. |
| harness | **31** | RED/GREEN arms driving those checkers with synthetic input. They claim nothing about the shipped surface; they say the checker above can still fail. |

**63 arms, all green.** No arm reports a verdict without reporting its own reach.

## How it got here — three designs, three rounds, two class-stops, two escalations

Designs 1 and 2 were both class-stopped on one shape: *a completeness claim certified by an oracle derived from the machinery it certifies.* Design 1 enumerated the surface by reflection and missed five measured routes that carry a consent sentence to a caller. Design 2 scanned source literals with a hand lexer that could not see inside an interpolation hole — 201 shipped lines outside the net, and the arm chartered to make that falsifiable structurally could not see it either.

- **Escalation 1** (round 1, the enumeration seam) → §4 ruling, Aaron-go 2026-08-25: scan **shipped source literals**, not reflected values. Option B (keep reflection, shrink the claims) declined by name.
- **Escalation 2** (round 1, the interpolation-hole seam) → §4 ruling, Aaron-go 2026-08-25: the completeness claim is certified by a **second, independently written reader**. Options B and C — land the const-concat alone (done separately as #430), and narrow the claims to ship a partial guard — both declined.
- **Round 2** fired the class-stop tripwire → advisor ruling 2026-08-25 ("the by-construction premise splits into two strata"), which split the claim into stratum (i) the enumeration net and stratum (ii) the interpretation layer, pre-decided a contingency for (ii), and recorded its own watch condition: *if silent-green overreach ever reappears in stratum (i), this ruling was wrong and it goes to Aaron as architecture.*
- **Round 3** returned two highs, **both in stratum (i)**, both reproduced at source. The session correctly did **not** fire the pre-decided contingency (wrong stratum — deleting the Class-2 arms fixes neither high) and correctly folded nothing (the §11 three-round cap was spent and the class had recurred). It escalated instead.
- **Advisor ruling, Aaron-go 2026-08-26**: the net's file membership is held against **the compiled artifact's own manifest**. Hand-parsing csproj rejected as the same disease one floor up; a printed boundary declined by name; the contingency stays unfired. Evaluate the PDB route first and record the ground either way. Complete the Append merge. Fix the reader-B blowup. Take the round-3 mediums per the record's triage. **No round 4** — this is directed work under a §4 ruling.

Round-3 findings and the escalation record live in the private `dev/` corpus (`SESSION_HANDOFF_2026-08-25_issue386-round3-STRATUM-I-OVERREACH-escalation.md`); the advisor entries are `dev/advisor/decisions/2026-08-25.md` and `2026-08-26.md`.

**This PR is the directed work. It had no review round of its own and is not entitled to one — Aaron's review on this PR is the gate.**

## H2 — membership from the compiler's receipt

The net's *file* set was a directory walk: match a directory under `src/` to a shipped assembly's name, take every `.cs` under it. Nothing asked the compilation which files it actually compiled. A `<Compile Include="..\prose-shared\X.cs" />` therefore put a shipped, compiled source file **outside** the net — measured, with a false consent claim shipping, **59/59 green and the file and literal counts unmoved**. `GREEN-ROOTS` and `SHIP-DERIVED` compare lists of *directories*; `INV6-AGREE` certifies only that the two readers agree about the files they were *handed*. The two readers agreed perfectly about files nobody gave them.

`MANIFEST-SET` now holds the scanned file set against each shipped assembly's **portable-PDB document table** — the compiler's own receipt for what went into the assembly — in **both directions**.

### The evaluation, run before the arm was written (the ruling required this, either way)

| question | measured answer |
|---|---|
| Are the PDBs readable at build time? | Yes. All three shipped assemblies emit **portable** PDBs (`BSJB`), and they sit beside the assemblies in the generator's own output directory, because the generator project-references all three. CI builds the solution and runs the generator from exactly that directory. |
| What does it cost? | **No new dependency.** `System.Reflection.Metadata` is in the shared framework — the evaluation harness read all four PDBs with an empty `<ItemGroup>`. Contrast Roslyn (reader A), which needed a pinned `PackageReference`. |
| What about generated code? | Exactly **three documents per assembly**, all under `obj/`: `*.GlobalUsings.g.cs`, `*.AssemblyInfo.cs`, and the framework attributes file. No source generator runs in the shipped trees; **zero** documents outside the repo root and **zero** naming a path that does not exist on disk. So the exclusion is one stated rule — a document under `obj/` or `bin/` is build output — and it is `IsBuildOutput`, now the **one home** shared with the file walk, so the net and the receipt cannot drift about what "build output" means. |
| Does the receipt agree with the walk today? | **Exactly.** 73 + 49 + 1 = **123** repo documents against **123** scanned files. Both directions of the set equality are assertable on measured ground, not assumed. |

Two failure modes, both **loud**, both stated in the code:

- A document reaches the table only where the compiler emitted sequence points, so a `.cs` file with no method body could in principle be compiled and absent. That arrives as *"scanned but no manifest names it"* — a red naming the file. Zero such files today.
- **The path-shape condition, measured rather than assumed.** My first draft of this docstring said SourceLink is "set nowhere in this repo". **That was false and is corrected** — SourceLink *is* active (the SDK includes it); it only *anonymises* source paths when `ContinuousIntegrationBuild` is set, which nothing here or in `ci.yml` does. I built with `-p:ContinuousIntegrationBuild=true` to check: every document comes back as `/_/src/…` and `MANIFEST-SET` goes red naming the condition, cause and remedy. That red is **deliberate and is not repaired by stripping the deterministic root** — a build that has anonymised its source paths has removed the fact the arm needs, and reconstructing it from a prefix convention would put membership back on a pattern one floor above where the ruling rejected exactly that. Reporting was collapsed from one line per document (123 of them) to one line per manifest with the count, a specimen, the cause and the remedy.

Above the receipt there is no further enumeration to certify. What remains is **declared human/process residue, printed on every run**: the watchlist's own contents, that the guard is run at all, and free-form truth.

## H1 — the Append merge now covers what it claimed

Settled decision 10's amendment and the run's printed disclosure both said an unbroken run of `Append` calls on one receiver is one sentence. The code read only the **statement** form (`);` between calls), so a fluent chain was outside the merge and a phrase split across it shipped with INV1 green — the arm was green over its own gap. Measured in the shipped trees: **2,479** chained `).Append(` gaps against **257** statement-form ones.

Both forms merge now, told apart by whether a receiver was named (the fluent form's receiver *is* the previous call's result, so identity is structural). `AppendCallBefore` accepts a closing paren in the receiver position, without which a three-link chain read as its first two and dropped the tail silently. A run that switches form mid-way merges too and picks up the right name, because `MergeSentences` carries the run's start forward — which is also what stops the switch merging onto a *different* builder.

**SD10 amendment, as amended here:** a `+`-run, an unbroken `Append` run on one receiver in either written form or across a switch between them, **and** an unbroken `Console.Write` run, are each one sentence. Effect on the shipped trees: 10,099 → **10,056** sentences.

## The round-3 mediums and lows

**Fixed:**

- **M6** — `Console.Write` runs (housecarl-setup talks to a modder through 47 of them) now merge like `Append` runs, with the method name compared so a run that changes method stays two sentences and `WriteLine` stays out. A comment no longer ends an authored sentence: the gap between two literals is stripped of comments before the join test. Both shapes shipped banned phrases green and neither was on the printed `NotInReach` list.
- **M2** — the default-marker pattern had no **left boundary** and its token group only needs a letter, so it could start mid-word: `3d (default)` matched from the `d`, `0x800` from the `x`, `_id` from the `i`. Each was a token no author wrote **and a false RED on `INV4-MARKCOVER`, a class-1 arm**.
- **M7** — every unread parenthetical's residue line claimed "the value is INSIDE the parenthesis, so there is no token in front of it". False for one of the thirty-eight: `RecordsWalk.follow`'s `"*" (default)` has a token in front and is rejected for not starting with a letter. The backward walk now lives in one place that both `MarkerShaped` and the residue printer use.
- **M5** — a nested type is a member of its declaring type *and* a type in `GetTypes()`, so its `[Description]` would enter three censuses twice. **Latent, not live, and measured as such**: the surface has nested types but none carries a type-level `[Description]`, and no census moved. The enumeration is now a function of its type set so `RED-NESTEDTYPE` can drive both branches, which nothing on the live surface can.
- **M4** — both of INV3's counts come off one run pattern, so a recital that stops matching leaves numerator and denominator together (comma-joining a seven-verb recital took "16 read" to 15, green, named nowhere). Counting verb **names** is independent of run shape: **12 descriptions** naming two or more verbs in no run shape at all are now printed with their labels. Comma lists are still not read as recitals — SD16 stands.
- **M3** — `INV4-DEFAULT`'s denominator is its own `(default` pattern, so a default declared in prose was in no census at all: changing `verb defaults to Set` to `Merge` moved no number and stayed green. **30 prose defaults** are counted and named now. They are not compared against a slot (mapping a prose sentence to the member it describes is prose-parsing), but their number sits beside the marked-default coverage so that coverage reads against how many defaults the surface actually declares.
- **M1**, in part — the two `ask again` grounds said nothing honest needs the span. False: *"you may be asked again"* is honest and is refused. The rules stay blunt on purpose and the grounds now say so, naming the settled correction wording as the remedy.
- **The reader-B blow-up** — finding a hole's end *lexes* everything inside it, and the caller then handed the same span to `Scan`, which lexed it again. Every nesting level did its subtree's work twice: **2ⁿ**, with no depth bound and no timeout. Measured: a legal **296-byte file nested 28 deep took the guard from 1.8s to 21.1s**, each further level doubling. `FindHoleEnd` is memoized on `(from, to)` through a dictionary created per `Read` and threaded down — keyed by source index, it is only valid for the string it was built for, so a static cache would answer for the wrong file. **After: depth 28 is 1.64s, depth 60 is 1.97s**, with `INV6-AGREE` reading the extra literals at every level. No depth cap imposed — a cap would put a legal literal outside the net.
- **Four lows, all false sentences in this branch's own prose** (which is the class #386 is about): the lexer put `SkseConfigReferenceExtractor.cs` "one character away from the scanned trees" when it is *in* one; `SlotDefault` claimed every default is spellable when `Render` refuses nulls, unnamed enum values and non-primitives; the header restated #386's headline of *thirteen* stale claims beside an itemisation totalling nineteen (**the discrepancy is in the issue's own body** — its table reads 11 / 6 / 2 under a "13 stale claims" heading — so the counts are given per pass and the discrepancy named rather than propagated); and `PublishCall` claimed to read a project named as a path while its variable alternative made the `$` optional, so `dotnet publish src/housecarl-mcp` captured `src`.
- The RED section header claims every checker is driven synthetically. `INV4-MARKCOVER`'s machinery had no driver; `RED-MARKBOUND` is now it.

**Refused, with reasons:**

1. **`"OUTPUT (default)"` enters the marked-default census as a phantom row.** It does — and the census **prints it with the reason it was not compared**. Excluding ALL-CAPS tokens to suppress it would add a house-style heuristic that could silently skip a real all-caps default: trading a visible phantom for an invisible omission, which is the wrong direction for this guard.
2. **The escalation record's own refusal stands**: reviewer C's finding that `GREEN-ROOTS` shares the `GREEN-*` prefix used by synthetic-input arms is a naming preference, not a defect — the `[c1]` tag already disambiguates it in the output.
3. **Half of M1** — "the remedy this session added is inert for a `NoCompanion` rule". Not reproducible: the remedy sentence is already gated on a rule *having* companions, so it is never emitted for these two.

## The contract catching its own author

Three defects in this session's work were caught by the branch's own arms and discipline, which is the evidence that the contract does something:

- `RED-MARKBOUND` — the arm asserting each rejection names its own cause — **failed on my first classifier**, which reported `"*" (default)` as "stands alone" because "nothing in front" and "a token that is not identifier-shaped" were the same branch.
- `RED-NESTEDTYPE` **failed on my first assertion**, which was wrong-headed; writing the correct one surfaced the precondition the whole skip rests on (`Assembly.GetTypes()` lists nested types in its own right), which is now asserted.
- I claimed the fluent/statement form switch was a boundary that would not merge, and **wrote it into `NotInReach`**. The RED cell said otherwise: it does merge, and correctly. The false disclosure was removed rather than the code changed to match it.

Plus one process defect worth recording: I verified builds with `grep "error CS"`, which **missed a SourceLink `error :`** and let a stale binary be run and reported as a measurement. The build check now requires `Build succeeded.` and the sabotage sweep reports `BUILD-FAIL` explicitly, never as green. (Also inherited from the prior session: the item-1 arm reporting a 0-of-2 denominator when a redundant regex escape broke its input — the Class-2 contract working — and a duplicated `AddRange` caught by reading the diff before staging.)

## Verification — everything re-run at HEAD

**Sabotage sweep, all cells, from a committed state, restore grep-verified after each:**

```
  ok C0-baseline      GREEN  arms=-                      no sabotage — the harness must be able to report GREEN
  ok C1-treeset       RED    arms=GREEN-ROOTS,SHIP-DERIVED,RED-ROOTS,RED-SHIPDERIVE
  ok C2-slotdefault   RED    arms=INV4-DEFAULT           bool esl = true -> false
  ok C3-marker        RED    arms=INV4-DEFAULT           a QUOTED default marker disagreeing with its slot
  ok C4-appendsplit   RED    arms=INV1                   phrase split across a STATEMENT-form Append run
  ok C5-readerb       RED    arms=GREEN-READERS-AGREE    reader B mis-reading a verbatim string that opens on a quote
  ok C7-manifest      RED    arms=MANIFEST-SET           a <Compile Include>-linked file outside the net   [canary]
  ok C8-fluentsplit   RED    arms=INV1                   phrase split across a FLUENT Append chain          [canary]

  8/8 cells as expected
```

The six earlier directed cells are C1–C5 plus item 6, whose cell is the tier accounting itself: **27 c1 / 5 c2 / 31 harness = 63**, exact (was 26 / 5 / 28 = 59). **C0 is the green-direction control** — an all-red sweep proves nothing about a harness that cannot report a pass. **C7 and C8 are the known-RED canaries**, each proven red by hand *before* the script existed, and each also proven **green-before** by rebuilding the same sabotage in a throwaway worktree at the parent commit (C7: 59/59 PASS at `2ee5581` with the false consent claim shipping and counts unmoved; C8: 61/61 PASS at `efceaf9`).

C5's red lands on the **fixture** arm rather than on `INV6-AGREE`, because the shipped trees carry no `@"""` today — measured, and exactly why the fixture exists. My sweep's original expectation was wrong there, not the guard.

**Suites:**

- `description-vocab-guard`: **63 passed, 0 failed → PASS**, in **1.63s** over 123 files / 12,794 literals (334 in interpolation holes) / 10,056 sentences / 567k chars.
- `ci-all`: **130 CI probes, ALL PASS.**
- Manifest census: `housecarl-mcp` 52 documents → 49 repo + 3 generated; `housecarl-core` 76 → 73 + 3; `houseCARL-Setup` 4 → 1 + 3. **123 repo documents against 123 scanned files.**

## CHANGELOG

**No `## Unreleased` line, and I would not add one.** Re-verified mechanically at HEAD: `git diff main...HEAD` over `src/housecarl-core`, `src/housecarl-mcp` and `src/housecarl-setup`, filtered to lines that are not `///` comments, **returns nothing**. Three shipped files are touched — `SweepExclusion.cs`, `WriteVerbs.cs`, `WriteSentences.cs` — and every changed line in them is a docstring. No caller-facing text moves. Everything else on this branch is in `src/housecarl-generator`, which is build-time only and never published.

## Still open, for your review

- The `/_/` path-mapping condition above: correct and loud today, but anyone hardening CI with `ContinuousIntegrationBuild=true` will meet a red. Flagged deliberately rather than papered over — say if you want it handled differently.
- `dev/` housekeeping noticed in passing, not in this branch's warrant: `git worktree list` carries ~13 stale `agent-*` worktrees from the earlier review rounds.
- The prior session's un-cleared boot-rule flags stand: issues #429 and #432 carry no milestone, and rule 2 (deferred empirical items against `EMPIRICAL_LEDGER.md`) is still owed by whoever boots against the run order next.

---

## Aaron's review — three findings, all folded (2026-08-26)

Directed folds, so no new rounds follow them. Each reproduced at source **before** anything was touched, each its own commit, each sabotage RED-checked from a committed state.

### 1. `AppendRun`'s statement-form lookback (:994) — the class fix

Receiver identity, the verb, and adjacency now all come off **reader A's syntax tree** (`RoslynLiteralReader.AppendCalls`, new `AppendCall` record). `AppendGap`, `AppendCallBefore` and the 80-character lookback window are deleted. The chain's head receiver is a node whatever it is spelled like, so a value-arg call before the gap and an indexer-spelled receiver both classify by construction — there is no receiver spelling left to enumerate.

Reproduced first, each cell isolated so it was measured on its own:

```
REPRO-A value-arg call then statement-form run   FAIL
REPRO-B indexer receiver, statement form         FAIL
REPRO-C indexer, DIFFERENT receiver              PASS
```

The syntax route exposed two things the gap regex had been hiding, both folded here:

- **Over-merge.** Pure tree adjacency joined `DialogueTools.AppendTopic:163`'s `" — winner "` to `"  "` across `t.WinnerPlugin`, `'\n'` and `pad`. The statement branch now also requires the first call to be the last thing its chain does (`Node == Outer`) and the second to be the first thing its own does (`Inner < 0`) — cells in both directions, since either condition alone lets it through.
- **The argument must *be* the literal, not merely end with one.** `sb.Append(name == "…")` appends a bool and its literal is read by nobody; admitting those joins 114 further pairs.

**Printed claim and code reach now match in both directions.** The tree sees a boundary the text could not: a one-line `if` body's `Append` no longer merges into the statement after the `if`. The old pattern crossed that whenever the `if` sat before the run's first literal — reaching past a boundary `NotInReach` prints that the guard does not cross.

**Census 10,056 → 10,050**, measured by dumping every sentence span from both commits in throwaway worktrees and diffing the boundary sets. **12 newly merged, 6 no longer merged:**

*Newly merged* — all case (a), where the call before the gap ended on a value rather than a name:
`AssetTools.cs:117`, `RecordsTools.cs:1574`, `SetupTools.cs:93`, `SetupTools.cs:94`, `SkseTools.cs:992`, `StatusTools.cs:69`, `StatusTools.cs:71`, `StatusTools.cs:73`, `StatusTools.cs:74`, `StatusTools.cs:200`, `StatusTools.cs:201`, `WriteTools.cs:1113`

*No longer merged* — each verified at source as a single-statement `if`/`else if` body:
`DialogueTools.cs:303`, `NexusTools.cs:362`, `NifTools.cs:342`, `NifTools.cs:725`, `SkseTools.cs:919`, `UpdateStatusTools.cs:97`

No live instance of case (b) in the trees, so the indexer cell is a RED cell only — stated rather than implied.

### 2. `WriteLine`/`AppendLine` (:922)

Reproduced by flipping the three assertions to the direction the review measured and confirming `RED-APPENDRUN` goes red. A `Line` variant is now the same verb wearing a terminator — reader A records what a call *adds* separately from whether it breaks the line *after* — and the merge admits it in the run's **last position only**: `Write("a"); WriteLine("b")` merges; `WriteLine("a"); Write("b")`, `WriteLine`→`WriteLine` and `AppendLine`→`Append` stay two sentences.

Both false ground sentences corrected (`consoleLineSrc` and `RED-APPENDRUN`'s copy), and the `NotInReach` entry now states the rule so the printed boundary and the code agree.

New arm, held against the merge's **actual output** rather than the predicate that decides it — `AppendRun` split into `Contiguous` + `!first.EndsLine` so `LineClosedGaps` reuses the same test rather than copying it:

```
[PASS] [c1] GREEN-LINEBREAK  every run a Line call already CLOSED stayed two sentences
                             (35 WriteLine and 3 AppendLine gap(s) in the shipped trees, none merged)
```

**38, not the review's 58** — the review counted the textual gap shape; this counts pairs satisfying every *other* condition of a run, which is the set the rule can decide. Both figures are recorded in the docstring. The count is printed rather than pinned (a new `WriteLine` pair is ordinary authoring), but the arm **reds if it ever reaches zero**, so it cannot pass by having nothing to hold — its own sabotage cell proves that. **The shipped census does not move**: zero plain→Line gaps in the trees today.

### 3. `CiAll.cs:200`

Confirmed at source: the registry said "thirteen stale claims that three hand sweeps found" while the probe header, added in the same commit, itemises those three as 11 + 6 + 2 and declines to restate the thirteen. The registry now carries the per-pass itemisation and restates no total.

### Verification after the last fold

- `description-vocab-guard`: **64 passed, 0 failed → PASS** (was 63 — `GREEN-LINEBREAK` is new). Tiers **28 c1 / 5 c2**. Census 123 files / 12,794 literals (334 in holes) / **10,050 sentences** / 567k chars.
- **Sabotage sweep: 16/16 RED**, each restored from a committed state and the restore verified by grepping for a string the fold introduced. `canary-receiver` is the known-RED canary, proven red by hand before the script existed.

```
  canary-receiver    RED   two different receivers must not merge
  verb               RED   a run that changes verb is two sentences
  fluent-link        RED   a fluent chain is one run
  stmt-following     RED   an intervening statement breaks the run
  stmt-outer         RED   a value the first statement ends with breaks the run
  stmt-head          RED   a value the next statement begins with breaks the run
  line-closes        RED   a Line call CLOSES a run and cannot continue one
  line-gap-scan      RED   GREEN-LINEBREAK reds when it has no gap left to hold
  chain-head-walk    RED   the receiver is the chain HEAD, not the link in front of the literal
  arg-count          RED   a second argument breaks the run
  arg-is-literal     RED   an expression that merely ENDS on a literal is not a run link
  arg-interpolated   RED   an interpolated argument IS a run link
  head-text          RED   the receiver TEXT is what is compared
  write-verb         RED   Console.WriteLine is a run link too
  appendline-verb    RED   AppendLine is a run link too
  endsline-flag      RED   WriteLine is recorded as BREAKING the line
```

- **CHANGELOG re-verified mechanically at HEAD**, unchanged: `git diff -U0 main...HEAD` over `src/housecarl-core` and `src/housecarl-mcp`, filtered to lines that are not `//`/`///` comments, returns nothing, and no `plugin/` file is touched. No caller-facing text moves, so no `## Unreleased` line.

### DIFF
--- a/src/housecarl-generator/CiAll.cs
+++ b/src/housecarl-generator/CiAll.cs
@@ -198,8 +198,10 @@ public static class CiAll
         ("wire-names-guard", WireNamesProbe.RunGuard),
         // Caller-facing prose vocabulary (#386): nothing read the WORDS of the [Description] attributes a model
         // reads or of the consent prompts a modder reads, so the in-place consent fix left thirteen stale claims
-        // that four hand sweeps found in four homes. (wire-names-guard does reach the description surface, but for
-        // the brace shape declaration it parses out of one — never for what the sentence around it claims.) The net is every string literal in the three SHIPPED trees, read TWICE —
+        // that three hand sweeps found surface by surface, and a fourth sweep — run by vocabulary instead — found
+        // fourteen more in other homes. (wire-names-guard does reach the description surface, but for the brace
+        // shape declaration it parses out of one — never for what the sentence around it claims.)
+        // The net is every string literal in the three SHIPPED trees, read TWICE —
         // by Roslyn and by an independently written lexer — with INV6-AGREE holding the two against each other, so
         // a reader that stops early turns red instead of quietly shrinking the net. Self-contained (no corpus, no
         // MO2 instance); it reads source files, so it must run from the repo root.
--- a/src/housecarl-generator/CiAll.cs
+++ b/src/housecarl-generator/CiAll.cs
@@ -196,6 +196,13 @@ public static class CiAll
         // initializer, which never consults the attribute). Reflection-discovered, so a new spec object enrols
         // itself. Self-contained (no corpus, no MO2 instance).
         ("wire-names-guard", WireNamesProbe.RunGuard),
+        // Caller-facing prose vocabulary (#386): the [Description] attributes a model reads and the consent prompts
+        // a modder reads had no guard behind them, so the in-place consent fix left thirteen stale claims that four
+        // hand sweeps found in four homes. The net is every string literal in the three SHIPPED trees, read TWICE —
+        // by Roslyn and by an independently written lexer — with INV6-AGREE holding the two against each other, so
+        // a reader that stops early turns red instead of quietly shrinking the net. Self-contained (no corpus, no
+        // MO2 instance); it reads source files, so it must run from the repo root.
+        ("description-vocab-guard", DescriptionVocabularyGuardProbe.RunGuard),
         ("nullarm-guard", NullArmGuardProbe.RunGuard),
         ("formlink-null-guard", FormLinkNullProbe.RunGuard),
         ("formlink-remove-guard", FormLinkRemoveProbe.RunGuard),
--- a/src/housecarl-generator/CiAll.cs
+++ b/src/housecarl-generator/CiAll.cs
@@ -196,9 +196,10 @@ public static class CiAll
         // initializer, which never consults the attribute). Reflection-discovered, so a new spec object enrols
         // itself. Self-contained (no corpus, no MO2 instance).
         ("wire-names-guard", WireNamesProbe.RunGuard),
-        // Caller-facing prose vocabulary (#386): the [Description] attributes a model reads and the consent prompts
-        // a modder reads had no guard behind them, so the in-place consent fix left thirteen stale claims that four
-        // hand sweeps found in four homes. The net is every string literal in the three SHIPPED trees, read TWICE —
+        // Caller-facing prose vocabulary (#386): nothing read the WORDS of the [Description] attributes a model
+        // reads or of the consent prompts a modder reads, so the in-place consent fix left thirteen stale claims
+        // that four hand sweeps found in four homes. (wire-names-guard does reach the description surface, but for
+        // the brace shape declaration it parses out of one — never for what the sentence around it claims.) The net is every string literal in the three SHIPPED trees, read TWICE —
         // by Roslyn and by an independently written lexer — with INV6-AGREE holding the two against each other, so
         // a reader that stops early turns red instead of quietly shrinking the net. Self-contained (no corpus, no
         // MO2 instance); it reads source files, so it must run from the repo root.
--- a/src/housecarl-generator/CiAll.cs
+++ b/src/housecarl-generator/CiAll.cs
@@ -197,10 +197,14 @@ public static class CiAll
         // itself. Self-contained (no corpus, no MO2 instance).
         ("wire-names-guard", WireNamesProbe.RunGuard),
         // Caller-facing prose vocabulary (#386): nothing read the WORDS of the [Description] attributes a model
-        // reads or of the consent prompts a modder reads, so the in-place consent fix left thirteen stale claims
-        // that three hand sweeps found surface by surface, and a fourth sweep — run by vocabulary instead — found
-        // fourteen more in other homes. (wire-names-guard does reach the description surface, but for the brace
-        // shape declaration it parses out of one — never for what the sentence around it claims.)
+        // reads or of the consent prompts a modder reads, so the in-place consent fix left stale claims that took
+        // four sweeps to clear — three run surface by surface, finding 11 on the acknowledge= parameters, 6 in the
+        // WriteTools lane parentheticals and 2 in the handshake builders, and a fourth run repo-wide by VOCABULARY
+        // instead, finding fourteen more in homes the first three had no reason to look at. (The counts are given
+        // per pass, as the probe's own header gives them: #386's headline totals its first three as thirteen,
+        // which its table contradicts, so nothing here restates a total.) (wire-names-guard does reach the
+        // description surface, but for the brace shape declaration it parses out of one — never for what the
+        // sentence around it claims.)
         // The net is every string literal in the three SHIPPED trees, read TWICE —
         // by Roslyn and by an independently written lexer — with INV6-AGREE holding the two against each other, so
         // a reader that stops early turns red instead of quietly shrinking the net. Self-contained (no corpus, no
--- a/src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
+++ b/src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
@@ -62,9 +62,10 @@ namespace HousecarlGenerator;
 /// <list type="bullet">
 ///   <item><b>SOURCE</b> — every string-literal SENTENCE in the three shipped trees (<c>housecarl-mcp</c>,
 ///         <c>housecarl-core</c>, <c>housecarl-setup</c>), a sentence being a maximal run of adjacent literals
-///         joined by <c>+</c> OR a run of consecutive <c>Append</c> calls on one receiver: the unit an author
-///         writes. This answers ABSENCE (INV1) — a phrase absent from every literal is absent from every string
-///         built out of them by those two shapes. Assembly across control flow or through a helper is a DECLARED
+///         joined by <c>+</c> OR a run of consecutive <c>Append</c> / <c>Write</c> calls on one receiver,
+///         optionally finished by the <c>Line</c> form of that verb: the unit an author writes. This answers
+///         ABSENCE (INV1) — a phrase absent from every literal is absent from every string built out of them by
+///         those two shapes. Assembly across control flow or through a helper is a DECLARED
 ///         boundary, printed on every run (<see cref="NotInReach"/>), not a claim this quietly covers.</item>
 ///   <item><b>SURFACE</b> — the compiled <c>[Description]</c> attributes of the tool assembly. This answers
 ///         PRESENCE and COMPLETENESS (INV3/INV4): "which verbs does a caller read" needs the text assembled, and
@@ -250,8 +251,10 @@ public static class DescriptionVocabularyGuardProbe
         "the READMEs, the shipped skills, plugin/CHANGELOG.md, and the plugin / marketplace JSON metadata (not in a scanned tree)",
         "a phrase split around a VALUE — \"shown \" + n + \"once\", or across an interpolation hole (no fragment carries it)",
         "prose assembled ACROSS CONTROL FLOW or through a helper — a run of Append calls broken by an 'if', or a "
-            + "sentence one method starts and another finishes. A +-run and an unbroken Append run on one receiver "
-            + "ARE read as one sentence, written as separate statements or as a fluent chain; deciding which "
+            + "sentence one method starts and another finishes. A +-run and an unbroken Append or Write run on one "
+            + "receiver ARE read as one sentence, written as separate statements or as a fluent chain, and a run "
+            + "FINISHED with the Line form of its verb is one too, because that break lands after the last half; "
+            + "a Line call does not CONTINUE a run, since then the break falls between the halves. Deciding which "
             + "conditional arms run together is dataflow analysis, not a merge rule, so this edge is declared "
             + "rather than guessed at",
         "third-party library messages surfaced to a caller (not ours to author, and not ours to fix)",
@@ -698,6 +701,8 @@ public static class DescriptionVocabularyGuardProbe
         // Every file the net ENUMERATED, recorded before it is read: membership is about which files the net
         // reaches, and a file that then fails to read is INV6-PARSE's red, not a hole in the file set.
         var scannedFiles = new List<string>();
+        var lineProblems = new List<string>();
+        int lineClosedAppend = 0, lineClosedWrite = 0;
         long chars = 0;
         int files = 0, literals = 0, inHoles = 0;
 
@@ -738,12 +743,26 @@ public static class DescriptionVocabularyGuardProbe
                     agreeProblems.AddRange(disagreements);
                 }
 
-                foreach (var s in MergeSentences(text, a, calls))
+                var merged = MergeSentences(text, a, calls);
+                foreach (var s in merged)
                 {
                     if (string.IsNullOrWhiteSpace(s.Text)) continue;
                     sentences.Add(new Sentence($"{label}:{s.Line}", s.Text));
                     rootSentences++;
                     chars += s.Text.Length;
+                }
+                // The Line rule, held against what the merge actually PRODUCED rather than against the predicate
+                // that decides it. A sentence carries the START of its head, so a literal that opens a sentence
+                // was not merged into what came before it; a Line-closed gap whose second literal is missing from
+                // that set was joined across a line break the modder can see.
+                var opens = merged.Select(m => m.Start).ToHashSet();
+                foreach (var (lit, adds) in LineClosedGaps(a, calls))
+                {
+                    if (adds == "Append") lineClosedAppend++; else lineClosedWrite++;
+                    if (!opens.Contains(lit.Start))
+                        lineProblems.Add($"{label}:{lit.Line} — the call before this one already broke the line, and the two "
+                                       + "were merged anyway: the phrase this reports would be split across a line break on "
+                                       + "the caller's screen");
                 }
                 literals += a.Count;
                 rootHoles += a.Count(l => l.Depth > 0);
@@ -758,6 +777,19 @@ public static class DescriptionVocabularyGuardProbe
         Console.WriteLine($"        total: {files} file(s), {literals} literal(s) ({inHoles} inside interpolation holes), {sentences.Count} sentence(s), {chars / 1000}k char(s)");
         Console.WriteLine("        NOT in reach (the by-construction claim is exactly as honest as this list):");
         foreach (var n in NotInReach) Console.WriteLine($"          · {n}");
+
+        // A run FINISHED by a Line call is one sentence (the break lands after it); a run that a Line call already
+        // closed is two. The second half is the one with live instances in these trees, so it gets an arm over the
+        // merge's own output. The count is printed rather than pinned: a new WriteLine pair is ordinary authoring,
+        // and what must hold is that none of them merged — but if the count ever reaches zero this arm is asserting
+        // nothing about nothing, and says so instead of passing quietly.
+        int lineClosed = lineClosedAppend + lineClosedWrite;
+        if (lineClosed == 0)
+            lineProblems.Add("no run in the shipped trees is closed by a Line call any more — this arm has nothing left "
+                           + "to hold, and a merge that started crossing line breaks would now pass it silently");
+        Check($"GREEN-LINEBREAK  every run a Line call already CLOSED stayed two sentences ({lineClosedWrite} WriteLine "
+            + $"and {lineClosedAppend} AppendLine gap(s) in the shipped trees, none merged)",
+            lineProblems.Count == 0, lineProblems, tier: Tier.Construction);
 
         var scanned = roots.Select(r => Path.GetFileName(r)!).ToList();
         rootProblems.AddRange(TreeSetMismatch(scanned));
@@ -961,17 +993,44 @@ public static class DescriptionVocabularyGuardProbe
     /// cross. Six shipped literals stopped being merged that way when the receiver moved onto the tree, each of
     /// them the statement after a one-line <c>if</c> body.</para></summary>
     static bool AppendRun(IReadOnlyDictionary<int, AppendCall> calls, SourceLiteral prev, SourceLiteral next)
+        // A call that breaks the line can FINISH a run but never continue one: the break lands after its own
+        // text, so Write("a"); WriteLine("b"); reads "ab" on one line, and WriteLine("a"); Write("b"); does not.
+        => Contiguous(calls, prev, next, out var first) && !first.EndsLine;
+
+    /// <summary>Everything a run needs EXCEPT the line rule: the same verb, on the same receiver, with nothing
+    /// between. Split out from <see cref="AppendRun"/> so that <see cref="LineClosedGaps"/> can ask for the pairs
+    /// that differ from a run in exactly one respect, without a second copy of the test to drift from this
+    /// one. <paramref name="first"/> is the call the earlier literal belongs to.</summary>
+    static bool Contiguous(IReadOnlyDictionary<int, AppendCall> calls, SourceLiteral prev, SourceLiteral next,
+                           out AppendCall first)
     {
         // Keyed by the literal's END: a merged run carries its TAIL's end forward, so this looks up the call the
         // run last made rather than the one it started with.
+        first = default;
         if (!calls.TryGetValue(prev.End, out var a) || !calls.TryGetValue(next.End, out var b)) return false;
-        // The same METHOD both sides. Append and Write each add text with nothing between, but a run that changes
+        first = a;
+        // The same VERB both sides. Append and Write each add text with nothing between, but a run that changes
         // from one to the other is two different calls, and joining them would manufacture text nobody reads.
-        if (!string.Equals(a.Method, b.Method, StringComparison.Ordinal)) return false;
+        if (!string.Equals(a.Adds, b.Adds, StringComparison.Ordinal)) return false;
         if (!string.Equals(a.Receiver, b.Receiver, StringComparison.Ordinal)) return false;
         return b.Inner == a.Node
             || (a.Statement >= 0 && b.Statement >= 0 && a.Following == b.Statement
                 && a.Node == a.Outer && b.Inner < 0);
+    }
+
+    /// <summary>The adjacent literal pairs that are a run in every respect except that the FIRST call already
+    /// broke the line. These must stay two sentences, and there are enough of them in the shipped trees for that
+    /// to be worth checking against the merge's actual output rather than against the predicate that decides it —
+    /// which is what <c>GREEN-LINEBREAK</c> does.</summary>
+    static List<(SourceLiteral Next, string Adds)> LineClosedGaps(
+        List<SourceLiteral> lits, IReadOnlyDictionary<int, AppendCall> calls)
+    {
+        var outp = new List<(SourceLiteral, string)>();
+        var top = lits.Where(l => l.Depth == 0).OrderBy(l => l.Start).ToList();
+        for (int i = 1; i < top.Count; i++)
+            if (Contiguous(calls, top[i - 1], top[i], out var a) && a.EndsLine)
+                outp.Add((top[i], a.Adds));
+        return outp;
     }
 
     /// <summary>Adjacent TOP-LEVEL literals are ONE sentence when the author wrote them as one. Two shapes count:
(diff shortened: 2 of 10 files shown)


src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
```diff
@@ -0,0 +1,2365 @@
...
+        var window = src[Math.Max(0, prev.Start - 80)..prev.Start];
+        var before = AppendCallBefore.Match(window);
+        if (!before.Success) return false;
+        // The same METHOD both sides. Append and Write each add text with nothing between, but a run that changes
+        // from one to the other is two different calls, and joining them would manufacture text nobody reads.
+        if (!string.Equals(before.Groups[2].Value, gap.Groups[2].Value, StringComparison.Ordinal)) return false;
+        // Fluent gap: no receiver named, none needed — the chain IS the receiver.
+        if (!gap.Groups[1].Success) return true;
+        // Statement gap: the names must match. In front of an already-merged run that is the chain's head.
+        return before.Groups[1].Success
```
Avick3110: **`AppendRun` drops a statement-form run whenever the previous literal's own call is not preceded by a bare receiver name.**

`AppendCallBefore` yields a captured NAME only when the receiver is a dotted identifier. When the lookback lands on the `\)` alternative instead — which is what the middle of a chain looks like *and* what a preceding call with a non-literal argument leaves — `before.Groups[1].Success` is false, so a statement-form gap that does name a receiver is refused here and the run is not merged. An indexer-spelled receiver fails the same way, because `[A-Za-z_][A-Za-z0-9_.]*` cannot match `cells[i].Sb`.

Measured by replicating `MergeSentences`/`AppendRun`/`StripComments` verbatim over Roslyn's literal list:

```
value-arg call then statement-form run (same receiver): carries 'shown once' = False
      [this prompt is shown ]
      [once.]
all fluent with a value arg first:                      carries 'shown once' = True
indexer receiver, statement form:                       carries 'shown once' = False
plain statement-form control:                           carries 'shown once' = True
```

i.e. `sb.Append(count).Append("this prompt is shown ");` / `sb.Append("once.");` puts the phrase in front of a caller with INV1 green — no value sits *between* the two halves, so this is not the declared "phrase split around a VALUE" boundary, and `NotInReach` prints the opposite claim on every run: *"an unbroken Append run on one receiver ARE read as one sentence, written as separate statements or as a fluent chain"*.

30 gaps of the `)`-headed shape already exist in the scanned trees (they happen to carry no watchlist phrase today).

The fix is either to walk back to the chain's head for the name when the lookback lands on `)`, or to narrow the printed disclosure to what the pattern actually reaches.

src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
```diff
@@ -0,0 +1,2365 @@
...
+    /// printed claim TRUE rather than narrowing it, which is why the ruling directed the completion.</para>
+    /// <para>Anything else still breaks the run — a second argument, an intervening statement, a different
+    /// method, a different named receiver.</para></summary>
+    /// <para><b>The METHOD is captured too</b>, because <c>Append</c> is not the only call that puts a literal in
+    /// front of a caller with nothing between: <c>housecarl-setup</c> talks to a modder in 47 <c>Console.Write*</c>
+    /// calls, and consecutive <c>Console.Write("…"); Console.Write("…");</c> concatenates on the modder's screen
+    /// exactly as an Append run concatenates in a builder. Both are read; the two names must MATCH, so a run that
+    /// changes method is two sentences. <c>WriteLine</c> is deliberately not in the set — it puts a line break
+    /// between the two halves, so they are not one sentence on the screen either.</para></summary>
+    static readonly Regex AppendGap = new(
```
Avick3110: **The `WriteLine`/`AppendLine` exclusion rests on a false statement, and reopens the hole the merge exists to close.**

The ground given here and at `RED-APPENDRUN`'s `consoleLineSrc` is *"WriteLine ... puts a line break between the two halves, so they are not one sentence on the screen either."* That is true only for `WriteLine` → `Write`. In the other direction the break lands **after** the second half:

```csharp
Console.Write("this prompt is shown ");
Console.WriteLine("once.");   // the modder reads: this prompt is shown once.
```

Replicating the merge verbatim, both of these come back as two sentences with `Carries("shown once") = False`:

```
Console.Write then Console.WriteLine: carries 'shown once' = False
Append then AppendLine:              carries 'shown once' = False
```

So any run finished with the `Line` variant — the commoner way to end a console or builder run — is outside INV1 with nothing on the printed `NotInReach` list saying so. (No live instance in the trees today: the 53 `WriteLine` and 5 `AppendLine` gaps I measured are all Line→Line, which correctly must not merge.)

Admitting `WriteLine`/`AppendLine` as the **last** call of a run — matching only in the gap's method position, never in `AppendCallBefore` — would close it while keeping `WriteLine("a"); Write("b")` two sentences.

src/housecarl-generator/CiAll.cs
```diff
@@ -196,6 +196,16 @@ static readonly (string Name, Func<string[], int> Run)[] Probes =
         // initializer, which never consults the attribute). Reflection-discovered, so a new spec object enrols
         // itself. Self-contained (no corpus, no MO2 instance).
         ("wire-names-guard", WireNamesProbe.RunGuard),
+        // Caller-facing prose vocabulary (#386): nothing read the WORDS of the [Description] attributes a model
+        // reads or of the consent prompts a modder reads, so the in-place consent fix left thirteen stale claims
```
Avick3110: `thirteen` is the headline this PR's own probe declines to restate. `DescriptionVocabularyGuardProbe.cs:20-27` itemises the three hand sweeps as 11 + 6 + 2 = **19** and says so explicitly:

> the per-pass counts are #386's own table; its headline totals those three passes as thirteen, which is a discrepancy in the issue rather than something to average here — the counts are given per pass so nothing restates a total its own itemisation contradicts

The PR body lists this as one of the four fixed lows, but the fix landed only in the probe docstring; the same sentence is added unchanged here, so `ci-all`'s registry and the probe's own header now disagree about the count in the same commit.

src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
Avick3110: Folded — the class fix, not the regex patch. Receiver identity, the verb, and adjacency all come off reader A's syntax tree now (`RoslynLiteralReader.AppendCalls`, new `AppendCall` record); `AppendGap`, `AppendCallBefore` and the 80-character lookback window are gone. The chain's head receiver is a node whatever it's spelled like, so both of your cells classify by construction rather than by two more alternations, and there's no receiver spelling left to enumerate.

Both repros reproduced at source before anything was touched — isolated as separate arms so each was measured on its own:

```
REPRO-A value-arg call then statement-form run   FAIL
REPRO-B indexer receiver, statement form         FAIL
REPRO-C indexer, DIFFERENT receiver              PASS
```

Pinned in `RED-APPENDRUN` by your two cells plus the negative that keeps them honest (`cells[i].Sb` vs `other[i].Sb` must still not merge).

**Your fix exposed a second thing, and it went the other way.** Dropping the gap regex for pure tree adjacency initially over-merged: `DialogueTools.AppendTopic:163` joined `" — winner "` to `"  "` across `t.WinnerPlugin`, `'\n'` and `pad`. The statement branch now also requires the first call to be the last thing its chain does (`Node == Outer`) and the second to be the first thing its own does (`Inner < 0`) — both directions driven by cells, since either condition alone lets it through. A third: the argument must *be* the literal, not merely end with one — `sb.Append(name == "…")` appends a bool and its literal is read by nobody; admitting those joins 114 further pairs.

**Printed claim vs code reach, both directions.** The tree also sees a boundary the text could not: a one-line `if` body's Append no longer merges into the statement after the `if`. The old pattern crossed that whenever the `if` sat before the run's first literal — reaching past a boundary `NotInReach` prints that the guard does not cross. Six shipped literals; all six verified at source as single-statement `if`/`else if` bodies (`DialogueTools:303`, `NexusTools:362`, `NifTools:342`, `NifTools:725`, `SkseTools:919`, `UpdateStatusTools:97`).

**Census 10,056 → 10,050**, measured by dumping every sentence span from both commits in throwaway worktrees and diffing the boundary sets — 12 newly merged, 6 no longer merged:

*newly merged (all case (a) — the call before ended on a value, not a name):*
`AssetTools.cs:117`, `RecordsTools.cs:1574`, `SetupTools.cs:93`, `SetupTools.cs:94`, `SkseTools.cs:992`, `StatusTools.cs:69`, `StatusTools.cs:71`, `StatusTools.cs:73`, `StatusTools.cs:74`, `StatusTools.cs:200`, `StatusTools.cs:201`, `WriteTools.cs:1113`

*no longer merged (the if-body boundary above):* the six listed already.

No live instance of case (b) in the trees, so the indexer cell is a RED cell only — stated rather than implied.

Sabotage cells pinning this fold, each RED from a committed state: `canary-receiver`, `verb`, `fluent-link`, `stmt-following`, `stmt-outer`, `stmt-head`, `chain-head-walk`, `arg-count`, `arg-is-literal`, `arg-interpolated`, `head-text`.

src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
Avick3110: Folded. Reproduced first by flipping the three assertions to the direction you measured and confirming `RED-APPENDRUN` goes red — `Write`→`WriteLine`, `Append`→`AppendLine` and the chained form all currently refused to merge, exactly as you had it.

A `Line` variant is now the same verb wearing a terminator. Reader A records what a call *adds* (`Append`/`Write`) separately from whether it breaks the line *after* (`EndsLine`), and the merge admits the Line form in the run's last position only:

- `Write("a"); WriteLine("b")` → one sentence
- `WriteLine("a"); Write("b")` → two
- `WriteLine("a"); WriteLine("b")` → two
- `AppendLine("a"); Append("b")` → two

**Both false ground sentences corrected** — the `consoleLineSrc` fixture comment and `RED-APPENDRUN`'s copy. The `NotInReach` entry now states the rule too, so the printed boundary and the code agree: a run *finished* with the Line form is one sentence because the break lands after the last half; a Line call does not *continue* one because then it lands between.

**The Line→Line count, held against the merge's actual output rather than the predicate that decides it.** New `GREEN-LINEBREAK` arm: `AppendRun` was split into `Contiguous` (same verb, same receiver, nothing between) plus `!first.EndsLine`, so `LineClosedGaps` asks for the pairs that differ from a run in exactly one respect without a second copy of the test to drift from the first. For each, it checks the second literal still *opens* a sentence in the merged output.

```
[PASS] [c1] GREEN-LINEBREAK  every run a Line call already CLOSED stayed two sentences
                             (35 WriteLine and 3 AppendLine gap(s) in the shipped trees, none merged)
```

Two notes on that number, rather than quietly shipping a different one. **38, not your 58** — you counted the textual gap shape; this counts pairs satisfying every *other* condition of a run, which is the set the rule can actually decide. Both figures are recorded in the docstring. And the count is printed, not pinned: a new `WriteLine` pair is ordinary authoring, so pinning 38 would red on correct edits. What is asserted is that none of them merged — plus a red if the count ever reaches zero, so the arm can never pass by having nothing to hold. That last property has its own sabotage cell (`line-gap-scan`).

**The shipped census does not move.** The trees hold zero plain→Line gaps today, so this closes a real hole in the rule while changing no shipped sentence — as you said, no live instance.

Sabotage cells, each RED: `line-closes`, `line-gap-scan`, `write-verb`, `appendline-verb`, `endsline-flag`, `verb`.

src/housecarl-generator/CiAll.cs
Avick3110: Folded. Confirmed at source first — `CiAll.cs:200` said "thirteen stale claims that three hand sweeps found", while `DescriptionVocabularyGuardProbe.cs:20-27`, added in the same commit, itemises those three as 11 + 6 + 2 and explicitly declines to restate the thirteen.

The registry now carries the per-pass itemisation and restates no total, matching the probe header:

> three run surface by surface, finding 11 on the `acknowledge=` parameters, 6 in the `WriteTools` lane parentheticals and 2 in the handshake builders, and a fourth run repo-wide by VOCABULARY instead, finding fourteen more in homes the first three had no reason to look at. (The counts are given per pass, as the probe's own header gives them: #386's headline totals its first three as thirteen, which its table contradicts, so nothing here restates a total.)

Issue #435, opened 2026-08-26
[Bug] merge_plugins de-localizes a localized donor silently — no note, other languages and the plugin's nature dropped without a word


Avick3110:
## Summary

`merge_plugins` building M from a localized donor inlines the donor's resolved string values and emits a non-localized M — the donor's `.STRINGS` set no longer describes the output, any other language it shipped is not in the result, and **the report says nothing**. This is the compact lane's old behaviour, surviving in the merge lane.

## Ground (measured, 2026-08-26)

Found during the pre-PR review rounds for the #368/#373 branch (`claude/localized-write`), in a path that branch does not otherwise touch — the outside-the-branch shape (#324 class), filed rather than folded.

From that branch's #373 disposition table (verified against source at `WritePatchBuilder.cs:2714` on the branch; the lane is pre-existing on `main`): M is never flagged localized, so no string tables are emitted and #373's index mispairing is unreachable — donors are opened strings-aware, values inline intact. The defect is not corruption; it is a **silent change of the output's nature** with a silent drop of any non-active language, reported as an ordinary success.

## Relationship to existing issues

- Sibling of #363 (merge drops the donor's ESL/ESM header flag and the report never says so) — same lane, same silent-header-fact class.
- Sibling of the compact lane's loud note (the compact new-file lane states the de-localization in its report; merge states nothing).
- NOT #373: no tables are emitted, nothing mispairs.

## Fix shape (not verified)

The compact lane's loud-note treatment applied to merge: state the de-localization and what it drops, with the count and list computed from the actual output, not the source folder.

## Provenance

Review-round finding, review session 2026-08-26; confirmed against source by that session; filed by the advisor lane on Aaron's go (2026-08-26). Unrouted and unmilestoned — lands at the next routing stroke.

Avick3110:
Routed (2026-08-27, Aaron-go, routing delegated to the advisor lane): **RUN_ORDER row 8 / the W5 charter, FIRM**, beside #363 — same merge lane, same silent-header-fact class; #363's carry-vs-report fork and this issue's loud note are one conversation at the charter. Fix shape as filed (the compact lane's loud-note treatment, computed from the output). Milestoned houseCARL 2.0 in the same stroke (boot rule 4).

PR #436, opened 2026-08-26
Refuse every localized in-place write off the mod in memory, and cut the localized P-prime

### LINKED ISSUE #368: [Bug] compact_plugin's repoint lane blanks a localized external referencer, permanently and in place

### LINKED ISSUE #373: [Bug] WriteInPlace commits the plugin without the string tables its own serialize emitted — every localized value shifts to another record

### LINKED ISSUE #374: [Bug] compact_plugin's external-referencer refusal offers a repoint that a localized referencer will refuse

Avick3110:
Fixes #368
Fixes #373
Fixes #374

## What this branch is worth

**Read this first, because it is not what the engagement set out to build.** The deliverable is a
**refusal that cannot fail open, and read-side correctness**. It is not new write capability — every new
write capability designed in this engagement was built, measured working, reviewed, and then cut.

`housecarl_compact_plugin`, `housecarl_apply`, `housecarl_create`, `housecarl_forward` and
`housecarl_remove` no longer rewrite a localized plugin in place, whatever arrangement its `.STRINGS` files
are in — and the refusal now tells the modder where their text actually lives, in that arrangement's own
words. A plugin houseCARL cannot *read* refuses too, saying so.

### The engagement's own frequency sweep, on Aaron's real load order

`localized-shape-sweep` over ARR 2.0 — **3802 plugins scanned, 63 flagged localized (1.7%)**:

| Arrangement | Count |
|---|---|
| complete loose set beside the plugin | **1** |
| a language missing a table kind | 0 |
| loose set duplicated in game-Data | 0 |
| strings inside a `.bsa` | 31 |
| strings in game-Data only | 0 |
| houseCARL can find no source (MO2's VFS) | 31 |

The single complete-loose-set plugin is `ksws07_quest.esm`. **Every write arm designed in this engagement
served that one plugin of sixty-three.** That number is why both cuts below were the right call, and it is
the number that prices this PR.

The 31 "can find no source" are Creation Club plugins whose `.bsa` sits in a *different* mod folder — only
MO2's VFS puts them together, which no path houseCARL walks can see. That is #371's class with real
numbers, routed to W5, and it is worth more than this engagement's write-side win.

## Two §4 stops, both taken rather than folded through

**Round 1 → the in-place localized write was CUT.** Four independent reviewers converged on the same
machinery: the crash-window recovery half. Its staging directory was per-FOLDER while its backup protocol
assumed per-PLUGIN, so an in-place write of any *other* plugin in the same mod folder adopted the
interrupted plugin's staged tables and recursively deleted the backups — #373 reproduced by the code that
existed to prevent it. Its recovery instruction, followed after the plugin swap, produced the scramble and
then committed it permanently. And the guard that had cleared it was vacuous on the point that mattered:
deleting the backup copy outright left all 131 arms green.

**Round 2 → a class stop.** The round-1 fold had itself introduced the branch's highest finding: the
refusal had stopped deciding off `targetMod.UsingLocalization` and started deciding off a re-read of the
destination file, whose fault path answered "not localized". A reviewer reproduced the write proceeding with
the destination locked by a second process — the plugin replaced, every value empty, the emitted tables
stranded. Four of round 2's confirmed defects were in the reviewing session's own fold, and the sentence
class from round 1 recurred in volume. The rounds stopped at two under §5 #11's recurring-class rule rather
than folding a third time.

**Both cut arms were advisor-endorsed before review killed them.** The crash-window ordering was an advisor
ruling with a hole closed on its own directive; Q2-A was ratified by the advisor and made unconditional by
Aaron. Neither was a shortcut somebody took — they were designed, reviewed, and measured working, and
review still found them unsound. Recorded here because "we ruled this" was not sufficient either time.

## What this PR actually changes

**The refusal cannot fail open.** `WriteInPlace` decides off `targetMod.UsingLocalization` — a fact about
the mod already in memory, which cannot fail to be read. The shape lookup that re-opens the file on disk
supplies only the sentence.

**`WriteEngine.PluginIsLocalized` returns three answers, not two.** `Localized` / `NotLocalized` /
`Unreadable`. Its bool form answered `false` on any fault, justified by naming the write's choke point as
its backstop — but that read *had become* the choke point's own decision, so the net it named was the net it
was inside. The service pre-flights that share `LocalizedStrings.RefusalFor` now fail closed on `Unreadable`:
a plugin houseCARL could not read refuses, saying that, instead of being waved through. All four in-place
pre-flights were audited against the new contract; none needed a design change.

**Q2-A is cut.** A compacted P′ is de-localized again, with the loud note. Aaron's ground is a fact about
the world, not the code: multi-language mods do not exist in the wild — Nexus ships single-language
translations — so the language a read resolves *is* the mod's language, and carrying a rewritten table set
forward buys nothing to offset the surface it costs. Most of round 2's remaining defects lived in that
output path. `WriteOwnedOutput` and `LocalizedTableCommit.cs` go with it.

**`RemapEngine.OpenMutable` is dropped, not documented as dead.** It was #368's *named* fix — the
strings-aware mutable open — and it became unreachable when the lane started refusing every localized
target (measured: disabling the redirect outright left the registry at 131/131). **So #368 closes by the
refusal, not by the fix the issue named.** Its loss path is shut, and shut harder than the issue proposed,
because the lane now refuses before reading matters. Unreachable code with a claimed purpose is a future
stale claim, so it is gone; git history keeps it.

**#374 is fixed at both layers.** The refusal message stopped naming a remedy a localized referencer would
refuse — but `WriteTools.cs`'s tool *description*, which a calling agent reads *before* choosing
`repoint_externals`, still steered them there. Both layers are now true, and no write tool's description
was mentioning localization at all.

### Sentences — each fixed with a probe, never a rewording

- **The "cannot find its text" refusal no longer asserts an absence it did not check.** It had been false
  about the modder's disk twice in one sentence: "no Strings folder beside it" (the folder was there,
  holding a neighbour's tables), then, after that directed fix, "no `.STRINGS` files for this plugin beside
  it" (with `ZRef_ptbr.STRINGS` sitting in it — houseCARL had not *matched* it, which is a different
  claim). It now counts and names what is in the folder and claims only that nothing in it matched.
- **The shared "why not in place" justification is per shape.** It described an interruption leaving records
  reading other records' text — exact for a live set beside the plugin, and false where a write would
  replace no table at all. With the text in game-Data or inside an archive, a plugin write cannot reach it;
  what it would do is *shadow* it and leave it stale. Armed in both directions.
- **The archive refusal names both locations** when a loose set is beside the plugin too.
- **The compact note drops its count** — it said "including the 1 other language(s) it shipped (English,
  French)", a count of N−1 against a list of N, with the surviving language listed as lost.
- **The `ArchiveCache` invalidation comment was measurably false** and the cache is rekeyed: on NTFS an
  in-place `.bsa` rewrite leaves the parent directory's mtime byte-identical, so `bsa_repack` inside one
  process served a stale stem list. Now keyed on each archive's own name, length and write time. **Unarmed,
  and it says so with the ground**: observing stale-vs-fresh needs a `.bsa` that parses, and Mutagen exposes
  no in-process archive builder.

Also: the dead exception surface is gone (`Text`, `Why`, the public constructor — a second refusal sentence
still asserting the pre-refusal behaviour); the `ci-all` registry entry itemises what the suite now contains
rather than three passes that were deleted; the CHANGELOG no longer announces two fixes for regressions that
never shipped on `main`.

## Residue, filed rather than folded

**#435 — `merge_plugins` de-localizes a localized donor silently**, no note at all. It is the compact lane's
old behaviour and a silent change of the output's nature. It is *not* #373 — it emits no tables and mispairs
nothing — and it is pre-existing on `main` in a path this branch does not otherwise touch, so it went to the
issue lane under CLAUDE.md's outside-the-branch rule rather than into this diff.

## Verification

`ci-all` **131/131**, before the work and after. **Nine sabotage cells, all run from a committed state, each
restore `git checkout --` and grep-verified, tree confirmed clean between cells.** The sweep carried a
known-RED canary proven by hand first:

| Sabotage | Result |
|---|---|
| **CANARY** — the fail-open read restored | RED, 3 arms; the pre-flight prints `<PROCEEDED>` |
| the choke point decides off the re-read again | RED, 3 arms — the write reaches the locked file |
| the shadow shapes get the scramble sentence | RED, 5 arms |
| the "cannot find its text" sentence asserts an unchecked absence | RED, 2 arms |
| the archive refusal names one location while two exist | RED, 1 arm |
| the unmatched-table scan stops looking | RED, 2 arms |
| Q2-A returns — P′ flagged localized | RED, 3 arms |
| the compact note's count-vs-list contradiction returns | RED, 1 arm |
| the in-place refusal promises the cut lane again | RED, 2 arms |

Three cells first reported *setup-failed* (anchor missing, after a restore flipped a file to CRLF) and one
*did-not-compile*. Noted because the sweep reported those as broken cells rather than as passes — the
behaviour the canary rule exists to produce. All four were re-run to a real RED.

**The fail-open is armed in both directions**, on the reviewer's own repro shape — the destination held
`FileShare.None` for the duration of the write. Locked: classifies unreadable, pre-flight refuses, write
refuses, plugin and tables byte-identical, no orphaned staging directory. Unlocked and non-localized: no
refusal and the lane proceeds, so the refusal is not simply unconditional.

## Round accounting

- **Round 1** — 4 reviewers (correctness / filesystem-safety / guard-integrity / user-sentences), isolated
  worktrees, seeded with the decision record and settled-decisions list only. 3 distinct HIGH, 7 MEDIUM,
  8 LOW, 2 coverage gaps. Nothing refused as unreproducible. **Stopped under §4 before folding**; the
  advisor ruled the in-place arm cut.
- **Fold** — the cut, plus the sentence class, each fix probed rather than reworded. Four sabotage
  RED-checks from a committed state, each restore grep-verified.
- **Round 2** — 4 fresh reviewers, isolated worktrees, seeded with the amended settled list only. 1 HIGH
  (found independently by two of them), 8 MEDIUM, 12 LOW, plus 12 break-test mutations of which 10 went RED.
  Nothing refused as unreproducible — and that is called out as a *weak* signal rather than a clean one:
  most of round 2's findings were in code written by the session doing the triage, which is the blind spot
  the fresh-review-session rule exists to avoid, and the second reason it went to Aaron. **Stopped under
  §5 #11's recurring-class rule.**
- **Round 3 not run.** The §11 cap is spent and this wave is directed work under a §4 ruling — the #386
  precedent. Fresh eyes have had two passes; **Aaron's independent review is the gate.**

One claim in the round-2 escalation's own first draft ("the replacement guard is vacuous") was refuted by
that round's mutation table and withdrawn before this PR.

## Aaron's review — seven findings, folded at the CLASS

Advisor-triaged 7/7 verified at source, 0 refused. **Directed folds under Aaron's own review; no new
rounds** — the §11 cap was spent at two, and "no new rounds after directed folds" governs.

The seven are one defect wearing seven faces: **the fail-closed conversion collapsed a three-state fact
into a boolean at the RENDER layer.** `srcShape.Shape != LocalizedShape.NotLocalized` and its siblings are
CORRECT for the refusal DECISION — `Unreadable` must block, which is this branch's own fix — and WRONG for
any sentence, because they make `Unreadable` inherit `Localized`'s words: "is flagged LOCALIZED", "its text
lives in separate .STRINGS files", asserted about a file houseCARL explicitly could not open. The fail-open
sentence's mirror image, and harmless only until the four service pre-flights began failing CLOSED on it.

### The render-seam contract

> **Decisions may collapse; words may not.** No localization sentence renders from a collapsed boolean.
> Every render site switches exhaustively on `LocalizedShape`, so an arm exists per shape, `Unreadable`
> carries only its own honest words, and a `_` arm asserts NO localization state rather than inheriting
> another shape's.

1. **`LocalizedTargetUnsupportedException.ShapeBody` is the seam** — exhaustive switch; `Unreadable` gets
   its own body and its own close (`SettledUnreadable`, never `Settled`, which states that the file IS a
   localized plugin). `WhereTheTextIs` and `WhyNotInPlace` each carry an `Unreadable` arm too, so a direct
   caller cannot get localized reasoning either. Three `_` arms spelled "It is flagged LOCALIZED, so its
   text lives in separate .STRINGS files" — the same defect one level down — now assert nothing.
2. **`Shaped`'s REMEDY switches the same way** (`RemedyFor`). A lane's clause answers "this plugin is
   localized, so what instead"; appended to an unreadable destination it answers a question nobody
   established the answer to, and the remove lane's spelling turns a momentary file lock into a permanent
   dead end.
3. **`LocalizedStrings.ConfirmedLocalized(shape)`** — "was the flag actually READ and set" — is what a lane
   consults before its output ASSERTS, as against the wider fail-closed `!= NotLocalized` it consults
   before it REFUSES. An exhaustive switch, not an inverted operator, so a later shape must be answered.

### Fold table

| # | Finding | Commit | Probe result |
|---|---|---|---|
| 1 | `Unreadable` ShapeBody arm asserts the plugin is localized | `707c788` | GREEN; RED×2 — reroute `Unreadable` through the composed halves (3 arms), and drop the remedy switch (2 arms, reproducing the remove-lane dead end verbatim) |
| 2 | the capped list's length rendered as the folder's count | `4ecfb39` | GREEN; RED — collapse `Total` onto `Cap` (2 arms). New `ManyNeighbourTables` fixture: 12 files against a cap of 8 |
| 3 | an unlistable `Strings` folder indistinguishable from an empty one | `f5ecc1b` | GREEN; RED — map `Unlistable` back onto `Listed` (3 arms, reproducing "no .STRINGS files beside it" verbatim). ACL deny-listing fixture, verified to bite; a cell that cannot be built FAILS rather than skipping |
| 4 | compact's report note and in-place arm assert LOCALIZED about an unread source | `2f2b9a1` | GREEN; RED — route `Unreadable` into the localized arm (2 arms); RED — collapse `ConfirmedLocalized` (2 arms) |
| 5+6 | `LocalizedAmong`'s hits rendered as one class | `9bfeef4` | GREEN; RED — collapse the split (5 arms) |
| 7 | the sweep's pre-cut ALLOW column, and `Unreadable` inside the localized total | `527fa57` | GREEN; re-run on the real order — 3802 scanned, 63 flagged localized, 0 unreadable, ALLOW gone |
| — | CHANGELOG claims the folds changed | `602aa53` | ci-all green |

`ci-all` **131/131** before (`9bab724`) and after. Every sabotage from a committed state, every restore
grep-verified under the dirty-restore gate, tree confirmed clean between cells.

**One new shape** — `LocalizedShape.StringsFolderUnreadable`: flagged localized, `Strings` folder present
and unlistable. `ReadStringsFolder` is now the single enumeration behind both the matched languages and the
unmatched names, because two passes over one folder could disagree with each other. The game-Data side
stays two-answer deliberately, and the comment says why: nothing rendered off it asserts an absence.

### Two arms could not be built, and the guards SAY so rather than looking armed

1. **The compact report note's gate is not driven end to end.** `PluginIsLocalized` and
   `TryReadOriginatingKeys` make the identical `CreateFromBinaryOverlay` call on the same path a few lines
   apart, so a file the first cannot open the second cannot either — the held source fails the whole
   compaction and produces no report at all. **A first draft of that arm asserted "held, no localization
   note" and passed for exactly that reason**; its own RED-check came back GREEN and caught it, and it was
   replaced by one stating what it drove. What remains between the two reads is a sub-millisecond race
   inside one `_writeGate` hold. The gate is armed at the predicate instead.
2. **`LocalizedAmong`'s mixed hit list cannot be produced through the service** — measured, not assumed:
   hold a referencer `FileShare.None` and the identify pass cannot read it either, so it drops out of
   `ExternalPlugins` and the pre-flight is never asked about it. The `Unreadable` class arises only by a
   referencer becoming unreadable BETWEEN the identify pass and the pre-flight, which is why
   `LocalizedAmong` fails closed on it. The render is armed on the REAL renderers, internal to
   `housecarl-mcp` via `InternalsVisibleTo` (the `NativePairingProbe` pattern), with the mixed list the
   service cannot be made to produce; the localized-only path is still driven end to end.

Same shape as this branch's existing note about the four pre-flights being unreachable in practice —
recorded rather than hidden.

**Reachability note on (2), measured on request and NOT fixed here — it is outside this branch (the #324
shape).** How the identify pass drops a referencer it cannot open depends on WHEN the file became
unreadable, and only one of the two answers is loud:

- **Locked before the resolver's index was built — LOUD.** The resolver records it in `ExcludedPlugins`
  with the full reason (`could not be opened — IOException: … used by another process`),
  `IdentifyExternalReferencers` skips it at its `ExcludedPlugins` guard, and it is NOT counted in
  `PluginsScanned`. `housecarl_load_order_status` reports the exclusion, though the compaction's own answer
  does not repeat it.
- **Locked after the index was built — SILENT.** `LoadOrderResolver.RecordsIn` swallows the failed
  per-plugin re-open (`try { ov = OpenOverlay(…); } catch { continue; }`), so the identify pass's own
  per-plugin catch never fires: `unscannable=0`, `UnscannableSamples` empty, and `PluginsScanned` still
  counts the plugin as scanned. This is the live-server shape — holding a file does not change its mtime,
  so the index is not refreshed — and it means the external-referencer safety scan can under-count: a
  compaction could proceed reporting a clean scan while a real-but-unreadable referencer exists, whose
  references then break.

Measured with a throwaway probe driving `IdentifyExternalReferencers` directly, held for the whole call, so
the answer is not timing-dependent (the probe was removed; nothing of it is in this branch). Pre-existing on
`main` in a path this branch does not otherwise touch; routed to the advisor as a filing candidate rather
than folded here.


src/housecarl-core/WriteEngine.cs
```diff
@@ -3936,25 +3960,205 @@ public NullArmSerializeException(Exception inner)
...
+            + "Rewriting the plugin renumbers those indices, and houseCARL cannot tell whether a set written beside it "
+            + "would replace what the game reads or shadow it, nor what would be left stale either way." + Settled,
+
+        // The plugin itself could not be read. Nothing about its arrangement was looked at, so nothing about its
+        // arrangement is claimed.
+        LocalizedShape.Unreadable or LocalizedShape.NotLocalized =>
+            "A localized plugin's text is not in the plugin, and houseCARL could not establish where this one's is, so "
+            + "it cannot tell what rewriting the plugin would do to that text." + Settled,
+
+        // Enumerated above; a shape added later fails LOUD and generic rather than inheriting another's reason (Q3).
```
Avick3110: **The `Unreadable` arm asserts the plugin is localized, and the service pre-flights now act on it.**

`WhereTheTextIs`'s `Unreadable` arm is careful — "houseCARL could not read the file at that path to see where its text lives." `ShapeBody` then concatenates *this* arm, which opens "A localized plugin's text is not in the plugin, and houseCARL could not establish where **this one's** is" and closes on `Settled` — "It does not edit a localized plugin in place." Those two halves of one sentence disagree, and the second one states a fact about the modder's plugin that was never checked.

That was harmless while `PluginIsLocalized` answered `false` on a fault. It is not harmless now that the four service pre-flights fail closed on `Unreadable`.

**Scenario.** A modder runs `housecarl_set_field` with `in_place=true` on a plain, **non-localized** plugin while MO2 refreshes or an AV scanner holds it for the instant of the header read. `RefusalFor` returns non-null, and the lane fails with a message telling them their plugin is localized and to use the new-plugin lane. Before this branch the lane proceeded and surfaced the real open failure — which is the thing they can act on. On `housecarl_remove_record` the appended clause is `RemoveNoEquivalent` ("there is no way to remove this record without rewriting the plugin that defines it"), so a momentary file lock now reads as a permanent dead end.

The fail-closed decision is right; the sentence attached to it is the fail-open sentence's mirror image — it names a cause it did not establish instead of the one it did. Worth an `Unreadable`-specific arm that says only that the file could not be opened and that houseCARL will not write to a destination it cannot classify.

src/housecarl-core/WriteEngine.cs
```diff
@@ -3936,25 +3960,205 @@ public NullArmSerializeException(Exception inner)
...
+    /// classification.</summary>
+    /// <summary>What the <c>Strings\</c> folder beside the plugin actually holds, for a plugin nothing in it matched.
+    /// Names the files when there are some, so the modder is not told "nothing is there" while looking at a folder
+    /// with files in it — and asserts an absence only where the absence was checked.</summary>
+    static string NothingMatched(LocalizedAssessment a)
+    {
+        if (a.UnmatchedTables.Count == 0)
+            return "no .STRINGS files beside it";
+        return "the Strings folder beside it holds " + a.UnmatchedTables.Count + " .STRINGS file(s) — "
+             + string.Join(", ", a.UnmatchedTables)
```
Avick3110: **The count is the cap, not the folder.**

`a.UnmatchedTables` is produced by `UnmatchedTablesIn`, which ends `.Take(8).ToList()`. So `a.UnmatchedTables.Count` is `min(actual, 8)`, and this sentence renders it as what the folder holds — then lists the eight names with no ellipsis.

**Scenario.** The `ksws07_quest` / `ksws07_quest_shrubs` pair this branch is built around, but with the neighbour shipping ten languages (10 × 3 kinds = 30 table files) in the shared Strings folder. The `Nowhere` refusal for the first plugin reads:

> the Strings folder beside it holds 8 .STRINGS file(s) — a, b, c, d, e, f, g, h — but houseCARL matched none of them to this plugin in a language it recognises

The count is false about their disk and the list stops silently. This is the same class as the two falsehoods the arm was rewritten to remove ("no Strings folder beside it", then "no .STRINGS files for this plugin beside it") — an assertion about the folder that the code did not check.

The guard cannot catch it: `LocalizedWriteGuardProbe` asserts `files.All(n => msg.Contains(n))` against the *same* capped list, and both fixtures (`NeighbourTablesOnly`, `UnknownLanguageToken`) produce six files, under the cap.

Carrying the true count alongside the capped names (and saying "… and N more") is the fix; a fixture with more than eight unmatched files is what arms it.

src/housecarl-core/LocalizedStrings.cs
```diff
@@ -0,0 +1,422 @@
...
+    {
+        if (!Directory.Exists(stringsDir)) return Array.Empty<string>();
+        try
+        {
+            return Directory.EnumerateFiles(stringsDir)
+                            .Where(p => Parse(Path.GetFileName(p), stem) is not null)
+                            .ToList();
+        }
+        catch (IOException) { return Array.Empty<string>(); }
+        catch (UnauthorizedAccessException) { return Array.Empty<string>(); }
```
Avick3110: **An unlistable Strings folder is indistinguishable from an empty one, and both the classification and an absence claim ride on it.**

`TableFilesIn` swallows `IOException` / `UnauthorizedAccessException` and returns `Array.Empty<string>()`. `UnmatchedTablesIn` (line 268) does the same. Neither caller can tell "enumerated it and found nothing" from "could not enumerate it" — and this file's whole thesis is that those are different answers.

**Scenario.** A mod folder whose `Strings` path exceeds MAX_PATH (`Directory.EnumerateFiles` throws `PathTooLongException`, an `IOException`) — routine in a deep MO2 mods tree — or whose ACL denies listing.

1. `LanguagesIn` → `TableFilesIn` returns empty, so `own.Count == 0` and `Assess` falls straight through to `LocalizedShape.Nowhere`, even though a complete loose set is sitting right there.
2. `UnmatchedTablesIn` returns empty for the same reason, so `NothingMatched` picks its `Count == 0` branch and emits **"no .STRINGS files beside it"**.

The modder gets "houseCARL cannot find its text: no .STRINGS files beside it, and no archive beside it or in your game folder carrying its tables" — an unchecked absence, stated flatly, over a folder that is full. That is the exact falsehood this arm was rewritten twice to stop making; it just arrives through the exception path instead of the matching path.

`Assess` has three answers for the *plugin* and two for the *folder*. The folder needs the third one too — a "the folder is there and houseCARL could not read it" state that `NothingMatched` and the shape decision can both see.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6429,12 +6465,26 @@ public WritePatchBuilder.CompactOutcome CompactPlugin(
...
-                    $"'{name}' is flagged LOCALIZED — its text lives in separate .STRINGS files. The compacted plugin " +
-                    "houseCARL wrote is NOT localized: it carries whatever this read of the source produced, written " +
-                    "into the plugin itself, and the source's .STRINGS files do not describe it. Read the output before " +
-                    "you enable it in place of the original.");
+                    $"'{name}' is flagged LOCALIZED — its text lives in separate .STRINGS files rather than in the "
+                    + "plugin"
+                    + (srcShape.Languages.Count > 0 ? " (" + string.Join(", ", srcShape.Languages) + ")" : "")
+                    + ". The compacted plugin houseCARL wrote is NOT localized: it carries whatever this read of the "
+                    + "source produced, written into the plugin itself, with no .STRINGS files of its own — so the "
+                    + "source's .STRINGS files do not describe it, and any language it shipped that this read did not "
```
Avick3110, line 6489: **`srcLocalized` is true for `Unreadable`, so this note asserts LOCALIZED about a source houseCARL never read.**

Line 6225: `bool srcLocalized = srcShape.Shape != LocalizedShape.NotLocalized;` — that is deliberately fail-closed for the in-place refusal above, and right there. But the same boolean gates this **report note**, where the correct answer is not "assume the worse one", it is "say nothing".

**Scenario.** `housecarl_compact_plugin` on a plain non-localized plugin that is locked for the instant `Assess` reads it at 6224 (AV scan, MO2 refresh) and readable again by the time `TryReadOriginatingKeys` runs seventeen lines later. The compaction succeeds, and the report carries:

> 'X.esp' is flagged LOCALIZED — its text lives in separate .STRINGS files rather than in the plugin. The compacted plugin houseCARL wrote is NOT localized …

X.esp is not localized and has no .STRINGS files. The modder is sent looking for tables that do not exist and told to distrust an output that is exactly what they asked for.

Same boolean, same read, at line 6226: an unreadable source taking the in-place lane is told houseCARL "will not replace a translated plugin's .STRINGS files on your own copy" and pointed at the new-file lane — which will fail for the underlying reason nobody has been told about. That is the #374 shape (a remedy that dead-ends) reappearing on the `Unreadable` arm.

Gating the note on `srcShape.Shape` being one of the *found* shapes — rather than on "not NotLocalized" — keeps the refusal fail-closed while stopping the report from claiming something it did not observe.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6267,11 +6277,32 @@ public WritePatchBuilder.CompactOutcome CompactPlugin(
...
+                    // cannot rewrite, and the caller learns that here rather than by following the instruction and
+                    // meeting a second refusal. The check runs only on the referencers already named, and only on this
+                    // branch: the repoint branch below has its own, which refuses before anything is written.
+                    var blocked = RemapEngine.LocalizedAmong(resolver, id.ExternalPlugins);
+                    var repointClause = blocked.Count == 0
+                        ? "Re-run with repoint_externals=true AND in_place=true (+ acknowledge=true) to ALSO rewrite those plugins in place to follow "
+                          + "the renumber, or handle them yourself first."
+                        : $"Re-running with repoint_externals=true will NOT work here: {string.Join(", ", blocked.Take(25).Select(b => b.Plugin))}"
+                          + $"{(blocked.Count > 25 ? $", … (+{blocked.Count - 25} more)" : "")} "
+                          + (blocked.Count == 1 ? "is localized" : "are localized")
```
Avick3110: **`LocalizedAmong` now returns unreadable referencers too, and this clause calls all of them localized.**

The `RemapEngine.LocalizedAmong` change is the right one — an unclassifiable referencer must block. But its hits are no longer homogeneous, and this renders every one of them as "is localized" / "are localized".

**Scenario.** Compacting a plugin whose external referencer `Ref.esp` is briefly held by xEdit. `Assess` returns `Unreadable`, `RefusalReasonFor` returns non-null, and the caller reads:

> Re-running with repoint_externals=true will NOT work here: Ref.esp **is localized**, and houseCARL does not rewrite a localized plugin in place, so the repoint would refuse before touching anything. Where Ref.esp's text is: **houseCARL could not read the file at that path to see where its text lives.** …

The two halves of that sentence contradict each other, and the first half is a claim about Ref.esp that houseCARL explicitly did not establish. The attributed-reason clause immediately below it is honest — this line is what breaks it.

Splitting the hits ("N localized, M houseCARL could not read") keeps the block and stops the message asserting the shape.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6306,10 +6337,15 @@ public WritePatchBuilder.CompactOutcome CompactPlugin(
                     return WritePatchBuilder.CompactOutcome.Fail(
                         $"refused — compacting '{name}' means rewriting the plugins that reference it, and " +
                         $"{(localized.Count == 1 ? "one of them is" : $"{localized.Count} of them are")} flagged LOCALIZED: " +
```
Avick3110: **Same as the `!repointExternals` branch: `"N of them are flagged LOCALIZED"` is false for the `Unreadable` hits.**

`LocalizedAmong` deliberately fails closed on a referencer it could not read (that is this branch's own change, and it is correct). This sentence then reports the whole hit list as flagged LOCALIZED.

**Scenario.** `repoint_externals=true` where `A.esp` really is localized and `B.esp` cannot be opened (corrupt header, or held by the running game):

> refused — compacting 'X.esp' means rewriting the plugins that reference it, and **2 of them are flagged LOCALIZED: A.esp, B.esp**. Where A.esp's text is: …

B.esp is not localized as far as anyone knows; it is unreadable, which is a different problem with a different fix. And because only `localized[0]` gets an attributed reason, B.esp's actual condition never appears anywhere in the message — the modder goes looking for .STRINGS files instead of for the file they cannot open.

The count and the label both want splitting by shape here.

src/housecarl-generator/LocalizedShapeSweep.cs
```diff
@@ -0,0 +1,90 @@
...
+        int localized = counts.Values.Sum(l => l.Count);
+        Console.WriteLine($"instance : {instance}");
+        Console.WriteLine($"dataDir  : {dataDir ?? "<none>"}");
+        Console.WriteLine($"plugins  : {scanned} scanned, {failed} unreadable, {localized} flagged localized");
+        Console.WriteLine();
+        foreach (var shape in Enum.GetValues<LocalizedShape>())
+        {
+            if (shape == LocalizedShape.NotLocalized) continue;
+            var l = counts.TryGetValue(shape, out var x) ? x : new List<string>();
+            var verdict = shape == LocalizedShape.LooseComplete ? "ALLOW " : "REFUSE";
```
Avick3110: **The verdict column is from before the cut.** `LooseComplete` prints `ALLOW`; every other shape prints `REFUSE`. This branch cut the in-place write for **every** shape, `LooseComplete` included — so the one row labelled ALLOW is the one the PR body specifically says is now refused.

Anyone re-running `localized-shape-sweep` after this lands reads "ALLOW LooseComplete 1" and concludes houseCARL will rewrite `ksws07_quest.esm` in place. It will not. This is the tool whose table the PR body quotes, so the stale label outlives the branch.

Separately, line 56: `localized` sums every non-`NotLocalized` shape, which now includes `LocalizedShape.Unreadable` — so `"N flagged localized"` over-counts by however many plugins the sweep could not open. Those are exactly the plugins worth breaking out, since the sweep's whole job is pricing the population.

Issue #440, opened 2026-08-27
[GAP] typed refusal rendered per-transport by Guard.Tool — refusal coverage by construction


Avick3110:
**Direction filed out of the row-6 PR A round-1 escalation (framing (b), Aaron-ruled 2026-08-27).** Option (i) — fix the escaped sites and machine-check the population — landed on `claude/read-response-layer`. This is option (ii), the structurally correct answer, chartered separately rather than smuggled in as a review fold.

## The problem option (i) mitigates but does not remove

The read surface's refusal grammar is 80-odd explicit `Wire.Refuse(json, …)` call sites. Settled decision #5 chose that over a `Guard.Tool`-boundary post-process, because the alternative on the table sniffed a leading `"error: "` off the returned string — a string-prefix heuristic deciding whether a response is a refusal, which is a Q3 hazard and #308's shape.

The reasoning against *that* boundary still stands. What round 1 measured is that the option chosen instead has **no completeness property of its own**: a body returns `string`, so nothing prevents the next refusal from being returned unshaped. The population was found by a regex over two hand-named files and was short by fourteen — twelve refusals produced inside a helper and handed back verbatim, a ternary whose two arms were both bare (the sweep needed `return "error:` on one line; that line read `return comparisonForm`), and `housecarl_check`, a json-capable read tool that was never in the file set and whose own comment stated the rule its bare return was breaking.

A reviewer then proved by mutation that reverting **all fourteen** `ReadTools` call sites to prose left the entire 131-probe suite green.

That is CLAUDE.md §3's hand-wired-coverage failure mode restated on the response layer: *"comprehensive write access was always one more hand-port from done."*

## What (i) gives, and what it does not

`refusal-completeness-guard` (on the same branch) derives the population from the artifact — every `Guard.Tool` body consulting the format machinery — and enumerates bare whole-call refusal returns with a parser rather than a pattern, carrying the pre-fix ternary as a permanent known-RED fixture. A new json-capable tool enrols itself.

So coverage is now **checked** by construction. It is still not **true** by construction: a body can still return an unshaped refusal, and the guard catches it after the fact rather than the type system refusing to express it.

## The change

[...]

Issue #439, opened 2026-08-27
[Bug] truncation notices tell a housecarl_records caller to narrow with fields= — it has project.fields=


Avick3110:
A truncation notice on `housecarl_records` tells the caller to narrow with `fields=` and
lower `depth=`. That tool has neither parameter — it spells them `project.fields=` and
`project.depth=`. The caller is told to pull a lever that is not there (the #343 class:
a remedy predicts what a LATER call produces).

Both transports are affected, and they compose their notices separately.

Found by measurement while re-chartering RUN_ORDER row 6 (the refusal + remedy grammar).
Reproduced on branch `claude/read-response-layer` at `2c34a97`..`c1ed89e`.

## Mechanism

The json body renderers and the text field renderer are SHARED by both tool generations,
and the remedy sentence is composed inside the shared renderer, so it can only speak one
vocabulary:

- text: `ReadTools.cs:1850` (`Wire.AppendRecord`) — `"; narrow with fields=, lower depth=, or raise max_chars]"`.
  `RecordsTools` drives it through `Wire.RenderBatch` (6 call sites) and `Wire.RenderCrossQuery` (4).
- json: `JsonWire.cs:316` (`WriteFieldsArray`, reached via `WriteReadRecord`) —
  `"[truncated at max_chars: {i} of {n} fields shown; narrow with fields=, lower depth=, or raise max_chars]"`.
  `RecordsTools` reaches it through `JsonWire.RenderBatch` and `JsonWire.RenderCrossQuery`.

Two sibling notices in the same shared renderers name other 1.x-only levers to a
`housecarl_records` caller:

- `ReadTools.cs:989` (`Wire.RenderBatch`) — `"request fewer formids, pass fields= to slim each, or raise max_chars"`
- `ReadTools.cs:1065` (`Wire.RenderCrossQuery`) — `"lower limit=, drop fields=/conflict_tree, or raise max_chars"`
  (`conflict_tree` is also not a `housecarl_records` parameter.)

## Reproduction

A guard arm that fails on current `main` is attached below. Driving a max_chars-starved
`fields` read through `housecarl_records`:

```
housecarl_records(formids=[…], format="json", max_chars=220,
                  project={form:"fields", fields:["BasicStats.Damage","EditorID","Name"]})
```

observed notice:

```
[truncated at max_chars: 0 of 3 fields shown; narrow with fields=, lower depth=, or raise max_chars]
```

[...]

Issue #438, opened 2026-08-27
[NOTE] read surface: the non-refusal duplicated-sentence tail (13 facts, row-6 deferral)


Avick3110:
Deferred out of RUN_ORDER row 6 at its stage-1 count gate (Aaron-go 2026-08-27). Row 6
re-chartered as the **refusal + remedy grammar** migration (carrying #403, widened under §4(b));
this issue carries the rest of the measured duplication so the deferral is tracked rather than
implied. Routed **row 8 FIRM**, milestoned.

Measured at `main` @ `2fd7b6e` on branch `claude/read-response-layer`. Counter, lifetime split
and triage are reproducible from that branch's `dev-scratch/` scripts.

## The measurement

READ surface = `ReadTools.cs` + `RecordsTools.cs` + `JsonWire.cs`'s read region (split at its
first write-outcome renderer, `RenderPatchOutcome` :2302 — the write region is already on
`WriteSentences` from #337). Tool-schema `[Description(...)]` strings excluded throughout, per
#337's scoping of the tool schema as a different surface.

| File | raw user-facing literals | sentence-bearing |
|---|---|---|
| `ReadTools.cs` | 145 | 102 |
| `RecordsTools.cs` | 191 | 157 |
| `JsonWire.cs` (read region) | 13 | 13 |
| **total** | **349** | **272** |

**Duplicated sentence-facts: 36** — one fact stated in ≥2 places, the twin-drift unit.
Mechanically 28 exact-normalized groups + 21 paraphrase clusters = 49, hand-triaged to 36
(−10 where one sentence spans several literals and landed as several groups; −3 fuzzy false
positives that are genuinely different facts).

Calibration: run against the already-shared sources (`ReadSentences.cs`, `WriteSentences.cs`)
the same counter finds **0 exact duplicates**, which is what a working response layer should
look like. The 7 paraphrase clusters it finds there are the fuzzy half's false-positive rate,
and the reason the 49 is triaged rather than reported raw.

Split by family:

| Family | distinct facts | disposition |
|---|---|---|
| refusal grammar (`error: …` sentences) | 13 | **row 6** |
| remedy / truncation grammar | 10 | **row 6** |
| **the tail — descriptive facts about the read itself** | **13** | **this issue** |

## What this issue covers — the 13

| # | Fact | Sites |
|---|---|---|
[...]

PR #441, opened 2026-08-27
The read surface's refusal grammar: one shape, and a guard that keeps it

### LINKED ISSUE #403: [Bug] json refusal grammar: RenderError lacks the ok discriminant its own summary claims

A json caller gets `{error, epoch?}` from `JsonWire.RenderError` (`JsonWire.cs:186-197` — every pre-engine / parameter refusal) but `ok:false` plus lane fields from outcome refusals (`w.WriteBoolean("ok", ...)` at `:2309, :2410, :2520, :2603, :2665`), so a consumer branching on `ok` finds it **absent** on the first shape. `RenderError`'s own doc comment (`:182-185`) still claims it matches "the outcome-borne refusal shape so json consumers see ONE refusal grammar" — the discriminant is the gap. The other bare-refusal paths (`:962`, `:1077`) have the same shape.

History: found in PR #310's round-5 review, explicitly scoped out there (~39 call sites across `ReadTools`/`RecordsTools` — surface-wide). The fix is one edit in `RenderError` + re-running the json arms of `records-guard`/`artifact-guard`/`apply-guard`; decide at the same time whether `RenderError` also carries `lane`/`dry_run` as nulls or stays minimal.

**Routing:** step 6 (READ-surface response-layer migration — same surface-wide read-side shape work), or its own small PR if step 6 dissolves under its count-first escape hatch. Milestoned.

Avick3110:
Fixes #403

A read tool asked for `format="json"` answered some refusals with a plain sentence instead of a json document, so a caller parsing the response handled two shapes depending on which rule it broke. This routes those refusals through one render path, gives the json lane the `ok:false` discriminant #403 asked for, and — after the review found the first attempt's coverage was hand-wired — adds a guard that derives the covered surface instead of trusting a sweep.

## What changed

- **`Wire.Refuse` is the one render path.** A tool body states its refusal once, in the text spelling; the path decides the shape. Refusal **text is byte-identical by construction** — `Refuse` returns the message unchanged when `json` is false.
- **`JsonWire.WriteRefusal` is the one discriminant writer.** 10 document-root refusals; `RenderError`'s summary claimed a shape the document did not carry (#403).
- **Two refusal-ordering fixes.** `batch_record_detail` and `resolve` refused an empty `formids=` *above* the line deciding the transport, so that one refusal could not honour json however it was rendered.
- **The served census stops answering to the discriminant** — `counts_only=true` wrote `"ok": <int>` (a resolved-row count) on a served document. Renamed `resolved`.
- **Eight post-capture refusals carry the build they consulted** instead of asserting `epoch: null`.
- **`refusal-completeness-guard`** — new, and the reason this branch is worth more than its diff.

Per-row `error` fields never gain `ok` (12 sites, untouched): a row that failed inside a call that succeeded is not a refusal.

## Review rounds

**One round, three fresh reviewers** (`isolation: "worktree"`, seeded with the branch's settled-decisions list and nothing else), then a **§11 class stop** and a directed fold wave. Round 2 was deliberately not run — reasoning below.

| Reviewer | Lens | Findings |
|---|---|---|
| 1 | runtime behaviour | 1 high, 2 medium, 1 low |
| 2 | contract completeness | 2 high, 1 medium, 2 low |
| 3 | does the pinning bite | 2 high, 2 medium |

**All three found the same class, by three different methods, without seeing each other's work:** a whole-call refusal escaping the one render path because the covered population was enumerated by a regex over two hand-named files.

### What the round found

- **The branch was not green.** `ci-all` was 130/131. `BulkPrimitivesWave2Probe.cs:285` pinned `conflict_tree=true, format=json` as `StartsWith("error:")` — the exact refusal this branch converts to a document. "Branch-green" had been measured on three guards, not the suite.
- **14 refusals outside every settled exception.** Twelve helper-returned (`ExpandListInput` / `ValidateToFile` hand back an `"error: …"` string that the body returned verbatim — `RecordsTools.cs:381` had a bare one three lines above three wrapped ones); one ternary at `RecordsTools.cs:223` whose two arms were both bare, between wrapped statements on either side; and `housecarl_check`, a json-capable read tool never in the traced file set, whose own comment seven lines below the bare return already stated the rule it was breaking.
- **Nothing would have caught a regression.** Reviewer 3 proved by mutation that reverting **all 14** `ReadTools` `Wire.Refuse` sites to prose left the whole 131-probe suite green. The grammar was pinned at 2 of 68 sites in `RecordsTools` and 0 of 14 in `ReadTools`.
- **The discriminant collided with an existing key**, on a document no refusal probe rendered.

### Triage: what was refused or narrowed

Refusal rate was low, so here is where triage bit rather than obeyed:

- **The `epoch` finding was narrowed.** Reviewer 1 called `Wire.Refuse`'s unused `epoch` parameter a defect at all 81 sites; reviewer 3 dropped the same observation as chartered shape (settled #2). Both were partly right: the `epoch: null` *shape* is chartered, but passing null where a stamp is **in hand** is a site-level miss under settled #4. Fixed at the 8 sites holding one; `RecordsTools.cs:979` deliberately left alone, because `ProbeSourceArm` returns null there and settled #4's third bucket is correct.
- **`Guard.Named` prose refused as a fix.** Reviewer 2 raised it and correctly declined to file it, citing settled #5. Agreed — claim-scope, not a defect.
- **The dead conjunct: reviewer 3 over reviewer 1.** `StartsWith("notaform") == false` can never be false, but the `Contains` half does the stated work. Dead weight, not a defect — deleted as fix-or-drop.
- **Settled decision #10 was never tested.** No reviewer proposed remedy unification or remedy-text changes, so nothing was refused on that front. Remedy work stays #439's own PR.

### Why one round, not three

§11 forbids folding a third instance of a recurring class and sends it to Aaron instead. The class recurred three times *inside* round 1, via three independent methods — a stronger signal than two sequential rounds, not a weaker one. Round 2 would either find instance #4 (unfoldable by rule) or find nothing (which would not remove the 14 sites). Neither changes the decision, and #323's precedent — ten rounds, ~4M tokens, four of five highs self-inflicted by the folds — argues against grinding to satisfy the letter of the rule.

Escalated under §4 framing **(b)**: the goal stands, the assumption that per-site explicit calls achieve it was false as executed. **Aaron ruled**: framing (b) accepted, option (i) now with (ii) filed as the direction (#440), and the `ok` collision ruled a rename. The advisor verified the collision and the `CheckTools` site at source and endorsed the round-2 skip on the record. What follows is that directed fold wave — no new rounds after it, per §11.

## The guard

Two properties, both load-bearing:

- **The population derives itself.** Every `Guard.Tool` body that consults the format machinery — 15 bodies across 8 files, found by walking the tree. A new json-capable tool enrols itself the day it is written; a hand-named file set is what let `housecarl_check` sit outside the net. Deriving also reaches past the read lane into the write tools, which is why the `format=` parse exception is written surface-wide rather than per file.
- **The enumeration parses.** Roslyn walks return statements; no regex anywhere. What defeated the sweep was a ternary spanning three lines — exactly what a line-oriented pattern cannot see. That construct is preserved verbatim as a **permanent known-RED fixture** the enumerator must flag, with the wrapped form of the same shape as its counter-fixture. If the fixture ever passes, the guard is broken, not the tree.

Three shapes are excluded structurally rather than by hand, because an exception that restates a structural fact is a guard that has stopped deriving: the MO2 config prompt (trained guidance, not a refusal), the write tools' own local `Refuse` helper (same verb, different receiver), and a text-lane twin whose json arm returned above it. What stays bare cites the settled decision that permits it, and a stale entry is its own failure.

A fourth arm makes the discriminant unambiguous by construction: every `ok` property write in the json renderer must be the literal `false` (the discriminant, written in one place) or an outcome's `.Success` (the write surface's contract). Anything else is the census collision returning.

## Aaron's gate review — four directed folds

His independent review found four, all folded (no new rounds — §11 forbids reopening the loop after directed folds). Each fold is one commit with its own RED-check from a committed state, and each mutation he supplied is now a permanent cell or fixture.

| Sev | Finding | What landed |
|---|---|---|
| high | The binding net missed the **tuple deconstruction**. `var (…, xerr) = Artifacts.ExpandListInput(…)` is a `DeclarationExpressionSyntax` carrying a `ParenthesizedVariableDesignationSyntax` — neither an `out var` argument nor an `is { }` pattern. 8 live sites; reverting any of them left the guard green. | The designation joins the net; the shape joins the permanent fixture set with its wrapped counter-fixture. The comment above the walk called ExpandListInput's tuple an out argument — it now names all three bindings for what they are. |
| high | The derived population **stopped at `Guard.Tool` lambdas**, so a helper *handed* the transport sat outside it — it produces and returns its refusal internally, so no tool-body return ever carries the sentence. `RecordsTools.RenderListAggregate` reverted to prose, guard green. | A scope is policed when the transport is **decided in it or handed to it**: a method or local function declaring `bool json` or `QueryFormat` joins the population. Still derived — no file named. Three live scopes join; **nothing new is flagged, so no allowlist entry was needed** — the escalation condition the finding set out did not fire. A fixture proves the derivation both ways, including the `QueryFormat` half, which has no live parameter today and would otherwise be unproven. |
| low | Allowlist matching was an **unanchored substring across two kinds** of `Hit.Sentence` — an identifier for a binding hit, a whole message for a literal one. A future refusal sentence containing `ferr` ("…the record was transferred…") would have been ruled correct silently. | Hits and entries are typed. A binding entry must **equal** the identifier; a literal entry matches message text; no cross-kind match. The example is the fixture: one policed helper yielding both hits, the binding ruled, the sentence left unexplained. |
| low | The CHANGELOG sentence read as a **closed list**, and two prose paths were missing from it. | Both verified at source first — `Guard.Named` (`Guard.cs:40`) on any unhandled exception, and `svc.ConfigPromptOrNull()`'s MO2 setup prompt, returned first by every tool body. The claim now names them as *not refusals of your call*. The `Guard.Named` triage stays refused as claim-scope; this narrows the claim to match it, it does not reopen it. |

**What the two highs say about the guard, stated plainly.** Its shape net is an **enumerated list of syntactic shapes** with a known-RED fixture per shape — a literal at any depth, an `out var` binding, an `is { }` binding, and now a tuple designation — over an enumerated list of policed scopes, tool bodies plus transport-carrying signatures. That is much stronger than the regex it replaced and still weaker than a property: every blind spot found so far has been a shape or a scope missing from those lists, and each was found by someone mutating the tree rather than by the guard noticing. **#440 is the by-construction close** — a typed refusal rendered per-transport at the `Guard.Tool` boundary, where the enumeration stops being load-bearing. A fourth blind-spot instance after this merges promotes it.

## Verification

- **`ci-all`: 132/132 ALL PASS** (the suite grew by this guard). Branch-green means the suite now, not three guards.
- **Byte-identity, mechanically** (`dev-scratch/literal_identity.py`, rewritten and re-run over `2fd7b6e..HEAD`): every C# string literal extracted from both sides by a scanner rather than a line regex — comments skipped, regular / verbatim / interpolated / raw forms all recognised — and refusal-bearing ones compared as multisets across the whole of `src/housecarl-mcp`: **267 base / 268 head, zero changed sentences**. The single addition is the new `RefusalPrefix` constant, `"error: "`. Canary-tested: perturbing one head-side sentence by one character reports it. **Its scope is refusal text only**, so it is blind to served-schema changes by design and reports those separately — `resolved: 1→2` (the rename) and `error: 25→15` (the `WriteString("error", …)` sites collapsing into the one writer). The counts differ from the figures quoted before Aaron's review because that extractor scoped itself to the changed files; the deltas agree, and this one states its scope.
- **Mutation check, from a committed state, canary intact** (`dev-scratch/completeness_sweep.sh`), re-run in full after Aaron's folds:

| Cell | Verdict |
|---|---|
| canary — the `:223` ternary back to bare arms | FAIL (expected), 1 unexplained |
| tuple-deconstruction — `ReadTools.cs:101` → `return xerr;` | FAIL, 1 |
| transport-helper — `RenderListAggregate` → the bare form | FAIL, 1 |
| one `ReadTools` site → prose | FAIL, 1 |
| helper-returned `verr` revert | FAIL, 1 |
| `CheckTools` `famErr` revert | FAIL, 1 |
| ALL `ReadTools` sites → prose (20 bypassed) | FAIL, 14 |
| ALL `RecordsTools` sites → prose (74 bypassed) | FAIL, 66 |
| allowlist kind check removed | FAIL |
| control — an unrelated comment edit | PASS |

**The earlier table overclaimed the `ALL ReadTools` cell**, and Aaron's finding 1 is what exposed it: that cell reverted only the literal-bearing sites, so the binding arm of the net was never exercised by it and the tuple hole survived a sweep that read as exhaustive. The cell now rewrites **every** `Wire.Refuse(` in the file — all 20 — through a bypass helper whose name is not an approved renderer. 14 of the 20 come back unexplained; the other 6 pass an *identifier* (`xerr` x3, `verr` x3), so under the bypass the return is still an invocation and the guard has nothing to flag. Those 6 are the sites the dedicated `tuple-deconstruction` and `helper-returned-verr` cells cover, by reverting the return to the genuinely bare form. Stated here rather than left implicit: a mass cell is not a substitute for the per-shape cells.

The `transport-helper` cell was itself broken on first run — anchored on the refusal sentence, it hit the *identical* sentence at `RecordsTools.cs:239` inside a tool body and reddened for the wrong reason. Re-anchored on the enclosing `gb` check, and only then did it confirm the finding.

- **Both step-4 arms RED-checked both directions**; reverting the rename reddens the structural arm (naming `RenderCounts()` at file:line) and the runtime arm independently.

Two defects were found by the verification machinery rather than by reading: the guard's first draft took only `out var` bindings, so `ValidateToFile`'s `is { } verr` sites — the fold's own fix — were invisible, and the mutation cell for them flagged nothing. And `return (xerr);` slipped the net until parentheses were unwrapped.

## Trace corrections

- #403 recorded **~39** `RenderError` call sites. The real count is **54** repo-wide, **44** on the read surface; ~39 is `RecordsTools.cs` alone.
- The build session's stage-1 report said **27** bare-text refusals were json-reachable. The true figure was **84**, from a regex that missed single-line `if (…) return "error: …";`.
- **"161 sites traced, zero untraced" was scoped to two files**, not to the read surface — `disposition.py`'s `FILES` list. That scoping is what left `housecarl_check` out. The disposition artifact on disk is the *post-fix* re-measurement (73 sites), not the 161-site pre-fix table; its five residual "defect" rows are `ParsePole`'s, correctly handled at its two call sites, because the classifier is call-site-blind.

The completeness claim is no longer carried by a count. It is carried by a guard that derives its own population, and #440 records the change that would make it true by construction rather than merely checked.

## Not in this PR

Remedy-grammar work is **#439**'s own follow-up PR, verified disjoint (0 overlapping lines) and deliberately not folded here. Its failing arm stays at `dev-scratch/ARM7_remedy_repro.patch`, uncommitted. The non-refusal duplicated-sentence tail is **#438**. The MO2 config prompt is returned bare by the whole tool surface including the write tools — a separate surface-wide question, excluded from the guard by name with that reason stated.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -15,12 +15,26 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 - **A read tool asked for json now refuses in json.** Reads that take `format="json"` previously answered
   some refusals with a plain sentence instead of a json document, so a caller parsing the response had to
-  handle two shapes depending on which rule it broke. Every refusal those tools can reach on a json call is
-  now one document — `ok:false` plus the `error` sentence, and the build stamp where the refusal consulted
-  one. A row that failed inside a call that succeeded is not a refusal and keeps its own `error` field, with
-  no `ok`: that is how a served answer is told from a refused one. Text output is unchanged. The refusals
-  still outside this shape are the ones whose own render says so: the tools with no json format, the
-  `format='dense'` refusals, and an unreadable `format=` itself, which cannot know the shape you wanted.
+  handle two shapes depending on which rule it broke. Those refusals are now one document — `ok:false` plus
+  the `error` sentence, and the build stamp where the refusal consulted one. A row that failed inside a call
+  that succeeded is not a refusal and keeps its own `error` field, with no `ok`: that is how a served answer
+  is told from a refused one. The refusals still outside this shape are the ones whose own render says so:
+  the tools with no json format, the `format='dense'` refusals, and an unreadable `format=` itself, which
+  cannot know the shape you wanted. Which refusals are in and which are out is enforced by
+  `refusal-completeness-guard`, which derives the covered surface rather than working from a list.
+
+- **`housecarl_records` with `counts_only=true` renames one json field: `ok` is now `resolved`.** The census
+  reported `{count, ok, errors}` with `ok` holding the number of inputs that resolved — the same key the
+  refusal shape above uses to mean "this call was refused", in a different type. A caller reading the
+  response as a refusal check saw `"ok": 0` on a served census where nothing resolved. The field pairs with
+  `errors` beside it now. The `ok:false` refusal discriminant is unchanged, and the text render still says
+  `ok=`. **If you parse the json census, update the field name.**
+
+- **Two read tools refuse an unreadable `format=` before anything else.** `housecarl_batch_record_detail`
+  and `housecarl_resolve` checked for an empty `formids=` first, which meant that one refusal could not be
+  a json document however you asked for it. The transport is settled first now. If a call breaks both rules
+  at once — an unreadable `format=` *and* an empty `formids=` — you are told about the `format=` where you
+  used to be told about the `formids=`. Apart from that ordering, text output is unchanged.
 
 - **houseCARL no longer edits a localized plugin in place, and now tells you where that plugin's text
   actually lives.** A localized plugin keeps its text in separate `.STRINGS`/`.DLSTRINGS`/`.ILSTRINGS`
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -18,10 +18,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   handle two shapes depending on which rule it broke. Those refusals are now one document — `ok:false` plus
   the `error` sentence, and the build stamp where the refusal consulted one. A row that failed inside a call
   that succeeded is not a refusal and keeps its own `error` field, with no `ok`: that is how a served answer
-  is told from a refused one. The refusals still outside this shape are the ones whose own render says so:
-  the tools with no json format, the `format='dense'` refusals, and an unreadable `format=` itself, which
-  cannot know the shape you wanted. Which refusals are in and which are out is enforced by
-  `refusal-completeness-guard`, which derives the covered surface rather than working from a list.
+  is told from a refused one. The refusals still outside this shape are the tools with no json format, the
+  `format='dense'` refusals, and an unreadable `format=` itself, which cannot know the shape you wanted —
+  plus an internal failure or an unconfigured instance, which are not refusals of your call. Which refusals
+  are in and which are out is enforced by `refusal-completeness-guard`, which derives the covered surface
+  rather than working from a list.
 
 - **`housecarl_records` with `counts_only=true` renames one json field: `ok` is now `resolved`.** The census
   reported `{count, ok, errors}` with `ok` holding the number of inputs that resolved — the same key the
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,6 +13,15 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A read tool asked for json now refuses in json.** Reads that take `format="json"` previously answered
+  some refusals with a plain sentence instead of a json document, so a caller parsing the response had to
+  handle two shapes depending on which rule it broke. Every refusal those tools can reach on a json call is
+  now one document — `ok:false` plus the `error` sentence, and the build stamp where the refusal consulted
+  one. A row that failed inside a call that succeeded is not a refusal and keeps its own `error` field, with
+  no `ok`: that is how a served answer is told from a refused one. Text output is unchanged. The refusals
+  still outside this shape are the ones whose own render says so: the tools with no json format, the
+  `format='dense'` refusals, and an unreadable `format=` itself, which cannot know the shape you wanted.
+
 - **houseCARL no longer edits a localized plugin in place, and now tells you where that plugin's text
   actually lives.** A localized plugin keeps its text in separate `.STRINGS`/`.DLSTRINGS`/`.ILSTRINGS`
   files and carries only indices into them. houseCARL previously committed such a plugin without the
--- a/src/housecarl-generator/RefusalCompletenessGuardProbe.cs
+++ b/src/housecarl-generator/RefusalCompletenessGuardProbe.cs
@@ -264,17 +264,30 @@ public static class RefusalCompletenessGuardProbe
         var seen = new HashSet<int>();
         foreach (var scope in scopes)
         {
-            // Names bound by an `out var X` ARGUMENT — the shape of a refusal produced inside a helper and handed
-            // straight back. Deliberately NOT every SingleVariableDesignation: `is { } prompt` binds the same way
-            // and the MO2-not-configured prompt it binds is trained guidance addressed to the model, not a
-            // refusal sentence — it is returned bare by the whole tool surface, write tools included, and is a
-            // separate surface-wide question rather than part of this grammar.
+            // Names bound to a string a HELPER produced — the shape of a refusal handed straight back, which is
+            // how twelve of the fourteen escaped. Two bindings carry it, and both must be in the net: an
+            // `out var X` argument (Wire.WantsJson, ExpandListInput's tuple) and an `is { } X` pattern
+            // (Artifacts.ValidateToFile). An earlier draft took only the first, and the ValidateToFile sites —
+            // the fold's own fix — went invisible; the mutation sweep's cell for them came back with nothing
+            // flagged, which is how the gap was found.
+            //
+            // One binding is excluded, structurally and by name rather than by file: a designation bound from
+            // ConfigPromptOrNull(). What that returns is trained guidance addressed to the model when no MO2
+            // instance is configured — not a refusal sentence, returned bare by the whole tool surface including
+            // the write tools, and a separate surface-wide question rather than part of this grammar.
             var outNames = new HashSet<string>(StringComparer.Ordinal);
             foreach (var arg in scope.DescendantNodes().OfType<ArgumentSyntax>())
             {
                 if (!arg.RefKindKeyword.IsKind(SyntaxKind.OutKeyword)) continue;
                 if (arg.Expression is DeclarationExpressionSyntax { Designation: SingleVariableDesignationSyntax sv })
                     outNames.Add(sv.Identifier.Text);
+            }
+            foreach (var pat in scope.DescendantNodes().OfType<IsPatternExpressionSyntax>())
+            {
+                if (pat.Pattern is not RecursivePatternSyntax { Designation: SingleVariableDesignationSyntax pv })
+                    continue;
+                if (pat.Expression.ToString().Contains("ConfigPromptOrNull", StringComparison.Ordinal)) continue;
+                outNames.Add(pv.Identifier.Text);
             }
 
             foreach (var ret in scope.DescendantNodes().OfType<ReturnStatementSyntax>())
--- a/src/housecarl-generator/RefusalCompletenessGuardProbe.cs
+++ b/src/housecarl-generator/RefusalCompletenessGuardProbe.cs
@@ -108,6 +108,38 @@ public static class RefusalCompletenessGuardProbe
         }
         """;
 
+    /// <summary>THE SECOND KNOWN-RED FIXTURE — <c>ReadTools.cs:101</c> under Aaron's gate-review mutation: the
+    /// refusal <c>Artifacts.ExpandListInput</c> hands back through a tuple, returned bare. This shape defeated the
+    /// net until the deconstruction designation joined it, and the miss was invisible to every other cell in the
+    /// sweep because <c>is { } verr</c> — the shape one line below it in the same file — WAS in the net. The
+    /// enumerator must flag this; eight live sites carry it.</summary>
+    const string HelperTupleFixture = """
+        static class Fixture
+        {
+            static string Body()
+            {
+                bool json = Wire.WantsJson(format, out var ferr);
+                var (toks, demand, echoSrc, xerr) = Artifacts.ExpandListInput(formids, "formids");
+                if (xerr is not null) return xerr;
+                return Ok();
+            }
+        }
+        """;
+
+    /// <summary>Its counter-fixture: the same tuple, wrapped. Silence here is an arm too.</summary>
+    const string WrappedTupleFixture = """
+        static class Fixture
+        {
+            static string Body()
+            {
+                bool json = Wire.WantsJson(format, out var ferr);
+                var (toks, demand, echoSrc, xerr) = Artifacts.ExpandListInput(formids, "formids");
+                if (xerr is not null) return Wire.Refuse(json, xerr);
+                return Ok();
+            }
+        }
+        """;
+
     public static int RunGuard(string[] args)
     {
         _pass = _fail = 0;
@@ -125,6 +157,15 @@ public static class RefusalCompletenessGuardProbe
         Check(greenHits.Count == 0,
               $"…and the wrapped form of the same shape is NOT flagged (found {greenHits.Count}, expected 0) — "
               + "an enumerator that flags everything proves nothing");
+
+        var tupleRed = Enumerate("<fixture>", HelperTupleFixture, out var tupleErrors);
+        Check(tupleErrors.Count == 0, $"the tuple-deconstruction fixture parses ({string.Join("; ", tupleErrors)})");
+        Check(tupleRed.Count == 1,
+              $"KNOWN-RED FIXTURE: a refusal handed back through a deconstructed tuple is FLAGGED "
+              + $"(found {tupleRed.Count}, expected 1) — the binding is a designation, not an out argument");
+        var tupleGreen = Enumerate("<fixture>", WrappedTupleFixture, out _);
+        Check(tupleGreen.Count == 0,
+              $"…and the wrapped form of the same tuple is NOT flagged (found {tupleGreen.Count}, expected 0)");
 
         // ---- 2. the population, derived from the artifact ----------------------------------------------
         Console.WriteLine();
@@ -278,11 +319,13 @@ public static class RefusalCompletenessGuardProbe
         foreach (var scope in scopes)
         {
             // Names bound to a string a HELPER produced — the shape of a refusal handed straight back, which is
-            // how twelve of the fourteen escaped. Two bindings carry it, and both must be in the net: an
-            // `out var X` argument (Wire.WantsJson, ExpandListInput's tuple) and an `is { } X` pattern
-            // (Artifacts.ValidateToFile). An earlier draft took only the first, and the ValidateToFile sites —
-            // the fold's own fix — went invisible; the mutation sweep's cell for them came back with nothing
-            // flagged, which is how the gap was found.
+            // how twelve of the fourteen escaped. THREE bindings carry it on this surface, and each is its own
+            // syntax: an `out var X` argument (Wire.WantsJson), a tuple deconstruction `var (…, X) = Helper(…)`
+            // (Artifacts.ExpandListInput), and an `is { } X` recursive pattern (Artifacts.ValidateToFile). Each
+            // one was missing from a draft of this net and each miss was found the same way — by a mutation cell
+            // that reverted the site and came back with nothing flagged. ExpandListInput's tuple is NOT an out
+            // argument: its left side is a DeclarationExpressionSyntax carrying a ParenthesizedVariableDesignation,
+            // so the `out var` walk below cannot see it and an earlier comment here said otherwise.
             //
             // One binding is excluded, structurally and by name rather than by file: a designation bound from
             // ConfigPromptOrNull(). What that returns is trained guidance addressed to the model when no MO2
@@ -294,6 +337,14 @@ public static class RefusalCompletenessGuardProbe
                 if (!arg.RefKindKeyword.IsKind(SyntaxKind.OutKeyword)) continue;
                 if (arg.Expression is DeclarationExpressionSyntax { Designation: SingleVariableDesignationSyntax sv })
                     outNames.Add(sv.Identifier.Text);
+            }
+            foreach (var decl in scope.DescendantNodes().OfType<DeclarationExpressionSyntax>())
+            {
+                if (decl.Designation is not ParenthesizedVariableDesignationSyntax pd) continue;
+                if (decl.Parent is AssignmentExpressionSyntax asg
+                    && asg.Right.ToString().Contains("ConfigPromptOrNull", StringComparison.Ordinal)) continue;
+                foreach (var v in pd.DescendantNodesAndSelf().OfType<SingleVariableDesignationSyntax>())
+                    outNames.Add(v.Identifier.Text);
             }
             foreach (var pat in scope.DescendantNodes().OfType<IsPatternExpressionSyntax>())
             {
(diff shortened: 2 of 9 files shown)


src/housecarl-generator/RefusalCompletenessGuardProbe.cs
```diff
@@ -0,0 +1,456 @@
...
+            // (Artifacts.ValidateToFile). An earlier draft took only the first, and the ValidateToFile sites —
+            // the fold's own fix — went invisible; the mutation sweep's cell for them came back with nothing
+            // flagged, which is how the gap was found.
+            //
+            // One binding is excluded, structurally and by name rather than by file: a designation bound from
+            // ConfigPromptOrNull(). What that returns is trained guidance addressed to the model when no MO2
+            // instance is configured — not a refusal sentence, returned bare by the whole tool surface including
+            // the write tools, and a separate surface-wide question rather than part of this grammar.
+            var outNames = new HashSet<string>(StringComparer.Ordinal);
+            foreach (var arg in scope.DescendantNodes().OfType<ArgumentSyntax>())
```
Avick3110, line 433: **[high] The enumerator's helper-binding net misses the tuple deconstruction — 8 live sites, confirmed by mutation.**

`outNames` is filled from two shapes only: an `out var X` argument, and an `is { } X` recursive pattern. `Artifacts.ExpandListInput`'s binding is neither — `var (toks, demand, echoSrc, xerr) = Artifacts.ExpandListInput(...)` parses as an assignment whose left side is a `DeclarationExpressionSyntax` carrying a `ParenthesizedVariableDesignationSyntax`, not an `ArgumentSyntax` and not an `IsPatternExpressionSyntax`. So `xerr` never enters `outNames`, `RefusalLiteral` finds no literal on `return xerr;`, and the return is skipped.

The comment at line 281–284 states the opposite: *"an `out var X` argument (Wire.WantsJson, ExpandListInput's tuple)"*. ExpandListInput's tuple is not an out argument.

Verified on a worktree at 221d279:

```
sed -i '101s/return Wire.Refuse(json, xerr);/return xerr;/' src/housecarl-mcp/ReadTools.cs
dotnet run --project src/housecarl-generator -- refusal-completeness-guard
```
→ `no bare whole-call refusal in the population is without a ruling (found 0)` … `8 passed, 0 failed -> PASS`.

Live sites in the class: `ReadTools.cs:101, 282, 407`; `RecordsTools.cs:442, 957, 1012, 1307, 1327`. That is the same twelve-helper-returned class round 1 found; the mutation table's `ALL ReadTools sites → prose` cell reverted the literal-bearing sites, so this arm of the sweep was never exercised — the `verr` cell was, because `is { } verr` *is* in the net.

src/housecarl-generator/RefusalCompletenessGuardProbe.cs
```diff
@@ -0,0 +1,456 @@
...
+            }
+        }
+        return files;
+    }
+
+    /// <summary>A <c>Guard.Tool(...)</c> invocation's body lambda, when that body consults the format machinery.
+    /// This is the definition of "on the json-capable read surface" — a tool that never asks about the transport
+    /// has no transport to honour, which is why <c>housecarl_effect_chain</c> is correctly absent without an
+    /// allowlist entry, and why a NEW json-capable tool enrols itself the day it is written.</summary>
+    static IEnumerable<SyntaxNode> ToolBodies(SyntaxNode root)
```
Avick3110, line 356: **[high] The derived population stops at `Guard.Tool` lambdas, so a transport-aware helper is outside the net — confirmed by mutation.**

`ToolBodies` yields only lambdas passed to `Guard.Tool`. A refusal produced in a file-scope helper that is *handed* the transport is therefore never enumerated. `RecordsTools.RenderListAggregate(…, bool json, …)` at `RecordsTools.cs:1875` is exactly that shape.

Verified on a worktree at 221d279 — rewriting that line to the bare form:

```
return $"error: project.group_by='{groupBy}' is not a count key — use 'winner', 'type', or 'defined_in'.";
```
→ `found 0` unexplained, `8 passed, 0 failed -> PASS`.

The no-transport helper case (`ParsePole`) is covered because its call sites are in a tool body. The transport-carrying helper case is not covered anywhere: the refusal is produced *and* returned inside the helper, so no tool-body return ever carries it. Deriving on "a method that takes a `bool json`/`QueryFormat` parameter" as well as "a `Guard.Tool` body" would close it without adding a named file.

src/housecarl-generator/RefusalCompletenessGuardProbe.cs
```diff
@@ -0,0 +1,456 @@
...
+    // ================= allowlist matching =================
+
+    static (string File, string Fragment, string Decision, string Why)? Match(Hit h)
+    {
+        foreach (var a in Allow)
+            if (Matches(h, a)) return a;
+        return null;
+    }
+
+    static bool Matches(Hit h, (string File, string Fragment, string Decision, string Why) a)
```
Avick3110: **[low] Allowlist matching is an unanchored substring against two different kinds of `Hit.Sentence`.**

`Hit.Sentence` is the *identifier name* for a binding hit but the *full message* for a literal hit, and `Matches` accepts any hit containing the fragment. With the two surface-wide entries spelled `ferr` and `fmtErr`, a future refusal **literal** containing that substring is ruled correct instead of flagged — e.g. `"error: the record was transferred from a plugin outside the order …"` matches `("*", "ferr", "#7")` and drops out of `unexplained`.

Nothing live hits it today (no `"error: …"` literal in `src/housecarl-mcp` contains `ferr`/`referred`/`deferred`/`transferred`), so this is latent — but the allowlist is the guard's only escape hatch, and a hole in it is silent by construction. Distinguishing the two hit kinds (or requiring an exact identifier match for binding entries) removes the ambiguity.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,29 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A read tool asked for json now refuses in json.** Reads that take `format="json"` previously answered
+  some refusals with a plain sentence instead of a json document, so a caller parsing the response had to
+  handle two shapes depending on which rule it broke. Those refusals are now one document — `ok:false` plus
+  the `error` sentence, and the build stamp where the refusal consulted one. A row that failed inside a call
+  that succeeded is not a refusal and keeps its own `error` field, with no `ok`: that is how a served answer
+  is told from a refused one. The refusals still outside this shape are the ones whose own render says so:
```
Avick3110: **[low] This sentence reads as a closed list, and at least two `error:`-shaped prose paths are missing from it.**

A `format="json"` caller can still receive prose from:

- `Guard.Named` (`Guard.cs:40`) — `"error: <tool> failed unexpectedly — …"` on any unhandled exception inside a read tool body;
- `svc.ConfigPromptOrNull()` — returned first by every tool body when no MO2 instance is configured.

Neither is "a tool with no json format", a `format='dense'` refusal, or an unreadable `format=`. A consumer that reads this line and writes `JSON.parse` unconditionally still has two prose paths to handle, and neither is enforced by the guard (the config prompt is excluded by name; `Guard.Named` is outside the derived population).

This is about the changelog sentence, not about re-opening the `Guard.Named` triage — that was refused as claim-scope, which is fine; the fix here is narrowing the claim, e.g. "the refusals still outside this shape are …, and an internal failure or an unconfigured instance, which are not refusals of your call".

src/housecarl-generator/RefusalCompletenessGuardProbe.cs
Avick3110: **Folded — `21aee61`.** Confirmed at source before folding: your mutation on `221d279` returned `PASS 8/8` with `found 0`.

The deconstruction designation is in the net now — `DeclarationExpressionSyntax` whose `Designation` is a `ParenthesizedVariableDesignationSyntax`, every `SingleVariableDesignationSyntax` under it (nested tuples included), with the same `ConfigPromptOrNull` exclusion the `is { }` walk carries. The comment at :281-284 no longer calls it an out argument; it names all three bindings and what each one actually is.

Your mutation is a permanent fixture (`HelperTupleFixture`) with a wrapped counter-fixture beside it, and a permanent sweep cell. Post-fix, from a committed state: `tuple-deconstruction FAIL, 1 unexplained`; canary red, control green.

**You were right about the mutation table, and it was worse than a wrong number.** The `ALL ReadTools → prose` cell reverted only the literal-bearing sites, so it never exercised the binding arm at all — which is exactly why the hole survived a sweep that read as exhaustive. That cell now rewrites **every** `Wire.Refuse(` in the file through a bypass helper whose name is not an approved renderer: 20 sites, **14 unexplained**. The other 6 are the ones passing an identifier (`xerr` ×3, `verr` ×3) — under a bypass the return is still an invocation, so there is nothing for the guard to flag, and those 6 are covered by the two dedicated per-shape cells instead. The table is corrected in the PR body and says so, including that a mass cell is not a substitute for the per-shape cells.

src/housecarl-generator/RefusalCompletenessGuardProbe.cs
Avick3110: **Folded — `2142c62`.** Confirmed at source first. (Worth noting: my first cell reproducing this was broken — anchored on the refusal sentence, it hit the *identical* sentence at `RecordsTools.cs:239` inside a tool body and came back red for the wrong reason. Re-anchored on the enclosing `gb` check, the unmutated guard returned `PASS`, `found 0`, as you had it.)

Widened per your suggestion: a scope is policed when the transport is **decided in it** (a `Guard.Tool` body consulting the format machinery) **or handed to it** (a method or local function declaring `bool`/`bool? json`, or a `QueryFormat` of any name). Still derived from the artifact — no file named. Three live scopes join: `Wire.Refuse`, `RenderRecordsSummary`, `RenderListAggregate`.

**The escalation condition did not fire.** Nothing new came back flagged — `found 0` unexplained, 0 stale — so no allowlist entry was added, silently or otherwise. `ParsePole` stays correctly outside: no transport is decided in it or handed to it, and its call sites are policed as before.

One thing the fold added on its own: `QueryFormat` has no live parameter anywhere on the surface today, so that half of the rule would have shipped unproven. There is now a population fixture — three helpers, none of them a `Guard.Tool` body, enumerated with the population filter on: the `bool json` one and the `QueryFormat` one are policed, the no-transport one is not (settled #8).

Post-fix, from a committed state: `transport-helper FAIL, 1 unexplained`; canary red, control green. Your mutation is a permanent sweep cell.

src/housecarl-generator/RefusalCompletenessGuardProbe.cs
Avick3110: **Folded — `010119b`.** Entries are typed now. A `Hit` carries a `HitKind` (`Binding` when the sentence is an identifier the body returned, `Literal` when it is the message), each allowlist entry declares its kind, and `Matches` requires the kinds to agree: a binding entry must **equal** the identifier it names, a literal entry matches a substring of message text. No cross-kind match.

Your latent-hole example is the fixture, verbatim: one policed helper yielding both hits — the binding `ferr`, and `"error: the record was transferred from a plugin outside the order."`. Three arms: the fixture yields exactly one hit of each kind, the binding entry rules the identifier, and the sentence stays unexplained.

RED-checked from a committed state by reopening the hole rather than by breaking the tree — deleting the kind conjunct from `Matches` reddens that arm (`allowlist-kind-check FAIL`), with the canary red and the control green. It is a permanent sweep cell, the only one that sabotages the guard instead of the source, because that is where the hole was.

The live allowlist is unchanged in meaning: `ferr` and `fmtErr` are binding entries, the two `format='dense'` sentences are literal entries, and the run reports 0 stale.

plugin/CHANGELOG.md
Avick3110: **Folded — `3923159`.** Both paths verified at source before the sentence was touched, not taken on the finding's word:

- `Guard.cs:40` — `Named` returns `"error: {tool} failed unexpectedly — …"`, reached from `Guard.Tool`'s `catch (Exception ex)`, so it wraps every read tool body.
- `LoadOrderService.cs:2245` — `ConfigPromptOrNull()` returns `NotConfiguredText`, which is not even `error:`-shaped; 20 tool bodies return it first.

Reworded to your shape: *"…the tools with no json format, the `format='dense'` refusals, and an unreadable `format=` itself, which cannot know the shape you wanted — plus an internal failure or an unconfigured instance, which are not refusals of your call."*

The `refusal-completeness-guard` pointer stays, and stays true: what it enforces is which **refusals** are in and which are out, and neither new item is a refusal of the caller's call. The `Guard.Named` triage stays refused as claim-scope — this narrows the claim to match that ruling rather than reopening it.

Avick3110:
## Fold wave complete — verification, from a committed state

Four directed folds, one commit each, RED-checked individually as they landed. No new review rounds (§11 — fresh eyes are for before the PR opens).

| Commit | Fold |
|---|---|
| `21aee61` | the tuple deconstruction joins the binding net, with its fixture pair |
| `2142c62` | a helper handed the transport joins the policed population, with a derivation fixture |
| `010119b` | allowlist entries are typed; no cross-kind match |
| `3923159` | the CHANGELOG claim is narrowed to two verified prose paths |

### `ci-all` — 132/132 ALL PASS (2.05 min)

`refusal-completeness-guard` grew from 8 arms to 18: three new fixtures (tuple-deconstruction red + wrapped green, the population fixture, the allowlist-kind fixture) and their arms.

### Full mutation sweep — `dev-scratch/completeness_sweep.sh`, canary intact

Every cell from a committed state; each restore verified, not assumed.

| Cell | Verdict | Unexplained |
|---|---|---|
| canary — the `:223` ternary back to bare arms | FAIL (expected) | 1 |
| tuple-deconstruction — `ReadTools.cs:101` → `return xerr;` | FAIL | 1 |
| transport-helper — `RenderListAggregate` → bare | FAIL | 1 |
| one `ReadTools` site → prose | FAIL | 1 |
| helper-returned `verr` revert | FAIL | 1 |
| `CheckTools` `famErr` revert | FAIL | 1 |
| ALL `ReadTools` → prose (20 sites bypassed) | FAIL | 14 |
| ALL `RecordsTools` → prose (74 sites bypassed) | FAIL | 66 |
| allowlist kind check removed | FAIL | 0 |
| control — an unrelated comment edit | PASS | 0 |

Two corrections carried into the PR body rather than left in a log: the `ALL ReadTools` cell now genuinely reverts all 20 sites (it previously reverted only the literal-bearing ones, which is why the tuple hole survived it), and the 6 that stay unflagged under a mass bypass are accounted for explicitly — they pass an identifier, so the bypassed return is still an invocation, and the per-shape cells are what cover them.

### Byte-identity — `dev-scratch/literal_identity.py`, canary-tested

Rewritten (the earlier one was not in the worktree) and re-run over `2fd7b6e..HEAD`, whole of `src/housecarl-mcp`, scanner rather than line regex — comments skipped, regular / verbatim / interpolated / raw forms recognised:

- refusal-bearing literals: **267 base / 268 head, zero changed sentences**. The one addition is `RefusalPrefix` = `"error: "`.
- canary: perturbing one head-side sentence by a single character is reported.
- served-schema keys, reported separately and outside the identity verdict: `resolved: 1→2`, `error: 25→15`.

Counts differ from the pre-review figures because that extractor scoped itself to the changed files; the deltas agree, and this one states its scope in the output.

---

**Stopping here — your go is the merge gate.** On it: remove the worktree, `gh pr merge 441 --rebase --delete-branch`, confirm #403 closed, delete the local branch. #439 (PR B) is kicked off after that, by the advisor, from this branch's unmoved ground.

Issue #449, opened 2026-08-28
[Bug] to_file= artifact rows carry scoped field values without the scoped-vs-winner note


Avick3110:
All three inline transports warn a caller that a `plugins=`-scoped `fields` scan is showing the
scoped plugin's own values rather than the live winner. The artifact written by `to_file=` does
not, and it carries the same values.

## Repro

```
housecarl_records(types: ["WEAP"], plugins: { names: ["<a defining esp>"] },
                  project: { form: "fields", fields: ["BasicStats.Damage"] })
```

Inline (text, json and dense all carry it):
```
note: field values are each match's SCOPED plugin's OWN version, NOT the live load-order
winner — pass fields_source="winner" for load-order truth.
```

The same call with `to_file="<path>.jsonl"` writes rows whose field values are the scoped
bodies, and no such note appears in the manifest or the rows.

## Why it bites

This is the silent-wrong trap the note exists for — a defining esp's `Damage 38` versus the live
winner's `200`. A caller who spills to an artifact and reads it later gets the number with nothing
attached saying which body it came from. The per-row `source` column names the plugin, but the
inline note exists precisely because the column alone was judged insufficient.

## Where

`Artifacts.WriteCrossQuery` (`Artifacts.cs:130`) writes rows and a manifest; the P5 note is
composed in the render path (`JsonWire.ScopedFieldsNote`, reached from `ReadTools.cs:1057`,
`JsonWire.cs:1043`, `:1147`) and has no artifact equivalent. Pre-existing; the #439 branch threaded
the artifact writer's container hint but did not add the note.

## Fix

Carry the note into the artifact — the manifest is the natural home, since it is the row set's
one header and is what a re-entering caller reads first.

## Provenance

Found by round 2 of PR B of RUN_ORDER row 6 (#439). Outside that branch's changed lines — the
#324 shape.

Unrouted and unmilestoned: routing is Aaron's stroke.

Issue #448, opened 2026-08-28
[Bug] housecarl_records format=dense silently drops project.depth>1 (the 1.x sibling refuses it loud)


Avick3110:
A `housecarl_records` dense call that asks for expanded containers gets collapsed cells, no
refusal and no note. The 1.x sibling refuses the same combination by name.

## Repro

```
housecarl_records(types: ["SPEL"], format: "dense",
                  project: { form: "fields", fields: ["Effects"], depth: 4 })
```
→ depth-1 collapsed cells, no refusal, no note.

The identical call at `format="text"` expands to eight lines per record.

`housecarl_cross_plugin_query` with `format="dense"` and `depth>1` refuses:
`ReadTools.cs:274` — *"depth>1 is not carried in format='dense'…"*.

## Why

`RecordsTools.cs:356-362` refuses dense for the `everything`/`aggregate`/`delta`/`tree`/`info_order`
forms but has no depth check, and `RecordsTools.cs:1255` calls `JsonWire.RenderCrossQueryDense`,
whose signature carries no `depth` parameter at all — so the value has nowhere to go.

Q3: a silently degraded answer. The caller asked for depth 4, got depth 1, and nothing in the
response says so.

Present on base (`fbfe015:src/housecarl-mcp/RecordsTools.cs:1253`) — not a regression.

## Note

The dense container hint added by #439's fix (*"pass project.depth=2 with format=text/json to
expand (dense cells are positional)"*) tells the caller to hop formats, which is correct advice
and also the thing that papers over this: the hint explains the limitation while the parameter
still accepts a value it discards.

## Fix

Refuse `project.depth > 1` under `format="dense"` by name, matching the 1.x sibling's refusal.

## Provenance

Found by round 2 of PR B of RUN_ORDER row 6 (#439), driven. Outside that branch's changed lines.

Unrouted and unmilestoned: routing is Aaron's stroke.

Avick3110:
Checked while working this: already fixed on `main` by 64a730b ("dense refuses a depth it cannot render, instead of dropping it"), which added the `scanLane && dense && project.depth > 1` refusal in `RecordsTools.cs` plus three tests in `RecordsScanRefusalRepairTests.cs` (the pairing refused naming the depth, dense at the default depth still served, the same depth in `format='text'` still served). The list lane's own dense refusal covers the `formids=` half. Verified those tests pass on the current tree, so this is closeable as done rather than by #566.

Avick3110:
[...]

Issue #447, opened 2026-08-28
[GAP] records-guard arm 7 covers ~10 of 29 carrier seams; known-GREEN sabotage cells listed


Avick3110:
The remedy-vocabulary arm (`RecordsGuardProbe.cs`, arm 7) asserts *"no sentence this tool emits
names a lever it lacks"*. It drives a probe list, and the list does not reach most of the seams
the carrier is threaded through: `RecordsTools.cs` has 29 `LeverNames.Records` call sites and the
arm reaches roughly ten of them.

## Known-GREEN cells — sabotage that the guard does not catch

Measured on `claude/remedy-vocabulary` (round 2), each run from a committed tree with a
hand-proven RED canary first:

| Sabotage | Verdict |
|---|---|
| Revert `LeverNames.Records` → `Legacy` on 14 lines: `RecordsTools.cs:1156, 1172, 1187, 1192, 1224, 1225, 1263, 1414, 1430, 1431, 1456, 1463, 1464, 1470` | **GREEN** |
| Revert the `delta`-form hand-spelled sites to `Legacy.Fields` (before that branch's clause cells were added) | **GREEN** |
| `clause/delta-identical` — `Narrow with project.fields= to compare in full` (`RecordsTools.cs:1585`, `:1606`) | **GREEN, and un-drivable** |

With the first applied, `project.form="everything"` and the off-order **scan** lane emit
`narrow with fields=` and `pass depth=2 to expand` — the original #439 defect — while the guard
reports PASS.

## The uncovered call shapes

- `project={"form":"everything"}` on a scan — bodies at `RecordsTools.cs:1156/1172/1187/1192`,
  renders at `:1224/:1225`. (The defect itself is #445.)
- off-order file pole **+ scan** — `:1414/:1430/:1431`. Distinct from the `poleOff/*` probes,
  which pass `formids=` and land on `:552`.
- the ceiling auto-spill artifact writer — `:1263/:1470`. Only the explicit `to_file=` writer at
  `:1248` is read back.

## The un-drivable clause

`clause/delta-identical` covers *"the deep read was TRUNCATED … Narrow with project.fields= to
compare in full"*. That sentence needs `FieldsDiff` to hit its expansion cap, i.e. a record nested
deeper than `ConflictDiffDepth = 16`, and the guard fixture has no way to build one. It is
declared in the sweep as a known-GREEN cell rather than deleted, so the gap stays visible.

## Why this is filed rather than fixed

[...]

Issue #446, opened 2026-08-28
[Bug] housecarl_records walk/chain: 'raise limit=' is a no-op remedy for the walk.max_nodes bound


Avick3110:
`Wire.RenderEffectChain` (`ReadTools.cs:1143`) takes no lever vocabulary at all, and
`RecordsTools.cs:1772` renders records' walk/chain responses through it. The sentence it
composes names a 1.x lever that does nothing on this tool.

## Repro

```
housecarl_records(formids: [<MGEF>], walk: { direction: "reverse", max_nodes: 1 },
                  project: { form: "chain" })
```

```
[!] 1 seed(s) hit the walk.max_nodes carrier bound (1) — their rows are a prefix of
    carrier_total; raise walk.max_nodes.
…
effect_chain for 000805:… (HcRecMgefFire, MagicEffect): 2 carrier rows
    (showing first 1; raise limit= or narrow to see more)
```

Re-run with `limit=5000`: **byte-identical output, still `showing first 1`.**

## Why

On `housecarl_records`, `limit=`/`offset=` window the SEEDS; the carrier bound is
`walk.max_nodes` (`RecordsTools.cs:631` says so in as many words). The envelope line gets this
right — `raise walk.max_nodes` — and the row line, composed in the shared 1.x renderer, does not.
So one response carries two remedies for the same cap and the more prominent one is inert.

Q3 shape: a silently degraded outcome rather than a refusal. The caller raises `limit=`, sees no
change, and has no way to tell the knob was wrong from the response.

Text only; the json lane emits `capped: true` with no sentence.

## Fix

Thread the carrier into `RenderEffectChain` (it currently has no parameter for one) and name
`walk.max_nodes` for a records caller, `limit=` for `housecarl_effect_chain`'s own callers.

## Provenance

Found by round 2 of PR B of RUN_ORDER row 6 (#439), driven rather than read. This is the finding
that met the class-stop condition on that branch. Not folded: the renderer is outside the
branch's changed lines and the branch ships a claim scoped to exclude the walk/chain lane.

Unrouted and unmilestoned: routing is Aaron's stroke.

PR #450, opened 2026-08-28
Remedy sentences name the calling tool's own levers (scoped claim)

### LINKED ISSUE #439: [Bug] truncation notices tell a housecarl_records caller to narrow with fields= — it has project.fields=

Avick3110:
Fixes #439

A remedy predicts what a LATER call produces, so it is only true for a caller that has the
parameters it names. The body renderers are shared by both tool generations, so a remedy composed
inside one could only speak one vocabulary — and it spoke the 1.x one. A `housecarl_records`
caller was told to narrow with `fields=` and lower `depth=`, which on that tool are
`project.fields=` and `project.depth=`.

`LeverNames` carries the caller's spelling into the renderer. `Legacy` is the default at every
seam, so 1.x call sites render byte-identically; `housecarl_records` passes `Records`.

---

## The claim this PR makes

> `housecarl_records` truncation and expansion notices name that tool's own parameters on the
> **`fields`, `tree` and `delta`** forms and the **`summary` form's scan cut**, for the
> **collapsed-container hint on every `source=` lane**, and for the **scoped-vs-winner note** —
> verified on `format="text"` and `"json"`, with the container hint additionally verified on
> `"dense"` and in `to_file=` artifact rows. A truncation cut's own **selection** and **slim-down**
> clauses name the lever that is real for the CALL — its selection lane, and whether it passed
> anything to slim with — verified on `"text"`, the only format that writes them.

It is deliberately **not** a completeness claim. An earlier draft was, twice, and both times a
review round proved it false. What is **not** covered is named in the CHANGELOG and filed:
the `everything` form (#445) and the walk/chain lane (#446).

**Every clause has its own sabotage-RED cell.** A clause with no cell is not in the claim.

| Clause | Cell | Verdict |
|---|---|---|
| the records vocabulary itself (canary, hand-proven at `2b037a2`) | `canary/records-selector` | RED |
| the depth knob's spelling | `records-depth` | RED |
| the absent lever (`conflict_tree`) | `records-conflict-tree` | RED |
| the renamed lever (`winner_fields` to `fields_source`) | `records-winner-fields` | RED |
| the `=` on the selector | `clause/selector-equals` | RED |
| the scan slim-down stays actionable | `clause/slim-scan` | RED |
| `delta` form, text | `clause/delta-text` | RED |
| `tree` form, text | `clause/tree-text` | RED |
| field truncation, json / text | `json/WriteFieldsArray-carrier`, `text/AppendRecord-carrier` | RED |
| tool entry, json / text | `entry/json-RenderBatch`, `entry/text-RenderBatch` | RED |
| container hint: winner / active pole / off-order pole / overlay post | `entry/container-hint`, `pole/active-arm`, `pole/offorder-arm`, `overlay/post` | RED |
| the tree row's hand-diverged literal | `json/WriteTreeRow-carrier` | RED |
| dense transport | `dense/carrier` | RED |
| artifact rows | `spill/carrier` | RED |
| P5 note: text / json / dense | `p5/text`, `p5/json`, `p5/dense` | RED |
| the truncated-comparison sentence (was a declared GREEN) | `clause/delta-identical` | RED |
| the batch cut's selection clause: scan lane / formids lane / both deciding call sites | `gate1/scan-lane`, `gate1/formids-lane`, `gate1/entry-everything`, `gate1/entry-offorder` | RED |
| the harvest's capitalisation blind spot | `gate2/ignorecase` | RED |
| the scan cut's slim-down clause: omit branch / keep branch / both deciding call sites | `gate3/omit-branch`, `gate3/keep-branch`, `gate3/inorder-callsite`, `gate3/offorder-callsite` | RED |
| artifact ROW writers: cross-query / batch | `gate4/crossquery-row`, `gate4/batch-row` | RED |

**34 RED, canary RED, tree clean after every restore, and nothing declared GREEN.** The one cell
that was declared rather than deleted — `clause/delta-identical`, the "comparison TRUNCATED …
compare in full" sentence — is a live RED cell now: the gate review's second finding produced the
fixture that drives it, so that sentence is in the claim and #447 loses the item.

**1.x identity:** 14 of 14 sites compose byte-identically to their pre-fix literals under `Legacy`,
checked against the blobs at `fbfe015` — re-run after the gate folds and unchanged, which is the
point: the lane-aware clauses must not move any 1.x output, and they do not. One declared deviation:
`WriteTreeRow`, which was hand-written in the records spelling and is carrier output now
(RED-checked specifically). Two sites still hold a bare literal and are correct as they stand:
a 1.x tool's own `Description` prose, and `Wire.RenderPluginFile`, whose only caller is
`housecarl_read_plugin_file` — both unreachable from records, verified by caller grep.

`ci-all` **132/132 guards, 3012/3012 checks**. PR A's `refusal-completeness-guard` is green and
untouched — no arm of it changed on this branch.

## Overlap with PR A, re-measured against merged `main`

| File | PR-A-changed lines (`2fd7b6e..fbfe015`) | remedy-notice lines | overlap |
|---|---|---|---|
| `ReadTools.cs` | 56 | 10 | **0** |
| `JsonWire.cs` | 59 | 4 | **0** |
| `RecordsTools.cs` | 80 | 0 | **0** |

Non-vacuous: widening the range to the whole history makes the same detector report 10 overlaps.

---

## Review rounds

Two rounds, three fresh agents each, own worktrees, seeded with the decision record and the
settled-decisions list only. Both rounds ended in an escalation rather than a fold-and-continue.

### Round 1 — ~11 findings, 2 refused (18%)

Three seams the carrier had not been threaded through, found independently by all three:

- `ResolveBatchFromPole` and `OverlayPostBatch` had **no `containerHint` parameter at all**, so the
  four records call sites could not pass one — `source=` and the SkyPatcher post view kept saying
  `pass depth=2 to expand`. (`overlay:pre` was already correct, so two arms of one parameter
  disagreed.)
- `ScopedFieldsNote` named `winner_fields=true` on all three transports. That lever is *renamed* on
  records, not re-scoped.
- `format="dense"` and the spilled artifact rows were threaded but unasserted — removing the carrier
  from either left the guard **green**.

**Refused:** "the json-lane `conflict_tree` assertion is vacuous" — real but harmless, an
absence-assertion that cannot fail today and remains a live net. And "`drop project.fields=` is
only conditionally actionable" — **this refusal was wrong**, see round 2.

**Class stop → escalation.** The recurring class was the design property that `Legacy`-as-default
makes every missed seam silent. Aaron ruled: widen the fix, the guard and the claim on this branch;
file the no-default carrier design separately (#444).

### Round 2 — 12 findings, 0 refused

Nothing was refused because every finding was reproduced against source or driven as a probe.

- **The strike:** `Wire.RenderEffectChain` takes no carrier at all; a records reverse walk is told
  `raise limit=`, which on that tool windows the seeds while `walk.max_nodes` is the bound. Driven
  at `limit=5000` — byte-identical output. Filed as #446.
- **A new class, which #444 as filed would not have prevented:** the carrier is per-TOOL, records'
  levers are per-FORM. `form="everything"` names `project.fields=`, which that form refuses.
  Filed as #445.
- **The guard still covered ~10 of 29 seams.** Reverting the carrier on 14 lines left it green while
  `everything` and the off-order *scan* lane emitted the original defect. Filed as #447.
- **My overturned refusal:** on `form="fields"`, dropping `project.fields` refuses — the sentence's
  literal follow-up fails, which is this branch's own definition of the defect. Having the lever is
  not the test; the next call working is. Folded as `SlimScan`, which names `project= (summary rows)`
  for records and keeps `fields=/conflict_tree` byte-identical for 1.x. One assertion now *follows*
  the remedy rather than reading it: drop `project=` and require the summary render it promised.

**Class stop → escalation.** Aaron ruled (iii): land with the scoped claim, charter the
`(tool, form)` + required-argument redesign as its own work, and gate the push on every clause of
the scoped sentence having a sabotage-RED arm first.

### Three errors of mine, corrected on the branch

1. **A guard cell that went RED for the wrong reason.** `pole/offorder-arm`'s pattern matched twice
   and it was mutating an unrelated read path 2000 lines away. The sweep now asserts each cell's
   occurrence count. With that fixed the cell came back **GREEN** — no probe reached the off-order
   pole at all — which is commit `1aa74a3`.
2. **A CHANGELOG sentence that overclaimed** while correcting the previous one for overclaiming:
   "all of it applies on text, json and dense, and to artifact rows" was wrong in three places.
   Measured, dense composes no truncation notice and artifact rows carry the container hint alone.
3. **I widened a key allowlist instead of retiring it.** Round 2 then found `truncation_note` occurs
   **nowhere** in the product source — the spelling is `truncated_note` — so one of its three keys
   had never matched anything and `scan_note` was never listed. The harvest names no keys now.

---

## Gate review — Aaron's four findings, all folded

His independent review found four, all triaged FOLD. No new rounds were run (§11: fresh eyes are for
before the PR opens). One commit per finding, each RED-checked from a committed state.

| # | Finding | Fold | Cells |
|---|---|---|---|
| 1 | `request fewer formids` is inert on the two scan-derived lanes of `Wire.RenderBatch` | `d726117` — `LeverNames.BatchSelection` + `OnScanSelection()`; the call sites declare their lane | `gate1/*` (4) |
| 2 | `remedyLine` was case-sensitive, so a capitalised remedy was dropped before assertion | `5bc3a96` — `RegexOptions.IgnoreCase` + a permanent fixture that drives the hidden site | `gate2/ignorecase`, `clause/delta-identical` |
| 3 | `drop project= (summary rows)` is inert on the form this tool defaults to | `3e0026e` — `SlimScan` nullable; the clause is omitted where there is nothing to drop | `gate3/*` (4) |
| 4 | `levers` reaches the artifact read but not the row writer | `3117710` — both row writers take it, and a `rowCap` so the dormant note can be driven | `gate4/*` (2) |

**Finding 1 was inside the claim, not outside it.** The claim covers the `fields` form, and the
in-order body lane serves `form="fields"` under `plugins=`+`source=`. The claim as written was false
for that case; the fold is what makes it true.

**Finding 2 closed a declared GREEN.** `clause/delta-identical` was declared rather than dropped
because the fixture could not reach `!d.Complete`. The route is `ReadEngine`'s expansion budget, not
`ConflictDiffDepth`: a form list one element past `MaxExpandNodes` truncates, and a copy override
supplies the zero-delta half. That sentence is in the claim now, and #447 loses the item.

**Finding 4 was folded one method wider than written**, and deliberately: `Artifacts.WriteBatch` had
no `levers` parameter at all, writes rows through the same `WriteReadRecord`, and is reached by four
`housecarl_records` call sites. Aaron's stated reason for folding the cross-query line applies to it
word for word, so closing one and leaving the other would have left the hazard live. Called out on
the thread rather than widened quietly.

Findings 1 and 3 established a third axis for #444 — the vocabulary is a function of
**(tool, form, selection lane)** — commented there with both folds as the measured instances.

## Filed, not folded

| # | What |
|---|---|
| #442 | summary rows label override depth `depth=`, colliding with `project.depth=` (label class) |
| #443 | the `group_by=type` aggregate refusal names 1.x levers, composed in the scan engine |
| #444 | the no-default carrier design — **re-scoped** to `(tool, form)`, with #445 as its acceptance fixture |
| #445 | the per-form vocabulary class (`everything` form) |
| #446 | `RenderEffectChain`'s no-op `raise limit=` remedy |
| #447 | the guard's uncovered seams — the `clause/delta-identical` item is closed by the gate fold, the rest stand |
| #448 | dense silently drops `project.depth>1` where the 1.x sibling refuses loud |
| #449 | `to_file=` artifact rows carry scoped values without the scoped-vs-winner note |

**No regression anywhere.** On `main`, `form="everything"` says `narrow with fields=` and refuses as
an unknown parameter; on this branch it says `project.fields=` and refuses as a wrong-form
parameter. Equally broken, differently worded — and everywhere else the branch is a real
improvement with 1.x identity proven.

### DIFF
--- a/src/housecarl-generator/RecordsGuardProbe.cs
+++ b/src/housecarl-generator/RecordsGuardProbe.cs
@@ -902,20 +902,27 @@ internal static class RecordsGuardProbe
             var tinyJson = RecordsTools.Records(
                 svc, formids: weapons.Select(Fid).ToArray(), format: "json", max_chars: 220,
                 project: new RecordsTools.RecordsProject { form = "fields", fields = new[] { "BasicStats.Damage", "EditorID", "Name" } });
-            var notes = new List<string>();
-            void Harvest(JsonElement e)
+            // Every note-bearing key anywhere in the document — a remedy is wrong wherever it sits, and a
+            // harvester that only reads the top level would miss the per-row ones, which is where they are.
+            static List<string> HarvestNotes(string json)
             {
-                if (e.ValueKind == JsonValueKind.Object)
-                    foreach (var p in e.EnumerateObject())
-                    {
-                        if (p.Name is "note" or "truncation_note" && p.Value.ValueKind == JsonValueKind.String)
-                            notes.Add(p.Value.GetString()!);
-                        Harvest(p.Value);
-                    }
-                else if (e.ValueKind == JsonValueKind.Array)
-                    foreach (var it in e.EnumerateArray()) Harvest(it);
+                var found = new List<string>();
+                void Walk(JsonElement e)
+                {
+                    if (e.ValueKind == JsonValueKind.Object)
+                        foreach (var p in e.EnumerateObject())
+                        {
+                            if (p.Name is "note" or "truncation_note" or "error" && p.Value.ValueKind == JsonValueKind.String)
+                                found.Add(p.Value.GetString()!);
+                            Walk(p.Value);
+                        }
+                    else if (e.ValueKind == JsonValueKind.Array)
+                        foreach (var it in e.EnumerateArray()) Walk(it);
+                }
+                try { Walk(JsonDocument.Parse(json).RootElement); } catch { }
+                return found;
             }
-            try { Harvest(Je(tinyJson)); } catch { }
+            var notes = HarvestNotes(tinyJson);
 
             Check(notes.Count > 0, "a max_chars-starved fields read emits a truncation notice at all");
             // The claim: no notice this TOOL emits may advertise the OTHER generation's selector. Spelled
@@ -941,6 +948,35 @@ internal static class RecordsGuardProbe
             Check(textTruncated, "the text lane truncates the same starved read");
             Check(textTruncated && !textWrongLever,
                   "…and its notice names project.fields= too — one vocabulary per caller, both lanes");
+
+            // MEASURED SURFACE — a remedy predicts what a LATER call produces, so the sentences are read,
+            // not reasoned about. This prints every lever-naming sentence a housecarl_records caller can be
+            // handed across the site families, both lanes, on every run. The assertions above pin the two
+            // that regressed; this block is the evidence for the rest, and the place a NEW wrong lever shows
+            // up before anyone writes an arm for it.
+            var lever = new System.Text.RegularExpressions.Regex(
+                @"\bfields=|\bdepth=|\boffset=|\blimit=|\bmax_chars=|conflict_tree|\bto_file=|\bform=");
+            void Measure(string label, string resp)
+            {
+                var hits = resp.StartsWith("{", StringComparison.Ordinal)
+                    ? HarvestNotes(resp)
+                    : resp.Split('\n').Where(l => l.Contains('[') || l.Contains("error:")).ToList();
+                foreach (var h in hits.Where(h => lever.IsMatch(h)).Distinct())
+                    Console.WriteLine($"        [{label}] {h.Trim()}");
+            }
+            string[] wf = weapons.Select(Fid).ToArray();
+            var containerFields = new RecordsTools.RecordsProject { form = "fields", fields = new[] { "Effects" } };
+            foreach (var (label, resp) in new (string, string)[]
+            {
+                ("fields/json", tinyJson), ("fields/text", tinyText),
+                ("container/text", RecordsTools.Records(svc, formids: new[] { Fid(spellA.FormKey) }, project: containerFields)),
+                ("container/json", RecordsTools.Records(svc, formids: new[] { Fid(spellA.FormKey) }, format: "json", project: containerFields)),
+                ("tree/text",  RecordsTools.Records(svc, formids: wf, max_chars: 300, project: new RecordsTools.RecordsProject { form = "tree" })),
+                ("tree/json",  RecordsTools.Records(svc, formids: wf, max_chars: 300, format: "json", project: new RecordsTools.RecordsProject { form = "tree" })),
+                ("scan/text",  RecordsTools.Records(svc, types: new[] { "WEAP" }, max_chars: 300, project: new RecordsTools.RecordsProject { form = "fields", fields = new[] { "BasicStats.Damage" } })),
+                ("scan/json",  RecordsTools.Records(svc, types: new[] { "WEAP" }, max_chars: 300, format: "json", project: new RecordsTools.RecordsProject { form = "fields", fields = new[] { "BasicStats.Damage" } })),
+            })
+                Measure(label, resp);
 
             Console.WriteLine();
             Console.WriteLine(_fail == 0
--- a/src/housecarl-generator/RecordsGuardProbe.cs
+++ b/src/housecarl-generator/RecordsGuardProbe.cs
@@ -890,6 +890,58 @@ internal static class RecordsGuardProbe
             Check(offIsRefusal, "an off-order scan refusal reaches a json caller as a refusal document");
             Check(offStamped, "…carrying the build it consulted, not epoch:null (settled #4's post-capture bucket)");
 
+            // ============================================================================================
+            //  7 — REMEDY GRAMMAR: a truncation notice names levers THIS tool actually has
+            // ============================================================================================
+            // A remedy predicts what a later call produces. housecarl_records spells its field selector
+            // project.fields=; the 1.x read tools spell it fields=. The json body renderers are shared by
+            // both generations, so a remedy composed inside one can name the other's vocabulary.
+            Console.WriteLine();
+            Console.WriteLine("--- 7: remedy notices name this tool's own levers ---");
+
+            var tinyJson = RecordsTools.Records(
+                svc, formids: weapons.Select(Fid).ToArray(), format: "json", max_chars: 220,
+                project: new RecordsTools.RecordsProject { form = "fields", fields = new[] { "BasicStats.Damage", "EditorID", "Name" } });
+            var notes = new List<string>();
+            void Harvest(JsonElement e)
+            {
+                if (e.ValueKind == JsonValueKind.Object)
+                    foreach (var p in e.EnumerateObject())
+                    {
+                        if (p.Name is "note" or "truncation_note" && p.Value.ValueKind == JsonValueKind.String)
+                            notes.Add(p.Value.GetString()!);
+                        Harvest(p.Value);
+                    }
+                else if (e.ValueKind == JsonValueKind.Array)
+                    foreach (var it in e.EnumerateArray()) Harvest(it);
+            }
+            try { Harvest(Je(tinyJson)); } catch { }
+
+            Check(notes.Count > 0, "a max_chars-starved fields read emits a truncation notice at all");
+            // The claim: no notice this TOOL emits may advertise the OTHER generation's selector. Spelled
+            // out here rather than compared against the render's own constant.
+            var wrongLever = notes.Where(n => System.Text.RegularExpressions.Regex.IsMatch(n, @"(?<!project\.)\bfields=")).ToList();
+            foreach (var n in wrongLever) Console.WriteLine("        offending notice: " + n);
+            Check(notes.Count > 0 && wrongLever.Count == 0,
+                  "…and no notice tells a housecarl_records caller to narrow with 'fields=' (it has project.fields=)");
+
+            // The same question on the text lane — the two lanes compose their notices separately, so a
+            // vocabulary fix on one proves nothing about the other (D2).
+            var tinyText = RecordsTools.Records(
+                svc, formids: weapons.Select(Fid).ToArray(), max_chars: 220,
+                project: new RecordsTools.RecordsProject { form = "fields", fields = new[] { "BasicStats.Damage", "EditorID", "Name" } });
+            bool textTruncated = tinyText.Contains("truncated") || tinyText.Contains("max_chars=");
+            // Report the whole LINE each hit sits on, not a character window around it: the window
+            // prints the middle of a field render and says nothing about what actually matched.
+            var textHits = tinyText.Split('\n')
+                .Where(l => System.Text.RegularExpressions.Regex.IsMatch(l, @"(?<!project\.)\bfields="))
+                .ToList();
+            bool textWrongLever = textHits.Count > 0;
+            foreach (var l in textHits) Console.WriteLine("        offending line: " + l.Trim());
+            Check(textTruncated, "the text lane truncates the same starved read");
+            Check(textTruncated && !textWrongLever,
+                  "…and its notice names project.fields= too — one vocabulary per caller, both lanes");
+
             Console.WriteLine();
             Console.WriteLine(_fail == 0
                 ? "[records-guard] PASS — the 2.0 read surface's core contract holds."
--- a/src/housecarl-generator/RecordsGuardProbe.cs
+++ b/src/housecarl-generator/RecordsGuardProbe.cs
@@ -1084,6 +1084,34 @@ internal static class RecordsGuardProbe
             Check(!slimmed.StartsWith("error:") && slimmed.Contains("form=summary"),
                   "…and dropping project= as the scan notice says yields summary rows, not a refusal");
 
+            // The batch notice's SELECTION clause is a function of the selection LANE, not of the tool: the same
+            // Wire.RenderBatch serves the formids= lane and two scan-derived body lanes, and "request fewer formids"
+            // is inert on a call that named no formids. Both directions are asserted, because a fix that swapped the
+            // clause unconditionally would satisfy one arm and break the other. The wrong-lever patterns above cannot
+            // see this one — 'formids' is a lever housecarl_records genuinely HAS, so only the lane makes it wrong.
+            static string? BatchCut(string resp) =>
+                resp.Split('\n').FirstOrDefault(l => l.Contains("... [truncated: rendered") && l.Contains(" records before hitting"));
+            var scanBatch = RecordsTools.Records(svc, types: new[] { "WEAP" }, max_chars: 400,
+                                                 project: new RecordsTools.RecordsProject { form = "everything" });
+            var scanCut = BatchCut(scanBatch);
+            Check(scanCut is not null && scanCut.Contains("lower limit=") && !scanCut.Contains("formids"),
+                  "…a SCAN-selected batch's truncation notice names limit=, not a formids list the caller never wrote");
+            var fidBatch = RecordsTools.Records(svc, formids: wf, max_chars: 400,
+                                                project: new RecordsTools.RecordsProject { form = "everything" });
+            var fidCut = BatchCut(fidBatch);
+            Check(fidCut is not null && fidCut.Contains("request fewer formids") && !fidCut.Contains("lower limit="),
+                  "…and the formids= lane still says 'request fewer formids', which is true only there");
+            // The OFF-ORDER scan is the second scan-derived body lane and the in-order arm above cannot reach it:
+            // with only that one, reverting this lane's call site left the guard green. The off-order file holds a
+            // single weapon override by fixture design (several checks above depend on that count), so this batch
+            // has one row and the notice is driven by starving the cap rather than by a second row — the sentence
+            // it composes is the same one, which is what the arm reads.
+            var offBatch = RecordsTools.Records(svc, types: new[] { "WEAP" }, source: Je($"\"{oldName}\""), max_chars: 12,
+                                                project: new RecordsTools.RecordsProject { form = "everything" });
+            var offCut = BatchCut(offBatch);
+            Check(offCut is not null && offCut.Contains("lower limit=") && !offCut.Contains("formids"),
+                  "…and the OFF-ORDER scan's batch notice names limit= too (its own lane, its own arm)");
+
             Console.WriteLine();
             Console.WriteLine(_fail == 0
                 ? "[records-guard] PASS — the 2.0 read surface's core contract holds."
--- a/src/housecarl-generator/RecordsGuardProbe.cs
+++ b/src/housecarl-generator/RecordsGuardProbe.cs
@@ -1048,6 +1048,10 @@ internal static class RecordsGuardProbe
                 // LeverNames.Records left the guard green while six sentences reverted to the pre-fix form
                 // (round 2). The lookahead is the whole point — bare 'project.fields' in a remedy is the defect.
                 (@"\bproject\.fields(?!=)", "narrow with 'project.fields' without the '=' (the pre-fix spelling)"),
+                // Actionability, not spelling. "drop project.fields=" names a lever records genuinely HAS, so no
+                // name-based pattern above can catch it — and the 'fields' form refuses without its paths, so the
+                // call it predicts is rejected. Having the lever is not the test; the next call working is.
+                (@"drop project\.fields",  "drop 'project.fields' (the 'fields' form refuses without its paths)"),
             };
             // Per-PROBE, before the per-lane assertions: a lane aggregates several site families, so a family that
             // stops emitting (fixture rot, a container that no longer collapses) leaves the lane count healthy and
@@ -1072,6 +1076,13 @@ internal static class RecordsGuardProbe
                           $"…no {lane} sentence tells a housecarl_records caller to {claim}");
                 }
             }
+
+            // The scan remedy's predicted call, actually MADE. Every assertion above reads a sentence; this one
+            // follows it. The notice says to drop project= for slimmer rows, so drop it and require what the
+            // sentence promised: a summary render, not a refusal.
+            var slimmed = RecordsTools.Records(svc, types: new[] { "WEAP" }, max_chars: 300);
+            Check(!slimmed.StartsWith("error:") && slimmed.Contains("form=summary"),
+                  "…and dropping project= as the scan notice says yields summary rows, not a refusal");
 
             Console.WriteLine();
             Console.WriteLine(_fail == 0
--- a/src/housecarl-generator/RecordsGuardProbe.cs
+++ b/src/housecarl-generator/RecordsGuardProbe.cs
@@ -923,60 +923,72 @@ internal static class RecordsGuardProbe
                 return found;
             }
             var notes = HarvestNotes(tinyJson);
-
             Check(notes.Count > 0, "a max_chars-starved fields read emits a truncation notice at all");
-            // The claim: no notice this TOOL emits may advertise the OTHER generation's selector. Spelled
-            // out here rather than compared against the render's own constant.
-            var wrongLever = notes.Where(n => System.Text.RegularExpressions.Regex.IsMatch(n, @"(?<!project\.)\bfields=")).ToList();
-            foreach (var n in wrongLever) Console.WriteLine("        offending notice: " + n);
-            Check(notes.Count > 0 && wrongLever.Count == 0,
-                  "…and no notice tells a housecarl_records caller to narrow with 'fields=' (it has project.fields=)");
 
-            // The same question on the text lane — the two lanes compose their notices separately, so a
-            // vocabulary fix on one proves nothing about the other (D2).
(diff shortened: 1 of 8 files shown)


src/housecarl-mcp/ReadTools.cs
```diff
@@ -986,12 +991,12 @@ public static string RenderBatch(LoadOrderService svc, IReadOnlyList<ReadOutcome
                 truncated = true;
                 sb.Append("... [truncated: rendered ").Append(rendered).Append(" of ").Append(outcomes.Count)
                   .Append(" records before hitting max_chars=").Append(cap)
-                  .Append("; request fewer formids, pass fields= to slim each, or raise max_chars]\n");
+                  .Append("; request fewer formids, pass ").Append(lv.Fields).Append(" to slim each, or raise max_chars]\n");
```
Avick3110: **`request fewer formids` is inert on the records lanes that reach this line via a scan.**

Half this sentence was fixed and half was not. `lv.Fields` now carries the caller's spelling, but `request fewer formids` still assumes the call was selected by `formids=` — and two of the three records call sites of `Wire.RenderBatch` are scan-derived body lanes:

- `RecordsTools.cs:1225` — `form="everything"` on a scan, and `form="fields"` with `plugins=`+`source=`
- `RecordsTools.cs:1431` — the off-order scan's body lane

Concretely: `housecarl_records(types=["SPEL"], project={form:"everything"}, max_chars:400)` cuts with

```
... [truncated: rendered 3 of 41 records before hitting max_chars=400; request fewer formids, pass project.fields= to slim each, or raise max_chars]
```

There are no `formids=` to request fewer of. The lever that actually windows this response is `limit=`, which the sentence never names. That is round 2's own test — *"having the lever is not the test; the next call working is"* — failing on a line this PR rewrote. `RecordsTools.cs:589` (the real `formids=` lane) is the only site where the clause is true.

(The `pass project.fields=` half being refused by `form="everything"` is the already-filed #445; this is the other half, which I don't see covered by #442–#449.)

src/housecarl-generator/RecordsGuardProbe.cs
```diff
@@ -890,6 +890,200 @@ static HashSet<FormKey> Run(string[] where, IEnumerable<IMajorRecordGetter> bodi
...
+                ("spill/artifact", "artifact", artResp),
+            };
+
+            // Harvest with NO lever filter. A filter is where a wrong lever hides: the pre-fix tree notice
+            // said "narrow with project.fields" with no '=', and a regex keyed on '=' did not see it at all.
+            // json → every note/error string in the document; text → every line carrying remedy vocabulary.
+            // 'pass ' is in this alternation because leaving it out was a blind spot of exactly the kind the
+            // no-lever-filter rule exists to stop: the text P5 note reads "… — pass winner_fields=true for
+            // load-order truth" and matched none of the other words, so the text lane harvested it as nothing.
+            var remedyLine = new System.Text.RegularExpressions.Regex(
```
Avick3110, line 1039: **`remedyLine` is case-sensitive, so the text lane's harvest silently drops a capitalized remedy.**

The comment above this regex defends it as the one filter that survives — "on the LITERAL (a multi-word sentence vs a bare key), never on which lever a sentence names". But without `RegexOptions.IgnoreCase` it is also a filter on *capitalization*, and that hides a real site:

`RecordsTools.cs:1585` emits

```
  no differing fields in what was read, but the deep read was TRUNCATED at the cap — NOT a clean 'identical' (Q3). Narrow with project.fields= to compare in full.
```

That line has no lowercase `narrow`, no `max_chars`, and none of `raise `/`lower `/`drop `/`pass `. So `resp.Split('\n').Where(l => remedyLine.IsMatch(l))` discards it before `sentences` is built, and every per-lane `wrongLevers` Check runs over a set that cannot contain it. Revert `LeverNames.Records.Fields` → `"fields="` at that one site and the guard stays green.

This stacks with the declared `clause/delta-identical` GREEN cell rather than duplicating it: that cell is green because the fixture cannot reach `!d.Complete` past `ConflictDiffDepth = 16`. Even once #447 supplies that fixture, the assertion still would not see the sentence — the harvest drops it first. And any future records remedy opening with `Narrow`/`Pass`/`Drop`/`Raise`/`Lower` inherits the same hole.

Passing `System.Text.RegularExpressions.RegexOptions.IgnoreCase` to this constructor closes it.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -1062,7 +1065,7 @@ public static string RenderCrossQuery(LoadOrderService svc, CrossQueryOutcome q,
                 truncated = true;
                 sb.Append("... [truncated: rendered ").Append(rendered).Append(" of ").Append(q.Keys.Count)
                   .Append(" returned matches before hitting max_chars=").Append(cap)
-                  .Append("; lower limit=, drop fields=/conflict_tree, or raise max_chars]\n");
+                  .Append("; lower limit=, drop ").Append(lv.SlimScan).Append(", or raise max_chars]\n");
```
Avick3110: **`drop project= (summary rows)` is inert on the form this tool defaults to.**

`SlimScan` is right for `form="fields"` — that was round 2's fold. But this same sentence renders on every records scan cut, and two of its lanes never passed `project=` at all:

- `RecordsTools.cs:1257` with `form="summary"` — the **default** form, `projFields` is null
- `RecordsTools.cs:1464` — the off-order scan, which passes `fields: null` unconditionally

`housecarl_records(types=["WEAP"], max_chars:300)` cuts with

```
... [truncated: rendered 4 of 812 returned matches before hitting max_chars=300; lower limit=, drop project= (summary rows), or raise max_chars]
```

Dropping `project=` from that call yields a byte-identical call — there is nothing to drop, and the rows the caller is already looking at *are* the summary rows the remedy points at. The parenthetical makes it worse than the neutral case: it promises a render the caller already has.

This is strictly better than the pre-fix `drop fields=/conflict_tree` (which named parameters records does not have), so it is not a regression. But it is the same inert-remedy class round 2 declined to leave standing on `drop project.fields=`, and the CHANGELOG's "what it does not cover" list does not name it.

src/housecarl-mcp/Artifacts.cs
```diff
@@ -157,7 +158,8 @@ public static (SpillInfo? Spill, string? Error) WriteCrossQuery(
                 var fk = q.Keys[i];
                 string? matches = q.MatchedTargets is { } mt && i < mt.Count ? mt[i] : null;
                 var o = svc.ResolveReadOn(q, fk, winnerFields ? null : (q.Sources is { } src ? src[i] : null), fields, false, depth,
-                                          resolveNames: resolveNames, linkMemo: linkMemo);
+                                          resolveNames: resolveNames, linkMemo: linkMemo,
+                                          containerHint: (levers ?? LeverNames.Legacy).ContainerHint);   // an artifact row is read by the same caller (#439)
```
Avick3110, line 165: **`levers` reaches the read here but not the row writer eleven lines down.**

`WriteCrossQuery` now threads the carrier into `ResolveReadOn`'s `containerHint`, but line 173 still writes each row with

```csharp
writer.WriteRow((w, ms) => JsonWire.WriteReadRecord(w, o, ms, int.MaxValue, matches), o.Record!.Type);
```

— no `levers`, so `WriteFieldsArray` composes its `[truncated at max_chars: … narrow with fields=, lower depth=]` note in the Legacy vocabulary for a records artifact.

Not live today: `ms.Length >= cap` cannot trip at `int.MaxValue`, so no wrong sentence reaches a file. Flagging it because the seam is inconsistent *by construction* — the same artifact row already carries `— pass project.depth=2 to expand` from this line while its truncation note would say `depth=` — and `Legacy`-as-default means any future bound on the artifact writer's cap (a per-row ceiling, a memory guard) re-opens #439 in the `to_file=` lane with nothing failing. That is exactly the silent-seam class escalated in round 1 and filed as #444.

One extra argument on line 173 closes it now rather than leaving it for whoever bounds that cap.

src/housecarl-mcp/ReadTools.cs
Avick3110: Folded in `d726117`.

The selection clause is carried now, not literal: `LeverNames.BatchSelection`, with `OnScanSelection()` deriving the scan-lane vocabulary from the tool's. Both scan-derived call sites pass it — the in-order body lane and the off-order file scan — and the renderer infers nothing from state. `formids=` keeps "request fewer formids", which is true only there.

Your repro cuts with this now:

```
... [truncated: rendered 1 of 4 records before hitting max_chars=400; lower limit=, pass project.fields= to slim each, or raise max_chars]
```

Three arms, RED-checked from a committed state (`gate1/*`, four cells): the scan lane must name `limit=` and not formids; the formids lane must name formids and not `limit=`, so a fix that swapped the clause unconditionally fails one of them; and the off-order lane has its own arm, because with only the in-order one, reverting the off-order call site left the guard **green**. None of the wrong-lever patterns could ever see this class — `formids` is a lever `housecarl_records` genuinely has, and only the lane makes naming it wrong.

On scope: this is inside the shipped claim, not outside it. The claim covers the `fields` form, and the in-order body lane serves `form="fields"` under `plugins=`+`source=`. So the claim as written was false for that case, and this fold is what makes it true rather than an extension of it. The `everything` form's `pass project.fields=` half stays #445.

Text lane only — the json batch render signals a cut structurally and composes no remedy prose. The CHANGELOG's covers list now scopes the cut's own clauses to `"text"` instead of letting "on text and json" cover a sentence json does not write.

src/housecarl-generator/RecordsGuardProbe.cs
Avick3110: Folded in `5bc3a96` — `RegexOptions.IgnoreCase`, plus the fixture that makes it load-bearing.

You are right about the axis. The comment defended the filter as being on the literal rather than on the lever, and it was, but it was also a filter on capitalisation — which is a property of where a remedy sits in its sentence, not of whether it is one.

The permanent fixture drives `RecordsTools.cs:1585` itself rather than a stand-in. That branch needs zero deltas AND `!d.Complete`, and the incompleteness was declared unreachable — but the route is `ReadEngine`'s expansion budget, not `ConflictDiffDepth`: a form list one element past `MaxExpandNodes` truncates, and an override that copies it without changing a field supplies the zero-delta half. `deltaIdentical/text` is in the probe list now, and the harvest picks up:

```
  no differing fields in what was read, but the deep read was TRUNCATED at the cap — NOT a clean 'identical' (Q3). Narrow with project.fields= to compare in full.
```

That closes the cell round 2 had to declare GREEN. `clause/delta-identical` is a live RED cell under its own name in the sweep, so that sentence is inside the claim rather than listed on #447, and #447 loses that item.

Two cells, failing for different reasons: dropping `IgnoreCase` fails the probe's own bite check — that lane harvests nothing from it — and reverting the site's lever to `fields=` fails the wrong-lever pattern, which is the check you proved stayed green.

src/housecarl-mcp/ReadTools.cs
Avick3110: Folded in `3e0026e`.

`SlimScan` is nullable now, and null means the clause is omitted rather than rendered over nothing — the notice keeps the two levers that are real on such a call. The in-order cross-query lane asks its own parameters (field paths present ⇒ detail rows ⇒ `project=` is genuinely droppable); the off-order lane passes `WithNothingToDrop()` unconditionally, because it never passes paths.

Your repro cuts with this now:

```
... [truncated: rendered 2 of 4 returned matches before hitting max_chars=300; lower limit= or raise max_chars]
```

Four cells (`gate3/*`): one per branch of the conditional, one per deciding call site. Both directions matter — simply deleting the clause satisfies the summary arm and loses a remedy that is actionable on the `fields` form, so that arm is asserted too, and the off-order lane has its own.

CHANGELOG: the conditionality is stated where the entry describes the slim-down clause, and the covers list names the `summary` form's scan cut. It also scopes the cut's own clauses to `"text"` — `"json"` and `"dense"` report a cut as data, so claiming them "on text and json" would have claimed a sentence json does not write.

src/housecarl-mcp/Artifacts.cs
Avick3110: Folded in `3117710` — and one method wider than you asked, deliberately.

`WriteCrossQuery`'s row writer takes `levers` now. But `WriteBatch` eleven lines down is the same defect: it had no `levers` parameter at all, writes its rows through the same `WriteReadRecord`, and is reached by four `housecarl_records` call sites (`to_file=` and the ceiling auto-spill, on the formids lane and both scan body lanes). Your stated reason for folding — the seam is inconsistent by construction, and `Legacy`-as-default means whoever later bounds that cap reopens #439 in the `to_file=` lane with nothing failing — applies to it word for word, so closing one and leaving the other would have left exactly that hazard live. Flagging it here rather than widening quietly; say the word if you want it split back out.

Dormancy is why the arm needed a seam of its own. Both writers take a `rowCap` that production never passes, documented as existing so the note can be **driven** — a note nothing can reach is a note nothing can prove. Two arms, one per writer, each asserting the records spelling and the absence of the bare one, both RED-checked (`gate4/*`).

Avick3110:
## Gate folds landed — verification results

All four findings folded, one commit each, each RED-checked from a committed state before the next
was started. No new rounds (§11: fresh eyes are for before the PR opens). Replies are on the four
threads; the PR body carries the updated claim, clause table and a gate-review section.

| Commit | Finding |
|---|---|
| `d726117` | 1 — the batch cut's selection clause is lane-aware |
| `5bc3a96` | 2 — `RegexOptions.IgnoreCase`, and the fixture that drives the site it was hiding |
| `3e0026e` | 3 — the slim-down clause renders only where there is something to drop |
| `3117710` | 4 — both artifact row writers compose in the caller's vocabulary |
| `45bdce9` | CHANGELOG: the two clauses' conditionality, and the cut's clauses scoped to `"text"` |

### Sabotage sweep — 34 cells, 34 RED, canary RED, tree clean, nothing declared GREEN

One sweep with one canary: the pre-PR wave's 21 cells are carried into it (re-pointed where a gate
fold moved the line they name), so a second canary is not proving a second list. Ten cells are new.

Both directions of both new conditionals are cells, per §11 — a fix that swapped or deleted a clause
unconditionally fails the opposite arm:

```
  gate1/scan-lane          RED   the scan lane reverts to the formids clause
  gate1/formids-lane       RED   the formids lane gets the scan clause
  gate1/entry-everything   RED   the in-order body lane stops declaring its lane
  gate1/entry-offorder     RED   the off-order body lane stops declaring its lane
  gate2/ignorecase         RED   the harvest goes back to filtering on capitalisation
  clause/delta-identical   RED   was the wave's ONE declared GREEN; live now
  gate3/omit-branch        RED   the nothing-to-drop call gets the clause back
  gate3/keep-branch        RED   the clause is deleted everywhere
  gate3/inorder-callsite   RED   the in-order lane stops asking its own parameters
  gate3/offorder-callsite  RED   the off-order lane stops declaring it passes no paths
  gate4/crossquery-row     RED   the cross-query artifact row writer loses the vocabulary
  gate4/batch-row          RED   the batch artifact row writer, same seam one method down
```

Two of these were **green before the arm was written**, which is why they are worth the tokens:
`gate1/entry-offorder` (the in-order arm alone could not reach the off-order scan) and
`clause/delta-identical` (no fixture reached `!d.Complete`). Both are folds of the arm, not just of
the code.

### `ci-all` — 132/132 guards, 3012/3012 checks

Up from 3003 by the nine new arms. PR A's `refusal-completeness-guard` is green and untouched.

### 1.x identity — 14/14, unchanged

Re-run against the blobs at `fbfe015` after the folds. This is the check that had to hold: the
lane-aware clauses are new conditionals inside sentences 1.x callers also render, and every
`Legacy` composition is still byte-identical. The two sites that still hold a bare literal are the
same two as before — a 1.x tool's own `Description` prose and `Wire.RenderPluginFile` — both
unreachable from records.

### One thing folded wider than written

Finding 4 named `Artifacts.cs:173`. `Artifacts.WriteBatch` eleven lines down had no `levers`
parameter at all, writes rows through the same `WriteReadRecord`, and is reached by four
`housecarl_records` call sites. The stated reason for folding the first applies to it word for word,
so both are folded and it is called out on the thread rather than widened quietly — say the word if
you want it split back out.

Both writers also gained a `rowCap` that production never passes, so the dormant note could be
driven and read rather than reasoned about.

**Stopping here — your explicit go gates the merge.**

Avick3110:
Merged with `--rebase`, so the SHAs cited in the four thread replies are the pre-rebase ones. They still resolve here (the PR keeps them), but they are not on `main`. Mapping, for anyone reading this from `main`:

| reply cites | on `main` |
|---|---|
| `d726117` | `d718298` — the batch cut's selection clause |
| `5bc3a96` | `699bb00` — IgnoreCase + the delta-identical fixture |
| `3e0026e` | `ede900b` — the slim-down clause's conditional |
| `3117710` | `b194573` — the artifact row writers |
| `45bdce9` | `0201d8d` — CHANGELOG |

`main` is at `0201d8d`, linear. #439 closed.

Issue #463, opened 2026-09-01
[NOTE] Console modes and a CLI parser still live in the shipping write engine — the live-code residue of #453


Avick3110:
**Provenance:** split out of #453 at Aaron's direction (2026-09-01) — #453's own class was the
spike-era proof/discovery modes plus the hardcoded source path, all swept in its PR. This is the
residue, which is a different class: code that is alive, not dead.

`src/housecarl-core/WriteEngine.cs` still carries console entry points inside the shipping library:

- **`patch` / `show` / `condition-patch`** — hand-run dev harnesses that resolve a record and write
  a real reviewable `.esp`. Not dead code: `insert-at-index-guard` drives `WriteEngine.RunPatch`
  directly, and `condition-patch` is the xEdit lock for the FLOI capability.
- **`coerce-audit` / `coerce-selftest`** — live `ci-all` probes. Every other `ci-all` probe lives in
  housecarl-generator; these two do not.
- **`WriteEngine.ParseFlags`** — a `--flag value` CLI parser in the engine library, with nine call
  sites across six generator probes. The reason `patch`/`show`/`condition-patch` cannot simply move
  without a decision: relocating it is its own change, and leaving it behind means the library keeps
  a CLI parser even after the harnesses go.

Roughly 1,000 lines. Harmless at runtime — nothing on the tool surface reaches any of it — and the
engine's class header names all of it explicitly today, so it is not hidden. But it is a dev/CLI
surface compiled into the shipped core, which is the same smell #453 named one class of.

Not urgent and not a bug. The options, smallest first: move the two coercion probes to
housecarl-generator beside every other `ci-all` probe (mechanical, compiler-verified); then decide
`ParseFlags`'s home; then move the three harnesses, which that decision unblocks. Worth doing when
something else is already open in `WriteEngine.cs` rather than as a standalone sweep.

A sibling worth noting while in there: `WriteOracle` (housecarl-generator) carries the same
hardcoded `C:\Program Files (x86)\Steam\…\Skyrim.esm` default that #453 removed from the core. It is
already on the dev side, so it was out of #453's scope, but it fails the same way — naming a path
the caller never typed on any machine that does not have that exact install.

Avick3110:
**Generator-side default-path residue — the full set, from PR #464's gate review.**

[...]

PR #464, opened 2026-09-01
Move the spike-era proof and discovery modes out of the write engine

### LINKED ISSUE #453: [NOTE] Spike-era proof scaffolding still lives in the shipping core (WriteEngine harness, hardcoded Steam path)
**Provenance:** filed from the 2026-08-31 cloud fresh-eyes session, on Aaron's direction to bring the session's quality findings into the tracked lane.

`src/housecarl-core/WriteEngine.cs` still opens with the spike-era acceptance harness: `RunNpcSkillsProof` and the `write-api` discovery mode, including a hardcoded default source path

```
C:\Program Files (x86)\Steam\steamapps\common\Skyrim Special Edition\Data\Skyrim.esm
```

in the production library (`DefaultSourcePath`, `WriteEngine.cs:31`). The engine below it is the shipped generic write path; the proof harness was its birth certificate.

Harmless at runtime (nothing on the tool surface reaches it), but it is dev scaffolding inside the shipping core: a console mode, a machine-specific path, and step-4-plan commentary that describes the build order rather than the invariants. Options, smallest first: move the proof modes into housecarl-generator beside the probes they resemble; or delete them if the atomic-commit and write guards now cover what the proof proved. Worth a sweep for siblings of the same class while in there.

Avick3110:
Fixes #453

Moves the spike-era proof and discovery modes out of `housecarl-core`'s `WriteEngine.cs` (the shipping library) into `WriteDiagnosticsProbe.cs` in `housecarl-generator`, and removes the hardcoded Steam path from core.

## What moved

`npc-skills` (the step-4 acceptance proof), `write-api`, `poly-probe`, `substruct-probe` — all hand-run, all reachable only from `Program.cs` dispatch, none on the tool surface — plus their exclusive helpers. Cross-file calls are now `WriteEngine.`-qualified; no access level widened (the callees are `internal`, reached through the pre-existing `InternalsVisibleTo("housecarl-generator")`). The move in fact *narrows* the shipped surface: four `public` dev-only methods leave the library that ships in the plugin.

`DefaultSourcePath` is gone from `housecarl-core`. Its last user, `condition-patch`, now **requires `--source`**, matching its siblings `patch` and `show`. The old default failed with `source plugin not found: C:\Program Files (x86)\…` — naming a path the caller never typed.

`WriteEngine.cs`: 4,362 → 4,066 lines.

## What did NOT move, deliberately

`patch` / `show` / `condition-patch` (live dev harnesses; `patch` is driven by `insert-at-index-guard` in CI) and `coerce-audit` / `coerce-selftest` (live `ci-all` probes). Console modes in a library is the same smell, but these are live tools, not spike residue. `WriteEngine.ParseFlags` has nine call sites across six generator probes, so relocating it is its own change.

That residue is filed as **#463** and is not this PR's concern.

## Verification

- `ci-all` **132/132**.
- Relocation verified mechanically three times independently (mine + two reviewers): a line-set comparison of every removed `WriteEngine.cs` line against the new file, normalising the `WriteEngine.` qualifier, leaves exactly the intended deltas. Local `Sha`/`SafeTypes`/`Pretty` copies are semantically identical to the originals (per the generator's existing convention — `ConditionProbe`, `CoordCellProbe` and others each carry their own).
- Outputs byte-identical base vs branch for `write-api`, `substruct-probe`, `npc-skills`, `poly-probe <plugin>`, `condition-patch --source <plugin>`. `npc-skills` still prints `=== PASS: nested dict-in-substruct Set landed; original untouched ===`.
- `condition-patch` both directions: no `--source` → `error: missing required --source`; with one → writes the patch, reads the re-targeted condition back, source unchanged.
- Verified no access level widened, no overload re-binding, no static-init/lifetime change, no assembly-location dependence, and nothing anywhere in the repo still points at the moved code by its old home.

No sabotage sweep and no new guard: this branch adds no CI machinery, and the two refusals it does add are hand-verified in both directions. Writing a CI guard for a hand-run dev harness would be disproportionate — stated here rather than left as an unexplained gap.

## Review rounds (CLAUDE.md §5 #11)

Conducted by a fresh session, not the branch's author (rule 9 seam). Two rounds, four independent reviewers, each in its own worktree.

**Round 1** — two reviewers (move fidelity; claims-vs-code).
- Fidelity reviewer: **no highs, no mediums.** Verified the relocation clean on every axis above.
- Claims reviewer: one **high** — the rewritten header claimed "a request is validated against the corpus rulebook before anything is mutated" for every entry point. `ApplyVerb` is public and does no rulebook check; the engine's own comment says pre-flight only "normally" gates it, and this branch's own moved proof calls `rulebook.Validate` itself before `ApplyVerb`. **Folded** per DOC_HYGIENE §8 (a guarantee without a probe is a claim, not a guarantee).
- Also folded: `poly-probe` crashed with an unhandled Mutagen exception on the default path while a new comment said it couldn't — given the named refusal its sibling has, verified both directions; the review-round citation in source (§7 — the `#453` pointer stays, that's §7's stated exception); `oracle` presented as a standing proof when it is hand-run; `substruct-probe` mislabelled a build-start diagnostic.
- **Refused (2):** the "still here" inventory as a §6 duplicate — settled decision 4 puts it there deliberately and both reviewers verified it exact, with no unnamed sixth entry point. And, initially, the stale `WriteCensus`/`condition-patch` arm-navigation comments, on the grounds the branch didn't touch them.

**Round 2** — two fresh reviewers (behaviour; claims-vs-code).
- Behaviour reviewer: **no highs, no mediums.** Three lows, each verified as pre-existing and identical on base — a non-plugin file still crashing one case past the new guard, an empty-string argument echoing an empty path, and a value-less `--source` binding to the literal `"true"` (shared with `RunPatch`/`RunShow`).
- Claims reviewer: two mediums. My round-1 fold's "`ApplyVerb` validates nothing" was over-broad and contradicted the line above it — `ApplyVerb` does refuse structurally by name; what it omits is the *rulebook* check. **Folded.** And it corrected my round-1 refusal: the branch *does* touch the `condition-patch` comment block, so §8's "cleaned when a file is touched" applies. **Folded**, minimally — the claim measured false is deleted rather than replaced with a capability claim this branch has no probe for.
- The two reviewers disagreed on the `ApplyVerb` sentence; both were half-right, which is what produced the precision fix rather than a revert.

**Refusals stated with grounds, so the rate is visible:** three findings refused across the two rounds — the §6 inventory duplicate (settled by decision 4, verified accurate), and the three pre-existing lows above, which are defects in code this branch does not change and are shared with sibling harnesses. Folding them would widen a relocation branch.

Stopped at two rounds: round 2's behaviour pass was clean at medium and above, and what remained was one clause of the reviewing session's own prose plus pre-existing comments — convergence rather than a recurring class.

### DIFF
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -16,15 +16,15 @@ namespace HousecarlCore;
 /// list element, polymorphic arm). It is blind to which record it edits, which is what makes coverage a property
 /// of Mutagen's model rather than of a hand-written per-type map.
 ///
-/// <para>Two contracts every entry point here keeps: a request is validated against the corpus rulebook before
-/// anything is mutated, and a path the engine cannot navigate or a value it cannot coerce is refused by name
-/// rather than skipped. The default destination is a new mod; the in-place lane is the caller's explicit,
-/// consented request, gated above this layer. <c>apply-guard</c> drives all three from the tool surface.</para>
+/// <para>What it keeps: a path it cannot navigate or a value it cannot coerce is refused by name, never skipped.
+/// <b>Corpus-rulebook validation is the CALLER's, not this layer's</b> — <see cref="ApplyVerb"/> is public and
+/// validates nothing, so a direct or CLI caller that skips pre-flight mutates unvalidated. <c>apply-guard</c>
+/// proves the pre-flight, refusal and lane contracts for the <c>housecarl_apply</c> path only.</para>
 ///
-/// <para>What the engine writes is proven per KIND by <c>oracle</c> — each cell run twice, through here and
-/// through a hand-written typed setter, and required byte-identical.</para>
+/// <para>Per-KIND equivalence against a hand-written typed setter is proven by <c>oracle</c> — hand-run, not in
+/// <c>ci-all</c>.</para>
 ///
-/// <para>The acceptance proof and the build-start diagnostics moved to <c>WriteDiagnosticsProbe</c>
+/// <para>The acceptance proof and the hand-run diagnostics moved to <c>WriteDiagnosticsProbe</c>
 /// (housecarl-generator) with #453. Still here, and still console entry points in a library: the <c>patch</c> /
 /// <c>show</c> / <c>condition-patch</c> dev harnesses, and the <c>coerce-audit</c> / <c>coerce-selftest</c>
 /// CI probes.</para>
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -17,12 +17,13 @@ namespace HousecarlCore;
 /// of Mutagen's model rather than of a hand-written per-type map.
 ///
 /// <para>What it keeps: a path it cannot navigate or a value it cannot coerce is refused by name, never skipped.
-/// <b>Corpus-rulebook validation is the CALLER's, not this layer's</b> — <see cref="ApplyVerb"/> is public and
-/// validates nothing, so a direct or CLI caller that skips pre-flight mutates unvalidated. <c>apply-guard</c>
-/// proves the pre-flight, refusal and lane contracts for the <c>housecarl_apply</c> path only.</para>
+/// <b>Corpus-rulebook validation is the CALLER's, not this layer's</b> — <see cref="ApplyVerb"/> is public and does
+/// no rulebook check (it makes its own structural refusals, but nothing about identity fields, verb-vs-cardinality,
+/// enum-key legality or range), so a direct or CLI caller that skips pre-flight mutates unvalidated.
+/// <c>apply-guard</c> proves the pre-flight, refusal and lane contracts for the <c>housecarl_apply</c> path only.</para>
 ///
-/// <para>Per-KIND equivalence against a hand-written typed setter is proven by <c>oracle</c> — hand-run, not in
-/// <c>ci-all</c>.</para>
+/// <para>Byte-equivalence against a hand-written typed setter is proven by <c>oracle</c>, over the cells it
+/// enumerates — hand-run, not in <c>ci-all</c>.</para>
 ///
 /// <para>The acceptance proof and the hand-run diagnostics moved to <c>WriteDiagnosticsProbe</c>
 /// (housecarl-generator) with #453. Still here, and still console entry points in a library: the <c>patch</c> /
@@ -337,8 +338,7 @@ public static class WriteEngine
     //  RE-TARGET it to a different real form THROUGH THE ENGINE (pre-flight rooted at the arm + ApplyVerb -> SetFloi),
     //  and emit ONE reviewable single-master .esp. Aaron opens it in xEdit and confirms the condition now points at
     //  the new target; the original stays byte-for-byte untouched (Q3). The write is ARM-ROOTED — the same engine
-    //  entry BuildStruct's nested Sets use; reaching arm sub-fields from a record-root path is the broader arm-breadth
-    //  nav surface (reconciles at the final completion sweep), deliberately not this wave.
+    //  entry BuildStruct's nested Sets use.
     //    dotnet run --project src/housecarl-generator condition-patch \
     //        --source "<plugin>" [--target XXXXXX:Plugin.esp] [--out <path>] [--name <patch>]
     // ======================================================================
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -17,9 +17,14 @@ namespace HousecarlCore;
 /// of Mutagen's model rather than of a hand-written per-type map.
 ///
 /// <para>What it keeps: a path it cannot navigate or a value it cannot coerce is refused by name, never skipped.
-/// <b>Corpus-rulebook validation is the CALLER's, not this layer's</b> — <see cref="ApplyVerb"/> is public and does
-/// no rulebook check (it makes its own structural refusals, but nothing about identity fields, verb-vs-cardinality,
-/// enum-key legality or range), so a direct or CLI caller that skips pre-flight mutates unvalidated.
+/// <b>Corpus-rulebook validation is the CALLER's, not this layer's</b> — <see cref="ApplyVerb"/> is public and runs
+/// no rulebook check. It does refuse most of what pre-flight would, just with a CLR/structural message rather than
+/// a corpus-derived one: a verb the collection does not take (<c>ApplyDictVerb</c>'s <c>default:</c> arm), a dict
+/// key that is not a legal enum member (<c>Coerce</c> → <c>Enum.Parse</c>), a value out of its type's range
+/// (<c>byte.Parse</c>) all throw here. The genuinely rulebook-only residue is narrower: record IDENTITY —
+/// <c>FormKey</c> is settable on every concrete Mutagen record, so nothing in this layer stops a caller rewriting
+/// it and only the rulebook's flat identity reject does — and legality the corpus declares that no CLR type
+/// expresses. That residue is what a direct or CLI caller skipping pre-flight mutates unvalidated.
 /// <c>apply-guard</c> proves the pre-flight, refusal and lane contracts for the <c>housecarl_apply</c> path only.</para>
 ///
 /// <para>Byte-equivalence against a hand-written typed setter is proven by <c>oracle</c>, over the cells it
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -11,143 +11,26 @@ using Noggog;
 namespace HousecarlCore;
 
 /// <summary>
-/// Step 4 — the reflection-driven write engine.
+/// The reflection-driven write engine: overlay → <c>GetOrAddAsOverride</c> → reflect the property → coerce →
+/// set → write with masters, for <b>any</b> record group and along <b>nested</b> paths (substruct, dict-by-key,
+/// list element, polymorphic arm). It is blind to which record it edits, which is what makes coverage a property
+/// of Mutagen's model rather than of a hand-written per-type map.
 ///
-/// The spike (<c>dev/references/spike/ReflectionWrite.cs</c>) proved the core mechanism
-/// (overlay → GetOrAddAsOverride → reflect property → coerce → SetValue → write-with-masters,
-/// 4/4 byte-identical) but only via <i>typed</i> accessors on <c>Armor</c>. This engine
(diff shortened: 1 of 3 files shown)


src/housecarl-core/WriteEngine.cs
```diff
@@ -11,143 +11,27 @@
...
+/// of Mutagen's model rather than of a hand-written per-type map.
 ///
-/// The spike (<c>dev/references/spike/ReflectionWrite.cs</c>) proved the core mechanism
-/// (overlay → GetOrAddAsOverride → reflect property → coerce → SetValue → write-with-masters,
-/// 4/4 byte-identical) but only via <i>typed</i> accessors on <c>Armor</c>. This engine
-/// generalises it to <b>any</b> record group and to <b>nested</b> paths (substruct → dict-by-key),
-/// which is the net-new, highest-risk delta (step-4 plan §2 #1).
+/// <para>What it keeps: a path it cannot navigate or a value it cannot coerce is refused by name, never skipped.
+/// <b>Corpus-rulebook validation is the CALLER's, not this layer's</b> — <see cref="ApplyVerb"/> is public and does
+/// no rulebook check (it makes its own structural refusals, but nothing about identity fields, verb-vs-cardinality,
```
Avick3110: **[correctness / medium]** The parenthetical names three refusals `ApplyVerb` actually makes.

Traced in this same file on this branch:

- **verb-vs-cardinality** — `ApplyDictVerb`'s `default:` arm throws `Verb 'SetAtIndex' is not valid on dict 'SkillValues'`.
- **enum-key legality** — `ApplyDictVerb` runs the key through `Coerce(req.Key!, kType)` → `TryEnum`, so `Key = "BogusSkill"` throws.
- **range** — `TryPrimitive`'s `byte.Parse("999", …)` throws `OverflowException` before any `SetValue`.

So the closing clause — *"a direct or CLI caller that skips pre-flight mutates unvalidated"* — overstates the exposure: for three of the four named classes the engine refuses loudly, it just refuses with a CLR/structural message instead of a corpus-derived one. Only the identity-field item (and then only for *settable* identity fields such as `EditorID`; get-only `FormKey` is caught by `ApplyScalarVerb`'s `!prop.CanWrite` throw) and corpus-declared legality beyond CLR reach are genuinely rulebook-only.

**Scenario:** a maintainer auditing whether the pre-flight-bypassing path is dangerous reads this header, concludes those three classes land as unvalidated writes, and either builds a redundant guard or mis-scopes a Q3 review of the bypass.

Rounds 1 and 2 reversed direction on this sentence once already; round 2's correction ("what it omits is the *rulebook* check") is right — it's the enumeration left behind that the code contradicts.

src/housecarl-generator/WriteDiagnosticsProbe.cs
```diff
@@ -0,0 +1,359 @@
...
+
+        void Probe(string label, WriteRequest r, bool expectOk)
+        {
+            var msg = rulebook.Validate(r);
+            var ok = msg is null;
+            var mark = ok == expectOk ? "OK" : "!! UNEXPECTED";
+            Console.WriteLine($"  [{mark}] {label} -> {(ok ? "ACCEPT" : "REJECT: " + msg)}");
+        }
+
+        Console.WriteLine("Pre-flight probes (1 accept + 5 fail-loud rejects):");
```
Avick3110: **[correctness / low]** The banner miscounts the probes printed directly under it.

Lines 105–113 are nine `Probe()` calls: **2 accepts** (the real write, and `TranslatedString accept: Npc.Name=Bob`) and **7 rejects**. The banner says `1 accept + 5 fail-loud rejects`.

**Scenario:** `npc-skills` prints the banner and then nine result lines. Anyone reading this mode's output as coverage evidence — its stated purpose — cannot tell from the banner whether probes were added since it was written or whether the harness silently skipped some. That is exactly the ambiguity a fail-loud proof exists to remove.

The count was already stale on base, so this is not a regression — but it is being carried verbatim into a file this branch creates, which is the same ground round 2 used to fold the `condition-patch` comment block (DOC_HYGIENE §8, "cleaned when a file is touched").

```suggestion
        Console.WriteLine("Pre-flight probes (2 accepts + 7 fail-loud rejects):");
```

src/housecarl-generator/WriteDiagnosticsProbe.cs
```diff
@@ -0,0 +1,359 @@
...
+/// </list>
+///
+/// They lived in <see cref="WriteEngine"/> until #453 moved them here, beside the probes they resemble. The
+/// engine's own header lists what stayed behind.
+/// </summary>
+public static class WriteDiagnosticsProbe
+{
+    // A convenience default for the two modes that read a plugin; both take a path as argv[0] instead. It names
+    // one machine's install and is not assumed to exist — each of those two refuses by name when it does not.
+    const string DefaultSourcePath =
```
Avick3110, line 33: **[correctness / low]** #453's fix is partial, and this branch newly emits the message it was filed against.

Verified on a machine with no Skyrim at that path, base vs branch:

- **base:** `poly-probe` with no argument crashed — unhandled `DirectoryNotFoundException`, exit 127.
- **branch:** exit 1 with `error: source plugin not found: C:\Program Files (x86)\Steam\steamapps\common\Skyrim Special Edition\Data\Skyrim.esm`

So the round-1 fold is a strict improvement. The point is narrower: that message is verbatim the defect the PR body cites for `condition-patch` — *"naming a path the caller never typed"* — and `poly-probe` did not emit it before this branch. `npc-skills` emits it too (line 47).

The comment on line 31 blesses this ("is not assumed to exist — each of those two refuses by name"), which is accurate about the *guard* but sidesteps why the same default was removed from core. Flagging so the gate review can decide whether the sibling treatment — required positional/`--source`, no baked-in install path — should extend to these two, or whether the split is deliberate because they are hand-run generator diagnostics rather than shipping-library code.

Not blocking either way; the branch is verified clean otherwise (`ci-all` 132/132 reproduced independently, `write-api` and `substruct-probe` byte-identical base vs branch, `condition-patch` refusal confirmed in both directions).

src/housecarl-core/WriteEngine.cs
Avick3110: **Folded** — `6aa9f73`. All three traces reproduced in this file on this branch: `ApplyDictVerb`'s `default:` arm (`Verb '…' is not valid on dict '…'`), `Coerce(req.Key!, kType)` → `TryEnum` → `Enum.Parse` on a bogus key, and `byte.Parse` at `WriteEngine.cs:2959`. The sentence now says the engine refuses those with a CLR/structural message rather than a corpus-derived one, and names the residue — record identity, and legality the corpus declares that no CLR type expresses — as what a pre-flight-skipping caller actually mutates unvalidated.

**One correction to the finding, measured rather than taken.** The parenthetical puts get-only `FormKey` inside `ApplyScalarVerb`'s `!prop.CanWrite` throw. `FormKey` is **settable** on concrete Mutagen records — `Weapon.FormKey` and `Npc.FormKey` both report `CanWrite=True` by reflection, and `generated/corpus.json` carries **137** fields flagged identity *and* writable. So `!prop.CanWrite` does not catch it; only the rulebook's flat `IsIdentity` reject (`CorpusRulebook.cs:288`) does — which `write-diagnostics-probe`'s own `identity reject: Npc.FormKey` probe demonstrates. That makes identity a *larger* rulebook-only class than the finding stated, so the fold's direction is unchanged; the header is written from the measurement.

Also checked and deliberately not claimed: `EditorID` is not flagged identity by the corpus and no rulebook rule gates it, so it is not part of the residue.

Doc-only; no conditional added. `ci-all` 132/132.

src/housecarl-generator/WriteDiagnosticsProbe.cs
Avick3110: **Folded** — `fcebadf`. Counted at source: nine `Probe()` calls, `expectOk: true` on the real write and on `TranslatedString accept: Npc.Name=Bob`, `expectOk: false` on the other seven. Banner now reads `2 accepts + 7 fail-loud rejects`.

Agreed on the ground — stale on base, not a regression, but carried verbatim into a file this branch creates, same as round 2's `condition-patch` fold. `ci-all` 132/132.

src/housecarl-generator/WriteDiagnosticsProbe.cs
Avick3110: **Held — Aaron-ruled, no branch change.** Scope follows #453's issue text (production library only), which is the same line the body's recorded `WriteOracle` call sits on; the baked-in default stays on `poly-probe` and `npc-skills`.

Your measurement is confirmed and not in dispute: `WriteDiagnosticsProbe.DefaultSourcePath` is the constant #453 removed from the core, carried across with the two modes that moved, and both refuse by name when it does not exist.

The generator-side residue — `poly-probe`, `npc-skills`, `WriteOracle`, one constant spelled twice — is now noted on #463 as part of that issue's console-mode residue, citing this thread: https://github.com/Avick3110/houseCARL/issues/463#issuecomment-5494391634

Avick3110:
## Directed folds from the gate review — two folded, one held

Fresh fold session (2026-08-28 rule); directive carried verbatim from Aaron's ruling of 2026-09-01. **No new review rounds** (CLAUDE.md §5 #11). Both folds reproduced against source first; per-thread replies carry the traces.

| # | Finding | Commit | Disposition |
|---|---|---|---|
| 1 | header's bypass-exposure sentence overstates | `6aa9f73` | folded, **with one correction to the finding, measured** |
| 2 | probe banner miscounts (1+5 vs 2+7) | `fcebadf` | folded |
| 3 | baked-in default path on `poly-probe` / `npc-skills` | — | **HELD, Aaron-ruled** |

**Finding 1's correction.** The finding placed get-only `FormKey` inside `ApplyScalarVerb`'s `!prop.CanWrite` throw. `FormKey` is **settable** on concrete records — `Weapon.FormKey` / `Npc.FormKey` both `CanWrite=True`, and `generated/corpus.json` carries 137 fields flagged identity *and* writable — so only the rulebook's flat identity reject stops it. Identity is therefore a *larger* rulebook-only class than reported; the header is written from that measurement. `EditorID` was checked and deliberately excluded (not flagged identity, no rulebook rule gates it).

**Finding 3 held:** scope follows #453's issue text (production library only), the same line as the body's recorded `WriteOracle` call. The generator-side residue — `poly-probe`, `npc-skills`, `WriteOracle`, one constant spelled twice — is noted on #463 instead: https://github.com/Avick3110/houseCARL/issues/463#issuecomment-5494391634

Both folds are doc/comment-only; no conditional added, so no new arms. **`ci-all` 132/132.**

PR #465, opened 2026-09-01
fix(schema): published tool schemas carry no $ref, so strict providers can load the server

### LINKED ISSUE #451: [Bug] Recursive $ref in tool inputSchemas breaks strict providers — "Recursive JSON schemas are not currently supported" (Muse Spark 1.2 via opencode-go)
## Summary

Several houseCARL tools declare **self-referential (recursive) JSON schemas** in their `inputSchema` (e.g. `housecarl_cross_plugin_query` / `housecarl_bulk_apply` / `housecarl_bulk_create`, where `compose.sets[].compose` and `operations[].compose` reference their own type, and the `where` predicate is a tree). Some model providers that validate `tools/list` schemas strictly **reject the whole server at session start**, so no conversation with houseCARL enabled is possible on those models — the only workaround is disabling houseCARL per-session.

## Exact error

```
Error from provider (Console Go): Upstream request failed: [invalid_request_error] Recursive JSON schemas are not currently supported
```

- Provider: `opencode-go` (Console Go relay), model `muse-spark-1.2-contributor`
- Trigger: houseCARL MCP v1.9.0 enabled -> the provider validates every tool's `inputSchema` at `tools/list` time and fails the **entire session** (not just the offending tool call)
- Other models on the same provider (deepseek-v4-pro, glm-5.x, qwen3.7-plus, kimi-k2.x ...) accept the same schemas fine — only strictly-validating models reject recursion

## Repro

1. Enable housecarl MCP in opencode (local stdio, `housecarl-mcp.exe`)
2. Select model `muse-spark-1.2-contributor` via the `opencode-go` provider
3. Send any message -> the above error, every time
4. Disable housecarl -> the model works; re-enable -> error returns

The recursive shapes are the polymorphic `compose` structs — `compose.sets[].compose` -> `compose` (and the same inside `composes[]`), plus the nested `where` condition tree. Recursive $ref is legal JSON Schema and natively supported by Anthropic/OpenAI APIs, but a growing set of smaller/relay models validate schemas conservatively and reject recursion outright.

## Why it matters

- Users on strict providers cannot use houseCARL at all without toggling MCP off/on per session — easy to forget, and the error message does not name houseCARL as the culprit.

## Suggested fix

Flatten the recursive $ref at `tools/list` serialization time to a **bounded depth** (e.g. depth 3) of plain inline object/array schemas (matching what a strict validator accepts), while keeping `tools/call` parameter handling unchanged. Optionally a config switch (`HouseCarl__FlatToolSchemas=true`) so the recursive form stays available for providers that support it.

Alternatively, documenting a known-incompatible-provider list would help users self-diagnose faster.

## Environment

- houseCARL 1.9.0 (local stdio via `housecarl-mcp.exe`, opencode desktop)
- Windows 11, .NET 9
- Affected provider: opencode-go / Console Go, model `muse-spark-1.2-contributor`

Avick3110:
Fixes #451.

Five tools published a self-referential `inputSchema`. A provider that validates `tools/list` strictly answers `Recursive JSON schemas are not currently supported` and refuses the **whole server** — naming neither houseCARL nor a tool, so the reporter's only way through was turning the MCP server off for that model. Every same-document `$ref` is now inlined at registration: a cycle is expanded to a bound and closed with an open node that keeps the type, stops constraining, and says in its own description that nesting continues. Measured on the served surface: **30 `$ref` sites across 5 tools → 0 across all 51**, `tools/list` +5.93%, the five affected schemas +2.96–3.22 KB each.

Not a config switch. The report suggested one; it was declined, because the flattened form is accepted everywhere the recursive one is, so the knob buys nothing reachable, leaves the default broken for exactly the users who hit this, and doubles what the guards must hold.

---

## The population correction that decided the guard design

The staged charter for the final wave carried a reasoned-not-run route saying the emission-grammar arm should read the raw pre-flatten `InputSchema` for all 51 tools. Implemented literally, **that yields 18 ref sites. The pass actually handles 30.** `housecarl_apply` and `housecarl_create` emit no `$ref` at all until `RewriteFileListUnions` generates their element-array arm, so a raw-only arm would have left the entire generated-arm population — 12 of 30 sites — ungoverned, and would not have seen a `$defs` the union pass introduced either.

The arm now replays the union pass in the same order the server runs it, and prints both populations so the split stays visible:

```
…they carry $ref sites for the per-site rules to run over
(30 in the pass's input; 18 straight from the assembly scan, the rest added by the @file union pass)
```

The derived population contradicted a hand-carried number, and the contradiction was only visible because the subjects came from the data. That is the ruling this PR ships under doing its job on its first run.

## Rounds, the class-stop, and the ruling

**Two rounds, five independent reviewers, fresh agents in their own worktrees.** Round 1 returned one class from two reviewers (~9 instances, all on synthetic repros): `FlattenRefs` is written and documented as a general JSON-Schema `$ref` inliner but is only correct for this SDK generator's emissions. Disposition: **drop** the algorithm fixes (general implementation is a different charter), **fold** the claim narrowing so the code and probe docs say they normalize *this* generator's output, and ship the narrowing with its precondition **pinned** by an arm over the pre-flatten grammar — unreachability had been claimed by review attention, which never carries a completeness claim.

**Round 2 class-stopped.** One class recurred: *a claim quantified over a population, verified over a hand-named subset or a narrower predicate.* Second project-level appearance (first: PR A / #441, ruled 2026-08-27). Three of the instances were introduced by the review session's **own round-1 folds**, which is why folding stopped rather than continuing to a third round. Instances:

- the detector's predicate (`$ref` value `StartsWith('#')`) was the same test the flattener gates inlining on — anchor-form `{"$ref":"Node"}` injected into all 51 schemas passed **green**;
- the anti-amputation arm walked `housecarl_apply` alone, so four of the five ref-carrying tools had no expansion arm — a bound-0 collapse of the 1.x subtree stayed green on **both** guards;
- ARM 7 re-derived six hand-named DTO schemas through its own generator call instead of reading the real emission path, so its drift claim was never established for the path that actually drifts;
- prose naming a reader that does not exist (`ApplyTools.ReadListParam`; the reader is `ListParams.Read<T>`), and an ADR contradicting itself 17 lines apart on whether anything reads a published schema.

**Aaron ruled the revision:** *"Derive, don't enumerate. I thought we had learned our lesson here."* Option (i), derive-from-surface, with three pins — the invariant predicate is zero `$ref` **members**, any spelling; the expansion arm's subjects are enumerated from the pre-flatten surface at guard time; ARM 7 reads the real `ProtocolTool.InputSchema` for all 51 tools. **One directed wave executed it, and there was no round 4** (directed-completion rule): fresh eyes are for before the PR opens, and a directed fold does not reopen the loop.

Deriving also **removes the residual 1.x coupling**: the subject set shrinks by itself when the demolition deletes the three 1.x ref-carrying tools, where a hand-list would have gone stale and passed.

## What was refused, and on what ground

- **The output-size blowup mitigation — refused in round 1, and the refusal stands.** No mitigation code was written. A size ceiling over a case with no producer is the #426 shape: a conditional asserting what the codebase cannot emit. The measurements (a fully-connected mutual-reference clique: n=7 → 0.9 s / 1.6 MB, n=8 → 8.5 s / 13 MB, n=9 → 203 s / 122 MB) ride as documentation of why the general-inliner claims were narrowed, and for nothing else.
- **The general-inliner algorithm fixes — dropped, not folded.** Nine instances, all verified unreachable from today's generator (empty RFC 6901 tokens, `$anchor` misparse, no percent-decoding, value-position `$ref`, boolean schemas, nested `$defs` survival, sibling-merge widening). Making `FlattenRefs` a correct general inliner is a different charter; what shipped instead is the narrowed claim **plus an arm pinning the grammar it depends on**, so the unreachability is measured rather than asserted.
- **A runtime fail-safe that catches and republishes the recursive schema — refused.** Silent republication is the wrong side of Q3. Loudness here is guard-at-CI, this codebase's standing pattern.
- **Chained refs were not ruled in as separate work.** Nobody decided them; they are a **consequence** of the ruled predicate — a surviving chained ref is a `$ref` member, so the widened invariant reddens on it by construction.
- **`MaxSelfExpansions` stays 1** — at 0 the bound fires on the first occurrence and open-nodes the non-cyclic `fields`/`ctor_args` leaves too, losing real element shapes rather than merely deep nesting. Aaron saw the cost figure and settled it: *"keep 1."*

## Guards

- **`binding-shim-guard`, SCHEMA arms** — over the real served surface, driving the built `housecarl-mcp.exe` over stdio. The invariant is now "no published tool schema carries a `$ref` **member**, in any spelling", deliberately **wider** than the flattener's own resolve gate and sharing no code with it; resolvability is reported as detail, not as the test. The anti-amputation subjects are derived at guard time — this run: *30 pre-flatten `$ref` sites across 5 tools, 10 of them recursive*, with no tool named in the guard. The recursion **step** is derived too (the one site whose pointer is a prefix of its own path is the positional back-reference), which keeps the walk indifferent to `[JsonPropertyName]` declaration order. Both directions are asserted non-empty, so a derivation finding nothing fails instead of passing vacuously. The bound is spelled independently in the guard (`levels == 1`) rather than read from the constant — changing a published contract must be re-ratified, so it reddens.
- **`schema-flatten-guard`** (new, in `ci-all`) — 27 checks. The mechanism over synthetic documents for the shapes today's DTOs cannot produce (`$defs` recursion, mutual recursion, sibling merge, an unresolvable pointer); ARM 6 drives the **real reader** to measure the sentence the terminator publishes; ARM 7 pins the emission grammar over the real pre-flatten surface.

**ARM 6 measures rather than asserts.** The terminator writes *"it is accepted but not spelled out again here"* into the published schema — a claim about accept behaviour that a later call can falsify. Measured through `ListParams.Read<ApplyOp>`: 5/10/15/20 compose levels read with the graph intact to the innermost node; past the JSON reader's own `MaxDepth` (~20 levels, an order of magnitude past live composes, which nest 1–2) it is a **named refusal carrying the element path**, never a silent truncation. No wire-side twin, deliberately: `ops` is declared `JsonElement`, so an arm over the SDK binder could not fail — the #426 shape again.

## Sabotage — 9 cells, run from a committed state

**The canary came first.** C1 (anchor-form `{"$ref":"Node"}` injected into every published schema) was proven RED **by hand, before the sweep script existed**, and only then used as that script's known-RED canary. The ordering is the point: a canary invented alongside the harness it is meant to validate proves nothing about the harness. Builds are captured to a buffer, never piped into `grep -q` (a SIGPIPE'd build leaves the old binary and the guard then measures stale code); every restore is grep-verified and the sweep aborts if any file still holds sabotage.

| Cell | Sabotage | Guard | Verdict |
|---|---|---|---|
| C1 **canary** | anchor-form `{"$ref":"Node"}` in every published schema | binding-shim | RED — invariant arm, **51 found** |
| C2 | effective bound of 0 on the three 1.x tools' subtree | binding-shim | RED — anti-amputation, on **all three** 1.x ref-carrying tools, **24 checks** |
| C3 | the `FlattenRefs` call removed | binding-shim | RED — invariant arm, **30 found** (the pre-fix population exactly) |
| C4 | `MaxSelfExpansions = 2` | binding-shim | RED — "expands exactly 1 level (got 2)" |
| C5 | the terminator's honesty clause reworded | binding-shim | RED — "closes on an OPEN node" |
| C6 | ARM 7's ref-node member set tightened to `{$ref}` | schema-flatten | RED — merge-rule members |
| C7 | ARM 7's no-`$defs` check inverted | schema-flatten | RED — `$defs` arm |
| C8 | `PreFlattenSurface.RefNodes` emptied | binding-shim | RED — "derived subject set is non-empty — 0 sites across 0 tools" |
| C9 **inverse** | `compose`/`composes` `[JsonPropertyName]` order swapped | binding-shim | **GREEN**, as required — a member reorder is not an amputation |

**C1 and C2 are the two regressions that stayed fully green through both review rounds.** C2 now reddens on **`housecarl_create_record`, `housecarl_bulk_create` and `housecarl_bulk_apply` — 8 failing checks each, 24 in total**. Those are exactly the three 1.x tools whose ref pointers route through that subtree; `housecarl_apply` and `housecarl_create` address theirs through `/properties/ops/` and are untouched by this particular cell. **None of the three had an expansion arm before the wave** — the only arm walked `housecarl_apply`, which this sabotage does not touch, which is why it stayed green twice. That is the derived subject set covering three tools nobody would have thought to add by hand, stated as a measurement rather than as a claim.

*(The sabotage table in the branch's decision record named only `housecarl_create_record` here. The population was re-measured at source and corrected in that record's addendum 5 — an expectation narrower than its measured population being, pointedly, this branch's own failure class. The correction only strengthens the claim.)*

## Independent completion verification

The rounds, the wave, and this PR were each conducted by a different session; the branch's author ran none of them. A fresh verifier re-ran the acceptance tests from the committed state at `2db5789`, restoring and grep-verifying after each cell:

| Test | Expected | Measured |
|---|---|---|
| Anchor-form injection after the pass | invariant arm RED, 51 found | **RED, 51 found** |
| Bound-0 collapse on the 1.x subtree | derived expansion arm RED | **RED on 3 tools, 24 checks ("got 0")** |
| `FlattenRefs` call removed | invariant arm RED, 30 found | **RED, 30 found** |
| `MaxSelfExpansions = 2` | "got 2" | **RED, "the ruled bound; got 2"** |
| Terminator clause reworded | "closes on an OPEN node" | **RED** |
| ARM 7 members tightened to `{$ref}` | merge-rule arm RED | **RED, 26/27** |
| ARM 7 no-`$defs` check inverted | `$defs` arm RED | **RED, 26/27** |
| `compose`/`composes` order swapped | both guards GREEN | **GREEN** (subject set still 30/5) |
| `ci-all` | 133/133 | **133/133 ALL PASS** |

ADR 0002's figures were re-measured against the branch as it stands rather than taken from the record: 5 tools declare `compose` and they are exactly the 5 that carried refs; 10 terminator nodes on the surface = the 10 recursive sites; +5.93% on `tools/list` and +2.96–3.22 KB per affected schema; and `ToolCallShim` contains **zero** occurrences of `items` or `anyOf`, which is the ADR's "never descends" claim. The ADR is ARCHIVE the moment this merges, so this was its last edit window.

## Scope

Engine files only — `src/housecarl-mcp/ToolSchemas.cs` plus generator probes. **`RewriteFileListUnions` became `internal`** — the one non-probe code change, the same probe-access pattern `FlattenRefs` already uses, and it exists so ARM 7 replays the union pass rather than re-implementing it; a re-implementation would be a second emission resembling the real one, which is the defect the wave removed. No fix work in any 1.x tool body. Docs: the essay header moved out of `ToolSchemas.cs` to `docs/architecture/tool-schema-publication.md`, ADR 0002 records the invariant, and the `## Unreleased` CHANGELOG line is on the branch.

## For the reply to the reporter

- The recursion was real, is reproduced, and is fixed — thank you for the report; an error naming neither the server nor a tool is exactly why it needed measuring rather than guessing.
- **The `where=` half of the diagnosis is wrong.** `housecarl_cross_plugin_query` and `housecarl_records` publish `where=` as `{"type":["array","null"],"items":{"type":["string","null"]}}` — a string list, not a predicate tree. There are no refs anywhere in them. All 30 refs came from one shape: `StructInput` → the `sets[]` element → its `compose` → `StructInput`.
- **The fix is unconditional — there is nothing to configure.** The suggested `HouseCarl__FlatToolSchemas=true` was declined on purpose: the people who would need to set it are the ones who cannot load the server to set it.

### DIFF
--- a/docs/decisions/0002-published-tool-schemas-carry-no-ref.md
+++ b/docs/decisions/0002-published-tool-schemas-carry-no-ref.md
@@ -27,7 +27,8 @@ feature that some validators reject is a liability regardless of who is technica
 1. A cycle is expanded a bounded number of times and then **closed with an open node** — the
    target's `type`, the parameter's own description, and a clause saying nesting continues.
    Nothing is narrowed: the schema at the bound accepts what the recursive form accepted, and
-   what a tool accepts was never decided by the published schema.
+   nothing reads the nested part of a published schema, so rewriting it cannot move what a
+   tool accepts (see the Consequences below for what does read one).
 2. `$defs` is dropped once nothing refers to it — an unreferenced definition still carries its
    cycle to a validator that walks definitions.
 3. A pointer that does **not** resolve is left in place, not replaced by an open node. A broken
@@ -43,8 +44,13 @@ feature that some validators reject is a liability regardless of who is technica
   (~6% of `tools/list`). Raising the bound deepens every recursive branch of every schema.
 - The invariant is guarded generically over all tools, so a future recursive DTO is covered
   by construction rather than by remembering this decision.
-- Anything that consumes a published schema may now assume it is self-contained. Nothing
-  currently does — the argument-binding shim reads only top-level `properties`.
+- Anything that consumes a published schema may now assume it is self-contained. One thing
+  does: the argument-binding shim is schema-driven off `InputSchema`, coercing argument shapes
+  and refusing unknown or missing parameters. It reads only the top-level `properties` and
+  never descends into `items`/`anyOf`, which is the whole of what this pass rewrites — so the
+  flattening is invisible to it. That is a measured property of today's shim, not a rule
+  binding it: a future consumer that walks nested schemas is now free to, and would have been
+  reading a cyclic document before.
 - Should a provider one day reject something else in these schemas, the fix has a home: this
   is a publication layer that already normalizes what the generator emits.
 
--- a/docs/decisions/0002-published-tool-schemas-carry-no-ref.md
+++ b/docs/decisions/0002-published-tool-schemas-carry-no-ref.md
@@ -1,0 +1,51 @@
+# ADR 0002 — Published tool schemas carry no `$ref`
+
+**Class:** ARCHIVE (ADR). Immutable once merged; superseded by a later ADR, never edited.
+**Date:** 2026-09-01 · **Status:** accepted · **Issue:** #451
+
+## Context
+
+houseCARL's write DTOs are recursive by design: a composed struct carries nested edits, and a
+nested edit may compose another struct. The MCP SDK's schema generator does not expand a
+recursive type — it inlines one level and terminates the next with a positional back-reference
+(`$ref` at an ancestor), so five tools published a cyclic `inputSchema`.
+
+That is legal JSON Schema. The Anthropic and OpenAI APIs accept it, and houseCARL shipped it
+for months. But an external report (#451) measured a provider that validates `tools/list`
+strictly and answers `Recursive JSON schemas are not currently supported` — refusing the
+**entire server**, not the offending tool, with an error naming neither houseCARL nor the
+schema. The user's only workaround was toggling the MCP server off for that model.
+
+The set of clients between a model and a tool list is not ours and keeps growing. A schema
+feature that some validators reject is a liability regardless of who is technically correct.
+
+## Decision
+
+**No published tool schema contains a same-document `$ref`.** At registration, after the
+`@file` union rewrite, every same-document pointer is inlined:
+
+1. A cycle is expanded a bounded number of times and then **closed with an open node** — the
+   target's `type`, the parameter's own description, and a clause saying nesting continues.
+   Nothing is narrowed: the schema at the bound accepts what the recursive form accepted, and
+   what a tool accepts was never decided by the published schema.
+2. `$defs` is dropped once nothing refers to it — an unreferenced definition still carries its
+   cycle to a validator that walks definitions.
+3. A pointer that does **not** resolve is left in place, not replaced by an open node. A broken
+   rebase must stay visible rather than hide behind a schema that looks finished.
+4. This is unconditional, not a config switch. The flattened form is accepted everywhere the
+   recursive form is, so a switch would add a knob with no reachable benefit, leave the default
+   broken for the users who hit this, and double what the guards must hold.
+
+## Consequences
+
+- The recursive shape stays legible to a caller — it is spelled out concretely to the bound
+  rather than amputated — at the cost of schema size. The five affected tools grew ~3 KB each
+  (~6% of `tools/list`). Raising the bound deepens every recursive branch of every schema.
+- The invariant is guarded generically over all tools, so a future recursive DTO is covered
+  by construction rather than by remembering this decision.
+- Anything that consumes a published schema may now assume it is self-contained. Nothing
+  currently does — the argument-binding shim reads only top-level `properties`.
+- Should a provider one day reject something else in these schemas, the fix has a home: this
+  is a publication layer that already normalizes what the generator emits.
+
+Mechanics and the rest of the layer: `docs/architecture/tool-schema-publication.md`.
--- a/docs/decisions/0002-published-tool-schemas-carry-no-ref.md
+++ b/docs/decisions/0002-published-tool-schemas-carry-no-ref.md
@@ -21,8 +21,14 @@ feature that some validators reject is a liability regardless of who is technica
 
 ## Decision
 
-**No published tool schema contains a same-document `$ref`.** At registration, after the
-`@file` union rewrite, every same-document pointer is inlined:
+**No published tool schema carries a `$ref` member, in any spelling.** That is wider than
+"contains a same-document pointer" on purpose: spelling the invariant the same way as the
+flattener's own resolve gate is what made an earlier version of the guard vacuous — a `$ref`
+the pass could not resolve was also one the detector could not see, so an anchor-form
+`{"$ref":"Node"}` injected into all 51 schemas passed green. The guard asks only whether the
+member is there.
+
+At registration, after the `@file` union rewrite, every same-document pointer is inlined:
 
 1. A cycle is expanded a bounded number of times and then **closed with an open node** — the
    target's `type`, the parameter's own description, and a clause saying nesting continues.
@@ -31,8 +37,11 @@ feature that some validators reject is a liability regardless of who is technica
    tool accepts (see the Consequences below for what does read one).
 2. `$defs` is dropped once nothing refers to it — an unreferenced definition still carries its
    cycle to a validator that walks definitions.
-3. A pointer that does **not** resolve is left in place, not replaced by an open node. A broken
-   rebase must stay visible rather than hide behind a schema that looks finished.
+3. A `$ref` the pass does **not** handle — a pointer that does not resolve, a form that is not
+   a same-document pointer at all — is left in place, not replaced by an open node and not
+   thrown on. A broken rebase must stay visible rather than hide behind a schema that looks
+   finished, and the wider invariant above is what makes leaving it a report rather than a
+   silence.
 4. This is unconditional, not a config switch. The flattened form is accepted everywhere the
    recursive form is, so a switch would add a knob with no reachable benefit, leave the default
    broken for the users who hit this, and double what the guards must hold.
--- a/src/housecarl-generator/BindingShimProbe.cs
+++ b/src/housecarl-generator/BindingShimProbe.cs
@@ -572,6 +572,11 @@ public static class BindingShimProbe
     {
         int failures = 0;
 
+        // The derivation the expansion arms below stand on, fixtured against shapes today's DTOs cannot produce
+        // end-to-end. Its two rules are visible only under those shapes, so nothing on the real surface would
+        // redden if either were dropped.
+        failures += DerivationFixtureArm();
+
         // One row per ToolSchemas.FileListParams entry — (tool, parameter, a member that proves the generator ran
         // over the real C# type). PR #311 review [low]: this arm used to hardcode housecarl_apply, so
         // housecarl_create's records= row was covered only by the generic dangling-$ref sweep — and a dropped or
@@ -675,18 +680,18 @@ public static class BindingShimProbe
                 found, "tool absent from tools/list");
             if (!found) continue;
 
-            // The recursion STEP, derived: the one site whose pointer is a prefix of its own path is a positional
-            // back-reference to an ancestor, and the path segment between them IS one turn of the cycle. Every site
-            // aiming at that same pointer re-enters the same shape, whichever member the generator reached first —
-            // so the walk below is indifferent to [JsonPropertyName] declaration order, which an earlier fixed-level
-            // pin was not (reordering ApplyOp.Compose/Composes made it report an amputation that had not happened).
-            string? step = null, cycle = null;
-            foreach (var (path, node) in sites)
-                if (node["$ref"]?.GetValue<string>() is { } p && path.Length > p.Length
-                    && path.StartsWith(p, StringComparison.Ordinal))
-                { step = path[p.Length..]; cycle = p; }
-            failures += Check($"SCHEMA: {tool.Name} — a recursion step is derivable from its own pre-flatten refs",
-                step is not null, "no $ref points at one of its own ancestors; the expansion arm has nothing to measure");
+            // The recursion STEP and CYCLE, derived from this tool's own refs — the rules live in DeriveRecursions.
+            // EXACTLY one is required. A tool carrying two distinct cycles is refused here, never measured on
+            // whichever one the walk saw last: that is a claim quantified over a population and verified over a
+            // narrower subset, which is the class this branch was stopped on.
+            var derived = DeriveRecursions(sites);
+            failures += Check($"SCHEMA: {tool.Name} — exactly one recursion step is derivable from its own pre-flatten refs (got {derived.Count})",
+                derived.Count == 1,
+                derived.Count == 0
+                    ? "no $ref points at one of its own ancestors on a segment boundary; the expansion arm has nothing to measure"
+                    : "distinct cycles: " + string.Join(" | ", derived.Select(d => $"{d.Cycle} +{d.Step}")));
+            string? step = derived.Count == 1 ? derived[0].Step : null;
+            string? cycle = derived.Count == 1 ? derived[0].Cycle : null;
 
             foreach (var (path, node) in sites)
             {
@@ -696,7 +701,7 @@ public static class BindingShimProbe
                     pubNode is { ValueKind: JsonValueKind.Object } n && SpellsOutStructure(n),
                     pubNode is { } got ? Trunc(got) : "<path does not resolve in the published schema>");
 
-                if (step is null || node["$ref"]?.GetValue<string>() != cycle) continue;
+                if (step is null || RefPointer(node) != cycle) continue;
                 recursive++;
                 var (levels, closedOpen, detail) = WalkRecursion(published, path, step);
                 // The bound is a RULED value (decision record entry 15, Aaron: "keep 1"), spelled here independently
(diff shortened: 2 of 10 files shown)


src/housecarl-generator/BindingShimProbe.cs
```diff
@@ -608,62 +610,211 @@ static int SchemaArm(JsonElement toolsList)
...
+
+            // The recursion STEP, derived: the one site whose pointer is a prefix of its own path is a positional
+            // back-reference to an ancestor, and the path segment between them IS one turn of the cycle. Every site
+            // aiming at that same pointer re-enters the same shape, whichever member the generator reached first —
+            // so the walk below is indifferent to [JsonPropertyName] declaration order, which an earlier fixed-level
+            // pin was not (reordering ApplyOp.Compose/Composes made it report an amputation that had not happened).
+            string? step = null, cycle = null;
+            foreach (var (path, node) in sites)
+                if (node["$ref"]?.GetValue<string>() is { } p && path.Length > p.Length
+                    && path.StartsWith(p, StringComparison.Ordinal))
```
Avick3110: **[medium] The step/cycle derivation takes the *last* match and tests the ancestor relation with an unanchored string prefix.**

Two distinct failure paths, both of which the derived-subjects design is meant to close:

**(a) Silent under-coverage.** The loop overwrites `step`/`cycle` on every matching site rather than asserting the match is unique. If a tool ever carries two distinct recursive pointers (two recursive DTO families reachable from one tool), only the last one's sites clear `node["$ref"] != cycle` and reach the `levels == 1` / `closes on an OPEN node` checks. The other cycle's bound is never measured, `recursive > 0` still passes, and the guard reports coverage it does not have — a claim quantified over a population, verified over a narrower subset, which is the class this branch was stopped on.

**(b) False RED on a correct schema.** `path.StartsWith(p)` has no JSON-pointer segment boundary. Today it survives by one character: `#/…/properties/compose` vs `#/…/properties/composes/items` diverges at `/` vs `s`. Give `StructInput` a member whose wire name *extends* an existing ref-target member name — e.g. `sets_batch` beside `sets`, exactly the shape of the existing `compose`/`composes` pair — and the site at `#/…/properties/sets_batch/…` matches its own `$ref` `#/…/properties/sets` as a prefix. `step` becomes `_batch/…`, `WalkRecursion` then walks pointers that do not resolve, and the guard reddens on a schema the pass flattened correctly.

Both are cheap to close: require the prefix to end on a `/` boundary, and fail if more than one distinct `(step, cycle)` is derivable for a tool instead of letting the last write win.

src/housecarl-mcp/ToolSchemas.cs
```diff
@@ -164,6 +161,108 @@ static void RebaseRefs(JsonNode? node, string basePointer)
...
+                break;
+        }
+        return changed;
+    }
+
+    /// <summary>The replacement for one <c>$ref</c> node, or null if the node carries no same-document ref.</summary>
+    static JsonNode? Expand(JsonNode? node, JsonObject snapshot, Dictionary<string, int> spent)
+    {
+        if (node is not JsonObject refNode) return null;
+        if (refNode["$ref"]?.GetValue<string>() is not { } pointer || !pointer.StartsWith('#')) return null;
```
Avick3110: **[low] A non-string `$ref` throws here instead of being left alone — inside a `PostConfigure`, so it takes the server's startup with it.**

`JsonNode.GetValue<string>()` on a `$ref` that is not a JSON string throws `InvalidOperationException`. `PublishSchemas` runs this from `services.PostConfigure<McpServerOptions>(…)`, which executes while the host is being built and has no try/catch, so the exception surfaces as an opaque host-construction failure on both transports: the whole server fails to start, naming neither houseCARL nor a tool. That is the same user-facing shape as #451 itself.

The class summary lists these spellings (`$anchor` fragments, boolean targets, value-position refs) as "outside what it handles" — i.e. left in place for the invariant arm to fail on, which is the documented and correct posture. Crashing is not that posture.

`SchemaFlattenProbe.EmissionGrammarArm` already uses the safe spelling for exactly this node:

```csharp
node["$ref"] is JsonValue v && v.TryGetValue<string>(out var s) ? s : null
```

Same one-line fix applies here and at `RebaseRefs` (line 148, pre-existing). ARM 7 does pin the grammar at CI, so this is defense-in-depth rather than a live defect — but the cost is one `is JsonValue` and the failure it prevents is the one the branch exists to eliminate.

```suggestion
        if (refNode["$ref"] is not JsonValue refValue || !refValue.TryGetValue<string>(out var pointer)
            || !pointer.StartsWith('#')) return null;
```

docs/decisions/0002-published-tool-schemas-carry-no-ref.md
```diff
@@ -0,0 +1,57 @@
...
+strictly and answers `Recursive JSON schemas are not currently supported` — refusing the
+**entire server**, not the offending tool, with an error naming neither houseCARL nor the
+schema. The user's only workaround was toggling the MCP server off for that model.
+
+The set of clients between a model and a tool list is not ours and keeps growing. A schema
+feature that some validators reject is a liability regardless of who is technically correct.
+
+## Decision
+
+**No published tool schema contains a same-document `$ref`.** At registration, after the
```
Avick3110: **[low] The ADR records the narrower predicate — and this is its last edit window.**

The Decision says **"No published tool schema contains a same-document `$ref`"**. The invariant that was actually ruled and is actually enforced is wider: `binding-shim-guard` asserts *"no published tool schema carries a `$ref` **member**, in any spelling"*, and its own comment explains why the narrowing was removed — spelling the detector the same way as the flattener's resolve gate is what made the earlier arm vacuous, so an anchor-form `{"$ref":"Node"}` injected into all 51 schemas passed green.

The PR body states the ruled predicate correctly ("the invariant predicate is zero `$ref` **members**, any spelling"). The ADR — the document a future engineer reads to learn what the invariant *is* — states the one it was deliberately widened away from, and `docs/architecture/tool-schema-publication.md:69` repeats it ("the one invariant the guard asserts — no published tool schema carries a same-document `$ref`").

Concretely: a later change that reintroduces a non-pointer `$ref` reads as ADR-compliant while reddening CI, with the ADR offering no way to tell which of the two is right. The ADR is ARCHIVE the moment this merges, so it cannot be corrected afterwards.

```suggestion
**No published tool schema contains a `$ref` member, in any spelling.** At registration, after the
`@file` union rewrite, every same-document pointer is inlined:
```

src/housecarl-generator/BindingShimProbe.cs
Avick3110: **Folded** — `6c931e4`. Both paths reproduced, one of them further than the finding claimed.

**(a) silent under-coverage — reproduced end-to-end.** Sabotage: a second `NestedSet[]` member on `StructInput`. The generator emitted a second cycle on all five ref-carrying tools, the later pair overwrote the first, and the guard stayed **fully green** while the recursive population went from `10 of 30` to `5 of 45` — the original `compose/sets` cycle measured on *no tool at all*, not merely on fewer sites.

**(b) false RED — confirmed at the code level, not end-to-end.** The unanchored prefix is real and the one-character survival is real: `#/…/properties/composes/items/properties/sets` refs `#/…/properties/compose/properties/sets`, and they diverge at `s` vs `/`. But the specific `sets_batch` producer does not reproduce: STJ emitted `sets_batch` at the *shorter* path as a forward ref to the deeper first full emission, so `path.Length > p.Length` fails and no false match arises. What that sabotage produced instead was (a). Recording it as measured rather than folding the trace as written.

**The fold.** `DeriveRecursions` now (i) requires the prefix to end on a `/` segment boundary and (ii) returns every distinct `(step, cycle)`; the caller requires **exactly one** and refuses the tool otherwise — refusal-shaped, never handle-multiple. With `derived.Count != 1`, `step` stays null, no site is measured, and `at least one derived subject is recursive` fails too, so the refusal cannot read as coverage.

**RED cells** (each from a committed state, restores grep-verified):

| sabotage | result |
|---|---|
| second `NestedSet[]` on `StructInput` (the (a) reproducer) | **RED** — 6 checks, both cycles named per tool; was fully green before |
| drop the segment-boundary test | **RED** — boundary fixture only |
| restore last-write-wins in `DeriveRecursions` | **RED** — two-family fixture only |

Neither rule has an end-to-end producer as the DTOs stand, so both ship as fixtures built from the pointers the sabotage actually emitted — the same standing as `schema-flatten-guard`'s `$defs` and mutual-recursion arms. `binding-shim-guard` back to green at `10 of 30`; `ci-all` 133/133.

The derivation also reads `$ref` through the safe `JsonValue` spelling now (see the finding-2 thread): a non-string `$ref` is what the invariant arm reports, not something that should take the guard down before it gets there.

src/housecarl-mcp/ToolSchemas.cs
Avick3110: **Folded** — `05d3108`. Trace reproduced: `GetValue<string>()` throws `InvalidOperationException` on all four non-string forms — `An element of type 'True'/'Number' cannot be converted to a 'System.String'` for scalars, `The node must be of type 'JsonValue'` for an object or array — and `PublishSchemas` does run from `services.PostConfigure<McpServerOptions>(…)` with no try/catch (`ToolSchemas.cs:64`, `Program.cs:170`).

Taken slightly differently from the suggestion: rather than the same spelling twice, both passes now read `$ref` through **one** `RefPointer` helper. `RebaseRefs`'s input is always generator output, so its own arm could not be fixtured honestly — one conditional means one arm covers it, instead of a second branch nothing can reach.

`ARM 5` grows that arm and is renamed to what it now measures (`a form the pass does not handle is left in place`): a non-string `$ref` does not throw, survives verbatim in all four forms, and the string pointer beside it is still inlined. RED-checked both ways from a committed state, restores grep-verified:

| sabotage | result |
|---|---|
| revert `RefPointer` to `GetValue<string>()` | **RED** — 3 checks; the throw also aborts the pass, so the *valid* pointer beside it goes un-inlined |
| make `RefPointer` return null unconditionally (over-reject) | **RED** — 2 checks, both "still inlined" arms |

Agreed on the standing: ARM 7 pins the grammar at CI, so this is defense-in-depth. The `FlattenRefs` header now states the invariant's ruled wider predicate (no `$ref` **member**, any spelling) rather than the narrower same-document one — leaving an unhandled form only reports anything under the wider predicate, which is the finding-3 thread's point in code.

docs/decisions/0002-published-tool-schemas-carry-no-ref.md
Avick3110: **Folded** — `c7198c5`. Verified at source: the guard asserts `SCHEMA: no published tool schema carries a $ref member, in any spelling` (`BindingShimProbe.cs`), while the ADR's Decision and `tool-schema-publication.md:69` both stated the narrower same-document form.

Both now carry the ruled predicate **and** the reason it is wider — the vacuous-arm history, so a future reader can see it was widened deliberately rather than read the narrowing back in. Point 3 also grew to cover what the branch now does with a form the pass cannot read (leave it — not an open node, not a throw), which was previously only "a pointer that does not resolve".

Taken as the ADR's last edit window, as flagged.

Avick3110:
## Directed folds from the gate review — all three folded

Fresh fold session (2026-08-28 rule); directive carried verbatim from Aaron's ruling of 2026-09-01 ("agreed on your recommendation"; vouching record `dev/advisor/decisions/2026-09-01.md`). **No new review rounds** — CLAUDE.md §5 #11, fresh eyes are for before the PR opens. Every finding was reproduced against source before folding; per-thread replies carry the traces.

| # | Finding | Commit | Disposition |
|---|---|---|---|
| 1 | step/cycle derivation: unanchored prefix + last-write-wins | `6c931e4` | folded; **(a) reproduced end-to-end, (b) confirmed at code level only** |
| 2 | non-string `$ref` throws out of `PostConfigure` | `05d3108` | folded, taken through one shared helper |
| 3 | ADR 0002 + architecture doc state the narrower predicate | `c7198c5` | folded (ADR's last edit window) |

**Finding 1 measured harder than reported.** Sabotage — a second `NestedSet[]` member on `StructInput` — gave all five ref-carrying tools a second cycle; the later pair overwrote the first and the recursive population went `10 of 30` → `5 of 45`, with the original `compose/sets` cycle measured on *no tool at all*, guard fully green. The reviewer's specific `sets_batch` false-RED producer did **not** reproduce (STJ emitted the shorter path as a forward ref, so the length test excludes it); the code-level hazard and the one-character survival on today's surface are real, so the boundary rule ships as a fixture rather than as a folded trace.

**RED cells** (each from a committed state, restores grep-verified):

| sabotage | result |
|---|---|
| second `NestedSet[]` on `StructInput` | **RED** — 6 checks, both cycles named; fully green before |
| drop the segment-boundary test | **RED** — boundary fixture only |
| restore last-write-wins | **RED** — two-family fixture only |
| `RefPointer` → `GetValue<string>()` | **RED** — 3 ARM 5 checks; the throw also aborts the pass |
| `RefPointer` → always null (over-reject) | **RED** — 2 ARM 5 checks |

`binding-shim-guard` green at `10 of 30` recursive; `schema-flatten-guard` 30/30; **`ci-all` 133/133**.

Issue #473, opened 2026-09-02
[Bug] codex-umbrella-coverage-guard is one-directional — a router naming deleted tools passes green


Avick3110:
Reported independently by all five reviewers in the #468 PR 1 pre-PR review rounds, including the gate-replica reviewer seeded with nothing but "review this branch".

**What.** `CodexUmbrellaCoverageProbe` asserts one direction only: every REGISTERED tool and bundled skill is referenced in the Codex router (`MissingRefs(umbrella, tools, Allow)`, `CodexUmbrellaCoverageProbe.cs:84,89`). Nothing asserts the reverse — that every tool the router NAMES is a tool that exists.

**Failure scenario, live today.** `plugin/codex/housecarl/SKILL.md` routes every write to a tool that no longer exists. Its step 3, "Pick the narrowest write tool", offers exactly four bullets — `housecarl_set_field`, `housecarl_bulk_apply`, `housecarl_create_record`/`housecarl_bulk_create`, `housecarl_remove_record` — all deleted at #468; `:24` teaches `patch=` semantics off `housecarl_remove_record`; `:51` lists all six as the write surface; and `:46` states "The 1.x write tools below keep working through the 2.0 build". `:39` separately advertises `housecarl_check`, which the wire shows is not registered (#470). The probe passes 12/12 over all of it.

The router is Codex's single hand-maintained entry point, so this is not cosmetic: a Codex user asking to retune a weapon is routed to `housecarl_set_field` and hits the retired-name redirect on every write path.

**Scope note — the router content is deliberately NOT being fixed yet.** Aaron, 2026-09-02: *"I don't want to do anything about the skills until we get to the re-write, remember, I am not releasing any builds until 2.0 is ship ready so this will only affect people cloning the repo temporarily which I am fine with."* W6 rewrites the bundled skills and the router against the shipped surface, and rule 10(f) means no 1.x release ships between now and 2.0.0. This issue is the **guard gap**, not the content: the reason the drift was invisible rather than the drift itself.

[...]

Issue #472, opened 2026-09-02
[GAP] The retired-name redirect sweep derives from a maintained list, not from what 1.9 actually published


Avick3110:
Found by the gate-replica reviewer during the #468 PR 1 pre-PR review rounds. The arm itself is sound and was strengthened on that branch; this is the claim it CANNOT make, now stated in its own comment and needing an oracle to close.

**What.** `BindingShimProbe`'s D3 arm drives the retired-tool-name redirect over stdio and asserts every unregistered retired name answers with a REGISTERED successor. Its subject set is:

    AliasTable.AllRetiredTools  MINUS  the running server's registered set

The right-hand side is derived from the wire. The left-hand side is a **maintained list**. So D3 proves that every retired name the table knows about redirects to a live tool — it does not prove the table is complete.

**Failure scenario.** A later wave deletes a tool and nobody adds its `AliasTable` row. D3 sweeps the rows it has, reports a clean sweep of those, and stays green. A caller on old docs calling the deleted name gets a bare JSON-RPC `Unknown tool: 'housecarl_x'` — the exact dead end the redirect exists to prevent, with the guard that watches the redirect reporting success.

This is CLAUDE.md §5 #11's "a guard's subject set is derived from the surface it claims, never hand-enumerated" applied to a surface that is HISTORICAL rather than live, which is why it needs a fixture instead of reflection.

**Suggested shape** (advisor-proposed, 2026-09-02): capture the **1.9 release's published `tools/list`** once — driven off the shipped 1.9 server, committed as data — and make D3's subject set `(1.9 published set − currently registered set)`, asserting each member has a row whose successor is an exactly-matched registered name. Then a deleted tool with no row is a RED cell by construction rather than an invisible omission. "Shipped and verified" for rule 10(e) already means "published in `tools/list` and driven over the wire" (RUN_ORDER amendment 2026-09-01 (6)), so the fixture is the same oracle written down once.

**Sequencing note.** This wants to ride #470's PR, which already moves the binding-shim census oracle from reflection to `tools/list` — same class of fix, same file, one fixture.

Meanwhile D3's comment on the #468 branch states the limitation plainly rather than claiming a derivation it does not have.

Routing is Aaron's stroke.

Issue #471, opened 2026-09-02
[Bug] CreateOpNaming.Legacy is unreachable but still the default — a future caller silently gets 1.x op spellings


Avick3110:
Found during the #468 PR 1 pre-PR review rounds, in a code path the branch did not touch. Filed rather than folded: this is engine code, wider than the demolition cut, and it is #444's class — "the safe default is what hides the defect".

**What.** `LoadOrderService.CreateOpNaming.Legacy` (`src/housecarl-mcp/LoadOrderService.cs:7372`) spells create-op refusals in the 1.x vocabulary: element `op`, copy subject `CopyFrom / from_plugin`. It is still the DEFAULT in two places:

- `:7128` — `BuildCreateSpec(..., naming ?? CreateOpNaming.Legacy)` in `CreateRecordsBatch`
- `:7099` — hard-wired on the scalar `CreateRecords` path

`housecarl_create` passes the 2.0 naming explicitly (`CreateTools.cs:204`, `new CreateOpNaming("ops", "op=\"CopyFrom\"")`). The 1.x tools that used to reach the Legacy default — `housecarl_create_record` and `housecarl_bulk_create` — were deleted at #468, so no registered tool selects it any more. The scalar `CreateRecords` path is now reached only from a guard.

**Why it matters.** A future tool calling `CreateRecordsBatch` without `naming:` silently emits refusals naming `op[0]` and `from_plugin` — a member and a parameter no surviving surface declares. The caller cannot see them in their own schema and cannot act on them. Nothing goes red: the default is what makes it quiet.

This is the same shape as the `laneAsName` fork that WAS folded on the #468 branch (an unreachable arm left as the default of a flag, measured green under sabotage). That one was deleted because its whole population was inside the deletion; this one is not, so it is filed.

**Suggested shape.** Either make the 2.0 naming the default and require the caller to opt into Legacy (which nothing then does, so it deletes), or delete `Legacy` outright and let the one guard-only caller pass the 2.0 naming. Whichever lands, a guard arm should assert that no reachable path renders `from_plugin` in a create refusal.

**Related residue spotted at the same time**, same file, same cause: `LoadOrderService.RemoveRecords`' three lane-contract refusals (`:5760`, `:5763`, `:5766`) and its `patch is required` arm (`:5774`) are unreachable from `RemoveTools.Remove`, which refuses those shapes first.

Refs #444. Routing is Aaron's stroke.

---

## Update — a second instance of the same class, found in round 2 of the same review

[...]

Issue #470, opened 2026-09-01
[Bug] housecarl_check is declared but never published: CheckTools lacks [McpServerToolType]; tools/list returns 42 of 43


Avick3110:
`housecarl_check` — the merged derived-findings sweep, W3's deliverable — is declared in source but has never been reachable by a caller.

`src/housecarl-mcp/CheckTools.cs:32` carries the tool attribute on the method:

```csharp
[McpServerTool(Name = "housecarl_check", ReadOnly = true, Title = "Sweep the load order for derived findings"), …]
public static string CheckTool(…)
```

…but the enclosing `CheckTools` class carries **no `[McpServerToolType]`**, and `src/housecarl-mcp/Program.cs:166` registers tools with `WithToolsFromAssembly()`, which discovers only types marked with that attribute. The method is therefore never registered.

## Measured, not inferred

Driving the built `housecarl-mcp.exe` over stdio (`initialize` → `notifications/initialized` → `tools/list`) against a fresh empty data dir:

```
PUBLISHED TOOL COUNT: 42
housecarl_check                  published=False
housecarl_records                published=True
housecarl_apply                  published=True
housecarl_create                 published=True
housecarl_remove                 published=True
housecarl_forward                published=True
housecarl_copy                   published=True
```

The count reconciles exactly: **43 tools declared in source, 42 published, the missing one is `check`.** A census of all 24 tool-declaring files in `src/housecarl-mcp` finds `CheckTools.cs` is the **only** one that declares a tool without the type attribute.

## Scope

- Every other 2.0 tool is published. This is one file.
- `check` is not merely written — it is implemented and heavily guarded. `check-guard` is the slowest probe in `ci-all` (62s). Every one of those guards calls `CheckTools.CheckTool` **directly as a C# method**, which is why no existing test noticed: nothing has ever driven `check` over the wire.
- Consequently its published schema, its `findings=` enum binding, and its interaction with `ToolCallShim` are all unproven at the transport layer. Adding the attribute publishes a tool whose wire path has never run.

## Consequences elsewhere

- **RUN_ORDER row 4 records "4b PR #399 merged … W3 COMPLETE" on this tool.** From a caller's side W3 did not complete: nine build sessions and three §4 escalations produced a tool nobody can call.
[...]

PR #474, opened 2026-09-02
Delete the six 1.x write tools their 2.0 replacements absorbed

Avick3110:
The rule-10(e) catch-up demolition, PR 1 of 2. Deletes the six 1.x **write** tools whose 2.0 replacements already shipped and verified, repairs what the deletion orphaned, and derives the completeness claim from the surface instead of a survey.

`Refs #468` — PR 2 (the read family) closes it.

## What is deleted, and on whose ruling

`housecarl_set_field` · `housecarl_bulk_apply` · `housecarl_create_record` · `housecarl_bulk_create` · `housecarl_remove_record` · `housecarl_forward_record`

Their capability is on `housecarl_apply` (both field-edit tools), `housecarl_create` (both create tools), `housecarl_remove` and `housecarl_forward`, where one op or one record is a set of one. Deleted rather than deprecated, per **RUN_ORDER rule 10(e)** — a 1.x tool whose capability the finished 2.0 surface covers goes at 2.0.0 with no deprecation window, in the PR that ships its replacement — as re-shaped by amendments **2026-08-29 (2)** (rolling demolition, and the catch-up PR for W1–W3) and **2026-09-01 (5)/(6)** (two PRs; PR 1 write-only).

Calling a deleted name is not a dead end: the server answers with the successor and that tool's parameter migration. Nothing shipped to users changes — rule 10(f) means the 1.9 zip is the last 1.x release and the next bump is 2.0.0.

## The check family is HELD

`housecarl_check` is declared but has **never been published** — `CheckTools` lacks `[McpServerToolType]`, so `WithToolsFromAssembly()` never registers it and `tools/list` returns 42 of 43 declared tools (**#470**). Deleting the three check tools would strand their callers on a successor no client can reach, so that commit is held on **`claude/check-family-held` @ `20b44b2`** until #470 lands.

## Review rounds

Conducted by a fresh session per RUN_ORDER boot rule 9; `ci-all` re-run independently at the branch head before any reviewer saw the tree (133/133).

**Round 1 — five reviewers** (removal residue · verification integrity · caller-facing sentences · transport/packaging · plus the **gate-replica trial reviewer**, Opus/high, seeded with nothing but "review this branch"). Twelve findings confirmed at source, nine folded. Three reviewers independently sabotaged production behaviour and showed a guard staying green.

**Round 2 — three reviewers**, same seed, told nothing of round 1. It returned the **same failure class inside the fold** — a class-stop under CLAUDE.md §5 #11, so the remaining findings are listed open below rather than folded a third time. No round 3.

**Refused, with ground:**

- **The shipped skills and Codex router** (all five round-1 reviewers, again in round 2). Ruled by Aaron: *"I don't want to do anything about the skills until we get to the re-write, remember, I am not releasing any builds until 2.0 is ship ready so this will only affect people cloning the repo temporarily which I am fine with."* The **guard gap** behind it is filed as **#473** — `CodexUmbrellaCoverageProbe` asserts only *registered ⊆ router*, never the reverse, which is why a router naming six dead tools passes green.
- **The changelog "overclaim"** (2 reviewers) — a true consequence of the `WriteEngine` defect, not a separate one. Repairing the refusal made the bullet true.
- **Comments that name a deleted tool accurately as history** — `ApplyTools`/`CreateTools`/`RemoveTools`/`ForwardTools`' "Absorbs `housecarl_X`", `LoadOrderService`'s alias-machinery note. The line held throughout: **false-about-now gets fixed, historically-accurate stays.**

## Open at the gate — three classes, one root

**The root: the population was hand-noticed, never derived.** That is CLAUDE.md §3's hand-wired-coverage failure applied to a deletion's residue instead of to a record set. Aaron's caveat on the class-stop ruling, recorded because it is the right worry:

> "however I have a suspicion this is also probably not the smarter better overall solution, it sounds like a class that can continue to bite"

This PR is the branch's **disposition**, not the cure. The cure is the executing-arm rule ruled in RUN_ORDER amendment **2026-09-02 (2)** — *a remedy that names a call ships with an arm that MAKES the call as the sentence tells the caller to, against a fixture, asserting it is not refused* — plus the derivations below. The open findings are the first executing arms of that chartered work, not a permanent open list. A review of the whole decommissioning approach follows this PR, at Aaron's direction.

### Class A — assertions pinning a proxy for the property they claim

The property is always "a caller who follows this is not refused"; the proxy is always "the text contains these tokens".

| # | Where | Proxy pinned | Property claimed | What an executing arm would drive | Status |
|---|---|---|---|---|---|
| 1 | `BulkCreateGuardProbe.cs:133` | `Contains("parent")` — satisfied by prose sitting *before* the repaired clause | the refusal states the shape `housecarl_create` accepts | call `create` with what the sentence says to pass | **open** |
| 2 | `DescriptionVocabularyGuardProbe.cs:368` | tolerance set `declared − registered`, derived from the failure it tolerates, so it widens itself | only #470's instance is excused | fail on any *new* member | **open** (needs #472) |
| 3 | `BindingShimProbe.cs:199` | `registered.Any(...)` — any registered name; the row's own `Successor` never consulted | the redirect sends the caller to *this row's* successor | assert every tool the row names is registered | **folded, RED-checked** |
| 4 | `DialogueCkParityGuardProbe.cs:417` | required member *names* appear | the rendered `ops` element is followable | drive it through `apply` | **open** |
| 5 | `DialogueCkParityGuardProbe.cs:411` | required set regexed out of `[Description]` prose | what `ApplyOp`/`ListParams.Strict` enforce | derive from the enforcing reader | **open** |

### Class B — 1.x residue unreachable but still the DEFAULT

The next caller who omits a flag silently selects the deleted tool's behaviour. Each measured by rewriting the default to garbage and getting `ci-all` 133/133.

| Instance | Disposition |
|---|---|
| `InPlaceAgainHint`'s `laneAsName` fork | **folded — deleted, not guarded.** An arm that cannot fail is not made safe by watching it |
| `CreateOpNaming.Legacy`, `sourceParam = "from_plugin"` (×4), the `origins` default | **filed #471**, engine code, outside the cut |

**Derivable population:** *every parameter whose default encodes a 1.x spelling and whose every production caller overrides it.* #471 is filed as the class, not the instances.

### Class C — statements false about the current surface

Found by four reviewers across both rounds. Every instance a fold in this branch passed through is fixed (the shim's "nine tools" count, `AliasTable`'s census note and its retired-name summary, two renderer docstrings, `RemoveTools`' two-caller note, `description-vocab-guard`'s "no sentence is exempt" line, and two comments naming the retired `full_readback` spelling as the path's name). **Six remain open:** `AliasTable.cs:141`, `:90`, `ApplyTools.cs:345`, `CreateTools.cs:216`, `ForwardTools.cs:134`, `LoadOrderService.cs:5811` — each asserts, in the present tense, a two-caller or six-tool world that ended at `c6c53cd`.

**Derivable population:** *every comment asserting something about the retired set*, held against the live table and the live registered set — so "which today is nothing" cannot survive six rows going live.

### Class D — the cut orphaned caller-facing PARAMETER spellings — **folded**

The sentence-repair population was derived from deleted **tool names**; the same cut orphaned **parameter and op-member** spellings, invisible to that grep and to the `TOOLNAME` arm (which matches `housecarl_` tokens only). Six live sentences instructed callers in spellings nothing declares. The sharpest, `LoadOrderService:7504`, was correct for two of its three callers before this branch and for **none** after — a caller adding `from_plugin` lands on a second refusal. Merging that is Q3's silently-degraded mode, so it folded rather than going open.

**Population line: not statically derivable** — measured while trying. 133 declared members across 45 registered tools read off the wire; 7,750 prose literals scanned; 42 undeclared spellings at 133 sites, dominated by fields the *response reports* (`epoch=` alone is 44); an instruction-shaped filter leaves exactly **one** clean member. The other cases do not fall out because:

- `from_plugin` is genuinely undeclared, but its sentences read "CopyFrom **requires** `from_plugin`" — no instruction-verb filter catches that without swallowing prose;
- **`verb` IS declared** — the nested `compose=` set legitimately spells it — yet `WriteEngine:1935`'s "use `verb=Remove`" is still wrong, because that sentence sits at the **op** altitude where the member is `op`. The same token is right at one altitude and wrong at another, inside one tool.

So the correct oracle is not "is this spelling declared anywhere" but "is it declared by the tool **this sentence is reachable from**" — a per-sentence reachability question an executing arm answers by construction and no containment net can. That is why the derivation belongs to the chartered work and this PR claims six sentences, not a sweep.

## Verification

`ci-all` **133/133** at every commit point. Mutation sweeps run from a committed state, restores byte-exact and `git status`-verified:

- round-1 fold — **10 cells, all RED**, including a canary proven GREEN *before* the fold;
- D3's successor fold — **4 cells**: control GREEN, dead successor RED, two successors one dead RED, **two successors both live GREEN** (the cell proving it does not redden on any second name);
- class D — the four guard assertions that moved went **RED on the old text** first; each corrected spelling then driven once over the wire by hand (`readback` top-level, `from_source`/`op` as op members; none of `full_readback`/`from_plugin`/`verb` declared anywhere on `housecarl_apply`; a call carrying the corrected spellings reaches the configuration prompt, not a binding refusal). By hand because the arm that would do it does not exist yet.

Issues filed from the rounds, unmilestoned — routing is Aaron's stroke: **#471**, **#472**, **#473**.

The `## Unreleased` changelog entries are in `a49226b`.

---


src/housecarl-mcp/WriteTools.cs
Avick3110, line 1: **[low] The class summary (lines 10-16) is now false about every tool left in this file.**

> houseCARL write tools (§8.4 Beat C). **Both ride** the PROVEN public write cleave (`WritePatchBuilder.Apply`) through `LoadOrderService.ApplyEdits` … **`into=` extends an existing patch**

After `c6c53cd` this type declares only `housecarl_create_plugin`, `housecarl_compact_plugin` and `housecarl_merge_plugins`. None of them call `ApplyEdits`, none resolve a record's load-order winner, none pre-flight through the corpus rulebook, and none declare `into=` — merge is explicitly new-file-only, compact's second lane is `in_place=`+`acknowledge=`, and create_plugin writes a header with zero records.

This is the Class C shape the PR body enumerates ("each asserts, in the present tense, a … world that ended at `c6c53cd`"), sitting in the file the cut gutted, and it is not on the six-item open list.

src/housecarl-generator/BindingShimProbe.cs
Avick3110, line 1: **[low] The census narrative at line 469 contradicts the `CensusExpected` array this commit edited.**

> `*` the five create-operand hints (record_type/editorid/parent/collection/grid) fire on BOTH housecarl_create and 1.x housecarl_bulk_create

This same commit deleted all five `"housecarl_bulk_create: … => hint"` rows from `CensusExpected`, and `AliasTable.cs`'s twin comment *was* corrected for exactly this ("the demolition catch-up (#468) deleted that tool and those five census rows with it"). One half of the pair moved, the other did not.

The header of this array instructs "On any schema or table change: eyeball the printed diff, then update `CensusExpected`" — so the note a future maintainer reads while doing that is the one that is wrong. There is also no `#468` line in the running narrative recording the `163 → 143` drop, unlike every prior wave.

src/housecarl-core/WriteVerbs.cs
Avick3110, line 1: **[low] `AllRecital`'s doc (lines 99-127) still counts a description that no longer ships.**

> Three shipped descriptions concatenate it today … and the `BulkOp.verb` one.

and

> shipping a false claim in `housecarl_bulk_apply`'s schema

`housecarl_bulk_apply` is deleted, and this same PR added `BulkOp` to `WireNamesProbe.NonInputWireTypes` on the ground that it is "reachable from no tool's input schema". So `BulkOp.verb`'s `[Description]` — and the tail gloss the hazard paragraph is entirely about — reaches no caller. Two shipped sites now, not three, and the hazard's consequence clause names a surface that ended.

This file was edited by the branch (lines 23 and 228), so it is a stale claim a fold passed through rather than one outside the cut.

src/housecarl-generator/InPlaceProbe.cs
```diff
@@ -1265,4 +1270,28 @@ static List<string> ReadMasters(string path)
     }
 
     static string Trim(string? s) => (s ?? "(null)").Replace("\r", " ").Replace("\n", " ").Trim();
+
+    /// <summary>Every REGISTERED tool declaring an <c>in_place</c> parameter, reflected off the shipped
```
Avick3110: **[low] The summary says REGISTERED; the predicate derives DECLARED.**

`InPlaceDeclaringTools()` reads `[McpServerTool]` on the *method* only. It omits the `[McpServerToolType]` check on the declaring type — the check that `RegisteredToolNames()`, added in this same PR, documents as load-bearing:

> The type half is load-bearing and is why this is not the method-attribute-only derivation used elsewhere in the generator — `CheckTools` carries the method attribute and no type attribute … and a predicate reading the method alone reports it live.

Concretely: a tool method declaring `in_place` on a type missing `[McpServerToolType]` — the configuration `housecarl_check` is in today (#470) — is swept by arm H, printed in its `swept N:` detail, and passed under a label reading "opt-in by construction, every in_place-declaring tool" even though the SDK never registers it. The superset makes it safe rather than unsafe, but two derivations written in one PR disagree about the same predicate and only one of them says so.

src/housecarl-generator/BindingShimProbe.cs
```diff
@@ -35,6 +40,18 @@ namespace HousecarlGenerator;
 public static class BindingShimProbe
 {
     const string GenericError = "An error occurred invoking";          // the SDK's opaque text (measured live, HCBR-2026-06-11-01)
+
+    /// <summary>One-line clamp of a server response, for an arm's failure detail.</summary>
```
Avick3110: **[low] Two `<summary>` elements on one member; `Flatten` ends up undocumented.**

"One-line clamp of a server response, for an arm's failure detail" describes `Flatten` (declared at line 49), but it is stacked above `ToolToken`'s own summary, so `ToolToken` carries two `<summary>` elements and `Flatten` has none. Malformed doc XML (CS1710 if `/doc` is ever enabled), and a reader of `ToolToken` gets a sentence about response clamping that has nothing to do with a regex over successor teaching.

Same slip at `DescriptionVocabularyGuardProbe.cs:2505`.

src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
```diff
@@ -2457,6 +2478,113 @@ over two lines
...
+            foreach (var m in t.GetMethods(Flags))
+                if (m.GetCustomAttribute<ModelContextProtocol.Server.McpServerToolAttribute>(inherit: false)?.Name is { Length: > 0 } n)
+                    names.Add(n);
+        }
+        return names;
+    }
+
+    static readonly Regex ToolToken = new("housecarl_[a-z0-9_]+", RegexOptions.Compiled);
+
+    /// <summary>A remedy may only name a call that works (<see cref="CorpusRulebook"/>'s own rule), and a deletion
```
Avick3110: **[low] Two `<summary>` elements on `DeclaredButUnregisteredToolNames`.**

The TOOLNAME arm's rationale ("A remedy may only name a call that works … So the check is derived on both sides") is stacked directly on top of `DeclaredButUnregisteredToolNames`'s own summary. That member now has two `<summary>` elements, and the rationale explaining why the whole arm exists is filed under the helper that computes its *exemption* set — the one part of the arm the rationale is explicitly not about. It reads like it belongs on `ToolNameArm` (line 2531).

Same slip at `BindingShimProbe.cs:44`.

Avick3110:
## Gate folds — all six, `550b6ea` → `9e986d8`

Fresh fold session (CLAUDE.md §5 #11: directed folds go to a session that did not conduct the rounds). No new review rounds. `ci-all` **133/133** from the committed state before the first edit and from the committed tip after the last; tree clean at both points.

| # | Finding | Commit | What changed |
|---|---|---|---|
| 1 | `WriteTools.cs:10-16` — class summary false about every tool left in the file | `a784e17` | Rewritten to the three tools the type still declares: each rides its own whole-plugin builder (`WritePatchBuilder.CreatePlugin` / `CompactBuild` / `MergeBuild`), and the four things none of them does — no `ApplyEdits`, no winner resolution, no rulebook pre-flight, no `into=`. Output lanes stated per tool because they differ (compact's `in_place=` lane overwrites; the other two never touch an existing file). Second paragraph names where the cleave's tools actually live and why the render helpers keep this file alive. |
| 2 | `BindingShimProbe.cs:469` — census narrative contradicts the array this commit edited | `10f7844` | The create-operand bullet records the both-tools activation as history and says the five hints fire on `housecarl_create` alone. Adds the missing `#468` narrative line — see the number note below. |
| 3 | `WriteVerbs.cs:99-127` — counts a description that no longer ships | `9e986d8` | Two shipped sites (`ApplyTools`' `op=` prose, `ApplyOp.op`), with `BulkOp.verb`'s concatenation named as vestigial and why. The hazard paragraph keeps the mechanism and states that since #468 it has **no shipped consequence** — the only glued site is a description no caller is served; what remains is latent against the `BulkOp` → `ApplyOp` collapse `NonInputWireTypes` already names. **Five sites, not one:** the same fact had four more homes in `description-vocab-guard` (`PublishedDefault`'s summary, INV4-MARK's failure message, `TailGlossVerb`'s summary, `TailGlossArm`'s summary) plus that arm's runtime failure line. Fixing `WriteVerbs` and leaving the twins is the shape finding 2 is about. |
| 4 | `InPlaceProbe.cs:1274` — summary says REGISTERED, predicate derives DECLARED | `e2c6b7f` | Unified, not relabelled. New `RegisteredTools.cs` holds the SDK's predicate once (type attribute AND method attribute) and returns `(name, method)` pairs; arm H filters them for `in_place`, `RegisteredToolNames()` takes the names. |
| 5 | `BindingShimProbe.cs:44` — two `<summary>` on `ToolToken`, `Flatten` undocumented | `5a3b174` | Clamp sentence moved onto `Flatten`; text unchanged. |
| 6 | `DescriptionVocabularyGuardProbe.cs:2505` — same slip | `5a3b174` | The TOOLNAME rationale moved onto `ToolNameArm`; `DeclaredButUnregisteredToolNames` keeps its own summary. Text unchanged. |

### Arm H's count: 6 → 6, and why the predicted drop did not happen

The fold was expected to drop arm H's `swept N:` by exactly one, `housecarl_check` leaving the sweep. **It did not move: 6 before, 6 after** — `apply`, `compact_plugin`, `create`, `forward`, `nif_set`, `remove`.

`CheckTools` declares no `in_place` parameter (`grep in_place src/housecarl-mcp/CheckTools.cs` is empty), so `housecarl_check` was never in this sweep. The method-only predicate was a **latent** superset, not a live one: the next `in_place` tool landing on a type without `[McpServerToolType]` would have been swept and passed under a label reading "every in_place-declaring tool". Both summaries now say that rather than claiming a live instance — the finding's substance holds, only its "concretely" example doesn't instantiate today.

RED-checked from the committed state: with `[McpServerToolType]` removed from `WriteTools`, arm H sweeps **5** — `compact_plugin` leaves. Under the old method-only predicate that sabotage moved nothing. Restore verified by grep (no sabotage string, attribute back, the fold's own string present, `git status` clean).

### Finding 2's numbers: the narrative's running total was already stale

The finding asks for a `163 → 143` line. Measured on this branch, the drop is **`169 → 144`** — 25 rows deleted, none added (`set_field` 4, `bulk_apply` 3, `create_record` 3, `bulk_create` 7, `remove_record` 5, `forward_record` 3), and the census guard prints `144 activations`.

`163` is the last running total anyone wrote (PR #311 review 3). Six rows landed after it — #312's `write_seq` `output_dir` pair, the `records` tool's — with no narrative line of their own, so the total went stale before this branch. The new line records the measured numbers and says so; it does not reconstruct the entries nobody wrote.

### Verification

- `ci-all` **133/133** at `550b6ea` (baseline) and at `9e986d8` (tip).
- Per-probe after the folds that touch guards: `binding-shim-guard` CENSUS PASS at 144 activations; `description-vocab-guard` 68/68 with INV4-TAILGLOSS PASS at 1 glued of the 3 descriptions carrying the recital; `inplace-guard` arm H PASS, swept 6.
- One RED-check (finding 4, above). No pins added — findings 1/2/3/5/6 are statements, and 4 is a predicate with one home.
- Every commit staged by path and its diff read before committing; nothing near `claude/check-family-held` or `CheckTools.cs`'s attribute (#470).

### One thing for the merge decision

Finding 3 leaves `INV4-TAILGLOSS` pinning a description no caller is served. It is kept, not deleted, because the `BulkOp` → `ApplyOp` collapse carries that gloss onto a shipped description — but "an arm guarding nothing shipped" is a delete-or-re-point question for that change, not something a doc rewrite settles. Flagged here rather than decided.

PR #476, opened 2026-09-02
Publish housecarl_check, and the test project is born

### LINKED ISSUE #470: [Bug] housecarl_check is declared but never published: CheckTools lacks [McpServerToolType]; tools/list returns 42 of 43

### LINKED ISSUE #472: [GAP] The retired-name redirect sweep derives from a maintained list, not from what 1.9 actually published

Avick3110:
Fixes #470
Closes #472

`housecarl_check` was merged on 2026-08-21 and no client could ever call it. `CheckTools` carried `[McpServerTool]` on its method but no `[McpServerToolType]` on the class, and `WithToolsFromAssembly()` discovers only marked types — so the tool was declared, guarded, documented and unreachable for twelve days. Every existing check guard called `CheckTools.CheckTool` as a C# method, which is exactly why none of them noticed.

The product fix is one attribute. The rest of this PR is the oracle that makes "shipped" mean something a caller can verify, and the test project that houses it.

## What lands

- **`housecarl_check` is published**, and driven over stdio rather than called in-process.
- **`src/housecarl-mcp-tests` is born** — a standard xUnit project under `dotnet test`, with `RecordsGuardProbe` (1,189 lines) converted into it as the first family and deleted in the same commit.
- **A tool-surface census from both sides**: what the assembly declares, reflected; what `tools/list` publishes, off the running server. Asserted as set equality, not a count — a count is right or wrong for the wrong reasons and moves every wave.
- **The retired-name redirect swept against what 1.9 actually published**, replacing a maintained list (#472).
- **A residue countdown** that gates the old harness against growth, and **a bridge** so `dotnet test` runs `ci-all` too.
- **ADR 0003**, the public home for the two-harness decision.

## Green, measured from the committed tip (re-measured after the gate folds)

| gate | result |
|---|---|
| `dotnet build housecarl.sln -c Release` | 0 errors |
| `ci-all` | **132/132**, exit 0 |
| `dotnet test src/housecarl-mcp-tests` (all tiers) | **323/323** |
| tiers | unit 66 · integration 182 · stdio 74 · bridge 1 |

Read off the summary lines. One `ci-all` run in this session printed no summary and looked red because a `grep` in the pipeline returned its own exit code while a background run held the log; re-running it cleanly gave exit 0 and `132/132`. Flagging it because the same misread happened once during the build session, and the lesson is the same both times: the summary line is what you read.

## The residue countdown, with its derivation

Three numbers gate with exact equality in **both** directions — growth fails, and an unrecorded shrink fails too, because a baseline above the real figure stops being a countdown and becomes headroom.

| measure | value | derived from | gates? |
|---|---|---|---|
| `probeFiles` | 155 | `*Probe*.cs` under `src/housecarl-generator`, recursive, bin/obj excluded | yes |
| `ciAllRows` | 132 | `CiAll.ProbeNames.Count` — the dispatch registry off the compiled generator | yes |
| `guardFilesOutsideTheCount` | 1 | registry guards whose host file `probeFiles` does not count, via `CiAll.ProbeTypes` | yes |
| `probeLines` | 65,245 | total lines of those probe files | **no — reports only** |

`ciAllRows` independently equals `ci-all`'s own `132/132` summary, which is the cross-check the baseline file claims is available.

`probeLines` reports rather than gates on Aaron's ruling. It moves in both directions on ordinary in-place edits to an existing guard — which the two-harness rule *requires* — so gating it punished the correct act. It moved twice in this PR alone and nothing failed, which is the ruling working.

`guardFilesOutsideTheCount` is 1, and it is `WriteEngine`: `ci-all` runs two guard verbs out of `src/housecarl-core/WriteEngine.cs`. Its 4,079 lines are product code hosting two guards, so folding it into `probeFiles`/`probeLines` would inflate the countdown with code that is not residue. It gets its own gated measure instead, reported by name, and W7 is not complete while it is above zero.

**Named as outside the counter by definition:** six files under `src/housecarl-generator` expose a `Run` verb without matching the glob — `WriteOracle`, `WriteCensus`, `CopyDifferentialHarness`, `SkyPatcherHarness`, `LocalizedShapeSweep`, `ClassParentsEmitter`. They are manual/exploratory verbs, not CI guards, and none is a registry row. Their fate belongs to the row-10 walk, not here; the exclusion is stated so it is visible rather than silent.

## The 1.9 oracle's provenance

`src/housecarl-mcp-tests/data/tools-list-1.9.0.json` is a frozen wire capture, not a reconstruction. **Re-derived independently by this review session**, not taken on trust from the build session: the `v1.9.0` release asset was downloaded again (17,647,713 bytes, sha256 `56349058cdf6a090…` — matching the file's recorded hash exactly), the shipped `housecarl-mcp.exe` inside it was driven over stdio against a fresh empty `HOUSECARL_DATA_DIR`, and the result was an **identical 45-name set** with `serverInfo` `{"name":"houseCARL","version":"1.9.0"}`. Two round-1 reviewers listed this as the one thing they could not verify offline; it is now measured. 1.9.0 is the last 1.x release, so the capture is final.

## Review rounds

Conducted by a fresh session that did not write this code, per the pre-PR rounds rule.

**Round 1 — five reviewers** (four seeded with the settled-decisions list, plus one gate-replica seat seeded with nothing but "review this branch"). **Round 2 — two reviewers**, per Aaron's directive mid-session that rounds use two from then on; it was applied to round 2 immediately rather than deferred, by stopping a third that had just been spawned. All reviewers were Opus, each in its own worktree, none ever seeing the tree these commits were made from.

**Round 1 → six fold commits.**

| folded | why | evidence |
|---|---|---|
| `ciAllRows` reads the registry, not a regex over `CiAll.cs` | a row with a trailing comment counted as **zero** while `ci-all` ran it | reproduced, then RED after the fold |
| CI floor reading the run's own trx; countdown prints the measurement | `dotnet test` **exits 0 when its filter matches nothing** | measured both directions; also gave `--logger trx` a reader |
| six false-about-now sentences in five generator files | publishing the tool made them false; none was updated | population swept independently — found one nobody reported |
| shim's coerced **value** asserted; all 12 published parameters driven | a shim wrapping a bare string into an **empty array** passed the entire suite — a silent wrong answer | that sabotage now fails two tests by name while the wire arms stay green |
| duplicate-filename tests get a private world and split in two | they mutated the shared world; the one merged pair against one-test-per-assertion | suite green, hazard removed |
| bridge cannot hang on a pipe; two arms fixed | `.Result` after `WaitForExit` could block forever on an inherited handle | cleared the two analyzer warnings pointing at it |

**Refused in round 1, with reasons:** the trx-artifact upload (ci.yml packaging is row 8's lane — the flag was given a reader instead of being removed); two ported cannot-fail remedy arms and the `xUnit2030` warning sitting on one (verbatim ports of pre-existing probe arms, #426's chartered class — changing them alters what they assert without fixing the class); a cosmetic trait-placement nit.

**Round 2 → class stop.** Both reviewers independently found the same defect **in the arm round 1 added to fix round 1's version of it**: it derived "guard entry point" from a method named `RunGuard`, and its docstring claimed that was the CI guard shape. False — nine of the 132 registry rows dispatch something else, and two of those live in `src/housecarl-core`, where no naming discipline inside the generator could ever have seen them. The countdown could therefore have reached zero, W7's closing condition, with `ci-all` still running two guards.

Three shapes had now been tried for that one population — a filename glob, a regex over source text, a method name — each a proxy, each short, the third written by a session that had just been told about the first two. Under the rounds rule that is a design signal, not a fix queue, so **no third instance was folded**; it went to Aaron as a §4(b) escalation. His ruling: fix the live half here, charter the rest. That fold is the last code commit — the population now comes off `CiAll.ProbeTypes`, the dispatch delegates themselves, which names `WriteEngine` on the first run and crosses assemblies for free.

**Chartered, not fixed here** (per that ruling): guards wired as their own `ci.yml` step outside the registry — the shape `freshness-capture-guard` already uses — and the census's hand-scoped assembly, which is latent rather than live because `housecarl-core` carries no MCP SDK reference, so a marked type cannot exist there without a csproj change.

**Every fold was mutation-tested RED from a committed state**, each sweep carrying a known-RED canary proven by hand first and a green control cell; scripts are kept with the branch's working notes. Two disclosures. First: on the shim fold one hand sabotage was run against a **dirty tree**, which the rules forbid — it ended clean, but the round-trip silently added a UTF-8 BOM that only reading the diff caught, and that sweep was re-run properly from a committed state afterwards. Second: one sabotage cell was mis-designed (moving a file broke the build, so the run went red for a compile error rather than the arm) — the cell's needle check caught it, and it was redesigned as a rename, which is why cells carry needles.

## The ADR question

Answered by landing one. Two reviewers, in separate rounds, argued this PR makes an architectural decision — a second harness, a declared window with two required CI steps, a checked-in baseline that gates — and that `standards/HOUSECARL_DOC_HYGIENE.md` §8 requires the ADR in the same PR. They also named the sharper half: every place the new machinery explained itself cited a document that exists only in the private working corpus, which is precisely the failure §8 describes. **ADR 0003** is now the public home, and the citations in `ci.yml`, both harness test classes and the baseline file point at it instead of at private amendment numbers.

The ADR records what was already decided; it does not decide anything new. It also states the part of the evidence that came out *against* the original plan: converting one family grew 1,189 lines into 1,540, so the hoped-for shrink did not happen, and the case rests on mutation testing, named failures, isolation and derived grids instead.

## Gate folds — Aaron's five review findings

A fresh fold session (not the author, not the rounds session) executed these. All five were triaged as
confirmed; none was refused. Each is one commit, RED-checked from the committed state after its commit, with
a known-RED canary proven by hand first and a green control cell in every sweep. **No new review rounds** —
fresh eyes are for before the PR opens.

| # | finding | commit | RED evidence |
|---|---|---|---|
| 1 | RC 3918228003 — `ServerFixture.Rpc` orphans a `ReadLine` on timeout, and the orphan eats the next test's response | `b032cc2` | canary (the pre-fold shape, hand-proven): abandoned read takes `"A"`, next caller handed `"B"`. Post-fold sabotage: pump made to eat the late line ⇒ RED `Expected "A" / Actual "B"`; poison's early throw disabled ⇒ RED, no exception thrown. Control GREEN |
| 2 | RC 3918230909 — `RecordsWorld` never restores `CorpusRulebook.CorpusPath`, and `Dispose` deletes the directory it names | `fb26ede` | canary (hand-proven): restore removed ⇒ RED, the static still names the deleted world. Second cell: capture moved after the repoint ⇒ RED, "restores" the world's own dead path. Control GREEN |
| 3 | RC 3918233575 — the host file was still derived from the type NAME, a fourth proxy | `fdbae68` | canary (hand-proven): a `housecarl-core` guard type named after an existing generator probe file ⇒ RED reporting `ArtifactGuardProbe (src/housecarl-core/ArtifactGuardProbe.cs)`, where the old shape was GREEN. Lambda row ⇒ RED with *"register a method group, not a lambda"*. Second guard type declared inside an existing `*Probe*.cs` ⇒ GREEN (its file is counted; the old shape called it outside). Control GREEN |
| 4 | RC 3918235910 — the residue-countdown step had no empty-run guard | `bfb6b13` | the committed check text extracted out of `ci.yml` and run against both states: filter matching nothing ⇒ `dotnet test` exit 0, `residue.trx` `executed="0"`, check throws naming the filter; real filter ⇒ `executed="5"`, check passes and the console logger still prints all three countdown lines. Missing-trx arm throws too |
| 5 | RC 3918239449 — the text-lane arm still sliced before checking its prefix | `6da4262` | measured on the pre-fold code with `RefusalText()` returning empty: `System.ArgumentOutOfRangeException` from the range operator. Same sabotage after the fold: `Assert.StartsWith() Failure: String start does not match`. Control GREEN |

Fold 1's own RED check found a false sentence in a comment fold 1 had just written — with the poison removed
the second call *succeeds*, because the pump queues the late response and the id check skips it, so the arm's
elapsed bound is a floor on "immediately" rather than the proof. Corrected inside fold 1's commit, and the
test now says which half discriminates and which cascade is not honestly fixturable here.

**The residue baseline is unchanged** — `probeFiles` 155, `ciAllRows` 132, `guardFilesOutsideTheCount` 1,
`probeLines` 65,245, all still measuring exactly what they did before. Fold 3 changed how the host file is
derived, not what the count is; `guardFilesOutsideTheCount` is still 1 and still `WriteEngine`, now reported
by path as `WriteEngine (src/housecarl-core/WriteEngine.cs)`.

## Not in this PR

The check-family deletion rides the cut. No 1.x tool was deleted, no deletion-flagged tool body was touched, no skill was touched. #426 was appended to, not closed — its `fields_source` finding needs a fixture where a scoped scan's field values differ from the winner's, which is a scenario to design.

`plugin/CHANGELOG.md` carries the user-facing line. It deliberately says nothing about the test project: that file's own rule is to state what a tool now does or refuses, and an internal guard harness is not that.


src/housecarl-mcp-tests/ServerFixture.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        int id = Interlocked.Increment(ref _id);
+        Send(new { jsonrpc = "2.0", id, method, @params });
+
+        var deadline = DateTime.UtcNow + TimeSpan.FromSeconds(60);
+        while (DateTime.UtcNow < deadline)
+        {
+            var read = Task.Run(_out.ReadLine);
+            var remaining = deadline - DateTime.UtcNow;      // ONE shared budget; a per-line wait never extends it
+            if (remaining <= TimeSpan.Zero || !read.Wait(remaining) || read.Result is not { } line) break;
```
Avick3110: **A timed-out `Rpc` orphans its `ReadLine` task, and the orphan eats the next test's response.**

`read` is started, and on either bail-out (`remaining <= 0`, or `read.Wait(remaining)` returning false) the loop `break`s while that `Task.Run(_out.ReadLine)` is still pending on the shared stream. Nothing cancels it — `StreamReader.ReadLine` is not cancellable — so it stays parked and consumes the next line that arrives.

Concrete cascade: `ServerFixture` is an `ICollectionFixture` shared by all 73 stdio tests. One call exceeds the 60s deadline (slow runner, or a genuinely hung body). That test throws `TimeoutException`. The late response then lands in the orphan and is discarded. The *next* test sends its request and starts its own read — but its response was never the line the orphan took, so the accounting is now one line behind: there is no unread line for it, it waits the full 60s, throws, and orphans another reader. Every remaining stdio test fails at 60s each — roughly 70 minutes of CI, all reporting `no response to … within 60s`, with the original slow call indistinguishable from the wreckage it caused.

The cheap fix is to make the timeout terminal for the fixture rather than for one test: set a poisoned flag when a wait fails and have every later `Rpc` throw it immediately (naming the first timeout), so the run fails fast and names the real cause. A single background pump thread draining `_out` into a `BlockingCollection` removes the orphan class entirely.

src/housecarl-mcp-tests/RecordsWorld.cs
```diff
@@ -0,0 +1,202 @@
...
+        OldFile = Path.Combine(ModsDir, "OldMod", OldName);
+        var midFile = Path.Combine(ModsDir, "MidMod", MidName);
+        master.BeginWrite.ToPath(masterFile).WithLoadOrder(Array.Empty<ISkyrimModGetter>()).Write();
+        midMod.BeginWrite.ToPath(midFile).WithLoadOrder(new ISkyrimModGetter[] { master }).Write();
+        ovMod.BeginWrite.ToPath(OverrideFile).WithLoadOrder(new ISkyrimModGetter[] { master }).Write();
+        oldMod.BeginWrite.ToPath(OldFile).WithLoadOrder(new ISkyrimModGetter[] { master }).Write();
+
+        var genDir = Path.Combine(Root, "corpus-gen");
+        CorpusGenerator.GenerateAll(genDir, Path.Combine(Root, "corpus-ref"));
+        CorpusRulebook.CorpusPath = Path.Combine(genDir, "corpus.json");
```
Avick3110, line 149: **`CorpusRulebook.CorpusPath` is set here and never restored, and `Dispose` deletes the directory it points at.**

`AssemblyInfo.cs` names this static as the reason parallelisation is off, and the summary above says a mutating test "constructs its own instance" — but a private instance repoints the process-global on the way in and, in `Dispose` (line 170), deletes `Root` while the static still names `Root/corpus-gen/corpus.json`. After the last private world disposes, the rulebook points at a path that no longer exists.

This is the protection the retired harness carried and the conversion dropped. `CompactServiceGuardProbe.cs:54` captures `priorCorpusPath` and `:809` restores it in a `finally`, with the comment stating exactly this hazard — *"CorpusPath is process-global, and leaving it pointing into a deleted temp dir would break whichever probe runs next"* — and `CiAll.cs:507` resets it to the canonical corpus before every probe.

Why it does not bite today: `RecordsTransportTests` and `RecordsDuplicateFilenameTests` are their own collections, and every test that resolves `type=` lives in the `records` collection, whose fixture resets the static when it is constructed. That is an accident of scheduling, not an invariant. Add one `[Fact]` outside the `records` collection that calls `RecordsTools.Records(..., types: new[]{"WEAP"})` and whether it passes depends on which collection xUnit ran first.

Restoring the prior value in `Dispose` (capture in the ctor, restore in `Dispose`) costs two lines and makes the property hold by construction rather than by collection layout.

src/housecarl-mcp-tests/HarnessResidueTests.cs
```diff
@@ -0,0 +1,239 @@
...
+    // two guard verbs, and counting it as harness residue would inflate the countdown with code that is not
+    // residue. It gets its own gated measure instead, so it is visible, cannot grow silently, and has to reach
+    // zero along with the other two.
+
+    static string[] GuardFilesOutsideTheCount()
+    {
+        var counted = ProbeFiles().Select(Path.GetFileName).ToHashSet(StringComparer.OrdinalIgnoreCase);
+        return HousecarlGenerator.CiAll.ProbeTypes
+            .Select(t => t.Name)
+            .Where(n => !counted.Contains(n + ".cs"))
```
Avick3110: **The population now comes off the registry (right), but the *host file* is still derived from the type NAME — a fourth proxy.**

`CiAll.ProbeTypes` answers "which types host guards" honestly. `.Where(n => !counted.Contains(n + ".cs"))` then answers "which file does this type live in" by assuming the file is named after the type. That is a naming convention, which is the shape the escalation was about.

Two concrete outcomes.

**False RED.** Register a row as a lambda rather than a method group — `("x-guard", () => Foo.Run())`. `p.Run.Method.DeclaringType` is then `CiAll`, or a compiler-generated `<>c`. `EveryRegistryGuardResolvesToASourceFile` fails naming `<>c`, which points at nothing a reader can open, and this measure reports 2 against a baseline of 1. Same red if a `*Probe*.cs` file ever hosts a second registered guard type (`SecondGuard` declared inside `ArtifactGuardProbe.cs` is counted as outside, because no `SecondGuard.cs` exists).

**False GREEN — the case this measure exists for.** A guard type `FooProbe` hosted at `src/housecarl-core/FooProbe.cs` is treated as counted, because `src/housecarl-generator/FooProbe.cs` happens to exist. `counted` is a set of bare filenames (line 130 drops the directory), so a name collision across projects hides a cross-assembly guard — the WriteEngine case, re-created.

Both close by deriving the file from the type rather than from its name: `t.Assembly` already separates `housecarl-core` from `housecarl-generator` for free, and comparing full paths rather than `Path.GetFileName` kills the collision. Happy to leave this to the chartered row-10 walk if that is the call — flagging it because the comment above claims the registry answer is the honest one, and this line reintroduces a proxy one step later.

.github/workflows/ci.yml
```diff
@@ -53,6 +53,47 @@ jobs:
...
+      # The residue countdown (ADR 0003, docs/decisions/): how much of the OLD
+      # harness is left. This prints the MEASUREMENT, not the committed baseline file. Printing the file was
+      # the first shape, and it was only honest for the two gated numbers: probeLines is deliberately
+      # ungated, so a legal in-place guard edit moves it and CI would have gone on printing the stale figure
+      # as though it had been measured. Re-running the derivation costs about a second and cannot drift,
+      # which a second copy of the derivation written in yaml would. W7 is not complete while these are
+      # above zero.
+      - name: Residue countdown — what is left in the old probe harness
+        if: ${{ !cancelled() }}
+        run: dotnet test src/housecarl-mcp-tests --configuration Release --no-build --filter "FullyQualifiedName~HarnessResidueTests" --logger "console;verbosity=detailed"
```
Avick3110: **This step has no empty-run guard, and the step 25 lines above documents exactly why it needs one.**

`dotnet test … --filter "FullyQualifiedName~HarnessResidueTests"` is subject to the measured behaviour the previous step exists to close: *"`dotnet test` exits 0 when its filter matches NOTHING"*. Rename or split `HarnessResidueTests` — a plausible conversion-PR edit — and this filter matches nothing, the step exits 0, and the countdown silently stops being printed with no signal in the log that a measurement disappeared.

The gate survives (`HarnessResidueTests` is `tier=unit`, so the `tier!=bridge` step still runs it), so this is the report going quiet rather than the guard going quiet. But quiet-and-green is the shape the step above rejects, and the fix is the same one already written: `--logger "trx;LogFileName=residue.trx"` plus the same executed-count check, or simply drop the filter and read the countdown lines out of `guards.trx` from the run that already happened.

src/housecarl-mcp-tests/RecordsRefusalGrammarTests.cs
```diff
@@ -0,0 +1,125 @@
...
+                     "the json lane's error property opens with the text lane's 'error: ' prefix");
+
+    [Fact]
+    public void TheTextTwinIsStillProseOpeningErrorColon() =>
+        Assert.StartsWith("error: ", RefusalText());
+
+    [Fact]
+    public void BothTransportsStateOneSentenceNotTwoSpellingsOfIt() =>
+        Assert.Equal(Doc(RefusalJson()).GetProperty("error").GetString()!.Trim(),
+                     RefusalText()["error: ".Length..].Trim());
```
Avick3110: **The slice hazard the comment fourteen lines up says was removed is still here.**

Line 46 records the fold: *"StartsWith rather than a `[..7]` slice: the slice threw ArgumentOutOfRangeException on any sentence shorter than seven characters, turning a clean assertion failure into an unrelated exception."* That fold was applied to the json sentence. This arm still does `RefusalText()["error: ".Length..]` on the raw text response.

If a regression makes the text lane answer empty or shorter than seven characters — which is the regression this file exists to catch — this test throws `ArgumentOutOfRangeException` from the range operator instead of showing the two sentences that failed to match. Same outcome the fold was made to prevent, one test down.

```suggestion
    public void BothTransportsStateOneSentenceNotTwoSpellingsOfIt()
    {
        var text = RefusalText();
        Assert.StartsWith("error: ", text);
        Assert.Equal(Doc(RefusalJson()).GetProperty("error").GetString()!.Trim(),
                     text["error: ".Length..].Trim());
    }
```

PR #477, opened 2026-09-02
Tool names are compile-time constants — the registry

### LINKED ISSUE #470: [Bug] housecarl_check is declared but never published: CheckTools lacks [McpServerToolType]; tools/list returns 42 of 43

### LINKED ISSUE #475: [NOTE] Tool-name registry: every shipped tool-name literal is a compile-time constant (amendment 2026-09-02 base (i))
## The ruling

RUN_ORDER amendment **2026-09-02 (base), part (i)** — Aaron-ruled, in the advisor session
("I agree on both your recommendations since my last message. the wider call and sequencing"):

> a compile-time tool-name registry — one static class of constants, every shipped literal
> interpolates them, so deleting a tool is deleting a constant and every reference is a compile
> error; surviving code only (rule 10(a) — deletion-flagged bodies are skipped).

It is the first of the three mechanisms the demolition method was re-shaped around, and it runs
between #470's PR and the cut (amendment 2026-09-02 (5)/(9) sequence). Its purpose is the cut's
checklist: after this lands, deleting a 1.x tool means deleting its constant, and the compiler
names every surviving site that still refers to it — instead of a hand-noticed population, which
is the root both #474 class-stops shared.

## The derived population

Derived by script, not by hand: `scripts/tool-names/derive-tool-name-sites.py`, run over
`src/housecarl-mcp`, `src/housecarl-core` and `src/housecarl-mcp-tests` (the last does not exist
on `main` @ `8f520b9`; it arrives with #470). Measured on `8f520b9`.

**124 `.cs` files scanned · 69 carry a site · 609 occurrences of `housecarl_[a-z0-9_]+`.**

| where | count | disposition |
|---|---|---|
| inside a string literal | **330** | classified below |
| inside a comment or XML doc | **279** | not rewritten — a comment is not shipped text and cannot be a compile error |
| in code (an identifier) | 0 | — |

The 330 literal sites, in the charter's own classes:

| class | count | what it is |
|---|---|---|
| **(a)** | 35 | the tool's own `[McpServerTool(Name = …)]` argument |
| **(b)** | 165 | a description / refusal / remedy sentence naming a tool |
| **(c)** | 79 | a code-path literal — the literal IS the name (AliasTable rows, redirect tables, census lists) |
| **(d)** | 43 | inside a deletion-flagged 1.x tool body — **skipped by rule 10(a)** |
| **(e)** | 8 | a name that is not a registered tool at all |
| | **279** | **rewritable (a + b + c)** |

Of the 279 rewritable sites, **254 are in `housecarl-mcp` and 25 in `housecarl-core`** (11 files).
257 sit in plain string literals and 22 inside interpolated ones; there are no verbatim (`@"…"`)
and no raw (`"""…"""`) sites at all.

Heaviest files (rewritable): `AliasTable.cs` 37 · `WriteTools.cs` 23 · `NexusTools.cs` 20 ·
`RecordsTools.cs` 19 · `CreateTools.cs` 11 · `LoadOrderService.cs` 11 · `PlaceAssetTools.cs` 11.

### The constant population: 46 DECLARED, not 45 registered

The constants are one per **declared** tool — every `[McpServerTool].Name` in the shipped
assembly — which is **46** on this base. That is deliberately not the *registered* set the SDK
scans (`[McpServerToolType]` on the type **and** `[McpServerTool]` on the method), which is
**45**: `housecarl_check` is declared and has never been registered (#470). Its attribute names
it in source, so it needs a constant; a completeness arm written against the registered set would
report that constant as spurious until #470 lands. The oracle is the declared set.

### Class (d) — the rule-10(a) exclusions, by name

The eleven deletion-flagged 1.x tools are not touched. They die at the cut.

Read family: `housecarl_read_record`, `housecarl_batch_record_detail`, `housecarl_resolve`,
`housecarl_read_plugin_file`, `housecarl_cross_plugin_query`, `housecarl_diff_record`,
`housecarl_effect_chain`, `housecarl_skypatcher_read`.
Check family: `housecarl_check_errors`, `housecarl_validate_scripts`,
`housecarl_validate_dialogue`.

Their 43 in-body sites: `ReadTools.cs` 36, `DialogueTools.cs` 4, `SkyPatcherTools.cs` 3. Sites in
*surviving* code that merely NAME one of the eleven (engine services, AliasTable's retired rows)
are rewritten normally — that is the point: at the cut, deleting the constant reds exactly those
sentences.

### Class (e) — the eight non-tool names

Six are retired 1.x write spellings in `AliasTable.RetiredTools`' `Old` column, left over from
PR #474: `housecarl_set_field`, `housecarl_bulk_apply`, `housecarl_create_record`,
`housecarl_bulk_create`, `housecarl_remove_record`, `housecarl_forward_record`. Two are
`housecarl_artifact` in `ResultArtifact.cs` — a JSON property name, not a tool, and a false
positive of the regex.

## Routing

Avick3110:
Closes #475

Every tool name in shipped code now references a compile-time constant instead of spelling it. 46 constants — one per declared tool — in `src/housecarl-core/ToolNames.cs`. **292 sites in 45 files** were rewritten by script. Nothing a caller sees changed: `tools/list` off the built server is byte-identical before and after.

The point is the next PR. Deleting a 1.x tool now means deleting its constant, and the compiler names the surviving sites that still refer to it — instead of somebody noticing them by hand, which is the root both of PR #474's class-stops shared.

Serves RUN_ORDER amendment 2026-09-02 (base), part (i).

---

## The population, derived (never hand-listed)

`scripts/tool-names/derive-tool-name-sites.py` lexes each shipped `.cs` file and classifies every `housecarl_[a-z0-9_]+` occurrence. On the original base `8f520b9`: **124 files scanned, 69 carry a site, 609 occurrences.**

| class | n | what |
|---|---|---|
| (a) | 35 | the tool's own `[McpServerTool(Name = …)]` argument |
| (b) | 165 | description / refusal / remedy prose |
| (c) | 79 | code-path literals — alias rows, artifact writers, `ToolSchemas`, `ResultsStore` paths |
| (d) | **43** | inside a deletion-flagged 1.x tool body — **skipped**, rule 10(a) |
| (e) | 8 | names no declared tool |
| | **279** | rewritten (a + b + c) |
| | 279 | in a comment / XML doc — not rewritten; a comment is not shipped text and cannot be a compile error |

After rebasing onto #476 the test project appears and is swept too: **+13 sites in 3 files** (`CheckWirePathTests` 10, `ServerTransportTests` 2, `ToolSurfaceCensusTests` 1), all plain literals, render check 13/13 identical. **279 + 13 = 292.**

**Class (d), the 43 rule-10(a) skips, by file:** `ReadTools.cs` 36 · `DialogueTools.cs` 4 · `SkyPatcherTools.cs` 3.

**Class (e):** the six retired 1.x write spellings in the `AliasTable.RetiredTools` `Old` column, plus two `housecarl_artifact` in `ResultArtifact.cs` — a JSON property name, not a tool.

Re-derived independently by the review session on both sides: reproduces exactly. On the branch tip, REWRITABLE = **0**.

## The wire

`scripts/tool-names/dump-tools-list.py` drives the real `housecarl-mcp.exe` over stdio and dumps `tools/list` as canonical JSON — names, descriptions, parameter descriptions and input schemas exactly as a client receives them.

| | tools | size | result |
|---|---|---|---|
| pre-rebase, vs `8f520b9` | 45 | 211,064 chars | **byte-identical**, sha `20cb6bec1f089f81` |
| post-rebase, vs `c4cf555` | **46** | 225,609 chars | **byte-identical**, sha `2ad33504dd5232f` |

46 now, because #476 published `check`. This is why the branch carries no `CHANGELOG` line.

**The wire is not the whole oracle, and this PR should not pretend it is.** It renders only attribute-borne text — 104 of the 279 original sites. The other **175, in 42 files**, are refusal prose, alias rows, artifact manifests and path builders that `tools/list` never renders, and the sweep's own render check covers them using *the same lexer that performed the rewrite*. Independent coverage instead came from:

- **the compiled `#US` string heaps** of both assemblies, base vs branch — 261 of 279 sites folded byte-identically; 18 literals split, exactly the interpolated rewrites;
- **fragment presence** — all 18 reassemble with every fragment intact; canaries RED;
- **fragment order from the IL** (`ldstr` operand sequence, 14,990 operands recovered) — 18/18 correct; reversed-order canary RED;
- **135 driven `tools/call` refusal responses** (~44k chars of off-wire prose) against an unconfigured server — base vs branch identical;
- three reviewers' own inverse transforms (substitute constants back, compare to base bytes): **42/42 files identical**, each with its own RED canary.

## The RED proof — all 46, not a sample

Deleting one constant at a time from a committed state, building, then restoring from the committed blob with the restore **verified by grep**:

| | |
|---|---|
| constants RED on deletion | **46 / 46** |
| silently deletable | **0** |
| green control (unmodified build) | GREEN, before and after |
| restores verified by grep | 46 / 46 |

**Two separate things shorten the first error list, and the checklist is iterative because of both.**

1. **Binding phases.** C# binds declarations — attribute arguments, const initializers — before method bodies and stops once the first phase has failed. A constant with 34 references across 8 files reported 10 diagnostics in 5 files on the first build.
2. **Project order, which hides more.** A failed project skips everything downstream of it. The registry is in core and the tool surface depends on it, so a constant deleted in core fails core and the tool project is never compiled at all. Measured: a constant with 11 references across 6 files in **both** projects reported diagnostics in **exactly one file** — its single core reference; the five tool-project files appeared only once core built again.

**For the cut session: build, fix what is named, build again, until green. A short first error list is not "that was all of them", and an error list confined to one project is not evidence that the other is clean.**

## The completeness arm

`src/housecarl-mcp-tests/ToolNameRegistryTests.cs` holds three reflected sets equal — the ToolNames constants, the names the source **declares**, and the names the SDK **registers**:

| set | count |
|---|---|
| constants | 46 |
| declared | 46 |
| registered | **46** (was 45 before #476 published `check`) |

This is what makes the registry safe to stop regenerating, and it closes #470's class by construction: a future declared-but-unregistered tool fails here instead of shipping unreachable for twelve days.

Mutation-tested from a committed state: a **spurious constant** → RED; **stripping `[McpServerToolType]`** so declared ≠ registered → RED; unmutated control → GREEN (4/4). Restores grep-verified, tree clean.

## Review rounds

Conducted by a fresh session that did not write this code (RUN_ORDER rule 9). Every reviewer Opus, in its own isolated worktree, seeded with the worktree/branch/base, the conventions pointer, and the tagged settled-decisions list — nothing else.

**Round 1 — 5 reviewers** (4 seeded + 1 gate-replica seeded with nothing but "review this branch").

Folded, each with an executing arm, a green control and a RED canary:

| finding | fix and its proof |
|---|---|
| The registry file told maintainers to re-run `--emit-registry`; that command **crashed** on the tree it had produced. Worse was the **silent** half: a *new* literal on the swept tree was classified as a retired spelling and `--rewrite` skipped it saying nothing | the deriver resolves `Name = ToolNames.X` through the registry and refuses when it can resolve neither. Two regenerations byte-identical; constants unchanged 46/46; a new site visible again (REWRITABLE 1, was 0), clean-tree control 0 |
| The rewriter **silently corrupts verbatim/raw literals**, and its own render check is blind by construction (same escape model both sides). Latent — every site is `regular` | refuses `literal_kind != "regular"`; verbatim and raw both REFUSED, regular control still sweeps |
| A 29-line rationale essay in a **public** source file citing the private corpus, against the Aaron-ruled terse-comments standard; and no ADR | rationale → ADR; summary keeps the contract plus a pointer; private citations removed from the header and the 11 per-constant notes |
| At the cut, the compile error at the alias rows means **restore the literal**, not delete the row — the opposite remedy from every other site, and unwritten | one terse line at the array |
| The deletion-flag list is the one hand-written population | says why it is hand-written and what its one-directional check cannot catch |
| `sites.tmp.json` in the repo root with no `finally`; non-atomic rewrite; unused import and parameter | temp dir + `finally`; every file buffered before any is written |

Refused, with grounds: sweeping `housecarl-generator` (the ruled harness boundary); deriving the deletion flag (policy, with no source property to derive from — an honest note instead); rewriting commit messages (history is not shipped surface); the precedent defence of private lore (doc-hygiene §8 rule 3 binds new code prospectively); the `+`-run observation (a known measured consequence).

**Round 2 — 2 reviewers** (the two-per-round directive arrived mid-engagement and binds from here).

**A class-stop was called and escalated under §4, framing (a).** The class: *defects in the committed sweep tooling, while the shipped rewrite came back clean from every reviewer and every oracle.* Round 1 found three; round 2 found a fourth — and the fourth **was introduced by round 1's own fold**. Severity is explicitly not the signal, so a reporting defect in the same file, same lane, same root is the same class.

The ruled mechanism was untouched. The seam was the `inferred` rider that the committed scripts are a **re-runnable maintenance path** — the promise that turned every script defect into a live claim to keep true. **Resolution: drop the promise, harden nothing.** The generated header and all three script docstrings now say these scripts are the one-shot migration record of how the registry and the rewrites were derived, that they are not re-run, and that a new tool's constant is added by hand and held by the completeness arm above.

Also folded from round 2, all documentation: the ADR's missing second cause of the short error list (project order, measured); its reference counts re-measured (34 refs / 8 files, 10 diagnostics / 5 files); private-corpus citations in public scripts replaced by the public ADR; `279` corrected from a literal count to a site count over 265 literals; and the deriver's one lexer limit recorded — it does not model a nested string inside an interpolation hole, its own render check compares the mis-span against itself and cannot see it, and the class was closed from outside by the compiled string-heap and `ldstr`-order comparison, which does not use that lexer at all.

**No round 3** — the class-stop is the terminator, not the three-round cap.

**The report-line bug was fixed after all, on Aaron's word:** *"'never fold a third instance' to my understanding is supposed to be an escalation trigger not a 'don't fix anything more' trigger. If it has resulted in us leaving a known and easily fixable bug in our wake then that rule needs relaxed."* The escalation had already been ruled and the seam removed, so this is the leftover instance rather than a fourth fold of an open class. It lands as its own commit, with the reviewer's own reproduction as the RED cell: reverting the three test-project files and re-running `--rewrite` prints **one name three times before the fix, three distinct names after**.

## Open for the gate — not folded

| # | item |
|---|---|
| 1 | **The alias `Old` column is now half constants (11 deletion-flagged), half literals (6 retired).** At the cut those 11 need hand-conversion back to literals — a second hand-kept step inside the very table the ADR says must not have one. A reviewer argues they are future retired spellings and belong as literals *today*. The trade: literals-today breaks "every shipped literal interpolates them" (ruled) and the commit-is-the-script's-output property; constants-today costs 11 compiler-named restore-the-literal remedies at the cut, already commented at the site. **Advisor's recommendation: keep constants, and carry the remedy into the cut's kickoff.** |
| 2 | **The description-vocabulary guard does not merely go quiet — it prints an affirmatively false zero.** Its declared-but-unregistered line reported 3 references by file:line on base and reports `0` on the branch, while those sentences still ship. That is the Q3 shape (never a silent wrong answer). The property has moved to the completeness arm above; that was a deliberate one-path choice, because teaching the old guard to read the registry would be in-place growth in a harness being retired, which amendment (8) names as the thing that would invalidate the residue plan. Stated here rather than left as a silent window. |
| 3 | **A dated future breakage.** That arm's anti-vacuity floor (`carrying > 0`) is held up entirely by the 43 literals in the three deletion-flagged files. When the cut deletes them, `carrying` becomes 0 and **the arm goes RED on the cut commit**. It belongs in the cut checklist, not in the cut session's surprise budget. |
| 4 | **The `Co-Authored-By: Claude Fable 5.1` trailer is a harness artifact**, not evidence of a Fable agent: every build and review agent in this engagement was spawned Opus, and the advisor holds the spawn record. The trailer comes from the global attribution rule in CLAUDE.md, so changing it is self-governing config — surfaced rather than self-committed, per §5 #10. |

## ADR and CHANGELOG

- **ADR: yes** — `docs/decisions/0004-tool-names-are-compile-time-constants.md`. Round 1 found the branch had argued three architectural decisions (a registry exists; it lives in core; `const` not `static readonly`) inside a source-file header instead. Numbered **0004** because #476's `0003` landed first; both live pointers moved with it, and the derived grep for the old number returns zero.
- **CHANGELOG: no** — nothing consumer-observable changed, and the byte-identical wire at 46 tools is the evidence.

## Gates, measured from the committed tip

| | branch | rebased base `c4cf555` |
|---|---|---|
| `dotnet build -c Release --no-incremental` | 0 errors, 89 warnings | 0 errors, **89 warnings** |
| `ci-all` | **132/132**, exit 0 | **132/132**, exit 0 |
| `dotnet test` (all tiers) | **327/327 passed** | — |
| `tools/list` | 46 tools, byte-identical to base | |

The ci-all count is 132, not the 133 measured before the rebase: #476 moved a probe out of the old harness into the test project. Base and branch agree, so it is the new baseline rather than a regression.

---

## Gate folds

Aaron's gate review posted four findings; the advisor triaged all four as confirmed. Folded by a
fresh session that neither wrote nor reviewed this branch (CLAUDE.md §5 #11, RUN_ORDER rule 9).
The four review-comment ids were re-fetched and matched against the triaged set before any edit —
no un-triaged comment exists on the PR.

| # | finding | commit |
|---|---|---|
| RC 3919054134 | nothing pins a published name to a literal any more | `aae8df1` |
| RC 3919057104 | ADR's declared-but-unregistered instance expired at the rebase | `e25d28d` |
| RC 3919057207 | ADR still promised the scripts are a re-runnable maintenance path | `e25d28d` |
| RC 3919057334 | "it does not fail; it empties" — it also fails, at the cut | `e25d28d` |

Folds 2–4 are one docs commit: they are three sentences in one immutable file, and fold 3 required
reading the whole document for the same class.

### Fold 1 — the literal anchor, and its RED evidence

`ToolSurfaceCensusTests.ToolsListNamesHousecarlCheck_TheWholePointOf470` goes back to
`Assert.Contains("housecarl_check", …)`, with a comment saying the literal is deliberate and why.
The other 12 swept sites are untouched — they drive the tool and should follow the constant; one
anchor is the property.

**RED check**, from the committed state, restore grep-verified, tree clean after:

| sabotage | test project | ci-all |
|---|---|---|
| `ToolNames.Check` value → `"housecarl_chekc"` | **RED**, this cell only — `Assert.Contains() Failure … Not found: "housecarl_check"`, `ToolSurfaceCensusTests.cs:109`. `Failed: 2, Passed: 325` | RED: `binding-shim-guard`, `codex-umbrella-coverage-guard` |

The reviewer's scenario reproduced exactly: the census set-equality cell
(`EveryToolDeclaredInSourceIsPublishedOnTheWire…`) and all of `ToolNameRegistryTests` stayed
**GREEN** — both compare sets that moved together — while the restored literal named the defect.

### The 1.9 oracle — measured, and it covers 39 of 46

The frozen `data/tools-list-1.9.0.json` (45 names, literals by construction) **does** anchor
1.9-era names, through `RetiredNameRedirectTests`: a 1.9 name that stops being published must have
an `AliasTable` redirect row, and a renamed tool has none.

| sabotage | result |
|---|---|
| `ToolNames.AssetStatus` → `"housecarl_asset_sttaus"` (in the 1.9 fixture) | **RED** — `RetiredNameRedirectTests.EveryToolNineteenPublishedIsStillLiveOrRedirectsToRegisteredSuccessors` |
| `ToolNames.Apply` → `"housecarl_aply"` (**not** in the fixture) | **no test-project cell RED** — only the retiring harness (`write-surface-guard`, `binding-shim-guard`, `codex-umbrella-coverage-guard`, `description-vocab-guard`, `copy-parser-guard` via the ci-all bridge) |

Set arithmetic: 39 of the 46 current names appear in the 1.9 fixture and are anchored by it. The 7
that do not are the 2.0-era names — `housecarl_apply`, `check`, `copy`, `create`, `forward`,
`records`, `remove`. Fold 1 anchors `check`; the remaining **six are anchored today only inside the
guard harness that is being retired and converted**. Stated, not widened — the advisor carries it
as a chartered item.

### ADR 0004, re-read whole

Fold 3 also removed one other sentence implying re-runnability: the Decision line's "a single
**generated** class" → "a single class" (houseCARL has a real build-time generator project, so
"generated" reads as build-time emission, which is false of this file).

After the folds, no sentence in the ADR is false about the tree at this tip. One thing it is now
silent about: fold 1's deliberate literal. That is not a falsity — the ADR's claims are scoped to
"shipped code" and "the shipped assemblies", and the test project is neither.

### Gates, re-measured from the folded tip `e25d28d`

| | baseline `de733fb` | folded `e25d28d` |
|---|---|---|
| `dotnet build -c Release --no-incremental` | 0 errors, 89 warnings | 0 errors, 89 warnings |
| `ci-all` | 132/132, exit 0 | **132/132, exit 0** |
| `dotnet test` (all tiers) | 327/327 | **327/327** |

No new rounds were spawned — CLAUDE.md §5 #11, "no new rounds after directed folds".


src/housecarl-mcp-tests/ToolSurfaceCensusTests.cs
```diff
@@ -101,6 +101,6 @@ public void EveryToolDeclaredInSourceIsPublishedOnTheWire_AndNothingIsPublishedT
     [Fact]
     public void ToolsListNamesHousecarlCheck_TheWholePointOf470()
     {
-        Assert.Contains("housecarl_check", _s.PublishedNames);
+        Assert.Contains(ToolNames.Check, _s.PublishedNames);
```
Avick3110: **[medium] Nothing pins a published tool name to a literal any more.**

On base, `housecarl_check` as a *published wire string* was pinned by 13 literal assertions: this line, `CheckWirePathTests` x10 (which drives the real server over stdio), and `ServerTransportTests` x2. All 13 now read `ToolNames.Check`, so every one of them follows a change to the constant's **value**.

Scenario: `ToolNames.Check` becomes `"housecarl_chekc"` — a typo in a later rename, or a bad merge at the cut. `CheckWirePathTests` calls the renamed tool and finds it. `ToolSurfaceCensusTests`' `Assert.Equal(declared, published)` compares two sets that both moved. `ToolNameRegistryTests` holds constants == declared == registered, all three moved together. ci-all's TOOLNAME arm reads source literals and now sees no `housecarl_` token at these sites at all. Everything is green, `tools/list` publishes the new name, and every skill, doc and user invocation of `housecarl_check` gets "unknown tool". This test — named `TheWholePointOf470` — was RED on exactly that before the sweep.

The branch's own rationale ("a deleted tool should break its tests at compile time") holds for *deletion*; it does not cover *renaming*, and the renaming oracle was the only thing these literals were doing. The `dump-tools-list.py` byte-identity check was a one-shot measurement, not a guard.

One literal here restores the anchor at no cost to the deletion property — the census test's other cell already holds the set equality by reflection.

docs/decisions/0004-tool-names-are-compile-time-constants.md
```diff
@@ -0,0 +1,91 @@
...
+  `[Description(…)]` and similar attribute arguments, which must be constant expressions.
+  `"…" + ToolNames.Records + "…"` is one; a `static readonly` field is not. This is forced, not
+  preferred.
+- **The registry lives in `housecarl-core`, not `housecarl-mcp`.** The project reference runs
+  mcp → core, and 25 of the rewritten sites are in core, so a registry in mcp could not reach
+  them. Placement follows the derived population.
+- **The population is DECLARED, not registered.** A constant exists for every tool whose
+  `[McpServerTool(Name = …)]` attribute names it in source. That is deliberately not the set the
+  SDK actually scans, which additionally requires `[McpServerToolType]` on the declaring type.
+  One tool is declared and has never been registered; its attribute still spells its name, so it
```
Avick3110: **[medium] This claim was falsified by the rebase, and the ADR is immutable once merged.**

"One tool is declared and has never been registered" was true against the pre-rebase base `8f520b9` (declared 46, registered 45 — `housecarl_check`). After the rebase onto `c4cf555`, #476 registered it and all three sets are 46. `c070dbe`'s own commit message says so: *"All three are 46 now that #476 published check."* The ADR sentence was not updated with it.

Two consequences, in a document headed **ARCHIVE — immutable once merged**:

1. A reader following `ToolNames.cs`'s rationale pointer goes looking for the tool this names and finds none.
2. The next sentence — *"A completeness check written against the registered set would report that constant as spurious"* — is contradicted by this branch's own `ToolNameRegistryTests.DeclaredEqualsRegistered`, which asserts `declared == registered` and would pass identically against either set.

The *decision* (population = declared) is still right and still worth stating — the SDK's predicate is the wider one and a future unmarked type is exactly what the new test catches. It is the supporting instance that has expired. Rewriting it in the past tense ("until #476, `housecarl_check` was declared and never registered — the case that showed why") keeps the argument and stops the record asserting something false forever.

docs/decisions/0004-tool-names-are-compile-time-constants.md
```diff
@@ -0,0 +1,91 @@
...
+  is owed a constant. A completeness check written against the *registered* set would report
+  that constant as spurious.
+- **Retired spellings get no constant.** A name that no longer names a tool has no constant
+  whose deletion should break anything, and a second hand-kept population is the hazard this
+  decision exists to remove. Retired names stay literals in the alias table that redirects them.
+- **The registry is generated, and regeneration is idempotent.** It is emitted from the
```
Avick3110: **[medium] This is the promise round 2 withdrew, still standing in the document `ToolNames.cs` points at.**

`e800f18` was the class-stop disposition: *"drop the promise, harden nothing."* It rewrote `ToolNames.cs`'s header and all three script docstrings to say the scripts are the one-shot migration record, are **not re-run**, and that a new tool's constant is added **by hand**. It edited this ADR in the same commit — and left this bullet.

So the two documents now disagree, and the one that is wrong is the one the other points at:

- `ToolNames.cs`: *"Those scripts are not a maintenance path and are not re-run. A new tool's constant is added BY HAND … Rationale: docs/decisions/0004-…"*
- this ADR, at that rationale: *"The registry is generated, and regeneration is idempotent … it can be re-run on its own output."*

Scenario: the next session adds a tool, reads `ToolNames.cs`, follows the pointer for the how, and re-runs `--emit-registry` against scripts nobody maintains — which is the live claim the class-stop existed to retire. Worth saying instead what the decision actually is: the registry was *produced* by script once, and is maintained by hand under the completeness test.

docs/decisions/0004-tool-names-are-compile-time-constants.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Two residues remain, both by rule.** Deletion-flagged 1.x tool bodies keep their literals —
+they are condemned code, and editing them is work that dies with them. The guard suite is a
+separate project that is itself being retired and converted; nothing leaves it except by a
+conversion change. So the compile-error checklist covers the shipped assemblies, not every file
+in the repository.
+
+**A guard becomes redundant.** A source-scanning check that held every `housecarl_` token in
+caller-facing prose against the set of real tool names loses most of its population, because a
+constant reference cannot name a tool that does not exist. It does not fail; it empties. The one
```
Avick3110: **[low] "It does not fail; it empties" — the PR body says it does fail, on a date.**

`DescriptionVocabularyGuardProbe`'s TOOLNAME arm asserts `violations.Count == 0 && registered.Count > 0 && carrying > 0` — it carries an anti-vacuity floor, so emptying is precisely what it will not tolerate. On this branch `carrying` is held up only by the 43 class-(d) literals left in the three deletion-flagged files; when the cut deletes those bodies, `carrying` hits 0 and the arm goes RED. Open item 3 of the PR body says the same thing.

So the cut session reading this ADR for what to expect is told the guard quietly empties, and instead hits a RED ci-all with no explanation in the record it was pointed at. One clause ("…it empties — and its `carrying > 0` floor then fails, which is the cut's to dispose of") makes the immutable record agree with the checklist.

PR #481, opened 2026-09-03
The old harness's roster and the residue populations are derived, not hand-kept

### LINKED ISSUE #478: [NOTE] Old-harness roster and residue populations derived, not hand-kept (residue countdown, amendment 2026-09-02 (7))
The old probe harness keeps two hand-maintained populations, and both are the shape the project's own rule
forbids — a guard's subject set derived by hand rather than from the surface it claims (CLAUDE.md §5 #11).
They also make every conversion PR conflict with every other conversion PR, because both live in files each
one edits.

This is the chartered remainder of PR #476's class-stop: Aaron ruled "fix the live half there, charter the
rest". Items 3 and 4 are that rest, verbatim from #476's "Chartered, not fixed here" section. Item 5 is the
chartered remainder of PR #477's gate.

## 1. The roster is a hand-maintained table

`src/housecarl-generator/CiAll.cs` lists every CI probe as a row in a `(string Name, Func<string[], int> Run)[]`
literal. Adding a probe means adding a row; deleting one means remembering to remove it. The residue
countdown's `ciAllRows` reads that table, so the countdown's honesty rests on a list somebody maintains.

Each probe entry method gains a `[CiProbe("<name>")]` attribute and the runner enumerates them by reflection,
sorted by name. Deleting a probe file then deletes its row by construction.

## 2. The residue baseline stores four hand-kept totals

`src/housecarl-mcp-tests/harness-residue-baseline.json` holds four integers. Every conversion PR lowers all
four, so any two conversion PRs conflict on that file whatever families they convert. A per-file map — probe
file to its line count and its share of the roster — makes a conversion PR delete its own key instead, and
the totals derive from the map.

## 3. Guards wired as their own `ci.yml` step, outside the roster

`.github/workflows/ci.yml` runs `freshness-capture-guard` as its own cold step, deliberately outside the
roster. Nothing derives that population from the workflow, so a second such step would be invisible to the
countdown: the countdown could reach zero, W7's closing condition, with CI still running guards.

The population comes off the workflow file itself — every `housecarl-generator.dll <verb>` invocation is
either a roster row or an attributed standalone verb, and a step naming a verb with no attribute fails by
name.

## 4. The census scopes "declared" to an assembly by hand

`ToolSurfaceCensusTests` reflects `typeof(CheckTools).Assembly` while `Program.cs` registers with
`WithToolsFromAssembly()` (the calling assembly). The two agree today by coincidence of layout, not by
construction. Replaced with the invariant: the assembly the census reflects is the assembly the server
registers from, exposed once from the server's composition, so a tool surface split across assemblies fails
loudly. Latent today — `housecarl-core` carries no MCP SDK reference — so the arm is written against a
fixture that proves the failure direction anyway.

## 5. Six 2.0 tool names are anchored only inside the retiring harness

The frozen 1.9 wire capture anchors 39 of the 46 published names as literals. Six 2.0-era names
(`housecarl_apply`, `_copy`, `_create`, `_forward`, `_records`, `_remove`) plus `housecarl_check` are not in
it, so after PR #477 a typo in one of those constants' VALUES moves the declared set, the registered set and
the published set together and passes every cell in the test project. A frozen 2.0 wire capture
(`data/tools-list-2.0.json`, literal names, produced from the built server once) plus one test asserting
every name in it is published closes that, in the same shape as the 1.9 oracle.

## The populations, measured on `main` @ 9d897b3

| population | measured | derived from |
|---|---|---|
| probe files | 155 | `*Probe*.cs` under `src/housecarl-generator`, recursive, bin/obj excluded |
| roster rows | 132 | rows in the `CiAll.Probes` table |
| distinct host types behind those rows | 124 | the rows' dispatch delegates |
| guard host files the probe-file glob does not count | 1 | `src/housecarl-core/WriteEngine.cs`, two coerce verbs |
| `housecarl-generator.dll <verb>` invocations in `ci.yml` | 2 | `ci-all` and `freshness-capture-guard` |
| standalone (out-of-roster) CI guard steps | 1 | that second invocation |
| published tool names anchored as literals by the 1.9 capture | 39 of 46 | `data/tools-list-1.9.0.json` |

Avick3110:
Closes #478

Two hand-maintained populations in the old guard harness become derived, and several smaller hand-kept things go with them. `CiAll.cs` held a 132-row table of `(verb, delegate)` pairs; every guard entry point now carries `[CiProbe("verb")]` and the runner reflects them. The residue baseline held four totals that every conversion PR edits; it is a per-file map now, and the totals derive from it, so a conversion PR deletes its own key and two conversions merge cleanly.

Nothing left the old harness. No family converted, no probe deleted, no 1.x tool body touched, no skill touched, no shared test fixture touched. This changes how the harness is enumerated, not what is in it.

**Opened against `main` @ `7afc7c3`.** The branch was rebased onto it at PR-open time: main had gained eight commits since the merge base, all touching only `CLAUDE.md`, and this branch touches no file main touched (empty intersection, checked). All three gates were re-run after the rebase.

---

## The gates, measured

| gate | on `main` | on this branch |
|---|---|---|
| `dotnet build housecarl.sln -c Release --no-incremental` | — | **0 errors, 89 warnings** |
| `ci-all` | 132/132 | **132/132, exit 0** |
| `dotnet test src/housecarl-mcp-tests` | 327 | **383 passed, 0 failed** |
| tiers | — | unit 79 · integration 182 · stdio 121 · bridge 1 |

Every gate was re-run by this review session from the committed tip at boot, again after each fold, and again after the rebase.

**A note on the warning count, because the decision record says 4.** That figure came from an incremental build where nothing recompiled. A full `--no-incremental` build emits 89: 88 × CA1416 (platform-specific ACL calls at `PlaceAssetProbe.cs` ~1317–1383) and 1 × xUnit2030 (`RecordsRemedyGrammarTests.cs:66`). Both are outside this branch's hunks and pre-existing. 89 is what a reviewer will actually see.

## Every derived population, with its count

| population | value | derived from |
|---|---|---|
| `[CiProbe]` verbs in the repo | **133** | reflection over the repo's project assemblies |
| ci-all roster rows | **132** | the same, minus `Standalone` — equals ci-all's own `132/132` |
| standalone CI guard steps | **1** | `freshness-capture-guard`; its `ci.yml` step is now generated from the attribute |
| assemblies the runner walks | **4** | `houseCARL-Setup`, `housecarl-core`, `housecarl-generator`, `housecarl-mcp` |
| repo projects (the second derivation's source) | **5** | `src/*/*.csproj` |
| shipped assemblies (split-surface population) | **4** | the project list, less the test project |
| probe files | **155** | `*Probe*.cs` under `src/housecarl-generator`, bin/obj excluded |
| probe lines | **65,380** | the sum of the per-file map (report-only) |
| guard files outside the counted set | **1** | `src/housecarl-core/WriteEngine.cs`, two coerce verbs |
| published tool names | **46** | `tools/list` off the built server |
| names the 1.9 capture anchors as literals | **39 of 46** | `data/tools-list-1.9.0.json` |
| files scanned by the one-home guard | **336** | the project list |
| tool-surface type names it matches against | **182** | reflection over `ToolSurface.Assembly` |
| sites naming the surface by picking a type | **0** (was 8) | the same scan |

I re-derived the roster a third way, independently of both the runner and `CiProbeRosterTests`: 135 raw `[CiProbe(` matches in source, less 2 that are refusal-message text (`CiAll.cs`, `CiWorkflowGuardStepTests.cs`) = 133 attributes, of which 1 carries `Standalone`. Hosted in 130 generator probe files — `PerkRefsProbe.cs` carries two — plus `WriteEngine.cs`. It reproduces.

**One correction to the decision record's §2.** It says "131 files changed by exactly one line; one changed by four". Measured: 130 probe files changed — 128 by exactly one line, `PerkRefsProbe.cs` by two (it hosts two verbs), `FreshnessCaptureProbe.cs` by four. Every added line in a probe file is a `[CiProbe(...)]` except three comment lines on `FreshnessCaptureProbe`'s summary, and no probe line was removed.

## The measured negative on the ordering hypothesis

The old table was in `ci.yml` step order; the roster is name-sorted, which reorders 132 probes. The build session flagged this as a free test for inter-probe ordering dependencies. **Measured, and reproduced independently by three of the four reviewers: ci-all is 132/132, exit 0, under the new order.** Nothing was pinned to preserve the old sequence and nothing needed to be. The one comment that asserts an order (`ArmClassificationProbe.cs:270`, "sits BEFORE emit-match-guard") is still true ordinally.

## ADR 0003 — dated superseding note

ADR 0003 merged with #476 and is ARCHIVE, so it is amended by a dated note at the head of the file; its Decision section is untouched. Three descriptions had expired on this branch:

1. "Three of its numbers gate" — four do; the line total is now the fifth number, still derived and printed rather than asserted.
2. The baseline is described as totals; it is a per-file map the totals derive from.
3. "remove the registry row" — there is no registry. The note names the Consequences sentence a conversion PR reads as its recipe and states what that recipe is now.

The note also carries the bound on the fourth measure (below).

## The review rounds

**Two rounds, four reviewers, all Opus, each in its own worktree with its own `TMP`.** Round 1 was one seeded reviewer plus one gate-replica reviewer seeded with nothing but the branch and base; round 2 was two seeded reviewers. Full results are in the lane's scratch folder.

**Round 1 — 0 high, 6 medium, 8 low. 8 folded across 5 commits, 1 refused.**

Both reviewers independently re-derived the branch's load-bearing claim — main's 132-row hand table against the attributed set — and both found them identical, same verbs at the same entry points, plus `freshness-capture-guard` newly attributed `Standalone`.

- **The workflow-step guard hand-spelled the invocation form.** A `ci.yml` step spelled `dotnet run --project …` or `.exe` ran an unattributed verb in CI, uncounted, everything green — reproduced by both reviewers with different spellings. Folded, then superseded in round 2 (below).
- **The split-surface arm walked the tool surface's reference closure.** The direction is housecarl-generator → housecarl-mcp → housecarl-core, so housecarl-generator was never in the population — and it is the one shipped assembly that already compiles against the MCP SDK. A reviewer added a real `[McpServerToolType]` tool there: census tests and ci-all both green over an unreachable tool. Fixed by deriving from the csproj list, sharing one home with the roster's second derivation.
- That move also fixed a loader that enumerated `<project>/bin` directly, which throws `DirectoryNotFoundException` on a never-built project — the one case its written refusal was for, so the sentence was unreachable.
- **Six false-about-now statements**, each made true rather than reworded: `ToolSurface.cs`'s "one home" claim, the ADR's conversion recipe, a Program.cs comment contradicted by seven unreachable dispatch blocks (deleted; all seven then driven on the wire), inconsistent measure counts in three texts, a doc pointer to a nonexistent ADR filename, three references to the deleted `CiAll.Probes`.

I also caught and fixed a defect I introduced in my own fold, by reading the diff before RED-checking it: a static property initializer turned a written refusal into a `TypeInitializationException` that poisons every later test touching the class.

**Round 2 — 0 high, 3 medium, 7 low. Class-stop.**

Round 2 returned two more instances of round 1's first class — a guard's subject set derived by a convention rather than from the surface it claims — and both were green over a real offender: the workflow guard read **one hand-named file** (a reviewer added `nightly.yml` running a roster verb *and* an unattributed verb: 11/11 green), and its exclusion predicate was line-scoped, so a `dotnet build … && dotnet …dll <verb>` step dropped out of the population entirely (4/4 green). One reviewer wrote the class note itself: *"the honest shape is one derivation, not two more patches to a line filter."*

Four instances over two consecutive rounds, three of them green over a real offender, and three of the four were the same arm patched three times. **Stopped and escalated rather than folding a third patch.** No round 3.

## What the class-stop was ruled, and what it cost

<a name="open-1"></a>**OPEN FOR AARON, `inferred` (the advisor's, within charter) — the first item on this list, and the one to strike if it is wrong.**

The assumption that fell was that the out-of-registry population can be derived by *reading* `ci.yml`. That was the advisor's mechanism, not Aaron's; the ruled goal — the countdown cannot reach zero while CI runs a guard outside `ci-all` — is untouched. Ruled: **generate the steps from the attribute, state the bound plainly, and delete the parser.**

- `CiWorkflowGuardStepTests` **renders** the standalone-guard steps from `CiAll.StandaloneProbeNames` and asserts `ci.yml` carries that block byte-equal, markers included, exactly once. Flag a guard `Standalone` and its step is owed; hand-edit, delete, respell or duplicate one and the test is red naming the lines the workflow does not carry and printing the block to paste.
- **Deleted:** `GeneratorLines`, `RunsNoVerb`, `ByBuildOutput`, `ByProject`, `VerbOn`, `InvokedVerbs`, and the three arms built on them. An arm that cannot be made honest is deleted, not strengthened.
- **Kept:** one token arm — no ci-all roster verb may be named anywhere in the workflow. It reads tokens against a derived population and claims nothing about invocation syntax, so that claim is complete over its own surface.
- **The bound, stated once in the baseline's `_what` and the ADR note, and claimed nowhere more strongly:** `standaloneCiSteps` counts attributed standalone verbs; CI's steps for them are generated; **a guard invocation hand-added to a workflow outside the generated block is outside the countdown's reach.** Completeness over the sanctioned path, not over arbitrary shell.

The leftover instance was then fixed with its arm — and it is the clearest evidence the class was real, because **my own round-1 fold of it was itself an instance**. Round 1 said `ToolSurface.cs`'s "every census reflects the same property" was false; I repointed the two sites the reviewers named. Round 2's reviewer derived the population with a grep and found two more, both live `ci-all` guard rows. My own second pass by grep was *also* short: the derived arm named three more, including a second `WithToolsFromAssembly` call site in `PreFlattenSurface.cs` — a registration that could diverge from `Program.cs`'s. Eight sites in total, and the arm now derives both sides (type names by reflection, files from the project list) with 0 offenders over 336 files.

## Sabotage, from committed states

**Five sweeps, 23 cells, every one with a hand-proven canary and a green control.**

| sweep | cells | result |
|---|---|---|
| the workflow-line population (round-1 fold) | 8 | 8/8 |
| the shipped-assembly population | 3 | 3/3 |
| the generated fragment (the ruled fold) | 5 | 5/5 |
| the one-home arm + the clash refusal | 4 | 4/4 |
| reviewers' own sweeps (rounds 1 and 2) | 20 | every gated arm the branch adds or changes fired |

**One sweep came back with a green canary and was thrown out, not reported as a pass.** The fragment sweep's canary cell used `.` to stand for an em-dash in a `sed` pattern; the em-dash is three UTF-8 bytes, so the sabotage never applied and the cell read green. That is the exact failure the canary rule exists to catch. The pattern was fixed, a "sabotage did not apply" tripwire added, and the sweep re-run: 5/5 with the canary firing.

Cells that matter: the `ci.yml` step hand-renamed → red naming the missing step; the block duplicated → red saying CI would run the guard twice; a new attributed standalone verb with no step → red naming the verb; the standalone attribute removed → the vacuity canary fires; a legal `dotnet build …/housecarl-generator.csproj` line → green, no nuisance red. A site back on `typeof(ApplyOp).Assembly` → red naming the site; a new site in an untouched file → red; a duplicate verb outside the runner's closure → the written refusal rather than the BCL's `An item with the same key has already been added`. A typo in a `ToolNames` constant's value → both anchor arms red, naming the old and the new spelling, which is what the 2.0 capture exists for.

## What I refused, and why

**One finding refused across both rounds** (rounds folded at 8/9 and 6/6 respectively; the round-2 folds were all under a ruling).

`HarnessResidueTests.TypeDeclaration` matches `class Foo` inside a comment or string, so one ordinary cross-reference comment can redden five residue tests. Reproduced by both round-1 and round-2 reviewers. Refused: it fails **loud**, never toward green — `DeclaringFile` requires exactly one hit and names the type on zero or many, so the count cannot go quiet over it; the regex is the one #476 shipped and this branch does not touch it; and every candidate fix is another heuristic over raw text (skipping `//` lines leaves string literals; requiring a declarator token is a second guess). Trading a loud papercut for a quieter one in the derivation the whole countdown rests on is the wrong direction. The round-1 gate-replica reviewer reached the same conclusion independently and did not file it. **Named here as a sharp edge the conversion lanes will meet, not dropped silently.**

Reviewers also examined and refused, between them: the sorted-by-name reorder; `CiProbeAttribute` living in housecarl-core; `houseCARL-Setup` being loaded and scanned; the `SplitSurfaceFixture` tool name leaking into a census (checked against all six other reflectors); `GetReferencedAssemblies` pruning (measured: all four walked); a Debug/Release mtime hazard in `BuiltAssembly` (driven and found inert — `Assembly.LoadFrom` resolves out of the test's app base); and `RepoProjects` missing a project (checked: all five csprojs are one level deep).

## Measured, out of branch, not a finding

`check-guard` run standalone returns 1 (two `type=` refusal arms fail) while passing inside `ci-all`. Not this branch: `TryDispatch` intercepted it identically on `main`, verified. The cause is the co-hosting contract `CiAll` documents — the runner pre-generates a canonical corpus and sets `CorpusRulebook.CorpusPath`, which a standalone run has not got. Loud, with named arms, so a documented environment dependency rather than a defect.

## Deliberately not here

- **No CHANGELOG line.** Nothing a caller sees changed; `tools/list` is byte-identical.
- **No conversion.** No probe deleted, no family moved. `probeFiles` is still 155.
- **No fixture change.** `ServerFixture`, `HarnessPaths`, `AssemblyInfo`, `RecordsWorld`, `RecordsTestBase` untouched. `RepoProjects.cs` is a new file.
- **`scripts/tool-names/` untouched.** The registry's 46 constants are as #477 left them.
- **The scripts that stamped the attributes are not checked in.** One-shot migration mechanics; the tests print what they expect on failure, which is the maintenance path.

## Settled decisions carried into every reviewer prompt

- `ruled` — the residue countdown: bridge, counter, batch pass, closing condition at zero; `probeFiles` and `ciAllRows` gate, `probeLines` reports (Aaron, 2026-09-02: "agreed"); `guardFilesOutsideTheCount` gates (PR #476).
- `ruled` — the #470 class-stop: fix the live half in #476, charter the rest — this PR is the rest.
- `ruled` — new guards go in the test project; nothing leaves the old harness except by a conversion PR; shared fixtures frozen during the window.
- `ruled` — build agents are Opus 5 or lower, never Fable.
- `ruled` — a guard's subject set is derived from the surface it claims, never hand-enumerated.
- `ruled` — every shipped tool-name literal interpolates a `ToolNames` constant.
- `ruled` — an ADR says what is true of the tree it ships on; a stale sentence is amended by a dated superseding note in the PR that makes it stale.
- `inferred` — this PR's existence and its sequencing before the cut and the conversion lanes.
- `inferred` — the attribute's name, sorted-by-name order, the per-file baseline shape, the out-of-registry steps counted via the same attribute, the census invariant's home, the 2.0 wire capture as the rename oracle's shape.
- `inferred` — `ciAllRows` recorded per file, the fourth gated measure `standaloneCiSteps`, the second derivation of the roster population, and the reverse direction on the 2.0 capture. Reviewers were told they may re-litigate these four; none did.

## Open for your gate

| # | item | tag |
|---|---|---|
| 1 | **The class-stop ruling** — generate the workflow's standalone-guard steps from the attribute, state the bound, delete the parser. See [above](#open-1). The advisor's call within charter; you never spoke to it. Strikeable. | `inferred` |
| 2 | **The 2.0 capture's reverse direction** — `NothingIsPublishedThatTheCaptureDoesNotName_ANewToolAddsItsLiteralHere` goes beyond the charter, which asked only that every captured name be published. The build session flagged it as strikeable in one edit: delete that fact and adjust the file's `_maintained` key. | `inferred` |
| 3 | **`standaloneCiSteps` as a fourth gated measure**, and `ciAllRows` recorded per file rather than as a scalar. Both the build session's. | `inferred` |
| 4 | **The `TypeDeclaration` comment-matching sharp edge**, refused twice with reasons above. Loud-only, pre-existing, and every fix is another heuristic — but the conversion lanes will meet it. | refused, stated |

Everything else on this branch is folded, arm'd, and green.


src/housecarl-mcp-tests/RepoProjects.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        // Enumerated from the PROJECT directory — which exists by construction, its csproj having been read
+        // out of it — and then filtered to the bin tree. Enumerating bin/ directly threw
+        // DirectoryNotFoundException on a project that had never been built, which is the one case the
+        // refusal below is written for: the written sentence was unreachable and the reader got a bare path
+        // exception instead.
+        var binSegment = Path.DirectorySeparatorChar + "bin" + Path.DirectorySeparatorChar;
+        var dll = Directory.EnumerateFiles(directory, assemblyName + ".dll", SearchOption.AllDirectories)
+                           .Where(p => p.Contains(binSegment, StringComparison.OrdinalIgnoreCase))
+                           .OrderByDescending(File.GetLastWriteTimeUtc)
```
Avick3110, line 95: **BuiltAssembly can load a different configuration's build than the one under test.**

The search is `EnumerateFiles(directory, name + ".dll", AllDirectories)` filtered to a `/bin/` segment, then `OrderByDescending(File.GetLastWriteTimeUtc)` — so it picks the newest build output anywhere under the project, regardless of configuration. On the maintainer's own checkout right now:

```
src/housecarl-setup/bin/Debug/net9.0/houseCARL-Setup.dll   2026-08-27
src/housecarl-setup/bin/Release/net9.0/houseCARL-Setup.dll 2026-08-25
```

`dotnet test src/housecarl-mcp-tests --configuration Release --no-build` then does this: `CiProbeRosterTests.AttributedVerbsAcrossTheRepo()` runs `RepoProjects.AllAssemblies()` *before* it touches `CiAll` (line 50 evaluates before line 51), houseCARL-Setup is not loaded yet, and `Assembly.LoadFrom` pulls in the **Debug** dll — not the Release copy sitting in the test's own app base.

The second-order effect is the one that matters: `Assembly.LoadFrom` loads into the **default** ALC, so `CiAll.DiscoverAssemblies`' later `Assembly.Load(reference)` for `houseCARL-Setup` resolves to that same Debug instance. Both derivations — the runner's reference closure and the repo-wide csproj walk this class exists to be independent of — end up reflecting over one stale build. A `[CiProbe]` guard added to housecarl-setup and present only in the Release output is then invisible to `CiAll` *and* to `CiProbeRosterTests`, green, which is exactly the false-green the second derivation was written to catch.

It is also order-dependent: if any test touches `CiAll` first, the app-base Release copy wins and the same run behaves differently.

Preferring the dll beside the test assembly (`AppContext.BaseDirectory`) before falling back to the project tree would make this deterministic and match the configuration actually under test.

src/housecarl-generator/CiAll.cs
```diff
@@ -1,488 +1,172 @@
...
+                    && n.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)
+                    && !seen.ContainsKey(n))
+                    queue.Enqueue(Assembly.Load(reference));
+        }
+
+        return seen.Values.OrderBy(a => a.GetName().Name, StringComparer.Ordinal).ToArray();
+    }
+
+    /// <summary>Every attributed guard, roster and standalone alike, ordered by name.</summary>
+    static readonly Entry[] All = Discover();
```
Avick3110: **`Discover()`'s written refusals are thrown out of a type initializer.**

Both refusals in `Discover()` — the wrong-signature one and the verb-clash one — are carefully authored sentences, and this line makes them arrive wrapped. A `static readonly` field initializer runs in the type initializer, so the CLR wraps the `InvalidOperationException` in a `TypeInitializationException` **and caches it against the type** for the life of the process.

Concretely: add a second `[CiProbe("apply-guard")]` anywhere. In the generator that is one wrapped-but-readable stderr dump. In the xUnit process it is eight: `CiProbeRosterTests` (both facts), `CiWorkflowGuardStepTests` (both facts), `HarnessResidueTests`' three CiAll-backed measures and the bridge test all fail with `The type initializer for 'HousecarlGenerator.CiAll' threw an exception`, and the one sentence that names both hosts is buried in an inner exception repeated eight times.

This is the same defect `ecdbfa0` fixed one file over, and `RepoProjects.cs` documents it in so many words:

> A method rather than a property initializer on purpose: `BuiltAssembly` can refuse, and a refusal thrown out of a static initializer arrives as a TypeInitializationException that then poisons every other test touching this class for the rest of the run.

Same shape here — `All` could be a lazily-initialised method (`static Entry[] All() => _all ??= Discover();`) so the refusal reaches the caller as itself.

(`RepoProjects.All` at line 18 has the identical shape, with `Assert.True` inside `Discover()`. Its trigger — no csproj under `src/` — is not realistic, so it is a consistency note rather than a finding.)

src/housecarl-mcp-tests/ToolSurfaceCensusTests.cs
```diff
@@ -62,6 +66,178 @@ public void EveryDeclaredToolNamesItselfExplicitly_SoTheCensusHasSomethingToComp
...
+
+    /// <summary>
+    /// Whether a `typeof(<paramref name="named"/>)` expression names a type on the tool surface.
+    ///
+    /// <para>A bare name matches on the simple name. A qualified one has to be consistent with a real surface
+    /// type's full name, because a simple name is not unique across assemblies —
+    /// <c>typeof(HousecarlSetup.Program).Assembly</c> names a different project's <c>Program</c>, and matching
+    /// on the last segment alone reported it as an offender.</para>
+    /// </summary>
+    static bool NamesASurfaceType(string named, (HashSet<string> Simple, string[] Full) surface) =>
```
Avick3110: **The unqualified branch matches on simple name against every type in housecarl-mcp.**

The qualified branch is careful — the doc comment above names the `typeof(HousecarlSetup.Program).Assembly` case and requires consistency with a real full name. The bare branch does not have that protection: `surface.Simple` is every non-generic, non-compiler-generated type name in housecarl-mcp (~180 per the PR description), internal helpers included.

So the first time housecarl-core or housecarl-generator gains a type whose *simple* name collides with any of them, every `typeof(ThatType).Assembly` in that project is reported as "naming the tool surface by picking a type", and the failure message tells the author to read `HousecarlMcp.ToolSurface.Assembly` instead — which would change what the expression means. That is a nuisance red on a guard whose job is to be believed.

Narrowing the bare-name set to the types that could plausibly be meant (e.g. only `[McpServerToolType]`-carrying types, or public ones) would keep the arm's teeth without the collision surface.

src/housecarl-mcp-tests/ToolSurfaceCensusTests.cs
```diff
@@ -62,6 +66,178 @@ public void EveryDeclaredToolNamesItselfExplicitly_SoTheCensusHasSomethingToComp
...
+        // Vacuity canary: an empty name set would make every file below clean by arithmetic.
+        Assert.True(surfaceTypes.Simple.Count > 0,
+            "Reflection over ToolSurface.Assembly found no types, so nothing below can match and this arm " +
+            "passes over any spelling. The reflection is broken, not the source.");
+
+        var files = RepoProjects.All
+            .SelectMany(p => Directory.EnumerateFiles(p.Directory, "*.cs", SearchOption.AllDirectories))
+            .Where(NotBuildOutput)
+            // The one home itself: ToolSurface.cs is where this fact is allowed to be spelled.
+            .Where(f => !string.Equals(Path.GetFileName(f), "ToolSurface.cs", StringComparison.Ordinal))
```
Avick3110: **The one-home exemption is by bare filename, not by the one home's path.**

This filters out *any* file called `ToolSurface.cs` anywhere under any project in `src/`. Add `src/housecarl-generator/ToolSurface.cs` (or `src/housecarl-core/ToolSurface.cs`) containing `typeof(ReadTools).Assembly` and it never enters `files`, so the offender is never seen — the guard whose subject is "this fact lives in exactly one place" is defeated by giving the second place the same filename.

The home is already knowable without a convention: `RepoProjects.DirectoryFor(ToolSurface.Assembly.GetName().Name!)` plus the filename, compared as a full path. That keeps the derived-both-sides posture the rest of this arm has.

```suggestion
            .Where(f => !string.Equals(
                f,
                Path.Combine(RepoProjects.DirectoryFor(HousecarlMcp.ToolSurface.Assembly.GetName().Name!),
                             "ToolSurface.cs"),
                StringComparison.OrdinalIgnoreCase))
```

Avick3110:
Gate findings folded — one commit each, each with an executing arm RED-checked from a committed state (sabotage → RED → restore, grep-verified).

- **3922536403** (RepoProjects.cs:75, app-base preference) — `bc98eb6`. Arm: `RepoProjectsTests.TheDllBesideTheTestAssemblyWinsOverANewerOneInTheProjectTree_TheBuildUnderTestIsTheOneReflected` (+ a fallback cell). The choice is its own method now (`LocateAssembly`); the fixture plants an OLDER dll beside the test assembly and a NEWER decoy under a project tree's bin/Debug, asserts the app-base path is chosen, and asserts nothing of that name was loaded.
- **3922538326** (CiAll.cs:70, refusal out of a type initializer) — `f908511`. Arm: `CiAllRefusalTests.NoStaticOnCiAllIsFilledByTheTypeInitializer_SoARefusalReachesTheCallerAsItself` (the RED/GREEN one, derived: no `initonly` static may exist on CiAll), plus `TwoGuardsClaimingOneVerbRefuseWithBothHostsNamed_NotAsATypeInitializerFailure` and `AnEntryPointTheRunnerCannotDispatchRefusesWithItsOwnSentence` over an emitted fixture assembly through a new `RosterIn(assemblies)` seam. `ProbeHosts`, `StandaloneProbeHosts`, `ProbeNames`, `StandaloneProbeNames`, `GuardTypes` and `GuardAssemblies` were property initializers too and run in the same initializer, so they cache on first read as well — leaving any eager would have kept the wrapping. `RepoProjects.All` got the same shape (the consistency note).
- **3922540601** (ToolSurfaceCensusTests.cs:171, bare-name set) — `c6e3a73`. Arm: `ABareNameIsReadAsTheSurfaceOnlyWhenItCouldMeanIt_AnInternalHelpersNameIsNotThatSentence`. Bare names now match `[McpServerToolType]`-carrying types plus `ToolSurface` itself: 25 names, down from 182. **Measured cost, flagged rather than folded silently:** of the six discarded spellings the tip commit repointed, the narrowed reading still catches two (`typeof(HousecarlMcp.ReadTools)` qualified, and bare `typeof(ReadTools)`) and would no longer catch four — three `typeof(ApplyOp)` (public record, no attribute) and `typeof(ToolSchemas)` (internal helper), the latter being PreFlattenSurface's second `WithToolsFromAssembly` call site. Both narrowings you named drop `ToolSchemas`; only this one also drops `ApplyOp`. If you want the teeth back without the collision surface, the alternative is to keep the full simple-name set and skip a bare name only when the file's own project also declares a type of that name — derived on both sides, and it removes exactly the nuisance-red class you described. Say the word and it is one more commit.
- **3922540744** (ToolSurfaceCensusTests.cs:191, one-home exemption) — `3ba0937`. Your suggestion, as given. Arm: `ASecondFileCalledToolSurfaceCsIsStillScanned_TheExemptionIsAPathNotAFilename` plants a second `ToolSurface.cs` in another project's directory (both sides derived from `RepoProjects`), asserts the scan reports it, removes it in a `finally` and asserts its absence.

Gate from the tip (`3ba0937`): `dotnet build housecarl.sln -c Release` 0 errors; `ci-all` 132/132; `dotnet test src/housecarl-mcp-tests` 390/390 (383 before, +7 new cells).

Avick3110:
Follow-up on **3922540601**: fold 3 revised in `75ece6b` — the full simple-name set is back and all eight repointed sites are caught again. A bare name is skipped only where the scanned file's own project declares a type of that name (file → project off `RepoProjects.All`; that project's names off its built assembly), which answers the collision without costing teeth. Files in the tool-surface project get an empty local set, so bare spellings inside housecarl-mcp stay caught. Arms: `EverySpellingThisBranchRepointedIsStillCaught_TheCollisionFixCostsNoTeeth` (RED under the narrowed set: 4 of 8 — the three `ApplyOp` sites and `ToolSchemas`) and `ABareNameIsSkippedOnlyWhereTheFilesOwnProjectDeclaresIt_NoNuisanceRedNoLostTeeth`. Gate from `75ece6b`: build 0 errors, ci-all 132/132, `dotnet test` 391/391.

Issue #483, opened 2026-09-03
[GAP] The executing-arm population is derived, not hand-swept — every prose site interpolating a ToolNames constant is the subject set


Avick3110:
## The gap

RUN_ORDER amendment **2026-09-02 (2)** ruled the cure for the remedy-names-a-broken-call class: *a remedy that names a call ships with an arm that MAKES the call as the sentence tells the caller to, against a fixture, asserting it is not refused.*

The **posture** is ruled. The **population it applies to is not derived**. Nothing enumerates "every shipped prose site that names a call", so each PR hand-sweeps a subset and is short by exactly what nobody thought of — CLAUDE.md §3's hand-wired-coverage failure at the sentence layer.

## Measured evidence that a hand sweep is short

On the cut PR (#468), the review session swept the chain and scan lanes by hand, repaired six sentences and shipped executing arms for them. Two independent round-2 reviewers then derived and **drove** a further **seven** sentences of the same class — `housecarl_records source="…"` with no select term, a call the tool refuses — two of them in files the fold had edited in the same commit. One of the seven was a **published tool description**. A third instance family (the copy read-back remedy) named a call that cannot answer its own question at all.

Same class, two consecutive rounds, six instances. That is a class-stop under CLAUDE.md §5 #11, and the seam is this missing derivation.

PR #474 met the same seam and recorded the measured negative that makes it non-trivial:

> the correct oracle is not "is this spelling declared anywhere" but "is it declared by the tool **this sentence is reachable from**" — a per-sentence reachability question an executing arm answers by construction and no containment net can.

## The proposed subject set

**Every shipped prose site that interpolates a `ToolNames` constant.** That population is derivable today: it is the registry's class-(b) sites, and `scripts/tool-names/derive-tool-name-sites.py` already carries the classification that produced them. A sentence that names a call names the tool, and after ADR 0004 it names the tool through a constant — so the constant is the hook the derivation hangs on.

The guard then asserts each such site has an executing arm: a test that makes the call the sentence describes and asserts it is not refused. Sites whose sentence names no call are classified out explicitly, with the classification reviewable rather than implicit.

## First cells

- the **ten sites folded on the cut PR** (seven `source=`-without-select sentences, the three copy read-back siblings) — each already has an executing arm, so they are the worked examples;
[...]

Issue #482, opened 2026-09-03
[GAP] The alias-activation census's exact-set half is not derivable and is unguarded after the cut


Avick3110:
## The gap

The alias-activation census's **exact-set half** is not statically derivable and is unguarded after the cut.

`BindingShimProbe`'s census was a 144-row eyeballed literal guarding two failures:

1. **A row that silently fails to fire.** This half *is* derivable, and it is ported by the cut PR: it is computed by running the shipped `ResolveAliases` over the surface's own schemas and then driven over the wire, which is strictly more than the literal did.
2. **"No row fires anywhere unintended."** This half is **not derivable** — "unintended" was the human judgement the list encoded. There is no oracle in the code for it.

**Until this is closed, an activation landing on a tool nobody checked is unguarded.**

## What closing it honestly looks like

A generated-and-reviewed snapshot: the derived set is written to a data file by an update run, the file is checked in, and it is diffed in review. That makes the human judgement a reviewed artifact rather than a hand-kept literal, and it puts the "unintended" half where a person can actually see a change to it.

## Why it is not a fold on the cut PR

This is guard growth, not a conversion. The cut PR (#468) converts what the old probe asserted; inventing a new snapshot mechanism is a separate change with its own design. It is recorded in that PR's arms-owed list and filed here as its own issue.

## Provenance

Derived and measured by the #468 build session; recorded in that engagement's decision record (§3, "Owed, and why"). Filed by the #468 review session as the issue that record owes.

Routing and milestone are Aaron's; filed unmilestoned.

Avick3110:
Routed 2026-09-03 (Aaron: "I agree with the rest of the routing proposals"; ground: the row-8 disposition walk, advisor lane)

**Routed to RUN_ORDER row 9, deliberately unmilestoned** — a W7 acceptance item, beside #426 / #444 / #447. Per amendment 2026-09-02 (6), the guard-shaped cluster is held in row 9 as W7's **acceptance items, not preconditions**: W7 starts regardless and is not complete until each is closed by construction (with the test that proves it) or has its remaining work named; no work on them until a conversion PR touches the guard each describes.

Boot rule 4 keeps it unmilestoned: post-2.0 / acceptance-held issues carry no milestone.

Row 9's `Issues routed` cell is refreshed in the same stroke.

Avick3110:
[...]

PR #484, opened 2026-09-03
The cut: the eleven remaining 1.x read and check tools are deleted, their probes with them, and the surviving engine behaviour is tested in the test project

### LINKED ISSUE #468: [NOTE] Rolling-demolition catch-up: delete the 1.x tools whose 2.0 replacements shipped in W1-W3
The rolling-demolition rule (RUN_ORDER boot rule 10(e)) says each PR shipping a 2.0 replacement deletes the 1.x tools it replaces in the same PR. Waves W1–W3 landed their replacements before that rule existed, so their predecessors are still registered. This is the catch-up: the tools whose replacements have already shipped and been verified.

**Population, derived from the SPEC's absorption map against the registered surface** — never hand-enumerated (rule 10). Source of truth: `[McpServerTool(Name = …)]` attributes in `src/housecarl-mcp` (51 tools registered), mapped through SPEC §6.1/§6.2.

| Family | 1.x tools deleted | 2.0 replacement (shipped + verified) |
|---|---|---|
| read (8) | `read_record` `batch_record_detail` `resolve` `read_plugin_file` `cross_plugin_query` `diff_record` `effect_chain` `skypatcher_read` | `records` — including `project=info_order` and `source={overlay, state}` |
| check (3) | `check_errors` `validate_scripts` `validate_dialogue` | `check` (`findings=`), plus `validate_dialogue`'s class-8 render on `records` |
| write (6) | `set_field` `bulk_apply` `create_record` `bulk_create` `remove_record` `forward_record` | `apply` `create` `remove` `forward` |

**Not in the population, each for a stated reason:**

- `copy_npc_appearance` — gated on #387. Until that resolves, the two-donor carry exists only in the ancestor. It dies in the PR that ships `copy`'s final form (row 10, retimed by amendment 2026-08-31 (5)).
- `place_asset` / `bulk_place_asset` — their replacement (`place`, SPEC §6.2) is unshipped W4 work. Not a catch-up item.

## Shape: two PRs

Ruled by Aaron 2026-09-01 (RUN_ORDER amendment 2026-09-01 (5)) after a §4(b) escalation: rule 10(e)'s "small standalone" was the recording session's paraphrase, and measurement falsified it — roughly 340 probe touch points plus 58 user-facing sentences in *surviving* code naming a tool about to stop existing.

- **PR 1 — write + check families.** 9 tools, 25 CI probe sites, two hand-run harnesses that die wholesale, 33 sentences.
- **PR 2 — read family.** 8 tools, 121 CI probe sites, plus the stdio binding shim.

Each PR is rule-9 seamed (build session ends at branch-green; a fresh session runs the review rounds). PR 1 references this issue; PR 2 closes it.

## Why the probe work is per-assertion, not wholesale

Every affected CI assertion was classified into two classes before any code moved:

- **dies-by-design** — the assertion's subject is a 1.x-only construct with no spelling on the 2.0 surface (e.g. `operations` XOR `from_file`: SPEC §5.1 retires `from_file` into the `@file` convention, and "the inline-vs-file mutual exclusion has no spelling when they are one parameter"). Deleting it loses nothing. Each such row carries its SPEC citation.
- **repoint-to-2.0** — the assertion's subject is engine or transport behaviour that survives. Deleting it drops coverage silently.

The classification found **eight live behaviours in the two smallest families guarded only through a dying tool**, all of which still exist on the 2.0 surface:

| behaviour | guarded only in | survives on |
|---|---|---|
| `@file`: relative path refused | `DryRunProbe` (dies with `bulk_apply`) | `apply` `ops="@<path>"` |
| `@file`: unreadable file refuses naming the path | same | same |
| `@file`: non-array root refuses naming the shape | same | same |
| `@file`: empty array refuses naming the file | same | same |
| `exclude=`: `implicit` token reads the MO2 composition | `CheckErrorsProbe` tool-layer arms | `check exclude=` |
| `exclude=`: `base_masters` group token | same | same |
| `exclude=`: read-failure refusal | same | same |
| `exclude=`: unknown group refuses naming the real token set | same | same |

A wholesale probe delete would drop all eight with `ci-all` staying green — the "verification net fails toward green" shape. Each is repointed and mutation-tested RED instead.

The read family (PR 2) carries the larger instance: of `BindingShimProbe`'s 25 live stdio calls, 21 drive a doomed tool and only 2 drive `records`. That guard exists because of a real production failure (HCBR-2026-06-11-01 — `plugins=` sent as a bare string; the SDK genericised the throw and the calling agent abandoned the documented query path).

## Also in scope, and why

- **Sentence repair in surviving code.** 58 user-facing sentences in engine and 2.0-tool code name a tool that will not exist — e.g. `DialogueCkParity.cs` telling a caller to reach for `housecarl_set_field`, `WriteEngine.cs` offering `housecarl_remove_record` as a remedy. Left alone these become remedies pointing at nothing. Repairing wreckage the cut itself causes is part of the cut; each repair composes per the SPEC's own posture, never through the 1.x renderer (rule 10(d)).
- **`EMPIRICAL_LEDGER` EL-3** names `bulk_create` as its test vehicle. `create` drives the same engine cleave, so the line's spelling updates; the debt is unchanged and stays `OPEN`.

## Deliberately untouched

- **Bundled skills** (17 files under `.claude/skills/`, plus `plugin/codex/housecarl/SKILL.md`) name doomed tools. They are rewritten against the shipped surface at W6 by construction (row 10), and rule 10(f) means no 1.x release ships in between, so the dev tree is knowingly inconsistent until then.

## Close condition

PR 2 merged, with the read family's tools deleted and its probe dispositions recorded. The row-10 walk then confirms the deletion population is empty rather than performing a cut.

Filed by the row-7 build session at kickoff per rule 10(e). **Routing into RUN_ORDER and setting the milestone are one act under boot rule 4 and are Aaron's — this issue is deliberately filed unmilestoned.**

Avick3110:
STACKED on `claude/roster` (PR #481); merge that first — this PR retargets to `main` and rebases when it lands.

Deletes the eleven remaining 1.x tools whose 2.0 replacements have shipped — eight read, three check — together with the guard arms whose subject was one of their responses, and lands fresh tests in `src/housecarl-mcp-tests` for the behaviour that survives. The wire goes **46 → 35**.

---

# The three ruled items — all three disposed of

**No item on this PR remains open for ruling.** Item 2 (L1) and item 3 (L2) are folded — see their sections. Item 1 (A) was ruled, executed against the compiler, measured at 84% where its gate was a third, and then **ruled a follow-up** on 2026-09-03; ADR 0005 is corrected on this branch to say what is true of the tree, and the deletion and its 147 arms are chartered to issue #486 as the next two PRs. What is left for Aaron here is his gate review, not a decision he still owes.

## 1. The tools are gone from the wire; their render and service halves are not — RULED 2026-09-03 — FOLLOW-UP (issue #486)

> **RULED 2026-09-03 — follow-up.** Aaron: *"agree, its a follow up"* — on the advisor's recommendation of two sequenced PRs landing right after this one and before #440: a **fixtures PR** (the Papyrus fixture and the file-lock harness the test project does not have), then the **render-halves PR** (delete the eight members, let the compiler name the 197 sites, write the 147 arms, re-measure the two cap/floor baselines). Filed as **#486**, unmilestoned — routing and milestone are Aaron's.
>
> **Landed on this PR instead:** `1fca938` rewrites the two ADR 0005 sentences that were false of this tree, so the ADR states what holds here — the eleven tools and their tool-layer probes are gone, eight render and service members survive with no shipped caller, 197 old-harness assertions drive them, and the residue countdown carries the seven probe files that stayed with their line counts. Nothing of the deletion itself is on the branch: it was made in a throwaway tree, measured, and reverted to the committed tip.
>
> **How it got here.** The ruling that opened this item directed that class (b) — arms asserting an ENGINE fact reached only through the dead render — be counted before any fresh test was written, with the turn to end if (b) exceeded about a third. **It is 84%**, so the turn ended and the measurement went back for a ruling. The measurement is below and stands whichever session picks up #486.

### The derivation, re-run — and it is EIGHT members, not nine

Re-derived by script over `src/housecarl-mcp` + `src/housecarl-core`, printing every word-boundary hit with `file:line` so a declaration, a comment and a call are told apart by reading; run on `claude/cut` @ `8ebb03c` and on the base `claude/roster` @ `9adf61d` as the control. Eight of the nine confirm — zero shipped callers on the branch, a real shipped caller on the base.

**`LoadOrderService.ValidateDialogue` is not dead and stays:**

```
LoadOrderService.cs:2293  public DialogueCheckResult CheckDialogue(IReadOnlyList<string>? seeds, int limit, bool countsOnly = false)
LoadOrderService.cs:2294      => DialogueSweep.Run(ValidateDialogue, seeds, limit, countsOnly);
```

`CheckDialogue` is called from `CheckTools.cs:202` — the surviving `housecarl_check findings=["dialogue"]` family — and the same call is on the base, so the member's caller count did not move at the cut at all. The escalation report explicitly corrected the gate-replica reviewer on this point; **the reviewer was right.** The cause is visible in the numbers: that script excluded the declaring FILE rather than the declaring LINE, which hides a same-file caller. `ValidateDialogue` is the only one of the nine with one.

**A tenth zero-caller member the table does not name:** `LoadOrderService.ReadPluginFile` has zero shipped callers on `8ebb03c` — only its declaration. The decision record's "the ENGINE is intact" is true of the code and not of its reachability. Not one of the nine, not touched, stated because item 2's ruling pointed at it as the home of a computation.

### The compiler's list, and the classification

The eight were deleted (**657 lines of shipped code**) and the solution built. The compiler named **197 distinct sites, every one in `src/housecarl-generator`** — none in `housecarl-mcp`, none in `housecarl-core`, none in `housecarl-mcp-tests`. That is the zero-caller claim confirmed a second way, by the compiler rather than by a script.

Three independent read-only agents classified every named site, each given the surviving-surface map — that `Wire.RenderCheck` / `JsonWire.RenderCheck` still call the same `CheckAccounting`, `BoundedBody`, `AppendErrorsHead`, `AppendErrorsSection`, `ComposeErrorSection`, `AppendScriptsHead`/`Section` and `AppendExcludedPlugins` the deleted entry points called:

| file | sites | (a) dies | (b) owes a fresh test | *of which already covered* | (c) diagnostic only | arms named | arms untouched |
|---|---|---|---|---|---|---|---|
| `CheckErrorsProbe.cs` | 110 | 0 | 101 | 2 | 9 | 72 | 56 |
| `EpochGuardProbe.cs` | 29 | 17 | 12 | 0 | 0 | 28 | 45 |
| `ScriptPropertyCheckProbe.cs` | 28 | 0 | 27 | 0 | 1 | 15 | 25 |
| `BulkPrimitivesWave3Probe.cs` | 18 | 3 | 14 | 7 | 1 | 21 | 68 |
| `DialogueInfoOrderProbe.cs` | 6 | 0 | 6 | 0 | 0 | 9 | 17 |
| `DialogueValidateGuardProbe.cs` | 3 | 0 | 3 | 0 | 0 | 1 | 41 |
| `CheckMergeProbe.cs` | 2 | 0 | 2 | 0 | 0 | 1 | 45 |
| `SkyPatcherHarness.cs` | 1 | 1 | 0 | 0 | 0 | 0 (no arms in the file) | — |
| **TOTAL** | **197** | **21** | **165** | **9** | **11** | **147** | **297** |

- **(a) DIES** — the assertion's subject is the deleted tool's own response and nothing surviving carries it.
- **(b) OWES** — the assertion itself calls a deleted member, and what it asserts is engine or shared-machinery behaviour a surviving tool still drives.
- **(c) DIAGNOSTIC ONLY** — the dead call is in the `Check(...)` failure-detail string or a `Console.WriteLine`, never in the boolean. Nothing dies, nothing is owed.

**(b) is 165 of 197 — 84%, against a bound of about a third.** Net of the nine already covered by existing tests, **156 sites owe a fresh test, across 147 distinct arms.**

### Why, stated plainly

The escalation assumed these arms "assert the deleted tools' response format". Measured, they mostly do not. `housecarl_check` renders the same `ErrorCheckResult` and `ScriptCheckResult` through a different entry point into the **same** accounting, budget, section-composition and honesty machinery. An arm asserting that an excluded finding class renders "NOT CHECKED" rather than 0, or that a response's own accounting numbers agree with the document it printed, is asserting behaviour that is live today under `check` and reached only through the entry point being deleted. The clean (a) clusters are `EpochGuardProbe`'s diff-family arms (the pairwise-diff SERVICE, gone entirely) and the `#271` / `read_plugin_file` banner and JSON keys, which no 2.0 surface emits. Worth noting: **no arm in `CheckErrorsProbe` asserts the deleted banner** — the one string that unambiguously died there is asserted by nothing.

### What the fresh tests would actually cost

The sites collapse onto far fewer distinct facts than 156 — roughly seventeen in `CheckErrorsProbe`, twenty-seven arms across epoch + scripts, about eleven new test bodies for the rest. Most of it is transcription. Three items are not:

1. **`src/housecarl-mcp-tests` has no Papyrus fixture at all** — nothing there references `Pex`, `VirtualMachineAdapter` or `ScriptEntry`. Every scripts-family arm needs the probe's `.pex` writer and its child/base property pair ported first.
2. **No file-lock harness** — three dialogue arms fault a read by holding a file open; the test project has nothing for that.
3. **Two re-measurements, not re-typings.** The cap/floor sweeps measure a floor the merged surface changes (its fixed part now carries the title, scope sentence, three family lists and a per-family boundary), and `CheckMergeProbe`'s #394 allocation arm takes its baseline from the deleted renderer — re-deriving it from the merged one is a design call. Every hand-shaped `ErrorCheckResult` fixture also needs a `CheckSweep` wrapper and every JSON assertion needs repathing from the root to `families.errors.*`; one of them (`JSON-INBAND-FIELDS-ABSENT-UNDER-COUNTS-ONLY`) goes **vacuously green** if that repathing is skipped.

### Two findings that outlive the stop

- **Shipped cascade** the compiler cannot name (private members and type declarations): `Wire.PoleLine`, `JsonWire.WriteEpochWithCoverage`, `LoadOrderService.ResolveDiffPole`, `DiffRecordOutcome`, `SkyPatcherPostStateData`. Surviving and NOT to be deleted: `DiffPole` (live via `DeltaRow` / `JsonWire.WriteDiffPole`), `SkyPatcherFolderOutcome` (live via the layer replay), `WriteSweepEpoch` (live in `JsonWire.RenderCheck`).
- **The `conflictTree` parameter item 3 deferred to this ruling.** With `Wire.RenderRecord` gone, `AppendConflictTree` is reached only through `AppendRecordBlock`'s `if (conflictTree)` branch, and all five shipped call sites (`RecordsTools.cs:601, 1240, 1276, 1452, 1488`) pass `false`. The render-layer chain — `RenderBatch` x2, `RenderCrossQuery` x2, `AppendRecordBlock` — and `AppendConflictTree` go with it. The SERVICE-level `ResolveRead(..., conflictTree:)` keeps its live `true` caller in `SnapshotViewProbe` and is untouched.

### ADR 0005 — corrected on this PR (`1fca938`)

The two sentences that were false of this tree are rewritten. Both had said the deletion finishes in one change: *"Guards on a deleted tool die with the tool … in the same change"*, and *"only its tool-layer block is removed"*. The ADR now keeps the rule and names this change as its counter-example — eight render and service members survive with no shipped caller, 197 old-harness assertions drive them, the classification is stated (21 / 165 / 11, 147 arms owed), and the finishing work is chartered to **#486** as two changes, fixtures first. It also records that the residue countdown carries the seven probe files that stayed, by name and line count, so the arrears are a number in the checked-in baseline rather than something to remember.

Every other count in ADR 0005 was re-checked against the tree at the same time: 45 published 1.x tools (from the frozen `tools-list-1.9.0.json` capture), `housecarl_records` absorbs eight read tools, `housecarl_check` two more plus part of a third, four write tools absorb six (corrected from five in round 2), plugin version still `1.9.0`, and `housecarl_copy_npc_appearance` still the one chartered survivor. No count needed changing.

<details>
<summary>The escalation and the three options, as they stood before the ruling</summary>

Nine members are compiled and shipped in `housecarl-mcp` with **zero callers in shipped code**, and had a shipped caller on the base. Their only remaining drivers are old-harness probes.

Derived by script over `src/housecarl-mcp` + `src/housecarl-core` with declarations excluded, and re-derived independently by the review session with the base counted as the control:

| member | shipped callers on cut | shipped refs on base | probe refs |
|---|---|---|---|
| `Wire.RenderCheckErrors` + `JsonWire.RenderCheckErrors` | **0** | 2 | **121** |
| `Wire.RenderScriptCheck` + `JsonWire.RenderScriptCheck` | **0** | 2 | **31** |
| `LoadOrderService.DiffRecord` | **0** | 1 | 15 |
| `DialogueWire.Render` | **0** | 1 | 9 |
| `Wire.RenderDiffRecord` + `JsonWire.RenderDiffRecord` | **0** | 1 | 8 |
| `Wire.RenderPluginFile` + `JsonWire.RenderPluginFile` | **0** | 1 | 8 |
| `LoadOrderService.ValidateDialogue` | **0** | 1 | 5 |
| `Wire.RenderRecord` + `JsonWire.RenderRecord` | **0** | 1 | 4 |
| `LoadOrderService.SkyPatcherPostState` (+ `…Data`) | **0** | 1 | 1 |

**≈190 probe assertions across nine surviving probe files** — `CheckErrorsProbe` 110, `ScriptPropertyCheckProbe` 28, `EpochGuardProbe` 24, `BulkPrimitivesWave3Probe` 8, `DialogueInfoOrderProbe` 6, `DialogueValidateGuardProbe` 3, `DialogueSubtypeMarkerGuardProbe` 3, `DialogueCkParityGuardProbe` 2, `CheckMergeProbe` 2, plus `SkyPatcherHarness` 1.

These are not engine arms behind a convenient render. They assert the **deleted tools' response format**: `Wire.RenderCheckErrors` opens `"check_errors — load-order integrity sweep\n"`, `RenderScriptCheck` opens `"validate_scripts — VMAD script-property binding sweep\n"`, `RenderPluginFile` opens `"read_plugin_file — OUT-OF-LOAD-ORDER …"`, `DialogueWire.Render` emits `"validate_dialogue: …"`. The live `housecarl_check` renders through a different entry point (`Wire.RenderCheck`, from `CheckTools.cs:167/205`); the 2.0 record reads go through `RenderBatch` / `RenderCrossQuery` / `RenderResolve`.

**Two sentences in ADR 0005, shipping on this PR, are currently false about the tree they ship on:**

> **Guards on a deleted tool die with the tool; behaviour that survives gets a fresh test in the same change.** … So the arms are deleted, and each one whose behaviour the successor still carries is written fresh as a test against that successor, in the same change — never left as a promise for later.

> **The old guard harness shrinks by whole files, and only where the file's whole subject went.** A file that also guards engine behaviour keeps that half; its residue entry stays, and only its tool-layer block is removed.

The 152 `RenderCheckErrors` / `RenderScriptCheck` arms **are** the tool-layer block, and they stayed. Whichever option is chosen, those two sentences are rewritten before merge — an ADR says what is true of the tree it ships on.

**How it surfaced.** Both round-1 reviewers found it independently, from opposite ends; the gate-replica reviewer, seeded with nothing but the branch and the base, graded it high. It is also a landed re-litigation of the branch's own `inferred` decision — the settled list names the arm-level reading of "the probes die with the tools" as *the largest judgement call on the branch* and invites a reviewer to re-open it.

### The three ways out

**(A) Delete the halves and their arms now.** Honours the ruling literally. Costs a build engagement of branch-comparable size — ≈190 assertions classified per-assertion, with fresh tests for whatever among them is genuine engine behaviour — and it takes those arms out of the old harness outside a conversion PR, which cuts against the standing `ruled` rule that nothing leaves the old harness except by one. The `RenderPluginFile` half cannot be executed until item 2 below is ruled, because that renderer *is* item 2's code.

**(B) Defer the halves to the chartered conversion PRs, and make ADR 0005 say so** — naming the surviving population and its counts, so the residue is stated rather than silent, with the same table in this body. Consistent with amendment 2026-09-02 (7) (*W7 proper converts ON TOUCH, not wholesale*) and with the conversion-PR rule. Costs a weaker ADR claim and ≈190 arms reporting green on renders no caller reaches until their family converts. Lands naturally as a row-9 / W7 acceptance item.

**(C) Split by size.** Delete now the small, uncoupled members (`SkyPatcherPostState`+`Data`, `Wire.RenderRecord`, `RenderDiffRecord` + `LoadOrderService.DiffRecord`, `DialogueWire.Render`, `ValidateDialogue` — ≈42 arms in five probe files); defer `RenderCheckErrors` (121) and `RenderScriptCheck` (31) to their families' conversion PRs; hold `RenderPluginFile` pending item 2.

**Two recommendations, and they differ.**

- **The review session recommends (B).** (C) draws the population line by arm count rather than by the surface it claims, which is the hand-drawn-population failure this project keeps finding at guards. (A) is the honest literal reading, but it is a second engagement and it is partly blocked.
- **The advisor recommends (A), executed with the compiler as the checklist** — deleting the nine zero-caller members makes the compiler name every arm that called them, so the population is classified by construction rather than by anyone's judgement. That is the same mechanism the cut itself ran on ("compile errors are the checklist"), which is a real argument that the work is smaller than the assertion count suggests.

</details>

## 2. L1 — the declared-master advisory loses its install-vs-enable distinction — RULED 2026-09-03, FOLDED

> **RULED 2026-09-03 — compensate on `check`; folded (`47019de`).** Aaron: *"I agree with the rest of your calls"*.

**As executed.** `housecarl_check findings=["missing_masters"]` keeps ONE finding class, one count and one list. Its **remedy** now names, per unsatisfied master, which case holds:

```
missing master(s) NOT installed anywhere in the MO2 install: HcMmAbsent.esm   [install them — this plugin's refs into them dangle until you do]
missing master(s) installed but NOT ACTIVE in the load order (in a disabled mod, or unchecked): HcMmGhost.esm   [enable them — this plugin's refs into them dangle until you do]
```

replacing the one line that said `install/enable it` for every master at once. No new finding class, no new lever, no new parameter; the JSON lane is untouched.

**The computation gets one home.** The ruling said to re-home the split if its only home was the deleted render. It was not in the render — `LoadOrderService.ReadPluginFile` computed it inline off `Mo2LoadOrder.PluginFileExists` — but that method has no shipped caller (item 1), so its only home was itself unreachable. So: `Mo2LoadOrder.SplitUnsatisfiedMasters` is new and is the one home for the discriminant; `ReadPluginFile`'s inline loop now calls it, so the two surfaces cannot decide the same master differently; `PluginErrors` gains `IReadOnlyList<string>? InstalledButInactiveMasters` — the **subset** of the masters the sweep already reports, so no count and no list moves; `LoadOrderService.ClassifyMissingMasters` fills it in where the MO2 composition lives, keeping the core free of MO2 plumbing; and `Wire.ComposeErrorSection` splits the remedy when the subset is present. **null means not classified** — the render then prints the union sentence it printed before, rather than claiming every master is uninstalled. That is the codebase's existing null-is-not-empty convention (`Histogram`, `DanglingBySource`).

**The arms** — nine, in `src/housecarl-mcp-tests/CheckMasterRemedyTests.cs`, over a new synthetic MO2 instance carrying one master of each standing: `Skyrim.esm` satisfied (in `loadorder.txt`, absent from `plugins.txt`); `HcMmGhost.esm` installed but in a mod `modlist.txt` switches OFF; `HcMmAbsent.esm` written outside every MO2 layer. They assert that the absent master is named on the install line and not the enable line; the disabled-mod master on the enable line and not the install line; that the union remedy is not printed where the split was made; that the satisfied master is on neither line; that a plugin whose masters are all satisfied gets no remedy line while the head still reports `0 missing master(s)` (so the two absences are an answer, not a skip); that the class still reports both as one count; that the subset is exactly the installed one; that an unclassified report keeps the union remedy; and that `read_plugin_file`'s advisory and the check remedy agree on which master is which over the same fixture. Every expected sentence is a fixture-known value naming the master it must appear beside, so a remedy printed against the wrong master fails on the name, not merely on the wording.

**One deviation from the ruling's letter, stated rather than substituted.** It said to drive the arms "through `housecarl_check` off the built server". That cannot be done honestly here: `ServerFixture` runs an **unconfigured** server — its own summary says the config check answers before any argument value is interpreted — so no stdio-driven call reaches a sweep at all, and pointing that fixture at an instance would modify a frozen shared fixture, which the same instruction names as a STOP. The arms drive `CheckTools.CheckTool` instead — the method the MCP server publishes and binds arguments into, and the lane this branch's own `CheckExcludeGrammarTests` calls "the surface a caller reaches". Not a render call.

**Sabotage, from the committed state, both branches of the conditional.** (1) The two remedy sentences SWAPPED: both mapping arms RED, the other seven including the all-satisfied control green — the sabotage bit exactly the cells that claim the mapping. (2) The union remedy corrupted (the null branch): the unclassified arm RED, alone. Both restored with `HOUSECARL_ALLOW_DIRTY_RESTORE`, each restore grep-verified in both directions plus `git diff HEAD` empty.

**Gates from the committed tip `47019de`:** build 0 errors; `dotnet test` 824 → **833**; `ci-all` **130/130 ALL PASS** (no probe file touched); `tools/list` **35**, wire canonical size **177 099 chars** — byte-identical to `8ebb03c`, so no survivor description changed.

**In `plugin/CHANGELOG.md`:** the cut's `## Unreleased` entry gains one paragraph stating that the remedy now names install vs enable per master, that the class, count and list are unchanged, and how to check it.

<details>
<summary>The escalation as it stood before the ruling</summary>

`housecarl_read_plugin_file`'s render printed two different lines off `PluginFileOutcome`:

- `! declared master(s) NOT installed anywhere in the MO2 install … (install them …)`
- `! declared master(s) installed but NOT ACTIVE in the load order (in a disabled mod, or unchecked) … (enable them …)`

The second is the PR #148 false-negative fix: a master present only in a disabled mod is not satisfied, and its remedy is *enable*, not *install*.

Measured: `LoadOrderService.ReadPluginFile` still computes both sets, so the **engine is intact**; `Wire.RenderPluginFile` / `JsonWire.RenderPluginFile` are their only shipped readers and have no shipped caller after the cut (`RecordsTools`' off-order lane renders through its own path and emits neither); `housecarl_check findings=["missing_masters"]` still reports unsatisfied masters, but `ErrorFindingClass.MissingMasters` is defined as "a master not present in the ACTIVE order", which is the **union** of the two. The distinction, and the two different remedies hanging off it, have no expression on the 2.0 surface.

The build session measured this; a round-1 reviewer who had never been told about it rediscovered it independently and graded it high, so it is not an artefact of how the question was framed.

</details>

## 3. L2 — the precise owned-child tier is unreachable — RULED 2026-09-03

> **RULED 2026-09-03 — accepted; tier deleted (`918a127`); issue #485.** Aaron: *"agree accept, delete and file the issue"*.

**As executed.** The tier is gone from the tree: `LoadOrderService.ResolveTreeFill`, `TreeFill` and `ChildDeclarers`; `Wire.ApplyPreciseChildNotes` and the `conflictTree == true` branch in the shared render that called it; and `ReadSentences`' whole tier-2 block — `DeclaredBy`, `CarriedBy`, `CouldNotRead`, `MergeCollection`(+`Framing`), `SingleResolved`(+`Framing`), `DeclarerNameCap`, `DeclarersNote`. The population was derived with the compiler (delete the entry points, build, delete what it names, repeat) and the sweep for the tier's names over `.cs`/`.md`/`.json` returns 0 live references. With the `precise` flag gone, `ChildClause` collapsed to its one surviving kind, so `ChildNotes`, `ClauseOf` and `ClauseReserve` lost the branches no caller could take, and `AppendConflictTree` lost the `prefetched` parameter the tier was its only source of. Three probe arms over `DeclarersNote` died with the sentences; the SENTENCE net that covers them derives its subjects from `ReadSentences`' own consts, so it shrank by construction rather than by hand.

**What was deliberately NOT done, and why.** Every remaining caller of the render-layer `conflictTree` parameter passes `false`, so on the "a branch no caller can take is deleted, not kept" rule the parameter is a deletion candidate. It stays: its other branch is `AppendConflictTree`, reached only from `Wire.RenderRecord` — one of the eight surviving members item 1 hands to #486 — so deleting the parameter would edit a member this fold was told not to touch and would cascade into deleting the conflict-tree diff render. It goes with item 1's ruling, not this one. (The service-level `LoadOrderService.ResolveRead(…, conflictTree:)` parameter has a live `true` caller in `SnapshotViewProbe` and is untouched.)

**Caller-visible consequence, in `plugin/CHANGELOG.md`** (`8ebb03c`): the cut's own `## Unreleased` entry now states that `read_record conflict_tree=true`'s declarers note is not carried to 2.0 — `records` states that other plugins touch the record and that this read did not open their bodies, and `project={"form": "tree"}` shows each touching plugin's own content — pointing at #485.

**ADR 0005 is unchanged by this ruling**: it does not mention L2. Its two sentences that were false about this tree belong to item 1, and are rewritten there (`1fca938`).

**The loss as filed (#485).** What the tier did: a CELL owns its placed references, a DIAL its INFOs, a WRLD its cells, and the game assembles children from every declaring plugin — so a winner can show an empty list while lower plugins carry hundreds (#342's Dawnstar cell: `Temporary` 0 at the winner, 201 in `Skyrim.esm`). Why it was unreachable: only the deleted `read_record` conflict-tree mode set the flag. What `records project={"form":"tree"}` gives today and where it degrades: it names every touching plugin and shows what each holds in the field, so the caller reads declarations off per-plugin content — but it renders *differences*, so a lower plugin whose content is identical to the winner's is not visible as a declarer, and it does not state the "nobody declares anything here" answer in its own voice. Candidate shapes: a `records` lever that fires the tier, or the tree form emitting the declarers sentence automatically. Promotion condition, verbatim: *"promote if the tree form proves insufficient in daily use — Aaron's data"*. Routing and milestone are Aaron's; filed unmilestoned.

<details>
<summary>The escalation as it stood before the ruling</summary>

The cheap tier ("this read did not open the other plugins' bodies") fires on `housecarl_records` in full. The precise tier — naming WHICH plugins declare content in a child-bearing field, and saying **nothing** when none do, which the cheap tier cannot know — fires only under `conflictTree == true` in the shared render, and every `RecordsTools` call site passes `false`. `ResolveTreeFill`, `ApplyPreciseChildNotes` and `ReadSentences.DeclarersNote` all survive **with no caller**. There is no `conflict_tree` lever on the records surface. 19 arms die here.

Mitigation measured, not assumed: `records project={"form":"tree"}` answers the same *question* in a different shape — it names every touching plugin and shows what each holds in the child-bearing field — and that is now pinned by 7 arms. It renders *differences* rather than *declarations*, so it degrades where a lower plugin's content is identical to the winner's; in the case the annotation exists for, the difference IS the declaration.

</details>

## The framing for items 2 and 3, and the alternative

**§4 framing (b).** SPEC §4.2/§6.1 assume the named-plugin SOURCE pole absorbs `read_plugin_file`, and that `records` absorbs the read family. It absorbs the **reads**. It does not absorb these two advisories. RUN_ORDER rule 10 says a 1.x tool the SPEC does not cover is never a silent survivor; the mirror holds — a capability the successor does not carry is never a silent deletion.

**Three revisions preserve the goal, and all three are small.** Each adds caller-visible output to a shipped 2.0 tool, which is why none was taken without a go:

1. Carry the two advisory lines onto the `records` off-order lane.
2. Split `ErrorFindingClass.MissingMasters` so `check` reports installed-but-inactive separately. — **TAKEN, in a smaller shape. Ruled compensate-on-`check` 2026-09-03; see item 2.** The finding class was not split: one class, one count, one list, and the REMEDY names which case each master is in — which gives the caller the distinction without a second class to filter on.
3. Give `records` a lever that fires the precise tier. — **not taken. Ruled accept-and-delete 2026-09-03; see item 3.**

**The alternative is accept-and-delete:** rule that the loss is accepted, in which case the now-unreachable code (`ResolveTreeFill`, `ApplyPreciseChildNotes`, `ReadSentences.DeclarersNote`, and `RenderPluginFile`'s advisory arms) is deleted with it rather than left on the tree as scaffolding for a caller that does not exist. That answer folds item 2's half of item 1 into the same stroke.

---

`Closes #468`

# The review rounds

**Two rounds, four reviewers, all Opus, each in its own worktree with its own `TMP`.** Round 1 was one seeded reviewer plus one gate-replica reviewer seeded with nothing but the branch and the base; round 2 was two seeded reviewers. Full results are in the lane's scratch folder. **No round 3 — the hard cap, and the class-stop below.**

## Round 1 — 2 high, 8 medium, 11 low. 12 folded across 6 commits, 3 refused.

- **HIGH, driven end to end:** six live sentences on `housecarl_records` paths named deleted tools by their BARE spelling — the scan and chain lanes' unscannable-record note, the chain form's three refusals, and the chain render's header. One was a remedy naming `cross_plugin_query references=`, a call that no longer exists. Repaired with seven executing arms. The retired vocabulary arm never covered bare names: its own doc says it matched `housecarl_`-prefixed tokens only, which is exactly where all six sat.
- The branch's own new dense+depth refusal fired in the `formids=` lane, where its remedy ("drop `project.depth`") lands the caller on a second refusal. Gated to the scan lane, with the lane decided once and the dispatch reading the same value; an arm per branch.
- The artifact-manifest provenance repair had **no arm anywhere** — reverting both stamps stayed 777/777 and 130/130 green.
- A fold gutted the `enum Tier` doc comment (its `<summary>`, `<list>` opener and whole `Construction` item), invisible to every gate because the project generates no documentation file. Found by both reviewers independently.
- `AliasTable`'s summary still said the read and check rows "stay dormant" and pointed a maintainer at a probe this branch deletes; `CONTRIBUTING.md`, `docs/architecture/tool-schema-publication.md` and `ToolSchemas`' comments named the deleted `binding-shim-guard`; `README.md` named two deleted tools; the residue baseline's `lines` drifted on ten files this branch trimmed (base drift: 0).

**Refused, with ground — 3 of 15:** the retired-name sweep's narrowed subject set (measured: no hole today, all 17 rows are 1.9.0 names, and it is a shape question about a test the roster PR shipped); `SeqTimestampRefreshed` at that point (read at source: it describes what the sweep does, it does not instruct a call — **round 2 overturned this**); and the gate-replica reviewer's declared-master finding, which is item 2 above and already open.

## Round 2 — 3 high, 7 medium, 7 low. CLASS-STOP.

Both reviewers independently derived and **drove** the same finding: **seven sentences this branch wrote spell the successor call as `housecarl_records source="…"` with no select term**, and the tool refuses a call carrying only a source pole. Both made the calls exactly as written and got `error: select something - formids= … types=, plugins=, conflicts_only=true, where=, references=.` One of the seven is a **published tool description**. A second finding: the copy read-back remedy tells the caller to confirm masters via `housecarl_records`, which **renders no plugin's master list in any form or transport** — driven four ways. A third: `SeqTimestampRefreshed`, the sentence round 1 refused, and round 1's own fold had pinned its incomplete spelling into `seq-write-guard`.

**That is the same failure class in two consecutive rounds** — *a caller-facing remedy names a call that does not work*, three instances in round 1 and three in round 2 — so the rounds stopped and the seam was escalated rather than a third instance folded.

### The seam, and how it was ruled

Amendment 2026-09-02 (2) already ruled the cure — a remedy that names a call ships with an arm that MAKES the call — but nothing derives the POPULATION the rule applies to, so every PR hand-sweeps a subset and is short by exactly what nobody thought of. This PR's own round-1 fold is the proof: it swept the chain and scan lanes by hand and missed the `source=` family, two sites of which sit in files it edited in the same commit.

That seam was already put to you on PR #474 and ruled there — the derivation is chartered work, not a per-PR sweep — with your caveat recorded: *"however I have a suspicion this is also probably not the smarter better overall solution, it sounds like a class that can continue to bite."* Under your 2026-09-02 relaxation — *"'never fold a third instance' … is supposed to be an escalation trigger not a 'don't fix anything more' trigger"* — the leftover instances are fixed once the seam is ruled.

So: **the ten enumerated sites are folded with executing arms on this branch, and the derivation is filed as #483** — subject set: every shipped prose site interpolating a `ToolNames` constant, the registry's class-(b) sites; these ten plus PR #474's four open class-A findings are its first cells. **The population folded here is HAND-DRAWN and is described that way in the code**: it is what two reviewers derived and drove, not a set computed from the surface.

### Round 2's other findings, all folded with arms

- **The scan render's headers were unguarded in BOTH harnesses.** Sabotaging `scan:` back to `cross_plugin_query:` left 791/791 and 130/130 green — the highest-traffic render on the surviving surface, whose repair this PR's CHANGELOG advertises. Now driven over every transport the tool declares (`Wire.QueryFormat`'s own members) plus the group_by twin.
- **The artifact fold was short by one site.** THREE writers had their stamp changed, not two: the batch writer was unguarded and its sabotage left the file green. The population is derived twice now — every `writer.Save(path, …)` site must pass the constant (8 sites, vacuity floor), and every form that spills is driven with the form vocabulary parsed out of the tool's OWN refusal.
- **The under-specified-call arm hand-enumerated four subjects** against roughly two dozen refusal sites. One row per declared parameter now, off the same reflected method the declared-spelling half uses, with a served row skipped-and-counted and a printed coverage floor.
- **`AliasActivationTests` was blind to a live alias row going dormant** — one of the two classes its own docstring credits the retired literal with catching. Because the population comes from the alias table, a mistyped candidate removes the behaviour and its subject together: three activations vanish, nothing goes red. The two counts are pinned as well as printed. **This also corrects the arms-owed list below**: the "row that fails to fire" half is ported only against a table that moves with it.
- **ADR 0005 said "five write tools absorb six" — four do** (`apply`, `create`, `remove`, `forward`).

### Two things this PR's verification caught in its own folds

Recorded because they are the honest part of it.

1. The executing arms proved the CALLS work but not that the SENTENCES still name them — measured by sabotage, three repaired sentences could regress with both gates green. Three pins were added against the shipped text: the description off the publishing attribute, the advisory RENDERED by the shipped method, the copy line against its source.
2. The first version of the advisory pin **returned early when the advisory did not name the tool**, which made it vacuous — it skipped the branch it was written for and stayed green under sabotage. It asserts the branch was reached now. The sibling unticked-in-plugins.txt advisory stays unpinned, stated in the file, because this world has no installed-but-unticked plugin and inventing one is a fixture change rather than a pin.

### Sabotage

**Three sweeps, 19 cells, each with a canary hand-proven RED before the sweep ran, a green control, restores grep-verified, and the tree asserted clean after every cell.**

The second sweep's first run reported four false greens — its ci oracle grepped for `ALL PASS`, which a probe prints in its own output, so a 128/130 run read as green. Re-run against the `ci-all summary - 130/130` line: two of the four were genuine gaps (folded above) and two were sabotages too narrow to remove every select term from the sentence under test. That is the fails-toward-green family the canary rule exists for, caught by the canary rule.

# The populations, derived

Every number below was re-derived by the review session from the committed tip, independently of the build session, before any reviewer saw the tree.

## The eleven

The **deletion-flagged** constants in `src/housecarl-core/ToolNames.cs`, confirmed against the SPEC's absorption map in both directions:

- Base carries **11** constants commented `deletion-flagged: dies at the 1.x cut`; the cut carries **0**.
- Base declares **46** tool-name constants, the cut **35** — **11 removed, 0 added**, and the removed set is byte-equal to the flagged set.
- **Constants and the wire agree exactly on both sides**: 46 constants == 46 published names on the base, 35 == 35 on the cut, empty symmetric difference in both directions.
- SPEC §6.1: `records` absorbs `read_record`, `batch_record_detail`, `resolve`, `cross_plugin_query`, `read_plugin_file`, `diff_record`, `effect_chain`, `skypatcher_read` — the eight. `check` absorbs `check_errors` + `validate_scripts`; `validate_dialogue` SPLITS, its finding classes to `check findings=["dialogue"]` and its merged INFO order to `records project={"form":"info_order"}` — the three.
- **No other 1.x tool is absorbed by a shipped 2.0 tool and left standing.** All 35 survivors walked: `copy_npc_appearance` is the only one absorbed by a shipped tool (`copy`), and it is the recorded #387 gate. Every other absorber the SPEC names — `place` (S2), the merged SKSE reader (S7) — has not shipped, so its ancestors survive by construction.

## The wire

Both binaries driven over stdio (`initialize` → `notifications/initialized` → `tools/list`):

- **46 → 35**, **exactly the eleven gone**, **0 added**.
- **Exactly two descriptions changed**, at the lengths below. Every other surviving description is byte-identical, and no fold in the review rounds touched a tool description.

## The probes

- **Five probe files deleted**: `ArtifactGuardProbe`, `BindingShimProbe`, `CheckDifferentialProbe`, `CheckMeasureProbe`, `DialogueWidthProbe` (the last retired unconverted — no assertion, no ci-all row).
- **Probe files 155 → 150**; `[CiProbe]` attribute sites 137 → 135; ci-all rows **132 → 130** (the two lost rows are `ArtifactGuardProbe` and `BindingShimProbe`).
- The build session derived the driving population twice, by script, because two oracles see different things: a *method-call* arm (tool identity from `[McpServerTool(Name = …)]` paired with its method, 46/46 mapped) found **11 probe files / 188 call sites**, reproducing the W7 sizing pass's own figures exactly; a *wire* arm (a probe naming a tool as a STRING, which the method arm cannot see) found `BindingShimProbe` with **20 driving call sites**, matching the sizing pass §2.3. The review session re-derived by a third oracle — any of the eleven named by literal or by `ToolNames` constant — and got 20 base probe files, agreeing on the deleted set.

## Thirteen new test files, 395 tests

382 → 777 on the build session's tip, 791 after the review folds. Every new test file was written and sabotage-verified in its author's own isolated worktree, from a committed state, with a canary proven RED by hand first. **No frozen fixture was modified** — `ServerFixture`, `HarnessPaths`, `AssemblyInfo`, `RecordsWorld`, `RecordsTestBase`, `LinePump` are byte-identical to the base.

# The arms owed

Every assertion in the eleven-file population was classified as **dies-with-tool** (no expression on the 2.0 surface), **engine-survives** (written fresh against the 2.0 tool), or **already covered** (an existing test named).

| source probe | in-scope arms | dies | fresh test written | already covered | fate of the file |
|---|---|---|---|---|---|
| `ArtifactGuardProbe` | 64 | 4 | 56 | 4 | **deleted whole** |
| `BindingShimProbe` | 41 | 0 | 40 | 1 | **deleted whole** (`Converted-from:` marker) |
| `OwnedChildContentProbe` | 48 | 19 | 28 | 0 | survives, 12 engine arms |
| `BulkPrimitivesWave2Probe` | 43 | 3 | 35 | 5 | survives, 17 engine arms |
| `BulkQueryPrimitivesProbe` | 34 | 5 | 27 | 2 | survives, 45 engine arms |
| `CheckErrorsProbe` | 8 | 0 | 8 | 0 | survives, ~84 engine cells |
| `ReadPluginFileProbe` | 6 | 1 | 4 | 2 | survives, ~27 engine arms |
| `ExcludedMasterWriteProbe` | 2 | 0 | 2 | 0 | survives |
| `BulkPrimitivesWave3Probe` | 2 | 0 | 2 | 0 | survives |
| `DialogueWidthProbe` | 0 | — | — | — | **retired unconverted** |
| `CheckDifferentialProbe`, `CheckMeasureProbe` | 0 | — | — | — | **deleted whole** (no assertion, not in ci-all) |

**Owed, and why:**

1. **The alias-activation census's exact-set half.** `BindingShimProbe`'s census was a 144-row eyeballed literal guarding two failures. The first — a row that silently fails to fire — is derivable and IS ported, computed by running the shipped `ResolveAliases` over the surface's own schemas and then driven over the wire, which is strictly more than the literal did. The second — "no row fires anywhere unintended" — is not derivable: "unintended" was the human judgement the list encoded. **Filed as #482** by the review session, with the honest close (a generated-and-reviewed snapshot: derived set written to a data file by an update run, checked in, diffed in review). Until then an activation landing on a tool nobody checked is unguarded.
2. **The 19 precise-tier owned-child arms** — item 3 above. They die because the behaviour is unreachable from the 2.0 surface, not because they could not be fixtured.
3. **The declared-master advisory** — item 2 above.

# The two survivor descriptions that changed

Both were sentences on a surviving tool naming a tool that will not be there. These are the only two; every other published description is byte-identical to the base.

| tool | delta | change |
|---|---|---|
| `housecarl_check` | **−4 chars** | "…to inspect one record use `housecarl_read_record`." → "…use `housecarl_records`." |
| `housecarl_skypatcher_layer` | **+154 chars** | "For ONE record's computed post-SkyPatcher state use `housecarl_skypatcher_read`." became "…use `housecarl_records formids=["<FormID>"] source={"overlay": "skypatcher", "state": "post"}` — source= is a version pole, not a selection, so the read needs formids= (or a scan scope)." The delta grew from +42 to +154 in the round-2 fold: the first repair named the successor without a selection, and the call it described was refused. |

# Issue dispositions — measured, not assumed

- **#468** — the tracking issue. `Closes #468`.
- **#482** — **filed by this PR's review session**, the issue the arms-owed list owes: the alias-activation census's exact-set half is not statically derivable and is unguarded after the cut. Filed unmilestoned; routing is Aaron's.
- **#444** (`LeverNames.Legacy` should have no default) — **DOES NOT close by construction. Stays open.** Amendment 2026-09-02 (7) recorded the expectation that the cut would leave the compiler naming nothing outside deleted code. **Measured, and the expectation is falsified more strongly than the build session put it.** The record says 14 sites in five surviving files; re-measured over all of `src/` excluding the test project, for every spelling of the token, it is **11 sites in three files** — `JsonWire.cs` 5, `ReadTools.cs` 5, `Artifacts.cs` 1; `LoadOrderService.cs` and `WriteSentences.cs` carry zero. Traced across every commit on the branch and onto the base: **11 sites / 3 files at `claude/roster`, and 11/3 at every one of the branch's commits.** So the cut removed no `LeverNames.Legacy` seam at all. They are unreachable at runtime — every shipped caller passes explicit levers — but still the default, which is what #444 is about. Closing it means making `levers` required at those 11 seams: bounded, mechanical, and a different change.
- **#471** (`CreateOpNaming.Legacy` unreachable but default) — **stays open.** The cut did not delete its last reader. Two sites, both in `LoadOrderService`: the singular `CreateRecords` path that hardcodes `Legacy` has no shipped caller but is called by eight guard probes; `CreateRecordsBatch`'s `naming ?? Legacy` is unreachable from `CreateTools.Create`, which passes an explicit naming.
- **#473** (the Codex coverage guard is one-directional) — **not closed, and measured green over the gap.** `plugin/codex/housecarl/SKILL.md` names **11 distinct deleted tools across 14 sites** while `ci-all` is 130/130. The guard asserts registered ⊆ router and never the reverse, exactly as #473 says. The router is untouched under Aaron's PR #474 ruling.
- **#387** — untouched. `copy_npc_appearance` survives by that gate, confirmed against the SPEC.

# Skills and the Codex router — not edited, measured

Aaron's ruling on PR #474 holds: *"I don't want to do anything about the skills until we get to the re-write … this will only affect people cloning the repo temporarily which I am fine with."* Owed to W6 (RUN_ORDER row 9).

**96 references across 17 files**, re-derived by the review session (release history in `plugin/CHANGELOG.md` excluded, since it is history and stays):

`bulk-record-jobs` 30 · `dialogue-authoring` 14 (+13 across its `references/`) · `facegen-diagnostics` 11 (+1) · `tool-output-awareness` 4 · `skypatcher-authoring` 3 · `biped-slot-reference` 2 (+1) · `oar-authoring` 2 · `mutagen-reference` 1 · **`plugin/codex/housecarl/SKILL.md` 14**.

**No skill's frontmatter description names a deleted tool** — checked; every break is in a body or a reference file, so no skill's primary workflow is the thing that died.

# CHANGELOG and ADR 0005, as landed

`plugin/CHANGELOG.md` gains four `## Unreleased` bullets: the eleven deletions with what carries each one and how to check it from a client's tool list; the surviving-tool sentences that were repaired; the new `format='dense'` + `project.depth` refusal and which lane it belongs to; and the two renamed headers with the chain form's repaired refusals.

`docs/decisions/0005-the-1x-tools-are-deleted-not-deprecated.md` records why the 1.x tools are deleted rather than deprecated — the caller is a model reading tool descriptions, every surviving 1.x tool has to be maintained as if live, and there are no installed users to protect on this schedule (the 1.9.0 zip is the last 1.x release; the next bump is 2.0.0). **Its two sentences that were false about this tree are rewritten at `1fca938`** — see item 1. Its other counts were re-checked against the tree at the same commit and none needed changing.

# The gates

Every figure re-measured by the review session from the committed tip, at boot and again after the folds.

| gate | build session's tip `908bb22` | after the review folds ``2a7f64a`` |
|---|---|---|
| `dotnet build housecarl.sln -c Release` | 0 errors | **0 errors** |
| `dotnet build … --no-incremental` | 83 distinct warnings | **83 distinct warnings** |
| `ci-all` | 130/130, exit 0 | **130/130, exit 0** |
| `dotnet test src/housecarl-mcp-tests` | 777/777 | ****824/824** (+47 arms)** |
| `tools/list` off the built server | 35 (was 46) | **35, unchanged** |
| residue baseline | probeFiles 150, ciAllRows 130 | probeFiles 150, ciAllRows 130, **lines refreshed** |

**Re-measured at the current tip `1fca938`** (after the L1 fold `47019de` and the ADR fold): build **0 errors**, `dotnet test src/housecarl-mcp-tests` **833/833** (+9, all L1's), `ci-all` **130/130 ALL PASS**. The ADR fold changed no code, so it ships no sabotage cell — there is no conditional and no sentence a caller reads. Its verification is the three gates plus the counts it asserts, each re-derived from the tree (the frozen 1.9.0 capture for 45, `ToolNames` and the residue baseline for the rest).

**On warnings.** The decision record does not state a figure, and the first build reported zero because it was incremental and nothing recompiled — a fails-toward-green reading. A full `--no-incremental` build emits 89 warning lines / **83 distinct file+line+code sites**. The BASE was built `--no-incremental` as the control and the distinct sets diffed: **83 on the base, 83 on the cut, zero new, zero gone.** The branch introduces no warning and removes none.

# Settled decisions, with tags

Carried verbatim into every seeded reviewer prompt. `ruled` entries reviewers were told to honour; `inferred` entries they were told they MAY re-litigate — and one of them did, which is item 1 above.

- `ruled` — every 1.x tool whose capability the completed 2.0 surface covers is deleted, no deprecation window, rolling; guards on deleted tools die with them; surviving engine behaviour gets fresh tests in the test project in the SAME PR. (Aaron, 2026-08-29 rule 10; 2026-09-02 (5): *"I agree on all of this and ratify it is as the new plan"*.) — **the "same PR" half was amended 2026-09-03** once the cost was measured: Aaron ruled item 1's render halves and their 147 arms a follow-up (*"agree, its a follow up"*), #486. The rule stands; this PR is its measured counter-example, and ADR 0005 now says so.
- `ruled` — the cut is ONE PR: 8 read + 3 check, `20b44b2` re-applied, compile errors are the checklist. (Amendment 2026-09-02 (5).)
- `ruled` — no fix work in deletion-flagged tool bodies; skills and the Codex router untouched until the W6 rewrite. (Aaron, PR #474 body.)
- `ruled` — every shipped tool-name literal interpolates a `ToolNames` constant; the registry's scripts are a one-shot record, not re-run. (Amendment (base) (i); PR #477.)
- `ruled` — new guards go in the test project; nothing leaves the old harness except by a conversion PR; shared fixtures frozen. (2026-09-02 (5).)
- `ruled` — #440 runs AFTER this PR, before the conversion PRs. (2026-09-02 (7).)
- `ruled` — sessions and subagents are Opus or lower. (Aaron, 2026-09-02.) Honoured: every agent in both engagements was spawned `model: "opus"`.
- `ruled` — an ADR says what is true of the tree it ships on; a stale ADR sentence is amended by a dated superseding note in the PR that makes it stale. (Aaron's gate on PR #477.)
- `inferred` — stacking on `claude/roster` rather than waiting for its merge.
- `inferred` — retiring, not re-flooring, the vocabulary guard's TOOLNAME arm. Both round-1 reviewers checked the reasoning and upheld it: measured, shipped prose carries no `housecarl_` token outside a constant, so the arm's claim is true by construction now. Neither asked for it back. But the arm never covered BARE retired names, and that is where all six of the sentences this PR repaired sat.
- `inferred` — ADR 0005 belongs on this PR, because this is where the demolition becomes true of the public tree.
- `inferred` — the per-assertion classification scheme, and the arm-level reading of "the probes die with the tools". **Re-litigated and landed — see item 1.**
- `inferred` — #471's disposition.
- `inferred` — repairing sentences on surviving tools that named a deleted tool, and repairing the defects the cut exposed, rather than listing them open. Precedent: PR #474 folded its class-D refusal regression rather than merging Q3's silently-degraded mode.

# Everything left open

**No item on this PR is open for a ruling.** What follows is state, and issues that outlive the PR.

1. **Item 1 is closed as a follow-up.** Executed against the compiler and measured: eight dead members (not nine), 197 named sites, **class (b) at 84% against a bound of about a third** — 156 sites owing a fresh test across 147 arms, with a Papyrus fixture port, a file-lock harness and two re-measurements that are not mechanical. Aaron: *"agree, its a follow up"* (2026-09-03). Filed as **#486** — two sequenced PRs, fixtures first, landing after this one and before #440; unmilestoned, routing Aaron's. ADR 0005's two sentences are rewritten at `1fca938` to say what is true of this tree.
2. **Items 2 (L1) and 3 (L2) are closed** — folded at `47019de` and `918a127`.
3. **Four sibling dead-name sentence sites that live INSIDE the code item 1 is about** (`ReadTools.cs:578/1105/1384`, `DialogueTools.cs:39/51/65`, `LoadOrderService.cs:2717`, `SweepScope.cs:138/165`) plus `DialogueValidate.cs:113`'s comment — deliberately not repaired, because prose in code that may be deleted is work either wasted or undone. They are unreachable today: `housecarl_check` refuses an unknown finding class with its own correct sentence first, measured.
4. **#444, #471, #473** — open, with the measurements above.
5. **#482** — filed, unmilestoned, routing Aaron's.
6. **#483** — the executing-arm population derivation, filed by this PR's review session after the class-stop. The ten sites folded here are its first cells; PR #474's four open class-A findings are the rest. Filed unmilestoned.
7. **The unticked-in-plugins.txt advisory is unpinned** — the same sentence shape as the on-disk advisory beside it, but this world has no installed-but-unticked plugin and inventing one is a fixture change rather than a pin. Stated in the test file.
8. **The retired-name redirect's residual case** — `RetiredNameRedirectTests` derives its subjects from the frozen 1.9.0 capture, so a row added for a name 1.9.0 never published would have no cell. Measured: no such row exists today (all 17 are 1.9.0 names). Recorded in the file's own comment rather than guarded, because there is nothing to guard.

---

## Rebased onto main 2026-09-03

The base was `claude/roster` (PR #481). That merged, so this branch was rebased from `9adf61d` onto `main` @ `18709bb`. The 21 own commits replayed with no conflict — no guard file and no `harness-residue-baseline.json` hunk.

One conflict was semantic rather than textual. `ToolSurfaceCensusTests.EverySpellingThisBranchRepointedIsStillCaught_TheCollisionFixCostsNoTeeth`, added on the roster, pins eight sites this tree repointed and reads each one by reflecting over the built tool-surface assembly. Two of the eight name `ReadTools`, a type this branch deletes at `34f6434`, so the reading no longer reaches them and the test reported 6 of 8. Fixed at `d841259`: the two rows are removed from the fixture, not filtered around — a row for a deleted type is an arm that cannot fail. The assertion's count already derived from the fixture's length, so no number changed by hand.

Gates at `d841259`: build 0 errors; `dotnet test` **832 → 840** under `tier!=bridge` (833 → 841 unfiltered), the delta being the roster's eight gate-fold tests, with nothing on this branch lost — verified by diffing the two test-name listings; `ci-all` **130/130 ALL PASS**; `tools/list` **35**, canonical size **177 099 chars**, unchanged by the rebase.


Avick3110:
RULED 2026-09-03 — item 1 (the eight surviving render/service members and the 147 arms they owe) is a follow-up, not this PR. Aaron: "agree, its a follow up".
Follow-up filed as #486 — two sequenced PRs, fixtures first (Papyrus fixture, file-lock harness), then the deletion; after #484, before #440. Unmilestoned, routing Aaron's.
ADR 0005 corrected at 1fca938 — the two sentences that were false of this tree now say what holds here; gates from that tip: build 0 errors, 833/833 tests, ci-all 130/130.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8517,16 +8504,12 @@ public PluginFileOutcome ReadPluginFile(string plugin, string? formid, string? t
...
-            var inactive = new List<string>();
-            foreach (var m in masters)
-            {
-                if (!Mo2LoadOrder.PluginFileExists(comp, modsDir, dataDir, overwriteDir, m)) { missing.Add(m); continue; }
-                bool active = comp.ActivePluginNames.Contains(m)
-                              || comp.ImplicitPluginNames.Any(x => x.Equals(m, StringComparison.OrdinalIgnoreCase));
-                if (!active) inactive.Add(m);
-            }
+            // active-order notion housecarl_check's errors family uses — NOT mere on-disk presence, so the two agree.
+            var unsatisfied = masters.Where(m => !(comp.ActivePluginNames.Contains(m)
```
Avick3110, line 8528: **Confirmed regression: a ticked-but-absent master is now reported in neither list.**

The old loop tested `PluginFileExists` FIRST and put any master with no on-disk copy into `missing`, whatever plugins.txt said about it. The new `unsatisfied` filter runs the ACTIVE test first, so a master that is ticked in `plugins.txt` but whose file is gone from the install is filtered out before `SplitUnsatisfiedMasters` ever looks for it, and comes back in neither `MissingMasters` nor `InactiveMasters`.

Concrete case: `plugins.txt` carries `*HcTaAbsent.esm` (a stale tick after the mod folder was deleted outside MO2) and `HcTaPatch.esp` declares it as a master. Before: `! declared master(s) NOT installed anywhere in the MO2 install: HcTaAbsent.esm`. After: silence.

Verified on this branch — a test asserting the name appears in `MissingMasters ∪ InactiveMasters` FAILS at `d841259` and PASSES with the pre-refactor loop restored in place, everything else unchanged.

The arm that rendered both advisory lines was deleted in this same PR (`ReadPluginFileProbe`, "it has NO replacement"), so nothing catches this today — and it is the shared home #486's render-halves PR will re-expose.

The splitter's own doc says the caller "decides what unsatisfied means against its own notion of the active order" — so this caller's notion needs both halves:

```suggestion
            var unsatisfied = masters.Where(m => !(comp.ActivePluginNames.Contains(m)
                                                   || comp.ImplicitPluginNames.Any(x => x.Equals(m, StringComparison.OrdinalIgnoreCase)))
                                                 || !Mo2LoadOrder.PluginFileExists(comp, modsDir, dataDir, overwriteDir, m));
```

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4636,10 +4592,41 @@ public ErrorCheckResult CheckErrors(IReadOnlyList<string>? plugins, int limit,
...
+
+    /// <summary>Fill in each report's install-vs-enable split for the masters the sweep found unsatisfied. The core
+    /// sweep knows the ACTIVE ORDER and stops there; which of those masters is nonetheless sitting in the install —
+    /// in a disabled mod, or unticked — is a fact about the MO2 composition, which lives at this layer. Done here so
+    /// the split has one home shared with <c>read_plugin_file</c>'s advisory
+    /// (<see cref="Mo2LoadOrder.SplitUnsatisfiedMasters"/>) rather than a second spelling inside the core.
+    /// <para>A composition that cannot be read leaves every report's subset NULL — not empty. Empty would say "none
+    /// of these is merely disabled", which is a claim about an install nobody looked at; null says the split was not
+    /// made, and the render answers with the union remedy it printed before (Q3).</para></summary>
+    ErrorCheckResult ClassifyMissingMasters(ErrorCheckResult r)
```
Avick3110, line 4611: **Cost: this re-walks the whole install once per (plugin, master) pair.**

`SplitUnsatisfiedMasters` is called once per report with no memoization, and `PluginFileExists` short-circuits only on a HIT. So the expensive path is exactly the common one — a master that is NOT installed pays the enabled-mods loop, then the disabled-mods loop, then `UnlistedModFolders`' full `Directory.EnumerateDirectories(modsDir)` plus a stat per unlisted folder, and then returns false.

A whole-order `housecarl_check` where 30 plugins each declare the same 2 uninstalled masters runs 60 independent whole-install scans; on a 2000-mod instance that is on the order of 10^5 filesystem calls added to one sweep, for an answer that is the same 2 names every time.

`Mo2LoadOrder.AllPluginFileNames(comp, modsDir, dataDir, overwriteDir)` already builds the whole name set in one pass and is right there — or memoize per name in a `Dictionary<string,bool>(OrdinalIgnoreCase)` across the report loop and hand that to the splitter. Either keeps the one-home property while making the cost independent of how many plugins declare the same shortfall.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -1277,7 +583,21 @@ internal static string ComposeErrorSection(PluginErrors p)
         var fixedPart = new StringBuilder("\n[ERROR] ").Append(p.Plugin).Append('\n');
         if (p.ScanError is not null)
             fixedPart.Append("  scan error: ").Append(p.ScanError).Append('\n');
-        if (p.MissingMasters.Count > 0)
+        // The two shortfalls are named apart because their REMEDIES differ, and a caller told "install/enable it"
+        // has to go find out which (PR #148's distinction, carried onto the surface that still reports the finding).
+        // Null means the split was not made — say what was said before rather than assert the uninstalled case.
+        if (p.MissingMasters.Count > 0 && p.InstalledButInactiveMasters is { } inactive)
```
Avick3110, line 589: **The split is text-only — `format='json'` never carries it.**

`ClassifyMissingMasters` populates `PluginErrors.InstalledButInactiveMasters`, and this render uses it, but `JsonWire.WritePluginHead` (JsonWire.cs:1461) still writes only the undifferentiated array:

```csharp
WriteStringArray(w, "missing_masters", p.MissingMasters);
```

So `housecarl_check plugins=["HcMmPatch.esp"] findings=["missing_masters"] format="json"` returns `"missing_masters": ["HcMmAbsent.esm", "HcMmGhost.esm"]` with nothing to tell them apart, while the identical call in text now says "NOT installed anywhere … [install them]" for one and "installed but NOT ACTIVE … [enable them]" for the other. The json caller cannot pick the remedy that will work, and every arm in `CheckMasterRemedyTests` drives the text render, so the divergence is unguarded.

Adding a `installed_but_inactive_masters` member (null when the split was not made — the same null-is-not-empty rule the DTO doc states) would keep the two transports saying the same thing. It costs a `PluginHeadCost` re-measure, which is why it is worth doing here rather than as a follow-up.

Avick3110:
Gate findings folded — one commit each, each reproduced as a RED test from the committed tip before the fix, then RED-checked again by sabotage from a committed state (sabotage → RED → `git checkout --` → grep-verified). All three were confirmed at source before this session started; nothing was refused.

- **3923605905** (`LoadOrderService.cs:8508`, a ticked-but-absent master in neither list) — `8dea77d`. Your suggestion applied verbatim: the `unsatisfied` filter is now "not active OR no on-disk copy". Arms in a new world whose absent master carries a stale tick (the L1 world cannot hold it — one `plugins.txt` cannot have a name both ticked and unticked): `CheckMasterTickedButAbsentTests.ReadPluginFileNamesATickedButAbsentMasterAsMissing_NotInNeitherList` is your reproduction, RED at `d841259` with `Assert.Contains() Failure … Collection: []` over `MissingMasters ∪ InactiveMasters`; plus `ReadPluginFileLeavesTheForceLoadedMasterOutOfBothShortfalls`. Sabotage (drop the `|| !PluginFileExists(...)` clause) → the first arm RED again. `ReadPluginFile` was fixed in place, not deleted or re-homed.

  **Your third arm was already green at the tip, and is now armed rather than trusted.** `TheCheckRemedyNamesTheTickedButAbsentMasterOnTheInstallLine` — the same fixture through `housecarl_check plugins=[patch] findings=["missing_masters"]` — passes at `d841259`. The reason is a second derivation: `check`'s active order is the BUILT one, and `Mo2LoadOrder.Build` drops a listed name nothing on disk provides (`Mo2LoadOrder.cs:104-109`), so the name arrives at `ErrorCheck`'s `!view.ContainsPlugin(m)` already unsatisfied and the splitter stats it to `NotInstalled`. Two derivations of the same question, one install, both armed now.

- **3923608780** (`LoadOrderService.cs:4611`, one whole-install walk per (plugin, master) pair) — `bf14612`. `SplitUnsatisfiedMasters` gains an overload taking the install's plugin-name set; the existing signature is now a thin wrapper that builds it via `AllPluginFileNames` once. `ClassifyMissingMasters` builds it once per call (inside the existing `try`, so no new catch and no new branch) and passes it for every report; `ReadPluginFile` builds it once and uses it for both its unsatisfied filter and its split. No caching across calls, no held state, and both split paths still draw from `CandidateFolders` — said in the doc comment, not a second walk.

  **The two cost numbers** (a report, not an arm — a wall-clock assertion does not belong in CI). Throwaway harness, kept out of the tree: 400 enabled mod folders, 200 disabled, 100 unlisted, 30 plugins each declaring the SAME 2 uninstalled masters, whole-scope `CheckErrors(findings=["missing_masters"])`, warmed once, median of five: **833.6 ms at `d841259` → 74.8 ms after** (runs 816.7/822.4/833.6/841.0/924.0 and 70.0/71.8/74.8/77.1/77.3).

  Equivalence arms over an install holding one plugin file in EVERY candidate layer — overwrite, an enabled mod, a disabled mod, an unlisted folder, game Data — plus one name in none of them, asserting fixture-known lists per entry point rather than one overload against the other: `MasterSplitInstallLocationsTests.TheWalkingSplitFilesEachMasterByWhichInstallLayerHoldsItsFile`, `TheSetBackedSplitFilesTheSameMastersFromANameSetTheCallerAlreadyHas`, `TheSetBackedSplitMatchesCaseInsensitivelyEvenFromAPlainList`, `ANameIsReducedToItsFilenameBeforeItIsLookedUp_ByBothEntryPoints`. Three sabotages, each RED: comparer → `Ordinal` reds the case arm; dropping the UNLISTED layer from `CandidateFolders` reds both location arms; dropping the `Path.GetFileName(m?.Trim())` normalization reds the fourth.

  **Residue this fold creates, flagged rather than left quiet:** `Mo2LoadOrder.PluginFileExists` and the 5-arg `SplitUnsatisfiedMasters` wrapper now have no shipped caller (both shipped call sites hold the name set already). Both are driven by the arms above and neither was deleted — that is not one of these three findings — but they belong on #486's roster of surviving zero-caller members.

- **3923608921** (`ReadTools.cs:589` / `JsonWire.cs:1461`, the split is text-only) — `7eebfc3`. `WritePluginHead` gains `installed_but_inactive_masters`, written through a new `WriteNullableStringArray` — the array twin of the `WriteNullable` the head already uses for `scan_error`. Array where the split was made, `null` where it was not.

  **The `PluginHeadCost` claim checks out.** It writes `WritePluginHead` itself into its scratch document through `MeasureUnit` (`JsonWire.cs:1482`), and `PluginHeadCostFor` (`:2951`) delegates to it — the cost follows the write by construction, and there is no hand-kept number for this head anywhere. ci-all's cap and floor arms are green unchanged.

  Arms (all three RED at `d841259` — the member does not exist): `CheckMasterRemedyJsonTests.TheJsonPluginObjectCarriesTheInstalledButInactiveSubsetAlongsideTheUnionList`, `TheJsonSubsetAndTheTextRemedyPutTheSameMasterOnTheEnableSide`, `AnUnclassifiedReportLeavesTheJsonSubsetNull_NotAnEmptyArray`. Sabotages, each RED: removing the member reds all three; making the helper write `[]` for null reds the null arm; writing `p.MissingMasters` into the subset reds all three. The null arm is driven at the render, and says so in its own summary: reaching `ClassifyMissingMasters`'s `catch { return r; }` end to end needs a composition read that THROWS, which a fixture can only produce by making the instance unreadable mid-call — so it is driven from an unclassified report, the same input that catch hands the render, exactly as the text lane's twin arm is. `plugin/CHANGELOG.md`'s L1 sentence now says "in both transports" and names the json member and its null.

**Gate from `7eebfc3`:** build 0 errors · `dotnet test --filter "tier!=bridge"` **840 → 850** (the ten arms named above) · `ci-all` **130/130** · `tools/list` off the built server **35 tools**, and no tool description moved — no `[Description]` string is touched in the three commits, and `check`'s own description does not enumerate json keys (checked against the dumped wire).

---

**Fold 2's residue is deleted, not parked — `83d7f4c`.** `Mo2LoadOrder.PluginFileExists` and the 5-arg `SplitUnsatisfiedMasters` wrapper are gone; the compiler named no caller in `housecarl-core`, `housecarl-mcp`, `housecarl-generator` or the test project. Their arms moved onto the surviving set-taking overload — the two location arms were the same fixture-known expectations put to two entry points and merge into one, so **849** rather than 850, with no assertion dropped (every layer, the case-insensitive comparison and the filename normalization are still named). The "same `CandidateFolders`" claim now has one home: `AllPluginFileNames` states it and names both its readers, and the split's doc points at it. Sabotage from `83d7f4c` — drop the UNLISTED layer from `CandidateFolders` → `MasterSplitInstallLocationsTests.TheSplitFilesEachMasterByWhichInstallLayerHoldsItsFile [FAIL]`; restored, grep-verified. Gate: build 0 errors, `dotnet test` 849, ci-all 130/130, CI green.

Issue #486, opened 2026-09-03
[NOTE] Read-family render halves: delete the eight zero-caller members and write the 147 engine arms they hid — fixtures PR first (Papyrus fixture, file-lock harness), then the deletion PR; sequenced after #484, before #440


Avick3110:
PR #484 deletes the eleven remaining 1.x read and check tools from the wire (46 → 35) and deletes their tool-layer probes. It does **not** delete the render and service members those tools called. Eight of them survive with **no shipped caller**, and the old guard harness still drives them from 197 assertions. This issue is the follow-up that finishes the job.

**Ruled by Aaron, 2026-09-03:** *"agree, its a follow up"* — on the advisor's recommendation of two sequenced PRs, landing right after #484 and before #440.

---

## The eight zero-caller members

Re-derived by script over `src/housecarl-mcp` + `src/housecarl-core` (bin/obj excluded), printing every word-boundary hit with `file:line` so a declaration, a comment and a call are told apart by reading; run on `claude/cut` @ `8ebb03c` and on the base `claude/roster` @ `9adf61d` as a control. Each has zero shipped callers on the branch and a real shipped caller (the deleted tool's body) on the base. Line numbers are as of `claude/cut` @ `47019de`.

| # | member | declared |
|---|---|---|
| 1 | `Wire.RenderCheckErrors` | `src/housecarl-mcp/ReadTools.cs:501` |
| 2 | `JsonWire.RenderCheckErrors` | `src/housecarl-mcp/JsonWire.cs:1299` |
| 3 | `Wire.RenderScriptCheck` / `JsonWire.RenderScriptCheck` | `ReadTools.cs:1041`, `JsonWire.cs:1857` |
| 4 | `Wire.RenderDiffRecord` / `JsonWire.RenderDiffRecord` | `ReadTools.cs:102`, `JsonWire.cs:137` |
| 5 | `Wire.RenderPluginFile` / `JsonWire.RenderPluginFile` | `ReadTools.cs:1312`, `JsonWire.cs:2278` |
| 6 | `Wire.RenderRecord` / `JsonWire.RenderRecord` | `ReadTools.cs:195`, `JsonWire.cs:379` |
| 7 | `LoadOrderService.DiffRecord` | `LoadOrderService.cs:2668` |
| 8 | `LoadOrderService.SkyPatcherPostState` | `LoadOrderService.cs:993` |
[...]

PR #488, opened 2026-09-03
#486 PR 1 of 2: the test project gets a Papyrus fixture and a file-lock harness

Avick3110:
Lands after #484 on Aaron's sequence; merge that first.

Part of #486 (PR 1 of 2; the issue closes at PR 2).

Two pieces of test machinery `src/housecarl-mcp-tests` did not have — a Papyrus fixture (a `.pex` writer plus the child/base property pair, re-homed as an MO2-instance world) and a file-lock harness — each in its own new file, plus the seven tests that prove the machinery is what it claims to be. Nothing is deleted. No file under `src/housecarl-generator`, `src/housecarl-mcp` or `src/housecarl-core` is touched, and `harness-residue-baseline.json` is not in the diff.

`docs/architecture/test-project-fixtures.md` is new and carries the rationale (ADR 0001's named home).

## Gates, measured from the committed tip

| gate | `main` @ `18709bb` | branch @ `a012b1e` |
|---|---|---|
| `dotnet build housecarl.sln -c Release` | 0 errors | 0 errors |
| `ci-all` | 132/132 ALL PASS | 132/132 ALL PASS |
| `dotnet test src/housecarl-mcp-tests` | 391 passed | 398 passed, 0 failed |
| published tool names | 46 | 46 (this branch publishes and unpublishes nothing) |

Measured again at each of the three tips, not once. Temp residue after every full run: zero `hc-lock-tests-*`, zero `hc-scripts-tests-*`.

## The rounds

Two rounds, four independent reviewers, each in its own isolated worktree, none of them the branch's author. Round 1 was one seeded reviewer plus one gate-replica (seeded with nothing but the branch and base, standing in for a first look with no conventions pointer).

**Round 1 — 0 high, 3 medium, 3 low, 2 nit. All 8 folded (`b7ca93a`), nothing refused** — every finding reproduced in the source when triaged against the code the reviewer cited.

- Two mediums, found independently from different angles: the loose-layer arm sampled findings with `Assert.Single(collection, predicate)` and never read the sizes, leaving `MyDefaulted` (the baked-initializer branch) and `MyAliasBound` (the alias binding) — two fixture properties whose whole point is producing **no** finding — guarded by nothing; and the acquisition-failure arm asserted the path with a bare substring that the inner exception's own message already satisfied. Both sets are now pinned whole; the path assertion is anchored to the sentence `HeldOpen` composes.
- The third medium: `LockWorld.Dispose` swallowed its `Directory.Delete`, so the overlay-leak fix in the previous commit could be reverted with the suite still green.
- Lows and nits: the sharing-violation assertion rested on .NET's localizable resource string and now rests on `ERROR_SHARING_VIOLATION` (HResult `0x80070020`, measured); the wire test carried two `tier` values because xUnit traits are additive and the class carried one; `HeldOpen.Path` was read by nothing and shadowed `System.IO.Path` inside the class; the ADR 0003 overlap and a cross-reference were stated where they belong.

**Round 2 — 0 high, 1 medium, 5 low, 4 nit — and two instances of round 1's class.**

Round 1's class was *an assertion whose real carrier is wider than the subject the arm names*. Round 2 found it twice more: the wire arm asserted the bare phrase `declared but NOT bound`, which `ReadTools` emits from **both** the object and the scalar branch (the fixture carries an unbound scalar, so the scalar branch alone satisfied it); and the lock arm asserted the plugin's file name inside an `IOException` message the BCL composes — the same defect round 1 fixed one method down, which that fold did not sweep for.

That is the same class in two consecutive rounds, so the rounds stopped there under CLAUDE.md §5 #11 and the seam went up for a ruling rather than a third fold. **No round 3.**

### Directed folds (`c3ce6f3`, `a012b1e`)

- The wire arm now asserts the **whole object-branch line**, spelled in the test from fixture-known values, which the scalar branch cannot satisfy. The structured form of the same fact is already asserted by the service arm; the wire arm keeps one text line for reachability.
- The lock arm's BCL-message substring is gone. The exact exception type plus the HResult are the discriminator, and neither is .NET's prose.
- The loud delete became `AssertNoHandlesLeft()`, called as each arm's last line after the hold is released, with `Dispose` best-effort again — a throwing `Dispose` was replacing a real assertion failure with its own.
- Three nits folded: `PluginPath` carries the constraint that an arm holding it open builds its own world; the sentence describing a mutating consumer that does not exist here is gone; the wire arm's count assertion has the clause saying why it is kept (a cheap canary that a scripts response came back at all).

### Refused

**One, and it stands after the ruling:** renaming the wire test so its name lists the unbound-property assertion too. A test name that undersells what it asserts is not a false claim, the name is already ninety characters, and the doc that *did* overstate was corrected instead.

## The derived population

The class's population was derived rather than hand-listed: every assertion on the branch whose expected text is a fragment of something composed elsewhere. **Nine**, all in the two test files.

| disposition | count | which |
|---|---|---|
| fixed this round | 2 | the bare `declared but NOT bound` phrase → the whole object-branch line (**corrected at your gate — see below**); the BCL file-name substring → deleted, type + HResult carry it |
| already anchored to a **fixture-known value** | 4 | the instance path and profile name in the `set_mo2_instance` confirmation; the record count; the composed `'Scripts\HcSpNoPex.pex' is not on disk` reason line, whose script name is the fixture's |
| already anchored to a **whole composed clause** | 2 | the `MySpell (Spell) on script HcSpChild` fragment, now merged into the whole line; `could not take an exclusive hold on '<path>'`, which `HeldOpen` composes (fixed in round 1) |
| not of the class | 1 | `Assert.DoesNotContain(ServerFixture.ConfigPrompt, …)` — a negative assertion against a shared constant, not a hand-spelled fragment |

The remaining assertions on the branch compare against fixture-known constants or structured result members, so none takes its expected value from the expression it checks.

**Correction (gate fold `aff8291`, your comment 3924563619).** The object-branch row above claimed a disposition the assertion had not earned. The whole line ruled out the *scalar* branch as a rival carrier; it did not rule out the second **record**. `HcSpAliasQuest` attaches the same `HcSpChild` through `QuestFragmentAlias.Scripts` and binds nothing, and because `MySpell`'s declaring script equals its attached script `ComposeScriptRecordUnit` omits the `[declared in …]` clause for both — so the quest renders a byte-identical line. The carrier was the line, and the line had two producers. What anchors it now is the footgun's own `[UNBOUND]` record header, asserted **immediately followed by** that line as one composed span: the header carries the FormKey, the EditorID and the plugin, which no other record can produce. Measured with your sabotage — the footgun's `VirtualMachineAdapter` dropped from `ScriptsWorld` (and `RecordsWithScripts` made consistent at 3, so the count canary does not mask the branch under test): at `2ae36cd` the line-only assertion stayed green off the quest; the anchored one is RED at `ScriptsWorldTests.cs:164`. Restored, grep-verified.

Every fixed or changed arm was sabotage-tested RED from a committed state, restored with a grep-verified `git checkout --`, tree clean after each:

| cell | sabotage | result |
|---|---|---|
| 1 | `PexWriter`'s baked initializer always Null | RED — the loose-layer arm (green before round 1's set pin) |
| 2 | `ScriptsWorld.AliasProp` forced to `Alias = -1` | RED — the loose-layer arm (likewise) |
| 3 | `HeldOpen`'s `'{path}'` deleted from its message | RED — the acquisition arm |
| 4 | the overlay dispose removed from `Read<T>` | RED — the leak reported by name |
| 5 | `OpenOverlay(path + ".nope")` | RED, but at the type assertion — `OpenOverlay` rejects a non-plugin extension with `ArgumentException` |
| 6 | `HeldOpen` `FileShare.None` → `FileShare.Read` | RED ×2 — both lock arms |
| 7 | `ReadTools.cs:1824`, the object-branch phrase | **RED — the wire arm.** This is the exact cell a round-2 reviewer measured GREEN before the fold |

**One honest gap, stated rather than papered over.** The HResult assertion cannot be isolated further: `Assert.Throws<IOException>` pins the exact type, so every producible fixture error (`FileNotFoundException`, `DirectoryNotFoundException`, `PathTooLongException`) is excluded before that line is reached, and no honest fixture yields a different plain `IOException` from this read. Its teeth are cell 6. The same was true of the message text it replaced, so nothing was traded away — but it is not a fully isolated arm.

The tier-trait fold was measured, not reasoned: `tier=stdio` selects the wire arm, `tier=integration` now excludes it (it did not before), and CI's `tier!=bridge` still selects it.

## Process observation — a fold's subject set should be derived, not hand-enumerated

`standards/HOUSECARL_REVIEW_ROUNDS.md` §2 already requires a **guard's** subject set to be derived from the surface it claims, never hand-enumerated. Nothing says the same of a **fold's** subject set, and this branch is the evidence for why it should: round 1 fixed the BCL-message assertion the reviewer pointed at and left its identical twin eleven lines away in the same file, which round 2 then found. A fold that changes an assertion of shape X should derive every assertion of shape X on the branch and dispose of the population, which is what the table above now does. Proposing it here rather than editing the standard, since that text is self-governing and yours to gate.

## Open for your gate

- **ADR 0003 rule 2 — the scripts family is briefly in both harnesses.** This branch's PEX-ROUNDTRIP arm re-homes an arm `ScriptPropertyCheckProbe` still runs, and the `RecordsWithScripts == 4` / unverifiable facts duplicate that probe's `RECORDS-COUNT` and `UNVERIFIABLE` arms. The guard is green, and correctly so: no `Converted-from:` marker was added, because adding one while the probes still exist would make `EveryConversionMarkerNamesAProbeThatIsGone_...` red. The overlap is real, deliberate, and a direct consequence of the ruled two-PR shape; it lasts exactly until PR 2 deletes `ScriptPropertyCheckProbe.cs`, adds the marker, and drops its baseline key.
- **The comment register beyond this branch.** ADR 0001 is prospective-immediate, so this branch complies now: the rationale moved to `docs/architecture/test-project-fixtures.md` and the five files' comment density fell from 26–44% to 18–23%. The test project's *neighbouring* files were written after the ADR in the register it retired, which is not this branch's to fix — recorded here as a standing item.

## Settled decisions the reviewers were given

`ruled` — Aaron decided it, in his own words; `inferred` — a session concluded it and Aaron never spoke. Reviewers honour `ruled`; `inferred` they may re-litigate.

1. **`ruled`** — the shape is a follow-up, not part of #484 (*"agree, its a follow up"*, 2026-09-03).
2. **`ruled`** — PR 1's scope: the Papyrus fixture and the file-lock harness as new machinery in their own files, nothing deleted.
3. **`inferred`** — the lane started off `main` in parallel with the #484 merge; its landing still waits for #484.
4. **`inferred`** — the `.pex` writer is ported, not referenced, because PR 2 may delete the probe that holds it.
5. **`inferred`** — the scripts world is an MO2-instance world so `housecarl_check` can drive it, with one wire-path smoke test proving reachability.
6. **`inferred`** — the fixture proofs are the fixture's own tests, not any of the 147 arms.
7. **`inferred`** — no CHANGELOG line: test machinery only, no shipped behaviour, no tool surface.
8. **`inferred`** — no ADR: the architectural decision this sits under is #486's own two-PR shape.
9. **`inferred`** — no `.csproj` change: `Mutagen.Bethesda.Pex` reaches the test project transitively.
10. **`inferred`** — `ScriptsWorld` exposes fixture-known values no test here reads; they are the vocabulary PR 2's arms spell, which PR 1's scope asks for.
11. **`inferred`** — the wire smoke test spins its own server; the shared one is deliberately unconfigured.
12. **`inferred`** — `ScriptsWorld` generates no corpus: the sweep never reads `CorpusRulebook`.
13. **`inferred`** — the lock proofs use `LoadOrderResolver.OpenOverlay` as "an engine read", keeping them about the harness rather than pre-empting PR 2.
14. **`inferred`** — the no-pex record keeps the probe's `HcSpNoPex` EditorID, which collides with the script name; the fixture is unchanged and the assertion is anchored to the composed reason line instead.

### DIFF
--- a/src/housecarl-mcp-tests/HeldOpenTests.cs
+++ b/src/housecarl-mcp-tests/HeldOpenTests.cs
@@ -41,9 +41,17 @@ public sealed class HeldOpenTests
     }
 
     /// <summary>The engine's single-plugin open, with the overlay disposed after — a memory-mapped overlay
-    /// left undisposed keeps a handle on the plugin for the rest of the process.</summary>
+    /// left undisposed keeps a handle on the plugin for the rest of the process.
+    /// <para><paramref name="use"/> must MATERIALISE everything it returns: the overlay is unmapped before the
+    /// value crosses the return. Why a mod or a record getter is refused rather than trusted:
+    /// <c>docs/architecture/test-project-fixtures.md</c>.</para></summary>
     static T Read<T>(string path, Func<ISkyrimModGetter, T> use)
     {
+        if (typeof(IModGetter).IsAssignableFrom(typeof(T)) || typeof(IMajorRecordGetter).IsAssignableFrom(typeof(T)))
+            throw new InvalidOperationException(
+                $"Read<{typeof(T).Name}> would return a view on an overlay this helper unmaps before the return — " +
+                "materialise what you need inside `use` (a string, a count, a copy) and return that instead.");
+
         var overlay = LoadOrderResolver.OpenOverlay(path, null);
         try { return use(overlay); }
         finally { (overlay as IDisposable)?.Dispose(); }
@@ -83,6 +91,24 @@ public sealed class HeldOpenTests
 
         var editorId = Read(world.PluginPath, mod => Assert.Single(mod.Weapons).EditorID);
         Assert.Equal(LockWorld.WeaponEditorId, editorId);
+
+        world.AssertNoHandlesLeft();
+    }
+
+    /// <summary>The materialise contract on <c>Read&lt;T&gt;</c>, made loud: a <c>T</c> that is still a view on
+    /// the overlay is refused before the plugin is opened at all.</summary>
+    [Fact]
+    public void ReadRefusesAValueThatWouldStillBeAViewOnTheUnmappedOverlay()
+    {
+        using var world = new LockWorld();
+
+        // Compiles identically to the materialising call below — which is the whole hazard.
+        var ex = Assert.Throws<InvalidOperationException>(
+            () => Read(world.PluginPath, mod => Assert.Single(mod.Weapons)));
+        Assert.Contains("materialise what you need inside `use`", ex.Message, StringComparison.Ordinal);
+
+        // Vacuity: the same read, materialised, is untouched by the guard.
+        Assert.Equal(LockWorld.WeaponEditorId, Read(world.PluginPath, mod => Assert.Single(mod.Weapons).EditorID));
 
         world.AssertNoHandlesLeft();
     }
--- a/src/housecarl-mcp-tests/HeldOpenTests.cs
+++ b/src/housecarl-mcp-tests/HeldOpenTests.cs
@@ -1,0 +1,102 @@
+using Mutagen.Bethesda;
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Records;
+using Mutagen.Bethesda.Skyrim;
+using Xunit;
+
+namespace HousecarlMcpTests;
+
+/// <summary>
+/// The file-lock harness's own proofs (#486 PR 1). They establish the two facts every arm built on
+/// <see cref="HeldOpen"/> rests on — that the hold really blocks an engine read, and that releasing it
+/// really unblocks the same read — plus the one that keeps a broken hold from passing quietly.
+///
+/// <para>Every arm here builds its OWN world and holds its OWN plugin. A lock is a mutation of shared state
+/// by any other name: a held file is not readable by anything else in the process, so holding one belonging
+/// to a shared fixture would make the world's readability depend on test scheduling. That is the same reason
+/// <see cref="RecordsWorld"/> gives for its mutating arms constructing their own instance.</para>
+/// </summary>
+public sealed class HeldOpenTests
+{
+    /// <summary>A throwaway one-plugin world this test owns outright, deleted when it is done with it.</summary>
+    sealed class LockWorld : IDisposable
+    {
+        public string Root { get; }
+        public string PluginPath { get; }
+        public const string WeaponEditorId = "HcLockW0";
+
+        public LockWorld()
+        {
+            Root = Path.Combine(Path.GetTempPath(), "hc-lock-tests-" + Guid.NewGuid().ToString("N"));
+            Directory.CreateDirectory(Root);
+
+            var key = new ModKey("HcLock", ModType.Plugin);
+            var mod = new SkyrimMod(key, SkyrimRelease.SkyrimSE);
+            var w = mod.Weapons.AddNew();
+            w.EditorID = WeaponEditorId;
+
+            PluginPath = Path.Combine(Root, key.FileName.String);
+            mod.BeginWrite.ToPath(PluginPath).WithLoadOrder(Array.Empty<ISkyrimModGetter>()).Write();
+        }
+
+        public void Dispose() { try { Directory.Delete(Root, true); } catch { /* temp cleanup best-effort */ } }
+    }
+
+    /// <summary>The engine's own single-plugin open, the read the hold is supposed to block.</summary>
+    static ISkyrimModGetter Read(string path) => LoadOrderResolver.OpenOverlay(path, null);
+
+    /// <summary>
+    /// While the hold is alive the engine read FAULTS — measured, not assumed: the type is
+    /// <see cref="IOException"/> and the message is Windows' own sharing-violation text. The second half is
+    /// the Q3 half and the reason this is not just <c>Assert.Throws</c> with a shrug: the failure mode worth
+    /// guarding is not "some exception", it is a read that comes back looking like a plugin with nothing in
+    /// it. <c>Assert.Throws</c> states that no value was returned at all.
+    /// </summary>
+    [Fact]
+    public void WhileHeldTheEngineReadFaultsLoudly_AndReturnsNoQuietlyEmptyPlugin()
+    {
+        using var world = new LockWorld();
+        using var hold = HeldOpen.Hold(world.PluginPath);
+
+        var ex = Assert.Throws<IOException>(() => Read(world.PluginPath));
+
+        Assert.Contains(Path.GetFileName(world.PluginPath), ex.Message, StringComparison.Ordinal);
+        Assert.Contains("being used by another process", ex.Message, StringComparison.Ordinal);
+    }
+
+    /// <summary>
+    /// After the hold is disposed the SAME read succeeds and yields the plugin's content. This is what stops
+    /// the arm above from passing on a file that was never readable in the first place — a fault before the
+    /// hold and a fault after it are indistinguishable from a working lock, until this runs.
+    /// </summary>
+    [Fact]
+    public void AfterTheHoldIsDisposedTheSameReadSucceeds()
+    {
+        using var world = new LockWorld();
+
+        var hold = HeldOpen.Hold(world.PluginPath);
+        Assert.Throws<IOException>(() => Read(world.PluginPath));
+        hold.Dispose();
+
+        var mod = Read(world.PluginPath);
+        var weapon = Assert.Single(mod.Weapons);
+        Assert.Equal(LockWorld.WeaponEditorId, weapon.EditorID);
+    }
+
+    /// <summary>
+    /// A hold that cannot be taken is itself loud, and names the path. The three probe arms this harness is
+    /// ported from each fail their own assertion when the lock cannot be acquired, for the reason
+    /// <see cref="HeldOpen"/> spells out: a test that silently proceeds without the lock asserts nothing while
+    /// still reporting a pass. A path that does not exist is the cheapest way to reach that branch.
+    /// </summary>
+    [Fact]
+    public void HoldingAPathThatDoesNotExistThrowsAndNamesThePath()
+    {
+        var missing = Path.Combine(Path.GetTempPath(), "hc-lock-absent-" + Guid.NewGuid().ToString("N") + ".esp");
+
+        var ex = Assert.Throws<InvalidOperationException>(() => HeldOpen.Hold(missing));
+
+        Assert.Contains(missing, ex.Message, StringComparison.Ordinal);
+        Assert.IsType<FileNotFoundException>(ex.InnerException);
+    }
+}
--- a/src/housecarl-mcp-tests/HeldOpenTests.cs
+++ b/src/housecarl-mcp-tests/HeldOpenTests.cs
@@ -40,7 +40,14 @@ public sealed class HeldOpenTests
             mod.BeginWrite.ToPath(PluginPath).WithLoadOrder(Array.Empty<ISkyrimModGetter>()).Write();
         }
 
-        public void Dispose() { try { Directory.Delete(Root, true); } catch { /* temp cleanup best-effort */ } }
+        /// <summary>The delete is LOUD, and that is the guard on this file's whole subject. A binary overlay
+        /// is memory-mapped: an undisposed one keeps a handle on the plugin, and Windows will not delete a
+        /// directory holding an open file — so the delete failing IS the leak, reported. Swallowed, the one
+        /// signal that proves the file-handle claim is thrown away, and dropping the overlay dispose in
+        /// <see cref="Read{T}"/> leaves the suite green (measured). The other worlds in this project keep a
+        /// best-effort swallow for a different reason — their deletes race a killed server process and they
+        /// claim no file-handle property; this one does.</summary>
+        public void Dispose() => Directory.Delete(Root, true);
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


src/housecarl-mcp-tests/HeldOpenTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        public void Dispose() { try { Directory.Delete(Root, true); } catch { /* temp cleanup best-effort */ } }
+    }
+
+    /// <summary>The engine's single-plugin open, with the overlay disposed after — a memory-mapped overlay
+    /// left undisposed keeps a handle on the plugin for the rest of the process.</summary>
+    static T Read<T>(string path, Func<ISkyrimModGetter, T> use)
+    {
+        var overlay = LoadOrderResolver.OpenOverlay(path, null);
+        try { return use(overlay); }
+        finally { (overlay as IDisposable)?.Dispose(); }
```
Avick3110, line 57: **Low — `Read<T>` disposes the overlay before returning `T`.**

The `finally` unmaps the overlay and *then* the value crosses the return boundary, so the helper is only safe for a `T` that is fully materialised inside `use`. Today's one value-returning call site returns `EditorID` (a real `string`), so it is green.

The hazard is the next caller. `Read(path, mod => Assert.Single(mod.Weapons))` compiles identically and reads the mapped view after `Dispose` — an `ObjectDisposedException` at best, an `AccessViolationException` that takes the whole runner down at worst, with no assertion failure pointing at the cause. The product states this contract explicitly on `OverlaySession` ("the service reads fields off a fetched body before its session disposes"); this helper repeats the shape without restating it, and it is the seam PR 2's three dialogue lock arms are built on.

A sentence on `Read<T>` saying `use` must materialise everything it returns would close it.

src/housecarl-mcp-tests/PexWriter.cs
```diff
@@ -0,0 +1,62 @@
...
+            Name = name,
+            TypeName = typeName,
+            DocString = "",
+            Flags = PropertyFlags.Read | PropertyFlags.Write | PropertyFlags.AutoVar,
+            AutoVarName = $"::{name}_var",
+        };
+        var data = initInt is int v
+            ? new PexObjectVariableData { VariableType = VariableType.Integer, IntValue = v }
+            : new PexObjectVariableData { VariableType = VariableType.Null };
+        var backing = new PexObjectVariable { Name = $"::{name}_var", TypeName = typeName, VariableData = data };
```
Avick3110, line 40: **Low — the initializer branch is Integer-typed regardless of the declared scalar type.**

`Auto` writes `VariableType.Integer` for any non-null initializer, but stamps the backing variable's `TypeName` from the caller's `typeName`. So `AutoScalar("MyFlag", "Bool", 1)` produces `::MyFlag_var` with `TypeName = "Bool"` and `VariableData.VariableType = Integer` — a pairing no Papyrus compiler emits.

It matters because the product keys off `VariableType`, not the declared type: `PapyrusDecompiler.InitText` would render `Bool Property MyFlag = 1 Auto`, and `ScriptPropertyCheck` would read `HasInitializer = true`. An arm built on that fixture asserts product behaviour against a `.pex` shape the game never produces, and stays green while claiming to model a defaulted Bool.

Nothing is wrong in this branch — the fixture only ever passes `Int` — but this is new public machinery that PR 2's 28 script-property arms will spell against, and the parameter name `initInt` is currently the only thing carrying the restriction. Either constrain it (`AutoScalar` refusing a non-`Int` `typeName` with an initializer) or say so in the doc comment.

src/housecarl-mcp-tests/ScriptsWorldTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        // The unverifiable attribution, spelled the way the sweep composes it — names the .pex it looked for.
+        Assert.Contains($@"'Scripts\{ScriptsWorld.MissingScript}.pex' is not on disk", r.Text, StringComparison.Ordinal);
+
+        // The WHOLE object-branch line, spelled from fixture-known values. The scalar branch emits the same
+        // "declared but NOT bound" phrase, so a fragment of it is satisfied by MyChance; only this line is.
+        Assert.Contains(
+            $"{ScriptsWorld.ObjectProperty} (Spell) on script {ScriptsWorld.ChildScript}"
+            + " — declared but NOT bound → None at runtime (HIGH: object/form type — the silent no-op)",
+            r.Text, StringComparison.Ordinal);
```
Avick3110, line 168: **Low — the whole object-branch line has two carrier records in this fixture, not one.**

The fold ruled out the *scalar branch* as a rival carrier, which is right. It did not rule out the second record.

`HcSpAliasQuest` attaches `HcSpChild` through `QuestFragmentAlias.Scripts` and binds nothing, so every declared property comes back unbound for it too. For `MySpell` the declaring script equals the attached script, so `ComposeScriptRecordUnit` skips the `[declared in …]` clause (the thing that distinguishes `InheritedThing`) and renders, verbatim:

```
  ! MySpell (Spell) on script HcSpChild — declared but NOT bound → None at runtime (HIGH: object/form type — the silent no-op)
```

…in the quest's section as well as the footgun's.

Concrete consequence: drop the footgun's `VirtualMachineAdapter` from the fixture, or break `IHaveVirtualMachineAdapterGetter` sweeping for weapons in the product, and this arm stays green off the quest. `TheServiceSweeps…` catches that; the wire arm does not.

As a reachability canary for the object render branch it is doing its job, so this may well be fine as-is — but the PR body's derived-population table lists this assertion as *disposed of*, and its carrier is still wider than one record. Either the table's disposition wants the qualifier, or the assertion wants the record header line prefixed to it (`[UNBOUND] … 'HcSpFootgun') in HcSp.esp`), which no other record can produce.

Avick3110:
Rebased onto main @ 46a59c1 after #484 merged; tests 850 → 857 (the 7 new facts, measured both sides); ci-all 130/130; wire unchanged (35 tools, byte-identical dump).

Avick3110:
## Gate folds — all three, one commit each

Tip `2ae36cd` → `aff8291`. Append-only; no force-push. Every arm sabotage-tested RED from a committed state, restored with a grep-verified `git checkout --`, tree clean after each.

### 1 — `3924559807`, `Read<T>` disposes the overlay before returning `T` — `f1c4752`

The contract is now a sentence on the helper (`use` must materialise everything it returns), with the reason in `docs/architecture/test-project-fixtures.md`, and the helper **refuses** a `T` that is an `IModGetter` or an `IMajorRecordGetter` before it opens the plugin at all. Refused rather than only documented because the failure is silent — measured at `2ae36cd`, `Read(path, mod => Assert.Single(mod.Weapons))` returned a record whose `EditorID` then read back *correctly* off the unmapped view and the arm reported a pass. Nothing warned; that silence is the finding. The doc states the guard's limit: it is a static check on `typeof(T)`, so a caller who erases the type walks past it.

- **Arm:** `ReadRefusesAValueThatWouldStillBeAViewOnTheUnmappedOverlay` — the refused call and the materialising call, which compile identically.
- **RED:** guard condition → `if (false)` ⇒ `Assert.Throws() Failure: No exception was thrown`.

### 2 — `3924561464`, the initializer branch is Integer-typed regardless of the declared scalar type — `e549e8f`

Constrained, the stronger of your two options, since PR 2 spells 28 arms against this machinery. `Auto` refuses a non-`Int` declared type carrying an initializer with an `ArgumentException` naming the property, the type and the value. Public signatures unchanged; nothing the fixture passes today is refused.

- **Arms, one per branch:** `TheWriterRefusesABakedInitializerOnANonIntScalar` (`AutoScalar("MyFlag", "Bool", 1)`, whole composed sentence asserted) and `TheWriterStillBakesAnIntScalarInitializerAndItRoundTrips` (`"Int"` with `5` still writes and reads back as `VariableType.Integer` under `TypeName = "Int"`).
- **RED:** at `2ae36cd`, before the constraint ⇒ `Assert.Throws() Failure: No exception was thrown`. After the fold, the condition inverted (`!string.Equals` → `string.Equals`) ⇒ RED ×2, both branches.

### 3 — `3924563619`, the object-branch line has two carrier records — `aff8291`

You were right that the fold ruled out the scalar branch and not the second record. The wire arm now asserts the footgun's `[UNBOUND]` header **immediately followed by** the `MySpell` line as one composed span — measured adjacent, since `MySpell` sorts first among the footgun's object-type findings. The header carries the FormKey, the EditorID and the plugin, which no other record can produce. The PR body's derived-population table now carries the correction with what the carrier was and what anchors it.

- **RED, your own sabotage:** the footgun's `VirtualMachineAdapter` dropped from `ScriptsWorld`, with `RecordsWithScripts` made consistent at 3 so the count canary does not mask the branch under test. At `2ae36cd` the line-only assertion **stayed green off the quest** (measured before the fix). After the fold: RED at `ScriptsWorldTests.cs:164`, `Assert.Contains() Failure: Sub-string not found`.

### Gates at `aff8291`

| gate | result |
|---|---|
| `dotnet build housecarl.sln -c Release` | 0 errors (1 warning, `RecordsRemedyGrammarTests.cs:66`, pre-existing on `main` from `5e38fc0`, not in this branch's diff) |
| `dotnet test --filter "tier!=bridge"` | 856 → **859**, 0 failed |
| `ci-all` | **130/130 ALL PASS** |
| `tools/list` off the built server | **35, md5 `f3ffa21301c698d3f55184c463cc345a` — byte-identical to base `46a59c1`** (dumped from a throwaway worktree at the base, since removed) |

Three new tests: `ReadRefusesAValueThatWouldStillBeAViewOnTheUnmappedOverlay`, `TheWriterRefusesABakedInitializerOnANonIntScalar`, `TheWriterStillBakesAnIntScalarInitializerAndItRoundTrips`.

No new review rounds — directed folds on a PR that has had its rounds plus your review (CLAUDE.md §5 #11).

Issue #487, opened 2026-09-03
[GAP] "What is live in this cell": Persistent/Temporary returns the winning override own list, not the engine additive union — #342 stage-2, split out


Avick3110:
**Split out of #342 on Aaron's routing go 2026-09-03** (verbatim: "I agree with the rest of the routing proposals"; ground: the row-8 disposition walk, advisor lane) **so the milestone carries stage-1 only.** The walk's finding, quoted: "#342 cannot be represented by one milestone. Stage-1 is FIRM/milestoned; stage-2 is a routing call. The milestone claims membership for the whole issue. Either stage-2 splits out as its own issue or the milestone over-claims." This issue is stage-2; #342 keeps stage-1 and its `houseCARL 2.0` milestone. This one is filed **unmilestoned**.

## The gap — the walk's #342 stage-2 row, quoted

| # | Routing line (row-8, short) | 10(b) tag — site | 7(b) — ground · promotion condition |
|---|---|---|---|
| **#342 s2** | "What's live in this cell" — FormKey-keyed union, per-reference winners | `engine-survives` — resolver read model; `housecarl_records` inherits it | **taken-into-W5**, highest priority inside it. Ground: the only Q3 silent-wrong-answer in the population, external-reported, engine-level, reproduced on the shipped 2.0 tool. See the top flag. Sanctioned `effect_chain` posture applies — naive concat triple-counts, so this is real design, not a patch. |

## Why it was surfaced at tagging time — the walk's top-flag row, quoted

The 2026-08-29 (4) ruling: "An engine-level instance feeding live ARR decisions surfaces to Aaron at tagging time, not at the charter." This is one of the three that qualify.

| # | What is silently wrong | Site (read) | Why ARR |
|---|---|---|---|
| **#342** | A cell's `Persistent`/`Temporary` returns the **winning override's own list**, not the engine's additive union. A cell with content reads as empty. | Resolver read model; `housecarl_records` reproduces it wholesale (the 1.x trio that filed it is condemned, the model is not) | A **Q3 silent wrong answer**, already proven on a real order: an external reporter was told a Dawnstar cell was empty when it was not, chasing a missing caravan. ARR is a 3800-plugin order with occlusion/worldspace patches — the exact shape that triggers it. Any "what's in this cell" answer given for ARR is suspect today. |

[...]

PR #495, opened 2026-09-03
The tree form names which providers declare a record's children, and says when none do

### LINKED ISSUE #485: [GAP] housecarl_records: no declarers note for owned-child fields — the precise tier (which plugins declare children in a CELL/DIAL/WRLD field) was deleted at the 1.x cut; the tree form shows each plugin's content instead
Filed by the fold session on PR #484 (the 1.x cut, `Closes #468`), on Aaron's ruling. Not a live-session report — a capability the 2.0 surface does not carry, recorded so it is not a silent deletion.

## Goal

Ask, of a record that owns child records, **which other plugins declare children in this field** — and be told nothing at all when none do.

Child records are declared per plugin and the game assembles them from every plugin that declares them: a CELL owns its placed references, a DIAL its INFO lines, a WRLD its cells. A plugin can override the parent for an unrelated reason — occlusion, lighting, music — and carry no children of its own. Reading that winner then reports an empty collection the game fills. Reproduced on a real order (#342): `008EB5:Skyrim.esm` reads `Temporary = [list: 0 item(s)]` at its winner (`Occlusion.esp`, override depth 42) while `Skyrim.esm`'s own body carries 201. A caller auditing "what is in this cell" through the winner got a confidently wrong answer.

## What fell short

houseCARL had two tiers of answer for this, and only the cheap one survives.

**The cheap tier** — live today on `housecarl_records`, on every read that renders a child-bearing field. It states what the index settles for free: that other plugins touch this record, and that this read did not open their bodies to see what they declare. It is honest and it costs nothing, but it cannot distinguish "they declare children here" from "they don't".

**The precise tier** — deleted at the cut. It named WHICH plugins declare content in each annotated field (collection-shaped fields by name, singular-shaped ones by count), reported separately any touching plugin whose body could not be read, and — the half the cheap tier can never do — said **nothing at all** when no other plugin declares anything there. It cost no extra fetch: it read the bodies the conflict tree had already materialised.

**Why it was unreachable.** The precise tier fired only under `conflictTree == true` in the shared record render. The only surface that ever set that flag was 1.x's `read_record` with `conflict_tree=true`. `housecarl_records` has no `conflict_tree` lever and every one of its call sites passes `false`, so from the day `records` became the read surface the tier had no caller. It was deleted on PR #484 rather than left on the tree as scaffolding.

## Workaround used

`housecarl_records project={"form":"tree"}` answers the same *question* in a different shape, and it is pinned by tests today: it names every plugin touching the record and shows what each one holds in the child-bearing field, so a caller can read the declarations off the per-plugin content.

**Where it degrades.** The tree form renders **differences**, not declarations — a lower plugin whose content in that field is identical to the winner's is omitted from the diff, so its declaration is not visible as a declaration. In the case the annotation exists for (a winner carrying 0 beside a base carrying 201) the difference *is* the declaration, so the tree form covers it. The gap is the identical-content case, and the "nobody declares anything here" answer, which the tree form does not state in its own voice.

## Ideal shape

Two candidates, either of which restores the answer on the 2.0 surface:

1. **A lever on `records`** that fires the precise tier — the flag the tier already reads, given a caller.
2. **The tree form emits the declarers sentence automatically** — it has already paid for every touching body, so the sentence is free where the form is already being rendered, and it needs no new parameter.

**Promotion condition, verbatim:** "promote if the tree form proves insufficient in daily use — Aaron's data".

## Provenance

Measured on PR #484's branch by the build session and independently by a round-1 reviewer, and escalated under CLAUDE.md §4 framing (b) as one of two capability losses the cut creates. Aaron's ruling, 2026-09-03: **"agree accept, delete and file the issue"**.

Avick3110:
Closes #485

Aaron's "we need it back at least in part to get us going" (2026-09-03) is what this PR restores; the remainder of the honest capability gap is #487, by his same words ("the true honest capability gap is something that does need closed before 2.0 is released").

## What the branch does

`housecarl_records project={"form": "tree"}` now states, for a record whose type owns child records, which providers DECLARE children per child-bearing field — naming them for a COLLECTION field, counting them for a SINGULAR one — and says so in its own words when none do. It reads the provider bodies the tree has already fetched for its diff, so it adds no fetch. The block renders above the diff on text, as a `child_declarers` array on json, and as a `child_declarers?` column plus a manifest note in the `to_file` artifact. The default read is unchanged; its existing note points here.

One `housecarl_records` description sentence changed on the wire (35 tools, 177387 chars canonical, unchanged across every fold below):

> 'tree' (every provider of each record in priority order, winner last, each diffed against the reference pole — default the winner; takes fields=. On a record type that OWNS child records — a cell's placed references, a topic's INFO lines, a worldspace's cells — it also states, per such field, which providers DECLARE children there (a COLLECTION field) or how many do (a SINGULAR one, e.g. Cell.Landscape), and says so when none do)

## The pre-green fold (before the rounds)

The build session ended at branch-green with a decision record; a Sonnet fold session — the Opus build session that would ordinarily have run it was lost to API overload after both pre-green reviews were written — picked up cold from that record, the two reviews and the diff. Recorded in the decision record §7.1/§7.2. It folded one HIGH and five MEDIUMs: the block growing a response past `max_chars` with no cap check on either text or json (H1/M6, one shape in two transports, folded together per the fold-population rule); the lead reaching json and the artifact rather than text only (M2/M3, folded in the same commit as one coherent "the precise tier's transport completeness" change); the SINGULAR field's empty answer keeping its own voice instead of borrowing the COLLECTION negative's plural; and the tree description matching what a SINGULAR field actually gets. Gates at branch-green: build 0 errors, 868 tests, ci-all 130/130.

## The rounds

Four rounds' worth of independent agents, all in isolated worktrees on Opus, none told what an earlier round found, all seeded with the settled-decisions list and its tags. Findings are per reviewer.

**Round 0 — rebase and gate re-run.** Rebased onto `origin/main`, clean. Build 0 errors / 91 warnings (all pre-existing), 878 tests (868 branch + 10 that arrived on `main` from PR #488, each accounted for by name), ci-all 130/130, wire 35 tools / 177387 chars.

**Round 1 — 0 high, 2 medium + 7 low (seeded-A); 0 high, 2 medium + 7 low (gate-B, deliberately contextless).** Folded, with arms: the text lane repeating the 290-char framing line once per ROW where json states it once per response (found independently by both, same 20.1%-of-bytes measurement) — arm `TheLeadIsStatedOnceAcrossMultipleRowsInText_NotPerRow`; the block's own TAIL having no cap check on either transport (measured +37 text, +536 json) — arms `ATextRowsDeclarersBlockTailAloneCanTripMaxChars_...` and `Json_TheResponseLevelLeadIsDroppedRatherThanOverrunningCap_...`; a cut block falling through into an empty diff header plus a second redundant notice — arm `ACutDeclarersBlockEndsTheRow_NoEmptyDiffHeaderAndNoSecondCutNotice`; the new architecture note duplicating three source comments instead of replacing them (DOC_HYGIENE §8 rule 2), with the two copies already drifted; the SINGULAR zero reading in the COLLECTION plural voice — arm `ASingularFieldNobodyCarriesIsCountedZero_...`; the COLLECTION overflow naming no remedy; `row_schema` missing the `?` optional marker — arm `Artifact_ThePreciseTiersRowSchemaNamesTheChildDeclarersColumnAsOptional`; two stale comments; a false CHANGELOG claim; an overclaiming probe docstring.

*Refused:* that the three names shown at the COLLECTION cap are the lowest-priority declarers rather than the most informative end — the reviewer's own words were "not reproducible on a three-plugin fixture", and the ordering matches the touchers list's own convention in the same render, so inverting it would be a new inconsistency, not a fix. *Dropped* on the reviewer's own recommendation: the dead `conflict_tree=true` lane computing and discarding declarers (settled item 6, #486's half). Gates after: 885 tests, ci-all 130/130, wire unchanged.

**Round 2 — 0 high, 4 medium + 4 low (seeded-A); 0 high, 4 medium + 6 low (seeded-B).** Folded, with arms: the branch's own CHANGELOG entry contradicting an older #484-cut bullet in the same `## Unreleased` section; `WriteTreeRow`'s two `truncated=true` assignments shadowing each other, the nodes-branch one unguarded by anything else (found independently by both) — arm `Json_ANoChildrenRowsNodesCutStillSetsTruncated_NotCoveredByTheDeclarersLeadCheck` at the measured 1077/1078 boundary on a WEAP-shaped row; rows after the lead-carrying one sitting unlabelled — arm `TheSecondRowsBlockCarriesTheShortHeaderNotTheFullLeadAgain`; the overflow remedy baked into `DeclarersNote` so json and the artifact carried a pointer to the format the reader was already in — arms `TheOverflowRemedyNamesFormatJson_TextOnly` and `Json_TheOverflowIsNeverCapped_AndCarriesNoRemedyText`, driven against a hand-built 5-declarer row; the documented `fields=` narrowing having zero guard (deleting the filter left the whole 885-test suite green) — arms `FieldsNarrowsTheBlockToTheNamedTopLevelField_NotTheWholeType` and `FieldsWithABracketedPathInsideAChildBearingFieldYieldsNoBlockAtAll`; and the four new cut-notice sentences never entering the wrong-lever grammar grid — arm `TheChildDeclarersCutNoticesNameNoWrongLever`. Plus four measured comment/baseline corrections.

*Deferred, not folded:* two of `AppendChildDeclarers`' three cap checks were individually removable with the suite green (both reviewers, independently, both hedging "not asserting it is wrong — only that nothing measures it"). *Not folded:* a CHANGELOG cross-reference the doc already states correctly. *Dropped:* the `conflict_tree` lane again; and a text lead reserve, on the reviewer's own verdict that the field loop's check always catches it. Gates after: 893 tests, ci-all 130/130, wire unchanged.

**Round 3 (final, hard cap) — 0 high, 2 medium + 2 low (seeded-A); 0 high, 1 medium + 4 low (seeded-B).** Both reviewers ran the full suite and ci-all green, drove the built server over stdio to confirm the changed sentence is on the wire, re-measured every pinned numeric boundary on the current code (all correct, none stale from rounds 1–2), and ran scripted sabotage sweeps — A's 15 mutations across 4 batches with the predicted RED set stated before each batch and matched exactly; B's 18-case sweep carrying a known-RED canary.

Folded, with arms:

- **B MEDIUM1 — the text block's tail notice was false in every case it could fire, and it swallowed the row's diff.** The field loop checks the cap before each line and returns from inside, so the tail is reachable only when every declarer line was written — nothing was ever cut there, yet it printed "[child declarers cut at max_chars=N]" and pointed at `fields=` to narrow a block that was already complete. The caller read that return as "end this row", so on a multi-provider row what actually went was the DIFF, silently. Reproduced: CellC (sole provider) false-claimed across caps 806–845, CellF (3 touchers) across 812–864 with its whole diff section absent. Fixed: the tail returns "the row ends here" without a claim, and the caller — which knows whether there is a diff to lose — names the nodes it dropped or says nothing. Arms, one per branch: `ASoleProviderRowWhoseCompleteBlockEndsPastTheCapClaimsNothingWasCut`, `AMultiProviderRowWhoseCompleteBlockEndsPastTheCapNamesTheDiffItLost`, `ABlockCutMidWayStillSaysTheDeclarersWereCut`. All three RED-checked from the committed state.
- **A MEDIUM1 — the text lane's framing line had no `max_chars` reserve** (settled item 13, `inferred`, re-litigated with a measurement the round-2 dismissal lacked). The entry check passed at 294 chars and then wrote 290 more unconditionally: at `max_chars=295` a sole-provider row returned a 683-char body. Fixed by reserving the framing line's own length the way `JsonWire.RenderTree` reserves the identical sentence and the cheap tier reserves its own clause. Boundary re-derived on the current render: 587 refuses the framing, 588 is the first cap it fits inside. Arms both directions: `TheFramingLineIsReservedAgainstMaxChars_NotWrittenPastIt`, `TheFramingLineRidesAtTheFirstCapItFitsInside`. The reserve subsumes the old entry check, which disposes of round 2's deferred item by deletion rather than by a new guard.
- **B LOW1 — the cut notice was a render literal in three byte-identical copies inside one 36-line method**, and the wrong-lever guard's two hand-chosen caps reach only one of them. All five instances of that composition in the file (the branch's three, plus the pre-existing `nodes` and `delta lines` siblings — population derived by grep, all five disposed of) now go through one `AppendCutNotice`, so the guard's single harvested hit is sufficient by construction rather than by the copies happening to agree.
- **A LOW1 / B LOW3 (first half) — three source comments restated sections of the new architecture note** (DOC_HYGIENE §8 rule 2, the same rule round 1 folded once already). `ReadSentences`' class summary, `JsonWire`'s seven-line reservation essay and `AppendChildDeclarers`' summary are now pointers. `TreeRow`'s `<para>` was kept: it states a constructor contract the note does not carry, which §8 puts in the code.
- **B LOW3 (second half) — 29 review-round discovery citations on the branch's own added lines** (DOC_HYGIENE §7), three of them narrating what an earlier version of the comment used to say. Population derived by script over the branch diff and disposed of in full — 7 with the behavioural commit, 22 with the docs commit; re-running the derivation returns nothing. Pre-existing citations elsewhere (~110) untouched, per §7's no-bulk-scrub rule.
- **B LOW4 — the CHANGELOG entry said the block covers every child-bearing field of the type** without saying `fields=` narrows it, one clause after naming `fields=` as a lever. It now says so.

*Refused / dropped, with grounds:*

- **A LOW2 — a row whose block was cut still counts toward `rendered N of M`.** Dropped on the reviewer's own read ("my own read is: drop it"): it matches the pre-existing delta-cut arm in the same file, so it is consistent rather than novel.
- **A MEDIUM2 — the wrong-lever guard's population is hand-enumerated.** Not folded here; it is a construction gap outside this branch. See the open items below.

Gates after round 3's two folds: build 0 errors / 91 warnings (all pre-existing), **898 tests** (+5 arms), **ci-all 130/130**, **wire 35 tools / 177387 chars, byte-identical** to the pre-fold dump.

### Derived populations, with counts

| Population | Derived by | Count | Disposed |
|---|---|---|---|
| Instances of the text lane's cut-notice composition in `RecordsTools.cs` | `grep 'cut at max_chars='` over `src/`, excluding tests | 5 (3 branch-added, 2 pre-existing siblings; a 6th in the chain form has a different remedy tail and is not this shape) | all 5 routed through one `AppendCutNotice` |
| Review-round discovery citations on branch-added lines | scripted regex over `git diff f307f7f..HEAD` | 29 | all 29 (7 + 22 across the two fold commits); re-derivation returns empty |
| Multi-line comment runs the branch added to shipped source | scripted scan of the branch diff | 17 | 4 judged to restate the architecture note; 3 trimmed to pointers, 1 (`TreeRow`) kept as a contract |
| Call sites of `AppendChildDeclarers` after its signature changed | `grep` over `src/` | 2 (one production, one test) | both updated; compiler-enforced |

### Gates, before and after

| Gate | branch-green | round 0 | after round 1 | after round 2 (`4c7d036`) | after round 3 (`2972cb2`) |
|---|---|---|---|---|---|
| build | 0 errors | 0 errors, 91 warnings | 0 errors, 91 warnings | 0 errors, 91 warnings | 0 errors, 91 warnings |
| `dotnet test` | 868 | 878 | 885 | 893 | **898** |
| `ci-all` | 130/130 | 130/130 | 130/130 | 130/130 | **130/130** |
| wire | 35 / 177387 | 35 / 177387 | 35 / 177387 | 35 / 177387 | **35 / 177387** |

All warnings are pre-existing CA1416 / xUnit2030, unchanged by this branch. Every gate above was re-run independently by this close-out session at `4c7d036` before folding and again at `2972cb2` after.

## Settled decisions, with tags

`ruled` = Aaron's own words, cited to a record. `inferred` = a session concluded it and Aaron never spoke; a reviewer may re-litigate these. Items 1–21 are the decision record's §8 list; 22–25 were added by the rounds (reconstructed from the round fold reports — the round-3 reviewers were seeded with them and cite item 25 by number).

| # | tag | decision |
|---|---|---|
| 1 | `ruled` | The tier comes back now, at least in part; the full capability gap is separate (#487). |
| 2 | `ruled` | The earlier deletion was accepted and filed as #485. |
| 3 | `ruled` | #485 is milestoned houseCARL 2.0. |
| 4 | `inferred` (advisor) | The shape is #485's option 2, "the tree form emits the sentence", unless the build measured it cannot state the negative without an extra fetch. It measured that it can. |
| 5 | `inferred` (advisor) | No new tool parameter, source/versus pole, or path-grammar vocabulary. Held — reviewers confirmed. |
| 6 | `inferred` (advisor) | The render halves #486 owns are not touched or revived. Held. |
| 7 | `inferred` | The negative is a sentence, not silence. |
| 8 | `inferred` | The block renders above the diff, not inside it. |
| 9 | `inferred` | json and the `to_file` artifact carry it, from the shared `WriteTreeRow`. |
| 10 | `inferred` | `MergeCollection` / `SingleResolved` are not restored; one lead line carries both shapes. |
| 11 | `inferred` | The block is emitted for every row whose type owns children, sole-provider rows included. |
| 12 | `inferred` | The singular line reads "carried by N provider(s)", not the deleted tier's "N other plugin(s)" — the tree has no subject body for "other" to be relative to. |
| 13 | `inferred` | **No `max_chars` reserve for the block, matching the tree lane's existing behaviour.** Round 3 re-litigated this with a measurement and it is now folded — see the open items below. |
| 14 | `inferred` | The block follows `project.fields` the way the cheap tier follows its rendered paths. Round 3 tried to find a divergence between the two tiers and could not. |
| 15 | `inferred` | Commits say `Refs #485`; whether the PR carries a closing keyword is the PR author's and Aaron's call. This PR carries `Closes`. |
| 16 | `inferred` | The pre-green fold ran on Sonnet, the Opus build session having been lost to API overload. |
| 17 | `inferred` | H1 and M6 are one shape in two transports and were folded in one commit. |
| 18 | `inferred` | M2 and M3 were folded in that same commit as one coherent transport-completeness change. |
| 19 | `inferred` | json's per-field `declaring` list is deliberately NOT capped at `DeclarerNameCap`; the response-level check bounds the document instead. |
| 20 | `inferred` | The architecture note is a new file, not a merge into the concurrent #486 branch's fixture doc. |
| 21 | `inferred` | The true merge-base correction was re-derived and confirmed, not taken on trust. |
| 22 | `inferred` (round 1) | The COLLECTION cap shows the LOWEST-priority declarers first, matching the touchers list's own winner-last convention in the same render. A UX design call, not a defect. |
| 23 | `inferred` (round 1) | The dead `conflict_tree=true` lane computing and discarding declarers is out of scope — #486's half. |
| 24 | `inferred` (round 2) | The entry and per-field cap checks in `AppendChildDeclarers` were mutually redundant, deferred rather than fixed blind. **Round 3 disposed of this** by replacing the entry check with the reserve, which is independently RED-checkable. |
| 25 | `inferred` (round 2) | `RemedyHarvest`'s tree probes never reach a child-bearing type because `RecordsWorld` — a frozen shared fixture — has none, so the branch's cut notices were covered by a hand-written replacement arm instead. Known gap, left for whoever next touches that fixture. |

## Open for Aaron's gate

1. **The wrong-lever guard's population is hand-enumerated, and this is the third instance of that class** (round 3, reviewer A, MEDIUM). `RecordsRemedyGrammarTests`' harvest reaches the tree form only through `RecordsWorld`, which has no child-bearing record type, so this branch's cut notices never entered the wrong-lever grid and the grid stayed green over the gap. Round 2's answer was to hand-write one replacement arm against this project's own fixture. CLAUDE.md §3's third cornerstone — elevated on this branch's own date — names exactly this shape, and §8 says the disposal is "make the shape unwritable, or **file the construction as a gap**". `RecordsWorld` is frozen (settled item 25), so the branch cannot make it unwritable; the second half was never done. I have not filed the issue, because filing on the public repo is an outward-facing act this session was not asked to take. **Recommended: a `[GAP]` issue against `RemedyHarvest`'s population** — derive its tree-form subject set from the record types the records surface can render a declarers block for, rather than from whatever `RecordsWorld` happens to contain. One word from you and it gets filed. (Round 3's `AppendCutNotice` fold is the by-construction fix at the one level this branch *could* reach: the guard's single harvested hit is now sufficient because there is one literal, not three.)

2. **A class recurred across rounds 2 and 3: "a conditional this work added is covered for by a neighbour, so it cannot be RED-checked on its own."** Under §5 #11's relaxed rule the instances are folded with their arms and the class is named here rather than folded silently.
   - Round 2's instance (the text lane's entry vs per-field checks) is **disposed of by deletion** — the reserve replaced the entry check, and it RED-checks on its own.
   - Round 3's surviving instance is **`JsonWire.cs`'s `truncated = true;` inside the `child_declarers` cut branch**: deleting it leaves the suite green, because the nodes loop always runs on a non-error row and sets the flag one statement later, with `leadOverCap` behind that. I did not fold it. The seam: it is a *defensive* assignment against an invariant ("a non-error row has at least one node") that nothing structurally guarantees, so deleting it trades a real safety margin for a testable one; and the only fixture that would isolate it is a hand-built row with declarers and zero nodes — a shape production cannot produce, which makes the arm theatre rather than a measurement. The honest options are to leave it as documented-and-unpinned, or to make the invariant structural so the branch cannot exist. That is a design call, not a fold.

3. **Settled item 13 was the builder's knowing choice, and round 3 overturned it.** The record's own words were: "this is the one place the branch knowingly matched existing behaviour instead of improving it" — the tree lane checks the cap per row and per node and already appends an unbounded toucher list past it, so the block matched the lane. Round 3 measured the actual consequence (a 683-char body at `max_chars=295`), which round 2's reviewer had dismissed on a wrong premise, and the reserve is now folded. Flagging it because it is an `inferred` decision the rounds reversed rather than upheld: if you would rather the block stay symmetric with the toucher list's existing overrun than with json's reserve, this is the commit to say so on (`b76f000`).

4. **Two SKILL.md files still instruct callers to use the deleted `conflict_tree: true` 1.x parameter** (round 2, out-of-scope note). Pre-existing, outside this diff, not folded and not filed. Minor, non-functional doc staleness.

## Notes on how the rounds were run

Rounds 1–3 were conducted by a fresh session, not the branch's author, per §5 #11 and `standards/HOUSECARL_REVIEW_ROUNDS.md`; that session reached its context seam after launching round 3, and this close-out session — which also did not write the code — triaged and folded round 3 and opened the PR. Every reviewer ran in its own isolated worktree. Every sabotage RED-check in every round ran from a committed state and was restored under `HOUSECARL_ALLOW_DIRTY_RESTORE`, with the restore verified by grepping for strings the fold introduced rather than assumed. Round 3 was the third of the hard cap of three; there is no round 4, and everything above under "Open for Aaron's gate" is listed rather than folded.

---

## Gate folds

Aaron's gate review posted four findings; the advisor triaged all four as confirmed. A fresh fold session (which did not write the code and did not conduct the rounds) executed them, one atomic commit each, each RED-checked from the committed state and restored under `HOUSECARL_ALLOW_DIRTY_RESTORE` with the restore verified by grep. No new rounds (§5 #11's fourth stop rule). PR open item 2 — `JsonWire`'s `truncated = true` in the `child_declarers` cut branch — was looked at and left: it is Aaron's design call, not a fold.

| # | Finding | Commit | RED check |
|---|---|---|---|
| 1 | [3927945079](https://github.com/Avick3110/houseCARL/pull/495#discussion_r3927945079) — a complete sole-provider row is still marked truncated, and spills | `0592742` | Restored the unconditional `truncated = true` → `ASoleProviderRowWhoseCompleteBlockEndsPastTheCapClaimsNothingWasCut` fails on its flipped last assertion (1 of 78 arms), the other 77 green |
| 2 | [3927946742](https://github.com/Avick3110/houseCARL/pull/495#discussion_r3927946742) — the `!declarersCut` guard leaves the diff loss unnamed on the cut branch | `4afbdec` | Restored the guard → 3 fail: the flipped `ABlockCutMidWayStillSaysTheDeclarersWereCut` (CellF 660), the renamed `ACutDeclarersBlockEndsTheRow_…` (CellF 600), and the new framing-reserve arm (CellF 625). The sole-provider control (CellC 700) stays green — that is what makes it the control |
| 3 | [3927948208](https://github.com/Avick3110/houseCARL/pull/495#discussion_r3927948208) — the `unreadable` overflow elides names with no remedy in text | `b0d3411` | Reverted the condition → `TheOverflowRemedyAlsoNamesFormatJsonForAnUnreadableOverflow_SingularShape` fails. `BothOverflowsOnOneLineStillCarryASingleRemedy` stays green under that sabotage, which is its "one remedy, not two" control role |
| 4 | [3927949679](https://github.com/Avick3110/houseCARL/pull/495#discussion_r3927949679) — the reserve is measured one property too early | `6f3fcbc` | Removed the `truncated` term → the property arm goes red at **caps 1889–1910** (22 caps, overshooting 22 chars down to 1 — the ~21-byte window the finding predicted), plus the re-pinned 1911 arm. Removed the `Framing.RootClose` term instead → red at **caps 1909–1910**. Both terms RED-check independently |

**Finding 4 needed one term more than the finding named, and this is the surfacing rather than a silent fold.** With only the `truncated` property reserved, CellC's json tree still ended 2 chars past cap at `max_chars=1909` and 1 char past at 1910 — the framing still overshooting, in a narrower window. The missing term is the root close, which the chain form's own cap checks already add (`Size + RootClose`) and this one did not; it is the same omission `MeasureFraming`'s docstring in the same file already records ("the comparison that decides whether a notice fires at all stayed one character short"). Both terms are now in the check, both measured off the writer. If you would rather the fold had stopped at the term the finding named and left the 2-char window, this is the commit to say so on (`6f3fcbc`).

The bespoke `MeasureDeclarersLeadReserve` became `MeasureRootProperty(Action<Utf8JsonWriter>)` and both reserves share it, so a reserve cannot drift from the write it stands for. The architecture note's reserve sentence described one term; it now names all three, which makes its "invariant framing does not [overshoot]" sentence true again.

### Derived populations, with counts

Per `HOUSECARL_REVIEW_ROUNDS.md` §2's fold-population rule — each fold derived every assertion of the shape it changed and disposed of the population, not the instance the finding named.

| Population | Derived by | Count | Disposed |
|---|---|---|---|
| `truncated` / `rowsTruncated = true` assignments the branch added | `git diff f307f7f..HEAD -- src/` | 5 | 1 is the finding (fixed); 1 is PR open item 2 (drops entries when it fires — Aaron's call, untouched); 3 set it only on a real loss. No other site sets it on a row that lost nothing |
| Assertions that a complete sole-provider block at the tail is marked truncated | grep for every spill/`truncated` assertion in `RecordsOwnedChildTests.cs` (14 sites) | 2 (both CellC @845) | one flipped to `DoesNotContain`; its twin `ATextRowsDeclarersBlockTailAloneCanTripMaxChars_…` moved to CellF @830, the row shape where the tail genuinely loses a diff |
| Assertions that a row carries no nodes-cut notice | `grep '"\[nodes cut'` over `src/housecarl-mcp-tests` | 3 | the 2 on a multi-provider row flip to `Contains`; the 1 on a sole-provider row stays, correctly |
| Appends of `DeclarersOverflowRemedy` in `src/` | grep | 1 | the sole caller, by design (`[NoClaims]` says the caller appends it, never `DeclarersNote`) — the one site fixed |

### Measured boundaries, re-derived on the current render

The fold session re-measured rather than trusting the pinned numbers: a sweep of caps 250–900 over CellC and CellF for text, 1840–1960 for json (probes run and deleted, not committed).

- **text** — CellC (sole provider): framing refused at ≤587, block cut 588–805, complete at 806+. CellF (3 touchers): framing refused at ≤625, block cut 626–811, complete at 812+.
- **json** — CellC's tree is 1911 chars. With the corrected reserve, **1911** is the last cap that drops the note and spills and **1912** the first that keeps it, at which point the document is 1911 chars — inside the cap rather than one line past it. The two pinned boundary arms move from 1884/1886 to 1911/1912.

The new arm `Json_WhereverTheResponseLevelLeadIsPresent_TheDocumentIsWithinCap` asserts the property those two caps are samples of, over every cap from 1871 to 1952 (the boundary ±40), and asserts the band straddles the boundary so it cannot pass vacuously.

### Gates after the gate folds

| Gate | after round 3 (`2972cb2`) | after the gate folds (`6f3fcbc`) |
|---|---|---|
| build | 0 errors, 91 warnings | 0 errors, 91 warnings (all pre-existing CA1416 / xUnit2030) |
| `dotnet test --filter "tier!=bridge"` | 897 | **903** (+6 arms) |
| `dotnet test` (unfiltered, incl. the 1 bridge test) | 898 | **904** |
| `ci-all` | 130/130 | **130/130** |
| wire | 35 tools / 177387 chars | **35 tools / 177387 chars, byte-identical** |

The 897/903 figures are the `tier!=bridge` count CI runs; the table above under "Gates, before and after" quotes the unfiltered count, which is one higher. Both were re-measured on this branch rather than carried forward.

**Wire.** `tools/list` was dumped off the built server over stdio (`scripts/tool-names/dump-tools-list.py`) at `6f3fcbc` and again from a scratch worktree built at `2972cb2`, and compared with `cmp`, not against a remembered constant: **byte-identical**. Against a scratch worktree of `origin/main` (`f307f7f`) exactly one tool differs — `housecarl_records`, in the `project.form` description — which is the one sentence this PR documents at the top. No fold touched a tool description.


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1668,6 +1669,14 @@ static string RenderRecordsTree(IReadOnlyList<LoadOrderService.TreeRow> rows, in
             for (int i = 0; i < row.Touchers.Count; i++)
                 sb.Append("    ").Append(i + 1).Append(". ").Append(row.Touchers[i])
                   .Append(i == row.Touchers.Count - 1 ? "  (winner)" : "").Append('\n');
+            if (AppendChildDeclarers(sb, row, cap, ref declarersLeadWritten, out bool declarersCut))
+            {
+                truncated = true;
```
Avick3110: **A complete sole-provider row is still marked truncated, and spills.**

`truncated = true` is set unconditionally on a `true` return, including the case this branch itself defines as "the block is complete and the row lost nothing" — the sole-provider row whose comment two lines down says it "loses nothing — it says nothing."

It does say something, just not here: `truncated` reaches `TreeResponse`, which writes a JSONL artifact to the results store and re-renders with `spilled: complete result`.

Scenario — the branch's own boundary, `Tree(_w.CellC, maxChars: 845)` (pinned by `ASoleProviderRowWhoseCompleteBlockEndsPastTheCapClaimsNothingWasCut`): all four declarer lines are written, the row has no diff, and `TrimEnd('
')` brings the body to 844 chars, *inside* the cap. Nothing was dropped, yet the caller gets a disk artifact and is told the answer did not fit. The arm's own name says nothing was cut while its last assertion pins `spilled: complete result` — the same falsity the tail-notice fold removed, moved into a different sentence.

Rows *after* this one are already covered by the loop's own top-of-loop cap check (which sets `truncated` and emits `rendered N of M`), so the flag is load-bearing here only when this row actually lost something.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1668,6 +1669,14 @@ static string RenderRecordsTree(IReadOnlyList<LoadOrderService.TreeRow> rows, in
...
+                // The block ran to completion and ended past cap: what this row loses is its DIFF, so that is
+                // what the notice names. A sole-provider row has no diff, and loses nothing — it says nothing.
+                if (!declarersCut && row.Nodes.Count > 1) AppendCutNotice(sb, "nodes", cap);
```
Avick3110: **The `!declarersCut` guard leaves the diff loss unnamed on the cut branch.**

Round 3's B MEDIUM1 established that a row ending here loses its DIFF and that the response must name it. That holds only when the block ran to completion. When declarer lines really were dropped, the row loses its diff too — and this guard suppresses the notice, so nothing names it.

Scenario — `Tree(_w.CellF, maxChars: 660)`, the branch's own arm `ABlockCutMidWayStillSaysTheDeclarersWereCut`, which asserts `DoesNotContain("[nodes cut")`: a 3-toucher row prints `[child declarers cut at max_chars=660 …]` and stops. The reader is told the *declarers* were cut and handed `fields=` as the remedy; the whole `diff (field deltas vs …)` section is gone with nothing in the response saying a diff existed. Same on the framing-reserve branch at line 1739, where none of the block was written at all.

Mitigated by the spill (the complete result does reach a file), which is why this is a low rather than a repeat of B MEDIUM1 — but the asymmetry is arbitrary: the caller knows `row.Nodes.Count > 1` in both branches.

src/housecarl-mcp/ReadSentences.cs
```diff
@@ -79,6 +72,80 @@ internal static class ReadSentences
...
+    {
+        string head = shape == OwnedChildShape.Singular
+            ? $"{CarriedBy} {declaring.Count} provider(s)"
+            : declaring.Count == 0 ? NoDeclarers
+                : $"{DeclaredBy} {string.Join(", ", declaring.Take(DeclarerNameCap))}"
+                  + (declaring.Count > DeclarerNameCap ? $" (+{declaring.Count - DeclarerNameCap} more)" : "");
+        return unreadable.Count == 0 ? head
+            : head + $"; {unreadable.Count} provider(s) {CouldNotRead} "
+              + $"({string.Join(", ", unreadable.Take(DeclarerNameCap))}"
+              + (unreadable.Count > DeclarerNameCap ? ", …" : "") + ")";
```
Avick3110, line 146: **The `unreadable` overflow elides names with no remedy in text.**

This line drops names past `DeclarerNameCap` behind a bare `", …"`, but the text caller appends `DeclarersOverflowRemedy` only for a COLLECTION field's `declaring` overflow:

```csharp
if (cd.Shape == OwnedChildShape.Collection && cd.Declaring.Count > ReadSentences.DeclarerNameCap)
    sb.Append(ReadSentences.DeclarersOverflowRemedy);
```

So the structurally identical elision one clause earlier in the same sentence gets the pointer and this one does not.

Scenario: a cell whose `Landscape` (SINGULAR — the `Collection` guard never fires) or whose `Temporary` with 2 declarers has 5 providers whose bodies could not be read. Text renders `…; 5 provider(s) could NOT be read (A.esp, B.esp, C.esp, …)` and stops. json and the artifact carry the full `unreadable` array; the text caller is never told that.

The `[NoClaims]` reasoning for `DeclarersOverflowRemedy` ("the caller appends it") holds either way — the caller just needs the second condition.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -740,19 +772,26 @@ public static string RenderTree(IReadOnlyList<LoadOrderService.TreeRow> rows, in
...
+                if (WriteTreeRow(w, row, ms, cap, levers)) rowsTruncated = true;
+                if (row.ChildDeclarers.Count > 0) anyDeclarers = true;
                 rendered++;
             }
             w.WriteEndArray();
             w.WriteNumber("rendered", rendered);
+            // The framing line is stated once per response, and its reserve MUST be checked before `truncated`
+            // is written: a Utf8JsonWriter cannot un-write a property once appended.
+            w.Flush();
+            bool leadOverCap = anyDeclarers && ms.Length + DeclarersLeadReserve >= cap;
```
Avick3110: **The reserve is measured one property too early.**

`leadOverCap` is checked against `ms.Length` here, but `child_declarers_note` is written at line 794 — *after* `w.WriteBoolean("truncated", …)` at 792. The boolean's own encoded bytes (~21 with `Indented = true`) are never reserved, so the framing sentence can still land past `cap`.

Scenario: a response where `ms.Length + DeclarersLeadReserve == cap - 1` at this line. The check passes, `"truncated": false` is appended, then the note on top — document ends ~21 bytes past `cap` with `truncated` reported false, so nothing spills either.

Small in bytes, but it falsifies the claim the branch's own architecture note makes: "Content lines still overshoot the cap by at most one line, which is the whole lane's existing tolerance; **invariant framing does not**." Reserving `DeclarersLeadReserve` plus the `truncated` property's own `MeasureUnit`-style cost restores it, and keeps the reserve derived rather than hand-counted.

Issue #494, opened 2026-09-03
[GAP] The renderer-free arms deleted with CheckErrorsProbe.cs and ScriptPropertyCheckProbe.cs are re-homed in the test project, structure-keyed (#492's shape)


Avick3110:
**Ruled by Aaron, 2026-09-03** (*"agreed"*, on the advisor's recommendation) at #486 PR 2's round-1 CLAUDE.md §4(b) stop. Milestone `houseCARL 2.0`. Sequenced **with #492, before the seven conversion PRs**.

## The class

#486 PR 2 deletes the eight zero-caller render halves, with the compiler as the checklist — Aaron's ruling at PR #484's gate. The compiler names the sites that CALL a deleted member, and every one of those 197/204 sites was classified (a)/(b)/(c) and carried to a disposition.

Phases 2 and 3 then deleted `CheckErrorsProbe.cs` and `ScriptPropertyCheckProbe.cs` **whole**. A whole-file deletion also removes every arm in the file the compiler never named — arms that drove `ErrorCheck.Run` / `ScriptPropertyCheck.Run` or a product class directly and had nothing to do with the render halves. Those arms went with no disposition and no replacement.

Round 1 of PR 2's review rounds found it. The sharpest instance was measured there: `ErrorCheck.UntypedOwnerVariableData` — the #207 fix, about forty lines of subtle shipped code where over-exempting stops the owner form being swept and under-exempting brings #207 back — had **no guard anywhere** on the branch.

This is the third cornerstone's shape (CLAUDE.md §3, elevated the same day): a population taken as given — the file — rather than derived.

## The counts

| file | arms | compiler-named lines | coupled (dispositioned) | RENDERER-FREE |
|---|---|---|---|---|
| `CheckErrorsProbe.cs` | 127 | 110 | 61 | **66** |
| `ScriptPropertyCheckProbe.cs` | 42 | 28 | 15 | **27** |
| **total** | **169** | **138** | **76** | **93** |

**93 renderer-free arms.** Three of them — `OWNER-RANK`, `OWNER-DATA-CHECKED`, `OWNER-RANK-TOTAL` — are
**already re-homed on this branch** as `OwnerRankExemptionTests`, at Aaron's ruling: they guarded #207's
`UntypedOwnerVariableData` exemption, about forty lines of subtle shipped code that had no guard anywhere
after the deletion. **90 remain**, and they are what the port charter covers.

[...]

Issue #492, opened 2026-09-03
[GAP] Test-side by construction: facts assert structure keyed by record, sentences are tested once at their catalogue, prose literals in fact tests are refused by a derived guard


Avick3110:
**Ruled by Aaron, 2026-09-03** ("yes on the by construction charter"), on the advisor's recommendation after the anchoring class recurred four times on #486 PR 2 and on PR #488 twice the same day. Routed row 7, **before the seven conversion PRs** (which would otherwise write hundreds of arms in the old shape); milestone `houseCARL 2.0`. The test-side twin of #440 option B.

## The class this closes

A test asserts a fragment of prose the product composes. It stays green when the thing it names is broken, because: the span can never occur by fixture construction (vacuous `DoesNotContain`); only a prefix of the sentence is pinned (a tail sabotage stays green); a second render branch or a second RECORD in the fixture emits the same line; or the expected text is carried by something else (a BCL message, a seed's own plugin name). Found by mutation on `claude/fixtures` (round 2 + Aaron's gate) and `claude/render-halves` (pre-green reviews 1a/1b: 3 HIGH, 3 MEDIUM, all this class). The interim rule — whole composed lines anchored to the record, sabotage per branch and per record, the fold-population sweep (`HOUSECARL_REVIEW_ROUNDS.md` §2) — is a discipline, not a construction: it is enforced by eye and by rounds, and the rounds keep finding it.

## The construction — three parts, each load-bearing

1. **Facts assert structure, keyed by record identity.** Every engine fact is asserted on the JSON document (`families.*`) or the result object, keyed by the record/plugin it is about. No span to prefix-pin; a second record cannot satisfy it; a vacuous negative has nowhere to hide.
2. **Sentences are tested once, at their home.** Rendered wording lives in the sentence catalogue (`ReadSentences` and its `[MustState]` machinery). Each sentence gets ONE test there, against a fixture-known, independently-written second spelling (the case-law rule), plus its reachability from each surface that emits it. Wording is proven once, not re-proven per fact.
3. **Prose literals in fact tests are refused by a guard, by construction.** A derived-population check over `src/housecarl-mcp-tests` (every `Assert.Contains`/`DoesNotContain`/`StartsWith`… whose expected value is a string literal or interpolation over rendered text, outside the catalogue tests) fails the build. A new arm in the old shape cannot land; the population is derived from the test sources, never hand-listed.

[...]

PR #496, opened 2026-09-03
The eight zero-caller render halves are deleted and the engine facts their arms reached are tested in the test project

### LINKED ISSUE #486: [NOTE] Read-family render halves: delete the eight zero-caller members and write the 147 engine arms they hid — fixtures PR first (Papyrus fixture, file-lock harness), then the deletion PR; sequenced after #484, before #440

Avick3110:
Closes #486

PR 2 of 2 on issue #486 (PR 1 was #488, the fixtures PR, merged). PR #484's gate found eight render and
service members left with no shipped caller when the eleven 1.x read and check tools came off the wire, and
Aaron ruled they go now, with the compiler as the checklist. This deletes them and writes the engine facts
their old-harness arms were reaching as tests against the surfaces that render those facts today.

Built across four phases in three sessions; phases 2–4 ran on Sonnet because Opus was refused by API
overload. The review rounds below are the quality gate on that.

## The census

Deleting the eight (plus their cascade `Wire.Close`, `DialogueWire.SumConditioned`,
`DialogueWire.AppendStandingLimits`, and the `conflictTree` render chain) broke the old harness at **200
distinct file:line sites, 204 compiler complaints** — matching #486's own pre-measurement of 197 sites, with
3 more that were arity fixes on surviving renderers rather than deletions. Three read-only agents classified
every site:

| bucket | sites | what it means |
|---|---|---|
| (a) DIES | 17 | the assertion's subject is the deleted tool's own response; nothing survives it |
| (b) OWES A FRESH TEST | 175 | it calls a deleted member, but asserts an engine or shared-machinery fact a surviving tool still drives |
| (c) DIAGNOSTIC ONLY | 12 | the deleted call appears only in a failure-detail string, never in the boolean |

The one disagreement with #486's own count (17 (a) vs its 21) was resolved in favour of (b): four
sweep-refusal and off-order arms are emitted TODAY by `Wire.RenderCheck` / `JsonWire.RenderCheck`.

The 175 (b) sites consolidate to about 65 distinct FACTS — one test per fact, not per old assertion.
**60 xUnit facts** land here: 57 new, plus the 3 owner-rank arms ported on Aaron's round-1 ruling (below).

## Every (b) site's disposition

The complete old-arm-to-new-test table is in the branch's decision record
(`dev/session-handoffs/SESSION_HANDOFF_2026-09-03_486-render-halves_BRANCH-GREEN_decision-record.md` §3,
private corpus), row by row — every (b) site is one row ending in a test name, an "already covered" naming
an existing test, or a reclassification to (a) with its reason. The counts:

| originating probe | (b) sites | fresh tests | already covered | reclassified (a), with reason |
|---|---|---|---|---|
| `CheckErrorsProbe.cs` | 110 | 28 (`CheckErrorsFamilyTests`) | 2 (`CheckMergeProbe`'s CAP-LADDER and RESERVE-DECLARED) | 0 |
| `ScriptPropertyCheckProbe.cs` | 27 | 14 (`ScriptsFamilyTests`) | 0 | 0 |
| `EpochGuardProbe.cs` | 19 | 5 (`EpochCheckSweepTests`) | 2 | 1 (E3, covered on the live records surface) + 3 arity sites fixed in place |
| `BulkPrimitivesWave3Probe.cs` | 14 | 3 (`RecordsOffOrderPathTests`) | 7 | 4 (`Wire.RenderPluginFile`'s banner — see the open items) |
| `DialogueInfoOrderProbe.cs` | 6 | 4 (`DialogueFamilyTests`) | 0 | the helper carrying ~8 more arms was fixed in place onto the surviving `AppendInfoOrderView` |
| `DialogueValidateGuardProbe.cs` | 3 | 1 (`DialogueFamilyTests.FactV1`) | 0 | 0 |
| `CheckMergeProbe.cs` | 2 | fixed in place | — | 0 |
| `SkyPatcherHarness.cs` | 1 | — | — | 1 |

No site was silently dropped, stubbed, or left both broken and untested. `#207`'s owner-rank exemption is
the one (b) fact that was NOT in the compiler's list and is carried anyway — see the round-1 ruling.

## What the compiler's list does not measure — the renderer-free arms

Two probe files were deleted WHOLE (`CheckErrorsProbe.cs`, 2436 lines; `ScriptPropertyCheckProbe.cs`, 836).
The compiler named the arms that CALLED a deleted member. The arms in those files that called nothing
deleted went with the file and are invisible to that list. Round 1 found this; the population was then
derived rather than estimated:

| file | arms | compiler-named lines | coupled | renderer-free |
|---|---|---|---|---|
| `CheckErrorsProbe.cs` | 127 | 110 | 61 | **66** |
| `ScriptPropertyCheckProbe.cs` | 42 | 28 | 15 | **27** |
| **total** | **169** | **138** | **76** | **93** |

An arm counts as coupled when a compiler-named line falls in its own span, **or** when it calls one of the
eleven file-local helper methods whose bodies contain such a line — the second clause is what two earlier
hand-tuned estimates (48 and 97) each got wrong in opposite directions. Both round-2 reviewers rebuilt the
derivation independently from the branch and reproduced it to within one arm each (67/27 and 65/26 against
66/27), in the safe direction.

**Three of the 93 are ported in this PR** (the #207 cluster, on Aaron's ruling). The remaining **90 are
chartered to #494**, with the derived list, the counts and the per-cluster notes in that issue's body.

## Two re-measurements

- **#394's allocation baseline** (`ALLOCATION-SECOND-FAMILY-DOES-NOT-WAIT-ITS-TURN`) took its number from
  the deleted single-family renderer, and is re-derived through the merged renderer over an errors-only
  `CheckSweep`. Round-2 reviewer A instrumented both ends and measured it **byte-identical**, not merely
  "no loss" by assertion: `mergedFloor` 2397 and `errorsWholeBody` 3610 at both `498e4f4` and the branch.
- **The cap/floor sweeps.** The errors family's instance is fact 22; the scripts family's is `FactS13`
  (caps 1..12000 across three fixture shapes, both transports). `CheckMergeProbe`'s existing
  `[matrix]`/`CapSweep` machinery already covers the merged surface across all three families, unmodified.

## Round 0 — the two pre-green reviews, folded

The build session ended at branch-green with two independent Opus reviews written but NOT folded; a fresh
session folded them first. **9 findings: 2 high, 4 medium, 3 low. 8 folded, 1 refused.**

The four HIGH/MEDIUM findings were one class — an assertion that cannot fail, or that a second record
satisfies — so the fold derived the whole population of that shape by script rather than fixing the named
instances: **120 assertion sites** across the six new test files, every one dispositioned (13 fixed over 9
arms; the other 107 classified, with the reason, into anchored-whole-line, fixture-computed-value,
negative-with-a-named-positive, structural-absence-with-a-positive, wording-retraction, subject-scoped, and
loop-variable). Nine sabotage cells from the committed state, canary included, all RED.

Refused: `FactD2` narrows its original because the fixture has no uncontested topic — ground, the reviewer's
own mitigation is the whole answer: the two strings appear nowhere in shipped code, so a second topic would
add a second render of a string that has no writer, and would change an `Order` sequence three other facts
read. The fixture half of a sixth finding was refused too (the `EpochGuardProbe` world's `badName` setup is
still written into `loadorder.txt` and two arms fingerprint that world's shape); its comment half was folded.

Full record: `dev/session-handoffs/render-halves-scratch/round0-report.md`.

## Round 1 — one seeded reviewer, one gate-replica

**14 findings: 2 high, 7 medium, 5 low. 10 folded, 2 refused, 2 carried to this gate, 1 escalated.**

- **HIGH — `Fact13` was green under an under-count.** The two reviewers disagreed and the disagreement was
  settled by measurement: over-count RED, under-count **GREEN** at cap 400, where the errors family's json
  carries `plugins: []` and both agreements read `0 == 0`. Under-reporting is the Q3 direction. Folded —
  `Fact13` walks 400..20000 for a cap producing a PARTIAL cut and asserts there, restoring three conjuncts
  the conversion had dropped. Seven sabotage cells from the committed state, canary included, all RED,
  including the under-count that was green before.
- **HIGH — the renderer-free arms.** Escalated under §4 rather than folded; ruled below.
- **MED — five zero-caller production members.** Folded over a re-derived population: **seven**, not five
  (`TextBoundaryReserve` and `SampleOf` are cascades). All seven deleted.
- **MED — three files cite a document that does not exist.** Folded: `docs/architecture/check-family-tests.md`
  is written, with its DTO-lane population derived from the source rather than asserted.
- **MED — `docs/architecture/test-project-fixtures.md` (LIVING) falsified in three places.** Folded.
- **MED/LOW — `SkyPatcherPostState` was not zero-caller.** **Refused as a defect, carried as disclosure:**
  `SkyPatcherHarness.Run` and the `skypatcher-post-state` verb went with it, and that verb was the Wave-1
  empirical-gate artifact — but `dev/plans/SKYPATCHER_DISTRIBUTOR_TOOL_PLAN_2026-07-08.md` records "EMPIRICAL
  GATE CONFIRMED by Aaron 2026-07-09", so the gate has fired, and the replay core stays reachable through the
  surviving `skypatcher-layer` harness. Round-2 reviewer B confirmed independently that the 2.0
  `source={overlay:"skypatcher",state:"post"}` pole builds its own replay context and never routed through the
  deleted member. Stated here because removing a named empirical-gate instrument should not live only in a
  code comment.
- **MED/LOW — `ResolveTree` and the public `ResolveRead(FormKey, …)` overload.** Carried to this gate beside
  `ReadPluginFile` rather than decided: all three are the same question, and deciding two and handing over the
  third would be the inconsistency.
- **LOW — refused:** the #394 baseline re-derivation is unobservable on a green run. Ground: the reviewer's own
  analysis is the ground — a larger `errorsWholeBody` makes the arm easier to satisfy, so the re-derivation is
  weaker rather than wrong, and producing the literal pair needs the deletion temporarily reverted. Round 2
  produced the pair anyway, by instrumentation; it is identical.
- Also folded: two probe headers describing moved arms, a dead `GapSigs`, the relocated `.pex` helpers'
  dropped `AutoScalar` caveat, and an editing artifact in `DialogueWorld.cs`.

Full record: `dev/session-handoffs/render-halves-scratch/round1-report.md`.

### The §4 escalation, and Aaron's ruling

Round 1's HIGH-1 was escalated rather than folded: the two whole-file deletions took renderer-free arms with
them, and three clusters had no replacement — `#207`'s owner-rank exemption (~40 lines of shipped `ErrorCheck`
code with **no guard anywhere on the branch**), both sweeps' `SweepScope` narrowing, and eleven
`SweepEmission`/`BoundedBody` allocation unit arms.

**Aaron ruled: "agreed"** — charter the bulk, port the #207 cluster now. Executed:

- `OwnerRankExemptionTests`, three facts, its own fixture, driven at DTO level on `ErrorCheck.Run` and
  **structure-keyed per #492's shape** — the assertions read `Dangling`'s FormKeys and `MissingMasters`' names
  off the result object; not one touches a rendered sentence. Both directions of the exemption are armed and
  both were proven RED from a committed state with a known-RED canary beside them: under-exempt (#207 returns)
  RED on 2 of 3, over-exempt (a genuinely broken owner goes silent) RED on 2 of 3. Both round-2 reviewers
  reproduced both directions independently. Reviewer B also found the port **stronger** than the arms it
  replaces: the old `OWNER-RANK` read `ownerRep is null || !ownerRep.Dangling.Any(...)`, so a vanished report
  passed it vacuously; the new one takes the report through `Assert.Single` and adds a present-faction control.
- The renderer-free population derived (above) and chartered as **#494**.
- ADR 0005 given a dated addendum narrowing its claim to the compiler-named population.

## Round 2 — two fresh seeded reviewers

**9 findings: 0 high, 2 medium, 6 low, 1 nit. 8 folded in full, 1 folded in part, 0 refused outright.**

**Why nothing was refused outright.** Every finding reproduced against the source on first attempt, three of
them wider than the reviewer stated — so the honest disposition was to fold, and the triage signal shows up
instead as three folds whose SUBJECT SET is larger than what was reported, and one remedy declined with its
reason. The one partial: A-LOW-3 asked for the generator's `AutoScalar` to match its hardened sibling. The
comment is corrected (it described a hardened copy as if it were not); the **hardening is refused** — a
refusal is a conditional, the generator has no unit harness to arm both branches from, and the review-rounds
standard says an unarmable conditional is not a thing a fold adds. Both call sites pass `"Int"`, and the file
retires with its own conversion. The reason is in the code, not only here.

| # | sev | finding | fold |
|---|---|---|---|
| A-MED-1 | MED | `DialogueFamilyTests.FactV1`'s block extractor bleeds across seeds; two of three CK-parity assertions cannot fail | **FOLDED**, `SeedBlock` |
| B-MED-1 | MED | eight in-code comments name deleted members in the present tense; two are unresolvable `<see cref>`s | **FOLDED** over a derived population — **ten**, not eight |
| A-LOW-1 | LOW | `test-project-fixtures.md` cites the wrong commit for the ADR-0003 rule-2 closure | FOLDED |
| A-LOW-2 / B-NIT-3 | LOW/NIT | five (reviewer A) / four (reviewer B) newly added test helpers are dead | FOLDED, five, derived |
| A-LOW-3 | LOW | `CheckMergeProbe`'s `AutoScalar` mis-states what its hardened sibling does | FOLDED (comment); hardening refused, with grounds |
| A-LOW-4 | LOW | `test-project-fixtures.md`'s opening carries PR 1's forecast, now falsified | FOLDED, and its unnamed twin with it |
| A-LOW-5 | LOW | the DTO-lane population misses one hand-shaped DTO | FOLDED |
| B-LOW-2 | LOW | `DialogueInfoOrderProbe.RenderOrderOnly`'s render cap silently dropped 10× in the rewrite | FOLDED |

**A-MED-1, in detail, because it is the one that mattered.** `FactV1` is the re-homed `RENDER-SCOPE-PIN` and
the only guard that DERIVES the checked-subrecord set from `Missing*Defaults` rather than restating it. Its
span helper terminated on the next BLANK line, and the merged dialogue render emits its seed heads
contiguously — so the first blank line in a three-seed response comes after the LAST seed, and every block
spanned every seed under it. Reviewer A measured it: deleting the DLVW verdict outright left the arm GREEN,
satisfied by the DLBR seed's identical `CK-parity: OK`; and `quest CK-parity: OK` contains that substring, so
the branch block was in the same position. Only the `ENAM`/`TNAM` halves bit, and only by the accident that
those two tokens are unique in the region — `DNAM`, which both sentences carry, anchored nothing.

Folding the terminator then turned the test RED for a second, deeper reason the old span had been running
past: the START was "the first line containing the marker", and the DLVW seed's scope sentence names
"quest (QUST)" as the record to validate for a graph surface, so the quest block resolved to the wrong seed
entirely. Both halves are fixed; the start is anchored to a head line. **Four sabotage cells from the
committed state — each of the three seed verdicts in turn, plus a canary already proven RED by hand — all
RED, with the file restored byte-identically (md5 checked).** Two of those three were green before the fold.

**The derived populations, since three of the folds are wider than what was reported.**

- The stale-comment population was never derived by round 1's fold, which derived three others. Deriving it —
  every identifier declared in the changed `src/housecarl-mcp` files at the base and declared nowhere at the
  tip, then every mention across `src/`, `docs/`, `standards/`, `plugin/`, `.github/`, classified by hand as
  historical or present-tense — found **ten**, including two outside the reviewer's eight
  (`JsonWire.cs:311`'s "see RenderRecord" and `RecordsPluginFileSourceTests`' provenance line). Re-running the
  derivation over the fixed tree leaves five mentions, all correctly past-tense.
- The dead-helper population was derived to a **transitive closure**, because
  `Accounting → AccountingLine → ErrorsSection` and `HeadLine → ErrorsSection` form a cluster that only calls
  itself. Five members, ~45 lines. Worth recording: the script first reported `HeadLine` LIVE, because
  `CheckExcludeGrammarTests` declares its own private `HeadLine` and a name-keyed reference count cannot tell
  two same-named members apart — checked by hand against the qualified call sites.
- The doc-number population was derived across both architecture documents rather than the three instances
  named, which found the twin of A-LOW-4: "28 script-property arms" appears **twice**, and both were wrong
  (14). `197 → 200/204`, `147 arms → 60 facts`, the SHA replaced by a commit subject (it would not have
  resolved after a `--rebase` merge either), `8 DTO sites → 9`.

Reviewer reports: `dev/session-handoffs/render-halves-scratch/round2-results/{rh-r2-a.md,rh-r2-b.md}`.

### The rounds are closed at 2 of 3, by the class-recurrence stop rule — and the class is for Aaron

Round 1's class was **a population taken as given rather than derived** (four instances in one round). Round 2
returns it three times:

1. **B-MED-1** — the round-1 fold derived three populations by script and left this one undone, and the
   reviewer's own list of it was in turn short by two.
2. **A-LOW-5** — a stated derivation ("derived from every `Result(...)` call … 8 sites") whose own subject set
   was short by a site that does not call `Result(`.
3. **A-MED-1** — round 0's fold classified `LineAfter`'s 9 sites as "scoped to one subject's own block" by
   reading the helper's doc comment. The doc comment was wrong. The classification was asserted, not measured.

That is CLAUDE.md §5 #11's two-consecutive-round recurrence, so this is where the rounds stop; the third
instance is an escalation trigger, not a fix freeze, so each instance is folded with its arm and the class
comes here instead of into a round 3.

**The seam, stated and not solved.** Every one of these is a population that a fold or a document CLAIMED to
have derived. The claim is prose; nothing holds the claim to the tree, so a derivation that is short by two —
or asserted from a doc comment — reads exactly like one that is complete, and the next round finds the
remainder one instance at a time. CLAUDE.md §3's third cornerstone says a detector for a recurring class is a
bug report against the construction, so **no mechanism is proposed here**. The class, its five instances
across two rounds, and the seam are Aaron's to rule on.

## Gates

| measure | `main` @ `f307f7f` | this tip |
|---|---|---|
| `dotnet build houseCARL.sln` | 0 errors | 0 errors |
| `dotnet test --filter "tier!=bridge"` | 859 | **919** (+60 — the 57 new facts and the 3 ported owner-rank arms) |
| `dotnet test`, unfiltered | — | **920** |
| `ci-all` | 130/130, ALL PASS | **128/128, ALL PASS** |
| `tools/list` | 35 tools | 35 tools — **the two dumps are byte-identical** |
| residue `probeFiles` | 150 | 148 |

**The `tools/list` claim is a comparison, not a constant.** `origin/main` was built in a scratch worktree the
same way, both servers were driven over stdio with `initialize` / `notifications/initialized` / `tools/list`,
and the two sorted, pretty-printed dumps were compared with `cmp` — identical, 201,553 bytes. (They also
reproduce the md5 the decision record recorded, but round 1 showed that constant does not reproduce under a
different dumper, so the comparison is the claim.)

The countdown 150 → 148 is exactly the two whole-file deletions, each with its `Converted-from:` marker landed
and its `harness-residue-baseline.json` key removed in the same commit. `ci-all` 130 → 128 is exactly those
two files' `check-errors-guard` and `script-property-check-guard` rows retiring; every other probe file this
PR touched keeps its row. All seven `HarnessResidueTests` guards pass, so the baseline is confirmed against
the source tree rather than hand-edited to agree.

**No CHANGELOG entry.** `tools/list` is byte-identical to `main`: nothing a caller can reach changed. The
renderer output is unchanged too — `conflictTree` was `false` at all five shipped call sites, so
`detail = fields is {Count:>0} || conflictTree` collapses to its first term, and `AppendRecordBlock` /
`Wire.RenderBatch` losing `svc`/`fields` is behaviour-preserving. Both round-2 reviewers checked this
independently and agreed no entry is owed.

## ADR 0005, true of this tree

> "Guards on a deleted tool die with the tool; behaviour that survives gets a fresh test."

Every site the deletion broke is now either deleted with the file it lived in, carried forward as a new xUnit
test against the surface that renders the same fact today, or named as already covered by an existing test.
`dotnet build` at 0 errors is the compiler confirming no reference to a deleted member survives; 919/919 is
every fresh and pre-existing test passing. ADR 0005's second sentence — its own "not always in the same
change" measurement, which chartered #486 as the completion of the rule for the 1.x cut — is discharged too.

**With the addendum this PR lands**, dated and appended without rewriting anything above it: the rule is true
as stated **for every site the compiler named**, which is what this PR discharges. The 93 renderer-free arms
of the two whole-file deletions are a cost the compiler's list does not measure — three re-homed here, 90
chartered to #494.

## Settled decisions

`ruled` = Aaron's own words, cited to a record the tagging session read. `inferred` = a session concluded it
and Aaron never spoke; a reviewer may re-litigate an `inferred` entry.

- **`ruled`** — the render halves are deleted now, with the compiler as the checklist. Aaron at PR #484's
  gate: *"I agree with the rest of your calls."*
- **`ruled`** — two sequenced PRs, fixtures first, then the deletion, landing before #440. Aaron, in #486's
  body: *"agree, its a follow up."*
- **`ruled`** — the renderer-free arms: charter the bulk to an issue, port the #207 cluster now. Aaron on the
  round-1 escalation, 2026-09-03: *"agreed."*
- **`ruled`** — a self-review finding is fix-or-drop, never filed. Aaron, CLAUDE.md §2.
- **`inferred`** — the phase order; `ReadPluginFile`'s engine half left in place and carried here; the #394
  baseline re-derived by measurement, with only a change in the allocation PROPERTIES a stop; the anchoring
  rule (anchor to record/response identity wherever a fixture has more than one carrier of the same shape, and
  a sabotage must corrupt INSIDE the matched span).
- **`inferred`** — `Wire.Close`, `SumConditioned` and `AppendStandingLimits` are cascade of the eight;
  `AppendRecordBlock`'s and `RenderBatch`'s `svc`/`fields` go with `conflictTree` (beyond the charter's literal
  list); the (c) diagnostic sites in the two deleted files are not fixed; the three `EpochGuardProbe` arity
  sites are repaired in place; the EpochGuardProbe four-site disagreement resolves to (b).
- **`inferred`** — fact 24 gets no dedicated arm; structural (json-shape/numeric) assertions are not
  individually sabotage-tested; `EpochGuardProbe.cs` is not deleted and owes no `Converted-from:` marker;
  `CheckMergeProbe`'s two (b) sites are fixed in phase 3; the four `.pex` helpers are re-derived as
  `CheckMergeProbe`'s own members rather than moved to a third shared file; S11/S12's cap values are searched
  across a band rather than pinned; the four `Wire.RenderPluginFile` sites die (a); B1/B2/B3/B6/B8/B9/B10 are
  already covered; `DialogueWorld`'s lock tests call `Svc.Stats()` once before taking a hold (#353); no
  dedicated merged-surface-wide cap/floor arm beyond the per-family instances.

Round-2 reviewer A examined four of the `inferred` entries and declined to re-litigate any, with reasons.

## Open for Aaron's gate

1. **Three shipped public members with zero shipped callers — delete them, or give them a 2.0 caller.**
   `LoadOrderService.ReadPluginFile` (its `WhyNotActive` / off-order-cause composition; its engine half's
   missing-master split does have a live home in `Mo2LoadOrder.SplitUnsatisfiedMasters`), `ResolveTree` (its
   only shipped caller was `AppendConflictTree`, deleted here), and the public `ResolveRead(FormKey, …)`
   overload (measurably unexercised). All three are the same question; this PR deliberately decided none of
   them. Until they are decided, `ReadPluginFile`'s DTO-level facts sit in `BulkPrimitivesWave3Probe.cs`
   surviving only because those lines never called a deleted member — not because anyone confirmed the
   mechanism is reachable from a live tool. That narrowing is stated in the probe's own comments rather than
   hidden: the `mod=` address lane keeps no render arm at all.
2. **#353 — the index build silently excludes an unreadable plugin,** changing which plugin the index calls
   the winner rather than surfacing a read failure. Met while writing this branch's lock tests; already filed
   and routed W5/E4, so cited, not re-filed. The tests work around it with an explicit `Svc.Stats()` before
   the hold, and `DialogueWorld`'s doc now says so.
3. **The anchoring class — #492 is the ruled by-construction answer.** It recurred four times on this branch
   (twice in `ab4c3e5`, once in phase 3, once in round 0) and once more in round 2 as A-MED-1. Structure-keyed
   tests are the answer already ruled; the #207 port here is written that way. Under CLAUDE.md §3's third
   cornerstone a detector is explicitly the wrong shape, so none is proposed.
4. **#494 — the 90 renderer-free arms** deleted with the two probe files, chartered with the derived list and
   sequenced with #492, before the conversion PRs.
   *One correction owed to that issue:* the arm line numbers in its body are **+2** relative to the real file
   for everything below L147 (`OWNER-RANK`'s `Check(` is at L530, listed L532). The red-line index the
   derivation consumed is correct and re-measured, so the classification and the counts are unaffected — only
   the citations do not resolve. Not edited here: this session's charter was to open the PR, and editing a
   filed issue is a separate outward-facing act.
5. **The class recurrence named above** — populations claimed as derived, five instances across rounds 1 and
   2. The rounds stopped on it; the seam is stated and no mechanism is proposed.

## Gate folds

Aaron's gate review posted **nine** findings. Eight were triaged as confirmed folds and executed by a fresh
fold session; the ninth is his to rule and is untouched. Every fold that changes an assertion was RED-checked
from a **committed** state, with the reviewer's own sabotage where he gave one, then restored and the restore
verified (grep for a string the fold introduced, or an md5 of the sabotaged file).

| # | comment | finding | fold | commit | RED-check |
|---|---|---|---|---|---|
| 1 | 3927993641 | `EntryLines` has no positive control — Fact9's emptiness assertion can go vacuous | Fact9 asserts the ample-cap render yields six entry lines before asserting the 400-cap render yields none | `902f00f` | reviewer's own sabotage (drop "by any" from the needle) → **RED on the control**, `Assert.Equal() Expected 6 Actual 0` at the new line, not on the `Assert.Empty` |
| 2 | 3927993875 | `ReadSentences.SweepBoundaryLabel` orphaned by this branch | const deleted; the neighbouring `<see cref>` rewritten so it names no deleted member | `9cc6555` | no assertion changed — the scan below is the evidence, and it was canary-checked against the pre-fold file |
| 3+4 | 3927994116, 3927994328 | three `ScriptsFixtures` helpers and `CheckErrorsFixtures.Plugin`/`Ref` land with zero callers | one derived population, five members deleted | `7f66d06` | the compiler is the check — build at 0 errors, 919/919 unchanged |
| 5 | 3927994545 | `w2Fid` and `masterName` are now unused locals in `DiffArm` | both deleted, plus the `w2Fk` cascade | `e1b183b` | build 0 errors, 0 new warnings (140 → 140), ci-all 128/128 |
| 6 | 3927994775 | `IClassFixture` builds `CheckErrorsWorld` twice per run | `[CollectionDefinition("check-errors")]` + `ICollectionFixture`, `[Collection]` on both classes; the doc names the collection | `7a1f9f5` | measured with a temporary construction counter, never committed: **2 constructions before, 1 after** |
| 7 | 3927994965 | the doc names one terminator; `SeedBlock`'s predicate has four | `\|\| l.Trim().Length == 0` dropped — the clause, not the doc | `ba292ec` | round 2's three seed-verdict cells (`DialogueViewParityOk`, `DialogueBranchParityOk`, `DialogueQuestParityOk` deleted in turn) plus reviewer A's hand-proven canary (`ENAM`→`ZZQM`): **4 of 4 RED**, `ReadSentences.cs` restored byte-identically (md5 `fe472b5b5125dc82cb4d555c0a753f42` before and after) |
| 8 | 3927995113 | `CheckMergeProbe.AutoScalar` is the unhardened twin of `PexWriter.Auto` | the `int? initInt` parameter and the initializer branch are removed — the shape goes rather than being described; `Auto` always writes `VariableType.Null` | `0bb9d33` | compiler as the checklist: build 0 errors, ci-all 128/128 |
| 9 | 3927995298 | `ReadPluginFile` loses every render arm and its last product caller | **open — Aaron's call.** This is open item 1 above, and his own words on it are "Not a blocker on this PR". `ReadPluginFile`, `PluginFileOutcome`, `ResolveTree`, the `ResolveRead` overload and the probe comment at that line are untouched by these folds | — | — |

### The derived populations

Three folds derived their subject set by script rather than fixing the instance named.

- **Fold 1 — helpers with no positive control.** Every test-project helper that returns a value (**77
  declarations** across the 68 files of `src/housecarl-mcp-tests`) whose call sites are consumed ONLY by
  `Assert.Empty` / `Assert.DoesNotContain` / `Assert.False` / `Assert.Null`: **one**, `EntryLines` itself.
  The population is a singleton; there is no twin to dispose of. (Helpers with no caller at all are a
  different population — folds 3+4.)

- **Fold 2 — `ReadSentences` members with no reader.** Aaron's own scan re-run over `src/ docs/ standards/
  plugin/ .github/ scripts/`: every `internal const` / `static` member of `ReadSentences` whose only CODE
  occurrence repo-wide is its own definition, doc-comment mentions excluded exactly as his finding excluded
  them ("that occurrence is its own definition (plus a `<see cref>` from a neighbouring doc)"). **1 before
  the fold** (`SweepBoundaryLabel`, code-occurrences 1, all-occurrences 2), **0 after**, over 76 → 75
  members. The scan was run against the pre-fold file first as its canary, so an all-clear is a working
  detector rather than a broken one.

- **Folds 3+4 — zero-caller fixture members.** Every member declared in the two fixture classes, with its
  caller count outside its own declaration, **class-scoped and overload-aware** and iterated to transitive
  closure. Scope matters because `Sel`, `Result`, `Text`, `Json`, `Head` and `Plugin` are declared in BOTH
  classes and `Plugin`/`Text`/`Json` also in `RecordsTestBase` and `RemedyHarvest` — a name-keyed count is
  the trap round 2 recorded. **22 members; 5 dead:**

  | member | decl | callers | |
  |---|---|---|---|
  | `CheckErrorsFixtures.Epoch` | :17 | 6 | live |
  | `CheckErrorsFixtures.Sel` | :20 | 16 | live |
  | `CheckErrorsFixtures.Ref` | :28 | 0 | **DEAD** (read only by `Plugin`) |
  | `CheckErrorsFixtures.Plugin` | :33 | 0 | **DEAD** |
  | `CheckErrorsFixtures.Result` | :38 | 4 | live |
  | `CheckErrorsFixtures.Text` (params) | :60 | 33 | live |
  | `CheckErrorsFixtures.Text` (histogramLimit) | :64 | 3 | live — `Fact16_18` ×2, `Fact19` |
  | `CheckErrorsFixtures.Json` (params) | :68 | 10 | live |
  | `CheckErrorsFixtures.Json` (histogramLimit) | :71 | 1 | live — `Fact16_18` |
  | `CheckErrorsFixtures.ErrorsFamily` | :75 | 14 | live |
  | `CheckErrorsFixtures.PluginSection` | :85 | 2 | live |
  | `CheckErrorsFixtures.EntryLines` | :96 | 2 | live |
  | `CheckErrorsFixtures.Head` | :102 | 1 | live — read by `PluginSection` |
  | `ScriptsFixtures.Epoch` | :16 | 1 | live |
  | `ScriptsFixtures.Sel` | :19 | 18 | live |
  | `ScriptsFixtures.Result` | :27 | 4 | live |
  | `ScriptsFixtures.Text` (params) | :52 | 4 | live |
  | `ScriptsFixtures.Text` (histogramLimit) | :56 | 0 | **DEAD** |
  | `ScriptsFixtures.Json` (params) | :60 | 1 | live |
  | `ScriptsFixtures.Json` (histogramLimit) | :63 | 0 | **DEAD** |
  | `ScriptsFixtures.ScriptsFamily` | :67 | 8 | live |
  | `ScriptsFixtures.Head` | :71 | 0 | **DEAD** |

  Exactly the five the two findings named, and no more; the errors family's histogram overloads are
  confirmed live, which is what the scripts finding predicted. `CheckErrorsFixtures`' `using
  Mutagen.Bethesda.Plugins` goes with `Ref`, its last reader. The script's first pass called those two
  overloads dead because it read `histogramLimit: 1` as a string argument — fixed before it was believed.

### Gates after the folds

| measure | before the folds (`f72c2f0`) | after (`0bb9d33`) |
|---|---|---|
| `dotnet build houseCARL.sln -c Release` | 0 errors | 0 errors |
| `dotnet test --filter "tier!=bridge"` | 919 | **919** |
| `ci-all` | 128/128, ALL PASS | **128/128, ALL PASS** |
| `HarnessResidueTests` | 7/7 | **7/7** |
| residue `probeFiles` | 148 | **148** |
| generator warnings (Release, `--no-incremental`) | 140 | **140** |

**`tools/list` re-measured as a comparison, not a constant.** `origin/main` (`f307f7f`) was built in a
scratch worktree, both servers driven over stdio with `initialize` / `notifications/initialized` /
`tools/list`, and the two sorted, pretty-printed dumps compared with `cmp` — **identical**, 35 tools,
223,452 bytes, md5 `791aa1e546a0e611ba73e887cac87459` on both. (The byte count differs from the figure
recorded when the PR opened because the dumper differs; that is exactly why the claim is the comparison.)

### Noticed and NOT folded

Two other fixtures in this project are shared by two test classes each through `IClassFixture`, each under a
doc sentence reading "built once for the class" — `CheckWorldFixture` (`CheckExcludeGrammarTests`,
`SeqStandingLimitRemedyTests`) and `CheckMasterRemedyFixture` (`CheckMasterRemedyTests`,
`CheckMasterRemedyJsonTests`). Both predate this branch on `main`, so they are outside the branch under
review and outside these folds. Recorded here rather than fixed, for Aaron.

### Rebase

Rebased onto `e8847bf` (post-#495) 2026-09-03: one comment-level conflict in `ReadTools.cs` resolved for the
branch's deletion; gates re-run: build 0 errors, 963 tests (base `f307f7f` 859 + this branch's 60 + #495's
44 — the pre-rebase 919 plus what #495 brought), `ci-all` 128/128 (`origin/main` measures 130; #495 added no
probe row, so the delta is still this branch's two retired rows), `HarnessResidueTests` 7/7, residue
`probeFiles` 148; wire byte-identical to main — both servers driven over stdio with
`scripts/tool-names/dump-tools-list.py`, 35 tools, md5 `60ee99828c4e8090c60bf79d3c7cab62` on both dumps.


src/housecarl-mcp-tests/CheckErrorsFixtures.cs
```diff
@@ -0,0 +1,107 @@
...
+        var lines = response.Split('\n');
+        int start = Array.FindIndex(lines, l => l.StartsWith("[ERROR] " + plugin, StringComparison.Ordinal));
+        Assert.True(start >= 0, $"no [ERROR] section for {plugin}: " + Head(response));
+        int end = start + 1;
+        while (end < lines.Length && lines[end].StartsWith("  ", StringComparison.Ordinal)) end++;
+        return string.Join("\n", lines[start..end]);
+    }
+
+    /// <summary>The dangling entry lines a text response actually emitted.</summary>
+    internal static string[] EntryLines(string response) =>
```
Avick3110, line 85: **[test-coverage] `EntryLines` has no positive control — `Fact9`'s emptiness assertion can go vacuous.**

This helper has exactly one consumer in the repo, and it is an `Assert.Empty` (`Fact9_MaxCharsOnlyCut_...`, line 209). Nothing anywhere proves the needle `[target not defined by any active plugin]` ever matches a real entry line.

Reword the dangling-entry render — drop the `any`, or rename the bracket to `[target not defined in the active order]` — and `EntryLines` returns an empty array for *every* response. `Assert.Empty(EntryLines(text))`, the assertion that says "no entries were emitted at max_chars=400", then passes for every possible render, including one that emits all six. The accounting assertion above it is partial cover only; the emptiness arm itself is the one that goes silent.

This is the shape `HarnessResidueTests` guards against with its own vacuity canary (`markers.Length > 0`). The cheapest fix is the same one: assert the helper finds entries in the ample-cap control before asserting it finds none in the cut one.

src/housecarl-mcp/CheckAccounting.cs
```diff
@@ -291,17 +292,14 @@ internal IReadOnlyList<SweepCount> MissingBySource
...
-    /// <summary>The accounting LINE's own room, without the boundary's. Separate because a merged response has one
-    /// boundary block and one accounting PER FAMILY: summing whole TextReserves would reserve the boundary once per
-    /// family, and room held for a sentence written once is a subtraction from the answer.</summary>
+    /// <c>max_chars=1200</c>, the response came back 315 chars under its own cap with both axes at zero rows.</para>
+    ///
+    /// <para>The <c>TextReserve</c> that summed this with the boundary's room went with the 1.x single-family
+    /// renderers (#486): a merged response has one boundary block and one accounting PER FAMILY, so the two are
+    /// only ever reserved apart, and summing them per family reserved the boundary once per family — room held for
+    /// a sentence written once, which is a subtraction from the answer.</para></summary>
     internal int TextAccountingReserve => _textReserve ??= CanStateAccounting ? Compose(Worst(escaped: false)).Length + TextWrap : 0;
```
Avick3110, line 301: **[dead-code] `ReadSentences.SweepBoundaryLabel` is orphaned by this hunk.**

`SweepBoundaryLabel` (`ReadSentences.cs:476`) had exactly two consumers: `Wire.Close` and the `CheckAccounting.TextBoundaryReserve` deleted right here. This PR deletes both. A scan of every `internal const`/`static` in `ReadSentences` over the branch turns up exactly one member with a single occurrence repo-wide — this one, and that occurrence is its own definition (plus a `<see cref>` from a neighbouring doc).

The PR's method is "the compiler as the checklist", and a `const string` is precisely what the compiler cannot name. What survives is a sentence with no writer, sitting beside the `SweepBoundaryLabelFor` the merged reserve loop actually uses — an invitation for a later reader to reintroduce the single-family label the merge deliberately dropped.

src/housecarl-mcp-tests/ScriptsFixtures.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    internal static string Json(ScriptCheckResult r, int maxChars, int histogramLimit, params string[] tokens) =>
+        JsonWire.RenderCheck(new CheckSweep(Sel(tokens), Scripts: r), maxChars, histogramLimit);
+
+    /// <summary>The scripts family's own object in a merged json document.</summary>
+    internal static JsonElement ScriptsFamily(string json) =>
+        JsonDocument.Parse(json).RootElement.GetProperty("families")
+                    .GetProperty(SweepFamilySelection.Token(SweepFamily.Scripts));
+
+    internal static string Head(string response)
```
Avick3110: **[dead-code] Three helpers in this new file land with zero callers.**

`ScriptsFamilyTests` is this file's only consumer, and its calls all bind to the two-argument overloads:

- `Head` (line 71) — no caller at all.
- `Text(ScriptCheckResult, int, int histogramLimit, params string[])` (line 56) — the four `Text(` sites are `Text(r, 0)`, `Text(excludedFat, cap)` ×2, `Text(excludedFat, 0)`.
- `Json(ScriptCheckResult, int, int histogramLimit, params string[])` (line 63) — the one `Json(` site is `Json(scanned, cap)`.

(`CheckErrorsFixtures`' histogram overloads *are* used — `Fact16_18`, `Fact19` — so this is the scripts twin only.)

Same population as commit 8c382d9 ("the five fixture helpers nothing calls are removed"), and the same population the whole PR exists to remove from the product. Worth the sweep before the PR opens rather than leaving the next reader to work out whether these are dead or merely not-used-yet.

src/housecarl-mcp-tests/CheckErrorsFixtures.cs
```diff
@@ -0,0 +1,107 @@
...
+        return selection!;
+    }
+
+    /// <summary>One dangling ref from <paramref name="plugin"/>, numbered so a rendered entry is identifiable.</summary>
+    internal static DanglingRef Ref(int i, string plugin) =>
+        new(FormKey.Factory($"{0x000800 + i:X6}:{plugin}"), "Npc", $"HcEd{i}",
+            FormKey.Factory("0E0E0E:Skyrim.esm"));
+
+    /// <summary>A plugin report carrying <paramref name="n"/> dangling refs and nothing else.</summary>
+    internal static PluginErrors Plugin(string name, int n) =>
```
Avick3110: **[dead-code] `Plugin` and `Ref` are dead on arrival.**

`Plugin(string, int)` has no caller in the repo; `Ref(int, string)` (line 28) is called only from `Plugin`. The DTO-lane arms build their `PluginErrors` inline instead — `Fact7` (the unread row) and `Fact22` (the narrow/wide floor pair).

Worth deleting with the ScriptsFixtures set: `RecordsTestBase.Plugin(string)` already exists in this namespace meaning something else entirely (a JSON pole), so a later writer reaching for `Plugin(` in an errors-family test gets a name collision to resolve over a helper nothing uses.

src/housecarl-generator/BulkPrimitivesWave3Probe.cs
```diff
@@ -566,53 +560,14 @@ static void DiffArm(string dir)
         string w2Fid = $"{w2Fk.ID:X6}:{w2Fk.ModKey.FileName}";
         string masterName = mKey.FileName.String, replName = rKey.FileName.String;
```
Avick3110: **[dead-code] `w2Fid` and `masterName` are now unused locals.**

Both were read only by the `svc.DiffRecord(...)` arms this PR deletes (`diff same plugin both sides`, `refusal: a plugin doesn't define the record`, `diff master vs repl`). Nothing in the remainder of `DiffArm` touches either — `replName`, declared on the same line as `masterName`, is still live at the `readTypo` and `mergeUnticked` arms.

Neither is initialised from a constant expression, so no CS0219 fires and the compiler-as-checklist pass cannot see them. The half-dead declaration on line 561 is the one that reads wrong at a glance.

src/housecarl-mcp-tests/CheckErrorsWorld.cs
```diff
@@ -0,0 +1,132 @@
...
+
+    public void Dispose()
+    {
+        Svc.Dispose();
+        try { Directory.Delete(Root, true); } catch { /* temp cleanup best-effort */ }
+    }
+}
+
+/// <summary>The world, built once for the class. Every arm over it is read-only.</summary>
+public sealed class CheckErrorsWorldFixture : IDisposable
```
Avick3110, line 131: **[efficiency] `IClassFixture` here builds the world twice per run.**

`CheckErrorsWorldFixture` is consumed through `IClassFixture<>` by both `CheckErrorsFamilyTests` and `CheckErrorsWorldTests`. xUnit instantiates a *class* fixture once per test class, so the synthetic MO2 instance — four Mutagen plugin writes plus an index build — is constructed and torn down twice.

Every other world in this project shares through `ICollectionFixture` + `[CollectionDefinition]` for exactly this reason (`ScriptsWorld` → `"scripts"`, `RecordsWorld` → `"records"`, and this PR's own `EpochWorld` → `"epoch"` and `DialogueWorld` → `"dialogue"`). `CheckErrorsWorld`'s own doc already says "built once for the class" while two classes claim it. A collection definition costs three lines and makes the doc true.

src/housecarl-mcp-tests/DialogueFamilyTests.cs
```diff
@@ -0,0 +1,232 @@
...
+    /// it, and an assertion that a seed states its own verdict was satisfied by a sibling's. Measured — deleting
+    /// the DLVW verdict outright left the arm green. The terminator is the next head, which is what
+    /// "this seed's own report" was always supposed to mean.</para>
+    ///
+    /// <para>The START is anchored to a HEAD line for the same reason, and that is not belt-and-braces: with the
+    /// terminator fixed and the start still "the first line containing the marker", the quest block resolved to the
+    /// DLVW seed's scope sentence, which names "quest (QUST)" as the record to validate for a graph surface. The
+    /// old span ran past it to the real quest head and passed on that. A kind label is a head's vocabulary, so the
+    /// head is what the search reads.</para></summary>
+    static string SeedBlock(string response, string marker)
```
Avick3110, line 215: **[correctness, low] The doc names one terminator; the predicate has four.**

The doc block above states the round-2 fix as *"The terminator is the next head, which is what 'this seed's own report' was always supposed to mean"* — but the `FindIndex` predicate still ends in `|| l.Trim().Length == 0`, the blank-line terminator that paragraph is describing as the bug.

The extra terminator only shortens a block, so it cannot turn a real failure green. But if a seed head ever grows a blank line before its verdict (a spacing change in `DialogueSweepRender`), `FactV1`'s `Assert.Contains("CK-parity: OK", byViewLine)` goes red for a reason the doc says was designed out, and whoever debugs it is reading a rule the code does not implement. Either drop the clause or amend the paragraph to say the blank line is kept as a belt-and-braces floor.

src/housecarl-generator/CheckMergeProbe.cs
```diff
@@ -2011,4 +2016,80 @@ static void CollectDuplicateKeys(JsonElement e, string where, List<string> into)
...
+    /// declared type, so <c>AutoScalar("MyFlag", "Bool", 1)</c> produces a lie.
+    /// The test project's copy does NOT merely record this — <c>PexWriter.Auto</c> REFUSES a
+    /// non-<c>Int</c> declared type carrying an initializer, with an <c>ArgumentException</c> naming the property,
+    /// the type and the value, pinned in both directions by
+    /// <c>ScriptsWorldTests.TheWriterRefusesABakedInitializerOnANonIntScalar</c> and its round-trip twin;
+    /// <c>docs/architecture/test-project-fixtures.md</c> spends a paragraph on why refusing beats documenting.
+    /// This copy is not hardened to match, deliberately: the refusal is a conditional, the generator has no unit
+    /// harness to arm both of its branches from, and an unarmed conditional is what the review-rounds standard
+    /// says not to add. Both call sites here pass <c>"Int"</c>, and this file retires with its own conversion.</para></summary>
+    internal static Decl AutoScalar(string name, string typeName, int? initInt) => Auto(name, typeName, initInt);
```
Avick3110: **[correctness, latent] This is the unhardened twin of `PexWriter.Auto`.**

The doc is candid that the two copies differ — this one records the non-`Int`-with-initializer lie in prose where the test project's copy throws `ArgumentException`. The stated reason (an unarmed conditional is what the review-rounds standard says not to add) is fair, but the result is two copies of one fact with divergent behaviour, which is the shape §3's third cornerstone asks to make unwritable rather than document.

Concretely: a later arm writing `AutoScalar("HcCmFlag", "Bool", 1)` gets a backing variable stamped `TypeName = "Bool"` with `VariableType.Integer`; the decompiler renders `Bool Property HcCmFlag = 1 Auto`, and the arm asserts product behaviour against a `.pex` shape the game never produces — green while claiming to model a defaulted Bool. Latent today: both call sites pass `"Int"` with `initInt: null`, so the initializer branch never fires at all.

If arming both branches is out of scope, dropping the `int? initInt` parameter here (no caller passes a value) removes the shape instead of describing it.

src/housecarl-generator/BulkPrimitivesWave3Probe.cs
```diff
@@ -690,21 +645,13 @@ dPathDisabled.Error is null && !dPathDisabled.A!.InOrder
...
-              renderLive.Contains("OUT-OF-LOAD-ORDER") && renderLive.Contains("not resolved through the load order")
-              && !renderLive.Contains("the game does not load this file"));
-        Check("#271 render: a live file addressed by path says so positively — #269's symptom, inverted",
-              renderLive.Contains("the game loads this file"));
-        var renderUnticked = Wire.RenderPluginFile(rpfUntickedByName, 8000);
-        Check("#271 render: the mod-vs-plugin pairing is disambiguated in words, and carries the cause",
-              renderUnticked.Contains("mod 'DiffUnticked' (enabled)")
-              && renderUnticked.Contains("the game does NOT load this file")
-              && renderUnticked.Contains("UNTICKED in plugins.txt"));
+        // The RENDERED header — the "[mod 'X' (enabled); NOT active]" pairing named a folder and a file in one
```
Avick3110, line 647: **[test-coverage] `ReadPluginFile` loses every render arm *and* its last product caller here.**

With `Wire.RenderPluginFile` / `JsonWire.RenderPluginFile` gone, `LoadOrderService.ReadPluginFile` has zero callers under `src/housecarl-mcp` — this probe is the only thing left driving it — and the facts the banner carried have no test anywhere: #271 items 3/4, B12/B13's filename and `mod=` counting lanes, and the json `why_not_active` arm.

The narrowing is honestly stated (this comment, and the note two hunks down: *"The mod= address lane keeps no arm of its own on this branch"*), and `RecordsOffOrderPathTests.FactB4` does carry the counting forward — but for the PATH address form only. The filename and `mod=` forms are unarmed, and the `mod=` lane is precisely where the #271 duplication survived a review round once already, because the equality test that was supposed to catch it failed silently there.

Not a blocker on this PR — it is in the body as an open item for Aaron. Flagging it so the choice (delete `ReadPluginFile` and its `PluginFileOutcome` chain too, or give it a 2.0 caller) is recorded against a line rather than only in a PR description that stops being read after merge.

Avick3110:
Finding 9 (`ReadPluginFile`, review comment 3927995298) — Aaron: "agreed on finding 9". Filed as #497 (the zero-caller `LoadOrderService` members after this deletion, population derived rather than the three named; carries the filename/`mod=` capability question), sequenced in row 7 after #492.

PR #500, opened 2026-09-03
test(records): the wrong-lever grid drives every lane housecarl_records renders

### LINKED ISSUE #498: [GAP] RecordsRemedyGrammarTests: RemedyHarvest's tree-form subject set is hand-enumerated via RecordsWorld — derive it from the record types the surface renders a declarers block for
Filed 2026-09-03 by the advisor lane on Aaron's "work autonomously from here", from PR #495's open item 1 (round-3 reviewer A, MEDIUM: "the wrong-lever guard's population is hand-enumerated"). Third-cornerstone class — the same shape as #483 (guard populations derived from the surface they claim).

## The gap

`RecordsRemedyGrammarTests`' `RemedyHarvest` reaches the tree form only through `RecordsWorld`, a frozen shared fixture that has **no child-bearing record type**. So every cut notice the #485 interim tier added (`[child declarers cut at max_chars=N …]`, and the `nodes` notice a cut row now carries beside it after PR #495's gate fold) never entered the wrong-lever grammar grid, and the grid stayed green over the gap. Round 2 covered the branch with one hand-written replacement arm (`TheChildDeclarersCutNoticesNameNoWrongLever`) against the branch's own fixture; round 3's `AppendCutNotice` fold made the guard's single harvested hit sufficient at the one level the branch could reach (one literal, not three). Neither derives the population.

## What this issue asks for

Derive `RemedyHarvest`'s tree-form subject set from **the record types the records surface can render a declarers block for** — `WriteEngine.ChildBearingProperties` / `OwnedChildContent.Fields`, the same by-construction set `ReadSentences.FieldList` already consumes — rather than from whatever `RecordsWorld` happens to contain. Either the harvest drives a fixture that carries at least one record per child-bearing shape (COLLECTION and SINGULAR), or the harvest refuses to run when its subject set does not cover the surface's shapes (a derived guard on its own completeness, the #483 pattern). `RecordsWorld` stays frozen; the fixture that carries the shapes is `OwnedChildWorld` (PR #495) or a derived extension of it.

Row 7, beside #483 and #492. Milestone houseCARL 2.0.

Avick3110:
`OwnedChildRemedyGrammarTests` drives the wrong-lever grid over the tree form on every child-bearing owner type, with its subjects found in the fixture rather than named, and a completeness arm that fails when a shape the surface renders has no subject — the gap #498 was filed about, where the grid ran green because `RecordsWorld` holds no child-bearing record at all. The grid's lane list comes from `RemedyHarvest.Lanes`, which is now the literal keywords `Wire.CrossQueryFormat` accepts (`text`, `json`, `dense`) plus the artifact lane; it was derived from `Wire.QueryFormat`, but the parser compares against hard-coded keywords and never reads that enum, so the derivation only agreed by coincidence. The arm that compared the lane list against the enum it was derived from is gone because it could not fail, and the prose guard, sentence catalogue, reachability countdown and their baselines are gone because the project no longer builds tests about tests.

Closes #498

## Gates

- `dotnet build housecarl.sln -c Release` — 0 errors.
- `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` — 995 passed, 0 failed.


src/housecarl-mcp-tests/read-sentence-reachability-unreached.json
```diff
@@ -0,0 +1,72 @@
...
+  "_why_most_of_them":"This class holds one fixture, the owned-child world, because the other worlds (CheckErrorsWorld, DialogueWorld, ScriptsWorld, EpochWorld) live in their own xUnit collections over process-wide statics. Reaching their render sites means one reachability class per collection - existing fixtures, no new ones - which is the following engagement's work. Four entries are a different thing entirely: sentences with no unbroken literal run long enough to be found by identity at all, which is a finding about the sentence. CarriedBy joined them when Facts.LongestRun was corrected to measure a contiguous run rather than a count of non-space characters - its arm had been passing on the count.",
```
Avick3110: **Eight of these entries are not fixture-reachability debt — they are unreachable-by-identity, mislabelled.**

`_why_most_of_them` says four entries are "sentences with no unbroken literal run long enough to be found by identity at all". Running `Facts.LongestRun` over every value-shaped `ReadSentences` member gives **twelve** below `IdentifiableRun = 8`:

| member | run | labelled? |
|---|---|---|
| `SweepClose` | 1 | yes |
| `SweepFamilyNotRun` | 1 | yes |
| `SweepFamilySectionHead` | 1 | yes |
| `CarriedBy` | 7 | yes |
| `CouldNotRead` | 5 | **no** |
| `SweepRemedyLimit` | 6 | **no** |
| `DialogueScopeAllSeeds` | 7 | **no** |
| `DialogueScopeSomeSeeds` | 7 | **no** |
| `SweepNoSectionRule` | 7 | **no** |
| `SweepOmittedByBudget` | 7 | **no** |
| `SweepRosterLead` | 7 | **no** |
| `SweepRosterCut` | 7 | **no** |

The eight unlabelled ones carry "the check-errors world … needs its own reachability class" or "needs a provider body that could not be READ". That reason is false: no fixture can drive them, because `TheSentenceReachesAResponse` refuses them at the `Identifiable` gate before it ever consults the harvest.

Concrete consequence — this is the PR's own "Open for Aaron" item: the follow-up engagement adds `CheckErrorsReachabilityTests` / `DialogueReachabilityTests`, deletes those keys per the countdown's instruction, and the theory cases go red on the identifiability refusal rather than passing. The countdown therefore overstates how much of the 65-entry remainder is reachable work by eight.

Nothing is red today (they are all recorded, and `NoRecordedPairIsActuallyReachedByThisClass` skips non-identifiable sentences), so this is a correction to the reasons and to `_why_most_of_them`, not to the numbers.

src/housecarl-mcp-tests/RemedyHarvest.cs
```diff
@@ -32,7 +32,51 @@ public static readonly (string Pattern, string Claim)[] WrongLevers =
...
+    /// <summary>The lane that is not a <c>format=</c> value at all — a <c>to_file</c> spill, whose rows are
+    /// read back off the artifact rather than out of the response.</summary>
+    public const string ArtifactLane = "artifact";
+
+    /// <summary>
+    /// Every lane housecarl_records renders, DERIVED from the product's own format vocabulary plus the
+    /// artifact lane. A format the product gains is in this list the day it lands, and every consumer of the
+    /// list gets its rows without being edited.
+    /// </summary>
+    public static readonly string[] Lanes =
```
Avick3110: **The lane vocabulary is derived off the enum's identifiers, not off the `format=` keywords the product actually parses.**

`Wire.CrossQueryFormat` decides `format=` by comparing against hard-coded literals `"text"` / `"json"` / `"dense"` — nothing there reads `Enum.GetNames`. So `Lanes` and the product's real vocabulary agree by naming coincidence, and both completeness arms are blind to a divergence.

Failure scenario. Rename `Wire.QueryFormat.Dense` → `Columnar`, wire keyword unchanged:

1. `Lanes` becomes `text/json/columnar/artifact`.
2. Every columnar call in `OwnedChildRemedyGrammarTests.Harvest` hits `if (ferr is not null) return ferr;` (`RecordsTools.cs:198`) and gets back the plain-text `error: format='columnar' is not recognized — use 'text' (the default), 'json', or 'dense'.`
3. That string matches no `WrongLevers` pattern and no `RemedyLine` keyword → `harvested == 0`, and it starts with `error:` → `refusals == responses`, so `EveryLaneEitherRendersSentencesThisGridReads_OrRefusesTheTreeFormByName` passes on its second disjunct.
4. `TheGridDrivesEveryLaneTheSurfaceRenders` compares `Enum.GetNames<Wire.QueryFormat>()` against `Lanes` — the derivation against a verbatim copy of itself — so it passes too.

Net: the dense lane's six wrong-lever rows run over refusals forever, green, which is the vacuity the derived lane list was meant to expose.

Two ways to close it, either enough: derive `Lanes` from the keywords `CrossQueryFormat` accepts, or add an arm asserting that every non-artifact lane name round-trips through `Wire.CrossQueryFormat` with `error == null`.

src/housecarl-mcp-tests/RemedyHarvest.cs
```diff
@@ -32,7 +32,51 @@ public static readonly (string Pattern, string Claim)[] WrongLevers =
...
+        Enum.GetNames<Wire.QueryFormat>().Select(n => n.ToLowerInvariant()).Append(ArtifactLane).ToArray();
+
+    /// <summary>
+    /// HOW ONE LANE IS HARVESTED — the one home for the discriminant. The artifact lane's rows are read off
+    /// the file the call spilled. Every other lane is decided by the RESPONSE rather than by the lane's name:
+    /// a response that parses as a JSON document is walked for its strings, and one that does not is read
+    /// line by line.
+    ///
+    /// <para>Asking the response is what keeps this from drifting. <c>dense</c> is named like a text lane and
```
Avick3110: **This paragraph describes code that was never in the tree.** The replaced discriminant was:

```csharp
var hits = lane switch
{
    "artifact" => HarvestArtifact(ArtifactPath),
    "text"     => resp.Split('\n').Where(l => RemedyLine.IsMatch(l)).ToList(),
    _          => HarvestAllStrings(resp),
};
```

It keyed on `"text"`, not on `"json"` — `dense` fell to the `_` arm and already took the document walk. So dense was never narrowed to the `RemedyLine` lines, and the folded HIGH finding in the PR body ("`dense` … moved from *every string in the document* to *only the lines already matching* `RemedyLine`") did not describe live behaviour.

The lane whose behaviour actually changes here is **text**: a text response that happens to parse as JSON is now walked as a document instead of line-filtered. (I checked the two that could have mattered — `JsonWire.RenderCrossQueryDense` writes a `Utf8JsonWriter` object, and dense refusals go through `Wire.Refuse(json: false, …)` because `bool json = fmt is Wire.QueryFormat.Json` excludes Dense — so no current lane's harvest changes at all.)

Asking the response rather than the name is still the right discriminant; it's the stated justification that needs correcting, here and in `RemedyHarvestTests`' summary ("a discriminant keyed on the name \"json\" **dropped** every dense sentence…", past tense).

src/housecarl-mcp-tests/TestProseGuardTests.cs
```diff
@@ -0,0 +1,647 @@
...
+    };
+
+    static string Classify(string expected, string actual, IReadOnlyList<string> literals,
+                           IReadOnlySet<string> wire, IReadOnlySet<string> fixtureValues, Regex? symbolRe)
+    {
+        if (CatalogueRe.IsMatch(expected)) return "ii-catalogue";
+        if (literals.Count == 0) return "iii-structural";
+        if (literals.All(wire.Contains)) return "iii-structural";
+        if (CollectionRe.IsMatch(expected)) return "iii-structural";
+        if (DocRe.IsMatch(actual) && !actual.TrimEnd().EndsWith(".GetString()", StringComparison.Ordinal))
```
Avick3110: **AXIS 3 matches `DocRe` against the whole flattened subject, so a substring assertion over composed prose can bucket as structural.**

`DocRe` includes `\.ToList\s*\(`, `\.ToArray\s*\(`, `\.Select\s*\(`, `\.Count\b`, `\.Length\b`. It is tested against the entire subject expression, so a subject that merely *mentions* one of those anywhere qualifies — it does not have to be the outermost call.

```csharp
Assert.Contains("child records — declared per plugin", string.Join("\n", lines.ToList()));
```

`actual` is `string.Join("\n", lines.ToList())`; `DocRe` matches `.ToList(`; the expression does not end with `.GetString()`; `Classify` returns `iii-structural` at this line. That is still `Assert.Contains(string, string)` — a substring check on a fragment of prose the product composed, i.e. exactly #492's class — and it never reaches the `i-prose` bucket, so the file's baseline does not move and the new site lands green.

The rationale for AXIS 3 is sound where the subject really is a collection (`Assert.Contains<T>(T, IEnumerable<T>)` is an exact element match, so it is whole-line by construction). The gap is that the syntactic proxy cannot tell that case from a joined string. Two options: match `DocRe` only against the *outermost* accessor rather than the whole expression, or add this shape to the class doc's "What it does not see" paragraph — the other limits (`static readonly`, helper parameters, `Assert.True(x.Contains(…))`, lambdas, `using static`) are all named there and this one is not.

PR #501, opened 2026-09-03
docs: rewrite the operating manual and standards from a blank page

Avick3110:
Docs-only. CLAUDE.md is rewritten as one page: what houseCARL is, four design principles (coverage generated from Mutagen, one grammar closed under composition, plain one-sentence errors, lazy reads with mtime freshness), how to work, where things live, five don'ts. AGENTS.md becomes a pointer to it. `standards/` now holds a one-page TESTING.md and a half-page NAMING.md; the doc-hygiene, review-rounds, and old naming standards are removed. ADR 0001 and 0003 recorded process rather than architecture and are removed, along with the three code comments that pointed at them. The gap-report issue template becomes a plain feature request. CONTRIBUTING.md asks for a test rather than a probe or guard.

Why: Aaron's ruling of 2026-09-04. The governance layer had grown past the product it was meant to protect, and sessions imitate what they read at boot. Every operating doc is rewritten against one test: what a reasonable senior engineer would set up for a high-quality tool. Nothing is carried by copy. The old docs are kept in the private `dev/archive/2026-09-reset/` and in git history.

No behaviour change; no CHANGELOG line.


CLAUDE.md
```diff
@@ -1,142 +1,57 @@
...
+   Tests drive the built server. How to write one: [standards/TESTING.md](standards/TESTING.md).
+3. **Commit small.** One change per commit, plain imperative subject under 72 characters.
+4. **Open the PR.** A paragraph saying what changed and why, `Closes #N` for the issue, and a line under `## Unreleased` in `plugin/CHANGELOG.md` if a user would notice the change. Run one review pass (`/code-review`) and fix what is real. Aaron reviews and merges on his word with `gh pr merge <N> --rebase --delete-branch`; then remove the worktree.
+5. **Write plainly.** Comments are one line saying what. Commit messages, PR bodies, and issues use ordinary words: bug, test, review, fix. No project jargon, no lore.
+
+## Where things live
+
+| Path | What |
+|---|---|
+| `src/housecarl-mcp/` | The MCP server and tool surface |
```
Avick3110: **Where-things-live misroutes the load-order engine.**

The table gives `src/housecarl-mcp/` as "The MCP server and tool surface" and `src/housecarl-core/` as "Load-order, record, asset, and write engine". But `src/housecarl-mcp/LoadOrderService.cs` is 9,459 lines and holds most of the load-order logic; core has `LoadOrderResolver.cs` / `Mo2LoadOrder.cs` / `ReadEngine.cs` / `WriteEngine.cs` but not the service.

A new engineer orienting from this table to change load-order behaviour searches core, finds the logic thin, and misses the file it is actually in. The old manual's "don't append a new domain to a large existing file" rule — which existed because of that file — is removed in the same commit, so nothing else in the repo points at it.

standards/TESTING.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Running
+
+```
+dotnet build housecarl.sln -c Release
+dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"
+dotnet src/housecarl-generator/bin/Release/net9.0/housecarl-generator.dll ci-all
+```
+
+The one `bridge`-tier test shells the generator's `ci-all` itself; CI runs `ci-all` as its own step, so the filter leaves it out there.
```
Avick3110: **The documented local command excludes the bridge tier this sentence explains.**

The Running block above gives `--filter "tier!=bridge"` as the local command; this line then says the bridge test exists so that a local `dotnet test` also runs `ci-all`, and that CI's filter leaves it out "there". Run the three commands exactly as written and the bridge tier runs nowhere — the one test whose stated purpose is a single local entry point is filtered out locally and excluded in CI.

Either drop the filter from the local command, or say plainly that the block runs `ci-all` explicitly and the bridge test is for a no-filter local run.

Avick3110:
## Code review — 9 findings

Two are posted inline (`CLAUDE.md`, `standards/TESTING.md`). The seven below are in files this PR does not touch, so they cannot be anchored to a diff line. Six of them are one theme: **ADR 0001 and 0003 are deleted here, but eleven live references to them survive across the repo.** The PR body says the removal took "the three code comments that pointed at them" — it took two files' worth (`CiProbeAttribute.cs`, `CiAll.cs`) and one bullet in `check-family-tests.md`.

CI stays green either way — nothing asserts on the doc paths, and `CiAll.cs` / `CiProbeAttribute.cs` are not in the gated residue baseline — so this is silent.

---

**1. `standards/HOUSECARL_SKILL_AUTHORING.md:204` — the one standard not rewritten cites the deleted naming standard**

> Tools are namespace-qualified (per `HOUSECARL_NAMING.md`, P1.8).

`HOUSECARL_NAMING.md` is deleted in this PR, and the replacement `standards/NAMING.md` has no P-numbered sections and does not carry the namespace-qualification rule, so the pointer resolves to nothing. The same line cites `housecarl_create_patch` (not in `ToolNames.cs`) and `mo2_mcp/tools_records.py:335` (the retired Python build). `CLAUDE.md` advertises `standards/` as "Testing, naming, and skill authoring", so a session is sent here. It is also now the only `HOUSECARL_`-prefixed file in a directory whose two new siblings dropped the prefix.

**2. `.github/workflows/ci.yml:87` — the residue-countdown step's rationale points at deleted ADR 0003**

> `# The residue countdown (ADR 0003, docs/decisions/): how much of the OLD`

That comment is the only explanation of a required CI step, and it names a file that is gone.

**3. `docs/decisions/0005-the-1x-tools-are-deleted-not-deprecated.md:63` and `:101` — a surviving ADR cross-references the deleted one**

> what the guard rewrite (ADR 0003) exists to leave behind

`docs/README.md` in this same PR says "A change of mind is a new record that names the one it replaces" — deleting a record other records still cite is the shape that sentence rules out.

**4. `docs/architecture/test-project-fixtures.md:115` — a section heading names the deleted ADR**

> `### ADR 0003 rule 2 — the scripts family was briefly in both harnesses; the overlap is closed`

"Rule 2" is defined only in the deleted ADR. The constraint survives as a heading with nothing behind it.

**5. `docs/architecture/check-family-tests.md:9` — inline ADR 0001 reference left after the link list was cleaned**

The PR removes the ADR 0003 bullet from this file's reference list at the bottom, but line 9 still reads "under ADR 0001 (a terse comment is terse because the knowledge has a public home)". Same file, same pass, missed citation.

**6. Five more ADR pointers in the test project**

`src/housecarl-mcp-tests/HarnessResidueTests.cs:11` and `:24`, `HarnessBridgeTests.cs:8`, `ScriptsWorldTests.cs:11`, `harness-residue-baseline.json:2`. Whoever executes the backlog item that deletes this machinery reads "the rules in ADR 0003 (docs/decisions/)" and cannot retrieve the rules those tests enforce.

**7. `src/housecarl-mcp/LoadOrderService.cs:5068` — a shipped refusal names the gap-report path this PR deletes**

> "If you have a case where it does have a defined meaning here, file it as a gap report — that is what would define it."

`gap-report.yml` is replaced by `feature-request.yml` here, so a modder who follows the instruction opens the issue chooser and sees only "Bug report" and "Feature request". `SourceChainProbe.cs:208` asserts the message contains "gap report", so the probe stays green while the named route no longer exists.

---

Suggested resolution for 2–6: either keep 0001 and 0003 as superseded records (the shape `docs/README.md` describes), or clean the eleven references in this PR. Leaving them dangling is the one option that costs a later reader time.

Avick3110:
All nine folded in the third commit.

- CLAUDE.md table: `src/housecarl-mcp/` now says `LoadOrderService.cs` holds most of the load-order logic; a sixth don't says not to add a domain to it.
- TESTING.md: the bridge-tier sentence now matches the command block (the block runs `ci-all` itself; the bridge test is for an unfiltered run).
- ADR 0001 and 0003 restored with `Status: superseded 2026-09-04` and one line each saying why they are kept. That resolves findings 2 through 6 without touching eleven files. 0003 can go once the backlog item that deletes the residue machinery has landed and nothing cites it.
- Skill-authoring standard line 204 points at `ToolNames.cs` and `standards/NAMING.md`; the dead tool name and Python path are gone. The file keeps its old name because twenty skill eval files cite it by that name.
- Issue template renamed "Gap report / feature request" so the shipped "file it as a gap report" message still routes to a form.

PR #502, opened 2026-09-04
Delete the tests that test the test suite

### LINKED ISSUE #426: [Bug] two apply-guard arms assert the absence of strings the codebase cannot emit, so they cannot fail

Avick3110:
Deletes the tests that test the test suite and the CI steps that ran them. `HarnessResidueTests` and its `harness-residue-baseline.json` counted probe files, lines and CI steps; `CiWorkflowGuardStepTests` held a generated block of `ci.yml` byte-equal; `CiProbeRosterTests` counted roster entries against a second derivation of the same list. None of them tests product behaviour. `ci.yml` loses both "Residue countdown" steps and the comment block explaining them; build, `ci-all`, `dotnet test` and the "the run was not empty" check stay.

Also deletes the two `apply-guard` arms from #426. Both asserted the absence of `"NOT landed"` / `"NOT carried by the written file"`, strings no production path emits, so neither could fail. ARM 6's superseded-clause assertion already carried the claim; ARM 7 was left counting ranks only, so it gets a render check on wording the code really emits (`the re-opened file did not answer for this op`), which is what its summary and banner have always claimed.

`docs/decisions/0003-guards-move-to-a-standard-test-project.md` is **kept**, as the superseded record it already is: ADR 0005 describes the countdown 0003 defined, and both are ARCHIVE class. Comments elsewhere that cited 0003 as a live rule are rewritten without the citation.

Closes #426

### Kept

- `ToolSurfaceCensusTests.cs` — its wire half drives the built server and asserts `tools/list` against what the tool-surface assembly declares, and the unit arms guard the product's MCP attributes (#470), not the suite's shape.
- `RepoProjectsTests.cs` — `RepoProjects` is not a helper for the deleted populations alone: the census above calls it at seven sites, all routing through `LocateAssembly`, and these two arms pin the app-base-first rule that once loaded a stale Debug dll.
- `freshness-capture-guard` — a real product probe over freshness and write capture, and `ci-all` does **not** run it (`Standalone = true`, its deferral arm needs a cold process). Deleting the step would have dropped real product coverage, so the generated block is replaced by a plain step. `CiAll.cs` and `ci.yml` now say that block is hand-maintained, and that probes are discovered only in the generator, core, mcp and setup assemblies.

### Gates

- `dotnet build housecarl.sln -c Release` — 0 errors.
- `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` — 984 passed, 0 failed (995 before; the 11 deleted cases are `HarnessResidueTests`, `CiWorkflowGuardStepTests` and `CiProbeRosterTests`).
- `ci-all` — 128/128 passed (128 before; no probe was removed).


src/housecarl-generator/CiAll.cs
```diff
@@ -37,8 +37,7 @@ public static class CiAll
     /// </summary>
     /// <remarks>
     /// A guard's assembly has to reference housecarl-core to see the attribute at all, and this project
-    /// references every houseCARL project, so the closure is the whole population. It is not taken on trust:
-    /// CiProbeRosterTests derives the same population from the repo's project list and holds the two equal.
+    /// references every houseCARL project, so the closure is the whole population.
```
Avick3110: **A `[CiProbe]` outside the generator's reference closure now runs nowhere, green.**

The sentence this hunk keeps — "this project references every houseCARL project, so the closure is the whole population" — is not true. `housecarl-generator.csproj` references `housecarl-core`, `housecarl-mcp` and `housecarl-setup`; it does not reference `housecarl-mcp-tests`. The sentence deleted right after it ("It is not taken on trust: `CiProbeRosterTests` derives the same population from the repo's project list and holds the two equal") was the thing that made the claim safe, and it is the claim that is load-bearing, not the countdown.

Concretely: put a `[CiProbe("foo")] public static int Foo(string[] args)` on a type in `housecarl-mcp-tests` — a natural place during a conversion, since that is where guards are meant to go now — and `DiscoverAssemblies` never walks that assembly. The verb is not in the roster, `ci-all` does not run it, `housecarl-generator.dll foo` refuses it as an unknown mode, and CI is green. `CiProbeRosterTests.EveryAttributedGuardInTheRepoIsOneTheRunnerFound_NotOneNothingRuns` derived the population from the csproj files instead of from a reference closure and failed naming the guard and its assembly; after this PR nothing does.

That arm's subject is "does CI run this guard", not "what shape is the suite", so it does not fall under the PR's thesis the way the residue counters do. Either keep it (it needs only `RepoProjects`, which survives) or make the remark honest about the bound it now has.

.github/workflows/ci.yml
```diff
@@ -84,51 +84,9 @@ jobs:
...
-      # outside the residue countdown's reach — the countdown claims completeness over this path, not over
-      # arbitrary shell.
-      # >>> generated: standalone CI guard steps — do not hand-edit
-      # Rendered from CiAll.StandaloneProbeNames. `if: !cancelled()` so one red run still shows whether these
-      # passed — the same "report every failure" posture as the ci-all runner itself.
-      - name: CI guard (standalone, own cold process) — freshness-capture-guard
+      # freshness-capture-guard runs as its own cold process: its deferral arm needs a write slow enough to
+      # straddle a fixed ~100ms sleep, which only holds in a cold process. The warm ci-all runner finishes the
+      # write too fast to land that race, so this probe is carved out here and ci-all does not run it.
+      - name: Probe — freshness-capture-guard (own cold process)
```
Avick3110, line 94: **A probe flagged `Standalone = true` from here on runs in neither harness, and nothing goes red.**

The rename hazard is covered — respell the verb and `Program.cs` refuses it as an unknown mode, so the step exits non-zero. The hazard the generated block covered is the other one: a *new* standalone probe.

`CiAll.Roster` filters `Standalone` entries out, so `ci-all` deliberately does not run them; this step is now hand-written, so nothing owes a second standalone probe a step. Add one — or flip an existing probe to `Standalone = true`, which is the documented way to carve a probe out of the warm runner — and it runs nowhere at all, with the job green and no line in the log saying a guard went missing. `CiWorkflowGuardStepTests` rendered this block from `CiAll.StandaloneProbeNames` and went red naming the verb whose step was owed.

`CiAll.cs:13` still states the invariant — "Standalone probes are dispatchable and counted but do not run here; **CI gives them their own step**" — and after this PR nothing enforces it. If the generated block is going, that sentence should say the step is hand-maintained, so the next person flagging a probe standalone knows they also owe this file an edit.

src/housecarl-mcp-tests/RepoProjectsTests.cs
```diff
@@ -1,85 +0,0 @@
...
-    ///
-    /// <para>The decoy is the failure as it happened: a project tree carrying a second configuration's build,
-    /// more recently written than the one under test. "Newest under bin/" loaded that one, into the default
-    /// load context, where every later Assembly.Load for the same name resolved to it too.</para>
-    ///
-    /// <para>Nothing is loaded here — the assertion is over the path chosen, and the arm ends by confirming no
-    /// assembly of that name entered the process.</para>
-    /// </summary>
-    [Fact]
-    public void TheDllBesideTheTestAssemblyWinsOverANewerOneInTheProjectTree_TheBuildUnderTestIsTheOneReflected()
```
Avick3110, line 33: **`RepoProjects` is not a helper that exists only to build the deleted populations — it is load-bearing for a test this PR keeps.**

The PR body's justification for this file is that it "covered a helper that exists only to build those populations". `RepoProjects.cs` survives, and `ToolSurfaceCensusTests` — explicitly on the Kept list — calls it at seven sites (`AllAssemblies()`, `All`, `BuiltAssembly`, `DirectoryFor`), every one of which routes through `LocateAssembly`. So the class stays, its consumer stays, and only its coverage goes.

What that coverage pinned is the app-base-first rule, and its docstring records the failure it was written for: "newest under bin/" once loaded a stale Debug dll during a Release run, `LoadFrom` put it in the default load context, and every later `Assembly.Load` for that name resolved to it too. Both derivations then reflected a build nobody asked for, order-dependently, and green. Simplify or reorder `LocateAssembly` after this merge and `ToolSurfaceCensusTests` asserts `tools/list` against the wrong assembly — passing or failing for reasons unrelated to the change under test, which is exactly the silent-wrong-answer shape the census exists to prevent.

The two arms are self-contained (plant a dll, assert the path, delete), load nothing into the process, and cost nothing to keep.

src/housecarl-generator/ApplyGuardProbe.cs
```diff
@@ -204,9 +201,6 @@ string RemoveRank(string key) =>
         int after = RanksIn(fx.ReplacerPath, fx.FactionFid);
         Check("two key-addressed Removes in one call: BOTH land (the count drops by two)",
             after == before - 2, $"{before} -> {after}\n{twoRemoves}");
-        Check("…and NEITHER is reported as not landed — the earlier op's count is behind the file's, not wrong",
-            !twoRemoves.Contains("NOT landed", StringComparison.Ordinal)
-            && !twoRemoves.Contains("NOT carried by the written file", StringComparison.Ordinal), twoRemoves);
     }
```
Avick3110, line 206: **The deletion is right about the arm being inert, but ARM 7's own docs still claim the coverage it just lost.**

Confirmed: no production path emits `NOT landed` or `NOT carried by the written file` — grepping `housecarl-core` and `housecarl-mcp` finds those strings only in comments describing the *old* divergence verdict, which ARM 8's summary says is gone. So the assertion could not fail and deleting it is correct.

What is left behind is that ARM 7 now never looks at `twoRemoves` at all: its only surviving assertion is `after == before - 2`, which is true whether or not the response slanders the earlier Remove. Meanwhile the summary above still says "Driven end-to-end, **and the disk count is asserted too, so the arm proves the write was right rather than merely that the render stayed quiet**" and the banner still reads "two key-addressed REMOVES in one call are **not slandered**". Both now describe a check that is not there, and the next change to `LaterOpTouchesSameLeaf` / `CountNeutralKeyedVerb` will read them as cover.

The live wording does exist, so the arm can be made honest rather than hollow — `LandedProvenance` emits `[as applied — the re-opened file did not answer for this op]`, which is what the earlier Remove would carry if the count-neutral exemption regressed:

```suggestion
        Check("two key-addressed Removes in one call: BOTH land (the count drops by two)",
            after == before - 2, $"{before} -> {after}\n{twoRemoves}");
        Check("…and NEITHER is slandered — no op's clause says the re-opened file could not answer for it",
            !twoRemoves.Contains("the re-opened file did not answer for this op", StringComparison.Ordinal), twoRemoves);
    }
```

If you would rather not add an assertion in a deletion PR, the alternative is to fix the two doc lines so they stop claiming a render check.

docs/decisions/0005-the-1x-tools-are-deleted-not-deprecated.md
```diff
@@ -98,7 +98,7 @@ Four consequences of the shape, each load-bearing:
   file whose subject is split shrinks in two steps, not one. A file that also guards engine behaviour
   keeps that half and its residue entry. On this change, the eight harness files enumerated on #486 also
   keep their **tool-layer** blocks: those blocks compile against the surviving render and service members,
-  and they go when those do. The countdown that measures the harness's remaining size (ADR 0003) is what
+  and they go when those do. The countdown that measures the harness's remaining size is what
```
Avick3110: **This ADR's own header forbids this edit, and the edit leaves the paragraph pointing at things the PR deletes.**

Line 3 of this file: "**Class:** ARCHIVE (ADR). Immutable once merged; superseded by a later ADR, never edited." This PR edits it twice, to drop two `(ADR 0003)` parentheticals.

Stripping the citations does not remove the dependency. The surrounding sentences still read "keeps that half and **its residue entry**", "is not **a countdown key**", "the arrears sit in **the checked-in baseline** as a number rather than as something a later session has to remember", and here, "**the countdown that measures the harness's remaining size** is what makes both states visible". After this PR `harness-residue-baseline.json` is gone, the countdown is gone, and ADR 0003 — which defined all three terms — is deleted rather than superseded, so `docs/decisions/` has a hole at 0003 and a reader tracing "the checked-in baseline" finds nothing anywhere in the tree or in the ADR sequence.

ADR 0003's own status line, added the day before this PR, says it was "kept because other records and tests cite it". The tests are gone; the record citation is still here. Under the class rule the retirement shape is a new ADR (0006) that marks 0003 superseded and states what replaced the countdown — nothing, deliberately — leaving 0003 in place and both sibling ADRs untouched. That costs one short file and keeps the trail resolvable.

Avick3110:
Five findings, dispositions (commits 5163cf2 and 75330b1):

- **RepoProjectsTests — folded, restored.** You are right: `RepoProjects` is called at seven sites by the kept census test, and the two arms pin a rule with a recorded real failure.
- **ApplyGuardProbe ARM 7 — folded, your suggestion taken verbatim.** The arm now asserts the live wording, so its summary and banner are true again.
- **CiAll reference closure — folded as a comment, refused as a test.** The sentence now names the four assemblies probes are discovered in and says a `[CiProbe]` anywhere else is not run. The roster-equality test stays deleted: a test that a test runner found every test is a guard over guards, which is what the reset rules out. Tests belong in the xUnit project, where `dotnet test` discovers them; nobody should be writing `[CiProbe]` in it.
- **Standalone CI step — folded as a comment, refused as enforcement.** `CiAll.cs` and the `ci.yml` comment now say the step is hand-maintained, so a new `Standalone` probe also needs a step added by hand. Same reasoning as above.
- **ADR 0003 / 0005 — folded differently.** Both restored from main. 0003 stays as the superseded record it already was, because 0005 describes the countdown it defined; no ADR 0006.

Tests 984 (11 cases deleted), ci-all 128/128, build clean.

PR #509, opened 2026-09-04
Report a plugin the compact/merge safety scan could not read

### LINKED ISSUE #437: [Bug] the external-referencer safety scan silently under-counts — a plugin unreadable after the index was built is skipped yet still counted as scanned
## Summary

`RemapEngine.IdentifyExternalReferencers` is the safety scan behind `housecarl_compact_plugin`'s promise: every plugin in the order was checked for references into the compact target. Its per-plugin catch would count and sample a failed plugin — but it never fires, because the swallow happens one layer down: `LoadOrderResolver.cs:884` is `try { ov = OpenOverlay(…); } catch { continue; }`.

Result: a plugin that becomes unreadable **after** the resolver's index was built is silently skipped by the scan while `PluginsScanned` still counts it — `unscannable=0`, `UnscannableSamples` empty, the plugin actively reported as covered.

## Measured (2026-08-27, one target + one referencer, hold spanning the whole identify call — deterministic, not timing-fragile)

```
unheld:             externalPlugins=[IdRef.esp] scanned=2 unscannable=0
locked-before-build: externalPlugins=[]         scanned=1 unscannable=0, exclusion named (LOUD)
locked-after-build:  externalPlugins=[]         scanned=2 unscannable=0 samples=[]   (SILENT)
```

The locked-after-build row is the live-server shape: holding a file does not change its mtime, so `RefreshIfStale` does not rebuild, and a plugin locked by xEdit / MO2 / the running game after the index was built takes the silent path.

## Impact

Q3: a compaction can proceed reporting a clean scan while a real-but-unreadable referencer exists; the renumber then breaks that referencer's references into the target. This is the same hazard family PR #436's branch closed at the write layer, sitting one layer upstream in pre-existing `main` code — merged into `main` untouched by that branch (outside-the-branch rule, the #324 shape).

## Relationship to existing issues

- **#353** — same family: a FAILURE-degraded read indistinguishable from a legitimate one; this is that class inside the safety scan, where the stake is a destructive proceed rather than a wrong read.
- **#367** — the identify pass's other blind spot (declarer-only dependents), different mechanism.
- **#435** — fellow residue filed from the same engagement's review rounds.

## Fix shape (not verified)

The resolver's per-plugin open failure must surface to the caller (count + sample), so the identify pass's existing unscannable accounting actually fires; the compact refusal/report can then state the scan's true coverage.

## Provenance

Found by the PR #436 fold session while measuring a reachability question the advisor lane raised; measured with a deterministic probe (since removed, tree clean); recorded in PR #436's body and the decision record's ADDENDUM 3. Filed by the advisor lane 2026-08-27. Unrouted and unmilestoned — lands at the next routing stroke, candidate beside #435.

Avick3110:
The external-reference pass that runs before a compact or merge renumber promises that every plugin in the order was checked for references into the target. It was not keeping that promise: `LoadOrderResolver.RecordsIn` swallowed a per-plugin open failure and simply moved on, so a plugin that opened when the index was built but is unreadable now — held open by xEdit, MO2 or the running game, with an unchanged mtime so nothing rebuilds — was skipped without a word while the identify pass still counted it as scanned. Its own unscannable count and samples never fired, and a compaction could proceed on a report of a clean scan and break that plugin's references. The stream now throws, naming the plugin; the identify pass counts it, keeps it out of the scanned total, and lists it separately with its cause and reason. The other callers of the stream were checked: `check_errors` and `validate_scripts` already turn a plugin-level fault into a named per-plugin scan error, `cross_plugin_query` already fails the call loudly with a named reason, and the dialogue merge gets a catch so its existing absentee accounting still fires.

Two things the pass reports follow from that. It has two causes, not one — a file that would not open, and a record throwing on top-level enumeration — so they are caught separately and each gets its own sentence: the first names the programs to close, the second gives the exception it faulted with and claims nothing about anything holding the file. And the plugin's reason is recorded unconditionally rather than sharing the 5-entry per-record sample budget, so a named plugin always carries its reason.

Compact's in-place lane refuses rather than warns. Overwriting the target or repointing referencers has no backup and no review step, the referencer list is only as good as the scan behind it, and `acknowledge=true` on the first call skips the confirm prompt entirely — so a run whose pass left any plugin unread is now refused, naming each plugin and its reason, before anything is written and whatever `acknowledge` says. This matches the existing refusal for a referencer houseCARL identified but cannot rewrite. The new-plugin lane keeps the note, since its output is reviewed before it replaces anything, and merge only ever writes a new file so its note stays a note.

Closes #437

`IdentifyScanCoverageTests.AReferencerLockedAfterTheIndexWasBuiltIsReportedUnscannable` builds a two-plugin world, builds the index, then holds the referencer open across the identify call. It fails on `main` at `UnscannableRecords` being 0 — the reported symptom — and passes here. `CompactUnscannableReferencerTests` drives compact end to end against a synthetic MO2 instance with the referencer held open: `InPlaceCompactRefusesWhenAReferencerCouldNotBeRead` checks the in-place refusal with `acknowledge=true` and that the target file is untouched, and `TheReportSaysAnUnreadablePluginCouldNotBeOpened` checks the new-plugin lane's report names the plugin and the open failure. Build clean, 987/987 tests, ci-all 128/128.

### DIFF
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -61,6 +61,19 @@ public readonly record struct WinnerInfo(FormKey FormKey, string WinnerPlugin, i
 public readonly record struct RecordStatus(
     FormKey FormKey, string RecordType, bool PluginWins, int OverrideDepth, IReadOnlyList<string> TouchingPlugins);
 
+/// <summary>A plugin in the load order could not be opened during a scan, although it opened when the index was
+/// built — it was moved, or another program (xEdit, MO2, the running game) is holding it. Thrown from the
+/// plugin-scoped record stream so a scan reports the plugin it could not read rather than skipping it silently.</summary>
+public sealed class PluginUnreadableException : Exception
+{
+    public string PluginName { get; }
+    public PluginUnreadableException(string pluginName, Exception inner)
+        : base($"could not read '{pluginName}': it opened when the load order was indexed but not now — another " +
+               "program is probably holding it open. Close that program and run this again. " +
+               $"({inner.GetType().Name}: {inner.Message})", inner)
+        => PluginName = pluginName;
+}
+
 /// <summary>A BASELINE master (Skyrim.esm / Update.esm) is active but cannot be opened. Every written plugin
 /// force-includes the baselines, and that force-include is derived from the known-master list, so quietly omitting one
 /// would emit a plugin missing a mandatory master. Thrown from the master-set builders — the single point every write
@@ -857,8 +870,10 @@ public sealed class LoadOrderResolver : IDisposable
         foreach (int i in ScopeIndices(plugins, s))
         {
             ISkyrimModGetter ov;
+            // A plugin that opened at build time but not now became unreadable after it — moved, or held open by
+            // another program. Surfaced with the name so a caller reports a partial scan instead of a clean one.
             try { ov = OpenOverlay(_paths[i], _dataDir); }
-            catch { continue; }
+            catch (Exception ex) { throw new PluginUnreadableException(_names[i], ex); }
             try
             {
                 IEnumerable<IMajorRecordGetter> recs = getterTypes is null
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -859,7 +859,10 @@ public sealed class LoadOrderResolver : IDisposable
     /// with that PLUGIN'S body in hand — the plugin-scoped path behind a plugin dump or content audit. A FormKey
     /// touched by more than one scoped plugin is yielded once per scoped plugin; the SERVICE de-dupes.
     /// Yields (FormKey, whole-order override-depth, the scoped plugin's body, the scoped plugin's
-    /// filename — so a caller can DISPLAY from the same body it filtered, not the winner). Holds nothing.</summary>
+    /// filename — so a caller can DISPLAY from the same body it filtered, not the winner). Holds nothing.
+    /// Throws <see cref="PluginUnreadableException"/> on a scoped plugin that opened at index-build time but cannot be
+    /// opened now, ending the stream — a caller must report that plugin rather than treat the scan as complete. With a
+    /// multi-plugin scope the throw ends the whole stream, so the plugins after it are unread too.</summary>
     public IEnumerable<(FormKey fk, int depth, IMajorRecordGetter body, string source)> RecordsIn(
         IReadOnlyList<string> plugins, IReadOnlyList<Type>? getterTypes)
         => RecordsIn(plugins, getterTypes, _snap);                         // ONE build for the whole scan (captured here, at the call)
--- a/src/housecarl-core/RemapEngine.cs
+++ b/src/housecarl-core/RemapEngine.cs
@@ -60,11 +60,23 @@ public static class RemapEngine
     /// the repoint path, which would report a false success. <paramref name="Record"/> is the overridden FormKey.</summary>
     public sealed record ExternalOverride(string Plugin, FormKey Record, string RecordType);
 
+    /// <summary>Why the identify pass could not read a plugin through. Two causes with two different remedies:
+    /// <see cref="Unopenable"/> is a file that would not open at all — almost always another program holding it, and
+    /// closing that program fixes the run; <see cref="EnumerationFault"/> is a file that opened and then threw part
+    /// way through, which no amount of closing programs changes. Telling a modder to close xEdit for the second
+    /// sends them round a loop that never ends, so the cause is carried, never assumed.</summary>
+    public enum UnscannableCause { Unopenable, EnumerationFault }
+
+    /// <summary>One plugin the identify pass could not read through: its name, which of the two causes, and the
+    /// underlying reason. The reason is recorded here unconditionally — it never competes for the per-record sample
+    /// budget, because this is the fault that decides whether an in-place compaction may proceed at all.</summary>
+    public sealed record UnscannablePlugin(string Plugin, UnscannableCause Cause, string Reason);
+
     /// <summary>The identify-pass result: every external reference found, the DISTINCT referencing plugins in load
     /// order (the opt-in-rewrite set), the external OVERRIDERS (detect and warn, not repointable), how many plugins
     /// were scanned THROUGH (a plugin whose scan faulted is not one of them), and the fault accounting — a record
-    /// whose link walk threw is counted and sampled, and a plugin that could not be read at all is named in
-    /// <see cref="UnscannablePlugins"/>; neither is ever silently skipped.</summary>
+    /// whose link walk threw is counted and sampled, and a plugin that could not be read through is named, with its
+    /// cause and reason, in <see cref="UnscannablePlugins"/>; neither is ever silently skipped.</summary>
     public sealed record IdentifyResult(
         IReadOnlyList<ExternalRef> Refs,
         IReadOnlyList<string> ExternalPlugins,
@@ -73,7 +85,7 @@ public static class RemapEngine
         IReadOnlyList<string> UnscannableSamples,
         IReadOnlyList<ExternalOverride> Overrides,
         IReadOnlyList<string> ExternalOverriders,
-        IReadOnlyList<string>? UnscannablePlugins = null)
+        IReadOnlyList<UnscannablePlugin>? UnscannablePlugins = null)
     {
         /// <summary>True when at least one plugin OUTSIDE the transform set references a remapped FormKey — the
         /// signal the default new-plugin path must not pass silently: fail loud and offer the opt-in rewrite.</summary>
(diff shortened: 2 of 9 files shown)


src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -857,8 +873,10 @@ IMajorRecordGetter FetchBody(OverlaySession session, int overlayIdx, FormKey fk)
         foreach (int i in ScopeIndices(plugins, s))
         {
             ISkyrimModGetter ov;
+            // A plugin that opened at build time but not now became unreadable after it — moved, or held open by
+            // another program. Surfaced with the name so a caller reports a partial scan instead of a clean one.
             try { ov = OpenOverlay(_paths[i], _dataDir); }
-            catch { continue; }
+            catch (Exception ex) { throw new PluginUnreadableException(_names[i], ex); }
```
Avick3110, line 879: **The sibling winner stream still swallows the same open failure.**

This fixes the plugin-scoped stream, but `WinnerRecordsOfType` a few lines up (line 846) still has `catch { continue; }` on the identical "opened at index-build time, unopenable now" case, with the comment "an unopenable plugin wins nothing (surfaced at build)" — which is exactly the assumption this PR disproves.

Scenario: MO2 or the running game holds `Foo.esp` open; the mtime is unchanged so nothing rebuilds. `housecarl_cross_plugin_query type=WEAP` takes the `WinnerRecordsOfType` arm (LoadOrderService.cs:3850), the overlay open for `Foo.esp` throws, it is swallowed at line 846, and every record `Foo.esp` wins is silently absent from the result with no note — a clean-looking scan of the whole order. `housecarl_effect_chain` (EffectChain.cs:119) and the dialogue winner sweep (DialogueValidate.cs:299) drop the same records the same way.

Same fail-open, same cause, one stream over. Worth closing here or tracking as a follow-up issue.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5894,6 +5894,14 @@ public WritePatchBuilder.CompactOutcome CompactPlugin(
                     foreach (var pl in id.ExternalPlugins.Take(25)) c.Append($"      · {pl}\n");
                     if (id.ExternalPlugins.Count > 25) c.Append($"      · … (+{id.ExternalPlugins.Count - 25} more)\n");
                 }
+                // The referencer list is only as good as the scan behind it, and this prompt is the last point the
```
Avick3110: **The unread plugin only warns on the irreversible lane, and this gate is skipped when `acknowledge=true` comes in on the first call.**

The PR body's failure — "a compaction could proceed on a report of a clean scan and break that plugin's references" — is still reachable:

`housecarl_compact_plugin plugin=Foo.esp in_place=true acknowledge=true`, with xEdit holding `Bar.esp` open. `Bar.esp` references records `Foo.esp` is about to renumber, but the identify pass could not read it, so `id.ExternalPlugins` is empty and `HasExternalReferencers` is false. The referencer refusal at 5814 does not fire. The `LocalizedAmong` pre-flight at 5871 — which deliberately fails closed on a referencer it cannot read — never sees `Bar.esp`, because it is fed only `id.ExternalPlugins`. And this gate is skipped because `acknowledge` is already true. `Foo.esp` is overwritten in place with no backup, `Bar.esp`'s links dangle, and the modder finds out from `AppendUnscannablePlugins` in the success render, after the fact.

The asymmetry is the part worth deciding on: compact *refuses* in the better-known case (a referencer it identified but cannot rewrite) and only *warns* in the strictly less-known one (a plugin it could not read at all, which may or may not be a referencer). On the in-place lane that looks like it should be a refusal, not a note — the modder can always re-run after closing the program.

src/housecarl-core/RemapEngine.cs
```diff
@@ -159,17 +161,20 @@ public static IdentifyResult IdentifyExternalReferencers(
...
             }
-            // The plugin enumeration itself faulting (a record that throws on top-level enumeration rather than on a
-            // link walk) is counted per-plugin and the pass continues — never an opaque whole-pass abort.
+            // The plugin enumeration itself faulting — a record that throws on top-level enumeration, or a file that
+            // cannot be opened at all — is counted per-plugin and the pass continues, never an opaque whole-pass
+            // abort. The plugin is NOT counted as scanned: its coverage is what the caller has to know about.
             catch (Exception ex)
             {
                 unscannable++;
+                unscannablePlugins.Add(plugin);
```
Avick3110: **This catch has two causes but the renders assert one of them as fact.**

It covers both an unopenable file (the new `PluginUnreadableException`) and a record that throws on top-level enumeration — the comment two lines up says so. But `WriteTools.AppendUnscannablePlugins` prints "probably held open by another program, so close xEdit, MO2 or Skyrim and run this again", and the confirm prompt at LoadOrderService.cs:5903 says the same.

Scenario: `Bar.esp` is perfectly readable but holds one record Mutagen throws on. It lands here, and the report tells the modder to close programs that are not holding anything. They close everything, re-run, get the identical message, and loop. The real reason survives only in `UnscannableSamples` — which the confirm prompt does not print at all.

Either split the two at the catch (`catch (PluginUnreadableException)` vs `catch (Exception)`) and give each its own sentence, or drop the cause claim from the prose and let the sample carry it.

src/housecarl-core/RemapEngine.cs
```diff
@@ -159,17 +161,20 @@ public static IdentifyResult IdentifyExternalReferencers(
...
                 if (unscannableSamples.Count < 5) unscannableSamples.Add($"{plugin} — record enumeration aborted: {ex.GetType().Name}: {ex.Message}");
```
Avick3110: **The plugin's reason can be dropped by the shared sample budget.**

This line shares the 5-entry `unscannableSamples` cap with the per-record faults at line 161. On a full-modlist scan where five records have already failed their link walk, a plugin that then cannot be opened gets named in `UnscannablePlugins` by line 172 but records no reason anywhere: the report names `Bar.esp` with only the generic "probably held open" prose, while the five samples shown all describe unrelated records in other plugins.

A per-plugin fault is a strictly bigger deal than a per-record one, so it should not lose the tie. Give the plugin-level faults their own small budget, or always reserve a slot for them.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5894,6 +5894,14 @@ public WritePatchBuilder.CompactOutcome CompactPlugin(
...
+                // modder can decline — so a plugin the pass could not read is named HERE, not only in the report.
+                if (id.UnscannablePlugins is { Count: > 0 } unread)
+                {
+                    c.Append($"  ! the external-reference pass could not fully read {string.Join(", ", unread.Take(25))}");
+                    if (unread.Count > 25) c.Append($" (+{unread.Count - 25} more)");
+                    c.Append(", probably held open by another program, so the list above does not cover it. Close xEdit, MO2 or Skyrim and run this again.\n");
```
Avick3110: **"the list above" is not always there.**

The referencer list is printed only under `willRepoint` (the block at 5890-5894). On an in-place compaction without `repoint_externals` the prompt reads:

```
CONFIRM in-place rewrite (your ORIGINAL file(s) will be rewritten - no houseCARL backup or undo; keep your own):
  - 'Foo.esp' will be OVERWRITTEN in place with its compacted form.
  ! the external-reference pass could not fully read Bar.esp, probably held open by another program, so the list above does not cover it. ...
```

There is no list above. This is the one screen the modder reads before authorising an unbackuppable overwrite, so the sentence should name the thing it means rather than a list that may not have been rendered — e.g. "so the external-referencer check does not cover it", which reads correctly in both shapes.

src/housecarl-core/LoadOrderResolver.cs
Avick3110: Real, and the same cause. Kept out of this PR: it is a different stream with three read-tool callers (cross_plugin_query, effect_chain, the dialogue winner sweep), each needing its own call on loud failure versus a named coverage gap plus a test, while #437 is scoped to the compact/merge safety scan. Filed as #510 with the fix shape. If you would rather it land here, say so and it goes on this branch. The other four findings are being folded now.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded — the in-place lane now refuses. When the target is overwritten or referencers are repointed and the identify pass left any plugin unread, compact returns a Fail naming each plugin and its reason, before anything is written and regardless of acknowledge. The new-plugin lane keeps the note, and the now-unreachable note in the confirm prompt is gone. Merge only ever writes a new file, so its note stays a note.

src/housecarl-core/RemapEngine.cs
Avick3110: Folded — the catch is split. A PluginUnreadableException gets the held-open sentence with the inner exception as the reason; a fault on top-level enumeration gets 'could not fully read <plugin>: <type>: <message>' and no held-open claim. The cause rides on each UnscannablePlugins entry, and one helper (WriteSentences.UnscannablePlugin) renders it for the report and the new refusal alike.

src/housecarl-core/RemapEngine.cs
Avick3110: Folded — UnscannablePlugins entries now carry name, cause and reason, recorded unconditionally, so a plugin-level fault can never be named without its reason. Plugin faults no longer write into unscannableSamples at all; the 5-entry cap is left as is for the per-record faults.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded — the prompt's note is gone, since that lane now refuses instead. The surviving render (the report, on the new-plugin lane) says 'so the external-referencer check does not cover it', which is true in both shapes, and lists each plugin with its own reason.

PR #512, opened 2026-09-04
Generate the third-party notices component table from the publish output

### LINKED ISSUE #423: THIRD-PARTY-NOTICES is out of date against the publish output

Avick3110:
`THIRD-PARTY-NOTICES.txt` claims to name every third-party component the plugin bundles in `server/`, with an exact version. Nothing derived that claim from what actually ships, and it drifted three times. This regenerates the file against a real `dotnet publish` of `housecarl-mcp` and makes the component table a build artifact rather than a hand-maintained list: `scripts/generate-notices.ps1` reads the publish output's `deps.json`, keeps only the packages that emitted a DLL beside the exe, resolves each to its NuGet licence expression, and rewrites the marked regions of both tracked copies from one string so they stay byte-identical. It runs as a new step 3 of `scripts/build-plugin.ps1`, inside the `-not $PluginTreeOnly` block right after the publish it reads, so a `-PluginTreeOnly` build never calls it with no publish. The GPLv3 section 6 corresponding-source line — the release and commit of the bundled Mutagen — is written the same way, from the Mutagen version the publish output ships; it is a second kind of generated region rather than a second mechanism. Only the licence texts and the per-component prose stay hand-authored.

Three data files carry what the publish output cannot say: `packaging/notices-display-names.json` (package id to listed name, for `Nifly` and `miniball-devel0`), `packaging/notices-licence-exceptions.json` (the nine packages whose metadata has no SPDX expression — vendored licence files, the deprecated `licenseUrl` form, and `TransparentValueObjects.Abstractions`, which declares nothing at all — each with the determined licence, how it was determined, and the version it was determined against), and `packaging/notices-mutagen-commits.json` (Mutagen release version to the repository commit that release is tagged at; the one entry, `0.54.4` to `0188012c607ce8bb283d2704400d37737f089134`, is the existing hand-written value, checked against the Mutagen GitHub tag). The script stops the build with the component named when a shipped DLL resolves to no package, when a package has neither an expression nor an exception entry, when an exception entry's version is not the version that ships (a package can relicense across versions — Reloaded.Memory was LGPLv3 before May 2023), when a copyleft component's metadata names no source URL, when a licence has no section in the file, or when the shipped Mutagen version has no recorded commit. The copyleft test matches "contains GPL", so LGPL and AGPL count.

CI keeps the tracked copies fresh. `generate-notices.ps1 -Check` generates into memory, writes nothing, and fails with one sentence naming the file, the first differing line and the fix. A new CI step publishes the server (Release, win-x64, framework-dependent — the same shape the build script uses) and runs it, so a dependency bump that leaves the notices file stale cannot merge.

Against the current graph the regenerated file gains nothing and loses five entries that do not ship: `System.Threading.Tasks.Extensions`, `Microsoft.Win32.Registry`, `Microsoft.NETCore.Platforms`, `System.Security.AccessControl` and `System.Security.Principal.Windows`, along with the caveat paragraph that covered them. All 54 shipped third-party versions were already correct — the file had been hand-swept during the Mutagen bump. Entries are now ordered by name.

Verified with `dotnet build housecarl.sln -c Release`, `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` (987 passed), `housecarl-generator ci-all` (all pass), and a full `scripts/build-plugin.ps1` run: 54 components written, and the repo-root copy, `plugin/` copy and `dist/housecarl/THIRD-PARTY-NOTICES.txt` compare byte-identical. Re-running the generator against the same publish leaves the files unchanged. `-Check` passes on that fresh tree and fails on a one-line hand edit of either tracked copy, naming the line. Emptying the commits data file, running with a missing `-PublishDir`, and pointing an exception entry at a version that does not ship were each checked by hand and each stop with their own sentence.

Closes #423

### DIFF
--- a/packaging/notices-licence-exceptions.json
+++ b/packaging/notices-licence-exceptions.json
@@ -1,0 +1,40 @@
+{
+  "_comment": "Shipped packages whose NuGet metadata carries no SPDX licence expression, with the licence determined by hand and how. Read by scripts/generate-notices.ps1: a package that ships with no expression and no entry here stops the build. Keep the reason with the entry so no determination has to be made twice.",
+
+  "Loqui": {
+    "licence": "GPL-3.0-only",
+    "why": "No licence element of any kind in the nuspec. github.com/Noggog/Loqui is GPL-3.0-only, as is the rest of the Mutagen first-party ecosystem it belongs to."
+  },
+  "Noggog.CSharpExt": {
+    "licence": "GPL-3.0-only",
+    "why": "No licence element of any kind in the nuspec. github.com/Noggog/CSharpExt is GPL-3.0-only, as is the rest of the Mutagen first-party ecosystem it belongs to."
+  },
+  "Nifly": {
+    "licence": "GPL-3.0-only",
+    "why": "The nuspec embeds the licence as a file rather than an expression. The bundled LICENSE and github.com/ousnius/NiflySharp are GPL-3.0-only."
+  },
+  "Reloaded.Memory": {
+    "licence": "GPL-3.0-only",
+    "why": "The nuspec embeds the licence as a file rather than an expression. The bundled LICENSE.md is GPL-3.0-only. It reaches the package as a dependency of NexusMods.Paths."
+  },
+  "K4os.Compression.LZ4": {
+    "licence": "MIT",
+    "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT LICENSE in github.com/MiloszKrajewski/K4os.Compression.LZ4."
+  },
+  "K4os.Compression.LZ4.Streams": {
+    "licence": "MIT",
+    "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT LICENSE in github.com/MiloszKrajewski/K4os.Compression.LZ4."
+  },
+  "K4os.Hash.xxHash": {
+    "licence": "MIT",
+    "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT LICENSE in github.com/MiloszKrajewski/K4os.Hash.xxHash."
+  },
+  "OneOf": {
+    "licence": "MIT",
+    "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT licence.md in github.com/mcintyre321/OneOf."
+  },
+  "TransparentValueObjects.Abstractions": {
+    "licence": "MIT",
+    "why": "No licence element, no licenseUrl and no projectUrl in the nuspec. The package owner on nuget.org is erri120, the upstream github.com/erri120/TransparentValueObjects is MIT, and the repository commit the nuspec itself records - ff32127019b0f53480f5e9c1867924b14138999c - resolves in that repository. It reaches the package as a dependency of GameFinder and NexusMods.Paths."
+  }
+}
--- a/packaging/notices-licence-exceptions.json
+++ b/packaging/notices-licence-exceptions.json
@@ -1,39 +1,48 @@
 {
-  "_comment": "Shipped packages whose NuGet metadata carries no SPDX licence expression, with the licence determined by hand and how. Read by scripts/generate-notices.ps1: a package that ships with no expression and no entry here stops the build. Keep the reason with the entry so no determination has to be made twice.",
+  "_comment": "Shipped packages whose NuGet metadata carries no SPDX licence expression, with the licence determined by hand and how. Read by scripts/generate-notices.ps1: a package that ships with no expression and no entry here stops the build, and so does an entry whose version is not the version that ships - a package can relicense across versions, so each determination is recorded against the version it was made against. Keep the reason with the entry so no determination has to be made twice.",
 
   "Loqui": {
+    "version": "3.7.0",
     "licence": "GPL-3.0-only",
     "why": "No licence element of any kind in the nuspec. github.com/Noggog/Loqui is GPL-3.0-only, as is the rest of the Mutagen first-party ecosystem it belongs to."
   },
   "Noggog.CSharpExt": {
+    "version": "4.3.0",
     "licence": "GPL-3.0-only",
     "why": "No licence element of any kind in the nuspec. github.com/Noggog/CSharpExt is GPL-3.0-only, as is the rest of the Mutagen first-party ecosystem it belongs to."
   },
   "Nifly": {
+    "version": "1.1.0",
     "licence": "GPL-3.0-only",
     "why": "The nuspec embeds the licence as a file rather than an expression. The bundled LICENSE and github.com/ousnius/NiflySharp are GPL-3.0-only."
   },
   "Reloaded.Memory": {
+    "version": "9.4.2",
     "licence": "GPL-3.0-only",
-    "why": "The nuspec embeds the licence as a file rather than an expression. The bundled LICENSE.md is GPL-3.0-only. It reaches the package as a dependency of NexusMods.Paths."
+    "why": "The nuspec embeds the licence as a file rather than an expression. The bundled LICENSE.md is GPL-3.0-only, and says the project was LGPLv3 before May 2023 - so this determination is version-specific. It reaches the package as a dependency of NexusMods.Paths."
   },
   "K4os.Compression.LZ4": {
+    "version": "1.3.8",
     "licence": "MIT",
     "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT LICENSE in github.com/MiloszKrajewski/K4os.Compression.LZ4."
   },
   "K4os.Compression.LZ4.Streams": {
+    "version": "1.3.8",
     "licence": "MIT",
     "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT LICENSE in github.com/MiloszKrajewski/K4os.Compression.LZ4."
   },
   "K4os.Hash.xxHash": {
+    "version": "1.0.8",
     "licence": "MIT",
     "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT LICENSE in github.com/MiloszKrajewski/K4os.Hash.xxHash."
   },
   "OneOf": {
+    "version": "3.0.271",
     "licence": "MIT",
     "why": "The nuspec uses the deprecated licenseUrl form, pointing at the MIT licence.md in github.com/mcintyre321/OneOf."
   },
   "TransparentValueObjects.Abstractions": {
+    "version": "1.1.0",
     "licence": "MIT",
     "why": "No licence element, no licenseUrl and no projectUrl in the nuspec. The package owner on nuget.org is erri120, the upstream github.com/erri120/TransparentValueObjects is MIT, and the repository commit the nuspec itself records - ff32127019b0f53480f5e9c1867924b14138999c - resolves in that repository. It reaches the package as a dependency of GameFinder and NexusMods.Paths."
   }
--- a/scripts/build-plugin.ps1
+++ b/scripts/build-plugin.ps1
@@ -67,16 +67,16 @@ Write-Host ("Building houseCARL v{0}" -f $Version) -ForegroundColor Green
 # Clean the WHOLE package root, not just dist/housecarl: stale package-root extras (codex/,
 # START-HERE.txt, a previous setup exe) would otherwise survive into the zip - a stale nested
 # codex/codex/ subtree did exactly that at the 1.2.2 build.
-Step '0/10' 'Clean dist/ (package root)'
+Step '0/11' 'Clean dist/ (package root)'
 if (Test-Path $PkgRoot) { Remove-Item $PkgRoot -Recurse -Force }
 New-Item -ItemType Directory -Path $DistRoot -Force | Out-Null
 
-# Steps 1-3 build the SERVER half of the tree. -PluginTreeOnly skips them: nothing the plugin
+# Steps 1-4 build the SERVER half of the tree. -PluginTreeOnly skips them: nothing the plugin
 # validator reads comes out of them, and they are the slow part (corpus reflection + two publishes).
 if (-not $PluginTreeOnly) {
 
   # ---- 1. regenerate the rulebook (corpus.json + mutagen-reference shards, in sync by construction)
-  Step '1/10' 'Regenerate corpus.json (generator)'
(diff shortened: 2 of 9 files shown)


scripts/build-plugin.ps1
```diff
@@ -86,13 +86,21 @@ if (-not (Test-Path $Exe)) { throw "server exe not produced at $Exe" }
...
 Get-ChildItem $ServerDir -Filter 'appsettings*.json' -ErrorAction SilentlyContinue | Remove-Item -Force
 
-# ---- 3. corpus beside the exe (the proven #1 requirement) ------------------
+# ---- 3. generated notices regions from the publish output ------------------
+# THIRD-PARTY-NOTICES.txt names every third-party component in server/ with its version, and names the
+# Mutagen release its corresponding-source line points at. Both are written here from the publish just
+# made, so neither can drift from what ships; the licence texts stay hand-authored. Both tracked copies
+# (repo root and plugin/) are rewritten from one string, so they stay byte-identical.
+Step '3/11' 'Write the generated notices regions from the publish output'
+& (Join-Path $PSScriptRoot 'generate-notices.ps1') -PublishDir $ServerDir -RepoRoot $RepoRoot
```
Avick3110: **This step needs the `-PluginTreeOnly` guard after the rebase.**

This branch's merge base is `2b2fc6a`. `origin/main` has moved on three commits since, and one of them (`af7fa30`) added `param([switch]$PluginTreeOnly)` and wrapped steps 1-3 in `if (-not $PluginTreeOnly) { ... }`; another (`2e9e5ea`) added a CI step that runs `./scripts/build-plugin.ps1 -PluginTreeOnly`.

This PR inserts its notices call at exactly that spot, unguarded. If the merge resolution leaves line 95 outside the guard, `-PluginTreeOnly` skips the publish, step 0 has already wiped `dist/`, and `generate-notices.ps1` is handed `dist/housecarl/server`, which does not exist. I ran the script with a missing `-PublishDir` and it dies at `Resolve-Path`:

```
Cannot find path 'C:\...\server' because it does not exist.
```

So the "Assemble the plugin tree" job turns red on every CI run and `claude plugin validate --strict` is skipped behind it.

Rebase onto current `main` and put the step inside the `if (-not $PluginTreeOnly)` block, beside the publish it reads.

scripts/generate-notices.ps1
```diff
@@ -0,0 +1,240 @@
+#requires -Version 5.1
+<#
+  generate-notices.ps1 - write the component table of THIRD-PARTY-NOTICES.txt from the publish output.
+
+  The notices file claims to list every third-party component the plugin bundles in server/, with an
+  exact version. That claim used to be hand-maintained and drifted three times (stale versions, missing
+  components, an unattributed Apache-2.0 component). This script derives the list instead: it reads the
+  publish output's deps.json, keeps only the packages that actually emitted a DLL beside the exe, resolves
+  each to its licence, and rewrites the marked regions of plugin/THIRD-PARTY-NOTICES.txt in place. The
```
Avick3110, line 9: **Nothing regenerates or verifies this in CI, so the tracked copies can still drift.**

The generation runs only inside a full local `build-plugin.ps1`. CI (`.github/workflows/ci.yml:118`) runs `build-plugin.ps1 -PluginTreeOnly`, which on current `main` skips the server publish and therefore this step, and no test drives the script (`grep -rn THIRD-PARTY --include=*.cs src/` finds nothing).

Concretely: bump Mutagen to 0.55.0 in a PR that does not happen to run a full local build. CI goes green, the PR merges, and both `THIRD-PARTY-NOTICES.txt` copies still claim `release 0.54.4; repository commit 0188012c...` and list the 0.54.4 versions of every Mutagen assembly. The shipped zip is right at release time because the release build regenerates it, but the repo copy - the one anyone reading the GitHub repo sees, and the one the licence claim is actually made in - is wrong until then. That is the drift #423 is about, just moved from "hand-maintained" to "regenerated only on the release path".

A `-Check` mode that regenerates into memory and diffs against the tracked files, run in CI after a publish, would close it.

scripts/generate-notices.ps1
```diff
@@ -0,0 +1,240 @@
...
+  and fails to parse.
+#>
+param(
+  [Parameter(Mandatory = $true)][string] $PublishDir,
+  [string] $RepoRoot
+)
+$ErrorActionPreference = 'Stop'
+
+if (-not $RepoRoot) { $RepoRoot = (Resolve-Path (Join-Path $PSScriptRoot '..')).Path }
+$PublishDir = (Resolve-Path $PublishDir).Path
```
Avick3110, line 52: **A missing `-PublishDir` gets a raw PowerShell error, not one plain sentence.**

Every other stop in this script names the component and says what to do - `"Mutagen 0.54.4 ships but packaging/notices-mutagen-commits.json records no commit for it; add ..."`. This line does not. Point it at a publish directory that was never produced (a `-PluginTreeOnly` build, a standalone run after a clean, a typo) and you get, reproduced:

```
Resolve-Path : Cannot find path 'C:\...\server' because it does not exist.
```

which names neither the script nor the fix. The `Test-Path` loop immediately below already produces the right shape of message for the six data files; `$PublishDir` wants the same treatment above it, e.g.

```suggestion
if (-not (Test-Path $PublishDir)) { throw "notices generation needs a publish directory, and there is nothing at $PublishDir; publish the server first" }
$PublishDir = (Resolve-Path $PublishDir).Path
```

packaging/notices-licence-exceptions.json
```diff
@@ -0,0 +1,40 @@
+{
+  "_comment": "Shipped packages whose NuGet metadata carries no SPDX licence expression, with the licence determined by hand and how. Read by scripts/generate-notices.ps1: a package that ships with no expression and no entry here stops the build. Keep the reason with the entry so no determination has to be made twice.",
```
Avick3110: **These entries are keyed by package id with no version, so a determination is carried forward silently across bumps.**

`generate-notices.ps1:136` does `Get-Prop $exceptions $id` with no version check, so once a package is in this file its hand-recorded licence applies to every future version that still ships no SPDX expression - which is the same staleness the generated table exists to remove, just pushed into the data file.

Not hypothetical for this graph. Reloaded.Memory's own bundled `LICENSE.md` (read out of the NuGet cache) says:

> Components of the Reloaded-Project are governed by the *GPLv3* license as of May 2023. Prior versions of the project were under the LGPLv3 license.

So that project has relicensed across versions before, and it still ships `<license type="file">` rather than an expression. A future bump that moves it again would be attributed under the stale licence with no build failure and no signal at all.

`packaging/notices-mutagen-commits.json` next door is version-keyed and stops the build on an unrecorded version. Keying these the same way - or at least recording the version each determination was made against and stopping when the shipped version is newer - gets the same loud failure.

scripts/generate-notices.ps1
```diff
@@ -0,0 +1,240 @@
...
+  if (-not $licence) {
+    throw "$id $version ships but declares no licence expression in its NuGet metadata; record the determined licence in packaging/notices-licence-exceptions.json"
+  }
+  $name = Get-Prop $displayNames $id
+  if (-not $name) { $name = $id }
+
+  $sourceUrl = $null
+  if ($facts) { $sourceUrl = Format-SourceUrl $facts.Url }
+  # A copyleft component with no source pointer would leave the GPLv3 section 6 claim unbacked.
+  if ($licence -like 'GPL*' -and -not $sourceUrl) {
```
Avick3110: **`-like 'GPL*'` misses `LGPL-*` and `AGPL-*`.**

`'LGPL-3.0-only' -like 'GPL*'` is `$false`. If a dependency ever resolves to LGPL or AGPL - Reloaded.Memory was LGPLv3 before May 2023, per its own bundled `LICENSE.md` - the build does stop once, for the missing licence section. But as soon as the maintainer adds that section, two checks stay silently off: this line never demands a repository/project URL, and line 229 (`$lic -like 'GPL*'`) renders the table through the no-source branch.

The file then carries a copyleft component with no source pointer while the prose two sections up still says "The other GPL components' sources are linked beside each entry below" - a silently degraded mode, which is the thing the design says a tool never does.

Matching on "contains GPL" fixes both sites:

```suggestion
  if ($licence -match 'GPL' -and -not $sourceUrl) {
```

scripts/build-plugin.ps1
Avick3110: Rebased onto current main; the notices step now sits inside the `if (-not $PluginTreeOnly)` block, right after the publish it reads (step 3 of 11), so a tree-only build never calls it.

scripts/generate-notices.ps1
Avick3110: Added your suggested Test-Path throw above Resolve-Path: "notices generation needs a publish directory, and there is nothing at <dir>; publish the server first".

scripts/generate-notices.ps1
Avick3110: Both sites now use `-match 'GPL'`, so LGPL and AGPL demand a source URL and render through the source branch.

scripts/generate-notices.ps1
Avick3110: Added a -Check mode: it generates into memory, writes nothing, and fails naming the file, the first differing line and the fix. CI now publishes the server (Release, win-x64, framework-dependent, same shape the build script uses) and runs it, so a dependency bump cannot merge with a stale notices file. Verified it passes on a fresh tree and fails on a one-line hand edit of either tracked copy.

packaging/notices-licence-exceptions.json
Avick3110: Each entry now records the version its determination was made against, and the generator stops the build when the shipped version differs, naming the package, both versions and the file. Filled in the nine current versions; Reloaded.Memory's reason now notes the pre-2023 LGPLv3 history you found.

PR #513, opened 2026-09-04
Report a plugin the whole-order winner scan could not read

### LINKED ISSUE #510: [Bug] the winner stream swallows a plugin unopenable after the index was built, so whole-order reads silently drop its records
`LoadOrderResolver.WinnerRecordsOfType` has the same `try { OpenOverlay } catch { continue; }` that #437 removed from `RecordsIn`. A plugin held open by MO2, xEdit or the running game after the index was built (mtime unchanged, so no rebuild) is skipped, and every record it wins is absent from the result with no note.

Reached by `housecarl_cross_plugin_query` (the `WinnerRecordsOfType` arm in `LoadOrderService`), `housecarl_effect_chain` (`EffectChain.cs`), and the dialogue winner sweep (`DialogueValidate.cs`). Each looks like a clean scan of the whole order.

Found in Aaron's review of PR #509. Kept out of that PR because the three callers are read tools with their own renders, each needing its own decision on loud failure versus a named coverage gap, and a test per caller. Fix shape: the resolver throws the same `PluginUnreadableException` #509 introduced, and each caller either fails naming the plugin or reports it as not covered.

Avick3110:
`LoadOrderResolver.WinnerRecordsOfType` swallowed a plugin that opened when the index was built but cannot be opened now — held open by xEdit, MO2 or the running game, with its last-write time unchanged so nothing rebuilds. Every record that plugin wins was simply absent, and each caller rendered a clean scan of the whole order: `housecarl_records` / `housecarl_cross_plugin_query` scanning by type, `housecarl_effect_chain`, the dialogue quest fan-out, and the SkyPatcher EditorID sweep behind the layer's no-op scan and the post-state read. All four now name the plugin and say what is missing.

**Iterator shape.** #509 made `RecordsIn` throw, because its caller wraps each plugin. This stream spans the whole order, so a throw mid-stream would lose every plugin after the bad one — trading a silent gap for a bigger one. `WinnerRecordsOfType` instead takes an optional collector: a caller that passes one gets the fault recorded and the scan carried on past that plugin, and a caller that passes none gets the throw. That keeps one primitive rather than a per-caller special case, and keeps the default loud: nothing can drop a plugin silently, whether or not it has somewhere to report it.

What each caller does with the gap: the cross-query scan and the effect chain append a named coverage gap to the scan note beside the existing per-record accounting, and keep scanning; the dialogue quest fan-out carries it on the quest report's own `ScanGaps` list; the SkyPatcher EditorID sweep marks the types whose sweep came up short, so a replay is refused by name only when its own lookups read one of those tables — a record the layer addresses by FormID is unaffected — and the layer's no-op scan names the plugin in its notes.

**Naming the gap is not enough if the line above it says the opposite.** Four renders asserted a definitive negative off a zero count alone, one line above the new note, so the note never retracted them. Each is now gated on the gap:

- The quest seed body no longer says the quest owns NO dialogue topics; with a gap it says none were found in the plugins the sweep could read, and prints a `coverage gap:` line per unread plugin.
- The unread-plugin gap left `InputIssues`, which is the CK-parity channel — the seed render gates the parity verdict on it and a json consumer reads it as the parity result. It has its own `ScanGaps` carrier and its own `scan_gaps` json key now, so a locked file no longer reads as a parity failure and the parity verdict still prints.
- The effect chain no longer says the effect is applied by no SPEL/ENCH/ALCH/SCRL/INGR in the active order; with a gap it names the scope it covered and the plugins it could not read.
- The paged scan header no longer says NO records match at any offset and to check the filter; with a gap it names the unread plugin instead, which is the one thing that is not the filter.

Closes #510

Tests are in `WinnerScanCoverageTests`, each holding the plugin open across the call with the index already built. One per caller for the gap itself: `TheCrossQueryScanNamesAPluginItCouldNotRead`, `TheEffectChainNamesAPluginItCouldNotRead`, `TheSkyPatcherNoOpScanNamesAPluginItCouldNotRead`, `TheQuestFanOutNamesAPluginItCouldNotRead`; each first asserts the unheld call finds the record, so it is about the lock and not an empty world. One per render for the false negative, asserting the definitive sentence is absent and the truthful one present: `ThePagedHeaderDoesNotBlameTheFilterWhenAPluginWasUnread`, `TheEffectChainRenderDoesNotClaimNoCarrierWhenAPluginWasUnread`, `TheQuestSeedRenderDoesNotClaimTheQuestOwnsNoTopics`, `TheQuestParityVerdictSurvivesAPluginItCouldNotRead`. All eight fail against `origin/main` and pass here. Build clean, 995/995 tests, ci-all 128/128.

### DIFF
--- a/src/housecarl-core/DialogueValidate.cs
+++ b/src/housecarl-core/DialogueValidate.cs
@@ -106,6 +106,12 @@ public sealed record DialogueValidationReport(
     /// for those input kinds). Empty for a DIAL input and for a gap-free record.</summary>
     public IReadOnlyList<DialogueIssue> InputIssues { get; init; } = Array.Empty<DialogueIssue>();
 
+    /// <summary>Plugins the topic fan-out could not read, one sentence each. NOT a finding about the input record,
+    /// so deliberately not in <see cref="InputIssues"/>: that is the CK-parity channel, and a file lock is not a
+    /// parity failure. It bounds what the report covers — a report carrying one lists fewer topics than the order
+    /// really holds, so no render may state "owns none" over it.</summary>
+    public IReadOnlyList<string> ScanGaps { get; init; } = Array.Empty<string>();
+
     /// <summary>The SEQ staleness/coverage lint, set only for a Start-Game-Enabled QUEST input; null otherwise
     /// (a non-SGE quest, or a DIAL input — a topic isn't a quest, so a .seq isn't its concern).</summary>
     public SeqLintFinding? SeqLint { get; init; }
@@ -331,13 +337,13 @@ public static class DialogueValidate
                 var questGaps = DialogueCkParity.MissingQuestDefaults(quest)
                     .Select(g => GapIssue("Quest", fk, g)).ToList();
                 // A plugin the fan-out could not read may own topics of this quest, so the topic list below it is
-                // incomplete — a Problem, because a clean pass over a short list is the misleading answer here.
-                foreach (var u in unreadable)
-                    questGaps.Add(new DialogueIssue(DialogueIssueSeverity.Problem,
-                        $"{u.Message} Any topic of this quest that plugin owns is missing from this report."));
+                // incomplete. Its own carrier, not the parity channel: a file lock is not a CK-parity failure.
+                var scanGaps = unreadable
+                    .Select(u => $"{u.Message} Any topic of this quest that plugin owns is missing from this report.")
+                    .ToList();
 
                 return new DialogueValidationReport(fk, "quest", quest.EditorID ?? "", win.Value.WinnerPlugin, topics)
-                    { ReadIncomplete = av.ReadIncomplete, SeqLint = seqLint, InputIssues = questGaps };
+                    { ReadIncomplete = av.ReadIncomplete, SeqLint = seqLint, InputIssues = questGaps, ScanGaps = scanGaps };
             }
 
             // DLVW / DLBR inputs: a RECORD-LEVEL CK-parity check — these carry no INFO list, so there is no topic
--- a/src/housecarl-core/DialogueValidate.cs
+++ b/src/housecarl-core/DialogueValidate.cs
@@ -302,7 +302,11 @@ public static class DialogueValidate
                 var seqLint = CheckSeq(view, av, fk, quest, win.Value.WinnerPlugin);
 
                 var topics = new List<TopicValidation>();
-                foreach (var (tfk, _, tbody) in view.WinnerRecordsOfType(DialTypes))
+                // A plugin that cannot be read now contributes no topics to the fan-out. Collected so the sweep
+                // covers the rest of the order, and named below — the same never-a-silently-thinner-answer rule the
+                // per-topic unread accounting follows.
+                var unreadable = new List<PluginUnreadableException>();
+                foreach (var (tfk, _, tbody) in view.WinnerRecordsOfType(DialTypes, unreadable))
                 {
                     if (tbody is not IDialogTopicGetter dt) continue;
                     if (NonNull(dt.Quest.FormKeyNullable) is not { } qk || qk != fk) continue;
@@ -326,6 +330,11 @@ public static class DialogueValidate
                 // never judges the ANAM VALUE (an edited quest's ANAM is a legitimate CK high-water mark).
                 var questGaps = DialogueCkParity.MissingQuestDefaults(quest)
                     .Select(g => GapIssue("Quest", fk, g)).ToList();
+                // A plugin the fan-out could not read may own topics of this quest, so the topic list below it is
+                // incomplete — a Problem, because a clean pass over a short list is the misleading answer here.
+                foreach (var u in unreadable)
+                    questGaps.Add(new DialogueIssue(DialogueIssueSeverity.Problem,
+                        $"{u.Message} Any topic of this quest that plugin owns is missing from this report."));
 
                 return new DialogueValidationReport(fk, "quest", quest.EditorID ?? "", win.Value.WinnerPlugin, topics)
                     { ReadIncomplete = av.ReadIncomplete, SeqLint = seqLint, InputIssues = questGaps };
--- a/src/housecarl-core/EffectChain.cs
+++ b/src/housecarl-core/EffectChain.cs
@@ -114,9 +114,12 @@ public static class EffectChain
         var rows = new List<EffectChainRow>();
         int total = 0, unscannable = 0;
         var samples = new List<string>();
+        // A plugin that cannot be opened now wins carriers this scan cannot see. Collected so the scan covers the
+        // rest of the order and the note names the gap, rather than the whole call ending at the bad plugin.
+        var unreadable = new List<PluginUnreadableException>();
         try
         {
-            foreach (var (fk, _, body) in view.WinnerRecordsOfType(scope))
+            foreach (var (fk, _, body) in view.WinnerRecordsOfType(scope, unreadable))
             {
                 // PER-RECORD FAULT ISOLATION (twin of the cross_plugin_query scan): Match lazily parses subrecord
                 // content, so ONE record Mutagen can't parse is excluded + accounted, not an opaque whole-call abort.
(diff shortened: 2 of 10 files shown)


src/housecarl-core/DialogueValidate.cs
```diff
@@ -326,6 +330,11 @@ Dictionary<FormKey, InfoOrderView> OrdersFor(IReadOnlyCollection<FormKey> topicF
                 // never judges the ANAM VALUE (an edited quest's ANAM is a legitimate CK high-water mark).
                 var questGaps = DialogueCkParity.MissingQuestDefaults(quest)
                     .Select(g => GapIssue("Quest", fk, g)).ToList();
+                // A plugin the fan-out could not read may own topics of this quest, so the topic list below it is
+                // incomplete — a Problem, because a clean pass over a short list is the misleading answer here.
+                foreach (var u in unreadable)
+                    questGaps.Add(new DialogueIssue(DialogueIssueSeverity.Problem,
```
Avick3110: **The unread-plugin gap lands in the CK-parity channel.**

`InputIssues` is not a general findings bag on a quest report — it is the record-parity result, and two consumers read it that way:

- `DialogueSweepRender.ComposeSeedBody` gates the parity verdict on `checks.HasFlag(DialogueChecks.RecordParity) && r.InputIssues.Count == 0` (DialogueSweepRender.cs:107). Adding this Problem suppresses `DialogueQuestParityOk`.
- `DialogueKindChecks.Names` says so in its own doc comment: "Without it a consumer reading an empty `input_issues` could not tell a check that ran and passed from one that never ran."

Concretely, in this PR's own `TheQuestFanOutNamesAPluginItCouldNotRead` world: `HcWsQuest`'s ANAM and objective FNAMs are fine, but with `HcWsHeld.esp` held open the seed body no longer prints "quest CK parity: present and correct" — it prints only the file-lock note. A JSON consumer sees `checks: [record_parity, topic_graph]` with a non-empty `input_issues` and reads a file lock as a parity failure.

The gap needs its own carrier on the report (or `ReadIncomplete`), not the parity channel.

src/housecarl-core/DialogueValidate.cs
```diff
@@ -302,7 +302,11 @@ Dictionary<FormKey, InfoOrderView> OrdersFor(IReadOnlyCollection<FormKey> topicF
                 var seqLint = CheckSeq(view, av, fk, quest, win.Value.WinnerPlugin);
 
                 var topics = new List<TopicValidation>();
-                foreach (var (tfk, _, tbody) in view.WinnerRecordsOfType(DialTypes))
+                // A plugin that cannot be read now contributes no topics to the fan-out. Collected so the sweep
+                // covers the rest of the order, and named below — the same never-a-silently-thinner-answer rule the
+                // per-topic unread accounting follows.
+                var unreadable = new List<PluginUnreadableException>();
+                foreach (var (tfk, _, tbody) in view.WinnerRecordsOfType(DialTypes, unreadable))
```
Avick3110, line 315: **The quest render still asserts the quest owns NO topics.**

When the collector catches the plugin, this fan-out yields nothing and `Topics.Count == 0`. `DialogueSweepRender.cs:102` keys purely on `r.InputKind == "quest" && r.Topics.Count == 0` and emits `ReadSentences.DialogueSeedNoTopics`:

> this quest owns NO dialogue topics in the active load order — nothing to validate. If you expected some, check those topics set DialogTopic.Quest to this quest and that their plugin is enabled.

In `TheQuestFanOutNamesAPluginItCouldNotRead`'s own world that is false on all three counts: `HcWsHeldTopic` exists, it does set `Quest` to `HcWsQuest`, and `HcWsHeld.esp` is enabled. The remedy sends the user to audit records when the fix is to close the program holding the file.

The new `InputIssues` Problem prints after this sentence, so it never retracts it. The line needs suppressing (or rewording to "none in the plugins this sweep could read") when the report carries an unread-plugin gap — this is the same silent-complete-answer shape the PR is closing, one layer up.

src/housecarl-core/EffectChain.cs
```diff
@@ -150,6 +153,13 @@ public static EffectChainResult Resolve(LoadOrderResolver resolver, FormKey mgef
               + (unscannable > samples.Count ? $"; and {unscannable - samples.Count} more" : "")
               + $". Inspect one with {ToolNames.Records} formids=[the FormID] (per-field fault isolation applies).";
 
+        string? coverageNote = unreadable.Count == 0 ? null
+            : $"coverage gap: {unreadable.Count} plugin(s) could not be read, so any carrier they win is missing from this chain: "
+              + string.Join("; ", unreadable.Select(u => u.Message));
```
Avick3110, line 158: **The chain render still asserts the effect is carried by nothing.**

`RenderEffectChain` prints the "none" line off `r.Total == 0` alone (ReadTools.cs:360), one line *before* it appends this note (ReadTools.cs:361):

> none — HcWsEffect is a valid MagicEffect but is applied by no SPEL/ENCH/ALCH/SCRL/INGR in the active order.

That is exactly the state `TheEffectChainNamesAPluginItCouldNotRead` produces: `HcWsHeldSpell` does carry `HcWsEffect`, the plugin is just locked. The reader gets a flat, definitive negative and then a note that does not withdraw it — and "this effect is carried by nothing" is the kind of answer a modder acts on (deleting the MGEF, skipping the patch).

Worth gating that line on `r.ScanNote is null`, or rewording it to name the scope it actually covers.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3984,6 +4025,14 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
               + (unscannable > unscannableSamples.Count ? $"; and {unscannable - unscannableSamples.Count} more" : "")
               + $". Inspect one with {ToolNames.Records} formids=[the FormID] (per-field fault isolation applies).";
+        // Whole-plugin coverage gap: the scan carried on past a plugin it could not open, so the answer covers the
+        // rest of the order but not that plugin's winners. Named here so the result never reads as a clean scan.
+        if (unreadablePlugins.Count > 0)
+        {
+            string gap = $"coverage gap: {unreadablePlugins.Count} plugin(s) could not be read, so any record they win is missing from this answer: "
+                       + string.Join("; ", unreadablePlugins.Select(u => u.Message));
+            scanNote = scanNote is null ? gap : scanNote + " " + gap;
+        }
```
Avick3110, line 4035: **The paged header still blames the filter.**

The note this block builds is correct, but on the `offset=` path the header above it already made the opposite claim. `ReadTools.cs:241` decides off `q.Total == 0` alone:

> (offset=10 had nothing to skip — NO records match at any offset; check the filter, not the paging)

Run `housecarl_cross_plugin_query type=Weapon offset=10` against `WinnerScanCoverageTests`' own world with `HcWsHeld.esp` held open: total is 0 purely because of the lock, and the user is told in so many words that the cause is their filter — the one thing it is not.

Same fix shape as the note itself: gate that wording on `unreadablePlugins.Count == 0`.

src/housecarl-core/DialogueValidate.cs
Avick3110: Fixed in 6ed1cdb. The gap now rides a new `ScanGaps` list on the quest report instead of `InputIssues`, so `ComposeSeedBody` still prints `DialogueQuestParityOk` when the ANAM and objective FNAMs are fine and only the file is locked; the json seed gets its own `scan_gaps` array beside an empty `input_issues`. `TheQuestParityVerdictSurvivesAPluginItCouldNotRead` asserts both transports.

src/housecarl-core/DialogueValidate.cs
Avick3110: Fixed in 6ed1cdb. With a scan gap on the report the seed body prints `DialogueSeedNoTopicsRead` instead — none found in the plugins the sweep could read, not the whole load order — and a `coverage gap:` line per unread plugin follows it. `TheQuestSeedRenderDoesNotClaimTheQuestOwnsNoTopics` covers it.

src/housecarl-core/EffectChain.cs
Avick3110: Fixed in c52707b. `EffectChainResult` now carries `UnreadPlugins`, and `RenderEffectChain` prints the whole-order negative only when that is empty; otherwise it says none in the plugins the scan could read and names the ones it could not. Gating on the list rather than `ScanNote` keeps the per-record unscannable note from suppressing a true negative. `TheEffectChainRenderDoesNotClaimNoCarrierWhenAPluginWasUnread` covers it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Fixed in 5664c66. `CrossQueryOutcome` now carries `UnreadPlugins` and the offset header branches on it: with a plugin unread it says no records match in the plugins the scan could read, names them, and points at the note, instead of blaming the filter. `ThePagedHeaderDoesNotBlameTheFilterWhenAPluginWasUnread` asserts on the header line alone.

PR #514, opened 2026-09-04
Report how many files bsa_repack packed, and what it skipped

### LINKED ISSUE #404: [Bug] bsa_repack reports "packed" with no count while BSArch silently drops root-level loose files

BSArch silently drops loose files at the source ROOT (only subfoldered files archive). `BsaArchive.Pack` (`BsaArchive.cs:198-235`) never enumerates the source folder — it shells `pack <src> <tmp> <fmt> -mt`, checks only temp-exists/non-empty/fresh-mtime (`:220-221`), and reports success off `File.Exists && Length > 0` (`:234`). The success message (`BsaTools.cs:156-160`) reports `packed <name> → <path>` with **no file count** and no root-vs-subfolder distinction anywhere.

Fix shape (as recorded): enumerate the source, report "packed N of M, root files skipped" when they diverge, + a `BsaContractProbe` arm. Low-bite (real trees are subfoldered) but a silent-omission-as-success shape. Note: `bsa_list`/`Unpack` already carry an independent header-count cross-check (`:66-68`, `:125-127`) — Pack is the one lane without one.

**Routing:** row 7/W4 (file-substrate wave). Milestoned.

Avick3110:
BSArch packs only files that sit under a subfolder of the source, so loose files at the source root were silently dropped — and `housecarl_bsa_repack` reported a plain "packed <name>" with no count, so a vanished file read as success. `BsaArchive.Pack` now enumerates the source before packing, separating the files BSArch will archive from the root-level ones it will drop, and reads the produced archive's own header count back before the temp is moved over the target — the same cross-check `List` and `Unpack` already do. The success line says "packed N file(s) into <name>"; when root files exist, a sentence names how many were left out and what to do about them, on the failed pack as well as the successful one (an all-root source makes BSArch fail outright, which is where that case lands). A count that disagrees with the source refuses with both numbers and leaves any existing archive untouched. The header oracle reads `.bsa` only, so the `fo4`/`fo4dds`/`sf1`/`sf1dds`/`tes3` formats write an archive houseCARL cannot count: those packs report no count and say so rather than printing the source's number as the archive's. A `.bsa` that should have carried a header but could not be read is a different case, and gets a WARNING that the archive was placed unchecked.

The count matches what BSArch actually takes: it drops every `*.db` file (a `Thumbs.db` in a textures folder) and every file with no extension wherever they sit, so those are left out of the expected count and no longer make the cross-check refuse a correct pack. The count is a cross-check, not a precondition — a source folder that cannot be fully listed (a denied ACL, a dangling junction) packs unchecked with a sentence saying so, rather than refusing a pack that used to work. A run that produced no archive reports no count at all rather than zero.

`Pack` now returns `BsaPackResult` (counts plus the dropped root files) instead of the shared `BsaResult`; the probes that call it use only `Ran`/`Success`/`Raw`/`RunError` and are unchanged.

Closes #404

### DIFF
--- a/src/housecarl-core/BsaArchive.cs
+++ b/src/housecarl-core/BsaArchive.cs
@@ -22,17 +22,19 @@ public sealed record BsaResult(bool Success, string Raw, string? RunError)
 }
 
 /// <summary>What a pack source folder holds: <see cref="Archivable"/> is the files BSArch will archive (anything under a
-/// subfolder) and <see cref="RootFiles"/> names the loose files sitting at the source root, which BSArch silently drops.</summary>
+/// subfolder that BSArch does not drop for its name) and <see cref="RootFiles"/> names the loose files sitting at the
+/// source root, which BSArch silently drops — the full root listing, not just the ones BSArch would have taken.</summary>
 public sealed record BsaSourceScan(int Archivable, IReadOnlyList<string> RootFiles);
 
 /// <summary>The result of a pack (BSArch). <see cref="RunError"/> non-null ⇒ the pack never really ran (BSArch couldn't
-/// be launched / a stuck stale scratch or an unreadable source refused up front); <see cref="CountError"/> non-null ⇒ it
+/// be launched / a stuck stale scratch refused up front); <see cref="CountError"/> non-null ⇒ it
 /// ran but the produced archive's header count disagreed with the source, so nothing was placed. <see cref="Packed"/> is
 /// the produced archive's own file count, or null when it could not be read — the header oracle reads .bsa only, so a
-/// BA2 (fo4/sf1) or Morrowind archive packs unverified. <see cref="Expected"/> is what the source offered and
-/// <see cref="RootSkipped"/> the root-level files BSArch dropped.</summary>
+/// BA2 (fo4/sf1) or Morrowind archive packs unverified. <see cref="Expected"/> is what the source offered, or null when
+/// the source could not be fully scanned (so nothing was cross-checked), and <see cref="RootSkipped"/> the root-level
+/// files BSArch dropped.</summary>
 public sealed record BsaPackResult(
-    bool Success, int? Packed, int Expected, IReadOnlyList<string> RootSkipped, string Raw, string? RunError, string? CountError)
+    bool Success, int? Packed, int? Expected, IReadOnlyList<string> RootSkipped, string Raw, string? RunError, string? CountError)
 {
     public bool Ran => RunError is null;
 }
@@ -209,7 +211,7 @@ public static class BsaArchive
     /// NTFS-class timestamp resolution — on a FAT-class target a same-second pack could read as stale and fail LOUD (never
     /// falsely succeed). The source is enumerated first and the produced archive's own header count is checked against it
     /// before the move, so a short pack refuses instead of reporting success; files at the source ROOT are counted apart
-    /// because BSArch drops them. NOTE the caller must surface BSArch's caveat: a COMPRESSED archive breaks any
+    /// because BSArch drops them, and a source that cannot be fully enumerated packs unverified rather than refusing. NOTE the caller must surface BSArch's caveat: a COMPRESSED archive breaks any
     /// sounds/voices it contains.</summary>
     public static BsaPackResult Pack(string bsarchExe, string srcFolder, string archive, string formatFlag, bool compress, int timeoutMs = 600_000)
     {
@@ -223,18 +225,16 @@ public static class BsaArchive
             // "tmp exists and is non-empty" pass on the PREVIOUS run's bytes when BSArch fails this
             // run — a false success that ships wrong content over the target. Refuse loud instead;
             // nothing is packed, the prior archive is untouched.
-            return new BsaPackResult(false, 0, 0, nothing, "",
+            return new BsaPackResult(false, null, null, nothing, "",
                 $"a stale houseCARL scratch from a previous run is stuck at '{tmp}' and could not be removed " +
                 "(another process may hold it). Delete it and retry — this run packed nothing; the existing archive, if any, is untouched.", null);
 
-        // Count what the source offers before packing, so the produced archive can be cross-checked against it.
-        BsaSourceScan scan;
+        // Count what the source offers before packing, so the produced archive can be cross-checked against it. The count
+        // is a cross-check, not a precondition: a folder that cannot be fully listed (denied ACL, dangling junction) packs
+        // unverified and the caller says so, rather than refusing a pack that used to work.
+        BsaSourceScan? scan;
         try { scan = ScanPackSource(srcFolder); }
-        catch (Exception ex)
-        {
-            return new BsaPackResult(false, 0, 0, nothing, "",
-                $"could not read the source folder '{srcFolder}' ({ex.GetType().Name}: {ex.Message}) — check it exists and is readable.", null);
-        }
+        catch { scan = null; }
         var baselineUtc = DateTime.UtcNow;
 
         var args = new List<string> { "pack", srcFolder, tmp, formatFlag, "-mt" };
@@ -248,7 +248,7 @@ public static class BsaArchive
         if (!packed)   // BSArch couldn't run, or produced no/empty output — leave any prior archive untouched
         {
             try { if (File.Exists(tmp)) File.Delete(tmp); } catch { /* best-effort */ }
-            return new BsaPackResult(false, null, scan.Archivable, scan.RootFiles, (run.stdout + "\n" + run.stderr).Trim(), run.runError, null);
+            return new BsaPackResult(false, null, scan?.Archivable, scan?.RootFiles ?? nothing, (run.stdout + "\n" + run.stderr).Trim(), run.runError, null);
         }
 
         // Cross-check the produced archive's own header count against the source, the same oracle List/Unpack use. A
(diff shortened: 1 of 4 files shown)


src/housecarl-core/BsaArchive.cs
```diff
@@ -221,18 +248,48 @@ public static BsaResult Pack(string bsarchExe, string srcFolder, string archive,
...
     }
 
+    /// <summary>Count what a pack of <paramref name="srcFolder"/> will archive and what it will drop: BSArch packs only
+    /// files under a subfolder, so loose files at the source root are silently left out.</summary>
+    public static BsaSourceScan ScanPackSource(string srcFolder)
+    {
+        if (!Directory.Exists(srcFolder)) return new BsaSourceScan(0, Array.Empty<string>());
+        var root = Directory.GetFiles(srcFolder, "*", SearchOption.TopDirectoryOnly).Select(p => Path.GetFileName(p)).ToList();
+        int all = Directory.GetFiles(srcFolder, "*", SearchOption.AllDirectories).Length;
+        return new BsaSourceScan(all - root.Count, root);
```
Avick3110: **The scan counts files BSArch never packs, so the new cross-check refuses correct packs.**

Being under a subfolder is not the only thing BSArch requires. Checked against the shipped BSArch v0.9c: it also drops every `*.db` file (case-insensitively) and every file with no extension, wherever they sit. `.bak`, `.tmp`, `.7z`, `.zip`, `.log`, `.md`, `.psc`, `.dll` and a 0-byte `.nif` all pack fine, so the exclusion really is just those two.

Run end-to-end through `BsaArchive.Pack` with a source of `textures/a.dds` + `textures/Thumbs.db`:

```
Ran=True Success=False Packed=1 Expected=2
CountError='Out.bsa' packed 1 file(s) but the source folder offers 2 - refusing to place it;
           check the source folder for unreadable or locked files and repack.
target exists = False
```

Windows writes `Thumbs.db` into any folder of images opened in Explorer, so a `textures\` tree is the likely place to find one. Before this PR that repack succeeded; now it produces no archive at all, and the refusal names a cause ("unreadable or locked files") that has nothing to do with what happened, leaving nothing for the user to act on.

Filtering the scan the way BSArch filters also folds the second full recursive walk of the tree into the count:

```suggestion
        // BSArch packs nothing with a .db extension and nothing without one (checked against BSArch v0.9c).
        static bool Packs(string p) => Path.GetExtension(p) is { Length: > 0 } e && !e.Equals(".db", StringComparison.OrdinalIgnoreCase);
        var root = Directory.GetFiles(srcFolder, "*", SearchOption.TopDirectoryOnly).Select(p => Path.GetFileName(p)).ToList();
        int all = Directory.GetFiles(srcFolder, "*", SearchOption.AllDirectories).Count(Packs);
        return new BsaSourceScan(all - root.Count(Packs), root);
```

(`RootFiles` staying the full root listing is right - the note is about what the user left loose, not about what BSArch would have taken.)

src/housecarl-mcp/BsaTools.cs
```diff
@@ -146,13 +146,33 @@ string Refuse(string msg)
...
+    {
         var sb = new StringBuilder();
-        sb.Append("packed ").Append(name).Append(" (").Append(fmtFlag.TrimStart('-')).Append(compress ? ", compressed" : ", uncompressed").Append(") → ").Append(archive).Append('\n');
+        sb.Append("packed ");
+        if (r.Packed is { } packed) sb.Append(packed).Append(" file(s) into ");
+        sb.Append(name)
+          .Append(" (").Append(fmtFlag.TrimStart('-')).Append(compress ? ", compressed" : ", uncompressed").Append(") → ").Append(archive).Append('\n');
         sb.Append("(a new houseCARL mod folder — enable it in MO2 to use the archive.)");
+        if (r.Packed is null)
+            sb.Append("\nhouseCARL reads file counts from .bsa headers only, so this archive's contents were not counted or checked against the source.");
```
Avick3110: **A `.bsa` whose header could not be read reports as a format limitation.**

`Packed` is null for two different reasons: the format genuinely has no BSA header (`fo4`/`fo4dds`/`sf1`/`sf1dds`/`tes3`), or `ReadBsaHeader(tmp)` failed on something that was supposed to be one. An `-sse` pack that lands on the second - BSArch wrote a file without the `BSA\0` magic, or the scratch could not be reopened - prints plain success plus a sentence telling the user the miss was expected. The archive is placed with no cross-check and nothing says so.

`fmtFlag` is already a parameter here, so the two cases can be told apart:

```suggestion
        if (r.Packed is null)
            sb.Append(fmtFlag is "-fo4" or "-fo4dds" or "-sf1" or "-sf1dds" or "-tes3"
                ? "\nhouseCARL reads file counts from .bsa headers only, so this archive's contents were not counted or checked against the source."
                : $"\nWARNING: '{name}' was written but houseCARL could not read its .bsa header, so its contents were not counted or checked against the source - list it before relying on it.");
```

src/housecarl-core/BsaArchive.cs
```diff
@@ -205,9 +223,18 @@ public static BsaResult Pack(string bsarchExe, string srcFolder, string archive,
...
             // nothing is packed, the prior archive is untouched.
-            return new BsaResult(false, "",
+            return new BsaPackResult(false, 0, 0, nothing, "",
                 $"a stale houseCARL scratch from a previous run is stuck at '{tmp}' and could not be removed " +
-                "(another process may hold it). Delete it and retry — this run packed nothing; the existing archive, if any, is untouched.");
+                "(another process may hold it). Delete it and retry — this run packed nothing; the existing archive, if any, is untouched.", null);
+
+        // Count what the source offers before packing, so the produced archive can be cross-checked against it.
+        BsaSourceScan scan;
+        try { scan = ScanPackSource(srcFolder); }
```
Avick3110: **An enumeration that throws now blocks a pack that used to work.**

The `SearchOption` overloads of `Directory.GetFiles` run on `EnumerationOptions.Compatible`, which sets `IgnoreInaccessible = false`. One subdirectory the process cannot list - a denied ACL, a dangling junction - throws `UnauthorizedAccessException`/`IOException` out of `ScanPackSource`, and this catch turns it into "could not read the source folder ... check it exists and is readable", packing nothing. The previous code handed the folder straight to BSArch, which archived everything it could reach.

Counting is a cross-check, not a precondition, so a scan that cannot complete is a reason to pack unverified and say so - the same shape as the uncountable-format case - rather than a reason to refuse.

src/housecarl-core/BsaArchive.cs
```diff
@@ -205,9 +223,18 @@ public static BsaResult Pack(string bsarchExe, string srcFolder, string archive,
             // "tmp exists and is non-empty" pass on the PREVIOUS run's bytes when BSArch fails this
             // run — a false success that ships wrong content over the target. Refuse loud instead;
             // nothing is packed, the prior archive is untouched.
-            return new BsaResult(false, "",
+            return new BsaPackResult(false, 0, 0, nothing, "",
```
Avick3110: `Packed: 0` here (and at the source-scan refusal below) contradicts the record's own doc comment, which says `Packed` is "the produced archive's own file count, or null when it could not be read". Nothing was produced on either path, so the count is unknown, not zero - a caller rendering `Packed` without first checking `Ran` prints "packed 0 file(s)" for a run that never packed. The `!packed` path at line 251 already passes `null`; these two should match it.

```suggestion
            return new BsaPackResult(false, null, 0, nothing, "",
```

src/housecarl-mcp-tests/BsaPackCountTests.cs
```diff
@@ -0,0 +1,118 @@
+using HousecarlCore;
+using HousecarlMcp;
+using Xunit;
+
+namespace HousecarlMcpTests;
+
+/// <summary>The repack file-count lane: BSArch archives only files under a subfolder, so a pack enumerates the source
+/// first, names the root-level files it drops, and refuses when the produced archive's count disagrees with the source.
+/// No BSArch needed — the enumeration, the refusal sentence and the tool's own render are exercised directly.</summary>
```
Avick3110, line 11: Issue #404's recorded fix shape asks for a `BsaContractProbe` arm alongside the enumeration change, and that probe is the one that locks the pack contract - it still covers only provenance and format refusal. These xUnit tests are good, but nothing in `ci-all` exercises the source scan or the header cross-check, so a later change that makes `ScanPackSource` undercount, or that drops the cross-check, passes the probe suite unnoticed. That is the same gap that let the silent-drop bug ship.

Avick3110: Refuted, for Aaron to rule on. Since the reset, tests in src/housecarl-mcp-tests are the contract (standards/TESTING.md) and probes convert to tests on touch only; the source scan and the header cross-check are pinned by BsaPackCountTests, which CI runs. Adding a probe arm would put the same contract in two places. The real gap the review names, that nothing exercises the read-back against an archive BSArch actually produced, is true of a probe arm too (the probe's stub packer always fails); closing it needs a fake-packer seam or the BSA builder lifted out of BsaExtractProbe, and is worth its own issue if wanted. The other four findings are being folded now.

src/housecarl-core/BsaArchive.cs
Avick3110: Folded, with your `Packs` predicate — the scan now leaves out every `*.db` file and every extensionless file, in one recursive walk instead of two, and `RootFiles` stays the full root listing. New tests: `ADbFileIsNotArchivableBecauseBsarchDropsIt` (textures/a.dds + textures/Thumbs.db → 1) and `AFileWithNoExtensionIsNotArchivableBecauseBsarchDropsIt`; both fail on the old count.

src/housecarl-mcp/BsaTools.cs
Avick3110: Folded — `PackReport` now branches on `fmtFlag`: the header-less formats keep the "counts read from .bsa headers only" sentence, and anything else with a null count gets your WARNING naming the archive and telling the user to list it. Both renders are tested (`AnUncountableArchiveFormatReportsNoCount`, `ABsaWhoseHeaderCouldNotBeReadWarnsInsteadOfBlamingTheFormat`).

src/housecarl-core/BsaArchive.cs
Avick3110: Folded — a throwing `ScanPackSource` now leaves the scan unknown instead of refusing: the pack runs, the cross-check is skipped, and the report says the source could not be fully scanned so the archive was not checked against it (`Expected` is null, the same shape as the uncountable-format case). A test does produce an unlistable directory on Windows: `ASourceThatCannotBeListedPacksUncheckedInsteadOfRefusing` puts a deny-ListDirectory ACE on a subfolder (the same trick `PlaceAssetProbe` already uses) and restores it in a finally; it fails against the old refusal.

src/housecarl-core/BsaArchive.cs
Avick3110: Folded — the stale-scratch refusal now passes `null` for both `Packed` and `Expected`; the source-scan refusal below is gone entirely (the scan no longer blocks a pack). `ARunThatPackedNothingReportsNoCountAtAll` holds the scratch open so the pack refuses up front and asserts both are null.

PR #515, opened 2026-09-04
Accept and print the runtime FormID the game and the logs use

### LINKED ISSUE #457: [GAP] cannot look up a record by the ESL runtime FormID (FE0xxYYY) that logs, the console, and crash logs print
> *Filed from an external live-session gap report by **Jouten** (running 1.9.0), with their OK to be named; triaged, root-caused against source, and posted by AI. Machine paths and the reporter's in-progress mod details scrubbed.*

### Goal

Take a FormID straight out of `Papyrus.0.log` and read the record it names. The log printed a follower's skin as `[Armor < (FE028B19)>]`; the goal was that ARMO's armature and biped slots.

### What fell short

`housecarl_read_record` and `housecarl_cross_plugin_query` take `XXXXXX:Plugin.esp` only. A light plugin's **runtime** FormID is `FE0xxYYY` (`FE` + 3-hex light index + 3-hex local id) — which is the ONLY form the game, Papyrus logs, the console, and SKSE plugin logs ever show. There is no way to hand that string to the tool, and no documented helper that converts it. The user has to know that `FE028B19` means "light index 0x028, local 0xB19", find which plugin holds light slot 0x028, and rebuild `000B19:ThatPlugin.esp` by hand.

Tool calls tried:

    housecarl_read_record formid="028B19:<reporter's plugin>.esp"
    -> "028B19:<reporter's plugin>.esp is not present in the load order"

(The refusal is CORRECT for the string given — the gap is that the string a log hands you cannot be given at all.)

### Workaround used

`housecarl_cross_plugin_query type="Armor" editorid_contains="<shared prefix>"` — an EditorID substring search, possible only because the records happened to share a guessable prefix. It would not have worked for a record with an unknown or non-obvious EditorID, which is the normal case when you are chasing something a log named and nothing else.

### Extra sting — the hand-derived translation goes stale

The light index is **not stable across sessions**. The same skin record printed `FE028B19` on 2026-08-29 and `FE025B19` on 2026-08-30 after unrelated load-order movement. A hand-derived mapping rots; the tool (which reads the live order) always knows the current one.

### Frequency

Every time a log or the console is the starting point for a record lookup — most debugging sessions. ESL/ESPFE plugins are the majority of a modern load order.

### Ideal shape (reporter's)

Accept `FE0xxYYY` (with or without an `0x`) anywhere a FormID is accepted, resolve it against the active order's light-slot table, and report which plugin it resolved to. A `housecarl_resolve`-style "what is this runtime FormID" answer would also do.

- date: 2026-08-30  |  houseCARL version: 1.9.0

---

Valid, reproducible, non-duplicate. Real at HEAD (`94bf378`), on the same 1.9.0 the reporter runs.

**Input is FormKey-only at every door, and there is no shared parser.** Each tool calls Mutagen's `FormKey.Factory` on the raw string directly — `ReadTools.cs:51-52`, `LoadOrderService.cs:2694-2696` (`resolve`), and ~15 further sites across `batch_record_detail`, `diff_record`, walk seeds, `info_order`, and the bulk ops, plus the predicate lists in `FieldPredicate.cs:421,440,793`. `FE028B19` has no colon, so `FormKey.Factory` throws and the caller gets "bad FormID … Expected 'XXXXXX:Plugin.esp'". **No tool accepts a runtime FormID, and no tool emits one** — `ReadEngine.cs` renders `record.FormKey.ToString()` only. The bridge is missing in both directions, so you cannot round-trip *out* to the console either.

**It is a data gap as well as a plumbing gap.** `LoadOrderResolver.BuildIndex` (`LoadOrderResolver.cs:480-539`) does not track the light flag at all, and does not need it to read ESL plugins: on-disk ESL references are master-list-index encoded, never `0xFE` — pinned empirically over 1.55M real records by our own `EslFormIdProbe.cs:469-515`. `0xFE` is purely a runtime address the engine computes. So the light-index table is cheap to derive (active order + header flag) but **is not computed anywhere today**.

**Half the math already has a shared home.** `FormIdRange.cs:51-83` defines `LightMasterIndexPrefix = 0xFE000000` and `LocalObjectId(runtimeFormId)`. It is used in exactly two places — `SkseConfigReferenceExtractor.cs:103` and `SkyPatcherOverlay.cs:810` — both of which get the plugin name *from the token itself* (`Plugin.esp|0xFExxxYYY`). What is missing is the light-index -> plugin table needed when the plugin name is not supplied, which is precisely the reporter's case.

**2.0 does not close it.** The charter is explicit — *"**Out:** the engine. Resolver, WriteEngine, reflection generator, corpus rulebook"* — and the light-flag capture belongs inside `LoadOrderResolver.BuildIndex`. The SPEC's SELECT axis changes FormID input only in *quantity* (`SPEC.md:53`, "FormID set (inline / `@file`)"), not notation, and 2.0's own tool repeats the requirement verbatim: `RecordsTools.cs:100` — *"A FormID is 'XXXXXX:Plugin.esp' — 6 hex digits, a colon, the defining master's filename."* A user pasting `FE028B19` into `housecarl_records` gets the same refusal.

### Class, not a one-off

The shape is **"the identifier form the game, logs, and console actually emit is not an accepted input"**. Every instance below currently fails:

| Instance | Status |
|---|---|
| ESL runtime `FExxxYYY` (Papyrus log, console, SKSE logs) | this report — refused |
| Full-plugin runtime `XX######` (xEdit's FormID column, console) | identically refused, same missing map |
| Console-printed refIDs for placed references | same form, same refusal |
| **Crash-log FormIDs** | no parser, no bridge — so the stated crash-diagnosis capability terminates at this same door |
| Output direction (houseCARL -> console/xEdit) | nothing emits a runtime FormID; no round-trip out |
| `FF` dynamic forms | unhandled; would need a loud refusal |
| Distributor-config `Plugin.esp\|0xFExxxYYY` | **the one lane that works**, because the plugin name rides in the token |

No existing issue covers any of it — searched open and closed for *runtime FormID, FE0, light index, console, crash log, xEdit FormID, ESL*. Nearest neighbours are unrelated: #364 (compact/merge silent on the INI distributor layer's addressing) and #363 (merge drops the ESL header flag).

### Why this has no issue number until now

`dev/plans/RUNTIME_FORMID_BRIDGE_PLAN_2026-06-08.md` is a complete design for this capability, honestly marked *"Not yet implemented"*, that has sat parked since June. It predates the 2026-08-22 migration that made GitHub Issues the sole gap lane and was never carried across — so it has been invisible to every routing pass since. There is no plan-vs-code discrepancy; both say it was never built.

### Planning posture — the existing plan is NOT a build-go (maintainer directive, 2026-09-01)

**Do not pick up `RUNTIME_FORMID_BRIDGE_PLAN_2026-06-08.md` and build it.** Aaron has declined to endorse it blindly; the approach is to be **planned from scratch**. Treat the June plan as prior art to read and argue with — not as a design to execute, and not as a default whose conclusions carry unless disproven. In particular its resolved forks (the write-acceptance Option A, the ±2 step-1 proof requirement from Heisen's design review) are **re-openable**, not settled.

One piece of evidence the June plan did not have, contributed by this report and worth carrying into whatever replaces it: the same record printed `FE028B19` and `FE025B19` on consecutive days after unrelated load-order movement. That is live field confirmation that any mapping must be recomputed against the live order and never cached — and a direct answer to the design review's "the indices shuffle, they aren't reliable" objection.

**Not a cornerstone escalation.** The ask is an I/O notation at the door, not a bespoke verb, and it serves §1's "beneath the human-interface tools" rather than straining it. Routing and milestone are Aaron's call.

Avick3110:
houseCARL now takes the FormID the game itself uses. A modder reading a crash log, a Papyrus log, an SKSE log or the in-game console has eight hex digits and no plugin name — `FExxxYYY` for a light plugin, `XX######` for a full one — and until now had to translate that by hand before they could ask about the record. Every parameter that holds nothing but FormIDs accepts that form, resolved against the load order as it stands at the call: the resolver builds the light and full runtime index tables from the order's own plugin kinds, so nothing is cached and the light index moves when the order does. The `XXXXXX:Plugin.esp` form is unchanged and stays the way to read a plugin that is not active.

It comes back out too. Every rendered record carries its runtime FormID beside its FormKey — `runtime=` in the text lanes, `runtime_formid` in json, in the scan's dense columns and in spilled artifacts — so a record found here can be carried straight to the console or xEdit. An index no active plugin occupies is refused by name, an `FF` dynamic form is refused as belonging to no plugin, and a plugin whose kind could not be read blocks only the slots at or after its own position.

Two cases where the honest answer is not an eight-hex form:

- **A light-flagged plugin that was never compacted.** Its record at `0x001801` masks into the same runtime form as its neighbour at `0x000801`, and reading that form back returns the other record. Those records get one sentence saying the plugin is above the ESL window and to address it by plugin, rather than a form that lies.
- **A plugin outside the active order.** No runtime address exists, and none is printed.

One `FormIdDoor` per call parses every token in that call: it answers the plugin-qualified form without touching the load order at all, reaches for the index only when a runtime FormID actually arrives, and then once — so a call's own tokens cannot resolve against two adjacent index builds. The records tool hands that build to the scan it then runs, so the tokens it parsed and the records it matches come from one build.

All numeric FormID window facts — the ESL window, the index-byte mask, the light and full index shifts, the dynamic prefix — live in `FormIdRange`, and the resolver's local-id split goes through `FormIdRange.LocalObjectId` rather than restating the hex.

Verification:

```
dotnet build housecarl.sln -c Release                                    0 errors, 89 warnings (unchanged)
dotnet test src/housecarl-mcp-tests -c Release --no-build \
    --filter "tier!=bridge"                                             1029 passed, 0 failed
                                                                        128/128 passed
```

Closes #457

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,6 +13,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A FormID from the game, a log or a crash log now works as an input, and comes back as an output.**
+  Everywhere houseCARL takes a FormID it also takes the runtime form the game, the console, Papyrus logs,
+  SKSE logs and crash logs print: eight hex digits and no plugin name — `FExxxYYY` for a light plugin,
+  `XX######` for a full one, with or without a leading `0x`. It is resolved against the load order as it
+  stands at the call (the light index moves whenever the order does, so nothing is cached), and the answer
+  names the plugin it resolved to. Every rendered record now carries its runtime FormID beside the
+  `XXXXXX:Plugin.esp` form — `runtime=` in the text lanes, `runtime_formid` in json — so a record read here
+  can be carried straight back to the console or xEdit. An index no active plugin occupies is refused by
+  name, and so is an `FF` dynamic form, which exists only in a save game. The `XXXXXX:Plugin.esp` form is
+  unchanged, and stays the way to read a plugin that is not active.
+
 - **`THIRD-PARTY-NOTICES.txt` lists what the plugin actually bundles.** The component tables are now written
   by the build from the publish output, so the names and versions in the file are the ones in `server/`. The
   file no longer lists five packages that do not ship (`System.Threading.Tasks.Extensions`,
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -19,8 +19,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `XX######` for a full one, with or without a leading `0x`. It is resolved against the load order as it
   stands at the call (the light index moves whenever the order does, so nothing is cached), and the answer
   names the plugin it resolved to. Every rendered record now carries its runtime FormID beside the
-  `XXXXXX:Plugin.esp` form — `runtime=` in the text lanes, `runtime_formid` in json — so a record read here
-  can be carried straight back to the console or xEdit. An index no active plugin occupies is refused by
+  `XXXXXX:Plugin.esp` form — `runtime=` in the text lanes, `runtime_formid` in json and in the scan's dense
+  columns and spilled artifacts — so a record read here can be carried straight back to the console or xEdit.
+  A record in a plugin flagged light but never compacted, whose object ID is above the ESL window, gets a
+  sentence saying so where the form would have been rather than a form that names its in-window neighbour
+  too. An index no active plugin occupies is refused by
   name, and so is an `FF` dynamic form, which exists only in a save game. The `XXXXXX:Plugin.esp` form is
   unchanged, and stays the way to read a plugin that is not active. A `where=` operand compared against a field
   holds numbers and enum names too, so it takes the plugin-qualified form only; the `formid in [...]` list takes
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -14,15 +14,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
 ## Unreleased
 
 - **A FormID from the game, a log or a crash log now works as an input, and comes back as an output.**
-  Everywhere houseCARL takes a FormID it also takes the runtime form the game, the console, Papyrus logs,
-  SKSE logs and crash logs print: eight hex digits and no plugin name — `FExxxYYY` for a light plugin,
+  Wherever a parameter holds nothing but FormIDs, houseCARL now also takes the runtime form the game, the
+  console, Papyrus logs, SKSE logs and crash logs print: eight hex digits and no plugin name — `FExxxYYY` for a light plugin,
   `XX######` for a full one, with or without a leading `0x`. It is resolved against the load order as it
   stands at the call (the light index moves whenever the order does, so nothing is cached), and the answer
   names the plugin it resolved to. Every rendered record now carries its runtime FormID beside the
   `XXXXXX:Plugin.esp` form — `runtime=` in the text lanes, `runtime_formid` in json — so a record read here
   can be carried straight back to the console or xEdit. An index no active plugin occupies is refused by
   name, and so is an `FF` dynamic form, which exists only in a save game. The `XXXXXX:Plugin.esp` form is
-  unchanged, and stays the way to read a plugin that is not active.
+  unchanged, and stays the way to read a plugin that is not active. A `where=` operand compared against a field
+  holds numbers and enum names too, so it takes the plugin-qualified form only; the `formid in [...]` list takes
+  both.
 
 - **`THIRD-PARTY-NOTICES.txt` lists what the plugin actually bundles.** The component tables are now written
   by the build from the publish output, so the names and versions in the file are the ones in `server/`. The
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -144,10 +144,71 @@ public sealed class LoadOrderResolver : IDisposable
         public readonly int MaxDepth;
         public readonly string Epoch;                                         // this build's fingerprint — immutable with the snapshot
 
+        /// <summary>Per plugin index: is it a LIGHT plugin? Read off the header while the build already had the
+        /// overlay open (the .esl extension counts too — the engine force-treats it as light). This is what decides
+        /// whether the game addresses the plugin through the shared 0xFE light block or through its own load
+        /// index, so it is the only extra fact the runtime-FormID tables need.</summary>
+        public readonly bool[] Light;
+
+        /// <summary>Active plugins whose light flag could NOT be read, because the file could not be opened at all.
+        /// Their kind decides every runtime index AFTER them, so a runtime FormID cannot be answered while one is in
+        /// the order — named here so the refusal can say which plugin, rather than guessing from the extension.</summary>
+        public readonly IReadOnlyList<string> LightFlagUnknown;
+
+        /// <summary>The runtime address tables, built on first use (a load order that never sees a runtime FormID
+        /// never pays for them) and thrown away with this snapshot, so they refresh with the index.</summary>
+        public readonly Lazy<RuntimeSlots> Slots;
+
         public IndexSnapshot(Dictionary<FormKey, (int winner, int count)> index, Dictionary<FormKey, int[]> overriders,
                              List<string> loadFailures, HashSet<int> excluded, HashSet<int> unopenable,
-                             Dictionary<string, string> excludedPlugins, int maxDepth, string epoch)
-        { Index = index; Overriders = overriders; LoadFailures = loadFailures; Excluded = excluded; Unopenable = unopenable; ExcludedPlugins = excludedPlugins; MaxDepth = maxDepth; Epoch = epoch; }
+                             Dictionary<string, string> excludedPlugins, int maxDepth, string epoch,
+                             bool[] light, IReadOnlyList<string> lightFlagUnknown)
+        {
+            Index = index; Overriders = overriders; LoadFailures = loadFailures; Excluded = excluded; Unopenable = unopenable;
+            ExcludedPlugins = excludedPlugins; MaxDepth = maxDepth; Epoch = epoch; Light = light; LightFlagUnknown = lightFlagUnknown;
+            Slots = new Lazy<RuntimeSlots>(() => RuntimeSlots.Build(light));
+        }
+    }
+
+    /// <summary>The runtime address tables for ONE index build: which plugin the game loads at each load index and at
+    /// each light index. The engine numbers the two kinds separately — full plugins take 0x00, 0x01, … in order,
+    /// light plugins all share the 0xFE index and take 0x000, 0x001, … in order among themselves — so a runtime
+    /// FormID's high byte says which table to read.</summary>
+    internal sealed class RuntimeSlots
+    {
+        /// <summary>Load index → plugin index; -1 where no active plugin occupies the slot.</summary>
+        public readonly int[] FullToPlugin;
+        /// <summary>Light index → plugin index; -1 where no active plugin occupies the slot.</summary>
+        public readonly int[] LightToPlugin;
+        /// <summary>Plugin index → its own runtime slot; -1 for a plugin past the end of its block.</summary>
+        public readonly int[] SlotOfPlugin;
+        public readonly int FullCount;
+        public readonly int LightCount;
+
+        RuntimeSlots(int[] fullToPlugin, int[] lightToPlugin, int[] slotOfPlugin, int fullCount, int lightCount)
+        { FullToPlugin = fullToPlugin; LightToPlugin = lightToPlugin; SlotOfPlugin = slotOfPlugin; FullCount = fullCount; LightCount = lightCount; }
+
+        /// <summary>0xFE is the light block and 0xFF the runtime-dynamic block, so a full plugin can only load at
+        /// 0x00–0xFD.</summary>
+        public const int MaxFullSlots = 0xFE;
+        /// <summary>The light index is 12 bits.</summary>
+        public const int MaxLightSlots = 0x1000;
+
+        public static RuntimeSlots Build(bool[] light)
+        {
+            var full = new int[MaxFullSlots];
+            var lite = new int[MaxLightSlots];
+            Array.Fill(full, -1);
+            Array.Fill(lite, -1);
+            var slotOf = new int[light.Length];
+            int nFull = 0, nLight = 0;
+            for (int i = 0; i < light.Length; i++)
+            {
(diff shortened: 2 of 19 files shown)


plugin/CHANGELOG.md
```diff
@@ -13,6 +13,19 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A FormID from the game, a log or a crash log now works as an input, and comes back as an output.**
+  Wherever a parameter holds nothing but FormIDs, houseCARL now also takes the runtime form the game, the
+  console, Papyrus logs, SKSE logs and crash logs print: eight hex digits and no plugin name — `FExxxYYY` for a light plugin,
+  `XX######` for a full one, with or without a leading `0x`. It is resolved against the load order as it
+  stands at the call (the light index moves whenever the order does, so nothing is cached), and the answer
+  names the plugin it resolved to. Every rendered record now carries its runtime FormID beside the
```
Avick3110, line 21: **"Every rendered record now carries its runtime FormID" is not true of the scan lanes.**

The runtime form was added to four renderers only: `JsonWire.WriteReadRecord` (287), `JsonWire.RenderRecordsSummary` (427), `ReadTools.AppendRecord` (933) and `RecordsTools.RenderRecordsSummary` (1928). The cross-plugin scan's *summary* projection renders its identity elsewhere and was not touched — `JsonWire.WriteSummaryRow` (1181, reached from JsonWire.cs:1010) and its text twin in `ReadTools.cs:299-305` — and neither is the dense lane's fixed column list (`JsonWire.cs:1096`, `formid`/`editorid`/fields). Spilled artifact schemas (`Artifacts.cs:141`, `:231`) are unchanged too.

Concrete: `housecarl_records types=["WEAP"] where=["BasicStats.Damage >= 20"]` returns rows whose only identity is `XXXXXX:Plugin.esp`, while the same records fetched with `formids=` come back with `runtime_formid`. A skill that finds records by scan and then tells the modder what to type into the console has nothing to print, and a json consumer that reads `runtime_formid` gets `undefined` on exactly the lane most likely to produce a list worth checking in game.

Either carry it in the summary row and the dense columns, or narrow the changelog sentence to the lanes that actually have it.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6429,7 +6441,7 @@ public NpcCopyOutcome CopyNpcAppearance(
         FormKey targetFk = default;
         if (apply)
         {
-            try { targetFk = FormKey.Factory(targetFormid!.Trim()); }
+            try { targetFk = OpenFormIdDoor().Parse(targetFormid); }
```
Avick3110: **Two doors in one call, which is the thing `FormIdDoor` was written to prevent.**

`FormIdDoor`'s own docstring says: *"Reaching for the resolver per token re-stats every plugin in the order on every one, and lets the index rebuild mid-list so two tokens in one call resolve against different builds."* But the donor (6432) and the target (this line) each call `OpenFormIdDoor()`, so each gets its own `CaptureView()` — and `LoadOrderService.Resolver` runs its mtime freshness check and can swap in a rebuilt index on every access. The same shape is at `CopyTools.cs:91`/`:103`, and at `RecordsTools.cs:953`/`:1012` (and `:1328`/`:1311`), where `formids=` and `references=` in one call parse through separate doors — and neither of those doors is the view the scan itself later captures.

Concrete: `housecarl_copy_npc_appearance source_formid="FE012801" target_formid="FE013802"`. If MO2 re-sorts, or any plugin's mtime changes, between the two captures, the second door rebuilds the index and the two light indices are read off different orders — the donor comes from build A, the target from build B, and the copy writes a face onto a record the caller never named. There is no refusal, because each token resolved cleanly against the build it saw.

Open one door per call and pass it down (as `ApplyEdits` at 4519 and `RemoveRecords` at 5428 already do).

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -601,6 +680,72 @@ static string Concise(Exception ex)
...
+        {
+            int slot = (int)RuntimeFormId.LightIndex(value);
+            int p = slot < slots.LightToPlugin.Length ? slots.LightToPlugin[slot] : -1;
+            if (p < 0)
+                throw new FormatException(
+                    $"{RuntimeFormId.Format(value)} names light index 0x{slot:X3}, which no active plugin occupies " +
+                    $"(the load order has {slots.LightCount} light plugin(s)) — the plugin may be inactive; read an " +
+                    "inactive plugin with 'XXXXXX:Plugin.esp'.");
+            EnsureKindKnownUpTo(s, p, value);
+            return FormKey.Factory($"{value & FormIdRange.LightObjectIdMask:X6}:{_names[p]}");
```
Avick3110: **The light-vs-full local-id split is restated here instead of read from its single home.**

`FormIdRange.LocalObjectId` already does exactly this masking, and its own doc-comment says why it exists: *"This is the SINGLE home for the distributor-config FormID normalization the SkyPatcher overlay and the SKSE config audit both apply ... Getting the light-vs-full split wrong inverts every resolve verdict, so both consumers read it here rather than restating the hex."* This line and 721 restate it, and `RuntimeFormId.Compose` (RuntimeFormId.cs) restates it a third time in the inverse direction. `RuntimeFormId.DynamicPrefix` / `IndexByteMask` likewise restate the `0xFF000000` byte mask that `LocalObjectId` already carries as a literal.

Concrete: the next change to that window — a fix to `LocalObjectId`, or Skyrim gaining an overlay/medium master tier — updates `FormIdRange` and silently leaves these three sites on the old split. The SkyPatcher overlay and the SKSE config audit would then resolve `FExxxYYY` to one record and the new runtime-FormID door to another, in the same session, with nothing to tell the reader they disagreed.

`value & FormIdRange.LightObjectIdMask` here is `FormIdRange.LocalObjectId(value)`; use it, and give `Compose` the constants rather than fresh copies.

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -601,6 +680,72 @@ static string Concise(Exception ex)
...
+    /// gives it no address: its plugin is not active, a plugin whose kind could not be read sits at or before it (see
+    /// <see cref="IndexSnapshot.FirstUnknownKind"/>), or it sits past the last runtime slot.</summary>
+    string? RuntimeFormIdFor(IndexSnapshot s, FormKey fk)
+    {
+        if (fk.IsNull) return null;
+        if (!_nameToIdx.TryGetValue(fk.ModKey.FileName.ToString(), out int p)) return null;
+        if (s.FirstUnknownKind >= 0 && p >= s.FirstUnknownKind) return null;
+        int slot = s.Slots.Value.SlotOfPlugin[p];
+        if (slot < 0) return null;
+        return RuntimeFormId.Format(RuntimeFormId.Compose(s.Light[p], slot, fk.ID));
```
Avick3110: **A light-flagged plugin whose records sit outside the ESL window gets a printed runtime FormID that does not read back to it.**

`Compose` masks `fk.ID` with `LightObjectIdMask` (0xFFF) unconditionally, and nothing here checks that the record is actually inside the 0x800–0xFFF window. An esp-fe that was light-flagged without being compacted — the exact plugin class `housecarl_compact_plugin` and the ESL checks exist for, and one that is common enough in the wild — carries records at e.g. `0x001801`.

Concrete: that record renders as `runtime=FE0xx801`, the same eight hex digits the record at `0x000801` in the same plugin renders as. Feed either string back and `RuntimeToFormKey` returns `000801:Plugin.esp` — the other record, or none. The tool description promises the printed form "can be carried straight back to the console or xEdit", and `ThePrintedRuntimeFormIdReadsBackTheSameRecord` asserts the round-trip, but only over a well-formed fixture.

(The engine masks the same way, so the printed value may match what the console shows; the wrong part is that houseCARL states it as an unqualified answer. Either suppress the runtime form when `fk.ID > FormIdRange.EslWindowCeiling` on a light plugin, or say the plugin is out of window.)

plugin/CHANGELOG.md
Avick3110: Folded — the scan lanes carry it now. `JsonWire.WriteSummaryRow` and its text twin in `ReadTools` write the runtime form on the summary row, the dense lane gained a `runtime_formid` column in both its detail and summary column lists, and all three spilled artifact schemas name it. `RecordSummary` carries the form the same way `ReadOutcome` does, filled from the scan's own view. New tests: `TheScanSummaryRowCarriesTheRuntimeFormId` and `TheDenseScanLaneCarriesTheRuntimeFormIdColumn` (both red before). The changelog sentence stands and now also names the dense columns and artifacts.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. `CopyNpcAppearance` and `CopyTools` each open one door and parse both tokens through it, and `RecordsTools.Records` opens one door at the top of the call that every lane below uses — the walk seeds, `formids=` and `references=` in both the in-order and off-order lanes. `FormIdDoor` now exposes the build it captured, and the records tool passes it to `CrossQuery`/`OffOrderQuery` as a pinned view, so the scan runs on the same build its tokens were parsed against. Grepped every `OpenFormIdDoor()` call: no call site opens two now.

src/housecarl-core/LoadOrderResolver.cs
Avick3110: Folded. Both lines in `RuntimeToFormKey` go through `FormIdRange.LocalObjectId` now. `FormIdRange` gained `IndexByteMask`, `DynamicIndexPrefix`, `LightIndexShift`, `LightIndexMask` and `FullIndexShift`, `LocalObjectId` uses `IndexByteMask` instead of its own literal, and `RuntimeFormId` takes all of them from there — `DynamicPrefix`, `IndexByteMask`, `LightIndex`, `LoadIndex` and the composer. No numeric FormID constant is restated outside `FormIdRange`.

src/housecarl-core/LoadOrderResolver.cs
Avick3110: Folded, and it reproduced exactly as described: the fixture's `001801:HcRtWide.esp` printed `runtime=FE001801`, the same eight digits as its `000801` neighbour. `Compose` is now `TryCompose` and returns false when a light plugin's object ID is above `FormIdRange.EslWindowCeiling`; the resolver then carries a note instead of a form, and the lanes print that sentence where the runtime form would have been (`runtime=(...)` in text, `runtime_formid: null` plus `runtime_formid_note` in json, the parenthetical cell in dense) rather than an unqualified answer. New fixture: a plugin written full with records at `0x000801` and `0x001801`, light-flagged in its header afterwards, because Mutagen refuses to write the out-of-window record into a light master. Tests `ARecordOutsideTheEslWindowGetsNoRuntimeFormIdButAReason` (red before) and `TheInWindowRecordOfTheSameUncompactedPluginStillPrintsItsRuntimeFormId`.

Issue #271, opened 2026-07-22
[NOTE] enabled / NOT active on a located plugin means its mod folder is enabled, not that the plugin is ticked in plugins.txt


Avick3110:
> *Raised by the independent review of #270 (finding 6) as a pre-existing issue outside that PR's scope; verified against source and filed by AI.*

### Summary

The `Enabled` flag on a located plugin file means **"the layer providing this file is active in the VFS"** — the overwrite folder, an enabled mod folder, or the game Data folder. It does **not** mean the plugin is checked in `plugins.txt`. Two renderers state it as though it did:

| Site | Renders `!Enabled` as |
|---|---|
| `ReadTools.cs:1022` — `read_plugin_file` header | `NOT active` |
| `LoadOrderService.cs:2246` — `diff_record` off-order pole | `disabled` |

So a plugin that sits in an **enabled mod folder but is unticked in MO2's right pane** reports as enabled / active — while the game does not load it. That is the false-negative direction: the label reassures where it should warn.

- severity: cosmetic (misleading output), pre-existing — not introduced by #270
- class: misleading output

### Where the flag comes from

`Mo2LoadOrder.LocatePlugin` (`src/housecarl-core/Mo2LoadOrder.cs:271-276`) stamps `enabled` per candidate layer:

    TryDir(overwriteDir, "overwrite", enabled: true);
    foreach (var mod in comp.EnabledMods)  TryDir(..., enabled: true);
    foreach (var mod in comp.DisabledMods) TryDir(..., enabled: false);
    foreach (var dir in UnlistedModFolders(...)) TryDir(..., enabled: false);
    TryDir(dataDir, "game Data", enabled: true);

Every one of those is a statement about the *mod folder*, never about the plugin's tick state. The `mod=` lane (`LoadOrderService.cs:5276`) does the same thing via `comp.EnabledMods`.

### Why it's worth fixing rather than just renaming

The composition already carries the missing half — `comp.ActivePluginNames` / `comp.InactivePluginNames` — and `read_plugin_file` **already uses it**, one screen below, to classify the file's declared MASTERS:

    bool active = comp.ActivePluginNames.Contains(m)
                  || comp.ImplicitPluginNames.Any(x => x.Equals(m, StringComparison.OrdinalIgnoreCase));

[...]

Issue #269, opened 2026-07-22
[Bug] diff_record labels an ACTIVE, load-order-winning plugin OUT-OF-LOAD-ORDER / disabled when passed by path


Avick3110:
> *Filed from an external live-session bug report; triaged, root-caused against source, and posted by AI.*

### Summary

A plugin that is **enabled and load-order-winning** gets labeled `OUT-OF-LOAD-ORDER (direct path, disabled)` when it is addressed by its absolute file path instead of its filename. The diff *values* are correct — only the provenance label lies, and it lies in the one direction that matters (it says the live file isn't live).

- severity: cosmetic (misleading output) — Q3-adjacent: mid-diagnosis it briefly suggested the wrong file was the active one
- class: misleading output
- date: 2026-07-21  |  houseCARL version: 1.9.0+444df77a

### Repro

    housecarl_diff_record formid=<FormID>
      plugin_a=<absolute path to an old ESP backup — genuinely off-order>
      plugin_b=<absolute path to the SAME plugin's live copy in an ENABLED MO2 mod folder>

- expected: side b labeled as the active plugin — or at minimum not "disabled". The file *is* the enabled, load-order-winning plugin, merely addressed by path.
- actual: `b: …\Some Mod.esp [OUT-OF-LOAD-ORDER (direct path, disabled), Quest …]`
- reproduces for any `diff_record` where a pole is the absolute path of a plugin that is enabled in the load order.
- xEdit check: n/a — the diff values were correct; only the provenance label is wrong.

### Root cause (verified against source, current `main`)

Two independent contributions, both in the direct-path lane:

1. **The shared locate contract hardcodes `Enabled: false` for a direct path** — `src/housecarl-mcp/LoadOrderService.cs:5269`:

       if (LooksLikePath(plugin))
       {
           if (!File.Exists(plugin)) return new(null, "", false, null, $"no file at path '{plugin}'.");
           return new(Path.GetFullPath(plugin), "direct path", false, null, null);
           //                                                 ^^^^^ Enabled — never checked
       }

   The filename and `mod=` lanes below it both compute `Enabled` honestly (from `comp.EnabledMods` / the hit); the path lane just asserts `false`. The XML doc directly above (`:5257-5262`) records that this contract was unified *because* an inline copy had already diverged on exactly this flag — the path branch was the one case the unification never gave a real answer.

[...]

PR #516, opened 2026-09-04
Delete the LoadOrderService members with no shipped caller

### LINKED ISSUE #497: Delete the LoadOrderService members with no shipped caller
Filed 2026-09-03 on Aaron's word ("agreed on finding 9") at PR #496's gate — his review comment on `BulkPrimitivesWave3Probe.cs:648` and the PR body's open item 1. Follow-up to #486; sequenced in row 7 after #492.

## What is left behind

PR #496 deleted the eight zero-caller render/service members and their cascade. Three public `LoadOrderService` members were named on the PR as the same question and deliberately not decided there:

- `LoadOrderService.ReadPluginFile` (+ its `PluginFileOutcome` chain and the `WhyNotActive` / off-order-cause composition). Zero callers under `src/housecarl-mcp`. Driven from **twelve** files under `src/housecarl-generator` and `src/housecarl-mcp-tests` (`ReadPluginFileProbe.cs`, `BulkPrimitivesWave3Probe.cs`, `EpochGuardProbe.cs`, `ResolveHarness.cs`, …), so its deletion is a compiler-named cascade of its own. The facts its banner carried — #271 items 3/4, B12/B13's filename and `mod=` counting lanes, the json `why_not_active` arm — have no test anywhere today; `RecordsOffOrderPathTests.FactB4` carries the counting forward for the PATH address form only.
- `LoadOrderService.ResolveTree` — its only shipped caller was `AppendConflictTree`, deleted on #496.
- the public `ResolveRead(FormKey, …)` overload — named on the PR as "measurably unexercised", but it **has** an internal caller (`LoadOrderService.cs:2506`, the non-pinned branch of the epoch split). "Zero shipped callers" is true of two of the three, not three.

## What this issue asks for

1. **Derive the population, do not take the trio as given.** Every public/internal `LoadOrderService` member with no caller under `src/housecarl-mcp` (the shipped process), by script, committed with the PR. The three above are the seed, not the list — the PR body's own list was short by one direction already.
2. For each member: **delete** it (the compiler names every generator/test arm that reached it; each arm is dispositioned the #486 way — dies with the member, or owes a fresh structure-keyed test in the test project, #492's shape), or **give it a 2.0 caller** where a shipped surface is meant to reach it.
3. **The capability question, Aaron's:** is the 2.0 records surface meant to address a plugin by *filename* or by *`mod=`* (the way 1.x `read_plugin_file` did)? If yes, `ReadPluginFile`'s engine half owes a caller and the two unarmed address lanes owe arms; if no, it dies and the `mod=` / filename facts die with it as (a). #271's history is the reason this is not obvious: the `mod=` lane is where a duplication once survived a review round because the equality test failed silently there.

Ground: CLAUDE.md §3 third cornerstone (the population is derived from the surface it claims); ADR 0005 ("guards on a deleted tool die with the tool; behaviour that survives gets a fresh test").

Avick3110:
Deletes the `LoadOrderService` members that nothing in the shipped process calls, and moves the coverage that reached them onto surfaces that do ship. `scripts/find-unused-service-members.ps1` derives the population: it reads every public/internal member declared in the class body, then counts each mention under `src/housecarl-mcp` as a call only when the name is used bare or through a receiver holding the service — so `WritePatchBuilder.CreateRecords` and `p.View.ResolveTree` stop reading as calls to the service's own members of those names. That receiver test is what found two members the issue's seed did not name. No shipped tool changes what it answers: everything deleted was unreachable from the tool layer, and the one arm that had no shipped equivalent is called out below rather than quietly dropped.

**The derived population.** The script's declaration parser was wrong twice, and both were fixed in review: a generic parameter list, and then (review round 2) a tuple return type, which recorded every tuple-returning member under the name `public` or `static` and so could never report one dead. Corrected, run on `main`: **114 public/internal members, 110 distinct names, 8 uncalled**.

| Member | Disposition |
|---|---|
| `ReadPluginFile` (+ `PluginFileOutcome`, `PluginRecordRow`, `PluginTypeCount`) | deleted |
| `ResolveTree` | deleted — its shipped caller went on #496; `ResolveTreePinned` is what `records` uses |
| `CreateRecords` (the single-record overload) | deleted — `housecarl_create` goes through `CreateRecordsBatch` |
| `OrderWarnings` | deleted — no mention anywhere; the `_orderWarnings` field it exposed is still read by `StatusData` |
| `ForGuard` | kept. Zero shipped callers by design: it is the seam nine CI guards construct the service through. Deleting it deletes coverage of `CrossQuery`, not dead code. |
| `WithSourceChainForGuard` | kept — the second guard seam, surfaced by the generic-parameter fix. |
| `Dispose` | kept — the language calls it under `using` rather than any code naming it. |
| `Stats` | kept — surfaced by the tuple fix, and the third seam: the probes and tests warm the lazy index through it and read the epoch and counters off it (~100 call sites). Its doc comment claimed a stand-up/health check the product does not do; it now says what it is. |

The tuple fix surfaced eight members in total; the other seven (`PapyrusSourceImportDirs`, `SetInstance`, `ClassParentsForDecompile`, `Ladder`, `Classify`, `ScriptOutputContract`, `SeqOutputContract`) all have shipped callers. All three kept-but-uncalled kinds are named in the script's header, so the next reader is not left to rediscover them.

On this branch after the deletions the script reports **110 members, 106 distinct names, 4 uncalled** — `Dispose`, `ForGuard`, `Stats`, `WithSourceChainForGuard`, each explained above and in the script header. Nothing else is left unexplained.

**`ReadPluginFile` and the SPEC.** SPEC §2.2's S1 SOURCE axis says `named(plugin)` is "one pole, resolved wherever the plugin lives: active in the order *or* an on-disk file off-order (absorbs `read_plugin_file`)", and the shipped surface has it: `source="X.esp"`, `source={"file": "X.esp", "mod": "<folder>"}` and a direct path all resolve through `ProbeSourceArm`, which composes the off-order label from the same `PluginLocateResult.WhyNotActive` the deleted member rendered. So there is no open question about the address forms — the SPEC's promise ships. The one genuine narrowing is stated below rather than left to be discovered.

**Dispositions, one line each:**

- `ReadPluginFileProbe.cs` — deleted; its whole subject was the member. Its facts hold on the `records` source pole and are covered there (below), except the master advisory, whose splitter (`Mo2LoadOrder.SplitUnsatisfiedMasters`) is unchanged and covered by `CheckMasterRemedyTests`.
- `BulkPrimitivesWave3Probe`'s #271 arms — kept, repointed at `svc.ProbeSourceArm`, which is what `housecarl_records` resolves `source=` with. Every cause they pinned (unticked / shadowed / mod switched off / mod unlisted / unregistered / not an install copy, and the address forms agreeing on the cause) still runs, now against the shipped composed label instead of a DTO field.
- The same probe's "the shadowed copy still READS (66)" — kept, rewritten as a `RecordsTools.Records` read off that path, because locating a file and opening it are different steps.
- The same probe's two ACTIVE-lane arms (`replPath`, `dataServedPath`) — kept, but they now test the path-identity rule, not the locate: a pole naming the active order's own copy resolves active before any folder is searched. The locate's enabled-layer ordering (a DISABLED folder walked ahead of game Data must not decide which copy serves) therefore lost its red, so a new arm carries it on a name the order does **not** hold — a second game-Data plugin with a copy in the same disabled decoy folder, addressed by path. Breaking `JudgeServed` to take the first hit rather than the first enabled hit turns that arm red; restored, it passes.
- The same probe's `mod=` arms for a copy of an **ACTIVE** plugin name (the shadowed and the serving copy of `HcW3DiffRepl.esp`) — **deleted, a real narrowing.** `source=` is one pole, and a plugin active in the order resolves there before any folder is searched, so there is no shipped way to address a different on-disk copy of an active plugin's filename. Stated in the probe and here.
- `ConflictDiffProbe` — kept, repointed at `ResolveTreePinned` through a fresh capture of its own resolver; the arms are unchanged.
- `BulkCreateGuardProbe`, `DialogueCkParityGuardProbe`, `DialogueSubtypeMarkerGuardProbe`, `InPlaceProbe` and `BulkPrimitivesWave3Probe`'s create arm — kept, repointed at `CreateRecordsBatch` through `ProbeCreate.CreateOne`, a batch of one carrying the same origin label and refusal wording `housecarl_create` passes, so no arm asserts on a wording the shipped tool never emits.
- `CheckMasterRemedyTests.TheReadPluginFileAdvisoryAndTheCheckRemedyAgreeOnWhichMasterIsWhich` — deleted; it compared two surfaces and one is gone. The install/enable split it checked is already held on the check lane by three tests in the same file.
- `CheckMasterTickedButAbsentTests`' two engine-half tests — deleted; the fact that the satisfied master stays out of the shortfall moved onto the check-lane test in the same file, which now also asserts `Skyrim.esm` is not named on the install line.
- New: `RecordsPluginFileSourceTests.APluginNamedOutOfASwitchedOffModSaysThatModFolderIsOff` — the filename lane's not-active cause, and the counting `FactB4` holds for the path lane: the remedy stated once **and** the mod named once. The mod-name count belongs on this lane, since the filename label already names the folder (`mod 'OldMod' (DISABLED)`) while the path lane's `Where` is the constant `direct path` and names none — so this is where the duplication can arise.
- New: `RecordsDuplicateFilenameSourceTests` — a filename two mod folders provide is refused with both named and the disambiguating spelling handed over, and `mod=` decides which copy is read. Its own instance; the shared world has no duplicate filename.

`LoadOrderService.cs` loses 221 lines (9,119 to 8,898). No CHANGELOG line: no shipped tool's behaviour changes.

**Verification:** `dotnet build -c Release` clean; `dotnet test --filter "tier!=bridge"` 1030 passed, 0 failed; `ci-all` 127/127 (128 before — `read-plugin-file-guard` is the one that died with its subject).

Closes #497


scripts/find-unused-service-members.ps1
```diff
@@ -0,0 +1,181 @@
...
+    $head = $line
+    $cut = ($head.IndexOf('{'), $head.IndexOf('=>') | Where-Object { $_ -ge 0 } | Measure-Object -Minimum).Minimum
+    if ($null -ne $cut) { $head = $head.Substring(0, $cut) }
+
+    # The parameter list is the first "(" an identifier or a generic argument list touches. A "(" preceded by
+    # anything else opens a tuple RETURN type, so skip it and its contents.
+    $paramAt = -1
+    for ($c = 0; $c -lt $head.Length; $c++) {
+        if ($head[$c] -ne '(') { continue }
+        if ($head.Substring(0, $c) -match '[A-Za-z0-9_>]\s*$') { $paramAt = $c; break }
```
Avick3110: **The tuple-return guard still misparses, and it hides eight members from the population.**

The test allows trailing whitespace, so `    public (int plugins, ...) Stats()` gives prefix `"    public "` — `[A-Za-z0-9_>]` matches the `c`, `\s*` eats the space, `$` closes. The tuple's paren is taken as the parameter list, `$decl` becomes `"    public "`, and the member is recorded under the name `public`.

Running the script on this repo as it stands:

```
114 public/internal members, 104 distinct names
...
public   method   285,1687,2073,7719   479
static   method   864,887,7379,7394    672
```

Those eight lines are `PapyrusSourceImportDirs`, `Stats`, `SetInstance`, `ClassParentsForDecompile`, `Ladder`, `Classify`, `ScriptOutputContract`, `SeqOutputContract` — every tuple-returning member in the class. They are collapsed under two names that count every `public`/`static` keyword in scope as a call, so none of them can ever be reported dead.

Dropping the `\s*` makes all eight appear — and `Stats` comes back with **Calls = 0** (its only callers are probes and tests, like `ForGuard`). That is a member the derived population was meant to surface and silently did not. The PR body's "the script's declaration parser was fixed in review" holds for the generic-parameter case; the tuple case is still broken, and the stated 114/101 counts overstate what was actually scanned.

```suggestion
        if ($head.Substring(0, $c) -match '[A-Za-z0-9_>]$') { $paramAt = $c; break }
```

src/housecarl-generator/BulkPrimitivesWave3Probe.cs
```diff
@@ -568,125 +568,80 @@ static void DiffArm(string dir)
...
         // filename is listed ahead of game Data, and must not decide the live plugin's provenance.
-        string dswFid = $"{dswFk.ID:X6}:{dswFk.ModKey.FileName}";
-        var rpfData = svc.ReadPluginFile(dataServedPath, dswFid, null, null, new[] { "BasicStats.Damage" }, 1, null, 10);
-        Check("read_plugin_file: a game-Data-served plugin listed BEHIND a disabled copy still reports enabled=true",
-              rpfData.Error is null && rpfData.Where == "direct path" && rpfData.Enabled);
-        var rpfDecoy = svc.ReadPluginFile(decoyPath, dswFid, null, null, new[] { "BasicStats.Damage" }, 1, null, 10);
-        Check("read_plugin_file: that disabled copy itself reports enabled=false",
-              rpfDecoy.Error is null && rpfDecoy.Where == "direct path" && !rpfDecoy.Enabled);
-        string uwFid = $"{uwFk.ID:X6}:{uwFk.ModKey.FileName}";
+        Check("source pole: a game-Data-served plugin listed BEHIND a disabled copy still resolves ACTIVE",
```
Avick3110, line 623: **These two ACTIVE-lane arms can no longer go red on the fact they were written for.**

`ResolvePoleArm` short-circuits before the locate ever runs:

```csharp
if (LooksLikePath(plugin) && ActiveNameForPath(view, plugin) is { } activeName) plugin = activeName;
if (view.ContainsPlugin(plugin))
    return (new PoleInfo(plugin, "active in the load order", ...), null);
```

`ActiveNameForPath` compares the given path against `view.PluginPath(name)`. So `Arm(dataServedPath)` returns `"active in the load order"` purely because the path equals the active order's own path for that plugin — `LocatePluginFileOnDisk` is never called.

That is exactly the fact the arm replaced. `rpfData.Enabled` was the locate's own flag, and it required the locate to skip the DISABLED `DataServedDecoy` folder listed ahead of game Data. If that ordering regressed and the decoy were treated as the serving copy, the old arm went red; the new one still says "active". Same for `Arm(replPath)` on line 594, which carried the second-consumer half of #269.

`Cause(decoyPath)` does not cover it — a decoy sitting in a switched-off folder renders "switched OFF" either way. `Cause(shadowPath)` covers the enabled-vs-enabled serving choice, but not the enabled-layer-vs-disabled-layer one.

The PR body enumerates its narrowings carefully; this one isn't among them. Either state it, or keep one arm on the locate directly (`svc.ProbeSourceArm` on a path that is not the active order's copy, or the mod= form) so the ordering still has a red.

src/housecarl-mcp-tests/RecordsPluginFileSourceTests.cs
```diff
@@ -47,6 +47,18 @@ public void AnOffOrderFileReadInJsonCarriesTheSameFieldTokenAsTheTextRender()
...
+    /// <summary>The FILENAME lane's not-active cause. A plugin addressed by NAME out of a switched-off mod names
+    /// the mod folder's state and its remedy, and says each once — the twin of <c>RecordsOffOrderPathTests.FactB4</c>,
+    /// whose path lane has no mod folder in its label to lean on.</summary>
+    [Fact]
+    public void APluginNamedOutOfASwitchedOffModSaysThatModFolderIsOff()
+    {
+        var r = OffOrderFields();
+        Served(r, "OUT-OF-LOAD-ORDER", "NOT active — that mod folder is switched OFF in MO2 — switch it on, then re-sort");
+        var subject = Assert.Single(r.Split('\n'), l => l.Contains("OUT-OF-LOAD-ORDER", StringComparison.Ordinal));
+        Assert.Equal(1, CountOf(subject, "switch it on"));
```
Avick3110, line 59: **The counting this test is offered as only carries half of it — and drops the half that belongs on this lane.**

The PR body says this test holds "the counting (mod and remedy stated once) that `FactB4` holds for the path lane". `FactB4` counts both:

```csharp
Assert.Equal(1, CountOf(subject, "'OldMod'"));
Assert.Equal(1, CountOf(subject, "switch it on"));
```

This one counts only the remedy. The mod-name count is the one that matters here: on the filename lane `loc.Where` is `mod 'OldMod' (DISABLED)` (`Mo2LoadOrder.CandidateFolders`), so the label already names the folder; on the path lane `Where` is the constant `"direct path"` and names none. The deleted probe arm said as much — "armed on the lane where the bug can actually appear".

Concretely: reword the layer-off cause to the phrasing the path lane already uses and the filename label becomes

```
OUT-OF-LOAD-ORDER (mod 'OldMod' (DISABLED); NOT active — it is provided by mod 'OldMod', which is switched OFF in MO2 — switch it on, then re-sort)
```

Both assertions here still pass, and `FactB4` is on the lane where the duplication cannot arise. One more line closes it:

```suggestion
        Assert.Equal(1, CountOf(subject, "switch it on"));
        Assert.Equal(1, CountOf(subject, "'OldMod'"));
```

scripts/find-unused-service-members.ps1
Avick3110: Folded — dropped the `\s*` as suggested. The corrected population on main is 114 members / 110 distinct names / 8 uncalled; the tuple fix surfaced all eight tuple-returning members, seven of which have shipped callers. `Stats` came back with zero, and is kept as a third seam (the probes and tests warm the lazy index through it and read the epoch and counters off it, ~100 call sites) — noted in the script header, and its doc comment no longer claims a stand-up/health check the product does not do. On the branch the script now reports 110 / 106 / 4: Dispose, ForGuard, Stats, WithSourceChainForGuard, each explained.

src/housecarl-generator/BulkPrimitivesWave3Probe.cs
Avick3110: Right — both arms now only test the path-identity rule, and their comments say so. The ordering keeps a red through a new arm on the locate itself: a second game-Data plugin (HcW3DataOff.esp) with a copy in the same disabled DataServedDecoy folder, addressed by path, on a name the order does not carry, so ResolvePoleArm cannot short-circuit. Confirmed by making JudgeServed take the first hit rather than the first enabled hit — that arm goes red (the copy reads SHADOWED); restored, ci-all is 127/127.

src/housecarl-mcp-tests/RecordsPluginFileSourceTests.cs
Avick3110: Added `Assert.Equal(1, CountOf(subject, \OldMod\))` — the filename lane is where the duplication can arise, since its label already names the folder. Red-checked by appending the mod name to the layer-off cause with the existing phrasing intact: the count fails 2 vs 1.

PR #520, opened 2026-09-04
records: read the named mod folder's copy under a {file, mod} source pole

### LINKED ISSUE #517: records: the mod-folder address form must read the named copy even when the filename is active
The `{"file","mod"}` source form on `housecarl_records` exists to read a specific on-disk copy of a plugin when several mod folders ship the same filename (ESP replacers) and MO2 picks one. Today the source pole resolves the active copy first, so when the filename is active, naming a different mod folder still reads the active copy. The form only does its job for a filename that is not active. PR #516 stated this narrowing rather than hiding it.

Aaron's ruling (2026-09-05): restore it. When both a filename and a mod folder are given, the copy in that folder is read, with the response saying it is not the active copy and which mod is. One line in `dev/projects/tool-surface-2.0/SPEC.md` §2.2 (Aaron edits), a test on the `records` lane with two enabled mods carrying one filename, its own PR.

Avick3110:
The `{"file", "mod"}` source form on `housecarl_records` exists to read a specific on-disk copy of a plugin when several mod folders ship the same filename and MO2 serves one of them. The source pole resolved the active copy first, so when the filename was active the form read the served copy whatever folder was named — it only did its job for a filename that was not active. The pole now runs the on-disk locate first for that form: the named folder's copy is read, and when it is not the copy the game loads the response carries the existing off-order label, which names the mod folder that provides the served copy. A folder that does not carry the filename is still refused, with a sentence that says the filename is active rather than claiming it is not. The plain-filename and direct-path forms are unchanged; a `file=` that is a path still addresses that path even with a `mod=` beside it.

One knock-on from the new state: an off-order subject can now carry a filename that IS active, and `versus="previous_provider"` is a position in the active touching stack that such a copy does not hold. It is refused instead of anchoring by name on the served copy.

Tests: two enabled mods carrying one filename with different Damage per copy, the higher-priority one served. Naming the shadowed mod reads its copy and says which mod is active; naming the serving mod reads the active copy on the active arm; a folder without the filename is refused; a path beside a `mod=` reads the path; `previous_provider` against the shadowed copy is refused.

Closes #517

### DIFF
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -2583,31 +2583,45 @@ public sealed class LoadOrderService : IDisposable
 
     /// <summary>Resolve a `records` source= pole against ONE captured view: active in the order, else located on disk
     /// across the whole install. A non-null error means it was found in neither place (naming both), or is ambiguous
-    /// across mod folders (naming them and the {file, mod} disambiguator).</summary>
+    /// across mod folders (naming them and the {file, mod} disambiguator).
+    /// <para>The {file, mod} form addresses ONE on-disk copy, which is the whole reason it exists: several mod
+    /// folders ship the same filename and MO2 serves one. So it does NOT short-circuit to the active copy of the
+    /// filename — the locate runs first, and the named copy resolves to the active arm only when it IS the copy the
+    /// game loads. A plain filename and a direct path are unchanged.</para></summary>
     (PoleInfo? Pole, string? Error) ResolvePoleArm(LoadOrderResolver.IndexView view, string plugin, string? mod)
     {
         // A pole addressed by path that IS the active order's file resolves back to its plugin name.
         if (LooksLikePath(plugin) && ActiveNameForPath(view, plugin) is { } activeName) plugin = activeName;
 
-        if (view.ContainsPlugin(plugin))
+        bool namesMod = !string.IsNullOrWhiteSpace(mod) && !LooksLikePath(plugin);
+        bool activeFilename = view.ContainsPlugin(plugin);
+        if (!namesMod && activeFilename)
             return (new PoleInfo(plugin, "active in the load order", InOrder: true, EpochCoversPole: true), null);
 
         string modsDir, dataDir, overwriteDir, profileDir;
         try { lock (_gate) { EnsurePathsDerived(); modsDir = _modsDir; dataDir = _dataDir; overwriteDir = _overwriteDir; profileDir = _profileDir; } }
         catch (Exception ex)
         {
-            return (null, $"source '{plugin}' is not active in the load order, and the MO2 roots couldn't be derived to search for it on disk: {ex.Message}");
+            return (null, activeFilename
+                ? $"source '{plugin}' names a mod folder, and the MO2 roots couldn't be derived to read that folder's copy: {ex.Message}"
+                : $"source '{plugin}' is not active in the load order, and the MO2 roots couldn't be derived to search for it on disk: {ex.Message}");
         }
         var comp = Mo2LoadOrder.ReadComposition(profileDir);
         var loc = LocatePluginFileOnDisk(comp, modsDir, dataDir, overwriteDir, plugin, mod);
         if (loc.Error is not null)
-            // A pole found in neither place names both places searched.
-            return (null, $"source '{plugin}' resolves in NEITHER place the one-pole rule searches: it is not ACTIVE in " +
-                          $"the load order ({view.PluginCount} plugins), and on disk {loc.Error}");
+            // A pole found in neither place names both places searched. When the filename IS active, the named mod
+            // folder is the only place searched, so the sentence says that instead of claiming it is not active.
+            return (null, activeFilename
+                ? $"source '{plugin}': {loc.Error} The filename IS active in the load order — drop mod= to read the copy the game loads."
+                : $"source '{plugin}' resolves in NEITHER place the one-pole rule searches: it is not ACTIVE in " +
+                  $"the load order ({view.PluginCount} plugins), and on disk {loc.Error}");
         if (loc.Ambiguous is not null)
             return (null, $"source '{plugin}' is not active in the load order and its filename is provided by SEVERAL mod " +
                           $"folders on disk: {string.Join(", ", loc.Ambiguous.Select(h => $"'{h.Where}'"))} — disambiguate " +
                           "with source={\"file\": \"" + plugin + "\", \"mod\": \"<mod folder>\"}.");
+        // The named copy IS the copy the game loads, so it is read out of the order like any other active pole.
+        if (activeFilename && loc.Enabled)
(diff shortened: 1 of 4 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2583,31 +2583,48 @@ public sealed record PoleInfo(string Plugin, string Where, bool InOrder, bool Ep
...
+            return (null, activeFilename
+                ? $"source '{plugin}': {loc.Error} The filename IS active in the load order — drop mod= to read the copy the game loads."
+                : $"source '{plugin}' resolves in NEITHER place the one-pole rule searches: it is not ACTIVE in " +
+                  $"the load order ({view.PluginCount} plugins), and on disk {loc.Error}");
         if (loc.Ambiguous is not null)
             return (null, $"source '{plugin}' is not active in the load order and its filename is provided by SEVERAL mod " +
                           $"folders on disk: {string.Join(", ", loc.Ambiguous.Select(h => $"'{h.Where}'"))} — disambiguate " +
                           "with source={\"file\": \"" + plugin + "\", \"mod\": \"<mod folder>\"}.");
+        // The named copy IS the copy the game loads, so it is read out of the order like any other active pole.
+        if (activeFilename && loc.Enabled)
```
Avick3110, line 2643: **form=tree now reports a real difference as "identical to the reference".**

`TreeBatch` decides which provider node IS the reference pole by plugin-NAME equality alone (`LoadOrderService.cs:3322-3323`: `string.Equals(node.Plugin, refPlugin, OrdinalIgnoreCase)`). That was safe until this line, because an off-order named pole could never carry a filename that is also active — a bare filename that was active short-circuited here, and a path pole keeps its path as `Plugin`. Now it can.

Repro on this branch's own `ShadowedCopyWorld` fixture (added a scratch test, ran it, removed it):

```
records formids=[000800:HcSMaster.esm] versus={"file":"HcSReplacer.esp","mod":"SLoser"} project.form=tree

  2 plugin(s) touch this record (load order, winner last):
    1. HcSMaster.esm
    2. HcSReplacer.esp  (winner)
  diff (field deltas vs HcSReplacer.esp; identical fields omitted; ...):
    HcSMaster.esm: BasicStats.Damage=10 (HcSReplacer.esp 44)
```

The active `HcSReplacer.esp` (Damage 99) is flagged `IsReference: true` with empty `Deltas`, which per `TreeNodeDelta`'s own doc means "genuinely identical to the reference" — but the reference is SLoser's copy at Damage 44. The 99-vs-44 delta is silently dropped. The delta form on the same two poles reports it correctly, so the two forms now disagree.

The node-is-the-reference test needs the reference's arm as well as its name — e.g. only treat a node as the reference when the reference pole resolved in-order (`r.Pole!.InOrder`), since an off-order pole is never one of the active providers.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2832,8 +2849,14 @@ public IReadOnlyList<DeltaRow> DeltaBatch(
             var s = sReader(fk, null);
             if (s.Error is not null) { rows.Add(new DeltaRow(fk.ToString(), s.Pole, null, null, null, null, "subject: " + s.Error)); continue; }
             // previous_provider is measured from the SUBJECT, so hand the reference reader the subject's resolved
-            // plugin for this record and it anchors on the right stack position.
-            var r = rReader(fk, s.Pole!.Plugin);
+            // plugin for this record and it anchors on the right stack position. An off-order subject holds no
+            // position in that stack — and its filename can be active as a DIFFERENT file — so it is refused rather
+            // than anchored by name on whatever the order serves.
+            var r = reference.Kind == PoleKind.PreviousProvider && !s.Pole!.InOrder
```
Avick3110: The refusal is decided per record, but what it tests — `reference.Kind` and the subject arm's `InOrder` — is uniform for the whole batch. Each row still pays `sReader(fk, null)` first, so on the off-order arm the file's overlay is opened and swept and every requested record is deep-read, then thrown away for a refusal that was knowable before the loop.

On the list lane that is a handful of records. On the scan-driven delta over an off-order file it is the whole selection: `RecordsTools.OffOrderScan` sets `offLimit = int.MaxValue` for a comparison form, so a `source={"file","mod"}` + `versus="previous_provider"` scan reads every match off disk and then errors on every row.

Every other uniform-arm problem in this method comes out as a single whole-call `refusal` next to `sErr`/`rErr`. This one reads the same way to a caller but costs the whole batch.

src/housecarl-mcp/LoadOrderService.cs
Avick3110, line 2643: Knock-on for the delta render: a subject and a reference can now be two different files sharing one filename, and the diff-line grammar names both by that filename only. Same fixture, `source="HcSReplacer.esp"` vs `versus={"file":"HcSReplacer.esp","mod":"SLoser"}`:

```
  subject:   HcSReplacer.esp [active in the load order]
  reference: HcSReplacer.esp [OUT-OF-LOAD-ORDER (mod 'SLoser'; ...)]
  1 difference — each line: HcSReplacer.esp's value (reference = HcSReplacer.esp):
    - BasicStats.Damage=99 (HcSReplacer.esp 44)
```

The answer is correct, but "HcSReplacer.esp's value (reference = HcSReplacer.esp)" and `=99 (HcSReplacer.esp 44)` carry no way to tell the two sides apart from the delta line itself — the pole lines above are the only disambiguator. Before this change the two poles could not collide on a name while being different files. Worth qualifying the label with the mod folder (or `(off-order)`) when the two poles resolve to the same filename on different arms.

Avick3110: Real — fixed. A node is the reference only when the reference pole resolved in-order, so an off-order pole is never treated as an active provider; the active copy now shows its delta. Covered by ATreeAgainstTheShadowedCopyShowsTheActiveCopysDeltaRatherThanCallingItTheReference on ShadowedCopyWorld (red before, green after).

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Agreed — hoisted. MakePoleReader now reports the resolved off-order arm, and DeltaBatch refuses the whole call beside sErr/rErr before the loop, so no record is read. PreviousProviderAgainstAnOffOrderSubjectIsRefusedOncePerCallNotOncePerRecord holds it to one sentence over a two-record batch.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Fixed. The off-order pole carries the locate's own layer label as a fact (PoleInfo.Layer, no string parsing), and DiffPole.LabelVersus qualifies a side only when the other arm carries the same filename — so the delta line reads HcSReplacer.esp (mod 'SLoser'). Applied to the tree render too, since the tree fix creates the same collision. Test: ADeltaBetweenTwoCopiesOfOneFilenameQualifiesTheOffOrderSideOnTheDeltaLine.

PR #243, opened 2026-07-18
feat(nif): mesh_paths batch on housecarl_nif_inspect — one call for a whole facegen sweep

Avick3110:
Fixes #229

## What

`housecarl_nif_inspect`'s `mesh_path` becomes **`mesh_paths`** — an `asset_status`-style array (a single path is simply a batch of one). This closes the last per-file loop in the load-order-wide dark-face scan: the flagged subset of a whole-load-order sweep is now **one call**, not hundreds.

Of the issue's two ideal shapes, this takes the first (array on the existing tool) — it mirrors the `asset_status` precedent exactly and follows the bulk-primitives convention of extending existing tools rather than minting `bulk_*` twins.

## How

- **Service** ([LoadOrderService.cs](../blob/claude/bulk-nif-inspect/src/housecarl-mcp/LoadOrderService.cs)): **one `Assets.Capture()` under the gate pins the whole batch** (this is the real win — the load order resolves once, not per mesh); per-path resolve → byte read → NIF parse runs outside the gate on the pinned handle-free view. Every per-path failure (empty/invalid path, ABSENT, a `mod=` that doesn't provide it, unreadable bytes, a parse refusal) is a **named per-path error that never aborts the rest** (Q3). Batch-level caveats (BSA read failures, discovery warnings, profile) move to a new `NifInspectBatchData`; `NifInspectData` slims to per-path fields.
- **Tool + wire** ([NifTools.cs](../blob/claude/bulk-nif-inspect/src/housecarl-mcp/NifTools.cs)): results render in **input order**, one block per mesh; the Q3 alarms render **once, first**, so a long batch can't truncate them away; an over-cap output cuts with the **omitted-mesh count named** — never silent. `sections=` / `mod=` / `max_chars` apply batch-wide, per the existing bulk conventions.
- **Probe**: new `nif-inspect-batch-guard` locks the four batch contracts (input order; per-path error isolation; one-shot alarms-first; explicit cut). All three teeth verified **mutation-RED at authoring** (abort-on-error → arm 2 fails; alarms-per-mesh → arm 3 fails; silent cut → arm 4 fails). `NifServiceGuardProbe`'s render arms now feed the wire a batch of one — all its arms unchanged and green, including the Lucien corpus smoke.
[...]

PR #521, opened 2026-09-04
Render both batch tools through one shared skeleton

### LINKED ISSUE #414: [Bug] no shared batch-render skeleton — and the PR #243 first-item fix landed in NifWire only, so asset_status can cut its first path
Migrated from the retired `dev/BACKLOG.md` (advisor sweep, Aaron-go 2026-08-22); re-verified at `main` @ `a783b2f` — **and the re-verify found a live defect the backlog item didn't have**, which is why this files as [Bug] not cleanup.

`AssetWire.Render` (`AssetTools.cs:56-79`) and `NifWire.Render` (`NifTools.cs:277-307`) are two hand-rolled copies of the same four-part batch shape (header count line, alarms-first-once, cap-checked per-item loop, omitted-count cut). Measured drift:
- **Loop guard:** `if (sb.Length >= cap)` (`AssetTools.cs:69`) vs `if (shown > 0 && sb.Length >= cap)` (`NifTools.cs:297`) — the first-item-always-renders fix from PR #243 exists only in NifWire. **A tight `max_chars` on `asset_status` can render zero items where `nif_inspect` guarantees one.**
- Cut marker: ASCII `... [N more path(s) omitted` vs U+2026 `… [N more mesh(es) omitted`, different indent.
- The alarms helpers are themselves duplicated (`AppendReadFailures`/`AppendDiscoveryWarnings` at `AssetTools.cs:132,147` AND `NifTools.cs:666,678`).

This is the twin-drift class (#337's lesson) on the batch surface, caught mid-drift. **W4 is about to add a third copy** (#246's S2 glob is a batch conversion) — extract the shared skeleton + one shared guard BEFORE that conversion, and the first-item fix reaches asset_status by construction.

**Routing:** row 7/W4, sequenced ahead of #246's conversion. Milestoned.

### LINKED ISSUE #525: [Bug] the batch alarms block has no first-line exemption, so a tight max_chars can name zero read failures while a per-path hedge points at them
Found by the review pass on PR #521 (the shared batch render). `BatchRender.AppendReadFailures` / `AppendDiscoveryWarnings` (shared by `asset_status` and `nif_inspect`) cut the alarm list on the cap with no first-line exemption, unlike the item loop, which always renders its first item. At a very tight `max_chars` the read-failure block can name zero archives while a per-path hedge below still points at it. Pre-existing in both tools before #521; the refactor kept the behaviour rather than fold a change in.

Fix shape: the alarm block gets the same first-line guard as the item loop, so at least one failure is always named, with the omitted count. One test under a cap tighter than the first alarm line, red first. Unmilestoned; Aaron promotes if it bites.

Avick3110:
`AssetWire.Render` and `NifWire.Render` were two hand-rolled copies of the same batch shape — header count line, alarms first and once, cap-checked per-item loop, omitted-count cut — and they had drifted. The first-item-always-renders guard from PR #243 existed only in `NifWire`, so a tight `max_chars` on `asset_status` could render zero items where `nif_inspect` guaranteed one; the cut markers and the `AppendReadFailures` / `AppendDiscoveryWarnings` helpers were duplicated too. Both tools now render through one `BatchRender` skeleton (`src/housecarl-mcp/BatchRender.cs`): the caller supplies a header, an alarms block, an item renderer, a cap, and the noun the omitted-count line counts, and the skeleton owns the first-item guard and the single cut marker. `asset_status` gets the fix by construction, and the W4 `place` tool lands as the third caller without a fourth copy. The only behaviour changes are that fix and the marker unification (`asset_status` now cuts with the same `  … [` marker `nif_inspect` uses); `NifTools`'s own detail cut delegates to the same marker rather than repeating it.

The alarm list had the same gap and is fixed here too: `AppendLines` now carries the same `shown > 0` exemption as the item loop, so at a very tight `max_chars` the read-failure block always names at least one archive before the omitted-count cut, instead of naming zero while a per-path hedge points at it.

New tests in `src/housecarl-mcp-tests/BatchRenderCapTests.cs` (tier `unit`): `AssetStatusUnderATightCapStillRendersItsFirstPath` is red on the unfixed renderer (verified by reverting both tool files — it cuts to zero paths) and green now; `BothBatchRendersKeepTheirFirstItemAndCutWithTheSameMarker` pins the parity the skeleton buys; `AReadFailureListUnderATightCapStillRendersItsFirstEntry` covers the alarm-list fix, and was red before it. `nif_inspect`'s own first-mesh contract stays where it already lives, arm 6 of `nif-inspect-batch-guard`. No probe pinned the old ASCII marker, so no probe needed updating.

Verified: `dotnet build -c Release` 0 errors, `dotnet test --filter tier!=bridge` 1042 passed / 0 failed, `ci-all` all pass.

Closes #414
Closes #525

### DIFF
--- a/src/housecarl-mcp/BatchRender.cs
+++ b/src/housecarl-mcp/BatchRender.cs
@@ -2,7 +2,7 @@ using System.Text;
 
 namespace HousecarlMcp;
 
-/// <summary>The one batch-render skeleton every multi-item tool renders through: a header count line, the batch-level
+/// <summary>The shared batch-render skeleton behind housecarl_asset_status and housecarl_nif_inspect: a header count line, the batch-level
 /// alarms once and first (so a long batch cannot truncate them away), a cap-checked per-item loop whose FIRST item
 /// always renders its core answer, and an explicit omitted-count cut. max_chars bounds the batch tail, never a
 /// single-item call's own answer, and a cut is always named — never a silent truncation. Callers supply the header,
@@ -58,7 +58,7 @@ static class BatchRender
         if (failures.Count == 0) return;
         sb.Append("\n[!] ").Append(failures.Count).Append(" archive(s) could NOT be read this build — ")
           .Append(subjectPhrase).Append(" present only in these may read as ABSENT below:\n");
-        AppendLines(sb, failures, cap);
+        AppendLines(sb, failures, "archive(s)", cap);
     }
 
     /// <summary>Archive-discovery warnings, e.g. a Skyrim.ini whose [Archive] base-archive list could not be found, so
@@ -67,16 +67,16 @@ static class BatchRender
     {
         if (warnings.Count == 0) return;
         sb.Append("\n[!] discovery (").Append(warnings.Count).Append("):\n");
-        AppendLines(sb, warnings, cap);
+        AppendLines(sb, warnings, "warning(s)", cap);
     }
 
     /// <summary>A capped bullet list inside an alarm block, cut with the same named marker.</summary>
-    static void AppendLines(StringBuilder sb, IReadOnlyList<string> lines, int cap)
+    static void AppendLines(StringBuilder sb, IReadOnlyList<string> lines, string itemNoun, int cap)
     {
         int shown = 0;
         foreach (var line in lines)
         {
-            if (sb.Length >= cap) { AppendCut(sb, lines.Count - shown, "", cap); break; }
+            if (sb.Length >= cap) { AppendCut(sb, lines.Count - shown, itemNoun, cap); break; }
             sb.Append("  - ").Append(line).Append('\n'); shown++;
         }
     }
--- a/src/housecarl-mcp/BatchRender.cs
+++ b/src/housecarl-mcp/BatchRender.cs
@@ -76,7 +76,8 @@ static class BatchRender
         int shown = 0;
         foreach (var line in lines)
         {
-            if (sb.Length >= cap) { AppendCut(sb, lines.Count - shown, itemNoun, cap); break; }
+            // shown > 0: the first line always renders even when the header and the alarm heading alone exhausted the cap.
+            if (shown > 0 && sb.Length >= cap) { AppendCut(sb, lines.Count - shown, itemNoun, cap); break; }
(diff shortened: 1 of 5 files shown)


src/housecarl-mcp/BatchRender.cs
```diff
@@ -0,0 +1,83 @@
...
+        AppendLines(sb, warnings, "warning(s)", cap);
+    }
+
+    /// <summary>A capped bullet list inside an alarm block, cut with the same named marker.</summary>
+    static void AppendLines(StringBuilder sb, IReadOnlyList<string> lines, string itemNoun, int cap)
+    {
+        int shown = 0;
+        foreach (var line in lines)
+        {
+            if (sb.Length >= cap) { AppendCut(sb, lines.Count - shown, itemNoun, cap); break; }
```
Avick3110: The alarm list has no first-line exemption, so the block the per-path hedge points at can name zero archives.

`Render` at line 31 guards its item loop with `shown > 0 &&`; this loop does not. With one BSA read failure at `max_chars=100`, the header plus the alarm's own heading already exceed the cap, so this cuts before its first entry — verified by rendering both tools:

```
[!] 1 archive(s) could NOT be read this build — an asset present only in these may read as ABSENT below:
  … [1 more archive(s) omitted at max_chars=100; raise max_chars to see all]

meshes/a/first.nif
  ABSENT — no active mod or BSA provides this path
  [!] but an archive failed to read this build (see the read-failure note above), so "absent" may be incomplete
```

The hedge points at a note that names nothing. I know the PR body calls this out as pre-existing and defers it to its own issue — raising it because the deferral was written when it was two duplicated loops; now that both tools share this one method, it is the same one-token fix the item loop already carries, and the class doc's "a cut is always named" reads oddly next to a named cut of everything.

```suggestion
            if (shown > 0 && sb.Length >= cap) { AppendCut(sb, lines.Count - shown, itemNoun, cap); break; }
```

Avick3110: Real — fixed here: the alarm list's AppendLines now has the same shown > 0 first-line exemption as the item loop, with a new red-first test in BatchRenderCapTests.cs, and the PR now closes #525 too.

PR #523, opened 2026-09-04
Test the bsa_repack header read-back against a real archive

### LINKED ISSUE #519: bsa_repack: cover the header read-back against a real archive in the test suite
PR #514 made `bsa_repack` count the source, read the produced archive's header count back, and refuse on a mismatch. The suite has no BSArch, so the tests cover the counting and the refusal wording, and the read-back itself was verified by hand against BSArch 0.9c on Aaron's install. Nothing in CI produces a real archive and reads it back, so a later break of the read-back stays green.

Aaron's ruling (2026-09-05): close it. Either a packer seam on `BsaArchive.Pack` so a test can substitute a packer that writes a known archive, or the in-memory BSA builder lifted out of `BsaExtractProbe` into a shared test helper. One test that packs, reads back, and goes red when the read-back is broken. Small PR in W4.

### LINKED ISSUE #522: bsa_repack: a fast pack can fail with no reason on the scratch mtime gate
`BsaArchive.Pack` reads `baselineUtc = DateTime.UtcNow` before running the packer and then requires `File.GetLastWriteTimeUtc(tmp) >= baselineUtc` before it will place the archive. `DateTime.UtcNow` comes from the precise system clock; NTFS stamps a last-write time from the coarse ~15.6ms tick, so a file written immediately after the baseline routinely carries an *earlier* mtime. A review pass on #519 measured 1211 of 2000 immediate writes landing below a baseline read microseconds earlier.

When that happens the pack deletes a perfectly good scratch and returns `Success=false` with an empty `Raw`, `RunError=null` and `CountError=null` — a failure with no sentence saying what went wrong. Real BSArch usually clears the window because launching a process costs more than a tick, but a fast pack of a small source folder sits inside it.

Fix at the source: give the baseline a tolerance (`DateTime.UtcNow.AddSeconds(-2)`), or capture the scratch's pre-run mtime and require strictly newer. The provenance the gate exists for survives either way.

Surfaced by #519, whose substitute packer has to stamp the scratch's write time itself to get past the gate; that stamp comes out once this is fixed.

Avick3110:
PR #514 made `bsa_repack` read the produced archive's own header count back and refuse when it disagrees with the source, but the suite has no BSArch, so nothing in CI ever produced an archive and counted it — a later break of the read-back stayed green. `BsaArchive.Pack` now takes an optional packer delegate, defaulting to the BSArch shell it already ran, so a test can substitute a packer that writes a known archive; the in-memory BSA builder that `BsaExtractProbe` hand-rolls moves to a shared `BsaBuilder` the probe and the tests both use, so the archive the read-back parses is a real one rather than a stub header. Two tests: one packs a three-file source through the seam, reads back 3 and reports success; one makes the archive claim 2 and asserts the refusal names both numbers and leaves the existing archive untouched.

Both shapes the issue offered turned out to be needed rather than alternatives — the seam is the only way to drive `Pack` without BSArch, and a substituted packer needs something real to write — so the smaller move was to take the seam and lift the builder instead of duplicating sixty lines of byte layout into the test project. The builder is a verbatim move; the probe keeps working through it.

Two review findings and one issue folded on top:

- The packer tuple now carries the run's **exit code**. It was dropped, so BSArch aborting partway while leaving a scratch whose header count happened to match the source would ship that scratch over the user's archive and report success. A non-zero exit is now a failed pack whatever it left on disk, refused in one sentence naming the code and what BSArch printed, with nothing committed over the target.
- `Pack`'s summary now says the write runs through `packer`, defaulting to the BSArch shell, so the "via BSArch" paragraph above it reads as that default.
- **#522**: the scratch freshness gate compared an NTFS write stamp against a baseline read from the precise clock, so a pack finishing inside one filesystem tick read as stale and failed with an empty message. The scratch is cleared before the run and a stuck one already refuses, so existence plus non-empty plus a clean exit now proves provenance without a timestamp. The test packer's write-time workaround comes out, so the tests exercise the real gate.

Proved red by making the read-back's `ReadBsaHeader(tmp)` return null: `PackReadsTheProducedArchiveCountBack` failed on `Packed` (expected 3, got null) and `PackRefusesAnArchiveThatCountsShort` failed on the refusal being gone. The two new tests were proved red against the old gate — with the exit check off and the mtime compare back, `PackRefusesAScratchFromANonZeroExit` and `PackAcceptsAScratchWrittenImmediately` (20 immediate writes in a loop) both failed. All green after. Full suite 1043 passed, `ci-all` 127/127. `BsaContractProbe`'s stub-BSArch check moved with the behaviour: a failing stub now returns a refusal naming the exit code rather than a bare Ran-but-not-Success.

Closes #519
Closes #522

### DIFF
--- a/src/housecarl-core/BsaArchive.cs
+++ b/src/housecarl-core/BsaArchive.cs
@@ -212,12 +212,11 @@ public static class BsaArchive
     /// <summary>Pack <paramref name="srcFolder"/> into a .bsa at <paramref name="archive"/> with the given format flag
     /// (e.g. "-sse") and optional compression, via BSArch (Mutagen has no BSA writer). NON-DESTRUCTIVE:
     /// an existing archive at the target is NEVER overwritten unless this run successfully packs a new one — BSArch writes
-    /// to a houseCARL-internal temp beside the target, and only a clean pack THIS RUN (temp exists, non-empty, AND written
-    /// at/after the run's mtime baseline) is moved over the target; a stale scratch from a previous run that cannot be
-    /// removed REFUSES up front (nothing runs, the prior .bsa untouched), and any failure (BSArch error, non-zero exit,
-    /// timeout, empty output, stale-mtime scratch) deletes the temp and leaves the prior .bsa untouched. The mtime gate
-    /// assumes an NTFS-class timestamp resolution — on a FAT-class target a same-second pack could read as stale and fail
-    /// LOUD (never falsely succeed). The source is enumerated first and the produced archive's own header count is checked against it
+    /// to a houseCARL-internal temp beside the target, and only a clean pack THIS RUN (the scratch is cleared before the
+    /// run, so the temp existing, non-empty, after a zero-exit run is this run's) is moved over the target; a stale
+    /// scratch from a previous run that cannot be removed REFUSES up front (nothing runs, the prior .bsa untouched), and
+    /// any failure (BSArch error, non-zero exit, timeout, empty
+    /// output) deletes the temp and leaves the prior .bsa untouched. The source is enumerated first and the produced archive's own header count is checked against it
     /// before the move, so a short pack refuses instead of reporting success; files at the source ROOT are counted apart
     /// because BSArch drops them, and a source that cannot be fully enumerated packs unverified rather than refusing. NOTE the caller must surface BSArch's caveat: a COMPRESSED archive breaks any
     /// sounds/voices it contains. The write itself runs through <paramref name="packer"/>, which defaults to the BSArch
@@ -246,8 +245,6 @@ public static class BsaArchive
         try { scan = ScanPackSource(srcFolder); }
         catch { scan = null; }
 
-        var baselineUtc = DateTime.UtcNow;
-
         var run = (packer ?? ShellBsArch)(bsarchExe, srcFolder, tmp, formatFlag, compress, timeoutMs);
 
         // A non-zero exit is a failed pack whatever it left behind: BSArch can abort partway and still leave a scratch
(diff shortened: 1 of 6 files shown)


src/housecarl-core/BsaArchive.cs
```diff
@@ -39,6 +39,13 @@ public sealed record BsaPackResult(
     public bool Ran => RunError is null;
 }
 
+/// <summary>How a pack writes the archive: produce <paramref name="tmpArchive"/> from <paramref name="srcFolder"/> and
+/// report what the run said, with <c>runError</c> non-null when it never really ran. <see cref="BsaArchive.Pack"/>
+/// defaults to the BSArch shell; a test substitutes a packer that writes a known archive, so the count read-back around
+/// it can be exercised without BSArch.</summary>
+public delegate (string stdout, string stderr, string? runError) BsaPacker(
```
Avick3110: The packer tuple carries only `(stdout, stderr, runError)`, so BSArch's exit code — which `Run` still computes and `ShellBsArch` now discards — is structurally unavailable to `Pack`.

Concrete case: BSArch aborts partway (exit 1) after writing a scratch that happens to declare the same file count the source scan produced. `packed` is true (runError null, non-empty, fresh mtime), the header cross-check agrees, and `Pack` commits the bad archive over the user's existing `.bsa` and reports success — BSArch's error text only in `Raw`.

Pre-existing (`Pack` never read `run.exit`), so nothing regressed here. Flagging it because the seam turns the omission into a public contract: fixing it later means changing the delegate signature, not just `Pack`. Worth carrying `exit` through now while the delegate is new — and the seam would let a test drive exactly that case.

src/housecarl-core/BsaArchive.cs
```diff
@@ -213,7 +220,7 @@ static bool SameOnDisk(string path, byte[] body)
     /// before the move, so a short pack refuses instead of reporting success; files at the source ROOT are counted apart
     /// because BSArch drops them, and a source that cannot be fully enumerated packs unverified rather than refusing. NOTE the caller must surface BSArch's caveat: a COMPRESSED archive breaks any
     /// sounds/voices it contains.</summary>
-    public static BsaPackResult Pack(string bsarchExe, string srcFolder, string archive, string formatFlag, bool compress, int timeoutMs = 600_000)
+    public static BsaPackResult Pack(string bsarchExe, string srcFolder, string archive, string formatFlag, bool compress, int timeoutMs = 600_000, BsaPacker? packer = null)
```
Avick3110, line 225: `Pack` gained a public `packer` parameter but the XML summary above is unchanged: it still says unconditionally "via BSArch" and "BSArch writes to a houseCARL-internal temp", and never says what passing a packer does or that null means the BSArch shell.

In this codebase those summaries are the contract, so the next person to touch the seam has to read the body to learn that the whole "via BSArch" paragraph is conditional. A sentence at the end — the pack runs through `packer`, defaulting to the BSArch shell — would close it.

src/housecarl-core/BsaArchive.cs
Avick3110: Folded: the delegate now carries `exit`, and a non-zero exit is a failed pack whatever it left on disk — refused in one sentence naming the code and BSArch's stderr, with a test driving exactly that case through the seam.

src/housecarl-core/BsaArchive.cs
Avick3110: Folded: the summary now ends by saying the write runs through `packer`, defaulting to the BSArch shell, so everything above about BSArch reads as that default.

PR #524, opened 2026-09-04
Refuse a runtime FormID at every write door

### LINKED ISSUE #518: Write verbs refuse a runtime FormID and give the plugin form to use
PR #515 accepts runtime FormIDs (`FExxxYYY` / `XX######`) at every FormID door, including write verbs. A runtime FormID in a write resolves at parse time, before the write gate refreshes the load order, so a reorder inside the call could land the write on the record the old numbering named. Plugin-form FormIDs cannot go wrong this way.

Aaron's ruling (2026-09-05): runtime FormIDs are read-only. Every write verb (`apply`, `create`, `remove`, `forward`, `copy`, and any other that takes a FormID as the target) refuses a runtime FormID in one plain sentence that gives the translated `XXXXXX:Plugin.esp` form to use, so the user pastes it back. Reads are unchanged. SPEC §2.2 updated in place. One test per write door, red first. Its own PR.

Avick3110:
A runtime FormID (`FExxxYYY` / `XX######`) names a slot in the load order as it stands, not a record, so a re-sort between the parse and the write could land a write on whatever the old numbering pointed at. Aaron's ruling is that they are read-only. The FormID door gains a write mode: it still translates the token, then refuses with the `XXXXXX:Plugin.esp` form to paste back in its place. `apply`, `remove`, `forward`, `copy`, `copy_npc_appearance` and both `place_asset` doors open the write door instead of the read one; `create`'s `parent=` also takes an EditorID, so it goes through a runtime-only check at the same door rather than a parse. Reads are unchanged, and the refusal is a sentence of its own rather than being wrapped in the callers' "bad FormID … expected" framing, which called a well-formed token malformed. `apply`'s tool description no longer advertises the runtime form on `ops[].formid` and `from=`.

Tests: one per write door in `RuntimeFormIdTests`, each giving the light plugin's runtime FormID, asserting the refusal names the translated plugin form and that no patch folder was written, plus one asserting the same token still reads. All eight write-door tests were red on the unfixed code (they got as far as the record's own type/field errors, so the token had been accepted); the read test passed throughout.

Verified: build clean, `dotnet test --filter "tier!=bridge"` 1048/1048, `ci-all` 127/127.

Closes #518

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -35,7 +35,7 @@ saying it sets an expectation their install may contradict. Say what is known, a
   off-order `source=` is refused once for the call, before any record is read, instead of once per record.
 
 - **A FormID from the game, a log or a crash log now works as an input, and comes back as an output.**
-  Wherever a parameter holds nothing but FormIDs, houseCARL now also takes the runtime form the game, the
+  Wherever a FormID is read, houseCARL now also takes the runtime form the game, the
   console, Papyrus logs, SKSE logs and crash logs print: eight hex digits and no plugin name — `FExxxYYY` for a light plugin,
   `XX######` for a full one, with or without a leading `0x`. It is resolved against the load order as it
   stands at the call (the light index moves whenever the order does, so nothing is cached), and the answer
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -17,6 +17,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
   answer instead of cutting to none.** `housecarl_nif_inspect` already did; both tools now render through one batch
   skeleton, so the first item always answers and the omitted-count cut reads the same in both.
 
+- **A runtime FormID (`FExxxYYY` / `XX######`) is now read-only: the write verbs refuse one and hand back the
+  `XXXXXX:Plugin.esp` form to use in its place.** A runtime FormID names a slot in the load order as it stands, so
+  a re-sort between the parse and the write would point the same eight digits at a different record; the plugin form
+  cannot move that way. The refusal names the translated form, so it can be pasted straight back into the call.
+  Reading by a runtime FormID is unchanged, and the doors that refuse are the ones the write verbs parse their
+  targets through — see `FormIdDoor.ForWrite`.
+
 - **`housecarl_records` with `source={"file", "mod"}` now reads the copy in the named mod folder, even when that
   filename is active in the load order.** Several mod folders can ship the same plugin filename (an ESP replacer),
   and MO2 serves one of them; naming a folder addresses that folder's file. When the named copy is not the one the
--- a/src/housecarl-mcp/FormIdDoor.cs
+++ b/src/housecarl-mcp/FormIdDoor.cs
@@ -62,6 +62,9 @@ internal sealed class FormIdDoor
         if (!RuntimeFormId.TryParse(raw, out _)) return null;
         try { Parse(raw); return null; }
         catch (WriteRefusal ex) { return ex.Message; }
+        // A well-formed runtime token the index cannot translate (an FF dynamic form, an unoccupied index) is bad
+        // input, not an internal fault — hand its sentence back as this record's problem.
+        catch (FormatException ex) { return ex.Message; }
     }
 
     /// <summary>What to report for a token this door threw on. A write refusal is a WELL-FORMED token being
--- a/src/housecarl-mcp/FormIdDoor.cs
+++ b/src/housecarl-mcp/FormIdDoor.cs
@@ -15,9 +15,11 @@ namespace HousecarlMcp;
 internal sealed class FormIdDoor
 {
     readonly LoadOrderService? _svc;
+    readonly bool _write;
     LoadOrderResolver.IndexView? _view;
 
-    FormIdDoor(LoadOrderService? svc, LoadOrderResolver.IndexView? view) { _svc = svc; _view = view; }
+    FormIdDoor(LoadOrderService? svc, LoadOrderResolver.IndexView? view, bool write = false)
+    { _svc = svc; _view = view; _write = write; }
 
     /// <summary>A door pinned to a build the caller already captured, so its FormIDs and its records describe the
     /// same index.</summary>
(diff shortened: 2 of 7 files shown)


src/housecarl-mcp/FormIdDoor.cs
```diff
@@ -26,17 +28,49 @@ internal sealed class FormIdDoor
...
+    }
+
+    /// <summary>Refuse a runtime FormID in a slot that may hold something OTHER than a FormID (a create's
+    /// <c>parent=</c> takes an EditorID too), returning the sentence or null. Anything the door cannot recognise as
+    /// a runtime FormID is left to the caller's own parser.</summary>
+    public string? RuntimeRefusal(string? raw)
+    {
+        if (!RuntimeFormId.TryParse(raw, out _)) return null;
+        try { Parse(raw); return null; }
+        catch (WriteRefusal ex) { return ex.Message; }
```
Avick3110, line 64: `RuntimeRefusal` catches only `WriteRefusal`, but `Parse` can throw a plain `FormatException` before it ever reaches the write branch — `RuntimeToFormKey` throws for an `FF` dynamic form, for an index no active plugin occupies, and for the unknown-kind case.

`create` with `parent="FF001234"`: `RuntimeFormId.TryParse` accepts the token, `Parse` is called, `ParseFormId` throws `FormatException`, and this catch does not match. It escapes `BuildCreateSpec` and `CreateRecordsBatch` to `Guard.Tool`, which answers `error: … failed unexpectedly — FormatException: … This is an internal houseCARL failure (the arguments bound fine), not bad input. Retry once…` — bad input reported as an internal fault, and the whole batch aborts instead of collecting a per-record problem. Before this PR `parent` was never parsed as a FormID here, so it is a regression on that path.

```suggestion
        catch (FormatException ex) { return ex.Message; }
```
(That also hands the translation failure back as a named per-record problem rather than an internal error.)

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6483,10 +6487,10 @@ public NpcCopyOutcome CopyNpcAppearance(
         // ---- argument shape: exactly one target mode ----
         // ONE door for both tokens: two doors would each capture their own build, and a re-sort between them
         // would read the donor off one order and the target off another.
-        var door = OpenFormIdDoor();
+        var door = OpenWriteFormIdDoor();
         FormKey donorFk;
         try { donorFk = door.Parse(sourceFormid); }
-        catch (Exception ex) { return NpcCopyOutcome.Fail($"bad source formid '{sourceFormid}': {ex.Message}. Expected 'XXXXXX:Plugin.esp', e.g. '000D62:Vivace.esp'."); }
+        catch (Exception ex) { return NpcCopyOutcome.Fail(FormIdDoor.Sentence(ex, "", $"bad source formid '{sourceFormid}': {ex.Message}. Expected 'XXXXXX:Plugin.esp', e.g. '000D62:Vivace.esp'.")); }
```
Avick3110, line 6493: **(the finding is on line 6505, which is outside the diff — commenting here on the nearest changed line)**

The `source_formid` catch was converted, but the `target_formid` catch on the same door at line 6505 was not:

```csharp
catch (Exception ex) { return NpcCopyOutcome.Fail($"bad target formid '{targetFormid}': {ex.Message}."); }
```

So `copy_npc_appearance(source_formid="000D62:Vivace.esp", target_formid="FE003801")` returns `bad target formid 'FE003801': 'FE003801' is a runtime FormID, which houseCARL accepts for reading but not for writing … — write to '000801:Some.esp' instead..` — a well-formed token called malformed, with a doubled full stop. That is the exact wording problem this PR set out to remove, on the one write door that parses two tokens. No test covers the slot: `CopyNpcAppearanceRefusesARuntimeFormIdAndNamesThePluginForm` drives the donor only.

Line 6505 wants the same treatment: `FormIdDoor.Sentence(ex, "", $"bad target formid '{targetFormid}': {ex.Message}.")`.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6777,12 +6784,16 @@ public WritePatchBuilder.CreateOutcome CreateRecordsBatch(IReadOnlyList<CreateOp
...
-    WritePatchBuilder.CreateSpec? BuildCreateSpec(string? recordType, string? editorid, IReadOnlyList<BulkOp> operations,
+    WritePatchBuilder.CreateSpec? BuildCreateSpec(FormIdDoor door, string? recordType, string? editorid, IReadOnlyList<BulkOp> operations,
         string? parent, string? collection, string? grid, string? where, List<string> problems, CreateOpNaming naming)
     {
         var prefix = where is null ? "" : where + ": ";
         int before = problems.Count;
 
+        // parent= takes an EditorID as well as a FormID, so only a runtime FormID is judged here; everything else
+        // is left to the core's own parse.
+        if (door.RuntimeRefusal(parent) is { } parentRefusal) problems.Add($"{prefix}parent: {parentRefusal}");
```
Avick3110: The comment above says it right — `parent=` takes an EditorID as well as a FormID — but `RuntimeRefusal`'s test is `RuntimeFormId.TryParse`, which accepts *any* eight hex characters. Every other slot the door guards holds nothing but a FormID, so the ambiguity rule is safe there; `parent` is the one that does not.

A nested `create` whose parent sibling has an eight-hex-character EditorID (`DEADBEEF`, `BAADF00D`, `0000BEEF` — modders write these) is now refused. On an order with a plugin at load index `0xDE` the translation even succeeds, so the refusal reads `parent: 'DEADBEEF' is a runtime FormID … write to 'ADBEEF:SomeMod.esp' instead`, naming an unrelated record. That call worked before this PR, where `parent` went straight to the core's own EditorID/FormID parse.

One way out: only refuse when the token cannot also be read as an EditorID reference — e.g. leave `parent` to the core parse and let the core refuse a resolved runtime form, or refuse only after the sibling/EditorID lookup misses.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A runtime FormID (`FExxxYYY` / `XX######`) is now read-only: the write verbs refuse one and hand back the
+  `XXXXXX:Plugin.esp` form to use in its place.** A runtime FormID names a slot in the load order as it stands, so
+  a re-sort between the parse and the write would point the same eight digits at a different record; the plugin form
+  cannot move that way. The refusal names the translated form, so it can be pasted straight back into the call.
+  Reading by a runtime FormID is unchanged, and the doors that refuse are the ones the write verbs parse their
+  targets through — see `FormIdDoor.ForWrite`.
```
Avick3110, line 25: **(the finding is on line 34, below in the same Unreleased block)**

The entry three items down still says:

> Wherever a parameter holds nothing but FormIDs, houseCARL now also takes the runtime form the game, the console, Papyrus logs, SKSE logs and crash logs print…

Neither entry has shipped, so one release's notes assert the fact and then withdraw it. A reader hits line 34 and cannot tell whether `apply`/`remove`/`forward` take the runtime form. The conventions block at the top of this file rules that out — a second copy of a fact that will be edited once.

Since the older entry is unreleased, amend it in place (say the runtime form is taken wherever a FormID is *read*) rather than adding a correction above it.

src/housecarl-mcp/FormIdDoor.cs
Avick3110: Fixed in 5a57fd3: RuntimeRefusal now catches FormatException too and returns its sentence as the per-record problem, covered by CreateReportsAnUntranslatableRuntimeParentAsThisRecordsProblem.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Fixed in 55c9302: the target_formid catch now goes through FormIdDoor.Sentence, with CopyNpcAppearanceRefusesARuntimeTargetFormIdAndNamesThePluginForm driving that slot.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Fixed in 5a57fd3: the runtime refusal on parent= now runs only after the same-call editorid lookup misses, so a sibling parent named DEADBEEF still nests — CreateTakesAnEightHexSiblingEditoridAsAParent.

plugin/CHANGELOG.md
Avick3110: Fixed in 12b0fa8: the older unreleased entry now says the runtime form is taken wherever a FormID is read, so the block no longer asserts then withdraws.

PR #535, opened 2026-09-04
Fold the place tools into one, and name the folder in every rider refusal

### LINKED ISSUE #357: [GAP] the file-placing lanes have no discoverable way to name the patch folder they create

Avick3110:
`housecarl_place` is the S2 write tool: one call places any number of chosen file copies into one houseCARL-owned MO2 mod folder so they can win the virtual file system. It absorbs `housecarl_place_asset` and `housecarl_bulk_place_asset`, which are deleted here — the last single/bulk tool pair on the surface. There is no `single` flag, no `mode`, and no second tool: one destination is a set of one, written the same way as forty.

## The schema

One composition, four axes.

**SELECT — `assets=`**, the set of destinations. Each member addresses one destination as either `path` (any Data-relative file) or `formid` (an NPC, whose FaceGen mesh and tint paths houseCARL derives). A member may carry its own `source` and `source_provider`. A `formid` with no `kind` places BOTH FaceGen files; `kind` narrows it to one. This is the only place-shaped parameter, and it is set-valued at every size. It reads through the same strict list reader as `ops=` and `records=`, so it also takes `"@<absolute path>"`, and a member the shape does not declare is refused by name at its element rather than dropped — a silently dropped `source_provider` would auto-resolve some other provider's copy and read back as a successful place.

**SOURCE — `source_provider=`**, the pole, set-level with a per-member override. `source=` (which copy, as a file) is per member only: a source names ONE file, and a set of destinations is many files, so a set-level `source=` would be a value that is wrong for every member but the first. `kind=` is likewise set-level with a per-member override, and is judged once under its own name rather than per member.

A set-level pole fills in only where a pole can apply — a member whose `source` is one exact file on disk already names the copy, and the placer refuses a pole against it. Fanning the call's pole onto such a member would refuse it over input the caller never wrote there; a member's OWN pole still reaches that refusal, because the caller did write it.

**ACT** — place. Nothing else on this substrate writes.

**LANE — `patch=` / `into=`.** `patch=` names the new mod folder (default `houseCARL_Assets`, auto-suffixed if taken); `into=` extends an existing houseCARL patch folder.

**TRANSPORT — `max_chars=`**, and in-band accounting on every response.

One path, from the winner's copy:

```json
{ "assets": [ { "path": "textures/armor/iron/cuirass_1.dds" } ], "source_provider": "*winner" }
```

One NPC, both FaceGen files carried out of a mod MO2 is not loading:

```json
{ "assets": [ { "formid": "0013481:Skyrim.esm" } ], "source_provider": "Some NPC Overhaul", "patch": "MyFaceFixes" }
```

A mixed set, each member naming its own source:

```json
{
  "assets": [
    { "formid": "0013481:Skyrim.esm", "kind": "mesh",
      "source": "meshes/actors/character/facegendata/facegeom/Other.esp/00001234.nif" },
    { "path": "textures/armor/iron/cuirass_1.dds", "source_provider": "AnotherMod" },
    { "path": "meshes/clutter/common/tankard01.nif", "source": "C:\\loose\\tankard01.nif" },
    { "path": "meshes/clutter/common/goblet01.nif", "source": "C:\\mods\\X\\X - Meshes.bsa|meshes/clutter/common/goblet01.nif" }
  ],
  "into": "MyFixes.esp"
}
```

## The pole grammar, as accepted

`source_provider=` takes `*winner` — the sigil is part of the token, per §5.5 — for whichever copy currently wins the VFS, or a provider's NAME ALONE: a mod folder, `overwrite`, `Data`, or a BSA filename, matched exactly. A bare name always means a provider of that name, so no mod folder is unreachable because of what it is called. Naming a mod MO2 is not loading reads that mod's own copy off disk, and the response says so; with the pole omitted, only the mods MO2 loads are considered and contention is refused rather than guessed.

`source=` takes a Data-relative path resolved through the VFS (a path different from the destination is a rename — how a baked FaceGen head is carried onto another NPC's path), a fully-qualified loose file path, `<archive.bsa>|<entry>`, or a bare `.bsa` path whose entry is taken to be the destination.

The source pole stays **singular**: one provider, not an ordered two-folder list. #387 is not in scope here, and it is the gate on deleting `copy_npc_appearance` — that tool's ordered source universe has no spelling on `place` until #387 lands.

## The refusals

One sentence each, what went wrong and what to try.

- `assets=` empty — pass one or more members.
- A member with neither or both of `path` and `formid`; the whole call is refused with the per-member reason and nothing is placed.
- A malformed FormID, or a runtime FormID: translated first, then refused with the plugin-qualified form to paste in its place, because eight hex digits name a slot in the order as it stands rather than a record.
- A `kind` that is not `mesh` or `tint`.
- A `formid` with no `kind` and an explicit `source` that is not a fully-qualified `.bsa` — one file cannot serve two slots; set `kind`, or drop `source` and let the pole say whose copy.
- `source_provider` against an on-disk `source` — that source already names one exact copy.
- More than one provider supplies a path and no pole was named: the providers are listed by name, quoted exactly as `source_provider=` takes them back, and houseCARL will not guess which copy is correct.
- A named provider that does not supply the path: refused, never silently replaced, and the refusal says which places were looked in.
- `into=` naming a folder houseCARL does not own, or no such folder — see below.

## #357: the rider lanes can now be told how to name their folder

`ResolvePatchModFolder`'s not-found refusal said only "Omit into= to create it fresh, or check the name." That is #343's problem left standing on the rider lanes: it never named the parameter that names the folder, and its record-lane sibling's "houseCARL names it Patch" is false on every rider, each of which passes its own stem.

Each rider now hands the resolver its own naming — the parameter it actually declares, its own default stem, and any caveat that parameter carries — so the sentence is true per lane rather than shared and wrong:

| Lane | Parameter named | Default stem |
|---|---|---|
| `place` | `patch=` | `houseCARL_Assets` |
| `compile_script` | `patch_name=` | `houseCARL_Scripts` |
| `decompile_script` | `patch_name=` | `houseCARL_Scripts` |
| `nif_set` | `patch_name=` | `houseCARL_NifEdit` |
| `write_seq` | `patch=` | `houseCARL_SEQ` |
| `bsa_repack` | `patch_name=` | `houseCARL_Archive` |

`bsa_repack` is the measured exception the issue turns on: the §5.3 alias candidate order routes a bare `patch=` there to `archive_name`, the `.bsa` filename, so its sentence names `patch_name=` and says out loud that `patch=` on that tool names the archive instead. A rider whose `into=` can never be non-empty (`bsa_extract`, the merge and compact folder cuts) passes no naming and keeps the weakest true remedy, "Check the name."

## What the deleted tools did that `place` does not

No capability. It did lose three sentences of the description, which the review caught and this branch restores: that a file inside an active mod's archive is listed and reached under the ARCHIVE's name and not the mod's (and an archive MO2 loads no plugin for is listed under neither, reachable only as an on-disk `source=`), that a name the active order already provides files under is answered by the active order so a same-named mod folder is not consulted, and the unticked-mod sentence on the member's own `source_provider` description. Every capability survives, and one refusal disappears rather than an ability: `place_asset`'s "kind is required with formid, use bulk_place_asset for both" was a consequence of the split, and a FormID with no kind now places both files at any set size. `patch_name=` and the `assets=` member shape carry over unchanged in meaning; `patch_name=` is spelled `patch=` on the new tool, per §5.1, and the alias layer already routes the old spelling. The two old names answer from the retired-tool layer with the migration, and nothing more. One accidental protection did have to be restated: `place_asset` declared `patch_name`, so the `plugin_name` alias row could not reach it; `place` declares `patch`, which on this tool names a mod FOLDER, so the row is excepted here the same way it already was on `write_seq`.

## Verification

```
dotnet build housecarl.sln -c Release                                    0 errors
dotnet test src/housecarl-mcp-tests -c Release --no-build
    --filter "tier!=bridge"                                             1061 passed, 0 failed
                                                                        127/127 passed
```

Closes #357


Avick3110:
Findings 1 to 7 folded in e8dd28a, 5087c81, 23166e0. Finding 8 (a set carrying both a bare `formid` member and a `formid`+`kind` member for the same NPC is not deduped, so the second write wins) is pre-existing bulk behaviour that predates this fold — noted, not changed here.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -0,0 +1,298 @@
...
+        var door = svc.OpenWriteFormIdDoor();
+        for (int i = 0; i < assets.Length; i++)
+        {
+            var a = assets[i];
+            var reqs = MapTarget(door.Parse, a, source_provider, kind, $"assets[{i}]: ", out var err, out var withheld);
+            if (err is not null) problems.Add(err);
+            else
+            {
+                all.AddRange(reqs!);
+                if (withheld) foreach (var r in reqs!) poleWithheld.Add(r.AssetPath);
```
Avick3110: **The withheld-pole note is silently dropped for a forward-slash `path=`.**

`poleWithheld` is keyed here by `PlaceRequest.AssetPath` — the caller's raw member `path`, only `.Trim()`ed. The lookup at line 223 uses `PlaceResult.AssetPath`, which `PlaceOne` has already put through `AssetResolver.ValidateRelPath` → `Normalize` (`/` becomes a backslash, leading separator stripped). The set is `OrdinalIgnoreCase`, but the two separators are different characters, so the keys never match for a destination spelled with forward slashes.

Scenario — the spelling this tool's own description and the PR body use in every example:

```json
{ "assets": [ { "path": "meshes/clutter/common/tankard01.nif", "source": "C:\\loose\\tankard01.nif" } ],
  "source_provider": "SomeMod" }
```

`MapTarget` correctly withholds the set-level pole (the source is one exact file on disk) and adds `meshes/clutter/common/tankard01.nif` to the set. The result row comes back back-slashed, `Contains` is false, and `AppendResult` prints a plain `OK` line. The caller reads a successful placement and believes the `source_provider=` they passed was honoured — which is exactly what the comment at lines 101-102 ("Said on their row rather than dropped") and the pole's own refusal sentence say must not happen.

The FormID lane is unaffected: `FaceGenPath.For` already emits backslashes. `PlaceProbe.cs:349` only exercises that spelling too, which is why the arm stays green.

Normalizing on both sides fixes it — run the key through `AssetResolver.ValidateRelPath` when populating the set (guarded, since an invalid path throws and is reported raw).

src/housecarl-generator/CopyDifferentialHarness.cs
```diff
@@ -225,11 +225,11 @@ static void Fixture2(string root)
...
-        var placed = PlaceAssetTools.BulkPlaceAsset(svc, new[]
+        var placed = PlaceTools.Place(svc, new[]
         {
-            new PlaceAssetSpec { Formid = newKey.Value.ToString(), Kind = "mesh", Source = meshRel, SourceProvider = "DonorMod" },
-            new PlaceAssetSpec { Formid = newKey.Value.ToString(), Kind = "tint", Source = tintRel, SourceProvider = "DonorMod" },
-            new PlaceAssetSpec { AssetPath = HarvestedRel, Source = HarvestedRel, SourceProvider = "DonorMod" },
+            new PlaceTarget { Formid = newKey.Value.ToString(), Kind = "mesh", Source = meshRel, SourceProvider = "DonorMod" },
+            new PlaceTarget { Formid = newKey.Value.ToString(), Kind = "tint", Source = tintRel, SourceProvider = "DonorMod" },
+            new PlaceTarget { Path = HarvestedRel, Source = HarvestedRel, SourceProvider = "DonorMod" },
         }, null, Path.GetFileName(newPatch!));
```
Avick3110: **These positional arguments now bind to `kind=`, not `patch=`/`into=`.**

The call above was switched from `BulkPlaceAsset(svc, assets, patch_name, into)` to `PlaceTools.Place`, but this trailing line was left as context. The new signature is `Place(svc, assets, source_provider, kind, patch, into, max_chars)`, so `null` lands on `source_provider` and `Path.GetFileName(newPatch!)` lands on `kind` — `into=` is dropped entirely.

Running `dotnet run --project src/housecarl-generator copy-differential` reaches this arm with `kind = "F2New.esp"`. `PlaceTargets` validates the set-level kind first and returns `error: kind 'F2New.esp' is not valid — use 'mesh' (the head .nif) or 'tint' (the face .dds).`, so the `Check(!placed.StartsWith("error") ...)` on the next line fails, nothing is placed, and the later `FindPatchFolder(mods, "F2New")` assertions fail for an unrelated reason. The harness is not in `ci-all`, so the green run in the PR body does not cover it.

The sibling call at line 387 was converted to `into:` correctly — same fix here.

```suggestion
        }, into: Path.GetFileName(newPatch!));
```

src/housecarl-generator/F1MeasureProbe.cs
```diff
@@ -144,9 +144,9 @@ static void Measure(string root)
         if (newKey is { } nk)
         {
             Say("M2b  the successor's ASSET carry — bulk_place_asset renaming the donor's facegen onto the clone",
-                PlaceAssetTools.BulkPlaceAsset(svc, new[]
+                PlaceTools.Place(svc, new[]
                 {
-                    new PlaceAssetSpec { Formid = nk.ToString(), Kind = "mesh", Source = faceRel, SourceProvider = DonorFolder },
+                    new PlaceTarget { Formid = nk.ToString(), Kind = "mesh", Source = faceRel, SourceProvider = DonorFolder },
                 }, null, Path.GetFileName(newPatch!)));
```
Avick3110: **Same missed positional migration as `CopyDifferentialHarness.cs:233`.**

Every other call in this file was converted to named arguments; this one still passes `null, Path.GetFileName(newPatch!)` positionally into `Place(svc, assets, source_provider, kind, patch, into, max_chars)`, so the patch filename binds to `kind=` and `into=` is dropped.

`dotnet run --project src/housecarl-generator f1-measure` therefore prints for M2b `error: kind '<patch>.esp' is not valid — use 'mesh' (the head .nif) or 'tint' (the face .dds).` instead of performing the carry, and the following `successor carried :` line reports `NOTHING` — the placement never ran, and would not have gone into `newPatchFolder` even if the kind token had been valid. The measurement this arm exists to produce is lost silently; the probe is not in `ci-all`.

```suggestion
                }, into: Path.GetFileName(newPatch!)));
```

Avick3110:
All three folded.

1. **Withheld-pole note keyed on the raw path** (ff76486) — the set now takes the key through `AssetResolver.ValidateRelPath`, the same normalizer the result row's `AssetPath` goes through, so a `path="meshes/x.nif"` member's note lands on its row instead of being dropped. Guarded: a path the validator rejects is keyed raw, matching the raw path its failure row reports. New probe arm in place-asset-guard drives a forward-slash destination with an on-disk source and a set-level pole and asserts the note renders — confirmed red against the old keying, green with the fix.

2. **copy-differential harness dropped into=** (06df2e4) — `}, into: Path.GetFileName(newPatch!));` as suggested. `dotnet run --project src/housecarl-generator copy-differential`: the arm passes ("the successor's carry places all three from the switched-off donor — placed 3 of 3 asset(s)"), whole harness ends DIFFERENTIAL CLEAN.

3. **f1-measure probe, same missed migration** (b954b73) — `}, into: Path.GetFileName(newPatch!)));`. `dotnet run --project src/housecarl-generator f1-measure`: M2b performs the carry — placed 1 of 1, 34 bytes read from the disabled DonorMod, and "successor carried: 34B == donor's? True".

Grepped the tree for other positional `PlaceTools.Place` calls: those two were the only ones; everything else was already named.

Build clean, 1061 tests pass (tier!=bridge), ci-all 127/127.

Issue #251, opened 2026-07-19
[GAP] cross_plugin_query where_source=winner re-fetches each winner via untyped whole-overlay enumeration — O(candidates × overlay) for broad scopes


Avick3110:
### Context

`housecarl_cross_plugin_query` gained `where_source=winner` in #250 (closes #233) — under a `plugins=` scope the body filters decide the match against the live load-order **winner** instead of the scoped body. Correct and shipped; this issue tracks its scalability.

### The gap

The winner-source path resolves each candidate's winner body with `view.GetRecord(session, winnerPlugin, fk)`, which calls an **untyped, whole-overlay** enumeration and linear-scans for the FormKey (`src/housecarl-core/LoadOrderResolver.cs:579` — `session.Overlay(idx).EnumerateMajorRecords()` with no type filter). So for a scan with **N** scoped candidates, the winner fetch is **O(N × overlay-size)** — one full re-walk of a winner plugin's records per candidate.

This is the exact anti-pattern the resolver's own comment warns about (`LoadOrderResolver.cs:613-617`): *"the naive 'get each winner body separately' was measured at ~100 s over 9k weapons because GetRecord re-enumerates a whole overlay per call."* For a broad post-patch audit — the feature's headline use case (e.g. all NPCs a large mod touches) — this is asymptotically **worse** than the client-side workaround it was meant to replace (a single typed `WinnerRecordsOfType` pass over the order, O(order)).

The session caches the overlay **open** (mmap) but not the **enumeration**, so each `GetRecord` re-walks. The current `bulk-query-primitives` guard fixture is 3 weapons, so this is unmeasured.

### Ideal shape

Make winner-source O(order), preserving the current contract (defined_in= scope + `winner_fields=` display decoupling — winner-source records the scoped source so `winner_fields=false` still shows the origin). Candidate approaches:

- **Two-phase**: collect candidate FKs (+ their scoped source/type) from the scoped stream in one pass, then a **single** typed `WinnerRecordsOfType(candidateTypes)` pass filtered to the candidate FK set — run the body filters on the winner bodies there. O(order), streams (no per-plugin record map held).
- **Typed GetRecord** (lighter, constant-factor only): pass the record's type so the fetch enumerates just its GRUP instead of the whole overlay — removes the untyped-whole-overlay walk but stays O(N × type-group-per-plugin). A stopgap, not the asymptotic fix.

[...]

PR #536, opened 2026-09-04
Report the running server's build version in load_order_status

### LINKED ISSUE #299: [GAP] Report the running server's build version in-band (load_order_status)
## The gap

No houseCARL tool reports the running MCP server's build version. A session verifying that the installed dev package is the build it expects must read `ProductVersion` off the installed binary out-of-band (`housecarl-mcp.exe` file properties) — neither `housecarl_load_order_status` (the natural home) nor `housecarl_update_status` carries a server/version line.

This bites in practice: the 2026-07-23 session captured against a stale staged install (1.9.3-dev where 1.9.5-dev was expected), and the 2026-07-29 #251 baseline capture had to do the out-of-band binary read to rule a repeat out.

## The ask

`housecarl_load_order_status` gains a server-version line reporting the informational `ProductVersion` the binary already embeds — e.g. `1.9.5-dev+e942910...` — so installed-build-vs-`main` verification never leaves the tool surface. (S9 session/instance substrate; no new tool.)

## Provenance

Side finding of the #251 baseline capture, 2026-07-29 (`dev/projects/tool-surface-2.0/research/BASELINE_251_2026-07-29.md` — private corpus). Aaron-directed: chartered as a **final task of houseCARL 2.0** (this milestone).

Avick3110:
`housecarl_load_order_status` now renders a `server:` line in its header, naming the running server's build:

```
load order status — profile 'Default'
server:   0.0.0-dev+8ef38472df129ef7dadb2283a4be51b64a82118e
instance: C:\MO2\Skyrim SE
```

The value is the informational version the binary embeds — the release version `build-plugin.ps1` stamps from `plugin.json`, then `+` and the full commit sha (a shipped build reads `1.9.5-dev+e942910…`; the `0.0.0-dev` above is what a plain `dotnet build` stamps). Verifying that an installed houseCARL is the build you expect no longer means reading `ProductVersion` off `housecarl-mcp.exe` out of band, which is what the 2026-07-23 stale-install session and the #251 baseline capture both had to do. No new tool, no new parameter; the line is read once from the assembly attribute at first call and renders in both the summary and `lookup=` paths.

Notes on the two judgement calls:

- **Assembly.** The attribute is read from `ToolSurface.Assembly`, not `Assembly.GetEntryAssembly()`. They are the same binary for the shipped server, but under a test host the entry assembly is the host, so the line would have reported the wrong build in any in-process test. `ToolSurface.Assembly` is also the repo's one home for "which assembly is the tool surface" (there is a test enforcing that).
- **Absent attribute.** If the assembly carries no informational version, the line says so and points at the exe's file properties rather than printing a blank, as #299 asks. In practice the SDK always emits the attribute, so this is defensive.

`load_order_status` has no json lane — its response is the rendered text only — so "text and json both" collapses to the one lane here.

Not done, deliberately: `Program.cs` has its own `ServerVersion()` for the MCP handshake's `ServerInfo.Version`, which reads the same attribute but trims the `+metadata` suffix and falls back to `0.0.0-dev`. `ServerBuild` would be the better home for both, but `Program.cs` is outside this lane's file scope, so the two readers stay separate for now. They do not disagree — the handshake value is a prefix of the status line's.

One unit test: the rendered header carries `server:` plus exactly the informational version the built assembly embeds.

Closes #299

### DIFF
--- a/src/housecarl-mcp/StatusTools.cs
+++ b/src/housecarl-mcp/StatusTools.cs
@@ -40,7 +40,9 @@ public static class StatusTools
         [Description("Optional. Max characters before name lists are cut with an explicit notice. 0 = the server default (~80k).")]
             int max_chars = 0) => Guard.Tool(ToolNames.LoadOrderStatus, () =>
     {
-        if (svc.ConfigPromptOrNull() is { } prompt) return prompt;
+        // The server line goes ahead of the config prompt, not after it: an unconfigured server is exactly the stale
+        // staged install this line exists to catch, and that answer returns before anything else renders.
+        if (svc.ConfigPromptOrNull() is { } prompt) return StatusWire.ServerLine + prompt;
         var data = svc.StatusData();
         var logs = StatusWire.LogFolders(tools);                 // resolved Papyrus/crash log dirs (pure — no persist)
         var profiles = svc.NamedProfileComposition(profile);     // available-profile discovery + inactive-profile inspection: text parse only, no index build, no switch
@@ -61,10 +63,9 @@ public static class ServerBuild
             is [System.Reflection.AssemblyInformationalVersionAttribute a, ..] && !string.IsNullOrWhiteSpace(a.InformationalVersion)
             ? a.InformationalVersion : null;
 
-    /// <summary>What the status line prints: the version, or a sentence naming the absent attribute.</summary>
+    /// <summary>What the status line prints: the version, or one short clause saying where to read it instead.</summary>
     public static string Line { get; } =
-        Version ?? "unstamped build — this binary carries no informational version, so houseCARL cannot name its own build; " +
-                   "read the version off housecarl-mcp.exe's file properties instead.";
+        Version ?? "(no build stamp; read the exe's file properties)";
 }
 
 /// <summary>Renders <see cref="LoadOrderStatusData"/>: a header line per category, then the name lists (disabled mods,
@@ -72,6 +73,10 @@ public static class ServerBuild
 /// single mod/plugin verdict.</summary>
 static class StatusWire
 {
+    /// <summary>The running server's own build, so an installed-build-vs-source check never leaves the tool surface.
+    /// Public because the unconfigured answer returns before <see cref="Render"/> runs and prints this itself.</summary>
+    public static string ServerLine => "server:   " + ServerBuild.Line + "\n";
+
     public static string Render(LoadOrderStatusData d, IReadOnlyList<LogFolderView> logs, NamedProfileResult profiles, string? lookup, int cap)
     {
         var c = d.Composition;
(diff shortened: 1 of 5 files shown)


Avick3110:
Review pass: findings 1 to 3 are folded in 3398545.

1. The server line now renders ahead of the config prompt, so an unconfigured server (the stale staged install this line exists for) still names its build. Both the prompt path and the render share one `StatusWire.ServerLine`.
2. The tautological assert is gone. The test now renders with a non-null `lookup=` as well and asserts the server line in both, so the early-returning lookup branch is covered too.
3. The unstamped fallback is one clause: `server:   (no build stamp; read the exe's file properties)`.

Finding 4 (header lowers max_chars headroom) declined: the header has always been unbounded and every header line shares that, so this is pre-existing behaviour rather than something the server line introduces, and it is only reachable at a max_chars small enough that the header alone overruns it. Happy to be overruled if you want the whole header brought under the cap, but that is its own change.

Build clean, 1060 tests pass (tier!=bridge), ci-all 127/127.

src/housecarl-mcp/StatusTools.cs
```diff
@@ -37,19 +40,43 @@ public static string LoadOrderStatus(
...
+public static class ServerBuild
+{
+    /// <summary>The informational version verbatim, metadata suffix and all; null on an unstamped build.</summary>
+    public static string? Version { get; } =
+        ToolSurface.Assembly.GetCustomAttributes(typeof(System.Reflection.AssemblyInformationalVersionAttribute), inherit: false)
+            is [System.Reflection.AssemblyInformationalVersionAttribute a, ..] && !string.IsNullOrWhiteSpace(a.InformationalVersion)
+            ? a.InformationalVersion : null;
+
+    /// <summary>What the status line prints: the version, or one short clause saying where to read it instead.</summary>
+    public static string Line { get; } =
```
Avick3110: `ServerBuild` is a second, independent reader of `AssemblyInformationalVersionAttribute`, and its no-stamp fallback contradicts the one `Program.cs` `ServerVersion()` already has.

On a binary carrying no informational version — the case both fallbacks exist for — the MCP handshake's `ServerInfo.Version` reports `0.0.0-dev` while this line reports `(no build stamp; read the exe's file properties)`. A user doing exactly the installed-build-vs-source check this feature was added for, comparing the client's reported server version against the status line, then gets two contradictory answers about one process. The PR body's invariant ("the handshake value is a prefix of the status line's") does not hold there.

The PR already names the fix — `ServerBuild` as the one home for both — so this is about whether the file-scope reason for deferring it outweighs the repo's one-home rule. Folding it would also put the type in `ServerBuild.cs` rather than `StatusTools.cs`, which is what `standards/NAMING.md` asks for ("file name = class name").

Avick3110:
Folded in b65c5ab: `ServerBuild` moved to its own file `src/housecarl-mcp/ServerBuild.cs` and is now the single reader of `AssemblyInformationalVersionAttribute`, exposing the full version for the status line and the handshake form trimmed at `+`, with `Program.cs`'s `ServerVersion()` calling it. On an unstamped binary both fall back to `0.0.0-dev` (the status line adding a short `(no build stamp)` clause), so the handshake value is a prefix of the status line value in every case, and a new test asserts that invariant on the built assembly.

PR #537, opened 2026-09-04
Fold the three SKSE tools into housecarl_skse

### LINKED ISSUE #417: [GAP] native_pairing_audit: a debug-built DLL renders clean [LOADS] on the author's own machine, with no debug-build mention

`NativePairingWire.Judge` (`SkseTools.cs:807-831`) reads only `LoadBlocker`/`Kind`/`Version`; `DebugCrtImports` appears nowhere in Judge or `DllLine` (`:949-958`). On a box where the debug runtime resolves, `DebugCrtBlocker` returns null, so the DLL renders `[LOADS] … version-independent` with no debug mention. The VERDICT is right — it does load there — but the miss is the plugin author who ships a debug build, gets "your mod does nothing" reports, runs the audit locally, and is told everything is healthy. `skse_inventory` already shouts `[!] DEBUG-BUILD` (`SkseTools.cs:184-195`), so today the two tools disagree in tone on the same file.

Fix is cheap and needs no machine probe: reaching Judge's Loads path with `Info.DebugCrtImports.Count > 0` already implies the runtime resolved here, so the arm is a one-line append — "imports the debug CRT — loads here, error 126 for anyone without Visual Studio" — verdict untouched. ~4 lines + a guard arm. Originally held back only because it was a behavior addition to a merged tool mid-review (PR #213).

**Routing:** row 8's W5 routing-call list, deliberately unmilestoned. Promotion condition: the author scenario bites a real user report.

Avick3110:
`housecarl_skse_inventory`, `housecarl_native_pairing_audit` and `housecarl_skse_config_audit` were three tools over one substrate (SPEC §1 S7), each paying for the same declared-versus-runtime ceiling teaching. They are now one tool, `housecarl_skse`, with `findings=` selecting the family — `inventory` (the default), `pairing`, `config` — per SPEC §2.6 S7 and §6.3's Aaron-decided merge. The ceiling is written once in the tool description; each family's own detail (what it scans, what it leads with, where its boundary is) lives on the `findings=` parameter. SELECT is the one `filter=` the three shared, its match domain stated per family; SOURCE stays the implicit installed VFS state; ACT is return. A call runs exactly one family and every response ends on a line naming the family it ran and the exact `findings=` spelling of the two it did not, so the omitted-`findings=` default narrows only because the response says so. That footer is priced into `max_chars` rather than appended past it. `peek=` stays the inventory family's: passed with `pairing` or `config` it is refused with a sentence rather than silently ignored, and it still requires `filter=`. The three old names are deleted from `ToolNames`, from the published surface and from the Codex umbrella router, and each gets a retired-name row naming the `findings=` value that replaces it.

The PR also carries #417. `NativePairingWire.Judge` read only the load blocker, the plugin kind and the version lock, so a debug-built DLL on the author's own machine — where the debug runtime resolves, so there is no blocker and every other check passes it — rendered a clean `[LOADS] … version-independent` with no mention of the debug build. The verdict was right and is untouched: it does load there. What is new is one appended clause naming the debug CRT it imports and error 126 for anyone without that runtime. Reaching the loads path with `DebugCrtImports.Count > 0` already implies the runtime resolved here, so no machine probe is needed. The inventory family already shouted this, so the two families now agree about the same file.

Build clean, `dotnet test --filter "tier!=bridge"` 1075 passed / 0 failed, `ci-all` 127/127. New coverage: `SkseFamilySelectionTests` (the default, the three tokens, the unknown-value refusal, the `peek=` family refusal, the footer) and `native-pairing-guard` arms A3a-A3i and A4a-A4c, which pin that the loads verdict survives, that the debug clause appears, and that a clean version-independent DLL gets no such note. The `AliasActivationTests` rename pin moved 91 → 89 because the `lookup` → `filter` row now activates on one tool instead of three; the retired-name sweep is derived from the 1.9 capture, so the three redirects are covered without a new list.

**Not delivered here, and not claimed:** the §2.1 TRANSPORT remainder on this substrate. `housecarl_skse` carries `max_chars=` and nothing else — no `format='json'`, no `limit=`/`offset=`, no in-band accounting block. That is exactly what the three tools carried before the merge, so nothing regressed, but the merge does not close the §2.1 gap for S7, and doing so needs a JSON twin for each of the three family renders. Worth its own issue rather than a silent claim of compliance.

Closes #417

### DIFF
--- a/src/housecarl-mcp/SkseTools.cs
+++ b/src/housecarl-mcp/SkseTools.cs
@@ -71,6 +71,43 @@ public static class SkseTools
         return $"\n\n(this call ran findings='{mine}'. NOT run: {a}; {b}.)";
     }
 
+    /// <summary>The three family renders, one method each. The dispatch below picks one and appends the footer; behind
+    /// this seam it can be driven without a live MO2 instance, so a test can pin which family a findings= value runs
+    /// and that the answer ends on the footer. <see cref="ServiceRenders"/> is the live implementation.</summary>
+    internal interface IFamilyRenders
+    {
+        string Inventory(string? filter, bool peek, int cap);
+        string Pairing(string? filter, int cap);
+        string Config(string? filter, int cap);
+    }
+
+    /// <summary>The live renders: each family's data read from the service, handed to its own wire class.</summary>
+    sealed class ServiceRenders(LoadOrderService svc) : IFamilyRenders
+    {
+        public string Inventory(string? filter, bool peek, int cap)
+            => SkseInventoryWire.Render(svc.SkseInventory(peek ? filter!.Trim() : null), filter, cap);
+        public string Pairing(string? filter, int cap) => NativePairingWire.Render(svc.NativePairingAudit(), filter, cap);
+        public string Config(string? filter, int cap) => SkseConfigAuditWire.Render(svc.SkseConfigAudit(), filter, cap);
+    }
+
+    /// <summary>Runs the selected family and appends the footer. The footer's length is subtracted from max_chars so it
+    /// is paid for rather than added past the cap; the renders themselves test the cap BEFORE each list item and then
+    /// append their scope note, caveats and filter hint unconditionally, so a response can still overrun a tight cap by
+    /// one item plus that tail. Under a cap too small to hold both, the footer wins: it is the line that says which
+    /// family answered.</summary>
+    internal static string Dispatch(IFamilyRenders renders, SkseFamily family, string? filter, bool peek, int max_chars)
+    {
+        var footer = FamilyFooter(family);
+        int cap = Math.Max(1, (max_chars > 0 ? max_chars : 80_000) - footer.Length);
+        var body = family switch
+        {
+            SkseFamily.Inventory => renders.Inventory(filter, peek, cap),
+            SkseFamily.Pairing => renders.Pairing(filter, cap),
+            _ => renders.Config(filter, cap),
+        };
+        return body + footer;
+    }
+
     [McpServerTool(Name = ToolNames.Skse, ReadOnly = true, Title = "The SKSE layer: DLL/config inventory, native pairing, config references"),
      Description(
          // ---- what it is, and what selects a family ------------------------------------------------
@@ -170,17 +207,7 @@ public static class SkseTools
         if (PeekFamilyError(peek, family) is { } peekErr) return peekErr;
         if (SkseInventoryWire.PeekArgError(peek, filter) is { } err) return err;
 
-        // The footer is priced into the cap rather than appended past it, so max_chars bounds the whole response. Under
-        // a cap too small to hold both, the footer wins: it is the line that says which family answered.
-        var footer = FamilyFooter(family);
-        int cap = Math.Max(1, (max_chars > 0 ? max_chars : 80_000) - footer.Length);
-        var body = family switch
-        {
-            SkseFamily.Inventory => SkseInventoryWire.Render(svc.SkseInventory(peek ? filter!.Trim() : null), filter, cap),
-            SkseFamily.Pairing => NativePairingWire.Render(svc.NativePairingAudit(), filter, cap),
-            _ => SkseConfigAuditWire.Render(svc.SkseConfigAudit(), filter, cap),
-        };
-        return body + footer;
+        return Dispatch(new ServiceRenders(svc), family, filter, peek, max_chars);
     });
 }
 
--- a/src/housecarl-mcp/SkseTools.cs
+++ b/src/housecarl-mcp/SkseTools.cs
@@ -904,6 +904,21 @@ static class NativePairingWire
         return best;
     }
 
+    /// <summary>True when every candidate DLL that actually loads is a debug build, and at least one does — the class
+    /// whose implementation exists on the author's machine alone (#417). A class with one clean loading candidate is
+    /// healthy however many debug siblings sit beside it; those siblings still carry the note on their own line.</summary>
+    static bool IsDebugOnly(NativeClassEntry c, string? runtime)
+    {
+        bool any = false;
+        foreach (var dll in c.PairedDlls)
+        {
+            if (Judge(dll, runtime).Fate != DllFate.Loads) continue;
+            if (dll.Info is not { } i || i.DebugCrtImports.Count == 0) return false;
+            any = true;
+        }
+        return any;
+    }
+
     public static string Render(NativePairingAuditData d, string? filter, int cap)
     {
         if (filter is { Length: > 0 }) return RenderFiltered(d, filter.Trim(), cap);
@@ -921,8 +936,10 @@ static class NativePairingWire
         var loads = byFate[DllFate.Loads].ToList();
         // #417: a class whose only loadable candidate is a DEBUG build loads on THIS machine and nowhere else, so it is
         // a finding in its own right rather than a line of the healthy roster — which prints class names, not DLLs, and
-        // would have carried the clean checkmark over exactly the file the author needs to hear about.
-        var debugBuilds = loads.Where(c => c.PairedDlls.Any(x => x.Info is { } i && i.DebugCrtImports.Count > 0)).ToList();
+        // would have carried the clean checkmark over exactly the file the author needs to hear about. EVERY loading
+        // candidate must be debug-built: one clean DLL that loads keeps the class healthy, because that DLL implements
+        // the class for everyone.
+        var debugBuilds = loads.Where(c => IsDebugOnly(c, d.InstalledRuntime)).ToList();
         var healthy = loads.Except(debugBuilds).ToList();
 
         var sb = new StringBuilder();
--- a/src/housecarl-mcp/SkseTools.cs
+++ b/src/housecarl-mcp/SkseTools.cs
@@ -36,9 +36,14 @@ public static class SkseTools
             case "pairing": family = SkseFamily.Pairing; return true;
             case "config": family = SkseFamily.Config; return true;
         }
+        // A list-shaped value is the housecarl_check habit, where findings= names several families. Naming the shape is
+        // what turns the refusal into a fix; "not a family" alone reads as the wrong word rather than the wrong shape.
+        var shape = token.Contains(',') || token.Contains('[')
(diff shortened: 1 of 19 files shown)


Avick3110:
One `/code-review` pass, three findings, all folded:

- **The #417 clause never reached the default pairing view.** `Judge` is called from `DllLine`, and the unfiltered render calls `DllLine` only for the dead and version-verify sections — a class whose best fate is `Loads` goes to the healthy roster, which prints class names and no DLL lines. So the author of a debug build ran the audit, saw "✓ every third-party native class pairs to a mod whose DLL statically loads", and the new sentence only appeared if they happened to pass `filter=`. That is #417's own scenario surviving the fix. A class paired only to a debug build is now a finding of its own: out of the healthy roster, named in the summary line, and printed with its DLL verdict in a `DEBUG BUILD` section. Probe arms A3e/A3f pin the default view, A3g pins that a clean DLL still earns the checkmark.
- **A small `max_chars` overshot.** The footer was priced in against a 1,000-character floor, so `max_chars=300` returned about 1,230. The floor is gone; under a cap too small for both, the footer wins, because it is the line that says which family answered.
- **`findings=` is scalar here and `string[]` on `housecarl_check`.** A caller carrying that habit gets a type error with no hint. The parameter text now says one string and names the difference, and a comma-list or bracketed value is refused naming the shape rather than only the word.

Rebuild green: 1070 tests passed / 0 failed, `ci-all` 127/127.

Avick3110:
All five review findings are folded.

1. **`debugBuilds` was Any, not All** — 838e5e3. A class with a clean release DLL that loads plus a debug sibling was reported "paired only to a DEBUG BUILD" and stripped from the healthy roster. The predicate is now `IsDebugOnly`: every candidate whose Judge fate is Loads must be debug-built, and at least one must exist. A class with any clean loading DLL stays healthy; the debug sibling's own line still carries the clause. Probe arms A3h (healthy, and the summary does not name a debug-only class) and A3i (the sibling's line keeps the clause under `filter=`) — both fail on the old predicate.

2. **The `findings=` dispatch and the footer append had no test** — 37c533c. The three service calls sit behind `SkseTools.IFamilyRenders`, with `Dispatch` doing the switch and the footer; `ServiceRenders` is the live implementation. New tests drive a recording fake: each `findings=` value runs its own family's render, the answer ends on that family's footer, and the cap handed to the render is short by the footer's length.

3. **The refusal said "one family per call" twice** — 012df55. The shape clause now names only the shape and the following sentence carries the rule. `SkseInventoryWire.PeekArgError` gains the `error: ` prefix, so all three of this tool's refusals read alike. The two tests that pinned the duplicate now assert the rule appears exactly once.

4. **The debug-CRT clause was gated on the loads verdict** — 4347bee. A version-locked debug DLL on the VERIFY path got no debug mention while `findings='inventory'` flagged the same file. The clause now rides VERIFY too, worded for that case ("even where the version matches it loads HERE only"); the verdict is unchanged. Probe arms A4a-A4c.

5. **The `max_chars` comment overclaimed** — 37c533c. It now says what is true: the footer is priced into the cap, and a render can still overrun by one item plus the tail it appends unconditionally (scope, caveats, filter hint). The PR body's test count is corrected to 1075 (it was stale at 1068; the tree was at 1070 before these five, +5 new tests).

Build clean, `dotnet test --filter "tier!=bridge"` 1075 passed / 0 failed, `ci-all` 127/127.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -5,119 +5,209 @@
...
+        switch (token.ToLowerInvariant())
+        {
+            case "inventory": family = SkseFamily.Inventory; return true;
+            case "pairing": family = SkseFamily.Pairing; return true;
+            case "config": family = SkseFamily.Config; return true;
+        }
+        // A list-shaped value is the housecarl_check habit, where findings= names several families. Naming the shape is
+        // what turns the refusal into a fix; "not a family" alone reads as the wrong word rather than the wrong shape.
+        var shape = token.Contains(',') || token.Contains('[')
+            ? " findings= here takes ONE value, not a list."
```
Avick3110, line 56: **findings= as a real JSON array never reaches this refusal.**

`findings` is declared `["string","null"]`, and `ToolCallShim.Coerce` has no array→string rule. So `{"findings": ["inventory"]}` — the exact shape the `housecarl_check` habit produces — is caught by `TypeMismatches` (ToolCallShim.cs:358), which runs before the handler, and comes back as:

> parameter whose type could not be bound: findings (expects string, received array). Fix the argument's TYPE to match the schema (array parameters take JSON arrays — a single bare string is auto-wrapped) …

That names no family, no one-family-per-call rule, and its remedy sentence points the caller the wrong way. This `token.Contains('[')` branch only fires when the array arrives as the *string* `"[\"inventory\"]"`, which is the rarer shape — and `SkseFamilySelectionTests.AListShapedFindingsValueIsRefusedNamingTheShape` pins only that string form, so the real-array path is untested as well as untaught.

The tool description at line 138 explicitly promises `findings=['inventory']` is refused with the family teaching, so today the description and the wire disagree.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -884,6 +1021,14 @@ public static string Render(NativePairingAuditData d, string? filter, int cap)
             AppendCapped(sb, unpaired, cap, c =>
                 $"  - {c.ClassName} ({c.NativeCount} native fn) ← {c.WinningProvider ?? "(no provider)"} ({c.ProviderKind})");
         }
+        if (debugBuilds.Count > 0)
+        {
+            sb.Append("\nDEBUG BUILD — these load on THIS machine and nowhere else (").Append(debugBuilds.Count)
```
Avick3110, line 1050: **"nowhere else" can be false when a sibling candidate is [VERIFY].**

`IsDebugOnly` only inspects candidates whose fate is `Loads`; a clean, non-debug candidate sitting on `Verify` is skipped entirely. So with `d.InstalledRuntime == null` (game version unresolvable) and a class carrying two candidates —

- `Dbg.dll` — version-independent, debug CRT, no blocker → `Loads`
- `Clean.dll` — version-LOCKED, clean → `Verify`

— `BestFate` returns `Loads`, `IsDebugOnly` returns true, and the class lands here under "these load on THIS machine and nowhere else … every native these scripts declare is a silent no-op there". On a machine whose runtime matches `Clean.dll`'s lock, `Clean.dll` implements the class fine, so the header states a universal the scan did not establish.

Arm A4 covers a single-DLL VERIFY class; it doesn't cover this mixed one. Either treat a clean `Verify` sibling the way a clean `Loads` sibling is treated (not debug-only), or soften the header when the installed runtime is unknown.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -5,119 +5,209 @@
...
+    /// and that the answer ends on the footer. <see cref="ServiceRenders"/> is the live implementation.</summary>
+    internal interface IFamilyRenders
+    {
+        string Inventory(string? filter, bool peek, int cap);
+        string Pairing(string? filter, int cap);
+        string Config(string? filter, int cap);
+    }
+
+    /// <summary>The live renders: each family's data read from the service, handed to its own wire class.</summary>
+    sealed class ServiceRenders(LoadOrderService svc) : IFamilyRenders
```
Avick3110, line 99: **`ServiceRenders` is the one seam nothing exercises.**

This class is the only place the three families are bound to their service call and their wire class, and no test reaches it: `SkseFamilySelectionTests` substitutes `RecordingRenders` for `IFamilyRenders` wholesale, the generator probes drive the wire classes from synthetic data without the service, and the only end-to-end call of the merged tool (`AliasActivationTests`, via the `lookup`→`filter` row) runs the default inventory family. `findings='pairing'` and `findings='config'` are never invoked through the published tool at all.

So a mis-wiring that typechecks — the wrong `filter` passed, `cap` dropped, a family arm pointed at the wrong service call — ships silently. One wire-level call per family against the test fixture would close it.

Avick3110:
All three folded.

**1. `findings=` refused an array before the handler saw it — 134165c.** It is now published as `type: ["string","array","null"]` (a new `ToolSchemas.ShapeUnionParams` row) and bound as a `JsonElement`, so a real JSON array reaches the handler and gets the tool's own one-sentence refusal naming the three families and the one-family rule. **Decision: a one-element array is refused with the rest, not taken as its scalar** — the tool description says `findings=['inventory']` is refused, and accepting it would make that line false and teach a shape the tool does not run.

Two supporting changes came with it. The shim no longer wraps a bare string into an array for a parameter that declares both string and array (otherwise `findings="pairing"` would have been coerced into the very shape the handler refuses). And the argument checks now run before the config prompt: a wrong `findings=` is the same mistake configured or not, and answering with the prompt first would send the caller to configure an instance only to meet the same refusal.

Test is at the wire, through the built server: the array shape, the two-element array, the scalar control for each family, the published type, and the object shape (which stays the shim's, now naming both accepted types). Against the unfixed source, 5 of the 8 fail.

**2. A clean version-LOCKED sibling on VERIFY no longer leaves a class debug-only — 375e93d.** `IsDebugOnly` looked only at LOADS candidates, so with the installed runtime unknown a class carrying `Dbg.dll` (loads) plus `Clean.dll` (version-locked, verify) was printed under "these load on THIS machine and nowhere else" — false on any machine matching `Clean.dll`'s lock. VERIFY now disqualifies exactly as LOADS does; only DEAD is passed over, since a DEAD candidate implements the class nowhere. Probe arm A3j/A3k covers the mixed class, and A3j fails before the fix.

**3. Nothing bound the three families to their service call and wire class — 1cb2fef.** The selection tests drive the switch behind a recording seam, so swapping two arms of `ServiceRenders` broke no test. Added one call per family through the tool method against the records world (bridge-free, no SKSE layer — what is asserted is which render answered), pinning each family's own header and its own footer, plus the omitted-`findings=` default. With `Pairing` and `Config` swapped in `ServiceRenders`, two of the four fail and every existing selection test still passes.

Build clean, 1087 tests (tier!=bridge) pass, ci-all 127/127. No changelog line: the unreleased `housecarl_skse` bullets already describe this behaviour, and all three are corrections inside that same unreleased change.

PR #538, opened 2026-09-04
Give asset_status a directory and glob select form

### LINKED ISSUE #246: [GAP] asset_status: no directory/glob query for bulk sweeps
**Surfaced by:** the 2026-07-18 dark-face load-order scan - Phase 3 checked 14,958 facegen paths (facegeom nif + facetint dds for 7,479 NPCs).

**Gap.** `asset_paths` takes explicit paths only. A load-order-wide facegen sweep therefore has to ship ~15k fully-spelled paths through the tool interface (~40 calls of 375 paths; each path ~95 chars of boilerplate). Since every path in a sweep lives under a handful of directories (`meshes\actors\character\facegendata\facegeom\<Master>\` and the facetint equivalent), one directory-level call - "resolve every file under this VFS folder, with winners and providers" - would collapse the whole phase to ~20 calls (one per defining master), no path lists at all.

**Secondary papercut:** at the default `max_chars`, a large `asset_paths` array truncates the per-path list with an easy-to-miss "N more omitted" notice; a bulk consumer that keys results by path can silently under-count if it doesn't cross-check sent vs parsed counts. Consider promoting the truncation to a louder structured marker (or an error when the array exceeds what fits).

**Ask:** a `dir=` / glob form on `asset_status` (VFS-resolved, winners + provider chains per file found), keeping `asset_paths` for point queries.

Avick3110:
`housecarl_asset_status` gains `under=`, the directory and glob form of its SELECT axis (SPEC §2.3), so a sweep names a folder instead of a path list. A selector is a Data-relative directory — every file the VFS provides beneath it, loose and archive both — or a glob anchored under one, where `*` matches within a path segment, `?` one character, and `**` any run of whole segments including none. `AssetResolver.EnumerateUnder` already did the enumeration; this exposes it, with the trailing-separator and Data-root spellings fixed so `facegeom/Skyrim.esm/` and `facegeom/Skyrim.esm` answer alike. It is one value on the existing axis, not a mode and not a second tool: `asset_paths=` and `under=` compose in one call, explicit paths keep their order and place, and a path the sweep already named is not resolved twice. The measure from the issue is a test — one call over `meshes\actors\character\facegendata\facegeom\<Master>` returns the same files and the same winners as spelling every path out, which is what collapses the ~40-call facegen phase to one call per defining master.

The secondary half of the issue is the accounting. Every response now ends with one structured line — `[accounting] total= rendered= capped= truncated= offset= remaining= notes=` — written after the `max_chars` cut, so it can never itself be truncated away, and a consumer keying results by path compares numbers instead of parsing the prose notice. `limit=`/`offset=` page the selection and only the window is resolved, so a folder-wide sweep pays for what it renders; the line says where the next page starts, and says when there isn't one. A selector that matched nothing says so rather than passing as a clean sweep, an unanchored glob is refused rather than walking every mod root and archive table, and a drive-rooted or `..`-escaping selector goes through the same guard every asset path does.

Build, 1071 tests and `ci-all` (127/127) green. Fifteen new tests over a synthetic MO2 world whose facegen set is split across two loose mods and one authored BSA.

Closes #246

### DIFF
--- a/src/housecarl-core/AssetGlob.cs
+++ b/src/housecarl-core/AssetGlob.cs
@@ -36,8 +36,15 @@ public static class AssetGlob
     /// <summary>Every Data-relative path the selector names, sorted. Throws ArgumentException (naming the input, never
     /// a parameter) for a drive-rooted or parent-escaping selector, the same gate every other asset query passes, and
     /// for one that names no directory at all.</summary>
-    public static IReadOnlyList<string> Select(AssetResolver.AssetView view, string selector)
+    public static IReadOnlyList<string> Select(AssetResolver.AssetView view, string selector) =>
+        Select(view, selector, out _);
+
+    /// <summary>As <see cref="Select(AssetResolver.AssetView, string)"/>, and says whether the selector turned out to
+    /// name one FILE rather than a folder — the caller pasted a path, and the answer is that path. The caller renders
+    /// that as a note, so the selection and the wording agree about what happened.</summary>
+    public static IReadOnlyList<string> Select(AssetResolver.AssetView view, string selector, out bool namedOneFile)
     {
+        namedOneFile = false;
         var norm = AssetResolver.ValidateRelPath(selector).TrimEnd('\\');
         var prefix = LiteralPrefix(norm);
         // A selector with no literal directory in front of it would enumerate the whole VFS — every loose file in every
@@ -49,7 +56,17 @@ public static class AssetGlob
                 "under has to be anchored under a directory, or it sweeps the whole load order — " +
                 $"name a folder, e.g. 'meshes/actors/character' or 'meshes/actors/character/**/*.nif': '{selector}'");
 
-        if (!HasWildcard(norm)) return Sorted(view.EnumerateUnder(norm));
+        if (!HasWildcard(norm))
+        {
+            var beneath = Sorted(view.EnumerateUnder(norm));
+            if (beneath.Count > 0) return beneath;
+            // Nothing beneath it, but the string may BE a file the load order provides — a path pasted into under=
+            // instead of asset_paths=. Answer it as that one file rather than claiming no mod provides the folder,
+            // which would contradict what asset_paths= says about the very same string. One resolve, only on the
+            // empty branch, so the ordinary folder sweep pays nothing for it.
+            if (view.Resolve(norm).Exists) { namedOneFile = true; return new[] { norm }; }
+            return beneath;
+        }
 
         var rx = ToRegex(norm);                                  // compiled ONCE, not per candidate path
         return Sorted(view.EnumerateUnder(prefix).Where(p => rx.IsMatch(p)));
--- a/src/housecarl-core/AssetGlob.cs
+++ b/src/housecarl-core/AssetGlob.cs
@@ -33,20 +33,23 @@ public static class AssetGlob
     /// normalized to backslashes.</summary>
     public static bool IsMatch(string pattern, string path) => ToRegex(pattern).IsMatch(path);
 
-    /// <summary>Every Data-relative path the selector names, sorted. Throws ArgumentException (naming the input) for a
-    /// drive-rooted or parent-escaping selector, the same gate every other asset query passes.</summary>
+    /// <summary>Every Data-relative path the selector names, sorted. Throws ArgumentException (naming the input, never
+    /// a parameter) for a drive-rooted or parent-escaping selector, the same gate every other asset query passes, and
+    /// for one that names no directory at all.</summary>
     public static IReadOnlyList<string> Select(AssetResolver.AssetView view, string selector)
     {
         var norm = AssetResolver.ValidateRelPath(selector).TrimEnd('\\');
-        if (!HasWildcard(norm)) return Sorted(view.EnumerateUnder(norm));
-
         var prefix = LiteralPrefix(norm);
-        // A glob with no literal directory in front of it would enumerate the whole VFS — every loose file in every
+        // A selector with no literal directory in front of it would enumerate the whole VFS — every loose file in every
         // enabled mod and every entry of every archive — before a single path is rendered. Refused rather than paid.
+        // Tested on the PREFIX, not on the presence of a wildcard: "/", "\" and "//" all normalize to the Data root and
+        // sweep exactly as wide as an unanchored glob does.
         if (prefix.Length == 0)
             throw new ArgumentException(
-                "a glob has to be anchored under a directory, or it sweeps the whole load order — " +
-                $"put a folder in front of it, e.g. 'meshes/actors/character/**/*.nif': '{selector}'", nameof(selector));
+                "under has to be anchored under a directory, or it sweeps the whole load order — " +
+                $"name a folder, e.g. 'meshes/actors/character' or 'meshes/actors/character/**/*.nif': '{selector}'");
+
+        if (!HasWildcard(norm)) return Sorted(view.EnumerateUnder(norm));
 
         var rx = ToRegex(norm);                                  // compiled ONCE, not per candidate path
         return Sorted(view.EnumerateUnder(prefix).Where(p => rx.IsMatch(p)));
--- a/src/housecarl-core/AssetGlob.cs
+++ b/src/housecarl-core/AssetGlob.cs
@@ -1,0 +1,63 @@
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace HousecarlCore;
+
+/// <summary>The directory / glob SELECT form over the VFS: turn one Data-relative selector into the set of paths it
+/// names, by enumerating the selector's literal directory prefix through <see cref="AssetResolver.EnumerateUnder(string)"/>
+/// and, when the selector carries wildcards, keeping the paths the pattern matches. A selector with no wildcard IS a
+/// directory and takes everything beneath it. Pure over the resolver's own enumeration — this class opens nothing and
+/// holds nothing.
+/// <para>Wildcards: <c>*</c> matches any run of characters within one path segment, <c>?</c> exactly one such
+/// character, and <c>**</c> matches across separators. Matching is case-insensitive and runs against the whole
+/// Data-relative path, the same spelling the enumeration returns.</para></summary>
+public static class AssetGlob
+{
+    static readonly char[] Wildcards = { '*', '?' };
+
+    /// <summary>Does this selector carry a wildcard — i.e. is it a pattern rather than a plain directory?</summary>
+    public static bool HasWildcard(string selector) => (selector ?? "").IndexOfAny(Wildcards) >= 0;
+
+    /// <summary>The literal directory prefix of a selector: everything before the separator that precedes the first
+    /// wildcard. "meshes/actors/*/x.nif" prefixes to "meshes\actors"; a wildcard in the first segment prefixes to the
+    /// Data root (""). A selector with no wildcard is its own prefix.</summary>
+    public static string LiteralPrefix(string normalized)
+    {
+        int wild = normalized.IndexOfAny(Wildcards);
+        if (wild < 0) return normalized;
+        int sep = normalized.LastIndexOf('\\', Math.Max(wild - 1, 0));
+        return wild == 0 || sep < 0 ? "" : normalized.Substring(0, sep);
+    }
+
+    /// <summary>Does <paramref name="path"/> match <paramref name="pattern"/>? Both are Data-relative and already
+    /// normalized to backslashes.</summary>
+    public static bool IsMatch(string pattern, string path) => ToRegex(pattern).IsMatch(path);
+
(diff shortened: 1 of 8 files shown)


Avick3110:
All six findings folded, in two commits.

**343f519 — Refuse an under= selector that names no directory**

1. (finding 2) `AssetGlob.Select` tested `HasWildcard` before the anchoring refusal, so a wildcard-free selector normalizing to `""` — `/`, `\`, `//` — went straight to `EnumerateUnder("")` and walked every loose root and every archive table. The refusal now tests the literal prefix, which covers both shapes. Test: `ASelectorThatNamesNoDirectoryIsRefusedTheSameWayAnUnanchoredGlobIs`.

2. (finding 3) A `.` segment survived `NormalizeQueryPath`, so `./meshes` matched in the loose walk but never against a BSA table and quietly dropped the archive-provided files. `.` and empty segments are now collapsed. Test: `ADotSegmentSelectsTheSameSetAsThePlainSpellingIncludingArchiveOnlyFiles` — both spellings select the same set, BSA-only file included.

3. (finding 4) Both refusals threw with a parameter name, rendering `(Parameter 'selector')` / `(Parameter 'relPath')` on the end of the note. The parameter names are gone; the sentence names `under`. Test: `ARefusedSelectorReadsAsOnePlainSentenceWithNoParameterSuffix`.

**edfb28d — Count the accounting line off what asset_status rendered**

4. (finding 1) The next-page offset and `remaining` came from `d.Results.Count`, not from `rendered`, so under a max_chars cut the advice skipped the trimmed paths and still said `remaining=0`. Both now count rendered paths: next offset is `d.Offset + rendered`, and only `d.Offset + rendered < d.Selected` offers a next page. Test: `WalkingBySuccessiveNextPageOffsetsRendersEveryPathExactlyOnce` — limit 3, max_chars 500 over the 5-path world, walking by the line's own advice.

5. (finding 5) `capped` conflated skipped-before-window with left-after. Split into `skipped=` (what `offset=` stepped over) and `capped=` (what `limit=` left behind), so `rendered + skipped + capped + truncated == total`. XML comment and changelog updated.

6. (finding 6) The accounting block was appended after the body had spent all of max_chars. Room for its widest spelling is now reserved before the paths render. The reserve lives in `BatchRender.Render` as an optional `reserve:` argument rather than in the caller — otherwise the cut marker would have named the reduced budget instead of the max_chars the caller passed, which `BatchRenderCapTests` caught. Test: `TheAccountingBlockIsPricedInsideMaxCharsRatherThanAppendedPastIt`.

One note on the shape of the finding-6 test. The batch skeleton has always been allowed to overshoot by the item that crossed the line plus its cut marker — that is its standing contract, shared with `nif_inspect` — so an absolute `Render(d, cap).Length <= cap` does not hold at every cap either before or after this change. The test asserts the property that actually changed: what the accounting block writes fits inside the room reserved for it, at five different caps, plus that a cap large enough for the whole answer holds it whole. Findings 1 to 4 have tests that fail on the parent commit and pass here; I verified each by reverting the fix and re-running.

1076 unit tests pass (`tier!=bridge`), ci-all 127/127.

src/housecarl-core/AssetGlob.cs
```diff
@@ -0,0 +1,85 @@
...
+        // A selector with no literal directory in front of it would enumerate the whole VFS — every loose file in every
+        // enabled mod and every entry of every archive — before a single path is rendered. Refused rather than paid.
+        // Tested on the PREFIX, not on the presence of a wildcard: "/", "\" and "//" all normalize to the Data root and
+        // sweep exactly as wide as an unanchored glob does.
+        if (prefix.Length == 0)
+            throw new ArgumentException(
+                "under has to be anchored under a directory, or it sweeps the whole load order — " +
+                $"name a folder, e.g. 'meshes/actors/character' or 'meshes/actors/character/**/*.nif': '{selector}'");
+
+        if (!HasWildcard(norm)) return Sorted(view.EnumerateUnder(norm));
```
Avick3110: A wildcard-free `under=` that names an existing **file** enumerates empty, and the caller is told the load order provides nothing there — which contradicts what `asset_paths=` answers for the identical string.

Verified against this PR's own `AssetSelectWorld`:

```
under = ["meshes\\actors\\character\\facegendata\\facegeom\\HcMaster.esm\\0001.nif"]
  -> selected=0, note: "under '...0001.nif' matched no file in the active load order
                        — check the spelling, or nothing enabled provides that folder."

asset_paths = [same string]
  -> exists=True, WINS: FaceBase (loose)
```

`EnumerateUnder` only matches entries under `pre + separator`, so a file path can never match. Both parameters take Data-relative paths, so pasting one into `under=` is a natural mistake — and the answer a modder gets is a wrong statement about the load order, not a wrong-shape complaint.

Cheapest fix: in the no-wildcard branch, when the enumeration comes back empty, `view.Resolve(norm)` once and, if it exists, say so ("that is a file, not a folder — pass it in `asset_paths`") or just fold it into the selection.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -320,14 +320,50 @@ AssetResolver BuildAssetResolverLocked()
...
+                        notes.Add($"under '{sel}' matched no file in the active load order — check the spelling, or nothing enabled provides that folder.");
+                    foreach (var m in matched) if (seen.Add(m)) selected.Add(m);
+                }
+                catch (ArgumentException ex) { notes.Add($"under '{sel}': {ex.Message}"); }
+            }
+
+            // Page over the SELECTED set and resolve only the window, so a 15k-file sweep pays for what it renders.
+            var total = selected.Count;
+            var start = Math.Min(Math.Max(offset, 0), total);
+            var window = selected.Skip(start).Take(limit > 0 ? limit : int.MaxValue).ToList();
```
Avick3110, line 368: The doc above says paging means "only the window is resolved and a folder-wide sweep pays for what it renders". Two things undercut that:

1. **The selection is re-enumerated in full on every page.** `AssetGlob.Select` -> `AssetResolver.EnumerateUnder` does a fresh `Directory.EnumerateFiles(baseDir, "*", AllDirectories)` per loose root plus a scan of every entry of every BSA table, and nothing is cached across calls. Paging a 15k-file folder 100 at a time is 150 full VFS walks, not one.

2. **The accounting's own advice never mentions `limit=`.** It says `re-call with offset=N`. A caller following it literally keeps `limit=0`, so every page resolves the whole remainder: page 1 resolves 15000, page 2 resolves 14960, and so on — quadratic, and exactly the cost the paging was added to avoid.

For (2), the next-page sentence could carry the limit the caller actually used (`re-call with limit=<L> offset=<N>`), or default the advice to a limit when none was passed. For (1), memoizing the enumeration per (selector, snapshot) would make a paged walk pay once.

src/housecarl-mcp/AssetTools.cs
```diff
@@ -52,18 +73,100 @@ public static string Render(AssetStatusData d, int cap)
...
+    }
+
+    /// <summary>What each under= selector had to say for itself — a selector that matched nothing, or was rejected.
+    /// Above the per-path list, with the other alarms, so a truncated sweep cannot cut it away. Uncapped, and bounded
+    /// by the call's own input: there is at most one note per selector the caller wrote.</summary>
+    static void AppendSelectorNotes(StringBuilder sb, IReadOnlyList<string>? notes)
+    {
+        if (notes is not { Count: > 0 }) return;
+        sb.Append("\n[!] under (").Append(notes.Count).Append("):\n");
+        foreach (var n in notes) sb.Append("  - ").Append(n).Append('\n');
```
Avick3110: This block is uncapped, unlike the two alarm blocks written immediately before it — `AppendReadFailures` and `AppendDiscoveryWarnings` both go through `BatchRender.AppendLines`, which honours `cap`.

The "bounded by the call's own input" argument is what protects it, but the input is an unbounded `string[]`. `under=` with 2,000 mistyped selectors produces 2,000 notes of ~140 chars each: ~280KB appended before the budget check ever runs, so the response overshoots `max_chars` by two orders of magnitude — the exact failure the reserve machinery in this same PR exists to prevent.

Routing it through `BatchRender.AppendLines(sb, notes, "selector(s)", cap)` would make it read and behave like its siblings.

src/housecarl-core/AssetResolver.cs
```diff
@@ -421,8 +421,10 @@ public IReadOnlyList<AssetHit> ResolveMany(IEnumerable<string> relPaths)
 
     IReadOnlyCollection<string> EnumerateUnder(string prefix, Snapshot snap)
     {
-        var pre = NormalizeQueryPath(prefix);                    // drive-root / '..' rejected loud, backslash-normalized
-        var withSep = pre + "\\";                                // match a SUBTREE, not a sibling whose name starts with 'pre'
+        var pre = NormalizeQueryPath(prefix).TrimEnd('\\');      // drive-root / '..' rejected loud, backslash-normalized
+        // Match a SUBTREE, not a sibling whose name starts with 'pre'. An empty prefix is the Data root, where every
+        // entry is under it and the separator test would exclude them all.
+        var withSep = pre.Length == 0 ? "" : pre + "\\";
```
Avick3110, line 473: (anchored here; the line at issue is the loose-lane `found.Add(Normalize(f.Substring(rootDir.Length)))` a few lines below.)

Not introduced here, but this PR makes it user-visible for the first time: `baseDir` is `Path.Combine(rootDir, pre)`, so the prefix portion of every loose result carries the **caller's** spelling, while BSA-lane entries carry the archive's. One response can therefore mix two spellings of the same folder.

Verified on this PR's fixture with `under = "MESHES\\ACTORS\\**\\*.NIF"`:

```
MESHES\ACTORS\character\facegendata\facegeom\HcMaster.esm\0001.nif   (loose)
...
meshes\actors\character\facegendata\facegeom\HcMaster.esm\0005.nif   (BSA)
```

Before this PR `EnumerateUnder` was only ever called with fixed literal prefixes, so the caller's casing was always canonical. Now the PR's headline consumer — "a consumer keying results by path" — gets keys whose case depends on how the selector was typed, and an ordinal compare against paths from `asset_paths=`, from a record, or from an earlier differently-cased sweep will miss.

Re-rooting the returned path on the normalized `pre` (`pre + f.Substring(baseDir.Length)`) would make the answer independent of how the selector was spelled.

Avick3110:
All four folded, each with a test that fails without its fix (verified by stashing the source changes and re-running: all four failed, then passed with them back).

**1. `under=` naming an existing file — eb4f184.** In the no-wildcard branch, when the enumeration comes back empty the selector is resolved once as a path; an existing one is folded into the selection as that single file and a note says under= took it as a file. One resolve, only on the empty branch, so the ordinary folder sweep pays nothing for it. Test asserts the answer agrees with `asset_paths=` on the same string, winner and all, and that the "matched no file" note is gone.

**2b. Next-page advice — 4f38b6e.** The advice now reads `re-call with limit=<L> offset=<N>`, where `<L>` is the limit the caller passed, or 200 when none was. `AssetStatusData` carries the asked-for limit alongside the asked-for offset; the accounting reserve measures the wider of the two so the line still fits inside its held-back room.

**2a. Enumeration memoization — not built; doc narrowed instead.** The cache itself is small (a dictionary on `Snapshot`, which dies with the snapshot on rebuild). The cost is elsewhere: `RefreshIfStale` only detects a loose change by re-statting the dirs it recorded at warm time, and `EnumerateUnder` walks `AllDirectories`. A file added into a nested folder under a cached prefix moves that folder's mtime, not the prefix's, so a memoized enumeration would keep answering without it — an answer that is silently wrong and stays wrong for the life of the snapshot. Making it correct means stamping every directory the walk crossed and checking them all on every `RefreshIfStale`, which runs per asset call, including single-path resolves. That is the enumeration cost back again on the cheap path, and it is well past a small bounded cache. So: no memoization, and the doc sentence in `LoadOrderService.AssetStatus` now says what is true — the window bounds the per-path RESOLUTION, the expensive half; the selector enumeration runs once per page.

**3. Selector-notes cap — 231188a.** Routed through `BatchRender.AppendLines(sb, notes, "selector(s)", cap)` like its two siblings. Test drives 2,000 bad selectors at `max_chars=2000` and asserts the render stays under 10KB with the named cut marker, while `notes=2000` is still stated in full in the accounting line.

**4. One spelling per answer — e423a41.** One correction to the finding: the loose-lane change alone is a no-op. `Directory.EnumerateFiles` echoes back the directory string it was handed, so `f.Substring(rootDir.Length)` and `pre + f.Substring(baseDir.Length)` produce the same string — both already carry the caller's spelling. The mixing came from the other lane: BSA entries were added verbatim from the archive tables. So both lanes are now re-rooted on the normalized prefix (`withSep + …`), which is what makes one response spell one folder one way. The prefix casing is the selector's own; what the fix removes is the two-spellings-in-one-answer split. Test uses an upper-cased selector and asserts every row, loose and archive alike, starts with that one prefix, and that typing changes no file in the set.

Changelog updated under the two existing Unreleased entries rather than adding new ones. Build clean, 1080 unit tests pass (`tier!=bridge`), ci-all 127/127.

Issue #376, opened 2026-08-19
[GAP] no tool reports a plugin's localized header flag — the in-place refusal cannot be anticipated


Avick3110:
## The gap

No houseCARL tool reports whether a plugin carries the **localized** header flag. Measured: `ModHeader.Flags` is referenced nowhere in `src/housecarl-mcp/` — only in the core write paths — and a case-insensitive search for "localiz" across the MCP layer finds only refusal text. So there is no read, scan, or status surface that answers "is this plugin localized?".

## Why it matters now

The in-place record-editing lanes (`housecarl_apply` / `housecarl_create` / `housecarl_remove` / `housecarl_forward` with `in_place=`, and `housecarl_compact_plugin(in_place=true)`) now **refuse** a plugin carrying that flag, because re-serializing one corrupts its text. That refusal is correct, but a modder cannot find out in advance which of their plugins it will apply to without opening files in xEdit.

Which plugins carry the flag depends on the install: the official masters always do, and translated releases commonly do, because that is what the translation toolchain emits. On a translated install this is a meaningful share of the order, not an edge case — so "which of my plugins can I not edit in place?" is a question the tool surface should be able to answer and currently cannot.

## Shape

`housecarl_load_order_status` is the natural home — it already walks the order per plugin and reports per-plugin state, and the flag is a header read (no strings resolution involved, so it needs none of the game-Data fallback machinery). A per-plugin boolean, or a listing of the localized ones, would close it.

Worth considering alongside it: the light/ESL and master flags are the same kind of header fact and are also not surfaced, so the row may be "report the header flags" rather than this one bit.

## Provenance

Raised in Aaron's review of the PR that added the refusal, where the changelog was setting an expectation the reader had no way to check. The changelog now states plainly that no tool reports the flag today, rather than naming one that would not answer — so this gap is what makes that sentence retractable.

Deliberately not built in that PR: the refusal was scoped to a refusal, and adding a read surface to it would have been scope creep. Unrouted and unmilestoned — it lands at the next routing stroke.

Avick3110:
[...]

PR #540, opened 2026-09-04
Strings and localization: the redirect gate, and what merge and compact do not say

### LINKED ISSUE #367: [GAP] the identify pass cannot see a plugin that only DECLARES a donor as a master — compact/merge report it as having no dependents
## Summary

`RemapEngine.IdentifyExternalReferencers` finds two kinds of dependent: a plugin whose record **links** to a target FormKey, and a plugin whose record **shares** a target FormKey (an override). Both come from walking `view.RecordsIn(...)` and `EnumerateFormLinks()`. It never opens a plugin **header**, so a plugin that lists a donor in its master list without referencing any of its records is invisible to it.

Both `merge_plugins` and `compact_plugin` then tell the user there is nothing to worry about.

## Failure scenario (Aaron's, from the PR #365 review)

`MyPatch Compat.esp` declares `MyPatch.esp` as a master but references none of its records — a normal shape for a compat patch that exists to be load-ordered, or one whose overrides were later trimmed.

```
housecarl_merge_plugins plugins=["MyPatch.esp"] output="My Patch v2.esp"
```

The report says `external referencers: none`. The user follows the swap instruction, deactivates `MyPatch.esp`, and `MyPatch Compat.esp` is now missing a master — the game refuses to load it.

## Why the rename lane makes this sharper

Pre-existing and merge-wide, but a **pure-override donor** — which #345's PR body calls "what most mis-named patches are" — originates no records, so `plan.Dict` is empty, so `targets` is empty, and the pass returns nothing *by construction*. The case where the pass has the least to say is now the headline case of the feature.

## What shipped in PR #365, and what did not

PR #365 folded the honesty half: the identify-pass line now states its own coverage —

> identify-pass scanned N plugin(s) — it reads record links and record identity, NOT declared masters or INI distributor configs (SPID/KID/SkyPatcher), so a plugin that only lists a donor as a master, or only names one in an .ini, is not counted above.

That stops the report claiming more than it measured. It does not make the dependent detectable.

## What the fix actually involves

Not a one-line addition:

- A **new detection category** — a declared-master dependent is neither a referencer nor an overrider — through `IdentifyResult`, both outcome records, and both renders (`RenderMerge` and `RenderCompact`).
- Reading each candidate plugin's **header** rather than its records. Cheap per plugin, but a different read from the one the pass does today.
- A **policy question for the shared compact lane**: compact *refuses* on external referencers where merge warns. Whether a declared-master dependent should refuse, warn, or be listed separately is a design call, not an implementation detail.

Related blind spots in the same pass: #364 (INI distributor configs, same "not a record link" shape).

## Provenance

Aaron's review of PR #365, finding F3. Confirmed at source: `IdentifyExternalReferencers` walks `RecordsIn` + `EnumerateFormLinks` and opens no header. Pre-existing on both verbs — the #324 shape, so filed rather than folded into that branch.

### LINKED ISSUE #369: [Bug] OpenOverlay's strings redirect is suppressed by any .bsa beside the plugin, including one holding no strings

### LINKED ISSUE #371: [GAP] merge/compact write blanks silently when a localized donor's strings resolve nowhere

### LINKED ISSUE #435: [Bug] merge_plugins de-localizes a localized donor silently — no note, other languages and the plugin's nature dropped without a word

Avick3110:
Six issues that share one detection: whether a plugin's strings actually resolve, and what merge and compact leave behind when they do not.

The shared piece is `LocalizedStrings.OwnFolderCarriesStringsFor` — "does this plugin's own folder carry a strings source FOR THIS PLUGIN". The overlay open's redirect gate now delegates to it instead of answering on the existence of any `Strings\` folder or any `.bsa`, so an asset-only archive no longer suppresses the game-Data lookup for a localized plugin (#369). Its sibling `ResolvesNowhere` reads the shape classifier that was already there, and merge and compact consume it: a localized source whose `.STRINGS` houseCARL can find nowhere is refused instead of having its blanks written into a plugin the caller keeps (#371). Merge also states the de-localization it performs on a donor whose strings do resolve — the compact lane already did (#435). The identify pass now reads each candidate's declared masters as well as its records, so a plugin that only lists a donor as a master is reported instead of counting as no dependent at all (#367); merge names those, compact does not, because a compaction keeps the source's basename and a declarer's master survives it. The `.esl` merge refusal no longer gives a reason that is false for the light single donor a rename makes ordinary (#363, the wording half).

Tests: 1071 pass (13 new — eight on the redirect gate against real authored archives, five on the declarer detection and its render). `ci-all` is ALL PASS (127/127); `strings-decision-guard` was updated to the fixed contract, the merge and compact guards gained arms for the refusal and the de-localization note (the merge guard's old RESIDUAL arm, which pinned the silent blanks, now pins the refusal), the merge guard gained an arm driving a real declarer plugin through `MergePlugins` to the rendered warning, and the compact guard one on the refusal's remedy for a `Strings\` folder that will not list.

Not done here, and why:

- **#376 (no tool reports the localized header flag)** is untouched. Its only home is `housecarl_load_order_status` — the tool and its render both live in `StatusTools.cs`, which this session was told not to edit. Nothing else on the surface walks the order per plugin. It needs a session that owns that file.
- **#363's carry decision** stays open. What remains after the wording fix is whether a single-donor merge should CARRY the donor's ESL/ESM flag, which forks on the recorded no-ESL-in-v1 scope call at PR #158 — a ruling, not a fold. The report already names the dropped flag.

One cost worth naming: the redirect gate can now open a `.bsa` beside a plugin to ask whether it embeds that plugin's tables. It is the folder-keyed archive cache the strings classifier already builds, so it is one walk per mod folder per process rather than per open — cheaper across a session than re-reading each plugin's header would be, but it is a real first-index cost on an order with many archives.

Closes #369
Closes #371
Closes #435
Closes #367

### DIFF
--- a/src/housecarl-core/LocalizedStrings.cs
+++ b/src/housecarl-core/LocalizedStrings.cs
@@ -287,6 +287,53 @@ public static class LocalizedStrings
         return (a.Shape, LocalizedTargetUnsupportedException.ShapeBody(a));
     }
 
+    /// <summary>Does the plugin's OWN folder carry a strings source FOR THIS PLUGIN — a loose table beside it, or a
+    /// <c>.bsa</c> there that embeds one? The read side's redirect gate (<see cref="LoadOrderResolver.OpenOverlay"/>):
+    /// false means the folder-adjacent lookup would resolve nothing for this plugin, so the open points at game-Data
+    /// instead.
+    ///
+    /// <para>Answers for the NAMED plugin, never for the folder in general. A <c>Strings\</c> folder holding only a
+    /// NEIGHBOUR's tables, and an asset-only <c>.bsa</c> (meshes/textures — the common case), are not this plugin's
+    /// source; counting either suppressed the redirect for a whole population of localized plugins whose text was
+    /// reachable in game-Data and read back blank everywhere (#369).</para>
+    ///
+    /// <para>Cost: the loose look is one directory listing, and the archive look is the folder-keyed
+    /// <see cref="ArchiveStrings"/> cache the classifier already builds — so a mod folder's archives are enumerated
+    /// once per server process rather than once per open. Any IO fault answers TRUE, keeping the unchanged
+    /// folder-adjacent open: we only ever redirect on a clean, empty read.</para></summary>
+    public static bool OwnFolderCarriesStringsFor(string pluginPath)
+    {
+        try
+        {
+            var folder = Path.GetDirectoryName(pluginPath);
+            if (folder is null) return true;
+            var stem = Path.GetFileNameWithoutExtension(pluginPath);
+            var loose = ReadStringsFolder(Path.Combine(folder, "Strings"), stem);
+            // A folder that is there and could not be LISTED tells us nothing about what is in it — keep the default.
+            if (loose.Read == StringsFolderRead.Unlistable) return true;
+            if (loose.Languages.Count > 0) return true;
+            // BsaEmbedding returns the archive's path both when it embeds this plugin's tables and when it could not
+            // be parsed at all, so a non-null answer covers "found it" and "cannot see in there" alike — and the
+            // second keeps the unchanged open rather than redirecting past an archive that may hold the strings.
+            return BsaEmbedding(folder, stem).Path is not null;
+        }
+        catch { return true; }
+    }
+
+    /// <summary>Could houseCARL find NO strings source at all for a plugin it read and found flagged localized? True
+    /// for the two shapes where a read resolves nothing: nothing found anywhere, and a <c>Strings\</c> folder beside
+    /// the plugin that could not be listed. Every value such a plugin carries reads EMPTY, so a lane that would bake
+    /// that read into a new file refuses rather than writing blanks (#371).
+    ///
+    /// <para><see cref="LocalizedShape.Unreadable"/> is deliberately NOT here: the plugin's own header was never
+    /// read, so nothing is known about its strings, and a lane that must act on that answers it as unreadable in its
+    /// own words.</para></summary>
+    public static bool ResolvesNowhere(LocalizedShape shape) => shape switch
+    {
+        LocalizedShape.Nowhere or LocalizedShape.StringsFolderUnreadable => true,
+        _ => false,
+    };
+
     /// <summary>The loose table files this plugin's own folder carries — what a read resolves against, and what a
     /// refusal has to leave byte-identical.</summary>
     public static IReadOnlyList<string> OwnTableFiles(string pluginPath)
--- a/src/housecarl-core/LocalizedStrings.cs
+++ b/src/housecarl-core/LocalizedStrings.cs
@@ -306,7 +306,9 @@ public static class LocalizedStrings
         try
         {
             var folder = Path.GetDirectoryName(pluginPath);
-            if (folder is null) return true;
+            // A folder that is not there is not a clean, empty read — it is a bad one, and the redirect is only ever
+            // taken on a clean one.
+            if (folder is null || !Directory.Exists(folder)) return true;
             var stem = Path.GetFileNameWithoutExtension(pluginPath);
             var loose = ReadStringsFolder(Path.Combine(folder, "Strings"), stem);
             // A folder that is there and could not be LISTED tells us nothing about what is in it — keep the default.
--- a/src/housecarl-core/LocalizedStrings.cs
+++ b/src/housecarl-core/LocalizedStrings.cs
@@ -182,7 +182,11 @@ public static class LocalizedStrings
         // An archive the write cannot rewrite decides the shape regardless of what else is on disk — a loose set beside
         // a BSA that also embeds this plugin's strings is an ambiguous source, and which one Mutagen prefers is not
         // something this classifier should be guessing at when the answer is "refuse either way".
-        var (bsaPath, bsaUnreadable) = BsaEmbedding(folder, stem);
+        // The folder-unlistable answer is not consumed here: a mod folder that will not list also hides its
+        // Strings\ folder, so this classification falls through to the shapes that resolve NOWHERE and every lane
+        // reading it refuses. Fail-CLOSED is the right default for a write; the read side's gate takes the third
+        // answer, because its default is to leave the open unchanged.
+        var (bsaPath, bsaUnreadable, _) = BsaEmbedding(folder, stem);
         if (bsaPath is not null)
             return new LocalizedAssessment(LocalizedShape.BsaEmbedded, Names(own), Incomplete(own), Names(gameData),
                                            bsaPath, bsaUnreadable, dataDir is null, UnmatchedTables: unmatched);
@@ -215,7 +219,7 @@ public static class LocalizedStrings
         // nothing nearer was found, so an ordinary plugin never pays for it.
         if (dataDir is not null)
         {
-            var (gdBsa, gdUnreadable) = BsaEmbedding(dataDir, stem);
+            var (gdBsa, gdUnreadable, _) = BsaEmbedding(dataDir, stem);
             if (gdBsa is not null)
                 return new LocalizedAssessment(LocalizedShape.BsaEmbedded, Array.Empty<string>(), NoneIncomplete,
                                                Array.Empty<string>(), gdBsa, gdUnreadable, false, BsaInGameData: true,
@@ -317,7 +321,11 @@ public static class LocalizedStrings
             // BsaEmbedding returns the archive's path both when it embeds this plugin's tables and when it could not
             // be parsed at all, so a non-null answer covers "found it" and "cannot see in there" alike — and the
             // second keeps the unchanged open rather than redirecting past an archive that may hold the strings.
-            return BsaEmbedding(folder, stem).Path is not null;
+            var bsa = BsaEmbedding(folder, stem);
+            // The mod folder itself would not list — the same fact as the unlistable Strings\ folder above, and the
+            // same answer: nothing is known about what is beside the plugin, so the default open stands.
+            if (bsa.FolderUnlistable) return true;
+            return bsa.Path is not null;
         }
         catch { return true; }
     }
@@ -447,19 +455,25 @@ public static class LocalizedStrings
     /// when the archive could not be parsed at all — an archive we cannot see into is refused rather than assumed
     /// harmless (a malformed one is not merely opaque: it takes the plugin's own open down with an ArchiveException).
     /// No archive is opened unless one is present, which is the common case at ~0.06 ms.</summary>
-    static (string? Path, bool Unreadable) BsaEmbedding(string folder, string stem)
+    static (string? Path, bool Unreadable, bool FolderUnlistable) BsaEmbedding(string folder, string stem)
     {
+        var (folderUnlistable, entries) = ArchiveStrings(folder);
+        // The folder would not list, so which archives are in it — and whether any carries this stem — was never
+        // established. Reported as its own answer rather than as "no archive": there is no archive to name, and a
+        // caller that must not conclude an absence needs to be told it cannot.
+        if (folderUnlistable) return (null, false, true);
+
         // A POSITIVE hit wins over an unreadable one: returning on the first unreadable archive would let a single
         // unparseable .bsa anywhere in game-Data answer for every plugin that searched there, naming an archive that
         // neither sits beside the plugin nor carries its tables. An archive we cannot see into still refuses, but only
         // once no archive has actually claimed the stem.
         string? unreadable = null;
-        foreach (var e in ArchiveStrings(folder))
+        foreach (var e in entries)
         {
             if (e.Unreadable) { unreadable ??= e.Archive; continue; }
-            if (e.Stems.Contains(stem)) return (e.Archive, false);
+            if (e.Stems.Contains(stem)) return (e.Archive, false, false);
         }
-        return unreadable is null ? (null, false) : (unreadable, true);
+        return unreadable is null ? (null, false, false) : (unreadable, true, false);
     }
 
     /// <summary>Per archive in a folder: which plugin stems it embeds strings for. Cached by folder, because game-Data
@@ -502,17 +516,20 @@ public static class LocalizedStrings
         return string.Join(" ", parts);
     }
 
-    static List<ArchiveEntry> ArchiveStrings(string folder)
+    static (bool Unlistable, List<ArchiveEntry> Entries) ArchiveStrings(string folder)
     {
         string[] archives;
+        // THE SAME THREE ANSWERS the loose look has, for the same reason: a folder that could not be listed is not a
+        // folder found to hold no archives, and returning the empty list for both made the read side's gate conclude
+        // an absence it never checked.
         try { archives = Directory.GetFiles(folder, "*.bsa"); }
-        catch (IOException) { return NoArchives; }
-        catch (UnauthorizedAccessException) { return NoArchives; }
-        if (archives.Length == 0) return NoArchives;
+        catch (IOException) { return (true, NoArchives); }
+        catch (UnauthorizedAccessException) { return (true, NoArchives); }
+        if (archives.Length == 0) return (false, NoArchives);
         var stamp = ArchiveStamp(archives);
 
         lock (ArchiveCache)
-            if (ArchiveCache.TryGetValue(folder, out var hit) && hit.Stamp == stamp) return hit.Entries;
+            if (ArchiveCache.TryGetValue(folder, out var hit) && hit.Stamp == stamp) return (false, hit.Entries);
 
         var entries = new List<ArchiveEntry>();
         foreach (var a in archives)
@@ -538,7 +555,7 @@ public static class LocalizedStrings
         }
 
         lock (ArchiveCache) ArchiveCache[folder] = (stamp, entries);
-        return entries;
+        return (false, entries);
     }
 
     static IReadOnlyList<string> Names(Dictionary<string, List<string>> m) => m.Keys.OrderBy(x => x).ToList();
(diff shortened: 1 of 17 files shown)


Avick3110:
All eight review findings folded. `dotnet test` 1071 pass, `ci-all` 127/127 ALL PASS.

| # | Finding | Commit |
|---|---|---|
| 1 | `DeclaredMasters` opened the overlay bare, so a held-open plugin reached the declarer scan as a generic exception and was reported as an enumeration fault ("closing programs will not help") instead of unopenable. Wrapped the open the way `RecordsIn` does. | 34b12a2 |
| 2 | The held-open test asserted only that something was reported; it now asserts the plugin name and `UnscannableCause.Unopenable`. Failed before the fix with `Expected: Unopenable / Actual: EnumerationFault`. | 34b12a2 |
| 3 | `ArchiveStrings` swallowed `IOException`/`UnauthorizedAccessException` from the archive listing and returned "no archives", so an unlistable mod folder made the redirect gate answer false — the opposite of the loose side's rule. The fault is now its own answer and the gate takes it. New test denies `ListDirectory` on a real mod folder; it failed before the fix. `Assess` deliberately does not consume the new answer: the write side stays fail-closed there. | 4e2157a |
| 4 | The #371 refusal's remedy was shared across shapes, so `StringsFolderUnreadable` was told to place tables in the folder houseCARL had just failed to open. Remedy is now per shape ("close whatever is holding it open, or fix its permissions"). New `compact-service-guard` arm with a real deny ACE; it failed before the fix on the old sentence. | f8c436f |
| 5 | New `merge-service-guard` arm on its own instance: builds a real declarer-only plugin, runs `MergePlugins`, and asserts the plugin is in `MasterDeclarers` (and in neither existing list) and named in the rendered WARNING. Verified it fails when `readDeclaredMasters` is flipped off. The probe's stale RESIDUAL doc bullet went with it. | 1807dd7 |
| 6 | UTF-8 BOM removed from `LoadOrderService.cs` line 1. Checked every changed file for the same drift; no others. | 62ab0dc |
| 7 | PR body corrected: 13 new tests, five on the declarer detection (it was 12/five at review time; finding 3 added one). | body edited |
| 8 | The "unrelated plugin" test scanned a plugin declaring no masters at all, so it passed with or without the transform-set filter. It now scans a plugin declaring a real master outside the set; verified it fails when `.Where(transformSet.Contains)` is removed. | 4f9aca5 |

Nothing refuted — all eight reproduced on the branch.

src/housecarl-core/LocalizedStrings.cs
```diff
@@ -182,7 +182,11 @@ public static LocalizedAssessment Assess(string pluginPath, string? dataDir)
         // An archive the write cannot rewrite decides the shape regardless of what else is on disk — a loose set beside
         // a BSA that also embeds this plugin's strings is an ambiguous source, and which one Mutagen prefers is not
         // something this classifier should be guessing at when the answer is "refuse either way".
-        var (bsaPath, bsaUnreadable) = BsaEmbedding(folder, stem);
+        // The folder-unlistable answer is not consumed here: a mod folder that will not list also hides its
+        // Strings\ folder, so this classification falls through to the shapes that resolve NOWHERE and every lane
+        // reading it refuses. Fail-CLOSED is the right default for a write; the read side's gate takes the third
+        // answer, because its default is to leave the open unchanged.
+        var (bsaPath, bsaUnreadable, _) = BsaEmbedding(folder, stem);
```
Avick3110: **The third answer is discarded here, and the shape it falls through to asserts an absence nothing checked.**

`BsaEmbedding` now has three answers, but `Assess` takes only two. The comment says a mod folder that will not list also hides its `Strings` folder — that is not so on Windows: `ListDirectory` is a separate right from traverse, so `Directory.Exists(folder/Strings)` and the enumeration *inside* it still succeed while `Directory.GetFiles(folder, "*.bsa")` throws. So the fall-through is not to "every shape resolves nowhere", it is specifically to `Nowhere`.

Scenario: a localized plugin in a mod folder on a share (or carrying a deny-list ACE), no loose `Strings` set, no game-Data set, tables in a `.bsa` in that same folder. `BsaEmbedding` returns `FolderUnlistable: true`; `Assess` throws it away, `bsaPath` is null, and the classification lands on `Nowhere`. Two consequences:

1. `ResolvesNowhere(Nowhere)` is true, so `housecarl_merge_plugins` and `housecarl_compact_plugin` refuse — with `WhereTheTextIs`'s `Nowhere` sentence, which states "*and no archive beside it ... carrying its tables*". Nothing checked that; the folder would not list. `StringsFolderUnreadable` was added precisely so an unreadable folder cannot arrive at that sentence and be described as an empty one; the mod-folder case now has exactly that gap. The remedy is wrong for the same reason — "place them in a Strings folder beside the plugin" instead of "free the folder", which is the distinction `UnresolvableStringsRefusal` is careful to draw for the other shape.
2. `LoadOrderResolver.FolderHasOwnStrings` answers **true** for the identical folder (`bsa.FolderUnlistable` → keep the default open). The doc-comment this PR adds there says the two sides "cannot disagree about whether a source exists". For this input they do.

Fail-closed is the right posture; the shape it fails closed *on* is the problem. Consider a shape (or reusing `StringsFolderUnreadable`'s wording) that says the folder could not be read, so the refusal stops claiming an absence and points at the permissions.

src/housecarl-core/RemapEngine.cs
```diff
@@ -131,6 +153,37 @@ public static IdentifyResult IdentifyExternalReferencers(
...
+                    continue;
+                }
+                catch (Exception ex)
+                {
+                    // It opened and the master table would not parse. Closing programs does not fix that, so it
+                    // takes the other cause, exactly as a record enumeration that throws part way does.
+                    unscannable++;
+                    unscannablePlugins.Add(new UnscannablePlugin(plugin, UnscannableCause.EnumerationFault,
+                                                                 $"master table unreadable: {ex.GetType().Name}: {ex.Message}"));
+                    continue;
```
Avick3110: **This `continue` costs the record walk, and only for the caller that opted in.**

The block's own comment justifies skipping with "a header that will not read means the file will not open at all — the record scan below would fail the same way". That is the argument for the `PluginUnreadableException` arm above. This arm is the other case, and its comment says so explicitly: *"It opened and the master table would not parse."* A plugin whose TES4 master table is malformed can still enumerate its records fine, and `continue` throws that scan away.

Scenario: a plugin with a corrupt master table also carries a FormList linking to a donor record. Through `housecarl_compact_plugin` (`readDeclaredMasters: false`) the record walk runs and the plugin lands in `ExternalPlugins` with the repoint remedy. Through `housecarl_merge_plugins` (`readDeclaredMasters: true`, the only caller that sets it) the same plugin is `continue`d, and `RenderMerge` prints "external referencers: none" for a plugin that does reference donor records. Turning a reporting feature on reduces the coverage of the pass it was added to.

It is named in `UnscannablePlugins` and the "'none' may be incomplete" caveat does fire, so this is a soft loss rather than a silent one — but recording the header fault and *falling through* to the record scan (rather than `continue`) keeps both facts, and `alreadyFound` already excludes anything in `unscannablePlugins` from the declarer list, so the declarer category stays honest either way.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6297,9 +6346,13 @@ public WritePatchBuilder.MergeOutcome MergePlugins(
         catch (Exception ex) { return WritePatchBuilder.MergeOutcome.Fail($"'{outName}' is not a valid plugin filename ({ex.Message})."); }
         if (outKey.Type == ModType.Light)
             return WritePatchBuilder.MergeOutcome.Fail(
+                // The reason is what the merge does NOT do, not a claim about where the donors' ids are. A single
+                // already-light donor's ids are all inside the window by definition, so "a merge keeps the ids in the
+                // full range" was false on exactly the path a rename made ordinary.
                 $"refused — '{outName}' has the .esl extension, which the game engine force-treats as a LIGHT master regardless " +
-                "of the header flag, but a merge keeps the donors' object ids in the full range (ids above 0xFFF would be misread " +
-                "in game). Merge to a '.esp' instead, then run " + ToolNames.CompactPlugin + " on it to make it light (the tools compose). Nothing was written.");
+                "of the header flag, but a merge does not renumber into the light range: it keeps each donor's object ids where " +
```
Avick3110: **The replacement reason is false on a different path than the one it fixed.**

"it keeps each donor's object ids where they already are" is not what `BuildMergeRemap` does. Pass 1 keeps an id only when it is `>= floor`, `<= ceiling` and unclaimed; everything else queues as a collision, and pass 2 allocates those from `floor` (`RemapEngine.EslFloor` = 0x800) upward. `RenderMerge`'s own per-donor accounting prints "N renumbered (id collisions / below-floor)" a few lines later, so the refusal contradicts the report the same tool writes on success.

Scenario: a user merges two donors that both originate 0x000800 into `Out.esl`. They are told the merge keeps each donor's ids where they already are; merged to `.esp` instead, the second donor's 0x800 would have been renumbered. Same for any donor with ids below 0x800.

"a merge does not renumber into the light range" is loose for the same reason — collision allocation starts *at* 0x800, which is inside the light window; what it does not do is *bound* ids to it. The claim that holds on every path is that a merge never constrains object ids to the light window (`compact_plugin` does), which is also what the second half of the sentence already says.

Avick3110:
Folded all three findings.

**1. `LocalizedStrings.Assess` dropped `BsaEmbedding`'s folder-unlistable answer** (cd3f0f5). A localized plugin in a mod folder that will not list, with its tables in a `.bsa` there, fell through to `Nowhere` and was refused with "no archive beside it" plus the wrong remedy, while the read side's gate answered true for the same folder. There is now a `ModFolderUnreadable` shape with its own arm at every render seam: `WhereTheTextIs` says the folder the plugin sits in could not be read, `WhyNotInPlace` names the hazard it actually carries, and `UnresolvableStringsRefusal`'s remedy points at freeing or fixing permissions on that folder, named. It still refuses (fail-closed) and `ResolvesNowhere` still covers it, but it claims no absence. New unit test `LocalizedModFolderUnreadableTests` — a deny-ACE on the mod folder with a real `.bsa` carrying the tables — asserts the shape, the refusal text, the remedy text, and that the read-side gate and the classifier agree a source may exist; verified failing against the old classifier. Probe arm `localized-write-guard :: a MOD folder houseCARL cannot LIST`, both directions.

**2. `RemapEngine` skipped the record walk after a header-read fault** (fc0ac21). The `continue` is gone: the fault is recorded in `UnscannablePlugins` and the walk falls through, so merge no longer reports "external referencers: none" for a plugin whose records it never looked at. A plugin already named by the header fault is not named (or counted) a second time by the record catches, and it is not counted as scanned through.

One disagreement, on the case rather than the fix. **The corrupt-TES4-master-table scenario does not reach that arm, and merge and compact do not diverge for it.** Measured: Mutagen parses the master list while OPENING a plugin, so every master-table corruption I could build (MAST length overrun, zero length, unterminated name, wrong length) throws at `CreateFromBinaryOverlay`; the resolver excludes the plugin at build and the identify pass skips it before either header branch, identically with `readDeclaredMasters` true and false. The same holds for a held-open file — `RecordsIn` re-opens, so it fails exactly as the header read does. What is left for the `EnumerationFault` arm is a fault the header read meets AFTER a successful open, which I could not fixture. So the test pins what actually happens rather than a scenario that cannot occur: `ACorruptMasterTableReadsTheSameForBothCallers` asserts the plugin is excluded at build and that the two callers return the same referencers, unscannable count and scanned count. The code change stays — it is correct for any post-open header fault and costs nothing — with the measurement recorded in the comment.

**3. The `.esl` merge refusal stated a false reason** (2017ec2). "It keeps each donor's object ids where they already are" is false whenever `BuildMergeRemap` renumbers a collision or a below-floor id from 0x800 up, a count `RenderMerge` prints. It now says a merge never constrains object ids to the light window and renumbers only what it must, and points at `compact_plugin`, which does. The merge-service-guard arm pins the new claim and the absence of the old one, and the changelog entry (which had already overshot) is corrected.

Drive-by in cd3f0f5: `LocalizedStrings.cs` carried a literal NUL byte where the archive cache key's separator space belongs, which made git treat the file as binary and print no diffs for it. Replaced with the space; behaviour unchanged.

Build clean, 1076 tests pass (`tier!=bridge`), ci-all 127/127.

PR #541, opened 2026-09-04
Run every step from the worktree; temp files in the scratchpad

Avick3110:
Two sentences in CLAUDE.md, rewritten after review. Step 1: run every command from inside the worktree, because a build, test, or review run from the primary checkout reads the wrong tree. That is the cause behind two sessions today reviewing the wrong branch, and it would have had them building and testing the wrong tree too. Step 3: temp files go in the session scratchpad; the shared temp directory, under any spelling, is where two parallel lanes overwrote each other's PR body today. The review sequence itself is unchanged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@ Code comments call the first two of these the cornerstones. A design question th
 
 ## How to work
 
-1. **Branch first.** `main` is read-only. Work in a worktree at `.claude/worktrees/<name>` on branch `claude/<name>`. Check `git branch --show-current` before the first edit.
+1. **Branch first.** `main` is read-only. Work in a worktree at `.claude/worktrees/<name>` on branch `claude/<name>`, and run every command from inside it: a build, test, or review run from the primary checkout reads `main`, not your branch. Check `git branch --show-current` before the first edit.
 2. **Build and test.**
    ```
    dotnet build housecarl.sln -c Release
@@ -25,7 +25,7 @@ Code comments call the first two of these the cornerstones. A design question th
    dotnet src/housecarl-generator/bin/Release/net9.0/housecarl-generator.dll ci-all
    ```
    Tests drive the built server. How to write one: [standards/TESTING.md](standards/TESTING.md).
-3. **Commit small.** One change per commit, plain imperative subject under 72 characters.
+3. **Commit small.** One change per commit, plain imperative subject under 72 characters. Temp files go in the session's own scratchpad directory; the shared temp directory, under any spelling, is where parallel sessions overwrite each other.
 4. **Open the PR.** A paragraph saying what changed and why, `Closes #N` for the issue, and a line under `## Unreleased` in `plugin/CHANGELOG.md` if a user would notice the change. Run one review pass (`/code-review`) and fix what is real. Aaron then reviews; his findings come from a fresh agent reading the PR blind, so a finding can be wrong. Fold what is real; refute the rest in a reply on the PR with the reason, and Aaron decides. He merges on his word with `gh pr merge <N> --rebase --delete-branch`; then remove the worktree.
 5. **Write plainly.** Comments are one line saying what. Commit messages, PR bodies, and issues use ordinary words: bug, test, review, fix. No project jargon, no lore.
 


CLAUDE.md
```diff
@@ -25,8 +25,8 @@ Code comments call the first two of these the cornerstones. A design question th
...
+3. **Commit small.** One change per commit, plain imperative subject under 72 characters. Temp files go in the session's own scratchpad directory, never `/tmp`, which parallel sessions share.
+4. **Open the PR.** A paragraph saying what changed and why, `Closes #N` for the issue, and a line under `## Unreleased` in `plugin/CHANGELOG.md` if a user would notice the change. Run one review pass (`/code-review <branch>`, naming the branch, since the skill reads the main checkout by default) and fix what is real. Aaron then reviews; his findings come from a fresh agent reading the PR blind, so a finding can be wrong. Fold what is real; refute the rest in a reply on the PR with the reason, and Aaron decides. He merges on his word with `gh pr merge <N> --rebase --delete-branch`; then remove the worktree.
```
Avick3110: **"since the skill reads the main checkout by default" is not what the skill does.**

The review skill derives its diff from the session's working directory — it runs `git diff @{upstream}...HEAD` (falling back to `main...HEAD` / `HEAD~1`) in the cwd. A session that followed step 1 and has its cwd in `.claude/worktrees/<name>` gets exactly the right diff from a bare `/code-review`.

The two bad reviews today were sessions whose cwd stayed at the primary checkout while `git worktree add` created the tree elsewhere. Writing that up as skill behaviour teaches the next session the wrong model: it keeps working from the primary checkout, believes the branch argument makes that safe, and carries the same mistake into every other step.

Suggest naming the actual condition — the review reads whatever tree the session is standing in — rather than a property of the skill.

Avick3110: **Naming the branch drops the working-tree pass.**

With no argument the skill also reviews uncommitted work: "If there are uncommitted changes, or the range diff is empty, also run `git diff HEAD` and include the working-tree changes in scope — the review often runs before the commit." Passing `<branch>` scopes it to committed refs and that pass goes away.

Concretely: session commits, opens the PR, makes an uncommitted fixup in the worktree, then runs the review pass as step 4 instructs. `/code-review claude/<name>` reads the pushed ref and never sees the edit — and run from the primary checkout it cannot see the worktree's dirty state at all. Aaron's blind pass reads the PR too, so that hunk gets no review from either.

A short caveat that the branch form only covers committed work would cover it.

Avick3110: **This fixes the visible symptom; the same cause breaks step 2 harder.**

The sentence is premised on a session whose cwd is the primary checkout. If that is true, step 2 already ran against the wrong tree: `dotnet build housecarl.sln` and `dotnet test` compiled and tested `main`'s unchanged code and passed. Step 4's `/code-review claude/<name>` then reads the correct diff and reports it clean, so the PR is opened claiming a green build for code that was never built. That is worse than a mis-scoped review and this change does not touch it.

This is also the pattern the file's own "Don't" list rules out — "Don't add a guard, sweep, or process rule to catch a mistake. Fix the mistake; if it recurs, fix the code that allows it." The mistake is the session not standing in the worktree. Fixing that (cd into it before the first command, or use the worktree tool) makes every step in the list correct by construction, including the review, and needs no argument to remember.

CLAUDE.md
```diff
@@ -25,8 +25,8 @@ Code comments call the first two of these the cornerstones. A design question th
    dotnet src/housecarl-generator/bin/Release/net9.0/housecarl-generator.dll ci-all
    ```
    Tests drive the built server. How to write one: [standards/TESTING.md](standards/TESTING.md).
-3. **Commit small.** One change per commit, plain imperative subject under 72 characters.
-4. **Open the PR.** A paragraph saying what changed and why, `Closes #N` for the issue, and a line under `## Unreleased` in `plugin/CHANGELOG.md` if a user would notice the change. Run one review pass (`/code-review`) and fix what is real. Aaron then reviews; his findings come from a fresh agent reading the PR blind, so a finding can be wrong. Fold what is real; refute the rest in a reply on the PR with the reason, and Aaron decides. He merges on his word with `gh pr merge <N> --rebase --delete-branch`; then remove the worktree.
+3. **Commit small.** One change per commit, plain imperative subject under 72 characters. Temp files go in the session's own scratchpad directory, never `/tmp`, which parallel sessions share.
```
Avick3110: **Banning `/tmp` bans one spelling of a directory with several.**

Verified on this machine: `/tmp`, `%TEMP%`, `$TEMP` and Python's `tempfile.gettempdir()` all resolve to `C:\Users\compl\AppData\Local\Temp`. A session that writes `%TEMP%\pr_body.md`, or runs any script calling `gettempdir()`, obeys this sentence to the letter and still overwrites the parallel lane's PR body — the exact failure it exists to prevent.

The hazard is also the fixed filename, not the directory: `tempfile.mkstemp` in that same shared dir never collides, while a fixed name collides anywhere two lanes both write it.

Pointing the rule at the scratchpad path the environment hands each session, rather than at the `/tmp` spelling, closes both holes.

Avick3110:
Folded. Findings 1 and 3 named the real cause, so the branch-argument sentence is gone and step 1 now says to run every command from inside the worktree; that covers build and test as well as the review. Finding 2 falls away with it. Finding 4: the sentence now names the scratchpad and says the shared temp directory under any spelling is the hazard, rather than banning one spelling.

PR #543, opened 2026-09-04
Fix the housecarl_records remedies and refusals

### LINKED ISSUE #442: [Bug] housecarl_records summary rows label override depth 'depth=', colliding with project.depth=
`housecarl_records` summary rows label the override depth `depth=`, colliding with the
tool's own `project.depth=` expansion knob. Every other renderer of the same fact —
both transports, and the artifact schemas — spells it `override_depth`.

## Where

`RecordsTools.cs:1856` (`RenderRecordsSummary`, the default summary form, text lane):

```
sb.Append("  winner=").Append(o.WinnerPlugin).Append("  depth=").Append(o.OverrideDepth);
```

renders a row like:

```
000D2B:Skyrim.esm  Weapon  IronSword  source=Requiem.esp  winner=Requiem.esp  depth=2
```

Every sibling says `override_depth`:

| Site | Lane | Spelling |
|---|---|---|
| `ReadTools.cs:1848` (`AppendRecord`) | text | `override_depth=` |
| `ReadTools.cs:1094` (`RenderCrossQuery` summary row) | text | `override_depth=` |
| `JsonWire.cs:366` / `:537` / `:1254` | json | `"override_depth"` |
| `Artifacts.cs:153` / `:179` / `:238` | artifact schema | `override_depth` |

So the odd one out is the row a caller sees FIRST — summary is the default form.

## Why it bites

`depth` on this tool already means something else: `project.depth=` is the field-content
expansion knob, and `walk.depth=` is the hop budget. A row reading `depth=2` invites the
reader to think the call expanded to depth 2, when it is reporting how many plugins deep
the winning override sits. The two facts are unrelated and the label is the only thing
distinguishing them.

Not the #439 class — this is not a remedy (it predicts no later call), which is why it
was not folded into that fix. It is a display-label inconsistency, one word wide.

## Suggested fix

`depth=` → `override_depth=` at `RecordsTools.cs:1856`, matching every other renderer.
Text lane only; no json or artifact change (they are already right).

Found by the derived lever-reachability sweep while building the #439 fix (PR B of
RUN_ORDER row 6) — the site is outside that branch's change, so it is filed rather than
folded. Unrouted and unmilestoned: routing is Aaron's stroke.

### LINKED ISSUE #443: [Bug] housecarl_records: the group_by=type aggregate refusal names 1.x levers (group_by=, type=)
A `housecarl_records` call can reach a refusal composed in the shared scan engine that
names two levers in the 1.x spelling. Same class as #439 (a sentence telling the caller to
pass something this tool does not accept), reached through a seam that fix does not cover.

## Repro

Driven against the records-guard fixture:

```
housecarl_records(conflicts_only: true,
                  project: { form: "aggregate", group_by: "type" })
```

returns:

```
records  form=aggregate  source=winner
error: group_by=type needs each match's type, which requires a body-bearing scope —
add type= or plugins= (winner/defined_in group without a body).
```

## What is wrong

Two levers, both in the 1.x spelling:

| Named | `housecarl_records` spells it |
|---|---|
| `group_by=` | `project.group_by` (form-scoped; `project.group_by` is refused elsewhere by name) |
| `add type=` | `types=` (plural) |

`plugins=` is correct on both generations.

## Where

`LoadOrderService.cs:4029`, in `CrossQuery` — the scan engine shared by `housecarl_records`
and `housecarl_cross_plugin_query`. The engine composes its refusals from string literals
and takes no lever vocabulary from its caller at all, so a sentence composed there can only
speak one generation's spelling.

## Why most of its siblings are fine

`housecarl_records` pre-validates and refuses first, in its own correct spelling, for nearly
every neighbouring engine refusal — the no-scope case (`RecordsTools.cs:344`), `offset=` with
the aggregate form (`:375`), the unknown-`group_by` key (`:239`), and the body-scan-needs-a-
bound case (measured: a `where=`-only call returns the records-spelled *"must be combined
with types=, plugins=, or a formids= set"*, not the engine's `type=` version).

This one has no records-side pre-check, so the engine's sentence is what the caller sees.
Measured, not derived — a `where=`-only probe and an `aggregate`+`group_by=type` probe were
driven and their responses read.

## Suggested fix

Either a records-side pre-check in the tool's own spelling, matching how the sibling cases
are already handled, or give the engine the caller's lever vocabulary. The first matches
existing practice and is one condition; the second is the structural direction and is
covered by the no-default-carrier issue filed alongside this one.

## Provenance

Found by the re-derived lever-reachability sweep during the round-1 wave of PR B of
RUN_ORDER row 6 (#439). The line is outside that branch's change and the branch's guard
does not reach it, so it is filed rather than folded — the #324/#442 shape.

Unrouted and unmilestoned: routing is Aaron's stroke.

### LINKED ISSUE #445: [Bug] housecarl_records: the everything form's remedy names project.fields=, which that form refuses
The lever-vocabulary carrier (#439) is per-TOOL. `housecarl_records`' levers are per-FORM.
So a spelling that is correct for the tool can still be wrong for the call.

## Repro

```
housecarl_records(formids: [3 WEAP], project: { form: "everything" }, max_chars: 220)
```

text:
```
... [truncated: showing 2 of 41 field lines at max_chars=220; narrow with project.fields=,
     lower project.depth=, or raise max_chars]
... [truncated: rendered 1 of 3 records before hitting max_chars=220; request fewer formids,
     pass project.fields= to slim each, or raise max_chars]
```
json: `"[truncated at max_chars: 0 of 41 fields shown; narrow with project.fields=, lower project.depth=, …]"`

Following it — `project: { form: "everything", fields: ["EditorID"] }`:
```
error: project.fields belongs to the 'fields'/'delta'/'tree' forms (got form='everything').
Set project.form, or drop fields.
```

Same on the scan lane (`types=["WEAP"], project={"form":"everything"}`).

## Why

`RecordsTools.cs:588-589` (list lane) and `:1224-1225` (everything-on-a-scan) pass
`LeverNames.Records` for every non-summary form. `RecordsTools.cs:213-214` refuses the selector
on `everything`, and `projFields` is nulled for that form at `:243` — the form genuinely has no
field selector. `lv.Depth` is honest there (`project.depth` is legal on `fields`/`everything`);
`lv.Fields` is not.

Exposure is text and json: `dense` + `everything` is refused up front (`RecordsTools.cs:356`),
and artifact rows are written at `int.MaxValue` so the notice is unreachable there.

Note this is **not** a missed seam — the seam is threaded and passes a vocabulary that is wrong
for the form. Threading more, or requiring the argument (#444), does not fix it.

## The shape of the fix

The vocabulary is a function of **(tool, form)**, not **(tool)**. Either `LeverNames` gains a
form dimension and each call site passes the form it is rendering, or the `everything` form gets
its own carrier value whose `Fields` names no selector at all (the form has none, so the honest
remedy is `lower project.depth=` and `raise max_chars` only).

## Provenance

Found independently by all three round-2 reviewers on PR B of RUN_ORDER row 6 (#439), each
driving the call. It is the reason that branch ships a scoped claim naming `fields`/`tree`/`delta`
rather than a completeness claim. Not a regression: on `main` the same call says `narrow with
fields=`, which refuses as an unknown parameter.

Unrouted and unmilestoned: routing is Aaron's stroke.

### LINKED ISSUE #446: [Bug] housecarl_records walk/chain: 'raise limit=' is a no-op remedy for the walk.max_nodes bound

### LINKED ISSUE #460: [Bug] record-not-found refusal does not say whether the PLUGIN resolved — one sentence for two opposite causes
> *Filed from an external live-session gap report by **Jouten** (running 1.9.0), with their OK to be named; triaged, root-caused against source, and posted by AI. Machine paths and third-party details scrubbed.*

### What happened

Resolving a RaceMenu preset's head-part reference (`0086E4` in `HG Hairdos 2.esp`) to find which hair style a jslot uses:

    housecarl_read_record formid="0086E4:HG Hairdos 2.esp"
    -> error: FormID 0086E4:HG Hairdos 2.esp is not present in the load order (437 plugins).

The plugin **was** active — confirmed by `load_order_status` the same minute (`as a plugin: ACTIVE (checked in plugins.txt)`). The wording reads equally as *"plugin unknown"* and *"record absent from a known plugin"*, so the actual cause took a second tool call to establish: the installed copy is the **ESL-flagged edition** whose records are compacted to 0x800+, so `0086E4` cannot exist in it.

Diagnosis came from `housecarl_cross_plugin_query type=HeadPart plugins=["HG Hairdos 2.esp"] defined_in=true`, which enumerated records at 000802+ — proving the compacted edition.

### Frequency

Every cross-edition FormID mismatch: preset debugging, patch auditing against a different release of the same mod. ESP vs ESL-flagged editions are common for hair and asset packs, and any jslot/preset/patch that stores FormIDs from a different *edition* of a mod lands here.

### Ideal shape (reporter's)

When the FormID's plugin IS in the load order, say so and characterize the miss — e.g. `plugin is ACTIVE but defines no record 0086E4 (its defined FormIDs span 000800-0008xx -- ESL-compacted?)`. The range alone would have made the edition mismatch one read instead of two.

- date: 2026-08-25  |  houseCARL version: 1.9.0

---

Valid, reproducible, non-duplicate. Real and unchanged on `main` @ `94bf378`. The string dates to `9fe4d80` (1.0.0); the 2.0 W2/W3 commits added *copies*, not variants.

**Everything needed to disambiguate is already in hand at the emit site.** `ResolveRead` (`LoadOrderService.cs:2347-2355`) calls `view.ContainsPlugin(plugin)` twenty lines earlier at `:2317`, and the absence-explainer apparatus (`view.ExplainAbsence` / `AbsenceClause` / `NameSuggestion`, `LoadOrderResolver.cs:646-660`) is already wired for the *named-plugin* arm. The winner-null arm checks only `ExcludedPlugins` and then falls through to the bare sentence — it never asks whether `fk.ModKey.FileName` is in the order, which is an O(1) dictionary hit.

Emitters of the same conflating sentence:

| Site | Path |
|---|---|
| `LoadOrderService.cs:2354` | `ResolveRead` — **the reporter's path** (`read_record`, `batch_record_detail`, and 2.0 `housecarl_records` via `ResolveBatch` / `ResolveBatchFromPole`) |
| `LoadOrderService.cs:1034` | `ReplaySkyPatcher` (SkyPatcher post-state) |
| `LoadOrderService.cs:3487` | SkyPatcher overlay batch |
| `LoadOrderService.cs:2559` | `ResolveSummary` (`resolve`, effect_chain summaries) |
| `LoadOrderService.cs:3189, 3311, 3890` | diff poles, SkyPatcher pre-pole, info_order rows |
| `ReadTools.cs:823` | `RenderResolve` default row error |

**We already draw exactly this distinction elsewhere.** The SKSE config audit separates `SkseRefVerdict.PluginMissing` from `Dangling` and renders them as `[MISSING]` vs `DANGLING — plugin present but no such record (a dead reference)` (`SkseTools.cs:605-607`, `:664-665`, `:764-765`). The record read has the same two states and collapses them.

What is available at the failure point:

- **Plugin presence** — knowable, O(1) (`IndexView.ContainsPlugin`).
- **ESL / light flag** — not currently read for an active plugin (`IsSmallMaster` appears only in write and probe code), but cheaply obtainable via `IndexView.PluginPath` + a header-only overlay open, the same trick `DeclaredMasters` already uses. No tool reports it today — the same shape as open issue **#376** (localized header flag unreported).
- **Defined FormID range** — not precomputed. The index is `Dictionary<FormKey,(winner,count)>`, so a per-plugin span costs a full scan today; a per-plugin min/max could be accumulated at build time for free, since the build already enumerates every record per plugin.

### Second defect in the same lines — the ordering bug

Because `:2347` runs **before** the good `does not touch … Touched by:` message at `:2360-2371`, even an **explicitly named, present plugin** (`plugin=` / 2.0 `source=`) gets the ambiguous sentence when the FormID resolves nowhere. The informative message only ever renders when the record exists somewhere else in the order. This is folded into this issue rather than filed separately because it is the same edit at the same lines.

### 2.0 does not close it — and inherits it

`housecarl_records` calls `svc.ResolveBatch` / `ResolveBatchFromPole` (`RecordsTools.cs:542`, `:550`, `:1155-1192`), both funnelling into the same `ResolveRead:2354`. SPEC §4.2 covers only the adjacent case — *"a record the pole does not touch is a per-item refusal naming the actual touchers"* (`SPEC.md:602`), which is the record-exists-elsewhere branch that is already good. The resolves-nowhere branch appears nowhere in `SPEC.md`, `CHARTER.md`, `RUN_ORDER.md`, or row 6's shard.

Row 6 — the READ-surface response-layer / refusal-grammar row — is **DONE** (PR #441 `Fixes #403`, PR #450 `Fixes #439`), and its scope was explicitly the *json shape* and *lever vocabulary*, not cause discrimination: #403 was "RenderError lacks the ok discriminant", #439 was shared renderers "speaking one lane's lever vocabulary". Neither touches what a refusal **knows**. `RUN_ORDER.md:145-149` rule 10(d) is a wiring rule, not a mandate to state a diagnosis.

Disposition under rule 10(b): **engine-survives**. `ResolveRead` is not tool-body code, so this is fixed normally — it is not closed by the 1.x demolition.

### Class, with a worse sibling

The shape is **"the refusal states a negative the tool could have qualified"**. Worst offenders:

1. **`ReadTools.cs:1879`** (`LinkText`, used by `resolve_names` and the dense render) — `"unresolved: target not in the active order"`. Worse than the reported one: it does not merely omit the plugin state, it **asserts** one that is frequently wrong — the plugin usually *is* in the order; the record is not. The correct sibling wording already exists at `LoadOrderService.cs:3766`: *"unresolved — no active plugin defines this target (a missing endpoint)"*.
2. The seven copies of the bare winner-null sentence above, none consulting `ContainsPlugin`.
3. **`ReadTools.cs:823`** — `resolve`'s row fallback, i.e. **the reporter's own workaround tool answers with the same ambiguity**.
4. **`ScriptPropertyCheck.cs:415`** — *"'{rel}' is not on disk (the script is not compiled, or not in the load order)"* — an explicit hand-written conflation of two causes in one parenthesis.

No open or closed issue covers this. Prior error/remedy work: #403, #439 (row 6, both closed — json refusal grammar and lever-vocabulary threading), #271 (closed — a plugin state label that misreported ticked-ness; the `AbsenceClause` machinery a fix here would reuse), #344 / #361 (closed — `check_errors` response layer). Still-open remedy-quality family: #359, #377, #380, #319, #408, plus row-6 residue #442-#449 and #440. Adjacent but distinct: **#376** (no tool reports a plugin's localized header flag — same "header flag unreported" shape as the ESL flag needed here).

### Fix direction

Add a `ContainsPlugin(fk.ModKey.FileName)` arm to the winner-null branch at `LoadOrderService.cs:2347-2355`, reusing the `ExplainAbsence` / `AbsenceClause` pattern already used twenty lines above, so the sentence becomes *"the plugin IS active but defines no record 0086E4"* vs *"the plugin is not in the order, because …"*. The light-flag read (`PluginPath` + header overlay, `IsSmallMaster`) is the optional second clause that names the reporter's actual cause. Fix the ordering bug in the same edit. Fold in the sibling sites — `ReadTools.cs:1879` and `:823` at minimum, since `resolve` and `resolve_names` are the tools a caller reaches for *after* the read fails.

The defined-FormID-range ask is the expensive half (needs a per-plugin min/max accumulated at index build) and is a stretch, not the fix.

Routing and milestone are Aaron's call; the row-6 residue bugs it would sit naturally beside are milestoned houseCARL 2.0.

### LINKED ISSUE #461: [Bug] a mid-path list hop in a where= predicate is misdiagnosed as a mistyped field, discarding the bracket hint
> *Found while triaging an external gap report from **Jouten**; verified against source and posted by AI. Not the reported gap itself — this is a defect noticed in the code path the report exercised, filed separately so it is not blocked on the larger capability plan it sits next to.*

### The defect

A `where=` predicate whose path hops **through** a list mid-path gets the *"you mistyped the field name"* refusal, discarding the bracket hint the per-candidate walk actually generated. The result is backwards: **the deeper and more specific the wrong path, the worse the advice.**

Two paths, same underlying mistake, opposite quality of answer:

| Predicate path | Classified as | Message the caller gets |
|---|---|---|
| `Conditions` (leaf container) | `EvalKind.Container` | *"it resolves to a container/list here, not a scalar leaf; filter on a scalar sub-path (e.g. `Conditions[0]` …)"* — **correct and actionable** |
| `Conditions.Data.Function` (hop through the list) | `EvalKind.NoField` | *"it is NOT A FIELD on these records (a mistyped path, or a field that doesn't exist on this record type); check the field name against the record's schema"* — **misdiagnosis** |

The second is what a user hits in practice, because nobody filters on a bare list — they filter on something inside it. `Conditions.Data.Function` is not a mistyped field name and is not absent from the record type; it is a correct field reached by a path missing its brackets. Sending the caller to "check the field name against the record's schema" points them away from the fix.

### Root cause

The good hint exists and is thrown away at the rollup.

- The per-candidate walk already produces it: `ReadEngine.NoFieldNote` (`ReadEngine.cs:356-367`) detects that the owner is a list/dict and emits *"'Conditions' is a list/dict — index an element with brackets, e.g. 'Conditions[Data]', not 'Conditions.Data'"* — described in its own doc comment as the read analog of the write pre-flight's bracket hint.
- But a mid-path list hop is bucketed as `EvalKind.NoField` (`FieldPredicate.cs:507`), not `Container` — `Container` is only reached when the **leaf** is a container.
- So the rollup takes the `_noField[k] == _scanned` branch (`FieldPredicate.cs:827-829`) and renders the generic mistyped-path sentence. The `_container[k] == _scanned` branch at `:830-832` — the one carrying the bracket advice — is unreachable for this shape.

The counters are otherwise well-built (four distinct causes, each with its own actionable reason, all preserving the loud "yielded no readable value on any of" marker). This is one bucket resolving to the wrong arm, not a design problem with the rollup.

### Fix direction

Either propagate the list-hop signal so it lands in a bracket-advising arm, or thread `NoFieldNote`'s already-computed hint through to the rollup so the `NoField` branch can render the specific reason when every miss was a collection hop. The second keeps one home for the wording (`ReadEngine`) rather than restating it in `FieldPredicate`.

Worth a probe case either way — `ValuePredicateProbe` currently has no coverage for a mid-path collection hop.

### Second item, same surface — the deferral is invisible to callers

Related and equally cheap: list wildcards are a known-unbuilt capability (`FieldPredicate.cs:38`, *"A wildcard over a list … is a deliberate future extension"*), but that note lives **only in an internal doc comment**. Nothing in the tool description says so, so a caller has no way to learn the capability does not exist and burns calls discovering it empirically — which is exactly what happened in the report this was found in (~14 calls to establish one fact).

houseCARL's own convention is to name known-unbuilt capabilities in the tool description: the reverse-reference index does precisely this (`RecordsTools.cs:113`, `:947`, `LoadOrderService.cs:3973` — *"a known future capability"*). Applying the same treatment to the list wildcard is a one-line description edit and a small §5 #7 (no silent failure) recovery.

Both items are read-side refusal quality on the same surface, hence one issue. Neither depends on the wildcard capability itself landing — that is **#458**, a separate, larger, plan-first piece of work.

Same class as **#460** (a record-not-found refusal that states a negative it could have qualified); this is another instance of it on the predicate surface.

### LINKED ISSUE #469: [Bug] the @file lane loses its file identity in two of four ListParams refusals
`ListParams.Read<T>` is the one reader behind every list-valued input that accepts the `@file` convention — `apply`'s `ops=` / `bundle=` / `assignments=`, `create`'s `records=`. It composes four post-read refusals. Two of them name the file lane; two do not, and the split does not track anything.

`src/housecarl-mcp/ListParams.cs:72-78`:

| refusal | line | spelling used | what a caller who passed `ops="@C:/jobs/x.json"` reads |
|---|---|---|---|
| parse error | :72 | `origin` | `the file named by ops could not be parsed at line 1, byte 13 in the line (element $[0].formid): …` |
| parsed to JSON null | :75 | `origin` | `the file named by ops parsed to JSON null — expected a JSON array of …` |
| **empty array** | :76 | `param` | **`ops is an empty array — give at least one {formid, field_path, …}`** |
| **null element** | :78 | `param` | **`ops: element [0] is null — every element must be an object.`** |

`origin` is set to `$"the file named by {param}"` on the @file lane (`:34`, `:55`) and to the bare `param` on the inline lane (`:57`). So for the last two rows a caller who supplied no inline array at all is told the fault is in `ops`, with nothing saying a file was read.

**Why it matters, stated at its real size.** This is not ambiguity — `ListParams` refuses a mixed inline/`@file` array by name (`:51`), so at most one file is ever in play and the caller knows which path they typed. It is that two of four adjacent refusals silently drop the fact that the content came from disk, which is the kind of thing that sends someone looking at their call instead of their file. The inline lane reads correctly in all four.

**Not a regression introduced by anything.** Both spellings have been there since the `@file` convention landed. It surfaced during the #468 demolition catch-up: `bulk_apply`'s `from_file=` lane named the file in its refusals (`dry-run-guard` asserted `.Contains("bad.json")` and `.Contains("empty.json")`), and when those assertions were repointed onto `apply`'s `ops="@path"` they failed — the 2.0 twin makes a weaker claim than the 1.x one it replaces.

**Deliberately not fixed in that PR** (advisor ruling, 2026-09-01): it is surviving-code behaviour the deletion did not cause — the #324 file-don't-fold shape. The catch-up PR's sentence-repair scope covers sentences naming a *dead* tool, which these are not. The two repointed assertions there state today's behaviour with this issue's number in the comment, so the accommodation is visible rather than baked in.

**The fix is not assumed to be `param` → `origin`.** That is the obvious two-token change, but CLAUDE.md §5 #11 says a finding about a user-facing sentence gets a probe, not a rewording — what these sentences should say is that PR's measurement. Whatever lands has to hold across every `ListParams.Read` caller (`apply` ×3 parameters, `create` ×1), both lanes, since the same four refusals serve inline and file callers alike.

---

**Sibling note, same area, no action implied.** `ApplyOp`/`BulkOp` and `CreateRecordSpec`/`CreateOp` can now collapse into one type each. `ApplyOp`'s own summary says it exists as a separate type "so the 1.x tools' published schemas stay untouched through the build waves" — the reason expired when `bulk_apply`/`bulk_create` were deleted in #468. Both `BulkOp` and `CreateOp` are now internal service shapes with vestigial `[JsonPropertyName]` attributes, parked in `WireNamesProbe.NonInputWireTypes` with that reason recorded. Collapsing them is a reshape of live 2.0 types, not a deletion, so it wants its own change.

Avick3110:
Seven wording and remedy bugs on the `housecarl_records` read surface and the refusal paths behind it. Each one is a sentence a caller reads, so each is driven and asserted rather than derived: a summary row's own label, the aggregate scope pre-check, the field selector a form does not have, the walk's carrier bound, the two refusals that stated a negative they could have qualified, and the `@file` lane's identity. The one structural piece is `LeverNames`: the lever vocabulary a remedy may name turns out to be a function of (tool, form), not of the tool alone, so it gains a `HasFieldSelector` axis and the `everything` form drops the "narrow with `project.fields=`" clause instead of naming a lever that form refuses.

While in the file: `project.form='chain'`'s description still said the traversal form was refused by name until it landed. It landed — accepted in the form switch and rendered by the chain lane — so the description now says what the form is.

Also folded from the review pass: the off-order scan's `everything` lane kept the tool-wide vocabulary (one of three lanes missed); `LinkText` asserted that no plugin defines a target in the case where the ref itself names the winner that does; a presence operator (`exists`/`missing`) on a list-hop path bypassed the new classification and got the old advice; the bracket advice now needs one list hop rather than all of them, so a mixed-type scan keeps it; and the two SkyPatcher lanes still emitted the undifferentiated not-present sentence, which is now one shared helper.

Tests: 1072 pass (`tier!=bridge`, 14 new), `ci-all` 127/127. `dry-run-guard`'s two `#469` accommodation assertions are tightened onto the file identity, and a third arm asserts the inline lane still names the bare parameter.

Closes #442
Closes #443
Closes #445
Closes #446
Closes #460
Closes #461
Closes #469


Avick3110:
All six folded. Build, `tier!=bridge` (1087 pass, 0 fail) and `ci-all` (127/127) green.

| # | Finding | Commit |
|---|---|---|
| 1 | ESL clause asserted unconditionally on the present-plugin arm | `b775472` |
| 6 | `AbsenceClause` re-implemented inline; spelling hint over a real cause | `b775472` |
| 2 | Remaining emitters composed their own "not present in the active order" | `1c3e992` |
| 3 | Bracket arm asserted a missing bracket without checking the element type | `ec2699b` |
| 4 | `group_by=type` pre-check fired on the off-order lane with a false reason | `6907c8d` |
| 5 | `RenderEffectChain`'s carrier bound defaulted to `limit=` | `cb16efb` |

Changelog reworded to match in `2648a72`.

**1 + 6.** `IndexView` now exposes `IsLightFlagged` (the same per-plugin flag `RuntimeAddressFor` composes runtime addresses from). The compaction clause is stated only when the index says the plugin is light; otherwise the sentence states that the plugin defines no such record and names the `defined_in` call. The absent arm uses a new `AbsenceClause(name, out cause)` overload — one wording, one explainer invocation — and drops the filename hint when a cause follows. `AMissingRecordInAPresentPluginSaysThePluginIsThere` no longer pins the compaction cause against the fixture's non-ESL master; the light-flagged case (and its unflagged neighbour) is driven in `RuntimeFormIdTests`, whose world already has an ESL — the records fixture has no light plugin and adding one would have moved every runtime index in it.

**2.** All five lanes plus the identity form route through `UnresolvedFormId`, one test each driving the same ghost FormID on a present plugin. Two things came out of doing it:

- The walk seed's `Fetch` returns null for two conditions. Only "no winner" is the unresolved case; a named winner whose body did not come back keeps the fetch sentence, or the seed would be told its plugin defines no such record when it does.
- The identity row carries the reason as its `Error`, so `ResolvedRef.Error` now means "the reason", not "the input was malformed" — doc updated, and the two wave2 probe arms that asserted `Error is null` on an absent form now assert it is present and is not the malformed-input one. The `resolve_names` link annotation is unaffected: it renders from `Resolved`/`Winner`. The explainer is memoised per plugin so a batch of dangling refs into one missing plugin does not pay for a profile parse per row.

**3.** `ReadEngine.ListHopRemedy` resolves the trailing segment against the collection's element type: a real field gets `Effects[0].Data`, anything else is named as not a field on that element type with the nearest real field if there is one, and a numeric segment keeps the `.0`-vs-`[0]` advice. The predicate rollup quotes that remedy rather than composing a second copy, so the leaf note and the scan sentence cannot drift.

**4.** The pre-check moved to the in-order scan, after the source probe, and reads off the resolved `scanPlugins` — so an in-order `source=` plugin now counts as the scope it is, which the old condition missed too. The off-order call gets back the refusal it had earned (`conflicts_only= has no meaning on an out-of-load-order file`), pinned by a test.

**One I only half-folded, for a decision.** "Restore `where=` and `references=` to the in-order remedy list": I did not make them satisfy the check, because they do not. `RecordsTools.cs:947` refuses `where=`/`references=` unless combined with `types=`, `plugins=` or a `formids=` set, and the engine's own body-bearing test is `hasType || hasPlugins || hasFormidSet` — so a caller told to "add `where=`" would have walked straight into a second refusal. They are named in the sentence with what is actually true of them: *"where= and references= read bodies too, but each takes one of those as its own bound"*. If the intent was that a `where=` scan should count as body-bearing on its own, that is a change to both guards and I would rather file it than smuggle it in here.

**5.** `carrierBound` is required; the two probes and one test that relied on the default now state `"limit="` themselves.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2531,15 +2567,18 @@ internal RecordSummary ResolveSummaryOn(CrossQueryOutcome q, FormKey fk)
...
         if (memo.TryGetValue(fk, out var hit)) return hit;
         ResolvedRef result;
         var w = view.ResolveWinner(fk);
         if (w is null)
             result = EngineImplicit.TryDescribe(fk, out var eiType, out var eiEditorId)
                 ? new ResolvedRef(fk.ToString(), Resolved: true, Type: eiType, EditorId: eiEditorId, Winner: "<engine>")   // engine-implicit: hardcoded, real, defined by no plugin
-                : new ResolvedRef(fk.ToString(), Resolved: false);   // valid FormKey, but no active plugin defines it (a dangling target)
+                // Valid FormKey, no active plugin defines it. The reason is the three-cause sentence every other
+                // lane states, so the identity form's row says WHICH cause instead of a bare "not present".
+                : new ResolvedRef(fk.ToString(), Resolved: false, Error: UnresolvedFormId(view, fk, absenceMemo));
```
Avick3110: `ResolveRefOne` is shared by two callers, and only one of them reads `Error`. `AnnotateLinks` (line 2359) calls it as `ResolveRefOne(view, session, fk, memo)` — no `absenceMemo` — and neither render of an annotation ever shows the string: `Wire.LinkText` branches on `Resolved`/`Winner`, and `JsonWire`'s `link` object writes `resolved` + identity only. So on the `project.resolve_names=true` path the sentence is computed and thrown away.

That matters because computing it is not free. When the target's plugin is absent from the order, `UnresolvedFormId` falls through to `view.AbsenceClause(defining, out cause)` → `ExplainPluginAbsence`, which does `Mo2LoadOrder.ReadComposition(profileDir)` (profile parse off disk) and, on the ticked/on-disk arms, `Mo2LoadOrder.LocatePlugin` (a sweep of the mods/overwrite/data folders). Without the memo that runs once per **distinct** dangling FormKey, not once per plugin.

Concrete: `housecarl_records formids=[…200 records…] project={form:'everything', resolve_names:true}` over a plugin that declares a master which is not installed — the exact case `check_errors`' MISSING MASTER category exists for. Every distinct dangling link re-parses the MO2 profile and re-sweeps the mod folders, for a string no renderer prints. Either skip the `Error` when the caller is the annotate lane, or thread the same `absenceMemo` through `AnnotateLinks` the way `ResolveRefs` does.

src/housecarl-core/ReadEngine.cs
```diff
@@ -339,12 +341,71 @@ static string NoFieldNote(object owner, string segName, string? precedingField)
...
         }
         return $"(no field {segName})";
     }
 
+    /// <summary>What to actually DO about a path that dotted THROUGH a list/dict — checked against the element
+    /// type, never asserted. A missing bracket and a wrong leaf name look identical at the dead-end, and they need
+    /// opposite next moves, so the segment is resolved against the collection's element type first: it exists, and
+    /// the remedy is the exact bracketed spelling; it does not, and the remedy says so and offers the nearest real
+    /// field. Only where the element type cannot be determined does this fall back to the shape advice alone.</summary>
+    internal static string ListHopRemedy(object owner, string segName, string pf)
```
Avick3110: `NoFieldNote` is composed eagerly on every leaf miss — `ReadLeaf` builds it at lines 239 and 248 before returning, and the predicate scan calls `ReadLeaf` once per scanned record. So this remedy is now recomputed for every record in the scan, while `FieldPredicateSet.Matches` keeps only the first one (`_listHopRemedy[k] ??= _lastListHopRemedy`).

What it costs per record: `ElementType(owner)` (two `GetInterfaces()` passes, plus enumerating the collection's first element on the untyped fallback), then `WriteEngine.ResolveProperty(et, segName)` — which is uncached and allocates a `List`, a `HashSet`, a `Queue`, calls `GetInterfaces()` again, does a `GetProperty` per interface and finishes with a LINQ `OrderByDescending`. On the miss branch it also runs `ElementFieldNames(et)` (`GetProperties` across every interface) and a Levenshtein sweep in `PluginNameSuggest.Nearest`.

The miss branch is the common one during a scan: a wrong list-hop path is wrong for *every* record, so `where=["Effects.NotOnAnEffect > 0"]` with `types=["SPEL","NPC_","WEAP"]` over a large order pays the full reflection + suggester sweep tens of thousands of times to produce one sentence. Before this change the same path was a string interpolation.

Cheapest fix that keeps the behaviour: memoise on `(elementType, segName)` — the answer only depends on those two, and a scan asks the same question every time.

src/housecarl-core/ReadEngine.cs
```diff
@@ -339,12 +341,71 @@ static string NoFieldNote(object owner, string segName, string? precedingField)
...
+        // check applies.
+        if (segName.Length > 0 && segName.All(char.IsDigit))
+            return $"index an element with brackets, e.g. '{pf}[{segName}]', not '{pf}.{segName}'";
+
+        var et = ElementType(owner);
+        if (et is null)
+            return $"index an element with brackets, e.g. '{pf}[0].{segName}', not '{pf}.{segName}'";
+
+        var etName = RecordNaming.StripOverlay(et.Name);
+        if (WriteEngine.ResolveProperty(et, segName) is not null)
```
Avick3110: The bracketed path this prints stops at the dead-end segment, so it is a prefix of the caller's path, not the fixed path — and following it literally lands them in a different refusal.

`where=["Effects.Data.Magnitude > 0"]` dead-ends at `Data` (the walk is at `path[1]`, line 239), so `segName="Data"`, `pf="Effects"`, and the remedy reads `index an element with brackets: 'Effects[0].Data', not 'Effects.Data'`. A caller who copies that gets `predicate field 'Effects[0].Data' … it resolves to a container/list here, not a scalar leaf` from the same accounting rollup, and has to work out for themselves that `.Magnitude` needed re-appending. The working path is `Effects[0].Data.Magnitude`.

The changelog states this as "the exact bracketed path is printed (`Effects[0].Data`)", and `APredicateHoppingThroughAListIsToldTheExactBracketedSpelling` asserts the prefix — so this may be deliberate, but the sentence is a remedy the caller is meant to paste, and as written it is not one. `ReadLeaf` has the full `path` array in hand at both call sites; passing the trailing segments in would let this print `Effects[0].Data.Magnitude`.

Same for the did-you-mean arm below: `Effects.Dat.Magnitude` would be answered `did you mean 'Effects[0].Data'?` rather than `'Effects[0].Data.Magnitude'`.

Avick3110:
Folded all three cost/remedy findings, each with a test that fails without its fix (verified by reverting the fix and re-running).

- `01ce500` — **resolve_names explained an absent master once per dangling FormKey.** `AnnotateLinks` called `ResolveRefOne` without an absence memo, so `UnresolvedFormId` re-ran the explainer (MO2 profile parse + mod-folder sweep) for every link into the same missing plugin. The link cache is now one `LoadOrderService.LinkMemo` carrying both the per-FormKey identities and the per-plugin absence tails, so every lane that already shared a link memo shares the absence cache too. Test: `AbsentMasterLinkTests.ResolveNamesExplainsAnAbsentMasterOncePerPluginNotOncePerDanglingLink` reads six records whose links each dangle into one uninstalled master and asserts the explainer ran once (counter on the service). Before the fix: six.
- `89599bd` — **the bracketed remedy printed a prefix, not the fixed path.** The read walk now hands the note the trailing segments, so `Effects.Data.Magnitude` is answered with `'Effects[0].Data.Magnitude'`, and the did-you-mean arm carries them as well (`Effects.Dat.Magnitude` → `did you mean 'Effects[0].Data.Magnitude'?`). `APredicateHoppingThroughAListIsToldTheExactBracketedSpelling` and its two siblings now assert the whole path, `ANearMissMidPathNamesTheWholeCorrectedSpelling` covers the did-you-mean arm, and the changelog entry says the printed spelling is one a caller can paste back.
- `359f84b` — **the list-hop remedy was composed on every leaf miss.** The element-type verdict (property resolve, element field names, nearest-name sweep) is now memoised on (element type, segment); only the cheap formatting of the caller's own path runs per miss. Test: `AScanComputesOneListHopRemedyForTheWholeScan` scans three SPEL records with a wrong list-hop path and asserts one verdict computation. Before the fix: three.

Build, `dotnet test --filter "tier!=bridge"` (1160 passed) and `ci-all` (128/128) all green, rebased onto current `main`.

Issue #489, opened 2026-09-03
[GAP] Reverse-reference index: unbounded and transitive "who references X" — built lazily, kept current per plugin (E1's engine half)


Avick3110:
**Direction ruled by Aaron, 2026-09-03**, on the advisor's recommendation after the E1 lookup/traversal plan: *"lazily and yes it needs to stay up to date or it is useless, memory is trivial so I agree on all."* Filed by the advisor lane; routed row 8 (W5) as E1's engine half; milestone `houseCARL 2.0`.

## The gap

"Who references X" is on the 2.0 surface today as `housecarl_records references=[…]` (reverse, one step) and `walk.direction='reverse'` (depth 1). Both require a bounding `types=`/`plugins=` scope, because without an index a reverse lookup is a typed-winner scan: **0.4–5.2 s per large plugin, ≈25 s order-wide** (reverse-ref precheck, 2026-06-26, on the live 3,520-plugin order). Transitive reverse (`depth > 1`) is refused outright, naming the reverse-reference index as the future capability (SPEC §3.2, "a declared cost-refusal"). The precheck's own decision was "the index is a feature, not a prerequisite — sequenced on its own value" — and it was never given an issue or a row, so it has been the unscheduled piece behind every "xEdit's referenced-by, but for agents" conversation since.

For an agent this is the difference between an audit that costs one call and one that costs a scope guess per plugin. The result shape already exists — identity rows (FormID, EditorID, type, which target was hit) the agent drills into with a normal read — so the missing half is the engine's, not the door's.

## What ships

A reverse-reference index on the resolver: `FormKey → set of referencing (record, plugin)` over every non-null FormLink in the order, built by the same whole-order link-walk `check`'s dangling sweep already runs. Measured at the precheck: 3,288,460 records · 27,044,696 links · 629,069 distinct targets · 1,513,154 (target, plugin) pairs · **~149 MB held**, ~25 s to build once. Derived data, no file handles at rest.

Three design points, each ruled:

1. **Built lazily** — on the first reverse call that needs it (an unbounded `references=`, or any `depth > 1` reverse), with the one-time cost declared in the response; never at startup.
2. **Stays current under churn** — keyed by plugin, so the epoch/mtime re-check invalidates and recomputes only the changed plugins' entries (0.4–5 s each), never the whole order. An index that goes stale silently is worse than none (Q3); an index that rebuilds wholesale on every MO2 touch is unusable on a 3,800-plugin order.
[...]

Issue #459, opened 2026-08-31
[GAP] no parent-of-this-record read for group-nested containment — DIAL->INFO, CELL->REFR, WRLD->CELL


Avick3110:
> *Filed from an external live-session gap report by **Jouten** (running 1.9.0), with their OK to be named; triaged, root-caused against source, and posted by AI. Machine paths, the reporter's in-progress mod details, and their project's internal voice-map specifics scrubbed.*

### Goal

A test round surfaced a silent dialogue line by its response text. The INFO was found by content (`cross_plugin_query where=["Responses[0].Text contains ..."]`); what was needed next was its **owning DIAL topic** — for the voice filename prefix and a pool audit.

### What fell short

DIAL -> INFO ownership is **group nesting, not a FormLink**, so `cross_plugin_query type=DialogTopic references=[<INFO formid>]` returns 0 matches, and the INFO record itself carries `Topic=(null link)`. There is no "parent of this record" read.

Tool calls tried:

    housecarl_cross_plugin_query type=DialogTopic references=["<INFO formid>"]   -> 0 matches
    housecarl_read_record <the INFO>                                             -> no parent field populated

### Workaround used

Guessed the topic family by EditorID substring (`cross_plugin_query type=DialogTopic editorid_contains=... fields=["Responses"] depth=2`) and eyeballed which topic's `Responses` list contained the INFO. Worked, but only because the naming was guessable.

### Frequency

Recurs whenever a line is found by content or FormID first — test-round triage, voice audits. The voice-filename rule needs the owning topic every time.

**Seen again 2026-08-26:** needed the owning DIALs of a hello-pool's INFOs to set `DoAllBeforeRepeating` on the topics; resolved via the project's own voice-map JSON (which happens to record topic per line), not via houseCARL.

### Ideal shape (reporter's)

`housecarl_read_record` on an INFO reports its owning DIAL (and quest), or `references=` treats group ownership as a link for DIAL/INFO.

- date: 2026-08-26  |  houseCARL version: 1.9.0

---

### Triage / root cause (verified against source)

Valid, reproducible, non-duplicate, and **structural**. Real on the same 1.9.0 the reporter runs.

[...]

PR #544, opened 2026-09-04
Add a probe measuring child-to-parent capture cost

Avick3110:
A manual generator harness, `parent-in-hand`, that answers two questions about child records on a real load order.

**1. Is the parent in hand during the flat `EnumerateMajorRecords` walk the index build runs today?** For one child kind, yes: a `NavigationMesh` names its cell at `Data.Parent`, whose `CellNavmeshParent` arm carries the link. For everything else, no. The check descends nested structs and polymorphic arms, samples every distinct child type under each child-bearing property (20 bodies per pair, or all if fewer), and runs on live bodies while the overlay is still open. The parent types it walks and the rows it reports come from one source — `WriteEngine.ChildBearingProperties` over `WriteSurfaceGuardProbe.ConcreteRecordTypes()` — so a Mutagen bump that adds a child-bearing type cannot leave a row unmeasured. Measured on the ARR 2.0 order:

| Property | Child type | Names its parent | Via |
|---|---|---|---|
| `Cell.Landscape` | `Landscape` | NO | — |
| `Cell.NavigationMeshes` | `NavigationMesh` | YES | `Data[NavigationMeshData].Parent[CellNavmeshParent].Parent` |
| `Cell.Persistent` | `PlacedHazard` | NO | — |
| `Cell.Persistent` | `PlacedNpc` | NO | — |
| `Cell.Persistent` | `PlacedObject` | NO | — |
| `Cell.Persistent` | `PlacedTrap` | NO | — |
| `Cell.Temporary` | `PlacedHazard` | NO | — |
| `Cell.Temporary` | `PlacedNpc` | NO | — |
| `Cell.Temporary` | `PlacedObject` | NO | — |
| `Cell.Temporary` | `PlacedTrap` | NO | — |
| `DialogTopic.Responses` | `DialogResponses` | NO | — |
| `Worldspace.SubCells` | `Cell` | NO | — |
| `Worldspace.TopCell` | `Cell` | NO | — |

**2. What does capturing the parent cost?** Two timed passes only: A, today's flat walk, and E, `EnumerateMajorRecordContexts`, which carries the parent, needs no link cache, and builds the packed ulong-to-ulong child-to-parent map. They run three times each, alternating (A E A E A E), with everything released between runs, after a discarded warm-up; the median of each is reported. Nothing but the map build runs inside the stopwatch — the by-type report and the re-parent count are derived afterwards from staged data.

On the ARR 2.0 order (3,801 plugins, 3,691,385 records) the flat walk takes 7.98 s and the context walk 8.41 s — 0.43 s more, 1.05x — for 2,711,713 child-to-parent edges. The two walks are close enough that the machine's own run-to-run spread is a few tenths of the gap: a second run of the same build gave 8.75 s / 9.12 s, 0.37 s more, 1.04x. Stored packed the map is 77.3 MB (~30 B/entry). Across the order 2,363 distinct children end up under a different parent than the first plugin put them under, so a single whole-order map is not well defined and the map has to be per plugin.

The harness needs a live MO2 instance (or `--plugin`), so it is dispatched on its own and is not wired into `ci-all`. Not user-visible, so no changelog line.

Refs #459 #489

### DIFF
--- a/src/housecarl-generator/ParentInHandProbe.cs
+++ b/src/housecarl-generator/ParentInHandProbe.cs
@@ -28,29 +28,34 @@ static class ParentInHandProbe
         if (paths is null) return 1;
         Console.WriteLine($"parent-in-hand: {label} — {paths.Count} plugin(s)\n");
 
+        // A discarded warm-up sweep runs first. Without it the first timed pass alone pays the JIT of both
+        // enumerators and the cold file cache for every plugin, which is bigger than the difference being measured.
+        Sweep(paths, dataDir, new WarmupPass(), "warm-up");
+        Console.WriteLine("warm-up sweep done and discarded; every timing below is warm.\n");
+
         // ---- pass A: the flat walk BuildIndex runs today ------------------------------------------------------
         var flat = new FlatPass();
-        var passA = Sweep(paths, dataDir, flat);
+        var passA = Sweep(paths, dataDir, flat, "A");
 
         Console.WriteLine($"A. flat pass (today's BuildIndex walk)");
         Console.WriteLine($"   records          : {flat.Records:N0}");
         Console.WriteLine($"   wall             : {passA.Wall.TotalSeconds:N2} s");
         Console.WriteLine($"   retained after   : {passA.RetainedMb:N1} MB   (FormKey buffer only, as today)");
-        Console.WriteLine($"   plugins skipped  : {passA.Skipped}\n");
+        Console.WriteLine($"   plugins excluded : {passA.Skipped}\n");
 
         // ---- pass B: flat + containment, fused into the SAME open ---------------------------------------------
         var both = new FusedPass();
-        var passB = Sweep(paths, dataDir, both);
+        var passB = Sweep(paths, dataDir, both, "B");
 
         Console.WriteLine($"B. fused pass (flat walk + containment walk, one open per plugin)");
         Console.WriteLine($"   records          : {both.Records:N0}");
         Console.WriteLine($"   parents visited  : {both.Parents:N0}");
         Console.WriteLine($"   child->parent    : {both.Map.Count:N0} distinct children");
         Console.WriteLine($"   wall             : {passB.Wall.TotalSeconds:N2} s");
-        Console.WriteLine($"   retained after   : {passB.RetainedMb:N1} MB   (includes the child->parent map)");
-        Console.WriteLine($"   plugins skipped  : {passB.Skipped}");
-        Console.WriteLine($"   MARGINAL time    : {(passB.Wall - passA.Wall).TotalSeconds:N2} s " +
-                          $"({(passA.Wall.TotalSeconds > 0 ? (passB.Wall.TotalSeconds / passA.Wall.TotalSeconds) : 0):N2}x the flat pass)");
+        Console.WriteLine($"   retained after   : {passB.RetainedMb:N1} MB   (the child->parent map)");
+        Console.WriteLine($"   plugins excluded : {passB.Skipped}");
+        Console.WriteLine($"   MARGINAL time    : {(passB.Wall - passA.Wall).TotalSeconds:N2} s more than the flat pass " +
+                          $"(B in total is {(passA.Wall.TotalSeconds > 0 ? (passB.Wall.TotalSeconds / passA.Wall.TotalSeconds) : 0):N2}x A)");
         Console.WriteLine($"   MARGINAL memory  : {(passB.RetainedMb - passA.RetainedMb):N1} MB\n");
 
         // ---- pass C: the same map, packed --------------------------------------------------------------------
@@ -58,16 +63,29 @@ static class ParentInHandProbe
         // pays for two of those per entry. The index already has a dense plugin index, so pack (modIdx, id) into
         // one ulong and measure what the SAME map costs stored that way.
         var packed = new PackedPass();
-        var passC = Sweep(paths, dataDir, packed);
-        Console.WriteLine($"C. packed map only (ulong -> ulong, (modIdx<<32|id))");
+        var passC = Sweep(paths, dataDir, packed, "C");
+        Console.WriteLine($"C. packed map (ulong -> ulong, (modIdx<<32|id)); same work as B, map stored packed");
         Console.WriteLine($"   entries          : {packed.Map.Count:N0}");
-        Console.WriteLine($"   wall             : {passC.Wall.TotalSeconds:N2} s");
-        Console.WriteLine($"   retained after   : {passC.RetainedMb:N1} MB   ({passC.RetainedMb * 1024 * 1024 / Math.Max(1, packed.Map.Count):N0} B/entry)");
-        Console.WriteLine($"   without Cell.Temporary : {both.Map.Count - both.PerProperty.GetValueOrDefault("Cell.Temporary"):N0} entries\n");
+        Console.WriteLine($"   wall             : {passC.Wall.TotalSeconds:N2} s   (flat walk + containment walk, as B)");
+        Console.WriteLine($"   retained after   : {passC.RetainedMb:N1} MB   (map plus the small ModKey index; " +
+                          $"~{passC.RetainedMb * 1024 * 1024 / Math.Max(1, packed.Map.Count):N0} B/entry)\n");
 
-        Console.WriteLine("   per child-bearing property (children reached by the containment walk):");
-        foreach (var kv in both.PerProperty.OrderByDescending(k => k.Value))
-            Console.WriteLine($"     {kv.Key,-34} {kv.Value,10:N0}");
+        // ---- the descriptive breakdown, deliberately OUTSIDE the timed passes --------------------------------
+        // Per-property distinct counts and one live sample body per property. Both would distort the figures
+        // above — the sample pins a whole overlay graph, the counts need a second set per property — so they are
+        // collected in their own sweep whose time and memory are not reported.
+        var describe = new DescribePass();
+        Sweep(paths, dataDir, describe, "describe");
+
+        var temporary = describe.PerProperty.GetValueOrDefault("Cell.Temporary")?.Count ?? 0;
+        var nonTemporary = describe.PerProperty.Where(k => k.Key != "Cell.Temporary")
+                                               .SelectMany(k => k.Value).Distinct().Count();
+        Console.WriteLine($"   Cell.Temporary children : {temporary:N0} distinct");
+        Console.WriteLine($"   children reached by any OTHER property : {nonTemporary:N0} distinct\n");
+
+        Console.WriteLine("   per child-bearing property (distinct children reached by the containment walk):");
+        foreach (var kv in describe.PerProperty.OrderByDescending(k => k.Value.Count))
+            Console.WriteLine($"     {kv.Key,-34} {kv.Value.Count,10:N0}");
         Console.WriteLine();
 
         // ---- pass E: Mutagen's own context enumeration, which carries a Parent ---------------------------------
@@ -75,7 +93,7 @@ static class ParentInHandProbe
         // IModContext<IMajorRecordGetter>, whose IModContext.Parent is the containing record's context. If that is
         // populated, the containment map needs no custom walk at all — one enumeration replaces both passes.
         var ctx = new ContextPass();
-        var passE = Sweep(paths, dataDir, ctx);
+        var passE = Sweep(paths, dataDir, ctx, "E");
         Console.WriteLine($"E. Mutagen context enumeration (EnumerateMajorRecordContexts, no link cache)");
         Console.WriteLine($"   contexts         : {ctx.Contexts:N0}");
         Console.WriteLine($"   with a Parent    : {ctx.WithParent:N0}");
@@ -84,7 +102,7 @@ static class ParentInHandProbe
         Console.WriteLine($"   re-parented      : {ctx.Reparented:N0} (a later plugin put the same child under a DIFFERENT parent)");
         Console.WriteLine($"   wall             : {passE.Wall.TotalSeconds:N2} s");
         Console.WriteLine($"   retained after   : {passE.RetainedMb:N1} MB");
-        Console.WriteLine($"   plugins skipped  : {passE.Skipped}");
+        Console.WriteLine($"   plugins excluded : {passE.Skipped}");
         Console.WriteLine($"   vs the flat pass : {(passE.Wall - passA.Wall).TotalSeconds:N2} s more");
         Console.WriteLine("   edges the context walk hands over (child -> parent, by type):");
         foreach (var kv in ctx.Edges.Where(k => !k.Key.EndsWith("(no parent)", StringComparison.Ordinal))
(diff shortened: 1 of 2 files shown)


Avick3110:
All twelve review findings folded in `ab4690a`, and the numbers below are a fresh run of the corrected harness against the live ARR 2.0 order (profile Custom 1).

1. Parent-naming check missed nested members — the walk now descends nested structs to depth 5 and reads each value's runtime type, so `NavigationMesh.Data.Parent` is found.
2. Polymorphic arms invisible — the runtime type is what is reflected over, and the arm is spelled in the reported path (`Data[NavigationMeshData].Parent[CellNavmeshParent].Parent`).
3. Unfair B vs A comparison — pass B deleted.
4. Unfair C vs A comparison — pass C deleted; the packed-map memory figure comes from E, whose re-parent bookkeeping is dropped before memory is read.
5. Single-run timings with an ordering advantage — A and E now run three times each, alternating A E A E A E, everything released between runs, median reported (the three runs are printed too).
6. `FlatPass`'s retained-memory line measured nothing — deleted.
7. Only two timed passes remain, A and E, as asked.
8. `DescribePass` half-counted a plugin that threw mid-walk — it now stages per plugin and merges only on completion, like the other passes.
9. "Re-parented" counted overwrite events — it now counts distinct children whose final parent differs from the first one seen, and the line says so. On this order that is 2,363, not the 2,832 previously reported.
10. The parent check ran on retained bodies after dispose — it now runs inside the sweep while the overlay is open.
11. A run where every plugin failed to open exited 0 — it now exits 1 with one sentence.
12. A negative difference printed "-0.03 s more" — it prints the absolute value with "less" or "more".

Corrected numbers: flat walk 7.79 s, context walk 8.76 s (0.97 s more, 1.13x), 3,691,385 records over 3,801 plugins, 2,711,713 child-to-parent edges, 77.3 MB packed. Question 1 now answers YES for `NavigationMesh` and NO for the other twelve (property, child type) pairs; the full table is in the PR body.

Build, tests (`tier!=bridge`, 1058 passed) and `ci-all` are green.

src/housecarl-generator/ParentInHandProbe.cs
```diff
@@ -0,0 +1,580 @@
...
+        {
+            long contexts = 0, withParent = 0, parentIsRecord = 0;
+            var edges = new Dictionary<string, long>(StringComparer.Ordinal);
+            var staged = new List<(FormKey Child, FormKey Parent)>();
+            void Bump(string k) => edges[k] = edges.GetValueOrDefault(k) + 1;
+
+            foreach (var c in ov.EnumerateMajorRecordContexts())
+            {
+                contexts++;
+                var child = Short(c.Record.GetType());
```
Avick3110: **The E pass's timed region builds the descriptive edge histogram, so the headline delta is inflated.**

For every one of the ~3.7M contexts this loop calls `Short(c.Record.GetType())` — which allocates a fresh `string[2]` per call (line 517) plus a `Substring` — interpolates a `"child -> parent"` key, and hashes that key twice (`GetValueOrDefault` then the indexer set). Lines 193, 199 and 211 do the same. The A pass does none of it.

Measured standalone on this machine, that bookkeeping alone costs **~0.45 s for 3,691,385 iterations** in a best case (5 distinct keys, tiny dictionary, no reflection). The reported delta is 0.97 s. So roughly half of "0.97 s more, 1.13x" is the by-type breakdown, not the cost of carrying the parent.

This is the same argument the file already makes for question 1 (lines 107–109: "Both would distort the figures above, so they get their own untimed sweep") — the histogram was left inside. Either stage the `(childType, parentType)` pair as two `Type` references and build the string keys after `sw.Stop()`, or move the histogram into the untimed `describe` sweep.

src/housecarl-generator/ParentInHandProbe.cs
```diff
@@ -0,0 +1,580 @@
...
+        return order.OrderedPaths;
+    }
+
+    /// <summary>The instance housecarl_set_mo2_instance saved, if any — the same file the server reads.</summary>
+    static string? UserConfiguredInstance()
+    {
+        foreach (var dir in new[]
+        {
+            Environment.GetEnvironmentVariable("HOUSECARL_DATA_DIR"),
+            Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "housecarl", "server"),
```
Avick3110, line 638: **This is the Codex install's path, not the one the server reads.**

The doc comment says "the same file the server reads", but `src/housecarl-mcp/Program.cs:77-78` resolves the user config to `HOUSECARL_DATA_DIR` when set, **else `AppContext.BaseDirectory` — beside the exe**. `housecarl-setup` puts the Claude server at `%USERPROFILE%\.claude\skills\housecarl\server\` (`ClaudeDestExe`); `%LOCALAPPDATA%\houseCARL\server` is only `CodexServerDir`.

`HOUSECARL_DATA_DIR` is unset in a plain shell, so a no-argument `parent-in-hand` never sees the Claude install's config. On a Claude-only machine it prints "no MO2 instance given and none saved" although `housecarl_set_mo2_instance` did save one. Where both installs exist it silently measures whatever the *stale* Codex copy names — on this machine that file is three months older than the live one and carries a different set of keys.

Two smaller points on the same block: hand-parsing with `JsonDocument` bypasses `UserConfigStore`, which `UserConfig.cs:43` declares "the single OWNER of houseCARL.user.json — every read and write of that file goes through here"; and the bare `catch { }` folds a corrupt file into the same "none saved" message instead of saying the file was unreadable.

src/housecarl-generator/ParentInHandProbe.cs
```diff
@@ -0,0 +1,580 @@
...
+                          "(distinct children, not overwrite events)");
+        Console.WriteLine($"   wall             : {medE.TotalSeconds:N2} s   (runs: {string.Join(", ", eWall.Select(w => $"{w.TotalSeconds:N2}"))})");
+        Console.WriteLine($"   retained after   : {Median(eMemory):N1} MB   (the packed map alone; " +
+                          $"~{Median(eMemory) * 1024 * 1024 / Math.Max(1, ctxFinal.Map.Count):N0} B/entry)");
+        Console.WriteLine($"   plugins excluded : {eSkipped}");
+        var delta = (medE - medA).TotalSeconds;
+        Console.WriteLine($"   vs the flat pass : {Math.Abs(delta):N2} s {(delta < 0 ? "less" : "more")} " +
+                          $"({(medA.TotalSeconds > 0 ? medE.TotalSeconds / medA.TotalSeconds : 0):N2}x A)");
+        Console.WriteLine("   edges the context walk hands over (child -> parent, by type):");
+        foreach (var kv in ctxFinal.Edges.Where(k => !k.Key.EndsWith("(no parent)", StringComparison.Ordinal))
```
Avick3110: **`(parent chain holds no record)` rows are printed as if they were edges.**

`Bump` writes three key shapes, not two: `-> (no parent)` (line 193), a real parent type, and `-> (parent chain holds no record)` (line 213) for a context whose parent is a group and whose chain either runs out or exceeds the 6-hop cap. This filter only excludes the first, so the third is printed under the heading "edges the context walk hands over (child -> parent, by type)" even though nothing was staged into `Map` for it — and line 102's "(no parent, N top-level records)" tally skips it too.

A reader summing the printed table gets a number larger than `Map.Count` with no way to tell which rows are not edges. Filter on a marker that covers both non-edge shapes, or print them under their own heading.

src/housecarl-generator/ParentInHandProbe.cs
```diff
@@ -0,0 +1,580 @@
...
+                tally.Unreadable += kv.Value.Unreadable;
+                tally.Paths.UnionWith(kv.Value.Paths);
+            }
+            foreach (var kv in sample) Sample.TryAdd(kv.Key, kv.Value);
+        }
+    }
+
+    /// <summary>Every record in the plugin that BEARS children — the three parent types, asked of the corpus rather
+    /// than named here: a type is a parent iff OwnedChildContent.Fields answers non-empty for it.</summary>
+    static IEnumerable<IMajorRecordGetter> ParentBodies(ISkyrimModGetter ov)
```
Avick3110, line 362: **The comment says the parent set is asked of the corpus; the code names it.**

`ChildBearingSurface` (line 504) derives the report's row set from `WriteEngine.ChildBearingProperties` over every concrete `IMajorRecord` type — the set `OwnedChildContent`'s own doc says "a Mutagen bump that grows one is picked up without an edit here". `ParentBodies` enumerates a hand-written three.

After such a bump the new type's properties still appear in the row set, reach `ReportParentInHand` with zero sampled rows, and print `no children in this order (0 seen) — not measured` (line 412). That asserts the order contains none, when the probe never enumerated a single body of that type. The two halves have to come from the same place, or the report has to distinguish "nothing found" from "never looked".

Driving the loop off `WriteSurfaceGuardProbe.ConcreteRecordTypes().Where(t => WriteEngine.ChildBearingProperties(t).Count > 0)` and calling `ov.EnumerateMajorRecords(getterType)` per type would make one source answer both.

src/housecarl-generator/ParentInHandProbe.cs
```diff
@@ -0,0 +1,580 @@
...
+        Console.WriteLine($"     (no parent, {ctxFinal.Edges.Where(k => k.Key.EndsWith("(no parent)", StringComparison.Ordinal)).Sum(k => k.Value):N0} top-level records across {ctxFinal.Edges.Count(k => k.Key.EndsWith("(no parent)", StringComparison.Ordinal))} types)\n");
+
+        lastCtx = null;
+        GC.Collect(); GC.WaitForPendingFinalizers(); GC.Collect();
+
+        // ---- the descriptive breakdown and question 1, deliberately OUTSIDE the timed passes --------------------
+        // Per-property distinct counts, and the parent-naming check run on live bodies while the overlay is still
+        // open. Both would distort the figures above, so they get their own untimed sweep.
+        var describe = new DescribePass();
+        Sweep(paths, dataDir, describe, "describe");
```
Avick3110: **The describe sweep's `PassResult` is dropped, so this section has neither the zero guard nor the excluded count the timed passes get.**

`Sweep` returns `Skipped`/`Opened`; lines 39–43, 60 and 67 use them, this call discards them. If plugins throw partway through `ParentBodies`/`DirectChildren` — a parse fault in a cell group is exactly the class `BuildIndex` already tolerates — those plugins are excluded and the surviving counts are still printed as `({count:N0} distinct children in the order)` (line 415) and "per child-bearing property (distinct children reached by the containment walk)" (line 119), with no "plugins excluded : N" line to qualify them. If every plugin failed to open, the whole section prints zeros and "no children in this order (0 seen)" for every row instead of the refusal A and E give.

src/housecarl-generator/ParentInHandProbe.cs
```diff
@@ -0,0 +1,580 @@
...
+        }
+
+        // Does a LATER plugin ever leave the same child under a DIFFERENT parent? That is what decides whether one
+        // whole-order map is well-defined or the map has to be per-plugin. Counted as DISTINCT children whose final
+        // parent differs from the first one seen — not as overwrite events, which double-count a child moved twice
+        // and count a child moved away and back as two.
+        void Put(FormKey child, FormKey parent)
+        {
+            var ck = Pack(child); var pk = Pack(parent);
+            _first!.TryAdd(ck, pk);
```
Avick3110: **`_first` is excluded from the memory figure but not from the wall figure.**

The comment at lines 177–178 and `Finish()` go out of their way to drop `_first` before memory is read, so "retained after" is "the packed map alone". The same argument applies to the timing and was not made there: this `TryAdd` runs for all 2,711,713 staged edges inside `pass.Plugin`, which is inside `Sweep`'s stopwatch, so E's wall time includes a second full dictionary the map build itself would never do. Measured standalone that is ~0.03 s per pass on 2.7M entries — small beside the histogram above, but charged to the context walk all the same.

Staging the re-parent check outside the timed region (or deriving it from `Edges`-style staging in `Finish`) would keep the two figures measuring the same thing.

src/housecarl-generator/ParentInHandProbe.cs
```diff
@@ -0,0 +1,580 @@
...
+        var paths = ResolvePaths(args, out var label, out var dataDir);
+        if (paths is null) return 1;
+        Console.WriteLine($"parent-in-hand: {label} — {paths.Count} plugin(s)\n");
+
+        // A discarded warm-up sweep runs first. Without it the first timed pass alone pays the JIT of both
+        // enumerators and the cold file cache for every plugin, which is bigger than the difference being measured.
+        var warm = Sweep(paths, dataDir, new WarmupPass(), "warm-up");
+        if (warm.Opened == 0)
+        {
+            Console.WriteLine("every plugin failed to open, so nothing was measured — check the instance path and the profile's load order.");
```
Avick3110, line 41: **A `--plugin` run is sent to an MO2 profile it never used.**

`ResolvePaths` returns the `--plugin` path at line 532 without checking the file exists, so `parent-in-hand --plugin C:\missing.esp` throws in `OpenOverlay`, leaves `Opened` at 0, and lands here — printing "check the instance path and the profile's load order" for a run that consulted neither. The `--plugin` arm needs its own sentence naming the path that would not open.

Avick3110:
All seven folded. Re-run against the live ARR 2.0 instance (profile Custom 1); build, tests (tier!=bridge, 1058 passed) and ci-all (127/127) green.

1. **Histogram inside E's timed region** — `12608c4`. The by-type counter is now keyed on `(Type child, Type? parent, int hops)` while the clock runs, and the string keys are spelled in `Finish`, after the stopwatch stops. This was the bulk of the reported delta.
2. **Instance lookup hand-parsing the wrong file** — `91ba010`. Goes through `UserConfigStore` now, so `houseCARL.user.json` keeps its single owner and a corrupt file reports "unreadable" instead of "none saved". The dirs searched are the server's own: `HOUSECARL_DATA_DIR`, then beside the server exe — both install layouts, `%USERPROFILE%\.claude\skills\housecarl\server` first, the Codex `%LOCALAPPDATA%` one after. Doc comment rewritten.
3. **Non-edge rows under the edges heading** — `445c8b8`. Two tables now. The edge table carries only rows that staged an edge; the second heading carries the rows that staged nothing (top-level records collapsed to a tally, `(parent chain holds no record)` named individually, 9,459 `Cell` contexts). One note: the edge table sums to 3,051,457, the "parent is a record" count — those are edge *events*, and `Map.Count` is 2,711,713, one per *distinct* child, the difference being children an override re-parented or restated. I made the heading say both figures rather than force the table to sum to `Map.Count`, since a per-type table of distinct children would need a second de-duplicating pass over the whole order.
4. **`ParentBodies` hand-writing three types** — `4b438e3`. Driven off `WriteSurfaceGuardProbe.ConcreteRecordTypes().Where(t => WriteEngine.ChildBearingProperties(t).Count > 0)`, mapped to the getter interface via `WriteEngine.PrimaryGetter` and enumerated with `ov.EnumerateMajorRecords(getterType)`. `ChildBearingSurface` (the row set) now reads off the same helper, so one source really does answer both. A child-bearing type with no getter interface is printed as a gap rather than dropped.
5. **Describe sweep's `PassResult` dropped** — `573dd30`. It prints `plugins excluded : N` and refuses on zero opened, like A and E.
6. **Re-parent `TryAdd` inside E's timed region** — `12608c4`, same commit as 1 since it is the same loop. The staged edges are appended to a list in plugin order inside the timed region (a list append, not a second dictionary write), and `Finish` replays it to derive the first-seen parent and the re-parent count, then drops the list before memory is read.
7. **Missing `--plugin` path** — `a641074`. `ResolvePaths` checks existence and says `no plugin file at <path> — pass the path to an existing .esp/.esm/.esl.` instead of falling through to the instance-and-profile sentence.

**New numbers.** The delta was mostly the harness measuring itself. A 7.98 s, E 8.41 s — **0.43 s more, 1.05x**, down from 0.97 s / 1.13x. A second run gave 8.75 s / 9.12 s, 0.37 s / 1.04x, so the run-to-run spread on this machine is itself a good fraction of the gap; the honest reading is "within about half a second". Everything else is unchanged: 3,691,385 records, 3,060,916 contexts with a parent, 3,051,457 with a record parent, 2,711,713 distinct children, 2,363 re-parented, 77.3 MB packed. Question 1's answer is unchanged — `NavigationMesh` names its cell, nothing else names anything. PR body and the design note's engine paragraph updated.

PR #545, opened 2026-09-04
Fix the NIF surface's provider addressing, and add npc= and header-string writes

### LINKED ISSUE #340: [Bug] nif_inspect / nif_set: the provider spelling mod= prints is not the spelling it accepts

### LINKED ISSUE #388: [GAP] a mod's own root archives are reachable by name only when MO2 is not loading the mod

### LINKED ISSUE #412: [GAP] nif_inspect: no npc= input — resolving an NPC record to its facegen mesh is the caller's job

`nif_inspect`'s params are `mesh_paths[]` / `sections` / `mod` / `max_chars` (`NifTools.cs:61-82`; the batch conversion landed since the original filing, but no NPC input did). Resolving an NPC record → facegen `.nif` path is exactly the derivation the facegen-diagnostics flow does by hand every time; an `npc=` input would close the loop in one call.

**Routing:** row 7/W4 (S3/nif substrate, beside #340/#388). Milestoned.

### LINKED ISSUE #413: [GAP] nif_set: texture-slot mode only — material/xml ref path-swaps remain read-only

The op whitelist (`NifTools.cs:256`) is `rename_shape, rename_node, set_flags, set_scale, set_partition, set_alpha, set_path, set_shader_value`; `set_path` (`:226-229`) requires a numeric `texture_slot` — `BSShaderTextureSet` slots only. Material (`.bgsm`), `.tri`, and physics-xml refs are readable via `sections=strings` (`:68-75`) but not writable, though a material path swap is mechanically a header-string edit like a rename.

**Routing:** row 7/W4 (S3/nif substrate). Milestoned.

Avick3110:
Four issues on the NIF surface, in the order they build on each other.

**#340 — the printed provider name is the accepted one.** `nif_inspect` and `nif_set` listed providers as `ModA (loose)` and matched `mod=` on the bare `ModA`, so following the message verbatim always refused, and "strip the parenthetical" is not a rule a caller can safely infer when `SkyUI (SE)` is a real mod folder. Both sites now print through the same formatter the asset surface uses — the name inside double quotes, a character a Windows name cannot contain, with the kind outside them — and take `mod=` through the same source policy `place_asset`'s `source_provider=` rides, so the two cannot drift again. The pair was the same code twice and had already drifted once; it is now one helper.

**#388 — naming a mod reaches its own archives.** An archive now carries the MO2 layer its file lives in, and the named pole matches that as well as the archive's filename, so an enabled mod whose only copy is inside its own `.bsa` is reached by naming the mod rather than the archive. The folder-scan lane is no longer gated away from an enabled mod's name either, which is what closes part (iii): a mod's name reaches the same places ticked or unticked. The gate that remains is narrower and is the one that was actually load-bearing — a name that means a *layer* ("overwrite", "Data", an active archive's filename) still never falls through to a folder called that, so `mods\Data` cannot shadow the Data folder. On the NIF tools `mod=` is also answered before the ABSENT check, so part (ii)'s dead end — a donor outside the active set reported as "no active mod provides this mesh path", one step before reading absent as "the donor has no mesh" — is now a refusal that names the mod and says where houseCARL looked.

That reach comes with a disclosure, because bytes read out of a copy the game is not loading must never present as what the game shows: the NIF response says so and says which of the two reasons it is (the mod is unticked, or the archive is one no active plugin binds), and `nif_set`'s `in_place` lane refuses such a copy outright — its consent handshake is written about the winning file, and overwriting an original the caller reached by naming a mod is not that.

**#412 — `nif_inspect npc=`.** An NPC FormID (plugin-qualified or the runtime form) becomes the FaceGen head mesh path its FormKey defines — the folder is the plugin that *defines* the NPC, never the load-order winner — and joins the same batch as any `mesh_paths` in the call. A value on SELECT, not a mode: one of the two is required and both may appear together.

**#413 — `set_path` takes a header-string target.** Same op, one more addressing form, chosen by whether `texture_slot` was given: without it, `target=` is the header string to replace, so a material (`.bgsm`), a `.tri` or a physics-xml ref is swappable. The whitelist stays a whitelist — three refusals guard it: a shape's or node's name goes to `rename_shape` / `rename_node`, which carry the rename-onto-an-existing-name guard this form would otherwise walk around; an extra-data block's `Name` is the key the engine looks it up by, not a path; and a replacement already in the table is refused, because merging two entries renumbers every later index and the verification cannot tell that from a collateral write. Both existing gates apply, and read-back checks the reloaded mesh's own string table rather than the refs the write went through, so a partial swap cannot verify green.

**Tests.** `dotnet build` clean, 1056 tests pass, `ci-all` 128/128. A new `nif-source-lane-guard` probe drives both tools through the real tool layer over a synthetic MO2 instance with deliberately hostile mod names (`JK's Skyrim`, `Donor Mod (SE)`): every provider name is parsed back out of the rendered chain by its delimiter and fed straight into `mod=`, and each must select that provider — a substring check would not catch this bug, since `ModA (loose)` contains `ModA`. The same probe covers the enabled-mod-archive reach, the unticked-mod read and its provenance line, the `in_place` refusal, `*winner`, and `npc=`. `nif-set-guard` gains the header-string arms including the length-changing swap that a byte-position diff would false-abort on. Two probes that pinned the old spelling and the old reach are updated to the new contract, and the `facegen-diagnostics` and `npc-appearance-copy` skills lose the claims these fixes retire (material/`.tri`/xml refs are no longer NifSkope's, and a switched-off donor no longer needs switching on for the mesh read).

**One thing I could not do.** `place_asset` renders its own off-order provenance line in `PlaceAssetTools.cs`, which this lane was told not to touch. The sentence now takes a flag for the enabled-mod-unloaded-archive case and the NIF surface passes it; `place_asset`'s call site still omits it, so that case reads "NOT enabled in MO2" there. The fix is one argument at that call site — there is a `TODO(#388)` on the sentence naming it — and it belongs with the `place` rewrite that owns the file this wave. Related: `asset_status` renders the same annotated provider form and #340's kickoff asked for it to be aligned here, but that render lives in `AssetTools.cs`, also off-limits to this lane; it is unchanged.

Closes #340
Closes #388
Closes #412
Closes #413

### DIFF
--- a/src/housecarl-core/AssetResolver.cs
+++ b/src/housecarl-core/AssetResolver.cs
@@ -57,9 +57,12 @@ public sealed record AssetHit(string RelPath, bool Exists, AssetProvider? Winner
 /// <para><see cref="OffOrder"/> is true only for a source resolved by <see cref="OffOrderAssetSource"/> — a mod folder
 /// the ACTIVE profile does not include, reachable only because a caller named it. The response must state this: bytes
 /// read out of an unticked mod are correct bytes from a mod the game is not loading, and a caller cannot tell that
-/// from the provider name alone.</para></summary>
+/// from the provider name alone.</para>
+/// <para><see cref="OwningMod"/> is the MO2 layer a BSA source's archive file physically lives in — a mod folder
+/// name, "overwrite" or "Data" — so naming that mod addresses its own archives, not only its loose files (#388).
+/// Null for a loose source, whose <see cref="ProviderName"/> already IS that layer.</para></summary>
 public sealed record PlacementSource(string ProviderName, AssetKind Kind, string? LooseFilePath, string? ArchivePath,
-                                     string EntryPath, bool OffOrder = false);
+                                     string EntryPath, bool OffOrder = false, string? OwningMod = null);
 
 /// <summary>Concrete-source resolution of one asset path for PLACEMENT (place_asset's auto-resolve when no explicit
 /// source= is given): every provider with its on-disk descriptor, winner FIRST (the same precedence
@@ -68,9 +71,11 @@ public sealed record PlacementSource(string ProviderName, AssetKind Kind, string
 /// unscanned). <see cref="Sources"/> empty ⇒ nothing active provides this path.</summary>
 public sealed record PlacementResolution(string RelPath, IReadOnlyList<PlacementSource> Sources, bool Ambiguous, bool ReadIncomplete);
 
-/// <summary>An active BSA the resolver should consider: its full path, the plugin it loads with, and that plugin's
-/// load-order rank (higher = later in load order = wins among BSAs). The service derives these.</summary>
-public sealed record ActiveArchive(string Path, string OwningPlugin, int PluginRank);
+/// <summary>An active BSA the resolver should consider: its full path, the plugin it loads with, that plugin's
+/// load-order rank (higher = later in load order = wins among BSAs), and the MO2 layer the archive FILE lives in
+/// (a mod folder name, "overwrite" or "Data") so a caller can address it by naming that mod. The service derives
+/// these.</summary>
+public sealed record ActiveArchive(string Path, string OwningPlugin, int PluginRank, string? OwningMod = null);
 
 public sealed class AssetResolver : IDisposable
 {
@@ -350,7 +355,7 @@ public sealed class AssetResolver : IDisposable
         foreach (var a in _archives)
             if (snap.Tables.TryGetValue(a.Path, out var t) && t.Contains(rel))
                 bsa.Add((new PlacementSource(Path.GetFileName(a.Path), AssetKind.Bsa,
-                    LooseFilePath: null, ArchivePath: a.Path, EntryPath: rel), a.PluginRank));
+                    LooseFilePath: null, ArchivePath: a.Path, EntryPath: rel, OwningMod: a.OwningMod), a.PluginRank));
         // Higher plugin rank wins; the archive filename is a DETERMINISTIC tie-break so equal-rank BSAs (a plugin can
         // ship more than one) order stably across runs rather than by hash/enumeration order.
         var bsaOrdered = bsa.OrderByDescending(b => b.rank)
@@ -379,28 +384,53 @@ public sealed class AssetResolver : IDisposable
         return new PlacementResolution(rel, sources, sources.Count > 1, snap.Failures.Count > 0);
     }
 
-    /// <summary>Is <paramref name="name"/> a provider name this build already knows — an enabled mod folder,
-    /// "overwrite", "Data", or an ACTIVE archive's filename? A name that passes here is answered by the built
-    /// universe and never reaches disk, which keeps the off-order lane (<see cref="OffOrderAssetSource"/>) additive
-    /// rather than a change to any enabled name. Matched OrdinalIgnoreCase, the same way the providers themselves
-    /// are, so this test and the match cannot disagree.</summary>
-    public bool IsUniverseProviderName(string name)
+    /// <summary>Is <paramref name="name"/> a RESERVED provider name — one that names something other than a mod
+    /// folder under <c>mods\</c>: MO2's "overwrite" layer, the game's "Data" folder, or an ACTIVE archive's
+    /// filename? These are the names the folder-scan lane must never be handed, because a mod folder literally
+    /// called <c>Data</c> would then be served for the name "Data" and shadow the layer that name means.
+    ///
+    /// <para>An enabled MOD FOLDER name is deliberately NOT reserved. The named pole reaches the folder scan only
+    /// after the active universe has already failed to answer for that name at that path, so falling through costs
+    /// nothing an enabled name already had and gains the one thing a built universe cannot supply: the folder's own
+    /// root archives, including any the engine does not load (#388, part iii — an enabled mod's name and the same
+    /// mod unticked now reach the same places).</para>
+    ///
+    /// <para>Matched OrdinalIgnoreCase, the same way the providers themselves are, so this test and the match cannot
+    /// disagree.</para></summary>
+    public bool IsReservedProviderName(string name)
     {
-        foreach (var (rootName, _) in _looseRoots)
-            if (string.Equals(rootName, name, StringComparison.OrdinalIgnoreCase)) return true;
+        if (_overwriteDir.Length > 0 && string.Equals("overwrite", name, StringComparison.OrdinalIgnoreCase)) return true;
+        if (_dataDir.Length > 0 && string.Equals("Data", name, StringComparison.OrdinalIgnoreCase)) return true;
         foreach (var a in _archives)
             if (string.Equals(Path.GetFileName(a.Path), name, StringComparison.OrdinalIgnoreCase)) return true;
         return false;
     }
 
+    /// <summary>Is <paramref name="name"/> a mod folder the active profile enables?</summary>
+    bool IsEnabledMod(string name)
+    {
+        foreach (var mod in _enabledMods)
+            if (string.Equals(mod, name, StringComparison.OrdinalIgnoreCase)) return true;
+        return false;
+    }
+
     /// <summary>The copy of <paramref name="relPath"/> inside the mod folder <paramref name="providerName"/> names,
-    /// when that name is NOT one this build knows — the off-order source lane. Thin delegation:
-    /// <see cref="OffOrderAssetSource"/> owns the lane, this resolver only supplies the two facts it holds (where the
-    /// mods folder is, and what the universe already answers for). Null when the name is a universe name, is not a
+    /// when the built universe has no answer for that name at that path — the named-folder source lane. Thin
+    /// delegation: <see cref="OffOrderAssetSource"/> owns the lane, this resolver only supplies the two facts it
+    /// holds (where the mods folder is, and which names are reserved). Null when the name is reserved, is not a
     /// plain folder name, or that folder supplies no copy — and the result says WHICH of those it was, because a
-    /// refusal may only claim the folder was searched when it was.</summary>
+    /// refusal may only claim the folder was searched when it was.
+    ///
+    /// <para>The lane reaches an ENABLED mod's folder too (see <see cref="IsReservedProviderName"/>), so the
+    /// <see cref="PlacementSource.OffOrder"/> flag is corrected here: a copy read out of a mod MO2 is loading is not
+    /// off-order, and labelling it so would tell the caller the game is not loading that mod.</para></summary>
     public OffOrderLookup TryResolveOffOrderProvider(string? providerName, string relPath)
-        => OffOrderAssetSource.Resolve(_modsDir, IsUniverseProviderName, providerName, relPath);
+    {
+        var look = OffOrderAssetSource.Resolve(_modsDir, IsReservedProviderName, providerName, relPath);
+        return look.Source is { } s && IsEnabledMod(s.ProviderName)
+            ? look with { Source = s with { OffOrder = false } }
+            : look;
+    }
 
     /// <summary>Resolve many paths in one call against ONE pinned build, so the whole scan stays internally
     /// consistent even if a RefreshIfStale lands mid-scan. To pair the scan with its read-failure list off the same
--- a/src/housecarl-core/AssetResolver.cs
+++ b/src/housecarl-core/AssetResolver.cs
@@ -60,9 +60,13 @@ public sealed record AssetHit(string RelPath, bool Exists, AssetProvider? Winner
 /// from the provider name alone.</para>
 /// <para><see cref="OwningMod"/> is the MO2 layer a BSA source's archive file physically lives in — a mod folder
 /// name, "overwrite" or "Data" — so naming that mod addresses its own archives, not only its loose files (#388).
-/// Null for a loose source, whose <see cref="ProviderName"/> already IS that layer.</para></summary>
+/// Null for a loose source, whose <see cref="ProviderName"/> already IS that layer.</para>
+/// <para><see cref="OwnerEnabled"/> qualifies <see cref="OffOrder"/> with WHY the game is not loading this copy,
+/// which the two cases answer differently: MO2 does not tick the folder, or it does and this is a root archive no
+/// active plugin binds. Meaningless unless <see cref="OffOrder"/> is true.</para></summary>
 public sealed record PlacementSource(string ProviderName, AssetKind Kind, string? LooseFilePath, string? ArchivePath,
-                                     string EntryPath, bool OffOrder = false, string? OwningMod = null);
+                                     string EntryPath, bool OffOrder = false, string? OwningMod = null,
+                                     bool OwnerEnabled = false);
 
 /// <summary>Concrete-source resolution of one asset path for PLACEMENT (place_asset's auto-resolve when no explicit
 /// source= is given): every provider with its on-disk descriptor, winner FIRST (the same precedence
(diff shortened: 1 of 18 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1220,40 +1220,75 @@ static NifInspectData NifInspectOne(AssetResolver.AssetView view, string rel, st
...
         if (!string.IsNullOrWhiteSpace(mod))
         {
-            var pick = place.Sources.FirstOrDefault(s => s.ProviderName.Equals(mod!.Trim(), StringComparison.OrdinalIgnoreCase));
-            if (pick is null)
-                return new NifInspectData(rel, null, providers, place.Ambiguous, false, null,
-                    $"mod '{mod!.Trim()}' does not provide this path. Providers (winner first): {string.Join(", ", providers.Select(p => p.Name + " (" + p.Kind + ")"))}.");
-            chosen = pick;
+            var pick = NifPick(view, place, rel, mod!.Trim());
+            if (pick.Error is not null)
+                return new NifInspectData(rel, null, providers, place.Ambiguous, false, null, pick.Error);
```
Avick3110: **`mod=` refusals drop the scan-incomplete hedge from an ABSENT.**

Every `mod=` error path returns `Absent: false`, but `NifPick`'s winner-pole arm (line 1284) produces an ABSENT sentence of its own. The renderer keys the batch-level caveats on `d.Absent` (`NifTools.cs:355`), so those caveats are suppressed exactly where they apply.

Scenario: a BSA fails to read this build and the mesh lives inside it. `nif_inspect mesh_paths=[X]` prints the ABSENT plus `[!] but an archive failed to read this build ... "ABSENT" may be incomplete`. The same call with `mod="*winner"` prints a flat, unqualified ABSENT.

Worth noting the asymmetry inside `NifPick` itself: the named-miss arm (line 1286) threads `place.ReadIncomplete` into its sentence; the winner-pole arm at 1284 threads nothing, and neither sets `Absent`.

src/housecarl-core/AssetResolver.cs
```diff
@@ -379,28 +388,56 @@ internal PlacementResolution ResolveForPlacement(string relPath, Snapshot snap)
...
+    /// finds there is still one the game is NOT loading: the built universe already answers for an enabled mod's
+    /// loose tree and every archive the engine loads, so what is left to find is a root archive no active plugin
+    /// binds. <see cref="PlacementSource.OffOrder"/> therefore stays set, and
+    /// <see cref="PlacementSource.OwnerEnabled"/> is filled in here so a caller's sentence can say WHICH of the two
+    /// reasons it is rather than hedging between them.</para></summary>
     public OffOrderLookup TryResolveOffOrderProvider(string? providerName, string relPath)
-        => OffOrderAssetSource.Resolve(_modsDir, IsUniverseProviderName, providerName, relPath);
+    {
+        var look = OffOrderAssetSource.Resolve(_modsDir, IsReservedProviderName, providerName, relPath);
+        return look.Source is { } s && IsEnabledMod(s.ProviderName)
```
Avick3110: **`OwnerEnabled` is set without checking the source is a BSA, so the enabled-mod provenance sentence can fire for a loose file and assert something false about it.**

`OffOrderAssetSource.Resolve` probes the loose file FIRST (`OffOrderAssetSource.cs:106-110`). The doc above says the built universe always answers for an enabled mod's loose tree, but `SafeListFilenames` (`AssetResolver.cs:588`) swallows every enumeration failure and returns null — a deny-ACL'd subtree, a disconnected drive, a transiently locked folder all contribute nothing to `Present`. The universe then has no source under that name, the named pole falls through to this lane, and the loose arm finds the file on the direct path.

`OwnerEnabled` is then true on a `Loose` source, and `WriteSentences.PlaceSourceOffOrder` renders "read from 'ModX', out of a root archive the engine does NOT load (no active plugin binds it)" for a plain loose file the game *is* loading.

```suggestion
        return look.Source is { } s && s.Kind == AssetKind.Bsa && IsEnabledMod(s.ProviderName)
```

Separately: no probe covers the `OwnerEnabled: true` arm at all. `NifSourceLaneProbe`'s enabled-archive mod (`Donor Mod (SE)`) is bound by an active plugin, so it resolves in-universe with `OffOrder: false` — the archive arm this PR adds to `PlaceSourceOffOrder` is never rendered by CI.

src/housecarl-mcp/WriteSentences.cs
```diff
@@ -532,9 +533,17 @@ internal static string PlaceSourceNamedAbsent(string provider, string rel, IRead
     /// <summary>The provenance line for bytes read out of a mod the active profile does NOT include. About the SOURCE
     /// and nothing else: the placed copy's own "does not win until you enable + sort" is the render's separate,
     /// unconditional line, and neither fact may be stated twice.</summary>
-    internal static string PlaceSourceOffOrder(string provider) =>
-        $"read from '{provider}', a mod folder that is NOT enabled in MO2 — you named it, so houseCARL read it off "
-      + "disk; the bytes are that mod's, and enabling it is not required for the copy just placed";
+    // TODO(#388): place_asset's render still calls this without the flag, so an enabled mod's unloaded archive gets
+    // the "NOT enabled in MO2" arm there. The fix is one argument at PlaceAssetTools.cs's call site, which the place
+    // rewrite owns this wave. The NIF surface passes it.
+    internal static string PlaceSourceOffOrder(string provider, bool ownerEnabled = false) => ownerEnabled
```
Avick3110, line 536: **The `TODO` is not just an unfinished nicety — this PR makes `place_asset` print a statement about the user's install that is false.**

Dropping enabled mod names from the reserved-name gate is what makes it reachable. Scenario: mod "DonorMod" is ticked in MO2 and its only copy of a mesh sits inside `mods\DonorMod\Extra.bsa`, which no active plugin binds. Before this change `IsUniverseProviderName` answered that name and the folder scan never ran. Now it falls through, the archive arm finds the entry, the placement succeeds with `OffOrder: true`, and `PlaceAssetTools.cs:223` renders the default arm: *"read from 'DonorMod', a mod folder that is NOT enabled in MO2"*.

That is the sentence `OffOrderAssetSource.IsPlainFolderName`'s comment called "the one sentence this lane exists to keep honest" — a line this PR deletes. Given `PlaceSourceOffOrder` already takes the flag, threading it at the one call site looks smaller than the lane boundary implies; if it genuinely has to wait for the `place` rewrite, the alternative is to keep the default arm honest by hedging it until the flag is passed everywhere.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1372,7 +1418,7 @@ public NifSetResult NifSet(string relPath, IReadOnlyList<NifSetOp> ops, string?
                 return NifSetResult.Fail($"wrote '{rel}' but its on-disk size ({size}) does not match the {editedBytes.Length} verified byte(s) — verify before relying on it.", providers, profileName);
             }
 
-            string? winner = providers.Count > 0 ? $"{providers[0].Name} ({providers[0].Kind})" : null;
+            string? winner = providers.Count > 0 ? providers[0].Text : null;
```
Avick3110, line 1424: **`winner` is now newly reachable as null, and the render turns that into an instruction that cannot be followed.**

Answering `mod=` ahead of the ABSENT return means a successful `nif_set` can finish with `place.Sources` empty. `winner` is then null, and `NifSetWire` falls back to the literal phrase: *"TO MAKE IT WIN: enable this folder in MO2 and sort it ABOVE the current winner"* — there is no current winner to sort above.

Concretely: `nif_set` on a mesh only an unticked donor supplies, with `mod="Unticked Donor"`. `place_asset` already handles the same state honestly (`PlaceAssetTools.cs:227`: *"nothing else provides this path — once '<folder>' is enabled, the placed copy wins"*); the NIF lane wants the same branch.

Avick3110:
All four findings folded, each with an arm that was red before the fix.

**1. `nif_inspect`'s `mod=` error paths never set `Absent`** — 4c95599. `NifPick` now returns the flag, true on the winner-pole arm (that sentence is the same absence the no-`mod=` arm reports) and false on a named miss (that one says which mod and carries its own inline scan caveat, so it must not also draw the ABSENT-worded hedges). Probe arm in `NifSourceLaneProbe`: a build with an unreadable archive, `mod="*winner"` over a path nothing active provides — the plain ABSENT hedges, and now so does the pole.

One deliberate departure. You also asked me to thread `place.ReadIncomplete` into the winner-pole sentence the way the named-miss arm does. I did not: `place.ReadIncomplete` and the batch-level `d.BsaFailures.Count > 0` that keys the hedge are the same fact (`_snap.Failures.Count > 0` both), so doing both prints "an archive failed to read this build" twice in one mesh block. `Absent` alone is strictly the better half — it also draws the *discovery*-incomplete hedge, which the inline sentence has no equivalent for. The named-miss arm threads it inline precisely because it cannot use `Absent`. Say the word if you want both anyway.

**2. `OwnerEnabled` set without checking the source is a BSA** — fcddd23. Applied your suggestion verbatim. Probe arms in `PlaceAssetProbe` section N, over an enabled mod that holds both a loose file and a root archive no active plugin binds: the loose find comes back `OwnerEnabled=false`, the archive find `true`. There was also no arm anywhere for the `OwnerEnabled:true` case at all, so N1 adds the end-to-end one.

**3. The TODO at `WriteSentences.cs` ~539** — 8949188. `PlaceResult` gained `SourceOffOrderOwnerEnabled`, set from `pick.Source.OwnerEnabled`, and `PlaceAssetTools.cs` passes it. TODO deleted. N1 is this arm too: an enabled mod whose bytes came from its unbound archive is now described as an archive the engine skips, not as a mod that is not ticked.

Note for whoever merges second: **#535 replaces `PlaceAssetTools.cs` with `PlaceTools.cs`.** Whichever of #535 and #545 lands second has to carry this one argument (`r.SourceOffOrderOwnerEnabled` into `WriteSentences.PlaceSourceOffOrder`) across onto `PlaceTools` on the rebase. The `PlaceResult` field itself lives in `LoadOrderService.cs` and is unaffected either way.

**4. `nif_set` can succeed with a null winner** — 27dfe24. Added the branch `place_asset` has: with no current winner the render now says nothing else provides the path and names the folder to enable, instead of "sort it ABOVE the current winner". Probe arm: `nif_set` with `mod=` naming an unticked donor that is the path's sole provider.

The two `nif_*` probe arms landed in the same commit as finding 1's fix rather than split across 4c95599 and 27dfe24 — they are one `--- #545 ---` block in `NifSourceLaneProbe` and splitting the block read worse than the commit boundary was worth.

Build clean, `dotnet test --filter "tier!=bridge"` 1056/1056, `ci-all` 128/128. Changelog entry added under Unreleased.

PR #548, opened 2026-09-05
Name the working call in a collection element's refusals

### LINKED ISSUE #319: [GAP] a polymorphic collection element's refusals don't name the working call — 4 guesses to edit one dict element
*Reported by **jouten** (external report, 2026-08-07), filed here largely verbatim.*

### Goal

Flip two booleans on an AI Package's `Data` (a dict of `APackageData` arms) — `Data[7]` and `Data[27]` from False to True, to match a vanilla sandbox package.

### What fell short

Every refusal was CORRECT and clearly worded, but each one pointed at a shape that then failed for a DIFFERENT reason, so it took four `bulk_apply` calls to land the edit. The sequence:

1. `field_path: "Data[7].Data"` →
   > Field 'Data' exists on several arms of 'APackageData' with CONFLICTING shapes ('PackageDataBool': scalar bool vs 'PackageDataInt': scalar uint) — the validator cannot pick one statically. Read the element first to learn its concrete arm, then target a field whose shape is unambiguous.

   I HAD already read it (it is `PackageDataBool`); the advice does not say how to *use* that knowledge.

2. `field_path: "Data"`, `verb: SetAtIndex`, `key: "7"`, compose `PackageDataBool` →
   > SetAtIndex is only valid on list; 'Data' is dict.

3. `field_path: "Data[7]"`, `verb: Set`, compose →
   > Path 'Data[7]' brackets a collection element at the LEAF; brackets navigate mid-path only. To edit a list/dict element, target the collection field and use the verb + Key (SetAtIndex/Set/Remove).

4. `field_path: "Data"`, `verb: Set`, `key: "7"`, compose `PackageDataBool {Data: "True"}` → **worked.**

All four via `housecarl_bulk_apply` in-place (1–3 refused, 4 applied).

**Frequency: recurs.** The SAME class hit the reporter earlier the same day on a different surface: setting `Conditions[i].ComparisonValue` on a CTDA refused with the identical "several arms with CONFLICTING shapes" text (`ConditionGlobal` formlink vs `ConditionFloat` float), and the fix there was to recompose the whole element. Polymorphic collection elements are common (package data, conditions, VMAD properties), so this is a whole-class papercut rather than a one-off.

### Workaround used

None needed in the end — shape 4 is correct, it just is not discoverable from the refusals. Cost: 3 wasted calls per edit of this class.

### Ideal shape

Make the refusals name **the working call for the container they are actually on**. Both fixes are purely corpus-static — no on-disk read, no instance awareness:

- **Refusal #3** ([`CorpusRulebook.cs:226`](https://github.com/Avick3110/houseCARL/blob/main/src/housecarl-core/CorpusRulebook.cs#L226)) has the collection's `Cardinality` in hand at that point, having just walked it. It should emit the container-specific call — `field_path='Data', verb='Set', key='7'` for a dict — instead of the generic `SetAtIndex/Set/Remove`, which names the one verb that is *wrong* here (see refusal #2). `WriteEngine.cs:1729`'s twin already half-splits list vs dict; the two recognisers must stay in agreement, per the existing pattern in that file.
- **Refusal #1** ([`CorpusRulebook.cs:1055`](https://github.com/Avick3110/houseCARL/blob/main/src/housecarl-core/CorpusRulebook.cs#L1055)) likewise knows from the path walk that the bracket sat on a dict, so it can name the call shape and say *"compose the arm you read"*. It genuinely cannot name the arm — that is the point of the refusal — but naming the shape is what turns the advice into an executable remedy.

Either fix alone collapses the sequence 4 calls → 2; both → 1.

### Scoped OUT (decided at triage)

The reporter's secondary ask — *allow the scalar path `Data[7].Data` when the element's concrete arm on disk is unambiguous* — is **not** taken. Pre-flight is corpus-static by design; the engine resolves on the element's runtime type and fails loud on a real mismatch. Making the validator instance-aware to save a compose is an architecture change for an ergonomic win the two message fixes already deliver.

### Notes

- Reproduces unchanged on the 2.0 surface: W3 PR 1 moved `bulk_apply` → `housecarl_apply`, but the path/verb/key grammar and every `CorpusRulebook` pre-flight refusal text carry over untouched.
- This is the PR #311 review doctrine — *a notice must name an **executable** remedy* — applied to the pre-flight refusal lane rather than the render lane.
- Adjacent but distinct: #302 (insert-at-index on modeled lists), #303 (compact per-row condition view).
- Size: a `CorpusRulebook` message change plus guard arms. Natural home is W3 PR 4 (`check`) rather than its own PR.

Avick3110:
Editing one element of a polymorphic collection took four calls: every refusal along the way was correct, and none of them named a call that works. `Data[7].Data` on an AI package refused because the arms disagree on the field's shape, and then told the caller to read the element first — which they had already done — without saying how to use that. `Data[7]` at the leaf refused with the rule and a verb menu but no path and no key. Both refusals already hold, from the path walk they just did, everything the working call needs.

Now they print it. The over-arms conflict remembers the bracketed hop that produced the type it is standing on — the container's dotted path, the key the caller typed, and the field's schema — and ends with the container call to make once the arm is known, with the verbs derived from that collection's shape through `WriteVerbs`, so a dict caller cannot be handed `SetAtIndex`. The hop is cleared on every plain descent, so the remedy only ever describes the element the caller is actually inside; with no bracketed hop it states the rule as before. The leaf-bracket refusal gains the same two facts, `field_path=` and `key=`, and its engine twin in `WriteEngine` gains them too so the two recognisers keep saying the same thing.

Both fixes are corpus-static — no on-disk read, no instance awareness. The reporter's secondary ask (accepting the scalar path when the element's arm on disk is unambiguous) is not taken; it was scoped out at triage.

The wrong-cardinality verb refusals (`SetAtIndex is only valid on list; 'Data' is dict.`) are the same class and still dead-end. They were left alone: with these two fixes the reported sequence is one call, so naming the working verbs there does not save one.

Closes #319

### DIFF
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -513,10 +513,15 @@ public sealed class CorpusRulebook
     /// lane cannot produce.</summary>
     static string OwnedChildSetRefusal(FieldSchema leaf) =>
         $"'{leaf.Name}' holds an owned child RECORD ({leaf.TypeRef}): a record is not a part of its parent, so it is " +
-        "neither built from parts (compose= / composes=) nor set from a value (value=). Address the child record " +
-        $"itself by its own FormID — read the parent at depth=2 and the '{leaf.Name}' field shows it. A path through " +
+        $"neither built from parts (compose= / composes=) nor set from a value (value=). {AddressChildByFormId(leaf.Name)} A path through " +
         "the parent reaches a child only when the record being written already carries one, which a patch's fresh " +
         "override of a parent never does; giving a parent a child it lacks is an open gap (#350).";
+
+    /// <summary>How an owned child record that ALREADY EXISTS is written: on the record axis, by its own FormID.
+    /// One sentence, shared by the value-shaped Set refusal and the element remedy, so the two doors cannot drift
+    /// on where the caller is being sent.</summary>
+    static string AddressChildByFormId(string fieldName) =>
+        $"Address the child record itself by its own FormID — read the parent at depth=2 and the '{fieldName}' field shows it.";
 
     /// <summary>True iff a leaf is the COLLECTION form of the owned-child shape — a list/dict whose ELEMENT is an
     /// owned child record (<c>Cell.Persistent</c>, <c>DialogTopic.Responses</c>, the typed record groups). The
@@ -1157,14 +1162,22 @@ public sealed class CorpusRulebook
     /// collection element the WORKING call is a corpus fact the walk already has, so it is named: the container's
     /// path, the caller's own key, and the verbs that shape takes. Without a bracketed hop there is no container to
     /// name, so it states the rule instead.
-    /// <para>The verbs are the PLACING-one filter, not the keyed one: this sentence promises to write the whole
-    /// element in one call, and <see cref="WriteVerbs.HowToAddress"/> would answer it with Remove — a verb that
-    /// composes nothing and deletes the element the caller came to edit. <see cref="WriteVerbs.HowToPlaceOne"/> is
-    /// the filter that matches the sentence.</para></summary>
+    /// <para>The verbs are the placing-one-AT-a-key filter, not the keyed one: this sentence promises to write the
+    /// whole element in one call, and <see cref="WriteVerbs.HowToAddress"/> would answer it with Remove — a verb
+    /// that composes nothing and deletes the element the caller came to edit. It is not the plain
+    /// <see cref="WriteVerbs.HowToPlaceOne"/> either: that names a list's keyless Add first, and the caller reading
+    /// this has just been handed a key. <see cref="WriteVerbs.HowToPlaceOneAt"/> is the filter that matches the
+    /// sentence.</para>
+    /// <para>An OWNED-RECORD element has no such call at all — the element is a record, reached on the record axis
+    /// by its own FormID — so that shape names no container path and no key: printing them would offer a call
+    /// nothing consumes.</para></summary>
     string ElementRemedy(ElementHop? elementHop) =>
         elementHop is { } h && WriteVerbs.OfField(h.Field, _corpus) is { } shape
-            ? $"Read the element to learn its concrete arm, then write the whole element in one call, composing that " +
-              $"arm: {h.Slot}='{h.Path}', key='{h.Key}' — {WriteVerbs.HowToPlaceOne(shape)}."
+            ? shape.Element == ElementPlacement.OwnedRecord
+                ? $"'{h.Field.Name}' holds owned child RECORDS ({h.Field.ElementTypeRef}), so no write verb reaches " +
+                  $"the element through its parent. {AddressChildByFormId(h.Field.Name)}"
+                : $"Read the element to learn its concrete arm, then write the whole element in one call, composing " +
+                  $"that arm: {h.Slot}='{h.Path}', key='{h.Key}' — {WriteVerbs.HowToPlaceOneAt(shape)}."
             : "Read the element first to learn its concrete arm, then target a field whose shape is unambiguous.";
 
     string FieldNotFound(TypeSchema owner, string name)
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -118,10 +118,16 @@ public sealed class CorpusRulebook
         // sit mid-path: the hops through it are where a verb acts on a record the call never named.
         FieldSchema? ownedChildHop = null;
         TypeSchema? ownedChildHopOwner = null;
+        // …and the bracketed hop that produced the type being walked RIGHT NOW, for the one refusal that cannot name
+        // the arm it needs (FindField's conflicting-shapes reject). Cleared on every plain hop, so it only ever
+        // describes the element the caller is standing in.
+        ElementHop? elementHop = null;
         for (int i = 0; i < req.Path.Length - 1; i++)
         {
             if (!TrySeg(req.Path[i], out var segName, out var segKey, out var segErr)) return segErr;
-            var field = FindField(current, segName, out _, out var polyErr);
+            var incoming = elementHop;
+            elementHop = null;
+            var field = FindField(current, segName, out _, out var polyErr, incoming);
             if (polyErr is not null) return polyErr;
             if (field is null) return FieldNotFound(current, segName);
             if (ownedChildHop is null && SchemaClassifier.IsOwnedChildRecord(field, _corpus))
@@ -204,6 +210,7 @@ public sealed class CorpusRulebook
                     && CheckValue(field.KeyType, segKey, $"dict key for '{segName}'", DictKeyType(field)?.AssemblyQualifiedName) is { } ke)
                     return ke;
                 current = elem;
+                elementHop = new ElementHop(PathTo(req.Path, i, segName), segKey, field);
             }
         }
 
@@ -225,12 +232,15 @@ public sealed class CorpusRulebook
             // caller is never handed index verbs nor a list caller key verbs. The verbless fallback is still reachable
             // — a bracketed typo ('Nope[0]') resolves no field — so it states the rule without naming verbs.
             var head = $"Path '{req.Path[^1]}' brackets a collection element at the LEAF; brackets navigate mid-path only. ";
+            // The remedy carries the caller's OWN key, not just the rule: the bracket they typed already says which
+            // element they meant, so the message can hand back the call that works rather than a shape to fill in.
             if (bracketed is not null && WriteVerbs.OfField(bracketed, _corpus) is { } bshape)
-                return head + $"Target the collection field '{leafName}' itself and address the element with the verb "
-                       + $"+ Key — {WriteVerbs.HowToAddress(bshape)}.";
+                return head + $"Target the collection field itself and address the element with the verb + Key: "
+                       + $"field_path='{PathTo(req.Path, req.Path.Length - 1, leafName)}', key='{leafKey}' — "
+                       + $"{WriteVerbs.HowToAddress(bshape)}.";
             return head + "Target the collection field itself and address the element with the verb + Key.";
         }
-        var leaf = FindField(current, leafName, out var leafOwner, out var leafPolyErr);
+        var leaf = FindField(current, leafName, out var leafOwner, out var leafPolyErr, elementHop);
         if (leafPolyErr is not null) return leafPolyErr;
         if (leaf is null) return FieldNotFound(current, leafName);
 
@@ -1108,6 +1118,27 @@ public sealed class CorpusRulebook
         catch (Exception ex) { name = segment; key = null; error = ex.Message; return false; }
     }
 
+    /// <summary>The bracketed hop that produced the type currently being walked: the collection field's own dotted
+    /// path, the key the caller indexed it with, and its schema. Held for the IMMEDIATELY preceding hop only, so a
+    /// refusal raised on the element's type can name the call that rewrites that element and nothing else.</summary>
+    readonly record struct ElementHop(string Path, string Key, FieldSchema Field);
+
+    /// <summary>The dotted field_path of the hop at <paramref name="index"/>, with its own bracket dropped —
+    /// earlier hops keep theirs, because they are still navigation.</summary>
+    static string PathTo(string[] path, int index, string name) =>
+        string.Join(".", path.Take(index).Append(name));
+
+    /// <summary>What to do about a field the over-arms search refused to pick an arm for. The refusal genuinely
+    /// cannot name the arm — that is what it is declining to guess — but when the caller is standing inside a
+    /// collection element the WORKING call is a corpus fact the walk already has, so it is named: the container's
+    /// path, the caller's own key, and the verbs that shape takes. Without a bracketed hop there is no container to
+    /// name, so it states the rule instead.</summary>
+    string ElementRemedy(ElementHop? elementHop) =>
+        elementHop is { } h && WriteVerbs.OfField(h.Field, _corpus) is { } shape
+            ? $"Read the element to learn its concrete arm, then write the whole element in one call, composing that " +
+              $"arm: field_path='{h.Path}', key='{h.Key}' — {WriteVerbs.HowToAddress(shape)}."
+            : "Read the element first to learn its concrete arm, then target a field whose shape is unambiguous.";
+
     string FieldNotFound(TypeSchema owner, string name)
     {
         var sample = owner.Fields.Select(f => f.Name).Take(12).ToList();
@@ -1128,7 +1159,8 @@ public sealed class CorpusRulebook
     /// turns out to be — the engine then resolves on the element's RUNTIME type and fails loud on a real mismatch);
     /// arms that DISAGREE reject by name, never guess. <paramref name="effectiveOwner"/> is the schema the found
     /// field belongs to (the arm for an arm-found field), so downstream messages name the real owner.</summary>
-    FieldSchema? FindField(TypeSchema owner, string name, out TypeSchema effectiveOwner, out string? error)
+    FieldSchema? FindField(TypeSchema owner, string name, out TypeSchema effectiveOwner, out string? error,
+        ElementHop? elementHop = null)
     {
         effectiveOwner = owner; error = null;
         if (owner.Fields.FirstOrDefault(f => f.Name == name) is { } direct) return direct;
@@ -1157,8 +1189,7 @@ public sealed class CorpusRulebook
             {
                 error = $"Field '{name}' exists on several arms of '{owner.Name}' with CONFLICTING shapes " +
                         $"('{firstArm.Name}': {firstField.Cardinality} {firstField.Type} vs '{arm.Name}': {f.Cardinality} {f.Type}) — " +
-                        "the validator cannot pick one statically. Read the element first to learn its concrete arm, " +
-                        "then target a field whose shape is unambiguous.";
+                        "the validator cannot pick one statically. " + ElementRemedy(elementHop);
                 return null;
             }
         effectiveOwner = firstArm;
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -105,8 +105,13 @@ public sealed class CorpusRulebook
     }
 
     /// <summary>Validate a write rooted at an arbitrary type — a record OR a struct being built from parts (so a
-    /// <see cref="StructSpec"/>'s nested writes validate by the identical leaf/path rules, recursively).</summary>
-    string? ValidateFromType(TypeSchema root, WriteRequest req, IReadOnlyCollection<string>? siblingEditorIds = null)
+    /// <see cref="StructSpec"/>'s nested writes validate by the identical leaf/path rules, recursively).
+    /// <para><paramref name="pathSlot"/> is what the caller's OWN input slot for <see cref="WriteRequest.Path"/> is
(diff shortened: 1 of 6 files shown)


src/housecarl-core/CorpusRulebook.cs
```diff
@@ -225,12 +232,15 @@ public static Corpus LoadCorpus(string corpusJsonPath)
             // caller is never handed index verbs nor a list caller key verbs. The verbless fallback is still reachable
             // — a bracketed typo ('Nope[0]') resolves no field — so it states the rule without naming verbs.
             var head = $"Path '{req.Path[^1]}' brackets a collection element at the LEAF; brackets navigate mid-path only. ";
+            // The remedy carries the caller's OWN key, not just the rule: the bracket they typed already says which
+            // element they meant, so the message can hand back the call that works rather than a shape to fill in.
             if (bracketed is not null && WriteVerbs.OfField(bracketed, _corpus) is { } bshape)
-                return head + $"Target the collection field '{leafName}' itself and address the element with the verb "
-                       + $"+ Key — {WriteVerbs.HowToAddress(bshape)}.";
+                return head + $"Target the collection field itself and address the element with the verb + Key: "
+                       + $"field_path='{PathTo(req.Path, req.Path.Length - 1, leafName)}', key='{leafKey}' — "
```
Avick3110: **`field_path=` is wrong when this refusal is raised inside a compose's nested `sets`.**

`ValidateFromType` is shared: `StructSpecContents` recurses into it for every `StructSpec.Sets` entry, with `req.Path` rooted at the struct being built, not at the record. Both new messages now print a concrete `field_path='…'` built from that struct-relative path — and inside a compose the slot is spelled `path=`, not `field_path=`.

Concrete: `op=Add, field_path='VirtualMachineAdapter.Scripts', compose={type:'ScriptEntry', sets:[{path:'Properties[0]', op:'Set'}]}` refuses with

> … Target the collection field itself and address the element with the verb + Key: **field_path='Properties'**, key='0' — SetAtIndex …

A caller who follows that literally sends `field_path='Properties'` at the top level and gets `No field 'Properties' on 'Npc'` — the message names a call that does not work, which is exactly the failure mode this PR is fixing elsewhere. Same applies to `ElementRemedy` at line 1139.

The walk knows which root it is validating against; either thread that through and suppress the `field_path=` form for the struct-rooted recursion, or spell the slot for the context.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -225,12 +232,15 @@ public static Corpus LoadCorpus(string corpusJsonPath)
...
+                       + $"{WriteVerbs.HowToAddress(bshape)}.";
```
Avick3110: **The leaf-bracket remedy echoes the key back unvalidated, so it can name a call that cannot work.**

The mid-path bracketed hop validates key SHAPE before using it (the `IsValidListIndexValue` check for lists and the `DictKeyType` + `CheckValue` check for dicts, both with comments explaining why). The leaf branch has no such check — it never needed one, because the old message only stated the rule. Now it prints `key='{leafKey}'` as part of a call to make.

Measured on this branch:

- `Package` / `Data[notasbyte]` → `… field_path='Data', key='notasbyte' — Set (compose= + key=) sets that entry …`. `Data`'s key is an `sbyte`; the suggested call throws `FormatException` at apply, the exact accept-then-throw the mid-path comment calls out.
- `Perk` / `Effects[-3]` → `… field_path='Effects', key='-3' — SetAtIndex …`. A negative index is refused by the leaf key gate.

Either run the same key-shape recognisers here before promoting the key into the remedy, or fall back to the verbless/keyless form when the key does not pass them.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1827,7 +1827,8 @@ public static void ApplyVerb(object record, WriteRequest req)
                        + "mid-path only; ";
             if (bracketProp is not null && WriteVerbs.OfRuntimeType(bracketProp.PropertyType) is { } bshape)
                 throw new InvalidOperationException(head
-                    + $"to operate on an element of '{leafName}', target the field itself and use the verb + Key — "
+                    + "to operate on that element, target the field itself and use the verb + Key: "
```
Avick3110: **The engine twin's change has no test, and nothing checks that the two recognisers agree.**

`CorpusRulebook`'s leaf-bracket refusal and this one are described in the PR body as twins kept "saying the same thing", but there is no test over this throw — before or after. Grepping the test tree for `brackets a collection element`, `verb + Key`, or `to operate on an element of` returns nothing; `ElementRefusalRemedyTests` drives `CorpusRulebook.Validate` only. So the new `field_path=`/`key=` facts here are asserted by the PR body and by nothing else, and the path-join is a hand-rolled second copy of `CorpusRulebook.PathTo` (`req.Path[..^1].Append(leafName)`) that can drift from it silently.

Two smaller things that follow from that: the sibling recognisers elsewhere in this file carry an explicit "its twin is X — two recognisers that must agree" comment (see the `GenderedItem<` pair), and this one does not; and if the throw is in fact unreachable because the rulebook always refuses first, that is worth saying in a comment rather than leaving the reader to work out whether the edit is live.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1108,6 +1118,27 @@ static bool TrySeg(string segment, out string name, out string? key, out string?
...
+
+    /// <summary>What to do about a field the over-arms search refused to pick an arm for. The refusal genuinely
+    /// cannot name the arm — that is what it is declining to guess — but when the caller is standing inside a
+    /// collection element the WORKING call is a corpus fact the walk already has, so it is named: the container's
+    /// path, the caller's own key, and the verbs that shape takes. Without a bracketed hop there is no container to
+    /// name, so it states the rule instead.</summary>
+    string ElementRemedy(ElementHop? elementHop) =>
+        elementHop is { } h && WriteVerbs.OfField(h.Field, _corpus) is { } shape
+            ? $"Read the element to learn its concrete arm, then write the whole element in one call, composing that " +
+              $"arm: field_path='{h.Path}', key='{h.Key}' — {WriteVerbs.HowToAddress(shape)}."
```
Avick3110: **The sentence promises a compose and then hands back a menu that includes `Remove`.**

`HowToAddress` is the *keyed* filter — "how do I reach ONE element" — so on `Package.Data` it renders `Set (compose= + key=) …; Add (compose= + key=) …; Remove (key=) drops that entry`. That menu fits the leaf-bracket site, which frames itself as "address the element". It does not fit this one, which frames itself as "write the whole element in one call, composing that arm" and then names a verb that composes nothing and deletes the element the caller was trying to edit.

`WriteVerbs` already carries the fact needed to fix this without reciting names: the placing filter (`Places`) or its singular form `HowToPlaceOne`, which is precisely "how do I put in ONE element" and drops both `Remove` and the batch slots. On a composed dict that yields `Set (compose= + key=)` and `Add (compose= + key=)` — the two calls this remedy is actually describing.

Avick3110:
All four findings are real, and all four are folded in 0ce4c6b. Nothing refuted.

**1. `field_path=` is the wrong slot inside a compose's nested `sets`.** Correct, and it applied to both new messages plus the engine twin, which is on the same lane (`Instantiate` replays a spec's `Sets` through `ApplyVerb` against the built struct). Took the "spell the slot for the context" option rather than suppressing the form: a nested `sets` entry is a full `WriteRequest`, so `path='Properties', key='0'` there is a call that works, and suppressing it would drop a fact the caller can use. `ValidateFromType` and `ApplyVerb` each take the slot name, defaulting to `field_path`; `StructSpecContents` and the nested-`Sets` replay pass `path`. The slot travels on `ElementHop` alongside the path, because neither means anything without the root it was built against.

**2. The leaf bracket echoed the key unvalidated.** Correct, and both cases reproduced. The mid-path shape checks are now one `KeyShapeError` helper that the mid-path hop and the leaf bracket both call, so they cannot drift on what a usable key is. A key that fails it falls back to the pre-PR sentence — the rule plus the shape-derived verb menu, no path and no key — so the remedy never names a call that throws at apply.

**3. The engine twin was untested and hand-rolled.** Correct on all three counts. It now joins the container path with `CorpusRulebook.PathTo` (made `internal`) instead of a second copy; it carries the "its twin is X — two recognisers that must agree" comment its sibling recognisers carry; and it is driven by tests for the first time. The throw is reachable, not dead — a direct/CLI call bypasses pre-flight, and so does the nested-`Sets` replay — so rather than a comment saying so, there is a test that asks both recognisers about the same bracketed leaf and requires the same container path, the same key, and the same verb menu.

**4. `HowToAddress` contradicts "composing that arm".** Correct. `ElementRemedy` now uses `HowToPlaceOne`, which is the filter that matches the sentence; the leaf-bracket site keeps `HowToAddress`, which matches its own. Tested that the container remedy names no `Remove`.

**Also, unposted but real:** `ANestedContainerIsNamedByItsWholePath` asserted the path and key without asserting which refusal produced them, so a walk that started refusing earlier could have satisfied it. It now pins the leaf-bracket head first.

Seven tests added (11 in the file). Full suite 1171 passed, `ci-all` all pass, changelog updated for the key gate and the compose slot.

Avick3110:
Blind review folded (4 of 4 real, 0 refuted). Queued for Aaron's review. Stacked follow-on: #557 (#350).

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1108,6 +1122,51 @@ static bool TrySeg(string segment, out string name, out string? key, out string?
...
+    /// path, the caller's own key, and the verbs that shape takes. Without a bracketed hop there is no container to
+    /// name, so it states the rule instead.
+    /// <para>The verbs are the PLACING-one filter, not the keyed one: this sentence promises to write the whole
+    /// element in one call, and <see cref="WriteVerbs.HowToAddress"/> would answer it with Remove — a verb that
+    /// composes nothing and deletes the element the caller came to edit. <see cref="WriteVerbs.HowToPlaceOne"/> is
+    /// the filter that matches the sentence.</para></summary>
+    string ElementRemedy(ElementHop? elementHop) =>
+        elementHop is { } h && WriteVerbs.OfField(h.Field, _corpus) is { } shape
+            ? $"Read the element to learn its concrete arm, then write the whole element in one call, composing that " +
+              $"arm: {h.Slot}='{h.Path}', key='{h.Key}' — {WriteVerbs.HowToPlaceOne(shape)}."
```
Avick3110: **The list remedy leads with a verb that appends instead of overwriting — and it succeeds.**

`HowToPlaceOne` returns the `On()` table in table order, and for a composed LIST that order is `Add` first. So the remedy prints a key and then names a keyless verb ahead of the in-place one:

```
GATE Faction.Conditions[0].ComparisonValue:
Field 'ComparisonValue' exists on several arms of 'Condition' with CONFLICTING shapes
('ConditionGlobal': formlink FormLink<IGlobalGetter> vs 'ConditionFloat': scalar float)
— the validator cannot pick one statically. Read the element to learn its concrete arm,
then write the whole element in one call, composing that arm:
field_path='Conditions', key='0'
— Add (compose=) appends a new element at the END;
  SetAtIndex (compose= + key=) overwrites the element already at that index, in place;
  InsertAtIndex (compose= + key=) …
```

A caller who typed `Conditions[0]` bracketed an index that already holds an element. Reading top-down they take `Add`, which the gate ACCEPTS (RemedyVerbsGuardProbe pins every verb this site names as accepted): the list comes back one element longer, element 0 untouched, which on a CTDA OR-run silently changes what the record gates on. It is the one wrong first choice on this branch that does not refuse.

This is the same hazard `WriteVerbs.cs` already calls out for the keyed menu — "SetAtIndex before InsertAtIndex, deliberately … Leading with insert is the one wrong first choice on this branch that does not refuse" — and SITE-BRACKET-ORDER pins that order. `HowToPlaceOne` carries no equivalent guarantee, and this is the site that prints `key='0'` immediately before a verb that ignores it.

Dict containers are fine (`Set` leads). Every list container in the corpus reaches this: `Conditions` on Faction / Effect / ConstructibleObject / DialogResponses / CameraPath, and VMAD `Properties`.

Suggest either ordering the placing-one filter so the address-an-existing-element verb leads on a list, or dropping `key=` from the printed call when the leading verb does not take one.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1108,6 +1122,51 @@ static bool TrySeg(string segment, out string name, out string? key, out string?
...
+    /// <summary>What to do about a field the over-arms search refused to pick an arm for. The refusal genuinely
+    /// cannot name the arm — that is what it is declining to guess — but when the caller is standing inside a
+    /// collection element the WORKING call is a corpus fact the walk already has, so it is named: the container's
+    /// path, the caller's own key, and the verbs that shape takes. Without a bracketed hop there is no container to
+    /// name, so it states the rule instead.
+    /// <para>The verbs are the PLACING-one filter, not the keyed one: this sentence promises to write the whole
+    /// element in one call, and <see cref="WriteVerbs.HowToAddress"/> would answer it with Remove — a verb that
+    /// composes nothing and deletes the element the caller came to edit. <see cref="WriteVerbs.HowToPlaceOne"/> is
+    /// the filter that matches the sentence.</para></summary>
+    string ElementRemedy(ElementHop? elementHop) =>
```
Avick3110, line 1174: **`ElementRemedy` has no arm for an owned-record collection, so it names a call that does not exist.**

`WriteVerbs.HowToPlace`/`HowToPlaceOne` special-case `ElementPlacement.OwnedRecord` precisely because there is no write verb to name — they return a sentence instead of a verb list. The wrapper sentence added here does not, so the two halves contradict each other:

```
GATE Cell.Persistent[0].MajorFlags:
Field 'MajorFlags' exists on several arms of 'Placed' with CONFLICTING shapes
('PlacedNpc': enum PlacedNpc.MajorFlag vs 'PlacedArrow': enum APlacedTrap.MajorFlag)
— the validator cannot pick one statically. Read the element to learn its concrete arm,
then write the whole element in one call, composing that arm:
field_path='Persistent', key='0'
— its elements are owned child RECORDS, which are created on the record axis
  — use housecarl_create with parent= the parent's FormID in its records= element,
    not a write verb.
```

`field_path='Persistent', key='0'` is presented as the call to make and nothing consumes it: no verb takes it, and the real fix is to resolve the placed reference on the record axis and never touch `Cell.Persistent` at all. Reachable on `Cell.Persistent` and `Cell.Temporary` — the `Placed` arms disagree on `MajorFlags`.

The remedy should fall through to the record-axis sentence without naming a container path/key when the shape is `OwnedRecord`.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1819,16 +1823,22 @@ public static void ApplyVerb(object record, WriteRequest req)
...
+            // A property that resolves to no collection at all still gets the rule, with no verb named.
             var head = $"Path segment '{req.Path[^1]}' brackets a collection element at the LEAF. Brackets navigate "
                        + "mid-path only; ";
             if (bracketProp is not null && WriteVerbs.OfRuntimeType(bracketProp.PropertyType) is { } bshape)
                 throw new InvalidOperationException(head
-                    + $"to operate on an element of '{leafName}', target the field itself and use the verb + Key — "
-                    + $"{WriteVerbs.HowToAddress(bshape)}.");
+                    + "to operate on that element, target the field itself and use the verb + Key: "
+                    + $"{pathSlot}='{CorpusRulebook.PathTo(req.Path, req.Path.Length - 1, leafName)}', "
+                    + $"key='{leafKey}' — {WriteVerbs.HowToAddress(bshape)}.");
```
Avick3110: **The engine twin echoes a key the rulebook deliberately withholds, so it names a call that throws at apply.**

The rulebook gates the promoted key through `KeyShapeError` before handing it back — that is what `AMalformedDictKeyIsNotEchoedAsACallToMake` and `ANegativeListIndexIsNotEchoedAsACallToMake` pin. The engine has no equivalent gate; it interpolates `leafKey` unconditionally. Driven directly (the direct/CLI `--op` lane the comment above says this message exists for):

```
GATE bad-dict-leaf: Path 'Data[notasbyte]' brackets a collection element at the LEAF; …
  Target the collection field 'Data' itself and address the element with the verb + Key — …
ENG  bad-dict-leaf: Path segment 'Data[notasbyte]' brackets a collection element at the LEAF. …
  to operate on that element, target the field itself and use the verb + Key:
  field_path='Data', key='notasbyte' — Set (compose= + key=) …

GATE neg-list-leaf: Path 'Effects[-3]' … Target the collection field 'Effects' itself … — …
ENG  neg-list-leaf: … field_path='Effects', key='-3' — SetAtIndex (compose= + key=) …
```

`Package.Data` is sbyte-keyed, so `key='notasbyte'` throws `FormatException` at apply; `key='-3'` is refused by `StepIntoElement` and surfaces as the misleading "real inconsistency" wrapper. Both are exactly the accept-then-throw dead end this PR closes on the gate side.

The comment added just above asserts the two recognisers say "same facts, same order, same slot", and `TheEngineTwinNamesTheSameCallAsTheRulebook` only exercises a well-formed key, so nothing catches the divergence. The engine cannot reach `KeyShapeError` (schema-blind by design), but it can apply the same runtime recognisers it already owns — `IsValidListIndexValue` for a list, and a coercion check against the dict's key type — before promoting the key.

Avick3110:
All three findings reproduce as written, all three are folded in 2cf365b (fix) and ea9fd5e (guard). Nothing refuted.

**1. The list remedy led with a verb that appends.** Real, and measured: `Faction.Conditions[0].ComparisonValue` printed `key='0'` and then `Add (compose=) appends a new element at the END` first. Took the "order the placing-one filter" option, one step further — a new `WriteVerbs.HowToPlaceOneAt` is `HowToPlaceOne` minus the verbs that take no key, so the site that has already printed `key=` cannot name a verb that ignores it. On a list that leaves `SetAtIndex` then `InsertAtIndex` (table order, so SITE-BRACKET-ORDER's guarantee carries); on a dict every placing verb is keyed, so that menu is unchanged. `HowToPlaceOne` itself is untouched — the composes= parenthetical that uses it has no key to contradict.

**2. `ElementRemedy` had no owned-record arm.** Real; `Cell.Persistent[0].MajorFlags` and `Cell.Temporary[0].MajorFlags` both printed a container path and key nothing consumes. The remedy now falls through on `ElementPlacement.OwnedRecord` and names neither: it says the field holds owned child records and sends the caller to the record axis by the element's own FormID. That last sentence is now one shared helper with `OwnedChildSetRefusal`, which already said it — the element remedy is about a child that EXISTS, so it points where that refusal points (address it by FormID), not at the create-oriented sentence `HowToPlace` gives.

**3. The engine twin echoed a key the rulebook withholds.** Real, both cases reproduced. The engine cannot reach `KeyShapeError`, so it applies the two runtime recognisers apply itself keys on — `Coerce(key, keyType)` for a dict, `IsValidListIndexValue` for a list — in one `KeyShapeUsable` helper, and falls back to the keyless form when the key fails. `Data[notasbyte]` and `Effects[-3]` now get the rule and the shape-derived menu with no `field_path=`/`key=`; `Effects[3]` still names the call.

Tests: 4 new in `ElementRefusalRemedyTests` (the keyed-only list menu with its order assertion, the owned-record remedy naming no call, and a two-case theory over the engine's withheld keys) — 11 → 15. Probes: 5 new arms in `remedy-verbs-guard` (`SITE-ELEMENT-CONFLICT-LIST`, `-LIST-KEYED`, `-OWNEDRECORD`, `SITE-BRACKET-ENGINE-BADKEY`, `-WITHHELD`) — 26 → 31; that site had no arm before, which is why finding 1 could ship. Build, 1175 tests and ci-all 128/128 green.

Issue #542, opened 2026-09-04
[Bug] the S2 tools (place, asset_status) have no format=json lane (SPEC §2.1 TRANSPORT)


Avick3110:
Stated in PR #535 (the `place` fold) rather than papered over, and the same on its S2 sibling.

Neither `housecarl_place` nor `housecarl_asset_status` has a `format=json` lane. SPEC §2.1 makes TRANSPORT uniform across every tool and says a surface missing one of these is a bug, not a parity feature request. The two old place tools had no json either, so #535 regressed nothing; #538 gave `asset_status` `limit`/`offset` and accounting but no json.

Fix: one json render for the S2 substrate shared by both tools (per-file rows with winner and provider chain; per-member place outcome rows), with the same accounting fields the text lane carries. Sibling of #539 (the same gap on `skse`).

PR #549, opened 2026-09-05
Carry a merge's light flag, report the localized flag, fix asset_status provider tokens

### LINKED ISSUE #363: [Bug] merge_plugins drops the donor's ESL/ESM header flag, and the report never says so

### LINKED ISSUE #376: [GAP] no tool reports a plugin's localized header flag — the in-place refusal cannot be anticipated

Avick3110:
Backlog item 18, the strings remainder, as one small PR: three independent fixes that each close an issue.

`merge_plugins` used to drop a donor's ESL flag unconditionally, so merging a set of light plugins silently cost a load-order slot. It now carries the flag under the 2026-09-05 ruling: only when every donor was light AND every merged object id landed inside the light window `0x800-0xFFF` (the same bound compact enforces — the merge remap runs to the full 24-bit range, so a large merge lands ids above the ceiling that a light write would throw on). Otherwise the flag is dropped and the report names which of the two conditions failed, because the two take different remedies: a full donor in the set is fixed by running `compact_plugin` on the output, while ids overflowing the window is the one case compact cannot rescue, so the report says that rather than pointing at it. The ESM flag is never carried and that drop is stated as before, and an output written light no longer gets the "want it light? run compact" tail.

`load_order_status`'s `lookup=` now reports a plugin's LOCALIZED header flag when the name resolved to a plugin file, so the in-place write refusal that flag causes is visible before a job runs into it. Three answers rather than a bool: an unreadable header is reported as unknown, never as "not localized".

`asset_status` prints provider names through the same formatter `place_asset` and the NIF tools use — the name inside double quotes, a character a Windows folder or file name cannot contain, with the kind outside them. This is the third surface of #340 (already closed), left unchanged by PR #545 because the render lives in a file that lane was told not to touch.

**Tests.** `dotnet build` clean, 1171 tests pass (1160 baseline + 11), `ci-all` 128/128. New: `MergeLightCarryTests` over its own instance of light, full and master donors covering the carry, both drop reasons and the ESM drop; `StatusLocalizedLookupTests` over a plugin whose TES4 header carries the localized bit; `AssetProviderTokenTests` over a mod folder named `Face Extras (SE)`, whose parenthetical is the case that makes a bare `Name (kind)` render unreadable. The merge probe's HEADER arms pinned the old unconditional drop and are updated to the new contract.

Closes #363
Closes #376

### DIFF
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -1833,7 +1833,8 @@ public sealed class LoadOrderService : IDisposable
     /// included — never a bool.
     /// <para>An INACTIVE plugin gets an answer too: the resolver indexes only the active order, so its path comes from
     /// the same on-disk locate every other lane uses. Reading an inactive plugin is a surface houseCARL advertises, and
-    /// a plugin answer with the localized half silently missing is the worse outcome.</para></summary>
+    /// a plugin answer with the localized half silently missing is the worse outcome. A name several mod folders
+    /// provide is answered from the copy MO2 priority serves, not called unreadable.</para></summary>
     public LocalizedFlagRead? PluginLocalizedFlag(string pluginName)
     {
         LoadOrderResolver.IndexView view;
@@ -1853,9 +1854,15 @@ public sealed class LoadOrderService : IDisposable
                         || comp.ImplicitPluginNames.Any(n => n.Equals(pluginName, StringComparison.OrdinalIgnoreCase));
         if (!isPlugin) return null;
         var loc = LocatePluginFileOnDisk(comp, modsDir, dataDir, overwriteDir, pluginName, null, offerModParam: false);
-        // Listed as a plugin but no single file behind the name (missing, or several folders provide it): the flag is
-        // not established, which is exactly what the third value says.
-        return loc.Path is null ? LocalizedFlagRead.Unreadable : WriteEngine.PluginIsLocalized(loc.Path);
+        if (loc.Path is { } path) return WriteEngine.PluginIsLocalized(path);
+        // Several folders provide the name: every copy is readable and MO2 priority already decides which one serves,
+        // so the served copy answers — the same rule the locate itself judges 'serves' by. UNKNOWN here would claim a
+        // header read failed when none was attempted.
+        if (loc.Ambiguous is { Count: > 0 } hits && hits.FirstOrDefault(h => h.Enabled) is { } served)
+            return WriteEngine.PluginIsLocalized(served.Path);
+        // Listed as a plugin, and no file behind the name serves it: the flag is not established, which is what the
+        // third value says.
+        return LocalizedFlagRead.Unreadable;
     }
 
     /// <summary>Read MO2's OWN local Nexus update cache — the modid / version / newestVersion / ignoredVersion /
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -1829,13 +1829,33 @@ public sealed class LoadOrderService : IDisposable
     /// <summary>The LOCALIZED header flag of ONE plugin, for housecarl_load_order_status' lookup= (#376): a localized
     /// plugin's text lives in .STRINGS files rather than in the plugin, which is what the in-place write lanes refuse
     /// on, so a caller can see that refusal coming instead of meeting it mid-job. Null when the name is not a plugin
-    /// this order resolved to a file; otherwise the three-way read, Unreadable included — never a bool.</summary>
+    /// at all (a mod folder, a typo — nothing has a header to read); otherwise the three-way read, Unreadable
+    /// included — never a bool.
+    /// <para>An INACTIVE plugin gets an answer too: the resolver indexes only the active order, so its path comes from
+    /// the same on-disk locate every other lane uses. Reading an inactive plugin is a surface houseCARL advertises, and
+    /// a plugin answer with the localized half silently missing is the worse outcome.</para></summary>
     public LocalizedFlagRead? PluginLocalizedFlag(string pluginName)
     {
         LoadOrderResolver.IndexView view;
         lock (_gate) { view = Resolver.Capture(); }
-        var path = view.PluginPath(pluginName);
-        return path is null ? null : WriteEngine.PluginIsLocalized(path);
+        if (view.PluginPath(pluginName) is { } activePath) return WriteEngine.PluginIsLocalized(activePath);
+
+        string modsDir, dataDir, overwriteDir, profileDir;
+        try { lock (_gate) { EnsurePathsDerived(); modsDir = _modsDir; dataDir = _dataDir; overwriteDir = _overwriteDir; profileDir = _profileDir; } }
+        catch { return null; }
+        Mo2Composition comp;
+        try { comp = Mo2LoadOrder.ReadComposition(profileDir); }
+        catch { return null; }
+        // Only a name the profile lists as a plugin: a mod folder and a typo both have no header, and inventing an
+        // UNKNOWN line for them would claim there is a plugin here whose flag simply could not be read.
+        bool isPlugin = comp.OrderedPluginNames.Any(n => n.Equals(pluginName, StringComparison.OrdinalIgnoreCase))
+                        || comp.InactivePluginNames.Any(n => n.Equals(pluginName, StringComparison.OrdinalIgnoreCase))
+                        || comp.ImplicitPluginNames.Any(n => n.Equals(pluginName, StringComparison.OrdinalIgnoreCase));
+        if (!isPlugin) return null;
+        var loc = LocatePluginFileOnDisk(comp, modsDir, dataDir, overwriteDir, pluginName, null, offerModParam: false);
+        // Listed as a plugin but no single file behind the name (missing, or several folders provide it): the flag is
+        // not established, which is exactly what the third value says.
+        return loc.Path is null ? LocalizedFlagRead.Unreadable : WriteEngine.PluginIsLocalized(loc.Path);
     }
 
     /// <summary>Read MO2's OWN local Nexus update cache — the modid / version / newestVersion / ignoredVersion /
@@ -6535,8 +6555,9 @@ public sealed class LoadOrderService : IDisposable
                 $"refused — '{outName}' has the .esl extension, which the game engine force-treats as a LIGHT master regardless " +
                 "of the header flag, but a merge never constrains object ids to the light window: it renumbers only what it must " +
                 "(cross-donor collisions, and ids below the write floor), so an id above 0xFFF would be misread in game. Merge to " +
-                "a '.esp' instead, then run " + ToolNames.CompactPlugin + " on it to make it light — that renumbers every id into " +
-                "the light window (the tools compose). Nothing was written.");
+                "a '.esp' instead: if every donor was light and every merged id landed in the window, the output is written LIGHT " +
+                "already; otherwise the report says so, and " + ToolNames.CompactPlugin + " on it renumbers every id into the light " +
+                "window (the tools compose). Nothing was written.");
         if (donorsRaw.Any(d => string.Equals(d, outName, StringComparison.OrdinalIgnoreCase)))
             return WritePatchBuilder.MergeOutcome.Fail($"the output '{outName}' cannot also be a donor — name a NEW plugin file.");
 
@@ -6723,7 +6744,7 @@ public sealed class LoadOrderService : IDisposable
                 plan.Donors, build.Conflicts, id.ExternalPlugins, id.ExternalOverriders,
                 id.PluginsScanned, id.UnscannableRecords, id.UnscannableSamples, build.Bytes, note,
                 assetRename, voiceRename, seqRegen, build.LightDonors, build.HeaderMetaDonors, build.MasterDonors,
-                id.UnscannablePlugins, localizedDonors, id.MasterDeclarers, build.LightCarried);
+                id.UnscannablePlugins, localizedDonors, id.MasterDeclarers, build.LightCarried, build.OriginatingRecords);
         }
     }
 
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -6723,7 +6723,7 @@ public sealed class LoadOrderService : IDisposable
                 plan.Donors, build.Conflicts, id.ExternalPlugins, id.ExternalOverriders,
                 id.PluginsScanned, id.UnscannableRecords, id.UnscannableSamples, build.Bytes, note,
                 assetRename, voiceRename, seqRegen, build.LightDonors, build.HeaderMetaDonors, build.MasterDonors,
-                id.UnscannablePlugins, localizedDonors, id.MasterDeclarers);
+                id.UnscannablePlugins, localizedDonors, id.MasterDeclarers, build.LightCarried);
         }
     }
 
--- a/src/housecarl-mcp/LoadOrderService.cs
+++ b/src/housecarl-mcp/LoadOrderService.cs
@@ -1826,6 +1826,18 @@ public sealed class LoadOrderService : IDisposable
             view.Epoch);
     }
 
+    /// <summary>The LOCALIZED header flag of ONE plugin, for housecarl_load_order_status' lookup= (#376): a localized
+    /// plugin's text lives in .STRINGS files rather than in the plugin, which is what the in-place write lanes refuse
+    /// on, so a caller can see that refusal coming instead of meeting it mid-job. Null when the name is not a plugin
+    /// this order resolved to a file; otherwise the three-way read, Unreadable included — never a bool.</summary>
+    public LocalizedFlagRead? PluginLocalizedFlag(string pluginName)
+    {
+        LoadOrderResolver.IndexView view;
+        lock (_gate) { view = Resolver.Capture(); }
+        var path = view.PluginPath(pluginName);
+        return path is null ? null : WriteEngine.PluginIsLocalized(path);
+    }
+
     /// <summary>Read MO2's OWN local Nexus update cache — the modid / version / newestVersion / ignoredVersion /
     /// lastNexusUpdate fields in every managed mod's meta.ini — with NO network (MO2 already paid the API cost). The
     /// cheap local pre-filter for update triage: it names which mods MO2 already learned a newer version for, plus the
--- a/src/housecarl-mcp/WriteTools.cs
+++ b/src/housecarl-mcp/WriteTools.cs
@@ -132,8 +132,8 @@ public static class WriteTools
          "A donor's HEADER mostly does not come along: master (ESM) status and Author/Description are always dropped, and the " +
          "report names each one it actually dropped. Light (ESL) status IS carried when every donor was light and every merged " +
          "object id fits the light window 0x800–0xFFF; otherwise it is dropped and the report says which of the two reasons it " +
-         "was. Want a dropped light flag back? Run " + ToolNames.CompactPlugin + " on the merged " +
-         "plugin afterward (the tools compose) — but it renumbers object ids from 0x800 upward, so ids the merge kept move.")]
+         "was, and whether " + ToolNames.CompactPlugin + " can still make it light — it can, by renumbering every id into that " +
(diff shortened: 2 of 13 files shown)


src/housecarl-mcp/WriteTools.cs
```diff
@@ -788,16 +790,34 @@ internal static string RenderMerge(WritePatchBuilder.MergeOutcome o)
...
             sb.Append("NOTE — ").Append(string.Join(", ", light.Take(10)));
             if (light.Count > 10) sb.Append(" (+").Append(light.Count - 10).Append(" more)");
             sb.Append(" carried the LIGHT (ESL) status; ")
               .Append(o.OutputName).Append(" does NOT — it is written as a full plugin and takes a full load-order slot. ")
-              .Append("To make it light again run " + ToolNames.CompactPlugin + " on '").Append(o.OutputName)
-              .Append("' (its esl defaults true) — but that renumbers object ids from 0x800 upward, so the ids listed as ")
-              .Append("kept above will move.\n");
+              // Why the flag was not carried, never a bare drop — and the two reasons take different remedies, so the
+              // remedy is written per reason rather than once for both.
+              .Append(light.Count < o.Donors.Count
```
Avick3110, line 820: The mixed-donor branch recommends a remedy that can fail. `MergeBuild` computes two independent conditions (`lightDonors.Count == donors.Count` and `lightIdsFit`) but collapses them into the single `LightCarried` bool, so the render can only tell "all donors light" from "not all donors light" — it cannot see the id-window result once a full donor is in the set. When both conditions fail, this branch fires and tells the caller to run `compact_plugin` on the output, but compact runs `BuildSequentialRemap` against `EslCeiling`, so it refuses outright above 2048 originating records.

Concrete: merge `Small.esl` (light, 10 records) with `Big.esp` (full, 3000 originating records). The report prints "Not every donor was light (1 of 2 were) … To make it light run housecarl_compact_plugin on 'Merged.esp' (its esl defaults true)". Running that gets `cannot merge 3010 originating records into the window 0x800–0xFFF` — the exact case the sibling branch below is careful to say compact cannot rescue.

Carrying `lightIdsFit` out of `MergeBuild` as its own field (rather than only folded into `LightCarried`) lets this branch pick the honest remedy for all four combinations instead of two.

Avick3110:
Blind review pass, additional findings. One finding from the same pass is already inline at WriteTools.cs:811 (the two carry conditions collapsed into one bool, so a full donor **plus** overflowing ids is sent to `compact_plugin`, which then refuses).

### 1. `src/housecarl-mcp/WriteTools.cs:816` — medium. The overflow sentence states a cause that is not established.

Overflow of the record count is not the only way the window check fails. `BuildMergeRemap` is called with the window `0x800`–`0xFFFFFF` (`LoadOrderService.cs:6603`), and it keeps an id whenever it is in that window and unclaimed (`RemapEngine.cs:655`) — so an originating id **above** `0xFFF` with nothing to collide with is carried through unchanged, however few records there are.

The light measurement deliberately counts a donor as light by extension as well as by flag (`WritePatchBuilder.cs:2582`), so: one donor, `X.esl`, header bit unset, one record at `0x5000`. Every donor is light, `lightIdsFit` is false, and the report says

> the donors together define more originating records than one light plugin holds … Merge fewer donors if a light output is what you need.

Both halves are false for a one-donor, one-record merge. That is a silently wrong answer in a sentence whose whole job is to say why. Either key the wording on whether `seen.Count` actually exceeded the window capacity, or say the neutral true thing (some merged id lands outside `0x800–0xFFF`) and drop the "merge fewer donors" advice when there is one donor.

### 2. `src/housecarl-mcp-tests/MergeLightCarryTests.cs:78` — medium. The overflow arm tests the render against a hand-written outcome, not the code.

The test merges the two in-window donors (which fit), then at line 83 does `o with { LightCarried = false }` and asserts on `RenderMerge`. Two consequences:

- The engine's ids-fit check (`WritePatchBuilder.cs:2603-2605`) has no coverage at all. Invert the comparison, or delete the loop and write `bool lightCarried = lightDonors.Count == donorsByLoadOrder.Count;`, and every test in this file plus every HEADER arm of `MergeServiceGuardProbe` still passes. The condition the ruling turns on is the one condition nothing exercises.
- The render arm it does reach picks the overflow branch from `light.Count == o.Donors.Count`, not from any ids-fit signal — so the assertion restates the fixture rather than showing the feature works. The comment "the same outcome, re-stated as the overflow shape" is accurate, and that is the problem: the engine cannot produce that outcome for these donors.

The fixture already writes object ids by hand (`Write(..., 0x801, ...)`). A third light donor at, say, `0x5000` makes this a real end-to-end arm for a couple of lines, and would also have surfaced finding 1.

### 3. `src/housecarl-mcp-tests/StatusLocalizedLookupTests.cs` — low. The three-valued read's third value is untested.

The PR body sells the unknown answer as the point ("an unreadable header is reported as unknown, never as 'not localized'"), but nothing asserts it. Localized and NotLocalized are covered end-to-end and the non-plugin case is covered via `null`; `LocalizedFlagRead.Unreadable` reaches no assertion, so the "UNKNOWN — houseCARL could not read this plugin's header" branch in `StatusWire.Render` is dead as far as the suite knows. One direct `StatusWire.Render(..., localized: LocalizedFlagRead.Unreadable, ...)` call covers it.

### 4. `src/housecarl-mcp/LoadOrderService.cs:1833` — low. An inactive plugin gets no localized line and no reason.

`IndexView.PluginPath` resolves only plugins in the active order (`LoadOrderResolver.cs:835`), so `PluginLocalizedFlag` returns `null` for an inactive plugin exactly as it does for a mod folder. The lookup block above prints an INACTIVE plugin verdict, and then the localized line is simply absent — the caller gets a plugin answer with no localized answer and nothing saying why one is missing. Reading inactive plugins is a surface houseCARL advertises, and the enum already has the value for "no answer established". Either resolve the path for inactive plugins too, or render `UNKNOWN` with the reason for a plugin the order did not resolve to a file.

### 5. `src/housecarl-mcp/WriteTools.cs:135` and `src/housecarl-mcp/LoadOrderService.cs:6526` — low. Two compact remedies the PR makes over-broad.

- The tool description now introduces the overflow reason and then closes with "Want a dropped light flag back? Run `compact_plugin` on the merged plugin afterward" — unqualified, though the text two sentences earlier names the case compact cannot fix. The rendered report is careful here; the description is not.
- The `.esl`-output refusal still ends "Merge to a '.esp' instead, then run `compact_plugin` on it to make it light". With the carry in place, an all-light donor set merged to `.esp` is already written light, so the advised compact is a pure renumber for no gain, and it contradicts the same call's new "is written LIGHT too" note.

### Checked and clean

The ids-fit scope is right — `plan.Dict` is built from `TryReadOriginatingKeys` per donor, so overrides are correctly excluded and no record with the output ModKey escapes the check, which is what keeps the light serialize from throwing. `NextFormID` landing at `0x1000` for a full light plugin is an already-accepted case. `IsSmallMaster` being set before the dangling-ref refusal is harmless (nothing is written on refusal). The asset change is clean: both sites in `AssetTools` now go through `AssetSourceSelection.Describe`, no bare `Name (kind)` remains, and the double-quote delimiter is the right choice for a name that may hold parentheses. No hand-written per-record-type coverage anywhere in the diff.

Avick3110:
Review fold — all six findings folded, none refuted. One commit: `4997147`.

**Inline, WriteTools.cs:811 — the mixed-donor branch recommends a remedy that can fail. Folded**, by a different
mechanism than suggested. The reason and the remedy turn out to be different questions: the reason is *which*
condition failed (donor mix, or a merged id outside the window), and the remedy is only ever "can compact fit this
plugin into the window", which is a question about the record count and nothing else — compact renumbers every id, so
it answers both reasons until the count itself overflows. So `MergeBuildResult`/`MergeOutcome` now carry
`OriginatingRecords` (the remap dict's size, which is exactly the count a later compact faces), the reason is written
per failed condition and the remedy per count, and all four combinations get a true sentence. `lightIdsFit` itself is
not carried out: with `LightCarried` and the donor counts already in the outcome, "all donors light and the flag not
carried" *is* the ids-fit answer, and a second field for it would be a second copy of the same fact.

**1. The overflow sentence states a cause that is not established. Folded.** The all-light drop now reads "not every
merged object id landed inside the light window (0x800–0xFFF) — a merge renumbers only what it must, so an id already
above the ceiling is kept where it is", which is the neutral true thing. The record-count claim and the "merge fewer
donors" advice now appear only on the arm where the count really did overflow, whichever reason dropped the flag. The
comment on the ids-fit check in `WritePatchBuilder` says the same thing: the fit check does not distinguish the two
ways an id lands outside the window, and the report reads the count to tell which.

**2. MergeLightCarryTests fabricates the overflow outcome. Folded**, with a real fixture, not a seam. Two donors added:
`HcLcLightC.esl`, light by extension with its one record at `0x5000` (written under the `.esp` ModKey and renamed,
because Mutagen will not serialize an id above the ceiling under a light ModKey; a plugin file carries no copy of its
own name); and `HcLcCrowd.esp`, a full donor with 2049 originating records. Two new arms replace the `with`:
`AllDonorsLightWithAnIdAboveTheCeilingDropsTheFlagAndStillOffersCompact` runs the engine's ids-fit loop for real (all
donors light, `LightCarried` false, compact still the right remedy at two records), and
`TooManyRecordsForTheWindowSaysCompactCannotRescueIt` is the inline finding's concrete case end to end. Inverting the
ids-fit comparison now fails the first of those.

**3. The Unreadable arm reaches no assertion. Folded**, end to end rather than by a direct `Render` call. The status
fixture gained an unticked file that is not a plugin at all, so `PluginIsLocalized` actually fails on it and
`AnUnreadableHeaderIsReportedAsUnknown` asserts the UNKNOWN line through `StatusTools.LoadOrderStatus`.

**4. An inactive plugin gets no localized line and no reason. Folded** the first way you offered: the path is resolved
for inactive plugins too, through the same `LocatePluginFileOnDisk` contract every other lane uses. `null` (no line at
all) now means only "this name is not a plugin" — a mod folder, a typo. A name the profile lists as a plugin that no
single file backs renders UNKNOWN rather than nothing. `AnInactivePluginStillGetsItsLocalizedAnswer` covers it.

**5. Two over-broad compact remedies. Folded.** The tool description now says compact can make a dropped flag light by
renumbering into the window "unless the merged plugin defines more than 2048 records, which no light plugin holds".
The `.esl`-output refusal now says merging to `.esp` writes the output light already when every donor was light and
every id landed in the window, and points at compact otherwise.

Build, `dotnet test --filter "tier!=bridge"` (1174 passed) and `ci-all` (128/128) are green.

Avick3110:
Blind review folded (6 of 6 real, 0 refuted). Queued for Aaron's review. Stacked follow-on: #559 (#542).

src/housecarl-mcp/WriteTools.cs
```diff
@@ -788,16 +790,41 @@ internal static string RenderMerge(WritePatchBuilder.MergeOutcome o)
...
+                            HousecarlCore.FormIdRange.EslWindowCeiling.ToString("X3");
             sb.Append(" carried the LIGHT (ESL) status; ")
               .Append(o.OutputName).Append(" does NOT — it is written as a full plugin and takes a full load-order slot. ")
-              .Append("To make it light again run " + ToolNames.CompactPlugin + " on '").Append(o.OutputName)
-              .Append("' (its esl defaults true) — but that renumbers object ids from 0x800 upward, so the ids listed as ")
-              .Append("kept above will move.\n");
+              .Append(light.Count < o.Donors.Count
+                  ? "Not every donor was light (" + light.Count + " of " + o.Donors.Count + " were), and merged content " +
+                    "that was never light-legal as a whole cannot be carried as light. "
+                  : "Every donor was light, but not every merged object id landed inside the light window (" + window +
```
Avick3110: **Correctness (medium).** When every donor is light but the donors together hold more than 2048 records, this reason clause is false. `BuildMergeRemap` runs with ceiling `0xFFFFFF`, so with two light ESPFEs of 1500 records each — every donor id inside `0x800–0xFFF` — pass 2 allocates the 1449 collisions from `0x1000` upward because 3000 > 2048. `light.Count == o.Donors.Count`, so the report prints "an id already above the ceiling is kept where it is", blaming a donor id that does not exist, and then the remedy clause immediately blames the record count instead — the two sentences contradict each other.

`MergeBuild`'s own comment says the record count is what tells the two causes apart, but only the remedy clause reads `OriginatingRecords`; the reason clause does not. Splitting the reason on the same count would make all four combinations true.

Neither new test arm covers this: `AllDonorsLightWithAnIdAboveTheCeiling…` uses a mis-flagged donor (the clause is true there), and `TooManyRecordsForTheWindow…` includes a full donor, so it takes the other reason branch.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1826,6 +1826,38 @@ public LoadOrderStatusData StatusData()
...
+        // Only a name the profile lists as a plugin: a mod folder and a typo both have no header, and inventing an
+        // UNKNOWN line for them would claim there is a plugin here whose flag simply could not be read.
+        bool isPlugin = comp.OrderedPluginNames.Any(n => n.Equals(pluginName, StringComparison.OrdinalIgnoreCase))
+                        || comp.InactivePluginNames.Any(n => n.Equals(pluginName, StringComparison.OrdinalIgnoreCase))
+                        || comp.ImplicitPluginNames.Any(n => n.Equals(pluginName, StringComparison.OrdinalIgnoreCase));
+        if (!isPlugin) return null;
+        var loc = LocatePluginFileOnDisk(comp, modsDir, dataDir, overwriteDir, pluginName, null, offerModParam: false);
+        // Listed as a plugin but no single file behind the name (missing, or several folders provide it): the flag is
+        // not established, which is exactly what the third value says.
+        return loc.Path is null ? LocalizedFlagRead.Unreadable : WriteEngine.PluginIsLocalized(loc.Path);
```
Avick3110: **Correctness (low).** `hits.Count > 1` also comes back with `Path` null, so a plugin name several mod folders provide collapses into `Unreadable` — which renders as "UNKNOWN — houseCARL could not read this plugin's header, so neither answer is established."

An unticked `Foo.esp` shipped by both a mod folder and an older backup folder hits this: every copy is perfectly readable, and MO2 priority already decides which one serves (`JudgeServed` computes exactly that from the same hits). The caller is told a header read failed when none was attempted. The missing-file case shares the value but at least matches the sentence.

Either the ambiguous arm could read the served copy's flag, or the UNKNOWN sentence needs a second spelling that says several folders provide this name rather than claiming a read failure.

Avick3110:
Aaron's review fold — both findings folded, none refuted. Two commits: `ff0772b`, `f1cc96a`.

**`WriteTools.cs:820` — the all-light reason clause blames an id above the ceiling when the record count is what
failed. Folded.** Real, and reachable exactly as described: `BuildMergeRemap` runs to `0xFFFFFF`, so two in-window
light donors totalling more than 2048 records get their collisions renumbered past `0xFFF` with no donor id ever
above the ceiling. The ids-fit condition fails two ways and only the count tells them apart, so the reason clause now
reads the count as the remedy clause already did: all donors light and the count fits keeps the "an id already above
the ceiling is kept where it is" wording, and all donors light with the count over the window says the donors
together define more records than the window holds. The two sentences no longer contradict each other, and the four
combinations each get a true one. New arm
`AllDonorsLightButTooManyRecordsBlamesTheCountNotAHighId` runs it end to end off a new light donor that fills the
window exactly (`HcLcLightFull.esl`, 2048 records at `0x800–0xFFF`, so every donor is light and every donor id is in
window) merged with `HcLcLightA.esp`; it asserts the count sentence and asserts the high-id sentence is absent.

**`LoadOrderService.cs:1858` — a name several mod folders provide reports UNKNOWN with a read-failure sentence.
Folded** the first way you offered. The ambiguous arm now reads the copy MO2 priority serves — `Ambiguous`'s first
enabled hit, which is the same copy `JudgeServed` calls "serves", off the same hits the locate already returned — so
no read failure is claimed where no read was attempted. `Unreadable` is left for the cases that really are
unestablished: no file behind the name, no copy served, or a header that would not parse. New arm
`APluginSeveralModFoldersProvideAnswersFromTheServedCopy` puts the same unticked filename in two enabled mod folders
with the LOCALIZED bit set on the higher-priority copy only, so an answer read from the loser would say the opposite.

Build, `dotnet test --filter "tier!=bridge"` (1176 passed, +2) and `ci-all` (128/128) are green.

Issue #258, opened 2026-07-21
[GAP] read_record/depth: polymorphic Conditions[].Data arm params never expand via list expansion (only via a direct path)


Avick3110:
> *Filed from an external live-session gap report (dialogue-authoring pass); triaged, root-caused against source, and posted by AI.*

### Goal

Dump a dialogue INFO's full condition stack (function + params) to replay it verbatim on new records (recruit-opener donor reads, ~40 rows).

### What fell short

`read_record fields=["Conditions"] depth=3` (text AND `format=json`) expands each condition row's scalars (op, value, flags) and shows `Data = [GetFactionRankConditionData]`, but NEVER the arm's parameter fields (Faction / Global / Object / Quest / Stage). The params appear ONLY when the arm is addressed directly — `Conditions[2].Data` at `depth=2`, or `Conditions[2].Data.Faction`.

Tool calls tried:

    housecarl_read_record formid=<INFO> fields=["Conditions"] depth=3 format=json        # → Data params absent
    housecarl_read_record formid=<INFO> fields=["Conditions[2].Data","Conditions[3].Data",...]   # → works, one path per row

Evidence: `depth=3` json returned `{"path":"Conditions[2].Data","note":"[GetFactionRankConditionData] Reference=Null"}` with no child fields, while `fields=["Conditions[2].Data"]` at `depth=2` printed `Faction=…` + `RunOnType` etc.

### Workaround used

Hand-enumerated per-row `Data` paths (25 paths in one call) — fine once, awkward for every donor read on a 30-row stack. Recurs on every condition-donor read; dialogue work does this constantly.

### Ideal shape

`depth>=3` on a `Conditions` list opens the polymorphic `Data` one bounded level, surfacing the arm's parameter fields — the params ARE shown when `Data` is addressed directly, so the renderer can already do it.

- date: 2026-07-20  |  houseCARL version: 1.9.0+444df77a

---

### Triage / root cause (verified against source)

Valid, reproducible, non-duplicate. Next member of the **#198 / #252** ReadEngine depth-expansion family — the class where a nested structure renders as a bare `[Type]` summary one level shallower than the `depth=` contract leads you to expect.

The mechanism is precise. `ReadEngine.Expand` (`src/housecarl-core/ReadEngine.cs`) has a depth-floor exception that surfaces a substruct's value ONE bounded level deeper even at `depth<=1` — but it is scoped to VMAD script properties only (`ReadEngine.cs:368-373`):

[...]

PR #550, opened 2026-09-05
Add a rows projection form for per-element list views

### LINKED ISSUE #303: [GAP] no compact per-row condition view — auditing a 40-row CTDA list needs index-probing across ~8 calls
> *Filed from an external live-session gap report (dialogue condition-surgery pass); triaged, root-caused against source, and posted by AI.*

### Goal

Before editing 4 dialogue INFOs with ~40-row condition stacks, locate the essential-gate OR group inside each: which row is `IsEssential`, which rows carry the `Or` flag, where each group terminates.

### What fell short

There is no compact per-row view of a condition list. Every available read is either opaque or unreadably verbose:

- `read_record depth=2` renders each row as a bare `[ConditionFloat]` — no function, no operator, no value, no flags. Useless for locating a row.
- `depth=3+` (post-#258) now opens the arm and dumps ~25 lines **per row**, most of them absent parameters. At 40 rows that blows `max_chars` and truncates.
- `validate_dialogue` decodes conditions, but only reports the **malformed** ones — there is no row table.

What actually worked was index-probing: hand-built `fields` arrays of 15-40 paths (`Conditions[i].Data.Function`, `Conditions[i].Flags`, …), ~8 calls across 4 records, to find 4 row indexes.

Tool calls tried:

    housecarl_read_record ... depth=2                                        # opaque
    housecarl_read_record ... depth=3                                        # too verbose, truncates
    housecarl_read_record ... fields=["Conditions[N].Data.Function","Conditions[N].Flags",...]   # worked, tedious

### Workaround used

Index-probing as above, plus an external dialogue dump for the earlier records.

### Frequency

Every condition-surgery task. 30-40-row condition stacks are normal on gate INFOs.

### Ideal shape

A **one line per row** condition view — function, key parameter resolved, operator, value, `Or` flag, RunOn — as either a `read_record` display mode or a `validate_dialogue` `decode` option. The decode semantics already exist as data in the `dialogue-authoring` skill's `references/condition-functions.md`.

- date: 2026-07-31  |  houseCARL version: unknown

---

Valid, reproducible, non-duplicate. **Adjacent to #258 but the opposite end of it**, and worth reading as its follow-on: #258 was "depth=3 stops at the bare arm type"; the fix (`6015faa`, `ReadEngine.cs:363-380`) opened `Conditions[].Data` one bounded level at the depth floor. That closed the *missing-params* half — and it is precisely why the reporter now hits the *too-many-params* half. The renderer emits every modeled parameter field per arm whether set or absent, so the row count that fits under `max_chars` fell.

So both reported failure modes are real and structural, not user error:

- `depth=2` genuinely has no function on a row — the function lives on the polymorphic `Data` arm, which at that depth renders as the summary token only.
- `depth=3` genuinely costs ~25 lines/row for the same reason the arm now expands at all.
- `validate_dialogue` genuinely has no row table: `DialogueValidate` treats CTDA semantics as explicitly out of scope (`DialogueValidate.cs:44-47` — "the CTDA conditions that gate when a line fires are semantic and only the game evaluates them"), so it reports malformedness and nothing else. That boundary is deliberate and should stay; a **decode** view is not an evaluation and does not cross it.

There is a real asymmetry here worth naming: houseCARL can *compose* a condition row from parts, but cannot *render* one back in the same compact shape. The write surface has a canonical row form; the read surface has only the reflection dump.

Fix direction (either home works, the first is narrower):

1. `validate_dialogue` gains an opt-in `decode` that emits a row table per INFO — `idx | Or | function | key param (resolved to EditorID) | op | value | RunOn`. It already walks every INFO and already resolves links.
2. `read_record` gains a compact display mode for condition lists specifically — same one-line-per-row shape, at the same place `ReadEngine` special-cases the `ConditionData` arm family today.

Either way the suppression rule is the payload: **absent parameters are omitted**, which is what turns 25 lines into 1.

Avick3110:
Reading a struct list off a record costs one line per modeled sub-field per element, so a 40-row condition stack runs to about a thousand lines and truncates before it is readable — most of those lines are parameters the arm does not carry. The way through was to hand-build a `fields` array of 15-40 indexed paths per record and probe, about eight calls to find four row indexes. This adds `project.form='rows'`: a list field folded to one line per element. Each line is the element's type followed by every sub-field that is actually there — the function, the operator, the value, the `Or` flag, the run-on — with absent optionals and null links omitted. That is the whole rule, and it is what turns 24 lines per condition into one.

It lands as a value on the existing project axis, not as a tool and not as a condition special case: the fold is over emitted paths, so it works on `Effects`, `Perks`, or any other list the same way, and conditions are only the case that motivated it. The form reads the caller's own `project.fields` like the `fields` form and takes the same `project.depth` (default 4 here, since a shallower read leaves every element as a bare type; `depth=1` is refused rather than answering with no rows at all), `resolve_names` annotates a link inside the row, and the form composes with the list lane, the scan lane and the off-order lane. `format='dense'` refuses it by name — a folded row has no fixed column set. An unreadable sub-field keeps its note, because a read fault is not an absence, and the list's own header keeps its true count.

No engine change was needed: #258's arm-opening at the depth floor already surfaces a condition arm's parameters, and this is the other half of that — #258 was "the params never appear", this is "they all appear, one line each".

Tests: 14 new, against a three-row condition stack added to the records world (one row `Or`-flagged, one whose arm carries no parameters) and the spell effect beside it for the general case. Build clean, 1174 tests pass, 128/128 probes pass.

Closes #303


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -240,14 +249,20 @@ public static string Records(
...
         }
-        if (project is { resolve_names: true } && form is not ("fields" or "everything"))
-            return Wire.Refuse(json, $"error: project.resolve_names annotates field values and belongs to the 'fields'/'everything' forms (got form='{form}').");
-        int depth = project?.depth is { } d && d > 0 ? d : 1;
-        var projFields = form is "fields" or "delta" or "tree" ? project?.fields : null;
+        if (project is { resolve_names: true } && form is not ("fields" or "rows" or "everything"))
+            return Wire.Refuse(json, $"error: project.resolve_names annotates field values and belongs to the 'fields'/'rows'/'everything' forms (got form='{form}').");
+        // The rows form's default is its own: at depth 1 every element renders as a bare arm type, which is the
+        // gap the form exists to close. Its cost is per-row TEXT, not per-row lines — the fold is one line either way.
+        int depth = project?.depth is { } d && d > 0 ? d : (form == "rows" ? RowProjection.DefaultDepth : 1);
```
Avick3110, line 286: depth=4 by default runs this form straight into `ReadEngine.MaxExpandNodes` on exactly the lists it exists for. That budget is 2000 emitted lines per record across all targets, and the PR's own figure is ~24 lines per condition -- so a stack past roughly 80 elements folds only its first ~80 and the remainder becomes a single `(expansion truncated at 2000 lines ...)` note. A CELL's `Temporary`, a long `Effects` list, or a 100+ row condition stack hits this. The cut is declared rather than silent, but the remedy the note offers is "narrow with a field path or a lower depth", and on this form depth=1 is refused and depth=2 renders bare arm types -- depth=3 is the only usable step down, and neither the note nor the depth=1 refusal says so.

src/housecarl-mcp/RowProjection.cs
```diff
@@ -0,0 +1,107 @@
...
+        && (path[owner.Length] == '.' || path[owner.Length] == '[');
+
+    /// <summary>The element path a line belongs to — a requested root plus the first bracketed segment after it —
+    /// or null when the line is the list's own summary, a non-list field, or the truncation note.</summary>
+    static string? RowKey(string path, IReadOnlyList<string> roots)
+    {
+        foreach (var r in roots)
+        {
+            if (path.Length <= r.Length || !path.StartsWith(r, StringComparison.Ordinal)) continue;
+            if (path[r.Length] != '[') continue;
```
Avick3110: A root that is not a plain list path never folds, and nothing says so. `RowKey` matches only when the character right after the root is `[`, so `project={"form":"rows","fields":["Conditions[0]"]}` (the natural way to probe one row) or a non-list path such as `["Name"]` produces no key for any emitted line: every line falls through the `key is null` branch and the response is byte-for-byte the `fields` form -- no note, no refusal. A caller who asked for one line per element gets the ~24-line unfolded read and no explanation. An empty list is a legitimate zero-row answer, so the check cannot be "no rows were produced"; it has to be on the root itself -- refuse an already-indexed root, and say when a named root resolved to something that is not a list.

src/housecarl-mcp/RowProjection.cs
```diff
@@ -0,0 +1,107 @@
...
+                cells.Add(Cell(f, openRow, trimToType: kept.Any(g => IsUnder(g.Path, f.Path))));
+            outp.Add(new FieldValue(openRow, false, null, string.Join(Sep, cells), Present: true));
+            openRow = null; kept = null;
+        }
+
+        foreach (var f in fields)
+        {
+            var key = RowKey(f.Path, roots);
+            if (key is null) { Close(); outp.Add(f); continue; }
+            if (key != openRow) { Close(); openRow = key; kept = new List<FieldValue>(); }
```
Avick3110: Rows are keyed by contiguity, so two overlapping roots split one element into two rows that share a path. With `fields=["Effects","Effects[0].Conditions"]` the longer root wins for the nested lines (Apply sorts by length), so emission runs `Effects[0]`, ..., `Effects[0].Conditions[0]`, ... and then returns to `Effects[0].<sub-field>` for any of the element's own sub-fields emitted after the nested list -- sub-field order comes from `ReflectedFieldNames`, not from the caller. `key != openRow` then closes the first row and opens a second one under the same `Effects[0]` key, so the JSON `fields` array carries two entries with an identical `path`, and the second has no element summary leading it because `isElement` is false for all of its lines.

src/housecarl-mcp/RowProjection.cs
```diff
@@ -0,0 +1,107 @@
...
+            bool isElement = f.Path.Length == key.Length;
+            // The element's own line always leads the row — it names the arm type, and dropping it could leave a
+            // row with nothing to render at all. Its sub-fields are kept only when something is there.
+            if (isElement || f.Present || !f.Readable) kept!.Add(f);
```
Avick3110: `!f.Readable` is unreachable on this lane, so the "a read fault is not an absence" guarantee in the class doc and the changelog is not actually delivered. `Readable: false` comes only from `LeafRead.Unreadable`, which lives on `ReadEngine`'s depth-1 leaf branch; the rows form always reads at depth >= 2, where `Expand` swallows a nested property-get fault outright (`catch { continue; }`) and the only `(unreadable: ...)` line it can emit sits on the top-level target path -- the list header, never inside a row. `AnUnreadableSubFieldIsNotAnAbsentOne` passes because it hand-builds a `FieldValue` the engine cannot produce on this path.

Separately, the `f.Present` filter drops a declared-but-null FormLink, which is the exact signal the None-link diagnostics key on: `form='rows'` over `VirtualMachineAdapter.Scripts` renders a script property the author left None identically to a property that does not exist, so "which properties are unfilled" cannot be answered from a row and a reader can conclude the wrong thing.

Avick3110:
Three more from a second read, on top of the inline comments.

**medium — `src/housecarl-mcp/RowProjection.cs:56`: the folded row puts its data in `Note`, so the json render calls it a note, not a value.**
`new FieldValue(openRow, false, null, string.Join(Sep, cells), Present: true)` is `HasValue: false`, `Token: null`, `Note: <the row>`, `Count: null`. `JsonWire.cs:260` then writes `{"path":"Conditions[0]","note":"[ConditionFloat] | Data.Function=…"}`. `FieldValue`'s own contract (`ReadEngine.cs:29`) says the value/note split is carried structurally precisely so "a consumer deciding presence from the note text would be parsing prose" — the rows form hands a json consumer exactly that. Two consequences beyond the shape: the per-cell `Display` and `Link` annotations are flattened into the string, so `resolve_names` under `format='json'` no longer yields the structured `link` object it yields on the `fields` form, and `Count` is dropped where the ordinary container line carries it. `TheJsonDocumentCarriesOneEntryPerElement` asserts the paths only, so none of this is pinned by a test.

**low — `src/housecarl-mcp-tests/RecordsRowsFormTests.cs:73`: the generality claim rests on a one-element list.**
`TheFoldIsGeneralNotConditionSpecific` and `ResolveNamesAnnotatesInsideARow` both read `SpellA.Effects`, which has a single effect, so the non-Conditions case never exercises the multi-element fold, the row-close/reopen boundary, or an absent optional on a non-condition element — the three things the fold actually does. A second effect on `SpellA` (and an assertion on `Effects[1]`) would make the claim carry. Also unexercised: the off-order lane, `to_file=`, and a `Conditions` read on a record with a nested list inside the element.

**low — design, `src/housecarl-mcp/RecordsTools.cs:34`: `rows` is a render shape sitting on the PROJECT axis.**
The form reads exactly what `form='fields'` reads (same `project.fields`, same `depth`, same engine call) and differs only in how the emitted lines are printed — the fold is over emitted paths, by the file's own description. That is TRANSPORT's job (`format`), and the tell is that `rows` and `format` now have to police each other: `format='dense'` needs a new refusal by name, and `format='json'` gets a shape it cannot express structurally (above). Worth stating why it is a PROJECT value rather than a `format` value, since the answer decides where the next compact render lands. The counter-case — that the fold drops absent lines and so changes *what* is projected, not only how — is real; it just is not the argument the PR body makes.

Avick3110:
Review fold. Build clean, 1186 tests pass (12 new), 128/128 probes.

**Folded**

1. **The silent fields-form fallback (`RowProjection.cs:86`).** Both halves, but the indexed root is fixed the other way round: `fields=["Conditions[0]"]` now *folds that one element* rather than being refused — it is the natural probe, it is the only way to narrow inside a huge list, and making it work removes the silent fallback with more capability rather than less. A root that resolved to something that is not a list now fails that record loud and names it ("`'EditorID'` on MagicEffect holds a single value — read it with `project.form='fields'`"). Per record, not per call: only the read knows what the path resolved to on that record's type. An absent or no-such-field root still passes through — that is the read's answer, not a misuse of the form.

2. **`depth=4` into `MaxExpandNodes` (`RecordsTools.cs:256`).** The remedy is fixed: on this form the truncation note now says the fold runs *after* the read, so the elements past the cut are missing, and points at an indexed element or `depth=3` (naming that `depth=2` leaves bare types). The deeper problem is real and **not** fixed: the budget is spent on lines the fold then discards, so a big stack still folds only its first ~80 rows. Folding during expansion is an engine change (`ReadEngine.Expand` would need the fold's rule) and did not belong in this PR — flagging it rather than half-doing it.

3. **Data in `Note` under json.** Fixed structurally. `FieldValue` gains `Cells` — the leaves a fold put on one line — and the json render emits a row as `{"path": "Effects[0]", "cells": [...]}`, each cell with its own `value`/`note`/`display`/`link`. `resolve_names` yields the same link object it yields on the `fields` form, counts ride on the cells that carry them, and no row writes a `note`. Pinned by a test that asserts the absence of `note`, the `value`, and the `link.editorid`.

4. **Duplicate json paths from overlapping roots.** Rows are keyed by path, not by contiguity: a re-opened element rejoins its own row. Unit-tested at `Fold` with the exact interleaving you traced (the engine's own field order does not produce it on any fixture record, so pinning it at the render would have pinned nothing).

5. **The `!Readable` keep and the `Present` filter.** The null-link half is fixed: only an ABSENT optional is dropped now, so a declared-but-null link is a cell and `VirtualMachineAdapter.Scripts` answers "which properties are unfilled". It costs one cell per condition (`Data.Reference=(null link)`), which is characters, not lines — the compression this form sells is the fold, not the omission. On `!Readable`: you are right that the lane cannot produce one, because `ReadEngine.Expand` swallows a nested property-get fault (`catch { continue; }`) rather than naming it. I kept the guard and removed the claim instead — the class doc and the changelog now point at that engine bound, and the test says out loud that it pins the fold and not the lane. The swallow itself is a cornerstone problem in core (a read fault rendered as nothing there) and wants its own issue.

6. **The one-element generality claim.** `SpellA` gains a second effect: a declared-but-null `BaseEffect`, and its own `Conditions` list, so the non-condition case now exercises the multi-element fold, the row boundary, a null link and a nested list inside an element. Also now tested: the off-order arm (the old file carries a copy of the spell), a walked reached set, and `to_file`.

8. **The walk refusal** now lists rows among the forms that consume a reached set.

**Left for you — (7), rows on PROJECT vs TRANSPORT**

My honest read: it belongs on PROJECT, and the tell you named is not the one that decides it.

- §2.1 defines TRANSPORT as uniform across all substrates, full stop. `rows` means nothing on assets, NIF, Papyrus or Nexus — a value that only exists for a record's list field fails that test on its face, while `format`, `limit`, `max_chars` and `to_file` pass it.
- The fold changes what is projected, not only how it prints: it drops absent optionals, refuses a non-list root, and folds an indexed one. Two calls with identical `fields`/`depth` now return different sets of facts.
- The `format='dense'` refusal is not evidence of an axis collision: dense already refuses `everything` and `aggregate` by name, both uncontested PROJECT forms. Any form without a fixed column set refuses the columnar transport — that rule pre-dates this PR.
- The json half of your tell was real and is now gone: a row renders in both transports, structurally, which is what a PROJECT value must do and a `format` value could not (one format cannot compose with another).

You are right that the PR body argued the wrong thing — it argued generality, which is not the axis argument. The axis argument is the drop rule, and it is stronger after this fold than it was when you read it.

If it does move, my view is the alternative is not a `format` value but a sub-parameter of the fields form (`project={"form":"fields","fold":"rows"}`), which states plainly "same read, one folding rule" — at the cost of a knob whose legality depends on the named field being a list. I have not restructured anything; your call.

Avick3110:
Blind review folded (6 folded, 2 partly refuted with reasons, the PROJECT-vs-TRANSPORT question left to Aaron). Queued for Aaron's review.

src/housecarl-mcp/RowProjection.cs
```diff
@@ -0,0 +1,165 @@
...
+            if (key is null) { outp.Add(Passthrough(f)); continue; }
+            if (!slotOf.TryGetValue(key, out int slot))
+            {
+                slot = outp.Count; slotOf[key] = slot; cells[key] = new List<FieldValue>();
+                outp.Add(f);                                                    // placeholder, rewritten below
+            }
+            // The element's own line always leads the row — it names the arm type, and dropping it could leave a
+            // row with nothing to render at all. A sub-field is dropped only when it is an ABSENT optional.
+            bool isElement = f.Path.Length == key.Length;
+            if (isElement || f.Present || !f.Readable || f.Note != ReadEngine.AbsentNote) cells[key].Add(f);
```
Avick3110, line 99: **Overlapping roots append every folded cell twice.**

`ReadEngine.ReadFields` walks each requested path independently and never deduplicates (`foreach (var p in targets) EmitWithDepth(...)`), so when `project.fields` names both a list and a list nested inside it, every line under the inner root is emitted **twice** — once inside the outer root's expansion, once from its own root. `Fold` keys rows by path, so both copies land in the same `cells[key]`.

Concretely, `project={"form":"rows","fields":["Effects","Effects[1].Conditions"]}` on `SpellA`:

- the `Effects[1].Conditions[0]` row renders `[ConditionFloat] | [ConditionFloat] | Data=… | …` — every sub-field repeated;
- the `Effects[1]` row carries `Conditions=[list: 1 item(s)]` twice.

`TheJsonDocumentNeverCarriesTwoEntriesForOnePath` drives exactly this input, but only asserts that the top-level paths are distinct, so it passes over the duplication. Deduplicating by cell path when appending would close it.

src/housecarl-mcp/RowProjection.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// offers a remedy written for the fields form; on this form it would send a caller to a depth that renders no
+    /// rows, so the remedy — and only the remedy — is restated.</summary>
+    static FieldValue Passthrough(FieldValue f)
+    {
+        if (f.Note is not { } n || !n.StartsWith("(expansion truncated", StringComparison.Ordinal)) return f;
+        int dash = n.IndexOf('—');
+        if (dash < 0) return f;
+        return f with { Note = n[..(dash + 1)] + " the fold runs AFTER the read, so the elements past the cut are " +
+                               "missing from the rows: name one element to fold it alone (e.g. \"Conditions[0]\"), " +
+                               "or lower depth to 3 (depth 2 leaves every element a bare type))" };
```
Avick3110: **The rewritten remedy hard-codes "lower depth to 3" regardless of the depth the call ran at.**

`project={"form":"rows","fields":["Items"],"depth":2}` on a leveled list with more than `MaxExpandNodes` entries truncates, and the note then tells the caller to "lower depth to 3" — an *increase* for a depth-2 call, which truncates sooner; at depth 3 it is a no-op. At those depths the only remedy that helps is the "name one element" clause beside it.

Conditioning the depth clause on the depth in scope (or dropping it when `depth <= 3`) keeps the sentence true at every depth the form accepts.

src/housecarl-mcp/RowProjection.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// <summary>Separator between a row's cells. Values carry spaces of their own, so the cells need a mark that
+    /// a value cannot be mistaken for.</summary>
+    const string Sep = " | ";
+
+    /// <summary>Fold every outcome's field list. A failed read has no body and passes through; a root that
+    /// resolved to something that is NOT a list fails that record loud, because the form's whole answer is rows
+    /// and there are none to give.</summary>
+    internal static IReadOnlyList<ReadOutcome> Apply(IReadOnlyList<ReadOutcome> outcomes, IReadOnlyList<string> roots)
+    {
+        var order = roots.OrderByDescending(r => r.Length).ToList();   // a root that prefixes another must not claim its rows
```
Avick3110, line 41: **A null entry in `project.fields` throws here, where the `fields` form degrades gracefully.**

`project={"form":"rows","fields":["Conditions",null]}` — System.Text.Json binds a null array element to a null string, and `project?.fields is { Length: > 0 }` only checks the array. `ReadEngine.ReadFields` absorbs the null per path inside its own try/catch (the `fields` form answers with an "(unreadable…)" note), but `roots.OrderByDescending(r => r.Length)` dereferences it and throws `NullReferenceException`.

`Guard.Tool` then returns "…failed unexpectedly — NullReferenceException … This is an internal houseCARL failure (**the arguments bound fine**), not bad input" — which asserts the opposite of what happened. Skipping or refusing null/blank roots keeps that wording honest.

Avick3110:
Second review fold. Build clean, 1190 tests pass (4 new), 128/128 probes.

All three findings were real; all three are folded.

**Folded**

1. **Overlapping roots appended every folded cell twice** (`RowProjection.cs`). Confirmed at the source you named: `ReadFields` runs `foreach (var p in targets) EmitWithDepth(...)` with no dedup, so every line under a nested root arrives once per root that covers it, and the path-keyed row took both copies. A row now keys its cells by path too and keeps the first at each. Two tests, both of which fail without the fix: the interleaving at `Fold` (`OverlappingRootsDoNotPrintACellTwice`), and the real read you traced — `fields=["Effects","Effects[1].Conditions"]` on `SpellA` — asserting the element's type token appears once in the rendered row (`ARowReadWithOverlappingRootsCarriesEachCellOnce`). `TheJsonDocumentNeverCarriesTwoEntriesForOnePath` stays as the distinct-paths assertion it was; it was never the one that could see this.

2. **The remedy hard-coded "lower depth to 3."** Right — at depth 3 it is a no-op and at depth 2 an increase, and the note had no depth in scope to know. `Apply`/`Fold`/`Passthrough` now carry the depth the call ran at, and the depth clause is emitted only above 3; below it the sentence ends at "name one element to fold it alone". Tested at `depth=2` on `BigList` (truncates, offers the element, says nothing about depth), with the default-depth test unchanged.

3. **A null entry in `project.fields` threw.** Confirmed — `roots.OrderByDescending(r => r.Length)` dereferences it and `Guard.Tool` then asserts the arguments bound fine, which is the opposite of what happened. Refused by index before any read instead of skipped, because the entry names a list to fold and an empty one names nothing: `project.fields[1] is empty — the 'rows' form folds the list each entry names, so every entry must be a field path`. Blank strings go the same way. Scoped to this form; the `fields` form's "(unreadable…)" answer for a null path is unchanged and is its own question.

Nothing refuted.

Issue #458, opened 2026-08-31
[GAP] where= cannot ask "does ANY element of this list match?" — the general lookup / referenced-by capability


Avick3110:
> *Filed from an external live-session gap report by **Jouten** (running 1.9.0), with their OK to be named; triaged, root-caused against source, and posted by AI. Machine paths scrubbed.*

### Goal

Prove a claim before shipping a dialogue fix: *"the condition function `IsGuard` appears exactly N times in this plugin, and here is every record carrying it."* This is the standard audit before changing a shared gate — find every site, not just the one that bit you. Same shape as "every INFO conditioned on quest X", "every MGEF with an `IsEssential` row", "every leveled list containing item Y".

### What fell short

`where=` predicates need a CONCRETE index. `Conditions[12].Data.Function = IsGuard` works; `Conditions.Data.Function = IsGuard` is correctly refused as not-a-field. But there is no way to express **"any element"**. Since a condition can sit at any index in a 43-long list, proving absence means sweeping every index one call at a time.

Tool calls tried:

    housecarl_cross_plugin_query type=DialogResponses plugins=[...] where=["Conditions.Data.Function = IsGuard"]
    -> refused, not a field (correct, but it is the query I actually wanted)

    ... then NINE separate calls, where=["Conditions[N].Data.Function = IsGuard"] for N = 10..18

    ... which still only covered indices 10-18, while the big gates carry up to 43 conditions -- so the
        audit was still incomplete and I nearly shipped an overclaim on the strength of it

    ... to close it: where=["Conditions[19] exists"] to find the 11 records with long lists, then one
        housecarl_batch_record_detail hand-enumerating 24 field paths
        (Conditions[19].Data.Function ... Conditions[42].Data.Function) across those 11 records

Verbatim refusal:

    predicate field 'Conditions.Data.Function' yielded no readable value on any of 1,095 scanned
    record(s) - it is NOT A FIELD on these records (a mistyped path, or a field that doesn't exist on
    this record type); check the field name against the record's schema. 0 matches on that basis is
    NOT a confirmed 'nothing matches'.

The reporter notes that last sentence is genuinely valuable and is the reason they did not treat the empty result as proof. The gap is that the query itself cannot be expressed.

### Workaround used

[...]

PR #551, opened 2026-09-05
Add the quantified where= step and the SELECT exclusion terms

### LINKED ISSUE #531: [GAP] SELECT has no exclusion terms: has is all-bits-only (no has_any/has_none) and references= cannot be negated
From a Discord report (2026-08-27, houseCARL 1.9). After 2.0. Vocabulary, so it lands before W6 if promoted, beside #458 in the E1 design page.

Two missing SELECT terms, one issue:
- `has <mask>` is all-bits-required (`FieldPredicate.cs`); there is no ANY or NONE form.
- `references=` is reverse, one step, positive only; there is no negated spelling.

The reporter's job ("ARMO on any slot 44-60, no armor slot, not already carrying keyword X") took 18 queries plus a script. Both are values missing from existing SELECT operators, the one-general-primitive case, not a job-shaped tool.

Not #458 (any-element wildcard), not #234 (resolved-link predicate), not #226 (closed, `formid not in`).

Avick3110:
A `where=` path step can now declare its multiplicity and its fold where the list binds, and `where=`/`references=` gain the exclusion spellings that go with it. This is step 1 of E1, the parser and the evaluation over what the current flat index already answers; containment (`*parent`, #459) and the reverse-reference index (#489) are the next two steps and are not touched here.

Four tokens land in the step: `Conditions[*any].Data.Function = IsGuard` matches when some element does, `[*all]` when every one does (vacuously true on an empty list), `[*none]` when none does — the "absence, proved" query, `Effects[*none].BaseEffect->editorid startswith REQ_` — and `[*count]` yields the number of elements for a numeric or membership comparison. The fold is the one `EvalLinkStep` already ran over link targets with the fan-out source swapped to the step's elements, so it composes with the `->` link step and with another quantified step for free; an absent list reads as empty, and so does one whose parent substruct is absent, so `VirtualMachineAdapter.Scripts[*count] = 0` finds the records that carry no script at all. The step's shape is judged on its DECLARED type, so a null non-list field is refused exactly as a carried one is, and a dict or a raw byte block is named rather than folded over. Where one element cannot be judged, the fold reports no verdict rather than claiming one — a universal over a list with an unread element is not proved.

The refusals: the bare `[*]` is the element set and is refused in a predicate naming the three fold tokens; an unknown quantifier word is refused naming all four; `[*count]` refuses anything after it and refuses an operator that is not a number comparison; a quantifier on a step that is not a list is refused from the schema wherever `types=` says which record type it lands on (naming the step's real cardinality), and where the scope cannot say — no `types=`, or a mixed one whose arms disagree — the scan's per-record accounting names what that step read instead. A fold token in `project.fields` is refused too: a boolean is not a row. The same token reaching the concrete-index walk now says which parameter each token belongs to rather than reporting an unbuilt capability as a malformed index, and the list-hop remedy offers the quantified spelling beside the bracketed one. The quantified step's cost — per-candidate work times list length, and one winner fetch per element under `->` — is named in the tool description beside the tokens, per SPEC §2's amendment.

For the exclusion half: `has` gains `has_any` (at least one operand bit set) and `has_none` (none set) — the same any/all/none fold over the bits of one leaf, three branches on an op enum that already evaluated `Has` bitwise — and a `!` before a `references=` entry inverts it, so `references=["!XXXXXX:A.esm"]` keeps only the records that do not reference that target. Plain and negated entries in one call compose by AND, and the negated form takes the same bounding `types=`/`plugins=`/`formids=` scope the plain form takes. A record with no live body is kept by the negated term — it references nothing at all, which is the strongest match the term has — rather than riding the skip the positive term needs.

**What is left of #458.** The PROJECT half is not here, so this PR references #458 rather than closing it. `fields` taking `[*]` and `[*count]`, and `dense`'s row-per-element reading, are named in SPEC §2.2's amendment but not specified far enough to build blind: neither the design page nor the SPEC says what `[*]` renders in `format=text`/`json`, what happens when two quantified paths in one `fields=` list have different element counts (a product, an alignment, or a refusal), or which identity columns the extra dense rows carry. Rather than guess a row shape that a skill would then be authored against, this PR ships the half that is specified — the `where=` vocabulary — and refuses the tokens in `project.fields` by name so nothing reads them silently. The remainder is #458's PROJECT half, and it is vocabulary, so it still needs to land before W6.

One thing the design page and the SPEC did not settle: the *spelling* of the negated `references=`. Both say the term "gains a negated spelling" without naming it. I took `!<formid>` on the §5.5 sigil rule — a FormID token can never begin with `!`, so the sigil is disjoint from the co-resident space by construction — and made mixed plain/negated entries compose by AND rather than refusing, since the AND is well defined and a refusal would leave "references A but not B" unaskable. If Aaron wants a different spelling it is a one-line change in `RecordsTools.SplitReferenceTargets` plus the tool description, and it is vocabulary, so it should be settled before W6. Its *meaning* is a second question, raised in the review and left for Aaron in a comment below.

Build, tests and probes green: 1202 tests (baseline 1160) and 128/128 probes. Every token and every refusal has a test.

Refs #458 (SELECT half; the PROJECT half stated above remains)
Closes #531

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -49,8 +49,9 @@ saying it sets an expectation their install may contradict. Say what is known, a
   A quantified step composes with the `->` link step and with another quantified step, at the cost the tool
   description declares (per-candidate work times list length, and one winner fetch per element under `->`). The bare
   `[*]` is the element SET and is refused in `where=`, naming the three fold tokens; a quantifier pointed at a field
-  that is not a list — a scalar, a dict, a raw byte block — is named in the scan's accounting with what that step read
-  instead, never counted as a silent non-match. An absent list reads as EMPTY, and so does one whose parent substruct
+  that is not a list — a scalar, a dict, a raw byte block — refuses the call by naming the step's real cardinality
+  wherever `types=` says which record type it lands on, and where the scope cannot say (no `types=`, or a mixed one)
+  the scan's accounting names what that step read instead; either way, never a silent non-match. An absent list reads as EMPTY, and so does one whose parent substruct
   is absent, so `VirtualMachineAdapter.Scripts[*count] = 0` finds the records that carry no script at all; where an
   element cannot be judged, the fold reports no verdict rather than claiming one. Reaching a list with a dotted
   segment still refuses, and its remedy now offers the quantified spelling alongside the bracketed index.
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -42,6 +42,22 @@ saying it sets an expectation their install may contradict. Say what is known, a
   collection whose elements are owned child records (`Cell.Persistent`) is sent to the record axis rather than handed
   a container call no verb takes. Inside
   a `compose=`'s nested `sets`, the path is named in the `path` slot it belongs to rather than `field_path`.
+- **A `where=` path step can now quantify a list, so "does ANY element match?" is one call.** Four tokens go in the
+  step where the list binds: `Conditions[*any].Data.Function = IsGuard` (some element), `[*all]` (every one, and
+  vacuously true on an empty list), `[*none]` (absence, proved — `Effects[*none].BaseEffect->editorid startswith REQ_`),
+  and `[*count]`, which yields the number of elements for a numeric or membership comparison (`Effects[*count] > 2`).
+  A quantified step composes with the `->` link step and with another quantified step, at the cost the tool
+  description declares (per-candidate work times list length, and one winner fetch per element under `->`). The bare
+  `[*]` is the element SET and is refused in `where=`, naming the three fold tokens; a quantifier pointed at a field
+  that is not a list on the scanned records is named in the scan's accounting with what that step read instead, never
+  counted as a silent non-match. Reaching a list with a dotted segment still refuses, and its remedy now offers the
+  quantified spelling alongside the bracketed index.
+- **`where=` gains the two exclusion folds of `has`, and `references=` gains a negated entry.**
+  `BodyTemplate.FirstPersonFlags has_any 44,45` is true when at least one operand bit is set and `has_none` when
+  none is, beside the existing `has` (every operand bit set). A `!` before a `references=` entry inverts it —
+  `references=["!XXXXXX:A.esm"]` keeps only the records that do NOT reference that target — and plain and negated
+  entries in one call compose by AND. The negated form takes the same bounding `types=`/`plugins=`/`formids=` scope
+  the plain form takes.
 - **`housecarl_load_order_status` now names the running server's build in its header** — a `server:` line carrying
   the informational version the binary embeds (the release version `build-plugin.ps1` stamps from `plugin.json`,
   then `+` and the full commit sha: `1.9.5-dev+e942910…`), so checking that the installed houseCARL is the build you expect no longer means
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -61,8 +61,9 @@ saying it sets an expectation their install may contradict. Say what is known, a
   none is, beside the existing `has` (every operand bit set). The operand names flag bits — a flag name, a
   comma-separated list of them, or a bit mask (`0xC000`) — never a biped slot number. A `!` before a `references=` entry inverts it —
   `references=["!XXXXXX:A.esm"]` keeps only the records that do NOT reference that target — and plain and negated
-  entries in one call compose by AND. The negated form takes the same bounding `types=`/`plugins=`/`formids=` scope
-  the plain form takes.
+  entries in one call compose by AND. The sigil takes the `@file` spelling the plain entry takes, so
+  `references=["!@C:/work/targets.jsonl"]` excludes every target a list file or result artifact names. The negated
+  form takes the same bounding `types=`/`plugins=`/`formids=` scope the plain form takes.
 - **`housecarl_load_order_status` now names the running server's build in its header** — a `server:` line carrying
   the informational version the binary embeds (the release version `build-plugin.ps1` stamps from `plugin.json`,
   then `+` and the full commit sha: `1.9.5-dev+e942910…`), so checking that the installed houseCARL is the build you expect no longer means
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -49,12 +49,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
   A quantified step composes with the `->` link step and with another quantified step, at the cost the tool
   description declares (per-candidate work times list length, and one winner fetch per element under `->`). The bare
   `[*]` is the element SET and is refused in `where=`, naming the three fold tokens; a quantifier pointed at a field
-  that is not a list on the scanned records is named in the scan's accounting with what that step read instead, never
-  counted as a silent non-match. Reaching a list with a dotted segment still refuses, and its remedy now offers the
-  quantified spelling alongside the bracketed index.
+  that is not a list — a scalar, a dict, a raw byte block — is named in the scan's accounting with what that step read
+  instead, never counted as a silent non-match. An absent list reads as EMPTY, and so does one whose parent substruct
+  is absent, so `VirtualMachineAdapter.Scripts[*count] = 0` finds the records that carry no script at all; where an
+  element cannot be judged, the fold reports no verdict rather than claiming one. Reaching a list with a dotted
+  segment still refuses, and its remedy now offers the quantified spelling alongside the bracketed index.
+  `project.fields` does not take these tokens yet and refuses them by name.
 - **`where=` gains the two exclusion folds of `has`, and `references=` gains a negated entry.**
-  `BodyTemplate.FirstPersonFlags has_any 44,45` is true when at least one operand bit is set and `has_none` when
-  none is, beside the existing `has` (every operand bit set). A `!` before a `references=` entry inverts it —
+  `BodyTemplate.FirstPersonFlags has_any Head,Hands` is true when at least one operand bit is set and `has_none` when
+  none is, beside the existing `has` (every operand bit set). The operand names flag bits — a flag name, a
+  comma-separated list of them, or a bit mask (`0xC000`) — never a biped slot number. A `!` before a `references=` entry inverts it —
   `references=["!XXXXXX:A.esm"]` keeps only the records that do NOT reference that target — and plain and negated
   entries in one call compose by AND. The negated form takes the same bounding `types=`/`plugins=`/`formids=` scope
   the plain form takes.
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -802,7 +802,9 @@ public sealed class FieldPredicateSet
         if (WriteEngine.ClosedInterface(t, typeof(IList<>)) is not null
             || WriteEngine.ClosedInterface(t, typeof(IReadOnlyList<>)) is not null)
             return null;
-        return $"a single {RecordNaming.StripOverlay((val?.GetType() ?? t).Name)} value";
+        // Both strips, in that order: a CARRIED value reflects as BodyTemplateBinaryOverlay and a null one falls back
+        // to the declared IBodyTemplateGetter, so without both the same field is named two ways on two records.
+        return $"a single {RecordNaming.StripGetterInterface(RecordNaming.StripOverlay((val?.GetType() ?? t).Name))} value";
     }
 
     /// <summary>Fold one step's element verdicts into the record's. Same accounting shape
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -180,10 +180,14 @@ public sealed class FieldPredicateSet
     /// <summary>One quantified step of one predicate: the segments of the side it sits on, which one carries the
     /// fold, how it is spelled, and the predicate's own text for the message. A scan with a NAMED type scope walks
     /// these against the schema, so "that step is not a list on this type" refuses the call rather than becoming a
-    /// whole-scan accounting note.</summary>
-    public readonly record struct QuantifiedStep(IReadOnlyList<string> Path, int Index, string Token, string Text);
+    /// whole-scan accounting note. <paramref name="OnScannedType"/> says whether the step is rooted at the SCANNED
+    /// record type: the right side of a <c>-&gt;</c> is rooted at the link TARGET's type instead, and asking the
+    /// scanned type's schema about it would judge a field it was never on.</summary>
+    public readonly record struct QuantifiedStep(IReadOnlyList<string> Path, int Index, string Token, string Text,
+                                                 bool OnScannedType);
 
-    /// <summary>Every quantified step in the set, both sides of a <c>-&gt;</c> included.</summary>
+    /// <summary>Every quantified step in the set, both sides of a <c>-&gt;</c> included — each saying which type it
+    /// is rooted at.</summary>
     public IReadOnlyList<QuantifiedStep> QuantifiedSteps
     {
         get
@@ -191,16 +195,18 @@ public sealed class FieldPredicateSet
             var steps = new List<QuantifiedStep>();
             foreach (var p in _predicates)
             {
-                Collect(p.LinkPath, p.LinkFolds, p);
-                Collect(p.PathSegments, p.PathFolds, p);
+                // The link side always roots at the scanned type; the path side does only when there is no link,
+                // because with one it is the tail read on the resolved target.
+                Collect(p.LinkPath, p.LinkFolds, p, true);
+                Collect(p.PathSegments, p.PathFolds, p, p.LinkPath is null);
             }
             return steps;
 
-            void Collect(string[]? segs, Fold[]? folds, Predicate p)
+            void Collect(string[]? segs, Fold[]? folds, Predicate p, bool onScanned)
             {
                 if (segs is null || folds is null) return;
                 for (int i = 0; i < segs.Length && i < folds.Length; i++)
-                    if (folds[i] != Fold.None) steps.Add(new QuantifiedStep(segs, i, FoldToken(folds[i]), p.Text));
+                    if (folds[i] != Fold.None) steps.Add(new QuantifiedStep(segs, i, FoldToken(folds[i]), p.Text, onScanned));
             }
         }
     }
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -177,6 +177,34 @@ public sealed class FieldPredicateSet
     /// <summary>Candidate bodies tested so far — the denominator the accounting reports against.</summary>
     public long Scanned => _scanned;
 
+    /// <summary>One quantified step of one predicate: the segments of the side it sits on, which one carries the
+    /// fold, how it is spelled, and the predicate's own text for the message. A scan with a NAMED type scope walks
+    /// these against the schema, so "that step is not a list on this type" refuses the call rather than becoming a
+    /// whole-scan accounting note.</summary>
+    public readonly record struct QuantifiedStep(IReadOnlyList<string> Path, int Index, string Token, string Text);
+
+    /// <summary>Every quantified step in the set, both sides of a <c>-&gt;</c> included.</summary>
+    public IReadOnlyList<QuantifiedStep> QuantifiedSteps
+    {
+        get
+        {
+            var steps = new List<QuantifiedStep>();
+            foreach (var p in _predicates)
+            {
+                Collect(p.LinkPath, p.LinkFolds, p);
+                Collect(p.PathSegments, p.PathFolds, p);
+            }
+            return steps;
+
+            void Collect(string[]? segs, Fold[]? folds, Predicate p)
+            {
+                if (segs is null || folds is null) return;
+                for (int i = 0; i < segs.Length && i < folds.Length; i++)
+                    if (folds[i] != Fold.None) steps.Add(new QuantifiedStep(segs, i, FoldToken(folds[i]), p.Text));
+            }
+        }
+    }
+
     // ======================================================================
     //  PARSE — "<path> <op> <value>", longest-match the operator.
     //  The path is dotted/bracketed identifiers (no whitespace, no operator
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -33,8 +33,14 @@ namespace HousecarlCore;
 /// membership is <c>references=</c>'s job. The PRESENCE operators (<c>exists</c>/<c>missing</c>) are the exception:
 /// they DO match a carried substruct/list leaf (present and non-empty), the "which records carry a VMAD/Effects"
 /// query. The MEMBERSHIP operators (<c>formid in</c>/<c>formid not in</c> a supplied list) are the other non-leaf
-/// case: they test the record's IDENTITY against a pre-parsed FormKey set, no read walk at all.
-/// A wildcard over a list (<c>Effects[*].Magnitude &gt; 50</c>) is a deliberate future extension.</para>
+/// case: they test the record's IDENTITY against a pre-parsed FormKey set, no read walk at all.</para>
+///
+/// <para><b>The quantified step.</b> A path step may declare its multiplicity and its fold where it binds:
+/// <c>Conditions[*any].Data.Function = IsGuard</c>, <c>Effects[*none].BaseEffect-&gt;editorid startswith REQ_</c>,
+/// <c>Effects[*count] &gt; 2</c>. <c>[*any]</c>/<c>[*all]</c>/<c>[*none]</c> fold the elements into a boolean and
+/// <c>[*count]</c> into their number; the bare <c>[*]</c> is the element SET and is refused here (a set is not a
+/// boolean). The fold is the same one <see cref="EvalLinkStep"/> already runs over link targets, with the fan-out
+/// source swapped to the step's elements — so it composes with the <c>-&gt;</c> link step and with itself.</para>
 /// </summary>
 public sealed class FieldPredicateSet
 {
@@ -44,7 +50,9 @@ public sealed class FieldPredicateSet
     /// <see cref="Has"/> is a BITWISE set-test for a <c>[Flags]</c> enum (or plain integer) leaf — true iff every
     /// bit of the operand is set on the field, regardless of other bits — so a multi-slot BodyTemplate still
     /// matches the one slot asked for, which <see cref="Eq"/> (exact value) and the range ops cannot express. Its
-    /// operand is a bit value (decimal or <c>0x</c> hex) or a flag NAME.
+    /// operand is a bit value (decimal or <c>0x</c> hex) or a flag NAME. <see cref="HasAny"/> and
+    /// <see cref="HasNone"/> are the other two folds over the SAME bits — any bit of the operand set, and none of
+    /// them set — the exclusion terms a slot sweep needs, spelled to match the path step's any/all/none.
     /// <see cref="Exists"/>/<see cref="Missing"/> are PRESENCE tests that take NO operand — true iff the path
     /// resolves to a present, NON-EMPTY value (a scalar OR a carried substruct/list) / its complement. They are the
     /// only operators that MATCH a no-value container leaf: the "which records CARRY a VirtualMachineAdapter /
@@ -55,7 +63,13 @@ public sealed class FieldPredicateSet
     /// read cleave where identity sits beside Fields) and a list operand: inline comma-separated FormIDs, or
     /// <c>@&lt;absolute path&gt;</c> naming a file of them. Restricted to <c>formid</c> at parse (a named refusal on any
     /// other path) so a future generalization to leaf-value membership is an extension, not a behavior change.</summary>
-    enum Op { Eq, Ne, Gt, Ge, Lt, Le, Contains, StartsWith, Has, Exists, Missing, In, NotIn }
+    enum Op { Eq, Ne, Gt, Ge, Lt, Le, Contains, StartsWith, Has, HasAny, HasNone, Exists, Missing, In, NotIn }
+
+    /// <summary>A path step's declared multiplicity and fold. <see cref="None"/> is an ordinary step;
+    /// <see cref="Set"/> is the bare <c>[*]</c> (the element set — a PROJECT reading, refused in a predicate);
+    /// <see cref="Any"/>/<see cref="All"/>/<see cref="NoneOf"/> fold the elements into a boolean and
+    /// <see cref="Count"/> into their number.</summary>
+    enum Fold { None, Set, Any, All, NoneOf, Count }
 
     /// <summary>One parsed predicate: the split path segments (fed straight to <see cref="ReadEngine.ReadLeaf"/>),
     /// the operator, the raw operand, and — for a numeric operator — the operand pre-parsed to a double (validated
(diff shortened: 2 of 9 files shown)


src/housecarl-core/FieldPredicate.cs
```diff
@@ -609,8 +700,92 @@ EvalKind ClassifyNoField(string note)
...
+            var (sat, kind) = eval(e);
+            if (_fatal is not null) return (false, EvalKind.Definite);
+            if (kind == EvalKind.Definite) { anyVerdict = true; if (sat) anyTrue = true; else anyFalse = true; }
+            else if (kind is EvalKind.NoField or EvalKind.ListHop or EvalKind.NotAList)
+            { anyNoField = true; anyListHop |= kind == EvalKind.ListHop; }
+        }
+        if (fold == Fold.Any && anyTrue) return (true, EvalKind.Definite);
+        if (fold == Fold.NoneOf && anyTrue) return (false, EvalKind.Definite);
+        if (fold == Fold.All && anyFalse) return (false, EvalKind.Definite);
+        if (anyVerdict) return (fold != Fold.Any, EvalKind.Definite);
```
Avick3110, line 842: `[*all]` / `[*none]` report a DEFINITE verdict while silently discarding the elements that produced no verdict.

Scenario: `where=["Effects[*none].Data.Magnitude > 50"]` on a record with three effects, where effect #1 throws inside Mutagen (Unreadable) or its sub-path is not a field on that element's concrete arm (NoField), and effects #2/#3 read fine and are both false. The loop sets `anyVerdict = true`, `anyTrue = false`, so this line returns `(true, Definite)` — the record is reported as a proved absence even though one element was never judged. Same for `[*all]` returning true when an unjudged element sits in the list. That is the false-positive direction on the query the PR sells as "absence, proved", and the summary rollup stays silent because the record counted as `_valueRead`.

The method's own doc says "where no element could be judged the no-verdict class carries out" — but that only fires when NO element was judged. For `[*all]`/`[*none]` a single unjudged element already invalidates the universal claim, so the no-verdict class should carry out whenever an element went unjudged and the fold has not already been decided false.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -609,8 +700,92 @@ EvalKind ClassifyNoField(string note)
...
+    (bool Satisfied, EvalKind Kind) FoldOver(Predicate p, List<object> elems, Fold fold, Func<object, (bool, EvalKind)> eval)
+    {
+        if (elems.Count == 0) return (fold != Fold.Any, EvalKind.Definite);
+        bool anyVerdict = false, anyTrue = false, anyFalse = false, anyNoField = false, anyListHop = false;
+        foreach (var e in elems)
+        {
+            var (sat, kind) = eval(e);
+            if (_fatal is not null) return (false, EvalKind.Definite);
+            if (kind == EvalKind.Definite) { anyVerdict = true; if (sat) anyTrue = true; else anyFalse = true; }
+            else if (kind is EvalKind.NoField or EvalKind.ListHop or EvalKind.NotAList)
```
Avick3110: An element whose verdict is `Unset` falls through both arms: it is not counted in `anyVerdict` and not counted in `anyNoField`. When EVERY element is Unset the fold ends at the last line returning `EvalKind.Unreadable`.

Concrete scenario, using the PR's own headline example: `where=["Effects[*none].BaseEffect->editorid startswith REQ_"]`. Each element goes through `JudgeTargets`, which returns `(false, EvalKind.Unset)` for `links.Count == 0` and for the `(absent)` note. A record whose effects all carry a null/absent `BaseEffect` link therefore produces Unset for every element, so the record is (a) dropped, though `[*none]` is vacuously true over "no REQ_ base effect anywhere", and (b) charged to `_unreadable[k]`, which the rollup renders as a Mutagen read fault — a wrong diagnosis, since nothing failed to read.

The same shape bites `[*all]`: an all-Unset list should be vacuously true (it already is for an EMPTY list, three lines up), but here it is Unreadable. Unset elements need an explicit decision — either treated as "no counterexample" for `all`/`none`, or carried out as `Unset` rather than `Unreadable`.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4036,7 +4041,7 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
                         }
                         if (conflictsOnly && (view.TouchingPlugins(fk)?.Count ?? 0) <= 1) continue;
                         if (DeletedRecordRule.HasNoLiveBody(body)
-                            && (refSet is not null || predicate is { NeedsLiveBody: true })) continue;
+                            && (refSet is not null || refNone is not null || predicate is { NeedsLiveBody: true })) continue;
```
Avick3110: Adding `refNone` to this skip inverts its meaning. For the POSITIVE `refSet` the skip is right: a deleted record has no live body, so it cannot link to the target and could never match. For the NEGATED term the same record is the strongest possible match — it references nothing at all, so "keeps only the records that do NOT reference that target" must keep it.

Scenario: `types=["WEAP"], references=["!000ABC:Skyrim.esm"]` over a scope containing a deleted WEAP override. The record silently disappears from the result set, and no accounting line says so.

`ExcludedByReference`'s own doc comment states the intent — "A record that carries no links at all references nothing and is kept — that is the whole point of the term" — which this guard contradicts. Same at lines 4166 and 4416: `refNone` should be left out of the `HasNoLiveBody` skip, since a no-live-body record simply passes the exclusion test.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,22 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
+  vacuously true on an empty list), `[*none]` (absence, proved — `Effects[*none].BaseEffect->editorid startswith REQ_`),
+  and `[*count]`, which yields the number of elements for a numeric or membership comparison (`Effects[*count] > 2`).
+  A quantified step composes with the `->` link step and with another quantified step, at the cost the tool
+  description declares (per-candidate work times list length, and one winner fetch per element under `->`). The bare
+  `[*]` is the element SET and is refused in `where=`, naming the three fold tokens; a quantifier pointed at a field
+  that is not a list on the scanned records is named in the scan's accounting with what that step read instead, never
+  counted as a silent non-match. Reaching a list with a dotted segment still refuses, and its remedy now offers the
+  quantified spelling alongside the bracketed index.
+- **`where=` gains the two exclusion folds of `has`, and `references=` gains a negated entry.**
+  `BodyTemplate.FirstPersonFlags has_any 44,45` is true when at least one operand bit is set and `has_none` when
```
Avick3110: `BodyTemplate.FirstPersonFlags has_any 44,45` does not mean what a reader of this changelog will take it to mean, and the call it teaches silently returns the wrong records.

`TryResolveBits` first tries `TryBits("44,45")`, which fails on the comma (`NumberStyles.None`), then falls through to `ReadEngine.TryEnumBitsFromName` -> `Enum.Parse(BipedObjectFlag, "44,45")`. `Enum.Parse` accepts NUMERIC components in a comma list and ORs them, so the operand resolves to bit mask `44 | 45 = 45` (bits 0, 2, 3, 5). Meanwhile `FlagSlotDisplay` in ReadEngine documents this project's own convention that on `BipedObjectFlag` the numbers a modder reasons in are SLOTS, with `slot = 30 + bit index` — so biped slots 44 and 45 are `0x4000 | 0x8000` = `49152`, nowhere near 45.

Use flag NAMES (as the tests do: `has_any Head,Hands`) or the real mask (`has_any 0xC000`), and say the operand is a bit value, never a slot number.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -609,8 +700,92 @@ EvalKind ClassifyNoField(string note)
...
+
+    /// <summary>Navigate to a quantified step's collection and hand back its elements (with the collection's owning
+    /// parent, which the element's token emit needs). An ABSENT collection reads as EMPTY — the same reading
+    /// <see cref="ReadEngine.KeywordKeys"/> already gives a record with no list. A step that is not a list at all
+    /// is a named no-verdict, never a silent non-match.</summary>
+    (List<object>? Elements, object? Parent, EvalKind? Miss) ElementsAt(object obj, string[] segs, Fold[] folds, int from, int q)
+    {
+        var (ok, val, parent, note) = ReadEngine.NavigateTo(obj, segs[from..(q + 1)]);
+        if (!ok) return (null, null, ClassifyMiss(note ?? ""));
+        if (val is null) return (new List<object>(), parent, null);
```
Avick3110: "An absent collection reads as EMPTY" only holds when the LEAF property itself returns null. If the collection is absent because a mid-path substruct is null, `NavigateValue` returns `ok = false` with `AbsentNote`, so `ClassifyMiss` classifies it `Unset` and the record gets no verdict at all.

Scenario: `where=["VirtualMachineAdapter.Scripts[*count] = 0"]` — the natural "which records carry no scripts" query. A record with a VMAD and an empty Scripts list matches (count 0). A record with NO VirtualMachineAdapter at all — the much larger group, and the one the caller most wants — is dropped with no verdict, silently, because the rollup only shouts when every scanned record was no-value. Same asymmetry for `Foo.Bar[*none] ...`.

Either the mid-path `(absent)` case should also read as an empty collection for a quantified step, or the doc comment and the tool description should not claim absence reads as empty.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -609,8 +700,92 @@ EvalKind ClassifyNoField(string note)
...
+        if (val is string || val is not System.Collections.IEnumerable en)
```
Avick3110: "Is a list" is decided by bare `IEnumerable`, which is broader than the read engine's own notion of a collection (`WriteEngine.ClosedInterface(..., IReadOnlyList<>/IDictionary<>)`, used in `EmitToken`). Two shapes slip through as lists instead of getting the new, correct `NotAList` refusal:

- Noggog byte slices. `MemorySlice<byte>`/`ReadOnlyMemorySlice<byte>` implement `IReadOnlyList<byte>`, so `where=["Model.Data[*count] > 0"]` counts BYTES and `Model.Data[*any] = 5` bit-picks one, both reported as definite verdicts rather than "that step is not a list".
- Dictionaries. A dict-valued step enumerates `KeyValuePair<,>` elements; `EmitToken` has no branch for those, so each falls to `LeafRead.Container(...)`, `DecideLeaf` sees `!HasValue` with a `[`-leading note and returns `Container` — so `SomeDict[*any] = X` can never match and never says why (`Container` is not one of the classes `FoldOver` carries out, so the fold ends at `Unreadable`).

Gating on the same `ClosedInterface` test the emit side uses would put both in the `NotAList` bucket the PR just built.

Avick3110:
Second pass, blind, against CLAUDE.md's design section, the E1 design page and the SPEC §2 / §2.2 / §5.3 blocks dated 2026-09-05. Six correctness findings are already posted inline; these are one more correctness bug and five conformance gaps that do not attach cleanly to a single changed line.

### 1. high — `src/housecarl-core/FieldPredicate.cs:742` — a quantifier on a NULL non-list field reads as an empty list, so `[*all]`/`[*none]` match vacuously

`ElementsAt` tests `val is null` before it tests "is this a list", so the new `NotAList` verdict is only reachable when the misused field happens to be non-null on that record. `BodyTemplate[*all].FirstPersonFlags has Head` returns a definite TRUE for every ARMO with no BodyTemplate, and `Name[*count] = 0` matches every record whose Name is null. The loud accounting sentence never fires for those records — it is bypassed exactly where the answer is a false positive. `WhereQuantifierTests`' non-list case only covers armors that all carry a BodyTemplate. Order the two checks the other way, and test the null case for both folds and for `[*count]`.

### 2. medium — conformance: SPEC §2.2 files "any quantifier on a non-list step" under "Refusals, all schema-declared"; this ships it as a whole-scan accounting note

Every record gets a definite non-match and the caller learns about it only when `_noField == _scanned`. The generator already carries the per-type cardinality the refusal would name. Either the refusal lands, or the SPEC line changes on Aaron's word — as written the PR takes the second silently.

### 3. medium — conformance: `Closes #458` with the PROJECT half absent

The design page frames #458 as "SELECT predicate grammar **+ PROJECT paths**", and SPEC §2.2's amendment is explicit: `fields` takes `[*]` and `[*count]`, `dense` gains its row-per-element reading, and `[*any]` in `project.fields` is refused as "a boolean is not a row". None of that is here and no such refusal exists. Step 1 of 3 is fine — but then #458 should not be closed by this PR, or the remainder should be filed.

### 4. medium — `src/housecarl-core/WriteEngine.cs:2751` — the new refusal asserts the fold tokens "are where= path steps"

Per SPEC §2.2, `[*]` and `[*count]` are PROJECT and walk-path tokens, and `[*]` is *only* legal there. So the sentence a caller gets for a spelling the SPEC calls legal tells them it belongs to a different parameter. The guard also sits on the list branch only — a dict-typed field with `[*any]` still comes back as a malformed key.

### 5. medium — conformance: SPEC §2's amendment requires the quantified step's cost "named in the schema beside the token"

`src/housecarl-mcp/RecordsTools.cs:102` gained the four tokens with no mention of the list-length multiplier or the one-winner-fetch-per-element under `->`. The changelog entry claims the opposite ("at the cost the tool description declares"). One sentence closes it.

### 6. medium — the `!<formid>` spelling settles #531 but forecloses what the design page says #489 buys it

The PR reads the negated term as "does not reference this target", which is exactly #531's reporter's job and is right. But SPEC §2.2's gloss for the same term is "referenced by nothing in the scope", and the design page §4 says the unbounded version is "referenced by nothing anywhere — the orphan sweep", arriving "on the same spelling" under §3.2. Those are different questions: unbounding `references=["!X"]` yields "every record in the order that does not reference X", not the orphans. Worth Aaron's word now rather than at #489, because it decides whether the orphan sweep needs a second spelling.

**Two smaller ones, no action implied.** "An absent list reads as empty" is a new schema-visible semantic and the description states only that `[*all]` is vacuously true on an *empty* list. And `has_none` cannot answer #531's own "no armor slot" case: a flags leaf with no value is a no-verdict, not a match.

Avick3110:
Review fold — thirteen findings read, eleven folded, one turned into a stated remainder, one left for Aaron. Build, 1202 tests (was 1190) and 128/128 probes green. Five commits, no history rewritten (a stacked branch is building on this one).

## Folded

**1 (high) — a quantifier on a null non-list field read as an empty list.** Real, and worse than the null case alone: the shape was decided on the live value, so anything that merely enumerates passed too. `ElementsAt` now judges the step on its DECLARED type through the same `ClosedInterface` test the emit side uses, before the null check. A null `BodyTemplate[*all]` is refused as a non-list; a null list is still empty. Tests: `ANullNonListStepIsStillNamedNotAList_NeverAVacuouslyTrueEmptyList`, `ANullScalarWithCountIsNotAnEmptyList`.

**Byte slices and dicts treated as lists.** Same fix, same place. A `MemorySlice`/`ReadOnlyMemorySlice` reads as "a raw block of bytes" and a dict as "a dict of keyed entries", both in the `NotAList` bucket. Tests: `AByteBlockStepIsNamedBytes_NotCountedElementByElement`, `ADictStepIsNamedADict_NotFoldedOverItsEntries`.

**Null mid-path substruct gave Unset, so `[*count] = 0` missed the VMAD-less records.** Real. A mid-path `(absent)` now reads as an absent collection, which reads as empty — the doc comment's claim is now true of both spellings. Test: `AnAbsentParentSubstructReadsAsAnEmptyCollection` (every armor with no VMAD answers `VirtualMachineAdapter.Scripts[*count] = 0`).

**`[*all]`/`[*none]` returned Definite while discarding no-verdict elements.** Real, and the same hole in the other direction for `[*any]`. `FoldOver` now carries the no-verdict class out whenever an element went unjudged and the judged ones have not already decided the fold: a true settles `any`/`none`, a false settles `all`, anything short of that is a claim over an element nobody read. Test: `OneUnjudgedElementStopsAllAndNoneClaimingAVerdict`.

**Unset element verdicts fell through both arms and ended as Unreadable.** Real — a wrong diagnosis, since nothing failed to read. Unset is now one of the classes the fold carries out, ranked below a read fault and a schema miss, so an all-null-link list reports no verdict rather than a Mutagen fault. I took your second option (carry Unset out) over the first (treat it as no counterexample): at the top level an Unset leaf yields no verdict for the record, and a fold should not be the one place where a value-less read silently becomes a proof. Test: `AnAllUnsetListIsNoVerdict_NotAReadFault`.

**`refNone` in the `HasNoLiveBody` skip inverted the negated term.** Real. The three skips are back to `refSet`-only, and `ExcludedByReference` returns false for a record with no live body — it references nothing, which is the strongest match the term has. Test: `ANegatedReferenceKeepsADeletedRecord_WhichReferencesNothingAtAll`, verified against the pre-fix code (it fails there).

**CHANGELOG `has_any 44,45`.** Real and embarrassing — the example taught slots where the operand is a bit value. Now `has_any Head,Hands`, with a sentence saying the operand is a flag name, a comma list of names, or a mask, never a slot number. (The `Enum.Parse` numeric-comma-OR behaviour itself is `TryResolveBits`, shared with the existing `has`, and predates this PR; I have not changed it.)

**2 — the non-list case belongs in schema-declared refusals, not an accounting note.** Folded rather than sent to Aaron, because the corpus does carry the cardinality. `CorpusRulebook.StepCardinality` walks the quantified step against the schema and the scan refuses before it starts, naming the step's real cardinality ("it is a substruct on Armor"). It only fires where `types=` names the type set and the step is a non-list on *every* named type — a union arm that does hold a list still runs, and an unscoped or undecidable scan keeps the per-record accounting as the backstop, since no schema answer exists there. Tests: `AQuantifierOnANonListStepIsRefusedFromTheSchemaWhenTheTypeIsNamed`, `AQuantifierOnAListStepPassesTheSchemaCheck`.

**4 — the `StepIntoElement` refusal misfiled the tokens and sat on the list branch only.** Real on both counts. The guard moved above the gendered and dict branches so every shape gets it, and the sentence now says where each token is legal: `[*any]`/`[*all]`/`[*none]` fold a list in `where=`, `[*]` and `[*count]` are project/walk steps.

**5 — the quantified step's cost is not named in the schema.** Real. The tool description now carries it beside the tokens: per-candidate work times list length, and one winner fetch per element under `->`, with the `Temporary[*any]->…` example the SPEC uses.

**The two "smaller ones".** Both folded into the changelog and the tool description: the absent-list-reads-as-empty semantic is now stated (with the parent-substruct case), and the CHANGELOG no longer implies `has_none` answers "no armor slot" — a flags leaf with no value is still a no-verdict, which is the honest answer for a record that carries no BodyTemplate at all.

## Turned into a stated remainder, not closed

**3 — `Closes #458` with the PROJECT half absent.** You are right that the half is missing. I did not build it: SPEC §2.2's amendment names `fields` taking `[*]`/`[*count]` and `dense`'s row-per-element reading, but does not say what `[*]` renders in `format=text`/`json`, what two quantified paths of different lengths in one `fields=` list produce (a product, an alignment, or a refusal), or which identity columns the extra dense rows carry. Guessing a row shape that a W6 skill would then be authored against is the failure the vocabulary gate exists to prevent. So the PR body now says `Refs #458 (SELECT half)` with the remainder written out, and `project.fields` refuses the tokens by name — `[*any]` as "a boolean is not a row" per the SPEC, `[*]`/`[*count]` as "not built yet" — so nothing reads them silently in the meantime. Tests for both refusals.

## Left for Aaron

**6 — what `!<formid>` means, and what unbounding buys #489.** You are right that these are two different questions on one spelling, and the SPEC gloss ("referenced by nothing in the scope") reads as the second while the implementation answers the first.

My view, plainly: the two questions differ only in what is quantified. "Does not reference X" quantifies over the *candidate's* links with the target fixed. "Referenced by nothing" quantifies over the *whole scope's* links with the candidate fixed. Today's bounded term can only answer the first, because there is no target to fix in the second — and the orphan sweep is the second with the scope unbounded. So unbounding `references=["!X"]` gives "every record in the order that does not reference X", which is a very large and not very interesting answer, and not the orphan sweep.

That leaves three ways out, and it is Aaron's call which:
1. Keep `!<formid>` as "does not reference this target" (what #531 asked for and what this PR ships) and give the orphan sweep its own spelling — something like `references=["!*"]` or a `referenced_by=` term with an empty target set. Two spellings, each unambiguous, and the orphan sweep is honest about being a different question.
2. Read `!X` as "referenced by nothing in the scope" and drop the target — but then `!X` names a target it does not use, which is a spelling that lies.
3. Keep one spelling and let the presence of a bounding scope switch the meaning. I would argue against this one: the same call means two different things depending on another parameter, which is the silently-wrong-answer shape.

I would take (1), and I would settle it now rather than at #489, because if the orphan sweep needs a second spelling then that spelling is vocabulary and belongs before W6. Nothing in this PR blocks any of the three; the change is confined to `SplitReferenceTargets` and the tool description.

## Refuted

Nothing. Every finding was either real as written or real in a wider form than written.

Avick3110:
Blind review folded (11 folded, 2 absorbed into wider fixes, 0 refuted). #458 PROJECT half deliberately left as a stated remainder. Queued for Aaron's review; the !<formid> semantics ruling is on this PR. Stacked follow-ons: #558 (#459), then #489.

Avick3110:
Update on the !<formid> question: #562's fold read SPEC §2.2's amendment and the design page §4, both of which already say the unbounded negated form is the orphan sweep, and implemented that. Bounded !X keeps the does-not-reference reading shipped here; the response says which question it answered. Still Aaron's to confirm, but no longer unruled.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4272,6 +4282,48 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
+            foreach (var ts in Rulebook.RecordTypesNamed(token))
+                if (!schemas.Contains(ts)) schemas.Add(ts);
+        if (schemas.Count == 0) return null;
+
+        foreach (var step in predicate.QuantifiedSteps)
+        {
+            var whatItIs = new List<string>();
+            foreach (var ts in schemas)
+            {
+                var card = Rulebook.StepCardinality(ts, step.Path, step.Index);
```
Avick3110, line 4345: **The link step's right-hand path is checked against the wrong record type.**

`QuantifiedSteps` collects both sides with no marker for which side a step came from:

```
Collect(p.LinkPath, p.LinkFolds, p);
Collect(p.PathSegments, p.PathFolds, p);
```

For a `->` predicate, `p.PathSegments` is the RIGHT side, rooted at the link TARGET's type — but `StepCardinality(ts, step.Path, step.Index)` here roots every collected step at the scanned `types=` schema.

Scenario: `types=["PlacedObject"], where=["EnableParent.Reference->DistantLodData[*count] > 0"]`. The step arrives as `(Path=["DistantLodData"], Index=0)`, so the schema is asked about `PlacedObject.DistantLodData`, which the corpus gives as cardinality `value` — the call is refused with "it is a value on PlacedObject", even though the step lands on the target (`PlacedBarrier`/`PlacedTrap`/`PlacedBeam`, where it IS a list). The mirror case is a silently missed refusal: a right-side step whose name happens to be a list on the scanned type passes unchecked.

There is a third consequence: `if (card is null) return null;` abandons the whole method, so one link predicate whose right-side field simply does not exist on the scanned type also silences the check for every legitimately checkable path-side step in the same call.

The fix is a side flag on `QuantifiedStep` (or collecting only `p.LinkPath` plus, when `LinkPath is null`, `p.PathSegments`) so a right-side step is skipped rather than checked against a type it was never rooted at.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -609,8 +728,135 @@ EvalKind ClassifyNoField(string note)
...
+        var name = t.Name;
+        if (name.StartsWith("MemorySlice", StringComparison.Ordinal) || name.StartsWith("ReadOnlyMemorySlice", StringComparison.Ordinal))
+            return "a raw block of bytes";
+        if (WriteEngine.ClosedInterface(t, typeof(IDictionary<,>)) is not null
+            || WriteEngine.ClosedInterface(t, typeof(IReadOnlyDictionary<,>)) is not null)
+            return "a dict of keyed entries";
+        if (WriteEngine.ClosedInterface(t, typeof(IList<>)) is not null
+            || WriteEngine.ClosedInterface(t, typeof(IReadOnlyList<>)) is not null)
+            return null;
+        return $"a single {RecordNaming.StripOverlay((val?.GetType() ?? t).Name)} value";
```
Avick3110: **The same field gets two different names depending on whether it is carried.**

`RecordNaming.StripOverlay` strips only a trailing `BinaryOverlay`, so:

- carried `BodyTemplate` → `val.GetType().Name` is `BodyTemplateBinaryOverlay` → "a single **BodyTemplate** value"
- null `BodyTemplate` → falls back to `t.Name`, the declared `IBodyTemplateGetter` → "a single **IBodyTemplateGetter** value"

Both tests here only assert `Contains("not a list")`, so the split is invisible to the suite. `RecordNaming.StripGetterInterface` (`INpcGetter → Npc`) is the existing strip for the declared side — composing it with `StripOverlay` makes `BodyTemplate[*all]` read the same whichever record it lands on.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -213,6 +222,14 @@ public static string Records(
...
+        // The quantifier tokens are where='s, not a projection's — refused by name here so a caller who tries one
+        // gets the rule rather than an unreadable-field note from the read walk.
+        if (project?.fields is { Length: > 0 } pf)
+            foreach (var path in pf)
+                if (QuantifierToken(path) is { } qt)
+                    return Wire.Refuse(json, qt.ToLowerInvariant() is "[*any]" or "[*all]" or "[*none]"
+                        ? $"error: project.fields path '{path}' folds the elements into a boolean, and a boolean is not a row — use {qt} in where= to SELECT the records, and name a concrete element ('Effects[0].Data.Magnitude') or the list itself to read one."
+                        : $"error: project.fields path '{path}' uses '{qt}', which is a where= step today — the projection half of the quantified step (a row per element, and the element count as a cell) is not built yet. Name a concrete element ('Effects[0].Data.Magnitude') or the list itself, which renders as a summary with project.depth.");
```
Avick3110, line 239: **A token that is not a quantifier at all gets the "not built yet" answer.**

`QuantifierToken` returns any bracket key beginning `*`, and the refusal has only two branches: the three boolean folds, and everything else. So `project.fields=["Effects[*sum].Data.Magnitude"]` is refused with "uses '[*sum]', which is a where= step today — the projection half of the quantified step … is not built yet".

`[*sum]` is not a `where=` step: `FieldPredicateSet.SplitFolds` refuses it as "'[*sum]' is not a quantifier — the tokens are [*any], [*all], [*none] and [*count]". The caller is pointed at a future capability that will never cover their token instead of being told they mistyped it. A third branch for anything outside `[*]`/`[*count]`/`[*any]`/`[*all]`/`[*none]` — reusing the parser's wording — keeps the two surfaces saying the same thing about the same token.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1,4 +1,4 @@
-using System.Globalization;
+﻿using System.Globalization;
```
Avick3110: **A UTF-8 BOM was added here (and to `LoadOrderService.cs`) by accident.**

```
git show main:src/housecarl-core/WriteEngine.cs | head -c 3 | od -c   ->   u   s   i
git show HEAD:src/housecarl-core/WriteEngine.cs | head -c 3 | od -c   -> 357 273 277
```

Neither file had a BOM on `main`. That is why both diffs open with a whole-line rewrite of an unchanged `using …;` — the diff claims a touched line that was not edited, and the tree ends up with two differently-encoded C# files. Worth stripping before merge (the two new test files carry one too, though those are new so they only need to match the repo's convention).

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1016,17 +1033,13 @@ string ScanLane()
...
             {
-                var list = new List<FormKey>();
-                foreach (var r in refs)
-                {
-                    if (string.IsNullOrWhiteSpace(r)) continue;
-                    try { list.Add(door.Parse(r)); }
-                    catch (Exception ex) { return Wire.Refuse(json, $"error: bad references FormID '{r}': {ex.Message}. Expected 'XXXXXX:Plugin.esp'."); }
-                }
-                if (list.Count > 0) refFks = list.Distinct().ToList();
+                var (pos, neg, rerr) = SplitReferenceTargets(refs, door);
```
Avick3110, line 1047: **A negated entry cannot use the `@file` spelling `references=` already documents.**

`Artifacts.ExpandListInput` decides an element is an artifact/file by its FIRST character:

```
int atCount = items.Count(i => i is not null && i.TrimStart() is { Length: > 0 } t && t[0] == '@');
if (atCount == 0) return (items, null, null, null);
```

A `!@<path>` element starts with `!`, so `atCount` is 0 and the array passes through untouched. `SplitReferenceTargets` then strips the `!` and hands `@C:/work/reqmgef.jsonl` to `door.Parse`, which throws — the caller gets `error: bad references FormID '!@C:/work/reqmgef.jsonl' … Expected 'XXXXXX:Plugin.esp'`.

So "exclude everything in this artifact" is unaskable, and the refusal names the wrong problem (it reads as though the path were a malformed FormID rather than an unsupported combination). Either strip the sigil before expansion, or refuse `!@` by name so the caller learns the rule.

Avick3110:
Aaron's review folded — five findings read, five folded, none refuted. Build, 1207 tests (was 1202) and 128/128 probes green. Five commits appended, no history rewritten (#558 and #562 are stacked on this branch).

## Folded

**The link step's right-hand path checked against the wrong record type.** Real, all three consequences. `QuantifiedStep` now carries `OnScannedType`: the `->` LEFT side always roots at the scanned type, the path side only when there is no link, and `QuantifierShapeRefusal` skips a step it is not rooted at rather than asking the scanned type's schema about a field the step never landed on. The third consequence went with it — `card is null` now skips that STEP instead of returning from the whole method, so one uncheckable right-side step no longer silences every checkable step in the same call. Tests: `ALinkPredicatesRightSideIsNotJudgedAgainstTheScannedType` (`Effects[*any].BaseEffect->Name[*count] > 0` on `types=["SPEL"]` was refused as "a substruct on Spell"), `AnUncheckableRightSideStepDoesNotSilenceACheckableOne`. Both fail on the pre-fold tree.

**The same field named two ways depending on whether it is carried.** Real. `NotListShape` now composes `StripGetterInterface` over `StripOverlay`, so `BodyTemplateBinaryOverlay` and the declared `IBodyTemplateGetter` both read "a single BodyTemplate value". The suite could not see it because the in-memory fixtures hold concrete Mutagen classes, where both spellings already agree — the new test writes the fixture mod out and reads it back as a binary overlay, which is the only shape that splits them. Test: `ANonListStepIsNamedTheSameWhetherItIsCarriedOrNull`; it fails on the pre-fold tree with `IBodyTemplateGetter` in the note.

**A token that is not a quantifier got the "not built yet" answer.** Real. A third branch now fires for anything outside `[*]` / `[*count]` / `[*any]` / `[*all]` / `[*none]`, in the parser's own words — "'[*sum]' is not a quantifier — the tokens are …" — so `where=` and `project.fields` judge the same token the same way instead of pointing the caller at a capability that will never cover it. Test: `ANonQuantifierTokenInProjectFieldsIsNamedATypo_NotAnUnbuiltCapability`.

**A UTF-8 BOM added by accident.** Real, and confirmed against `main`: neither file had one, and no `.cs` file in the repo does. Stripped from `WriteEngine.cs` and `LoadOrderService.cs` (the phantom `using …;` line is gone from both diffs) and from the two new test files, which now match the repo's convention.

**A negated entry could not use the `@file` spelling.** Real — `ExpandListInput` decides on the first character, so `!@<path>` passed through untouched and the path reached `door.Parse` as a malformed FormID. Folded as support rather than a refusal, since `references=` already documents the spelling: `ExpandReferenceList` strips the sigil before the expander sees it and puts it back on every token the file expanded to, so "exclude everything in this artifact" is one call. A `!@` mixed with inline entries still gets the expander's own mixing refusal. Tool description and the changelog entry now say it. Test: `ANegatedReferenceTakesTheAtFileSpellingToo`.

## Refuted

None.

PR #552, opened 2026-09-05
Give both extend refusals candidates and a lane

### LINKED ISSUE #359: [Bug] the un-owned-folder refusal still dead-ends — it names no parameter and no lane

### LINKED ISSUE #380: [GAP] the extend refusal's always-true remedy is empty — nothing can list houseCARL-owned patches, so "Check the name." names no candidates
## Summary

When the `into=`-extend resolver refuses because the named patch does not exist, the always-true remedy is `"Check the name."` — and it carries no information. It names no candidate, quotes nothing back, and points at no tool, because **no tool lists houseCARL-owned patches** (measured on PR #379: the sentence was deliberately left bare rather than naming a tool that cannot answer).

## Why it matters

With the safe remedy empty, the only ACTIONABLE clause in the composed refusal is the consent-gated in-place lane that rewrites an original — and the likeliest cause of the refusal by a wide margin is a typo or an auto-suffixed name, exactly the caller who needs the candidate list. This is the substance behind PR #379's contested clause ordering (two rounds objected in opposite directions): order was never the variable — information is. If the refusal could carry the owned-patch names near the stem, the prominent clause would be the safe one on merit rather than by placement, and the ordering question mostly dissolves.

## Shape (design belongs to #359's engagement, not fixed here)

Two candidate halves, not mutually exclusive:
- **Refusal-side:** the resolver already enumerates houseCARL-owned folders (`OwnedFoldersHolding` / the canonical scan) — the not-found arm could quote near-stem candidates directly, the way the ambiguous arm already quotes its choices.
- **Surface-side:** a read that lists houseCARL-owned patches (natural home to be decided — `load_order_status` or an existing read's scope).

How much a refusal may enumerate, and whether a consent-gated destructive lane may be recommended **while the safe remedy is empty**, is #359's sharpened fork — this gap rides that engagement.

## Provenance

Named by the PR #379 review's third-reader observation (2026-08-19, inline comment on LoadOrderService.cs:8044); filed by the advisor lane on Aaron's go. Evidence deposit on #359 points here.

Routed (same stroke, Aaron-go, boot rule 4): **RUN_ORDER row 8/W5, beside #359 — same engagement** (its fix is the input that dissolves #359's ordering half). Milestoned `houseCARL 2.0`.

Avick3110:
Both arms of the `into=` extend resolver dead-ended. The not-found arm's always-true tail was "Check the name.", which named no candidate and pointed at no tool, because nothing lists the patches houseCARL owns; the un-owned-folder arm ended "Use a different patch name.", naming no parameter and no lane, on the path a caller reaches when their `into=` lands on an ordinary mod folder. Both arms now close with the owned patches rendered as `into=` spellings, each with the plugins it holds, nearest the name the caller typed first (ranked by the shared name-suggester), capped at eight with a "+N more" count, and with folders holding no plugin left out of the record lane, which could not extend them anyway. Every spelling printed is checked against this resolver's own arms before it is printed, so a row a caller copies resolves back to the patch it was printed for: a folder whose own name would resolve elsewhere, or one holding several plugins that the record lane would then refuse, is listed by its plugins instead, and a patch no spelling reaches is counted in the tail rather than offered a spelling that fails.

The un-owned arm additionally names the calling lane's own fresh-patch parameter (`patch=` on the record lanes, the rider's own on the rider lanes) and, where the calling tool declares one, the in-place lane, since reaching a foreign folder by name does not say which of the two readings was meant. The in-place clause is offered only where it would work: `in_place=` resolves a plugin FILENAME through the active load order, so a folder holding no plugin, or only ones MO2 has not enabled, gets no sentence claiming a lane it has — and where the lane does exist, the sentence names the plugin rather than a `<plugin filename>` placeholder. The in-place clause also reads LAST on that arm, behind the fresh-patch parameter and the candidate list: it is the one clause there that rewrites a file the caller did not author, and #380's argument was that the prominent clause should be the safe one. Removal's un-owned arm also states its own no-fresh-patch rule, which previously rode the not-found arm only. It deliberately does not hand the caller their own stem back: `IsStemFree` tests only for an owned folder and an *active* plugin of that name, so a patch minted under the colliding stem would sit beside an inactive foreign plugin of the same name — two plugins that cannot both be active. The in-place spelling is handed down from the tool, the way removal already did it, rather than picked by the service, so a lane that has no such parameter is never told to pass one.

**Scoped out, said plainly:** #359's second half — "may need `IsStemFree` to consider inactive plugins in foreign folders" — is NOT fixed here. `IsStemFree` still tests only for a `houseCARL - <stem>` folder and an active `<stem>.esp`, so a caller who follows the new remedy with a stem that collides with an inactive plugin in someone else's mod folder still mints a plugin that cannot be active beside it, with no refusal. The refusal avoids walking them into it (it does not hand the stem back), and the CHANGELOG now says outright that houseCARL does not check the name for that collision, but the check itself is a change to the fresh-write path rather than to a refusal's wording, with its own blast radius on every `patch=` call — it belongs in its own issue rather than riding this one.

Coverage went into `extend-resolve-guard`, which already owns this resolver's refusals end-to-end against a synthetic MO2 instance, rather than into a new xUnit class that would duplicate it (TESTING.md: no duplicate of a probe). Three new arms: 9b calls all four write tools against an un-owned folder holding an active plugin, so each tool's hand-down of the in-place sentence has its own arm and deleting any one of them reddens; 9c builds the two shapes that make a rendered candidate false and FOLLOWS the spellings the list prints; 9d covers the pluginless folder on both lanes and the cap. The list assertions pin the colon form, which the empty-inventory clause does not satisfy. `write-surface-guard` gained the un-owned apply refusal, on a foreign folder that now holds an active plugin so the sentence it observes is one the render really carries.

Closes #359
Closes #380

Tests 1160 passed, probes 128/128 passed, build clean.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -242,8 +242,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   folders holding no plugin, which it could not extend. Where houseCARL owns patches but no single `into=` spelling
   reaches any of them, the refusal says how many there are and to rename one in MO2, instead of saying it owns none;
   where a mod folder could not be read while the patches were being counted, no candidate is offered at all rather
-  than the half the scan reached. `housecarl_remove` cannot create a patch, so on an install owning none yet it now
-  names the writes that can make one instead of stopping at what went wrong.
+  than the half the scan reached, and the refusal says the scan failed rather than that houseCARL owns no patch — a
+  claim it could not have established. `housecarl_remove` cannot create a patch, so on an install owning none yet it
+  now names the writes that can make one instead of stopping at what went wrong. The fresh-patch remedy says to drop
+  `into=` as well as which parameter to pass: every lane takes the extend branch whenever `into=` is set, so merely
+  adding the parameter to the failed call returns the same refusal.
   Naming a mod folder houseCARL does not own is still refused with the originals untouched; that refusal does not
   hand back the name you passed, since a patch minted under that stem would sit beside the foreign plugin of the
   same name. houseCARL does not check the name you pass `patch=` next against inactive plugins in other people's mod
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -233,13 +233,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
   union above affordable on every read; the conflict tree takes the same path. Measured on a synthetic order of
   one 20k-record master and nine 2k plugins, ten touchers of one cell: 80 ms of body fetches became 4 ms.
 - **An `into=` that names no houseCARL patch now lists the patches houseCARL owns.** Both refusals close with the
-  owned patch folders as `into=` spellings, with the plugins each holds, so a typo or an auto-suffixed name is
-  answered with the candidates instead of "Check the name." The list is capped, and says so when it is; the record
-  lane leaves out folders holding no plugin, which it could not extend. Naming a mod folder houseCARL does **not**
-  own is still refused with the originals untouched, but that refusal now names the calling tool's own fresh-patch
-  parameter and, on `housecarl_apply` / `housecarl_create` / `housecarl_forward` / `housecarl_remove`, the
-  `in_place=` lane — the other reading of a name that landed on someone else's folder. It does not hand back the
-  name you passed: a patch minted under that stem would sit beside the foreign plugin of the same name.
+  owned patches as `into=` spellings, nearest what you typed first, and every spelling printed is one that resolves
+  back to the patch it is printed for — a folder holding two plugins is listed by each plugin, not by a folder name
+  the next call would refuse. The list is capped, and says so when it is; the record lane leaves out folders holding
+  no plugin, which it could not extend. Naming a mod folder houseCARL does **not** own is still refused with the
+  originals untouched, but that refusal now names the calling tool's own fresh-patch parameter and, where that
+  folder holds a plugin your load order actually has active, the `in_place=` lane for that plugin by name — the
+  other reading of a name that landed on someone else's folder. It does not hand back the name you passed: a patch
+  minted under that stem would sit beside the foreign plugin of the same name. houseCARL does not check the name you
+  pass `patch=` next against inactive plugins in other people's mod folders, so that collision is still yours to
+  avoid.
 
 - **`housecarl_load_order_status` now names the running server's build in its header** — a `server:` line carrying
   the informational version the binary embeds (the release version `build-plugin.ps1` stamps from `plugin.json`,
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -232,6 +232,15 @@ saying it sets an expectation their install may contradict. Say what is known, a
   plugin.** Finding one cell no longer steps over every placed reference in the plugin, which is what makes the
   union above affordable on every read; the conflict tree takes the same path. Measured on a synthetic order of
   one 20k-record master and nine 2k plugins, ten touchers of one cell: 80 ms of body fetches became 4 ms.
+- **An `into=` that names no houseCARL patch now lists the patches houseCARL owns.** Both refusals close with the
+  owned patch folders as `into=` spellings, with the plugins each holds, so a typo or an auto-suffixed name is
+  answered with the candidates instead of "Check the name." The list is capped, and says so when it is; the record
+  lane leaves out folders holding no plugin, which it could not extend. Naming a mod folder houseCARL does **not**
+  own is still refused with the originals untouched, but that refusal now names the calling tool's own fresh-patch
+  parameter and, on `housecarl_apply` / `housecarl_create` / `housecarl_forward` / `housecarl_remove`, the
+  `in_place=` lane — the other reading of a name that landed on someone else's folder. It does not hand back the
+  name you passed: a patch minted under that stem would sit beside the foreign plugin of the same name.
+
 - **`housecarl_load_order_status` now names the running server's build in its header** — a `server:` line carrying
   the informational version the binary embeds (the release version `build-plugin.ps1` stamps from `plugin.json`,
   then `+` and the full commit sha: `1.9.5-dev+e942910…`), so checking that the installed houseCARL is the build you expect no longer means
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -236,10 +236,15 @@ saying it sets an expectation their install may contradict. Say what is known, a
   owned patches as `into=` spellings, nearest what you typed first, and every spelling printed is one that resolves
   back to the patch it is printed for — a folder holding two plugins is listed by each plugin, not by a folder name
   the next call would refuse. The list is capped, and says so when it is; the record lane leaves out folders holding
-  no plugin, which it could not extend. Naming a mod folder houseCARL does **not** own is still refused with the
-  originals untouched, but that refusal now names the calling tool's own fresh-patch parameter and, where that
-  folder holds a plugin your load order actually has active, the `in_place=` lane for that plugin by name — the
-  other reading of a name that landed on someone else's folder. It does not hand back the name you passed: a patch
+  no plugin, which it could not extend. Where houseCARL owns patches but no single `into=` spelling reaches any of
+  them, the refusal says how many there are instead of saying it owns none. Naming a mod folder houseCARL does
+  **not** own is still refused with the originals untouched, but that refusal now names the calling tool's own
+  fresh-patch parameter and, where the plugin your load order resolves to that folder is one that folder holds, the
+  `in_place=` lane for that plugin by name — the other reading of a name that landed on someone else's folder. A
+  folder shipping a copy of a plugin some other mod wins gets no such sentence, since the lane would open the other
+  mod's file. That lane reads last on both refusals, behind the candidates you can take without touching anyone
+  else's plugin, and names a few plugins with a count rather than every plugin in a large folder. It does not hand
+  back the name you passed: a patch
   minted under that stem would sit beside the foreign plugin of the same name. houseCARL does not check the name you
   pass `patch=` next against inactive plugins in other people's mod folders, so that collision is still yours to
   avoid.
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -232,22 +232,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
   plugin.** Finding one cell no longer steps over every placed reference in the plugin, which is what makes the
   union above affordable on every read; the conflict tree takes the same path. Measured on a synthetic order of
   one 20k-record master and nine 2k plugins, ten touchers of one cell: 80 ms of body fetches became 4 ms.
-- **An `into=` that names no houseCARL patch now lists the patches houseCARL owns.** Both refusals close with the
-  owned patches as `into=` spellings, nearest what you typed first, and every spelling printed is one that resolves
-  back to the patch it is printed for — a folder holding two plugins is listed by each plugin, not by a folder name
-  the next call would refuse. The list is capped, and says so when it is; the record lane leaves out folders holding
-  no plugin, which it could not extend. Where houseCARL owns patches but no single `into=` spelling reaches any of
-  them, the refusal says how many there are instead of saying it owns none. Naming a mod folder houseCARL does
-  **not** own is still refused with the originals untouched, but that refusal now names the calling tool's own
-  fresh-patch parameter and, where the plugin your load order resolves to that folder is one that folder holds, the
-  `in_place=` lane for that plugin by name — the other reading of a name that landed on someone else's folder. A
-  folder shipping a copy of a plugin some other mod wins gets no such sentence, since the lane would open the other
-  mod's file. That lane reads last on both refusals, behind the candidates you can take without touching anyone
-  else's plugin, and names a few plugins with a count rather than every plugin in a large folder. It does not hand
-  back the name you passed: a patch
-  minted under that stem would sit beside the foreign plugin of the same name. houseCARL does not check the name you
-  pass `patch=` next against inactive plugins in other people's mod folders, so that collision is still yours to
-  avoid.
+- **An `into=` that names no houseCARL patch, or names a folder houseCARL does not own, now tells you what to try.**
+  Each refusal is one sentence: what went wrong, then the nearest houseCARL-owned patches to extend instead, named
+  as `into=` spellings (at most three, nearest what you typed first), and the calling tool's own fresh-patch
+  parameter. Every spelling offered is one that resolves back to the patch it stands for — a folder holding two
+  plugins is offered by a plugin, not by a folder name the next call would refuse — and the record lane leaves out
+  folders holding no plugin, which it could not extend. Where houseCARL owns patches but no single `into=` spelling
+  reaches any of them, the refusal says how many there are and to rename one in MO2, instead of saying it owns none.
+  Naming a mod folder houseCARL does not own is still refused with the originals untouched; that refusal does not
+  hand back the name you passed, since a patch minted under that stem would sit beside the foreign plugin of the
+  same name. houseCARL does not check the name you pass `patch=` next against inactive plugins in other people's mod
+  folders, so that collision is still yours to avoid.
 
 - **`housecarl_load_order_status` now names the running server's build in its header** — a `server:` line carrying
   the informational version the binary embeds (the release version `build-plugin.ps1` stamps from `plugin.json`,
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -234,11 +234,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
   one 20k-record master and nine 2k plugins, ten touchers of one cell: 80 ms of body fetches became 4 ms.
 - **An `into=` that names no houseCARL patch, or names a folder houseCARL does not own, now tells you what to try.**
   Each refusal is one sentence: what went wrong, then the nearest houseCARL-owned patches to extend instead, named
-  as `into=` spellings (at most three, nearest what you typed first), and the calling tool's own fresh-patch
-  parameter. Every spelling offered is one that resolves back to the patch it stands for — a folder holding two
+  as `into=` spellings (at most three, nearest what you typed first, with a count of the ones the cap left out so the
+  three do not read as the whole inventory), and the calling tool's own fresh-patch parameter. Nearest is measured
+  even where the name-suggester vouches for nothing, so the three are never whichever names happen to sort first.
+  Every spelling offered is one that resolves back to the patch it stands for — a folder holding two
   plugins is offered by a plugin, not by a folder name the next call would refuse — and the record lane leaves out
   folders holding no plugin, which it could not extend. Where houseCARL owns patches but no single `into=` spelling
-  reaches any of them, the refusal says how many there are and to rename one in MO2, instead of saying it owns none.
+  reaches any of them, the refusal says how many there are and to rename one in MO2, instead of saying it owns none;
+  where a mod folder could not be read while the patches were being counted, no candidate is offered at all rather
+  than the half the scan reached. `housecarl_remove` cannot create a patch, so on an install owning none yet it now
+  names the writes that can make one instead of stopping at what went wrong.
   Naming a mod folder houseCARL does not own is still refused with the originals untouched; that refusal does not
   hand back the name you passed, since a patch minted under that stem would sit beside the foreign plugin of the
   same name. houseCARL does not check the name you pass `patch=` next against inactive plugins in other people's mod
--- a/src/housecarl-generator/ExtendResolveProbe.cs
+++ b/src/housecarl-generator/ExtendResolveProbe.cs
@@ -325,16 +325,20 @@ internal static class ExtendResolveProbe
             //      unreachable for them — but they pass null naming rather than borrowing a sibling's sentence, and
             //      null must degrade to the weakest true remedy rather than to a wrong one.
             Console.WriteLine();
-            Console.WriteLine("--- 8c2: a rider with no naming falls back to 'Check the name.' ---");
+            Console.WriteLine("--- 8c2: a rider with no naming falls back to the owned-patch list ---");
             {
                 string bareErr = "";
                 try { svc.ResolvePatchModFolder(null, "GhostRider", "houseCARL_Extract", naming: null); }
                 catch (InvalidOperationException ex) { bareErr = ex.Message; }
                 Check(bareErr.Contains("GhostRider.esp", StringComparison.Ordinal)
-                      && bareErr.Contains("Check the name.", StringComparison.Ordinal)
                       && !bareErr.Contains("patch_name=", StringComparison.Ordinal)
                       && !bareErr.Contains("patch=", StringComparison.Ordinal),
                       "a null naming refuses without offering a parameter it was never told about");
+                // #380: the always-true tail used to be "Check the name.", which named no candidate at all. It is now
+                // the owned-patch list — the one thing true for every caller that also tells them something.
+                Check(bareErr.Contains("houseCARL owns:", StringComparison.Ordinal)
+                      && bareErr.Contains("into=\"", StringComparison.Ordinal),
+                      "…and closes with the patches houseCARL owns, as into= spellings (#380)");
             }
 
             // ---- 8d: the REMOVAL lane must not get the naming sentence either ----
@@ -360,8 +364,8 @@ internal static class ExtendResolveProbe
                 // a lane that starts stating CreatedByOmittingInto, or a default that starts assuming one, reddens here.
                 Check(r.Error is not null && !r.Error.Contains("create it fresh", StringComparison.OrdinalIgnoreCase),
                       "…and no longer offers 'create it fresh' — the remedy this lane could not perform (#356)");
-                Check(r.Error is not null && r.Error.Contains(WriteSentences.ExtendCheckTheName, StringComparison.Ordinal),
-                      "…it falls back to the always-true default tail (check what you typed)");
+                Check(r.Error is not null && r.Error.Contains("houseCARL owns", StringComparison.Ordinal),
+                      "…it falls back to the always-true default tail, which is now the owned-patch list (#380)");
                 Check(r.Error is not null && r.Error.Contains(WriteSentences.RemoveNoFreshPatch, StringComparison.Ordinal),
                       "…and the LANE states why there is no create route here (removal needs a patch that already carries it)");
                 // A direct SERVICE call names no in-place spelling at all — the two tools spell that lane
@@ -489,11 +493,32 @@ internal static class ExtendResolveProbe
                 Check(!r.Success && refused, "into=\"Foreign\" (un-owned folder) is REFUSED — houseCARL won't edit a folder it doesn't own");
                 Check(File.ReadAllBytes(foreignEsp).SequenceEqual(foreignBefore), "the un-owned plugin is byte-untouched after the refusal");
 
+                // #359: the refusal must not dead-end. It names the fresh lane's own parameter, the owned patches to
+                // extend instead, and — at the TOOL altitude, where the spelling is known — the in-place lane, the
+                // other reading of a name that landed on someone else's folder.
+                Check(r.Error is not null && r.Error.Contains("pass patch= a name no mod folder already uses", StringComparison.Ordinal),
+                      "…the un-owned refusal names the fresh lane's parameter (#359)");
+                Check(r.Error is not null && !r.Error.Contains("patch=\"Foreign\"", StringComparison.Ordinal),
+                      "…and does NOT hand back the colliding stem, which would shadow the foreign plugin (#359)");
+                Check(r.Error is not null && r.Error.Contains("houseCARL owns", StringComparison.Ordinal),
+                      "…and lists the patches houseCARL owns instead of 'Use a different patch name.'");
+                Check(r.Error is not null && !r.Error.Contains("in_place", StringComparison.Ordinal),
+                      "…and at the SERVICE altitude names no in-place spelling (the tools own that sentence)");
+
+                var tooled = RemoveTools.Remove(svc, new[] { fid }, into: "Foreign");
+                Check(tooled.Contains(WriteSentences.RemoveInPlaceLane, StringComparison.Ordinal)
+                      && tooled.Contains("houseCARL owns", StringComparison.Ordinal),
+                      "housecarl_remove's un-owned refusal names the in-place lane IT declares, beside the owned patches (#359)");
+
                 // the rider lane refuses the un-owned folder too (shared resolver — same gate)
-                bool riderRefused = false;
+                string riderErr2 = "";
                 try { svc.ResolvePatchModFolder(null, "Foreign", "houseCARL_Archive", BsaTools.RepackNaming); }
-                catch (InvalidOperationException ex) { riderRefused = ex.Message.Contains("NOT created by houseCARL", StringComparison.Ordinal); }
-                Check(riderRefused, "the RIDER lane also refuses the un-owned folder (same ownership gate, no foreign-plugin door)");
+                catch (InvalidOperationException ex) { riderErr2 = ex.Message; }
+                Check(riderErr2.Contains("NOT created by houseCARL", StringComparison.Ordinal),
+                      "the RIDER lane also refuses the un-owned folder (same ownership gate, no foreign-plugin door)");
+                Check(riderErr2.Contains("pass patch_name= a name no mod folder already uses", StringComparison.Ordinal)
+                      && riderErr2.Contains("houseCARL owns", StringComparison.Ordinal),
+                      "…and names THAT lane's own parameter plus the owned patches, never patch= or a dead end (#359)");
             }
 
             // ---- 10: ORIGINALS — the master plugin never moved a byte (extends only wrote the patch folder) ----
--- a/src/housecarl-generator/ExtendResolveProbe.cs
+++ b/src/housecarl-generator/ExtendResolveProbe.cs
@@ -398,6 +398,12 @@ internal static class ExtendResolveProbe
                       "housecarl_remove names in_place=\"<plugin filename>\" — the lane it actually declares");
                 Check(!modern.Contains("target=", StringComparison.Ordinal),
                       "…and never names target=, which that tool does not declare");
+                // Both arms of this resolver read the same way about emphasis: the consent-gated lane comes after
+                // the candidates the caller can take without touching anyone else's file (#380). The not-found arm
+                // used to carry it inside the lane clause, ahead of the list.
+                int ownsFirst = modern.IndexOf("houseCARL owns", StringComparison.Ordinal);
+                Check(ownsFirst >= 0 && ownsFirst < modern.IndexOf("in_place=", StringComparison.Ordinal),
+                      "…and the NOT-FOUND arm reads the candidates before the in-place lane, like the un-owned one");
 
                 // The 2.0 tool refuses first, in its own spelling. Kept as a statement about the TOOL — it is
                 // no longer what makes anything in the service correct.
@@ -544,8 +550,10 @@ internal static class ExtendResolveProbe
                 Check(ap.Contains(WriteSentences.ExtendInPlaceLane, StringComparison.Ordinal) && ap.Contains(named, StringComparison.Ordinal),
                       "housecarl_apply names the in-place lane AND the plugin in that folder the lane can take");
                 // The one clause here that rewrites a file the caller did not author reads LAST, behind both lanes
-                // that leave every original untouched (#380).
-                Check(ap.IndexOf("houseCARL owns:", StringComparison.Ordinal) < ap.IndexOf("in_place=", StringComparison.Ordinal),
+                // that leave every original untouched (#380). The list has to be PRESENT for the order to mean
+                // anything — IndexOf returns -1 for a missing substring, which would satisfy any "before" test.
+                int ownsAt = ap.IndexOf("houseCARL owns:", StringComparison.Ordinal);
+                Check(ownsAt >= 0 && ownsAt < ap.IndexOf("in_place=", StringComparison.Ordinal),
                       "…and the safe candidate list is read BEFORE the consent-gated lane");
 
                 var cr = CreateTools.Create(svc, records: Doc("[{\"record_type\":\"Keyword\",\"editorid\":\"HcExtUnowned\"}]"),
@@ -642,6 +650,88 @@ internal static class ExtendResolveProbe
                 int rowCount = capText.Split("into=\"", StringSplitOptions.None).Length - 1;
                 Check(rowCount == 8 && capText.Contains(" more.", StringComparison.Ordinal),
                       $"the list stops at 8 rows and counts the rest ({rowCount} rows rendered)");
+            }
+
+            // ---- 9e: the in-place lane is offered only for the copy the game actually LOADS ----
+            //      Two mod folders can ship the same plugin basename and only one wins. Offering the lane on a
+            //      filename match alone points a consent-gated write at the OTHER folder's file — a plugin the
+            //      sentence never named. The twin here is not enabled, so RemoveHereMod's copy stays the winner.
+            Console.WriteLine();
+            Console.WriteLine("--- 9e: a folder holding a losing copy of an active plugin gets NO in-place lane ---");
+            {
+                var twin = Path.Combine(mods, "TwinRemoveMod");                // un-owned, and NOT in modlist.txt
+                Directory.CreateDirectory(twin);
+                File.Copy(Path.Combine(mods, "RemoveHereMod", "RemoveHere.esp"), Path.Combine(twin, "RemoveHere.esp"));
+
+                var twinned = RemoveTools.Remove(svc, new[] { fid }, into: "TwinRemoveMod");
+                Check(twinned.Contains("NOT created by houseCARL", StringComparison.Ordinal)
+                      && !twinned.Contains("in_place=", StringComparison.Ordinal),
+                      "the un-owned refusal offers no in-place lane for a copy the load order does not resolve here");
+                Check(twinned.Contains("houseCARL owns", StringComparison.Ordinal),
+                      "…and still closes with the owned patches rather than dead-ending");
+
+                // The folder that DOES win keeps its sentence, so the gate is a path check and not a blanket refusal.
+                var winner = RemoveTools.Remove(svc, new[] { fid }, into: "RemoveHereMod");
+                Check(winner.Contains("in_place=\"RemoveHere.esp\"", StringComparison.Ordinal),
+                      "…while the folder whose copy the game loads still names the lane, with the plugin");
+            }
+
+            // ---- 9f: owned patches that no single into= spelling reaches are COUNTED, not denied ----
+            //      rows.Count == 0 has two causes and only one of them is "there is nothing to extend". Its own
+            //      instance, because the claim is about the WHOLE inventory being unreachable.
+            Console.WriteLine();
+            Console.WriteLine("--- 9f: an inventory nothing reaches says so, instead of claiming none exists ---");
+            {
+                string inst2 = Path.Combine(root, "instance2");
+                string prof2 = Path.Combine(inst2, "profiles", "Default");
+                string mods2 = Path.Combine(inst2, "mods");
+                Directory.CreateDirectory(prof2); Directory.CreateDirectory(mods2);
+                File.WriteAllText(Path.Combine(inst2, "ModOrganizer.ini"),
+                    "[General]\r\ngameName=Skyrim Special Edition\r\nselected_profile=@ByteArray(Default)\r\ngamePath=@ByteArray("
+                    + Path.Combine(root, "game").Replace(@"\", @"\\") + ")\r\n");
+                Directory.CreateDirectory(Path.Combine(mods2, "MasterMod"));
+                File.Copy(masterPath, Path.Combine(mods2, "MasterMod", mKey.FileName.String));
+                File.WriteAllText(Path.Combine(prof2, "loadorder.txt"), "# header\r\n" + mKey.FileName + "\r\n");
+                File.WriteAllText(Path.Combine(prof2, "plugins.txt"), "*" + mKey.FileName + "\r\n");
+                File.WriteAllText(Path.Combine(prof2, "modlist.txt"), "# header\r\n+MasterMod\r\n");
+
+                // Two owned twins, each holding the SAME two plugins: no folder name resolves (neither holds
+                // "<folder>.esp"), and each plugin name is ambiguous across the pair. Every spelling is refused.
+                foreach (var name in new[] { "houseCARL - Twin", "houseCARL - Twin backup" })
+                {
+                    var d = Path.Combine(mods2, name);
+                    Directory.CreateDirectory(d);
+                    File.WriteAllText(Path.Combine(d, "meta.ini"),
+                        $"{HousecarlOwnerMeta.Section}\r\ngenerated=true\r\nplugin=Alpha.esp\r\n");
+                    File.WriteAllText(Path.Combine(d, "Alpha.esp"), "a");
+                    File.WriteAllText(Path.Combine(d, "Beta.esp"), "b");
+                }
+
+                var store2 = new UserConfigStore(Path.Combine(root, "houseCARL.user2.json"));
+                using var svc2 = LoadOrderService.WithInstance(inst2, 0, store2);
+                svc2.Stats();
+                var unreachable = svc2.ApplyEdits(new[] { Wgt(1) }, null, "GhostNowhere");
+                string text = unreachable.Error ?? "";
+                Check(text.Contains("houseCARL owns 2 patches, none reachable", StringComparison.Ordinal),
+                      $"the refusal counts the patches no spelling reaches ({text})");
+                Check(!text.Contains("owns no patch", StringComparison.Ordinal),
+                      "…rather than telling the caller houseCARL owns none, which would send them to mint a duplicate");
+            }
+
+            // ---- 9g: the in-place lane sentence is capped like the list beside it ----
+            //      A compilation folder ships dozens of active plugins; the sibling clause caps at 8 rows for the
+            //      same reason. Rendered directly — building 30 active plugins to see one sentence is not the claim.
+            Console.WriteLine();
+            Console.WriteLine("--- 9g: the in-place lane names a few plugins and counts the rest ---");
+            {
+                var many = Enumerable.Range(0, 25).Select(i => $"Pack{i:D2}.esp").ToList();
+                var sentence = WriteSentences.InPlaceLane(WriteSentences.ExtendInPlaceLane, many);
+                int quoted = sentence.Split('"').Length / 2;
+                Check(quoted == 4 && sentence.Contains("or one of 21 others in that folder", StringComparison.Ordinal),
+                      $"25 active plugins render 4 names plus a count, not 25 filenames ({sentence})");
+                var few = WriteSentences.InPlaceLane(WriteSentences.ExtendInPlaceLane, new[] { "OnlyOne.esp" });
+                Check(few.Contains("\"OnlyOne.esp\".", StringComparison.Ordinal) && !few.Contains("others", StringComparison.Ordinal),
+                      "…and a folder under the cap still names every plugin, with no tail");
             }
 
             // ---- 10: ORIGINALS — the master plugin never moved a byte (extends only wrote the patch folder) ----
--- a/src/housecarl-generator/ExtendResolveProbe.cs
+++ b/src/housecarl-generator/ExtendResolveProbe.cs
@@ -1,3 +1,4 @@
+using System.Text.Json;
 using Mutagen.Bethesda;
 using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Plugins.Records;
@@ -199,8 +200,11 @@ internal static class ExtendResolveProbe
                 string twoEsp = Path.Combine(mods, "houseCARL - TwoEsp");
                 Directory.CreateDirectory(twoEsp);
                 MarkOwned(twoEsp, "Alpha.esp");
-                File.WriteAllText(Path.Combine(twoEsp, "Alpha.esp"), "x");
-                File.WriteAllText(Path.Combine(twoEsp, "Beta.esp"), "y");
+                // Real plugins, not placeholder bytes: arm 9c FOLLOWS the spelling this folder is rendered as, and a
+                // remedy is checked by making the call.
+                foreach (var n in new[] { "Alpha", "Beta" })
+                    new SkyrimMod(new ModKey(n, ModType.Plugin), SkyrimRelease.SkyrimSE)
+                        .BeginWrite.ToPath(Path.Combine(twoEsp, n + ".esp")).WithLoadOrder(Array.Empty<ISkyrimModGetter>()).Write();
 
                 var r = svc.ApplyEdits(new[] { Wgt(1) }, null, "TwoEsp");
                 bool named = r.Error is not null
@@ -505,10 +509,13 @@ internal static class ExtendResolveProbe
                 Check(r.Error is not null && !r.Error.Contains("in_place", StringComparison.Ordinal),
                       "…and at the SERVICE altitude names no in-place spelling (the tools own that sentence)");
 
+                // This folder's plugin is NOT in the active order, so the in-place lane cannot take it — in_place=
+                // resolves a filename through the load order. The sentence is withheld rather than offered false,
+                // which is the whole point of gating it on what the folder actually holds.
                 var tooled = RemoveTools.Remove(svc, new[] { fid }, into: "Foreign");
-                Check(tooled.Contains(WriteSentences.RemoveInPlaceLane, StringComparison.Ordinal)
+                Check(!tooled.Contains("in_place=", StringComparison.Ordinal)
                       && tooled.Contains("houseCARL owns", StringComparison.Ordinal),
-                      "housecarl_remove's un-owned refusal names the in-place lane IT declares, beside the owned patches (#359)");
+                      "housecarl_remove's un-owned refusal offers NO in-place lane for a folder with no active plugin, and still lists the owned patches");
 
                 // the rider lane refuses the un-owned folder too (shared resolver — same gate)
                 string riderErr2 = "";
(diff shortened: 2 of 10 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8166,9 +8176,26 @@ string ResolveOwnedPatchFolder(string into, bool needEsp,
...
+                        ? $"Omit into= and pass {fr.Param}= a name no mod folder already uses to create a fresh one. "
+                          + (fr.Caveat is null ? "" : fr.Caveat + " ")
+                        : freshPatch switch
+                        {
+                            FreshPatchRemedy.NamedByPatchParam =>
+                                "Omit into= and pass patch= a name no mod folder already uses to create a fresh patch. ",
+                            FreshPatchRemedy.CreatedByOmittingInto => "Omit into= to create a fresh patch. ",
+                            _ => "",
+                        }) +
+                    (inPlaceLane is null ? "" : inPlaceLane + " ") +
```
Avick3110: The in-place clause is appended unconditionally, without checking that the foreign folder holds a plugin the in-place lane could actually take.

`in_place=` needs a plugin filename that is in the ACTIVE load order (`housecarl_remove`'s own description: "the opt-in lane for ANY existing **active** plugin"). Two ordinary shapes of `cand` cannot satisfy that:

- the folder holds no plugin at all — a texture/mesh-only mod, or a SkyPatcher/SPID INI mod. `housecarl_apply(into="SomeTextureMod")` now answers "To write into an existing plugin IN PLACE instead, pass in_place=\"<plugin filename>\"" for a folder with no plugin in it. The new `write-surface-guard` fixture is exactly this case: `W2TwinForeign` is a bare `Directory.CreateDirectory` with nothing in it, so the render this PR adds to prove the sentence reaches a lane is a render of the sentence being false.
- the folder is not enabled in MO2 — which is the case the PR body itself builds the fresh-lane wording around ("a patch minted under the colliding stem would sit beside an *inactive* foreign plugin of the same name"). An inactive plugin is not reachable by `in_place=` either, so the caller gets a second refusal.

This is the failure mode the not-found arm's own comment forty lines below warns against: "telling such a caller to omit the lane sends them into a second refusal."

Gating the clause on the folder actually holding a plugin — and naming it, rather than `<plugin filename>` — would make the sentence true when it is offered and absent when it is not.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8226,6 +8255,42 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+            foreach (var dir in Directory.EnumerateDirectories(_modsDir).OrderBy(d => d, StringComparer.OrdinalIgnoreCase))
+            {
+                if (!IsHouseCarlOwned(dir)) continue;
+                var plugins = Directory.EnumerateFiles(dir)
+                    .Where(f => PluginExts.Any(ext => f.EndsWith(ext, StringComparison.OrdinalIgnoreCase)))
+                    .Select(Path.GetFileName).OrderBy(n => n, StringComparer.OrdinalIgnoreCase).ToList();
+                if (needEsp && plugins.Count == 0) continue;
+                if (rows.Count == cap) { more++; continue; }
+                var name = Path.GetFileName(dir);
+                rows.Add(plugins.Count == 0
```
Avick3110: A folder holding more than one plugin is rendered as a single `into="<folder>"` spelling, which the record lane then refuses.

Follow that candidate on the record lane and `ResolveOutputPath` resolves the folder, finds no `<folder-stem>.esp`, and hands it to `SoleEspInFolder`, which returns null with `holds N plugins (...) — name the one to extend by passing its filename as into=`. So a caller who takes the offered spelling literally lands in a second refusal — the exact dead-end this PR set out to remove. Concrete: a merge or a hand-renamed patch folder carrying `A.esp` and `B.esp` renders `into="houseCARL - Merge" (A.esp, B.esp)`, and `into="houseCARL - Merge"` is refused.

The plugin names are already in hand at this point. Emitting them as the tokens when there is more than one — `into="A"  |  into="B"` (each resolves through the by-plugin arm) — and keeping the folder spelling only for the single-plugin and pluginless (`needEsp:false`) cases would make every row a spelling that actually resolves.

src/housecarl-generator/ExtendResolveProbe.cs
```diff
@@ -489,11 +493,32 @@ void MarkOwned(string folder, string plugin) =>
...
                 Check(File.ReadAllBytes(foreignEsp).SequenceEqual(foreignBefore), "the un-owned plugin is byte-untouched after the refusal");
 
+                // #359: the refusal must not dead-end. It names the fresh lane's own parameter, the owned patches to
+                // extend instead, and — at the TOOL altitude, where the spelling is known — the in-place lane, the
+                // other reading of a name that landed on someone else's folder.
+                Check(r.Error is not null && r.Error.Contains("pass patch= a name no mod folder already uses", StringComparison.Ordinal),
+                      "…the un-owned refusal names the fresh lane's parameter (#359)");
+                Check(r.Error is not null && !r.Error.Contains("patch=\"Foreign\"", StringComparison.Ordinal),
+                      "…and does NOT hand back the colliding stem, which would shadow the foreign plugin (#359)");
+                Check(r.Error is not null && r.Error.Contains("houseCARL owns", StringComparison.Ordinal),
```
Avick3110: `Contains("houseCARL owns")` is satisfied by the EMPTY-list clauses too, so this arm and its three siblings pass whether or not any candidate is rendered.

`OwnedPatchCandidates` returns `"houseCARL owns no patch holding a plugin yet."` / `"houseCARL owns no patch folder yet."` when `rows` is empty — both contain the asserted substring. So a regression that makes the enumeration yield nothing (the ownership filter inverted, `needEsp` dropping everything, the `catch (IOException) {}` swallowing a real failure on a live ModsDir) leaves all four of these green, and the refusal is back to the dead end #380 is about. Same for lines 510 and 520 here, and line 367.

Only the 8c2 arm (`"houseCARL owns:"` with the colon, plus `into=\"`) actually pins the list form. Asserting the colon form here would close it.

Related gap: the changelog claims the record lane "leaves out folders holding no plugin", but nothing exercises `needEsp:true` against an owned pluginless folder — the only arm that reads the list form is the rider (`needEsp:false`) one.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8226,6 +8255,42 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+        try
+        {
+            foreach (var dir in Directory.EnumerateDirectories(_modsDir).OrderBy(d => d, StringComparer.OrdinalIgnoreCase))
+            {
+                if (!IsHouseCarlOwned(dir)) continue;
+                var plugins = Directory.EnumerateFiles(dir)
+                    .Where(f => PluginExts.Any(ext => f.EndsWith(ext, StringComparison.OrdinalIgnoreCase)))
+                    .Select(Path.GetFileName).OrderBy(n => n, StringComparer.OrdinalIgnoreCase).ToList();
+                if (needEsp && plugins.Count == 0) continue;
+                if (rows.Count == cap) { more++; continue; }
```
Avick3110: The cap is checked AFTER the per-folder file enumeration, so it caps the sentence but not the work.

On a real MO2 instance this walk already reads `meta.ini` for every mod folder (`IsHouseCarlOwned`) — thousands of them on a big list — and now also materialises a sorted plugin list for every owned folder, including the ninth onward that only bump `more`. A refusal is rare enough that this is not urgent, but the ninth-onward rows only need to know whether the folder holds a plugin, not which ones: hoisting `if (rows.Count == cap)` above the enumeration and using `Directory.EnumerateFiles(dir).Any(...)` on that branch keeps the `needEsp` filter exact at a fraction of the IO.

Avick3110:
Second review pass (fresh reader, blind to how this was built). These are **in addition** to the four inline findings already posted — no overlap with them.

### 1. medium — `src/housecarl-mcp/LoadOrderService.cs:8272` — the candidate list is alphabetical-first-8, not near-stem, so #380's acceptance is only half met

#380 asks for **"near-stem candidates"** ("the not-found arm could quote near-stem candidates directly, the way the ambiguous arm already quotes its choices"). `OwnedPatchCandidates` orders by folder name and keeps the first eight, with no reference to the stem the caller actually typed. On a working MO2 instance with more than eight houseCARL patches — the normal state after a few sessions — the row the caller needs is exactly as likely to be behind `+N more` as in front of it, and a caller who typo'd `MyPatch2` gets eight `A…`-`C…` folders. That is the same dead end #380 describes, one step politer. Sorting the rows by distance from `stem` (or at minimum floating the rows whose folder or plugin name contains the stem) is what makes the list answer the question the refusal was raised by.

### 2. medium — `src/housecarl-mcp/LoadOrderService.cs:8259-8261` — the rendered `into=` token is not guaranteed to reach the folder it was rendered from

The docstring claims the folder's own name "resolves through the folder catch-all whether or not the folder was renamed". The catch-all (`ResolveOwnedFolderByName`) is the **fourth** arm of `ResolveOwnedPatchFolder`; two earlier arms can win first:

- canonical: an owned `houseCARL - Foo` holding `Foo.esp`;
- by-plugin: any owned folder holding `Foo.esp`.

So for a renamed owned folder literally named `Foo` holding `Bar.esp`, `into="Foo"` resolves to whichever *other* owned folder holds `Foo.esp`, silently extending a different patch than the row the caller picked. Both rows can be printed in the same sentence, and one of the two tokens is then unreachable. This is a silently-wrong answer rather than a refusal, which is the cornerstone this list was added to serve. Either render the plugin filename as the token where it is unambiguous, or state the claim the resolver actually makes.

### 3. medium — `src/housecarl-mcp/CreateTools.cs:176` and `src/housecarl-mcp/ForwardTools.cs:121` — the two new hand-downs are proved by nothing

`ApplyTools`' hand-down is held up by `write-surface-guard`'s outer-sentence coverage (the only render observing `ExtendInPlaceLane`), and removal's by `extend-resolve-guard` arm 9. Create and forward have **no** un-owned refusal at tool altitude anywhere: arm 8e reaches both, but only on the *not-found* arm. Delete `inPlaceRemedy:` from either of these two call sites and the whole suite stays green. Arm 8e exists precisely for this shape ("a deleted statement is a silently weakened refusal") — the un-owned arm needs its own 8e.

### 4. medium — `src/housecarl-mcp/LoadOrderService.cs:8193-8198` — the consent-gated lane is still rendered ahead of the safe remedy

Composition order on the un-owned arm is: fresh clause, then `in_place=`, then the owned-patch list. #380's whole argument was that with the safe remedy empty the only actionable clause was the destructive one, and "if the refusal could carry the owned-patch names near the stem, the prominent clause would be the safe one **on merit**". Filling the list but leaving it last, behind the in-place sentence, keeps the destructive lane in front in reading order — the fix lands the information but not the consequence the issue was filed for.

### 5. low-medium — #359's second half is neither fixed nor said

#359 names two design calls: name the in-place lane, and "may need `IsStemFree` to consider inactive plugins in foreign folders". The PR takes the first and side-steps the second by not handing the stem back, and the body explains why — but `IsStemFree` is still blind to an inactive plugin in a foreign folder, and the new remedy delegates the check to the caller: *"pass patch= a name no mod folder already uses"*. houseCARL does not enforce that. A caller who follows the sentence with a colliding name still mints a plugin that cannot be active beside the foreign one, with no refusal. Either fix the check or say loudly in the refusal that houseCARL cannot verify it — the CHANGELOG's "It does not hand back the name you passed" reads as though the hazard is closed.

### 6. low — `src/housecarl-mcp/LoadOrderService.cs:5675` — removal's un-owned refusal drops `RemoveNoFreshPatch`

`laneClause` rides the not-found arm only, so `housecarl_remove`'s un-owned refusal now says nothing about why there is no fresh-patch route; it offers `in_place=` and the candidate list. A caller who reads "pass a name no mod folder uses" nowhere and reaches for `patch=` anyway gets a second refusal. Correct in what it states, thin in what it omits, on the one lane where the omission is load-bearing.

### 7. design — "errors are one plain sentence" has drifted

CLAUDE.md's third cornerstone. The un-owned refusal is now four sentences plus a pipe-separated inventory of up to eight rows each carrying its folder name and every plugin in it — several hundred characters, three different next steps. That may still be the right trade for #380, but it is a cornerstone being spent, and nothing in the PR body says so. Worth Aaron's ruling rather than an author's judgement call.

**Clean:** no per-record-type hand coverage; no new domain in `LoadOrderService.cs` (the new helper is the same resolver's rendering, not a subsystem); `ExtendCheckTheName` fully removed with no dangling reference; the closure-copy and NPC-copy call sites correctly pass no `inPlaceLane` (no such parameter on those tools); `nif_set` correctly gets none too — its `in_place` is a bool, and the shared sentence spells a filename.

Avick3110:
Review fold. Eight folded, one refuted, two for your call. Build clean, tests 1160 passed, probes 128/128.

## Folded

**Inline 1 — the in-place clause was rendered unconditionally.** Real, and the fixture proved your point twice over. `in_place=` resolves a plugin FILENAME through the ACTIVE load order, so the clause is now gated on the foreign folder holding a plugin that is actually active, and it NAMES that plugin instead of `<plugin filename>`. A folder holding no plugin, or only ones MO2 has not enabled, gets no in-place sentence at all. `WriteSentences.ExtendInPlaceLane` / `RemoveInPlaceLane` are now the lane prefix, completed either with the real filename or (where none is known — removal's `patch is required` arm and its not-found arm) with `InPlaceAnyFilename`, so the outer-sentence coverage arm still observes them literally. `write-surface-guard`'s `W2TwinForeign` now holds an active plugin, and `extend-resolve-guard` arm 9 asserts the OPPOSITE case explicitly: `houseCARL - Foreign` holds an inactive plugin, so the refusal must offer no `in_place=` at all.

**Inline 2 + inline 4 — the rendered token could resolve elsewhere, or be refused.** Both real, and they are the same bug: the folder name was printed unconditionally. Every candidate is now run through this resolver's own three arms (canonical, by-plugin, folder catch-all) against the folders read for the list, plus the record lane's own "does this land on ONE plugin" test, and only spellings that resolve back to the row they are printed for are emitted. Where the folder name does not reach it — another owned folder holds `<folder>.esp`, or the folder holds several plugins — the row is printed by its plugins instead (`into="Alpha.esp" (in houseCARL - TwoEsp)`). A patch no spelling reaches is counted in the tail rather than printed with a spelling that fails. Arm 9c builds both shapes and FOLLOWS the printed spellings; before the fix it renders `into="SeedA"` for a folder that resolves to a different patch.

**Inline 3 (summary 6) — removal's un-owned arm dropped `RemoveNoFreshPatch`.** Real. The resolver gained `unownedFresh`, riding the un-owned arm the way `laneClause` rides the not-found one, and removal hands it its no-create rule. Its in-place half moved out of `laneClause` into the named clause above, so the sentence is not said twice. Pinned in arm 9b.

**Summary 1 — alphabetical-first-8, not near-stem.** Real and #380 asks for it. Rows are ranked by `PluginNameSuggest.Nearest` over both the folder names and the plugin names — the same suggester the missed-plugin lookups use, rather than a second hand-rolled distance — and anything it does not rank keeps its alphabetical place behind them. Arm 9c pins it with a typo'd stem.

**Summary 3 (finding 10) — create and forward hand-downs proved by nothing.** Correct, and it was exactly the shape 8e exists for. Arm 9b is the un-owned arm's 8e: all four write tools called at TOOL altitude against an un-owned folder holding an active plugin, so deleting `inPlaceRemedy:` from any one call site reddens only that tool's check.

**Summary 4 — the consent-gated lane led the safe remedy.** Real, and done. Composition on the un-owned arm is now the fresh-lane clause, then the owned-patch candidates, then `in_place=` last — the one clause that rewrites a file the caller did not author reads behind both lanes that leave every original untouched. It is also absent far more often now (inline 1's gate), which does as much for #380's argument as the order does. Arm 9b pins the order, so a later edit that floats it back reddens. The NOT-FOUND arm keeps its existing composition: its in-place sentence arrives inside removal's own `laneClause`, whose position #379 already litigated, and moving it would hand the sentence to three tools that do not render it there today.

**Inline 3 on the probe (summary 11) — `Contains("houseCARL owns")` passes on the empty clause.** Correct, all four arms were vacuous. The list-form arms now pin `"houseCARL owns: "` (with the colon) plus `into="`. Arm 9d adds the two unexercised halves: an owned pluginless folder is absent from the record lane's list and present in the rider's, and nine surplus owned folders prove the cap holds at 8 rows with the rest counted.

## Refuted

**Inline 4 — the cap is applied after the IO.** Real as written against the old code, but no longer applicable, and I'd rather say so than fake the fix. Deciding whether a candidate spelling RESOLVES needs to know which owned folders hold a given plugin — that is the by-plugin arm's uniqueness test — so the plugin list is required for every owned folder, including the ones past the cap that only bump the count. Hoisting the cap above the enumeration would put the correctness fix for inline 2 back. What is left of the cost is small in the right place: the unavoidable expense on a big list is the ownership scan (`IsHouseCarlOwned` reads `meta.ini` for every mod folder, thousands of them), and `EnumerateFiles` runs only on the owned ones — houseCARL's own patches, a handful to a few dozen. Happy to reverse this if you'd rather have the cheaper walk and a weaker token guarantee.

## For your call

**Summary 5 (#359's second half, `IsStemFree`).** Not fixed, and now stated plainly in the PR body as a scoped-out half with the reason: `IsStemFree` still cannot see an inactive plugin in a foreign folder, so a caller who follows the remedy with a colliding stem still mints a plugin that cannot be active beside it. The refusal does not walk them into it (it never hands the stem back), and the CHANGELOG now says outright that houseCARL does not check the name for that collision instead of implying the hazard is closed. Teaching `IsStemFree` about foreign inactive plugins is a change to the fresh-write path on every `patch=` call, not to a refusal's wording; I'd rather it were its own issue than a rider here, but say the word and I'll fold it.

**Summary 7 — the cornerstone.** Your call, and here is my honest read. The refusal is now four sentences plus the inventory; a real one, after the reorder, measures like this:

> cannot extend: mod folder 'RemoveHereMod' exists but was NOT created by houseCARL (no marker) — refusing to write into a folder houseCARL doesn't own (originals untouched, Q3). Omit into= and pass patch= a name no mod folder already uses to create a fresh patch. houseCARL owns: into="houseCARL - DupHome" (SeedA.esp)  |  into="houseCARL - GhostPatch" (GhostPatch.esp)  |  into="Alpha.esp" (in houseCARL - TwoEsp)  |  …  To write into an existing plugin IN PLACE instead, pass in_place="RemoveHere.esp".

I think the cornerstone is about not returning a silently wrong answer or a degraded mode, and it is holding: every clause here is now true when printed and absent when not. But "one plain sentence" is being spent, and the spend is real — this is the longest refusal on the write surface. If you want it back under control the lever I'd reach for is the inventory, not the prose: cap it at three or four near-stem rows rather than eight, since the ranking now means the row the caller wants is at the front, and the tail already says how many were cut. That would take a typical refusal to about half its current length with no loss to the caller who typo'd. I have not done it because eight was my number, picked in this PR with no evidence behind it, and how much a refusal may enumerate is the judgement #380 explicitly routed to you.

Avick3110:
Blind review folded (8 folded, 1 refuted with reason, 2 left for Aaron: the IsStemFree half of #359 and the refusal length). Queued for Aaron's review.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8226,6 +8272,136 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+    /// a FILENAME and resolves it through the ACTIVE load order, so a plugin MO2 has not enabled is not reachable by
+    /// it. Empty when the folder holds no plugin or none of them is active — which is the whole point, since the
+    /// caller is then offered no in-place sentence at all rather than one that leads to a second refusal.</summary>
+    IReadOnlyList<string> ActiveInPlaceTargets(string folder)
+    {
+        var active = ActivePluginBasenames();
+        try
+        {
+            return Directory.EnumerateFiles(folder).Select(f => Path.GetFileName(f)!)
+                .Where(n => PluginExts.Any(ext => n.EndsWith(ext, StringComparison.OrdinalIgnoreCase)) && active.Contains(n))
```
Avick3110: **`in_place=` is offered for a plugin that lives in a different folder.**

`ActiveInPlaceTargets` matches the foreign folder's plugin filenames against `ActivePluginBasenames()` — a set of *basenames*, with no check that the active plugin of that name is the copy in *this* folder.

Scenario: the user has `OldWeapons` (disabled, or lower priority) and `OldWeapons Updated` (enabled), both shipping `CoolWeapons.esp`. `into="OldWeapons"` hits the un-owned arm; this method enumerates `OldWeapons\CoolWeapons.esp`, sees `CoolWeapons.esp` in the active set (supplied by the *other* folder), and the refusal closes with `pass in_place="CoolWeapons.esp"`. Following it lands in `ApplyEditsInPlace`, which resolves the filename through the load order — "in-place edits the file the game actually loads" — and rewrites `OldWeapons Updated\CoolWeapons.esp`, not the folder the caller named. That is a consent-gated in-place write on a file the sentence never claimed.

The new `WriteSurfaceGuardProbe` arm is literally this shape: `W2TwinForeign` holds a **copy** of `W2Repl`'s plugin and is not enabled, yet the sentence is emitted and observed as if the lane would take that file. `Mo2LoadOrder.Build(...).OrderedPaths` already carries the resolving path, so the gate could compare paths rather than names (the `_resolver.PluginNames` branch would need the same).

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8226,6 +8272,136 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+            if (tokens.Count == 0) { more++; continue; }                    // no spelling reaches it — never print a false one
+            foreach (var t in tokens)
+            {
+                if (rows.Count == cap) { more++; continue; }
+                rows.Add(t);
+            }
+        }
+        if (rows.Count == 0)
+            return needEsp
+                ? "houseCARL owns no patch holding a plugin yet."
```
Avick3110: **The empty-inventory clause is false when every owned patch was merely unreachable.**

`rows.Count == 0` is reached two ways: houseCARL genuinely owns nothing, and every owned patch was dropped by `ReachingTokens` (`more > 0`). Both render "houseCARL owns no patch holding a plugin yet.", which is a wrong statement in the second case — and it throws away the `more` count computed one line above for exactly this purpose.

Scenario: the user duplicates a two-plugin houseCARL patch folder in MO2 to back it up, so the marker is copied too — `houseCARL - Foo` and `houseCARL - Foo backup` both hold `Alpha.esp` and `Beta.esp`, and no other owned patch exists. `Reaches` rejects each folder name (`needEsp`, `Plugins.Count == 2`, no `<folder>.esp`) and each plugin token (`byEsp.Count > 1` → the ambiguous arm), so `tokens` is empty for both, `more == 2`, `rows` empty. The refusal tells the caller houseCARL owns no patch holding a plugin, so they mint a fresh one — when two extendable patches exist and the real answer is that their names are ambiguous.

A third branch for `rows.Count == 0 && more > 0` ("houseCARL owns N patch(es), none reachable by a single into= spelling") keeps the refusal honest.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5674,7 +5676,11 @@ public WritePatchBuilder.RemovalOutcome RemoveRecords(IReadOnlyList<string> form
             string outPath;
             try { outPath = ResolveOutputPath(patchName: null, into: patch, out _, out _,
                                               laneClause: WriteSentences.RemoveNoFreshPatch
-                                                          + (inPlaceRemedy is null ? "" : " " + inPlaceRemedy)); }
+                                                          + (inPlaceRemedy is null ? "" : " " + inPlaceRemedy + WriteSentences.InPlaceAnyFilename),
```
Avick3110: **The not-found arm puts the consent-gated in-place clause ahead of the candidate list, which is the ordering this PR argues against.**

`laneClause` carries `RemoveNoFreshPatch + in-place sentence`, and the not-found arm renders `laneClause` **before** `OwnedPatchCandidates`. So `housecarl_remove` with `patch="Typo"` reads:

> …'Typo'. \<no-fresh-patch rule\> To remove from an existing plugin IN PLACE instead, pass in_place="\<plugin filename\>". houseCARL owns: into=… .

The un-owned arm goes to real trouble (and a probe assertion at `ExtendResolveProbe.cs:548`) to put the clause that rewrites a file the caller did not author **last**, behind the safe candidates. The two arms of the same resolver now disagree about which remedy gets the emphasis. Splitting `laneClause` so the in-place half is appended after `OwnedPatchCandidates` would make them agree.

src/housecarl-mcp/WriteSentences.cs
```diff
@@ -574,10 +577,21 @@ internal static string PlaceSourceOffOrder(string provider, bool ownerEnabled =
...
+        "To remove from an existing plugin IN PLACE instead, pass in_place=";
+
+    /// <summary>The completion when no plugin can be named — the refusal knows the lane but not the file.</summary>
+    [NoClaims("a completion fragment; the claim lives in the lane sentence it finishes")]
+    internal const string InPlaceAnyFilename = "\"<plugin filename>\".";
+
+    /// <summary>One lane sentence, finished. <paramref name="plugins"/> are the plugins the in-place lane could
+    /// actually take here — a plugin must be in the ACTIVE load order for in_place= to resolve it — so an empty set
+    /// means the lane is unavailable and the sentence is not offered at all.</summary>
+    internal static string InPlaceLane(string lane, IReadOnlyList<string> plugins)
```
Avick3110: **Uncapped join, next to a list that is capped at 8.**

`OwnedPatchCandidates` caps its rows at 8 with a `+N more` tail because a long sentence stops being a readable next step. This one joins *every* active plugin in the foreign folder with `" or "` and no cap.

`into=` naming a large enabled mod folder — a compilation or a merged pack shipping, say, 25 active plugins — ends the refusal with one sentence carrying 25 quoted filenames. Same overflow, same arm, opposite treatment. A cap plus "or one of N others" would match the sibling clause.

src/housecarl-generator/ExtendResolveProbe.cs
```diff
@@ -489,11 +497,151 @@ void MarkOwned(string folder, string plugin) =>
...
+                static JsonElement Doc(string s) => JsonDocument.Parse(s).RootElement;
+                string named = "in_place=\"RemoveHere.esp\"";
+
+                var ap = ApplyTools.Apply(svc, ops: Doc($"[{{\"formid\":\"{fid}\",\"field_path\":\"BasicStats.Weight\",\"value\":\"2\"}}]"),
+                                          into: "RemoveHereMod");
+                Check(ap.Contains(WriteSentences.ExtendInPlaceLane, StringComparison.Ordinal) && ap.Contains(named, StringComparison.Ordinal),
+                      "housecarl_apply names the in-place lane AND the plugin in that folder the lane can take");
+                // The one clause here that rewrites a file the caller did not author reads LAST, behind both lanes
+                // that leave every original untouched (#380).
+                Check(ap.IndexOf("houseCARL owns:", StringComparison.Ordinal) < ap.IndexOf("in_place=", StringComparison.Ordinal),
```
Avick3110: **This ordering assertion passes vacuously when the list is gone.**

`IndexOf` returns `-1` for a missing substring, and `-1 < <any real index>` is true. So if `OwnedPatchCandidates` ever returns the empty-inventory clause on this arm — e.g. the owned-folder scan starts throwing an exception type the `catch (IOException) { } catch (UnauthorizedAccessException) { }` swallows — `ap.IndexOf("houseCARL owns:")` is `-1`, `ap.IndexOf("in_place=")` is positive, and the arm that exists to pin safe-clause-first ordering reports green on a render with no candidate list at all.

```suggestion
                int ownsAt = ap.IndexOf("houseCARL owns:", StringComparison.Ordinal);
                Check(ownsAt >= 0 && ownsAt < ap.IndexOf("in_place=", StringComparison.Ordinal),
```

Avick3110:
Second fold — your five inline findings. All five folded, nothing refuted. Build clean, tests 1160 passed, probes 128/128.

## Folded

**`in_place=` offered for a plugin that lives in a different folder** (`LoadOrderService.cs:8285`). Real, and the worse half is that following the sentence is a consent-gated write on a file it never named. `ActiveInPlaceTargets` now matches on the RESOLVED PATH, not the basename: a new `ActivePluginPaths()` builds filename → the path the active order actually resolves it to (the built resolver's `view.PluginPath` when there is one, else `Mo2LoadOrder.Build(...).OrderedPaths`, same best-effort composition as `ActivePluginBasenames`), and a plugin is offered only when the copy in THIS folder is the one that map points at. A losing copy gets no sentence.

Arm 9e is the shape you described: `TwinRemoveMod` is un-owned, not in modlist.txt, and holds a copy of `RemoveHere.esp`, whose winner stays in `RemoveHereMod`. The refusal must carry no `in_place=` at all, and the winning folder's refusal in the same arm must still carry it, so the gate is a path check and not a blanket withdrawal. Pre-fix the first check reddens.

You were also right that `write-surface-guard`'s new arm was a render of the sentence being false — `W2TwinForeign` held a copy and was never enabled. The twin folder is gone; the observed call is `into="W2Repl"` instead, an ordinary enabled un-owned mod folder whose replacer plugin IS the winner, so the sentence the two transports observe is one the render really carries and the fixture gains nothing new.

**The empty-inventory clause is false when every owned patch was merely unreachable** (`LoadOrderService.cs:8353`). Real, and the count was already computed. A third branch: `rows.Count == 0 && more > 0` renders "houseCARL owns N patches, none reachable by a single into= spelling (their folder and plugin names collide) — rename one of them in MO2 to name it." No create remedy is appended — this tail closes every lane and a create route is false on the ones that cannot create, so renaming is the one step true at every altitude.

Arm 9f builds your backup-duplicate scenario, on its own synthetic instance because the claim is about the WHOLE inventory: two owned folders each holding `Alpha.esp` + `Beta.esp`, no folder name resolving, both plugin names ambiguous. Pre-fix it reads "houseCARL owns no patch holding a plugin yet."

**The not-found arm puts the consent-gated clause ahead of the candidates** (`LoadOrderService.cs:5679`). Real — and yes, this is the ordering the un-owned arm was reworked to get, so the two arms disagreeing was mine to fix rather than defend. I refuted the shape of this last round on the grounds that moving the sentence would hand it to three tools that do not render it there; that argument does not apply to splitting the clause, which is what this needed. `laneClause` now carries `RemoveNoFreshPatch` alone and the in-place half is handed in as a separate `laneInPlace`, appended LAST, behind `OwnedPatchCandidates`. Removal is still the only caller that passes either, so no other tool's sentence moves. Pinned in arm 8d3 with the same present-and-before test.

**Uncapped join next to a list capped at 8** (`WriteSentences.cs:592`). Real. `InPlaceLane` caps at 4 names and closes "or one of N others in that folder". Arm 9g renders it directly for 25 plugins and for one — building 25 active plugins to read one sentence would make the fixture the claim.

**The ordering assertion passes vacuously** (`ExtendResolveProbe.cs:548`). Correct, `-1` satisfies it. Took your suggestion verbatim, and the new not-found ordering check in 8d3 is written the same way.

## Counts

New probe arms: 9e (in-place lane gated on the winning copy), 9f (the unreachable-inventory clause), 9g (the lane sentence's cap), plus two checks folded into existing arms (8d3's not-found ordering, 9b's non-vacuous ordering). No new xUnit tests — this is the resolver's refusal wording end to end, which `extend-resolve-guard` already owns against a synthetic MO2 instance, and TESTING.md says not to duplicate a probe.

Avick3110:
Reworked to Aaron's ruling of 2026-09-06: these refusals are one sentence — what went wrong and what to try, with the nearest owned candidates named inside it, no inventory block and no trailing in-place clause. The in-place lane is gone from both arms entirely (it stays discoverable on the tool that declares it); the reachability check and the nearest-name ranking that pick the candidates stay, capped at three and ordered nearest first inside a folder as well as across folders. Both arms now read, measured off `extend-resolve-guard`'s synthetic instance:

**The un-owned-folder arm** (record lane; the rider lane says `patch_name=` and "a fresh folder"):

> cannot extend: mod folder 'houseCARL - Foreign' exists but was NOT created by houseCARL (no marker), so writing into it would touch a mod houseCARL doesn't own (originals untouched, Q3); try into="houseCARL - DupHome", into="houseCARL - GhostActive\_001" or into="houseCARL - GhostPatch", or patch= a name no mod folder already uses for a fresh patch.

**The not-found arm, record lane:**

> cannot extend: no houseCARL patch named 'Soloo' — no owned folder holds 'Soloo.esp' and none is named 'houseCARL - Soloo' or 'Soloo'; try into="Solo.esp", into="houseCARL - DupHome" or into="houseCARL - GhostActive\_001", or patch="Soloo" for a fresh patch (auto-suffixed if that name is taken).

**The not-found arm, removal lane** (no fresh route, and it says why):

> cannot extend: no houseCARL patch named 'GhostRemove' — no owned folder holds 'GhostRemove.esp' and none is named 'houseCARL - GhostRemove' or 'GhostRemove', and houseCARL will not create a patch here, since a removal only drops a record the patch ITSELF already carries; try into="houseCARL - DupHome", into="houseCARL - GhostActive\_001" or into="houseCARL - GhostPatch".

**The all-unreachable arm** (owned patches exist, no single `into=` spelling reaches any of them):

> cannot extend: no houseCARL patch named 'GhostNowhere' — no owned folder holds 'GhostNowhere.esp' and none is named 'houseCARL - GhostNowhere' or 'GhostNowhere'; try renaming one of the 2 patches houseCARL owns in MO2, since no single into= spelling reaches any of them (their folder and plugin names collide), or patch="GhostNowhere" for a fresh patch (auto-suffixed if that name is taken).

Probe arms that pinned the old shape were rewritten rather than deleted: each arm now pins one terminating period, the `; try ` remedy, the ordering, the cap of three, and the absence of any in-place clause — at service altitude and at tool altitude on all four write tools. Arm 9e (the in-place lane's path check) and 9g (its cap) went with the clause they covered, and `WriteSentences.ExtendInPlaceLane`, `InPlaceLane()` and `InPlaceAnyFilename` are gone. `RemoveInPlaceLane` went too: with the extend arms no longer carrying it, its only remaining site was removal's "patch is required" arm, which `housecarl_remove` answers in its own words before the service is reached — the shared-sentence guard caught it as unrendered.

Tests 1160 passed, probes 128/128 passed, build clean.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8226,6 +8263,113 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+            foreach (var dir in Directory.EnumerateDirectories(_modsDir).OrderBy(d => d, StringComparer.OrdinalIgnoreCase))
+            {
+                if (!IsHouseCarlOwned(dir)) continue;
+                var plugins = Directory.EnumerateFiles(dir)
+                    .Where(f => PluginExts.Any(ext => f.EndsWith(ext, StringComparison.OrdinalIgnoreCase)))
+                    .Select(f => Path.GetFileName(f)!).OrderBy(n => n, StringComparer.OrdinalIgnoreCase).ToList();
+                owned.Add(new OwnedPatch(dir, Path.GetFileName(dir), plugins));
+            }
+        }
+        catch (IOException) { } catch (UnauthorizedAccessException) { }
```
Avick3110: **medium** — a partial folder scan here can print an `into=` spelling that the next call refuses, which is exactly the guarantee this module claims.

The `try` wraps the whole enumeration, so a throw partway (MO2 holds `meta.ini` open while it updates mod metadata, so `IsHouseCarlOwned`'s `File.ReadLines` raising `IOException` is not exotic; `EnumerateDirectories` also throws lazily mid-iteration) leaves `owned` **partially** populated — and the method then goes on to build candidates from it. `ReachingTokens.Reaches` decides reachability by counting how many *other* owned folders hold `<stem>.esp`; drop one of a colliding pair from `owned` and an ambiguous token looks unambiguous. Concretely: two owned folders both hold `Alpha.esp`, the second one's `meta.ini` is momentarily locked, and the refusal prints `into="Alpha.esp"` — following it hits `cannot extend: 2 houseCARL folders carry 'Alpha.esp' — ambiguous`. That is the second refusal the doc comment above says a caller "never meets", and the summary's own promise ("an unreadable ModsDir yields **no** candidates") is not what the code does.

Dropping the partial set on the way out matches the documented behaviour:

```suggestion
        catch (IOException) { owned.Clear(); } catch (UnauthorizedAccessException) { owned.Clear(); }
```

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8226,6 +8263,113 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+    /// folders read here, and on the record lane it must also land on a definite plugin, so a caller who takes one
+    /// literally never meets a second refusal. A patch no token reaches is counted instead, and the sentence says so
+    /// when there is nothing to name. Nearest <paramref name="stem"/> first — the caller who reached this refusal
+    /// typed something close to what they meant — ranked by the shared name-suggester, and capped at three so the
+    /// refusal stays one readable sentence. <paramref name="needEsp"/> drops the folders holding no plugin, which the
+    /// record lane could not extend anyway. Best-effort: an unreadable ModsDir yields no candidates rather than
+    /// failing the refusal.</summary>
+    PatchCandidates OwnedPatchCandidates(bool needEsp, string stem)
+    {
+        const int cap = 3;
```
Avick3110, line 8525: **low** — the cap drops candidates silently, while its sibling branch counts them.

`ExtendRefusal`'s unreachable branch says "one of the **2** patches houseCARL owns", but past this cap the sentence just stops: a user with 40 owned patches sees three names and nothing saying there are 37 more. Probe 9d shows the shape it produces on a stem the suggester ranks nothing for — `into="GhostCapped"` answers `try into="houseCARL - Bulk00", into="houseCARL - Bulk01" or into="houseCARL - Bulk02"`, i.e. the alphabetically-first three, with `houseCARL - GhostPatch` (the actual near-miss) not named and no hint that it exists. The PR body describes a "+N more" count; the code has none. A short `(+N more)` after the list would keep it one sentence and stop the list reading as exhaustive.

Related, same loop: once `rows.Count == cap` the loop still runs `ReachingTokens` for every remaining owned folder and keeps incrementing `unreachable` — harmless today only because `Unreachable` is read only when `Tokens` is empty.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8179,35 +8191,60 @@ string ResolveOwnedPatchFolder(string into, bool needEsp,
...
+        {
+            var t = candidates.Tokens;
+            tries.Add(t.Count == 1 ? t[0] : string.Join(", ", t.Take(t.Count - 1)) + " or " + t[^1]);
+        }
+        else if (candidates.Unreachable > 0)
+            tries.Add($"renaming one of the {candidates.Unreachable} patch{(candidates.Unreachable == 1 ? "" : "es")} houseCARL owns "
+                    + "in MO2, since no single into= spelling reaches any of them (their folder and plugin names collide)");
+        if (fresh is not null) tries.Add(fresh);
+
+        var nothingOwned = candidates.Tokens.Count == 0 && candidates.Unreachable == 0
```
Avick3110: **low** — the one case this PR set out to fix still dead-ends, and reads badly doing it.

On the removal lane `noFreshRule` is set and `fresh` is null, so a brand-new install with no owned patches yet (`housecarl_remove into="X"`) produces `tries.Count == 0` **and** `nothingOwned`, and both clauses open with ", and":

> cannot extend: no houseCARL patch named 'X' — no owned folder holds 'X.esp' and none is named 'houseCARL - X' or 'X', and houseCARL will not create a patch here, since a removal only drops a record the patch ITSELF already carries, and houseCARL owns no patch holding a plugin yet.

Two stacked ", and" clauses, and no "what to try" at all — the dead-end #359/#380 are about, on the likeliest first-time path. There is genuinely nothing to *extend* here, but there is something to say: the patch has to be created by an apply/create/forward first, which is the fact the caller is missing. Worth one clause on this branch rather than a full stop.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8179,35 +8191,60 @@ string ResolveOwnedPatchFolder(string into, bool needEsp,
...
+        // lanes', where the spelling is the same on every caller (#357).
+        throw new InvalidOperationException(ExtendRefusal(
+            $"no houseCARL patch named '{stem}' — no owned folder holds '{espName}' and none is named "
+            + $"'{ModFolderName(stem)}'"
+            + (string.Equals(bareName, ModFolderName(stem), StringComparison.OrdinalIgnoreCase) ? "" : $" or '{bareName}'"),
+            noFreshRule,
+            OwnedPatchCandidates(needEsp, stem),
+            riderNaming is { } rn
+                ? $"{rn.Param}=\"{stem}\" for a fresh folder (auto-suffixed if that name is taken"
+                  + (rn.Caveat is null ? ")" : $"; {rn.Caveat.TrimEnd('.')})")
```
Avick3110: **low** — the caveat is a standalone sentence spliced mid-clause, so it lands with a capital after a semicolon. Probe 8c's own output:

> … or patch_name="GhostRider" for a fresh folder (auto-suffixed if that name is taken; **On** this tool a bare patch= names the ARCHIVE (archive_name=), not the folder).

`TrimEnd('.')` handles the tail but not the head. Given the whole point of this change is one clean sentence, lowercasing the first character (or storing the caveat as a clause the way `RemoveNoFreshPatch` now is) would finish the job. Same splice on the un-owned arm at L8177.

plugin/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Changelog
+﻿# Changelog
```
Avick3110: **low** — unintended UTF-8 BOM churn. Six files in this diff gained a leading `EF BB BF` with no other change on line 1:

```
src/housecarl-mcp/ApplyTools.cs        ef bb bf
src/housecarl-mcp/CreateTools.cs       ef bb bf
src/housecarl-mcp/ForwardTools.cs      ef bb bf
src/housecarl-mcp/LoadOrderService.cs  ef bb bf
src/housecarl-mcp/RemoveTools.cs       ef bb bf
plugin/CHANGELOG.md                    ef bb bf
```

Untouched siblings (`RecordsTools.cs`, `WriteSentences.cs`) have none, so this is a repo-wide inconsistency introduced by an editor rather than a deliberate change. `plugin/CHANGELOG.md` is the one that actually matters — it ships, and some Markdown renderers surface the BOM as a stray glyph before the `#`. Worth stripping all six.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8226,6 +8263,113 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+                owned.Add(new OwnedPatch(dir, Path.GetFileName(dir), plugins));
+            }
+        }
+        catch (IOException) { } catch (UnauthorizedAccessException) { }
+
+        // Near-stem first (#380 asks for candidates near what the caller typed, not the alphabet's first eight).
+        // Ranked by the same suggester the missed-plugin lookups use, over both the folder names and the plugin
+        // names, so a typo'd folder and a typo'd plugin both float. Anything it does not rank keeps its alphabetical
+        // place behind them.
+        var near = PluginNameSuggest.Nearest(stem, owned.SelectMany(o => o.Plugins.Prepend(o.Name)), max: 32);
```
Avick3110, line 8552: **medium** — the "nearest first" ranking silently degrades to the alphabet, which is the thing #380 named.

`PluginNameSuggest.Nearest` is documented to return **empty** when nothing clears its relevance bar — "a wrong 'did you mean' is worse than none" (`PluginNameSuggest.cs:15`). That is a deliberate refusal to guess. Here the empty result does not propagate: every unranked folder gets `NearRank == int.MaxValue`, `OrderBy(Rank)` is stable, so the list falls back to the enumeration order — which line 8272 has already sorted `OrderBy(d => d, OrdinalIgnoreCase)`. The refusal then names the alphabetically-first three owned patches and calls them the nearest.

This is not hypothetical; it is what the probe itself prints. From arm 9d on this branch:

```
into="GhostCapped" ->
  ...; try into="houseCARL - Bulk00", into="houseCARL - Bulk01" or into="houseCARL - Bulk02", or patch="GhostCapped" ...
```

The fixture also holds `houseCARL - GhostPatch`, `houseCARL - GhostRemove` and `houseCARL - GhostActive_001`. None of them is offered; three `Bulk` folders are, purely because `B` sorts before `D` and `G`. Arm 9c is the milder version of the same thing: for `into="Soloo"` only `into="Solo.esp"` is actually ranked, and the other two slots (`houseCARL - DupHome`, `houseCARL - GhostActive_001`) are alphabetical filler.

The comment on 8297 says #380 asks for candidates near what the caller typed, "not the alphabet's first eight". With the cap at three and no relevance floor, this is the alphabet's first three — and CHANGELOG's "nearest what you typed first" plus the `OwnedPatchCandidates` doc-comment ("Nearest `stem` first ... ranked by the shared name-suggester") are both untrue on that path. The refusal is one plain sentence, but it is not a true one, which is the cornerstone this PR is being measured against.

Two shapes that would hold: drop unranked folders entirely (offer only what the suggester vouches for, and fall through to the count/no-candidates wording when it vouches for nothing), or keep the fallback but stop calling it nearest.

src/housecarl-generator/ExtendResolveProbe.cs
```diff
@@ -489,11 +503,193 @@ void MarkOwned(string folder, string plugin) =>
...
+                    File.WriteAllText(Path.Combine(d, "Alpha.esp"), "a");
+                    File.WriteAllText(Path.Combine(d, "Beta.esp"), "b");
+                }
+
+                var store2 = new UserConfigStore(Path.Combine(root, "houseCARL.user2.json"));
+                using var svc2 = LoadOrderService.WithInstance(inst2, 0, store2);
+                svc2.Stats();
+                var unreachable = svc2.ApplyEdits(new[] { Wgt(1) }, null, "GhostNowhere");
+                string text = unreachable.Error ?? "";
+                Check(text.Contains("try renaming one of the 2 patches houseCARL owns in MO2", StringComparison.Ordinal)
```
Avick3110, line 732: **low** — this is the one remedy in the file that is asserted but never followed.

The arm's own neighbours set the bar explicitly: 8b "a remedy is a claim about what a call does, so it is checked by MAKING the call", 8d2 follows the in-place lane, and 9c follows both rendered `into=` spellings. Here the assertion stops at `text.Contains("try renaming one of the 2 patches houseCARL owns in MO2")` — it pins the wording, not the claim. Nothing in the suite shows that renaming one of the twins in MO2 actually makes an `into=` spelling resolve, so the sentence could go on being printed after a change that falsifies it.

It is cheap to close on this fixture: rename `houseCARL - Twin backup` to `houseCARL - Alpha`, re-`Stats()`, and assert `into="Alpha"` now resolves to that folder. That turns the last unfollowed remedy in this resolver into a followed one.

src/housecarl-generator/ExtendResolveProbe.cs
```diff
@@ -1,3 +1,4 @@
+using System.Text.Json;
```
Avick3110, line 1: **low** — stale header at line 34 of this file (not in a diff hunk, so anchored here).

The not-found arm no longer says "create it fresh" on the record lanes; it now says `patch="<stem>" for a fresh patch`, and it also names the owned candidates, which this line does not mention at all. The header is the map a reader uses before reading 700 lines of arms, so it should read something like: refuse, naming both places searched, then the nearest owned patches and the lane's own fresh-patch parameter — one sentence.

Avick3110:
Folded all eight findings from this review; nothing refuted. Rebased on main (changelog conflict kept both sides, the extend entry taking its reworked wording). Build, 1365 tests and 128/128 probes green.

**Folded**

1. **"nearest first" degrading to the alphabet** (medium). `PluginNameSuggest.Nearest` returning empty left every folder tied at `int.MaxValue`, so the stable sort fell back to the alphabetical enumeration order while the doc comment and changelog still called it nearest. Added `PluginNameSuggest.StemDistance`, an edit distance that always answers, as the tiebreak behind the suggester's own ranking — on folders and on the tokens inside one. Arm 9d now prints `into="houseCARL - GhostPatch"` first, the near-miss the review said went unnamed, and asserts it does.

2. **Partial folder set on IOException** (medium). `catch (IOException) { owned.Clear(); }` on both arms, so the documented degradation ("an unreadable ModsDir yields no candidates") is what the code does. New probe arm locks a twin's `meta.ini` with `FileShare.None` mid-scan and asserts no candidate is offered — in particular not `into="Alpha.esp"`, the spelling that only looks unambiguous because the colliding folder went missing.

3. **The cap dropping candidates with no count** (low). `PatchCandidates` carries `BeyondCap`, rendered as `(+N more)`. 9d asserts the shape and that the count tracks the inventory: one more offerable patch, one higher.

4. **Removal lane with nothing owned** (low). The facts are now clauses joined with a single trailing `, and`, so two cannot stack, and the branch with no fresh route and no candidates names the writes that can make a patch (`housecarl_apply`, `housecarl_create_record`, `housecarl_forward_record`) instead of stopping dead. New arm 9g, its own instance owning nothing.

5. **Rider caveat spliced mid-clause with a capital** (low). `AsClause` lowers the leading capital and drops the terminating period, leaving an acronym alone (`ARCHIVE` survives). Both splice sites use it; 8c asserts the rendered `; on this tool a bare patch= names the ARCHIVE`.

6. **UTF-8 BOM churn** (low). Stripped from all six files; the whole branch diff is BOM-free.

7. **9f's rename remedy asserted but never followed** (low). The twins hold real plugins now, and the arm renames `houseCARL - Twin backup` to `houseCARL - Alpha`, re-`Stats()`, and asserts `into="Alpha"` resolves to that folder and writes there.

8. **Stale probe header** (low). The NOT-FOUND line now says: refuse in one sentence, naming both places searched, then the nearest owned patches as `into=` spellings and the lane's own fresh-patch parameter.

Avick3110:
Reworked to the one-sentence ruling, fresh blind review folded (8 of 8). Ready for Aaron's second review, since the refusal text is new since his first pass.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8331,35 +8343,84 @@ string ResolveOwnedPatchFolder(string into, bool needEsp,
...
+        else if (candidates.Unreachable > 0)
+            tries.Add($"renaming one of the {candidates.Unreachable} patch{(candidates.Unreachable == 1 ? "" : "es")} houseCARL owns "
+                    + "in MO2, since no single into= spelling reaches any of them (their folder and plugin names collide)");
+        if (fresh is not null) tries.Add(fresh);
+        // A lane with no fresh route, on an install owning nothing yet, otherwise stops dead — the likeliest first-time
+        // path, and the dead end #359/#380 are about. There is nothing to extend, but there is something to say: the
+        // patch has to exist before this lane can touch it, and only a writing lane can make one.
+        if (tries.Count == 0 && noFreshRule is not null && nothingOwned)
+            tries.Add("making the patch first with a write that creates one (housecarl_apply, housecarl_create_record "
+                    + "or housecarl_forward_record), then naming it here");
```
Avick3110: **Retired tool names in the empty-inventory remedy.**

This clause names `housecarl_create_record` and `housecarl_forward_record`. Both are retired on the 2.0 surface — `AliasTable.RetiredTools` has rows redirecting them into `ToolNames.Create` (`housecarl_create`) and `ToolNames.Forward` (`housecarl_forward`).

Scenario (arm 9g's own fixture): `housecarl_remove(formids=[...], into="Nothing")` on an install owning no patch refuses with `; try making the patch first with a write that creates one (housecarl_apply, housecarl_create_record or housecarl_forward_record)`. A caller who follows it gets `error: housecarl_create_record is not on this surface — absorbed into housecarl_create: ...` — a second refusal handed out by the first one. `housecarl_apply` is correct; the other two are dead.

`RetiredNameRedirectTests` only sweeps `AliasTable` rows against `tools/list`, so free-text mentions like this one are not covered. Using the `ToolNames` constants would also make this un-driftable.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8331,35 +8343,84 @@ string ResolveOwnedPatchFolder(string into, bool needEsp,
...
+        // lanes', where the spelling is the same on every caller (#357).
+        throw new InvalidOperationException(ExtendRefusal(
+            $"no houseCARL patch named '{stem}' — no owned folder holds '{espName}' and none is named "
+            + $"'{ModFolderName(stem)}'"
+            + (string.Equals(bareName, ModFolderName(stem), StringComparison.OrdinalIgnoreCase) ? "" : $" or '{bareName}'"),
+            noFreshRule,
+            OwnedPatchCandidates(needEsp, stem),
+            riderNaming is { } rn
+                ? $"{rn.Param}=\"{stem}\" for a fresh folder (auto-suffixed if that name is taken"
+                  + (rn.Caveat is null ? ")" : $"; {AsClause(rn.Caveat)})")
```
Avick3110: **The fresh-write remedy no longer tells the caller to drop `into=`, and on this lane that makes it a loop.**

The old sentence led with "Omit into= and pass `patch_name=\"X\"` ...". The rewrite kept only the parameter. But `ResolvePatchModFolder` takes the extend branch whenever `into` is non-blank (line 7735) and never looks at `patchName`, so a caller who literally does what this clause says — adds `patch_name="GhostRider"` to the call that just failed — re-enters the same branch and gets the **identical refusal back**. `housecarl_bsa_repack` declares no into=/patch_name= exclusivity check, so nothing intercepts it: it is a loop, not a second explaining refusal.

Same omission on the record lanes (line 8366 and the un-owned arm at 8328/8332), where it is less severe only because `ApplyTools`:149 / `CreateTools`:123 / `ForwardTools`:95 refuse `patch=` + `into=` with "Drop patch= to extend, or drop into= to write fresh."

Worth noting arm 8b "follows" this remedy as `svc.ApplyEdits(new[] { Wgt(4) }, "GhostActive", null)` — with `into` **null**. It proves a call the sentence does not describe, which is why the gap survived a probe written to catch exactly this.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8378,6 +8439,124 @@ List<string> OwnedFoldersHolding(string espFileName)
...
+                var plugins = Directory.EnumerateFiles(dir)
+                    .Where(f => PluginExts.Any(ext => f.EndsWith(ext, StringComparison.OrdinalIgnoreCase)))
+                    .Select(f => Path.GetFileName(f)!).OrderBy(n => n, StringComparer.OrdinalIgnoreCase).ToList();
+                owned.Add(new OwnedPatch(dir, Path.GetFileName(dir), plugins));
+            }
+        }
+        // A throw partway through leaves a PARTIAL set, and reachability is decided by counting how many OTHER owned
+        // folders hold the same plugin — so a dropped folder can make an ambiguous token look unambiguous and print a
+        // spelling the next call refuses. No candidates is the documented degradation; half of them is a wrong answer.
+        catch (IOException) { owned.Clear(); } catch (UnauthorizedAccessException) { owned.Clear(); }
```
Avick3110: **A failed scan renders as the positive claim "houseCARL owns no patch" — the exact sentence the `Unreachable` counter exists to avoid.**

The comment just above says "Two different facts reach an empty list, and only one of them is 'there is nothing to extend'." There are three. This `owned.Clear()` is the third, and it is folded into the first: with `Tokens=0` and `Unreachable=0`, `ExtendRefusal` computes `nothingOwned = true` and adds the fact `houseCARL owns no patch holding a plugin yet` (line 8409) — and on the removal lane also appends `; try making the patch first with a write that creates one` (line 8402).

Scenario: MO2 holds one owned folder's `meta.ini` open, or one mod folder carries a denying ACL. `IsHouseCarlOwned` (no try/catch of its own) or `EnumerateFiles` throws, every candidate is discarded, and a user with 40 houseCARL patches is told houseCARL owns none — then told to go create one, minting a duplicate beside patches they could have extended. That is the failure mode lines 8393-8395 spend a paragraph preventing for the unreachable case.

The XML doc and the CHANGELOG both promise only that "no candidate is offered at all rather than the half the scan reached" — which is right. Suppressing the candidates is fine; asserting the inventory is empty is not. A third state (scan failed) would let the sentence say so, or at least stay silent.

Arm 9f currently pins the wrong wording: `Check(Candidates(locked).Count == 0 && locked.Contains("houseCARL owns no patch", ...))`.

Avick3110:
Third fold — your three findings from the second review. All three folded, nothing refuted. Rebased on main (clean). Build clean, 1443 tests, 128/128 probes.

## Folded

**Retired tool names in the empty-inventory remedy** (`LoadOrderService.cs:8404`). Real: `housecarl_create_record` and `housecarl_forward_record` are both absorbed rows in `AliasTable.RetiredTools`, so following the remedy got a second refusal handed out by the first. The clause is now built from `ToolNames.Apply`, `ToolNames.Create` and `ToolNames.Forward`, so a rename cannot drift it again. Arm 9g asserts the rendered names against the same constants and that neither retired spelling appears.

**A failed scan rendering as "houseCARL owns no patch"** (`LoadOrderService.cs:8476`). Real, and the third empty list you identified is now a state rather than a fold into the first. `PatchCandidates` carries `ScanFailed`, both `catch` arms set it, `nothingOwned` excludes it, and the sentence says `try again once the mods folder can be read, since houseCARL could not scan it just now and so cannot name a patch or tell whether it owns any` — no inventory claim and, on the removal lane, no "go make one" either, which was the half that would have minted a duplicate beside forty existing patches. Arm 9f's locked-`meta.ini` block now pins the new wording instead of the wrong one, and adds the removal-lane render, which is where the false claim did the damage:

> cannot extend: no houseCARL patch named 'GhostLocked' — no owned folder holds 'GhostLocked.esp' and none is named 'houseCARL - GhostLocked' or 'GhostLocked', and houseCARL will not create a patch here, since a removal only drops a record the patch ITSELF already carries; try again once the mods folder can be read, since houseCARL could not scan it just now and so cannot name a patch or tell whether it owns any.

**The fresh-write remedy no longer telling the caller to drop `into=`** (`LoadOrderService.cs:8363`). Real, and a loop on the rider lanes exactly as you describe — `ResolvePatchModFolder` takes the extend branch on any non-blank `into=` and never reads `patchName`, and no rider tool declares an exclusivity check. All four fresh clauses (not-found and un-owned, record lane and rider lane) now read `dropping into= and passing patch="X" ...` / `... patch_name= ...`. Arm 8c pins the wording, then makes the call that proves the finding — `ResolvePatchModFolder("GhostRider", "GhostRider", ...)` returns the byte-identical refusal — and then follows the corrected remedy for real: `into` dropped creates `houseCARL - GhostRider` fresh, and the arm removes it again so the later arms' inventory count is unchanged. That also retires your note on 8b proving a call the sentence did not describe: the sentence now describes it.

## Counts

Three findings, three folded, none refuted. Two probe checks added to arm 8c (the identical-refusal loop and the followed remedy), two to 9f (the new scan-failed wording and the removal-lane render), one to 9g (the tool names against the constants); arm 9d's cap assertion updated to the new fresh clause. No new xUnit tests — this is the resolver's refusal wording end to end, which `extend-resolve-guard` already owns against a synthetic MO2 instance.

PR #553, opened 2026-09-05
Read a cell's child records as the additive union the game assembles

### LINKED ISSUE #342: Cell.Persistent/Temporary reads only the winning override's own list, not the additive union real cell references use

### LINKED ISSUE #354: [GAP] no targeted per-plugin record fetch — GetRecord enumerates a whole overlay per call

### LINKED ISSUE #487: [GAP] "What is live in this cell": Persistent/Temporary returns the winning override own list, not the engine additive union — #342 stage-2, split out

Avick3110:
A cell's `Persistent`/`Temporary`, a topic's `Responses` and a worldspace's `SubCells` hold records that are declared per plugin, and the engine assembles the parent's from every plugin that declares any. Reading only the winner's own list therefore reported an empty cell the game fills with hundreds of references whenever the winning override existed for an unrelated reason — an occlusion or lighting patch. The read now states the additive union beside the value: every distinct child the whole order declares, keyed by FormID so a child two overrides both declare counts once rather than twice, with each contributing plugin's own count and how much of the union the read body carries. The field's value is deliberately unchanged — it is still that body's own list in its own order, because those are the indices a write addresses — and the union is labelled as a separate quantity so the two cannot be confused. A singular owned child (`Cell.Landscape`, `Worldspace.TopCell`) is not a union, since its declarers override one record, so it says which plugin's copy is live instead of adding counts that would be a fiction. A plugin whose body could not be read is named beside the union, never absorbed into it. In json the union also carries the member FormIDs, so a caller reads them straight back through `formids=`. The union claims declaration, not liveness: whether a member's own winner is deleted or initially disabled is a fact about that child record and asserting it here would cost one fetch per member.

#354 was needed and is done. The union needs one body per touching plugin, and `GetRecord` fetched a body by scanning every record in the plugin, which is the cost that made the earlier index-only tier the only affordable answer (588 ms for one Dawnstar cell on a real order). `GetRecord` and the conflict tree now take the record's getter type and use Mutagen's typed group enumeration, falling back to the flat scan when the typed walk yields nothing so a type Mutagen does not route cannot read as absent. Measured on a #354-shaped synthetic order (one 20k-record master, nine 2k plugins, ten touchers of one cell whose union is 209 references): the ten bodies cost 80 ms fetched flat and 4 ms seeked by type; the whole union read answers in 3.3 ms against 1.0 ms for the same read projecting `EditorID` alone. The union is also computed only for child-bearing fields the read actually emitted, so a projection naming none pays nothing.

New engine logic is `src/housecarl-core/OwnedChildUnion.cs`; `WriteEngine`'s child-reachability walk gained a sibling that answers *which* record type a field owns (needed so a worldspace block enumerates to its cells rather than to everything inside them), leaving the write-side property set untouched. `OwnedChildContentProbe`'s arm pinning that the default read could not open a body is gone — that is the behaviour this PR reverses — and the rendered coverage is in `RecordsOwnedChildTests`, including the overlapping-declarer case a naive concatenation would double-count.

Closes #342
Closes #487
Closes #354

### DIFF
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -261,6 +261,12 @@ public sealed class LoadOrderResolver : IDisposable
     /// keeps results valid by holding the session open until it has materialised what it returns (the service reads
     /// fields off a fetched body before its session disposes; the write path keeps the source body + link cache valid
     /// through serialize).</summary>
+    /// <summary>How many overlay OPENS sessions have paid in this process, counted where one actually happens.
+    /// Whether a call opened a plugin once or once per row is invisible in every answer — the bytes are the same,
+    /// only the clock differs — so this is the one thing a test can hold a session-reuse claim to. Read-only to
+    /// everything but <see cref="OverlaySession.Overlay"/>.</summary>
+    internal static long SessionOverlayOpens;
+
     public sealed class OverlaySession : IDisposable
     {
         readonly LoadOrderResolver _r;
@@ -272,7 +278,10 @@ public sealed class LoadOrderResolver : IDisposable
         internal ISkyrimModGetter Overlay(int idx)
         {
             if (!_open.TryGetValue(idx, out var ov))
+            {
+                System.Threading.Interlocked.Increment(ref SessionOverlayOpens);
                 _open[idx] = ov = OpenOverlay(_r._paths[idx], _r._dataDir);
+            }
             return ov;
         }
 
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -876,8 +876,8 @@ public sealed class LoadOrderResolver : IDisposable
         /// excluded-plugin check judged against THIS view's build — so a winner this view resolved and the body
         /// fetched for it can never be vetted by two different builds. The write path resolves every edit of one call
         /// off ONE view; reads pin their fetch to the same view they resolved with.</summary>
-        public IMajorRecordGetter? GetRecord(OverlaySession session, string pluginName, FormKey fk)
-            => _r.GetRecord(session, pluginName, fk, _s);
+        public IMajorRecordGetter? GetRecord(OverlaySession session, string pluginName, FormKey fk, Type? getterType = null)
+            => _r.GetRecord(session, pluginName, fk, _s, getterType);
 
         /// <summary>The master filenames a plugin DECLARES in its header (the master table), in declared order — opens
         /// the overlay, reads the header, disposes — the header parses without enumerating records. Throws
@@ -920,11 +920,15 @@ public sealed class LoadOrderResolver : IDisposable
         var overlayIdxs = e.count == 1 ? new[] { e.winner } : s.Overriders[fk];
         var nodes = new ConflictNode[overlayIdxs.Length];
         string? recType = null;
+        // The FIRST body is fetched blind; every one after it seeks by the type that body turned out to be, which is
+        // the same record's type in every provider. On a cell that turns N flat scans into one plus N-1 typed seeks.
+        Type? getterType = null;
         for (int n = 0; n < overlayIdxs.Length; n++)
         {
             int oi = overlayIdxs[n];
-            var rec = FetchBody(session, oi, fk);
+            var rec = FetchBody(session, oi, fk, getterType);
             recType ??= RecordNaming.StripOverlay(rec.GetType().Name);
+            getterType ??= WriteEngine.PrimaryGetter(rec.GetType());
             nodes[n] = new ConflictNode(_names[oi], rec);
         }
         return new ConflictTree(fk, recType ?? "?", nodes);
@@ -962,14 +966,29 @@ public sealed class LoadOrderResolver : IDisposable
     /// the private <see cref="FetchBody"/> (which throws on an index inconsistency). The server's read_record uses this
     /// for an explicit-plugin read, and for the winner's body off <see cref="ResolveWinner"/>. The returned body is
     /// backed by the session's overlay — read it before the session disposes.</summary>
-    public IMajorRecordGetter? GetRecord(OverlaySession session, string pluginName, FormKey fk)
-        => GetRecord(session, pluginName, fk, _snap);                      // single-shot: this call = its own build (the IndexView overload pins a whole operation)
+    public IMajorRecordGetter? GetRecord(OverlaySession session, string pluginName, FormKey fk, Type? getterType = null)
+        => GetRecord(session, pluginName, fk, _snap, getterType);          // single-shot: this call = its own build (the IndexView overload pins a whole operation)
 
-    IMajorRecordGetter? GetRecord(OverlaySession session, string pluginName, FormKey fk, IndexSnapshot s)
+    IMajorRecordGetter? GetRecord(OverlaySession session, string pluginName, FormKey fk, IndexSnapshot s, Type? getterType)
     {
         if (!_nameToIdx.TryGetValue(pluginName, out int idx)) return null;
         if (s.Excluded.Contains(idx)) return null;                         // excluded plugin — never re-enumerate it (would re-throw); the service reports the reason via ExcludedPlugins
-        foreach (var rec in session.Overlay(idx).EnumerateMajorRecords())
+        return SeekBody(session.Overlay(idx), fk, getterType);
+    }
+
+    /// <summary>Find one record in one open overlay. With <paramref name="getterType"/> in hand this is a TYPED
+    /// group seek: Mutagen walks only the GRUPs that can hold that type, so finding a cell does not step over the
+    /// placed references that outnumber it fifty to one (#354 — the cost that shaped the owned-child read's two
+    /// tiers). Without a type it is the flat scan, unchanged.
+    /// <para>A typed walk that yields nothing falls back to the flat one. Mutagen's typed enumeration is a switch
+    /// over the types it models, and a type it does not route would otherwise turn "this plugin has the record"
+    /// into a silent "it does not" — the fallback costs a second pass on a miss and never a wrong answer.</para></summary>
+    static IMajorRecordGetter? SeekBody(ISkyrimModGetter ov, FormKey fk, Type? getterType)
+    {
+        if (getterType is not null)
+            foreach (var rec in ov.EnumerateMajorRecords(getterType, throwIfUnknown: false))
+                if (rec.FormKey == fk) return rec;
+        foreach (var rec in ov.EnumerateMajorRecords())
             if (rec.FormKey == fk) return rec;
         return null;
     }
@@ -1000,10 +1019,9 @@ public sealed class LoadOrderResolver : IDisposable
 
     /// <summary>Fetch one record body from one overlay by re-enumerating it into the session. Throws if the overlay
     /// can't yield a FormKey the index says it contains — a real inconsistency, named rather than swallowed.</summary>
-    IMajorRecordGetter FetchBody(OverlaySession session, int overlayIdx, FormKey fk)
+    IMajorRecordGetter FetchBody(OverlaySession session, int overlayIdx, FormKey fk, Type? getterType = null)
     {
-        foreach (var rec in session.Overlay(overlayIdx).EnumerateMajorRecords())
-            if (rec.FormKey == fk) return rec;
+        if (SeekBody(session.Overlay(overlayIdx), fk, getterType) is { } rec) return rec;
         throw new InvalidOperationException(
             $"body-fetch inconsistency: {_names[overlayIdx]} is indexed as containing {fk} but did not yield it on re-enumeration.");
     }
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -928,7 +928,7 @@ public sealed class LoadOrderResolver : IDisposable
             int oi = overlayIdxs[n];
             var rec = FetchBody(session, oi, fk, getterType);
             recType ??= RecordNaming.StripOverlay(rec.GetType().Name);
-            getterType ??= WriteEngine.PrimaryGetter(rec.GetType());
+            getterType ??= WriteEngine.SeekTypeFor(rec);
             nodes[n] = new ConflictNode(_names[oi], rec);
         }
         return new ConflictTree(fk, recType ?? "?", nodes);
--- a/src/housecarl-core/OwnedChildUnion.cs
+++ b/src/housecarl-core/OwnedChildUnion.cs
@@ -31,9 +31,6 @@ public sealed record ChildUnion(
     /// <summary>Distinct child records the order declares for this field.</summary>
     public int Total => Members.Count;
 
-    /// <summary>Does the body the read was taken from carry less than the whole? The #342 case in one predicate:
-    /// the winner touched the parent for an unrelated reason and its own list reads short or empty.</summary>
-    public bool OwnIsPartial => OwnCount < Total;
 }
 
 /// <summary>
@@ -65,7 +62,7 @@ public static class OwnedChildUnion
         var touching = view.TouchingPlugins(fk);
         if (touching is null || touching.Count <= 1) return null;
 
-        var getterType = WriteEngine.PrimaryGetter(subjectBody.GetType());
+        var getterType = WriteEngine.SeekTypeFor(subjectBody);
         var members = new Dictionary<string, List<FormKey>>(StringComparer.Ordinal);
         var seen = new Dictionary<string, HashSet<FormKey>>(StringComparer.Ordinal);
         var declarers = new Dictionary<string, List<ChildUnionDeclarer>>(StringComparer.Ordinal);
@@ -84,9 +81,15 @@ public static class OwnedChildUnion
         // shape's live copy is simply the last declarer standing.
         foreach (var plugin in touching)
         {
-            var body = string.Equals(plugin, subjectPlugin, StringComparison.OrdinalIgnoreCase)
-                ? subjectBody
-                : view.GetRecord(session, plugin, fk, getterType);
+            // A sibling plugin that will not OPEN must not fault the read. The subject's own body is already in
+            // hand, so the answer the caller asked for survives; this plugin becomes an unreadable the note names.
+            // The transient case is the common one — xEdit or MO2 holding an exclusive handle, an AV scan — and
+            // before the union the default read opened only the source, so nothing here could ever throw.
+            IMajorRecordGetter? body;
+            if (string.Equals(plugin, subjectPlugin, StringComparison.OrdinalIgnoreCase)) body = subjectBody;
+            else
+                try { body = view.GetRecord(session, plugin, fk, getterType); }
+                catch { body = null; }
             foreach (var f in fields.Keys)
             {
                 // Null is "could not look", never "declares nothing" — a provider counted into the negative would
@@ -143,7 +146,14 @@ public static class OwnedChildUnion
         // TYPE this field owns, so the walk stops at the cells rather than descending into their contents.
         if (val is IMajorRecordGetterEnumerable en && childType is not null)
         {
+            int before = sink.Count;
             foreach (var child in en.EnumerateMajorRecords(childType, throwIfUnknown: false)) sink.Add(child.FormKey);
+            // A typed enumeration Mutagen does not route yields an EMPTY sequence rather than throwing, which would
+            // turn "I could not look" into "it declares nothing" — the #308 rule inverted. So an empty typed walk
+            // over a container that holds records at all is a miss, not a negative: fail closed and let the caller
+            // say the body could not be read. The untyped check runs only on that empty branch and stops at the
+            // first record it finds, so the walk this arm exists to avoid is never paid on the normal path.
+            if (sink.Count == before && en.EnumerateMajorRecords().Any()) return false;
             return true;
         }
         if (val is System.Collections.IEnumerable seq)
--- a/src/housecarl-core/OwnedChildUnion.cs
+++ b/src/housecarl-core/OwnedChildUnion.cs
@@ -1,0 +1,157 @@
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Records;
+
+namespace HousecarlCore;
+
+/// <summary>One plugin's contribution to a field's additive union: how many child records ITS body declares there.
+/// A child two plugins both declare is counted by both — the counts are contributions, not a partition, which is
+/// why <see cref="ChildUnion.Total"/> is carried rather than summed from these.</summary>
+public sealed record ChildUnionDeclarer(string Plugin, int Count);
+
+/// <summary>What the game assembles for ONE child-bearing field of ONE record, across every plugin that touches it.
+///
+/// <para>For a <see cref="OwnedChildShape.Collection"/> field the answer is a UNION keyed by FormID:
+/// <see cref="Members"/> is every distinct child the order declares, in load order of first declaration, each
+/// counted once however many plugins declare it. For a <see cref="OwnedChildShape.Singular"/> field there is no
+/// union — the declarers override one record — so <see cref="Members"/> is the one live child and
+/// <see cref="LivePlugin"/> is the highest plugin declaring it.</para>
+///
+/// <para><b>What it claims is DECLARATION, not liveness.</b> A child in this set is one some plugin declares for
+/// this parent. Whether that child's own winner is deleted or initially disabled is a fact about the child record,
+/// readable by reading it, and asserting it here would cost one fetch per member.</para></summary>
+public sealed record ChildUnion(
+    string Field,
+    OwnedChildShape Shape,
+    IReadOnlyList<FormKey> Members,
+    int OwnCount,
+    IReadOnlyList<ChildUnionDeclarer> Declarers,
+    IReadOnlyList<string> Unreadable,
+    string? LivePlugin)
+{
+    /// <summary>Distinct child records the order declares for this field.</summary>
+    public int Total => Members.Count;
+
+    /// <summary>Does the body the read was taken from carry less than the whole? The #342 case in one predicate:
+    /// the winner touched the parent for an unrelated reason and its own list reads short or empty.</summary>
+    public bool OwnIsPartial => OwnCount < Total;
+}
+
+/// <summary>
+/// The additive union a child-bearing field really holds (#342 / #487).
+///
+/// <para>An owned child record is declared PER PLUGIN, and for a collection field the game assembles the parent's
+/// children from every plugin that declares any. A cell override that exists to add occlusion data carries no
+/// placed references and deletes none, so reading its <c>Temporary</c> reports an empty cell the game fills with
+/// hundreds — the read model's one silent wrong answer. This computes what the engine assembles instead: the
+/// FormID-keyed union over every touching plugin's own body, so a child two overrides both declare is counted
+/// once rather than twice.</para>
+///
+/// <para><b>Cost.</b> One body per touching plugin, seeked by the record's own type
+/// (<see cref="LoadOrderResolver.IndexView.GetRecord"/> takes it, #354) so finding a cell in a worldspace plugin
+/// does not step over the placed references that outnumber it. Nothing is held: the bodies are read through the
+/// caller's open session and only FormKeys and counts survive the call.</para>
+/// </summary>
+public static class OwnedChildUnion
+{
+    /// <summary>The additive union of each named child-bearing field of <paramref name="fk"/>, or NULL when the
+    /// record has one toucher — its own body is then the whole story and there is nothing to assemble.
+    /// <paramref name="subjectBody"/> is the body the read was taken from (the winner, or a <c>plugin=</c>-scoped
+    /// copy); it is reused rather than re-fetched, and it is what <see cref="ChildUnion.OwnCount"/> is about.</summary>
+    public static IReadOnlyDictionary<string, ChildUnion>? Compute(
+        LoadOrderResolver.IndexView view, LoadOrderResolver.OverlaySession session, FormKey fk,
+        string subjectPlugin, IMajorRecordGetter subjectBody, IReadOnlyDictionary<string, OwnedChildShape> fields)
+    {
+        if (fields.Count == 0) return null;
+        var touching = view.TouchingPlugins(fk);
+        if (touching is null || touching.Count <= 1) return null;
+
+        var getterType = WriteEngine.PrimaryGetter(subjectBody.GetType());
+        var members = new Dictionary<string, List<FormKey>>(StringComparer.Ordinal);
+        var seen = new Dictionary<string, HashSet<FormKey>>(StringComparer.Ordinal);
+        var declarers = new Dictionary<string, List<ChildUnionDeclarer>>(StringComparer.Ordinal);
+        var unreadable = new Dictionary<string, List<string>>(StringComparer.Ordinal);
+        var live = new Dictionary<string, string?>(StringComparer.Ordinal);
+        var liveKeys = new Dictionary<string, IReadOnlyList<FormKey>>(StringComparer.Ordinal);
+        var own = new Dictionary<string, int>(StringComparer.Ordinal);
+        foreach (var f in fields.Keys)
+        {
+            members[f] = new List<FormKey>(); seen[f] = new HashSet<FormKey>();
+            declarers[f] = new List<ChildUnionDeclarer>(); unreadable[f] = new List<string>();
+            live[f] = null; liveKeys[f] = Array.Empty<FormKey>(); own[f] = 0;
+        }
+
+        // Priority order, low to high: the union's member order is first-declaration order, and the SINGULAR
+        // shape's live copy is simply the last declarer standing.
+        foreach (var plugin in touching)
+        {
+            var body = string.Equals(plugin, subjectPlugin, StringComparison.OrdinalIgnoreCase)
+                ? subjectBody
+                : view.GetRecord(session, plugin, fk, getterType);
+            foreach (var f in fields.Keys)
+            {
+                // Null is "could not look", never "declares nothing" — a provider counted into the negative would
+                // turn an unreadable body into evidence the field is empty (#308's rule, one level down).
+                var keys = body is null ? null : ChildKeys(body, f);
+                if (keys is null) { unreadable[f].Add(plugin); continue; }
+                if (keys.Count == 0) continue;
+                declarers[f].Add(new ChildUnionDeclarer(plugin, keys.Count));
+                live[f] = plugin;
+                liveKeys[f] = keys;
+                if (ReferenceEquals(body, subjectBody)) own[f] = keys.Count;
+                foreach (var k in keys)
+                    if (seen[f].Add(k)) members[f].Add(k);
+            }
+        }
+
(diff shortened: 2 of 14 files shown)


src/housecarl-core/OwnedChildUnion.cs
```diff
@@ -0,0 +1,157 @@
...
+        foreach (var f in fields.Keys)
+        {
+            members[f] = new List<FormKey>(); seen[f] = new HashSet<FormKey>();
+            declarers[f] = new List<ChildUnionDeclarer>(); unreadable[f] = new List<string>();
+            live[f] = null; liveKeys[f] = Array.Empty<FormKey>(); own[f] = 0;
+        }
+
+        // Priority order, low to high: the union's member order is first-declaration order, and the SINGULAR
+        // shape's live copy is simply the last declarer standing.
+        foreach (var plugin in touching)
```
Avick3110, line 88: An overlay that throws here faults the whole read, and the `Unreadable` arm below it is unreachable.

`view.GetRecord(session, plugin, fk, getterType)` reaches `OverlaySession.Overlay(idx)` -> `OpenOverlay(...)`. Nothing in `Compute` catches; `ChildKeys`'s try/catch is inside the callee and never sees this. Before this PR the default read opened exactly one plugin (the source), so an open failure on a *sibling* plugin could not affect it. Now: index builds cleanly, then MO2/xEdit/an AV scan takes an exclusive handle on one of the ten plugins touching a Dawnstar cell -> `records formids=<cell>` throws out of `ResolveRead` instead of returning the record. The epoch doc calls that exact transient case out as expected ("xEdit or MO2 holding an exclusive handle, an AV scan").

Second half: `body is null` on line 94 is only reachable via `!_nameToIdx.TryGetValue(...)` (impossible - the names came from `_names`) or `s.Excluded.Contains(idx)`. But `Exclude()` drops every one of a plugin's index entries, so an excluded plugin never appears in `TouchingPlugins(fk)`. So `Unreadable` is always empty, the `unreadable` json array and the `; N plugin(s) could NOT be read (...)` tail in `UnionNote` are dead, and the PR body's claim that "a plugin whose body could not be read is named beside the union" is not something this code can deliver. Wrapping the fetch in a catch that feeds `unreadable[f]` would make the claim true and fix the fault above in one move.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2400,24 +2400,25 @@ static RecordFields AnnotateOwnedChildContent(RecordFields rf, IMajorRecordGette
             if (owning.ContainsKey(rf.Fields[i].Path)) (hits ??= new List<int>()).Add(i);
         if (hits is null) return rf;
 
-        var touching = view.TouchingPlugins(fk);
-        if (touching is null || touching.Count <= 1) return rf;   // sole toucher: its own body IS the whole story
+        // Narrowed to the fields this read emitted: the union opens a body per touching plugin, and assembling a
+        // worldspace's cells for a read that asked for EditorID would be a cost nobody asked for.
+        var wanted = new Dictionary<string, OwnedChildShape>(hits.Count, StringComparer.Ordinal);
+        foreach (var i in hits) wanted[rf.Fields[i].Path] = owning[rf.Fields[i].Path];
+        var unions = OwnedChildUnion.Compute(view, session, fk, source, body, wanted);
```
Avick3110: This turns a per-record index lookup into one body fetch per touching plugin, on lanes that run it per row.

`ResolveRead` is shared by the single read, `batch_record_detail` (line 2738), the `cross_plugin_query` detail path (`ResolveReadOn`, line 2516) and the dense grid (`JsonWire` line ~1167). None of them memoise across rows, and each `ResolveRead` opens its own `OverlaySession`, so the overlays are re-mmapped per row. A `cross_plugin_query` over CELL with the default projection (no `fields=`) emits `Persistent`/`Temporary`, so N result rows x K touching plugins full typed-group seeks, each seek walking that plugin's whole cell tree. On a real order Skyrim.esm has tens of thousands of cells, and the measurement in the PR body is against a 20k-record synthetic master, which does not exercise this.

`Worldspace.SubCells` is worse, and the member cap does not help: `ChildKeys` calls `p.GetValue(body)` and then `EnumerateMajorRecords(ICellGetter)` per block, materialising every exterior cell of *every* touching plugin. `ChildUnionMemberCap` caps the output, not the work - so reading Tamriel WRLD now does that walk across the whole order.

Worth either memoising the union per (fk, plugin-set) across a bulk call, or gating the union to the single-record lane and leaving batch/dense on the index-only note.

src/housecarl-core/OwnedChildUnion.cs
```diff
@@ -0,0 +1,157 @@
...
+    static bool Collect(object? val, Type? childType, List<FormKey> sink, int depth)
+    {
+        if (val is null) return true;                            // absent optional / empty — declares nothing
+        if (depth > MaxDepth) return false;                      // nested deeper than any known shape — unknown, not "no"
+        if (val is IFormLinkGetter) return true;                 // a reference, not a child (the type walk's own cut)
+        if (val is IMajorRecordGetter rec) { sink.Add(rec.FormKey); return true; }
+        if (val is string) return true;
+        // A non-record container that knows its own containment (a worldspace block): ask Mutagen for the child
+        // TYPE this field owns, so the walk stops at the cells rather than descending into their contents.
+        if (val is IMajorRecordGetterEnumerable en && childType is not null)
```
Avick3110, line 156: This branch reports "declares nothing" when it should report "could not look".

`EnumerateMajorRecords(childType, throwIfUnknown: false)` returns an empty sequence for a type Mutagen's containment switch does not route - and this arm then `return true` with an empty `sink`, which `Compute` reads as a plugin that declares no children. That is exactly the #308 rule the file's own comment on line 93 invokes, inverted. Concretely: a Mutagen bump (or a container whose owned type resolves to an abstract getter interface, which `OwnedRecordTypeOf` will happily return since it stops at the first `IMajorRecordGetter`-assignable type it reaches) makes a worldspace read say `additive union: no plugin touching this record declares child records here` for a worldspace full of cells - a silently wrong answer, not a degraded one.

`SeekBody` in `LoadOrderResolver` guards the same Mutagen behaviour with a flat-scan fallback and says why. This walk has no equivalent. Returning `false` (i.e. null keys -> unreadable) when the typed enumeration yields nothing but the container is non-empty would keep it fail-closed.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -278,6 +279,44 @@ static void WriteFieldsArray(Utf8JsonWriter w, RecordFields r, MemoryStream ms,
...
+    /// the same <c>formids=</c> door it came in by. <c>members</c> is capped at
+    /// <see cref="ChildUnionMemberCap"/> with <c>members_omitted</c> naming the rest, because a worldspace's union
+    /// runs to tens of thousands and a field annotation is not a listing surface.</summary>
+    static void WriteChildUnion(Utf8JsonWriter w, ChildUnion u)
+    {
+        w.WriteStartObject("owned_child_union");
+        w.WriteString("shape", u.Shape.ToString());
+        w.WriteNumber("total", u.Total);
+        w.WriteNumber("own", u.OwnCount);
+        WriteNullable(w, "live_plugin", u.LivePlugin);
```
Avick3110: `live_plugin` is written for `Collection` unions too, where it is a false claim.

For a `Collection` shape `Compute` sets `live[f]` to the last plugin in priority order that declared *any* child - but an additive union has no single live declarer; every declarer's children are live. A json consumer reading `{"shape":"Collection","total":209,"live_plugin":"Occlusion.esp"}` reasonably concludes Occlusion.esp's list is the live one, which is the exact #342 misreading this PR exists to remove. The prose `UnionNote` is careful here (it only names a live plugin on the `Singular` arm); the json is not. Either omit the key unless `u.Shape == Singular`, or rename it for the collection case (e.g. `highest_declarer`).

src/housecarl-mcp/JsonWire.cs
```diff
@@ -260,6 +260,7 @@ static void WriteFieldsArray(Utf8JsonWriter w, RecordFields r, MemoryStream ms,
             if (f.HasValue) w.WriteString("value", f.Token);   // round-trip parity: identical token to the text render
             else WriteNullable(w, "note", f.Note);
             if (f.Display is not null) w.WriteString("display", f.Display);
+            if (annotated is not null && annotated.TryGetValue(f.Path, out var union)) WriteChildUnion(w, union);
```
Avick3110: A union field object can overshoot `max_chars` by a couple of KB with no `truncated` signal.

`WriteFieldsArray` only checks `ms.Length >= cap` *before* writing each field. Field objects used to be small, so the overshoot was one short line. `WriteChildUnion` now adds up to `ChildUnionMemberCap` (100) FormKey strings at ~28 bytes each plus the declarer array - ~2.5-3KB for one field. With a small `max_chars` (the text lane tests here use 1400) the json response can more than double the cap before the loop notices on the next iteration, and `ClauseReserve` reserves nothing for it. The auto-spill trigger reads the truncated flag, so the overrun is invisible to it. Either check the budget after writing the union, or scale the member cap to the remaining budget.

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -920,11 +920,15 @@ public bool IsLightFlagged(string pluginName)
...
+        // The FIRST body is fetched blind; every one after it seeks by the type that body turned out to be, which is
+        // the same record's type in every provider. On a cell that turns N flat scans into one plus N-1 typed seeks.
+        Type? getterType = null;
         for (int n = 0; n < overlayIdxs.Length; n++)
         {
             int oi = overlayIdxs[n];
-            var rec = FetchBody(session, oi, fk);
+            var rec = FetchBody(session, oi, fk, getterType);
             recType ??= RecordNaming.StripOverlay(rec.GetType().Name);
+            getterType ??= WriteEngine.PrimaryGetter(rec.GetType());
```
Avick3110: `PrimaryGetter` is the type this file already documents as the wrong one to hand Mutagen's typed routing.

`RemovalTypeFor` (WriteEngine) exists because passing a concrete subclass getter of an abstract-group `T` - `IGlobalShortGetter` under `SkyrimGroup<Global>` - makes Mutagen's typed routing silently no-op even with `throwIfUnknown: true`. `PrimaryGetter` returns exactly that subclass interface. So for GLOB/GMST (and any other abstract-base group) every body after the first pays an empty typed walk *and* the full flat scan - strictly slower than before this change, on the conflict tree lane that is the hottest user of `FetchBody`. `SeekBody`'s fallback keeps it correct, so this is cost only, but reusing `RemovalTypeFor`'s flat-group resolution would make the optimisation apply to those types instead of penalising them.

Avick3110:
Four more, on top of the six inline ones — docs, dead code and test reality rather than code paths.

**`docs/architecture/records-owned-child-declarers.md:36-44` — medium. The measurement that justifies making the union the default does not cover the shapes that caused the withdrawal.**
The old text withdrew the per-toucher walk on five numbers: 27ms→588ms for a Dawnstar cell, 21ms→1.3s for Tamriel, *a worldspace read to 2.5s*, *6.3s→126s for a 200-cell query*, and *an unbounded artifact job that never finished*. The new text keeps the first two and drops the last three, then argues the typed seek "is the ceiling those numbers were paying" from one synthetic single-cell read (209 members, one 20k master). The three dropped cases are exactly the ones the typed seek does not help with: the seek finds the parent cheaply, but `ChildKeys` still materialises every child of every toucher, and the 200-cell query and the artifact job multiply that by N rows (see the inline finding on `LoadOrderService.cs:2407`). Either re-measure a worldspace read and a 200-cell query on the same synthetic, or say plainly that those shapes are unmeasured.

**`plugin/CHANGELOG.md:23-24` — medium. The json claim omits the cap.**
"In json the union also carries the member FormIDs, so they can be read straight back through `formids=`" — but `JsonWire.ChildUnionMemberCap` is 100, and the doc's own example union is 209. A caller who takes the changelog at its word round-trips a little under half of a Dawnstar cell and gets no signal in the prose that they did. `members_omitted` is in the payload; the user-facing sentence should say the array is a sample.

**`src/housecarl-core/OwnedChildUnion.cs:36` — low. `ChildUnion.OwnIsPartial` is dead.**
Declared and documented ("the #342 case in one predicate") and referenced by nothing — not `ReadSentences.UnionNote`, not `JsonWire.WriteChildUnion`, not a test. The #342 case is instead re-derived as prose in `UnionNote` (`this body's own list carries {OwnCount}` against `Total`). Either use it or drop it; a documented predicate nothing reads is a claim with no owner.

**Test reality — two arms of the new shape are unpinned.**
- No test exercises `ChildUnionMemberCap`. `RecordsOwnedChildTests` asserts a 3-member and a 4-member union; nothing builds one over 100, so the cap branch and `members_omitted` ship untested.
- No test exercises `Unreadable` / `ReadSentences.CouldNotRead`. Per the inline finding on `OwnedChildUnion.cs:85` that path may be unreachable by construction — which is itself the thing a test would have caught, since the PR body states "a plugin whose body could not be read is named beside the union" as delivered behaviour.

The rest of the claimed surface holds up on reading: the FormID-keyed dedup is genuinely keyed (`seen[f].Add(k)`), `TouchingPlugins` really is low-to-high so the member order and the `Singular` live-copy pick are both right, inactive plugins never enter the index so the union can't count them, the field value is untouched (`Display` only) and pinned by `TheUnionNeverReplacesTheBodysOwnList`, the worldspace walk stops at `ICellGetter` rather than sweeping cell contents, and nothing survives the call but `FormKey`s and counts — no overlay or plugin handle is held at rest.

Avick3110:
Review fold. All ten findings were real; nothing is refuted. Six commits, `cb33509..90a4468`. Build, tests (1175, +8) and `ci-all` (128/128) green.

**1 — high, `OwnedChildUnion.cs:85`: a locked sibling faulted the read, and `Unreadable` was dead.** Folded in `c036b91`. The sibling fetch is wrapped, and the plugin becomes an unreadable the note names, which is what makes the arm reachable at all. Pinned by a test that takes an exclusive handle on the mid plugin for the duration of one read, and a second that shows the same read names nothing once the handle is gone.

**2 — high, `LoadOrderService.cs:2407`: the union ran per row on the bulk lanes.** Folded in `849c7a4`. The union now runs only where the caller NAMED the records — the single read, `batch_record_detail`, and the `records` source pole — with one `ChildUnionMemo` per call so a formid named twice assembles once. **What the bulk lanes do now:** the scan lanes (`cross_plugin_query` detail rows, the dense grid, and the artifact spill of either) state the index-only note again — other plugins touch this record, their declarations for this field were not read — with a response clause naming the `formids=` lane that does assemble it. They still annotate, because the field is a false-empty either way; they just never claim a quantity they did not compute. The reported #342 repro is `batch_record_detail`, which is on the union side of that line. Two tests pin the split.

**3 — medium, `OwnedChildUnion.cs:144`: an unrouted `childType` read as "declares nothing".** Folded in `c036b91`. An empty typed enumeration over a container that holds records at all now returns false (could not look) rather than true with an empty sink. The untyped check runs only on that empty branch and stops at the first record, so the walk the arm exists to avoid is never paid on the normal path.

**4 — medium, `JsonWire.cs:293`: `live_plugin` on a Collection union.** Folded in `7943838`. `live_plugin` is written only on the Singular shape; a Collection carries `highest_declarer` instead. Pinned both ways in one test.

**5 — medium, `JsonWire.cs:263`: the union block could overshoot `max_chars`.** Folded in `7943838`. The member array is bounded by what is left of the cap as well as by the flat 100, with `members_omitted` counting whatever was not listed either way. Tested at `max_chars=1400`, where the sample shrinks and the omission count still adds up to the true total.

**6 — medium, the doc dropped three of the withdrawal's five measurements.** Folded in `90a4468`. All five are addressed, off a synthetic order built for this: one 19.2k-record master and nine ~2k plugins, a worldspace of 200 exterior cells each carrying 40 references, every plugin overriding the worldspace and every cell while declaring none — so every cell has ten touchers.

| shape | without the union | with |
|---|---|---|
| one cell, `EditorID` / `Temporary` | 1.9 ms | 5.7 ms |
| the worldspace, `EditorID` / `SubCells` (union of 200 cells) | 1.8 ms | 8.8 ms |
| 200 cells named by `formids=` | — | 1.04 s |
| the same 200 cells as a `types=["CELL"]` scan | 0.24 s (index-only) | not run |

The worldspace read is 8.8 ms, not seconds: the seek finds the parent without stepping over the references, and `ChildKeys` stops at the cells. The 200-cell query and the artifact job are answered by the lane split from finding 2 rather than by the seek. The doc labels all of it as synthetic and says what a real order adds.

**7 — medium, the CHANGELOG omitted the cap.** Folded in `90a4468`. The json line now says the members are a sample of up to 100 (fewer under a tight `max_chars`) with `members_omitted` counting the rest, and a second bullet says which lanes assemble the union.

**8 — low, `PrimaryGetter` penalises abstract-base groups.** Folded in `318efd1`. `WriteEngine.SeekTypeFor` answers with the flat group's getter interface where one matches — the same `EnumerateFlatGroups` resolution `RemovalTypeFor` uses — so GLOB/GMST get a working typed walk instead of an empty one plus the flat scan. Both the conflict tree and the union take it.

**9 — low, `OwnIsPartial` is dead.** Folded in `c036b91` — dropped. `UnionNote` states `OwnCount` against `Total`, which says more than the predicate did.

**10 — low, the cap branch and `Unreadable` are untested.** Folded in `abd040c`. The world gains a cell whose union is 101, one past the cap, and generates a corpus rulebook so the scan lanes can be driven from it. Eight new tests: the cap sample and `members_omitted`, the budget-shrunk sample, the unreadable arm and its release, `highest_declarer` vs `live_plugin`, and the scan lanes' tier on both transports.

Avick3110:
Blind review folded (10 of 10 real, 0 refuted). Queued for Aaron's review; the declaration-vs-liveness call is on this PR. Stacked follow-on: #569 (#251).

src/housecarl-mcp/ReadSentences.cs
```diff
@@ -17,34 +17,97 @@ namespace HousecarlMcp;
...
+            head = u.LivePlugin is null
+                ? $"no plugin touching this record declares this {ChildContent}"
+                : $"one {ChildContent}: {u.Declarers.Count} plugin(s) hold a copy and OVERRIDE each other; " +
+                  $"the live copy is {u.LivePlugin}'s";
+        else if (u.Total == 0) head = NoUnionMembers;
+        else
+            head = $"{UnionLabel}: {u.Total} {ChildContent}(s) across {u.Declarers.Count} plugin(s) — "
+                 + string.Join(", ", u.Declarers.Take(UnionDeclarerCap).Select(d => $"{d.Plugin} {d.Count}"))
+                 + (u.Declarers.Count > UnionDeclarerCap ? $" (+{u.Declarers.Count - UnionDeclarerCap} more)" : "")
+                 + $"; this body's own list carries {u.OwnCount}";
```
Avick3110: **`this body's own list carries N` counts a different unit than the value it sits beside (Worldspace.SubCells).**

Run against this PR's own fixture, `housecarl_records formids=[<Worldspace>]` renders:

```
SubCells = [list: 2 item(s)] - pass project.depth=2 to expand   (additive union: 3 child record(s) across 1 plugin(s) - HcOcBase.esm 3; this body's own list carries 0)
```

The winner declares 2 blocks holding 0 cells. `OwnCount` (and `Total`) count the CELLS two container levels down, which is exactly what `OwnedChildUnion.ChildKeys` is built to do; the rendered value counts the top-level list, i.e. the BLOCKS. So the line says `2 item(s)` and `own list carries 0` at once, and `UnionFraming` then tells the caller "The VALUE shown for such a field is this body's own list" - which is the one thing that is not true here.

`Cell.Persistent`/`Temporary` and `DialogTopic.Responses` are unaffected: their children are the list's own elements. `Worldspace.SubCells` is the only field whose child records sit below the level the value renders, so either the note needs to name its unit on that shape, or `OwnCount` needs to be stated as something other than "this body's own list".

src/housecarl-mcp-tests/RecordsOwnedChildTests.cs
```diff
@@ -181,6 +214,11 @@ public OwnedChildWorld()
         File.WriteAllText(Path.Combine(profiles, "plugins.txt"), "*" + BaseName + "\r\n*" + MidName + "\r\n*" + TopName + "\r\n");
         File.WriteAllText(Path.Combine(profiles, "modlist.txt"), "# header\r\n+TopMod\r\n+MidMod\r\n+BaseMod\r\n");
 
+        // The scan lanes validate against the corpus rulebook, so this world generates one like the others.
+        var genDir = Path.Combine(Root, "corpus-gen");
+        CorpusGenerator.GenerateAll(genDir, Path.Combine(Root, "corpus-ref"));
+        CorpusRulebook.CorpusPath = Path.Combine(genDir, "corpus.json");
```
Avick3110, line 228: **`OwnedChildWorld` repoints `CorpusRulebook.CorpusPath` without capturing or restoring it, and then deletes what it points at.**

`Dispose` deletes `Root`, and this corpus lives at `Root/corpus-gen/corpus.json`, so once the fixture is gone the process-global names a file that no longer exists.

Every other world here does the capture/restore pair - `RecordsWorld.cs:63`/`176`, `BulkRecordsWorld.cs:63`/`143` and `199`/`217`, `RuntimeFormIdTests.cs:396` - and `RecordsWorldLifecycleTests` exists for nothing else: "an unrestored static leaves everything after it resolving types against a path that no longer exists."

Concretely: any class that runs after `OwnedChildFixture` disposes and resolves `type=` through the rulebook without repointing it first loads from the deleted path. `CheckErrorsWorld` documents that it deliberately does not repoint. The suite passes today only because of the order the collections happen to run in.

```suggestion
        CorpusRulebook.CorpusPath = Path.Combine(genDir, "corpus.json");
```
(the fix is the pair the other worlds have: a `_priorCorpusPath` field set from `CorpusRulebook.CorpusPath` before this line, and `CorpusRulebook.CorpusPath = _priorCorpusPath;` in `Dispose` before the `Directory.Delete`.)

src/housecarl-core/OwnedChildUnion.cs
```diff
@@ -0,0 +1,167 @@
...
+        foreach (var plugin in touching)
+        {
+            // A sibling plugin that will not OPEN must not fault the read. The subject's own body is already in
+            // hand, so the answer the caller asked for survives; this plugin becomes an unreadable the note names.
+            // The transient case is the common one — xEdit or MO2 holding an exclusive handle, an AV scan — and
+            // before the union the default read opened only the source, so nothing here could ever throw.
+            IMajorRecordGetter? body;
+            if (string.Equals(plugin, subjectPlugin, StringComparison.OrdinalIgnoreCase)) body = subjectBody;
+            else
+                try { body = view.GetRecord(session, plugin, fk, getterType); }
```
Avick3110, line 97: **A `formids=` batch re-opens and re-seeks every toucher once per record.**

`ResolveRead` does `using var session = resolver.OpenSession()` per record, so the session that caches overlays dies with each record, and `ChildUnionMemo` dedupes only a repeated FormID, not a repeated plugin. `housecarl_records formids=[...100 exterior cells...]` therefore pays ~100 x (touchers) overlay opens plus 100 x (touchers) typed `ICellGetter` seeks, and each of those seeks walks the CELL group tree of whichever master it lands in.

The measurement in the doc is 1.04 s for 200 cells on a 19.2k-record synthetic master, and the doc says so honestly - but the numbers that got the earlier per-toucher walk withdrawn (27ms -> 588ms for a Dawnstar cell, 21ms -> 1.3s for Tamriel) came from a real order, where Skyrim.esm alone carries ~65k cells. The typed seek is a real 20x, but it is being asked to absorb a much larger corpus than the synthetic has, and this is the lane #354 was opened about.

Worth either one real-order measurement of the `formids=` batch shape before merge, or hoisting the `OverlaySession` to the batch so each toucher is opened once per call rather than once per record.

Avick3110:
Second review fold. All three findings were real; nothing is refuted. Three commits, `0b923ea..2fd9fec`. Build, tests (1181, +6) and `ci-all` (128/128) green.

**1 — `ReadSentences.cs:104`: `this body's own list carries N` counted a different unit than the value beside it.** Folded in `51c45f0`. Confirmed on the fixture: the worldspace winner declares two EMPTY blocks, so the line rendered `SubCells = [list: 2 item(s)]` next to `own list carries 0` — two units, one sentence.

The fix is the unit, answered structurally rather than by naming `SubCells` in prose. `OwnedChildContent.NestedFields` asks the property whether its ELEMENTS are the child records (a cell's `Temporary`, a topic's `Responses`) or containers holding them deeper (`Worldspace.SubCells`) — the element type off the same reflected child-bearing set the shape comes from, so a Mutagen bump that adds a nested field is picked up without an edit. `ChildUnion.Nested` carries it; the nested arm says `this body declares N of them — the value beside this counts the CONTAINERS holding them, not the child records themselves`, and json gets `"nested": true` for a consumer comparing `own`/`total` against `value` with no sentence to read. Flat fields keep the existing wording unchanged.

I left `UnionFraming` alone. Its "The VALUE shown for such a field is this body's own list" is literally true of `SubCells` — the value IS the body's own list, of blocks — and the contradiction was the two numbers sitting side by side unsaid, which is now said at the line where they sit. Three tests pin it: the nested note, the flat note still reading as a share of the value, and the json key present on one shape and absent on the other.

**2 — `RecordsOwnedChildTests.cs:220`: `OwnedChildWorld` repointed `CorpusRulebook.CorpusPath` without restoring it.** Folded in `0b923ea`. Correct on all counts — it was the one repointing site without the capture/restore pair, and `Dispose` deletes the directory the static was left naming. Added the pair, and generalised `RecordsWorldLifecycleTests` over a world factory so this world gets the same arm `RecordsWorld` has rather than the test existing for one world by accident.

**3 — `OwnedChildUnion.cs:91`: a `formids=` batch re-opened and re-seeked every toucher once per record.** Folded in `2fd9fec`. Verified: `ResolveRead` did `using var session = resolver.OpenSession()` per record, so the overlay cache died with each row while the union opened a body per toucher, and `ChildUnionMemo` dedupes a repeated FormID, not a repeated plugin.

Rather than pick between your two remedies I did both. `ResolveRead` now takes an optional session; the two lanes that read records the caller NAMED (`ResolveBatch` and the `records` source pole) open one for the call and hand it down. A single read still opens and closes its own, so nothing is held between calls. Measured on the same synthetic the doc's table uses, the 200-cell `formids=` batch:

| | overlay opens | wall |
|---|---|---|
| a session per record | 2,050 | 2.12 s |
| one per call | 10 | 1.37 s |

Opens now scale with the ORDER's plugin count instead of with rows, so the real-order shape you name — more touchers, bigger plugins — costs more only in the body walk itself, which is what the caller asked for. The remaining second is that walk. I did not get a real-order measurement; the doc says plainly that the numbers are synthetic, and that claim is unchanged.

Pinning a cost needed something observable, since session reuse returns identical bytes: `LoadOrderResolver.SessionOverlayOpens` counts opens where one happens, and the test asserts a two-record batch pays no more opens than the order has plugins. Checked it bites — 6 opens without the fix, 3 with.

No behaviour on this PR turns on the declaration-vs-liveness question, so nothing here pre-empts your call on it.

Issue #406, opened 2026-08-22
[Bug] read-cache freshness key is mtime-only — a same-mtime different-size edit is invisible


Avick3110:
Migrated from the retired `dev/BACKLOG.md` (advisor sweep, Aaron-go 2026-08-22); re-verified at `main` @ `a783b2f`.

`RefreshIfStale` (`LoadOrderResolver.cs:924-934`) compares only `SafeMtime(_paths[i]) != _mtimes[i]` (`SafeMtime` = `File.GetLastWriteTimeUtc`, `:936-939`); the epoch fingerprint (`ComputeEpoch`, `:564-576`) is likewise name|path|mtime.Ticks. No size component anywhere — so an edit that lands within mtime granularity (or a tool that restores mtime) serves stale parsed state silently. Strengthen the key to `(mtime, size)`. Originally deferred as HCBR item 4.1 PART-2; the known write-during-rebuild race is documented at `:561-563`.

**Routing:** row 8's W5 routing-call list, deliberately unmilestoned until the charter — read-honesty family (#353's neighborhood). Promotion condition: any observed real-world stale read.

Avick3110:
Routed 2026-09-03 (Aaron: "I agree with the rest of the routing proposals"; ground: the row-8 disposition walk, advisor lane)

**Rule-7(b) disposition: taken-into-W5**, riding the read-honesty engagement.

**Rule-10(b) tag: `engine-survives`** — site, quoted from the walk: "`LoadOrderResolver.cs:924-934` (`RefreshIfStale` compares `SafeMtime` only), `:936-939` (`SafeMtime` = `GetLastWriteTimeUtc`), `:564-576` (`ComputeEpoch` = name|path|mtime.Ticks); verified, no size term anywhere".

**Ground, quoted:** "this is a **read-honesty family member at engine level** — the epoch is a coverage claim, and a claim keyed on mtime alone silently over-claims freshness. The fix is a size term in two derived keys: hours, not design."

**The alternative the walk recorded, quoted:** "deferred — its own promotion condition (*'any observed real-world stale read'*) has **not fired**. **What decides:** whether the charter's derived-population sweep counts a freshness key as an accounting line. I read rule 10(g) as yes."

[...]

Issue #405, opened 2026-08-22
[Bug] nexus_check_updates renders REMOVED-category files as "— current"


Avick3110:
Migrated from the retired `dev/BACKLOG.md` (advisor sweep, Aaron-go 2026-08-22); re-verified at `main` @ `a783b2f`.

`IsSuperseded` (`NexusClient.cs:246`) is a closed set — `category is "OLD_VERSION" or "ARCHIVED"` — so every other raw category, REMOVED/DELETED included, falls through to `FileVerdict.Live` (`:292`) and the renderer appends `" — current"` (`NexusTools.cs:604`). A removed file renders `'name' v1 [REMOVED] — current` — self-contradictory in one line.

Mitigation on record: the raw category is always printed first (`:603`; doc note `:594` frames the catch-all as intentional forward-compat, "visible, not hidden") — so this is a contradiction, not a hidden state. The decision is whether REMOVED/DELETED join the retired set or get their own verdict.

**Routing:** row 8/W5 firm, beside the read-honesty family (#353/#396). Milestoned.

PR #554, opened 2026-09-05
Give the scripts findings family an off-order lane

### LINKED ISSUE #395: [GAP] the scripts findings family has no off-order lane, so a pre-enable verify sweep cannot check a fresh patch's script bindings

Avick3110:
The scripts family of `housecarl_check` refused a `plugins=` name that was not in the active load order, so the one artifact whose script bindings are most worth checking — a patch houseCARL has just written, not yet enabled — could not be checked at all without enabling it first. It now has the same off-order lane the errors family has: a named plugin the active order does not hold is located on disk (the shared `LocatePluginFileOnDisk` contract, ambiguity refusal included) and swept from its own binary overlay, its VMAD attachments cross-checked against the `.pex` chain the active order supplies. The merged surface hands both swept families the one `plugins=` list whole, so the "handed the active subset, states what it did not sweep" machinery is gone along with its sentence; each family now names in its own section the file it swept off-order, and says its epoch stamp does not cover content located outside the index. `exclude=` judges an off-order file as part of the scope, and a scope that resolves entirely off-order is not widened to the whole order. The `.pex` chain is deliberately still read from the active order, so a script that ships only inside the not-yet-enabled mod is reported UNVERIFIABLE rather than clean — the boundary is stated in the tool description, the render and the changelog rather than papered over.

Three shared pieces came out of the change rather than being copied: the `plugins=` split (`SweepOffOrderScope`), the off-order record stream (`SweepScope.RecordsFrom`, previously private to `ErrorCheck`), and the scripts sweep's per-record cross-check, which both lanes now call so they cannot drift on what counts as a finding.

Two probe cells changed subject because the asymmetry they asserted no longer exists: the off-order cells now assert the lane in both families, and the end-to-end family-local-refusal cell moved to the errors + dialogue pair, which is the pair that still scopes differently — a scope refusal shared by the two swept families is one ground and correctly collapses to one error.

Build, 1168 tests (8 new) and 128/128 probes green.

Closes #395


src/housecarl-generator/CheckShapeMatrix.cs
```diff
@@ -73,55 +73,54 @@ internal static IReadOnlyList<Shape> Build(ErrorCheckResult errors, ScriptCheckR
...
+            new CheckSweep(Sel("errors", "scripts", "dialogue"), e, sc, multi));
         Add("all three, listing, two seeds, roster",
             new CheckSweep(Sel("errors", "scripts", "dialogue"),
-                           e with { ExcludedPlugins = roster }, sc with { ExcludedPlugins = roster }, null, multi));
+                           e with { ExcludedPlugins = roster }, sc with { ExcludedPlugins = roster }, multi));
 
         // ---- OFF-ORDER: a named plugin the scripts family has no lane for ---------------------------
         Add("errors+scripts, listing, off-order",
-            new CheckSweep(Sel("errors", "scripts"), e, sc, new[] { "HcMxFresh.esp" }));
+            new CheckSweep(Sel("errors", "scripts"), e, sc));
```
Avick3110: The four shapes still named `off-order` no longer carry any off-order data. The old `CheckSweep` positional slot was deleted but nothing replaced it on the family results, so `errors+scripts, listing, off-order` is now byte-identical to `errors+scripts, listing`, `all three, counts_only, off-order` to `all three, counts_only`, and likewise for the refused and roster variants. Four inventory rows are duplicates and the inventory contains **no** shape whose scripts or errors result has `OffOrderScanned` set.

Concretely: `AppendScriptsHead` now writes a two-line "swept OFF-ORDER (on disk, not in the active load order): … [the file's own records; each attached script's .pex read from the ACTIVE order …]" block, and `WriteScriptsHead` gained `epoch_covers_all_inputs` plus `off_order_scanned`. All of that lands in the fixed part / JSON demand, and the comment kept at lines 119-121 says the fixed-part pass has to measure it — but with no shape carrying the data, `MATRIX-INSIDE-ITS-CAP` and the allocation properties never render it. A cap overrun on an off-order response at a biting `max_chars` would go unseen, which is the exact class of defect this matrix exists to catch.

Suggest putting the data back where it now lives, e.g. `sc with { OffOrderScanned = new[] { "HcMxFresh.esp" } }` (and `e with { … }` for the errors side), and updating the class doc at line 18, which still describes off-order as "a named plugin the scripts family has no lane for".

src/housecarl-core/ScriptPropertyCheck.cs
```diff
@@ -186,112 +199,112 @@ public static ScriptCheckResult Run(LoadOrderResolver resolver, LoadOrderResolve
...
-                    catch (Exception ex)
-                    {
-                        scanError = (scanError is null ? "" : scanError + "; ")
-                                  + $"a record's script adapter could not be read ({fk} — {ex.GetType().Name}: {ex.Message})";
-                    }
+            if (unbound.Count == 0 && nulls.Count == 0 && unver.Count == 0) return;
+
+            // Apply the shared finding budget (unbound + null are the capped population; unverifiable notes
+            // are few and always kept). The true totals are counted regardless of the cap.
+            totalUnbound += unbound.Count;
```
Avick3110, line 285: UNVERIFIABLE notes are deliberately outside `findingBudget` — fine when they are rare, but the new lane makes them the normal case for its own headline scenario. A pending mod is typically DISABLED (the new `ScriptsOffOrderTests` fixture is exactly this: `-PendingMod` in modlist.txt), so its whole `Scripts/` folder is outside the VFS and `ResolveChain` returns `OwnLoadError` for **every** script the file attaches.

So `housecarl_check plugins=["FreshMod.esp"] findings=["scripts"]` over a disabled mod with, say, 3,000 scripted records produces 3,000 uncapped report rows that all say the same thing, and `limit=` cannot trim any of them. The `max_chars` cut then falls on whatever comes last — including the unbound-property findings the caller actually asked for, since reports are in record order and the unbound rows are not privileged over the notes. The pre-enable verify sweep returns a wall of "could not read its .pex" with the real findings truncated out.

Worth collapsing repeated identical `OwnLoadError` notes per script class on the off-order lane (one note naming the class and how many records carry it), or letting them share the budget once the same class has been reported.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -854,6 +850,10 @@ static void AppendScriptsHead(StringBuilder sb, ScriptCheckResult r)
         if (r.Epoch is not null) sb.Append(" · epoch=").Append(r.Epoch).Append(EpochOffOrderQualifier(r));
         sb.Append('\n');
         if (r.FilterNote is not null) sb.Append(r.FilterNote).Append('\n');
+        if (r.OffOrderScanned is { Count: > 0 } off)
+            sb.Append("swept OFF-ORDER (on disk, not in the active load order): ").Append(string.Join(", ", off))
```
Avick3110: This sentence is now written verbatim in two places — here and in `AppendErrorsHead` at line 431 — differing only in the bracketed tail. The sentence it replaces (`ReadSentences.SweepFamilyOffOrderSkipped`) lived in `ReadSentences` carrying `[MustState("off-order", "did NOT sweep")]`, so deleting it drops the claim from the registry: the new sentence makes a coverage claim ("the file's own records", ".pex read from the ACTIVE order") that nothing now registers or holds to a spelling.

Both probe arms and `ScriptsOffOrderTests` match the literal by hand, so a wording change in one copy and not the other reddens them, but the two copies are still free to drift on the bracketed half. Suggest one `ReadSentences` constant parameterised on the per-family tail, with the `[MustState]` claim restored.

src/housecarl-mcp/SweepOffOrderScope.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// nowhere, and a name several mod folders provide each refuse before anything is swept.</summary>
+    internal static Refusal? Split(LoadOrderResolver.IndexView view, IReadOnlyList<string> plugins,
+                                   string modsDir, string dataDir, string overwriteDir, string profileDir,
+                                   out List<string> active, out List<(string Name, string Path)> offOrder)
+    {
+        active = new List<string>();
+        offOrder = new List<(string Name, string Path)>();
+        Mo2Composition? comp = null;
+        foreach (var name in plugins)
+        {
```
Avick3110, line 62: Routing the scripts family through this split silently drops its did-you-mean clause. `ScriptPropertyCheck.Run`'s membership refusal was `plugin not in the load order: {name}.{view.AbsenceClause(name)}` — `AbsenceClause` is the near-miss hint. A name that is neither in the order nor on disk now takes this branch instead and gets `plugin not in the load order: X — and no on-disk copy was found either (…)` with no suggestion, so the commonest failure (a typo, or the `.esp`/`.esl` extension guessed wrong) loses the one thing that fixes it. The new test `ANameThatIsNeitherInTheOrderNorOnDiskRefusesSayingBoth` pins the new sentence, so the regression is locked in.

The errors family had the same gap before this PR, so the cheap fix helps both: append `view.AbsenceClause(n)` to this message.

Avick3110:
Two more findings that fall outside the diff hunks, so they could not be posted inline.

**`src/housecarl-mcp/CheckTools.cs:128-132` — the `exclude=` description now contradicts the behaviour.** It still says: "A FILENAME YOU NAMED that nothing in scope matches is refused by each family that HAS a scope to match it against — where a family's scope is empty because every plugin you named is outside it, that family reports the empty scope instead, since the exclusion is not why it swept nothing." After this change no swept family can be in that state: both hold off-order files in scope, and the new probe arm `ORCH-EXCLUDE-JUDGES-AN-OFF-ORDER-SCOPE` asserts the opposite outcome on that very call shape — `plugins=["HcOrchOff.esp"] exclude=["HcOrchTwo.esp"]` refuses as unmatched, and `exclude=["HcOrchOff.esp"]` returns "removed every plugin this sweep would have covered (1 in scope, all excluded)". A caller reading the tool description will expect a report where they now get a refusal.

**`src/housecarl-generator/CheckMergeProbe.cs:41` — stale arm index.** The header inventory still lists `OFF-ORDER-STATED-PER-FAMILY — the scripts family has no off-order lane, and the response says which files it did not sweep`. The arm was renamed to `OFF-ORDER-NAMED-PER-FAMILY` and its claim inverted. Same family of staleness at line 1436 (`It was filtered out before \`noneInScope\` was computed` — that variable is gone) and `CheckShapeMatrix.cs:18` (`OFF-ORDER (a named plugin the scripts family has no lane for)`).

---
Rest of the diff read clean: the split is symmetric with `CheckErrors` (same `inScope` union, same `scopeBefore` arithmetic, same stamped/unstamped refusal handling), the off-order overlay is disposed on every path, `ScanRecord`'s captured counters and `findingBudget` behave as the inlined body did, the empty-`targets` guard correctly refuses to widen, and the new head sentence is inside `Compose` so the skeleton pass does measure it.

Avick3110:
Second review pass (fresh reader, blind). Two findings beyond the six already on this PR.

**1. `src/housecarl-mcp/JsonWire.cs:1770` — medium. The `.pex`-from-the-active-order boundary is stated in the text transport only.**

The PR body says the boundary is "stated in the tool description, the render and the changelog rather than papered over". It is stated in *one* render. `ReadTools.AppendScriptsHead` appends it inline:

```
swept OFF-ORDER (...): X   [the file's own records; each attached script's .pex read from the ACTIVE
order, so a script that ships only inside the not-yet-enabled mod reads UNVERIFIABLE, not clean]
```

`WriteScriptsHead` writes only `off_order_scanned` (the array) and `epoch_covers_all_inputs`. The shared `ReadSentences.SweepScriptBoundary`, which *is* written to both transports, says only "a script whose .pex is not on disk is reported unverifiable" — it never says the chain came from the active order, so it does not connect the off-order roster to the unverifiable count. A `format="json"` consumer therefore gets `off_order_scanned: ["Pending.esp"]`, `unverifiable: 1`, and nothing tying them together: the one caveat that makes this lane's verdict readable is absent from the machine transport. The existing probe idiom ("in both transports") is what this misses.

Both halves of the same root cause as the low finding on `ReadTools.cs:854`: the sentence is an inline literal rather than a `ReadSentences` const with a `[MustState]` claim, so the JSON lane never picked it up. Fixing it in `ReadSentences` fixes both.

Related, same area: `EpochOffOrderQualifier(ErrorCheckResult)` (ReadTools:431 region) and `EpochOffOrderQualifier(ScriptCheckResult)` are now byte-identical bodies. The doc comment justifying the second overload ("the overload exists so the two sweep headers stay textually identical") no longer describes what it does; one method over a `IReadOnlyList<string>?` would do.

**2. `src/housecarl-generator/CheckMergeProbe.cs:1340` — low. No cell now covers two swept families refusing on scope while a seeded family answers.**

The reasoning for moving `ORCH-A-FAMILY-LOCAL-REFUSAL-DOES-NOT-REFUSE-THE-CALL` to the errors + dialogue pair is sound as far as it goes — with one shared scope, an `exclude=` refusal really is one ground across errors and scripts, so the collapse is correct and nothing family-local is being hidden there. But the shape this PR newly makes ordinary is `findings=['errors','scripts','dialogue']` with an `exclude=` that empties the shared swept scope: two families refuse on the *same* ground, dialogue answers, and the grounds-are-one rule must *not* fire because a family answered. That is the one arrangement where the collapse could go wrong, and neither the moved cell (errors + dialogue, cost refusal) nor `ORCH-EXCLUDE-JUDGES-AN-OFF-ORDER-SCOPE` (scripts alone, so one family is the whole call) reaches it. Worth one arm.

Also note `ORCH-EXCLUDE-JUDGES-AN-OFF-ORDER-SCOPE` dropped the old cell's `!emptyScopeRealName.StartsWith("error")` assertion. With `findings=['scripts']` that assertion could not have survived anyway, but nothing replaced it elsewhere.

**Checked and clean**, for the record: the `SweepOffOrderScope` split is the same locate contract in both families, ambiguity refusal included, and its stamped/unstamped distinction is right; `ScanRecord`'s captured counters and `findingBudget` behave as the inlined body did on both lanes; the empty-`targets` guard correctly refuses to widen an entirely off-order scope, and the ordering of the `scope` / `offOrder` / whole-order branches handles the mixed case; the off-order overlay is disposed on every path; no live caller still reads `ScriptsSkippedOffOrder` (only stale comments, already flagged); the new subsystem went into its own file rather than into `LoadOrderService.cs`; and the eight new tests do drive the real overlay through `LoadOrderService.ValidateScripts` over a synthetic MO2 instance with a genuinely disabled mod, so the "`.pex` outside the VFS ⇒ UNVERIFIABLE" half is exercised for real and not simulated. The lane's untested corners are `counts_only=` and `type=` over the off-order stream, and the ambiguous-filename refusal.

Avick3110:
Review fold. All eight findings were real; all eight are folded, nothing refuted.

**Folded**

1. **`CheckShapeMatrix` off-order shapes were duplicates (medium).** The four shapes now carry the data: `OffOrderScanned = ["HcMxFresh.esp"]` on the errors and scripts results (the errors side alone on the scripts-refused shape, since a refusal has no roster), and the scripts side of the listing shape also carries a collapsed-unverifiable count so the third off-order fixed-part line is measured too. `OffOrderFile` is a name no other shape prints. Class doc updated. 31 shapes, still ALL PASS.

2. **UNVERIFIABLE notes bypassed `findingBudget` (medium).** Repeats are now collapsed sweep-wide, not just on the off-order lane: the first record carrying a given (script class, reason) lists it, every later one is counted into `ScriptCheckResult.UnverifiableCollapsed`, and a record whose only findings were repeats is not listed at all. The total still counts every one. The head states the collapse (`ReadSentences.SweepScriptUnverifiableCollapsed`), json carries `unverifiable_collapsed`. Test: `RepeatedUnverifiableNotesAreCollapsedAndTheCollapseIsStated`, over a fixture where three records attach the pending mod's own script.

3. **The `.pex`-from-the-active-order boundary was text-only (medium).** Both coverage tails are now `ReadSentences` consts, and json writes the family's tail as `off_order_coverage` beside `off_order_scanned` (null when nothing was swept off-order, so the key is never a caveat over a lane that did not run). One `WriteOffOrder` helper writes the pair for both families. Test: `TheJsonTransportTiesTheOffOrderRosterToTheUnverifiableCount`.

4. **`exclude=` description promised removed behaviour (medium).** Rewritten: the swept families share one scope, off-order files included, so an unmatched name is a typo on every family that could have run, and an exclusion that removes the whole scope is refused rather than sweeping nothing in silence.

5. **Inline literal, lost `[MustState]`, byte-identical overloads (low).** The head sentence is `ReadSentences.SweepOffOrderScanned` parameterised on the per-family tail, with `[MustState]` on it and on both tails. The two `EpochOffOrderQualifier` overloads are one method over `IReadOnlyList<string>?`.

6. **Stale arm-index text (low).** `CheckMergeProbe`'s header inventory row is now `OFF-ORDER-NAMED-PER-FAMILY` with the inverted claim, the `noneInScope` comment no longer names a variable that is gone, and `CheckShapeMatrix`'s class doc is updated (see 1).

7. **`view.AbsenceClause` dropped by the shared split (low).** Appended to the locate refusal in `SweepOffOrderScope.Split`, so the errors family gains it as well. New test `ANearMissNameKeepsItsDidYouMean` pins the suggestion; the old test still pins both halves of the sentence.

8. **No cell for two swept families refusing while a seeded one answers (low).** New arm `GROUNDS-ARE-ONE-DOES-NOT-FIRE-WHILE-A-FAMILY-ANSWERED`, next to `GROUNDS-ARE-ONE`: errors and scripts both refuse on the one ground their shared scope gives them, dialogue answers, and the response must stay three sections in both transports with `families_ran` naming one and `families_refused` naming two. Driven at the render layer rather than through the tool, because this fixture's dialogue seed refuses on its own ground and could not be the family that answered.

**Refuted:** none.

**Counts:** build clean; `dotnet test --filter "tier!=bridge"` 1171 passed, 0 failed (was 1168); `ci-all` ALL PASS, `check-guard` ALL PASS with one new arm.

Avick3110:
Blind review folded (8 of 8 real, 0 refuted). Queued for Aaron's review. Stacked follow-on: #564 (#405 #406).

src/housecarl-core/ScriptPropertyCheck.cs
```diff
@@ -179,119 +196,136 @@ public static ScriptCheckResult Run(LoadOrderResolver resolver, LoadOrderResolve
...
+            foreach (var n in nulls)   { if (findingBudget > 0) { keptNull.Add(n);   findingBudget--; } else capped = true; }
+
+            // Unverifiable notes ride outside the finding budget, and one unreadable script class hits EVERY record
+            // that attaches it — a disabled mod puts all of them outside the VFS at once, so the same sentence would
+            // fill the listing and push the unbound findings past max_chars. The first record carrying a given
+            // class+reason lists it; the repeats are counted and the collapse is stated in the head.
+            var keptUnver = new List<ScriptUnverifiable>();
+            foreach (var uv in unver)
+            {
+                if (seenUnverifiable.Add((uv.Script, uv.Reason))) keptUnver.Add(uv);
```
Avick3110: The collapse key is `(Script, Reason)`, and one of the two note kinds has a **constant** script name: the nameless-attachment note at line 231 is always `ScriptUnverifiable("(unnamed)", "the script attachment carries no class name — can't resolve its .pex to check its properties.")`. So across the whole sweep only the FIRST record with a nameless script entry is ever listed.

Scenario: a plugin has 50 records whose VMAD carries a script entry with no class name (a corrupt or hand-edited adapter). Record 1 is reported. For records 2–50 `seenUnverifiable.Add` returns false, `keptUnver` is empty, and the guard on line 311 (`unbound.Count == 0 && nulls.Count == 0 && keptUnver.Count == 0`) returns before `reports.Add` — so those 49 records never appear in the listing at all, at any `limit=` or `max_chars`.

That is the right answer for the case the collapse was written for: an unreadable `.pex` is one file, the class name identifies it, and repeating the sentence per record adds nothing. It is the wrong answer for `(unnamed)`, where the RECORD is the identity of the defect — the caller is left with a collapse count and no way to learn which records to open.

Exempting the nameless note from the collapse would fix it — collapse only notes whose `Script` is a real class name (the string is worth hoisting to a constant shared with line 231):

```suggestion
                if (uv.Script == "(unnamed)" || seenUnverifiable.Add((uv.Script, uv.Reason))) keptUnver.Add(uv);
```

src/housecarl-mcp/CheckTools.cs
Avick3110, line 1: The `limit=` parameter description (line 142, outside this diff) still says:

> Master-table findings and unverifiable notes are always listed in full (they are few).

The scripts family no longer lists them in full — repeats for the same script class are collapsed. Every other surface was updated to say so (the family description, the render head, `unverifiable_collapsed`, the changelog); this knob description was not, so an agent planning against the tool description is told the one thing the change made untrue.

Suggest amending that sentence, e.g. "…always listed in full (they are few); on the scripts family a note repeating one already reported for the same script class is collapsed and counted instead."

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4691,15 +4678,27 @@ public ScriptCheckResult ValidateScripts(IReadOnlyList<string>? plugins, int lim
...
-    /// sweep's own file, not here.</summary>
-    internal IReadOnlyList<string> ActivePluginNames => Resolver.PluginNames;
+        // The off-order lane, resolved exactly as CheckErrors resolves it: a named plugin the active order does not
+        // hold is located on disk and swept from its own file, which is how a fresh patch's script bindings can be
+        // checked BEFORE it is enabled.
+        if (plugins is { Count: > 0 })
+        {
+            string modsDir, dataDir, overwriteDir, profileDir;
+            lock (_gate) { EnsurePathsDerived(); modsDir = _modsDir; dataDir = _dataDir; overwriteDir = _overwriteDir; profileDir = _profileDir; }
+            if (SweepOffOrderScope.Split(view, plugins, modsDir, dataDir, overwriteDir, profileDir,
```
Avick3110, line 4795: Both swept families now resolve the same `plugins=` list independently: `CheckErrors` calls `SweepOffOrderScope.Split` at line 4541 and `ValidateScripts` calls it again here. `Split` is lazy about the composition, so an all-active scope costs nothing extra — but the moment a name is off-order it reads the MO2 composition and runs `Mo2LoadOrder.LocatePlugin`, a sweep over every mod folder in the install.

So the call this PR exists for — `housecarl_check(plugins=['FreshPatch.esp'], findings=['errors','scripts'])`, the default findings set — pays that composition read and whole-install folder sweep twice, for an answer that is identical both times. On a 3,800-plugin install that is not free.

The merged surface already knows both families take the same list; splitting once in `CheckTools` and handing both families the resolved `(active, offOrder)` pair would also make it structurally impossible for the two to disagree about which names resolved — which is the property `SweepOffOrderScope`'s doc comment claims for the shared code.

Avick3110:
Aaron's review fold. All three findings were real; all three are folded, nothing refuted.

**Folded**

1. **`ScriptPropertyCheck.cs:307` — the collapse swallowed all but the first nameless attachment.** Confirmed: the nameless note's script name is a constant, so `(Script, Reason)` was the same key on every record and the guard below returned before the record was listed. `(unnamed)` is now hoisted to `ScriptPropertyCheck.NamelessScript` and exempted from the collapse — it names no class, so the record is the only identity the defect has. Test: `ANamelessAttachmentIsListedOnEveryRecordRatherThanCollapsed`, over a new off-order fixture plugin whose three records each carry a class-less script entry (its own file, so the counts the other tests pin stay put). Doc, and the changelog entry for the collapse, say the carve-out.

2. **`CheckTools.cs:142` — the `limit=` description promised a listing the collapse made untrue.** It now says the notes are outside the cap but that a repeat for the same script class is collapsed to a count, with the nameless carve-out. Test: `TheLimitDescriptionStatesTheUnverifiableCollapseRatherThanAFullListing`, which also pins that the "always listed in full" claim is gone rather than qualified further down.

3. **`LoadOrderService.cs:4690` — both families resolved the same list independently.** Confirmed: the default findings set read the MO2 composition and ran `LocatePlugin` over every mod folder twice for an identical answer. `SweepOffOrderScope.Split` now takes an optional `SweepOffOrderMemo`, which `CheckTools` creates once and hands to both families; the memo answers only for the build and the very list it was filled against, and lives for the one call, so nothing caches a locate across calls. Test: `TheTwoSweptFamiliesResolveTheOffOrderScopeOnceBetweenThem` — after the errors family runs, a second disabled mod folder is planted with the same filename, so a fresh locate would refuse as ambiguous (asserted as the control) while the family handed the memo sweeps off the split already made.

**Refuted**

None.

Build, `tier!=bridge` tests (1174) and `ci-all` (128/128) green.

Issue #527, opened 2026-09-04
[Bug] a concrete-arm type filter (GlobalShort, GameSettingInt) returns the whole abstract group: the in-order scan lane never re-checks the arm


Avick3110:
From a Discord report (2026-08-25, houseCARL 1.9). Reproduces on `main`.

`type="GlobalShort"` and `type="GlobalFloat"` both return the identical whole-GLOB set, mixed, with no note that the filter widened. `ResolveTypeFilter` maps the name to `IGlobalShortGetter` correctly, but the in-order scan lane hands the type straight to Mutagen's `EnumerateMajorRecords` (`LoadOrderResolver.cs`, around lines 1032 and 1065) with no arm re-check. The off-order lane does re-check (`LoadOrderService.cs`, around line 4164). Generalises to every abstract-group arm (Global*, GameSetting*).

Fix: a one-line post-filter on the in-order lane mirroring the off-order lane. The corrupted denominator is invisible unless the caller reads per-row type labels.

Not #424 (that is generator-side type resolution). Rides W5 item 19.

PR #555, opened 2026-09-05
Give housecarl_skse the uniform transport lane

### LINKED ISSUE #539: [Bug] housecarl_skse carries max_chars= only: no format=json, no limit/offset, no in-band accounting (SPEC §2.1 TRANSPORT)
Surfaced by PR #537 (the `housecarl_skse` fold), stated there rather than papered over.

`housecarl_skse` carries `max_chars=` only. It has no `format=json`, no `limit`/`offset`, and no in-band accounting line. SPEC §2.1 makes TRANSPORT uniform across every tool and says a surface missing one of these is a bug, not a parity feature request. The three tools it replaced had the same gap, so #537 regressed nothing, but the fold did not close it either.

Fix: a JSON twin per family render (inventory, pairing, config), paging over the per-family row list, and the `total / rendered / capped / truncated / notes` accounting on every answer. Text render unchanged.

Avick3110:
`housecarl_skse` carried `max_chars=` and nothing else: no `format=json`, no `limit=`/`offset=`, no in-band accounting. SPEC §2.1 makes those axes uniform across every tool, so the gap is a bug rather than a parity request. This gives the tool the same lane the recently landed surfaces have, out of shared code rather than a bespoke variant: `housecarl_asset_status` owned the only accounting block on the read surface, so its counts, composer, reserve and paging window move into a shared `TransportAccounting` that both tools now render through — the extraction is behaviour-preserving for `asset_status` apart from one word in its truncation sentence, which is now parameterised on the row noun.

On `housecarl_skse` itself: `format='json'` returns the machine-readable twin of whichever family ran, carrying the same census, the same windowed rows and the same numbers in named fields, plus the family that ran and the two that did not — the text footer's prose cannot ride a json document, so it becomes `family` and `not_run`. `limit=`/`offset=` page each family's row list: the DLLs for `findings='inventory'` (with `filter=`, its DLL and config matches, the window walking the DLLs first and continuing into the configs), the native-declaring classes for `'pairing'`, the config files for `'config'`. The census above the rows keeps stating the whole layer while the rows are the window — the `asset_status` contract — and every answer ends on `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window or a cap left out is a number rather than a silence. `rendered` counts distinct rows that reached the page, because the inventory view lists one DLL in more than one section. An unrecognized `format=` is refused naming the two shapes the tool renders, a negative `limit=`/`offset=` is refused naming both knobs, and every one of the tool's own refusals now comes back in the shape the caller asked for.

Each family's json serializer lives beside its text render rather than in `JsonWire`, deliberately: the pairing twin has to classify with the same `Judge`/`BestFate` the text render uses, and the inventory twin with the same DLL split, so a copy in another file would be free to drift — which is the one thing a twin may not do. The shared skeleton the three documents have in common is in `SkseJsonDoc`. Text output with neither knob passed is unchanged apart from the closing accounting line.

Closes #539

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -95,7 +95,10 @@ saying it sets an expectation their install may contradict. Say what is known, a
   answer now ends on `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window
   or a cap left out is a number rather than a silence. An unrecognized `format=` is refused naming the two the tool
   renders, and a negative `limit=`/`offset=` is refused naming both. Text output with neither knob passed is unchanged
-  apart from that closing accounting line.
+  apart from that closing accounting line. `max_chars` bounds the json document the way it bounds the text render:
+  what it cuts from a list the accounting does not count — one config's references, the inventory's config folders,
+  the pairing's unreadable `.pex` entries — is named on the spot as `references_truncated`,
+  `config_folders_truncated` and `unreadable_pex_truncated`.
 - **`housecarl_load_order_status` now names the running server's build in its header** — a `server:` line carrying
   the informational version the binary embeds (the release version `build-plugin.ps1` stamps from `plugin.json`,
   then `+` and the full commit sha: `1.9.5-dev+e942910…`), so checking that the installed houseCARL is the build you expect no longer means
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -86,6 +86,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
   both halves now ends with the near-miss suggestion, so a typo or a guessed `.esp`/`.esl` extension names the
   plugin you meant.
 
+- **`housecarl_skse` now carries the same transport every other tool has: `format='json'`, `limit=`/`offset=` paging,
+  and an in-band accounting line on every answer.** `format='json'` returns the machine-readable twin of whichever
+  family ran — the same census, the same rows, the same numbers, in named fields, plus the family that ran and the two
+  that did not (the text footer's prose cannot ride a json document). `limit=`/`offset=` page the family's row list:
+  the DLLs for `findings='inventory'` (with `filter=`, its DLL and config matches), the native-declaring classes for
+  `'pairing'`, the config files for `'config'`. The census above the rows keeps stating the whole layer, and every
+  answer now ends on `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window
+  or a cap left out is a number rather than a silence. An unrecognized `format=` is refused naming the two the tool
+  renders, and a negative `limit=`/`offset=` is refused naming both. Text output with neither knob passed is unchanged
+  apart from that closing accounting line.
 - **`housecarl_load_order_status` now names the running server's build in its header** — a `server:` line carrying
   the informational version the binary embeds (the release version `build-plugin.ps1` stamps from `plugin.json`,
   then `+` and the full commit sha: `1.9.5-dev+e942910…`), so checking that the installed houseCARL is the build you expect no longer means
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -92,10 +92,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
   that did not (the text footer's prose cannot ride a json document). `limit=`/`offset=` page the family's row list:
   the DLLs for `findings='inventory'` (with `filter=`, its DLL and config matches), the native-declaring classes for
   `'pairing'`, the config files for `'config'`. The census above the rows keeps stating the whole layer, and every
-  answer now ends on `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window
-  or a cap left out is a number rather than a silence. An unrecognized `format=` is refused naming the two the tool
+  answer now carries `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window
+  or a cap left out is a number rather than a silence. It closes the render: on the text lane the one-line family
+  footer still follows it, on the json lane the `accounting` object is the document's last member. An unrecognized
+  `format=` is refused naming the two the tool
   renders, and a negative `limit=`/`offset=` is refused naming both. Text output with neither knob passed is unchanged
-  apart from that closing accounting line. `max_chars` bounds the json document the way it bounds the text render:
+  apart from that accounting line. `max_chars` bounds the json document the way it bounds the text render:
   what it cuts from a list the accounting does not count — one config's references, the inventory's config folders,
   the pairing's unreadable `.pex` entries — is named on the spot as `references_truncated`,
   `config_folders_truncated` and `unreadable_pex_truncated`.
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -91,12 +91,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   family ran — the same census, the same rows, the same numbers, in named fields, plus the family that ran and the two
   that did not (the text footer's prose cannot ride a json document). `limit=`/`offset=` page the family's row list:
   the DLLs for `findings='inventory'` (with `filter=`, its DLL and config matches), the native-declaring classes for
-  `'pairing'`, the config files for `'config'`. The census above the rows keeps stating the whole layer, and every
-  answer now carries `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window
-  or a cap left out is a number rather than a silence. It closes the render: on the text lane the one-line family
-  footer still follows it, on the json lane the `accounting` object is the document's last member. An unrecognized
-  `format=` is refused naming the two the tool
-  renders, and a negative `limit=`/`offset=` is refused naming both. Text output with neither knob passed is unchanged
+  `'pairing'`, the config files for `'config'`. The census above the rows states the whole population the answer is
+  about — the layer, or, under `filter=`, the filter's matches — while the rows below it are the window; a count a
+  filter cannot scope (the inventory's uncategorized-file total, the pairing's `.pex` scan) is left out of a filtered
+  answer rather than stated for the whole layer beside numbers that are not. Every answer now carries
+  `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window or a cap left out
+  is a number rather than a silence. It closes the render: on the text lane the one-line family footer still follows
+  it, on the json lane the `accounting` object is the document's last member. An unrecognized `format=` is refused
+  naming the two the tool renders, and a negative `limit=`/`offset=` is refused naming both. Text output with neither knob passed is unchanged
   apart from that accounting line. `max_chars` bounds the json document the way it bounds the text render:
   what it cuts from a list the accounting does not count — one config's references, the inventory's config folders,
   the pairing's unreadable `.pex` entries — is named on the spot as `references_truncated`,
--- a/src/housecarl-mcp/SkseTools.cs
+++ b/src/housecarl-mcp/SkseTools.cs
@@ -1,5 +1,6 @@
 using System.ComponentModel;
 using System.Text;
+using System.Text.Json;
 using HousecarlCore;
 using ModelContextProtocol.Server;
 
@@ -90,18 +91,39 @@ public static class SkseTools
     /// and that the answer ends on the footer. <see cref="ServiceRenders"/> is the live implementation.</summary>
     internal interface IFamilyRenders
     {
-        string Inventory(string? filter, bool peek, int cap);
-        string Pairing(string? filter, int cap);
-        string Config(string? filter, int cap);
+        string Inventory(FamilyCall c);
+        string Pairing(FamilyCall c);
+        string Config(FamilyCall c);
     }
+
+    /// <summary>One call's shared render context: what to narrow to, the char budget, the TRANSPORT paging window,
+    /// and which format was asked for. A record so a new TRANSPORT axis lands here rather than as a fourth positional
+    /// argument on all three renders.</summary>
+    internal readonly record struct FamilyCall(string? Filter, bool Peek, int Cap, RowWindow Window, bool Json);
 
     /// <summary>The live renders: each family's data read from the service, handed to its own wire class.</summary>
     sealed class ServiceRenders(LoadOrderService svc) : IFamilyRenders
     {
-        public string Inventory(string? filter, bool peek, int cap)
-            => SkseInventoryWire.Render(svc.SkseInventory(peek ? filter!.Trim() : null), filter, cap);
-        public string Pairing(string? filter, int cap) => NativePairingWire.Render(svc.NativePairingAudit(), filter, cap);
-        public string Config(string? filter, int cap) => SkseConfigAuditWire.Render(svc.SkseConfigAudit(), filter, cap);
+        public string Inventory(FamilyCall c)
+        {
+            var d = svc.SkseInventory(c.Peek ? c.Filter!.Trim() : null);
+            return c.Json ? SkseInventoryWire.RenderJson(d, c.Filter, c.Cap, c.Window)
+                          : SkseInventoryWire.Render(d, c.Filter, c.Cap, c.Window);
+        }
+
+        public string Pairing(FamilyCall c)
+        {
+            var d = svc.NativePairingAudit();
+            return c.Json ? NativePairingWire.RenderJson(d, c.Filter, c.Cap, c.Window)
+                          : NativePairingWire.Render(d, c.Filter, c.Cap, c.Window);
+        }
+
+        public string Config(FamilyCall c)
+        {
+            var d = svc.SkseConfigAudit();
+            return c.Json ? SkseConfigAuditWire.RenderJson(d, c.Filter, c.Cap, c.Window)
+                          : SkseConfigAuditWire.Render(d, c.Filter, c.Cap, c.Window);
+        }
     }
 
     /// <summary>Runs the selected family and appends the footer. The footer's length is subtracted from max_chars so it
@@ -109,18 +131,28 @@ public static class SkseTools
     /// append their scope note, caveats and filter hint unconditionally, so a response can still overrun a tight cap by
     /// one item plus that tail. Under a cap too small to hold both, the footer wins: it is the line that says which
     /// family answered.</summary>
-    internal static string Dispatch(IFamilyRenders renders, SkseFamily family, string? filter, bool peek, int max_chars)
+    internal static string Dispatch(IFamilyRenders renders, SkseFamily family, string? filter, bool peek, int max_chars,
+                                    bool json = false, RowWindow window = default)
     {
-        var footer = FamilyFooter(family);
+        // The json document states the family and the two that did not run in-band, so appending the text footer to
+        // it would only break the document.
+        var footer = json ? "" : FamilyFooter(family);
         int cap = Math.Max(1, (max_chars > 0 ? max_chars : 80_000) - footer.Length);
+        var call = new FamilyCall(filter, peek, cap, window, json);
         var body = family switch
         {
-            SkseFamily.Inventory => renders.Inventory(filter, peek, cap),
-            SkseFamily.Pairing => renders.Pairing(filter, cap),
-            _ => renders.Config(filter, cap),
+            SkseFamily.Inventory => renders.Inventory(call),
+            SkseFamily.Pairing => renders.Pairing(call),
+            _ => renders.Config(call),
         };
         return body + footer;
     }
+
+    /// <summary>The two families this call did not run, in the spelling that would run them — the json twin of
+    /// <see cref="FamilyFooter"/>, so a json consumer learns the default narrowed exactly as a text one does.</summary>
+    internal static string[] NotRun(SkseFamily ran) =>
+        new[] { SkseFamily.Inventory, SkseFamily.Pairing, SkseFamily.Config }
+            .Where(f => f != ran).Select(f => f.ToString().ToLowerInvariant()).ToArray();
 
     [McpServerTool(Name = ToolNames.Skse, ReadOnly = true, Title = "The SKSE layer: DLL/config inventory, native pairing, config references"),
      Description(
@@ -143,7 +175,12 @@ public static class SkseTools
          "filter= narrows whichever family ran (its match domain is that family's own — see filter= below); peek= " +
          "belongs to the inventory family alone and is refused, not ignored, on the other two. NOT COVERED by any " +
          "family: distributor INIs in Data\\ root (SPID *_DISTR, KID *_KID) — they live outside SKSE\\Plugins and are " +
-         "owned by the spid-authoring / kid-authoring skills.")]
+         "owned by the spid-authoring / kid-authoring skills. " +
+         // ---- transport, the same axes every tool carries -------------------------------------------
+         "TRANSPORT: format= 'text' | 'json' (the same rows and accounting, machine-readable); limit=/offset= page " +
+         "the family's row list; max_chars= caps the render. Every response ends on the in-band accounting — " +
+         "total / rendered / skipped / capped / truncated / offset / remaining / notes — so what a window or a cap " +
+         "left out is a number, never a silence.")]
     public static string Skse(
         LoadOrderService svc,
         [Description(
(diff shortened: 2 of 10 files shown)


src/housecarl-mcp/TransportAccounting.cs
```diff
@@ -0,0 +1,150 @@
...
+        if (Limit > 0) q = q.Take(Limit);
+        return q.ToList();
+    }
+
+    /// <summary>The window over a SECOND list that continues the first — the shape a family whose row list is two
+    /// concatenated populations needs (SKSE inventory: DLLs then configs). <paramref name="consumed"/> is how many
+    /// rows the first list held, so the offset lands where the first list stopped and the limit counts what the
+    /// first list already spent.</summary>
+    internal RowWindow After(int consumed, int taken) =>
+        new(Math.Max(Offset - consumed, 0), Limit <= 0 ? 0 : Math.Max(Limit - taken, 0));
```
Avick3110: **An exhausted `limit=` turns into "no limit" for the second list.**

`After` returns `Limit = 0` whenever the first list already spent the whole budget (`Math.Max(Limit - taken, 0)`), and `Apply` reads `Limit <= 0` as *unlimited*. So the continuation window over the second population is unbounded exactly when it should be empty.

Both call sites hit it: `SkseTools.cs:482` (text) and `SkseTools.cs:792` (json).

Concrete: an inventory layer with 2 DLLs and 30 configs all matching `filter='Mod'`, called with `limit=2`. The 2 DLL matches consume the limit, `After(2, 2)` yields `(0, 0)`, and all 30 configs render. Verified against the branch — the answer ends on `[accounting] total=32 rendered=32 skipped=0 capped=0 truncated=0`, so the accounting also asserts the window was honoured when it was not.

The sentinel needs to distinguish "no limit was asked for" from "the limit is used up":

```suggestion
    internal RowWindow After(int consumed, int taken) =>
        new(Math.Max(Offset - consumed, 0), Limit <= 0 ? 0 : Math.Max(Limit - taken, -1));
```

(That is one way — a `-1`/`Exhausted` marker that `Apply` treats as "take nothing". Any encoding works as long as a spent limit does not collide with the unlimited sentinel.)

Avick3110: Correction to the sketch above — `Math.Max(Limit - taken, -1)` still yields `0` when the limit is exactly spent, so it does not fix it. The distinction has to be encoded explicitly, e.g.:

```csharp
internal RowWindow After(int consumed, int taken) =>
    new(Math.Max(Offset - consumed, 0), Limit <= 0 ? 0 : Math.Max(Limit - taken, 0) is var left && left == 0 ? -1 : left);
```

with `Apply` returning an empty list for `Limit < 0`. Any encoding is fine as long as "budget spent" and "no budget asked for" are not the same value. The finding itself stands as described.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -653,23 +751,177 @@ static string Provider(SkseFileEntry e) =>
...
+                    SkseJsonDoc.Strings(w, "providers", g.Providers!);
+                    w.WriteNumber("contested", g.Contested);
+                    w.WriteEndObject();
+                }
+            w.WriteEndArray();
+
+            if (d.PeekRequested && allDlls.Count == 0)
+                w.WriteString("peek_note", "peek=true matched no DLL at all — nothing was peeked.");
+            SkseJsonDoc.Caveats(w, d.ReadIncomplete, d.Warnings, d.BsaFailures);
+            TransportAccounting.WriteJson(w, TransportAccounting.Tally(total, windowed, rendered, window, notes));
```
Avick3110, line 883: **The json lane holds back no reserve, so `max_chars` is overshot by the tail.**

Every text render subtracts `TransportAccounting.Reserve(...)` from `cap` before the row loop, precisely so the closing accounting block is paid for inside `max_chars` rather than appended past it. None of the three `RenderJson` methods does — `cap` goes straight into `SkseJsonDoc.Over`, and then `Caveats` and `WriteJson` are written unconditionally after the arrays close.

Concrete: `format='json' max_chars=1500` on a build with a handful of BSA read failures and warnings. The row loops fill the document to just under 1500, then the `caveats` object (one string per warning and per archive failure) plus the eight-field `accounting` object push the returned document well past the cap the caller set. Same at `SkseTools.cs:1231` and `SkseTools.cs:1700`.

The branch's own `OverMaxCharsTheJsonTwinDropsRowsAndStaysValid` test asserts the document parses and that rows were dropped, but never asserts its length against the cap, so this passes.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -433,10 +529,12 @@ static string RenderFiltered(SkseInventoryData d, string filter, int cap)
                 sb.Append("    - ").Append(e.FileName);
                 if (e.ProviderCount > 1) sb.Append(": ").Append(Chain(e));   // contested config → the full winner→loser chain
                 else sb.Append(Provider(e));
-                sb.Append('\n'); shown++;
+                sb.Append('\n'); shown++; tally.Mark(e.RelPath);
             }
         }
-        return sb.ToString().TrimEnd('\n');
+        // A config already shown as a DLL's paired config is a rendered row too, so it counts.
+        foreach (var e in cfgHits) if (shownCfg.Contains(e.RelPath)) tally.Mark(e.RelPath);
```
Avick3110, line 536: **A config rendered as a DLL's paired config is never tallied unless it also lands inside the window, so `rendered` undercounts and the next-page advice can re-show rows.**

`AppendDetail` (line 598) collects paired configs by stem prefix over **all** of `d.Configs`, not over `cfgHits`, and adds each to `shownCfg`. Line 536 then only marks the ones that are also in `cfgHits`. A paired config that sits past the window's end is printed in full under its DLL but counted as neither rendered nor windowed.

Concrete: `filter='SkyPatcher' limit=1` where the one DLL match pulls in five paired configs, all of which fall after the window. All five are printed; the accounting reports `rendered=1`, and the advice says `offset=1` for the next page — which then re-renders those same five configs as fresh rows. A caller paging by the tool's own advice sees them twice.

The same asymmetry means `shownCfg` can suppress a windowed config from `rest` while the config that displaced it was never counted, so `truncated` absorbs the difference and reads as a `max_chars` cut that never happened.

Avick3110: Retracting this one — I misread it. Line 536 does mark windowed configs that were shown under a DLL, so `rendered` and `truncated` are correct, and a paired config outside the window is legitimately not a row of this page. The only residue is cosmetic: such a config appears once inline under its DLL and again as a row on a later page. Not a bug. Sorry for the noise.

Avick3110:
Three further findings from a second read, all lower severity than the `RowWindow.After` one already posted inline.

**1. `src/housecarl-mcp-tests/SkseFamilySelectionTests.cs:130` — low (test gap).** The recorder gains `public SkseTools.FamilyCall Call;` and every family setter assigns it, but nothing in the file ever reads it. That is the one seam where a test could assert that `limit=`/`offset=`/`format=` actually reach the renders through `Dispatch`, and it is left unasserted — so nothing pins that the tool's new knobs are wired to the `FamilyCall` at all. Of the 36 new tests, 29 call the three wire classes directly and the 7 that go through the server all land on refusals or the schema; no test drives a successful paged or json answer over the wire. Either assert on `Call` (window round-trips, `Json` is set for `format='json'`, footer suppressed) or drop the field.

**2. `src/housecarl-mcp/SkseTools.cs:1212` — low.** In `SkseConfigAuditWire.RenderJson` the `SkseJsonDoc.Over(w, ms, cap)` check is per-FILE, but the inner `foreach (var r in file.Refs)` loop has no cap check. One config with tens of thousands of form tokens (a large SkyPatcher INI is exactly that) writes its whole reference array past `max_chars` in a single unbroken pass. The text render bounds refs per line; the twin does not.

**3. `src/housecarl-mcp/SkseTools.cs:836` and `:1687` — low.** The `config_folders` (inventory) and `unreadable_pex` (pairing) json arrays `break` at the cap with no marker of their own. They are not row-list rows, so the `accounting` block does not count them, and the text renders emit an explicit `... [showing N of M folders; raise max_chars]` notice in the same place. It is recoverable — `totals.config_folders` and `totals.unreadable_pex` state the true counts, so a consumer comparing array length to the total can tell — but the json lane loses the "what was cut" notice the text lane gives, which is the asymmetry §2.1 exists to remove.

Checked and clean beyond what the inline review covered: the `AssetWire` extraction moves nothing but the accounting (the `Render` body, the reserve subtraction and `AccountingReserve`'s signature are byte-identical, and `"resolved path(s)"` is the sole wording change — no test asserts it); `Tally` reconciles `skipped + rendered + truncated + capped == total` at `offset` past the end, at `limit=0`, and at `offset=0 limit=0`; `Split`/`Classify` run the same function over the census and the window so the two cannot drift; the pairing twin's `VerdictName` routes through the same `Judge`/`BestFate`/`IsDebugOnly` the text sections use.

Avick3110:
Review fold. All five findings were real; nothing refuted.

**1 (high) — `RowWindow.After` turned a spent limit into no limit.** Folded. `RowWindow` gains a third member, `Spent`: `After` sets it when the first population used the whole budget, and `Apply` returns nothing for a spent window. `Limit = 0` keeps meaning "no limit was asked for", so the two states no longer collide. Tests: `ALimitTheFirstPopulationSpendsLeavesNothingForTheSecond` (the reported case — 2 DLLs, 30 configs, `filter='Mod' limit=2`, now `total=32 rendered=2 capped=30`), its json twin, plus two controls — a partly spent limit still continues into the second population, and an offset landing inside it still renders it.

**2 (low) — the json lane held back no reserve.** Folded. `SkseJsonDoc.TailReserve` measures the widest tail each document will close on (the caveats object, the accounting object, and that family's own conditional members) and all three `RenderJson` methods subtract it from `cap` before the row loops, the way every text render subtracts `TransportAccounting.Reserve`. `TransportAccounting.Widest` became internal so the two reserves measure off one composer. Test: `TheJsonTailIsPaidForInsideMaxCharsRatherThanAppendedPastIt` over all three families — a ~12 KB caveat tail inside a 20 KB cap, asserting the document length against the cap, which is what the old test never did.

**3 (low) — `FamilyCall` never read, no successful wire answer.** Folded both halves. `TheTransportKnobsReachTheRenderOnTheCall` and `AJsonCallSetsTheJsonFlagAndTakesNoFooter` assert the filter, peek, window, cap and json flag the dispatch composes. New `SkseTransportWireTests` builds an MO2 instance of six SKSE configs, configures its own server (the shared fixture is deliberately unconfigured) and drives `findings='config' limit=2 offset=1` in both formats: the json document comes back parseable with `f2.ini`/`f3.ini` and `total=6 rendered=2 skipped=1 capped=3 offset=1`, and the text answer carries the same numbers and still ends on the family footer.

**4 (low) — the inner `Refs` loop was unbounded.** Folded. The reference loop tests the cap before each reference, and the file states `references_truncated` for what was cut, since the accounting counts files rather than references. Test: `MaxCharsBoundsTheReferenceArrayInsideOneConfigFileToo` — one config of 4,000 refs at `max_chars=3000`.

**5 (low) — `config_folders` and `unreadable_pex` broke with no marker.** Folded. Both now emit `config_folders_truncated` / `unreadable_pex_truncated` when the cap cut them, and nothing when it did not. Tests: `TheNonRowJsonArraysMarkWhatMaxCharsCutFromThem` and the no-cut control.

Each fold has a test that fails on the pre-fold tree (verified: 7 of the new assertions fail without the source changes). The changelog entry gained a sentence for the three new cut markers.

Build, `tier!=bridge` tests (1210 passed) and `ci-all` are green.

Avick3110:
Blind review folded (5 of 5 real, 0 refuted). Queued for Aaron's review. Stacked follow-on: #566 (#527 #449; #448 found already fixed on main by 64a730b, close by hand).

src/housecarl-mcp/SkseTools.cs
```diff
@@ -863,18 +1161,113 @@ bool Match(SkseConfigFileAudit f) =>
...
+        var flatAll = d.Files.SelectMany(x => x.Refs).ToList();
+        int notes = NoteCount(d);
+        int rendered = 0;
+        // The caveats and accounting tail is paid for inside max_chars rather than appended past it.
+        cap = Math.Max(1, cap - SkseJsonDoc.TailReserve(d.ReadIncomplete, d.Warnings, d.BsaFailures,
+            TransportAccounting.Widest(allFiles.Count, files.Count, window, notes)));
+
+        return SkseJsonDoc.Write(SkseTools.SkseFamily.Config, filter, d.ProfileName, (w, ms) =>
+        {
+            int Verdicts(SkseRefVerdict v) => flatAll.Count(r => r.Verdict == v);
```
Avick3110, line 1216: **The json `totals` block ignores `filter=` and states the whole layer.**

`flatAll` (line 1193) is built from `d.Files`, not from `allFiles`, so every number in `totals` — `configs_scanned`, `files_with_references`, `references_checked`, `ok`/`dangling`/`unparseable`/`plugin_missing`, `broken`, `inert`, `read_errors` — describes the whole audit while `files` and `accounting` describe the filter's matches.

Driven over synthetic data: a layer of 5 configs, 2 in folder `Alpha` (one OK ref, one dangling) and 3 in `Beta` (two dangling, one unparseable). `SkseConfigAuditWire.RenderJson(d, "Alpha", 200_000)` gives

```
files.length = 2   accounting.total = 2
totals = { configs_scanned: 5, references_checked: 5, dangling: 3, unparseable: 1, broken: 4 }
```

A consumer that filters to `Alpha` and reads `totals.broken` is told the folder carries 4 broken references; it carries 1. The text lane does not publish these counts under a filter at all — its header restates the population as `2 config(s) match` — so the two lanes differ in more than formatting here.

The same shape is in the other two json twins (`SkseInventoryWire.RenderJson` line 798, `NativePairingWire.RenderJson` line 1641), commented separately.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -653,23 +751,192 @@ static string Provider(SkseFileEntry e) =>
...
+        // appended past it, exactly as the text render's own reserve does.
+        cap = Math.Max(1, cap - SkseJsonDoc.TailReserve(d.ReadIncomplete, d.Warnings, d.BsaFailures,
+            TransportAccounting.Widest(total, windowed, window, notes),
+            tw => { tw.WriteString("peek_note", PeekNoDllNote); tw.WriteNumber("config_folders_truncated", folderCount); }));
+
+        return SkseJsonDoc.Write(SkseTools.SkseFamily.Inventory, filter, d.ProfileName, (w, ms) =>
+        {
+            SkseJsonDoc.Nullable(w, "installed_runtime", d.InstalledRuntime);
+            w.WriteStartObject("totals");
+            w.WriteNumber("dlls", all.Loaded.Count);
```
Avick3110, line 817: **`totals` here is the whole layer even under `filter=`.**

`all = Split(d.Dlls)` (line 798) is computed unconditionally, so `dlls`, `subfolder_dlls`, `modern`, `legacy_query`, `non_plugin`, `bsa_only`, `unreadable`, `address_library`, `signature_scanning` and `version_locked` all count the whole layer while the `dlls`/`configs` arrays and `accounting` count the filter's matches.

With 5 loader-scoped DLLs, `RenderJson(d, "p3", 200_000)` gives `dlls.length = 1`, `accounting.total = 1`, but `totals.dlls = 5`, `totals.modern = 5` — while the text lane's header for the same call reads `filter 'p3' — 1 DLL + 0 config match(es)`.

Second, smaller: `config_folders` is opened unconditionally but only filled `if (!filtered)`, so a filtered document carries `"config_folders": []` next to a non-zero `totals.config_folders`, which reads as "this layer has no config folders" rather than "this view does not list them".

src/housecarl-mcp/SkseTools.cs
```diff
@@ -1165,25 +1603,153 @@ bool Match(NativeClassEntry c) => In(c.ClassName) || In(c.RelPath) || In(c.Winni
...
+            TransportAccounting.Widest(allClasses.Count, classes.Count, window, notes),
+            tw => tw.WriteNumber("unreadable_pex_truncated", d.Unreadable.Count)));
+
+        return SkseJsonDoc.Write(SkseTools.SkseFamily.Pairing, filter, d.ProfileName, (w, ms) =>
+        {
+            SkseJsonDoc.Nullable(w, "installed_runtime", d.InstalledRuntime);
+            // Tri-state: null is "the check itself failed", never a checked-and-absent verdict.
+            if (d.SkseLoaderSeen is { } seen) w.WriteBoolean("skse_loader_seen", seen); else w.WriteNull("skse_loader_seen");
+            w.WriteStartObject("totals");
+            w.WriteNumber("classes", d.Classes.Count);
```
Avick3110: **Same as the other two twins: `totals` is the whole audit under `filter=`.**

`all = Classify(d.Classes, d.InstalledRuntime)` (line 1641) classifies every class, and `classes`/`pex_scanned`/`unreadable_pex` come straight off `d`, so `engine`, `skse_core`, `unpaired`, `dead`, `verify`, `debug_build` and `healthy` state the whole audit while the `classes` array and `accounting.total` state the filter's matches.

An audit of 40 classes with 3 dead, called `findings='pairing' format='json' filter='MyMod'` where MyMod contributes one healthy class, returns `classes.length = 1`, `accounting.total = 1`, `totals.dead = 3`, `totals.classes = 40`. A consumer scoping to one mod reads three dead pairings that belong to other mods.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -979,28 +1372,62 @@ static bool IsDebugOnly(NativeClassEntry c, string? runtime)
...
+        // accounting block is held back out of the cap so it is paid for rather than appended past it.
+        int notes = NoteCount(d);
+        var rows = window.Apply(d.Classes);
+        int reserve = TransportAccounting.Reserve(d.Classes.Count, rows.Count, window, notes, RowNoun);
+        cap = Math.Max(1, cap - reserve);
+        var tally = new RowTally();
+
+        var all = Classify(d.Classes, d.InstalledRuntime);
+        var w = Classify(rows, d.InstalledRuntime);
+        var engine = all.Engine; var skseCore = all.SkseCore;
```
Avick3110: **The "accounted for" block now mixes a whole-audit count and a windowed count on adjacent lines.**

`engine`/`skseCore` come from `all`; `healthy` (line 1427) comes from `w`. The block's own comment says it is "everything that is not a finding, so nothing is dropped" — a reconciliation — and under a window it stops reconciling.

Driven over synthetic data: 10 classes, 6 Engine and 4 healthy third-party, `RowWindow(0, 8)` renders

```
accounted for: 6 engine class(es) … · 0 SKSE-core class(es)
paired healthy (2 class(es)) — implementing mod ← its classes:
```

6 is the layer, 2 is the page, and nothing in the two lines says which is which. It gets sharper at `limit=4 offset=6`: the window then holds only the 4 third-party classes, yet the answer still prints `accounted for: 6 engine class(es)` — already printed in full on page 1 — above `paired healthy (4 class(es))`, the page's own count.

The other two families keep this straight: the config render's whole "accounted for" line is layer-wide, and the inventory's census is layer-wide with only the listed sections windowed.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,19 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **`housecarl_skse` now carries the same transport every other tool has: `format='json'`, `limit=`/`offset=` paging,
+  and an in-band accounting line on every answer.** `format='json'` returns the machine-readable twin of whichever
+  family ran — the same census, the same rows, the same numbers, in named fields, plus the family that ran and the two
+  that did not (the text footer's prose cannot ride a json document). `limit=`/`offset=` page the family's row list:
+  the DLLs for `findings='inventory'` (with `filter=`, its DLL and config matches), the native-declaring classes for
+  `'pairing'`, the config files for `'config'`. The census above the rows keeps stating the whole layer, and every
+  answer now ends on `total / rendered / skipped / capped / truncated / offset / remaining / notes`, so what a window
```
Avick3110: **A text answer does not end on the accounting — the family footer does.**

`Dispatch` returns `body + footer`, and `body` is what already carries the accounting, so on the text lane the order is `… [accounting] … \n\n(this ran the X family; …)`. `SkseTransportWireTests.ThePagedTextAnswerCarriesTheSameWindowAndEndsOnTheFooter` asserts exactly that with `Assert.EndsWith(FamilyFooter(...))`.

Only the json lane actually closes on the accounting. `TransportAccounting.Compose`'s own doc comment ("The one machine-readable accounting line, always last") carries the same inaccuracy. Either reword both to "every answer carries" / "last before the footer", or move the footer above the accounting.

src/housecarl-mcp/TransportAccounting.cs
```diff
@@ -0,0 +1,157 @@
...
+            Limit <= 0 ? 0 : Math.Max(Limit - taken, 0),
+            Spent || (Limit > 0 && taken >= Limit));
+
+    /// <summary>The window's own refusal, or null when both values are legal. One sentence naming both knobs,
+    /// because a caller who got one wrong usually typed the other in the same call.</summary>
+    internal string? Error =>
+        Offset < 0 || Limit < 0
+            ? $"error: limit={Limit} offset={Offset} — neither can be negative. Pass limit=0 for no limit and offset=0 to start at the beginning of the selection."
+            : null;
+}
```
Avick3110, line 43: **One string the extraction left duplicated.** `RowWindow.Error` now owns this sentence, but `AssetTools.cs:58-60` still carries a byte-identical inline copy:

```csharp
if (limit < 0 || offset < 0)
    return $"error: limit={limit} offset={offset} — neither can be negative. Pass limit=0 for no limit and " +
           "offset=0 to start at the beginning of the selection.";
```

The PR's premise is that `asset_status` and `housecarl_skse` render through one accounting; this refusal is the last piece still spelled twice, and rewording one leaves the other answering the old sentence for the same input class. `asset_status` can take the same `new RowWindow(offset, limit).Error` check `housecarl_skse` does — it already builds a `RowWindow` from those two values a few lines later in `AssetWire.Tally`.

Avick3110:
Aaron's review folded. All six findings were real; nothing refuted.

**1–3 (the three `totals` blocks state the whole layer under `filter=`).** Folded, one rule across the three twins: the census states the population the document answers over — the filter's matches when there is a filter, the whole layer when there is not — so no number in it describes a wider set than the `files`/`dlls`/`classes` array and the `accounting` beside it. `flatAll` is now taken off `allFiles`, `Split` off `allDlls`, `Classify` off `allClasses`; `configs_scanned`, `files_with_references`, `read_errors`, `configs`, `config_folders` and `classes` all count the matches. A count a filter has nothing to match on is left out of a filtered answer rather than stated for the whole layer beside numbers that are not: `other_files` (counted, never listed) and `pex_scanned`. `unreadable_pex` follows the same rule for its count; its *array* stays whole in both modes, because an unparseable `.pex` has no class name, provider or paired mod to filter on and dropping it would hide the one caveat saying those files were not counted as native-free. Unfiltered output is unchanged. Tests: `TheConfigTwinsCensusCountsTheFiltersReferencesNotTheWholeAudit` (your `Alpha` case in miniature — 5 configs, filter matches 1, `references_checked=2 inert=2` instead of the audit's 10), `ThePairingTwinsCensusClassifiesTheFiltersClassesNotTheWholeAudit` (1 healthy MyMod class beside 3 dead ones belonging to other mods: `dead=0` under the filter, `dead=3` unfiltered), `TheInventoryTwinsCensusCountsItsMatchesAndOmitsWhatAFilterCannotScope`.

The second half of the inventory finding — `"config_folders": []` next to a non-zero `totals.config_folders` — folded with it: a filtered document now omits the folder table rather than writing it empty, since it lists its matching configs individually above and `[]` reads as "this layer has no config folders". Asserted in the same test.

**4 (the "accounted for" block mixed a whole-audit count and a windowed count).** Folded. `engine` and `skseCore` now come from the windowed `Classify`, like every other section below the summary, so the block reconciles the page: its rows are its findings plus its baseline counts plus its healthy roster. The whole-audit numbers stay where they already were, in the summary above. The missing-loader alarm is the one thing that stays layer-wide — it is a build-level fact, not a row, and a window holding none of the SKSE-core classes must not silence it. Tests: `TheAccountedForBaselineCountsThePagesRowsNotTheWholeAudit` (your sharper case, `limit=4 offset=6` over 6 engine + 4 healthy — now `0 engine` above `paired healthy (4 class(es))`) and `TheMissingLoaderAlarmRidesTheWholeAuditNotTheWindow`.

**5 (a text answer does not end on the accounting).** Folded as a reword of both, since the footer is deliberately last: the changelog now says every answer *carries* the block and that it closes the render — the family footer following it on the text lane, the `accounting` object closing the document on the json lane — and `Compose`'s doc comment says it closes the render body, naming the footer as what a surface may write after it.

**6 (the negative-window refusal spelled twice).** Folded. `asset_status` now returns `new RowWindow(offset, limit).Error`; the inline copy is gone. Test: `TheNegativeWindowRefusalIsTheOneTheWindowItselfSpells` asserts the tool's answer is equal to the window's own sentence, for a negative `limit=` and a negative `offset=`, so a reword of one cannot leave the other behind.

Six new tests. Each of the five source folds has one that fails on the pre-fold tree (verified by restoring `SkseTools.cs` and re-running: the three census tests and the two baseline tests fail without the changes); the sixth is the refusal equality, which fails against the duplicated string only if the two are reworded apart, which is the point of it.

Build, `tier!=bridge` tests (1216 passed, was 1210) and `ci-all` (128/128) are green.

PR #557, opened 2026-09-05
Add the owned-child record lifecycle: create one, delete one

### LINKED ISSUE #350: [GAP] owned child record lifecycle — a singular one can't be created where the parent has none, and can't be deleted at all

Avick3110:
Stacked on #548 (`claude/polymorphic-refusal`) — review that one first; this branch's diff against it is the owned-child work alone.

A cell's terrain and a worldspace's top cell are single owned child records, and neither end of their life worked. A cell carrying no `Landscape` could not be given one: `parent=` resolved into child *collections* only, filtering to `IList`-shaped properties before it ever asked about the child type, so a slot holding exactly one record was invisible to create while `ChildBearingProperties`, `SchemaClassifier` and the read side could all see it. And an existing one could not be deleted: Mutagen's typed `Remove(FormKey, Type)` routes by group, a singular child sits in none, so the removal found nothing to drop and returned without throwing — the silent no-op the survivor check caught and reported as an engine inconsistency.

Both halves land on the record axis, as values on the verbs that already exist. `create` with `parent=` now resolves either shape of child slot; `collection=` names a singular slot exactly as it names a collection, and the shape only decides how the child is attached, which is the engine's business rather than the caller's. `remove` with the child's own FormID now reaches it: a new `OwnedChildLifecycle` finds the slot holding a given child and detaches it — clearing the property for a singular slot, dropping the element for a collection one — and both remove lanes run it between the typed remove and the survivor check, so it reaches only what the typed remove left behind and the survivor check still speaks for every ordinary record.

Neither slot set is a type list. Both come from the same reflected child-bearing walk the forward lanes preserve children with, so a Mutagen bump that adds a child-bearing property is covered without an edit here, and a test asserts every such property over every modeled record type classifies as one of the two shapes.

Two boundaries the shape creates, both refused rather than guessed. A singular slot that already holds a child is refused — appending is not available and overwriting would drop the record that is there, with everything under it, as a side effect of a call that said "create" — and the refusal names the two real moves. A `Cell` under a `Worldspace` has two routes that build different things, its single `TopCell` and its coordinate-keyed block tree, so one reflected slot matching is not enough to resolve it; silently taking the slot would build a top cell for the far commoner request whose `grid=` was forgotten.

The three field-level refusals that used to end at "an open gap (#350)" now name these calls, and the probe arm that pinned the old contract — CopyFrom's refusal must NOT mention `housecarl_create` — asserts the new one. Field-level `Set`/`Remove`/`CopyFrom` on the slot stay refused: a record is not a field's value, and that rule has not changed. What changed is that their remedies now exist.

Eleven tests, engine-level and world-free since both halves are decided by reflection. One of them pins the measurement this work rests on: the typed remove silently misses a singular owned child and does not throw, so a Mutagen bump that fixes it surfaces here rather than leaving dead code behind.

1182 tests pass, `ci-all` 128/128.

Closes #350

### DIFF
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -307,15 +307,15 @@ public sealed class CorpusRulebook
         // under it — a Worldspace's TopCell takes its persistent references with it. The list form of the same family
         // does delete an owned child (Cell.Persistent Remove key=0), but by INDEX, so the caller names which one; a
         // keyless clear of a singular child deletes a whole subtree implicitly, in one call, with no backup on the
-        // in-place lane. Deliberate owned-child deletion is an open gap, #350.
+        // in-place lane. Deliberate owned-child deletion is on the RECORD axis, where the caller names the record.
         if (string.Equals(req.Verb, "Remove", StringComparison.Ordinal)
             && SchemaClassifier.IsOwnedChildRecord(leaf, _corpus))
             return $"'{leaf.Name}' on '{leafOwner.Name}' holds an owned child RECORD ({leaf.TypeRef}): clearing the " +
                    "field deletes that record and every record under it, implicitly and in one call. The list form of " +
                    "this family deletes a child too, but by INDEX — you name which one; there is no such target here. " +
-                   "Deliberate deletion of an owned child record is an open gap (#350), not something clearing a " +
-                   $"field decides. (To see which record this is: read the parent at depth=2 — the '{leaf.Name}' " +
-                   "field shows the child's FormID.)";
+                   "Delete it on the record axis instead, where the record is named: " + ToolNames.Remove +
+                   " with the child's own FormID. (To see which record this is: read the parent at depth=2 — the " +
+                   $"'{leaf.Name}' field shows the child's FormID.)";
 
         // (3b) record identity (FormKey/ModKey) is a flat, honest reject regardless of Mutagen's setter.
         if (leaf.IsIdentity)
@@ -454,8 +454,9 @@ public sealed class CorpusRulebook
         // owned child records "coercible values". Answered from the same classification the collection verbs decide
         // with, and NOT verb-scoped: a record is not built from parts under any verb.
         //
-        // The two shapes carry different remedies, so they cannot share one sentence: create with parent= is refused
-        // for a singular child (which routes to the open gap #350) and is exactly the shape the collection serves.
+        // The two shapes carry different remedies, so they cannot share one sentence: both are created with parent=,
+        // but a singular slot holds exactly one and a collection takes another, so what to do about an occupied
+        // parent differs and the sentences say so separately.
         if (IsOwnedChildRecordCollection(leaf))
             return OwnedChildRecordCollectionRefusal(leaf);
         if (leaf.Cardinality != "list")
@@ -496,10 +497,10 @@ public sealed class CorpusRulebook
     /// Everything else — scalar/enum/value, formlink, formlink/modeled list, sub-struct, polymorphic arm —
     /// WriteEngine.CopyField transplants by construction.
     /// <para/>
-    /// The singular and collection arms word their remedies differently because only one remedy works per shape:
-    /// <c>housecarl_forward</c> carries the child record itself across for both, but <c>create</c> with
-    /// <c>parent=</c> is refused for a singular child, so only the collection arm may offer it; the singular case
-    /// routes to the lifecycle gap #350 instead.</summary>
+    /// The singular and collection arms word their remedies differently because the shapes differ in what a caller
+    /// does next: <c>housecarl_forward</c> carries the child record itself across for both, and <c>create</c> with
+    /// <c>parent=</c> authors a fresh one for both — but a singular slot names itself with <c>collection=</c> and
+    /// holds exactly one, so the singular sentence points at the slot by name.</summary>
     string? CopyFromLegality(FieldSchema leaf, TypeSchema owner, FieldSchema? ownedChildHop = null, TypeSchema? hopOwner = null)
     {
         if (leaf.IsIdentity)
@@ -526,8 +527,9 @@ public sealed class CorpusRulebook
                    "FIELD's value, not a record — copying it here would write another plugin's record, with its own " +
                    "FormID and everything under it, in as this parent's child. To carry that record across from " +
                    "another plugin use " + ToolNames.Forward + " on the CHILD record itself; read the parent at " +
-                   $"depth=2 and the '{leaf.Name}' field shows the child's FormID. Giving a parent a child it does " +
-                   "not have is an open gap (#350).";
+                   $"depth=2 and the '{leaf.Name}' field shows the child's FormID. To give a parent a child it does " +
+                   "not have, create one on the record axis: " + ToolNames.Create + " with parent= the parent's " +
+                   $"FormID and collection='{leaf.Name}' in its records= element.";
         // The same recogniser as the other two collection doors; the SENTENCE stays this door's own, because
         // CopyFrom's remedy is housecarl_forward (carrying across a record that already exists) rather than
         // housecarl_create with parent= alone. Shared predicate, per-door remedy.
@@ -554,7 +556,8 @@ public sealed class CorpusRulebook
         $"'{leaf.Name}' holds an owned child RECORD ({leaf.TypeRef}): a record is not a part of its parent, so it is " +
         $"neither built from parts (compose= / composes=) nor set from a value (value=). {AddressChildByFormId(leaf.Name)} A path through " +
         "the parent reaches a child only when the record being written already carries one, which a patch's fresh " +
-        "override of a parent never does; giving a parent a child it lacks is an open gap (#350).";
+        "override of a parent never does; to give a parent a child it lacks, create one on the record axis — " +
+        ToolNames.Create + $" with parent= the parent's FormID and collection='{leaf.Name}'.";
 
     /// <summary>How an owned child record that ALREADY EXISTS is written: on the record axis, by its own FormID.
     /// One sentence, shared by the value-shaped Set refusal and the element remedy, so the two doors cannot drift
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -314,8 +314,9 @@ public sealed class CorpusRulebook
                    "field deletes that record and every record under it, implicitly and in one call. The list form of " +
                    "this family deletes a child too, but by INDEX — you name which one; there is no such target here. " +
                    "Delete it on the record axis instead, where the record is named: " + ToolNames.Remove +
-                   " with the child's own FormID. (To see which record this is: read the parent at depth=2 — the " +
-                   $"'{leaf.Name}' field shows the child's FormID.)";
+                   " with the child's own FormID, plus the FormIDs of any records that child carries — the delete " +
+                   "refuses until every record it would drop is named. (To see which record this is: read the " +
+                   $"parent at depth=2 — the '{leaf.Name}' field shows the child's FormID.)";
 
         // (3b) record identity (FormKey/ModKey) is a flat, honest reject regardless of Mutagen's setter.
         if (leaf.IsIdentity)
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -1268,14 +1268,24 @@ public static class WriteEngine
     /// the model, not a hand-coded per-family list. Every false (no such containment, an ambiguous unnamed target, an
     /// unknown collection name) names what it checked.</summary>
     public static bool CanCreateNested(string childCatalogName, Type parentType, string? collectionName, out string? reason)
+        => TryResolveChildSlot(childCatalogName, parentType, collectionName, out _, out _, out reason);
+
+    /// <summary>The same question <see cref="CanCreateNested"/> asks, answering with WHICH slot resolved and what
+    /// SHAPE it has. Pre-flight needs both: a singular slot holds exactly one child, so whether the parent's real
+    /// body already fills it is knowable before anything is allocated.</summary>
+    public static bool TryResolveChildSlot(string childCatalogName, Type parentType, string? collectionName,
+        out string? slotName, out OwnedChildShape shape, out string? reason)
     {
+        slotName = null; shape = OwnedChildShape.None;
         var childType = ResolveConcreteRecordType(childCatalogName);
         if (childType is null)
         {
             reason = $"'{childCatalogName}' is absent from the Mutagen corpus — a real coverage gap to surface, not a value to guess.";
             return false;
         }
-        return TryFindNestedCollection(parentType, childType, collectionName, out _, out _, out reason);
+        if (!TryFindChildSlot(parentType, childType, collectionName, out var prop, out shape, out _, out reason)) return false;
+        slotName = prop!.Name;
+        return true;
     }
 
     /// <summary>Find the parent type's SETTABLE child-collection to allocate a <paramref name="childType"/> into — the
@@ -1330,13 +1340,20 @@ public static class WriteEngine
         }
         var parentName = parentType.Name;   // concrete class name, already clean (never an I…Getter here)
         var childName = childType.Name;
+        // A parent may ALSO file this child by COORDINATE, in a block tree no slot name reaches — the route
+        // AddExteriorCell serves. Derived by asking whether any child-bearing property reaches the child type
+        // through containers rather than holding it, so no record type is named and a Mutagen bump that nests
+        // another child that way is covered without an edit.
+        bool coordinateRoute = ReachesChildThroughContainers(parentType, childType);
         if (hits.Count == 0)
         {
             error = $"'{childName}' cannot be created under a '{parentName}' — that parent models no child slot that " +
                     "holds it (a real containment boundary, surfaced not guessed, Q3)." +
-                    (childType == typeof(Cell)
-                        ? " A Cell is coordinate-keyed, not collection-nested: create an EXTERIOR cell with parent=<Worldspace> + grid=<X,Y>, or an INTERIOR cell with no parent."
-                        : "");
+                    (coordinateRoute
+                        ? $" A {childName} is filed by coordinate under a {parentName}, not held in a slot: create it with parent=<{parentName}> + grid=<X,Y>."
+                        : childType == typeof(Cell)
+                            ? " A Cell is coordinate-keyed, not collection-nested: create an EXTERIOR cell with parent=<Worldspace> + grid=<X,Y>, or an INTERIOR cell with no parent."
+                            : "");
             return false;
         }
         if (collectionName is not null)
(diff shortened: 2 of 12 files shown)


src/housecarl-core/WriteEngine.cs
```diff
@@ -1285,23 +1314,44 @@ static bool TryFindNestedCollection(Type parentType, Type childType, string? col
...
             }
-            prop = named; return true;
+            (prop, shape) = named[0]; return true;
+        }
+        // A CELL under a parent that also files cells by COORDINATE is ambiguous even at one hit, and the two
+        // routes are not interchangeable: a worldspace's TopCell is one record, its exterior cells are a block
+        // tree keyed by grid. Resolving the single slot silently would build the wrong one for the far commoner
+        // request (an exterior cell whose grid= was forgotten), so it is named rather than guessed — the same Q3
+        // rule the several-hits arm below applies, on a route the reflected slot set cannot see.
+        if (childType == typeof(Cell) && hits.Count == 1 && HasCoordinateKeyedCells(parentType))
```
Avick3110: **[high] The `collection=TopCell` route this arm points at is unreachable from the tool.**

`WritePatchBuilder` Phase 1 intercepts *every* spec whose `RecordType` is `Cell` before it ever calls `CanCreateNested`:

```
if (IsCellType(s.RecordType))
{
    if (s.Grid is null) { problems.Add($"Cell '{s.EditorId}': a Cell with parent= but no grid= is ambiguous ..."); continue; }
    ...
    cellKinds[i] = CellCreate.Exterior;
}
else if (!WriteEngine.CanCreateNested(...))
```

(`WritePatchBuilder.cs` ~2949-2956, and Phase 3 dispatches on `cellKinds`, never reaching `NestedAddNew` for a Cell.)

So:

* This ambiguity message never renders on the production path — the only caller that can reach it is a direct `CanCreateNested` call, i.e. the new unit tests.
* A caller who *does* see it (or who reads the new refusal text in `CorpusRulebook`) and runs `housecarl_create parent=<Worldspace> collection=TopCell` with no `grid=` gets refused by the Phase-1 branch with "a Cell with parent= but no grid= is ambiguous" — advice that contradicts the advice they were just given.
* The CHANGELOG entry ("a worldspace's top cell ... could not be given any ... Both now work on the record axis") is therefore only half true: `Landscape` works, `TopCell` does not.

Either route the `IsCellType` branch through the slot resolver when `s.IntoCollection` is set, or drop the TopCell half of the claim from the ambiguity text, the rulebook refusals and the changelog.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -516,7 +518,8 @@ static string OwnedChildSetRefusal(FieldSchema leaf) =>
         "neither built from parts (compose= / composes=) nor set from a value (value=). Address the child record " +
         $"itself by its own FormID — read the parent at depth=2 and the '{leaf.Name}' field shows it. A path through " +
         "the parent reaches a child only when the record being written already carries one, which a patch's fresh " +
-        "override of a parent never does; giving a parent a child it lacks is an open gap (#350).";
+        "override of a parent never does; to give a parent a child it lacks, create one on the record axis — " +
+        ToolNames.Create + $" with parent= the parent's FormID and collection='{leaf.Name}'.";
```
Avick3110, line 561: **[medium] For `Worldspace.TopCell` this remedy is refused by the call it names.**

`OwnedChildSetRefusal` (and the twin sentence in `CopyFromLegality`, ~line 490) is reached for every singular owned child, which is `Cell.Landscape` **and** `Worldspace.TopCell`. For `TopCell` the rendered advice is:

> ... create one on the record axis — `housecarl_create` with parent= the parent's FormID and collection='TopCell'.

Running that fails: `WritePatchBuilder` Phase 1 refuses any `Cell` spec carrying `parent=` and no `grid=` before the slot resolver runs (`a Cell with parent= but no grid= is ambiguous`). The old "#350 open gap" text was at least accurate for this leaf; the new text sends the caller into a refusal loop.

Same for the `WriteEngine.cs` ~2764 `ExpectedApplyRejectionException`, which interpolates `collection='{segment}'` and hits `TopCell` the same way.

Either make the create lane honour `collection=` for a Cell, or word these so the TopCell case is not promised.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1294,6 +1294,8 @@ static RemovalOutcome RemoveRecordsCore(
                 $"present-check passed but Remove threw — a real engine inconsistency, surfaced not swallowed (Q3): "
                 + $"{ex.GetType().Name}: {ex.Message}");
         }
+        if (DetachOwnedChildren(patchMod, toRemove) is { } detachFailed)
```
Avick3110, line 1301: **[medium] The detach deletes a whole subtree, and the removal report accounts for one record.**

`DetachOwnedChildren` clears the parent's property for a singular slot. For `Cell.Landscape` that is one record and harmless, but the same path now serves `Worldspace.TopCell` — and clearing `TopCell` drops every persistent reference filed under it, in one call, with no per-record accounting: `RemovalOutcome` is built from `toRemove`, which holds only the FormIDs the caller named. The user is told "removed 1 record" while N are gone (only the raw `remaining` count hints at it).

That is precisely the hazard `CorpusRulebook`'s field-lane refusal cites as the reason it will not clear the field — "clearing the field deletes that record and every record under it, implicitly and in one call" — and that refusal now redirects the caller here. Before this PR the record axis failed loud on the survivor check; now it succeeds quietly.

Suggest either enumerating the descendants of a detached child into the removal report, or refusing a singular detach whose child carries children of its own unless the caller names them too.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1331,6 +1333,31 @@ static RemovalOutcome RemoveRecordsCore(
     /// effect and only the post-write verify catches it. Checking the mutable mod first keeps a no-op loud with the
```
Avick3110: **[low] This block was inserted between `RemoveSurvivors`' doc comment and the new method, so `DetachOwnedChildren` now carries two `<summary>` tags and `RemoveSurvivors` carries none.**

The preceding summary ("IN-MEMORY absence verify run BEFORE any serialize ... Null ⇒ all targets gone") describes `RemoveSurvivors`, but it is now the first of two summaries on `DetachOwnedChildren`. A reader (or an IDE tooltip) on `DetachOwnedChildren` gets the wrong description first, and `RemoveSurvivors` — the method the new code is explicitly reasoning about — is undocumented. Move the new summary above `DetachOwnedChildren` only and leave the old one attached to `RemoveSurvivors`.

While here: the comment above the typed `Remove` loop (~line 1281) still claims it "reaches NESTED records (Cell/Placed*/INFO/Navmesh/Landscape) too". This PR's own `TheTypedRemoveDoesNotReachASingularOwnedChild` test measures that `Landscape` is not reached — that clause should go, or the new detach path reads as redundant to anyone trusting the comment.

Avick3110:
Blind review pass, second half — four more findings on top of the four inline ones.

**[medium] `src/housecarl-core/WriteEngine.cs:1409` — the occupied-slot guard reads the patch's copy of the parent, and nothing measures whether that copy carries the child.**

`if (shape == Singular && prop.GetValue(parentInPatch) is IMajorRecordGetter occupant)` asks the parent *in the patch*, which on the `into=`/new-patch lane is whatever `MakeSettableParent` → `GenericGetOrAddAsOverride` produced. Two places in the tree say opposite things about what that copy holds:

- `CorpusRulebook.OwnedChildSetRefusal` and `WriteEngine.cs:~2764` both tell the caller, as a fact, that "a patch's fresh override of a parent does not bring the parent's child records with it".
- `NestedCreateProof` N2 measures the opposite for a `DialogTopic`: the overridden topic arrives "carrying its original lines", and the arm asserts the original responses survive.

If the refusal text is the true one for a `Cell`, the guard never fires on the ordinary lane: `create Landscape parent=<a cell that already has terrain>` sees `null`, allocates, and the patch ships a cell carrying a second, empty LAND — which is precisely the "drop the record that is there, with everything under it" outcome the guard's own message says it exists to prevent. Only the in-place lane, where the parent really is the record that holds the child, is protected. The occupancy is knowable at pre-flight from the winner body Phase 1 already resolves (`parentPlans[i].body`), which would also put it in the batched "NOTHING created" problem list instead of the Phase-3 catch that is documented as the engine-inconsistency channel. Either way this needs a measurement, and the tests do not have one: every create test builds its parent with `AddInteriorCell` in the same in-memory mod.

**[medium] `src/housecarl-core/WriteEngine.cs:1330` — `childType == typeof(Cell)` is a named record type deciding behaviour, and it is derivable.**

The structural distinction is real — a single slot and a coordinate-keyed block tree are different routes and must not be silently collapsed — but the test for it is half reflected and half hand-written. `HasCoordinateKeyedCells` asks the model (a child-bearing property whose list element is not itself a record) and then throws the element type away, so it cannot tell which child that block tree actually leads to; the `typeof(Cell)` literal is doing the coverage work the reflection was supposed to do. Deriving it — does any child-bearing property *reach* `childType` through non-record containers, the same question `ReachesOwnedRecord` already answers — removes the named type and makes the arm correct for whatever Mutagen nests next. As written it also silently stops applying when `hits.Count > 1`: the several-hits message lists the slots and never mentions `grid=`, so a second Cell-typed slot would drop the second route from the sentence.

**[low] `src/housecarl-mcp-tests/OwnedChildLifecycleTests.cs:171` — the derivation test guards a different walk from the one create uses.**

`EveryChildBearingPropertyClassifiesAsOneOfTheTwoShapes` iterates `WriteEngine.ChildBearingProperties`, but the create lane does not: `TryFindChildSlot` runs its own `GetProperties` with its own predicates (`GetGetMethod() != null`, `CanWrite`, `IMajorRecordGetter`-typed, element-assignable + `IList`). The two agree today on the singular half and deliberately differ on the collection half, which is exactly why a drift test is worth having — but the drift it would catch is on the walk create does not call. Asserting through `CanCreateNested`/`TryFindChildSlot` for every child-bearing property would pin the thing the lane depends on.

**[low] `src/housecarl-mcp-tests/OwnedChildLifecycleTests.cs:1` — engine-level only is not enough for this change, and the gap is where the first finding landed.**

Eleven unit-tier tests call `WriteEngine.NestedAddNew`, `OwnedChildLifecycle.TryFindSlot` and `Detach` directly. Nothing drives either half through the surface:

- `DetachOwnedChildren` — the whole point of the delete half — has no test at all. Neither `RemoveRecords` nor `RemoveRecordsInPlaceCore` is ever called with an owned-child target, so the ordering against the survivor check, the failure text, and the "file UNTOUCHED" claim are unexercised.
- The create half never goes through `WritePatchBuilder.CreateRecords`, so a load-order parent, the override copy, the serialize and the re-open are all untested — and it is Phase 1, not the engine, that makes the advertised `collection=TopCell` route unreachable.
- `NestedCreateProof` exists on the stated ground that "the Mutagen primitive in isolation proves nothing about the tool" and covers N1–N7 for the collection shape. The singular shape adds no arm. An N8 (create a `Landscape` under an existing cell → re-open, it is there; and under a cell that already has one → refused) would have caught the top finding and would settle the override-carries-children question above.
- The integration-tier `OwnedChildRemedyGrammarTests` / `OwnedChildWorld` already derive their subject set from the same child-bearing walk and grade remedy sentences; the three refusals this PR rewrote to name `housecarl_create`/`housecarl_remove` are graded there by shape, but nothing checks that the named call actually accepts the sentence's own arguments — which for `TopCell` it does not.

Avick3110:
Review fold — all eight findings addressed. Seven folded, one folded with a correction to its supporting evidence.

**[high] the `collection=TopCell` route is unreachable — FOLDED.** `WritePatchBuilder` Phase 1 now intercepts a Cell spec only when it carries `grid=`; every other cell spec with a parent goes through the slot resolver like any other child, so `collection='TopCell'` reaches `NestedAddNew` and works. Naming `grid=` and `collection=` together is refused rather than ranked. The changelog claim is now true for both leaves, and it says which token picks which route. Measured end to end: the new `OwnedChildLifecycleDrivenTests.TheTopCellRouteTheRefusalsNameIsReachableThroughTheTool` (a worldspace with no top cell gets one, re-opened from disk) and the proof's N12 against Skyrim.esm.

**[medium] the rewritten refusals route to a dead route — FOLDED** by the above; the sentences are now accurate for `Worldspace.TopCell` as well as `Cell.Landscape`, and `create`'s own `collection=` description says a single-child slot is a legal value.

**[medium] the detach deletes a subtree the report accounts one record for — FOLDED, by refusing rather than by reporting.** Of the two remedies you offered I took the second: a detach whose child still carries records the call did not name is refused, naming them and how many. Reporting afterwards would still delete records the caller never asked for; refusing keeps the property the rulebook's field-lane refusal cites as the list family's virtue — the caller names what dies — and when they do name them all, every one is in `Removed`, so the accounting is exact by construction. Driven test: `RemovingAnOwnedChildThatCarriesUnnamedRecordsIsRefusedUntilTheyAreNamedToo` (refused, then both named, `removed=2`, slot empty on re-open).

**[medium] the occupied-slot guard reads the patch's copy — FOLDED, with one correction.** The check now runs at pre-flight against `parentPlans[i].body` (the definer/winner body Phase 1 already resolves), so it lands in the batched "NOTHING created" list; the Phase-3 guard stays as a backstop and its comment says so. A same-call double claim on one singular slot is refused there too.

The correction is to the evidence, not the finding: N2 does **not** assert the original responses survive — the arm's own comment says it asserts the mechanism only, and the run reports `override-carries=1 of 2`. What you read was a stale line in the file's header doc, which I have fixed. So the refusal text is the true one — a fresh override carries no children — which is exactly why the old guard could never fire on the `into=` lane. Finding stands, measured: `AnOccupiedSingularSlotIsRefusedEvenThoughThePatchCopyCarriesNoChild` uses CellA, whose winner declares nothing and whose definer declares the LAND, plus the proof's N11.

**[medium] `childType == typeof(Cell)` and the `hits.Count > 1` cliff — FOLDED.** The arm now asks whether the parent reaches the child type *through containers* (`ReachesChildThroughContainers`, the same walk and bound as `ReachesOwnedRecord`, asked of one target type), and the coordinate route is folded into the several-routes message, so it applies at any hit count and lists every route. No record type decides behaviour there. The one remaining `typeof(Cell)` is the pre-existing hint on the zero-slot arm (create a Cell under a Weapon → the interior-cell route), which fires only when no coordinate route exists.

**[low] doc-comment placement and the stale reach claim — FOLDED.** `RemoveSurvivors` has its summary back, `DetachOwnedChildren` has one of its own, and all three comments claiming the typed `Remove` reaches a Landscape now say it does not and point at the detach path.

**[low] the derivation test walks a different set — FOLDED.** It now asserts through `CanCreateNested` for every child-bearing property: a property that holds a record must be nameable as a slot for that record, and a container tree must be refused by name and, asked with no discriminator, name `grid=`. Both halves must be non-empty for the test to pass.

**[low] no tool-surface or NestedCreateProof coverage — FOLDED.** New `OwnedChildLifecycleDrivenTests` (integration tier, 7 tests) drives both halves through `LoadOrderService` against a real load order, and `NestedCreateProof` gains N10 (singular child created), N11 (occupied slot refused), N12 (neither cell route named), N13 (the child removed by its own FormID, slot empty on re-open). N8/N9 were already taken, hence the numbering.

Nothing refuted outright.

Green: build 0 errors, `dotnet test` 1189/1189, `ci-all` 128/128, `nested-create-proof` 13/13 arms against Skyrim.esm (byte-untouched).

Avick3110:
Blind review folded (8 of 8 real; one factual claim in finding 4 corrected, the finding itself folded). Queued for Aaron's review; the refuse-vs-report call on a subtree delete is on this PR.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2937,17 +2985,38 @@ static CreateOutcome CreateRecordsCore(
...
-                else if (!WriteEngine.CanCreateNested(s.RecordType, parentType, s.IntoCollection, out var nestedWhy)) { problems.Add($"{s.RecordType} '{s.EditorId}': {nestedWhy}"); continue; }
+                else if (!WriteEngine.TryResolveChildSlot(s.RecordType, parentType, s.IntoCollection, out var slotName, out var slotShape, out var nestedWhy))
+                { problems.Add($"{s.RecordType} '{s.EditorId}': {nestedWhy}"); continue; }
+                // A SINGULAR slot holds exactly one child, so an occupied one is not something create can resolve —
+                // and occupancy is knowable HERE, from the parent body this call will host in, before a FormKey is
+                // allocated. Reading the patch's copy instead would answer for whatever the override carried, which
+                // on the into= lane is not the parent the caller is looking at.
+                else if (slotShape == OwnedChildShape.Singular)
+                {
+                    var body = parentPlans[i]!.Value.body ?? (IMajorRecordGetter?)parentPlans[i]!.Value.patchParent;
```
Avick3110: **[high] On the `into=` / already-carried lane this guard reads the very copy it was written to see past, so a second singular child can still be created.**

`body ?? patchParent` resolves to the destination's own record whenever `AlreadyCarried` matched — and that record is the child-less fresh override N11 and `AnOccupiedSingularSlotIsRefusedEvenThoughThePatchCopyCarriesNoChild` were measured against.

Sequence:

1. `housecarl_create` a `PlacedObject` into cell `000ABC:Skyrim.esm` (which declares a LAND), `patch=P`. `GenericGetOrAddAsOverride` copies the cell in without its children, so `P`'s Cell 000ABC has `Landscape == null`.
2. `housecarl_create record_type='Landscape' parent='000ABC:Skyrim.esm' into=P`. Phase 1 takes the `AlreadyCarried` branch (`parentPlans[i] = (null, null, null, already)`), so `body` is null and this line falls through to `patchParent` — `P`'s own copy. `OccupantOf` returns null; the slot reads free.
3. `NestedAddNew`'s backstop (`WriteEngine.cs:1443`) asks the same copy, also sees null, allocates.

`P` now ships a cell override declaring a brand-new LAND while `Skyrim.esm`'s cell declares its own — exactly the duplicate-terrain outcome the fresh-patch arm refuses.

What makes it more than a corner: the rewritten `MaterializeSubstruct` refusal (`WriteEngine.cs:2803`) tells a caller to run this exact call, and that refusal fires precisely when the patch already carries the parent override. Followed literally, the sentence produces the duplicate.

The in-place branch above has the same shape: a target plugin that overrides the cell without carrying its LAND reads as a free slot.

The real body is knowable in both branches — the parent's load-order winner/definer is still resolvable by `parentFk`. Consulting it (and treating the carried copy as the additional, not the only, source) would make the answer the same on every lane.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1348,15 +1430,30 @@ public static IMajorRecord NestedAddNew(SkyrimMod patchMod, IMajorRecord parentI
...
+                $"nested create: '{parentName}.{prop.Name}' already holds a {childType.Name} ({occupant.FormKey}" +
+                (occupant.EditorID is { } oe ? $" editorid={oe}" : "") + ") and it holds exactly one, so there is no " +
+                "room to create another. Edit the one that is there by its own FormID, or remove it first with " +
+                ToolNames.Remove + " and create again.");
 
         EnsureFormIdFloor(patchMod);   // a rehydrated (into=) counter below 0x800 would hand out engine-reserved IDs
         EnsureAllocatable(patchMod);
         var child = ConstructRecord(childType, AllocateNextFormKey(patchMod));
         if (editorId is not null) child.EditorID = editorId;
+        if (shape == OwnedChildShape.Singular) { prop!.SetValue(parentInPatch, child); return child; }
```
Avick3110, line 1477: **[medium] The singular-slot route creates a `Cell` without the cell-EditorID dedup both other cell routes run.**

`AddInteriorCell` and `AddExteriorCell` both open with `EnsureNoDuplicateCellEditorId`, whose own comment says why: cells carry a stable EditorID, cell create does not upsert, so an `into=` re-run would otherwise silently append a second cell at the same identity. `collection='TopCell'` is now a third route to a `Cell` and it comes through `NestedAddNew`, which never asks.

Sequence:

1. `housecarl_create record_type='Cell' editorid='MyCell'` (no parent) into patch `P` — an interior cell.
2. `housecarl_create record_type='Cell' editorid='MyCell' parent=<Worldspace> collection='TopCell' into=P`.

Phase 1's `seenEdid` set is per-call, so nothing catches it; `P` carries two cells named `MyCell`. The occupancy guard does not cover this — the collision is with a different cell, not with the slot.

The cheap fix is to run the same guard on the singular path when the constructed type is a `Cell`; the general one is to run it from `NestedAddNew` for any child type that carries a stable EditorID.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -268,15 +268,15 @@ public static Corpus LoadCorpus(string corpusJsonPath)
...
+        // in-place lane. Deliberate owned-child deletion is on the RECORD axis, where the caller names the record.
         if (string.Equals(req.Verb, "Remove", StringComparison.Ordinal)
             && SchemaClassifier.IsOwnedChildRecord(leaf, _corpus))
             return $"'{leaf.Name}' on '{leafOwner.Name}' holds an owned child RECORD ({leaf.TypeRef}): clearing the " +
                    "field deletes that record and every record under it, implicitly and in one call. The list form of " +
                    "this family deletes a child too, but by INDEX — you name which one; there is no such target here. " +
-                   "Deliberate deletion of an owned child record is an open gap (#350), not something clearing a " +
-                   $"field decides. (To see which record this is: read the parent at depth=2 — the '{leaf.Name}' " +
-                   "field shows the child's FormID.)";
+                   "Delete it on the record axis instead, where the record is named: " + ToolNames.Remove +
```
Avick3110, line 316: **[low] For `Worldspace.TopCell` the remedy this sentence names refuses on the first try.**

This refusal renders for every singular owned child, which is `Cell.Landscape` **and** `Worldspace.TopCell`. For a top cell, `housecarl_remove` with the child's own FormID hits `DetachOwnedChildren`'s subtree guard and refuses with "you named none of them", because a top cell carries the worldspace's persistent references.

So a caller who follows the sentence exactly gets a refusal, then has to read the subtree and re-issue. The create arm of this same pair names its discriminator (`collection='...'`) so the call it points at works first time; this one could say the equivalent — that a child carrying records of its own needs those named in the same call — and keep the parity.

(Not a claim about the refuse-vs-report decision itself, which is on this PR for Aaron. Only that the sentence promises a one-call fix that is one call for `Landscape` and two for `TopCell`.)

src/housecarl-core/WriteEngine.cs
```diff
@@ -1259,47 +1269,119 @@ public static bool CanCreateNested(string childCatalogName, Type parentType, str
     /// null on success.</summary>
     static bool TryFindNestedCollection(Type parentType, Type childType, string? collectionName,
         out PropertyInfo? prop, out List<string> matches, out string? error)
+        => TryFindChildSlot(parentType, childType, collectionName, out prop, out _, out matches, out error);
```
Avick3110: **[low] `TryFindNestedCollection` has no callers left.**

`WritePatchBuilder` moved to `TryResolveChildSlot` and `NestedAddNew` calls `TryFindChildSlot` directly, so this wrapper is now unreachable private code — which the compiler will not flag. The two surviving `<see cref="TryFindNestedCollection"/>` references (lines 1229 and 1426) also point a reader at the wrapper rather than at `TryFindChildSlot`, where the singular-slot behaviour they describe actually lives.

Dropping it and re-pointing the two crefs leaves one name for one resolver.

Avick3110:
Aaron's review folded — all four findings real, nothing refuted.

**[high] the occupancy guard reads the carried copy on the into=/in-place lanes — FOLDED.** Verified: on the `AlreadyCarried` branches `parentPlans[i]` is `(null, null, null, already)`, so `body ?? patchParent` was the patch's own override, which carries none of the parent's children. Pre-flight now asks BOTH copies — the carried one and the parent's real body in the order (`OrderBodyOf`: the definer's copy, else the winner's, on the same memo). Same answer on every lane, including in-place. Measured two ways: `AnOccupiedSingularSlotIsRefusedWhenThePatchAlreadyCarriesTheParent` (a placed ref puts the cell in the patch, then a Landscape under the same cell is refused, `NOTHING created`) and the proof's new N14 against Skyrim.esm (`already holds ... (00A0E3:Skyrim.esm)`, no LAND in the patch). Both fail without the change.

**[medium] the singular route skips the cell-editorid dedup — FOLDED.** Reproduced. `EnsureNoDuplicateCellEditorId` is now `EnsureNoDuplicateEditorId(patchMod, type, editorId)` and `NestedAddNew` runs it for any SINGULAR slot rather than only for the two coordinate-keyed cell routes, so no record type decides the behaviour and the rule reads: a slot that does not upsert does not let a second record at the same editorid in behind it. The COLLECTION route keeps its documented append (those children carry no stable editorid handle). Test: `ACellThroughTheSingularSlotGetsTheSameEditorIdDedupAsTheOtherCellRoutes` — an interior `MyCell`, then `parent=<Worldspace> collection='TopCell'` with the same editorid, refused. Fails without the change. One clause added to the changelog entry.

**[low] the remove remedy refuses on the first try for a top cell — FOLDED.** The field-lane Remove refusal now says to name the child's own records in the same call and why ("the delete refuses until every record it would drop is named"), so it has the same first-try-works property as its create twin. Test: `TheFieldLaneRemoveRefusalSaysToNameTheRecordsUnderTheChild`, asked of `Worldspace.TopCell`.

**[low] `TryFindNestedCollection` has no callers — FOLDED.** Dropped; the two `<see cref>`s and the header comment now point at `TryFindChildSlot`. One name for one resolver.

Green: build 0 errors, `dotnet test` 1212/1212, `ci-all` 128/128, `nested-create-proof` 14/14 arms against Skyrim.esm (byte-untouched).

PR #558, opened 2026-09-05
Add the *parent containment step

### LINKED ISSUE #459: [GAP] no parent-of-this-record read for group-nested containment — DIAL->INFO, CELL->REFR, WRLD->CELL

Avick3110:
Containment becomes the second edge kind beside the form link, spelled `*parent`. DIAL→INFO, CELL→REFR/ACHR and WRLD→CELL ownership is group nesting, so `references=` correctly returns nothing for it and the child body names no owner — only a NavigationMesh does, at `Data.Parent`. The index build therefore captures it: `LoadOrderResolver.BuildIndex` swaps its flat `EnumerateMajorRecords` for `EnumerateMajorRecordContexts`, which yields the identical record set with the containing context attached, climbs group/block contexts to the nearest record ancestor (and stages nothing where the chain holds no record, rather than guessing), and merges one packed child→parent map per plugin, later-wins — the 2,363 children a real order re-parents are what makes per-plugin later-wins the right merge rather than one whole-order map. `*parent` then lands as a value on the axes that already exist: a leading, chaining step on `where=` paths, on `project.fields`, and on a walk's `seed_paths`/`follow`. It is a step, so everything below the hop — the `winner` term, `editorid`, `formid` membership, leaves, quantified steps from #458 — reads the containing record with no second rule. There is no per-record-type parent map anywhere in it: the set is Mutagen's own containment walk, and the refusals name `WriteEngine.ChildBearingProperties` as the reason, derived, so a Mutagen bump grows the reach with no edit here.

Refusals, each one sentence: `*parent` after a field step (it leads a path by definition — the containing record is a property of the record, not of a field value), carrying a quantifier (there is only ever one), with nothing after it, on the left of `->` with nothing after it, and any other `*` token. A record nothing contains is a named no-verdict whose scan rollup says so and names the child-bearing properties — never a silent non-match, and never a null on the read side.

Measured on the live ARR 2.0 order (3,801 plugins, 3.69M records, `parent-in-hand` probe, medians of three alternating runs): the context walk costs **0.31 s more than the flat walk, 1.04×** (7.91 s → 8.22 s), inside the run-to-run spread, and the packed map retains **77.3 MB** for 2,711,713 distinct children at ~30 B/entry. On the synthetic test fixture (3 plugins) the delta is below timer noise. 1,214 tests (baseline 1,190) and 128/128 probes pass.

Not in this PR: #489's reverse-reference index, which is E1 step 3.

Closes #459

### DIFF
--- a/src/housecarl-core/ContainmentIndex.cs
+++ b/src/housecarl-core/ContainmentIndex.cs
@@ -1,0 +1,91 @@
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Plugins.Records;
+using Mutagen.Bethesda.Skyrim;
+
+namespace HousecarlCore;
+
+/// <summary>
+/// The child → parent map: the second edge kind, beside the form link. A DIAL owns its INFOs, a CELL its placed
+/// references and navmeshes, a WRLD its cells — group nesting, not a link, so <c>references=</c> cannot see it and
+/// the child body names nothing (only a NavigationMesh does, at <c>Data.Parent</c>). This is what the
+/// <c>*parent</c> path step reads.
+///
+/// <para><b>Captured at index build, from Mutagen's own containment walk.</b>
+/// <see cref="LoadOrderResolver"/> already enumerates every plugin once; swapping that walk from the flat
+/// <c>EnumerateMajorRecords</c> to <c>EnumerateMajorRecordContexts</c> yields the identical record set with the
+/// containing context attached, for about 0.4 s more over a 3,800-plugin order. There is no per-record-type parent
+/// map here and none is possible: the parent arrives from Mutagen, so a Mutagen bump that adds a child-bearing
+/// container grows <c>*parent</c>'s reach with no edit — the first cornerstone applied to an edge kind.</para>
+///
+/// <para><b>Later plugin wins, like everything else.</b> Across a real order thousands of children end up under a
+/// different parent than the first plugin put them under (a placed reference moved between cells), so the map is
+/// merged per plugin in priority order and the last declaration stands — the same rule the winner index follows.
+/// A plugin that throws part-way through its walk merges nothing, so a half-read plugin never leaves partial
+/// containment behind.</para>
+///
+/// <para><b>Packed.</b> Entries are ulong → ulong ((interned mod index &lt;&lt; 32) | FormID), not FormKey → FormKey:
+/// ~44 B an entry, about 77 MB for the 2.7M children of a large order, of which <c>Cell.Temporary</c> is four
+/// fifths. Pure data — no bodies, no file handles.</para></summary>
+public sealed class ContainmentIndex
+{
+    readonly Dictionary<ulong, ulong> _map = new();
+    readonly Dictionary<ModKey, int> _modToIdx = new();
+    readonly List<ModKey> _idxToMod = new();
+
+    /// <summary>How far a group/block context is climbed to reach the nearest record ancestor. Mutagen hands a
+    /// worldspace's cells up through block and sub-block contexts that carry no record of their own; the deepest
+    /// real chain is two hops. A chain that holds no record inside the bound stages no edge, so <c>*parent</c>
+    /// says it has none rather than guessing at one.</summary>
+    const int MaxGroupHops = 6;
+
+    /// <summary>Distinct children with a recorded parent.</summary>
+    public int Count => _map.Count;
+
+    /// <summary>The containing record of <paramref name="child"/>, or null when this build recorded none — the
+    /// record is top-level, or its context chain held no record ancestor.</summary>
+    public FormKey? ParentOf(FormKey child)
+    {
+        if (!_modToIdx.TryGetValue(child.ModKey, out int i)) return null;
+        if (!_map.TryGetValue(((ulong)(uint)i << 32) | child.ID, out var packed)) return null;
+        return new FormKey(_idxToMod[(int)(packed >> 32)], (uint)packed);
+    }
+
+    /// <summary>Stage one plugin's containment edges, read off the context walk. Returns the pairs for the caller to
+    /// merge only if the whole plugin enumerated — the same plugin-atomic rule the winner index follows.</summary>
+    internal static void Stage(IModContext context, List<(FormKey Child, FormKey Parent)> into)
+    {
+        if (context.Record is not IMajorRecordGetter child) return;
+        var up = context.Parent;
+        for (int hops = 0; up is not null && hops <= MaxGroupHops; hops++, up = up.Parent)
+            if (up.Record is IMajorRecordGetter ancestor) { into.Add((child.FormKey, ancestor.FormKey)); return; }
+    }
+
+    /// <summary>Merge one fully-enumerated plugin's edges, later-wins.</summary>
+    internal void Merge(IReadOnlyList<(FormKey Child, FormKey Parent)> edges)
+    {
+        foreach (var (child, parent) in edges) _map[Pack(child)] = Pack(parent);
+    }
+
+    ulong Pack(FormKey k)
+    {
+        if (!_modToIdx.TryGetValue(k.ModKey, out int i))
+        {
+            _modToIdx[k.ModKey] = i = _idxToMod.Count;
+            _idxToMod.Add(k.ModKey);
+        }
+        return ((ulong)(uint)i << 32) | k.ID;
+    }
+
+    /// <summary>The child-bearing property surface, spelled <c>Type.Property</c> — what a <c>*parent</c> refusal
+    /// names as the reason a record has no containing record. Derived from
+    /// <see cref="WriteEngine.ChildBearingProperties"/> over every concrete record type Mutagen models, the same
+    /// source the write surface's child preservation runs on, so it is never a hand list.</summary>
+    public static string ChildBearingSurface() => _surface ??= string.Join(", ",
+        typeof(Weapon).Assembly.GetTypes()
+            .Where(t => t.IsClass && !t.IsAbstract && !t.Name.EndsWith("BinaryOverlay", StringComparison.Ordinal)
+                        && typeof(IMajorRecord).IsAssignableFrom(t))
+            .SelectMany(t => WriteEngine.ChildBearingProperties(t).Select(p => $"{t.Name}.{p.Name}"))
+            .OrderBy(s => s, StringComparer.Ordinal));
+    static string? _surface;
+}
--- a/src/housecarl-core/ContainmentIndex.cs
+++ b/src/housecarl-core/ContainmentIndex.cs
@@ -14,7 +14,9 @@ namespace HousecarlCore;
 /// <para><b>Captured at index build, from Mutagen's own containment walk.</b>
 /// <see cref="LoadOrderResolver"/> already enumerates every plugin once; swapping that walk from the flat
 /// <c>EnumerateMajorRecords</c> to <c>EnumerateMajorRecordContexts</c> yields the identical record set with the
-/// containing context attached, for about 0.4 s more over a 3,800-plugin order. There is no per-record-type parent
+/// containing context attached, for about 0.3 s more over a 3,800-plugin order — 0.31 s to 0.43 s across repeated
+/// alternating runs on the ARR 2.0 order (3,801 plugins, 3.69M records), which is that machine's own run-to-run
+/// spread; the changelog quotes the low end. There is no per-record-type parent
 /// map here and none is possible: the parent arrives from Mutagen, so a Mutagen bump that adds a child-bearing
 /// container grows <c>*parent</c>'s reach with no edit — the first cornerstone applied to an edge kind.</para>
 ///
@@ -25,8 +27,9 @@ namespace HousecarlCore;
 /// containment behind.</para>
 ///
 /// <para><b>Packed.</b> Entries are ulong → ulong ((interned mod index &lt;&lt; 32) | FormID), not FormKey → FormKey:
-/// ~44 B an entry, about 77 MB for the 2.7M children of a large order, of which <c>Cell.Temporary</c> is four
-/// fifths. Pure data — no bodies, no file handles.</para></summary>
+/// ~30 B an entry, about 77 MB measured for the 2,711,713 children of a large order, of which
+/// <c>Cell.Temporary</c> is four fifths. Pure data — no bodies, no file handles. The count is reported in band on
+/// <c>housecarl_load_order_status</c>, so the cost is declared rather than discovered.</para></summary>
 public sealed class ContainmentIndex
 {
     /// <summary>The one spelling of the containment step, shared by every surface that takes a path — a
@@ -35,15 +38,47 @@ public sealed class ContainmentIndex
     /// takeable.</summary>
     public const string ParentToken = "*parent";
 
+    /// <summary>True when this path segment IS the containment step.</summary>
+    public static bool IsParentStep(string seg) => string.Equals(seg, ParentToken, StringComparison.OrdinalIgnoreCase);
+
+    /// <summary>The ONE grammar check for a <c>*parent</c> run, shared by every surface that takes a path — the
+    /// <c>where=</c> predicate and the <c>project.fields</c> read walk both call this, so an identical mistake
+    /// gets an identical sentence instead of a typo hint on one surface and a named refusal on the other.
+    /// Returns how many hops lead <paramref name="segs"/>, or the refusal (unprefixed, so each surface wraps it
+    /// in its own voice). A hop leads a path by definition — the containing record is a property of the RECORD,
+    /// not of a field value — so a <c>*parent</c> anywhere else refuses by name, as does one carrying a
+    /// quantifier, one with nothing after it, and any other <c>*</c> token.</summary>
+    /// <param name="display">How the caller spelled this side, for the message.</param>
+    /// <param name="isLinkLeft">The left side of a <c>-&gt;</c> step, whose all-hops case wants a link, not a value.</param>
+    /// <param name="allowBare">A walk's seed path or follow, where a path of nothing but hops IS the edge crossed —
+    /// the containing record is the walk's next node, not a value to read.</param>
+    public static (int Hops, string? Error) SplitHops(string[] segs, string display, bool isLinkLeft = false, bool allowBare = false)
+    {
+        int hops = 0;
+        while (hops < segs.Length && IsParentStep(segs[hops])) hops++;
+        for (int i = hops; i < segs.Length; i++)
+        {
+            var s = segs[i];
+            if (IsParentStep(s))
+                return (0, $"'{ParentToken}' is the record that CONTAINS this one, so it can only lead a path — " +
+                           $"in '{display}' it follows a field step. Write the hops first ('{ParentToken}.EditorID').");
+            int open = s.IndexOf('[');
+            if (open > 0 && s.EndsWith("]", StringComparison.Ordinal)
+                && string.Equals(s[..open], ParentToken, StringComparison.OrdinalIgnoreCase))
+                return (0, $"'{s}' — '{ParentToken}' names ONE containing record, not a list, so it takes no quantifier. Write '{ParentToken}'.");
+            if (s.Length > 0 && s[0] == '*' && open != 0)
+                return (0, $"'{s}' is not a path token — the tokens are '{ParentToken}' (the containing record) and the quantifiers [*any], [*all], [*none] and [*count] on a list step.");
+        }
+        if (hops == segs.Length && hops > 0 && !allowBare)
+            return (0, isLinkLeft
+                ? $"'{display}' is the containing record, which is not a link-bearing field — name one on it ('{display}.Quest->editorid')."
+                : $"'{display}' names the containing record, not a value — follow it with a field ('{display}.EditorID').");
+        return (hops, null);
+    }
+
     readonly Dictionary<ulong, ulong> _map = new();
     readonly Dictionary<ModKey, int> _modToIdx = new();
     readonly List<ModKey> _idxToMod = new();
-
-    /// <summary>How far a group/block context is climbed to reach the nearest record ancestor. Mutagen hands a
-    /// worldspace's cells up through block and sub-block contexts that carry no record of their own; the deepest
-    /// real chain is two hops. A chain that holds no record inside the bound stages no edge, so <c>*parent</c>
-    /// says it has none rather than guessing at one.</summary>
-    const int MaxGroupHops = 6;
 
     /// <summary>Distinct children with a recorded parent.</summary>
     public int Count => _map.Count;
@@ -58,12 +93,19 @@ public sealed class ContainmentIndex
     }
 
     /// <summary>Stage one plugin's containment edges, read off the context walk. Returns the pairs for the caller to
-    /// merge only if the whole plugin enumerated — the same plugin-atomic rule the winner index follows.</summary>
+    /// merge only if the whole plugin enumerated — the same plugin-atomic rule the winner index follows.
+    ///
+    /// <para>The climb to the nearest record ancestor is UNBOUNDED, deliberately. Mutagen hands a worldspace's
+    /// cells up through block and sub-block contexts that carry no record of their own, and the deepest chain
+    /// today is two such hops — but a cap here would turn a chain one level deeper than the cap into a staged
+    /// nothing, which <c>ParentOf</c> then reports as "no record contains this", a false statement rather than a
+    /// missing one. The chain is finite and Mutagen-owned (each context's Parent terminates at the mod), so
+    /// walking it to the end makes "no containing record" true by construction and lets a Mutagen bump that nests
+    /// one level deeper keep working with no edit here.</para></summary>
     internal static void Stage(IModContext context, List<(FormKey Child, FormKey Parent)> into)
     {
         if (context.Record is not IMajorRecordGetter child) return;
-        var up = context.Parent;
-        for (int hops = 0; up is not null && hops <= MaxGroupHops; hops++, up = up.Parent)
+        for (var up = context.Parent; up is not null; up = up.Parent)
             if (up.Record is IMajorRecordGetter ancestor) { into.Add((child.FormKey, ancestor.FormKey)); return; }
     }
 
--- a/src/housecarl-core/ContainmentIndex.cs
+++ b/src/housecarl-core/ContainmentIndex.cs
@@ -29,6 +29,12 @@ namespace HousecarlCore;
 /// fifths. Pure data — no bodies, no file handles.</para></summary>
 public sealed class ContainmentIndex
 {
+    /// <summary>The one spelling of the containment step, shared by every surface that takes a path — a
+    /// predicate, a field projection, a walk's seed paths and follow. The <c>*</c> sigil is §5.5's rule: the
+    /// co-resident space is field names, and <c>Worldspace.Parent</c> is a real field, so a bare <c>parent</c> is
+    /// takeable.</summary>
+    public const string ParentToken = "*parent";
+
     readonly Dictionary<ulong, ulong> _map = new();
     readonly Dictionary<ModKey, int> _modToIdx = new();
     readonly List<ModKey> _idxToMod = new();
@@ -76,6 +82,26 @@ public sealed class ContainmentIndex
         }
         return ((ulong)(uint)i << 32) | k.ID;
     }
+
+    /// <summary>The <c>*parent</c> hop a field read takes: this build's containment map, then the containing
+    /// record's winner body through the caller's own session, so a read that already holds one plugin open pays
+    /// nothing extra. A record nothing contains comes back with the plain sentence saying so and naming the
+    /// properties containment runs from — never a null the render has to guess at.</summary>
+    public static Func<IMajorRecordGetter, (IMajorRecordGetter? Parent, string? Why)> ReadHop(
+        LoadOrderResolver.IndexView view, LoadOrderResolver.OverlaySession session) => child =>
+    {
+        var pk = view.ParentOf(child.FormKey);
+        if (pk is null)
+            return (null, $"no record contains this {RecordNaming.StripOverlay(child.GetType().Name)} — containment runs " +
+                          $"from these properties only: {ChildBearingSurface()}");
+        var winner = view.ResolveWinner(pk.Value);
+        if (winner is null)
+            return (null, $"the containing record {pk.Value} is not in the active load order");
+        var body = view.GetRecord(session, winner.Value.WinnerPlugin, pk.Value);
+        return body is null
+            ? (null, $"the containing record {pk.Value} would not fetch from its winner '{winner.Value.WinnerPlugin}'")
+            : (body, null);
+    };
 
     /// <summary>The child-bearing property surface, spelled <c>Type.Property</c> — what a <c>*parent</c> refusal
     /// names as the reason a record has no containing record. Derived from
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -489,9 +489,6 @@ public sealed class FieldPredicateSet
         return (outSegs, folds, null);
     }
 
-    /// <summary>The containment step token: reverse containment, spelled with the <c>*</c> sigil because
-    /// <c>Worldspace.Parent</c> is a real field and a bare <c>parent</c> is takeable.</summary>
-    internal const string ParentToken = "*parent";
 
     /// <summary>Strip a side's leading <c>*parent</c> hops and hand back the rest of the path. A hop leads a path
     /// by definition — the containing record is a property of the RECORD, not of a field value — so a <c>*parent</c>
@@ -504,15 +501,15 @@ public sealed class FieldPredicateSet
         for (int i = hops; i < segs.Length; i++)
         {
             if (IsParentStep(segs[i]))
-                return (segs, 0, $"predicate '{raw}': '{ParentToken}' is the record that CONTAINS this one, so it can only lead a path — " +
-                                 $"in '{display}' it follows a field step. Write the hops first ('{ParentToken}.EditorID').");
+                return (segs, 0, $"predicate '{raw}': '{ContainmentIndex.ParentToken}' is the record that CONTAINS this one, so it can only lead a path — " +
+                                 $"in '{display}' it follows a field step. Write the hops first ('{ContainmentIndex.ParentToken}.EditorID').");
             var s = segs[i];
             int open = s.IndexOf('[');
             if (open > 0 && s.EndsWith("]", StringComparison.Ordinal)
-                && string.Equals(s[..open], ParentToken, StringComparison.OrdinalIgnoreCase))
-                return (segs, 0, $"predicate '{raw}': '{s}' — '{ParentToken}' names ONE containing record, not a list, so it takes no quantifier. Write '{ParentToken}'.");
+                && string.Equals(s[..open], ContainmentIndex.ParentToken, StringComparison.OrdinalIgnoreCase))
+                return (segs, 0, $"predicate '{raw}': '{s}' — '{ContainmentIndex.ParentToken}' names ONE containing record, not a list, so it takes no quantifier. Write '{ContainmentIndex.ParentToken}'.");
             if (s.Length > 0 && s[0] == '*' && open != 0)
-                return (segs, 0, $"predicate '{raw}': '{s}' is not a path token — the tokens are '{ParentToken}' (the containing record) and the quantifiers [*any], [*all], [*none] and [*count] on a list step.");
+                return (segs, 0, $"predicate '{raw}': '{s}' is not a path token — the tokens are '{ContainmentIndex.ParentToken}' (the containing record) and the quantifiers [*any], [*all], [*none] and [*count] on a list step.");
         }
         if (hops == 0) return (segs, 0, null);
         if (hops == segs.Length)
@@ -522,7 +519,7 @@ public sealed class FieldPredicateSet
         return (segs[hops..], hops, null);
     }
 
-    static bool IsParentStep(string seg) => string.Equals(seg, ParentToken, StringComparison.OrdinalIgnoreCase);
+    static bool IsParentStep(string seg) => string.Equals(seg, ContainmentIndex.ParentToken, StringComparison.OrdinalIgnoreCase);
 
     /// <summary>The token a fold is spelled with, for a message.</summary>
     static string FoldToken(Fold f) => f switch
@@ -1040,7 +1037,7 @@ public sealed class FieldPredicateSet
     {
         if (_parentOf is null || _fetchWinnerBody is null)
         {
-            _fatal = $"internal: a '{ParentToken}' containment predicate was evaluated without a bound resolution context — this scan surface does not support it.";
+            _fatal = $"internal: a '{ContainmentIndex.ParentToken}' containment predicate was evaluated without a bound resolution context — this scan surface does not support it.";
             return (null, EvalKind.Definite);
         }
         for (int i = 0; i < hops; i++)
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -1,4 +1,4 @@
-﻿using System.Globalization;
+using System.Globalization;
 using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Plugins.Records;
 
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -1,4 +1,4 @@
-using System.Globalization;
+﻿using System.Globalization;
 using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Plugins.Records;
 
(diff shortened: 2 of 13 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4395,7 +4417,8 @@ public CrossQueryOutcome OffOrderQuery(PoleInfo pole, IReadOnlyList<string>? typ
                             var w = view.ResolveWinner(fk);
                             return w is null ? null : view.GetRecord(sess!, w.Value.WinnerPlugin, fk);
                         }
-                        : null);
+                        : null,
+                    predicate.NeedsContainment ? fk => view.ParentOf(fk) : null);
```
Avick3110: **OffOrderQuery binds `*parent` to the ACTIVE order's containment index while scanning an off-order file, so the answer is wrong either way.**

`view.ParentOf(fk)` reads `IndexSnapshot.Containment`, which is built only from the plugins in the active load order. Two concrete failures on this lane:

1. A record *defined* in the off-order plugin (say a new `PlacedObject` in a patch houseCARL just wrote, or a disabled mod's REFR) has a ModKey that was never packed into `_modToIdx`, so `ParentOf` returns null and the predicate reports the named no-verdict "no record CONTAINS a PlacedObject — containment runs from these properties only: …". That sentence is simply false: the REFR *is* in a cell, in the very file being scanned. The real reason (this file's containment was never indexed) is never stated.

2. Worse, when the off-order file shares a filename with an active plugin — the routine MO2 case of inspecting an older/disabled copy of `Foo.esp`, and the case `where_source`/pole scans exist for — the ModKey *does* resolve, and `ParentOf` hands back the **active** order's parent for that FormID. The scan then filters the off-order body against a containment edge read out of a different file, with nothing in the output saying so.

Either refuse `*parent` on this lane by name (`predicate.NeedsContainment` is already computed right here), or stage containment for the pole file during its own enumeration. Silently answering from the wrong index is the case the second cornerstone rules out.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -125,26 +141,35 @@ enum PseudoPath { None, EditorId, Winner, FormId }
...
-    /// winner name) — the call site opens an overlay session for the fetch when true.</summary>
-    public bool NeedsBodyResolution => _predicates.Any(p => p.LinkPath is not null);
+    static int Hops(Predicate p) => p.ParentHops + p.LinkParentHops;
 
     /// <summary>Whether any predicate reads record BODY content (a leaf walk or a link step) — false when every
     /// term is header/resolution-only (`editorid`, `winner`, `formid` membership). The scan's deleted-record check
     /// keys on this: a deleted record has no live body for the CONTENT filters, but its EditorID and its winner
     /// resolution are real facts, so a header-only predicate set must still see it.</summary>
-    public bool NeedsLiveBody => _predicates.Any(p => p.LinkPath is not null || p.Pseudo == PseudoPath.None);
+    public bool NeedsLiveBody => _predicates.Any(p => p.LinkPath is not null || p.Pseudo == PseudoPath.None || Hops(p) > 0);
```
Avick3110: **`NeedsLiveBody` now excludes deleted records from a `*parent`-only predicate, which drops the PR's own motivating case.**

Adding `|| Hops(p) > 0` makes `where=["*parent.EditorID = SomeCell"]` a body-content predicate. All three scan lanes then run `if (DeletedRecordRule.HasNoLiveBody(body) && … predicate is { NeedsLiveBody: true }) continue;` (LoadOrderService.cs:4064, :4187, :4438) and skip the record with no accounting — the comment there calls it "a clean non-match".

But the containment hop never touches the child's body: it needs `body.FormKey` and nothing else, and the terms below the hop read the *parent's* body, which is live. A REFR a patch marked deleted (very common — patches delete placed references constantly) is exactly the crash-log lookup the changelog advertises, and it now silently disappears from the results.

The hop should be treated like `editorid`/`winner` — header-only for the child. `NeedsLiveBody` should key on whether the terms *below* the hop read body content, i.e. `p.LinkPath is not null || (p.Pseudo == PseudoPath.None && …)`, not on the hop count. Note `NeedsResolution`/`NeedsBodyResolution`/`NeedsContainment` are all correct as written; it is only this one that overreaches.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -310,12 +352,13 @@ public static (FieldPredicateSet? Set, string? Error) Parse(IReadOnlyList<string
...
         // plugin WINS the record, resolution not content), 'formid' (the membership ops' identity path).
-        var pseudo = segs.Length == 1 && !path.Contains('[')
-            ? (path.Equals("editorid", StringComparison.OrdinalIgnoreCase) ? PseudoPath.EditorId
-               : path.Equals("winner", StringComparison.OrdinalIgnoreCase) ? PseudoPath.Winner
-               : path.Equals("formid", StringComparison.OrdinalIgnoreCase) ? PseudoPath.FormId
-               : PseudoPath.None)
-            : PseudoPath.None;
+        // Classified off the segment left AFTER the '*parent' hops, not the raw path, so '*parent.editorid' reads
+        // the containing record's identity rather than falling through to a case-sensitive field walk.
+        var term = segs.Length == 1 && !segs[0].Contains('[') ? segs[0] : "";
```
Avick3110: **Moving the pseudo-path classification after `SplitFolds` makes it read fold-stripped segments, so a quantifier on an identity term is now silently dropped.**

`SplitFolds` rewrites `outSegs[i] = s[..open]`, so by the time this line runs, `editorid[*any]` has become `segs = ["editorid"]` with `pathFolds = [Any]`. The old test was `!path.Contains('[')` against the raw display path, which caught it and left `Pseudo.None`.

Consequences:
- `editorid[*any] = Foo` / `winner[*all] = Bar.esp` / `formid[*any] in [...]` now classify as the pseudo term, and `EvalCore`'s pseudo branches (lines 748, 762, 789) return before `EvalOwnPath` ever looks at `PathFolds`. The quantifier is accepted and ignored — a silently different filter than the one the caller wrote.
- `editorid[*count] > 2` now hits the `Pseudo.EditorId` op check and refuses with "'editorid' is a text term", which names the wrong problem.

The fix is to keep the fold out of the classification: gate on `pathFolds is null || pathFolds[0] == Fold.None` as well as `segs.Length == 1`. The `*parent` intent (classify off the segment left after the hops) is fine — it is the reordering past `SplitFolds` that costs this.

src/housecarl-core/ContainmentIndex.cs
```diff
@@ -0,0 +1,117 @@
...
+        return new FormKey(_idxToMod[(int)(packed >> 32)], (uint)packed);
+    }
+
+    /// <summary>Stage one plugin's containment edges, read off the context walk. Returns the pairs for the caller to
+    /// merge only if the whole plugin enumerated — the same plugin-atomic rule the winner index follows.</summary>
+    internal static void Stage(IModContext context, List<(FormKey Child, FormKey Parent)> into)
+    {
+        if (context.Record is not IMajorRecordGetter child) return;
+        var up = context.Parent;
+        for (int hops = 0; up is not null && hops <= MaxGroupHops; hops++, up = up.Parent)
```
Avick3110: **A context chain deeper than `MaxGroupHops` stages no edge and is then reported as "no record contains this", which is a wrong statement rather than a missing one.**

The loop bails after 7 iterations and `Stage` returns having added nothing. The doc above says that is the "chain holds no record" case and that `*parent` will "say it has none rather than guessing" — but the two are not the same thing. A chain that *does* hold a record ancestor beyond hop 6 produces the identical outcome as a genuinely top-level record: `ParentOf` returns null, and both the read hop and the predicate rollup then assert "no record contains this X — containment runs from these properties only: …", which is false.

The bound also has no derivation behind it (the comment says the deepest real chain is two hops, and the constant is 6), so a Mutagen version that nests one level deeper turns into a wrong answer with no signal — the opposite of the "a Mutagen bump grows the reach with no edit here" property this file claims. Either drop the cap (the chain is finite and Mutagen-owned) or record the truncation so the no-parent sentence can distinguish the two cases.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -850,9 +942,38 @@ EvalKind ClassifyMiss(string note)
...
+        if (_parentOf is null || _fetchWinnerBody is null)
+        {
+            _fatal = $"internal: a '{ContainmentIndex.ParentToken}' containment predicate was evaluated without a bound resolution context — this scan surface does not support it.";
+            return (null, EvalKind.Definite);
+        }
+        for (int i = 0; i < hops; i++)
+        {
+            var pk = _parentOf(body.FormKey);
+            if (pk is null) { _lastNoParent = RecordNaming.StripOverlay(body.GetType().Name); return (null, EvalKind.NoParent); }
+            if (!_targetCache.TryGetValue(pk.Value, out var parent))
```
Avick3110, line 1054: **The parent hop reuses `_targetCache`, whose documented bound does not hold for containment.**

The comment on `_targetCache` (line 134) justifies being unbounded on the grounds that the distinct link-target population is "hundreds-to-low-thousands" and that "a pathological whole-order high-fan-out path is bounded by the scope the grammar already requires (types= / plugins=)".

Neither holds here. `types=["PlacedObject"], where=["*parent.EditorID startswith Whiterun"]` over the order caches one entry per distinct **CELL**, and Skyrim's cell population is five figures before any mod is added; `types=` bounds the child type, not the parent population, so the stated escape hatch does not apply. A `*parent.*parent` chain adds every worldspace on top.

Not a wrong answer, but it is memory retained for the whole call on a path the grammar now makes easy to reach. Worth either bounding the cache or amending the comment so the next reader is not relying on a bound that containment breaks.

src/housecarl-core/ReadEngine.cs
```diff
@@ -213,11 +216,43 @@ public static RecordFields ReadFields(IMajorRecordGetter record, IReadOnlyList<s
...
+    /// <summary>The <c>*parent</c> containment step on a read path: strip the leading hops, climb to the record
+    /// that CONTAINS this one, and hand back what the rest of the path should be read on. The parent set is
+    /// Mutagen's own containment walk, captured at index build — never a per-record-type map here. Returns a note
+    /// instead when the hop cannot be taken: the record has no containing record, or this read surface carries no
+    /// load-order index to ask.</summary>
+    static (IMajorRecordGetter On, string[] Tail, string? Note) HopToParent(
+        IMajorRecordGetter record, string[] segs,
+        Func<IMajorRecordGetter, (IMajorRecordGetter? Parent, string? Why)>? parentOf)
+    {
+        int hops = 0;
```
Avick3110: **The read-side hop does not enforce the grammar refusals the changelog promises for `project.fields`.**

`FieldPredicate.SplitParentHops` refuses `*parent` after a field step, `*parent` carrying a quantifier, and any other `*` token, each by name. This function only strips a *leading* run and hands everything else to the ordinary path walk, so on the `fields=` surface:

- `fields=["Responses.*parent.EditorID"]` → a generic "no such field '*parent'" leaf note.
- `fields=["*parent[*any].EditorID"]` → same.
- `fields=["*owner.EditorID"]` → same.

The changelog entry says "A `*parent` after a field step, carrying a quantifier, or with nothing after it refuses by name" and lists `project={"form":"fields"}` as one of the surfaces the step lands on, so a caller who mistypes on the read surface gets the typo-hint path instead of the sentence the where= surface would give them for the identical mistake. The two surfaces should share one validator.

src/housecarl-core/ContainmentIndex.cs
```diff
@@ -0,0 +1,117 @@
...
+/// container grows <c>*parent</c>'s reach with no edit — the first cornerstone applied to an edge kind.</para>
+///
+/// <para><b>Later plugin wins, like everything else.</b> Across a real order thousands of children end up under a
+/// different parent than the first plugin put them under (a placed reference moved between cells), so the map is
+/// merged per plugin in priority order and the last declaration stands — the same rule the winner index follows.
+/// A plugin that throws part-way through its walk merges nothing, so a half-read plugin never leaves partial
+/// containment behind.</para>
+///
+/// <para><b>Packed.</b> Entries are ulong → ulong ((interned mod index &lt;&lt; 32) | FormID), not FormKey → FormKey:
+/// ~44 B an entry, about 77 MB for the 2.7M children of a large order, of which <c>Cell.Temporary</c> is four
```
Avick3110: **The two measured figures in this file's doc contradict themselves and the PR body.**

- Line 28: "~44 B an entry, about 77 MB for the 2.7M children". 2.7M × 44 B is ~119 MB, not 77 MB; 77 MB over 2,711,713 entries is ~28.5 B, which matches the PR body's "~30 B/entry". One of the two numbers on this line is wrong.
- Line 17 (and LoadOrderResolver.cs:585) says the context walk costs "about 0.4 s more over a 3,800-plugin order", while the PR body and the changelog both say the measured median is 0.31 s / "about 0.3 s".

These are load-bearing claims in a cornerstone comment — a later reader will size the map off them. Worth reconciling to the measured values.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -1,4 +1,4 @@
-using System.Globalization;
+﻿using System.Globalization;
```
Avick3110: **A UTF-8 BOM was added to the head of four files by this PR** — `FieldPredicate.cs`, `ReadEngine.cs`, `LoadOrderService.cs` and `RecordsTools.cs` each show a `using` → `\ufeffusing` change as their first hunk.

It looks like an editor artifact rather than an intended change: the four files are unrelated apart from having been opened, and no other file in the PR gained one. It puts a spurious first-hunk diff on the two largest files in the repo and will show up in every future blame on line 1. Worth stripping before merge.

Avick3110:
Blind review, second pass — coverage and conformance findings that sit alongside the eight inline comments, not repeats of them.

**`src/housecarl-core/LoadOrderResolver.cs:590` — nothing pins that the context walk yields the same record set as the flat one.** This swap is not scoped to `*parent`: the winner index, the overrider lists, `maxDepth` and the whole resolution surface are now built from `EnumerateMajorRecordContexts()` instead of `EnumerateMajorRecords()`. The comment asserts "it yields the same record set"; no test or probe asserts it. If Mutagen's context enumeration ever omits or duplicates a record kind (a Landscape, a navmesh, a DIAL's INFOs under some group shape), the failure is a mis-resolved order, silently, everywhere — not a missing `*parent`. A single equality assertion over both enumerations on the fixture (and a `ci-all` probe over a real order) would make that a caught regression rather than a discovered one.

**`src/housecarl-mcp-tests/RecordsContainmentTests.cs:73` — `TheChainReadsThroughTwoHops` does not take two hops.** It reads `*parent.EditorID` on `0xF10`, which the test above it establishes is a worldspace cell whose parent *is* the worldspace — one hop. So there is no end-to-end `*parent.*parent` against the real index anywhere; the chain is covered only by `WhereContainmentTests`, against a hand-built in-memory dictionary. The crash-log REFR -> CELL -> WRLD case the changelog and the design page both headline is therefore untested against the machinery that actually answers it.

**`src/housecarl-mcp-tests/RecordsContainmentTests.cs` — the child kinds the SPEC amendment names are only one-third covered.** The amendment's set is seven properties over three types. Tested: `DialogTopic.Responses`, `Cell.Temporary`, `Worldspace.SubCells`. Untested, and all present-or-addable in `OwnedChildWorld`: `Cell.Persistent` (`0xC1A` exists in the fixture and is never asserted), `Cell.Landscape` (`0xC1B`, likewise), `Cell.NavigationMeshes` — the one child kind that also names its parent on its own body, so the only place the index and the record body can disagree — and `Worldspace.TopCell`. ACHR is named in the issue title, the SPEC amendment and the PR body, and the fixture contains no `PlacedNpc` at all, so "CELL->REFR/ACHR" is half claimed.

**`src/housecarl-core/ContainmentIndex.cs:44` — the per-plugin later-wins merge, which the PR body calls the reason for the design, has no test.** `OwnedChildWorld` never re-parents anything: Top re-declares INFO `0xD10` under the same `HcOcTopic`, and Mid's cells add new children rather than moving existing ones. So the branch that 2,363 real children exercise is unexercised here, and the inverse case matters as much — Mid re-declaring `CellA` with an empty `Temporary` must *not* erase base's `0xC10 -> CellA` edge (it doesn't, because absent means no edge staged, but that is load-bearing behaviour standing on nothing).

**`src/housecarl-core/LoadOrderResolver.cs:873` — `ContainedRecordCount` is dead, and the cost it would declare is nowhere in band.** No caller anywhere in `src/`. Meanwhile `plugin/CHANGELOG.md` tells the user to "check yours against the build time `housecarl_load_order_status` reports" — but neither the containment map's entry count nor its share of the build time is reported by anything. SPEC §2.1's amendment makes in-band declaration of an index's cost the rule for #489; the same reasoning applies to a 77 MB map every user now pays for whether or not they ever spell `*parent`. Either surface the count on `load_order_status` (the accessor is already written for it) or drop the accessor and the changelog sentence.

**`src/housecarl-core/LoadOrderResolver.cs:593` — `if (ctx is IModContext c)` is a silent filter on a cast that cannot fail.** `EnumerateMajorRecordContexts()` yields `IModContext<...>`, which is an `IModContext`. If the guard ever did anything, it would drop that record's containment edge with no note — the failure mode the file's own doc comment says it avoids. Either drop the cast or make the else-branch say something.

**`src/housecarl-mcp/LoadOrderService.cs:4137` — a `plugins=`-scoped `where=` reads containment from the whole order, not the scoped plugin.** The `plugins=` lane exists to filter *that plugin's own body* — the comment two lines up says so ("the scoped plugin's own body ... rather than the winner"). `view.ParentOf` answers from the later-wins whole-order map, so `plugins=["A.esp"] where=["*parent.EditorID = X"]` can filter A.esp's records by an edge B.esp declared. That may well be the intended reading of containment, but it is a different pole from every other term in that lane and it is stated nowhere.

Nothing here blocks on its own; the first two are the ones I would not merge without.

Avick3110:
Review fold. All fifteen findings are real and all fifteen are folded; one is folded by the other arm the finding itself offered, which is called out below. Six commits on top, no rebase (`claude/e1-reverse` is building on this branch).

## Correctness

**1 (high) — `OffOrderQuery` binds the active order's `ParentOf`.** Real, both halves. `*parent` now refuses by name on the off-order lane, at the point `predicate.NeedsContainment` is computed, before the file is opened. Staging containment for the pole file during its own enumeration is the better answer eventually, but it is a different piece of work and refusing is the honest interim. Covered by `AParentFilterOverAnOutOfLoadOrderFileRefusesRatherThanAnsweringFromTheActiveIndex`, which uses the worst case you named — a same-named copy of an active plugin, so the ModKey *does* resolve.

**2 (high) — `NeedsLiveBody |= Hops > 0` drops deleted records.** Real, and your diagnosis of where it should key was right. It now asks whether the CANDIDATE's own body is read: `LinkPath is not null ? LinkParentHops == 0 : (ParentHops == 0 && Pseudo == None)`. A hop leading either side makes that side header-only on the child, which separates `*parent.X` and `*parent.X->y` from `Perks->*parent.X` — the last one *does* read the child's body, through the link's left path, and still reports true. The fixture gained a genuinely deleted `PlacedObject`; the test asserts both directions — a body predicate drops it, the `*parent` predicate keeps it — so it cannot pass vacuously.

**3 (medium) — pseudo classification reads fold-stripped segments.** Real. Gated on `pathFolds[0] == Fold.None` as you suggested. One step further than restoring the old behaviour: `editorid[*any]` would otherwise fall through to a case-sensitive walk for a lowercase field named `editorid` and report a typo, so a quantifier on any of the three identity terms now refuses by name. Four `[Theory]` rows in `WhereQuantifierTests`, plus one asserting the unquantified spellings still classify as the identity terms.

**4 (medium) — nothing pins that the two walks yield the same record set.** Real. Two arms:
- `TheContextWalkYieldsTheSameRecordSetAsTheFlatWalk` compares both enumerations per fixture plugin, through `LoadOrderResolver.OpenOverlay` — the same door the index build opens.
- `parent-in-hand` gained an untimed `E2` pass comparing them as **multisets** per plugin over a real order, so a drop, a duplicate and a substitution are each a difference and each names the file it is in. It exits non-zero on a difference. It is not in `ci-all` because it needs a live MO2 order and `ci-all` has none to run it over.

**5 (medium) — a chain deeper than `MaxGroupHops` reads as "nothing contains this".** Real, and folded by your **first** arm rather than your second: the cap is gone. Recording the truncation would mean a second per-child set, a second accessor, and widening `FieldPredicate`'s `_parentOf` from `Func<FormKey, FormKey?>` to a reason-carrying shape across three call sites — all to distinguish a state that then still has to be explained to the caller. Dropping the cap removes the state instead: the context chain is finite and Mutagen-owned (`Parent` terminates at the mod), so "no record ancestor anywhere in the chain" becomes true by construction and the sentence can never be wrong. It also restores the property the file claims — a Mutagen bump that nests one level deeper keeps working with no edit here — which a cap of 6 did not actually give. Consequence: there is no truncation test, because after this there is no truncation. Say the word if you want the accounting arm instead.

**6 (medium) — the read-side hop skips the grammar refusals.** Real. There is now one validator, `ContainmentIndex.SplitHops`, and all three surfaces call it: `where=`, `project.fields`, and the walk's seed paths (which passes `allowBare: true`, since a path of nothing but hops IS the edge a walk crosses). Each surface adds only its own voice. Pinned by a `[Theory]` asserting the SAME sentence fragment comes back from both surfaces for `Responses.*parent.EditorID`, `*parent[*any].EditorID`, `*owner.EditorID` and a bare `*parent`.

**7 (low) — `_targetCache`'s documented bound does not hold for containment.** Real. Folded as the comment arm: the containment case is now stated in full where the bound is claimed — `types=` bounds the child type, not the parent population; five figures of cells before any mod; a chain adds every worldspace on top. Not bounded, because the entries are bodies the scan fetches anyway and the lifetime is one call, so it is retention rather than repeated work — but the next reader is no longer relying on a bound containment breaks.

**8 (low) — `plugins=`-scoped `where=` reads containment from the whole-order map.** Real that it was stated nowhere. It is intended, and it is now argued at the bind site rather than left to be inferred: containment is a fact about the assembled order, not about one file — thousands of children across a real order sit under a different parent than the plugin that first declared them — so the whole-order map is the only well-defined reading, and a later plugin's re-parenting is what answers.

**9 (low) — `if (ctx is IModContext c)` cannot fail.** Real. Dropped; the compiler agrees.

## Tests

- **ACHR** — the fixture had no `PlacedNpc` at all. `CellG` now carries one, asserted through the index and through a `where=` filter, so "CELL to REFR/ACHR" is claimed in full.
- **`Cell.Persistent`, `Cell.Landscape`, `Cell.NavigationMeshes`, `Worldspace.TopCell`** — one row each. The navmesh test is the one you flagged: it reads the index AND the record's own `Data.Parent` (the fixture sets the `CellNavmeshParent` arm) and asserts they name the same cell, which is the only place the two could disagree. All seven properties of the amendment are now covered.
- **The two-hop test now takes two hops.** It reads a placed reference sitting inside a worldspace cell and asserts both that one hop lands on the cell and that two land on the worldspace, so a chain that quietly stops after one cannot pass. There is a matching `where=` assertion, so the chain is pinned on both surfaces against the real index rather than only against the in-memory dictionary in `WhereContainmentTests`.
- **The later-wins merge, both directions.** Mid declares base's reference under a cell of its own (the edge moves) and re-declares `CellG` carrying nothing (the edges base staged there survive). The second half is the behaviour you noted was standing on nothing.

## Hygiene

- The figures are reconciled to the measurements: ~30 B an entry over 2,711,713 children for the 77 MB, and the walk's cost stated as the measured spread (0.31–0.43 s across repeated alternating runs) with a note that the changelog quotes the low end — rather than two numbers contradicting each other and the PR body.
- `ContainedRecordCount` is no longer dead: `housecarl_load_order_status` reports the map's entry count on its own line, so the cost every build pays is declared in band per §2.1 rather than discovered as memory. The changelog sentence is rewritten to point at what actually gets reported; it had promised a build time the status line does not carry.
- The four UTF-8 BOMs are stripped, in their own first commit so they do not sit in front of anything else.

Build, `dotnet test --filter "tier!=bridge"` (1,235 passed, 21 of them new) and `ci-all` (128/128) are green.

Avick3110:
Blind review folded (15 of 15 real, 0 refuted; finding 5 taken as drop-the-cap, reversible). Queued for Aaron's review. Merge after #551; stacked follow-on: #562 (#489).

Avick3110:
Restacked onto main after #551 merged. Conflict resolution in FieldPredicate.cs, RecordsTools.cs, CHANGELOG and WhereQuantifierTests kept both sides; one extra commit (28daadfd) makes the new project.fields quantifier gate skip the *parent token so the containment grammar answers. Ready for Aaron's review.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2371,7 +2371,9 @@ ReadOutcome ResolveRead(LoadOrderResolver resolver, LoadOrderResolver.IndexView
                 $"Touched by (load order, winner last): {string.Join(", ", touchers)}.");
         }
 
-        var record = ReadEngine.ReadFields(rec, fields, depth, containerHint);   // materialise while the session (overlay) is open
+        // materialise while the session (overlay) is open; the *parent hop climbs the index's containment map and
+        // fetches the containing record's winner body through the same session
+        var record = ReadEngine.ReadFields(rec, fields, depth, containerHint, ContainmentIndex.ReadHop(view, session));
```
Avick3110: **A `*parent` read bypasses the owned-child declarers note.**

The next line, `AnnotateOwnedChildContent(record, rec, view, fk, out var childFields)`, keys on `OwnedChildContent.Fields(rec)` — the body of the record being read — and then matches `owning.ContainsKey(rf.Fields[i].Path)`. Neither test survives a hop:

- `formids=[<a REFR>] project={"form":"fields","fields":["*parent.Temporary"]}` — the rows come from the containing CELL, but `rec` is the PlacedObject, so `owning` is empty and the function returns `rf` unchanged at the first guard.
- Even reading a CELL, the emitted row path is `*parent.Temporary` (ReadEngine keeps the caller's spelling as the display), which is not a key in `owning`, so `hits` is null.

So the response lists the winner cell's `Temporary` contents with no "other plugins declare children here and this read did not open them" note and with `OwnedChildFields` null. That is the same answer `formids=[<the CELL>] fields=["Temporary"]` annotates — the two spellings of one question now disagree, and the unannotated one is the reading this PR headlines (crash-log REFR → its cell). AnnotateOwnedChildContent's own comment calls that a silently wrong answer.

Annotating would mean resolving the hop's target key (which `ReadHop` already has) and matching on the path with the leading hops stripped.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -80,12 +87,16 @@ enum Fold { None, Set, Any, All, NoneOf, Count }
...
+    /// <paramref name="ParentHops"/> / <paramref name="LinkParentHops"/> are how many leading <c>*parent</c>
+    /// containment steps that side opens with; the segments after them are stored bare, so once the hop lands on
+    /// the containing record everything downstream reads an ordinary path.</summary>
     sealed record Predicate(string Text, string[] PathSegments, string PathDisplay, Op Op, string Operand, double NumericOperand,
                             HashSet<FormKey>? FormIds = null, ArtifactDemand? Artifact = null,
                             string[]? LinkPath = null, string? LinkPathDisplay = null,
                             PseudoPath Pseudo = PseudoPath.None, IReadOnlyList<string>? RawMembers = null,
-                            Fold[]? PathFolds = null, Fold[]? LinkFolds = null);
+                            Fold[]? PathFolds = null, Fold[]? LinkFolds = null,
+                            int ParentHops = 0, int LinkParentHops = 0);
```
Avick3110, line 99: **`QuantifiedSteps` doesn't know about these two, so the schema refusal judges the wrong record type.**

The hops are stripped before the fold split, so a side with `ParentHops > 0` has `PathSegments` (or `LinkPath`) rooted at the *containing* record's type. But `QuantifiedSteps` (unchanged, ~line 241) still says:

```csharp
Collect(p.LinkPath, p.LinkFolds, p, true);
Collect(p.PathSegments, p.PathFolds, p, p.LinkPath is null);
```

`QuantifierShapeRefusal` then asks the *scanned* type's schema about a step that never lands on it:

`types=["Cell"] where=["*parent.Music[*any]->EditorID = X"]` reads `Worldspace.Music`, but the check runs `StepCardinality(Cell, ["Music"], 0)` and refuses with *"'Music[*any]' quantifies a step that is not a list — it is a link on Cell"* — naming a type the step is not on.

The wrong type doesn't only mis-word the message, it decides the verdict. Wherever a child and its container both carry a field of that name with different cardinality, a valid call is refused (list on the parent, scalar on the child) or a genuine refusal is silenced (the mirror case). Today's Mutagen shapes make the second the common one because an unknown name returns `null` and skips, but that is luck, not a guarantee — a Mutagen bump that adds a colliding name changes it.

`->`'s right side already has the answer: pass `onScanned: false` for a side whose hop count is non-zero, and those steps fall through to the runtime accounting.

src/housecarl-core/ReadEngine.cs
```diff
@@ -186,7 +186,8 @@ public static int RunRead(string[] args)
     /// so the server's reads inherit the read-proof by construction. Per-leaf fault isolation: an unreadable field
     /// names itself in its <see cref="FieldValue.Note"/>, never throws out of the record read.</summary>
     public static RecordFields ReadFields(IMajorRecordGetter record, IReadOnlyList<string>? paths = null, int depth = 1,
-                                          string? containerHint = DepthExpandHint)
+                                          string? containerHint = DepthExpandHint,
+                                          Func<IMajorRecordGetter, (IMajorRecordGetter? Parent, string? Why)>? parentOf = null)
```
Avick3110, line 198: **`parentOf` is wired on two `ReadFields` call sites; the other in-order ones refuse `*parent` with advice that doesn't apply.**

Threaded in at `LoadOrderService.cs:2376` (ReadRecordDetail) and `:3444` (OverlayPostBatch). Not at:

- `:2588` — the conflict-tree read (`project.form="tree"`, which takes `fields=`)
- `:3131`, `:3161`, `:3194`, `:3254`, `:3304`, `:3312` — every delta pole (`form="delta"`, which takes `fields=` to narrow)

All of those hold the same captured `view` and open `session` the wired lanes do — the off-order lane at `:2986` is the only one that legitimately has no index.

So `records formids=[<an INFO>] project={"form":"tree","fields":["*parent.EditorID"]}` gets *"'*parent' needs the load-order index, which this read does not carry — read the record through the load order instead"* on a read that is through the load order. The `fields=` description and the changelog line both advertise `*parent` on `project.fields` unqualified.

Either pass `ContainmentIndex.ReadHop(view, session)` on those lanes too, or name the forms it isn't available on — the current sentence sends a caller to a remedy they already used.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -344,12 +400,25 @@ public static (FieldPredicateSet? Set, string? Error) Parse(IReadOnlyList<string
...
-               : path.Equals("winner", StringComparison.OrdinalIgnoreCase) ? PseudoPath.Winner
-               : path.Equals("formid", StringComparison.OrdinalIgnoreCase) ? PseudoPath.FormId
-               : PseudoPath.None)
-            : PseudoPath.None;
+        // Classified off the segment left AFTER the '*parent' hops, not the raw path, so '*parent.editorid' reads
+        // the containing record's identity rather than falling through to a case-sensitive field walk. A step that
+        // carried a quantifier is NOT an identity term, and the fold must be read off pathFolds rather than off the
+        // segment: SplitFolds has already stripped the bracket, so 'editorid[*any]' reaches here spelled 'editorid'
+        // and would otherwise classify as the pseudo term with its quantifier silently dropped.
+        var term = segs.Length == 1 && !segs[0].Contains('[') && (pathFolds is null || pathFolds[0] == Fold.None)
```
Avick3110, line 409: Minor, and half of it predates this PR: because `term` is read off `segs[0]` after the hops are stripped, `*parent.EditorID` classifies as `PseudoPath.EditorId` and so reaches the presence-test refusal at line 439 — *"'*parent.EditorID' always exists (every record has an identity and a winner) — a presence test on it can never filter."*

EditorID is nullable, and exterior CELLs overwhelmingly have none, so `types=["PlacedObject"] where=["*parent.EditorID missing"]` — placed references in unnamed exterior cells — is a filter that would partition a real scan, refused on a premise that isn't true of the field. The `winner` half of that sentence is right; the identity half isn't.

The bare `editorid missing` spelling has the same problem, but `*parent` is what makes it worth noticing: reaching a container by identity is the case where "does it have an EditorID at all" is the discriminator.

Avick3110:
Second review fold (the four findings posted after the restack). Three folded, one refuted. Five commits on top, no rebase — `claude/e1-reverse` is still building on this branch.

## Folded

**1 — `*parent` bypasses the owned-child declarers note** (`LoadOrderService.cs:2376`). Real. `AnnotateOwnedChildContent` keyed `OwnedChildContent.Fields` on the child's body and matched `owning.ContainsKey(f.Path)`, so neither test survived a hop. It now groups the read's rows by how many hops their path opens with, climbs the read's own `ReadHop` once per group, and judges each group against the record its rows were actually read on — `owning` and `TouchingPlugins` both off that record, matched on the path below the hops, and keyed in the outcome map by the row's display path so the render still matches it. `formids=[<a REFR>] fields=["*parent.Temporary"]` and `formids=[<the CELL>] fields=["Temporary"]` now carry the same note. Test: `AReadThroughTheHopCarriesTheContainingCellsOwnedChildNote`, which asserts both spellings.

**2 — `QuantifiedSteps` judges the wrong record type** (`FieldPredicate.cs:99`). Real, and the fix is the one the finding named. A side that opens with a hop is rooted at the containing record's type exactly as the right side of a `->` is rooted at the target's, so `OnScannedType` is now `p.LinkParentHops == 0` for the link side and `p.LinkPath is null && p.ParentHops == 0` for the path side; those steps fall through to the runtime accounting. Test: `AQuantifiedStepBelowAHopIsNotRootedAtTheScannedType`, on the property directly, covering the plain side, a hopped path side and a hopped link side.

**3 — `parentOf` wired on two `ReadFields` call sites only** (`ReadEngine.cs:190`). Real. `ContainmentIndex.ReadHop(view, session)` is now passed on the conflict-tree read (`:2588`) and on all six delta poles — winner, previous_provider, named in-order, overlay pre and post. Every one of them already held the same captured view and open session. The off-order lane at `:2986` and the off-order pole keep the note, which is now accurate: it is the only read form that carries no index, and its sentence's advice ("read the record through the load order instead") is the right one there. Test: `TheComparisonFormsReadTheHopOnTheirInOrderArm`, a theory over `delta` and `tree`, measuring an in-order arm against an off-order file on one call so it pins both halves — the in-order side reads `HcOcCellH`, the off-order side says it cannot. Changelog line updated to say `fields=` takes the hop on every in-order form.

## Refuted

**4 — a presence test on `*parent.EditorID` is refused** (`FieldPredicate.cs:408`). Not real on this branch. The presence-op refusal at `:438` reads `if (pseudo is PseudoPath.Winner or PseudoPath.FormId)` — `EditorId` is deliberately not in it, and `EvalCore`'s EditorID branch handles `Exists`/`Missing` explicitly on `!string.IsNullOrEmpty(eid)`. So `*parent.EditorID missing` parses and filters, and so does the bare `editorid missing`. The quoted sentence exists but only fires for `winner` and `formid`, where "always exists" is true. Added `APresenceTestOnTheParentsEditorIdIsAFilter_NotARefusal` to pin it rather than leaving the claim untested.

Build, 1288 tests and 128/128 probes green.

PR #559, opened 2026-09-05
Give the S2 tools a format=json lane

### LINKED ISSUE #542: [Bug] the S2 tools (place, asset_status) have no format=json lane (SPEC §2.1 TRANSPORT)

Avick3110:
`housecarl_asset_status` and `housecarl_place` had no `format=json`. SPEC §2.1 makes TRANSPORT uniform across every
tool and calls a surface missing one of the axes a bug, so both tools get the lane through the same shared code every
other tool uses — `Wire.WantsJson` for the parameter, `Wire.Refuse`/`JsonWire.RenderError` for a whole-call refusal, a
`JsonWire` renderer beside the existing twins, `WriteSentences.Cap` for the write budget. No bespoke serializer and no
second budget rule.

**asset_status** (a read document, so no `ok` on a served answer): profile, the build-level caveats
(`read_incomplete`, `bsa_failures`, `warnings`, `selector_notes`) first, then one row per path — `winner` and the
full `providers` chain as `{name, kind}`, `ambiguous`, and for an absent path its `prefix_suggestions` plus the two
hedges the text render prints at the point of use, `absent_may_be_incomplete_read_failure` and
`absent_may_be_incomplete_undiscovered_archives`, kept apart because they are two facts with two remedies. Then the
same eight counters the text accounting line states, under one `accounting` object, so
`skipped + rendered + truncated + capped == total` on both lanes, with `next_limit`/`next_offset` measured off what
was rendered beside it.

The counters come from `AssetWire.Tally` — the composer the text line already goes through — and its `Widest` sizes a
reserve, so the accounting, the caveat blocks and the advice are all paid for INSIDE `max_chars` rather than written
past it. The three caveat blocks are capped and cut with the same named marker the text lane uses (an `under=` of
thousands of selectors is otherwise a block `max_chars` never reaches), and the first row always renders, so a small
`max_chars` still answers the path that was asked about.

A provider is `{name, kind}` rather than the printed `"Name" (loose)` token on purpose: the reusable value is the
name alone, which is what `place`'s `source_provider=` accepts, so a consumer reads a field instead of parsing the
display token apart.

**place** (a write document, so `ok` on both outcomes): the mod folder, warnings, one row per destination with its
bytes, source, `current_winner` and the off-order-source note, plus `set_provider_withheld` — and `next_step`, the
"a placed file does not win until you enable and sort the mod" instruction. That sentence and the leftover-folder
note now have one home each in `PlaceWire`, called by both lanes: a json caller who was told only that the write
succeeded would enable nothing, which is the degraded mode json is not allowed to be. Its row loop renders a first
row whatever `max_chars` says too — that row is the only place `current_winner` is stated.

`truncated` at the document root stays the boolean every json document carries; the text lane's `truncated=N` count
is `accounting.truncated`. A document that hits `max_chars` drops trailing rows and stays valid JSON rather than
cutting a string mid-object.

`PlaceOutcome` gains a `Success` property (`Error is null`) so the `ok` key is written from a write outcome, which
is what `refusal-completeness-guard` requires of every `ok` write in `JsonWire`.

**On #555.** That PR extracts these same counters into a shared `TransportAccounting` with a JSON twin. This PR's
json accounting is written in that shape deliberately — nested `accounting` object, the same eight names in the same
order, `truncated` as the count — so whichever merges second is a mechanical rebase: `AssetWire.WriteJson` becomes a
call to `TransportAccounting.WriteJson`, and `AssetWire.Tally`/`Widest` become `TransportAccounting.Tally`/`Widest`.
After #555 merges, this lane rewires through `TransportAccounting` rather than keeping its own composer.
`next_limit`/`next_offset` and the two advice sentences stay siblings of the object rather than members of it, which
is where a rewire leaves them untouched.

Tests: `S2JsonLaneTests.cs` — nine arms over the real `asset_status` json lane (winner and chain as data, the eight
counters summing to the total, a cut document that still parses, the first row rendering at a tiny `max_chars`, the
caveat blocks capped, the tail landing inside the cap, a refusal document, an unrecognized `format`, and `place`'s
refusal document through the same door), one driving `place` end to end to a SERVED write with a withheld set-level
pole, one at the render seam for the two ABSENT hedges, and five at `place`'s render seam, including the first-row
rule and one holding the `next_step` sentence identical across the two lanes. Suite 1190 passed, `ci-all` 128/128.

Stacked on `claude/strings-remainder` (#549) — review that one first; this PR's base moves to `main` when it merges.

Closes #542

### DIFF
--- a/src/housecarl-mcp/JsonWire.cs
+++ b/src/housecarl-mcp/JsonWire.cs
@@ -228,28 +228,30 @@ static class JsonWire
     }
 
     /// <summary>The json twin of <see cref="BatchRender.AppendLines"/>: a caveat list bounded by the SAME budget the
-    /// row loop is bounded by, cut with the same named marker as a final element. An unbounded list here is a list
-    /// max_chars does not reach — one under= of thousands of selectors writes megabytes before the row loop takes
-    /// its first budget reading, and the rows the caller asked about are then all cut.
-    /// <para><paramref name="cap"/> is what the marker names — the number the caller passed and would raise —
-    /// while <paramref name="budget"/> is what it is measured against.</para></summary>
-    static void WriteCappedStringArray(Utf8JsonWriter w, MemoryStream ms, string name, IReadOnlyList<string> items,
-                                       string itemNoun, int cap, int budget)
+    /// row loop is bounded by. An unbounded list here is a list max_chars does not reach — one under= of thousands of
+    /// selectors writes megabytes before the row loop takes its first budget reading, and the rows the caller asked
+    /// about are then all cut.
+    /// <para>The cut is a sibling <c>&lt;name&gt;_omitted</c> count, not a prose element inside the array — the
+    /// capped-list-plus-count shape <see cref="Wire.ContestedHostsShown"/> already uses. A marker element would be
+    /// handed to a consumer iterating the array as if it were one of the entries, and the array length would stop
+    /// matching the count the accounting states.</para>
+    /// <para>Returns how many entries were omitted, so the caller can say so at the document root.</para></summary>
+    static int WriteCappedStringArray(Utf8JsonWriter w, MemoryStream ms, string name, IReadOnlyList<string> items,
+                                      int budget)
     {
         w.WriteStartArray(name);
         int shown = 0;
         foreach (var s in items)
         {
             // shown > 0: the first line always renders, exactly as it does on the text lane.
-            if (shown > 0 && Over(w, ms, budget))
-            {
-                w.WriteStringValue($"… [{items.Count - shown} more {itemNoun} omitted at max_chars={cap}; raise max_chars to see all]");
-                break;
-            }
+            if (shown > 0 && Over(w, ms, budget)) break;
             w.WriteStringValue(s);
             shown++;
         }
         w.WriteEndArray();
+        int omitted = items.Count - shown;
+        w.WriteNumber(name + "_omitted", omitted);
+        return omitted;
     }
 
     // ---- shared record + field writers --------------------------------------------------------------
@@ -2579,10 +2581,10 @@ static class JsonWire
             // against the same budget its text twin caps against — an under= of thousands of selectors would
             // otherwise write megabytes here before the row loop ever checked the budget.
             w.WriteBoolean("read_incomplete", d.ReadIncomplete);
-            WriteCappedStringArray(w, ms, "bsa_failures", d.BsaFailures, "archive(s)", cap, budget);
-            WriteCappedStringArray(w, ms, "warnings", d.Warnings, "warning(s)", cap, budget);
-            if (d.SelectorNotes is null) w.WriteNull("selector_notes");
-            else WriteCappedStringArray(w, ms, "selector_notes", d.SelectorNotes, "selector(s)", cap, budget);
+            int caveatsOmitted = WriteCappedStringArray(w, ms, "bsa_failures", d.BsaFailures, budget)
+                               + WriteCappedStringArray(w, ms, "warnings", d.Warnings, budget);
+            if (d.SelectorNotes is null) { w.WriteNull("selector_notes"); w.WriteNumber("selector_notes_omitted", 0); }
+            else caveatsOmitted += WriteCappedStringArray(w, ms, "selector_notes", d.SelectorNotes, budget);
 
             w.WriteStartArray("results");
             int rendered = 0;
@@ -2598,8 +2600,10 @@ static class JsonWire
 
             var counts = AssetWire.Tally(d, rendered);
             TransportAccounting.WriteJson(w, counts);
-            w.WriteBoolean("truncated", counts.Truncated > 0);
-            WriteAssetAdvice(w, counts, cap);
+            // The document's own flag, so a consumer branching on it re-calls when ANYTHING was dropped: a caveat
+            // block the budget cut is a loss the row counters cannot see.
+            w.WriteBoolean("truncated", counts.Truncated > 0 || caveatsOmitted > 0);
+            WriteAssetAdvice(w, counts, cap, caveatsOmitted > 0);
             w.WriteEndObject();
         }
         return Finish(ms);
@@ -2609,8 +2613,10 @@ static class JsonWire
     /// page (measured off what was RENDERED, so a consumer paging by these lands on the first path it has not seen),
     /// an offset past the end, and the max_chars cut. Siblings of the <c>accounting</c> object rather than members of
     /// it, because that object is the eight counters and nothing else.</summary>
+    /// <param name="caveatsCut">a caveat block lost entries to the budget — the cut the row counters cannot see.</param>
     /// <param name="everySentence">write them all, whatever the counts say — the widest case the reserve measures.</param>
-    static void WriteAssetAdvice(Utf8JsonWriter w, TransportCounts c, int cap, bool everySentence = false)
+    static void WriteAssetAdvice(Utf8JsonWriter w, TransportCounts c, int cap, bool caveatsCut,
+                                 bool everySentence = false)
     {
         if (everySentence || c.Remaining > 0)
         {
@@ -2619,15 +2625,17 @@ static class JsonWire
         }
         if (everySentence || (c.Remaining == 0 && c.Total > 0 && c.Offset >= c.Total))
             w.WriteString("offset_note", $"offset={c.Offset} is past the end of the selection ({c.Total} path(s)) — the last page starts before it.");
-        if (everySentence || c.Truncated > 0)
+        if (everySentence || c.Truncated > 0 || caveatsCut)
             w.WriteString("truncated_note",
-                $"max_chars={cap} cut trailing resolved path(s) from this document — raise max_chars, or page with limit=/offset=.");
+                $"max_chars={cap} cut content from this document — accounting.truncated names the resolved path(s) " +
+                "dropped and each *_omitted counter the caveat entries; raise max_chars, or page with limit=/offset=.");
     }
 
-    /// <summary>The chars held back from max_chars for the tail this document closes on — the accounting object, the
-    /// truncation flag and every advice sentence. Measured by serializing the WIDEST tail under the response's own
-    /// writer options, so no rendering of it can outgrow its own room (measuring unindented what is written indented
-    /// under-reserves by the whole indentation).</summary>
+    /// <summary>The chars held back from max_chars for what this document writes outside the budgeted body — the
+    /// accounting object, the truncation flag, every advice sentence, and the three caveat <c>_omitted</c> counters,
+    /// which are written after their array has already spent the budget. Measured by serializing the WIDEST tail
+    /// under the response's own writer options, so no rendering of it can outgrow its own room (measuring unindented
+    /// what is written indented under-reserves by the whole indentation).</summary>
     static int AssetTailReserve(AssetStatusData d, int cap)
     {
         var widest = AssetWire.Widest(d);
@@ -2636,9 +2644,13 @@ static class JsonWire
         {
             w.WriteStartObject();
             w.WriteString("before", "");   // the tail is never a document's first member, so it pays the separator one owes
+            // Each block can omit at most its own entries, so its own count is the widest number it can write.
+            w.WriteNumber("bsa_failures_omitted", d.BsaFailures.Count);
+            w.WriteNumber("warnings_omitted", d.Warnings.Count);
+            w.WriteNumber("selector_notes_omitted", d.SelectorNotes?.Count ?? 0);
             TransportAccounting.WriteJson(w, widest);
             w.WriteBoolean("truncated", true);
-            WriteAssetAdvice(w, widest, cap, everySentence: true);
+            WriteAssetAdvice(w, widest, cap, caveatsCut: true, everySentence: true);
             w.WriteEndObject();
         }
         return (int)ms.Length;
--- a/src/housecarl-mcp/JsonWire.cs
+++ b/src/housecarl-mcp/JsonWire.cs
@@ -227,6 +227,31 @@ static class JsonWire
         w.WriteEndArray();
     }
 
+    /// <summary>The json twin of <see cref="BatchRender.AppendLines"/>: a caveat list bounded by the SAME budget the
+    /// row loop is bounded by, cut with the same named marker as a final element. An unbounded list here is a list
+    /// max_chars does not reach — one under= of thousands of selectors writes megabytes before the row loop takes
+    /// its first budget reading, and the rows the caller asked about are then all cut.
+    /// <para><paramref name="cap"/> is what the marker names — the number the caller passed and would raise —
+    /// while <paramref name="budget"/> is what it is measured against.</para></summary>
+    static void WriteCappedStringArray(Utf8JsonWriter w, MemoryStream ms, string name, IReadOnlyList<string> items,
+                                       string itemNoun, int cap, int budget)
+    {
+        w.WriteStartArray(name);
+        int shown = 0;
+        foreach (var s in items)
+        {
+            // shown > 0: the first line always renders, exactly as it does on the text lane.
+            if (shown > 0 && Over(w, ms, budget))
+            {
+                w.WriteStringValue($"… [{items.Count - shown} more {itemNoun} omitted at max_chars={cap}; raise max_chars to see all]");
+                break;
+            }
+            w.WriteStringValue(s);
+            shown++;
+        }
+        w.WriteEndArray();
+    }
+
     // ---- shared record + field writers --------------------------------------------------------------
     /// <summary>Serialize the fields array. Each leaf is <c>{path, value}</c> for a round-trippable leaf (value = the
     /// SAME wire token the text mode emits) or <c>{path, note}</c> for a no-value leaf; the display-only <c>display</c>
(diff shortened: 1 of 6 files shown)


src/housecarl-mcp/JsonWire.cs
```diff
@@ -2515,6 +2515,191 @@ public static string RenderForwardOutcome(WritePatchBuilder.ForwardOutcome o, in
...
+        using (var w = new Utf8JsonWriter(ms, Opts))
+        {
+            w.WriteStartObject();
+            w.WriteString("profile", d.ProfileName.Length > 0 ? d.ProfileName : "(unconfigured)");
+            // The caveats lead, as they do in the text render: an ABSENT below is only authoritative when both are
+            // empty, and a document that truncates its rows must not be able to cut the reason away.
+            w.WriteBoolean("read_incomplete", d.ReadIncomplete);
+            WriteStringArray(w, "bsa_failures", d.BsaFailures);
+            WriteStringArray(w, "warnings", d.Warnings);
+            WriteNullableStringArray(w, "selector_notes", d.SelectorNotes);
```
Avick3110: **medium — `max_chars` is not honoured for the build-level caveats on the json lane.**

`bsa_failures`, `warnings` and `selector_notes` go out through `WriteStringArray`, which is unbounded, and they are written *before* the row loop's first `Over(w, ms, cap)` check. The text twin deliberately caps each of these three blocks against the same `cap` (`AppendReadFailures` / `AppendDiscoveryWarnings` / `AppendSelectorNotes` all forward `cap` into `BatchRender.AppendLines`), with the comment on `AppendSelectorNotes` naming the exact scenario: *"that input can be thousands of selectors, which would write megabytes before the per-path loop ever checks the budget."*

Concrete case: `asset_status(under=[…3000 selectors that match nothing…], format='json', max_chars=2000)`. The text lane cuts the note list at 2000 chars with a named marker; the json lane emits all 3000 notes — a multi-hundred-KB document past the host's response limit — and then, because the stream is already far over `cap`, breaks on the very first result row and reports `truncated:true` with `results:[]`. Same shape for a build with many unreadable archives.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2515,6 +2515,191 @@ public static string RenderForwardOutcome(WritePatchBuilder.ForwardOutcome o, in
...
+
+            w.WriteStartArray("results");
+            int rendered = 0;
+            bool truncated = false;
+            foreach (var r in d.Results)
+            {
+                if (Over(w, ms, cap)) { truncated = true; break; }
```
Avick3110: **medium — the json row loop drops the "first row always renders" guarantee the text lane makes.**

`BatchRender.Render` guards its cut with `shown > 0`, and says why in its own summary: *"a cap-checked per-item loop whose FIRST item always renders its core answer … max_chars bounds the batch tail, never a single-item call's own answer."* The json loop checks `Over` unconditionally at the top, so nothing guarantees a row.

Concrete case: `asset_status(asset_paths=['textures/x.dds'], format='json', max_chars=400)` on a build carrying a couple of BSA read failures. The caveat block alone clears 400 chars, so the loop breaks before the single row and the caller gets `results: []`, `truncated: true` — the one question they asked is unanswered, while the same call in `text` answers it. The unbounded caveats above make this reachable at much larger `max_chars` too.

`RenderPlaceOutcome`'s loop (L2653) has the same shape, and there the dropped row is the only place `current_winner` — the mod the caller must sort above — is stated.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -59,40 +59,55 @@ public static string Place(
...
-            return "error: assets is empty. Pass one or more { path|formid, kind?, source?, source_provider? } destinations.";
-        return PlaceTargets(svc, assets, source_provider, kind, patch, into, max_chars);
+            return Refuse(json, "assets is empty. Pass one or more { path|formid, kind?, source?, source_provider? } destinations.");
+        return PlaceTargets(svc, assets, source_provider, kind, patch, into, max_chars, json);
     });
 
+    /// <summary>The write surface's one refusal shape, in the transport the caller asked for — the text lane's
+    /// "error: " prefix, or the json refusal document.</summary>
+    static string Refuse(bool json, string message)
+        => json ? JsonWire.RenderError(message, null) : "error: " + message;
```
Avick3110: **low — this duplicates `Wire.Refuse`.**

`ReadTools.cs:53` already is "the write/read surface's one refusal shape, in the transport the caller asked for": `Wire.Refuse(bool json, string message, string? epoch = null)`, which returns text unchanged and strips `RefusalPrefix` before handing the sentence to `JsonWire.RenderError`. The refusal-completeness guard even lists both `Wire.Refuse` and a bare `Refuse` as approved renderers precisely so a local helper isn't a hole — but a second definition is a second place the shape can drift, and this one inverts the convention (callers pass the message *without* the `error: ` prefix here, *with* it everywhere else, so `RefusalPrefix` no longer has one owner).

Suggest calling the shared one and dropping this helper:

```suggestion
```

…with the call sites reading `Wire.Refuse(json, "error: " + message)` — or, if a local shorthand is wanted, `static string Refuse(bool json, string message) => Wire.Refuse(json, Wire.RefusalPrefix + message);` (`AssetTools`' local function has the same issue).

Avick3110:
## Blind review — findings beyond the three inline comments

Read only this PR's own diff (`gh pr diff 559`), plus CLAUDE.md, #542 and SPEC §2.1.

### 1. `src/housecarl-mcp/JsonWire.cs:2555-2575` — medium. The json lane re-implements `AssetWire.Tally`, and drops its reserve.

`AssetWire` deliberately puts the eight counters behind ONE composer, and says so at the type:

> "A record so the real line and the widest-case line the reserve measures go through ONE composer — a second formatter would be a second spelling, and the reserve would stop bounding what is written."

`RenderAssetStatus` writes a second spelling inline: `Math.Min(d.Offset, d.Selected)`, `Math.Max(d.Selected - d.Offset - d.Results.Count, 0)`, `Math.Max(d.Results.Count - rendered, 0)`, `Math.Max(d.Selected - (d.Offset + rendered), 0)` and the `d.Limit > 0 ? d.Limit : AssetWire.DefaultPageLimit` next-page rule are all copies of `AssetWire.Tally`/`Compose`, and the two conditional sentences (`offset_note`, the next-page pair) copy `Compose`'s two conditions. The arithmetic is correct today; the objection is that it is now maintained twice on the same tool, which is the thing the text lane's own comment forbids.

The consequence that is not just style: `AssetWire.Render` passes `reserve: AccountingReserve(d)` into `BatchRender.Render` precisely so the accounting is paid for INSIDE `max_chars` —

> "max_chars then means the same on this tool as on every other."

The json lane has no reserve. The header block, the eight counters, `offset_note` and `truncated_note` are all written past the `Over(...)` check, so a `format='json'` answer overruns the caller's `max_chars` by the whole accounting block — the test at `S2JsonLaneTests.cs:63` passes `max_chars=300` and the document it asserts on is several times that. Either reserve, as the text twin does, or state in the doc-comment that the json lane's `max_chars` bounds the rows only.

### 2. `src/housecarl-mcp/JsonWire.cs:2555` — medium. Collides with #555, which moves exactly this into `TransportAccounting`.

Parallel PR #555 (`claude/skse-transport`, also `main`-based, also closing a §2.1 gap) extracts `asset_status`'s counts, composer, reserve and paging window into a shared `TransportAccounting`, and gives it a JSON twin: `TransportAccounting.WriteJson` emits the eight numbers **nested under an `accounting` object**, with `truncated` as the **count**. This PR emits them **flat at the document root**, with `truncated` as a **boolean** and the count renamed `truncated_rows`. `DefaultPageLimit` also moves (`AssetWire.DefaultPageLimit` → `TransportAccounting.DefaultPageLimit`), which this file now references.

So whichever merges second, `housecarl_skse` and `housecarl_asset_status` ship two different json accounting shapes for the same eight §2.1 counters, and the second merge has a textual conflict in `AssetTools.cs` on top. This needs one shape agreed before either lands — and once `TransportAccounting.WriteJson` exists, `RenderAssetStatus` should call it rather than keep its own copy (which also fixes finding 1).

### 3. `src/housecarl-mcp-tests/S2JsonLaneTests.cs:110-169` — medium. `place`'s json lane is never driven end to end on a successful write.

Six arms go through the real `asset_status` tool; of the four `place` arms, three call `JsonWire.RenderPlaceOutcome` directly at the render seam and the fourth (`S2JsonLaneTests.cs:96`) is a refusal. Nothing drives `PlaceTools.Place(..., format: "json")` to a served outcome, and no probe does either (`git grep 'format: "json"' src/housecarl-generator` hits only `ApplyGuardProbe`). Untested as a result: that `PlaceTargets` routes `json` to the json renderer at all (`PlaceTools.cs:136`), and that `poleWithheld` reaches it — `set_provider_withheld` is never asserted `true` on either lane in this file, so the one field the write lane added for the json document has no coverage of its interesting value. One arm placing a real file with a set-level `source_provider` that a member cannot use would close both.

### 4. `src/housecarl-mcp/JsonWire.cs:2606` — low. The two ABSENT hedges collapse into one boolean.

`JsonWire`'s own class contract is that "text and JSON can differ only in formatting, never in data". The text render prints two distinct hedges on an ABSENT row — an archive that failed to read, vs. base archives never discovered — and `absent_may_be_incomplete` is `readIncomplete || discoveryIncomplete`, so a json consumer cannot tell which applies without re-deriving it from the top-level `read_incomplete` / `warnings`. Two booleans, or reuse the top-level pair by name, would keep the lanes at parity.

### Checked and clean

- `PlaceOutcome.Success => Error is null` is right: `Error` is non-null only on `PlaceOutcome.Fail`, which carries no results, and it matches the `Success` spelling on the six other outcome records. It satisfies the refusal guard's §4 arm (`ok` is literal `false` or `o.Success`).
- The provider `{name, kind}` claim holds. `AssetSourceSelection.Matches` compares the selector against `ProviderName` **or** `OwningMod`, and `AssetProvider.Source` is the mod folder / "overwrite" / "Data" / the BSA filename — every one of which `source_provider=` accepts. The name alone really is the token.
- `PlaceWire.EnableAndSort` / `LeftoverNote` are a genuine extraction, not a copy; `TheTwoLanesCarryTheSameNextStepSentence` holds them identical, and `next_step` is gated on `placed > 0` on both lanes.
- No caller of `PlaceTools.Place` or `AssetTools.AssetStatus` passes `max_chars` positionally, so inserting `format` ahead of it is safe.

Avick3110:
## Review fold — 7 findings, 7 folded, 0 refuted

**Folded**

1. **Unbounded caveat blocks (inline, `JsonWire.cs:2540`).** Real: `bsa_failures`, `warnings` and `selector_notes`
   went out through the unbounded `WriteStringArray` before the first budget reading, so an `under=` of thousands of
   selectors wrote the lot and then cut every row. New `WriteCappedStringArray` — the json twin of
   `BatchRender.AppendLines`: the same budget, the same "first line always renders" rule, and the cut named as a
   final element (`… [N more selector(s) omitted at max_chars=X; raise max_chars to see all]`). Test:
   `TheCaveatBlocksAreCappedByMaxCharsToo` — 400 selectors at `max_chars=2000`, the list is cut, the document is a
   quarter of the unbounded one, and `accounting.notes` still counts all 400.

2. **The dropped "first row always renders" rule (inline, `JsonWire.cs:2547`).** Real on both renders. Both loops now
   check `rendered > 0 && Over(...)`. Tests: `TheFirstRowRendersHoweverSmallMaxCharsIs` (a one-path call at
   `max_chars=120` is still answered) and `TheFirstDestinationRowRendersHoweverSmallMaxCharsIs` (at `max_chars=1`, and
   it asserts `current_winner` is on that row — the thing the cut was hiding).

3. **The duplicated `Refuse` (inline, `PlaceTools.cs:101`).** Real. Both helpers now delegate:
   `Refuse(...) => Wire.Refuse(json, Wire.RefusalPrefix + message)`, the shorthand form you offered, so
   `RefusalPrefix` has one owner again and the call sites are unchanged.

4. **The inline copy of `Tally`/`Compose`, and the missing reserve (summary 1).** Real. The counters now come from
   `AssetWire.Tally` itself, and `AssetWire.Widest` sizes a reserve measured by serializing the widest tail under the
   response's own writer options (the `CheckAccounting.MeasureJson` pattern), held back before the caveats and rows
   write. `AssetWire` gained one method — `WriteJson`, the json twin of `Compose`, placed beside it so the reserve
   measures both through the same `Counts`. One qualification on the remedy: the reserve makes the TAIL fit, but the
   document can still end up to one row past `max_chars` on the row that was in flight when the budget ran out —
   which is exactly the overshoot the text lane makes, so the two lanes now mean the same thing by `max_chars`. The
   test asserts that rather than an exact bound: `TheDocumentFitsTheMaxCharsItWasGiven`.

5. **The #555 collision (summary 2).** Cannot take that branch into this one, so instead this PR's json accounting is
   written in the shape #555 ships: nested `accounting` object, the same eight names in the same order, `truncated` as
   the count (`truncated_rows` is gone; the root `truncated` boolean stays, as every json document carries one).
   Whichever merges second is then a mechanical rebase — `AssetWire.WriteJson`/`Tally`/`Widest` become the
   `TransportAccounting` calls, and `DefaultPageLimit` moves with them. Stated in the PR body: after #555 merges, this
   lane rewires through `TransportAccounting` rather than keeping its own composer. `next_limit`/`next_offset` and the
   two advice sentences are siblings of the object, not members, which is where a rewire leaves them untouched — the
   text lane states that advice, so json cannot drop it.

6. **`place`'s json never driven to a served write (summary 3).** Real. New `PlaceJsonServedLaneTests` — its own
   fixture instance, because the arm writes a mod folder — drives `PlaceTools.Place(..., format: "json")` to a real
   placement whose on-disk `source=` the set-level `source_provider=` cannot apply to: it asserts the document parses
   (so `PlaceTargets` routed json at all), `ok`, `placed=1`, `set_provider_withheld=true`, a null `current_winner`,
   and `next_step`.

7. **The collapsed ABSENT hedge (summary 4).** Real. Two booleans now:
   `absent_may_be_incomplete_read_failure` and `absent_may_be_incomplete_undiscovered_archives`, one per text hedge.
   Test: `AssetStatusJsonHedgeTests.TheTwoAbsentHedgesAreStatedApart`, posing each caveat alone at the render seam.

**Refuted:** nothing — all seven hold.

Build clean, suite 1190 passed (1184 + 6 new), `ci-all` 128/128.

Avick3110:
Blind review folded (7 of 7 real, 0 refuted; json accounting shape aligned to #555's). Queued for Aaron's review. Merge after #549 and rebase onto #555 to rewire through TransportAccounting.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2515,6 +2540,227 @@ public static string RenderForwardOutcome(WritePatchBuilder.ForwardOutcome o, in
...
+                // the budget — the same rule BatchRender makes on the text lane, so a one-path call is answered.
+                if (rendered > 0 && Over(w, ms, budget)) break;
+                WriteAssetRow(w, r, d.ReadIncomplete, d.Warnings.Count > 0);
+                rendered++;
+            }
+            w.WriteEndArray();
+
+            var counts = AssetWire.Tally(d, rendered);
+            AssetWire.WriteJson(w, counts);
+            w.WriteBoolean("truncated", counts.Truncated > 0);
```
Avick3110: **`truncated:false` on a document that was cut.**

`counts.Truncated` is `max(d.Results.Count - rendered, 0)` — rows only. The three `WriteCappedStringArray` calls above drop entries from `bsa_failures` / `warnings` / `selector_notes` without touching it, so a document whose caveats were cut but whose rows all fit reports itself complete, and `WriteAssetAdvice` writes no `truncated_note` either.

The PR's own test hits it: `TheCaveatBlocksAreCappedByMaxCharsToo` passes 400 non-matching selectors at `max_chars=2000`, which resolves zero paths — `rendered=0`, `Results.Count=0`, `accounting.truncated=0`, root `truncated=false` — while 389 selector notes were dropped. A consumer branching on `truncated` to decide whether to re-call with a larger `max_chars` concludes nothing was lost. The only remaining signal is the prose marker inside the array, which is the parsing this lane exists to remove.

A sibling omitted-count per caveat block (see the comment on `WriteCappedStringArray`) would fix both this and the marker.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -227,6 +227,31 @@ static void WriteStringArray(Utf8JsonWriter w, string name, IReadOnlyList<string
...
+                                       string itemNoun, int cap, int budget)
+    {
+        w.WriteStartArray(name);
+        int shown = 0;
+        foreach (var s in items)
+        {
+            // shown > 0: the first line always renders, exactly as it does on the text lane.
+            if (shown > 0 && Over(w, ms, budget))
+            {
+                w.WriteStringValue($"… [{items.Count - shown} more {itemNoun} omitted at max_chars={cap}; raise max_chars to see all]");
```
Avick3110: **The cut marker is written as a data element of the array.**

A consumer iterating `selector_notes` to map each note back to the selector it came from receives `"… [389 more selector(s) omitted at max_chars=2000; raise max_chars to see all]"` as note #12, and either mis-attributes it or has to substring-match prose to reject it. `accounting.notes` (400) then no longer equals the array length, and nothing else in the document explains the gap.

This file already carries the right pattern for the same problem: `Wire.ContestedHostsShown`'s doc says "Both lanes publish the full distinct count beside the capped list" — a capped list plus a sibling count, not an in-band sentence. `bsa_failures_omitted` / `warnings_omitted` / `selector_notes_omitted` (or a single `caveats_truncated: true`) would keep the arrays pure data and give the document root something honest to set `truncated` from.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2515,6 +2540,227 @@ public static string RenderForwardOutcome(WritePatchBuilder.ForwardOutcome o, in
...
+            w.WriteBoolean("truncated", truncated);
+            // Not "raise max_chars to see the rest": the bytes are already on disk, so a re-issue would place again.
+            if (truncated)
+                w.WriteString("truncated_note",
+                    $"the render hit max_chars={cap} and dropped trailing destination rows — the WRITE is unaffected; " +
+                    "raise max_chars and re-read with " + ToolNames.AssetStatus + " rather than placing again.");
+            if (o.LeftoverFolder is not null)
+                w.WriteString("leftover_folder_note", PlaceWire.LeftoverNote(o.LeftoverFolder));
+            WriteNullable(w, "leftover_folder", o.LeftoverFolder);
+            if (placed > 0) w.WriteString("next_step", PlaceWire.EnableAndSort(o, modFolder));
```
Avick3110: **`next_step` can cite winners the document dropped.**

`PlaceWire.EnableAndSort` scans every entry of `o.Results` for `r.Placed && r.CurrentWinner is not null`, but the row loop above renders only what fits `cap`. Place 200 assets at `max_chars=800` where row 0 is uncontended and row 137 has a `CurrentWinner`: the loop renders row 0 and breaks, then `next_step` says "SORT it (left pane) ABOVE the current winner(s) **listed above**" while no `current_winner` appears anywhere in the document. The caller is told to sort above a mod the response never names.

The first-row rule doesn't cover this — it guarantees *a* row, not the contended one. Computing `anyContended` over the rendered rows (or naming the winners inside `next_step`) would make the sentence true of the document it ships in. The text lane has the same wording, but this is the lane whose contract is that json is never the degraded mode.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2515,6 +2540,227 @@ public static string RenderForwardOutcome(WritePatchBuilder.ForwardOutcome o, in
...
+    {
+        int cap = WriteSentences.Cap(maxChars);   // the WRITE budget rule, shared with the text twin
+        using var ms = new MemoryStream();
+        using (var w = new Utf8JsonWriter(ms, Opts))
+        {
+            w.WriteStartObject();
+            w.WriteBoolean("ok", o.Success);
+            if (!o.Success)
+            {
+                w.WriteString("error", o.Error);
```
Avick3110, line 2717: **`place` has two refusal doors that emit two document shapes.**

This branch writes `{ok, error}`. Every other refusal the tool can produce goes through `Wire.Refuse` → `JsonWire.RenderError`, which writes `{ok, error, epoch}`.

So `place` with a malformed destination answers `{"ok":false,"error":…,"epoch":null}`, and `place` with valid destinations whose `svc.PlaceAssets` fails (`PlaceOutcome.Fail("no assets to place.")`, or the rider-folder / asset-layer failures in `PlaceAssets`) answers `{"ok":false,"error":…}` with no `epoch`. A consumer reading `doc["epoch"]` on a refusal throws on one of the two.

The PR's own rule for this is "the read/write surface's one refusal shape, through its one owner" — writing the epoch key here (null, as `RenderError` does) closes it.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2515,6 +2540,227 @@ public static string RenderForwardOutcome(WritePatchBuilder.ForwardOutcome o, in
...
+            {
+                // rendered > 0: the FIRST row always renders, as it does on the text lane — and on this document it
+                // is the only place current_winner, the mod the caller has to sort above, is stated.
+                if (rendered > 0 && Over(w, ms, cap)) { truncated = true; break; }
+                WritePlaceRow(w, r, modFolder, poleWithheld?.Contains(r.AssetPath) == true);
+                rendered++;
+            }
+            w.WriteEndArray();
+
+            w.WriteNumber("total", o.Results.Count);
```
Avick3110: **The two lanes added in this PR spell their accounting differently.**

`asset_status` nests its counters under an `accounting` object; `place` writes `total` / `rendered` / `placed` / `failed` flat at the document root. A consumer written against both S2 tools reads `doc.accounting.total` for one and `doc.total` for the other.

More concretely for the #555 story in the PR body: `asset_status`'s shape was chosen so the rebase onto `TransportAccounting` is mechanical, but `place`'s counters are not in that shape, so this lane gets a second — and this time breaking — migration when #555 lands. Nesting them here now costs nothing (no consumer exists yet) and is the drift the shared composer was introduced to prevent.

Avick3110:
## Aaron's review — 5 findings, 5 folded, 0 refuted. Plus the #555 rewire.

Rebased onto `main` first (#549 and #555 are both in). Two conflicts in `AssetTools.cs`, both kept on both sides:
the empty-`asset_paths` refusal now goes through this PR's transport-aware `Refuse` while the negative
`limit`/`offset` refusal comes from `main`'s new `RowWindow.Error` (routed through `Wire.Refuse` so a json caller
still gets a document), and `AssetWire`'s private `Counts`/`Tally`/`Widest`/`Compose`/`DefaultPageLimit` are gone in
favour of `main`'s shared `TransportAccounting`.

**The #555 rewire, as the PR body promised.** `AssetWire.WriteJson` is deleted; `JsonWire.RenderAssetStatus` and its
tail reserve call `TransportAccounting.WriteJson` directly, over `TransportCounts` from `TransportAccounting.Tally` /
`Widest`. No second copy of the eight counters anywhere on the lane.

**Folded**

1. **`truncated:false` on a document that was cut (`JsonWire.cs:2591`).** Real — `counts.Truncated` counts rows, and
   the three capped caveat arrays dropped entries without touching it. The root flag is now
   `counts.Truncated > 0 || caveatsOmitted > 0`, and `truncated_note` fires on either. Test:
   `ADocumentWhoseCaveatsWereCutSaysItWasTruncated` — 400 non-matching selectors at `max_chars=2000`,
   `accounting.truncated=0`, notes dropped, root `truncated=true` with the note naming the cap.

2. **The cut marker as a data element (`JsonWire.cs:246`).** Real, and the same fix. `WriteCappedStringArray` writes
   no prose element; it returns what it dropped and writes a sibling `bsa_failures_omitted` /`warnings_omitted` /
   `selector_notes_omitted` — the capped-list-plus-count shape `Wire.ContestedHostsShown` uses. Array length plus
   the count equals the whole, so `accounting.notes` and the array no longer disagree with nothing to explain the
   gap. The three counters are written after their array has spent the budget, so `AssetTailReserve` now pays for
   them too. Test: `TheCaveatBlocksAreCappedByMaxCharsToo`, rewritten — the count reconciles, no entry contains the
   old marker text, and an uncut document reports `0`.

3. **`next_step` citing winners the document dropped (`JsonWire.cs:2733`).** Real, and on the text lane as well.
   `PlaceWire.EnableAndSort` now takes the rendered row count. A contended row that WAS rendered still gets "listed
   above"; a contended row the cap cut gets "max_chars cut the row(s) naming them from this render" instead. It does
   not fall through to the "nothing else provided these path(s)" sentence, which would have been false rather than
   merely dangling. Test: `TheEnableStepNeverCitesAWinnerTheDocumentDropped` — 41 rows, contention only past row 0,
   at `max_chars=1`; the same outcome rendered whole still says "listed above".

4. **Two refusal doors, two shapes (`JsonWire.cs:2695`).** Real. The outcome-level branch writes `epoch` (null), the
   third key `RenderError` writes. Test: `BothOfPlacesRefusalDoorsWriteTheSameDocumentShape` compares the property
   names of a `PlaceOutcome.Fail` document against a `JsonWire.RenderError` one.

5. **The two lanes spelling accounting differently (`JsonWire.cs:2720`).** Real. `place`'s counters now go through
   `TransportAccounting.WriteJson` under the same `accounting` object, so `doc.accounting.total` reads the same on
   both S2 tools and there is no second migration when a shared composer changes. This lane pages nothing, so the
   window is `RowWindow.All` and only `max_chars` can omit a row. `placed`/`failed` stay siblings of the object, not
   members — that object is the eight shared counters and nothing else, the same rule `next_limit`/`next_offset`
   already follow on `asset_status`. Test: the accounting arm in `PlaceJsonRenderTests` asserts the nesting and that
   no flat `total` remains.

One knock-on: `TheDocumentFitsTheMaxCharsItWasGiven` moved from `max_chars=1500` to `2000`. The tail genuinely grew
(three `_omitted` counters and a longer `truncated_note`), so at 1500 the reserve left room for only one row and the
arm's point — more than one row rendered, tail beginning inside the cap — no longer had anything to measure.

Build clean, suite 1327 passed, `ci-all` 128/128.

PR #562, opened 2026-09-05
Answer an unbounded references= off a lazy reverse-reference index

Avick3110:
"Who references X" over the whole order was a refusal: the reverse direction had no index, so `references=` demanded a bounding `types=`/`plugins=` scope and an agent had to guess a scope per plugin. This adds the reverse-reference index — a target-to-referencer map built off the same whole-order link walk the dangling sweep already runs — and makes the unbounded `references=` an answer. It is E1 step 3 and stacks on #558; the diff here is only this step. No vocabulary: two declared cost-refusals become values on axes that already have their spelling, exactly as the accepted E1 design page and SPEC §2.1 / §3.2 amendments of 2026-09-05 describe.

The index is lazy — nothing builds it at startup, and the first call that needs it pays and says what it paid. It is held beside the snapshot rather than on it, partitioned per plugin and keyed on that plugin's `(path, mtime)`, because `RefreshIfStale` is all-or-nothing and `epoch` is order-wide: an index keyed on those would die wholesale on every MO2 touch. A snapshot swap — and, since the review fold, a RESOLVER swap, which is what ticking a plugin in MO2 actually causes — keeps every partition whose key did not change, and a plugin whose bytes changed rebuilds only its own slice. A plugin whose enumeration throws contributes nothing (plugin-atomic, like the winner index) and is named in the accounting, so a short answer never reads as a complete one. The index supplies the scan's *candidate* universe only; the existing scan still reads the body it means to judge, so a later override that dropped the link is not counted as a referencer. Every other body filter keeps its bound — a `where=` or `editorid_contains=` with no scope is still refused, with the sentence reworded to say that only `references=` is unbounded — because the index knows links, not field values.

A refresh publishes a whole new generation of the index in one assignment and readers take it once, so an unbounded call running concurrently with a refresh sees one consistent state and takes no build lock. Partitions are keyed on the plugin path rather than its filename, because the resolver tolerates two entries sharing a name. The accounting line rides every answer the index serves, not only the call that paid the build: the build clause is conditional, the freshness key and the coverage disclosures are not.

**The negated spelling is the orphan sweep, unbounded.** SPEC §2.2's amendment ("`references=` gains a negated spelling … What the reverse-reference index buys it is the *unbounded* question — the orphan sweep") and the design page §4 ("the *unbounded* question ('referenced by nothing anywhere') — the orphan sweep") both read it that way, so that is what ships: with no `types=`/`plugins=`/`formids=` bound, the universe is every record nothing in the order references, and the named target then excludes any of those that link it. A bounded negated `references=` is unchanged — records in that scope that do not link the target — and the response states which of the two questions it answered. `HasAnyReferencer` is now one lookup against a per-generation set of every referenced target, not a scan over every partition. The sweep is refused in combination with `delta`, `tree` and `info_order`, which consume every match uncapped.

**Measured, and above the accepted figure.** On the live ARR 2.0 order (3,801 plugins, 3,214,664 records): first build 24.4 s, 1,655,669 target slots, 14,720,188 target→referencer pairs, ~203 MB measured heap, 0 unreadable plugins, 1 unscannable record; a second call on an unchanged order costs 18 ms and rebuilds nothing. SPEC §3.2's amendment and the design page state ~149 MB (629,069 distinct targets, 1,513,154 (target, plugin) pairs). The shipped index is 36% over that because it holds referencing **records**, not referencing plugins — it has to, to return rows. Memory was accepted flat with no ceiling knob, so this is a documentation delta rather than a blocker; both documents are Aaron's and are not amended here. On the test fixture: 3 partitions, 14 ms, 10 pairs.

**Left, and stated rather than worked around:** the depth > 1 reverse walk. SPEC §3.2's amendment deletes that cost-refusal too, but `walk.direction='reverse'` on this surface is the typed MGEF carrier lane (MGEF seeds, per-carrier payload rows), and the general reverse spelling is `references=` by the lane's own refusal. Making `depth > 1` generic there would mean the same spelling meaning two different things by depth, which is a shape question the design page and the SPEC do not settle. It stays refused, with the refusal now saying truthfully that the index exists and this lane is not wired to it. So this PR ships one of #489's two "What it lifts" entries; say the word and the other lands as its own step.

Part of #489 — the second half of its "What it lifts" list (the depth > 1 reverse walk) is not in this PR, so it does not auto-close.

### DIFF
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -121,6 +121,8 @@ public sealed class LoadOrderResolver : IDisposable
     readonly Dictionary<string, int> _nameToIdx;       // plugin filename → index (last copy of a duplicate name wins = priority)
     FileStamp[] _stamps;                               // last-write AND length at the last index build, per path (freshness baseline)
     readonly string? _dataDir;                         // real game-Data folder (Skyrim.esm's dir) — localized-strings fallback source (OpenOverlay)
+    ReverseReferenceIndex? _reverse;                   // lazy, held BESIDE the snapshot so a snapshot swap does not drop it
+    readonly object _reverseGate = new();              // one build at a time; a second caller waits rather than walking the order twice
 
     /// <summary>The real game-Data folder this order resolved, for callers OUTSIDE the session that must open a plugin
     /// through <see cref="OpenOverlay"/> and get the same strings resolution the index reads with. The in-place write's
@@ -914,6 +916,20 @@ public sealed class LoadOrderResolver : IDisposable
         /// <summary>Distinct children this build recorded a containing record for.</summary>
         public int ContainedRecordCount => _s.Containment.Count;
 
+        /// <summary>Every FormKey in the order, in no promised order beyond being stable for one build — the
+        /// universe an unbounded selection starts from when nothing narrows it.</summary>
+        public IEnumerable<FormKey> RecordKeys() => _s.Index.Keys;
+
+        /// <summary>Build the reverse-reference index if nothing has needed it yet, refresh the partitions whose
+        /// plugin bytes changed, and hand back what that cost — see
+        /// <see cref="LoadOrderResolver.EnsureReverseIndex"/>. The index is the resolver's, not this build's, so a
+        /// freshness rebuild between two calls keeps every partition it did not invalidate.</summary>
+        public ReverseReferenceIndex.Refreshed EnsureReverseIndex() => _r.EnsureReverseIndex(_s);
+
+        /// <summary>The reverse-reference index itself, once <see cref="EnsureReverseIndex"/> has built it; null
+        /// before that.</summary>
+        public ReverseReferenceIndex? ReverseIndex => _r.ReverseIndex;
+
         /// <summary>Every FormKey overridden by more than one plugin (the whole-order conflict set).</summary>
         public IEnumerable<FormKey> ConflictKeys() => _s.Overriders.Keys;
 
@@ -1273,6 +1289,25 @@ public sealed class LoadOrderResolver : IDisposable
         _snap = BuildIndex();                                              // ONE reference write — in-flight readers keep their captured build
         return true;
     }
+
+    // ---- The reverse-reference index: lazy, per-plugin, beside the snapshot ----------------
+
+    /// <summary>Build the reverse-reference index if this is the first call that needs it, and bring its per-plugin
+    /// partitions up to date against the files on disk. Held on the RESOLVER, not on the snapshot: the snapshot's
+    /// epoch is order-wide and a snapshot swap would drop the whole index, where this refresh rebuilds only the
+    /// partitions whose own (path, mtime) changed. Returns what the refresh cost and what it holds, for the
+    /// response's accounting.</summary>
+    internal ReverseReferenceIndex.Refreshed EnsureReverseIndex(IndexSnapshot s)
+    {
+        lock (_reverseGate)
+        {
+            _reverse ??= new ReverseReferenceIndex();
+            return _reverse.Refresh(_names, _paths, i => OpenOverlay(_paths[i], _dataDir), s.Excluded);
+        }
+    }
+
+    /// <summary>The current reverse index, or null when nothing has needed it yet.</summary>
+    internal ReverseReferenceIndex? ReverseIndex { get { lock (_reverseGate) return _reverse; } }
 
     /// <summary>The resolver holds NO plugin file handles at rest (only the pure-data index), so there is nothing to
     /// release — Dispose is a no-op, kept so the service can treat a resolver as a disposable resource it builds and
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -1309,6 +1309,17 @@ public sealed class LoadOrderResolver : IDisposable
     /// <summary>The current reverse index, or null when nothing has needed it yet.</summary>
     internal ReverseReferenceIndex? ReverseIndex { get { lock (_reverseGate) return _reverse; } }
 
+    /// <summary>Take over the reverse index of the resolver this one replaces. Ticking a plugin in MO2 rebuilds the
+    /// RESOLVER, not just the snapshot, so without this the index dies on the commonest touch there is and the next
+    /// unbounded call re-walks 3,800 unchanged plugins. The partitions are keyed on (path, mtime), so the next
+    /// refresh keeps every one whose file did not move and builds only the plugins this order added.</summary>
+    public void AdoptReverseIndexFrom(LoadOrderResolver previous)
+    {
+        var carried = previous.ReverseIndex;
+        if (carried is null) return;
+        lock (_reverseGate) _reverse ??= carried;
+    }
+
     /// <summary>The resolver holds NO plugin file handles at rest (only the pure-data index), so there is nothing to
     /// release — Dispose is a no-op, kept so the service can treat a resolver as a disposable resource it builds and
     /// swaps over its lifetime, and so `using var resolver = …` call sites stay valid.</summary>
--- a/src/housecarl-core/ReverseReferenceIndex.cs
+++ b/src/housecarl-core/ReverseReferenceIndex.cs
@@ -18,15 +18,21 @@ namespace HousecarlCore;
 /// <para><b>Partitioned by plugin, keyed on (path, mtime).</b> This is the one place the resolver's existing
 /// freshness machinery does not fit: <c>RefreshIfStale</c> is all-or-nothing and <c>Epoch</c> is order-wide, so an
 /// index hung off the snapshot would die wholesale on every MO2 touch and pay the full walk again. The index is
-/// held BESIDE the snapshot and carried across a snapshot swap; a refresh drops and rebuilds only the partitions
-/// whose own key changed. Stale-and-silent is what the partition key makes structurally impossible.</para>
+/// held BESIDE the snapshot and carried across a snapshot swap AND across a resolver swap; a refresh drops and
+/// rebuilds only the partitions whose own key changed. Stale-and-silent is what the partition key makes
+/// structurally impossible.</para>
+///
+/// <para><b>Generational, so a read is never torn.</b> A refresh builds whole new collections and publishes them
+/// with one field assignment; a reader takes the current generation once and works off it. No reader ever sees a
+/// half-rebuilt order, and no reader takes the build lock.</para>
 ///
 /// <para><b>Plugin-atomic.</b> A plugin whose enumeration throws part-way contributes NO partition content and is
 /// named in the report, so a half-read plugin never leaves partial reverse edges behind and a caller can say the
 /// answer is short rather than call it complete.</para>
 ///
 /// <para><b>Packed.</b> Entries are ulong ((interned mod index &lt;&lt; 32) | FormID), not FormKey: 8 bytes a pair
-/// instead of 24. Pure derived data — no bodies, no file handles at rest.</para>
+/// instead of 24. Pure derived data — no bodies, no file handles at rest. It is resident until the resolver it
+/// hangs off is dropped: there is no eviction policy and no ceiling knob, which is the accepted ruling.</para>
 ///
 /// <para><b>What it is not.</b> Not the position-seek index (a different question), and not reverse over the
 /// runtime layers (SkyPatcher, SPID): plugin FormLinks only, like the sweep it rides.</para></summary>
@@ -35,6 +41,7 @@ public sealed class ReverseReferenceIndex
     sealed class Partition
     {
         public string Path = "";
+        public string Name = "";
         public DateTime Mtime;
         public Dictionary<ulong, ulong[]> ByTarget = new();
         public long Pairs;
@@ -42,31 +49,66 @@ public sealed class ReverseReferenceIndex
         public string? Unreadable;                     // set ⇒ this plugin contributed nothing, and why
     }
 
-    readonly Dictionary<string, Partition> _byPlugin = new(StringComparer.OrdinalIgnoreCase);
-    readonly List<string> _order = new();              // plugin names in priority order — what makes a candidate list deterministic
-    readonly Dictionary<ModKey, int> _modToIdx = new();
-    readonly List<ModKey> _idxToMod = new();
+    /// <summary>One published, immutable state of the index. Everything a read touches lives here, so a read takes
+    /// the field once and is consistent for its whole run even while a refresh builds the next one.</summary>
+    sealed class Generation
+    {
+        // Keyed on the plugin PATH, not its filename: the resolver's order is addressed by position and tolerates
+        // two entries sharing a filename (last copy wins = priority), so a filename key would silently drop the
+        // lower-priority copy's edges and re-walk both plugins on every call.
+        public Dictionary<string, Partition> ByPath = new(StringComparer.OrdinalIgnoreCase);
+        public List<string> Order = new();             // partition paths in priority order — what makes a candidate list deterministic
+        public Dictionary<ModKey, int> ModToIdx = new();
+        public List<ModKey> IdxToMod = new();
+        public string Key = "";
+
+        Lazy<HashSet<ulong>>? _referenced;
+
+        /// <summary>Every target something in the order links, as one set — the orphan question answered in a
+        /// lookup rather than a scan over every partition. Built on first ask and thrown away with the
+        /// generation.</summary>
+        public HashSet<ulong> Referenced =>
+            (_referenced ??= new Lazy<HashSet<ulong>>(BuildReferenced, LazyThreadSafetyMode.ExecutionAndPublication)).Value;
+
+        HashSet<ulong> BuildReferenced()
+        {
+            var set = new HashSet<ulong>();
+            foreach (var path in Order)
+                if (ByPath.TryGetValue(path, out var p))
+                    foreach (var t in p.ByTarget.Keys) set.Add(t);
+            return set;
+        }
+
+        public bool TryPack(FormKey k, out ulong packed)
+        {
+            if (!ModToIdx.TryGetValue(k.ModKey, out int i)) { packed = 0; return false; }
+            packed = ((ulong)(uint)i << 32) | k.ID;
+            return true;
+        }
+
+        public FormKey Unpack(ulong packed) => new(IdxToMod[(int)(packed >> 32)], (uint)packed);
+    }
+
+    volatile Generation _gen = new();
 
     /// <summary>Plugins with a partition in this index.</summary>
-    public int PartitionCount => _byPlugin.Count;
+    public int PartitionCount => _gen.ByPath.Count;
 
     /// <summary>Distinct (target, plugin) slots — one per target a plugin links at all.</summary>
-    public int TargetSlotCount => _byPlugin.Values.Sum(p => p.ByTarget.Count);
+    public int TargetSlotCount => _gen.ByPath.Values.Sum(p => p.ByTarget.Count);
 
     /// <summary>Distinct (target, referencing record) pairs held.</summary>
-    public long PairCount => _byPlugin.Values.Sum(p => p.Pairs);
+    public long PairCount => _gen.ByPath.Values.Sum(p => p.Pairs);
 
     /// <summary>What the index holds, to the nearest useful order of magnitude: eight bytes a pair plus the
     /// per-target slot overhead of the dictionaries. Reported, not enforced — there is no ceiling knob.</summary>
     public long ApproxBytes => PairCount * 8 + (long)TargetSlotCount * 40;
 
-    /// <summary>Does anything in the order link to this record? The whole of the orphan question.</summary>
+    /// <summary>Does anything in the order link to this record? The whole of the orphan question, in one lookup.</summary>
     public bool HasAnyReferencer(FormKey target)
     {
-        if (!TryPack(target, out var pt)) return false;
-        foreach (var name in _order)
-            if (_byPlugin.TryGetValue(name, out var p) && p.ByTarget.ContainsKey(pt)) return true;
-        return false;
+        var g = _gen;
+        return g.TryPack(target, out var pt) && g.Referenced.Contains(pt);
     }
 
     /// <summary>Every record that links to ANY of these targets, deduped, in load order then plugin-enumeration
@@ -75,17 +117,18 @@ public sealed class ReverseReferenceIndex
     /// still decides on the body it means to judge (a later override may have dropped it).</summary>
     public IReadOnlyList<FormKey> ReferencersOf(IReadOnlyList<FormKey> targets)
     {
+        var g = _gen;
         var packed = new List<ulong>(targets.Count);
-        foreach (var t in targets) if (TryPack(t, out var pt)) packed.Add(pt);
+        foreach (var t in targets) if (g.TryPack(t, out var pt)) packed.Add(pt);
         var seen = new HashSet<ulong>();
         var outp = new List<FormKey>();
-        foreach (var name in _order)
+        foreach (var path in g.Order)
         {
-            if (!_byPlugin.TryGetValue(name, out var p)) continue;
+            if (!g.ByPath.TryGetValue(path, out var p)) continue;
             foreach (var pt in packed)
                 if (p.ByTarget.TryGetValue(pt, out var arr))
                     foreach (var r in arr)
-                        if (seen.Add(r)) outp.Add(Unpack(r));
+                        if (seen.Add(r)) outp.Add(g.Unpack(r));
         }
         return outp;
     }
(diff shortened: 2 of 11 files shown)


src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -0,0 +1,263 @@
...
+        foreach (var name in _order)
+            if (_byPlugin.TryGetValue(name, out var p) && p.ByTarget.ContainsKey(pt)) return true;
+        return false;
+    }
+
+    /// <summary>Every record that links to ANY of these targets, deduped, in load order then plugin-enumeration
+    /// order — deterministic for an unchanged order, which is what lets a caller's offset/limit windows tile. The
+    /// answer is a CANDIDATE set: the index says some plugin's copy of the record carries the link, and the caller
+    /// still decides on the body it means to judge (a later override may have dropped it).</summary>
+    public IReadOnlyList<FormKey> ReferencersOf(IReadOnlyList<FormKey> targets)
```
Avick3110, line 137: **Reads of the index are unsynchronized against a concurrent `Refresh`.**

`Refresh` mutates `_order` (it calls `_order.Clear()` on line 122, unconditionally, on *every* refresh), `_byPlugin`, `_modToIdx` and `_idxToMod`, and is serialised by the resolver's `_reverseGate`. But `ReferencersOf`, `HasAnyReferencer` and `Unpack` are called from `CrossQuery` **outside** that lock (`LoadOrderService.cs:3926`, via `view.ReverseIndex!`), and `CrossQuery` holds no service lock while it scans — `Resolver` takes and releases `_gate` in its getter, and the whole `_writeGate`/`_gate` design of this file presumes concurrent tool calls.

Scenario: two `records references=[...]` calls land on different threads. Thread A returns from `EnsureReverseIndex` and starts `foreach (var name in _order)` inside `ReferencersOf`. Thread B enters `Refresh`, hits `_order.Clear()`. Thread A either throws `InvalidOperationException: Collection was modified` out of the tool as an opaque transport error, or — if B has already re-added some names — silently iterates a partial plugin list and returns a **short referencer set with no note saying so**. The same window lets `Pack` grow `_idxToMod` (a `List<ModKey>`) while thread A is inside `Unpack`'s `_idxToMod[i]`, which can read a stale backing array or throw `ArgumentOutOfRangeException`.

Either hold `_reverseGate` across the read too, or make the index copy-on-write: build the new `_byPlugin`/`_order` into fresh collections and publish them with a single volatile field assignment, so a reader is always looking at one consistent generation.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /// <summary>What one refresh did, for the response's in-band accounting: the build cost when this call is the
+    /// one that paid it, the per-plugin freshness key, and every plugin the walk could not read.</summary>
+    public sealed record Refreshed(int Partitions, int Rebuilt, long ElapsedMs, int TargetSlots, long Pairs,
+                                   long ApproxBytes, IReadOnlyList<string> Unreadable, int UnscannableRecords,
+                                   string Key)
+    {
+        /// <summary>The one accounting line, or null when this call changed nothing and so paid nothing.</summary>
+        public string? Note => Rebuilt == 0 ? null
```
Avick3110: **Gating the whole note on `Rebuilt == 0` drops the coverage disclosure on every cached call, not just the build cost.**

`Note` returns null when nothing was rebuilt, so a second and every later unbounded `references=` call emits *no* reverse-index line at all. That silently loses three things the PR and the changelog say the response carries:

1. **The freshness key.** The changelog states the response "carries the index's own freshness key" — it only does so on the one call that happened to pay a rebuild.
2. **`Unreadable`** — the "the answer is short by whatever they reference" sentence. Scenario: the build call names `Foo.esp` as unreadable and says the answer is short. The agent reruns the same query (or asks a different target) on the unchanged order; `Rebuilt == 0`, the note vanishes, and the same short answer now reads as complete. `Refresh` deliberately re-populates `unreadable` from retained partitions on line 133-134, so the data is there and is then thrown away by this ternary.
3. **`UnscannableRecords`** — same, and the PR's own measurement reports 1 such record on the live order.

That is a silently degraded answer, which the cornerstone in CLAUDE.md rules out. Suggest splitting: keep the `built … in … ms` clause behind `Rebuilt > 0`, but emit `key=` plus the unreadable/unscannable clauses unconditionally.

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -104,6 +104,8 @@ public sealed class LoadOrderResolver : IDisposable
     readonly Dictionary<string, int> _nameToIdx;       // plugin filename → index (last copy of a duplicate name wins = priority)
     DateTime[] _mtimes;                                // last-write at the last index build, per path (freshness baseline)
     readonly string? _dataDir;                         // real game-Data folder (Skyrim.esm's dir) — localized-strings fallback source (OpenOverlay)
+    ReverseReferenceIndex? _reverse;                   // lazy, held BESIDE the snapshot so a snapshot swap does not drop it
```
Avick3110, line 124: **"a snapshot swap does not drop it" is true, but the resolver itself is swapped on the commonest MO2 touch — and that does drop the whole index.**

Hanging the index off the resolver survives `RefreshIfStale` (a plugin's *bytes* changed), which is what `ReverseIndexLifecycleTests.AChangedPluginRebuildsOnlyItsOwnPartition` exercises with `File.SetLastWriteTimeUtc`. But `LoadOrderService.ReResolve()` (LoadOrderService.cs:2024) does `var rebuilt = LoadOrderResolver.Build(paths, …); _resolver.Dispose(); _resolver = rebuilt;` whenever the resolved plugin *set or order* differs — i.e. whenever the user ticks or unticks a plugin, or re-sorts, in MO2. The new resolver has `_reverse == null`.

Scenario on the order the PR measured: agent runs an unbounded `references=`, pays 24.4 s. User enables one small ESP in MO2. Next tool call re-resolves, the resolver is replaced, and the next unbounded `references=` pays the full 24.4 s whole-order walk again — re-reading 3,800 plugin files whose `(path, mtime)` did not change. That is precisely the "die wholesale on every MO2 touch" failure the partition key was introduced to avoid, and no test covers it.

Carrying `_reverse` across in `ReResolve` (hand it to the rebuilt resolver, then let `Refresh` prune departed plugins and build the new ones — which it already does correctly via `live`) would make the design claim hold for the case that actually happens.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -954,10 +957,13 @@ string ScanLane()
             // subject, info_order the merge, a walk its re-entered read. The tree has no subject, so the scan's
             // selection statement stays — it is what discloses an off-order or scoped selection universe.
             bool pipelineArms = form == "delta" || form == "info_order" || walk is not null;
-            if (hasBodyFilter && !hasTypes && !hasScope && !hasFormids)
-                return Wire.Refuse(json, "error: where=/references= is a body scan and must be combined with types=, plugins=, or a formids= set to bound the work " +
-                       "(conflicts_only= alone is not enough — an unbounded body scan over the whole order is refused; the " +
-                       "reverse-reference index that would lift this is a known future capability).");
+            // An unbounded references= is answered off the reverse-reference index; every other body filter still
+            // needs a bound, because the index knows links, not field values.
+            bool onlyReverseFilter = where is not { Length: > 0 };
```
Avick3110, line 1007: **Lifting the bound here makes the whole-order universe reachable by the forms whose scan is deliberately UNCAPPED.**

Line 1065: `int effLimit = wantFile || derivedSelection ? int.MaxValue : …`, where `derivedSelection = comparisonForm || form == "info_order" || walk is not null`. That `int.MaxValue` was safe because a body filter could not reach the whole order — the refusal this hunk removes was the bound.

Concrete: `records references=["!0006BBD3:Skyrim.esm"] form=delta versus=…` is now accepted. `references` splits to `refNoneFks` only, so `CrossQuery` sees `references == null`, `ReverseSelection.Universe` returns `view.RecordKeys().ToList()` — the whole order — and the set-alone branch runs with `limit == int.MaxValue`. On the order the PR measured that is 3,214,664 winner-body fetches accumulating into `keys` **and** `prefilled` (a `RecordSummary` per match, `LoadOrderService.cs:4032`), then handed wholesale to `DeltaBatch`. Same for `form=tree`, `form=info_order`, `walk=`, and any `to_file`. `WholeOrderNote` declares "each winner body read", which is the *scan* cost; it does not declare that the derived form then compares or walks every one of them, and on these forms the note is not rendered at all (see the separate comment).

At minimum the negated-only unbounded spelling should be refused in combination with `derivedSelection`/`wantFile`, or capped with a stated ceiling.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -954,10 +957,13 @@ string ScanLane()
...
+            if (hasBodyFilter && !hasTypes && !hasScope && !hasFormids && !onlyReverseFilter)
```
Avick3110, line 1008: **`ReverseIndexNote` is only wired into the two `CrossQuery` renderers; every pipeline form on the scan discards it.**

The note is read in `ReadTools.RenderCrossQuery` and `RenderCrossQueryGroups`, and in `JsonWire.WriteNotes`. But once the scan is a *selection* for another lane, the outcome is consumed for `outcome.Keys`/`outcome.Total` and the rest is dropped:

- `walk is not null` → `WalkLane(seedKeys, …)` (line 1108)
- `comparisonForm` → `DeltaResponse` / `TreeResponse` (line 1121)
- `form == "info_order"` → `InfoOrderResponse` (line 1156)
- `form == "everything"` (and `fields` under `scopePlusPole`) → `Wire.RenderBatch` / `JsonWire.RenderBatch` (line 1175)

Scenario: `records references=["0006BBD3:Skyrim.esm"] form=everything` on a cold order. The call blocks for 24 s building the index, holds ~200 MB, and the response says nothing about either — no build cost, no freshness key, no unreadable-plugin disclosure. Worse for the negated spelling: `records references=["!…"] walk={...}` swallows `WholeOrderNote` entirely, so the caller is never told the universe was the whole order.

Since these forms already append to `headerLine`/`envelope` before dispatching, appending `outcome.ReverseIndexNote` there (once, in a helper, so no form forgets — same pattern as `CoverageNote`) would close it.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -0,0 +1,263 @@
...
+        var live = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        _order.Clear();
+        for (int i = 0; i < names.Count; i++)
+        {
+            if (excluded.Contains(i)) continue;                    // excluded from the index build; excluded here too
+            var name = names[i];
+            live.Add(name);
+            _order.Add(name);
+            var mtime = SafeMtime(paths[i]);
+            if (_byPlugin.TryGetValue(name, out var have) && have.Path == paths[i] && have.Mtime == mtime)
```
Avick3110: **Partitions are keyed on the plugin FILENAME, but the resolver's order is indexed by position and explicitly tolerates duplicate names.**

`LoadOrderResolver.Build` comments its own map as "last copy of a duplicate name wins = priority" and `BuildIndex` walks every *position*, so two entries in `_names` can share a filename with different `_paths`. This loop writes `_byPlugin[name]`, so with a duplicate:

- the lower-priority copy's reverse edges are **silently lost** — the second iteration's `have.Path != paths[i]` mismatch rebuilds and overwrites the first copy's partition, and nothing is reported;
- the mismatch is permanent, so `Rebuilt` is never 0 for those slots: **every** unbounded call re-walks both plugins, and `Refreshed.Key` changes on every call even though nothing on disk moved, breaking the "a second call rebuilds nothing / 18 ms" property this PR measures;
- `_order` gets the name twice, so `ReferencersOf` scans the surviving partition twice (harmless — `seen` dedupes — but wasted).

Keying the partition dictionary on the plugin *index* or on `paths[i]` rather than `names[i]` would match how the rest of the resolver addresses the order.

Avick3110:
Second pass on this PR, against issue #489, the accepted E1 design page and the 2026-09-05 SPEC amendments (§2.1, §2.2, §3.2). The six correctness findings are already inline; these are conformance and coverage findings that do not sit on a single changed line.

**1. Three refusal/description strings still say the index does not exist — medium.**
`src/housecarl-mcp/RecordsTools.cs:314` refuses a depth>1 reverse walk with "no index exists for it today ... (the reverse-reference index is the known future capability that lifts this)". It does exist as of this PR. Line 318 and the `walk.direction` description at line 61 likewise still say `references=` is "bounded by types=/plugins=", directly contradicting the `references=` description this PR rewrote sixty lines above. Whatever is decided about lifting the depth>1 refusal, an error message that states something untrue about the server's own capability is the one thing the errors cornerstone rules out, and the two `references=` descriptions cannot both stand.

**2. The negated-unbounded reading looks already ruled, the other way — medium.**
`ReverseSelection.NegatedMeansUnreferencedByAnything = false` (`ReverseReferenceIndex.cs:242`) is presented as awaiting a ruling, but SPEC §2.2's 2026-09-05 amendment reads: "What the reverse-reference index buys it is the *unbounded* question — the orphan sweep — which arrives on the same spelling under §3.2's amendment today." The design page §4 says the same in the same words. That is the `true` reading. Under the shipped `false`, the index buys the negated spelling nothing at all: the universe is the whole order and every winner body is read anyway, which is the outcome the amendment says the index replaces.

Two things follow. The `true` branch is dead and untested — no test exercises `HasAnyReferencer`. And as written it would not survive being switched on: `HasAnyReferencer` (`:64`) walks every partition per key, so the orphan sweep is 3,801 dictionary probes times 3.2M keys before a single body is read. Flipping the constant is not "one line plus its tests" until that is a per-target lookup rather than a per-plugin scan.

**3. The measured memory is 36% over the accepted figure and no document was amended — low.**
SPEC §3.2's amendment states the bound as fact: "~149 MB measured (629,069 distinct targets, 1,513,154 (target, plugin) pairs)". The shipped index holds 14.7M (target, record) pairs at ~203 MB. The PR body explains why and I agree the change is right — pairs are what returns rows — but the SPEC and design-page numbers are now wrong for the thing that shipped, and the diff amends neither. Memory was accepted flat, so this is documentation, not a blocker; still, the next reader takes 149 MB as the contract.

Against CLAUDE.md's "nothing holds the load order in memory": the index is derived data and is freed with its resolver, so the letter holds, but it is a second ~200 MB resident structure beside the ~150 MB winner index and nothing in the code says out loud that it is never released short of a resolver swap. One line on the class doc would settle it.

**4. `walk.direction='reverse'` with depth>1 stays refused — noted, agreed as flagged.**
SPEC §3.2's amendment deletes that cost-refusal explicitly ("Both were category-(c) declared cost-refusals with the index named as their lifter; the lifter shipping is what retires them"). The PR ships half the amendment and says so in the body rather than papering over it, which is the right call given the MGEF-lane collision it names — but as it stands this PR does not close #489, whose "What it lifts" list has two entries.

**5. Test coverage against the staleness cases — the eleven tests cover three of them.**
Covered: first build reports cost and key; a second call rebuilds nothing; a changed plugin rebuilds exactly one partition; the index survives a snapshot swap and answers from the new bytes.

Not covered:
- **A plugin removed from the order.** `Refresh`'s drop loop (`:140`) has no test, and per the inline finding on `LoadOrderResolver.cs:107` it is unreachable in production anyway — the resolver is rebuilt before it could run.
- **A plugin re-ordered.** Same path; `_order` rebuild ordering is what makes `ReferencersOf` deterministic enough for offset/limit windows to tile, and nothing asserts it.
- **A same-mtime edit.** Inherent to a `(path, mtime)` key and disclosed in the CHANGELOG, but there is no test pinning the disclosure, so the day the key changes shape nothing notices.
- **An mtime change on a master other partitions' FormKeys depend on.** Correct by construction here (a target FormKey is master-qualified and independent of the referencing partition), but that is the non-obvious half of the per-plugin key and it is asserted nowhere.
- **An unreadable plugin.** No test asserts the `Unreadable` sentence appears at all, which is why the `Rebuilt == 0` bug in the inline finding could ship.
- **The json lane.** `WriteNotes` gained the note but no test reads it back out of a json response, so text and json can drift.
- **Concurrency**, per the inline finding.

**6. Two comments no longer describe the code — low.**
`LoadOrderService.cs:4046` says "the whole-order universe is never copied into a second set" as the justification for the `!indexUniverse` guard on `setFilter`. The set-alone branch then builds `seenSet` (`:4066`) and adds every key of that same universe to it, so the copy happens ten lines later. `ReverseSelection.Universe` also materializes `view.RecordKeys().ToList()` — a second full FormKey list — before that. The guard is still worth having; the reason given for it is not the true one.

Avick3110:
Review fold. Eleven of the twelve findings are folded, one half of one is refuted, one is left as you flagged it. Build green, 1,238 tests pass (was 1,225), ci-all 128/128.

## Folded

**1 — unsynchronized reads (high).** Real, and the `_order.Clear()` window is the worst of it. `Refresh` now stages a whole new generation — partitions, order, the mod-key interning, the freshness key — and publishes it with one volatile field assignment. A reader takes the generation once and works off it, so it sees the old state or the new one and never a mixture, and it takes no build lock. Retained partitions are shared across generations, and the interning list is copied and only ever appended to, so a carried partition's packed ulongs keep meaning what they meant. Test: three reader threads against forty refreshes, asserting no throw and no empty answer.

**2 — `Note` null when nothing rebuilt (high).** Real. Split as you suggested: the build clause is behind `Rebuilt > 0` (a cached call says "unchanged, N plugin partition(s) held"), and `key=` plus the unreadable/unscannable clauses are unconditional. Two tests: a cached call still carries the key and does not claim a build; an unreadable plugin is named on the second call as well as the first.

While testing that I found the disclosure was narrower than it looked: a plugin that cannot be opened is *excluded at snapshot build*, so it never reached `BuildPartition` and the reverse index said nothing about it at all. Excluded plugins are now reported through the same sentence, which is what makes the test above possible.

**3 — the index dies with the resolver (medium).** Real, and it was the case that actually happens. `ReResolve` now hands the index to the rebuilt resolver (`AdoptReverseIndexFrom`), and the partition keys do the rest. Two tests: unticking a plugin drops exactly its partition and rebuilds nothing; re-sorting the order rebuilds nothing and still answers. Both would report a full rebuild without the carry.

**4 — the lifted bound reaches the uncapped forms (medium), in part.** Folded for `derivedSelection`: a negated-only unbounded `references=` with `delta`, `tree` or `info_order` is refused, naming the bound and pointing at the `to_file` + `formids=["@file"]` re-entry. `walk=` cannot reach it — `walk=` with `references=` is already refused as the same construct one lane up. Refutation for `to_file` below.

**5 — the note only reaches the two CrossQuery renderers (medium).** Real. One block before the dispatch chain appends it to `headerLine` and the envelope for walk / delta / tree / info_order / everything, guarded so the forms that fall through to the scan renderers (including `counts_only`) do not print it twice. Tests on `everything`, `tree` and the json lane.

**6 — partitions keyed on the filename (low).** Real. Keyed on the path now, and `_order` holds paths. Test: two mod folders shipping one filename get a partition each, and the second call rebuilds nothing — with a name key that pair rebuilt on every call forever, which as you say also breaks the freshness key's meaning.

**7 — three strings say the index does not exist (medium).** Real; all three reworded. The depth>1 refusal now says the index ships and this lane is not wired to it, which is the true statement.

**8 — the negated-unbounded reading (medium).** You are right and I have implemented it. SPEC §2.2's amendment and design page §4 both name the orphan sweep in the same words, so that is the ruling and the `unruled` language is gone with the constant. Shipped semantics: with no `types=`/`plugins=`/`formids=` bound the universe is every record nothing in the order references, and the named target then excludes any of those that link it (the ordinary AND the term is everywhere else); a bounded negated `references=` is unchanged, and the response states which of the two it answered, since one spelling asking two questions on the presence of a scope is exactly the thing that must not be silent. `HasAnyReferencer` is now a single lookup against a per-generation set of every referenced target, built on first ask — your 3,801 × 3.2M probe count is gone. Tests: the sweep excludes a record something links, keeps an orphan, and declares itself.

**9 — 203 MB against the accepted 149 MB (low).** Agreed, and the PR body now states both numbers and why they differ. SPEC and the design page are Aaron's; not amended here. The class doc now says out loud that the index is resident until the resolver it hangs off is dropped, with no eviction policy — that was the missing sentence.

**11 — the uncovered staleness cases (medium).** All seven covered: plugin removed, order re-sorted, an edit that preserves the mtime (pinning the disclosed limit of the key), a touched master rebuilding only its own partition while other partitions' answers about its records still stand, the unreadable disclosure, the json note, and concurrency. Your note that the drop loop was unreachable in production was correct at the time; with the resolver carry in finding 3 it is now the live path, and the removed-plugin test drives it.

**12 — the false comment (low).** Real. Rewritten to say what the guard actually buys (an unread pre-filter copy), and it no longer claims nothing copies the universe.

## Refuted

**4, the `to_file` half.** `wantFile` is not refused with the sweep. The orphan list is precisely the artifact this feature exists to produce, and `to_file` adds no per-match work beyond the scan the response already declares — the derived forms are refused because they *compare or merge* every match on top of it, which is a different order of cost. The scan's own note (whole-order universe, build cost, freshness key) does render on the `to_file` lane, so nothing about it is undeclared. If you want a ceiling there anyway it is a knob, not a bug.

## Left

**10 — the depth > 1 reverse walk.** As flagged, and `Closes #489` is now `Part of #489` so it does not auto-close an issue whose second entry is unshipped.

## Counts

+13 tests (24 in the reverse-index file, was 11), no new probes; 128/128 ci-all unchanged.

Avick3110:
Blind review folded (11 folded, one half-finding refuted with reason, the reverse-walk depth half left; the negated unbounded form is now the orphan sweep per SPEC §2.2 and the design page). Queued for Aaron's review. Merge after #558. #489 stays open for the reverse-walk depth half.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1054,6 +1063,13 @@ string ScanLane()
                 if (neg!.Count > 0) refNoneFks = neg.Distinct().ToList();
             }
 
+            // The negated-only unbounded spelling selects the whole orphan set — millions of records on a real
+            // order — and the derived forms consume EVERY match uncapped (effLimit below is int.MaxValue for them),
+            // so each match would also be compared, merged or walked. Refused with the bound named, rather than
+            // run. A positive references= is not this: its universe is what links the target.
+            if (refNoneFks is not null && refFks is null && !hasTypes && !hasScope && !hasFormids && derivedSelection)
```
Avick3110: **Correctness — an in-order `source=` bounds the scan, but this guard does not see it.**

The test is `hasScope` (i.e. `plugins.names`), not the `scanPlugins` the call actually scans with. Three lines down, `scanPlugins = scopePlusPole ? plugins!.names : (srcName is not null ? new[] { srcName } : plugins?.names)` — so a named in-order `source=` reaches `CrossQuery` as `plugins=[srcName]`, takes the bounded `hasPlugins` path, and never touches the reverse index at all.

Concrete: `records references=["!0006BBD3:Skyrim.esm"] source="MyPatch.esp" project.form="tree"` is refused with *"a negated references= with no types=/plugins=/formids= bound is the orphan sweep … the 'tree' form compares or merges EVERY match, uncapped"* — but nothing about that call is the orphan sweep; it is one plugin's records. I ran it against `RecordsWorld` to confirm.

The `group_by='type'` guard eleven lines below already reads `scanPlugins` for exactly this reason, and its comment says so: *"read off scanPlugins, so an in-order source= plugin counts as the scope it is."* Same fix here.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1054,6 +1063,13 @@ string ScanLane()
...
+                return Wire.Refuse(json, $"error: a negated references= with no types=/plugins=/formids= bound is the orphan sweep — every record nothing in the order references — and the '{form}' form compares or merges EVERY match, uncapped. Add types= or plugins= to bound it, or run the sweep as a plain scan with to_file= and re-enter the artifact via formids=[\"@<file>\"].");
+
             var scanPlugins = scopePlusPole ? plugins!.names : (srcName is not null ? new[] { srcName } : plugins?.names);
             bool definedIn = plugins?.defined_in ?? false;
             // group_by=type names each match's record type, which only a body-bearing scope can supply. The engine
```
Avick3110, line 1100: **Stale refusal (the guard just below, lines 1080–1086) — `group_by='type'` still turns away an unbounded `references=`.**

The engine's own equivalent check (`LoadOrderService.cs:4034`) now *passes* for an unbounded `references=`, because the index branch sets `hasFormidSet = true` and the set-alone branch has each match's body in hand to name its type. But this tool-level pre-check tests `formidSet is null` — the caller's list, not the index universe — so it refuses first.

Concrete: `records references=["0006BBD3:Skyrim.esm"] project.form="aggregate" project.group_by="type"` returns *"project.group_by='type' counts each match's record TYPE, which only a body-bearing scope can name — add types=, plugins=, formids=, or an in-order source= plugin. (where= and references= read bodies too, but each takes one of those as its own bound; …)"*. The same call with `group_by='winner'` is served, index accounting and all — so the capability is there and only this pre-check blocks it.

The parenthetical in that message ("references= … takes one of those as its own bound") is also a sentence this PR made false.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -0,0 +1,325 @@
...
+            {
+                var sb = new StringBuilder("reverse-reference index: ");
+                sb.Append(Rebuilt > 0 ? $"built {Rebuilt} plugin partition(s) in {ElapsedMs} ms"
+                                      : $"unchanged, {Partitions} plugin partition(s) held");
+                sb.Append($" ({Pairs} target→referencer pairs over {TargetSlots} target slots, ~{ApproxBytes / (1024 * 1024)} MB held), ");
+                sb.Append($"key={Key} (per plugin, path+mtime — beside the order-wide epoch, not riding it).");
+                if (Unreadable.Count > 0)
+                    sb.Append($" {Unreadable.Count} plugin(s) contributed nothing because the walk could not read them: ")
+                      .Append(string.Join(", ", Unreadable))
+                      .Append(" — the answer is short by whatever they reference.");
```
Avick3110: **The direction of this disclosure is inverted for the orphan sweep.**

"the answer is short by whatever they reference" is right for a positive `references=`: edges the walk could not read mean referencers are missing from the result. The orphan sweep inverts it — a record is an orphan *because* nothing links it, so an unreadable plugin's missing edges make records look unreferenced. The sweep is **over**-inclusive there, not short.

Concrete: plugin `U` is the only thing in the order linking record `R`, and `U` cannot be opened or enumerated by the walk. `records references=["!XXXXXX:A.esm"]` lists `R` among "the N record(s) nothing in the order references", and the sentence beside it tells the caller the answer is *short* by what `U` references. A caller who reads the disclosure concludes `R` is a confirmed orphan and acts on it.

I confirmed the sentence is emitted verbatim on an orphan-sweep response after corrupting a plugin in `RecordsWorld` — the note and the universe note are simply concatenated, with no branch on which question was asked. `Refreshed.Note` doesn't know which lane it is serving; `ReverseSelection.UniverseNote` does, so the over-reporting wording probably belongs there.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -0,0 +1,325 @@
...
+    /// nothing in the order references, which is the unbounded question SPEC §2.2's and §3.2's 2026-09-05
+    /// amendments name for this spelling. The named negated targets still apply after it, as the AND term they are
+    /// everywhere else.</summary>
+    public static IReadOnlyList<FormKey> Universe(LoadOrderResolver.IndexView view, ReverseReferenceIndex index,
+                                                  IReadOnlyList<FormKey>? references)
+    {
+        if (references is { Count: > 0 }) return index.ReferencersOf(references);
+        var orphans = new List<FormKey>();
+        foreach (var k in view.RecordKeys())
+            if (!index.HasAnyReferencer(k)) orphans.Add(k);
```
Avick3110: **The orphan universe is not built against one generation, though the class promises it is.**

The header says *"a reader takes the current generation once and works off it. No reader ever sees a half-rebuilt order"*, and `ReferencersOf` honours that (`var g = _gen;` once). This loop does not: each `index.HasAnyReferencer(k)` re-reads the volatile `_gen` and re-takes `g.Referenced`, so the universe is assembled from however many generations happen to be published while it runs.

Concrete: a second tool call refreshes the index (a plugin's mtime moved) while a sweep is walking 3.2M keys. Keys enumerated before the swap are judged against the pre-rebuild edges and keys after against the post-rebuild ones — a record whose only referencer lives in the rebuilt plugin lands on either side depending on where the swap fell. The `key=` the response reports names a single generation, so the answer is not reproducible from the key it cites, which is the one thing that key is for.

Also a cost cliff rather than a correctness one: crossing a generation mid-loop discards the memoised `Referenced` set and rebuilds a ~1.6M-entry `HashSet` inline.

Hoisting the generation once — an overload of `HasAnyReferencer` that takes a captured `Generation`, or lifting `Referenced` out of the loop into a local — closes both.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -0,0 +1,325 @@
...
+        public List<ModKey> IdxToMod = new();
+        public string Key = "";
+
+        Lazy<HashSet<ulong>>? _referenced;
+
+        /// <summary>Every target something in the order links, as one set — the orphan question answered in a
+        /// lookup rather than a scan over every partition. Built on first ask and thrown away with the
+        /// generation.</summary>
+        public HashSet<ulong> Referenced =>
+            (_referenced ??= new Lazy<HashSet<ulong>>(BuildReferenced, LazyThreadSafetyMode.ExecutionAndPublication)).Value;
```
Avick3110: **Nit / thread-safety: this memo can be built more than once.**

`_referenced ??= …` is a non-atomic check-then-assign on a field of a generation every reader shares, and the field is not `volatile`. `LazyThreadSafetyMode.ExecutionAndPublication` guards the `Lazy` instance a thread already holds; it does not guard the field write that publishes it, so two readers can each construct a `Lazy` and each run `BuildReferenced`. (No correctness problem — each thread uses the instance its own `??=` evaluated to.)

Concrete: two orphan sweeps overlap on the 3,801-plugin order. Both see null, both walk every partition to build a ~1.6M-entry `HashSet`, and both are resident at once — on top of the ~200 MB the index already holds, in a design that deliberately ships no ceiling knob.

`Interlocked.CompareExchange` on the field, or just making it a `readonly Lazy<…>` initialised with the generation, makes it once-only.

Avick3110:
Aaron's review folded. All five findings are real; nothing refuted. Rebased on `claude/e1-parent` (clean, no conflicts). Build green, 1,317 tests pass, ci-all 128/128.

## Folded

**1 — an in-order `source=` bounds the scan, but the sweep guard did not see it (`RecordsTools.cs:1070`).** Real. `scanPlugins` is now computed above the guard and the guard reads it, so `source="MyPatch.esp"` counts as the scope it is and its one plugin's records are not called the sweep. Test: a negated `references=` under an in-order `source=` with `form='tree'` is served, not refused.

**2 — the stale `group_by='type'` refusal (`RecordsTools.cs:1075`).** Real, and the capability was there behind the pre-check — the engine's own equivalent passes because the index branch sets `hasFormidSet`. The pre-check now exempts `references=` (positive or negated), names `references=` among the bounds that work, and the parenthetical no longer claims `references=` takes a bound of its own. Test: an unbounded `references=` groups by type and carries the index accounting.

**3 — the unreadable disclosure is inverted for the sweep (`ReverseReferenceIndex.cs:157`).** Real, and it is the one direction where a caller acts on the sentence. `Refreshed` gained `NoteFor(bool orphanSweep)` (`Note` is the positive reading); the sweep now reads "the sweep is OVER-inclusive by whatever they reference: a record only they link is listed here as an orphan." The lane is known at the call site in `CrossQuery`, so the note is told for the question that was asked. You put the wording in `UniverseNote`; I kept it on the disclosure itself so the fact and its consequence stay in one sentence rather than the consequence arriving in a second note that may or may not be appended. Test: an unreadable plugin says OVER-inclusive on the sweep and short on the positive question, and neither says the other.

**4 — the sweep universe is not built against one generation (`ReverseReferenceIndex.cs:312`).** Real, both halves. The per-key `HasAnyReferencer` loop is replaced by `ReverseReferenceIndex.Orphans(candidates)`, which takes `_gen` and its memoised `Referenced` once for the whole pass — so the answer matches the `key=` it cites, and a crossing no longer rebuilds the ~1.6M-entry set inline. Test: a candidate sequence that publishes a new generation mid-pass (the plugin holding every edge made unreadable) still judges the later keys against the generation the sweep started on. Confirmed it fails on the old per-key form.

**5 — the memo can be built more than once (`ReverseReferenceIndex.cs:71`).** Real. `_referenced` is now a `readonly Lazy<…>` built in `Generation`'s constructor, so once-only is structural rather than racing a check-then-assign. No dedicated assertion — there is nothing observable to assert once a generation can only hold one `Lazy` — but a new test runs four sweeps in parallel off one generation and asserts they agree.

## Counts

+5 tests (29 in the reverse-index file, was 24), no new probes; 128/128 ci-all unchanged. The changelog's two `references=` entries gained the `group_by='type'` composition, the in-order `source=` bound, and the sweep's over-inclusive disclosure.

PR #564, opened 2026-09-05
Name a withdrawn Nexus file, and add size to the freshness key

### LINKED ISSUE #405: [Bug] nexus_check_updates renders REMOVED-category files as "— current"

### LINKED ISSUE #406: [Bug] read-cache freshness key is mtime-only — a same-mtime different-size edit is invisible

Avick3110:
Two small read-honesty bugs, stacked on #554 (`claude/check-offorder-395`) because it is the branch in flight; nothing here depends on that work.

**#405 — `nexus_check_updates` rendered a REMOVED file as current.** `IsSuperseded` is the closed `OLD_VERSION`/`ARCHIVED` set, so a `REMOVED` or `DELETED` file fell through to `FileVerdict.Live` and the renderer appended `" — current"`: `'name' v1 [REMOVED] — current`, one line contradicting itself. Withdrawal is now its own bucket (`IsRemoved`, `FileVerdict.Removed`, `UpdateVerdict.FileRemoved`), with its own group at the top of the report and its own count in the summary. The row says the file is no longer offered and to read the page — description, posts, changelog — for why before installing anything in its place; a live file with that exact name is named as a lead, never as the fix. A withdrawn file is also no longer offered as the replacement for a *retired* one: the same-name search tests `IsLive` rather than merely not-retired. An unfamiliar category is still treated as live and printed as-is, which was the deliberate forward-compat behaviour and is unchanged. Five new cells in `nexus-file-check-guard`, including one that asserts the render never says "current" beside a `[REMOVED]` category.

**#406 — the read-cache freshness key was mtime-only.** `RefreshIfStale` compared `File.GetLastWriteTimeUtc` alone and `ComputeEpoch` hashed `name|path|mtime.Ticks`, so an edit inside the filesystem's timestamp granularity, or one whose tool restored the timestamp it found, served stale parsed state silently. Both keys now carry the file's length beside its last-write, through one `FileStamp` read by a single `FileInfo` stat — no extra syscall on the cheap no-change path. A content hash was not added: reading every plugin's header on every freshness sweep costs an open per plugin per call, which the "freshness is cheap" rule rules out, and size catches the case the issue names. The remaining bounds — an edit that changes neither term, and the known write-during-rebuild race — are stated at `ComputeEpoch` rather than papered over. `FreshnessKeyTests` rewrites a plugin to a different length and puts the timestamp back, then asserts the rebuild fires, the epoch moves, and an untouched order is still reported unchanged.

Build clean; 1174 tests pass (was 1171); `ci-all` ALL PASS.

Closes #405
Closes #406


src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -629,15 +630,17 @@ IndexSnapshot BuildIndex()
...
+    /// re-reading them, so a plugin rewritten mid-rebuild can pair its new stamp with old content until its next
+    /// change — the pre-existing freshness-baseline race, which this fingerprint shares by construction. An edit
+    /// that changes neither the last-write time nor the LENGTH is still invisible to both.</para></summary>
+    static string ComputeEpoch(string[] names, string[] paths, FileStamp[] stamps, Dictionary<string, string> excludedPlugins)
     {
         var sb = new StringBuilder(names.Length * 96);
         for (int i = 0; i < names.Length; i++)
-            sb.Append(names[i]).Append('|').Append(paths[i]).Append('|').Append(mtimes[i].Ticks).Append('\n');
+            sb.Append(names[i]).Append('|').Append(paths[i]).Append('|').Append(stamps[i].Mtime.Ticks)
+              .Append('|').Append(stamps[i].Size).Append('\n');
```
Avick3110, line 643: Adding the size term changes the epoch *format*, and epochs are persisted on disk in result artifacts (`ResultArtifact.cs:148` writes `epoch` into the manifest; `ArtifactDemand` re-checks it on every `@<path>` re-entry). After a user upgrades to this build, every artifact written by an older houseCARL fingerprints differently even over a genuinely unchanged load order, so `LoadOrderService.ArtifactEpochMismatch` refuses it with a claim that is false: "the load order changed since the artifact was written". There is deliberately no stale-override switch, so the only way out is re-running the producing query — for a large `to_file=` catalogue that is the expensive path, taken for a reason the message never names.

Concrete: user runs `cross_plugin_query ... to_file=armor.jsonl` on Monday, upgrades houseCARL, and on Tuesday `formids=@armor.jsonl` refuses, with nothing on disk having changed.

Worth either a changelog line telling users to regenerate artifacts after this release, or a version term in the epoch input so the mismatch message can distinguish "the formula changed" from "your load order changed".

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -1,4 +1,4 @@
-using System.Text;
+﻿using System.Text;
```
Avick3110: This hunk adds a UTF-8 BOM (`EF BB BF`) to the head of the file — the old line is bare `using System.Text;`, the new one is BOM + the same text. No other .cs file in `src/housecarl-core` has one (checked `AssetResolver.cs`, `WritePatchBuilder.cs`, `ResultArtifact.cs`). It compiles fine, but it is an unintended whole-file encoding change from an editing tool, and it makes this file the odd one out for anything that reads sources as bytes. Worth stripping before merge.

Avick3110:
Second review pass (blind, own diff only). Two findings are already posted inline (the epoch-format break against week-old artifacts, and the UTF-8 BOM added to `LoadOrderResolver.cs`). Four more, all low, none blocking:

**1. Not every freshness reader carries the size term, but `CLAUDE.md` now states it as the project-wide model.** The design line was changed to "a change on disk is picked up by a last-write-plus-size check", and that is now true only of the plugin read cache. Still last-write alone, in the same shape #406 names:

- `src/housecarl-mcp/LoadOrderService.cs:1948` / `:1961` — the MO2 profile gate (`plugins.txt`, `loadorder.txt`, `modlist.txt`) via its own `SafeMtime`. Lower stakes than the resolver, since `RefreshIfStale` still runs per call at `:145` and a pure reorder does not change size either — but a same-mtime `modlist.txt` rewrite still leaves the set behind.
- `src/housecarl-core/AssetResolver.cs:559`, `:577`, `:654` — the BSA archive and loose-subtree caches, which are the sibling read cache and have no other freshness signal.

`src/housecarl-core/SkyPatcherDiscovery.cs:80` already keys on `(mtime, length)`, so the repo now holds three different freshness keys for the same job. Either widen the two above or narrow the `CLAUDE.md` sentence to the plugin read cache. Relatedly, `README.md:158` still says "cheap mtime checks" and was not updated with `CLAUDE.md`.

**2. `src/housecarl-mcp-tests/FreshnessKeyTests.cs:40` — the premise is asserted before the act, so the test can go green for the wrong reason.** `EditWithoutMovingMtime` restores the timestamp and asserts it took, then the caller runs `RefreshIfStale()`. If the last-write ever drifts between those two points (a lazily flushed write, a virus scanner touching the file), the refresh reports stale on the *mtime* term and the assertions all pass while proving nothing about size. Cheap hardening: re-assert `File.GetLastWriteTimeUtc(_path) == before` after the refresh, or assert directly that the two stamps differ only in length. As written the test does fail against the pre-fix key (verified by reading it; the restore is exact on NTFS) — this is about it staying a real test.

**3. `src/housecarl-mcp/NexusTools.cs:527` — the summary line and the groups disagree on order.** The summary prints `N outdated · N file-removed · …` while the groups below it, and the method's own doc comment at `:509`, both put file-removed first as the most actionable. Lead the summary with file-removed too.

**4. `src/housecarl-mcp/NexusClient.cs:136-139` — the "what each verdict means" doc was not updated for the new verdict.** `ComputeStatus` points readers at `CheckUpdatesAsync` for the verdict semantics, and that summary still reads "a live category (MAIN, UPDATE, OPTIONAL, MISCELLANEOUS) is Current; OLD_VERSION or ARCHIVED is Outdated and points at the newest same-name file; absent from the page is FileGone" — no withdrawal bucket. The enum's own doc was updated; this one is the home the code cites.

Verified against the branch: build clean, `FreshnessKeyTests` 3/3 pass, `nexus-file-check-guard` all PASS including the five new K cells. `SafeStamp` is genuinely one stat and no open — on Windows `FileInfo.Exists` refreshes through `GetFileAttributesEx`, and `LastWriteTimeUtc` and `Length` both read that cached result, so the no-change path costs exactly what the mtime-only path cost. The withdrawal bucket itself reads correct: closed sets, `Removed` outranking `Superseded` outranking `Missing`, the replacement search testing `IsLive`, unknown categories still Live and printed as-is, and all eight verdicts counted in the summary.

Avick3110:
Review fold. All six findings were real; all six are folded, none refuted.

**1. The epoch format break against artifacts on disk (medium).** Real, and the sentence was the wrong one. The epoch string now carries a format tag (`e2-` + the 16 hex chars), so an epoch written before the formula changed is recognisable rather than merely unequal. `ArtifactEpochMismatch` picks its middle sentence off that: a comparable epoch still reads "the load order changed since the artifact was written"; an incomparable one reads "that epoch was written by an OLDER houseCARL, before the fingerprint formula changed, so the two cannot be compared: your load order may be untouched, and this build still cannot tell." The next move (re-run the producing query) and the no-override posture are unchanged, and a CHANGELOG entry tells users to re-run artifacts once after this release. No stale-override switch was added.

Two knock-ons. `epoch-guard`'s shape check now asserts the tag instead of bare 16-hex, plus a new cell that a legacy-shaped epoch is distinguishable from one this build wrote. And the three-character-longer envelope moved four hand-pinned cap boundaries in `RecordsOwnedChildTests` (587/588, 1077/1078, 1911/1912 → 590/591, 1080/1081, 1914/1915); those are calibration constants for the reserve logic and their comments moved with them.

**2. The UTF-8 BOM (low).** Stripped; `LoadOrderResolver.cs` is bare `using System.Text;` again.

**3. Three freshness keys for one job (low).** Widened rather than narrowed, as the finding's first option. There is now one `FileStamp` (`src/housecarl-core/FileStamp.cs`) — `(last-write, length)`, one stat, one absent sentinel — and everything reads it: the plugin read cache, the MO2 profile gate and the ModOrganizer.ini baseline (`LoadOrderService`), the BSA tables and the loose-subtree cache (`AssetResolver`), and `SkyPatcherDiscovery.ParseCache`, which had the pair inline. Directories go through `FileStamp.OfDirectory` — a directory has no length, so the size term is pinned at 0 and the last-write carries the signal, with the sentinel still separating present from absent. `README.md`'s "cheap mtime checks" is now "cheap last-write-plus-size checks", and the same prose in `AssetTools`, `AssetResolver` and `LoadOrderService`'s headers went with it. Two new facts in `FreshnessKeyTests` pin the shared key: two files differing only in length stamp differently, and the absent sentinel behaves for files and directories alike.

**4. The premise asserted before the act (low).** `EditWithoutMovingMtime` now returns the timestamp it restored and both tests re-assert it after the refresh, so a stamp nudged in between cannot leave the test green off the mtime term.

**5. Summary order (low).** The counts now lead with file-removed, matching the groups and the method's own doc. A new probe cell renders a response carrying both verdicts and pins summary order against group order.

**6. The `CheckUpdatesAsync` verdict doc (low).** Updated: REMOVED/DELETED is FileRemoved, the same-name live file is a lead and not a replacement, and FileRemoved outranks Outdated when a mod has both.

Build clean, `dotnet test --filter "tier!=bridge"` 1177/1177 (three new tests), `ci-all` 128/128 ALL PASS.

Avick3110:
Blind review folded (6 of 6 real, 0 refuted; epoch string now carries a format tag, artifacts from older builds get an honest sentence). Queued for Aaron's review. Merge after #554.

src/housecarl-mcp/NexusTools.cs
```diff
@@ -505,26 +507,31 @@ static string HumanSize(long bytes)
...
         int outdated = results.Count(r => r.Verdict == UpdateVerdict.Outdated);
+        int removed = results.Count(r => r.Verdict == UpdateVerdict.FileRemoved);
         int current = results.Count(r => r.Verdict == UpdateVerdict.Current);
         int fileGone = results.Count(r => r.Verdict == UpdateVerdict.FileGone);
         int noFileId = results.Count(r => r.Verdict == UpdateVerdict.NoFileId);
         int latest = results.Count(r => r.Verdict == UpdateVerdict.LatestOnly);
         int notFound = results.Count(r => r.Verdict == UpdateVerdict.NotFound);
         int errored = results.Count(r => r.Verdict == UpdateVerdict.Error);
 
+        // Same order as the groups below and the doc above: the most actionable verdict leads both.
```
Avick3110, line 526: **[low] The comment states an order invariant the code does not hold.**

The summary emits `file-removed · outdated · current · file-gone · no-fileid · latest-only · not-found`, but the `AppendUpdateGroup` calls below run `FILE REMOVED, OUTDATED, FILE GONE, NO FILEID, current, latest version, not found`. `current` is third in the counts and fifth in the groups.

Concretely: a check whose results include one Current mod and one FileGone mod prints "current" as the third count and the fifth group. The K-cell probe only asserts `file-removed < outdated`, so nothing checks the sentence, and the next person adding a verdict "in the same slot in both" will put it in the wrong one.

Either reorder the summary to match the groups, or narrow the comment to what is true — that the most actionable verdict leads both.

src/housecarl-generator/EpochGuardProbe.cs
```diff
@@ -42,7 +42,11 @@ public static int RunGuard(string[] args)
         Console.WriteLine();
         int fail = 0;
         void Check(bool c, string label) { Console.WriteLine((c ? "  PASS  " : "  FAIL  ") + label); if (!c) fail++; }
-        static bool IsEpochToken(string? e) => e is { Length: 16 } && e.All(ch => ch is (>= '0' and <= '9') or (>= 'a' and <= 'f'));
+        // A format tag, a dash, then 16 hex chars. The tag is what lets a consumer tell an epoch this build cannot
+        // compare against (one written before the formula changed) from one that simply names a different world.
+        static bool IsEpochToken(string? e) =>
```
Avick3110, line 49: **[low] The guard hard-codes the tag width it exists to police.**

`Length: 19` and `Substring(3)` are both silently derived from `EpochFormat.Length == 2`. `EpochFormat`'s own doc says it is "bumped whenever `ComputeEpoch`'s input changes shape" — so a bump is the expected event, and the first one to a three-character tag (`e10`, or any `v2`-style scheme) makes every valid epoch 20 chars: `IsEpochToken` returns false, arm 1 fails claiming the epoch is not an opaque token, and the hex check starts at the wrong offset. The guard fires on the change it was written to survive.

Deriving both from the format itself keeps them in step:

```suggestion
        static bool IsEpochToken(string? e) =>
            e is not null && LoadOrderResolver.IsCurrentEpochFormat(e)
            && e.AsSpan(e.IndexOf('-') + 1) is { Length: 16 } hex
            && hex.ToString().All(ch => ch is (>= '0' and <= '9') or (>= 'a' and <= 'f'));
```

Separately, the arm-1 doc at lines 15–17 still says the fingerprint is over "plugin names + mtimes" and that the format contract is "16 lowercase hex" — both are now stale.

src/housecarl-mcp/NexusClient.cs
```diff
@@ -256,21 +269,25 @@ internal static NexusUpdateStatus ComputeStatus(int modId, bool found, string? n
...
                 {
                     // Point to the newest live file with the same name — the variant line's replacement. Left null if
                     // the author renamed or dropped the variant, rather than guessing the wrong file.
                     string? rn = null, rv = null; long rd = 0;
                     foreach (var f in files)
-                        if (!IsSuperseded(f.category) && string.Equals(f.name, hit.name, StringComparison.OrdinalIgnoreCase) && (rn is null || f.date > rd))
+                        if (IsLive(f.category) && string.Equals(f.name, hit.name, StringComparison.OrdinalIgnoreCase) && (rn is null || f.date > rd))
                             { rn = f.name; rv = f.version ?? "?"; rd = f.date; }
-                    detail.Add(new InstalledFileCurrency(fid, hit.name, hit.version, hit.category, FileVerdict.Superseded, rn, rv, rd));
+                    var withdrawn = IsRemoved(hit.category) ? FileVerdict.Removed : FileVerdict.Superseded;
```
Avick3110: **[low] `withdrawn` holds `Superseded` on the retirement path.**

This branch now serves both buckets, and for an `OLD_VERSION`/`ARCHIVED` hit the expression evaluates to `FileVerdict.Superseded` — so the local named `withdrawn` carries the value that is explicitly *not* a withdrawal, on what is likely the more common half of the traffic. A reader skimming the branch takes the name as the branch's meaning and concludes retired files are bucketed as withdrawals, which is the exact conflation the rest of this PR is separating.

```suggestion
                    var fileVerdict = IsRemoved(hit.category) ? FileVerdict.Removed : FileVerdict.Superseded;
                    detail.Add(new InstalledFileCurrency(fid, hit.name, hit.version, hit.category, fileVerdict, rn, rv, rd));
```

Avick3110:
Aaron's review fold. All three findings were real; all three folded, none refuted.

**1. `NexusTools.cs:526` — the comment stated an order invariant the code did not hold.** Real: the counts read `file-removed · outdated · current · file-gone · no-fileid · …` while the groups (and the method's own doc) run `FILE REMOVED, OUTDATED, FILE GONE, NO FILEID, current, …` — `current` third in one and fifth in the other. Folded by reordering the summary to the groups, which is the order the doc already claims, rather than narrowing the comment. The K cell that only pinned `file-removed < outdated` is widened: one render carrying all seven verdicts, asserting the count tokens and the group labels each appear in the same rising order, so the next verdict added in the wrong slot fails here. A changelog clause says the counts read in group order.

**2. `EpochGuardProbe.cs:47` — the guard hard-coded the tag width it exists to police.** Real: `Length: 19` and `Substring(3)` were both silently `EpochFormat.Length == 2`, so the first three-character tag (`e10`) would make `IsEpochToken` reject every valid epoch and read the hex from the wrong offset — the guard firing on the change it was written to survive. Both are now read off the dash: `e.Substring(e.IndexOf('-') + 1) is { Length: 16 } hex`. A new cell pins the tag's width out of the contract (`Length == IndexOf('-') + 17`). The stale arm-1 doc went with it: the fingerprint is over plugin names + last-write + size, and the format contract is the tag then 16 lowercase hex.

**3. `NexusClient.cs:280` — `withdrawn` held `Superseded` on the retirement path.** Real, and it is the conflation the rest of the PR separates. Renamed to `fileVerdict`, with a line saying the branch serves both buckets.

Build clean, `dotnet test --filter "tier!=bridge"` 1258/1258, `ci-all` 128/128 ALL PASS. Two new probe cells, no new xUnit tests — both findings are probe-side.

Issue #526, opened 2026-09-04
[Bug] Add appends a FormLink the list already carries, and the readback cannot be told from a clean add


Avick3110:
From a Discord report (2026-08-25, houseCARL 1.9). Reproduces on `main`.

`verb=Add` on a FormLink list appends the element unconditionally: the list Add in `WriteEngine.cs` is a plain `List.Add` with no membership test. `housecarl_apply op=Add` reaches the same code. The readback prints only the new count, so a duplicating add and a clean add look identical. The reporter wrote duplicate keywords across 126 records before noticing.

Fix in scope: a `duplicate: already present` note on the op line, in the readback. No new vocabulary.

Out of scope, separate decision for Aaron: whether a `verb=AddUnique` or `dedupe=` flag should exist. That is a SPEC change and would need to land before W6.

Rides W5 item 19 (silent wrong answers).

PR #566, opened 2026-09-05
Fix two silent wrong answers on the records lane

### LINKED ISSUE #448: [Bug] housecarl_records format=dense silently drops project.depth>1 (the 1.x sibling refuses it loud)

### LINKED ISSUE #449: [Bug] to_file= artifact rows carry scoped field values without the scoped-vs-winner note

### LINKED ISSUE #527: [Bug] a concrete-arm type filter (GlobalShort, GameSettingInt) returns the whole abstract group: the in-order scan lane never re-checks the arm

Avick3110:
Two silent-wrong-answer bugs on the records lane. Both returned or wrote something that read as an answer to the question asked and was an answer to a different one.

**#527 — a concrete-arm type filter returned the whole abstract group.** `types=["GlobalShort"]` and `types=["GlobalFloat"]` both returned the identical whole-GLOB set, mixed, with nothing saying the filter had widened. `ResolveTypeFilter` maps the arm to `IGlobalShortGetter` correctly; the in-order scan then hands that type to Mutagen's `EnumerateMajorRecords`, which seeks the GRUP — and for an abstract group the GRUP is every arm of it. The off-order lane already re-checks the arm per record (`LoadOrderService.cs`, `types.Any(t => t.IsInstanceOfType(rec))`); the two in-order enumerations in `LoadOrderResolver.cs` — `WinnerRecordsOfType` and the plugin-scoped `RecordsIn` — now do the same. The group's own name (`GLOB`, `GMST`) still resolves to every arm.

**#449 — `to_file=` artifact rows carried scoped field values without the scoped-vs-winner note.** The text, json and dense renders all state which body a `plugins=`-scoped fields scan read its values from; the artifact carried the same values and said nothing, so the number came back out of the file later with nothing attached naming its source. The note now rides the manifest — line 1, what a re-entering caller reads first — from `JsonWire.ScopedFieldsNote`, the same composer and the same "any row read from a scoped body" test the inline renders use, so the four-way (`fields_source=`, `where_source=`) matrix cannot drift between the two.

**#448 — not folded: already fixed on `main`.** The refusal the issue asks for landed in 64a730b ("dense refuses a depth it cannot render, instead of dropping it") with three tests in `RecordsScanRefusalRepairTests.cs`. `RecordsTools.cs` refuses `format='dense'` with `project.depth > 1` by name on the scan lane, and the list lane's own dense refusal covers the other half. Verified on this branch: those tests pass. The issue can be closed as done rather than by this PR, which is why there is no `Closes #448` here.

New tests, each failing on the tree before its fix: `RecordsTypeArmTests` over a new `TypeArmWorld` (a GLOB group of all three arms and a GMST group of two) covers both arms of a group selecting their own records, a second abstract group, the plugin-scoped lane, and the group name still returning everything — 4 of its 5 failed before the fix. `RecordsArtifactScopedNoteTests` covers the note on the manifest, the display-pole arm, and the two cases owed no note — 2 of its 4 failed before the fix.

Closes #527
Closes #449

Stacked on #555 (`claude/skse-transport`), which this branch is based on and targets.

Build, `tier!=bridge` tests (1219 passed) and `ci-all` green.

### DIFF
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -1093,12 +1093,11 @@ public sealed class LoadOrderResolver : IDisposable
             }
             try
             {
-                foreach (var t in getterTypes)
-                    foreach (var rec in ov.EnumerateMajorRecords(t, throwIfUnknown: true))
-                        // A typed enumeration seeks the GRUP, which for an abstract group (Global, GameSetting) holds
-                        // every arm of it — so an arm's own type is re-checked here rather than trusted from the seek.
-                        if (t.IsInstanceOfType(rec) && s.Index.TryGetValue(rec.FormKey, out var e) && e.winner == i)   // this overlay's instance wins
-                            yield return (rec.FormKey, e.count, rec);
+                // RecordArms is the shared arm re-check: a typed enumeration seeks the GRUP, which for an abstract
+                // group (Global, GameSetting) holds every arm of it.
+                foreach (var rec in RecordArms.OfTypes(ov, getterTypes))
+                    if (s.Index.TryGetValue(rec.FormKey, out var e) && e.winner == i)   // this overlay's instance wins
+                        yield return (rec.FormKey, e.count, rec);
             }
             finally { (ov as IDisposable)?.Dispose(); }
         }
@@ -1128,11 +1127,9 @@ public sealed class LoadOrderResolver : IDisposable
             catch (Exception ex) { throw new PluginUnreadableException(_names[i], ex); }
             try
             {
-                // The arm re-check is the one WinnerRecordsOfType does, for the same reason: a typed enumeration over an
-                // abstract group's arm seeks the whole GRUP.
                 IEnumerable<IMajorRecordGetter> recs = getterTypes is null
                     ? ov.EnumerateMajorRecords()
-                    : getterTypes.SelectMany(t => ov.EnumerateMajorRecords(t, throwIfUnknown: true).Where(t.IsInstanceOfType));
+                    : RecordArms.OfTypes(ov, getterTypes);   // the shared arm re-check
                 foreach (var rec in recs)
                     if (s.Index.TryGetValue(rec.FormKey, out var e))
                         yield return (rec.FormKey, e.count, rec, _names[i]);   // _names[i] = this scoped plugin's filename (the source body)
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -1095,7 +1095,9 @@ public sealed class LoadOrderResolver : IDisposable
             {
                 foreach (var t in getterTypes)
                     foreach (var rec in ov.EnumerateMajorRecords(t, throwIfUnknown: true))
-                        if (s.Index.TryGetValue(rec.FormKey, out var e) && e.winner == i)   // this overlay's instance wins
+                        // A typed enumeration seeks the GRUP, which for an abstract group (Global, GameSetting) holds
+                        // every arm of it — so an arm's own type is re-checked here rather than trusted from the seek.
+                        if (t.IsInstanceOfType(rec) && s.Index.TryGetValue(rec.FormKey, out var e) && e.winner == i)   // this overlay's instance wins
                             yield return (rec.FormKey, e.count, rec);
             }
             finally { (ov as IDisposable)?.Dispose(); }
@@ -1126,9 +1128,11 @@ public sealed class LoadOrderResolver : IDisposable
             catch (Exception ex) { throw new PluginUnreadableException(_names[i], ex); }
             try
             {
+                // The arm re-check is the one WinnerRecordsOfType does, for the same reason: a typed enumeration over an
+                // abstract group's arm seeks the whole GRUP.
                 IEnumerable<IMajorRecordGetter> recs = getterTypes is null
                     ? ov.EnumerateMajorRecords()
-                    : getterTypes.SelectMany(t => ov.EnumerateMajorRecords(t, throwIfUnknown: true));
+                    : getterTypes.SelectMany(t => ov.EnumerateMajorRecords(t, throwIfUnknown: true).Where(t.IsInstanceOfType));
                 foreach (var rec in recs)
                     if (s.Index.TryGetValue(rec.FormKey, out var e))
                         yield return (rec.FormKey, e.count, rec, _names[i]);   // _names[i] = this scoped plugin's filename (the source body)
--- a/src/housecarl-mcp/Artifacts.cs
+++ b/src/housecarl-mcp/Artifacts.cs
@@ -184,7 +184,7 @@ internal static class Artifacts
         // The manifest stamps which tool wrote the artifact; see WriteResolve for why it names records.
         var (manifest, err) = writer.Save(path, ToolNames.Records, query, identity, schema, sort,
                                           q.Groups is not null ? q.Groups.Count : q.Total, q.Epoch ?? "",
-                                          OwnedChildNotes(annotated, annotatedUnioned));
+                                          CrossQueryNotes(q, fields, winnerFields, annotated, annotatedUnioned, levers));
         return err is not null ? (null, err) : (new SpillInfo(path, manifest!, reason), null);
     }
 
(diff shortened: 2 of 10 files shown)


src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -1083,9 +1085,11 @@ IMajorRecordGetter FetchBody(OverlaySession session, int overlayIdx, FormKey fk)
             catch (Exception ex) { throw new PluginUnreadableException(_names[i], ex); }
             try
             {
+                // The arm re-check is the one WinnerRecordsOfType does, for the same reason: a typed enumeration over an
+                // abstract group's arm seeks the whole GRUP.
                 IEnumerable<IMajorRecordGetter> recs = getterTypes is null
                     ? ov.EnumerateMajorRecords()
-                    : getterTypes.SelectMany(t => ov.EnumerateMajorRecords(t, throwIfUnknown: true));
+                    : getterTypes.SelectMany(t => ov.EnumerateMajorRecords(t, throwIfUnknown: true).Where(t.IsInstanceOfType));
```
Avick3110: The third lane with this exact widening is left unfixed: `ErrorCheck.OffOrderRecords` (`src/housecarl-core/ErrorCheck.cs:496-499`) still does `types.SelectMany(t => ov.EnumerateMajorRecords(t, throwIfUnknown: true))` with no arm re-check, and `SweepScope.Matches` deliberately refuses to re-test `Types` on the grounds that it "is applied at the stream" (`SweepScope.cs:52-53`) - which is now true only of the two lanes this PR fixed.

Concrete scenario: `housecarl_check_errors` over an off-order file with `type='GlobalShort'` (or any other arm, or any signature that resolves to more than one getter type) sweeps and counts the whole GRUP - the plugin lands in `examined` and in the per-plugin `bySource` tally on records outside the caller's filter, and with a multi-arm resolution (`type='GLOB'`) each record is walked once per arm, so any finding it produces is multiplied. Same class of silently-widened filter the PR is closing, on the one lane the active sweep does not share (`ErrorCheck.cs:266` goes through the now-fixed `RecordsIn`).

Worth either folding the same `.Where(t.IsInstanceOfType)` into `OffOrderRecords` or updating the `SweepScope.Matches` comment to say which lanes guarantee it.

src/housecarl-mcp/Artifacts.cs
```diff
@@ -191,6 +191,21 @@ public static (SpillInfo? Spill, string? Error) WriteCrossQuery(
...
+    /// <summary>The manifest notes a scan artifact carries: the scoped-vs-winner field-source note the three inline
+    /// transports state, then the owned-child clause. The artifact holds the same values the inline render would have
+    /// shown and is read later with no conversation attached, so the sentence saying WHOSE body those values came
+    /// from has to travel with them. The scoped test is the one the inline renders use, so the two cannot disagree
+    /// about when the note is owed.</summary>
+    static IReadOnlyList<string>? CrossQueryNotes(CrossQueryOutcome q, IReadOnlyList<string>? fields, bool winnerFields,
+                                                  IReadOnlyCollection<string> annotatedFields, LeverNames? levers)
+    {
+        var notes = new List<string>();
+        if (fields is { Count: > 0 } && q.Sources is { } sources && sources.Take(q.Keys.Count).Any(s => s is not null))
```
Avick3110: The PR body's guarantee is that the artifact uses "the same scoped test the inline renders use", but this is a fourth hand-copy of that predicate, not a shared one: `JsonWire.cs:989`, `JsonWire.cs:1085` and `ReadTools.cs:237` each spell `detail && q.Sources is { } ss && ss.Take(q.Keys.Count).Any(s => s is not null)` independently, and this line spells it again with `detail` inlined as `fields is { Count: > 0 }`. They agree today (verified character-for-character), so this is not a bug - but the copy is exactly the drift the note is meant to be immune to. One `internal static bool AnyScopedFieldRow(CrossQueryOutcome q, IReadOnlyList<string>? fields)` on `JsonWire`, called from all four, would make the guarantee structural rather than a claim.

Minor, same line: `CrossQueryNotes` tests `fields` without excluding the `q.Groups is not null` branch that `WriteCrossQuery` takes first. It is inert only because a `group_by=` outcome never fills `Keys` (`LoadOrderService.cs:4188`), so `Take(0).Any()` is false - an invariant held one call away, not by this method.

Avick3110:
Review fold — both findings folded, nothing refuted.

**Medium, the third lane.** Right, and the fix now lives in one place rather than a third copy: `RecordArms.OfTypes` in housecarl-core is the single typed enumeration, and `ErrorCheck.OffOrderRecords`, `LoadOrderResolver.WinnerRecordsOfType` and `RecordsIn` all go through it. `SweepScope.Matches` keeps declining to re-test `Types`, and its comment now names why that is safe: every stream a sweep reads narrows through `RecordArms`.

New test `AnArmTypeFilterOnTheOffOrderSweepExaminesThatArmAlone` — `check_errors` over an off-order file holding a GlobalFloat and a GlobalInt, scoped `type='GlobalShort'`, must examine nothing. Red before the fix (`BaseMastersSwept` came back `["Update.esm"]`, i.e. the whole GLOB group swept and tallied), green after. `TheArmTheOffOrderFileHoldsIsExamined` is the control, so the first assertion is a filter working rather than the lane never reaching the records. The changelog entry now says the fix covers `check_errors` with `type=` on both lanes.

**Low, the fourth hand-copy.** Folded as suggested: `JsonWire.AnyScopedFieldRow(q, fields)` is now the one predicate, called from `JsonWire.cs:989`, `JsonWire.cs:1085`, `ReadTools.cs:237` and `Artifacts.CrossQueryNotes`. It also answers false outright when `q.Groups is not null`, so the group_by branch is excluded structurally rather than resting on that outcome leaving `Keys` empty one call away.

Build clean, 1221 tests pass (`tier!=bridge`), ci-all 128/128.

Avick3110:
Blind review folded (2 of 2 real, 0 refuted). Queued for Aaron's review. Merge after #555; stacked follow-on: #568 (#526 #382).

PR #568, opened 2026-09-05
Say when an Add duplicates and when an in-place create grows the masters

### LINKED ISSUE #382: [Bug] in-place create grows the master header without the re-sort note its sibling lanes emit

### LINKED ISSUE #526: [Bug] Add appends a FormLink the list already carries, and the readback cannot be told from a clean add

Avick3110:
Two writes that answered without saying what they had done.

`verb=Add` on a list appended an element the list already carried, and nothing in the response or the read-back could tell that from a clean add — both print a longer list — so the same duplicate got repeated across a bulk run. The fix keeps Add's meaning (it still appends; no new verb, no new parameter) and makes it report: the engine's list Add now tests membership before appending and hands the caller a `duplicate: …` note, carried on that op's line in the text render and as `apply_note` in the json render. A dry run says it before anything is written, in the conditional voice the rest of that render uses. Every Add form is checked, `compose=` and `composes=` included — Mutagen's element classes (ContainerEntry, LeveledItemEntry, Condition, Effect) override `Equals` structurally, so `Contains` finds a freshly built identical element; a composes= batch counts its duplicates and points the remedy at Remove-by-index. A membership test that cannot be asked — an element type that compares by reference only — says so rather than reading as clean.

The in-place create lane grew the target's master header without the re-sort note its sibling lanes emit. `MasterGrowNote` had two call sites, the in-place edit and the in-place forward, and the create lane's success construction never computed a before-state. It does now, for both lanes that open an existing file — in place, and `into=` extending a patch the caller already enabled and sorted — and the service's in-place create block joins `outcome.Note` first, which #382 recorded as a required rider: without it the note this adds would be dropped by the very `JoinNotes` call the issue's measurement cleared. A fresh patch keeps no before-state and needs none: its whole header is new by construction.

Tests: twelve, in `WriteSilentAddAndMasterGrowTests`, driving the tools as a caller does. The in-place one asserts the sibling lanes' own sentence rather than the word "re-sort", which the mod-folder line already carried in the misleading form ("re-sort only if a winner changed") that this note exists to correct.

1231 tests (baseline 1219), 128/128 probes, build clean.

Closes #526
Closes #382

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -142,12 +142,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   the verb is unchanged — but a duplicating add and a clean one printed the same longer list, so neither the
   response nor the read-back could tell them apart, and a bulk run repeated the duplicate across every record.
   The op line now carries `duplicate: …`, and `format='json'` carries it as `apply_note`; a dry run says it before
-  anything is written. The check is on the plain-value form of `Add`, the one that can compare elements — a
-  `compose=`-built element is a fresh instance and is not compared.
-- **An in-place create that grows the target's master header now emits the same re-sort note the in-place edit and
-  forward lanes emit.** A new record whose link points into a plugin the target did not already master adds that
-  master, and the file will not load until the order is re-sorted; the create lane said nothing, so the only line
-  about re-sorting was the mod-folder line's "only if a winner changed", which is not this case.
+  anything is written. `compose=` and `composes=` elements are compared too — a duplicated condition row, container
+  entry or leveled-list entry is named the same way, counted over the batch. A list whose elements cannot be compared
+  by value says that instead of reading as clean.
+- **A create that grows an existing plugin's master header now emits the same re-sort note the in-place edit and
+  forward lanes emit — in place, and when extending a patch with `into=`.** A new record whose link points into a
+  plugin the file did not already master adds that master, and the file will not load until the order is re-sorted;
+  the create lane said nothing, so the only line about re-sorting was the mod-folder line's "only if a winner
+  changed", which is not this case.
 
 - **A `types=` filter naming one arm of an abstract record group — `GlobalShort`, `GlobalFloat`, `GameSettingInt` and
   every other Mutagen arm — now answers over that arm alone.** It returned the whole group instead: the plugin
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -143,13 +143,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
   response nor the read-back could tell them apart, and a bulk run repeated the duplicate across every record.
   The op line now carries `duplicate: …`, and `format='json'` carries it as `apply_note`; a dry run says it before
   anything is written. `compose=` and `composes=` elements are compared too — a duplicated condition row, container
-  entry or leveled-list entry is named the same way, counted over the batch. A list whose elements cannot be compared
-  by value says that instead of reading as clean.
-- **A create that grows an existing plugin's master header now emits the same re-sort note the in-place edit and
-  forward lanes emit — in place, and when extending a patch with `into=`.** A new record whose link points into a
-  plugin the file did not already master adds that master, and the file will not load until the order is re-sorted;
-  the create lane said nothing, so the only line about re-sorting was the mod-folder line's "only if a winner
-  changed", which is not this case.
+  entry or leveled-list entry is named the same way, counted over the batch. A `composes=` batch that repeats an
+  element of its own says that as its own clause, apart from what the list already carried.
+- **A write that grows an existing plugin's master header now emits the same re-sort note on every lane that opens
+  one — a create in place or with `into=`, and an apply or forward extending a patch with `into=`.** A record or
+  edit whose link points into a plugin the file did not already master adds that master, and the file will not load
+  until the order is re-sorted; those lanes said nothing, so the only line about re-sorting was the mod-folder line's
+  "only if a winner changed", which is not this case. A dry run of any of them says what the real write would add.
 
 - **A `types=` filter naming one arm of an abstract record group — `GlobalShort`, `GlobalFloat`, `GameSettingInt` and
   every other Mutagen arm — now answers over that arm alone.** It returned the whole group instead: the plugin
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -138,6 +138,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
 - **A `plugins=` name found neither in the active order nor on disk keeps its did-you-mean.** The refusal that says
   both halves now ends with the near-miss suggestion, so a typo or a guessed `.esp`/`.esl` extension names the
   plugin you meant.
+- **An `Add` that appends an element the list already carried now says so on that op's line.** It still appends —
+  the verb is unchanged — but a duplicating add and a clean one printed the same longer list, so neither the
+  response nor the read-back could tell them apart, and a bulk run repeated the duplicate across every record.
+  The op line now carries `duplicate: …`, and `format='json'` carries it as `apply_note`; a dry run says it before
+  anything is written. The check is on the plain-value form of `Add`, the one that can compare elements — a
+  `compose=`-built element is a fresh instance and is not compared.
+- **An in-place create that grows the target's master header now emits the same re-sort note the in-place edit and
+  forward lanes emit.** A new record whose link points into a plugin the target did not already master adds that
+  master, and the file will not load until the order is re-sorted; the create lane said nothing, so the only line
+  about re-sorting was the mod-folder line's "only if a winner changed", which is not this case.
 
 - **A `types=` filter naming one arm of an abstract record group — `GlobalShort`, `GlobalFloat`, `GameSettingInt` and
   every other Mutagen arm — now answers over that arm alone.** It returned the whole group instead: the plugin
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -1918,9 +1918,9 @@ public static class WriteEngine
     /// can actually re-send; it mirrors the same parameter on <c>CorpusRulebook.ValidateFromType</c>.</para>
     /// </summary>
     /// <returns>An apply-time note about what the write DID that the written file cannot express afterwards — today
-    /// only the duplicate-Add note (a list Add that appended an element the list already carried; the file shows a
-    /// longer list either way). Null when there is nothing to say, which is the common case. Callers that render an
-    /// op line carry it there; the rest ignore it.</returns>
+    /// only the list Add's membership answer (it appended an element the list already carried, or it could not
+    /// compare the element at all; the file shows a longer list either way). Null when the check ran and found
+    /// nothing, which is the common case. Callers that render an op line carry it there; the rest ignore it.</returns>
     public static string? ApplyVerb(object record, WriteRequest req, string pathSlot = "field_path")
     {
         object current = record;
@@ -2630,7 +2630,8 @@ public static class WriteEngine
         }
     }
 
-    /// <returns>The apply-time note <see cref="ApplyVerb"/> hands back — non-null only for a duplicate Add.</returns>
+    /// <returns>The apply-time note <see cref="ApplyVerb"/> hands back — non-null only for an Add whose membership
+    /// test found the element already there, or could not be asked.</returns>
     static string? ApplyListVerb(object parent, PropertyInfo prop, Type listIface, WriteRequest req)
     {
         // ARRAY-backed collection (a C# T[], e.g. Weather.CloudTextures / Weather.Clouds — fixed-size game
@@ -2674,30 +2675,51 @@ public static class WriteEngine
                 // struct-element list (modeled-struct elements) → build the new element FROM PARTS; coercible-element
                 // list → coerce the plain value. composes= appends MANY built elements in ONE op (each pre-flighted
                 // by ComposesLegality).
+                // EVERY Add form reports whether the list ALREADY carried what it appended. Add still appends — the
+                // verb's meaning is unchanged and no verb is added — but a duplicating add and a clean one otherwise
+                // render identically (both print a longer list), so the caller cannot tell one from the other and
+                // repeats it across a bulk run. Composed elements are checked too: Mutagen's Loqui element classes
+                // (ContainerEntry, LeveledItemEntry, Condition, Effect, …) override Equals structurally, so Contains
+                // finds a freshly built identical element. A type that compares by reference only cannot be asked,
+                // and that is its own reported outcome — never a clean answer.
                 if (req.Structs is { } addSpecs)
                 {
                     var addM = AddMethod(lt, elem);
-                    foreach (var s in addSpecs) addM.Invoke(list, new[] { BuildStruct(s) });
-                    break;
+                    int dup = 0, unknown = 0;
+                    foreach (var s in addSpecs)
+                    {
+                        var built = BuildStruct(s);
+                        switch (ListCarries(elem, list, built))
+                        {
+                            case Carried.Yes: dup++; break;
+                            case Carried.Unknown: unknown++; break;
+                        }
+                        addM.Invoke(list, new[] { built });
+                    }
+                    return ComposedAddNote(prop.Name, addSpecs.Count, dup, unknown);
                 }
                 if (req.Struct is not null)
                 {
-                    AddMethod(lt, elem).Invoke(list, new[] { BuildStruct(req.Struct) });
-                    break;
+                    var built = BuildStruct(req.Struct);
+                    var carriedStruct = ListCarries(elem, list, built);
+                    AddMethod(lt, elem).Invoke(list, new[] { built });
+                    return ComposedAddNote(prop.Name, 1,
+                        carriedStruct == Carried.Yes ? 1 : 0, carriedStruct == Carried.Unknown ? 1 : 0);
                 }
-                // A PLAIN-VALUE Add reports whether the list ALREADY carried this element. Add still appends — the
-                // verb's meaning is unchanged and no verb is added — but a duplicating add and a clean one otherwise
-                // render identically (both print a longer list), so the caller cannot tell one from the other and
-                // repeats it across a bulk run. Only the plain-value form is checked: a BUILT struct element is a
-                // fresh instance whose equality is the modeled type's, so a membership test there would answer no on
-                // every real duplicate and say nothing useful.
                 var addValue = Coerce(req.Value!, elem);
-                bool already = ListCarries(elem, list, addValue);
+                var already = ListCarries(elem, list, addValue);
                 AddMethod(lt, elem).Invoke(list, new[] { addValue });
-                if (already)
-                    return $"duplicate: '{req.Value}' was already in '{prop.Name}' — Add appends, so the list now "
-                           + "carries it twice; Remove it by value if you meant the list to hold one.";
-                break;
+                // CONDITIONAL voice, and no count: the same string renders under the dry-run header, where nothing was
+                // written and "Remove it" would delete the record's only copy, and Contains answers presence, not
+                // multiplicity — a list that already held two now holds three.
+                return already switch
+                {
+                    Carried.Yes =>
+                        $"duplicate: '{req.Value}' is already in '{prop.Name}', and Add appends rather than replacing "
+                        + "— once this write lands the list carries another copy, so Remove it by value if it should hold one.",
+                    Carried.Unknown => UncomparableNote(prop.Name),
+                    _ => null,
+                };
             case "SetAtIndex":
             {
                 // EXPECTED apply rejection (live length): pre-flight gates the index SHAPE (parseable non-negative int)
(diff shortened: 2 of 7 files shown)


src/housecarl-core/WriteEngine.cs
```diff
@@ -2532,8 +2538,23 @@ static void ApplyListVerb(object parent, PropertyInfo prop, Type listIface, Writ
...
+                // verb's meaning is unchanged and no verb is added — but a duplicating add and a clean one otherwise
+                // render identically (both print a longer list), so the caller cannot tell one from the other and
+                // repeats it across a bulk run. Only the plain-value form is checked: a BUILT struct element is a
+                // fresh instance whose equality is the modeled type's, so a membership test there would answer no on
+                // every real duplicate and say nothing useful.
+                var addValue = Coerce(req.Value!, elem);
+                bool already = ListCarries(elem, list, addValue);
+                AddMethod(lt, elem).Invoke(list, new[] { addValue });
+                if (already)
+                    return $"duplicate: '{req.Value}' was already in '{prop.Name}' — Add appends, so the list now "
```
Avick3110: The duplicate note is written in the past tense and its remedy assumes the append already happened, but the same string is rendered on a **dry run**, where nothing was written. `ADryRunAlsoSaysTheAddWouldDuplicate` exercises exactly that path, and the output reads:

```
DRY RUN - nothing written...
  Weapon 000803:HcDupMaster.esm  Keywords Add  -> would become ...  [duplicate: '000800:HcDupMaster.esm' was already in 'Keywords' - Add appends, so the list now carries it twice; Remove it by value if you meant the list to hold one.]
```

Both halves are wrong there: the list does not now carry it twice, and a caller who follows "Remove it by value" after a dry run deletes the record's only copy of the keyword. Every other clause in `RenderDryRun` switches to the conditional voice (`would become`, `would apply`) for exactly this reason. Suggest phrasing it neutrally - "the list already carries this element, and Add appends, so it would hold two" - or having the dry-run render restate it.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -322,13 +322,17 @@ static void AppendCompactReadback(StringBuilder sb, IReadOnlyList<WritePatchBuil
             // The per-op clause is the FILE's answer when the file gave one (LandedOnDisk), and is marked as the
             // applied edit's claim when it did not — the banner above says "re-read off the written file".
             var landed = ops.Where(op => op.Target == r.Target && (op.LandedOnDisk ?? op.Landed) is not null)
-                             .Select(op => $"{op.Label}: {op.LandedOnDisk ?? op.Landed}" + LandedProvenance(op))
+                             .Select(op => $"{op.Label}: {op.LandedOnDisk ?? op.Landed}" + LandedProvenance(op) + ApplyNote(op))
```
Avick3110: `ApplyNote(op)` is appended here **and** on the op line in `Render` (line 181), so a single response prints the same sentence twice for the same op. In-place writes force `readback: true`, so every in-place duplicate Add hits this; on the patch lane it happens whenever `readback=true` is passed. Concretely, one Add of a keyword the record already carries yields:

```
  Weapon 000803:...  Keywords Add  -> ...  [duplicate: ... carries it twice; Remove it by value ...]
read back off the written file:
  Weapon 000803:...; Keywords Add: ...  [duplicate: ... carries it twice; Remove it by value ...]
```

The note belongs in one place. `LandedProvenance` next to it is deliberately terse for the same reason.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2749,6 +2757,14 @@ static CreateOutcome CreateRecordsCore(
         }
         if (!string.Equals(patchMod.ModKey.FileName.String, fileName, StringComparison.OrdinalIgnoreCase))
             return CreateOutcome.Fail($"{(inPlace ? "in-place" : "patch")} ModKey '{patchMod.ModKey.FileName}' must match {(inPlace ? "the target filename" : "output filename")} '{fileName}'.");
+        // IN PLACE ONLY: the author's DECLARED masters, captured before any mutation, so Phase 5 can diff the re-opened
+        // header and emit the SAME re-sort note the in-place edit and forward lanes emit (MasterGrowNote). A new record
+        // whose FormLink points into a plugin the target did not already master grows the header, and the file will not
+        // load until the order is re-sorted. The patch lanes have no before-state to diff: a fresh patch's whole header
+        // is new by construction, and an extend already tells the caller the patch grew.
+        var mastersBefore = inPlace
```
Avick3110: The comment justifies skipping the patch lanes with "an extend already tells the caller the patch grew", but the extend render does not say that. `RenderCreate` prints `WriteSentences.Masters(o.Masters)` - the full master list, with no before/after delta and no re-sort guidance - so the extend case has the same failure the in-place case has: a user extends `MyPatch.esp` they already enabled and sorted, the new record's FormLink pulls in a master the patch did not have, and the plugin now refuses to load until the order is re-sorted, with nothing in the response saying so. An extend has a real before-state (`patchMod` is opened from the existing file on the branch just above), so the same `mastersBefore` capture would work for `extend` too. Worth either extending the capture to `inPlace || extend` or correcting the comment to say the extend gap is knowingly left open.

src/housecarl-mcp-tests/WriteSilentAddAndMasterGrowTests.cs
```diff
@@ -0,0 +1,139 @@
...
+using Xunit;
+
+namespace HousecarlMcpTests;
+
+/// <summary>
+/// Two writes that used to answer without saying what they had done: an Add that appended an element the list
+/// already carried (#526), and an in-place create that grew the target's master header without the re-sort note its
+/// sibling lanes emit (#382).
+///
+/// <para>The world is built per class and mutated by the in-place create, so it is this class's own: an in-place
```
Avick3110: "built per class" is not what happens: xUnit constructs the test class **once per test method**, so this constructor - including `CorpusGenerator.GenerateAll`, both plugin writes and the MO2 instance - runs five times, and `Dispose` tears it all down five times. `RuntimeFormIdTests` states the same shape correctly ("built per test"). Only `AnInPlaceCreateThatGrowsTheMasterHeaderSaysToReSort` actually mutates the world; the four Add tests are read-mostly and could share an `IClassFixture` world if the corpus generation is expensive. At minimum the comment should say "per test", since the stated reason (isolation from the in-place rewrite) is what the current shape happens to buy.

src/housecarl-core/WriteEngine.cs
```diff
@@ -2532,8 +2538,23 @@ static void ApplyListVerb(object parent, PropertyInfo prop, Type listIface, Writ
                     foreach (var s in addSpecs) addM.Invoke(list, new[] { BuildStruct(s) });
                     break;
                 }
-                AddMethod(lt, elem).Invoke(list,
-                    new[] { req.Struct is not null ? BuildStruct(req.Struct) : Coerce(req.Value!, elem) });
+                if (req.Struct is not null)
```
Avick3110, line 2702: **high** — the reason given for skipping the membership test on a composed element is not true of Mutagen's element types, and the skip is itself silent.

The comment below says a built element "is a fresh instance whose equality is the modeled type's, so a membership test there would answer no on every real duplicate". Measured against Mutagen.Bethesda.Skyrim 0.54.4 (the pinned version), the opposite holds: the Loqui-generated element classes override `Equals(object)` structurally, and `ICollection<T>.Contains` finds a freshly built, identically-populated element.

```
EqualsDeclaringType[ContainerEntry]   = ContainerEntry
EqualsDeclaringType[LeveledItemEntry] = LeveledItemEntry
EqualsDeclaringType[ConditionFloat]   = ConditionFloat
EqualsDeclaringType[Effect]           = Effect
ContainerEntry   list.Contains(fresh duplicate) = True
LeveledItemEntry list.Contains(fresh duplicate) = True
Condition        list.Contains(fresh duplicate) = True
```

(Two independently constructed elements with the same field values, added to an `ExtendedList<T>`, asked through the same closed `ICollection<T>` interface `ListCarries` uses.)

So the exclusion drops the check on exactly the families where a repeated bulk write bites hardest — a duplicated CTDA row, a duplicated container entry, a duplicated leveled entry — and #526's "the reporter wrote duplicate keywords across 126 records" scenario reproduces through `compose=` with nothing said. `req.Structs` (`composes=`, line 2535) is the same.

Two ways out, either is fine:

1. Run `ListCarries` for the built element too (it is already defensive — a comparer that throws answers no). If you do, note that a duplicate entry in an LVLI/LVLN is legitimate weighting, so the note's wording should suit that family rather than reading as "you made a mistake".
2. Keep the exclusion, but say it on the op line — "duplicate check skipped (composed element)". As it stands the caller cannot tell "checked, clean" from "not checked", which is the silently-degraded mode the cornerstone rules out and the one this PR exists to close.

Verified against the PR head; the four other new tests, the full suite (1224) and `ci-all` all pass, so this is a gap in coverage, not a regression.

src/housecarl-core/WriteEngine.cs
```diff
@@ -2610,6 +2631,20 @@ static void ApplyListVerb(object parent, PropertyInfo prop, Type listIface, Writ
...
+    /// CLOSED INTERFACE, so an explicit implementation still answers, and a failure to ask is answered NO — the
+    /// duplicate note is an extra sentence on a write that has to land either way, never a reason to fail one.</summary>
+    static bool ListCarries(Type elem, object list, object? value)
+    {
+        try
+        {
+            var contains = typeof(ICollection<>).MakeGenericType(elem).GetMethod("Contains", new[] { elem });
+            return contains?.Invoke(list, new[] { value }) is true;
+        }
+        catch { return false; }
```
Avick3110, line 2808: **low** — the check has three outcomes and the caller is shown two.

`Contains` can answer yes, answer no, or fail to be asked (the reflection lookup returning null, or a comparer that throws). The catch folds the third into the second, and a caller reading the op line then sees exactly what a verified-clean add looks like. Not asking is right — the write has to land either way — but by the same rule the note exists for, an unaskable check should say so rather than borrow the clean reading.

src/housecarl-core/WriteEngine.cs
```diff
@@ -2532,8 +2538,23 @@ static void ApplyListVerb(object parent, PropertyInfo prop, Type listIface, Writ
...
+                // render identically (both print a longer list), so the caller cannot tell one from the other and
+                // repeats it across a bulk run. Only the plain-value form is checked: a BUILT struct element is a
+                // fresh instance whose equality is the modeled type's, so a membership test there would answer no on
+                // every real duplicate and say nothing useful.
+                var addValue = Coerce(req.Value!, elem);
+                bool already = ListCarries(elem, list, addValue);
+                AddMethod(lt, elem).Invoke(list, new[] { addValue });
+                if (already)
+                    return $"duplicate: '{req.Value}' was already in '{prop.Name}' — Add appends, so the list now "
+                           + "carries it twice; Remove it by value if you meant the list to hold one.";
```
Avick3110: **low** — "carries it twice" states a count the check did not establish.

`Contains` answers presence, not multiplicity. A list that already carried the element twice now carries it three times, and the note still says twice. The rest of the sentence does not need the number: "was already in 'X' — Add appends, so this is another copy" says the same thing without asserting a count nothing measured.

Avick3110:
Review fold. All seven findings were real; all seven are folded, nothing refuted. 1231 tests (was 1224), 128/128 probes, build clean.

**1 (high) — the `compose=` exclusion rests on a false premise.** Confirmed independently against the pinned Mutagen.Bethesda.Skyrim 0.54.4, with two identically-populated instances built separately and asked through the same closed `ICollection<T>` interface `ListCarries` uses:

```
EqDecl[ContainerEntry]   = ContainerEntry     ContainerEntry   Contains(fresh dup) = True
EqDecl[LeveledItemEntry] = LeveledItemEntry   LeveledItemEntry Contains(fresh dup) = True
EqDecl[ConditionFloat]   = ConditionFloat     Condition        Contains(fresh dup) = True
EqDecl[Effect]           = Effect             Effect           Contains(fresh dup) = True
```

The check now runs on every Add form — `compose=`, `composes=`, and the plain value. A `composes=` batch counts over the batch ("1 of the 2 composed elements are already in 'Entries'"), and its remedy is Remove **by index**, because Remove-by-value takes a plain value a composed element does not have. The wording reports rather than scolds, since a repeated leveled-list entry is legitimate weighting. Two tests: a composed Add that duplicates, and one that does not.

**5 (low) — `ListCarries` folded "could not ask" into "not a duplicate".** Now a three-valued answer (`No` / `Yes` / `Unknown`), and `Unknown` gets its own sentence: "duplicate not checked: houseCARL cannot compare 'X' elements by value, so it cannot say whether this Add appends a copy the list already carried." Unknown covers the reflection lookup failing, a comparer throwing, and — new — a value whose runtime type declares no `Equals(object)` of its own, which is the case your finding 1 named. Asked of the value's runtime type, not the list's element type: an element type that is an interface (`IFormLinkGetter<T>`) declares no `Equals` at all while the value behind it does.

One honest gap: that arm has no end-to-end test, because I could not find a reachable case. Across every concrete public class in the Skyrim assembly the 1742 without a structural `Equals` are translation/mask internals, not element types; every element family I could reach overrides it. The arm is defensive, and it is now reported rather than silent, which is what the finding asked for.

**2 (medium) — past tense under the dry-run header.** The note is now conditional and holds in both renders: "duplicate: 'X' is already in 'Keywords', and Add appends rather than replacing — once this write lands the list carries another copy, so Remove it by value if it should hold one." A test pins the dry run to `is already in` + `once this write lands`, refuses `was already in` and `the list now`, and asserts the dry-run note is the **same string** as the applied one, so the two cannot drift.

**6 (low) — "carries it twice" asserts a multiplicity never measured.** Dropped; "another copy" replaces it, tested.

**3 (low) — the note printed twice.** Removed from `AppendCompactReadback`'s per-op clause; it stays on the op line. Tested on the in-place lane, which is where it actually bites (readback is forced there — `readback: true` on the patch lane takes the *full* dump, which never carried it).

**4 (low) — the "an extend already tells the caller the patch grew" premise.** Took your first option and closed the gap rather than documenting it: `mastersBefore` is now captured for `inPlace || extend`, both lanes that open an existing file. A fresh patch still keeps none and the comment now says why (whole header new by construction). Tested: a fresh patch says nothing about a grown master; extending it with a record whose FormLink pulls in `HcDupMaster.esm` gets the sibling lanes' sentence. Note the create core has no dry-run path, so there is no tense problem here.

**7 (low) — "built per class".** Corrected to per test method. I measured the sharing question rather than assuming: the class runs 3 s for one test and 5 s for five, so a construction costs about 0.5 s, and an `IClassFixture` would save under 2 s on a 1231-test suite while coupling four tests to a world one of them rewrites. Isolation kept, and the comment now records the measurement instead of an untested reason.

Changelog updated for the broadened check and the extend lane.

Avick3110:
Blind review folded (7 of 7 real, 0 refuted; the compose= exclusion was wrong and is gone). Queued for Aaron's review. Merge after #566.

Avick3110:
Removed the unreachable "could not check for a duplicate" arm on list Add, per Aaron's ruling (dev/DECISIONS.md, 2026-09-06): deleted, not kept as defensive code. `ListCarries` is now two-valued (returns bool) instead of the three-valued `Carried` enum; the `Unknown` outcome, its `UncomparableNote` sentence, the `HasStructuralEquality` pre-check that only existed to reach it, and the unknown counting in `ComposedAddNote` are all gone. No test or comment existed solely for that arm. Build clean, 1231 tests pass, ci-all 128/128.

src/housecarl-core/WriteEngine.cs
```diff
@@ -2546,15 +2553,41 @@ static void ApplyListVerb(object parent, PropertyInfo prop, Type listIface, Writ
...
                 if (req.Structs is { } addSpecs)
                 {
                     var addM = AddMethod(lt, elem);
-                    foreach (var s in addSpecs) addM.Invoke(list, new[] { BuildStruct(s) });
-                    break;
+                    int dup = 0;
+                    foreach (var s in addSpecs)
+                    {
+                        var built = BuildStruct(s);
+                        if (ListCarries(elem, list, built)) dup++;
```
Avick3110: **composes= counts intra-batch repeats as duplicates *of the list*.** `ListCarries` is asked after the previous element of the same batch has already been appended, so a repeat within the batch is attributed to the file's prior state.

Scenario: `composes=[Entry(9), Entry(9)]` against a leveled list that does not carry Entry(9). Iteration 1 appends it; iteration 2's `Contains` finds the element the loop just added; the op line reads `1 of the 2 composed elements are already in 'Entries'` — false about what the list carried before this write — and its remedy, "Remove by index if it should hold one of each", points at undoing deliberate leveled-list weighting, which is exactly the family the note's wording was written to suit.

Test membership against a snapshot of the list taken before the loop, or count the batch's own repeats separately and word them as such.

plugin/CHANGELOG.md
```diff
@@ -85,6 +85,18 @@ saying it sets an expectation their install may contradict. Say what is known, a
 - **A `plugins=` name found neither in the active order nor on disk keeps its did-you-mean.** The refusal that says
   both halves now ends with the near-miss suggestion, so a typo or a guessed `.esp`/`.esl` extension names the
   plugin you meant.
+- **An `Add` that appends an element the list already carried now says so on that op's line.** It still appends —
+  the verb is unchanged — but a duplicating add and a clean one printed the same longer list, so neither the
+  response nor the read-back could tell them apart, and a bulk run repeated the duplicate across every record.
+  The op line now carries `duplicate: …`, and `format='json'` carries it as `apply_note`; a dry run says it before
+  anything is written. `compose=` and `composes=` elements are compared too — a duplicated condition row, container
+  entry or leveled-list entry is named the same way, counted over the batch. A list whose elements cannot be compared
+  by value says that instead of reading as clean.
```
Avick3110: **This promises an outcome the code no longer has.** `b36b47f7` dropped the `Carried.Unknown` arm, so `ListCarries` now returns plain `false` for an element type that cannot answer by value — which renders exactly like a verified-clean add, the thing the removed arm existed to prevent.

Scenario: a caller reads this line, Adds an element to a list whose element type compares by reference, sees a clean op line, and treats it as "checked, no duplicate" when no membership answer was possible.

The same stale claim is in two code docs: `ApplyVerb`'s `<returns>` (`WriteEngine.cs:1799`, "or it could not compare the element at all") and `OpResult.ApplyNote` (`WritePatchBuilder.cs:94`, "or its elements could not be compared"). If the arm was right to drop as unreachable, all three should say so; if it is reachable, the arm is the fix.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2769,6 +2777,15 @@ static CreateOutcome CreateRecordsCore(
         }
         if (!string.Equals(patchMod.ModKey.FileName.String, fileName, StringComparison.OrdinalIgnoreCase))
             return CreateOutcome.Fail($"{(inPlace ? "in-place" : "patch")} ModKey '{patchMod.ModKey.FileName}' must match {(inPlace ? "the target filename" : "output filename")} '{fileName}'.");
+        // EVERY lane that OPENED AN EXISTING FILE: the author's DECLARED masters, captured before any mutation, so
+        // Phase 5 can diff the re-opened header and emit the SAME re-sort note the in-place edit and forward lanes
+        // emit (MasterGrowNote). A new record whose FormLink points into a plugin the file did not already master
+        // grows the header, and the file will not load until the order is re-sorted — true of an extended patch the
+        // caller already enabled and sorted, not only of an in-place target. A FRESH patch has no before-state and
+        // needs none: its whole header is new by construction and the render lists it in full.
+        var mastersBefore = inPlace || extend
```
Avick3110, line 2861: **The `extend` rationale in this comment covers two more lanes that stay silent.** `ApplyCore` (line 617) and `ForwardRecordsCore` (line 2314) also open an existing patch under `into=` and can pull a plugin into its header, and neither captures a before-state nor emits `MasterGrowNote` — so the fix lands on one of the three extend lanes.

Scenario: `housecarl_apply into="MyPatch.esp"` on a patch the caller already enabled and sorted, with an op whose value is a FormID in `Dawnguard.esm` that the patch did not master. The header grows, the response prints the full master list with nothing marking which entry is new, and the plugin will not load until the order is re-sorted — the exact case this hunk argues is worth saying for create.

Not necessarily this PR's job, but worth an issue if it is deferred.

src/housecarl-core/WriteEngine.cs
```diff
@@ -2630,8 +2663,32 @@ static void ApplyListVerb(object parent, PropertyInfo prop, Type listIface, Writ
...
+    }
+
+    /// <summary>Does the list already carry this element? Asked through <c>ICollection&lt;T&gt;.Contains</c> on the
+    /// CLOSED INTERFACE, so an explicit implementation still answers — never a reason to fail the write, which has
+    /// to land either way.</summary>
+    static bool ListCarries(Type elem, object list, object? value)
+    {
+        try
+        {
+            var contains = typeof(ICollection<>).MakeGenericType(elem).GetMethod("Contains", new[] { elem });
```
Avick3110: **The `MethodInfo` is resolved per element, and the check makes Add linear.** `MakeGenericType` + `GetMethod` run on every call — including once per element inside the `composes=` loop — and `Contains` is itself a linear scan with structural `Equals`, so Add went from O(1) to O(n) per element and a composed batch to O(n·m).

Scenario: a `bulk_apply` building a FormList of a few thousand entries, one Add per op — each op scans the list it just grew (~12M FormLink comparisons over 5000 Adds); a `composes=` of several hundred `LeveledItemEntry` rows does the same with a deep structural compare per element.

Hoisting the `MethodInfo` out of the loop is the cheap half; if the scan cost bites, the composes= path could build one comparison set up front.

Avick3110:
Second review fold. Four findings, all four real, all four folded, nothing refuted. 1335 tests (1331 before this fold; the count moved because the base branch did), 128/128 probes, build clean.

**1 — `composes=` counted intra-batch repeats as duplicates of the list.** Real, and the wording made it worse: `ListCarries` ran mid-loop, so the second `Entry(9)` of `composes=[Entry(9), Entry(9)]` found the element the loop had just appended and the op line claimed the *file* already carried it, with a remedy pointing at undoing the caller's own weighting.

Took the second option you named, because the two are different facts and folding either into the other lies about one of them: every element is now built and asked *before* any is appended (so the "already in 'X'" count is measured against the list's real before-state), and a repeat within the batch is counted separately and gets its own clause — "1 of the 2 composed elements is a repeat of another element in this same op". Both clauses join into one sentence when both apply. Silence would have been the other failure: a `composes=` batch that duplicates itself creates two copies, which is exactly #526.

Test: `AComposesBatchThatRepeatsItsOwnElementSaysSoSeparately` — two identical entries against a list carrying neither, asserting the repeat clause is present and `already in 'Entries'` is absent. The existing batch test (one duplicate of the list, one new) still pins the other half.

**2 — the changelog and two code docs promised the deleted unknown arm.** Real, all three. Fixed to describe what the code does now: the changelog line about "a list whose elements cannot be compared by value" is gone (replaced by the batch-repeat clause), and `ApplyVerb`'s `<returns>`, `ApplyListVerb`'s `<returns>` and `OpResult.ApplyNote` no longer mention a comparison that could not be made. The arm was right to drop per Aaron's ruling (dev/DECISIONS.md, 2026-09-06); the docs were the leftover.

**3 — `ApplyCore` and `ForwardRecordsCore` extend under `into=` and stay silent.** Real, and folded rather than deferred to an issue: all three extend lanes open a file the caller already enabled and sorted, and the asymmetry (create says it, apply and forward do not) is worse than either uniform answer. Both now capture `mastersBefore` on the extend arm and set `Note = MasterGrowNote(...)` on the success outcome; the plumbing was already there (`PatchOutcome.Note`, `ForwardOutcome.Note`, and both renders print it) and unused on those lanes.

Their dry runs say it too, predictively. That put a third and fourth copy of the "would ADD" sentence in the file, so it is now one helper, `MasterGrowWouldNote`, which the two in-place dry runs also call — same string, no behaviour change there.

Tests: `AnApplyExtendingAPatchThatGrowsItsMasterHeaderSaysToReSort`, `AForwardExtendingAPatchThatGrowsItsMasterHeaderSaysToReSort`, and `ADryRunOfAnExtendSaysWhatTheRealWriteWouldAdd`, each starting from a fresh patch that references nothing so the extend is the only thing that can pull a master in.

**4 — the `MethodInfo` was resolved per element.** Real. `ICollection<T>.Contains` is memoised per element type in a `ConcurrentDictionary`, so `MakeGenericType` + `GetMethod` run once per element type for the process, not once per element and once per op.

The other half — that `Contains` itself is a linear scan — I have folded only the cheap part, deliberately: the scan *is* the check #526 asked for, and dropping it on large lists would put back the silent mode this PR exists to close. A comparison-set built up front would help only the `composes=` path (the bulk-apply case is one Add per op, each op its own call, so there is nothing to amortise across), and it would need its own equality story rather than borrowing `Contains`. Flagging rather than deciding: if the scan on a several-thousand-entry FormList turns out to bite, it wants a measurement and its own issue, not a guess here.

PR #569, opened 2026-09-05
Gather winner bodies a plugin at a time on where_source=winner

### LINKED ISSUE #251: [GAP] cross_plugin_query where_source=winner re-fetches each winner via untyped whole-overlay enumeration — O(candidates × overlay) for broad scopes

Avick3110:
`where_source=winner` decides a scan's match on the live load-order winner rather than on the body the scan streams. It got that winner with `view.GetRecord(session, winnerPlugin, fk)`, which walks the winner plugin's whole record list to find one record. For N candidates that is N walks of a plugin, so the cost grew with the candidate count — on the feature's headline use, a broad post-patch audit, that is minutes of nothing but re-reading, and it is the exact anti-pattern the resolver's own comment thirty lines away warns about.

The candidate keys are settled by index-only filters (`formids=` set intersection, `conflicts_only=`, `defined_in=`), so the scan collects them off the record stream it is already walking and gathers their winner bodies **by plugin** rather than by record: `LoadOrderResolver.CollectRecords` walks one plugin's overlay once and picks every wanted key out of it, stopping as soon as it has them all.

**It gathers one CHUNK of candidates at a time — 10,000 — and drains it before reading on.** Gathering the whole candidate set first would hold a key and a pinned overlay getter per candidate for the scan's whole length, which on `plugins=["Skyrim.esm"]` with no `types=` is every record in the order; the chunk is the memory bound, and the measurement below is what it is worth. Two more things keep the work down: the chunk is filled from the **one** scoped stream the scan already runs, so the scope is walked once rather than twice, and a candidate the **scoped plugin itself wins** is never gathered — the streamed body already IS the winner's, which on the headline shape is most of them. With a `types=` scope in hand each walk is typed as well, and falls through to the flat one for whatever it did not find, per key.

Nothing is held past the scan's own session: the bodies are backed by the overlays it already has open. A winner plugin that will not open leaves its candidates absent and hands back the **reason**, so a file another program is holding open reports that — with the IOException under it, and the plugin named once as a coverage gap — instead of every one of its records reading as index staleness. A record the order gives no winner at all is a clean non-match again, not an unscannable row naming a winner `'?'` that was never fetched.

## Measured

A synthetic order built for this: one master and nine plugins each overriding a 2,000-record slice of it (step 1,800, so the slices overlap and the winners spread across all ten). Three builds against the same bytes — `main`'s per-record fetch, the gather-all this PR first proposed, and the chunked gather it now has. `peak heap` is the highest sampled managed heap during the scan, with gen0 forced small so the sample tracks the live set.

| scan (19,200-record master) | per-record | gather-all | chunked |
|---|---|---|---|
| every record the master touches, `where_source=winner` | 34.0 s | 0.5 s | **0.4 s** |
| the same scan, `where_source` default | 0.1 s | 0.2 s | 0.2 s |
| the nine plugins as the scope, `where_source=winner` | 10.3 s | 0.4 s | **0.2 s** |
| the same scan, `where_source` default | 0.1 s | 0.2 s | 0.2 s |

So the winner-source penalty is gone and the two sources cost the same. The second scope walk the gather-all paid shows up in the 0.5 → 0.4 and 0.4 → 0.2.

The memory bound wants a bigger order, so the same shape at a **200,000-record master** (two samples each):

| 200k master audit | gather-all | chunked |
|---|---|---|
| `where_source=winner` | 2.4 s / 2.5 s, **417.7 MB / 403.4 MB** | 1.7 s / 1.8 s, **214.4 MB / 198.0 MB** |
| `where_source` default (no gather at all) | 1.2 s, 199.3 MB | 1.2 s, 198.9 MB |

The default-source arm is the floor. Gather-all sits ~210 MB above it — about 1.05 KB per candidate, which extrapolates to roughly a gigabyte on the ~1M-record untyped `Skyrim.esm` scope the review named. The chunked gather sits **on** the floor: its live set is one chunk, so at 200k candidates it is the same peak as a scan that fetches no winner bodies, and it does not move with the candidate count.

**The bound, stated:** at most 10,000 buffered rows and at most 10,000 gathered winner bodies are live at once, whatever the scan considers. The trade is that a winner plugin holding candidates in several chunks is enumerated once per chunk; the self-won skip is what keeps that small, since on a master audit the only records gathered at all are the ones something overrides.

## Tests

`WinnerSourceScanTests` is the parity fixture the issue asked for: a master defining 30 weapons, one plugin overriding the first ten and another overriding an overlapping middle ten, so five records win in the first plugin, ten in the second and fifteen in the master itself. A gather that took the first declarer, the scoped body, or only the plugins above the scope answers a different set on each of those arms. Eleven tests pin all three partitions, the de-dup under a multi-plugin scope, the type-scoped walk, a winner plugin another program is holding open, one that opens but faults part-way through its record walk, a plugin outside the order leaving the gather empty rather than throwing, and a typed gather handed a type that does not cover every wanted key.

1416 tests and 128/128 probes pass.

Stacked on #553 (`claude/cell-union`) — that branch's typed seek is in the diff base, and it does not by itself close this: the winner-source call site passed no type, and a typed seek would only be a constant factor over the same per-candidate walk, which is what the issue calls the stopgap.

Closes #251

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -181,7 +181,9 @@ saying it sets an expectation their install may contradict. Say what is known, a
 - **A `where_source=winner` scan now says WHY a winner plugin did not give up its record.** A plugin another
   program is holding open reports that, with the error underneath it and the plugin named once as a coverage gap,
   instead of every one of its records reading as a winner the index named that did not re-resolve — a different
-  problem, and not the one the user can act on.
+  problem, and not the one the user can act on. A plugin that opens but whose records cannot be read to the end —
+  it changed on disk since the load order was indexed — reports that instead, and is not described as held open by
+  a program that is holding nothing. Either way the gap is one plugin's: the rest of the scan still answers.
 
 - **Reading a cell's `Persistent`/`Temporary` — or a topic's `Responses`, or a worldspace's `SubCells` — now
   states the additive union the game assembles, not just the read body's own list.** These children are declared
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -169,6 +169,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   None-property diagnostics read it), and the list's own header keeps its true count. Under `format='json'` a row
   is one entry carrying its folded leaves as `cells`, each with its own value, `link` and count — the values are
   never handed over as a sentence to parse.
+- **A broad `where_source=winner` scan is no longer slower the more records it considers.** Deciding the match on
+  the live winner used to re-read each candidate's winner by walking that plugin's whole record list, once per
+  candidate, so a post-patch audit over a large master could spend minutes doing nothing but re-reading. The
+  winner bodies are now gathered a plugin at a time — one pass per winner plugin however many of its records are
+  wanted — so the cost is bounded by the load order instead of by the number of candidates. On a synthetic order
+  of one 19.2k-record master and nine plugins, an audit of every record the master touches went from 165 s to
+  51 s, which is what the same scan costs with `where_source` left at its default. Same matches, same rows.
+
 - **Reading a cell's `Persistent`/`Temporary` — or a topic's `Responses`, or a worldspace's `SubCells` — now
   states the additive union the game assembles, not just the read body's own list.** These children are declared
   per plugin and the engine assembles a parent's from every plugin that declares any, so a cell whose winner is
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -171,11 +171,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
   never handed over as a sentence to parse.
 - **A broad `where_source=winner` scan is no longer slower the more records it considers.** Deciding the match on
   the live winner used to re-read each candidate's winner by walking that plugin's whole record list, once per
-  candidate, so a post-patch audit over a large master could spend minutes doing nothing but re-reading. The
-  winner bodies are now gathered a plugin at a time — one pass per winner plugin however many of its records are
-  wanted — so the cost is bounded by the load order instead of by the number of candidates. On a synthetic order
-  of one 19.2k-record master and nine plugins, an audit of every record the master touches went from 165 s to
-  51 s, which is what the same scan costs with `where_source` left at its default. Same matches, same rows.
+  candidate, so a post-patch audit over a large master could spend minutes doing nothing but re-reading. The winner
+  bodies are now gathered a chunk of candidates at a time — one pass per winner plugin in the chunk — and a
+  candidate the scoped plugin itself wins is not re-read at all, since the body the scan already streamed IS the
+  winner's. So the cost is bounded by the load order instead of by the number of candidates, and the scan holds one
+  chunk of bodies at a time rather than one per candidate. On a synthetic order of one 19,200-record master and
+  nine plugins, an audit of every record the master touches went from 34 s to 0.4 s, against the 0.2 s the same
+  scan costs with `where_source` left at its default. Same matches, same rows.
+- **A `where_source=winner` scan now says WHY a winner plugin did not give up its record.** A plugin another
+  program is holding open reports that, with the error underneath it and the plugin named once as a coverage gap,
+  instead of every one of its records reading as a winner the index named that did not re-resolve — a different
+  problem, and not the one the user can act on.
 
 - **Reading a cell's `Persistent`/`Temporary` — or a topic's `Responses`, or a worldspace's `SubCells` — now
   states the additive union the game assembles, not just the read body's own list.** These children are declared
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -64,14 +64,31 @@ public readonly record struct RecordStatus(
 /// <summary>A plugin in the load order could not be opened during a scan, although it opened when the index was
 /// built — it was moved, or another program (xEdit, MO2, the running game) is holding it. Thrown from the
 /// plugin-scoped record stream so a scan reports the plugin it could not read rather than skipping it silently.</summary>
-public sealed class PluginUnreadableException : Exception
+public class PluginUnreadableException : Exception
 {
     public string PluginName { get; }
     public PluginUnreadableException(string pluginName, Exception inner)
-        : base($"could not read '{pluginName}': it opened when the load order was indexed but not now — another " +
+        : this(pluginName, inner,
+               $"could not read '{pluginName}': it opened when the load order was indexed but not now — another " +
                "program is probably holding it open. Close that program and run this again. " +
-               $"({inner.GetType().Name}: {inner.Message})", inner)
-        => PluginName = pluginName;
+               $"({inner.GetType().Name}: {inner.Message})") { }
+
+    protected PluginUnreadableException(string pluginName, Exception inner, string message)
+        : base(message, inner) => PluginName = pluginName;
+}
+
+/// <summary>A plugin OPENED but its records could not be walked to the end — the file changed under the index, or
+/// Mutagen cannot step over something in it. A different fault from <see cref="PluginUnreadableException"/> and
+/// worded as one: telling the user to close a program that is holding nothing would send them after the wrong
+/// thing. Reported the same way, as a whole-plugin coverage gap, because a walk that died part-way says nothing
+/// about the records after it.</summary>
+public sealed class PluginUnscannableException : PluginUnreadableException
+{
+    public PluginUnscannableException(string pluginName, Exception inner)
+        : base(pluginName, inner,
+               $"could not finish reading '{pluginName}': it opened, but its records could not be walked to the end " +
+               "— the file has probably changed since the load order was indexed. Run this again, and check the " +
+               $"plugin in xEdit if it repeats. ({inner.GetType().Name}: {inner.Message})") { }
 }
 
 /// <summary>A BASELINE master (Skyrim.esm / Update.esm) is active but cannot be opened. Every written plugin
@@ -1054,7 +1071,9 @@ public sealed class LoadOrderResolver : IDisposable
     /// them cost N walks (#251), while this costs one walk however many are wanted, and stops as soon as it has them
     /// all. <paramref name="getterTypes"/> narrows the walk to the GRUPs those types live in when the caller knows
     /// them; anything the typed walk did not produce falls through to the flat one, per KEY, for the same reason
-    /// <see cref="SeekBody"/> does. A key this plugin does not hold is simply absent from the sink.</summary>
+    /// <see cref="SeekBody"/> does. A key this plugin does not hold is simply absent from the sink. Throws
+    /// <see cref="PluginUnreadableException"/> on a plugin that opened at index-build time but cannot be opened now;
+    /// a fault from the WALK is left as it came, so a caller does not report it as a held-open file.</summary>
     public void CollectRecords(OverlaySession session, string pluginName, IReadOnlyCollection<FormKey> wanted,
                                IReadOnlyList<Type>? getterTypes, IDictionary<FormKey, IMajorRecordGetter> sink)
         => CollectRecords(session, pluginName, wanted, getterTypes, sink, _snap);
@@ -1065,7 +1084,11 @@ public sealed class LoadOrderResolver : IDisposable
         if (wanted.Count == 0) return;
         if (!_nameToIdx.TryGetValue(pluginName, out int idx)) return;
         if (s.Excluded.Contains(idx)) return;                              // excluded plugin — never re-enumerate it (would re-throw)
-        var ov = session.Overlay(idx);
+        ISkyrimModGetter ov;
+        // Only the OPEN is named as unopenable, exactly as RecordsIn and WinnerRecordsOfType name it. A fault from
+        // the walk BELOW is a different problem, and calling it a held-open file would send the user after nothing.
+        try { ov = session.Overlay(idx); }
+        catch (Exception ex) { throw new PluginUnreadableException(_names[idx], ex); }
         var want = wanted as HashSet<FormKey> ?? new HashSet<FormKey>(wanted);
         int got = 0;
         foreach (var fk in want) if (sink.ContainsKey(fk)) got++;          // already in hand from an earlier call
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -1053,7 +1053,7 @@ public sealed class LoadOrderResolver : IDisposable
     /// the caller wants many records from the same plugin: <see cref="GetRecord"/> costs one walk per record, so N of
     /// them cost N walks (#251), while this costs one walk however many are wanted, and stops as soon as it has them
     /// all. <paramref name="getterTypes"/> narrows the walk to the GRUPs those types live in when the caller knows
-    /// them; a typed walk that finds none of the wanted keys falls back to the flat one for the same reason
+    /// them; anything the typed walk did not produce falls through to the flat one, per KEY, for the same reason
     /// <see cref="SeekBody"/> does. A key this plugin does not hold is simply absent from the sink.</summary>
     public void CollectRecords(OverlaySession session, string pluginName, IReadOnlyCollection<FormKey> wanted,
                                IReadOnlyList<Type>? getterTypes, IDictionary<FormKey, IMajorRecordGetter> sink)
@@ -1068,6 +1068,8 @@ public sealed class LoadOrderResolver : IDisposable
         var ov = session.Overlay(idx);
         var want = wanted as HashSet<FormKey> ?? new HashSet<FormKey>(wanted);
         int got = 0;
+        foreach (var fk in want) if (sink.ContainsKey(fk)) got++;          // already in hand from an earlier call
+        if (got == want.Count) return;
         if (getterTypes is { Count: > 0 })
         {
             foreach (var t in getterTypes)
@@ -1077,7 +1079,11 @@ public sealed class LoadOrderResolver : IDisposable
                         sink[rec.FormKey] = rec;
                         if (++got == want.Count) return;
                     }
-            if (got > 0) return;                                           // the typed walk routed; what it did not find, this plugin does not hold
+            // Fall through for whatever is STILL missing, per key. "The typed walk routed something, so what it
+            // did not find is not here" is the inference SeekBody's own doc comment warns against: with two types
+            // where Mutagen routes one and silently routes nothing for the other, the routed type's hits would
+            // otherwise declare the unrouted type's records absent. The flat pass costs the same as SeekBody's
+            // fallback on a miss, and it cannot answer wrong.
         }
         foreach (var rec in ov.EnumerateMajorRecords())
             if (want.Contains(rec.FormKey) && !sink.ContainsKey(rec.FormKey))
--- a/src/housecarl-core/LoadOrderResolver.cs
+++ b/src/housecarl-core/LoadOrderResolver.cs
@@ -904,6 +904,12 @@ public sealed class LoadOrderResolver : IDisposable
         public IMajorRecordGetter? GetRecord(OverlaySession session, string pluginName, FormKey fk, Type? getterType = null)
             => _r.GetRecord(session, pluginName, fk, _s, getterType);
 
+        /// <summary>Many records from ONE plugin in ONE enumeration (<see cref="LoadOrderResolver.CollectRecords"/>),
+        /// judged against THIS view's build.</summary>
+        public void CollectRecords(OverlaySession session, string pluginName, IReadOnlyCollection<FormKey> wanted,
+                                   IReadOnlyList<Type>? getterTypes, IDictionary<FormKey, IMajorRecordGetter> sink)
+            => _r.CollectRecords(session, pluginName, wanted, getterTypes, sink, _s);
+
         /// <summary>The master filenames a plugin DECLARES in its header (the master table), in declared order — opens
         /// the overlay, reads the header, disposes — the header parses without enumerating records. Throws
         /// on a name not in the order or excluded this build. The integrity sweep diffs this against the masters a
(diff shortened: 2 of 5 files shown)


src/housecarl-mcp-tests/WinnerSourceScanTests.cs
```diff
@@ -0,0 +1,205 @@
...
+        using var doc = JsonDocument.Parse(json);
+        Assert.Equal(Expected(5, 15),
+                     doc.RootElement.GetProperty("matches").EnumerateArray()
+                        .Select(m => m.GetProperty("formid").GetString()!).OrderBy(x => x, StringComparer.Ordinal).ToArray());
+    }
+
+    /// <summary>The gather itself, under the tool: every candidate's winner body, in one enumeration per winner
+    /// plugin. Its answer must be the winner's body — the damage each record's winner declares — for all thirty.</summary>
+    [Fact]
+    public void EveryCandidateGetsItsOwnWinnersBody_NotSomeOtherPluginsCopy()
```
Avick3110: This test does not exercise the code the PR changes. It passes `formids:` only — no `types:`, no `plugins:`, no `where:`, no `where_source:` — so `CrossQuery` takes the `!hasType && !hasPlugins && hasFormidSet` set-alone branch at LoadOrderService.cs:4039, which still does a per-key `view.GetRecord(setSession, ...)` and never touches `WinnerBodies`. `whereWinnerActive` is also false here (it requires `whereWinner && hasPlugins`), so the entire winner-gather path is bypassed.

The doc comment above it claims the opposite: "The gather itself, under the tool: every candidate's winner body, in one enumeration per winner plugin." As written the test would still pass if `WinnerBodies.For` returned an empty dictionary. Give it `plugins:` + `where:` + `where_source: "winner"` (or drop the claim) so it actually pins the batched gather.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4141,8 +4141,26 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
+            // at a time is a whole-overlay walk each (#251). The candidate keys are settled by the index-only
+            // filters below, so they are collected in a pre-pass that touches no body, and the winner bodies are
+            // then gathered by plugin — one enumeration per distinct winner plugin, so the whole fetch is bounded
+            // by the order rather than by the candidate count.
+            Dictionary<FormKey, IMajorRecordGetter>? winnerBodies = null;
             try
             {
+                if (whereWinnerActive)
+                {
+                    var candidates = new HashSet<FormKey>();
```
Avick3110: The pre-pass turns a previously O(1)-memory scan into an O(candidates)-memory one, and it does so on exactly the headline use.

Before, each winner body was fetched, filtered, and dropped, so the scan's live set was one record. Now `candidates` holds every surviving FormKey, and `winnerBodies` then holds a getter for each one, both alive for the whole scan. Concretely: `plugins=["Skyrim.esm"], where=..., where_source=winner` with no `types=` — the "broad post-patch audit" the PR description names — considers roughly a million records on a real order. That is a ~1M-entry `HashSet<FormKey>` plus a ~1M-entry `Dictionary<FormKey, IMajorRecordGetter>` whose values each pin an overlay record object, comfortably into the hundreds of MB, held simultaneously. The synthetic benchmark (19.2k master) is three orders of magnitude short of hitting this.

Worth either bounding it (gather per winner plugin and drain, rather than gather-all-then-stream) or stating the ceiling, since the failure mode is an OOM on the exact scan the feature is for.

src/housecarl-core/WinnerBodies.cs
```diff
@@ -0,0 +1,48 @@
...
+            if (view.ResolveWinner(fk) is not { } w) continue;
+            if (!byPlugin.TryGetValue(w.WinnerPlugin, out var set)) byPlugin[w.WinnerPlugin] = set = new HashSet<FormKey>();
+            set.Add(fk);
+        }
+        foreach (var (plugin, wanted) in byPlugin)
+            // A winner plugin that will not OPEN leaves its candidates absent rather than ending the scan, which is
+            // what the per-record fetch this replaces did: each such candidate is then the caller's own
+            // "the index named a winner that did not re-resolve" row, and the rest of the scan still answers.
+            try { view.CollectRecords(session, plugin, wanted, getterTypes, bodies); }
+            catch { /* absent; the caller reports each candidate */ }
```
Avick3110: The bare `catch` discards the only diagnostic the caller had.

Under the old per-record path, a winner plugin that would not open threw out of `session.Overlay(idx)` inside the per-record `try` in CrossQuery, and the sample line was `$"{fk} — {ex.GetType().Name}: {ex.Message}"` — i.e. "IOException: the process cannot access the file ... because it is being used by another process", which tells the user to close xEdit. Now the exception is swallowed here and every candidate of that plugin instead reports "winner 'X.esp' did not yield the record on winner-source re-fetch", which points at index staleness rather than the held-open file. Same accounting, strictly worse cause.

It also swallows `ArgumentException` (plugin not in the order / excluded), which the enclosing `catch (ArgumentException ex) => CrossQueryOutcome.Fail(ex.Message)` at LoadOrderService.cs:4258 would otherwise have surfaced as a named failure — that is the "never a silently degraded mode" cornerstone.

Suggest capturing the exception per plugin and handing it back (an out-parameter or a second dictionary keyed by plugin) so the caller can put the real reason in `unscannableSamples`.

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -1017,6 +1023,45 @@ IReadOnlyList<string> DeclaredMasters(string pluginName, IndexSnapshot s)
...
+        if (getterTypes is { Count: > 0 })
+        {
+            foreach (var t in getterTypes)
+                foreach (var rec in ov.EnumerateMajorRecords(t, throwIfUnknown: false))
+                    if (want.Contains(rec.FormKey) && !sink.ContainsKey(rec.FormKey))
+                    {
+                        sink[rec.FormKey] = rec;
+                        if (++got == want.Count) return;
+                    }
+            if (got > 0) return;                                           // the typed walk routed; what it did not find, this plugin does not hold
```
Avick3110: This is a weaker fallback than the `SeekBody` rule the doc comment says it copies, and the inline justification is not sound.

`SeekBody` decides the fallback per KEY: if the typed walk did not produce this key, run the flat walk. Here the decision is made once for the whole plugin, on "did the typed walk produce anything at all". With `getterTypes = [A, B]` where Mutagen's typed enumeration routes A but not B (`throwIfUnknown: false` makes a non-routing type yield silently), a plugin holding wanted keys of both types sets `got > 0` from A's records and returns, and every B candidate is silently absent — surfacing to the user as `unscannable` "the index named a winner that did not re-resolve" rather than as a match.

The comment "what it did not find, this plugin does not hold" is exactly the inference `SeekBody`'s own doc comment warns against ("a type it does not route would otherwise turn 'this plugin has the record' into a silent 'it does not'").

The cheap fix keeps the batching: fall through to the flat walk whenever `got < want.Count`, not only when `got == 0` — same cost on a miss as `SeekBody` pays, and it cannot answer wrong.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4141,8 +4141,26 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
+                    foreach (var (fk, depth, _, _) in view.RecordsIn(plugins!, types))
```
Avick3110: The pre-pass re-enumerates the whole scope, so the scoped walk is now paid twice. `RecordsIn` opens each scoped plugin's overlay, walks it, and disposes it (LoadOrderResolver.cs:1137-1157) — it is not cached — so this `foreach` is a second full open-and-walk of every plugin in `plugins=`, and the stream below immediately does it again.

That is a clear net win when the candidate count is large (the PR's numbers), but it regresses the narrow shapes: `plugins=["Skyrim.esm"], formids=[one key], where=..., where_source=winner` went from one Skyrim.esm walk plus one winner walk to TWO Skyrim.esm walks plus one winner walk — roughly 2x, on a scan the user framed as a single-record question.

Materialising the pre-pass tuples (fk, depth, body, source) into a list and streaming from that instead of calling `RecordsIn` twice would keep the batching win and drop the second walk, at the cost of the list — which, given the dictionary built right below it is already O(candidates), is not a new order of magnitude.

src/housecarl-mcp-tests/WinnerSourceScanTests.cs
```diff
@@ -0,0 +1,205 @@
...
+            master.Weapons.Add(new Weapon(fk, SkyrimRelease.SkyrimSE)
+            {
+                EditorID = "HcWsrcWeap" + i,
+                BasicStats = new WeaponBasicStats { Damage = MasterDamage, Value = 10, Weight = 1 },
+            });
+        }
+        Keys = keys;
+        var masterPath = Path.Combine(masterDir, MasterName);
+        master.BeginWrite.ToPath(masterPath).WithLoadOrder(Array.Empty<ISkyrimModGetter>()).Write();
+        var masterOv = SkyrimMod.CreateFromBinaryOverlay(masterPath, SkyrimRelease.SkyrimSE);
```
Avick3110: `masterOv` is a binary overlay (memory-mapped) and is never disposed. It is only needed for the two `Write` calls below, but it stays alive for the fixture's whole lifetime and past it, holding a file handle on `mods/BaseMod/HcWsrcBase.esm`.

Consequence: `Directory.Delete(Root, true)` in `Dispose` (line 97) throws `IOException` on Windows and is swallowed by the best-effort catch, so every run of this test class leaves a full temp instance tree (master + two plugins + a generated corpus) under `%TEMP%` forever. Across CI runs that accumulates.

Wrap it — `using var masterOv = ...`, or `(masterOv as IDisposable)?.Dispose()` after the second `Write`.

src/housecarl-mcp-tests/WinnerSourceScanTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    /// <summary>A scope spanning several plugins still de-dups to one verdict per record: the winner answer is
+    /// FormKey-intrinsic, so a key touched by two scoped plugins must not match twice.</summary>
+    [Fact]
+    public void AMultiPluginScopeStillAnswersOncePerRecord() =>
+        Assert.Equal(Expected(5, 15),
+                     Matched(WinnerSourceWorld.MidDamage, new[] { _w.MasterName, _w.LowName, _w.MidName }));
+
+    /// <summary>The same answer with a type scope, which narrows each winner plugin's walk to that type's GRUP.</summary>
+    [Fact]
+    public void ATypeScopedWinnerSourceScanAnswersTheSameSet()
```
Avick3110: Unlike the `Matched` helper above, this one skips `Assert.False(doc.RootElement.TryGetProperty("error", out _), json)` before reaching for `"matches"`. If the type-scoped scan fails — which is the scenario this test exists to catch, since it is the only arm that passes `getterTypes` down into `CollectRecords`'s typed walk — the failure surfaces as a bare `KeyNotFoundException: matches`, with the tool's own error message (the thing that would say what went wrong) thrown away. Reuse `Matched` with a `types` parameter, or add the same assertion.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A broad `where_source=winner` scan is no longer slower the more records it considers.** Deciding the match on
+  the live winner used to re-read each candidate's winner by walking that plugin's whole record list, once per
+  candidate, so a post-patch audit over a large master could spend minutes doing nothing but re-reading. The
+  winner bodies are now gathered a plugin at a time — one pass per winner plugin however many of its records are
+  wanted — so the cost is bounded by the load order instead of by the number of candidates. On a synthetic order
+  of one 19.2k-record master and nine plugins, an audit of every record the master touches went from 165 s to
```
Avick3110: The number does not match the measurement. The PR body's table gives the after figure for that scan as **46.9 s** (and 50.0 s for the same scan on the default source); the changelog says 51 s. It reads as if 51 s were the measured result rather than the default-source comparison it is borrowed from. Either quote 47 s, or say "to 47 s — the same 50 s the scan costs with `where_source` left at its default".

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4170,14 +4188,11 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
                         IMajorRecordGetter filterBody = body;
                         if (whereWinnerActive)
                         {
-                            var w = view.ResolveWinner(fk);
-                            if (w is null) continue;                              // the key came from the order, so a winner must exist; defensive
-                            var wb = view.GetRecord(winnerSession!, w.Value.WinnerPlugin, fk);
-                            if (wb is null)
+                            if (!winnerBodies!.TryGetValue(fk, out var wb))
```
Avick3110: **Low/Medium — a record with no winner changes from a clean non-match to an unscannable row with a nonsense sentence.**

The old code handled the two absences separately:

```
if (w is null) continue;                    // clean skip, no accounting
var wb = view.GetRecord(...);
if (wb is null) { unscannable++; ... }      // accounted
```

`WinnerBodies.For` folds them together: a candidate whose `ResolveWinner` returns null is `continue`d out of `byPlugin`, so it is absent from `winnerBodies` and arrives here as an unscannable row reading `<fk> — winner '?' did not yield the record on winner-source re-fetch`. That sentence is not true of that case (nothing was re-fetched, and there was no winner to name), and it inflates the `unscannable` count and the scan note for something the old path treated as a silent non-match. Both branches call it defensive/impossible, but the two arms now report differently, and the `'?'` sentence is the one a user would have to act on.

Keeping them apart is cheap: have `For` return the unresolved keys too, or check `view.ResolveWinner(fk) is null` here first and `continue` as before.

Related, in the same hunk: `predicate.FatalError` (`break` at the `predicate.Matches` call below) used to abort within the first record or two. The pre-pass now runs the entire candidate collection **and** the whole by-plugin gather before the first predicate evaluation, so a `where=` that can never match anything — a numeric op on a non-numeric field — costs the full gather before it is named. Not wrong, but it turns a fast failure into a slow one on exactly the broad scans this PR is about.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4141,8 +4141,26 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
+                {
+                    var candidates = new HashSet<FormKey>();
+                    foreach (var (fk, depth, _, _) in view.RecordsIn(plugins!, types))
+                    {
+                        if (setFilter is not null && !setFilter.Contains(fk)) continue;
+                        if (conflictsOnly && depth <= 1) continue;
+                        if (definedIn && !scopedModKeys!.Contains(fk.ModKey)) continue;
+                        candidates.Add(fk);
+                    }
+                    winnerBodies = WinnerBodies.For(view, winnerSession!, candidates, types);
```
Avick3110: **Low — the gather re-fetches bodies the scan already has in hand.**

Under `plugins=`, the stream below yields each record with the *scoped plugin's own* body and its filename in `source`. For every candidate the scoped plugin also **wins** — which on the headline case (`plugins=[the master]`, a broad audit) is most of them, and in this PR's own fixture is 15 of the 30 — the winner body and the streamed body are the same record, so the gather walks a plugin again to re-find records the stream is about to hand over for free.

Skipping those keys when building the candidate set (`view.ResolveWinner(fk)?.WinnerPlugin` equals the yielding `source`, then use `body` as `filterBody` in the loop) drops both the biggest plugin's gather walk and its share of the retained `winnerBodies` entries — which is also the memory point raised on line 4154. It would have shown up in the measurements: the 46.9 s after-figure for the master audit is essentially the cost of walking that master twice more.

Worth checking against the reported numbers rather than taking on faith — but as written the by-plugin gather does strictly more work than the information already in the stream requires.

Avick3110:
Review fold: all ten findings were real and all ten are folded. Nothing refuted. Two of them carry a caveat about what could be tested, below.

**Folded**

1. **The test did not exercise the gather.** It passed `formids:` only, taking the set-alone branch. It is now a `plugins=` + `where=` + `where_source=winner` scan whose filter the scoped master's own body cannot satisfy — all thirty of its weapons declare the same damage — so the fifteen matches can only come through the gather. `Pin the gather under the tool, the held-open winner and the typed miss`.

2. **O(candidates) memory.** The pre-pass is gone. The scan now buffers a chunk of 10,000 candidates off the one scoped stream, gathers that chunk's winner bodies and drains it before reading on, so at most 10,000 rows and 10,000 winner getters are live at once whatever the scan considers. Re-measured at a 200,000-record master: gather-all peaked ~210 MB above the no-gather floor (≈1.05 KB per candidate — about a gigabyte extrapolated to the 1M-record scope you named), the chunked gather peaks **on** the floor. Full table in the PR body. `Gather the winner bodies one chunk of candidates at a time`.

3. **The bare catch discarded the cause.** `WinnerBodies.For` now hands back a per-plugin `PluginUnreadableException`, and `ArgumentException` is rethrown so the enclosing handler still names it. A held-open winner reports "could not read 'X.esp' … being used by another process", and the plugin is named once as a coverage gap rather than only sampled three rows deep. Test: `AWinnerPluginHeldOpenNamesTheHeldFile_NotAStaleIndex`.

4. **Flat fallback decided per plugin.** `CollectRecords` falls through for whatever the typed walk did not produce, per key, exactly as `SeekBody` does; `got` is also seeded from keys already in the sink so a partly-filled sink cannot fake completion. Test: `ATypedGatherFallsThroughForAWantedKeyThatTypeDoesNotCover`. *Caveat:* I could not build the failure through the tool — `types=` resolves to types Mutagen routes, so the silent-absence case needs a non-routing type and the test drives `CollectRecords` directly. The bug is real at the primitive, which is public.

5. **Null winner became an unscannable row.** `view.ResolveWinner(fk) is not { } w → continue` now happens while the chunk is filled, before anything is gathered, so it is a clean non-match again and the `'?'` sentence is unreachable. *Caveat:* no test — `RecordsIn` only yields keys the index holds, and `ResolveWinner` is null only for a key the index does not hold, so the arm cannot be reached through the tool. I left it as a skip rather than delete it. The `FatalError` half is fixed by the same chunking: a `where=` that can never match is now named after one chunk instead of after the whole gather.

6. **Re-fetching what the stream already had.** A candidate whose winner plugin equals the streamed `source` is not gathered at all; its streamed body is the winner's. On the 19.2k master audit that is 1,200 of 19,200 gathered instead of 19,200. You were right that it shows in the numbers — see (7).

7. **The scope walked twice.** The pre-pass was the second walk and it is gone; the chunk fills from the single stream. Measured: the master audit went 0.5 s → 0.4 s and the nine-plugin scope 0.4 s → 0.2 s against the gather-all build on the same bytes.

8. **`masterOv` never disposed.** Wrapped in `try/finally` inside the fixture constructor, so the temp tree deletes.

9. **Type-scoped arm skipped the error assertion.** It reuses `Matched` with a `types` parameter now.

10. **CHANGELOG said 51 s.** The whole entry is rewritten against the re-measurement, and a second entry covers the cause a held-open winner plugin now reports.

**Re-measured**

Three builds against the same synthetic bytes (`main`'s per-record fetch, the gather-all you reviewed, the chunked gather). 19,200-record master, nine plugins × 2,000 overrides: the master audit under `where_source=winner` is 34.0 s → 0.5 s → **0.4 s**, against 0.2 s for the same scan on the default source. The prior 165 s / 46.9 s figures were a different scan shape and are not comparable; the PR body now carries only what I measured, including the memory table.

1184 tests (was 1182 — the fixed gather test plus two new arms), 128/128 probes.

Avick3110:
Blind review folded (10 of 10 real, 0 refuted; chunked gather bounds memory to 10,000 rows regardless of candidate count). Queued for Aaron's review. Merge after #553.

src/housecarl-core/WinnerBodies.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+        foreach (var (plugin, wanted) in byPlugin)
+        {
+            // A winner plugin that will not open leaves its candidates absent rather than ending the scan, which is
+            // what the per-record fetch this replaces did — but the CAUSE travels with it, so a file another program
+            // is holding open reads as that and not as a stale index. An ArgumentException is a named failure about
+            // the request itself (a plugin outside the order), so it is left to the caller's own handler.
+            try { view.CollectRecords(session, plugin, wanted, getterTypes, bodies); }
+            catch (ArgumentException) { throw; }
+            catch (Exception ex) { unreadable[plugin] = new PluginUnreadableException(plugin, ex); }
```
Avick3110: **[correctness | medium]** This catch-all wraps *every* exception from `CollectRecords` — including one raised while enumerating an already-opened plugin — in `PluginUnreadableException`, whose message asserts the opposite of what happened: *"it opened when the load order was indexed but not now — another program is probably holding it open. Close that program and run this again."*

Concrete case: a winner plugin opens fine (`session.Overlay(idx)` succeeds) but throws mid-walk in `ov.EnumerateMajorRecords` — a record or GRUP whose structure Mutagen cannot step over, e.g. the file was rewritten by another tool after the index build, which the last-write-plus-size check has not yet picked up. The scan then tells the user to close a program that is holding nothing, names the plugin as a whole-plugin coverage gap in `unreadablePlugins`, and counts every one of its chunk candidates unscannable. `main` reported that same fault as one unscannable record carrying the real exception type and message.

Every other site in the resolver that raises this exception wraps **only** the open call, for exactly this reason — `RecordsIn` (LoadOrderResolver.cs:1125), `WinnerRecordsOfType` (:1083), `DeclaredMasters` (:1039). Since the PR's headline is "says WHY a winner plugin did not give up its record", it seems worth catching around `session.Overlay(idx)` inside `CollectRecords` and letting enumeration faults out unwrapped, so the scan's own per-record handler accounts them with their real cause.

Avick3110: **[correctness | low]** The premise this guard states is not true of the callee: `CollectRecords` never throws `ArgumentException` for a plugin outside the order or an excluded one — it returns silently, mirroring `GetRecord`:

```csharp
if (!_nameToIdx.TryGetValue(pluginName, out int idx)) return;
if (s.Excluded.Contains(idx)) return;
```
(LoadOrderResolver.cs:1066-1067)

So the rethrow can only ever fire on an unrelated `ArgumentException` — one Mutagen raises from somewhere under `ov.EnumerateMajorRecords` for a single winner plugin — and there it defeats the per-plugin isolation it sits inside: the exception escapes `WinnerBodies.For` and `DrainChunk` and lands in the scan's outer `catch (ArgumentException ex) { return CrossQueryOutcome.Fail(ex.Message); }`, the handler labelled *plugin not in order / unknown type*. The whole scan then fails with a bare framework message naming neither the plugin nor the record, instead of that plugin's candidates going absent while the rest of the chunk scans.

Dropping the line leaves the general catch below to isolate it.

```suggestion
            try { view.CollectRecords(session, plugin, wanted, getterTypes, bodies); }
            catch (Exception ex) { unreadable[plugin] = new PluginUnreadableException(plugin, ex); }
```

Avick3110:
Aaron's review: both findings folded, nothing refuted — though the first is folded with a different mechanism than suggested, for the reason below.

**Folded — the catch-all mislabelled a mid-walk fault as a held-open file.** Real. `CollectRecords` opened and walked inside one `try`, so a plugin that opened fine and then threw out of `EnumerateMajorRecords` was reported with the held-open wording, telling the user to close a program holding nothing. `CollectRecords` now wraps only `session.Overlay(idx)` in `PluginUnreadableException`, exactly as `RecordsIn`, `WinnerRecordsOfType` and `DeclaredMasters` do, and its doc comment says a walk fault is left as it came.

Where I differed: you suggested letting the walk fault out unwrapped so the scan's per-record handler accounts it with its real cause. There is no per-record handler over the gather — `ScanRow`'s `try` covers the filtering, not the fetch — so unwrapped it reaches the outer `catch (Exception)` and aborts the whole scan, which is worse isolation than `main` had. Instead the gather keeps the fault per plugin and gives it its own wording: `PluginUnscannableException` (a subclass, so every `PluginUnreadableException` collector and the `PluginName` plumbing are unchanged) says the plugin opened but its records could not be walked to the end and has probably changed since the index was built. Same accounting shape as before — one plugin's coverage gap, the rest of the scan still answers — with the cause it actually had. Test: `AWinnerPluginThatFaultsMidWalkNamesTheChangedFile_NotAHeldOpenOne` overstates the first WEAP record's length in the mid plugin after the index is built, so the file opens and the walk dies on it; it asserts the fault is isolated, names the file as changed, and does not say "holding it open".

**Folded — the `ArgumentException` rethrow guarded a premise the callee does not hold.** Real, and verified: `CollectRecords` returns silently for a plugin not in `_nameToIdx` and for an excluded one, so the rethrow could only ever fire on an `ArgumentException` Mutagen raised mid-walk for a single plugin, and would then fail the whole scan through the handler labelled "plugin not in order / unknown type", naming neither plugin nor record. The line is gone; that fault is now one plugin's, like any other. Test: `APluginOutsideTheOrderLeavesTheGatherEmptyRatherThanThrowing` pins the premise directly — a gather against a plugin outside the order leaves the sink empty and throws nothing.

No change to the chunk size or the memory bound, so the measurements stand. 1416 tests (was 1414 on this base) and 128/128 probes pass.

PR #571, opened 2026-09-05
Stamp the dialogue check with the build it read

### LINKED ISSUE #396: [GAP] a dialogue verdict straddles the record and asset substrates, so no epoch is stamped and none is owned

Avick3110:
The dialogue family of `housecarl_check` was the one section of a merged response with no freshness claim at all: it stamped no epoch, because half its verdicts — whether a line's `.fuz` is on disk, whether a result script's `.pex` chain exists, whether the `.seq` lists the quest and is newer than its plugin — come off the asset substrate, which the record fingerprint does not cover. Omitting the stamp was a worse answer than declaring the bound. `DialogueCheckResult` now carries `Epoch`, taken off one pinned view that `CheckDialogue` captures and hands to every seed's validation, so the stamp names the build the response was actually answered from rather than whichever one the last seed happened to catch. Both renders write it through the shape the errors and scripts families already use: json gets `epoch`, `epoch_covers_all_inputs: false` and `epoch_uncovered` naming the three classes; the text render states the same stamp and the same three classes on its own line. `WriteSweepEpoch` gained the uncovered arm rather than a second writer, and the class names are one array both transports read, so the two cannot name different sets. The "no honest stamp exists" remarks in `DialogueCheck.cs` and `LoadOrderService.cs` are gone — the SPEC sentence they pointed at exists now (§2.1, amended 2026-09-06).

A refusal still carries no stamp: a family that validated nothing has no build to claim, which is what the sibling families do.

Test: `FactV2_TheStampDeclaresItsBound` drives a dialogue check on the shared world and asserts the stamp and the three names in both transports, with the uncovered set read off the same array the render writes.

Closes #396

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -64,9 +64,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `epoch_covers_all_inputs: false` and `epoch_uncovered`, which names the verdicts read off disk rather than off the
   record index: voiced lines (.fuz on disk), the result-script .pex chain, and .seq coverage and staleness. The text
   render states the same stamp and the same three classes on its own line. Compare two responses' `epoch` to know
-  whether the record half changed between them; the named classes are outside that comparison. A refusal — no seeds,
-  or every seed unresolvable — carries the bare `epoch` with no coverage claim beside it, which is what the errors and
-  scripts families' refusals carry.
+  whether the record half changed between them; the named classes are outside that comparison. A call whose every seed
+  was a dialogue view (DLVW) or branch (DLBR) reads nothing off disk, so it stamps `epoch_covers_all_inputs: true` and
+  names no classes. A dialogue-only call refused because every seed was unresolvable carries the bare `epoch` — the
+  build it was decided against — as the errors and scripts families' post-index refusals do; a call with no `seeds=` is
+  refused on the arguments alone before any build is read, and names none, as theirs do not.
 
 - **`housecarl_asset_status` and `housecarl_place` take `format='json'`.** Both answered text only, so a caller
   reading them from a script parsed prose. The json document carries the same data as the text render: for
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -64,7 +64,9 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `epoch_covers_all_inputs: false` and `epoch_uncovered`, which names the verdicts read off disk rather than off the
   record index: voiced lines (.fuz on disk), the result-script .pex chain, and .seq coverage and staleness. The text
   render states the same stamp and the same three classes on its own line. Compare two responses' `epoch` to know
-  whether the record half changed between them; the named classes are outside that comparison.
+  whether the record half changed between them; the named classes are outside that comparison. A refusal — no seeds,
+  or every seed unresolvable — carries the bare `epoch` with no coverage claim beside it, which is what the errors and
+  scripts families' refusals carry.
 
 - **`housecarl_asset_status` and `housecarl_place` take `format='json'`.** Both answered text only, so a caller
   reading them from a script parsed prose. The json document carries the same data as the text render: for
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -58,6 +58,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   reads as one covering all of them. `walk.max_nodes` says which reading it spends: per seed on the carrier
   walk, one shared budget across every seed and hop on the transitive walk.
 
+- **The dialogue check stamps the build it read, and says what that stamp does not cover.** `housecarl_check
+  findings=["dialogue"]` was the one family answering with no freshness claim at all. It now carries `epoch` — the
+  record build every seed was validated against, one pinned build for the whole call — beside
+  `epoch_covers_all_inputs: false` and `epoch_uncovered`, which names the verdicts read off disk rather than off the
+  record index: voiced lines (.fuz on disk), the result-script .pex chain, and .seq coverage and staleness. The text
+  render states the same stamp and the same three classes on its own line. Compare two responses' `epoch` to know
+  whether the record half changed between them; the named classes are outside that comparison.
+
 - **`housecarl_asset_status` and `housecarl_place` take `format='json'`.** Both answered text only, so a caller
   reading them from a script parsed prose. The json document carries the same data as the text render: for
   `asset_status`, the build-level caveats, one row per path with its winner and full provider chain, and the same
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -26,8 +26,8 @@ saying it sets an expectation their install may contradict. Say what is known, a
   beside the epoch, never inside it, so two builds that differ only in health still compare as different builds.
   A healthy order carries nothing. The sentence names up to ten plugins and counts the rest —
   `housecarl_load_order_status` lists them all. `housecarl_check` states the sentence once, at the top of the
-  response, so a dialogue-only check — the one family that carries no epoch — says it too; each swept family
-  carries the flag and the plugin count beside its own `epoch`, the same fact its text head states.
+  response, so a dialogue-only check says it too; each family carries the flag and the plugin count beside its own
+  `epoch`, the same fact its text head states.
 - **`housecarl_records` walks the reverse direction to any depth.** `walk={"direction":"reverse","depth":N}`
   under a reading form now answers "what points at the seeds, and what points at that" — every link at every
   hop, off the reverse-reference index, with no bounding `types=`/`plugins=` scope; the depth>1 refusal is gone.
--- a/src/housecarl-core/DialogueCheck.cs
+++ b/src/housecarl-core/DialogueCheck.cs
@@ -46,12 +46,11 @@ public sealed record DialogueSeedResult(string Seed, DialogueValidationReport? R
 /// <param name="CountsOnly">the response carries totals and the unreachable-seed roster, and no topic blocks. The
 /// merged tool's own mode applied to this family — a family that ignored it would render a full listing under a
 /// parameter documented to suppress one.</param>
-///
-/// <remarks><b>No EPOCH, deliberately.</b> The two sweep families stamp the record build they read. A dialogue
-/// validation cannot honestly carry that stamp: core pins one view, but half of what it reports — whether a line's
-/// <c>.fuz</c> is on disk, whether a result script's <c>.pex</c> exists — comes off the ASSET substrate, which the
-/// record fingerprint does not cover. A stamp covering half the answer would be a freshness claim the response
-/// cannot support. An honest stamp spanning both substrates is still unowned.</remarks>
+/// <param name="Epoch">the RECORD build every seed was validated against — the one pinned view the whole call reads,
+/// stamped like the sibling families stamp theirs. It names the record build and nothing else: the verdicts taken off
+/// the ASSET substrate (a line's <c>.fuz</c>, a result script's <c>.pex</c> chain, <c>.seq</c> coverage and staleness)
+/// are outside the fingerprint, so both renders write it with <c>epoch_covers_all_inputs: false</c> and name those
+/// classes rather than omitting the stamp. Null only on a refusal, which validated nothing.</param>
 public sealed record DialogueCheckResult(
     IReadOnlyList<DialogueSeedResult> Seeds,
     int TopicsFound,
@@ -60,7 +59,8 @@ public sealed record DialogueCheckResult(
     string? Error = null,
     int Limit = 0,
     int SeedsNamed = 0,
-    bool CountsOnly = false)
+    bool CountsOnly = false,
+    string? Epoch = null)
 {
     public bool Success => Error is null;
 
--- a/src/housecarl-core/DialogueCheck.cs
+++ b/src/housecarl-core/DialogueCheck.cs
@@ -50,7 +50,8 @@ public sealed record DialogueSeedResult(string Seed, DialogueValidationReport? R
 /// stamped like the sibling families stamp theirs. It names the record build and nothing else: the verdicts taken off
 /// the ASSET substrate (a line's <c>.fuz</c>, a result script's <c>.pex</c> chain, <c>.seq</c> coverage and staleness)
 /// are outside the fingerprint, so both renders write it with <c>epoch_covers_all_inputs: false</c> and name those
-/// classes rather than omitting the stamp. Null only on a refusal, which validated nothing.</param>
+/// classes rather than omitting the stamp. A refusal carries the bare stamp — the build it was decided against, with
+/// no coverage claim beside it, which is what the sibling families' refusals carry.</param>
 public sealed record DialogueCheckResult(
     IReadOnlyList<DialogueSeedResult> Seeds,
     int TopicsFound,
@@ -80,6 +81,8 @@ public sealed record DialogueCheckResult(
     /// could not evaluate, not a description of the listing.</summary>
     public int ConditionedInfos => Topics.Sum(x => x.Topic.ConditionedInfoCount);
 
-    public static DialogueCheckResult Fail(string error) =>
-        new(Array.Empty<DialogueSeedResult>(), 0, 0, false, error);
+    /// <summary>The family's refusal, stamped with the build it was decided against — the sibling families stamp
+    /// theirs, and a refusal decided after the view was captured has a build to name.</summary>
+    public static DialogueCheckResult Fail(string error, string? epoch = null) =>
+        new(Array.Empty<DialogueSeedResult>(), 0, 0, false, error) { Epoch = epoch };
 }
--- a/src/housecarl-core/DialogueCheck.cs
+++ b/src/housecarl-core/DialogueCheck.cs
@@ -50,8 +50,8 @@ public sealed record DialogueSeedResult(string Seed, DialogueValidationReport? R
 /// stamped like the sibling families stamp theirs. It names the record build and nothing else: the verdicts taken off
 /// the ASSET substrate (a line's <c>.fuz</c>, a result script's <c>.pex</c> chain, <c>.seq</c> coverage and staleness)
 /// are outside the fingerprint, so both renders write it with <c>epoch_covers_all_inputs: false</c> and name those
-/// classes rather than omitting the stamp. A refusal carries the bare stamp — the build it was decided against, with
-/// no coverage claim beside it, which is what the sibling families' refusals carry.</param>
+/// classes rather than omitting the stamp. Null on a refusal decided before any build was read, which is where the
+/// sibling families leave theirs null too.</param>
 public sealed record DialogueCheckResult(
     IReadOnlyList<DialogueSeedResult> Seeds,
     int TopicsFound,
@@ -81,8 +81,9 @@ public sealed record DialogueCheckResult(
     /// could not evaluate, not a description of the listing.</summary>
     public int ConditionedInfos => Topics.Sum(x => x.Topic.ConditionedInfoCount);
 
-    /// <summary>The family's refusal, stamped with the build it was decided against — the sibling families stamp
-    /// theirs, and a refusal decided after the view was captured has a build to name.</summary>
+    /// <summary>The family's refusal. It carries a stamp only when it was decided AFTER a build was read — the split
+    /// the sibling families keep: a refusal decided on the arguments alone has no build to name, and reading one to
+    /// stamp it would build the index for a call that is about to refuse.</summary>
     public static DialogueCheckResult Fail(string error, string? epoch = null) =>
         new(Array.Empty<DialogueSeedResult>(), 0, 0, false, error) { Epoch = epoch };
 }
--- a/src/housecarl-mcp-tests/DialogueFamilyTests.cs
+++ b/src/housecarl-mcp-tests/DialogueFamilyTests.cs
@@ -201,8 +201,8 @@ public sealed class DialogueFamilyTests
         Assert.Equal(DialogueSweepRender.EpochUncovered,
                      fam.GetProperty("epoch_uncovered").EnumerateArray().Select(e => e.GetString()).ToArray());
 
-        // A refusal carries the BARE stamp, as the sibling families' refusals do (EpochCheckSweepTests.FactE5_6):
-        // the view is pinned before the seeds are read, so there is a build to name, and nothing was covered by it.
+        // A refusal reached AFTER the build was read carries the bare stamp, as the sibling families' post-capture
+        // refusals do (EpochCheckSweepTests.FactE5_6): there is a build to name, and nothing was covered by it.
         var refused = Svc.CheckDialogue(new[] { "not-a-formid" }, 1000);
         Assert.NotNull(refused.Error);
         Assert.Equal(result.Epoch, refused.Epoch);
@@ -211,6 +211,13 @@ public sealed class DialogueFamilyTests
         var refusedJson = JsonWire.RenderCheck(new CheckSweep(DialogueSel(), Dialogue: refused), 20000);
         Assert.Equal(refused.Epoch, JsonDocument.Parse(refusedJson).RootElement.GetProperty("epoch").GetString());
         Assert.DoesNotContain("epoch_covers_all_inputs", refusedJson);
+
+        // …and a refusal decided on the ARGUMENTS alone names no build, which is the split the siblings keep: it is
+        // answered without reading one, so there is nothing honest to stamp.
+        var unseeded = Svc.CheckDialogue(Array.Empty<string>(), 1000);
+        Assert.NotNull(unseeded.Error);
+        Assert.Null(unseeded.Epoch);
+        Assert.DoesNotContain("epoch=", Wire.RenderCheck(new CheckSweep(DialogueSel(), Dialogue: unseeded), 20000));
     }
 
     /// <summary>Everything AFTER the seed prefix the sweep echoes — the composed refusal itself. The sweep writes
--- a/src/housecarl-mcp-tests/DialogueFamilyTests.cs
+++ b/src/housecarl-mcp-tests/DialogueFamilyTests.cs
@@ -1,3 +1,4 @@
+using System.Text.Json;
 using System.Text.RegularExpressions;
 using Mutagen.Bethesda.Skyrim;
 using HousecarlCore;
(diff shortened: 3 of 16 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2294,17 +2294,21 @@ public ReadOutcome ResolveRead(FormKey fk, string? plugin, IReadOnlyList<string>
...
     public DialogueCheckResult CheckDialogue(IReadOnlyList<string>? seeds, int limit, bool countsOnly = false)
-        => DialogueSweep.Run(ValidateDialogue, OpenFormIdDoor().Parse, seeds, limit, countsOnly);
+    {
+        // One resolver and one view for the whole call, the contract the sibling families keep: every seed is validated
+        // against the same build, so the stamp below names the build the response was actually answered from rather
+        // than whichever one the last seed happened to catch.
+        var resolver = Resolver;
+        var view = resolver.Capture();
+        return DialogueSweep.Run(fk => DialogueValidate.Run(resolver, Assets, fk, view),
+                                 OpenFormIdDoor().Parse, seeds, limit, countsOnly, view.Epoch);
```
Avick3110: The seed parse still captures its OWN build, so the stamp does not name the build the seeds were resolved against. `OpenFormIdDoor()` returns a `FormIdDoor.For(this)`, and on the first *runtime* FormID seed its `Parse` calls `_svc.CaptureView()` -> the `Resolver` property -> `RefreshOnProfileChange()` + `RefreshIfStale()`, which can rebuild the snapshot. Concretely: `check findings=["dialogue"] seeds=["FE012800"]` captures view A here, then the door captures view B a moment later; if a plugin was ticked or the order re-sorted in between, `FE012800` maps to a different record under B than under A, and the response stamps A. It also pays a second whole-order stat sweep per call. Pinning the door to the view already captured fixes both.

src/housecarl-mcp/DialogueSweep.cs
```diff
@@ -17,9 +17,12 @@ internal static class DialogueSweep
     /// <param name="limit">how many seeds this call may expand. Over it, the extra seeds are not validated and the
     /// response states how many and which knob moves them.</param>
     /// <param name="countsOnly">carry the totals and the unreachable-seed roster, and no topic blocks.</param>
+    /// <param name="epoch">the record build every seed was validated against, stamped onto the result. Left off the
+    /// refusals, as the sibling families leave it off theirs: a family that validated nothing has no build to claim.</param>
```
Avick3110: "as the sibling families leave it off theirs" is not what the sibling families do — they stamp their refusals. `EpochCheckSweepTests.FactE5_6` asserts exactly that: `Svc.CheckErrors(new[]{"Nope.esp"},1000)` refuses and still carries `Assert.Equal(current, locate.Epoch)`, the refusal json carries `epoch`, and the scripts not-in-order refusal does the same (only `epoch_covers_all_inputs` is success-path-only). So this family now diverges from the contract the comment claims to follow: `check findings=["dialogue"] seeds=["not-a-formid"]` returns a refusal with no stamp, while `check findings=["errors"] plugins=["Nope.esp"]` returns one with a stamp — and by the time the second `Fail` below (line 70, every seed unresolvable) is reached, `CheckDialogue` has already captured the build, so there is a build to claim. Both `Fail` sites can take `epoch`.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2294,17 +2294,21 @@ public ReadOutcome ResolveRead(FormKey fk, string? plugin, IReadOnlyList<string>
     /// a verify step: a mid-run resolve or asset failure rides
     /// <see cref="DialogueValidationReport.CheckError"/>, and a not-in-order or wrong-type input is a named
     /// <see cref="DialogueValidationReport.Error"/>.</summary>
-    // No epoch is stamped: the report is one build's answer, but many of its verdicts come off the ASSET substrate,
-    // outside the record fingerprint — the .pex chain behind each result script, the .fuz checks, and the .seq
-    // staleness verdict, which is a file-mtime comparison no record fingerprint expresses. A record fingerprint here
-    // would claim freshness for verdicts it does not describe.
     public DialogueValidationReport ValidateDialogue(FormKey fk) => DialogueValidate.Run(Resolver, Assets, fk);
```
Avick3110, line 2306: The deleted comment above this line explained why `ValidateDialogue` — the `housecarl_validate_dialogue` path — carries no stamp, and that path still carries none. The PR removes the rationale but not the omission, so the two surfaces now disagree with nothing saying why: the same verdicts stamped under `check findings=["dialogue"]`, unstamped under `validate_dialogue`. Either stamp it here the same way (with the same uncovered classes) or keep a one-line note saying why this one does not.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in cd1c1ce9. `CheckDialogue` now parses seeds through `FormIdDoor.On(view).Parse`, the view it already captured, so a runtime FormID seed resolves against the build the response stamps and the second whole-order stat sweep is gone.

src/housecarl-mcp/DialogueSweep.cs
Avick3110: Folded in 97e96782. Both `Fail` sites take the pinned epoch, `DialogueCheckResult.Fail` carries it, and `CheckSweep.Epoch` now reads the dialogue family too so a dialogue-only refusal is stamped in both transports. The refusal carries the bare stamp with no `epoch_covers_all_inputs` beside it, matching FactE5_6. The doc comment says that instead of the old claim, and `FactV2_TheStampDeclaresItsBound` asserts the refusal stamp and the absent coverage key. Changelog line updated.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Not folded — there is no `housecarl_validate_dialogue` surface to disagree with. That tool was retired in the fold; `AliasTable` redirects the name into `check findings=["dialogue"]` plus `records project=info_order`, and no MCP tool binds `LoadOrderService.ValidateDialogue` any more. Its only callers are tests (`WinnerScanCoverageTests`) and two generator probes, none of which render a response, so there is nothing that could carry a stamp and no consumer seeing an unstamped one. The two surfaces the comment compares are one surface. The deleted comment was describing a non-stamp on a path that no longer reaches a caller, which is why it went rather than being re-pointed; the design page and the #396 decision line use the retired name as shorthand for exactly the dialogue family this PR stamps.

src/housecarl-mcp/DialogueSweepRender.cs
```diff
@@ -55,9 +55,21 @@ internal static void AppendHead(StringBuilder sb, CheckOutcome o)
...
                                 d.FindingsFound));
+        if (o.Sweep.Dialogue?.Epoch is { } epoch)
+            sb.Append(string.Format(ReadSentences.DialogueEpochBound, epoch, string.Join(", ", EpochUncovered)));
         if (d.CountsOnly) sb.Append(ReadSentences.DialogueCountsOnly);
     }
 
+    /// <summary>The verdict classes this family reports that the record fingerprint does not describe — the ASSET
+    /// substrate half of the answer. Data, read by both transports, so the text sentence and <c>epoch_uncovered</c>
+    /// cannot name different sets.</summary>
+    internal static readonly string[] EpochUncovered =
```
Avick3110, line 73: **The uncovered set is unconditional, so it contradicts the family's own per-kind table.**

`EpochUncovered` is a constant array, written on every dialogue success. But the three classes it names all live behind `DialogueChecks.TopicGraph`, and that flag does not run for every seed kind.

Concretely: `housecarl_check findings=['dialogue'] seeds=['000E01:Mod.esp']` where the seed is a DLVW (or DLBR). `DialogueValidate.Run`'s view/branch arm returns a report with no `TopicValidation`, no `VoiceLines`, no `ScriptFindings`, no `SeqLint` — it does not even set `ReadIncomplete`. That answer is 100% record substrate; the fingerprint covers all of it. Yet the head line says the stamp "does not cover: voiced lines (.fuz on disk), result-script .pex chain, .seq coverage and staleness", and json says `epoch_covers_all_inputs: false` — a few lines above `DialogueBoundaryRecordLevel`, which says "no dialogue graph, voice file, result script or condition was checked here". The response contradicts itself within one section.

Narrower version of the same defect on a DIAL-only call: `CheckSeq` runs only in the QUST arm, so `.seq coverage and staleness` is named as uncovered for a verdict the response never produced.

`o.Dialogue!.Value.ChecksRun` (CheckOutcome.cs:77) is the union already computed for exactly this distinction, and both write sites already do `var d = o.Dialogue!.Value;`. Forking the set on `d.ChecksRun.HasFlag(DialogueChecks.TopicGraph)` — and passing `null` for `uncovered` when it did not run, which `WriteSweepEpoch` already handles by flipping the coverage flag back to true — makes the stamp say what this response actually answered.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2294,17 +2294,23 @@ public ReadOutcome ResolveRead(FormKey fk, string? plugin, IReadOnlyList<string>
...
     /// <summary>The merged <c>check</c> surface's dialogue family: <see cref="ValidateDialogue"/> over a seed list,
     /// tallied for one section of a merged response. Deliberately thin — the family's own grammar (seed parse,
     /// cost refusal, seed budget, tally) lives in <see cref="DialogueSweep"/> rather than in this file.</summary>
     public DialogueCheckResult CheckDialogue(IReadOnlyList<string>? seeds, int limit, bool countsOnly = false)
-        => DialogueSweep.Run(ValidateDialogue, OpenFormIdDoor().Parse, seeds, limit, countsOnly);
+    {
+        // One resolver and one view for the whole call, the contract the sibling families keep: every seed is validated
+        // against the same build, so the stamp below names the build the response was actually answered from rather
+        // than whichever one the last seed happened to catch.
+        var resolver = Resolver;
```
Avick3110: **The eager capture puts a pure input refusal behind a whole index build.**

`var resolver = Resolver;` runs before `DialogueSweep.Run` ever reaches its `named.Length == 0` check.

Before this change, `housecarl_check findings=['dialogue']` with no `seeds=` touched the resolver zero times: `OpenFormIdDoor()` is `FormIdDoor.For(this)`, which captures only when a runtime FormID actually arrives; `ValidateDialogue` was an uninvoked method group; and `SweepSharedInput.Error` short-circuits when no `formids=`/`type=` was passed. The call answered `DialogueNeedsSeeds` immediately. Now it pays a full `Mo2LoadOrder.Build` + `LoadOrderResolver.Build` over the whole order to produce that same sentence.

Worse than the cost: when the profile yields no active plugins — MO2 mid-resort, `loadorder.txt` not yet written — the `Resolver` getter throws `InvalidOperationException("No active plugins resolved from the MO2 profile…")`. `Guard.Tool` catches it and answers *"error: housecarl_check failed unexpectedly — … This is an internal houseCARL failure (the arguments bound fine), not bad input."* for a call whose arguments were exactly the problem. And because that is a bare string, a `format='json'` caller gets prose instead of a document — which is the failure CheckTools.cs:159-164 deliberately routes shared-input refusals through the render to avoid.

`SweepSharedInput` states the rule this breaks in its own comment: don't read the MO2 composition "for a call that may be about to refuse anyway".

Capturing inside `DialogueSweep.Run` after the seed check, or handing the sweep a lazy `Func<string>` for the epoch, keeps the stamp on the refusals that have work behind them and keeps the no-seeds refusal free.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2294,17 +2294,23 @@ public ReadOutcome ResolveRead(FormKey fk, string? plugin, IReadOnlyList<string>
...
+        var view = resolver.Capture();
+        // The seed door is pinned to that same view: a door of its own would capture a second build on the first
+        // runtime FormID, so the seeds could name records from a build other than the one the response stamps.
+        return DialogueSweep.Run(fk => DialogueValidate.Run(resolver, Assets, fk, view),
```
Avick3110: **The record half is pinned; the asset half is still re-taken per seed.**

`Assets` is a property, and it is evaluated inside the lambda — so it is re-gated on every seed. Its getter runs `RefreshOnProfileChange()` and `_assetResolver.RefreshIfStale()`, and can drop and rebuild `_assetResolver` outright.

Concretely: `housecarl_check findings=['dialogue'] seeds=[questA, questB]` over a large order. Between validating questA and questB, a repacked BSA or a toggled loose-file mod trips the asset freshness check. questA's `.fuz` and `.pex` verdicts then come off one asset build and questB's off another, inside one response carrying a single tally and a single stamp — the same cross-value inconsistency `Capture()`'s doc comment exists to prevent, one substrate over.

The sibling families do not have this: both resolve `Assets` once at the call site and thread the resolved `AssetResolver` down (`ScriptPropertyCheck.Run(resolver, view, Assets, …)` at LoadOrderService.cs:5031). Hoisting `var assets = Assets;` next to `var resolver = Resolver;` and closing over it gives the dialogue family the same guarantee for one line.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The dialogue check stamps the build it read, and says what that stamp does not cover.** `housecarl_check
+  findings=["dialogue"]` was the one family answering with no freshness claim at all. It now carries `epoch` — the
+  record build every seed was validated against, one pinned build for the whole call — beside
+  `epoch_covers_all_inputs: false` and `epoch_uncovered`, which names the verdicts read off disk rather than off the
+  record index: voiced lines (.fuz on disk), the result-script .pex chain, and .seq coverage and staleness. The text
+  render states the same stamp and the same three classes on its own line. Compare two responses' `epoch` to know
+  whether the record half changed between them; the named classes are outside that comparison. A refusal — no seeds,
```
Avick3110: **"A refusal … carries the bare `epoch`" is only true when dialogue is the only family selected.**

`housecarl_check findings=['errors','dialogue']` with no `seeds=`: the errors family answers, so `CheckOutcome` does not collapse to a whole-call `Error` and dialogue renders as its own refusal section. The text lane writes only the refusal (the `o.Refusal(f) is { } refusal` arm in `ReadTools.Compose`) and the json lane writes only `"refused"` (JsonWire.cs:1827). `DialogueSweepRender.AppendHead` / `WriteHead` — the only two places the dialogue epoch is written — are never reached.

So a caller who reads this line and parses `families.dialogue.epoch` off a refusal gets nothing in exactly the case the merged surface exists for. Either narrow the sentence to a dialogue-only call, or write the bare stamp in the family refusal arm too.

src/housecarl-core/DialogueCheck.cs
```diff
@@ -80,6 +81,8 @@ public sealed record DialogueCheckResult(
     /// could not evaluate, not a description of the listing.</summary>
     public int ConditionedInfos => Topics.Sum(x => x.Topic.ConditionedInfoCount);
 
-    public static DialogueCheckResult Fail(string error) =>
-        new(Array.Empty<DialogueSeedResult>(), 0, 0, false, error);
+    /// <summary>The family's refusal, stamped with the build it was decided against — the sibling families stamp
```
Avick3110: **"the sibling families stamp theirs" is the opposite of what they do for this class of refusal.**

The siblings deliberately split a refusal decided *on input* from one decided *against a build*. `ErrorCheckResult.Fail(scopeErr)` and `Fail(classErr)` (LoadOrderService.cs:4846-4848) return before `resolver.Capture()` and carry no epoch at all; the scripts family stamps only when `splitErr.Stamped` — i.e. only for a refusal reached after the capture (LoadOrderService.cs:5027-5029).

`DialogueNeedsSeeds` is squarely the first kind, and it is the refusal that forced the eager `Resolver` capture at LoadOrderService.cs:2307. So this comment both misdescribes the convention it cites and is the justification a future change will reach for when stamping the next pre-capture refusal.

src/housecarl-mcp-tests/DialogueFamilyTests.cs
```diff
@@ -176,6 +177,42 @@ public void FactV1_ParityOkProseNamesEveryCoveredSubrecord()
         foreach (var sig in pinQuest) Assert.Contains(sig, byQuestLine);
     }
 
+    // ---- fact V2 --------------------------------------------------------------------------------------
+    // The family stamps the record build every seed was validated against, and DECLARES the bound rather than
+    // omitting the stamp: both transports name the three verdict classes the record fingerprint does not describe.
+
+    [Fact]
+    public void FactV2_TheStampDeclaresItsBound()
```
Avick3110, line 185: **This xUnit fact is the only thing that renders the new line — `ci-all` never does.**

`CheckMergeProbe.DialogueSweep_Run`, `RecordLevelFixture`, `RecordLevelFailingFixture` and `MixedKindFixture` (CheckMergeProbe.cs:1685-1712) all call `DialogueSweep.Run` with the new `epoch` parameter left at its default null, and `CheckShapeMatrix.Build` derives every dialogue shape from those. So `AppendHead`'s new head line and `WriteHead`'s `epoch` / `epoch_covers_all_inputs` / `epoch_uncovered` keys are absent from every shape the matrix sweeps its cap, fixed-part and allocation properties over.

That is why `ci-all` is green: `RecordLevelFixture` is the very shape that would print the stamp sentence next to `DialogueBoundaryRecordLevel` and expose the contradiction flagged on DialogueSweepRender.cs:66, and it renders without the sentence.

Threading a non-null epoch through those four fixtures is a one-argument change and puts the new line inside the cap band the matrix already drives at every integer cap.

src/housecarl-mcp/DialogueSweepRender.cs
Avick3110: Folded in 6e1f782d, with the narrower version refuted below.

`DialogueSweepRender.UncoveredBy(d)` now forks on `d.ChecksRun.HasFlag(DialogueChecks.TopicGraph)` — the union `CheckOutcome` already computes and the boundary already reads. A DLVW/DLBR-only call gets an empty set, so `WriteSweepEpoch` flips `epoch_covers_all_inputs` back to true and writes no `epoch_uncovered`, and the text head takes a second sentence (`DialogueEpochWhole`: "it covers every verdict here") instead of the bound one. `RecordLevelFixture` no longer prints an uncovered set above a boundary saying nothing off disk was checked. New ci-all arm `DIALOGUE-THE-STAMP-DECLARES-ONLY-THE-BOUND-THIS-RESPONSE-HAS` holds both sides (record-level covers whole, quest and mixed name the three), and the existing json arm now checks the two coverage keys fork with it.

**Not folded: `.seq` named on a DIAL-only call.** There is no axis to fork it on. `.seq` is quest-only in the validator, but `DialogueChecks` has no term for it — `DialogueKindChecks.For` puts it inside `TopicGraph`, and the union `ChecksRun` carries cannot separate "a quest ran" from "a topic ran": a mixed DLVW + DIAL call yields `RecordParity | TopicGraph`, the same value a quest yields, so any test off that value is wrong on one of the two. Separating it means a third flag, which changes `DialogueKindChecks.Names` — the per-seed `checks_run` array a consumer parses — to caveat one word. The wide boundary sentence does not fork on `.seq` either: it names the graph, `.fuz`, result-script and condition checks and never mentions `.seq`, so a DIAL-only response has no sentence the standing class contradicts. That is why this one is unlike the DLVW case, where the boundary says in so many words that no voice file or result script was checked.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 516fef88.

`CheckDialogue` no longer captures anything itself. `DialogueSweep.Run` now takes a `Func<Binding>` — the validate lambda, the pinned seed parse and the stamp, taken together so the three cannot come off different builds — and calls it only after `named.Length == 0` has been checked. So `findings=["dialogue"]` with no `seeds=` answers `DialogueNeedsSeeds` having touched neither `Resolver` nor `Assets`, as it did before the stamp: no `Mo2LoadOrder.Build`, and no `InvalidOperationException` from the getter turning an input refusal into "an internal houseCARL failure (the arguments bound fine)" and a bare string where `format='json'` was asked for.

The refusal that stays stamped is the one reached after the binding ran (every seed unresolvable), which is the sibling split — see the reply on DialogueCheck.cs. `FactV2_TheStampDeclaresItsBound` now asserts both: `not-a-formid` carries the stamp, the unseeded call has `Epoch is null` and renders no `epoch=`.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in c8110b6c. `var assets = Assets;` is taken once inside the binding, beside `var resolver = Resolver;`, and the lambda closes over the resolved `AssetResolver` — the same shape `ScriptPropertyCheck.Run(resolver, view, Assets, …)` gets at the call site. The property's `RefreshOnProfileChange()` / `RefreshIfStale()` and its outright rebuild now run at most once per call, so every seed's `.fuz` and `.pex` verdicts come off one asset build under the one tally and the one stamp.

plugin/CHANGELOG.md
Avick3110: Folded in d8dceaee, by narrowing the sentence rather than writing the stamp into the family refusal arm.

Narrowing, because the refusal arm is not a dialogue divergence: `o.Refusal(f)` is a plain string for every family, so `findings=["errors","scripts"]` where scripts refuses writes `"refused"` and no epoch either. Stamping only dialogue there would make this family the odd one out in the arm the comment is about, and stamping all three is a change to the merged surface's refusal shape, not to this PR's subject.

The line now says: a dialogue-only call refused because every seed was unresolvable carries the bare `epoch`, as the errors and scripts families' post-index refusals do; a call with no `seeds=` is refused on the arguments alone before any build is read and names none, as theirs do not. The sentence about the full-coverage case (every seed a DLVW or DLBR) is added too.

src/housecarl-core/DialogueCheck.cs
Avick3110: Folded in 516fef88 — you are right about the split, and the code moved to it rather than only the comment.

`DialogueCheckResult.Fail` now reads: it carries a stamp only when the refusal was decided AFTER a build was read; a refusal decided on the arguments alone has no build to name, and reading one to stamp it would build the index for a call that is about to refuse. `DialogueNeedsSeeds` is now that first kind in fact, not just in wording — the binding that pins the view is called after the seed check, so the eager `Resolver` at LoadOrderService.cs:2307 is gone and the sentence no longer describes a convention the siblings do not keep. The `Epoch` parameter doc on the record says the same, and the sweep's `bind` doc names `SweepSharedInput`'s rule as the reason.

src/housecarl-mcp-tests/DialogueFamilyTests.cs
Avick3110: Folded in 516fef88 and 6e1f782d.

All four fixtures — `DialogueSweep_Run`, `RecordLevelFixture`, `RecordLevelFailingFixture`, `MixedKindFixture` — now go through one `Bind(validate)` helper that supplies the stub validator, the plugin-qualified parse and a fixed `ProbeEpoch` of a real stamp's width (16 hex characters), so every dialogue shape `CheckShapeMatrix.Build` derives carries the head line and the coverage keys, and the matrix sweeps them at every integer cap with a width that does not move between runs. The signature change made this unavoidable rather than optional: `Binding.Epoch` is non-nullable, so a fixture cannot leave it out.

That is also what exposed the DialogueSweepRender.cs:66 defect where you said it would: with the epoch threaded, `RecordLevelFixture` printed the uncovered set above the record-level boundary. The new arm `DIALOGUE-THE-STAMP-DECLARES-ONLY-THE-BOUND-THIS-RESPONSE-HAS` fails on that render, and the existing json arm now checks `epoch_covers_all_inputs` / `epoch_uncovered` fork the same way. ci-all 128/128, tests 1523/1523.

PR #572, opened 2026-09-05
Say when a read was answered off a failure-degraded order

### LINKED ISSUE #353: [GAP] a read served from a FAILURE-degraded order looks identical to one served from a legitimately changed order

Avick3110:
Say when a read was answered off a failure-degraded order

When an active plugin becomes unopenable or unparseable, the next index build drops it and every read after answers normally off the narrowed order — shallower override depth, fewer touching plugins, sometimes a different winner. The epoch stamp changes, but a legitimate reorder changes it too, so inside one response a failure-degraded order and a reordered one were indistinguishable; the only in-band signal was the refusal a caller got if they happened to name the failed plugin. Every stamped response now carries `order_degraded: true` and one sentence naming how many plugins are missing, which ones, that this is a load failure rather than a change the caller made, and that `housecarl_load_order_status` gives the reason. The text lane appends the count beside `epoch=`. It is a sibling of the epoch, never encoded into it: the epoch is opaque and compared only for equality, so folding health into the string would leave two builds that differ only in health comparing as merely "different", which is the ambiguity this fixes, re-spelled. Silent on a healthy order.

How it is carried. The epoch and the plugins a build lost are one value, `OrderStamp(Epoch, ExcludedPlugins)`, built once in the `IndexSnapshot` constructor — the single funnel every build passes through — and handed out as `IndexView.Stamp`. Every outcome type carries that stamp where it used to carry a bare epoch string, and exposes `Epoch` as a computed read through it. So the stamp sites are `Stamp = view.Stamp`, a site that tries to set an epoch without its health is a compile error rather than a lane that forgot, and the shared writers (`JsonWire.WriteEpoch` for json, `Wire.EpochLine`/`EpochInline` for text) take the fact rather than looking it up. The two sweep results already hold their own `ExcludedPlugins` roster, so they keep a bare epoch and their head lines count that roster.

`housecarl_check` states the marker at the response ROOT as well as on each swept family. The root is written from the order the whole call answered from, captured once before the families are dispatched, so a dialogue-only check says it too — that family carries no epoch by design, and the root marker is a statement about the order rather than a freshness stamp, so its no-epoch contract is unchanged.

One wording change came with it: the errors and scripts sweep heads already rendered `· N plugin(s) excluded (unparseable)` before their `epoch=`, which would now sit beside the new clause saying the same thing. Those two are replaced by the shared clause, which says `(load failure)` — the excluded set includes plugins that could not be *opened* at all, which "unparseable" never described.

Tests: a case per lane the claim covers — the record read, the scan, the check root and its errors family, a dialogue-only check, and the write lane's dry run — each with a healthy-order twin asserting silence.

`dotnet build`, `dotnet test --filter "tier!=bridge"` (1531 passed) and `ci-all` (128/128) all green.

Closes #353


src/housecarl-mcp/JsonWire.cs
```diff
@@ -1658,7 +1675,7 @@ static int MeasureMember(int depth, Action<Utf8JsonWriter> write)
     static void WriteSweepEpoch(Utf8JsonWriter w, string? epoch, IReadOnlyList<string>? offOrderScanned)
     {
         if (epoch is null) return;
-        w.WriteString("epoch", epoch);
+        WriteEpoch(w, epoch);
```
Avick3110: `housecarl_check` never carries the marker where the changelog says every response does. `WriteSweepEpoch` is only ever called from `WriteErrorsHead` (1365) and `WriteScriptsHead` (1880), both of which write into the *per-family* object under `families`, and `Compose` writes no epoch at the document root. So on a merged check against a degraded order the response has `families.errors.order_degraded` and `families.scripts.order_degraded` (the same fact, twice) and nothing at the root — a consumer following the changelog's `doc["order_degraded"]` sees nothing on the one tool where a missing plugin most directly changes the answer (fewer scanned plugins, fewer dangling refs, a smaller `scanned_plugins`).

Worse for the dialogue lane: `DialogueCheckResult` carries no epoch at all, so `CheckSweepRender.Epoch => Errors?.Epoch ?? Scripts?.Epoch` is null and `DialogueSweepRender.WriteHead` writes no stamp. A `housecarl_check` that ran only the dialogue family, against an order that lost a plugin, is silent in both lanes — exactly the failure #353 exists to end, and the PR body's "no lane can miss it" does not hold for it.

Either hoist the marker to the root of the check document beside `findings_scope` (it is a response-level fact, like the excluded roster already written there), or say in the changelog that check states it per family and give the dialogue family a stamp.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -428,9 +428,7 @@ static void AppendErrorsHead(StringBuilder sb, ErrorCheckResult r, CheckAccounti
           .Append(didDangling ? $"{r.TotalDangling} dangling ref(s)" : "dangling refs NOT CHECKED (findings= excluded 'dangling')").Append(" · ")
           .Append(didMasters ? $"{r.TotalMissingMasters} missing master(s)" : "missing masters NOT CHECKED (findings= excluded 'missing_masters')").Append(" · ")
           .Append(didDangling ? $"{r.TotalUnscannableRecords} unscannable record(s)" : "unscannable records NOT COUNTED (the record walk was skipped)");
-        if (r.ExcludedPlugins.Count > 0)
-            sb.Append(" · ").Append(r.ExcludedPlugins.Count).Append(" plugin(s) excluded (unparseable)");
-        if (r.Epoch is not null) sb.Append(" · epoch=").Append(r.Epoch).Append(EpochOffOrderQualifier(r.OffOrderScanned));
+        if (r.Epoch is not null) sb.Append(" · epoch=").Append(r.Epoch).Append(OrderHealth.ClauseFor(r.Epoch)).Append(EpochOffOrderQualifier(r.OffOrderScanned));
```
Avick3110: This replaces a fact carried **on the result** with a lookup in a bounded, process-global side table, and the two are not equivalent in the failure case. The old line printed `r.ExcludedPlugins.Count` unconditionally whenever the sweep had excluded plugins; `OrderHealth.ClauseFor(r.Epoch)` returns `""` whenever the epoch is not in `Notes` — and `Notes` is capped at 32 and evicted oldest-first (`OrderHealth.cs:48`). A miss renders a head line claiming a clean scope while `WriteExcluded` / the `ExcludedRows` section below still lists the plugins that were dropped, which is the silently-degraded render the cornerstone forbids.

The eviction window is not hypothetical during a patching session: houseCARL writes plugins, every write changes a stamp, every stamp change is a new epoch, and with one locked or unparseable plugin present every one of those builds is degraded — 33 writes is an ordinary afternoon. Any response stamped with an epoch older than the last 32 builds (a re-rendered artifact head, a scan/compare seam that carries `outcome.Epoch`) loses the clause with no trace.

The result already has the count and the roster in hand here. Suggest keeping the head line driven by `r.ExcludedPlugins` and using `OrderHealth` only for the *wording* — or, if the shared writer is the point, having `ClauseFor` take the count as a fallback so a table miss cannot turn into silence. Same line at `ReadTools.cs:848` for the scripts sweep.

src/housecarl-core/OrderHealth.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// A healthy build records nothing.</summary>
+    public static void Record(string epoch, IReadOnlyCollection<string> excludedPlugins)
+    {
+        if (excludedPlugins.Count == 0) return;
+        var entry = new Degraded(excludedPlugins.Count, Sentence(excludedPlugins));
+        lock (Gate)
+        {
+            if (!Notes.TryAdd(epoch, entry)) return;     // same build seen again — same fact, nothing to update
+            Order.Enqueue(epoch);
+            while (Order.Count > Capacity) Notes.Remove(Order.Dequeue());
```
Avick3110: The registry is a lossy cache of something the answer already has in hand.

Every stamp site gets its `epoch` from a view that also carries the excluded set: the sweep outcomes hold `ExcludedPlugins` outright (`ReadTools.cs:546`, `:775`, `:879` all read it on the same object whose `.Epoch` is being stamped two lines earlier), and every other outcome's epoch comes from the same `IndexSnapshot` that `Record` was called from. So "was this build degraded" is not only a function of the epoch — it is a field already sitting beside it. `OrderHealth` re-derives it through a process-wide, capacity-32, evicting side table, which can answer *nothing* where the view can only answer correctly.

That trade buys one thing: not touching 48 stamp sites. But the cost lands exactly where the issue does — a lookup miss is silent, and silence is the failure #353 exists to end, so the mechanism reintroduces the bug's shape at a lower rate. It also means the marker is now unavailable to any lane that has an epoch string but no snapshot, and correct only for as long as the last 32 degraded builds.

If the 48 sites are genuinely the blocker, the cheaper honest shape is to carry the excluded count on the one thing all seven outcome types already share (or thread the snapshot into the two shared writers), not to build a second source of truth that can disagree with the first by returning empty. At minimum, `NoteFor`/`ClauseFor` should not be able to answer "healthy" — a table miss on an epoch that names a real build is an unknown, not a clean bill of health, and the two are currently indistinguishable to the writer.

Worth checking on merits: two builds with the same epoch and different health cannot happen (`ComputeEpoch` folds the sorted excluded names in), and the locking is sound. Staleness is harmless for the same reason. Eviction is the only way this is wrong, and it is wrong silently.

src/housecarl-mcp-tests/DegradedOrderMarkerTests.cs
```diff
@@ -0,0 +1,71 @@
...
+        Assert.Contains("housecarl_load_order_status", note);       // where the reason is
+    }
+
+    [Fact]
+    public void ADegradedBuildsTextHeadCarriesTheClauseBesideTheEpoch()
+    {
+        var text = RecordsTools.Records(Svc, formids: new[] { Fid }, project: Eid);
+
+        Assert.Contains($"epoch={Svc.Stats().epoch} · 1 plugin(s) excluded (load failure)", text);
+    }
```
Avick3110, line 123: The tests cover one lane; the PR's claim is about all of them.

Both cases drive `RecordsTools.Records` — one outcome type of the seven, through `WriteEpoch`'s single-read top-level call. The PR body's load-bearing claim is "no lane can miss it", and that is precisely what is untested: the `housecarl_check` root and the dialogue-only check both stamp nothing today (see the `JsonWire.cs` comment), and this suite passes anyway. A test that cannot fail on the behaviour being claimed is not testing that behaviour — per `standards/TESTING.md`, assert on what the caller looks for, and here the caller is every stamped lane, not one.

A per-lane case on `CheckErrorsWorld` (or a single table-driven case over the outcome types) would have caught it before review.

Also untested: eviction. `Capacity = 32` with oldest-first removal is the one path where `ClauseFor` returns `""` for a genuinely degraded build, and nothing exercises it — a unit test that records 33 degraded epochs and asks for the first is cheap and would pin whether a miss is meant to read as healthy.

src/housecarl-core/OrderHealth.cs
Avick3110: Folded. OrderHealth.cs is deleted. The epoch and the plugins the build lost now travel as one value — OrderStamp(Epoch, ExcludedPlugins), built once in the IndexSnapshot constructor and exposed as IndexView.Stamp. The outcome types carry that stamp instead of a bare epoch string and expose Epoch as a computed read through it, so the ~48 stamp sites became 'Stamp = view.Stamp' and any site that tries to set an epoch without its health is a compile error rather than a lane that forgot. The shared writers take the stamp: JsonWire.WriteEpoch(w, stamp) for json, Wire.EpochLine/EpochInline for text. No table, no eviction, no lookup that can miss. The two sweep results were the exception you named: they already hold ExcludedPlugins, so they keep a bare Epoch and the marker is read off that roster.

src/housecarl-mcp/ReadTools.cs
Avick3110: Folded, by the first route you suggested. Both sweep head lines are back on the carried roster — OrderDegraded.Clause(r.ExcludedPlugins.Count) at ReadTools.cs:431 and :848 — so the head line and the ExcludedRows section below it are driven by the same value and cannot disagree. The wording change stands: the clause says '(load failure)', which covers the could-not-be-opened class 'unparseable' never described.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded, by the first option: the marker is hoisted to the root of the check document, beside findings_scope, written from the order the whole call answered from (CheckTools captures it once before dispatch; CheckSweep carries it as Order). The text lane states the same sentence under the scope line. That makes a dialogue-only check say it too, without giving DialogueCheckResult an epoch it cannot honestly claim — the root marker is a statement about the ORDER, not a freshness stamp, so the no-epoch note on that record still holds. The per-family heads keep their own marker beside their own epoch.

src/housecarl-mcp-tests/DegradedOrderMarkerTests.cs
Avick3110: Folded. There is now a case per lane: the record read (json and text), the scan, the check root AND its errors family, a dialogue-only check on both transports, and the write lane on a dry run — each with a healthy-order twin asserting silence. Eviction is not tested because there is no longer anything to evict: the registry is gone (see the OrderHealth.cs thread). EpochWorld gains a corpus so the scan and write lanes run against it. 1531 tests pass, ci-all 128/128.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -467,8 +467,8 @@ void Arm(string statement)
         // downstream form's epoch is compared against the deriving step's, and a divergence refuses loud rather
         // than mixing builds.
         string? expectEpoch = null;
-        string? SeamTear(string? epoch) =>
-            expectEpoch is not null && epoch is not null && epoch != expectEpoch
+        string? SeamTear(OrderStamp? epoch) =>
+            expectEpoch is not null && epoch is not null && epoch.Epoch != expectEpoch
                 ? $"the load order changed between deriving the selection (epoch={expectEpoch}) and reading it (epoch={epoch}) — the two halves would mix builds. Retry the call."
```
Avick3110: `SeamTear` prints the OrderStamp object where the epoch belongs. The parameter changed from `string?` to `OrderStamp?`, and the comparison was updated to `epoch.Epoch != expectEpoch`, but the message still interpolates `{epoch}` — which now renders the record's `ToString()`.

A `records` call with `walk=` that straddles a load-order change refuses with:

```
the load order changed between deriving the selection (epoch=e2-abc123) and reading it (epoch=OrderStamp { Epoch = e2-def456, ExcludedPlugins = System.String[] }) — the two halves would mix builds. Retry the call.
```

Every sibling seam-tear in this file (1134, 1200, 1214, 1234, 1311, 1484, 1498) was converted to `.Epoch`; the compiler can't catch this one because interpolation accepts `object`.

```suggestion
                ? $"the load order changed between deriving the selection (epoch={expectEpoch}) and reading it (epoch={epoch.Epoch}) — the two halves would mix builds. Retry the call."
```

src/housecarl-mcp/JsonWire.cs
```diff
@@ -1655,10 +1684,11 @@ static int MeasureMember(int depth, Action<Utf8JsonWriter> write)
     /// off-order files were swept beside the index, since their content is outside the fingerprint and
     /// <c>off_order_scanned</c> names them. Both swept families have that lane, so both write it through here.
     /// Success path only; a refusal swept nothing and carries the bare stamp.</summary>
-    static void WriteSweepEpoch(Utf8JsonWriter w, string? epoch, IReadOnlyList<string>? offOrderScanned)
+    static void WriteSweepEpoch(Utf8JsonWriter w, string? epoch, IReadOnlyDictionary<string, string> excluded,
+                                IReadOnlyList<string>? offOrderScanned)
     {
         if (epoch is null) return;
```
Avick3110, line 1690: On the json check surface the full ~250-character sentence lands three times in one document: at the root (`Compose`, line 1844), in the errors head through here, and in the scripts head through here again — beside the excluded-plugin roster, which already names the same plugins *with their reasons*.

That is ~750 characters of fixed part, and the fixed part is subtracted from the row budget, so at a given `max_chars` fewer findings fit. The PR's own `ScriptsFamilyTests` band is the evidence: it had to be widened 4200 → 5000 to still reach the uncapped case.

The root marker earns its place — it is the only lane a dialogue-only check has. The family heads already say the count in-band (`epoch=… · N plugin(s) excluded (load failure)`), so the sentence there is the third telling of a fact the reader has twice already. Consider `WriteSweepEpoch` writing the boolean + count only, and leaving the sentence to the root.

src/housecarl-mcp/CheckTools.cs
```diff
@@ -167,6 +167,10 @@ public static string CheckTool(
         // that split, so the default findings set does not read the MO2 composition and sweep every mod folder twice
         // for an answer that is identical both times — and the two cannot disagree about which names resolved.
         var offOrderMemo = new SweepOffOrderMemo();
+        // The order every family below answers from, captured once so the response root can say whether it had lost
+        // plugins — the dialogue family carries no epoch, so this is the only lane that can state it for a
+        // dialogue-only call (#353).
+        var order = svc.CaptureView().Stamp;
```
Avick3110, line 173: The root stamp is its own capture; each family captures again below. The PR body says the root is "the order the whole call answered from", but nothing holds the two together — a freshness rebuild between this line and `svc.CheckErrors(...)` gives the families a different build.

The response then carries `order_degraded` computed from build A beside `epoch=<B>` on the errors family: it names plugins the answering build did not lose, or stays silent about ones it did. Narrow, but it is the exact ambiguity #353 exists to end, and every other multi-step lane in the codebase compares the two epochs and refuses loud instead.

Passing this view down to the families (or comparing `order.Epoch` against `sweep.Epoch` and refusing on a divergence, like the records seams do) would close it.

src/housecarl-core/OrderStamp.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// <summary>What the caller needs to know in one sentence: how many plugins are missing, which ones, that this
+    /// is a FAILURE rather than something they did, and where the reason is.</summary>
+    public static string Sentence(IReadOnlyCollection<string> excludedPlugins)
+    {
+        var names = excludedPlugins.OrderBy(n => n, StringComparer.OrdinalIgnoreCase).ToList();
+        var shown = string.Join(", ", names.Take(NamesShown));
+        if (excludedPlugins.Count > NamesShown) shown += $", and {excludedPlugins.Count - NamesShown} more";
+        return $"{excludedPlugins.Count} plugin(s) could not be loaded for this build and are absent from the order " +
+               $"this answer describes ({shown}) — this is a load FAILURE, not a change you made; " +
```
Avick3110: "not a change you made" is asserted unconditionally, but the resolver excludes a plugin for two reasons (`LoadOrderResolver.Exclude` call sites, 602 and 636) and only one of them fits the claim:

- `contains a record Mutagen cannot parse` — true, nothing the caller did.
- `could not be opened` — most often exactly what the caller did: disabled or removed the mod in MO2 while the plugin stayed ticked in plugins.txt.

For the second case the sentence steers the reader away from the real cause, which is the opposite of what the marker is for. The snapshot already keeps the two apart (`Unopenable` is its own set precisely because the distinction is load-bearing), so this can name which without a new fact — or just drop the "not a change you made" half and let `housecarl_load_order_status` carry the reason, which the sentence already points at.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, fixed in 6e462cb. The interpolation now reads `{epoch.Epoch}`, so the refusal prints the epoch string on both halves of the seam. Confirmed the other seven seam-tears in the file were already on `.Epoch`.

src/housecarl-core/OrderStamp.cs
Avick3110: Real, fixed in 52ed1e5. The clause is now "this is a load FAILURE, not a reorder; housecarl_load_order_status gives the reason for each" — it still separates a failure from a legitimate reorder, which is the distinction the marker exists for, and stops asserting the caller did not cause it. The doc comment and the changelog now say why: the parse class is nothing the caller did, the open class usually is (mod off, plugins.txt still listing it), and which applies is per plugin, so the status tool answers it. Test pins the absent phrase.

src/housecarl-mcp/CheckTools.cs
Avick3110: Real, fixed in f90c836 by the second remedy you name. `check` now compares the captured stamp against every family that stamped one, and a divergence refuses the whole call: "the load order changed while this check was running (epoch=A when the call started, epoch=B when the errors family answered) — the response would describe two builds. Retry the call." The ground rides a new `OrderSeamError` on `CheckSweep` rather than `SharedInputError`, which is documented as decided before dispatch; `CheckOutcome` short-circuits the ground collapse on it, so no section is emitted and the refusal carries no marker. It also closes the errors-vs-scripts case, which `CheckSweep.Epoch` and the roster picker were already assuming — both comments now say the assumption is enforced. New fact E9 in EpochCheckSweepTests.

src/housecarl-mcp/JsonWire.cs
Avick3110: Real, folded in 14f836b. `WriteSweepEpoch` now writes `order_degraded: true` and `order_degraded_plugins: N` and leaves the sentence to the root, so the ~250 characters land once per document instead of three times, and the family json now says what the family text head says. The family heads are only ever written inside the merged document (the two call sites are both in `Compose`), so the root sentence is always there to carry it — and after f90c836 a call whose families read a different build refuses, so the root fact really is the families' fact. The `ScriptsFamilyTests` band is back to 300/3600/3900/4200 with the fixed part shrunk; full suite 1532 passed, ci-all 128/128.

PR #573, opened 2026-09-05
Walk the reverse direction to any depth

### LINKED ISSUE #489: [GAP] Reverse-reference index: unbounded and transitive "who references X" — built lazily, kept current per plugin (E1's engine half)

Avick3110:
`walk={"direction":"reverse","depth":N}` was refused past depth 1, naming the reverse-reference index as the thing that would answer it. That index shipped in #562, so the refusal is deleted and the walk answers instead. `depth` means the same thing it means everywhere — the follow rule repeated at each hop.

**`walk.follow` picks the reverse walk; the form only picks the view.** `follow` names the edges a walk crosses, which is as meaningful backwards as forwards, so the surface's own "follow shapes a FORWARD expansion" refusal narrows to `seed_paths`/`exclusions`. On reverse:

- `follow` unset or `"*"` — the transitive walk: every link, at every hop, off the index, with no bounding `types=`/`plugins=` scope.
- `follow="Effects[].BaseEffect"` (SPEC §3.3's spelling) — the typed MGEF carrier walk: per-carrier magnitude/area/duration, `types=` narrowing the carrier types. It crosses the effect link at every hop and a carrier is not a magic effect, so it reaches nothing past hop 1 — a consequence of the follow the caller named, not a special case.
- anything else — refused, naming the two the index can serve (it holds every link and no per-path breakdown).

The reached set is then the same under every form for a given follow, and both walks hand their reached set to the same two views: `form='chain'` renders the walk's rows, a reading form consumes the reached set. The carrier walk under a reading form is new here — it used to be chain-only.

**No default follow implies a walk, under any form.** `form='chain'` on a reverse walk with `follow` unset refuses in one sentence naming the two follows the index can serve — `"Effects[].BaseEffect"` for the typed carrier walk, which chain renders, and `"*"` for every link, which chain cannot render because the transitive walk expands one shared frontier and has no per-seed path to draw — and points that reading at a reading form. `form='chain'` with an explicit `follow="*"` keeps its own refusal. So a reverse chain call says its follow outright, including the one the retired `housecarl_effect_chain` redirect teaches, which now carries the token. Aaron's ruling of 2026-09-06: a token means one thing under every form, and the earlier draft's carry-the-old-spelling default is gone.

`walk.max_nodes` now says which reading it spends: per seed on the carrier walk, one budget shared across every seed and every hop on the transitive walk.

**The reached set is verified, not raw candidates.** The index answers in candidates — some plugin's copy carries the link. `references=` re-tests each candidate against the body it judges; so does this walk now, against the load-order winner, and a candidate that fails is neither reported nor expanded, so the two spellings agree and a false hop-1 node cannot seed a false subtree. The response says how many candidates the check dropped.

**The hop census is honest.** The node budget is tested before the node is consumed, so a raised budget on a retry sees the same graph; the hop the cut landed on is marked `(cut by walk.max_nodes)` instead of reading as `hop N: 0`; a capped walk is never also called exhausted; the carrier walk's hop-1 count is the census total, not the rows written; and the empty-hop prose is told only when a depth was actually asked for, never off the default 16. Seeds are deduplicated, so two spellings of one key are one seed.

The traversal itself is `ReverseSelection.Transitive` beside the index it reads, and the lane's batch is a new file rather than another domain in `LoadOrderService.cs`. No new parameter and no new token.

Tests: reverse depth 2 reaches a second-hop referrer and depth 1 does not; an exhausted walk names the hops it did not walk; a capped walk names the cut hop and is not called exhausted; a candidate whose winner dropped the link is excluded by both `references=` and the walk, and the drop is reported; repeated seeds are deduplicated; `follow="*"` is the transitive walk and `follow="Effects[].BaseEffect"` the carrier walk under a reading form; an unservable follow, chain-over-transitive and chain-with-follow-unset all refuse naming the alternatives, and the same chain call with the carrier follow said outright is served; a bad direction no longer teaches the deleted depth-1 rule; a carrier call with no depth prints no hop census. The fixture gains a three-link chain (MgefHop <- SpellHop <- ListHop) and a list whose winning override drops its link.

Closes #489

### DIFF
--- a/src/housecarl-core/ReverseReferenceIndex.cs
+++ b/src/housecarl-core/ReverseReferenceIndex.cs
@@ -336,6 +336,41 @@ public static class ReverseSelection
         return index.Orphans(view.RecordKeys());
     }
 
+    /// <summary>One hop of a transitive reverse walk: the records reached at this distance from the seeds, first
+    /// arrival wins. An EMPTY hop is a fact and is kept in the list, so a walk that ran out of referrers says so
+    /// rather than trailing off.</summary>
+    public sealed record Hop(int Depth, IReadOnlyList<FormKey> Reached);
+
+    /// <summary>The transitive reverse walk: who references the seeds, then who references those, hop after hop.
+    /// The follow rule — every link — is the same at every hop, which is what <c>depth</c> means here and
+    /// everywhere. Records already reached are not re-reported and not re-expanded, so a reference cycle
+    /// terminates. <paramref name="maxNodes"/> bounds the TOTAL reached across all hops; a breach keeps what was
+    /// proved, stops there, and sets <paramref name="capped"/> so the response can say the cut happened.</summary>
+    public static IReadOnlyList<Hop> Transitive(ReverseReferenceIndex index, IReadOnlyList<FormKey> seeds,
+                                                int depth, int maxNodes, out bool capped)
+    {
+        capped = false;
+        var hops = new List<Hop>();
+        var visited = new HashSet<FormKey>(seeds);
+        IReadOnlyList<FormKey> frontier = seeds;
+        int reached = 0;
+        for (int d = 1; d <= depth; d++)
+        {
+            var next = new List<FormKey>();
+            foreach (var k in index.ReferencersOf(frontier))
+            {
+                if (!visited.Add(k)) continue;
+                if (reached >= maxNodes) { capped = true; break; }
+                next.Add(k);
+                reached++;
+            }
+            hops.Add(new Hop(d, next));
+            if (capped || next.Count == 0) break;
+            frontier = next;
+        }
+        return hops;
+    }
+
     /// <summary>The sentence a caller gets for the negated-only unbounded form: its universe is the orphan set, not
     /// the whole order and not the same question a bounded negated <c>references=</c> asks. Declared, never
     /// discovered.</summary>
--- a/src/housecarl-core/ReverseReferenceIndex.cs
+++ b/src/housecarl-core/ReverseReferenceIndex.cs
@@ -348,8 +348,9 @@ public static class ReverseSelection
     /// terminates. The index answers in CANDIDATES, so <paramref name="verify"/> re-tests each one against the body
     /// the caller means to judge — the same second step <c>references=</c> takes — and a candidate that fails it is
     /// neither reported nor expanded, so a dropped link cannot seed a false subtree. <paramref name="maxNodes"/>
-    /// bounds the TOTAL reached across all hops; the budget is tested BEFORE the node is consumed, so a raised
-    /// budget on a retry sees the same graph, and the hop the cut landed on is marked
+    /// bounds the TOTAL reached across all hops; the budget is tested BEFORE the candidate is verified or
+    /// consumed, so a spent budget bounds the body reads as well as the reach, a raised budget on a retry sees
+    /// the same graph, and the hop the cut landed on is marked
     /// <see cref="Hop.Cut"/> rather than reading as a hop that reached nothing.</summary>
     public static IReadOnlyList<Hop> Transitive(ReverseReferenceIndex index, IReadOnlyList<FormKey> seeds,
                                                 int depth, int maxNodes,
@@ -368,8 +369,10 @@ public static class ReverseSelection
             foreach (var k in index.ReferencersOf(frontier))
             {
                 if (visited.Contains(k)) continue;
-                if (verify is not null && !verify(k, frontierSet)) continue;
+                // The budget is spent before the candidate is verified, so a spent budget stops the body reads
+                // too — verification has no ordering effect, so a raised budget on a retry still sees this graph.
                 if (reached >= maxNodes) { capped = true; cut = true; break; }
+                if (verify is not null && !verify(k, frontierSet)) continue;
                 visited.Add(k);
                 next.Add(k);
                 reached++;
--- a/src/housecarl-core/ReverseReferenceIndex.cs
+++ b/src/housecarl-core/ReverseReferenceIndex.cs
@@ -338,16 +338,22 @@ public static class ReverseSelection
 
     /// <summary>One hop of a transitive reverse walk: the records reached at this distance from the seeds, first
     /// arrival wins. An EMPTY hop is a fact and is kept in the list, so a walk that ran out of referrers says so
-    /// rather than trailing off.</summary>
-    public sealed record Hop(int Depth, IReadOnlyList<FormKey> Reached);
+    /// rather than trailing off — unless <paramref name="Cut"/>, which says the node budget ended this hop and its
+    /// count is a prefix, not a finding.</summary>
+    public sealed record Hop(int Depth, IReadOnlyList<FormKey> Reached, bool Cut);
 
     /// <summary>The transitive reverse walk: who references the seeds, then who references those, hop after hop.
     /// The follow rule — every link — is the same at every hop, which is what <c>depth</c> means here and
     /// everywhere. Records already reached are not re-reported and not re-expanded, so a reference cycle
-    /// terminates. <paramref name="maxNodes"/> bounds the TOTAL reached across all hops; a breach keeps what was
-    /// proved, stops there, and sets <paramref name="capped"/> so the response can say the cut happened.</summary>
+    /// terminates. The index answers in CANDIDATES, so <paramref name="verify"/> re-tests each one against the body
+    /// the caller means to judge — the same second step <c>references=</c> takes — and a candidate that fails it is
+    /// neither reported nor expanded, so a dropped link cannot seed a false subtree. <paramref name="maxNodes"/>
+    /// bounds the TOTAL reached across all hops; the budget is tested BEFORE the node is consumed, so a raised
+    /// budget on a retry sees the same graph, and the hop the cut landed on is marked
+    /// <see cref="Hop.Cut"/> rather than reading as a hop that reached nothing.</summary>
     public static IReadOnlyList<Hop> Transitive(ReverseReferenceIndex index, IReadOnlyList<FormKey> seeds,
-                                                int depth, int maxNodes, out bool capped)
+                                                int depth, int maxNodes,
+                                                Func<FormKey, IReadOnlySet<FormKey>, bool>? verify, out bool capped)
     {
         capped = false;
         var hops = new List<Hop>();
@@ -356,15 +362,19 @@ public static class ReverseSelection
         int reached = 0;
         for (int d = 1; d <= depth; d++)
         {
+            var frontierSet = new HashSet<FormKey>(frontier);
             var next = new List<FormKey>();
+            bool cut = false;
             foreach (var k in index.ReferencersOf(frontier))
             {
-                if (!visited.Add(k)) continue;
-                if (reached >= maxNodes) { capped = true; break; }
+                if (visited.Contains(k)) continue;
+                if (verify is not null && !verify(k, frontierSet)) continue;
+                if (reached >= maxNodes) { capped = true; cut = true; break; }
+                visited.Add(k);
                 next.Add(k);
                 reached++;
             }
-            hops.Add(new Hop(d, next));
+            hops.Add(new Hop(d, next, cut));
             if (capped || next.Count == 0) break;
             frontier = next;
         }
--- a/src/housecarl-mcp/RecordsTools.cs
+++ b/src/housecarl-mcp/RecordsTools.cs
@@ -55,16 +55,16 @@ public static class RecordsTools
         [Description("Link-bearing field paths that start the walk from each seed, e.g. [\"HeadParts\", \"WornArmor\"]. '*parent' crosses the containment edge instead — the record that CONTAINS the seed (a REFR from a crash log to its CELL; '*parent.*parent' to the worldspace). Omit for every link on the seed.")]
         public string[]? seed_paths { get; set; }
 
-        [Description("The link path followed at every LATER hop. \"*\" (default) walks every link — full closure. A named path restricts to one chain, e.g. \"Template\" for NPC template inheritance, or \"*parent\" to climb containment.")]
+        [Description("The link path followed at every LATER hop — the edges this walk crosses, in either direction. \"*\" (default) walks every link — full closure. A named path restricts to one chain, e.g. \"Template\" for NPC template inheritance, or \"*parent\" to climb containment. On direction='reverse' the two legal values are \"*\" (unset; the transitive walk over every link, off the reverse-reference index) and \"Effects[].BaseEffect\" (the typed MGEF carrier walk).")]
         public string? follow { get; set; }
 
-        [Description("'forward' (default) — what the seeds point AT (cheap: each hop is one link resolve). 'reverse' — what points AT the seeds, at any depth, needing no bounding scope: it follows EVERY link at every hop, off the reverse-reference index, which is built on the first such call at the cost of one whole-order link-walk and reports that cost in the response. The FORM picks the reverse lane, as it does forward: a reading form (summary/fields/rows/everything/aggregate) consumes the reached set and the response names the count at each hop, empty hops included; form='chain' is the typed MGEF carrier lane instead — magic-effect seeds get per-carrier magnitude/area/duration (types= narrows the carrier types; walk.max_nodes bounds each seed's carrier rows; limit=/offset= window the SEEDS), and it reaches nothing past hop 1 because a carrier is not a magic effect. references= is the same reverse question as one step of SELECT.")]
+        [Description("'forward' (default) — what the seeds point AT (cheap: each hop is one link resolve). 'reverse' — what points AT the seeds, at any depth, needing no bounding scope. walk.follow picks the reverse walk, exactly as it does forward: unset or \"*\" follows EVERY link at every hop off the reverse-reference index, which is built on the first such call at the cost of one whole-order link-walk and reports that cost in the response; follow=\"Effects[].BaseEffect\" is the typed MGEF carrier walk — magic-effect seeds, per-carrier magnitude/area/duration, types= narrowing the carrier types — which reaches nothing past hop 1 because a carrier is not a magic effect. The FORM then only picks the view: 'chain' renders the walk's own rows, a reading form (summary/fields/rows/everything/aggregate) consumes the same reached set. One exception, declared: form='chain' with follow unset means the carrier walk, since the transitive walk has no per-seed path render. references= is the same reverse question as one step of SELECT.")]
         public string? direction { get; set; }
 
         [Description("Maximum hops from a seed (default 16). Nodes AT the cap are recorded, not entered, and the response says the cap cut the walk — never a silent stop.")]
         public int? depth { get; set; }
 
-        [Description("Maximum nodes reached per seed (default 2000, the read-expansion budget). A breach keeps what was proved and says so.")]
+        [Description("The node budget (default 2000, the read-expansion budget). Per seed on a forward walk and on the reverse carrier walk (each seed's carrier rows); ONE budget shared across every seed and every hop on the transitive reverse walk, whose hops are one frontier and not a per-seed expansion. A breach keeps what was proved and says which reading it spent.")]
         public int? max_nodes { get; set; }
 
         [Description("Node classes the walk must not enter, as data: [{\"match\": \"Race\", \"severity\": \"stop\"|\"refuse\"}] — match is the record type name a read reports; stop prunes there (recorded as a boundary), refuse fails the whole call loud.")]
@@ -324,17 +324,22 @@ public static class RecordsTools
         // ---- walk= (the traversal construct) ----
         string walkDirection = "forward";
         int walkDepth = 16, walkMaxNodes = 2000;
+        bool walkDepthAsked = false;
+        // The reverse carrier walk's follow, spelled as SPEC §3.3 spells it. It is the discriminator between the
+        // two reverse walks, so it is one constant read by the validation, the lane and every remedy sentence.
+        const string CarrierFollow = "Effects[].BaseEffect";
         var walkExclusions = new List<(string Match, bool Refuse)>();
         if (walk is not null)
         {
             var dir = walk.direction?.Trim().ToLowerInvariant();
             if (dir is not (null or "" or "forward" or "reverse"))
-                return Wire.Refuse(json, $"error: walk.direction='{walk.direction}' — use 'forward' (what the seeds point at) or 'reverse' (what points at them; depth 1).");
+                return Wire.Refuse(json, $"error: walk.direction='{walk.direction}' — use 'forward' (what the seeds point at) or 'reverse' (what points at them, at any depth; walk.follow picks which reverse edges: \"*\" for every link, \"{CarrierFollow}\" for the typed MGEF carriers).");
             if (!string.IsNullOrEmpty(dir)) walkDirection = dir!;
             if (walk.depth is { } wd)
             {
                 if (wd < 1) return Wire.Refuse(json, $"error: walk.depth={wd} — depth must be >= 1 (hops from the seed).");
                 walkDepth = wd;
+                walkDepthAsked = true;
             }
             if (walk.max_nodes is { } wn)
             {
@@ -352,8 +357,15 @@ public static class RecordsTools
             }
             if (walkDirection == "reverse")
             {
-                if (walk.seed_paths is { Length: > 0 } || walk.exclusions is { Length: > 0 } || walk.follow is not null)
-                    return Wire.Refuse(json, "error: walk.seed_paths/follow/exclusions shape a FORWARD expansion — a reverse walk scans TOWARD the seeds and follows every link at every hop. Drop them.");
+                // seed_paths and exclusions really do shape a forward expansion; follow does not — it names which
+                // edges the walk crosses, which is as meaningful backwards as forwards, so it stays legal here and
+                // is what tells the two reverse walks apart.
+                if (walk.seed_paths is { Length: > 0 } || walk.exclusions is { Length: > 0 })
+                    return Wire.Refuse(json, "error: walk.seed_paths/exclusions shape a FORWARD expansion — a reverse walk scans TOWARD the seeds. Drop them (walk.follow stays: it names the edges this walk crosses).");
+                var revFollow = walk.follow?.Trim();
+                if (!string.IsNullOrEmpty(revFollow) && revFollow != "*"
+                    && !string.Equals(revFollow, CarrierFollow, StringComparison.OrdinalIgnoreCase))
+                    return Wire.Refuse(json, $"error: walk.follow='{walk.follow}' on a reverse walk — the reverse edges come from the reverse-reference index, which holds every link and no per-path breakdown, so the two follows it can serve are \"*\" (every link, the transitive walk) and \"{CarrierFollow}\" (the typed MGEF carriers). To narrow the transitive walk's reach by path, walk with to_file= and re-enter the artifact on a where= over the field.");
             }
             if (references is { Length: > 0 })
                 return Wire.Refuse(json, "error: walk= and references= are the same construct (references= IS the reverse walk at depth 1) — use one spelling per call.");
@@ -365,7 +377,7 @@ public static class RecordsTools
                 return Wire.Refuse(json, "error: walk= composed with where= (filtering the reached set by predicate) — walk with to_file=, then re-enter the artifact on a bounded scan via where=[\"formid in @<file>\", …]; the reached set becomes the scan's identity list.");
         }
         if (form == "chain" && walk is null)
-            return Wire.Refuse(json, "error: the 'chain' form renders a walk's paths — pass walk= (e.g. walk={\"follow\": \"Template\"} over NPC seeds; reverse MGEF carriers: walk={\"direction\": \"reverse\"} with MGEF formids=).");
+            return Wire.Refuse(json, $"error: the 'chain' form renders a walk's paths — pass walk= (e.g. walk={{\"follow\": \"Template\"}} over NPC seeds; reverse MGEF carriers: walk={{\"direction\": \"reverse\", \"follow\": \"{CarrierFollow}\"}} with MGEF formids=).");
 
         // The existing single-pole lanes below drive off the named-pole fields; the richer specs dispatch to the
         // comparison/overlay lanes before reaching them.
@@ -401,14 +413,21 @@ public static class RecordsTools
         // formids= composes with the scan terms: the identity set intersects the scan's selection, or is the
         // universe when it is the only bound. The reverse MGEF walk keeps its own lane, where formids are seeds.
         bool reverseWalk = walk is not null && walkDirection == "reverse";
-        // The reverse direction has two lanes and they are told apart by the FORM, exactly as the forward direction
-        // is: 'chain' is the typed MGEF carrier lane (per-carrier magnitude/area/duration), every reading form
-        // consumes the reached set of the transitive reverse walk off the reverse-reference index.
-        bool reverseCarrier = reverseWalk && form == "chain";
+        // The reverse direction has two walks and walk.follow tells them apart, exactly as it does forward: "*" (or
+        // unset) is the transitive walk over every link off the reverse-reference index, the carrier path is the
+        // typed MGEF walk. The FORM then only picks the view. One declared exception keeps the shipped carrier
+        // spelling working: form='chain' with follow unset means the carrier walk, because the transitive walk —
+        // one shared frontier, not a per-seed expansion — has no per-seed path render for chain to draw.
+        bool followAsked = !string.IsNullOrWhiteSpace(walk?.follow);
+        bool reverseCarrier = reverseWalk
+            && (string.Equals(walk!.follow?.Trim(), CarrierFollow, StringComparison.OrdinalIgnoreCase)
+                || (!followAsked && form == "chain"));
+        if (reverseWalk && form == "chain" && !reverseCarrier)
+            return Wire.Refuse(json, $"error: the 'chain' form renders a walk's own per-seed paths, and the transitive reverse walk (follow=\"*\") expands ONE shared frontier rather than a path per seed, so it has none to draw — read it with a reading form (summary/fields/rows/everything/aggregate) over the reached set, or ask for the typed MGEF carrier chain with walk.follow=\"{CarrierFollow}\".");
         if (reverseWalk && (plugins?.names is { Length: > 0 } || conflicts_only || where is { Length: > 0 }))
             return Wire.Refuse(json, "error: a reverse walk takes formids= (its seeds) and nothing else as a scope — the bounded reverse over other scan terms is the references= spelling.");
         if (reverseWalk && !reverseCarrier && types is { Length: > 0 })
-            return Wire.Refuse(json, "error: types= narrows the CARRIER types of the typed MGEF lane (form='chain'), and this reverse walk reaches records of every type — walk with to_file= and re-enter the artifact via formids=[\"@<file>\"] with types= to keep only the types you want.");
+            return Wire.Refuse(json, $"error: types= narrows the CARRIER types of the typed MGEF walk (walk.follow=\"{CarrierFollow}\"), and this reverse walk follows every link and reaches records of every type — walk with to_file= and re-enter the artifact via formids=[\"@<file>\"] with types= to keep only the types you want.");
         if (reverseWalk && !hasFormids)
             return Wire.Refuse(json, reverseCarrier
                 ? "error: the reverse walk needs its seeds — pass formids= (the MGEF(s) whose carriers to trace)."
@@ -669,9 +688,9 @@ public static class RecordsTools
 
         // ================================================================================================
         //  WALK lane — forward walks expand the winner link graph (the chain form renders the paths; any other
-        //  form consumes the reached set as its selection). Reverse splits on the same form line: chain is the
-        //  typed MGEF lane tracing carriers with their per-hit payload, every reading form is the transitive
-        //  reverse walk off the reverse-reference index.
+        //  form consumes the reached set as its selection). Reverse splits on walk.follow, not on the form: "*" is
+        //  the transitive walk off the reverse-reference index, the carrier path is the typed MGEF walk, and each
+        //  hands its reached set to the same two views.
         // ================================================================================================
         string WalkLane(string[] ids, HousecarlCore.ArtifactDemand? demand, string? echoSrc)
         {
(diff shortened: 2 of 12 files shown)


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -682,7 +687,35 @@ List<KeyValuePair<string, string>> Echo()
...
+                var rev = ReverseWalkBatch.Run(svc, ids, walkDepth, walkMaxNodes, demand);
+                if (rev.Refusal is not null)
+                    return json ? JsonWire.RenderError(rev.Refusal, rev.Epoch) : "error: " + rev.Refusal + (rev.Epoch is not null ? $"\nepoch={rev.Epoch}" : "");
+                if (SeamTear(rev.Epoch) is { } rTear)
+                    return json ? JsonWire.RenderError(rTear, rev.Epoch) : "error: " + rTear;
+                // Every hop is named with its count, empty ones included: a walk that ran out of referrers says so
+                // on the response instead of trailing off, so an empty hop 2 is visible and not silent.
+                var hopLine = string.Join(", ", rev.Hops.Select(h => $"hop {h.Depth}: {h.Reached.Count}"));
+                if (rev.Hops.Count < walkDepth)
+                    hopLine += $" (nothing left to expand, so hops {rev.Hops.Count + 1}–{walkDepth} were not walked)";
```
Avick3110, line 726: A walk cut by `walk.max_nodes` claims it was exhausted. `rev.Hops.Count < walkDepth` is true for a capped walk too, so the "nothing left to expand" sentence fires alongside the `[!] budget was reached` banner, and the response asserts both.

Verified against the branch — `formids=[MgefHop], walk={"direction":"reverse","depth":6,"max_nodes":1}`:

```
walk=reverse (every link) depth=6: hop 1: 1, hop 2: 0 (nothing left to expand, so hops 3–6 were not walked) — selection = 2 record(s) (1 referrer(s), seeds included)
[!] the walk.max_nodes budget (1) was reached — what is listed IS reached and proved; raise walk.max_nodes to walk further.
```

ListHop *is* a hop-2 referrer; it was cut by the budget, not absent. The exhaustion clause should be gated on `!rev.Capped`.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -336,6 +336,41 @@ public static IReadOnlyList<FormKey> Universe(LoadOrderResolver.IndexView view,
...
+        var visited = new HashSet<FormKey>(seeds);
+        IReadOnlyList<FormKey> frontier = seeds;
+        int reached = 0;
+        for (int d = 1; d <= depth; d++)
+        {
+            var next = new List<FormKey>();
+            foreach (var k in index.ReferencersOf(frontier))
+            {
+                if (!visited.Add(k)) continue;
+                if (reached >= maxNodes) { capped = true; break; }
```
Avick3110: The budget check sits *after* `visited.Add(k)`, so the node that trips the cap is marked visited, dropped from `next`, and never reported — and because the break leaves the current hop with the nodes it had gathered so far, a hop the budget cut is emitted as `hop N: 0`, indistinguishable from a hop that genuinely reached nothing.

Concretely, when the referrer count at hop 1 exactly equals `maxNodes`: hop 1 fills to the budget and ends normally (`capped` still false), then hop 2 trips the cap on its first candidate and is recorded as `Hop(2, [])`. The response then prints `hop 2: 0` for a hop that has referrers. The PR's own premise is that an empty hop is a fact; here it is an artifact of the cut.

Two fixes worth making together: test the budget before consuming the node into `visited` (so a raised `max_nodes` on a retry sees the same graph), and do not append a hop that was terminated by the cap — or mark it so the render can say "cut" rather than "0".

src/housecarl-mcp/ReverseWalkBatch.cs
```diff
@@ -0,0 +1,54 @@
...
+                return new Result(Array.Empty<ReverseSelection.Hop>(), Array.Empty<string>(), false, null, epoch,
+                                  $"bad FormID '{raw}': {ex.Message} — every seed of a reverse walk must parse before the walk starts.");
+            }
+        }
+
+        var built = view.EnsureReverseIndex();
+        var hops = ReverseSelection.Transitive(view.ReverseIndex!, seedKeys, depth, maxNodes, out var capped);
+
+        // Seeds first, then each hop in order: the selection reads in walk order, and the render says the seeds
+        // are in it.
```
Avick3110, line 114: The reached set is the index's raw output, and the index is documented as a **candidate** set, not an answer. `ReverseReferenceIndex.ReferencersOf` says so ("The answer is a CANDIDATE set: the index says some plugin's copy of the record carries the link, and the caller still decides on the body it means to judge (a later override may have dropped it)"), and `ReverseSelection.Universe` says the same for the lane that already ships: "the scan that follows is the existing one, so the verdict still comes from the body the caller means to judge and a later override that dropped the link is not counted."

`references=` honours that — the index supplies the universe and the ordinary body scan re-verifies each candidate. This lane skips the verify: `Transitive`'s output goes straight into `Selection` and out through `ListLane`. So for a seed X referenced by a record R only in a *losing* plugin — the winning override having dropped the link — `references=[X]` correctly excludes R, while `walk={"direction":"reverse","depth":1}` lists it. Two spellings of the same question, described in the PR as "the same construct", give different answers, and the walk's is the wrong one with nothing said.

It compounds transitively: a false hop-1 node seeds a whole false subtree at hops 2+.

Either re-verify each hop's candidates against the body the caller judges (the winner, or `source=`), or say in the response that this lane's reach is unverified candidate edges — but it cannot be silently different from `references=`.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -714,11 +747,18 @@ List<KeyValuePair<string, string>> Echo()
...
-                envelope.Add(new("walk", "reverse, depth 1 — the typed MGEF carrier lane"));
-                headerLine += "\nwalk=reverse (per seed: every SPEL/ENCH/ALCH/SCRL/INGR applying it, with the MATCHING entry's magnitude/area/duration — reported AS AUTHORED; conditions are not evaluated, so a row means 'defines it at this strength', not 'it will fire')";
                 // The census separates written rows from the true total and names capped seeds, so the artifact
                 // and its census cannot disagree and a walk.max_nodes cut is always declared.
                 int carrierRows = results.Sum(r => r.Result.Error is null ? r.Result.Rows.Count : 0);
+                // This lane follows the effect link at every hop, and a carrier is not a magic effect — so hop 2
+                // and beyond reach nothing. Said out loud when a depth was asked for, never a silent stop.
+                var carrierHops = $"hop 1: {carrierRows}";
+                if (walkDepth > 1)
+                    carrierHops += string.Concat(Enumerable.Range(2, walkDepth - 1).Select(d => $", hop {d}: 0"))
```
Avick3110: `walkDepth` defaults to **16**, not 1, so `walkDepth > 1` is true on every carrier-lane call that did not pass a depth — the comment above says "Said out loud when a depth was asked for", but the code cannot tell a default from an ask.

Verified on the branch: the plain, previously-clean `formids=[MGEF], walk={"direction":"reverse"}, form='chain'` call now renders

```
hop 1: 2, hop 2: 0, hop 3: 0, hop 4: 0, hop 5: 0, hop 6: 0, hop 7: 0, hop 8: 0, hop 9: 0, hop 10: 0, hop 11: 0, hop 12: 0, hop 13: 0, hop 14: 0, hop 15: 0, hop 16: 0 — the carrier lane follows the effect link at every hop and a carrier is not a magic effect, ...
```

in both the header and the envelope. That is fifteen zero hops and a paragraph of prose answering a question the caller never asked, on the surface's most common reverse call. Gate it on an explicit `walk.depth` (keep the parsed value alongside a flag), not on `walkDepth > 1`.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -714,11 +747,18 @@ List<KeyValuePair<string, string>> Echo()
...
                 }
                 Arm("winner (carriers are the load-order-effective versions)");
-                envelope.Add(new("walk", "reverse, depth 1 — the typed MGEF carrier lane"));
-                headerLine += "\nwalk=reverse (per seed: every SPEL/ENCH/ALCH/SCRL/INGR applying it, with the MATCHING entry's magnitude/area/duration — reported AS AUTHORED; conditions are not evaluated, so a row means 'defines it at this strength', not 'it will fire')";
                 // The census separates written rows from the true total and names capped seeds, so the artifact
                 // and its census cannot disagree and a walk.max_nodes cut is always declared.
                 int carrierRows = results.Sum(r => r.Result.Error is null ? r.Result.Rows.Count : 0);
+                // This lane follows the effect link at every hop, and a carrier is not a magic effect — so hop 2
+                // and beyond reach nothing. Said out loud when a depth was asked for, never a silent stop.
+                var carrierHops = $"hop 1: {carrierRows}";
```
Avick3110: The hop-1 count is `carrierRows` — the rows actually written — not `carrierTotal`. When a seed hits the per-seed `walk.max_nodes` carrier bound, `Result.Rows.Count` is a prefix of `Result.Total`, so the same response says `hop 1: <prefix>` a few lines above a census reading `carrier_rows=<prefix> carrier_total=<bigger>` and a `[!] … their rows are a prefix of carrier_total` banner.

A caller reading "hop 1: N" reasonably takes N as what hop 1 reached; here it is what got printed. Use `carrierTotal` for the hop census (or name both), so the hop line and the census cannot disagree.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -58,7 +58,7 @@ public sealed class RecordsWalk
...
+        [Description("'forward' (default) — what the seeds point AT (cheap: each hop is one link resolve). 'reverse' — what points AT the seeds, at any depth, needing no bounding scope: it follows EVERY link at every hop, off the reverse-reference index, which is built on the first such call at the cost of one whole-order link-walk and reports that cost in the response. The FORM picks the reverse lane, as it does forward: a reading form (summary/fields/rows/everything/aggregate) consumes the reached set and the response names the count at each hop, empty hops included; form='chain' is the typed MGEF carrier lane instead — magic-effect seeds get per-carrier magnitude/area/duration (types= narrows the carrier types; walk.max_nodes bounds each seed's carrier rows; limit=/offset= window the SEEDS), and it reaches nothing past hop 1 because a carrier is not a magic effect. references= is the same reverse question as one step of SELECT.")]
```
Avick3110: Two descriptions that the PR leaves saying the opposite of what now happens:

1. The bad-direction refusal a few lines down still teaches the deleted rule — `walk.direction='{x}' — use 'forward' (what the seeds point at) or 'reverse' (what points at them; **depth 1**).` A caller who typos the direction is told reverse is depth-1 only, which this PR is exactly the removal of.

2. `max_nodes`'s own description reads "Maximum nodes reached **per seed** (default 2000…)". That is still true of the carrier lane, but `ReverseSelection.Transitive` spends one shared budget across all seeds and all hops. Ten seeds under the transitive lane get 2000 nodes total, not 2000 each, and the cap banner names `walk.max_nodes` without saying which reading applies.

src/housecarl-mcp/ReverseWalkBatch.cs
```diff
@@ -0,0 +1,54 @@
...
+            catch (Exception ex)
+            {
+                return new Result(Array.Empty<ReverseSelection.Hop>(), Array.Empty<string>(), false, null, epoch,
+                                  $"bad FormID '{raw}': {ex.Message} — every seed of a reverse walk must parse before the walk starts.");
+            }
+        }
+
+        var built = view.EnsureReverseIndex();
+        var hops = ReverseSelection.Transitive(view.ReverseIndex!, seedKeys, depth, maxNodes, out var capped);
+
```
Avick3110, line 112: The seed half of `Selection` is not deduplicated, so `formids=["000805:X.esm", "000805:X.esm"]` (or two spellings of one key — a runtime FormID and its `ID:Plugin` form both parse to the same `FormKey`) puts the same record in the selection twice and the reading form lists it twice. `visited` dedupes it for the traversal, so only the render is wrong.

The forward lane does dedupe its handoff (`seen` / `combined` in `WalkLane`); this lane should do the same — dedupe `seedKeys` before building `selection`, which also keeps `reachedRev = Selection.Count - ids.Length` honest.

Avick3110:
**Design point, asked against the 2026-09-06 #489 ruling: is telling the two reverse lanes apart by `form` "one spelling meaning two things"?**

Yes, and it does not mirror the forward direction. The answer is decided by *where the form branch sits relative to the traversal*.

**Forward.** `RecordsTools.cs:796` runs `svc.WalkForwardBatch(ids, walk.seed_paths, walk.follow, walkDepth, walkMaxNodes, walkExclusions, ...)` **once**, and only then, at line 803, does `if (form == "chain")` split. Chain renders the rows; every reading form consumes the same rows as a selection. Same `walk=` arguments, same traversal, same reached set, two renderings. That is the property that makes "the FORM picks the lane, as it does forward" true forward: form picks a *view*, never a *walk*.

**Reverse, as this PR builds it.** `reverseCarrier` (line 407) is decided *before* any traversal, and the two arms in `WalkLane` are two different walks:

| | `form='chain'` | a reading form |
|---|---|---|
| traversal | `EffectChain` typed carrier scan | `ReverseSelection.Transitive` off the index |
| legal seeds | MGEF only (a non-MGEF seed fails loud per item) | any record type |
| edges followed | the effect link | every link |
| `types=` | legal, narrows carrier types | refused |
| `walk.max_nodes` | per seed, carrier rows | one budget shared across all seeds and all hops |
| reached at depth 3 | hop 1 only, hops 2–3 declared 0 | genuinely transitive |

So `walk={"direction":"reverse","depth":3}` — the identical argument — produces a different reached set, accepts a different seed population, and gives `max_nodes` a different unit, depending on a parameter that forward uses only to choose a renderer. That is the ruling's "no spelling means two things", squarely: not one spelling with two views, but one spelling with two walks.

**And the ruling already names the discriminator this PR set aside.** The 2026-09-06 line is explicit: "Transitive reverse is `direction=reverse` with `follow` unset (every link); the typed MGEF carrier walk keeps its own follow." SPEC §3.3 spells both the same way — `effect_chain` is "direction=reverse, depth=1, follow=`Effects[].BaseEffect`", and "`references=` is the same construct with follow=`*`". The lanes were specified to split on `follow`, which is the axis that already means "which edges this walk crosses" in both directions.

The PR's stated reason is that `follow` is not settable on reverse — but that is this surface's own refusal, not a property of the reverse direction: line 356 still refuses `walk.follow` on reverse with "walk.seed_paths/follow/exclusions shape a FORWARD expansion … Drop them", a sentence that contradicts §3.3 and that this PR edited without revisiting. The remedy that obeys the ruling is to narrow that refusal to `seed_paths`/`exclusions` (which really are forward-shaping) and let `follow` through on reverse: unset/`*` = the transitive index walk, `follow="Effects[].BaseEffect"` = the carrier lane. Then `form` goes back to meaning exactly what it means forward, `depth` and `max_nodes` keep one meaning each, and the carrier lane's "reaches nothing past hop 1" prose becomes a consequence of the follow the caller named rather than a special case the tool has to explain.

Two knock-on symptoms dissolve with the same change, both already visible in the diff: `types=` legal on one reverse call and refused on the other, and `max_nodes` documented as "per seed" while one of the two lanes spends it as a single shared budget.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. The exhaustion clause is gated on `!rev.Capped`, so a capped walk is never also called exhausted — and the hop the cut landed on now renders `hop N: <count> (cut by walk.max_nodes)`, so the two sentences cannot both be told. Test: `ACappedReverseWalkNamesTheCutHopAndIsNotCalledExhausted`.

src/housecarl-core/ReverseReferenceIndex.cs
Avick3110: Folded, both halves. The budget is tested before `visited.Add`, so a node the cut rejects stays unvisited and a raised `max_nodes` on a retry sees the same graph; and `Hop` gains a `Cut` flag that the render turns into `(cut by walk.max_nodes)` beside the count, so a hop the budget ended can never read as a hop that reached nothing.

src/housecarl-mcp/ReverseWalkBatch.cs
Avick3110: Folded. `ReverseSelection.Transitive` now takes a verifier and `ReverseWalkBatch` supplies it: each index candidate has its winner body fetched and its `EnumerateFormLinks` tested against the current frontier — the same second step `references=` takes, with the same deleted-body rule — and a candidate that fails is neither reported nor expanded, so a false hop-1 node cannot seed a subtree. The response says how many candidates the check dropped. The fixture gains `HcRecListDropped`: it carries the link in the master and the winning override clears it, and the test asserts `references=` and the walk both exclude it.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. `walkDepthAsked` is set only when `walk.depth` was actually parsed, and the empty-hop census and its prose are gated on it, so a plain carrier call prints no hop line at all. Test: `TheCarrierLaneWithNoDepthAskedPrintsNoHopCensus` asserts the response contains neither `hop 2: 0` nor `hop 16`.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. The hop census reads `hop 1: {carrierTotal}` — what hop 1 reached — and `carrier_rows` stays in the census as what was written, so the hop line and the census can no longer disagree. The capped banner names the total the selection is a prefix of.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded, both. (1) The bad-direction refusal now teaches the follow that picks the reverse walk instead of the deleted depth-1 rule; a test asserts the refusal does not contain "depth 1". (2) `max_nodes`'s description states both readings — per seed on a forward walk and on the carrier walk, one budget shared across every seed and hop on the transitive walk — and each cap banner names which reading it spent.

src/housecarl-mcp/ReverseWalkBatch.cs
Avick3110: Folded. `ReverseWalkBatch` dedupes `seedKeys` on the parsed `FormKey` before the walk, and `Result` carries the deduplicated `Seeds` count so `reachedRev = Selection.Count - Seeds` is honest. Test: `RepeatedSeedsAreDeduplicated`.

Avick3110:
Folded, as directed. The refusal at line 356 is narrowed to `seed_paths`/`exclusions` — the two that really do shape a forward expansion — and `follow` is legal on reverse, where it is now the only thing that picks the walk:

- unset or `"*"` — the transitive walk over every link, off the index.
- `"Effects[].BaseEffect"` — the typed MGEF carrier walk (SPEC §3.3's spelling, hoisted to one `CarrierFollow` constant that the validation, the lane and every remedy sentence read).
- anything else — refused, naming the two the index can serve: it holds every link and no per-path breakdown, so there is no third follow to honour.

The knock-ons dissolve as you said they would. `types=` is now legal on the carrier follow and refused on the transitive one — keyed on follow, not on form. `max_nodes` documents both readings and each cap banner names which one it spent. The carrier walk's "reaches nothing past hop 1" prose is now phrased as a consequence of `walk.follow="Effects[].BaseEffect"` crossing the effect link at every hop, and it is only told when a depth was actually asked for.

`form` is back to picking a view: both walks hand their reached set to the same two renderers, and the carrier walk under a reading form (new — it was chain-only) consumes the carrier records as its selection exactly as a forward walk consumes its reached set.

**One residue, declared rather than left silent.** `form='chain'` with `follow` unset still means the carrier walk. Making follow-unset mean `"*"` under every form would have turned every shipped `walk={"direction":"reverse"}, form='chain'` call into a different walk with no word said, and the transitive walk has no per-seed path for `chain` to draw anyway — it expands one shared frontier, not an expansion per seed. So that spelling keeps its meaning, it is stated in the `direction` and `follow` descriptions and echoed on every response's walk line, and `form='chain'` with an explicit `follow="*"` refuses naming both alternatives rather than quietly rendering something else. If you would rather that combination refuse too and the carrier lane always name its follow, say so and I will make it refuse — it is a one-line change and one test.

Avick3110:
The residue is gone, per your 2026-09-06 ruling: no default follow implies a walk under any form.

`form='chain'` on a reverse walk with `follow` unset now refuses in one sentence that names the two follows the index can serve — `"Effects[].BaseEffect"` for the typed carrier walk, which chain renders, and `"*"` for every link, which chain cannot render because the transitive walk expands one shared frontier and has no per-seed path to draw — and points that reading at a reading form. `form='chain'` with an explicit `follow="*"` keeps its own refusal, unchanged.

So the carrier lane always names its follow. The `direction` and `follow` descriptions say there is no default under `chain`, the retired `housecarl_effect_chain` redirect now teaches `walk={"direction": "reverse", "follow": "Effects[].BaseEffect"}`, and every reverse chain test says the token outright. No skill text under `.claude/skills` spells a reverse chain call, so there was nothing to update there.

New tests: the follow-unset chain call refuses naming both follows, and the same call with the carrier follow said outright is served, so the refusal is about the missing follow and not about chain on reverse. 1539 tests pass, ci-all 128/128.

src/housecarl-mcp/ReverseWalkBatch.cs
```diff
@@ -0,0 +1,86 @@
...
+        // The index answers in candidates — it says SOME plugin's copy carries the link. references= then re-tests
+        // each candidate against the body it judges, and so does this: a record whose winner dropped the link is
+        // neither listed nor expanded, so the two spellings of the reverse question cannot disagree and a false
+        // hop-1 node cannot seed a false subtree.
+        bool Verify(FormKey candidate, IReadOnlySet<FormKey> frontier)
+        {
+            var w = view.ResolveWinner(candidate);
+            if (w is null) { dropped++; return false; }
+            IMajorRecordGetter? body;
+            try { body = view.GetRecord(session, w.Value.WinnerPlugin, candidate); }
```
Avick3110: **Verify catches only `PluginUnreadableException`, so any other throw kills the whole walk.**

`GetRecord` ends in `SeekBody` -> `ov.EnumerateMajorRecords()`, a lazy Mutagen overlay enumeration that throws on a malformed subrecord or an unparseable body. The `references=` lane wraps exactly this fetch in `catch (Exception ex)` and counts the record as `unscannable` (`LoadOrderService.cs:4275`), so one corrupt record leaves the answer complete minus that record.

Here the same throw propagates out of `Verify`, out of `ReverseSelection.Transitive`, out of `Run`, and the whole reverse walk dies in `Guard.Tool` with a generic fault. So the two spellings this PR says agree do not agree on the case that actually happens in the field, and one bad record makes the lane unusable rather than short by one.

```suggestion
            try { body = view.GetRecord(session, w.Value.WinnerPlugin, candidate); }
            catch (Exception) { dropped++; return false; }
```

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -714,17 +771,53 @@ List<KeyValuePair<string, string>> Echo()
...
+                        {
+                            var key = row.Carrier.ToString();
+                            if (carrierSeen.Add(key)) carrierSel.Add(key);
+                        }
+                    }
+                    if (carrierSel.Count == 0)
+                        return json ? JsonWire.RenderError($"the reverse carrier walk reached nothing readable ({seedErrs2} seed error(s) — run form='chain' to see each seed's outcome).", epochR)
+                                    : $"error: the reverse carrier walk reached nothing readable ({seedErrs2} seed error(s) — run form='chain' to see each seed's outcome)." + (epochR is not null ? $"\nepoch={epochR}" : "");
+                    envelope.Add(new("selection", $"the {carrierSel.Count} record(s) the carrier walk reached (seeds included)"));
+                    headerLine += $"\nwalk: selection = {carrierSel.Count} reached record(s) (seeds included)";
```
Avick3110, line 827: **The carrier reading form drops seed errors silently.**

`seedErrs2` is computed at line 779 and then never surfaced on this path unless the selection is completely empty.

Scenario: `formids=[<MGEF>, <WEAP>]` with `walk={"direction":"reverse","follow":"Effects[].BaseEffect"}` and `form='summary'`. The WEAP seed produces `Result.Error`, the loop above skips it with `continue`, `carrierSel` is non-empty from the MGEF, so the `carrierSel.Count == 0` guard does not fire — and neither the envelope nor `headerLine` mentions the failed seed. The caller gets a clean-looking answer covering one of two seeds.

The forward walk's identical selection-consumption path does report it (its envelope appends `"; {seedErrs} seed error(s), listed via form='chain'"`), and the chain form reports it as an `errors` count, so the reverse reading form is the only view that swallows it. It is also the exact case this lane's own comment says must not happen ("a non-MGEF seed fails loud per item rather than reading as '0 carriers'").

```suggestion
                    headerLine += $"\nwalk: selection = {carrierSel.Count} reached record(s) (seeds included)";
                    if (seedErrs2 > 0)
                        headerLine += $"\n[!] {seedErrs2} seed(s) failed and contributed no carriers — run form='chain' to see each seed's outcome.";
```

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -682,7 +707,39 @@ List<KeyValuePair<string, string>> Echo()
...
+                    return json ? JsonWire.RenderError(rev.Refusal, rev.Epoch) : "error: " + rev.Refusal + (rev.Epoch is not null ? $"\nepoch={rev.Epoch}" : "");
+                if (SeamTear(rev.Epoch) is { } rTear)
+                    return json ? JsonWire.RenderError(rTear, rev.Epoch) : "error: " + rTear;
+                // Every hop is named with its count, empty ones included: a walk that ran out of referrers says so
+                // on the response instead of trailing off, so an empty hop 2 is visible and not silent. A hop the
+                // budget ended says 'cut' instead, so a prefix is never read as a finding, and the exhaustion
+                // clause is only told of a walk that actually ran out.
+                var hopLine = string.Join(", ", rev.Hops.Select(h => $"hop {h.Depth}: {h.Reached.Count}{(h.Cut ? " (cut by walk.max_nodes)" : "")}"));
+                if (!rev.Capped && rev.Hops.Count < walkDepth)
+                    hopLine += $" (nothing left to expand, so hops {rev.Hops.Count + 1}–{walkDepth} were not walked)";
```
Avick3110, line 726: **The depth cap never announces itself on this lane, so a truncated walk reads as a complete one.**

The exhaustion clause fires only when `rev.Hops.Count < walkDepth`. The complementary case — the walk stopped because it hit `walk.depth` — adds nothing at all.

Scenario: `walk={"direction":"reverse","depth":2}` over a record with a five-hop referrer chain renders `hop 1: 1, hop 2: 1` and stops. `hops.Count == walkDepth` so this clause is skipped; `rev.Capped` is false so the `max_nodes` warning is skipped too. A record whose graph genuinely ends at hop 2 renders identically, so the caller cannot tell a prefix from a finding — and against the exhaustion sentence the walk *does* print elsewhere, the silence reads as a positive claim of completeness.

`walk.depth`'s own description promises the opposite: "Nodes AT the cap are recorded, not entered, and the response says the cap cut the walk — never a silent stop."

A last hop with a non-empty `Reached` and `hops.Count == walkDepth` is exactly the cut case and can be said in one clause here.

src/housecarl-mcp/ReverseWalkBatch.cs
```diff
@@ -0,0 +1,86 @@
...
+        int dropped = 0;
+        using var session = pin.Resolver.OpenSession();
+        // The index answers in candidates — it says SOME plugin's copy carries the link. references= then re-tests
+        // each candidate against the body it judges, and so does this: a record whose winner dropped the link is
+        // neither listed nor expanded, so the two spellings of the reverse question cannot disagree and a false
+        // hop-1 node cannot seed a false subtree.
+        bool Verify(FormKey candidate, IReadOnlySet<FormKey> frontier)
+        {
+            var w = view.ResolveWinner(candidate);
+            if (w is null) { dropped++; return false; }
```
Avick3110: **`dropped` collects four different causes; the sentence rendered for it asserts one.**

A candidate is counted here when its winner cannot be resolved (line 60), when the winning plugin has become unreadable since the index was built (line 63), when the winner is deleted or is not a link container (lines 64-68), and when the winner genuinely dropped the link (line 71).

`RecordsTools.cs:733` then renders all four as:

> N index candidate(s) were dropped: the index says some plugin's copy carries the link, and the body this walk judges — the load-order winner — does not

So a user whose plugin went unreadable mid-session is told the winner dropped the link. That is a positively wrong explanation for what is actually a coverage gap — and it is the case `references=` deliberately keeps distinct, reporting it as `unscannable` with its own sample sentence rather than as a non-match.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -336,6 +336,51 @@ public static IReadOnlyList<FormKey> Universe(LoadOrderResolver.IndexView view,
...
+        for (int d = 1; d <= depth; d++)
+        {
+            var frontierSet = new HashSet<FormKey>(frontier);
+            var next = new List<FormKey>();
+            bool cut = false;
+            foreach (var k in index.ReferencersOf(frontier))
+            {
+                if (visited.Contains(k)) continue;
+                if (verify is not null && !verify(k, frontierSet)) continue;
+                if (reached >= maxNodes) { capped = true; cut = true; break; }
```
Avick3110: **The budget is tested after `verify`, so once it is spent it bounds nothing.**

The break only happens on a candidate that PASSES verification. Every candidate ahead of that one which fails — a dropped link, a deleted record, a null winner — is still verified, and verification is a winner resolve plus a full overlay seek for the body.

Scenario: `walk.max_nodes=1` against a frontier whose candidate list is mostly stale overrides. `reached` hits 1 on the first hop, and the walk then reads records for the whole remaining candidate list before it happens on one that verifies. Each of those reads also increments `dropped`, so the response reports thousands of "index candidate(s) were dropped" for a walk that had already stopped reaching.

Swapping the two lines keeps the stated "budget tested BEFORE the node is consumed" property (a raised budget on a retry still sees the same graph, since verification has no ordering effect) while making `max_nodes` actually bound the work.

```suggestion
                if (reached >= maxNodes) { capped = true; cut = true; break; }
                if (verify is not null && !verify(k, frontierSet)) continue;
```

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -336,6 +336,51 @@ public static IReadOnlyList<FormKey> Universe(LoadOrderResolver.IndexView view,
...
+        int reached = 0;
+        for (int d = 1; d <= depth; d++)
+        {
+            var frontierSet = new HashSet<FormKey>(frontier);
+            var next = new List<FormKey>();
+            bool cut = false;
+            foreach (var k in index.ReferencersOf(frontier))
+            {
+                if (visited.Contains(k)) continue;
+                if (verify is not null && !verify(k, frontierSet)) continue;
```
Avick3110: **A candidate that fails `verify` is never recorded, so it is re-fetched and re-counted at every later hop.**

Only verified nodes go into `visited`. Record X that is a candidate for both hop 1's frontier and hop 2's frontier — its master copy links into both, its winner links into neither — is verified twice, and `dropped` reads 2 for one record. So the response's "N index candidate(s) were dropped" is a count of verification failures, not of records, and cannot be reconciled against the index's candidate list.

The mirror case is worse to read: a record dropped at hop 1 but legitimately reached at hop 2 appears in the selection *and* in the drop total.

A `rejected` set beside `visited` would fix both the count and the repeated body reads.

src/housecarl-mcp/ReverseWalkBatch.cs
Avick3110: Folded. The fetch now catches any throw, not `PluginUnreadableException` alone, so a malformed subrecord out of the lazy overlay seek costs that one record and not the walk — the same rule `references=` keeps at LoadOrderService.cs:4287. The link enumeration that follows is wrapped too, since it is the same lazy read one step later. Those records land in the drop census as `whose winning plugin could not be read — a coverage gap, not a verdict` (the comment below), so they are never rendered as a winner that dropped the link. Commit 8265fec2, with the census in 1f8e3be0.

No test: making a winner body throw needs a plugin corrupted after the index was built, and the freshness check rebuilds the index on the changed bytes before the walk ever reads it, so the fixture would be asserting on the index's own unreadable-partition path instead of this one. Said plainly rather than papered over with a test that proves something else.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded, as suggested. The carrier reading form now appends `[!] {n} seed(s) failed and contributed no carriers — run form='chain' to see each seed's outcome.` to the header, and the envelope's selection line carries `; {n} seed error(s), listed via form='chain'` the way the forward walk's does. An answer over one of two seeds can no longer read as an answer over both. Test: `TheCarrierReadingFormSaysASeedFailed` — an MGEF seed and a non-MGEF seed under a reading form, asserting the served rows and the failed-seed line together. Commit 2d87c5c2.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. The complementary clause is there now: when the walk is not capped and the last hop still found records at `hops.Count == walkDepth`, the hop line says `(walk.depth=N was reached with hop N still finding records, so they were recorded and not expanded — raise walk.depth to walk further)`. It is an `else if` on the exhaustion clause, so exactly one of the two is ever told, and `walk.depth`'s description is true again. Tests: `AWalkStoppedByWalkDepthSaysTheCapCutIt` (depth 1 over the hop chain: says the cap, never "nothing left to expand") and `AWalkThatRanOutBeforeTheCapIsStillCalledExhausted` (depth 5: the exhaustion sentence, not the cap one). Commit 4afd0230.

src/housecarl-mcp/ReverseWalkBatch.cs
Avick3110: Folded. `Result.Dropped` is a `DropCensus(NoLink, Unreadable, NoLiveBody, NoWinner)` and the render names each non-zero cause with its own count:

> N index candidate(s) were dropped — the index names a plugin copy that carries the link, and this walk judges the load-order winner (the same second step references= takes): a whose winner does not carry the link; b whose winning plugin could not be read — a coverage gap, not a verdict; c whose winner is deleted or carries no links; d with no resolvable winner. None of them was reached or expanded.

The lead sentence no longer asserts a cause, and the unreadable arm says outright that it is a coverage gap rather than a verdict — the distinction `references=` keeps as `unscannable`. A null fetch moved out of the deleted/no-links arm into the unreadable one, matching the `winner did not yield the record on fetch` sample `references=` writes for the same case. Test: `TheDropLineNamesTheCauseRatherThanAssertingOne`. Commit 1f8e3be0.

src/housecarl-core/ReverseReferenceIndex.cs
Avick3110: Folded, as suggested — the two lines are swapped, so the budget is spent before the candidate is verified and a spent budget bounds the body reads, not only the reach. The property the doc comment claims still holds: verification has no ordering effect on the graph, so a raised `max_nodes` on a retry sees the same candidates in the same order. A hop can now be marked cut on a candidate that would have failed verification, which claims "there may be more" where the walk might in fact have been exhausted — the conservative direction, and never the reverse.

Test: `ASpentBudgetStopsVerifyingTheCandidatesBehindIt` — `formids=[SpellHop], depth=1, max_nodes=1`, where `HcRecListHop` fills the budget and `HcRecListDropped` sits behind it: the response reaches ListHop and reports no dropped candidates at all, where before the fix it read the stale body and counted it. Commit db603b5c.

src/housecarl-core/ReverseReferenceIndex.cs
Avick3110: Folded, both halves, though not with a bare `rejected` set — that would have caused the silent short answer you name in the mirror case. Whether a winner carries a link is frontier-relative, so a record rejected at hop 1 can be a genuine referrer at hop 2, and skipping it on a remembered rejection would drop it from the selection.

What landed instead is a per-record verdict cache: `linksOf` maps each candidate to its winner's link set, or to null when the winner cannot be judged at all (no winner, unreadable, deleted or not a link container — all three permanent). So the body is fetched exactly once per record however many hops name it, and those three causes count records by construction. The frontier test is re-asked per hop off the remembered set, which costs a set lookup and no read.

The count: no-link drops are a `HashSet<FormKey>` rather than a counter, and a candidate that later verifies is removed from it — so the mirror case is fixed too, and a record cannot appear in both the selection and the drop total. `DropCensus.NoLink` is `noLink.Count`, a count of records that reconciles against the index's candidate list.

Test: `ACandidateNamedAtTwoHopsIsDroppedOnce`. The fixture's `HcRecListDropped` now carries both `MgefHop` and `SpellHop` in the master (the winner still clears the list), so a two-hop walk from `MgefHop` names it a candidate at hop 1 and again at hop 2; the response says `1 index candidate(s) were dropped`, where before the fix it said 2. Commit 2152a1ff.

PR #574, opened 2026-09-05
Read the quantified step in project.fields

### LINKED ISSUE #458: [GAP] where= cannot ask "does ANY element of this list match?" — the general lookup / referenced-by capability

Avick3110:
`where=` has taken the quantifier tokens since E1; the projection half was refused as unbuilt. This adds it: `project.fields=["Effects[*count]"]` renders one number per record (the element count of that list), and `["Effects[*]"]` renders one row per element in the same row shape `project.form='rows'` produces — the fold IS `RowProjection`'s, so a row here and a row under the rows form are the same row rather than two renderings of one idea. A sub-path after the token (`"Effects[*].Data.Magnitude"`) renders that leaf per element instead. Under `format='dense'` the extra rows repeat the record's identity columns, and an unquantified column repeats its own cell beside them; the text and json lanes take the existing body lane, so the render, the artifact and the json document all see the same folded fields.

The token vocabulary moved into `PathFoldGrammar` in core and `FieldPredicate` now reads through it, so `where=` and `project.fields` cannot drift on what a token means — the SELECT half's tokenizer is reused, not copied. `[*any]`/`[*all]`/`[*none]` fold to a boolean, which is not a row, and stay refused here by name; a quantifier on a step that is not a list fails that record with one sentence naming the path, what the step actually holds, and the two ways to read it.

Tests: `RecordsFieldFoldTests` covers the count on a list, the bare `[*]` in the rows shape and in dense, the sub-path form, composition with an ordinary path, and every refusal. Full suite green (1537), `ci-all` 128/128.

Closes #458

### DIFF
--- a/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
+++ b/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
@@ -1,0 +1,136 @@
+using HousecarlMcp;
+using Xunit;
+
+namespace HousecarlMcpTests;
+
+/// <summary>
+/// The PROJECT half of the quantified step: <c>project.fields</c>' <c>[*count]</c> (one number per record) and
+/// <c>[*]</c> (one row per element), driven end to end through the tool over the fixture's two-effect spell.
+/// </summary>
+[Collection("records")]
+[Trait("tier", "integration")]
+public sealed class RecordsFieldFoldTests : RecordsTestBase
+{
+    public RecordsFieldFoldTests(RecordsFixture f) : base(f) { }
+
+    string Spell(params string[] paths) =>
+        RecordsTools.Records(Svc, formids: new[] { Fid(W.SpellA) }, project: Fields(paths));
+
+    [Fact]
+    public void CountOnAListIsOneNumberPerRecord()
+    {
+        var r = Spell("Effects[*count]");
+        Served(r, "Effects[*count] = 2");
+        // The number, not the list: no element line survives.
+        Assert.DoesNotContain("Effects[0]", r);
+    }
+
+    [Fact]
+    public void CountReadsTheListWithoutOpeningIt()
+    {
+        // A count needs no expansion, so the call costs the depth-1 read even though [*] would force depth 4.
+        var r = RecordsTools.Records(Svc, types: new[] { "SPEL" }, format: "dense", project: Fields("Effects[*count]"));
+        var doc = Je(r);
+        Assert.Equal(new[] { "formid", "runtime_formid", "editorid", "Effects[*count]" },
+                     doc.GetProperty("columns").EnumerateArray().Select(c => c.GetString()).ToArray());
+        var spellA = doc.GetProperty("rows").EnumerateArray().First(x => x[2].GetString() == "HcRecSpellA");
+        Assert.Equal("2", spellA[3].GetString());
+    }
+
+    [Fact]
+    public void TheBareStarIsOneRowPerElement_TheRowsFormsOwnRowShape()
+    {
+        var r = Spell("Effects[*]");
+        Served(r, "Effects[0] = ", "Effects[1] = ");
+        // Each row carries the element's sub-fields inline, exactly as form='rows' renders them.
+        Assert.Contains("Data.Magnitude=5", RowLine(r, "Effects[0]"));
+        Assert.Contains("Data.Magnitude=4", RowLine(r, "Effects[1]"));
+        // And never as lines of their own — that is the whole difference from naming the list.
+        Assert.DoesNotContain("Effects[0].", r);
+    }
+
+    [Fact]
+    public void AQuantifiedPathComposesWithAnOrdinaryOne()
+    {
+        var r = Spell("EditorID", "Effects[*]");
+        Served(r, "EditorID = HcRecSpellA", "Effects[0] = ", "Effects[1] = ");
+    }
+
+    [Fact]
+    public void ASubPathAfterTheTokenIsThatLeafPerElement()
+    {
+        var r = Spell("Effects[*].Data.Magnitude");
+        Served(r, "Effects[0].Data.Magnitude = 5", "Effects[1].Data.Magnitude = 4");
+    }
+
+    [Fact]
+    public void DenseCarriesTheExtraRowsWithTheIdentityColumnsRepeated()
+    {
+        var doc = Je(RecordsTools.Records(Svc, formids: new[] { Fid(W.SpellA) }, types: new[] { "SPEL" },
+                                          format: "dense", project: Fields("EditorID", "Effects[*]")));
+        Assert.Equal(new[] { "formid", "runtime_formid", "editorid", "EditorID", "Effects[*]" },
+                     doc.GetProperty("columns").EnumerateArray().Select(c => c.GetString()).ToArray());
+        var rows = doc.GetProperty("rows").EnumerateArray().ToList();
+        Assert.Equal(2, rows.Count);                                    // one row per effect
+        Assert.Equal(rows[0][0].GetString(), rows[1][0].GetString());   // identity repeated, not omitted
+        Assert.Equal("HcRecSpellA", rows[0][3].GetString());            // the unquantified column repeats too
+        Assert.Equal("HcRecSpellA", rows[1][3].GetString());
+        Assert.Contains("Data.Magnitude=5", rows[0][4].GetString()!);
+        Assert.Contains("Data.Magnitude=4", rows[1][4].GetString()!);
+    }
+
+    [Fact]
+    public void AQuantifierOnANonListPathIsRefusedByName()
+    {
+        var r = Spell("EditorID[*]");
+        Assert.Contains("'EditorID[*]' quantifies a LIST", r);
+        Assert.Contains("'EditorID' on Spell", r);
+        Assert.Contains("drop the quantifier", r);
+    }
+
+    [Fact]
+    public void CountOnANonListPathIsRefusedTheSameWay() =>
+        Assert.Contains("'EditorID[*count]' quantifies a LIST", Spell("EditorID[*count]"));
+
+    [Fact]
+    public void AnAbsentListIsTheReadsAnswerNotAMisuseOfTheToken() =>
+        // No Effects field on a WEAP at all: the read says so under the caller's own spelling.
+        Served(RecordsTools.Records(Svc, formids: new[] { Fid(W.Weapons[1]) }, project: Fields("Effects[*count]")),
+               "Effects[*count]");
+
+    [Fact]
+    public void ABooleanFoldStaysInWhere() =>
+        Refused(Spell("Effects[*any].Data.Magnitude"), "not a row", "where=");
+
+    [Fact]
+    public void ANonQuantifierBracketKeyIsNamedATypo() =>
+        Refused(Spell("Effects[*sum]"), "is not a quantifier");
+
+    [Fact]
+    public void NothingCanFollowCount() =>
+        Refused(Spell("Effects[*count].Data"), "nothing can follow '[*count]'");
+
+    [Fact]
+    public void TwoQuantifiedStepsInOnePathAreRefused() =>
+        Refused(Spell("Effects[*].Conditions[*]"), "quantifies two steps");
+
+    [Fact]
+    public void TheTokenBelongsToTheFieldsForm() =>
+        Refused(RecordsTools.Records(Svc, formids: new[] { Fid(W.SpellA) },
+                    project: new RecordsTools.RecordsProject { form = "rows", fields = new[] { "Effects[*]" } }),
+                "the 'rows' form already folds");
+
+    [Fact]
+    public void DepthOneWouldCollapseTheElementsAway() =>
+        Refused(RecordsTools.Records(Svc, formids: new[] { Fid(W.SpellA) },
+                    project: new RecordsTools.RecordsProject { form = "fields", fields = new[] { "Effects[*]" }, depth = 1 }),
+                "Effects[*count]");
+
+    /// <summary>The one rendered line for a row, by its path.</summary>
+    static string RowLine(string response, string path)
+    {
+        var line = response.Split('\n').FirstOrDefault(l => l.TrimStart().StartsWith(path + " = ", StringComparison.Ordinal));
+        Assert.NotNull(line);
+        return line!;
+    }
+}
--- a/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
+++ b/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
@@ -210,6 +210,32 @@ public sealed class RecordsFieldFoldTests : RecordsTestBase
         Assert.NotEqual(JsonValueKind.Null, rows[1][4].ValueKind);
     }
 
+    /// <summary>A DICT list is keyed by its own keys, not by a position: a package's two Data arms keyed 2 and 7
+    /// are two rows, not eight with six empty ones, and the count the document reports is the same two.</summary>
+    [Fact]
+    public void ADictListLaysOneRowPerPairAndNotOnePerKeyValue()
+    {
+        var doc = Je(RecordsTools.Records(Svc, formids: new[] { Fid(W.Package) }, types: new[] { "PACK" },
+                                          format: "dense", project: Fields("Data[*]")));
+        var rows = doc.GetProperty("rows").EnumerateArray().ToList();
+        Assert.Equal(2, rows.Count);
+        Assert.Equal(2, doc.GetProperty("rows_rendered").GetInt32());
+        Assert.Contains("HcRecDatumTwo", rows[0][3].GetString()!);
+        Assert.Contains("HcRecDatumSeven", rows[1][3].GetString()!);
+    }
+
+    /// <summary>The key a row is laid on is the bracket segment as text — the one shape that covers a positional
+    /// list and a dict alike.</summary>
+    [Fact]
+    public void AnElementKeyIsTheBracketSegmentItself()
+    {
+        Assert.Equal("0", FoldPlan.ElementKey("Effects[0].Data", "Effects"));
+        Assert.Equal("7", FoldPlan.ElementKey("Data[7]", "Data"));
+        Assert.Equal("Male", FoldPlan.ElementKey("WorldModel[Male]", "WorldModel"));
+        Assert.Null(FoldPlan.ElementKey("Effects", "Effects"));
+        Assert.Null(FoldPlan.ElementKey("Effects.Data", "Effects"));
+    }
+
     [Fact]
     public void TwoDifferentListsCannotShareOneDenseRow() =>
         Refused(RecordsTools.Records(Svc, types: new[] { "SPEL" }, format: "dense",
--- a/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
+++ b/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
@@ -66,6 +66,26 @@ public sealed class RecordsFieldFoldTests : RecordsTestBase
         Served(r, "Effects[0].Data.Magnitude = 5", "Effects[1].Data.Magnitude = 4");
     }
 
+    /// <summary>A bracketed index inside the sub-path costs an expansion level of its own, so a depth counted per
+    /// dotted SEGMENT stops the read one level short and the record answers "no element carries it" — a confident
+    /// wrong answer about a field that is right there.</summary>
+    [Fact]
+    public void AnIndexedSubPathIsReadDeepEnoughToReachIt()
+    {
+        Served(Spell("Effects[*].Conditions[0].Data"), "Effects[1].Conditions[0].Data = ");
+        Assert.DoesNotContain("no element of", Spell("Effects[*].Conditions[0].Data"));
+    }
+
+    /// <summary>The same need, at a depth the caller named: the default of 4 covers the shallower spelling, so the
+    /// claim only bites once the caller asks for less than the token needs.</summary>
+    [Fact]
+    public void TheTokensOwnDepthWinsOverALowerOneTheCallerNamed()
+    {
+        var r = RecordsTools.Records(Svc, formids: new[] { Fid(W.SpellA) },
+                    project: new RecordsTools.RecordsProject { form = "fields", fields = new[] { "Effects[*].Conditions[0]" }, depth = 2 });
+        Served(r, "Effects[1].Conditions[0] = ");
+    }
+
     [Fact]
     public void DenseCarriesTheExtraRowsWithTheIdentityColumnsRepeated()
     {
--- a/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
+++ b/src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
@@ -1,3 +1,5 @@
+using System.Text.Json;
+using HousecarlCore;
 using HousecarlMcp;
 using Xunit;
 
@@ -28,12 +30,13 @@ public sealed class RecordsFieldFoldTests : RecordsTestBase
     [Fact]
     public void CountReadsTheListWithoutOpeningIt()
     {
-        // A count needs no expansion, so the call costs the depth-1 read even though [*] would force depth 4.
-        var r = RecordsTools.Records(Svc, types: new[] { "SPEL" }, format: "dense", project: Fields("Effects[*count]"));
-        var doc = Je(r);
-        Assert.Equal(new[] { "formid", "runtime_formid", "editorid", "Effects[*count]" },
-                     doc.GetProperty("columns").EnumerateArray().Select(c => c.GetString()).ToArray());
-        var spellA = doc.GetProperty("rows").EnumerateArray().First(x => x[2].GetString() == "HcRecSpellA");
+        // The cost claim, on the one fixture that can prove it: the over-budget list truncates the moment it is
+        // expanded, so a count that reports the number WITHOUT that note never opened it.
+        var count = RecordsTools.Records(Svc, formids: new[] { Fid(W.BigList) }, project: Fields("Items[*count]"));
+        Served(count, "Items[*count] = " + (ReadEngine.MaxExpandNodes + 1));
+        Assert.DoesNotContain("expansion truncated", count);
+        var dense = Je(RecordsTools.Records(Svc, types: new[] { "SPEL" }, format: "dense", project: Fields("Effects[*count]")));
+        var spellA = dense.GetProperty("rows").EnumerateArray().First(x => x[2].GetString() == "HcRecSpellA");
         Assert.Equal("2", spellA[3].GetString());
     }
 
@@ -93,10 +96,15 @@ public sealed class RecordsFieldFoldTests : RecordsTestBase
         Assert.Contains("'EditorID[*count]' quantifies a LIST", Spell("EditorID[*count]"));
 
     [Fact]
-    public void AnAbsentListIsTheReadsAnswerNotAMisuseOfTheToken() =>
-        // No Effects field on a WEAP at all: the read says so under the caller's own spelling.
-        Served(RecordsTools.Records(Svc, formids: new[] { Fid(W.Weapons[1]) }, project: Fields("Effects[*count]")),
-               "Effects[*count]");
+    public void AnAbsentListIsTheReadsAnswerNotAMisuseOfTheToken()
+    {
+        // No Effects field on a WEAP at all: the read's own note carries out under the caller's spelling, and the
+        // count is NOT answered with a number — a missing field and an empty list are different answers.
+        var r = RecordsTools.Records(Svc, formids: new[] { Fid(W.Weapons[1]) }, project: Fields("Effects[*count]"));
+        Served(r, "Effects[*count]");
+        Assert.DoesNotContain("Effects[*count] = 0", r);
+        Assert.Contains("no field", Line(r, "Effects[*count]").ToLowerInvariant());
+    }
 
     [Fact]
     public void ABooleanFoldStaysInWhere() =>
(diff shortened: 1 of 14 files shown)


src/housecarl-mcp/FieldFold.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <summary>One record's lines, grouped per REQUESTED path and in the caller's own order: a quantified path
+    /// contributes its count cell or its element rows, an ordinary path the lines the read emitted for it.
+    /// Grouped rather than flat because the columnar render needs to know which column varies per element.</summary>
+    internal (IReadOnlyList<FieldValue>[]? Columns, string? Error) Columns(RecordFields rec)
+    {
+        var setRoots = Folds.Where(f => f is { Fold: PathFold.Set }).Select(f => f!.Root).Distinct(StringComparer.Ordinal)
+                            .OrderByDescending(r => r.Length).ToList();
+        // The element rows come from the 'rows' fold itself, run over the same lines — so a row here and a row
+        // under form='rows' are the same row, never two renderings of one idea.
+        var rows = setRoots.Count > 0 ? RowProjection.Fold(rec.Fields, setRoots, Depth) : rec.Fields;
```
Avick3110, line 69: The read's expansion-truncation note is silently dropped by the fold.

`ReadEngine.Emit` names its cut with a `FieldValue` whose `Path` is `"…"` (ReadEngine.cs:652). `RowProjection.Fold` deliberately passes that line through (`Passthrough`), but neither consumer here keeps it: `Elements` yields only lines where `RowKey(f.Path, roots) == f.Path` (or `== key + tail`), and `Lines` only lines where `f.Path == path || IsUnder(f.Path, path)`. `"…"` matches neither, and `Apply` then replaces `Record.Fields` with exactly `cols.SelectMany(...)`.

Concrete: `project.fields=["Effects[*]"]` (or any quantified path) on a record whose depth-4 expansion exceeds `MaxExpandNodes` — the elements past the cut are missing from the rows and the response says nothing at all about it. Every lane is affected (text, json, dense, and the artifact via `Artifacts.WriteCrossQuery`), and a `[*count]`-only plan drags the same loss onto the unquantified columns beside it.

That is the "silently degraded mode" the design forbids. The note needs to survive `Columns` — e.g. append any field the fold did not claim and whose path is not under a requested path, or carry it explicitly.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -1231,10 +1232,40 @@ public static string RenderCrossQueryDense(LoadOrderService svc, CrossQueryOutco
...
                         if (o.Error is not null) { (errors ??= new()).Add((fk.ToString(), o.Error)); rendered++; continue; }
                         var r = o.Record!;
+                        if (fold is not null)
+                        {
+                            // A quantified path makes the requested paths PER ELEMENT, so the record answers with
+                            // one row per element and repeats its identity columns on each. The unquantified
+                            // columns keep their own single cell, which is the same cell an unfolded call renders.
+                            var (cols, ferr) = fold.Columns(r);
+                            if (ferr is not null) { (errors ??= new()).Add((fk.ToString(), ferr)); rendered++; continue; }
+                            int elems = 1;
```
Avick3110: Dense pairs two quantified columns by ORDINAL, not by element, so columns whose line sets differ silently mis-pair.

`elems` is the max over the Set columns and each cell is `col[e]`. That is only correct if every Set column emits exactly one entry per element, in element order — but `Elements`' tail branch filters `raw` for `f.Path == key + tail` and simply skips an element that emitted no such line. `ReadEngine.Expand` does skip: a null list entry, an arm whose type does not carry the named sub-field (a polymorphic `Conditions[].Data` arm), and any value that fails `IsModeledContent` all produce no line at that path.

Concrete: `format='dense'`, `project.fields=["Conditions[*]", "Conditions[*].Data.Faction"]` on a record whose condition 0 is not a faction arm — the Faction column has one entry, from element 1, and dense writes it on row e=0 beside `Conditions[0]`. The row reads as "condition 0 tests faction X", which is false, and nothing in the document says the columns came from different elements.

The same holds for two `[*]` paths on different lists of different lengths. Either key the element rows by their element index rather than by position, or refuse more than one distinct `[*]` root / a tail column that did not emit one line per element.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -1231,10 +1232,40 @@ public static string RenderCrossQueryDense(LoadOrderService svc, CrossQueryOutco
...
                     if (detail)
                     {
-                        var o = svc.ResolveReadOn(q, fk, winnerFields ? null : (q.Sources is { } src ? src[i] : null), fields, false,
-                                                  resolveNames: resolveNames, linkMemo: linkMemo, containerHint: (levers ?? LeverNames.Legacy).DenseContainerHint);   // dense refuses depth>1, so the hint names the format hop; pinned to the scan's build
+                        var o = svc.ResolveReadOn(q, fk, winnerFields ? null : (q.Sources is { } src ? src[i] : null), fields, false, fold?.Depth ?? 1,
+                                                  resolveNames: resolveNames, linkMemo: linkMemo, containerHint: (levers ?? LeverNames.Legacy).DenseContainerHint);   // dense refuses depth>1 unless a quantifier asks for it; pinned to the scan's build
                         if (o.Error is not null) { (errors ??= new()).Add((fk.ToString(), o.Error)); rendered++; continue; }
                         var r = o.Record!;
+                        if (fold is not null)
+                        {
```
Avick3110, line 1273: The folded dense branch never registers the owned-child fields, so the clause that explains those cells is dropped.

The unfolded path below registers at emission (`if (o.OwnedChildFields?.ContainsKey(f.Path) == true) { childFields.Add(...); childUnioned |= ...; }`, line 1278) and `WriteOwnedChildNote` at line 1310 turns that into the sentence saying the values are index-only / unioned. The fold branch `continue`s before ever touching `childFields`, so with a fold the set stays empty and the note is omitted.

Concrete: `format='dense'`, `types=["CELL"]`, `project.fields=["Temporary[*]"]` — the rows carry the owned-child cells and the document no longer states the clause that says what they are. `Artifacts.WriteCrossQuery` still collects it off the folded outcome, so the spilled artifact and the inline render now disagree about the same rows.

Registration needs to happen per emitted cell inside the element loop too (the `FieldValue`s in `cols` keep their paths, but a `[*]` row's path is the element path, so key on `fold.Root` / the row's `Cells` paths).

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1240,14 +1255,16 @@ List<KeyValuePair<string, string>> Echo()
             // ---- form=everything on a scan: selection here, bodies via the batch lane, window-bounded.
             // counts_only skips the body lane entirely — its census is the scan render below. ----
             // The rows form always takes this lane: its fold is over a body read, and the scan render fills
-            // detail rows of its own that never pass through it.
-            if ((form == "everything" || form == "rows" || (form == "fields" && scopePlusPole)) && !counts_only && outcome.Error is null && outcome.Groups is null)
+            // detail rows of its own that never pass through it. A quantified fields path folds the same way, so
+            // it takes the lane too — except under dense, whose own render carries the per-element rows.
+            if ((form == "everything" || form == "rows" || (form == "fields" && (scopePlusPole || (foldPlan is not null && !dense))))
```
Avick3110: This lane now swallows the reverse-index accounting note.

The body-lane condition here gained `(foldPlan is not null && !dense)`, but the shared reverse-index gate at line 1182 still reads `(form == "everything" || (form == "fields" && scopePlusPole)) && !counts_only`. Its own comment says the note lives in one place so no form forgets it — and the two CrossQuery renderers read it off the outcome themselves, which is why plain `form='fields'` was fine before.

Concrete: `references=[...]`, `project={form:'fields', fields:['Effects[*count]']}`, `format='text'` — the call now takes this body lane and renders through `Wire.RenderBatch`, whose header and envelope carry only what was added above. `outcome.ReverseIndexNote` (the index build cost and per-plugin freshness key) is never added, so it is dropped from the response entirely.

Line 1182's condition wants the same `foldPlan is not null && !dense` clause (and, for what it's worth, `form == "rows"` looks like it was already missing there).

src/housecarl-mcp/JsonWire.cs
```diff
@@ -1231,10 +1232,40 @@ public static string RenderCrossQueryDense(LoadOrderService svc, CrossQueryOutco
...
+                            // A quantified path makes the requested paths PER ELEMENT, so the record answers with
+                            // one row per element and repeats its identity columns on each. The unquantified
+                            // columns keep their own single cell, which is the same cell an unfolded call renders.
+                            var (cols, ferr) = fold.Columns(r);
+                            if (ferr is not null) { (errors ??= new()).Add((fk.ToString(), ferr)); rendered++; continue; }
+                            int elems = 1;
+                            for (int c = 0; c < cols!.Length; c++) if (fold.Folds[c] is { Fold: HousecarlCore.PathFold.Set }) elems = Math.Max(elems, cols[c].Count);
+                            for (int e = 0; e < elems; e++)
```
Avick3110, line 1311: The element loop is outside the `max_chars` cap check, so one record can overrun the ceiling by an unbounded amount.

The `ms.Length >= cap` test runs once per RECORD at the top of the outer loop. Every other dense row is one record, so that was tight; a `[*]` record now emits `elems` rows before control returns to the check.

Concrete: `format='dense'`, `types=["CELL"]`, `project.fields=["Temporary[*]"]` with a default `max_chars` — a worldspace cell holding a few thousand placed refs writes a few thousand rows in this loop before the cap is consulted again, so the response blows past the caller's ceiling in one go.

Also note `rendered` still counts records while `rows` now holds one entry per element, so a consumer cannot get the row count from the accounting. Hoisting the flush/cap check into the `for (int e = ...)` loop (setting `rowsTruncated` and breaking out) fixes both the overrun and the truncation signal.

src/housecarl-mcp/FieldFold.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+        var tail = "." + string.Join(".", fold.Tail);
+        foreach (var f in raw)
+            if (RowProjection.RowKey(f.Path, roots) is { } key && f.Path == key + tail && seen.Add(f.Path))
+                yield return f;
+    }
+
+    /// <summary>The lines an ordinary (unquantified) path contributed: its own, plus whatever the depth read
+    /// emitted under it.</summary>
+    static IReadOnlyList<FieldValue> Lines(IReadOnlyList<FieldValue> fields, string path)
```
Avick3110: An unquantified path in the same call is now read at the fold's depth, which can be a large unannounced expansion.

`depth` is raised to `RowProjection.DefaultDepth` (4) for the whole call as soon as any path carries `[*]`, and `Lines` here returns `f.Path == path || IsUnder(f.Path, path)` — i.e. everything the depth-4 read emitted under the unquantified path, not just its own line.

Concrete: `project.fields=["Conditions[*]", "Effects"]` with no `project.depth`. Before, `Effects` was a one-line depth-1 summary; now it renders its full four-level expansion (every effect's Data sub-fields, one line each) purely because a different path was quantified. The `Conditions` fold is what asked for the depth, but the cost lands on `Effects`.

Worth deciding deliberately: either keep unquantified columns at the caller's own depth (read them on a separate pass), or say in the description that a quantifier raises the default depth for every path in the call.

src/housecarl-mcp/FieldFold.cs
```diff
@@ -0,0 +1,151 @@
...
+            // and it fails the record by name.
+            if (head is null || !head.Present || !head.Readable)
+            {
+                cols[i] = new[] { (head ?? new FieldValue(fold.Root, false, null, ReadEngine.AbsentNote, Present: false)) with { Path = fold.Requested } };
+                continue;
+            }
+            if (head.Count is null) return (null, NotAList(rec, fold, head));
+            cols[i] = fold.Fold == PathFold.Count
+                ? new[] { new FieldValue(fold.Requested, true, head.Count.Value.ToString(), null) }
+                : Elements(rows, rec.Fields, fold).ToList();
```
Avick3110: **high — an empty list under `[*]` vanishes with nothing said.**

An empty list reads back `Present: true, Count: 0` (`ReadEngine.LeafRead.Container`, "0 = present-but-EMPTY"), so it is not `head.Count is null` and never reaches `NotAList`. It falls to `Elements(...)`, which yields only paths where `RowProjection.RowKey(f.Path, roots) == f.Path` — the list's own summary line has a null `RowKey` — so the column comes back empty, and `Apply`'s `cols.SelectMany(...)` drops the requested path out of `Record.Fields` entirely. `ReadTools.RenderBatch` then prints the record's identity header and no field line at all.

So `project.fields=["Effects[*]"]` over a mixed selection is silently indistinguishable between "this record's list is empty" and "this path was never asked for". Under `format='dense'` it is worse in a quieter way: `elems` stays 1 and the cell is written `null`, which is the same cell a null value writes.

Two things make this look unintended rather than chosen:

- The class comment two lines above says "A root that resolved to something with no elements at all IS a misuse, and it fails the record by name." The code does not do that for an empty list — only for a non-list.
- The `rows` form, whose row shape this claims to reuse, does **not** drop it: `RowProjection.Fold` sends the list's own summary line through `Passthrough`, so `form='rows'` on an empty `Effects` still renders `Effects` with its count. Two spellings of the same row shape disagree on the empty case.

Either keep the list's summary line when the element set is empty (matching `rows`), or emit one line under the caller's own spelling saying the list holds no elements. Right now it is a silently degraded answer.

No test covers an empty list under `[*]`, or `[*count]` returning 0.

src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    [Fact]
+    public void CountOnANonListPathIsRefusedTheSameWay() =>
+        Assert.Contains("'EditorID[*count]' quantifies a LIST", Spell("EditorID[*count]"));
+
+    [Fact]
+    public void AnAbsentListIsTheReadsAnswerNotAMisuseOfTheToken() =>
+        // No Effects field on a WEAP at all: the read says so under the caller's own spelling.
+        Served(RecordsTools.Records(Svc, formids: new[] { Fid(W.Weapons[1]) }, project: Fields("Effects[*count]")),
+               "Effects[*count]");
```
Avick3110: **medium — this does not test what its name claims.**

`Served` is `Assert.Contains` (RecordsTestBase.cs:49), and the only thing named is `"Effects[*count]"` — the caller's own path, which the fold echoes back on *every* branch. The test passes whether the response says the field is absent, renders a blank cell, or renders `= 0`. The behaviour under review is the one it cannot see: that a missing list under `[*count]` says so rather than answering with a number.

The production code does the right thing (`Columns` carries `head`'s own note out, or `ReadEngine.AbsentNote` when the read emitted no line at all), so name that: assert the absent note, and assert the response does **not** contain `Effects[*count] = 0` — a missing field and an empty list are different answers and this is the test that has to keep them apart.

Same class of gap on `CountReadsTheListWithoutOpeningIt` (line 29): the name asserts a cost property — that the call runs at depth 1 and never expands the list — and the body asserts only the dense column list and the cell `"2"`. Nothing in it fails if the read expands to depth 4; it is a second rendering of `CountOnAListIsOneNumberPerRecord`. Either pin the depth (the plan's `Depth` is 1 for a count-only fold and observable) or give it the name it actually earns.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1553,7 +1570,7 @@ string RenderOff(SpillState? sp, out bool trunc) => json
             SpillState? spill = null;
             if (wantFile && outcome.Error is null)
             {
-                var (sp, aerr) = Artifacts.WriteCrossQuery(svc, outcome, null, false, false, 1, toFile!, "to_file", Echo(), LeverNames.Records);
+                var (sp, aerr) = Artifacts.WriteCrossQuery(svc, outcome, null, false, false, 1, toFile!, "to_file", Echo(), LeverNames.Records, fold: foldPlan);
```
Avick3110: **low — `fold:` is threaded into the one artifact call that cannot honour it, and the artifact lane has no test.**

This is the off-order scan's summary/aggregate spill: `fields` is `null` and `depth` is `1`. `Artifacts.WriteCrossQuery` runs `ResolveReadOn(..., fields, ..., depth)` and then `fold.Apply(o)` on the result, and at depth 1 a list renders as its summary line only — no element paths exist, so every `[*]` column folds to nothing and the spilled rows come back short. It is dead today (the `form != "fields"` gate above refuses a `foldPlan` before this lane is reachable), but it is dead the way a loaded gun is unloaded: the day that gate is widened, this writes a silently short artifact rather than refusing. Passing `null` here says what is true.

Separately, the PR body claims "the render, the artifact and the json document all see the same folded fields," and `RecordsFieldFoldTests` drives only the text render and `format='dense'`. There is no `to_file` test and no `format='json'` test for a quantified path, so the two lanes the sentence is about are the two with no coverage — and they take different routes (json/text go through the body lane and `Artifacts.WriteBatch`, which never sees `foldPlan`; dense goes through `WriteCrossQuery`, which does). One test per lane asserting the same element rows would pin the claim.

src/housecarl-mcp/FieldFold.cs
Avick3110: Folded.

`FoldPlan.Columns` now returns, beside the columns, the lines no column claims and no requested path covers — the read's own `"…"` note above all — and `Apply` puts them at the FRONT of `Record.Fields`. Front, not back: a render that hits its own ceiling part-way down the element rows would drop a note written after them, which is the same silent shortening one layer up.

When a `[*]` fold ran, the note carried out is `RowProjection.Passthrough`'s restated one ("the fold runs AFTER the read, so the elements past the cut are missing from the rows…"), because the engine's own remedy sends the caller to a depth that renders no rows. `format='dense'` has no field line to carry it, so it states the same note once per document as `read_note`, next to `owned_child_note`.

Tests over the fixture's over-budget FormList: `ATruncatedExpansionStillSaysSoBesideTheShortRows` (text, and it asserts the fold's own remedy wording) and `TheDenseDocumentCarriesTheCutToo`.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded, both halves, one each way.

Within ONE list a row is now keyed by its element INDEX: `FoldPlan.ElementIndex` reads the index off the cell's own path and dense fills row `e` from that map, so a sub-path column that emitted no line for an element leaves that row's cell null instead of shifting the next element's value onto it.

Across two DIFFERENT lists there is no shared element to key on, so `format='dense'` with two distinct `[*]` roots is refused by name at parse time — one row cannot be an element of both. The text lane still takes them, because there each line carries its own full path and nothing is being paired.

Tests: `ADenseCellSitsBesideItsOwnElementNeverTheNextOne` (`Effects[*]` + `Effects[*].Conditions[0]` on the fixture spell whose effect 0 carries no conditions — row 0's cell is null, row 1's is not) and `TwoDifferentListsCannotShareOneDenseRow`.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded.

The fold branch now registers per COLUMN before it writes the rows, keyed on the column's own list path (`fold.Root`) rather than a cell path — a `[*]` row's path is the element's, so keying on the cell would never match `OwnedChildFields`. An unquantified column falls back to its own cell path, which is the unfolded lane's key.

Test: `RecordsOwnedChildTests.ADenseFoldStatesTheSameClauseTheUnfoldedRowsDo` — `types=["CELL"]`, `format='dense'`, `project.fields=["Temporary[*]"]`, asserting the index-only clause names `Temporary`.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded, and not by copying the clause into the second place.

The forms the body lane renders are now one expression — `bodyLaneForm` — computed above the gate and read by both it and the lane itself, so the two cannot drift again. That also gives `form='rows'` the note, which as you say it was already missing.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded, both parts.

The flush and the cap test now run per ROW inside the element loop; a record cut part-way sets `truncated` and breaks out, and is NOT counted in `rendered` — a half-written record is not a rendered one.

`rendered` keeps its meaning (records), and `rows_rendered` is written beside it whenever a fold is in play, so a consumer reads the row count rather than inferring it.

Test: `DenseCapsTheElementRowsAndNotOnlyTheRecords` — the over-budget FormList under `Items[*]` with `max_chars=4000`, asserting the response length, `truncated`, and `rows_rendered` against the rows actually written.

src/housecarl-mcp/FieldFold.cs
Avick3110: Folded — the first option, and without the second read pass.

The plan now carries `CallerDepth` (the caller's own `project.depth`, else 1), and `Lines` keeps only the lines within that depth of an unquantified path: `Levels` counts one level per `.` or `[` after the path, which is the unit `project.depth` counts (depth 1 = its own line, 2 = each element's type, 4 = `Data.Magnitude`). So `project.fields=["Conditions[*]", "Effects"]` renders `Effects` exactly as it renders alone, while the quantified path still gets the depth its token needs.

The read itself is still the deeper one — the token's declared cost — but no column expands for a reason the caller did not ask that column for.

src/housecarl-mcp/FieldFold.cs
Avick3110: Folded — and you read the intent right: the class comment was wrong, not the requirement.

An empty list (`Present`, `Count 0`) now carries its OWN summary line out under the caller's spelling, which is exactly what `RowProjection.Fold` passes through for `form='rows'`, so the two spellings of the row shape agree on the empty case. Only a root that is not a list at all still fails the record by name; the comment is corrected to say that.

The neighbouring hole is closed the same way: a sub-path that NO element carries (list non-empty, column empty) gets one note saying so rather than an empty column. A requested path can no longer drop out of `Record.Fields`, in any lane.

Tests: `AnEmptyListRendersItsOwnLineAndNotNothing` (effect 0 has no conditions, effect 1 does — the same call proves both halves) and `CountOnAnEmptyListIsZero`.

src/housecarl-mcp-tests/RecordsFieldFoldTests.cs
Avick3110: Folded, both.

`AnAbsentListIsTheReadsAnswerNotAMisuseOfTheToken` now asserts the read's own "no field" note on that line, and that the response does NOT contain `Effects[*count] = 0` — the two answers it has to keep apart.

`CountReadsTheListWithoutOpeningIt` now proves the cost property on the one fixture that can observe it: the over-budget FormList truncates the moment it is expanded, so a count that reports 2001 with no truncation note demonstrably never opened the list. (The plan's `Depth` is not observable inline — the `depth` stamp is written into the artifact echo, not the response envelope — so the cut note is the witness.)

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded, both parts.

The off-order lane passes no fold now — `fields` is null and `depth` is 1 there, so passing the plan says something that is not true of that call.

And the two lanes the PR body's sentence is about now have a test each, both asserting the record's fields are exactly `Effects[0]`, `Effects[1]`: `TheJsonDocumentCarriesTheSameElementRows` (`format='json'`, the body lane through `JsonWire.RenderBatch`) and `RecordsFieldFoldArtifactTests.TheArtifactCarriesTheSameElementRows` (`to_file`, through `Artifacts.WriteBatch`).

src/housecarl-mcp/FieldFold.cs
```diff
@@ -0,0 +1,198 @@
...
+                any = true;
+            }
+        }
+        if (!any) return (null, null);
+        // The depth the TOKENS require: an element lives one level under its list, and each sub-path step after
+        // the token is one more. A [*count]-only plan needs no expansion at all — the list's own line carries the
+        // count. The caller's own depth is folded in above this, where it can still be deeper.
+        int need = 1;
+        foreach (var f in folds)
+            if (f is { Fold: PathFold.Set }) need = Math.Max(need, 2 + f.Tail.Length);
```
Avick3110: `need = 2 + f.Tail.Length` counts one expansion level per tail SEGMENT, but a segment that carries a bracket index is two levels in `ReadEngine.Expand` (one for the `.` step, one for the `[`). `FoldPlan.Levels` right below already counts it the other way — one per `.` or `[` — so the two halves of this file disagree on what a level is.

A node at level L below the read root needs depth L+1. `Effects[*].Data.Magnitude` is level 3, need computes 4, fine. `Effects[*].Conditions[0].Data` is level 4 and needs depth 5, but Tail.Length is 2 so need is 4 — and the default for a `[*]` plan is also 4, so `Math.Max` does not save it. The read stops one level short, `Elements` matches nothing, and since the list is non-empty the record answers with the `NoElement` sentence: "(no element of 'Effects' carries 'Conditions[0].Data' — the list holds 2)". That is a confident, wrong answer.

The shallower `Effects[*].Conditions[0]` (need 3, true need 4) hits it too as soon as the caller passes `project.depth=2` or `3` — `ADenseCellSitsBesideItsOwnElementNeverTheNextOne` only passes because the depth default of 4 covers for it.

src/housecarl-mcp/FieldFold.cs
```diff
@@ -0,0 +1,198 @@
...
+                if (bare.Length == 0)
+                    return (null, $"project.fields path '{path}': '{segs[s]}' has no field name before '[' — a quantifier binds to a list field, e.g. 'Conditions[*]'.");
+                if (fold == PathFold.None)
+                    return (null, $"project.fields path '{path}': '[{key}]' is not a quantifier — the tokens are [*], [*count], [*any], [*all] and [*none] (the last three fold to a boolean and belong in where=).");
+                if (fold is PathFold.Any or PathFold.All or PathFold.NoneOf)
+                    return (null, $"project.fields path '{path}' folds the elements into a boolean, and a boolean is not a row — use {PathFoldGrammar.Token(fold)} in where= to SELECT the records, and [*] here to read one row per element.");
+                if (fold == PathFold.Count && s != segs.Length - 1)
+                    return (null, $"project.fields path '{path}': nothing can follow '[*count]' — it yields how MANY elements there are, not an element to step into.");
+                folds[i] = new FieldFold(path, string.Join(".", segs[..s].Append(bare)), segs[(s + 1)..], fold);
+                readPaths[i] = folds[i]!.Root;
```
Avick3110, line 214: Every quantified path rewrites its read path to the list `Root`, so two columns that quantify the same list hand `ReadFields` the same path twice. `ReadEngine.ReadFields` does not de-duplicate `targets`, and it allocates `int budget = MaxExpandNodes` ONCE for the whole record (ReadEngine.cs:226) before looping `foreach (var p in targets)`.

So `project.fields=["Effects[*]", "Effects[*].Data.Magnitude"]` walks `Effects` twice into one 2000-line budget. On a large list the second walk pushes past the cut: the caller gets the expansion-truncation note and roughly half the element rows the same call with a single column would return. The fold's own de-duplication (`RowProjection.Fold`'s `seen`, `Elements`' `seen`) hides the duplicate LINES, so nothing in the output says why the rows are short.

Distinct-ing `readPaths` before the read (keeping the per-column mapping) would fix both the double walk and the halved budget.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -293,16 +303,31 @@ public static string Records(
...
-        int depth = project?.depth is { } d && d > 0 ? d : (form == "rows" ? RowProjection.DefaultDepth : 1);
+        bool foldsElements = foldPlan?.RendersElements ?? false;
+        int depth = project?.depth is { } d && d > 0 ? d : (form == "rows" || foldsElements ? RowProjection.DefaultDepth : 1);
+        // A sub-path after [*] names exactly what to read, so reaching it is the token's own requirement rather
+        // than expansion the caller has to budget for.
+        // CallerDepth is what an UNQUANTIFIED column beside a quantified one renders at: the token raises the
+        // read's depth for the paths that need it, and a sibling path must not expand deeper for that reason.
+        if (foldPlan is not null)
+        {
+            depth = Math.Max(depth, foldPlan.Depth);
```
Avick3110, line 314: `CallerDepth` trims the RENDER (`FoldPlan.Lines`) but not the READ. `ReadEngine.ReadFields` takes one `depth` and applies it to every requested path, off one shared `MaxExpandNodes` budget — so raising `depth` here expands the unquantified siblings too, and their extra lines are then thrown away.

`project.fields=["VirtualMachineAdapter", "Effects[*]"]` with no `project.depth`: `foldsElements` makes depth 4, so VMAD is walked to depth 4 (a script-heavy record is hundreds of lines) before the walk reaches `Effects`. `Lines` then keeps exactly one line for it — the depth-1 summary. The budget is spent either way, so the `Effects` walk can hit the cut and the caller gets truncated element rows for a list that would have folded whole had they asked for `Effects[*]` alone. Reversing the two entries changes the answer, which is the tell.

Reading the quantified paths at the raised depth and the sibling paths at `CallerDepth` (two reads, or a per-path depth) would make the render's claim true of the read as well.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -1231,10 +1234,68 @@ public static string RenderCrossQueryDense(LoadOrderService svc, CrossQueryOutco
...
+                            for (int c = 0; c < cols.Length; c++)
+                            {
+                                if (fold.Folds[c] is not { Fold: HousecarlCore.PathFold.Set } fc) continue;
+                                var map = byIndex[c] = new Dictionary<int, HousecarlCore.FieldValue>();
+                                for (int k = 0; k < cols[c].Count; k++)
+                                {
+                                    int idx = FoldPlan.ElementIndex(cols[c][k].Path, fc.Root);
+                                    map[idx < 0 ? k : idx] = cols[c][k];
+                                }
+                                foreach (var idx in map.Keys) elems = Math.Max(elems, idx + 1);
```
Avick3110: `elems` is the largest parsed index + 1, and `FoldPlan.ElementIndex` parses the bracket segment as a decimal int. That holds for a positional list, but `ReadEngine.Expand` emits DICT elements as `path[<key>]`, and a dict passes the `head.Count is null` list check in `FoldPlan.Columns` (its `deepCount` is the pair count).

A Package's `Data` is a dict of `APackageData` arms keyed by int — `ElementRefusalRemedyTests` uses `Data[7]`. `format='dense'` with `project.fields=["Data[*]"]` on a package whose keys are {2, 7} gives map keys {2, 7}, so `elems` is 8: the document emits 8 rows for a 2-pair dict, 6 of them with every cell null, and `rows_rendered` says 8.

The other half: a non-numeric key makes `ElementIndex` return -1 and `map[idx < 0 ? k : idx]` falls back to `k`, the position in the column — which is exactly the pairing the comment three lines up says must never happen, and which `ADenseCellSitsBesideItsOwnElementNeverTheNextOne` pins for list roots. A sub-path column that skips an element then lands its cell on a different element's row.

Keying rows by the bracket segment as a STRING (in first-seen order) would cover list and dict alike.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1162,9 +1187,12 @@ List<KeyValuePair<string, string>> Echo()
             // The reverse-reference index's accounting belongs to the response whatever consumes the scan. The two
             // CrossQuery renderers read it off the outcome themselves; every form BELOW renders its own pipeline's
             // response and would drop it, so it rides their header and envelope. One place, so no form forgets.
+            // The forms whose bodies the batch lane below renders: the note rides their envelope, so the gate
+            // and the lane read ONE condition and cannot drift on which forms those are.
+            bool bodyLaneForm = form == "everything" || form == "rows"
```
Avick3110, line 1312: Lower confidence, but worth a look: adding a quantifier now moves `form='fields'` off the CrossQuery render and onto the body lane, and the two lanes do not carry the same columns. `JsonWire.RenderCrossQuery` emits a per-row `matches` cell for a multi-target `references=` (JsonWire.cs:1107); `JsonWire.RenderBatch` has no such column.

So `references=["A","B"]`, `form='fields'`, `format='json'` tells the caller which target each row matched — until they add `project.fields=["Effects[*count]"]`, at which point the column disappears with nothing in the response saying so. The same is already true of `everything`/`rows`, so this may be a deliberate widening of an existing gap rather than a new one; flagging it because a quantifier reads like a projection choice, not a lane choice.

src/housecarl-mcp/FieldFold.cs
Avick3110: Real, folded in 67a35bd3. The two halves did disagree: `FoldPlan.Levels` counts one per `.` or `[`, and the need counted one per tail SEGMENT. `FieldFold.TailLevels` now counts the same way — one per dotted step plus one per bracket in it — so `Effects[*].Conditions[0].Data` needs 5 and `Effects[*].Conditions[0]` needs 4.

Both cases confirmed before the fix: `Effects[*].Conditions[0].Data` answered with the `NoElement` sentence at the default depth, and `Effects[*].Conditions[0]` did the same at `project.depth=2`. Two tests pin them (`AnIndexedSubPathIsReadDeepEnoughToReachIt`, `TheTokensOwnDepthWinsOverALowerOneTheCallerNamed`); both fail on the old arithmetic.

src/housecarl-mcp/FieldFold.cs
Avick3110: Real and folded in fa895db3, though not with the symptom you predicted — worth stating precisely because the two look alike.

The double walk is real: `readPaths` repeated the list root once per column, `ReadEngine.ReadFields` does not de-duplicate `targets`, and the `MaxExpandNodes` budget is allocated once for the whole record. But the rows do not come back halved. The budget is spent in order, so the FIRST walk fills it and the second emits nothing, and both walks emit the same paths, which the fold's `seen` sets then collapse. The rows are therefore the same rows a single-column call renders. What the caller actually gets is a list that fits the budget once but not twice — say 1,200 lines — coming back with an "expansion truncated" note attached to an answer that is complete. A cut the read names that the answer did not suffer is the same defect from the other side, so the fix is the one you named.

`FoldPlan.Read()` now hands the read each distinct path once, and the per-column arrays keep their own mapping (`FoldPlan.Paths`), so nothing about the columns changes. `OneListQuantifiedTwiceIsReadOnceAndSiblingsAtTheCallersDepth` pins it.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, folded in fa895db3 alongside the de-duplication — they are the same computation, so `FoldPlan.Read()` returns the distinct paths and the depth each one needs together.

`ReadEngine.ReadFields` takes an optional `depths` array parallel to `paths`: a quantified path reads at the plan's depth, an unquantified one at `CallerDepth`, and a path both columns name takes the higher. It is threaded through `ResolveRead`/`ResolveReadOn`/`ResolveBatch`/`ResolveBatchFromPole`/`OverlayPostBatch`, so every lane the fold reaches — formids, scan body, dense, artifact — reads the same way. A mismatched array is ignored rather than applied, so it can never quietly read a path at the wrong depth. Sibling rendering is unchanged: with a fold present those paths were already going through the deep branch, and `Expand` at depth 1 emits exactly the lines `Lines(..., CallerDepth)` was keeping.

`AnUnquantifiedColumnIsReadAtTheCallersOwnDepth` pins it on the over-budget FormList: `project.fields=["Items","Effects[*]"]` used to come back with an "expansion truncated" note for a column that renders one collapsed line. It fails without the per-path depths.

src/housecarl-mcp/JsonWire.cs
Avick3110: Real, folded in c07ca4fc. Both halves reproduce.

`FoldPlan.ElementIndex` is gone; `FoldPlan.ElementKey` returns the bracket segment as TEXT, and the dense render lays one row per distinct key rather than per index. Keys that all parse as integers are ordered numerically, so a positional list reads in index order whatever order the columns are named in; anything else keeps read order. The positional fallback for a key the render cannot parse is gone with it, so a sub-path column can no longer land a cell on another element's row.

The fixture now carries a package whose `Data` holds arms keyed 2 and 7. `ADictListLaysOneRowPerPairAndNotOnePerKeyValue` asserts two rows and `rows_rendered=2`; on the old keying it renders eight, six of them empty. `AnElementKeyIsTheBracketSegmentItself` pins the key itself. The changelog line said "keyed by its element index" and now says what it actually is.

src/housecarl-mcp/RecordsTools.cs
Avick3110: The gap is real, but it is the batch lane's and not the quantifier's, so I have changed nothing here — flagging it for you rather than widening this PR.

The evidence: on `main`, `RecordsTools.cs:1244` reads `form == "everything" || form == "rows" || (form == "fields" && scopePlusPole)`. All three already route to `JsonWire.RenderBatch`, which carries no `matches` — so `references=["A","B"]` with `form='rows'` already loses the column today, without any token. This PR adds `foldPlan is not null && !dense` to that same condition; the quantifier joins the lane, it does not create the gap. (The artifact lane is not affected either way: `Artifacts.WriteCrossQuery`'s schema carries `matches?` and it fills it.)

Fixing it properly means giving the batch body lane a per-row `matches` — which lands on `everything`, `rows` and scoped `fields` at the same time, and is a change to a shared render rather than to this feature. Happy to open an issue for it if you want it done; I did not want to fold four forms' worth of render change into the fields-fold PR without your word.

PR #575, opened 2026-09-05
docs(spid-authoring): verifying a distribution actually happened

### LINKED ISSUE #383: spid-authoring: nothing on verifying a distribution actually landed

Avick3110:
The `spid-authoring` skill said nothing about positively verifying that a rule landed on an actor, which is where several SPID behaviours bite. This lands juggernaunt46-star's PR #384 text with corrections. The new "Verifying a distribution actually happened" section says that a parsed `_DISTR.ini` and a clean SPID log show what SPID looked up, not what an actor received, and then what to assert instead: the player is never a distribution target; a dynamically created keyword has no plugin FormID but is still found by name from a script; and sampling an interior cell with known occupants is testing practice, marked as such rather than as a SPID fact. The pre-existing-save bullet from #384 is rewritten because it was wrong as written — SPID redistributes from scratch every launch, so an old save's actors do get a newly added rule; the real constraint is that the INIs are read once per game launch, so a rule added mid-session needs a restart and the actor has to load in that new session. The section sits after "Common mistakes" per the section order in `standards/HOUSECARL_SKILL_AUTHORING.md` §4.2, and cites the `papyrus-reference` skill for the `Keyword.GetKeyword` signature per §4.4.

The facts themselves are grounded in the reference corpus rather than asserted only in the skill body: player exclusion and the once-per-launch config read go into `references/grammar-core.md` §3, the dynamic keyword's missing plugin FormID and its reachability by name go into `references/form-types.md`, both with the corpus's existing **[source]** marker naming SPID's own files (`DistributeManager.cpp` `should_process_NPC`, `FormData.h` `kCreateIfMissing`, `main.cpp` config timing). `_CORPUS_STATUS.md` no longer claims "Gaps: none in the article's scope" — article 6617 covers grammar, not verification behaviour — and lists the new source-resolved facts and the files they came from.

Docs only, no build. Frontmatter and description are untouched, so the trigger surface is unchanged and the eval set is unaffected; no eval re-run per §6.5, which gates re-measurement on a changed description. `SKILL.md` is 158 lines, under the §4.1 cap. No houseCARL tool is named in the added text.

Closes #383

### DIFF
--- a/.claude/skills/spid-authoring/SKILL.md
+++ b/.claude/skills/spid-authoring/SKILL.md
@@ -139,6 +139,14 @@ what SPID looked up; only a live actor shows what SPID applied.
   it by name, which is what makes a no-ESP keyword distribution verifiable from script at all. For the
   signature, look it up in the `papyrus-reference` skill; the SPID side
   is in `references/form-types.md`.
+- **Chance and a Level Filter skip actors by design.** A `Chance` below 100 is *supposed* to miss some
+  NPCs, so two sampled actors without the form is an ordinary outcome for `|||||50`, not a broken
+  rule — and the roll is re-made each session unless `!` pins it per NPC (`!` makes re-sampling the
+  same actor repeatable; it does not make the form land). A Level Filter does the same
+  deterministically: the line moves to the Leveled Distribution pass and only auto-leveled NPCs whose
+  level/skills fall in the range receive it. Before calling a rule failed, set Chance to 100 and check
+  the actors you sampled pass the Level Filter (`references/grammar-core.md` §10,
+  `references/filters.md` §3).
 - **Sample somewhere NPCs must be** — testing practice, not a SPID rule. An interior cell with known
   occupants beats an exterior spawn marker: finding no NPC within scan range there is inconclusive
   rather than a failing distribution.
--- a/.claude/skills/spid-authoring/SKILL.md
+++ b/.claude/skills/spid-authoring/SKILL.md
@@ -125,9 +125,10 @@ SPID documentation.
 A parsed `_DISTR.ini` and a clean SPID log are **not** proof that an NPC got the thing. The log shows
 what SPID looked up; only a live actor shows what SPID applied.
 
-- **SPID does not distribute to the player.** Every distribution path is gated on a check that
-  excludes the player, so asserting on `Game.GetPlayer()` reports failure for a rule that is working
-  perfectly — sample NPCs instead (`references/grammar-core.md` §3).
+- **SPID does not distribute to the player.** The ordinary on-load path is gated on an explicit
+  `IsPlayer()` check, and the other paths into `Distribute()` exclude the player by their own guards,
+  so asserting on `Game.GetPlayer()` reports failure for a rule that is working perfectly — sample
+  NPCs instead (`references/grammar-core.md` §3).
 - **A newly added rule reaches old saves, but not mid-session.** SPID distributes from scratch each
   launch and writes nothing into the save, so actors in a pre-existing save do get a rule added
   afterwards (`references/grammar-core.md` §3). What bites is that the INIs are read **once per game
--- a/.claude/skills/spid-authoring/SKILL.md
+++ b/.claude/skills/spid-authoring/SKILL.md
@@ -128,14 +128,15 @@ what SPID looked up; only a live actor shows what SPID applied.
 - **SPID does not distribute to the player.** Every distribution path is gated on a check that
   excludes the player, so asserting on `Game.GetPlayer()` reports failure for a rule that is working
   perfectly — sample NPCs instead (`references/grammar-core.md` §3).
-- **A newly added rule reaches old saves, but not mid-session.** SPID redistributes from scratch each
-  launch, so actors in a pre-existing save do get a rule added afterwards. What bites is that the INIs
-  are read **once per game launch**: a rule added while the game is running needs a restart, and the
-  actor has to load in that new session for the lazy per-actor distribution to run.
+- **A newly added rule reaches old saves, but not mid-session.** SPID distributes from scratch each
+  launch and writes nothing into the save, so actors in a pre-existing save do get a rule added
+  afterwards (`references/grammar-core.md` §3). What bites is that the INIs are read **once per game
+  launch**: a rule added while the game is running needs a restart, and the actor has to load in that
+  new session for the lazy per-actor distribution to run.
 - **A dynamic keyword IS reachable.** A keyword SPID creates at runtime has no plugin FormID — nothing
   in xEdit to point at — but it is in the game's keyword array, so `Keyword.GetKeyword("<name>")` finds
   it by name, which is what makes a no-ESP keyword distribution verifiable from script at all. For the
-  signature, look it up in the `papyrus-reference` skill (`references/skse/Keyword.md`); the SPID side
+  signature, look it up in the `papyrus-reference` skill; the SPID side
   is in `references/form-types.md`.
 - **Sample somewhere NPCs must be** — testing practice, not a SPID rule. An interior cell with known
   occupants beats an exterior spawn marker: finding no NPC within scan range there is inconclusive
--- a/.claude/skills/spid-authoring/SKILL.md
+++ b/.claude/skills/spid-authoring/SKILL.md
@@ -120,6 +120,27 @@ SPID documentation.
   one of them (text → String, form → Form, level/skill → Level, trait → Trait). Look it up in
   `filters.md` rather than assuming a filter exists.
 
+## Verifying a distribution actually happened
+
+A parsed `_DISTR.ini` and a clean SPID log are **not** proof that an NPC got the thing. The log shows
+what SPID looked up; only a live actor shows what SPID applied.
+
+- **SPID does not distribute to the player.** Every distribution path is gated on a check that
+  excludes the player, so asserting on `Game.GetPlayer()` reports failure for a rule that is working
+  perfectly — sample NPCs instead (`references/grammar-core.md` §3).
+- **A newly added rule reaches old saves, but not mid-session.** SPID redistributes from scratch each
+  launch, so actors in a pre-existing save do get a rule added afterwards. What bites is that the INIs
+  are read **once per game launch**: a rule added while the game is running needs a restart, and the
+  actor has to load in that new session for the lazy per-actor distribution to run.
+- **A dynamic keyword IS reachable.** A keyword SPID creates at runtime has no plugin FormID — nothing
+  in xEdit to point at — but it is in the game's keyword array, so `Keyword.GetKeyword("<name>")` finds
+  it by name, which is what makes a no-ESP keyword distribution verifiable from script at all. For the
+  signature, look it up in the `papyrus-reference` skill (`references/skse/Keyword.md`); the SPID side
+  is in `references/form-types.md`.
+- **Sample somewhere NPCs must be** — testing practice, not a SPID rule. An interior cell with known
+  occupants beats an exterior spawn marker: finding no NPC within scan range there is inconclusive
+  rather than a failing distribution.
+
 ## Notes
 
 - **Provenance.** The `references/` corpus is reconstructed from "SPID: The Complete Reference"
--- a/.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
+++ b/.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
@@ -75,8 +75,10 @@ only — no code vendored. Files consulted (all `SPID/src/`): `LookupConfigs.cpp
 chain), `LookupConfigs.h` (`TraitsFilterComponentParser`, `ChanceComponentParser`,
 `LevelFiltersComponentParser`), `Defs.h` (`Traits` struct), `DistributeManager.cpp`
 (`detail::should_process_NPC`, the load hooks), `FormData.h` (`kCreateIfMissing`), `main.cpp`
-(config read and lookup timing). MIT (unlike SkyPatcher's unlicensed repo)
-would permit vendoring, but the corpus documents grammar, it doesn't embed source.
+(config read and lookup timing). The verification-behaviour files were read at commit
+`6e66908` (`master`, 2026-09-02); the earlier parsing files at `master` as of 2026-06-02. MIT (unlike
+SkyPatcher's unlicensed repo) would permit vendoring, but the corpus documents grammar, it doesn't
+embed source.
 
 ## Structure
 
--- a/.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
+++ b/.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
@@ -40,7 +40,7 @@ the per-record files.
 | Distributable form types | **10 / 10** (`form-types.md`) |
 | Filter sections | **4 / 4** (`filters.md`) |
 | Flat enums | skill indices, trait letters, package-list types, form signatures, distribution order, defaults (`value-tables.md`) |
-| **Gaps** | **The article does not cover verification behaviour.** Four details a user hits when checking whether a rule landed — player exclusion, the once-per-launch config read, the dynamic keyword's missing plugin FormID, and its reachability by name — are absent from article 6617 and were resolved from source (marked **[source]**; see Confidence). Two other article-*silent* details (comment syntax, whitespace tolerance) came from source the same way. Expect further gaps outside the article's grammar scope. |
+| **Gaps** | **The article does not cover verification behaviour.** Five details a user hits when checking whether a rule landed — player exclusion, the once-per-launch config read, the runtime `SPID_Processed` marker, the dynamic keyword's missing plugin FormID, and its reachability by name — are absent from article 6617 and were resolved from source (marked **[source]**; see Confidence). Two other article-*silent* details (comment syntax, whitespace tolerance) came from source the same way. Expect further gaps outside the article's grammar scope. |
 
 ## Confidence
 
@@ -61,6 +61,10 @@ the per-record files.
     article.
   - **Configs are read once per game launch** (`main.cpp`, `kPostLoad`/`kDataLoaded`), so a rule added
     mid-session needs a restart. Not in the article.
+  - **The "already handled this NPC" marker is a runtime keyword, not a persisted one** — `Setup()` in
+    `DistributeManager.cpp` creates `SPID_Processed` via `IFormFactory` each launch (declared in
+    `DistributeManager.h`), and `distribute_on_load` skips an NPC that already carries it. The
+    article states the from-scratch-per-launch behaviour; the mechanism behind it is source-only.
   - **A dynamically created keyword has no plugin FormID but is pushed into the game's keyword array**
     (`FormData.h`, `kCreateIfMissing`). The other half — that SKSE's `Keyword.GetKeyword` resolves a
     name against that same array — is read from SKSE's own source (`ianpatt/skse64`,
@@ -79,11 +83,17 @@ the per-record files.
 only — no code vendored. Files consulted (all `SPID/src/`): `LookupConfigs.cpp` (`sanitize()`, parser
 chain), `LookupConfigs.h` (`TraitsFilterComponentParser`, `ChanceComponentParser`,
 `LevelFiltersComponentParser`), `Defs.h` (`Traits` struct), `DistributeManager.cpp`
-(`detail::should_process_NPC`, the load hooks), `FormData.h` (`kCreateIfMissing`), `main.cpp`
-(config read and lookup timing). The verification-behaviour files were read at commit
-`6e66908` (`master`, 2026-09-02); the earlier parsing files at `master` as of 2026-06-02. MIT (unlike
-SkyPatcher's unlicensed repo) would permit vendoring, but the corpus documents grammar, it doesn't
-embed source.
+(`detail::should_process_NPC`, `Setup()`, the load hooks), `DistributeManager.h` (the `processed`
+keyword declaration), `DistributePCLevelMult.cpp` and `DeathDistribution.cpp` (the remaining
+`Distribute()` call sites and their guards), `Distribute.cpp` (the call-site inventory),
+`FormData.h` (`kCreateIfMissing`), `main.cpp` (config read and lookup timing). The
+verification-behaviour files were read at commit `6e66908` (`master`, 2026-09-02); the earlier parsing
+files at `master` as of 2026-06-02. MIT (unlike SkyPatcher's unlicensed repo) would permit vendoring,
+but the corpus documents grammar, it doesn't embed source.
+
+One fact reaches outside SPID's repo: **SKSE** (`ianpatt/skse64`, `master` `4cd2e34`, read 2026-09-05)
+— `skse64/PapyrusKeyword.cpp`, for how `Keyword.GetKeyword` resolves a name. Cited in `form-types.md`
+so the by-name-reachability claim does not span an unread seam.
 
 ## Structure
 
--- a/.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
+++ b/.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
@@ -62,8 +62,10 @@ the per-record files.
   - **Configs are read once per game launch** (`main.cpp`, `kPostLoad`/`kDataLoaded`), so a rule added
     mid-session needs a restart. Not in the article.
   - **A dynamically created keyword has no plugin FormID but is pushed into the game's keyword array**
-    (`FormData.h`, `kCreateIfMissing`), which is why SKSE's `Keyword.GetKeyword` finds it by name. The
-    article documents dynamic creation but neither consequence.
+    (`FormData.h`, `kCreateIfMissing`). The other half — that SKSE's `Keyword.GetKeyword` resolves a
+    name against that same array — is read from SKSE's own source (`ianpatt/skse64`,
+    `skse64/PapyrusKeyword.cpp`), not inferred; its one-shot, never-invalidated cache is recorded as a
+    caveat in `form-types.md`. The article documents dynamic creation but neither consequence.
 - **One residual confidence caveat:** comment syntax. SPID parses configs via **CSimpleIniA**
   (source-confirmed); the **`;`** line-comment character is CSimpleIni's documented *library default*
   rather than a SPID line of code we read. Stated in the corpus as "standard INI `;` comments (via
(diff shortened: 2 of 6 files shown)


.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
```diff
@@ -54,6 +54,13 @@ the per-record files.
   - **Whitespace around `|` and `,` is stripped**, **`" - "` ⇒ `~`** (xEdit paste form), and FormID
     zero-padding is forgiven — from `sanitize()`.
   - **Chance `!`** deterministic flag and **Level `w`** weight prefix — confirmed in the parsers.
+  - **The player is never distributed to** — `should_process_NPC` in `DistributeManager.cpp` gates
+    every path on `!IsPlayer() && !IsDeleted()`. Not in the article.
+  - **Configs are read once per game launch** (`main.cpp`, `kPostLoad`/`kDataLoaded`), so a rule added
+    mid-session needs a restart. Not in the article.
+  - **A dynamically created keyword has no plugin FormID but is pushed into the game's keyword array**
+    (`FormData.h`, `kCreateIfMissing`), which is why SKSE's `Keyword.GetKeyword` finds it by name. The
+    article documents dynamic creation but neither consequence.
```
Avick3110: The three bullets land under a preamble that reads "Source-resolved beyond the article (the no-guesswork wins, **all in `source-notes.md`**)", and the "Files consulted" list below now names `DistributeManager.cpp`, `FormData.h` and `main.cpp`. Neither is true of the capture: `dev/references/SPID/source-notes.md` has no mention of `should_process_NPC`, `IsPlayer`, `kCreateIfMissing`, `kPostLoad` or `kDataLoaded`, and its header records a single 2026-06-02 fetch of `LookupConfigs.cpp`/`LookupConfigs.h`/`Defs.h`.

That file states its own invariant: "This file records exactly what was read and where, so every corpus claim traces to either the article or this excerpt." With this change, four **[source]**-marked claims trace to nothing a reader can check, and the preamble asserting otherwise is now false. Concrete failure: on the next SPID version bump someone re-derives the corpus against `source-notes.md`, finds no record of the three new files, and cannot tell whether the player-exclusion and once-per-launch facts were verified or written from memory. Add the excerpts to `source-notes.md` (and its fetch line) in the same shape as the `Traits` and trait-letter blocks, or drop the **[source]** markers.

.claude/skills/spid-authoring/references/grammar-core.md
```diff
@@ -52,7 +52,14 @@ comment stripping are CSimpleIniA's, not SPID's own code.)
...
 
 **When:** distribution is **lazy** — it runs when NPCs are loaded into the world (typically on cell
-load), not all at once at game start.
+load), not all at once at game start. The configs themselves are read **once per game launch**, so an
+INI edited while the game is running takes effect only after a restart. **[source]**
+(`SPID/src/main.cpp` — configs are read at `kPostLoad` and forms looked up at `kDataLoaded`.)
+
+**Who — never the player.** Every distribution path is gated on
+`detail::should_process_NPC`, which is `!a_npc->IsPlayer() && !a_npc->IsDeleted()`, so the player
+character is not a distribution target and never receives a SPID form. **[source]**
```
Avick3110: This quotes an exact C++ expression — "`detail::should_process_NPC`, which is `!a_npc->IsPlayer() && !a_npc->IsDeleted()`" — as if verbatim, but unlike every other **[source]** claim in this corpus there is no verbatim excerpt behind it anywhere in the repo (`dev/references/SPID/source-notes.md` carries the `Traits` struct and the trait-letter switch verbatim; it has no `should_process_NPC`).

The risk is specific, not stylistic: in SPID the player skip and the "already processed" keyword guard are not one expression, so if `should_process_NPC` is not literally that conjunction, a reader who greps the MIT source for it finds nothing and the **[source]** marker — the thing that makes the rest of this corpus trustworthy — is what they stop believing. Either paste the real body into `source-notes.md` and cite it, or state the behaviour ("the player is excluded before any distribution runs") without quoting code.

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -120,6 +120,27 @@ SPID documentation.
...
 
+## Verifying a distribution actually happened
+
+A parsed `_DISTR.ini` and a clean SPID log are **not** proof that an NPC got the thing. The log shows
+what SPID looked up; only a live actor shows what SPID applied.
+
+- **SPID does not distribute to the player.** Every distribution path is gated on a check that
+  excludes the player, so asserting on `Game.GetPlayer()` reports failure for a rule that is working
+  perfectly — sample NPCs instead (`references/grammar-core.md` §3).
+- **A newly added rule reaches old saves, but not mid-session.** SPID redistributes from scratch each
```
Avick3110: The load-bearing half of this bullet — "SPID redistributes from scratch each launch, so actors in a pre-existing save do get a rule added afterwards" — is not in the corpus. `grammar-core.md` §3 gained only "the configs are read once per game launch", and that does not entail anything about pre-existing saves: a config read at `kPostLoad` is consistent with an implementation that never revisits an already-distributed actor. Nothing in `grammar-core.md`, `form-types.md` or `_CORPUS_STATUS.md` states the from-scratch-per-launch behaviour, and this bullet carries no **[source]** marker.

That makes it exactly the case the skill's own "Bundled-or-warn — never invent SPID grammar" section (line 91) covers: "If a form type, filter, trait, or **behavior** isn't in it, say so — don't fabricate a plausible token." Concrete cost: a user asks whether they need a new game after adding a `_DISTR.ini` rule, Claude answers "no, SPID redistributes from scratch" with the confidence of a sourced fact, and if the underlying reasoning is wrong there is no corpus entry to catch it. Add the fact to `grammar-core.md` §3 with its source alongside the once-per-launch line, and have this bullet point at it like the player bullet does.

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -120,6 +120,27 @@ SPID documentation.
...
+  launch, so actors in a pre-existing save do get a rule added afterwards. What bites is that the INIs
+  are read **once per game launch**: a rule added while the game is running needs a restart, and the
+  actor has to load in that new session for the lazy per-actor distribution to run.
+- **A dynamic keyword IS reachable.** A keyword SPID creates at runtime has no plugin FormID — nothing
+  in xEdit to point at — but it is in the game's keyword array, so `Keyword.GetKeyword("<name>")` finds
+  it by name, which is what makes a no-ESP keyword distribution verifiable from script at all. For the
+  signature, look it up in the `papyrus-reference` skill (`references/skse/Keyword.md`); the SPID side
+  is in `references/form-types.md`.
```
Avick3110, line 141: `standards/HOUSECARL_SKILL_AUTHORING.md` §4.4 is explicit that a cross-skill reference is by skill name only: "Other Housecarl skills: reference by skill name in backticks. **Don't link by relative path — paths drift across plugin/repo split.**" The eight existing cross-references to `papyrus-reference` in `papyrus-optimization` and `skse-plugin-authoring` all follow that and name no path.

Beyond the rule, the path is actively misleading here: read inside `spid-authoring`, `references/` is *this* skill's own directory, which contains `grammar-core.md`, `form-types.md`, `filters.md`, `value-tables.md` and `index.jsonl` — no `skse/`. The same bullet then writes `references/form-types.md` meaning the local dir, so the two `references/` in one sentence resolve to different skills. A reader following the first one greps for a file that does not exist where it is looking.

```suggestion
  it by name, which is what makes a no-ESP keyword distribution verifiable from script at all. For the
  signature, look it up in the `papyrus-reference` skill; the SPID side
  is in `references/form-types.md`.
```

.claude/skills/spid-authoring/references/grammar-core.md
```diff
@@ -52,7 +52,14 @@ comment stripping are CSimpleIniA's, not SPID's own code.)
 ## 3. When distribution happens, and in what order
 
 **When:** distribution is **lazy** — it runs when NPCs are loaded into the world (typically on cell
-load), not all at once at game start.
+load), not all at once at game start. The configs themselves are read **once per game launch**, so an
+INI edited while the game is running takes effect only after a restart. **[source]**
```
Avick3110, line 56: New facts landed in `grammar-core.md` §3 and `form-types.md`, but `references/index.jsonl` was not updated. `SKILL.md:46` sells that file as the way in — "a grep-friendly router ... Grep it to jump straight to the right section" — and its 27 entries have no topic covering distribution timing, the player exclusion, or the dynamic keyword's script reachability. Grepping it for `player` hits only the Trait-filter "Player's Teammate" rows; `restart`, `launch`, `GetKeyword` and `verify` hit nothing at all.

So the routed path lands a reader on `filters.md` (Player's Teammate) for a player question, and the actual answer in §3 is only reachable by reading `grammar-core.md` whole — which the First-step section explicitly tells them not to do ("don't bulk-load everything"). Two lines fix it, e.g. a `{"topic": "when distribution happens", "kind": "grammar", "file": "grammar-core.md", "aliases": ["lazy", "cell load", "once per launch", "restart", "player excluded", "never the player"]}` entry, and `"GetKeyword"`/`"no plugin FormID"` added to the existing `Keyword` row's aliases.

.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
Avick3110: Folded. `dev/references/SPID/source-notes.md` now carries the verbatim excerpts the preamble promised: `detail::should_process_NPC` with `distribute_on_load` and the `DistributeManager.h` declaration, `Setup()` creating the `SPID_Processed` keyword, the `kPostLoad`/`kDataLoaded` cases from `main.cpp`, and the `kCreateIfMissing` enum plus the `find_or_create_keyword` lambda from `FormData.h`. Its fetch record now lists two fetches: the original 2026-06-02 parsing files, and a 2026-09-05 read of the four new files at commit `6e669085994d23c78f66d8ffb0b5e1812af1af6b` (`master`, committed 2026-09-02). `_CORPUS_STATUS.md` names that commit too, so a re-derivation at the next version bump knows exactly what was read and at which revision.

.claude/skills/spid-authoring/references/grammar-core.md
Avick3110: Folded via the same capture. The quote is verbatim — the body at `6e66908` is exactly `return !a_npc->IsPlayer() && !a_npc->IsDeleted();` — and that excerpt is now in `source-notes.md` alongside `distribute_on_load`. The notes also spell out the point you flagged: the player skip and the already-processed keyword guard are two separate, nested checks, not one expression, so the corpus quote cannot be read as covering both.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded, and it turns out the fact is article-stated rather than inferred. Article 6617's "General Distribution Info" says SPID "distributes all configured forms at runtime, it doesn't have effect on either loaded plugins nor on save" and "distributes everything from scratch each time you launch the game". `grammar-core.md` §3 now carries that as its own paragraph, with the source half beside it: `Setup()` creates the `SPID_Processed` marker keyword fresh each launch via `IFormFactory`, and `distribute_on_load` skips an NPC already carrying it — so nothing crosses sessions. Both excerpts are in `source-notes.md`. The SKILL.md bullet now points at `references/grammar-core.md` §3 the way the player bullet does, and drops "redistributes" for the article's own wording.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded as suggested — the bullet now says "look it up in the `papyrus-reference` skill" with no path, matching §4.4 and the eight existing cross-references. The remaining `references/form-types.md` in the same sentence is unambiguous now that it is the only `references/` in it.

.claude/skills/spid-authoring/references/grammar-core.md
Avick3110: Folded. `index.jsonl` gains a `when distribution happens` grammar row pointing at `grammar-core.md`, with aliases `lazy`, `cell load`, `once per launch`, `restart`, `from scratch`, `existing save`, `new game needed`, `player excluded`, `never the player`, `verify` — so a player-timing question no longer routes to the Trait filter. The `Keyword` form-type row gains `GetKeyword`, `no plugin FormID`, `not in xEdit` and `keyword array`. Both rows parse as JSON.

.claude/skills/spid-authoring/references/grammar-core.md
```diff
@@ -52,7 +52,23 @@ comment stripping are CSimpleIniA's, not SPID's own code.)
...
+(`SPID/src/main.cpp` — configs are read at `kPostLoad` and forms looked up at `kDataLoaded`.)
+
+**Every launch, from scratch.** SPID "distributes everything from scratch each time you launch the
```
Avick3110, line 59: **Article quotes are marked `[source]` against a .cpp file.**

The two quoted sentences here are verbatim article text, but the paragraph's only citation is a `[source]` marker pointing at `SPID/src/DistributeManager.cpp`. In this corpus `[source]` means "resolved from the MIT source, beyond article 6617" (`_CORPUS_STATUS.md` § Confidence).

"distributes everything from scratch each time you launch the game" and "doesn't have effect on either loaded plugins nor on save" are from `dev/references/SPID/_extracted/spid-complete-reference-6617.md:85` ("General Distribution Info") — `dev/references/SPID/source-notes.md:196-199` says so explicitly, quoting them as *the article's own statement*.

Scenario: the Provenance note tells the reader to re-derive on a SPID version bump. Whoever does it opens DistributeManager.cpp looking for those sentences, does not find them, and either drops a still-valid article fact or re-marks it unverifiable. Only the `SPID_Processed`/`Setup()` half of this paragraph is source-grounded — scope the marker to that and cite the article for the quotes.

.claude/skills/spid-authoring/references/grammar-core.md
```diff
@@ -52,7 +52,23 @@ comment stripping are CSimpleIniA's, not SPID's own code.)
...
+game" and "doesn't have effect on either loaded plugins nor on save" — nothing it distributes is
+written into the save, so a rule added to a `_DISTR.ini` after a save was made applies to that save's
+actors on the next launch, when they next load in. The "already handled this NPC" marker is a runtime
+`SPID_Processed` keyword created fresh at each launch, so it carries nothing across sessions.
+**[source]** (`SPID/src/DistributeManager.cpp` — `Setup()` creates the keyword via `IFormFactory`;
+`distribute_on_load` skips an NPC that already has it.) The practical limit is the once-per-launch
+config read above: no restart, no new rule.
+
+**Who — never the player.** Every distribution path is gated on
```
Avick3110: **"Every distribution path" asserts a universal the excerpt does not establish.**

The recorded evidence (`dev/references/SPID/source-notes.md:138-172`) shows exactly one caller of `should_process_NPC` — `distribute_on_load` — and its own grounding note claims only that: "`should_process_NPC` is literally `!a_npc->IsPlayer() && !a_npc->IsDeleted()`, and `distribute_on_load` calls it before any `Distribute()`." SPID's other entry points into `Distribute()` were not read.

Scenario: if any of them reaches `Distribute()` without the guard, a reader who trusts "never receives a SPID form" as source-verified stops looking at the player when debugging — and the corpus has shipped an unverified universal under a `[source]` marker, which is the guesswork the corpus exists to prevent. `_CORPUS_STATUS.md:57` repeats the same "every path" wording.

Either scope the claim to the on-load path that was read, or read the remaining call sites and keep the universal.

.claude/skills/spid-authoring/references/form-types.md
```diff
@@ -48,6 +48,12 @@ any plugin:
 Keyword = MyVerySpecialKeyword
 ```
 
+A keyword created this way is a **dynamic form: it has no owning plugin, so there is no plugin FormID
```
Avick3110, line 51: **The `[source]` marker vouches for an SKSE claim that was never read from SKSE.**

`FormData.h` proves SPID pushes the created keyword into `dataHandler->GetFormArray<BGSKeyword>()`. It says nothing about how SKSE's `Keyword.GetKeyword` resolves a name — yet that is the load-bearing half of the sentence the marker covers. `dev/references/SPID/source-notes.md:278-282` makes the same inference ("the same array a by-name keyword lookup searches") with no SKSE excerpt, and `papyrus-reference` bundles the signature only (`references/skse/Keyword.md:9`), not the implementation.

Scenario: if SKSE resolves by an editorID map, or by a cache built at data-load that a form pushed later is absent from, a user following the new SKILL.md bullet writes a verification script that returns None for a working distribution — exactly the false negative the section exists to prevent — while the skill states the mechanism as source-backed fact.

Either read SKSE's `papyrusKeyword` implementation and cite it alongside `FormData.h`, or state the SKSE half as inference rather than under `[source]`.

.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
```diff
@@ -40,7 +40,7 @@ the per-record files.
 | Distributable form types | **10 / 10** (`form-types.md`) |
 | Filter sections | **4 / 4** (`filters.md`) |
 | Flat enums | skill indices, trait letters, package-list types, form signatures, distribution order, defaults (`value-tables.md`) |
-| **Gaps** | **none in the article's scope.** Two article-*silent* details (comment syntax, whitespace tolerance) were resolved from source — see Confidence. |
+| **Gaps** | **The article does not cover verification behaviour.** Four details a user hits when checking whether a rule landed — player exclusion, the once-per-launch config read, the dynamic keyword's missing plugin FormID, and its reachability by name — are absent from article 6617 and were resolved from source (marked **[source]**; see Confidence). Two other article-*silent* details (comment syntax, whitespace tolerance) came from source the same way. Expect further gaps outside the article's grammar scope. |
```
Avick3110: **The ledger accounts for four `[source]` facts; the corpus now carries five.**

This row says four details "were resolved from source (marked **[source]**; see Confidence)". But `grammar-core.md:62-64` also marks the runtime `SPID_Processed` keyword — `Setup()` creating it via `IFormFactory`, `distribute_on_load` skipping an NPC that already has it — as `[source]`, and that claim has no entry in the § Confidence list this row points readers at.

Related: `dev/references/SPID/source-notes.md:160-172` quotes `SPID/src/DistributeManager.h` verbatim for the `processed` declaration, but § Source-repo cross-check does not name that file among those consulted.

Scenario: a reader auditing the corpus the way this file instructs — every `[source]` claim traced through Confidence — finds a marked claim with no provenance entry and cannot tell whether it was read or inferred.

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -120,6 +120,28 @@ SPID documentation.
...
+  launch and writes nothing into the save, so actors in a pre-existing save do get a rule added
+  afterwards (`references/grammar-core.md` §3). What bites is that the INIs are read **once per game
+  launch**: a rule added while the game is running needs a restart, and the actor has to load in that
+  new session for the lazy per-actor distribution to run.
+- **A dynamic keyword IS reachable.** A keyword SPID creates at runtime has no plugin FormID — nothing
+  in xEdit to point at — but it is in the game's keyword array, so `Keyword.GetKeyword("<name>")` finds
+  it by name, which is what makes a no-ESP keyword distribution verifiable from script at all. For the
+  signature, look it up in the `papyrus-reference` skill; the SPID side
+  is in `references/form-types.md`.
+- **Sample somewhere NPCs must be** — testing practice, not a SPID rule. An interior cell with known
```
Avick3110, line 150: **Chance (and a Level filter) is the missing false-negative cause.**

The section enumerates the ways a sample wrongly reads as failure — the player, a mid-session edit, no plugin FormID, an empty exterior — but never mentions the two that skip an actor by design.

Scenario: a user writes `Spell = MySpell|||||50`, follows this bullet to an interior cell with known occupants, samples two NPCs, finds the spell on neither, and concludes the rule is broken — when a 50% chance, re-rolled per session unless `!` is appended, makes that outcome ordinary. Same shape for a rule pushed into Leveled Distribution by a Level filter the sampled actors miss.

One line closes it: before concluding a rule failed, check Chance is 100 (or deterministic via `!`) and that the sampled actors pass the Level filter — both already in `grammar-core.md`.

.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
Avick3110: **(Anchored here because the lines are outside the diff.) The file still says SKILL.md and index.jsonl do not exist.**

Pre-existing, but this PR is the accuracy pass on this file and edits both artifacts it declares missing:

- § What this is: "The eventual `SKILL.md` (not yet authored — see Layering)"
- § Structure: "`index.jsonl` ← lookup routing — **NOT YET GENERATED**"
- § Layering: "**Layer 2 — not started.**"

Scenario: the file's opening line makes it the entry point for "a session picking this up." That session reads that no SKILL.md or router exists, and either re-authors what already ships or skips updating `index.jsonl` when it adds a corpus fact — the exact step this PR had to do by hand. Two lines of correction while the file is open.

.claude/skills/spid-authoring/references/grammar-core.md
Avick3110: Folded (02602846). The two quoted sentences now carry their own citation — "article 6617, § General Distribution Info, quoted verbatim" — and the **[source]** marker is scoped to the sentence it actually covers, the `SPID_Processed` / `Setup()` half. Also named `DistributeManager.h` there, since the `processed` declaration is what `source-notes.md` quotes for it.

Whoever re-derives on a version bump now opens the article for the quotes and DistributeManager.cpp for the marker, which is where each fact lives.

.claude/skills/spid-authoring/references/grammar-core.md
Avick3110: Folded (96ed0d21) — and you were right that the universal was false, not just unproven. I read the remaining call sites at `master` `6e66908`:

| File | Call site | Player guard |
|---|---|---|
| `DistributeManager.cpp:18` | `distribute_on_load` | `should_process_NPC` |
| `DistributePCLevelMult.cpp:22` | `HandleUpdatePlayerLevel` | none of its own — runs only `if (npc->HasKeyword(processed))` |
| `DistributePCLevelMult.cpp:135` | `LoadGame` | same `HasKeyword(processed)` condition |
| `DeathDistribution.cpp:303` | `ProcessEvent(TESDeathEvent)` | `!IsPlayerRef() && !IsSummoned()` |
| `DeathDistribution.cpp:214` | `ShouldBackgroundClone` hook | **none** |

`AddKeyword(processed)` appears exactly once, inside the `should_process_NPC` branch, so the two PC-level-mult hooks inherit the exclusion transitively. On-death uses a different check entirely, and its `ShouldBackgroundClone` entry has no player check at all — the player not being background-cloned is a game fact, not a SPID guard.

So `should_process_NPC` was never the shared gate. grammar-core §3 now states it per-path and says so, `_CORPUS_STATUS.md` matches, and `source-notes.md` carries the call-site table with the commit it was read at. The SKILL.md verification bullet lost its "every distribution path" wording too.

.claude/skills/spid-authoring/references/form-types.md
Avick3110: Folded by reading it (b3828252) — `ianpatt/skse64`, `skse64/PapyrusKeyword.cpp`, `master` `4cd2e34`. `GetKeyword` does walk `DataHandler::keywords`, so the claim holds and stays **[source]** — but it now cites SKSE for the SKSE half instead of stretching `FormData.h` across the seam, and form-types.md splits the sentence into the two facts with a marker each.

Your false-negative scenario was closer than the fix: it is not an editorID map, it is a cache — SKSE builds a `BSFixedString → BGSKeyword*` map from the array on the **first** `GetKeyword` call of the session and never invalidates it (the invalidation branch is commented out in SKSE's source). Anything pushed into the array after that first call is unfindable for the rest of the session. SPID creates its keywords during form lookup at `kDataLoaded`, before Papyrus runs, so a SPID keyword is in before the cache is built — that ordering is why the mechanism works, and it is now stated rather than assumed. form-types.md carries it as an explicit caveat, since it is the one way this lookup returns `None` for a keyword that exists.

.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
Avick3110: Folded (835f4f06). Gaps row now says five, and lists the runtime `SPID_Processed` marker among them; § Confidence gains its entry (`Setup()` creating it via `IFormFactory` each launch, declared in `DistributeManager.h`, `distribute_on_load` skipping an NPC that carries it), so every **[source]** claim in the corpus traces through Confidence.

§ Source-repo cross-check now names `DistributeManager.h`, plus `DistributePCLevelMult.cpp`, `DeathDistribution.cpp` and `Distribute.cpp` read for #3941109402, and carries a separate line for the one fact outside SPID's repo (SKSE `PapyrusKeyword.cpp`, with its own commit and read date) so a reader is not left inferring which repo a citation points at.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded (af085019) — new bullet in the verification section, grounded in grammar-core §10 and filters.md §3.

One correction to the wording you suggested: `!` does not rescue the sample. grammar-core §10 has it making the roll consistent for the same NPC + savegame across sessions — so a `50!` rule still misses half the NPCs, just repeatably. The bullet says a Chance below 100 is *supposed* to miss, that `!` makes re-sampling the same actor repeatable rather than making the form land, and that the way to settle it is to set Chance to 100. The Level half is as you had it: the line moves to the Leveled Distribution pass, and only auto-leveled NPCs inside the range receive it.

Also added `"only some NPCs got it"` / `"didn't land on everyone"` / `"random miss"` to the Chance line in `index.jsonl`, so the debugging phrasing routes there rather than only the word "chance".

.claude/skills/spid-authoring/references/_CORPUS_STATUS.md
Avick3110: Folded (2692c5ae). All three lines corrected:

- § What this is — "The eventual `SKILL.md` (not yet authored)" → "The skill's `SKILL.md` drives lookups against it, routed through `index.jsonl`".
- § Structure — the `index.jsonl` row drops **NOT YET GENERATED** and says what the file is, plus one line under the tree: add or update a topic's `index.jsonl` line when you add a fact, since a fact the router cannot reach is one the skill will not find. That is the by-hand step this PR had to do.
- § Layering — Layer 2 is done, not "not started": SKILL.md, index.jsonl and evals/eval_set.json all ship.

PR #577, opened 2026-09-05
Name the copy source's mod folder, and delete copy_npc_appearance

### LINKED ISSUE #387: [GAP] the asset source pole is singular, so a carry cannot follow an ordered from_source across two mod folders

Avick3110:
`housecarl_copy`'s readback now names the MO2 mod folder each source arm resolved from, beside the plugin filename it already printed — `Donor.esp (MO2 mod folder 'DisabledB')`, or `from the active load order` where the arm resolved through the order. That closes the flow gap #387 measured: with a look read through an override, the record comes from the first arm while the FaceGen sits beside the second, and nothing in the response said which mod folder held it, so the caller had to already know the answer before naming `source_provider=` on the following `housecarl_place`. The folder is carried as data on each source arm (read off the located path by the shared layer rule, not parsed back out of a sentence) and rendered wherever a source is named — the sources line, the from-record line, and the miss and fault refusals. The ordered-list half of the issue is waived by Aaron: `source_provider=` stays singular. In the same PR `housecarl_copy_npc_appearance` is deleted — the tool, its service and record halves, its CI guard and its differential harness — since the same job is `housecarl_copy` for the records plus `housecarl_place` for the FaceGen; the retired name redirects to that two-call shape, and the appearance skill and the codex map teach it. The asset-link harvest the closure copy still uses moves to its own file, and the alias census count drops by the one activation the deleted tool carried.

A new test builds the two-disabled-mod instance the issue describes (the `f1-i14` shape) and follows the readback rather than reading it: it takes the folder named for the `Donor.esp` arm, proves that folder holds the FaceGen on disk, places with it and compares the carried bytes, and shows the folder of the arm the record came from does not supply the file.

Closes #387


src/housecarl-mcp/WriteSentences.cs
```diff
@@ -209,16 +211,33 @@ internal static string Masters(IReadOnlyList<string> masters) =>
...
+    /// <summary>…and the other arm: the source resolved through the ACTIVE order, which is not one folder. Said
+    /// rather than left blank, so a missing folder name is never read as an omission.</summary>
+    [MustState("active load order")]
+    internal const string CopyArmActive = "from the active load order";
+
+    /// <summary>One source arm as a caller acts on it: their own spelling, and where it actually resolved.</summary>
+    internal static string CopyArm(SourceArmRef arm) =>
+        arm.Provider is { } folder
+            ? $"{arm.Spelling} ({CopyArmFolderLead}'{folder}')"
+            : $"{arm.Spelling} ({CopyArmActive})";
```
Avick3110: A null `Provider` is rendered as `from the active load order`, but null carries two different meanings and only one of them is that.

`LoadOrderService.cs:5508` fills `Provider` for a **File** arm from `LayerOfInstallPath`, which returns null whenever the located path is not under mods/overwrite/Data. That happens for real inputs:

- an absolute path in `from_source=` — the ambiguity refusal a few lines up literally instructs the caller to "Put the full path to the copy you mean in this element", and a backup/staging copy outside the instance resolves fine and lands here;
- a plugin sitting directly in the mods root (`slash > 0 ? rest[..slash] : null`);
- any path reaching the instance through a junction or symlink — `LocatePluginFileOnDisk` notes explicitly that such a path "will not string-match".

In all three the readback then says `Backup.esp (from the active load order)` about a plugin that is by construction **not** in the active load order — and it says it on the success line, the sources-consulted line, the miss refusal and the fault refusal alike. That is a silently wrong answer of exactly the kind the arm's own `Where` (`file 'X.esp' (direct path)`) already states correctly but never gets rendered.

The arm knows which case it is: `SourceArmKind.ActiveOrder` is the only kind entitled to `CopyArmActive`. A File arm with no resolved layer needs its own third phrasing (or fall back to `Where`) rather than borrowing the active-order sentence.

src/housecarl-mcp/WriteSentences.cs
```diff
@@ -209,16 +211,33 @@ internal static string Masters(IReadOnlyList<string> masters) =>
...
+
+    /// <summary>…and the other arm: the source resolved through the ACTIVE order, which is not one folder. Said
+    /// rather than left blank, so a missing folder name is never read as an omission.</summary>
+    [MustState("active load order")]
+    internal const string CopyArmActive = "from the active load order";
+
+    /// <summary>One source arm as a caller acts on it: their own spelling, and where it actually resolved.</summary>
+    internal static string CopyArm(SourceArmRef arm) =>
+        arm.Provider is { } folder
+            ? $"{arm.Spelling} ({CopyArmFolderLead}'{folder}')"
```
Avick3110: The folder is delimited with single quotes inside parentheses, but the codebase already has one delimiter for a provider name and it is double quotes: `AssetSourceSelection.Describe` renders `"<name>" (loose|BSA)`, and `AssetTools.cs:174` says why — "A mod folder can legitimately hold a parenthetical (\"SkyUI (SE)\"), so the delimiter is what tells a caller where the name ends" (#340).

This new sentence reintroduces the problem it solved, in the one place the name is meant to be copied verbatim into the next call:

- a folder with an apostrophe — `Bijin's NPCs`, `Immersive Wenches' Patch` — renders as `Donor.esp (MO2 mod folder 'Bijin's NPCs')` and the token has no unambiguous end;
- the parentheses are the delimiter and the folder may contain them: `SkyUI (SE)`.

The PR's own test shows the cost: `CopySourceFolderReadbackTests.FolderNamedFor` parses with `'([^']+)'`, which silently returns `Bijin` for the first case. And `npc-appearance-copy/SKILL.md` tells the caller, for the very next call, to "Name the mod exactly as the providers chain prints it, **inside the double quotes**" — so the workflow now hands the reader two different quoting conventions for the same name.

Rendering through the shared `AssetSourceSelection.Describe` formatter (or at minimum double quotes) keeps the printed token the token `source_provider=` accepts.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5503,8 +5503,11 @@ string Fail(string message)
             // later arm's version of a record this arm actually carries would be a silently wrong answer.
             var cache = ov.ToImmutableLinkCache();
             var where = $"file '{Path.GetFileName(loc.Path!)}' ({loc.Where}{(loc.WhyNotActive is { } why ? $"; NOT active — {why}" : "")})";
+            // The layer this file physically sits in, read off the path by the shared rule rather than parsed back
+            // out of `where`: it is the name a following asset placement passes as its provider.
+            var provider = LayerOfInstallPath(loc.Path!, modsDir, overwriteDir, dataDir);
```
Avick3110: `LayerOfInstallPath` answers with a LAYER, and two of its three answers are not mod folders: it returns `"overwrite"` for the overwrite layer and `"Data"` for the game Data folder. Both are reachable here — an unticked plugin in Data, or a plugin dropped in overwrite, takes the File arm — and the readback then prints `MyMod.esp (MO2 mod folder 'Data')`, naming a folder that is not an MO2 mod.

The token itself still works downstream (`AssetResolver.IsReservedProviderName` reserves exactly `overwrite` and `Data`, so `source_provider="Data"` resolves against the active universe), so this is the sentence being wrong rather than the value — but the sentence is the whole feature. The rename from `ShipperOfArchivePath` to `LayerOfInstallPath` says as much; the caller-facing phrase did not follow it.

src/housecarl-generator/WriteSurfaceGuardProbe.cs
```diff
@@ -2880,17 +2880,19 @@ static List<string> CopyOutcomeRenders()
...
                                IReadOnlyList<StripEntry>? attach = null, IReadOnlyList<WalkBoundary>? kept = null,
-                               IReadOnlyList<string>? assets = null, IReadOnlyList<string>? srcs = null,
+                               IReadOnlyList<string>? assets = null, IReadOnlyList<SourceArmRef>? srcs = null,
                                bool nothingBound = false) => new(
             true, null, null, null, strips.Count > 0 ? "clone" : "attach",
             new FormKey(src, 0x803), new FormKey(patch, 0x900), outPath, false,
             copied, kept ?? Array.Empty<WalkBoundary>(), Array.Empty<WalkCycle>(),
             attach ?? Array.Empty<StripEntry>(), strips, srcs ?? sources,
-            "CopySrc.esp", assets ?? Array.Empty<string>(),
+            sources[0], assets ?? Array.Empty<string>(),
```
Avick3110: `Make` now hardcodes the from-arm to `sources[0]` — the outer array — while `srcs:` is overridable, so the multi-source sample contradicts itself. The two-source call below passes `srcs: [("Override.esp","TheOverhaulMod"), ("CopySrc.esp", null)]`, and the rendered response then reads:

```
sources (in order, first hit wins): Override.esp (MO2 mod folder 'TheOverhaulMod') -> CopySrc.esp (from the active load order)
the source record was read from CopySrc.esp (MO2 mod folder 'TheDonorMod').
```

Same arm, two different origins, one response. This corpus is the pinned sentence-reach evidence, so a self-contradicting sample is what a future reader takes as the intended shape. Pass the from-arm alongside `srcs` (or default it to `srcs[0]`) so the two lines cannot disagree.

scripts/tool-names/sweep-tool-names.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+﻿#!/usr/bin/env python3
```
Avick3110: A UTF-8 BOM was added ahead of the shebang. The kernel matches the first two bytes literally, so `EF BB BF 23 21` is no longer a shebang and `./sweep-tool-names.py` fails with an exec-format error on Linux/macOS (`python3 sweep-tool-names.py` still works — Python opens source as utf-8-sig).

This looks unintentional rather than authored: nine files in this PR gained a BOM they did not have on main, all of them otherwise-untouched first lines —

```
scripts/tool-names/sweep-tool-names.py
src/housecarl-core/OffOrderAssetSource.cs
src/housecarl-generator/BulkPrimitivesWave3Probe.cs
src/housecarl-generator/ClosureWalkProbe.cs
src/housecarl-generator/Program.cs
src/housecarl-generator/WriteSurfaceGuardProbe.cs
src/housecarl-mcp-tests/RuntimeFormIdTests.cs
src/housecarl-mcp-tests/SourcePoleRemedyTests.cs
src/housecarl-mcp/LoadOrderService.cs
```

(plus the new `AssetLinkHarvest.cs`, whose source file `NpcAppearanceAssets.cs` had none). Worth stripping them so the diff is only the change.

src/housecarl-core/SourceChain.cs
```diff
@@ -52,12 +52,29 @@ public enum SourceArmKind
...
+/// <summary>One arm as a READBACK names it: the caller's own spelling and the mod folder behind it, with the fetch
+/// left off. Carried into an outcome so a response can name where each source resolved from after the chain and its
+/// overlays are gone.</summary>
+public sealed record SourceArmRef(string Spelling, string? Provider)
+{
+    /// <summary>The arm, minus its fetch.</summary>
+    public static SourceArmRef Of(SourceArm arm) => new(arm.Spelling, arm.Provider);
+
+    /// <summary>A source named with no folder behind it — what a caller-side test or a synthetic outcome spells.</summary>
+    public static SourceArmRef Named(string spelling) => new(spelling, null);
```
Avick3110: `SourceArmRef.Named` has no callers anywhere in the tree — the probe fixtures it was presumably added for construct `new SourceArmRef(spelling, null)` directly (`WriteSurfaceGuardProbe.cs:2917`, `:2922`, `:2928`). Either point those at it or drop it; a named constructor that nothing uses is a shape the next reader has to decide about.

src/housecarl-mcp/AliasTable.cs
```diff
@@ -247,6 +247,11 @@ static readonly (string Old, string Successor)[] RetiredTools =
...
          "absorbed into " + ToolNames.Place + ": assets= is unchanged in shape except asset_path= is path=, and source_provider=/kind= may now be given once for the whole set instead of per member. patch_name= is patch=; into= is unchanged."),
 
+        // The standalone NPC copy was one verb over two substrates, so its successor is two calls and the row says
+        // both — a row naming only the record half would leave the caller with a face and no FaceGen.
+        ("housecarl_copy_npc_appearance",
+         "split in two. The records are " + ToolNames.Copy + ": from= is source_formid=, seed_paths=[\"HeadParts\", \"HairColor\", \"HeadTexture\", \"WornArmor\"] walks the appearance, exclude_types=[\"Race:refuse\"] keeps the walk out of the race, and new_editorid=/target= are the same two destinations (source_plugin=+source_mod= is from_source=, an ORDERED list — name the override then the defining plugin). The FaceGen and textures are " + ToolNames.Place + ": the copy lists the asset paths and names, per source, the MO2 MOD FOLDER it read from — pass that folder as source_provider= and the files come out of it even when MO2 does not load that mod."),
```
Avick3110: **medium.** The row inverts the old-is-new direction every sibling row uses, in its own first clause.

Every other row reads OLD` = `NEW: `asset_path= is the member's path=`, `operations= is ops=`, `patch_name= is patch=`, `from_plugin= is source=`. This row opens with `from= is source_formid=` — `from=` is the *new* `housecarl_copy` parameter and `source_formid=` was the retired one, so the mapping reads backwards. A caller arriving with `source_formid=` habits (which is the whole audience for a retired-name row) is told to rename the parameter it already has into the one it does not.

The same sentence gets it right two clauses later — `source_plugin=+source_mod= is from_source=` is old-is-new — so the row disagrees with itself as well as with its siblings. Suggest `source_formid= is from=`.

Avick3110: **medium.** `new_name=` is the one deleted parameter with no successor named anywhere, and the deletion is only correct if nothing is lost.

The retired tool took `new_name` ("the clone's display Name; defaults to the donor's"). `housecarl_copy` has no name parameter, so a clone minted through the two-call shape silently carries the donor's display name — the wrong answer for the standalone follower this skill's headline job is, and one the modder will not notice until they see the donor's name in-game.

It is recoverable — `housecarl_apply` with an op on `Name`, which the flow already calls in step 2 for the tint bundle — but nothing says so. This row names the migration for every other parameter it moves, and the skill's clone flow never mentions naming the clone at all. Either the row gains the clause (`new_name= is an op on Name in the same housecarl_apply call`) or the skill's step 2 does; leaving it in neither is the silent half of the loss the two-call shape was supposed to have none of.

(`patch_name= is patch=` and `target_formid= is target=` are also unstated here — the row says `new_editorid=/target= are the same two destinations`, which glosses the rename rather than teaching it. Both are cosmetic beside `new_name=`.)

src/housecarl-mcp/WriteSentences.cs
Avick3110: Folded. `SourceArmRef` now carries the arm's `Kind`, so the render answers from what the arm actually was rather than from a null: an `ActiveOrder` arm gets "from the active load order", a `File` arm with a layer gets that layer, and a `File` arm with none gets its own sentence — "read straight off disk, outside your mods, overwrite and Data folders — no MO2 mod folder to pass as source_provider= for its files". `SourceArmRef.Named` went with it (your last finding); every construction now states the kind. The probe sample carries all five shapes so each sentence reaches a render.

src/housecarl-mcp/WriteSentences.cs
Avick3110: Folded — double quotes, the #340 delimiter: `Donor.esp (MO2 mod folder "TheDonorMod")`. The test regex follows, and its fixture folder is now `Bijin's NPCs (SE)` so the apostrophe and the parenthesis are both in the asserted path rather than in a comment. The skill's example and its "inside the double quotes" instruction now agree, and the changelog says the delimiter.

src/housecarl-mcp/AliasTable.cs
Avick3110: Folded — the row now opens `source_formid= is from=`, old-is-new like every sibling.

Avick3110: Folded in both places. The row now says `new_name=` has no successor parameter: a clone carries the DONOR's display name, and renaming it is an op on `Name` in the same `housecarl_apply` call that copies the tint bundle — and it also picks up `patch_name= is patch=`. The skill's Step 2 gains a "Name the clone here" block with the op spelled out, and the changelog bullet says the same. `target=` was already named in the row as a destination, so I left that clause alone.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The render is layer-aware now: a real mod folder keeps "MO2 mod folder", the overwrite layer prints "MO2's overwrite folder, \"overwrite\"" and the game Data folder "the game's Data folder, \"Data\"" — the quoted token stays exactly what `source_provider=` takes, only the phrase stops calling it a mod.

src/housecarl-generator/WriteSurfaceGuardProbe.cs
Avick3110: Folded — `Make` takes its from-arm from `(srcs ?? sources)[0]`, so the two lines cannot disagree. The multi-source sample is now one shared array covering all five arm shapes (mod folder, active order, overwrite, Data, a file outside them all).

scripts/tool-names/sweep-tool-names.py
Avick3110: Folded — stripped from all ten (the nine you listed plus the new `AssetLinkHarvest.cs`). `F1MeasureProbe.cs` and `NativePairingProbe.cs` also carry BOMs but had them on `main`, so they are untouched; the diff's first lines are back to just the change.

src/housecarl-core/SourceChain.cs
Avick3110: Folded — `Named` is deleted. `SourceArmRef` is now `(Spelling, Kind, Provider)` and every construction, probe fixtures included, states the kind, which is what the first finding needed anyway.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: **A named ACTIVE plugin arm still names no folder, which is the enabled-donor case.**

Only this off-order FILE arm gets a `Provider`. The two ACTIVE arms above (`arms.Add(new SourceArm(spelling, SourceArmKind.ActiveOrder, …))` at ~5447, and the named-by-path one at ~5456) are left at `Provider = null`, so `CopyArm` renders them as `from the active load order` and the caller is back where #387 found them.

The `SourceArm` doc justifies null as "no single folder stands behind it". That is true of the bare `winner` pole — it is the whole order — but it is not true of a named active plugin, which sits in exactly one mod folder, nor of the path-named arm, where the caller literally typed the folder in.

Concrete: donor mod ENABLED, its FaceGen path contested by a face replacer sorted above it.

```
housecarl_copy(from="000805:Donor.esp", from_source=["Donor.esp"], seed_paths=[…], new_editorid="X")
→ sources (in order, first hit wins): Donor.esp (from the active load order)
```

Step 3 of the skill then says "Name the donor with `source_provider=`, do not take the VFS winner" — but nothing named a folder, so the call omits it, resolution takes the winner, and the replacer's mesh and tint land on top of the head-part and tint records copied from the donor. That is the row already in the skill's own mistakes table ("Letting the FaceGen source default to the VFS winner").

The folder is derivable with the same one-liner used here:

```csharp
LayerOfInstallPath(view.PluginPath(active)!, modsDir, overwriteDir, dataDir)
```

(`PluginPath` returns null for a plugin not in the order, which cannot happen on this branch, and the roots would need deriving before the `comp is null` block — but the rule is the shared one either way.)

src/housecarl-mcp/CopyTools.cs
```diff
@@ -49,7 +49,9 @@ public static class CopyTools
          "resolves whether that plugin is active OR sitting on disk in a DISABLED mod. Naming several is how you " +
          "read a look from an override patch while its records come from the defining plugin: " +
          "from_source=['Override.esp','Donor.esp']. This is FALLBACK, never a merge — a record present in several " +
-         "sources comes from the FIRST, and the result names which source produced each record. A record no source " +
+         "sources comes from the FIRST, and the result names which source produced each record — and, for each " +
+         "source, the MO2 MOD FOLDER it was read from, which is the name to pass as the provider when you place " +
```
Avick3110: **The description promises a folder for *each* source; three of the four arm shapes never have one.**

`CopyArm` names a mod folder only for a FILE arm that resolved under `mods\`. The other three say something else: the default `winner` pole and a named active plugin both render `(from the active load order)`, and a file outside mods/overwrite/Data renders `read straight off disk … no MO2 mod folder`.

So a caller on the DEFAULT (`from_source` omitted → `["winner"]`) reads this sentence, expects a folder, and gets none. The two outcomes are the ones the skill warns about: invent a folder name for `source_provider=` ("a guess wearing a path"), or omit it and take the VFS winner.

The caveats do exist — but only in `npc-appearance-copy/SKILL.md`, three paragraphs deep. The tool description is where the model meets the claim first, and it asserts unconditionally. Same unconditional wording in `plugin/codex/housecarl/SKILL.md` ("The readback names, per source, the MO2 mod folder it resolved from") and in the `housecarl_copy_npc_appearance` redirect hint in `AliasTable.cs`.

A clause is enough: *…and, for a source read out of a mod folder, the MO2 mod folder it was read from — the name to pass as the provider when you place that record's files.*

src/housecarl-mcp/WriteSentences.cs
```diff
@@ -209,16 +211,62 @@ internal static string Masters(IReadOnlyList<string> masters) =>
...
+    /// apostrophe (<c>JK's Skyrim</c>) or parentheses (<c>SkyUI (SE)</c>), and this token is copied verbatim into
+    /// the next call's <c>source_provider=</c>, so the boundary has to be a character the name cannot contain.</para></summary>
+    internal static string CopyArm(SourceArmRef arm) => arm.Kind == SourceArmKind.ActiveOrder
+        ? $"{arm.Spelling} ({CopyArmActive})"
+        : arm.Provider is { } layer
+            ? $"{arm.Spelling} ({LayerPhrase(layer)}\"{layer}\")"
+            : $"{arm.Spelling} ({CopyArmNoFolder})";
+
+    /// <summary>What to call the layer a file arm resolved into — a mod folder, or one of the two that are not.</summary>
+    static string LayerPhrase(string layer) =>
```
Avick3110: **A mod folder named `Data` or `overwrite` is described as the layer it is not — and the name handed back is reserved.**

`LayerPhrase` decides which of the three sentences to use by comparing the folder NAME, but `LayerOfInstallPath` returns that same string for a directory under `mods\` as it does for the two reserved layers. It never says which branch produced it.

A donor plugin at `mods\Data\Donor.esp`:

- `LayerOfInstallPath` takes the `Under(archivePath, modsDir, …)` branch, `rest = "Data\Donor.esp"`, returns `"Data"`.
- The readback prints `Donor.esp (the game's Data folder, "Data")` — a false claim about where that plugin is.
- The caller passes `source_provider="Data"` to `housecarl_place`. `AssetResolver.IsReservedProviderName("Data")` is true, so `OffOrderAssetSource` returns `ReservedName` and the folder scan never reaches `mods\Data`. Resolution falls to the game Data root instead: a false "does not supply", or — if the game Data folder holds a file at the same relative path — the wrong bytes placed silently.

`AssetResolver.IsReservedProviderName`'s own doc names this collision as the reason reserved names exist ("a mod folder literally called `Data` would then be served for the name `Data` and shadow the layer that name means"). The new readback creates a path that walks the caller into it, and does so while asserting the wrong layer.

The fix belongs at the source rather than here: have `LayerOfInstallPath` return which branch answered (a small enum or a `(Kind, Name)` pair) so the sentence keys on the branch, not on the string. Failing that, a `mods\` folder whose name is reserved should be reported as having no folder to pass rather than as the layer.

src/housecarl-mcp/CopyTools.cs
```diff
@@ -239,11 +241,13 @@ internal static string Render(ClosureCopyOutcome o)
...
     }
 
-    static string RenderWalkRefusal(WalkRefusal w, IReadOnlyList<string> sources) => w.Kind switch
+    static string RenderWalkRefusal(WalkRefusal w, IReadOnlyList<SourceArmRef> sources) => w.Kind switch
     {
         WalkRefusalKind.SourceMiss => WriteSentences.CopySourceMiss($"{w.Key} (reached via {w.PulledBy})", sources),
         WalkRefusalKind.SourceFault => WriteSentences.CopySourceFault(
-            $"{w.Key}", w.Fault?.Arm.Spelling ?? "a source", w.Detail),
+            $"{w.Key}",
+            w.Fault is { } f ? WriteSentences.CopyArm(SourceArmRef.Of(f.Arm)) : "a source",
```
Avick3110: **The fault refusal single-quotes the whole composed phrase, so it reads as the source's name.**

`CopySourceFault` renders its `source` argument as `'{source}'`, and it is now handed the full `CopyArm(...)` phrase instead of the bare spelling. The rendered refusal:

```
error: 'Donor.esp (MO2 mod folder "TheDonorMod")' carries 000805:Donor.esp but it could not be read — …
… Repair or replace that plugin, or name a different one in from_source=.
```

The quoted token is not a value `from_source=` accepts, so a caller acting on that remedy has to strip the bracket by hand. It also works against the #340 delimiter argument this PR leans on — the folder name is supposed to end at the double quote, and here the whole thing is wrapped in another quote pair.

Passing `f.Arm.Spelling` and appending the folder outside the quotes keeps both facts:

```
'Donor.esp' (MO2 mod folder "TheDonorMod") carries 000805:Donor.esp but it could not be read — …
```

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real, and folded in fbf23e4f.

A named ACTIVE arm now derives its layer the same way the file arm does. `ActiveLayer(name)` in `BuildSourceChain` takes `view.PluginPath(name)` through the shared rule and hands the result to both active arms — the named one and the named-by-path one. It is best effort: roots that will not derive cost the caller the folder name, never the source.

The bare `winner` pole keeps no folder, because there it really is the whole order.

`CopyArmWhere` now says both facts rather than choosing:

```
sources (in order, first hit wins): Donor.esp (from the active load order, MO2 mod folder "TheDonorMod")
```

Pinned end to end by `CopyEnabledDonorFolderTests`, on the same fixture world with `enableDefining: true`: the active arm names the folder; the folder it names is proven to be the one that really holds that plugin's FaceGen; and `winner` still names none (`DoesNotContain("winner (from the active load order,")`).

src/housecarl-mcp/CopyTools.cs
Avick3110: Real, and folded in 396f9994.

The description now promises the folder only where one exists, and says the rest say so outright:

> …the result names which source produced each record — and, for a source that resolved into an MO2 MOD FOLDER (active or disabled), that folder, which is the name to pass as the provider when you place that record's files (its FaceGen, say) with `housecarl_place`. A source with no such folder — the 'winner' pole, MO2's overwrite, the game's Data folder, a file outside all of them — says so instead, so you never have to invent a folder name.

"(active or disabled)" is now true of the readback as well: fbf23e4f gives an enabled named plugin its folder too, so the set with no folder is the one this sentence lists rather than three of the four shapes.

The same two echoes are corrected: `plugin/codex/housecarl/SKILL.md` and the `housecarl_copy_npc_appearance` redirect hint in `AliasTable.cs`. The `npc-appearance-copy` skill's own paragraph is updated in fbf23e4f for the enabled-donor and reserved-name cases.

src/housecarl-mcp/WriteSentences.cs
Avick3110: Real, both halves, and folded at the source in fbf23e4f.

The branch now travels with the name. `InstallLayerOfPath` returns `SourceLayer(SourceLayerKind Kind, string Name)` — ModFolder / Overwrite / GameData — and `LayerOfInstallPath` is a one-line wrapper over it for the callers that only want the name. `SourceArm`/`SourceArmRef` carry `SourceLayer? Layer` in place of `string? Provider`, so the sentence keys on the branch and can no longer call a mod folder the layer it merely spells like.

The second half — the name handed back is one the placement will not answer — is said rather than left implied. A ModFolder whose name is reserved renders:

```
Collide.esp (MO2 mod folder "Data" — but that name is RESERVED for MO2's overwrite layer and the game's Data
folder, so it cannot be passed as source_provider=: a placement given it reads the layer, not this mod. Rename
that mod folder in MO2)
```

The reserved test itself is stated once: `AssetResolver.IsReservedLayerName` is the two layer names, `IsReservedProviderName` now spells them through the same two constants, so the test that decides what a placement refuses and the test that decides what a readback promises cannot drift.

`CopyReservedFolderNameTests` builds `mods\Data\Collide.esp` for real and pins both: the readback does not say "the game's Data folder" for that arm and does carry the withdrawal; and — with a copy of the FaceGen placed inside `mods\Data` so a folder scan would succeed — `housecarl_place` handed `source_provider="Data"` still cannot reach it. The folder is never scanned, which is the claim the sentence makes.

src/housecarl-mcp/CopyTools.cs
Avick3110: Real, and folded in a6f24853.

`CopySourceFault` now takes the arm rather than a rendered phrase, and quotes only the spelling:

```csharp
internal static string CopySourceFault(string what, SourceArmRef? arm, string cause) =>
    $"'{arm?.Spelling ?? "a source"}'{(arm is { } a ? $" ({CopyArmWhere(a)})" : "")} carries {what}"
    + CopySourceFaultLead + cause + CopySourceFaultRemedy;
```

which renders the shape you wrote:

```
error: 'Donor.esp' (MO2 mod folder "TheDonorMod") carries 000805:Donor.esp but it could not be read — …
```

`CopyArm` is now `$"{Spelling} ({CopyArmWhere(arm)})"` over a shared `CopyArmWhere`, so the sources line and this refusal state where an arm resolved from one place — a sentence that quotes the spelling itself appends the location outside its own quotes instead of swallowing the whole phrase. The write-surface-guard's fault sample carries a mod folder now, so the composed shape is the one the sentence-reach net renders.

PR #579, opened 2026-09-05
Remake the skill-authoring standard from the current official docs

### LINKED ISSUE #402: [Docs] HOUSECARL_SKILL_AUTHORING.md still carries legacy scaffolding

`standards/HOUSECARL_SKILL_AUTHORING.md` (444 lines) still reads as the predecessor project's document: `Claude_MO2` framing ×11 (:3, :11, :13, :160, :290, :293, :299, :375, :377, :383, :385, :433), old-repo references ×17 (:79, :171, :208, :249, :255, :261, :316, :412, :434–:440), `daemon-foundation`/`daemon-side`/daemon nouns ×4 (:65, :125, :270, :293), and 9 lines of `Phase N` migration language.

**Triage nuance (verified):** a subset of hits is intentional — :123/:125 cite phase tags and daemon nouns *as forbidden examples* in an anti-pattern table, and :208 explicitly scopes old-repo refs as acceptable in standards docs. The genuinely stale-shaped set is the forward-looking Phase 2/2.7/3.2/3.6 planning references (:269, :270, :277, :279, :364, :365), the `<old-repo>` citation anchors, and the pervasive `Claude_MO2` framing. A ground-up native rewrite was deferred as Aaron's call (recorded 2026-07-15).

**Routing:** row 9 rider — W6 is the skill-authoring wave; the methodology doc it authors against gets its rewrite there. Milestoned.

Avick3110:
The skill-authoring standard is rewritten from a blank page against the current Anthropic and OpenAI documentation. The old file was removed first and not read by anyone in this phase, so nothing carried over by habit. houseCARL ships one bundle to two hosts, so host difference is a first-class axis: every rule says whether it holds for Claude Code, for Codex, or for both, and where the hosts differ it says what the bundle does so both are satisfied. Every rule cites the page it rests on; a rule that rests on no page is labelled a house choice with its reason, and all 68 of them are collected in one table at the end for you to rule on, alongside a list of every URL cited. The order runs: what a skill is for, layout and packaging, frontmatter, the description and loading, the body and style, testing. Two things moved from the section they were drafted in: "one skill, one job" sits in "what a skill is for" rather than in the body section, because it answers the scope question a reader has before writing anything; and the precedence rule was merged into one rule in the body section instead of being stated in two places.

Closes #402

## House choices for Aaron

| # | Rule | Choice | Reason |
|---|---|---|---|
| 1 | 1 | Pre-decision knowledge does not go in a skill | The body is read after the decision is made |
| 2 | 4 | The reference-versus-procedure split is used on Codex too | Codex is silent and nothing there argues against it |
| 3 | 6 | houseCARL skills carry no hooks | One tree, two hosts; a skill-carried hook exists on one host only |
| 4 | 7 | A skill stays a procedure; isolation is said in the procedure | Skill preloading into a subagent is Claude Code only |
| 5 | §1 table, row 3 | That houseCARL exists lives in standing context, not a skill | Rule 1; no vendor page addresses a skill bundle over an MCP server |
| 6 | §1 silences | `CLAUDE.md` and `AGENTS.md` say the same facts rather than dividing them | Nobody documents what happens when both are present |
| 7 | §1 silences | A skill stands alone and never assumes another's body is loaded | Only Claude Code documents chaining |
| 8 | 11 | Folder name equals frontmatter `name`, kebab-case | One string, so disk and invocation cannot drift |
| 9 | 12 | No version number in a folder name, and no rename on rewrite | The name is what a user types; a rename needs a marketplace `renames` entry |
| 10 | 13 | `references/`, `scripts/`, `assets/` on both hosts | Codex and the spec name them; Claude Code has no competing convention |
| 11 | 14 | Eval set at `evals/eval_set.json` in the skill folder | Tests travel with what they test; the documented name is `evals/evals.json` and nothing but us reads the file |
| 12 | 20 | Every skill name is distinctive without a prefix | Codex does not resolve collisions |
| 13 | 22 | An umbrella skill carries the bundle's identity on Codex | Codex gives locally installed skills no namespace |
| 14 | 23 | Install a copy under `~/.agents/skills/` rather than expect Codex to read the Claude tree | Codex is silent on `.claude/skills` |
| 15 | 28 | No root-level `SKILL.md`, no flat `commands/` | One layout both hosts accept beats two and a conversion step |
| 16 | 30 | Default `skills/` layout, so the manifest `skills` field is not load-bearing | The two hosts' rules for the field do not reconcile |
| 17 | 31 | Real directories, never symlinks | A layout that only works before packaging fails at packaging |
| 18 | 32 | Two packages generated from the one `.claude/skills/` source | Codex excludes MCP wiring from a skills-only upload |
| 19 | 34 | Ship a Codex package rather than rely on conversion or import | Both routes are stated to change what they carry |
| 20 | §2 silences | Follow `$HOME/.agents/skills` over `~/.codex/skills` | It is the path in the discovery page's scope table and the cross-client convention |
| 21 | §2 silences | `.claude-plugin/marketplace.json` is a Claude Code file only | The guide governing our direction says not to rely on it |
| 22 | 38 | User-visible changes go in `plugin/CHANGELOG.md` under `## Unreleased` | The file is for people; no page states a location or format |
| 23 | 40 | `SKILL.md` uppercase, no BOM | The uppercase spelling is the one every source accepts |
| 24 | 43 | Always write `name`, equal to the folder name | Required on two surfaces, free on the third |
| 25 | 44 | ASCII kebab-case, no reserved words, no angle brackets in `name` | The tighter rule is the one every surface accepts, and it costs nothing |
| 26 | 45 | `name` at most 54 characters | `housecarl:` spends 10 of Codex's 64-character identity cap |
| 27 | 48 | No angle brackets in `description` | The listing wraps and escapes the value; Anthropic's own validator bans them |
| 28 | 49 | The description cap is 1,024, not the Help Center's 200 | 1,024 is stated by every surface the bundle reaches; 200 governs the claude.ai upload path |
| 29 | 50 | Frontmatter is `name`, `description`, `license` and nothing else | One file valid everywhere; no field whose behaviour depends on the host |
| 30 | 51 | No `allowed-tools` | It grants permissions on one host and does nothing on the other |
| 31 | 52 | No `compatibility`, no `metadata`; declare the server where a host acts | A `compatibility` string is text nothing reads |
| 32 | 53 | Write `license`, short, same on every skill | Legal everywhere, and it travels with a skill copied out of the bundle |
| 33 | 55 | `interface.short_description` is human text, not the trigger | One line is read by a person, the other by a model |
| 34 | 57 | Booleans spelled `true` / `false` | The strict spelling is the one both hosts read |
| 35 | §3 silences | Write only fields inside the allowed six rather than answer the unknown-key question | Nobody states what a host does with an unknown key |
| 36 | §3 silences | The kebab identifier, not a human display label | It is also the folder name and the command segment |
| 37 | §3 silences | Write to the specification where OpenAI defers but does not enforce | Nothing valid under the specification is invalid under Codex's checks |
| 38 | §3 silences | Do not use `display-name`, `default-enabled`, `fallback`, `dependencies`, `SKILL.json` | Undocumented behaviour, and outside the allowed six |
| 39 | 65 | Every description fits in 400 characters | Derived from Codex's 8,000-character worst case; survives on the tightest host |
| 40 | 67 | A boundary clause where a skill's subject borders another's | The cheapest fix for a wrong load; costs a dozen words |
| 41 | 70 | No declared arguments; a skill works when invoked with none | The placeholder grammar exists on one host only |
| 42 | 71 | No skill depends on being stacked | Stacking is Claude Code only |
| 43 | 72 | Neither invocation switch is set on either host | No skill performs the side effect itself, and hiding one defeats triggering |
| 44 | 74 | No skill sets `context: fork` | These skills carry guidance as well as steps, and Codex has no counterpart |
| 45 | 76 | No description or body claims system-level authority | On one host the text sits at user-message priority |
| 46 | §4 silences | Assume shortening cuts from the end and a skill can vanish | The cheap assumption; the opposite is unrecoverable at runtime |
| 47 | §4 silences | Author against Codex's 8,000 characters, not its token figure | The smaller and more concrete of two unreconciled units |
| 48 | §4 silences | Write `policy.allow_implicit_invocation` if it is ever set | It is the spelling the validator publishes |
| 49 | §4 silences | Follow the enterprise page on how many skills is too many | It is the one that names a failure mode |
| 50 | 77 | Body under 500 lines and 5,000 tokens on both hosts | The only published number; no Codex rule competes |
| 51 | 82 | Adopt Codex's six-item body list on both hosts | The only required-content list either vendor publishes |
| 52 | 87 | Table of contents over 100 lines, not 300 | The stated reason, partial reads, applies at 100 |
| 53 | 88 | Game-generic and Skyrim-specific knowledge split structurally | Reuse: a future game target keeps the generic half |
| 54 | 89 | Domain knowledge as tables and lists, not prose | A lookup should not require reading a paragraph |
| 55 | 98 | An error or refusal is one plain sentence | Matches the houseCARL cornerstone; the message shapes the next attempt |
| 56 | 102 | A reason clause only where it changes what the model would do | Splits the two vendors' opposite advice along their own fragility line |
| 57 | 103 | No all-caps, no MUST/ALWAYS/NEVER | Two current-generation pages and skill-creator against one platform page |
| 58 | 112 | No precedence boilerplate; write so the question does not arise | Fourteen copies of a line that matters only when a skill is already wrong |
| 59 | 113 | Provider-neutral body, no Claude-only body syntax, written that way in the source | Text that silently means nothing is worse than text that fails |
| 60 | 114 | Bare `housecarl_<snake_case>` plus the server named once, not `ServerName:tool_name` | The prefix is host-specific; `standards/NAMING.md` fixes the bare name |
| 61 | 117 | Hold the bundle to the union of the three structural checks | One tree, two hosts; only the strictest clears all of them |
| 62 | 119 | Codex's five categories, Anthropic's numbers | Categories are a shape; Codex publishes no measurement |
| 63 | 125 | A blind comparison decides a version change | Only the comparator is written to be read blind |
| 64 | 126 | Two or three outcome cases to start; twenty trigger queries | The only trigger-set size published with a scoring method |
| 65 | 127 | Test on the models the bundle runs under | A trigger rate on a model nobody uses is not the number that matters |
| 66 | 128 | Anthropic's method, re-run once against Codex's five categories | Codex publishes a procedure but no measurement |
| 67 | §6 silences | Name no eval command as a method | A method the standard cannot cite is one a reader cannot run |
| 68 | §6 silences | Treat the comparison harness as a plugin, not a host feature | Codex has no harness, so the method must be runnable by hand |

## Conflicts resolved

- **Description length.** The field's hard cap is 1,024 (rule 47), the working budget is 400 characters (rule 65), and the claude.ai Help Center's 200 applies only to that upload path (rule 49). The 400 sits inside the 1,024, so the two are stated once each and cross-referenced rather than restated.
- **Tool naming.** The Anthropic platform says to write `ServerName:tool_name`; `standards/NAMING.md` fixes the project spelling as `housecarl_<snake_case>`. The standard follows NAMING.md and names the server once in prose (rule 114), keeping the platform citation for the other side.
- **The install facts.** The brief's addendum corrects what the first draft inferred: the Claude install copies the whole bundled plugin folder to `~/.claude/skills/housecarl/` and carries a `.claude-plugin/marketplace.json`, and Codex gets the fourteen skills flat under `~/.agents/skills/` with the umbrella beside them and the server in a neutral per-user directory. Rule 21 states those facts and drops the `@skills-dir` inference.
- **Eval filename.** The bundle ships `evals/eval_set.json`; the documented name is `evals/evals.json`. Rule 14 keeps the shipped name as a house choice and records the documented one; rule 123 describes the content.
- **Duplication across sections.** The supporting-directory names, the 54-character name arithmetic, the `agents/openai.yaml` location, the compaction behaviour, the provider-neutral body rule and the precedence rule were each stated in two or three sections; each is now stated once, in the section that owns it, with the citations from both sides kept.

## Left unanswered by the sources

- Whether the Codex portal's conversion of a Claude manifest fixes a `skills` value shaped for Claude Code.
- The exact form of a namespaced invocation for a plugin-bundled skill on Codex, beyond the `$`/`@` sigils and the 64-character identity cap.
- Whether Codex reads a per-skill `agents/openai.yaml` for a locally installed skill under `~/.agents/skills/` the same way it does for a submitted plugin. Every shape rule and error code comes from the submission page.
- `SKILL.json` on the Codex app-server path: no published schema, and no statement of how it relates to `agents/openai.yaml` when both exist.
- What Claude Code does with an unknown key in `SKILL.md` frontmatter.
- Whether Claude Code's 1%-of-context listing budget is counted in characters or tokens, so no worst-case number can be computed for that host.
- Whether the file paths Codex shows in the listing are absolute or relative, so the 80–100 character per-entry overhead behind the 400-character budget is an estimate.
- What `interface.default_prompt` does at invocation time.
- How a description is shortened on either host, and which skills Codex omits first when the list overflows.
- Whether Codex re-reads a skill body on later turns.
- Whether Codex resolves relative supporting-file paths for the model.
- Any threshold at which a body should be split by domain, so the generic/Skyrim split is stated as an always.
- Whether the bundle's `evals/eval_set.json` currently holds trigger queries or outcome expectations. Nobody in this phase could read the skill tree; if it holds trigger queries, the outcome set described in rule 123 does not exist yet.
- Anything about a skill bundle sitting on top of an MCP server, which is houseCARL's actual shape.

## Uncertain

Each drafter named what they were least sure of. These are the rules most worth a second look.

- **Layout and packaging.** The Claude install mechanism. The draft inferred a `@skills-dir` plugin from the `/housecarl:<name>` invocation; the brief's addendum supplied the checked facts and rule 21 now states those instead, but nobody in this phase read the installer or `build-plugin.ps1`. The same drafter called the `evals/` location (rule 14) the section's weakest-sourced rule.
- **Frontmatter.** Whether `license` should be written at all (rule 53). It is inert on both hosts; the whole justification is that it is legal everywhere and travels with a copied skill. Dropping it would make the frontmatter exactly two fields.
- **The description and loading.** The 400-character budget (rule 65). It is arithmetic over one stated budget plus two estimates — the per-entry path overhead, and the assumption that other people's skills share the catalogue — and it is the one number here a reader will treat as authoritative.
- **The body and style.** The tool-name rule (rule 114) writes a house choice against a stated Anthropic rule; if either host actually needs a server prefix for the houseCARL server, that fact should replace the reasoning, because a rule that produces "tool not found" is worse than a little duplication. Rule 103 (no all-caps) is the strongest position in the standard and overrides a live Anthropic page.
- **What a skill is for.** Rule 112 declines a thing OpenAI explicitly recommends — stating in the skill that user instructions take precedence — on the grounds that fourteen copies of that sentence cost more than they buy.


standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
-### 3.4 — What the description must NOT contain
-
-These items are forbidden in any Housecarl skill description and a reviewer rejects on first sight:
-
-| Forbidden item | Example (don't ship this) | Why |
+# Skill authoring
+
+houseCARL ships one set of skill sources to two hosts: Claude Code and OpenAI Codex. The source of
+truth is `.claude/skills/<name>/SKILL.md` in this repository; the build produces one package per
+host from it. Host difference is a first-class axis here: every rule says which host it holds for,
```
Avick3110, line 5: The opening promises that every rule says which host it holds for, and a large share of the rules never name one. Rules 11, 12, 14, 20, 22, 30, 38, 43, 45, 49, 50, 65, 82, 87, 88, 89, 98, 102, 103, 112, 113, 114, 119, 123 and 125 to 128 open with the bold claim and a House choice label and no host clause at all. Most of them read as both hosts, but the author has to infer it — which is the thing the contract promises they will not have to do. Fix: add the host clause to those rules, or narrow the promise to "every rule that turns on a host difference says which host".

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+    19. Codex's only stated uniqueness guarantee is inside one plugin — "`skill_identity_duplicate` -
+    Each skill `name` must be unique within the plugin" — and it documents `plugin-name:skill-name`
+    only as a 64-character identity cap, stating no rule that the prefix prevents a collision with a
+    local skill (https://developers.openai.com/plugins/deploy/submission-errors.md). Where two skills
+    of the same name both load, the user tells them apart by path: "In Codex, the initial list also
+    includes each skill's file path" (https://learn.chatgpt.com/docs/build-skills). So a skill name
+    says what the skill does specifically enough to survive without a prefix: `facegen-diagnostics`,
+    not `diagnostics`.
+
+21. **The bundle lands at `~/.claude/skills/housecarl/` on Claude Code and under `~/.agents/skills/`
```
Avick3110, line 262: Rule 21 states the two install layouts as bare fact: no citation, no House choice label, and no row in the House choices table at the end. It is the only rule in the file of that kind, and it carries load — where the bundle lands, that the installer copies the whole plugin folder, that a marketplace.json ships, that the skills are copied flat on Codex. The PR body says nobody in this phase read the installer or the build script, so it is unsourced on both sides. Fix: cite the file that defines the behaviour (the installer, build-plugin.ps1) the way every other rule cites a page, so a reader can check it and so it gets corrected when the build changes.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+    on Codex.** Both hosts, one landing place each. On Claude Code the installer copies the whole
+    bundled plugin folder — skills, server, corpus — to `~/.claude/skills/housecarl/`, and the
+    package carries a `.claude-plugin/marketplace.json` so the folder can be added with `claude
+    plugin marketplace add <folder>`; the skills are invoked as `/housecarl:<name>`. On Codex, which
```
Avick3110, line 266: This landing place cannot be squared with four earlier rules. Rule 17 says `~/.claude/skills/<skill-name>/SKILL.md` is the personal-skill location; rule 10 says a directory there with no `SKILL.md` in it is not a skill; rule 18 says the bundle ships as a plugin because the plugin namespace is the collision answer; rule 25 says an installed plugin runs from `~/.claude/plugins/cache/...` and the plugin root is ephemeral and must not be written to. A plugin folder sitting at `~/.claude/skills/housecarl/` is a fifth arrangement none of those describe, and the reader cannot tell which of the three mechanisms is in force. Fix: say which one the install actually uses, and make rules 18 and 25 agree with it.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+    is different: Anthropic's schemas and the spec both name `evals/evals.json` for outcome evals
+    (skill-creator `references/schemas.md`; https://agentskills.io/specification), and the only other
+    trace is Anthropic's packager excluding `evals/` from a `.skill` zip "only at the skill root"
+    (https://raw.githubusercontent.com/anthropics/skills/main/skills/skill-creator/scripts/package_skill.py),
+    which shows the location is a convention rather than a requirement. Reason for keeping the
+    bundle's own filename: the tests travel with the thing they test, the exclusion rule means they
+    cost nothing at packaging time, and nothing reads the file but us. What the file holds is rule
+    123.
+
+15. **A Codex-only sidecar lives at `skills/<skill>/agents/openai.yaml`, one directory level inside
```
Avick3110, line 216: Rule 15 requires the `agents/openai.yaml` sidecar in both packages, and rule 56 requires `dependencies.tools` in it on every skill. Every citation for the file comes from the plugin submission path, and the PR body lists as unanswered whether Codex reads the file at all for a locally installed skill under `~/.agents/skills/` — which rules 21, 23 and 34 say is the only Codex path the bundle uses. So the standard mandates a file it cannot say does anything on the path it ships to. Fix: say plainly that the file is written for a possible submission and is not known to be read on the local install path, or hold both rules until the behaviour is known.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+63. **Claude Code's listing budget is 1% of the context window, with a hard per-entry cap of 1,536
+    characters.** Claude Code only: "The budget scales at 1% of the model's context window. When the
+    listing overflows, Claude Code drops descriptions starting with the skills you invoke least, so
+    the skills you use most keep their full text"; "each entry's combined text is capped at 1,536
+    characters regardless of budget. The cap is configurable with `skillListingMaxDescChars`"; the
+    listing "always contains every skill name", and shortening "can strip the keywords Claude needs
+    to match your request" (https://code.claude.com/docs/en/skills.md). The settings are
+    `skillListingBudgetFraction` (default `0.01`), `skillListingMaxDescChars` (default `1536`), and
+    the `SLASH_COMMAND_TOOL_CHAR_BUDGET` environment variable as a fixed character count
+    (https://code.claude.com/docs/en/settings-reference.md). Fourteen skills at the cap would present
```
Avick3110, line 846: Rule 63 computes the Claude Code listing over fourteen entries; rule 64, twelve lines later, computes the Codex listing over fifteen — the fourteen skills plus the umbrella. Only one can be right about the umbrella: either it ships in both packages and rule 63 is short by one, or it is Codex-only, which rule 22 implies but never states. Fix: say in rule 22 which packages carry the umbrella, and use the same count in both arithmetics.

Avick3110 (commented):
18 findings: 2 contract breaks, 5 contradictions or number mismatches, 1 process rule, 1 unsourced added rule, 5 rules an author cannot act on, 3 duplicated passages, 1 broken sentence.
Most important: rule 21 states the install layout for both hosts with no source and no house-choice label, and its Claude Code landing place cannot be squared with rules 10, 17, 18 and 25 — the fixed fact the whole packaging section rests on is the one fact nobody checked.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+    omit some skills from the initial list and show a warning"
+    (https://learn.chatgpt.com/docs/build-skills.md). The budget is set by
+    `skills.max_context_tokens` — "Defaults to 2% of the model's context window. Explicit values are
+    capped at `10000` tokens" (https://learn.chatgpt.com/docs/config-file/config-reference.md). Codex
+    installs fifteen entries — the fourteen skills plus the umbrella — each carrying name,
+    description and path. At the 8,000-character floor that is 533 characters per entry for
+    everything, and name plus absolute path costs on the order of 80 to 100, leaving roughly 430 to
+    450 characters of description before Codex starts shortening across the whole catalogue.
+
+65. **Every houseCARL description fits in 400 characters.** **House choice**, derived from rule 64;
```
Avick3110, line 863: The arithmetic does not leave the margin the reason claims. Fifteen entries at 400 is 6,000, plus the roughly 1,350 of names and paths this rule counts, is 7,350 of the 8,000 floor — leaving about 650 characters for every other skill on the user's machine, which by rule 64's own per-entry figure is one entry, not margin. The 400 also goes silently wrong the moment a fifteenth skill is added, which line 1038 explicitly contemplates. Fix: express the budget as the floor divided by the entry count, or state the count the 400 assumes so the next author knows to recompute it.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+  v2.1.199 changelog says "up to 5". The standard picks no number, because rule 71 forbids depending
+  on stacking at all.
+- **Anthropic contradicts itself on how many skills is too many.** The overview says "you can install
+  many Skills without context penalty"
+  (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview); the enterprise page
+  says "limit the number of Skills loaded simultaneously to maintain reliable recall accuracy … stop
+  adding when performance degrades"
+  (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/enterprise). **House choice:**
+  follow the enterprise page, because it is the one that names a failure mode and it agrees with both
+  hosts' published budgets. A new houseCARL skill earns its catalogue entry against the fourteen
```
Avick3110, line 1038: "A new houseCARL skill earns its catalogue entry against the fourteen already there, rather than merely being useful" is a gate, not a rule an author can follow: it names no test for "earns" and nobody to decide. CLAUDE.md says not to add a process rule to catch a mistake. Fix: state the measurable consequence instead — every added skill shrinks the per-description budget in rule 65 — and let the number do the work.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+50. **A houseCARL skill writes `name`, `description`, and `license`, and nothing else.** **House
+    choice.** Claude Code documents twenty fields, and the same page says that outside Claude Code
+    "packaging or upload fails with a hard error instead of ignoring the field", quoting the error
+    naming one of its own documented fields: "Unexpected key(s) in SKILL.md frontmatter:
+    argument-hint. Allowed properties are: allowed-tools, compatibility, description, license,
+    metadata, name" (https://code.claude.com/docs/en/skills.md). The reference validator errors the
+    same way. Codex is silent on unknown keys in either direction; the only positive signal is
+    practice — OpenAI's curated `linear` skill ships a `metadata` key. Reason: one file that is valid
+    everywhere, and no field whose behaviour depends on which host opened it. A Claude Code extension
+    field is added only if a skill genuinely needs it, and then the skill says out loud that it has
```
Avick3110, line 640: "and then the skill says out loud that it has left the portable set" — an author cannot act on this. Where does it say it: a line in the body, a comment beside the frontmatter, the changelog? And to whom, given the body is read by a model rather than by a maintainer. Fix: name the place, or cut the clause and let the field's presence in the file be the record.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+
+105. **Pick one default; do not offer a menu.** Both hosts. "**Avoid offering too many options** —
+     Don't present multiple approaches unless necessary"
+     (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices); "**Provide
+     defaults, not menus.** When multiple tools or approaches could work, pick a default and mention
+     alternatives briefly rather than presenting them as equal options", one of the three named
+     causes of an agent wasting time being "too many options presented without a clear default"
+     (https://agentskills.io/skill-creation/best-practices.md).
+
+106. **Use one term for one thing, throughout the skill, and never state the same rule twice in
```
Avick3110, line 1344: The bolded rule has two halves and only the first is in the citations. "Choose one term and use it throughout the Skill" supports one term for one thing. "never state the same rule twice in different words" is a separate and much stronger rule; neither quote supports it, and it carries no House choice label, which the opening contract requires of a rule resting on no page. Fix: split it out and label it with its reason, or cut it.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+     say once in the skill which server it comes from.** **House choice** against a stated rule. The
+     platform advice is to qualify: "always use fully qualified tool names to avoid 'tool not found'
+     errors. **Format:** `ServerName:tool_name`… Without the server prefix, Claude may fail to locate
+     the tool" (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices).
+     Codex is silent on tool naming inside a body. Reason: the server prefix is host-specific, a
+     prefix written for one host is wrong text in the other, and `standards/NAMING.md` fixes the bare
+     name as the project's spelling. Where the prefix genuinely matters, name it in prose rather than
+     baking it into every mention.
+
+115. **Stay inside Codex's byte ceilings.** Codex sets hard ones for skills imported through MCP —
```
Avick3110, line 1440: These ceilings are stated for skills imported through MCP — a path rules 21, 23 and 34 say the bundle does not use — yet the rule binds the bundle to them on the grounds that "They are the only hard numbers". That is exactly the reasoning rule 49 refuses when it declines the Help Center's 200-character cap because it governs an upload path the bundle does not ship to. Fix: apply the same test — say the limits bind only if a skill is ever imported through MCP — or drop the rule.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+     threshold"; a 60/40 train/validation split, "Select the best iteration by its validation pass
+     rate", "Five iterations is usually enough"
+     (https://agentskills.io/skill-creation/optimizing-descriptions.md). The negative cases carry the
+     weight: "The most valuable negative test cases are **near-misses** — queries that share keywords
+     or concepts with your skill but actually need something different", while "'Write a fibonacci
+     function' — obviously irrelevant, tests nothing". And the fix for a failure is not the failed
+     wording: "Avoid adding specific keywords from failed queries — that's overfitting. Instead, find
+     the general category or concept those queries represent" (same page).
+
+119. **The eval set covers Codex's five categories and is scored with Anthropic's numbers.** **House
```
Avick3110, line 1523: Anthropic's numbers in rule 118 are trigger rates over should-trigger and should-not-trigger queries. Codex's category 3 (incomplete input that should draw a follow-up question) and category 5 (edge cases where the skill must not invent information) are not trigger questions and cannot be scored by a trigger rate. As written the rule tells an author to score five categories with a method that fits three of them. Fix: say that categories 1, 2 and 4 are scored by rule 118's trigger rate and that 3 and 5 are graded by rule 121's pass or fail with quoted evidence.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+     not the number the standard cares about.
+
+128. **The bundle is tested by the Anthropic method and then re-run once on Codex against Codex's
+     five categories.** **House choice** on a silence. Codex has no eval harness, no test runner, no
+     trigger-rate metric, no CLI validator and no scoring rubric; its published mechanical bar is the
+     submission code list plus an automated policy and security scan of skill content, which "may
+     block submission or require remediation", and its published behavioural bar is human review of
+     submitted cases (https://developers.openai.com/plugins/deploy/submission). Reason: Codex
+     publishes a procedure but no measurement, and running the Claude Code measurement twice is
+     cheaper than inventing a second one. The testing consequence of the two hosts' different
```
Avick3110, line 1637: "The testing consequence of the two hosts' different invocation switches (rule 72) is the same: a measurement is run per host to confirm what took." Rule 72 sets neither switch on either host, so there is no switch whose effect could be confirmed, and the sentence does not parse — "confirm what took" is missing a word. Fix: cut it, or say what the per-host measurement is and why it is needed.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+   the two apart: "**Skills** are reusable content you can load into any context; **Subagents** are
+   isolated workers that run separately from your main conversation"
+   (https://code.claude.com/docs/en/features-overview.md), and "Use a skill when you want the
+   procedure to play out inside the main thread so you can see and steer each step"
+   (https://claude.com/blog/steering-claude-code-skills-hooks-rules-subagents-and-more). **House
+   choice:** where a job wants isolation the skill says so in the procedure rather than becoming a
+   worker definition. Reason: Claude Code can preload a named skill into a subagent — "The full
+   content of each listed skill is injected into the subagent's context at startup"
+   (https://code.claude.com/docs/en/sub-agents.md) — and no Codex page describes preloading a named
+   skill into a subagent, so a skill that assumed it would work on one host only.
```
Avick3110, line 97: The sentence ends mid-thought: "and no Codex page describes preloading a named skill into a subagent, so a skill that assumed it would work on one host only." Fix: finish it — presumably that a skill assuming preloading would work on one host only.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+    choice** on a within-vendor difference. Anthropic states two thresholds: "For reference files
+    longer than 100 lines, include a table of contents at the top. This ensures Claude can see the
+    full scope of available information even when previewing with partial reads"
+    (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices) and "For large
+    reference files (>300 lines), include a table of contents"
+    (https://raw.githubusercontent.com/anthropics/skills/main/skills/skill-creator/SKILL.md). Codex
+    is silent. Reason: the stated reason — partial reads — applies at 100 lines as much as at 300,
+    and a contents list costs a few lines.
+
+88. **Split game-generic knowledge from Skyrim-specific knowledge structurally: separate files, or
```
Avick3110, line 1154: Stated as an always with no threshold, and line 1475 admits no source gives one. Several skills in the bundle are entirely Skyrim-specific; their author cannot act on this rule, and splitting a file to hold a generic half that does not exist yet costs the reader a second file for nothing. The reuse case — a future game target — is not a fact the standard establishes anywhere. Fix: make it conditional (split where a skill actually carries both kinds of knowledge), or drop it.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+    of these directories — its examples show `reference.md`, `examples.md`, `scripts/helper.py` at
+    the skill root (https://code.claude.com/docs/en/skills.md). **House choice:** use the three
+    spec and Codex names on both hosts, because Codex names them and Claude Code has no competing
+    convention to violate. Anthropic's own shipped skills break this — `pdf` puts `reference.md` and
+    `forms.md` at the skill root, `mcp-builder` uses `reference/` singular, and the platform's
+    good-practice example writes `reference/finance.md`
+    (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices). That is
+    practice, not rule, and not the practice to copy.
+
+14. **Each skill carries its own eval set at `evals/eval_set.json` inside the skill folder.** **House
```
Avick3110, line 205: Rule 14 mandates a file at a name the standard itself says is not the documented one, and its own reason is that nothing but us reads it; rule 123 then says the baseline result "was recorded with the evidence that produced it" without naming where. So an author is told to create a file at a non-standard name, in a place no tool looks, for a reader nobody names. Judgement call rather than a defect, but the cheap fix is the documented `evals/evals.json` plus a sentence saying which command or person reads it; if the answer is nobody, the rule is not paying for its lines.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+- **Nothing says what an unknown frontmatter key does in Claude Code or Codex**, while `skills-ref`
+  errors on one. The testing consequence is that a structural pass on one validator is not a pass on
+  the others, which is why rule 117 takes the union.
+- **Where the skill listing sits is stated three ways.** The Anthropic platform says name and
+  description "are loaded into the system prompt"; Claude Code says only that it "loads a listing of
+  skill names and descriptions into context"; the spec's client guide treats placement as a host
+  choice. Nothing in the bundle depends on which is true; it relies only on the fact that name and
+  description are present at decision time and the body is not.
+
+## House choices
```
Avick3110, line 1667: The House choices table restates all 68 choices and their reasons in words different from the rules themselves — 74 lines, and by rule 106's own second half, the same rule stated twice. It is also the part most likely to drift from the rule text, since a rule can be edited without the table noticing. Fix: make it an index — choice number, rule number, short title — and let each rule carry its reason once.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+     files per skill, up to five skills per import
+     (https://learn.chatgpt.com/docs/codex-manual.md) — plus the archive caps recorded in the
+     packaging silences above.
+     No Anthropic page states a byte ceiling on bundled files; its nearest statement is about
+     read-time context cost, not storage. They are the only hard numbers and nothing on the Claude
+     side conflicts.
+
+### Where the sources are silent or disagree
+
+- **All-caps.** Anthropic contradicts itself, OpenAI is silent, and both vendors' shipped skills use
```
Avick3110, line 1451: These four bullets say nothing the rules do not. Each restates a disagreement the rule already states and then points back at it: "House position and reason in rule 103", "House position in rule 102", "House choice: 100 (rule 87)", "House choice: adopt the Codex list (rule 82)". Fix: cut all four. The same pattern is at lines 469-470, 488-489 and 490-493 and can go the same way.

standards/HOUSECARL_SKILL_AUTHORING.md
```diff
@@ -1,444 +1,1819 @@
...
+    them.** **House choice**; the documentation is one sentence long. Claude Code says "Document
+    changes in a `CHANGELOG.md`" and nothing more — no location, no format, and no statement that
+    anything reads it (https://code.claude.com/docs/en/plugins-reference.md). The Agent Skills spec,
+    the Anthropic platform docs, and every Codex page are silent on a changelog. Reason: the file is
+    for people, so it lives where the shipped package lives and says what a user would notice,
+    matching the repository's existing practice.
+
+### Where the hosts differ
+
+| Point | Claude Code | Codex | What the bundle does |
```
Avick3110, line 443: Judgement call rather than a defect: this table and the one at line 998 repeat rules already given in prose, with the same URLs pasted a second time, and the last column repeats each rule's decision along with its rule number. That is roughly twenty of the file's longest lines carrying nothing new for a reader who has just read those rules. Fix: if the table is meant as a quick index, drop the citations and the restated decisions and keep point / Claude Code / Codex / rule number.

Avick3110:
Closed unmerged. The deliverable was misdefined: the skill surface gets a ground-up audit (does each skill deserve to exist, evidence gathered by doing the thing first) rather than a single standard. The review-hardened documentation corpus behind this draft is kept in dev/references; the draft is discarded.

Issue #583, opened 2026-09-05
[Bug] Claude Code cuts every MCP tool description at 2,048 characters; twelve 2.0 tools lose most of their grammar


Avick3110:
Measured 2026-09-05 on the 2.0.0-beta.1 install, Claude desktop app. The `housecarl_records` description as delivered to a session (or to a subagent through ToolSearch) ends mid-sentence at character 2,048 of 8,378 with a `[truncated]` marker. The parameter schemas (up to 14.9 KB for `records`) arrive whole.

From the published tools/list, twelve of the 31 tools exceed the cap: records 8,378 / apply 7,886 / check 7,489 / create 6,511 / copy 5,243 / forward 4,132 / merge_plugins 3,363 / nif_inspect 3,038 / compact_plugin 3,014 / place 2,976 / remove 2,378 / nif_set 2,116 characters. Everything below 2,048 is intact.

What is lost sits in the tail: for `records`, the rest of the where= grammar (quantified steps' declared cost, `*parent`, membership, the `->` link step, the `winner` provenance term), the whole SOURCE section, the reverse-reference index cost declaration, the union-arm tip. The descriptions were written on the assumption that they are read whole, and the front-loaded part is the SELECT summary.

Not a cause of #582 (the agent's call there was correct by the grammar), but every session on this host has been composing calls against a quarter of the `records` description. Whether the cap is the desktop app's, the CLI's, or the SDK's is not established; the ToolSearch path shows it, and the number is exactly 2,048.

Options are the surface's to choose: move grammar out of the description into the parameter descriptions (which arrive whole); shorten; or split. This decides how the 2.0 descriptions are shaped, so it belongs before item 27 rewrites the skills against them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Avick3110:
Agreed approach (Aaron, 2026-09-06), on record here so it survives the session:

1. Every tool description is bounded at 2,048 characters, the cap the client cuts at.
2. Grammar that currently lives in a description moves into the parameter descriptions it belongs to: for `records`, the where= predicate language into `where`, the SOURCE section into `source` and `versus`, walk semantics into `walk`, form details into `project`. The description keeps the axes sentence, the FormID forms, and the composition rule. Parameter descriptions arrive whole (measured up to 14.9 KB), so nothing is lost, only moved.
[...]

Issue #570, opened 2026-09-05
[Bug] ci-all probes write to fixed paths under the shared temp dir, so parallel runs collide


Avick3110:
## Summary

Several probes in `ci-all` write their fixtures to fixed paths under the shared `%TEMP%\hc-*` directories. With two or more sessions running `ci-all` at once, a run fails with file-in-use or access-denied on `inplace-guard`, `mo2instance-probe`, `pkcu-regression`, `compact-readback-guard` (and a different set the next time), then passes clean when rerun alone. Seen four times on 2026-09-06 across parallel fold sessions.

## Fix shape

Each probe run takes a per-process root (`Path.GetTempPath()/hc-<pid>-<name>` or the session scratchpad when set) instead of a fixed `hc-<name>` path, and cleans it up on exit. No probe should share a fixture directory with another process.

PR #585, opened 2026-09-06
Bring the records tool description under the 2,048-character cap

Avick3110:
`housecarl_records` published an 8,378-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the predicate language, the source rules, the projection forms and the transport rules were all being described in a part of the text the model never saw. Nothing is deleted here: each paragraph moves into the description of the parameter it is about, and where the parameter already said the same thing the description's copy is dropped as a duplicate rather than pasted in twice. The description keeps the one-read-surface sentence, the FormID forms, the set-valued and `@<absolute path>` rule, the composition rule, one line per axis naming its parameters, and the pointer to the write tools. A new wire test holds the bound on the published `tools/list` and names the eleven tools still over it, one line each, so the next eleven PRs each delete their own line.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_records` description | 8,378 | **2,036** |
| Top-level parameter descriptions | 5,573 | 9,094 |
| All schema descriptions (parameters + nested DTO members) | 11,975 | 16,394 |

The description shed 6,342 characters and the schema gained 4,419; the 1,923-character difference is text the description was repeating from a parameter that already carried it (the axis lists, the `rows` form details, the quantified-step note, `previous_provider`, the SkyPatcher overlay, `counts_only=`, `to_file=`).

## Harvest

|---|---|
| "Read Bethesda records from the load order — ONE read surface: … compose in a single call." | kept in description |
| "A FormID is 'XXXXXX:Plugin.esp' — 6 hex digits, a colon, the defining master's filename." | kept in description |
| "The RUNTIME form the game, the console, Papyrus logs, SKSE logs and crash logs print is accepted too … prints each record's runtime FormID beside its own." | kept in description |
| "It is taken wherever a parameter holds nothing but FormIDs … so it takes the plugin-qualified form only." | kept in description |
| "Every list-valued parameter is set-valued … epoch-checked against the then-current build)." | kept in description |
| "SELECT: formids= \| types= \| plugins= \| conflicts_only= \| where= \| references= \| walk=" (the axis list) | kept in description (the SELECT axis line) |
| "plugins= (scope: which records are considered)" | `plugins` (already said there in the same words) |
| where= comparisons: 'BasicStats.Damage >= 50', 'editorid contains Iron', 'editorid startswith REQ_' | `where` |
| where= flag tests: 'FirstPersonFlags has Body', `has_any`, `has_none` | `where` |
| where= quantified steps `[*any]/[*all]/[*none]/[*count]`, the empty-list rule, and their declared cost (list length; one winner fetch per element under `->`) | `where` |
| where= containment step `*parent`, the INFO/DIAL and worldspace examples, leads-and-chains | `where` |
| where= presence 'VirtualMachineAdapter exists' | `where` |
| where= membership `in`/`not in`, `@file`, and the comma/bracket caveat | `where` |
| where= the single `->` link step | `where` |
| where= the `winner = X.esp` provenance term and its declared cost | `where` |
| references= reverse-reference index: built on the first call, one whole-order link-walk, reports that cost and its per-plugin freshness key; a bounded call is cheaper | `references` |
| references= the `!` negation, AND composition, and the `!@file` spelling | `references` |
| references= the orphan sweep | `references` |
| the UNION-ARM tip (`Configuration.Level.LevelMult >= 0`) | `where` |
| "formids= COMPOSES with the scan terms … needs no types=/plugins=)." | kept in description |
| "The walk= construct is a SELECT term too: what it reaches is a selection any reading form can consume." | kept in description |
| "SOURCE decides WHOSE version you read." | `source` |
| "Default: the load-order winner." | `source` |
| "Naming a plugin … reads THAT plugin's version wherever it lives … the response STATES which arm resolved (active, or out-of-load-order and from where)." | `source` |
| "A plugin found in neither place is refused naming both places searched." | `source` |
| "A record the named plugin does not touch is refused naming the plugins that DO touch it — never silently absent." | `source` |
| "An off-order file's content sits OUTSIDE the epoch fingerprint and the response says so." | `source`, folded into the one epoch-fingerprint sentence that already covered the SkyPatcher overlay, so the rule is stated once for both |
| "versus= is the comparison REFERENCE pole (delta/tree)" | `versus` (already said there) |
| "{\"overlay\": \"skypatcher\", \"state\": \"pre\"\|\"post\"} reads around the SkyPatcher INI layer" | `source` and `versus` (already said on both) |
| "\"previous_provider\" (a versus= value) is the plugin immediately below the SUBJECT in the record's touching stack." | `versus` (already said there, with its four declared cases) |
| "PROJECT is a single form (see project=): identity \| summary (default) \| … \| info_order." | `project`, as a bare roster of the ten names; `project.form` remains the one place each form is described |
| "Sub-parameters live INSIDE the form that uses them … no flat spelling for an illegal pairing." | `project` |
| "form='rows' is the compact per-row view of a LIST field … ABSENT optionals omitted." | `project.form` (already said there) |
| "Auditing a 40-row condition stack … is one call, not an index probe per row" | `project.form` |
| "an indexed path (project.fields=[\"Conditions[0]\"]) folds that one element" | `project.form` (already said there) |
| "a field that is not a list is refused by name" | `project.form` (its weaker "fails loud" replaced) |
| "On form='fields' a path may QUANTIFY its step … dense repeats the identity columns across those rows." | `project.fields` (already said there) |
| delta: "each delta line shows the subject's value with the reference's labeled by its plugin; versus=\"previous_provider\" answers 'what did this plugin change relative to what sat beneath it'." | `project.form` |
| tree: "lists EVERY plugin touching each record (load order, winner last) with each provider's only-the-fields-that-differ against the reference — the conflict-resolution view." | `project.form` |
| "Both compare by the content-keyed, truncation-honest engine (list reorders flagged; a truncated deep read is reported, never claimed 'identical')." | `project.form` |
| info_order: the merged sequence, first passing line, the PNAM re-listing rule, invisible to a diff | `project.form` |
| "A quest's topics select by composition: types=[\"DIAL\"] where=[\"Quest = <quest formid>\"]." | `project.form` |
| "format= 'text' \| 'json' \| 'dense' (positional columnar cells 1:1 with the requested fields — depth expansion and the everything form are inexpressible in it)" | `format` |
| "limit=/offset= page a scan in exact windows WITHIN one epoch (… re-run from offset=0, do not stitch)" | `limit` and `offset` |
| "counts_only= returns the accounting without rows" | `counts_only` (already said there) |
| "max_chars= caps the RENDER … SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, row schema, and epoch) and the response names the file" | `max_chars` |
| "to_file= forces that disposition to your own path" | `to_file` (already said there) |
| "Every response carries epoch=<hex> — the identity of the index build it was answered from." | `format`, with both spellings named: `epoch=<hex>` on text and dense, an `epoch` member on json |
| "This tool never writes. Authoring goes through the write tools (…)." | kept in description |

`walk=` needed no move: the only walk sentence in the description was the composition rule, which stays, and the traversal semantics were already on `walk` and its DTO members.

The description gained one thing it did not carry before: the SOURCE axis line now names `where_source=` and `fields_source=` beside `source=` and `versus=`. The roster is the description's only index of the parameters now, and those two were on neither line, so the lever that makes `where=` match the live winner was unreachable from it.

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`, `[Collection("server")]` `[Trait("tier", "stdio")]`:

- `EveryPublishedDescriptionFitsTheClientsCut` — a theory over every published tool, asserting the served description is at most 2,048 .NET characters. Tools in `StillOversized` are skipped. Ran once before the change: `housecarl_records` was the only failure outside that set, at 8,378.
- `EveryToolNamedOversizedStillIs` — a `Fact` iterating `StillOversized`, asserting each named tool is still over the bound, so a line cannot go stale. Verified it fails on a stale entry by temporarily adding `ToolNames.Records`: "housecarl_records's published description is 2036 characters, which is inside the 2048 bound — delete its line from StillOversized". A `Fact` rather than a theory because the set is designed to empty out, and xunit fails an empty `MemberData` source with "No data found" — the last PR of the wave would have gone red for a reason unrelated to what it changed.

The theory rows come off the checked-in name capture, because `MemberData` is evaluated before the fixture exists. `PublishedNameAnchorTests` already holds that capture equal to the published set in both directions and names the tool that moved, so this file reuses its `Captured()` rather than restating the property; `Description()` looks the tool up with `TryGetValue`, so a line left behind for a retired tool gets the sentence written for that case instead of a `KeyNotFoundException`.

Both print each tool's measured length to the test output.

## What was run

```
    1620 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. No checked-in capture snapshots the descriptions: `src/housecarl-mcp-tests/data/tools-list-2.0.json` holds tool names only, so nothing needed regenerating.


plugin/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **Tool descriptions now fit inside the 2,048 characters Claude Code shows, and the grammar that was being
```
Avick3110: **The headline claim is false as shipped.** "Tool descriptions now fit inside the 2,048 characters Claude Code shows" reads as a statement about the surface, but only `housecarl_records` was brought under the bound. Measured on this branch's `tools/list`, 11 of the 31 published descriptions are still over it: `housecarl_apply` 7,886, `housecarl_check` 7,489, `housecarl_create` 6,511, `housecarl_copy` 5,243, `housecarl_forward` 4,132, `housecarl_merge_plugins` 3,363, `housecarl_nif_inspect` 3,038, `housecarl_compact_plugin` 3,014, `housecarl_place` 2,976, `housecarl_remove` 2,378, `housecarl_nif_set` 2,116 — that is `StillOversized` verbatim. A user reading the changelog concludes their whole tool surface now arrives intact and stops looking for truncated grammar on `apply` or `create`. The last sentence ("names the tools still over it") only walks it back for someone who reads to the end of the entry. Say what changed: the records description came under the bound, and a test now holds the bound and names the eleven that have not.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
+  cut off lives on the parameters it belongs to.** A description longer than that cap lost its tail on the way
+  to the model, so the predicate language, the source and comparison rules, the projection forms and the
+  transport rules were being described in a part of the text nobody read. None of it was deleted: each piece
+  moved into the description of the parameter it is about — `where=`, `references=`, `source=`, `versus=`,
+  `walk=`, `project=` and the transport parameters — which the client delivers whole however long they are.
```
Avick3110: **`walk=` did not receive any moved grammar — this diff does not touch it.** The `walk` parameter's `[Description]` in `RecordsTools.cs` is unchanged context in this PR, and the PR body says so explicitly: "`walk=` needed no move: the only walk sentence in the description was the composition rule, which stays." Listing it beside `where=`/`references=`/`source=`/`versus=`/`project=` tells a reader that walk semantics that used to be in the description now live on the parameter, so someone chasing a missing traversal sentence looks at a parameter that gained nothing. Drop `walk=` from the list.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
```diff
@@ -0,0 +1,102 @@
...
+                  .ToArray();
+    }
+
+    public static IEnumerable<object[]> EveryPublishedTool() =>
+        CapturedNames().Select(n => new object[] { n });
+
+    public static IEnumerable<object[]> OversizedTools() =>
+        StillOversized.OrderBy(n => n, StringComparer.Ordinal).Select(n => new object[] { n });
+
+    string Description(string tool) => _s.PublishedTools[tool].GetProperty("description").GetString()!;
```
Avick3110: **A retired tool turns the stale-entry guard into an opaque crash.** `Description` indexes `_s.PublishedTools[tool]` with no containment check, and `OversizedTools()` feeds it names straight out of `StillOversized` without ever checking them against the published set. Concrete scenario: `housecarl_nif_set` is retired or renamed and its `StillOversized` line is left behind — `EveryToolNamedOversizedStillIs` dies on `KeyNotFoundException: The given key was not present in the dictionary`, and the actionable sentence written for exactly this case (line 99, "delete its line from `StillOversized`") never runs. The same happens in `EveryPublishedDescriptionFitsTheClientsCut` for a captured name the server no longer publishes; the `Fact` reports that case well, but the theory rows throw first. One `TryGetValue` with a sentence naming the tool and where its line lives would keep the failure legible.

```suggestion
    string Description(string tool)
    {
        Assert.True(_s.PublishedTools.TryGetValue(tool, out var t),
            $"'{tool}' is named here but the server does not publish it — it was renamed or retired. " +
            "Update its line in StillOversized and in data/tools-list-2.0.json.");
        return t.GetProperty("description").GetString()!;
    }
```

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -95,81 +95,14 @@ public sealed class RecordsWalkExclusion
...
-         "TRANSPORT: format= 'text' | 'json' | 'dense' (scan lane: positional columnar cells 1:1 with the requested " +
-         "fields — by that definition depth expansion and the everything form are inexpressible in it); limit=/" +
-         "offset= page a scan in exact windows WITHIN one epoch (different epochs across pages ⇒ the order changed " +
-         "— re-run from offset=0, do not stitch); counts_only= returns the accounting without rows; max_chars= caps " +
-         "the RENDER — an over-cap result is not truncated, it SPILLS in full to a server-side JSONL artifact " +
-         "(line 1 = manifest with the query echo, row schema, and epoch) and the response names the file; to_file= " +
-         "forces that disposition to your own path. Every response carries epoch=<hex> — the identity of the index " +
-         "build it was answered from.\n\n" +
+         "Each axis's grammar is on its own parameters:\n" +
+         "SELECT — formids= | types= | plugins= | conflicts_only= | where= | references= | walk=.\n" +
```
Avick3110, line 99: **The axis roster is now the description's only index of the parameters, and it is missing two of them.** `where_source=` and `fields_source=` appear on neither the SELECT line nor the SOURCE line below, and neither parameter's own description opens with an axis tag the way every other one does (`where_source` starts "Which BODY the where= predicates decide the MATCH on", `fields_source` "The pole field VALUES display from"). Before this change the description was flowing prose and the lists were incidental; now they are the map. Concrete: a caller wanting the post-patch audit body runs a `plugins=`-scoped scan, reads the SELECT roster to find the lever that makes `where=` match the live winner, sees seven parameters none of which is it, and never learns `where_source="winner"` exists. Add both to the SOURCE line — they are the two poles that split matching from display.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -31,7 +31,7 @@ public sealed class RecordsScope
...
...OWNS child records — a cell's placed references, a topic's INFO lines, a worldspace's cells — it also states, per such field, which providers DECLARE children there (a COLLECTION field) or how many do (a SINGULAR one, e.g. Cell.Landscape), and says so when none do) | 'info_order' (DIAL topics only: the effective MERGED INFO sequence across every touching plugin, with MOVED annotations — the 'why does the wrong line play' diagnostic. The game walks the sequence top to bottom and plays the FIRST passing line; re-listing a line appends it to the BOTTOM unless the plugin also carries its PNAM, so a reorder changes which line answers while every field stays identical — invisible to a diff, which is what this form is for. A quest's topics select by composition: types=[\"DIAL\"] where=[\"Quest = <quest formid>\"]) | 'chain' (a walk's own paths, endpoints and cycles rather than the records it reached; needs walk=, and carries the NPC-template inheritance report and the reverse MGEF carrier rows). The comparison forms 'delta' and 'tree' both compare by the content-keyed, truncation-honest engine: a list reorder is flagged, and a truncated deep read is reported, never claimed 'identical'.")]
```
Avick3110: **The moved sentence lost its `project.` qualifier.** The tool description spelled this `project.fields=["Conditions"]`; the copy that landed here says `fields=["Conditions"]`. `housecarl_records` has no top-level `fields=` — `RemedyHarvest.WrongLevers` names a bare `fields=` as a wrong lever for exactly this reason (`"narrow with 'fields=' (it has project.fields=)"`), and the suite fails any refusal sentence that spells it this way. A model reading this example and sending `{"project": {"form": "rows"}, "fields": ["Conditions"]}` gets a bind failure on a call the description just told it to make. (The neighbouring "takes fields=" and "takes depth=" are pre-existing and read as sub-parameter names in context; a full `key=[value]` example does not.)

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -182,29 +115,29 @@ public static string Records(
...
             RecordsProject? project = null,
         [Description("SELECT: the traversal construct — follow record-to-record links from this call's own SELECT (the seeds) and select what the walk reaches; project.form='chain' renders the paths/endpoints/cycles themselves (with, for NPC template chains, the per-category active-vs-masked inheritance report), while any other form reads the reached set like any selection. The walk expands on the WINNER's link graph; source= governs whose version the form then reads.")]
             RecordsWalk? walk = null,
-        [Description("TRANSPORT: 'text' (default) | 'json' (machine-readable document; same accounting in-band) | 'dense' (scan lane: columnar positional rows — the compact bulk-enumeration form).")]
+        [Description("TRANSPORT: 'text' (default) | 'json' (machine-readable document; same accounting in-band) | 'dense' (scan lane: positional columnar cells 1:1 with the requested fields — the compact bulk-enumeration form; by that definition depth expansion and the 'everything' form are inexpressible in it). Every response, whatever the format, carries epoch=<hex> — the identity of the index build it was answered from.")]
```
Avick3110: **"whatever the format" is the one qualifier this sentence cannot carry.** The literal `epoch=<hex>` is the text lane's spelling (`ReadTools.EpochLine`/`EpochInline`); the json lane writes it as a document member, `"epoch": "<hex>"` (`JsonWire.cs:48`, `:1770`). The old description asserted this globally too, but it sat in flowing prose; parked on the parameter that chooses between text, json and dense, "whatever the format" reads as a promise about all three. Concrete: a caller runs `format="json"`, greps the response for `epoch=` to check the build its rows came from, finds nothing, and concludes the stamp is missing. "Every response carries the epoch stamp — `epoch=<hex>` on text, an `epoch` member on json" says the same thing without the trap.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -182,29 +115,29 @@ public static string Records(
...
+        [Description("SOURCE decides whose version you read; this is the SUBJECT of the call. Omit or \"winner\" for the load-order winner (the default). A plugin filename (e.g. \"OldPatch.esp\") reads THAT plugin's version WHEREVER the plugin lives — active in your order, or sitting on disk unticked — you do not have to know which, and the response STATES which arm resolved (active, or out-of-load-order and from where); use {\"file\": \"X.esp\", \"mod\": \"<mod folder>\"} when two mods ship the same filename. A plugin found in neither place is refused naming both places searched. A record the named plugin does not touch is refused naming the plugins that DO touch it — never silently absent. An off-order file's content sits OUTSIDE the epoch fingerprint and the response says so. {\"overlay\": \"skypatcher\", \"state\": \"pre\"|\"post\"} reads around the SkyPatcher INI layer (post = after it replays; INI content sits OUTSIDE the epoch fingerprint and the response says so). \"previous_provider\" is a versus= value only — it is measured FROM the subject this parameter names.")]
```
Avick3110: **The move left this parameter saying the same thing twice, one sentence apart.** The harvested sentence "An off-order file's content sits OUTSIDE the epoch fingerprint and the response says so" lands immediately before the pre-existing overlay clause, which ends "...INI content sits OUTSIDE the epoch fingerprint and the response says so" — the identical predicate, twice, about two different subjects, ~25 words apart. That is exactly the second copy the PR body says it avoided ("where the parameter already said the same thing the description's copy is dropped as a duplicate rather than pasted in twice"); it was not caught because the pre-existing copy is inside a parenthetical. It matters beyond tidiness: two spellings of one fact are two things to edit when the epoch rule changes, and a model reading them consecutively cannot tell whether the second is a restatement or a second, narrower rule. State the epoch-fingerprint rule once, covering both — e.g. "Content read from outside the load order — an off-order file, or the SkyPatcher INI layer — sits OUTSIDE the epoch fingerprint, and the response says so" — and drop the parenthetical's copy.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -182,29 +115,29 @@ public static string Records(
...
             JsonElement? versus = null,
         [Description("The pole field VALUES display from, when it differs from the matching pole: \"winner\" shows the live winner's values on a plugins=-scoped scan (the old winner_fields=true). Display only — where_source= governs matching.")]
             string? fields_source = null,
-        [Description("PROJECT: the shape of the answer — a single form plus its own sub-parameters. Omit for summary rows.")]
+        [Description("PROJECT: the shape of the answer — a SINGLE form plus that form's own sub-parameters: identity | summary (default) | fields | rows | everything | aggregate | delta | tree | chain | info_order. Sub-parameters live INSIDE the form that uses them (depth belongs to fields/rows/everything, group_by to aggregate, fields to fields/rows/delta/tree), so there is no flat spelling for an illegal pairing. Omit for summary rows.")]
```
Avick3110, line 131: **The PR body counts the form list as a duplicate it dropped; the diff pastes it in.** The accounting paragraph lists "the `project=` form list" among the 2,034 characters that were "text the description was repeating from a parameter that already carried it" — i.e. dropped, not moved. But this line gains `identity | summary (default) | fields | rows | everything | aggregate | delta | tree | chain | info_order` verbatim, while `project.form` two hundred lines up still enumerates all ten forms in full with their sub-parameters. So the published schema now carries the enumeration twice, and the body's arithmetic is describing a different change than the one shipped. Either drop the list here and let `project.form` be the single place the forms are named (this parameter's job is "a single form plus that form's own sub-parameters"), or keep it and correct the accounting paragraph so the numbers and the harvest table agree with the diff.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    [Fact]
+    public void TheToolsDrivenAreExactlyTheOnesPublished_SoTheSweepIsNotShort()
```
Avick3110: **This claim is already made, twice, by `PublishedNameAnchorTests`.** That class holds the same capture against the same `_s.PublishedNames` in both directions: `ACapturedNameIsStillPublished_ATypoInItsConstantsValueGoesRedHere` (a theory row per captured name) and `NothingIsPublishedThatTheCaptureDoesNotName_ANewToolAddsItsLiteralHere`. Between them, captured == published is already red on either divergence, and with the offending tool named rather than as a set difference. This `Fact` re-asserts set equality, and `CapturedNames()` above re-implements `PublishedNameAnchorTests.Captured()` — same path, same `GetProperty("tools")`, same ordinal `OrderBy` — minus its non-empty guard, and under a name that already means something different in that class (`CapturedNames()` there returns theory rows). Two readers of one file drift when the file's shape changes. Reuse the existing helper (make `Captured()`/`CapturePath` internal and call it) and delete the `Fact`; the sweep is already pinned by the class that exists to pin it, and this file's own comment can point there instead of restating the property.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        return doc.RootElement.GetProperty("tools").EnumerateArray()
+                  .Select(t => t.GetString()!)
+                  .OrderBy(n => n, StringComparer.Ordinal)
+                  .ToArray();
+    }
+
+    public static IEnumerable<object[]> EveryPublishedTool() =>
+        CapturedNames().Select(n => new object[] { n });
+
+    public static IEnumerable<object[]> OversizedTools() =>
```
Avick3110: **The last PR of the wave gets a red test with a message about none of this.** `StillOversized` is designed to empty out — eleven PRs, one line each — and when the eleventh deletes the last line, `OversizedTools()` yields no rows. xunit 2.9.2 (this project's version) reports an empty `MemberData` source as a failed test: `System.InvalidOperationException : No data found for HousecarlMcpTests.PublishedDescriptionBoundTests.EveryToolNamedOversizedStillIs`. So the PR that finally completes #583 goes red for a reason unrelated to what it changed, and the author has to work out that the fix is to delete the theory. Either say so in the comment above `StillOversized` ("when this empties, delete `EveryToolNamedOversizedStillIs` with the last line"), or give the theory a row it always has — e.g. drive it off `StillOversized.DefaultIfEmpty(null)` and assert the empty case passes.

src/housecarl-mcp/RecordsTools.cs
Avick3110: **A code comment still points at the tool description for two promises this PR moved out of it.** `RenderListAggregate`'s summary (RecordsTools.cs:2128, outside the diff so I cannot anchor there) reads: "it carries the same response envelope as every other form, including the resolved source statement and the epoch-coverage qualifier **the tool description promises unconditionally**." Both promises left the description in this change — the resolved-source statement is now on `source=` ("the response STATES which arm resolved"), the epoch qualifier is on this parameter. The comment is now a pointer to text that is not there, which is worse than a restatement: the next reader greps the description for the promise, does not find it, and cannot tell whether the render's obligation was dropped or merely relocated. One-line fix: name the parameters instead, e.g. "...the epoch-coverage qualifier `format=` promises unconditionally".

plugin/CHANGELOG.md
Avick3110: Folded in 1962f00e: the headline now names `housecarl_records` as the tool whose description came under the cap, and says eleven others are still over it and that the test names each.

plugin/CHANGELOG.md
Avick3110: Folded in 1962f00e: `walk=` is out of the list, which now reads `where=`, `references=`, `source=`, `versus=`, `project=`, `format=` and the other transport parameters.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 5f74d5b6: the example is `project.fields=["Conditions"]` again.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 5f74d5b6: both copies are gone, replaced by one sentence covering both subjects — "Content read from outside the load order — an off-order file, or the SkyPatcher INI layer — sits OUTSIDE the epoch fingerprint, and the response says so."

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 5f74d5b6: the qualifier is gone and both spellings are named — `epoch=<hex>` on text and dense, an `epoch` member on json.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: Folded in 4d1697ce: `Description()` uses `TryGetValue` and names the tool and the two places its name is spelled — its `StillOversized` line and the capture.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: Folded in 4d1697ce: the `Fact` and the private `CapturedNames()` are gone, `PublishedNameAnchorTests.Captured()` and `CapturePath` are internal, and the rows read them, so the capture is parsed in one place with its empty-capture guard. The class comment points at that class for the pin.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: Folded in 4d1697ce, so the follow-on PRs need no thought about it: `EveryToolNamedOversizedStillIs` is now a single `Fact` that iterates the set, prints each length and passes trivially when the set is empty, with a comment saying the set empties as the wave lands and the Fact then guards nothing on purpose. The delete-your-line comment stays on the set.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 5f74d5b6: the summary now reads "the resolved source statement `source=` promises and the epoch-coverage qualifier `format=` promises unconditionally".

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 5f74d5b6: the SOURCE line reads "source= is the subject, versus= the reference pole; where_source= | fields_source= split matching from display". Measured on the published tools/list the description is 2,036 characters, so it fits.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Not folded — the accounting was wrong, not the diff. What sits on `project` is a bare roster of the ten names with no per-form text; `project.form` is still the one place each form is described, and a caller reading `project=` needs to know which names are legal there. So it is a pointer, not a second copy. The PR body is corrected instead: the `project=` form list is out of the dropped-duplicates list, the harvest row says the roster lands on `project` with `project.form` keeping the descriptions, and the arithmetic is re-measured against this commit (description 2,036, top-level parameters 9,094, all schema descriptions 16,394).

PR #586, opened 2026-09-06
Give each ci-all run its own fixture directory

### LINKED ISSUE #570: [Bug] ci-all probes write to fixed paths under the shared temp dir, so parallel runs collide

Avick3110:
Probes name their fixture directories with fixed names, so two `ci-all` runs at once opened the same files and failed each other on file-in-use and access-denied. A run now points its own process at a temp directory named for its pid before any probe starts, so every `Path.GetTempPath()` a probe reaches for lands under a root no other process can name, and deletes that root when the run ends (a file another process still holds open is reported, not thrown, and never changes the exit code). Three probes built their scratch directory from the working directory instead of the temp path, which collided the same way from one checkout; they now build it from the temp path too, so the two ignore rules that existed only to hide those directories are gone. No probe and no assertion changed, and nothing outside the probes referenced any of these paths: the standalone `freshness-capture-guard` step in `.github/workflows/ci.yml` runs through the same dispatch and gets its own root.

A run that is killed never reaches its own cleanup, so the redirect also drops the roots whose pid no process holds — otherwise naming the root for the pid would trade the old fixed names' bounded residue for an unbounded one. Only a definite "no such process" counts as dead: a process the run cannot inspect, such as one at a higher integrity level, keeps its root, because leaving a directory behind is residue the next run picks up while deleting one is a running run's fixtures going mid-run. Residue under our own pid that will not go is reported in the same one line the cleanup uses, and the run takes a fresh sibling root rather than start inside fixtures it failed to empty.

Closes #570

**Reproduction (before the fix).** Two `ci-all` runs started at the same time from one worktree, first attempt, both red: run A failed 10 probes (`import-order-guard`, `inplace-guard`, `nested-create-guard`, `pkcu-regression`, `snapshot-view-guard`, `source-display-guard`, `subclass-remove-guard`, `substruct-nullable-clear-guard`, `upsert-guard`, `verify-loop-guard`) and run B failed 9 (`compile-probe`, `import-order-guard`, `inplace-guard`, `nested-create-guard`, `snapshot-view-guard`, `source-display-guard`, `subclass-remove-guard`, `upsert-guard`, `verify-loop-guard`). The messages name the shared paths directly, e.g. `IOException: The process cannot access the file 'C:\...\Temp\hc-substruct-nullable-clear-guard\corpus-gen\corpus.json' because it is being used by another process` and `IOException: Access to the path 'C:\...\Temp\hc-subclass-remove-guard\A' is denied`.

**Proof.** With the temp root alone, two-at-once dropped to 1 and 2 failures — `import-order-guard` in both and `compile-probe` in one, which is the second collision, on `.import-order-probe-gen` and `.compile-probe-gen` in the checkout root: one run's cleanup deletes the other's scratch mid-compile, and `compile-probe` then dereferences a null `PexPath`. With both commits, two `ci-all` runs at once are 127/127 each, and a run alone is 127/127. That null `PexPath` deref is now guarded too, so the same environmental failure would report the probe's own failed check rather than a throw.

**What was run.** `dotnet build housecarl.sln -c Release`; `dotnet src/housecarl-generator/bin/Release/net9.0/housecarl-generator.dll ci-all` twice concurrently (127/127 each, both exit 0) and once alone (127/127); `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` (1591 passed, 0 failed). The cleanup reported nothing left behind in any run, and the temp directory holds no `hc-*` fixture directory from them. The sweep of dead runs' roots was checked by planting `hc-999999` — not a possible live pid — and watching a solo run remove it.

**Tests.** The liveness check that decides whether a root is residue or a running run's scratch is pinned by `ProbeTempSweepTests`: a pid no process holds is dead, this process is live, and a running process the run cannot open is live. That third case fails before the fix and passes after. The redirect itself still has no test — the only way to reach it is to call it, which repoints the whole test process's temp directory and then deletes it out from under every other test in that process. The honest proof for the redirect is running `ci-all` twice at once, which is above.

### DIFF
--- a/src/housecarl-generator/BsaProbe.cs
+++ b/src/housecarl-generator/BsaProbe.cs
@@ -34,7 +34,7 @@ internal static class BsaProbe
         if (!File.Exists(bsarch)) { Console.WriteLine($"  SKIP  no BSArch at '{bsarch}' (pass its path as arg 1, or set HOUSECARL_BSARCH)"); return 0; }
         if (!File.Exists(bsa)) { Console.WriteLine($"  SKIP  no test archive at '{bsa}' (pass one as arg 2, or set HOUSECARL_TEST_BSA)"); return 0; }
 
-        var work = Path.Combine(Environment.CurrentDirectory, ".bsa-probe");
+        var work = Path.Combine(Path.GetTempPath(), ".bsa-probe");
         var unpacked = Path.Combine(work, "unpacked");            // Mutagen unpack
         var bsarchDir = Path.Combine(work, "bsarch-unpacked");    // BSArch unpack (the parity oracle)
         var repacked = Path.Combine(work, "repacked.bsa");
--- a/src/housecarl-generator/BsaProbe.cs
+++ b/src/housecarl-generator/BsaProbe.cs
@@ -85,7 +85,7 @@ internal static class BsaProbe
             Check(File.Exists(repacked) && new FileInfo(repacked).Length == beforeLen,
                   "non-destructive: the PRIOR archive is left intact after a failed pack");
         }
-        finally { try { Directory.Delete(work, recursive: true); } catch { /* in-dir scratch; non-fatal */ } }
+        finally { try { Directory.Delete(work, recursive: true); } catch { /* temp scratch; non-fatal */ } }
 
         Console.WriteLine();
         Console.WriteLine(fail == 0 ? "================ ALL PASS ================" : $"================ {fail} CHECK(S) FAILED ================");
--- a/src/housecarl-generator/CompileProbe.cs
+++ b/src/housecarl-generator/CompileProbe.cs
@@ -59,7 +59,7 @@ internal static class CompileProbe
             var vanilla = Path.Combine(gameRoot!, "Data", "Source", "Scripts");
             Check(Directory.Exists(vanilla), $"vanilla sources derived from compiler dir exist ({vanilla})");
 
-            var work = Path.Combine(Environment.CurrentDirectory, ".compile-probe-gen");
+            var work = Path.Combine(Path.GetTempPath(), ".compile-probe-gen");
             var outDir = Path.Combine(work, "out");
             Directory.CreateDirectory(outDir);
             File.WriteAllText(Path.Combine(work, "HCGood.psc"),
@@ -95,7 +95,7 @@ internal static class CompileProbe
                 Check(!reBad.Success && reBad.PexPath is null, "recompile (now broken): reports failure, no new .pex");
                 Check(File.Exists(pex), "recompile (now broken): the PRIOR .pex is LEFT INTACT (non-destructive)");
             }
-            finally { try { Directory.Delete(work, recursive: true); } catch { /* in-dir scratch; non-fatal */ } }
+            finally { try { Directory.Delete(work, recursive: true); } catch { /* temp scratch; non-fatal */ } }
         }
 
         Console.WriteLine();
--- a/src/housecarl-generator/CompileProbe.cs
+++ b/src/housecarl-generator/CompileProbe.cs
@@ -83,17 +83,26 @@ internal static class CompileProbe
                 // RECOMPILE behaviour (Aaron 2026-06-06): a prior .pex is LEFT in place — a successful recompile updates
                 // it (success = THIS run wrote it, detected by the write-time advancing, NOT by a pre-delete); a FAILED
                 // recompile must NOT destroy the last good build (non-destructive — the user deletes outputs at will).
-                var pex = good.PexPath!;
-                var firstWriteUtc = File.GetLastWriteTimeUtc(pex);
-                var reGood = PapyrusCompile.CompileObject(compiler, "HCGood", imports, outDir);
-                Check(reGood.Success && File.Exists(pex), "recompile (still valid): Success + .pex present");
-                Check(File.GetLastWriteTimeUtc(pex) > firstWriteUtc, "recompile (still valid): .pex rewritten this run (write-time advanced; no pre-delete)");
+                // Without a .pex from the good compile there is nothing to recompile over; say that rather than
+                // dereference the null and report the whole probe as a throw.
+                if (good.PexPath is null)
+                {
+                    Console.WriteLine("  SKIP  recompile checks: the good compile produced no .pex (see the failed check above)");
+                }
+                else
+                {
+                    var pex = good.PexPath;
+                    var firstWriteUtc = File.GetLastWriteTimeUtc(pex);
+                    var reGood = PapyrusCompile.CompileObject(compiler, "HCGood", imports, outDir);
+                    Check(reGood.Success && File.Exists(pex), "recompile (still valid): Success + .pex present");
+                    Check(File.GetLastWriteTimeUtc(pex) > firstWriteUtc, "recompile (still valid): .pex rewritten this run (write-time advanced; no pre-delete)");
 
-                File.WriteAllText(Path.Combine(work, "HCGood.psc"),
-                    "Scriptname HCGood extends Quest\n\nFunction Foo()\n    @@@ broken edit\nEndFunction\n");
-                var reBad = PapyrusCompile.CompileObject(compiler, "HCGood", imports, outDir);
-                Check(!reBad.Success && reBad.PexPath is null, "recompile (now broken): reports failure, no new .pex");
-                Check(File.Exists(pex), "recompile (now broken): the PRIOR .pex is LEFT INTACT (non-destructive)");
+                    File.WriteAllText(Path.Combine(work, "HCGood.psc"),
+                        "Scriptname HCGood extends Quest\n\nFunction Foo()\n    @@@ broken edit\nEndFunction\n");
+                    var reBad = PapyrusCompile.CompileObject(compiler, "HCGood", imports, outDir);
+                    Check(!reBad.Success && reBad.PexPath is null, "recompile (now broken): reports failure, no new .pex");
+                    Check(File.Exists(pex), "recompile (now broken): the PRIOR .pex is LEFT INTACT (non-destructive)");
+                }
             }
             finally { try { Directory.Delete(work, recursive: true); } catch { /* temp scratch; non-fatal */ } }
         }
--- a/src/housecarl-generator/ProbeTemp.cs
+++ b/src/housecarl-generator/ProbeTemp.cs
@@ -12,8 +12,11 @@ namespace HousecarlGenerator;
 /// spawns inherit the same root.</para>
 ///
 /// <para>The run also stops leaving fixtures behind: the whole root goes at the end, entry by entry so one
-/// file another process still holds open strands only itself. A file that will not delete is reported and the
-/// run's exit code is unaffected — cleanup never turns a green run red.</para>
+/// file another process still holds open strands only itself. A run that is killed — the bridge test's
+/// 20-minute cap, Ctrl-C, a cancelled CI job — never reaches that, so <see cref="Redirect"/> also drops the
+/// roots whose pid is no longer a live process; residue under our own pid that will not go is reported and
+/// the run takes a fresh sibling root rather than work in another run's fixtures. Nothing here throws: a file
+/// that will not delete is reported and the run's exit code is unaffected — cleanup never turns a run red.</para>
 /// </summary>
 public static class ProbeTemp
 {
@@ -25,9 +28,20 @@ public static class ProbeTemp
     {
         if (_root != null) return _root;
 
-        var root = Path.Combine(Path.GetTempPath(), "hc-" + Environment.ProcessId);
-        // A root under this pid can only be residue from a dead process: no live process shares our pid.
-        if (Directory.Exists(root)) Remove(root);
+        var temp = Path.GetTempPath();
+        SweepDeadRoots(temp);
+
+        // A root under this pid can only be residue from a dead process: no live process shares our pid. If any
+        // of it will not go, say so and take a fresh sibling — a run inside fixtures it failed to empty is the
+        // collision this class exists to remove, and silent.
+        var root = Path.Combine(temp, "hc-" + Environment.ProcessId);
+        for (int n = 2; Directory.Exists(root); n++)
+        {
+            var (left, first) = Remove(root);
+            if (left == 0) break;
+            ReportLeft(root, left, first);
+            root = Path.Combine(temp, $"hc-{Environment.ProcessId}-{n}");
+        }
         Directory.CreateDirectory(root);
 
         _tmpWas = Environment.GetEnvironmentVariable("TMP");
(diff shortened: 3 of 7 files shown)


src/housecarl-generator/ProbeTemp.cs
```diff
@@ -0,0 +1,85 @@
...
+        _root = null;
+    }
+
+    /// <summary>Delete a directory's contents and then itself, counting what would not go.</summary>
+    static (int Left, string? FirstError) Remove(string dir)
+    {
+        int left = 0;
+        string? first = null;
+
+        foreach (var entry in Directory.EnumerateFileSystemEntries(dir))
```
Avick3110: `Directory.EnumerateFileSystemEntries(dir)` sits outside the try, so the enumerator's own failure is not caught: only the per-entry delete is. `Cleanup()` runs from `RunAll`'s and `TryDispatch`'s `finally`, so anything thrown here escapes as an unhandled exception and replaces the run's computed exit code and its 127/127 summary with a crash. That contradicts the two promises written directly above it ("reported, not thrown", "cleanup never turns a green run red").

Concrete: `ci-all` spawns `housecarl-mcp.exe` children that HarnessBridgeTests notes can outlive the run; if one of them removes or recreates a directory in the root while the enumerator is walking it, or if the root has already gone (any probe that cleans the temp path itself), the `foreach`'s `MoveNext` throws `DirectoryNotFoundException`/`IOException` and a green run reports a crash. The same call is reached from `Redirect()` line 30, where it would kill the run before the first probe.

Wrapping the whole body of `Remove` (enumeration included) in a try/catch that folds into `left`/`first` keeps the stated contract.

src/housecarl-generator/ProbeTemp.cs
```diff
@@ -0,0 +1,85 @@
...
+    static string? _tmpWas, _tempWas, _tmpdirWas;
+
+    /// <summary>Point this process's temp directory at a root only this process uses, and return it.</summary>
+    public static string Redirect()
+    {
+        if (_root != null) return _root;
+
+        var root = Path.Combine(Path.GetTempPath(), "hc-" + Environment.ProcessId);
+        // A root under this pid can only be residue from a dead process: no live process shares our pid.
+        if (Directory.Exists(root)) Remove(root);
```
Avick3110: `Remove(root)`'s `(Left, FirstError)` is discarded here, and `Directory.CreateDirectory(root)` on the next line succeeds whether or not the residue actually went. So a run can start inside a root it did not empty, and nothing says so.

Concrete: a previous `ci-all` at this pid was killed rather than exiting (HarnessBridgeTests kills it at the 20-minute cap; Ctrl-C; a cancelled CI job), leaving `hc-<pid>` full of `hc-inplace-guard`, `.compile-probe-gen` and friends, some still held open by an orphaned `housecarl-mcp.exe` child. Windows recycles pids quickly, so a later run lands on the same number, `Remove` fails on the held subtrees, and the probes then run against another run's stale fixtures — the exact file-in-use failure this PR exists to remove, except now invisible because the fixture names are no longer shared and nobody is looking.

At minimum print what would not go, the way `Cleanup` does.

src/housecarl-generator/ProbeTemp.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    static string? _root;
+    static string? _tmpWas, _tempWas, _tmpdirWas;
+
+    /// <summary>Point this process's temp directory at a root only this process uses, and return it.</summary>
+    public static string Redirect()
+    {
+        if (_root != null) return _root;
+
+        var root = Path.Combine(Path.GetTempPath(), "hc-" + Environment.ProcessId);
```
Avick3110: Naming the root for the pid makes the leak on an abnormal exit unbounded, where the old scheme's was bounded. Before this change the fixture names were a fixed set of ~50 (`hc-inplace-guard`, `hc-mo2instance-probe`, …) that every run reused, so a killed run left at most that set behind and the next run overwrote it. Now each run picks a fresh name, `Cleanup()` only runs on a normal return, and `Redirect()` only ever removes the root for its own pid — nothing sweeps the others.

Concrete: HarnessBridgeTests kills `ci-all` at its 20-minute cap, and Ctrl-C during a local run does the same; each such run strands a full `hc-<pid>` fixture tree in `%TEMP%` permanently, and they accumulate run after run. A cheap fix is to also drop `hc-*` roots whose pid is not a live process at `Redirect()` time, or to register the cleanup on `AppDomain.CurrentDomain.ProcessExit`. Worth at least noting in the class doc, which currently says the root is "removed when the run ends" without qualification.

src/housecarl-generator/BsaProbe.cs
```diff
@@ -34,7 +34,7 @@ public static int Run(string[] args)
         if (!File.Exists(bsarch)) { Console.WriteLine($"  SKIP  no BSArch at '{bsarch}' (pass its path as arg 1, or set HOUSECARL_BSARCH)"); return 0; }
         if (!File.Exists(bsa)) { Console.WriteLine($"  SKIP  no test archive at '{bsa}' (pass one as arg 2, or set HOUSECARL_TEST_BSA)"); return 0; }
 
-        var work = Path.Combine(Environment.CurrentDirectory, ".bsa-probe");
+        var work = Path.Combine(Path.GetTempPath(), ".bsa-probe");
```
Avick3110, line 37: Moving `work` under the temp path leaves the comment on the matching cleanup at line 88 saying `/* in-dir scratch; non-fatal */`, which is no longer true — the directory is no longer in the checkout. The other two probes moved in the same commit had that exact comment updated to `temp scratch`; this one was missed, so the file now describes a location it does not use. Change line 88's comment to match.

src/housecarl-generator/CompileProbe.cs
```diff
@@ -95,7 +95,7 @@ public static int Run(string[] args)
                 Check(!reBad.Success && reBad.PexPath is null, "recompile (now broken): reports failure, no new .pex");
```
Avick3110: About the null `PexPath` the PR body names as the second collision's symptom: the trigger is gone, but the deref that turned it into a crash is still unguarded. Line 86, `var pex = good.PexPath!;`, runs even when the check on line 74 has already recorded that `good.PexPath` is null, so `File.GetLastWriteTimeUtc(pex)` on line 87 throws and the probe reports `THREW: ArgumentNullException` instead of the two named failed checks that say what actually broke. Any environmental failure of the good compile, not just a deleted scratch dir, lands there. Returning early (or skipping the recompile arm) when `good.PexPath is null` keeps the diagnosis in the probe's own words. Pre-existing and outside this PR's subject, so take it or leave it.

src/housecarl-generator/ProbeTemp.cs
Avick3110: Real and folded in 546914c0: the enumeration is inside the try with the deletes now, a root that goes while being walked returns clean rather than counting a phantom entry, and Remove never throws.

src/housecarl-generator/ProbeTemp.cs
Avick3110: Real and folded in 546914c0: Redirect now reads what Remove reported, prints the same one-line "temp cleanup left N entries" as Cleanup, and takes a fresh sibling root (hc-<pid>-2) rather than starting inside residue it failed to empty — the never-throws contract holds and the run is not silently degraded.

src/housecarl-generator/ProbeTemp.cs
Avick3110: Real and folded in 546914c0: Redirect now sweeps the hc-<pid> roots whose pid is no longer a live process before taking its own, so a killed run stays bounded, and the class doc says that instead of the unqualified "removed when the run ends". Verified by planting hc-999999 (not a possible live pid) and watching a solo ci-all remove it.

src/housecarl-generator/BsaProbe.cs
Avick3110: Right, fixed in 92408f46: the comment now reads temp scratch, matching the other two probes moved in the same commit.

src/housecarl-generator/CompileProbe.cs
Avick3110: Taken, in 0d9c620b: the recompile arm is skipped with a SKIP line when good.PexPath is null, so the probe reports its own failed check instead of THREW: ArgumentNullException.

src/housecarl-generator/ProbeTemp.cs
```diff
@@ -0,0 +1,149 @@
...
+        if (name.Length <= 3) return false;
+        var text = name.AsSpan(3);
+        var dash = text.IndexOf('-');
+        if (dash >= 0) text = text[..dash];
+        return int.TryParse(text, out pid);
+    }
+
+    static bool IsLive(int pid)
+    {
+        try { using var p = System.Diagnostics.Process.GetProcessById(pid); return !p.HasExited; }
```
Avick3110: **`IsLive` treats "cannot inspect the process" as "dead", so the sweep can delete a still-running run's root.**

Two `ci-all` runs on the same Windows user, one launched from an elevated terminal (pid 4321) and one not. The non-elevated run calls `IsLive(4321)`: `Process.GetProcessById` succeeds (the enumeration behind it needs no handle), but `HasExited` cannot `OpenProcess` a higher-integrity process — the handle comes back invalid, `HasExited` returns `true` — so `IsLive` returns `false` and `SweepDeadRoots` calls `Remove` on the live run's `hc-4321` root, deleting every fixture it is not holding open at that instant. That is the mid-run scratch deletion #570 is about, arriving from the other direction. The blanket `catch { return false; }` does the same for any transient permission or enumeration error.

The default is the wrong way round: only a definite "no such process" (`ArgumentException`) is evidence of death; everything else — an exception, or a `HasExited` reached through an unopenable handle — should count as live and leave the root alone. Worst case that is bounded residue the next run picks up, which is the trade this class already makes elsewhere.

src/housecarl-generator/ProbeTemp.cs
```diff
@@ -0,0 +1,149 @@
...
+                try
+                {
+                    if (Directory.Exists(entry)) Directory.Delete(entry, recursive: true);
+                    else File.Delete(entry);
+                }
+                catch (Exception ex)
+                {
+                    left++;
+                    first ??= ex.Message;
+                }
```
Avick3110: **A concurrent deleter's race is counted as residue.**

Two `ci-all` runs start together — the case this PR is for — and both sweep the same dead `hc-<pid>` root. Run A enumerates the entries, run B deletes one, and run A's `Directory.Delete(entry, recursive: true)` throws `DirectoryNotFoundException`, counted as `left++`. Run A then prints `(temp sweep left 1 entry from an earlier run ...)` naming a root that is in fact gone.

The same race on the run's **own** root in `Redirect`'s loop (line 40) is worse than cosmetic: `left > 0` makes the run report phantom residue and take `hc-<pid>-2` even though `hc-<pid>` was cleared.

An entry that is already gone is the outcome the delete wanted:

```suggestion
                catch (Exception ex) when (ex is not (DirectoryNotFoundException or FileNotFoundException))
                {
                    left++;
                    first ??= ex.Message;
                }
```

src/housecarl-generator/ProbeTemp.cs
```diff
@@ -0,0 +1,149 @@
...
+        Directory.CreateDirectory(root);
+
+        _tmpWas = Environment.GetEnvironmentVariable("TMP");
+        _tempWas = Environment.GetEnvironmentVariable("TEMP");
+        _tmpdirWas = Environment.GetEnvironmentVariable("TMPDIR");
+        Environment.SetEnvironmentVariable("TMP", root);
+        Environment.SetEnvironmentVariable("TEMP", root);
+        Environment.SetEnvironmentVariable("TMPDIR", root);   // what GetTempPath reads off Windows
+
+        _root = root;
```
Avick3110: **A partial `Redirect` leaves the process redirected with nothing to clean it up.**

`_root` is assigned last, after the three `SetEnvironmentVariable` calls, and both callers run `ProbeTemp.Redirect()` *outside* the `try` whose `finally` calls `Cleanup()`. So if `Directory.CreateDirectory(root)` fails (a read-only or full temp volume, or a plain file already named `hc-<pid>` — `Directory.Exists` is false for that, so the loop above does not catch it), or a `SetEnvironmentVariable` throws after the first has taken effect, then: `TMP`/`TEMP` already point at the new root, `_root` is still null so `Cleanup()` returns immediately, the root is never deleted, and `ci-all` exits on a raw stack trace before any probe runs rather than the one plain sentence the project asks for.

Assigning `_root = root` right after `Directory.CreateDirectory` and moving `Redirect()` inside the `try` in `RunAll`/`TryDispatch` closes both halves.

src/housecarl-generator/ProbeTemp.cs
Avick3110: Folded in 730bca34. Confirmed the mechanism here first: on .NET 9 `Process.GetProcessById(4)` (the System process) succeeds and `HasExited` then throws `Win32Exception: Access is denied`, which the blanket catch turned into "dead" — so the outcome is yours exactly, by a throw rather than a `true`. `IsLive` now returns false only on `ArgumentException` (no such process) and true on anything else; `HasExited` is gone entirely, since reaching it needs a handle we may not get and a process that has exited already throws `ArgumentException` out of `GetProcessById`. New unit test `ProbeTempSweepTests` pins all three cases — free pid dead, own process live, running-but-unopenable process live. The third fails before the fix and passes after (verified locally by reverting the fix and rerunning).

src/housecarl-generator/ProbeTemp.cs
Avick3110: Folded in e5d534fb, your suggestion as written: the inner catch now filters out `DirectoryNotFoundException` and `FileNotFoundException`, so an entry that is already gone is not counted. That covers both call sites you name — `SweepDeadRoots` no longer prints a root that is in fact gone, and `Redirect`'s own-root loop no longer abandons a cleared `hc-<pid>` for `hc-<pid>-2` on a phantom count. The outer catch and the final `Directory.Delete` already handled the same case. No test: the race needs an entry to vanish between `EnumerateFileSystemEntries` returning it and the delete reaching it, and I could not produce that deterministically without a fault-injection seam, which is not worth adding here.

src/housecarl-generator/ProbeTemp.cs
Avick3110: Folded in d0a42201, both halves as you prescribe: `_root = root` now sits immediately after `Directory.CreateDirectory(root)`, and `ProbeTemp.Redirect()` moved inside the `try` in both `RunAll` and `TryDispatch`, so the `finally` that calls `Cleanup()` covers it. A `SetEnvironmentVariable` that throws after the first has taken now leaves a root `Cleanup` still knows to remove. No test: forcing a partial redirect needs fault injection on `CreateDirectory` or `SetEnvironmentVariable`, and driving `Redirect()` from a test repoints the test process's own temp directory and sweeps the real temp path, which is the hazard this class exists to remove. Left alone: the raw stack trace half of your note. `CreateDirectory` failing still throws out of `Program.cs` unwrapped, so `ci-all` dies on a trace rather than one plain sentence. That is a separate change to how the runner reports a failed start, and I would rather you rule on it than fold it in silently here.

PR #587, opened 2026-09-06
Bring the merge_plugins tool description under the 2,048-character cap

Avick3110:
`housecarl_merge_plugins` published a 3,363-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the renumber rule, the cross-donor conflict grammar, the asset carry, the refusal list and the save warning were all sitting in text the model never saw. Nothing is deleted: each paragraph moves into the description of the parameter it is about — the donor rules to `plugins`, the new-name consequences to `output` — and where the parameter already said the same thing weakly, the description's fuller version replaces it rather than being pasted in beside it. The description keeps what belongs to no single parameter: the one-sentence purpose, the post-merge MO2 procedure with its BSA caveat, the header rules (ESM/Author dropped, ESL carried or not, and whether `compact_plugin` can still make it light), and one new line naming which parameter carries which grammar. `merge_plugins`'s line comes out of `StillOversized`.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_merge_plugins` description | 3,363 | **1,742** |
| Parameter descriptions (`plugins` + `output` + `patch_name`) | 688 | 2,662 |
| Both together | 4,051 | 4,404 |

The description shed 1,931 characters of moved text and gained a 304-character parameter roster it did not carry before; the parameters gained 1,974. The 43-character difference is the fold wording in the three rows below that say so.

## Harvest

|---|---|
| "MERGE one or more ACTIVE plugins into ONE NEW plugin — a RECORDS operation (the zMerge/'Merge Plugins' job) … the donor FILES and their mods are NEVER touched (new-file lane only, no in-place)." | kept in description |
| "TO RENAME a plugin, pass ONE donor: with nothing to combine the merge IS a rename … facegen/voice/seq carried to the new name." | `plugins`, folded onto its weaker "A SINGLE donor renames it into output=." — that sentence keeps the pointer to `output=` and the description's version supplies the grammar behind it |
| "RENUMBER is collision-first (zMerge's default): the donor EARLIEST in the load order keeps its FormID object ids … (all records necessarily move to the new plugin's identity)." | `plugins` (under the "houseCARL uses LOAD order for id priority" sentence already there, which is now its topic line) |
| "Cross-donor conflicts on the SAME record resolve to the LOAD-ORDER WINNER and are each REPORTED; a losing donor's nested children … are GRAFTED into the winner's copy — so merging a mod WITH its patches is the intended use." | `plugins` |
| "ASSETS follow the renumber: every donor NPC's facegen and every voiced line are carried into the new plugin-name folders … and a .seq is refreshed when any donor shipped one." | `output` — the plugin NAME the paths embed is the one this parameter sets, so it reads "ASSETS follow the renumber into this name" |
| "THE SAFETY (Q3): plugins OUTSIDE the merge that reference or override donor records are WARNED and NAMED, never refused … the remedy is to include those patches in the merge set or re-point them before disabling the donors." | `plugins` — the remedy is a value of this parameter |
| "Refuses loud + writes nothing on: a donor unparseable / not on disk; … a dangling donor-internal reference (a donor referencing a FormID no donor defines); a declared master not active." | `plugins` |
| …the same sentence's "a donor not active" clause | `plugins` (already said there as "each must be active in your load order"; that sentence was the weaker one — it stated the requirement without the refusal — so it now carries the description's force: "each must be active in your load order: a donor not active is refused loud and nothing is written") |
| …the same sentence's "an output name already in the load order" clause | `output` (already said there as "must NOT already exist in the load order"; that sentence was the weaker one — it stated the requirement without the refusal — so it now carries the description's force: "an output name already there is refused loud and nothing is written") |
| "AFTER: review the merged plugin in xEdit, enable its mod folder in MO2, then deactivate the donor PLUGINS (right pane) but KEEP the donor MOD FOLDERS enabled (left pane) — merge carries only the FormID-keyed files the rename breaks … loads from the donor folders." | kept in description |
| "Caveat: a donor .bsa stops auto-loading once its same-named plugin is inactive — extract it into the mod folder (`housecarl_bsa_extract`) or load it via a same-named dummy plugin (`housecarl_create_plugin`)." | kept in description (it is a caveat on the AFTER step above it, and carries two cross-tool pointers) |
| "Existing SAVES that depend on the donors will NOT survive (the records now live under a different plugin name, and any id that had to be renumbered moved with it) — best for a new game." | `output` — "a different plugin name" is this parameter's value, so it reads "under this plugin name" |
| "A donor's HEADER mostly does not come along: master (ESM) status and Author/Description are always dropped, and the report names each one it actually dropped." | kept in description |
| "Light (ESL) status IS carried when every donor was light and every merged object id fits the light window 0x800–0xFFF; otherwise it is dropped and the report says which of the two reasons it was, and whether `housecarl_compact_plugin` can still make it light … unless the merged plugin defines more than 2048 records, which no light plugin holds." | kept in description |

`patch_name` received nothing: no sentence of the description was about the mod folder, and its own description already covers the default names.

The description gained one thing it did not carry before: a line naming which parameter holds which grammar (`plugins=` the donor set, `output=` the new filename, `patch_name=` the new mod folder). With the grammar moved out, that roster is the description's only index of the parameters, and a caller who reads only the 2,048 characters the client shows needs to know where the rest is.

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`: the `ToolNames.MergePlugins` line comes out of `StillOversized`, and nothing else in the file changes. Run with that line gone and the description unchanged, `EveryPublishedDescriptionFitsTheClientsCut` failed on `housecarl_merge_plugins` at 3,363 characters; after the change the class is green and prints 1,742.

## What was run

```
    1620 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. No CHANGELOG line: the records PR's entry covers the wave.


src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
```diff
@@ -36,7 +36,6 @@ public sealed class PublishedDescriptionBoundTests
         ToolNames.Copy,
         ToolNames.Create,
         ToolNames.Forward,
-        ToolNames.MergePlugins,
```
Avick3110, line 39: The changelog's count goes stale with this line. `plugin/CHANGELOG.md` line 23 (the records entry under `## Unreleased`, already on `main`) says "Eleven other tools are still over the cap" — that eleven was `StillOversized`'s eleven entries, including `merge_plugins`. Deleting this line makes it ten, and the PR ships no changelog change, so the released Unreleased section states a number the code contradicts. Concretely: a user reading the shipped changelog after this merges is told eleven tools remain oversized while the test names ten.

The entry's own rule ("point at the bound rather than restating it, so the two cannot drift apart") argues against just decrementing to "Ten" — every PR in this wave would have to edit it again. Better to reword that sentence to point at `PublishedDescriptionBoundTests.StillOversized` for the current set instead of carrying a count.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -106,39 +106,26 @@ public static string CompactPlugin(
...
...t order does not matter: houseCARL uses LOAD order for id priority and conflict resolution. RENUMBER is collision-first (zMerge's default): the donor EARLIEST in the load order keeps its FormID object ids; later donors renumber ids already taken, and ANY donor's ids below the 0x800 floor renumber too (all records necessarily move to the new plugin's identity). Cross-donor conflicts on the SAME record resolve to the LOAD-ORDER WINNER and are each REPORTED; a losing donor's nested children the winner doesn't re-list (a base mod's dialogue lines under a patched topic; placed refs under a patched cell) are GRAFTED into the winner's copy — so merging a mod WITH its patches is the intended use. THE SAFETY (Q3): plugins OUTSIDE the merge that reference or override donor records are WARNED and NAMED, never refused — the donors stay active until you swap in MO2, so nothing breaks at write time; the remedy is to include those patches in the merge set or re-point them before disabling the donors. Refuses loud + writes nothing on: a donor not active / unparseable / not on disk; a dangling donor-internal reference (a donor referencing a FormID no donor defines); a declared master not active.")]
```
Avick3110: **low** — `plugins` now states the active-donor rule twice: "each must be active in your load order" in the first sentence, and "a donor not active" in the refusal list at the end. That is the same requirement/refusal overlap the PR folded on `output` into one sentence ("must NOT already exist in the load order: an output name already there is refused loud and nothing is written"), so the same fact is handled two ways in one change, and the harvest table does not list this overlap at all. Fold it the same way here: either drop "each must be active in your load order" and let the refusal list carry it, or give the opening sentence the refusal's force and drop "a donor not active" from the list.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: Refuting: by design no PR of this wave except the first and the last touches the CHANGELOG, so parallel PRs cannot conflict on it, and the last PR of the wave (forward) rewrites that line for the whole wave and will point at the test's set rather than a number, as you suggest.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in fc3faa64: `plugins` now carries the active-donor rule once, the way `output` does — "each must be active in your load order: a donor not active is refused loud and nothing is written" — and the refusal list drops "not active"; the harvest table has the row, and the published description is unchanged at 1,742.

PR #588, opened 2026-09-06
Bring the place tool description under the 2,048-character cap

Avick3110:
`housecarl_place` published a 2,976-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the source grammar — the per-member rule, the BSA-entry shorthand, the rename, and the refusal an omitted `source=` raises when several providers contend — sat in a part of the text the model never saw. Nothing is deleted: each paragraph moves into the description of the parameter it is about, and where the parameter already said the same thing the description's copy is dropped as a duplicate rather than pasted in twice. The description keeps the purpose sentence, the `housecarl_asset_status` pointer, the composition sentence, the set-valued rule, one line per axis naming its parameters, and the two facts that belong to the whole call — the write is crash-atomic, and a placement does not win until the mod is enabled and sorted. `housecarl_place`'s line comes out of `StillOversized`, so the bound now holds it.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_place` description | 2,976 | **1,146** |
| Top-level parameter descriptions | 2,609 | 3,405 |
| All schema descriptions (parameters + `assets[]` members) | 4,471 | 5,831 |

The description shed 1,830 characters and the schema gained 1,360. The 470-character difference is text the description was repeating from a parameter that already carried it: the LANE paragraph (both halves are on `patch=` and `into=`), the `assets=` member shape (on `assets=`), the formid/path exclusivity and the both-FaceGen-files rule (on the `formid`, `kind` and `path` members), and the `*winner` / NAME ALONE token grammar (on `source_provider=`) — less the text the review fold below added back on both sides.

Both columns of the third row are re-measured by one sweep (the top-level parameter descriptions plus the five `PlaceTarget` member descriptions); the pair posted earlier, 4,560 and 5,839, came from a sweep counting 89 characters this one does not attribute to the schema, and the row's delta is unchanged by the correction.

## Harvest

|---|---|
| "Place chosen copies of files — ANY Data-relative file … so the copy YOU pick wins the virtual file system." | kept in description |
| "The WRITE counterpart to `housecarl_asset_status` (which reports which copy currently wins)." | kept in description |
| "ONE surface: WHERE the bytes land (assets=) x WHOSE copy to read (the SOURCE pole) x WHICH folder it goes in (the LANE) x how it reads back (TRANSPORT)." | kept in description, verbatim and above the four axis lines — the axis lines enumerate the axes, this says a single call composes them |
| "One file is a set of one — the same call shape places forty." | kept in description |
| "assets= is the SET OF DESTINATIONS, each member { path? \| formid?, kind?, source?, source_provider? }." | `assets` (already said there, as the typed member roster) |
| "Give a member EITHER path (a Data-relative destination) OR formid (an NPC's 'XXXXXX:Plugin.esp', whose FaceGen paths houseCARL computes)" | `assets[].formid` and `assets[].path` (already said on both — "Provide this OR path" / "Provide this OR formid", with the FaceGen computation on `formid`) |
| "with a formid and NO kind BOTH the head mesh and the face tint are placed; kind='mesh' or 'tint' narrows it to one." | `kind` and `assets[].kind` (already said on both) |
| "All members land in ONE reviewable mod folder." | `assets` — its "all placed into ONE mod folder" gains the description's "reviewable" |
| "SOURCE — whose copy to read. source_provider= names it: `*winner` (the sigil is part of the token) for the current VFS winner, or a provider's NAME ALONE — a mod folder, 'overwrite', 'Data', or a BSA filename — matched exactly." | `source_provider` — already said there except "(the sigil is part of the token)", which moves in |
| "Naming a mod folder MO2 is NOT currently loading reads that mod's own copy off disk … For a mod MO2 IS loading, nothing changes …" (`WriteSentences.PlaceSourceNameReachesUnticked`) | `source_provider` — it was already on the per-member `assets[].source_provider` but not on the call-level pole, whose "a mod folder of that same name is not consulted" sentence presupposes it. The pole's own "a file inside an active mod's archive is listed under the ARCHIVE's name" sentence said the constant's tail over again, so it is trimmed to the part that is new: the archive MO2 loads no plugin for |
| "Set it once for the whole set, or per member." | kept in description (the SOURCE axis line: "once for the whole set or per member") |
| "source= is PER MEMBER (a source names ONE file, and a set of destinations is many)" | `assets[].source` |
| "a DATA-RELATIVE path resolved through the VFS, a full loose file path, '<archive.bsa path>\|<entry inside>', or just a '.bsa' path (the entry is taken to be the destination — a quick way to pull ONE file out of a BSA as a loose override)" | `assets[].source` — its shorter list had no BSA-entry gloss, so the description's fuller spelling replaces it |
| "A source path DIFFERENT from the destination is a RENAME: the bytes of one file land under another file's name, which is how a baked FaceGen head is carried onto a different NPC's FormID path." | `assets[].source` — replacing its weaker "different from the destination = a rename", which carried no example |
| "With no source=, the DESTINATION path is resolved instead: the sole provider, or the one source_provider= names, REFUSING (and listing the providers) when several contend and none was named — it will not guess which is correct." | `assets[].source` — replacing its weaker "Omit to resolve the destination path through the VFS", which named no refusal. The VFS qualifier stays on the sole-provider clause, since the `source_provider=` branch can read an off-order mod's copy off disk |
| "LANE — patch= names the NEW mod folder (default 'houseCARL_Assets'; auto-suffixed if taken)" | `patch` (already said there) |
| "into= adds to an EXISTING houseCARL patch folder instead, so calls accumulate in one mod." | `into` (already said there, as "accumulate across calls") |
| "A malformed member (bad FormID, bad kind, neither or both of formid and path) refuses the WHOLE call with per-member reasons and places nothing; a source that is ambiguous, absent or unreadable is a PER-MEMBER error and the rest still place." | `assets`, beside the undeclared-member refusal it already carried, so the three refusal shapes read together — and with `MapTarget`'s fourth whole-call case named: a `formid` member with no `kind` whose `source=` is not a full '.bsa' path |
| "The write is crash-atomic and originals are never touched." | kept in description |
| "IMPORTANT (and reported back): the placed copies do NOT win on write — you must ENABLE the new mod in MO2 and SORT it above the current winner." | kept in description |

Every parameter the description now names has a description of its own, and every parameter with a description is named on an axis line — `assets=` and `kind=` on DESTINATION, `source_provider=` and the members' `source=` on SOURCE, `patch=` and `into=` on LANE, `format=` and `max_chars=` on TRANSPORT.

Two code comments in `src/housecarl-generator/PlaceProbe.cs` pointed at the tool description for grammar this change moves out of it, and are repointed at the parameter that now carries it: the no-`source=` shape (line 635) to the member's `source=` description, and the archive-filename pole (line 1338) to `source_provider=`.

## Review fold

Five of the six review findings are folded, each in its own commit:

- cf12188 restores the composition sentence to the description, above the axis lines.
- 1c20c58 names the source-shape whole-call refusal in `assets=`: `MapTarget` refuses the whole call for a `formid` member with no `kind` whose `source=` is a loose file, a BSA entry, or a relative path, so the malformed-member list was reading as exhaustive without it.
- 38bcbb4 scopes the VFS qualifier on `assets[].source` to the sole-provider clause, since the `source_provider=` branch reads an off-order mod's copy off disk.
- 5f3bb44 says on the call-level `source_provider=` that the pole is withheld from a member whose own `source=` is on disk, matching what `MapTarget` computes through `SourceTakesAProvider` and what the per-member description already said.
- eeb92bf trims the pole's archive-name sentence to the one thing the inserted constant does not already say.

The sixth, on the changelog's "Eleven other tools" count, is refuted: by design only the first and last PR of the wave touch the CHANGELOG, and the last one rewrites that line to point at the test's set rather than number it.

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs` loses its `ToolNames.Place` line from `StillOversized`, and nothing else.

- With the line gone and the old description in place, `EveryPublishedDescriptionFitsTheClientsCut(tool: "housecarl_place")` failed: "housecarl_place's published description is 2976 characters; Claude Code shows the first 2048 and drops the rest." 1 failed, 31 passed.
- With the description moved and the fold applied, the class is green and the theory prints `housecarl_place: 1146 characters (bound 2048)`.

## What was run

```
    1620 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. No CHANGELOG line: the records PR's entry covers the wave and the last PR of it generalises the entry.


src/housecarl-mcp/PlaceTools.cs
```diff
@@ -22,36 +22,22 @@ public static class PlaceTools
...
+        [Description("SELECT: the destinations, all placed into ONE reviewable mod folder. Each: { formid?: 'XXXXXX:Plugin.esp', kind?: 'mesh'|'tint' (omit with formid to place BOTH FaceGen files), path?: 'meshes/...', source?: '<loose path>' | '<archive.bsa>|<entry>' | '<archive.bsa>' | '<Data-relative path>', source_provider?: 'SomeMod' | 'X - Textures.bsa' | '" + AssetSourceChoice.WinnerToken + "' } — or \"@<absolute path>\" to read that SAME array from a JSON file. Set-valued at every size — one destination is a set of one. A member the shape does not declare is refused BY NAME at its element, never silently dropped. A malformed member — a bad FormID, a bad kind, or neither or both of formid and path — refuses the WHOLE call with per-member reasons and places nothing; a source that is ambiguous, absent or unreadable is a PER-MEMBER error and the rest still place. Each member's own description says what it takes.")]
```
Avick3110: The refusal partition this sentence promotes into `assets=` is not the one the code implements, and that matters more now than it did before: reconstructing the old description gives 2,976 characters with this paragraph starting around offset 2,450, so it sat past the 2,048 cut and the model never read it. Delivered for the first time, its incompleteness becomes load-bearing.

`MapTarget` also returns a whole-call `error` for a source whose SHAPE is wrong — a `formid` member with no `kind` whose `source=` is a loose file, a BSA entry, or a relative path: "with formid and no kind (placing BOTH mesh and tint), an explicit source= must be a FULL '.bsa' path". That error goes into `problems`, and `PlaceTargets` then refuses the whole call and places nothing. So a source problem is not always per-member, and the malformed-member list here reads as exhaustive without covering it.

Scenario: 40 destinations, 39 fine, one a `formid` member without `kind` carrying a loose-file `source=`. A caller reading this expects 39 placed and one per-member failure; nothing is placed.

Adding the source-shape case to the malformed-member list, or narrowing "a source that is ambiguous, absent or unreadable" so it does not read as covering every source problem, would close it.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -333,7 +319,7 @@ public sealed record PlaceTarget
...
+    [JsonPropertyName("source"), Description("The copy to place, for THIS destination — a source names ONE file and a set of destinations is many, so it is PER MEMBER: a DATA-RELATIVE path resolved through the VFS, a full loose file path, '<archive.bsa path>|<entry inside>', or just a '.bsa' path (the entry is taken to be the destination — a quick way to pull ONE file out of a BSA as a loose override). A source path DIFFERENT from the destination is a RENAME: the bytes of one file land under another file's name, which is how a baked FaceGen head is carried onto a different NPC's FormID path. With no source=, the DESTINATION path is resolved through the VFS instead: the sole provider, or the one source_provider= names, REFUSING (and listing the providers) when several contend and none was named — it will not guess which is correct. With formid and no kind, an explicit source must be a FULLY-QUALIFIED '.bsa' path (a relative one is a Data-relative asset path, and one path cannot serve both slots).")]
```
Avick3110: "through the VFS" is new in this clause and changes what it claims. The old tool description said only "the DESTINATION path is resolved instead: the sole provider, or the one source_provider= names", and the old member description's "through the VFS" attached to the bare omit case with no mention of a pole. Merged, the qualifier now governs the `source_provider=` branch as well — and that branch is precisely the one that is NOT VFS resolution: `WriteSentences.PlaceSourceNameReachesUnticked`, which this PR adds to both `source_provider` descriptions, says naming a mod MO2 is not currently loading "reads that mod's own copy off disk".

Scenario: a caller omits `source=` and passes `source_provider='SomeDisabledMod'`, reads this sentence, and concludes the name has to be one the VFS currently serves — the exact case the off-order lane was built for, talked out of using it by the description.

Dropping "through the VFS" here (as the old description had it), or attaching it only to the sole-provider clause, restores the accurate reading.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -22,36 +22,22 @@ public static class PlaceTools
...
+        [Description("SOURCE: whose copy to read, for EVERY member that does not name its own. " + AssetSourceChoice.WinnerToken + " (the sigil is part of the token) for whichever copy currently wins the VFS, or the provider's NAME ALONE — a mod folder, 'overwrite', 'Data', or a BSA filename like 'X - Textures.bsa' — matched exactly, without " + ToolNames.AssetStatus + "'s ' (loose)' / ' (BSA)' annotation. A bare name ALWAYS means a provider of that name. " + WriteSentences.PlaceSourceNameReachesUnticked + " Note that a file inside an active mod's archive is listed (and reached) under the ARCHIVE's name, not the mod's — and an archive MO2 loads no plugin for is listed under neither, so it is reachable only as an on-disk source= path. A name the active order already provides files under is answered by the active order, so a mod folder of that same name is not consulted. Omitted = the sole provider, refused if more than one contends.")]
```
Avick3110: The opening clause "for EVERY member that does not name its own" is carried over unchanged, but this PR makes this description the authoritative home for the SOURCE grammar, so the overstatement is worth fixing while the line is being edited.

`MapTarget` computes `setApplies = LoadOrderService.SourceTakesAProvider(src)` and then `prov = ownProv ?? (setApplies ? NullIfBlank(setProvider) : null)`. A member whose own `source=` is an on-disk path has the call-level pole WITHHELD — not refused, withheld — and the caller learns of it only afterwards, on that member's result row ("note: set-level source_provider not applied: source is one exact file").

Scenario: 20 members, `source_provider='SomeMod'` set once for the set, three of them carrying absolute loose-file `source=` paths. The caller expects the pole to govern all 20; it governs 17, and nothing says otherwise until the write has already happened.

The per-member `assets[].source_provider` description does state the rule ("Not valid with an on-disk source"), so a matching one-clause caveat here would make the two agree.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -22,36 +22,22 @@ public static class PlaceTools
...
-         "LANE — patch= names the NEW mod folder (default 'houseCARL_Assets'; auto-suffixed if taken); into= adds to an " +
-         "EXISTING houseCARL patch folder instead, so calls accumulate in one mod.\n\n" +
-         "A malformed member (bad FormID, bad kind, neither or both of formid and path) refuses the WHOLE call with " +
-         "per-member reasons and places nothing; a source that is ambiguous, absent or unreadable is a PER-MEMBER error " +
-         "and the rest still place. The write is crash-atomic and originals are never touched. IMPORTANT (and reported " +
-         "back): the placed copies do NOT win on write — you must ENABLE the new mod in MO2 and SORT it above the " +
-         "current winner.")]
+         "counterpart to " + ToolNames.AssetStatus + " (which reports which copy currently wins). One file is a set of " +
+         "one — the same call shape places forty.\n\n" +
+         "Each axis's grammar is on its own parameters:\n" +
```
Avick3110, line 28: This is the one sentence of the old description that now lands nowhere. I reconstructed the base description (2,976 characters, confirmed) and checked every sentence of it against every description on the branch; all of them have a home except `ONE surface: WHERE the bytes land (assets=) x WHOSE copy to read (the SOURCE pole) x WHICH folder it goes in (the LANE) x how it reads back (TRANSPORT).` The harvest table records it as "kept in description, as the four axis lines that replace it", but the axis lines only enumerate the axes -- they never say that a single call composes all four, which is the claim that sentence carries and the reason `assets=` x source x lane is one call rather than four tools. #583's ruling has the description keeping "the axes sentence ... and the composition rule", and the merged `records` pattern keeps both: `ONE read surface: which records (SELECT) x whose version (SOURCE) x what shape of answer (PROJECT) compose in a single call.` sits above its axis lines, with a `COMPOSITION:` paragraph besides. `apply`, `create`, `forward` and `remove` all still carry `ONE surface:` too, so place would be the only write tool without it. The description measures 993 characters, so restoring one sentence above the axis lines costs nothing against the 2,048 bound.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Inserting `WriteSentences.PlaceSourceNameReachesUnticked` here makes this parameter state one fact twice, back to back. The constant ends `For a mod MO2 IS loading, nothing changes: its loose files are reached by the mod's name and a file inside its archive by that archive's filename, as before.` and the very next sentence in this description opens `Note that a file inside an active mod's archive is listed (and reached) under the ARCHIVE's name, not the mod's`. That is the same rule in two spellings in adjacent sentences, and the PR body's own principle is that where the parameter already said the same thing the copy is dropped rather than pasted in twice. Only the tail of the `Note that` sentence -- the archive MO2 loads no plugin for, reachable only as an on-disk `source=` path -- is new information. Either trim it to that tail, or move the constant down to sit directly before `A name the active order already provides files under is answered by the active order...`, which is the sentence the PR says presupposes it; the per-member `source_provider` two hundred lines down keeps the constant without this collision, so the duplication is specific to this pole.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
```diff
@@ -39,7 +39,6 @@ public sealed class PublishedDescriptionBoundTests
         ToolNames.MergePlugins,
         ToolNames.NifInspect,
         ToolNames.NifSet,
```
Avick3110, line 41: Dropping `ToolNames.Place` from this set takes the set to ten, which falsifies a sentence already shipped in the changelog: the merged `housecarl_records` entry under `## Unreleased` says "Eleven other tools are still over the cap: a test on the published tool list holds the bound and names each of them". I agree no changelog line belongs in this PR, so this is not a request for one -- but that count is a restatement of what this set holds rather than a pointer at it, which is the defect the changelog's own writing guide names, and every remaining PR of the wave will make it wrong again. Worth deciding now whether the wave fixes the count as each tool lands or repoints the sentence at the test instead of numbering it.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in 1c20c58 — the assets= refusal list now names the fourth whole-call case, a formid member with no kind whose source= is not a full '.bsa' path.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in 38bcbb4 — the VFS qualifier now sits on the sole-provider clause only, since the source_provider= branch can read an off-order mod's copy off disk.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in 5f3bb44 — the call-level pole now says it is withheld from a member whose own source= is on disk, matching the per-member wording and what MapTarget computes.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in cf12188 — the composition sentence is back in the description, verbatim and above the axis lines.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in eeb92bf — the Note sentence is trimmed to its new tail, the archive MO2 loads no plugin for, and the constant carries the rest.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: Not folded: by design only the first and last PR of the wave touch the CHANGELOG, and the last one rewrites that line to point at the test's set instead of numbering it, so the count is stale only until the wave ends.

PR #589, opened 2026-09-06
Bring the copy tool description under the 2,048-character cap

Avick3110:
`housecarl_copy` published a 5,243-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the source-universe rules, the internalize rule, the seed shapes, the exclusion consequences and the two destinations were all being described in a part of the text the model never saw. Nothing is deleted here: each paragraph moves into the description of the parameter it is about, and where the parameter already said the same thing the two are stated once rather than pasted in twice. The description keeps the purpose sentence and the whole-record-duplicate mechanism, one line per axis naming its parameters, the pointer to `housecarl_place` for the files that go with the copied records, and the never-touches-originals guarantee. `housecarl_copy`'s line comes out of `StillOversized`, and the copy parser guard's comment — which said the pins were "a sentence the Description promises" — now names both kinds it holds: the sentences the parameter descriptions promise, and the parse tolerances only a refusal or a default states.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_copy` description | 5,243 | **1,104** |
| Top-level parameter descriptions | 1,141 | 5,158 |
| All schema descriptions (parameters + nested DTO members) | 1,141 | 5,158 |

`housecarl_copy` has no object-valued parameters, so the second and third rows are the same number.

The description shed 4,139 characters and the schema gained 4,017. Two things account for the 122-character gap. Text both copies carried — the ordered/first-hit-wins source rule and its two element forms, the record-link seed shape, the unset-seed clear, the `'Type:stop'`/`'Type:refuse'` spellings, the blank-`exclude_types`-entry refusal, and the clone/target destination pair — is now stated once, in the stronger of the two wordings. Against that, the review fold below spent 54 characters in the description and 182 in the schema.

## Harvest

|---|---|
| "Copy a record together with the records it DEPENDS ON — its link closure — out of one plugin universe and into a houseCARL patch, under NEW FormIDs, so the result no longer masters the plugin you copied from." | kept in description |
| "The mechanism is a whole-record duplicate, so every field carries by construction; EditorIDs are preserved on the copies." | kept in description |
| "WHAT GETS COPIED is decided by a rule, not a list: a linked record is INTERNALIZED (duplicated under a new FormID) when it is defined in the plugin(s) you are copying away from, or when it does not resolve in your active load order; every other link is KEPT and mastered normally." | `from_source` — stated there as the set `CopyClosure` actually binds: `from=`'s own defining plugin plus every plugin named in this parameter, with the `'winner'` pole naming none |
| "So vanilla and active shared-resource records stay links, and only what would VANISH with the source is copied." | `from_source` |
| "A record marked for internalizing that NO source in from_source= can produce — a dangling link, typically — refuses the whole copy naming every source consulted, rather than writing a patch with a hole in it." | `from_source`, folded with "A record no source has is refused, naming every source consulted" (below) into the one sentence that covers both subjects, so the rule is stated once |
| "WHERE THE WALK STARTS is yours: seed_paths= names the fields to walk from (e.g. ['HeadParts','HairColor','HeadTexture','WornArmor'] for an NPC's appearance)." | `seed_paths` (its weaker two-element example replaced by this four-element one with its "for an NPC's appearance" qualifier) |
| "Each must be a RECORD LINK or a LIST OF RECORD LINKS, judged on the field's DECLARED type — so a field the source happens to carry none of is still that shape." | `seed_paths` (its version lacked the declared-type caveat) |
| "A path that is not a field, or whose entries are structures rather than links (Factions, Perks, Items), is REFUSED BY NAME rather than quietly seeding nothing: for those use housecarl_apply's bundle=/assignments= zip, where op=Merge and op=ReplaceAll are your choice of merging into the target's entries or replacing them." | `seed_paths` (it replaces the bare "anything else is refused by name", which said the same thing with no route out) |
| "A seed the source leaves UNSET or EMPTY clears the target's, and says so — the result is the source's look, not a mixture of both." | `seed_paths` (its version stopped at "and says so") |
| "exclude_types= names record types the walk must not enter, each as 'Type:stop' (prune it, keep the link) or 'Type:refuse' (fail the whole copy) — a RACE is the standing 'refuse' case, since a race pulls skeletons and sibling races rather than an appearance subtree." | `exclude_types` (the spellings were already there with the `['Race:refuse']` example; the reason a RACE is the standing case was not) |
| "'stop' KEEPS the link, so wherever the link SURVIVES it needs a plugin the patch can master: with target=, pruning a record that is NOT in your active load order is refused up front, because an artifact cannot master a plugin the game does not load." | `exclude_types` |
| "With new_editorid= the clone's strip removes links into the source anyway, so a pruned off-order record is not refused there — unless one survives the strip." | `exclude_types` |
| "Any off-order link the finished patch still carries refuses, and the refusal names WHICH cause: your 'stop', a field seed_paths never named so the copy carried the link across, or a record an earlier call already put in the patch you are extending." | `exclude_types` |
| "Every element of exclude_types must name a type; a blank entry is refused rather than dropped, because dropping one silently applies fewer exclusions than you passed." | `exclude_types` (its version was "A blank entry is refused, not dropped" with no reason) |
| "FROM WHERE — from_source= is an ORDERED LIST of sources, tried in order, FIRST HIT WINS." | `from_source` (already said there) |
| "Each element is either 'winner' (the active load order's winning version of each record) or a plugin FILENAME, which resolves whether that plugin is active OR sitting on disk in a DISABLED mod." | `from_source` (already said there; this fuller spelling kept) |
| "Naming several is how you read a look from an override patch while its records come from the defining plugin: from_source=['Override.esp','Donor.esp']." | `from_source` |
| "This is FALLBACK, never a merge — a record present in several sources comes from the FIRST, and the result names which source produced each record — and, for a source that resolved into an MO2 MOD FOLDER (active or disabled), that folder, which is the name to pass as the provider when you place that record's files (its FaceGen, say) with housecarl_place." | `from_source` |
| "A source with no such folder — the 'winner' pole, MO2's overwrite, the game's Data folder, a file outside all of them — says so instead, so you never have to invent a folder name." | `from_source` |
| "A record no source has is refused, naming every source consulted." | `from_source`, folded into the one refusal sentence above that now covers both it and the internalizing case |
| "('winner' is the bare word: plugin names always carry an extension, so the two can never collide. 'previous_provider' is refused here — it is relative to a subject plugin, and a walk reaching records through links has no subject for it to be relative to.)" | `from_source` |
| "TWO DESTINATIONS (pass exactly one)" | kept in description, as the DESTINATION axis line — it is the composition rule between `target=` and `new_editorid=`; each parameter keeps its own "Pass this OR …" so the pairing survives the cut |
| "target= copies onto an EXISTING record — an active one, or a record in the patch itself when you pass into=" | `target` (already said there) |
| "the seed fields are set on it pointing at the internalized copies" | `target` |
| "new_editorid= mints a CLONE of the source record instead, and every link on it still pointing into the source is STRIPPED and reported by name — including when clearing one takes a WHOLE property with it (a script adapter, say), which the report says out loud." | `new_editorid` (the mint half was already there; the strip half was not) |
| "A link the record model REQUIRES cannot be stripped, so that refuses loud rather than writing an invented null or silently mastering the source." | `new_editorid` |
| "When NOTHING is being copied away from — the source is a base-game master and every source you named is one too — the result is reported as an appearance transplant rather than a standalone-ization, because an always-loaded master is not being removed from anything." | `from_source` |
| "Name a MOD in from_source= and that mod IS being copied away from, so the ordinary standalone report applies even for a base-game FormID." | `from_source` |
| "Writes a NEW plugin (folder-per-patch) or extends one with into=." | kept in description, as the OUTPUT axis line naming `patch=` and `into=`; `patch=` and `into=` are unchanged, so "folder-per-patch" is stated once |
| "Originals are never touched, and a refusal writes nothing." | kept in description |

Two notes on what the description gained and did not gain:

- The axis roster is the description's only index of the parameters now, so it names all eight — `from=`, `from_source=`, `seed_paths=`, `exclude_types=`, `target=`, `new_editorid=`, `patch=`, `into=`. The `WHAT` line says outright which plugins are being copied away from — `from=`'s own plus every plugin named in `from_source=` — because that rule left the description and a caller reading the roster needs to know which parameter it went to.
- The tool-boundary sentence ("This tool copies RECORDS. The FILES that go with them — an NPC's FaceGen mesh and tint, say — are placed with `housecarl_place`.") is new, and is the description's copy of the `housecarl_place` pointer. The old description carried that pointer only inside the `from_source` clause about which mod folder to pass as the provider; that clause moves whole to `from_source`, so the new sentence states the boundary without restating the folder rule. `housecarl_copy` never carried a `copy_npc_appearance` retirement note in its description — the alias layer handles that name — so there was none to keep.

No `@<absolute path>` rule, set-valued rule or runtime-FormID form was in this description to keep: `housecarl_copy` is a write verb, its FormID parameters take the plugin-qualified form only, and each already spells it. Nothing was invented.

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`: the `ToolNames.Copy` line comes out of `StillOversized`, and nothing else changes. Run before the description change, with the line already gone, `housecarl_copy` was the only failure: "housecarl_copy's published description is 5243 characters; Claude Code shows the first 2048 and drops the rest." After the change, and after the review fold below, the class is green and prints `housecarl_copy: 1104 characters (bound 2048)`.

## Review fold

- **The internalize gloss misstated the bound set** (`CopyTools.cs`). Real, and checked against `CopyClosure`: it binds `from=`'s own defining plugin as well as every named non-base-master arm, and skips the `'winner'` pole, so under the default `from_source=['winner']` nothing is named yet the source plugin is still internalized. The `from_source` description and the description's `WHAT` line now state that set. Published description re-measured: **1,104** characters.
- **`seed_paths` is required but nothing said so** (`CopyParserProbe.cs`, `CopyTools.cs`). Real: it is nullable, so the schema marks it optional, while the tool refuses without it. Its description now opens `REQUIRED.`. The same gap holds for pins no description promises either — the case-insensitive severity token, the whitespace-only `target=`, the empty `from_source=` taking the default — so the comment's header names both kinds rather than claiming all the pins are in prose.

## What was run

```
    1620 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. `src/housecarl-mcp-tests/data/tools-list-2.0.json` holds tool names only, so no capture needed regenerating. No CHANGELOG line: #585's entry covers the wave and the last PR of it generalises the wording.


src/housecarl-mcp/CopyTools.cs
```diff
@@ -19,70 +19,29 @@ public static class CopyTools
...
... extension, so the two can never collide. 'previous_provider' is refused here — it is relative to a subject plugin, and a walk reaching records through links has no subject for it to be relative to.) WHAT GETS COPIED is decided by a rule, not a list: a linked record is INTERNALIZED (duplicated under a new FormID) when it is defined in the plugin(s) you are copying away from — the ones named here — or when it does not resolve in your active load order; every other link is KEPT and mastered normally. So vanilla and active shared-resource records stay links, and only what would VANISH with the source is copied. A record no source can produce — a dangling link marked for internalizing, typically — refuses the whole copy naming every source consulted, rather than writing a patch with a hole in it. When NOTHING is being copied away from — the source is a base-game master and every source you named is one too — the result is reported as an appearance transplant rather than a standalone-ization, because an always-loaded master is not being removed from anything. Name a MOD here and that mod IS being copied away from, so the ordinary standalone report applies even for a base-game FormID.")]
```
Avick3110: The gloss added here — "the plugin(s) you are copying away from — **the ones named here**" — misstates the internalize set.

`CopyClosure` builds the bound set from **two** sources (`src/housecarl-mcp/LoadOrderService.cs`, ~5606-5623): `from=`'s own defining plugin, *plus* every non-base-master arm named in `from_source=`. The `winner` pole is explicitly skipped:

```csharp
if (!baseMasters.Contains(sourceKey.ModKey)) bound.Add(sourceKey.ModKey);
foreach (var arm in chain.Arms)
{
    if (string.Equals(arm.Spelling, SourcePoles.Winner, ...)) continue;
    ...
    if (!baseMasters.Contains(mk)) bound.Add(mk);
}
```

So with the documented default `from_source=['winner']` — the ordinary case — nothing is named here, yet the source plugin is still bound and its records are still internalized. Concretely: a caller runs `housecarl_copy(from='000D62:Vivace.esp', seed_paths=[...], new_editorid='X')`, reads "the ones named here", and concludes the walk will internalize nothing and that they must add `from_source=['Vivace.esp']` to get a standalone — which is redundant, since `Vivace.esp` is bound either way.

It also contradicts a sentence three later in this same description, which correctly requires *both* halves: "the source is a base-game master **and** every source you named is one too". The old wording ("the plugin(s) you are copying away from", full stop) was right; the added gloss is what breaks it.

The description's WHAT line (lines 23-24) carries the same error: "from_source= … names what is being copied AWAY FROM, which decides what the walk internalizes" — it does not name all of it, since `from=`'s plugin is bound whether or not it appears here.

src/housecarl-generator/CopyParserProbe.cs
```diff
@@ -12,15 +12,15 @@ namespace HousecarlGenerator;
...
+/// <para><b>Why this guard exists as a separate lane.</b> Every claim the parameter descriptions make below the record
 /// layer is a claim about a mapping — caller string → typed argument — and a mapping is exactly what a
 /// service-level fixture cannot see: <c>copy-service-guard</c> hands <see cref="HousecarlCore.WalkExclusion"/>
 /// values straight to <c>CopyClosure</c>, so deleting the <c>'Type:stop'</c> / <c>'Type:refuse'</c> parse would
 /// leave it green while the documented spelling stopped working. That is #339's lesson stated one layer up: an arm
 /// that drives the service does not test the wire.</para>
 ///
-/// <para>Pins, each of them a sentence the Description promises:
-///   EXCLUSIONS  — <c>'Type:refuse'</c> fails the whole copy, <c>'Type:stop'</c> prunes and keeps the link, and the
+/// <para>Pins, each of them a sentence a parameter's description promises:
```
Avick3110: This line now claims every pin is "a sentence a parameter's description promises", but the SEED PATHS pin below is promised by no description on the surface.

`seed_paths` is declared `string[]? seed_paths = null`, so the published schema marks it optional. Its description (`CopyTools.cs:38`) says only "WHERE THE WALK STARTS is yours: the fields to walk from…", and the tool description's THE WALK line says only "seed_paths= is where it starts". Neither states it is required. Concretely: a model omits `seed_paths` because the schema says optional and no prose says otherwise, and gets back `error: seed_paths is required — name the link-bearing field(s) the walk starts from`.

Two EXCLUSIONS sub-claims are in the same position — "The severity token is case-insensitive" and "an unknown one refuses by name" appear in no description either. (Relatedly, the parser defaults a bare `Race` with no colon to `refuse`, which no description mentions.)

This was already loose when the line pointed at "the Description", but the PR's whole point is to make the parameter descriptions the source of truth for these pins, so it is worth closing rather than repointing: either add "Required." to the `seed_paths` description, or soften this line to say the pins are behaviours the guard holds, not all of them promised in prose.

src/housecarl-mcp/CopyTools.cs
Avick3110: Right, and verified against CopyClosure — the bound set is from='s own defining plugin plus every named non-base-master arm, with the winner pole skipped, so the gloss was wrong in the default case; both the from_source description and the WHAT line now state that set (45e85d94).

src/housecarl-generator/CopyParserProbe.cs
Avick3110: Folded both ways: seed_paths now opens with REQUIRED, and since the case-insensitive severity token, the whitespace-only target= and the empty-from_source= default are pins no description promises either, the header now says the pins are the described sentences plus those parse tolerances rather than claiming all of them are in prose (d1d9f644).

PR #590, opened 2026-09-06
Bring the apply tool description under the 2,048-character cap

Avick3110:
`housecarl_apply` published a 7,886-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the op grammar, the copy zip, the lane rules, the dry-run contract and the transport rules were all being described in a part of the text the model never saw. Nothing is deleted here: each paragraph moves into the description of the parameter it is about — the ops[] DTO members for the op grammar, `bundle`/`assignments` for the zip, `into`/`in_place`/`acknowledge`/`dry_run` for the lane, `readback`/`format` for transport — and where the parameter already said the same thing the description's copy is dropped as a duplicate rather than pasted in twice. The description keeps the one-surface sentence, the FormID forms, the `@<absolute path>` rule with each list input's own spelling, the whole-call write model and the all-or-nothing guarantee, one line per axis naming its parameters, and the pointers to the other write tools. `housecarl_apply` comes off the bound test's `StillOversized` list, which now names ten tools.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_apply` description | 7,886 | **2,032** |
| Top-level parameter descriptions | 3,142 | 5,031 |
| All schema descriptions (parameters + nested DTO members) | 9,819 | 13,164 |

Re-measured after the review fold; the description was 2,040 before it.

The description shed 5,854 characters and the schema gained 3,345; the 2,509-character difference is text the description was repeating from a parameter that already carried it (the ops member roster, the verb recital, `values=`/`entries=`/`key=`, the ZIP-not-a-product rule, the bundle's by-construction rule, the `patch=` default, the `into=` opening, most of the `in_place=` and `acknowledge=` sentences, `format='json'`, `max_chars`).

## Harvest

|---|---|
| "Edit fields on one or many records and write the result to a NEW patch plugin (originals untouched by default)." | kept in description |
| "ONE surface: what to change (ops=, or the bundle=/assignments= copy zip) x WHERE it lands (the LANE: …) x how it reads back (TRANSPORT)." | kept in description |
| "A FormID is 'XXXXXX:Plugin.esp' — 6 hex digits, a colon, the defining master's filename, and every FormID this tool takes (ops[].formid, from=, a field VALUE) is that form." | kept in description |
| "The RUNTIME form the game, the console and the logs print … is refused with the 'XXXXXX:Plugin.esp' form to use in its place." | kept in description |
| "`housecarl_records` reads by either form." | kept in description |
| "Every edit resolves the record's load-order WINNER and overrides it into the patch; all edits land in ONE reviewable .esp, whose master header spans every plugin the edits reference (cross-master merge, derived not declared)." | kept in description |
| "OPS. ops= is the edit list — {formid, field_path, op?, …} each; one op is a set of one." | `ops` (already said there in the same words) |
| "field_path is dotted ('BasicStats.Damage', 'Name', 'Keywords'); step INTO a list/dict element MID-path with brackets ('Effects[0].Data.Magnitude'), but at the LEAF use op + key instead of brackets." | `ops[].field_path`, which already said the bracket rule; it gains the 'Name' and 'Keywords' examples |
| "value is coerced to the field's real type — a number, an enum name ('OneHanded'), or a FormID for a reference." | `ops[].value` (its bare "coerced to the field's type" replaced) |
| "op is Set (default) \| Add \| … \| CopyFrom." | `ops[].op` (already said there, off the same `WriteVerbs.AllRecital` const) |
| "InsertAtIndex inserts a NEW element AT key= and shifts the rest right (key = the list's length appends); use it to grow a POSITION-CONTIGUOUS run in place, e.g. adding an arm to an existing CTDA OR-group, where Add would land the row at the end as a separate AND-group." | `ops[].op`, which had the first clause; it gains the OR-group example |
| "On a [Flags] enum (SPEL Flags, NPC Configuration.Flags, WEAP Data.Flags...) Add SETS a bit and Remove CLEARS one … to turn all bits off, Set the field to '0'." | `ops[].op` (its weaker one-line version replaced) |
| "Omit value only for a Remove that whole-clears a NULLABLE field." | `ops[].value` |
| "values= is the whole new list for a list ReplaceAll" | `ops[].values` (already said there) |
| "entries= (key->value) drives a dict Merge or dict ReplaceAll" | `ops[].entries` (already said there) |
| "key= is the dict key or list index at the leaf." | `ops[].key` (already said there) |
| "compose= builds a MODELED struct for an Add / InsertAtIndex / SetAtIndex, or a polymorphic Set — a leveled-list entry, an effect, a condition row, or a polymorphic list element by its CONCRETE arm type" | `ops[].compose` (its weaker version replaced; the arm-type rule and its two examples are stated once, not twice, each example next to the row kind it names) |
| the VMAD worked example (`op=Add, field_path='VirtualMachineAdapter.Scripts[0].Properties', compose={type:'ScriptObjectProperty', fields:{…}}`) | `ops[].compose` |
| the leveled-list worked example (`op=Add, field_path='Entries', compose={type:'LeveledItemEntry', sets:[…]}`) | `ops[].compose` |
| "composes= is the BATCH sibling — a LIST built in ONE op: with Add it APPENDS each in order (a whole block of condition rows at once), with ReplaceAll it CLEARS the list then appends each (the way to replace a whole modeled list); composes=[] with ReplaceAll clears the list to empty. compose and composes are mutually exclusive." | `ops[].composes`, which had the mechanics; it gains the two parentheticals |
| "COPYING A FIELD: op=CopyFrom takes no value — the source IS another record's version." | `ops[].op`, where the verb recital lives (the no-value half was already on `ops[].value`) |
| "from_source= names the plugin to copy from (an ACTIVE plugin, or a plugin FILE on disk that isn't in the load order — a disabled old patch you want to re-assert a field from)." | `ops[].from_source` (already said there; it gains "you want to re-assert a field from") |
| "from= names a DIFFERENT source RECORD (defaulting to from_source's version of the record being edited); with from= and no from_source= the source is that record's load-order winner." | `ops[].from` and `ops[].from_source` (already said on both) |
| "Cross-record pairs must be the SAME record type — refused by name otherwise." | `ops[].from` (its "must be the same record type" gains "refused by name otherwise") |
| "CopyFrom copies a WHOLE field (scalar, formlink, modeled list, sub-struct); it cannot copy owned child records (forward the whole record with `housecarl_forward` instead)." | `ops[].op` (a verb rule, so it sits with the verb; `ops[].from` points at it) |
| "THE COPY ZIP (bundle= x assignments=). To copy the SAME set of fields from one record to another — many pairs in one call — name the paths once and pair them explicitly: bundle=[…], assignments=[…]." | `bundle` for the framing and the paths-once rule; both examples were already on `bundle` and `assignments` |
| "It is a ZIP, never a product: each target takes its OWN source." | `assignments` (already said there) |
| "Identity and everything outside the bundle are untouched BY CONSTRUCTION — a bundle only names what it copies." | `bundle` (already said there; it gains the "a bundle only names what it copies" clause) |
| "Which paths form an appearance set or a balance frame is knowledge a skill carries, not a verb this tool owns." | `bundle` |
| "The zip composes with ops= in one call." | kept in description (the COMPOSITION line) |
| "LANE — where the write lands. Default: a NEW patch named patch= (auto-suffixed if taken, so a prior patch is never overwritten)." | `patch` (already said there) |
| "into='<an existing patch's filename>' EXTENDS that patch instead — the way to accumulate across calls and sessions." | `into` (already said there) |
| "PRECEDENCE with into= (pinned): a FormKey the patch ALREADY CARRIES is edited AS-IS in the patch; only a FormKey it does NOT yet carry copies the load-order winner in first." | `into` |
| "So `housecarl_forward` from a source + apply into= the same patch is THE recipe to build on a specific plugin's version while a stale winner sits above it." | `into` |
| "in_place='<plugin filename>' is the opt-in THIRD lane: houseCARL rewrites your ORIGINAL file (incl. a mod it didn't author) — no new patch, and NO houseCARL backup or undo (keep your own)." | `in_place` (already said there; it gains "(keep your own)") |
| "It re-lays-out the whole plugin the way xEdit/CK do on save, VERIFIES the records you edit, trusts Mutagen for the untouched rest, and refuses a file it can't parse or that holds engine-reserved (sub-0x800) records." | `in_place` |
| "The FIRST in-place edit of a given plugin returns a confirmation prompt (re-call with acknowledge=true)" | `acknowledge` |
| "that consent covers touching your original ONLY — it NEVER skips the record verify." | `acknowledge` (already said there) |
| "dry_run=true runs the FULL pipeline — winner resolve, schema pre-flight, every op applied in memory, the reference-resolution check — and STOPS before anything touches disk." | `dry_run` (its "run the whole real pipeline" replaced by the four named stages) |
| "It returns what WOULD change and the expected masters, or EXACTLY the refusal the real call would give: catch a bad field path before the first write of a big batch, not after the last." | `dry_run`, which had the first clause; it gains the big-batch reason |
| "Works on every lane (an in-place dry run needs no acknowledge and never records consent)." | `dry_run` (it gains the parenthetical) |
| "Not a disk guarantee — a serialize/commit fault still surfaces only for real." | `dry_run` |
| "ALL-OR-NOTHING (Q3): if ANY op is malformed or fails pre-flight, the whole call is refused with per-op reasons and NOTHING is written. No partial patches, ever." | kept in description — it governs ops= and the zip alike, so it belongs to neither alone |
| "TRANSPORT: readback=true expands the read-back to the FULL deep field-by-field dump of every touched record (in place, the verify ALWAYS runs and shows compactly by default; readback widens it) — confirm composed structures landed and nothing else was disturbed WITHOUT enabling the patch in MO2." | `readback`, which had the dump sentence; it gains the in-place clause and the reason |
| "The read-back is the WRITTEN FILE's content, NOT load-order truth: the patch wins nothing until enabled + sorted in MO2." | `readback` (its shorter "not load-order truth" replaced) |
| "max_chars caps the render with an explicit notice, never a silent cut." | `max_chars` (already said there, with the render's two lanes spelled out) |
| "Every response carries epoch=<hex> — the identity of the index build the winners were resolved from." | `format`, with both spellings named: `epoch=<hex>` on text, an `epoch` member on json |
| "ops= (like every list-valued input) also accepts \"@<absolute path>\" in place of the inline array: the SAME array as a JSON manifest on disk." | kept in description, naming all three list inputs with the spelling each takes — `"@<absolute path>"` on `ops=` and `assignments=`, `["@<absolute path>"]` on `bundle=`, which is typed `string[]`; `ops` already carried its own half |
| "Write a big job's ops once, dry-run the file, then apply it — and re-run the same manifest to recover an interrupted write (overrides are idempotent)." | `ops` |
| "The path must be ABSOLUTE (the server resolves relative paths against its OWN working directory, not yours), and the file is read at CALL time, so re-dry-run it after editing." | kept in description (it governs all three list inputs) |
| "This tool edits EXISTING records' fields. New records are `housecarl_create`; dropping whole records is `housecarl_remove`; copying a whole record verbatim is `housecarl_forward`. Read first with `housecarl_records`." | kept in description |

The description gained one thing it did not carry before: the three axis lines (WHAT / LANE / TRANSPORT), naming every parameter. The roster is the description's only index of the parameters now, and `patch=`, `acknowledge=`, `readback=`, `format=` and `max_chars=` were reachable from the old text only as prose mentions.

## Stale pointers repointed

Removing `"op is " + WriteVerbs.AllRecital + ". "` from the description left three comments and one refusal sentence naming text that is no longer there:

- `WriteVerbs.cs` and `DescriptionVocabularyGuardProbe.cs` both said the const's other two conversion sites are "one appends after a full stop, one reads `"op is " + AllRecital + ". "`". There is one other site now — `ApplyOp.op` — and both comments name it.
- The probe's `INV4-MARK` doc comment and failure message said the const "feeds two shipped descriptions"; it feeds one, plus the vestigial `BulkOp.verb` attribute they already name.
- `CorpusRulebook.PlacingRemedy`'s fallback sentence said "the verbs this field takes are in the tool description". It now says "in the op member's description", which is where both shipped write surfaces recite them. (Unreachable today by construction — no corpus field has a declined element kind — but it would have pointed at nothing.)

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`: the `ToolNames.Apply` line comes out of `StillOversized`, and nothing else changes. Ran the class with the line gone and the description untouched — `EveryPublishedDescriptionFitsTheClientsCut(tool: "housecarl_apply")` failed at 7,886, which is the first commit here. It passes at 2,032 after the review fold.

## Review fold

- Named each list input's `@file` spelling in the description. `ops=` and `assignments=` bind as `JsonElement`, so a bare `"@<path>"` reaches `ListParams.Read`; `bundle=` is `string[]` and `ReadBundlePaths` takes the one-element `["@<path>"]`, which is what `bundle`'s own description says. The old generalisation contradicted it, and the bare string would have died in the SDK binder.
- Moved the `compose=` example `'LeveledItemEntry'` off "a condition row" (whose arm is `ConditionFloat`) and onto "a leveled-list entry".
- Moved "CopyFrom takes no value" and the whole-field / no-owned-child-records limit off `ops[].from` — which the ordinary same-record copy omits — onto `ops[].op`, where the verb recital is; `from` keeps the same-record-type rule and points at `op=`.
- Corrected the tail-gloss docstring's "one site instead of three" to match the "the other conversion site" line the PR changed a sentence below it.

The description's margin under the cap is 16 characters, up from 8. It embeds `ToolNames` five times (`housecarl_records` twice, plus `housecarl_create`, `housecarl_forward`, `housecarl_remove`), so a rename spending 17 characters across those sites would put it back over — the bound test catches that, but in a PR about the rename.

## What was run

```
    1620 passed, 0 failed
    64 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. No checked-in capture snapshots the descriptions: `src/housecarl-mcp-tests/data/tools-list-2.0.json` holds tool names only. No `plugin/CHANGELOG.md` line — the records PR's entry covers this wave.


src/housecarl-mcp/ApplyTools.cs
```diff
@@ -25,100 +25,45 @@ public static class ApplyTools
          "this tool takes (ops[].formid, from=, a field VALUE) is that form. The RUNTIME form the game, the console " +
          "and the logs print — eight hex digits and no plugin name, 'FExxxYYY' / 'XX######' — is READ-ONLY, because " +
          "it names a slot in the load order as it stands rather than a record: pass one here and it is refused with " +
-         "the 'XXXXXX:Plugin.esp' form to use in its place. " + ToolNames.Records + " reads by either form. " +
+         "the 'XXXXXX:Plugin.esp' form to use in its place. " + ToolNames.Records + " reads by either form. Every " +
+         "list-valued input (ops=, bundle=, assignments=) also accepts \"@<absolute path>\" in place of the inline " +
```
Avick3110: The generalisation is wrong for `bundle=`. `ops=` and `assignments=` are `JsonElement?`, so a bare `"@<absolute path>"` binds and `ListParams` resolves it — but `bundle` is `string[]?`, and `ReadBundlePaths` only fires the file branch at `bundle.Length == 1 && bundle[0].StartsWith('@')`, i.e. the `["@<path>"]` spelling. `bundle`'s own parameter description still says `Accepts ["@<absolute path>"]`, so the two descriptions now contradict each other on the same call.

Concrete failure: a caller who reads this sentence sends `bundle: "@C:/jobs/appearance.json"`. STJ cannot deserialize a JSON string into `string[]`, so the call dies in the SDK binder before `Apply` runs — the caller gets a schema/binder error rather than one of this tool's plain sentences, and `ReadBundlePaths`'s carefully worded refusal never gets a chance to fire.

Suggest scoping the sentence back to the two inputs that take the bare form, and naming bundle's array spelling.

src/housecarl-mcp/ApplyTools.cs
```diff
@@ -337,16 +282,16 @@ public sealed record ApplyOp
     [JsonPropertyName("entries"), Description("Key->value pairs for a dict Merge or dict ReplaceAll.")]
     public Dictionary<string, string>? Entries { get; init; }
 
-    [JsonPropertyName("compose"), Description("Build a modeled struct: the arm for a polymorphic Set, or the element for a struct-element Add / InsertAtIndex / SetAtIndex (e.g. 'LeveledItemEntry'; for a polymorphic list, the element's CONCRETE arm type such as 'ScriptObjectProperty').")]
+    [JsonPropertyName("compose"), Description("Build a MODELED struct for an Add / InsertAtIndex / SetAtIndex, or a polymorphic Set — a leveled-list entry, an effect, a condition row (e.g. 'LeveledItemEntry'), or a polymorphic list element by its CONCRETE arm type (e.g. 'ScriptObjectProperty'). A VMAD script property: op=Add, field_path='VirtualMachineAdapter.Scripts[0].Properties', compose={type:'ScriptObjectProperty', fields:{Name:'MyProp', Flags:'Edited', Object:'XXXXXX:Plugin.esp', Alias:'-1'}}. Merging a weapon into a leveled list: op=Add, field_path='Entries', compose={type:'LeveledItemEntry', sets:[{path:'Data.Level',value:'1'},{path:'Data.Count',value:'1'},{path:'Data.Reference',value:'<weapon FormID>'}]}.")]
```
Avick3110: The `(e.g. 'LeveledItemEntry')` parenthetical moved and now attaches to the wrong item. It reads as the example for "a condition row", the noun immediately before it — but a condition row's concrete type is `ConditionFloat` (or another `Condition` arm); `LeveledItemEntry` is the example for "a leveled-list entry", three nouns earlier. The old wording had the e.g. after the whole Add/InsertAtIndex/SetAtIndex clause, where it read as generic.

Concrete failure: a caller composing a CTDA follows the nearest example and sends `compose={type:'LeveledItemEntry', …}` on a `Conditions` list. Pre-flight refuses the whole call (all-or-nothing), so it costs a round trip on a description that was supposed to prevent exactly that guess.

Moving the parenthetical next to "a leveled-list entry" — or dropping it, since the worked example two sentences later already shows `LeveledItemEntry` on `Entries` — fixes it.

src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
```diff
@@ -1655,8 +1655,8 @@ static void DefaultSlotsArm(List<(SurfaceSite Site, string Verb)> marks)
     /// happens to end with. Appending a ninth verb to the const, which is exactly the edit the const exists to
     /// make sufficient at one site instead of three, moves the gloss onto the new verb and strips it off the old
     /// one, and no other arm sees it, because the recital is still complete and every token in it is still a real
-    /// verb. Reordering does the same. The other two conversion sites are position-independent — one appends after
-    /// a full stop, one reads <c>"op is " + AllRecital + ". "</c>.</para>
+    /// verb. Reordering does the same. The other conversion site, <c>ApplyOp.op</c>, is position-independent — it
```
Avick3110, line 1658: The stale-pointer sweep stopped one sentence short. The line above still says the const "exists to make sufficient at one site instead of three", and this PR is what made that three a two: `AllRecital` is now concatenated at `BulkOp.verb` and `ApplyOp.op` only, which is exactly why the next clause changed from "the other two conversion sites" to "the other conversion site". The paragraph now contradicts itself about the same count.

Same class of thing as the `WriteVerbs.cs` and INV4-MARK edits this PR already made — worth folding in while the paragraph is open.

src/housecarl-mcp/ApplyTools.cs
```diff
@@ -337,16 +282,16 @@ public sealed record ApplyOp
...
     public StructInput[]? Composes { get; init; }
 
-    [JsonPropertyName("from"), Description("op='CopyFrom' only: the SOURCE RECORD to copy the field from, as 'XXXXXX:Plugin.esp' — a DIFFERENT record from formid (SPEC §4.5's cross-record copy). Omit to copy this same record's version from another plugin (name it in from_source). Source and target must be the same record type.")]
+    [JsonPropertyName("from"), Description("op='CopyFrom' only: the SOURCE RECORD to copy the field from, as 'XXXXXX:Plugin.esp' — a DIFFERENT record from formid (SPEC §4.5's cross-record copy). CopyFrom takes no value: the source IS another record's version. Omit to copy this same record's version from another plugin (name it in from_source). Source and target must be the SAME record type — refused by name otherwise. CopyFrom copies a WHOLE field (scalar, formlink, modeled list, sub-struct); it cannot copy owned child records (forward the whole record with " + ToolNames.Forward + " instead).")]
```
Avick3110: medium: the CopyFrom rules landed on a member the common CopyFrom caller omits. `CopyFrom takes no value` and `CopyFrom copies a WHOLE field (scalar, formlink, modeled list, sub-struct); it cannot copy owned child records` are properties of the verb, not of `from=`, but they now live only here — and this member's own second sentence says to omit it for the same-record copy (`from_source=` alone, no `from=`), which is the ordinary way to re-assert a field from another plugin. That caller reads `ops[].op` and `ops[].from_source`, neither of which says either thing, so the child-record limit that used to be in the tool description is now unreachable for them and they find it only in the refusal. `Cross-record pairs must be the SAME record type` genuinely belongs here; move the other two to `ops[].op` (which recites CopyFrom but glosses it nowhere) or repeat the whole-field rule on `ops[].from_source`.

src/housecarl-mcp/ApplyTools.cs
```diff
@@ -25,100 +25,45 @@ public static class ApplyTools
...
+         "declared). ALL-OR-NOTHING (Q3): if ANY op is malformed or fails pre-flight, the whole call is refused " +
+         "with per-op reasons and NOTHING is written. No partial patches, ever.\n\n" +
+         "Each axis's grammar is on its own parameters:\n" +
+         "WHAT — ops= (the edit list), or the copy zip bundle= x assignments=.\n" +
+         "LANE — patch= | into= | in_place= with acknowledge= | dry_run=.\n" +
+         "TRANSPORT — readback= | format= | max_chars=.\n\n" +
+         "COMPOSITION: the zip composes with ops= in one call.\n\n" +
          "This tool edits EXISTING records' fields. New records are " + ToolNames.Create + "; dropping whole records is " +
          ToolNames.Remove + "; copying a whole record verbatim is " + ToolNames.Forward + ". Read first with " +
          ToolNames.Records + ".")]
```
Avick3110, line 47: note, not a defect: I rebuilt the worktree and measured the published description independently at 2,040 characters, so the PR's number holds and the `StillOversized` line comes out correctly. The 8-character margin does depend on `ToolNames`: the description embeds `housecarl_records` twice plus `housecarl_create`, `housecarl_forward` and `housecarl_remove`, five occurrences in all. Renaming `ToolNames.Records` to anything five characters longer, or adding nine characters spread across those five sites, puts the description back over the cap — the bound test would catch it, but the failure would land in a PR about a rename rather than about this description. Recording it here so the next person is not surprised by it.

src/housecarl-mcp/ApplyTools.cs
Avick3110: Folded in c225fdec. The sentence now names both spellings: ops= and assignments= take "@<absolute path>", bundle= takes ["@<absolute path>"], which is what bundle's own description says.

src/housecarl-mcp/ApplyTools.cs
Avick3110: Folded in bb540a10 — the example now sits on "a leveled-list entry"; a condition row's arm is ConditionFloat.

src/housecarl-generator/DescriptionVocabularyGuardProbe.cs
Avick3110: Folded in 5b47b17e — the line now reads "at one site instead of two", matching the sentence below it.

src/housecarl-mcp/ApplyTools.cs
Avick3110: Folded in 199f21fe — "CopyFrom takes no value" and the whole-field / no-owned-child-records limit are now on ops[].op with the verb recital; ops[].from keeps the same-record-type rule and points at op=.

src/housecarl-mcp/ApplyTools.cs
Avick3110: Noted, and recorded in the PR body; the fold above cut the description to 2,032, so the margin is now 16 characters rather than 8.

PR #591, opened 2026-09-06
Refuse a fresh patch name that shadows an inactive plugin

### LINKED ISSUE #561: [Bug] a fresh patch stem is not checked against inactive plugins in foreign folders (#359's second half)
## Summary

The second half of #359, split out on the 2026-09-05 overnight run so PR #552 could close the refusal side.

`IsStemFree` tests only for an owned `houseCARL - <stem>` folder and an ACTIVE plugin of that name. `patch="My Cool Patch"` beside a foreign folder holding an inactive `My Cool Patch.esp` still mints a stem that shadows that file once the user enables both. PR #552 stops the refusal from walking the caller into it and says in the CHANGELOG that the name is not checked against inactive plugins in foreign folders; the uniqueness rule itself is unchanged.

## Why it is its own issue

Changing `IsStemFree` changes the fresh-write path on every `patch=` call, not just the refusal text; it wants its own test against a foreign folder with an inactive plugin and a decision on whether a shadowing stem is refused or renamed.

Avick3110:
A new plugin's filename was checked against two things only: the mod folders houseCARL owns, and the plugins the active order loads. Neither sees a plugin sitting in someone else's mod folder that MO2 has switched off or left unticked, so `patch="My Cool Patch"` beside a foreign folder holding an inactive `My Cool Patch.esp` minted a second plugin of that name — two files that can never both be active — and said nothing. The exact filename a fresh write is about to create now goes through a shadow check, and a collision on a name the caller chose is refused, never suffixed around: suffixing would hide the very file the caller may have meant. The check lives in its own small file (`PatchStemShadow`) and is called from `UniqueStem`, the one place both fresh-write lanes allocate a stem, plus `CreatePlugin`, which mints its name by hand because a header-only trigger's basename is load-bearing. It reuses the existing off-order plugin locate (`Mo2LoadOrder.LocatePlugin`) rather than adding a second directory walk, so it searches the same layers a plugin read reaches: the overwrite folder, every mod folder enabled, switched off or not yet listed by MO2, and the game `Data` folder.

Closes #561

## What is checked, and on which tool

The test is on the plugin FILE a call will write, not on the mod-folder stem, because what makes two plugins a collision is the filename. Each lane states the file it emits and the parameter its caller changes to move it:

| Lane | File checked | Parameter the refusal names |
|---|---|---|
| `housecarl_create` / `_copy` / `_forward` / `_apply` (fresh `patch=`) | `<stem>.esp` | `patch=` |
| `housecarl_merge_plugins` | `output=` | `output=` |
| `housecarl_create_plugin` | `<plugin_name>` as a BASENAME — `.esp`, `.esm` and `.esl` | `plugin_name=` |
| `housecarl_compile_script`, `_decompile_script`, `_bsa_repack`, `_bsa_extract`, `_place`, `_nif_set`, `_write_seq` | none — these write scripts, an archive, a mesh, loose files or a `.seq` | no shadow refusal at all |
| `housecarl_compact_plugin` | none — see below | no shadow refusal at all |

`create_plugin` is the one lane that sweeps all three extensions, and it does so for the reason its own active-order check already does: a header-only trigger resolves by BASENAME, so a foreign `MyTrigger.esm` is exactly the second file that check exists to prevent. The record lanes stay on `.esp` alone, because two different filenames coexist fine — `AForeignEsmOfTheSameStemIsNoShadowForTheEspTheLaneWrites` pins that.

`compact_plugin` is deliberately out. Its output keeps the source's exact basename so the source's external masters still resolve, and an inactive source is a documented target of the tool, so checking that filename would refuse the tool's own purpose on the very lane it was built for. Its output is a same-named twin the user swaps in, not a new name houseCARL invented.

Ownership is not an axis. Any folder holding the exact filename this call will write is a shadow, houseCARL's own folders included, so a houseCARL patch whose folder the user RENAMED no longer slips through. The folder a fresh write lands in is created after the check and holds nothing, so it can never be its own shadow, and the `into=` extend lane never reaches the check. There is no `meta.ini` read on this path.

## The refusal, as it renders

```
cannot write 'My Cool Patch.esp': mod 'Foreign Patch Mod' (DISABLED) already holds 'My Cool Patch.esp', which the load order is not loading, and two plugins cannot share one filename — pass patch= a name nothing on your install already uses, or enable 'My Cool Patch.esp' and target that plugin if it is the one you meant.
```

The layer label is the one `LocatePlugin` already renders, so a hit in the overwrite folder or in `Data` names itself rather than being called a mod — and the remedy says "nothing on your install" for the same reason, since two of the three places it searches are not mod folders. The parameter spelling is the calling lane's own, so no caller is sent to a parameter their tool does not declare. Where the file found is not the file being written — only `create_plugin`, sweeping the basename — the sentence names both, and the lane states why the two collide ("a header-only trigger needs a UNIQUE basename"). The consent-gated in-place lane is deliberately not spelled here, the way Aaron's 2026-09-05 ruling keeps it out of the un-owned extend refusal; the second remedy says to enable the plugin and target it, in whatever lane the caller's tool has.

## Two cases deliberately left alone

- **A plugin the order IS loading.** That is what the existing suffix loop is for, and a generic default stem (`patch=` omitted, stem `Patch`) depends on it. The shadow check skips a name already in the active order, so that behaviour is unchanged.
- **A name houseCARL chose or derived.** The ruling is about the name the caller passed. A shadow on `Patch_001` when the caller passed no name at all is not a name they can be told to avoid, so the allocator steps to the next suffix as it always did; only the caller's own `patch=` is refused. That includes the copy lane's fallback: with `patch=` omitted, `housecarl_copy` derives the stem from `new_editorid=`, and a shadow on that derived name steps rather than refusing on a parameter the call never passed. The one exception is a lane whose filename does not follow the stem (`merge_plugins`, whose `output=` is fixed), where every suffix would land on the same file and stepping could never clear it.

## Cost, and when the check does not run

Paid once per FRESH write on a lane that actually writes a plugin, never on an `into=` extend, and inside the same `_gate` the stem allocation already holds: one profile-composition read (three small text files) plus one `LocatePlugin` sweep for ONE filename — one filename stat per candidate mod folder, plus one ModsDir listing for the not-yet-listed layer, opening no plugin. No `meta.ini` is read. A lane that writes no plugin passes no shadow target and does not pay the composition read at all.

The check is skipped, leaving the pre-existing folder and active-order uniqueness in place, when the order's active plugins cannot be read at all, or when the profile composition is not USABLE — a profile that names no mod in either list, which is what a missing `modlist.txt` produces. That second guard is on the composition's content, not on whether the read threw, because `ReadComposition` does not throw on a missing `modlist.txt`: it returns empty mod lists, and then every folder under ModsDir reads as UNLISTED and a genuinely loaded plugin would be refused as one the order is not loading. A `modlist.txt` truncated to a PARTIAL list is not detectable here and is not claimed to be.

## Tests

`src/housecarl-mcp-tests/PatchStemShadowTests.cs`, driven through `housecarl_create`'s engine entry — the shortest real `patch=` write — and through `housecarl_merge_plugins`, `housecarl_create_plugin` and `housecarl_copy`. Each test is on its own world, since they add mod folders and rewrite the profile's modlist. Every one fails before the change it covers and passes after.

- `AStemThatWouldShadowAnInactivePluginInAForeignModFolderIsRefused` — a foreign, switched-off mod folder holds an inactive `My Cool Patch.esp`; `patch="My Cool Patch"` is refused naming that folder and that file, and no patch folder of either spelling is left behind.
- `AForeignEsmOfTheSameStemIsNoShadowForTheEspTheLaneWrites` — a foreign folder holds `My Cool Patch.esm`, a different filename that coexists fine; the record lane's `.esp` write proceeds.
- `AShadowOnAStemTheCallerDidNotChooseStepsToTheNextSuffix` — `patch=` omitted and a foreign folder holds `Patch.esp`; the allocator lands on `Patch_001.esp` instead of failing.
- `AShadowOnAnEditoridDerivedStemStepsToTheNextSuffix` — `housecarl_copy` with `new_editorid="Donor"`, no `patch=`, against a switched-off donor holding `Donor.esp`; it lands on `Donor_001` rather than refusing on a `patch=` the call never passed.
- `AHeaderOnlyCreateIsRefusedByAnyExtensionSharingItsBasename` — a switched-off mod holds `HcShadowTrigger.esm`; `create_plugin("HcShadowTrigger")` is refused naming `plugin_name=` and the `.esm` it found.
- `AProfileNamingNoModAtAllSkipsTheShadowSweep` — `modlist.txt` deleted while the order still loads a plugin from the game `Data` folder; the write proceeds instead of being refused on a folder the sweep has misread as UNLISTED.
- `AMergeIsRefusedOnTheOutputFilenameItWritesNotItsFolderStem` — a foreign folder holds an inactive `HcShadowMerge.esp` and `output="HcShadowMerge.esp"` is refused naming `output=`.

## Also changed

The `patch=` parameters on the record lanes now name all three places the sweep reaches. The parameters on the lanes that write no plugin no longer advertise a refusal they cannot get, `merge_plugins` states it on `output=` and `create_plugin` on `plugin_name=` — on the parameters, never on the tool descriptions, so the description cap test stays green. `plugin/CHANGELOG.md` gains one `Unreleased` line, and PR #552's line points at it instead of saying the name is not checked against inactive plugins in foreign folders.

## What was run

From the worktree, on `claude/stem-shadow`:

```
dotnet build housecarl.sln -c Release     -> 0 errors
                                          -> 1626 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with the parallel build sessions on this machine); CI runs it. The probes' foreign-folder arms were read rather than run: every one of them reaches the foreign folder through `into=`, not through a fresh write, and the one that holds an active plugin is a case this check skips.


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7713,7 +7714,10 @@ public RiderFolder ResolvePatchModFolder(string? patchName, string? into, string
                 return new RiderFolder(folder, folder, CreatedFresh: false);   // reused — the user owns it; cleanup leaves it
             }
 
-            var newStem = UniqueStem(PatchStem(string.IsNullOrWhiteSpace(patchName) ? defaultStem : patchName!));
+            // The rider's OWN folder parameter, the way its extend refusal already names it — patch= is wrong on a
+            // lane where a bare patch= binds to something else.
+            var newStem = UniqueStem(PatchStem(string.IsNullOrWhiteSpace(patchName) ? defaultStem : patchName!),
+                                     naming?.Param ?? "patch");
```
Avick3110: **Three rider lanes pass `naming: null`, so the refusal names `patch=` on tools that do not declare it.**

`naming?.Param ?? "patch"` treats a null `RiderNaming` as "this lane declares patch=", but null is exactly the case the type documents as *unknown* ("A lane whose `into=` can never be non-empty passes null, and keeps the weakest true remedy"). Three callers pass null:

- `LoadOrderService.cs:6611` — **`housecarl_compact_plugin`**, which declares `patch_name=`.
- `LoadOrderService.cs:7021` — **`housecarl_merge_plugins`**, which declares `patch_name=`.
- `BsaTools.cs:71` — **`housecarl_bsa_extract`**, which declares *no* folder-name parameter at all; the stem is derived from the archive filename.

Scenario: a disabled mod folder holds an inactive `Foo compacted.esp`; `housecarl_compact_plugin plugin="Foo.esp" patch_name="Foo compacted"` is refused with "pass `patch=` a name no mod folder already uses" — a parameter that tool does not declare, which is the precise failure the PR body says the `patchParam` argument exists to prevent. This PR even rewrites `compact_plugin`'s and `merge_plugins`' `patch_name` descriptions to advertise the refusal while the refusal itself names `patch=`.

For `bsa_extract` the remedy is not merely misspelled but unreachable: the caller cannot influence the stem at all, so the only real escape is `dest=`, which the sentence never mentions.

Making the `RiderNaming?` parameter carry a real spelling at those three call sites (or making `patchParam` non-defaultable) would close it.

src/housecarl-mcp/PatchStemShadow.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <summary>The first thing <paramref name="stem"/> would shadow, or null when the stem is clear. Searches the
+    /// same layers a plugin read reaches — the overwrite folder, every mod folder enabled, disabled or not yet listed
+    /// by MO2, and the game Data folder — so a stem is judged against the whole install rather than the active order
+    /// alone. houseCARL's own patches are skipped: a same-named plugin in a folder we own is the <c>into=</c> extend
+    /// lane's, not a foreign collision.</summary>
+    internal static Hit? Find(Mo2Composition comp, string modsDir, string dataDir, string overwriteDir,
+                              string stem, string canonicalFolder,
+                              IReadOnlySet<string> activePlugins, Func<string, bool> isOwned)
+    {
+        foreach (var ext in PluginFile.Extensions)
```
Avick3110: **Sweeping `.esm` and `.esl` over-refuses, and the refusal's stated reason is false for those hits.**

The fresh record lane only ever mints `<stem>.esp` (`ResolveOutputPath`: `var plugin = freeStem + ".esp";`), and `IsStemFree` correspondingly checks only `stem + ".esp"` against the active order. This sweep checks all three of `PluginFile.Extensions`.

Scenario: a switched-off mod folder holds an inactive `Skyrim Extended Cut.esl`. `patch="Skyrim Extended Cut"` would have written `Skyrim Extended Cut.esp` — a *different* filename, which coexists with the `.esl` perfectly well — but it is refused with "already holds 'Skyrim Extended Cut.esl' … and two plugins cannot share one filename". They do not share one filename; nothing collides. The check refuses a legal write and explains it with a claim that is not true of the case it fired on.

Either restrict the sweep to the extension the lane actually mints, or (if the intent is the shared-BSA-stem hazard) say that instead — the current sentence is the wrong reason.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7713,7 +7714,10 @@ public RiderFolder ResolvePatchModFolder(string? patchName, string? into, string
                 return new RiderFolder(folder, folder, CreatedFresh: false);   // reused — the user owns it; cleanup leaves it
             }
 
-            var newStem = UniqueStem(PatchStem(string.IsNullOrWhiteSpace(patchName) ? defaultStem : patchName!));
+            // The rider's OWN folder parameter, the way its extend refusal already names it — patch= is wrong on a
+            // lane where a bare patch= binds to something else.
+            var newStem = UniqueStem(PatchStem(string.IsNullOrWhiteSpace(patchName) ? defaultStem : patchName!),
```
Avick3110, line 7738: **The rider lane never writes a plugin, so a plugin-filename shadow cannot apply to it.**

Two lines below, `WriteOwnerMeta(newFolder, "(houseCARL output)")` carries the comment "this folder may hold scripts / a .bsa / loose files, **not an .esp**". Yet `UniqueStem` now refuses this lane on the grounds that a foreign folder holds `<stem>.esp`.

Scenario: a switched-off mod folder holds an inactive `My Cool Patch.esp`. `housecarl_compile_script patch_name="My Cool Patch"` is refused with "…two plugins cannot share one filename" — but the call would only have created `houseCARL - My Cool Patch\Scripts\Foo.pex`. No plugin exists on either side of the claimed collision, and the second remedy ("enable '<file>' and target that plugin") is meaningless for a compile.

Same for `bsa_repack`, `decompile_script`, `place`, `nif_set` and `write_seq`. If the intent is folder-name consistency across lanes, the sentence needs to say that; as written the rider lanes get a refusal whose reason does not hold for them.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8190,17 +8194,41 @@ static string PatchStem(string raw)
...
     /// sharing a basename, and mod-folder uniqueness alone never sees a same-named plugin in another mod. into=
-    /// remains the way to grow an existing patch; this is only the fresh path.</summary>
-    string UniqueStem(string stem)
+    /// remains the way to grow an existing patch; this is only the fresh path.
+    /// <para>Neither test sees a plugin the active order is NOT loading, so the stem about to be claimed is also put
+    /// through <see cref="PatchStemShadow"/>, which REFUSES rather than suffixing (#561): a stem suffixed around a
+    /// foreign inactive plugin hides the file the caller may have meant. <paramref name="patchParam"/> is the calling
+    /// lane's own name for its fresh-patch parameter, so the refusal never sends a caller to a parameter their tool
+    /// does not declare.</para></summary>
+    string UniqueStem(string stem, string patchParam)
```
Avick3110: **The check tests the folder stem, so `merge_plugins` still mints the duplicate #561 is about.**

`UniqueStem` is handed the *mod-folder* stem. On the record lane that stem is also the plugin basename, so the check lands on the right name. On `merge_plugins` it is not: the stem is `patchName` or `"<output> merged"` (line 7021-7022), while the plugin actually written is `outName` (line 7025, `Path.Combine(rf.OutputDir, outName)`). `compact_plugin` is the same shape — the output keeps the source's basename (line 6615), not the stem.

So for `merge_plugins`:

- **Missed collision.** `output="MyMerge.esp"` with a switched-off mod folder holding an inactive `MyMerge.esp` still writes a second `MyMerge.esp` — two plugins that can never both be active, silently. The only pre-existing guard is `view.ContainsPlugin(outName)` at line 6921, which sees the *active* order only, which is exactly the gap #561 names. The shadow check ran against `"MyMerge merged"` and saw nothing.
- **False refusal.** Conversely, a foreign folder holding an inactive `MyMerge merged.esp` refuses the call, though nothing named that would ever be written.

The check wants the plugin basename the lane will emit, not the folder stem, on the lanes where those differ.

src/housecarl-mcp/PatchStemShadow.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                var dir = Path.GetDirectoryName(hit.Path);
+                if (dir is not null && isOwned(dir)) continue;
+                return new Hit(hit.Where, file);
+            }
+        }
+        // A folder literally named "houseCARL - <stem>" that carries no marker is a user's, and a fresh patch under
+        // this stem would claim its name. It holds no plugin of the name (the sweep above would have found one), so
+        // the refusal names the folder alone.
+        var canonical = Path.Combine(modsDir, canonicalFolder);
```
Avick3110: **This folder arm is unreachable, and with it the `hit.File is null` half of `Refusal`.**

`Find` is only ever called from `ClaimStem`, and `ClaimStem` only runs on a stem that just passed `IsStemFree`:

```cs
bool IsStemFree(string stem, IReadOnlySet<string> activePlugins)
    => !Directory.Exists(Path.Combine(_modsDir, ModFolderName(stem))) && !activePlugins.Contains(stem + ".esp");
```

That first clause is ownership-blind — it rejects *any* existing `houseCARL - <stem>` folder, the user's included — and it uses the same `_modsDir` + `ModFolderName(stem)` this line rebuilds. So `Directory.Exists(canonical)` is always false here, the `Hit(where, null)` is never constructed, and the first branch of `Refusal` (lines 57-59) is dead code.

Not a runtime misbehaviour, but it is a guard the file's comments present as live coverage, and it will read as covering something on the next change to `IsStemFree`.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8190,17 +8194,41 @@ static string PatchStem(string raw)
...
     {
         var active = ActivePluginBasenames();
-        if (IsStemFree(stem, active)) return stem;
+        var comp = ReadCompositionForShadow();
+        if (IsStemFree(stem, active)) return ClaimStem(stem);
         for (int i = 1; i < 10000; i++)
         {
             var cand = $"{stem}_{i:D3}";
-            if (IsStemFree(cand, active)) return cand;
+            if (IsStemFree(cand, active)) return ClaimStem(cand);
```
Avick3110: **A shadow on an auto-chosen suffix hard-stops the loop instead of advancing to the next one.**

The ruling ("refuse, never suffix around") is about the name *the caller passed*. Inside the loop the name is houseCARL's own invention, and `ClaimStem` throwing there kills the search rather than trying `_002`.

Scenario, with `patch=` omitted so the stem defaults to `Patch`: houseCARL already owns `houseCARL - Patch`, so the loop moves to `Patch_001`; a switched-off foreign folder happens to hold an inactive `Patch_001.esp`; the call now fails outright, even though `Patch_002` is clear and is what the allocator would have produced a moment later. The remedy sentence then tells the caller to "pass `patch=` a name no mod folder already uses" — but they passed nothing, and the name they would have to avoid is one they never chose.

`AnOwnedPatchIsNoShadow_ButTheStemTheSuffixLandsOnIsStillChecked` pins this behaviour, so it is deliberate; flagging it because it converts the default lane's automatic recovery into a hard failure. Refusing on the caller's own stem and continuing the loop on a generated suffix keeps both halves of the ruling.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8190,17 +8194,41 @@ static string PatchStem(string raw)
...
+    /// <para>Neither test sees a plugin the active order is NOT loading, so the stem about to be claimed is also put
+    /// through <see cref="PatchStemShadow"/>, which REFUSES rather than suffixing (#561): a stem suffixed around a
+    /// foreign inactive plugin hides the file the caller may have meant. <paramref name="patchParam"/> is the calling
+    /// lane's own name for its fresh-patch parameter, so the refusal never sends a caller to a parameter their tool
+    /// does not declare.</para></summary>
+    string UniqueStem(string stem, string patchParam)
     {
         var active = ActivePluginBasenames();
-        if (IsStemFree(stem, active)) return stem;
+        var comp = ReadCompositionForShadow();
```
Avick3110: **When the active set degrades to empty, the new check inverts from permissive to refuse-everything — with a false sentence.**

`ActivePluginBasenames` is documented as best-effort and returns an **empty set** rather than throwing when the order cannot be read. That happens without any exception: a missing or unwritten `loadorder.txt` makes `ReadLoadOrderNames` add a warning and return nothing, so `Mo2LoadOrder.Build` yields zero `OrderedPaths` and the `catch` never fires. Meanwhile `ReadCompositionForShadow` reads `modlist.txt` fine and returns a non-null `comp`, so the sweep runs at full strength with *no knowledge of what is active*.

With an empty `active`, `PatchStemShadow.Find`'s `if (activePlugins.Contains(file)) continue;` never skips anything, so every same-named plugin anywhere — including one the order **is** loading, out of an enabled mod folder — becomes a "shadow".

Scenario: `loadorder.txt` is missing (MO2 hasn't re-sorted after a profile switch) and an enabled, active mod holds `Foo.esp`. `patch="Foo"` used to fall back to folder-only uniqueness and write `Foo.esp` / suffix; it now hard-fails with "…already holds 'Foo.esp', **which the load order is not loading**" — a statement that is the opposite of true, on a call that previously worked. That contradicts `ReadCompositionForShadow`'s own doc ("a call that worked before never fails because the extra check could not run"): the guard covers the composition read but not the active-set read the check equally depends on.

Skipping the shadow sweep when `active.Count == 0` would restore the documented degradation.

src/housecarl-mcp/PatchStemShadow.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        foreach (var ext in PluginFile.Extensions)
+        {
+            var file = stem + ext;
+            // An active plugin of this name is the suffix loop's case, not a silent shadow — leave it alone.
+            if (activePlugins.Contains(file)) continue;
+            foreach (var hit in Mo2LoadOrder.LocatePlugin(comp, modsDir, dataDir, overwriteDir, file))
+            {
+                var dir = Path.GetDirectoryName(hit.Path);
+                if (dir is not null && isOwned(dir)) continue;
```
Avick3110: **`isOwned` can throw straight out of a write that would previously have succeeded.**

`IsHouseCarlOwned` opens `meta.ini` with `File.ReadLines` and catches nothing, so an `IOException` (the file locked while MO2 is writing it — the common case, since MO2 rewrites `meta.ini` on refresh) or an `UnauthorizedAccessException` propagates through `Find` → `ClaimStem` → `UniqueStem` and aborts the whole write.

`LocatePlugin` swallows a bad candidate folder itself, and `ReadCompositionForShadow` wraps its own read, so this is the one unguarded IO on the new path — and it fires only on the folders that actually hold the name, i.e. exactly when the check matters. The declared contract ("a call that worked before never fails because the extra check could not run") does not hold for it.

A `try`/`catch` around the ownership probe, treating an unreadable `meta.ini` as un-owned or as no-hit, restores it.

src/housecarl-mcp/PatchStemShadow.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// another name, or enable the plugin already there — because the caller who reached this may have meant either.
+    /// <paramref name="patchParam"/> is the calling lane's OWN parameter spelling, since the rider lanes name their
+    /// fresh folder with something other than <c>patch=</c>.</summary>
+    internal static string Refusal(string stem, Hit hit, string patchParam)
+        => hit.File is null
+            ? $"cannot create a fresh patch named '{stem}': {hit.Where} already exists and was not created by "
+              + $"houseCARL, so a fresh patch would claim its name — pass {patchParam}= a name no mod folder already uses."
+            : $"cannot create a fresh patch named '{stem}': {hit.Where} already holds '{hit.File}', which the load "
+              + "order is not loading, and two plugins cannot share one filename — pass "
+              + $"{patchParam}= a name no mod folder already uses, or enable '{hit.File}' and target that plugin if "
```
Avick3110: **low** — the remedy says "a name no mod folder already uses", but the sweep also reaches the overwrite folder and the game `Data` folder, neither of which is a mod folder, so a hit rendered as `overwrite already holds 'Patch.esp'` or `game Data already holds 'Patch.esp'` ends with an instruction that misdescribes what the caller has to avoid — and the CHANGELOG entry likewise promises "one sentence naming the mod folder it found". It matters because the overwrite hit is the likeliest one on a real install (xEdit and Synthesis drop output there) and the caller is told to check the wrong place. Either say "a name nothing on the install already uses", or key the noun off the layer the hit came from.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7630,7 +7630,8 @@ string ResolveOutputPath(string? patchName, string? into, out bool extend, out b
 
             extend = false;
             var baseStem = PatchStem(string.IsNullOrWhiteSpace(patchName) ? "Patch" : patchName!);
-            var freeStem = UniqueStem(baseStem);
+            // Every record lane that reaches here declares patch=, so that is the spelling the shadow refusal names.
+            var freeStem = UniqueStem(baseStem, "patch");
```
Avick3110: **low** — `UniqueStem` is not the only fresh-write stem allocator: `CreatePlugin` (this file, ~6320-6331) builds `ModFolderName(stem)` and `stem + ".esp"` by hand and screens the name against `view.ContainsPlugin` (the active order) and the canonical owned folder only — the same two tests #561 says are blind to an inactive plugin in a foreign mod folder. It matters more there than on the patch lane, because that refusal's own words are that a header-only trigger's "exact basename is what makes the trigger resolve", so the shadow silently breaks the thing the tool exists to make. Either route it through the shared check or say in the PR that `plugin_name=` is out of scope and file it.

src/housecarl-mcp-tests/PatchStemShadowTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        Assert.Contains("My Cool Patch.esp", o.Error);
+        Assert.Contains("patch=", o.Error);
+        // Refused, never renamed: no patch folder of either spelling was left behind.
+        Assert.Empty(Directory.EnumerateDirectories(w.ModsDir, "houseCARL - My Cool Patch*"));
+    }
+
+    /// <summary>The same name with no foreign plugin behind it is an ordinary write — the check must not refuse a
+    /// stem merely because a folder of some similar name exists.</summary>
+    [Fact]
+    public void TheSameStemProceedsWhenNoForeignFolderHoldsThePlugin()
```
Avick3110: **low** — standards/TESTING.md: "A test must fail before the fix and pass after. If it cannot fail, it is not a test; delete it." The PR body concedes this one passes before and after and keeps it as an over-refusal guard. It matters because the standard is the repo's own line on exactly this trade, so the arm should either go or be argued to Aaron rather than settled in the PR body. If it stays, note that the guard only covers the record lane, and the over-refusals actually reachable here are on the rider lanes (`ResolvePatchModFolder`) and on `.esm`/`.esl`, which no test touches.

src/housecarl-mcp/PatchStemShadow.cs
```diff
@@ -0,0 +1,64 @@
...
+        foreach (var ext in PluginFile.Extensions)
+        {
+            var file = stem + ext;
+            // An active plugin of this name is the suffix loop's case, not a silent shadow — leave it alone.
+            if (activePlugins.Contains(file)) continue;
+            foreach (var hit in Mo2LoadOrder.LocatePlugin(comp, modsDir, dataDir, overwriteDir, file))
+            {
+                var dir = Path.GetDirectoryName(hit.Path);
+                if (dir is not null && isOwned(dir)) continue;
+                return new Hit(hit.Where, file);
```
Avick3110: **low** — the ownership skip is keyed on the wrong axis, and the residual the PR calls "houseCARL's own collision with a different remedy" is the same failure #561 describes. A houseCARL patch whose folder the user renamed in MO2 keeps its `meta.ini` marker, so `isOwned` skips it; if its `<stem>.esp` is not active, `IsStemFree` also passes (the canonical folder name is gone), and `patch=<stem>` mints a second `<stem>.esp` beside it — two plugins that cannot both be active, which is the whole of the issue. What decides the hazard is the filename, not who owns the folder. Worth saying in the PR that the residual is the identical failure with a different remedy (`into=`), not a lesser one, so Aaron's call is priced right.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in d22bd7c6: the check is keyed on the plugin file a lane writes, so compact_plugin and bsa_extract take no shadow refusal at all and merge_plugins names output=, the parameter that actually moves its filename. Both compact_plugin and merge_plugins keep naming: null, which is still correct under RiderNaming's documented contract (their into= can never be non-empty), because naming no longer feeds the shadow sentence.

src/housecarl-mcp/PatchStemShadow.cs
Avick3110: Folded in d22bd7c6: the sweep is one filename now, the one the lane will write, so a foreign .esm no longer refuses an .esp write; AForeignEsmOfTheSameStemIsNoShadowForTheEspTheLaneWrites covers it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in d22bd7c6: a lane that writes no plugin passes no shadow target, so compile_script, decompile_script, bsa_repack, bsa_extract, place, nif_set and write_seq take no plugin-shadow refusal, and 842c670e takes the sentence back off their parameters.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in d22bd7c6: merge_plugins hands the check outName rather than the folder stem, so the missed collision refuses and the false one does not; AMergeIsRefusedOnTheOutputFilenameItWritesNotItsFolderStem covers it. compact_plugin is the other way and I have overturned it there: its output deliberately keeps the source's exact basename so external masters still resolve, and an inactive source is a documented target, so checking that filename would refuse the tool's own purpose. It takes no shadow check, and the changelog says so.

src/housecarl-mcp/PatchStemShadow.cs
Avick3110: Folded in d22bd7c6: the canonical-folder arm, the Hit type and the file-less branch of Refusal are deleted; Find returns the PluginFileHit it found.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in d22bd7c6: a shadow on a suffix houseCARL invented steps to the next suffix, and only a stem the caller passed is refused. AShadowOnAStemTheCallerDidNotChooseStepsToTheNextSuffix covers it. The one exception is a lane whose filename does not follow the stem (merge_plugins' output=), where every suffix would land on the same file.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in d22bd7c6: with no active plugin known the sweep does not run at all, so the pre-existing folder and active-order uniqueness stand — the same degradation the composition read already documents, rather than a third behaviour.

src/housecarl-mcp/PatchStemShadow.cs
Avick3110: Folded in d22bd7c6: the ownership axis is gone with finding 12, so there is no meta.ini read on this path to throw.

src/housecarl-mcp/PatchStemShadow.cs
Avick3110: Folded in d22bd7c6 and 9d3248a0: the remedy says "a name nothing on your install already uses", and the layer label already names which of the three it was; the changelog line says the same.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in d22bd7c6: create_plugin takes the same check before it creates the folder, refusing on plugin_name= rather than renaming, since the exact basename is what makes the trigger resolve.

src/housecarl-mcp-tests/PatchStemShadowTests.cs
Avick3110: Folded in 2f93c4fe: that arm is deleted. Three replace it, each failing before its own fold and passing after — a foreign .esm not refusing the .esp write, the auto-suffix stepping past a shadow, and a merge refused on its output filename naming output=.

src/housecarl-mcp/PatchStemShadow.cs
Avick3110: Folded in d22bd7c6: the ownership axis is dropped entirely. Any folder holding the exact filename the call will write is a shadow, houseCARL's own included; the fresh folder is created after the check and holds nothing, so it can never be its own shadow. The residual you name is closed rather than repriced.

Avick3110:
Rebased onto the merged description wave (tip `cd6b3139`). All eight commits replayed in order, none squashed; the net diff is byte-identical in shape to before the rebase (9 files, +246/-19).

Two files conflicted:

- `src/housecarl-mcp/ForwardTools.cs` — conflicted twice, once in each of the two description commits. The wave had dropped the "Mutually exclusive with into= and in_place=" sentence from `patch=` here (only here; the same sentence still stands on `apply`, `create` and the other lanes). Kept the wave's shortened text and re-applied the shadow clause onto `patch=` where it had been, so the sentence the wave removed is not resurrected.
- `plugin/CHANGELOG.md` — the wave replaced the per-tool `housecarl_records` entry this PR's entry had been written above with the consolidated "Every tool description now fits inside the 2,048 characters" entry. Kept both, in the file's order: this PR's shadow entry first under `## Unreleased`, then the wave's entry. The edit to #552's line further down applied cleanly.

Everything else auto-merged: `ApplyTools.cs`, `CopyTools.cs`, `CreateTools.cs`, `WriteTools.cs`, `BsaTools.cs`, `CompileTools.cs`, `DecompileTools.cs`, `NifTools.cs`, `PlaceTools.cs`, `SeqTools.cs`, `LoadOrderService.cs`, `PatchStemShadow.cs`, `PatchStemShadowTests.cs`. `PublishedDescriptionBoundTests.cs` was not touched by this PR, so the wave's version stands unchanged.

Re-run from the worktree after the rebase:

```
dotnet build housecarl.sln -c Release   -> 0 errors
                                        -> 1623 passed, 0 failed
```

`PublishedDescriptionBoundTests` and `PatchStemShadowTests` pass — every clause this PR adds is on a parameter, so no tool description grew.

`ci-all` was not run locally (#570); CI runs it.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8191,17 +8214,53 @@ static string PatchStem(string raw)
...
+            return false;
+        }
+    }
+
+    /// <summary>The profile composition the shadow sweep walks, read once per fresh write. Best-effort in the same way
+    /// <see cref="ActivePluginBasenames"/> is: an unreadable profile yields null and the pre-existing folder and
+    /// active-order uniqueness, so a call that worked before never fails because the extra check could not run.</summary>
+    Mo2Composition? ReadCompositionForShadow()
+    {
+        try { return Mo2LoadOrder.ReadComposition(_profileDir); }
```
Avick3110: **The sweep's "unreadable profile" fallback never actually engages.**

`ReadCompositionForShadow` documents that an unreadable profile yields `null` and the check is skipped. But `Mo2LoadOrder.ReadComposition` doesn't throw on a missing profile file — `ParseModlist`, `ParsePlugins` and `ReadLoadOrderNames` each return early with a warning, so the composition comes back non-null and empty. The `try/catch` only fires on a real IO error mid-`ReadAllLines`.

Concrete failure: `modlist.txt` is missing or truncated, but `Skyrim.esm` and friends still resolve from the game `Data` folder, so `Mo2LoadOrder.Build` still returns active plugins and `ActivePluginBasenames()` is non-empty — the `active.Count == 0` guard does not fire either. The composition then has `EnabledMods` and `DisabledMods` both empty, so `UnlistedModFolders` yields *every* directory under ModsDir. A `patch="Foo"` write against an install whose enabled mod `CoolMod` holds a genuinely loaded `Foo.esp` is refused with:

> cannot write 'Foo.esp': mod 'CoolMod' (UNLISTED) already holds 'Foo.esp', which the load order is not loading…

Both the `(UNLISTED)` label and "the load order is not loading" are false, and this is exactly the case the PR body promises cannot happen ("a call that worked before never fails because the extra check could not run").

The guard wants to be on whether the composition is *usable*, not on whether the read threw — e.g. skip the sweep when `modlist.txt` is absent, or when `comp.EnabledMods.Count == 0` while the order has active plugins.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6329,9 +6330,16 @@ public WritePatchBuilder.CreatePluginOutcome CreatePlugin(string pluginName, boo
             if (Directory.Exists(folder))
                 return WritePatchBuilder.CreatePluginOutcome.Fail(
                     $"a houseCARL output folder '{ModFolderName(stem)}' already exists — houseCARL won't auto-rename a header-only plugin (its exact basename is what makes the trigger resolve). Remove that folder in MO2, or choose a different name.");
+            // (c) a plugin of this exact filename sits somewhere the order is NOT loading — the same shadow the fresh
+            //     patch lanes take (#561), and it bites harder here: the exact basename is what makes a trigger
+            //     resolve, so a second file of that name is the thing this tool exists to avoid.
+            var plugin = stem + ".esp";
```
Avick3110, line 6339: **On this tool the shadow check is narrower than the collision check right above it.**

Check (a) at line 6323 loops `PluginExts` — `.esp`, `.esm`, `.esl` — against the active order, on the stated ground that a header-only trigger needs a unique *basename*. The new check (c) tests only `stem + ".esp"`.

So: a switched-off mod folder holds `MyTrigger.esm`. `create_plugin(plugin_name="MyTrigger")` sails past (c) and writes `MyTrigger.esp`. Had that same `MyTrigger.esm` been active, (a) would have refused outright — "a plugin named 'MyTrigger.esm' is already active … a header-only trigger needs a UNIQUE basename". Whatever binds to the basename now has two files on the install to resolve against, which is the state (a) exists to prevent.

The `.esp`-only scope is right for the record lanes (`AForeignEsmOfTheSameStemIsNoShadowForTheEspTheLaneWrites` justifies it: two different filenames coexist fine). It just doesn't match this tool's own premise. Either (c) should sweep the three extensions here, or (a)'s three-extension loop is the one that wants explaining.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7631,7 +7642,10 @@ string ResolveOutputPath(string? patchName, string? into, out bool extend, out b
 
             extend = false;
             var baseStem = PatchStem(string.IsNullOrWhiteSpace(patchName) ? "Patch" : patchName!);
-            var freeStem = UniqueStem(baseStem);
+            // Every record lane that reaches here declares patch= and writes "<stem>.esp", so that is the file the
+            // shadow check tests and the spelling its refusal names.
+            var freeStem = UniqueStem(baseStem, !string.IsNullOrWhiteSpace(patchName),
```
Avick3110: **`stemFromCaller` is true for a stem the caller never spelled as `patch=`.**

It is derived from whatever `patchName` argument `ResolveOutputPath` received — but the closure-copy lane passes a coalesced value:

```csharp
// line 5643
outPath = ResolveOutputPath(patchName ?? (into is null ? newEditorid?.Trim() : null), into, …)
```

So with `patch=` omitted, the record's EditorID becomes the base stem *and* is flagged caller-chosen, which makes a shadow on it a hard refusal rather than a step to `_001`.

The scenario is the one `LocatePlugin`'s own doc calls realistic: the user is standalone-izing a follower out of a mod they have switched **off** in MO2, whose plugin is `MyFollower.esp`, and calls `housecarl_copy` with `new_editorid="MyFollower"` and no `patch=`. `UniqueStem` gets `stem="MyFollower"`, `stemFromCaller: true`, `Find` hits the disabled donor folder, and the whole copy is refused with "pass `patch=` a name nothing on your install already uses" — a parameter that was not in the call, and whose description (`CopyTools.cs:47`) is the only place the new refusal is documented.

`AShadowOnAStemTheCallerDidNotChooseStepsToTheNextSuffix` establishes that an unspecified `patch=` steps rather than fails; this lane is the one place that rule quietly doesn't hold. Either pass `stemFromCaller: !string.IsNullOrWhiteSpace(patchName)` from the caller's *own* `patch=` (so an editorid-derived stem steps), or let the lane state its own parameter the way `merge_plugins` states `output=`.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8191,17 +8214,53 @@ static string PatchStem(string raw)
...
+    /// lane's default, which is what makes a shadow on it refusable.</para></summary>
+    string UniqueStem(string stem, bool stemFromCaller, PatchStemShadow.Target? writes)
     {
         var active = ActivePluginBasenames();
-        if (IsStemFree(stem, active)) return stem;
+        // The sweep can only tell a shadow from the active plugin the suffix loop already dodges if it knows what the
+        // order loads. With no composition, or no active plugin known at all, it does not run and the pre-existing
+        // folder and active-order uniqueness stand — the same best-effort degradation both reads already document,
+        // so a call that worked before never fails because the extra check could not run.
+        var comp = active.Count == 0 ? null : ReadCompositionForShadow();
```
Avick3110: **The composition is read even when the lane can never use it.**

`Takeable` discards `comp` immediately at `writes is not { } w`, but the read above is unconditional. Every lane that passes no shadow target — `compile_script`, `decompile_script`, `bsa_repack`, `bsa_extract`, `place`, `nif_set`, `write_seq`, `compact_plugin` — now re-parses `modlist.txt`, `plugins.txt` and `loadorder.txt` per fresh write for a value that is thrown away. On a cold resolver `ActivePluginBasenames()` on the line above has just parsed the same three files through `Mo2LoadOrder.Build`, so that is two full profile parses where the lane needs zero.

```suggestion
        var comp = active.Count == 0 || writes is null ? null : ReadCompositionForShadow();
```

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 284485dc. You are right that the try/catch never fires for a missing modlist.txt — ReadComposition returns empty mod lists, every folder under ModsDir then reads as UNLISTED, and a loaded plugin gets refused as one the order is not loading. ReadCompositionForShadow now guards on whether the composition is usable: a profile naming no mod in either list (missing modlist.txt, or truncated to nothing) yields null and the sweep is skipped, leaving folder and active-order uniqueness. Pinned by AProfileNamingNoModAtAllSkipsTheShadowSweep, which deletes modlist.txt while the order still loads a plugin out of the game Data folder, so the active.Count guard is not the one doing the work; it fails before the change with exactly the sentence you quoted. One honest limit, and a decision if you want it to go further: a modlist.txt truncated to a PARTIAL list still reads as usable, so a mod whose line was lost still reads as UNLISTED. The doc comment says so rather than claiming otherwise. The discriminant that would close that gap is plugins.txt — a ticked filename means the profile intends to load that name, so "the load order is not loading" would be false — but it also weakens a real refusal (a ticked plugin whose mod folder is switched off is genuinely not loading and IS a shadow), so I left it out rather than change what shadow means on my own judgement.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 4002ec6a. Check (c) now sweeps PluginExts like (a) right above it, and for (a)'s own reason: a header-only trigger resolves by BASENAME, so a switched-off MyTrigger.esm is precisely the second file (a) exists to prevent. The record lanes stay on .esp alone — AForeignEsmOfTheSameStemIsNoShadowForTheEspTheLaneWrites still pins that, and the two premises are now each stated where they hold. PatchStemShadow.Refusal took an optional found= (the file actually sitting there, when it is not the file being written) and an optional clash clause, so the sentence names both names and the create lane supplies its own reason; every other caller renders exactly as before. New test AHeaderOnlyCreateIsRefusedByAnyExtensionSharingItsBasename. plugin_name='s parameter description now says .esp/.esm/.esl outright; the tool description is untouched, so the cap test stays green.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 960ef01b, the first option. ResolveOutputPath gained an optional stemFromCaller, defaulting to reading it off patchName as before; the closure-copy lane passes its own patch= instead, so a stem coalesced from new_editorid is no longer flagged caller-chosen and a shadow on it steps to the next suffix. New test AShadowOnAnEditoridDerivedStemStepsToTheNextSuffix builds exactly your scenario on TwoDisabledDonorsWorld — new_editorid="Donor", no patch=, against the switched-off donor holding Donor.esp — and lands on houseCARL - Donor_001; it fails before the change with the refusal naming patch=. CopyTools.cs:47 is unchanged because it is now true as written: patch= is refused only when the caller actually passed it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in dc77d9d4, your suggestion as written: the read is now guarded on writes as well, so a lane passing no shadow target pays no profile parse. No test — the change is a read that was thrown away, with no observable behaviour either side of it, and the only way to pin it would be to count file reads, which is a guard around a mistake rather than a test of the code. The build and full suite are green.

PR #592, opened 2026-09-06
Bring the remove tool description under the 2,048-character cap

Avick3110:
`housecarl_remove` published a 2,378-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the removal-legality rule, the all-or-nothing refusal, the lane grammar, the consent handshake and the response shape were all being described in a part of the text the model never saw. Nothing is deleted here: each paragraph moves into the description of the parameter it is about — the legality and `@file` rules to `formids`, the patch lane to `into`, the in-place trade-off to `in_place`, the confirmation prompt and the absence verify to `acknowledge`, the response shape and the epoch stamp to `format` — and where the parameter already said the same thing the description's copy is dropped as a duplicate rather than pasted in twice. The description keeps the one-sentence purpose, the `ONE surface:` composition sentence, the write model (what dropping this file's override means, and the automatic master prune), one line per axis naming its parameters, and the pointers to `housecarl_apply` and `housecarl_records`. `housecarl_remove` comes off the bound test's `StillOversized` list, which now names six tools.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_remove` description | 2,378 | **1,223** |
| Top-level parameter descriptions | 1,563 | 2,618 |
| All schema descriptions (parameters + nested DTO members) | 1,563 | 2,618 |

The description shed 1,155 characters and the schema gained 1,055; the 100-character difference is text the description was repeating from a parameter that already carried it (the set-valued sentence, the `["@<absolute path>"]` spelling, "it must be a patch houseCARL created", the in-place rewrite and no-backup sentence, "the FIRST in-place write to a given plugin", "that consent covers touching your original ONLY", and the two mutually-exclusive statements the lanes already carry).

`housecarl_remove` has no nested DTO members, so the last two rows are the same number.

## Harvest

|---|---|
| "Remove WHOLE records — a literal drop-from-plugin, NOT a flag-as-deleted stub." | kept in description |
| "The counterpart to `housecarl_apply`: where apply ADDS an override into a patch, this drops one OUT of it." | kept in description |
| "ONE surface: what to drop (formids=) x WHERE from (the LANE: into= a houseCARL patch \| in_place=\"X.esp\") x how it reads back (TRANSPORT)." | kept in description |
| "WHAT CAN BE REMOVED. Only a record the named file ITSELF carries …" | `formids`, as "Only a record the LANE FILE itself carries may be named" — stated lane-neutrally, because it holds on both lanes |
| "… — one houseCARL created, or an override the patch accumulated via a prior apply/forward into= it." | `into`, which is the lane those two provenances describe (the `in_place` lane takes any active plugin's own records, and `in_place` already says so) |
| "You CANNOT remove a record that lives in a master or another mod; you can only drop THIS file's override of it, which makes the load-order winner revert (the file stops touching that record)." | kept in description — it is what a removal MEANS, not a rule about one parameter |
| "A FormID the file doesn't carry is REFUSED loud with nothing written (Q3)." | `formids`, merged with the ALL-OR-NOTHING sentence below into one statement rather than two adjacent ones saying the same thing |
| "Reaches records in ANY group (cells, placed references, dialogue, navmesh)." | `formids` |
| "formids= is SET-VALUED — drop many records in ONE re-serialize" | `formids` (already said there in the same words) |
| "(one is a set of one)" | `formids` |
| "ALL-OR-NOTHING (Q3): if ANY target isn't carried, the whole call is refused with per-record reasons and NOTHING is written." | `formids` |
| "It also accepts [\"@<absolute path>\"] — the same list from a file" | `formids` (already said there) |
| "one FormID per line" | `formids` |
| "LANE — where the removal happens." | kept in description, as the LANE axis line |
| "into='<a houseCARL patch's filename>' drops the records from that patch" | `into` (already said there) |
| "(the same name you pass to `housecarl_apply`'s into=)" | `into` |
| "it must be a patch houseCARL created." | `into` (already said there) |
| "in_place='<plugin filename>' is the opt-in lane for ANY existing active plugin, including one houseCARL didn't author" | `in_place` (already said there; its "that existing active plugin" gains the ANY) |
| "your ORIGINAL file is rewritten — no houseCARL backup or undo (keep your own)." | `in_place` (already said there; it gains "(keep your own)") |
| "The FIRST in-place write to a given plugin returns a confirmation prompt (re-call with acknowledge=true)" | `acknowledge`, which had the "needed only on the FIRST in-place write" half; it gains the prompt mechanic, in the wording `housecarl_apply` settled on |
| "that consent covers touching your original ONLY — it NEVER skips the absence verify" | `acknowledge` (already said there) |
| ", which confirms on the re-opened file that every record you dropped is actually gone." | `acknowledge` |
| "Exactly one lane per call." | kept in description, on the LANE axis line, with the reason the tool has no `patch=` lane at all |
| "Unused masters are pruned automatically: if a removed record held the file's last reference to a master, that master drops from the header on the re-write." | kept in description — the write model, and it belongs to no parameter |
| "Returns what was removed, the remaining masters, and how many records remain (0 = the file is an inert shell)." | `format`, the TRANSPORT parameter that decides how that document is rendered; both renders carry all three facts |
| "Every response carries epoch=<hex> — the identity of the index build this removal's master context came from." | `format`, with both spellings named: `epoch=<hex>` on `'text'`, an `'epoch'` member on `'json'` (checked against `WriteSentences.Epoch` and `JsonWire.WriteEpoch`) |
| "To remove a list ENTRY (a keyword, an item, a leveled-list line) rather than a whole record, use `housecarl_apply` with op='Remove' instead." | kept in description |
| "Read first with `housecarl_records`." | kept in description |

The description gained one thing it did not carry before: the three axis lines (WHAT / LANE / TRANSPORT), naming every parameter, which is now the description's only index of them — `format=` and `max_chars=` were reachable from the old text only as the bare word TRANSPORT.

## #377 checked, nothing to correct

#377 is about `housecarl_remove_record`, the 1.x tool, which declared `patch=` while the shared extend resolver's ambiguous and multi-plugin refusals told the caller to "pass into=". The 2.0 `housecarl_remove` declares `into=`, so that mismatch does not reach this surface: no sentence in the description or in any parameter description names a parameter the tool does not declare (the six are `formids`, `into`, `in_place`, `acknowledge`, `format`, `max_chars`; the description also names `housecarl_apply`'s `into=` and `op='Remove'`, both correct for that tool). The `into` description's own ambiguity remedy — "for two patches sharing a filename, pass the mod-folder name here instead" — names `here`, the parameter it sits on. Nothing changed for #377; the issue's fix is in the refusal text in `LoadOrderService.cs`, which this PR does not touch.

## Stale pointers

`grep -rn "tool description" src/ --include=*.cs` finds no comment or refusal naming text this PR moved. Nothing outside `RemoveTools.cs` reads the description; the generator probes that name `RemoveTools.Remove` drive the tool and assert on its output, not on its description string.

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`: the `ToolNames.Remove` line comes out of `StillOversized`, and nothing else changes. Ran the class with the line gone and the description untouched — `EveryPublishedDescriptionFitsTheClientsCut(tool: "housecarl_remove")` failed at 2,378, which is the first commit here. It passes at 1,223 after the second.

## What was run

```
    1620 passed, 0 failed
    64 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. No `plugin/CHANGELOG.md` line — the records PR's entry covers this wave.


src/housecarl-mcp/RemoveTools.cs
```diff
@@ -16,38 +16,28 @@ public static class RemoveTools
...
             bool acknowledge = false,
-        [Description("TRANSPORT: 'text' (default) | 'json' (the same data, machine-readable, accounting in-band).")]
+        [Description("TRANSPORT: 'text' (default) | 'json' (the same data, machine-readable, accounting in-band). Either way the response states what was removed, the remaining masters, and how many records remain (0 = the file is an inert shell). Every response carries the epoch stamp — the identity of the index build this removal's master context came from — spelled epoch=<hex> on 'text', and as an 'epoch' member on 'json'.")]
```
Avick3110: The epoch sentence overstates what the pre-capture refusals carry. `housecarl_remove` answers six refusals before any build is consulted — both lanes named, no lane named, `acknowledge=` without `in_place=`, empty `formids=`, an artifact path in `formids=`, and the malformed-FormID refusal `LoadOrderService.RemoveRecords` returns with no `Stamp` — and none of them has an epoch to state. On `'text'` that is literal absence: `WriteTools.RenderRemoval` appends `WriteSentences.Epoch(o)`, which is `ReadTools.EpochLine` returning `""` for a null stamp, and the `Refuse(...)` path returns a bare `"error: …"` string that never reaches a render at all — so no `epoch=<hex>` line. On `'json'` the member is present but null (`WriteNullable` writes `w.WriteNull`), not the hex identity the sentence promises. `JsonWire.RenderError`'s own docstring states the contract this contradicts: "pre-capture validation refusals omit it because they consulted no build."

Concrete: a caller passes `formids=[]` with `format='json'`, reads `epoch` expecting a build id per this description, and gets `null`; the same call on `'text'` gets no epoch line whatsoever.

The claim is inherited from the old description, so this is not a regression — but the sentence is being restated here more specifically (it now names both spellings), and this is the PR whose job is description accuracy. "Every response that reached the index build carries the epoch stamp" would be true as written.

src/housecarl-mcp/RemoveTools.cs
```diff
@@ -16,38 +16,28 @@ public static class RemoveTools
...
+         "TRANSPORT — format= | max_chars=.\n\n" +
          "To remove a list ENTRY (a keyword, an item, a leveled-list line) rather than a whole record, use " +
          ToolNames.Apply + " with op='Remove' instead. Read first with " + ToolNames.Records + ".")]
     public static string Remove(
         LoadOrderService svc,
-        [Description("The record(s) to drop, each 'XXXXXX:Plugin.esp' (6 hex digits, the defining master's filename) — set-valued, so many records drop in ONE re-serialize. Also accepts [\"@<absolute path>\"] to read the same list from a file.")]
+        [Description("The record(s) to drop, each 'XXXXXX:Plugin.esp' (6 hex digits, the defining master's filename) — SET-VALUED, so many records drop in ONE re-serialize (one is a set of one). Only a record the LANE FILE itself carries may be named; it reaches records in ANY group (cells, placed references, dialogue, navmesh). ALL-OR-NOTHING (Q3): a FormID the file doesn't carry is REFUSED loud — if ANY target isn't carried, the whole call is refused with per-record reasons and NOTHING is written. Also accepts [\"@<absolute path>\"] to read the same list from a file, one FormID per line.")]
```
Avick3110: low. The description now points at `formids=` for "the rule for which ones may be named", and this is the whole rule it states — carried by the lane file, any group, all-or-nothing on a FormID the file doesn't carry. The engine has a second naming rule that is not here: `DetachOwnedChildren` (`src/housecarl-core/WritePatchBuilder.cs`, run on both remove lanes) refuses a target that is a singular owned child — a cell's Landscape, a worldspace's TopCell — when the records under it were not named too, with "Name them in the same call so the removal reports every record it drops". That record IS carried, so a caller reading this parameter has no reason to expect the refusal, and the surface already promises it elsewhere: `OwnedChildLifecycleDrivenTests.TheFieldLaneRemoveRefusalSaysToNameTheRecordsUnderTheChild` asserts the field-lane refusal sends the caller to `housecarl_remove` saying it "refuses until every record it would drop is named". Inherited rather than introduced — the old description did not carry it either — but this PR is what makes `formids=` the authority, so a clause for it belongs here.

src/housecarl-mcp/RemoveTools.cs
Avick3110: Folded in 7639dce6: format= now says every response answered from a build carries the epoch stamp, and that a refusal that consulted no build carries none, keeping both spellings.

src/housecarl-mcp/RemoveTools.cs
Avick3110: Folded in 7639dce6: formids= now carries the owned-child clause in the engine refusal's own words — name them in the same call so the removal reports every record it drops.

PR #595, opened 2026-09-06
Bring the create tool description under the 2,048-character cap

Avick3110:
`housecarl_create` published a 6,511-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the record-spec grammar, the nesting rules, the `'@editorid'` sibling references, the lane rules and the transport rules were all being described in a part of the text the model never saw. Nothing is deleted here: each paragraph moves into the description of the parameter it is about — the `records=` element members for the spec grammar, `parent`/`value` for nesting and same-call references, `into`/`in_place`/`acknowledge` for the lane, `readback`/`format` for transport — and where the parameter already said the same thing the description's copy is dropped as a duplicate rather than pasted in twice. The description keeps the one-surface sentence, the whole-call all-or-nothing rule, one line per axis naming its parameters (with `records=`'s own `"@<absolute path>"` spelling, which is what a `JsonElement` input binds), the composition rule, and the pointers to the other write tools. `housecarl_create` comes off the bound test's `StillOversized` list, which now names six tools.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_create` description | 6,511 | **1,225** |
| Top-level parameter descriptions | 2,509 | 4,458 |
| All schema descriptions (parameters + nested DTO members) | 9,314 | 13,069 |

The description shed 5,286 characters and the schema gained 3,755; the 1,531-character difference is text the description was repeating from a parameter that already carried it (the records= shape and set-of-one rule, the parent-before-children rule, the `record_type` catalog/signature sentence, most of `patch=`, the `into=` opening, the in-place lane opening, the consent sentence, `format='json'`, `max_chars`, and the `@file` half `records` already stated).

## Harvest

|---|---|
| "Create BRAND-NEW records (new FormIDs) and write them to a NEW patch plugin (originals untouched by default) — the net-new authoring tool, the companion to `housecarl_apply` (which edits EXISTING records)." | kept in description |
| "ONE surface: what to author (records=) x WHERE it lands (the LANE: a new patch \| into= an existing one \| in_place=\"X.esp\") x how it reads back (TRANSPORT)." | kept in description |
| "RECORDS. records= is the spec list — {record_type, editorid, ops?, parent?, collection?, grid?} each; ONE record is a set of one." | `records` (already said there in the same words) |
| "record_type is a catalog name ('Keyword', 'Spell', 'LeveledItem', 'DialogTopic') or a 4-char signature ('KYWD')." | `records[].record_type` (already said there; it gains 'LeveledItem', the one example its own list lacked) |
| "editorid is REQUIRED — the EditorID the record is referenced by (in SkyPatcher/SPID, in xEdit); choose a clear, prefixed name." | `records[].editorid`, which had the first half; it gains where the EditorID is referenced from and the naming advice |
| "Any flat top-level record is fair game — a keyword, spell, perk, magic effect, faction, armor, weapon, leveled list, global, quest — as is a nested one (see NESTING)." | `records[].record_type`; the "see NESTING" pointer is repointed to "see parent=", the parameter that grammar now lives on |
| "For an ABSTRACT record group name the CONCRETE subtype directly ('GlobalFloat'/'GlobalInt'/'GlobalShort', 'GameSettingFloat'/'GameSettingInt'/'GameSettingString') — that is how a global variable or game setting is created." | `records[].record_type` (its weaker two-example version replaced) |
| "Each new FormID is auto-allocated in the patch's own 0x800+ range and REPORTED BACK: it is how you reference the record afterwards — to make ANOTHER record point at it, call this tool or `housecarl_apply` again with into='<this patch>' and that FormID as the value." | `records`, with "the patch's own" widened to "the receiving plugin's own" so the sentence is true on the in-place lane too (the 0x800 floor is enforced for every allocation); `in_place` keeps the lane-specific half below |
| "OPS set the new record's fields — the same shape `housecarl_apply` takes MINUS formid (the record has no id yet): {field_path, op?, value?, key?, values?, entries?, compose?, composes?}." | `records[].ops`, which had the shape and the minus-formid rule; it gains "(the record has no id yet)" |
| "e.g. ops=[{field_path:'Name', value:'My Spell'}, {field_path:'EffectList', op:'Add', compose:{...}}]" | `records[].ops` |
| "compose builds a modeled struct (a leveled-list entry, an effect, a condition row, or a polymorphic element by its CONCRETE arm type)" | `records[].ops[].compose`, which had the arm-type rule; it gains the three struct kinds, and its 'LeveledItemEntry' example is now tied to "a leveled-list entry" by name rather than sitting next to whichever noun preceded it |
| "composes is its batch sibling." | `records[].ops[].composes` (already said there) |
| "Copying a field from another record is `housecarl_apply`'s CopyFrom — there is no other version of a record that does not exist yet." | `records[].ops[].op`, where this surface's verb recital lives and where a caller looking for CopyFrom will not find it |
| "NESTING. parent= makes the record a CHILD (a dialogue line under a topic, a placed ref in a cell): the parent is an EXISTING record's FormID 'XXXXXX:Plugin.esp' OR the editorid of a record declared EARLIER in this same records= array (a same-call sibling)" | `records[].parent`, which had both parent forms; it gains the CHILD gloss and its two examples |
| "— which is how a topic AND its lines are authored in ONE call: records=[{record_type:'DialogTopic', …}, {record_type:'DialogResponses', …}] (declare the parent BEFORE its children)." | `records[].parent` (the worked example); the declare-the-parent-first rule was already on `records` |
| "A FormLink field VALUE can reference a same-call sibling too, written '@editorid' — so a line's order-chain and its topic back-link are authored in the same call (ops:[{field_path:'Topic', value:'@MyTopic'}, {field_path:'PreviousDialog', value:'@MyTopic_L1'}])." | `records[].ops[].value`, which already named the '@<editorid>' form; it gains the worked example |
| "'@editorid' must name a record declared EARLIER in the array or the record being created ITSELF (self-reference — e.g. a quest's VMAD alias fragment whose Property.Object is the quest). Only on FormLink fields, including inside a compose spec." | `records[].ops[].value` |
| "collection= names WHICH of the parent's child slots to add into — a child LIST (e.g. a cell's 'Persistent'/'Temporary') or a SINGLE-child slot (a cell's 'Landscape', a worldspace's 'TopCell', which hold exactly one and refuse rather than overwrite when occupied) — needed only when more than one fits." | `records[].collection` (its weaker one-example version replaced) |
| "grid= is the EXTERIOR cell's \"X,Y\" (record_type 'Cell' with parent= a Worldspace; houseCARL files it into the worldspace's block tree). A 'Cell' with NO parent and NO grid is an INTERIOR cell." | `records[].grid` (already said there; it gains the block-tree clause) |
| "LANE — where the write lands. Default: a NEW patch named patch= (auto-suffixed if taken, so a prior patch is never overwritten)." | `patch` (already said there in the same words); the LANE roster stays in the description as an axis line |
| "into='<an existing patch's filename>' ADDS these records to that patch instead — the way to accumulate across calls and sessions, and a parent created in a PRIOR into= call can be the parent here too (it resolves from the patch being extended, not only from the load order)." | `into` (its shorter parenthetical replaced by this fuller one) |
| "in_place='<plugin filename>' is the opt-in THIRD lane: the records are created straight INTO your ORIGINAL file (incl. a mod houseCARL didn't author) — no new patch, and NO houseCARL backup or undo (keep your own)." | `in_place` (already said there; it gains "(keep your own)") |
| "Full create parity in place, nesting included: a parent the target already owns hosts the child, a parent from another plugin is overridden in." | `in_place` |
| "Each new record gets a fresh FormID in THAT plugin's own range." | `in_place`, as the lane half of the allocation rule ("takes its fresh FormID from THAT plugin's own range, not a patch's") |
| "The FIRST in-place write to a given plugin returns a confirmation prompt (re-call with acknowledge=true); that consent covers touching your original ONLY — it NEVER skips the record verify." | `acknowledge`, which already carried the consent half; it gains the prompt sentence, in the wording `housecarl_apply`'s `acknowledge` uses |
| "ALL-OR-NOTHING (Q3): if ANY spec is malformed or fails pre-flight … the whole call is refused with per-record reasons and NOTHING is written. No partial patches, ever." | kept in description — it governs the whole call, not one member of one spec |
| the refusal causes: "an unknown or ambiguous record_type, a missing editorid, an illegal field op, a nested child with no resolvable parent, an ambiguous collection, a type that cannot be created (the bare abstract base 'Global'/'GameSetting' needs a concrete subtype; the refusal names them)" | `records`. Written as "the pre-flight refusals include …", which is what the old em-dash list was — instances, not the closed set; the engine refuses more than these six |
| "WHAT ELSE IS REPORTED (never silent, Q3): creating dialogue lines reports VOICE coverage (a response with no .fuz plays SILENT in game — the audio is yours to provide) and RESULT-SCRIPT binding (a bound script that is unwired or uncompiled runs nothing); creating cells reports the world content houseCARL does NOT author (lighting, terrain, water, navmesh — Creation Kit work)." | `records` — which reports come out is decided by what you put in records= |
| "TRANSPORT: readback=true expands the read-back to the FULL deep field-by-field dump of every created record (in place the verify ALWAYS runs and shows compactly by default; readback widens it) — confirm composed structures landed WITHOUT enabling the patch in MO2." | `readback`, which had the dump sentence; it gains the in-place clause and the reason |
| "The read-back is the WRITTEN FILE's content, NOT load-order truth: the patch wins nothing until enabled + sorted in MO2." | `readback` (its shorter "not load-order truth" replaced) |
| "max_chars caps the render with an explicit notice, never a silent cut." | `max_chars` (already said there, with the render's three blocks and the rendered-vs-total census spelled out) |
| "Every response carries epoch=<hex> — the identity of the index build parents and link values resolved from." | `format`, with both spellings named: `epoch=<hex>` on text, an `epoch` member on json |
| "records= (like every list-valued input) also accepts \"@<absolute path>\" in place of the inline array: the SAME array as a JSON manifest on disk." | `records` (already said there), and the description's WHAT axis line names the spelling. The "like every list-valued input" aside is now stated exactly rather than generally: `records=` is this tool's only list-valued input, and it binds as `JsonElement`, so the bare `"@<absolute path>"` is the correct spelling for it |
| "The path must be ABSOLUTE (the server resolves relative paths against its OWN working directory, not yours), and the file is read at CALL time." | `records` |
| "This tool authors NEW records. Editing an existing record's fields is `housecarl_apply`; dropping a whole record is `housecarl_remove`; an EMPTY header-only trigger plugin (no records at all) is `housecarl_create_plugin`. Read first with `housecarl_records`." | kept in description |

The description gained two things it did not carry before: the three axis lines (WHAT / LANE / TRANSPORT), naming every parameter — the roster is the description's only index of them now, and `patch=`, `acknowledge=`, `readback=`, `format=` and `max_chars=` were reachable from the old text only as prose mentions — and a COMPOSITION line for the nested one-shot, pointing at `parent=`, where the worked example now lives.

No stale pointers to repoint: nothing else in `src/` says the create tool description promises text this PR moved, and no generator probe reads it (`DialogueCkParityGuardProbe` derives its required op members from `housecarl_apply`'s `ops=`, not this tool's).

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`: the `ToolNames.Create` line comes out of `StillOversized`, and nothing else changes. Ran the class with the line gone and the description untouched — `EveryPublishedDescriptionFitsTheClientsCut(tool: "housecarl_create")` failed at 6,511, which is the first commit here. It passes at **1,225**.

## What was run

```
    1620 passed, 0 failed
dotnet .../housecarl-generator.dll description-vocab-guard    64 passed, 0 failed
dotnet .../housecarl-generator.dll wire-names-guard           28 passed, 0 failed
```

`wire-names-guard` because this PR moves brace-list examples into descriptions the guard parses: `records`'s and `records[].ops`'s new examples declare only real `CreateRecordSpec` / `CreateFieldOp` wire names, so INV4 and INV6 stay green. `ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. No checked-in capture snapshots the descriptions: `src/housecarl-mcp-tests/data/tools-list-2.0.json` holds tool names only. No `plugin/CHANGELOG.md` line — the records PR's entry covers this wave.

## Review fold

Both review findings were real and are folded.

- **`records=`'s FormID remedy was patch-only** (3c81b1e1). This PR widened the allocation half of the sentence to "the receiving plugin's own 0x800+ range" so it covers the in-place lane, but left the remedy half saying `into='<this patch>'`. `into=` is ownership-gated — `OwnedFoldersHolding` keeps only folders passing `IsHouseCarlOwned` — so an in-place caller who followed it got a `cannot extend:` refusal instead of a write. The remedy now names the lane the caller is on: `into='<this patch>'`, or `in_place='<that plugin>'` in place. With that sentence lane-agnostic, `in_place`'s "Each new record takes its fresh FormID from THAT plugin's own range, not a patch's" says nothing the widened version does not, so it comes out as a duplicate.
- **The epoch claim on `format=` was absolute** (27602b73). `Create`'s local `Refuse` returns `"error: " + message` with no epoch, so the text half of the promise is false for every tool-layer refusal — the lane-exclusivity ones, the missing-`records=` one, and every `ListParams.Read` one. Qualified the way `housecarl_remove` was in #592: every response answered from a build carries the stamp, a refusal that consulted no build carries none. Both spellings kept.

The parent-before-children rule stays on all three of its sites, because no two of them are the same sentence. `records=` says the child's `parent=` is what names the editorid, which is the routing from the top-level list down to the member; `records[].parent` carries it as the parenthetical that closes its worked example; and the COMPOSITION line is the description-level statement that a parent and its children fit in one call at all, plus the pointer to `parent=`. Dropping any one of the three loses something the other two do not say.

The published `housecarl_create` description is unchanged at **1,225** characters — both edits landed in parameter descriptions, which are delivered whole.

Re-ran from the worktree after the fold: `dotnet build` 0 errors, `tier!=bridge` 1,620 passed / 0 failed, `wire-names-guard` 28 passed / 0 failed.


src/housecarl-mcp/CreateTools.cs
```diff
@@ -19,85 +19,33 @@ public static class CreateTools
...
...One record is a set of one. For a nested one-shot, declare the parent BEFORE the children whose parent= names its editorid. A member the shape does not declare is refused BY NAME at its element, never silently dropped. Each new record's FormID is auto-allocated in the receiving plugin's own 0x800+ range and REPORTED BACK: that is how you reference the record afterwards — to make ANOTHER record point at it, call this tool or " + ToolNames.Apply + " again with into='<this patch>' and that FormID as the value. The pre-flight refusals include an unknown or ambiguous record_type, a missing editorid, an illegal field op, a nested child with no resolvable parent, an ambiguous collection, and a type that cannot be created (the bare abstract base 'Global'/'GameSetting' needs a concrete subtype; the refusal names them). What else is reported, never silently: creating dialogue lines reports VOICE coverage (a response with no .fuz plays SILENT in game — the audio is yours to provide) and RESULT-SCRIPT binding (a bound script that is unwired or uncompiled runs nothing); creating cells reports the world content houseCARL does NOT author (lighting, terrain, water, navmesh — Creation Kit work).")]
```
Avick3110: The harvest widened `the patch's own 0x800+ range` to `the receiving plugin's own 0x800+ range` so the allocation sentence covers the in-place lane too, but the remedy half of the same sentence stayed patch-only: `call this tool or housecarl_apply again with into='<this patch>' and that FormID as the value`.

`into=` is ownership-gated — `LoadOrderService.OwnedFoldersHolding` keeps only folders passing `IsHouseCarlOwned`, and the extend refusal offers only houseCARL-owned patches. So on the in-place lane, a caller who follows this sentence gets a `cannot extend:` refusal instead of a write. The tool's own completion hint already says the right thing for that lane (`WriteTools.InPlaceAgainHint`: "pass in_place=\"X.esp\" again").

Scenario: a model creates a Spell in place into `CoolWeapons.esp`, reads back its FormID, then wants a new Perk to point at it — it re-calls with `into='CoolWeapons.esp'` as instructed and is refused, because that plugin is not a houseCARL patch.

Either re-scope the sentence to the patch/into lanes, or name both continuations (`into='<this patch>'`, or `in_place='<that plugin>'` on the in-place lane).

src/housecarl-mcp/CreateTools.cs
```diff
@@ -19,85 +19,33 @@ public static class CreateTools
...
             bool acknowledge = false,
-        [Description("TRANSPORT: expand the read-back to the FULL deep field-by-field dump of every record this call created (not just the fields you set). The written file's content, not load-order truth.")]
+        [Description("TRANSPORT: expand the read-back to the FULL deep field-by-field dump of every record this call created (not just the fields you set) — confirm composed structures landed WITHOUT enabling the patch in MO2. In place, the verify ALWAYS runs and shows compactly by default; this widens it. The read-back is the WRITTEN FILE's content, NOT load-order truth: the patch wins nothing until enabled + sorted in MO2.")]
             bool readback = false,
-        [Description("TRANSPORT: 'text' (default) | 'json' (the same data, machine-readable, accounting in-band).")]
+        [Description("TRANSPORT: 'text' (default) | 'json' (the same data, machine-readable, accounting in-band). Every response carries the epoch stamp — the identity of the index build parents and link values resolved from — spelled epoch=<hex> on 'text', and as an 'epoch' member on 'json'.")]
```
Avick3110: The move sharpened the epoch claim into a per-lane promise (`spelled epoch=<hex> on 'text', and as an 'epoch' member on 'json'`), and the text half of that promise is false for every tool-layer refusal on this surface.

`Create`'s local `Refuse` returns `"error: " + message` on the text lane with no epoch appended — the comment right above it says so ("Epoch is null on all of them: none has consulted a build yet"). Every lane-exclusivity refusal, the missing-`records=` refusal, and every `ListParams.Read` refusal go through it. On the json lane `JsonWire.RenderError(message, null)` does emit the member, but as `"epoch": null` (`WriteNullable` writes a null rather than omitting), so "carries the epoch stamp" is only true in the shape sense there.

Scenario: a caller passes both `into=` and `in_place=`, gets `error: into= and in_place= are different lanes …`, and finds no `epoch=<hex>` anywhere — despite the parameter description saying every response carries one.

The old description made the same absolute claim, so this is inherited rather than introduced; but since this PR is a fidelity pass over exactly this sentence, "every response that reaches a build" (or "every non-refusal response") would make it true.

src/housecarl-mcp/CreateTools.cs
Avick3110: Right, and fixed in 3c81b1e1: the remedy now names the lane the caller is on (into='<this patch>', or in_place='<that plugin>' in place), and in_place='s restatement of the allocation rule went with it as a duplicate of the widened sentence.

src/housecarl-mcp/CreateTools.cs
Avick3110: Right, and fixed in 27602b73: the claim is now qualified to responses answered from a build, with a refusal that consulted no build carrying none, in the wording housecarl_remove took in #592; both spellings kept.

PR #596, opened 2026-09-06
Make a records scan render cheap, declared and cancellable

### LINKED ISSUE #582: [Bug] records over 66,856 NPC rows with resolve_names: silent for 30 minutes, and the work keeps running after the client aborts
Seen 2026-09-05 on the 2.0.0-beta.1 install against the ARR 2.0 order (3801 plugins, 66,856 NPC_ winners), during the skill-surface audit's control run for bulk-record-jobs.

**What happened.** `housecarl_records` with `types=["NPC_"]`, `format="dense"`, `to_file=...`, and `project={"form":"fields","resolve_names":true,"fields":[EditorID, Name, Race, Configuration.Level.Level, Configuration.Level.LevelMult, Configuration.CalcMinLevel, Configuration.CalcMaxLevel]}` sent no response and no progress for 1800 s, at which point the Claude Code client aborted the call (`sent no response or progress for 1800s; aborting`). Three earlier full-set NPC_ scans in the same session, without `resolve_names`, returned in seconds, so the scan is not the cost; the per-row Race winner lookup is. No partial file was written.

**After the abort** the server process kept working: ~5 cores and 1.3 GB for more than 30 minutes after the client had given up, then ~2.4 cores an hour later. The client's abort does not cancel the server-side call, and a retry competes with the orphaned work.

**Why it matters.** A tool never returns a silently degraded mode. Thirty minutes of silence on a call whose cost is linear in row count, with no progress and no cancellation, is that. Either the call should refuse up front (row count times a per-row resolve above a bound, with the shape that would work), report progress, or be cancellable on client abort. #422 (long-running sweep support) is the nearest open issue; this is the concrete case.

**Evidence.** The agent's call log with the exact arguments and the verbatim error is in the private audit corpus (`dev/skill-validation/audit-2026-09/bulk-record-jobs/control/log.md`), not in the repo.

Not milestoned; Aaron decides whether it promotes into W5 item 19.

Avick3110:
A `housecarl_records` scan's detail render read each row's body with a call that opened a fresh overlay session and then walked the whole winning plugin looking for that one record, so the per-row cost was linear in the winning plugin's record count and a whole-order NPC catalogue took about 45 minutes — past the 30-minute idle timeout the client gives a call, with no progress and no way to stop it. Every lane that reads a body per row — the four scan renders (text, json, dense, artifact) and the batch lane behind `form='everything'`, `form='rows'` and a scoped or folded `fields` projection — now gathers bodies through one chunked, per-plugin reader instead. On top of that the call declares what the render cost, refuses up front when the projected render is past a bound, and honours a client's cancellation inside one record.

Closes #582

## Profile

`resolve_names` is not the cost, and neither is the mmap: `LoadOrderResolver.SeekBody` enumerates **every major record in the plugin** to find one FormKey, and both reading lanes called it once per row. Measured on one synthetic world — a single master holding 30,000 records (2,000 WEAP, 28,000 AMMO), the scan selecting the 2,000 WEAP:

| the row cost | ms/row |
|---|---|
| per-row `GetRecord` — the flat whole-plugin seek every rendered row used to pay | 1.15 – 1.25 |
| the chunked per-plugin gather, untyped | 0.005 – 0.006 |
| the chunked per-plugin gather, with the scan's own types | 0.004 – 0.005 |

End to end, per row, over the same 2,000 rows:

| call | ms/row |
|---|---|
| `form='fields'`, dense, three field paths | 0.052 – 0.060 |
| the body lane, `ResolveBatch` with a three-field projection | 0.013 |
| the body lane, `ResolveBatch` reading the WHOLE record | 29.9 |
| `form='everything'`, json | 31.6 – 32.0 |

Two things follow, and they are the two substantive changes in this fold.

**The batch lane needed the same gather.** It was reading each row's body with the same per-row plugin walk the scan render had stopped doing, so `form='everything'` over the #582 selection would have reproduced the silent call with the bound's blessing. It now shares one `BodyPrefetch` with the scan render: a chunk of rows at a time, one enumeration per source plugin, and `LoadOrderResolver.BodySeeks` counts the per-record walks so a test can hold both lanes to it.

**One bound over both lanes was wrong.** A named-fields row costs 0.013 ms; a whole-record row costs 30 ms, and the gather does not move that number — the cost is materialising every field, not finding the record. At 40 ms declared, `form='everything'` gets its own bound (below), because 300,000 whole-record rows is over three hours, not ten minutes.

**The gather waits for the row.** A chunk is 2,000 rows and the default render window is 500, so filling the chunk on the first row meant walking every plugin the selection touched — including the plugins whose rows a `max_chars` cut never reached, at a full enumeration each, three times over on the auto-spill path. Each plugin is now walked on the first row that actually asks for it, covering that plugin's whole share of the chunk, so the cost is the plugins the RENDERED rows came from and never more than the per-row seek this replaced. `LoadOrderResolver.CollectPasses` counts those walks, and `AChunkWalksOnlyThePluginsTheRenderedRowsCameFrom` holds five rows from one plugin to one walk (twelve before).

Passing the scan's types down to the gather (a typed GRUP seek rather than a flat walk) is folded and measures 11 ms → 9 ms for 2,000 bodies on this world — small here because the flat walk stops as soon as it has every wanted key and the WEAP GRUP is enumerated first. The win is where the wanted records are not at the front of the file, which is the #354 shape the argument comes from; it can never cost more than the flat walk, since that is its own fallback.

I did not make `offset=` seek. Post-fix the scan itself is the cheap half — a summary scan of a whole type at `offset=1500` measured the same 5–6 ms as at `offset=0` — and remembering a scan position across calls means holding per-query state at rest between calls, which the fourth design sentence rules out. So it is declared instead, on the `offset=` parameter: `offset=` re-scans from the start, every window pays the whole scan again, and narrowing the scan terms beats paging far into one.

## The bounds and how they were chosen

Same rule for both, from two stated numbers in `RenderBudget`: a declared per-row cost and a ten-minute render ceiling. Ten minutes is a third of the client's 30-minute idle timeout, leaving the scan itself, the artifact write and one retry inside the same window.

| lane | declared per-row cost | measured | bound |
|---|---|---|---|
| a `fields`/`rows` row — the named field paths off a body | 2 ms | 0.013 – 0.06 ms | **300,000 rows** |
| a `form='everything'` row — the whole record materialised | 40 ms | 29.9 – 32 ms | **15,000 rows** |

The 2 ms is deliberately pessimistic: it is the reported real-order 40 ms divided by the *smallest* speedup measured, so it carries better than an order of magnitude over the extrapolated real figure. The 40 ms is the measurement plus a margin for a record type fatter than a weapon; it is the one number here with no real-order check behind it, and `render_ms` is how it gets one. The 66,856-row catalogue that started this issue runs rather than refuses on the fields lane, which is the intended outcome; the same selection as `form='everything'` refuses, and says to name the fields.

The refusals name the shapes that fit and the caveats that decide between them. The whole-record one leads with the move between the lanes:

> error: this call renders 400,000 rows and each one reads a WHOLE record body — about 267 minutes of render, past the 15,000-row bound form='everything' is given (a client stops waiting at 30 minutes). Name the fields you need instead — project.form='fields' with fields=[…] reads a body per row too but costs a fraction of a whole-record read, and is bounded at 300,000 rows. Or narrow the scan terms (types=, plugins=, where=), or take the selection in windows with limit= and offset= — offset= re-scans the selection from the start, so a window costs more the further in it is. to_file= captures the COMPLETE selection rather than a window, so it renders every row and does not combine with offset=: narrow the scan until the whole set fits, then write it in one call.

`to_file=` is named as the whole-set shape, not as a way under the bound, because it is not one: it renders every selected row and is held to the same bound. The bound is checked **below** the walk lane, which returns before it — a walk renders the set it reaches under its own node budget, not the seed scan the count is of, so measuring it there refused calls for a cost they were never going to pay.

**Where the bound applies** (the fold of Aaron's review — the numbers are unchanged, what moved is the set of lanes):

| lane | held to the bound | why |
|---|---|---|
| the in-order scan | yes | it reads a body per rendered row |
| an off-order `source=` scan | yes | the same read, off a file outside the order — it also reports `render_ms` now |
| a `formids=` list | yes, over the LIST's length | it reads a body per id BEFORE `limit=`/`offset=` window the render, so paging does not lower the cost and its refusal says so instead of naming scan terms it does not have. `formids=["@<file>"]` is a list of any length, and it is the path the whole-record refusal and the `to_file=` description send the caller to |
| a `formids=` list a WALK derived | no | `walk.max_nodes` bounds that set already, the same reason the seed scan is not measured |
| `counts_only` on a scan | no | the census reads no bodies |
| `counts_only` on a `formids=` list | yes | that lane reads every id's body whatever it renders |

## What each call actually spent

Every `fields`/`rows`/`everything` scan reports it, on every transport, because a bound calibrated on one machine is only checkable if the truth comes back:

| transport | what it carries |
|---|---|
| text scan render | `rendered N rows in X ms`, held back from `max_chars` rather than appended past it |
| json / dense scan render | `rendered`, `render_ms` |
| `to_file=` manifest | `rendered_to_file`, `render_ms` — timed on the artifact write, which is where a to_file= call actually renders its rows |
| the body lane (`rows`/`everything`), all transports | `rows_read`, `render_ms` / `read N record bodies in X ms` |
| the off-order `source=` body lane, all transports | the same pair — it was reporting nothing |

The `formids=` list lane does not report `render_ms`. It reads the whole list and renders a `limit=`/`offset=` window of it, so the count beside the milliseconds would be the window's rather than what was read; passing the read count through the shared batch render is a second parameter on that surface, and I left it for a ruling rather than shipping a number that means something different on one lane. Its cost is declared on `limit=` instead.

## Cancellation, and what the server can and cannot detect

The MCP SDK (ModelContextProtocol 2.2.0) binds a `CancellationToken` parameter on an `[McpServerTool]` method to the token it cancels on `notifications/cancelled` for that request id, and leaves it out of the published JSON schema. `Records` takes one and it now reaches every loop that reads a body:

| where | polled | a cancel there finishes as | test |
|---|---|---|---|
| the in-order scan loop | yes | a cancellation (the catch-all rethrows it) | `ACancelIsNotReportedAsAnInternalFailure` |
| the off-order scan loop | yes | a cancellation — was a `file '…' could not be fully read` refusal | `ACancelledOffOrderScanIsNotReportedAsAFileFault` |
| the scan render, per row | yes | a cancellation | `ACancelStopsTheRenderWithinOneRow` |
| the batch read, per row | yes — the token now reaches it from all five call sites | a cancellation | `ABatchBodyReadStopsWhenTheClientCancels` |
| the off-order body read (file walk and per-record read) | yes | a cancellation — and not a file fault, the same rule the scan loop follows | `ACancelStopsTheOffOrderBodyRead` |
| the SkyPatcher overlay read, per record | yes | a cancellation | `ACancelStopsTheOverlayPostRead` |
| the artifact write | yes | no artifact, and no stray `.tmp-*` | `ACancelledToFileCallLeavesNoArtifact` |
| the auto-spill's reserved name | released on the way out | nothing left in the results directory | `ACancelledCallLeavesNothingInTheResultsDirectory` |

The last two rows are the fold of Aaron's review: the off-order body read took no token and polled none of its own, and `OverlayPostBatch` was the one body lane in the tool still without one — so whether a call stopped depended on which `source=` arm it took.

A cancelled `to_file=` call leaves nothing behind, and not by cleanup: `ResultArtifact.Writer` buffers rows in memory and `Save` writes manifest plus rows to a temp file and moves it into place in one step, so a cancel during the row loop throws before any bytes reach the destination. The auto-spill path additionally reserves its name on disk (`ResultsStore.NextPath` creates the file), and now gives that name back if a cancel lands inside the write.

**Transport close.** `WithStdioServerTransport` runs the server as a single-session service that exits when stdin closes, so a client that closes the transport takes the process down and the work with it. What the server cannot detect is the case this issue actually observed: the client gave up waiting, sent no `notifications/cancelled`, and kept the process alive. Nothing in the protocol or the SDK reports a client that has stopped reading while the pipe is still open, so that case is fixed only from the client side, or by the client sending the cancel notification this PR honours — which is why the bound exists. I could not reproduce a real `notifications/cancelled` end to end without driving the live client, so cancellation is verified at the tool body and below.

## Tests

`src/housecarl-mcp-tests/RecordsRenderCostTests.cs`, on a `RenderCostWorld` of sixty weapons in one master, one plugin in a switched-off mod folder (the off-order lane needed a world), and twelve small plugins of five AMMO each — a claim about which PLUGINS a render walks needs rows whose winners live in more than one, and their own record type keeps every weapon count the other tests pin. Tagged `tier=integration`.

| test | before | after |
|---|---|---|
| `ADetailRenderOpensOneOverlayPerPluginNotPerRow` | fails (60 opens for 60 rows) | passes (1) |
| `AnEverythingRenderReadsEachPluginOnceNotOncePerRow` | fails (60 per-record plugin walks) | passes (0) |
| `TheDenseAccountingReportsTheRowsRenderedAndWhatTheRenderCost` | fails (no `render_ms`) | passes |
| `TheJsonAccountingReportsTheRowsRenderedAndWhatTheRenderCost` | fails (no `render_ms`) | passes |
| `TheTextAccountingReportsTheRowsRenderedAndWhatTheRenderCost` | fails (no line) | passes |
| `TheEverythingAccountingReportsWhatReadingItsBodiesCost` | fails (the body lane reported no cost) | passes |
| `AToFileCallReportsWhatWritingItsRowsCost` | fails (`render_ms` suppressed on `to_file=`) | passes |
| `TheAccountingLineIsReservedFromTheRowBudget` | fails (no reserve for the line) | passes |
| `ARenderOverTheBoundRefusesNamingTheWindowAndTheWholeSetShapes` | fails (serves the rows) | passes |
| `AnEverythingRenderIsBoundedByItsOwnWholeRecordCost` | fails (no whole-record bound) | passes |
| `TheFieldsBoundDoesNotRefuseAnEverythingRenderThatFitsItsOwn` | fails (one bound refused it) | passes |
| `AWalkIsNotRefusedForTheSizeOfItsSeedScan` | fails (refused for its seed count) | passes |
| `AToFileCallOverTheBoundRefusesAndWritesNoArtifact` | fails (writes the artifact) | passes |
| `ACancelStopsTheRenderWithinOneRow` | fails (reads every row) | passes |
| `ABatchBodyReadStopsWhenTheClientCancels` | did not compile — the overload took no token | passes |
| `ACancelledOffOrderScanIsNotReportedAsAFileFault` | fails (returns the file-fault refusal) | passes |
| `ACancelledToFileCallLeavesNoArtifact` | fails (writes the artifact) | passes |
| `ACancelIsNotReportedAsAnInternalFailure` | fails (returns normally) | passes |
| `ACancelStopsAnEverythingRenderToo` | passes both (the scan raises first) | passes |
| `ACancelledCallLeavesNothingInTheResultsDirectory` | passes both | passes |
| `ARenderUnderTheBoundIsUntouched` | passes both | passes |
| `TheCensusIsNotHeldToTheRenderBound` | passes both | passes |

Seven more came out of Aaron's review:

| test | before the fold | after |
|---|---|---|
| `ACancelStopsTheOffOrderBodyRead` | fails (the off-order pole read polled no token at all) | passes |
| `AnOffOrderRenderOverTheBoundRefusesToo` | fails (serves the rows) | passes |
| `AnOffOrderBodyRenderReportsWhatItsBodiesCost` | fails (no `render_ms` on this lane) | passes |
| `AFormidsReadOverTheBoundRefuses` | fails (serves the rows) | passes |
| `AFormidsCensusIsHeldToTheBoundBecauseItStillReadsEveryBody` | fails (serves the census) | passes |
| `ACancelStopsTheOverlayPostRead` | fails (returns normally) | passes |
| `AChunkWalksOnlyThePluginsTheRenderedRowsCameFrom` | fails (5 rows from one plugin walked 12 plugins) | passes |

Fail-before was verified by disabling the mechanism in place and re-running for the gather, the off-order guard, the walk placement and the deferred plugin walk; the rest assert a value or a refusal that did not exist before the change, and the last four are no-regression guards that pass by design. The profiling harness was a temporary test class; it is not in the branch.

## Descriptions

Three sentences, each on the parameter it describes — `limit=` (the per-row render cost, `render_ms`, and the two bounds), `offset=` (the re-scan), `to_file=` (renders every row, same bound). `limit=` now also names the lanes the bound holds on and says that a `formids=` read costs the whole list rather than the window, since that is the one lane where paging does not lower the cost. The tool description is untouched and `PublishedDescriptionBoundTests` passes, so `housecarl_records` stays under 2,048.

## What was run

From inside `.claude/worktrees/records-cost`:

```
```

1,649 passed, 0 failed. `ci-all` was not run locally (#570); CI runs it. Nothing was run against the live MO2 instance.


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4749,6 +4769,7 @@ public CrossQueryOutcome OffOrderQuery(PoleInfo pole, IReadOnlyList<string>? typ
             var seen = new HashSet<FormKey>();
             foreach (var rec in ov.EnumerateMajorRecords())
             {
+                ct.ThrowIfCancellationRequested();   // a client that aborted stops the scan inside one record
```
Avick3110, line 4824: `CrossQuery` got the guard that keeps a cancel from being renamed a scan fault (line 4535: `catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }`), but `OffOrderQuery` did not. This new `ThrowIfCancellationRequested` sits inside the try whose outer catch at line 4832 is `catch (Exception ex) { return CrossQueryOutcome.Fail($"file '{pole.Plugin}' could not be fully read — {ex.GetType().Name}: {ex.Message}") ... }`.

So cancelling a `source="SomeOffOrder.esp"` scan (RecordsTools.cs:1539 passes `ct` into this method) does not finish the request as cancelled — it returns the refusal `file 'SomeOffOrder.esp' could not be fully read — OperationCanceledException: The operation was canceled.` The client sees a data failure naming a file that is perfectly readable, which is exactly the mislabelling the PR body says it set out to avoid.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2907,6 +2917,7 @@ public IReadOnlyList<ReadOutcome> ResolveBatch(IReadOnlyList<string> formids, IR
         var outcomes = new List<ReadOutcome>(formids.Count);
         foreach (var raw in formids)
         {
+            ct.ThrowIfCancellationRequested();   // a client that aborted stops the batch inside one record
```
Avick3110, line 2935: This poll can never fire: no caller supplies the token. The four production call sites are `RecordsTools.cs:598`, `1358`, `1373` and `1378`, and all four go through the *simple* `ResolveBatch` overload at line 2884, which forwards `null, out _, out _, containerHint, depths` and stops there — it neither takes nor passes a `CancellationToken`. `ResolveBatchFromPole` (line 3022, used at RecordsTools.cs:1342) has no token parameter at all.

That matters because `bodyLaneForm` (RecordsTools.cs:1247) routes `project.form='everything'` and `'rows'` — and `'fields'` under a scoped pole or a fold — through this batch lane and returns at line 1430, before any of the cancellable render paths. So `project={form:"everything"}` on a large scan keeps running after the client cancels, which is the case `ACancelIsNotReportedAsAnInternalFailure` does not cover (it uses `form="fields"`, which takes the `ScanDetailReader` lane instead).

Either thread `ct` through the simple overload and `ResolveBatchFromPole` to those four call sites, or drop the parameter so it does not read as coverage that exists.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1252,6 +1253,14 @@ List<KeyValuePair<string, string>> Echo()
                 headerLine += "\n" + outcome.ReverseIndexNote;
             }
 
+            // ---- the render's own bound. The scan terms bound the SCAN; every rendered row of a body form then
+            // reads a record body, and that cost was undeclared until a call went silent past the client's idle
+            // timeout (#582). Checked before any body is read, so the caller learns the shape instead of waiting.
+            if ((bodyFields || form == "everything") && !counts_only
```
Avick3110: The gate is `bodyFields || form == "everything"`, i.e. `fields`, `rows` and `everything`. But only *some* of those reach the `ScanDetailReader` the 2 ms/row figure was measured on. `bodyLaneForm` at line 1247 sends `rows`, `everything`, and scoped/folded `fields` to the batch lane at 1336, which reads through `svc.ResolveBatch` → `ResolveRead` → `view.GetRecord` → `SeekBody` — the same per-row whole-plugin walk this PR removed from the scan render, still there and untouched.

Concretely: `types=["NPC_"] project={form:"everything"} to_file="..."` over the 66,856-record selection from #582 sets `effLimit = int.MaxValue` (line 1201), passes this check (66,856 < 300,000), and then does 66,856 whole-overlay seeks against masters holding ~250,000 records — the same ~45-minute silent call the issue reports, now with the bound's blessing. At the measured pre-fix ~40 ms a row, the 300,000-row bound is over three hours for these two forms, not the ten minutes `RenderBudget.MaxRenderRows` is derived from.

Either bound the two lanes separately (a per-row cost per lane), or route the batch lane through a chunked gather too so the one number is true of both.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1252,6 +1253,14 @@ List<KeyValuePair<string, string>> Echo()
...
+                && outcome.Error is null && outcome.Groups is null
+                && RenderBudget.Refuse(outcome.Keys.Count) is { } tooBig)
```
Avick3110: This runs *before* the walk branch at line 1266, so on a `walk=` call it measures the wrong set. `derivedSelection` includes `walk is not null` (line 1080), so `effLimit` is `int.MaxValue` (line 1201) and `outcome.Keys` is the full **seed** scan, not what the walk renders — the walk lane windows its own rows by `limit=`.

So `types=["REFR"] walk={...} project={form:"fields", ...}` on a real order (REFR alone is well past 300,000 records) is now refused with `this call renders 1,2XX,XXX rows and each one reads a record body`, which is not what the call would have done: the walk would have rendered at most `limit` rows. The refusal also sends the caller to `limit=`/`offset=`, neither of which changes the seed count that actually tripped it.

Moving the check below the walk/comparison branches — next to the render it is about — would keep it honest.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4061,7 +4072,8 @@ public CrossQueryOutcome CrossQuery(string? type, IReadOnlyList<FormKey>? refere
                                         bool conflictsOnly, IReadOnlyList<string>? plugins, IReadOnlyList<string>? where, int limit,
                                         bool definedIn = false, string? groupBy = null, int offset = 0, string? whereSource = null,
                                         IReadOnlyList<ArtifactDemand>? artifactDemands = null,
-                                        IReadOnlyList<FormKey>? referencesNone = null)
+                                        IReadOnlyList<FormKey>? referencesNone = null,
+                                        CancellationToken ct = default)
         => CrossQuery(type is null ? null : new[] { type }, references, editoridContains, conflictsOnly, plugins, where,
                       limit, definedIn, groupBy, offset, whereSource, artifactDemands, referencesNone: referencesNone);
```
Avick3110: The parameter is added but the forwarding call does not pass it, so the set-valued overload silently binds `CancellationToken.None`. Any caller of the single-`type` overload that hands in a live token gets no cancellation and nothing says so — the signature claims otherwise.

```suggestion
                                        CancellationToken ct = default)
        => CrossQuery(type is null ? null : new[] { type }, references, editoridContains, conflictsOnly, plugins, where,
                      limit, definedIn, groupBy, offset, whereSource, artifactDemands, referencesNone: referencesNone, ct: ct);
```

src/housecarl-mcp/ScanDetailReader.cs
```diff
@@ -0,0 +1,111 @@
...
+            set.Add(fk);
+        }
+
+        var sink = new Dictionary<FormKey, IMajorRecordGetter>(end - start);
+        foreach (var (plugin, wanted) in byPlugin)
+        {
+            _ct.ThrowIfCancellationRequested();
+            // A plugin that cannot be read leaves its rows unfetched here and the row's own read raises the same
+            // fault it always did — the prefetch is an optimisation and must never become a second error path.
+            try { view.CollectRecords(_session, plugin, wanted, null, sink); }
```
Avick3110: `getterTypes` is passed as `null`, so every prefetch is `CollectRecords`' *flat* whole-plugin enumeration. The shape this class says it copies does not do that: `WinnerBodies.cs:51` passes the scan's `types` through, and `CollectRecords`' own doc comment says the argument "narrows the walk to the GRUPs those types live in when the caller knows them" — the #354 cost, where a master's placed references outnumber the records wanted fifty to one.

So a `types=["NPC_"]` catalogue pays `ceil(rows / 2000)` flat walks *per contributing plugin*: on a 66,856-row selection whose winners spread across, say, 40 plugins, that is up to 34 x 40 = 1,360 full enumerations of large masters rather than 1,360 typed GRUP seeks. Much better than one walk per row, but several times more than it needs to be — and the class exists to cut exactly this cost. The scan already resolved the type filter; threading it down to the reader would close the gap.

src/housecarl-mcp/ReadTools.cs
```diff
@@ -326,7 +330,13 @@ public static string RenderCrossQuery(LoadOrderService svc, CrossQueryOutcome q,
             }
             rendered++;
         }
+        renderClock.Stop();
         AppendOwnedChildNotes(sb, notes);
+        // What the RENDER cost, stated in-band: the scan terms bound the scan, and the per-row body read is the
+        // other half of a call's cost, which nothing said before (#582).
+        if (detail && !(spill?.ManifestOnly ?? false))
```
Avick3110: Two things on this line, both about the shape the bound is aimed at.

**`to_file=` reports no render cost at all.** `spill.ManifestOnly` is true exactly on the `to_file=` disposition (RecordsTools.cs:1440), and that call *did* render every selected row — through `Artifacts.WriteCrossQuery`, which now runs the same `ScanDetailReader`. The row loop above is skipped, so the clock reads ~0 and the line is suppressed. Same condition in `JsonWire.cs:1163` and `1387`. The PR argues the constant is "a guess calibrated on one machine" that `render_ms` exists to check against a real order — but the whole-set shape, the one that renders the most rows and is the reason for the bound, is the one shape that reports nothing. Timing the artifact write and carrying that number into the manifest-only response would fix it.

**The line has no `cap` reserve.** It is appended after the loop, outside the `sb.Length >= cap - notes.Reserve` test that `ChildNotes.Reserve` exists to make honest. A render that filled to `max_chars` now returns roughly 25-30 characters past the declared ceiling.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
 ## Unreleased
 
+- **A `housecarl_records` scan now says what its render cost, refuses a render too big to finish, and stops
+  when you abort the call.** Reading a record body per rendered row used to walk the whole winning plugin for
+  each one, so a projection of a few fields over a big selection could run for tens of minutes with nothing
+  coming back; the bodies are now gathered a plugin at a time and the same catalogue renders in a fraction of
+  the time. Every `fields`/`rows`/`everything` scan reports the rows it rendered and the milliseconds the
+  render took, and a call whose render would exceed the bound `limit=` names refuses up front, saying whether
+  to narrow the scan terms, take windows with `limit=`/`offset=`, or write the whole set with `to_file=` —
+  with the caveats that decide between them on the `offset=` and `to_file=` parameters. Aborting the call in
```
Avick3110: medium: "Aborting the call in the client now stops the scan and the render" promises the reported symptom is fixed, but the PR body says the opposite for the case #582 actually observed - the client gave up waiting, sent no `notifications/cancelled`, and the work carried on - and that a real cancel notification could not be reproduced end to end. A user who reads this entry and then hits the same 30-minute silence will believe the fix regressed. Say what is known: the call stops when the client sends a cancel, and name what the reader can check (the `render_ms` line), per this file's own head guidance.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1435,15 +1444,15 @@ string RenderEv(SpillState? sp, out bool trunc) => json
...
+                Wire.QueryFormat.Dense or Wire.QueryFormat.Json => JsonWire.RenderCrossQuery(svc, outcome, projFields, max_chars, resolveNames, winnerFields, depth, sp, out trunc, envelope, qLevers, ct),
+                _ => headerLine + "\n" + Wire.RenderCrossQuery(svc, outcome, projFields, max_chars, resolveNames, winnerFields, depth, sp, out trunc, qLevers, ct),
             };
             var rendered = Render(spill, out var truncated);
             if (spill is null && truncated && outcome.Error is null)
             {
                 var path = ResultsStore.NextPath(ToolNames.Records, outcome.Epoch ?? "none");
-                var (s, aerr) = Artifacts.WriteCrossQuery(svc, outcome, readPaths, resolveNames, winnerFields, depth, path, "ceiling", Echo(), LeverNames.Records, fold: foldPlan);
+                var (s, aerr) = Artifacts.WriteCrossQuery(svc, outcome, readPaths, resolveNames, winnerFields, depth, path, "ceiling", Echo(), LeverNames.Records, fold: foldPlan, ct: ct);
```
Avick3110: low: a cancel inside this ceiling spill leaves a stray artifact after all. `ResultsStore.NextPath` on the line above reserves the name by creating an empty `.jsonl` on disk (`FileMode.CreateNew`), and it is removed only by `ResultsStore.Release(path)` on the line below, which the `OperationCanceledException` skips - so a cancel during this re-render strands a zero-byte spill in the results directory until the seven-day prune. `ACancelledToFileCallLeavesNoArtifact` covers the caller-named `to_file=` path, which needs no reservation, so this one is untested. Wrapping the write in try/catch(OperationCanceledException) { Release(path); throw; } would make the "a cancelled call writes nothing" claim hold on both paths.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 462b3ee0 and a7ae33b9. The batch lane now shares one chunked per-plugin gather with the scan render, and measuring what was left showed the bound itself was wrong: a named-fields row costs 0.013 ms while a form='everything' row costs 30 ms — the whole record materialised, which the gather does not move — so 'everything' gets its own 15,000-row bound and a refusal that names project.form='fields' as the move between them. Both numbers and the measurements are in the PR body; `AnEverythingRenderReadsEachPluginOnceNotOncePerRow` and `AnEverythingRenderIsBoundedByItsOwnWholeRecordCost` cover it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 5ec89f10. The token is threaded through the simple ResolveBatch overload and ResolveBatchFromPole to all five call sites (RecordsTools.cs:598, 608, 1358, 1373, 1378). The tokenless overload stays — it is the one EpochGuardProbe and the remedy tests use — but it now takes a defaulted token rather than swallowing the question. `ABatchBodyReadStopsWhenTheClientCancels` is the test; it would not have compiled before.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 5ec89f10 — the same rethrow guard CrossQuery has, and `ACancelledOffOrderScanIsNotReportedAsAFileFault` on the off-order world the fixture now carries (a plugin in a switched-off mod folder). Verified fail-before by removing the guard: it returned the file-fault refusal.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in a7ae33b9 — exempted, and the comment says which. The check now sits below the walk lane, which returns before it: a walk renders the set it reaches under its own node budget, not the seed scan the count is of, so measuring it there refused calls for a cost they were never going to pay. `AWalkIsNotRefusedForTheSizeOfItsSeedScan`, verified fail-before by moving the check back above the walk.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 5ec89f10, exactly as suggested.

src/housecarl-mcp/ScanDetailReader.cs
Avick3110: Folded in 462b3ee0: the scan's resolved getter types ride the outcome and both lanes pass them down. Re-measured, and the gain on my synthetic world is small — 11 ms to 9 ms for 2,000 bodies — because CollectRecords stops as soon as it has every wanted key and the WEAP GRUP is enumerated first, so the flat walk short-circuits. The win is where the wanted records are not at the front of the file, which is the #354 shape your argument comes from, and the typed walk can never cost more since the flat one is its own fallback. Profile table updated in the PR body.

src/housecarl-mcp/ReadTools.cs
Avick3110: Folded in 3db794c3, the first part exactly: the to_file= manifest now carries `rendered_to_file` and `render_ms` timed on the artifact write, and the body lane (rows/everything, which reported nothing at all) carries `rows_read` and `render_ms` on every transport. `AToFileCallReportsWhatWritingItsRowsCost` and `TheEverythingAccountingReportsWhatReadingItsBodiesCost`.

The second part I have folded but must scope down the claim: the accounting line now comes out of the row budget (the loop's cap test holds back `RenderBudget.AccountingReserve`, and a test holds that constant to the line at its widest). It does not make a filled render fit inside max_chars, and could not — the render already answers past the ceiling by the last record block, which is appended whole after the test, plus the truncation notice and the `spilled:` block, all of which predate this PR. So: the line is no longer the thing pushing it over, but 'never exceeds the ceiling' is not a claim this render can make, and I did not want to write a test asserting it.

plugin/CHANGELOG.md
Avick3110: Folded in f22d5c56. The entry now says a cancel sent by the client stops the work inside one record, and states plainly that a client which stops waiting WITHOUT sending a cancel is not detectable from the server — nothing in the protocol reports it — which is what the render bound is there for. It points at limit=/offset=/to_file= for the bounds rather than restating them.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 5ec89f10 — try/catch(OperationCanceledException) { Release(path); throw; } around the ceiling write. The test I added (`ACancelledCallLeavesNothingInTheResultsDirectory`) asserts the invariant at the call level and passes before as well as after: the cancel it sends lands in the render, and a cancel landing between the render and the spill write is a window I cannot time without adding a seam for it. Flagging that rather than dressing it up as a fail-before test.

Avick3110:
Rebased onto the merged description wave (default tip `cd6b3139`). Old tip `f22d5c56` → new tip `64f15ecb`; ten commits, none squashed.

One file conflicted, twice — once in each of the two commits that touch it:

- `plugin/CHANGELOG.md` — kept main's single wave entry ("Every tool description now fits inside the 2,048 characters…") and this branch's render-cost entry, in that file order, and dropped the branch's stale per-tool copy of the entry (the `housecarl_records`-only description entry that the wave entry replaced). The second conflict was the same file again, on the commit that rewrites the render-cost entry to say what a cancel actually stops; resolved the same way, keeping the rewritten wording.

Nothing else conflicted. `RecordsTools.cs` merged clean — the wave did not touch that file after this branch's merge base, so the three sentences on `limit=`, `offset=` and `to_file=` sit where they were, once each. `LoadOrderService.cs` merged clean too; the wave's edit there is in the forward path, not the read path this branch changes.

From the rebased worktree:

```
```

1,642 passed, 0 failed. `PublishedDescriptionBoundTests` passes (31 tests), so `housecarl_records` is still under 2,048. `ci-all` not run locally; CI runs it.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1527,7 +1550,7 @@ string OffOrderScan(LoadOrderService.PoleInfo pole)
             var outcome = svc.OffOrderQuery(pole, types, refFks, null, plugins?.names,
                                             plugins?.defined_in ?? false, where, offLimit, offGroupBy, offset,
                                             formidSet, offDemands.Count > 0 ? offDemands : null, door.CapturedView,
-                                            refNoneFks);
+                                            refNoneFks, ct);
```
Avick3110, line 1568: **The off-order body lane never sees the token.** `OffOrderQuery` gets `ct` here, but the branch that actually reads the bodies — `ResolveBatchFromPole` at line 1615 — is still called without it, so `ct` defaults to `default` and the read runs to completion.

Concretely: `source="SomeOffOrder.esp" types=["NPC_"] project.form='everything' to_file=…`. `offLimit` is `int.MaxValue` for a `to_file` call (line 1545), so every match's whole body is materialised. A `notifications/cancelled` stops the scan loop above and then the call goes silent anyway in the half that costs ~30 ms a row. That is the #582 shape, on the sibling lane. Every other body read in this tool now takes the token, and the PR's cancellation table reads as though this one does too.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1527,7 +1550,7 @@ string OffOrderScan(LoadOrderService.PoleInfo pole)
             var outcome = svc.OffOrderQuery(pole, types, refFks, null, plugins?.names,
                                             plugins?.defined_in ?? false, where, offLimit, offGroupBy, offset,
                                             formidSet, offDemands.Count > 0 ? offDemands : null, door.CapturedView,
```
Avick3110, line 1567: **The off-order lane is not held to the render bound, and reports no `render_ms`.** `RenderBudget.Refuse` is placed in `ScanLane` only (line 1275). The off-order lane's `form is ("fields" or "rows" or "everything")` branch at line 1612 reads a body per row exactly the same way, and nothing measures it.

An off-order plugin with 40,000 records as `project.form='everything' to_file=…`: `offLimit` is `int.MaxValue` (line 1545), so all 40,000 whole bodies are materialised — ~20 minutes at the PR's own 30 ms/row — with no up-front refusal. The `limit=` description now states flatly that a render too big to finish REFUSES up front and that `form='everything'` is bounded at 15,000 rows; a caller who reads that and points the same call at an off-order `source=` gets the old silent call. Same for the accounting: the changelog says every `fields`/`rows`/`everything` scan reports its rows and milliseconds *on every transport*, and `RenderOff` (line 1636) passes no `renderMs`.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -594,7 +595,7 @@ string Render(SpillState? sp, out bool trunc) => json
             else if (srcName is null)
             {
                 if (srcOverlay) Arm("skypatcher overlay (pre) = winner — the body the INI layer starts from");
-                outcomes = svc.ResolveBatch(ids, readFields, false, depth, resolveNames, null, demand, out var refusal, out var refusalEpoch, LeverNames.Records.ContainerHint, readFieldDepths);
+                outcomes = svc.ResolveBatch(ids, readFields, false, depth, resolveNames, null, demand, out var refusal, out var refusalEpoch, LeverNames.Records.ContainerHint, readFieldDepths, ct);
```
Avick3110, line 610: **The `formids=` list lane takes the token but not the bound.** `RenderBudget.Refuse` only runs in `ScanLane`; this lane reads a whole body per id with nothing checking the count first, and `ids` can come from `formids=["@<artifact>"]`, which is unbounded by construction.

That matters because it is the path the surface itself recommends. The `to_file=` description says "Re-enter it later via `formids=["@<path>"]`", and the new whole-record refusal ends by telling the caller to write the complete selection with `to_file=`. So: write a 400,000-row artifact, re-enter it as `formids=["@out.jsonl"] project.form='everything'`, and 400,000 whole records are materialised — ~4.4 hours at the declared 40 ms — while the scan lane refuses the same selection at 15,000. The `limit=` description states the bound unconditionally, so the one door left open is the one the refusal points at.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -594,7 +595,7 @@ string Render(SpillState? sp, out bool trunc) => json
             else if (srcName is null)
             {
                 if (srcOverlay) Arm("skypatcher overlay (pre) = winner — the body the INI layer starts from");
```
Avick3110, line 609: **`OverlayPostBatch` (line 585) is the one body lane in this tool left without `ct`.** Its two siblings ten lines below both take it now.

`formids=["@big.jsonl"] source='skypatcher_overlay' project.form='everything'` replays each winner through the SkyPatcher layer and reads its whole body; a client cancel is honoured on the `ResolveBatch` and `ResolveBatchFromPole` arms and dropped on this one, so whether the call stops depends only on which `source=` arm it took. The PR's cancellation table says the token now reaches the batch read from all five call sites.

src/housecarl-mcp/ScanDetailReader.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    void FillChunk(int i)
+    {
+        if (_view is not { } view || _session is null) return;   // unpinned: the per-row path answers
+        int start = BodyPrefetch.ChunkStart(i);
+        if (start == _chunkStart) return;
+        _chunkStart = start;
+        // The scan's own resolved types narrow each plugin's walk to the GRUPs they live in, which is the whole
+        // point of the gather on a master whose placed references outnumber the records wanted.
+        _chunk = BodyPrefetch.Gather(view, _session, _q.Keys, start,
```
Avick3110, line 81: **The gather is eager over the chunk; the render is lazy to `max_chars`.** `Row(0)` fills the whole chunk, so bodies for rows the render never reaches are gathered anyway — and each of those rows' source plugins costs a full enumeration.

`BodyPrefetch`'s doc says "a render that stops at max_chars pays for the chunk it reached", but that only holds when `ChunkRows` is smaller than the selection. At 2,000 against the default `limit=500` the chunk *is* the whole selection, so it never holds on the common path.

Worked case: a `limit=500` fields scan whose winners live in ~200 plugins, `max_chars` cutting the render at ~40 rows. Before: ~40 typed seeks in ~40 plugins. After: ~200 plugin enumerations, 160 of them for rows nobody sees. It compounds on the auto-spill path, where `Render()` runs twice around the artifact write (RecordsTools.cs:1461-1471) — three gathers of the same chunk.

Passing the render's own row budget down (or `Math.Min(ChunkRows, limit)`) would keep the per-plugin win without paying for rows the render will not reach.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in ed3ce9ad. The call site now passes `ct` — and it needed more than the argument: the off-order arm of `ResolveBatchFromPole` polled no token anywhere, so passing it alone would have been inert. Both of its loops check it now (the file walk that collects the wanted keys, and the per-record `ReadFields` loop that is the ~30 ms half), and `OperationCanceledException` is rethrown ahead of the walk's catch-all so a cancel does not come back as `file "…" could not be fully read` — the same rule the off-order scan loop already followed. Pinned by `ACancelStopsTheOffOrderBodyRead`, which fails before the fold because nothing on that arm threw.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 06cab04b. `RenderBudget.Refuse` now runs in the off-order lane too, on the same two numbers and before a body is read, and the read is clocked and reports `render_ms` / `rows_read` on both transports. Tests: `AnOffOrderRenderOverTheBoundRefusesToo` and `AnOffOrderBodyRenderReportsWhatItsBodiesCost`. The `limit=` description and the changelog line now say the bound holds on a scan, on an off-order `source=` and on a `formids=` list alike.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 9dd1c055. The list lane refuses on the same two bounds, checked over the LIST length before any body is read, with a remedy of its own: this lane reads every id before `limit=`/`offset=` window the render, so paging does not lower the cost and telling the caller to narrow scan terms it does not have would be wrong. `counts_only` is not exempt here (it reads the bodies anyway, unlike the scan lane's census), and a `formids=` set a WALK derived is exempt — `walk.max_nodes` bounds that already, and holding it to this bound broke `AWalkIsNotRefusedForTheSizeOfItsSeedScan`, which is the same decision under a different spelling. Tests: `AFormidsReadOverTheBoundRefuses`, `AFormidsCensusIsHeldToTheBoundBecauseItStillReadsEveryBody`.

One thing I did not do: this lane still reports no `render_ms`. It reads the whole list and renders a window of it, so the row count beside the milliseconds would be the window's and not what was read — an honest number needs a second parameter on the shared batch render. Your call; the cost is declared on `limit=` in the meantime.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 08028453. `OverlayPostBatch` takes a `CancellationToken` and polls it at the top of its per-record replay loop; the call site passes `ct`. Pinned end to end by `ACancelStopsTheOverlayPostRead`, which drives the tool with `source={"overlay":"skypatcher"}` and a cancelled token — it returned normally before. The PR's cancellation table now has a row for this lane and one for the off-order body read.

src/housecarl-mcp/ScanDetailReader.cs
Avick3110: Folded in 767618ec, though not by the knob you suggested. `Math.Min(ChunkRows, limit)` does not fix your own worked case: at `limit=500` the chunk becomes exactly the 500-row selection, and a render `max_chars` cuts at 40 rows still pays all ~200 plugin enumerations. Chunking smaller trades the other way — a render that completes then re-walks the same plugins once per chunk.

So the gather is deferred instead of resized. `BodyPrefetch.Gather` now builds only the row-to-plugin map; a plugin is enumerated on the first row that actually asks for its body, and that one walk covers that plugin's whole share of the chunk. Cost becomes the plugins the RENDERED rows came from: your case is ~40 walks instead of 200, a completed 500-row render is the same 200 it is today, and it can never exceed the per-row seek this replaced. The auto-spill's three renders each pay only for what they render.

Pinned by `AChunkWalksOnlyThePluginsTheRenderedRowsCameFrom` on a new `LoadOrderResolver.CollectPasses` counter, over twelve small AMMO plugins added to the world (one plugin cannot show this): five rows from one plugin cost 1 walk, and 12 with the gather made eager again.

PR #597, opened 2026-09-06
Spell every write refusal in the member the caller wrote

Avick3110:
Four places still spelled a refusal in the deleted 1.x vocabulary — a member or a parameter no 2.0 surface declares, so a caller reading one was handed a name they could not act on. #471 enumerates them; this branch takes them together, because they are one defect wearing four hats: a value or default that no shipped caller selects, and that nothing asserts, so it stays wrong quietly.

Part of #471

**1. `CreateOpNaming.Legacy` — deleted.** The struct let a calling tool hand the create engine its own words for an op, and `Legacy` spelled them the 1.x way (`op[i]`, `CopyFrom / from_plugin`) while also being the default. It was constructed at exactly two sites, both with the same value `("ops", "op=\"CopyFrom\"")`: `CreateTools.cs` (the only production caller, `housecarl_create`) and `ProbeCreate.cs`. `Legacy` was named at one site only, the `naming ?? CreateOpNaming.Legacy` default in `CreateRecordsBatch`. `CreateRecordsBatch` is the sole caller of `BuildCreateSpec`, which is the sole caller of `MapCreateEdit`, which is where `naming` was read. The 1.x tools that used to select the default — `housecarl_create_record` and `housecarl_bulk_create` — were deleted at #468, and the service's single-record overload at #497. One value left is not a choice, so the struct, the `naming` parameter on the three methods, and the two argument sites are gone, and `MapCreateEdit` spells `ops[i]` and `op="CopyFrom"` directly.

**2. The `origins` fallback — deleted (`fa86d2a3`).** `CreateRecordsBatch` also took an `origins` list so a caller could label each spec in its own words, and fell back to `record[{r}]` when it was absent — the singular of the `records=` member the caller actually wrote. Both callers that passed it passed exactly `records[i]`, index for index: `CreateTools.cs` adds `records[{i}]` for every spec in the loop that builds the wire list, and `ProbeCreate.CreateOne` passes `records[0]` for its batch of one. So the parameter held one value, and the fallback was the only way to render anything else — the eight probe and test call sites that omit `origins` took it, this PR's own first test among them. The parameter is deleted and the label spelled `records[{r}]` at the one point it is read; `BuildCreateSpec`'s `where` is now non-null, since its null branch had no caller either. `CreateOpSpellingTests` now pins `records[0]: ops[0]: field_path is required` and the absence of both `op[0]:` and `record[0]:`.

**3. The `sourceParam = "from_plugin"` defaults — deleted (`7559a809`).** The forward engines take the calling tool's spelling for the source pole so a shared refusal names a parameter the caller can see, and all four entry points defaulted it to the deleted `housecarl_forward_record`'s word. Per site:

- `WritePatchBuilder.ForwardRecordsInPlace` — one caller in the tree, `LoadOrderService.ForwardRecordsInPlace`, which passes it.
- `LoadOrderService.ForwardRecordsInPlace` (private) — one caller, `LoadOrderService.ForwardRecords`, which passes it.
- `LoadOrderService.ForwardRecords` — `ForwardTools.cs`, the only shipped caller, passed `source=`; the fifteen probe and test callers omitted it and took `from_plugin`.
- `WritePatchBuilder.ForwardRecords` — `LoadOrderService` passes it; eleven `ForwardFromPluginProbe` calls omitted it.

Disposition, since it differs by layer: the two service entry points hold `source=` directly (one const, `ForwardSourceParam`), because one value is not a choice and that layer owns the surface word — `ForwardTools` no longer passes anything. The two core entry points now REQUIRE the spelling rather than defaulting it, because `housecarl-core` has no surface of its own to name; the forward probe passes `sourceParam: "source="`, the word the shipped tool passes, for the same reason `ProbeCreate` matches the create tool. `WriteSurfaceGuardProbe.cs:742/752` already asserted the shipped forward tool must never say `from_plugin`; nothing asserted the defaults, which is why they survived.

**4. `op[i]` on an inline apply op — fixed (`843b4dff`), and this one was live.** `ApplyTools.cs:146` rendered `e.Origin ?? $"op[{i}]"`, and `Origin` is set only for zip-generated edits, so every ordinary `housecarl_apply` call took the fallback and was refused against a member spelled `ops=`. `LoadOrderService.MapEdit` repeated it, so the engine's own refusals (`formid is required`, `field_path is required`) carried it too. Both now spell `ops[{i}]`. `ApplyOpSpellingTests` covers one altitude each, through the tool: the `from=` check in `ApplyTools`, and `formid is required` from the engine. Measured: both fail before the change (`ops[0]: from= names the SOURCE RECORD` and `ops[0]: formid is required` are not in the refusals, which say `op[0]`) and pass after. This is the one part of the branch a user can see, so it has a changelog line.

**What is left of #471, and why this is `Part of` rather than `Closes`.**

- **`RemoveRecords`' lane refusals are reachable and pinned — left standing.** The issue lists the three lane-contract arms and the `patch is required` arm as residue. They are unreachable from `RemoveTools.Remove`, which refuses each shape first, but not unreachable from the service: `InPlaceProbe`'s arm U calls `LoadOrderService.RemoveRecords` three times to drive all three lane arms and asserts each sentence (`requires target=`, `mutually exclusive`, `only meaningful with in_place`), and `ExtendResolveProbe`'s 8d drives the `patch is required` arm and asserts it names no lane spelling — the arm Aaron's 2026-09-05 ruling is about ("a direct service call hands none, so it names none"). They are an asserted contract at the service altitude, not unguarded residue, and they render no 1.x word. Deleting them would redden two probes, so they stay.
- **The derivation the issue actually asks for is not here.** Its closing ask is the population computed rather than remembered — every parameter whose default encodes a 1.x spelling and whose every production caller overrides it — plus a guard asserting that set is empty. That is a separate cut, and the four instances above are what a contained change could take.

**Run**, all from the worktree: `dotnet build housecarl.sln -c Release` (0 errors); `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` — 1623 passed, 0 failed. `ci-all` left to CI (#570); the probes on the paths this branch touches were run alone and all pass: `write-surface-guard` (233 passed, 0 failed), `bulk-create-guard`, `extend-resolve-guard`, `dialogue-ckparity-guard`, `forward-from-plugin-guard`, `apply-guard` (80 passed, 0 failed), `inplace-guard`, `dry-run-guard`.


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7135,7 +7134,7 @@ public WritePatchBuilder.CreateOutcome CreateRecordsBatch(IReadOnlyList<CreateOp
             // origins[r] is the caller's own spelling for this spec, carried parallel to the list for the same
             // reason ApplyEdits carries opOrigins: a refusal must never name an index shape the caller did not write.
             var where = origins is not null && r < origins.Count && origins[r] is { } o ? o : $"record[{r}]";
```
Avick3110: [medium] The `origins` fallback on this line is the same defect as the `naming ?? CreateOpNaming.Legacy` default deleted on the next line, and #471 names it explicitly. No shipped caller omits `origins`: `CreateTools.cs:119` adds `records[i]` for every spec, and `ProbeCreate.cs:27` supplies `records[0]`. So the only thing `$"record[{r}]"` can ever render is `record[0]` -- a label no 2.0 surface declares, and the singular of the `records=` member the caller actually wrote.

The PR's own new test renders it: the string `CreateOpSpellingTests` asserts against is `record[0]: ops[0]: field_path is required.` The test that proves the fix is standing on the residue.

This is one line away from the change, in the same method, in the same call. Close it the same way -- make `origins` a required parameter alongside `records` (both surviving callers already pass it), or refuse when it is absent or short -- so the class does not survive in the method it was just removed from.

Avick3110 (commented):
Blind review. Built the branch (`dotnet build housecarl.sln -c Release`, 0 errors), ran the new test (1 passed), `write-surface-guard` (233/0), `bulk-create-guard` (PASS), `bulk-primitives-wave3-guard` (57/0), `extend-resolve-guard` (ALL PASS), `dialogue-ckparity-guard` (PASS), and the two driven test families that call `CreateRecordsBatch` without `naming:` (`OwnedChildLifecycleDrivenTests` + `RuntimeFormIdTests`, 44/0). `ci-all` not run.

What holds. `CreateOpNaming` has zero remaining references in the tree. The direct spellings match what `CreateTools` passed exactly: `Element "ops"` -> `$"ops[{index}]"`, `CopySubject "op=\"CopyFrom\""` -> the same literal in the refusal, so `WriteSurfaceGuardProbe.cs:1859` (which pins `op="CopyFrom" copies from an EXISTING record` and the absence of `from_plugin`) is unaffected. Unreachability holds for shipped callers: `CreateRecordsBatch` is called from exactly one place in `src/housecarl-mcp` (`CreateTools.cs:122`); the other eight call sites are probes and tests, and none of them asserts on the `op[i]` spelling. The new test does discriminate -- before the change the same call renders `record[0]: op[0]: field_path is required.`, so `Assert.Contains("ops[0]: field_path is required")` fails, and `"op[0]:"` is not a substring of `"ops[0]:"` so the negative assert is not vacuous.

Two findings below, one inline.

**[medium] `Closes #471` closes an issue whose remaining instances the PR neither fixes nor names.** The issue body enumerates four instances, not one: `CreateOpNaming.Legacy` (fixed here), the `origins` default (inline comment below), `sourceParam = "from_plugin"` at `src/housecarl-core/WritePatchBuilder.cs:1751` and `:2231` and `src/housecarl-mcp/LoadOrderService.cs:6141` and `:6220` -- which the issue reports as measured dead, both defaults rewritten to `ZZDEADPARAM` with `ci-all` 133/133 -- and `RemoveRecords`' unreachable lane refusals. All four are still present on this branch; I confirmed the `sourceParam` defaults by grep. The issue's own closing ask is the derivation plus a guard over the class. The ratified disposition for this cluster is close-by-construction *or* remaining work named; the PR body names nothing. Either narrow to `Refs #471` and file the remainder, or add a paragraph saying what is left and why it is a separate cut.

**[medium, adjacent -- a follow-up, not a change here]** The 1.x op label is still live on a *shipped* surface. `src/housecarl-mcp/ApplyTools.cs:146` uses `var where = e.Origin ?? $"op[{i}]"`, and `Origin` is set only for zip-generated edits (`ApplyTools.cs:249`), so an inline op -- the ordinary `housecarl_apply` call -- is refused as `op[0]: from= names the SOURCE RECORD ...` against a member spelled `ops=`. `LoadOrderService.MapEdit:7427` repeats the same `origin ?? $"op[{index}]"`, so the engine-side refusals (`op[0]: formid is required.`, `op[0]: field_path is required.`) carry it too. Nothing pins this in the probes or the test project. It is the exact caller-invisible spelling this PR removes from create, still reachable on apply, and it makes the two sibling surfaces disagree about the same `ops=` member.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in fa86d2a3: the `origins` parameter is deleted rather than re-spelled, because both callers passed exactly `records[i]` index for index, so `CreateRecordsBatch` now spells `records[{r}]` at the one point it is read (and `BuildCreateSpec`'s `where` is non-null, its null branch having had no caller either) — the test's expected string is now `records[0]: ops[0]: field_path is required`, and it asserts the absence of `record[0]:` as well as `op[0]:`.

Avick3110:
Both body findings folded; the inline one is answered on its own thread.

**`Closes #471` with three instances untouched** — folded. The `sourceParam` defaults are gone (7559a809: the two service entry points hold `source=` directly, the two core ones now require the spelling and the forward probe passes it) and the `origins` fallback with them (fa86d2a3), so three of the four instances are in. The fourth is not: `RemoveRecords`' lane refusals are unreachable from `RemoveTools.Remove` but reachable and asserted from the service — `InPlaceProbe` arm U drives all three lane arms and pins each sentence, and `ExtendResolveProbe` 8d pins the `patch is required` arm — so they are a guarded contract rather than residue, and deleting them would redden two probes. With that left standing and the issue's real ask (the derived population plus a guard over it) out of scope for this cut, the PR is now `Part of #471` and the body names both.

**The 1.x op label still live on apply** — folded here rather than filed (843b4dff). `ApplyTools.cs:146` and `LoadOrderService.MapEdit` both spell `ops[i]` now, with a test per altitude in `ApplyOpSpellingTests`; measured, both fail before the change (the refusals say `op[0]`) and pass after, and the apply wording is the one user-visible part of the branch, so it has a changelog line.

Avick3110:
Rebased onto the merged description wave (`cd6b3139`). All six commits replayed in order, none squashed; the branch is now `MERGEABLE`.

One file conflicted. Resolution per file:

- **`plugin/CHANGELOG.md`** — conflict. Both sides added at the top of `## Unreleased`, and main's `a8dbb8f6` rewrote the very entry this branch was carrying under it. Kept main's consolidated wave entry ("Every tool description now fits inside the 2,048 characters…") and dropped this branch's now-superseded single-tool `housecarl_records` copy of it, which was the same entry in its pre-wave wording. This branch's own `housecarl_apply` / `ops[i]` entry is kept, above it, in the file's newest-first order. Net: both sides' current entries present, no duplicate.
- **`src/housecarl-mcp/ForwardTools.cs`** — no conflict. Main's rewrite is confined to the tool and parameter descriptions (lines 21–70); this branch's only hunk is the `svc.ForwardRecords(...)` call at line 97, which dropped `sourceParam: "source="`. Auto-merged, both changes present.
- **`src/housecarl-mcp/LoadOrderService.cs`** — no conflict. Main's `5f0c2d6c` added `with { Stamp = offEpoch }` to the `ResolveOutputPath` catch at 6202; this branch's hunks sit on the lines either side (`ForwardRecordsInPlace(...)` above, `WritePatchBuilder.ForwardRecords(..., ForwardSourceParam, ...)` below). Auto-merged, both present and verified by reading the block.

No description needed realigning to the code. Main's new parameter text spells the forward source pole `source=` throughout and never names an op or record index, and a grep of `src/housecarl-mcp`, `src/housecarl-core` and `src/housecarl-generator` finds no remaining `op[` or `record[` label outside the two probe comments that deliberately assert the old spelling is *absent* (`ApplyGuardProbe.cs:724-731`, `WriteSurfaceGuardProbe.cs:1844-1851`) and the changelog sentence quoting the old name. The `from_plugin` hits that remain are the declared `ApplyOp.from_plugin` member and its alias-table entry, not the deleted default.

**Run**, all from the worktree, after the rebase: `dotnet build housecarl.sln -c Release` — 0 errors. `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` — 1622 passed, 0 failed; `PublishedDescriptionBoundTests`, `ApplyOpSpellingTests` and `CreateOpSpellingTests` re-run alone together, 34 passed, 0 failed. The eight probes on the touched paths, each run alone: `write-surface-guard` (233 passed, 0 failed), `bulk-create-guard` (PASS), `extend-resolve-guard` (ALL PASS), `dialogue-ckparity-guard` (PASS), `forward-from-plugin-guard` (ALL PASS), `apply-guard` (80 passed, 0 failed), `inplace-guard` (PASS), `dry-run-guard` (ALL PASS). `ci-all` left to CI.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -6133,16 +6137,13 @@ WritePatchBuilder.RemovalOutcome RemoveRecordsInPlace(
...
-    /// expose.</param>
     public WritePatchBuilder.ForwardOutcome ForwardRecords(IReadOnlyList<string> formids, string fromPlugin, string? patchName, string? into,
         bool fullReadback = false, string? target = null, bool inPlace = false, bool acknowledge = false,
-        bool dryRun = false, string sourceParam = "from_plugin")
+        bool dryRun = false)
     {
         if (string.IsNullOrWhiteSpace(fromPlugin))
             return WritePatchBuilder.ForwardOutcome.Fail(
-                $"{sourceParam} is required — name the plugin whose version of the record(s) to forward (the earlier override, or a master to revert to vanilla).");
+                $"{ForwardSourceParam} is required — name the plugin whose version of the record(s) to forward (the earlier override, or a master to revert to vanilla).");
```
Avick3110, line 6146: The same defect this PR takes for `op[i]` is still live twenty-five lines below this hunk. `ForwardRecords` labels a malformed FormID `formid[{i}]:` at lines 6171 and 6173, and `RemoveRecords` does the same at 6039/6041 — but both shipped tools declare `formids=`, plural. A caller who passes a bad FormID gets

```
refused — 1 of 1 formid(s) malformed; NOTHING forwarded:
  - formid[0]: 'ZZZ:Mod.esp' …
```

and goes looking for a `formid` parameter that no 2.0 surface publishes — the same "handle that points at nothing in the call" the changelog line is about. It reads as an in-scope fifth hat on the defect rather than part of the computed derivation left to a later cut, and it is in one of the two methods this PR already edits.

Avick3110:
**`src/housecarl-generator/ApplyGuardProbe.cs:731` — the arm's negative pin now names a string nothing can emit.** (Posted here rather than inline: that file is not in this PR's diff.)

```csharp
// REVIEW FOLD [medium]: a zip-generated op must never be refused at an op[i] the caller never wrote.
// Two real ops here, so a naive index would say "op[2]" — a line that does not exist in the call.
…
    badPair.StartsWith("error:") && badPair.Contains("assignments[0]") && !badPair.Contains("op[2]"), badPair);
```

After this branch, the phantom label that regression would produce is `ops[2]`, and `"ops[2]"` does not contain `"op[2]"` — so `!badPair.Contains("op[2]")` is vacuously true and the clause written for this exact regression can no longer fire. If a later change drops the `Origin` thread and a zip edit falls back to its wire index, only the sibling `Contains("assignments[0]")` catches it; the pin the comment describes does not. Spelling it `!badPair.Contains("ops[2]")` restores it, and it is the same residue class the branch is clearing.

(For contrast, `WriteSurfaceGuardProbe.cs:547`'s `!badType.Contains("record[1]:")` stays a live regression pin — the singular could still be reintroduced by hand — so that one reads fine as-is.)

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 4d9652d0. Real, and confirmed by reading both sites: `ForwardRecords` (LoadOrderService.cs 6171/6173) and `RemoveRecords` (6039/6041) built the label as `formid[{i}]`, while `ForwardTools` and `RemoveTools` both declare `formids=`. All four sites now spell `formids[{i}]`, on both the empty-token branch and the `FormIdDoor.Sentence` prefix/malformed pair.

New `FormidsSpellingTests` adds one test per tool, driving the shipped surface so the assertion is on the caller-visible string: each passes a malformed FormID, asserts `formids[0]` is named and `formid[0]` is absent. Both refusals land at the parse, before the write gate, so nothing is written. Existing changelog line retitled from the apply-only wording to cover all three tools.

Full suite 1624 passed, 0 failed.

Avick3110:
Folded in d1e96f7a. Real: `ops[2]` does not contain `op[2]` (the character after `op` is `s`, not `[`), so after this branch `!badPair.Contains("op[2]")" could never be false and the clause the comment describes was dead. Line 731 now pins `!badPair.Contains("ops[2]")`, and the two comment lines above it say `ops[i]` / `ops[2]` to match.

Ran the probe alone from the worktree: `housecarl-generator.dll apply-guard` -> 80 passed, 0 failed, including the arm itself ("a bad FormID in an assignment is refused NAMING THE ASSIGNMENT, not a phantom op index"). The refusal genuinely carries `assignments[0]` and no `ops[2]`, so the pin is live rather than vacuous now.

Agreed on `WriteSurfaceGuardProbe.cs:547` — `record[1]:` is a spelling that could still be reintroduced by hand, so it stays as-is. Left untouched.

Full suite: 1624 passed, 0 failed.

PR #598, opened 2026-09-06
Bring the compact_plugin tool description under the 2,048-character cap

Avick3110:
`housecarl_compact_plugin` published a 3,014-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the output lanes, the external-referencer safety and its cost, the localized-plugin rules, the off-order resolve, the flag-only lane and the whole refusal list were sitting in text the model never saw. Nothing is deleted: each paragraph moves into the description of the parameter it is about — the target rules to `plugin`, the window and the ESL lanes to `esl`, the two output lanes and the localized rules to `in_place`, the external-referencer safety to `repoint_externals` — and where the parameter already said the same thing weakly, the description's fuller version replaces it rather than being pasted in beside it. The description keeps what belongs to no single parameter: the one-sentence purpose with what a compaction does to FormIDs, the Papyrus caveat, and one new line naming which parameter carries which grammar. `compact_plugin`'s line comes out of `StillOversized`.

Part of #583.

| | Before | After |
|---|---|---|
| `housecarl_compact_plugin` description | 3,014 | **960** |
| Parameter descriptions (all six) | 2,017 | 3,709 |
| Both together | 5,031 | 4,669 |

The description shed 2,054 characters and the parameters gained 1,692; the 362-character difference is text the description was repeating from a parameter that already carried it, plus the two folds named in the table below.

## Harvest

|---|---|
| "COMPACT a plugin's FormIDs — the data-layer twin of xEdit's 'Compact FormIDs for ESL'." | kept in description |
| "Renumbers EVERY record the plugin DEFINES (its originating records — flat AND nested: cells, placed references, dialogue lines, navmesh, landscape) … repoints every reference WITHIN the plugin, leaves its overrides of other mods at their master FormIDs" | kept in description — this is what a compaction does whatever `esl=` is |
| …the same sentence's "into the light/ESL range 0x800–0xFFF (the 2048-ID window)" and "flags the result a light master (ESPFE) so it frees a load-order slot" clauses | `esl` — both are the `esl=true` branch, and `esl` already carried the range and the ESPFE flag in the same words; only "which frees a load-order slot" was new there, so that clause is what got added |
| "esl=false instead renumbers contiguously from 0x800 with NO light flag/ceiling (to close FormID gaps)." | `esl` (already said there in the same words) |
| "OUTPUT (default): a NEW plugin keeping the SOURCE'S EXACT basename (so other mods that list it as a master still resolve) in a fresh houseCARL mod folder — your ORIGINAL is untouched" | `in_place`, folded onto its weaker "OMIT (the default) to write a NEW plugin (same basename, fresh mod folder) and leave the original untouched for review" — that sentence stated the lane without saying why the basename is kept, so the description's version supplies it. The "(default)" heading is dropped as the marker word: `in_place`'s own text already spells its default, and the parameter is where a caller reads it |
| "review the new one in xEdit, then in MO2 enable its folder and DISABLE the original mod (same basename — MO2 serves one)" | `in_place` — this is the new-plugin lane's swap procedure and it is false for `in_place=true`, where there is no folder to enable and nothing to disable, so it does not belong to the whole operation the way merge's AFTER paragraph does |
| "in_place=true instead OVERWRITES the original (xEdit's norm; rides the in-place consent, NO backup; needs acknowledge=true)." | `in_place` (already said there, except "rides the in-place consent", which is folded into its "Rides the in-place consent: requires acknowledge=true") |
| "THE SAFETY (Q3): renumbering breaks any reference from OUTSIDE this plugin (they'd point at FormIDs that vanish)." | `repoint_externals` |
| "houseCARL scans the WHOLE load order for such external referencers (a one-pass walk — can take ~25s on a big order): if NONE, it's a clean compaction; if SOME, the call is REFUSED and lists them" | `repoint_externals`, folded onto its weaker "the call REFUSES (listing them) by default" and scoped: the engine skips the whole-order walk when the remap is empty, so it reads "whenever there is anything to renumber" rather than claiming the scan is unconditional |
| "UNLESS repoint_externals=true, which ALSO rewrites each of them in place to follow the renumber (needs acknowledge=true; no backup of them either)." | `repoint_externals` (already said there in the same words, in the second person the parameter uses) |
| "LOCALIZED PLUGINS: houseCARL does not rewrite one in place — its text lives in separate .STRINGS files it cannot swap together with the plugin — so in_place=true on a localized target is refused" | `in_place`, folded onto its weaker "A LOCALIZED plugin is REFUSED in this lane whatever arrangement its .STRINGS files are in" — that sentence stated the refusal without the reason, so the description's version supplies it |
| "…and so is repoint_externals=true when any referencer is localized (the refusal names it and where its text is, so you learn that BEFORE choosing the flag)." | `repoint_externals`, folded onto its weaker "the default refusal above says so up front rather than sending you here", which now carries the description's detail: "names it and where its text is" |
| "The DEFAULT new-file lane still compacts a localized plugin: the output is DE-LOCALIZED — the text this read resolved is written into the plugin itself and the source's .STRINGS files no longer describe it — and the report says so." | `in_place`, folded onto its weaker "the new-plugin lane compacts it, DE-LOCALIZED, and says so" |
| "Read the output before swapping it in." | `in_place`, folded onto the swap procedure two sentences above it so the instruction is given once: it now points at that review step and says what to read there ("read the output's TEXT"), rather than restating "review before you swap" a second time in the same parameter |
| "The target need NOT be active: a plugin on disk but not (yet) in the load order — e.g. the patch houseCARL just wrote, before the MO2 refresh — is resolved by filename across ALL mod folders and compacted OFF-ORDER" | `plugin`, folded onto its weaker "a plugin on disk but not in the order (a fresh houseCARL patch, a disabled mod) is resolved by filename and compacted OFF-ORDER" — the parameter's version had neither "across ALL mod folders" nor the "before the MO2 refresh" qualifier on its example, and both are now on it |
| "(its declared masters must still be active)" | `plugin` (already said there as "its declared masters must still be active"; that sentence was the weaker one — it stated the requirement without the refusal — so it now carries the refusal list's force: "a declared master not active is refused loud and nothing is written", and the refusal list drops that clause) |
| "An override-only plugin with esl=true takes the FLAG-ONLY lane: nothing to renumber, every record copies verbatim, the ESL flag is set (always valid — the light window only constrains originating records)." | `esl` |
| "Refuses loud + writes nothing on: the plugin found nowhere on disk / ambiguous across folders / unparseable … a serialize fault." | `plugin` — every clause of that list is about the target plugin's own resolve and write |
| …the same sentence's "an override-only plugin with esl=false (nothing to do)" clause | `esl`, said beside the flag-only lane it is the other half of |
| …the same sentence's "MORE records than the light range holds (the hard 2048 ESL ceiling — named, never truncated)" clause | `esl`, scoped "with esl=true": with `esl=false` the ceiling is the full 24-bit object-ID range, not the light window |
| …the same sentence's "a declared master not active" clause | `plugin` (folded into the masters sentence, per the row above) |
| "Note: references compiled into Papyrus scripts (.pex hardcoded FormIDs / GetFormFromFile) are NOT remappable — verify scripted records after compacting." | kept in description — it is a property of any renumber, in either lane and at either window |

`acknowledge` and `patch_name` received nothing: no sentence of the description was about the consent flag or the mod folder beyond the "needs acknowledge=true" clauses their own descriptions already carry.

The description gained one thing it did not carry before: a line naming which parameter holds which grammar. With the grammar moved out, that roster is the description's only index of the parameters, and a caller who reads only the 2,048 characters the client shows needs to know where the rest is.

Two notes on what is not here. `compact_plugin`'s description carries no "ONE surface" composition sentence and no cross-tool pointer, so there was none to keep — `merge_plugins` points at this tool, not the other way round. It also carries no sentence about the facegen / voice / .seq carry, which the operation does perform and which the report states; adding one would be new text, not a move, so this PR does not. Nor does it carry a sentence about the runtime INI distributor layer (#364) — that reminder is rendered from `WriteSentences.CompactRuntimeConfigs` and has never been in the description, so there was nothing to place.

## Test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`: the `ToolNames.CompactPlugin` line comes out of `StillOversized`, and nothing else in the file changes. Run with that line gone and the description unchanged, `EveryPublishedDescriptionFitsTheClientsCut` failed on `housecarl_compact_plugin` at 3,014 characters and was the only failure in the class; after the change the class is green and prints 960.

## What was run

```
    1620 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. No CHANGELOG line: the records PR's entry covers the wave.


src/housecarl-mcp/WriteTools.cs
```diff
@@ -55,41 +55,24 @@ public static string CreatePlugin(
...
+        [Description("Optional, default false. THE SAFETY (Q3): renumbering breaks any reference from OUTSIDE this plugin (they'd point at FormIDs that vanish), so whenever there is anything to renumber houseCARL scans the WHOLE load order for such external referencers (a one-pass walk — can take ~25s on a big order): if NONE, it's a clean compaction; if SOME, the call REFUSES (listing them) by default. Set true to ALSO rewrite those external referencers IN PLACE to follow the renumber (requires acknowledge=true; no backup of them either). Refused when any referencer is LOCALIZED — houseCARL does not rewrite one in place — and the default refusal above names it and where its text is, so you learn that BEFORE choosing this flag rather than being sent here.")]
```
Avick3110: **medium** — `repoint_externals` is now the sole home of the external-referencer grammar, but it still does not say the flag requires `in_place=true`.

`LoadOrderService.CompactPlugin` refuses the pairing the parameter describes:

```csharp
if (id.HasExternalReferencers)
{
    if (!repointExternals) { ...refuse and list them... }
    if (!inPlace)
        return WritePatchBuilder.CompactOutcome.Fail(
            $"refused — repoint_externals requires in_place=true. ...");
}
```

Scenario: a caller reads only this parameter (the whole point of the move) and calls `compact_plugin(plugin: "CoolMod.esp", repoint_externals: true, acknowledge: true)`. Whenever there actually are external referencers — the only case the flag exists for — the call is refused and nothing is written. The engine's own default refusal spells the remedy out ("Re-run with repoint_externals=true AND in_place=true (+ acknowledge=true)"), so the fact is stated at runtime but missing from the parameter that is supposed to carry the grammar; the description's own roster line points a caller here for "the external-referencer safety".

Worth adding `requires in_place=true` alongside the `requires acknowledge=true` clause that is already there.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -55,41 +55,24 @@ public static string CreatePlugin(
...
+        [Description("Optional, default false. IN-PLACE LANE (opt-in): OVERWRITE the original plugin with its compacted form (xEdit's norm) instead of writing a new file — NO houseCARL backup or undo (keep your own). Rides the in-place consent: requires acknowledge=true. OMIT (the default) to write a NEW plugin instead, keeping the SOURCE'S EXACT basename (so other mods that list it as a master still resolve) in a fresh houseCARL mod folder, leaving the original untouched: review the new plugin in xEdit, then in MO2 enable its folder and DISABLE the original mod (same basename — MO2 serves one). LOCALIZED PLUGINS: houseCARL does not rewrite one in place — its text lives in separate .STRINGS files it cannot swap together with the plugin — so a LOCALIZED plugin is REFUSED in this lane whatever arrangement its .STRINGS files are in. The new-plugin lane still compacts it: the output is DE-LOCALIZED — the text this read resolved is written into the plugin itself and the source's .STRINGS files no longer describe it — and the report says so. The review step above is where you catch that: read the output's TEXT before swapping it in.")]
```
Avick3110: **medium** — "The new-plugin lane still compacts it" is stated unconditionally, and the engine has a new-file-lane refusal that contradicts it.

`LoadOrderService.CompactPlugin`, immediately after the in-place localized refusal:

```csharp
if (LocalizedStrings.ResolvesNowhere(srcShape.Shape))
    return WritePatchBuilder.CompactOutcome.Fail(
        UnresolvableStringsRefusal(name, srcShape, "compact"));
```

and `LocalizedStrings.ResolvesNowhere` is true for `Nowhere`, `StringsFolderUnreadable` and `ModFolderUnreadable`.

Scenario: a localized plugin whose `Strings\` folder is missing or unreadable, compacted with the default `in_place=false`. The parameter promises a DE-LOCALIZED output plus a report saying so; the call is refused instead, because the read would bake blanks in. A caller who trusted this text plans a review-and-swap that never gets an output to review.

The old text carried the same absolute claim, so this is not introduced here — but this PR makes `in_place` the single home of the localized grammar and expands the sentence, so it is the place to scope it (e.g. "…unless its .STRINGS resolve nowhere, which this lane refuses too").

src/housecarl-mcp/WriteTools.cs
```diff
@@ -55,41 +55,24 @@ public static string CreatePlugin(
...
          "verify scripted records after compacting.")]
     public static string CompactPlugin(
         LoadOrderService svc,
-        [Description("The plugin's filename to compact (e.g. 'CoolMod.esp'). Usually active in your load order; a plugin on disk but not in the order (a fresh houseCARL patch, a disabled mod) is resolved by filename and compacted OFF-ORDER — its declared masters must still be active. The compacted output keeps this EXACT basename.")]
+        [Description("The plugin's filename to compact (e.g. 'CoolMod.esp'). The target need NOT be active: usually it is in your load order, but a plugin on disk and not (yet) in it — the patch houseCARL just wrote, before the MO2 refresh; a plugin inside a disabled mod — is resolved by filename across ALL mod folders and compacted OFF-ORDER. Its declared masters must still be active: a declared master not active is refused loud and nothing is written. The compacted output keeps this EXACT basename. Refuses loud + writes nothing on: this plugin found nowhere on disk / ambiguous across folders / unparseable; a serialize fault.")]
```
Avick3110: **low** — the master-must-be-active rule reads as scoped to the OFF-ORDER lane, but the engine applies it on every lane.

The old description carried "a declared master not active" as an unscoped item in its refusal list; the harvest folds it into the off-order sentence and drops it from the list. Read as written, "Its declared masters must still be active" takes its subject from the preceding clause — the plugin "on disk and not (yet) in it".

The refusal lives in `WritePatchBuilder.CompactBuild`, which runs for every target, active or not:

```csharp
var mp = resolveMasterPath(mfn);
if (mp is null)
    return CompactBuildResult.Fail(
        $"cannot compact '{modKey.FileName}': its declared master '{mfn}' is not active in the load order, ...");
```

Scenario: an ordinary ACTIVE plugin whose master the user unticked in MO2 (MO2 permits this — it only warns). Compacting it is refused for a reason the parameter now presents as an off-order-only condition. Either restate the clause in the refusal list or make the sentence's subject the target plugin generally.

src/housecarl-mcp/WriteTools.cs
Avick3110: **low** — "Refused when any referencer is LOCALIZED" names only half the blocking class, and the in-place lane has a third refusal this parameter does not reach.

`RemapEngine.LocalizedAmong` fails closed on a referencer it could not open, and both refusal sites say so in their own words: "houseCARL rewrites neither a localized plugin nor one it cannot read in place" — `BlockedReferencerCensus` exists precisely because the hits are not homogeneous. A referencer that is merely unreadable blocks the repoint just as hard, and the parameter attributes the refusal to localization only.

Separately, `LoadOrderService.CompactPlugin` refuses ANY in-place run when the identify pass could not read some plugin through:

```csharp
if ((willOverwriteTarget || willRepoint) && id.UnscannablePlugins is { Count: > 0 } unread)
    return WritePatchBuilder.CompactOutcome.Fail(...);
```

This fires before the consent gate and whatever `acknowledge` says, so a caller who set `in_place=true, acknowledge=true` on the strength of these descriptions can be refused for a condition none of the six parameters mentions. Since this PR designates `repoint_externals` as the home of "the external-referencer safety", the unscannable case belongs here (or on `in_place`).

Avick3110: Folded in 2d0591d4 — repoint_externals now says it requires in_place=true (so the target and its referrers move together) alongside acknowledge=true.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 2d0591d4 — the new-plugin lane sentence is now scoped: it refuses when the .STRINGS resolve nowhere, in the refusal's own words about reading back empty (or unknowable from a folder nothing could open).

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 2d0591d4 — the clause now takes the target as its subject and says so lane-neutrally: whichever lane it came from, active or not, its declared masters must still be active.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 2d0591d4 — repoint_externals now refuses on a referencer that is localized OR one houseCARL cannot read, and the unscannable-plugins refusal went on in_place, which is what the engine gates it on (the repoint arm is unreachable without in_place=true).

Issue #528, opened 2026-09-04
[Mutagen gap] Ingestible (ALCH) models no VirtualMachineAdapter, and the read note cannot tell not-modelled from a typo


Avick3110:
From a Discord report (2026-08-25, houseCARL 1.9).

Two halves.

1. Mutagen gap. Mutagen models `VirtualMachineAdapter` on 33 of 133 Skyrim record types and `Ingestible` (ALCH) is not one of them, though the game allows VMAD on ALCH. Scripted potions are invisible to houseCARL and `where=["VirtualMachineAdapter exists"]` on ALCH cannot work. Where Mutagen lags xEdit we say so; this issue is that statement.

2. Read note. On the `where=` lane this is already handled (`FieldPredicate.cs` says the field is not on these records). On the projection lane (`fields=[...]`), `ReadEngine.NoFieldNote` returns the same `(no field X)` for a typo and for a field the type does not model. The note should say which.

The note half is a candidate for W5 item 19 if Aaron promotes it. Not #35, not #408.

PR #600, opened 2026-09-06
Name an unreadable sub-field instead of dropping it

### LINKED ISSUE #560: [Bug] ReadEngine.Expand swallows nested read faults, so an unreadable sub-field renders as absent
## Summary

Found by the blind review of PR #550 and confirmed by the fold. `ReadEngine.Expand` swallows a read fault on a nested field (`catch { continue; }`), so a sub-field that throws during expansion at depth >= 2 is rendered as absent rather than as an unreadable field with a note.

## Why it matters

An absent optional and a field that could not be read look the same to the caller. Any consumer of the expanded body (the `rows` fold in #550, artifacts, the where= quantifiers from #551 that read "absent as empty") inherits the confusion. "A tool never returns a silently wrong answer" says the fault has to surface on the field.

## Fix shape

Expand emits the sub-field with `Readable=false` and the exception message as the note, the way the top-level read already does, and the callers that assume nested faults cannot happen (noted in #550's fold) get their claims back.

Avick3110:
A deep read (`project.depth` 2 or more) expands a substruct by reflecting its fields, and a getter that threw was skipped with `catch { continue; }`, so the sub-field was left out of the result entirely — indistinguishable from a field the record does not carry, which is the silently wrong answer the read surface is not allowed to give. The expansion now emits that sub-field's own line with the same `(unreadable: …)` note and the same `Readable=false` the depth-1 read already emits, and carries on with the siblings, so a batch, a probe or a catalogue keeps every other field on the row. The same treatment covers the two neighbouring silent skips in the same walk (a gendered arm that will not resolve or throws) and the reason is unwrapped out of reflection's `TargetInvocationException`, whose own message names nothing a caller can act on. Where the walk can say which it is — a name that came off the type's own reflection is modelled, so a miss there is a fault and not a coverage gap — it says so outright; #528's "cannot tell not-modelled from a typo" seam is untouched. `NavigateValue` now carries its readable/absent classification structurally rather than leaving the deep path's own fault line claiming `Readable=true`, so no render decides "could not look" versus "nothing there" by matching prose.

The isolation is per CHILD, not per getter: every recursion (substruct field, list element, dict value, gendered arm) goes through `ExpandChild`, which wraps `Expand` in the same try/catch. A fault under one child — its own token emit, its element count, a grandchild's getter — names that child's path and the sibling walk carries on. Unwinding to the caller instead dropped every later sibling (each then reading as absent) and pinned a second line on the parent's own path, which both row folds discard, so the fault could vanish entirely.

And the comparison forms carry it. `FieldsDiff` had no notion of `Readable`: a fault note entered the exact-path map as if it were a token, so two poles unreadable at the same path compared identical (`identical to <ref> (whole record)`), and an unreadable-vs-read pair was reported as a value difference — the engine asserting a reading it does not have. An unreadable leaf is now kept out of the comparison and clears `Complete`, and its path is named as its own delta with the read's own reason (`BasicStats.Damage: UNREADABLE here — not compared (unreadable: …)`), so such a record can be neither counted `identical` nor reported as differing in value: a no-verdict is a third state, carried on `FieldsDiff.Result` and counted as `no_verdict` rather than folded into `differing`. What it suppresses is scoped to that path and everything under it, plus the element comparison of a list it sits inside — a cap, whose whole line set is unreliable, keeps its record-wide suppression, but an unreadable leaf removes exactly one known path and the rest of the record still compares. Both comparison forms say the comparison was incomplete whenever it was, deltas or no deltas.

The two `Readable=false` answers are separated by the bit, not by prose kept per consumer: everything `Readable=false` is a no-verdict except the one comparable answer, a no-such-field (`ReadEngine.NoFieldPrefix` / `ReadEngine.IsNoSuchFieldNote`), which says something true about the record. Keying on the fault note's own prefix instead covered only the getter throw and let the walk's other fault notes — an FLOI it cannot read, a `*parent` hop it cannot take — back in as comparable values.

Closes #560

## A deliberate behaviour change

`ReflectedFieldNames` yields a name and `ResolveProperty` then resolves it; where it returned null, the old walk expanded a null value and the line rendered `(absent)`. It now renders as a fault. That branch is unreachable by construction, which is why the fault is the right answer for it rather than an absence: `ReflectedFieldNames` collects properties off `PrimaryGetter(t)` plus that interface's own interfaces (or, for a substruct, `t` plus its interfaces), and `ResolveProperty(t, name)` searches `t` plus `t.GetInterfaces()` — which is the transitive closure and therefore a superset of the set the name came from. A name that reached the resolve call has a declaration inside the closure being searched. If that ever stops holding it is a fault in the read walk, and saying "the record does not carry this field" would be the silently wrong answer.

## How it renders

Text (`format='text'`, `project.form='fields'`):

```
  BasicStats.Value = 34
  BasicStats.Weight = 5.5
  BasicStats.Damage = (unreadable: Specified argument was out of the range of valid values.)
```

JSON (`format='json'`) — the field object carries the note in place of a value, exactly as a top-level fault does:

```json
{ "path": "BasicStats.Damage", "note": "(unreadable: Specified argument was out of the range of valid values.)" }
```

`rows` (whose fold already keeps a line when `Readable` is false) and `everything` render from the same emitted lines. `delta` and `tree` render from the comparison, which names the path and refuses the identity claim.

## Tests

`src/housecarl-mcp-tests/RecordsUnreadableSubFieldTests.cs`. The fault is a real one, not an injected hook: the world writes a master and an override with Mutagen, then cuts the override's DATA declared length from 10 to 8 (with the enclosing record and group sizes), which is what a truncated or badly merged plugin looks like on disk. The weapon's record opens, its `BasicStats` substruct opens, `Value` and `Weight` read, and `Damage` throws inside Mutagen because it sits past the end of its own subrecord. The plugin is read through the off-order lane out of a switched-off mod folder — a record that faults during the active order's index build excludes its whole plugin, which is separate, already-loud behaviour — and the clean master it overrides is the delta's reference pole.

- `TheFieldsFormNamesTheSubFieldAndTheReasonAndKeepsItsSiblings` — before: no `BasicStats.Damage` line at all, assertion fails. After: passes.
- `TheJsonRenderCarriesTheFaultAsThatFieldsOwnNote` — before: no field object with that path, `Single` throws. After: passes.
- `AnEverythingReadKeepsTheRestOfTheRecordAndStillNamesTheFault` — before: the record's other fields are there but the fault line is missing, assertion fails. After: passes.
- `ADeltaNamesTheUnreadableFieldInsteadOfCallingItAValueDifference` — before: the delta reads `BasicStats.Damage=(unreadable: …)` as a value difference and the comparison claims to be complete, assertion fails. After: passes.
- `UnreadableLeafComparisonTests.TwoUnreadablePolesAtOnePathAreNotIdentical` — the pole-symmetric half, driven at `FieldsDiff` because one truncated plugin cannot be both poles of a lane call. Before: `Complete` is true and there is no delta, so the record renders identical; assertion fails. After: passes. `ANoSuchFieldStillCompares` is its control — the other `Readable=false` answer stays comparable.

New from Aaron's round, in the same file: `ADifferingListBesideAnUnreadableLeafIsStillCompared` and `AnUnreadableLeafInsideAnElementSuppressesOnlyThatList` pin the scoping, `AnFloiFaultIsNotComparedEither` pins the gate on the bit rather than the prose, `UnreadableLinkLeafReadTests.AnUnreadableConditionTargetIsNotReadAsAbsent` pins the readable bit through the deep read's link leaf (an index-mode FLOI, which is a live fault today), `ANoVerdictIsCountedApartFromValueDifferences` pins `differing=0 identical=0 no_verdict=1`, and `TreeIncompleteNoteTests` drives the tree render against a hand-built row for the note beside deltas.

`RecordsRowsFormTests.AnUnreadableSubFieldIsNotAnAbsentOne` carried a comment saying no read could produce such a line; that is no longer true, so it points at the new tests instead.

## Review

Seven findings from the blind review folded, one refuted (an arm in `RecordsWorldLifecycleTests`: nine worlds repoint the process-global corpus path and two have arms there, so a tenth would be one more per-world guard rather than the general arm the suite would actually need). Reasons are on each thread.

Aaron's round: all five folded — the gate keyed on the readable bit, the readable bit carried through the deep read's link leaf, suppression scoped to the unreadable path, the no-verdict counted as its own state, and the tree's incompleteness note moved beside its deltas. The scoped suppression is the judgement call of the five: it restores what this PR had been quieting (a differing list beside an unreadable scalar) and keeps `Complete` false, but reverting to record-wide suppression is a one-line change if that trade is wanted the other way. Reasons are on each thread.

## Run

From the worktree: `dotnet build housecarl.sln -c Release` clean, and `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` green at 1632 passed. `ci-all` was not run locally (#570); CI runs it.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -55,10 +55,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
   one. A client that stops waiting WITHOUT sending a cancel is not something the server can detect — nothing
   in the protocol reports it — which is what the render bound is there for.
 - **A field inside a record that cannot be read now says so, instead of rendering as absent.** Reading a
-  record deeply (`project.depth` of 2 or more) dropped the line for any sub-field whose read threw — a
-  truncated or badly merged plugin, most often — so a field that could not be read looked exactly like a
-  field the record does not carry. It now renders as `(unreadable: <reason>)` on its own line, naming the
-  sub-field and Mutagen's own reason, in every form and format; the sub-fields beside it still read.
+  record deeply (`project.depth` of 2 or more) dropped the line for any sub-field whose read threw — what a
+  truncated or badly merged plugin does to one field — so a field that could not be read looked exactly like
+  a field the record does not carry. It now renders as `(unreadable: <reason>)` on its own line, naming the
+  sub-field and Mutagen's own reason, and the fields beside it still read. `delta` and `tree` name it too:
+  an unreadable field is reported as not compared and leaves the comparison incomplete, so a record carrying
+  one is never called `identical`.
 
 - **Every tool description now fits inside the 2,048 characters Claude Code shows, and the grammar that was
   being cut off lives on the parameters it belongs to.** A description longer than that cap lost its tail on
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -54,6 +54,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
   the server running, and a `to_file=` call stopped that way writes no artifact at all rather than a partial
   one. A client that stops waiting WITHOUT sending a cancel is not something the server can detect — nothing
   in the protocol reports it — which is what the render bound is there for.
+- **A field inside a record that cannot be read now says so, instead of rendering as absent.** Reading a
+  record deeply (`project.depth` of 2 or more) dropped the line for any sub-field whose read threw — a
+  truncated or badly merged plugin, most often — so a field that could not be read looked exactly like a
+  field the record does not carry. It now renders as `(unreadable: <reason>)` on its own line, naming the
+  sub-field and Mutagen's own reason, in every form and format; the sub-fields beside it still read.
+
 - **Every tool description now fits inside the 2,048 characters Claude Code shows, and the grammar that was
   being cut off lives on the parameters it belongs to.** A description longer than that cap lost its tail on
   the way to the model, so predicate languages, source and comparison rules, refusal lists, projection forms
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -59,8 +59,10 @@ saying it sets an expectation their install may contradict. Say what is known, a
   truncated or badly merged plugin does to one field — so a field that could not be read looked exactly like
   a field the record does not carry. It now renders as `(unreadable: <reason>)` on its own line, naming the
   sub-field and Mutagen's own reason, and the fields beside it still read. `delta` and `tree` name it too:
-  an unreadable field is reported as not compared and leaves the comparison incomplete, so a record carrying
-  one is never called `identical`.
+  an unreadable field is reported as not compared, with the reason, and leaves the comparison incomplete, so
+  a record carrying one is never called `identical`. The rest of that record still compares — only the
+  unreadable path and the list it sits in drop out — and such a record is counted as neither differing nor
+  identical but under its own `no_verdict` count.
 
 - **Every tool description now fits inside the 2,048 characters Claude Code shows, and the grammar that was
   being cut off lives on the parameters it belongs to.** A description longer than that cap lost its tail on
--- a/src/housecarl-core/FieldsDiff.cs
+++ b/src/housecarl-core/FieldsDiff.cs
@@ -19,7 +19,8 @@ namespace HousecarlCore;
 ///     emitted: for record types where list order IS the semantics (a DIAL's INFO children decide which
 ///     line plays), a pure reorder must not read as identical. Nested list reordering INSIDE an element is
 ///     not canonicalised — it can over-report as a content delta, never under-report;
-///   • if either side's deep read hit the expansion cap, <see cref="Result.Complete"/> is false: list
+///   • if either side's deep read hit the expansion cap, OR either side carries a leaf it could not read,
+///     <see cref="Result.Complete"/> is false: list
 ///     comparison and one-sided-presence deltas are SUPPRESSED (where the two caps fell would otherwise
 ///     fabricate differences), only value mismatches observed on both sides are reported, and the caller
 ///     must not claim identity beyond what was actually compared.
@@ -27,8 +28,10 @@ namespace HousecarlCore;
 public static class FieldsDiff
 {
     /// <summary>Field-level deltas, preformatted for the conflict-tree render. <see cref="Complete"/> false ⇒
-    /// at least one side's read was truncated at the expansion cap, so an empty <see cref="Deltas"/> must NOT
-    /// be rendered as "identical to winner" — never claim knowledge the comparison doesn't have.
+    /// at least one side's read was truncated at the expansion cap, or carried a leaf that could not be read, so an
+    /// empty <see cref="Deltas"/> must NOT be rendered as "identical to winner" — never claim knowledge the
+    /// comparison doesn't have. An unreadable leaf also gets its own named line in <see cref="Deltas"/>, so that
+    /// case never presents as an empty delta list.
     ///
     /// <para><see cref="AgreedCount"/> + <see cref="AgreedSample"/> are the present-==-winner signal:
     /// how many VALUE LEAVES the node carries that exactly equal the winner's — i.e. ITM-restated fields,
@@ -61,8 +64,10 @@ public static class FieldsDiff
     {
         var tValueLeaves = new HashSet<string>(StringComparer.Ordinal);
         var wValueLeaves = new HashSet<string>(StringComparer.Ordinal);
-        var (tLines, tComplete) = CleanLines(theirs, tValueLeaves);
-        var (wLines, wComplete) = CleanLines(winner, wValueLeaves);
+        var tUnreadable = new HashSet<string>(StringComparer.Ordinal);
+        var wUnreadable = new HashSet<string>(StringComparer.Ordinal);
+        var (tLines, tComplete) = CleanLines(theirs, tValueLeaves, tUnreadable);
+        var (wLines, wComplete) = CleanLines(winner, wValueLeaves, wUnreadable);
         bool complete = tComplete && wComplete;
 
         // Numeric-bracket roots seen on EITHER side (the union, so a 0-item-vs-N-item list is still compared
@@ -90,6 +95,17 @@ public static class FieldsDiff
         }
 
         var deltas = new List<string>();
+
+        // ---- unreadable leaves: a NO-VERDICT, named. The path is out of the comparison above (its note is a
+        //      reason, not a value) and Complete is already false, so the record can never be counted identical;
+        //      the line says WHICH field and on which side, because "the comparison is incomplete" alone does not
+        //      tell a caller where to look.
+        foreach (var path in tUnreadable.Union(wUnreadable).OrderBy(p => p, StringComparer.Ordinal))
+            deltas.Add(tUnreadable.Contains(path) && wUnreadable.Contains(path)
+                ? $"{path}: UNREADABLE on both sides — not compared"
+                : tUnreadable.Contains(path)
+                    ? $"{path}: UNREADABLE here — not compared"
+                    : $"{path}: UNREADABLE in {referenceLabel} — not compared");
 
         // ---- exact-path comparison: scalars, substructs, dict children (bracket = a semantic key) --------
         // On a TRUNCATED comparison the list roots' own summary lines join the exact-path set: a root count
@@ -193,18 +209,25 @@ public static class FieldsDiff
         return null;
     }
 
-    /// <summary>The read's lines minus the expansion-cap sentinel; each value is the round-trippable token or,
-    /// for a non-leaf/absent line, its note. <paramref name="valueLeaves"/> collects the paths that carry a real
-    /// VALUE (<c>HasValue</c>) — the only lines an agreement count may consider, so a container summary line
-    /// ("[3 item(s)]") or an absent/null-link note is never miscounted as a present field. Complete=false iff the
-    /// expansion-cap sentinel was present.</summary>
-    static (List<(string path, string val)> lines, bool complete) CleanLines(RecordFields rf, HashSet<string> valueLeaves)
+    /// <summary>The read's lines minus the expansion-cap sentinel and the UNREADABLE ones; each value is the
+    /// round-trippable token or, for a non-leaf/absent line, its note. <paramref name="valueLeaves"/> collects the
+    /// paths that carry a real VALUE (<c>HasValue</c>) — the only lines an agreement count may consider, so a
+    /// container summary line ("[3 item(s)]") or an absent/null-link note is never miscounted as a present field.
+    /// <paramref name="unreadable"/> collects the paths this side could not read: a fault note is a REASON, not a
+    /// value, so comparing it as one makes two unreadable poles agree and an unreadable-vs-read pair a value
+    /// difference. It leaves the comparison incomplete at that path, exactly as the cap does — Complete is false
+    /// when either was present. Keyed on <see cref="ReadEngine.IsUnreadableNote"/>, not on <c>Readable</c> alone:
+    /// the other Readable=false answer is a no-such-field, which IS knowledge about the record and stays a
+    /// comparable shape difference.</summary>
+    static (List<(string path, string val)> lines, bool complete) CleanLines(RecordFields rf,
+        HashSet<string> valueLeaves, HashSet<string> unreadable)
     {
         var lines = new List<(string, string)>(rf.Fields.Count);
         bool complete = true;
         foreach (var f in rf.Fields)
         {
             if (f.Path == "…") { complete = false; continue; }         // ReadEngine's expansion-cap sentinel
+            if (!f.Readable && ReadEngine.IsUnreadableNote(f.Note)) { unreadable.Add(f.Path); complete = false; continue; }
             if (f.HasValue) valueLeaves.Add(f.Path);
             lines.Add((f.Path, f.HasValue ? f.Token ?? "" : f.Note ?? ""));
         }
--- a/src/housecarl-core/FieldsDiff.cs
+++ b/src/housecarl-core/FieldsDiff.cs
@@ -53,8 +53,12 @@ public static class FieldsDiff
     /// presence bit, so a leaf that EQUALS the winner is counted as agreement (it IS the same modeled value) but
     /// the render never claims the contributor "carries" it as a distinct subrecord — there is no bit to prove
     /// that. The ABSENT render fires only on the explicit sentinels, which only nullable fields produce.</para></summary>
+    /// <param name="NoVerdictCount">How many of <paramref name="Deltas"/> are UNREADABLE no-verdict lines rather
+    /// than value differences. A no-verdict is a THIRD state — the record can be neither counted identical nor
+    /// reported as differing in value — so a caller counting differences subtracts it here rather than re-deriving
+    /// it from the line text, which only this comparison actually knows.</param>
     public sealed record Result(IReadOnlyList<string> Deltas, bool Complete,
-        int AgreedCount, IReadOnlyList<string> AgreedSample);
+        int AgreedCount, IReadOnlyList<string> AgreedSample, int NoVerdictCount);
 
     /// <summary>True when a CleanLines value is a read-engine "no value here" note sentinel — the field is
     /// modeled but the contributor carries nothing (absent optional, or a present-but-null link). Treated as a
@@ -222,7 +226,7 @@ public static class FieldsDiff
         // artifact — suppress it: a partial read must not claim "N fields match the winner". An unreadable leaf
         // does not touch it: every path it counts was read on both sides, and Complete still refuses identity.
         if (capped) { agreedCount = 0; agreedSample.Clear(); }
-        return new Result(deltas, complete, agreedCount, agreedSample);
+        return new Result(deltas, complete, agreedCount, agreedSample, noVerdict.Count);
     }
 
     /// <summary>How many agreed-field paths to keep for the render — a small sample, not the full set (a deep
--- a/src/housecarl-core/FieldsDiff.cs
+++ b/src/housecarl-core/FieldsDiff.cs
@@ -216,9 +216,11 @@ public static class FieldsDiff
     /// <paramref name="unreadable"/> collects the paths this side could not read: a fault note is a REASON, not a
     /// value, so comparing it as one makes two unreadable poles agree and an unreadable-vs-read pair a value
     /// difference. It leaves the comparison incomplete at that path, exactly as the cap does — Complete is false
-    /// when either was present. Keyed on <see cref="ReadEngine.IsUnreadableNote"/>, not on <c>Readable</c> alone:
-    /// the other Readable=false answer is a no-such-field, which IS knowledge about the record and stays a
-    /// comparable shape difference.</summary>
+    /// when either was present. Keyed on <c>Readable</c>, which carries the fault by construction, with ONE
+    /// carve-out: a no-such-field (<see cref="ReadEngine.IsNoSuchFieldNote"/>) IS knowledge about the record and
+    /// stays a comparable shape difference. Testing the fault note's prose instead would cover only the getter
+    /// throw and let the walk's other fault notes — an FLOI it cannot read, a <c>*parent</c> hop it cannot
+    /// take — back in as comparable values.</summary>
     static (List<(string path, string val)> lines, bool complete) CleanLines(RecordFields rf,
         HashSet<string> valueLeaves, HashSet<string> unreadable)
     {
@@ -227,7 +229,7 @@ public static class FieldsDiff
         foreach (var f in rf.Fields)
         {
             if (f.Path == "…") { complete = false; continue; }         // ReadEngine's expansion-cap sentinel
-            if (!f.Readable && ReadEngine.IsUnreadableNote(f.Note)) { unreadable.Add(f.Path); complete = false; continue; }
+            if (!f.Readable && !ReadEngine.IsNoSuchFieldNote(f.Note)) { unreadable.Add(f.Path); complete = false; continue; }
             if (f.HasValue) valueLeaves.Add(f.Path);
             lines.Add((f.Path, f.HasValue ? f.Token ?? "" : f.Note ?? ""));
         }
--- a/src/housecarl-core/FieldsDiff.cs
+++ b/src/housecarl-core/FieldsDiff.cs
@@ -20,10 +20,14 @@ namespace HousecarlCore;
 ///     line plays), a pure reorder must not read as identical. Nested list reordering INSIDE an element is
 ///     not canonicalised — it can over-report as a content delta, never under-report;
 ///   • if either side's deep read hit the expansion cap, OR either side carries a leaf it could not read,
-///     <see cref="Result.Complete"/> is false: list
-///     comparison and one-sided-presence deltas are SUPPRESSED (where the two caps fell would otherwise
-///     fabricate differences), only value mismatches observed on both sides are reported, and the caller
-///     must not claim identity beyond what was actually compared.
+///     <see cref="Result.Complete"/> is false and the caller must not claim identity beyond what was actually
+///     compared. What gets suppressed differs by cause, because the two are not alike. A CAP makes the whole
+///     line set unreliable — where it fell decides which lines exist — so list comparison and
+///     one-sided-presence deltas are suppressed RECORD-WIDE. An UNREADABLE leaf removes exactly one known
+///     path, so suppression is scoped to that path and everything under it (and, where it sits inside a
+///     positional list, to that list's element comparison, whose fingerprints it would otherwise skew): the
+///     rest of the record still compares in full, and a differing list beside an unreadable scalar is still
+///     reported.
 /// </summary>
 public static class FieldsDiff
 {
@@ -39,8 +43,9 @@ public static class FieldsDiff
     /// identically (an ITM override) was indistinguishable from one that simply doesn't carry it. Counts only
     /// exact-path value leaves read on BOTH sides (never container summaries, never a side's <c>(absent)</c> /
     /// <c>(null link)</c> sentinel — an absent field is NOT an agreement). <see cref="AgreedSample"/> is up to a
-    /// few of those paths for the render. Both are 0/empty on a truncated comparison — the agreed set, like the
-    /// one-sided deltas, would be a where-the-cap-fell artifact.</para>
+    /// few of those paths for the render. Both are 0/empty on a CAPPED comparison — the agreed set, like the
+    /// one-sided deltas, would be a where-the-cap-fell artifact. An unreadable leaf leaves them standing: every
+    /// path they count was really read on both sides.</para>
     ///
     /// <para><b>Honest limit:</b> per-field presence is reliable only for NULLABLE fields, whose absence the
     /// read engine surfaces as a distinct <c>(absent)</c> / <c>(null link)</c> note. Non-nullable scalars grouped
@@ -64,11 +69,20 @@ public static class FieldsDiff
     {
         var tValueLeaves = new HashSet<string>(StringComparer.Ordinal);
         var wValueLeaves = new HashSet<string>(StringComparer.Ordinal);
-        var tUnreadable = new HashSet<string>(StringComparer.Ordinal);
-        var wUnreadable = new HashSet<string>(StringComparer.Ordinal);
-        var (tLines, tComplete) = CleanLines(theirs, tValueLeaves, tUnreadable);
-        var (wLines, wComplete) = CleanLines(winner, wValueLeaves, wUnreadable);
-        bool complete = tComplete && wComplete;
+        var tUnreadable = new Dictionary<string, string>(StringComparer.Ordinal);
+        var wUnreadable = new Dictionary<string, string>(StringComparer.Ordinal);
+        var (tLines, tCapped) = CleanLines(theirs, tValueLeaves, tUnreadable);
+        var (wLines, wCapped) = CleanLines(winner, wValueLeaves, wUnreadable);
+        bool capped = tCapped || wCapped;
+        var noVerdict = tUnreadable.Keys.Union(wUnreadable.Keys).OrderBy(p => p, StringComparer.Ordinal).ToList();
+        bool complete = !capped && noVerdict.Count == 0;
+
+        // A path is OUT of the comparison when it, or a path it hangs under, could not be read on either side:
+        // the faulting side carries no line there and none below it, so comparing would report the other side's
+        // lines as a one-sided difference that is not one. Scoped to those paths only — a cap is the wholesale
+        // case, and it is handled separately.
+        bool Suppressed(string p) => noVerdict.Any(s => p.StartsWith(s, StringComparison.Ordinal)
+                                                     && (p.Length == s.Length || p[s.Length] is '.' or '['));
 
         // Numeric-bracket roots seen on EITHER side (the union, so a 0-item-vs-N-item list is still compared
         // as elements), then split DICT-vs-LIST by the read engine's in-band container marker: a numeric-KEYED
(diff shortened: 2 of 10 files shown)


src/housecarl-core/ReadEngine.cs
```diff
@@ -281,6 +281,11 @@ record = parent;
     /// <c>StepIntoElement</c>) but is READ-ONLY — an absent optional substruct is surfaced, never
     /// materialised. Per-leaf fault isolation: any reflection/parse failure names itself and never
     /// throws out, so one Mutagen-unparseable field can't crash a record read.</summary>
+    /// <summary>The reason an unreadable note reports. A getter's own throw comes back from reflection wrapped in a
+    /// <see cref="TargetInvocationException"/> whose message ("Exception has been thrown by the target of an
+    /// invocation") names nothing a caller can act on, so the note carries the inner exception's message instead.</summary>
+    static string Reason(Exception ex) => (ex as TargetInvocationException)?.InnerException?.Message ?? ex.Message;
```
Avick3110, line 302: `Reason` is inserted between `ReadLeaf`'s doc comment and `ReadLeaf` itself, so the "Read one leaf path off a located record..." summary now documents `Reason`, and `ReadLeaf` is left with no doc comment at all. Concretely: hovering `ReadLeaf` (or following the several `<see cref="ReadLeaf"/>` links in this file, e.g. from `NavigateValue` and `NavigateTo`) now shows nothing, and `Reason` carries two `<summary>` tags of which the first describes a different method. Move `Reason` above the `ReadLeaf` summary block (or below `ReadLeaf`) so each doc stays on its own member.

src/housecarl-core/ReadEngine.cs
```diff
@@ -647,8 +673,19 @@ static void Expand(object? val, Type declaredType, object parent, string path, i
...
+                if (prop is null)
+                {
+                    Emit(sink, ref budget, Fault($"{path}.{fname}",
+                        $"{RecordNaming.StripOverlay(val.GetType().Name)} declares '{fname}' but the read walk cannot resolve it"));
+                    continue;
+                }
+                object? fv;
+                try { fv = prop.GetValue(val); }
+                catch (Exception ex) { Emit(sink, ref budget, Fault($"{path}.{fname}", ex)); continue; }
+                Expand(fv, prop.PropertyType, val, $"{path}.{fname}", childDepth, sink, ref budget);
```
Avick3110: The isolation only wraps the getter CALL, not the recursion, so the same "a dropped line reads as absent" failure survives one level up. `Expand` on the child is outside the try, and the child's first two acts can both throw: `EmitToken(val, declaredType, parent)` (line 571) and `de.Cast<object?>().Count()` (line 591), the latter being exactly the "enumerator that faults mid-list" the doc above names.

Concrete scenario, the same shape as the new test but one field over: a substruct whose fields are `Value`, `Effects` (a list) and `Weight`, in a plugin whose list count is corrupt. `Value` emits a line; `Expand` recurses into `Effects`, whose count enumeration throws; the exception unwinds through this `foreach` (no try) into `EmitWithDepth`'s catch. Result: no line for `Effects`, no line for `Weight` — `Weight` is dropped silently and reads as a field the record does not carry — plus a SECOND line on the PARENT's own path, contradicting the parent summary line already emitted at line 592 (`Present: true` vs `Present: false`).

That duplicate is then discarded by both fold consumers, so the fault can vanish entirely: `FieldFold.Columns` takes `rec.Fields.FirstOrDefault(f => f.Path == fold.Root)` and gets the earlier `Present: true` summary; `RowProjection.Fold`'s `seen[key].Add(f.Path)` drops the second line with the same path.

Wrapping the recursive `Expand` call in the same try/catch as the getter (emitting `Fault($"{path}.{fname}", ex)`) closes it for the sub-field walk, and the gendered/list/dict arms need the same treatment.

src/housecarl-core/ReadEngine.cs
```diff
@@ -670,8 +707,11 @@ static bool Emit(List<FieldValue> sink, ref int budget, FieldValue fv)
 
     /// <summary>Navigate a path READ-ONLY to its target, returning the live value object (+ declared type +
     /// owning parent) for recursion, or a miss note. Same walk as <see cref="ReadLeaf"/> but yields the object
-    /// instead of a token, so the expander can descend into it. Fault-isolated.</summary>
-    static (bool ok, object? val, Type type, object parent, string? note) NavigateValue(object record, string[] path)
+    /// instead of a token, so the expander can descend into it. Fault-isolated.
+    /// <para><c>readable</c> classifies the miss exactly as <see cref="ReadLeaf"/> does — false for a no-such-field
+    /// or a throw, true for a genuinely absent optional — carried structurally so a caller never decides "could not
```
Avick3110: This claims the classification is "carried structurally so a caller never decides 'could not look' versus 'nothing there' by matching the note's prose", but neither boundary out of this file carries it, so no caller can actually act on it:

1. `NavigateTo` (line 408), the only wrapper other callers use, still returns the 5-tuple without `readable`. Its one consumer, `FieldPredicate.ElementsAt` (FieldPredicate.cs:860), therefore still classifies with `if (note == ReadEngine.AbsentNote)` — prose matching, and the branch it picks decides whether a quantified step reads as an empty list or as a no-verdict.
2. `JsonWire.WriteLeaf` (JsonWire.cs:336-338) writes only `path` and `value`/`note`. Neither `Present` nor `Readable` is serialized, so an MCP client reading `format='json'` gets `{"path":"BasicStats.Damage","note":"(unreadable: ...)"}` and `{"path":"X","note":"(absent)"}` and can still only tell them apart by parsing the sentence.

Either propagate `readable` through `NavigateTo` and onto the JSON wire, or soften the comment to say the flag is internal to this file — as written it promises a guarantee the surface does not give.

src/housecarl-mcp-tests/RecordsUnreadableSubFieldTests.cs
```diff
@@ -0,0 +1,164 @@
...
+        return -1;
+    }
+
+    readonly string _priorCorpusPath;
+
+    public TruncatedSubFieldWorld()
+    {
+        // CorpusRulebook.CorpusPath is a process-global: capture before repointing, and restore before the
+        // directory the new value names is deleted.
+        _priorCorpusPath = CorpusRulebook.CorpusPath;
```
Avick3110, line 47: This world repoints the process-global `CorpusRulebook.CorpusPath` and deletes the directory the new value names on Dispose, which is exactly the criterion `RecordsWorldLifecycleTests` states an arm is required for: "Every world that repoints the static gets an arm here; the suite otherwise passes only on the order the collections happen to run in." `TruncatedSubFieldWorld` gets no arm.

Concrete failure it leaves uncovered: a later edit that reorders `Dispose` (restore after `Directory.Delete`, or drops the restore when `Svc.Dispose()` throws first) leaves `CorpusRulebook.CorpusPath` naming a deleted temp path, and every test class that runs after this one resolves types against a corpus that no longer exists — with no arm to say which world did it. Add `ADisposedTruncatedSubFieldWorldPutsTheGlobalCorpusPathBack` alongside the RecordsWorld/OwnedChildWorld arms.

src/housecarl-mcp-tests/RecordsUnreadableSubFieldTests.cs
```diff
@@ -0,0 +1,164 @@
...
+        // …and now cut two bytes off DATA, so the last field in the struct reads past the end of its own
+        // subrecord. The record, the group and the file all stay internally consistent: nothing but that one
+        // field is broken.
+        var bytes = File.ReadAllBytes(path);
+        int grup = Find(bytes, "GRUP", 0);           // the WEAP group header; its own label is "WEAP" at +8,
+        int rec = Find(bytes, "WEAP", grup + 24);    // so the record itself is searched for past the header
+        int sub = Find(bytes, "DATA", rec + 24);
+        int len = BitConverter.ToUInt16(bytes, sub + 4);
+        const int Keep = 8;
+        int cut = len - Keep;
```
Avick3110, line 99: The byte surgery trusts three unchecked results, and each failure mode corrupts the fixture into something that still writes a file, so the test fails somewhere far from the cause.

- `Find` returns -1 on a miss, and nothing checks it. `Find(bytes, "DATA", rec + 24)` returning -1 makes `BitConverter.ToUInt16(bytes, 3)` read two arbitrary bytes of the TES4 header as `len`, and the writes at `sub + 4` / the range `bytes[..(sub + 6 + Keep)]` then land at offsets 3 and 14 of the file header.
- `len` is never asserted to be 10. If a Mutagen bump changes WEAP's DATA layout (or an earlier subrecord's payload happens to contain the ASCII "DATA", which `Find` cannot distinguish from a header), `cut = len - Keep` can go negative; `(uint)cut` then underflows to ~4.29e9 and the record/group sizes are set to garbage, while `bytes[..(sub+6+8)].Concat(bytes[(sub+6+len)..])` DUPLICATES bytes instead of removing them.

One `Assert.True(grup >= 0 && rec >= 0 && sub >= 0)` plus `Assert.Equal(10, len)` before the mutation turns all of those into a one-line "the fixture assumption moved" failure.

src/housecarl-core/ReadEngine.cs
```diff
@@ -538,15 +545,26 @@ static void EmitWithDepth(object record, string path, int depth, List<FieldValue
...
             Expand(nav.val, nav.type, nav.parent, shown, depth, sink, ref budget);
         }
-        catch (Exception ex) { Emit(sink, ref budget, new FieldValue(shown, false, null, $"(unreadable: {ex.Message})", Present: false)); }
+        catch (Exception ex) { Emit(sink, ref budget, Fault(shown, ex)); }
     }
 
+    /// <summary>The ONE unreadable line the deep walk emits — same sentence and same structural flags the depth-1
+    /// read's <see cref="LeafRead.Unreadable"/> carries, so a nested fault and a top-level one read alike.</summary>
+    static FieldValue Fault(string path, string reason) =>
+        new(path, false, null, $"(unreadable: {reason})", Present: false, Readable: false);
```
Avick3110: **medium** — the comparison forms do not carry this line; on `delta`/`tree` an unreadable leaf can still be reported as `identical`.

`FieldsDiff` has no notion of `Readable`. `CleanLines` (FieldsDiff.cs:209) reduces every field to `f.HasValue ? f.Token : f.Note`, and `Complete` is set false only by the `"…"` expansion-cap sentinel (FieldsDiff.cs:207). So the new `(unreadable: …)` note enters the exact-path map as if it were a value token. I ran it:

```
both sides unreadable at the same path -> deltas=0 complete=True
unreadable vs a real value            -> deltas=1 complete=True
                                         "BasicStats.Damage=(unreadable: …) (winner 12)"
```

With zero deltas and `Complete` true, `RecordsTools.cs:947` counts the record in `identical=` and the renders print `identical to <ref> (whole record; N leaf/leaves agree)` (RecordsTools.cs:1775, 1866) — for a record one of whose fields could not be read on either pole. That contradicts `housecarl_records`' own description ("the content-keyed, truncation-honest engine: … a truncated deep read is reported, never claimed 'identical'") and the PR body's claim that `delta` and `tree` "carry it without a second spelling". The asymmetric case is also stated as a value difference — the engine asserts the subject's value differs from the reference's, when what it knows is that it could not read one of them.

Before this PR the same paths were simply missing, so a both-unreadable record also compared identical — but the PR is what makes the fault a line the comparison sees, and it is what claims the comparison forms now carry it.

Fix: give `CleanLines` the `Readable` bit and treat a `Readable=false` leaf as incomparable — clearing `Complete` reuses the honest wording the renders already have ("no differing fields in what was read, but the deep read was TRUNCATED … NOT a clean 'identical'"), or emit an explicit delta naming the unreadable side. Either way a test on `form='delta'` over the new truncated world would pin it.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A field inside a record that cannot be read now says so, instead of rendering as absent.** Reading a
+  record deeply (`project.depth` of 2 or more) dropped the line for any sub-field whose read threw — a
+  truncated or badly merged plugin, most often — so a field that could not be read looked exactly like a
```
Avick3110: **low** — "most often" is the frequency adjective this file's own head guidance rules out: "No frequency or coverage adjectives (\"rare\", \"most\", \"only affects…\"): how often something bites depends on the reader's install, which the entry cannot see" (CHANGELOG.md:10-12). Drop it — "a truncated or badly merged plugin" reads the same and claims nothing about the reader's install.

Same entry, next line: "in every form and format" is an unpointed coverage claim, and it is not currently true of `delta`/`tree` (see the comment on `ReadEngine.cs`). Point at the bound rather than restating a scope.

src/housecarl-core/ReadEngine.cs
Avick3110: Folded in bf28fc32: an unreadable leaf is now kept out of the comparison and clears Complete, and its path is named as its own `UNREADABLE … — not compared` delta, so neither pole can read as identical or as a value difference; a delta over the truncated world and a both-poles-unreadable case at FieldsDiff pin it, both failing before.

src/housecarl-core/ReadEngine.cs
Avick3110: Folded in 7e827a93: every recursion (substruct, list, dict and gendered arm) now goes through ExpandChild, which wraps Expand in the same try/catch, so a fault under a child names that child's path and the siblings keep reading. No test: the fixture's fault is a short subrecord, which throws when the child's own count or token emit runs and is already caught at that child — I could not make the fixture fault deeper without an injected hook, which the world deliberately avoids.

src/housecarl-core/ReadEngine.cs
Avick3110: Folded in 1a50f5b9: Reason and its summary moved above ReadLeaf's doc block, so each member carries its own.

src/housecarl-core/ReadEngine.cs
Avick3110: Folded the claim, in 1ab4dc94: the comment now says the bit is carried onto the emitted leaf's Readable (which the rows fold reads, and which the conflict diff now reads too) and that NavigateTo drops it, so its callers still classify from the note. I did not carry it through the three sites: the JSON wire gaining present/readable members is a user-visible surface change with its own changelog line, and FieldPredicate's ClassifyMiss is a four-way no-verdict taxonomy that a bool cannot replace — both belong to their own issue. What the diff needed to separate a fault from a no-such-field is now a shared ReadEngine.IsUnreadableNote over one constant prefix, not a private copy of the sentence.

src/housecarl-mcp-tests/RecordsUnreadableSubFieldTests.cs
Avick3110: Folded in 9db8ae4b: the three Find results are asserted non-negative and DATA's length asserted to be 10, before any arithmetic uses them.

plugin/CHANGELOG.md
Avick3110: Folded in c4f5e3dc: "most often" is gone, and the coverage claim now points at the comparison forms and what they do (named as not compared, comparison incomplete, never identical) instead of asserting every form and format.

src/housecarl-mcp-tests/RecordsUnreadableSubFieldTests.cs
Avick3110: Refuted. The criterion that comment states is not the pattern the file actually follows: nine test worlds repoint CorpusRulebook.CorpusPath (BulkRecordsWorld twice, EpochWorld, TypeArmWorld, RecordsOwnedChildTests, RuntimeFormIdTests, WinnerScanCoverageTests, WinnerSourceScanTests, WriteSilentAddAndMasterGrowTests, RecordsWorld) and exactly two have an arm here. A tenth arm would be one more per-world guard against a mistake rather than a fix for what allows it — if the suite wants this covered, the honest shape is one arm that builds and disposes every world, not another special case. TruncatedSubFieldWorld restores the static before the delete in the same order the two armed worlds do.

Avick3110:
Rebased onto the current default tip (`cd6b3139`), which now carries the twelve merged description PRs. No squash: the eight commits are unchanged apart from the one conflict resolution below, and the diffstat against the new base is identical to what it was before the rebase — 6 files, +367/-36.

One file conflicted:

- `plugin/CHANGELOG.md` — kept main's consolidated wave entry ("Every tool description now fits inside the 2,048 characters…") whole, and re-applied this PR's own entry above it, which is where the PR had put it. The per-tool `housecarl_records` description entry that this branch was sitting on top of was dropped, because main deliberately replaced it with the consolidated one; that entry is not this PR's and nothing of it is lost. The only difference from main's `CHANGELOG.md` is the seven added lines of this PR's entry.

Nothing else conflicted. Main did not touch `ReadEngine.cs`, `FieldsDiff.cs`, `RecordsTools.cs` or either test file in the wave, so those took the PR's side unchanged. The `RecordsTools.cs` hunk is the delta form's incomplete-comparison note, a rendered string rather than a tool description, so the description bound is not in play for it.

Verified from the worktree on the rebased branch: `dotnet build housecarl.sln -c Release` clean, 0 errors; `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` green at 1625 passed, 0 failed, including `PublishedDescriptionBoundTests` (31 passed) and the new unreadable-sub-field tests (16 passed). `ci-all` not run locally (#570); CI runs it.

GitHub now reports the PR MERGEABLE.

src/housecarl-core/FieldsDiff.cs
```diff
@@ -193,18 +209,25 @@ public static Result Compare(RecordFields theirs, RecordFields winner, string re
...
+    /// comparable shape difference.</summary>
+    static (List<(string path, string val)> lines, bool complete) CleanLines(RecordFields rf,
+        HashSet<string> valueLeaves, HashSet<string> unreadable)
     {
         var lines = new List<(string, string)>(rf.Fields.Count);
         bool complete = true;
         foreach (var f in rf.Fields)
         {
             if (f.Path == "…") { complete = false; continue; }         // ReadEngine's expansion-cap sentinel
+            if (!f.Readable && ReadEngine.IsUnreadableNote(f.Note)) { unreadable.Add(f.Path); complete = false; continue; }
```
Avick3110: **The unreadable gate keys on the note's prose, so it misses the read walk's other fault notes.**

`CleanLines` requires BOTH `!f.Readable` AND `IsUnreadableNote(f.Note)`. But `(unreadable: ` is not the only sentence the read walk puts on a `Readable=false` leaf:

- `EmitFloi` returns `LeafRead.Unreadable` three times with a `(floi: ...)` note (`ReadEngine.cs:1167`, `:1176`, `:1181`) — a parent with no `UseAliases`/`UsePackageData` discriminator, a form-mode FLOI whose FormKey will not read, an index-mode FLOI whose index accessor is unresolved. All three are read faults by construction.
- `ReadFields` emits a `*parent` hop failure with `Readable: false` and a `(no *parent ...)`-class note (`ReadEngine.cs:216`, `:242`).

None of those are excluded, so they are still compared as tokens. Two poles that both fault on `Conditions[0].Data.Reference` compare their identical `(floi: form mode, null or unreadable FormKey on ...)` note as equal values: `Complete` stays true, `Deltas` is empty, and `tree` prints `identical to <ref> (whole record)` for a field neither side could read. The asymmetric pair emits `Conditions[0].Data.Reference=(floi: ...) (winner 0012A4:Skyrim.esm)` — a value difference asserted from a value that was never read. That is the same silently-wrong answer this PR closes for `(unreadable: ...)`.

The classification the diff actually wants is "fault vs no-such-field", and `Readable` already carries the first half. Keying on `!f.Readable` with an explicit carve-out for the no-such-field note (the one comparable `Readable=false` answer, and the one `ANoSuchFieldStillCompares` pins) covers every fault class by construction; a prefix test covers only the one that happens to use that prefix today.

*(Note this is currently masked at depth by the sibling issue on `ReadEngine.cs:600` — a deep FLOI fault arrives here claiming `Readable=true` as well. Both need fixing.)*

src/housecarl-core/ReadEngine.cs
```diff
@@ -538,15 +558,37 @@ static void EmitWithDepth(object record, string path, int depth, List<FieldValue
...
     /// a link → its note (not opened); a container/substruct → an identity-enriched summary line, then (while
-    /// depth allows) one child line per element (bracketed) or sub-field (dotted), each recursed at depth-1.</summary>
+    /// depth allows) one child line per element (bracketed) or sub-field (dotted), each recursed at depth-1.
+    /// <para>A child that cannot be read — its getter, or anything under it (see <see cref="ExpandChild"/>) — gets
+    /// its own "(unreadable …)" line, the same sentence and the same Readable=false the depth-1 read emits, and the
+    /// walk carries on with its siblings. It is never skipped: a dropped line is indistinguishable from an absent
+    /// optional, which would make an unreadable field read as evidence that nothing is there.</para></summary>
     static void Expand(object? val, Type declaredType, object parent, string path, int depth, List<FieldValue> sink, ref int budget)
     {
         if (budget < 0) return;
```
Avick3110, line 601: *(Anchored here; the line is 600, just past the hunk.)*

**`Expand`'s link/FLOI branch drops `leaf.Readable`.**

```csharp
if (val is IFormLinkGetter || WriteEngine.IsFormLinkOrIndex(Nullable.GetUnderlyingType(declaredType) ?? declaredType))
{ Emit(sink, ref budget, new FieldValue(path, false, null, leaf.Note, Present: leaf.Present)); return; }
```

`Present` is carried, `Readable` is not, so it defaults to `true`. For a plain link that is right. For a FLOI it is not: `EmitToken` dispatches a `FormLinkOrIndex` to `EmitFloi` at `:949`, and `EmitFloi` returns `LeafRead.Unreadable(...)` — `Present=false, Readable=false` — for a mode, index accessor or link it cannot read. The value is non-null and `IsFormLinkOrIndex(declaredType)` is true, so it lands on this branch and the false bit is lost.

The result is a depth-2+ leaf that says "nothing is there" when the walk could not look — precisely the structural confusion this bit exists to prevent, and precisely what this PR fixes one line up in `EmitWithDepth` (`Readable: nav.readable`). Every consumer reading the bit structurally instead of matching prose gets the wrong answer: `FieldsDiff` runs on deep reads (`ConflictDiffDepth`), and the `rows` fold's `!f.Readable` keep-rule never fires.

`Readable: leaf.Readable` alongside `Present: leaf.Present` is the whole fix, and it keeps the two flags travelling together the way the depth-1 read already does at `:226`.

src/housecarl-core/FieldsDiff.cs
```diff
@@ -61,8 +64,10 @@ public static Result Compare(RecordFields theirs, RecordFields winner, string re
     {
         var tValueLeaves = new HashSet<string>(StringComparer.Ordinal);
         var wValueLeaves = new HashSet<string>(StringComparer.Ordinal);
-        var (tLines, tComplete) = CleanLines(theirs, tValueLeaves);
-        var (wLines, wComplete) = CleanLines(winner, wValueLeaves);
+        var tUnreadable = new HashSet<string>(StringComparer.Ordinal);
+        var wUnreadable = new HashSet<string>(StringComparer.Ordinal);
+        var (tLines, tComplete) = CleanLines(theirs, tValueLeaves, tUnreadable);
+        var (wLines, wComplete) = CleanLines(winner, wValueLeaves, wUnreadable);
         bool complete = tComplete && wComplete;
```
Avick3110: **One unreadable leaf now suppresses list comparison for the whole record.**

`CleanLines` sets `complete = false` on an unreadable leaf, and `complete` gates the positional-list loop, the one-sided-presence deltas on both sides, and the agreed count. So the suppression is record-wide, not path-wide.

Concretely: a truncated plugin overrides a WEAP whose `BasicStats.Damage` cannot be read AND whose `Keywords` list differs from the winner's by two entries. `delta` now prints `BasicStats.Damage: UNREADABLE here — not compared` and the incomplete note, and says nothing at all about the keywords. Before this PR the keyword delta was reported (next to a bogus value delta on Damage). The user ran the diff for the keywords, and the answer got quieter.

The cap earns its blanket suppression because a cap makes the whole line set unreliable — where it fell decides which lines exist. An unreadable leaf is not like that: it removes exactly one path. The fabrication risk it creates is bounded — a one-sided-presence phantom at that path, and a wrong fingerprint for the element containing it — so scoping the suppression to that path (and, when the path sits under a numeric bracket, to its list root) keeps the rest of the comparison honest and complete. `Complete` can still go false so no caller claims identity.

If the record-wide sweep is the deliberate call, the class doc at the top says the cap is the cause of suppression; it would be worth saying there that an unreadable leaf suppresses everything too, and why that is the right trade against under-reporting.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1790,7 +1790,7 @@ static string RenderRecordsDelta(IReadOnlyList<LoadOrderService.DeltaRow> rows,
                     sb.Append("    - ").Append(delta).Append('\n');
```
Avick3110, line 1851: *(Anchored on the changed line; the code is at `:946` and `:1779`.)*

**The `differing` count now counts a no-verdict as a difference.**

```csharp
int differing = rows.Count(x => x.Error is null && x.Diff!.Deltas.Count > 0);
int identical = rows.Count(x => x.Error is null && x.Diff!.Deltas.Count == 0 && x.Diff.Complete);
```

`identical` correctly excludes an unreadable record (`Complete` is false). `differing` includes it, because the UNREADABLE line lives in `Deltas`. So `records project.form='delta' counts_only=true` over a modlist with one truncated plugin reports `differing=1 identical=0` for an override that is byte-identical to its reference apart from a field nobody could read. The PR body's own contract is that such a record "can be neither counted `identical` nor reported as differing in value" — the count says otherwise.

The same conflation reaches the text render at `:1779`, which heads the block "N differences — each line: `<subject>`'s value (reference = `<ref>`)" and then prints a line that is explicitly not a value, and `delta_count` in the JSON wire.

A no-verdict is a third state, so it wants a third number: count only deltas that are not UNREADABLE lines as `differing`, and either carry the no-verdict rows as their own count or leave them out of both. Since `FieldsDiff.Result` is the one place that knows which deltas are no-verdicts, exposing that count on `Result` beats having the render re-derive it from the line text.

Avick3110, line 1851: *(Anchored on the changed line; the code is at `:1859`–`:1866`.)*

**The `tree` form never prints the incompleteness note when the node has deltas.**

```csharp
if (n.Deltas.Count > 0)
    sb.Append(string.Join("; ", n.Deltas)).Append('
');
else if (!n.Complete)
    sb.Append("no differing fields in what was read, but the deep read was TRUNCATED — not a clean 'identical'.
");
```

The `!n.Complete` branch is reachable only when there are no deltas. That was already a small gap for a cap-truncated node with value mismatches; this PR makes it the *normal* shape, because an unreadable leaf always produces a delta line and always sets `Complete` false.

So a contributor with an unreadable `BasicStats.Damage` and a differing `Keywords` list renders as:

```
    PluginA.esp: BasicStats.Damage: UNREADABLE here — not compared
```

and nothing else. The keyword list was never compared (the `complete` guards in `FieldsDiff` skipped the list loop), and the node line does not say so. The `delta` form's equivalent path was reworded in this PR to say exactly that (`:1792`); the PR body lists `tree` as carrying the same fact, and it does not.

Appending the note whenever `!n.Complete`, independent of the delta count, matches `delta` and keeps the two renders on the same wording rule the method's own doc comment claims.

src/housecarl-core/FieldsDiff.cs
Avick3110: Folded in 3ba3bbaf. The gate is `!f.Readable && !ReadEngine.IsNoSuchFieldNote(f.Note)` now, with `NoFieldPrefix`/`IsNoSuchFieldNote` as the one carve-out and `IsUnreadableNote` gone (it had no other consumer). So every fault class is covered by construction — the three `(floi: …)` returns and the `*parent` hop notes included — and only the comparable answer is left in. `AnFloiFaultIsNotComparedEither` pins the FLOI class; `ANoSuchFieldStillCompares` is unchanged as the control.

Two knock-ons worth flagging. `RecordsContainmentTests.TheComparisonFormsReadTheHopOnTheirInOrderArm` measured the hop on a delta where the off-order pole cannot take it: that used to render as a value difference showing both sides’ text, and is now a no-verdict, so the assertion is on `*parent.EditorID: UNREADABLE in <ref>` plus not-`on both sides` — which pins the same two halves more tightly than the old substring did. And because that sentence carries its own remedy ("read the record through the load order instead"), the no-verdict line now carries the read’s reason after "not compared" rather than dropping it (a15880de); when both sides fault with different reasons it names both.

src/housecarl-core/ReadEngine.cs
Avick3110: Folded in e3db07c5 — `Readable: leaf.Readable` beside `Present: leaf.Present`, one line, exactly as suggested.

It is not only reachable, it fires on a plain in-memory record today: a `GetInFaction` condition with `UseAliases` set reads `Conditions[0].Data.Faction` as `(floi: index mode, index accessor unresolved — refined in oracle)` with `Present=false, Readable=true` before the fix. That is the new test, `UnreadableLinkLeafReadTests.AnUnreadableConditionTargetIsNotReadAsAbsent` — no fixture plugin needed, and it fails on the old line.

src/housecarl-core/FieldsDiff.cs
Avick3110: Folded in a15880de, your way: suppression is now scoped, the cap keeps its record-wide sweep, and `Complete` stays false either way. This is the judgement call of the five, so the reasoning, and the escape hatch if you want it the other way.

Why scoped. The cap and an unreadable leaf are not the same kind of ignorance. Where the cap fell decides which lines exist, so nothing about the line set can be trusted; an unreadable leaf removes exactly one KNOWN path. `CleanLines` now returns `capped` alone, the unreadable paths are collected separately, and a path is out of the comparison when it, or a path it hangs under, could not be read on either side (`Suppressed`) — that covers the one real fabrication risk on the exact-path side, including the case you did not raise: a fault ON a substruct emits one line and no descendants, so the other side’s children would otherwise all report as one-sided. Element comparison drops for a root holding an unreadable leaf, since a missing line changes that element’s fingerprint; other roots compare as usual, and a suppressed root still compares its own count token. `AgreedCount` survives an unreadable leaf (every path it counts was read on both sides) and still dies on a cap.

Two tests: `ADifferingListBesideAnUnreadableLeafIsStillCompared` (your WEAP example — the keywords delta is reported again, and the winner’s readable Damage is not reported as one-sided presence) and `AnUnreadableLeafInsideAnElementSuppressesOnlyThatList`.

The class doc now states both rules and why they differ, and the delta form’s note says what actually held: "a field above could not be read (nothing at or under it was compared), or the deep read hit the cap (which suppresses list-content and one-sided-presence deltas for the whole record)".

If you would rather keep record-wide suppression, it is `if (capped)` → `if (!complete)` in three places plus the doc; say the word and I will put it back.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 0f4b61db. `FieldsDiff.Result` carries `NoVerdictCount` — the comparison is the one place that knows which lines are no-verdicts, so nothing re-derives it from the line text.

- `differing` counts a row only when `Deltas.Count > NoVerdictCount`; `identical` is unchanged (`Complete` already excluded it).
- The census gains `no_verdict` (rows with at least one), text and json: `count=1 differing=0 identical=0 no_verdict=1 errors=0` for your truncated-plugin case.
- The text block header heads the VALUE differences and names the rest apart: "1 field that could not be read — each value line: …". With both: "2 differences and 1 field that could not be read — each value line: …".
- The record-count line names the no-verdict rows only when there are any, so the two counts above it do not look like they lost a record.
- The json row gains `no_verdict_count` beside `delta_count` (and the delta artifact manifest lists it).

`ANoVerdictIsCountedApartFromValueDifferences` pins both halves. One test moved with the wording: `FactB7` asserts "each value line".

Avick3110: Folded in b00f3a3f. The note is appended whenever `!n.Complete`, deltas or not, and it names both causes the way the delta form does. The empty-delta branch keeps its own sentence, since the cap is the only cause that can reach it now.

Pinned by `TreeIncompleteNoteTests`, driven against a hand-built `TreeRow` — no fixture makes this node shape (the truncated plugin is off-order, so its record has one in-order toucher and the tree has nothing to diff), so `RenderRecordsTree` is internal for the same reason `AppendChildDeclarers` is. Two arms: a node with deltas still says the comparison was incomplete, and the no-delta node still names the cap.

PR #601, opened 2026-09-06
Bring the forward tool description under the 2,048-character cap

### LINKED ISSUE #583: [Bug] Claude Code cuts every MCP tool description at 2,048 characters; twelve 2.0 tools lose most of their grammar

Avick3110:
﻿`housecarl_forward` published a 4,132-character tool description. Claude Code shows the first 2,048 characters of a description and delivers parameter descriptions whole, so the set-valued rule, the source resolution and its refusals, the lane semantics and the whole ALL-OR-NOTHING refusal list were being described in a part of the text the model never saw. The grammar moves rather than going away: each paragraph lands in the description of the parameter it is about, where the parameter already said the same thing the description's copy is dropped as a duplicate rather than pasted in twice, and where a description's sentence was stated wider than the code enforces it is corrected to the code (the two cases are listed below). The description keeps the one-sentence purpose, what the tool is for, the composition sentence, a line per axis naming its parameters, the call-level all-or-nothing rule, the stale-winner recipe and the pointers to the sibling tools. This is the last of the twelve: the bound test's exemption set is deleted outright, so every published description is held with nothing exempt, and the wave's CHANGELOG entry is rewritten to cover the whole wave rather than `housecarl_records` alone.

Closes #583.

| | Before | After |
|---|---|---|
| `housecarl_forward` description | 4,132 | **1,968** |
| Parameter descriptions | 2,656 | 6,168 |

The description shed 2,164 characters and the schema gained 3,512. It gained more than the description shed because the refusal list, which was one run-on sentence sharing a single subject ("if source is ..."), becomes a clause per parameter, each of which has to name its own subject; because the description's version of a rule won wherever the parameter's version was weaker; and because the review fold put the two lanes' own refusals on `into=` and `in_place=`, which the old description never listed anywhere. `housecarl_forward` has no nested DTO parameters, so the parameter total is the whole schema.

## Harvest

|---|---|
| "Forward a SPECIFIC plugin's version of one-or-more records as an override — xEdit's 'copy as override into', the INVERSE of `housecarl_apply`." | kept in description |
| "apply edits the load-order WINNER; this copies source's whole record VERBATIM, so SOURCE's version (not the winner) becomes the patch's content." | kept in description |
| "ONE surface: what to copy (formids=) x WHOSE version (source=) x WHERE it lands (the LANE) x how it reads back (TRANSPORT)." | kept in description |
| "WHAT IT IS FOR: RE-ASSERT an earlier mod's version over a later override (a late total-overhaul re-architected a record an earlier list patch had balanced — …)." | kept in description, with its worked example |
| "or REVERT a record to vanilla (name a master — Skyrim.esm/Update.esm/… — as source)." | kept in description |
| "This copies the record WHOLE: it does NOT edit fields (that's apply) and needs no field pre-flight, because a complete source record is legal by construction." | kept in description |
| "formids= is set-valued — one or more 'XXXXXX:Plugin.esp', ALL copied from the SAME source" | `formids` (already said there; "set-valued, so many records copy in ONE write" replaced the bare "Set-valued") |
| "call again with into= to forward from a different source into the same patch" | `formids`, and `into` already said it from its own end |
| "It also accepts ['@<absolute path>']." | `formids` (already said there, with the one-FormID-per-line rule) |
| "source= is any plugin that DEFINES or overrides each record" | `source` (already said there; "a record it doesn't touch is refused by name, never silently absent" is folded in from the refusal list below) |
| "ACTIVE, or a file that is only ON DISK (a DISABLED mod, an unticked plugin, a folder MO2 never registered; pass its full path if several folders provide that filename)." | `source` (the unregistered-folder arm and the full-path remedy were not there before) |
| "Re-asserting a disabled old patch's version is a first-class use of this tool, not an edge case." | `source` |
| "Forwarding does NOT add source as a master: the patch overrides the record's ORIGIN FormKey with the copied body, so the header carries the origin master + whatever the body references (exactly xEdit's copy-as-override-into-a-new-patch)." | `source` |
| "The record's ORIGIN plugin must still be active — a patch can't master a plugin that isn't loaded" | `formids`, where the origin is read off the FormID the caller types; `source` points at it for the master model. Narrowed to match `ResolveForwardSources`, which exempts an origin that IS the file being written |
| "and an off-order read STATES which copy on disk it opened." | `source` (already said there) |
| "LANE — where the write lands. Default: a NEW patch named patch= (auto-suffixed if taken)." | the description's LANE axis line (which now says the new-patch lane is the default) and `patch` (already said there). `patch=`'s own copy of the lane-exclusivity rule is dropped in the fold, because the description states it and neither `into=` nor `in_place=` ever did |
| "into='…' EXTENDS that patch — and if it ALREADY carries a forwarded FormKey, its existing override is REPLACED by source's body (xEdit's copy-as-override overwrite, flagged per record)." | `into` (the xEdit naming and "flagged per record" added to what it already said) |
| "in_place='<plugin filename>' forwards INTO an existing plugin's OWN file (incl. one houseCARL didn't author) — same replace-on-collision semantics, your ORIGINAL rewritten, no houseCARL backup or undo, and the same acknowledge= consent as the sibling write tools." | `in_place` (the replace-on-collision cross-reference and the acknowledge= pointer added) |
| "THE STALE-WINNER BYPASS RECIPE (pinned): forward from the source you want, then apply into= the same patch — …" | kept in description |
| "ALL-OR-NOTHING (Q3): the whole call is refused with a named reason and NOTHING is written…" (the rule) | kept in description, and restated on `formids` for the targets it owns |
| …"if source is found in NEITHER the load order NOR on disk (both places are named)" | `source`, with the did-you-mean the locate actually appends |
| …"matches several mod folders" | `source`, with the pass-the-path remedy |
| …"was excluded as unparseable WHEN NAMED (addressing that same file by PATH reads it directly instead — copying one record out is not the whole-file re-serialize the refusal guards; the response says so)" | `source` |
| …"is the artifact being written itself" | `source` ("the file this call is writing itself" — one sentence for both lanes, which is what the engine's two messages say) |
| …"names a target twice" | `formids` |
| …"simply doesn't DEFINE/override a given record (nothing there to forward)" | `source` |
| …"or names a record whose ORIGIN plugin isn't active (the patch would need it as a master)" | `formids` |
| "dry_run=true resolves every record from source, copies each into the in-memory would-be artifact, and STOPS before anything touches disk — what WOULD be forwarded, or EXACTLY the refusal the real call would give." | `dry_run` (which said the rest already; the resolve-and-copy detail is the description's, and it wins) |
| "Returns, per record, what was copied and the current winner it will out-rank once enabled (a forward whose version is ALREADY winning is flagged redundant, never silently a no-op)." | `format`, which is where the response shape is described on the sibling write tools |
| "readback=true additionally returns each forwarded record IN FULL off the written file — the pre-enable verification that the copy is exactly the source's, WITHOUT enabling the patch in MO2 (the written file's content, not load-order truth)." | `readback` (already said there, minus "WITHOUT enabling the patch in MO2", which is folded in) |
| "max_chars caps the render with an explicit notice." | `max_chars` (already said there, in more detail) |
| "Every response carries epoch=<hex> — the identity of the index build the sources and the out-ranked winners were resolved from." | `format`, corrected the way `housecarl_remove`'s was: every response answered from a build carries it, spelled `epoch=<hex>` on `'text'` and as an `epoch` member on `'json'`, and a refusal that consulted no build carries none |

Two things the description gained. The axis roster is the description's only index of the parameters now, so it names `dry_run=` and `acknowledge=`, which the old prose mentioned but no axis line pointed at — `dry_run=` on the LANE line, which is where `housecarl_apply` files it. And it ends with a pointer to `housecarl_records` — `project.form='tree'` for every plugin touching the record, `'delta'` against a `versus=` reference — because "see what the forward will out-rank before writing it" is the read that belongs in front of this call and the description named no read tool at all.

Two things were narrowed rather than moved, both because the old sentence was wider than the code:

- The refusal list read as exhaustive ("the whole call is refused … if source is A, B, C, …") but omitted cases the engine refuses: a malformed FormID, a result-artifact path in `formids=`, a source file that cannot be opened or read, an in-place target that is not active, a localized in-place target, an unwritable parent, and `into=` naming a patch that does not exist. The description now states the all-or-nothing RULE and says each parameter carries its own refusals; every case is on the parameter that owns it, including the two lanes' own — `into=` names the missing patch, the foreign folder and the ambiguous owned match; `in_place=` names the inactive target, the excluded one, the localized one and the unwritable parent.
- "The record's ORIGIN plugin must still be active" is unconditional, but `ResolveForwardSources` exempts an origin that is the artifact being written (re-asserting an old copy's body of a record the patch you are authoring originates, before that patch is enabled). `formids` now says "must be active, or be the very file this call writes".

## Parameter spellings

`ForwardTools` passes `sourceParam: "source="`, so every engine refusal that names the source pole renders the tool's own word. Three refusals in `LoadOrderService.ForwardRecords` name `target=` and `in_place=true`, neither of which `housecarl_forward` exposes — but all three arms are unreachable from this tool (`hasInPlace` is derived from `in_place` being non-empty, and the lane conflicts are refused in `ForwardTools` first), so nothing was changed for them here. Noting it because #597 is in flight on the same path.

## One engine fix

`format=` says every response answered from a build carries the epoch stamp, and a refusal that consulted no build carries none. One refusal on the forward path broke that: `ResolveOutputPath` throwing — `into=` naming a plugin houseCARL does not own, or one it holds no .esp for — was caught as `ForwardOutcome.Fail(ex.Message)` with no `Stamp`, even though the source resolve had already captured a view. It now returns `with { Stamp = offEpoch }`, like every other post-capture outcome on that path, so the sentence holds rather than needing a qualifier.

## The bound test

`src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs`: `ToolNames.Forward` was the last line of `StillOversized`, and with the wave landed the set, its early return in the theory, and the `EveryToolNamedOversizedStillIs` Fact that iterated it are all deleted. `EveryPublishedDescriptionFitsTheClientsCut` — the theory over every tool the published-name capture holds — is now the whole test, and there is no longer a way for a published description to exceed the bound without going red. The failure message the theory shares points only at the capture, which is where its rows come from.

Run before the description change, with the line already deleted: `housecarl_forward` failed at 4,132, the only failure in the class. Run after: 31 of 31 green.

## Every published description, longest first

| Tool | Characters |
|---|---|
| `housecarl_nif_inspect` | 2,037 |
| `housecarl_records` | 2,036 |
| `housecarl_apply` | 2,032 |
| `housecarl_write_seq` | 1,997 |
| `housecarl_skse` | 1,993 |
| `housecarl_forward` | 1,968 |
| `housecarl_nif_set` | 1,830 |
| `housecarl_compile_script` | 1,783 |
| `housecarl_nexus_check_updates` | 1,743 |
| `housecarl_merge_plugins` | 1,742 |
| `housecarl_asset_status` | 1,659 |
| `housecarl_check` | 1,649 |
| `housecarl_nexus_mod` | 1,633 |
| `housecarl_skypatcher_layer` | 1,573 |
| `housecarl_decompile_script` | 1,334 |
| `housecarl_load_order_status` | 1,331 |
| `housecarl_create_plugin` | 1,248 |
| `housecarl_create` | 1,225 |
| `housecarl_remove` | 1,223 |
| `housecarl_place` | 1,146 |
| `housecarl_copy` | 1,104 |
| `housecarl_nexus_graphql` | 993 |
| `housecarl_compact_plugin` | 960 |
| `housecarl_update_status` | 951 |
| `housecarl_nexus_search` | 914 |
| `housecarl_set_tool_path` | 910 |
| `housecarl_set_mo2_instance` | 867 |
| `housecarl_nexus_identify` | 692 |
| `housecarl_bsa_repack` | 524 |
| `housecarl_bsa_extract` | 438 |
| `housecarl_bsa_list` | 272 |

Measured on the `tools/list` the built server serves over stdio, not on the source literals. The bound is 2,048; the tightest margin on the surface is `housecarl_nif_inspect`'s 11 characters.

## Review fold

Seven findings from the blind review, all folded, one commit each:

| Finding | Commit |
|---|---|
| "each parameter above carries its own refusals" was untrue of `into=` and `in_place=` | `b1589c87` — each lane now names its own refusals, in the engine's words |
| The lane-exclusivity rule was stated twice, in the description and on `patch=` | `e2ce05a0` — dropped from `patch=`; `into=` and `in_place=` never state it, so the description's copy is the one that reaches those callers |
| `dry_run=` filed under TRANSPORT, while `housecarl_apply` lists it on LANE | `b90a4936` — moved to the LANE line |
| An `into=` refusal out of `ResolveOwnedPatchFolder` returned unstamped though a view had been captured | `cb9d9352` — the catch now stamps, so `format=`'s epoch sentence needs no qualifier |
| The emptied `StillOversized` set and its early return were the one documented way past the bound, and the Fact iterated nothing | `f2975f75` — set, early return and Fact deleted |
| The theory's failure message pointed at `StillOversized` for rows that come from the name capture | `f2975f75` — points only at the capture |
| The CHANGELOG's "none of the text was deleted" was untrue as shipped, since two rules were narrowed | `68fa4721` — says the grammar moved, and that a rule stated wider than the code was corrected to the code |

## What was run

```
    1619 passed, 0 failed   (one fewer than before: the deleted Fact)
dotnet src/housecarl-generator/bin/Release/net9.0/housecarl-generator.dll description-vocab-guard
    64 passed, 0 failed
```

`ci-all` was not run locally (#570 — its probes collide with parallel runs); CI runs it here. `description-vocab-guard` was run alone because it is the one probe that reads `[Description]` prose; no generator probe reads `housecarl_forward`'s description text, and no test or comment in `src/` quotes it. `src/housecarl-mcp-tests/data/tools-list-2.0.json` holds tool names only, so nothing needed regenerating.


src/housecarl-mcp/ForwardTools.cs
```diff
@@ -21,59 +21,38 @@ public static class ForwardTools
...
+        [Description("TRANSPORT: also return each forwarded record IN FULL, read back from the written file on disk (every field, deep) — the pre-enable verify that the copy is exactly the source's, WITHOUT enabling the patch in MO2. The written file's content, not load-order truth.")]
             bool readback = false,
-        [Description("TRANSPORT: 'text' (default) | 'json' (the same data, machine-readable, accounting in-band).")]
+        [Description("TRANSPORT: 'text' (default) | 'json' (the same data, machine-readable, accounting in-band). Either way the response states, per record, what was copied and the current winner it will out-rank once enabled — a forward whose version is ALREADY winning is flagged redundant, never silently a no-op. Every response answered from a build carries the epoch stamp — the identity of the index build the sources and the out-ranked winners were resolved from — spelled epoch=<hex> on 'text', and as an 'epoch' member on 'json'; a refusal that consulted no build carries none.")]
```
Avick3110, line 56: The new epoch sentence is slightly wider than the code. `into=` naming a plugin houseCARL does not own (or a folder it cannot resolve an .esp in) is refused by `ResolveOwnedPatchFolder` throwing out of `ResolveOutputPath`, and `LoadOrderService.ForwardRecords` catches it as `ForwardOutcome.Fail(ex.Message)` with **no** `Stamp` — even though `Resolver` was built and `ResolveOffOrderForwardSource` already did `resolver.Capture()` on the way in. So `housecarl_forward(formids=[...], source='X.esp', into='NotAHouseCarlPatch.esp')` returns a refusal with no `epoch=` after a build was consulted, which is the case this sentence says carries one. Every other post-capture refusal on this path is stamped (`offError`, the engine's own refusals via `ForwardRecordsCore`), so this is the one hole. Either narrow the sentence to the refusals the engine answers, or stamp the `ResolveOutputPath` catch the way the `offError` return is stamped.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
```diff
@@ -27,11 +28,9 @@ public sealed class PublishedDescriptionBoundTests
...
     const int Bound = 2048;
 
-    // Tools whose descriptions are still over the bound, one line each. Each description PR deletes its own line.
-    static readonly HashSet<string> StillOversized = new(StringComparer.Ordinal)
-    {
-        ToolNames.Forward,
-    };
+    // Tools whose descriptions are still over the bound, one line each. Empty on purpose: the wave that brought
+    // the twelve oversized descriptions under the bound has landed, so nothing is exempt from the theory.
+    static readonly HashSet<string> StillOversized = new(StringComparer.Ordinal);
```
Avick3110: Emptying the set leaves the `Description(string tool)` failure message (line 41) stale, and it is now the only way that message can ever fire. Its rows come from `PublishedNameAnchorTests.Captured()`, not from this file, so when a tool is renamed in `ToolNames` and the capture is not regenerated, `EveryPublishedDescriptionFitsTheClientsCut` fails with "'<tool>' is named in this test file but the server does not publish it ... Update its line in StillOversized and in '<capture>'" — but the tool is not named in this file, and `StillOversized` has no lines to update. The reader is sent to a set that is empty on purpose. Worth dropping the `StillOversized` half of that sentence so the message points only at the capture.

src/housecarl-mcp/ForwardTools.cs
```diff
@@ -21,59 +21,38 @@ public static class ForwardTools
...
-         "author) — same replace-on-collision semantics, your ORIGINAL rewritten, no houseCARL backup or undo, and " +
-         "the same acknowledge= consent as the sibling write tools.\n\n" +
+         "Each axis's grammar is on its own parameters:\n" +
+         "WHAT — formids=, the set of records to copy and the rule for which ones may be named.\n" +
+         "WHOSE — source=, the ONE plugin every record in the call is copied from, active or only on disk.\n" +
+         "LANE — the default, a NEW patch= | into= an existing houseCARL patch | in_place= an existing plugin " +
+         "(opt-in), with acknowledge=. Exactly one lane per call; naming two is refused, never silently ignored.\n" +
+         "TRANSPORT — dry_run= | readback= | format= | max_chars=.\n\n" +
+         "ALL-OR-NOTHING (Q3): one rejected target refuses the WHOLE call, with a reason per record, and NOTHING is " +
+         "written; each parameter above carries its own refusals.\n\n" +
```
Avick3110, line 32: "each parameter above carries its own refusals" is not true of the two lane parameters. `in_place=` names none of the refusals the engine raises on that lane (target not an active plugin, `LoadOrderService.cs:6228`; target excluded as unparseable, `WritePatchBuilder.cs:1785`; a localized target, `LoadOrderService.cs:6237`; a parent folder that is missing or unwritable, `LoadOrderService.cs:5877-5899`), and `into=` names none of its own either (no houseCARL patch by that name, `LoadOrderService.cs:8330`; a folder houseCARL does not own, 8300; several owned folders carrying the same .esp, 8268). That matters because the sentence replaced a list that at least read as closed: a caller sent to the parameter for the refusal now finds nothing there. Either name those refusals on `into=` and `in_place=`, or say the parameters carry the refusals they document rather than promising all of them.

src/housecarl-mcp/ForwardTools.cs
```diff
@@ -21,59 +21,38 @@ public static class ForwardTools
...
-         "existing patch's filename>' EXTENDS that patch — and if it ALREADY carries a forwarded FormKey, its " +
-         "existing override is REPLACED by source's body (xEdit's copy-as-override overwrite, flagged per record). " +
-         "in_place='<plugin filename>' forwards INTO an existing plugin's OWN file (incl. one houseCARL didn't " +
-         "author) — same replace-on-collision semantics, your ORIGINAL rewritten, no houseCARL backup or undo, and " +
-         "the same acknowledge= consent as the sibling write tools.\n\n" +
+         "Each axis's grammar is on its own parameters:\n" +
+         "WHAT — formids=, the set of records to copy and the rule for which ones may be named.\n" +
+         "WHOSE — source=, the ONE plugin every record in the call is copied from, active or only on disk.\n" +
+         "LANE — the default, a NEW patch= | into= an existing houseCARL patch | in_place= an existing plugin " +
+         "(opt-in), with acknowledge=. Exactly one lane per call; naming two is refused, never silently ignored.\n" +
```
Avick3110: "Exactly one lane per call; naming two is refused, never silently ignored" is now the second copy of the rule: `patch=` (line 43) already says "Mutually exclusive with into= and in_place= — naming both lanes is refused, never silently ignored." The wave's method everywhere else has been to keep one copy (e.g. `ecc396cc`, "Drop the archive-name rule the pole now states twice"). Here the description's copy is the better one to keep, because `into=` and `in_place=` never state the rule at all and a caller reading either of those learns it only from the description — so drop the sentence from `patch=` instead.

src/housecarl-mcp/ForwardTools.cs
```diff
@@ -21,59 +21,38 @@ public static class ForwardTools
...
+         "TRANSPORT — dry_run= | readback= | format= | max_chars=.\n\n" +
```
Avick3110: `dry_run=` is filed under TRANSPORT here, but its own parameter opens "DRY RUN:" (line 51) while `readback=`/`format=`/`max_chars=` open "TRANSPORT:", and `housecarl_apply` — the tool this description calls its inverse, and the only other tool with the parameter — lists it on the LANE line instead (`ApplyTools.cs:42`, "LANE — patch= | into= | in_place= with acknowledge= | dry_run="). The roster is now the description's only index of the parameters, so the axis it assigns is the one the caller learns; two sibling tools filing the same parameter under different axes is the drift the roster exists to prevent. Move it to the LANE line, as apply has it.

plugin/CHANGELOG.md
```diff
@@ -13,15 +13,18 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
-  forms, and a line per axis naming its parameters. Eleven other tools are still over the cap: a test on the
-  published tool list holds the bound and names each of them, and they come under it one tool at a time.
+- **Every tool description now fits inside the 2,048 characters Claude Code shows, and the grammar that was
+  being cut off lives on the parameters it belongs to.** A description longer than that cap lost its tail on
+  the way to the model, so predicate languages, source and comparison rules, refusal lists, projection forms
+  and transport rules were being described in a part of the text nobody read. Twelve tools were over it —
+  `housecarl_records`, `housecarl_apply`, `housecarl_check`, `housecarl_create`, `housecarl_copy`,
+  `housecarl_forward`, `housecarl_merge_plugins`, `housecarl_nif_inspect`, `housecarl_nif_set`,
+  `housecarl_compact_plugin`, `housecarl_place` and `housecarl_remove` — and each came under it in its own
+  change. None of the text was deleted: each piece moved into the description of the parameter it is about,
```
Avick3110: "None of the text was deleted" is not true as shipped, and this PR's own body says so: it narrowed two rules rather than moving them — the refusal list stopped reading as exhaustive, and "the record's ORIGIN plugin must still be active" gained the exemption for an origin that is the file being written. Earlier changes in the wave dropped a description's copy of a rule the parameter already carried. The file's head guidance asks for entries that are true, and a reader who relied on the old unconditional sentence would go looking for it. Say what actually happened — the grammar moved onto the parameters, and where a description's sentence was wider than the code it was corrected — rather than promising nothing was cut.

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: With the set empty and the comment saying "A tool that goes over is fixed rather than listed here", the machinery it feeds is now dead in one direction and an escape hatch in the other: `if (StillOversized.Contains(tool)) return;` (line 53) is the only way a published description can exceed the bound without turning the theory red, and the comment right here forbids anyone using it. The Fact then iterates nothing. The CHANGELOG entry in this same PR promises "a description cannot grow back past the cut without going red" — which is true only while nobody adds a line. Deleting the set, the early return and the Fact makes that promise unconditional and costs no coverage: the theory already drives all 31 published tools off the name capture. This also matches CLAUDE.md's "don't add a guard to catch a mistake; fix the mistake", and it removes the stale message the other review comment on this line is about. If you would rather keep the scaffolding for a future wave, that is a defensible call — but then the comment should not say the set must stay empty.

src/housecarl-mcp/ForwardTools.cs
Avick3110: Fixed rather than narrowed: the ResolveOutputPath catch now returns with { Stamp = offEpoch }, so that refusal carries the epoch like every other post-capture one and the sentence holds (cb9d9352).

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: Folded with the set itself: the message now points only at the capture (f2975f75).

src/housecarl-mcp/ForwardTools.cs
Avick3110: Folded: into= and in_place= now each name the refusals the engine raises on that lane, in the engine's words (b1589c87).

src/housecarl-mcp/ForwardTools.cs
Avick3110: Folded: dropped from patch=, kept in the description (e2ce05a0).

src/housecarl-mcp/ForwardTools.cs
Avick3110: Folded: dry_run= now sits on the LANE line, as apply has it (b90a4936).

plugin/CHANGELOG.md
Avick3110: Folded: the entry now says the grammar moved and that a rule stated wider than the code was corrected to the code (68fa4721).

src/housecarl-mcp-tests/PublishedDescriptionBoundTests.cs
Avick3110: Folded: the set, the early return and the Fact are gone, so the theory over every published tool is the whole test (f2975f75).

PR #602, opened 2026-09-06
Gather a walk's bodies per hop so a big selection fits in memory

### LINKED ISSUE #556: [Bug] plugins=-scoped selection feeding the chain walk throws OutOfMemoryException at 2235 NPCs
## Summary

Found while verifying #236 on a `main` build (a43a0cd) against the live order (2026-09-05 overnight, note "houseCARL W4-13 verification.md" in the Claude Output vault).

The single-call form of the template-chain walk, `housecarl_records` with a `plugins=`-scoped selection feeding `chain`, dies with an internal `OutOfMemoryException` at the issue's own scale: Requiem.esp's 2235 NPCs. Reproduced twice uncontended, with and without `seed_paths`.

The two-step split works at comparable size: scan to `to_file`, then `formids=["@file"]` into the walk (1687 seeds, 4582 nodes, 0 errors). So the blow-up is in the scan-scoped selection feeding the walk, not in the walk itself.

## Why it matters

An internal exception is not one plain sentence saying what to try; the two-step remedy exists but nothing in the response names it. Either the scan-scoped selection streams into the walk at that scale, or the call refuses at the scale with the two-step remedy named.

## Repro

Live order, `plugins=["Requiem.esp"]`, type NPC_, the chain walk over Template, no other filters. Exact bodies are in the vault note.

Avick3110:
A `housecarl_records` walk read each seed's and each reached node's body with `LoadOrderResolver.SeekBody` — one enumeration of the whole winning plugin to find one record, the same per-row cost #582 took out of the scan and batch lanes but never out of the walk. A walk's cost was therefore the number of seeds times the winning plugin's record count, in time and in allocation, which is what ran a 2,235-NPC template-chain walk out of memory. The walk now advances every seed one hop together, so each hop's bodies come from one enumeration per source plugin through the same `BodyPrefetch` gather the other lanes use, and a walk costs what its seeds cost. The engine entry also takes the call's `CancellationToken` and polls it per seed and per hop, like every other loop that reads a body.

Closes #556

## The measurement

Synthetic world, one master plugin, NPC seeds on a shared three-deep template chain, walked with `project.form='chain'`. `LoadOrderResolver.BodySeeks` counts the per-record whole-plugin walks; allocation is `GC.GetTotalAllocatedBytes(precise: true)` and the peak is the heap right after the call.

**The cost is the seeds times the plugin, not the seeds.** Holding the walk fixed and growing only the plugin's *other* records moves the per-seek cost in step, which is the whole finding:

| seeds | records in the plugin | seeks | allocated per seek |
|---|---|---|---|
| 402 | ~405 | 402 | 0.33 MB |
| 402 | ~1,605 | 402 | 1.06 MB |
| 402 | ~3,205 | 402 | 2.00 MB |

So the call's total is quadratic when the seeds grow with the plugin they live in — 400 → 132 MB, 1,600 → 1,913 MB, 3,200 → 7,567 MB allocated, with 38 MB / 579 MB / 2,596 MB still on the heap when the call returns. Requiem.esp's 2,235 NPCs in a plugin of that mod's size is well past where a 64-bit process gives up.

**Before and after, same worlds, identical answers** (`seeds=`, `reached=`, `errors=` match to the row in every case):

| the walk | allocated before | allocated after | heap at return, before → after |
|---|---|---|---|
| 400 seeds, template chain | 132 MB | 12 MB | 38 MB → 12 MB |
| 1,600 seeds, template chain | 1,913 MB | 48 MB | 579 MB → 48 MB |
| 3,200 seeds, template chain | 7,567 MB | 98 MB | 2,596 MB → 57 MB |
| 400 seeds, 2,800 unrelated records beside them | 804 MB | 14 MB | 92 MB → 14 MB |
| 800 seeds, closure walk reaching 3,200 distinct nodes | 6,123 MB | 20 MB | 1,362 MB → 20 MB |

Seeks went from `seeds + distinct nodes` to 0 in every one of those, and growth is now linear in the seeds (400 → 12 MB, 1,600 → 48 MB, 3,200 → 98 MB).

**On the issue's own framing.** #556 reads the two-step split (`to_file=`, then `formids=["@file"]`) succeeding at 1,687 seeds as evidence that the blow-up is in the scan-scoped selection feeding the walk. It is not: the two shapes hand `WalkForwardBatch` the same seed list and, measured side by side on every world above, cost the same to within 1% both before and after this change (ratio 1.00). The two-step run survived because it was smaller — 1,687 seeds against 2,235 is a 1.75x difference on a quadratic curve — and because its scan half had already released. The cost was in the walk itself, and it is fixed there, so both shapes now run and neither is a workaround for the other.

## What changed in the walk

The per-seed loop became: parse and gather every seed's body, start each seed's first hop, then advance all seeds one hop at a time, gathering the hop's frontier in `BodyPrefetch.ChunkRows`-sized chunks. Per-seed state (visited set, node rows, cycles, node budget, truncation note) is unchanged and lives on a `WalkSeedState`; results come back in seed order as before. A key the gather does not return stays out of the cache, so the per-record read still raises whatever it always raised — the gather is an optimisation, not a second error path.

**The gather is bounded by the node budget, not by the frontier.** `walk.max_nodes` is consulted on dequeue, so gathering a hop's whole frontier first meant a seed that had already filled its cap still made the call read and pin bodies that never reached a row — the same allocation axis the issue died on. Each seed's remaining budget now rides into the gather: a seed at its cap contributes nothing, a seed near it contributes only what it can still admit. Over 302 seeds capped at one node each, the call asks the gather for 601 bodies where it asked for 1,202. The cap itself is unchanged — a node at the cap is recorded, not entered, with the same truncation sentence.

**A finished seed's bookkeeping goes back at once.** A seed is finished the moment its frontier empties, and the results loop reads neither its visited set nor its queue, so both are released there rather than at return. The old per-seed loop released both when that seed finished; holding all of them for the whole call would have grown with the seed count exactly the way the seek cost did.

One behaviour does shift, and it is in the code comment: an exclusion with `severity='refuse'` ends the whole call, and the sentence now names the first seed in seed order that reaches the excluded class at the shallowest hop any seed reaches it. Before, seeds ran to completion one after another, so an earlier seed's deep refusal outranked a later seed's shallow one. Either way the call refuses and returns nothing; only which seed and chain the sentence quotes can differ.

## The bound on the reached set

A walk that no longer runs out of memory needs something else to stop it. The walk lane returns above the scan's render bound, deliberately, because the seed scan's count is not the rendered count. The **reached** count is: when a body form consumes the walk, the reached set becomes the list lane's `formids` and every row of it reads a record body — up to seeds × `walk.max_nodes` with no ceiling at all once the OOM is gone.

So the reached count is now measured against the same bound, with the same two lanes (a named-fields row and an `everything` row are two orders of magnitude apart), before a single body is read. That holds on every walk lane and for every reading form that consumes the reached set: `summary` and `aggregate` read one cheap leaf, but they read it off a body per row exactly as `fields`, `rows` and `everything` do, and both reverse walks hand the list lane their reached set the same way the forward walk does.

The remedy sentence is the walk's own rather than the scan's, since `limit=`/`offset=` window the render and not the walk, and each lane names only levers that move it: the seeds the caller passed (named without the scan terms, because a walk is seeded from the `formids=` lane too, where those parameters are not part of the call), `walk.depth` and `walk.max_nodes` where they apply, and `project.form='chain'` on the forward and carrier lanes — the two chain can draw. The transitive reverse lane names `walk.depth` instead and says why chain is not a lever there. `form='chain'` itself stays exempt: it renders the walk's own rows, so it pays no second body read on top of the walk's own — the walk reads one per reached node whatever the form, which is what the remedy now says. The sentence lives on `walk.max_nodes`, where the budget it bounds is described; `housecarl_records`'s own description is untouched.

## Tests

`src/housecarl-mcp-tests/RecordsWalkCostTests.cs`, on a `WalkCostWorld` of 300 NPC seeds on one template chain, each carrying 3 items of its own, in a master of ~1,500 records. Tagged `tier=integration`. `BodyPrefetch.KeysWanted` is a new process counter of the bodies the bulk gather has been asked for, for the reason `LoadOrderResolver.BodySeeks` is one: gathering only what a lane can use, or a whole frontier it then throws away, is invisible in the answer.

| test | before | after |
|---|---|---|
| `AWalkGathersItsSeedBodiesOncePerPluginNotOncePerSeed` | fails — 302 per-record plugin walks for 302 seeds | passes (0) |
| `AWalkGathersEachHopsBodiesTogetherNotOnePerNode` | fails — 1,202 per-record plugin walks for a closure walk | passes (0) |
| `AWalkCostsWithItsSeedsNotWithThePluginBehindThem` | fails — allocates 240 MB, past the 37 MB the seed count allows | passes (9 MB) |
| `ACappedSeedGathersNoBodiesPastItsCap` | fails — 1,202 bodies gathered for 302 seeds allowed one node each | passes (601, and no fallback seek) |
| `AWalkCancelledBeforeItStartsStopsInTheSeedGather` | did not compile — the engine entry took no token | passes |
| `AWalkStopsBetweenHopsWhenTheClientCancels` | fails with the hop loop's two polls deleted | passes |
| `AReachedSetTooBigToRenderRefusesNamingTheChainFormAndNarrowerSeeds` | fails — the reached set renders unbounded | passes |
| `TheChainFormIsNotHeldToTheBodyRenderBound` | passes — the chain form is exempt and stays exempt | passes |
| `AReachedSetUnderTheBoundRendersAsBefore` | passes — under the bound nothing changes | passes |
| `AHopDoesNotSpendItsGatherOnNodesTheSeedAlreadyVisited` | fails — 40 per-record plugin walks when a hop's frontier leads with revisits | passes (0) |
| `EveryReadingFormOverTheReachedSetIsHeldToTheBound` (summary, aggregate) | fails — neither form was gated | passes |
| `AFormidsSeededWalkIsRefusedWithoutNamingTheScanTerms` | fails — the refusal opened on `types=`/`plugins=`/`where=` | passes |
| `ATransitiveReverseWalkTooBigToRenderRefuses` | fails — the lane rendered unbounded | passes |
| `AReverseCarrierWalkTooBigToRenderRefusesNamingTheChainForm` | fails — the lane rendered unbounded | passes |

Fail-before was verified in place for each: making the gather budget unbounded, deleting the two hop-loop polls, and neutralising the reached-set bound, each re-run on its own; the numbers in the table are those runs' own messages. The allocation test is bytes, not a clock, so there is no timing assertion to go flaky: it calls the walk once to pay the index build and the JIT, then measures a second identical call and holds it to 128 KB per seed — about 4x above what it now spends and about 6x below what it spent before. Assembly-wide parallelisation is already off, so the process-wide allocation, seek and gather counters are these tests' alone.

The base branch's `AWalkIsNotRefusedForTheSizeOfItsSeedScan` pinned the walk lane as exempt from the render bound. Its claim — a walk is not refused for the size of the scan that seeded it — still holds and is what it still asserts; it is renamed `AWalkIsMeasuredOnWhatItReachedAndNotOnItsSeedScan` and now checks that the refusal it gets is the walk lane's, naming the walk's levers and not the scan window's.

## What was run

From inside `.claude/worktrees/walk-oom`:

```
```

0 build errors, 1,692 tests passed, 0 failed, and `ci-all` 127/127. `housecarl_records`'s own description is untouched and `PublishedDescriptionBoundTests` passes. Nothing was run against the live MO2 instance.

### DIFF
--- a/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
+++ b/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
@@ -1,0 +1,197 @@
+using Mutagen.Bethesda;
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Skyrim;
+using HousecarlCore;
+using HousecarlGenerator;
+using HousecarlMcp;
+using Xunit;
+
+namespace HousecarlMcpTests;
+
+/// <summary>
+/// A world whose point is the SHAPE a walk pays for: many seeds sitting in one plugin that also holds many other
+/// records. A walk's old per-node body read was a whole-plugin seek, so its cost was the seed count TIMES the
+/// plugin's record count; a handful of records in a small plugin cannot show that, and the shared worlds are both.
+/// </summary>
+public sealed class WalkCostWorld : IDisposable
+{
+    /// <summary>NPC seeds, each on the same template chain.</summary>
+    public const int Seeds = 300;
+
+    /// <summary>Distinct items per seed, so a closure walk's first hop reaches records nothing else reaches.</summary>
+    public const int ItemsPerSeed = 3;
+
+    public string Root { get; }
+    public string MasterName { get; }
+    public LoadOrderService Svc { get; }
+
+    readonly string _priorCorpusPath;
+
+    public WalkCostWorld()
+    {
+        _priorCorpusPath = CorpusRulebook.CorpusPath;
+        Root = Path.Combine(Path.GetTempPath(), "hc-walkcost-tests-" + Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(Path.Combine(Root, "game", "Data"));
+
+        var masterKey = new ModKey("HcWalkCostMaster", ModType.Master);
+        MasterName = masterKey.FileName.String;
+        var master = new SkyrimMod(masterKey, SkyrimRelease.SkyrimSE);
+
+        var top = master.Npcs.AddNew();
+        top.EditorID = "HcWalkTemplateTop";
+        var mid = master.Npcs.AddNew();
+        mid.EditorID = "HcWalkTemplateMid";
+        mid.Template.SetTo(top);
+        mid.Configuration.TemplateFlags = NpcConfiguration.TemplateFlag.Stats;
+
+        for (int i = 0; i < Seeds; i++)
+        {
+            var n = master.Npcs.AddNew();
+            n.EditorID = "HcWalkSeed" + i;
+            n.Template.SetTo(mid);
+            n.Configuration.TemplateFlags = NpcConfiguration.TemplateFlag.Stats;
+            n.Items = new Noggog.ExtendedList<ContainerEntry>();
+            for (int j = 0; j < ItemsPerSeed; j++)
+            {
+                var item = master.Ammunitions.AddNew();
+                item.EditorID = $"HcWalkItem{i}_{j}";
+                n.Items.Add(new ContainerEntry { Item = new ContainerItem { Item = item.ToLink(), Count = 1 } });
+            }
+        }
+
+        var instance = Path.Combine(Root, "inst");
+        var mods = Path.Combine(instance, "mods");
+        Directory.CreateDirectory(Path.Combine(mods, "WalkCostMod"));
+        master.BeginWrite.ToPath(Path.Combine(mods, "WalkCostMod", MasterName))
+              .WithLoadOrder(Array.Empty<ISkyrimModGetter>()).Write();
+
+        var genDir = Path.Combine(Root, "corpus-gen");
+        CorpusGenerator.GenerateAll(genDir, Path.Combine(Root, "corpus-ref"));
+        CorpusRulebook.CorpusPath = Path.Combine(genDir, "corpus.json");
+
+        File.WriteAllText(Path.Combine(instance, "ModOrganizer.ini"),
+            "[General]\r\ngameName=Skyrim Special Edition\r\nselected_profile=@ByteArray(Default)\r\ngamePath=@ByteArray("
+            + Path.Combine(Root, "game").Replace(@"\", @"\\") + ")\r\n");
+        var prof = Path.Combine(instance, "profiles", "Default");
+        Directory.CreateDirectory(prof);
+        File.WriteAllText(Path.Combine(prof, "loadorder.txt"), "# header\r\n" + MasterName + "\r\n");
+        File.WriteAllText(Path.Combine(prof, "plugins.txt"), "*" + MasterName + "\r\n");
+        File.WriteAllText(Path.Combine(prof, "modlist.txt"), "# header\r\n+WalkCostMod\r\n");
+
+        Svc = LoadOrderService.WithInstance(instance, 0, new UserConfigStore(Path.Combine(Root, "user.json")));
+    }
+
+    public void Dispose()
+    {
+        CorpusRulebook.CorpusPath = _priorCorpusPath;
+        Svc.Dispose();
+        try { Directory.Delete(Root, true); } catch { /* temp cleanup best-effort */ }
+    }
+}
+
+/// <summary>The shared walk-cost world. One build per collection.</summary>
+public sealed class WalkCostFixture : IDisposable
+{
+    public WalkCostWorld W { get; } = new();
+    public void Dispose() => W.Dispose();
+}
+
+/// <summary>Its own collection: <c>CorpusRulebook.CorpusPath</c> is a process-global and one world owns it at a time.</summary>
+[CollectionDefinition("walk-cost")]
+public sealed class WalkCostCollection : ICollectionFixture<WalkCostFixture> { }
+
+/// <summary>
+/// What a walk over a big selection costs (#556). A scan-scoped selection feeding the chain walk died with an
+/// internal OutOfMemoryException at 2,235 NPC seeds, because the walk read each node's body with the whole-plugin
+/// seek #582 took out of the scan and batch lanes: one walk of the winning plugin per seed and per reached node, so
+/// the call's cost was the seed count times the plugin's record count.
+/// </summary>
+[Collection("walk-cost")]
+[Trait("tier", "integration")]
+public sealed class RecordsWalkCostTests
+{
+    readonly WalkCostWorld _w;
+    public RecordsWalkCostTests(WalkCostFixture f) => _w = f.W;
+
+    LoadOrderService Svc => _w.Svc;
+    static readonly string[] Npc = { "NPC_" };
+    static RecordsTools.RecordsProject Chain() => new() { form = "chain" };
+    RecordsTools.RecordsScope Scope() => new() { names = new[] { _w.MasterName } };
+
+    /// <summary>The template chain, the shape the issue was reported on.</summary>
+    static RecordsTools.RecordsWalk TemplateWalk() =>
+        new() { follow = "Template", seed_paths = new[] { "Template" } };
+
+    // ---- the cost itself ---------------------------------------------------------------------------
+
+    /// <summary>The seeds are known before the walk starts, so their bodies come from one enumeration per source
+    /// plugin. Before, each seed cost a whole-plugin seek of its own.</summary>
+    [Fact]
+    public void AWalkGathersItsSeedBodiesOncePerPluginNotOncePerSeed()
+    {
+        var before = LoadOrderResolver.BodySeeks;
+        var response = RecordsTools.Records(Svc, types: Npc, plugins: Scope(), walk: TemplateWalk(),
+                                            project: Chain(), counts_only: true);
+        var seeks = LoadOrderResolver.BodySeeks - before;
+
+        Assert.Contains($"seeds={WalkCostWorld.Seeds + 2}", response);
+        Assert.True(seeks <= 1, $"{WalkCostWorld.Seeds + 2} walk seeds cost {seeks} per-record plugin walks.");
+    }
+
+    /// <summary>Every seed advances one hop together, so a hop's reached nodes are one gather too. Before, a closure
+    /// walk paid a whole-plugin seek per distinct node as well as per seed — the same call without seed_paths, which
+    /// is how the issue was also reproduced.</summary>
+    [Fact]
+    public void AWalkGathersEachHopsBodiesTogetherNotOnePerNode()
+    {
+        var before = LoadOrderResolver.BodySeeks;
+        var response = RecordsTools.Records(Svc, types: Npc, plugins: Scope(),
+                                            walk: new RecordsTools.RecordsWalk { depth = 1 },
+                                            project: Chain(), counts_only: true);
+        var seeks = LoadOrderResolver.BodySeeks - before;
+
+        Assert.Contains($"reached={WalkCostWorld.Seeds * (WalkCostWorld.ItemsPerSeed + 1) + 1}", response);
+        Assert.True(seeks <= 1, $"a closure walk over {WalkCostWorld.Seeds} seeds cost {seeks} per-record plugin walks.");
+    }
+
+    /// <summary>What the seeks meant in memory. A walk's allocation now scales with the SEEDS, not with the seeds
+    /// times the plugin behind them: 9 MB over these 302 seeds, where the per-node seek spent 240 MB on the same
+    /// call and rose with every record added to the plugin. The bound is allocated bytes, which a run reports
+    /// exactly — no clock, so no flaky timing.</summary>
+    [Fact]
+    public void AWalkCostsWithItsSeedsNotWithThePluginBehindThem()
+    {
+        // The same call once first: the index build and the JIT of everything below are one-time costs of the
+        // process, not of a walk, and measuring them would make the number depend on which test ran first.
+        string Walk() => RecordsTools.Records(Svc, types: Npc, plugins: Scope(), walk: TemplateWalk(),
+                                              project: Chain(), counts_only: true);
+        Walk();
+
+        var before = GC.GetTotalAllocatedBytes(precise: true);
+        Walk();
+        var allocated = GC.GetTotalAllocatedBytes(precise: true) - before;
+
+        const long perSeedCeiling = 128 * 1024;
+        var ceiling = perSeedCeiling * (WalkCostWorld.Seeds + 2);
+        Assert.True(allocated < ceiling,
+                    $"the walk allocated {allocated / 1048576} MB over {WalkCostWorld.Seeds + 2} seeds — past the {ceiling / 1048576} MB this world's seed count allows.");
+    }
+
+    // ---- and it stops when the client stops waiting -------------------------------------------------
+
+    /// <summary>The walk's hop loop honours the call's cancellation token, like every other loop that reads a body
+    /// (#582). Its engine entry took no token at all before.</summary>
+    [Fact]
+    public void AWalkStopsWhenTheClientCancels()
+    {
+        using var cts = new CancellationTokenSource();
+        cts.Cancel();
+        var q = Svc.CrossQuery(Npc, null, null, false, new[] { _w.MasterName }, null, int.MaxValue);
+        Assert.Null(q.Error);
+        var seeds = q.Keys.Select(k => k.ToString()).ToArray();
+
+        Assert.Throws<OperationCanceledException>(() =>
+            Svc.WalkForwardBatch(seeds, new[] { "Template" }, "Template", 16, 2000,
+                                 Array.Empty<(string, bool)>(), null, out _, out _, cts.Token));
+    }
+}
--- a/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
+++ b/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
@@ -116,6 +116,7 @@ public sealed class RecordsWalkCostTests
     LoadOrderService Svc => _w.Svc;
     static readonly string[] Npc = { "NPC_" };
     static RecordsTools.RecordsProject Chain() => new() { form = "chain" };
+    static RecordsTools.RecordsProject Fields() => new() { form = "fields", fields = new[] { "EditorID" } };
     RecordsTools.RecordsScope Scope() => new() { names = new[] { _w.MasterName } };
 
     /// <summary>The template chain, the shape the issue was reported on.</summary>
@@ -204,6 +205,49 @@ public sealed class RecordsWalkCostTests
         Assert.True(seeks <= 1, $"a capped walk cost {seeks} per-record plugin walks.");
     }
 
+    // ---- what the reached set costs to RENDER --------------------------------------------------------
+
+    /// <summary>The walk lane returns above the scan's render bound, because the seed count is not the rendered
+    /// count. The reached count is, so a reading form that consumes the reached set is measured against the same
+    /// bound — otherwise a walk that no longer runs out of memory hands the list lane seeds times walk.max_nodes
+    /// bodies with no ceiling at all. The remedy is the walk's own, not the scan's window.</summary>
+    [Fact]
+    public void AReachedSetTooBigToRenderRefusesNamingTheChainFormAndNarrowerSeeds()
+    {
+        var response = WithBound(10, () =>
+            RecordsTools.Records(Svc, types: Npc, plugins: Scope(), walk: TemplateWalk(), project: Fields()));
+
+        Assert.StartsWith("error:", response);
+        Assert.Contains("project.form='chain'", response);
+        Assert.Contains("walk.max_nodes", response);
+        Assert.Contains("walk.depth", response);
+        Assert.DoesNotContain("re-scans", response);          // the scan's window is not what moves a walk
+        Assert.DoesNotContain("does not combine", response);
+    }
+
+    /// <summary>The chain form stays exempt: it renders the walk's own rows and reads no record body, so the same
+    /// call under the same bound serves.</summary>
+    [Fact]
+    public void TheChainFormIsNotHeldToTheBodyRenderBound()
+    {
+        var response = WithBound(10, () =>
+            RecordsTools.Records(Svc, types: Npc, plugins: Scope(), walk: TemplateWalk(),
+                                 project: Chain(), counts_only: true));
+
+        Assert.DoesNotContain("error:", response);
+        Assert.Contains($"seeds={WalkCostWorld.Seeds + 2}", response);
+    }
+
+    /// <summary>Under the bound nothing changes: the same reading form serves the set the walk reached.</summary>
+    [Fact]
+    public void AReachedSetUnderTheBoundRendersAsBefore()
+    {
+        var response = WithBound(WalkCostWorld.Seeds * 4, () =>
+            RecordsTools.Records(Svc, types: Npc, plugins: Scope(), walk: TemplateWalk(), project: Fields()));
+
+        Assert.DoesNotContain("error:", response);
+    }
+
     // ---- and it stops when the client stops waiting -------------------------------------------------
 
     /// <summary>The engine entry takes the call's token and hands it to the body gather, so a walk cancelled before
@@ -237,6 +281,16 @@ public sealed class RecordsWalkCostTests
     }
 
     // ---- helpers -----------------------------------------------------------------------------------
+
+    /// <summary>Run one call with the named-fields render bound moved, restored whatever happens — building
+    /// 300,000 reachable records to meet the real one is not a test.</summary>
+    static string WithBound(int rows, Func<string> call)
+    {
+        var prior = RenderBudget.MaxRenderRows;
+        RenderBudget.MaxRenderRows = rows;
+        try { return call(); }
+        finally { RenderBudget.MaxRenderRows = prior; }
+    }
 
     /// <summary>Every NPC in the world, as walk seeds.</summary>
     string[] Seeds()
--- a/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
+++ b/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
@@ -206,19 +206,61 @@ public sealed class RecordsWalkCostTests
 
     // ---- and it stops when the client stops waiting -------------------------------------------------
 
-    /// <summary>The walk's hop loop honours the call's cancellation token, like every other loop that reads a body
-    /// (#582). Its engine entry took no token at all before.</summary>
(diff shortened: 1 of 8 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3911,72 +3958,97 @@ seedWin is null
...
+        // A node at the depth cap is recorded and not entered, so nothing is ever queued past `depth`.
+        for (int d = 1; d <= depth; d++)
+        {
+            ct.ThrowIfCancellationRequested();
+            var frontier = new List<FormKey>();
+            foreach (var s in states)
+                if (s is not null)
+                    foreach (var q in s.Frontier) frontier.Add(q.Key);
+            if (frontier.Count == 0) break;
+            Prefetch(frontier);
```
Avick3110: **The `max_nodes` cap no longer bounds how many bodies a walk reads or retains.**

`Prefetch(frontier)` gathers the entire hop frontier of every seed, and each gathered body is pinned in `bodyCache` for the rest of the call. The per-seed cap is only consulted later, on dequeue.

Concrete: a closure walk whose seed fills the default 2000-node cap on hop 1. Those 2000 nodes are all *expanded*, so the seed's frontier carries 2000 x (links per record) keys — tens of thousands. Hop 2 then prefetches and caches every one of them, and only after that does the first dequeue trip `st.Nodes.Count >= maxNodes`, write the truncation note and `Clear()` the frontier. Not one of those bodies reaches a row. The old loop read exactly one body past the cap; this reads and retains the whole frontier, times the seed count — the same allocation axis #556 died on, and it also defeats the chunk bound `BodyPrefetch`'s own doc comment promises ("the map, and the getters it pins, are bounded by the chunk").

Folding the cap into the gather fixes both halves — a seed already at the cap is finished here instead of a hop later, and a seed near it contributes only what it can still record. Keys left out simply stay uncached, which is the behaviour the `Prefetch` comment already relies on.

```suggestion
            var frontier = new List<FormKey>();
            foreach (var s in states)
            {
                if (s is null || s.Frontier.Count == 0) continue;
                if (s.Nodes.Count >= maxNodes)
                {
                    s.Truncation = $"walk truncated: the {maxNodes}-node cap was reached — what is listed IS reached and proved; raise walk.max_nodes to walk further.";
                    s.Frontier.Clear();
                    continue;
                }
                int room = maxNodes - s.Nodes.Count, took = 0;
                foreach (var q in s.Frontier) { frontier.Add(q.Key); if (++took >= room) break; }
            }
            if (frontier.Count == 0) break;
            Prefetch(frontier);
```

src/housecarl-mcp-tests/RecordsWalkCostTests.cs
```diff
@@ -0,0 +1,197 @@
...
+    [Fact]
+    public void AWalkStopsWhenTheClientCancels()
+    {
+        using var cts = new CancellationTokenSource();
+        cts.Cancel();
+        var q = Svc.CrossQuery(Npc, null, null, false, new[] { _w.MasterName }, null, int.MaxValue);
+        Assert.Null(q.Error);
+        var seeds = q.Keys.Select(k => k.ToString()).ToArray();
+
+        Assert.Throws<OperationCanceledException>(() =>
```
Avick3110, line 345: The test cancels before the call, so the throw comes out of the very first `Prefetch(seedGather)` — `BodyPrefetch.Gather` polls the token in its per-plugin loop, before a single hop has run. The two `ct.ThrowIfCancellationRequested()` calls this PR adds to the hop loop (`LoadOrderService.cs:3976` and `:3987`) are never reached, so deleting both of them leaves this test green and the summary's "polls it per seed and per hop" unproved.

What it actually proves is that the engine entry now takes a token and hands it to the gather, which is worth having — but the doc comment claims the hop loop. Either say that, or cancel from inside the walk so the hop loop is the thing that throws: a large enough seed set with a `CancellationTokenSource` cancelled on a timer, or a token cancelled from a callback the walk crosses, would reach it.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3799,16 +3799,36 @@ public sealed record WalkSeedResult(string Seed, string? Type, string? EditorId,
...
+    sealed class WalkSeedState
+    {
+        public FormKey Key;
+        public IMajorRecordGetter Body = null!;
+        public string Type = "";
+        public string Label = "";
+        public List<WalkNodeRow> Nodes = new();
+        public List<string> Cycles = new();
+        public string? Truncation;
+        public HashSet<FormKey> Visited = new();
```
Avick3110, line 3823: Every seed's `Visited` set and `Frontier` queue now live for the whole call; the old per-seed loop released both the moment that seed finished. Under `walk.max_nodes` at its default 2000 across the 2,235-seed selection from #556 that is up to ~4.5M `FormKey` entries of pure bookkeeping held at once, on top of the node rows the call actually returns — new retention on the axis this PR is fixing, and it grows with the seed count exactly the way the old seek cost did.

A seed is finished the moment its frontier empties (nothing but its own processing ever enqueues into it), and neither `Visited` nor `Frontier` is read again in the results loop — so a `st.Visited.Clear()` at the bottom of each seed's turn when `st.Frontier.Count == 0` costs nothing and gives most of it back.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -834,7 +834,7 @@ string RenderRev(SpillState? sp, out bool trunc) => json
             // Forward: one engine batch, one captured build. The chain form renders it; every other form
             // consumes the reached set — seeds included, and the render says so — through the normal lanes.
             var rows = svc.WalkForwardBatch(ids, walk!.seed_paths, walk.follow, walkDepth, walkMaxNodes,
-                                            walkExclusions, demand, out var wRefusal, out var wEpoch);
+                                            walkExclusions, demand, out var wRefusal, out var wEpoch, ct);
```
Avick3110, line 866: **A walk that is no longer stopped by memory is still stopped by nothing else.**

The walk lane returns above the render bound (`RenderBudget.Refuse`, RecordsTools.cs:1276-1277) — deliberately, per the comment there — and when the reached set is consumed by a body form the call falls through to `ListLane()`, whose `svc.ResolveBatch(ids, ...)` reads *every* reached record's body before `Windowed()` throws most of them away. Until this PR the OOM was the de-facto bound on that path at the issue's scale; now `WalkForwardBatch` returns quickly and hands `ListLane` up to `seeds x walk.max_nodes` records (2,235 x 2,000 with the defaults), which is exactly the silent multi-hour render #582/#596 set out to stop, on the one lane the bound skips.

Either measure `combined.Count` against `RenderBudget` before `formids = combined; return ListLane();`, or open a follow-up issue saying the walk-consume path is knowingly unbounded — but the branch should not leave it unsaid now that the walk itself finishes.

(Severity: low — pre-existing on the base branch, but this PR is what makes it reachable at #556's scale.)

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 150a63d4: the remaining per-seed budget now rides into the gather, so a seed at its cap gathers nothing and a seed near it gathers only what it can still admit — over 302 seeds capped at one node each that is 601 bodies asked of the gather where it was 1,202, and the cap semantics and sentence are untouched.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 71e708c8: neither the visited set nor the frontier is read again in the results loop, so both are released the moment a seed's frontier empties.

src/housecarl-mcp-tests/RecordsWalkCostTests.cs
Avick3110: Folded in 68f01525: the token is now tripped from inside the walk, through the exclusion list the walk reads once per reached node, so the hop loop's per-seed poll is what throws; deleting the two polls makes the new test fail. The before-entry case keeps a test of its own saying only what it proves.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 019a095c: the reached set is measured against the same two-lane render bound before it reaches ListLane, with the walk's own levers named rather than the scan window's — form=chain, which reads no body, and narrower seeds or lower caps. The chain form itself stays exempt, since it renders the walk's rows and not the records.

Avick3110:
Moved onto the rebased #596 (`claude/records-cost` is now `64f15ecb`, itself rebased onto the merged description wave). Old tip `019a095c` → new tip `4dabea12`; seven commits, none squashed.

No file conflicted — the replay was clean, including `plugin/CHANGELOG.md`, `RecordsTools.cs` and `RenderBudget.cs`.

From the rebased worktree:

```
```

1,651 passed, 0 failed. `ci-all` not run locally; CI runs it.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -64,7 +64,7 @@ public sealed class RecordsWalk
...
         public int? depth { get; set; }
 
-        [Description("The node budget (default 2000, the read-expansion budget). Per seed on a forward walk and on the reverse carrier walk (each seed's carrier rows); ONE budget shared across every seed and every hop on the transitive reverse walk, whose hops are one frontier and not a per-seed expansion. A breach keeps what was proved and says which reading it spent.")]
+        [Description("The node budget (default 2000, the read-expansion budget). Per seed on a forward walk and on the reverse carrier walk (each seed's carrier rows); ONE budget shared across every seed and every hop on the transitive reverse walk, whose hops are one frontier and not a per-seed expansion. A breach keeps what was proved and says which reading it spent. A reading form (summary/fields/rows/everything/aggregate) then renders the whole reached set — seeds times this budget at the worst — and reads a body per row, so it is held to the same render bound a scan is and refuses up front naming project.form='chain', which lists the same set without reading a body.")]
```
Avick3110: **The description names five reading forms; the gate covers three.**

This sentence says "A reading form (summary/fields/rows/everything/aggregate) then renders the whole reached set ... so it is held to the same render bound a scan is and refuses up front naming project.form='chain'". The gate added at `RecordsTools.cs:897` is `(bodyFields || form == "everything")`, and `bodyFields` is `form is "fields" or "rows"` (line 171). `summary` and `aggregate` are not held to anything.

They are not free rows either: line 573 routes both to `readFields = new[] { "EditorID" }` and they go through `ResolveBatch`, so a body is read per row exactly as the sentence says.

Scenario: `types=["NPC_"] walk={follow:"Template"} project={form:"summary"}` over 2,235 seeds. `combined.Count` is up to seeds x `max_nodes`, no refusal is emitted, and the list lane reads a body for every one of those rows. The published description promises a refusal the code does not emit — either widen the gate to every form that consumes the reached set, or narrow the parenthetical to the forms actually bounded.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -888,6 +888,16 @@ string Render(SpillState? sp, out bool trunc) => json
             if (combined.Count == 0)
                 return json ? JsonWire.RenderError($"the walk reached nothing readable ({seedErrs} seed error(s) — run form='chain' to see each seed's outcome).", wEpoch)
                             : $"error: the walk reached nothing readable ({seedErrs} seed error(s) — run form='chain' to see each seed's outcome)." + Wire.EpochLine(wEpoch);
+
+            // ---- the reached set's own render bound. The seed scan's count was never this count, which is why the
+            // scan's bound below the walk lane exempts a walk; the REACHED count is it — up to seeds x
+            // walk.max_nodes rows, each one a body read by the list lane. Same two lanes as the scan's bound, its
+            // own remedy, since the scan window is not what moves a walk. form='chain' stays exempt and returned
+            // above: it renders the walk's own rows, not the records' bodies.
+            if ((bodyFields || form == "everything") && !counts_only
```
Avick3110: **The bound lands on the forward lane only, but the description it is written for covers the reverse carrier lane too.**

The reverse carrier lane, ~90 lines above, ends the same way and gets no bound:

```csharp
expectEpoch = epochR?.Epoch;
formids = carrierSel.ToArray();
walk = null;
return ListLane();
```

And `max_nodes`'s own text (line 67) says the budget is "Per seed on a forward walk **and on the reverse carrier walk**" before going on to claim the reading form "is held to the same render bound a scan is".

Scenario: `walk={direction:"reverse", follow:"Effects[].BaseEffect"}` with a few hundred MGEF seeds and `project.form='everything'`. `carrierSel` is seeds x `max_nodes` at the worst, far past `MaxWholeRecordRows` (15,000), and every row is rendered with no refusal — the same unbounded body render this PR closes on the forward lane.

(The transitive reverse lane at ~721 shares one budget across all seeds so its set stays near `max_nodes`, but it is also unbounded here, and the remedy this sentence names — `project.form='chain'` — is explicitly refused for that lane at line 392, so the sentence would point a transitive-walk caller at a shape that refuses.)

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3911,72 +3958,117 @@ seedWin is null
...
+                if (s is null || s.Frontier.Count == 0) continue;
+                pending = true;
+                int room = maxNodes - s.Nodes.Count;
+                if (room <= 0) continue;                                  // at its cap: its turn below records the cut
+                int took = 0;
+                foreach (var q in s.Frontier)
                 {
-                    truncation ??= $"walk reached its depth cap ({depth}) on at least one chain — nodes at the cap are recorded, not entered; raise walk.depth to walk deeper.";
-                    continue;
+                    if (q.Key.IsNull || !gatherSeen.Add(q.Key)) continue;
```
Avick3110: **The gather budget is spent on frontier entries the seed is going to skip, which pushes the nodes it does record back onto the per-record seek.**

This fill loop screens an entry on `q.Key.IsNull` and the cross-seed `gatherSeen`. It never asks `s.Visited`. A revisit is therefore added to the gather — a body fetched and immediately discarded at dequeue — *and* it burns one of the seed's `room` slots.

The processing loop below skips revisits before the node counter, so the nodes the seed actually records come from further down the queue than the gathered prefix reaches, and each of those falls through `Prefetch`'s cache miss into `Fetch` → `view.GetRecord` → the whole-plugin `SeekBody` this PR exists to remove.

Scenario: closure walk, one seed, `maxNodes=2000`, `Nodes.Count=100` so `room=1900`, a hop-2 frontier of 5,000 entries of which ~3,000 are diamond revisits interleaved through the queue. The first 1,900 distinct keys are gathered; roughly 1,140 of them are revisits that never become rows; the ~1,140 nodes recorded from beyond that prefix each pay a whole-plugin walk.

Neither new test covers it: `WalkCostWorld`'s template chain has no revisits, and `ACappedSeedGathersNoBodiesPastItsCap` caps at one node so the prefix is never outrun.

```suggestion
                    if (q.Key.IsNull || s.Visited.Contains(q.Key) || !gatherSeen.Add(q.Key)) continue;
```

src/housecarl-mcp/RenderBudget.cs
```diff
@@ -70,11 +70,29 @@ internal static string Projected(int rows, bool wholeRecord)
...
+        "in windows with limit= and offset= — offset= re-scans the selection from the start, so a window costs " +
+        "more the further in it is. to_file= captures the COMPLETE selection rather than a window, so it renders " +
+        "every row and does not combine with offset=: narrow the scan until the whole set fits, then write it in " +
+        "one call.";
+
+    /// <summary>What moves a WALK's row count. The rows are what the walk REACHED, not what the scan selected, so
+    /// the scan window is the wrong lever: the seeds, the walk's own caps, or the chain form, which lists the same
+    /// reached set without reading a body.</summary>
+    internal const string WalkRemedy =
+        "the seeds (types=, plugins=, where=), or lower walk.depth or " +
```
Avick3110: **The first lever this remedy names does not exist on half the calls that can hit it.**

`WalkLane` is reached from the formids lane as well as the scan lane — `RecordsTools.cs:509`, `if (walk is not null) return WalkLane(ids, demand, echoSrc);`. On that path `types=`, `plugins=` and `where=` are not part of the call at all (`formids=` selects the seeds instead), so the refusal opens by telling the caller to narrow three parameters they did not pass and could not pass alongside `formids=`.

Scenario: `formids=["@seeds.json"] walk={follow:"Template"} project={form:"everything"}` over a large seed file refuses with "Or narrow the seeds (types=, plugins=, where=)". The two levers that *do* move it on that path — a smaller seed file, and `walk.depth` / `walk.max_nodes` — are second in the sentence, behind three that do nothing.

The scan's own remedy did not have this problem because the scan bound only ever fired on the scan lane. Naming the seeds without the parenthesis ("the seeds you passed") would hold on both entries.

src/housecarl-mcp/RenderBudget.cs
```diff
@@ -70,11 +70,29 @@ internal static string Projected(int rows, bool wholeRecord)
...
+        "walk.max_nodes, until the set the walk reaches fits — the rows are what the walk reached, seeds included, " +
+        "so limit= and offset= window the render and not the walk. project.form='chain' lists the same reached set " +
+        "without reading a body, which is what to run first.";
```
Avick3110: **"without reading a body" is not what `form='chain'` costs.**

The walk reads a body for every reached node whatever the form: `WalkForwardBatch` calls `Fetch(key)` on each dequeued node to get its type, EditorID and outgoing links, and it does that before the form is ever consulted. `form='chain'` skips the *render's* second read of those same records, not the walk's own.

So the remedy — and the matching CHANGELOG line, "`chain` itself reads no record body and is not held to it" — tells a caller refused at, say, 400,000 reached rows that the shape to run first costs nothing, when it costs the ~400,000 body reads the walk was always going to make. Roughly half of what was refused, not approximately none. This sentence is the only cost guidance the refusal carries.

"lists the same reached set without reading a body **per rendered row**" would be true and would keep the recommendation intact.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in e6d06bac. Widened the gate rather than narrowing the sentence: `summary` and `aggregate` do read a body per row, so the walk lane is the same OOM shape for them. `bodyForm` is now `fields | rows | everything | summary | aggregate` — one named predicate beside `bodyFields`, since it is the set a derived selection's bound is measured over — and the walk lane gates on it. `identity` stays out: it returns above on `ResolveRefs` and reads no body. Test `EveryReadingFormOverTheReachedSetIsHeldToTheBound` is a Theory over summary and aggregate; both refuse before the fold and pass after.

Left the `formids=` lane's own bound (line 576) alone — its published sentence is `limit=`'s "a BODY form (fields/rows/everything)", which matches what it gates on, so widening it there would be a second change against a description this PR does not own.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in b3ba7107. Both reverse lanes now take the same bound before `return ListLane()`, each with its own remedy, because their levers genuinely differ:

- **Reverse carrier** (`RenderBudget.ReverseCarrierRemedy`): fewer `formids=` seeds, lower `walk.max_nodes` (per seed here), `types=` narrows the carrier types, and `project.form='chain'` as the shape to run first. `walk.depth` is deliberately absent — that walk reaches nothing past hop 1, so naming it would be the same defect as the one on the other thread.
- **Transitive reverse** (`RenderBudget.ReverseTransitiveRemedy`): fewer seeds, lower `walk.depth` or `walk.max_nodes` (one shared budget), and it says outright that `chain` is not a lever here because the walk expands one shared frontier and has no per-seed path to draw — the same reason line 392 refuses it.

`walk.max_nodes`'s own sentence was the other half of this, and it no longer promises `chain` on every lane: it now says the refusal names that lane's own levers — its seeds, this budget, and `project.form='chain'` on the forward and carrier walks, the two chain can draw. Tests `ATransitiveReverseWalkTooBigToRenderRefuses` and `AReverseCarrierWalkTooBigToRenderRefusesNamingTheChainForm` in `RecordsReverseIndexTests.cs` pin each lane's refusal and, negatively, the lever it must not name.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 1bf1f9d3 — your suggestion applied as written.

Test `AHopDoesNotSpendItsGatherOnNodesTheSeedAlreadyVisited` pins it. `WalkCostWorld` grows a revisit fan of FLSTs: 40 hubs, each pointing at every hub (itself included) and at one terminal of its own, plus a seed list of the 40 hubs. A closure walk from that seed at `depth=2`, `max_nodes=80` records the hubs at hop 1, so `room` is 40 at hop 2 and the hop-2 frontier is the 40 hubs — all revisits — ahead of the 40 terminals it can still record. Measured with the fix reverted: 40 per-record plugin walks, exactly the fall-through you describe; with it, 0. The answer is identical either way (`reached=80`), which is the point — only the cost differed.

The fan is FLST-only and reachable from nothing else, so the NPC-seeded tests' counts are untouched.

src/housecarl-mcp/RenderBudget.cs
Avick3110: Folded in f320702a. `WalkRemedy` now opens "narrow the seeds you passed, or lower walk.depth or walk.max_nodes …" — the parenthesis is gone, and it holds on both entries to `WalkLane`.

Test `AFormidsSeededWalkIsRefusedWithoutNamingTheScanTerms`: a `formids=`-seeded walk over the bound refuses naming the seeds and `walk.max_nodes`, and asserts the refusal contains no `types=`, `plugins=` or `where=`.

One structural note, since it touched the same line: #596's fold had added a `listSelection` bool to `Refuse` while this branch was adding a `remedy` string, and the rebase collided there. Resolved to the string — the `formids=` lane's wording moved out to `RenderBudget.ListRemedy` and the bool is gone, so each lane's lever now sits beside the others as a named constant. Every remedy carries its own leading verb in lower case and `Refuse` capitalises it when the sentence starts there, so all four published strings are byte-identical to what they were apart from this change.

src/housecarl-mcp/RenderBudget.cs
Avick3110: Folded in e5f53f97. `WalkRemedy` now reads "project.form='chain' lists the same reached set without reading a body per rendered row, which is what to run first", and the three other places that made the same claim were corrected with it:

- `walk.max_nodes`'s description: "… without reading a body per rendered row (the walk reads one per reached node whatever the form)".
- The CHANGELOG line: "`chain` reads no body per rendered row and is not held to the bound, though the walk itself still reads one per reached node whatever the form."
- The code comment at the walk-lane gate, which said chain renders "not the records' bodies" — now "it pays no SECOND body read on top of the walk's own".

The two new reverse remedies were written to the corrected wording from the start.

PR #603, opened 2026-09-06
Make max_chars a ceiling on the batch, skse and records renders

### LINKED ISSUE #546: [GAP] max_chars= is not a hard bound: BatchRender and the skse renders overshoot by the item that crossed the cap
Surfaced twice on 2026-09-05, by the reviews of PR #537 and PR #538.

`max_chars=` is not a hard bound on any render that tests the cap before appending an item: `BatchRender` (asset_status, nif_inspect, place) overshoots by the item that crossed the line plus its cut marker, and the `skse` family renders append scope, caveats and the filter hint after the cap test. Footers and accounting blocks are now priced inside the cap (#537, #538), but the item overshoot remains, so a caller that sizes a buffer to `max_chars` can still receive more.

Fix belongs in `BatchRender` once, with the item that would cross the line dropped and counted as truncated instead of emitted, and the same rule applied to the `skse` renders. Not a silent wrong answer (the accounting says what was cut); a contract note. After 2.0 unless a caller's buffer sizing bites.

Avick3110:
#596 has merged; this branch is rebased onto main, so the diff below is its own commits only (four, plus eight folding a blind review and eight more folding Aaron's).

`max_chars=` was a test a render took *before* writing the next item, not a bound on what came back: the record block, path, mesh or row that crossed the line was written whole, and the truncation notice, the accounting line, the `spilled:` block, the skse scope note/caveats/filter hint and place's whole trailer were appended on top of it — so a caller who sized a buffer to `max_chars` could receive several hundred characters more. #596's fold recorded exactly this on finding 7's thread. The layout is now the bound: everything a render writes below its rows is composed and charged before the first row is laid, an item that would cross what is left is taken back out whole and counted in the notice, and one item wider than the whole budget is named with the `max_chars` that clears it in one step rather than dropped — "wider than the whole budget" claimed only when the item would not have fitted an empty page either, and the remedy settled against its own rendered width so a cap with fewer digits than its answer does not send the caller round twice. Nothing is trimmed mid-token and no second cap was added. The one arm that can still exceed the cap is a `max_chars` too small for what a response carries whatever the budget — its header, the notices it owes, its accounting — and that now says so in one settled sentence naming the number that clears it, the same shape the check sweep's `max_chars_overrun` has carried since #537.

Closes #546

## What changed, per render

| render | what it did | what it does |
|---|---|---|
| `BatchRender` (asset_status, nif_inspect, place — text) | cap tested before the item; the crossing item written whole; the cut marker appended past the budget; the first item exempt from the cap entirely | the caller's trailer **and** the cut notice at its widest spelling are charged first; the item is written, measured, and taken back out whole if it crossed; `out int shown` replaces the caller-side counter that used to count an item the render then discarded |
| `BatchRender.AppendLines` / read-failure / discovery alarms | first line exempt; the block heading unconditional | the heading is unconditional again — it carries the count and the count IS the alarm, so it is never dropped to meet the cap (a heading that only fits sometimes hands back a clean-looking sweep over a build whose archive failed to read); whole detail lines inside the budget, the marker's room charged first, and a cap too small for the alarm is named by `Settle` |
| `nif_inspect` per-mesh block | inner sections cut against the caller's `max_chars`, so a mesh could fill the whole cap and then be appended to a buffer already holding the header and alarms | sections cut against the room the batch has left; a requested section with no room to start is counted in one line; the one-line block/shape lists charge their own cut notice |
| `place` | the counts line, the leftover note and the enable-and-sort block sat outside the cap **on purpose** | the widest spelling of all three is charged before the rows; all three still always render |
| `skse` inventory / pairing / config (text) | cap tested before each row; the scope note, caveats and filter hint appended unconditionally; each row list's own `showing N of M` notice appended without checking; the three families overwrote `cap` with the reduced budget and quoted THAT in their notices; `filter=` on the config family still wrote the crossing file block whole; pairing's tail was composed but never charged at all | the tail is composed and charged before the rows; every row list measures the row it is about to write AND holds back the widest spelling of its own cut notice, and a section starts only where that notice fits too; the two sections written whatever the rows cost — the plugin roster, the config-folder table — hold theirs with their headings; `cap` stays the caller's number and a separate `budget` carries the room, so every notice quotes what the caller passed; the config `filter=` view writes, measures and rolls back the whole file block, and rolls its `shownCfg` marks back with it; the family footer settles the fixed-part arm |
| `records` scan / batch / resolve / group (text) | cap tested before the row; the truncation notice, the accounting line and the `spilled:` block appended after it | the notice and the spill block are composed and charged first; the row that would cross is taken back out **with the owned-child clause it earned** (`ChildNotes.Mark`/`Restore`); the field loop measures the line it is about to write and still quotes the caller's `max_chars`, not the reduced budget |
| `records` summary (`project.form='summary'`) | the list lane's own render still carried the pre-fix shape: cap tested before the row, the crossing row written whole, the notice and the `spilled:` block appended on top, and no `Settle` on this lane at all | the notice and the spill block are charged first, the row that would cross is taken back out, and an overrun the fixed part causes is named |
| `records` header line | six forms did `headerLine + "\n" + Wire.Render…(max_chars)`, adding ~40 characters the ceiling never counted | the render takes the header and writes it inside the budget |
| `TransportAccounting` | a window that got no row onto the page still advised `re-call … offset=<the same offset>` | a window that rendered nothing offers no next page, and instead names the cap to raise for the row at this offset and the `offset=` that steps past it — where one row is wider than the whole budget, that offset is the only knob that moves |

The base branch's `rendered` / `render_ms` accounting stays inside the budget: `RenderBudget.AccountingReserve` is still charged, and the new charges are added to it rather than replacing it.

**New file:** `src/housecarl-mcp/RenderCap.cs` — the `(Cap, Budget)` pair, so the number a notice quotes and the number a unit is measured against cannot be confused, plus `Settle`, the fixed-part overrun sentence.

## Tests

`BatchRenderCapTests` extended in place (14 cases), `SkseTransportTests` and `RecordsArtifactTests` gained the same contract on their own worlds.

| test | before | after |
|---|---|---|
| `AnAssetStatusRenderFilledPastItsCapAnswersInsideIt` (3 caps) | would fail † — 400 contested paths return past the cap at every one | passes |
| `ANifInspectRenderFilledPastItsCapAnswersInsideIt` (3 caps) | would fail † — same, by one whole mesh block | passes |
| `APlaceRenderFilledPastItsCapAnswersInsideItAndStillCarriesItsTrailer` (2 caps) | would fail † — the trailer alone put it over | passes, trailer still present |
| `TheNoticeCountsExactlyWhatTheRenderHeldBack` | would fail † — `rendered` counted an item that was written past the cap | passes (rendered + omitted = the selection) |
| `AnItemWiderThanTheWholeBudgetIsNamedNotSilentlyDropped` | would fail † — no such sentence existed; the item was written past the cap instead | passes |
| `FollowingThatRemedyOnceRendersTheItem` | would fail † — no remedy to follow | passes (one step) |
| `ACapTooSmallForTheFixedPartSaysSoAndNamesTheCapThatClearsIt` | would fail † — the render answered over the cap in silence | passes, and the number it names equals its own length |
| `AReadFailureListIsBoundedToAndCutsWithTheNamedMarker` | would fail † — the alarm list ran past the budget by one line | passes |
| `BothBatchRendersCutWithTheSameMarker` | passes both — the marker's spelling is unchanged | passes |
| `FollowingThatRemedyOnceRendersTheItem` (now 4 caps: 900, 950, 990, 1200) | fails at 900/950/990 — the remedy was short by the digits it grew | passes at all four |
| `AlarmsThatFillTheBudgetAreNotBlamedOnTheFirstItem` | fails — an ordinary item was named as wider than the whole budget when the alarms had filled it | passes |
| `ACutReadFailureListNamesTheArchivesItHeldBack` | replaces the cap that no longer cut: a cap that DOES cut three failures, pinning the marker's spelling and that listed + held back is the whole list | passes |
| `AnAlarmTooWideForTheBudgetIsStillSaidAndTheOverrunIsNamed` | fails — the block vanished whole, with the answer comfortably inside the cap and nothing saying an archive had failed | passes |
| `ASummaryRenderIsNeverWiderThanItsCap` (3 caps) | passes at these caps before and after — the length pin on the lane | passes |
| `ASummaryTooSmallForWhatItMustCarrySaysSoAndNamesTheCapThatClearsIt` † | fails — the summary lane answered ~1,100 chars over `max_chars=900` in silence, never having called `Settle` | passes |
| `EachFamilysTextRenderFilledPastItsCapAnswersInsideIt` (9 cases: 3 skse families x 3 caps) | fails on all nine — 26 to 450 chars over | passes |
| `ACaveatBlockThatOutgrowsTheCapIsChargedBeforeTheRowsNotAppendedPastThem` | fails — the caveats landed past the ceiling | passes |
| `TheWholeDispatchedResponseFooterAndAllIsInsideMaxChars` (2 caps x 3 families) | fails — footer plus one row over | passes |
| `ACapTooSmallForTheFixedPartSaysSoInsteadOfOverrunningSilently` (skse) | fails — silent overrun | passes |
| `AScanRenderIsNeverWiderThanItsCap` (4 caps) | fails — the last record block plus the notice plus the spill block | passes |
| `AScanCutByItsCapSaysHowManyRecordsItHeldBack` | passes both — the notice's wording is unchanged | passes |
| `TheSpillStillCarriesTheCompleteResultWhenTheRenderIsHeldToItsCap` | fails — the render was over the cap; the manifest was already complete | passes, and the manifest still holds every row |
| `ACapTooSmallForTheSpillBlockSaysSoAndNamesTheCapThatClearsIt` | fails — ~900 chars over in silence | passes |

† marks a case NOT run against the pre-fix build. Fail-before, stated as observed rather than claimed wholesale. The nine skse cases and the four records cases were watched failing on the part-fixed tree and their messages carry the measured overshoot — 26 to 450 characters for the three skse families, and 15 characters for a records scan at `max_chars=2000` before the header line moved inside the render. The asset_status / nif_inspect / place ceiling cases were NOT run against the pre-fix build: their assertion is the length bound, and the pre-fix loop wrote the crossing item after the test by construction. What I did watch fail on the first build of this change is the three first-item tests they replace, which is the same fact from the other side. The cases asserting a sentence — the oversize notice, the overrun notice, the one-step remedy — assert wording that did not exist before this branch.

**Existing tests whose measurement moved.** `AssetSelectTests.TheAccountingBlockIsPricedInsideMaxCharsRatherThanAppendedPastIt` measured from `[accounting]` to the end of the response; the overrun sentence a too-small cap earns is written after the block and is not part of it, so the measurement now stops where it starts. The alarm heading being unconditional is what makes that sentence appear at `max_chars=400` on that world, and it is the documented arm.

**Existing tests whose cap moved.** Six tests drove caps that can no longer put a row on the page, because the row that used to be written past the cap now is not: `BatchRenderCapTests`' three first-item tests are replaced by the contract above; `PlaceRenderTests.ACutListStillNames…`, `AssetSelectTests.TheNextPageAdviceCarriesTheLimit…` and `…WalkingBySuccessiveNextPageOffsets…`, `RecordsOwnedChildTests.ACapThatTruncatesTheAnnotatedFieldAway…` and `RecordsRemedyRepairTests.TheEverythingFormsTruncationNotice…` keep their assertions at a cap that renders a row. The owned-child one also puts its annotated field last behind filler, so what it asks — the clause is earned by a line the caller can see — no longer depends on where the cut happens to land.

**Probes.** `nif-inspect-batch-guard` arm 4 gains `4c`, the ceiling itself, and its arm 6 (`FIRST MESH ALWAYS ANSWERS`) is deleted: that contract is what this PR changes, and its replacement — an item wider than the budget is named, and the remedy works in one step — lives in `BatchRenderCapTests`. `nif-service-guard`'s two section-cut arms moved to caps that still cut.

## Fold of Aaron's review

Seven findings, seven commits, plus one on the changelog entry they made stale.

| finding | verdict | commit |
|---|---|---|
| the nif section cut marker is charged nowhere, so a cut mesh could land one marker past the budget and be taken back out entire | folded — the marker's room is charged beside the missed-sections line; `AMeshWhoseSectionsCutIsStillRendered` fails on the old code at all three caps | `Charge the section cut marker so a cut mesh still renders` |
| `Needed`'s fixed point assumes a fixed-width item, so the one-step promise does not terminate where the item or the alarms grow with the budget | folded — the remedy is measured against the widest form: the header, the alarms and every item through the one that crossed, rendered against a budget none can exhaust. A sweep of 1,700 nif caps and 3,700 asset caps had 970 one-step failures before, none after | `Measure the oversize remedy against the widest form of the render` |
| `SlotNamingCaveat` appended with the `TryAppend` result discarded | folded — charged with the heading, so a section that cannot hold both does not start and is counted as missed; same at `RenderPaths` | `Write the slot-naming caveat with its heading or not at all` |
| `housecarl_skse` hands the renders `cap = max_chars - footer`, so its notices quote a number nobody typed | folded — the footer rides down as the render's trailer and the cap stays the caller's; `EveryNoticeQuotesTheMaxCharsTheCallerPassed` asserts one number per response over all three families | `Quote the caller's max_chars in every housecarl_skse notice` |
| the paging advice is dropped when `Rendered == 0`, which removes the only sentence naming an offset in the case where `offset=` is the only escape | folded — one sentence for the rendered-nothing case naming both the cap to raise and `offset=Offset+1`, gated on `Truncated > 0` so a spent limit is not blamed on max_chars | `Name the offset that moves a window that rendered nothing` |
| the oversize sentence is gated on `shown == 0`, so an oversize item after the first gets a remedy that cannot work | folded — the guard is dropped, the width test kept, the remainder counted, and the sentence now says "the next one" rather than "the first" | `Name an oversize item wherever it falls in the batch` |
| `line(e)` evaluated twice per row in the pairing render | folded — the row is built once, the shape the other two renders already use | `Compose each pairing row once instead of twice` |

Two second-order effects, both intended: the rendered-nothing sentence widens the accounting reserve, so three batch-render cap tests that pinned a single `max_chars` now sweep a range of caps instead of sitting on a number the reserve moved; and `FamilyCall` gained a `Trailer`, so the test double that stood in for `ServiceRenders` had to pass it down like the real one does.

## What was run

From inside `.claude/worktrees/max-chars`:

```
```

1,745 passed, 0 failed after the fold and the rebase onto the current main. `ci-all` now runs locally as well: 127/127 passed. The probes that could see this change were run by name and pass: `nif-inspect-batch-guard`, `nif-service-guard`, `asset-status-guard`, `place-asset-guard`, `render-clamp-guard`, `native-pairing-guard`, `skse-config-audit-guard`, `skse-peek-guard`, `nif-sections-guard`, `bulk-primitives-wave2-guard`, `bulk-primitives-wave3-guard`, `bulk-query-primitives-guard`, `owned-child-content-guard`, `effect-chain-guard`, `apply-guard`, `write-surface-guard`, `bulk-create-guard`, `conflict-diff-guard`, `loadorder-status-guard`, `source-chain-guard`, `source-display-guard`, `dialogue-info-order-guard`, `dialogue-validate-guard`, `upsert-guard`, `readback-count-guard`, `compact-readback-guard`, `epoch-guard`, `emit-match-guard`, `refusal-completeness-guard`, `depth-leak-guard`, `flag-bits-display-guard`, `poly-field-descend-guard`. `check-guard` and `remedy-verbs-guard` fail in this worktree only because `generated/corpus.json` has not been produced here — the failure is `FileNotFoundException: corpus.json`, a generated input this worktree has never produced, not an assertion about a render; CI generates it before running them. Nothing was run against the live MO2 instance.

## What this does NOT cover

`housecarl_apply`, `housecarl_create`, `housecarl_forward`, `housecarl_remove`, `housecarl_write_seq`, `housecarl_bsa_list`, `housecarl_load_order_status`, `housecarl_update_status` and `housecarl_skypatcher_layer` also take `max_chars=` and also answer past it — three of them (`remove`, `write_seq`, `apply`'s read-back) put their trailer outside the budget in a comment that says so deliberately, which is a call to make rather than a bug to fix here. The json lanes of the tools above were already bounded and are untouched. #546 names `BatchRender` and the skse renders, and the records lane is what finding 7's thread named; the rest is a separate issue if Aaron wants the same rule everywhere, and the changelog entry names only the five tools this PR actually holds.

Inside `housecarl_records` itself, five text renders still test the ceiling before the row and so can answer past it by the block that crossed: the comparison forms `delta` and `tree`, the walk lane's `chain` and reverse effect-chain renders, and `info_order`. They live in `RecordsTools.cs` and each has inner loops (deltas, nodes, declarers, the shared `AppendInfoOrderView` / `RenderEffectChain`) that would have to be charged too, which is a change of its own rather than a line in this one. The `max_chars=` parameter on the tool and the changelog entry both name them, so nothing claims a ceiling the code does not hold. Worth an issue.


Avick3110:
Moved onto the rebased #596 (`claude/records-cost` is now `64f15ecb`, itself rebased onto the merged description wave). Old tip `fa1388e8` → new tip `2aced128`; five commits, none squashed.

One file conflicted:

- `src/housecarl-mcp/NifTools.cs`, the `nif_inspect` `max_chars=` description — the wave had just moved "one cap over the WHOLE batch's render, not per mesh" out of the tool description onto this parameter, and this branch rewrites the same line to say what the ceiling now covers. Kept main's moved fact and this branch's sentences as one text: "TRANSPORT: character CEILING on the whole response — one cap over the WHOLE batch's render, not per mesh: the mesh whose block would cross it is not written at all, …". I kept the branch's `TRANSPORT:` opening rather than the file's older `Optional.`, so this parameter reads like the four sibling `max_chars=` descriptions this branch rewrites (`asset_status`, `place`, `records`, `skse`); say the word if you want the file's own spelling back instead.

`PlaceTools.cs`, `RecordsTools.cs`, `AssetTools.cs`, `SkseTools.cs` and `plugin/CHANGELOG.md` merged clean — the wave did not touch those `max_chars=` lines, so no description text was lost there. Render code is untouched by the resolution.

From the rebased worktree:

```
```

1,673 passed, 0 failed. `ci-all` not run locally; CI runs it.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -332,7 +334,21 @@ public static string Render(SkseInventoryData d, string? filter, int cap, RowWin
...
+                   "grouped by folder above. Non-config content (animation/mesh/etc.) is counted in the 'other file(s)' total.)\n" +
+                   Caveats(d) +
+                   "\n→ filter='<plugin/mod/DLL name>' for a plugin's full detail, or filter='<folder>' (e.g. SkyPatcher, OStim) to list a config group.";
+        // The two sections this view always writes below its rows — the plugin roster and the config-folder table —
+        // are headings the budget owes whatever the rows cost, so they are charged here with the tail.
+        string rosterHead = "\nplugins with metadata (" + Split(rows).Modern.Count + ") — name · version · compat · winning mod:\n";
+        int folderGroups = d.Configs.Select(e => e.Group).Distinct(StringComparer.OrdinalIgnoreCase).Count();
+        string foldersHead = d.Configs.Count == 0 ? ""
+            : "\nconfig folders (" + folderGroups + ") — folder: files ← provider(s):\n";
+        cap = Math.Max(1, cap - reserve - tail.Length - rosterHead.Length - foldersHead.Length
```
Avick3110: **The inventory render still answers past `max_chars`.** The reserve here charges `reserve + tail + rosterHead + foldersHead + SectionsMissed`, but not the cut notices the row lists themselves append. `AppendCapped` (L806) stops when the *next* row would cross `cap` and then appends `"  ... [showing N of M; raise max_chars…]"` **without checking**; the config-folder loop (L468) does the same. Three of those can fire in one render — one diagnostic subset, the plugin roster, the folder table — and none of them is charged.

Measured on the built branch, a layer with several non-empty diagnostic subsets (locked / legacy / non-plugin / unreadable / subfolder / contested) plus 300 configs in 60 folders:

| max_chars | returned | over |
|---|---|---|
| 1500 | 1576 | +76 |
| 2000 | 2074 | +74 |
| 3000 | 3074 | +74 |
| 5000 | 5076 | +76 |
| 8000 | 8074 | +74 |

(12000 passed.) `EachFamilysTextRenderFilledPastItsCapAnswersInsideIt` misses this because its `Inventory(300)` fixture builds 300 identical Modern/address-library DLLs, so every diagnostic subset is empty and only one unreserved notice ever fires — the accounting reserve's slack (`Reserve` composes with `everySentence: true`, which is always wider than the line actually written) absorbs it.

Through `Dispatch` the caller does not get a silent overrun — `RenderCap.Settle` catches it — but they get the *fixed-part* sentence, which misdiagnoses this: the fixed part fits; three unreserved cut notices are what pushed it over.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -1143,7 +1208,8 @@ string Accounting() => TransportAccounting.Compose(
         int shownFiles = 0;
         foreach (var f in hits)
         {
-            if (sb.Length >= cap) { sb.Append("\n  ... [showing ").Append(shownFiles).Append(" of ").Append(hits.Count).Append(" files; raise max_chars]\n"); break; }
+            int mark = sb.Length;
```
Avick3110, line 1289: **This arm was renamed but never rewired — `mark` is dead and the block that crosses the cap is still written whole.** The line below still tests *before* the file block (`if (mark > cap)`), and `mark` is never read again, so `housecarl_skse family=config filter=…` keeps exactly the behaviour #546 is about. Every other loop in this PR moved the test after the unit and rolls back with `sb.Length = mark` (L521, L543, L1690).

Measured on the built branch (60 files, 12 refs each, `filter='Group'`):

| max_chars | returned | over |
|---|---|---|
| 800 | 1415 | +615 |
| 1500 | 2454 | +954 |
| 3000 | 3493 | +493 |

The caller then gets the `Settle` overrun notice blaming the fixed part, when what actually overran is one file block plus its reference lines. This view also charges only `reserve + tail`, so it has no room reserved for its own `"showing N of M files"` notice either.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -743,6 +762,20 @@ static string CompatTag(SksePluginReader.SkseVersionInfo v)
...
+    internal static bool Head(StringBuilder sb, int cap, string head)
+    {
+        if (sb.Length + head.Length > cap) return false;
+        sb.Append(head);
+        return true;
+    }
+
+    /// <summary>The line that says how many sections the budget could not start. Spelled once so its room can be
+    /// charged before the sections render.</summary>
+    internal static string SectionsMissed(int missed, int cap) =>
```
Avick3110, line 818: **This notice names a `max_chars` the caller never passed.** All three skse families overwrite `cap` with the reduced budget (`cap = Math.Max(1, cap - reserve - tail…)`, L349 / L1029 / L1529) and then hand that same variable to `SectionsMissed` at L474 / L1165 / L1619, and to the `"omitted at max_chars=" + cap` notice in `RenderDetail`.

Observed output for a caller who passed `max_chars=1500`:

```
  ... [5 section(s) omitted at max_chars=1082; raise max_chars to see them]
```

and at `max_chars=5000`: `omitted at max_chars=4082`. A caller who trusts the number and raises to just above it lands below their previous cap. The reduction was smaller before this PR (`reserve` only); adding `tail` + the two headings makes the quoted number 25–30% off.

This is the exact confusion `RenderCap(Cap, Budget)` was introduced to prevent — the skse families are the ones that did not adopt it. Keeping the caller's `max_chars` in a separate local and quoting *that* in every notice would settle all of them.

src/housecarl-mcp/BatchRender.cs
```diff
@@ -3,87 +3,128 @@
...
-            }
-            appendItem(sb, item);
-            shown++;
+            int mark = sb.Length;
+            bool roomBefore = mark <= budget.Budget;
+            if (roomBefore) appendItem(sb, items[i], budget);
+            if (roomBefore && sb.Length <= budget.Budget) { shown++; continue; }
+
+            // Whole items only: the one that crossed is taken back out, and what it would have needed is named.
+            int needed = cap + (sb.Length - budget.Budget);
```
Avick3110: **`needed` is short whenever the remedy crosses a decimal-digit boundary, so "clears it in one step" is not true.**

`needed = cap + (sb.Length - budget.Budget)` = `itemEnd + reserve + NoticeReserve(cap)`. But `NoticeReserve` embeds `cap`'s own digits (both `Cut` and `Oversize` print `max_chars=<cap>`), so re-rendering at `cap' = needed` recomputes a reserve that is `digits(needed) - digits(cap)` characters *wider*. The item then misses the budget by exactly that many characters and is taken back out again.

Measured on the built branch, `AssetWire.Render(Paths(3, providers: 60), cap)`:

| max_chars | notice says | re-render at that value |
|---|---|---|
| 900 | 4842 | still omitted |
| 950 | 4842 | still omitted |
| 990 | 4842 | still omitted |
| 1200 | 4842 | renders ✓ |

`FollowingThatRemedyOnceRendersTheItem` only drives 1200, where `cap` and `needed` are both 4 digits, so the growth is zero. Any cap with fewer digits than the answer (a plain `max_chars=900`, or `500`, or `80`) sends the caller round twice.

Charging `NoticeReserve` at a fixed width — or adding `digits(needed) - digits(cap)` to `needed` — closes it.

src/housecarl-mcp/BatchRender.cs
```diff
@@ -3,87 +3,128 @@
...
+    public static void AppendReadFailures(StringBuilder sb, IReadOnlyList<string> failures, string subjectPhrase, RenderCap cap)
     {
         if (failures.Count == 0) return;
-        sb.Append("\n[!] ").Append(failures.Count).Append(" archive(s) could NOT be read this build — ")
-          .Append(subjectPhrase).Append(" present only in these may read as ABSENT below:\n");
+        // The heading carries the count, so it goes in whole or not at all: a heading with no lines under it still
+        // says how many archives failed, which is the alarm.
+        var head = new StringBuilder("\n[!] ").Append(failures.Count).Append(" archive(s) could NOT be read this build — ")
+            .Append(subjectPhrase).Append(" present only in these may read as ABSENT below:\n").ToString();
+        if (!cap.TryAppend(sb, head)) return;
```
Avick3110: **The archive-read-failure alarm can now vanish with nothing said, and the ABSENT rows below it go unhedged.** The heading used to be unconditional; now a heading that does not fit the budget returns early and the whole block — heading *and* lines — is gone. Nothing counts it: `AppendCut` is never reached, and `RenderCap.Settle` does not fire because the response is comfortably *inside* `max_chars`.

Measured on the built branch, `AssetWire.Render` over one path and two BSA read failures:

| max_chars | returned | contains "could NOT be read this build" |
|---|---|---|
| 200 | 533 | no |
| 300 | 533 | no |
| 400 | 310 | no |
| 500 | 310 | no |
| 700 | 310 | no |

At `max_chars=500` the answer is 310 characters long — 190 characters of headroom — and the caller reads `ABSENT` / provider rows with no sign that an archive failed to read. That is the class of answer the file's own doc comment ("so a long batch cannot truncate them away") and the project's "never a silently degraded mode" rule exist to prevent, and it is a regression from the pre-PR behaviour.

`AppendDiscoveryWarnings` (L112) and `AssetWire.AppendSelectorNotes` have the identical shape. The alarms are the one thing that should outrank the rows: charge their room with the trailer, or fall back to a one-line `[!] N archive read failure(s) — raise max_chars` that always fits.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -231,11 +231,14 @@ public static string Render(PlaceOutcome o, int cap, IReadOnlySet<string>? poleW
...
             .Append(modFolder is null ? "" : $"\nmod folder: {modFolder}").ToString();
 
-        int rendered = 0;
+        // Everything below the rows — the counts line, the leftover note, the enable-and-sort block — is charged
+        // before the first row is laid, so a filled render answers inside max_chars instead of past it.
         var body = BatchRender.Render(
             header, o.Results, "asset(s)", cap,
-            sb => { foreach (var w in o.Warnings) sb.Append("[!] discovery: ").Append(w).Append('\n'); },
-            (sb, r) => { rendered++; AppendResult(sb, r, modFolder, poleWithheld?.Contains(r.AssetPath) == true); });
+            (sb, room) => { foreach (var w in o.Warnings) room.TryAppend(sb, "[!] discovery: " + w + "\n"); },
```
Avick3110: **`TryAppend`'s result is dropped inside a loop, so discovery warnings can go missing silently — and out of order.** The loop keeps going after a warning does not fit, so a long warning #2 is skipped while a short warning #3 still lands: the caller reads a contiguous-looking list that is missing an entry, with no count and no cut marker. Before this PR every warning was written.

Same in `NifWire`'s `unknownSections` block (single `TryAppend`, result ignored): the "unrecognized section(s) ignored" line can disappear, and the caller believes their `sections=` was honoured.

Either `break` on the first failure and append a `"N discovery warning(s) omitted"` line, or route these through `BatchRender.AppendLines`, which already reserves room for its own cut marker.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -138,7 +138,7 @@ public static string Records(
...
             int offset = 0,
-        [Description("TRANSPORT: character ceiling on the RENDER. Never truncates the result — an over-ceiling result SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, the row schema, and the epoch) and the response names the file. 0 = the server default (~80k).")]
+        [Description("TRANSPORT: character CEILING on the RENDER, and a hard one — the record block that would cross it is not written, and the truncation notice, the accounting line and the spilled: block are charged before the rows are laid. Never truncates the RESULT: an over-ceiling result SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, the row schema, and the epoch) and the response names the file, so what the ceiling held back inline is in the file. 0 = the server default (~80k).")]
```
Avick3110: **This says "a hard one" on a tool that still has an unbounded text form: `project.form='summary'`.** `RenderRecordsSummary` (RecordsTools.cs:2107) was not touched by this PR and still carries the pre-#546 shape verbatim — `if (sb.Length >= cap)` is tested *before* the row (L2125), so the row that crosses is written whole; the `"... [rendered N of M at max_chars=...]"` notice is appended on top of it; and `Artifacts.AppendSpillStateText` (L2145) adds the whole `spilled:` block after that, uncharged. A `housecarl_records ... project={form:'summary'} max_chars=2000` call over a large selection comes back over 2000 with no `Settle` sentence, because this lane never calls `RenderCap.Settle`.

That makes both this parameter description and the CHANGELOG line ("`max_chars=` is now a hard ceiling on `housecarl_records` … The one case that can still come back over the cap is a `max_chars` too small for what a response carries whatever the budget") wrong for one of the tool's own project forms, and there is no test on the summary lane's length. Either hold `RenderRecordsSummary` to the same rule as the other four record renders, or name the summary form as out of scope here and in the changelog so the claim matches what the code does.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -502,8 +518,9 @@ string Accounting() => TransportAccounting.Compose(
         var shownCfg = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         foreach (var e in dllHits)
         {
-            if (sb.Length >= cap) { sb.Append("\n  ... [remaining DLL matches omitted at max_chars=").Append(cap).Append("]\n"); break; }
+            int mark = sb.Length;
             AppendDetail(sb, e, d, shownCfg);
```
Avick3110: **The rollback here takes the text back out but leaves `shownCfg` written, so configs the response never showed are both hidden and counted as rendered.** `AppendDetail` adds every config it pairs to this DLL into `shownCfg`; the next line rolls the buffer back to `mark` but nothing rolls `shownCfg` back. Two consequences downstream:

- L534 `rest = cfgHits.Where(e => !shownCfg.Contains(e.RelPath))` — those configs are now filtered out of the "matching configs" list as if they had already been shown. They appear nowhere in the response and no notice counts them.
- L555 `foreach (var e in cfgHits) if (shownCfg.Contains(e.RelPath)) tally.Mark(e.RelPath);` — they are marked as rendered rows, so the accounting reports `rendered=` higher and `truncated=` lower than what is on the page.

That is the same defect `ChildNotes.Mark`/`Restore` was added for in the records lane, and it contradicts this PR's own rule that `shown` agrees with what the reader can see. Snapshot the set (or its count) before `AppendDetail` and restore it alongside `sb.Length = mark`.

src/housecarl-mcp/BatchRender.cs
```diff
@@ -3,87 +3,128 @@
...
+            int mark = sb.Length;
+            bool roomBefore = mark <= budget.Budget;
+            if (roomBefore) appendItem(sb, items[i], budget);
+            if (roomBefore && sb.Length <= budget.Budget) { shown++; continue; }
+
+            // Whole items only: the one that crossed is taken back out, and what it would have needed is named.
+            int needed = cap + (sb.Length - budget.Budget);
+            sb.Length = mark;
+            sb.Append('\n');
+            if (shown == 0 && roomBefore) sb.Append(Oversize(items.Count, itemNoun, cap, needed));
```
Avick3110: **`shown == 0` is not the same fact as "this item is wider than the whole budget", so this sentence can name the wrong cause.** The alarms block above ran against the same `budget` and can legitimately fill it to the last character (`AppendLines` appends whole lines up to `room` and then its own cut marker). When it does, `mark <= budget.Budget` still holds, so `roomBefore` is true, the first ordinary-sized item crosses, `shown` is still 0 — and the caller is told "the first alone is wider than this response's whole budget", which is false: the alarms were.

Concretely: an `asset_status` build with many BSA read failures or many `under=` selector notes at a modest `max_chars` yields that sentence for a two-line path block. The remedy number is still computed off `sb.Length`, so it is not wildly wrong, but the diagnosis is, and it points the caller away from the real reason their render is empty.

Distinguishing the two is cheap — compare `mark` against the header length, or record whether the alarms block was itself cut — and the alternative wording already exists (`AppendCut`).

src/housecarl-mcp-tests/BatchRenderCapTests.cs
```diff
@@ -45,53 +49,158 @@ public class BatchRenderCapTests
...
+    // ---- the alarm lists ------------------------------------------------------------------------------
+
+    /// <summary>The alarm block above the items is bounded the same way, whole lines only, and counts what it
+    /// dropped.</summary>
     [Fact]
-    public void AReadFailureListUnderATightCapStillRendersItsFirstEntry()
+    public void AReadFailureListIsBoundedToAndCutsWithTheNamedMarker()
     {
-        var text = AssetWire.Render(ThreeReadFailures(), AlarmCap);
+        var text = AssetWire.Render(ThreeReadFailures(), 1_500);
```
Avick3110, line 254: **This test is named for a cut it no longer makes, and the claim the old one carried is gone.** `ThreeReadFailures()` is three short lines; at `max_chars=1500` the alarm block fits with room to spare, so no `archive(s) omitted at max_chars=` marker is produced and none is asserted — "…AndCutsWithTheNamedMarker" is not exercised by anything here or elsewhere in the file, and the alarm list's own cut marker is now uncovered.

The replaced test (`AReadFailureListUnderATightCapStillRendersItsFirstEntry`, cap 60) pinned two things: the marker's exact spelling *and* that a tight cap still names one failed archive. The first contract has no replacement; the second is what changed in `AppendReadFailures`, and the change is now unpinned in the other direction too — nothing asserts what a tight cap does to the alarm block. Either drive a cap that actually cuts three failures, or rename to what is being asked.

src/housecarl-mcp/PlaceTools.cs
Avick3110, line 1: **Two doc comments now describe the behaviour this PR removed.**

- `PlaceWire`'s class summary (PlaceTools.cs:225-229): "…then the §2.1 accounting and the explicit … instruction. **Those last two sit OUTSIDE the cap on purpose**". They are charged inside the cap now — `TrailerReserve` exists for exactly that — and the `max_chars=` description a few lines up was updated to say so, so the two disagree on the same file.
- `SkseTools.Dispatch`'s summary (SkseTools.cs:128-133): "…**the renders themselves test the cap BEFORE each list item and then append their scope note, caveats and filter hint unconditionally, so a response can still overrun a tight cap by one item plus that tail**". That is a description of the bug this PR fixes, sitting directly above the new `RenderCap.Settle` call.

Both are the kind of comment the next reader will trust over the code.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded in b0b40094: every row list now holds back the widest spelling of its own cut notice before laying a row, a section starts only where that notice fits too, and the roster and folder table hold theirs with their headings, so the three that could fire in one render are all charged.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded in 277f06f8: the file block is written, measured and rolled back entire like the other renders, and this view's own cut notice is charged before the first block.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded in b0b40094: cap is left alone as the caller's number and a separate budget carries the room, so SectionsMissed, the DLL-matches notice and every showing-N-of-M line quote what the caller passed.

src/housecarl-mcp/BatchRender.cs
Avick3110: Folded in 2f43aebc: the remedy is settled against its own rendered width, and FollowingThatRemedyOnceRendersTheItem now drives 900, 950 and 990 as well as 1200.

src/housecarl-mcp/BatchRender.cs
Avick3110: Folded in d9bcc19a: the alarm heading is unconditional again on all four blocks, only the detail lines are cut, and a cap too small for it is named by Settle rather than met by dropping the alarm.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in 0c7ac5ef for place, where the warning list stops at the first warning that does not fit and counts the rest with a line whose room is charged first, and in d9bcc19a for the unrecognized-sections line, which is written whatever the budget.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 40f0a8c1, which brings RenderRecordsSummary under the same shape with tests, and in 55336adb, which scopes the parameter text and the changelog to the renders that hold the ceiling and names the five that do not: delta, tree, the walk lane's chain and effect-chain renders, and info_order.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded in b0b40094: AppendDetail records the configs it newly marked and the rollback removes them alongside sb.Length = mark, so a config the reader never saw is neither filtered out of the list below nor counted as rendered.

src/housecarl-mcp/BatchRender.cs
Avick3110: Folded in 2f43aebc: the oversize sentence is claimed only when the item would not have fitted an empty page either, otherwise the cut marker is what the caller reads, pinned by AlarmsThatFillTheBudgetAreNotBlamedOnTheFirstItem.

src/housecarl-mcp-tests/BatchRenderCapTests.cs
Avick3110: Folded in d9bcc19a: ACutReadFailureListNamesTheArchivesItHeldBack drives a cap that does cut three failures, pinning the marker's spelling and that listed plus held back is the whole list, and AnAlarmTooWideForTheBudgetIsStillSaidAndTheOverrunIsNamed covers the tight-cap end.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in 9baf15b8: both comments now say what the code does.

src/housecarl-mcp/NifTools.cs
```diff
@@ -708,27 +737,31 @@ static string BitList(uint f)
...
         sb.Append('\n');
     }
 
-    /// <summary>True (and appends the cut notice) when the buffer has hit the cap mid-section — the per-item loop breaks.</summary>
-    static bool Cut(StringBuilder sb, int cap, int remaining)
+    /// <summary>True (and appends the cut notice) when the unit written since <paramref name="mark"/> crossed the
+    /// budget — the unit is taken back out whole and the per-item loop breaks. The test runs AFTER the unit is
+    /// written because that is the only point its size is known; a test before it knows only what the buffer
+    /// already holds.</summary>
+    static bool Cut(StringBuilder sb, RenderCap cap, int mark, int remaining)
```
Avick3110, line 764: **The section cut marker is the one notice this PR does not charge, and it costs the whole mesh.**

`Cut` rolls the crossing unit back to `mark` (which is `<= cap.Budget`) and then appends `BatchRender.AppendCut(...)` (~62-68 chars) with no room reserved for it. `AppendMesh` only reserves `SectionsMissed(want.Count, ...)` (~85 chars) at line 444, so after a section cuts, `sb.Length` can sit at `budget.Budget - 85 + 65`. If a later requested section then cannot start (`missed > 0`), the ~85-char `SectionsMissed` line at line 459 pushes the mesh past `budget.Budget`, and `BatchRender.Render` takes the **entire mesh block back out**.

Scenario: `housecarl_nif_inspect` on one mesh, `sections=shapes,nodes,strings`, `max_chars=2000`. The shapes list fills the room and cuts (notice appended, +65); `nodes` and `strings` no longer fit their headings (missed=2); `SectionsMissed(2, 2000)` is appended; the mesh now exceeds `budget.Budget`, so `BatchRender` discards it and answers `1 mesh(es) omitted at max_chars=2000: the first alone is wider than this response's whole budget`. Before this PR the caller got the mesh with cut sections; now they get the path line and nothing else — no version, no providers, no resolution.

Charging the marker alongside `SectionsMissed` in the `room` computed at line 444 (`cap.Less(SectionsMissed(...).Length + Cut-marker width)`) makes the mesh fit where the sections were cut to make it fit.

src/housecarl-mcp/BatchRender.cs
```diff
@@ -3,87 +3,148 @@
...
+            else AppendCut(sb, items.Count - shown, itemNoun, cap);
+            break;
         }
         return sb.ToString().TrimEnd('\n');
     }
 
+    /// <summary>The max_chars that clears this item in ONE step. The notice quoting it prints the cap as well as the
+    /// remedy, so a wider answer widens the reserve the item then has to clear: the number is settled against its own
+    /// rendered width rather than named short by the digits it grew.</summary>
+    static int Needed(int floor, int count, string itemNoun, int cap)
```
Avick3110, line 86: **`Needed` settles against the digits it grows, but not against an item whose rendered width grows with the budget — on `nif_inspect` the promised one-step remedy never terminates.**

The fixed point here assumes `floor` (header + alarms + item + reserve) is constant as `cap` rises. Two of the three callers break that assumption:

1. **The item.** `NifWire`'s `appendItem` is `AppendMesh(sb, r, want, room, ...)`, and a mesh with `sections=` expands to fill whatever budget it is given. Raise `max_chars` to the named number and the mesh renders more shape/node lines, so it still does not fit and the same sentence comes back with a slightly larger number. Following the remedy is an unbounded loop, not one step.
2. **The alarms.** `mark` is measured after `appendAlarms` ran against the *current* budget. When the alarm block was itself cut (hundreds of `under=` selector notes on `asset_status`, or a long BSA-failure list), a larger cap renders more alarm lines, pushing `mark` up by roughly what the item gained — again short.

The oversize sentence is the one place the render promises a number that works; where the item's width is budget-dependent it would be more honest to name the cap that fits the item's *maximum* form, or to drop the "in one step" claim for that lane.

`FollowingThatRemedyOnceRendersTheItem` passes because its fixture items are fixed-width; neither the nif-with-sections case nor the cut-alarms case is covered.

src/housecarl-mcp/NifTools.cs
```diff
@@ -462,14 +475,14 @@ static void AppendProviders(StringBuilder sb, IReadOnlyList<NifProvider> provide
         sb.Append('\n');
     }
 
-    static void RenderShapesDetail(StringBuilder sb, NifInspect nif, int cap)
+    static bool RenderShapesDetail(StringBuilder sb, NifInspect nif, RenderCap cap)
     {
-        sb.Append("\n--- shapes (").Append(nif.Shapes.Count).Append(") ---\n");
-        if (SlotNamingCaveat(nif) is { } shapesCaveat) sb.Append(shapesCaveat);
+        if (!cap.TryAppend(sb, "\n--- shapes (" + nif.Shapes.Count + ") ---\n")) return false;
+        if (SlotNamingCaveat(nif) is { } shapesCaveat) cap.TryAppend(sb, shapesCaveat);
```
Avick3110: **A caveat that changes how the rows below it read is now dropped silently when it does not fit.**

The section heading is gated (`if (!cap.TryAppend(...)) return false`), so a section that cannot start is counted by `SectionsMissed`. But the `SlotNamingCaveat` right after it is appended with a bare `cap.TryAppend(...)` whose result is discarded: when the heading fits and the caveat does not, the section renders its shape rows **without** the caveat and nothing says so.

That caveat is exactly the line that stops a misreading: without it a bare `tex[2]:` reads as "this Skyrim shader does not determine slot 2" instead of "this layout is not modelled at all" — the ambiguity its own doc comment says it exists to resolve. Same shape at line 520 in `RenderPaths`.

This also contradicts the rule the rest of the PR holds to (`BatchRender.AppendReadFailures`, `AppendDiscoveryWarnings`, `AppendSelectorNotes` all write their heading whatever the budget, and the changelog says "an alarm is never dropped to meet the ceiling"). Either charge the caveat with the heading — `TryAppend(head + caveat)` as one unit — or refuse to start the section when the caveat does not fit.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -145,7 +145,9 @@ internal static string Dispatch(IFamilyRenders renders, SkseFamily family, strin
             SkseFamily.Pairing => renders.Pairing(call),
             _ => renders.Config(call),
         };
-        return body + footer;
+        // The one arm a bounded render may still exceed on — a cap too small for what the response carries whatever
+        // the budget — is named rather than left for the caller to discover by measuring.
+        return RenderCap.Settle(body + footer, max_chars > 0 ? max_chars : 80_000);
```
Avick3110, line 154: **One `housecarl_skse` response can now quote two different "max_chars" numbers, and the one the notices quote is not the caller's.**

Line 140 hands the renders `cap = max_chars - footer.Length`, and `FamilyFooter` is ~190 chars. Every notice inside the family renders takes that reduced number as "the number the caller passed": `Showing(...)` prints `at max_chars=<cap>`, `SectionsMissed(missed, cap)`, `dllCut`'s `omitted at max_chars=<cap>`, `AppendHits`' cut lines. `Settle` on this line, added by this PR, quotes the real `max_chars`.

So `housecarl_skse(findings='config', max_chars=2000)` that both cuts a list and overruns comes back saying `... omitted at max_chars=1810 ...` and `over the max_chars=2000 it was given`. 1810 is a number the caller never typed, and the whole point of the `(Cap, Budget)` split introduced in `RenderCap.cs` — with comments at lines 336, 513, 1104 and 1596 all promising "cap stays the CALLER's max_chars, the number every notice quotes" — is defeated one frame above them.

The footer is a trailer like any other: pass the caller's `max_chars` down as `cap` and charge `footer.Length` as the reserve inside each family render (each already computes a `budget = cap - reserve - tail...`), rather than pre-reducing `cap`.

src/housecarl-mcp/TransportAccounting.cs
```diff
@@ -126,7 +126,9 @@ internal static string Compose(TransportCounts c, string rowNoun, bool everySent
         // Only what is still AHEAD of what was rendered earns a next page, and the next offset starts at the first
         // row this response did not show — so a caller following the advice sees every row exactly once. The advice
         // carries limit= as well: without it the next call resolves the whole remainder instead of one page.
-        if (everySentence || c.Remaining > 0)
+        // A window that got NO row onto the page has no next page to name: the advice would send the caller back to
+        // the offset they just used, and the max_chars sentence below is the remedy that actually moves them.
+        if (everySentence || (c.Remaining > 0 && c.Rendered > 0))
```
Avick3110, line 131: **A window blocked by one oversized row now gets no advice at all that moves it.**

Dropping the advice when `Rendered == 0` is right for the case it was written for (the advice used to point back at the offset the caller just used). But it also removes the only sentence that names an offset in the case where `offset=` is the *only* escape: one row wider than the whole budget.

Scenario: `housecarl_skse(findings='config', filter='<folder>', limit=5, offset=0)` where the first matching config declares thousands of references. `RenderFiltered` writes the block, measures it, rolls it back (`SkseTools.cs:1298`), and answers with `showing 0 of 5 files`. `Rendered = 0`, `Remaining > 0`, so the block now says only `max_chars cut 5 file(s) from the render: raise max_chars, or page with limit=/offset=` — with no offset. Raising `max_chars` does not help (the caller cannot know how far), lowering `limit` does not help, and nothing tells them `offset=1` shows the rest of their filter's matches. The rows are silently unreachable through the accounting block's own advice.

Worth naming `offset=Offset + 1` (skip the row that did not fit) rather than suppressing the sentence, or adding one sentence for the rendered-nothing case that says which knob actually moves.

src/housecarl-mcp/BatchRender.cs
```diff
@@ -3,87 +3,148 @@
...
+            if (roomBefore && sb.Length <= budget.Budget) { shown++; continue; }
+
+            // Whole items only: the one that crossed is taken back out, and what it would have needed is named.
+            int itemLength = sb.Length - mark;
+            sb.Length = mark;
+            sb.Append('\n');
+            // "Wider than the whole budget" is a claim about the item, so it is made only when the item would not
+            // have fitted an empty page either. When the alarms above it are what filled the budget, the item is
+            // ordinary and the cut marker is the honest line.
+            if (shown == 0 && roomBefore && headerEnd + itemLength > budget.Budget)
```
Avick3110: **The oversize sentence is gated on `shown == 0`, so an item wider than the whole budget anywhere but first gets the remedy that cannot work.**

When items 1..k fit and item k+1 is wider than the budget could ever make room for, `shown > 0` and the render falls to `AppendCut`, which says `raise max_chars to see all`. The caller raises `max_chars`; items 1..k grow to fill it and the same oversize item is again the one that crosses; same advice, forever. The `Oversize` branch exists precisely to break that loop, and it is exactly as true of item k+1 as of item 1 — the `shown == 0` guard is about not *blaming* an ordinary item when the alarms filled the page, which `headerEnd + itemLength > budget.Budget` already tests on its own.

Dropping `shown == 0` (keeping the `headerEnd + itemLength` test, and counting the remainder rather than `items.Count` in the sentence) covers the general case; the alarms-filled-the-page arm `AlarmsThatFillTheBudgetAreNotBlamedOnTheFirstItem` pins stays covered by the width test.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -1618,26 +1800,28 @@ string Accounting() => TransportAccounting.Compose(
...
     static void AppendCapped<T>(StringBuilder sb, IReadOnlyList<T> items, int cap, Func<T, string> line,
                                 RowTally? tally = null, Func<T, string>? key = null)
     {
+        // The cut notice is charged before the first row, so a list that cuts says so inside max_chars.
+        int room = cap - SkseInventoryWire.CutRoom(items.Count, hint: SkseInventoryWire.FilterHint);
         int shown = 0;
         foreach (var e in items)
         {
-            if (sb.Length >= cap) { sb.Append("  ... [showing ").Append(shown).Append(" of ").Append(items.Count).Append("; raise max_chars or use filter= to see all]\n"); break; }
+            if (sb.Length + line(e).Length + 1 > room) { sb.Append(SkseInventoryWire.Showing(shown, items.Count, hint: SkseInventoryWire.FilterHint)); break; }
```
Avick3110: `line(e)` is now evaluated twice per row — once to measure, once to write. For `dead`, `verify` and `debugBuilds` the delegate is `DeadLine`, which allocates a `StringBuilder` and walks `c.PairedDlls` formatting a fate line per candidate DLL, so a pairing audit over a large layer does that work twice for every row it renders.

`var row = line(e) + "\n";` then measure and append `row` — the shape `SkseInventoryWire.AppendCapped` and `SkseConfigAuditWire.AppendHits` already use two frames away — keeps it to one call.

src/housecarl-mcp/NifTools.cs
Avick3110: Folded in ec84d6a4. Real, and the drop was the whole mesh: the room at `AppendMesh` charged only `SectionsMissed`, so a section that cut left `sb` at up to `budget.Budget + 65` and the batch took the mesh back out entire. The marker's widest spelling is now charged beside the missed-sections line (`BatchRender.CutReserve`). Swept over caps 700..12,000 on a mesh with `sections=shapes,nodes,strings`: the mesh was dropped at 5,823 caps before and at none above the cap where its fixed part genuinely does not fit. `AMeshWhoseSectionsCutIsStillRendered` (2,000 / 3,000 / 4,500) fails on the old code at all three.

src/housecarl-mcp/BatchRender.cs
Avick3110: Folded in 6800e5fa, taking the first of your two options: the remedy names the cap that fits the item's maximum form rather than dropping the one-step claim.

`Widest` renders the header, the alarms and every item through the one that crossed against a budget none of them can exhaust, and that is the floor `Needed` settles the digits on. So the number covers the item's widest form, the alarm block's widest form, and — after the `shown == 0` guard came off next door — the items already on the page, which also grow. It runs only on the path that has already given up on the response, so it costs one extra pass exactly when the answer was going to be a refusal sentence anyway.

Measured: sweeping the nif fixture over caps 300..2,000 and an asset one over 300..3,000, 970 caps named a number that did not clear the item in one step before; none after. `TheRemedyHoldsWhereTheRenderedWidthGrowsWithTheBudget` and `TheRemedyCountsTheAlarmLinesAWiderCapWouldRender` sweep those ranges and check the second render for both lanes.

One thing the fix trades: where the alarm block was itself cut, the named number is now large enough to render every alarm line too, which is more than the item strictly needs. It is an "at least" claim and it works in one step, which insufficiency did not.

src/housecarl-mcp/NifTools.cs
Avick3110: Folded in db7621ed, charging the caveat with the heading as one unit — a section that cannot hold both does not start and is counted by `SectionsMissed`, the same rule the rest of the PR holds to. Same at `RenderPaths`. `ASlotSectionRendersItsCaveatOrDoesNotStart` renders a mesh whose shaders read as the FO4 layout and asserts the caveat wherever either heading appears; on the old code 245 caps in 400..3,000 rendered the shapes section without it, starting at 924.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded in 17017a1e. `FamilyCall` gained a `Trailer`, `Dispatch` passes the caller's `max_chars` as `Cap` and the footer's length as that trailer, and each family render subtracts it inside its own `budget = cap - trailer - reserve - …`. So `Showing`, `SectionsMissed`, `dllCut` and `AppendHits` quote the number the caller typed, and `Settle` quotes the same one. `EveryNoticeQuotesTheMaxCharsTheCallerPassed` regexes every `max_chars=<n>` out of a dispatched response for all three families at two caps and asserts they are all the caller's. `TheFooterLengthIsSubtractedFromTheCapHandedToTheRender` pinned the old contract, so it is replaced by `TheFooterIsChargedAsTheRendersTrailerNotTakenOffItsCap`; `StubRenders` in the transport tests had to pass the trailer down like `ServiceRenders` does.

src/housecarl-mcp/TransportAccounting.cs
Avick3110: Folded in fe6a76ca, as the second option: one sentence for the rendered-nothing case rather than an offset spliced into the next-page advice. It names both knobs — raise `max_chars` for the row at this offset, or skip it with `offset=<Offset+1>` — and says "skip", since that row is one the caller has not seen. Gated on `Truncated > 0` as well as `Rendered == 0`, so a window that got no rows because its limit was already spent (the inventory family's second list) is not told max_chars did it. The reserve still bounds it: `Widest` composes every sentence, and it derives `Offset + 1` from the same offset, so the digits match exactly. `AWindowThatRenderedNothingNamesTheOffsetThatStepsPastTheRowThatDidNotFit` drives the config family with a first file wide enough to roll back.

src/housecarl-mcp/BatchRender.cs
Avick3110: Folded in e66bf466. The `shown == 0` guard is gone, the `headerEnd + itemLength > budget.Budget` width test stays as the thing that keeps alarms from being blamed on an ordinary item, and the sentence counts `items.Count - shown`. It also had to stop saying "the first": it now reads "the next one alone is wider than this response's whole budget", true in both positions. `AnOversizeItemAfterTheFirstIsNamedToo` sweeps caps over three narrow paths followed by a 60-provider one, asserts a narrow item is on the page with the sentence, and follows the number once to get the wide one. `AlarmsThatFillTheBudgetAreNotBlamedOnTheFirstItem` still passes on the width test alone.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded in 8dced26f: `var row = line(e) + "\n";` then measure and append `row`, which is exactly what `SkseInventoryWire.AppendCapped` and `SkseConfigAuditWire.AppendHits` already do — this was the one of the three that did not.

PR #606, opened 2026-09-06
Say why a ci-all run cannot make its fixture directory

### LINKED ISSUE #605: [Bug] ci-all throws a raw stack trace when it cannot make its fixture directory
PR #586 gives each `ci-all` run its own fixture directory under the system temp. If `Directory.CreateDirectory` fails there — a full or read-only temp volume, or a plain file already named `hc-<pid>`, which the `Directory.Exists` check does not see — the run dies on an unhandled exception out of the generator instead of saying what happened.

A failure should be one plain sentence naming the path it tried, the underlying reason, and what to try, and the run should exit non-zero. Both callers are affected: `ci-all` and the single-probe dispatch.

Avick3110:
Each `ci-all` run works in its own fixture directory under the system temp (#586). Making that directory could still fail — a full, read-only or missing temp volume, or a plain file already named `hc-<pid>`, which the `Directory.Exists` check ahead of it does not see — and the run then died on a raw stack trace out of the generator. `ProbeTemp.Redirect` becomes `TryRedirect(out string? error)`: the `CreateDirectory` call is caught and turned into one sentence naming the path it tried, the reason underneath it, and what to try, and both callers in `CiAll` report it as an `::error::` line and exit 1, the same shape a failed probe gets. Nothing is redirected on that path — the environment variables are set after the directory exists, so a failed start leaves TMP/TEMP/TMPDIR where they were.

Three things came out of review. `SweepDeadRoots` runs first, and in these states `Directory.EnumerateDirectories` throws; its catch-all counted that as one entry left from an earlier run, so every failure here printed a false residue line ahead of the start sentence. That is no longer counted as residue. Making it silent instead was wrong, though: the string-pattern overload lists with `IgnoreInaccessible = false`, so a temp directory this process may traverse and write but not *read* — an ACL granting create and denying list — denies the enumeration while the `CreateDirectory` right after it succeeds. The run then goes green with no start sentence, the sweep is skipped every run, and the roots of killed runs pile up with nothing saying why. A sweep that cannot list its temp now prints one line naming the directory and the reason, worded as a sweep that could not look rather than as entries left behind, and says what to try. The remedy in the start sentence also named only two of the causes: freeing space and removing what is at the path fix a full volume and a plain file, but not a temp location that is missing, read-only or denied, so it now also says to point TMP/TEMP at a writable directory.

Closes #605

Verified from the worktree, rebased on the current base: `dotnet build housecarl.sln -c Release` (0 errors); `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` — 1696 passed, 0 failed; `ci-all` — 127/127 passed in 1.38 min.

Verified the failures by hand rather than in a test, never touching the real temp. A plain file was written at a path in this session's scratchpad, and the generator was run twice with `TMP`/`TEMP`/`TMPDIR` pointed at that file — once as `ci-all`, once as the single probe `apply-guard`. Both printed exactly one line, `::error::could not create the fixture directory …\faketemp\hc-25180 this run works in (Cannot create '…\faketemp' because a file or directory with the same name already exists.) — free space on the temp volume, remove whatever is already at that path, or point TMP/TEMP at a writable directory, then run again.`, and exited 1. With the three variables pointed at `Z:\nope`, a missing volume, the run prints two true lines and exits 1: the sweep's `  (temp sweep could not look in Z:\nope\ for the roots killed runs leave behind (Could not find a part of the path 'Z:\nope'.) — grant this run list access there or point TMP/TEMP at a directory it can read, or those roots go unswept every run)` and then the start sentence for `Z:\nope\hc-33448`. The list-denied state was reproduced with `icacls <scratch dir> /deny "<user>:(RD)"` and the same three variables pointed at it: `apply-guard` printed that same sweep line naming the scratch directory and `Access to the path … is denied.`, then ran to `80 passed, 0 failed -> PASS` and exit 0 — the run survives, which is exactly why the line has to be there.

The tests cover the two extracted sentence builders, reflected the way `ProbeTempSweepTests` already reflects the sweep's liveness check: the start sentence names the path, the reason and what to try, and the sweep line names the directory, the reason and what to try while not reading as entries left behind. `TryRedirect` itself still cannot be driven from a test without repointing the test process's own temp directory, which is the judgement the #586 fold made and I agree with; the by-hand runs above are what exercise the real path. No CHANGELOG line: this is developer tooling and no user of the shipped plugin sees it.

### DIFF
--- a/src/housecarl-generator/ProbeTemp.cs
+++ b/src/housecarl-generator/ProbeTemp.cs
@@ -88,6 +88,7 @@ public static class ProbeTemp
     {
         int left = 0;
         string? first = null;
+        Exception? unlistable = null;
 
         try
         {
@@ -100,13 +101,22 @@ public static class ProbeTemp
                 first ??= f;
             }
         }
-        // A temp path that will not list has nothing to sweep and no residue to name: the enumeration is the
-        // only thing here that throws, and the start failure right after it is the run's one sentence.
-        catch { }
+        // Listing temp is the only thing here that throws, and it throws in states the run survives: a temp
+        // this process may traverse and write but not read denies the enumeration while CreateDirectory below
+        // succeeds. Say the sweep could not look rather than count entries it never saw — silence there skips
+        // the sweep every run and lets dead runs' roots pile up with nothing saying why.
+        catch (Exception ex) { unlistable = ex; }
 
         if (left > 0)
             Console.WriteLine($"  (temp sweep left {left} entr{(left == 1 ? "y" : "ies")} from an earlier run under {temp}: {first})");
+        if (unlistable != null) Console.WriteLine($"  ({SweepUnlistable(temp, unlistable)})");
     }
+
+    /// <summary>The one line a run prints when it cannot list temp to sweep the roots earlier runs left.</summary>
+    static string SweepUnlistable(string temp, Exception ex) =>
+        $"temp sweep could not look in {temp} for the roots killed runs leave behind ({ex.Message.TrimEnd()}) — " +
+        "grant this run list access there or point TMP/TEMP at a directory it can read, or those roots go " +
+        "unswept every run";
 
     /// <summary>Read the pid out of an <c>hc-&lt;pid&gt;</c> or <c>hc-&lt;pid&gt;-&lt;n&gt;</c> root name.</summary>
     static bool TryReadPid(string name, out int pid)
--- a/src/housecarl-generator/ProbeTemp.cs
+++ b/src/housecarl-generator/ProbeTemp.cs
@@ -44,8 +44,8 @@ public static class ProbeTemp
             ReportLeft(root, left, first);
             root = Path.Combine(temp, $"hc-{Environment.ProcessId}-{n}");
         }
-        // A full or read-only temp volume, or a plain FILE already named hc-<pid> (which Directory.Exists above
-        // does not see), stops the run before any probe: say which path and why, not a stack trace.
+        // A full, read-only or missing temp volume, or a plain FILE already named hc-<pid> (which
+        // Directory.Exists above does not see), stops the run before any probe: say which path and why.
         try { Directory.CreateDirectory(root); }
         catch (Exception ex) { error = StartFailure(root, ex); return false; }
 
@@ -66,7 +66,8 @@ public static class ProbeTemp
     /// <summary>The one sentence a run prints when its fixture root cannot be made.</summary>
     static string StartFailure(string root, Exception ex) =>
         $"could not create the fixture directory {root} this run works in ({ex.Message.TrimEnd()}) — " +
-        "free space on the temp volume, or remove whatever is already at that path, then run again.";
+        "free space on the temp volume, remove whatever is already at that path, or point TMP/TEMP at a " +
+        "writable directory, then run again.";
 
     /// <summary>Put the temp directory back and delete the root. A file still held open is reported, not thrown.</summary>
     public static void Cleanup()
--- a/src/housecarl-generator/ProbeTemp.cs
+++ b/src/housecarl-generator/ProbeTemp.cs
@@ -99,7 +99,9 @@ public static class ProbeTemp
                 first ??= f;
             }
         }
-        catch (Exception ex) { left++; first ??= ex.Message; }
+        // A temp path that will not list has nothing to sweep and no residue to name: the enumeration is the
+        // only thing here that throws, and the start failure right after it is the run's one sentence.
+        catch { }
 
         if (left > 0)
             Console.WriteLine($"  (temp sweep left {left} entr{(left == 1 ? "y" : "ies")} from an earlier run under {temp}: {first})");
--- a/src/housecarl-generator/ProbeTemp.cs
+++ b/src/housecarl-generator/ProbeTemp.cs
@@ -13,7 +13,7 @@ namespace HousecarlGenerator;
 ///
 /// <para>The run also stops leaving fixtures behind: the whole root goes at the end, entry by entry so one
 /// file another process still holds open strands only itself. A run that is killed — the bridge test's
-/// 20-minute cap, Ctrl-C, a cancelled CI job — never reaches that, so <see cref="Redirect"/> also drops the
+/// 20-minute cap, Ctrl-C, a cancelled CI job — never reaches that, so <see cref="TryRedirect"/> also drops the
 /// roots whose pid is no longer a live process; residue under our own pid that will not go is reported and
 /// the run takes a fresh sibling root rather than work in another run's fixtures. Nothing here throws: a file
 /// that will not delete is reported and the run's exit code is unaffected — cleanup never turns a run red.</para>
@@ -23,10 +23,12 @@ public static class ProbeTemp
     static string? _root;
     static string? _tmpWas, _tempWas, _tmpdirWas;
 
-    /// <summary>Point this process's temp directory at a root only this process uses, and return it.</summary>
-    public static string Redirect()
+    /// <summary>Point this process's temp directory at a root only this process uses. False with a one-sentence
+    /// reason if that root could not be made — the temp directory is then left exactly where it was.</summary>
+    public static bool TryRedirect(out string? error)
     {
-        if (_root != null) return _root;
+        error = null;
+        if (_root != null) return true;
 
         var temp = Path.GetTempPath();
         SweepDeadRoots(temp);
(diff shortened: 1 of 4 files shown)


src/housecarl-generator/ProbeTemp.cs
```diff
@@ -23,10 +23,12 @@ public static class ProbeTemp
...
+    /// <summary>Point this process's temp directory at a root only this process uses. False with a one-sentence
+    /// reason if that root could not be made — the temp directory is then left exactly where it was.</summary>
+    public static bool TryRedirect(out string? error)
     {
-        if (_root != null) return _root;
+        error = null;
+        if (_root != null) return true;
 
         var temp = Path.GetTempPath();
         SweepDeadRoots(temp);
```
Avick3110, line 34: Every fixture-root failure this PR now handles is preceded by a line from `SweepDeadRoots` that says something untrue. `Directory.EnumerateDirectories(temp, "hc-*")` throws when `temp` is not a listable directory — exactly the states this PR is about — and the catch-all counts that as residue, so the run prints:

```
  (temp sweep left 1 entry from an earlier run under ...\faketemp\: The parameter is incorrect. : '...\faketemp'.)
::error::could not create the fixture directory ...\faketemp\hc-38256 this run works in (...)
```

Reproduced on this branch both ways: TMP/TEMP/TMPDIR pointed at a plain file, and pointed at `Z:\nope` (`Could not find a part of the path 'Z:\nope'.`). There is no earlier run and nothing was left behind — the temp path simply could not be listed — so the first line, which is the first thing anyone sees, sends them after residue that does not exist. It is also why the two by-hand runs in the PR body did not in fact print exactly one line.

The fix belongs in `SweepDeadRoots`: an enumeration that fails is not the same fact as an entry that would not delete, and should either stay silent or say that the temp directory could not be listed.

src/housecarl-generator/ProbeTemp.cs
```diff
@@ -54,9 +60,14 @@ public static string Redirect()
...
         Environment.SetEnvironmentVariable("TMPDIR", root);   // what GetTempPath reads off Windows
 
-        return root;
+        return true;
     }
 
+    /// <summary>The one sentence a run prints when its fixture root cannot be made.</summary>
+    static string StartFailure(string root, Exception ex) =>
+        $"could not create the fixture directory {root} this run works in ({ex.Message.TrimEnd()}) — " +
+        "free space on the temp volume, or remove whatever is already at that path, then run again.";
```
Avick3110: The remedy covers only two of the three causes the comment above names. On a temp location that is missing, read-only, or denied, neither "free space on the temp volume" nor "remove whatever is already at that path" is the fix — the fix is to point TMP/TEMP somewhere writable, which the sentence never mentions.

Reproduced on this branch with `TMP`/`TEMP`/`TMPDIR` set to `Z:\nope`:

```
::error::could not create the fixture directory Z:\nope\hc-17344 this run works in (Could not find a part of the path 'Z:\nope\hc-17344'.) — free space on the temp volume, or remove whatever is already at that path, then run again.
```

Freeing space does nothing and there is nothing at that path to remove, so a run in CI on a container with a read-only or unmounted temp gets a sentence whose "what to try" half is false. Adding the third remedy — point TMP/TEMP at a writable directory — makes it true for every cause `CreateDirectory` can fail with.

src/housecarl-generator/ProbeTemp.cs
Avick3110: Real, reproduced, folded in 432f8de0. `Directory.EnumerateDirectories` is the only thing in that `try` that throws — `Remove`, `TryReadPid` and `IsLive` all swallow — so the catch-all was only ever catching an enumeration failure and turning it into one entry of residue. It now catches and says nothing: there was nothing to sweep, and the start failure right after it is the run's one sentence. Residue counted before the enumeration threw is still reported, since that part is real. No pre-check added — the enumeration is still just tried.

All three by-hand states now print exactly one line: temp at a plain file run as `ci-all`, the same run as the single probe `apply-guard`, and `Z:\nope` run as `ci-all`. The PR body's verification paragraph is updated with the exact output.

src/housecarl-generator/ProbeTemp.cs
Avick3110: Real, folded in 0a3c639d. The sentence now reads `— free space on the temp volume, remove whatever is already at that path, or point TMP/TEMP at a writable directory, then run again.`, which covers the missing, read-only and denied cases the first two remedies did nothing for. The comment above `CreateDirectory` and the test class doc now say "full, read-only or missing" too, and `SaysWhatToTry` asserts the third remedy.

Confirmed against your `Z:\nope` run: `::error::could not create the fixture directory Z:\nope\hc-50780 this run works in (Could not find a part of the path 'Z:\nope\hc-50780'.) — free space on the temp volume, remove whatever is already at that path, or point TMP/TEMP at a writable directory, then run again.`

src/housecarl-generator/ProbeTemp.cs
```diff
@@ -88,7 +100,9 @@ static void SweepDeadRoots(string temp)
                 first ??= f;
             }
         }
-        catch (Exception ex) { left++; first ??= ex.Message; }
+        // A temp path that will not list has nothing to sweep and no residue to name: the enumeration is the
+        // only thing here that throws, and the start failure right after it is the run's one sentence.
+        catch { }
```
Avick3110: A sweep that cannot list temp is now completely silent, and that state is not always followed by the start failure the comment relies on.

`Directory.EnumerateDirectories(temp, "hc-*")` on line 94 uses the string-pattern overload, whose `EnumerationOptions.Compatible` has `IgnoreInaccessible = false`, so a temp directory the process may traverse and write but not *read* throws `UnauthorizedAccessException` here. Creating a subdirectory needs only `w+x`, so the `Directory.CreateDirectory(root)` right after it succeeds — the run goes green, no start sentence is printed, and nothing is said about the sweep.

Concrete: a runner with `TMPDIR` pointed at a mode-0733 directory, or any temp whose ACL grants create and denies list. Every run silently skips the sweep, and the roots left behind by killed runs accumulate under temp indefinitely with no line anywhere saying why. The old `catch (Exception ex) { left++; first ??= ex.Message; }` miscounted it as residue, but it did at least name the reason.

Worth either keeping the reason and reporting it as a sweep that could not run rather than as an entry left behind, or narrowing the catch to the states where `CreateDirectory` fails too.

Avick3110: Real, and reproduced: on a scratch directory with `icacls <dir> /deny "<user>:(RD)"`, `Directory.EnumerateDirectories(dir, "hc-*")` throws `UnauthorizedAccessException` while `Directory.CreateDirectory(<dir>\hc-<pid>)` succeeds. The run goes green, so no start sentence follows and the silence was the whole answer.

Folded in 139117b0 (the branch is rebased on the current base). The catch keeps the reason and reports it as a sweep that could not run: it holds the exception, and after the residue line the sweep prints one line built by a new `SweepUnlistable`, `  (temp sweep could not look in <temp> for the roots killed runs leave behind (<reason>) — grant this run list access there or point TMP/TEMP at a directory it can read, or those roots go unswept every run)`. It is not counted as residue, and `left` is untouched, so a run that swept some entries and then lost the listing prints both lines, each about its own thing. I did not narrow the catch: the narrow set is not knowable up front, and reporting whatever the reason is cannot go silent.

By hand, with `TMP`/`TEMP`/`TMPDIR` pointed at the denied directory, `apply-guard` prints

```
  (temp sweep could not look in …\scratchpad\nolist\ for the roots killed runs leave behind (Access to the path '…\scratchpad\nolist' is denied.) — grant this run list access there or point TMP/TEMP at a directory it can read, or those roots go unswept every run)
```

then runs to `=== apply-guard: 80 passed, 0 failed -> PASS ===` and exit 0. With them pointed at `Z:\nope`, `ci-all` prints that line for `Z:\nope\` with `Could not find a part of the path 'Z:\nope'.` and then the start sentence for `Z:\nope\hc-33448`, exit 1 — two true lines, which I left rather than suppress one behind a check on the other. The real temp was never touched, and the deny ACE and the scratch directory are gone.

Pinned in `ProbeTempSweepTests` the way the start sentence is pinned in `ProbeTempStartTests`, reflecting the private builder: one test that the line names the directory, the reason and what to try, one that it does not read as entries left behind. Build 0 errors, 1696 tests passed, `ci-all` 127/127. PR body updated.

PR #609, opened 2026-09-06
Bound and account for the formids= lane on every reading form

### LINKED ISSUE #607: [GAP] the formids= lane's render bound skips summary and aggregate, which read a body per row too
Found while folding Aaron's review on PR #602 (finding 1). PR #596 put a render bound on the `formids=` list lane of `housecarl_records` (`RecordsTools.cs`, the bound near the list lane's body read), gating on `project.form` in `fields` / `rows` / `everything`. `summary` and `aggregate` route through the same `ResolveBatch` with `readFields = ["EditorID"]`, so they read a body per row as well, and a `formids=["@big.jsonl"] project.form='summary'` call is unbounded where the same call with `form='fields'` refuses.

PR #602 widened the walk lane's gate to all five body-reading forms (`bodyForm`), but left this lane alone because `limit=`'s published sentence ("a BODY form (fields/rows/everything)") matches what the lane gates on today, and changing it is a description change #602 did not own.

Fix: gate the `formids=` lane on the same `bodyForm` predicate the walk lane uses, and amend `limit=`'s sentence to match (description stays at or under 2,048 characters; `PublishedDescriptionBoundTests` pins it). One test on the list lane with `form='summary'` over the bound.

Avick3110:
The `formids=` list lane of `housecarl_records` checked the render bound only for `fields`, `rows` and `everything`, but `summary` and `aggregate` route through the same `ResolveBatch` with `readFields = ["EditorID"]` and read a body per row as well — so the same list that refused as `fields` ran unbounded as `summary`. The lane is now gated on the `bodyForm` predicate the walk lane already uses, and the `limit=` description says so. The declared constants are unchanged for those five: `summary` and `aggregate` read ONE cheap leaf per body here, which is the body seek the named-fields tier is set against, so they fall under that tier — 300,000 rows at 2 ms — and only `everything` takes the 15,000-row whole-record tier.

Second half, on Aaron's ruling in the issue comment: this lane now reports `rows_read` and `render_ms` like every other body render, on every form it renders — `summary` (this lane's default) and `aggregate` included. The read count travels with the milliseconds as one value into the renderers instead of being taken off the rendered list, because the renderers only ever see a `limit=`/`offset=` window. That count is the bodies actually READ, which is not the list's length either: under a named `source=` pole an id the pole has no version of refuses per item without a read, and a malformed token never reaches one, so a 2,000-id list over a pole holding three used to print `read 2000 record bodies in 6 ms` and divide the clock by rows that cost nothing. The bound itself is still measured on the list — that is what the lane will read at worst, and the refusal states it. The scan lane and the off-order lane pass the count they already knew, so what they report is unchanged.

`identity` does NOT stay out, contrary to an earlier version of this description. `ResolveRefOne` fetches the winner's body for the type, editorid and name, and `IndexSnapshot` carries no record type, so `GetRecord` reaches `SeekBody` with no `getterType` — the untyped whole-plugin scan, one per id. Measured on the live ARR order (3,801 plugins) over 3,000 WEAP ids, twice: **12.5 and 14.0 ms a row for `identity`, against 0.054 and 0.072 ms for the same ids read as three named fields, and 0.045–0.050 ms for `summary`'s one leaf** — the fields and summary reads cost 0 body seeks because they gather per plugin, the identity read cost 3,000. So `identity` is the dearest row on this lane, not a free one, and it is now bounded on a tier of its own: 15 ms declared, 40,000 rows. A typed seek was the other option and is not available — the index holds `(winner, count)` per FormKey and no type, so knowing the type would itself cost the scan. Making `ResolveRefs` gather per plugin the way `ResolveBatch` does would remove the cost rather than bound it; that is a separate change, filed as #610, which is what would retire this tier. Its accounting counts the ids that RESOLVED, for the same reason: a malformed or absent id never reaches a read.

The identity refusal names `project.form='summary'` first, the way the whole-record refusal names `form='fields'`: summary answers the same type/editorid/winner question off a gathered read at a fraction of the cost. The shared `ListRemedy` is unchanged — for a `summary` refusal, "pass fewer ids" is the only lever this lane has, since `to_file=` and `counts_only` are bounded on it too.

Also folded from the review: a walk with `counts_only=true` was bounded by nothing. The three walk gates skipped the bound on `!counts_only`, on the scan lane's reasoning that a census reads no bodies — but a walk does not render its own reached set, it hands it to the list lane, which reads a body per id before it counts anything, and the list lane's own gate is skipped on `walkDerived`. So `walk={"max_nodes": 500000}` with a body form and `counts_only=true` materialised every reached body. All three gates now measure the census like the render.

And the refusal a census gets is written for a census. `RenderBudget.Refuse`'s lead opened `this call renders N rows and each one reads a record body`, which is false of a call that renders nothing, and it left out the fact that explains the refusal — the list lane reads every reached body BEFORE it counts them — while the remedy pointed at a `limit=`/`offset=` window over a render the caller never asked for. The lead now has a `counts_only` arm, and each lane's remedy a census twin composed from the same lever constant so the two cannot drift. All four gates that can refuse a census pass it: the three walk gates and the `formids=` lane's own, which hands out the same lead.

Closes #607

Verified from the worktree: `dotnet build housecarl.sln -c Release` (0 errors), `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` (1,767 passed, 0 failed), and `dotnet .../housecarl-generator.dll ci-all` (127 probes, ALL PASS). Each new bound and each new sentence was measured against the code before it: the summary and aggregate refusals fail with the gate reverted; `AWalkCensusIsHeldToTheBoundBecauseTheListLaneStillReadsEveryBody` returns a rendered census rather than a refusal with the `!counts_only` exemption in place; the identity tests fail with no identity gate; the two accounting tests report 65 and 63 rather than 5 and 60 with the count back at `ids.Length`; and the two census-wording tests get the render's lead with the census arm reverted. Tests added, one per claim, all through the shipped tool: the summary and aggregate refusals and their accounting on text and json; the identity refusal, its tier's separation from the named-fields one, and its accounting; the walk census refusal and its wording, the `formids=` census wording, and a pin that a rendered walk still speaks of its render; the pole and malformed-token accounting; and two pins that a rows-form fold does not inflate `rows_read`.

One test outside this pair moved: `RecordsOwnedChildTests`'s two padded-cell reads now pass `max_chars: 1500` instead of 1400, because the lane reserves room for its accounting line the same way the scan lane does and the annotated field no longer fit under the old cap. The claims those tests make are unchanged.


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -657,8 +663,8 @@ string Render(SpillState? sp, out bool trunc) => json
             }
             string Render2(SpillState? sp, out bool trunc) => form == "summary"
                 ? RenderRecordsSummary(winOutcomes, json, headerLine, envelope, max_chars, sp, out trunc)
-                : json ? JsonWire.RenderBatch(winOutcomes, max_chars, sp, out trunc, envelope, formLevers)
-                       : Wire.RenderBatch(winOutcomes, max_chars, sp, out trunc, formLevers, header: headerLine);
+                : json ? JsonWire.RenderBatch(winOutcomes, max_chars, sp, out trunc, envelope, formLevers, (ids.Length, listClock.ElapsedMilliseconds))
```
Avick3110: **The two forms this PR newly bounds are the two that still report nothing.**

`Render2` hands `(ids.Length, listClock.ElapsedMilliseconds)` to `JsonWire.RenderBatch`/`Wire.RenderBatch`, but `form == "summary"` short-circuits to `RenderRecordsSummary(...)`, which takes no cost argument, and `form == "aggregate"` returned at line 646 through `RenderListAggregate(...)`, which takes none either. So the accounting lands only on `fields`, `rows` and `everything` â€” the three forms that already had it via the scan lane's precedent.

`summary` is this lane's DEFAULT form. Concretely: `housecarl_records(formids=["@big.jsonl"])` with no `project` at all now reads a body per id, is refused at 300,000, and on a list just under the bound renders with no `rows_read` and no `render_ms` â€” the caller who was just refused and sliced the list has no number to check the slice against.

The CHANGELOG bullet added in this PR reads "And the response reports the rows it read and the milliseconds they took, on text and json alike" as the second half of a bullet whose subject is all five reading forms, and the PR body says "this lane now reports `rows_read` and `render_ms` like every other body render." Both are true of three forms out of five.

Either thread `listClock` into `RenderRecordsSummary` and `RenderListAggregate` the way it is threaded here, or narrow the changelog sentence to the forms that actually carry it.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1464,8 +1470,8 @@ List<KeyValuePair<string, string>> Echo()
                 // clause must name limit= — the knob that actually windows this response.
                 var evLevers = formLevers.OnScanSelection();
                 string RenderEv(SpillState? sp, out bool trunc) => json
-                    ? JsonWire.RenderBatch(bodies, max_chars, sp, out trunc, envelope, evLevers, bodyClock.ElapsedMilliseconds)
-                    : Wire.RenderBatch(bodies, max_chars, sp, out trunc, evLevers, bodyClock.ElapsedMilliseconds, headerLine);
+                    ? JsonWire.RenderBatch(bodies, max_chars, sp, out trunc, envelope, evLevers, (bodies.Count, bodyClock.ElapsedMilliseconds))
```
Avick3110, line 1501: **The new `RowsRead` parameter is documented as "what was read", and two of its three call sites pass what was RENDERED.**

The doc comment this PR adds to both renderers says the count "travels WITH the milliseconds rather than being taken off this list" so that "the count beside the cost is what was read". The formids= lane honours that — it passes `ids.Length`. This lane, and the off-order lane at line 1693, pass `bodies.Count` — and `bodies` was reassigned by `bodies = FoldRows(bodies)` a few lines above (line 1444 here, 1675 there), AFTER `bodyClock.Stop()`.

`FoldRows` expands: `form == "rows"` runs `RowProjection.Apply`, which turns one outcome into one row per list element, and a quantified `project.fields` path runs `foldPlan.Apply`, which does the same. So a scan with `project.form='rows'`, `fields=["Conditions"]` over 3 records carrying 10 conditions each reports `rows_read: 30` and prints `read 30 record bodies in N ms` — 3 bodies were read, and `N` is the time for 3.

That over-reports by the fold factor, and it makes the same JSON member mean two different things depending on which lane answered: bodies on `formids=`, folded element rows on a scan. It also breaks the stated purpose of the number, which is to be checkable against the bound the lane is measured on — the bound counts `outcome.Keys.Count`, not folded rows.

`keys.Count` (already in scope on both lanes, and the exact value each lane's `RenderBudget.Refuse` was measured against) is the count that matches the parameter's contract.

The behaviour predates this PR, but this PR is what turns the count into a named, separately-passed value with a contract — so it is the place to make the three call sites agree.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -575,7 +575,9 @@ string Render(SpillState? sp, out bool trunc) => json
             // lane does, and formids=["@<file>"] is a list of any length by construction — the very path the
             // whole-record refusal and to_file= point at. Checked before any body is read, and counts_only pays it
             // too: this lane reads the list whatever it renders, where the scan lane's census reads no bodies.
-            if ((bodyFields || form == "everything") && !walkDerived
+            // Measured over ALL five reading forms: summary and aggregate read one cheap leaf off a body here, so
+            // they pay the body seek the named-fields tier is set against (#607).
+            if (bodyForm && !walkDerived
```
Avick3110, line 591: **`!walkDerived` leaves the hole this line closes still open on the walk path — and it is the `counts_only` half specifically.**

The comment directly above says the gate is checked "before any body is read, and `counts_only` pays it too: this lane reads the list whatever it renders". That is exactly right for a caller-passed list, and dropping `!counts_only` from this gate is what makes it true.

But when the list came from a walk, `walkDerived` is true and this check is skipped, and the walk lane's own three gates (lines 742, 831, 932) are each spelled `bodyForm && !counts_only`. So a walk with `counts_only: true` is measured by neither:

```
housecarl_records(formids=[…seeds…],
                  walk={"direction":"forward","depth":4,"max_nodes":500000},
                  project={"form":"everything"},
                  counts_only=true)
```

The walk lane skips its `Refuse(combined.Count, …)` on `!counts_only`, sets `walkDerived = true` and re-enters `ListLane`, which skips this gate on `!walkDerived`. `svc.ResolveBatch(ids, readFields, …)` at line 618 then runs over the whole reached set BEFORE the `counts_only` census returns at line 648 — so 500,000 whole-record bodies are materialised unbounded, which is the silent-past-the-idle-timeout failure #582 and this PR are both about.

The walk gates' `!counts_only` rests on the same assumption this line just retired: that a census reads no bodies. It does on the scan lane, whose census comes off the index; it does not on the list lane, which reads the bodies either way. Dropping `!counts_only` from the three walk gates makes the reasoning in the comment above hold everywhere it is claimed to.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -575,7 +575,9 @@ string Render(SpillState? sp, out bool trunc) => json
             // lane does, and formids=["@<file>"] is a list of any length by construction — the very path the
             // whole-record refusal and to_file= point at. Checked before any body is read, and counts_only pays it
             // too: this lane reads the list whatever it renders, where the scan lane's census reads no bodies.
-            if ((bodyFields || form == "everything") && !walkDerived
+            // Measured over ALL five reading forms: summary and aggregate read one cheap leaf off a body here, so
```
Avick3110, line 589: **Bounding `summary` here makes `ListRemedy` incomplete: the refusal never names the form that answers the same question for free.**

Before this change a big `formids=` list had a working shape — `project.form='summary'`, unbounded. That shape is now refused, and the refusal it gets is `RenderBudget.ListRemedy` (RenderBudget.cs:85), which only ever says *shrink the list*: "pass fewer `formids=` entries … Re-enter a big artifact a slice at a time, or narrow the selection that wrote it."

But `form='identity'` returns type / editorid / name / winner per FormID off `ResolveRefs`, reads no body, and is excluded from `bodyForm` — so it is not bounded at all. For a caller who asked for `summary`, that is very nearly the same answer at zero body cost, and it is the one remedy that does not require them to give up rows.

So the concrete misbehaviour: `housecarl_records(formids=["@400k.jsonl"])` used to work and now returns a refusal that tells the caller to slice a 400,000-id artifact by hand, while a one-word change to the call would have answered it whole. That is the tool naming a lever that costs the caller their result over one that does not — the opposite of what the errors cornerstone asks for.

`ListRemedy` is shared with `fields`/`rows`/`everything`, where `identity` is not equivalent, so this probably wants a summary-specific lead rather than an edit to the shared constant — the same way `Refuse` already gives `form='everything'` its own "name the fields you need instead" first sentence before falling through to the shared lever.

src/housecarl-mcp/RecordsTools.cs
Avick3110, line 591: **`identity` does read a body — and by the untyped flat seek, the most expensive one on this lane.**

The PR body says "`identity` reads no body and stays out", and the CHANGELOG headline says the lane "is now held to the render bound on every form that reads a body". `bodyForm` (line 174) excludes `identity`, so the identity arm at line 536 passes this gate untouched.

But `svc.ResolveRefs(ids, …)` at line 541 lands in `LoadOrderService.ResolveRefOne` (LoadOrderService.cs:2757), which does:

```csharp
var body = view.GetRecord(session, w.Value.WinnerPlugin, fk);
… EditorId: body.EditorID, Name: ReadDisplayName(body) …
```

`GetRecord` with no `getterType` reaches `SeekBody` (LoadOrderResolver.cs:1100), which increments `BodySeeks` and runs `ov.EnumerateMajorRecords()` — the **flat** whole-plugin scan, per id. That is the per-row whole-plugin seek `RenderBudget.MillisPerRow`'s own doc comment cites as having measured ~40 ms a row on the ARR order before it was removed elsewhere. So `identity` is not the one form below the bound; on this lane it is plausibly the one *above* the tier the other four were just put in.

What that leaves:

- `formids=["@400k.jsonl"] project.form='summary'` now refuses. The same list with `form='identity'` reads 400,000 bodies with no bound and no accounting — the silent hang #582 exists to prevent, still reachable, and now by the one form a caller is steered toward when `summary` refuses.
- `limit=`'s new sentence, "On a formids= read EVERY reading form reads a body — summary and aggregate one cheap leaf too — so the bound holds there for all five", is not true as written: `identity` is a reading form, it reads a body, and the bound does not hold on it. The sentence a few clauses earlier — "every rendered row of a BODY form (fields/rows/everything) READS that record's body" — is left unedited and now disagrees with it too, naming three forms where the new one names five.

The fix is either the gate covering `identity` at the named-fields tier, or, if it is meant to stay out, a stated reason that survives reading `ResolveRefOne` — "reads no body" does not.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded — dd344b90.

Right on both arms. `summary` is this lane's default form and `aggregate` returns earlier still, so the two forms the bound was just extended to were the two reporting nothing. The clock is now taken once as `listCost` and every form this lane renders is handed it: `RenderRecordsSummary` and `RenderListAggregate` both take the cost, write `rows_read`/`render_ms` on json, and append the `read N record bodies in M ms` line on text. The summary text render also holds `RenderBudget.AccountingReserve` back from its row budget, the way the batch render does, so the line is charged rather than appended past `max_chars`.

Three tests: `AFormidsSummaryRenderReportsWhatItsBodiesCost`, `AFormidsSummaryTextRenderReportsWhatItsBodiesCost`, `AFormidsAggregateRenderReportsWhatItsBodiesCost`. The changelog sentence stands as written now.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Refuted — `FoldRows` does not expand the outcome list, so `bodies.Count` is `keys.Count`.

`RowProjection.Apply` (RowProjection.cs:39) does one `folded.Add` per input outcome and rewrites that outcome's FIELD list: `o with { Record = o.Record with { Fields = Fold(...) } }`. The per-element lines live inside one record's row. `FoldPlan.Apply` (FieldFold.cs:118) is `outcomes.Select(Apply).ToList()` — 1:1 too. So a rows-form scan over 3 records carrying 10 conditions each renders 3 records, not 30, and reports `rows_read: 3`.

Measured rather than argued. I built the fold into the render-cost world (3 keywords on every weapon) and asserted the shape, against the code as it stands: `AFoldedScanReportsTheBodiesItReadAndNotTheRowsTheFoldMade` gets `rendered: 60`, `rows_read: 60`, and 4 field lines on the first record (the root count line plus one per keyword); `AFoldedOffOrderScanReportsTheBodiesItReadToo` gets 12 and 12. Both pass unchanged — c39ae785 adds them as pins rather than a fix, so a fold that ever does expand rows lands as a red test instead of a wrong number.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded — 461e6773. Same one-line shape on all three walk gates.

The trace holds. `!counts_only` on the walk gates rests on the scan lane's reasoning, and it is wrong here for the reason this PR just retired it on the list gate: a walk does not render its own reached set, it hands it to `ListLane`, which runs `ResolveBatch` over every id before the `counts_only` census returns. Read against the code: the list gate is skipped on `walkDerived`, the walk gate on `!counts_only`, so the shape you wrote was measured by neither.

Measured before the change: `AWalkCensusIsHeldToTheBoundBecauseTheListLaneStillReadsEveryBody` — a walk with `counts_only: true`, a fields form and the bound at 10 — came back with a rendered census over 60 bodies rather than a refusal. With `!counts_only` dropped from all three gates it refuses, naming the walk lane's own remedy (`project.form='chain'`) and not the scan's. Full suite and ci-all clean, so no shape that legitimately counted without reading is now refused.

The comment above the list gate is true as written; the three walk gates now carry the same one-line reason.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Refuted as written, but the remedy work landed the other way round — b24b427c.

The suggestion rests on `identity` being the cheap answer, and it is the opposite one (see the reply on the `identity` thread). Measured on the live ARR order, 3,000 ids: `identity` 12.5–14.0 ms a row, `summary` 0.045–0.050 ms. So a caller refused on `summary` cannot be sent to `identity` — that swap costs them 250x and, before this PR, no bound at all.

Which leaves `ListRemedy` correct and complete for a `summary` refusal: this lane has no scan terms, `to_file=` renders the whole selection and is held to the same bound, and `counts_only` reads the same bodies and is held to it too. "Pass fewer ids, re-enter a big artifact a slice at a time, or narrow the selection that wrote it" is the whole lever set. Adding a summary-specific lead would mean naming a shape that does not exist.

The form-specific lead you were after does exist now, pointing the other way: `RenderBudget.RefuseIdentity` names `project.form='summary'` first — the same facts off a read gathered per plugin, bounded at 300,000 rows — before falling through to the shared lever, exactly as the whole-record lead names `form='fields'`.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded — b24b427c. The trace is right, and the measurement is worse than the estimate you extrapolated.

Verified: `ResolveRefOne` (LoadOrderService.cs:2757) calls `view.GetRecord(session, w.Value.WinnerPlugin, fk)` with no `getterType`; `GetRecord` hands that null straight to `SeekBody` (LoadOrderResolver.cs:1101), which increments `BodySeeks` and runs the flat `ov.EnumerateMajorRecords()`. One untyped whole-plugin scan per id.

Measured on the live ARR order (3,801 plugins), 3,000 WEAP ids, two runs:

| read | per row | body seeks |
|---|---|---|
| `identity` (`ResolveRefs`) | 14.03 / 12.55 ms | 3,000 |
| `fields`, three paths (`ResolveBatch`) | 0.072 / 0.054 ms | 0 |
| `summary`'s one leaf | 0.050 / 0.045 ms | 0 |

Zero seeks on the other two because they gather per plugin. So `identity` is ~250x the form it was described as being cheaper than.

I took (a). (b) is not available: `IndexSnapshot.Index` holds `(winner, count)` per FormKey and no record type, so a typed seek would need a scan to learn the type. What (b) would really be is making `ResolveRefs` gather per plugin the way `ResolveBatch` already does — that removes the cost instead of bounding it, and it is a change to a service method several lanes share, so it is filed as #610 with these numbers rather than folded here.

What landed: `identity` gets its own tier — `MillisPerIdentityRow = 15.0` (measurement plus margin), `DefaultMaxIdentityRows = 40_000` (ten minutes at that rate). Its own lead too, because "reads a WHOLE record body" and "name the fields you need" are both false for it: `RefuseIdentity` says the row is an untyped whole-plugin seek, names `project.form='summary'` as the same facts off a gathered read at the 300,000-row tier, mentions `fields=["Name"]` for the display name summary does not carry, then falls through to `ListRemedy`. It reports `rows_read`/`render_ms` now as well, so the new tier is checkable against a real order rather than trusted.

Both `limit=` sentences are rewritten together: the earlier clause no longer enumerates three body forms, the tier list carries all three numbers, and `identity` is named as the dearest row rather than the exempt one. Tests: `AnIdentityReadOverItsOwnBoundRefuses`, `TheIdentityBoundAndTheFieldsBoundAreSeparate`, `AnIdentityReadReportsWhatResolvingItsFormidsCost`.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -631,13 +643,16 @@ string Render(SpillState? sp, out bool trunc) => json
                     headerLine += "\n(the off-order file's content is OUTSIDE the epoch fingerprint — an edit to it changes answers without changing the epoch)";
                 }
             }
+            listClock.Stop();
+            // One cost for every form this lane renders, so the accounting cannot land on some of them and not others.
+            var listCost = (ids.Length, listClock.ElapsedMilliseconds);
```
Avick3110: `rows_read` is `ids.Length`, but this lane does not read a body for every id. Under a named `source=` pole, `ResolveBatchFromPole` returns a per-item refusal for any id the pole has no version of (`'X.esp' does not touch <fk>`, LoadOrderService.cs:2414) without reading a body; a malformed token never reaches a read at all.

So `formids=[2,000 ids] source="SmallPatch.esp" project.form='fields'` where the pole overrides 3 of them prints `read 2000 record bodies in 6 ms` — 3 bodies read, 2,000 claimed. Dividing that gives ~0.003 ms a body against the 2 ms/row the 300,000-row tier is declared at, which is the opposite of what the accounting is for: the PR adds it so a tier calibrated on one machine is checkable where the truth comes back, and on the pole path it is off by the touched/untouched ratio.

Line 548 has the same shape on the identity lane — `ResolveRefs` skips a malformed token before `ResolveRefOne`, so a list of bad FormIDs still reports `read N record bodies`.

The count the reads actually cost is available on both lanes: `outcomes.Count(o => o.Record is not null)` here, `rows.Count(r => r.Resolved)` there. If the intent is that the number stays the LIST's length because that is what the bound is measured against, then the line's own words (`read N record bodies`) are what needs changing, not the count.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -923,7 +940,8 @@ string Render(SpillState? sp, out bool trunc) => json
             // walk.max_nodes rows, each one a body read by the list lane. Same two lanes as the scan's bound, its
             // own remedy, since the scan window is not what moves a walk. form='chain' stays exempt and returned
             // above: it renders the walk's own rows, so it pays no SECOND body read on top of the walk's own.
-            if (bodyForm && !counts_only
+            // counts_only pays it too: the list lane reads every reached body whether it renders them or counts them.
+            if (bodyForm
                 && RenderBudget.Refuse(combined.Count, form == "everything", RenderBudget.WalkRemedy) is { } walkTooBig)
```
Avick3110: Bounding the census is right, but it now routes through a sentence written for a render. `RenderBudget.Refuse`'s non-whole-record lead (RenderBudget.cs:134) opens `this call renders {rows} rows and each one reads a record body`, and a `counts_only` call renders no rows.

`housecarl_records(types=["WEAP"], walk={"max_nodes": 500000}, project={form:"summary"}, counts_only=true)` over a 412,000-node reached set now answers `error: this call renders 412,000 rows and each one reads a record body — about 14 minutes of render…`. The caller asked for a count and is told about a render they did not request; the one fact that explains the refusal — the list lane reads every reached body *before* it counts them, which is the reasoning in the comment right above this gate — is the one thing the sentence does not say. The remedy then adds `limit= and offset= window the render and not the walk`, which reads as advice about a render that was never in the call.

Same three gates: 753 and 843 too. A `counts_only` arm on the lead ("counts N records and each one reads a record body before it is counted") would make the sentence true on both paths.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 495daacb. The count is now the bodies actually read on both lanes: `outcomes.Count(o => o.Record is not null)` on the pole/batch lane, `rows.Count(r => r.Resolved)` on identity. Took the first of your two shapes rather than rewording the line — the accounting exists to check the declared tier against the truth, and the bodies read is the number that does that; the refusal already states the list length, so nothing is lost by the line not repeating it. The BOUND still measures `ids.Length`, which is what the lane will read at worst.

Two tests, measured against the code before: `AFormidsPoleReadCountsOnlyTheBodiesThePoleHeld` (5 ammo the pole defines plus all 60 weapons it does not, under `source="HcCostSpread0.esp"`) reported 65 and now reports 5; `AnIdentityReadCountsOnlyTheIdsThatResolved` (three bad tokens plus 60 weapons) reported 63 and now reports 60. Both assert `count` is still the full list, so the two numbers stay distinguishable. The `limit=` description, the changelog bullet and the `rows_read` doc comments now say bodies read rather than list length.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in 9c5a0cc0. `RenderBudget.Refuse` takes a `census` flag: the lead becomes `this call counts N records and each one reads a record body before it is counted — about X of reading, past the …-row bound`. Each lane keeps its render remedy and gains a census twin, both composed from one shared lever constant (`WalkLever`, `ReverseCarrierLever`, `ReverseTransitiveLever`, `ListLever`) so the pair cannot drift: the walk twin reads `… until the set the walk reaches fits — the count is what the walk reached, seeds included, and counts_only= does not lower it: the bodies are read before they are counted. project.form='chain' counts the same reached set without reading a body per record`. Checked that chain honours `counts_only` before pointing at it — it answers `seeds/reached/errors`.

Four gates pass it, not three: the `formids=` lane's own gate at 588 hands out the same lead and had the same untruth on a direct `formids=[…] counts_only=true` refusal, so it takes `ListCensusRemedy`. `RefuseIdentity` gets the flag too, for the one word it shares (`of render` → `of reading`). Say if you would rather the fourth gate came out.

Three tests: the walk census and the `formids=` census both assert the census words and assert `renders`/`window the render` are absent (they fail against the previous commit — `error: this call renders 63 rows and each…`), and `ARenderedWalkRefusalStillSpeaksOfItsRender` pins that the render arm is untouched.

PR #618, opened 2026-09-07
Read a record with a draft SkyPatcher INI folded into the layer

### LINKED ISSUE #613: [GAP] Offline SkyPatcher draft check: verify an INI before it is placed
skypatcher-authoring's verify step calls `housecarl_skypatcher_read`, which no longer exists. The live replacement, `housecarl_records` with `source={"overlay":"skypatcher","state":"post"}`, needs the INI placed in a mod folder and that mod enabled, so a draft cannot be checked before it is written into a live install. The skill's one claimed safety step is a guaranteed failed call until a draft check exists.

Wanted: a way to read a record as it would be after a SkyPatcher INI that is not yet placed replays over it, offline, without writing into the mods directory. Found by the skill eval (`dev/skill-validation/eval-2026-09/skypatcher-authoring/REPORT.md`, D4.41 and D5.51); ruled into 2.0, wave 0 of the skill rewrite (DECISIONS 2026-09-07, rewrite ruling 6).

Avick3110:
`housecarl_records` could already read a record after the SkyPatcher INI layer replays, but only for INIs discovered under `SKSE\Plugins\SkyPatcher` in an enabled mod — so a draft INI could not be checked until it had been written into a mod folder, which put the skill's verify step after the write instead of before it. This adds one value to the existing overlay pole: `source={"overlay": "skypatcher", "state": "post", "ini": "<absolute path to a draft .ini>", "subfolder": "weapon"}` reads the record as the game would see it once the draft is placed — the live layer plus the draft, sorted into its type folder by filename among the live files with the same ordinal sort discovery uses, and the same `Plugin.esp.ini` filename gate. `subfolder` may be omitted when the draft's parent directory is already a documented type folder, and the arm line says the folder came from there. The pole is legal as `versus=` too, so a draft's post state against the plain post state is the draft's own effect and nothing else. Refusals: an undocumented subfolder (naming the documented list), a filename already placed in that folder (placing it would shadow or be shadowed depending on mod order, so the answer would not be the draft's), `state: "pre"` with a draft, a missing file, and `subfolder` with no `ini`. The scan-lane refusal for an overlay pole is unchanged; the draft rides the `formids=` lane only.

The draft-to-`IniFile` construction and the merge into a `LayerScan` live in a new `src/housecarl-core/SkyPatcherDraft.cs`; the pole parsing sits beside the existing overlay parsing in `RecordsTools.cs`, and `LoadOrderService` only passes the plan through to the two places that discover the layer. One thing had to come with it: the replay's per-line warnings (unknown key, unmapped op, unresolved filter, parse note) went nowhere on the records lane, so a skipped draft line would have read as "the draft changes nothing". They are now collected and rendered beside the answer on both the read and delta lanes, each already carrying its own file and line — which for a draft is the draft's path.

Tests: `RecordsSkyPatcherDraftTests` covers a draft that sets a leaf, the inferred subfolder, the delta composition, an unknown key warning naming the draft, a draft gated on an inactive plugin, and each refusal. `RecordsWorld` gains one comment-only placed INI so the layer has a real type folder and a filename to collide with; it patches nothing, so no existing test changes. Full suite 1778 passed, `ci-all` 127/127.

Closes #613

### DIFF
--- a/src/housecarl-core/SkyPatcherDraft.cs
+++ b/src/housecarl-core/SkyPatcherDraft.cs
@@ -60,7 +60,7 @@ public static class SkyPatcherDraft
             {
                 refusal = $"the draft's filename '{name}' is already live in the '{Subfolder}' folder as '{clash.RelPath}'"
                         + (clash.WinningProvider is null ? "" : $" (from '{clash.WinningProvider}')")
-                        + ", so placing the draft there would shadow that file or be shadowed by it depending on mod order — rename the draft, or nest it under a mod-specific subfolder.";
+                        + ", so placing the draft there would shadow that file or be shadowed by it depending on mod order — give the draft a filename not yet placed in that folder, which is also what decides where it sorts among them.";
                 return live;
             }
 
--- a/src/housecarl-core/SkyPatcherDraft.cs
+++ b/src/housecarl-core/SkyPatcherDraft.cs
@@ -38,6 +38,20 @@ public static class SkyPatcherDraft
             for (int i = 0; i < live.Folders.Count; i++)
                 if (live.Folders[i].Subfolder.Equals(Subfolder, StringComparison.OrdinalIgnoreCase)) { at = i; break; }
 
+            // A draft that IS one of the layer's live files: folding it in would put the same file in the folder
+            // twice and replay every one of its lines twice, a post state the game never produces. The filename
+            // clash below catches this only when the live copy sits flat in the type folder, since a nested one
+            // sorts under 'MyMod\Blades.ini' and no bare filename matches it.
+            var placed = live.Folders.SelectMany(f => f.Files)
+                             .FirstOrDefault(f => f.LooseFilePath is { } p && SamePath(p, IniPath));
+            if (placed is not null)
+            {
+                refusal = $"the draft '{IniPath}' is already placed — it is the file the layer reads as '{placed.RelPath}'"
+                        + (placed.WinningProvider is null ? "" : $" (from '{placed.WinningProvider}')")
+                        + ", so folding it in would replay its lines twice; read it with the plain post state by dropping \"ini\".";
+                return live;
+            }
+
             // Same relative path once placed = the VFS same-path collision: one copy wins by mod order and the
             // other is never read, so which body this call would answer with is not the draft's to decide.
             var clash = at < 0 ? null
@@ -68,7 +82,7 @@ public static class SkyPatcherDraft
                 warnings?.Add($"{IniPath}: the draft is {notApplied} — SkyPatcher would not read it once placed, so this post state is the plain winner.");
             }
 
-            var draft = new SkyPatcherDiscovery.IniFile(IniPath, Subfolder, name, DraftProvider,
+            var draft = new SkyPatcherDiscovery.IniFile(IniPath, Subfolder, name, DraftProvider, IniPath,
                                                         Array.Empty<string>(), gate, notApplied, lines);
 
             // The SkyPatcher.ini [Patcher] toggle for this type. A folder the live scan built carries it already;
@@ -100,6 +114,13 @@ public static class SkyPatcherDraft
                 "so this reads what the game would see once it is placed there.").ToList();
             return new SkyPatcherDiscovery.LayerScan(folders, notes, live.ReadIncomplete, live.PatcherToggles);
         }
+    }
+
+    /// <summary>Two paths naming the same file, compared as the filesystem does here: normalized, case-insensitive.</summary>
+    static bool SamePath(string a, string b)
+    {
+        try { return Path.GetFullPath(a).Equals(Path.GetFullPath(b), StringComparison.OrdinalIgnoreCase); }
+        catch { return false; }
     }
 
     /// <summary>What the draft's provider column says: it comes from no mod, and saying so keeps a render from
--- a/src/housecarl-core/SkyPatcherDraft.cs
+++ b/src/housecarl-core/SkyPatcherDraft.cs
@@ -25,8 +25,9 @@ public static class SkyPatcherDraft
         /// <summary>Fold this draft into a live layer scan: its type folder gains the draft, sorted among the
         /// live files by filename exactly as <see cref="SkyPatcherDiscovery.Scan"/> sorts them. Returns the live
         /// scan unchanged when <paramref name="refusal"/> is set. <paramref name="warnings"/> collects the facts
-        /// a replay would otherwise swallow — a gated-off draft applies nothing, and silence there would read as
-        /// "the draft changes nothing".</summary>
+        /// a replay would otherwise swallow — a gated-off draft, or one in a type the <c>SkyPatcher.ini</c>
+        /// <c>[Patcher]</c> toggles off, applies nothing, and silence there would read as "the draft changes
+        /// nothing".</summary>
         public SkyPatcherDiscovery.LayerScan Fold(SkyPatcherDiscovery.LayerScan live, SkyPatcherCatalog catalog,
                                                   Func<string, bool> pluginPresent, out string? refusal,
                                                   List<string>? warnings = null)
@@ -70,10 +71,20 @@ public static class SkyPatcherDraft
             var draft = new SkyPatcherDiscovery.IniFile(IniPath, Subfolder, name, DraftProvider,
                                                         Array.Empty<string>(), gate, notApplied, lines);
 
+            // The SkyPatcher.ini [Patcher] toggle for this type. A folder the live scan built carries it already;
+            // a folder invented here for a type with no live INI has no scan to carry it, so it is read off the
+            // layer's toggle map — assuming enabled there would read a disabled type as applied.
+            bool folderEnabled = at >= 0
+                ? live.Folders[at].PatchingEnabled
+                : SkyPatcherDiscovery.ToggleEnabled(live.PatcherToggles, Subfolder);
+            if (!folderEnabled)
+                warnings?.Add($"{IniPath}: SkyPatcher.ini disables '{Subfolder}' patching (iEnable…Patching=0) — the DLL skips the whole folder, " +
+                              "so the draft would apply nothing once placed and this post state is the plain winner.");
+
             var folders = new List<SkyPatcherDiscovery.FolderScan>(live.Folders);
             if (at < 0)
             {
-                folders.Add(new SkyPatcherDiscovery.FolderScan(Subfolder, catalog.ForSubfolder(Subfolder), true,
+                folders.Add(new SkyPatcherDiscovery.FolderScan(Subfolder, catalog.ForSubfolder(Subfolder), folderEnabled,
                                                                new List<SkyPatcherDiscovery.IniFile> { draft }));
                 folders = folders.OrderBy(f => f.Subfolder, StringComparer.OrdinalIgnoreCase).ToList();
             }
@@ -87,7 +98,7 @@ public static class SkyPatcherDraft
             var notes = live.Notes.Append(
                 $"the draft INI '{IniPath}' was folded into the '{Subfolder}' folder as '{name}' — it is NOT on disk in a mod, " +
                 "so this reads what the game would see once it is placed there.").ToList();
-            return new SkyPatcherDiscovery.LayerScan(folders, notes, live.ReadIncomplete);
+            return new SkyPatcherDiscovery.LayerScan(folders, notes, live.ReadIncomplete, live.PatcherToggles);
         }
     }
 
--- a/src/housecarl-core/SkyPatcherDraft.cs
+++ b/src/housecarl-core/SkyPatcherDraft.cs
@@ -1,0 +1,136 @@
+namespace HousecarlCore;
+
+/// <summary>
+/// DRAFT INI: fold one not-yet-placed SkyPatcher INI into the discovered layer, so a record can be read
+/// as the game would see it once the draft is placed in a mod. A draft is a file anywhere on disk plus
+/// the type subfolder it would be placed in; everything after that is the live layer's own machinery,
+/// unchanged — the <c>Plugin.esp.ini</c> filename gate, the ordinal sort within the type folder, the
+/// per-line warnings. The draft's file NAME in the layer is its full path, so every warning the replay
+/// produces for one of its lines names the draft rather than a Data-relative path that does not exist.
+///
+/// <para><b>No shadowing.</b> A draft whose filename collides with a live INI at the same relative path
+/// is refused: once placed, one of the two would win the VFS and which one depends on mod order, so the
+/// answer would not be the draft's.</para>
+/// </summary>
+public static class SkyPatcherDraft
+{
+    /// <summary>A validated draft: the file, the type folder it would sit in, and whether that folder was
+    /// taken from the file's parent directory rather than named — which the arm line says out loud.</summary>
+    public sealed record Plan(string IniPath, string Subfolder, bool SubfolderInferred)
+    {
+        /// <summary>The arm clause a response leads with, so the answer always names the draft it included.</summary>
+        public string Arm => $"the draft INI '{IniPath}' placed in the '{Subfolder}' folder"
+                           + (SubfolderInferred ? " (subfolder taken from the draft's parent directory)" : "");
+
+        /// <summary>Fold this draft into a live layer scan: its type folder gains the draft, sorted among the
+        /// live files by filename exactly as <see cref="SkyPatcherDiscovery.Scan"/> sorts them. Returns the live
+        /// scan unchanged when <paramref name="refusal"/> is set. <paramref name="warnings"/> collects the facts
+        /// a replay would otherwise swallow — a gated-off draft applies nothing, and silence there would read as
+        /// "the draft changes nothing".</summary>
+        public SkyPatcherDiscovery.LayerScan Fold(SkyPatcherDiscovery.LayerScan live, SkyPatcherCatalog catalog,
+                                                  Func<string, bool> pluginPresent, out string? refusal,
+                                                  List<string>? warnings = null)
+        {
+            refusal = null;
+            var name = Path.GetFileName(IniPath);
+            int at = -1;
+            for (int i = 0; i < live.Folders.Count; i++)
+                if (live.Folders[i].Subfolder.Equals(Subfolder, StringComparison.OrdinalIgnoreCase)) { at = i; break; }
+
+            // Same relative path once placed = the VFS same-path collision: one copy wins by mod order and the
+            // other is never read, so which body this call would answer with is not the draft's to decide.
+            var clash = at < 0 ? null
+                : live.Folders[at].Files.FirstOrDefault(f => f.SortKey.Equals(name, StringComparison.OrdinalIgnoreCase));
+            if (clash is not null)
+            {
+                refusal = $"the draft's filename '{name}' is already live in the '{Subfolder}' folder as '{clash.RelPath}'"
+                        + (clash.WinningProvider is null ? "" : $" (from '{clash.WinningProvider}')")
+                        + ", so placing the draft there would shadow that file or be shadowed by it depending on mod order — rename the draft, or nest it under a mod-specific subfolder.";
+                return live;
+            }
+
+            IReadOnlyList<SkyPatcherLine> lines;
+            try { lines = SkyPatcherParse.ParseFile(File.ReadAllText(IniPath)); }
+            catch (Exception ex)
+            {
+                refusal = $"the draft '{IniPath}' could not be read: {ex.Message}";
+                return live;
+            }
+
+            // The filename gate applies to the draft as to any file: '<Plugin>.esp.ini' loads only when that
+            // plugin is active.
+            var gate = SkyPatcherDiscovery.GatePluginOf(name);
+            string? notApplied = null;
+            if (gate is not null && !pluginPresent(gate))
+            {
+                notApplied = $"filename-gated on plugin '{gate}', which is not in the active load order";
+                warnings?.Add($"{IniPath}: the draft is {notApplied} — SkyPatcher would not read it once placed, so this post state is the plain winner.");
+            }
+
+            var draft = new SkyPatcherDiscovery.IniFile(IniPath, Subfolder, name, DraftProvider,
+                                                        Array.Empty<string>(), gate, notApplied, lines);
+
+            var folders = new List<SkyPatcherDiscovery.FolderScan>(live.Folders);
+            if (at < 0)
+            {
+                folders.Add(new SkyPatcherDiscovery.FolderScan(Subfolder, catalog.ForSubfolder(Subfolder), true,
+                                                               new List<SkyPatcherDiscovery.IniFile> { draft }));
+                folders = folders.OrderBy(f => f.Subfolder, StringComparer.OrdinalIgnoreCase).ToList();
+            }
+            else
+            {
+                var files = folders[at].Files.Append(draft)
+                                             .OrderBy(f => f.SortKey, StringComparer.OrdinalIgnoreCase).ToList();
+                folders[at] = folders[at] with { Files = files };
+            }
+
+            var notes = live.Notes.Append(
+                $"the draft INI '{IniPath}' was folded into the '{Subfolder}' folder as '{name}' — it is NOT on disk in a mod, " +
+                "so this reads what the game would see once it is placed there.").ToList();
+            return new SkyPatcherDiscovery.LayerScan(folders, notes, live.ReadIncomplete);
+        }
+    }
+
+    /// <summary>What the draft's provider column says: it comes from no mod, and saying so keeps a render from
+    /// naming a mod folder that does not have it.</summary>
+    public const string DraftProvider = "the draft (not placed in a mod)";
+
+    /// <summary>Validate a draft pole expression into a <see cref="Plan"/>. Returns the refusal sentence, or null
+    /// on success. <paramref name="subfolder"/> may be null, in which case the file's parent directory name is
+    /// used when it is a documented type folder and the call is refused when it is not.</summary>
+    public static string? Prepare(string? iniPath, string? subfolder, SkyPatcherCatalog catalog, out Plan? plan)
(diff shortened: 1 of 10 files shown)


src/housecarl-core/SkyPatcherDraft.cs
```diff
@@ -0,0 +1,136 @@
...
+                warnings?.Add($"{IniPath}: the draft is {notApplied} — SkyPatcher would not read it once placed, so this post state is the plain winner.");
+            }
+
+            var draft = new SkyPatcherDiscovery.IniFile(IniPath, Subfolder, name, DraftProvider,
+                                                        Array.Empty<string>(), gate, notApplied, lines);
+
+            var folders = new List<SkyPatcherDiscovery.FolderScan>(live.Folders);
+            if (at < 0)
+            {
+                folders.Add(new SkyPatcherDiscovery.FolderScan(Subfolder, catalog.ForSubfolder(Subfolder), true,
```
Avick3110: A draft folded into a type folder that is not already live ignores the `SkyPatcher.ini` `[Patcher]` toggle, so a draft in a disabled type reads as applied when the game would skip it.

`PatchingEnabled` is hardcoded `true` here. The toggles are read only inside `SkyPatcherDiscovery.Scan` (`ReadPatcherToggles` / `ToggleEnabled`, SkyPatcherDiscovery.cs:154-168), and they are applied only to folders built from `byFolder` — folders that exist because at least one INI was found there. A type with no live INI produces no `FolderScan` at all, so the toggle for it is never consulted and this `Fold` invents the folder as enabled.

Concrete failure: the install's `SKSE/Plugins/SkyPatcher.ini` contains `iEnableWeaponPatching=0`, and no enabled mod ships any `SkyPatcher/weapon/*.ini`. Call `housecarl_records` with `formids=["<a WEAP>"]`, `source={"overlay":"skypatcher","state":"post","ini":"C:/work/MyWeapons.ini","subfolder":"weapon"}` and a draft line `filterByWeapons=<edid>:attackDamage=123`. The response reports `BasicStats.Damage = 123`. Once the draft is placed, the DLL skips the whole `weapon` subfolder and the value stays at the winner's — the answer is wrong in exactly the direction the verify step is meant to catch, and it is silent about it. `OrderedLines` gates on `PatchingEnabled` (SkyPatcherDiscovery.cs:189), so passing the real toggle through here is what decides it.

Related, same root: when the type folder *does* exist and is toggled off, the draft is appended to the disabled folder and correctly contributes nothing — but the reason lives in `LayerScan.Notes`, which the records lane does not render. That is the "the draft changes nothing" silence the rest of this PR set out to remove; the gate refusal gets a warning here, the toggle case gets none.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1052,6 +1052,17 @@ internal static (NativeProvenance Provenance, NativePairingRung? Rung, string? P
...
 
+    /// <summary>Carry one replay's warnings (unknown key, unmapped op, unresolved filter, parse note) into the
+    /// caller's sink, deduplicated: a line the layer could not apply belongs beside the answer, not in silence.
+    /// Each warning already names its own file and line, so a draft's warnings name the draft's path.</summary>
+    static void CollectOverlayWarnings(IReadOnlyList<SkyPatcherFolderOutcome> folders, List<string>? sink)
+    {
+        if (sink is null) return;
+        foreach (var f in folders)
+            foreach (var w in f.Result?.Warnings ?? Array.Empty<string>())
+                if (!sink.Contains(w)) sink.Add(w);
```
Avick3110: `sink.Contains(w)` is a linear scan over a list that now grows with the batch size, and the overlay replay emits one *distinct* warning per record for the commonest warning there is — so a large `formids=` read under an overlay post source goes quadratic.

`SkyPatcherOverlay.Apply` line 151 emits `"{file}:{line}: line skipped — carries filter(s) with no static evaluation ({names}); whether it applies to {Ident(fk, editorId)} is UNRESOLVED"`. `Ident` is the record's FormKey and EditorID, so every record in the batch yields a different string for the *same* INI line — nothing dedupes. `CollectOverlayWarnings` is called once per record inside `OverlayPostBatch`'s loop and inside the delta pole reader, so the sink reaches N entries and costs N^2/2 comparisons. Worse, all N strings share the same ~180-character prefix (same file, same line number), so each comparison runs the full prefix before it differs.

Concrete failure: an order whose SkyPatcher layer has one line carrying a filter with no static evaluation (`LinesSkippedUnresolvedFilter` exists precisely because this is ordinary). Call `housecarl_records` with `formids=["@ids.jsonl"]` holding 50,000 ids of that type, `source={"overlay":"skypatcher","state":"post"}`, `project={"form":"fields",...}`. The list bound allows it (`RenderBudget.MaxRenderRows` is 300,000 on this lane). The call now spends ~1.25e9 string comparisons over ~180-char prefixes purely on the dedup, and holds ~50,000 x ~250 chars of warning text, to render at most 20 of them. Before this PR the sink was null and `CollectOverlayWarnings` returned immediately, so this is a regression on the plain overlay-post lane, not only on the draft path.

A `HashSet<string>` for the membership test, and stopping collection once the cap plus an overflow counter is reached, both fix it without changing what is rendered.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -992,7 +1010,8 @@ List<KeyValuePair<string, string>> Echo()
             if (form == "delta")
             {
                 var rows = svc.DeltaBatch(ids, srcSpec, versusSpec!, projFields, demand,
-                                          out var sArm, out var rArm, out var covers, out var refusal, out var epoch);
+                                          out var sArm, out var rArm, out var covers, out var refusal, out var epoch,
+                                          overlayWarnings);
```
Avick3110, line 1013: The delta lane gets the sink, but the tree lane nine lines down does not — so a draft read through `form='tree'` swallows every replay warning, which is the failure this PR set out to remove.

`svc.TreeBatch(ids, versusSpec!, ...)` (line 1023) calls `MakePoleReader` internally (LoadOrderService.cs:3747) with the new `overlayWarnings` parameter left at its default `null`, and `TreeResponse` never calls `StateOverlayWarnings()`. Both halves have to change for the tree to say anything.

An overlay pole is reachable on this lane: `ScanLane` refuses `versusSpec.Kind == Overlay` in both its arms (lines 1168 and 1591), but the `formids=` tree branch refuses only a non-winner `source=`. So this call is accepted:

`housecarl_records` with `formids=["<a WEAP>"]`, `project={"form":"tree","fields":["BasicStats.Damage"]}`, `versus={"overlay":"skypatcher","state":"post","ini":"C:/work/Typo.ini","subfolder":"weapon"}` where the draft's one line is `filterByWeapons=<edid>:attakDamage=5`.

The draft is folded in, the line is skipped for the unknown key, the tree renders every provider as identical to the reference, and no warning appears anywhere. Same inputs on `form='fields'` produce the warning that `ADraftLineWithAnUnknownKeyWarnsAndNamesTheDraftsPath` asserts. The gated-on-an-inactive-plugin warning from `Plan.Fold` disappears on this lane too, so a draft named for a plugin not in the order reads as "changes nothing" with no reason given.

src/housecarl-core/SkyPatcherDraft.cs
Avick3110: Real, and fixed in c0c0fad.

Reproduced with a test before the fix: the world's `SkyPatcher.ini` now carries `[Patcher] iEnableArmorPatching=0` and nothing is placed in `armor`, so a draft folded there is exactly the invented folder. Pre-fix the read answered `ArmorRating = 99` with no warning; post-fix it answers the winner's value and says the toggle skipped the folder.

The toggle map now rides `LayerScan` (`PatcherToggles`) rather than staying local to `Scan`, `ToggleEnabled` is public, and `Fold` takes the folder's enabled flag from the live `FolderScan` when there is one and from the map when it invents one. The second half is covered too: enabled or not, when the draft's folder is toggled off the fold adds a warning naming the toggle, so the toggle case now speaks where only the gate case did.

Test: `ADraftInATypeToggledOffInSkyPatcherIniAppliesNothingAndSaysSo`.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real, and fixed in b325a55.

Measured it: 50,000 distinct warnings sharing a 180-character prefix take 8.8s through the old `List` + `Contains`, and 38ms through the fix. The cost test bounds it at 3s, so it fails before and passes after.

The sink is now a small class, `SkyPatcherOverlay.WarningSink`: a `HashSet` membership test, twenty warnings kept (the cap moved out of `RecordsTools`), and the rest counted. Once the cap is reached the seen set stops growing too, so neither the kept list nor the set grows with the batch.

One consequence worth naming: past the cap, `Overflow` counts further warnings rather than further *distinct* ones — deduplicating them exactly is what would need the 50,000 strings kept. It is only ever rendered as "there were more" ("+N more not listed"), so the render is unchanged in what it claims; the class says this out loud.

Tests: `SkyPatcherWarningSinkTests` (cap, dedupe, and the cost bound).

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, and fixed in e81047a.

Both halves changed, as you said: `TreeBatch` takes the sink and hands it to `MakePoleReader`, and `TreeResponse` calls `StateOverlayWarnings()`. The two scan-lane `TreeBatch` calls are left alone — that lane refuses an overlay pole at both arms.

Reproduced before the fix with your exact call shape: `form='tree'` with an overlay `versus=` carrying a draft whose one line has a misspelled key rendered every provider as identical with nothing said, and the gate case read as "changes nothing" the same way. Both fail without the render and pass with it.

Tests: `ADraftLineTheLayerSkipsIsNamedOnTheTreeFormToo` (the replay's own warning) and `ADraftGatedOnAnInactivePluginSaysSoOnTheTreeFormToo` (the fold's, which comes from a different place).

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1052,6 +1052,17 @@ internal static (NativeProvenance Provenance, NativePairingRung? Rung, string? P
         return (typeName, winner.Value.WinnerPlugin, body.EditorID, folders, null, copy);
     }
 
+    /// <summary>Carry one replay's warnings (unknown key, unmapped op, unresolved filter, parse note) into the
+    /// caller's sink, deduplicated: a line the layer could not apply belongs beside the answer, not in silence.
+    /// Each warning already names its own file and line, so a draft's warnings name the draft's path.</summary>
```
Avick3110, line 1057: **Filter-class warnings carry no file or line, so a draft's warnings cannot be told apart from a live INI's.**

The doc line says "Each warning already names its own file and line, so a draft's warnings name the draft's path", but that only holds for the warnings `Apply` raises directly, which prefix `where = $"{line.File}:{line.LineNumber}"`. The filter-evaluation warnings are built without it: `SkyPatcherOverlay.cs:334` (`filter '{cls.BaseKey}' has no static evaluation — {spec.Unmapped}`), `:417` (`has no evaluation (neither built-in nor in the filter map)`), and the `:364` / `:386` / `:393` connective and `modNamesLastOverridden` warnings. `EvaluateFilters` / `EvaluateOneFilter` are never passed `where`.

Scenario: a modder checks a draft with `source={"overlay":"skypatcher","state":"post","ini":"C:/work/draft.ini","subfolder":"weapon"}`. The draft carries an unmapped filter, and so does an already-placed INI in the same folder. Both replays produce the byte-identical string, so `WarningSink` dedupes them to a single rendered line:

```
[!] skypatcher: filter 'filterByAlternateTextures' has no static evaluation — …
```

Nothing in that line says whose it is. The modder cannot tell whether their draft is the broken one or someone else's INI is — which is the question the draft pole exists to answer. Since the whole point of this hunk is to stop a skipped draft line reading as "the draft changes nothing", the attribution gap matters most exactly here. Either thread `where` down into the filter evaluators, or have `CollectOverlayWarnings` tag each warning with the file it came from when the source line doesn't.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3489,6 +3507,11 @@ PoleReader MakeOverlayPoleReader(LoadOrderResolver.IndexView view, LoadOrderReso
                     fieldMap = SkyPatcherFieldMap.Load();
                     catalog = SkyPatcherCatalog.Load();
                     scan = SkyPatcherDiscovery.Scan(assets, catalog, view.ContainsPlugin, _skyPatcherParseCache);
+                    if (spec.Draft is not null)
+                    {
+                        scan = spec.Draft.Fold(scan, catalog, view.ContainsPlugin, out var draftRefusal, overlayWarnings);
+                        if (draftRefusal is not null) { setupError = draftRefusal; return new PoleReading(null, null, null, setupError); }
```
Avick3110: **A draft refusal is a per-record error on the delta/tree lane, and `counts_only` never renders it.**

`OverlayPostBatch` (line 3645) turns the same `Fold` refusal into the call's `refusal`, so the `formids=` read lane refuses by name. Here it only reaches `setupError`, which becomes each row's `PoleReading` error and then `DeltaRow.Error`.

Scenario: `source={"overlay":"skypatcher","state":"post","ini":"C:/work/placed.ini","subfolder":"weapon"}` where `placed.ini` already exists in the live `weapon` folder, `versus="winner"`, `project={"form":"delta"}`, `counts_only=true`. `DeltaResponse` takes the `counts_only` branch, which renders only

```
count=1 differing=0 identical=0 no_verdict=0 errors=1
```

The per-row error text is never rendered on that path, so the caller is told the comparison ran and produced one error, with no mention of the filename collision — a silently degraded mode. Without `counts_only` it still reads as N failed rows rather than the refusal the PR body describes.

The PR's own composition (`source=` draft post vs `versus=` plain post) is exactly the delta lane, and `RecordsSkyPatcherDraftTests` only exercises the collision through `ReadW0`, so this divergence is untested. Suggest surfacing the fold refusal through `MakePoleReader`'s `error` out-param so the delta and tree lanes refuse the way the read lane does.

src/housecarl-core/SkyPatcherDraft.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            refusal = null;
+            var name = Path.GetFileName(IniPath);
+            int at = -1;
+            for (int i = 0; i < live.Folders.Count; i++)
+                if (live.Folders[i].Subfolder.Equals(Subfolder, StringComparison.OrdinalIgnoreCase)) { at = i; break; }
+
+            // Same relative path once placed = the VFS same-path collision: one copy wins by mod order and the
+            // other is never read, so which body this call would answer with is not the draft's to decide.
+            var clash = at < 0 ? null
```
Avick3110, line 57: **A draft path that is itself an already-placed, nested INI is folded in alongside the live copy — its lines apply twice.**

The clash test compares the draft's bare `Path.GetFileName` against live `SortKey`s. `SkyPatcherDiscovery.Scan` sets `SortKey` to the path *relative to the type folder*, so a nested live file at `SKSE/Plugins/SkyPatcher/weapon/MyMod/Blades.ini` has `SortKey = "MyMod\Blades.ini"`.

Scenario: the modder edits their INI in place inside the mod folder (exactly what the refusal below recommends — nesting under a mod-specific subfolder) and then checks it with `ini="<that absolute path>", subfolder="weapon"`. `name` is `"Blades.ini"`, which matches no live `SortKey`, so no clash is raised and the file joins the folder a second time. `ReplaySkyPatcher` then replays every one of its lines twice: a plain SET is idempotent by luck, but an `AddEntry`-class op appends the entry twice and `Mult` / `AddNumeric` compound, so the reported post state is one the game would never produce — a silently wrong answer rather than a named refusal.

Neither `Prepare` nor `Fold` compares `IniPath` against the layer's winning loose paths. The cheapest guard is probably in `Prepare`: refuse a draft whose normalized path is under the Data-side `SKSE\Plugins\SkyPatcher` tree at all, naming that it is already placed and should be read with the plain post state.

src/housecarl-core/SkyPatcherDraft.cs
```diff
@@ -0,0 +1,147 @@
...
+                : live.Folders[at].Files.FirstOrDefault(f => f.SortKey.Equals(name, StringComparison.OrdinalIgnoreCase));
+            if (clash is not null)
+            {
+                refusal = $"the draft's filename '{name}' is already live in the '{Subfolder}' folder as '{clash.RelPath}'"
+                        + (clash.WinningProvider is null ? "" : $" (from '{clash.WinningProvider}')")
+                        + ", so placing the draft there would shadow that file or be shadowed by it depending on mod order — rename the draft, or nest it under a mod-specific subfolder.";
```
Avick3110: **Both remedies in this refusal fail: one cannot be expressed on the pole, the other changes the answer.**

"nest it under a mod-specific subfolder" is unreachable. `Prepare` resolves `subfolder` through `SkyPatcherCatalog.ForSubfolder`, an exact dictionary lookup (`SkyPatcherCatalog.cs:63`), so `subfolder="weapon/MyMod"` (or the backslash spelling) is refused as "not a documented SkyPatcher record type", and `Fold` always gives the draft a bare-filename `SortKey`. A caller who follows the advice hits a second refusal with no path forward.

"rename the draft" is reachable but answer-changing: files sort by name within a type folder, so a renamed draft sits at a different point in the apply order than the file the modder actually intends to place. The post state read back is then not the one the real placement produces — the specific thing this pole promises.

Given the surface's error contract ("what went wrong and what to try"), either accept a nested relative placement as the `subfolder` value (`"weapon/MyMod"`, with the draft's `SortKey` becoming `MyMod\<name>`), or drop the nesting half of the sentence and say the draft must be given a name that is not yet placed, noting that the name decides its sort position.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1819,7 +1840,25 @@ static bool TryEditorIdContains(string clause, out string? text)
...
+                {
+                    if (st.Equals("pre", StringComparison.OrdinalIgnoreCase))
+                        return $"error: {param}= overlay state \"pre\" IS the plain load-order winner, the body the INI layer starts from, so a draft INI cannot change it — read the draft with state \"post\".";
+                    if (SkyPatcherDraft.Prepare(draftIni, draftSub, SkyPatcherCatalog.Load(), out plan) is { } derr)
+                        return $"error: {param}= {derr}";
+                }
+                spec = new LoadOrderService.PoleSpec(LoadOrderService.PoleKind.Overlay, OverlayState: st.ToLowerInvariant(), Draft: plan);
                 return null;
             }
             if (!e.TryGetProperty("file", out var fEl) || fEl.ValueKind != JsonValueKind.String)
```
Avick3110, line 1869: **`ini` / `subfolder` on the `{"file", "mod"}` pole are accepted and silently ignored.**

The two new keys are only read inside the `overlay` branch above. An object pole that carries them without `"overlay"` falls straight through to this `file` branch and builds a plain `PoleKind.Named` spec.

Scenario: `source={"file": "MyMod.esp", "ini": "C:/work/draft.ini", "subfolder": "weapon"}` — a natural mix-up, since the draft is *for* that plugin. The call returns `MyMod.esp`'s own record with no SkyPatcher layer and no draft; neither the arm line nor the envelope mentions that a draft was named and dropped, so the answer reads as the draft's post state.

This is the "accepted and ignored" failure the surface refuses elsewhere by name — including in the overlay branch itself, which already refuses `"subfolder"` with no `"ini"`, and in this branch, which refuses a missing or non-string `"file"`. A one-line guard before the `file` lookup (refuse an object carrying `"ini"` or `"subfolder"` without `"overlay"`, naming that the draft rides the SkyPatcher overlay pole) would close it. The error sentence on line 1865 could also grow the draft keys, since it is now the message a caller sees after mis-spelling the overlay form.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real, and fixed in 99cb5d5.

Confirmed by reading: every `warnings.Add` inside `Apply`'s line loop carries `where`, and every one reached through `EvaluateFilters` does not — the whole `dedupe`-guarded set, the six you named included. The `CollectOverlayWarnings` doc line claimed otherwise, so it was wrong for exactly the warnings the draft pole most needs attributed.

The `(List<string> warnings, HashSet<string> dedupe)` pair the filter chain threaded is now one collector, `FilterWarnings`. `Apply` tells it the file and `file:line` at the top of each line iteration; `Add(key, text)` prefixes the text with `file:line` and scopes the dedupe key to the file. So a draft and a placed INI raising the same filter warning produce two distinct strings, each naming its own path, and the `WarningSink` no longer folds them into one.

The dedupe scope is the one behaviour change worth naming: it was once per replay, it is now once per INI per key. A filter no evaluator handles is still named once per file rather than once per line, so a file with fifty such lines still warns once — but two files now warn twice, which is the point.

Test: `AFilterTheLayerCannotEvaluateWarnsWithTheDraftsPathAndLine`, on `filterByHasAmmoFromWeaponList` (the weapon folder's one explicitly unmapped filter). It asserts the warning reads `<draft path>:1: filter 'filterByHasAmmoFromWeaponList'`; before the fix the response carries the line-level skip with that prefix but the filter warning without one, so it fails.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real, and fixed in 8b547b9.

Reproduced with your call shape: `source=` a draft whose filename collides with the placed INI, `versus="post"`, `form='delta'`, `counts_only=true` returned `count=1 differing=0 identical=0 no_verdict=0 errors=1` and never said what the error was. The tree lane did the same through `versus=`.

The cause was the timing, as you said: `MakeOverlayPoleReader` built its whole replay context on the first record, so the fold's refusal could only ever be a `PoleReading` error. That setup is now a local `Setup()`, and a pole carrying a draft calls it up front, before the reader is handed back — whether a draft can be folded at all is a fact about the call, not about a record. A refusal there sets the `error` out-param, which `DeltaBatch` and `TreeBatch` already turn into `"source: …"` / `"versus: …"` on the call. Laziness is unchanged for a plain overlay pole: no draft, no eager setup.

Tests: `ADraftTheFoldRefusesRefusesTheDeltaCall` (with `counts_only=true`, the shape that hid it) and `ADraftTheFoldRefusesRefusesTheTreeCall`, since the tree lane maps the same out-param at its own call site. Both fail before the fix.

src/housecarl-core/SkyPatcherDraft.cs
Avick3110: Real, and fixed in b559619.

The reading holds: `Scan` sets `SortKey` to the path relative to the type folder, so a nested live INI sorts under `MyMod/Blades.ini` and the bare-filename clash test never sees it. Pointing the pole at that file folded it in beside itself, and the replay walked its lines twice.

I did not take the path-shape guard, because it over-refuses: a draft sitting in a DISABLED mod's own `SKSE/Plugins/SkyPatcher/weapon` folder is not in the layer, and folding it is exactly the check-before-enabling case the pole is for. The guard is identity instead. `IniFile` now carries `LooseFilePath` — the winning loose copy's path, null for a BSA-only file — and `Fold` refuses when the draft's normalized path is one of them, naming the relative path the layer reads it as and pointing at the plain post state. That catches the nested case and the flat one under the same sentence, ahead of the filename clash, which is now only about two different files wanting the same name.

Test: `ADraftThatIsAlreadyPlacedIsRefusedRatherThanFoldedInTwice`. `RecordsWorld` gains a second placed INI, `weapon/MyMod/nested.ini`, comment-only like the first, so it is a live file whose sort key is a relative path and no existing test's post state moves.

src/housecarl-core/SkyPatcherDraft.cs
Avick3110: Half real. The nesting half is fixed in 2b24340; I am refuting the renaming half.

**Nesting: real.** `ForSubfolder` is an exact dictionary lookup, so `subfolder="weapon/MyMod"` is refused as an undocumented record type and `Fold` always gives the draft a bare-filename sort key. The sentence advised something the pole cannot express. It now reads:

> … — give the draft a filename not yet placed in that folder, which is also what decides where it sorts among them.

One sentence, one remedy, and it names the sort consequence rather than leaving the caller to find it.

**Renaming changes the answer: not real, in this case.** The refusal only fires when the draft's filename is ALREADY LIVE at that relative path. That name is not one the modder can place: putting their file there is the same-path VFS collision the refusal is about — one copy wins by mod order and the other is never read. So there is no "the file the modder actually intends to place" under that name to be measured against; whatever name they end up placing it as is a different name either way, and checking under that name is the correct read, not a drifted one. The advice asks them to decide the name first, which they have to do regardless.

I did not take the other option you offered — accepting `"weapon/MyMod"` and giving the draft a nested sort key. It is a real capability the pole lacks (a draft that would be placed nested cannot be expressed at all), but it is a surface expansion rather than a fix to a wrong answer, so it is Aaron's call, not mine to fold into a review pass.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, and fixed in d5f9413.

Confirmed by reading: the two keys are read only inside the `overlay` branch, so an object without `"overlay"` fell through to the `file` branch and built a plain `PoleKind.Named` spec with the draft dropped, silently.

The object branch now refuses an object carrying `"ini"` or `"subfolder"` without `"overlay"`, before the `file` lookup, and names the pole they ride:

> error: source= names "ini"/"subfolder" without "overlay" — a draft INI is a value on the SkyPatcher overlay pole, so pass {"overlay": "skypatcher", "state": "post", "ini": "<absolute path>", "subfolder": "<type folder>"}.

The structured-pole sentence below it grew the draft keys too, as you suggested, since it is what a caller sees after misspelling the overlay form.

Test: `TheDraftKeysOnAPluginPoleAreRefusedNamingTheOverlayPole` — `source={"file": "<the override plugin>", "ini": "<a draft>"}`; before the fix it is served, not refused.

PR #619, opened 2026-09-07
Refuse an in-place create over an editorid the target defines

### LINKED ISSUE #611: bulk_create on the in-place lane replaces a record whose editorid the target already defines
Reported by a user (Sanguine's Trade), 2026-09-06, against the current 1.6 MCP build.

### What happened

A `bulk_create` on the in-place lane, into a plugin houseCARL did not author, was given five records. One of them used an editorid the target plugin already defined: `ST_InSceneFaction` (622D39), a faction that is part of the mod's published extension API and referenced by its scripts and docs. The user did not know it existed.

```
housecarl_bulk_create
  target="Sanguine's Trade - An Economy Mod.esp" in_place=true acknowledge=true
  records=[..., {record_type:"Faction", editorid:"ST_InSceneFaction",
                 operations:[{field_path:"Name", value:"Sanguine's Trade In Scene"}]}]
```

The call succeeded. The existing faction was re-created fresh at the same FormID carrying only the one `Name` op; every field it previously had was dropped. The user recovered it from a pre-edit backup.

The output said so plainly:

```
Faction 622D39 ... ST_InSceneFaction [REPLACED: this patch already defined this
editorid — re-created fresh at the same FormID; prior contents, including any
set_field edits since, were discarded]
```

### Why it does this

This is deliberate, and documented at `src/housecarl-core/WritePatchBuilder.cs:251`: a create whose editorid the artifact already defines is a re-run of the same create, so the record is rebuilt from the new spec at a stable FormKey. That is the right semantic for `into=` — a houseCARL patch being regenerated should be idempotent, not accumulate duplicates.

The in-place lane inherits it. There the file is someone else's, its records were not written by houseCARL, and a colliding editorid is far more likely to be a name the user did not know was taken than a re-run of their own create.

### The fix

On `in_place=true`, an editorid the target plugin already defines should refuse the whole call before anything is written, all-or-nothing, naming the record:

```
ST_InSceneFaction is already defined in Sanguine's Trade - An Economy Mod.esp at
622D39. Pass replace=true to overwrite it, or pick another editorid.
```

`replace=true` then does what today's default does. `into=` keeps its current rebuild behaviour unchanged — nothing about the patch-regeneration lane should move.

The report was honest and the user could see exactly what had happened, which is how they recovered. The problem is that the destructive step ran first: a create should not destroy a record on a file houseCARL does not own without being asked to.

### Test

A probe covering both lanes: an in-place create whose editorid collides refuses and leaves the file byte-identical; the same call with `replace=true` replaces and flags it; an `into=` re-run still replaces at a stable FormKey as it does now.

Avick3110:
A create on the in-place lane whose editorid the target plugin already defines used to re-create that record fresh at its own FormID from the call's spec, discarding everything else it held, and report it only afterwards. That rebuild is what makes an `into=` re-run of a houseCARL patch idempotent, so it stays there; on a file houseCARL does not own the colliding name is far likelier one the caller did not know was taken. The create pre-flight now looks the editorid up in the target's own defined records — the same phase that already refuses a missing editorid or an un-createable type — and refuses the whole call before anything is written, naming each colliding record's type and FormID. `replace=true` opts back into the rebuild; it answers this refusal and nothing else, so it is refused off the in-place lane rather than accepted and ignored. The check sits on the flat-create path only, because that is the one that upserts: cells and nested children already refuse or append instead of replacing, and an override the target carries is still refused loud at the upsert, which never replaces one.

Second commit: the bracket the render prints on a replace said "this patch already defined this editorid". In place the file is the caller's own plugin, so it names the file instead.

Tests: `InPlaceCreateEditoridClashTests` — the collision refuses and the target stays byte-identical (with a non-colliding sibling spec in the same batch, so all-or-nothing is what is measured); `replace=true` overwrites, flags it, and keeps the record's FormID; a free editorid still writes and leaves the existing record alone; an `into=` re-run still replaces at a stable FormID; `replace=` off the in-place lane is refused by name. Full suite 1774 passed, `ci-all` 127/127.

Closes #611


src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2944,6 +2959,18 @@ static CreateOutcome CreateRecordsCore(
...
+                // In place, a flat create whose editorid the target already defines would rebuild that record from this
+                // spec at its own FormID and discard everything else it held. That is right under into=, where the
+                // artifact is houseCARL's own and a re-run should be idempotent; on a file houseCARL does not own the
+                // name is far likelier one the caller did not know was taken. Refused before anything is written;
+                // replace= opts back in.
+                else if (inPlace && !replaceExisting && DefinedInTarget(s.EditorId) is { } clash)
+                {
+                    problems.Add($"{s.RecordType} '{s.EditorId}': {fileName} already defines "
+                        + $"{RecordNaming.StripOverlay(clash.GetType().Name)} {clash.FormKey.ID:X6} with that editorid. "
+                        + "Pass replace=true to overwrite it, or pick another editorid.");
```
Avick3110: The refusal offers `replace=true` for collisions that `replace=true` cannot resolve, so following the message lands on a second refusal.

`DefinedInTarget` matches on editorid alone (any record the target defines, any type). `GenericUpsertNew` — the only thing `replace=true` unlocks — will only replace when the existing record's type also matches, and otherwise throws the cross-type refusal at WriteEngine.cs:1113.

Repro: target `X.esp` defines a Keyword `MyThing`. Call

```
housecarl_create in_place="X.esp" acknowledge=true
  records=[{record_type:"Faction", editorid:"MyThing", ops:[{field_path:"Name", value:"…"}]}]
```

* Now: `Faction 'MyThing': X.esp already defines Keyword 000801 with that editorid. Pass replace=true to overwrite it, or pick another editorid.`
* Re-run with `replace=true`: `upsert refused: existing record 'MyThing' (000801:X.esp) is a Keyword, not a Faction — an EditorID collision across record types is a real authoring error…`

Before this change the caller got that accurate cross-type sentence on the first call; now they get a remedy that cannot work first, and the accurate one only after a second in-place call (which, on a first touch of the plugin, also costs a fresh `acknowledge=` since a refused call records nothing).

The same shape applies to the two other collisions the upsert refuses rather than absorbs: a carried override under the same editorid (filtered out of `definedHere` by the `FormKey.ModKey == patchMod.ModKey` clause, so that one still falls through correctly) and duplicate-editorid residue — `DefinedInTarget` reports `g.First()` and offers `replace=true`, but `matches.Count > 1` throws "which survives is your call" instead.

Cheapest fix: make the clash test type-aware — only offer `replace=true` when the clashing record is actually replace-eligible (same type, target's own ModKey, exactly one match), and say "pick another editorid" alone otherwise. There is no test for a cross-type clash in `InPlaceCreateEditoridClashTests`, which is why this is invisible in the suite.

Avick3110: Real, and fixed in 25feea26.

Confirmed by reading both sides: `DefinedInTarget` matched on editorid alone and took `g.First()`, while `GenericUpsertNew` only replaces when the stored record's concrete type matches, is exactly one, and is the mod's own. So the offer and the outcome could disagree three ways, not two — the mixed case you'd get from a plugin that both defines a record named X and carries an override named X passed the `ModKey` filter with one match and was offered `replace=true` too, then hit the override refusal.

Rather than mirror the eligibility test in the pre-flight, the decision now lives in one place: `WriteEngine.UpsertWouldReplace(patchMod, typeName, matches)`. The upsert gates its own replace on it (its two identical cross-type throws collapsed into one), and the pre-flight asks it before offering anything. The pre-flight index dropped the `ModKey` filter so it judges the same match set the upsert does, and words each refusal for its own lane:

- cross-type: "... already defines Keyword 000801 with that editorid — an editorid collision across record types, which no overwrite resolves: a Keyword cannot be re-created as a Faction. Pick another editorid."
- duplicates: names both FormIDs and points at `housecarl_remove`.
- carried override: points at `housecarl_apply`.

None of the three names `replace=`, and all three now refuse in the pre-flight whether or not `replace=true` was passed, so the accurate sentence arrives on the first call in every case.

Tests, both failing before the fix: `ACrossTypeClashIsRefusedWithoutOfferingReplace` (target defines a Keyword, call creates a Faction — refusal names Keyword, does not contain `replace=`) and `ADuplicateEditoridClashIsRefusedWithoutOfferingReplace` (names `housecarl_remove`, no `replace=`). I dropped a third I had written for `replace=true` over a cross-type clash: it passes before the fix too, since the upsert threw its own accurate sentence there and wrote nothing either way, so it is not a test.

Full suite 1776 passed, `ci-all` all pass. `housecarl_create`'s `replace=` description and the changelog entry now say the same thing.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2923,6 +2926,39 @@ static CreateOutcome CreateRecordsCore(
...
+                     + $"override would blank the original plugin's record. Edit it with {ToolNames.Apply}, or pick another editorid.";
+            if (clash.Count > 1)
+                return $"{fileName} defines {clash.Count} records with that editorid "
+                     + $"({string.Join(", ", clash.Select(c => c.FormKey.ID.ToString("X6")))}) — external references may point at "
+                     + $"either copy, so which survives is your call: remove the extra(s) with {ToolNames.Remove}, then re-run.";
+            var one = clash[0];
+            var itsType = RecordNaming.StripOverlay(one.GetType().Name);
+            return WriteEngine.UpsertWouldReplace(patchMod, wantType, clash)
+                ? $"{fileName} already defines {itsType} {one.FormKey.ID:X6} with that editorid. "
+                  + "Pass replace=true to overwrite it, or pick another editorid."
```
Avick3110: **The `replace=true` this line offers destroys a DialogTopic's or Worldspace's child records.**

`replace=true` routes into `WriteEngine.GenericUpsertNew`, whose replace arm is `InvokeRemove(group!, formKey)` then `InvokeAddNewWithFormKey` (WriteEngine.cs:1105-1113) — no `CaptureChildGroup` / `RestoreChildGroup`. `ChildGroupCarry`'s own doc names the three types that own children (WriteEngine.cs:541): Cell / DialogTopic / Worldspace, and the forward lanes go to explicit lengths to stop exactly this loss ("without this every child under the replaced record is destroyed while the call reports success: INFOs under a DIAL, placed refs under a CELL", WriteEngine.cs:528-529).

`DialogTopic` is a flat `SkyrimGroup<T>`, so it reaches `GenericUpsertNew`. Concrete run:

1. A user authors a dialogue unit into their own plugin — `housecarl_create in_place=MyMod.esp` with a `DialogTopic` "MyTopic" plus its INFO children (exactly probe arm P, and what the dialogue-authoring skill drives).
2. They re-run the same call.
3. This pre-flight refuses: *"MyMod.esp already defines DialogTopic 000801 with that editorid. Pass replace=true to overwrite it, or pick another editorid."*
4. They pass `replace=true`. The topic is removed from its group and re-added fresh — **every INFO under it is gone from their own plugin**, serialized in place.

The only notice is the after-the-fact `[REPLACED: … prior contents … were discarded]` bracket, by which point the file is already rewritten. Same shape for a `Worldspace` and its cells.

Either run the forward lane's capture/restore through the upsert's replace arm, or don't offer `replace=` for a record that owns children — refuse it the way the cross-type and duplicate collisions are refused, naming the children that would be lost.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2944,6 +2980,17 @@ static CreateOutcome CreateRecordsCore(
                     cellKinds[i] = CellCreate.Interior;
                 }
                 else if (!WriteEngine.CanCreateType(s.RecordType, out var why)) { problems.Add($"{s.RecordType} '{s.EditorId}': {why}"); continue; }
+                // In place, a flat create whose editorid the target already defines would rebuild that record from this
+                // spec at its own FormID and discard everything else it held. That is right under into=, where the
+                // artifact is houseCARL's own and a re-run should be idempotent; on a file houseCARL does not own the
+                // name is far likelier one the caller did not know was taken. Refused before anything is written;
+                // replace= opts back in, but only over a collision the upsert would actually overwrite.
+                else if (inPlace && CarriedUnder(s.EditorId) is { Count: > 0 } clash
```
Avick3110: **A parentless Cell create never reaches this guard, and the cell-side check it defers to is narrower than the PR body claims.**

This is the third arm of the `else if` chain, so `IsCellType(s.RecordType)` at line 2977 takes an interior cell out before the collision check runs. The PR body says cells "already refuse", but the only cell-side check is `EnsureNoDuplicateEditorId(patchMod, typeof(Cell), editorId)` (WriteEngine.cs:1554), which matches **only Cell-typed records** and compares with `StringComparison.Ordinal`.

Two holes, both landing in the caller's own plugin with no refusal:

- Cross-type. `HcUser.esp` defines a Keyword "MyPlace"; `housecarl_create in_place=HcUser.esp records=[{"record_type":"Cell","editorid":"MyPlace"}]` writes a Cell under an editorid the target already holds.
- Case. An existing Cell "myplace" does not stop a create of "MyPlace", even though this guard and `GenericUpsertNew` both match `OrdinalIgnoreCase`.

The second one also hits the `into=` lane, so it is older than this PR — but a plain `CarriedUnder(s.EditorId)` check ahead of the cell branch would close both for the lane this PR is about.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1089,40 +1089,54 @@ public static (IMajorRecord Record, bool Replaced) GenericUpsertNew(SkyrimMod pa
...
+                    var freshArm = AddConcreteArmToGroup(armGroup!, armType!, formKey, editorId);
                     return (freshArm, true);
                 }
 
-                object? group = null; Type? tMajor = null;
-                foreach (var (prop, tm, _) in EnumerateFlatGroups(patchMod.GetType()))
-                    if (string.Equals(tm.Name, typeName, StringComparison.OrdinalIgnoreCase)) { group = prop.GetValue(patchMod); tMajor = tm; break; }
-                if (group is null || tMajor is null)
+                if (!TryResolveFlatGroup(patchMod, typeName, out var group, out var tMajor))
                     throw new InvalidOperationException($"upsert: no flat group found for type '{typeName}'.");
```
Avick3110: This throw is now unreachable, and the case it covered reports as something else.

`UpsertWouldReplace` returns false for a fourth reason its own doc doesn't list: neither an abstract arm nor a flat group resolves for `typeName`. Because the guard at line 1095 now runs first, that case throws *"existing record 'E' (…) is a Faction, not a X — an EditorID collision across record types"* — pointing the caller at their editorid rather than at a group-resolution fault — and control never gets here.

`CanCreateType` (line 969) walks the same two enumerations a few lines above, so nothing reaches it today; this is latent, not live. But the dead throw reads as live coverage, and if those enumerations ever diverge the wrong message is what a user sees. Worth either folding the group-resolution failure into `UpsertWouldReplace`'s doc + a distinct message, or dropping the now-dead line.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2923,6 +2926,39 @@ static CreateOutcome CreateRecordsCore(
...
             return carried.TryGetValue(fk, out var rec) ? rec : null;
         }
+        // The target's records by editorid, indexed once — the in-place collision pre-flight below asks per spec, and
+        // enumerating the destination each time is the same O(specs x records) the parent index avoids. The WHOLE match
+        // set, carried overrides included, because that is the set the upsert judges. Lazy: nothing is built off the
+        // in-place lane.
+        Dictionary<string, List<IMajorRecord>>? carriedByEdid = null;
+        IReadOnlyList<IMajorRecord> CarriedUnder(string editorId)
+        {
+            carriedByEdid ??= patchMod.EnumerateMajorRecords()
```
Avick3110: This is a second full `patchMod.EnumerateMajorRecords()` walk over the same records `carried` (line 2923) already walks. An in-place call carrying both a FormKey-parented spec (which triggers `AlreadyCarried`) and a flat spec (which triggers this) enumerates the whole target plugin twice — and in place the target is the user's own file, the bigger of the two, which is the reason the comment above gives for indexing at all.

Both indexes are keyed off the same enumeration; one pass could fill both (FormKey → record, and editorid → records).

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Real, and measured before fixing it: with a target defining `HcClashTopic` and two INFOs under it, an in-place create of that topic with `replace=true` wrote the file and the written topic came back with `Responses.Count == 0`. Both lines gone, only the after-the-fact bracket said anything. Fixed in 5bcbf067.

I took your second option — don't offer `replace=` for a record that owns children — rather than running the forward lane's capture/restore through the upsert's replace arm, because the restore would break what the replace arm is for. Under `into=` the re-run re-creates the children in the same call (a DIAL spec plus its INFO specs); re-attaching the old children on top of that would leave two copies of every line, so the drop is exactly what keeps an `into=` re-run idempotent. The loss is only a loss in place, where the children are the user's own and no spec puts them back.

So the gate is in the in-place pre-flight, next to the other three, and asks about the record in front of it rather than its type: `ReplaceKeepsEverything` = `UpsertWouldReplace` and `WriteEngine.ChildCountOf(clash[0]) == 0`. A childless DialogTopic still replaces; a topic with lines under it refuses whether or not `replace=true` was passed, and the refusal names the count and up to five of the children:

> `HcClashUser.esp already defines DialogTopic 000807 with that editorid, and 2 record(s) live under it (HcClashLine0, HcClashLine1) — re-creating it drops the record and its children together, and no overwrite puts them back. Edit it with housecarl_apply, or pick another editorid.`

`ChildCountOf`/`ChildNamesOf` are the same walk `ChildGroupCarry` uses, so Cell / DialogTopic / Worldspace come from the pinned property set, not a hand list.

Test: `AClashOverARecordThatOwnsChildrenIsRefusedEvenWithReplace` — refused, names `HcClashLine0`, target byte-identical. It fails before the fix (the call succeeds), and with the assertion temporarily flipped it is what showed the 2 → 0.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Real, both holes, reproduced as tests before fixing them — each one wrote the cell and returned `created into HcClashUser.esp IN PLACE`, no refusal. Fixed in cfadee6e.

The check moved out of the `else if` chain and below it, so the cell arm falls through to it as well: the cell arm sets its routing and does not `continue`, the `CanCreateType` arm still `continue`s on failure, so an un-createable type keeps its own message and the collision question is asked once, in one place, for every parentless spec.

One wording change came with it. For a Cell, `UpsertWouldReplace` is false because no flat group resolves for `Cell` at all, not because the types differ — so the cross-type sentence would have read "a Cell cannot be re-created as a Cell". `ClashReason` now splits on whether the two type names match: same type and still not replaceable is the cell route, and gets "and a create never overwrites a Cell. Edit it with housecarl_apply, or pick another editorid." `replace=true` does not get through it either, which is right — the cell route files by coordinates and never overwrites anything.

Tests: `AnInPlaceCellCreateOverATakenEditoridIsRefused` (target holds the name on a Keyword; refusal names Keyword, target byte-identical) and `AnInPlaceCellCreateOverACellNamedInAnotherCaseIsRefused` (create `HcClashCell`, then `hcclashcell`; the second is refused and the file is unchanged from after the first). Both fail before the fix.

The `into=` half of the case hole is untouched, as you said — this is the in-place lane only.

src/housecarl-core/WriteEngine.cs
Avick3110: Real. `UpsertWouldReplace` returning true down the non-arm branch means `TryResolveFlatGroup` already returned true, and the call here is the same call with the same arguments — so the throw could not fire, and a group that failed to resolve would have been reported as an editorid collision by the guard above. Fixed in 56d705fc.

I kept the line and made it live rather than dropping it: the group is now resolved before the type verdict is asked for (`isArm` plus the flat resolution, once), the resolution fault throws its own sentence there, and the two arms below reuse what was resolved instead of resolving again. `UpsertWouldReplace`'s doc now names the fourth false reason and says both callers settle it first — the upsert by resolving before it asks, the pre-flight by running `CanCreateType` before it asks.

No test with this one, and I don't think there can be: the case is unreachable while `CanCreateType` and the two enumerations agree, which is the same reason you called it latent. Nothing else changes — full suite 1801 passed, `ci-all` all pass.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Real, and folded as you suggested in c429b537: one `IndexDestination()` walk fills both dictionaries — FormKey → record and editorid → records — and `AlreadyCarried` / `CarriedUnder` both call it. Still lazy on the pair (a call that asks neither builds neither), and the FormKey side keeps first-wins with `TryAdd`, which is what `GroupBy(...).First()` gave.

The one thing it costs: a patch-lane call with parented specs now builds the editorid dictionary it will never read. That is one pass over records already in hand against a whole second enumeration of the destination, which is the walk your comment is about, so I took it.

No test — it is behaviour-preserving, and the existing create tests (parented specs, prior-`into=` parents, the collision pre-flight) cover both readers. Full suite 1801 passed, `ci-all` all pass.

PR #620, opened 2026-09-07
Delete biped-slot-reference, papyrus-optimization and tool-output-awareness

### LINKED ISSUE #401: [Docs] tool-output-awareness needs a rewrite or scrapping
Migrated from the retired `dev/BACKLOG.md` (advisor sweep, Aaron-go 2026-08-22). Skill verified present: `.claude/skills/tool-output-awareness/SKILL.md` (188 lines + `evals/`).

Aaron, 2026-08-11: *"tool output awareness is a weak skill that needs re-written or scrapped for various reasons."* The reasons were not enumerated — **ask him before touching it**.

One concrete agreement gap already on record: since the #300 fix, `forward_record` declines to force a master or inline a winner, and this skill is what should tell the agent when the FEEDER shape is being asked for ("make this sit before the Reqtificator / Synthesis output" — master neither, forward neither, don't bake in what they'd regenerate). Tool and skill need to agree; today only one of them knows about the other.

**Routing:** row 9 — W6 rewrites every bundled skill against the shipped surface; this one's rewrite-or-scrap decision is part of that pass. Milestoned (row 9 is spine).

Avick3110:
Deletes three shipped skills whole, per the 2026-09-07 rulings: `biped-slot-reference`, `papyrus-optimization` and `tool-output-awareness`, with their evals and reference files — nothing is kept as a baseline for a replacement. They come out of the packager's ship list (`scripts/build-plugin.ps1`, now 11 skills), out of the Codex umbrella router (`plugin/codex/housecarl/SKILL.md`), out of both READMEs and both `THIRD-PARTY-NOTICES.txt` copies, and out of the four live pointers the remaining skills had at them (`bulk-record-jobs`'s routing table, `dialogue-authoring`'s result-script note, `npc-appearance-copy`'s generated-output note, `papyrus-reference`'s `silent-biters.md`). The `npc-appearance-copy` note keeps the fact it was making — copy the record from the mod that authored it — without the pointer. Skill counts stated in comments and prose move 14/13 → 11, and `npc-appearance-copy` is added to the root README's list, which had never named it (the count there said 13 while 14 shipped).

The one fold that rides this PR is the biped slot fact, and it needs no change: `ReadEngine.FlagSlotDisplay` (`src/housecarl-core/ReadEngine.cs:1049-1056`) already decodes a `BodyTemplate.FirstPersonFlags` value into its slot numbers with `slots.Add(30 + i)` over the set bits, gated to `BipedObjectFlag`, and `FlagBitsDisplayProbe` holds it in CI. So "slot N is bit N−30" is already in the rendering and nothing was added. Nothing from `papyrus-optimization` survives. The never-copy sentence and the generated-filename parenthetical from `tool-output-awareness` are not in this PR — they go into `ServerInstructions` in the separate standing-context PR, and `src/housecarl-mcp/Program.cs` is untouched here.

Dated eval records in the surviving skills (`eval_set.json` adjudicator notes, `bulk-record-jobs/evals/eval_results_2026-07-15.md`) still name the deleted skills as menu entries and expected routes. Those are records of runs that happened against that menu, so they are left as written rather than rewritten to hide it; the eval sets themselves are rewritten to one convention in a later wave.

Review fold: setup used to upgrade by merge-copy only, so a deleted skill survived an upgrade and kept loading. It now removes the skill folders a previous install left behind and names them — by directory diff at `~/.claude/skills/housecarl/skills/`, which is houseCARL's own root, and by a recorded install list at `~/.agents/skills/`, which is shared with other agents' skills. The Codex record covers every folder the Codex install writes under that shared root, the umbrella (`~/.agents/skills/housecarl/`) included, so a later version that stops shipping one takes it back; it starts in this version, so the changelog tells a pre-2.0.0 Codex upgrader to delete the three folders there by hand once. The prune is cleanup, and cleanup cannot fail an install: a folder that will not delete (a read-only file inside, a held handle) is caught per folder, the run goes on to the MCP registration, the folder's path is named at the end for the user to delete by hand, and the Codex record keeps it so the next upgrade retries. A package with no `skills/` folder is a broken download, not a version that ships no skills, so it prunes nothing, says so, leaves the record alone, and is refused by `Main`'s package validation alongside a missing manifest or exe. New probe `setup-skill-prune` drives the real `TryInstall` over a synthetic package, through seven cases including the read-only leftover, the skills-less package and the dropped umbrella. The changelog entry now says where each of the three skills' facts went, and the DrHeisen credit in both READMEs names `oar-authoring` plus two retired skills instead of a skill the reader cannot find.

Build, tests and `ci-all` pass (1767 tests, 128/128 probes), and a full `scripts/build-plugin.ps1` run packages the tree with `skills: 11` and a clean leak-check.

Closes #401


plugin/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **Three skills are removed: `biped-slot-reference`, `papyrus-optimization` and `tool-output-awareness`.**
```
Avick3110, line 36: **Upgraders keep all three deleted skills — "eleven skills ship now" is false for anyone who installs over an existing install.**

`houseCARL-Setup.exe` is the documented upgrade path (`packaging/START-HERE.txt`: "UPDATING (installing a newer houseCARL over an older one) — Same as installing"), and it upgrades by merge-copy only. `Program.cs`'s `CopyDirectory` (src/housecarl-setup/Program.cs:462) does `Directory.CreateDirectory` + `File.Copy(overwrite: true)` and never deletes; there is no `Directory.Delete`/`File.Delete` anywhere in the setup utility. So:

- Claude path — `InstallForClaude` copies the plugin tree over `~/.claude/skills/housecarl/`. `skills/papyrus-optimization/`, `skills/tool-output-awareness/` and `skills/biped-slot-reference/` survive from the previous version, keep their `SKILL.md` frontmatter, and keep loading. Their descriptions stay in always-loaded context and they keep triggering — including the 366-line `papyrus-performance-reference.md` this PR deleted.
- Codex path — `InstallForCodex` copies each shipped skill folder flat into `~/.agents/skills/<slug>`. The three stale folders remain there too, and now they are orphans: the umbrella router (`plugin/codex/housecarl/SKILL.md`) no longer names them, so `CodexUmbrellaCoverageProbe`'s invariant holds in the repo while the installed tree violates it.

This is the first PR that removes a skill, so the merge-copy has never mattered before. Either the setup utility needs to prune skill folders not in the shipped set before copying, or this changelog entry has to tell upgraders to delete the three folders by hand from `~/.claude/skills/housecarl/skills/` and `~/.agents/skills/`.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
+  Eleven skills ship now; the biped slot fact they were kept for is already in the server's own output, which
```
Avick3110: The rationale covers one of the three removals and leaves the other two unexplained. "the biped slot fact **they** were kept for" reads as the reason all three went, but only `biped-slot-reference` was about biped slots; `papyrus-optimization` (script cost review) and `tool-output-awareness` (keep regenerated output out of authored patches) had nothing to do with it. A user who was using either of those reads this entry and learns neither that their capability is gone nor where it went.

The `tool-output-awareness` half is the one with teeth: the PR body says its never-copy rule lands in `ServerInstructions` in a *separate* PR. If this PR ships first, the release removes the guardrail with nothing in its place except the single fact folded into `npc-appearance-copy` (which only fires on an NPC copy). Either say what replaces each removal, or hold this behind the standing-context PR.

README.md
```diff
@@ -225,7 +216,5 @@ pointers.
   for the bundled `papyrus-reference` skill. Thank you.
 - **Zzyxzz** (SkyPatcher) and **powerofthree** (SPID and KID) — whose public documentation the
   distributor-authoring grammar facts were drawn from.
-- **DrHeisen** — contributed the `papyrus-optimization` skill (houseCARL's first community-contributed
-  skill, a Papyrus performance reviewer), the `oar-authoring` skill (Open Animation Replacer config
-  authoring), and the `tool-output-awareness` skill (keeping generated-tool output out of authored patches).
-  Thank you.
+- **DrHeisen** — contributed the `oar-authoring` skill (Open Animation Replacer config authoring), and
```
Avick3110: The credit now names one thing and gestures at a second that nothing identifies. "contributed the `oar-authoring` skill (…), and houseCARL's first community-contributed skill" parses two ways and both are wrong: as one item it calls `oar-authoring` the first community skill, which it wasn't (the removed text says `papyrus-optimization` was); as two items it credits an unnamed skill the reader cannot find anywhere in the repo, because it's the one this PR deleted.

Either drop the trailing clause, or make it explicit that the first community-contributed skill was `papyrus-optimization` and it has since been retired. `plugin/README.md:137` carries the identical sentence and needs the same fix.

plugin/README.md
```diff
@@ -134,7 +134,5 @@ no linking exception), which houseCARL bundles. Every third-party component and
   the bundled `papyrus-reference` skill. Thank you.
 - **Zzyxzz** (SkyPatcher) and **powerofthree** (SPID and KID) — the public documentation behind the
   distributor-authoring skills.
-- **DrHeisen** — contributed the `papyrus-optimization` skill (houseCARL's first community-contributed
-  skill, a Papyrus performance reviewer), the `oar-authoring` skill (Open Animation Replacer config
-  authoring), and the `tool-output-awareness` skill (keeping generated-tool output out of authored patches).
-  Thank you.
+- **DrHeisen** — contributed the `oar-authoring` skill (Open Animation Replacer config authoring), and
```
Avick3110: Same dangling credit as `README.md:219` — "and houseCARL's first community-contributed skill" names nothing, and the skill it refers to (`papyrus-optimization`) is deleted by this PR. This is the copy that ships to users, so fix both together.

plugin/CHANGELOG.md
Avick3110: Real, and fixed in the installer rather than in prose: 88b91944.

`TryInstall` now removes, after the copy, the skill folders a previous install left behind, and names each one it removed.

- Claude — `~/.claude/skills/housecarl/skills/` is houseCARL's own root, so it is a directory diff: any folder there that the package does not ship goes. This covers today's three for every Claude upgrader.
- Codex — `~/.agents/skills/` holds other agents' skills too, so a directory diff there would delete things that are not ours. The install now records the folder names it put there (`installed-skills.txt`, in houseCARL's own data dir beside the Codex server dir) and the next upgrade takes back only recorded names that the new package no longer ships. Names that are not a bare folder name are dropped when the record is read, so a hand-edited record cannot aim the delete elsewhere.

The Codex record starts in this version, so a Codex install made before it has nothing recorded and keeps the three folders. The changelog now says that and tells those upgraders to delete `biped-slot-reference/`, `papyrus-optimization/` and `tool-output-awareness/` from `~/.agents/skills/` by hand once (bda819b9).

New CI probe `setup-skill-prune` (`src/housecarl-generator/SetupSkillPruneProbe.cs`), at the level `setup-update-lock-guard` uses: it drives the real `TryInstall` over a synthetic package + temp home, installs three skills for Both, re-installs a package shipping two, and asserts the dropped folder is gone from both locations, the kept ones and the umbrella survive, a foreign skill folder in `~/.agents/skills` is untouched, and the run said what it removed. ci-all: 128/128.

plugin/CHANGELOG.md
Avick3110: First half real — folded in bda819b9. The entry now has a sentence per removed skill saying where its surviving fact went: `biped-slot-reference`'s slot-to-bit fact is in the server's own output, which decodes `BodyTemplate.FirstPersonFlags` into slot numbers; `papyrus-optimization` leaves nothing behind, because its review was no better than the one you get without it; `tool-output-awareness`'s never-copy rule and its list of the tools that generate output move into the server's own instructions, with the NPC-copy case stated inline in `npc-appearance-copy`.

Second half — the release shipping with the guardrail gone — no. No release happens between this PR and the standing-context one: 2.0.0 releases at the end of item 30, and both PRs are in the same wave before it. Nothing installs from `Unreleased`, so there is no window in which a user has this deletion without the `ServerInstructions` text. Holding this PR behind the other one would only serialise two changes that land together anyway.

README.md
Avick3110: Real — fixed in 34928d09, in both copies. The line now reads:

> **DrHeisen** — contributed the `oar-authoring` skill (Open Animation Replacer config authoring), along with two earlier community-contributed skills that have since been retired. Thank you.

That is what the record says he contributed (`oar-authoring`, plus `papyrus-optimization` and `tool-output-awareness`, both deleted here), with no dangling referent to a skill the reader cannot find. The `oar-authoring` credit in `THIRD-PARTY-NOTICES.txt` names only that skill and is already true as written.

plugin/README.md
Avick3110: Real — fixed in 34928d09, in both copies. The line now reads:

> **DrHeisen** — contributed the `oar-authoring` skill (Open Animation Replacer config authoring), along with two earlier community-contributed skills that have since been retired. Thank you.

That is what the record says he contributed (`oar-authoring`, plus `papyrus-optimization` and `tool-output-awareness`, both deleted here), with no dangling referent to a skill the reader cannot find. The `oar-authoring` credit in `THIRD-PARTY-NOTICES.txt` names only that skill and is already true as written.

src/housecarl-setup/Program.cs
```diff
@@ -374,6 +374,19 @@ private static void InstallForClaude(string pluginSrc, string home)
...
+        // is a leftover.
+        string installedSkills = Path.Combine(skillsDest, "skills");
+        List<string> shipped = ShippedSkillNames(pluginSrc);
+        List<string> stale = Directory.Exists(installedSkills)
+            ? Directory.GetDirectories(installedSkills)
+                .Select(d => Path.GetFileName(d)!)
+                .Where(n => !shipped.Contains(n, StringComparer.OrdinalIgnoreCase))
+                .ToList()
+            : new List<string>();
+        ReportRemoved("Claude Code", installedSkills, RemoveSkillDirs(installedSkills, stale));
```
Avick3110: **A prune failure aborts the install after the copy, before the MCP registration.**

`RemoveSkillDirs` calls `Directory.Delete(dir, recursive: true)` unguarded, and this line sits between `CopyDirectory` and `RegisterClaudeMcpServer` (line 392).

Concrete: a leftover skill folder holds a read-only file (restored from a backup, extracted read-only, set by AV). `Directory.Delete` throws `UnauthorizedAccessException` — not the `IOException` + `IsSharingViolation` that `TryInstall` catches — so it escapes to `Main`'s generic handler. The plugin tree has already been copied, but `RegisterClaudeMcpServer` never runs; the user gets "houseCARL setup did not complete" and an unregistered server. Before this PR the leftover folder simply survived and the install succeeded.

The milder variant: a held handle on a `SKILL.md` *is* caught, but surfaced as "a server file is in use — fully quit Claude Code AND Codex", which is not what happened.

Cleanup shouldn't be able to fail an otherwise good install — wrap the prune (both locations) in try/catch and report what it couldn't remove, or move it after registration.

src/housecarl-setup/Program.cs
```diff
@@ -457,6 +482,57 @@ private static int Finish(int exitCode)
         return exitCode;
     }
 
+    // ---- stale skill folders ----------------------------------------------
+
+    /// <summary>The skill folder names this package ships.</summary>
+    private static List<string> ShippedSkillNames(string pluginSrc)
```
Avick3110: **An empty shipped list means "delete every installed skill", not "delete nothing".**

`ShippedSkillNames` returns an empty list when `pluginSrc/skills` is missing, and both prunes read that as "the package ships nothing".

Concrete: `Main`'s package validation checks only `.claude-plugin/plugin.json` and `server/housecarl-mcp.exe`, so a package whose `skills/` folder didn't make it (partial unzip, AV quarantine, a hand-trimmed copy) still installs. `shipped` is then empty, the Claude directory diff (382-388) classifies all 11 installed skill folders as stale and deletes them, and the Codex prune (432-435) removes every recorded folder. The installer prints "removed skills this version no longer ships", exits 0, and the user is left with a skill-less install. Line 437 then overwrites the Codex record with an empty list, so the next good install has nothing left to reconcile against and the leftovers stay silently.

A package that ships no skills is a broken package, not a package that dropped every skill — guard both prunes on `shipped.Count > 0` (and skip the record write in that case).

src/housecarl-setup/Program.cs
```diff
@@ -411,6 +424,18 @@ private static void InstallForCodex(string pluginSrc, string home, string? homeO
...
+        // shared with every other agent's skills, so this prunes ONLY the folder names houseCARL recorded
+        // installing — never a directory diff of a dir we do not own.
+        List<string> shippedSkills = ShippedSkillNames(pluginSrc);
+        string recordPath = CodexSkillRecord(home, homeOverride);
+        List<string> staleSkills = ReadSkillRecord(recordPath)
+            .Where(n => !shippedSkills.Contains(n, StringComparer.OrdinalIgnoreCase))
+            .ToList();
+        ReportRemoved("Codex", skillsRoot, RemoveSkillDirs(skillsRoot, staleSkills));
+        Directory.CreateDirectory(Path.GetDirectoryName(recordPath)!);
+        File.WriteAllLines(recordPath, shippedSkills);
```
Avick3110: **The record omits one folder the Codex install actually puts in the shared dir: the umbrella.**

`CodexSkillRecord`'s doc comment says it records "the skill folders it put in the shared `~/.agents/skills`, so a later upgrade can take back exactly those". This writes `ShippedSkillNames` only, but the umbrella is copied to `skillsRoot/housecarl` a few lines below (447-452) and never recorded.

Concrete: a later version stops shipping `codex/housecarl`, or `PluginFolderName` changes. The orphaned `~/.agents/skills/housecarl/` folder is outside the record, so no upgrade prunes it, and Codex keeps loading a `$housecarl` router that lists skills no longer installed — the exact failure this PR exists to fix, at the one folder the record does not cover.

Appending `PluginFolderName` to the recorded list when the umbrella is installed closes it.

plugin/CHANGELOG.md
Avick3110: Already folded, in `88b91944` (setup prunes) and `dd87f2d0` + `bda819b9` (the changelog entry). Setup no longer upgrades by merge-copy alone: `InstallForClaude` diffs `~/.claude/skills/housecarl/skills/` against the package's skill list and deletes what is not in it; `InstallForCodex` takes back only the folder names it recorded installing, and starts recording them in this version, so the changelog tells a pre-2.0.0 Codex upgrader to delete the three by hand. `plugin/CHANGELOG.md` now says:

> `~/.agents/skills/` (the Codex location) is shared with other agents' skills, so setup takes back only the folder names it recorded installing — and it starts recording them in this version, which means a Codex install made before it still has `biped-slot-reference/`, `papyrus-optimization/` and `tool-output-awareness/` there; delete those three folders by hand once.

`setup-skill-prune` drives the real `TryInstall` over a synthetic package and holds it. This round's `037d8b28` hardens the same prune against your three later comments.

plugin/CHANGELOG.md
Avick3110: Already folded, in `bda819b9`. The entry now says where each of the three went, and what replaces the `tool-output-awareness` rule:

> `biped-slot-reference` said slot N is `FirstPersonFlags` bit N−30; the server's own output already decodes a `BodyTemplate.FirstPersonFlags` value into its slot numbers, so that fact stays. `papyrus-optimization` leaves nothing behind — its script-cost review was no better than the review you get without it. `tool-output-awareness`'s rule — never copy a record or asset path a generated tool re-derives — and its list of the plugins that produce them move into the server's own instructions, which every session reads (a separate change in this release); the same fact for a copied NPC is now stated in `npc-appearance-copy`.

So the guardrail's replacement is named and is in the same release, rather than this shipping with nothing in its place.

README.md
Avick3110: Already folded, in `34928d09`, in both READMEs. `README.md:219-220` and `plugin/README.md:137-138` now read:

> **DrHeisen** — contributed the `oar-authoring` skill (Open Animation Replacer config authoring), along with two earlier community-contributed skills that have since been retired. Thank you.

That is a third wording rather than either of the two you offered, and here is the reason: DrHeisen contributed three skills — `papyrus-optimization` (the first community one) and `tool-output-awareness`, both deleted here, plus `oar-authoring` — so dropping the clause undercredits, and naming the retired two points the reader at folders the package no longer has. The wording keeps the credit, drops the "first" claim that was false of `oar-authoring`, and says the others are retired so nobody goes looking. If you would rather have `papyrus-optimization` named outright, say so and I will name both.

plugin/README.md
Avick3110: Already folded with the root README, in `34928d09` — the shipped copy at `plugin/README.md:137-138` carries the same sentence:

> **DrHeisen** — contributed the `oar-authoring` skill (Open Animation Replacer config authoring), along with two earlier community-contributed skills that have since been retired. Thank you.

Reasoning for that wording is on the `README.md` comment.

src/housecarl-setup/Program.cs
Avick3110: Real, folded in `037d8b28`. `RemoveSkillDirs` now catches `IOException` and `UnauthorizedAccessException` **per folder**: the loop goes on to the next name, the folder's full path is collected, and the install runs to the MCP registration exactly as before. `TryInstall` says at the end what is still there:

    NOTE: houseCARL installed, but these old skill folders could not be deleted (a file in
          them is read-only or open) — delete them by hand so they stop loading:
          - <path>

Cleanup cannot fail an install that otherwise worked, so the pre-PR behaviour (the leftover survives, the install succeeds) is what a delete failure now degrades to, plus the sentence naming it. On the Codex side a folder that would not delete stays on the record, so the next upgrade tries it again instead of forgetting a folder houseCARL put there.

`setup-skill-prune` gains T5: a stale folder holding a read-only file at both locations, both host config files deleted first so their reappearance proves the registration ran after the failed prune. It checks the outcome is `Installed`, both registrations happened, the folder is still there, the run names its full path, and the record kept it.

src/housecarl-setup/Program.cs
Avick3110: Real, folded in `037d8b28`. `ShippedSkillNames` now returns `null` when `pluginSrc/skills` is absent — "this package ships no skills folder", which is not the same as an empty list — and both prunes refuse on it, say why, and skip the Codex record write so the next good install still has a list to reconcile against:

    [Claude Code] this package has no skills folder, so no installed skill was removed
          (a package ships skills; unzip the download again if this is not what you expect)

On the package validation: I added it, and it is consistent with how `Main` validates today — the same `if`, next to `.claude-plugin/plugin.json` and `server/housecarl-mcp.exe`, reusing the same "couldn't find the houseCARL plugin next to this program" error, since a package without `skills/` is the same class of broken unzip that message already describes. Nothing else ships a package without skills: `build-plugin.ps1` always writes the folder, and the probes drive `TryInstall` rather than `Main`, so the refusal inside the prune is the load-bearing half and the validation is the earlier, clearer stop.

`setup-skill-prune` gains T6: the package's `skills/` is moved away and the install re-run — the installed skills at both locations survive, the umbrella survives, the record is not overwritten, and the run says why it pruned nothing.

src/housecarl-setup/Program.cs
Avick3110: Real, folded in `037d8b28`. `InstallForCodex` now installs the umbrella *before* the prune and records every folder it wrote under the shared root: `installedNames = shipped skills + PluginFolderName when the package ships codex/housecarl`. That list is both what the prune keeps and what the record holds, so a later package that stops shipping the umbrella (or renames it) takes the orphan back instead of leaving a `$housecarl` router listing skills that are gone. Folders the prune could not delete are appended to the record too, so they are retried rather than forgotten.

`setup-skill-prune` gains T7: the record is checked to contain `housecarl` after the install, then `codex/housecarl` is deleted from the package and the install re-run — the installed umbrella is gone from `~/.agents/skills`, is named in the output, is off the record, and both the shipped skills and the foreign skill folder are untouched.

PR #621, opened 2026-09-07
Check shipped skill pointers and give the Codex tree a skills/ root

Avick3110:
Wave 0 of the skill rewrite: the packager side of three rulings in `dev/DECISIONS.md` dated 2026-09-07. The leak scan in `scripts/build-plugin.ps1` now also checks what the shipped skill markdown points at, in **both** forms a pointer is written: the `references/<file>` hop, and the bare sibling name (`value-tables.md`) that most of the corpus uses. Every pointer a shipped `.md` writes must exist in that same shipped copy, and none may point at a file the packager strips (`evals/`, whose eval file is `evals/evals.json`, and `references/_CORPUS_STATUS.md`). It reads every `.md` under a skill, not only `SKILL.md` — a reference file pointing at a stripped file is the same dead link in the installed plugin — and it matches either slash in any case, with a backslash pointer reported as its own defect since skill pointers are written with forward slashes. Both regexes guard their left edge, so `retrievals/cache.json` and `cross-references/glossary.md` are not read as `evals/` and `references/` pointers. It is the same scan, not a new script — one failing line names the skill, the file and the pointer, and the build stops.

This is the **resolution** half of ruling 2. The **form** half — that a pointer is written as `references/<file>`, one hop — is enforced per skill by the regrade in that skill's rewrite wave, not by this scan: flagging form today would go red on every skill not yet rewritten. The scan comment and this body both say so rather than implying pointer form is closed.

The bare form resolves against every name the skill ships, and only `.md` and `.jsonl` names are candidates — those are the only extensions any shipped skill file has (measured: 254 `.md`, 11 `.jsonl`, nothing else). A bare `.json`, `.ini` or `.psc` name in these skills is a file in the modded game, not a pointer at a sibling document: OAR's `config.json`/`user.json` (58 mentions), SPID's `_DISTR.ini` (13), `SkyPatcher.ini` (6), vanilla `Form.psc`/`UI.psc`/`Input.psc` (13), `vcpkg.json` (13). Taking those extensions as pointers would hard-stop the build on 143 correct documents, so they are not candidates and the comment says why.

The Codex umbrella skill now ships as `dist/codex/skills/housecarl/SKILL.md` instead of `dist/codex/housecarl/SKILL.md`, so its skill directory is an immediate child of a `skills/` root, and the same scan covers that tree too; the installer and the setup-lock probe follow the new package path. The whole `plugin/codex` tree ships: every directory (hidden ones included) as a skill under `skills/`, and any loose file at the Codex package root, where a package-level Codex file belongs — nothing in that tree is dropped without saying so. The excluded-file scan covers the Codex package root, not just its `skills/`. The scan and the Codex copy both moved ahead of the `-PluginTreeOnly` return, which is what CI assembles with, so CI runs the check rather than only a full release build.

Both arms are proven on the real skill set. Green: 15 skills, 255 markdown files, across the two trees, leak-check clean; full build packs the zip with `houseCARL/codex/skills/housecarl/SKILL.md`. Red, six in one run on `oar-authoring/references/oar-config-reference.md`: a bare `_CORPUS_STATUS.md`, `evals/evals.json`, `references/does-not-exist.md`, `references\backslash-form.md`, a bare `gone-away.md` and a bare `gone-away.jsonl` each produce their own line and stop the build, while a bare `oar-config-reference.md` (which resolves) and the decoys `retrievals/cache.json`, `cross-references/glossary.md`, `config.json`, `Form.psc` and `_DISTR.ini` stay silent. A zero-byte `SKILL.md` is a seventh arm — it reports `oar-authoring ships an empty SKILL.md.` instead of throwing an `ArgumentNullException` out of the regex.

What the scan flagged on the current skills:

Five pointers at `_CORPUS_STATUS.md`, which the packager deletes from every shipped copy — the defect `dev/skill-validation/eval-2026-09/rewrite-inputs.md` section (b) item 1 records (the two bodies write it with the `references/` hop, the three reference files by bare filename, which is why the scan matches that name bare too):

- `dialogue-authoring/SKILL.md:68`
- `kid-authoring/SKILL.md:124`
- `dialogue-authoring/references/condition-functions.md:13`
- `dialogue-authoring/references/dialogue-branch.md:8`
- `dialogue-authoring/references/quest-objectives-tab.md:8`

Every one of those sentences already said the file is dev-side and unshipped, so this PR drops the path from each and keeps the sentence.

Four more from the bare form, each a shipped document naming another project's file by a bare name that reads as a sibling:

- `skse-plugin-authoring/references/multi-runtime.md` (18 mentions) and `references/plugin-dissection.md` (2) cite CommonLibSSE-NG's bundled `CLAUDE.md`. They now write `ng/CLAUDE.md`, the qualified form `references/hooking.md` in the same skill already used four times.
- `skse-plugin-authoring/references/hooking.md:466` cites `VERIFICATION_MIGRATION.md`; it now writes `docs/VERIFICATION_MIGRATION.md`, the form line 445 of the same file already used.
- `tool-output-awareness/SKILL.md:136` said "the modlist's own `CLAUDE.md` / project memory"; the bare name is dropped and the sentence keeps "the modlist's own project memory".

Nothing else flagged: no other pointer in any shipped markdown is broken, in either form. None of the three skills a parallel PR deletes (biped-slot-reference, papyrus-optimization, tool-output-awareness) has a broken pointer today. The rest of those bodies belongs to their own rewrite PRs; the ship list is untouched — that PR owns it.

The Codex package path is tested on both sides, not just written twice. The setup-lock probe's T1 asserts the umbrella lands at `~/.agents/skills/housecarl`, and the package fixture it installs from is the path **read out of `scripts/build-plugin.ps1`** (`$CodexRoot` / `$CodexSkills`, each a `Join-Path` of the one above it) rather than restated in the probe — so the probe pins the pair, not just the installer. Reverting the installer's path alone turns T1 red; moving the packager's `$CodexSkills` alone turns it red too, naming both sides: `FAIL Codex umbrella skill packed at codex/agents/housecarl landed at ~/.agents/skills/housecarl` (both proven). A script the probe cannot read, or an assignment that has gone away, is its own failing check with the reason — it never certifies the path pair from its own restatement.

The whole-tree excluded-file scan is back at step 10 as the backstop, and all three of its checks now walk the package root: an `appsettings.json` the step-9 setup-utility publish drops beside `houseCARL-Setup.exe` lands outside `dist/housecarl`, and is now caught rather than packed. Step 7's scan is the skill-scoped one CI runs.

No skill's `evals/eval_set.json` is renamed here; that lands per skill in its own PR. The packager never named an eval file by name, and still does not — it strips the whole `evals/` directory, and the comment now names `evals/evals.json` as the file that directory holds.

Build and tests from the worktree: `dotnet build` clean, full `build-plugin.ps1` clean (341 zip entries), 1789 tests passed, `ci-all` 127/127.

### DIFF
--- a/scripts/build-plugin.ps1
+++ b/scripts/build-plugin.ps1
@@ -155,11 +155,14 @@ foreach ($f in @('.mcp.json','LICENSE','THIRD-PARTY-NOTICES.txt','README.md','CH
 # ---- 7. leak-check the shipped skill trees (excluded files + markdown pointers) ----
 # Runs on both trees before the -PluginTreeOnly return, so CI (which assembles with that switch)
 # runs it too. Two invariants, from dev/DECISIONS.md 2026-09-07 ruling 2: nothing this script
-# excludes may survive into a shipped copy, and every references/<file> a shipped markdown file
-# writes must exist in that same copy and must not be a file this script excludes. The pointer scan
-# reads every .md in the skill, not just SKILL.md - a reference file pointing at a stripped file is
-# the same dead link. It matches either slash and any case, and a backslash pointer is itself a
-# defect (skill pointers are written with forward slashes).
+# excludes may survive into a shipped copy, and every file a shipped markdown file points at must
+# exist in that same copy and must not be a file this script excludes. Both pointer forms are read:
+# the references/<file> hop, and the bare sibling name that most of the corpus uses. The scan reads
+# every .md in the skill, not just SKILL.md - a reference file pointing at a stripped file is the
+# same dead link. It matches either slash and any case, and a backslash pointer is itself a defect
+# (skill pointers are written with forward slashes). This checks that a pointer RESOLVES; the ruling's
+# form half - that a pointer is written as references/<file> - is enforced per skill by the regrade in
+# that skill's rewrite wave, not here.
 Step '7/12' 'Leak-check skills (excluded files + markdown pointers)'
 $skillLeaks = @()
 foreach ($r in $LeakRoots) {
@@ -174,6 +177,9 @@ $pointerFails = @()
 $docCount = 0
 foreach ($r in $SkillRoots) {
   foreach ($skillDir in (Get-ChildItem $r -Directory)) {
+    # every file name this skill ships - what a bare sibling pointer has to resolve to
+    $shipped = @{}
+    Get-ChildItem $skillDir.FullName -Recurse -Force -File | ForEach-Object { $shipped[$_.Name] = $true }
     foreach ($doc in (Get-ChildItem $skillDir.FullName -Recurse -Force -File -Filter '*.md')) {
       $docCount++
       $rel     = ($doc.FullName.Substring($skillDir.FullName.Length).TrimStart('\')) -replace '\\','/'
@@ -194,6 +200,16 @@ foreach ($r in $SkillRoots) {
           $pointerFails += ("{0} ({1}) points at {2}, which this script strips from every shipped copy." -f $skillDir.Name, $rel, $ptr)
         } elseif (-not (Test-Path (Join-Path $skillDir.FullName $ptr))) {
           $pointerFails += ("{0} ({1}) points at {2}, which is not in the shipped copy." -f $skillDir.Name, $rel, $ptr)
+        }
+      }
+      # the bare sibling form, the dominant one in the corpus (a plain value-tables.md, no references/ hop):
+      # it resolves against every name the skill ships. Only .md and .jsonl are candidates - those are the
+      # extensions shipped skill files use, so a bare .json/.ini/.psc name here is a file in the modded game
+      # (OAR's config.json, SPID's _DISTR.ini, a vanilla Form.psc), never a pointer at a sibling document.
+      foreach ($m in [regex]::Matches($docText, '(?<![A-Za-z0-9_./\\-])[A-Za-z0-9_-]+\.(?:md|jsonl)(?![A-Za-z0-9])', 'IgnoreCase')) {
+        if ($m.Value -eq '_CORPUS_STATUS.md') { continue }   # the stripped-file rule above owns this name
+        if (-not $shipped.ContainsKey($m.Value)) {
+          $pointerFails += ("{0} ({1}) points at {2}, which is not in the shipped copy." -f $skillDir.Name, $rel, $m.Value)
         }
       }
     }
--- a/scripts/build-plugin.ps1
+++ b/scripts/build-plugin.ps1
@@ -10,6 +10,7 @@
 
   Outputs (all gitignored build artifacts, reproducible on demand):
       dist/housecarl/                         the plugin tree (what `claude plugin validate` checks)
+      dist/codex/skills/housecarl/            the Codex umbrella skill, under a skills/ root
       dist/houseCARL-Setup.exe                the no-CLI desktop installer
       dist/START-HERE.txt                     friend-facing note (version stamped from plugin.json)
       dist/.claude-plugin/marketplace.json    CLI-install descriptor (local marketplace)
@@ -21,9 +22,10 @@
   After it succeeds, run the validation gate (necessary, not sufficient):
       claude plugin validate ./dist/housecarl --strict
 
-  -PluginTreeOnly assembles just dist/housecarl (skills + plugin files) and stops: no corpus,
-  no server publish, no setup exe, no zip. That is the whole tree `claude plugin validate`
-  reads, and it is what CI runs the real validator against.
+  -PluginTreeOnly assembles just dist/housecarl (skills + plugin files) and the Codex skill tree,
+  runs the skill leak-check over both, and stops: no corpus, no server publish, no setup exe, no
+  zip. dist/housecarl is the whole tree `claude plugin validate` reads, and it is what CI runs the
+  real validator against; the leak-check runs here so CI runs it too.
 
   Reference:
       dev/plans/PLUGIN_BUILD_EXECUTION_2026-06-03.md   (the checklist this implements)
@@ -41,6 +43,8 @@ $PkgRoot      = Join-Path $RepoRoot 'dist'              # package root: ships ho
 $DistRoot     = Join-Path $PkgRoot 'housecarl'          # the plugin tree (what `claude plugin validate` checks)
 $ServerDir    = Join-Path $DistRoot 'server'
 $SkillsDir    = Join-Path $DistRoot 'skills'
+$CodexRoot    = Join-Path $PkgRoot 'codex'              # the Codex bundle root (beside, not inside, the plugin tree)
+$CodexSkills  = Join-Path $CodexRoot 'skills'           # Codex skill dirs are immediate children of a skills/ root
 $PluginSrc    = Join-Path $RepoRoot 'plugin'
 $GeneratedDir = Join-Path $RepoRoot 'generated'
 $CorpusSrc    = Join-Path $GeneratedDir 'corpus.json'
@@ -67,7 +71,7 @@ Write-Host ("Building houseCARL v{0}" -f $Version) -ForegroundColor Green
 # Clean the WHOLE package root, not just dist/housecarl: stale package-root extras (codex/,
 # START-HERE.txt, a previous setup exe) would otherwise survive into the zip - a stale nested
 # codex/codex/ subtree did exactly that at the 1.2.2 build.
-Step '0/11' 'Clean dist/ (package root)'
+Step '0/12' 'Clean dist/ (package root)'
 if (Test-Path $PkgRoot) { Remove-Item $PkgRoot -Recurse -Force }
 New-Item -ItemType Directory -Path $DistRoot -Force | Out-Null
 
@@ -76,7 +80,7 @@ New-Item -ItemType Directory -Path $DistRoot -Force | Out-Null
 if (-not $PluginTreeOnly) {
 
   # ---- 1. regenerate the rulebook (corpus.json + mutagen-reference shards, in sync by construction)
-  Step '1/11' 'Regenerate corpus.json (generator)'
+  Step '1/12' 'Regenerate corpus.json (generator)'
   # Explicit absolute args make this CWD-independent (Program.cs defaults are relative to CWD).
   dotnet run --project $GenProj -c Release -- $GeneratedDir $RefDir
   if ($LASTEXITCODE -ne 0) { throw "generator failed (exit $LASTEXITCODE)" }
@@ -86,7 +90,7 @@ if (-not $PluginTreeOnly) {
   # ---- 2. publish the server (framework-dependent; trimming OFF) -------------
   # -p:Version stamps the plugin.json version into the exe, which ServerInfo reports over MCP
   # (one version home; an unstamped dev build says 0.0.0-dev).
-  Step '2/11' 'Publish server (Release, win-x64, framework-dependent)'
+  Step '2/12' 'Publish server (Release, win-x64, framework-dependent)'
   dotnet publish $McpProj -c Release -r win-x64 --self-contained false -p:Version=$Version -o $ServerDir
   if ($LASTEXITCODE -ne 0) { throw "publish failed (exit $LASTEXITCODE)" }
   $Exe = Join-Path $ServerDir 'housecarl-mcp.exe'
@@ -100,48 +104,102 @@ if (-not $PluginTreeOnly) {
   # Mutagen release its corresponding-source line points at. Both are written here from the publish just
   # made, so neither can drift from what ships; the licence texts stay hand-authored. Both tracked copies
   # (repo root and plugin/) are rewritten from one string, so they stay byte-identical.
-  Step '3/11' 'Write the generated notices regions from the publish output'
+  Step '3/12' 'Write the generated notices regions from the publish output'
   & (Join-Path $PSScriptRoot 'generate-notices.ps1') -PublishDir $ServerDir -RepoRoot $RepoRoot
 
   # ---- 4. corpus beside the exe (the proven #1 requirement) ------------------
   # Reads survive without it via a reflection fallback, but writes + type-filtered queries need it.
-  Step '4/11' 'Copy corpus.json beside the exe'
+  Step '4/12' 'Copy corpus.json beside the exe'
   Copy-Item $CorpusSrc (Join-Path $ServerDir 'corpus.json') -Force
 }
 
-# ---- 5. bundle the 11 skills (exclude evals/ + _CORPUS_STATUS.md; KEEP all .jsonl) ----
-Step '5/11' 'Bundle skills'
+# ---- 5. bundle the skills, both trees (exclude evals/ + _CORPUS_STATUS.md; KEEP all .jsonl) ----
+# The Codex umbrella skill (the $housecarl entry point for Codex) ships at the PACKAGE ROOT, beside -
+# not inside - dist/housecarl/, so the Claude install (which copies the housecarl/ plugin tree
+# wholesale) never picks it up; only the setup utility's Codex path places it. Its skill dir is an
+# immediate child of a skills/ root (dist/codex/skills/housecarl), the shape Codex accepts.
+Step '5/12' 'Bundle skills (plugin tree + Codex umbrella)'
 New-Item -ItemType Directory -Path $SkillsDir -Force | Out-Null
 foreach ($s in $Skills) {
   $src = Join-Path $RepoRoot ".claude\skills\$s"
   if (-not (Test-Path $src)) { throw "skill not found: $src" }
   Copy-Item $src (Join-Path $SkillsDir $s) -Recurse -Force
 }
-# prune dev/QA meta from the copies (index.jsonl + the mutagen shards stay - load-bearing)
-Get-ChildItem $SkillsDir -Directory -Recurse -Filter 'evals' | Remove-Item -Recurse -Force
-Get-ChildItem $SkillsDir -File -Recurse -Filter '_CORPUS_STATUS.md' | Remove-Item -Force
+$CodexSrc = Join-Path $PluginSrc 'codex'
+$SkillRoots = @($SkillsDir)
+if (Test-Path $CodexSrc) {
+  New-Item -ItemType Directory -Path $CodexSkills -Force | Out-Null
+  Get-ChildItem $CodexSrc -Directory | ForEach-Object { Copy-Item $_.FullName $CodexSkills -Recurse -Force }
+  $SkillRoots += $CodexSkills
+}
+# prune dev/QA meta from the copies (index.jsonl + the mutagen shards stay - load-bearing). The eval
+# file is evals/evals.json (dev/DECISIONS.md, 2026-09-07, ruling 3); the whole directory is stripped.
+foreach ($r in $SkillRoots) {
+  Get-ChildItem $r -Directory -Recurse -Filter 'evals' | Remove-Item -Recurse -Force
+  Get-ChildItem $r -File -Recurse -Filter '_CORPUS_STATUS.md' | Remove-Item -Force
+}
 
 # ---- 6. plugin source files ------------------------------------------------
-Step '6/11' 'Copy plugin files'
+Step '6/12' 'Copy plugin files'
 Copy-Item (Join-Path $PluginSrc '.claude-plugin') $DistRoot -Recurse -Force   # -> dist/housecarl/.claude-plugin/plugin.json
 foreach ($f in @('.mcp.json','LICENSE','THIRD-PARTY-NOTICES.txt','README.md','CHANGELOG.md')) {
   Copy-Item (Join-Path $PluginSrc $f) (Join-Path $DistRoot $f) -Force
 }
 
+# ---- 7. leak-check the shipped skill trees (excluded files + SKILL.md pointers) ----
+# Runs on both trees before the -PluginTreeOnly return, so CI (which assembles with that switch)
+# runs it too. Two invariants, from dev/DECISIONS.md 2026-09-07 ruling 2: nothing this script
+# excludes may survive into a shipped copy, and every references/<file> a shipped SKILL.md body
+# writes must exist in that same copy and must not be a file this script excludes.
+Step '7/12' 'Leak-check skills (excluded files + SKILL.md pointers)'
+$skillLeaks = @()
+foreach ($r in $SkillRoots) {
+  $skillLeaks += Get-ChildItem $r -Recurse -Force -File -Filter '_CORPUS_STATUS.md'
+  $skillLeaks += Get-ChildItem $r -Recurse -Force -Directory -Filter 'evals'
+}
+if ($skillLeaks.Count -gt 0) {
+  $skillLeaks | ForEach-Object { Write-Host "  LEAK (excluded file present): $($_.FullName)" -ForegroundColor Red }
+  throw "leak-check failed: excluded files present in dist"
+}
+$pointerFails = @()
+foreach ($r in $SkillRoots) {
+  foreach ($skillDir in (Get-ChildItem $r -Directory)) {
+    $body = Join-Path $skillDir.FullName 'SKILL.md'
+    if (-not (Test-Path $body)) { continue }
+    $bodyText = Get-Content $body -Raw
+    foreach ($m in [regex]::Matches($bodyText, '(?:references|evals)/[A-Za-z0-9_./-]+')) {
+      $ptr  = $m.Value.TrimEnd('.', ',', ';', ':', ')')
+      $leaf = Split-Path $ptr -Leaf
+      if ($ptr.EndsWith('/') -or ($leaf -notmatch '\.')) { continue }   # names the directory, not a file
+      if ($ptr -like 'evals/*' -or $leaf -eq '_CORPUS_STATUS.md') {
+        $pointerFails += ("{0} points at {1}, which this script strips from every shipped copy." -f $skillDir.Name, $ptr)
+      } elseif (-not (Test-Path (Join-Path $skillDir.FullName $ptr))) {
+        $pointerFails += ("{0} points at {1}, which is not in the shipped copy." -f $skillDir.Name, $ptr)
+      }
+    }
+  }
+}
+if ($pointerFails.Count -gt 0) {
+  $pointerFails | Sort-Object -Unique | ForEach-Object { Write-Host "  POINTER: $_" -ForegroundColor Red }
+  throw "leak-check failed: a shipped SKILL.md points at a file that does not ship beside it"
+}
+Write-Host ("skill leak-check clean ({0} skills across {1} tree(s))." -f ($SkillRoots | ForEach-Object { (Get-ChildItem $_ -Directory).Count } | Measure-Object -Sum).Sum, $SkillRoots.Count) -ForegroundColor Green
+
 # The validator's whole input is assembled now. Stop here when only that was asked for.
 if ($PluginTreeOnly) {
   $skillCount = (Get-ChildItem $SkillsDir -Directory).Count
   Write-Host ("`nPlugin tree assembled: {0}   skills: {1}" -f $DistRoot, $skillCount) -ForegroundColor Green
+  Write-Host ("Codex skill tree:      {0}" -f $CodexSkills)
   Write-Host "Server, setup utility and zip were skipped (-PluginTreeOnly)."
   return
 }
 
-# ---- 7. package-root extras (START-HERE note + local marketplace.json) -----
+# ---- 8. package-root extras (START-HERE note + local marketplace.json) -----
 # These live in the PACKAGE ROOT (dist/), beside - not inside - dist/housecarl/, so the leak-check's
 # excluded-file scan (scoped to dist/housecarl) leaves them out, while the dev-path scan (step 9) still
 # covers them. Source: packaging/ (tracked). START-HERE.txt is version-stamped from plugin.json;
 # marketplace.json is the local CLI-install descriptor (`claude plugin marketplace add <this folder>`).
-Step '7/11' 'Package-root extras (START-HERE.txt + marketplace.json + codex umbrella)'
+Step '8/12' 'Package-root extras (START-HERE.txt + marketplace.json)'
 $startHere = (Get-Content (Join-Path $PackagingSrc 'START-HERE.txt') -Raw) -replace '\{\{VERSION\}\}', $Version
 if ($startHere -match '\{\{') { throw "START-HERE.txt has an unresolved {{token}} after substitution" }
 [System.IO.File]::WriteAllText((Join-Path $PkgRoot 'START-HERE.txt'), $startHere, (New-Object System.Text.UTF8Encoding($false)))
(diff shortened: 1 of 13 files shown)


scripts/build-plugin.ps1
```diff
@@ -100,62 +104,110 @@ if (-not $PluginTreeOnly) {
...
+  $skillLeaks += Get-ChildItem $r -Recurse -Force -File -Filter '_CORPUS_STATUS.md'
+  $skillLeaks += Get-ChildItem $r -Recurse -Force -Directory -Filter 'evals'
+}
+if ($skillLeaks.Count -gt 0) {
+  $skillLeaks | ForEach-Object { Write-Host "  LEAK (excluded file present): $($_.FullName)" -ForegroundColor Red }
+  throw "leak-check failed: excluded files present in dist"
+}
+$pointerFails = @()
+foreach ($r in $SkillRoots) {
+  foreach ($skillDir in (Get-ChildItem $r -Directory)) {
```
Avick3110, line 179: The scan reads only `SKILL.md`, so the defect it is meant to close still ships. Three shipped reference files under `dialogue-authoring/references/` say "see `_CORPUS_STATUS.md`" — `condition-functions.md:13`, `dialogue-branch.md:8`, `quest-objectives-tab.md:8` — and step 5 deletes `_CORPUS_STATUS.md` from every shipped copy. So a Claude session that opens `condition-functions.md` in the installed plugin is sent to a file that is not there, and this build prints `skill leak-check clean`. Two gaps compound: the loop only opens `SKILL.md`, and the regex requires a `references/` prefix, so a bare `_CORPUS_STATUS.md` inside a reference file would not match even if reference files were read. Either widen the scan to every shipped `.md` under each skill dir and match bare excluded filenames too, or fix those three lines in this PR — as it stands the PR body's "no other body pointer is broken" is true only of `SKILL.md`.

src/housecarl-generator/SetupUpdateLockProbe.cs
```diff
@@ -65,7 +65,7 @@ public static int RunGuard(string[] args)
             WriteFile(Path.Combine(src, "server", "housecarl-mcp.exe"), exeV1);
             WriteFile(Path.Combine(src, "server", "Mutagen.Bethesda.dll"), new byte[] { 9 }); // a sibling DLL (T4)
             WriteFile(Path.Combine(src, "skills", "demo-skill", "SKILL.md"), "demo");
-            WriteFile(Path.Combine(pkg, "codex", "housecarl", "SKILL.md"), "umbrella");
+            WriteFile(Path.Combine(pkg, "codex", "skills", "housecarl", "SKILL.md"), "umbrella");
```
Avick3110: This fixture line is the only thing in the repo that pins the new package path, and nothing asserts the umbrella actually installs. `InstallForCodex` guards the copy with `if (Directory.Exists(umbrellaSrc))`, and this probe never checks `~/.agents/skills/housecarl/SKILL.md` after T1 — it only checks the two server exes. So if the packager path and the installer path ever disagree (revert the `Program.cs` line alone and leave this one, or vice versa), the umbrella is silently not installed, setup prints no `[Codex] installing the houseCARL umbrella skill` line, and `ci-all` stays green. The path change is the whole point of the PR and it is untested. Add to T1:

```suggestion
            WriteFile(Path.Combine(pkg, "codex", "skills", "housecarl", "SKILL.md"), "umbrella");
            string umbrellaDest = Path.Combine(home, ".agents", "skills", "housecarl", "SKILL.md");
```

and a `Check(File.Exists(umbrellaDest), "Codex umbrella skill landed at ~/.agents/skills/housecarl");` beside the two existing T1 checks.

scripts/build-plugin.ps1
```diff
@@ -100,62 +104,110 @@ if (-not $PluginTreeOnly) {
...
   Copy-Item $src (Join-Path $SkillsDir $s) -Recurse -Force
 }
-# prune dev/QA meta from the copies (index.jsonl + the mutagen shards stay - load-bearing)
-Get-ChildItem $SkillsDir -Directory -Recurse -Filter 'evals' | Remove-Item -Recurse -Force
-Get-ChildItem $SkillsDir -File -Recurse -Filter '_CORPUS_STATUS.md' | Remove-Item -Force
+$CodexSrc = Join-Path $PluginSrc 'codex'
+$SkillRoots = @($SkillsDir)
+if (Test-Path $CodexSrc) {
+  New-Item -ItemType Directory -Path $CodexSkills -Force | Out-Null
+  Get-ChildItem $CodexSrc -Directory | ForEach-Object { Copy-Item $_.FullName $CodexSkills -Recurse -Force }
```
Avick3110: `-Directory` silently drops anything that is a file directly under `plugin/codex/`. The old line was `Copy-Item $CodexSrc (Join-Path $PkgRoot 'codex') -Recurse -Force`, which took the whole tree. Today `plugin/codex/` holds only `housecarl/`, so this is green, but the day someone adds `plugin/codex/AGENTS.md` or a `plugin/codex/README.md` it vanishes from the zip with no error and no leak-check hit — a silent packaging loss, the class this script otherwise throws on (`throw "skill not found: $src"` two lines up). Either enumerate files too and fail on an unexpected loose file, or keep the whole-tree copy and just place it under `$CodexSkills`.

scripts/build-plugin.ps1
```diff
@@ -100,62 +104,110 @@ if (-not $PluginTreeOnly) {
...
+  $skillLeaks | ForEach-Object { Write-Host "  LEAK (excluded file present): $($_.FullName)" -ForegroundColor Red }
+  throw "leak-check failed: excluded files present in dist"
+}
+$pointerFails = @()
+foreach ($r in $SkillRoots) {
+  foreach ($skillDir in (Get-ChildItem $r -Directory)) {
+    $body = Join-Path $skillDir.FullName 'SKILL.md'
+    if (-not (Test-Path $body)) { continue }
+    $bodyText = Get-Content $body -Raw
+    foreach ($m in [regex]::Matches($bodyText, '(?:references|evals)/[A-Za-z0-9_./-]+')) {
```
Avick3110: `[regex]::Matches` is case-sensitive and the pattern only accepts forward slashes, so the guard has two blind spots that read as "clean". A body writing ``References/foo.md``, ``Evals/evals.json`` or ``references\\foo.md`` matches nothing and is never checked, while `Test-Path` on Windows would have happily resolved the first two — so the very pointer the ruling is about can be written in a spelling the scan ignores. Since the resolution side is case-insensitive anyway, make the match agree with it: `[regex]::Matches($bodyText, '(?:references|evals)[/\\][A-Za-z0-9_./\\-]+', 'IgnoreCase')`, and normalise `\\` to `/` before the `-like 'evals/*'` test.

scripts/build-plugin.ps1
```diff
@@ -165,19 +217,19 @@ if (Test-Path $CodexSrc) { Copy-Item $CodexSrc (Join-Path $PkgRoot 'codex') -Rec
...
 Write-Host ("houseCARL-Setup.exe: {0:N2} MB" -f ((Get-Item $SetupExe).Length / 1MB))
 
-# ---- 9. leak-check ---------------------------------------------------------
-Step '9/11' 'Leak-check assembled tree'
+# ---- 10. leak-check the server half + the whole package root ----------------
+# The skill trees were leak-checked at step 7 (excluded files + body pointers); this covers what only
+# a full build produces: the published server, and every shipped text file under the package root.
+Step '10/12' 'Leak-check assembled tree'
 $leaks = @()
 $leaks += Get-ChildItem $DistRoot -Recurse -Force -Filter 'appsettings*.json'
```
Avick3110: Moving the `_CORPUS_STATUS.md` / `evals` scan to step 7 narrowed its scope from all of `dist/housecarl` to `dist/housecarl/skills` plus `dist/codex/skills`. Step 7 also runs *before* nothing relevant — the server publish (steps 2-4) has already written `dist/housecarl/server`, and step 6 copies the plugin files — yet neither is covered by either scan any more: the only thing left here is `appsettings*.json`. Concrete gap: if a `_CORPUS_STATUS.md` or an `evals/` directory ever lands outside `skills/` (a stray copy under `.claude-plugin/`, or one pulled in by a future publish), the full build now ships it and prints `leak-check clean.` where it previously threw. Cheap fix: keep the two `$DistRoot` lines here as well — the skill-scoped check at step 7 is what CI needs, this one is the backstop.

scripts/build-plugin.ps1
```diff
@@ -100,62 +104,110 @@ if (-not $PluginTreeOnly) {
...
+if ($skillLeaks.Count -gt 0) {
+  $skillLeaks | ForEach-Object { Write-Host "  LEAK (excluded file present): $($_.FullName)" -ForegroundColor Red }
+  throw "leak-check failed: excluded files present in dist"
+}
+$pointerFails = @()
+foreach ($r in $SkillRoots) {
+  foreach ($skillDir in (Get-ChildItem $r -Directory)) {
+    $body = Join-Path $skillDir.FullName 'SKILL.md'
+    if (-not (Test-Path $body)) { continue }
+    $bodyText = Get-Content $body -Raw
```
Avick3110: `Get-Content -Raw` on a zero-byte `SKILL.md` returns `$null`, and `[regex]::Matches($null, ...)` throws `ArgumentNullException`. With `$ErrorActionPreference = 'Stop'` the build dies on a .NET stack trace instead of the one plain sentence this script's other failures give (`throw "skill not found: ..."`). A truncated or empty `SKILL.md` is exactly the kind of mistake this scan exists to catch, and it is the one input that makes it crash rather than report. Guard it: `if (-not $bodyText) { $pointerFails += ("{0} has an empty SKILL.md." -f $skillDir.Name); continue }`.

scripts/build-plugin.ps1
```diff
@@ -100,62 +104,110 @@ if (-not $PluginTreeOnly) {
...
+  $pointerFails | Sort-Object -Unique | ForEach-Object { Write-Host "  POINTER: $_" -ForegroundColor Red }
+  throw "leak-check failed: a shipped SKILL.md points at a file that does not ship beside it"
+}
+Write-Host ("skill leak-check clean ({0} skills across {1} tree(s))." -f ($SkillRoots | ForEach-Object { (Get-ChildItem $_ -Directory).Count } | Measure-Object -Sum).Sum, $SkillRoots.Count) -ForegroundColor Green
+
 # The validator's whole input is assembled now. Stop here when only that was asked for.
 if ($PluginTreeOnly) {
   $skillCount = (Get-ChildItem $SkillsDir -Directory).Count
   Write-Host ("`nPlugin tree assembled: {0}   skills: {1}" -f $DistRoot, $skillCount) -ForegroundColor Green
+  Write-Host ("Codex skill tree:      {0}" -f $CodexSkills)
```
Avick3110: This prints unconditionally, but `$CodexSkills` is only created inside `if (Test-Path $CodexSrc)` at step 5. Delete or rename `plugin/codex/` and a `-PluginTreeOnly` run — the shape CI uses — still reports `Codex skill tree: ...\dist\codex\skills` for a directory that does not exist, and nothing else in the run mentions the omission (`$SkillRoots` just stays one element, and the leak-check happily says "1 tree(s)"). Gate the line on `Test-Path $CodexSkills`, or say "Codex skill tree: (none — plugin/codex not found)".

scripts/build-plugin.ps1
```diff
@@ -100,62 +104,110 @@ if (-not $PluginTreeOnly) {
...
   Write-Host "Server, setup utility and zip were skipped (-PluginTreeOnly)."
   return
 }
 
-# ---- 7. package-root extras (START-HERE note + local marketplace.json) -----
+# ---- 8. package-root extras (START-HERE note + local marketplace.json) -----
 # These live in the PACKAGE ROOT (dist/), beside - not inside - dist/housecarl/, so the leak-check's
 # excluded-file scan (scoped to dist/housecarl) leaves them out, while the dev-path scan (step 9) still
```
Avick3110: Two stale claims left by the renumber, both about a safety property a future editor would rely on. "the leak-check's excluded-file scan (scoped to dist/housecarl)" is no longer true — the step 7 scan is scoped to the two skill roots, and `dist/codex` is now inside its scope, not outside it; and "the dev-path scan (step 9)" now points at step 9's *setup-utility publish*, not the dev-path scan, which moved to step 10. Same class at line 214, which still says the leak-check "scans dist\housecarl only" while the dev-path scan walks `$PkgRoot`. Worth fixing while the numbering is being touched.

scripts/build-plugin.ps1
Avick3110: Not folding this one. `plugin/codex/` has no loose files and none are planned: `git ls-files plugin/codex` returns exactly `plugin/codex/housecarl/SKILL.md` and `plugin/codex/housecarl/agents/openai.yaml`, and the only thing the backlog and the rulings say about that tree (item 27, and the 2026-09-07 rewrite ruling 7) is the umbrella skill itself.

`-Directory` is also the right enumeration for what this now copies into, not an accident of the rewrite. The destination is a `skills/` root, and Codex reads a `skills/` root as skill directories; a loose `plugin/codex/AGENTS.md` copied there would be a stray file inside the skills root, not a shipped document. The old whole-tree copy landed at `dist/codex/`, where a loose file was harmless; `dist/codex/skills/` is a different kind of directory. So if someone does add a package-level Codex file later, the fix is a line placing it at `dist/codex/`, not widening this copy.

Adding a throw for a loose file that has never existed would be a guard against a hypothetical mistake, which CLAUDE.md tells us not to write.

scripts/build-plugin.ps1
```diff
@@ -100,84 +104,151 @@ if (-not $PluginTreeOnly) {
...
+$docCount = 0
+foreach ($r in $SkillRoots) {
+  foreach ($skillDir in (Get-ChildItem $r -Directory)) {
+    foreach ($doc in (Get-ChildItem $skillDir.FullName -Recurse -Force -File -Filter '*.md')) {
+      $docCount++
+      $rel     = ($doc.FullName.Substring($skillDir.FullName.Length).TrimStart('\')) -replace '\\','/'
+      $docText = Get-Content $doc.FullName -Raw
+      if (-not $docText) { $pointerFails += ("{0} ships an empty {1}." -f $skillDir.Name, $rel); continue }
+      # a references/ or evals/ pointer in any spelling, plus a bare mention of the stripped status note
+      $hits  = @([regex]::Matches($docText, '(?:references|evals)[/\\][A-Za-z0-9_./\\-]+', 'IgnoreCase') | ForEach-Object { $_.Value })
```
Avick3110: **The bare sibling pointer — the dominant form in the corpus — is never checked.**

The scan matches `references/<file>`, `evals/<file>`, and the bare name `_CORPUS_STATUS.md`. But most pointers in the shipped skills are written as a bare sibling filename, not with the `references/` hop. In this branch's shipped set that is roughly 250 mentions — 57 of `value-tables.md`, 47 of `grammar-core.md`, 18 each of `plugin-skeleton.md` and `multi-runtime.md`, 16 of `filters.md`, and so on. `kid-authoring/SKILL.md` writes both forms in the same file: `references/filters.md` at line 41 (checked) and a bare `filters.md` at line 65 (not checked).

Scenario: a later skill-rewrite PR renames or drops `.claude/skills/kid-authoring/references/value-tables.md`. Step 7 still reports `skill leak-check clean`, the build goes green, and 57 dead pointers ship in the installed plugin — the failure this check exists to stop.

Adding the bare form is a third regex plus a resolve that tries both the skill root and the document's own directory. If it is out of scope for wave 0, the scan message and the PR's invariant should say so — as written, "every `references/<file>` a shipped `.md` writes" reads as if pointer coverage is closed.

src/housecarl-generator/SetupUpdateLockProbe.cs
```diff
@@ -65,14 +70,15 @@ public static int RunGuard(string[] args)
...
```
Avick3110: **T1 pins the installer path only — it cannot make the packager agree.**

The probe writes its own package fixture at `codex/skills/housecarl` here, then asserts the umbrella lands at `~/.agents/skills/housecarl`. That constrains `Program.cs`. Nothing in the probe reads `scripts/build-plugin.ps1`, so the packager side is still unpinned — but the new comments at line 16 and line 59, and the PR body ("the packager path and the installer path must agree", "tested, not just written twice"), claim the pair is now held together.

Scenario: a later PR changes `$CodexSkills` in the packager to, say, `dist/codex/agents/housecarl` and leaves `Program.cs` alone. `ci-all` stays green — the probe writes its own umbrella at the hardcoded `codex/skills/housecarl` and the installer still reads exactly that. The shipped zip now carries the umbrella where `Directory.Exists(umbrellaSrc)` will not find it, the guard skips the copy in silence, and Codex users get no `$housecarl` entry point. That is the same silent failure the PR says is now proven to turn red.

Deriving the fixture path from the script (grep `$CodexSkills`/`$CodexRoot` out of `scripts/build-plugin.ps1`, the way `DescriptionVocabularyGuardProbe` derives the published trees) would make the claim true. Failing that, the two comments should say the probe pins the installer only.

scripts/build-plugin.ps1
```diff
@@ -100,84 +104,151 @@ if (-not $PluginTreeOnly) {
...
-# ---- 9. leak-check ---------------------------------------------------------
-Step '9/11' 'Leak-check assembled tree'
+# ---- 10. leak-check the server half + the whole package root ----------------
+# The skill trees were leak-checked at step 7 (excluded files + markdown pointers); this covers what
+# only a full build produces: the published server, and every shipped text file under the package
+# root. The excluded-file scan walks the whole package root, so an excluded file that lands outside
+# skills/ - a stray copy under .claude-plugin/, or one a future publish pulls in - is still caught.
+Step '10/12' 'Leak-check assembled tree'
 $leaks = @()
 $leaks += Get-ChildItem $DistRoot -Recurse -Force -Filter 'appsettings*.json'
```
Avick3110: **This line stayed `$DistRoot` while the comment above it now claims package-root coverage.**

The two lines below were widened to `$PkgRoot`; the `appsettings*.json` line was not. But the new step-10 comment reads "The excluded-file scan walks the whole package root, so an excluded file that lands outside `skills/` ... is still caught" — which is only true for `_CORPUS_STATUS.md` and `evals/`.

It matters because step 9 publishes straight into `$PkgRoot`. Scenario: the setup utility picks up a dependency (or an SDK change) that emits `appsettings.json` next to `houseCARL-Setup.exe`. It lands at `dist/appsettings.json`, outside `$DistRoot`, so nothing sees it; step 11 packs it as `houseCARL/appsettings.json` and the build prints `leak-check clean.` — the appsettings class of leak, shipped, from the one publish the widened scan was supposed to cover.

scripts/build-plugin.ps1
```diff
@@ -100,84 +104,151 @@ if (-not $PluginTreeOnly) {
...
   Copy-Item $src (Join-Path $SkillsDir $s) -Recurse -Force
 }
-# prune dev/QA meta from the copies (index.jsonl + the mutagen shards stay - load-bearing)
-Get-ChildItem $SkillsDir -Directory -Recurse -Filter 'evals' | Remove-Item -Recurse -Force
-Get-ChildItem $SkillsDir -File -Recurse -Filter '_CORPUS_STATUS.md' | Remove-Item -Force
+$CodexSrc = Join-Path $PluginSrc 'codex'
+$SkillRoots = @($SkillsDir)
+if (Test-Path $CodexSrc) {
+  New-Item -ItemType Directory -Path $CodexSkills -Force | Out-Null
+  Get-ChildItem $CodexSrc -Directory | ForEach-Object { Copy-Item $_.FullName $CodexSkills -Recurse -Force }
```
Avick3110: **This narrows what ships from `plugin/codex`, silently.**

The old line copied the whole `plugin/codex` tree. This enumerates directories only — and without `-Force` — so a file at the root of `plugin/codex`, or a hidden directory under it, is dropped from the package with no throw and no warning.

Nothing sits there today (`plugin/codex/` holds only `housecarl/`), so this is latent rather than broken. But the trap is quiet: add `plugin/codex/README.md`, or a shared `plugin/codex/agents.yaml` beside `housecarl/`, and it never reaches `dist/codex/` or the zip, while step 7 prints `skill leak-check clean` and the summary still says `codex skill: True`.

Copying the tree and then moving the skill dirs under `skills/` — or at least throwing when `Get-ChildItem $CodexSrc -File` is non-empty — keeps the packager honest about what it drops.

scripts/build-plugin.ps1
```diff
@@ -100,84 +104,151 @@ if (-not $PluginTreeOnly) {
...
+foreach ($r in $SkillRoots) {
+  foreach ($skillDir in (Get-ChildItem $r -Directory)) {
+    foreach ($doc in (Get-ChildItem $skillDir.FullName -Recurse -Force -File -Filter '*.md')) {
+      $docCount++
+      $rel     = ($doc.FullName.Substring($skillDir.FullName.Length).TrimStart('\')) -replace '\\','/'
+      $docText = Get-Content $doc.FullName -Raw
+      if (-not $docText) { $pointerFails += ("{0} ships an empty {1}." -f $skillDir.Name, $rel); continue }
+      # a references/ or evals/ pointer in any spelling, plus a bare mention of the stripped status note
+      $hits  = @([regex]::Matches($docText, '(?:references|evals)[/\\][A-Za-z0-9_./\\-]+', 'IgnoreCase') | ForEach-Object { $_.Value })
+      $hits += @([regex]::Matches($docText, '(?<![A-Za-z0-9_./\\-])_CORPUS_STATUS\.md', 'IgnoreCase') | ForEach-Object { $_.Value })
```
Avick3110, line 190: **This regex has the left-boundary guard the one above it lacks.**

This line guards its left edge with `(?<![A-Za-z0-9_./\-])`; the `(?:references|evals)[/\\]…` regex on the previous line does not. So it also matches the tail of any longer word that happens to end in `references` or `evals` and is followed by a slash.

Scenario: a doc writes `retrievals/cache.json` or `cross-references/glossary.md` — in prose or inside a fence. The first matches as `evals/cache.json` and is reported as "points at evals/cache.json, which this script strips from every shipped copy"; the second matches as `references/glossary.md` and fails `Test-Path`. Either one hard-stops the build on a correct document, naming a path that does not appear in the file.

Nothing in the corpus trips it today (verified: the scan is clean over all 255 shipped markdown files), so this is prevention, not a live break. The same lookbehind on the previous line closes it.

scripts/build-plugin.ps1
Avick3110: Folded in 80245197, and extended here. The scan reads every `.md` under each skill dir, not just `SKILL.md` (build-plugin.ps1:183), and matches a bare `_CORPUS_STATUS.md` as well as the hop form (line 190). The three `dialogue-authoring/references/` lines you named were dropped in f0dc1a35. Beyond that finding, 0073c265 adds the bare sibling form generally, so a bare pointer at any shipped file is now resolved too.

scripts/build-plugin.ps1
Avick3110: Folded in 80245197. build-plugin.ps1:189 is now `'(?<![A-Za-z0-9_./\\-])(?:references|evals)[/\\][A-Za-z0-9_./\\-]+', 'IgnoreCase'` — either slash, any case. A backslash pointer is reported as its own defect rather than normalised (line 194), since skill pointers are written with forward slashes; that is why there is no `-replace` before the `-like 'evals/*'` test. Proven red on `references\backslash-form.md` and on `References/also-missing.md`.

scripts/build-plugin.ps1
Avick3110: Folded in 80245197. build-plugin.ps1:187 guards it: `if (-not $docText) { $pointerFails += ("{0} ships an empty {1}." -f $skillDir.Name, $rel); continue }`. A zero-byte `SKILL.md` now reports `oar-authoring ships an empty SKILL.md.` and stops the build with the same one-sentence shape as the rest, instead of an ArgumentNullException stack.

scripts/build-plugin.ps1
Avick3110: Folded in 1baa6ee0. build-plugin.ps1:228-229 gates it: it prints the path when `Test-Path $CodexSkills`, and `Codex skill tree:      (none - plugin/codex not found)` otherwise.

scripts/build-plugin.ps1
Avick3110: Folded in 1baa6ee0. The step-8 comment (build-plugin.ps1:234-235) now says step 10 covers the package-root extras both ways, and the step-10 comment (lines 266-268) says all three of its excluded-file scans walk the whole package root — true as of 8c832883, which widened the last one.

scripts/build-plugin.ps1
Avick3110: Folded in 0073c265 — the resolution half. A bare sibling name in any shipped `.md` now resolves against every file name that skill ships; one that matches nothing is flagged with the same one-sentence message the hop form gets. `kid-authoring/SKILL.md:65`'s bare `filters.md` is checked now, and so are the ~250 mentions of `value-tables.md`, `grammar-core.md` and the rest — rename or drop one of those in a later rewrite PR and step 7 goes red instead of shipping the dead pointers.

Two bounds, both deliberate.

**Form is not checked here.** This scan says a pointer *resolves*. The FORM rule — that a pointer is written `references/<file>`, one hop (`dev/DECISIONS.md` 2026-09-07, rewrite ruling 2) — is enforced per skill by the regrade in that skill's rewrite wave, not by this scan today, because flagging form now would go red on every skill not yet rewritten. The scan comment and the PR body now say exactly that, so "pointer coverage" is not read as closed.

**Only `.md` and `.jsonl` bare names are candidates.** Those are the only extensions any shipped skill file has (254 `.md`, 11 `.jsonl`, nothing else), so they are the only names a bare pointer could resolve to. Taking `.json`, `.ini` and `.psc` as pointers too would hard-stop the build on 143 correct documents: OAR's `config.json` / `user.json` (58), SPID's `_DISTR.ini` (13), vanilla `Form.psc` / `UI.psc` / `Input.psc` (13), `vcpkg.json` (13), `SkyPatcher.ini` (6), `Skyrim.ini`, `meta.ini`, `NPC_Token.json`. In these skills a bare `.ini`/`.psc`/`.json` name is a file in the modded game, never a sibling document. The comment at the regex says so.

Arms proven on `oar-authoring/references/oar-config-reference.md`, six red in one run:

```
SKILL DOC: oar-authoring (references/oar-config-reference.md) points at _CORPUS_STATUS.md, which this script strips from every shipped copy.
SKILL DOC: oar-authoring (references/oar-config-reference.md) points at evals/evals.json, which this script strips from every shipped copy.
SKILL DOC: oar-authoring (references/oar-config-reference.md) points at gone-away.jsonl, which is not in the shipped copy.
SKILL DOC: oar-authoring (references/oar-config-reference.md) points at gone-away.md, which is not in the shipped copy.
SKILL DOC: oar-authoring (references/oar-config-reference.md) points at references/does-not-exist.md, which is not in the shipped copy.
SKILL DOC: oar-authoring (references/oar-config-reference.md) writes references\backslash-form.md with a backslash; skill pointers use forward slashes.
```

Green in the same run: a bare `oar-config-reference.md` (resolves, silent), and the decoys `retrievals/cache.json`, `cross-references/glossary.md`, `config.json`, `Form.psc`, `_DISTR.ini` — none of them trip.

What it flagged on today's skills, all four fixed in 1d11d3ec, each a shipped document naming another project's file by a bare name that reads as a sibling:

- `skse-plugin-authoring/references/multi-runtime.md` (18 mentions) and `references/plugin-dissection.md` (2) cite CommonLibSSE-NG's bundled `CLAUDE.md`. They now write `ng/CLAUDE.md`, the qualified form `references/hooking.md` in the same skill already used four times.
- `skse-plugin-authoring/references/hooking.md:466` cites `VERIFICATION_MIGRATION.md`; it now writes `docs/VERIFICATION_MIGRATION.md`, the form line 445 of the same file already used.
- `tool-output-awareness/SKILL.md:136` said "the modlist's own `CLAUDE.md` / project memory"; the bare name is dropped and the sentence keeps "the modlist's own project memory".

No dead sibling pointer at a file that used to ship: nothing in the corpus has been renamed out from under a bare mention yet. That is the case the check now stops. Listed in the PR body.

src/housecarl-generator/SetupUpdateLockProbe.cs
Avick3110: Folded in d3614bea, the first way — the probe now reads the package path out of `scripts/build-plugin.ps1` instead of restating it, so the claim is true rather than softened. Twelve lines.

`CodexPackagePath()` reads the script's text and takes the two assignments that compose the path — `$CodexRoot = Join-Path $PkgRoot 'codex'` and `$CodexSkills = Join-Path $CodexRoot 'skills'` — and returns `codex/skills`. T1 writes its fixture at that path, then asserts the umbrella lands at `~/.agents/skills/housecarl`. Read rather than run, for the same reason `DescriptionVocabularyGuardProbe` reads it: the answer is two lines of the script, and running a packaging build to learn it would cost minutes.

Your scenario now reds. Changing `$CodexSkills` to `agents` in the packager and leaving `Program.cs` alone:

```
PASS  packager's Codex umbrella path read from scripts/build-plugin.ps1: codex/agents
FAIL  Codex umbrella skill packed at codex/agents/housecarl landed at ~/.agents/skills/housecarl
================ 1 CHECK(S) FAILED ================
```

The check names both sides, so the failure says which pair disagreed. Reverting the installer's path alone still reds too (unchanged from before).

A script the probe cannot read, or an assignment that has gone away, is its own failing check with the reason (`not readable from '<cwd>' — run ci-all from the repo root`, or `no '$CodexSkills = Join-Path $CodexRoot ...' assignment found`) — it never certifies the path pair from its own restatement. The class-2 caveat applies as it does to the vocabulary guard: this reads PowerShell by pattern, and it cannot hide a shortfall, because failing to read the path fails the probe.

The comments at line 16 and the PR body are rewritten to say what is now pinned: the pair, both directions.

scripts/build-plugin.ps1
Avick3110: Folded in 800b62d5, and my earlier refutation on the sibling thread is withdrawn on your word.

The whole `plugin/codex` tree ships again. `Get-ChildItem $CodexSrc -Directory -Force` puts every directory, hidden ones included, under `dist/codex/skills/`; `Get-ChildItem $CodexSrc -File -Force` puts any loose file at `dist/codex/`, the package-level Codex location. Nothing in that tree is dropped without a line for it.

The leak scan covers the result. The excluded-file scan and the strip pass now walk a wider root list (`$LeakRoots` = the plugin skills dir plus the Codex package root) than the pointer scan does (`$SkillRoots` = the two skill-directory roots), so a `_CORPUS_STATUS.md` or an `evals/` directory arriving as a loose Codex file is stripped and then checked, not only what lands under `skills/`.

Full build after the change: `skill leak-check clean (15 skills, 255 markdown files, across 2 tree(s))`, `leak-check clean.`, zip packs 341 entries with `houseCARL/codex/skills/housecarl/SKILL.md`.

scripts/build-plugin.ps1
Avick3110: Folded in 8c832883. build-plugin.ps1:272 is now `$leaks += Get-ChildItem $PkgRoot -Recurse -Force -Filter 'appsettings*.json'`, so all three of step 10's excluded-file checks walk the package root and the comment above them is true of each. Your scenario — the step-9 setup-utility publish emitting `appsettings.json` beside `houseCARL-Setup.exe` at `dist/appsettings.json` — now fails the build instead of being packed as `houseCARL/appsettings.json`. The comment names that publish as the case it covers rather than leaving the reader to infer it.

scripts/build-plugin.ps1
Avick3110: Folded in 9d1ac741. build-plugin.ps1:189 now carries the same `(?<![A-Za-z0-9_./\\-])` lookbehind as the line below it, so `retrievals/cache.json` and `cross-references/glossary.md` no longer match as `evals/cache.json` and `references/glossary.md`. Both are in the red-arm probe as decoys and stay silent while the six real defects fire.

scripts/build-plugin.ps1
Avick3110: Withdrawn — folded in 800b62d5 on Aaron's word, after he raised it a second time on the sibling thread. The whole `plugin/codex` tree is copied again: every directory (hidden ones included) under `dist/codex/skills/`, any loose file at `dist/codex/`. The excluded-file scan walks the Codex package root, so a loose file is checked too. My argument above stands only as the reason `dist/codex/skills/` is the wrong destination for a loose file, which is why it now lands beside that directory rather than inside it.

src/housecarl-generator/SetupUpdateLockProbe.cs
Avick3110: Folded in ec604ba0: T1 now asserts `~/.agents/skills/housecarl/SKILL.md` exists after the clean install, beside the two exe checks, and e19e8288 says in the probe summary what that check proves. Reverting the installer's path alone turns it red. d3614bea then closes the other half you raised below — the fixture path T1 installs from is read out of `scripts/build-plugin.ps1`, so moving the packager alone reds too. Details on that thread.

PR #622, opened 2026-09-07
Resolve a generic type's getter interface, arity and all

### LINKED ISSUE #424: [Bug] GetterInterfaceFor resolves a getter interface by name only — generic arity and not-own-named shapes return null for extractable types

Avick3110:
`CorpusGenerator.GetterInterfaceFor` built its probe name straight off `Type.Name`, which carries the generic arity, so for a generic class it asked for the impossible ``IFoo`1Getter``, and both `GetInterfaces` fallbacks failed for the same reason — a generic getter interface is named ``IFooGetter`1``, whose name does not end in "Getter". Every generic modeled type therefore answered "no getter interface" in a walk whose whole claim is that coverage is Mutagen's coverage by construction. `MutableInterfaceFor` already carries this exact fix, with a comment naming `IGenderedItem<T>`; this is the same fix on its twin. The probe name now lives in one helper (`GetterProbeName`) that both the resolver and the anomaly line read, so the line reports the name that was actually looked up. The implemented interface is also asked before the assembly lookup, so a closed generic resolves the closed `IFooGetter<Bar>` rather than the open definition `Assembly.GetType` can only ever hand back.

The issue's second half — whether a class implementing a getter interface named after *another* type should resolve — is answered **no**, deliberately, and the code says why. That shape is exactly Mutagen's read-only projections (`SkyrimMultiModOverlay` implements `ISkyrimModDisposableGetter`, `MergedWorldspace` implements `IWorldspaceGetter`, `MergedCellBlock` implements `ICellBlockGetter`, and the `Consolidated*` / `Merged*` / `*Wrapper` group views likewise), and "owns no getter interface" is the structural signal `ClassifyArm` reads to spot one. Measured: widening the resolver to accept any implemented getter interface resolves one for 32 further types, 31 of which are borrowed-name shapes dominated by those read-only views. `SkyrimMultiModOverlay` and `MergedCellBlock` — the two exhibits this guard already pins as `ReadOnlyProjection` — flip to `Authorable` under that widening, i.e. re-enter as arms of unions they can never be composed into. That is a straight regression of #397's split, so the resolver stays own-named.

The cost of the narrow answer is the 32nd type: `ArmorAddonWeightSliderContainer` is not a projection, implements `IGenderedItemGetter<bool>`, and still resolves null. Its data is catalogued anyway, as `GenderedItem<Boolean>`, which is exactly why `UnextractableWarning` reports rather than diagnoses. The comments that called this an outstanding defect in `GetterInterfaceFor` now call it a settled boundary.

The fix also makes `ClassifyArm`'s `?? t` fallback unreachable. It was rescuing ``FormLinkNullableGetter`1`` and ``FormLinkOrIndexGetter`1``, whose own names end in "Getter" — both were failing to resolve their real getter interfaces for the arity reason, so that fallback was compensating for this bug. Measured across all three assemblies the walk visits: no class now reaches that line with no getter interface and a mutable twin of its own name. Removed rather than left as a dead branch under a comment saying it is load-bearing.

## The corpus diff, read

**There is none.** A full emit before and after the fix is byte-identical: `corpus.json`, `corpus.summary.md`, and all six committed `mutagen-reference` shards (`index`, `records`, `structs`, `arms`, `polymorphic`, `enums`). The generator's printed report is identical too, coverage-anomaly section included — still the same two lines, `MergedWorldspace` and `MergedWorldspaceCell`. Nothing was regenerated because nothing changed, and `emit-match-guard` is green against the committed shards untouched.

Read record type by record type, the answer is that **no record type's schema moved**, and here is why, from the measurement rather than from reasoning. Sixteen types resolve differently after the fix, and every one of them is an **open generic type definition**:

| type | before | after |
|---|---|---|
| ``SkyrimGroup`1`` | null | ``ISkyrimGroupGetter`1`` |
| ``SkyrimListGroup`1`` | null | ``ISkyrimListGroupGetter`1`` |
| ``GenderedItem`1`` | null | ``IGenderedItemGetter`1`` |
| ``FormLinkNullable`1`` | null | ``IFormLinkNullableGetter`1`` |
| ``FormLinkNullableGetter`1`` | null | ``IFormLinkNullableGetter`1`` |
| ``FormLinkOrIndex`1`` | null | ``IFormLinkOrIndexGetter`1`` |
| ``FormLinkOrIndexGetter`1`` | null | ``IFormLinkOrIndexGetter`1`` |
| ``SortingListDictionary`2`` | null | ``ISortingListDictionaryGetter`2`` |
| ``FormLink`1`` | `IFormLinkGetter` | ``IFormLinkGetter`1`` |
| ``FormLinkGetter`1`` | `IFormLinkGetter` | ``IFormLinkGetter`1`` |
| ``EDIDLink`1`` | `IEDIDLinkGetter` | ``IEDIDLinkGetter`1`` |
| ``AssetLink`1`` | `IAssetLinkGetter` | ``IAssetLinkGetter`1`` |
| ``AssetLinkGetter`1`` | `IAssetLinkGetter` | ``IAssetLinkGetter`1`` |
| ``MajorRecordIdentifier`1`` | `IMajorRecordIdentifierGetter` | ``IMajorRecordIdentifierGetter`1`` |
| ``LoadOrder`1`` | `ILoadOrderGetter` | ``ILoadOrderGetter`1`` |
| ``ModListing`1`` | `IModListingGetter` | ``IModListingGetter`1`` |

The five in the issue's table are all here, plus eleven the issue did not measure. That second group is types that *did* resolve before, through the normalized fallback, but resolved the non-generic same-named interface instead of the arity-matched one — `FormLink<T>` implements both `IFormLinkGetter<T>` and `IFormLinkGetter`, and the old code picked the latter. The arity-matched answer is the correct one, and it also does not move the corpus.

None of the sixteen reaches emission, for the reason the issue gives: `FindUnionArms` filters candidates with `getterIfc.IsAssignableFrom(t)`, and `Assembly.GetTypes()` yields open definitions, for which that is always false. Closed generics never appear in `GetTypes()`, and a closed generic sitting in a field position reaches the walk as its *getter interface* (`IGenderedItemGetter<bool>`), which `GetterInterfaceFor` returns unchanged on its first line. So the fix is exactly as latent as the issue said, and it lands with the corpus provably still.

That is also the point of landing it now. The next Mutagen release that adds a generic or differently-named modelled type is the run where this stops being latent, and it should walk into a resolver that answers correctly rather than into a name match.

## Tests

Three new arms on `arm-classification-guard`, the existing guard over this classifier: a generic type resolves its arity-matched getter interface (E1); the record-group container does too (E2), which matters because `IsList` hard-codes ``ISkyrimGroupGetter`1`` by full name and a resolver disagreeing with that list is the generator contradicting itself; and a closed generic resolves the closed interface rather than the open definition (E3). All three are red before the fix and green after — measured by reverting `CorpusGenerator.cs` and rerunning the guard.

The own-named boundary deliberately gets no arm of its own. Widening the resolver turns B4 red (its exhibit would resolve `IGenderedItemGetter<bool>`) and D0b red (both record-path exhibits would resolve a borrowed getter interface), so it is already pinned from two directions and a third check could not fail today. The guard header says that rather than leaving a reader to work it out.

This is generator-internal reflection with no server-side surface, so it is pinned where the rest of this classifier is pinned rather than in `housecarl-mcp-tests`, which the generator does not expose internals to.

Ran: `dotnet build housecarl.sln -c Release`; `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` (1767 passed); `ci-all` (127/127).

No `plugin/CHANGELOG.md` line: the emitted corpus and the shipped reference shards are byte-identical, so there is nothing a user would notice.

## Not done

The issue's "also in scope" item — collapsing `WritableSurfaceCount`'s local property walk back into `CollectAllProperties` — is conditional on that shared helper being tightened, which this change does not do. Left alone.

Closes #424

### DIFF
--- a/src/housecarl-generator/CorpusGenerator.cs
+++ b/src/housecarl-generator/CorpusGenerator.cs
@@ -558,10 +558,11 @@ public static class CorpusGenerator
         /// <summary>A read-only PROJECTION of a real writable type (Mutagen's multi-mod overlay, the
         /// merged-cell view). Nothing is lost by excluding it — it can never be composed.</summary>
         ReadOnlyProjection,
-        /// <summary>No <c>I{Name}Getter</c> resolved BY NAME, yet the class carries settable state. Reported,
-        /// not diagnosed: the name probe misses a type whose getter interface is generic or differently
-        /// named, so this verdict does NOT establish a coverage gap — only that a human should check whether
-        /// the data is reachable elsewhere in the catalogue.</summary>
+        /// <summary>No OWN-named getter interface resolved, yet the class carries settable state. Reported,
+        /// not diagnosed: <see cref="GetterInterfaceFor"/> deliberately answers only for the interface a class
+        /// owns, so a type borrowing another type's getter interface lands here, and this verdict does NOT
+        /// establish a coverage gap — only that a human should check whether the data is reachable elsewhere
+        /// in the catalogue.</summary>
         WritableButUnextractable,
     }
 
@@ -588,9 +589,8 @@ public static class CorpusGenerator
     /// <see cref="EmittedArmNames"/>.
     ///
     /// THE SPLIT. When a getter interface EXISTS, authorability is whether its mutable twin does; no twin
-    /// means a read-only projection, the correct exclusion. When NO getter interface exists, handing the
-    /// concrete class to <see cref="MutableInterfaceFor"/> always answers "excluded" — its
-    /// <c>EndsWith("Getter")</c> test a concrete class never passes — which makes "has no getter interface"
+    /// means a read-only projection, the correct exclusion. When NO getter interface exists there is nothing to
+    /// take a mutable twin of, so the type is excluded either way — which would make "has no getter interface"
     /// indistinguishable in the output from "is a read-only projection", though only the second is a correct
     /// reason to exclude. So split here on whether the class actually carries writable state.
     /// </summary>
@@ -598,12 +598,13 @@ public static class CorpusGenerator
     {
         var gi = GetterInterfaceFor(t);
 
-        // The `?? t` fallback is load-bearing, not the dead branch it looks like. A concrete class whose OWN name
-        // ends in "Getter" (FormLinkNullableGetter`1, FormLinkOrIndexGetter`1 in Mutagen.Bethesda.Core) passes
-        // MutableInterfaceFor's arity-stripped EndsWith test and resolves a mutable twin through it, so it is
-        // authorable and must stay so. Computing this from `gi` alone silently flips both types, and both are in
-        // the walk because BuildCorpus seeds IPexFileGetter from that same assembly.
-        if (MutableInterfaceFor(gi ?? t) != null) return ArmClass.Authorable;
+        // This used to read MutableInterfaceFor(gi ?? t), the `?? t` rescuing two concrete classes whose own
+        // names end in "Getter" — FormLinkNullableGetter`1 and FormLinkOrIndexGetter`1 in Mutagen.Bethesda.Core,
+        // in the walk because BuildCorpus seeds IPexFileGetter from that assembly. It was compensating for the
+        // arity bug in GetterInterfaceFor (#424): both DO own a getter interface, and now resolve it. Measured
+        // across all three assemblies the walk visits, no class reaches this line with no getter interface and a
+        // mutable twin of its own name, so the fallback is gone rather than left as a dead branch.
+        if (gi != null && MutableInterfaceFor(gi) != null) return ArmClass.Authorable;
 
         // Not authorable. Everything below only decides WHICH exclusion to report — never whether to exclude.
         if (gi != null) return ArmClass.ReadOnlyProjection;
@@ -680,11 +681,13 @@ public static class CorpusGenerator
     /// gap, and does NOT say the fix belongs upstream in Mutagen — neither is established by "no
     /// <c>I{Name}Getter</c> resolved by name", and the second is provably false for at least one type that
     /// reaches this line: ArmorAddonWeightSliderContainer implements <c>IGenderedItemGetter&lt;bool&gt;</c> and
-    /// its data ships in the reference today as <c>GenderedItem&lt;Boolean&gt;</c>, 2/2 writable. The name probe
-    /// is <c>I{Name}Getter</c> built from <c>Type.Name</c>, which carries the generic arity, so a type whose
-    /// getter interface is named differently — or is generic — resolves null here while being perfectly
-    /// extractable. That is a real defect in <see cref="GetterInterfaceFor"/>; until it is fixed this line must
-    /// not diagnose, only report, and hand the reader the one check that settles it.
+    /// its data ships in the reference today as <c>GenderedItem&lt;Boolean&gt;</c>, 2/2 writable.
+    /// <see cref="GetterInterfaceFor"/> resolves the getter interface a class OWNS and nothing else — a
+    /// deliberate boundary, not an oversight (#424): implementing someone else's getter interface is the
+    /// structural signal <see cref="ClassifyArm"/> uses to spot a read-only projection. So a type that is not a
+    /// projection yet borrows another type's getter interface resolves null here while being perfectly
+    /// extractable, and this line must not diagnose, only report, and hand the reader the one check that
+    /// settles it.
     ///
     /// Shared by the record path and the union-arm path so both name the same thing the same way.
     /// </summary>
@@ -693,9 +696,9 @@ public static class CorpusGenerator
         var (w, total) = WritableSurfaceCount(t);
         var name = t.FullName ?? t.Name;   // FullName: ~350 distinct nested types share the simple name "ErrorMask"
         if (w == 0)
-            return $"No I{t.Name}Getter resolved by name for {role} {name} — {w} of {total} properties are " +
+            return $"No {GetterProbeName(t)} resolved by name for {role} {name} — {w} of {total} properties are " +
                    $"authorable (public settable, or a mutable collection). {ReachabilityCheck}";
-        return $"UNEXTRACTABLE BY NAME: {role} {name} — no I{t.Name}Getter resolved by name, and {w} of {total} " +
+        return $"UNEXTRACTABLE BY NAME: {role} {name} — no {GetterProbeName(t)} resolved by name, and {w} of {total} " +
                $"properties are authorable (public settable, or a mutable collection). It is excluded from the " +
                $"catalog. {ReachabilityCheck}";
     }
@@ -728,15 +731,53 @@ public static class CorpusGenerator
         return gf?.GetValue(null) is RecordType g ? g.ToString() : null;
     }
 
-    /// <summary>The "I{Name}Getter" interface for a concrete Mutagen class.</summary>
+    /// <summary>
+    /// The name this resolver probes for: the getter interface a class of this name would OWN.
+    /// The arity suffix stays where the runtime puts it — <c>GenderedItem`1</c>'s getter interface is
+    /// <c>IGenderedItemGetter`1</c>, never the impossible <c>IGenderedItem`1Getter</c> a probe built straight
+    /// off <see cref="Type.Name"/> asks for. One source of truth, so the anomaly line reports the name that
+    /// was actually looked up.
+    /// </summary>
+    internal static string GetterProbeName(Type concrete)
+    {
+        var tick = concrete.Name.IndexOf('`');
+        return tick >= 0
+            ? $"I{concrete.Name[..tick]}Getter{concrete.Name[tick..]}"
+            : $"I{concrete.Name}Getter";
+    }
+
+    /// <summary>
+    /// The getter interface a concrete Mutagen class OWNS — the one named after it.
+    ///
+    /// Arity-aware throughout, the same trap <see cref="MutableInterfaceFor"/> already fixes: the probe name
+    /// carries the suffix in the runtime's position (see <see cref="GetterProbeName"/>), and the last fallback
+    /// strips it before testing for the "Getter" ending, which a generic interface's own name never has
+    /// (<c>ISkyrimGroupGetter`1</c>). Missing that answered "no getter interface" for every generic modeled
+    /// type — GenderedItem&lt;T&gt;, SkyrimGroup&lt;T&gt;, FormLinkNullable&lt;T&gt; — in a walk whose whole
+    /// claim is that coverage is Mutagen's coverage by construction.
+    ///
+    /// OWN-NAMED ONLY, deliberately. A class that implements a getter interface named after some OTHER type
+    /// does NOT resolve here, and must not: implementing only the getter side of a type you are not is exactly
+    /// the shape of Mutagen's read-only projections (SkyrimMultiModOverlay implements ISkyrimModDisposableGetter,
+    /// MergedWorldspace implements IWorldspaceGetter, MergedCellBlock implements ICellBlockGetter, and the
+    /// Consolidated*/Merged* group views likewise). <see cref="ClassifyArm"/> reads "no getter interface" as
+    /// precisely that structural signal, so widening this to any implemented getter interface would hand every
+    /// one of those a mutable twin and re-admit it as an authorable arm of a union it can never be composed
+    /// into. The cost of the narrow answer is that a NON-projection implementing someone else's getter
+    /// interface — ArmorAddonWeightSliderContainer, which implements IGenderedItemGetter&lt;bool&gt; — still
+    /// resolves null; its data is catalogued anyway, as GenderedItem&lt;Boolean&gt;, which is why
+    /// <see cref="UnextractableWarning"/> reports rather than diagnoses.
+    /// </summary>
     internal static Type? GetterInterfaceFor(Type concrete)
     {
         if (concrete.IsInterface) return concrete;
-        var direct = concrete.Assembly.GetType($"{concrete.Namespace}.I{concrete.Name}Getter");
-        if (direct != null) return direct;
-        return concrete.GetInterfaces()
-            .FirstOrDefault(i => i.Name == $"I{concrete.Name}Getter")
-            ?? concrete.GetInterfaces().FirstOrDefault(i => i.Name.EndsWith("Getter") && Normalize(i.Name) == Normalize(concrete.Name));
+        var probe = GetterProbeName(concrete);
+        var ifaces = concrete.GetInterfaces();
+        // The IMPLEMENTED interface is asked first, ahead of the assembly lookup: on a CLOSED generic class it
+        // is the closed IFooGetter<Bar>, where Assembly.GetType can only ever hand back the open definition.
+        return ifaces.FirstOrDefault(i => i.Name == probe)
+            ?? concrete.Assembly.GetType($"{concrete.Namespace}.{probe}")
+            ?? ifaces.FirstOrDefault(i => i.Name.Split('`')[0].EndsWith("Getter") && Normalize(i.Name) == Normalize(concrete.Name));
     }
 
     /// <summary>The mutable twin of a getter interface (strip the "Getter" suffix).</summary>
--- a/src/housecarl-generator/CorpusGenerator.cs
+++ b/src/housecarl-generator/CorpusGenerator.cs
@@ -754,7 +754,9 @@ public static class CorpusGenerator
     /// strips it before testing for the "Getter" ending, which a generic interface's own name never has
     /// (<c>ISkyrimGroupGetter`1</c>). Missing that answered "no getter interface" for every generic modeled
     /// type — GenderedItem&lt;T&gt;, SkyrimGroup&lt;T&gt;, FormLinkNullable&lt;T&gt; — in a walk whose whole
-    /// claim is that coverage is Mutagen's coverage by construction.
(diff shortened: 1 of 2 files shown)


Avick3110:
Blind review, no findings. Nothing survived that I could make fail, so nothing is posted inline.

What I checked, and how.

**The byte-identical claim, measured rather than read.** I emitted the corpus twice from two already-built generators — the pre-fix one in the primary checkout at `main` (62b2356, confirmed pre-fix: its `arm-classification-guard` has no E arms) and the one at this PR's head (7210f85) — both into scratch dirs with an explicit `refDir` so nothing tracked was touched. `corpus.json`, `corpus.summary.md` and all six `mutagen-reference` shards are identical, and the printed report is identical line for line, coverage-anomaly section included (same two lines, `MergedWorldspace` and `MergedWorldspaceCell` — both non-generic, so `GetterProbeName` renders the same string `UnextractableWarning` used to build by hand). `emit-match-guard` is green against the committed shards from the PR build, and `ci-all` is 127/127 on the branch. So the claim holds as stated, not just as reasoning.

**Why it holds, from the code.** The three behavioural deltas in the new resolver are (a) arity-correct probe naming, (b) implemented-interface-before-assembly-lookup, (c) the arity strip in the last fallback. Every one of them only changes an answer for a generic type. For a non-generic class the reorder is a no-op: the old first step took `{Namespace}.I{Name}Getter` from the class's own assembly and the old second step took an implemented interface of that exact simple name, so swapping their order can only differ if a class implements an `I{Name}Getter` that is a different type from the one sitting in its own namespace — which does not occur in the three assemblies the walk visits, as the identical emit shows. The generic answers stay off the emit path for the reason the issue gives: `FindUnionArms` filters with `IsAssignableFrom`, false for the open definitions `Assembly.GetTypes()` yields, and a closed generic in a field position arrives as its getter interface, which `GetterInterfaceFor` returns on line one. The one closed-generic case that could have moved output — `AssetLink<T>` reaching `ResolveGetterInterface` as a concrete class — lands on `ImplementsByName(getterIfc, "IAssetLinkGetter")` either way, because `IAssetLinkGetter<T>` carries the non-generic interface, and that branch hard-codes its `Type`, so even a changed resolution emits the same field.

**The NO on the issue's second half.** Verified the part it turns on rather than the 32-type count: both exhibits really would flip. `SkyrimMultiModOverlay` implements `ISkyrimModGetter` (BuildCorpus's own comment names it as the second implementer) and `ISkyrimMod` exists; `MergedCellBlock` implements `ICellBlockGetter` and `ICellBlock` exists — both appear as `MutableInterface` values in the emitted corpus. So widening to any implemented getter interface gives each a mutable twin, `ClassifyArm` calls both `Authorable`, and `SkyrimMod` reclassifies from a plain struct to a two-arm union. That is a corpus change and a straight regression of #397's split, so own-named-only is not merely defensible here, it is what keeps the emit still. The boundary is also genuinely pinned from two directions as the header says: widening turns B4 red, since `ArmorAddonWeightSliderContainer` would resolve `IGenderedItemGetter<bool>` whose twin `IGenderedItem<bool>` exists.

**Tests that would pass with the fix reverted.** None of the three. Under the old resolver `GetterInterfaceFor(GenderedItem<>)` is null on all three steps — the assembly probe asks for the impossible `IGenderedItem`1Getter`, and the last fallback's `EndsWith("Getter")` is false for `IGenderedItemGetter`1` — so E1 and E2 are red. E3 is the one that pins the reorder rather than the arity fix: `Assembly.GetType` on a backtick name hands back the open definition, so asking the assembly first would fail the `== typeof(IGenderedItemGetter<bool>)` equality. Each arm fails for the specific thing it names.

**The removed `?? t`.** The only behaviour difference is a class with no own getter interface whose own name ends in "Getter" and which has a same-named mutable twin — `MutableInterfaceFor(gi)` and `MutableInterfaceFor(gi ?? t)` agree everywhere else. `FormLinkNullableGetter`1` and `FormLinkOrIndexGetter`1` now resolve `IFormLinkNullableGetter`1` / `IFormLinkOrIndexGetter`1` through the arity-stripped fallback and stay `Authorable` through the `gi` path, so the rescue is genuinely spent. Confirmed no live consequence: identical emit, identical anomaly channel, 127/127.

Scope is the issue and nothing else — two generator files, no server surface, and the issue's conditional `WritableSurfaceCount` item correctly left alone with its condition unmet.

src/housecarl-generator/CorpusGenerator.cs
```diff
@@ -728,15 +731,53 @@ internal static string UnextractableWarning(string role, Type t)
...
-        return concrete.GetInterfaces()
-            .FirstOrDefault(i => i.Name == $"I{concrete.Name}Getter")
-            ?? concrete.GetInterfaces().FirstOrDefault(i => i.Name.EndsWith("Getter") && Normalize(i.Name) == Normalize(concrete.Name));
+        var probe = GetterProbeName(concrete);
+        var ifaces = concrete.GetInterfaces();
+        // The IMPLEMENTED interface is asked first, ahead of the assembly lookup: on a CLOSED generic class it
+        // is the closed IFooGetter<Bar>, where Assembly.GetType can only ever hand back the open definition.
+        return ifaces.FirstOrDefault(i => i.Name == probe)
+            ?? concrete.Assembly.GetType($"{concrete.Namespace}.{probe}")
+            ?? ifaces.FirstOrDefault(i => i.Name.Split('`')[0].EndsWith("Getter") && Normalize(i.Name) == Normalize(concrete.Name));
```
Avick3110: The normalized fallback has no arity preference: it returns whichever matching interface `GetInterfaces()` lists first.

Measured against the shipped Mutagen.Bethesda.Core, two live types take this path (their probe name is the impossible ``IFormLinkGetterGetter`1``), and each matches **both** candidates:

```
FormLinkGetter`1:  matches [IFormLinkGetter`1, IFormLinkGetter]
AssetLinkGetter`1: matches [IAssetLinkGetter`1, IAssetLinkGetter]
```

`Normalize` collapses both to the same key, and after the new `Split('`')[0]` both pass the "Getter" ending test. Today the runtime happens to list the generic one first, which is why the table in the description reads ``IFormLinkGetter`1`` — but `Type.GetInterfaces()` ordering is not specified, and a Mutagen recompile that reorders the interface list flips these to the non-generic `IFormLinkGetter`. `MutableInterfaceFor` would then hand back the non-generic `IFormLink`, i.e. a generic type's schema read off a zero-argument interface — the exact regression E3 pins, on the one path E3 does not cover.

Preferring an arity match before the untyped `FirstOrDefault` removes the order dependence and makes this line agree with the property E1–E3 assert.

src/housecarl-generator/CorpusGenerator.cs
```diff
@@ -728,15 +731,53 @@ internal static string UnextractableWarning(string role, Type t)
...
-        var direct = concrete.Assembly.GetType($"{concrete.Namespace}.I{concrete.Name}Getter");
-        if (direct != null) return direct;
-        return concrete.GetInterfaces()
-            .FirstOrDefault(i => i.Name == $"I{concrete.Name}Getter")
-            ?? concrete.GetInterfaces().FirstOrDefault(i => i.Name.EndsWith("Getter") && Normalize(i.Name) == Normalize(concrete.Name));
+        var probe = GetterProbeName(concrete);
+        var ifaces = concrete.GetInterfaces();
+        // The IMPLEMENTED interface is asked first, ahead of the assembly lookup: on a CLOSED generic class it
+        // is the closed IFooGetter<Bar>, where Assembly.GetType can only ever hand back the open definition.
+        return ifaces.FirstOrDefault(i => i.Name == probe)
```
Avick3110: Asking the implemented interface before the assembly lookup also changes what an **open generic definition** resolves to: the open *constructed* interface `ISkyrimGroupGetter<T>` rather than the generic type *definition* the assembly lookup returns. A constructed open type has `FullName == null` and `AssemblyQualifiedName == null`.

Measured over the walk's assemblies — all 11 open generic classes (``SkyrimGroup`1``, ``SkyrimListGroup`1``, ``FormLink`1``, ``GenderedItem`1``, ``LoadOrder`1``, …) now come back with a null `FullName`, where line 779 would have returned the definition named ``Mutagen.Bethesda.Skyrim.ISkyrimGroupGetter`1``.

Every consumer keyed on the full name degrades silently rather than failing: `ExtractType` would write `GetterInterface` as the bare ``ISkyrimGroupGetter`1`` with no namespace (and the same for the assembly-qualified field), and `IsList`'s hard-coded ``Mutagen.Bethesda.Skyrim.ISkyrimGroupGetter`1`` can never match a type whose `FullName` is null — which is exactly the agreement E2's comment says the resolver has to keep with that list.

Latent today only because `FindUnionArms` filters open definitions out with `IsAssignableFrom`, so it holds precisely as long as that filter does. Normalizing to `GetGenericTypeDefinition()` when `concrete.IsGenericTypeDefinition` keeps the closed-generic win of the reorder without the null-name shape.

src/housecarl-generator/CorpusGenerator.cs
Avick3110: Real, folded in 250c2ffc.

Measured it over the two assemblies the walk visits (Skyrim + Core). Seven classes reach the normalized fallback; exactly the two you name match more than one candidate, and both list the arity match first today:

```
FormLinkGetter`1   -> [IFormLinkGetter`1, IFormLinkGetter]
AssetLinkGetter`1  -> [IAssetLinkGetter`1, IAssetLinkGetter]
InterfaceMapGetter -> [IInterfaceMapGetter]
LoadOrderGetter    -> [ILoadOrderGetter]
FormVersionGetter  -> [IFormVersionGetter]
FormLinkOrIndexGetter`1   -> [IFormLinkOrIndexGetter`1]
FormLinkNullableGetter`1  -> [IFormLinkNullableGetter`1]
```

So the answer is right by the runtime's ordering rather than by construction, and the ordering is unspecified.

The selection moved into `OwnNamedGetterAmong(concrete, ifaces)`, which prefers a candidate whose arity equals the class's and falls back to the first match as before. The helper takes the interface list as a parameter for one reason: it is the only way to make the failure reproducible. E4 asks it for both classes over `GetInterfaces()` and over `GetInterfaces().Reverse()` and requires the same answer, generic, either way. With the selection back to a plain `FirstOrDefault` the arm is red and names both:

```
FAIL  E4. the normalized fallback prefers the arity match, whatever order GetInterfaces() lists
      -> order-dependent: FormLinkGetter`1, AssetLinkGetter`1
```

Corpus: unchanged. A full emit on the folded branch is byte-identical to one from the PR head — `corpus.json`, `corpus.summary.md`, all six shards, and the printed report line for line, the same two coverage-anomaly lines included. No record type's schema moved, so there is nothing to list type by type; today's ordering already gave the arity answer, and this only removes the dependence on it.

Ran on the fold: build 0 errors, `dotnet test --filter "tier!=bridge"` 1767 passed, `ci-all` 127/127 ALL PASS.

src/housecarl-generator/CorpusGenerator.cs
Avick3110: Real, folded in f80e6d68.

Measured it: under the head resolver exactly 15 classes come back with a null `FullName`, and every one is an open generic definition — ``SkyrimGroup`1``, ``SkyrimListGroup`1``, ``FormLink`1``, ``FormLinkGetter`1``, ``FormLinkNullable`1``, ``FormLinkNullableGetter`1``, ``FormLinkOrIndex`1``, ``FormLinkOrIndexGetter`1``, ``EDIDLink`1``, ``AssetLink`1``, ``AssetLinkGetter`1``, ``GenderedItem`1``, ``MajorRecordIdentifier`1``, ``LoadOrder`1``, ``ModListing`1``. For 11 of them the assembly lookup would have returned the named definition (``Mutagen.Bethesda.Skyrim.ISkyrimGroupGetter`1`` and so on); the other 4 are the ``*Getter`1`` classes, which have no own-named assembly type at all and reach the answer through the last fallback — the definition is still reachable from the constructed type, so the same normalization covers them.

The fix is the one you name: when `concrete.IsGenericTypeDefinition` and the answer is a constructed generic, return `GetGenericTypeDefinition()`. A closed generic is untouched, so E3's closed-interface win stands.

It is pinned by tightening E2 rather than by a new arm, because E2 was the arm that already claimed this. Its comment says the resolver has to agree with `IsList`'s hard-coded full name, but it asserted only `GetGenericTypeDefinition() == typeof(ISkyrimGroupGetter<>)`, which a nameless constructed answer satisfies. It now asserts the identity and the full name `IsList` actually matches:

```
groupOpen == typeof(ISkyrimGroupGetter<>)
&& groupOpen.FullName == "Mutagen.Bethesda.Skyrim.ISkyrimGroupGetter`1"
```

With the normalization removed that arm is the one red arm in the guard, so it fails for its own reason and nothing else moved.

Corpus, record type by record type: no record type's schema moved. A full emit on the folded branch is byte-identical to one from the PR head — `corpus.json`, `corpus.summary.md`, all six committed shards, and the printed report line for line, the same two coverage-anomaly lines (`MergedWorldspace`, `MergedWorldspaceCell`) included. Those 15 stay off the emit path for the reason the PR gives, so the fold is as latent as the change it corrects; `emit-match-guard` is green against the committed shards untouched.

Ran on the fold: build 0 errors, `dotnet test --filter "tier!=bridge"` 1767 passed, `ci-all` 127/127 ALL PASS.

PR #623, opened 2026-09-07
Fold the standing context into the server instructions and nine parameters

### LINKED ISSUE #401: [Docs] tool-output-awareness needs a rewrite or scrapping

Avick3110:
Wave 0 of the skill rewrite: the standing-context half of the fold ledger (`dev/skill-validation/eval-2026-09/FOLD-IN.md`, corrected 2026-09-08). Everything the ledger routes to `ServerInstructions` or to a parameter description lands here; nothing under `.claude/skills/` or `plugin/` is touched. The instructions string in `src/housecarl-mcp/Program.cs` keeps its existing 1,892 characters untouched and appends F1 (which framework owns a job: SPID / KID / CID / SkyPatcher), F2 (nothing houseCARL writes wins until the mod is enabled in MO2) and F3 (never copy generated output, with the fixed filenames), each preceded by one joining space. Nine parameter descriptions take the per-parameter semantics that were living in skill bodies: F4 on `housecarl_create.records`, F5 on `housecarl_check.findings`, F6/F7/F47 on `housecarl_apply.ops`, F8 on `housecarl_records.fields_source`, F9 on `housecarl_records.project`, F10/F50 on `housecarl_records.where`, F48 on `housecarl_forward.source`, and F49's `" + sorted"` deletion on `housecarl_apply.readback` and `housecarl_create.readback`. Every added sentence is the ledger's text, corrected where review showed the ledger's draft was not true of the code (below), with the two tool names spelled through `ToolNames` the way the rest of the surface spells them.

**ServerInstructions measures 3,133 characters on the wire**, not the ledger's 2,875. Two review findings account for the 258, and each buys back a case where the drafted sentence sent a caller the wrong way:

- **+121 on F2** — "a new mod loads LAST, so enabling is the step, not sorting" is true of the fresh-folder lane (`ResolveOutputPath` makes a new `<ModsDir>\houseCARL - <name>`, houseCARL never writes `modlist.txt`, and MO2 registers an unseen folder at the highest priority — top of `modlist.txt`, per this repo's own parse at `Mo2LoadOrder.cs:10`). It is not true of `place into=` / `write_seq into=`, which write into an existing patch folder whose priority is already fixed. The sentence now says "a NEW mod folder loads LAST, so enabling is the step, not sorting, while a write into an EXISTING mod keeps that mod's priority and may still need sorting above the current winner".
- **+137 on F3** — the folder signal is the skill's tier 3 and never shipped without its false-positive check: `DynDOLOD 3 Resources`, a `… DynDOLOD TexGen Fixes (…)`, a downloaded `… xEdit-Synthesis Patch` are inputs the generator consumes. Folded flat, the rule would have refused a patch against an ordinary authored mod, and a refusal is the one failure a description cannot self-correct. Appended: "A mod that only NAMES a generator ('… Resources', '… Fixes', a downloaded patch) is an INPUT it consumes, not output: patch it normally."

The instructions are measured, not asserted from the source literal: `ServerFixture` captures the `instructions` string off the `initialize` result the same way it captures `tools/list`, and `ServerInstructionsTests` writes the measured length to the test output, records 3,133 and its two reasons in its header, and holds each of the three folded facts to exactly one statement — the point of folding them was that they are said once. Measured on the wire: `create.records` 2,132, `create.readback` 509, `check.findings` 5,110, `apply.ops` 1,076, `apply.readback` 539, `records.fields_source` 433, `records.project` 664, `records.where` 2,694, `forward.source` 1,612. Two are over the ledger's prediction, from the other three findings:

- **`create.records` 2,132** (ledger 1,988). `SubtypeName` (SNAM) is a `DialogTopic` field, not an INFO one (`DialogueSubtype.NormalizeMarker` takes `IDialogTopic`; the generated schema has it on DIAL and not on INFO), so the clause moved next to the DIAL's `Priority`. And the list said "each field it filled" while naming 8 of the 11 `DialogueCkParity` fills — the missing DLBR `Flags` (DNAM) is the one whose absence publishes a branch to the player's menu. The list is now complete rather than hedged. Past 2,048 deliberately: that cut is Claude Code's *tool description* cut, and parameter descriptions arrive whole (`housecarl_check.findings` has shipped at 5,110 for as long as it has existed).
- **`records.project` 664** (ledger 545). The drafted "limit= caps only the RENDERED rows" is a general property of `limit=` stated inside a sentence scoped to the one form where it does not hold: on `aggregate` the scan is uncapped, `limit` is never consulted, and `offset=` is refused outright. The sentence now says that, and names the lane the render bound holds on (below).

No tool description changed — the three largest are still `housecarl_nif_inspect` 2,037, `housecarl_records` 2,036 and `housecarl_apply` 2,032, and `PublishedDescriptionBoundTests` passes on all 31.

**F22, checked in code and written.** `LoadOrderService.WriteSeq` compares the built bytes against the destination and returns `Unchanged` (refreshing the timestamp if the file was older than the plugin) *after* the lane has resolved the folder, so the skip is lane-independent and the `patch=`/`into=` lanes do behave the way `output_dir=` states. One thing the ledger did not anticipate: `patch=` resolves through `UniqueStem`, so its folder is always fresh and a `.seq` can never already be there — the state is real but unreachable on that lane. So the sentence is written onto `into=` only, where an existing patch folder can hold a prior `.seq`, and `patch=` is left alone rather than told about a case it cannot reach.

**F11 needs nothing.** `ReadEngine.FlagSlotDisplay` (`src/housecarl-core/ReadEngine.cs:1049`) already decodes a `BodyTemplate.FirstPersonFlags` value into slot numbers with `slots.Add(30 + i)` over the set bits, gated to `BipedObjectFlag` by name — "slot N is bit N−30" is shipped in the rendering, as #620 also found. Nothing was added.

**F51 is not here.** #620 already removed the live pointers into the deleted skills, and it carries `Closes #401`, so this PR references #401 without closing it: F47 and F48 are the re-expression of #401's sentence that #620's body says lands in the standing-context PR.

One review finding was refuted rather than folded, on the PR thread: the fresh-folder half of F2 stands, because a new mod folder outranks the previous winner the moment it is ticked. A separate note there, not changed here: `PlaceTools.EnableAndSort` overstates in the other direction, telling every caller with a contended row to sort above the winner even on the fresh-folder lane where it already wins — it branches on contention and never on the lane. Filed as #632.

**Second review round, four more findings, all four folded.** `ServerInstructions` is untouched by them and stays at 3,133; the four are parameter text, and two of them turned up code defects that are filed rather than fixed here, because a description-only PR should not change what a call does.

- **`apply.ops` 1,076** (+12) — `from_source=` was stated as a general per-op pole; only `op='CopyFrom'` honours it, and nothing refuses it elsewhere. `ApplyTools.cs:148` gates `from=` on the verb; `from_source` is a declared member, so it binds and maps to `BulkOp.FromPlugin` while every consumer (`PrepareCopyFromSources`, `IsOffOrderCopySource`, `ResolveCopyPole`) requires `CopyFrom` — a silent drop. The clause now reads "from_source= (op='CopyFrom' only)". The missing refusal is #630.
- **`apply.readback` 539 and `create.readback` 509** (+112 each) — F49's flat `" + sorted"` deletion left "wins nothing until enabled" true of the fresh-folder lane only, while both tools have an `into=` lane writing into a mod that is usually already enabled. Both now carry the instructions' own clause verbatim, so the three surfaces say one thing.
- **`records.project` 664** (+66) — "max_chars alone bounds the render" holds on the scan lane (`RenderCrossQueryGroups` budgets group rows and reports truncation) but not on the `formids=` lane: `RenderListAggregate` (`RecordsTools.cs:2316`) takes no cap and settles nothing. The sentence now names both lanes. Passing the cap in is #631, and when it lands the sentence collapses back to one clause.
- **`create.records` 2,132** (+23) — VTCK is filled only on a Reference alias (`DialogueCkParity.cs:428`), not on every alias; `Flags` (FNAM) is on every one. Two comments in `DialogueCkParity` itself carried the same error and are corrected with it.

Build clean; 1,793 tests pass; `ci-all` 128/128, `description-vocab-guard` green.

References #401


src/housecarl-mcp/Program.cs
```diff
@@ -151,7 +151,20 @@ static void AddMcp(IServiceCollection services, bool stdio)
...
+            "MD5. Prefer over a browser or web search; each tool's own description carries the specifics. " +
+            // The three cross-skill facts the skill rewrite folds here: which runtime framework owns a job, that a
+            // write wins nothing until enabled, and the never-copy rule for generated output.
+            "RUNTIME DISTRIBUTION LAYERS — which framework owns a job is decided by what RECEIVES the change: a " +
+            "spell, perk, item, keyword, outfit or faction onto NPCs is SPID; a keyword onto item records is KID; " +
+            "items into containers is CID; a record's OWN FIELDS is SkyPatcher, whose replayed layer " +
+            ToolNames.SkypatcherLayer + " reads. " +
+            "NOTHING houseCARL WRITES WINS UNTIL IT IS ENABLED. A patch plugin, a placed asset and a written .seq " +
+            "do nothing until the user enables that mod in MO2 — a new mod loads LAST, so enabling is the step, " +
+            "not sorting — and every read-back describes the WRITTEN FILE, not the load order. " +
```
Avick3110: **"enabling is the step, not sorting" is false for the placed-asset half of this sentence, and it contradicts what `housecarl_place` itself prints.**

`PlaceTools.EnableAndSort` (src/housecarl-mcp/PlaceTools.cs:292-313) renders, verbatim, to every caller of `housecarl_place`:

> Enable the mod '<folder>' in MO2 **and SORT it (left pane) ABOVE the current winner(s) listed above. Only then does the placed copy win.**

and only takes the "once enabled the placed copy wins" arm when *nothing else provided the path*. `LoadOrderService.PlaceOne`'s own summary says the same ("the placed file does NOT win until the mod is enabled + sorted", LoadOrderService.cs:1573), as does `PlaceResult.CurrentWinner` (LoadOrderService.cs:9766).

Concrete scenario: a session places a corrected facegen mesh over the one an NPC overhaul already provides. The standing instructions — which every session gets, whether or not a skill loads, and which outrank a tool response the model has already read past — say enabling is the step and sorting is not. The model tells the user to tick the mod in MO2; the new mod is not above the overhaul in the left pane, the overhaul keeps winning the VFS, and the dark face is still there. The tool's own correct instruction is now competing with a blanket claim that says it is unnecessary.

Secondarily, the `.seq` half is wrong on two of its three lanes: `into=` writes into an existing houseCARL patch mod and `output_dir=` typically writes into the plugin's own mod after an in-place edit — both already enabled by construction.

Suggest scoping the claim to the plugin lane, which is the one where "a new mod loads LAST" holds, and leaving the file lanes to the place/seq responses that already state them correctly.

src/housecarl-mcp/Program.cs
```diff
@@ -151,7 +151,20 @@ static void AddMcp(IServiceCollection services, bool stdio)
...
+            "NEVER COPY GENERATED OUTPUT: a tool re-derives it on its next run, so it is never copied into an " +
+            "authored patch and never the base for authored work — 'Requiem for the Indifferent.esp'; " +
+            "'PGPatcher.esp' / 'PG_1.esp'; 'DynDOLOD.esm' / 'DynDOLOD.esp' / 'Occlusion.esp'; a mod folder holding " +
+            "'NPC_Token.json' or 'ParallaxGen_Diff.json'; a Synthesis, TexGen or xLODGen output folder.";
```
Avick3110: **The fold drops the one qualifier the folder-based signals cannot ship without.**

The skill this is folded from grades its signals in three tiers (.claude/skills/tool-output-awareness/SKILL.md:44-83): fixed plugin filenames and marker files are marked "bulletproof", but tier 3 — the output *folder* — is explicitly labelled "needs the false-positive check", and the section immediately below lists the false positives by name: `DynDOLOD 3 Resources`, `DynDOLOD TexGen Fixes (…)`, a `… DynDOLOD Add-On`, a downloaded `… xEdit-Synthesis Patch`. Those are **inputs the generator consumes**, and the skill's ruling is "patch it like any normal mod."

This line folds tier 3 in flat — "a Synthesis, TexGen or xLODGen output folder" — with no decisive test beside it. The two marker-file signals on the line above are fine; this clause is the one that was never self-sufficient.

Concrete scenario: a user asks for a patch against `DynDOLOD TexGen Fixes (SMIM)` or a `… xEdit-Synthesis Patch` in a session where the skill has not loaded. The model reads NEVER COPY GENERATED OUTPUT, matches the folder name, and refuses to forward from or patch an ordinary authored mod — the exact miss the skill's false-positive section exists to prevent. A refusal-shaped error is the one failure mode a description cannot self-correct, because the call never runs.

Either carry the decisive test with the clause (does the mod hold the generator's own plugin or marker, or is it the directory the tool wrote into?), or fold only the two bulletproof tiers here and leave the folder signal to the skill.

src/housecarl-mcp/CreateTools.cs
```diff
@@ -33,7 +33,7 @@ public static class CreateTools
...
...rmID as the value and the same lane (into='<this patch>', or in_place='<that plugin>' in place). The pre-flight refusals include an unknown or ambiguous record_type, a missing editorid, an illegal field op, a nested child with no resolvable parent, an ambiguous collection, and a type that cannot be created (the bare abstract base 'Global'/'GameSetting' needs a concrete subtype; the refusal names them). What else is reported, never silently: creating dialogue lines reports VOICE coverage (a response with no .fuz plays SILENT in game — the audio is yours to provide) and RESULT-SCRIPT binding (a bound script that is unwired or uncompiled runs nothing); creating cells reports the world content houseCARL does NOT author (lighting, terrain, water, navmesh — Creation Kit work). Authoring dialogue also AUTO-FILLS the Creation Kit's own bookkeeping and reports each field it filled: the INFO's SubtypeName (SNAM) marker derived from Subtype, its FavorLevel (CNAM) and Flags (ENAM), a DLVW's DNAM and ENAM, a DLBR's Category (TNAM), the DIAL's Priority (PNAM), and the QUST's NextAliasID (ANAM) plus each objective's Flags (FNAM). Every auto-fill is NON-OVERRIDE: a value you passed always wins.")]
```
Avick3110: **SNAM is attributed to the wrong record type here.** The new sentence reads:

> the **INFO's** SubtypeName (SNAM) marker derived from Subtype, its FavorLevel (CNAM) and Flags (ENAM)

`SubtypeName` is a **DialogTopic (DIAL)** field, not a DialogResponses (INFO) one. `DialogueSubtype.NormalizeMarker` takes `IDialogTopic topic` and writes `topic.SubtypeName` (src/housecarl-core/DialogueSubtype.cs:183-190), and the CK-parity header calls it "the DialogTopic SNAM marker in DialogueSubtype" (src/housecarl-core/DialogueCkParity.cs:16). Only CNAM/ENAM in that clause are INFO fields. The sentence also names DIAL separately, for Priority (PNAM) — so a reader has no way to recover the right owner from context.

Concrete scenario: a caller authoring dialogue reads this, expects the marker on the line record, and issues `{record_type: "DialogResponses", …, ops: [{field_path: "SubtypeName", …}]}` or looks for SNAM in the INFO's read-back. `SubtypeName` is not a field on DialogResponses, so the pre-flight refuses by name — and the description that caused it is the one surface the caller is supposed to trust. The whole point of moving this fact out of the skill body is that the parameter is now the only place it is stated.

Moving the SNAM clause next to the DIAL's Priority clause fixes it.

Avick3110: **"reports each field it filled" then names 8 of the 11 the code fills — and the three missing ones include the sharpest.**

`DialogueCkParity` (src/housecarl-core/DialogueCkParity.cs:23-49) enumerates its defaults in three tiers. This sentence covers INFO CNAM/ENAM, DLVW DNAM/ENAM, DLBR TNAM, DIAL PNAM, QUST ANAM and QUST objective FNAM. Absent:

- **DLBR Flags (DNAM)** — the whole IN-GAME-BEHAVIOUR tier. The file's own note: an absent DNAM is read by the engine as TopLevel, so a branch the author never marked top-level is published to the player's dialogue menu anyway, and "the output is byte-valid and passes validate_dialogue, so nothing catches it before the game does."
- **QUST QuestAlias Flags (FNAM)** — materialised per alias, alongside the objective FNAM that *is* named.
- **QUST QuestAlias VoiceTypes (VTCK)**.

Concrete scenario: a caller creates a DLBR and, working from this list, believes the auto-fills it will get are Category (TNAM) only. It also gets Flags (DNAM) = 0, which is the fill that decides whether the branch shows up in the player's menu — the one auto-fill whose value the author most needs to know landed, and the one the list does not mention. A caller reading a sentence that says *each* is stated has no reason to go looking for the other three.

Either name all eleven, or drop "each" and say the list is the ones worth knowing about.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -126,9 +126,9 @@ public static string Records(
...
             string? fields_source = null,
-        [Description("PROJECT: the shape of the answer — a SINGLE form plus that form's own sub-parameters: identity | summary (default) | fields | rows | everything | aggregate | delta | tree | chain | info_order. Sub-parameters live INSIDE the form that uses them (depth belongs to fields/rows/everything, group_by to aggregate, fields to fields/rows/delta/tree), so there is no flat spelling for an illegal pairing. Omit for summary rows.")]
+        [Description("PROJECT: the shape of the answer — a SINGLE form plus that form's own sub-parameters: identity | summary (default) | fields | rows | everything | aggregate | delta | tree | chain | info_order. Sub-parameters live INSIDE the form that uses them (depth belongs to fields/rows/everything, group_by to aggregate, fields to fields/rows/delta/tree), so there is no flat spelling for an illegal pairing. Omit for summary rows. aggregate counts ALL matches regardless of limit=; limit= caps only the RENDERED rows, so a count is safe to size a job with.")]
```
Avick3110: **"limit= caps only the RENDERED rows" is true of the row forms, but false of the form this sentence is about.**

On `aggregate` (`group_by=`), `limit=` caps nothing at all. The scan is uncapped (LoadOrderService.cs:4652, "aggregate over all matches, no keys or prefill, no limit cap"), and the *render* is bounded by `max_chars`, not `limit=` — the group loop breaks on `ms.Length >= cap` and reports `truncated`, with `limit` never consulted (JsonWire.cs:1124-1143). `offset=` is refused outright on this form for the same reason (LoadOrderService.cs:4309).

So the first clause is right and worth having; the second states a general property of `limit=` inside a sentence scoped to the one form it does not hold for.

Concrete scenario: a caller runs an aggregate over a whole-order scan expecting a wide key space, sets `limit=20` on the strength of this sentence to keep the render small, and gets every group instead — trimmed, if at all, by the `max_chars` ceiling with a `truncated` flag rather than a clean 20-row window.

"aggregate counts ALL matches and `limit=` does not window its groups — `max_chars` bounds the render" says the true thing without losing the point.

src/housecarl-mcp/Program.cs
Avick3110: Half folded, half refuted — 21e90a3a.

**Refuted: the new-folder lane.** `place`'s default lane creates a *fresh* mod folder — `ResolveOutputPath` (LoadOrderService.cs:7815-7832) makes `<ModsDir>\houseCARL - <name>` and stamps its own `meta.ini` — and houseCARL never writes `modlist.txt` at all (it only ever reads it). MO2 registers a folder it has not seen at the highest priority, which is last in the left pane and the first line of `modlist.txt` — this repo's own parse says top is highest (Mo2LoadOrder.cs:10 and :122). So a placed copy in a new folder outranks the previous winner the moment the mod is ticked; enabling is the step and sorting is not, which is why Aaron ruled the sentence this way on 2026-09-07 (item 4) and why F2 records it as a correction of all three skill sites rather than a paraphrase of them. The scenario in the finding — the corrected facegen mesh over an overhaul — is the fresh-folder lane, so the standing sentence sends the user to the right action there.

**Folded: the existing-mod lane.** `place into=` and `write_seq into=` write into an *existing* houseCARL patch folder whose priority is already fixed, and there the placed copy really can sit below the current provider. That is the one case the sentence was missing, so it now carries a clause for it and nothing more:

> — a NEW mod folder loads LAST, so enabling is the step, not sorting, while a write into an EXISTING mod keeps that mod's priority and may still need sorting above the current winner —

+121 characters. Measured on the wire, ServerInstructions is now **3,133** (2,875 + 121 here + 137 for the never-copy input exception on the sibling thread).

**The .seq half.** `output_dir=` lands the .seq in the user's own mod — typically the plugin's own, after an in-place edit — which is already enabled, and `housecarl_write_seq` says exactly that rather than "enable the new mod" (SeqTools.cs:148). `into=` is an existing houseCARL patch. "Wins nothing until it is enabled" is a precondition and is true on all three lanes; what was actually missing was the priority half, which the clause above now states.

**Separate note, no change in this PR.** `PlaceTools.EnableAndSort` (PlaceTools.cs:298-313) overstates in the other direction: whenever *any* placed row had a current winner it prints "SORT it (left pane) ABOVE the current winner(s)", including on the fresh-folder lane where the new mod already outranks that winner. Its no-winner arm is the correct one. That is a sentence in the place response, not part of this fold — worth its own issue.

src/housecarl-mcp/Program.cs
Avick3110: Folded — f278589f. Real: the skill grades the folder signal tier 3 and pairs it with a named false-positive list (`tool-output-awareness/SKILL.md:74-79`), and the fold shipped the signal without the test. A refusal on an input mod is the failure mode a description cannot correct, because the call never runs.

The shortest clause that carries the test, appended after the filename list:

> A mod that only NAMES a generator ('… Resources', '… Fixes', a downloaded patch) is an INPUT it consumes, not output: patch it normally.

+137 characters. It reads against the list it follows: the list *is* the decisive test (a fixed plugin name, a marker file, or the directory the tool wrote into), so "only names" is the complement of it in one word rather than a restatement. `DynDOLOD 3 Resources`, `DynDOLOD TexGen Fixes (SMIM)`, a `… DynDOLOD Add-On` and a downloaded `… xEdit-Synthesis Patch` all land on the input side of it.

Measured on the wire, ServerInstructions is now **3,133** (2,875 + 137 here + 121 for the existing-mod clause on the sibling thread), stated in `ServerInstructionsTests`' header with the two reasons.

src/housecarl-mcp/CreateTools.cs
Avick3110: Folded — 7e8b71e2. Verified three ways before folding: `DialogueSubtype.NormalizeMarker` takes `IDialogTopic` and writes `topic.SubtypeName` (DialogueSubtype.cs:184-190); `DialogueCkParity`'s header calls it "the DialogTopic SNAM marker in DialogueSubtype" (DialogueCkParity.cs:16); and in the generated schema `SubtypeName` is a field of `DialogTopic` and does not appear on `DialogResponses` at all (`.claude/skills/mutagen-reference/references/records.jsonl`, the DIAL and INFO rows). A caller following the old wording would have been refused by name at the pre-flight.

The clause now sits beside the DIAL's own: "the DIAL's Priority (PNAM) and its SubtypeName (SNAM) marker derived from Subtype", with CNAM/ENAM left as the INFO's.

Avick3110: Folded — 7e8b71e2, and the list is complete rather than hedged with "including". Checked against `DialogueCkParity.cs:23-49`: the three absent fills were DLBR Flags (DNAM), QUST QuestAlias Flags (FNAM) and QUST QuestAlias VoiceTypes (VTCK). The sentence now reads:

> ... the INFO's FavorLevel (CNAM) and Flags (ENAM), a DLVW's DNAM and ENAM, a DLBR's Category (TNAM) and Flags (DNAM — an absent one publishes the branch as TopLevel), the DIAL's Priority (PNAM) and its SubtypeName (SNAM) marker derived from Subtype, and the QUST's NextAliasID (ANAM) plus each objective's Flags (FNAM) and each alias's Flags (FNAM) and VoiceTypes (VTCK).

DNAM carries its consequence in six words because that is the whole reason it is the sharpest of the three — a fill whose value decides whether the branch reaches the player's menu is worth more than its name.

On the budget: `housecarl_create.records` measures **2,109** on the wire, over 2,048, and that is fine — the 2,048 cut is Claude Code's *tool description* cut, and parameter descriptions are delivered whole. `PublishedDescriptionBoundTests`' own header says so, `housecarl_check.findings` already ships at 5,110 and `housecarl_records.where` at 2,694, and FOLD-IN §1 states the cut against tool descriptions only. The ledger predicted +416 for F4; the true list costs +537.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded — a9d64c49. Read the aggregate path end to end and the finding holds: the scan increments the group counts with no `limit` in the condition (LoadOrderService.cs:4652-4657), the json render breaks on `ms.Length >= cap` and never consults `limit` (JsonWire.cs:1124-1143), the text render bounds the group rows on the `max_chars` budget the same way (ReadTools.cs:434-466), and `offset=` is refused outright on this form (LoadOrderService.cs:4309-4311).

The sentence now says what is true of the form it is scoped to:

> aggregate counts ALL matches and limit= does not window its groups (offset= is refused there outright) — max_chars alone bounds the render, so a count is safe to size a job with.

The `offset=` half is there because the refusal is the caller's evidence for the rest: a form that refuses paging is a form nothing windows. +53 characters over the ledger's F9 draft; `housecarl_records.project` measures **598** on the wire.

src/housecarl-mcp/ApplyTools.cs
```diff
@@ -47,7 +47,7 @@ public static class ApplyTools
...
+        [Description("The edits, all into one artifact: [{formid, field_path, op?, value?, values?, key?, entries?, compose?, composes?, from?, from_source?}, …] — or \"@<absolute path>\" to read that SAME array from a JSON manifest file. One op is a set of one. An op member the shape does not declare is refused BY NAME at its element, never silently dropped. The manifest is how a big job is run: write the ops once, dry-run the file, then apply it — and re-run the same manifest to recover an interrupted write (overrides are idempotent). op='Remove' CLEARS a nullable field — a nullable substruct, a nullable polymorphic field — and is REFUSED BY NAME on a required one rather than writing an invented null. Which verb an op may use is decided by the field's CARDINALITY, and this pre-flight is the source of truth for it: read the cardinality off the schema before composing the op. from_source= on an op takes that field's value from the plugin you NAME rather than from the load-order winner — how an edit is based on an authored plugin when a generated one is winning the record.")]
```
Avick3110: **`from_source=` is stated generally here, but only `op='CopyFrom'` honours it — and nothing refuses it on another verb.**

The folded sentence says "from_source= on an op takes that field's value from the plugin you NAME rather than from the load-order winner". `ops=` is the parameter a caller reads for the op shape, so that reads as a general per-op pole. It isn't one: `ApplyOp.FromSource`'s own description says "op='CopyFrom' only".

The mismatch is silent, not refused. `ApplyTools.cs:148` gates only `from=` on CopyFrom; `from_source` is a declared member, so strict binding accepts it and it maps straight to `BulkOp.FromPlugin`. From there `PrepareCopyFromSources` early-returns when no op has `Verb == "CopyFrom"` (`LoadOrderService.cs:5415`), and both `IsOffOrderCopySource` (`WritePatchBuilder.cs:688`) and `ResolveCopyPole` (`WritePatchBuilder.cs:697`) require the same verb.

So `{formid:'0012AB:Mod.esp', field_path:'BasicStats.Damage', op:'Set', value:'12', from_source:'Authored.esp'}` writes, reports success, and bases the edit on the winner — the generated plugin this very sentence is telling the caller to get out from under. Either scope the sentence to CopyFrom, or refuse `from_source` on a non-CopyFrom verb the way `from=` already is.

src/housecarl-mcp/ApplyTools.cs
```diff
@@ -63,7 +63,7 @@ public static string Apply(
...
             bool dry_run = false,
-        [Description("TRANSPORT: expand the read-back to the FULL deep field-by-field dump of every record this call touched (not just the edited leaves) — confirm composed structures landed and nothing else was disturbed WITHOUT enabling the patch in MO2. In place, the verify ALWAYS runs and shows compactly by default; this widens it. The read-back is the WRITTEN FILE's content, NOT load-order truth: the patch wins nothing until enabled + sorted in MO2.")]
+        [Description("TRANSPORT: expand the read-back to the FULL deep field-by-field dump of every record this call touched (not just the edited leaves) — confirm composed structures landed and nothing else was disturbed WITHOUT enabling the patch in MO2. In place, the verify ALWAYS runs and shows compactly by default; this widens it. The read-back is the WRITTEN FILE's content, NOT load-order truth: the patch wins nothing until enabled in MO2.")]
```
Avick3110: **Dropping "+ sorted" makes this true of the fresh-folder lane only — and `into=` is the other one.**

F49's deletion leaves "the patch wins nothing until enabled in MO2" as a flat statement, but `apply` has an `into=` lane that writes into an EXISTING patch mod: that mod is typically already enabled, its priority is already fixed, and enabling is not the remaining step.

This is the same case the PR corrects 20 lines away in `Program.cs`, where the standing sentence now reads "a write into an EXISTING mod keeps that mod's priority and may still need sorting above the current winner". After this PR the two surfaces disagree.

Concretely: `housecarl_apply into='houseCARL - MyPatch'` where that mod sits below the current winner. The read-back says the patch wins once the mod is enabled; it already is, so the caller stops and the record still loses. Same wording and same lane on `CreateTools.cs:46`.

The minimal fix is the clause the instructions already carry — e.g. "…until enabled in MO2 (a write into an existing mod keeps that mod's priority and may still need sorting)".

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -126,9 +126,9 @@ public static string Records(
...
             string? fields_source = null,
-        [Description("PROJECT: the shape of the answer — a SINGLE form plus that form's own sub-parameters: identity | summary (default) | fields | rows | everything | aggregate | delta | tree | chain | info_order. Sub-parameters live INSIDE the form that uses them (depth belongs to fields/rows/everything, group_by to aggregate, fields to fields/rows/delta/tree), so there is no flat spelling for an illegal pairing. Omit for summary rows.")]
+        [Description("PROJECT: the shape of the answer — a SINGLE form plus that form's own sub-parameters: identity | summary (default) | fields | rows | everything | aggregate | delta | tree | chain | info_order. Sub-parameters live INSIDE the form that uses them (depth belongs to fields/rows/everything, group_by to aggregate, fields to fields/rows/delta/tree), so there is no flat spelling for an illegal pairing. Omit for summary rows. aggregate counts ALL matches and limit= does not window its groups (offset= is refused there outright) — max_chars alone bounds the render, so a count is safe to size a job with.")]
```
Avick3110: **"max_chars alone bounds the render" holds on the scan lane, but nothing bounds an aggregate on the `formids=` lane.**

The first two clauses check out: `RecordsTools.cs:432` refuses `offset=` on aggregate, `CrossQuery` bumps its groups over ALL matches with no limit cap (`LoadOrderService.cs:4438`, `:4696`), and the scan lane's `RenderCrossQueryGroups` (`ReadTools.cs:435`) budgets group rows against `cap` and emits a truncation notice.

The list lane does not. `RecordsTools.cs:681` returns `RenderListAggregate(...)` directly, and that method (`RecordsTools.cs:2317`) takes no `cap` parameter at all — it appends every group row and returns without `RenderCap.Settle`, unlike every sibling render on the same lane.

So `formids=[…500 ids…]` with `project={form:'aggregate', group_by:'defined_in'}` and `max_chars=2000` can render hundreds of group rows well past the cap, with no notice saying so. Either pass the cap into `RenderListAggregate` (making the sentence true on both lanes), or scope the claim to the scan lane.

src/housecarl-mcp/CreateTools.cs
```diff
@@ -33,7 +33,7 @@ public static class CreateTools
...
...include an unknown or ambiguous record_type, a missing editorid, an illegal field op, a nested child with no resolvable parent, an ambiguous collection, and a type that cannot be created (the bare abstract base 'Global'/'GameSetting' needs a concrete subtype; the refusal names them). What else is reported, never silently: creating dialogue lines reports VOICE coverage (a response with no .fuz plays SILENT in game — the audio is yours to provide) and RESULT-SCRIPT binding (a bound script that is unwired or uncompiled runs nothing); creating cells reports the world content houseCARL does NOT author (lighting, terrain, water, navmesh — Creation Kit work). Authoring dialogue also AUTO-FILLS the Creation Kit's own bookkeeping and reports each field it filled: the INFO's FavorLevel (CNAM) and Flags (ENAM), a DLVW's DNAM and ENAM, a DLBR's Category (TNAM) and Flags (DNAM — an absent one publishes the branch as TopLevel), the DIAL's Priority (PNAM) and its SubtypeName (SNAM) marker derived from Subtype, and the QUST's NextAliasID (ANAM) plus each objective's Flags (FNAM) and each alias's Flags (FNAM) and VoiceTypes (VTCK). Every auto-fill is NON-OVERRIDE: a value you passed always wins.")]
```
Avick3110: **VTCK is not filled on "each alias" — only on Reference aliases.**

The roster reads "each alias's Flags (FNAM) and VoiceTypes (VTCK)", which puts both fills on every alias. The code gates them differently: `DialogueCkParity.cs:419` fills `Flags` for every alias, while `:428` fills `VoiceTypes` only under `IsReferenceAlias(alias) && !HasAliasVoiceTypes(alias)` — the comment there says so outright ("the CK writes both on a REFERENCE alias").

Authoring a QUST with a Location alias therefore reports `Aliases[0] … Flags (FNAM)` and no VTCK. A caller checking the write against this list reads the absent VTCK as a fill that was dropped, and hand-writes a subrecord the CK would not put there.

"…and each REFERENCE alias's VoiceTypes (VTCK)" would carry the gate.

src/housecarl-mcp/ApplyTools.cs
Avick3110: Folded — 00f1bf52. Traced the whole path and the finding holds: `ApplyTools.cs:148` gates `from=` on `CopyFrom` and refuses it by name on any other verb; `from_source` has no such gate, is a declared `ApplyOp` member so strict binding accepts it, and maps to `BulkOp.FromPlugin` at `:157`. Downstream every consumer tests the verb — `PrepareCopyFromSources` early-returns with no `CopyFrom` op (`LoadOrderService.cs:5416`), and `IsOffOrderCopySource` (`WritePatchBuilder.cs:688`) and `ResolveCopyPole` (`:697`) each require it — so the pole is dropped without a word.

The description now reads:

> from_source= (op='CopyFrom' only) takes that field's value from the plugin you NAME rather than from the load-order winner — how an edit is based on an authored plugin when a generated one is winning the record.

+12 characters; `housecarl_apply.ops` measures **1,076** on the wire.

The refusal is the other half and is not in this PR — a description-only PR should not change what a call does, and the fix is a pre-flight clause beside the `from=` one, in one sentence naming the op index and the verb it got. Filed as #630.

src/housecarl-mcp/ApplyTools.cs
Avick3110: Folded — aebdba87, on both read-backs. `apply` and `create` each have an `into=` lane that writes into an existing houseCARL patch mod whose priority is already fixed, so F49's flat deletion left the sentence true of the fresh-folder lane only, and disagreeing with the standing instructions twenty lines away was the tell.

Both now carry the instructions' own clause, unchanged:

> The read-back is the WRITTEN FILE's content, NOT load-order truth: the patch wins nothing until enabled in MO2, and a write into an EXISTING mod keeps that mod's priority and may still need sorting above the current winner.

+112 characters each: `housecarl_apply.readback` measures **539** on the wire, `housecarl_create.readback` **509**. One sentence, one wording, three surfaces — the instructions, `apply.readback`, `create.readback` — so a later edit to it cannot land on one and not the others. The changelog line that described the deletion is updated to describe the bound instead (8da2b949).

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded — 06752042. Verified on the list lane: `RecordsTools.cs:682` returns `RenderListAggregate(...)` and that method (`:2316`) has no cap parameter at all — it appends a row per group and returns without `RenderCap.Settle`, while every sibling render on the same lane settles. The scan lane does hold it, in `RenderCrossQueryGroups` (`ReadTools.cs:435`).

Wrote the sentence that is true on both rather than picking one:

> aggregate counts ALL matches and limit= does not window its groups (offset= is refused there outright) — on a SCAN max_chars alone bounds the render, while on a formids= list nothing bounds the group rows, so a count is safe to size a job with.

+66 characters; `housecarl_records.project` measures **664** on the wire.

Passing the cap into `RenderListAggregate` is the real fix and is not in this PR: a render that overruns its stated ceiling without a notice is a silent degradation, not a wording problem, and a description-only PR should not change what a call returns. Filed as #631. Once that lands, this sentence collapses back to "max_chars alone bounds the render" on both lanes — the issue says so.

src/housecarl-mcp/CreateTools.cs
Avick3110: Folded — edbea683. Checked against the code: `DialogueCkParity.cs:419` fills `Flags` for every alias, `:428` gates `VoiceTypes` behind `IsReferenceAlias(alias) && !HasAliasVoiceTypes(alias)`, and the note above it gives the reason — VTCK is an actor/reference concept, and filling it on a Location alias would over-add and false-warn on real CK-authored location aliases.

The roster now reads "…and each alias's Flags (FNAM) and each REFERENCE alias's VoiceTypes (VTCK)". +23 characters; `housecarl_create.records` measures **2,132** on the wire.

Two comments in `DialogueCkParity` itself carried the same error and are fixed in the same commit: the byte-parity roster at `:39` said "materialised per alias", and the QUST doc-comment at `:355` said "each alias's Flags (FNAM) + VoiceTypes (VTCK)". The gate was only stated at `:369`, next to the code — which is why the description was written from the wrong one.

src/housecarl-mcp/Program.cs
Avick3110: Follow-up on the separate note above: filed as #632.

Confirmed in code before filing — `EnableAndSort` (`PlaceTools.cs:298`) picks its sentence on CONTENTION only, never on the lane, so a fresh-folder placement over an existing provider still gets "SORT it (left pane) ABOVE the current winner(s)". That is `place` overstating, not this PR understating: on the fresh-folder lane the new mod outranks the previous winner the moment it is ticked, and the method's own no-contention arm already says so. The standing sentence in this PR is the true one on both lanes; #632 moves the place response onto it, along with the same unlaned wording in `LoadOrderService.PlaceOne` and `PlaceResult.CurrentWinner`.

The branch has been rebased since the reply above, so its commit reads as 12f2a4bd now; the text is unchanged.

PR #624, opened 2026-09-07
Annotate the FormID an element summary renders

### LINKED ISSUE #530: [GAP] resolve_names skips a FormLink rendered inside a container element summary note
From a Discord report (2026-08-25, houseCARL 1.9). After 2.0.

`resolve_names=true` leaves the FormLink inside a depth=2 container element's summary note (`[Effect] BaseEffect=033975:Skyrim.esm`) unannotated. `AnnotateLinks` in `LoadOrderService.cs` annotates only fields with `HasValue` and a parseable token; the element summary line is emitted `HasValue: false` with the FormID in `Note` (`ReadEngine.cs`).

Workaround in one call: `project.depth=3` promotes `Effects[i].BaseEffect` to its own row, which is resolved. So this is a papercut on output size at the cheaper depth, not a blocker.

Not #230, #231, #252.

Avick3110:
At `project.depth=2` a container element renders as a one-line summary that carries its identity field, and where that field is a FormLink the line shows a FormID — `Effects[0] = [Effect] BaseEffect=033975:Skyrim.esm`. `AnnotateLinks` read the reference off `FieldValue.Token`, which that line does not have (it is a no-value line with the FormID inside its prose `Note`), so `resolve_names=true` left it bare and the editorid was reachable only by re-reading at `depth=3`. Rather than teach the annotator to parse the note, the read now carries the FormID it rendered on a new structural member, `FieldValue.NoteRef`, and `AnnotateLinks` resolves whichever carrier the line rendered its reference on — the token, or that. One rule, one shape of value, and every render site already appended `Link` regardless of `HasValue`, so the text, json, dense and rows lanes all show it with no per-lane change.

Two bounds, both in the code and the changelog entry: an owned child record's own FormKey (`[DialogResponses 4D9A74:Plugin.esp editorid=…]`) is the element's own identity, already spelled with its editorid, so it is not annotated with itself; and an element with two or more FormLink fields still renders as a bare `[Type]` and shows no FormID to annotate, for the reason `LoneFormLinkIdentity` already gives.

Three tests in `RecordsSummaryLinkTests` over the shared records world's two-effect spell: the text annotation beside the unchanged summary, the json `link` sibling of the `note`, and the un-annotated line when `resolve_names` is off. Two of the three fail before the fix.

Ran on this branch: `dotnet build -c Release` clean, `dotnet test --filter "tier!=bridge"` 1792/1792, `ci-all` 127/127.

Closes #530

### DIFF
--- a/src/housecarl-core/ReadEngine.cs
+++ b/src/housecarl-core/ReadEngine.cs
@@ -18,12 +18,13 @@ namespace HousecarlCore;
 /// read-proof, and diff. Used to decode a value that is correct but opaque — a biped-slot bitmask into its slot
 /// numbers — without disturbing the token that must round-trip. Null on every leaf that needs no annotation.</para>
 ///
-/// <para><see cref="Link"/> is the resolve_names annotation: when a leaf's <see cref="Token"/> is a
-/// form reference (a token that round-trips to a FormKey), the SERVICE layer resolves its target's identity
-/// (editorid/name) against the load order and hangs it here. Like <see cref="Display"/> it is DISPLAY-ONLY — never
-/// part of the round-trip <see cref="Token"/>, so it is invisible to write, read-proof, and diff. Populated by the
-/// service (which holds the resolver), never by the core read (which reads one record's bytes and cannot resolve a
-/// target). Null unless resolve_names was requested and the leaf carries a resolvable FormKey.</para></summary>
+/// <para><see cref="Link"/> is the resolve_names annotation: when a line RENDERS a form reference — its
+/// <see cref="Token"/>, or the <see cref="NoteRef"/> its summary note shows — the SERVICE layer resolves that
+/// target's identity (editorid/name) against the load order and hangs it here. Like <see cref="Display"/> it is
+/// DISPLAY-ONLY — never part of the round-trip <see cref="Token"/>, so it is invisible to write, read-proof, and
+/// diff. Populated by the service (which holds the resolver), never by the core read (which reads one record's
+/// bytes and cannot resolve a target). Null unless resolve_names was requested and the line renders a
+/// FormKey.</para></summary>
 /// <param name="Present">Is anything THERE? True for a value and for a present container/substruct (whose own
 /// <paramref name="Count"/> says how much); false only when the leaf holds nothing — absent, no such field,
 /// unreadable. Carried structurally because the two render alike (both are parenthesised notes), so a consumer
@@ -37,9 +38,14 @@ namespace HousecarlCore;
 /// carried structurally for the same reason <paramref name="Present"/> is: the folded line's TEXT is prose, and a
 /// consumer must be able to read the values, the resolve_names links and the counts out of the row without
 /// parsing it.</param>
+/// <param name="NoteRef">The form reference this line's <paramref name="Note"/> RENDERS, when the line carries no
+/// round-trip <paramref name="Token"/> — a container element's summary identity (<c>[Effect]
+/// BaseEffect=033975:Skyrim.esm</c>). Carried structurally for the same reason <paramref name="Present"/> is: the
+/// note is prose, and resolve_names must reach the FormID a line shows without parsing it. Null on every line
+/// whose note renders no form reference.</param>
 public sealed record FieldValue(string Path, bool HasValue, string? Token, string? Note, string? Display = null, ResolvedRef? Link = null,
                                 bool Present = true, int? Count = null, bool Readable = true,
-                                IReadOnlyList<FieldValue>? Cells = null);
+                                IReadOnlyList<FieldValue>? Cells = null, string? NoteRef = null);
 
 /// <summary>The resolved identity of a form reference — the shared contract behind housecarl_resolve (a full row)
 /// and the resolve_names field annotation. <see cref="Resolved"/> false ⇒ the FormKey is valid but not present in
@@ -620,8 +626,11 @@ public static class ReadEngine
         // a container or substruct — summarise (with an element identity where we can), then maybe open it.
         // Present/Count are set on the DEEP path too: they exist so a consumer never parses a note to decide
         // presence, and left at their defaults a depth-2 read of an EMPTY list claims content.
+        // NoteRef carries the FormID the summary spelled, so resolve_names annotates the reference this line SHOWS
+        // by the one rule it applies to a token — no second, note-shaped special case.
         int? deepCount = val is System.Collections.IEnumerable de and not string ? de.Cast<object?>().Count() : null;
-        if (!Emit(sink, ref budget, new FieldValue(path, false, null, ElementSummary(val, isDict), Present: true, Count: deepCount))) return;
+        var summary = ElementSummary(val, isDict, out var summaryRef);
+        if (!Emit(sink, ref budget, new FieldValue(path, false, null, summary, Present: true, Count: deepCount, NoteRef: summaryRef))) return;
 
         // Two POLYMORPHIC-ARM families normally stop here at their identity summary, hiding their VALUE, and both
         // surface it ONE bounded level deeper even at the depth floor so a read reaches parity with the write
@@ -836,8 +845,17 @@ public static class ReadEngine
     /// (<see cref="SummariseContainer"/>); for a struct, <c>[TypeName]</c> plus a representative identity
     /// field (Name/EditorID/Title) where present — so a list line like
     /// <c>Properties[5] = [ScriptObjectProperty] Name=DAK_HorseBuyPerk</c> reveals which element is which.</summary>
-    static string ElementSummary(object val, bool isDict = false)
+    static string ElementSummary(object val, bool isDict = false) => ElementSummary(val, isDict, out _);
+
+    /// <summary>Overload that also yields the form reference the summary RENDERED, so a caller building the line's
+    /// <see cref="FieldValue"/> can carry it on <see cref="FieldValue.NoteRef"/> and resolve_names annotates the
+    /// FormID a reader can see. It is the identity FIELD's target (an <c>Effect</c>'s BaseEffect, a
+    /// <c>PerkPlacement</c>'s Perk) — a reference OUT of this element, exactly what a leaf Token is. An owned child
+    /// RECORD's own FormKey is not one: it is this element's own identity, already spelled with its editorid, and
+    /// resolving it would annotate a line with itself.</summary>
+    static string ElementSummary(object val, bool isDict, out string? refToken)
     {
+        refToken = null;
         if (val is System.Collections.IEnumerable && val is not string) return SummariseContainer(val, isDict);
         var t = val.GetType();
         var typeName = RecordNaming.StripGetterInterface(RecordNaming.StripOverlay(t.Name));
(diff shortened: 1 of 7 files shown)


src/housecarl-core/ReadEngine.cs
```diff
@@ -620,8 +626,11 @@ static void Expand(object? val, Type declaredType, object parent, string path, i
         // a container or substruct — summarise (with an element identity where we can), then maybe open it.
         // Present/Count are set on the DEEP path too: they exist so a consumer never parses a note to decide
         // presence, and left at their defaults a depth-2 read of an EMPTY list claims content.
+        // NoteRef carries the FormID the summary spelled, so resolve_names annotates the reference this line SHOWS
+        // by the one rule it applies to a token — no second, note-shaped special case.
         int? deepCount = val is System.Collections.IEnumerable de and not string ? de.Cast<object?>().Count() : null;
-        if (!Emit(sink, ref budget, new FieldValue(path, false, null, ElementSummary(val, isDict), Present: true, Count: deepCount))) return;
+        var summary = ElementSummary(val, isDict, out var summaryRef);
+        if (!Emit(sink, ref budget, new FieldValue(path, false, null, summary, Present: true, Count: deepCount, NoteRef: summaryRef))) return;
```
Avick3110, line 634: The `rows` lane now prints the annotation on a cell that no longer shows the FormID it names, and prints it twice in the row.

`RowProjection.Cell` trims a container-summary cell back to its bare type whenever the row also holds sub-field cells under it (`trimToType: kept.Any(g => IsUnder(g.Path, f.Path))`), and it appends `f.Link` **after** that trim:

```csharp
if (trimToType && !f.HasValue && ...) val = n[..(close + 1)];   // "[Effect] BaseEffect=033975:Skyrim.esm" -> "[Effect]"
if (f.Link is not null) val += $" ({Wire.LinkText(f.Link)})";
```

Before this change the element line carried no `Link`, so the trim was clean. Now the element line is annotated from `NoteRef`, and the trimmed cell keeps the parenthetical.

Input: `housecarl_records formids=[<SpellA>] project={form:'rows', fields:['Effects'], resolve_names:true}` — `form='rows'` defaults to `RowProjection.DefaultDepth` = 4, so each element's sub-fields are emitted and `trimToType` is true for the element cell.

Output today:

```
Effects[0] = [Effect] | BaseEffect=033975:Skyrim.esm (-> HcRecMgefFire) | Data.Magnitude=5 | ...
```

Output on this branch:

```
Effects[0] = [Effect] (-> HcRecMgefFire) | BaseEffect=033975:Skyrim.esm (-> HcRecMgefFire) | Data.Magnitude=5 | ...
```

The first cell reads as an identity with nothing to attach it to — the FormID it resolves was deliberately trimmed out one line above — and it repeats the annotation on the very next cell, which is exactly the repetition `trimToType` exists to prevent ("On a row whose own sub-fields are right there it would just repeat them, so it keeps only its type."). Same shape for any lone-FormLink element, e.g. `PerkPlacement` -> `[PerkPlacement] (-> ...) | Perk=03AF81:Skyrim.esm (-> ...)`.

The existing rows coverage does not catch it: `RecordsRowsFormTests.ResolveNamesAnnotatesInsideARow` only asserts `Contains("HcRecMgefFire", RowLine(...))`, which passes with the annotation on either cell or both.

Either drop the `Link` when the cell is trimmed (the FormID it names is gone), or keep the untrimmed note on a cell that carries a link.

Avick3110: Real, and reproduced exactly as written. The rows read at `resolve_names=true` printed `Effects[0] = [Effect] (→ HcRecMgefFire) | BaseEffect=033975:Skyrim.esm (→ HcRecMgefFire) | …` — the first cell naming a FormID the trim had already taken out of it, and the name repeated on the next cell.

Fixed by the first of the two options: `RowProjection.Cell` now records whether the trim actually fired and skips the `Link` append when it did. The trim and the annotation describe the same FormID, so when the trim removes it the annotation goes with it; the sub-field cell that still shows the FormID carries the name, once. Only the trimmed cell is affected — an untrimmed summary cell (a collapsed element with no sub-field cells on the row) still shows both, and the json lane is untrimmed and unchanged, so its `cells` still carry the element's own `link` beside its full `note`.

Test: `RecordsRowsFormTests.ATrimmedElementCellKeepsNeitherTheFormIdNorItsName` asserts the row's first cell is a bare `[Effect] | BaseEffect=…` and that the resolved editorid appears exactly once in the row — the count is what the existing `ResolveNamesAnnotatesInsideARow` `Contains` could not see. It fails before the fix on the first assertion.

Build clean, `tier!=bridge` 1793/1793, `ci-all` all pass. 385a4117.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2455,7 +2458,10 @@ static RecordFields AnnotateLinks(RecordFields rf, LoadOrderResolver.IndexView v
         for (int i = 0; i < rf.Fields.Count; i++)
         {
             var f = rf.Fields[i];
-            if (f.HasValue && f.Token is { } tok && FormKey.TryFactory(tok, out var fk) && !fk.IsNull)
+            // Whichever carrier the line RENDERED its reference on: the round-trip token, or the FormID a
+            // container element's summary note spelled (FieldValue.NoteRef). One rule, one shape of value.
+            var rendered = f.HasValue ? f.Token : f.NoteRef;
```
Avick3110, line 2463: At `project.depth` >= 3 the same target is now annotated twice on consecutive lines:

```
Effects[0] = [Effect] BaseEffect=033975:Skyrim.esm   (→ HcRecMgefFire)
Effects[0].BaseEffect = 033975:Skyrim.esm   (→ HcRecMgefFire)
```

The rows lane treats exactly that repetition as noise and suppresses it (`RowProjection.cs:173` — "it would just repeat them"); the fields, json and dense lanes keep both.

Concrete cost: `ReadTools.cs` charges every field line against `room`, so `project={form:"fields", fields:["Effects"], depth:3, resolve_names:true}` on a many-effect spell (or an NPC's `Perks` at depth 3) now spends one duplicate parenthetical per element and truncates earlier than it did, dropping trailing fields that previously fitted under `max_chars`.

Either skip the NoteRef annotation when the element's own link leaf is also emitted, or say in the changelog that the fields lane repeats it deliberately.

src/housecarl-core/ReadEngine.cs
```diff
@@ -37,9 +38,14 @@ namespace HousecarlCore;
 /// carried structurally for the same reason <paramref name="Present"/> is: the folded line's TEXT is prose, and a
 /// consumer must be able to read the values, the resolve_names links and the counts out of the row without
 /// parsing it.</param>
+/// <param name="NoteRef">The form reference this line's <paramref name="Note"/> RENDERS, when the line carries no
+/// round-trip <paramref name="Token"/> — a container element's summary identity (<c>[Effect]
+/// BaseEffect=033975:Skyrim.esm</c>). Carried structurally for the same reason <paramref name="Present"/> is: the
+/// note is prose, and resolve_names must reach the FormID a line shows without parsing it. Null on every line
+/// whose note renders no form reference.</param>
```
Avick3110: The justification here is the `Present`/`Count` doctrine — "the note is prose, and resolve_names must reach the FormID a line shows without parsing it" — but `NoteRef` stops at the service: `JsonWire.WriteLeaf` never writes it.

So a json consumer at depth 2 gets

```json
{"path":"Effects[0]","note":"[Effect] BaseEffect=033975:Skyrim.esm","link":{"resolved":true,"editorid":"HcRecMgefFire"}}
```

— `link` names the target, but the FormID is only inside the prose. A caller that reads `Effects` at depth 2 and then wants to re-read each base effect through `formids=[…]` still has to regex `note`, which is the thing `Present`, `Count` and `Cells` exist to prevent.

Either emit it (a `note_ref` sibling of `note`), or drop the structural claim from the doc — as it stands the member is a service-internal carrier and the paragraph oversells it.

src/housecarl-mcp-tests/RecordsRowsFormTests.cs
```diff
@@ -261,6 +261,19 @@ public void ResolveNamesAnnotatesInsideARow()
...
+    public void ATrimmedElementCellKeepsNeitherTheFormIdNorItsName()
+    {
+        var r = RecordsTools.Records(Svc, formids: new[] { Fid(W.SpellA) },
+            project: new RecordsTools.RecordsProject { form = "rows", fields = new[] { "Effects" }, resolve_names = true });
+        var row = RowLine(r, "Effects[0]");
+        // The element cell is trimmed to its bare type because BaseEffect is a cell of its own on the same row.
+        // The name the trim's FormID resolved to goes with it: it names a FormID this cell no longer shows, and
+        // the cell that still shows it carries the same name one cell along.
+        Assert.Contains("[Effect] | BaseEffect=", row);
+        Assert.Equal(1, row.Split("HcRecMgefFire").Length - 1);
```
Avick3110: `RecordsTestBase.CountOf(haystack, needle)` already does this, and `Assert.Equal(1, CountOf(…))` is the suite's idiom for exactly this shape (`RecordsComparisonFormTests.cs:262`, `RecordsOffOrderPathTests.cs:45`, `RecordsPluginFileSourceTests.cs:59`).

```suggestion
        Assert.Equal(1, CountOf(row, "HcRecMgefFire"));
```

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real, and reproduced: at `depth=3` with `resolve_names=true` the fields lane prints

```
Effects[0] = [Effect] BaseEffect=000805:HcRecMaster.esm   (→ HcRecMgefFire)
Effects[0].BaseEffect = 000805:HcRecMaster.esm   (→ HcRecMgefFire)
```

Took the second option — the changelog now says the repetition is deliberate — rather than the first, for two reasons.

The FormID itself already repeats on those two lines, and did before this branch: at `depth=3` the fields lane prints the element's summary and then the leaf under it, and both spell `000805:HcRecMaster.esm`. Suppressing only the name would leave a line showing a FormID with no name beside it while the identical FormID one line down has one, and would make the same summary line annotated at `depth=2` and bare at `depth=3`. The name follows the FormID the line shows; the extra characters are the ones the fields lane already spends repeating the FormID.

The rows lane's suppression is not that rule. It drops the name because `RowProjection.Cell`'s trim took the FormID out of the cell — the cell shows no reference, so there is nothing left to name (385a4117). Where the FormID stays on the line, the name stays with it.

Changelog says so, and `RecordsSummaryLinkTests.ADeeperReadNamesTheSummaryAndItsLeafBecauseBothShowTheFormId` asserts both lines carry the parenthetical, so the repetition is pinned rather than only described. Being plain about it: that test fails against the base branch (where the summary line has no name) but cannot fail against a changelog sentence, so it pins the behaviour rather than proving a fix.

Build clean, `tier!=bridge` 1795/1795, `ci-all` all pass. cb2a9f85.

src/housecarl-core/ReadEngine.cs
Avick3110: Real — `JsonWire.WriteLeaf` wrote `path`, `value`/`note`, `display` and `link`, and never `NoteRef`, so a json consumer at depth 2 got the target named but not spelled.

Took the first option: `WriteLeaf` now emits a `note_ref` sibling of `note`.

```json
{"path":"Effects[0]","note":"[Effect] BaseEffect=000805:HcRecMaster.esm","note_ref":"000805:HcRecMaster.esm","link":{"resolved":true,"type":"MagicEffect","editorid":"HcRecMgefFire","name":null}}
```

A caller feeds that straight back through `formids=` instead of regexing the note. It is the line's own shape, not the annotation's, so it is written with `resolve_names` off as well; a folded row writes no `note` and writes no `note_ref` either — its leaves carry theirs in `cells`. The `NoteRef` paragraph now names the wire key rather than claiming the reach in the abstract, and the changelog entry says the json line carries it.

Test: `RecordsSummaryLinkTests.TheJsonSummaryLineCarriesTheFormIdItRenderedBesideTheProse` — fails before the fix (no such property on the element line), passes after. 7d7dcb88.

src/housecarl-mcp-tests/RecordsRowsFormTests.cs
Avick3110: Taken — the hand-rolled split is now `Assert.Equal(1, CountOf(row, "HcRecMgefFire"))`, the same shape as `RecordsComparisonFormTests.cs:262`. Suite green. 26c79caa.

PR #625, opened 2026-09-07
Read a short-circuit whose value is a mid-list call argument

### LINKED ISSUE #393: [GAP] decompile_script: a short-circuit arm whose operand evaluation emits statements fails the whole function
Reported externally via Discord (2026-08-15) and triaged before filing. Reporter unnamed here — their OK to attribute wasn't obtained. Third-party machine paths scrubbed.

## Goal

Decompile a live deployed `ST_Config.pex` to confirm the shipped bytecode matched current source (MCM blank-content-pane forensics on a third-party modlist).

## What fell short

`housecarl_decompile_script` returned PARTIAL — 57 functions, 1 FAILED:

```
decompile PARTIAL: ST_Config.pex -> <patch>\Source\Scripts\ST_Config.psc
57 function(s), 1 FAILED - each is a loud comment block with its raw bytecode in the .psc, which will NOT compile as-is:
  FAIL ST_Config.BuildIntegrationsBlock: short-circuit arm @35..36 produced statements
note: optimizer-compiled patterns detected (Caprica class) - source is correct; byte-identity vs the original .pex under the CK compiler is not expected.
```

Call: `housecarl_decompile_script` with `pex=<...>\Sanguine's Trade - An Economy Mod\Scripts\ST_Config.pex`, `patch_name=houseCARL_STConfig_Forensics`.

The failing function's body is unreadable (raw-bytecode comment block), so that one function could not be verified against source. Every other function in the same `.pex` reconstructed cleanly. **The tool reported the failure loudly and did not emit wrong source — behaving as documented** (`PEX_DECOMPILE_TOOL_PLAN` §1, per-function loud failures).

Frequency: one-off so far — 1 of 57 functions in one `.pex`; four other `.pex` files decompiled 100% clean in the same session (SkyUI + MCM Recorder `SKI_ConfigBase`, `SKI_ConfigManager`).

## Workaround used

None needed — the verification targets (`OnConfigOpen` / `OnConfigClose` / `OnPageReset` / `Pages` / `GetVersion`) all decompiled fine, so the task completed. Would have been a hard block had the answer lived in `BuildIntegrationsBlock`.

## Ideal shape

Reconstruct the arm rather than failing the function.

**Not** a best-effort body with an inline uncertainty comment, which is what the report originally asked for. The statement in question is *conditionally* evaluated; emitting it unconditionally yields source that reads plausibly and is semantically wrong. The reporter's own goal was verifying shipped bytecode against source — a plausible-wrong body would have silently passed that check, where a bytecode dump does not. Recording this so nobody folds the original wording verbatim.

Two acceptable resolutions:

1. Reconstruct the shape — inline pure materialized temps back into the arm's expression tree so the arm produces a value, not statements.
2. If the shape is genuinely irreducible, add it to the named-irreducible list with its specific reason, alongside `nl_mcm_module.Get`.

## Triage notes

- **Throw site:** `src/housecarl-core/PapyrusDecompiler.cs:580` — `Structure(...)` over the arm must yield only pending values; any statement is a deliberate loud fail. The report matches the code exactly.
- **Not a known duplicate.** No open decompile issue.
- **This is irreducible #2.** The class was the spike's residual #5 ("10 short-circuit arm with statements", `PEX_DECOMPILE_SPIKE_FINDINGS_2026-06-10` §5), fixed by the pre-if `FlushPendingCalls` drain — after which the bulk gate measured **100.0% structuring across 35,604 functions with exactly one named irreducible** (`nl_mcm_module.Get`). So this is a sub-shape outside a 3,400-plugin corpus.
- **Suspected mechanism (unverified — needs the repro):** the arm's sub-structure hits the value-reused-temp materialization (findings fix #4), which does `stmts.Add(assignment)`. That is the same hazard the code already keeps deliberately loud for `while` conditions — hoisting an assignment out of a lazily-evaluated region changes semantics. If that's the cause, the fail-loud is correct and resolution (2) may be the honest answer.
- **Repro is obtainable without the reporter's disk.** The mod is public on Nexus (`Sanguine's Trade - An Economy Mod`, mod id 185788) — pull it and extract `ST_Config.pex`. Version caveat: reported 2026-08-15, mod updated 2026-08-16, so the shipped `.pex` may no longer be the one that failed. Ask the reporter for the exact version, or for the `.pex` itself, before concluding the shape is gone.
- **Survives tool-surface 2.0.** `decompile_script` keeps its own tool and shape under S4; 2.0 does not touch the decompiler engine.
- **Priority: low.** One function of 57, loud, workaround available, no data-integrity risk.

Avick3110:
﻿The decompiler took a short-circuit only when the false-path jump landed *on* the instruction that consumes the temp. When the short-circuit's value is a call argument that is not the last one, the arguments after it evaluate before the call, so the jump lands on the next argument's first instruction and the consuming call sits further on. The reader took that for a plain `if`, found the condition temp read again at the call, and promoted it to a named local. That promotion is what broke the reported function: a later arm of the same expression then wrote a materialized local, which is a statement, and a statement inside a lazily-evaluated arm cannot be hoisted out — so the function failed to a raw-bytecode comment block. The fix broadens the match to the shape's actual definition: the arm rewrites the same temp without reading it first, and the join side reads it before rewriting it. The old jump-lands-on-the-consumer form stays accepted as-is. The "without reading it first" half keeps the broadened match off a guarded block that reads *and* rewrites its condition temp — that is a reused condition temp, which the promotion path owns and reads cleanly today — so nothing that worked before changes.

Two things follow. Because the CK's own compiler emits this form, the old reading also fired the optimizer-compiled ("Caprica class") marker on canonical CK output — the note in the issue's own transcript is that false positive, and it stops firing. And an arm that genuinely evaluates a statement still fails loud; its message now names the first statement, so the next report pins the sub-shape without the `.pex`.

Closes #393.

## Repro

The reporter's `ST_Config.pex` was not obtainable (mod updated the day after the report). A second instance was found instead, in Aaron's own ARR 2.0 modlist: `Devotion/Scripts/PDV__ManagerQuest.pex`, `SyncKhajiitEmphasisFamily`, failing at `short-circuit arm @34..37`. Tracing it gave the mechanism above, and it is the promotion — not an irreducible arm — that is the cause, so resolution (1) in the issue applies, not (2). The function now decompiles to

```papyrus
SyncRaceRewardSpell(playerRef, t1, isActive && activeTier == Self.TIER_SEEKER, "Khajiit " + label + " T1")
```

and the whole 1,287-function script comes back with zero failures and zero optimizer hints.

## What was run

- `dotnet build -c Release`, `dotnet test --filter "tier!=bridge"` — 1774 passed. `ci-all` — 127/127, decompile-guard's golden unchanged.
- **Round trip through the real compiler** (local; CI has no CK compiler). Authored `HC_ScArgProbe.psc` with the shape, compiled it with the CK's `PapyrusCompiler.exe`, decompiled the `.pex` with this branch, recompiled the decompiled source: **instruction streams byte-identical**. That stream is what `DecompileShortCircuitTests` pins by hand.
- **Corpus sweep**, 18,562 `.pex` files across both local Skyrim modding trees (47,580 functions after de-duplication), before and after:

  | | failed functions | optimizer hints | files flagged optimizer-compiled |
  |---|---|---|---|
  | before | 6 | 17 | 4 |
  | after | 5 | 9 | 3 |

  Re-swept over the larger of the two trees (14,424 unique `.pex`) after the review added the
  `!ReadsBeforeWrite(i + 1, …)` conjunct: totals and every per-script source hash are identical with
  and without it, so the narrowing costs the corpus nothing.

  The second review found two more shapes the broadened match stole from the plain-`if` path: an
  if/else whose then-block writes the condition temp (the arm swallowed the then-block's trailing
  `JMP`), and a temp the promotion path had already turned into a named local (writes to it are
  statements, so the arm could only ever fail). Both are declined now — a trailing `JMP` in the
  region, and `Materialized`. Both trees were swept again for each fix (18,408 unique `.pex`,
  57,561 functions): totals and every per-script source hash identical each time.

  The short-circuit class is gone; the two remaining classes are the known `nl_mcm_module.Get` irreducible and a `::NoneVar` case in `Condiexp_RandomFrostbite.Random` that this change does not touch. No new failure class, and the decompiled source of every other script in the corpus is byte-identical (SHA-256 per script, before vs after).

## Notes

- The `WritesDestIn(...)` half of the new match — the arm must write the temp — makes no difference anywhere in that corpus; the whole sweep's output is identical with and without it. It is kept because it is half of what a short-circuit arm *is* (the arm produces the right operand), and it does change what a block that never touches its condition temp is told: without it, that shape's loud failure blames a short-circuit arm instead of naming the temp. `DecompileShortCircuitTests` pins that. Plain-`if` reading stays covered by decompile-guard's exact golden.
- The `::NoneVar` failure class above (4 functions in one script) is a separate shape and is not filed as an issue yet.

### DIFF
--- a/src/housecarl-core/PapyrusDecompiler.cs
+++ b/src/housecarl-core/PapyrusDecompiler.cs
@@ -10,7 +10,8 @@ namespace HousecarlCore;
 ///   - jump offsets relative to the jump instruction itself
 ///   - while  = cond; JMPF -> E; body; JMP (backward); E:
 ///   - if     = cond; JMPF -> L; then; JMP -> M; L: else; M:   (JMP -> L means no else)
-///   - and/or = short-circuit JMPF/JMPT landing ON the final consuming conditional jump, same temp
+///   - and/or = short-circuit JMPF/JMPT whose arm rewrites the same temp, read again at the join —
+///     usually landing ON the consumer, further on when later call arguments evaluate in between
 ///   - auto-prop backing var ::Name_var; AutoReadOnly = GET returning a literal
 ///   - compiler-generated GotoState/GetState in the '' state (skipped on emit)
 ///   - FunctionFlags raw bits: bit0 = Global, bit1 = Native (Mutagen's enum names sit one off)
@@ -559,12 +560,18 @@ public sealed class PapyrusDecompiler
                             target = hi;
                         }
 
-                        // Short-circuit: the jump lands ON the instruction that consumes the temp as a
-                        // source — another conditional jump (plain &&/||), a CAST hop into a different
-                        // temp (nested mixed-temp conditions), an ASSIGN/RETURN/call-arg (`x = a || b`).
-                        // Temps only: a real-var condition is always a plain if.
+                        // Short-circuit: the arm leaves the right operand in the SAME temp and the
+                        // join side still reads it. The jump usually lands ON that consumer — another
+                        // conditional jump (plain &&/||), a CAST hop into a different temp (nested
+                        // mixed-temp conditions), an ASSIGN/RETURN/call-arg (`x = a || b`). When the
+                        // value is a call argument that is NOT the last one, the arguments after it
+                        // evaluate first and the consuming call sits further on, so also accept an
+                        // arm that writes the temp whose value is read before being rewritten at or
+                        // after the join. Temps only: a real-var condition is always a plain if.
                         if (condName is not null && IsTemp(condName) && target < hi
-                            && ConsumesAsSource(_ins[target], condName))
+                            && (ConsumesAsSource(_ins[target], condName)
+                                || (WritesDestIn(i + 1, target, condName)
+                                    && ReadsBeforeWrite(target, hi, condName))))
                         {
                             var (left, leftStart) = Consume(condName, i);
                             // Pre-if discarded-result calls may still pend here (their temps can be
@@ -573,7 +580,12 @@ public sealed class PapyrusDecompiler
                             FlushPendingCalls(stmts);
                             // Evaluate the right side (cur+1 .. target) — must produce only pending values.
                             var sub = Structure(i + 1, target, flushAtEnd: false, exits: new HashSet<int>(), cont: target);
-                            if (sub.Count > 0) throw new StructureException($"short-circuit arm @{i + 1}..{target} produced statements");
+                            // The arm is lazily evaluated, so a statement in it cannot be hoisted out
+                            // without changing semantics. Name the first one: it is what pins the
+                            // sub-shape when the pex itself is not to hand.
+                            if (sub.Count > 0)
+                                throw new StructureException(
+                                    $"short-circuit arm @{i + 1}..{target} evaluates a statement, not just a value (first: {sub[0].Trim()})");
                             var (right, _) = Consume(condName, target);
                             var combined = new EBin(op == InstructionOpcode.JMPF ? "&&" : "||", left, right);
                             SetPending(condName, combined, stmts, leftStart);
@@ -1041,6 +1053,16 @@ public sealed class PapyrusDecompiler
                 if (ConsumesAsSource(_ins[k], name)) return true;
                 if (WritesDest(_ins[k], name)) return false;
             }
+            return false;
+        }
+
+        /// <summary>Is <paramref name="name"/> written as a destination anywhere in [lo, hi)?
+        /// A short-circuit arm always writes its own temp; a plain if's condition temp is dead in
+        /// the block it guards.</summary>
+        bool WritesDestIn(int lo, int hi, string name)
+        {
+            for (int k = lo; k < hi && k < _ins.Count; k++)
+                if (WritesDest(_ins[k], name)) return true;
             return false;
         }
 
--- a/src/housecarl-core/PapyrusDecompiler.cs
+++ b/src/housecarl-core/PapyrusDecompiler.cs
@@ -1060,8 +1060,8 @@ public sealed class PapyrusDecompiler
         }
 
         /// <summary>Is <paramref name="name"/> written as a destination anywhere in [lo, hi)?
-        /// A short-circuit arm always writes its own temp; a plain if's condition temp is dead in
-        /// the block it guards.</summary>
+        /// A short-circuit arm always writes its own temp, so a guarded block that never touches
+        /// the condition temp is not an arm and must not be reported as one.</summary>
         bool WritesDestIn(int lo, int hi, string name)
         {
             for (int k = lo; k < hi && k < _ins.Count; k++)
--- a/src/housecarl-core/PapyrusDecompiler.cs
+++ b/src/housecarl-core/PapyrusDecompiler.cs
@@ -567,10 +567,13 @@ public sealed class PapyrusDecompiler
                         // value is a call argument that is NOT the last one, the arguments after it
                         // evaluate first and the consuming call sits further on, so also accept an
                         // arm that writes the temp whose value is read before being rewritten at or
-                        // after the join. Temps only: a real-var condition is always a plain if.
+                        // after the join. An arm that READS the temp first is not an arm at all: it
+                        // is a guarded block over a reused condition temp, which the promotion path
+                        // below owns. Temps only: a real-var condition is always a plain if.
                         if (condName is not null && IsTemp(condName) && target < hi
                             && (ConsumesAsSource(_ins[target], condName)
                                 || (WritesDestIn(i + 1, target, condName)
+                                    && !ReadsBeforeWrite(i + 1, target, condName)
                                     && ReadsBeforeWrite(target, hi, condName))))
                         {
                             var (left, leftStart) = Consume(condName, i);
--- a/src/housecarl-core/PapyrusDecompiler.cs
+++ b/src/housecarl-core/PapyrusDecompiler.cs
@@ -569,10 +569,13 @@ public sealed class PapyrusDecompiler
                         // arm that writes the temp whose value is read before being rewritten at or
                         // after the join. An arm that READS the temp first is not an arm at all: it
                         // is a guarded block over a reused condition temp, which the promotion path
-                        // below owns. Temps only: a real-var condition is always a plain if.
+                        // below owns. A real arm is straight-line, so a trailing JMP in the region
+                        // means an if/else or a while, not an arm. Temps only: a real-var condition
+                        // is always a plain if.
                         if (condName is not null && IsTemp(condName) && target < hi
                             && (ConsumesAsSource(_ins[target], condName)
                                 || (WritesDestIn(i + 1, target, condName)
+                                    && (target - 1 <= i || _ins[target - 1].OpCode != InstructionOpcode.JMP)
                                     && !ReadsBeforeWrite(i + 1, target, condName)
                                     && ReadsBeforeWrite(target, hi, condName))))
                         {
--- a/src/housecarl-core/PapyrusDecompiler.cs
+++ b/src/housecarl-core/PapyrusDecompiler.cs
@@ -571,8 +571,9 @@ public sealed class PapyrusDecompiler
                         // is a guarded block over a reused condition temp, which the promotion path
                         // below owns. A real arm is straight-line, so a trailing JMP in the region
                         // means an if/else or a while, not an arm. Temps only: a real-var condition
-                        // is always a plain if.
-                        if (condName is not null && IsTemp(condName) && target < hi
+                        // is always a plain if, and a temp the promotion path already took is a named
+                        // local — writes to it are statements, which an arm can never carry.
+                        if (condName is not null && IsTemp(condName) && !Materialized.Contains(condName) && target < hi
                             && (ConsumesAsSource(_ins[target], condName)
                                 || (WritesDestIn(i + 1, target, condName)
                                     && (target - 1 <= i || _ins[target - 1].OpCode != InstructionOpcode.JMP)
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -91,6 +91,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   skill directory that is an immediate child of `skills/`. The installer places it in the same place it always
   did (`~/.agents/skills/housecarl`), so an install run from `houseCARL-Setup.exe` is unchanged; a copy taken by
   hand out of the zip comes from the new path.
+- **`housecarl_decompile_script` now reconstructs a short-circuit (`&&` / `||`) whose value is a call argument
+  followed by another argument.** The arguments after it evaluate before the call, so the false-path jump lands on
+  the next argument's first instruction rather than on the call itself; the reader took only the jump-lands-on-the-
+  consumer form, read the rest as a plain `if`, and promoted the condition temp to a named local — which then made
+  a later arm of the same expression emit statements and fail the whole function to a raw-bytecode comment block.
+  The expression now comes back whole, and because the CK compiler itself emits this form, the optimizer-compiled
+  ("Caprica class") note it used to trigger no longer fires on it. An arm that genuinely evaluates a statement
+  still fails loud, and the failure now names the statement.
 
 - **A `housecarl_records` `formids=` read is now held to the render bound on every form that reads a body, and
   says what those bodies cost.** `summary` and `aggregate` read a body per id on this lane exactly as `fields`,
--- a/src/housecarl-mcp-tests/DecompileShortCircuitTests.cs
+++ b/src/housecarl-mcp-tests/DecompileShortCircuitTests.cs
@@ -1,0 +1,119 @@
+using Mutagen.Bethesda;
+using Mutagen.Bethesda.Pex;
+using Xunit;
+
+namespace HousecarlMcpTests;
+
+/// <summary>
+/// A short-circuit whose value is a call argument that is NOT the last one: the arguments after it
+/// evaluate between the arm and the call, so the false-path jump lands on the next argument's first
+/// instruction instead of on the consumer.
+///
+/// The Feed stream below is what the CK's own PapyrusCompiler emits for the source above it — checked
+/// by compiling that source, and by recompiling the decompiled output back to a byte-identical .pex.
+/// CI has no CK compiler, so the stream is pinned here by hand.
+/// </summary>
+[Trait("tier", "unit")]
+public class DecompileShortCircuitTests
+{
+    // Function Feed(bool isActive, int tier, string label)
+    //     Sink(isActive && tier >= Self.Threshold, "T " + label)
+    // EndFunction
+    static PexObjectFunction Feed()
+    {
+        var f = Fn(("Bool", "isActive"), ("Int", "tier"), ("String", "label"));
+        Local(f, "Int", "::temp0");
+        Local(f, "Bool", "::temp1");
+        Local(f, "String", "::temp2");
+        Local(f, "None", "::NoneVar");
+        Ins(f, InstructionOpcode.CAST, Id("::temp1"), Id("isActive"));
+        Ins(f, InstructionOpcode.JMPF, Id("::temp1"), Int(4));                       // -> 5, the STRCAT
+        Ins(f, InstructionOpcode.PROPGET, Id("Threshold"), Id("self"), Id("::temp0"));
+        Ins(f, InstructionOpcode.CMP_GTE, Id("::temp1"), Id("tier"), Id("::temp0"));
+        Ins(f, InstructionOpcode.CAST, Id("::temp1"), Id("::temp1"));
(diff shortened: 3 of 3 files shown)


src/housecarl-core/PapyrusDecompiler.cs
```diff
@@ -559,12 +560,18 @@ List<string> Structure(int lo, int hi, bool flushAtEnd, HashSet<int> exits, int
...
+                        // conditional jump (plain &&/||), a CAST hop into a different temp (nested
+                        // mixed-temp conditions), an ASSIGN/RETURN/call-arg (`x = a || b`). When the
+                        // value is a call argument that is NOT the last one, the arguments after it
+                        // evaluate first and the consuming call sits further on, so also accept an
+                        // arm that writes the temp whose value is read before being rewritten at or
+                        // after the join. Temps only: a real-var condition is always a plain if.
                         if (condName is not null && IsTemp(condName) && target < hi
-                            && ConsumesAsSource(_ins[target], condName))
+                            && (ConsumesAsSource(_ins[target], condName)
+                                || (WritesDestIn(i + 1, target, condName)
```
Avick3110, line 578: A stream that decompiles cleanly on `main` fails the whole function on this branch, so "nothing that worked before changes" does not hold. The old disjunct is unchanged, but the *new* one takes shapes that previously went down the plain-`if` path.

The new discriminator does not exclude the value-reused condition temp that the promotion block ~45 lines below (`ReadsBeforeWrite(i + 1, target, condName)` then `Materialized.Add(condName)`) exists to handle. When the guarded block reads the condition temp **and** rewrites it, and the temp is read again after the join, `WritesDestIn` is true and `ReadsBeforeWrite(target, hi)` is true, so this branch is taken before promotion gets a chance; `Structure` over the block then hits the already-consumed temp and the function falls back to a raw-bytecode comment block.

Stream (`hi` = 6):

```
0: CAST        ::temp0, flag
1: JMPF        ::temp0, +3                            -> 4
2: CALLMETHOD  Nudge, self, ::NoneVar, 1, ::temp0     ; reads the temp
3: CALLMETHOD  Ping,  self, ::temp0,   0              ; rewrites it -> WritesDestIn true
4: STRCAT      ::temp2, "T ", label                   ; join, does NOT consume ::temp0
5: CALLMETHOD  Sink,  self, ::NoneVar, 2, ::temp0, ::temp2   ; -> ReadsBeforeWrite true
```

Built as a `PexFile` with the same helpers as `DecompileShortCircuitTests` and run through `PapyrusDecompiler.DecompileFile`:

- `origin/main`: `FunctionsFailed = 0`, `OptimizerHints = 1`, source
  ```
  bool temp0 = flag
  if temp0
      Nudge(temp0)
      temp0 = Ping()
  endif
  Sink(temp0, "T " + label)
  ```
- this branch: `FunctionsFailed = 1`, `temp ::temp0 read with no pending value @2`, body emitted as the raw-bytecode block.

It fails loud rather than emitting wrong source, but it is the exact failure mode #393 was filed about, on a function that used to come back whole. The source shape is ordinary (`bool b = flag / if b / Nudge(b) / b = Ping() / endif / Sink(b, ...)`) under a compiler that keeps `b` in a temp slot, which is the premise of the promotion path; the sweep just did not hit one.

One separator that holds on both: also require `!ReadsBeforeWrite(i + 1, target, condName)` for the new disjunct, i.e. let promotion win whenever its own condition fires. In `Feed()` the arm writes `::temp1` at the `CMP_GTE` before any read of it, so that predicate is false there and the new test still passes; in the stream above it is true, and the plain-`if` reading is kept.

src/housecarl-core/PapyrusDecompiler.cs
```diff
@@ -1044,6 +1056,16 @@ bool ReadsBeforeWrite(int lo, int hi, string name)
             return false;
         }
 
+        /// <summary>Is <paramref name="name"/> written as a destination anywhere in [lo, hi)?
+        /// A short-circuit arm always writes its own temp; a plain if's condition temp is dead in
```
Avick3110: The PR notes say of this half: "makes no difference anywhere in that corpus ... It is kept because it is half of what a short-circuit arm *is* ..., not as a guard, and a test that pins it could not fail." A test does fail without it, and it is doing guard work.

Same builders as `DecompileShortCircuitTests`, `hi` = 5:

```
0: CAST        ::temp0, flag
1: JMPF        ::temp0, +2                            -> 3
2: CALLMETHOD  Nudge, self, ::NoneVar, 1, ::temp0     ; reads the temp, never writes it
3: STRCAT      ::temp2, "T ", label                   ; join
4: CALLMETHOD  Sink,  self, ::NoneVar, 2, ::temp0, ::temp2
```

- This branch as written: `FunctionsFailed = 0`, promotion fires, source `bool temp0 = flag / if temp0 / Nudge(temp0) / endif / Sink(temp0, "T " + label)`.
- With only the `WritesDestIn(...)` conjunct deleted from the match: `FunctionsFailed = 1`, `temp ::temp0 read with no pending value @2`, raw-bytecode block. The three new tests in this PR all still pass in that state, so nothing else in the suite catches it.

So it is what keeps the broadened match off the value-reused condition temp, and this docstring's "a plain if's condition temp is dead in the block it guards" is not true of that shape - the promotion branch in `Structure` exists precisely because the temp is read there. Worth a test, and worth correcting the note; see the other comment for the case where `WritesDestIn` is true and the guard therefore does not hold.

src/housecarl-core/PapyrusDecompiler.cs
Avick3110: Real, and folded as suggested. Your stream is now a test (`AGuardedBlockThatRewritesItsConditionTempStaysAPlainIf`): on e5ad18c4's parent it gives `FunctionsFailed = 1`, after the fix `FunctionsFailed = 0` and the source is the plain `if` with the promoted local, `Sink(temp0, "T " + label)` after the join.

The fix is the conjunct you named — the new disjunct also requires `!ReadsBeforeWrite(i + 1, target, condName)`, so promotion wins wherever its own condition fires. It reads as part of the definition rather than a patch: an arm that reads the temp before writing it is not producing the right operand, it is a guarded block over a reused condition temp.

I also re-ran the corpus sweep with and without the new conjunct over the larger of the two trees (14,424 unique `.pex`): totals and every per-script source hash identical, so the narrowing costs nothing that the broadening bought. The PR body now says so, and its "nothing that worked before changes" sentence is corrected to name what makes it true.

src/housecarl-core/PapyrusDecompiler.cs
Avick3110: Right on both counts, and folded in a41a6e79 — the docstring's second sentence is gone, and there is now a test. One correction to the mechanics, with the measurement.

Your stream stops separating `WritesDestIn` once the other comment's fix lands. There the block reads `::temp0` at instruction 2, so `!ReadsBeforeWrite(i + 1, target, condName)` is already false and the new disjunct cannot fire whatever `WritesDestIn` says. Ran it: with the fix in and `WritesDestIn(...)` deleted, that stream gives `FunctionsFailed = 0`, `OptimizerHints = 1` and the promoted source you quoted. The full suite and `ci-all` also pass in that state, so nothing pinned it.

What still separates it is the same source with the block's read taken away — `bool b = flag / if b / Nudge() / endif / Sink(b, "T " + label)`, `hi` = 5:

```
0: CAST        ::temp0, flag
1: JMPF        ::temp0, +2                            -> 3
2: CALLMETHOD  Nudge, self, ::NoneVar, 0              ; neither reads nor writes the temp
3: STRCAT      ::temp2, "T ", label
4: CALLMETHOD  Sink,  self, ::NoneVar, 2, ::temp0, ::temp2
```

That shape fails either way — the condition already spent the temp — but the message differs:

- with `WritesDestIn`: `temp ::temp0 read with no pending value @4`, same as `origin/main`
- without it: `short-circuit arm @2..3 evaluates a statement, not just a value (first: Nudge())`

The second blames a short-circuit arm on a block that is not one, so the conjunct is keeping a failure honest rather than only restating the definition. `AGuardedBlockThatNeverTouchesItsConditionTempIsNotReportedAsAnArm` pins it: the failure names the temp and does not say "short-circuit". The PR note that called it untestable is corrected.

src/housecarl-core/PapyrusDecompiler.cs
```diff
@@ -559,12 +560,21 @@ List<string> Structure(int lo, int hi, bool flushAtEnd, HashSet<int> exits, int
...
+                        // arm that writes the temp whose value is read before being rewritten at or
+                        // after the join. An arm that READS the temp first is not an arm at all: it
+                        // is a guarded block over a reused condition temp, which the promotion path
+                        // below owns. Temps only: a real-var condition is always a plain if.
                         if (condName is not null && IsTemp(condName) && target < hi
-                            && ConsumesAsSource(_ins[target], condName))
+                            && (ConsumesAsSource(_ins[target], condName)
+                                || (WritesDestIn(i + 1, target, condName)
+                                    && !ReadsBeforeWrite(i + 1, target, condName)
+                                    && ReadsBeforeWrite(target, hi, condName))))
```
Avick3110, line 581: **The broadened match steals a plain `if`/`else`, and the function then fails whole.**

When the then-block assigns the condition's temp slot without reading it, disjunct 2 matches — but the arm region `[i+1, target)` still contains the then-block's trailing `JMP` to the join, and `Structure` is called for the arm with an empty `exits` set, so that JMP matches none of the JMP cases and throws.

Stream (run through both branches):

```
0 CAST       ::temp0, a
1 JMPF       ::temp0, +3        -> 4
2 CALLMETHOD X, self, ::temp0, 0
3 JMP        +3                 -> 6
4 CALLMETHOD DoSomething, self, ::NoneVar, 0
5 NOP
6 CALLMETHOD Sink, self, ::NoneVar, 1, ::temp0
```

i.e. `if a / b = X() / else / DoSomething() / endif / Sink(b)` with `b` sharing the condition's temp slot. `WritesDestIn(2,4)` true, `!ReadsBeforeWrite(2,4)` true, `ReadsBeforeWrite(4,hi)` true → disjunct 2 matches → `Structure(2,4)` hits `JMP @3 -> 6`, which is not `i+1`, not in the arm's empty `exits`, and not `_ins.Count`.

- on `main`: `if a / temp0 = X() / else / DoSomething() / endif / Sink(temp0)`
- on this branch: `unmatched JMP @3 -> 6 (unknown flow pattern)` — the whole function drops to the raw-bytecode comment block.

The same hole covers a `while` whose body writes the condition temp: the arm swallows the backward JMP. That is the case the promotion path below carries an explicit `!isWhile` guard for, and this disjunct runs before `isWhile` is even computed.

A real short-circuit arm is straight-line, so declining on the if-else/while marker would close both:

```suggestion
                            && (ConsumesAsSource(_ins[target], condName)
                                || (WritesDestIn(i + 1, target, condName)
                                    && (target - 1 <= i || _ins[target - 1].OpCode != InstructionOpcode.JMP)
                                    && !ReadsBeforeWrite(i + 1, target, condName)
                                    && ReadsBeforeWrite(target, hi, condName))))
```

src/housecarl-core/PapyrusDecompiler.cs
```diff
@@ -559,12 +560,21 @@ List<string> Structure(int lo, int hi, bool flushAtEnd, HashSet<int> exits, int
...
+                        // join side still reads it. The jump usually lands ON that consumer — another
+                        // conditional jump (plain &&/||), a CAST hop into a different temp (nested
+                        // mixed-temp conditions), an ASSIGN/RETURN/call-arg (`x = a || b`). When the
+                        // value is a call argument that is NOT the last one, the arguments after it
+                        // evaluate first and the consuming call sits further on, so also accept an
+                        // arm that writes the temp whose value is read before being rewritten at or
+                        // after the join. An arm that READS the temp first is not an arm at all: it
+                        // is a guarded block over a reused condition temp, which the promotion path
+                        // below owns. Temps only: a real-var condition is always a plain if.
                         if (condName is not null && IsTemp(condName) && target < hi
```
Avick3110: **A condition temp already in `Materialized` is not excluded, and it can never satisfy the arm.**

Once a temp is promoted, writes to it are real assignments — statements — so an arm over a materialized temp always trips the `sub.Count > 0` failure below. The match reaches that state instead of declining.

Stream (run through both branches):

```
0 CALLMETHOD X,        self, ::temp0, 0
1 CALLMETHOD DoThing,  self, ::NoneVar, 0
2 JMPF       ::temp0, +3           -> 5
3 CALLMETHOD Y,        self, ::temp0, 0
4 NOP
5 NOP
6 CALLMETHOD Sink,     self, ::NoneVar, 1, ::temp0
```

The flush at instruction 1 materializes `::temp0` (its value is read downstream), so it is a named local from there on. At the JMPF, `WritesDestIn(3,5)` true, `!ReadsBeforeWrite(3,5)` true, `ReadsBeforeWrite(5,hi)` true → disjunct 2 matches.

- on `main`: `bool temp0 = X() / DoThing() / if temp0 / temp0 = Y() / endif / Sink(temp0)`
- on this branch: `short-circuit arm @3..5 evaluates a statement, not just a value (first: temp0 = Y())` — function failed.

The comment above says an arm that reads the temp first "is a guarded block over a reused condition temp, which the promotion path below owns". A temp the promotion path already took is the same case one step later, and `IsTemp` still returns true for it. Adding the exclusion to the `IsTemp` test covers both disjuncts:

```suggestion
                        if (condName is not null && IsTemp(condName) && !Materialized.Contains(condName) && target < hi
```

src/housecarl-core/PapyrusDecompiler.cs
Avick3110: Real, and folded as suggested in 3d21edfd. Your stream is now a test (`AnIfElseWhoseThenBlockWritesTheConditionTempStaysAnIfElse`): on a41a6e79 it gives `FunctionsFailed = 1` with exactly `unmatched JMP @3 -> 6 (unknown flow pattern)`; with the fix it gives `FunctionsFailed = 0` and `if a / temp0 = X() / else / DoSomething() / endif / Sink(temp0)`. I also ran the test against `origin/main`'s `PapyrusDecompiler.cs` with the rest of the branch in place: it passes there, so this is a worked-before/broken-now regression, not a shape the file never handled.

The fix is the conjunct you wrote, on the new disjunct only.

One correction on the `while` half. That case is not a regression — it fails on `main` too, and the fix restores `main`'s message rather than a whole function.

```
0 CAST       ::temp0, a
1 JMPF       ::temp0, +3        -> 4
2 CALLMETHOD X, self, ::temp0, 0
3 JMP        -3                 -> 0
4 NOP
5 CALLMETHOD Sink, self, ::NoneVar, 1, ::temp0
```

- `origin/main`: `temp ::temp0 read with no pending value @5`
- a41a6e79: `unmatched JMP @3 -> 0 (unknown flow pattern)` — the arm swallows the back-jump, as you said
- with the fix: `temp ::temp0 read with no pending value @5`, same as `main`

The reason it cannot come back whole either way is that the loop re-evaluates its condition, so the body's write to the temp is dead across the back edge and the post-loop read has nothing pending — the value-out-of-a-loop shape the `!isWhile` guard on the promotion path already declines. So there is no pass/fail test to pin for it, only the message, and it is the same message `main` gives. Your `target - 1` conjunct covers it because the marker it tests is the same one `isWhile` reads.

Corpus, to check the narrowing costs nothing: swept both trees (18,562 `.pex`, 18,408 unique after de-duplication, 57,561 functions) with a41a6e79 and with 3d21edfd. Failed functions, optimizer hints, files flagged and every per-script source SHA-256 are identical.

src/housecarl-core/PapyrusDecompiler.cs
Avick3110: Real, and folded as suggested in 0da6dd23 — the exclusion is on the `IsTemp` test, so it covers both disjuncts. Your stream is now a test (`AConditionTempAlreadyPromotedToALocalIsNotAnArm`): on 3d21edfd it gives `FunctionsFailed = 1` with exactly `short-circuit arm @3..5 evaluates a statement, not just a value (first: temp0 = Y())`; with the fix it gives `FunctionsFailed = 0` and `if temp0 / temp0 = Y() / endif / Sink(temp0)`. Run against `origin/main`'s `PapyrusDecompiler.cs` it passes, so this too is worked-before/broken-now.

On putting it on both disjuncts rather than only the new one, which is a change to behaviour that predates this PR: a materialized temp cannot satisfy the OLD disjunct either. `Consume` hands back a bare `EIdent` for it, so the match builds `temp0 && temp0` and then calls `SetPending` on a name that `Materialized` says is a real local — the one place in the file that can put a pending on a promoted temp. Where the arm is empty that renders as a redundant `x && x`; where the arm writes the temp it is your failure one instruction earlier. Neither is something to keep, and the exclusion is one predicate for both.

Corpus, since that half touches pre-existing behaviour: swept both trees (18,562 `.pex`, 18,408 unique, 57,561 functions) with 3d21edfd and with 0da6dd23. Failed functions, optimizer hints, files flagged and every per-script source SHA-256 are identical, so no script in either tree was reaching the old disjunct on a promoted temp.

Full suite 1774 passed, `ci-all` 127/127 with decompile-guard's golden unchanged.

PR #626, opened 2026-09-07
Say which references= target each batch-rendered row hit

### LINKED ISSUE #576: [GAP] the shared batch render carries no per-row matches column, so form=rows and scoped fields cannot say which references= target each row hit
## The gap

`housecarl_records references=["A","B"]` with `form='rows'`, a scoped `form='fields'`, or (since PR #574) a quantified `fields` projection routes through `JsonWire.RenderBatch` (`RecordsTools.cs`, the body-lane gate), which carries no `matches` column. The CrossQuery render does carry it. So the same query answers "which target did this row hit" under one form and stays silent under the others; nothing in the response says the column was dropped.

Found during Aaron's review of PR #574 (comment 3940880520). The PR joins an existing gap rather than opening one: the body lane already dropped `matches` on `main` before it, with no fold token involved.

## What ships

A per-row `matches` on the shared batch render, landing on every form that uses it at once. Not a per-form patch.

## Not this issue

The artifact lane already carries `matches?`; nothing to change there.

Avick3110:
A `references=["A","B"]` scan records which of the targets each match hit, and the scan render shows it as a per-row `matches` column. The forms that read a body per row — `project.form='rows'`, `'everything'`, a `fields` read under a `plugins=` scope with a named `source=`, and a `fields` projection with a quantified path — do not render through the scan render at all; they hand their bodies to the shared batch render, which had no such column. So the same two-target lookup could be un-merged back into per-target answers under one form and stayed silent under another, with nothing in the response saying the column had been dropped. The fix is one per-row `matches` on the shared batch render rather than a patch per form: `Wire.RenderBatch` and `JsonWire.RenderBatch` take the list parallel to their outcomes, and both body lanes in `RecordsTools.cs` (the in-order scan and the off-order file scan) pass the scan outcome's `MatchedTargets` down. The text line and the json key are the scan render's own, so the two transports cannot drift. A single-target `references=` still passes null and adds no column.

One thing beyond the issue's stated scope: the issue says the artifact lane already carries `matches?`, which is true of `Artifacts.WriteCrossQuery` but not of `Artifacts.WriteBatch` — the artifact the body lane spills to. Under `to_file=` the rows *are* the answer, so leaving that one out would have moved the gap rather than closed it; `WriteBatch` now takes the same list and names `matches?` in its row schema.

Tests are eight in `RecordsScanProjectionTests`, beside the existing multi-target reverse-lookup test: the rows form on text and on json, the quantified `fields` projection, the scoped `fields` read, `everything`, a spilled artifact's row plus its schema, the off-order file scan's rows, and the single-target negative. Seven of them fail at the branch point and pass after; the single-target one is an absence guard — it passes at the branch point too, and holds the column quiet once the other seven make it appear. The off-order one drives the second body lane, which the other seven never reach: `BulkRecordsWorld` gained a switched-off mod carrying the same two-target weapon pair out of the load order, so a `source=` naming that plugin takes the file-scan lane, and reverting `offMatches` in `RecordsTools.cs` fails that test alone.

Closes #576

### DIFF
--- a/src/housecarl-mcp/RecordsTools.cs
+++ b/src/housecarl-mcp/RecordsTools.cs
@@ -1518,13 +1518,16 @@ public static class RecordsTools
                 // These bodies were selected by a scan, not a formids list, so the batch notice's selection
                 // clause must name limit= — the knob that actually windows this response.
                 var evLevers = formLevers.OnScanSelection();
+                // These rows were selected by the scan, so they carry its multi-target references= un-merge too —
+                // one row per key, in key order, which is what makes the list parallel to the bodies.
+                var evMatches = outcome.MatchedTargets;
                 string RenderEv(SpillState? sp, out bool trunc) => json
-                    ? JsonWire.RenderBatch(bodies, max_chars, sp, out trunc, envelope, evLevers, (bodies.Count, bodyClock.ElapsedMilliseconds))
-                    : Wire.RenderBatch(bodies, max_chars, sp, out trunc, evLevers, (bodies.Count, bodyClock.ElapsedMilliseconds), headerLine);
+                    ? JsonWire.RenderBatch(bodies, max_chars, sp, out trunc, envelope, evLevers, (bodies.Count, bodyClock.ElapsedMilliseconds), evMatches)
+                    : Wire.RenderBatch(bodies, max_chars, sp, out trunc, evLevers, (bodies.Count, bodyClock.ElapsedMilliseconds), headerLine, evMatches);
                 SpillState? evSpill = null;
                 if (wantFile)
                 {
-                    var (s, aerr) = Artifacts.WriteBatch(bodies, toFile!, "to_file", Echo(), evLevers);
+                    var (s, aerr) = Artifacts.WriteBatch(bodies, toFile!, "to_file", Echo(), evLevers, matches: evMatches);
                     if (aerr is not null) return json ? JsonWire.RenderError(aerr, bodyEpoch) : "error: " + aerr;
                     evSpill = SpillState.Spilled(s!, manifestOnly: true);
                 }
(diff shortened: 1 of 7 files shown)


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1738,22 +1741,24 @@ List<KeyValuePair<string, string>> Echo()
                 headerLine += $"\n{outcome.Total} match(es); bodies for the {keys.Count}-row window below";
                 // Selected by the off-order file scan, so the remedy vocabulary matches the body lane above.
                 var offLevers = formLevers.OnScanSelection();
+                // Same rule as the in-order body lane: the file scan's rows carry its references= un-merge too.
+                var offMatches = outcome.MatchedTargets;
```
Avick3110, line 1745: The off-order file-scan half of this change is not pinned by any test. All seven new tests drive the in-order scan lane: they select with `types=` or `plugins=` over the active order, and the one that names a pole (`AScopedFieldsReadCarriesTheMatchesColumn`) names `W.MasterName`, which is in the load order, so it lands on `scopePlusPole` and the in-order body lane, not here. `OffOrderScan` is only reached from the `!probe.InOrder` branch, which nothing in the suite exercises with `references=`. Delete `offMatches` from the three call sites below and `dotnet test` still passes.

The lane is live, not dead code, so the gap is real rather than moot: `OffOrderScan` refuses `conflicts_only`, `info_order`, `walk` and `dense` but says nothing about `references=`, it passes `refFks` into `OffOrderQuery`, and that scan builds its own `matched` list parallel to `keys` for a two-or-more-target lookup exactly as the in-order scan does. One test — `source=` a plugin file outside the load order, `references=` two targets, `project.form='rows'`, `format='json'` — would pin it the way the six in-order ones pin their half.

Avick3110: Real, and reproduced exactly as you describe: setting `offMatches` to null here left the whole suite green. Folded in 077f59ff.

`BulkRecordsWorld` had no off-order file at all, so it gained one: a switched-off mod `BulkOffMod` holding `HcBulkOff.esp`, with the same shape the active order carries — `OffW1` on `KwA` alone, `OffW2` on both keywords. The mod is `-` in `modlist.txt` and its plugin is in neither profile list, so it is on disk and out of the order, and no active-order count moves (the WEAP scan is still 3).

The test is `TheOffOrderFileScansRowsCarryTheMatchesColumn`: `source=` that plugin, `references=` both keywords, `project.form='rows'`, `format='json'`, asserting `matches` on both rows — `KwA, KwB` on `OffW2` and `KwA` alone on `OffW1`, so it reads the un-merge and not just the column's presence. Reverting `offMatches` to null fails that test and no other.

The PR body's tests paragraph is corrected too: eight tests, seven failing at the branch point, with the single-target one named as the absence guard it is.

PR #627, opened 2026-09-08
Refuse an import directory that declares an RVA but no size

### LINKED ISSUE #416: [Bug] ReadImportNames: a delay-load directory with Size==0 but non-zero RVA silently drops its imports and reports a successful walk
Migrated from the retired `dev/BACKLOG.md` (advisor sweep, Aaron-go 2026-08-22); re-verified at `main` @ `a783b2f` — verbatim.

`SksePluginReader.cs:402`: `if (dirSize == 0 || dirRva == 0) return true;` — either-zero reads as "directory genuinely absent", so a PE declaring a delay-import RVA with zero Size drops its delay imports and still reports success (non-null Imports). Every other failure mode in that walk returns `false` → UNKNOWN; this is the one silent partial left in a method that tightened everywhere else (an under-declared non-zero Size already truncates into a loud `false` via the `:409` clamp — Size==0 is the only silent case).

**Deliberately unfixed pending a specimen:** the loader walks to the null terminator and treats Size as informational, so refusing on `rva != 0 && size == 0` might reject legal images — and the `297/297 walked · 0 UNKNOWN` evidence counts walk success, saying nothing about whether delay imports were found. The fix fork is undecidable without a real PE exhibiting the shape (the #393/#397 latent class).

**Routing:** row 8's W5 routing-call list, deliberately unmilestoned. Promotion condition: a specimen surfaces.

Avick3110:
`ReadImportNames` walks the PE import and delay-import directories bounded by the Size the header declares for each — that bound is the only thing stopping an unterminated table from reading on into adjacent `.rdata` and inventing descriptors. It treated `Size == 0` as "directory absent" and returned success, so a directory with a non-zero RVA and a zero Size had its whole table skipped while the walk still reported a clean result. The remaining names rendered as a complete `imports (N): …`, and `DebugCrtBlocker` over that list returned a clean bill of health for a DLL whose dropped delay imports could have been the debug CRT — the one silent partial in a method that answers UNKNOWN everywhere else. Now a declared RVA with no Size fails the walk like every other parse failure there, so the DLL reads `imports: UNKNOWN`. `RVA == 0` is unchanged: that is genuine absence, and refusing it would turn every DLL that delay-imports nothing into an UNKNOWN.

Closes #416.

## The specimen the issue was waiting on

The issue deferred the fix because refusing on `rva != 0 && size == 0` "might reject legal images" — the loader walks to the null terminator and treats Size as informational — and said the `297/297 walked · 0 UNKNOWN` evidence counted walk success, not whether delay imports were seen. That is measurable rather than undecidable, so I measured it on this machine: 3,598 PE images in `System32` and the 323 SKSE plugin DLLs in the ARR 2.0 modlist, 1,748 of which declare a delay-import directory at all (1,745 + 3). **Zero** carry the `rva != 0 && size == 0` shape, in either directory. No legal image in the population houseCARL reads is rejected by this. (Incidentally: only 3 of the 323 SKSE plugins declare a delay-import directory, which is why the old walk-success evidence could not have caught this.)

Because no real image has the shape, the tests build the PE bytes instead: a minimal x64 image with one `.rdata` section carrying a real import descriptor table, a real delay-load descriptor table, and their two name strings, so only what the header declares about them varies. Five cases — the well-formed delay directory (which imports `vcruntime140d.dll`, i.e. exactly what the silent drop was hiding from the debug-CRT check), the same directory with Size zeroed, the same shape on the ordinary import directory, an absent delay directory, and an image importing nothing. The two zero-Size cases fail on `main` with `["kernel32.dll"]` instead of null; the other three pass either way and are there to pin the refusal's edge — they fail against the naive `dirSize == 0 || dirRva == 0 → false`.

## Not done here

There is a third option the issue did not name: on `Size == 0`, walk the table to its null terminator bounded by the section remainder instead of refusing. That would answer instead of degrade, and the existing terminator and `MaxDescriptors` checks would still make a genuinely unterminated table loud. I did not take it — it drops the header bound this method deliberately keeps, and with zero specimens in ~3,900 images there is nothing to justify the extra surface. Worth an issue if a specimen ever does turn up.

## Run

`dotnet build housecarl.sln -c Release` clean; `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` 1794/1794; `ci-all` 127/127.

### DIFF
--- a/src/housecarl-core/SksePluginReader.cs
+++ b/src/housecarl-core/SksePluginReader.cs
@@ -459,12 +459,16 @@ public static class SksePluginReader
     /// design: the SKSE loader itself resolves symbols by exact unmangled name string, so a name lookup is all we need.
     /// Returns an EMPTY map for a DLL with genuinely no export table (→ classify NotSkse), but <c>null</c> when the directory
     /// is present yet CORRUPT (a parse failure — the caller classifies Unreadable, never silently as a bundled
-    /// dependency). Never throws.</summary>
+    /// dependency). ABSENCE is a zero RVA and nothing else: a declared RVA is read whatever the header's Size says,
+    /// since this walk is driven by the directory's own counts (capped by <c>MaxExports</c>) and never by that Size.
+    /// Never throws.</summary>
     static Dictionary<string, int>? ReadExportRvas(PEReader pe)
     {
         var byName = new Dictionary<string, int>(StringComparer.Ordinal);
         var dir = pe.PEHeaders.PEHeader!.ExportTableDirectory;
-        if (dir.Size == 0 || dir.RelativeVirtualAddress == 0) return byName;   // no export table → empty (genuinely no exports)
+        // A zero Size beside a declared RVA is NOT "no exports": the table is there, and reading it as empty would
+        // classify a real plugin NotSkse — "a bundled dependency DLL, not a plugin" — and skip its version checks.
+        if (dir.RelativeVirtualAddress == 0) return byName;                    // no export table → empty (genuinely no exports)
         try
         {
             var ed = pe.GetSectionData(dir.RelativeVirtualAddress).GetReader();
--- a/src/housecarl-core/SksePluginReader.cs
+++ b/src/housecarl-core/SksePluginReader.cs
@@ -388,7 +388,12 @@ public static class SksePluginReader
 
         bool Walk(int dirRva, int dirSize, int stride, int nameOff, bool delay)
         {
-            if (dirSize == 0 || dirRva == 0) return true;              // directory genuinely absent → nothing to add
+            if (dirRva == 0) return true;                              // directory genuinely absent → nothing to add
+            // A declared RVA with a zero Size is NOT absence: there is a table there, and this walk's only bound on it
+            // is that Size (see the limit below). Returning true would walk nothing and still report success — a whole
+            // directory of imports dropped behind a complete-looking "imports (N): …", the same silent partial the
+            // unresolvable-name case below refuses. Answer UNKNOWN instead, like every other failure here.
+            if (dirSize == 0) return false;
             try
             {
                 var block = pe.GetSectionData(dirRva);
--- a/src/housecarl-core/SksePluginReader.cs
+++ b/src/housecarl-core/SksePluginReader.cs
@@ -372,7 +372,8 @@ public static class SksePluginReader
     ///
     /// Both directories are arrays of fixed-size descriptors terminated by an all-zero entry, each carrying an RVA to the
     /// imported DLL's ASCII name. Delay-load descriptors predate the RVA convention: bit0 of their Attributes is
-    /// <c>RvaBased</c>, and a (long-obsolete) VA-based table is SKIPPED rather than misread as an RVA.</summary>
+    /// <c>RvaBased</c>, and a (long-obsolete) VA-based table carries absolute addresses, resolved here the way the
+    /// loader does — VA minus the image base — and refused when that lands outside the image.</summary>
     static List<string>? ReadImportNames(PEReader pe)
     {
         var names = new List<string>();
@@ -409,10 +410,19 @@ public static class SksePluginReader
                     uint attributes = delay ? rd.ReadUInt32() : 0;     // delay-load: Attributes precedes the name RVA
                     rd.Offset = i * stride + nameOff;
                     int nameRva = rd.ReadInt32();
-                    // The all-zero descriptor terminates the array. A delay-load table whose bit0 (RvaBased) is clear is
-                    // VA-based (pre-VS2015); its "RVA" is an absolute address we must NOT resolve — skip, don't guess.
+                    // The all-zero descriptor terminates the array.
                     if (nameRva == 0) return true;
-                    if (delay && (attributes & 1) == 0) continue;
+                    // A delay-load table whose bit0 (RvaBased) is clear is VA-based (pre-VS2015): the field is an
+                    // absolute address, so subtract the image base to get the RVA, which is what the loader does.
+                    // Skipping it instead would drop the name behind a complete-looking list — the same silent partial
+                    // the zero Size above refuses. A VA outside the image is unresolvable, so it fails the walk; a
+                    // 32-bit field cannot express a 64-bit image's addresses, which is why such a table is x86-only.
+                    if (delay && (attributes & 1) == 0)
+                    {
+                        long fromBase = (uint)nameRva - (long)hdr.ImageBase;
+                        if (fromBase is <= 0 or > int.MaxValue) return false;
+                        nameRva = (int)fromBase;
+                    }
                     // An unresolvable name is CORRUPTION, and skipping it would hand back a short list that renders as a
                     // complete "imports (N): …" — a silent partial answer, which is worse here than no answer: if the
                     // entry we dropped were vcruntime140d.dll, the Debug-CRT check would report a clean bill of health.
--- a/src/housecarl-core/SksePluginReader.cs
+++ b/src/housecarl-core/SksePluginReader.cs
@@ -365,10 +365,12 @@ public static class SksePluginReader
     }
 
     /// <summary>Walk the PE IMPORT + DELAY-LOAD directories and return the imported DLL names, lower-cased and
-    /// deduplicated, in image order. Same rule as <see cref="ReadExportRvas"/>: an ABSENT directory yields an empty
-    /// list (a real, if odd, "imports nothing"), while a PRESENT-but-CORRUPT one yields <c>null</c> — a parse failure
-    /// the caller renders as UNKNOWN, since a corrupt import table must not read as a clean bill of health.
-    /// Never throws.
+    /// deduplicated, in image order. Absence is a zero RVA and nothing else, as in <see cref="ReadExportRvas"/>: an
+    /// ABSENT directory yields an empty list (a real, if odd, "imports nothing"), while a PRESENT-but-CORRUPT one
+    /// yields <c>null</c> — a parse failure the caller renders as UNKNOWN, since a corrupt import table must not read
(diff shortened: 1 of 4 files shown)


src/housecarl-core/SksePluginReader.cs
```diff
@@ -388,7 +388,12 @@ static string AsciiZ(ReadOnlySpan<byte> buf, int off, int max)
 
         bool Walk(int dirRva, int dirSize, int stride, int nameOff, bool delay)
         {
-            if (dirSize == 0 || dirRva == 0) return true;              // directory genuinely absent → nothing to add
+            if (dirRva == 0) return true;                              // directory genuinely absent → nothing to add
+            // A declared RVA with a zero Size is NOT absence: there is a table there, and this walk's only bound on it
+            // is that Size (see the limit below). Returning true would walk nothing and still report success — a whole
+            // directory of imports dropped behind a complete-looking "imports (N): …", the same silent partial the
+            // unresolvable-name case below refuses. Answer UNKNOWN instead, like every other failure here.
+            if (dirSize == 0) return false;
```
Avick3110, line 399: The sibling silent partial this leaves open, two lines below the new guard: `if (delay && (attributes & 1) == 0) continue;`. A delay-load descriptor whose Attributes bit0 (RvaBased) is clear is skipped and the walk still returns true — the same failure this PR closes, in the same method.

I reproduced it with your own test image, one byte changed (Attributes 1 -> 0 at `At(DelayRva) + 0x00`), everything else identical:

```
VA-based  -> Imports=kernel32.dll                    DebugCrt=(empty)  Blocker=<none>
RVA-based -> Imports=kernel32.dll,vcruntime140d.dll                    Blocker=a DEBUG build - it imports vcruntime140d.dll ...
```

So `Imports` comes back non-null and renders as a complete `imports (1): kernel32.dll`, `DebugCrtImportsOf` is empty, and `DebugCrtBlocker` returns null — a clean bill of health for an image that delay-imports the debug CRT. That is the #416 failure verbatim, just via a different branch. It also makes the new comment's "Answer UNKNOWN instead, like every other failure here" not quite true of the branch it sits above.

The VA is not actually unresolvable, either: VA - ImageBase = RVA is what the loader does, so this one could answer rather than drop.

Evidence for how much it matters, on the same standard this PR set: I scanned 5,183 PE images in System32 and all 323 SKSE plugin DLLs in the ARR 2.0 modlist. 2,211 declare a delay-import directory; every one of them is RVA-based, none VA-based. (Same scan reproduces your other numbers: 0 images of the `rva != 0 && size == 0` shape in either directory, and 3 of the 323 SKSE DLLs declaring a delay directory.) So there is no specimen here either — not a merge blocker, but it is the last silent partial left in this method, and worth an issue if you would rather not widen the change.

Avick3110: Real, and folded in 23381f43 rather than filed, since it is the same silent partial #416 is about, two lines over, and answering it turned out to be five lines.

The walk now resolves the VA instead of skipping the descriptor:

```csharp
if (delay && (attributes & 1) == 0)
{
    long fromBase = (uint)nameRva - (long)hdr.ImageBase;
    if (fromBase is <= 0 or > int.MaxValue) return false;
    nameRva = (int)fromBase;
}
```

So the name is read where the address resolves, and where it does not the walk fails like every other parse failure there — no branch left that drops a name and returns true. The doc comment and the inline comment that said VA-based tables are skipped now say what happens instead.

Two tests off the same image builder, which grew two optional parameters (`delayAttributes`, `imageBase`) so the delay table is written the way a linker would: bit0 set writes the RVA, bit0 clear writes `imageBase + rva`. Both fail on the unfixed reader with exactly your `["kernel32.dll"]`:

- `VaBasedDelayDescriptorIsResolvedAgainstTheImageBase` — ImageBase `0x400000`, VA `0x401100`, resolves to `vcruntime140d.dll`, and `DebugCrtImportsOf` sees it.
- `VaBasedDelayDescriptorOutsideTheImageAnswersUnknown` — ImageBase `0x180000000`, so the 32-bit name field cannot hold the address, it lands outside the image, and `Imports` is null.

One correction to the repro, which does not change the finding: the PR's image leaves ImageBase at 0, so there VA and RVA are the same number and the subtraction is a no-op — flipping the byte reproduces the drop, but not the resolution. Giving the image a real base is what separates the two cases above, and it is also why the second one exists: a VA-based table is a 32-bit-only shape, because the descriptor's name field is a DWORD and cannot express a 64-bit image's addresses. So on the x64 images houseCARL actually reads, the honest answer to a VA-based delay table is UNKNOWN, not a resolved name — which the fix now gives, instead of a clean bill of health.

Your scan numbers agree with mine and there is still no specimen; the change is worth it for the branch it closes, not for a DLL either of us has seen.

Build clean; `tier!=bridge` 1796/1796; `ci-all` all pass.

Avick3110:
Review of the diff. The change itself is correct: the terminator check runs before the VA branch (so an all-zero descriptor still ends the walk in both forms), `(uint)nameRva - (long)hdr.ImageBase` cannot overflow and the `<= 0 or > int.MaxValue` guard catches both the wrapped-32-bit case and a base above the address, and the seven new tests pass on the branch and pin the edges the naive form would miss. Build clean, the new file 7/7, `skse-peek-guard` ALL PASS.

I also re-ran the population measurement independently, widened: 11,986 PE images (System32 + SysWOW64 + the whole `ARR 2.0/mods` tree), reading the header directly.

```
import dir declared (rva!=0): 10173, of which size==0: 0
export dir declared (rva!=0):  7625, of which size==0: 0
delay  dir declared (rva!=0):  3519, of which size==0: 0
delay descriptors VA-based (attrs bit0 clear): 0 (x64: 0)
```

That confirms the PR body's claim and extends it to the VA-based change, which the body does not measure — no real image on this machine is turned into an UNKNOWN by either half.

Two findings, both just below the diff.

---

**1. `src/housecarl-core/SksePluginReader.cs:467` — `ReadExportRvas` still reads a declared-RVA/zero-Size export directory as "no exports".**

```csharp
if (dir.Size == 0 || dir.RelativeVirtualAddress == 0) return byName;   // no export table → empty
```

Same shape this PR just refused on the import side, and here the wrong answer is worse: a DLL with `ExportTableDirectory.RelativeVirtualAddress != 0` and `Size == 0` gets an empty map, so `ReadStream` takes the `!hasVersion && !hasQuery && !hasLoad` branch and classifies it `NotSkse` — "no SKSE export (SKSEPlugin_Version/Query/Load) — a bundled dependency DLL, not a plugin". A real plugin is reported to the modder as a stray dependency, and its version/runtime checks never run. Not a short list: a definitive wrong classification.

And unlike the import walk, `dir.Size` is not load-bearing here at all — the walk is driven by `numFuncs`/`numNames` read out of the directory itself and capped by `MaxExports`, never by the header Size. So the `dir.Size == 0` disjunct buys no bound; it only manufactures the false "no exports". Dropping it (keeping the `RelativeVirtualAddress == 0` half, which is genuine absence) would make the two directories agree, with no new refusal surface.

Hardening, not a live break — 0/7,625 declared export directories on this machine carry the shape.

**2. `src/housecarl-core/SksePluginReader.cs:368` — the summary now claims a parity that no longer holds.**

> Same rule as `ReadExportRvas`: an ABSENT directory yields an empty list …

After this change the two disagree on what absent means: `ReadImportNames` refuses `rva != 0 && size == 0` (null → UNKNOWN), `ReadExportRvas` accepts it (empty → NotSkse). The next reader takes the cross-reference at face value and assumes the export path already carries the refusal, which is exactly how finding 1 stays invisible. Either state the divergence or fix the export path and keep the sentence true.

Avick3110:
Both findings folded; nothing refuted. Two commits on top of the branch.

**1. `ReadExportRvas` on a declared-RVA/zero-Size export directory — folded.** `298edc68`

Confirmed by reading the walk: the loop is driven by `numFuncs`/`numNames` read out of the directory itself and capped by `MaxExports`, and `dir.Size` is never consulted again after that guard. So the disjunct bought no bound and only manufactured the wrong answer — empty map → `!hasVersion && !hasQuery && !hasLoad` → `NotSkse`, "a bundled dependency DLL, not a plugin", with the version and runtime checks skipped. Dropped `dir.Size == 0 ||`; absence is now a zero RVA in both walks. A declared RVA that maps to no section still lands in the `catch` → `null` → `Unreadable`, which is what the same shape already did with a non-zero Size, so that path is unchanged.

Test: `ExportDirectoryWithZeroSizeIsStillRead` on a synthetic image whose export directory exports `SKSEPlugin_Query` — expects `LegacyQuery`, and on `main` it is `NotSkse`. It is the only test in the file that fails without the fix. Two more pin the edges: `ExportDirectoryWithItsSizeDeclaredIsRead` (the control — the header field is the only thing that varies between the two) and `AbsentExportDirectoryIsNotSkse` (the other half of the fork, so dropping the RVA half too would be caught).

The image builder is the one already in the file, extended with an export directory that is always written; only what the header declares about it varies. That made the file's name wrong, so it moved: `SkseImportDirectoryBoundTests.cs` → `SkseDirectoryReadTests.cs`, with the class doc rewritten to cover both walks. Git tracks it as a rename; the seven tests you read are unchanged. Also a changelog line, since a plugin named as a bundled dependency is something a modder would see.

**2. The `ReadImportNames` summary claiming a parity that no longer holds — folded.** `b3537af9`

Fixing the export path restores the parity the sentence was about, but not the whole of it, so the sentence is now explicit about both halves rather than left to be read as "these two do the same thing":

> Absence is a zero RVA and nothing else, as in `ReadExportRvas` … The two walks agree on what is present and part company only on what a DECLARED directory with a zero Size can be made to yield: that Size is this walk's ONLY bound, so it fails here, while the export walk is bounded by the directory's own counts and reads it.

`ReadExportRvas`'s own summary now says the same thing from its side.

Build clean; `dotnet test --filter "tier!=bridge"` 1799/1799; `ci-all` ALL PASS.

PR #628, opened 2026-09-08
Build a constructor-argument arm from the compose's own fields

### LINKED ISSUE #563: Can't construct a from-scratch Script-archetype MagicEffect
### What you were trying to do

I was trying to build a genuinely unique-feeling enchant for a weapon using Requiem's own "scripted effect" pattern (a MagicEffect whose behavior comes from a compiled Papyrus script wired in via VirtualMachineAdapter properties — the same mechanism Requiem's own real artifacts like Volendrung use), rather than a simpler numeric archetype.

### What you tried

```text
created a new MagicEffect record and attempted to set Archetype.Type to the Script arm via compose= in bulk_apply. This failed immediately with CompositionRequiredException — no parameterless constructor exists for that specific polymorphic arm, so the tool can't even instantiate an empty object for it. We then tried the workaround that's proven to work for other polymorphic union fields: Set one simple leaf field on the desired arm first, hoping that side-effect would force a default object into existence, then compose the rest in a follow-up call. This exact technique worked perfectly for a different arm on the same record type (MagicEffectCloakArchetype, tested side-by-side in the same session) — but for the Script arm specifically, every variation of it hit the identical construction error.
```

### What would have worked

What would have worked: we never found a way to build one from absolute scratch. The actual resolution was to abandon "from scratch" entirely and housecarl_forward_record an existing real Script-archetype effect Requiem or the base game already ships, then edit its already-populated script properties in place — copying a real object sidesteps needing to construct one from nothing. That's a workaround, not a fix: a truly novel scripted effect with no existing analog to fork still has no path in houseCARL today. Given houseCARL is built on Mutagen, this may be an upstream constructor gap in Mutagen's own MagicEffectArchetype class rather than something houseCARL's layer trivially owns — but at minimum, a clearer error pointing at the fork-instead-of-construct workaround would save the next person the same dead-end.

Avick3110:
A compose of a polymorphic arm whose type has no parameterless constructor was accepted by pre-flight and then threw at apply. Mutagen models a MagicEffect's plain `MagicEffectArchetype` with a single constructor, `MagicEffectArchetype(TypeEnum)`, so `compose={"type": "MagicEffectArchetype", "fields": {"Type": "Script"}}` reached `BuildStruct`, fell through `Instantiate` into `InstantiateComposition`, and came back as a `CompositionRequiredException` claiming the type was a `GenderedItem`/`Array2d` composition buildable only from its parts — which it is not. With no way to construct one, a Script-archetype effect could only be had by forwarding an effect that already carried one.

The fix is general and stays on the constructor, not the record type. `BuildStruct` now draws positional constructor args from the compose's OWN fields when no `ctor_args` were given: it picks the smallest public constructor every parameter of which a supplied field names (matched without regard to case, so the field `Type` supplies the parameter `type`) and whose value coerces, passes those to the constructor, and does not set them again in the field pass. Recognition is by the missing parameterless constructor, never by type name, so any arm Mutagen models this way is covered. `ctor_args` still supplies the same values positionally and is unchanged.

The gate now predicts the same thing rather than accepting and throwing: `StructSpecContents` asks `WriteEngine.TryRecognizeInstantiable`, which calls the very method the apply instantiates through, so the two cannot drift. A compose that names none of the constructor's arguments is refused before anything is written, naming the field to add and listing the type's constructors. The same value put in a nested `sets` entry is refused too — nested sets run against the already-built value, so they cannot reach the constructor.

Two things I did not change, both notes rather than diff:

- The discriminator refusal on a dotted `Archetype.Type` Set (P-DISC) is keyed on the polymorphic BASE, `AMagicEffectArchetype`, where `Type` is read-only; the live value may be a `MagicEffectArchetype`, where it is writable. That refusal is now correct advice rather than a dead end — it sends the caller to the arm Set, which works — so I left it alone.
- `CompositionRequiredException`'s message still says every no-parameterless-constructor type is a `GenderedItem`/`Array2d` composition. After this change the only path that can reach it with a constructor-argument type is `MaterializeSubstruct` on a substruct-typed leaf, and no such leaf exists in the Skyrim model, so the wording is unreachable rather than wrong in practice.

Tests: seven through the tool surface (dry runs against the shared world, so nothing is written) covering the reported call, the positional lane, the parameterless arm, and both refusals; three world-free engine tests for what the built arm IS, including a plugin written and re-read as a Script-archetype effect. Eight of the ten fail without the fix; the two that do not are the regression guards.

Run locally: `dotnet build` clean, `dotnet test --filter "tier!=bridge"` 1799/1799, `ci-all` 127/127.

Closes #563

### DIFF
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -1039,6 +1039,12 @@ public sealed class CorpusRulebook
             if (CheckValue(af.Type, f.Value, $"'{f.Key}' on '{spec.Type}'",
                     af.MutableTypeAssemblyQualified ?? af.GetterTypeAssemblyQualified) is { } e) return e;
         }
+        // With no ctor_args the type still has to be BUILDABLE: either it has a parameterless constructor, or the
+        // fields just checked satisfy one of its constructors (a discriminator arm). WriteEngine.TryRecognizeInstantiable
+        // calls the very method BuildStruct instantiates through, so gate and apply cannot drift. Runs AFTER the field
+        // loop so a field that is misspelled or won't coerce is reported as itself, not as a missing constructor arg.
+        if (spec.CtorArgs is null && WriteEngine.TryRecognizeInstantiable(spec.Type, spec.Fields) is { } buildErr)
+            return buildErr;
         foreach (var s in spec.Sets ?? new())
             // siblingEditorIds threads through — a same-call @editorid ref inside a COMPOSED struct's nested Sets
             // (e.g. a VMAD quest-fragment's Property.Object=@<own quest>) validates by the SAME gates as a top-level
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -1012,10 +1012,19 @@ public sealed class CorpusRulebook
         // before the per-field checks. Skipped when CtorArgs is null (the parameterless/fields-only compose path).
         if (spec.CtorArgs is { } ctorArgs && WriteEngine.TryRecognizeCtorArgs(spec.Type, ctorArgs) is { } ctorErr)
             return ctorErr;
+        // The fields BuildStruct hands to the constructor instead of setting — legal to name even when the property
+        // has no setter, and only on the no-ctor_args lane, which is exactly when BuildStruct skips them.
+        var ctorCarried = spec.CtorArgs is null
+            ? WriteEngine.CtorConsumedFields(spec.Type, spec.Fields)
+            : (IReadOnlySet<string>)new HashSet<string>();
         foreach (var f in spec.Fields ?? new())
         {
             var af = structSchema.Fields.FirstOrDefault(x => x.Name == f.Key);
             if (af is null) return FieldNotFound(structSchema, f.Key);
+            // A field the apply cannot set is refused HERE, not thrown mid-apply: BuildStruct's field pass rejects a
+            // read-only property, and a discriminator on an arm (Condition data's Function, an arm's AssociationKey)
+            // is the natural thing to copy out of a read and back into a compose.
+            if (!af.Writable && !ctorCarried.Contains(f.Key)) return WritabilityRejection(structSchema, af);
             // A '@editorid' same-call reference in a compose FIELD (the VMAD alias-fragment
             // Property.Object=@<the quest itself> shape) — legal on a singular FORMLINK field in CREATE context only,
             // mirroring the top-level singular-value gate exactly (formlink-only + declared-earlier-or-self); the
--- a/src/housecarl-core/WriteEngine.cs
+++ b/src/housecarl-core/WriteEngine.cs
@@ -2320,7 +2320,9 @@ public static class WriteEngine
         // ctor_args were given — the caller names the discriminator as a field, which is the natural spelling and the
         // one the discriminator refusal already sends them to.
         var fromFields = spec.CtorArgs is null ? CtorArgsFromFields(type, spec.Fields) : null;
-        var instance = Instantiate(type, spec.CtorArgs ?? fromFields?.Args);
+        // The constructor CtorArgsFromFields chose is the one invoked — not one re-derived from the arg count, which
+        // would pick a different overload of the same arity than the gate validated.
+        var instance = fromFields is { } ff ? Instantiate(ff.Ctor, ff.Args) : Instantiate(type, spec.CtorArgs);
         foreach (var (name, val) in spec.Fields ?? new())
         {
             // A field the constructor already carried is not re-set: it is written, and on an arm whose discriminator
@@ -2451,13 +2453,24 @@ public static class WriteEngine
             var ctor = t.GetConstructors().FirstOrDefault(c => c.GetParameters().Length == ctorArgs.Length)
                 ?? throw new InvalidOperationException(
                     $"{t.Name}: no constructor taking {ctorArgs.Length} arg(s). Ctors: {CtorList(t)}");
-            var ps = ctor.GetParameters();
-            return ctor.Invoke(ps.Select((p, i) => Coerce(ctorArgs[i], p.ParameterType)).ToArray());
+            return Instantiate(ctor, ctorArgs);
         }
         var paramless = t.GetConstructor(Type.EmptyTypes);
         if (paramless is not null) return paramless.Invoke(null);
         return InstantiateComposition(t);
     }
+
+    /// <summary>Invoke ONE already-chosen constructor with positional string args, each coerced to its parameter
+    /// type. Taking the <see cref="ConstructorInfo"/> rather than re-selecting it by arity is what makes the
+    /// fields-lane gate and apply the same choice: two overloads of the same arity are two different builds, and
+    /// arity alone cannot tell them apart.</summary>
+    static object Instantiate(ConstructorInfo ctor, string[] args) =>
+        ctor.Invoke(ctor.GetParameters().Select((p, i) => Coerce(args[i], p.ParameterType)).ToArray());
+
+    /// <summary>Build a type from the constructor its own compose fields satisfy, or null when none does — the seam
+    /// <see cref="BuildStruct"/>'s no-<c>ctor_args</c> lane is made of (choose a constructor, invoke THAT one).</summary>
+    internal static object? BuildFromFieldConstructor(Type t, IReadOnlyDictionary<string, string>? fields) =>
+        CtorArgsFromFields(t, fields) is { } ff ? Instantiate(ff.Ctor, ff.Args) : null;
 
     /// <summary>Recognition-only mirror of <see cref="Instantiate"/>'s ctor-args path — the write pre-flight gate's twin
     /// of the apply-time ctor build. Does this struct type have a constructor of the supplied arity, and does each
@@ -2491,10 +2504,12 @@ public static class WriteEngine
     /// (<c>MagicEffectArchetype(TypeEnum)</c>). Recognised by the missing parameterless ctor, never by type name.
     /// Picks the SMALLEST public constructor whose every parameter is named by a supplied field (matched
     /// case-insensitively, so the parameter <c>type</c> is satisfied by the field <c>Type</c>) and whose value
-    /// coerces; returns those args in positional order together with the field names it consumed, or null when no
-    /// constructor is satisfied. Reads nothing and builds nothing, so the pre-flight gate calls the same method the
-    /// apply does and the two cannot drift.</summary>
-    static (string[] Args, HashSet<string> Consumed)? CtorArgsFromFields(Type t, IReadOnlyDictionary<string, string>? fields)
+    /// coerces; returns THAT constructor with its args in positional order and the field names it consumed, or null
+    /// when no constructor is satisfied. The chosen <see cref="ConstructorInfo"/> travels with the args because the
+    /// args alone do not identify it — two overloads of the same arity would let the apply invoke one the gate never
+    /// validated. Reads nothing and builds nothing, so the pre-flight gate calls the same method the apply does and
+    /// the two cannot drift.</summary>
+    static (ConstructorInfo Ctor, string[] Args, HashSet<string> Consumed)? CtorArgsFromFields(Type t, IReadOnlyDictionary<string, string>? fields)
     {
         if (t.GetConstructor(Type.EmptyTypes) is not null || fields is not { Count: > 0 }) return null;
         foreach (var ctor in t.GetConstructors().Where(c => c.GetParameters().Length > 0)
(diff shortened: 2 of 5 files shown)


Avick3110:
Blind review, no context beyond the repo. **No findings** — nothing I could turn into a failing input survived. What I checked:

**Could any type now be built through a different constructor than before?** No. I enumerated every type name that can legally appear as a compose `spec.Type` — each corpus field's `TypeRef` and `ElementTypeRef` (excluding enum/formlink/scalar cardinalities), plus the arms of every polymorphic base named there — and resolved each through a copy of `ResolveStructType` against Mutagen 0.54.4. 892 names, of which 137 have no public parameterless constructor. All but one are unreachable through a compose gate:

- `Cell`, `Landscape`, `NavigationMesh`, `DialogResponses`, `PlacedObject`/`PlacedNpc`/the other nine `Placed*` arms, `Global*`, `GameSetting*` and the top-level `SkyrimMod.<group>` element types are all corpus Kind `record`, so `ClassifyElement` returns `ElementKind.Record` and `IsComposableElement` is false — they never reach `StructSpecContents`. `Cell.Landscape` is a record-kind substruct `TypeRef`, excluded by `IsComposableSubstructLeaf` for the same reason.
- `TranslatedString` (`APerkEffect.ButtonLabel`) is excluded by `CoercibleLeaf` — it keeps its plain-value Set.

That leaves `MagicEffectArchetype`, reached as an arm of `AMagicEffectArchetype` on `MagicEffect.Archetype`. It has exactly one public constructor. So the new arm changes what happens for exactly the type the issue reported and nothing else, and no existing compose can pick up a different constructor.

**Gate vs apply.** `StructSpecContents` and `BuildStruct` both call `CtorArgsFromFields` with the same `spec.Fields`, so the "does a constructor accept these fields" answer is genuinely one method. The one substitution that happens between gate and apply — `ResolveSiblingRefs` rewriting an `@editorid` field value, which the gate `continue`s past before `CheckValue` but which `TryCoerce` inside `CtorArgsFromFields` would reject — can only bite a formlink constructor parameter, and no reachable type has one.

**One asymmetry, not currently reachable, noting it only so it is on the record.** `CtorArgsFromFields` picks a specific `ConstructorInfo` (smallest arity every parameter of which a field names and coerces), but it returns only the argument strings — `Instantiate` then re-selects with `GetConstructors().FirstOrDefault(c => c.GetParameters().Length == ctorArgs.Length)`. For a type with two public constructors of the same arity the apply could invoke a different one than the gate validated, and `Coerce` would then be handed a value for the wrong parameter type. Nothing in the reachable set has an arity tie (the only reachable type has one constructor), so I cannot make it fail and it is not a finding — but the PR's claim that gate and apply "cannot drift" holds because of the current model, not because of the code. Returning the `ConstructorInfo` and invoking that would make it structural.

**Regression surface.** `CtorArgsFromFields` returns null immediately when a parameterless constructor exists, so the other 755 composable types are untouched. `GenderedItem<T>` / `Array2d<T>` names do not resolve through `ResolveStructType`, so `TryRecognizeInstantiable` catches and defers, and `IsPlainComposableStruct` is unchanged — the `G4-OK-NOCTORARGS` probe arm (`MagicEffectLightArchetype`, which does have a parameterless constructor) and `SUBSTRUCT-SET-ARRAY2D-REJECTED` both still take the same path.

**Tests fail with the fix reverted.** Dry runs do run the real in-memory apply (`WritePatchBuilder` applies, then halts before serialize), so the integration tests exercise `BuildStruct`, not just the rulebook. Reverted, `InstantiateComposition` throws `CompositionRequiredException`, whose own message contains "has no parameterless constructor" and which lands in `WritePatchBuilder`'s generic `catch (Exception)` that renders "pre-flight ACCEPTED it but the apply threw" — so both `DoesNotContain` assertions flip. The refusal tests key on `'Type'`, `fields=` and `NO patch written`, which only the new pre-flight path emits.

**Minor, checked and fine.** `MagicEffectArchetype.Type` is in fact writable (`CanWrite=True` on the arm, read-only only on the `AMagicEffectArchetype` base), so the `Consumed` skip in `BuildStruct` is defensive rather than load-bearing here; a case-differing duplicate field key cannot reach it because the gate's field lookup against the corpus is ordinal. The new write test's use of `Path.GetTempPath()` plus a GUID matches what every other fixture in `housecarl-mcp-tests` does.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1039,6 +1039,12 @@ bool IsValueCoercibleElement(FieldSchema leaf)
             if (CheckValue(af.Type, f.Value, $"'{f.Key}' on '{spec.Type}'",
                     af.MutableTypeAssemblyQualified ?? af.GetterTypeAssemblyQualified) is { } e) return e;
```
Avick3110, line 1049: **Compose fields still skip the writability gate — the same accept-then-throw this PR closes.**

The new line below makes the gate predict `Instantiate`. The field loop right above it still doesn't predict the field pass: it checks `FieldNotFound` and `CheckValue`, but never `af.Writable`. So a compose naming a read-only struct field is accepted by pre-flight and throws mid-apply.

Concretely:

```
op=Set field_path="Conditions[0].Data"
compose={"type":"GetActorValueConditionData",
         "fields":{"Function":"GetActorValue","ActorValue":"Destruction"}}
```

`Function` is `Writable=false` on every `*ConditionData` arm (`CorpusGenerator.PolyDiscriminators`, rule R2), but both gate checks pass. `BuildStruct` then hits `if (!p.CanWrite) throw new InvalidOperationException($"Field '{name}' on '{spec.Type}' is not writable")` (WriteEngine.cs:2301), and `WritePatchBuilder` renders it as `pre-flight ACCEPTED it but the apply threw — a real inconsistency`. Naming `Function` alongside `ActorValue` is the natural spelling for a caller copying what a read showed them.

This rulebook already owns the right refusal for it — `WritabilityRejection`'s P-DISC branch (line 603, "its value is fixed by which arm is selected") — but only the top-level leaf paths call it (lines 326, 509). It isn't a one-off: `MagicEffectArchetype.AssociationKey` is the same, and ~447 struct/arm corpus types carry at least one read-only field.

Not introduced here, but it's the same bug class as #563, one line above the fix, and the check is available:

```suggestion
            if (!af.Writable) return WritabilityRejection(structSchema, af);
            if (CheckValue(af.Type, f.Value, $"'{f.Key}' on '{spec.Type}'",
                    af.MutableTypeAssemblyQualified ?? af.GetterTypeAssemblyQualified) is { } e) return e;
```

src/housecarl-core/WriteEngine.cs
```diff
@@ -2455,6 +2463,73 @@ static object Instantiate(Type t, string[]? ctorArgs)
...
+    /// Picks the SMALLEST public constructor whose every parameter is named by a supplied field (matched
+    /// case-insensitively, so the parameter <c>type</c> is satisfied by the field <c>Type</c>) and whose value
+    /// coerces; returns those args in positional order together with the field names it consumed, or null when no
+    /// constructor is satisfied. Reads nothing and builds nothing, so the pre-flight gate calls the same method the
+    /// apply does and the two cannot drift.</summary>
+    static (string[] Args, HashSet<string> Consumed)? CtorArgsFromFields(Type t, IReadOnlyDictionary<string, string>? fields)
+    {
+        if (t.GetConstructor(Type.EmptyTypes) is not null || fields is not { Count: > 0 }) return null;
+        foreach (var ctor in t.GetConstructors().Where(c => c.GetParameters().Length > 0)
+                              .OrderBy(c => c.GetParameters().Length))
```
Avick3110: **The gate and the apply select the constructor by two different rules.**

`CtorArgsFromFields` picks the smallest constructor *whose parameters are all named and coercible*, then throws that choice away and returns only positional strings. `Instantiate` re-selects with `GetConstructors().FirstOrDefault(c => c.GetParameters().Length == ctorArgs.Length)` — arity alone. Those agree only while no candidate type has two constructors of the same arity.

When one does: overloads `(A a)` and `(B b)`, fields naming `b`. `CtorArgsFromFields` skips `(A a)` (no field names `a`), matches `(B b)`, returns one arg. `TryRecognizeInstantiable` sees non-null and pre-flight accepts. `BuildStruct` then calls `Instantiate(type, [value])`, which takes the *first* 1-arg ctor — `(A a)` — and `Coerce(value, typeof(A))` either throws mid-apply (accepted-then-threw again, the thing this PR exists to remove) or, if both types coerce, silently builds the wrong arm.

Latent today: `MagicEffectArchetype` has exactly one constructor, and I checked the rest — within Mutagen the no-parameterless-ctor types with duplicate-arity overloads are major records, which the corpus gates exclude from compose and `MaterializeSubstruct` guards by `IMajorRecordGetter`. But the PR sells this as recognition by shape, never by type name ("any arm Mutagen models this way is covered"), and gate/apply agreement here is by coincidence rather than by construction.

Cheap fix: return the `ConstructorInfo` you chose alongside `Consumed`, and give `Instantiate` an overload that invokes that constructor instead of re-deriving it from the arity.

src/housecarl-core/CorpusRulebook.cs
Avick3110: Folded — 01e72882. Real, and it reproduces exactly as written:

```
error: engine error applying [Set Conditions[0].Data] to MagicEffect 000806:HcRecMaster.esm:
pre-flight ACCEPTED it but the apply threw — a real inconsistency, surfaced not swallowed (Q3):
InvalidOperationException: Field 'Function' on 'GetActorValueConditionData' is not writable
```

The gate now runs `af.Writable` in the field loop, as you wrote it, with one addition: it skips the fields the constructor carries. `BuildStruct` skips those before its own `CanWrite` check, so a gate that did not skip them would refuse a compose the apply can build — an arm whose ctor parameter maps to a read-only property. Nothing today is that shape (`MagicEffectArchetype.Type` is `Writable=true`, so this PR's lane is unaffected either way), but it is the difference between the two agreeing and the two happening to agree. `WriteEngine.CtorConsumedFields` answers it from the same `CtorArgsFromFields` the apply builds through, and only on the no-`ctor_args` lane, which is exactly when `BuildStruct` skips anything.

Before folding I checked the gate's corpus writability against the CLR property the apply actually reflects on, for every field of every non-enum, non-record corpus type — 5,246 fields:

- corpus read-only / property settable: **0**. So the new refusal cannot refuse a compose the field pass would have written.
- corpus writable / property get-only: 116, every one of them a `SkyrimMod` record group or one of `PlayerSkills`' two arrays. That is the other direction (still accept-then-throw), none of them is a compose target, and this change does not touch it.

Test: `ComposeFieldWritabilityTests` — your exact call, dry run against the shared world. It asserts the refusal names `'Function'` and `discriminator`, and that the response no longer says the apply threw; a second test keeps the same compose *without* that field composing, so the refusal is about the one field and not the arm. Verified failing with the gate line disabled and passing with it.

One note, no change made: the refusal is `WritabilityRejection`'s P-DISC branch as you suggested, whose second sentence — "Set the parent polymorphic field to a different arm" — is what the caller is already doing when the compose *is* the arm Set. The first half ("its value is fixed by which arm is selected") carries the fix, and this file's own rule is that a per-door phrasing is where the doors start describing the field differently, so I left the shared wording alone. Say the word if you want a compose-door sentence instead.

`dotnet build` clean, `dotnet test --filter "tier!=bridge"` 1803/1803, `ci-all` 127/127.

src/housecarl-core/WriteEngine.cs
Avick3110: Folded — 90a352b6. `CtorArgsFromFields` now returns the `ConstructorInfo` it chose alongside the args and the consumed field names, and `BuildStruct` invokes *that* constructor through a new `Instantiate(ConstructorInfo, string[])` overload. The explicit `ctor_args` lane is unchanged: it still selects by arity, and now calls the same invoke helper, so there is one place that turns a chosen constructor into an instance.

Latent, as you said — I get the same answer from the other end: over the corpus, the only non-enum, non-record types with no public parameterless constructor are `MagicEffectArchetype` (one constructor) and `SkyrimMod` (one constructor, and not a compose target). No arity ties anywhere in reach, so there is no input that fails today.

Test: `ConstructorSelectionTests`, world-free. Two synthetic types with `(int alpha)` and `(string beta)` declared in the two possible orders, fields naming `Beta`. Arity-alone selection takes whichever one-arg constructor reflection reports first, so whatever that order is, one of the two types picks `(int alpha)` and `Coerce("hello", int)` throws — the test fails without the fix (verified: one of the two cases threw exactly there) and passes with it, without depending on reflection's constructor order. It drives `WriteEngine.BuildFromFieldConstructor`, the internal seam that is the pair under test — choose a constructor, invoke that one. The tool surface cannot reach it: `ResolveStructType` resolves Mutagen types only, so a compose can never name a synthetic type.

`dotnet build` clean, `dotnet test --filter "tier!=bridge"` 1803/1803, `ci-all` 127/127.

PR #629, opened 2026-09-08
Say whether a no-field read is a typo or an unmodeled field

### LINKED ISSUE #528: [Mutagen gap] Ingestible (ALCH) models no VirtualMachineAdapter, and the read note cannot tell not-modelled from a typo

Avick3110:
On the projection lane, a `project.fields` path that named no field came back as `(no field X)` whether the
name was misspelled or was a real Mutagen field the record type does not carry. Those need opposite next moves —
fix the spelling, versus nothing in the path is wrong and the data is out of reach — so the note now says which.
The verdict comes from the generated schema, read by field name instead of by type: a name the corpus carries on
other types is not mistyped, and a name it carries nowhere is. `VirtualMachineAdapter` on a Spell now reads
"not a mistyped name — Mutagen models 'VirtualMachineAdapter' on 34 other type(s) (Activator,
AlchemicalApparatus, Armor, …), just not on Spell"; `Effcts` reads "a mistyped name … did you mean 'Effects'?".
A case-only slip (`editorid`) gets the exact spelling, which `PluginNameSuggest.Nearest` declines on its own
because it is case-insensitive and treats the match as exact.

The by-field index is its own file (`src/housecarl-core/ModeledFieldIndex.cs`) rather than more lines in
`ReadEngine.cs`. It caches one slot keyed by `CorpusRulebook.CorpusPath` — a process-global the probes and test
worlds repoint — and memoises the verdict per (corpus, owner type, name), because a scan dead-ends the same way
on every record it crosses and the nearest-name sweep walks the owner's whole field list. Where the corpus is not
built the note stays the bare `(no field X)`; saying less is not saying something wrong. The list/dict arm of
`NoFieldNote` is untouched, as is `ReadEngine.IsNoSuchFieldNote` and the `where=` lane's parsing of the note.

This is the second half of the issue only. The first half is the Mutagen gap itself — `Ingestible` models no
`VirtualMachineAdapter` — which is not code work here.

Two things noticed and left alone. `FieldPredicate.AccountingNote`'s all-records-missed sentence still states the
same either/or the projection lane no longer has to ("a mistyped path, or a field that doesn't exist on this record
type"); it could route through the same verdict, but that is the `where=` lane and the issue calls it already
handled. And nothing surfaces the Mutagen-versus-xEdit gap directly — the note can say Mutagen models a field
elsewhere, never that the game has it here.

Closes #528

### DIFF
--- a/src/housecarl-core/ModeledFieldIndex.cs
+++ b/src/housecarl-core/ModeledFieldIndex.cs
@@ -20,8 +20,10 @@ public static class ModeledFieldIndex
 {
     /// <summary>What the schema says about a name that would not resolve on <c>OwnerType</c>.
     /// <para><see cref="OnOwner"/> true means the catalog DOES list the field on the owner type — the walk and the
-    /// schema disagree, which is a defect to name rather than a diagnosis to give.</para></summary>
-    public readonly record struct Verdict(bool OnOwner, IReadOnlyList<string> ModeledOn, string? Near);
+    /// schema disagree, which is a defect to name rather than a diagnosis to give.</para>
+    /// <para><see cref="NearIsCaseSlip"/> true means <see cref="Near"/> is the owner's own spelling of the very
+    /// name asked for, differing only in case — a certainty about the owner, not the nearest-name guess.</para></summary>
+    public readonly record struct Verdict(bool OnOwner, IReadOnlyList<string> ModeledOn, string? Near, bool NearIsCaseSlip);
 
     /// <summary>One slot, keyed by the corpus path it was built from. <c>CorpusRulebook.CorpusPath</c> is a
     /// process-global the probes and test worlds repoint at their own generated corpus, so a flat cache would
@@ -53,20 +55,22 @@ public static class ModeledFieldIndex
             var on = idx.ByField.TryGetValue(field, out var types) ? types : Array.Empty<string>();
             bool onOwner = on.Contains(owner, StringComparer.Ordinal);
             var others = onOwner ? on.Where(t => !string.Equals(t, owner, StringComparison.Ordinal)).ToArray() : on;
-            return new Verdict(onOwner, others, onOwner ? null : NearestOn(idx.ByType, owner, field));
+            var (near, caseSlip) = onOwner ? default : NearestOn(idx.ByType, owner, field);
+            return new Verdict(onOwner, others, near, caseSlip);
         });
     }
 
-    /// <summary>The owner type's own field that a miss most likely meant. A CASE-only difference is checked first
-    /// and answered exactly: field names are case-sensitive, and <see cref="PluginNameSuggest.Nearest"/> is
-    /// case-insensitive, so it declines the very match that explains the miss.</summary>
-    static string? NearestOn(Dictionary<string, string[]> byType, string ownerTypeName, string fieldName)
+    /// <summary>The owner type's own field that a miss most likely meant, and whether it is the CASE-only slip.
+    /// That case is checked first and answered exactly: field names are case-sensitive, and
+    /// <see cref="PluginNameSuggest.Nearest"/> is case-insensitive, so it declines the very match that explains
+    /// the miss.</summary>
+    static (string? Near, bool CaseSlip) NearestOn(Dictionary<string, string[]> byType, string ownerTypeName, string fieldName)
     {
-        if (!byType.TryGetValue(ownerTypeName, out var fields)) return null;
+        if (!byType.TryGetValue(ownerTypeName, out var fields)) return default;
         foreach (var f in fields)
-            if (string.Equals(f, fieldName, StringComparison.OrdinalIgnoreCase)) return f;
+            if (string.Equals(f, fieldName, StringComparison.OrdinalIgnoreCase)) return (f, true);
         var near = PluginNameSuggest.Nearest(fieldName, fields, 1);
-        return near.Count > 0 ? near[0] : null;
+        return near.Count > 0 ? (near[0], false) : default;
     }
 
     static (Dictionary<string, string[]> ByField, Dictionary<string, string[]> ByType)? Index()
--- a/src/housecarl-core/ModeledFieldIndex.cs
+++ b/src/housecarl-core/ModeledFieldIndex.cs
@@ -42,12 +42,15 @@ public static class ModeledFieldIndex
 
     /// <summary>What the schema knows about <paramref name="fieldName"/> not resolving on
     /// <paramref name="ownerTypeName"/> (a catalog name — <c>Ingestible</c>, not <c>IIngestibleGetter</c>).
-    /// Null when the corpus is not built or will not parse: the caller then says only what it knows, never a
-    /// guessed verdict.</summary>
+    /// Null when the corpus is not built or will not parse, and null for an owner type the catalog does not carry:
+    /// the caller then says only what it knows, never a guessed verdict.</summary>
     public static Verdict? Diagnose(string ownerTypeName, string fieldName)
     {
         var path = CorpusRulebook.CorpusPath;
         if (Index() is not { } idx) return null;
+        // A read walk lands on plenty the catalog does not model — a FormLink, a string. There is no schema there
+        // to weigh the name against, so any verdict about such an owner names a schema that does not exist.
+        if (!idx.ByType.ContainsKey(ownerTypeName)) return null;
         return Verdicts.GetOrAdd((path, ownerTypeName, fieldName), key =>
         {
             Interlocked.Increment(ref VerdictComputations);
--- a/src/housecarl-core/ModeledFieldIndex.cs
+++ b/src/housecarl-core/ModeledFieldIndex.cs
@@ -1,0 +1,99 @@
+using System.Collections.Concurrent;
+
+namespace HousecarlCore;
+
+/// <summary>
+/// Which catalog types model a field NAME — the generated schema read the other way round, by field instead of
+/// by type.
+///
+/// A path that dead-ends on a record has two causes that need opposite next moves, and the dead-end itself
+/// cannot tell them apart: the name is MISTYPED (fix the spelling), or the name is a real Mutagen field that
+/// this record type does not carry (Mutagen models <c>VirtualMachineAdapter</c> on some record types and not
+/// others, so a scripted ALCH is invisible and there is nothing to fix in the path). Only the schema can
+/// separate them, and it separates them by construction — the set of names here is the set Mutagen models,
+/// never a hand list.
+///
+/// Lookups are ORDINAL, matching <c>WriteEngine.ResolveProperty</c>'s own case-sensitive resolve, so the index
+/// and the resolver cannot disagree about what counts as a field.
+/// </summary>
+public static class ModeledFieldIndex
+{
+    /// <summary>What the schema says about a name that would not resolve on <c>OwnerType</c>.
+    /// <para><see cref="OnOwner"/> true means the catalog DOES list the field on the owner type — the walk and the
+    /// schema disagree, which is a defect to name rather than a diagnosis to give.</para></summary>
+    public readonly record struct Verdict(bool OnOwner, IReadOnlyList<string> ModeledOn, string? Near);
+
+    /// <summary>One slot, keyed by the corpus path it was built from. <c>CorpusRulebook.CorpusPath</c> is a
+    /// process-global the probes and test worlds repoint at their own generated corpus, so a flat cache would
+    /// answer from the previous world's schema; a slot per path grows without bound across a probe run. One
+    /// slot, rebuilt when the path changes, is both correct and bounded.</summary>
+    static (string Path, Dictionary<string, string[]> ByField, Dictionary<string, string[]> ByType)? _cache;
+    static readonly object Gate = new();
+
+    /// <summary>Memoised per (corpus path, owner type, field name): a scan hits the same dead-end once per scanned
+    /// record and only the first note is kept, so the nearest-name sweep must not run per record.</summary>
+    static readonly ConcurrentDictionary<(string, string, string), Verdict> Verdicts = new();
+
+    /// <summary>How many verdicts have actually been computed — the memo's own counter, so a test can say a scan
+    /// over many records diagnoses the dead-end once.</summary>
+    internal static int VerdictComputations;
+
+    /// <summary>What the schema knows about <paramref name="fieldName"/> not resolving on
+    /// <paramref name="ownerTypeName"/> (a catalog name — <c>Ingestible</c>, not <c>IIngestibleGetter</c>).
+    /// Null when the corpus is not built or will not parse: the caller then says only what it knows, never a
+    /// guessed verdict.</summary>
+    public static Verdict? Diagnose(string ownerTypeName, string fieldName)
+    {
+        var path = CorpusRulebook.CorpusPath;
+        if (Index() is not { } idx) return null;
+        return Verdicts.GetOrAdd((path, ownerTypeName, fieldName), key =>
+        {
+            Interlocked.Increment(ref VerdictComputations);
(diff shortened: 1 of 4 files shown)


src/housecarl-core/ReadEngine.cs
```diff
@@ -444,7 +449,23 @@ static string NoFieldNote(object owner, string segName, string? precedingField,
...
+        // No corpus (not built / unparseable) ⇒ the bare note. Saying less is not saying something wrong.
+        if (ModeledFieldIndex.Diagnose(typeName, segName) is not { } v) return $"{NoFieldPrefix}{segName})";
+
+        if (v.OnOwner)
+            return $"{NoFieldPrefix}{segName}: {typeName} declares '{segName}' but the read walk cannot resolve it)";
+
+        if (v.ModeledOn.Count > 0)
+        {
+            var shown = string.Join(", ", v.ModeledOn.Take(3)) + (v.ModeledOn.Count > 3 ? ", …" : "");
+            return $"{NoFieldPrefix}{segName}: not a mistyped name — Mutagen models '{segName}' on " +
```
Avick3110: A `project.fields` path that is the exact case-slip this change exists to catch gets told its spelling is fine.

`Weapon` models `Data`. The corpus also models a field spelled literally `DATA` — on `DialogResponses`, and on nothing else (checked in `generated/corpus.json`). So for `housecarl_records types=["WEAP"] project.fields=["DATA"]` the walk dead-ends, `ModeledOn.Count` is 1, and this branch emits:

```
(no field DATA: not a mistyped name - Mutagen models 'DATA' on 1 other type(s) (DialogResponses), just not on Weapon)
```

It is a mistyped name, and the right note is `did you mean 'Data'?`. `Verdict.Near` already holds it: `Diagnose` runs `NearestOn` for every `!OnOwner` verdict, so `Near == "Data"` here — the branch just never looks at it, and asserts the opposite of the truth instead. `DATA` is xEdit's own spelling of that subrecord, so it is a likely thing for a caller to type at a WEAP.

Same masking, same corpus, 16 record-type owners: `Package`, `MusicType`, `SoundOutputModel`, `NavigationMesh`, `Weapon` and the `GameSetting*` / `Global*` families (all `DATA` vs `Data`), `IdleAnimation` (`FileName` vs its real `Filename`), `NavigationMeshInfoMap` (`NavmeshVersion` vs `NavMeshVersion`); 57 owners counting structs. `ACaseOnlySlipIsOfferedTheExactSpelling` passes only because `editorid` happens to be modeled nowhere.

A case-insensitive match on the owner's OWN fields should win ahead of "modeled elsewhere" — the owner's spelling is the thing to fix regardless of what else the corpus carries under that casing.

src/housecarl-core/ReadEngine.cs
```diff
@@ -444,7 +449,23 @@ static string NoFieldNote(object owner, string segName, string? precedingField,
             var pf = precedingField ?? "<field>";
             return $"{NoFieldPrefix}'{segName}': '{pf}' is a list/dict — {ListHopRemedy(owner, segName, pf, trailing)})";
         }
-        return $"{NoFieldPrefix}{segName})";
+
+        var typeName = RecordNaming.StripGetterInterface(RecordNaming.StripOverlay(owner.GetType().Name));
```
Avick3110, line 460: The verdict is asserted about owner types the corpus does not model, and the sentence it produces is then false.

`owner` is whatever the walk is standing on, not necessarily a catalog type. `FormLink<T>` is one such: it is not `IEnumerable` (checked against Mutagen 0.54.4, the pinned version), so it falls past the collection branch, and its `Type.Name` is ``FormLink`1`` — a string no corpus key can ever equal.

So `housecarl_records types=["SPEL"] project.fields=["HalfCostPerk.Name"]` reads:

```
(no field Name: not a mistyped name - Mutagen models 'Name' on 83 other type(s) (APackageData, Activator, ActorValueInformation, ...), just not on FormLink`1)
```

A FormLink has no `Name` and never will. The path is wrong and the caller has to resolve the link; being told the name is right and the data is out of Mutagen's reach sends them the other way. `string` owners land the same way (`string` is excluded from the collection test above, so `EditorID.Name` on a Spell reports "just not on String").

The mistyped arm below has the mirror of the same problem: `check the name against {typeName}'s schema` names a type with no schema in the corpus, and `Near` is always null for these owners because `ByType` has no such key — so even a plain case slip like `HalfCostPerk.formkey` silently loses its `did you mean 'FormKey'?`.

`ModeledFieldIndex` can only speak about types the catalog carries; where `idx.ByType` has no entry for `typeName`, the bare `(no field X)` is the only thing that is not wrong.

src/housecarl-core/ReadEngine.cs
Avick3110: Real, and reproduced against the corpus: `ByField["DATA"]` is exactly `[DialogResponses]`, `Weapon` carries `Data` and not `DATA`, so `DATA` at a WEAP took the modeled-elsewhere arm with `Near == "Data"` sitting unread. Your masking census reproduces too: 57 owners, 16 of them record-kind, the same list you named.

Fixed in b322731e. `Verdict` now carries `NearIsCaseSlip`, set when `NearestOn` matched the owner's own field case-insensitively rather than by edit distance, and `NoFieldNote` takes that arm ahead of the modeled-elsewhere one. Only the case-only match jumps the queue; the fuzzy nearest still does not, so `VirtualMachineAdapter` on a Spell keeps its coverage-limit verdict. `DATA` at a WEAP now reads `(no field DATA: a mistyped name - field names are case-sensitive; did you mean 'Data'?)`.

Test `ACaseOnlySlipBeatsTheSameCasingModeledElsewhere` reads `DATA` off the fixture's weapon and asserts the exact spelling plus the absence of "not a mistyped name"; it fails on the pre-fix build with the note you quoted. `ACaseOnlySlipIsOfferedTheExactSpelling` still passes unchanged, now through the new arm.

src/housecarl-core/ReadEngine.cs
Avick3110: Real. The corpus has no key for ``FormLink`1`` or `String` (checked), so `ByType` misses, `Near` is null for these owners, and the modeled-elsewhere arm still fired off `ByField` alone — asserting a schema that does not exist and pointing the caller away from the link they have to resolve.

Fixed in bf77d838, at the place you named: `Diagnose` returns null when `idx.ByType` has no entry for the owner type, so `NoFieldNote` falls to the bare `(no field X)`. That covers the mistyped arm's mirror of the same problem as well — no "check the name against ``FormLink`1``'s schema" either. Nothing changes for owners the catalog does carry, structs included: the gate is the same `ByType` lookup `NearestOn` already made.

Test `AnOwnerTheCatalogDoesNotModelGetsTheBareNote` reads `HalfCostPerk.Name` off the fixture's Spell and asserts the bare note with no "mistyped name" in it; it fails on the pre-fix build with the 83-other-types sentence.

src/housecarl-core/ReadEngine.cs
```diff
@@ -444,7 +449,30 @@ static string NoFieldNote(object owner, string segName, string? precedingField,
...
+        if (v.NearIsCaseSlip)
+            return $"{NoFieldPrefix}{segName}: a mistyped name — field names are case-sensitive; " +
+                   $"did you mean '{v.Near}'?)";
+
+        if (v.ModeledOn.Count > 0)
+        {
+            var shown = string.Join(", ", v.ModeledOn.Take(3)) + (v.ModeledOn.Count > 3 ? ", …" : "");
+            return $"{NoFieldPrefix}{segName}: not a mistyped name — Mutagen models '{segName}' on " +
+                   $"{v.ModeledOn.Count:N0} other type(s) ({shown}), just not on {typeName})";
+        }
```
Avick3110, line 484: **The modeled-elsewhere branch shadows a real typo suggestion.**

This branch is tested before the nearest-name branch below it, so a plain typo whose spelling happens to exist as a field on any of the other 1,174 catalog types is asserted "not a mistyped name" — and the correction `Diagnose` already computed is thrown away.

Concretely, `project.fields=["Effect"]` on a Spell. Spell carries `Effects`, and `v.Near` is already `"Effects"` (PluginNameSuggest's PREFIX tier scores it 800). But the corpus models the exact name `Effect` on one other type, so this branch wins and the note reads:

> not a mistyped name — Mutagen models 'Effect' on 1 other type(s) (…), just not on Spell

i.e. "nothing in the path is wrong" when a one-character fix is sitting right there. Same shape for `Keyword` vs `Keywords` (25 other types), `Perk` vs `Perks` (4), `Faction` vs `Factions` (20), and `Description` / `Model` / `Sound` / `Item`. Singular-plural is the commonest field-path slip, so this is the case the change exists to diagnose, now given the opposite advice.

Keeping the confident claim only when the owner has no near match:

```suggestion
        if (v.ModeledOn.Count > 0)
        {
            var shown = string.Join(", ", v.ModeledOn.Take(3)) + (v.ModeledOn.Count > 3 ? ", …" : "");
            // The owner's own near field outranks the claim: a name modeled elsewhere can still be this
            // record's typo, and "not a mistyped name" over a one-character fix sends the caller away.
            var lead = v.Near is null ? "not a mistyped name — Mutagen models" : "Mutagen models";
            var alt = v.Near is { } nm ? $"; did you mean '{nm}'?" : "";
            return $"{NoFieldPrefix}{segName}: {lead} '{segName}' on " +
                   $"{v.ModeledOn.Count:N0} other type(s) ({shown}), just not on {typeName}{alt})";
        }
```

src/housecarl-core/ModeledFieldIndex.cs
```diff
@@ -0,0 +1,106 @@
...
+    internal static int VerdictComputations;
+
+    /// <summary>What the schema knows about <paramref name="fieldName"/> not resolving on
+    /// <paramref name="ownerTypeName"/> (a catalog name — <c>Ingestible</c>, not <c>IIngestibleGetter</c>).
+    /// Null when the corpus is not built or will not parse, and null for an owner type the catalog does not carry:
+    /// the caller then says only what it knows, never a guessed verdict.</summary>
+    public static Verdict? Diagnose(string ownerTypeName, string fieldName)
+    {
+        var path = CorpusRulebook.CorpusPath;
+        if (Index() is not { } idx) return null;
```
Avick3110: **The memo key can name a corpus the verdict did not come from.**

`path` is read here, then `Index()` reads the same process-global again on the next line. `CorpusRulebook.CorpusPath` is settable and the probe runner and test worlds repoint it at their own generated corpus — a repoint landing between these two reads stores a P2-derived verdict under key `(P1, owner, name)`.

It self-heals only by accident: the next `Index()` rebuild for P1 clears the whole memo. Reading the path once, out of `Index()`, closes the window — e.g. return it alongside the two dictionaries and key on `idx.Path`.

Avick3110, line 52: **The global lock is taken per dead-ended record, ahead of the memo.**

`Index()` is called before `Verdicts` is consulted, and it takes the static `Gate` unconditionally. So a `types=["SPEL"]` scan with a dead-ending `project.fields` path acquires the process-wide lock and allocates a tuple once per scanned record; only the nearest-name sweep is actually skipped. The sibling `ReadEngine.HopVerdicts` memo takes no lock on its hot path at all.

Consulting `Verdicts` for an existing entry first, or holding the built index in a volatile field read outside the lock, gets the cost the memo's own comment describes.

src/housecarl-core/ReadEngine.cs
Avick3110: Real, and reproduced against `generated/corpus.json`. `ByField["Effect"]` is exactly `[CriticalData]`, `Spell` carries `Effects`, and `Effects` scores 800 (PREFIX) as the nearest name on Spell — so `Effect` at a SPEL took the modeled-elsewhere arm with `Near == "Effects"` sitting unread, exactly as you quote it.

Fixed in 33095af6, with the branch you wrote: the confident claim is kept only when the owner offers nothing closer (`Near is null`), and where it does offer something the note states the coverage fact without the verdict and appends the spelling.

Two notes on the census, from replaying `NearestOn` over the corpus. The masking is per owner type, not per name: `Keyword` masks on Spell, Npc and Armor (all spell it `Keywords`), and `Perk` masks on Spell (nearest `HalfCostPerk`) — but `Faction`, `Model`, `Sound` and `Item` clear nothing on a Spell, so those kept the confident sentence and keep it now; `Description` is a real Spell field and never reached the arm. And the flagship cases are unaffected because `Near` is null for them: `VirtualMachineAdapter` on Spell and on Ingestible, and `ArmorRating` on Weapon, all still read "not a mistyped name", so `AFieldTheTypeDoesNotModelIsToldTheNameIsNotMistyped` and `TheTypesThatDoModelTheFieldAreNamed` pass unchanged.

Test `AnOwnerNearFieldBeatsTheModeledElsewhereClaim` reads `Effect` off the fixture's Spell and asserts the exact spelling plus the absence of "not a mistyped name"; it fails on the pre-fix build with the note you quoted. The changelog entry now says the note names the owner's near field rather than calling the name right.

src/housecarl-core/ModeledFieldIndex.cs
Avick3110: Real, and it does not only self-heal. The insert can outlive the clear that was supposed to cover it: A reads `path` as P1, is descheduled, `Index()` then returns P2's dictionaries, and A's `GetOrAdd` factory — which closes over that P2 index — can run *after* a later repoint back to P1 has rebuilt and cleared the memo. The P2-derived verdict then sits under key `(P1, owner, name)` with `_cache.Path == P1`, and the next P1 reader is served it.

Fixed in cce51606, the way you named: `Index()` returns the path alongside the two dictionaries (as one `Snapshot` record), and `Diagnose` keys on `idx.Path` instead of reading the global a second time. A verdict is now filed under the corpus it was computed from, whatever the global does in between.

No test comes with it, and I do not think one can: single-threaded the two reads always return the same string, so the old and new code are indistinguishable — the difference is only the interleaving above. A test that raced repoints against `Diagnose` would be timing-dependent, and a flaky test is worse than none. The change is a removal of the second read, so the existing no-field-note tests carry the behaviour; the suite and `ci-all` are green.

Avick3110: Real: `Index()` took `Gate` on every dead-ended record, so the memo skipped the nearest-name sweep and nothing else — the lock and a tuple per record stayed.

Fixed in cce51606, with the second of the two remedies you named, since it also closes the memo-key window in the sibling comment. The index is published as one immutable `Snapshot` reference in a `volatile` static; `Index()` reads that field and returns it when the path matches, so the gate is taken only on a rebuild, and no tuple is allocated on the hot path. `Diagnose` then does a plain `Verdicts.TryGetValue` before `GetOrAdd`, so a repeat dead-end also allocates no factory closure. That leaves `HopVerdicts` and this memo with the same shape on their hot paths.

Same caveat as the sibling: no new test. The change is the cost of a path that already had test coverage, not a behaviour, and `AScanComputesOneVerdictForTheWholeScan` still pins the one computation per scan. Suite and `ci-all` green.

Issue #614, opened 2026-09-07
[GAP] Who does this filter reach: preview the set a SPID or KID line selects


Avick3110:
Neither spid-authoring nor kid-authoring can tell a user which actors or items a `_DISTR.ini` or `_KID.ini` line will reach before the file is written. The audit could not distinguish a line reaching 17 actors from one reaching 4,024, and nothing on the surface reports the reached set.

Wanted: a query that takes a SPID or KID filter line and returns the set it selects on the live order, so a line can be checked before distribution. Found by the skill eval (spid-authoring D5.35, kid-authoring D4.41/D5.35). After 2.0 unless promoted (DECISIONS 2026-09-07, rewrite ruling 6).

PR #633, opened 2026-09-08
Rewrite the kid-authoring skill: ground a KID line before it ships

Avick3110:
Wave 1 of the skill rewrite: `kid-authoring`, built from its approved rewrite document with the verifier's overturned rows corrected as the verifier says. The skill had no houseCARL calls in it at all, so there was no 1.x rot to migrate; the defects were corpus numbering, a pointer the packager deletes from every shipped copy, an eval file in the wrong shape, and a missing trait-to-field column. The body is rebuilt around the one thing the audit says the skill did not do: it now **grounds** a line before the file ships. A composed line's type-plus-trait predicate is counted with `housecarl_records` (`types`, `where`, `counts_only=true`), the count is reported with the call that produced it, and that count is the stop condition — a census of zero, or one the user did not expect, re-cuts the filter instead of shipping. The same step states its own bound: it proves one predicate, not KID's evaluation order or `chance`, which is #614 and not in 2.0. Alongside it the frontmatter gains `compatibility` and a description that front-loads "Keyword Item Distributor (KID)" and the v3.5.0 floor (383 → 441 characters), the four-way distributor routing table leaves the body for ServerInstructions (F1) leaving one lane sentence and one sibling pointer, the trait-to-field mapping becomes a pointer at `housecarl:mutagen-reference` (F16), the dead `_CORPUS_STATUS.md` pointer goes and its provenance is inlined (F42), the wrong section numbers at the top of `grammar-core.md` are fixed (filters are section 2, traits section 3), the four 100-line reference files gain a section list, and `evals/eval_set.json` becomes `evals/evals.json` in the published shape with a without-skill arm on every case.

No issue is named as closing: the document names only #614, which it writes around rather than closes.

## Migration

The migration table runs the other way here — the rewrite *introduces* houseCARL calls for the first time. Every tool and parameter written into the body appears in `surface/tools.md` as written: `housecarl_records` with `types` / `where` / `counts_only` / `project` (form `identity`) / `limit` / `to_file` / `formids` (`["@<path>"]`) / `source`, and `housecarl_create` with `records` and `patch`. Row 7 (previewing the whole line as KID evaluates it) is not expressible on the 2.0 surface and the body says so, naming #614.

## Open questions, decided as follows or left for Aaron

1. **Rename, or accept D3.8?** — **left for Aaron.** A rename moves the folder, the frontmatter `name`, the identity, every sibling pointer in the other eleven rewrite documents and F16's destination sentence. Nothing here changes it; the description carries the full framework name, which is the half a description can carry, and D3.8 stays failing and declared.
2. **Is "ignored" a second behaviour, or the same one?** — **left for Aaron**, and the verifier's builder note followed: "ignored" is *not* retired from `references/types.md` or `references/traits.md`, because those are scoped claims about a traits section being dropped while the line still applies, and flattening them to "skipped" could lose a real distinction. The body uses **skipped** only for the malformed-entry case, on `grammar-core.md`'s authority, and refers the trait-less case to the types file rather than asserting a failure mode.
3. **Where the trait → field row physically lives** — **decided, by Aaron's 2026-09-08 answer to FOLD-IN Q7**: a hand-written notes file beside the generated shards, never in the generated corpus. The kid body carries F16's sentence verbatim; the row itself lands in `mutagen-reference`'s own rewrite, not here, so nothing outside this skill is touched.
4. **Does step 6 ship as an approximation?** — **decided: yes**, per rewrite ruling 6 ("the rewrites write around the rest, naming the issue"). Step 6 ships with the sentence naming #614.
5. **The log read has no houseCARL surface** — **decided for now: prose step**, and the gap is **left for Aaron**. Step 8 makes reading `po3_KeywordItemDistributor.log` a step and says plainly that no houseCARL tool reads it. Ruling 6 filed no issue for it and this PR files none.
6. **Does `_CORPUS_STATUS.md` stay in `references/`?** — **left for Aaron.** It stays where it is with corrections 3–5 applied, and nothing points at it any more, which is the arrangement ruling 2 anticipates. Moving it out of the skill tree is still open.
7. **Confirm the description trade, 383 → 441** — **decided: taken.** Measured at 441 characters, `name` + `description` at 454, both far inside every cap.

## The verifier's five builder items

- **B5** — the two tool names are written **bare** as ruling 1 requires, so **D4.42 is a declared carve-out on the regrade gate**, the way D1.28 is. It is not claimed to "not arise": once the body names a tool, the criterion is scored, and bare names do not meet its conform condition.
- **D7** — decided in favour of the D4.3 / D5.45 clear: the "Targeting NPCs" common-mistake bullet is **dropped**, so the routing rule now sits at two sites (the description clause and the one lane sentence), which is what F1's "stays in the skill" list admits. Seven mistakes are kept by promoting a non-duplicative one in its place — putting the file under `Data\SKSE\Plugins\`, the asymmetry the audit calls "not guessable".
- **I4 / I5** — **D2.35 is removed** from the must-clear list (it conforms today and the `compatibility` line keeps it conforming; ruling 7 is why it is added, not a partial). **D4.55 is placed and closed** by one line in the body: the user's own instructions outrank anything in the skill. **D5.26 is placed** in must-clear, closed by the output-quality arm's control. **D5.49 is demoted to recorded-not-converted**: no enforcement route exists for a plugin-shipped skill's always-load claim on either host, and inventing one would be a process rule rather than a fix.
- **G2** — see open question 2; "ignored" is left standing.
- **D1** — the body budget is restated as measured, not as the document's arithmetic: **161 lines / 9,762 bytes / ≈2,440 tokens**, against the corrected ~11,700-byte figure and D3.6's 5,000-token ceiling. The ≤150-line budget is exceeded by 11 lines, deliberately: section 5 carries **three** delimited intent→line pairs rather than one, because D4.50's conform condition is "three to five" and the verifier's D8 caveat says one in the body meets it only by counting pairs in files the body invites the reader to skip.

Two other judgement calls, recorded so a reviewer can push back:

- **`grammar-core.md`'s own routing note (§9) is left in place.** D4.3 counts it as a fourth site, but F1's ledger names only the three body sites and FOLD-IN's rule is that no writer folds what the ledger does not list. A reference file opened deliberately is a different audience from the always-loaded body.
- **The packager's pointer scan does not exist yet** on `origin/HEAD` — ruling 2 puts it in the wave 0 packager PR, which has not landed, and this PR does not touch the packager. The full `scripts/build-plugin.ps1` build ran and its leak check is clean, and the pointer invariant was checked by hand against the packaged tree instead: the shipped skill holds exactly six files, the body names exactly six, each once, one hop, and none of them is a packager-excluded file.

## Checks

`scripts/build-plugin.ps1` full build: leak-check clean, 11 skills bundled, zip packed. `dotnet build housecarl.sln -c Release`: 0 errors. `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"`: 1793 passed, 0 failed. Generator `ci-all`: 128/128 passed.

## Regrade

Regraded on this branch against the five checklists; the roll-up is
`dev/skill-validation/eval-2026-09/kid-authoring/REGRADE.md`, the per-angle files are under
`kid-authoring/regrade/`.

|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 38 | 0 | 1 | 3 |
| D3 — loading decision and invocation control | 33 | 28 | 0 | 1 | 4 |
| D4 — body, supporting files, instruction style | 64 | 38 | 3 | 2 | 21 |
| D5 — validation and standing context | 50 | 35 | 10 | 0 | 5 |
| **Total** | **241** | **179** | **13** | **4** | **45** |

Fails by class: factual-or-dead-call 1 (D4.48) · packager-only 1 (D1.28) · other 2 (D3.8, D4.42).

**Verdict: FAIL**, on one criterion — **D4.48**. The body states twice that SPID's INIs live under
`Data\SKSE\Plugins\` (`SKILL.md:90-91`, `:157-158`, echoed at `references/grammar-core.md:55`), and
they do not: they are flat in `Data\`, as this skill's own `_CORPUS_STATUS.md:116-117` and the
sibling `spid-authoring` both say. That is a factual fail on a load-bearing line, and the threshold
admits none on D4; the SkyPatcher half of each sentence is right. One edit closes it — strike
"SPID's and" from `SKILL.md:91` and "SPID and" from `:157` and `grammar-core.md:55`.

The other three fails do not carry the verdict. D1.28 (release tagging) is packager-only and
excluded by ruling 8. D3.8 (the bare `name` on a description drop) is the rewrite document's own
held-open naming question, closable only by a rename. D4.42 (bare MCP tool names) is ruling 1's
deviation from a criterion it cannot meet as written, classed *other* because the name is on the
published surface exactly as the body writes it; this PR declares it a carve-out, the regrade
records that as still open (Q3) — either way it is not a factual fail and not what fails the gate.

Not fails, still outstanding: D4's three partials (D4.13, D4.16, D4.43) all land on the two
packager-excluded files, and six of D5's ten partials (D5.19, D5.21, D5.23, D5.24, D5.25, D5.26) are
one line from conforming — `evals/evals.json:388` "Not yet run."

**Regrade hold fixed on this branch.** D4.48 is closed by the one edit the grader named: "SPID's and" struck from `SKILL.md:91` and "SPID and" from `:157` and `references/grammar-core.md:55`, so each sentence now attributes `Data\SKSE\Plugins\` to SkyPatcher alone; the ledger's body byte column was re-measured to 9,717 with it (D4.43).

## Regrade

Re-run 2026-09-08 after the D4.48 fix recorded above, on all five angles; this supersedes the section
above. Roll-up: `dev/skill-validation/eval-2026-09/kid-authoring/REGRADE.md`; per-angle files under
`kid-authoring/regrade/`.

|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 38 | 0 | 1 | 3 |
| D3 — loading and invocation | 35 (33 + 2 carried) | 30 | 0 | 1 | 4 |
| D4 — body, supporting files, instruction style | 64 | 38 | 5 | 0 | 21 |
| D5 — validation and standing context | 50 | 35 | 10 | 0 | 5 |
| **Total** | **243** | **181** | **15** | **2** | **45** |

Fails by class: factual-or-dead-call **0** · packager-only **1** (D1.28) · other **1** (D3.8).

**Verdict: PASS.** D4.48 now clears — the SPID attribution is corrected and D4 returns zero fails of
any class, as does D5 and D2. The two remaining fails are exactly the two carve-outs the rewrite
document names: D1.28 (release tagging, packager-only, excluded by ruling 8) and D3.8 (the bare `name`
surviving a Claude Code description drop, held open pending open question 1 — only a rename changes
it). D4.19 stays conforming, and the six in-scope fails that had to clear all do: D1.33, D4.48, D5.13
and D5.33 conform, D5.19 and D5.21 rise to partial with both arms and the control specified.

Still open, none of them a fail: the partials blocked on `evals/evals.json:388` "Not yet run."
(D5.22, D5.25, D5.52, and D5.19/D5.21/D5.26 with them), D5.49's absent enforcement route, D5.29's
`modding-basics` menu fiction at `evals/evals.json:274`, and D4's five partials, four of which land on
the two packager-excluded files.


.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,171 @@
...
+**Tag every dagger in the load order.** The trait, not the name, is the filter:
+
+```ini
+Keyword = HC_AuditDaggerTag|Weapon|NONE|OneHandDagger
+```
+
+Grounded on a 3,801-plugin order this reaches 786 WEAP records across 54 plugins. A `*Dagger` name
+filter is wrong in **both** directions on the same order: it misses `REQ_Artifact_Keening`,
+`REQ_Artifact_MehrunesRazor`, `REQ_Artifact_Nettlebane`, `BSKHatchet`, `zzzCrbAkaviriKodachi` and
+`BPUFXelzazKukri`, and it catches `DummyDagger`, `CWDummyDaggerSons`, `CWDummyDaggerImperial` (all
```
Avick3110: **The worked example's "catches" list is wrong for a name filter.**

`DummyDagger`, `CWDummyDaggerSons` and `CWDummyDaggerImperial` have no display name at all. Re-read on the same order this example cites (3,801 plugins, epoch `e2-f736766b42e4f032`):

```
editorid=DummyDagger            Name = (absent)  Data.AnimationType = OneHandSword
editorid=CWDummyDaggerSons      Name = (absent)  Data.AnimationType = OneHandSword
editorid=CWDummyDaggerImperial  Name = (absent)  Data.AnimationType = OneHandSword
```

A KID String filter matches the item **name** (`references/filters.md` section 1 - name / archetype / actor value / nif path; EditorID is a section 2 Form filter), which step 6 at `:83-84` states outright. So `*Dagger` catches none of these three. The only real false positives on that list are `DBMTWR_RiftenDaggerDummy` ("The Dagger of Riften (2)", HandToHand) and `zzzGHCrSkavenDaggers` ("Skaven Daggers", TwoHandSword) - both verified.

The evidence was taken from an EditorID scan and presented as a name-filter result: the exact conflation step 6 warns against, in the one worked example that demonstrates the new grounding discipline. A reader who trusts it believes a `*Dagger` line tags three nameless dummy weapons and re-cuts a filter that was already correct. The misses arm checks out (`REQ_Artifact_Keening` = "Keening", `BSKHatchet` = "Elven Hatchet", `BPUFXelzazKukri` = "Xelzaz's Saxhleel Kukri", all OneHandDagger).

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,171 @@
...
-guessed token yields a `_KID.ini` that quietly adds nothing — and nothing in-game points at the cause.
-A clear "that isn't in the KID reference" beats a confident wrong line that costs a debugging session.
-If the user is on a KID newer than v3.5.0 and asks about a feature not in the reference, surface that
-the reference may be behind and offer to re-derive from the current KID source/description.
+1. **Identify the keyword.** Reference it by EditorID (`WeapMaterialDwarven`) or FormID
+   (`0x12345~Plugin.esp`). An EditorID resolving to nothing is not an error: KID **creates the
+   keyword at runtime**, so a custom tag like `MyCursedGear` is valid. A runtime keyword has no
+   persistent FormID, so when other tooling must point at the tag, author the record instead —
+   `housecarl_create` with `records=[{"record_type": "Keyword", "editorid": "MyCursedGear"}]` and
+   `patch` naming the new plugin — and use the FormID the call reports back.
```
Avick3110: **The FormID `housecarl_create` reports is not KID syntax, and the mismatch fails silently.**

houseCARL reports a FormID as `XXXXXX:Plugin.esp` - colon form. `housecarl_records`' own description opens with that spelling, and a real read prints `formid=161D72:LegacyoftheDragonborn.esm`. KID's keyword section takes the tilde form with a `0x` prefix and leading zeros stripped - `0x12345~Plugin.esp`, as this same step writes at `:48-49` and `references/grammar-core.md` section 5 documents.

Step 1 ends "use the FormID the call reports back" with no conversion. Pasted as reported, `Keyword = 000800:MyPatch.esp|Weapon|...` contains no `~`, so KID reads the whole token as an EditorID, fails to resolve it, and - per grammar-core section 5, "a typo'd keyword name never errors: KID just creates a new, empty keyword instead" - silently creates a runtime keyword literally named `000800:MyPatch.esp`. The KYWD record just authored is never used, and nothing in game or in the log points at it.

One clause fixes it: convert `XXXXXX:Plugin.esp` to `0xXXXXX~Plugin.esp` before it goes in the line.

.claude/skills/kid-authoring/evals/evals.json
```diff
@@ -0,0 +1,389 @@
...
+   "prior_evidence": "2026-07-28 run under the compressed descriptions: specificity row, routed to SkyPatcher"
+  },
+  {
+   "id": "Q1",
+   "prompt": "Tag every dagger in this load order with a new keyword, using KID. Write the _KID.ini, and tell me what the line actually reaches.",
+   "should_trigger": true,
+   "kind": "output quality",
+   "expected_output": "A _KID.ini whose line selects on the WEAP animation-type trait, named with the _KID substring and placed flat in Data\\, plus a reached count backed by the housecarl_records call that produced it.",
+   "expectations": [
+    "the filter axis is the OneHandDagger trait, not a *Dagger name substring. Measured on the 3,801-plugin ARR order the name filter is wrong in both directions: it misses REQ_Artifact_Keening, REQ_Artifact_MehrunesRazor, REQ_Artifact_Nettlebane, BSKHatchet, zzzCrbAkaviriKodachi and BPUFXelzazKukri, and it catches DummyDagger, CWDummyDaggerSons and CWDummyDaggerImperial (OneHandSword), DBMTWR_RiftenDaggerDummy (HandToHand) and zzzGHCrSkavenDaggers (TwoHandSword)",
```
Avick3110: **This expectation carries the same wrong claim as `SKILL.md:113`, so it will fail a correct answer.**

`DummyDagger`, `CWDummyDaggerSons` and `CWDummyDaggerImperial` have `Name = (absent)` on the order this case names, and KID's String filter matches the item name, not the EditorID. A `*Dagger` name filter catches none of the three.

Because this is a grading expectation rather than prose, the cost is worse than in the body: an agent that grounds the comparison correctly - reporting that only `DBMTWR_RiftenDaggerDummy` and `zzzGHCrSkavenDaggers` are name-filter false positives - is marked wrong by Q1, and the eval rewards reproducing the error. Fix both sites together.

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,171 @@
...
+   winner is the record KID acts on. To show what a name filter would have caught instead, run the
+   same scan with `where=["editorid contains Dagger"]`, noting that KID's own String filter matches
+   the item's **display name**, not its EditorID, so that scan illustrates rather than repeats the
+   predicate. **This grounds one type-plus-trait predicate, not the whole line:** nothing on the 2.0
+   surface replays KID's evaluation order or `chance` — that is issue **#614**, not in 2.0, so say
+   so rather than implying the whole line was proved.
+
+7. **Place the file** at `Data\<name>_KID.ini`. The `_KID` substring is **mandatory** — a file
+   without it is never read — and the file sits flat in `Data\`, not under `Data\SKSE\Plugins\`
+   where SkyPatcher's INIs live. KID INIs have **no `[Section]` headers**; every line
```
Avick3110: **`Data\SKSE\Plugins\` is not where SkyPatcher reads.**

SkyPatcher scans `Data\SKSE\Plugins\SkyPatcher\<type folder>\`. The sibling `skypatcher-authoring/SKILL.md:65` places INIs at `Data/SKSE/Plugins/SkyPatcher/<subfolder>/<file>.ini`, and `housecarl_skypatcher_layer`'s own description says it "Scans Data\SKSE\Plugins\SkyPatcher exactly as the DLL reads it".

The negative claim about KID holds either way, but the attribution is repeated as positive guidance in the common-mistake bullet at `:157` ("That is where SkyPatcher reads from"). A reader acting on that puts a SkyPatcher INI two levels too high, where SkyPatcher never reads it and nothing reports the miss - the same silent-failure class this skill exists to prevent. This is also the same class as the D4.48 fail the regrade closed: striking "SPID's and" fixed who the folder belongs to without fixing the folder.

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,171 @@
...
+   ```
+   Sections are positional. `Keyword = MyKwd|Book|NONE|S,20` puts `S,20` in *traits*;
+   `Keyword = MyKwd|Armor|||50` puts `50` in *chance*. Leave an unused middle section blank or
+   `NONE`; a trailing unused section can be dropped. Set `chance` (0–100, default 100) only when you
+   want less than guaranteed.
+
+6. **Ground the line.** Measure the reach before the file ships, and report the number with the call
+   that produced it rather than as a claim. The census is `housecarl_records` with `types=["WEAP"]`,
+   `where=["Data.AnimationType = OneHandDagger"]` and `counts_only=true`. To let the user eyeball
+   the set, drop `counts_only` and pass `project` with form `identity` plus a `limit`; for a set too
```
Avick3110: **`project` form `identity` is the wrong form for "let the user eyeball the set", and the surface says so.**

From `housecarl_records`' `limit` documentation: form `identity`'s row "is an UNTYPED whole-plugin seek for the winner's body and is the dearest row here rather than a free one: form='summary' answers the same identity question off a gathered read". Its render bound is 40,000 rows against 300,000 for a named-fields row.

So the step teaches the most expensive projection for exactly the case the surface routes to the cheapest. The bite is on a broad predicate rather than the 786-record example: grounding a whole-type census (`types=["ARMO"]` with a loose trait) puts the reader on identity's per-row whole-plugin seek and can refuse up front at the 40,000-row bound, where `summary`, or `fields` naming EditorID and Name, returns the same eyeball list off a gathered read. Recent commits on `main` ("Bound form='identity', which reads the dearest body on the lane") are pinning this exact point.

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,171 @@
...
+live load order so its reach is measured before the file ships, and **places** the file where KID
+will read it. Every type, filter, trait and value comes from `references/`, never from memory — a
+fabricated token fails silently.
+
+## This skill's lane
+
+KID's lane is **a keyword onto an item record**. A keyword onto an NPC is SPID: load
+`housecarl:spid-authoring` for that job instead. Sibling skills are written `housecarl:<name>`, the
+Claude Code plugin form; a Codex install sees the same skill under its bare folder name
+(`spid-authoring`). MCP tool names are bare — `housecarl_records` — on every host.
```
Avick3110, line 32: **The CID and SkyPatcher lanes leave the body on a premise that does not hold.**

The PR body says the four-way distributor routing table "leaves the body for ServerInstructions (F1)". `src/housecarl-mcp/Program.cs:131-154` is the whole of `ServerInstructions`, and it carries no distributor routing - no KID, no SPID, no CID, no SkyPatcher lane. This PR touches no C#, so nothing added it.

What left the always-loaded body is the old `Scope boundary` list, the `Cross-tool routing` note, and the `Targeting NPCs` mistake bullet; what replaced them names SPID only. Concretely: "distribute this potion into every bandit chest with a runtime INI" loads this skill on items-plus-distribution, and nothing in the body says containers are CID - the composed answer is a keyword line that tags the potion instead. Same shape for "change this weapon's damage without an ESP" (SkyPatcher).

`grammar-core.md` section 9 still carries the routing, so the loss is confined to the always-loaded surface - which is the surface the routing was for.

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,171 @@
...
+   (`Magic Effect`, not `MagicEffect`). One line targets one type; repeat the line per type.
+
+3. **Choose the filters.** By name / archetype / actor value / nif path → a String filter; by record,
+   plugin or associated form → a Form filter. Modifiers: `+` requires all, `-` excludes, `*`
+   wildcard-substring (strings only), bare = match-any; KID evaluates requirements → exclusions →
+   matches → wildcards.
+
+4. **Add traits if the narrowing is a property of the item.** Only the 12 trait-bearing types accept
+   them. Which record field a KID trait tests is the schema's answer, not KID's: resolve the field
+   with `housecarl:mutagen-reference` before composing the filter.
```
Avick3110, line 67: **This routes the trait question to a skill that cannot answer it, and de-authorizes the one that can.**

`housecarl:mutagen-reference` is the generated Mutagen schema corpus - record, struct and enum shapes. It carries nothing about KID, and this PR's own open question 3 says the trait-to-field row "lands in `mutagen-reference`'s own rewrite, not here". So today the pointer resolves to a skill with no such mapping in it.

"the schema's answer, not KID's" also cuts against the Overview at `:24` ("Every type, filter, trait and value comes from `references/`, never from memory"). What a trait *tests* is KID's answer, documented in `traits.md` behind its `[source]` markers - `BLACK` is KID reading the black-soul flag, `AR(min/max)` is KID reading armorRating. mutagen-reference is the right tool for step 6's grounding predicate, not for what the trait means.

As written, a reader composing `|Soul Gem|NONE|BLACK` opens mutagen-reference, finds `SoulGem.Soul` and `SoulGem.MaximumCapacity` with nothing saying which one `BLACK` tests, and either guesses a `where=` or falls back to `traits.md` against this step's instruction. A guessed `where=` yields a wrong census, and step 8's stop condition then uses that census to block a line that was correct.

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,171 @@
...
+
+## Worked lines, end to end
+
+**Tag every dagger in the load order.** The trait, not the name, is the filter:
+
+```ini
+Keyword = HC_AuditDaggerTag|Weapon|NONE|OneHandDagger
+```
+
+Grounded on a 3,801-plugin order this reaches 786 WEAP records across 54 plugins. A `*Dagger` name
```
Avick3110: **786 reproduces; 54 does not, from the call this skill teaches.**

Re-run on the same order and epoch (`e2-f736766b42e4f032`):

- `types=["WEAP"], where=["Data.AnimationType = OneHandDagger"], counts_only=true` -> **786 matches**, exactly as written
- the same scan with `project={"form":"aggregate","group_by":"winner"}` -> **39 groups**
- `group_by="defined_in"` -> **54 groups**

So 54 counts the plugins that *define* those records, not the plugins whose versions KID acts on, which is 39. Step 6 tells the reader the winner is the record KID acts on and to "report the number with the call that produced it rather than as a claim" - and this is the one figure in the skill with no call beside it, phrased ("reaches 786 WEAP records across 54 plugins") so it reads as the winner count it is not. Either name `defined_in` or use 39. Same figure at `evals/evals.json:370` and `:385`.

.claude/skills/kid-authoring/evals/evals.json
```diff
@@ -0,0 +1,389 @@
...
+    "without-skill arm: removing kid-authoring does not change where this query lands"
+   ],
+   "why_this_case": "hard near-miss — names KID but is install",
+   "prior_evidence": "2026-07-28 run under the compressed descriptions: specificity row, routed to modding-basics"
+  },
+  {
+   "id": "T20",
+   "prompt": "i'm new to modding, what even is KID and do i need it for my load order",
+   "should_trigger": false,
+   "expected_output": "kid-authoring is NOT loaded; the query routes to modding-basics.",
```
Avick3110: **T19 and T20 cannot pass as specified.**

Both `expected_output`s say the query "routes to modding-basics", but neither declared menu contains it: `method.run_configurations` offers `isolation` ("kid-authoring alone plus NONE") and `coexistence` ("kid-authoring among its distributor-cluster siblings plus NONE"). There is also no `modding-basics` skill in `.claude/skills/` - it was a synthetic menu letter in the deleted `eval_set.json`'s 2026-06 fan-out.

In the isolation configuration the agent's only choices are kid-authoring and NONE, so the stated expected output is unreachable and the case is unfalsifiable; in coexistence it is unreachable too. The PR body records this as D5.29 still open - flagging it so it does not survive into the run. Naming NONE (or a sibling) as the destination makes both cases runnable.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `a4b71a24`, together with the eval site.

Re-read on the same order and epoch (`e2-f736766b42e4f032`) confirms it: `DummyDagger`, `CWDummyDaggerSons` and `CWDummyDaggerImperial` all read `Name = (absent)`, and the scan that actually asks the name-filter question — `types=["WEAP"]`, `where=["Name contains Dagger", "Data.AnimationType != OneHandDagger"]` — returns exactly two records, `DBMTWR_RiftenDaggerDummy` ("The Dagger of Riften (2)", HandToHand) and `zzzGHCrSkavenDaggers` ("Skaven Daggers", TwoHandSword). The six misses re-read as OneHandDagger with no "dagger" in the name.

The paragraph now names that `where=` beside the claim, and the nameless dummies are kept as the illustration of the conflation itself: an EditorID scan turns them up, a String filter never sees them.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `37309b59`. Step 1 now ends: convert the FormID the call reports, `XXXXXX:Plugin.esp`, to KID's tilde form `0xXXXXX~Plugin.esp` (leading zeros stripped) before it goes in the line — pasted as reported it reads as an EditorID and quietly creates an empty runtime keyword. That is the silent-failure class the step exists to close, and `grammar-core.md` section 5 is the authority for the tilde form.

.claude/skills/kid-authoring/evals/evals.json
Avick3110: Folded, in `a4b71a24`, at the same time as `SKILL.md`. The Q1 expectation now says KID's String filter matches the item NAME, names the two real false positives with their names and animation types, and states outright that an answer listing the three nameless dummies has read an EditorID scan as a name-filter result and is wrong — so the case grades the correct answer correct. The count expectation in the same block became "786 records, defined across 54 plugins, won by 39".

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `c0096be2`. Both body sites and `grammar-core.md` section 2 now name `Data/SKSE/Plugins/SkyPatcher/<type folder>/`, matching `skypatcher-authoring/SKILL.md:65` and `housecarl_skypatcher_layer`'s own description. The KID claim is unchanged — flat in `Data/` — but the mistake bullet no longer sends a SkyPatcher INI two levels too high.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `37e3c794`, and it is worse than dear — it is refused. Run against the live order:

```
housecarl_records types=["WEAP"] where=["Data.AnimationType = OneHandDagger"] project={"form":"identity"} limit=3
-> error: the identity form labels a formids= list; a scan's summary rows already carry each match's
   identity - use form='summary' (the default).
```

So the step taught a dead call. It now reads: drop `counts_only` and pass a `limit` — the default `summary` rows carry each match's identity, and form `fields` with `fields=["Name"]` puts the name beside it (form `identity` labels a `formids=` list, and a scan refuses it).

The rewrite document's migration row 2 does name form `identity`, but its own stated constraint is that every call appear in `surface/tools.md` as written, and `tools.md:1438` is the sentence you quote. The surface wins.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Refuted — the rewrite document settles the placement, and this PR is not the destination.

`REWRITE.md` fold-out **F1** names exactly what leaves the body ("`SKILL.md:22-29` … with its four-way item / NPC / container / fields table", restated at `:104-106` and `:132-134`) and exactly what stays ("the one-clause statement of **this skill's own lane**, the description's 'Keywords onto NPCs are SPID, not KID', and one sibling pointer to `housecarl:spid-authoring`"). The body carries that list and nothing else, which is what the ledger asks for. `FOLD-IN.md` §6 holds the drafted ServerInstructions string in full, including the CID and SkyPatcher clauses; appending it is the fold-in change, which lands in its own PR, and ruling 4 admits only the three ServerInstructions folds — no writer folds what the ledger does not list, and no writer keeps in the body what the ledger folds out.

So you are right that the destination has not landed yet on `origin/HEAD`; that is a sequencing fact about the wave, the same shape as the packager's pointer scan recorded in the PR body, not a defect in this skill. The routing also still stands in `grammar-core.md` section 9 in the meantime, which the PR body already records as a judgement call.

Flagging the window itself is fair — worth Aaron's call on whether wave 1 should wait on the ServerInstructions PR before merge.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Refuted — the sentence is fixed by the ledger, verbatim.

`REWRITE.md` fold-out **F16** gives it word for word as "the sentence that stays in kid-authoring": "Which record field a KID trait tests is the schema's answer, not KID's: resolve the field with `housecarl:mutagen-reference` before composing the filter." The body carries it unchanged. That it does not resolve to a row today is the ledger's own open question 3 (`FOLD-IN.md` Q7), decided by Aaron on 2026-09-08: the row is hand-written beside the generated shards and lands in `mutagen-reference`'s own rewrite, not here — so nothing outside this skill is touched by this PR, by design.

On the second half: what a trait *means* is not routed away. Section 3 of the body sends exactly that question to `references/traits.md` ("When the narrowing is a property of the item rather than its name — armor rating, animation type, soul size, effect school — read `references/traits.md`"), and step 6 only needs the field for the census predicate. `BLACK` is looked up in `traits.md` like every other token; `housecarl:mutagen-reference` is asked only which field the resulting `where=` reads.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `a4b71a24`. Re-run confirms both figures: `group_by="winner"` gives 786 matches across 39 groups, `group_by="defined_in"` gives 786 across 54. The sentence now reads "786 WEAP records, defined across 54 plugins and won by 39 (`project` form `aggregate`, `group_by` `defined_in` and `winner`)", so the one figure that had no call beside it now names both. Same correction at `evals/evals.json:370`; the `prior_evidence` line at `:385` is left as the audit's own wording, since it records what that run reported.

.claude/skills/kid-authoring/evals/evals.json
Avick3110: Folded, in `0983b548`. Both `expected_output`s now read "kid-authoring is NOT loaded; the pick is NONE on both declared menus", T19's adding "neither of which carries an install or orientation skill". That is reachable in isolation and in coexistence, so both cases are runnable and falsifiable. The `prior_evidence` / `prior_result` lines keep the `modding-basics` wording, because they record what the 2026-07-28 run did under that synthetic menu rather than specifying this one.

Avick3110:
## Review fold

Seven of the nine findings folded, two refuted; each has its reply on the thread.

**Folded**

| Finding | Fix | Commit |
|---|---|---|
| The worked example's "catches" list is wrong for a name filter | The three dummy daggers read `Name = (absent)`, so a String filter never sees them; the real false positives are the two returned by `where=["Name contains Dagger", "Data.AnimationType != OneHandDagger"]`. Same fix at the eval site | `a4b71a24` |
| "786 across 54 plugins" reads as a winner count | Now "defined across 54 plugins and won by 39", with the `aggregate` / `group_by` calls named. Re-run: 39 winner groups, 54 defined_in groups | `a4b71a24` |
| The reported FormID is not KID syntax | Step 1 converts `XXXXXX:Plugin.esp` to `0xXXXXX~Plugin.esp` before the line, and says what happens if it does not | `37309b59` |
| `Data/SKSE/Plugins/` is not where SkyPatcher reads | Both body sites and `grammar-core.md` section 2 name `SkyPatcher/<type folder>/` | `c0096be2` |
| `project` form `identity` on a scan | It is refused outright, not merely dear - the step taught a dead call. Now `summary` rows plus a `limit`, with `fields` for the name | `37e3c794` |
| T19 / T20 expect a `modding-basics` that is on no menu | Both now name NONE as the destination, so the cases are runnable | `0983b548` |
| (ledger) body figures moved with the edits | Body re-measured: **169 lines / 10,512 bytes**, `SKILL.md` 179 lines; description unchanged at 441 | `02791a98` |

**Refuted**

- **The CID / SkyPatcher lanes leaving the body.** `REWRITE.md` F1 lists exactly what leaves and exactly what stays, and `FOLD-IN.md` section 6 holds the ServerInstructions string that receives it. That the destination has not landed on the base branch yet is wave sequencing - the same shape as the packager's pointer scan already recorded above - not a defect here. Worth Aaron's call on whether wave 1 waits on that PR.
- **The `housecarl:mutagen-reference` pointer.** F16 fixes that sentence verbatim, and where the row physically lives is open question 3, decided on 2026-09-08 for `mutagen-reference`'s own rewrite. What a trait *means* is still routed to `references/traits.md` by section 3; only the field path goes to the schema.

**Checks after the fold.** `scripts/build-plugin.ps1`: leak-check clean (12 skills, 250 markdown files), zip packed. `dotnet build housecarl.sln -c Release`: 0 errors. `dotnet test --filter "tier!=bridge"`: 1793 passed, 0 failed. Generator `ci-all`: 128/128 pass.

The regrade files under `kid-authoring/regrade/` are left as the record of that run; the body-figure row they cite is the ledger's, which is updated here.

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,179 @@
...
+   that produced it rather than as a claim. The census is `housecarl_records` with `types=["WEAP"]`,
+   `where=["Data.AnimationType = OneHandDagger"]` and `counts_only=true`. To let the user eyeball
+   the set, drop `counts_only` and pass a `limit` — the default `summary` rows carry each match's
+   identity, and form `fields` with `fields=["Name"]` puts the name beside it (form `identity`
+   labels a `formids=` list, and a scan refuses it). For a set too
+   large to render inline, pass `to_file` an absolute `.jsonl` path and re-enter it later as
+   `formids=["@<that absolute path>"]`. Leave `source` omitted, or pass `"winner"` — the load-order
+   winner is the record KID acts on. To show what a name filter would have caught instead, run the
+   same scan with `where=["editorid contains Dagger"]`, noting that KID's own String filter matches
+   the item's **display name**, not its EditorID, so that scan illustrates rather than repeats the
```
Avick3110: **Step 6 prescribes the wrong predicate for the name-filter comparison.**

The sentence opens by saying it shows "what a name filter would have caught", tells the reader to scan `editorid contains Dagger`, then closes by conceding that KID's String filter matches the display name — so the call it prescribes does not answer the question it was introduced to answer. The supported predicate that does is `Name contains`, which the worked example at :119 uses (`where=["Name contains Dagger", "Data.AnimationType != OneHandDagger"]`).

Concrete failure: an agent follows step 6 verbatim, runs the editorid scan, and reports its hits — `DummyDagger`, `CWDummyDaggerSons`, `CWDummyDaggerImperial`, which have no Name at all — as what a `*Dagger` name filter would catch. That is exactly the answer this skill's own `evals/evals.json` Q1 grades wrong: "an answer that lists them as name-filter catches has read an EditorID scan as a name-filter result and is wrong". Step 6 is the always-loaded procedure; the correct call sits only in a worked example the reader is told to skim.

Suggest running `Name contains` here as the comparison scan, and keeping the `editorid contains` scan (with its caveat) as the separate illustration it already is at :121-122.

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,179 @@
...
+`zzzCrbAkaviriKodachi` and `BPUFXelzazKukri`, whose names carry no "dagger"; and by
+`where=["Name contains Dagger", "Data.AnimationType != OneHandDagger"]` it catches exactly two —
+`DBMTWR_RiftenDaggerDummy` ("The Dagger of Riften (2)", `HandToHand`) and `zzzGHCrSkavenDaggers`
+("Skaven Daggers", `TwoHandSword`). The `DummyDagger` records an EditorID scan turns up have no name
+at all, so a String filter never sees them: that scan illustrates the predicate, it does not repeat it.
+
+**Tag one mod's armor above rating 20**, filters and traits both used:
+
+```ini
+Keyword = MyLightSetTag|Armor|MyArmorMod.esp|AR(20/100)
```
Avick3110: **The caption says "above rating 20"; the line writes a closed 20-100 range.**

`references/traits.md:14` gives the idiom for an open-ended minimum: "One number — `X(10)` — sets only the minimum." `AR(20/100)` sets both ends, so anything rated above the upper bound is excluded.

Concrete failure: a user copies this example to tag every piece in their mod above rating 20. Pieces over the cap are silently skipped — no parse error, nothing in `po3_KeywordItemDistributor.log`, the keyword just never lands. This is the shipped model for the `AR()` trait, so the wrong idiom propagates to every armor line composed from it.

```suggestion
Keyword = MyLightSetTag|Armor|MyArmorMod.esp|AR(20)
```

.claude/skills/kid-authoring/SKILL.md
```diff
@@ -1,136 +1,179 @@
...
+   `Keyword = MyKwd|Armor|||50` puts `50` in *chance*. Leave an unused middle section blank or
+   `NONE`; a trailing unused section can be dropped. Set `chance` (0–100, default 100) only when you
+   want less than guaranteed.
+
+6. **Ground the line.** Measure the reach before the file ships, and report the number with the call
+   that produced it rather than as a claim. The census is `housecarl_records` with `types=["WEAP"]`,
+   `where=["Data.AnimationType = OneHandDagger"]` and `counts_only=true`. To let the user eyeball
+   the set, drop `counts_only` and pass a `limit` — the default `summary` rows carry each match's
+   identity, and form `fields` with `fields=["Name"]` puts the name beside it (form `identity`
+   labels a `formids=` list, and a scan refuses it). For a set too
```
Avick3110: **`fields` is a `project` sub-parameter, but it is written in the flat notation this sentence uses for every real top-level parameter.**

The same paragraph writes `types=`, `where=`, `counts_only=true`, `limit`, `to_file`, `source` and `formids=` — all genuine top-level parameters — and then `fields=["Name"]` beside them. The spec is explicit that sub-parameters live inside the form: `project={"form": "fields", "fields": ["Name"]}`.

Concrete failure: an agent passes a flat `fields=["Name"]` and the call is refused — `src/housecarl-mcp/AliasTable.cs:145`, "not a parameter here — field paths live inside the form". It costs a round trip on the one step this rewrite exists to add rather than producing a wrong answer, so it is minor, but writing the nested spelling once here would remove the trap. (Worth noting the parenthetical at :114-115 has the same shape for `group_by`.)

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `97e70ba1`.

Step 6 now runs `where=["Name contains Dagger"]` as the comparison scan and says why — KID's String filter matches the display name, so `Name` is the predicate that answers the question the sentence opens with. The `editorid contains` scan stays named as the different question it is ("turns up records with no name at all, which no String filter ever sees"), which is the illustration the worked example already carries at `SKILL.md:124-125`.

`contains` is a general field operator, not an editorid-only term — `src/housecarl-core/FieldPredicate.cs:294` lists it beside `=`, `startswith` and the rest, and `tools.md:1384` documents body-field comparisons. The rewrite document's migration row 5 does name `editorid contains`; the row's own constraint is that the call be expressible on the surface, and `Name contains` is, so the body no longer prescribes a call whose caveat cancels it.

Current lines: `SKILL.md:87-90`.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `2be6339a`, as suggested: the line is now `Keyword = MyLightSetTag|Armor|MyArmorMod.esp|AR(20)`.

`references/traits.md:14` is the authority — one number sets only the minimum — and the caption says "above rating 20", so the closed `AR(20/100)` was silently excluding every piece over the cap, with nothing in `po3_KeywordItemDistributor.log` to show for it. This was the skill's only `AR()` model line, so nothing else propagated it.

Current line: `SKILL.md:129`.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Folded, in `1675f28f`, both sites.

Step 6 now writes the whole object — `project={"form": "fields", "fields": ["Name"]}` — and adds the scoping rule in four words: field paths live inside the form, never flat. That is the refusal text itself (`src/housecarl-mcp/AliasTable.cs:145`), so an agent that reads the sentence cannot spend the round trip.

The parenthetical you flagged is fixed the same way: the worked example's aggregate now reads `project={"form": "aggregate", "group_by": "defined_in"}`, then `"winner"` — `group_by` is gated on `project` in the same table (`AliasTable.cs:147`).

Current lines: `SKILL.md:82-84` and `:117`. The ledger's measured body figures were re-taken after the three folds (`2a1d97d9`): 181 lines total, body 171 lines / 10,664 bytes.

Issue #294, opened 2026-07-28
Skill descriptions cost ~3,040 est. tokens of always-resident context (~1.5x the whole listing budget)


Avick3110:
## Problem

The 13 skill descriptions in this repo cost **~12,150 characters (~3,040 est. tokens)** of always-resident context, in every session, whether or not any skill is invoked.

Claude Code budgets roughly **1% of the context window (~2,000 tokens)** for the *entire* skill listing across all sources. houseCARL alone is ~1.5× that budget. Measured on a real setup (2026-07-28), the full listing came to ~8,300 est. tokens — ~4× budget — with houseCARL the single largest contributor.

When the listing exceeds its budget, entries get **truncated**. That's the part that actually hurts: a truncated description routes worse than a short one, so over-long descriptions don't just cost tokens, they degrade the routing they were written to improve.

## Current state

Mean description length is **935 characters**. Ten of thirteen are over 800.

| Skill | Description chars | Est. tokens |
|---|---:|---:|
| `dialogue-authoring` | 1081 | ~270 |
| `skypatcher-authoring` | 1016 | ~254 |
| `biped-slot-reference` | 1003 | ~251 |
| `oar-authoring` | 1000 | ~250 |
| `kid-authoring` | 997 | ~249 |
| `bulk-record-jobs` | 970 | ~243 |
| `skse-plugin-authoring` | 956 | ~239 |
| `spid-authoring` | 949 | ~237 |
| `mutagen-reference` | 884 | ~221 |
| `papyrus-reference` | 854 | ~214 |
| `tool-output-awareness` | 840 | ~210 |
| `facegen-diagnostics` | 824 | ~206 |
| `papyrus-optimization` | 776 | ~194 |
| **Total** | **12150** | **~3038** |

## What's eating the characters

Reading through them, the bulk is three patterns:

1. **Exhaustive trigger enumeration** — long "Use when the user wants to A, B, C, D, E, F, G, H, or I" chains, where the listed triggers are near-synonyms that a model already generalises across.
2. **Capability inventories** — the description restating what the skill *contains* (which filters, which record types, which tools) rather than when to reach for it. That belongs in the body, which loads on invocation and is free.
3. **The "load BEFORE, not after" coda** — a useful instruction, but it's currently spelled out at length in most descriptions. It compresses to a short clause.

## Suggested target

[...]

PR #634, opened 2026-09-08
Rewrite the spid-authoring skill on the approved document

Avick3110:
Wave 1 of the skill rewrite, executing the approved document for `spid-authoring` with the five rows its verifier overturned corrected as the verifier says. The substantive change is a factual one: the skill said a Level Filter moves a line into a leveled pass that only auto-levelled NPCs receive, which is wrong and understated a real line's reach by a factor of hundreds on the audited order. SPID checks the filter against every matching NPC's own level; the separate pass is a re-distribution restricted to NPCs on a PC level multiplier, so that a scaling actor is re-evaluated as the player levels. The corrected rule now lives once, in `references/filters.md` §3, with its citations into SPID's MIT source, and the body points at it instead of restating it. Around that: the description is rewritten to the document's draft (612 characters, third-person "what" plus an imperative "Use when", the load-timing imperative ahead of the trigger list, a negative clause naming KID/CID/SkyPatcher), `compatibility` is added, the four-way routing table collapses to one routing sentence now that ServerInstructions carries the rule, siblings are written `housecarl:<name>` with one line noting the Codex bare-folder form, two worked lines are added where the body carried no example, and a new section names the three read-only checks that actually exist — `housecarl_records` with `counts_only=true` for the population a filter names and for how many of those are on a PC level multiplier, and `housecarl_asset_status` for which mod wins the placed `_DISTR.ini` — while saying plainly that nothing previews a composed line's reach (#614) and nothing parses a draft `_DISTR.ini` offline. The version window becomes two-directional and names the call that reads the installed version, the EditorID caveat about a renaming winner lands in `grammar-core.md` §6 and one line of the workflow, the three reference files over 100 lines gain a section list, and `evals/eval_set.json` becomes `evals/evals.json` on the published field names, with a without-skill arm, three runs per query, per-row evidence, a Haiku pass, an isolation configuration, and the audit's own task quoted verbatim as the output-quality arm.

No issue closes with this: the document names none, and no wave-1 issue exists for this skill.

**1. Which SPID source settles the Level Filter — decided as follows.** The repository publishes no 7.0.0 and no 7.3.0 tag; its tags stop at `v6.8.5.rc9`, so neither version named in the question is readable as a tag. I read `master` and re-checked at `v6.8.5.rc9`, which brackets the 7.x window, and the logic is identical at both: `DistributeManager.cpp`'s `detail::distribute_on_load` calls `Distribute(npcData, false)` for every NPC passing `should_process_NPC` (`!IsPlayer() && !IsDeleted()`); `FormData.h`'s `Distributables<Form>::GetForms(bool)` returns the full `forms` list on `false` and the `formsWithLevels` subset only on `true`; `LookupFilters.cpp`'s `Data::passed_level_filters` tests the range against `NPC::Data::GetLevel()`; and `DistributePCLevelMult.cpp` gates its hooks on `npc->HasPCLevelMult()`. That is the whole answer, and it is cited in `references/filters.md` §3 so the next reader does not re-derive it.

**2. Whether the body may teach the `housecarl_skse` version read — decided as follows: yes.** The fold stays refused (ruling 4 bars a tool description, and this is not one of the three ServerInstructions folds), but the check is runnable and the body teaches it with the tool's own ceiling attached — what the file declares, never what the DLL does. Overrule and I will cut it back to pure prose.

**3. An offline SPID draft check — left for Aaron.** No issue is filed; #613 is SkyPatcher-scoped and #614 is the reach question, not a parse check. The body writes around the gap and names it, per ruling 6. Filing a SPID sibling of #613 is a ruling-6 call, so I did not make it.

**4. The opaque bare name (D3.8) — decided as the document's default:** the name stays `spid-authoring` on family consistency with `kid-authoring`, `skypatcher-authoring` and `oar-authoring`, and D3.8 stays partial by design. Nothing here forecloses a rename later.

**5. `references/_CORPUS_STATUS.md`'s fate — decided as follows:** it stays in the source tree as an unreferenced dev-side ledger, the shape F42 rules for `kid-authoring`. The packager already strips it, the body never points at it, and its one durable sentence — the corpus provenance — is inlined in Notes.

## The verifier's five corrections, as applied

Common mistakes carries **seven** traps, not eight — the routing bullet leaves with the fold and the cell named no replacement. The output-quality arm **quotes the audit's task verbatim**, including "never unique NPCs" and the reach estimate, so it still exercises the `-U` trait letter that assertion 1 scores. The new conceptual/beginner eval row is recorded as an **unbiased measurement, not a predicted pass**: the description's boundary clause is a target boundary against sibling frameworks and says nothing about beginner questions, which is the recorded diagnosis for that miss. The two document-side corrections — adding D4.48 to the D4 regrade row, and re-anchoring the `CLAUDE.md` sentence off D5.42 — change nothing that ships, but the regrade gate needs D4.48 in its D4 list or its own "zero partial" threshold is unreachable.

## Judgement calls

The body came out at 197 lines / 12.6k characters against the document's 146-line, ~10.5k-character target. Every section is present and none is padded; the overrun is bullets that wrap. Both binding limits pass with room — the 500-line target and the 5,000-token recommendation (~3.2k tokens at four characters per token, ~4.2k at three). Trim it further and the checks in §5 or the seven traps are what would go.

Ruling 2's pointer scan is not in `scripts/build-plugin.ps1` yet — it belongs to the wave 0 packager PR, which this branch does not touch. The existing leak check ran clean on the built package, and I verified by hand that every `references/<file>` the body names resolves inside `dist/housecarl/skills/spid-authoring/` and that neither `evals/` nor `_CORPUS_STATUS.md` is pointed at or shipped.

Build clean; 1793 tests pass; `ci-all` 128/128.

## Regrade

Assembled in `dev/skill-validation/eval-2026-09/spid-authoring/REGRADE.md` from the five grader files
in `regrade/`, against the regrade half of `REWRITE.md` §9a.

|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 36 | 0 | 1 | 5 |
| D3 — loading and invocation | 33 | 28 | 1 | 0 | 4 |
| D4 — body, supporting files, style | 64 | 41 | 5 | 1 | 17 |
| D5 — evidence and integration | 50 | 33 | 10 | 1 | 6 |

**Verdict: FAIL.** D2 and D3 carry zero fails; D1's only fail is D1.28, packager-only and excluded by
ruling 8. D4 and D5 each carry one fail, both classed factual-or-dead-call, and both are the same
defect: the second worked line at `SKILL.md:106` has one pipe too many, so the trait letter `F` lands
in `CountOrPackageIndex` and the line distributes to every `NordRace` NPC at 50% rather than to female
ones, while the read-back at `:109-111` numbers Chance as position 7 against the corpus's position 6.
Both graders name the same one-line fix — `Keyword = HC_FrostTouched||NordRace||F||50` with
"position 6" in the read-back — which also lifts D4.8 to conforms.

**Regrade holds fixed (312baca0).** The second worked line is now `Keyword = HC_FrostTouched||NordRace||F||50` and its read-back says Chance is position 6, on the 0-based numbering the body now states explicitly and shares with `references/`; the two `See the routing skill` pointers in `grammar-core.md` route by name to `housecarl:skypatcher-authoring` and `housecarl:kid-authoring`; and the body's `§5`/`§8` citations, against headings that carry no numbers, now name the headings. Package rebuilt (pointer scan clean), 1793 tests pass, ci-all 128/128.

## Regrade

Re-assembled from the five grader files in `regrade/` after the fixes above, and superseding the
earlier Regrade block in this body. Full write-up:
`dev/skill-validation/eval-2026-09/spid-authoring/REGRADE.md`.

|---|---:|---:|---:|---:|---:|
| D1 — discovery and packaging | 42 | 36 | 0 | 1 | 5 |
| D2 — frontmatter | 52 | 39 | 0 | 0 | 13 |
| D3 — loading and invocation | 33 | 28 | 1 | 0 | 4 |
| D4 — body, supporting files, instruction style | 64 | 43 | 3 | 1 | 17 |
| D5 — validation and standing context | 52 | 36 | 10 | 0 | 6 |

**Verdict: PASS.** Two fails in the whole regrade. D1.28 (release tagging) is packager-only, outside
the skill's files, and excluded by ruling 8. D4.42 (MCP tools written bare rather than
`ServerName:tool_name`) is classed "other" by its own grader — every tool and parameter the body names
is live on the shipped surface, so it is a naming form, not a wrong fact or a dead call; the D4
threshold bars only factual and dead-call fails. D2, D3 and D5 carry no fail of any class. The
remaining partials — D3.8 and D5.49 accepted by design, and the seven D5 rows that all wait on the
eval suite being run rather than on another edit — are recorded in `REGRADE.md`.


.claude/skills/spid-authoring/SKILL.md
```diff
@@ -10,159 +17,196 @@ description: >-
...
-  `references/filters.md` §3).
+  launch and writes nothing into the save, so actors in a pre-existing save do pick up a rule added
+  afterwards. The INIs are read **once per game launch**, though: a rule added while the game runs
+  needs a restart, and the actor has to load in that new session.
+- **A dynamic keyword IS reachable.** A keyword SPID creates at runtime has no plugin FormID —
+  nothing in xEdit to point at — but it is in the game's keyword array, so `Keyword.GetKeyword()`
+  finds it by name, which is what makes a no-ESP keyword distribution verifiable from script at all.
+  For the signature use `housecarl:papyrus-reference`; the SPID side is in `references/form-types.md`.
+- **Chance skips actors by design.** A `Chance` below 100 is *supposed* to miss some NPCs, so two
+  sampled actors without the form is an ordinary outcome for `|||||50` — and the roll is re-made
```
Avick3110: **`|||||50` is a Count of 50, not a Chance of 50 — the same pipe-count error the regrade just fixed in the worked line.**

`references/grammar-core.md` §4 is explicit: "`Item = MyItem|||||5` has five pipes, so `5` lands in position 5 (`CountOrPackageIndex`); Chance is omitted. `Form = 0x12345||||||60` has six pipes, so `60` lands in position 6 (`Chance`)." §10's own examples are all `||||||60`, `||||||0.01`, `||||||50!`. The fixed worked line at `:109` writes six pipes for the same reason.

Scenario: a reader following this bullet to diagnose "my rule only hits some NPCs" writes `Spell = MySpell|||||50` believing it is a 50% chance. Five pipes puts `50` in `CountOrPackageIndex`, Chance defaults to 100, and the line distributes to **every** matching NPC — the exact silent meaning-change the "Count the pipes before you save" bullet twelve lines above warns about. The body then teaches, in this same bullet, that missing the form on two sampled actors is "an ordinary outcome" for that line, so the reader stops investigating a rule that is in fact over-distributing.

```suggestion
  sampled actors without the form is an ordinary outcome for `||||||50` — and the roll is re-made
```

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -10,159 +17,196 @@ description: >-
...
+
+A SPID line is written blind unless you measure the population first. Three checks, all read-only:
+
+1. **Size the group the filter names**, before composing: `housecarl_records` with `types`, a
+   `where` predicate for the faction or keyword, and `counts_only=true` — the cheap census. A
+   faction or keyword test is one quantified step with a link step
+   (`Factions[*any].Faction->editorid = BanditFaction`); resolve the field path with
+   `housecarl:mutagen-reference` rather than guessing it.
+2. **Count how many of those are on a PC level multiplier**, when a Level Filter is in play:
+   `housecarl_records` with the same `types`, `where=["Configuration.Level.LevelMult >= 0"]` and
```
Avick3110: **Check 2 drops check 1's predicate, so it does not count "how many of *those*".**

The heading promises the intersection — how many of the filter's population are on a PC level multiplier. The call as written keeps only `types` and replaces `where` with the level-mult term alone, so it returns every NPC in scope on a multiplier, across the whole order, with no faction or keyword bound.

Scenario: the reader runs check 1 and gets 24,782 bandit-faction NPCs; runs check 2 verbatim and gets, say, 9,300 — the load order's entire PC-level-mult NPC population. The **Stop condition** immediately below tells them to "Quote the counts you measured", so they report 9,300 of 24,782 bandits as scaling actors. `where=` predicates are ANDed, so the correct call keeps both terms.

```suggestion
   `housecarl_records` with the same `types`, the same filter predicate AND
   `"Configuration.Level.LevelMult >= 0"` in `where`, and
```

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -10,159 +17,196 @@ description: >-
...
+are empty; every pipe before Chance is still written.
+
+## Check the reach before you write it
+
+A SPID line is written blind unless you measure the population first. Three checks, all read-only:
+
+1. **Size the group the filter names**, before composing: `housecarl_records` with `types`, a
+   `where` predicate for the faction or keyword, and `counts_only=true` — the cheap census. A
+   faction or keyword test is one quantified step with a link step
+   (`Factions[*any].Faction->editorid = BanditFaction`); resolve the field path with
```
Avick3110, line 123: **Both census predicates read template-masked fields, so the counts are not the population SPID reaches.**

`references/grammar-core.md` §11, in this same skill, states the rule: "**Filters always evaluate the final, resolved NPC** (after merging template attributes)." `Factions[*any].Faction->editorid` (check 1) and `Configuration.Level.LevelMult` (check 2) both read the record's *own* local data, and houseCARL itself documents that this data is masked when the matching `Configuration.TemplateFlags` bit is set — `RecordsTools.cs:2154`: "a SET flag means the category is INHERITED and the seed's own local data for it is MASKED". `Use Factions` masks the record's Factions list; `Use Stats` masks its ACBS level/level-mult arm.

Scenario: on a heavily overhauled order most encounter bandits are `LvlBandit*` records templated onto a base, with `Use Factions` and `Use Stats` set — SPID matches them (it sees the resolved NPC), the census does not (it sees the masked local record). The reader is then told to "Quote the counts you measured" and "never state a reach you did not measure", so a systematically low number is reported as the authoritative reach. `evals.json` assertion 2 scores an answer *correct* for quoting exactly these two numbers, which locks the defect in.

Either the predicates need to account for `Configuration.TemplateFlags` / follow `Template`, or the section needs to say plainly that these counts read local record data and undercount templated actors.

.claude/skills/spid-authoring/references/filters.md
```diff
@@ -162,10 +165,38 @@ Form = 0x12345|||w1(2/3)
...
+
+**The separate pass is a re-distribution, not the distribution.** SPID keeps a second, smaller list
+holding only the level-filtered entries, and replays it for NPCs whose level is a **PC level
+multiplier** — on player level-up, and on game load — so a scaling actor gains an entry when it grows
+into the range and loses it when it falls out. That pass is restricted to PC-level-mult NPCs; it
+adds re-evaluation to an already-distributed population, it does not shrink the population the line
+reached in the first place.
+
+**Why it matters in numbers.** On a heavily overhauled order the two populations are nowhere near
+each other — an audit of one live load order found 17 of 24,782 bandit-faction NPCs on a PC level
```
Avick3110: **"17 of 24,782" is very likely the template-masking artefact, and it is the load-bearing number for the whole correction.**

17/24,782 is 0.07%. Vanilla Skyrim's encounter bandits (`EncBandit*` / `LvlBandit*`) are the canonical PC-level-mult population — SPID gives them their own trait letter, `L` = "Leveled (Is PC Level Mult)" (`value-tables.md`). A bandit-faction population in which 1 in 1,458 scales with the player is not a plausible reading of a live order; it is what `Configuration.Level.LevelMult >= 0` returns when the `Use Stats` template flag has masked the local ACBS arm on nearly every templated bandit (see my comment on `SKILL.md:123`).

Scenario: a reader takes this paragraph at face value and tells a user "only 17 NPCs on your order are on a level multiplier", or reuses the 24,782 figure — itself larger than the total NPC_ count of many full modlists — as a sanity anchor for their own census. Both numbers ship inside the section the body now delegates the entire Level Filter answer to, and the "factor of hundreds" claim is derived from their ratio.

The *correction itself* is well-sourced and I do not dispute it; the quantitative claim needs to be re-measured with a predicate that follows templates, or dropped in favour of the source citation that already carries the argument.

.claude/skills/spid-authoring/evals/evals.json
```diff
@@ -0,0 +1,291 @@
...
+    },
+    {
+      "id": "without-skill",
+      "description": "The same queries with no skill in the catalogue at all. Required on every trigger eval: a with-skill-only measurement reports the skill's self-assessment, not its value. The pair is the only honest measure."
+    }
+  ],
+  "configurations": [
+    {
+      "id": "coexistence",
+      "description": "The nine-skill menu above — the deployment shape, with the three near-neighbour distributor skills present.",
```
Avick3110: **The nine-skill menu is not "the deployment shape" — four of its nine options do not ship, and eleven that do ship are absent.**

This tree ships 14 skills (`.claude/skills/`): biped-slot-reference, bulk-record-jobs, dialogue-authoring, facegen-diagnostics, kid-authoring, mutagen-reference, npc-appearance-copy, oar-authoring, papyrus-optimization, papyrus-reference, skse-plugin-authoring, skypatcher-authoring, spid-authoring, tool-output-awareness. The menu's `D=CID`, `F=crash-diagnostics`, `G=conflict-resolution` and `I=modding-basics` are none of them.

Scenario: the specificity arm is scored by whether a should-not query routes *away* to a named neighbour. Rows 13, 15, 16, 18, 19 and 27 all expect a route to CID, crash-diagnostics, conflict-resolution or modding-basics. In the real catalogue those destinations do not exist, so the only available answers are NONE or this skill — which is precisely the pressure the `isolation` configuration below says is where an over-broad description shows. A specificity number produced against four phantom escape hatches does not measure the shipped install, and row 19 / row 27 (the conceptual-question gap this suite is explicitly designed to detect) are the rows most affected.

Row 147 makes the contradiction internal: `expected_output` is "none of the distributor authoring skills (**CID has no skill**)" while the menu it is measured against offers `D=CID`.

.claude/skills/spid-authoring/evals/evals.json
```diff
@@ -0,0 +1,291 @@
...
+  },
+  "gate": {
+    "recall_threshold": 0.8,
+    "specificity_threshold": 0.5,
+    "runs_per_query": 3,
+    "reporting": "Report the trigger RATE over the three runs per query, not a single verdict. A query that triggers twice of three is a 0.67 row and is reported as one.",
+    "evidence": "Every row carries its own evidence — the adjudicated reasoning behind the pick — on hits as well as misses. A row with a pass and no evidence is not a measurement."
+  },
+  "rerun_triggers": [
+    "Any edit to the frontmatter description.",
```
Avick3110, line 43: **This PR fires its own first rerun trigger and deletes the only recorded evidence without replacing it.**

`rerun_triggers[0]` is "Any edit to the frontmatter description" — and the description is rewritten wholesale in this PR. The deleted `eval_set.json` carried measured results: the 2026-06-11 fan-out (recall 14/14, specificity 12/12, `passed: true`) and the 2026-07-28 revalidation for #294 (recall 0.93, specificity 0.92) with its skill-specific contamination-audit outcome ("q11 and q16 rows re-judged hardened: unchanged"). The new file carries a design and a `gate`, and no result at all.

Scenario: after merge, the shipped `description` — the only text that decides whether this skill is ever loaded — has no measurement behind it, and no measurement behind the version it replaced either. `dev/skill-validation/` holds `2026-06-11-read-triggers/`, `audit-2026-09/` and `eval-2026-09/`, but no `2026-07-28*` record, so the #294 revalidation and its contamination finding exist nowhere else and are lost with this file. The regrade's own "seven D5 rows that all wait on the eval suite being run" is the same gap seen from the other side.

At minimum the superseded results and the contamination-audit note belong in `dev/skill-validation/` before the file is replaced.

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -1,7 +1,14 @@
...
 description: >-
-  Author or interpret SPID (Spell Perk Item Distributor) `_DISTR.ini` files — runtime, no-ESP distribution of spells, perks, items, keywords, outfits, and factions to NPCs. Use when writing or auditing a `_DISTR.ini`, distributing forms to NPCs by faction / race / level / traits, or asking why a line isn't distributing. Load before any SPID line — a misread filter silently changes who's targeted.
+  Authors and interprets SPID (Spell Perk Item Distributor) `_DISTR.ini` files — runtime, no-ESP
+  distribution of spells, perks, items, keywords, outfits and factions to NPCs, on the bundled SPID
+  7.3.0 grammar. Load it before composing or judging any SPID line: a misread filter silently
+  changes who is targeted, and an unparseable one is skipped with no error. Use when writing or
+  auditing a `_DISTR.ini`, targeting NPCs by faction, race, level or trait, or asking why a line
+  isn't distributing. Not this skill: a keyword onto item records is KID, items into containers is
+  CID, a record's own fields is SkyPatcher.
+compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
```
Avick3110, line 11: **`compatibility:` is a frontmatter key no other skill in this tree uses, and nothing here validates it.**

All 13 sibling `SKILL.md` files carry `name` + `description` only — this is the sole occurrence of `compatibility` (or of any third key) across `.claude/skills/`. `scripts/build-plugin.ps1` copies skill trees verbatim and its leak-check only looks for `evals/`, `_CORPUS_STATUS.md`, `appsettings*.json` and dev paths; no step parses skill frontmatter, and `src/housecarl-mcp-tests/` has no frontmatter test, so nothing in the build would catch a key the loader rejects.

Scenario: if the skill loader validates frontmatter against a known key set, the packaged `spid-authoring` fails to parse and the skill does not load at all — a silent total loss of the skill, invisible to `dotnet build`, the test suite, `ci-all` and the packager's own checks, and visible only to a user whose install stops offering it. The PR body records that the package was rebuilt and the pointer scan run, but not that `plugin-validate` was run against the tree (`plugin/CHANGELOG.md:2188` shows that gate exists).

If the key is intentional, it is worth confirming against the loader and applying to the other 13 for consistency; if it is decorative, note that `name` + `description` load into every session's context and this adds a third line to that budget for every user.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded in 03a8f4ea — real, and the same defect the regrade fixed one worked line above. `grammar-core.md` §4 is the authority: five pipes lands 50 in `CountOrPackageIndex`. The bullet now reads `||||||50`.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded in 03a8f4ea. `surface/tools.md` on `where=` says "body predicates, ANDed", so the call as written counted every scaling NPC in scope rather than the intersection the heading promises. Check 2 now carries check 1's predicate alongside `"Configuration.Level.LevelMult >= 0"`, and says why dropping it changes the question. Measured on the audited order to be sure: the faction predicate alone is 24,782, the level-mult predicate alone is 963, and the two together are 17.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded in 03a8f4ea and fda9c052, on the second of the two options you name. The mechanism is real: `where=` reads the record's own local data, a set TemplateFlags bit masks that category, and this skill's own `grammar-core.md` §11 says SPID matches the resolved NPC. The section now says so where it teaches the check — "A census is a floor", naming §11 — and `filters.md` §3 repeats it beside the numbers. I did not make the predicates follow templates: a census that walks each candidate's template chain is a different call shape, and the honest statement of what the cheap census measures is what the skill needs.

Measured the size of the gap on the audited order rather than leaving it open: of the 24,782 bandit-faction NPCs, 1,073 carry `Configuration.TemplateFlags has Stats`, so that is the ceiling on the undercount for check 2 — not the whole population.

.claude/skills/spid-authoring/references/filters.md
Avick3110: Refuted on measurement — I re-ran the census on the same live order (`housecarl_records`, `types=["NPC_"]`, `counts_only=true`, epoch e2-f736766b42e4f032) rather than argue it:

- `Factions[*any].Faction->editorid = BanditFaction` → **24,782**
- `Configuration.Level.LevelMult >= 0` alone, whole order → **963**
- both predicates together → **17**

Both published figures reproduce exactly. The premise the finding rests on — that 17 is implausibly small for a bandit population — assumes a large PC-level-mult population to be masked out of; the order carries 963 PC-level-mult NPC records in total, across every faction, so 17 bandits among them is the ordinary reading. This is a Requiem order, which de-levels the world; that is what a de-levelled order looks like at the data layer.

The masking effect is real but bounded, and I folded it (see the `SKILL.md:123` thread): 1,073 of the 24,782 carry `Configuration.TemplateFlags has Stats`, which is the ceiling on the undercount, not a factor of hundreds. `filters.md` §3 now attributes both figures to a census, adds the whole-order 963 so the ratio can be judged, states the template floor, and says the source citation rather than the ratio is what settles the rule.

.claude/skills/spid-authoring/evals/evals.json
Avick3110: Folded in f215b778. The menu was carried over verbatim from the 2026-06-11 run and the new text called it "the deployment shape", which it is not — four options do not ship and eleven that do are absent. The `method` menu is now the 14 skills this tree ships, and the rows whose destination never existed are scored against NONE (13, 15, 16, 18, 19, 27), with row 16 also allowing `housecarl:bulk-record-jobs`. The method note keeps a line saying which four the old runs offered, so the superseded scores can still be read for what they were. Row 147's internal contradiction goes with it.

.claude/skills/spid-authoring/evals/evals.json
Avick3110: Folded in f215b778 — with one correction to the premise: nothing was lost, the prior file is `evals/eval_set.json` at `358c5717^` and git keeps it. Findability was the real complaint and it stands. `evals.json` now carries a `superseded_results` block with both runs (2026-06-11 recall 14/14 / specificity 12/12; the #294 revalidation 0.93 / 0.92), what description each measured, and the contamination audit including the skill-specific "q11 and q16 re-judged hardened: unchanged" outcome, with a pointer to the commit for the full file. It sits with the file that replaced them rather than in `dev/`, which is gitignored and would put the record outside the repo; the packager strips `evals/`, so it costs nothing shipped.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Refuted. `compatibility` is what the rewrite document specifies, not a decoration this PR invented: REWRITE.md §3 — "Three keys. `name` unchanged; `description` rewritten; `compatibility` added per ruling 7" — with the key measured at 76 characters against the 500 cap (D2.35) and recorded as inside the spec's closed six (D2.30), which is why the packaging and upload paths accept it. A finding that asks for something a ruling settled otherwise is refuted by citing it.

The runnable half of the concern I did check rather than assert: `claude plugin validate dist/housecarl --strict` passes on the rebuilt package with the key present, so the loader's own gate does not reject it. The wave-0 packager PR is where a frontmatter check would live if one is wanted; this branch does not touch `scripts/build-plugin.ps1`.

Applying it to the other 13 is a per-skill call for their own rewrite waves, not a change this PR should make on their behalf.

Avick3110:
Fold pass on the seven review comments: **five folded, two refuted**, each answered in its own thread.

Folded — `03a8f4ea` the Chance pipe count (`|||||50` is a Count) and check 2, which dropped check 1's predicate and so counted every scaling NPC in the order rather than the intersection, plus a "a census is a floor" paragraph saying both counts read each record's own local data that a template masks; `fda9c052` the same caveat beside the numbers in `filters.md` §3, with the figures attributed to a census and the whole-order PC-level-mult count added so the ratio can be judged; `f215b778` the eval menu, which claimed to be the deployment shape while offering four skills that do not ship, now the shipped 14 with the orphaned rows scored against NONE, and a `superseded_results` block keeping the 2026-06-11 and #294 scores and the contamination audit.

Refuted — the claim that 17 of 24,782 is a template-masking artefact, which I settled by re-running the census on the same order (24,782 / 963 / 17, both published figures reproducing, and 1,073 records carrying `TemplateFlags has Stats` as the ceiling on any undercount); and the `compatibility` frontmatter key, which REWRITE.md §3 specifies per ruling 7 and which `claude plugin validate --strict` accepts on the rebuilt package.

Package rebuilt (leak-check clean, 12 skills / 250 markdown files), `plugin validate --strict` passes, build clean, 1793 tests pass, `ci-all` 128/128.

.claude/skills/spid-authoring/evals/evals.json
```diff
@@ -0,0 +1,311 @@
...
+  "method": "Trigger arm: an independent fan-out of fresh-context agents, one per query per run, each shown a neutral menu of skill descriptions (this skill among realistic siblings, not flagged as under test) plus one query, asked which ONE skill to load (or NONE) and why. Verdicts are adjudicated from each agent's reasoning and its pick. Menu: the 14 skills this tree actually ships, each shown with its own description and none flagged as under test — biped-slot-reference, bulk-record-jobs, dialogue-authoring, facegen-diagnostics, kid-authoring, mutagen-reference, npc-appearance-copy, oar-authoring, papyrus-optimization, papyrus-reference, skse-plugin-authoring, skypatcher-authoring, spid-authoring [under test], tool-output-awareness — plus NONE. The 2026-06-11 and 2026-07-28 runs used a nine-letter menu containing CID, crash-diagnostics, conflict-resolution and modding-basics, none of which ship; a should-not query whose only honest destination is one of those must be scored against NONE, not against a skill the install does not offer.",
```
Avick3110: **The menu names three skills this tree does not ship.** It is specified as "the 14 skills this tree actually ships", listing `biped-slot-reference`, `papyrus-optimization` and `tool-output-awareness` among them. All three were deleted: `dev/DECISIONS.md` 2026-09-07 ruling 6b ("Deleted skills go whole: biped-slot-reference, papyrus-optimization and tool-output-awareness lose their eval sets and reference files with the skill"). `git ls-tree -d <branch> .claude/skills/` returns 11 directories, and `$Skills` in `scripts/build-plugin.ps1` lists the same 11.

This is the defect the note indicts two sentences earlier — the older runs "used a nine-letter menu containing CID, crash-diagnostics, conflict-resolution and modding-basics, none of which ship". The suite has not been run yet (the regrade records seven D5 partials waiting on it), so when it is, the menu as written cannot be built, and a specificity number measured against three phantom neighbours is not the deployment shape the arm claims to measure.

Line 183 carries the same count: "none of the **14** covers the question".

.claude/skills/spid-authoring/references/filters.md
```diff
@@ -162,10 +165,42 @@ Form = 0x12345|||w1(2/3)
...
+**The separate pass is a re-distribution, not the distribution.** SPID keeps a second, smaller list
+holding only the level-filtered entries, and replays it for NPCs whose level is a **PC level
+multiplier** — on player level-up, and on game load — so a scaling actor gains an entry when it grows
+into the range and loses it when it falls out. That pass is restricted to PC-level-mult NPCs; it
+adds re-evaluation to an already-distributed population, it does not shrink the population the line
+reached in the first place.
+
+**Why it matters in numbers.** On a heavily overhauled order the two populations are nowhere near
+each other — a census of one live load order (`housecarl_records`, `counts_only=true`) found 17 of
+24,782 bandit-faction NPCs on a PC level multiplier, against 4,024 matching the line's own level
```
Avick3110: **4,024 is not the level-bound population — it carries the `-U` term too.**

I re-measured all five numbers against the same order, same epoch (`e2-f736766b42e4f032`):

| claim | predicate | result |
|---|---|---|
| 24,782 bandit-faction NPCs | `Factions[*any].Faction->editorid = BanditFaction` | 24,782 OK |
| 17 on a PC level multiplier | + `Configuration.Level.LevelMult >= 0` | 17 OK |
| 963 PC-level-mult NPCs in total | 66,855 scanned - 65,892 unreadable | 963 OK |
| 1,073 inherit stats | + `Configuration.TemplateFlags has Stats` | 1,073 OK |
| **4,024 matching the line's own level bound** | + `Configuration.Level.Level >= 20` | **4,040** |

4,024 is that set plus a third predicate — `Configuration.Flags has_none Unique`. Because the other four reproduce exactly, this is not load-order drift: the `-U` term was carried into a label that names only the level bound.

It matters here more than elsewhere: this is the passage the body's check 2 points at when it teaches that `where` predicates are ANDed and that dropping a term changes what you counted, under a stop condition of "Quote the counts you measured and no others". Either relabel it ("4,024 matching the level bound and not Unique") or requote the two-term figure, 4,040 — the "factor of hundreds" argument holds on both.

.claude/skills/spid-authoring/references/filters.md
```diff
@@ -162,10 +165,42 @@ Form = 0x12345|||w1(2/3)
...
-filter discards the NPC or the chance roll fails). This is a distinct distribution path from regular
-(non-level-filtered) entries.
+### What a Level Filter reaches — and what it does not **[source]**
+
+This section owns the question; the skill body points here rather than restating it.
+
+**A Level Filter narrows by the NPC's own level. It does not narrow the line to auto-levelled NPCs.**
+On the ordinary on-load path SPID runs the **whole** entry list — level-filtered entries included —
+against every non-player, non-deleted NPC that loads, and compares the range to that NPC's own
+level. So a line reading `|20` reaches every matching NPC at level 20 or above, fixed-level actors
```
Avick3110: **`|20` is one pipe — that is the String section, not the Level section.** A Level Filter is position 3, so the line is `|||20`; thirty lines above, this same file writes "at least level 5" as `Form = 0x12345|||5`, and the header block states the positions.

An agent copying the shorthand out of the section that *owns* this rule emits a String filter whose term is `20`, which matches no NPC name, EditorID, template or keyword and is skipped with no log line — the silent failure the skill's first Common Mistake exists to prevent.

```suggestion
level. So a line reading `|||20` reaches every matching NPC at level 20 or above, fixed-level actors
```

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -10,159 +17,203 @@ description: >-
...
+   - Combine filters per `references/grammar-core.md`: **OR within a section, AND between
+     sections**, **exclusions (`-X`) always AND**. For a *union* of groups, write multiple lines
+     for one form.
+
+4. **Set Count / Index / Chance** if needed (`references/grammar-core.md`): item count or `min-max`
+   range; zero-based package index; package-list type `0`–`4`; a `0`–`100` chance (default 100).
+   Append `!` to the chance to make it deterministic (consistent per NPC + save across sessions).
+
+5. **Place the file** at `Data/<name>_DISTR.ini` — the **`_DISTR` suffix is mandatory** and the file
+   is flat in `Data/`, never in a per-type subfolder; without both it is never read. Files load
```
Avick3110: **"without both" reads as "only when both are missing".** Two independent requirements are joined here — the `_DISTR` suffix and flat placement in `Data/` — and each is on its own fatal. As written, `Data/SPID/MyMod_DISTR.ini` still satisfies the sentence (it has the suffix), but SPID never reads it.

The Common Mistakes bullet below states the rule correctly and the two now disagree: "A file without the suffix, **or** filed into a per-type subfolder, is ignored entirely".

```suggestion
   is flat in `Data/`, never in a per-type subfolder; without either it is never read. Files load
```

.claude/skills/spid-authoring/references/filters.md
Avick3110: Folded — `925c045b`. `references/filters.md:175` now reads `|||20`.

You are right and the file contradicted itself: the header block at `references/filters.md:5-10` puts Level Filters at position 3, and the section's own example thirty lines above writes "at least level 5" as `Form = 0x12345|||5` (`references/filters.md:159`). `|20` is a String filter whose term is `20`, which matches nothing and is skipped silently — the failure the first Common Mistake exists to prevent, emitted by the section that owns the rule.

.claude/skills/spid-authoring/references/filters.md
Avick3110: Folded — `6009aeda`. `references/filters.md:187` now quotes 4,040.

I re-measured both readings myself on the same order and the same epoch you name, `e2-f736766b42e4f032`, and they reproduce your table exactly:

- `types=["NPC_"]`, `where=["Factions[*any].Faction->editorid = BanditFaction", "Configuration.Level.Level >= 20"]`, `counts_only=true` → **4,040**
- the same call plus `"Configuration.Flags has_none Unique"` → **4,024**

So the label carried a third term it does not name. I took the second of your two options and requoted the two-term figure rather than lengthening the label, because this is the paragraph the body's check 2 points at when it teaches that `where` predicates are ANDed — a two-term label over a three-term count is the exact mistake the check warns about, and the shorter label is the one a reader can reproduce from the sentence. The other four figures are unchanged, and the "factor of hundreds" claim reads the same on 4,040.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded — `c9b0a782`. `SKILL.md:89` now reads "without either it is never read", your wording.

The two requirements are independent and each is on its own fatal, so "without both" both read as "only when both are missing" and let `Data/SPID/MyMod_DISTR.ini` satisfy the sentence. The Common Mistakes bullet at `SKILL.md:174` already said it correctly ("A file without the suffix, or filed into a per-type subfolder, is ignored entirely"); the two now agree.

.claude/skills/spid-authoring/evals/evals.json
Avick3110: Folded — `4635bfcc`. The menu is now the eleven skills that ship.

You are right, and this is my own fold from the blind-review pass overshooting: `f215b778` replaced the nine-letter phantom menu with a "shipped 14" that was itself wrong. `.claude/skills/` on this branch holds 11 directories and the `$Skills` list at `scripts/build-plugin.ps1:60` holds the same 11; biped-slot-reference, papyrus-optimization and tool-output-awareness are gone under `dev/DECISIONS.md` 2026-09-07 ruling 6b, and the just-rebuilt package reports `skills: 11`.

Three places changed:

- `evals/evals.json:5` — the menu is bulk-record-jobs, dialogue-authoring, facegen-diagnostics, kid-authoring, mutagen-reference, npc-appearance-copy, oar-authoring, papyrus-reference, skse-plugin-authoring, skypatcher-authoring, spid-authoring [under test], plus NONE, cited to `$Skills`, with one sentence saying the three deleted skills are not on it.
- `evals/evals.json:19` — the coexistence configuration says "shipped 11-skill menu".
- `evals/evals.json:183` — row 15's expected output says "none of the 11".

I re-checked every `expected_output` while I was in the file: no row routes to a deleted skill, so the rows themselves needed nothing.

Avick3110:
Fold pass on Aaron's four review comments: **all four folded**, nothing refuted, each answered in its own thread.

`925c045b` the Level Filter shorthand, which the section that owns the rule wrote as `|20` — one pipe, a String filter — against its own header block and its own `|||5` example. `6009aeda` the 4,024 label: I re-ran both readings on the same order and epoch (`e2-f736766b42e4f032`) and reproduce the table exactly — 4,040 on the two terms the label names, 4,024 once `Configuration.Flags has_none Unique` joins them — so the paragraph now quotes 4,040. `c9b0a782` the placement sentence, whose "without both" let `Data/SPID/MyMod_DISTR.ini` satisfy two requirements that are each on their own fatal. `4635bfcc` the trigger menu, which my own earlier fold overshot into a "shipped 14": `.claude/skills/` and `$Skills` at `scripts/build-plugin.ps1:60` both hold 11, so the menu, the coexistence configuration and row 15 now say 11 and name the three deleted skills as off the menu; no row routed to one of them.

Package rebuilt (skill leak-check and pointer scan clean, `skills: 11`), build clean, 1858 tests pass, `ci-all` 128/128.

Still open for Aaron, unchanged from the PR body: open question 2 (whether the body may teach the `housecarl_skse` version read — I decided yes; overrule and I cut it to prose) and open question 3 (whether to file a SPID sibling of #613 for an offline draft check, a ruling-6 call).

PR #635, opened 2026-09-08
Make max_chars a ceiling on the last five records renders

### LINKED ISSUE #604: [GAP] max_chars= still overshoots on the records delta, tree, chain, effect-chain and info_order text renders
PR #603 (#546) made `max_chars=` a hard ceiling on the BatchRender family, the skse renders, and the records scan, batch, resolve, group and summary text renders. Five `housecarl_records` text renders were left out because their inner loops and the shared `AppendInfoOrderView` / `RenderEffectChain` helpers would need the same charge-and-roll-back shape as a change of its own: `project.form='delta'`, `'tree'`, the walk lane's `'chain'` and reverse effect-chain renders, and `'info_order'`. Each still tests the cap before an item and appends the item and its notice past it.

The `max_chars=` parameter description and the CHANGELOG line from #603 name exactly these five as not held to the ceiling, so the surface is honest until this lands. The fix is the shape #603 used: compose and charge the trailer and notice first, write an item, measure, roll it back whole if it crossed, and route a too-small cap through `RenderCap.Settle`. Tests per form: fill past the cap and assert length at most `max_chars`.

Also outside #603 by scope, unchanged: `apply`, `create`, `forward`, `remove`, `write_seq`, `bsa_list`, `load_order_status`, `update_status`, `skypatcher_layer` take `max_chars=` and can still answer past it.

Avick3110:
`max_chars=` was a test these five renders took *before* writing the next thing, not a bound on what came back: the record block, node, delta line or topic that crossed the line went in whole, and the truncation notice and the `spilled:` block were appended on top of it. #603 gave every other render in the tool the charge-and-roll-back shape and named these five in the parameter description and the changelog as the ones it did not hold. They now carry the same shape, and both of those statements are gone.

The layout is the bound. Per render: the truncation notice at its widest spelling and the spill block are composed and charged before the first record is laid; every inner list (delta lines, tree nodes, child declarers, chain nodes, effect-chain carrier rows) composes the line it is about to write and prices it **together with its own cut notice**, so the notice lands inside the budget rather than a character past the line that crossed; a record that runs to completion and still crosses is taken back out whole and counted in the notice; and the response settles through `RenderCap.Settle`, so the one arm left — a `max_chars` too small for what the response carries whatever the budget, its spill block included — says so and names the number that clears it.

Two states a row can end in, and each is stated:

- **said** — the row stopped inside the budget and named what it held back (`[delta lines cut …]`, `[nodes cut …]`, `[child declarers cut …]`). The row stays; nothing more fits, so the render stops after it.
- **mute** — the row stopped with no room even for that notice. The whole row goes back out and the render's own `[rendered N of M …]` notice takes its place. A section never ends a row in silence.

To keep `mute` off the common paths, two sections now reserve their notice alongside the thing that starts them: the declarers block will not write its framing line unless the cut notice still fits under it, and the tree's `diff (field deltas vs …)` heading is only written where a node or that notice fits beneath it. A sweep of every cap from 150 to 1800 on both fixture cells is monotonic and inside the ceiling at every one.

Closes #604

## What changed, per render

| render | what it did | what it does |
|---|---|---|
| `delta` | cap tested before the record and before each delta line; the crossing one written whole; `[delta lines cut …]` and the `spilled:` block appended past the budget | notice and spill charged first; each delta line priced with its own cut notice; a record that completes and crosses is rolled back whole; `Settle` on the response |
| `tree` | same, plus `AppendChildDeclarers` checking only `sb.Length >= cap` before each declarer line, and the `diff` heading and `[nodes cut …]` written unconditionally | the block takes a `RenderCap` and the caller's `tailReserve` (the nodes notice a multi-provider row still owes), each declarer and node line is composed then priced with its notice, and the framing line and diff heading each reserve the notice that follows them |
| `chain` | cap tested before each node; cycles, the truncation note and the whole TemplateFlags report appended after it without a check | the node line is composed and priced with `[nodes cut …]`; a seed cut mid-list ends there rather than appending a report the budget did not hold; a seed that completes and crosses is rolled back whole |
| reverse effect chain | `Wire.RenderEffectChain` was handed the caller's whole `cap` per seed — so *each* seed could fill the entire budget — and built its own buffer the caller then appended past the ceiling | the shared render takes a `RenderCap` plus what the caller has already spent, prices each carrier row with its own cut notice, and still quotes the caller's `max_chars`; the seed block is rolled back whole if it crosses |
| `info_order` | `DialogueWire.AppendInfoOrderView` handed the whole `cap`, its `[truncated at max_chars]` marker and its trailing MOVED paragraph and order note appended past it | the shared view is bounded by what this render has left, and a topic that still crossed is taken back out whole and counted |

**Not changed:** `DialogueWire.AppendInfoOrderView` itself. Its tail — the MOVED paragraph and the order note — is appended after its row loop and its width is content-dependent, so there is no way to keep half a topic and stay under the ceiling without reshaping a helper the dialogue surface shares. A topic wider than the budget is therefore counted (`[rendered 0 of 1 rows at max_chars=…]`) rather than shown half, which is the same contract the scan and batch lanes have carried since #603. `RecordsTools.AppendChildDeclarers`'s tail return is now unreachable by construction — each declarer line is priced with the notice beside it, so a completed block always leaves room for it — and it returns `false` with that stated.

## Tests

`RecordsArtifactTests` gains the ceiling contract for `delta`, `tree`, `chain` and the reverse effect chain (four caps each, both arms of the bound), plus one case pinning that the notice quotes the caller's own `max_chars` and the artifact still holds the complete result. `DialogueFamilyTests` gains the same for `info_order`. `RecordsOwnedChildTests` gains a cap sweep over the whole band where a tree row is cut.

| test | before | after |
|---|---|---|
| `ADeltaRenderIsNeverWiderThanItsCap` (4 caps) | fails — 1,601 chars at `max_chars=1200`, and over at the others | passes |
| `ATreeRenderIsNeverWiderThanItsCap` (4 caps) | fails — 1,490 at `max_chars=1200` | passes |
| `AChainRenderIsNeverWiderThanItsCap` (4 caps) | fails | passes |
| `AReverseEffectChainRenderIsNeverWiderThanItsCap` (4 caps) | fails — 1,799 chars at `max_chars=400` | passes |
| `AnInfoOrderRenderIsNeverWiderThanItsCap` (5 caps) | fails at 400/700/1200 | passes |
| `EachOfTheseRendersSaysWhatItsCapHeldBackAndStaysInsideIt` | fails — the response ran past the cap | passes |
| `AnInfoOrderCutByItsCapSaysHowManyTopicsItHeldBack` | fails | passes |
| `TheTreeRenderIsNeverWiderThanItsCap` (2 rows × 236 caps) | fails across the band | passes at every cap |

**Caps that moved, and why the tests that pin them now drive the render.** Ten `RecordsOwnedChildTests` cases pin an exact `max_chars` and assert what the tree does at it. Under the ceiling, a cut response also carries the auto-spill block — and that block names an **absolute path**, three times. Pinning a cap through the tool would bake this checkout's path length into every one of those numbers, and CI's is different. So those ten now drive `RenderRecordsTree` directly through `LoadOrderService.TreeBatch`, with no artifact lane around it: the boundaries become properties of the render, identical everywhere. Their caps moved with it (590/591 → 631/632, 600 → 500, 625 → 755, 660 → 800, 700 → 800).

**Two whose premise the change removed.** `ASoleProviderRowWhoseCompleteBlockEndsPastTheCapClaimsNothingWasCut` and `ARowWhoseDeclarersBlockFitsExactlyAtTheTailIsNotMarkedTruncated` both pinned "the block ran to completion and the row ended *past* the cap" — the overshoot this PR removes. There is no such state now, so they are replaced by the pair that states the new boundary: at the first cap the whole row fits inside, nothing is cut and nothing is marked truncated; one character below it, the row is cut and says so. `AMultiProviderRowWhoseCompleteBlockEndsPastTheCapNamesTheDiffItLost` keeps its claim — a complete block that leaves no room for the diff still names the diff it lost — at a cap where that is what happens.

The two tests that drive `AppendChildDeclarers` directly pass `new RenderCap(100_000, 100_000)` and `tailReserve: 0` for the widened signature.

## What was run

From inside `.claude/worktrees/max-chars-five`:

```
```

1,813 passed, 0 failed. `ci-all`: 127/127. Nothing was run against the live MO2 instance.

## What this still does NOT cover

Unchanged and outside this issue, as #603 recorded: `housecarl_apply`, `housecarl_create`, `housecarl_forward`, `housecarl_remove`, `housecarl_write_seq`, `housecarl_bsa_list`, `housecarl_load_order_status`, `housecarl_update_status` and `housecarl_skypatcher_layer` also take `max_chars=` and can still answer past it. The json lanes of all five renders here were already bounded and are untouched.

### DIFF
--- a/src/housecarl-mcp/RecordsTools.cs
+++ b/src/housecarl-mcp/RecordsTools.cs
@@ -138,7 +138,7 @@ public static class RecordsTools
             int limit = 500,
         [Description("TRANSPORT: skip the first N matches (exact windows: offset=0/500/1000…). Windows tile only WITHIN one epoch — if two pages' epochs differ the load order changed mid-pagination; re-run from offset=0, do not stitch the pages. offset= RE-SCANS the selection from the start rather than seeking into it, so every window pays the whole scan again and a deep window costs more than a shallow one — narrowing the scan terms beats paging far into one.")]
             int offset = 0,
-        [Description("TRANSPORT: character CEILING on the RENDER, hard on every text render this tool has — the scan, batch, resolve, group_by and summary renders, the comparison forms (delta, tree), the walk lane's chain and effect-chain renders, and info_order. The record block, node or delta line that would cross it is not written, and the truncation notice, the accounting line and the spilled: block are charged before the rows are laid. The one answer that can still come back over it is a max_chars too small for what the response carries whatever the budget — its header, the notices it owes, its spilled: block — which says so and names the number that clears it. Never truncates the RESULT: an over-ceiling result SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, the row schema, and the epoch) and the response names the file, so what the ceiling held back inline is in the file. 0 = the server default (~80k).")]
+        [Description("TRANSPORT: character CEILING on the RENDER, hard on every text render this tool has — the scan, batch, resolve, group_by and summary renders, the comparison forms (delta, tree), the walk lane's chain and effect-chain renders, and info_order. The record block, node or delta line that would cross it is not written, and the truncation notice, the accounting line and the spilled: block are charged before the rows are laid — charged only where the whole render does not fit, so an answer that fits inside the max_chars you passed comes back complete, uncut and unspilled. The one answer that can still come back over it is a max_chars too small for what the response carries whatever the budget — its header, the notices it owes, its spilled: block — which says so and names the number that clears it. Never truncates the RESULT: an over-ceiling result SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, the row schema, and the epoch) and the response names the file, so what the ceiling held back inline is in the file. 0 = the server default (~80k).")]
             int max_chars = 0,
         [Description("TRANSPORT: return the accounting block and counts only, no rows — the cheap census.")]
             bool counts_only = false,
@@ -1891,10 +1891,17 @@ public static class RecordsTools
     /// enough delta lines to be cut) has no other way in.</summary>
     internal static string RenderRecordsDelta(IReadOnlyList<LoadOrderService.DeltaRow> rows, int total, int differing, int identical,
                                      int noVerdict, int errors,
-                                     string headerLine, OrderStamp? epoch, int maxChars, SpillState? spill, out bool truncated)
+                                     string headerLine, OrderStamp? epoch, int maxChars, SpillState? spill, out bool truncated,
+                                     bool unreserved = false)
     {
         truncated = false;
         int cap = maxChars > 0 ? maxChars : Wire.DefaultMaxChars;
+        if (!unreserved)
+        {
+            var whole = RenderRecordsDelta(rows, total, differing, identical, noVerdict, errors, headerLine, epoch,
+                                           maxChars, spill, out _, unreserved: true);
+            if (whole.Length <= cap) return whole;
+        }
         bool manifestOnly = spill?.ManifestOnly ?? false;
         var sb = new StringBuilder();
         sb.Append(headerLine).Append('\n');
@@ -1912,7 +1919,7 @@ public static class RecordsTools
         string Notice(int r) =>
             "... [rendered " + r + " of " + rows.Count + " rows at max_chars=" + cap + "]\n";
         var spillText = Wire.SpillText(spill);
-        int budget = Math.Max(cap - spillText.Length - Notice(rows.Count).Length, 0);
+        int budget = unreserved ? Unbounded : Math.Max(cap - spillText.Length - Notice(rows.Count).Length, 0);
         string deltaCut = CutNotice("delta lines", cap);
         foreach (var row in rows)
         {
@@ -1995,6 +2002,14 @@ public static class RecordsTools
         return RenderCap.Settle(sb.ToString().TrimEnd('\n'), cap);
     }
 
+    /// <summary>The budget of the unreserved pass every bounded render here makes first: no unit can cross it, so
+    /// that pass lays the COMPLETE render, and a complete render that fits max_chars is the answer. The reserves
+    /// the bounded pass holds back — the accounting line, each list's cut notice — are room for notices a complete
+    /// render never writes, so charging them against one that fits cuts an answer that fitted, spills it, and (the
+    /// spill block being charged in turn) comes back with no rows at all, naming a number far over what the whole
+    /// answer took. They are charged only once the whole thing is known not to fit at this cap.</summary>
+    const int Unbounded = int.MaxValue / 2;
+
     /// <summary>Whole units only: a unit written from <paramref name="mark"/> that crossed <paramref name="budget"/>
     /// — or that stopped early with no room to say so, <paramref name="force"/> — is taken back out entire and the
     /// caller's notice put in its place. True means the render stops here.</summary>
@@ -2036,10 +2051,16 @@ public static class RecordsTools
     /// on a record only one in-order plugin touches) has no other way in.</summary>
     internal static string RenderRecordsTree(IReadOnlyList<LoadOrderService.TreeRow> rows, int total, int contested, int errors,
                                     bool fieldsNarrow, string headerLine, OrderStamp? epoch, int maxChars,
-                                    SpillState? spill, out bool truncated)
+                                    SpillState? spill, out bool truncated, bool unreserved = false)
     {
         truncated = false;
         int cap = maxChars > 0 ? maxChars : Wire.DefaultMaxChars;
+        if (!unreserved)
+        {
+            var whole = RenderRecordsTree(rows, total, contested, errors, fieldsNarrow, headerLine, epoch, maxChars,
+                                          spill, out _, unreserved: true);
+            if (whole.Length <= cap) return whole;
+        }
         bool manifestOnly = spill?.ManifestOnly ?? false;
         var sb = new StringBuilder();
         sb.Append(headerLine).Append('\n');
@@ -2051,7 +2072,8 @@ public static class RecordsTools
         string Notice(int r) =>
             "... [rendered " + r + " of " + rows.Count + " rows at max_chars=" + cap + "]\n";
         var spillText = Wire.SpillText(spill);
-        var room = RenderCap.For(cap, spillText.Length + Notice(rows.Count).Length);
+        var room = unreserved ? new RenderCap(cap, Unbounded)
+                              : RenderCap.For(cap, spillText.Length + Notice(rows.Count).Length);
         int budget = room.Budget;
         string nodesCut = CutNotice("nodes", cap);
         foreach (var row in rows)
@@ -2228,10 +2250,16 @@ public static class RecordsTools
     /// with nodes enough for max_chars to cut) has no other way in.</summary>
     internal static string RenderRecordsChain(IReadOnlyList<LoadOrderService.WalkSeedResult> rows, int total, int reached,
                                      int errors, string headerLine, OrderStamp? epoch, int maxChars,
-                                     SpillState? spill, out bool truncated)
+                                     SpillState? spill, out bool truncated, bool unreserved = false)
     {
         truncated = false;
         int cap = maxChars > 0 ? maxChars : Wire.DefaultMaxChars;
+        if (!unreserved)
+        {
+            var whole = RenderRecordsChain(rows, total, reached, errors, headerLine, epoch, maxChars, spill, out _,
+                                           unreserved: true);
+            if (whole.Length <= cap) return whole;
+        }
         bool manifestOnly = spill?.ManifestOnly ?? false;
         var sb = new StringBuilder();
         sb.Append(headerLine).Append('\n');
@@ -2243,7 +2271,7 @@ public static class RecordsTools
             "... [rendered " + r + " of " + rows.Count + " seeds at max_chars=" + cap + "]\n";
         string nodesCut = "    ... [nodes cut at max_chars=" + cap + " — raise max_chars, or to_file= for the complete walk]\n";
         var spillText = Wire.SpillText(spill);
-        int budget = Math.Max(cap - spillText.Length - Notice(rows.Count).Length, 0);
+        int budget = unreserved ? Unbounded : Math.Max(cap - spillText.Length - Notice(rows.Count).Length, 0);
         foreach (var row in rows)
         {
             if (manifestOnly) break;
@@ -2330,13 +2358,23 @@ public static class RecordsTools
     /// seed's carriers through the shared effect-chain render, the standard explicit cut and spill marker.
     /// max_chars is a CEILING: the notice and the spill block are charged first, the shared render is told what
     /// this one has already spent, and a seed that would cross what is left is taken back out whole. A seed the
-    /// shared render cut inside its own buffer reports that back — nothing here can measure it.</summary>
-    static string RenderRecordsEffectChains(IReadOnlyList<(string Seed, EffectChainResult Result)> results,
+    /// shared render cut inside its own buffer reports that back — nothing here can measure it.
+    /// Internal so a test can drive it against a hand-built <see cref="EffectChainResult"/>, the same reason
+    /// <see cref="RenderRecordsTree"/> is: reaching a cut INSIDE the shared render through the tool needs a seed
+    /// whose carriers are wider than the auto-spill block, and no fixture has one.</summary>
+    internal static string RenderRecordsEffectChains(IReadOnlyList<(string Seed, EffectChainResult Result)> results,
                                             int totalSeeds, int carrierRows, int carrierTotal, int errors, string headerLine,
-                                            OrderStamp? epoch, int maxChars, SpillState? spill, out bool truncated)
+                                            OrderStamp? epoch, int maxChars, SpillState? spill, out bool truncated,
+                                            bool unreserved = false)
     {
         truncated = false;
         int cap = maxChars > 0 ? maxChars : Wire.DefaultMaxChars;
+        if (!unreserved)
+        {
+            var whole = RenderRecordsEffectChains(results, totalSeeds, carrierRows, carrierTotal, errors, headerLine,
+                                                  epoch, maxChars, spill, out _, unreserved: true);
+            if (whole.Length <= cap) return whole;
+        }
         bool manifestOnly = spill?.ManifestOnly ?? false;
         var sb = new StringBuilder();
         sb.Append(headerLine).Append('\n');
@@ -2349,7 +2387,8 @@ public static class RecordsTools
         string Notice(int r) =>
             "... [rendered " + r + " of " + results.Count + " seeds at max_chars=" + cap + "]\n";
         var spillText = Wire.SpillText(spill);
-        var room = RenderCap.For(cap, spillText.Length + Notice(results.Count).Length);
+        var room = unreserved ? new RenderCap(cap, Unbounded)
+                              : RenderCap.For(cap, spillText.Length + Notice(results.Count).Length);
         foreach (var (seed, result) in results)
         {
             if (manifestOnly) break;
@@ -2378,10 +2417,16 @@ public static class RecordsTools
     /// topic block, so there is no way to keep half of one and stay under the ceiling.</summary>
     static string RenderRecordsInfoOrder(IReadOnlyList<LoadOrderService.InfoOrderRow> rows, int total, int contested,
                                          int errors, string headerLine, OrderStamp? epoch, int maxChars,
-                                         SpillState? spill, out bool truncated)
+                                         SpillState? spill, out bool truncated, bool unreserved = false)
     {
         truncated = false;
         int cap = maxChars > 0 ? maxChars : Wire.DefaultMaxChars;
+        if (!unreserved)
+        {
+            var whole = RenderRecordsInfoOrder(rows, total, contested, errors, headerLine, epoch, maxChars, spill,
+                                               out _, unreserved: true);
+            if (whole.Length <= cap) return whole;
+        }
         bool manifestOnly = spill?.ManifestOnly ?? false;
         var sb = new StringBuilder();
         sb.Append(headerLine).Append('\n');
@@ -2392,7 +2437,7 @@ public static class RecordsTools
         string Notice(int r) =>
             "... [rendered " + r + " of " + rows.Count + " rows at max_chars=" + cap + "]\n";
         var spillText = Wire.SpillText(spill);
-        int budget = Math.Max(cap - spillText.Length - Notice(rows.Count).Length, 0);
+        int budget = unreserved ? Unbounded : Math.Max(cap - spillText.Length - Notice(rows.Count).Length, 0);
         foreach (var row in rows)
         {
             if (manifestOnly) break;
--- a/src/housecarl-mcp/RecordsTools.cs
+++ b/src/housecarl-mcp/RecordsTools.cs
@@ -138,7 +138,7 @@ public static class RecordsTools
             int limit = 500,
         [Description("TRANSPORT: skip the first N matches (exact windows: offset=0/500/1000…). Windows tile only WITHIN one epoch — if two pages' epochs differ the load order changed mid-pagination; re-run from offset=0, do not stitch the pages. offset= RE-SCANS the selection from the start rather than seeking into it, so every window pays the whole scan again and a deep window costs more than a shallow one — narrowing the scan terms beats paging far into one.")]
             int offset = 0,
-        [Description("TRANSPORT: character CEILING on the RENDER, and a hard one on the scan, batch, resolve, group_by and summary renders — the record block that would cross it is not written, and the truncation notice, the accounting line and the spilled: block are charged before the rows are laid. NOT yet hard on the comparison forms (delta, tree), the walk lane's chain and effect-chain renders, or info_order: those still test the ceiling before the row, so they can come back over it by the block that crossed. Never truncates the RESULT: an over-ceiling result SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, the row schema, and the epoch) and the response names the file, so what the ceiling held back inline is in the file. 0 = the server default (~80k).")]
+        [Description("TRANSPORT: character CEILING on the RENDER, hard on every text render this tool has — the scan, batch, resolve, group_by and summary renders, the comparison forms (delta, tree), the walk lane's chain and effect-chain renders, and info_order. The record block, node or delta line that would cross it is not written, and the truncation notice, the accounting line and the spilled: block are charged before the rows are laid. The one answer that can still come back over it is a max_chars too small for what the response carries whatever the budget — its header, the notices it owes, its spilled: block — which says so and names the number that clears it. Never truncates the RESULT: an over-ceiling result SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, the row schema, and the epoch) and the response names the file, so what the ceiling held back inline is in the file. 0 = the server default (~80k).")]
             int max_chars = 0,
         [Description("TRANSPORT: return the accounting block and counts only, no rows — the cheap census.")]
             bool counts_only = false,
@@ -1882,7 +1882,10 @@ public static class RecordsTools
 
     /// <summary>The delta form's text render: header counts, then per record the two pole lines, the stack-above
     /// fact stated neutrally rather than as advice, and the delta-line grammar — where a truncated deep read is
-    /// never rendered as 'identical'.</summary>
+    /// never rendered as 'identical'. max_chars is a CEILING here, the shape #603 gave the scan and batch renders:
+    /// the truncation notice and the spill block are charged before the first record is laid, a delta line is
+    /// written only where its own cut notice still fits beside it, and a record that would cross what is left is
+    /// taken back out whole and counted.</summary>
     static string RenderRecordsDelta(IReadOnlyList<LoadOrderService.DeltaRow> rows, int total, int differing, int identical,
                                      int noVerdict, int errors,
                                      string headerLine, OrderStamp? epoch, int maxChars, SpillState? spill, out bool truncated)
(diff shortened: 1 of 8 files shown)


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -2187,27 +2309,31 @@ static string RenderRecordsEffectChains(IReadOnlyList<(string Seed, EffectChainR
...
-                  .Append(" seeds at max_chars=").Append(cap).Append("]\n");
-                break;
-            }
+            int mark = sb.Length;
             sb.Append('\n').Append("seed ").Append(seed).Append('\n');
-            sb.Append(Wire.RenderEffectChain(result, cap, "walk.max_nodes")).Append('\n');
+            // The shared render builds its own buffer, so it is told what this one has already spent — and it
+            // still quotes the caller's max_chars in its own cut notice.
+            sb.Append(Wire.RenderEffectChain(result, room, sb.Length + 1, "walk.max_nodes")).Append('\n');
+            if (Crossed(sb, mark, room.Budget, Notice(rendered), ref truncated)) break;
```
Avick3110, line 2404: A reverse effect chain whose carrier rows are cut inside the shared render now comes back with rows missing, no count of the seeds, and no artifact holding what went.

`Wire.RenderEffectChain` cuts rows itself and reports nothing back about it. The only thing that sets `truncated` in this loop is `Crossed` — and now that the shared render keeps its own output inside `room.Budget`, the seed block no longer crosses, so `Crossed` returns false, `truncated` stays false, and `EffectChainsResponse` writes no auto-spill.

Measured through the tool on `RecordsWorld` (`formids=[<MgefA>]`, `walk={direction:"reverse", follow:"Effects[].BaseEffect"}`, `project.form='chain'`), at every `max_chars` from 685 to 875:

```
1 seed(s), 2 carrier row(s), 0 error(s)  epoch=e2-...

chain for 000805:HcRecMaster.esm (HcRecMgefFire, MagicEffect): 2 carrier rows  epoch=e2-...
Spell (2):
  ... [truncated: rendered 0 of 2 rows before hitting max_chars=685; lower limit= or raise max_chars]
```

635 chars, inside the cap, both carrier rows gone, and no `spilled:` block anywhere in the response. The uncut render of the same call is 729 chars, so at `max_chars=875` an answer that would have fit with 146 chars to spare comes back carrying none of its rows and naming no file that has them. On `main` this call returned the complete answer.

That is the one thing the ceiling contract promises it will not do, in the two sentences this PR itself writes: the new `max_chars=` description ("Never truncates the RESULT: an over-ceiling result SPILLS in full to a server-side JSONL artifact ... and the response names the file") and the changelog line ("what the ceiling held back inline is complete in the artifact the response names"). `JsonWire.RenderEffectChains` sets `outTruncated` when it cuts a carrier, so the same call in `format='json'` does spill — only the text lane loses the rows.

The fix is the shape every other render here already has: `RenderEffectChain` hands back whether it cut (an `out bool`), and this loop ors it into `truncated`.

Nothing catches it: `AReverseEffectChainRenderIsNeverWiderThanItsCap` asserts length only, and the reverse effect chain is the one form left out of `EachOfTheseRendersSaysWhatItsCapHeldBackAndStaysInsideIt`, which is where the `spilled: complete result` assertion lives.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1982,127 +2027,181 @@ internal static string RenderRecordsTree(IReadOnlyList<LoadOrderService.TreeRow>
...
+            if (mute || !said)
+            {
+                if (Crossed(sb, mark, budget, Notice(rendered), ref truncated, force: mute))
+                { declarersLeadWritten = leadMark; break; }
+                rendered++;
+                continue;
             }
+            // The row was laid to the budget and says what it held back: it stays, and nothing more fits.
+            truncated = true;
             rendered++;
```
Avick3110, line 2169: On the `said` path the render breaks without ever writing `Notice(rendered)`, so the rows after the kept-and-cut one are counted nowhere in the render — only the inner notice fires, and it claims a different loss.

Measured through the tool on `RecordsWorld`, `types=["WEAP"] project.form='tree' max_chars=1785` (4 matches):

```
000800:HcRecMaster.esm  Weapon  HcRecW0
  3 plugin(s) touch this record (load order, winner last):
    1. HcRecMaster.esm
    2. HcRecMid.esp
    3. HcRecOverride.esp  (winner)
    ... [nodes cut at max_chars=1785 — raise max_chars or narrow with project.fields=]

spilled: complete result (4 rows) -> ...
```

Three of the four records rendered. The fourth is absent and no notice says a record was held back: the only notice present names record 000800's diff nodes, which is a different thing from a whole record that never rendered. The lines this hunk replaces wrote `... [rendered 3 of 4 rows at max_chars=1785]` on the next turn of the loop, so before this change the response named both losses.

Same shape in `RenderRecordsDelta` (1977–1979 here: `truncated = true; rendered++; break;`) and `RenderRecordsChain` (2284–2286).

The spill block does say the rows above are a prefix of 4, so this is recoverable rather than lost — but the rendered-of-total count, which every other cut path in this family states and which the `max_chars=` description promises as "the truncation notice, the accounting line and the spilled: block are charged before the rows are laid", is charged out of the budget and then not written. If the intent is that a kept-and-cut row ends the render, the notice still has room reserved for it and belongs there.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, reproduced, folded in fb2612f8.

Measured before the fix on the exact call you name, at `max_chars` 685/700/800/875: the uncut answer is 729 chars, and at 800 and 875 the response came back 733 chars with one of the two carrier rows gone, `truncated` false and no `spilled:` block anywhere. At 685 and 700, both rows gone, same silence.

The fix is the one you named. `Wire.RenderEffectChain`'s bounded overload now takes an `out bool cut`, set where it appends its own truncation notice, and the seed loop treats a reported cut like every other cut unit in this family: the seed stays, the seeds below it are counted in `... [rendered N of M seeds at max_chars=...]`, and `truncated` drives the auto-spill. Same call at 875 now names an artifact holding both rows.

One thing the fix does NOT change, and it is worth stating rather than leaving for the next reader. At those caps the answer now comes back with no carrier rows at all and over the ceiling, saying `raise max_chars to at least 1778` — because the first render pass has no spill block to charge, cuts a row, and the second pass then has to charge the ~1,040-char spill block against 875, which leaves zero budget. That band exists because each row is priced with room for a cut notice that a row completing the list would never need. The tree render has the identical property (a 1,675-char answer at `max_chars=1785` comes back cut and spilled, which is what the ten pinned caps in `RecordsOwnedChildTests` encode), so this is the family's shape, not something this lane does alone. Relaxing the reserve on the last unit of each list would fix it everywhere and move most of those pinned caps again; I have not done it here, since it is a different change from the one you found. Say if you want it.

Test: `RecordsArtifactTests.AReverseEffectChainCutInsideTheSharedRenderIsCountedAndSpilled`. Its cap is derived, not pinned — the complete render's length plus 100, so the whole answer fits with room to spare and only the shared render can cut. It asserts the seed count and `spilled: complete result`, and fails on the parent commit (`Assert.Contains() Failure` on the spill marker).

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, reproduced, folded in 81211f7b.

Your call reproduces exactly: `types=["WEAP"] project.form='tree' max_chars=1785` rendered 000801, 000803 and a cut 000800, dropped 000802, and wrote no line saying a record was held back. The `said` path broke after `truncated = true; rendered++;` with the accounting line's room charged out of the budget and then unspent.

All three sites now end through one helper, `Stopped`, which sets `truncated` and writes `Notice(rendered)`. Same call now ends `... [rendered 3 of 4 rows at max_chars=1785]` before the spill block, at 1,771 chars — the same count and the same wording the pre-change loop wrote on its next turn.

One judgement call inside the fix, so it is visible rather than buried: the line is written only when `rendered < rows.Count`. A `said` row that was the LAST row left nothing uncounted, and `... [rendered 4 of 4 rows]` over a complete set of rows would be an accounting line claiming a loss that did not happen — every other site in this family writes the line only where rows were actually dropped. Both arms are pinned by the test.

Test: `RecordsOwnedChildTests.ARowKeptAndCutCountsTheRowsTheRenderNeverReached`, two rows through `RenderRecordsTree` directly, so no absolute path rides in the budget: at 900 CellF is kept and cut, CellC never renders, and the line says `rendered 1 of 2`; at 1906, the first cap CellC's own row fits inside, no line is written. It fails on the parent commit.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -2116,35 +2224,47 @@ static string RenderRecordsChain(IReadOnlyList<LoadOrderService.WalkSeedResult>
...
             {
-                if (sb.Length >= cap)
+                // Composed before it is priced, so the cut notice lands inside the budget rather than a character
+                // past the node that crossed it.
+                string line = "    d" + n.Depth + "  " + n.Key
+                              + (n.Type is not null ? "  " + n.Type + "  " + (n.EditorId ?? "<no editorid>") : "")
+                              + "  [" + n.Status + ']'
+                              + (n.Note is not null ? "  " + n.Note : "")
+                              + "  <- " + n.PulledBy + "\n";
+                if (sb.Length + line.Length + nodesCut.Length > budget)
```
Avick3110: **The cut-notice reserve is charged against the last unit too, so a render that fits is cut, spilled, and comes back with zero rows over the ceiling.**

The last node of the last seed is priced as `sb.Length + line.Length + nodesCut.Length > budget` even though nothing follows it that could be cut, and `budget` (2231, and the same at 1907 / 2042 / 2325 / 2368) subtracts `Notice(rows.Count)` even when every row renders. So a response whose complete output fits inside `max_chars` is cut anyway, `truncated` is set, the result auto-spills — and because the spill block names an absolute path three times (~1.1 KB), the re-render's budget goes to zero and the answer comes back with no rows at all, well over the cap.

Measured on this branch against the artifact fixture, `housecarl_records types=["SPEL"] walk={} project={form:"chain"}` (complete render = 837 chars):

```
cap=835  len=1530  spilled=True   ... [rendered 0 of 4 seeds at max_chars=835]
cap=900  len=1533  spilled=True   ... [rendered 0 of 4 seeds at max_chars=900]
cap=970  len=1533  spilled=True   ... [rendered 0 of 4 seeds at max_chars=970]
cap=975  len=837   spilled=False
```

On `main`, every cap in that band returns the complete 837-char answer inside the cap. The same holds for `project={form:"tree"}` over WEAP — complete 1,675, caps 1,676–1,809 truncate and spill on this branch, `main` returns 1,675 — and by construction for `delta` (1960) and the tree node loop (2113). The band is as wide as the reserved notices, ~137 chars here.

The overrun sentence that ships in that band is also misleading: it says "raise max_chars to at least 1533" when the complete answer is 837 and would have fitted.

Suggested shape: price the final element of a list without its cut notice (there is nothing after it to cut, and if it fits without one no notice is owed), and charge the `[rendered N of M …]` line only where a unit is actually dropped.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -2116,35 +2224,47 @@ static string RenderRecordsChain(IReadOnlyList<LoadOrderService.WalkSeedResult>
...
                 }
-                sb.Append("    d").Append(n.Depth).Append("  ").Append(n.Key);
-                if (n.Type is not null) sb.Append("  ").Append(n.Type).Append("  ").Append(n.EditorId ?? "<no editorid>");
-                sb.Append("  [").Append(n.Status).Append(']');
-                if (n.Note is not null) sb.Append("  ").Append(n.Note);
-                sb.Append("  <- ").Append(n.PulledBy).Append('\n');
+                sb.Append(line);
             }
+            // A seed cut mid-list ends there: what follows the nodes belongs to a seed the budget did not hold.
+            if (said || mute) goto measure;
```
Avick3110: **A seed cut mid-list also loses its `walk truncated` note, its cycles and its TemplateFlags report — and the notice left in their place points at a lever that cannot fix the loss.**

`goto measure` jumps past `row.Cycles`, `row.TruncationNote` and `row.TemplateReport`. `main` wrote all three after the node loop whether or not the nodes were cut.

`TruncationNote` is `walk truncated: the {maxNodes}-node cap was reached — what is listed IS reached and proved; raise walk.max_nodes to walk further.` (LoadOrderService.cs:4077). So on `housecarl_records project={form:"chain"} walk={max_nodes:50}` over a seed whose walk hit the 50-node cap, at a `max_chars` that also cuts the node list, the row now ends with only:

```
    ... [nodes cut at max_chars=N — raise max_chars, or to_file= for the complete walk]
```

The caller raises `max_chars`, or takes the `to_file=` the notice offers, and still gets a walk cut at 50 nodes — with nothing in the response saying the walk itself was incomplete. That is a second incompleteness, on a different axis, dropped in silence, and the surviving notice's remedy is wrong for it.

The NPC `TemplateFlags` inheritance report goes the same way, and that report is the reason `form=chain` is run on an NPC at all.

If the budget genuinely cannot hold them, the honest end is the `mute` path (take the seed back out whole and count it), not a seed that keeps its nodes and quietly sheds the statements about what it lost.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1945,26 +1954,71 @@ static string RenderRecordsDelta(IReadOnlyList<LoadOrderService.DeltaRow> rows,
...
-                        AppendCutNotice(sb, "delta lines", cap);
+                        said = Said(sb, deltaCut, budget);
+                        mute = !said;
                         break;
                     }
-                    sb.Append("    - ").Append(delta).Append('\n');
+                    sb.Append(line);
                 }
-                if (!d.Complete)
+                if (!said && !mute && !d.Complete)
```
Avick3110: **`!said && !mute` suppresses the INCOMPLETE note, which is a claim about a different axis than the cut.**

`d.Complete == false` means the deep read hit the cap, so list-content and one-sided-presence deltas were suppressed **for the whole record** — deltas that were never computed. `said` means the delta lines that *were* computed did not all fit. They are separate facts, and only the second one is stated now.

A record with both, rendered through `housecarl_records project={form:"delta"}` at a cap that cuts its delta lines, ends at:

```
    ... [delta lines cut at max_chars=N — raise max_chars or narrow with project.fields]
```

so the caller reads the shown deltas as a prefix of the differences that were found, raises `max_chars`, and gets the rest of a comparison that still never looked at list contents. `main` appended the note unconditionally after the delta loop.

The tree render keeps its equivalent — the INCOMPLETE clause rides inside each node's `body` string, so it cannot be dropped by a cut — which makes this the one render where the caveat is droppable.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -2221,33 +2362,36 @@ static string RenderRecordsInfoOrder(IReadOnlyList<LoadOrderService.InfoOrderRow
...
                 sb.Append("  no INFO lines — every touching plugin's child list is empty.\n");
-            else if (!DialogueWire.AppendInfoOrderView(sb, row.Order, "", cap, indent: false))
-            {
-                truncated = true;
-                break;
-            }
+            else
+                // The shared view is bounded by what this render has left, not by the whole cap, and its own tail
+                // rides inside the topic block — so a topic that still crossed goes back out whole below.
+                DialogueWire.AppendInfoOrderView(sb, row.Order, "", budget, indent: false);
```
Avick3110: **The view's own stop signal is discarded, and what replaces it holds only by accident of where the marker is written.**

`AppendInfoOrderView`'s contract is "Returns false at the cap so the caller stops" (DialogueTools.cs:79). Dropping it leaves `Crossed`'s `sb.Length > budget` as the only detector, which works today for one reason: the view checks `sb.Length >= cap` *before* a row and appends `... [truncated at max_chars]` only after that check has already been crossed, so the buffer is always past the budget when it returns false.

Give that helper the reserve-the-notice shape this PR gives every other inner list — the obvious next step, and what the rest of this file now does — and it returns false with `sb.Length` still inside the budget. Then `Crossed` returns false, `truncated` stays false, no artifact is written, and the response ships a topic ending in `[truncated at max_chars]` with no `spilled:` block and no way to ask for the rest: exactly the silent-degraded mode the ceiling work exists to remove.

Keeping the signal costs one line — `if (!DialogueWire.AppendInfoOrderView(...)) truncated = true;` beside the `Crossed` call.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -138,7 +138,7 @@ public static string Records(
...
+        [Description("TRANSPORT: character CEILING on the RENDER, hard on every text render this tool has — the scan, batch, resolve, group_by and summary renders, the comparison forms (delta, tree), the walk lane's chain and effect-chain renders, and info_order. The record block, node or delta line that would cross it is not written, and the truncation notice, the accounting line and the spilled: block are charged before the rows are laid. The one answer that can still come back over it is a max_chars too small for what the response carries whatever the budget — its header, the notices it owes, its spilled: block — which says so and names the number that clears it. Never truncates the RESULT: an over-ceiling result SPILLS in full to a server-side JSONL artifact (line 1 = manifest with the query echo, the row schema, and the epoch) and the response names the file, so what the ceiling held back inline is in the file. 0 = the server default (~80k).")]
```
Avick3110: **"hard on every text render this tool has" is wider than what holds: `counts_only` is neither budgeted nor settled.**

Every `counts_only` branch in this tool returns its string straight out — e.g.

```csharp
: $"{headerLine}\ncount={rows.Count} differing={differing} identical={identical} no_verdict={noVerdict} errors={errs}" + Wire.EpochLine(epoch);
```

with no budget and no `RenderCap.Settle`. `headerLine` alone carries the source and selection statements and runs to a few hundred characters, so `housecarl_records types=["SPEL"] counts_only=true max_chars=50` comes back several times over the cap with no `[!] this response is N chars, over the max_chars=50` sentence and no number to raise to — which is the one arm the sentence above says always says so. Same on the delta, tree, chain, info_order and reverse census branches.

Either route those returns through `RenderCap.Settle` (they are one-liners, so it is only the overrun sentence) or narrow the claim to the renders that carry it.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, reproduced exactly, folded in 266bbdbe — but with a wider bound than the shape you suggested, so the difference is stated rather than buried.

Before the fix, on the two cells you name: chain complete = 837, and caps 835/900/970 came back 1533/1539/1539 chars, spilled, `... [rendered 0 of 4 seeds ...]`; 975 came back 837 clean. Tree complete = 1675, and 1676/1700/1785 came back 1530/1530/1536, cut and spilled; 1809 came back 1675 clean. Your numbers, to the character.

**Why not the suggested pricing.** Pricing the last element without its cut notice and charging the accounting line only where a unit is dropped narrows the band but does not close it: the accounting line is still charged against every row except the last, so a render whose earlier rows end inside `cap` but past `cap - notice` is cut even though the whole thing fits. The property worth having is the one your measurement is really about — *if the complete render fits its cap, it IS the answer* — so each of the five renders now lays an unreserved pass first (a budget nothing can cross, which is the complete render) and charges the reserves only when that pass is over the cap. Reserves are room for notices a complete render never writes.

After: 837..974 → the complete 837, unspilled; the whole tree band → 1675. Below the complete length nothing changes — 835 still cuts and spills, which is correct, the answer really does not fit — and the misleading "raise max_chars to at least 1533" over an 837-char answer is gone with the band.

Tests: `RecordsArtifactTests.ARenderThatFitsItsCapIsNotCutByNoticesItNeverWrites` (delta, tree, chain, reverse effect chain, each at its own width and +1/+50/+140, asserting the response is byte-for-byte the complete render, unspilled, with no notice) and `DialogueFamilyTests.AnInfoOrderRenderThatFitsItsCapIsNotCutByNoticesItNeverWrites`. Both fail on the parent commit.

**Test caps that moved, and what moved them.**

- `ARowThatFitsWholeIsNotMarkedTruncated` 889 → 748 and `ARowOneCharacterTooWideIsCutAndSaysSo` 888 → 747. CellC's whole row is 748 chars: the old pair sat 140 above the render's own width, which is exactly the band this closes. The boundary is now the width itself.
- `ASoleProviderRowCutMidBlockSaysTheDeclarersWereCutAndNamesNoDiff` 800 → 700: at 800 CellC now renders whole. Across the whole band where its block IS cut (632..747) the cut lands after the first field line, so the witness lines move from `NavigationMeshes:`/`Temporary:` to `Landscape:`/`Temporary:`. Same claim, same two notices, same branch.
- `ARowKeptAndCutCountsTheRowsTheRenderNeverReached` keeps both caps (900, 1906); only its doc line moved. 1906 is now the first cap CellC's row is reached at all — its block is cut there, and being the last row nothing is left to count, which is still the arm the test pins. The pair fits whole at 2097.
- `AReverseEffectChainCutInsideTheSharedRenderIsCountedAndSpilled` lost its premise: a cut at a cap where the whole answer fits with 100 to spare is precisely the state this removes. And on that fixture no cap reaches a cut inside the shared render at all any more — the complete answer is 729 chars while the auto-spill block it would then have to carry is ~1 KB (three absolute paths), so every cap under 729 leaves the seed zero budget and the seed is rolled back whole instead. So it became a render-level test, `ReverseEffectChainSharedCutTests.ACutInsideTheSharedRenderMarksTheResponseTruncated`, driving `RenderRecordsEffectChains` against a hand-built 8-carrier result one char under its own width: the shared render cuts, and `truncated` — the flag the pipeline spills on — comes back true. It fails when the seed loop ignores the reported cut, i.e. on the state before fb2612f8. `RenderRecordsEffectChains` is internal for that, the same reason `RenderRecordsTree` is.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, folded in 7b391718.

`goto measure` did skip `row.Cycles`, `row.TruncationNote` and `row.TemplateReport`, and you are right about which of the two losses the surviving notice can claim: `raise max_chars, or to_file=` answers a nodes cut and answers nothing at all about a walk that stopped at `walk.max_nodes`.

The fix takes the first of your two options rather than the mute path, because the tail has a known width and can therefore be held back: `SeedTail(row)` composes the cycles, the walk-cap note and the TemplateFlags report up front, that width is reserved beside every node line, and the tail is written whether or not the node list was cut. The mute path is still the end when even that does not fit — the seed goes back out whole and is counted, which is the honest end you named.

One consequence worth stating rather than leaving for the next reader: charging the tail against every node line means a seed carrying a big template report renders fewer nodes at the same cap. That is deliberate. The report and the walk-cap note are statements about what the WALK lost; nodes are content, and the nodes notice already says content went. Trading a node for a statement no notice can make is the right way round.

Test: `ChainSeedTailTests.ASeedCutMidListStillSaysItsWalkHitItsOwnCap` — a hand-built seed (6 nodes, a 2-node walk-cap note, one inherited TemplateFlags category) rendered one char under its own uncapped width, so the node list is cut and both statements must survive. It fails on the parent commit (`Assert.Contains` on the walk-truncated sentence). `RenderRecordsChain` is internal for it, the same reason `RenderRecordsTree` is: no fixture makes a capped walk with nodes enough for max_chars to cut.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, folded in 3e3b40a0 — and it is the same defect as the chain seed's tail, on a different axis, so it gets the same shape.

The note is now composed before the delta loop, its width reserved beside every delta line, and it is written on the `said` path too. Only `mute` drops it, and a mute record goes back out whole and is counted, so nothing ends in silence. A record whose deep read was complete composes an empty note and nothing changes for it.

Test: `DeltaIncompleteNoteTests.ACutDeltaListStillSaysTheComparisonWasIncomplete` — a hand-built row with five delta lines and `Complete: false`, rendered one char under its own uncapped width, asserting both `[delta lines cut` and `the comparison is INCOMPLETE`. It fails on the parent commit.

`RenderRecordsDelta` is internal for that test, the same reason `RenderRecordsTree` already is: the one fixture that produces `Complete == false` (`TruncatedSubFieldWorld`) carries exactly one delta line, so an incomplete read with lines enough to cut has no other way in. It sits next to `TreeIncompleteNoteTests`, which drives the tree render the same way for the same claim.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded in dda67fc5 — one line, `else if (!DialogueWire.AppendInfoOrderView(...)) truncated = true;` — but with a caveat I would rather state than let the commit imply.

It fixes no live defect and no test can fail on it. The view returns false at exactly one place (`DialogueTools.cs:156`): `if (sb.Length >= cap) { append the 32-char marker; return false; }`. The marker is appended after the check, so on every false return the buffer is strictly past the budget and `Crossed`'s `sb.Length > budget` agrees, always. I did not write a test, because a test that cannot fail is not a test (TESTING.md), and I did not weaken anything to manufacture one.

I folded it anyway because the coupling you name is real: this render should not depend on where a helper it does not own writes its marker, and the day that helper gets the reserve-the-notice shape the rest of the file now has, the agreement ends silently. If you would rather not carry an untested line, it is one line to drop and I have no argument against that.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real, folded in 6a62dfe5 — routed rather than narrowed, since the wider sentence is the better contract and the census is a render like any other.

All seven text census returns — the identity list lane, the batch list lane, the reverse carrier census, the forward chain census, delta, tree and info_order — now go through one local `Census(body)` that settles the string against the caller's max_chars. The json arms are untouched; the sentence is about text renders.

`types=["SPEL"] counts_only=true max_chars=50` now ends with `[!] this response is N chars, over the max_chars=50 it was given ... raise max_chars to at least N`, which is the one arm the paragraph above it promises always says so.

Tests: `RecordsArtifactTests.ACensusTooBigForItsCapSaysSoAndNamesTheNumberThatClearsIt` walks six census sites at max_chars=50 and asserts each is over the cap, says so, and names its own length; `DialogueFamilyTests.AnInfoOrderCensusTooBigForItsCapSaysSoAndNamesTheNumberThatClearsIt` covers the seventh. Both fail with `Census` as the identity function.

The `max_chars=` description now names the census in its list, and the changelog entry says it too.

PR #636, opened 2026-09-08
Rewrite the skypatcher-authoring skill

Avick3110:
Wave 1 of the skill rewrite: `skypatcher-authoring`, built from the approved rewrite document with the six rows its verifier overturned corrected as the verifier says. The skill's verify step named `housecarl_skypatcher_read`, which does not exist on the 2.0 surface, so anyone following the skill literally made a failed tool call; every houseCARL mention now migrates to the call the migration table gives — `housecarl_records` for the target set, each address, the draft file and the pre/post delta, `housecarl_skypatcher_layer` for the file-level verdict, `housecarl_skse findings='config'` for the reference verdicts. The record-type router moved out of `references/index.jsonl` into the body so every reference pointer is one hop; the reference files are flat under `references/`, `reference.md` is renamed `placed-reference.md`, the file-and-folder half of `grammar-core.md` split into `placement-and-conflicts.md` (a one-line patch no longer pays for the whole syntax spine), and `index.jsonl` plus the packager-excluded `_CORPUS_STATUS.md` are gone with their two load-bearing sentences inlined into the body. The description gains the version floor, the front-loaded load rule and the SPID/KID/CID exclusion; the frontmatter gains the `compatibility` line from ruling 7. The thirteen lines describing the layer reader's output (F43) are deleted, the distributor routing table is not restated (F1, already in ServerInstructions), and the four worked examples that carry another record type's filter are marked where they sit rather than rewritten, so the corpus keeps its verbatim-from-the-article guarantee. `evals/eval_set.json` becomes `evals/evals.json` on the published `prompt`/`expected_output`/`expectations` field names, with 32 trigger cases (17 positive, 15 negative), a without-skill arm, per-case evidence, and the with/without output-quality task on the live order that the 2026-09 audit's veto skipped.

- **Q1, the shared `compatibility` wording — left for Aaron.** Written as the document proposes: "Requires the houseCARL MCP server and a configured Mod Organizer 2 instance." (76 characters). If Aaron supplies a different canonical string, this file and the other eleven change together.
- **Q2, router in the body or `index.jsonl` kept — decided: router in the body.** Ruling 2 wants one hop; the 28-row table costs ~30 lines per load and pays for the deleted index. The drift guard that read the index now reads the table (below).
- **Q3, `_CORPUS_STATUS.md`'s disposition — decided: deleted from the skill.** It was already stripped from every shipped copy, and its two load-bearing sentences (provenance/version floor; the primary filter beats a stray example) are inlined into the body. Git history is the ledger. Moving a copy into `dev/` instead is still open if Aaron wants one.
- **Q4, correct the four bad examples or mark them — decided: marked, not corrected.** Each affected file gains one "Corpus note" block under its examples naming the wrong filter and the right one. The corpus keeps "every worked example is preserved verbatim from the source article"; nothing ships a silently wrong line.
- **Q5, `housecarl_skse findings='config'` in the verify step — decided: kept.** It is the only machine check for the skill's own "truncated or wrong FormID" mistake, and it sits after the install with the layer reader, two lines.
- **Q6, the description at 606 characters — decided: all four additions kept.** Every cap is far off (1,024 description, 1,536 listing, 702 measured). The trigger set is re-measured either way, and six cases were added for the changed clauses.
- **Q7, who owns the SkyPatcher version bump — left for Aaron.** The floor is stated in the description as the document specifies, which makes staleness more visible; no owner is named anywhere, so this is still a real question.

## Judgement calls a reviewer should check

- **#613 landed before this branch, so the skill uses it instead of naming it as a gap.** The document was written when the offline draft check did not exist and told the body to name the gap by issue number. #613 was closed by #618, which is on this branch's base: `housecarl_records`'s overlay pole takes `"ini"` and `"subfolder"` on both `source=` and `versus=`, and the replay's per-line warnings render beside the answer with the draft's path as the file name. The "Check before you install" step therefore reads the draft rather than pointing at an unlanded issue, and migration rows 3 and 6 — the two the document marked **not expressible — #613** — are now expressed by the shipped surface rather than written around. Everything else in the migration table is unchanged.
- **The `skypatcher-catalog-guard` CI probe read `references/index.jsonl` and now reads the router table.** Deleting the index broke the probe's drift check (name/sig/subfolder/primary filter, catalog against reference). The same four columns are in the body table, so the probe parses that instead; the comparison is unchanged except that a row's signature and primary filter are matched component-wise, which the `LVLI / LVLN` row already needed. Three router rows were spelled to match the catalog's own names: "Reference (REFR)" rather than "Placed Reference" (the file is still `placed-reference.md`), "Race Hook (RACE)", and OMOD's subfolder written as `objectModification` *(unconfirmed)*, which is what the corpus records.
- **The body measures 173 lines / 11.9k characters / roughly 2,900 tokens against the document's ~147-line budget** (the verifier corrected 137 to 147). Inside the 500-line ceiling and the 5,000-token advisory. The overage is blank lines around the table and the draft-check bullets.
- **The eval file is authored, not run.** It says so at the top. The shipped 26 cases carry their last measured verdicts as a baseline; the six new ones have never been measured, and the description they measure changed here.

Build: `dotnet build housecarl.sln -c Release` clean, `dotnet test --filter "tier!=bridge"` 1793/1793 passed, `ci-all` 128/128 passed, `scripts/build-plugin.ps1` full package built with the leak check and the markdown pointer scan clean (12 skills, 251 markdown files, across 2 trees).

## Regrade

Assembled in `dev/skill-validation/eval-2026-09/skypatcher-authoring/REGRADE.md` from five blind grader passes over this branch's skill tree.

| Angle | Criteria | Conforms | Partial | Fails | n/a | Fails by class |
|---|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 36 | 0 | 1 | 5 | packager-only 1 (D1.28, excluded by ruling 8) |
| D2 — frontmatter | 52 | 39 | 0 | 0 | 13 | — |
| D3 — loading and invocation control | 33 | 29 | 0 | 0 | 4 | — |
| D4 — body, supporting files, style | 64 | 44 | 2 | 2 | 16 | factual-or-dead-call 1 (D4.43); other 1 (D4.42) |
| D5 — correctness and evals | 52 | 31 | 14 | 1 | 6 | factual-or-dead-call 1 (D5.51) |
| **Total** | **243** | **179** | **16** | **4** | **44** | factual-or-dead-call 2, other 1, packager-only 1 |

**The regrade's two counting fails were graded against a stale surface capture.** D4.43 and D5.51 are the same defect from two angles: `SKILL.md:110-116` calls `housecarl_records` with `source=`/`versus=` carrying `ini` and `subfolder`, and both graders read `dev/skill-validation/eval-2026-09/surface/tools.md` and nothing else, where those keys were absent. That capture was taken on 2026-09-07 before #618 merged, so it predates the pole the body uses.

The surface is now recaptured from the tip (`e2990653`): built in the worktree, server started over stdio with a fresh empty `HOUSECARL_DATA_DIR`, `initialize` at protocol `2025-06-18` then `tools/list`, rendered in the same shape. 31 tools, 179 parameters — the draft check added a value to an existing pole, not a parameter. `ini` and `subfolder` are on `housecarl_records.source` (`tools.md:1402`) and on `versus` (`:1408`, "`{\"overlay\", \"state\"[, \"ini\", \"subfolder\"]}`"). Thirteen other parameter descriptions moved with the tip as well. The date and tip sha are now stated at the top of `tools.md`.

Against the refreshed surface the body's draft-check claims hold, with one correction folded in this PR: the delta that isolates a draft's effect reads the draft as the **subject** — `source=` the draft pole, `versus={"overlay": "skypatcher", "state": "post"}` — and the step had the two poles the other way round, which reports the change backwards. The bullet also now says `subfolder` can be dropped when the draft already sits in a folder of that name. The dependent sentences are unchanged and check out: the replay's warnings render beside the answer on the read, delta and tree lanes (`RecordsTools.cs`, `StateOverlayWarnings`), each carrying its own file and line, which for a draft is the draft's path.

D4.42 (bare tool names) is recorded as a fail in the checklist's literal terms only; ruling 1 retires it, and it is not a factual or dead-call fail. D1.28 is packager-only and excluded by ruling 8.

## Regrade

Final assembly, replacing the counts above: `dev/skill-validation/eval-2026-09/skypatcher-authoring/REGRADE.md`, rebuilt from five fresh grader passes (`regrade/D1.md` … `D5.md`, 2026-09-08) taken against the recaptured `surface/tools.md`.

| Angle | Criteria | Conforms | Partial | Fails | n/a | Fails by class |
|---|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 36 | 0 | 1 | 5 | packager-only 1 (D1.28, excluded by ruling 8) |
| D2 — frontmatter | 52 | 40 | 0 | 0 | 12 | — |
| D3 — loading and invocation control | 33 | 29 | 0 | 0 | 4 | — |
| D4 — body, supporting files, style | 64 | 47 | 1 | 0 | 16 | — |
| D5 — validation and standing context | 52 | 32 | 14 | 0 | 6 | — |
| **Total** | **243** | **184** | **15** | **1** | **43** | packager-only 1, factual-or-dead-call 0, other 0 |

One fail across all five angles: D1.28 (release tags are `v1.9.0`, not `housecarl--v{version}`), which is packager-only, outside the skill's files, and excluded from the gate by ruling 8. D4.43 and D5.51 no longer fail — against the recaptured surface the draft-replay call is expressible as written — and D4.42 conforms under ruling 1. D2 and D3 carry no partial at all; D4's one partial is D4.50 (a single input/output pair against the platform's "3–5"); eight of D5's fourteen partials are the one fact that `evals/evals.json` is authored and not yet run.

Recorded but not a fail: the per-line replay-warning claim at `SKILL.md:116-118`, `:131`, `:141-142` is a diagnostic `surface/tools.md` does not itself promise (D5.51 grades it partial — the call still answers; the risk is a stop condition satisfied vacuously).

**Verdict: PASS.** Zero fails on D2 and D3; zero factual-or-dead-call fails on D4 and D5; D1 clean but for the excluded packager-only D1.28.


.claude/skills/skypatcher-authoring/SKILL.md
```diff
@@ -1,138 +1,173 @@
 ---
 name: skypatcher-authoring
 description: >-
-  Author or interpret SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set / add / remove their properties. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons / armor / NPCs / leveled lists without an ESP, or asking why a patch line isn't applying. Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing and filter names are non-obvious.
+  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, or asking why a patch line isn't applying. Not for distribution — forms onto NPCs is SPID, keywords onto items is KID, items into containers is CID.
```
Avick3110, line 4: **The new description tells the matcher to decline a job this skill fully covers.**

The added clause ends `… items into containers is CID.` But SkyPatcher has a Container patcher and this skill ships its reference: `references/container.md` documents `filterByContainers` with `addToContainers=<form~count>`, `addOnceToContainers`, `replaceInContainers`, `removeFromContainers` and `clear=true`; the router table below carries the `Container (CONT) | container | filterByContainers` row; and the shipped `skypatcher-catalog.json` carries the same row, which the probe in this PR now cross-checks against that table.

The description is the entire trigger surface. Concrete failure: a user asks "add 100 lockpicks to the Riverwood general store chest, no ESP" — the exclusion clause reads as an explicit non-trigger, `skypatcher-authoring` is not loaded, and the user is routed to a distribution framework for a one-line `filterByContainers=…:addToContainers=…` patch the skill has a whole reference file for. The old description carried no such clause; the old body mentioned CID only inside the loaded skill, where `container.md` sits one hop away to correct it.

SPID and KID are clean exclusions (SkyPatcher has no NPC-form or item-keyword distribution). Containers is not — suggest dropping that third item, or narrowing it to what SkyPatcher genuinely cannot do (distributing to containers *by rule* across the order, rather than editing a named container's object list).

.claude/skills/skypatcher-authoring/evals/evals.json
```diff
@@ -0,0 +1,305 @@
...
+      "prompt": "give every bandit a spell at runtime, no esp",
+      "expected_output": "Route to the SPID skill — forms onto NPCs is not a SkyPatcher job.",
+      "expectations": ["does not pick skypatcher-authoring", "routes to the forms-onto-NPCs skill"],
+      "should_trigger": false,
+      "evidence": "NEW 2026-09-08, never measured. The 'runtime, no esp' phrasing is the lure; the exclusion clause is what should decide it."
+    },
+    {
+      "id": 29,
+      "prompt": "put these items into a chest without an ESP",
+      "expected_output": "Route to the CID skill — items into containers is not a SkyPatcher job.",
```
Avick3110, line 260: **This negative case encodes the container carve-out as a hard fail, and names a destination that does not exist.**

Two problems:

1. "put these items into a chest without an ESP" is a SkyPatcher job. `references/container.md` in this same skill answers it with `filterByContainers=<chest>:addToContainers=<form~count>`. Grading "does not pick skypatcher-authoring" as the correct outcome measures the wrong answer, and if the suite is ever run and tuned to pass, the description gets pushed further away from container patching.

2. The expectation `routes to the items-into-containers skill` has no target. The eleven shipped skills (`plugin/CHANGELOG.md`, and `.claude/skills/`) contain no CID skill: `bulk-record-jobs, dialogue-authoring, facegen-diagnostics, kid-authoring, mutagen-reference, npc-appearance-copy, oar-authoring, papyrus-reference, skse-plugin-authoring, skypatcher-authoring, spid-authoring`. So the case cannot pass as written even on its own terms — unlike cases 11/12/25/27/28, whose SPID and KID targets both ship.

Cases 27 and 28 are good negatives and should stay. This one should be dropped, or turned into a *positive* case for the container patcher.

.claude/skills/skypatcher-authoring/references/enchantment.md
```diff
@@ -37,3 +37,8 @@ chargeTime=2.5
 enchantmentAmount=24
 clear=true
 ```
+
+> **Corpus note — one example above is wrong.** The `filterByAlchs=` line is an Ingestible example
```
Avick3110, line 41: **A fifth carried-over-wrong example was missed — `references/ingredient.md` line 28.**

That file's examples block contains:

```ini
filterByAlchs=Skyrim.esm|366BF:mgefsToChange=Skyrim.esm|397E~null~10~null~null
```

It is the same line, verbatim, that this note flags in `enchantment.md`: `filterByAlchs` is ALCH's primary filter, INGR's is `filterByIngs`, and `ingredient.md`'s own Filters list does not carry `filterByAlchs`. Cross-checking every record file's examples against its own declared filter list turns up exactly five wrong-record-type filters — `enchantment.md`, `magic-effect.md`, `scroll.md` (two) and `ingredient.md` — and only the first four got a note. (`ammo.md`'s `filterByKeywords`, `location.md`'s `filterByKeywordsOr` and `spell.md`'s `filterByFlagsOr/Excluded` are not this class: the latter two are declared connectives.)

The PR's stated invariant is "nothing ships a silently wrong line", and Q4's ruling was mark-don't-correct. `ingredient.md` needs the same block.

.claude/skills/skypatcher-authoring/SKILL.md
```diff
@@ -1,138 +1,173 @@
...
+compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
```
Avick3110, line 5: **Nothing reads `compatibility:`.**

It is the only non-`name`/`description` frontmatter key across all eleven skills, and no consumer in this repo looks at it: `PluginValidateProbe.ValidateSkillFrontmatter` only asserts the block parses and carries a non-empty `name` and `description`, and `scripts/build-plugin.ps1` copies the file verbatim. Per the note this repo already wrote for itself (`plugin/CHANGELOG.md:2236`), the harness surfaces `name` + `description` and nothing else — so this line ships as bytes the user never sees, in the one file that pays for them in every packaged copy.

The risk beyond dead weight: if any loader in the chain (Claude Code or the Codex path) validates the frontmatter against a closed key set, an unknown key drops the whole block — which is the exact silent failure `PluginValidateProbe`'s header describes, and its YAML-parses check would not catch it. Worth confirming against a real load before this becomes the shared convention on "the other eleven".

.claude/skills/skypatcher-authoring/SKILL.md
Avick3110: **Refuted — the rewrite document prescribes this clause verbatim.**

`REWRITE.md:100` gives the description as a single literal string ending "Not for distribution — forms onto NPCs is SPID, keywords onto items is KID, items into containers is CID.", and `:128-129` names that clause as item 5 of the five description additions, closing D3.9, D3.11 and D4.63 together. It is not a judgement call this PR made; changing it would put the shipped description out of step with the approved document and with the eval cases §9c specifies against it.

The observation underneath is real and worth a ruling, so recording it rather than burying it: `references/container.md` does ship `filterByContainers` with `addToContainers=<form~count>`, so "add 100 lockpicks to a named chest, no ESP" is a job the skill covers, and the clause reads as a non-trigger for it. The document's own eval evidence for case 29 acknowledges the tension — "SkyPatcher does patch CONT records, so the boundary is what RECEIVES the change" — and draws the line at rule-based distribution across the order (CID) versus editing a named container's object list (SkyPatcher). Whether that line survives contact with the matcher is exactly what the trigger arm measures, and the case is authored, not yet run.

Aaron's call. If the clause should narrow to "distributing items into containers by rule is CID", it is a one-line change here, in `REWRITE.md:100`, and in case 29 — but it also changes the shared wording template the other eleven skills inherit, so it belongs to him and not to this fold.

.claude/skills/skypatcher-authoring/evals/evals.json
Avick3110: **Half folded, half refuted.**

Point 2 is real and folded in 6ccff7e. `routes to the items-into-containers skill` is unsatisfiable — no CID skill ships, unlike cases 27 and 28 whose KID and SPID targets both do — so the case failed by construction even when the description worked exactly as designed. Case 29 now grades on `does not pick skypatcher-authoring` alone, and its `evidence` says why the routing expectation is absent where 27 and 28 carry one.

Point 1 is refuted. The case is not a judgement call this PR made: `REWRITE.md` §9c names the three added negatives by their prompt text, "put these items into a chest without an ESP" among them, as the cases that measure the new exclusion clause. Dropping it or flipping it positive would leave that clause unmeasured, which is the opposite of what §9c asks for. The substantive question — whether the clause should exclude container patching at all — is answered on the sibling thread against `SKILL.md:4`, and it is Aaron's to rule; if he narrows the clause, this case moves with it.

.claude/skills/skypatcher-authoring/references/enchantment.md
Avick3110: **Folded in 2b4343c.** Confirmed by cross-checking every `references/*.md` examples block against that file's own declared filter list: exactly five files carry a filter their list does not declare — `enchantment.md`, `magic-effect.md`, `scroll.md` (two) and `ingredient.md` — and only the first four had a note. `ingredient.md` now carries the same block, worded from `enchantment.md`'s and naming `filterByIngs` as INGR's primary filter.

Your exclusion of `ammo.md` is right and I kept it out: `filterByKeywords` there is a generic filter missing from that file's list, not another record type's primary filter, so it is a corpus gap rather than a wrong line and marking it would say something untrue about it.

.claude/skills/skypatcher-authoring/SKILL.md
Avick3110: **Refuted — ruling 7 puts the line there, knowing nothing reads it.**

`REWRITE.md:82` lists the `compatibility:` line as ruling 7's, with the disposition spelled out in the same row: "Both lanes; inert on Claude Code, silent on Codex, spec-legal everywhere (D2.35)". "Nothing reads it" is the ruling's own premise, not a defect it missed; `:101` gives the exact string and `:111` measures it against D2.35's 500-character cap. The only part still open is the wording, which is Q1 at `:431` and left for Aaron — if he supplies a different canonical string, this file and the other eleven change together.

The silent-drop risk does not apply here. `compatibility` is not an unknown key: `REWRITE.md:92-93` places it inside the skill spec's closed six alongside `name` and `description`, which is why the ruling could add it without a Claude Code extension key and keep the file legal on both lanes. A loader validating against a closed key set accepts it.

.claude/skills/skypatcher-authoring/SKILL.md
Avick3110, line 4: **The exclusion clause denies three things the skill''s own corpus documents.**

`Not for distribution — forms onto NPCs is SPID, keywords onto items is KID, items into containers is CID.`

All three are shipped SkyPatcher operations in this skill''s references:

- `references/container.md` — `addToContainers=<form~count>`, `addOnceToContainers`, `replaceInContainers`
- `references/npc.md` — `spellsToAdd`, `perksToAdd`, `objectsToAdd`, `factionsToAdd`, `shoutsToAdd`
- `keywordsToAdd` on weapon/armor/misc

Concrete failure: "add 500 gold to every bandit chest, no ESP". The description says items-into-containers is CID, so the matcher declines this skill and routes to CID — for which no skill ships (`evals/evals.json:264` concedes exactly this) — while `filterByContainers=…:addToContainers=…` is the documented answer sitting in `references/container.md`.

The pre-rewrite body carried the qualifier that made this a routing heuristic rather than a capability denial: *"If the user names SPID/KID/CID or asks to 'distribute' rather than 'edit/patch,' this is the wrong skill"*, plus the Common-mistakes bullet *"Check the verb ('distribute' vs 'edit/patch/buff') and any named framework."* Both are gone, and the description is the entire trigger surface.

Suggest restoring the qualifier — the boundary is the user naming a framework or the verb "distribute", not the shape of the change.

.claude/skills/skypatcher-authoring/evals/evals.json
```diff
@@ -0,0 +1,305 @@
...
+      "id": 28,
+      "prompt": "give every bandit a spell at runtime, no esp",
+      "expected_output": "Route to the SPID skill — forms onto NPCs is not a SkyPatcher job.",
+      "expectations": ["does not pick skypatcher-authoring", "routes to the forms-onto-NPCs skill"],
+      "should_trigger": false,
+      "evidence": "NEW 2026-09-08, never measured. The 'runtime, no esp' phrasing is the lure; the exclusion clause is what should decide it."
+    },
+    {
+      "id": 29,
+      "prompt": "put these items into a chest without an ESP",
```
Avick3110, line 259: **This case grades the correct answer as a failure, and routes to a skill that does not exist.**

`put these items into a chest without an ESP` is `filterByContainers=…:addToContainers=<form~count>` — documented in this skill''s own `references/container.md`. The case marks it `should_trigger: false` and expects a route to "the CID skill", while its own evidence line concedes `No CID skill ships`.

That makes the pass condition "decline a job the shipped reference answers, and leave the user with no destination." Since this suite is the instrument the description is tuned against, running it drives the description further from the corpus rather than closer.

Cases 27 and 28 are defensible — both name a distribution verb and both have a real destination. 29 has neither.

.claude/skills/skypatcher-authoring/SKILL.md
```diff
@@ -1,138 +1,173 @@
...
+  the draft's path.
+
+Stop when the count is what you meant, every address resolves, the delta is the change you intended,
+and the replay warned on none of your lines.
+
+## Check after you install
+
+Once the INI is installed, three reads prove it landed.
+
+- `housecarl_skypatcher_layer` with `filter=` the type folder or the INI filename, for the
```
Avick3110, line 127: **The post-install check dropped the three ITM classes the old verify step required.**

This bullet asks the layer read for "whether the file is read at all, where it sorts, and what it conflicts with". The pre-rewrite step 6 also required: no intra-file dead write, no cross-INI duplicate, no no-op write — "All three are authoring slips to fix at the source."

`housecarl_skypatcher_layer` still reports all three (`SkyPatcherTools.cs:135-156` dead writes, `:161` duplicates, `:183` no-ops), and nothing else in the rewritten body mentions them.

Concrete failure: an author writes `filterByWeapons=A.esp|800:attackDamage=10`, then later in the same file an unfiltered `attackDamage=20`. The draft delta shows 20 — the intended value. The layer verdict shows the file read, sorted, no set-conflict. Every stop condition this skill lists passes, while the tool has already flagged the first line as a fully dead write. A delta is exactly the view a dead write hides behind, so the "Check before you install" step does not cover it either.

src/housecarl-generator/SkyPatcherCatalogProbe.cs
```diff
@@ -13,9 +13,10 @@ namespace HousecarlGenerator;
 //  filter (with connective) and a known operation correctly, flags an
 //  unknown key as Unknown (bundled-or-warn), preserves the OMOD gap, and
 //  cross-checks the record dimension (name/sig/subfolder/primaryFilter)
-//  against the skypatcher-authoring reference's index.jsonl — the drift
-//  guard for when the skill reference updates. In-process; the catalog is
-//  an embedded resource, the index is read from the repo (CWD-relative).
+//  against the skypatcher-authoring skill's own router table in SKILL.md —
```
Avick3110, line 16: **A sibling comment was left pointing at the deleted file.**

This header now correctly names SKILL.md''s router table, but `src/housecarl-core/SkyPatcherCatalog.cs:18` still reads:

> the record dimension (recordType / subfolder / sig / primaryFilter) is cross-checked in CI against the reference `index.jsonl`

`references/index.jsonl` is deleted in this PR. Someone changing the catalog reads that line, greps for `index.jsonl` under the skill, finds nothing, and concludes the drift guard was retired — then changes a `recordType`/`sig`/`subfolder` without touching the router table and gets a `ci-all` failure the comment cannot help them locate.

.claude/skills/skypatcher-authoring/SKILL.md
Avick3110: **Folded in 6fd31cc9.** Confirmed on both halves: the reader still reports all three classes — dead writes at `SkyPatcherTools.cs:133-156` (with the fully-dead-only caveat at `:156`), cross-INI duplicates at `:159-178`, no-ops at `:181-196`, and the summary line at `:68-72` — and the rewritten body named none of them, so the stop conditions the skill lists all passed on a file the reader had already flagged.

The layer bullet now reads (`SKILL.md:127-131`):

> whether the file is read at all, where it sorts, what it conflicts with, and which of its ITM findings (an intra-file dead write, a cross-INI duplicate, a no-op write) name your file. The first two are authoring slips to fix at the source, and a dead write hides from the delta below, which shows only the value that survived it.

Four lines added, not the thirteen F43 deleted: the classes are named as things to look for, the caveats and the per-class explanations stay in the tool's own description, which is where F43 put them. The one-clause enumeration was already there — it listed three of the four things the reader reports, and the missing one was the only class a delta cannot show.

src/housecarl-generator/SkyPatcherCatalogProbe.cs
Avick3110: **Folded in 95bc8553.** `SkyPatcherCatalog.cs:17-18` did still name the deleted `index.jsonl`, and it also listed three columns where the probe checks four. Both fixed in one sentence:

> the record dimension (name / sig / subfolder / primaryFilter) is cross-checked in CI against the router table in the `skypatcher-authoring` skill's `SKILL.md`.

That is now the same four columns and the same source the probe header names (`SkyPatcherCatalogProbe.cs:14-19`). A grep for `index.jsonl` under the skill tree returns nothing after this PR, so the old line was the last pointer at it from the code.

.claude/skills/skypatcher-authoring/SKILL.md
Avick3110: **Left for Aaron — the clause is the approved document's, and the same string is wave 1's shared template.**

The observation is verified and I am not disputing it: `references/container.md:13` ships `addToContainers=<form~count>` under `filterByContainers` (`:8`), `references/npc.md:79` ships `spellsToAdd` / `shoutsToAdd`, and `keywordsToAdd` sits on the weapon, armor and misc references. So all three named jobs are things SkyPatcher does, and read as a capability statement the clause is wrong on all three.

What I cannot do inside this fold is change it. `REWRITE.md:100` gives the description as one literal string ending in that clause, and `:129-132` names it as item 5 of the five description additions, closing D3.9, D3.11 and D4.63 together; the approved document outranks a review finding. It is also not one skill's line: ruling 7 (`DECISIONS.md`, 2026-09-07) makes the frontmatter wording shared across all twelve rewrites, and §9c of the document specifies the three negatives that measure this clause, so narrowing it moves the document, this description, eval cases 27-29, and whatever the other eleven inherit.

The restore you suggest has a second collision worth your eye before you rule: the pre-rewrite qualifier lived in the body, and the body half of it is what FOLD-IN F1 moved into ServerInstructions (ruling 4, 2026-09-07) — "which framework owns a job is decided by what RECEIVES the change". Putting the verb heuristic back into the body would re-restate the fold that F1 collapsed. If the boundary should be "the user names a framework, or says distribute", the natural home is the description clause itself rather than a restored body bullet, which would also spare the other eleven a body-level exception.

Recorded for you as an open call; the eval sibling (case 29) is answered on its own thread and moves with whatever you rule here.

.claude/skills/skypatcher-authoring/evals/evals.json
Avick3110: **Half already folded, half left for Aaron.**

The unsatisfiable-target half is folded, in `1562b3e3` (the blind review raised the same point on the sibling thread). Case 29 no longer routes anywhere: `evals.json:261` is `["does not pick skypatcher-authoring"]` alone, and `:263` says why the routing expectation is absent where cases 27 and 28 carry one — no CID skill ships. Your read of the file matches what is there; the "routes to the CID skill" wording you quote is only in `expected_output` (`:260`) now, not in a graded expectation.

The other half — whether the case should exist at all — is the description clause, and I have left that for you on the `SKILL.md:4` thread. The case is not a judgement this PR made: `REWRITE.md:408` names "put these items into a chest without an ESP" by its prompt text as one of the three negatives that measure the new clause. Dropping or flipping it while the clause stands would leave the clause unmeasured. If you narrow the clause, this case moves with it in the same change — and it is authored, never run, so nothing has been tuned against it yet.

PR #637, opened 2026-09-08
Rewrite the OAR skill and rename it open-animation-replacer

Avick3110:
Wave 1 of the skill rewrite, for the OAR skill only. It carries two things: the rewrite of the skill's body, reference and evals, and the rename of the skill itself from `oar-authoring` to **`open-animation-replacer`**. This is the sweep the approved rewrite document asks for, not a rebuild: the OAR content carries forward and everything around it is redone.

## The rename

Aaron's call 5 of 2026-09-08 takes the rewrite document's Q2. The slash command is now `/housecarl:open-animation-replacer`; `/housecarl:oar-authoring` no longer exists. The reason is the D3.8 failure mode: a host that lists skills by name alone, with descriptions dropped, showed the old skill as "oar", which reads as the English word rather than the mod. The new name says which tool it is without the description. The name taken is `open-animation-replacer`, not the document's draft `open-animation-replacer-authoring` — the shorter form names the tool and drops a suffix that carries nothing.

Done whole: the folder under `.claude/skills/`, the frontmatter `name`, the body's H1, `evals/evals.json`'s `skill_name` and every case that names the skill, the sibling route in `papyrus-reference`'s eval set, the `$Skills` ship list in `scripts/build-plugin.ps1`, the Codex umbrella's skill list and router (`plugin/codex/housecarl/SKILL.md`), `README.md` and `plugin/README.md` (entry, credit, and the `/housecarl:` list), and both `THIRD-PARTY-NOTICES.txt` copies where DrHeisen is credited. The description keeps its "Open Animation Replacer (OAR)" wording and nothing the skill teaches changed. `plugin/CHANGELOG.md` carries a line under `## Unreleased` saying the slash command changed and that the old one is gone. Historical changelog entries under released versions keep the old name, because that is the name that shipped.

## The rewrite

The one step the skill delegated to a tool named `housecarl_read_record` and `housecarl_cross_plugin_query`, neither of which the server publishes any more, so that step was a guaranteed failed call; it now reads a condition's form with `housecarl_records` — `formids=["XXXXXX:Plugin.esp"]` or `types=[…]` plus `where=[…]`, projected `{"form":"identity"}` — and states the fallback for when the server is not there. Locating a submod was Glob and Grep, which cannot say which mod wins a file, so the first step and the verification step now sweep the OAR and DAR trees with `housecarl_asset_status` `under=`, which names the winning source per `config.json`; that call was the single most valuable one in the audit run and the skill did not suggest it. The reference gains the two things the inherited verdict named: a DAR-function → OAR-condition table, so `IsEquippedRight` → `IsEquipped` plus `"Left hand": false` is written down rather than inferred, and an explicit statement that how `AND` and `OR` bind in a mixed `_conditions.txt` chain is unsettled. Around those: a boundary clause in the description and the body (`.hkx`, Nemesis/FNIS behaviour files, and record edits routed to `housecarl:skypatcher-authoring` / `housecarl:spid-authoring`), a `compatibility` line, per-topic reference pointers carrying their section number, a second worked example that converts a real DAR folder end to end, a verification section that checks offline before it names the in-game Detected Problems panel as the enforcement, a loop and a stop condition, the overrides-mod rule and the "houseCARL can't see OAR configs" fact each stated once instead of three and two times, and the `IsEquippedType` values living only in the reference with the warning in the body. `references/index.jsonl` is deleted and `evals/eval_set.json` becomes `evals/evals.json` on the published field names.

No issue is named by the rewrite document as closing, so there is no `Closes` line.

**Decided as follows.**

- **Q1 — the `AND`/`OR` precedence rule.** Shipped as an explicit "unsettled" note in reference §8: no source available to this repository states it, both readings parse and load cleanly, so the instruction is to convert under one reading, say which in the submod's `description`, and say what the other would have meant. The body's DAR example does exactly that. Whether to authorise fetching OAR's source so the rule can be written source-verified is **left for Aaron** — that is a fetch decision, not a writing one.
- **Q2 — the rename.** **Taken**, on Aaron's call 5 of 2026-09-08, as `open-animation-replacer`. See the rename section above; it churns `plugin/codex/housecarl/SKILL.md`, `plugin/README.md`, `README.md`, both notices files, the ship list and any typed `/housecarl:oar-authoring`, and all of that is done on this branch.
- **Q3 — `references/index.jsonl`.** Deleted (the document's own recommendation, Branch A). No line of the body named it, its fifteenth row pointed back out at `../SKILL.md`, and the per-topic section-numbered pointers do its routing job inside the body the agent already has. It is one file in git history if you want it back; the counter-argument the document records — three sibling grammar skills advertise theirs — is a batch-consistency call you may want to make once for all twelve.
- **Q4 — the `compatibility` wording.** Written as the document drafts it: "Requires the houseCARL MCP server and a configured Mod Organizer 2 instance." **Confirm it as the batch string** — twelve documents need the same one, and this PR only sets it here.

**Left for Aaron.**

- **Q5 — where the path-redacted outcome record lives.** Nothing written. `evals/evals.json` records the tension rather than resolving it: the pre-rewrite 9-config round-trip has no control arm and its record is withheld for carrying a tester's load-order paths, and it is re-run with a control as eval 25.
- **Q6 — filing an issue for the OAR-introspection gap.** Not filed; filing is yours. The gap is stated in the body, once, beside what houseCARL is for here.

## Verifier overturns, applied

The eight rows `REWRITE-verify.md` overturned are corrected as the verifier says: the `DECISIONS.md` line citations are not used; the body-features host claim (Claude Code, not Codex) is not written into the skill; `compatibility` is not claimed to close D2.35 (it closes nothing here — ruling 7 is unconditional and that is the whole reason for it); the front-loaded expansion is not treated as a D3.8 mitigation; the "eighteen lines" figure is dropped (it is 83); the enum's authority is OAR's `Conditions.cpp`, not the audit's five-config grep, which was for argument names; and the group-2 mechanical check is "every shipped file is named **at least** once and every named path resolves in `dist/`", not "exactly once" — the design places twelve pointers at one file on purpose. Verified after the package build: the shipped tree is `SKILL.md` + `references/oar-config-reference.md`, the body names that file and nothing else, and nothing it names is missing from `dist/`.

## Two judgement calls worth your eye

- **D5.49** was inside the 37-of-37 regrade gate with nothing in the document that would regrade it (verifier row 66). Naming an enforcement route for a specific host would have broken D4.12, which conforms today because the prose names no host. So the body carries one host-neutral sentence instead — a subagent asked to do OAR work needs the skill loaded in its own context, named in the task that spawns it. Grade it as you see it; the alternative was recording D5.49 as a standing partial.
- **Body size.** The document budgets ~13,250 characters across eleven sections; the body lands at ~15,400 (~3,850 tokens), 16% over. The overage is content the document itself mandates — twelve section pointers, the whole migration step written out with its fallback, both worked examples, and the D5.49 sentence — and it is still well inside the only published ceiling ("< 5000 tokens"). I did not cut mandated content to hit an estimate written before the text existed, but the number is worth knowing.

Build and checks, all from the worktree, after the rename: `scripts/build-plugin.ps1` green, leak-check clean, 11 skills packaged with `dist/housecarl/skills/open-animation-replacer/` present and no `oar-authoring` folder left; `dotnet build housecarl.sln -c Release` 0 errors; `dotnet test --filter "tier!=bridge"` 1858 passed, 0 failed; generator `ci-all` 128/128, with `codex-umbrella-coverage-guard` 12/12 and `plugin-validate-guard` 20/20 both green on the new name. The pointer scan ruling 2 calls for is not on this branch — it belongs to the wave 0 packager PR, which has not merged — so the dangling-pointer check above was done by hand against `dist/`.

## Regrade

Five independent regradings of this branch, one per angle, each derived from disk with nothing
carried forward (`dev/skill-validation/eval-2026-09/oar-authoring/regrade/D1.md` … `D5.md`),
assembled in `dev/skill-validation/eval-2026-09/oar-authoring/REGRADE.md`.

| Angle | Conforms | Partial | Fails | n/a | Rows |
|---|---|---|---|---|---|
| D1 — discovery and packaging | 36 | 0 | 1 | 5 | 42 |
| D2 — frontmatter | 39 | 0 | 0 | 13 | 52 |
| D3 — loading decision and invocation control | 27 | 1 | 1 | 4 | 33 |
| D4 — body, supporting files, instruction style | 40 | 2 | 1 | 21 | 64 |
| D5 — validation and standing context | 34 | 10 | 0 | 6 | 50 |
| **Total** | **176** | **13** | **3** | **49** | **241** |

That table is the regrade as it stood **before** the rename. **D3.8 is closed by this PR**: the row failed on the bare name reading as the English word "oar" with descriptions dropped, and `REWRITE.md:350` regrades it if the Q2 rename is taken. It is taken, so D3.8 now conforms and D3 has zero fails.

The two remaining fails are both unmeetable-as-written rather than defects in the skill: D1.28 (release tagging, packager-only, excluded from the gate by ruling 8) and D4.42 (bare MCP tool names, written that way on ruling 1, which requires the name the criterion forbids). Neither blocks a threshold. Factual-or-dead-call fails: **0**.

**Verdict: PASS.** D3.8 was the only threshold missed, and the rename clears it.

## Regrade

Re-run on 2026-09-08 against the branch as it stands, under the skill's new name. Five fresh
gradings, each derived from disk with nothing carried forward:
`dev/skill-validation/eval-2026-09/open-animation-replacer/regrade/D1.md`, `D2.md`, `D3.md` and
`D5.md`, plus `dev/skill-validation/eval-2026-09/oar-authoring/regrade/D4.md` (its grader wrote D4
under the old eval-folder name and says so in its header; it grades the same worktree copy).
Assembled in `dev/skill-validation/eval-2026-09/open-animation-replacer/REGRADE.md`. This supersedes
the table above, which was measured before the rename.

| Angle | Conforms | Partial | Fails | n/a | Rows |
|---|---|---|---|---|---|
| D1 — discovery and packaging | 36 | 0 | 1 | 5 | 42 |
| D2 — frontmatter | 39 | 0 | 0 | 13 | 52 |
| D3 — loading decision and invocation control | 28 | 1 | 0 | 4 | 33 |
| D4 — body, supporting files, instruction style | 40 | 2 | 1 | 21 | 64 |
| D5 — validation and standing context | 38 | 8 | 0 | 6 | 52 |
| **Total** | **181** | **11** | **2** | **49** | **243** |

D5 reads 52 rows because its grader counted the two carried rows D5.51 and D5.52 inside its total;
the other four exclude the rows their graders added beyond the checklists.

D3.8 conforms on this tree — `SKILL.md:2` is `name: open-animation-replacer`, the condition
`REWRITE.md:350` set — so D3 has zero fails. The two remaining fails are both unmeetable as written
rather than defects in the skill: **D1.28** (release tagging; tags are `vX.Y.Z`, not
`housecarl--v{version}`; packager-only, excluded from the gate by ruling 8) and **D4.42** (both MCP
tools written bare, which ruling 1 requires and the criterion forbids; classed `other`, both names on
`surface/tools.md` with every parameter published). Factual-or-dead-call fails: **0** on every angle.

**Verdict: PASS.** Zero fails on D2 and D3; zero factual-or-dead-call fails on D4 and D5; D1 clean
but for the packager-only D1.28 that ruling 8 excludes.


.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+3. **Note required addons.** A condition name absent from the built-in roster comes from an addon DLL
+   (Math / RaySense / IED / Detection / Dialogue). Confirm it is installed under `…/SKSE/Plugins/`,
+   or the line is a dead no-op. Which conditions come from which addon DLL:
+   `references/oar-config-reference.md` §7.
+4. **Resolve the forms a condition names.** A perk, keyword, race, faction or magic effect needs
+   `{ "pluginName": …, "formID": … }` or `{ "editorID": … }`, where `formID` is the record's
+   **local** id in its defining plugin. Read it with `housecarl_records` and
+   `project={"form":"identity"}`: `formids=["XXXXXX:Plugin.esp"]` when you have the FormID (the
+   runtime spelling a console, Papyrus or crash log prints is accepted too, and the response names
+   the plugin it resolved to); `types=["PERK"]` (or `KYWD`, `RACE`, `FACT`, `MGEF`) with
```
Avick3110, line 80: **Dead call: `form='identity'` on a scan is refused outright.**

Step 4's second branch â€” `types=["PERK"]` plus `where=["editorid startswith REQ_"]`, no `formids=` â€” is a scan, and `housecarl_records` refuses the identity form there. `src/housecarl-mcp/RecordsTools.cs:1170`:

```csharp
if (form == "identity")
    return Wire.Refuse(json, "error: the identity form labels a formids= list; a scan's summary rows already carry each match's identity â€” use form='summary' (the default).");
```

The parameter doc agrees: identity is "the labeling form; **needs formids=**". So the EditorID branch â€” the one an author takes when all they have is `REQ_SomePerk` â€” is a guaranteed failed call. That is the same defect class this rewrite set out to remove from the old `housecarl_read_record` step, reintroduced on the replacement.

Fix: keep `project={"form":"identity"}` on the `formids=` branch only; the scan branch wants the default summary form, whose rows already carry identity.

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+| The global `OpenAnimationReplacer.ini` | §11 |
+| FormID form and the embedded-null gotcha | §12 |
+
+Every step below names the section it needs, so one section can be read on its own.
+
+## First step — orient before you touch a config
+
+1. **Locate the submod and see who wins the file.** Run `housecarl_asset_status` with
+   `under=["meshes/actors/character/animations/OpenAnimationReplacer/**/config.json"]`, and the same
+   selector under `DynamicAnimationReplacer/**` for legacy folders. It resolves every config the VFS
```
Avick3110: **The legacy-DAR half of the locate sweep matches nothing.**

The step says to run "the same selector under `DynamicAnimationReplacer/**`" â€” but the selector's leaf is `config.json`, and a legacy DAR submod has no `config.json`. It carries `_conditions.txt`; `references/oar-config-reference.md` Â§8 is explicit ("Missing `_conditions.txt` â‡’ the folder is skipped with a warning"), and the Form B folders carry neither file.

So on a load order full of DAR submods this returns zero rows, silently. That is load-bearing: the mental model (lines 84-85) and Verification step 2 both insist legacy DAR folders compete in the same global priority space, and this sweep is the only input to the winner answer. Follow the step as written and you conclude there are no legacy competitors, then hand back the wrong winner whenever a DAR folder outranks the OAR submod.

The DAR selector needs its own leaf, e.g. `.../DynamicAnimationReplacer/**/_conditions.txt`, and a word about the actor-base folder form that has no marker file at all.

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+| Priority and winner resolution | §10 |
+| The global `OpenAnimationReplacer.ini` | §11 |
+| FormID form and the embedded-null gotcha | §12 |
+
+Every step below names the section it needs, so one section can be read on its own.
+
+## First step — orient before you touch a config
+
+1. **Locate the submod and see who wins the file.** Run `housecarl_asset_status` with
+   `under=["meshes/actors/character/animations/OpenAnimationReplacer/**/config.json"]`, and the same
```
Avick3110: **The sweep pins `<project>` to `character`, so every creature project is invisible.**

`references/oar-config-reference.md` Â§1 â€” the section this very step points at two lines down â€” says the layout is `meshes/actors/<project>/animations/...` and that `<project>` is "almost always `character` for humanoids. Creatures use their own project folder (`canine`, `draugr`, `dragon`, â€¦)". The deleted skill globbed `<project>`; this selector hardcodes it.

Verification step 2 asks for "every competing `config.json`" and has only this sweep to work from, so a draugr or dragon moveset question gets an empty competition set with no warning that the scan was narrowed. `under=["meshes/actors/*/animations/OpenAnimationReplacer/**/config.json"]` covers it â€” `*` matches within one path segment, per the `under=` parameter doc.

plugin/CHANGELOG.md
```diff
@@ -2709,6 +2709,24 @@ offset, paging meant slicing by `editorid_contains`. Now:
   each candidate, which for an NPC defined in `Skyrim.esm` is not the folder the FaceGen path points at. The
   four Race cases and that provider test move to `references/race-and-provider-cases.md`, the body is about a
   third shorter, and the trigger set moves to the published `evals/evals.json` shape with a without-skill arm.
+- **`oar-authoring` names tools that exist, finds the config the game actually loads, and states what it does not
```
Avick3110: **Truncated leftover bullet.**

This is a half-written duplicate of the entry two bullets below. The sentence stops mid-clause at "states what it does not", the `**` it opens is never closed â€” so every following line in the section renders bold â€” and it still names `oar-authoring`, the name this PR removes everywhere else. Delete the line.

.claude/skills/papyrus-reference/evals/evals.json
```diff
@@ -130,7 +130,7 @@
     {"id": 15, "query": "wire a TIF result fragment onto this dialogue line so it advances the quest stage", "should_trigger": false, "expected_route": "dialogue-authoring"},
     {"id": 16, "query": "how do i install SKSE and PapyrusUtil in MO2?", "should_trigger": false, "expected_route": "NONE / basics — names corpus members but is an install question"},
     {"id": 17, "query": "decompile campfire's pex files and tell me how heavy its update loops are", "should_trigger": false, "expected_route": "NONE — the judgment asked is heaviness, and the skill that owned it was deleted; hard near-miss"},
-    {"id": 18, "query": "gate this dodge animation behind a perk check so only the player gets it", "should_trigger": false, "expected_route": "oar-authoring — animation condition, not a script"}
+    {"id": 18, "query": "gate this dodge animation behind a perk check so only the player gets it", "should_trigger": false, "expected_route": "open-animation-replacer — animation condition, not a script"}
```
Avick3110, line 133: The same stale route survives one file over. `.claude/skills/dialogue-authoring/evals/evals.json:235`:

```json
"expected_output": "dialogue-authoring is NOT loaded; the query routes to housecarl:oar-authoring."
```

That is live case T18, not a historical run record, so it now grades a pick against a skill the install no longer offers â€” the exact hazard `spid-authoring`'s own method note warns about ("a should-not query whose only honest destination is one of those must be scored against NONE, not against a skill the install does not offer").

The PR body says the rename covered "the sibling route in `papyrus-reference`'s eval set"; dialogue-authoring's needs the identical edit.

scripts/build-plugin.ps1
```diff
@@ -57,7 +57,7 @@ $ReleaseDir   = Join-Path $RepoRoot 'release'          # output dir for the ship
 $PluginManifest = Join-Path $PluginSrc '.claude-plugin\plugin.json'   # single source of truth for the version
 
 # the 11 shipped skills (the modlist-authoring cluster was removed; tool-surface skills wait)
-$Skills = @('mutagen-reference','papyrus-reference','skypatcher-authoring','spid-authoring','kid-authoring','facegen-diagnostics','dialogue-authoring','oar-authoring','skse-plugin-authoring','bulk-record-jobs','npc-appearance-copy')
+$Skills = @('mutagen-reference','papyrus-reference','skypatcher-authoring','spid-authoring','kid-authoring','facegen-diagnostics','dialogue-authoring','open-animation-replacer','skse-plugin-authoring','bulk-record-jobs','npc-appearance-copy')
```
Avick3110, line 60: With this line changed, `.claude/skills/spid-authoring/evals/evals.json:5` becomes false. It describes its judge menu as "the 11 skills this tree actually ships (the `$Skills` list in `scripts/build-plugin.ps1`)" and then enumerates them, including `oar-authoring`.

It reads as a present-tense description of the shipped menu rather than a dated run record, and the note's own argument is that a query must not be scored against a skill the install does not offer. Worth folding into the same rename sweep.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
```diff
@@ -335,9 +342,32 @@ NOT IsEquippedLeftType(4)
...
+|---|---|---|
+| `IsEquippedRight("P.esp" \| 0xF)` | `IsEquipped` | `"Form": { "pluginName": "P.esp", "formID": "<local hex>" }`, `"Left hand": false` |
+| `IsEquippedLeft("P.esp" \| 0xF)` | `IsEquipped` | same `Form`, `"Left hand": true` |
+| `IsEquippedRightType(n)` | `IsEquippedType` | `"Type": { "value": n.0 }`, `"Left hand": false` |
+| `IsEquippedLeftType(n)` | `IsEquippedType` | `"Type": { "value": n.0 }`, `"Left hand": true` |
+| `IsEquippedRightHasKeyword(kwd)` | `IsEquippedHasKeyword` | `"Keyword": { "editorID": … }` or a form, `"Left hand": false` |
+| `IsEquippedLeftHasKeyword(kwd)` | `IsEquippedHasKeyword` | same `Keyword`, `"Left hand": true` |
+| `IsActorBase("P.esp" \| 0xF)` | `IsActorBase` | `"Actor base": { "pluginName": …, "formID": … }` |
+| `IsWorn` / `IsWornHasKeyword` | `IsWorn` / `IsWornHasKeyword` | one Form / Keyword component; no hand flag |
+| `HasPerk` / `HasSpell` / `HasMagicEffect` / `IsInFaction` / `IsRace` / `IsClass` | same name | one Form component named for the thing (`"Perk"`, `"Spell"`, …) — confirm the exact argument name per §6 |
```
Avick3110: **Circular pointer.** This row defers the argument name to Â§6 â€” but Â§6 publishes only the condition *name* roster, and says the opposite of what is promised here: "Each takes its own component params; open the in-game editor (or read an existing config) to see the exact argument names for a given condition."

So the one row honest enough to admit it does not know the key sends the reader to a section that also does not know it, and a reader who follows the pointer either finds nothing or assumes the parenthesised guess (`"Perk"`, `"Spell"`) is confirmed. A wrong key name is the silent no-op this whole skill exists to prevent, and this table is the new artifact the rewrite added for by-hand conversion.

Point the row at the in-game editor / an existing config directly, the way Â§6 does.

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+      "Conditions": [
+        { "condition": "IsEquipped", "requiredVersion": "1.0.0.0",
+          "Form": { "pluginName": "Skyrim.esm", "formID": "2F2F4" }, "Left hand": false },
+        { "condition": "IsEquipped", "requiredVersion": "1.0.0.0",
+          "Form": { "pluginName": "Woodaxeweapons.esp", "formID": "5909" }, "Left hand": false } ] } ] }
+```
+
+The other reading — `AND` binding tighter than `OR` — puts the guard inside an `AND` with only the
+*first* equipped-form term, and wraps the whole in the `OR`: that fires the animation in combat for
+every weapon but the first. Record which you wrote in the submod's `description`.
```
Avick3110: The instruction is to record which `AND`/`OR` reading you implemented "in the submod's `description`" â€” but the config directly above (lines 222-231) carries `name` and `priority` and no `description` key at all.

The exemplar an author copies therefore violates the rule the same paragraph sets, and `evals/evals.json` eval 25 grades exactly this ("a run that emits one config without saying which binding it assumed FAILS"). Add `"description"` to the example JSON, carrying the sentence the rule asks for.

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+
+Legacy folder `…/DynamicAnimationReplacer/_CustomConditions/2000030002/` holds:
+
+```
+NOT IsInCombat() AND
+IsEquippedRight("Skyrim.esm" | 0x02F2F4) OR
+IsEquippedRight("Woodaxeweapons.esp" | 0x005909)
+```
+
+The folder name is the priority. Map each function to its OAR condition — the full table is
```
Avick3110, line 224: The example converts the conditions but never mentions the animation files, and the PR body sells it as converting "a real DAR folder end to end".

A `_CustomConditions/<n>/` folder holds `.hkx` files beside `_conditions.txt`, and a native submod only replaces anything when its `.hkx` mirror the original animation's path â€” this skill's own authoring step 1 (line 97) â€” or it sets `overrideAnimationsFolder`. Emit only the config shown here and you get a submod that parses, loads, wins its priority slot, and replaces nothing; the failure surfaces in game, not in Detected Problems.

One sentence saying to carry the `.hkx` across at their mirrored paths (or point `overrideAnimationsFolder` at the legacy folder) closes it.

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+  DAR folders included; the highest one whose conditions pass wins. If you cannot see every competing
+  submod, say so and name what you would need to see.
+- **"Why isn't it playing?"** Walk the list: is a higher-priority submod winning; do the conditions
+  actually pass in that situation (weapon hand, enum value, missing perk); is a required addon DLL
+  missing so the condition reads INVALID; does the `.hkx` path mirror the original; is a `user.json`
+  shadowing the `config.json` you are reading; is `disabled` set?
+
+## Common mistakes
+
+- **Write the submod's top-level array as lowercase `conditions` and every nested child array as
```
Avick3110, line 150: **This section is headed "Common mistakes" but every bullet states the correct action.**

The bolded leads are imperatives to follow, not errors to avoid: "Write the submod's top-level array as lowercase `conditions`...", "Take `IsEquippedType` values from the reference...", "Treat a submod with no `.hkx` as normal", and at line 148 "**Change the `priority` integer to change a winner.**"

An agent skimming bold leads under a heading that declares them mistakes reads that last one as "do not change the priority integer to change a winner" â€” inverting the one rule the skill puts at its centre. The deleted body phrased them as mistakes ("Editing load order to fix a winner. Pointless â€” OAR only reads `priority`"), which matched the heading.

Either restore the mistake phrasing, or rename the heading to what the bullets now are.

Avick3110: Refuted, with the reason.

The rewrite document mandates both halves of what this row objects to. Its body outline (`dev/skill-validation/eval-2026-09/oar-authoring/REWRITE.md`, §4, section 7) names the section **Common mistakes** and specifies "Seven traps, each **led by the correct move** with the mistake as the reason (D4.47)". D4.47 is a sources-disagree row — Anthropic prefers the positive, Codex writes prohibitions — and the document's §9 gate resolves it with "the correct-move-first bullet that keeps the mistake as the reason", explicitly so that "both shapes satisfy every stated style at once". Restoring the mistake phrasing would put D4.47 back to partial; renaming the heading would depart from the section name the document sets.

The one bullet the finding calls the sharp case, `:148`, carries its reason in the same sentence — "Editing MO2 or plugin order does nothing — OAR reads only `priority`" — so the correct action and the mistake it names are both on the line, which is the shape the ruling asks for.

The other nine findings on this PR are folded; the eight that were about the sweep, the `identity` form, the argument-name pointer, the eval routes and the changelog bullet are on the branch as of 00e8ad4c.

Avick3110:
## Fold pass

Nine of the ten findings folded, one refuted in a reply on its own thread.

**Folded** (db393006, 1eb05bfb, 6f559164, 00e8ad4c):

- **`form='identity'` on the EditorID branch.** Confirmed against `src/housecarl-mcp/RecordsTools.cs:1170` — the scan lane refuses the identity form outright. Step 4 now keeps `project={"form":"identity"}` on the `formids=` branch only and tells the scan branch to leave `project=` off and read the default summary rows, which already carry identity.
- **The DAR half of the locate sweep matched nothing.** The DAR selector's leaf is now `_conditions.txt`, per reference §8 Form A. The step also says Form B `<Plugin.esp>/<FormID>/` carries no marker file at all and names the `**/*.hkx` sweep for it.
- **`<project>` was pinned to `character`.** Both selectors now use `meshes/actors/*/animations/...`; `*` matches within one path segment per the `under=` parameter doc, so `canine`, `draugr` and `dragon` are swept too.
- **Circular pointer in the DAR→OAR table.** The `HasPerk` / `HasSpell` / … row deferred to §6, which itself defers to the in-game editor. It now points at the editor or an existing config directly and says the name is not published in this reference.
- **The DAR exemplar violated its own rule.** The config now carries a `description` saying which `AND`/`OR` reading it implemented and what the other would have meant — the key eval 25 grades and a valid submod key per reference §3.
- **The DAR example emitted no animations.** A closing paragraph says to carry the legacy folder's `.hkx` across at their mirrored paths, or point `overrideAnimationsFolder` at the legacy folder, and that a config alone parses, loads, wins its slot and replaces nothing.
- **Stale route in `dialogue-authoring`'s eval T18** and **the `$Skills` menu in `spid-authoring`'s method note** both now name `open-animation-replacer`.
- **The truncated changelog bullet** at `plugin/CHANGELOG.md:2712` is deleted.

**Refuted:** the "Common mistakes" heading-versus-bullets row — the rewrite document mandates both the heading and the correct-move-first phrasing. Reason on the thread.

Checks from the worktree after the fold: `scripts/build-plugin.ps1` green, leak-check clean, 11 skills; `dotnet build housecarl.sln -c Release` 0 errors; `dotnet test --filter "tier!=bridge"` 1890 passed, 0 failed; generator `ci-all` 128/128.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
```diff
@@ -335,9 +342,32 @@ NOT IsEquippedLeftType(4)
...
+### DAR function → OAR condition
+
+The mapping a by-hand conversion needs. Each row is this section's function roster crossed with the
+built-in roster (§6) and the value-component shapes (§4); the hand flag is §4's Bool component.
+
+| DAR function | OAR condition | Parameters |
+|---|---|---|
+| `IsEquippedRight("P.esp" \| 0xF)` | `IsEquipped` | `"Form": { "pluginName": "P.esp", "formID": "<local hex>" }`, `"Left hand": false` |
+| `IsEquippedLeft("P.esp" \| 0xF)` | `IsEquipped` | same `Form`, `"Left hand": true` |
+| `IsEquippedRightType(n)` | `IsEquippedType` | `"Type": { "value": n.0 }`, `"Left hand": false` |
```
Avick3110, line 354: **DAR's type number is mapped 1:1 onto OAR's `IsEquippedType`, with no note that the two enums may diverge.**

§5 of this same file says OAR's enum "deliberately differs" from the vanilla enum "at values 6/9/10/11", and splits battleaxe/warhammer by keyword precisely because the engine type `kTwoHandAxe` is a single value. This row asserts `n → n` across exactly that range without saying it was checked.

Concretely: converting `IsEquippedRightType(6)` emits `"Type": { "value": 6.0 }`, which under §5 is Battleaxe only — warhammer is 10. If DAR's 6 covers the whole `kTwoHandAxe` class (which is what §5's own "OAR splits them by keyword" implies about any engine-typed source), the converted submod silently stops applying to warhammers. The config parses, loads, wins its priority slot, and Detected Problems stays clean, so nothing flags the narrowing. 9/10/11 are worse: DAR has no such labels at all under the divergent reading.

Either verify DAR's enum against OAR's legacy parse in `Parsing.cpp`, or mark the numeric identity unverified at 6/9/10/11 the way §8 already marks the `AND`/`OR` binding — the same honesty rule, one row over.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
```diff
@@ -335,9 +342,32 @@ NOT IsEquippedLeftType(4)
...
+| `IsEquippedRightHasKeyword(kwd)` | `IsEquippedHasKeyword` | `"Keyword": { "editorID": … }` or a form, `"Left hand": false` |
+| `IsEquippedLeftHasKeyword(kwd)` | `IsEquippedHasKeyword` | same `Keyword`, `"Left hand": true` |
+| `IsActorBase("P.esp" \| 0xF)` | `IsActorBase` | `"Actor base": { "pluginName": …, "formID": … }` |
+| `IsWorn` / `IsWornHasKeyword` | `IsWorn` / `IsWornHasKeyword` | one Form / Keyword component; no hand flag |
+| `HasPerk` / `HasSpell` / `HasMagicEffect` / `IsInFaction` / `IsRace` / `IsClass` | same name | one Form component named for the thing — the exact argument name is not published in this reference; read it off the in-game editor or an existing config that uses the condition |
+| `IsInCombat()` / `IsChild()` / `IsInInterior()` / `IsPlayerTeammate()` | same name | no parameters |
+| Form B's `<Plugin.esp>/<FormID>/` folder pair | `IsActorBase` | auto-synthesized from the folder names; write it out explicitly when converting by hand |
+
+`NOT Fn(…)` becomes `"negated": true` on the converted condition. A DAR `0x0ABCDE` argument becomes
+the **local** hex string `"ABCDE"` in the `formID` — see §12.
```
Avick3110: **`see §12` tells the converter to strip a byte a DAR argument does not carry, and contradicts SKILL.md's own worked example.**

§12's rule for producing a local id is "strip the load-order byte". A DAR `_conditions.txt` argument is *already* local to the plugin named beside it — that is what the `"Plugin.esp" | 0xFormID` pair means — so applying §12's rule removes a byte of the real id.

On the example this file and `SKILL.md` share: `IsEquippedRight("Skyrim.esm" | 0x02F2F4)` (Woodcutter's Axe). §12's rule reads `02` as the load-order byte and gives `"formID": "F2F4"`; `SKILL.md:222` gives `"2F2F4"`. The §12 answer points OAR at `0x0000F2F4` — a different record that still *resolves*, so the condition is simply never true, the animation never plays, and Detected Problems reports nothing.

```suggestion
`NOT Fn(…)` becomes `"negated": true` on the converted condition. A DAR argument is ALREADY the local id
within the plugin named beside it — nothing to strip: `0x02F2F4` becomes `"2F2F4"`, dropping only the `0x`
and leading zeros. §12's load-order-byte rule is for a RUNTIME FormID, not for a DAR argument.
```

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,264 @@
...
+
+Loop steps 1-4 until Detected Problems is empty and the priority list shows your submod winning in
+the situation you care about. That pair is the stop condition.
+
+## Worked example — overriding a mod's conditions via `user.json`
+
+A mod's archery moveset should apply only after the player earns a specific perk, without editing the
+mod. The original `…/Bow Rapid Combo V3/Base/config.json` is `IsActorBase player` plus
+`IsEquippedType 7` at priority `9901000`. Put a `user.json` at the matching submod path in the
+dedicated overrides mod, holding the complete original config plus one condition:
```
Avick3110: **The example is introduced as "holding the complete original config" but carries only `priority` and `conditions`.**

Step 4 and reference §9 both state `user.json` is a full-document shadow that "must hold the complete config you want". This example then models the opposite shape while claiming to be the complete one.

A reader copying it over a submod whose `config.json` also sets `overrideAnimationsFolder` — a submod that ships no `.hkx` of its own, which this skill calls normal under "Common mistakes" — drops that key. The shadowed submod then wins its `9901000` slot and replaces nothing. Same for `replacementAnimDatas`: the mod's random variants disappear silently. This is the exact failure the DAR example warns about 40 lines later ("a config alone parses, loads and wins its priority slot while replacing nothing, and that failure shows up in game, not in Detected Problems").

Either show the example carrying the original's other keys, or change the lead-in so the elision is explicit — e.g. "…plus every other key the original carries (`name`, `description`, `overrideAnimationsFolder`, `replacementAnimDatas`, …), elided here".

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -0,0 +1,264 @@
...
+| Priority and winner resolution | §10 |
+| The global `OpenAnimationReplacer.ini` | §11 |
+| FormID form and the embedded-null gotcha | §12 |
+
+Every step below names the section it needs, so one section can be read on its own.
+
+## First step — orient before you touch a config
+
+1. **Locate the submod and see who wins the file.** Run `housecarl_asset_status` with
+   `under=["meshes/actors/*/animations/OpenAnimationReplacer/**/config.json",
```
Avick3110: **This selector anchors the sweep at `meshes/actors`, not at the OAR tree.**

`AssetGlob.LiteralPrefix` (`src/housecarl-core/AssetGlob.cs:24`) cuts a selector at the separator before its **first** wildcard. With `*` in the third segment, the literal prefix of both selectors is `meshesctors`, so `AssetGlob.Select` calls `EnumerateUnder("meshesctors")` — every loose file and every BSA entry under the whole actors tree — then regex-filters it and sorts the survivors.

`limit=` / `offset=` do not bound that: they window the *render*, after `Select()` has already enumerated and sorted. On the order this skill's own eval 25 names (ARR 2.0, 3,801 plugins) that is every facegen mesh, every body and armor mesh and every actor entry in every active archive, to find a few hundred `config.json` files — and the verification loop pays it again on each iteration.

A wildcard-free selector takes everything beneath it, so the cheap spelling is literal:

```
under=["meshes/actors/character/animations/OpenAnimationReplacer",
       "meshes/actors/character/animations/DynamicAnimationReplacer"]
```

naming the extra project folders (`canine`, `draugr`, …) explicitly when a creature is in play.

.claude/skills/open-animation-replacer/evals/evals.json
```diff
@@ -0,0 +1,262 @@
...
+    "run_from": "dist/housecarl/skills/open-animation-replacer/ - the PACKAGED tree, not .claude/skills/. The packager strips evals/, so the suite is driven from outside the tree it tests, which is also the check that no body pointer dangles in dist/.",
+    "arms": {
+      "trigger": "Which ONE skill a fresh-context agent loads for a query, measured with-skill and without-skill.",
+      "output_quality": "The audit's own DAR-to-OAR task on the live order, run as a pair: control (no housecarl:* skill in the catalogue) against with-skill. Eval id 25."
+    },
+    "method": {
+      "cases": "24 trigger cases - 12 should-trigger, 12 should-not - plus 1 output-quality case. The shipped 20 (10/10) survive with their queries unchanged; 4 are added for the description's changed clauses (2 positives naming the front-loaded terms, 2 near-miss negatives for the new exclusion clause).",
```
Avick3110: **`_meta` says 12 should-trigger cases; the array has 13.**

Eval 25 — the output-quality case, marked `"arm": "output_quality"` — also carries `"should_trigger": true`. Counting the array gives 13 should-trigger and 12 should-not, against this line's "12 should-trigger, 12 should-not — plus 1 output-quality case".

`should_trigger` is the field every other case uses and the only partition key 24 of the 25 carry, so a runner that splits on it puts the full DAR-conversion task into the trigger arm: it gets scored as a which-skill-do-you-load pick it was never written to answer, and recall is computed over 13 rather than the 12 the `recall_threshold: 0.8` gate was set against.

Drop `should_trigger` from eval 25 (leaving `arm` to place it), or correct this count to 13/12.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
Avick3110: Folded, on the second of the two routes you named — marked unverified, not verified.

Verifying it needs OAR's legacy parse in `Parsing.cpp`, and no copy of `ersh1/OpenAnimationReplacer` exists anywhere this branch can reach: `dev/` carries the audit inputs and the eval sets, not the source the reference was generated from. Fetching it is the same fetch decision Q1 already leaves for you, so I did not take it here either.

`references/oar-config-reference.md:364`, immediately under the DAR→OAR table, now says the `n → n` identity is UNVERIFIED at 6, 9, 10 and 11, names why (§5's four divergences — battleaxe 6 / warhammer 10 split by `WeapTypeWarhammer` out of the single engine type `kTwoHandAxe`, crossbow 9, shield 11), says the two agree outside that range, gives your concrete case (`IsEquippedRightType(6)` emitting `"Type": { "value": 6.0 }` silently stops applying to warhammers if DAR's 6 covered the whole class, and 9/10/11 may name nothing the DAR file meant), names `Parsing.cpp` as the source that would settle it, and puts it under the same rule §8 gives the `AND`/`OR` binding: convert under one reading, say which in the submod's `description`, test in game. It closes on the same failure sentence — parses, loads, wins its slot, never passes, Detected Problems clean.

Commit c1e55703.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
Avick3110: Folded, on your suggestion.

Confirmed the contradiction: §12 (`references/oar-config-reference.md:480`) reads "the record's **local** id within that plugin (strip the load-order byte)", the row pointed at it, and `SKILL.md:229-230` gives `0x02F2F4` → `"2F2F4"` — so following the pointer on the file's own shared example produces `"F2F4"`, a different record that resolves and is simply never true.

`references/oar-config-reference.md:376` now reads:

> `NOT Fn(…)` becomes `"negated": true` on the converted condition. A DAR argument is ALREADY the local id within the plugin named beside it — that is what the `"Plugin.esp" | 0xFormID` pair means, so there is nothing to strip: `0x02F2F4` becomes `"2F2F4"`, dropping only the `0x` and the leading zeros. §12's load-order-byte rule is for a RUNTIME FormID, not for a DAR argument; apply it here and you drop a real byte of the id, and the condition points at a record that still resolves and is simply never true.

§12 itself is unchanged — its rule is right for what it is about, it was the pointer that was wrong.

Commit dd4c9489.

.claude/skills/open-animation-replacer/SKILL.md
Avick3110: Folded, and the example is now complete rather than caveated.

The original is published in full one file over — `references/oar-config-reference.md` §9 shows `Bow Rapid Combo V3 / Base`'s `config.json` as `{ "name": "Base", "priority": 9901000, "keepRandomResultsOnLoop": true, "conditions": [...] }` — so there was no reason to elide anything. `SKILL.md:198-200` now opens the JSON with `"name": "Base"`, `"priority": 9901000` and `"keepRandomResultsOnLoop": true`, which is every key the original carries.

The lead-in (`SKILL.md:190-196`) states the general rule the example can only show for this one submod: hold **every key the original carries** plus the new condition, read the original first, and carry across whatever it has — `overrideAnimationsFolder` and `replacementAnimDatas` named — "or the shadowed submod wins its slot and replaces nothing", which is the same failure sentence the DAR example closes on.

Commit d9a7d1e1.

.claude/skills/open-animation-replacer/SKILL.md
Avick3110: Folded, with one deviation from the spelling you gave, stated below.

Confirmed the cost. `AssetGlob.LiteralPrefix` (`src/housecarl-core/AssetGlob.cs:24-30`) cuts at the separator before the first wildcard, so `meshes/actors/*/animations/...` prefixed to `meshes\actors` and `Select` called `EnumerateUnder` over the whole actors tree — facegen included — before anything rendered. `limit=`/`offset=` are documented on `under=` as windowing the selection's render, not the enumeration (`dev/skill-validation/eval-2026-09/surface/tools.md`, `housecarl_asset_status` → `limit`: "Max paths to resolve and render **from the selection**"), so they never bounded it. That `*` was mine — it came in folding the earlier finding that the sweep was pinned to `character` — and it bought creature coverage at the price of the whole tree.

`SKILL.md:52-61` now anchors both selectors at the project folder:

```
under=["meshes/actors/character/animations/OpenAnimationReplacer/**/config.json",
       "meshes/actors/character/animations/DynamicAnimationReplacer/**/_conditions.txt"]
```

and the step says to name a creature project explicitly — `canine`, `draugr`, `dragon` in place of `character` — and **not** to write `*` in the project segment, with your reason written out: the selector is enumerated from the literal directory in front of its first wildcard, so `meshes/actors/*/…` walks every facegen, body and armor mesh before a row renders, and `limit=`/`offset=` window the render, not that walk. That is the whole of the cost finding: the literal prefix is now `meshes\actors\character\animations\OpenAnimationReplacer`, so `EnumerateUnder` never leaves the animations tree.

The deviation: I kept the `**/config.json` and `**/_conditions.txt` leaves rather than going fully wildcard-free. With the prefix already at the animations tree the enumeration is identical either way, so nothing of your finding is left unpaid; what the leaves buy is the render. A bare directory selector takes everything beneath it (`AssetGlob.cs:7-8`), which is every `.hkx` in every submod — thousands of rows to page through — and it collapses the Form A / Form B distinction the step depends on: `_conditions.txt` is what makes a legacy row a Form A submod, and the separate `**/*.hkx` sweep two sentences down is what finds the Form B folders that carry no marker file at all. Say the word if you want the wildcard-free spelling anyway and I will take the leaves off.

Commit 12e32e76.

.claude/skills/open-animation-replacer/evals/evals.json
Avick3110: Folded, on the first of the two options — `should_trigger` dropped from eval 25.

Counted it before and after: 13 true / 12 false before, 12 / 12 now, with eval 25 the only case carrying no `should_trigger`. `arm: "output_quality"` (`evals/evals.json:255`) is left to place it, which is what `_meta.arms` already says the field is for.

`_meta.method.cases` (`evals/evals.json:12`) gains the clause that makes the partition explicit rather than implied: the 24 trigger cases are the ones carrying `should_trigger`; eval 25 carries `arm` and no `should_trigger`, so a runner splitting on that field never scores the DAR-conversion task as a which-skill pick, and recall is computed over the 12 the `recall_threshold: 0.8` gate was set against.

Commit 5d8f7595.

Avick3110:
## Fold pass — Aaron's review

All five findings folded; nothing refuted. Each has a reply on its own thread with the citation and the current line.

- **The DAR conversion stripped a byte off a form id** (dd4c9489). The `see §12` pointer sent the converter at §12's runtime rule; on this file's own shared example that turns `0x02F2F4` into `"F2F4"` and points the condition at a record that resolves and is never true. `references/oar-config-reference.md:376` now says a DAR argument is already local to the plugin named beside it, gives `0x02F2F4` → `"2F2F4"`, and says §12's rule is for a runtime FormID. §12 itself is unchanged.
- **The DAR type number was mapped 1:1 with no divergence note** (c1e55703). No copy of OAR's source is reachable from this branch, and fetching it is the Q1 decision still open, so this went the second route you named: `references/oar-config-reference.md:364` marks the `n → n` identity UNVERIFIED at 6, 9, 10 and 11, names §5's four divergences and `Parsing.cpp` as what would settle it, and puts it under §8's convert-under-one-reading-and-say-which rule.
- **The locate sweep anchored at `meshes/actors`** (12e32e76). Confirmed against `src/housecarl-core/AssetGlob.cs:24-30` and the `limit=` doc. Both selectors now anchor at the project folder, and the step says to name a creature project explicitly and not to write `*` in that segment, with the reason. One deviation, on its thread: the `**/config.json` and `**/_conditions.txt` leaves stay, because the enumeration is already bounded and the leaves are what keep Form A distinct from the marker-less Form B.
- **The `user.json` example was not the complete config it claimed to be** (d9a7d1e1). Reference §9 publishes the original in full, so the example now carries `name` and `keepRandomResultsOnLoop` too, and the lead-in states the general rule with `overrideAnimationsFolder` and `replacementAnimDatas` named.
- **Eval 25 sat in the trigger arm** (5d8f7595). `should_trigger` dropped from it; the counts are 12/12 with `arm` placing eval 25, and `_meta.method.cases` says so.

**Left for Aaron:** nothing new. The two calls the PR body already carries stand — Q1's fetch authorisation (the type-number row above is a second thing that only OAR's source can settle), and Q4's `compatibility` string as the batch wording.

Checks from the worktree after the fold: `scripts/build-plugin.ps1` green, leak-check clean (12 skills, 226 markdown files, both trees), 11 skills packaged with `dist/housecarl/skills/open-animation-replacer/` present; `dotnet build housecarl.sln -c Release` 0 errors; `dotnet test --filter "tier!=bridge"` 1890 passed, 0 failed; generator `ci-all` 128/128.

Issue #411, opened 2026-08-22
[GAP] skse-plugin-authoring teaches consuming an address, never discovering one


Avick3110:
Migrated from the retired `dev/BACKLOG.md` (advisor sweep, Aaron-go 2026-08-22); re-verified at `main` @ `a783b2f`.

The skill's references cover the whole REL consumption zoo, but across all 10 reference files (4,290 lines) there is zero Ghidra/IDA/disassembly methodology — an author following the skill can only copy offsets out of someone else's plugin. The text admits the hole twice: `references/hooking.md:545` (per-site resume offsets "are per-site properties needing disassembly" — and stops) and `:470` ("production ships raw offsets" — none of the three exemplars use the verified/pattern surface).

**Candidate fill** (recorded 2026-08-22, from Aaron): a tenth reference on address / struct-layout / call-site discovery, pointing at `alandtse/BethesdaGhidraScripts` as the setup route (clang AST-dump of CommonLib headers applied to your own exe in headless Ghidra: exact-offset structs ~99.75% typed, vtable layouts, Address Library IDs as symbols; SE/AE/VR + FO4/Starfield/FNV). The repo is a bench tool, not shippable reference (~371 MB, six submodules, per-machine output) — only a methodology doc could ship.

**Two gates before chartering (binding, from the corpus's own status):** (1) the corpus's V0–V3 empirical ladder is still owed (`references/_CORPUS_STATUS.md:52-82`) — a new lane inherits the same `[BUILD-TEST]` gate; (2) it tracks a living upstream, so it lands under the staleness duty, not the by-construction guarantee.

**Routing:** deliberately NOT routed (gated, standalone whenever — 2.0 makes it neither cheaper nor dearer; nearest-term value, AddWaterGuard, needs none of the skill work). Unmilestoned.

PR #638, opened 2026-09-08
Rewrite the skse-plugin-authoring skill

Avick3110:
Wave 1 of the skill rewrite: `skse-plugin-authoring`, built from the approved rewrite document with its verifier's three overturned rows applied. The skill now writes both halves of a native Papyrus function it is inventing — the `.psc` declaration grammar sits beside the C++ registration, with three worked cases (a global native, a latent function, and a return type that compiles and then silently truncates), because `papyrus-reference` is a lookup corpus for functions that already exist and by construction holds no entry for one that does not. The nine reference pointers are written `references/<file>.md` instead of as bare filenames, so they open as written from an installed plugin; every shipped file is named exactly once, in the reference section, and the repeated pointers in the mistakes section are gone. The body drops the host claim about Claude Code doing the whole loop, writes out the never-report-an-unrun-build rule instead of citing a rule name the shipped plugin does not carry, undates the lineage claims, and states each fact once. `references/multi-runtime.md` gains an "SE+AE only" five-row entry point for a target with no VR in it, and `references/toolchain-setup.md` gains `vcpkg.json`'s `builtin-baseline`, whose absence errors where classic mode is off. `_CORPUS_STATUS.md` leaves the skill directory for the working corpus — the packager already deleted it from every shipped copy, so a pointer at it was barred and leaving it unreferenced in the source tree was the finding. `evals/eval_set.json` becomes `evals/evals.json`: one file, a without-skill arm on every trigger query, an output-quality arm carrying the audit's own task with its six falsifiable predictions and three mechanical assertions, runs taken from the packaged tree, and no absolute user-home path.

Migration: the document measured this skill as naming no dead 1.x tool and no retired parameter, so the table is a call-form migration. All six `housecarl_*` sites (`housecarl_compile_script` ×5, `housecarl_compact_plugin` ×1) stay bare and live per ruling 1; the sites moved where the document said. All 18 `papyrus-reference` sites across the shipped body and references are touched: four in the body are deleted with the boundary paragraph, one survives as the single sentence in the native-function section, and the rest are renamed `housecarl:papyrus-reference` and narrowed from "owns the `.psc` surface" to "looks up functions that already exist". A grep for a bare `papyrus-reference` now returns one hit: the Codex bare-folder-name example the host note requires.

No issue is named as closing by the rewrite document, so there is no `Closes` line. #411 (this skill teaches consuming an address, never discovering one) is untouched and stays open.

**Q1 — does `compatibility` take this skill's toolchain clause? Decided as follows.** Taken as the document proposes: the shared 76-character sentence byte-identical, plus a second sentence naming MSVC, CMake, vcpkg-or-xmake and a CommonLibSSE-NG checkout. 192 characters against the 500 cap. Aaron can cut the second sentence to hold ruling 7 to "same wording throughout" without touching anything else.

**Q2 — is `housecarl_place` a deploy path for the built DLL? Left for Aaron.** The surface does not say whether an `assets` member's `source=` accepts a build-output path outside the VFS, so the body does not name the tool. It now says deploying is plain file placement and that no houseCARL tool takes a build output, which is the honest form of the sentence rather than the old "there is no deploy tool, and none is needed".

**Q3 — how far does N5's scoping sentence stretch? Decided as follows.** As proposed: the Overview names `housecarl_compile_script` as the only tool in the build lane and `housecarl_skse` / `housecarl_load_order_status` as the read tools for job (c) and load-failure triage. Holding N5 to its literal wording would leave the body claiming a tool surface the audit's own run contradicted.

**Q4 — does job (c) stay in this skill? Decided as follows.** It stays, and the description change is aimed at it. Recording the verifier's correction: job (c) is one of the three job shapes, not one third of the trigger set — one of the ten should-trigger rows is job (c). Splitting it out remains available and would want its own decision, not this PR.

**Q5 — CMake or xmake as the stated default? Decided as follows, and this is the judgement call most worth overturning.** CMake + vcpkg is stated as the default at the point of choice, with xmake taken when the request asks for it or the project already builds with it. The reason is the audit's: most requests phrase themselves in CMake, and the toolchain reference already teaches the CMake file-set with confidence while marking that no public template demonstrates it. The alternative reading — xmake is the proven path, so it is the default and CMake is the marked one — is defensible and is a one-line change.

**Q6 — does the V0–V3 in-game validation pass ride with this sweep? Left for Aaron.** The honesty rule survives at three lines and says the pass is still owed. Nothing here schedules it.

## Judgement calls beyond the questions

- The body is 12.3 KB / ~2,890 tokens against the document's ~10.9 KB / ~2,710 target, down from 13.3 KB / ~3,300. Two sections run over budget: Step 0 carries the CMake-versus-xmake default and the `/Zc:preprocessor`-plus-PCH obligation, and the lineage section carries the non-NG declarative-loader idiom. Both are content the document routes into the body from elsewhere; the alternative was losing two of the six saves the audit credits. Still a little over half the advisory 5,000-token ceiling.
- The mistakes section holds seven traps as budgeted, but not the same seven: the two multi-runtime traps are merged into one bullet and the `SKSE::Init` log-truncation footgun takes the freed slot, since it is one of the six itemised saves and appears nowhere in the shipped body.
- The packager's leak scan runs clean and the packaged tree was checked by hand: nine reference files ship, all nine pointers resolve inside `dist/housecarl/skills/skse-plugin-authoring/`, and neither `evals/` nor the ledger is present. The pointer-scan extension ruling 2 asks for is the wave 0 packager PR's and is not in this base, so that check was manual here.

## Regrade

Five independent graders re-ran D1-D5 against this branch; the assembly is `dev/skill-validation/eval-2026-09/skse-plugin-authoring/REGRADE.md`.

| Angle | Rows | Conforms | Partial | Fails | n/a |
|---|---|---|---|---|---|
| D1 - discovery and packaging | 42 | 36 | 0 | 1 | 5 |
| D2 - frontmatter | 52 | 39 | 1 | 0 | 12 |
| D3 - loading and invocation | 35 | 30 | 1 | 0 | 4 |
| D4 - body, supporting files, instruction style | 64 | 42 | 3 | 1 | 18 |
| D5 - evals, safety, context economy | 52 | 31 | 15 | 0 | 6 |
| **Total** | **245** | **178** | **20** | **2** | **45** |

**PASS** against the rewrite document's regrade thresholds. D2 and D3 return zero fails. D4 and D5 return zero factual-or-dead-call fails. D1 is clean but for D1.28, which is packager-only (the repo tags `vX.Y.Z` rather than `housecarl--v{version}`, outside the skill's files) and excluded by ruling 8.

The two surviving fails are the two the thresholds carve out in advance. D1.28 as above. D4.42 is a naming-form defect - the three tool names are written bare, `housecarl_compile_script` / `housecarl_skse` / `housecarl_load_order_status`, all three live on the surface as written, so it is not a dead call - and ruling 1 settles the bare form as the house form.

Every row the threshold table named as a fail to fix is conforms: D1.33, D3.13, D4.13, D4.20, D4.50, D5.13, D5.51. Recorded alongside the pass: D5.19 and D5.21 came back partial rather than conforms - both arms and the without-skill control are specified in `evals/evals.json` but neither set has been run, so the eval exists as a specification and not yet as a result. That run is also what would move nine of D5's fifteen partials.


.claude/skills/skse-plugin-authoring/references/multi-runtime.md
```diff
@@ -31,6 +32,30 @@ surprising claim traces back to source.
...
+## SE+AE only — the five rows {#se-ae-only}
+
+Most of this file is about VR, which is the runtime that shifts vtables and changes base classes. If the
+target is **SE + AE and no VR**, read these five rows and stop; the rest is for when VR joins the set.
+
+1. **The derivation is in `Common.h`, and it wins.** Authors set only `ENABLE_SKYRIM_SE` /
+   `ENABLE_SKYRIM_AE` / `ENABLE_SKYRIM_VR`; every `EXCLUSIVE_*` macro is derived from those at preprocess
+   time. Read the derivation verbatim in [the preprocessor define set](#the-preprocessor-define-set) rather
+   than any summary table — where a table and `Common.h` disagree, `Common.h` wins.
+2. **FLAT is not SE and is not AE.** A dual SE+AE build defines `EXCLUSIVE_SKYRIM_FLAT` **alone**; `_SE` and
```
Avick3110: **Row 2 tells the SE+AE reader to gate on macros its own build never defines.**

The row correctly states that a dual SE+AE build (the `flatrim` preset) defines `EXCLUSIVE_SKYRIM_FLAT` **alone** — and then, in the same sentence, says to "gate with `EXCLUSIVE_SKYRIM_SE` / `EXCLUSIVE_SKYRIM_AE` wherever the two layouts differ". In a flatrim build neither macro exists, so `#if defined(EXCLUSIVE_SKYRIM_SE) ... #elif defined(EXCLUSIVE_SKYRIM_AE) ... #endif` takes no arm: the `BaseExtraList` handling this row names as the canonical case compiles away to nothing, on both runtimes, with no diagnostic.

The section header says "read these five rows and stop", so the reader never reaches *Compile-time vs runtime — the decision rule*, which carries the mechanism that actually works for SE-vs-AE inside one DLL: `REL::Module::IsAE()` and `REL::Relocate(seAndVr, ae)` (that section notes `Relocate` "emits a `Module::IsAE()` branch in SE+AE"). Row 4's `RELOCATION_ID` two-arg form covers address IDs only, not offsets or behaviour.

Suggest row 2 stop at "cannot discriminate SE from AE" and then hand off explicitly: compile-time `_SE`/`_AE` gating applies only if you ship two separate builds (presets `se` and `ae`); for one dual DLL, discriminate at runtime.

.claude/skills/skse-plugin-authoring/SKILL.md
```diff
@@ -1,165 +1,193 @@
...
-  it's so you read borrowed code correctly — never as a recommendation to switch.
-- **Crash-diagnostics handoff.** `load-failures.md` (symptom → `skse64.log` line → cause → fix) and
-  `plugin-dissection.md` (DLL-name → mod attribution) are the surfaces houseCARL's future crash-diagnostics
-  skill consumes in reverse. When a task is really "diagnose this user's crash," those two references are the
-  bridge.
-- **houseCARL tool touch-points are thin.** The C++ build runs on the native compiler, not a houseCARL tool.
-  The only tool in this skill's lane is `housecarl_compile_script` for the **`.psc` consumer side** of a
-  native function (job b). Deploying the built DLL is plain file placement — copy it to a mod's
-  `SKSE/Plugins/<name>.dll`; there is no deploy tool, and none is needed.
+- **Gate with `EXCLUSIVE_SKYRIM_SE` / `_AE` where SE and AE differ, and read every runtime-varying
```
Avick3110: Same contradiction as `references/multi-runtime.md` row 2, compressed into a single bullet: it opens "Gate with `EXCLUSIVE_SKYRIM_SE` / `_AE` where SE and AE differ" and then states "A dual SE+AE build defines `EXCLUSIVE_SKYRIM_FLAT` alone". Both halves cannot be acted on at once.

The mistakes list is the part of the skill that is always in context, so this is the version an agent acts on. Writing `#if defined(EXCLUSIVE_SKYRIM_SE)` into a flatrim build leaves every arm unselected and the SE/AE-divergent code silently absent — exactly the "compiles clean" failure the bullet exists to prevent. Name the runtime discriminator (`REL::Module::IsAE()` / `REL::Relocate(seAndVr, ae)`) as the answer for one dual DLL, and reserve `_SE`/`_AE` for separate per-runtime builds.

.claude/skills/skse-plugin-authoring/SKILL.md
```diff
@@ -1,165 +1,193 @@
 ---
 name: skse-plugin-authoring
 description: >-
-  Author, build, or audit a native SKSE plugin DLL in C++ against CommonLibSSE-NG — "plugin" here means a compiled DLL, not an .esp or .psc. Use for writing or building an SKSE plugin or .dll, native Papyrus functions, BSTEventSinks, game hooks (trampoline, Address Library), or SE + AE + VR targeting. Load before any SKSE C++ — multi-runtime and lifecycle rules compile clean yet fail at load.
+  Author, build, or audit a native SKSE plugin DLL against CommonLibSSE-NG (alandtse ng) — "plugin"
+  here means a compiled DLL, not an .esp or .psc — and load this before any SKSE C++ is written. Use
+  for an SKSE plugin or .dll, declaring and registering a native Papyrus function, BSTEventSinks,
+  game hooks (trampoline, Address Library), SE + AE + VR targeting, or reading someone else's SKSE
+  DLL source. Multi-runtime and lifecycle rules compile clean yet fail at load.
```
Avick3110, line 8: The description body is **472 characters**. The standing target from issue #294 — cited in this file's own `evals/evals.json` history ("all 13 descriptions compressed to the amended 200-400 char target") and in every sibling skill's eval set — is 200-400. Every other shipped skill on this branch measures 369-431 including the `description:` prefix, and the version this PR replaces was 398.

Nothing in CI measures it: `PluginValidateProbe` only asserts `name` and `description` are non-empty and parse, so the overrun ships silently. The regrade table reports D2 (frontmatter) with zero fails, which suggests this row was not checked. Either trim to target or record the exception deliberately.

.claude/skills/skse-plugin-authoring/SKILL.md
```diff
@@ -1,165 +1,193 @@
...
+compatibility: >-
```
Avick3110, line 9: `compatibility` is a frontmatter key no other skill on this branch carries and nothing in the repo reads — grepping `src/`, `scripts/` and `standards/` finds no consumer, and `PluginValidateProbe` validates only `name` and `description`.

The one gate that could reject an unrecognized key is CI's `claude plugin validate ./dist/housecarl --strict` (`.github/workflows/ci.yml:136`), which is not part of the local build/test loop in CLAUDE.md — so if strict mode rejects unknown SKILL.md frontmatter fields, that surfaces as a red CI after the review, not during it.

Also worth reconciling with the PR body, which describes this as "the shared 76-character sentence byte-identical": no other skill carries such a sentence, so there is nothing shared here yet. Confirm against a strict-validate run before merge, or put the text in the body where it is certainly read.

.claude/skills/skse-plugin-authoring/SKILL.md
```diff
@@ -1,165 +1,193 @@
...
+
+The loop is build → deploy → verify. Deploying is plain file placement — copy the built DLL to a
+mod's `SKSE/Plugins/<name>.dll`; no houseCARL tool takes a build output. Then launch and read
+`skse64.log` for the plugin's own line, which answers "does it load?" in seconds;
+`housecarl_load_order_status` names the resolved SKSE crash-log and Papyrus script-log folders, so
+triage never starts with a filesystem hunt.
+
+## A native Papyrus function — both halves are written here
+
+The `.psc` declaration for a native you are inventing is written here, beside its C++ registration:
```
Avick3110, line 74: This skill now claims the `.psc` declaration of a new native, but the other side of the boundary is untouched in this PR.

`papyrus-reference`'s description still reads "Use for any `.psc` read or edit, any added or changed function call ... Unbundled functions get a warning, never an invented signature." A native you are inventing is by definition unbundled, so a request phrased script-side ("what's the `.psc` line for the native I'm adding?") can still route to `papyrus-reference`, which will then warn rather than write the declaration — the failure this rewrite exists to fix.

On Codex there is no `housecarl:` namespace and the umbrella router is authoritative; `plugin/codex/housecarl/SKILL.md:71` still lists `papyrus-reference` as owning "Papyrus / SKSE function signatures". Unless the carve-out is recorded on one of those two surfaces, the old handoff can still fire in the old direction.

.claude/skills/skse-plugin-authoring/evals/evals.json
```diff
@@ -0,0 +1,108 @@
...
+      "runs_per_query": 3,
+      "trigger_rate_threshold": 0.5,
+      "arms": ["with_skill", "without_skill"],
+      "without_skill_arm": "The same query, the same menu with this skill's entry removed, in a fresh context. A positive row that is answered correctly without the skill is evidence the skill is not earning that row; a negative row that goes wrong only in the with-skill arm is a false-trigger cost.",
+      "split": {"train": 0.6, "held_out": 0.4},
+      "held_out_ids": [6, 9, 10, 13, 17, 18, 19, 20],
+      "models": ["haiku", "sonnet", "opus"],
+      "deployment_envelope": "If a model is skipped, the record states which models the result holds for rather than generalizing.",
+      "contexts": ["isolation", "coexistence"],
+      "coexistence_note": "The coexistence set must include a query that exercises what used to be the papyrus-reference handoff — a new native's .psc declaration — because that is this skill's real coexistence risk now that it owns both halves.",
```
Avick3110: `coexistence_note` says the coexistence set "must include a query that exercises what used to be the `papyrus-reference` handoff — a new native's `.psc` declaration". No row in `evals` is that query. The closest, id 2, asks about the C++ side ("how do I register it?"), which is the half the skill already owned; nothing probes the half this PR newly claims.

The set is also `"status": "not yet run"`, so nothing else surfaces the gap. The file states a requirement it does not satisfy — add the row (something like "I'm adding a native to my SKSE dll, what does the `.psc` declaration for it look like?") rather than leaving the note as the only record of the risk.

.claude/skills/skse-plugin-authoring/evals/evals.json
```diff
@@ -0,0 +1,108 @@
...
+      {"id": 3, "should_trigger": true, "query": "my SKSE dll builds and runs fine on SE but CTDs on AE only. pretty sure it's a struct member offset thing. how do people handle this"},
+      {"id": 4, "should_trigger": true, "query": "can you scaffold a CommonLibSSE-NG project for me — vcpkg + cmake, targeting 1.6.1170, just a hello-world that logs on load"},
+      {"id": 5, "should_trigger": true, "query": "I want to hook the game so my code runs whenever an actor takes a hit. trampoline or vtable hook? and how do I not break the original call"},
+      {"id": 6, "should_trigger": true, "held_out": true, "query": "here's the github for an open-source combat overhaul's SKSE dll. can you read the source and tell me what it actually patches in the engine", "note": "The only miss of the 2026-07-28 revalidation, and the row the description's 'reading someone else's SKSE DLL source' clause was added for. Held out — it is not used to tune the description text."},
```
Avick3110: Row 6 is marked "Held out — it is not used to tune the description text", but the same file contradicts that: `revalidation_2026-07-28_294.adjudicator_notes` says the fix for this row ("surface read/reverse-engineer coverage in the trigger sentence") "is applied in the 2026-09 rewrite", and the new description now reads "... or reading someone else's SKSE DLL source".

The description was fitted to this row before it was held out, so the row cannot measure generalization. With only three positives in the held-out split (6, 9, 10), one pre-fitted row is a third of the held-out recall gate — a 3/3 there is not evidence of anything. Either move row 6 to train and hold out a positive that did not drive the edit, or drop the "not used to tune" claim.

.claude/skills/skse-plugin-authoring/evals/evals.json
```diff
@@ -0,0 +1,108 @@
...
+  "trigger_set": {
+    "purpose": "Does the description route the right requests here and keep the wrong ones out?",
+    "method": {
+      "shape": "Anonymized fresh-agent relevance fan-out: one fresh-context agent per query per run, each shown the lettered menu of all shipped skill descriptions (this skill unflagged) plus NONE, asked which ONE to load. Adjudicated on the agent's reasoning, not on the verdict token.",
+      "runs_per_query": 3,
+      "trigger_rate_threshold": 0.5,
+      "arms": ["with_skill", "without_skill"],
+      "without_skill_arm": "The same query, the same menu with this skill's entry removed, in a fresh context. A positive row that is answered correctly without the skill is evidence the skill is not earning that row; a negative row that goes wrong only in the with-skill arm is a false-trigger cost.",
+      "split": {"train": 0.6, "held_out": 0.4},
+      "held_out_ids": [6, 9, 10, 13, 17, 18, 19, 20],
```
Avick3110: Held-out membership is recorded twice, and the two copies already disagree: `held_out_ids` names eight rows (6, 9, 10, 13, 17, 18, 19, 20), while only id 6 carries `"held_out": true` on the row itself.

A runner keying on the per-row flag — the natural read for a per-row consumer — treats the other seven as training rows and reports a split that does not match `split: {"train": 0.6, "held_out": 0.4}` sitting four lines above. Keep one source of truth; the repo's own changelog rule applies here: point at the bound rather than restating it, because a second copy of a fact is the one that drifts.

.claude/skills/skse-plugin-authoring/references/toolchain-setup.md
```diff
@@ -228,7 +229,21 @@ target_precompile_headers(${PROJECT_NAME} PRIVATE PCH.h)                   # req
 ```
 
 Your `vcpkg.json` must carry `spdlog`, `fmt`, `directxtk`, `directxmath`, `rapidcsv` (rapidcsv because VR
-defaults on). One pairing hazard: the bare `<SKSE/Impl/PCH.h>`-only PCH is **incompatible** with an
+defaults on) — **and a `builtin-baseline`**, the commit sha of the vcpkg checkout the manifest pins its
+package versions to (`git -C %VCPKG_ROOT% rev-parse HEAD`). A manifest without it errors outright where
+classic mode is disabled, which is the default for manifest builds, and the message names versioning
```
Avick3110: The condition on this claim is vacuous, which makes the claim unconditional: manifest mode *is* classic mode being disabled, so "where classic mode is disabled, which is the default for manifest builds" reduces to "always, in manifest mode".

That is stronger than vcpkg behaves. A manifest with no `builtin-baseline` and no version constraints resolves against the ports in the current vcpkg checkout and configures fine; the `builtin-baseline` error is raised when the manifest actually uses versioning — `version>=`, `overrides`, or a registry. As written, a reader who omits the field is told to expect a hard failure that will not arrive, and then has no way to recognise the case where it does.

Two smaller things on the same addition: it carries no `path:line` provenance, which the rest of this file uses for surprising claims; and the prose requires five packages while the example JSON below lists six (`xbyak`).

.claude/skills/skse-plugin-authoring/SKILL.md
```diff
@@ -1,165 +1,193 @@
...
+```papyrus
+Int Function WaitForBounty(Faction akFaction) global native
+```
+
+**A narrowing return type.** `size_t`, `std::int64_t` and `double` all marshal — as Int, Int and
+Float — so this **compiles and then silently truncates**. Declare `std::int32_t` / `std::uint32_t`
+for Int and `float` for Float, and do the widening in C++ before you return.
+
+```cpp
+std::size_t CountRefs(RE::StaticFunctionTag*);   // WRONG: narrows to Int32 with no diagnostic
```
Avick3110: These two lines redeclare `CountRefs` differing only in return type. Pasted into a translation unit as shown they are a hard compile error (MSVC C2371) — the opposite of the "compiles and then silently truncates" point the block is making, and the one block in this section a reader is most likely to copy verbatim.

Give them distinct names, or comment out the wrong one, so the contrast survives the paste.

.claude/skills/skse-plugin-authoring/references/multi-runtime.md
Avick3110: Folded. Row 2 now stops at "cannot discriminate SE from AE" and hands off: in one dual DLL discriminate at runtime (`REL::Module::IsAE()`, `REL::Relocate(seAndVr, ae)`), with a pointer to the decision rule; compile-time `_SE`/`_AE` gating is marked as selecting an arm only in a separate `se` or `ae` build. Footgun 14 and the authoring rule at the tail carried the same unactionable instruction, so both take the same correction. d61cfd0.

.claude/skills/skse-plugin-authoring/SKILL.md
Avick3110: Folded, same correction as the reference. The bullet now leads with the runtime discriminator and says the macros are undefined in a flatrim build, so an `#if` on them drops the divergent code on both runtimes. d61cfd0.

.claude/skills/skse-plugin-authoring/SKILL.md
Avick3110: Refuted. The rewrite document measured this text and accepted the overrun: "**467 characters.** Name 21; combined listing entry 488 characters against Claude Code's 1,536 cap (D3.4) and the spec/Codex 1,024 description cap (D2.21, D5.3). Over the Help Center's 200, which the digest records as unreconciled — recorded, not treated as a fail." (`dev/skill-validation/eval-2026-09/skse-plugin-authoring/REWRITE.md:117-120`). The 200-400 band comes from #294's compression pass, not from a published cap; the caps the checklist actually grades are 1,024 and 1,536, and 467 clears both. D2 returning zero fails is consistent with that — the row is a partial the document moves, not a fail. I measure the description body at 467, not 472.

Aaron can still overrule it; the second sentence of the description is the cut if he wants one.

.claude/skills/skse-plugin-authoring/SKILL.md
Avick3110: Refuted on the risk, kept on the reconciliation.

`compatibility` is not an unrecognized key: it is one of the six the spec allows outside Claude Code — "only `name`, `description`, `license`, `allowed-tools`, `metadata` and `compatibility` are legal outside Claude Code, where any other key is a hard packaging/upload error" (`eval-2026-09/checklists/D2.md:240`), and D2.35 names it "the spec's stated home for runtime-level prerequisites". Carrying it is ruling 7, not a choice: "Every skill carries one compatibility frontmatter line (rewrite ruling 7) naming the houseCARL MCP server and an MO2 instance, same wording throughout" (`dev/DECISIONS.md:19`, 2026-09-07). No other skill on this branch carries it because this is wave 1 of twelve.

On the untested gate: I ran it. `claude plugin validate ./dist/housecarl --strict` against the rebuilt package returns "Validation passed". So the red-CI-after-review case does not arise.

The "shared sentence" wording in the PR body is fair to flag — it is shared prospectively, with the other eleven, not with anything already on the branch.

.claude/skills/skse-plugin-authoring/SKILL.md
Avick3110: Refuted as out of scope for this PR, and both halves are settled elsewhere.

Papyrus-reference's own rewrite document discharges F15 on its side and records why no edit is owed: "Discharged here by measurement rather than by an edit: the shipped body names no sibling skill (D5.48 conforms), so there is no handoff clause to remove — the obligation is that the rewrite adds none, and that the bundled-or-warn options list stays a lookup path for functions that exist rather than drifting into a declaration recipe for one that does not. The eval set's negative row q12 ... is the standing check on that" (`eval-2026-09/papyrus-reference/REWRITE.md:265-273`). Its description is that skill's wave to change, against its own regrade.

The Codex umbrella router is item 27, deferred behind #583, and has its own conditional rewrite document (`eval-2026-09/codex-umbrella/REWRITE.md`) — its verdict is "not a keep", so editing its routing table here would pre-empt a ruling that has not been made.

What this PR can carry it does: all 18 `papyrus-reference` sites in this skill's shipped tree are touched, and the surviving mention states the boundary in the new direction.

.claude/skills/skse-plugin-authoring/evals/evals.json
Avick3110: Folded. The note now has a row to point at: a `coexistence_set` beside the method, carrying "i'm adding a native to my SKSE dll — what does the .psc declaration for it look like on the script side?" with the expected pick, the wrong pick (`papyrus-reference`) and why that pick is the pre-rewrite failure reproducing. It sits outside `evals` so the trigger set stays at the 10/10 the rewrite document specifies. e7b7fe9.

.claude/skills/skse-plugin-authoring/evals/evals.json
Avick3110: Folded, taking the second option. The row stays held out — the document places it there (`REWRITE.md:483-486`) — but the "not used to tune" claim is gone: the note now says the description was fitted to this row, that a pass here checks the named fix landed rather than measuring generalization, and that rows 9 and 10 are the held-out positives that drove no edit and so carry the generalization claim. e7b7fe9.

.claude/skills/skse-plugin-authoring/evals/evals.json
Avick3110: Folded. The per-row `held_out` flag is removed; `held_out_ids` is the only record, with a note saying so and why. e7b7fe9.

.claude/skills/skse-plugin-authoring/references/toolchain-setup.md
Avick3110: Folded, all three. The condition is narrowed to what vcpkg actually does — the baseline is required once the manifest uses versioning (`version>=`, `overrides`, or a registry), and a manifest using none of those resolves against the current checkout and configures fine — with the reason to carry it anyway kept. The prose now names `xbyak`, matching the JSON's six.

On provenance: this is vcpkg behaviour, not a CommonLibSSE-NG one, so there is no `path:line` in the corpus to cite. Rather than invent one, it goes in the file's own "Not yet verified in-game" list, recorded as read from vcpkg's versioning docs and not observed failing on this machine. 19f04bf.

.claude/skills/skse-plugin-authoring/SKILL.md
Avick3110: Folded. The wrong declaration is commented out, so the block pastes as one declaration and the contrast survives. 1b25229.

.claude/skills/skse-plugin-authoring/SKILL.md
```diff
@@ -1,165 +1,195 @@
...
+  naive virtual call and an upcast each sit at a different offset or vtable slot per runtime; all of
+  it compiles clean under a single-runtime test.
+- **Export the whole triad.** A portable plugin exports `SKSEPlugin_Version` (AE data path),
+  `SKSEPlugin_Query` (SE/VR function path) and `SKSEPlugin_Load`. Drop `Query` and VR never sees the
+  plugin — silently skipped, no error dialog.
+- **Keep one metadata path.** When the build system generates the export triad
+  (`add_commonlibsse_plugin`, or the xmake plugin rule), a hand-written `SKSEPluginInfo` on top gives
+  duplicate exports and a load failure that names nothing.
+- **Set the log up before `SKSE::Init`, and never re-init it after.** `SKSE::Init(skse)` truncates a
+  log you have just configured, taking every pre-`Init` line with it.
```
Avick3110: **The fix is missing, so the bullet's imperative reproduces the bug it warns about.**

The body is the always-loaded surface; `plugin-skeleton.md` is a job-gated read. An author who follows this bullet writes:

```cpp
InitializeLog();
SKSE::Init(skse);
```

On an AE build the default `a_log = true` runs CommonLib's `log::init()`, which truncates `<PluginName>.log` and replaces the default logger, losing every pre-`Init` line. That is exactly the failure the bullet describes, and the actual remedy - passing `false` - appears nowhere in the body. `references/plugin-skeleton.md:242-244` states it, and the working sequence at `:172-174` is `SKSE::Init(a_skse, false);   // false = we own the logger`.

Worth noting the eval cost: `evals/evals.json` prediction P4 is "The control writes `SKSE::Init(a_skse)` after setting up logging and loses every pre-Init log line" - as written, the with-skill arm reproduces the control's miss.

Also: "never re-init it after" is not the reference's rule and does not appear in the corpus; the reference's rule is the `a_log` argument.

```suggestion
- **Set the log up before `SKSE::Init`, and pass `a_log = false`.** `SKSE::Init(skse)` with the
  default `a_log = true` runs CommonLib's own `log::init()`, which truncates the log you just
  configured and replaces your logger, taking every pre-`Init` line with it. Call
  `SKSE::Init(a_skse, false)` when you own the logger.
```

.claude/skills/skse-plugin-authoring/references/multi-runtime.md
```diff
@@ -31,6 +32,33 @@ surprising claim traces back to source.
...
+target is **SE + AE and no VR**, read these five rows and stop; the rest is for when VR joins the set.
+
+1. **The derivation is in `Common.h`, and it wins.** Authors set only `ENABLE_SKYRIM_SE` /
+   `ENABLE_SKYRIM_AE` / `ENABLE_SKYRIM_VR`; every `EXCLUSIVE_*` macro is derived from those at preprocess
+   time. Read the derivation verbatim in [the preprocessor define set](#the-preprocessor-define-set) rather
+   than any summary table — where a table and `Common.h` disagree, `Common.h` wins.
+2. **FLAT is not SE and is not AE.** A dual SE+AE build (preset `flatrim`) defines `EXCLUSIVE_SKYRIM_FLAT`
+   **alone**; `_SE` and `_AE` are *not* defined (a single-runtime build defines its own plus FLAT). FLAT
+   means "not VR", so it cannot discriminate SE from AE. In one dual DLL, discriminate at **runtime**:
+   `REL::Module::IsAE()` for a branch, `REL::Relocate(seAndVr, ae)` for an offset (`BaseExtraList` is the
```
Avick3110: **`BaseExtraList` is the wrong citation for the runtime-`Relocate` answer, and it contradicts this file's own decision rule.**

`BaseExtraList`'s SE/AE divergence is not an offset that `Relocate` can select between: per `:466-488`, AE 1.6.629+ promoted the dtor to virtual, adding a vtable pointer, so the *type's layout* changes - data at `0x00` size `0x10` on SE/VR, data at `0x08` size `0x18` on AE. The decision rule at `:176-180` puts exactly that class of difference in the must-be-`#if` bucket ("class / struct size", "vtable slot count", "base-class inheritance", "whether a member is present at all"), which is unexpressible in `flatrim` because neither `EXCLUSIVE_SKYRIM_SE` nor `_AE` is defined there.

Concretely: an author following this row writes `REL::Relocate(0x00, 0x08)` for the data offset and still has `sizeof`, any `STATIC_ASSERT_SIZE`, and every direct field access wrong on one runtime. For a layout divergence the honest answer is that `flatrim` cannot carry it - ship the `se` and `ae` presets (or stay on the generated `RUNTIME_DATA` accessor, which is what row 5 already says).

`SKILL.md:158` carries the same compressed advice with the layout exception dropped entirely ("Use `REL::Module::IsAE()` or `REL::Relocate(seAndVr, ae)`; the macros gate only a single-runtime build"), so the caveat is not recoverable from the body either.

.claude/skills/skse-plugin-authoring/evals/evals.json
```diff
@@ -0,0 +1,112 @@
...
```
Avick3110: **Row 6 is in the held-out split, and row 6 is the row the description was fitted to.**

Row 6's own note says so: "the row the description's 'reading someone else's SKSE DLL source' clause was written for. The description was therefore fitted to this row: a pass here checks that the named fix landed, and is not evidence of generalization."

Held-out is by definition the portion not used to fit, so a reported held-out score of 8/8 counts one row that was trained on and inflates the generalization claim by 1/8 - while the row's own note says the opposite. The same note already names the rows that do carry the claim: "Rows 9 and 10 are the held-out positives that drove no edit."

Swapping 6 for an untouched positive (1-5, 7 or 8) keeps the split at 8 of 20 and keeps the 3-positive / 5-negative shape. Which one to promote is a judgement call, so no suggestion block.

.claude/skills/skse-plugin-authoring/evals/evals.json
```diff
@@ -0,0 +1,112 @@
...
+    ],
+    "status": "not yet run against the rewritten description"
+  },
+
+  "output_quality_set": {
+    "purpose": "Does loading the skill produce a better artifact, and at what cost?",
+    "task": "Author a complete SKSE plugin source tree for a new native Papyrus function targeting SE (1.5.97) and AE (1.6.1170) against CommonLibSSE-NG, on the live ARR 2.0 order.",
+    "protocol": "Run twice in fresh contexts with a 40-call cap: once with the skill loaded from the packaged tree, once without it. Same task text, same model, same order.",
+    "arms": {
+      "with_skill": {"status": "exists", "evidence": "dev/skill-validation/audit-2026-09/skse-plugin-authoring/with-skill/ (deliverable HC_AuditNative/)"},
```
Avick3110: **The with-skill arm is credited to a run of the pre-rewrite skill, so the pass condition would compare a stale arm to a future control.**

The evidence path is the 2026-09 audit, which ran against the skill as it stood *before* this PR. But the protocol two lines above requires both arms "in fresh contexts", "Same task text, same model, same order", and `run_from` says the packaged tree - none of which the stored deliverable satisfies for the rewritten body.

It matters for the predictions this PR is the reason for: P4 (the `SKSE::Init` log truncation) and P6 (the narrowing return type) test content this PR adds. Scoring them against a with-skill artifact produced without that content measures the old skill while claiming to measure the new one. Whoever later runs "the missing control" would in good faith read `"status": "exists"` and skip re-running the with-skill arm.

Marking this arm stale rather than `exists` - or naming the commit it was run at - keeps the comparison honest.

.claude/skills/skse-plugin-authoring/references/toolchain-setup.md
```diff
@@ -228,7 +229,23 @@ target_precompile_headers(${PROJECT_NAME} PRIVATE PCH.h)                   # req
 ```
 
 Your `vcpkg.json` must carry `spdlog`, `fmt`, `directxtk`, `directxmath`, `rapidcsv` (rapidcsv because VR
-defaults on). One pairing hazard: the bare `<SKSE/Impl/PCH.h>`-only PCH is **incompatible** with an
+defaults on) and `xbyak` (the trampoline's assembler) — **and a `builtin-baseline`**, the commit sha of the
```
Avick3110: **`xbyak` is stated unconditionally here and conditionally at `:143` in the same file.**

Line 141-146 reads: "always `spdlog`, `fmt`, `directxtk`, `directxmath`; plus `rapidcsv` whenever VR support is on (the default); **`xbyak` only if you enable xbyak support**" - followed by a field note saying the practical reason to include it anyway is CommonLib's own `find_package` set, not the trampoline.

This line says "must carry ... `xbyak` (the trampoline's assembler)". A reader who reaches `:231` first always adds it; a reader who reaches `:143` first omits it. The gloss is also a new causal claim the file does not otherwise make.

Separately: the `builtin-baseline` addition in this same hunk got a provenance row at `:475` saying it was read from vcpkg's versioning docs and not observed failing on this machine. The `xbyak` change got no such row, and neither the PR body nor the changelog entry mentions it - so it lands uncited in a corpus whose discipline is `path:line` or an explicit not-yet-verified note.

Pointing at `:143` rather than restating the rule would also match the changelog's own "point at the bound rather than restating it".

.claude/skills/skse-plugin-authoring/references/_CORPUS_STATUS.md
```diff
@@ -1,130 +0,0 @@
-# skse-plugin-authoring — corpus status (dev-side; NOT shipped in the plugin)
```
Avick3110, line 1: **Deleting the ledger loses it from version control, and ruling 2 did not ask for the deletion.**

Ruling 2 bars a *pointer* at a packager-excluded file. That is satisfied by removing the pointers - which this PR also does. The file itself was already stripped from every shipped copy by `scripts/build-plugin.ps1:145`, so leaving it in the source tree cost nothing shipped.

What is lost: the reference-to-research-doc mapping, the pinned corpus date, and the staleness contract for all nine references. `dev/` is gitignored (`.gitignore:81`), so the copy the PR body says it moved to is untracked and machine-local; the next maintainer refreshing this corpus has to dig it out of this commit's parent.

It is also now inconsistent: `dialogue-authoring`, `kid-authoring`, `skypatcher-authoring` and `spid-authoring` all still carry `references/_CORPUS_STATUS.md` in the source tree, stripped by the same packager rule.

If the ledger really should leave `.claude/skills/`, that is a decision worth applying to all five at once rather than to this one skill.

.claude/skills/skse-plugin-authoring/SKILL.md
Avick3110: Folded, with your wording. The bullet names the argument now instead of an invented "never re-init" rule: `SKSE::Init(skse)` with the default `a_log = true` runs CommonLib's own `log::init()`, which truncates the log you just configured and replaces the logger, and the remedy `SKSE::Init(a_skse, false)` is written into the always-loaded body rather than living only at `references/plugin-skeleton.md:242-244`. `SKILL.md:170-173`. 4e1768fc.

P4 now has content in the with-skill arm to test.

.claude/skills/skse-plugin-authoring/references/multi-runtime.md
Avick3110: Folded, all three sites. `BaseExtraList` is a layout change — AE 1.6.629+ makes the dtor virtual, data moves `0x00` to `0x08` and size `0x10` to `0x18` (`references/multi-runtime.md:471-477`) — which the file's own decision rule at `:176-180` puts in the must-be-`#if` bucket, so it was the wrong citation for `Relocate`.

- Row 2 no longer names it as the `Relocate` case. It now says the runtime probe reaches an offset, not a layout — size, vtable slot count, base class, or a member present on one runtime only — and that `flatrim` cannot carry that difference because `sizeof`, `STATIC_ASSERT_SIZE` and direct field access stay wrong on one runtime; the answer is the generated accessor (row 5) or the `se` and `ae` presets. `references/multi-runtime.md:48-56`.
- The authoring rule at the tail carried the same implication one paragraph below the `BaseExtraList` case, so it takes the exception explicitly: `REL::Relocate(0x00, 0x08)` selects an offset, not a type's shape.
- `SKILL.md:158-161` carries the compressed form of the same caveat, so it is recoverable from the body.

8e4af812.

.claude/skills/skse-plugin-authoring/evals/evals.json
Avick3110: Folded, taking the swap. The earlier blind-review round raised the same row and was folded the other way (e7b7fe9: the "not used to tune" claim went, the row stayed held out on `REWRITE.md:483-486`) — which fixed the note and left the inflated score, your point here.

Row 6 moves to train; row 8 ("writing a `BSTEventSink` for `TESDeathEvent` — when exactly should I register it") takes the slot. It is a positive this rewrite aimed no edit at, so the split stays 8 of 20, 3 positives and 5 negatives. `evals/evals.json:18`. The rationale sits only in `held_out_ids_note`, and row 6's own note now says it is in train for the reason the note gives. Which positive to promote was the judgement call you flagged, so it is a cheap one-token change if you want a different one.

Recorded there too: this departs from `REWRITE.md:483-486`, since the document held row 6 out on a ground the earlier fold already established as false.

2c590ef1.

.claude/skills/skse-plugin-authoring/evals/evals.json
Avick3110: Folded. The arm reads `"status": "stale — re-run required"` with the evidence path kept and marked as run against the pre-rewrite skill, plus a `why_stale` saying what it costs: the stored deliverable carries none of the content P4 and P6 test, so scoring it against a fresh control would measure the old skill under the new skill's name. It also says both arms are re-run together, from the packaged tree, at the rewrite commit — so "run the missing control" no longer reads as a one-arm job. `evals/evals.json:60`. 2c590ef1.

.claude/skills/skse-plugin-authoring/references/toolchain-setup.md
Avick3110: Folded. The scaffold line no longer restates the dependency rule or glosses `xbyak` causally — it points at **Consumer obligations** (`references/toolchain-setup.md:141-146`), where the conditional rule and the field note on why a build carries it anyway already live, and names `xbyak` as riding that field note rather than as a requirement. The example JSON keeps its six, which is what the field note recommends. `references/toolchain-setup.md:231-233`. ca41cccc.

That also removes the uncited claim: there is no new causal statement left to cite.

.claude/skills/skse-plugin-authoring/references/_CORPUS_STATUS.md
Avick3110: Folded on the deletion. `references/_CORPUS_STATUS.md` is restored from e65109a2^ — ruling 2 bars a pointer at a packager-excluded file, and removing the pointers satisfies it; `scripts/build-plugin.ps1:145` strips the file from every shipped copy regardless, and the four sibling skills still carry theirs. A grep for `_CORPUS_STATUS` in the skill returns nothing but the file itself, and the packager's pointer scan is clean. 426dd4d3.

**Left for Aaron:** whether the ledger should leave `.claude/skills/` at all. That is the five-skill decision you name, not this skill's, so this PR puts it back rather than deciding it here.

Note the PR body still describes the ledger as moved out; that sentence is now wrong and the file is the record.

PR #642, opened 2026-09-08
Rewrite the papyrus-reference skill and fix its corpus

Avick3110:
Rewrites the `papyrus-reference` skill and fixes the corpus defects the 2026-09 audit found, executing the approved rewrite document with the ten rows its verifier overturned corrected as the verifier says. The body drops the Tier-2 gate keyed on a `requires_plugin` DLL filename — that filename is `PapyrusUtilSE.dll` in the index and `PapyrusUtil.dll` on an AE install, so following the old instruction literally refused a plugin that was present and working — and defers the availability question to `housecarl_skse` with `findings='pairing'`, carrying that tool's own ceiling (a declaration is not a runtime guarantee) into the sentence that makes the call. It names `housecarl_compile_script` as the deterministic check behind "never invent a signature", rewrites the bundled-miss options against the 2.0 surface (`housecarl_skse` pairing, `housecarl_decompile_script`, `housecarl_bsa_extract`, and the plain statement that parameter defaults do not survive a decompile), pulls the eight silent-biter rules inline behind a trigger condition, relaxes the block-read rhythm from one call per entry to line-exact reads that may be batched, cuts Tier 3 and the authoring note, and removes every figure nothing maintains. It also states one coverage bound outright: base-class events on `Form.psc` have no vanilla or skse row, so a miss there is reported as a corpus hole and never as an absence from Papyrus. In the corpus, 32 vanilla files that carried no index entry are gone, every reference file over 100 lines gained a `## Contents` table, and the index's line ranges were re-derived so all 7,345 rows still land on their own `###` block. The maintainer lane moved out of the body into `references/corpus-notes.md`, and `evals/eval_set.json` became `evals/evals.json` with a without-skill arm on every trigger query, both rows that routed to the deleted `papyrus-optimization` expecting NONE, and the corpus checks written as a verification script rather than as judgement.

No issue is named as closing by the rewrite document, so this PR closes none.

**Decided as follows.**

- **Q4, the 32 function-less vanilla stubs — dropped.** The verifier overturned this as already answered: the report's own fix says to drop them at regeneration because a dead read is all they can produce, and the document adopts the drop twice. Verified before deleting that no index row referenced any of them.
- **Q6, the `compatibility` line — carried as ruled, unvaried.** The verifier overturned the second half of the question: ruling 7 fixes the wording for all twelve skills. The body scopes it in one clause instead — the lookup works with no server, and only the availability half needs one.
- **Q5 in part, `paths` — not added.** The skill fires on a compile error with no `.psc` open, and no source makes `paths` an obligation.

**Left for Aaron.**

- **Q1, ruling 2 against a 149-file generated corpus.** Implemented as the document reads it: the body names `index.jsonl`, `silent-biters.md` and `corpus-notes.md` one hop each, declares the index as the intended second hop, and names the generated files as a class. The packaged pointer scan passes on both trees. Confirm that reading of "the body names every shipped file once".
- **Q2, where the missing base-class rows come from.** Not fixable here — the corpus is generated from upstream and nothing in the inputs says whether upstream carries a vanilla `Form#OnInit` row. The alternative the document names is implemented meanwhile: the body states the coverage bound, and `references/corpus-notes.md` records what closing it would take (an upstream release, or a hand-authored supplement in its own file) and says the bound leaves the body in the same change.
- **Q3, whether the body calls a tool at all.** Implemented as the document proposes, with the verifier's correction: the body now names **five** houseCARL tools, not four (`housecarl_skse`, `housecarl_compile_script`, `housecarl_decompile_script`, `housecarl_bsa_extract`, `housecarl_asset_status`), in the one skill that previously named none. Confirm the deferral or bound it to the availability question alone.
- **Q5, the invocation posture.** The skill stays model- and user-invocable and the body documents the function-name argument in one clause, but the `arguments` frontmatter key is **not** declared: the verifier records it as a Claude Code extension outside the spec's portable six, so declaring it would turn D2.30 from conforming into non-conforming, which the regrade gate does not allow. Declare it and take the portability cost, or leave it as prose?

## Verifier overturns folded in

- The existence argument is re-grounded (rows 6 and 49): `housecarl_compile_script` auto-discovers the enabled mods' own Papyrus source folders, so a control **can** read a default out of shipped `.psc`. The body now says the corpus is the cheap and complete route rather than the only one, and the eval's output-quality arm marks the three default expectations as graded but not assumed to differentiate, naming three others that carry the claim.
- Both q10 **and** q11 route to the deleted skill (row 50); both are edited, and q17's route was corrected for the same deletion.
- The `findings='inventory'` and `findings='pairing'` diagnostic lists are attributed to the right family (row 31): BSA-only and 32-bit are pairing's PAIRED-BUT-DEAD, not inventory's.
- The frontmatter carries three keys and does not call `arguments` one of the spec's six (row 18).
- The `## Contents` threshold is stated: over **100** lines, the stricter of the two thresholds the sources disagree on, so it clears both (row 28's caveat). 87 files, not the 48 a >300-only emit would have covered.
- The aggregates and the index size are recorded as a deliberate trade in `references/corpus-notes.md` (row 29, D4.66).

## Checks

`scripts/build-plugin.ps1` clean, pointer scan included (12 skills, 219 markdown files, both trees). `dotnet build -c Release` clean; 1,858 tests pass; `ci-all` 128/128. (Re-run after Aaron's review was folded.)

## Regrade

Assembled in `dev/skill-validation/eval-2026-09/papyrus-reference/REGRADE.md` from the five grader
files under `regrade/`, graded on this branch's copy of the skill.

| Angle | Rows | Conforms | Partial | Fails | n/a | Fails by type |
|---|---|---|---|---|---|---|
| D1 — discovery and packaging | 46 | 44 | 0 | 1 | 1 | 1 packager-only (D1.28) |
| D2 — frontmatter | 52 | 40 | 0 | 0 | 12 | — |
| D3 — loading and invocation | 33 | 28 | 0 | 0 | 5 | — |
| D4 — body, supporting files, style | 64 | 36 | 4 | 1 | 23 | 1 other (D4.42) |
| D5 — evals and integration | 53 | 29 | 15 | 2 | 7 | 2 other (D5.21, D5.26) |
| **Total** | **248** | **177** | **19** | **4** | **48** | 1 packager-only, 3 other, **0 factual-or-dead-call** |

**PASS** against the regrade thresholds. D2 and D3 have zero fails. D4 and D5 have zero
factual-or-dead-call fails: D4.42 is the bare-vs-prefixed MCP tool name, a standing project ruling,
and D5.21 and D5.26 are unrun measurements rather than wrong facts — every tool, parameter and value
the body names checks out against `surface/tools.md`. D1's only fail is D1.28, release tagging,
packager-only and excluded by ruling 8.

Outstanding but not gating: the with/without eval pair is specified in `evals/evals.json` and has not
been run, which is also the single shared reason behind nine D5 partials.


.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,217 @@
...
+   that compiles binds; a call the corpus lacks that compiles is a corpus hole to report.
+
+Never invent a signature, including under pressure to "just guess". A confidently wrong signature
+ships code that compiles and misbehaves at runtime, or turns an authoring error into a filed
+upstream bug — both cost far more than the non-answer.
+
+## Tier 2 — the corpus carries more than a modlist installs
+
+The corpus ships two tiers, and the tier is a fact about the corpus, not about the user's machine.
+Tier 1 is `vanilla/` and `skse/`, which any Skyrim + SKSE install has; their rows carry no
```
Avick3110: **The two-tier `requires_plugin` rule is false for 549 rows, and it fails open.**

The body makes the presence of `requires_plugin` the tier discriminator: Tier 1 rows carry none, Tier 2 rows do. Measured on this branch's `index.jsonl`, 549 rows from Tier-2 sources carry no `requires_plugin` at all — `clib` (515, cLib-Papyrus Function Library), `dynamicwetness` (28, Dynamic Wetness), `skyprompt` (6, SkyPrompt). All three are third-party mods; none of them ships with Skyrim + SKSE.

(The other half of the rule holds: zero `vanilla`/`skse` rows carry the key.)

Concrete failure: a user asks about `cArrayCreateActor`. The grep resolves to

```
{"name":"cArrayCreateActor","qualified":"cArrayCreateBase.cArrayCreateActor","source":"clib",...}
```

no `requires_plugin`. Applying this paragraph literally, the model classifies it Tier 1 — "which any Skyrim + SKSE install has" — and reports the function as always available, skipping the `housecarl_skse findings='pairing'` check the next paragraph exists to route it to. The user writes a call that fails to bind at runtime. This is exactly the wrong-answer class this section is here to prevent, and it fails silently in the permissive direction.

The reliable discriminator is already in every row: `source`. Tier 1 is `source` in {`vanilla`, `skse`}; everything else is Tier 2 whether or not the row names a DLL.

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,217 @@
...
+a real miss — it routes a present function into the warning path, the exact failure this skill
+exists to prevent. **Validate the instrument before you trust a zero:** grep a token you know is
+there (`"name":"OnInit"` resolves to several entries). Only once the pattern is proven is an empty
+result "not in the corpus".
+
+1. **Take the unqualified name from the call site.** `Self.GetActorValue("Health")` → `GetActorValue`;
+   `StringUtil.Substring(s, 0, 4)` → `Substring`.
+2. **Grep `references/index.jsonl` for `"name":"<Name>"`.** Three result shapes:
+   - **One match.** Take its `file`, `line_start`, `line_end`.
+   - **Several matches.** Disambiguate on `qualified`. A qualified call site is authoritative —
```
Avick3110, line 67: **`qualified` does not disambiguate — it collides on 1,860 rows, and the step's own example is one of the collisions.**

Measured on this branch's `index.jsonl`: 930 distinct `qualified` values map to two or more rows (1,860 rows, 25% of the corpus), almost all of them a `vanilla`/`skse` pair. The worked resolution this step gives — "`Self.GetActorValue(...)` in a script extending `Actor` is `Actor#GetActorValue`" — resolves to two rows, not one:

```
{"name":"GetActorValue","qualified":"Actor#GetActorValue","source":"skse","file":"references/skse/Actor.md","line_start":802,...}
{"name":"GetActorValue","qualified":"Actor#GetActorValue","source":"vanilla","file":"references/vanilla/Actor.md","line_start":734,...}
```

So step 2 terminates without a single row for a quarter of the corpus and the procedure says nothing about what to do next.

Usually harmless — 926 of the 930 groups render byte-identical blocks — but four disagree, and they disagree on exactly the thing the lookup exists to answer. `ObjectReference#RemoveAllStolenItems`:

- `references/skse/ObjectReference.md` 1630-1639 — `akTransferTo` `ObjectReference` **Required ✓**
- `references/vanilla/ObjectReference.md` 1502-1513 — `akTransferTo` `ObjectReference` **not required**, plus the doc-comment the skse block drops

The other three are `ObjectReference#GetAllItemsCount`, `#IsContainerEmpty`, `#SetContainerAllowStolenItems`.

Concrete failure: the model looks up `RemoveAllStolenItems`, grep prints the skse row first, it reports `akTransferTo` as required, and the user is told to pass a target for a call whose vanilla declaration takes none. Flip the grep order and it reports the opposite. Neither answer is marked uncertain.

Fix is a third clause on step 2: when `qualified` still ties, break on `source` — prefer `skse` (an SKSE install extends the vanilla class) — and say so when the two blocks differ.

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,217 @@
...
+That trap fails silently. A spaced `"name": "Dispel"`, or one missing the leading quote, matches
+**zero lines** for a function that is present, and a format-induced zero is indistinguishable from
+a real miss — it routes a present function into the warning path, the exact failure this skill
+exists to prevent. **Validate the instrument before you trust a zero:** grep a token you know is
+there (`"name":"OnInit"` resolves to several entries). Only once the pattern is proven is an empty
+result "not in the corpus".
+
+1. **Take the unqualified name from the call site.** `Self.GetActorValue("Health")` → `GetActorValue`;
+   `StringUtil.Substring(s, 0, 4)` → `Substring`.
+2. **Grep `references/index.jsonl` for `"name":"<Name>"`.** Three result shapes:
```
Avick3110: **The prescribed grep is case-sensitive; Papyrus identifiers are not. That is a second silent zero, and it is not covered.**

The section immediately above this one is devoted to the format-induced zero — "a format-induced zero is indistinguishable from a real miss — it routes a present function into the warning path, the exact failure this skill exists to prevent" — and prescribes a canary against it. But it treats spacing as the only way to manufacture one. Casing does it too, and casing is far more likely to appear at a real call site, because Papyrus is case-insensitive and modders write whatever they like.

Measured on this branch:

```
grep -c '"name":"Notification"'   index.jsonl  -> 1
grep -c '"name":"notification"'   index.jsonl  -> 0
grep -c '"name":"getactorvalue"'  index.jsonl  -> 0
```

Concrete failure: the `.psc` under edit contains `Debug.notification("...")` — legal Papyrus, and common in the wild. Step 1 says "take the unqualified name from the call site", which gives `notification`. Step 2's pattern returns zero. Step 2's third shape sends it to Bundled-or-warn, and the model emits "No Papyrus reference for `notification` — not in the bundled corpus" for a function that is in the corpus one line away. The canary does not catch it either: `"name":"OnInit"` is typed from this file in its correct casing, so the instrument validates while the real query is still miscased.

One clause fixes it: match case-insensitively (`grep -i '"name":"notification"'`), and take the corpus's casing off the row rather than the call site's.

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,217 @@
...
+- `qualified` — `Script.Function`, `Script#Method`, `Script.Event`, `Script.Property`. Disambiguates
+  a `name` that collides across sources.
+- `source` — the source directory the entry came from (`vanilla`, `skse`, `papyrusutil`, …).
+- `file` — the reference file holding the entry, rooted at the skill folder.
+- `kind` — `global`, `instance-method`, `event` or `property`.
+- `line_start` / `line_end` — 1-indexed inclusive block range inside `file`.
+- `requires_plugin` — Tier 2 only, and indicative (see above).
+
+**The index is the intended second hop.** Every other file under `references/` is reached through a
+row's `file` field and read at its `line_start`..`line_end`; each carries a `## Contents` table for
```
Avick3110: **Both halves of this sentence are false as written.**

*"each carries a `## Contents` table"* — 87 of the 151 `.md` files under `references/` carry one; 64 do not. The rule that was actually applied is "over 100 lines" (`corpus-notes.md` post-pass 3, and the PR body), so everything at or under 100 lines was skipped — including the ones a fallback read is most likely to land on: `skse/FormList.md` (99 lines), `vanilla/Weather.md` (97), `skse/Camera.md` (96), `skyregex.md` (95), `skyprompt.md` (90), `skse/LeveledActor.md` (84), `vanilla/FormList.md` (80).

*"Every other file under `references/` is reached through a row's `file` field"* — `silent-biters.md` and `corpus-notes.md` are under `references/` and neither is reachable from the index (`corpus-notes.md` says so itself: "Never resolved by the index"). The body contradicts this two sections earlier, where it instructs a direct read: "Read `references/silent-biters.md` before any call that…" — a whole-file read of a file this sentence says is only ever reached at a `line_start`..`line_end`.

Concrete failure: a malformed block sends the model to the fallback read this sentence promises, on `skse/FormList.md`. It opens the file expecting a `## Contents` table, finds none, and either re-reads whole-file (the waste the paragraph is written to prevent) or reports a corpus bug that isn't one.

Stating the actual bound fixes it: generated files **over 100 lines** carry a `## Contents` table; `silent-biters.md` and `corpus-notes.md` are hand-authored, are not index-reachable, and are read whole by name.

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,217 @@
...
 description: >-
-  Look up Papyrus function signatures, parameter and return types, and flags from bundled references — vanilla + SKSE + ~45 SKSE-plugin APIs (PapyrusUtil, JContainers, MCMHelper, …). Use for any .psc read or edit, any added or changed function call, or an unknown-identifier / type-mismatch compile error. Unbundled functions get a warning, never an invented signature.
+  Looks up Papyrus function, event and property signatures — parameter order, types, defaults,
+  return type and flags — in a bundled offline corpus covering vanilla, SKSE and the shipped
+  SKSE-plugin APIs (PapyrusUtil, JContainers, MCMHelper, po3, SkyUI). Use for any .psc read or
+  edit, any added or changed function call, and for an unknown-identifier or type-mismatch
+  compile error. Covers the API surface only — not reading a modlist's .psc source, not
+  compiling it. A function the corpus does not carry gets a warning and a check path, never an
+  invented signature.
+compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
```
Avick3110, line 11: **This `compatibility` gate is new in this PR, it is the only one in the tree, and it is false for the skill's main path.**

`git show e2990653:.claude/skills/papyrus-reference/SKILL.md` has no `compatibility` key, and no other shipped skill has one either:

```
$ grep -n '^compatibility:' .claude/skills/*/SKILL.md
.claude/skills/papyrus-reference/SKILL.md:11: ...
```

Eleven skills ship; ten of them genuinely do need the server and a configured instance, and none of them declares this. The one that declares it is the one whose primary path — grep `references/index.jsonl`, read a line range out of a bundled markdown file — needs neither. The description one line above even says so: "in a bundled offline corpus".

The PR body says the body "scopes it in one clause instead — the lookup works with no server, and only the availability half needs one." I can't find that clause in `SKILL.md`. "An offline corpus of…" describes the corpus, not the requirement, and nothing in the body says the lookup runs without a configured instance.

Concrete failure: a user with the plugin installed but no MO2 instance configured yet asks for the signature of `StringUtil.Substring`. The frontmatter says the skill requires a configured instance, so it is declined or the user is sent to `housecarl_set_mo2_instance` first — for a grep over a file that ships inside the skill. Only the two availability checks (Bundled-or-warn step 2, and the Tier-2 section) touch the server at all.

Either scope the line — "the lookup is offline; the Tier-2 availability check needs the houseCARL MCP server and a configured Mod Organizer 2 instance" — or drop the key and match the other ten.

.claude/skills/papyrus-reference/references/papyrusutil.md
```diff
@@ -4,6 +4,48 @@
...
+|---|---|---|
+| Global Functions | 5 | 49 |
+| `JsonUtil` | 0 | 118 |
+| Global Functions | 135 | 124 |
+| `MiscUtil` | 0 | 1739 |
+| Global Functions | 18 | 1745 |
+| `ObjectUtil` | 0 | 2003 |
+| `PapyrusUtil` | 0 | 2010 |
+| Global Functions | 3 | 2016 |
+| Return the total sum of all values stored in the given array | 4 | 2045 |
```
Avick3110: **The new Contents table on the largest Tier-2 file lists doc-comments as sections.**

20 of this table's rows are not sections. They are function doc-comments that upstream emitted as `## ` headings, and the generator that built this table walked `^## ` and took them at face value:

```
| Return the total sum of all values stored in the given array | 4 | 2045 |
| Get an array of values from ArrayValues1 that ARE NOT among ... | 0 | 2293 |
| Setting CompareBoth = true will change the behavior to also ... | 0 | 2294 |
| Setting IncludeDupes = true will allow the resulting array ... | 0 | 2295 |
```

Lines 2293-2295 are three consecutive lines of one function's doc-block, listed as three sections with 0 entries each. The three real sections in that stretch (`ActorUtil`, `JsonUtil`, `MiscUtil`, `PapyrusUtil`, …) are interleaved with them and also carry 0.

The underlying `##`-in-a-doc-comment defect is upstream and pre-existing — the entry blocks themselves are fine, and every one of this file's index rows still lands on its own `### ` heading — but the table is new here, and it is the one artifact the body promises for the fallback read on a 6,388-line file, the largest in the corpus. As shipped it makes that fallback worse than no table.

Two ways out, either fine: filter the table to headings that actually head a section (one with a nonzero entry count, or matching the known section names — `Properties`, `Events`, `Functions`, `Global Functions`, and the backticked class names), or add pass 3.5 to `corpus-notes.md` to normalize prose `##` lines to plain paragraphs at regeneration, which also stops the block for `PapyrusUtil.AddIntValues` and 19 others from spanning a stray heading.

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in a308e5ed. Both halves confirmed: 87 of 151 `.md` files under `references/` carry a `## Contents` table, and the index resolves 149 files, neither of them `silent-biters.md` or `corpus-notes.md`.

The paragraph now states the actual bound: every **generated** file is reached through a row's `file` field and read at its `line_start`..`line_end`; generated files **over 100 lines** carry a `## Contents` table for the rare fallback read, the shorter ones do not and that is not a corpus bug; and two files sit outside all of it â€” `silent-biters.md` and `corpus-notes.md` are hand-authored, reachable from no row, and read whole by name. That last clause also settles the contradiction you spotted with the direct read instruction two sections earlier.

.claude/skills/papyrus-reference/references/papyrusutil.md
Avick3110: Folded in 77e4f10d, taking the first of your two options.

Scope check first: `papyrusutil.md` is the only file in the corpus with prose `##` lines â€” a walk of every `## ` heading in `references/` finds 22 non-section headings and all 22 are in that file (`silent-biters.md` and `corpus-notes.md` are hand-authored and carry no table).

The table is rebuilt from the headings that head a section â€” `Properties`, `Events`, `Functions`, `Global Functions`, and the backticked class names â€” so it is 10 rows, not 32, and the entry counts the prose rows were splitting are whole again: `PapyrusUtil`'s Global Functions reads 97 rather than 3. The header shrank by 22 lines, so the file's 537 index rows were shifted with it, and pass 4's check re-run over all 7,345 rows: every `line_start` still lands on its own `### ` heading naming that row. Every `## Contents` line number in all 87 tables re-verified too.

`corpus-notes.md` pass 3 now carries the rule so a regeneration does not put the prose back, and `evals.json` gains corpus check c5, which walks every table row and fails on a heading that does not head a section. The recorded 6,388-line figure for the file is now 6,366.

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in a308e5ed. Confirmed on this branch's index: 549 rows from `clib` (515), `dynamicwetness` (28) and `skyprompt` (6) carry no `requires_plugin`, and no `vanilla`/`skse` row carries one.

The Tier 2 section now reads the tier off `source` — Tier 1 is `source` `vanilla` or `skse`, everything else is Tier 2 whether or not the row names a DLL — and says outright that reading a missing key as Tier 1 reports a third-party function as always present and skips the availability check. The `source` and `requires_plugin` field descriptions in "The index format" carry the same correction, and `evals.json` gains corpus check c6 so a regeneration that reintroduces the confusion fails mechanically.

The body keeps the count out (the rewrite drops figures nothing maintains); it names the three sources instead.

*(Edited: this reply briefly carried text belonging to another PR — a scratchpad collision on my side, not a change of position.)*

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in a308e5ed. Reproduced: 930 `qualified` values map to two or more rows (1,860 rows), and rendering each group's blocks shows exactly four groups whose text differs — `ObjectReference#RemoveAllStolenItems`, `#GetAllItemsCount`, `#IsContainerEmpty`, `#SetContainerAllowStolenItems` — all four on whether a parameter is required.

Step 2's "Several matches" branch now ends with the third clause: `qualified` does not always settle it, and when it ties, break on `source` and prefer `skse` because an SKSE install extends the vanilla class; almost every such pair renders the same block, and where the two differ they differ on whether a parameter is required, so read both and say which declaration you took. The "Getting it wrong" bullet and the `source` field description carry the same order — `name`, then `qualified`, then `source`.

*(Edited: this reply briefly carried text belonging to another PR — a scratchpad collision on my side, not a change of position.)*

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in a308e5ed. Reproduced: `grep -c '"name":"notification"'` is 0, `grep -ci` is 1.

The section now carries a second paragraph beside the spacing trap: casing manufactures the same silent zero and more often, because Papyrus is case-insensitive and a call site is written however its author liked; `Debug.notification("...")` is legal and the corpus row is `Notification`. The instruction is to grep case-insensitively (`grep -i`) and take the casing off the matched row, never off the call site. Step 2's imperative and the "Getting it wrong" bullet both say case-insensitively, and `evals.json` gains corpus check c7 for it.

You are right that the canary does not catch this one — it validates the pattern's shape, not the query's casing — which is why the fix is in the match rule rather than in another check.

*(Edited: this reply briefly carried text belonging to another PR — a scratchpad collision on my side, not a change of position.)*

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Half folded, half refuted.

**Refuted — the key stays, with this wording.** Dropping or varying the line is settled the other way by a ruling: `dev/DECISIONS.md` 2026-09-07 — "Every skill carries one compatibility frontmatter line (rewrite ruling 7) naming the houseCARL MCP server and an MO2 instance, same wording throughout" — and `REWRITE.md:78` prints the exact line this branch ships and answers the question you are raising in its own §8 Q6: carry the shared wording as written and let the body scope it in one clause. That this is the only skill carrying it today is the wave order, not a divergence: the ruling is a day old and this is the first of the twelve rewritten under it.

**Folded in a308e5ed — the clause you could not find.** You were right that it was not there; the Overview asserted nothing about the server. It now says: the lookup itself is offline, the index and reference files ship inside this skill, reading them needs no server and no configured instance, and the frontmatter's compatibility line is the prerequisite for the availability half only — the checks that ask whether a function is real on this machine call the server, and a lookup never has to wait on one. That is the sentence your `StringUtil.Substring` case needed.

.claude/skills/papyrus-reference/references/messagebox.md
```diff
@@ -4,6 +4,17 @@
...
 ---
 
+## Contents
+
+Entry blocks are reached from `index.jsonl` at their `line_start`..`line_end`;
+this table is for a fallback read, not the lookup path.
+
+| Section | Entries | Line |
+|---|---|---|
+| Global Functions | 23 | 18 |
```
Avick3110: **The `## Contents` table counts doc-comment sub-headings as entries.**

This table says the `Global Functions` section holds 23 entries. It holds 8: `Delete`, `GetResultIndex`, `GetResultText`, `IsMessageResultAvailable`, `Show`, `Show_NonBlocking`, `ShowArray`, `ShowArray_NonBlocking`. The other 15 `### ` lines in this file are prose inside doc-comments (`### Example`, `### Gotcha!`, `### Get name of selected button`, `### Button names`, `### Advanced`).

This is the same defect 77e4f10d fixed for `## ` headings in `papyrusutil.md`, one heading level down. The consequence is exactly what the table exists for: someone doing the fallback read is told to expect 23 functions and hunts for 15 that do not exist.

`evals/evals.json` check c5 cannot catch it either — it asserts that each row *names a heading that exists at the stated line and heads a real section*, never that the count is right. Every other Contents table in the tree has a correct count; this is the only one that does not, so c5 would want an entry-count clause to hold the line.

```suggestion
| Global Functions | 8 | 18 |
```

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,240 @@
...
+Read `references/papyrusutil.md` lines 5163-5175 and the block gives
+`FormListAdd(ObjKey, KeyName, value, allowDuplicate) → Int`, `Native Global`, with
+`allowDuplicate: Bool` defaulting to **`true`**. That default is the payoff: without a
+`FormListClear` first, a re-init silently doubles the list — and it still compiles.
+
+## Silent biters
+
+Eight traps a correct signature does not reveal. Each compiles clean and then misbehaves:
+
+- `Game.GetForm` returns `None` for the whole ESL range — use `Game.GetFormEx`.
```
Avick3110: **This drops half the `GetForm` trap, and it is the half a large order hits.**

`references/silent-biters.md:14` says `GetForm` returns `None` for any FormID `>= 0x80000000` — "the entire ESL / ESPFE light-plugin range *and* load index `0x80+`". The inline rule keeps only the ESL half.

The PR pulls these eight rules inline precisely so `silent-biters.md` is read conditionally, which makes this line what most sessions actually see. On an order past 128 regular plugins — ARR 2.0, the reference order, has 3,801 — a model checks that the plugin is not ESL-flagged, concludes `GetForm` is safe, and ships a quest that silently rebuilds `None` from stored FormIDs. That is the failure the rule exists to prevent.

```suggestion
- `Game.GetForm` returns `None` for any FormID `>= 0x80000000` — the whole ESL range *and* load
  index `0x80+` — use `Game.GetFormEx`.
```

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,240 @@
...
 description: >-
-  Look up Papyrus function signatures, parameter and return types, and flags from bundled references — vanilla + SKSE + ~45 SKSE-plugin APIs (PapyrusUtil, JContainers, MCMHelper, …). Use for any .psc read or edit, any added or changed function call, or an unknown-identifier / type-mismatch compile error. Unbundled functions get a warning, never an invented signature.
+  Looks up Papyrus function, event and property signatures — parameter order, types, defaults,
+  return type and flags — in a bundled offline corpus covering vanilla, SKSE and the shipped
+  SKSE-plugin APIs (PapyrusUtil, JContainers, MCMHelper, po3, SkyUI). Use for any .psc read or
+  edit, any added or changed function call, and for an unknown-identifier or type-mismatch
+  compile error. Covers the API surface only — not reading a modlist's .psc source, not
+  compiling it. A function the corpus does not carry gets a warning and a check path, never an
+  invented signature.
+compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
```
Avick3110, line 11: **The `compatibility` line states a prerequisite the body then spends a paragraph denying.**

Lines 23-24 say "reading them needs no server and no configured instance", and lines 24-26 exist only to walk this line back. Nothing in the repo consumes the key either — a grep over `scripts/`, `src/` and the other ten skills finds no reader — so its only effect is to tell a human, or a future loader that does read it, that a fully offline signature lookup is gated on a configured MO2 instance.

Scoping it to the availability half would make the frontmatter true on its own and let the Overview drop the correction:

```suggestion
compatibility: The lookup is offline. The availability checks require the houseCARL MCP server and a configured Mod Organizer 2 instance.
```

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,240 @@
...
+through a row's `file` field and read at its `line_start`..`line_end`, and no read off the index is
+ever whole-file. Generated files **over 100 lines** carry a `## Contents` table for the rare
+fallback read; the shorter ones do not, and that is not a corpus bug. The index itself is grepped,
+never bulk-loaded. Two files sit outside all of this: `silent-biters.md` and `corpus-notes.md` are
+hand-authored, are reachable from no row, and are read whole by name.
+
+## Getting it wrong, and the rule instead
+
+- Grep the index case-insensitively with the compact token, `grep -i '"name":"Dispel"'` (a spaced
+  `"name": "Dispel"` and a miscased `"name":"dispel"` both match nothing and read as absence).
```
Avick3110: **This bullet contradicts the `grep -i` it prescribes in the same sentence.**

`grep -i '"name":"dispel"' references/index.jsonl` resolves `ActiveMagicEffect#Dispel` in both the vanilla and the skse row. Under `-i` a miscased pattern does not match nothing.

Line 50 gets this right by scoping it — "a **case-exact** grep for `"name":"notification"` returns zero". Here the scoping is gone, so a reader who takes the bullet literally treats a genuine zero from a case-insensitive grep as a pattern defect and re-greps instead of entering bundled-or-warn.

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,240 @@
...
+2. **Grep `references/index.jsonl` case-insensitively for `"name":"<Name>"`.** Three result shapes:
+   - **One match.** Take its `file`, `line_start`, `line_end`.
+   - **Several matches.** Disambiguate on `qualified`. A qualified call site is authoritative —
+     `StringUtil.Substring` is the row whose `qualified` is `StringUtil.Substring`; `solveObjSetter`
+     exists on `JDB`, `JFormDB` and `JValue`, and only the call site says which. An unqualified call
+     inside a script body resolves up the calling script's `extends` chain — `Self.GetActorValue(...)`
+     in a script extending `Actor` is `Actor#GetActorValue`. **`qualified` does not always settle
+     it:** many class members carry both a `vanilla` and an `skse` row under one `qualified`, and
+     `Actor#GetActorValue` is one of them. When it still ties, break on `source` and prefer `skse` —
+     an SKSE install extends the vanilla class. Almost every such pair renders the same block; where
```
Avick3110, line 74: **"they differ on whether a parameter is required" holds for 1 of the 4 pairs that actually differ.**

Measured on this branch's corpus: 835 `qualified` names carry both a vanilla and an skse row. 831 render byte-identical blocks. Of the 4 that differ:

- `ObjectReference#RemoveAllStolenItems` — required-ness (vanilla leaves `akTransferTo` optional, skse marks it required). The case this sentence describes.
- `ObjectReference#GetAllItemsCount` — skse drops the doc-comment.
- `ObjectReference#SetContainerAllowStolenItems` — skse drops the doc-comment.
- `ObjectReference#IsContainerEmpty` — skse replaces the doc-comment with "added in 1.6.1126".

So three quarters of the time a model that reads both blocks, sees a difference and applies this rule will report a declaration conflict over a dropped sentence of prose. Naming the doc-comment case as the common one would keep the instruction honest.

.claude/skills/papyrus-reference/SKILL.md
```diff
@@ -1,115 +1,240 @@
...
+this skill ships. It is generated from BellCube's
+[papyrus-index](https://github.com/BellCubeDev/papyrus-index) and lives under `references/`.
+
+The lookup itself is offline: the index and the reference files ship inside this skill, and reading
+them needs no server and no configured instance. The frontmatter's compatibility line is the
+prerequisite for the availability half only — the checks that ask whether a function is real on this
+machine call the houseCARL MCP server, and a lookup never has to wait on one.
+
+This covers the **API surface**. Reading a modlist's actual `.psc` source is a file job (ask
+`housecarl_asset_status` which mod or BSA wins a `Scripts\...` path first, then read it with your
```
Avick3110, line 29: **"read it with your own file tool" dead-ends when the winner is BSA-packed.**

`housecarl_asset_status` explicitly reports whether the winning copy is loose or inside an active BSA, and framework scripts usually are packed — SkyUI, PapyrusUtil and most of the rest ship theirs in an archive. A plain file tool cannot open a `.bsa` member, so this sentence routes the more common case nowhere.

The skill already has the answer 60 lines down, for `.pex`: `housecarl_bsa_extract` with `archive=` and `dest=`, then read the extracted path. One clause here would close it.

.claude/skills/papyrus-reference/references/corpus-notes.md
```diff
@@ -0,0 +1,72 @@
...
+## Regenerating
+
+Regenerate from upstream, then apply the four post-passes below in order. All four are part of the
+regeneration, not one-off edits: skipping any of them puts the shipped tree back into a state the
+body's promises do not hold in.
+
+1. **Copy the hand-authored files back.** `silent-biters.md` and this file are outside the generated
+   set and are lost by a clean regeneration.
+2. **Drop function-less script files.** A generated file that carries no entry is reachable from
+   neither the body nor the index; all it can ever produce is a dead read. 32 such vanilla stubs
```
Avick3110: **"all it can ever produce is a dead read" understates what the stubs carried.**

Each dropped file was three lines, and the middle one was the class's extends edge — e.g. `vanilla/Book.md`: `**Source:** \`vanilla\` • **Extends:** \`Form\``.

Eleven of the 32 names still have an `skse/` counterpart carrying that edge. Twenty-one do not, and now appear in no file in the corpus at all: `Action`, `Activator`, `AssociationType`, `Class`, `Container`, `Door`, `EncounterZone`, `Explosion`, `Furniture`, `Hazard`, `Idle`, `ImpactDataSet`, `Key`, `LocationRefType`, `MiscObject`, `Projectile`, `Static`, `TalkingActivator`, `VoiceType`, `WordOfPower`, `WorldSpace`.

`SKILL.md:68` tells the model to resolve an unqualified call "up the calling script's `extends` chain". For a script extending `Container` or `Door` the corpus no longer records what the parent is. The drop still looks right — a dead lookup is a dead lookup — but this is a real loss the rationale does not cover, so it belongs either in the "Known coverage bound" section or in a kept header line.

.claude/skills/papyrus-reference/references/messagebox.md
Avick3110: Folded in f5ad39c7, taking the suggestion. `references/messagebox.md:14` now reads `| Global Functions | 8 | 18 |`.

Confirmed: the file carries 23 `### ` lines and 8 index rows; the other 15 are doc-comment prose (`### Example`, `### Gotcha!`, `### Button names`, …). Same defect as 77e4f10d one heading level down, as you say.

Scope check — I walked every table row in the tree against the index (span the row's section, count the rows whose `line_start` falls inside it): 87 tables, and `messagebox.md` was the only wrong count. `cube.md` carries 2 doc-comment `### ` lines too but is 45 lines, so it has no table.

c5 gains the entry-count clause you asked for: each row's count must equal the number of index rows whose `line_start` falls inside that section, so a doc-comment `### ` is never counted as an entry. `references/corpus-notes.md` pass 3 carries the same rule — "count entries, not `###` lines" — so a regeneration cannot put it back.

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in b2b17e89, taking the suggestion verbatim. `SKILL.md:117-118` now reads:

```
- `Game.GetForm` returns `None` for any FormID `>= 0x80000000` — the whole ESL range *and* load
  index `0x80+` — use `Game.GetFormEx`.
```

Confirmed against `references/silent-biters.md:14`, which states the `>= 0x80000000` bound and names both halves. You are right about which half a large order hits: the inline rule is what a session sees when the conditional read of `silent-biters.md` does not fire, and an order past 128 regular plugins reaches `0x80+` with nothing ESL-flagged involved.

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Left for Aaron — refuted here on the ruling, but it is your ruling to vary.

Both of your facts check out: nothing in `scripts/`, `src/` or `plugin/` reads the key, and this is the only skill in the tree carrying it today. Neither settles it.

**The wording is ruled.** `dev/DECISIONS.md:21`, 2026-09-07 — "Every skill carries one `compatibility` frontmatter line (rewrite ruling 7) naming the houseCARL MCP server and an MO2 instance, **same wording throughout**" — and `REWRITE.md:78-81` prints the exact line this branch ships. That a grep finds one occurrence is wave order: the ruling is a day old and this is the first of the twelve rewritten under it. A scoped variant here is a different line from the other eleven, which is the half of the ruling that is not about content.

**The walk-back is the ruling's own design, not a defect.** `REWRITE.md:84-89` anticipates exactly your reading — D2.35 graded the *absence* of the key as conforming because the lookup is offline, that reason expires when the Tier-2 check becomes a `housecarl_skse` call, and the document's answer is: carry the shared line, and have the body scope it "in one clause rather than promising a hard floor the corpus does not need". `SKILL.md:23-26` is that clause, added for the blind review's version of this finding. So the Overview correction is what the document asked for.

That said, changing ruling 7 is a call only you can make, and it is twelve skills wide, not one — so: leave it as ruled, or vary ruling 7 and I will scope the line in all twelve. I have not applied the suggestion.

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in 858cef3f. Reproduced on this branch's index: `grep -c '"name":"dispel"'` is 0, `grep -ic` is 2 (the vanilla and skse `ActiveMagicEffect#Dispel` rows), and the spaced pattern is 0 under `-i` as well.

`SKILL.md:226-229` now scopes it the way line 50 does:

```
- Grep the index case-insensitively with the compact token, `grep -i '"name":"Dispel"'` (a spaced
  `"name": "Dispel"` matches nothing at any casing, and a case-*exact* grep for `"name":"dispel"`
  matches nothing either — both read as absence; under `-i` the miscased pattern resolves, so a zero
  from it is a real miss).
```

That closes the loop you name: a zero from the prescribed grep enters bundled-or-warn rather than sending the reader back to re-grep.

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in 0941c2a9. Reproduced your numbers exactly by rendering both blocks for every `qualified` that carries one vanilla and one skse row: 835 pairs, 831 byte-identical, 4 differing, and only `ObjectReference#RemoveAllStolenItems` differs on required-ness (`akTransferTo` optional in vanilla, ✓ in skse). The other three are the doc-comment: `#GetAllItemsCount` drops it, `#SetContainerAllowStolenItems` drops it, `#IsContainerEmpty` replaces it with "added in 1.6.1126".

`SKILL.md:74-77` now reads:

> Almost every such pair renders the same block; where the two differ it is usually the doc-comment the `skse` block drops or replaces, and only rarely a parameter's required-ness. Read both, and call it a declaration conflict only when the signature differs, not when a sentence of prose does.

Same instruction — read both — without the three-in-four false conflict.

.claude/skills/papyrus-reference/SKILL.md
Avick3110: Folded in 7b1a17ed. `SKILL.md:28-32` now reads:

> Reading a modlist's actual `.psc` source is a file job (ask `housecarl_asset_status` which mod or BSA wins a `Scripts\...` path first, then read it with your own file tool — or, when the winner is inside an archive, which is where most framework scripts ship, `housecarl_bsa_extract` with `archive=` and `dest=` first and read the extracted path); compiling is `housecarl_compile_script`.

Checked against the surface: `housecarl_asset_status` reports "whether the asset is loose or in a BSA" and names every providing source including active BSAs, and `housecarl_bsa_extract` takes `archive=` (full path, required) and `dest=` (optional folder). Same two parameters the `.pex` step 60 lines down already names, so the clause adds no new tool to the body's five.

.claude/skills/papyrus-reference/references/corpus-notes.md
Avick3110: Folded in 13970099, taking the first of your two options.

Verified your split against the tree: of the 32 names dropped in 5574ca2f, 11 keep the `extends` edge on an `skse/` file (`Ammo`, `Apparatus`, `Book`, `ConstructibleObject`, `Flora`, `Outfit`, `Perk`, `Race`, `Shout`, `SoulGem`, `TextureSet`) and the other 21 are the list you name — they appear in no file in the corpus now.

`corpus-notes.md` pass 2 no longer says the stubs could only ever produce a dead read; it says a *lookup* could, and names the cost. The `## Known coverage bound` section is now `## Known coverage bounds` and leads with the `extends` edge: the 21 names, the fact that `SKILL.md` walks the `extends` chain for an unqualified call, and that closing it means keeping a stub's header line rather than dropping the file — a change to pass 2, not a re-run of it.

Kept out of the body: `SKILL.md`'s coverage bound is stated as one, and this bound is a maintainer fact about a regeneration pass rather than something a lookup can act on. Say the word if you want it in the body too, or the header lines kept instead.

PR #643, opened 2026-09-08
Rebuild the mutagen-reference skill on the 2.0 grammar

Avick3110:
Wave 2 of the skill rewrite: `mutagen-reference`, built to `dev/skill-validation/eval-2026-09/mutagen-reference/REWRITE.md` with `REWRITE-verify.md` beside it. The skill's measured value — the shard corpus, the index, and the one-line read discipline — is untouched; what changed is everything that had rotted around it. The body named four tools the server does not publish (`housecarl_set_field`, `housecarl_bulk_apply`, `housecarl_create_record`, `housecarl_read_record`), so an agent that trusted its write table called four nonexistent tools before finding the real ones. Addressing a field now names `housecarl_apply`'s `field_path` and the `op` verbs each cardinality accepts, with `housecarl_create` for new records, `housecarl_records` for the read, `housecarl_forward` plus `apply into=` for a whole record, and `op='CopyFrom'` with `from_source` for one field's value from a named plugin. Two things it taught that were wrong are gone: the Mutagen version pin nothing in the skill maintained (and the warning template's pointer that depended on it), and the claim that giving a parent a child record or dropping one is an open gap — `housecarl_create.records`' `parent` and `housecarl_remove.formids` both express it, and what survives is the narrower true rule. The scope sentence stops claiming the skill carries no write view while handing out write verbs; the description gains a third-person opening, an explicit schema-only exclusion clause and the ruled `compatibility` line; every shipped reference file is named once in the body with its own read condition, so the route body → index → shard no longer has an undeclared leg; the re-read rule is stated once instead of three times; worked examples go from one pair to three; and the write section falls from 37% of the body to 24%. `evals/eval_set.json` becomes `evals/evals.json` on the published field names, keeping all 21 cases (11 should-trigger, 10 near-miss negatives), each with checkable expectations, a without-skill arm on every trigger case, the packaged tree as the thing under test, and triggering and output quality as two separate result sets.

No issue is named as closing by the document, so there is no `Closes` line.

## Numbers

- `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` — 1793 passed, 0 failed.
- `housecarl-generator ci-all` — 128/128 probes passed.
- `scripts/build-plugin.ps1` — skill leak-check clean, 12 skills, 250 markdown files, both trees. Full package build clean.
- Body: 14,870 characters ≈ 4,130 tokens (was 19,496 ≈ 4.9–5.6k), 152 lines. `description` 531 characters; `name` + `description` 548. Section 8 is 3,497 characters, 24% of the body — above the document's ≈2,600 / ≈20% estimate, and the one budget I did not hit. Cutting further meant dropping either the owned-child rule or the index-versus-gap distinction between `SetAtIndex` and `InsertAtIndex`, both of which a write actually needs.

**Decided here, and why:**

1. **The four unpublished write verbs.** Kept in the verb column on the audit's evidence; no issue filed, and `housecarl_apply.ops` is not grown for them. The audit sentence the document weighs ("does match `housecarl_apply`'s `op` exactly") is now better evidence than it was: the surface was re-captured on 2026-09-08 at `e2990653`, and `housecarl_apply.ops` today publishes `from_source= (op='CopyFrom' only)`. That is the fifth verb the same sentence endorsed and the document declared dead on a 0-hit count against the older capture. So the sentence held where it could be checked, and the skill keeps the verb column it vouches for.
2. **`CopyFrom` (migration row 6, correction 2) is overtaken by that re-capture.** The rewrite would have replaced it with `from_source`; the live surface publishes both, as one op and its member. The body says so rather than teaching that a published verb is dead — a silently wrong answer is the one thing this skill exists to prevent. Flagging it because it is a departure from an approved document, on the strength of the source the brief names as authoritative.
3. **`compose`'s inner shape** stays on the skill's own authority (`compose:{type:…}` and `compose.sets` are what composing an element needs, and nothing else states them). **The CTDA index-mode operands go**: per §4 the note is reduced to the normalisation fact and "confirm it at the engine". No issue filed.
4. **#350** — the citation is dropped, as the document assumed; the narrower sentence carries what is still true.
5. **Invocation posture** — no `user-invocable`, no `arguments`. Frontmatter stays inside the spec's closed six, and D3.19 and D3.23 stand as permitted partials.

**Left for Aaron:**

6. **The name.** `mutagen-reference` still names the library rather than the task (D3.8). Unchanged here: renaming moves the folder, the identity and every inbound pointer at once.
7. **Haiku tier and isolation run** (D5.23, D5.24). Ruling 3 is silent on both; `evals.json` records the method it does have and does not invent a house rule for these two.
8. **Where a hand-authored row lands — this one blocked.** See below.

## What blocked: F16 has no home the tree will accept

FOLD-IN's settled answer to its Q7 (`dev/DECISIONS.md`, 2026-09-08) puts hand-authored rows in one hand-written notes file beside the shards, `mutagen-reference/references/hand-notes.md`, named with a one-hop pointer in this body. I wrote that file with F16's row in it (the KID trait `OneHandDagger` → `Weapon.Data.AnimationType`, verified through all three shards) and the generator refused it:

```
FAIL  hand-notes.md: present in the reference tree but not emitted
      -> the generator does not produce this file.
emit-match-guard: FAIL (1 shard(s) stale)
```

`emit-match-guard` (#351) holds the whole `references/` tree to what a fresh emit produces, so no hand-written file can sit there without the guard being taught about it. That is a design call — it is the first cornerstone with a probe behind it — so I did not work around it. The file and its pointer are backed out of this PR, and `REWRITE.md` §8's own instruction ("the rewrite does not write either row; it records where they land and stops") is what this PR follows. The choice is a guard that allows one declared hand-written file in that directory, or a different home for the row. Whichever it is, `kid-authoring`'s pointer sentence in its own wave depends on the answer.

**F11 needs nothing.** Its settled destination is the `FirstPersonFlags` rendering, and the check it asks the build session to make comes back positive: `src/housecarl-core/ReadEngine.cs:1040-1055` already decodes a `BodyTemplate.FirstPersonFlags` value into its slot numbers. Nothing was added.

## Permitted partials, with their reasons

D3.8 (the name — open question above), D3.19 and D3.23 (invocation posture, decided as the document assumed), D4.16 (no shard opens with a table of contents: a TOC is not expressible in JSONL, where every line is one self-contained record, and `index.jsonl` already discharges what a TOC is for; exact `line` addressing answers the partial-read hazard), D5.23 and D5.24 (left for Aaron above). D2.55 stays no source — the Codex compressed-bundle ceiling is unpublished.

## Two corrections to the document, both from the verifier

`REWRITE-verify.md`'s two overturns are folded: D4.54 is a partial the rewrite fixes, not a conforms it inherits, so the single re-read rule is written as a fix rather than preserved as-is; and N7 refuses only the parameter-description home for the write table, not a `references/` one — the table stays in the body on the document's own judgement, and N7 is not cited as the authority for it.

## Regrade

Five angles regraded first-hand against this branch, one grader per angle
(`dev/skill-validation/eval-2026-09/mutagen-reference/regrade/D1.md`–`D5.md`), assembled in
`REGRADE.md` beside them.

| Angle | conforms | partial | fails | n/a | criteria graded |
|---|---|---|---|---|---|
| D1 — discovery and packaging | 40 | 0 | 1 | 1 | 42 |
| D2 — frontmatter | 39 | 1 | 0 | 12 | 52 |
| D3 — loading and invocation control | 25 | 3 | 0 | 5 | 33 |
| D4 — body, supporting files, style | 42 | 2 | 0 | 23 | 67 |
| D5 — end-to-end and evals | 35 | 8 | 0 | 7 | 50 |
| **Total** | **181** | **14** | **1** | **48** | **244** |

**PASS.** The only fail on any angle is D1.28, release tagging — repository tags carry no
`housecarl--v{version}` form matching `plugin/.claude-plugin/plugin.json`'s `2.0.0-beta.1`. It is
packager-only, outside the skill's own files, and excluded by ruling 8. D2 and D3 have zero fails;
D4 and D5 have zero fails of any kind, so zero factual or dead-call fails — every tool and parameter
name the body uses resolves on the captured surface, and the four dead 1.x names are gone.

Five of the twenty-eight partials the document requires to clear did not: D5.20, D5.21, D5.22 and
D5.25 — `evals/evals.json` now specifies the right experiment and has not been run, so every result
field is `null` — and D5.49, no enforcement route named in the skill. D2.35 is one new partial, the
cost of ruling 7's `compatibility` line. None is a fail and none moves the verdict. Two graders also
confirmed independently that `CopyFrom` is live on the captured surface (`tools.md:709`), backing
open question 2 above.


.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -78,65 +101,59 @@ When the index has no match for a requested record type or signature:
...
-
-Two honesty boundaries the write tool enforces, worth stating when you compose against this schema:
+| `scalar` / `enum` / `formlink` / `value` | the dotted name — `ArmorRating`, `BasicStats.Damage` | `Set`; `Remove` to clear a nullable field | The value is coerced to the field's `t`: a number, an enum name from the referenced enum's `values`, or `XXXXXX:Plugin.esp` for a `formlink` |
+| `substruct` | descend to the leaf — `WorldModel.Male.Model.File`, not `WorldModel` | `Set` on the leaf | A write *at* the substruct itself is refused. Where the `ref` resolves to a `"kind":"record"` entry the field holds an **owned child record** (`Cell.Landscape`, `Worldspace.TopCell`): edit that child by its own FormID. A path *through* a parent reaches a sub-field only when the record being written already carries the child — a fresh override does not |
```
Avick3110: **"A write *at* the substruct itself is refused" is false, and the compression is what made it false.**

The prior wording was narrower â€” *"A direct `Set` on the substruct **itself** is refused"* â€” and the blanket restatement now contradicts the server on three counts:

- `CorpusRulebook.VerbLegality` (`Set` arm) returns `null` for a substruct leaf. `CorpusRulebook.cs:735` then routes a *composable* substruct leaf to `StructLeafLegality`: "an absent struct can be filled in ONE op" (`ObjectBounds`, `FaceParts`, `FaceMorph`).
- A *coercible* substruct takes a plain-value `Set` **at the field**. `TranslatedString` is one (`structs.jsonl:219`), so `Armor.Name` (`c:"substruct"`, `ref:"TranslatedString"`) is written as `{field_path:'Name', value:'â€¦'}` â€” which is verbatim the example in `housecarl_create`'s own `ops` description (`CreateTools.cs:139`).
- `CopyFromLegality` (`CorpusRulebook.cs:505`) refuses only owned child records and dicts. `housecarl_apply`'s `op=` description (`ApplyTools.cs:273`) says CopyFrom "copies a WHOLE field (scalar, formlink, modeled list, **sub-struct**)".
- `Remove` on a nullable substruct is admitted too (`VerbLegality`: `leaf.Nullable ? null : â€¦`) â€” that is how `Armor.Destructible` (`null:true`) is whole-cleared.

**Scenario.** Asked to rename a batch of armors, the reader block-reads ARMO line 9, sees `{"n":"Name","c":"substruct","ref":"TranslatedString"}`, and this row tells it a write at the field is refused. It either descends to `Name.String` or reports the rename as unsupported â€” when the published, documented form is a plain `Set` on `Name`. That is a confidently wrong answer produced *by the reference*, which is the one failure the skill exists to prevent.

`plugin/CHANGELOG.md:27` ships the same overbroad claim ("a write at the sub-struct field is refused").

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -78,65 +101,59 @@ When the index has no match for a requested record type or signature:
...
-
-Two honesty boundaries the write tool enforces, worth stating when you compose against this schema:
+| `scalar` / `enum` / `formlink` / `value` | the dotted name — `ArmorRating`, `BasicStats.Damage` | `Set`; `Remove` to clear a nullable field | The value is coerced to the field's `t`: a number, an enum name from the referenced enum's `values`, or `XXXXXX:Plugin.esp` for a `formlink` |
```
Avick3110: **The `[Flags]` enum bit verbs are missing from this row, and their absence is a silent data-loss path.**

`CorpusRulebook.VerbLegality` admits `Add` and `Remove` on any `[Flags]`-attributed enum leaf (`CorpusRulebook.cs:386` and `:398`, gated by `IsFlagsEnumLeaf` at `:353`): Add ORs the bit in, Remove ANDs it out, other bits preserved. `housecarl_apply`'s `op=` description (`ApplyTools.cs:273`) calls this "the way to flip one flag **WITHOUT** a Set silently dropping every bit you didn't mention; to turn all bits off, Set the field to '0'."

This row lists only `Set`, plus `Remove` described purely as a nullable-field clear â€” so the reference actively steers away from the correct verb.

**Scenario.** `ARMO.MajorFlags` reads `{"c":"enum","ref":"Armor.MajorFlag"}`. Asked to mark an armor non-playable, the reader follows this row and issues `op='Set', field_path='MajorFlags', value='NonPlayable'`, wiping every flag the record already carried. Nothing refuses it; the loss surfaces in game.

Same for `SPEL.Flags`, `NPC.Configuration.Flags`, `WEAP.Data.Flags`. (Pre-existing wording, but this PR's stated purpose is to make the table match the published surface.)

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -78,65 +101,59 @@ When the index has no match for a requested record type or signature:
...
-- **A composed record missing a required arm fails loud.** If you `compose` or `Add` a modeled element and leave a required polymorphic sub-field unset (a `Condition` composed without its `Data` arm, a leveled-list entry missing required data), the write is refused at serialize time with a **named** error and **nothing is written** — all-or-nothing, never a half-written patch. So when you compose an element, set its required sub-arm in the same `compose`.
+Brackets are for `list` and `dict` elements only. Taking one field's value from a plugin you name rather than from the load-order winner is `op='CopyFrom'` with `from_source`; copying a whole record is `housecarl_forward`, then `housecarl_apply` with `into` the same patch. New records are `housecarl_create`, and the read that precedes any of this is `housecarl_records`.
```
Avick3110: **The "a composed element must carry its required polymorphic arm" rule was dropped, and nothing replaces it.**

The prior body closed this section with two honesty boundaries. The PR description enumerates the intended cuts (the Mutagen version pin, the #350 citation, the CTDA index-mode operands) â€” this is not among them, and it is the one that costs a call.

It is true and load-bearing, and pre-flight cannot cover for it: `WriteEngine.cs:1659` notes that `Condition.Data` reads `Nullable=false` yet nullability "is NOT a 'required arm at serialize' signal", so the gate deliberately does not check it. The failure lands at `WriteEngine.cs:3817-3831` / `WritePatchBuilder.cs:1134`.

**Scenario.** The reader composes a CTDA â€” `op='Add'`, `field_path='Conditions'`, `compose:{type:'Condition', sets:[â€¦]}` â€” without setting the `Data` arm. Pre-flight passes. `dry_run` passes (`housecarl_apply`'s own `dry_run=` text: "Not a disk guarantee â€” a serialize/commit fault still surfaces only for real"). The write fails at serialize. This skill's own `evals.json` output-quality bar is "The write is composed first try".

One sentence restores it: when you compose an element, set its required sub-arm in the same `compose`.

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -1,72 +1,95 @@
 ---
 name: mutagen-reference
 description: >-
-  Look up the schema of any Skyrim record type — fields, types, writability, legal enum values — from the bundled by-construction reference. Use before reading, editing, or patching any record (ARMO, WEAP, MGEF, NPC_, …) and for any "what fields does X have" or xEdit-signature question. A type absent from the reference is a real coverage gap to surface, never something to guess.
+  Looks up the schema of any Skyrim record type — fields, types, cardinality, writability, and an enum field's legal values — from the by-construction reference bundled with this skill. Use before reading, editing, or patching any record (ARMO, WEAP, MGEF, NPC_, …) and for any "what fields does X have" or xEdit-signature question. Schema only, never instance values — the shape of a type, not what one record in one plugin holds. A type absent from the reference is a real library-coverage gap to surface, never something to guess.
```
Avick3110, line 4: **The description keeps, byte for byte, the clause the recorded eval miss blamed â€” while `evals.json` claims the rewrite fixed it.**

`evals/evals.json:251` records: *"the instance-value prompt (id 12 here) picked this skill: the description's 'reading' verb read as covering a plain value read with no edit intent. **The rewritten description answers it with an explicit schema-only clause.**"* The deleted `eval_set.json` carried the adjudicator's actual remedy: *"narrow the 'reading' verb in the description to reading a record *type's schema* rather than reading a record, so a value lookup no longer matches; that risks nothing on the recall side since all 11 recall rows keyed off field-name/type/writability/enum language, not the 'reading' clause."*

The clause is unchanged: `Use before reading, editing, or patching any record (ARMO, WEAP, MGEF, NPC_, â€¦)` is identical in the old and new description. A schema-only sentence was *added* elsewhere, but the phrase the judge quoted as its justification still sits in the imperative sentence that tells the model when to load.

**Scenario.** The conductor runs the pending fan-out; case 12 ("what's the armor rating value of Steel Armor in Skyrim.esm right now", `should_trigger: false`) picks this skill again on the same clause, and specificity lands at 9/10 for the same reason as 2026-07-28 â€” with the file asserting the cause was addressed.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,19 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The `mutagen-reference` skill is rebuilt against the 2.0 tool surface.** It named four tools that no longer
+  exist — `housecarl_set_field`, `housecarl_bulk_apply`, `housecarl_create_record` and `housecarl_read_record` —
+  so an edit composed from it called nothing. Addressing a field now names `housecarl_apply`'s `field_path` and
```
Avick3110: **"called nothing" is not what the server does with those four names.**

All four are live rows in `AliasTable.RetiredTools` â€” `housecarl_read_record` (`AliasTable.cs:192`), `housecarl_set_field` (`:220`), `housecarl_bulk_apply` (`:222`), `housecarl_create_record` (`:225`) â€” and `ToolCallShim.cs:27` answers a call naming an unregistered tool with that row's redirect, which spells out the successor's full call shape. The header comment on `RetiredTools` says exactly this: "turning the SDK's generic unknown-tool error into a one-hop redirect."

So the user-visible cost of the stale table was extra hops before the right call, not a dead one. The PR description gets this right ("called four nonexistent tools **before finding the real ones**"); the changelog line overstates it, and this file is the one users read.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: **This states the constraint and removes the only route past it.**

"A path *through* a parent reaches a sub-field only when the record being written already carries the child â€” a fresh override does not." True. But the old body at least named a remedy (wrongly â€” "an open gap (#350)"), and the rewrite deletes the claim without putting the real answer in its place.

The PR description and `plugin/CHANGELOG.md:25-26` both assert the route exists: "`housecarl_create`'s `parent` and `housecarl_remove`'s `formids` both express it." That is correct â€” `CreateTools.cs:142` says `parent` takes "an EXISTING parent's FormID 'XXXXXX:Plugin.esp'", and `:145` says `collection` names the single-child slot **by name**, citing "a cell's 'Landscape', a worldspace's 'TopCell'". Neither `parent` nor `collection` appears anywhere in the skill body.

**Scenario.** Asked to give a fresh CELL override a Landscape, the reader lands on this note, finds the write-at-the-field refused and the parent path unavailable, and reports it as not supported â€” while `housecarl_create` with `parent='<CELL formid>', collection='Landscape'` does it. The rewrite traded a wrong remedy for no remedy; the reader reaches the same dead end either way.

.claude/skills/mutagen-reference/evals/evals.json
```diff
@@ -0,0 +1,255 @@
+{
+  "skill_name": "mutagen-reference",
+  "gate": {
+    "recall_threshold": 0.8,
+    "specificity_threshold": 0.5
+  },
+  "method": {
+    "tree": "dist/housecarl/skills/mutagen-reference — every run loads the PACKAGED skill, never .claude/skills/, so a tool name or a pointer that does not ship is caught here (dev/DECISIONS.md, 2026-09-07, ruling 3)",
+    "arms": ["with_skill", "without_skill"],
+    "arm_rule": "every should_trigger case runs in both arms, each from a fresh context. An expectation that passes in both arms is not differentiating and is investigated rather than counted.",
```
Avick3110: **`arm_rule` is not executable against the triggering cases as `judging` defines them.**

`judging` (line 15) describes a menu-pick fan-out: "one fresh-context agent per prompt, shown the full skill menu plus NONE and asked which ONE to load." In the `without_skill` arm this skill is not in the menu, so the first expectation on every one of the 11 `should_trigger` cases â€” "The skill is selected" â€” cannot pass by construction. It is not a differentiating result; it is a tautology.

The rule then compounds it: "An expectation that passes in both arms is not differentiating and is investigated rather than counted" â€” which sends the conductor investigating nothing while the 11 vacuous runs still get run.

The two-arm design is real and valuable, but it belongs to `output_quality` (line 213), where it already sits with its own arms list and a genuine differentiator ("The `without_skill` arm fails at least one of the four above"). Scoping `arm_rule` to the `output_quality` set â€” and leaving the triggering set as the single-arm menu pick it actually is â€” would make the file runnable as written.

**Scenario.** The conductor picks this up to run the pending fan-out, executes 22 triggering runs instead of 11, and has to adjudicate 11 results that carry no information.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded in bd56b5a8. Confirmed at the rulebook: `VerbLegality` returns `null` for a substruct `Set` and `CorpusRulebook.cs:735` routes a composable leaf to `StructLeafLegality` ("an absent struct can be filled in ONE op"); a coercible one (`TranslatedString`) takes the plain-value `Set` below it; `Remove` is admitted on a nullable leaf; `CopyFromLegality` refuses only owned child records and dicts. The row now reads `Set` on the leaf, and at the field itself `Set` (by `compose`, or a plain value where the struct coerces), `Remove` on a nullable one, and `CopyFrom` — with the refusal scoped to a field holding an owned child record, which is the narrow statement `REWRITE.md` §7 correction 3 says survives. `plugin/CHANGELOG.md` is narrowed the same way in 7aa6cfa6.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded in bd56b5a8. `IsFlagsEnumLeaf` (`CorpusRulebook.cs:353`) gates `Add` at `:386` and `Remove` at `:398` on any `[Flags]` enum leaf, and the scalar row was steering a one-flag edit into a whole-value `Set`. The row now carries the bit verbs and says what each does: `Set` writes the whole value and drops every bit you did not name, `Add` sets one bit, `Remove` clears one. One honest limit stated with it — the generated block does not mark which enums carry bits, so the row names the shape (`MajorFlags`, `Flags`, an entry listing independent bits) and says pre-flight refuses a bit verb on a single-value enum by name.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded in bd56b5a8, and the document agrees: `REWRITE.md` §8's F46 fold-out deletes the two fail-loud boundaries but says "the composition instruction that survives — set an element's required sub-arm in the same `compose` — stays in section 8 as reference, not as a claim about tool behaviour." The rewrite dropped it with the boundaries. Restored as one sentence after the table, stated as composition rather than as what the server does when it refuses: "When you compose an element, set its required sub-arm in the same `compose` — a `Condition` carries its `Data` arm."

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Refuted: the rewrite document settles this the other way. `REWRITE.md` §3 gives the `description` verbatim as its final text, 531 characters, with the trigger clause "Use before reading, editing, or patching any record (ARMO, WEAP, MGEF, NPC_, …)" kept and listed among "trigger terms kept and still front-loaded"; the fix it chooses for the boundary is the added exclusion clause — "An explicit exclusion clause … moving the boundary out of the body into the only text a host sees at decision time". §9 names the same clause as the answer to this exact case: `misses[0]`, "what's the armor rating value of Steel Armor in Skyrim.esm right now", "is exactly the near-miss the description's new exclusion clause is aimed at". So the adjudicator's 2026-07-28 remedy (narrow the "reading" verb) is not the remedy the approved document took, and `evals.json:251` states what the document states rather than a fix that was not made. The description is a rewrite-document decision, not a build-session one; changing it here would depart from an approved document. Case 12 stays live in the pending fan-out and will say whether the exclusion clause carries it — if it misses again on the same clause, that is the evidence for narrowing the verb, and it is Aaron's call.

plugin/CHANGELOG.md
Avick3110: Folded in 7aa6cfa6. All four are `AliasTable.RetiredTools` rows and `ToolCallShim.cs:27` answers an unregistered name with that row's redirect, so the cost was a spent call, not a dead one. The line now reads: "so an edit composed from it spent its first call on a retired name, which the server answers with a redirect to the successor rather than with the edit."

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded in bd56b5a8. The body stated the dead end and named no way past it, while the changelog and `REWRITE.md` §6 row 10 both assert the route exists — a reader following the row reports "not supported" for something the surface does. The owned-child clause now ends: "edit that child by its own FormID, and author one the parent lacks with `housecarl_create`'s `parent` and `collection` (the singular slot's own name)". Both are members of `housecarl_create.records` on the captured surface (`tools.md:946`), and the narrow refusal `REWRITE.md` §7 correction 3 keeps is unchanged around it.

.claude/skills/mutagen-reference/evals/evals.json
Avick3110: Half folded in 0384223e; the proposed remedy is refuted. The incoherence is real — with the skill out of the without-skill menu, "The skill is selected" has nothing to score in that arm — but scoping `arm_rule` to `output_quality` would drop the without-skill arm from the trigger cases, which ruling 3 requires of every skill: "every trigger eval has a without-skill arm and runs from the packaged tree" (`dev/DECISIONS.md`, 2026-09-07), restated in `REWRITE.md` §9 as "carry a without-skill arm on every trigger case". So the arm stays and the rule says what it scores: selection is scored in the with-skill arm only, the without-skill arm scores the case's remaining expectations, and the pass-in-both-arms clause applies to those. No vacuous adjudications, no runs dropped.

Avick3110:
Review folded: six of seven findings, one refuted, each answered in its own thread.

Folded — bd56b5a8 (`SKILL.md`), 7aa6cfa6 (`plugin/CHANGELOG.md`), 0384223e (`evals/evals.json`):

- The substruct row's blanket "a write at the substruct itself is refused" was false and would have made the reference give a confidently wrong answer on `Armor.Name`. Scoped to a field holding an owned child record, with the forms admitted at the field named.
- The `[Flags]` bit verbs are back in the scalar row — `Add` sets one bit, `Remove` clears one, `Set` drops every bit not named — with the honest note that the block does not mark which enums carry bits.
- The composition instruction `REWRITE.md` §8 says survives F46 ("set an element's required sub-arm in the same `compose`") is restored, as reference rather than as a claim about tool behaviour.
- The owned-child clause names the route past it: `housecarl_create`'s `parent` and `collection`.
- The changelog no longer says a retired tool name "called nothing" — `AliasTable.RetiredTools` answers one with a redirect to its successor.
- `evals.json`'s `arm_rule` now says what the without-skill arm scores on a trigger case, so the selection expectation is not adjudicated where it cannot pass. The arm itself stays on every trigger case, which ruling 3 requires.

Refuted: narrowing the description's "reading" clause. `REWRITE.md` §3 fixes that text verbatim and §9 names the new exclusion clause as the answer to case 12; the pending fan-out is what settles whether it carries.

Numbers after the fold: body 15,861 characters (was 14,870), 153 lines; section 8 is 4,422 characters, 28% of the body — further above the document's ≈20% estimate, and every character of the rise is a correction to a row that was wrong. `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` — 1793 passed, 0 failed. `ci-all` — 128/128 probes. `scripts/build-plugin.ps1` — clean, leak-check clean, 12 skills, 250 markdown files.

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -78,65 +101,59 @@ When the index has no match for a requested record type or signature:
...
+| `substruct` | descend to the leaf — `WorldModel.Male.Model.File`, not `WorldModel` | `Set` on the leaf; at the field itself `Set`, `Remove` on a nullable one, `CopyFrom` | Setting the field itself fills an absent struct in one op — `compose:{type:'<Struct>', …}` for one built from parts (`ObjectBounds`, `FaceParts`), a plain value where the struct coerces from one (`Name`, a `TranslatedString`). Where the `ref` resolves to a `"kind":"record"` entry the field holds an **owned child record** (`Cell.Landscape`, `Worldspace.TopCell`): every write at the field is refused — edit that child by its own FormID, and author one the parent lacks with `housecarl_create`'s `parent` and `collection` (the singular slot's own name). A path *through* a parent reaches a sub-field only when the record being written already carries the child — a fresh override does not |
```
Avick3110: **The `Set`-at-the-field claim drops the refusal case, and the row's own example is one.**

The verbs column now says `Set` works at the substruct field itself, unconditionally. It doesn't for a substruct that is neither composable nor coercible — and `WorldModel`, the path example in the very same cell, is exactly that: `{"n":"WorldModel",...,"ref":"GenderedItem<ArmorModel>"}`.

`CorpusRulebook` line 735 gates the compose path on `SchemaClassifier.IsComposableSubstructLeaf`, which returns false for `GenderedItem<T>` (`WriteEngine.IsPlainComposableStruct` excludes it — no paramless ctor). `CanCoerce` is false too, so it falls to `CoercibilityReject`:

> `'WorldModel' is a GenderedItem<ArmorModel> substruct — a direct Set isn't supported; navigate into it and Set a sub-field.`

The old row stated that refusal outright ("A direct `Set` on the substruct *itself* is refused"). The rewrite is right that a *composable* or *coercible* substruct takes a `Set` at the field, but it now teaches that as the whole rule, so an agent composing against a gendered field spends its first call on a refusal.

---

**Second, smaller thing in the same cell:** `compose:{type:'<Struct>', …}` is illustrated with "(`ObjectBounds`, `FaceParts`)". `ObjectBounds` is fine — field and struct share a name. `FaceParts` is the *field* name; its `ref` is `NpcFaceParts`, and the index carries no `"name":"FaceParts"` entry at all. `compose:{type:'FaceParts'}` is refused by the struct-spec check: `Compose type 'FaceParts' does not match 'FaceParts' struct type 'NpcFaceParts'.`

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -78,65 +101,59 @@ When the index has no match for a requested record type or signature:
...
+| `list` | `[N]` mid-path (`Effects[0].Data.Magnitude`); at the leaf, name the list itself and pass the index as `key` | `Add`, `Remove`, `SetAtIndex`, `InsertAtIndex`, `ReplaceAll` — not `Set` | Add a modeled element with `compose:{type:'<ElementType>', …}`; a coercible-element list takes plain values. `Add` appends and takes no `key`; `SetAtIndex` overwrites at `key` (`0`..`count-1`); `InsertAtIndex` inserts at a gap (`0`..`count`) and shifts the rest right, which is what a CTDA `Or` chain needs |
+| `dict` | `[key]` mid-path; at the leaf, the verb with `key` | `Set` (with `key`), `Add`, `Remove`, `Merge`, `ReplaceAll` | `Merge` and `ReplaceAll` take an `entries` key-to-value map; the block's `key` gives the key type |
+| `polymorphic` as a **list element** | step into the element and name the field on its concrete arm — `Scripts[0].Properties[0].Object` | the `list` verbs | Each element is a concrete arm of the modeled base, resolved at apply time. `Add` one with `compose:{type:'<arm>', …}` |
```
Avick3110: **`Scripts[0].Properties[0].Object` is not a path any record accepts** — it's missing its root.

No record type in the bundled reference has a top-level `Scripts` field (a scan of `records.jsonl` returns zero). It lives under `VirtualMachineAdapter`, and `housecarl_apply`'s own `compose` description spells the rooted form: `field_path='VirtualMachineAdapter.Scripts[0].Properties'`.

An agent copying this cell verbatim gets `FindField`'s "no such field on 'Npc'" and has to discover the prefix by trial — which is the thing this section exists to remove. The old row had the same gap, but since the row is being rewritten anyway it's cheap to root it.

```suggestion
| `polymorphic` as a **list element** | step into the element and name the field on its concrete arm — `VirtualMachineAdapter.Scripts[0].Properties[0].Object` | the `list` verbs | Each element is a concrete arm of the modeled base, resolved at apply time. `Add` one with `compose:{type:'<arm>', …}` |
```

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -78,65 +101,59 @@ When the index has no match for a requested record type or signature:
...
+| `list` | `[N]` mid-path (`Effects[0].Data.Magnitude`); at the leaf, name the list itself and pass the index as `key` | `Add`, `Remove`, `SetAtIndex`, `InsertAtIndex`, `ReplaceAll` — not `Set` | Add a modeled element with `compose:{type:'<ElementType>', …}`; a coercible-element list takes plain values. `Add` appends and takes no `key`; `SetAtIndex` overwrites at `key` (`0`..`count-1`); `InsertAtIndex` inserts at a gap (`0`..`count`) and shifts the rest right, which is what a CTDA `Or` chain needs |
+| `dict` | `[key]` mid-path; at the leaf, the verb with `key` | `Set` (with `key`), `Add`, `Remove`, `Merge`, `ReplaceAll` | `Merge` and `ReplaceAll` take an `entries` key-to-value map; the block's `key` gives the key type |
```
Avick3110: **`Merge` / `ReplaceAll` are not on every dict.** `WriteVerbs.On` adds them only when the element is coercible:

```csharp
// ReplaceAll/Merge carry their elements in entries=, a plain key-to-value shape with no build-from-parts
// form — so on a MODELED-element dict they do not exist yet, and naming them would offer a verb that fails.
if (!composed) { dict.Add(ReplaceAll…); dict.Add(Merge…); }
```

The corpus has exactly one modeled-element dict — `Package.Data`, `elemRef: APackageData` — so on that field the legal set is `Set`, `Add`, `Remove` only. Narrow, but the row as written offers a verb the gate refuses by name, which is the failure mode this section is meant to close. A clause on the Notes cell ("`Merge` / `ReplaceAll` only where the element is coercible — a modeled-element dict takes `Set`/`Add`/`Remove`") covers it.

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -78,65 +101,59 @@ When the index has no match for a requested record type or signature:
...
-- **A composed record missing a required arm fails loud.** If you `compose` or `Add` a modeled element and leave a required polymorphic sub-field unset (a `Condition` composed without its `Data` arm, a leveled-list entry missing required data), the write is refused at serialize time with a **named** error and **nothing is written** — all-or-nothing, never a half-written patch. So when you compose an element, set its required sub-arm in the same `compose`.
+Brackets are for `list` and `dict` elements only. When you compose an element, set its required sub-arm in the same `compose` — a `Condition` carries its `Data` arm. Taking one field's value from a plugin you name rather than from the load-order winner is `op='CopyFrom'` with `from_source`; copying a whole record is `housecarl_forward`, then `housecarl_apply` with `into` the same patch. New records are `housecarl_create`, and the read that precedes any of this is `housecarl_records`.
```
Avick3110: **"Brackets are for `list` and `dict` elements only" is wrong mid-path** — and so is the Common-mistakes bullet at line 153 ("a bracket on either is refused").

`CorpusRulebook`'s mid-path hop has an explicit branch above the collection branch:

> Gendered field ([0]=male / [1]=female): a substruct whose TypeRef is `GenderedItem<T>`. The named arms (.Male/.Female) descend as plain hops; **[0]/[1] is the render-matching navigable alias.**

So `WorldModel[0].Model.File` is a legal `field_path` — a bad key there is refused *by naming the two legal indices*, not by rejecting the bracket. The claim is true at the **leaf** (`Set WorldModel[0]` is refused, pointed at `.Male`/`.Female`), and that leaf refusal is probably what the sentence is remembering; but as written it covers mid-path too, where it's false.

The practical cost: the renderer *shows* `[0]`/`[1]`, so an agent will meet these paths in a read and, on this skill's authority, rewrite a working one — or tell a user their correct path is refused. Scoping the sentence to the leaf ("a bracket at the **leaf** is refused; mid-path, a gendered substruct takes `[0]`/`[1]` as an alias for `.Male`/`.Female`") fixes both spots.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded — 2de181e0, `SKILL.md:135`.

Both halves verified. `CorpusRulebook.cs:735` routes a `Set` to the compose path only through `SchemaClassifier.IsComposableSubstructLeaf`, which at `SchemaClassifier.cs:150-155` defers to `WriteEngine.IsPlainComposableStruct` and so excludes `GenderedItem<T>`; with `CanCoerce` false as well, `WorldModel` falls to `CoercibilityReject` (`CorpusRulebook.cs:1070-1077`) and its "navigate into it and Set a sub-field" refusal. The verbs cell now reads `Set` at the field "where the struct composes or coerces (Notes)", and the Notes name the third case and its way in: a `GenderedItem<T>` (`WorldModel`) is written at `.Male` / `.Female`, or at `[0]` / `[1]`.

The compose type is the field's `ref`, not its name: `CorpusRulebook.cs:998` matches the spec type against the leaf's TypeRef, and `records.jsonl` carries `{"n":"FaceParts",...,"ref":"NpcFaceParts"}` with no `"name":"FaceParts"` block anywhere in the tree. The cell now says `<Struct>` is the field's own `ref` and not its name — `ObjectBounds` is both, the `FaceParts` field composes as `NpcFaceParts`.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded — bad88c9a, `SKILL.md:137`.

`WriteVerbs.On` adds `ReplaceAll` and `Merge` under `if (!composed)` at `WriteVerbs.cs:146-152`, so a modeled-element dict is offered `Set` / `Add` / `Remove` only. The corpus has exactly the one you name: a scan of `records.jsonl` / `structs.jsonl` / `polymorphic.jsonl` returns eight `"c":"dict"` fields, and `Package.Data` (`elemRef: APackageData`) is the only one with a modeled element.

The verbs cell now reads `Set` (with `key`), `Add`, `Remove`; `Merge`, `ReplaceAll` where the element coerces, and the Notes cell says why and names the one field it excludes.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded, as suggested — 4534666f, `SKILL.md:138`.

Confirmed: no record block in `references/records.jsonl` carries a top-level `Scripts` field (zero hits over every `kind:"record"` entry). The rooted form is the one `housecarl_apply`'s own `compose` description spells at `ApplyTools.cs:288` (`field_path='VirtualMachineAdapter.Scripts[0].Properties'`), and the cell now reads `VirtualMachineAdapter.Scripts[0].Properties[0].Object`.

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Folded in both spots — 6d7850ee, `SKILL.md:141` and `SKILL.md:153`.

`CorpusRulebook.cs:208-226` takes the gendered branch above the collection branch on a mid-path hop: a `substruct` whose TypeRef starts `GenderedItem<` accepts `[0]` / `[1]` and descends to the arm type, and a bad key there is refused by naming the two legal indices. The leaf is where the bracket is genuinely refused, at `CorpusRulebook.cs:262-266`, which points at `.Male` / `.Female`.

The paragraph now scopes the rule to the leaf and names the mid-path alias with a worked path (`WorldModel[0].Model.File`), noting that this is how a read renders it; the Common-mistakes bullet carries the same scoping.

Avick3110:
Aaron's review folded: all four findings, each answered in its own thread. No refutals.

- 2de181e0 — the substruct row claimed a `Set` at the field unconditionally. A substruct that neither composes nor coerces refuses it (`CorpusRulebook.cs:1070`, reached because `SchemaClassifier.cs:150` excludes `GenderedItem<T>`), and `WorldModel` — the row's own path example — is one. The row now scopes the verb and names the way in. Same cell: `compose:{type:…}` takes the field's `ref`, not its name, so the `FaceParts` example reads `NpcFaceParts`.
- bad88c9a — `Merge` / `ReplaceAll` are added only for a coercible element (`WriteVerbs.cs:146`), so the dict row no longer offers them on the corpus's one modeled-element dict, `Package.Data`.
- 4534666f — the script-property path is rooted: no record has a top-level `Scripts` field; it is `VirtualMachineAdapter.Scripts[0].Properties[0].Object`.
- 6d7850ee — the bracket rule is scoped to the leaf. Mid-path, a gendered substruct takes `[0]` / `[1]` as the alias for `.Male` / `.Female` (`CorpusRulebook.cs:208`), which is how a read renders it; only the leaf bracket is refused (`:262`). The Common-mistakes bullet carries the same scoping.

Still open for you, unchanged from the description: the skill's name (D3.8), the Haiku tier and isolation run (D5.23, D5.24), and where a hand-authored row lands — `emit-match-guard` refuses a hand-written file under `references/`.

After the fold: `scripts/build-plugin.ps1` clean (skill leak-check clean, 12 skills, 250 markdown files, both trees; assembled-tree leak-check clean), 1858 tests passed / 0 failed, `ci-all` 128/128.

Issue #584, opened 2026-09-05
[GAP] asset_status has no batch or file-sink mode; a whole-order facegen pairing sweep has to be rebuilt off-tool


Avick3110:
Seen 2026-09-05 in both facegen-diagnostics audit runs on the 2.0.0-beta.1 install against the ARR 2.0 order (3801 plugins, 2,706 loose facegeom and 2,996 loose facetint files).

**What happened.** Both agents (one without the skill, one with it) needed the same answer: for every NPC with loose FaceGen, which mod wins the head mesh and which wins the tint, so the two can be compared to each other and to the record winner. `housecarl_asset_status` answers that per path, in text, at a size that breaks the character budget past a few dozen paths. Both agents therefore walked the MO2 mods folder and modlist.txt themselves in Python to reconstruct VFS precedence for ~5,700 paths, then used `asset_status` only to spot-check. Each spent 38 of 40 calls and about half an hour. The with-skill run's finding (62 NPCs whose winning mesh names a tint that exists nowhere, 57 in Vigilant) came entirely from that off-tool map.

**Why it matters.** The reconstruction ignores BSA providers, so it can over-report "absent"; the agents knew that and re-verified samples, but the sweep itself was never a tool answer. Facegen pairing is the dark-face diagnosis, and it is a set question over the whole order, not a per-path one.

**What would close it.** `asset_status` taking a set (paths, or FormIDs with the facegen pair derived from each) with `to_file=` like `records`, returning per path the winning provider, the provider kind (loose or BSA), and for facegen the paired path's winner beside it. The set-valued and file-sink conventions already exist on `records`; this is the same axis on the asset lane.

Evidence: `dev/skill-validation/audit-2026-09/facegen-diagnostics/{control,with-skill}/log.md` (private).

Not milestoned; Aaron decides.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #644, opened 2026-09-08
Rebuild facegen-diagnostics on the pair diff and the 2.0 tools

Avick3110:
Rebuilds the `facegen-diagnostics` skill against the 2.0 tool surface, executing the approved rewrite document (`dev/skill-validation/eval-2026-09/facegen-diagnostics/REWRITE.md`) with its verification pass folded in. The flow now opens on the mesh-versus-tint pair diff — resolve the head `.nif` and the face `.dds` for one NPC in a single `housecarl_asset_status` call and branch on the two winners — and the record comparison it used to lead with is the second step, reached when the pair is clean. That re-ordering is the point of the rebuild: on the audited order the pair axis held every finding while the record axis was clean everywhere, and the sentence that carried the pair rule used to sit in a Notes bullet past the point a host re-reads. Five tool names the old body drove the whole flow on do not exist on this server; every call is migrated to `housecarl_records`, `housecarl_apply`, `housecarl_place` and `housecarl_forward`, including the one inside the two-precedence passage the fold ledger keeps in the body. Eight factual corrections land with it: the console-FormID "STOP" becomes "the read tools take the runtime form, a console click still selects the placed reference, and `housecarl_apply` refuses it"; "enable + sort" becomes "enable", since a new mod folder lands last; the sweep language no longer implies `housecarl_check` has a facegen finding family, because it does not; and the twelve "it can now…" currency claims are gone. The body is cut from 24,475 characters to 15,285 (~4,100 tokens), with the capability inventory deleted, the batch flow ceded to `housecarl:bulk-record-jobs`, and the mesh-side repairs carved into a second bundled reference, `references/mesh-repairs.md`. `evals/eval_set.json` becomes `evals/evals.json` on the published fields, with a without-skill arm, both run configurations, a named failure mode per gate dimension, and three new cases: two lead-axis near-misses that must not trigger, and one boundary case that cedes to `housecarl:npc-appearance-copy`.

No issue closes with this: the rewrite document names none, and `#584` (no file sink and no formid/kind pair mode on `housecarl_asset_status`) stays open — the body names it where the whole-order sweep is described rather than pretending the join is one call.

## The verifier's overturned rows

All four are folded as the verification pass says. **G0** — the missing migration row: `SKILL.md:57` named `housecarl_read_record` inside the passage the fold ledger keeps, so the kept frame now names `housecarl_records`. **D11** — the point-of-view row: the D2 checklist's own conform test is met by the one-word turn, so the description reads "Diagnoses and repairs…" while the imperative "Use for…" trigger clause is untouched; D2.26 and D4.61 close instead of being carried, which takes the carried rows to seven. **G20** and **J5** are a site column and a wording claim inside the document itself, with no consequence for what ships. The five optional precision notes are honoured too — the colon in the description is safe because it sits inside the folded `>-` scalar, and D4.13 is graded against the shipped tree, where `evals/` does not exist.

- **Q1, does `housecarl_compact_plugin` carry facegen files** — **left for Aaron.** The server instructions say it does and the tool's own description says nothing about assets. Nothing here asserts either way, and no sentence was added to the body on the strength of the unreconciled claim.
- **Q2, does #584 stay as written** — **left for Aaron.** The body names the issue and states exactly the half that still forces a manual join, so it reads correctly whether the issue keeps its scope or is narrowed.
- **Q3, the console-FormID rewrite** — **decided as follows.** The proposed replacement is in: read the runtime form, confirm the record it resolved to is the NPC the user means, then write with the `XXXXXX:Plugin.esp` form. The harder STOP is factually wrong on this surface, so it is not kept.
- **Q4, one reference file or two** — **decided as follows.** Two. `references/mesh-repairs.md` is the second, named once from the body under its own read condition.
- **Q5, what counts as beating the control** — **decided as the document proposes, and still Aaron's to change.** `evals/evals.json` records the gate as both fault families in one run. The cheaper alternative — gate on the pair family and run the record family as a second pass — is one edit to that block if he prefers it.
- **Q6, does the in-place `nif_set` lane stay in the body** — **decided as follows.** It does not. The body carries one sentence routing mesh-side repairs to the reference, and both lanes — the default new-mod folder and the consent-gated in-place one — are documented there.

## Judgement calls

- The rewrite document's migration table covers `SKILL.md` only, but the bundled cause-and-fix reference carried the same five dead tool names and the same currency framing. Since D5.13 and D4.43 are graded on the skill rather than on the body alone, the reference is migrated too, and its console-FormID paragraph is corrected to match correction 4.
- The cause-and-fix reference points at `mesh-repairs.md` by name where its old embedded-texture section used to be. That is a hop between two references, not a body pointer, so ruling 2's one-hop rule is untouched and the packager's pointer scan resolves it.
- `walk=`'s sub-parameters are not in `surface/tools.md`, so the template-chain read is written in prose naming only `walk=` and `project={"form":"chain"}` rather than inventing a literal shape.
- `housecarl:bulk-record-jobs` is named once, in the section that cedes to it, rather than repeated in the closing pointer table.
- The body lands at 15,285 characters, ~4,131 tokens, against the document's ~4,050 budget and the ~5,000 ceiling.

Package built with `scripts/build-plugin.ps1` (skill leak-check clean, 12 skills, 251 markdown files, both trees); `dotnet test` 1793 passed, 0 failed; generator `ci-all` 128/128.

## Regrade

Regraded against `checklists/D1.md`–`D5.md` on this branch at `48803b9c`. Full write-up:
`dev/skill-validation/eval-2026-09/facegen-diagnostics/REGRADE.md`.

|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 37 | 1 | 0 | 4 |
| D3 — loading decision and invocation control | 33 | 26 | 1 | 0 | 6 |
| D4 — body, supporting files, instruction style | 64 | 39 | 2 | 1 | 22 |
| D5 — validation and standing context | 50 | 35 | 8 | 1 | 6 |
| **Total** | **241** | **177** | **12** | **2** | **50** |

**FAIL** against the gate (zero fails on D2 and D3; zero factual-or-dead-call fails on D4 and D5;
D1 clean except packager-only, with D1.28 excluded by ruling 8). D1, D2 and D3 carry no fail —
D1's only partial is D1.28, which the gate excludes. D4's only fail is D4.42, the bare MCP tool
names, typed **other** and ruled by ruling 1, so D4 meets its threshold too. The gate breaks on
**D5.13**, typed **factual-or-dead-call**: `references/facegen-causes-and-fixes.md:204` still
names `housecarl_forward_record`, which is not on the 31-tool surface (the shipped tool is
`housecarl_forward`, taking `source=`), and `:198-199` and `:326` point at "Step 5" and "Step 3",
which the rebuilt body does not have. `SKILL.md` itself is clean on both counts; D4.43 is partial
on the same two lines. Fixing them closes D5.13 and D4.43 together and takes the regrade to PASS.

**Fixed in `d9f6efd8`.** `references/facegen-causes-and-fixes.md:204` now names `housecarl_forward`
with the real call shape (`formids=` the NPC, `source=` the appearance plugin, `into=` the patch);
`:198-199` points at SKILL.md §7, Step 4 (verify) and `:326` at SKILL.md §3, Step 1 and §5, Step 2.
A sweep of the whole skill tree — `SKILL.md`, both references, `evals/evals.json` — found no other
1.x name and no `housecarl_` token off the 31-tool index; the nine tools it names all resolve.
Package rebuilt (skill leak-check clean), `dotnet test` 1793 passed / 0 failed, `ci-all` 128/128.

## Regrade

Re-run on this branch at `d9f6efd8`, after the fix above. This table supersedes the `48803b9c` one
in the section above. Full write-up:
`dev/skill-validation/eval-2026-09/facegen-diagnostics/REGRADE.md`.

|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 37 | 1 | 0 | 4 |
| D3 — loading and invocation | 33 | 27 | 0 | 0 | 6 |
| D4 — body, supporting files, instruction style | 64 | 39 | 2 | 1 | 22 |
| D5 — validation and standing context | 50 | 37 | 7 | 0 | 6 |
| **Total** | **241** | **180** | **10** | **1** | **50** |

**PASS** against the gate (zero fails on D2 and D3; zero factual-or-dead-call fails on D4 and D5;
D1 clean except packager-only, with D1.28 excluded by ruling 8). D1, D2, D3 and D5 carry no fail
at all, and the factual-or-dead-call count is zero on all five angles — every tool name, parameter
and call shape in the shipped tree resolves on the surface, so D5.13 and D4.43 are now conforms.
The one fail left on the whole regrade is **D4.42**, the bare MCP tool names, typed **other** and
ruled by ruling 1, which the D4 threshold does not catch. D1's only partial is D1.28, excluded by
ruling 8. Of the other nine partials, four are carried by §9a with their reasons (D4.25, D5.23,
D5.40, D5.49), four are eval-file rows that close on the first run of the arms `evals.json`
already defines or on a one-word key rename (D5.16, D5.21, D5.25, D5.27), and one is the
enable+sort wording still in `references/facegen-causes-and-fixes.md:181` (D4.48).


.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,251 @@
...
+  patch="Facegen Fix")
+```
+
+`bundle=` is the field set — which paths form an appearance set is knowledge this skill carries, not
+something the tool owns — and `assignments=` pairs each target with the record it takes them from.
+When the whole record from that plugin is wanted instead, `housecarl_forward` with `source=` copies it
+verbatim.
+
+Then tell the user to **enable** the new mod in MO2. A new mod folder lands last in the priority order,
+so enabling is the step that makes the write win; re-resolve the path (§7) rather than asserting it.
```
Avick3110, line 190: **The sort step is dropped, and the claim that replaces it is not true of the contended case.**

`PlaceTools.EnableAndSort` emits, whenever a placed path already has a winner: `Enable the mod '<X>' in MO2 and SORT it (left pane) ABOVE the current winner(s) listed above. Only then does the placed copy win.` The tool's own description says the same (`you must ENABLE the new mod in MO2 and SORT it above the current winner`), `WriteSentences.cs:117` says `enable + sort it in MO2 to use the patch` for the nif_set lane, and `housecarl_apply`'s readback text says `the patch wins nothing until enabled + sorted in MO2`. Same claim at `references/mesh-repairs.md:89-90`.

Nothing in this repo backs "a new mod folder lands last in the priority order". `Mo2LoadOrder.CandidateFolders` ranks a freshly-written houseCARL folder as UNLISTED — below every enabled *and* disabled mod, above only game Data — until MO2 refreshes, and where MO2 then files it is not something houseCARL controls.

Failure scenario: Cause E, which §3's branch table routes straight here — a stale loose copy in MO2's **overwrite** folder wins the `.dds`. `housecarl_place` writes a new mod folder; the user follows this line, enables it, and stops. Overwrite sits at the top of MO2's VFS, so no mod-folder priority can beat it — the face stays dark, and §10's "place the pair from Bijin, enable it, then run the in-game check" reads as a completed fix. The body also contradicts itself: §1 says "among loose the higher-priority mod (then overwrite) wins", and §7's first verify row still says the tool "names the winner to rank above" — a rank-above step §6 no longer tells anyone to take.

The fabricated sample output at :153 (`ENABLE this mod in MO2 — the placed copies do not win on write.`) also teaches a tool response the tool does not emit.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,251 @@
 ---
 name: facegen-diagnostics
+compatibility: Requires the houseCARL MCP server and a configured MO2 instance.
 description: >-
-  Diagnose and repair the dark / grey / black-face NPC bug in Skyrim SE by comparing the record winner against the VFS facegen file winner. Use for any discolored face, neck seam, "fine in xEdit but wrong in game", NPCs gone dark after an ESL-compaction or merge, or any FaceGen / facegeom / facetint mention. Load before judging any face bug — record-vs-file precedence decides the fix.
+  Diagnoses and repairs the dark / grey / black-face NPC bug in Skyrim SE by diffing which mod wins the head .nif against which wins the face .dds for the same NPC, then the record winner behind them. Use for any discolored face, neck seam, "fine in xEdit but wrong in game", NPCs gone dark after an ESL-compaction or merge, or any FaceGen / facegeom / facetint mention. Not a purple or white face (a missing texture) and not player-only grey (RaceMenu/SKEE); copying a face onto another NPC is housecarl:npc-appearance-copy. Load before judging any face bug — the mesh-versus-tint pair decides the fix.
```
Avick3110, line 5: **The new negative-scope clause excludes a case this same PR's eval set requires to trigger, and that §2 still handles.**

The description now reads `not player-only grey (RaceMenu/SKEE)`. But `evals/evals.json:160` keeps `st-player-grey-after-update` — prompt "after the latest skyrim update my character's face is grey in the inventory menu but npcs are all fine" — with the expectations `The facegen-diagnostics skill is loaded.` and `The answer says houseCARL cannot fix this and names the runtime route.` §2 of the body carries that exact routing (skee64.dll, SKSE ↔ runtime ↔ RaceMenu, `skse64.log`), which is dead text if the description keeps the skill from loading.

Failure scenario: the router reads the description, sees player-only grey explicitly disclaimed, and abstains; the eval case fails on its first expectation, and the user gets no RaceMenu/SKEE route — the deliberate Q3 probe the old eval set called out as "the strongest exclusion rule is the skill's to own", which the prior fan-out scored as a recall hit.

Either drop the player-grey half of the exclusion (keeping the purple/white one), or drop the eval case and §2's bullet with it. Both cannot stand.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,251 @@
...
+```
+
+Then branch on the two winners:
+
+| The pair | What it is | Where to go |
+|---|---|---|
+| Same source | Not a pair fault | §5 — the record axis |
+| Different sources | A cross-bake: the head from one mod, the tint from another | §6 — re-place the pair from one source |
+| One present, one absent | The hard fault, and the one that accounts for nearly every finding on a big order | §6 — place the missing half from the source that has the other; if it exists nowhere, Ctrl+F4 |
+| Neither present | Nothing to place | §5 — does this NPC own a face at all; if it does, Ctrl+F4 |
```
Avick3110, line 73: **Mode ii has no branch in the rebuilt flow, yet §9 and §11 both point at one.**

The four rows here cover same-source, different-source, one-absent and neither-present. "Same source" routes to §5, whose two questions are *did the winning record change appearance* and *does this NPC own a face at all*. So the case the old body carried as an explicit third Step-3 bullet — a file wins **from the right source**, the record is right, and the face is still dark, i.e. baked shape names or the embedded slot-6 tint path disagreeing with the record — falls off the end of the flow with nothing to do.

Both downstream pointers presuppose the branch that is missing:
- §9: "Calling a green file status the all-clear ... read the winning mesh's shape names and embedded tint path" — no step in §3–§7 ever issues that `housecarl_nif_inspect` read.
- §11: read `references/mesh-repairs.md` "when §3 says the file wins but the face is still wrong" — §3 contains no such sentence.

Failure scenario: `st-modder-baked-still-black` ("baked the facegen with ctrl+f4 ... her head is still pitch black"). The pair resolves, both halves from the modder's own mod: row 1, same source, go to §5. The record is the only override, so the delta shows a normal appearance change; nothing owns a template. The run reports "pair clean, record clean" and stops — while the eval expects it to treat "a file wins at the path" as necessary-not-sufficient and read the winning mesh.

A fifth row, or a same-source sub-branch, that routes to `mesh-repairs.md` on `housecarl_nif_inspect` closes it.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,251 @@
...
+
+**Does the winning record change appearance at all?** Compare the appearance fields against the plugin
+immediately beneath the winner — a winner can change morph or tint while leaving `HeadParts` alone, so
+`HeadParts` on its own is not the test:
+
+```
+housecarl_records(formids=["013BBF:Skyrim.esm"], versus="previous_provider",
+  project={"form":"delta","fields":["HeadParts","FaceMorph","FaceParts","TintLayers",
+                                    "HairColor","HeadTexture","TextureLighting"]})
+```
```
Avick3110, line 128: **This call is refused on an uncontested NPC — which is one of the two ways §5 is entered.**

`versus="previous_provider"`'s contract (`RecordsTools.cs:127`) declares four cases, one of which is: *subject defines the record → refused naming it (never an empty diff that reads as 'no changes')*. With `source=` omitted the subject is the winner, so on any NPC no plugin overrides — the winner *is* the defining master — this delta comes back as a per-item error, not as "unchanged".

Failure scenario: §3's fourth row, "Neither present → §5 — does this NPC own a face at all". That row's population is dominated by NPCs nobody overhauled: a custom follower whose author never baked facegen, or a vanilla actor whose files exist nowhere. The record-axis step opens with this call, it refuses, and the body offers no next move — the flow stalls exactly where §10 expects a defensible verdict. The paragraph beneath ("Unchanged means the NPC rides the master's facegen at the same keyed path") reads as though a no-op diff is the answer for that case, when the tool refuses rather than returning one.

Worth one sentence saying that a refusal here *is* the "no plugin changed this record" answer, and to fall through to the master's-file check — or naming `project={"form":"tree"}`, which does not refuse an uncontested record.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,251 @@
...
+overhaul's folder. There is **no cross-folder fallback**: the engine reads that one keyed path and
+regenerates if nothing wins there. So ask "what wins this exact path", never "does it fall back".
+
+Resolving the NPC a user names:
+
+1. **Prefer EditorID, then display name.** `housecarl_records` with `types=["NPC_"]` and
+   `where=["editorid contains Lydia"]`. More than one hit ("Guard", "Bandit") → list the candidates and
+   let the user pick. Never auto-pick the first.
+2. **An xEdit-style FormID** the user read somewhere: drop the high byte (it is that person's
+   load-order index), keep the 6-hex local, and let houseCARL attach the defining master.
```
Avick3110: **"Keep the 6-hex local and let houseCARL attach the defining master" names a capability no parameter on the 2.0 surface has.**

A FormID token goes through `FormIdDoor.Parse`: `RuntimeFormId.TryParse` accepts **exactly eight** hex digits (`RuntimeFormId.cs:31,42`), and anything else falls to `FormKey.Factory`, which requires the `XXXXXX:Plugin.esp` spelling. A bare `013BBF` is neither, so `housecarl_records(formids=["013BBF"])` is a per-item "bad FormID" error. Nothing on the surface takes a local id and derives the master.

Failure scenario: a user pastes `2A013BBF` from a forum post. The reader follows this step, drops `2A`, calls `formids=["013BBF"]`, and gets a parse error with no fallback named — while §4's own path template needs the defining master to build the folder, so the flow cannot continue. (Passing the full eight digits *does* parse, but resolves the high byte against the reader's order, not the poster's, which is the trap this step exists to avoid — so "drop it" is right and the follow-through is what is missing.)

The honest move is the one step 1 already describes: use the local id as a hypothesis, find the record by EditorID or name with `housecarl_records` `types=["NPC_"]` + `where=`, and confirm the local id matches before acting. This step should say that rather than implying a tool does it.

(Carried over from the pre-2.0 body rather than introduced here, but the PR grades itself on "every tool name, parameter and call shape in the shipped tree resolves on the surface", and this one does not.)

.claude/skills/facegen-diagnostics/evals/evals.json
```diff
@@ -0,0 +1,366 @@
...
+    "grading": "Binary and evidence-bound: every expectation is PASS or FAIL with a quotation from that run's own transcript. No partial credit; a right-shaped answer with the wrong content FAILS. The hardened prompt (an explicit warning that the agent's own environment may list a stale copy of the description) is the default.",
+    "selection": "60/40 train / held-out split over the cases below; the description is selected on the held-out half only.",
+    "cost": "total_tokens and duration_ms are captured per run, in both configurations, and reported beside the pass-rate delta.",
+    "tier_scope": "Two model tiers today. A third tier is added, or the reason it is not is written down beside the results."
+  },
+  "gate": {
+    "per_query_trigger_floor": 0.5,
+    "recall_threshold": 0.8,
+    "specificity_threshold": 0.5
+  },
```
Avick3110, line 35: **The gate declares recall and specificity thresholds, but no case carries a machine-readable trigger label any more.**

The deleted `eval_set.json` gave every case `should_trigger: true|false`. `evals.json` drops the field entirely. The only remaining signal for which arm a case belongs to is the `st-`/`snt-` id prefix and English prose inside `expectations` ("The facegen-diagnostics skill is loaded." versus "... is NOT loaded.").

Failure scenario: anything reading this file to score the gate — `recall_threshold: 0.8`, `specificity_threshold: 0.5`, `per_query_trigger_floor: 0.5` — has no field to partition on, so it either string-matches the id prefix (undeclared anywhere in the file, and `st-copy-face-boundary` and `st-nifset-facetint-repair` both read like near-misses by name) or parses free text. The `gate_dimensions` entry for triggering accuracy names `snt-mesh-winner-generic` and `snt-pair-diff-generic` by id as the new probes, which only works if that prefix convention is load-bearing — but it is nowhere stated.

The rename also leaves this the only skill of the thirteen whose eval file is not `evals/eval_set.json`, which is worth being deliberate about since nothing in `standards/` pins the name either way.

Adding `should_trigger` back per case costs one line each and makes the declared gate computable.

.claude/skills/facegen-diagnostics/references/facegen-causes-and-fixes.md
```diff
@@ -305,49 +305,26 @@ surfacing that limit IS the Q3-honest move.
...
-material / `.tri` / xml ref → `set_path` with NO `texture_slot`, `target=` the string as `sections=strings`
-prints it; general missing tint → FDF; a **bulk** cross-mod rename → still faster in the
-batch tool. You no longer guess the broken slot, and for a single mesh you no longer hand off the edit.
+## 6. Embedded `.nif` texture references
+
+The head `.nif` carries its own asset references — the per-NPC FaceTint `.dds` at binary texture slot 6,
+the base skin diffuse/normal at slots 0/1, and the header strings a material, `.tri` or physics-xml ref
+lives in. Reading them, repairing them with `housecarl_nif_set`, the two verification gates a write
+passes, and the history of why those repairs used to be a NifSkope step are all in this skill's other
+reference, `mesh-repairs.md`. Causes F(b), I and R below, and the FaceTint half of Fix E, are resolved
```
Avick3110: **"below" points the wrong way.** In this file §3 (cause taxonomy A–X, incl. F, I and R) is at line 122 and §4 (fix taxonomy, incl. Fix E) at line 162 — both **above** this §6 at line 308. A reader who follows "below" scrolls into §7 Edge cases, §8 Community tools and §9 Residual uncertainty and finds none of them.

```suggestion
reference, `mesh-repairs.md`. Causes F(b), I and R in §3, and the FaceTint half of Fix E in §4, are resolved
```

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Refuted on the main claim, folded on the two halves that stand on their own.

**Enable, not sort, is a ruling.** `FOLD-IN.md` F2 carries Aaron's ruling of 2026-09-07 (item 4): "the folded sentence says **enabled**, not sorted: a new mod loads last by default, so sorting is not the step that makes the write win", and the sentence it folds into ServerInstructions reads "a new mod loads LAST, so enabling is the step, not sorting". The rewrite document executes that as correction 7, and says in the same breath what you found: "two shipped parameter descriptions still say 'enabled + sorted'" — those are F49, corrected in the same wave, not evidence against the ruling. `PlaceTools.EnableAndSort` and the `place` description are on the wrong side of a settled call, so the body is not changed back to match them.

Cause E is not an argument for sorting either: nothing a mod folder can be sorted to beats MO2's overwrite. What covers it is the sentence already in the body — "re-resolve the path (§7) rather than asserting it" — and §7's own row, which now reports who still beats the placed copy.

**Folded, both real:**
- The fabricated readback at `:153` is gone. It put a sentence in the tool's mouth the tool does not emit; the accounting line stays and the instruction is the body's own prose.
- §7's first verify row no longer says the tool "names the winner to rank above" — it now reads "Whether the placed copy now wins, and who still beats it if it does not", which is the check §6 actually asks for.

Also corrected while here: `references/facegen-causes-and-fixes.md:181` still said "a fresh enable+sort MO2 mod" (the regrade's D4.48 partial). `mesh-repairs.md:89-90` already says enable only and is left as written.

b5c29812.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded, on the eval side.

You are right that both cannot stand. The description text is fixed by the rewrite document — §3 gives it verbatim at 599 characters with the negative-scope clause named as what closes D3.11 — so the half that moves is the eval case. `st-player-grey-after-update` is now `snt-player-grey-after-update`, expecting the skill NOT to load and the player-only scope to be read as runtime state. That is the same shape as `snt-shiny-enb-face`, which is the other Step 0 exclusion sitting in the should-not arm already.

§2's bullet stays, and is not dead text: it is for the ambiguous arrival — a plain "grey face bug" prompt (`st-generic-greyface-bug`) loads the skill, and Step 0 is where the run finds out it is player-only and hands off to the RaceMenu/SKEE route. What the description excludes is the unambiguous prompt, which is exactly the one this case carries.

Counts go from 12 should-trigger / 18 should-not to 11 / 19, 30 cases in all.

b5c29812.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded. The branch was missing and both pointers presupposed it.

§3 now carries the sentence after the branch table:

> A file winning at the path is necessary, not sufficient. When the pair is same-source and §5 finds the record clean too, read the winning mesh itself with `housecarl_nif_inspect`: a baked shape name or an embedded FaceTint path that disagrees with the record is a mesh-side fault, repaired per §11.

That gives `st-modder-baked-still-black` its next move, makes §9's "read the winning mesh's shape names and embedded tint path" reachable from the flow, and makes §11's read condition — "when §3 says the file wins but the face is still wrong" — literally true of §3. It points at §11 rather than naming `mesh-repairs.md` a second time, so the body still names each reference exactly once (D4.14).

b5c29812.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded. `versus=`'s own contract declares it — "subject defines the record → refused naming it (never an empty diff that reads as 'no changes')" — and the fourth branch row sends uncontested NPCs straight into this call, so the flow stalled there.

§5 now says the refusal is the answer:

> `versus="previous_provider"` is **refused** when the subject defines the record — an uncontested NPC has no pole beneath it — and that refusal *is* the "no plugin changed this" answer, not an error. Unchanged, or refused that way, means the NPC rides the master's facegen at the same keyed path — benign, so confirm the master's file resolves.

Kept the `delta` call rather than switching to `project={"form":"tree"}`: the delta is what answers the question the step asks (did the winner change appearance), and the refusal now reads as a result instead of a dead end.

b5c29812.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded. `housecarl_records` states its two accepted spellings — "'XXXXXX:Plugin.esp' — 6 hex digits, a colon, the defining master's filename" or the eight-digit runtime form — and a bare local id is neither, so the step named a capability the surface does not have. Step 2 now reads:

> **An xEdit-style FormID** the user read somewhere: the high byte is *that person's* load-order index, not yours, so the 6-hex local under it is a hypothesis, not an address — no parameter takes a bare local id without its plugin. Find the record by EditorID or name as in step 1, then confirm the local id matches before acting on it.

That is the move step 1 already describes, and it keeps the reason the high byte is dropped rather than passed through.

b5c29812.

.claude/skills/facegen-diagnostics/evals/evals.json
Avick3110: Refuted: the rewrite document settles both halves.

§9c: "The file is renamed `evals/evals.json` and re-shaped to the published fields — `prompt`, `expected_output`, `files`, `expectations` — closing D4.22 (partial) and D5.16 (partial); the current `eval_set.json`'s house schema (`query`, `should_trigger`, `manual_predicted_trigger`, `rationale`) **does not survive**." One convention for all twelve is ruling 3, so the name is deliberate and the field's removal is the point, not an oversight — the same schema change lands on every skill in the wave.

The partition is not left to prose: `expectations[0]` is the machine-readable signal, and it is exact and generated — "The facegen-diagnostics skill is loaded." versus "The facegen-diagnostics skill is NOT loaded." — on every case in the file. The `st-`/`snt-` prefix follows it rather than carrying it, which is why `st-copy-face-boundary` reading like a near-miss is not a scoring hazard: its first expectation says loaded, and that is what recall counts.

If the harness wants a boolean field back, that is a decision for all twelve files at once, not this skill alone — worth raising against the convention rather than here. `st-player-grey-after-update` moved arms in this same push for the reason in the description thread, so the file's own two polarities now agree with the description.

b5c29812.

.claude/skills/facegen-diagnostics/references/facegen-causes-and-fixes.md
Avick3110: Folded, as suggested. §3 is at :122 and §4 at :162, both above this §6, so "below" sent the reader into the edge cases instead. It now reads "Causes F(b), I and R in §3, and the FaceTint half of Fix E in §4, are resolved there."

b5c29812.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,257 @@
...
+  patch="Facegen Fix")
+```
+
+`bundle=` is the field set — which paths form an appearance set is knowledge this skill carries, not
+something the tool owns — and `assignments=` pairs each target with the record it takes them from.
+When the whole record from that plugin is wanted instead, `housecarl_forward` with `source=` copies it
+verbatim.
+
+Then tell the user to **enable** the new mod in MO2. A new mod folder lands last in the priority order,
+so enabling is the step that makes the write win; re-resolve the path (§7) rather than asserting it.
```
Avick3110, line 190: **Dropping the sort step contradicts the tool's own output, and the claim under it is unsupported.**

On the contended path this section is about, `housecarl_place` renders (`src/housecarl-mcp/PlaceTools.cs:308`):

> Enable the mod 'Facegen Fix' in MO2 **and SORT it (left pane) ABOVE the current winner(s) listed above. Only then does the placed copy win.**

§3's own worked example has the head `.nif` winning from `Bijin NPCs (loose)`, so a facegen fix is contended by definition — that is the whole reason for the placement. An agent that relays "enable" alone sends the user to enable a mod that still loses the VFS, and the face stays dark. §7's re-resolve will report the miss, but the skill no longer carries the remedy anywhere in the tree (same wording now at `references/mesh-repairs.md:90` and `references/facegen-causes-and-fixes.md:198`).

"A new mod folder lands last in the priority order" is also not something houseCARL establishes: it only *reads* `modlist.txt` (`src/housecarl-core/Mo2LoadOrder.cs`) and never registers or positions the folder — where MO2 files a newly discovered folder on refresh is MO2's behaviour, not the server's. The wording this PR removed handled that explicitly ("trust its reported winner over the abstract rule").

Separately, the sentence says "the new mod" singular while the section opens with "often both are needed" — `housecarl_place` and `housecarl_apply` each write their **own** MO2 mod folder (`LoadOrderService.cs:9776` `OutputModFolder`), so both fixes leave two mods to enable.

.claude/skills/facegen-diagnostics/references/mesh-repairs.md
```diff
@@ -0,0 +1,120 @@
...
+Every write passes **two offset-immune verification gates before anything lands**: only the block and
+value the op claims to touch changed, a reload re-reads the new value, and the block census plus the SE
+stream are intact. A failed gate writes **nothing** and says why — there is no partial write to clean up.
+
+Two lanes:
+
+- **Default.** The verified mesh goes into a **new houseCARL MO2 mod folder** at the same path
+  (`patch_name=` names it, `into=` accumulates into an existing one). Originals are untouched, and a
+  BSA-packed source becomes a loose winning override this way. Enable that mod in MO2 — a new mod folder
+  lands last in the priority order, so enabling is the step that makes the edit win.
```
Avick3110, line 90: Same enable-only claim as `SKILL.md:185`, and `housecarl_nif_set` contradicts it in two places of its own:

- tool description, `src/housecarl-mcp/NifTools.cs:187` — "enable it and sort it **ABOVE the current winner** so the edit wins";
- response text, `NifTools.cs:839` — `TO MAKE IT WIN: enable this folder in MO2 and sort it ABOVE <winner>`.

The default lane exists precisely because another mod's copy currently wins the mesh, so this is the contended case every time. Enabling without sorting leaves the stale mesh winning and the face still grey.

.claude/skills/facegen-diagnostics/references/facegen-causes-and-fixes.md
```diff
@@ -193,20 +191,22 @@ into a fresh enable+sort MO2 mod. **Place BOTH `.nif` and `.dds` as a pair, from
...
 > re-masters). The engine DOES honor the embedded path, so the placed `.nif`'s slot-6 FaceTint path must be
-> repointed at the destination — which **houseCARL now does itself** with `nif_set set_path texture_slot=6`
-> (verified), no longer a NifSkope escalation. A bulk run across many such NPCs is still faster in
+> repointed at the destination — **`nif_set set_path texture_slot=6`**, verified before landing, rather than a
+> NifSkope escalation. A bulk run across many such NPCs is still faster in
 > FaceGenEslify.
 
-Then re-run `asset_status` and tell the user to enable + sort, and run the in-game verification (SKILL.md
-Step 5).
+Then re-run `asset_status` and tell the user to enable the new mod, and run the in-game verification (SKILL.md
```
Avick3110, line 198: Third site of the enable-only regression. Fix B is defined a few lines up as the case where "a correct copy exists somewhere (mod or BSA) but **loses VFS precedence**" — so another provider wins by construction. Enabling without sorting above it does nothing; the `asset_status` re-run this same sentence prescribes will show the placement still losing, and with the sort instruction now gone from `SKILL.md`, `mesh-repairs.md` and here, the skill carries no remedy for what that re-read reports.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,257 @@
...
+
+- **Player grey, NPCs fine** (after a reload, crash or game update) → RaceMenu/SKEE co-save state. If
+  every RaceMenu slider and overlay is gone game-wide, `skee64.dll` did not load: match SKSE ↔ runtime ↔
+  RaceMenu and read `skse64.log`. houseCARL is a no-op here.
+- **Brown face** matching nothing → weight baked into the save. Re-issue `setnpcweight`; if it is
+  save-baked, a new game or ReSaver.
+- **Purple or bright-white face** → a missing *texture* file. A different lane entirely.
+- **Shiny or oily face, ash pile** → specular/ENB or script state. Not facegen.
+- **An `FFxxxxxx` base id, or SPID/SkyPatcher-distributed appearance** → houseCARL reads *plugin*
+  records, so the winner it sees is not the in-game face. Warn; route to the distributed head parts.
```
Avick3110: **The hedge was hardened into a claim that is false for the SkyPatcher half.** `main` read "the winner it sees *may* not be the in-game face"; this now says "is not".

houseCARL reads the post-SkyPatcher state directly: `housecarl_records` takes `source={"overlay":"skypatcher","state":"post"}` (`RecordsTools.cs:125`), and there is a whole `housecarl_skypatcher_layer` tool. The NPC ops SkyPatcher can apply include `race` and `skin` (`src/housecarl-core/skypatcher-fieldmap.json`) — both face-relevant.

Failure: a user reports a head/body colour mismatch on an NPC whose `skin` is set by a SkyPatcher INI. The agent reads this bullet, declares houseCARL blind to the in-game face, and routes away — when one `source={"overlay":"skypatcher","state":"post"}` read would show the applied skin. The SPID half of the bullet is correct (no SPID overlay exists); it is the SkyPatcher half that now overclaims a blind spot the server does not have.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,302 +1,257 @@
...
+2. **An xEdit-style FormID** the user read somewhere: the high byte is *that person's* load-order
+   index, not yours, so the 6-hex local under it is a hypothesis, not an address — no parameter takes a
+   bare local id without its plugin. Find the record by EditorID or name as in step 1, then confirm the
+   local id matches before acting on it.
+3. **A runtime FormID** — the form the game, console, Papyrus and crash logs print, `FExxxYYY` or
+   `XX######` — is read directly by `housecarl_records` and by `housecarl_nif_inspect` `npc=`, resolved
+   against the current load order, and the response names the plugin it resolved to. Two traps remain: a
+   console click selects the **placed reference**, not the base `NPC_`, so read it back and confirm the
+   name before acting on it; and `housecarl_apply` refuses the runtime form by design, so write with the
+   `XXXXXX:Plugin.esp` form the read printed.
```
Avick3110, line 114: **Naming only `housecarl_apply` here understates the refusal, and it misses this skill's primary write.**

The write door is shared: `PlaceTools.cs:106` calls `svc.OpenWriteFormIdDoor()` — the same `FormIdDoor.ForWrite` as apply — so a `formid` member carrying a runtime form throws the same `WriteRefusal`, and the whole `housecarl_place` call is refused (`assets=` is all-or-nothing on a malformed member). `housecarl_forward` is on the same door.

Failure: a user pastes an `FE012ABC` from a crash log. The agent reads step 3, takes the refusal as apply-only, and goes to §6's `housecarl_place(assets=[{"formid": "FE012ABC"}], …)` — refused, nothing placed. Worth generalising to "the write tools (`housecarl_place`, `housecarl_apply`, `housecarl_forward`) refuse the runtime form by design", since §6 drives all three.

.claude/skills/facegen-diagnostics/evals/evals.json
```diff
@@ -0,0 +1,366 @@
...
+      "prompt": "after i compacted this follower mod to ESL the facetint texture path baked inside her head nif still points at the old form id and her face is grey, can you fix the embedded path",
+      "expected_output": "facegen-diagnostics is loaded and the repair is routed through the mesh-repair reference rather than to NifSkope.",
+      "files": [],
+      "expectations": [
+        "The facegen-diagnostics skill is loaded rather than a raw mesh-write.",
+        "The answer says a verified data-layer write is not a verified render and hands off the in-game check."
+      ]
+    },
+    {
+      "id": "st-copy-face-boundary",
```
Avick3110, line 180: **This new should-trigger case is graded against a description that cedes exactly this prompt.**

The description (`SKILL.md:5`) closes its negative scope with "copying a face onto another NPC is `housecarl:npc-appearance-copy`", and `npc-appearance-copy`'s own description claims "**a copied NPC rendering dark**". The prompt here — "i copied a face from an overhaul onto my custom follower and now her head renders dark … what's the right way to do that copy?" — matches both cede signals, so a model reading the descriptions correctly does *not* load `facegen-diagnostics`, and expectation 1 fails.

The `gate_dimensions` block then grades the other branch as a failure too: "`housecarl:npc-appearance-copy` loads instead of this skill on a diagnosis prompt" is listed as the coexistence failure mode. As written the case cannot pass either way.

Either the description needs a carve-back (the *diagnosis* of a dark copied NPC stays here; only the copy mechanics cede), or this case belongs in the should-not arm.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Refuted on the enable-versus-sort claim, folded on the singular, and two halves were already folded on the blind pass.

**Enable, not sort, is a ruling, and it still stands.** `dev/DECISIONS.md` 2026-09-07 (rewrite ruling 4) folds "nothing houseCARL writes wins until enabled in MO2" into ServerInstructions; the 2026-09-08 FOLD-IN acceptance turns "the two 'enabled + sorted' parameter descriptions" into "enabled" in the same PR; and the 2026-09-08 acceptance of #623 names the one exception, "a write into an existing mod keeps its priority and may need sorting". That is what shipped: `ApplyTools.cs:66` and `CreateTools.cs:48` read "the patch wins nothing until enabled in MO2, and a write into an EXISTING mod keeps that mod's priority and may still need sorting above the current winner." Section 6's lane is the fresh-folder one, so it takes the first half and not the second.

The reason the ruled sentence is true of the contended case is what "lands last" means: MO2 appends a newly discovered mod at the END of the left pane, which is the winning end of mod priority, so enabling is the step. `Mo2LoadOrder.CandidateFolders` ranking a fresh folder UNLISTED (`Mo2LoadOrder.cs:225`, overwrite then enabled then disabled then unlisted then Data) is houseCARL's own reading of a folder MO2 has not refreshed yet, not where MO2 files it once it has.

`PlaceTools.cs:36` and `:308`, `NifTools.cs:187` and `:837`, `WriteSentences.cs:117`, `WriteTools.cs:238`/`:483` and `LoadOrderService.cs:2396`/`:9774` are sites the ruling has not reached yet -- the same class as the two parameter descriptions #623 corrected, not evidence against it. They are code, not skill text; I have listed them for a code-side PR rather than reverting the body to match text the ruling supersedes.

Cause E is not an argument for sorting either: nothing a mod folder can be sorted to beats MO2's overwrite. What covers it is the re-resolve section 6 already prescribes, and the section 7 row that names who still beats the placed copy (`SKILL.md:203`).

**Folded -- the singular is wrong.** `place`'s `patch=` is "base name for the NEW houseCARL mod folder" and `apply`'s is "base filename for the NEW patch this call writes", each auto-suffixed if taken, so the two fixes section 6 calls "often both needed" do leave two mods. `SKILL.md:191` now says so and gives the one-mod recipe: write the record patch first, then pass its filename as `into=` on the `housecarl_place` call. c6acf78e.

**Already folded on the blind pass**, at `d13c716d`: the fabricated readback is gone -- `SKILL.md:163` is the accounting line only -- and section 7's first verify row now reads "Whether the placed copy now wins, and who still beats it if it does not" (`SKILL.md:203`).

.claude/skills/facegen-diagnostics/references/mesh-repairs.md
Avick3110: Refuted, on the same ruling as the `SKILL.md:189` thread.

`dev/DECISIONS.md` 2026-09-07 (rewrite ruling 4) folds "nothing houseCARL writes wins until enabled in MO2" into ServerInstructions, and the 2026-09-08 FOLD-IN acceptance turns the "enabled + sorted" parameter descriptions into "enabled"; the single exception the 2026-09-08 acceptance names is a write into an EXISTING mod, which keeps that mod's priority. `mesh-repairs.md:86-90` is the default lane -- a fresh houseCARL folder MO2 appends at the winning end of the left pane -- so it takes the ruled wording, and the in-place lane below it overwrites the winning file where it sits and needs no sort by construction.

`NifTools.cs:187` and `:837` are two more sites the ruling has not reached, the same class as the parameter descriptions #623 corrected. Listed with `PlaceTools.cs:36`/`:308`, `WriteSentences.cs:117`, `WriteTools.cs:238`/`:483` and `LoadOrderService.cs:2396`/`:9774` for a code-side PR, rather than reverting the reference to match text the ruling supersedes.

.claude/skills/facegen-diagnostics/references/facegen-causes-and-fixes.md
Avick3110: Refuted, on the same ruling as the two threads above.

Fix B is contended by construction, and the ruled sentence is written for exactly that case: "lands last" is last in MO2's left pane, which is the winning end of mod priority, so enabling a freshly appended folder puts it above the mod that held the path. `dev/DECISIONS.md` 2026-09-07 (rewrite ruling 4) folds that into ServerInstructions and the 2026-09-08 FOLD-IN acceptance corrects the "enabled + sorted" parameter descriptions to match; the exception it names is a write into an EXISTING mod, which is not this lane. `ApplyTools.cs:66` carries the shipped wording of both halves.

The `asset_status` re-run this sentence prescribes is therefore the check, not a dead end: if something still beats the placement -- MO2's overwrite folder being the case no mod priority can answer -- the re-read names it, and `SKILL.md:203` is where the skill says what that report means. The remaining enable+sort strings in the server (`PlaceTools.cs:36`/`:308`, `NifTools.cs:187`/`:837`, `WriteSentences.cs:117`, `WriteTools.cs:238`/`:483`, `LoadOrderService.cs:2396`/`:9774`) are listed for a code-side PR.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded. The hedge should not have been hardened, and the SkyPatcher half is readable.

Verified: `housecarl_records` takes `source={"overlay": "skypatcher", "state": "pre"|"post"}` (`RecordsTools.cs:125`, and the surface doc carries the same pole plus the `housecarl_skypatcher_layer` sentence pointing at it), and the NPC ops SkyPatcher can apply include `race` (set `Race`) and `skin` (set `WornArmor`) in `src/housecarl-core/skypatcher-fieldmap.json` -- both face-relevant. So a bullet that routes away from the whole class loses a read the server has.

`SKILL.md:43-47` now reads:

> - **An `FFxxxxxx` base id, or SPID-distributed appearance** -> houseCARL reads *plugin* records, so the winner it sees may not be the in-game face. There is no SPID overlay: warn, and route to the distributed head parts. **SkyPatcher it can read** -- `housecarl_records` with `source={"overlay":"skypatcher","state":"post"}` returns the record after the layer replays, race and skin included, so read that before declaring a blind spot.

"may not be" is restored, SkyPatcher comes out of the exclusion's subject line, and the SPID half keeps the warn-and-route it had. d493d6f2.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded. The door is shared, and section 6 drives all three tools.

Verified: `PlaceTools.cs:106` and the forward path (`LoadOrderService.cs:6393`) both open `svc.OpenWriteFormIdDoor()`, the same `FormIdDoor.ForWrite` as apply (`LoadOrderService.cs:115`), and `FormIdDoor.cs:49-53` throws the `WriteRefusal` on any runtime form at a write door -- handing back the plugin form to paste in its place. On `housecarl_place` a malformed member refuses the whole call, so the failure you describe is exact.

`SKILL.md:111-113` now reads "and the write tools -- `housecarl_place`, `housecarl_apply` and `housecarl_forward` share one write door -- refuse the runtime form by design, so write with the `XXXXXX:Plugin.esp` form the read printed." 8a5cfe74.

.claude/skills/facegen-diagnostics/evals/evals.json
Avick3110: **Left for you** -- both ways out overturn something the rewrite document rules, so I have changed neither.

The tension is real and it is between two ruled texts, not between the case and an accident:

- `REWRITE.md` section 9c prescribes the case as written: "**Added** for the rewrite: ... and one should-trigger boundary case whose correct outcome is the cede to `housecarl:npc-appearance-copy`." Moving it to the should-not arm deletes the probe the document asked for.
- `REWRITE.md` section 3 fixes the description verbatim at 599 characters, with "copying a face onto another NPC is housecarl:npc-appearance-copy" listed as the clause that closes D4.63 (fails) and D3.9 (partial). A carve-back edits ruled text.

And you are right that the sibling claims the symptom: `npc-appearance-copy`'s description names "a copied NPC rendering dark", so a host reading both descriptions correctly has two cede signals and one prompt.

If I had the call I would carve the description back rather than move the case -- "authoring a face copy onto another NPC is `housecarl:npc-appearance-copy`", one word, which keeps the diagnosis of a dark copied NPC here, keeps D4.63 closed because the overlap is still routed, and leaves the coexistence failure mode (`evals.json:50`, "`housecarl:npc-appearance-copy` loads instead of this skill on a diagnosis prompt") gradeable. The alternative -- renaming the case `snt-copy-face-boundary` and narrowing that failure mode to a prompt with no copy in it -- is one edit to each block if you prefer it.

Everything else in your review is folded or answered in the threads above; this is the only row waiting on you.

Issue #615, opened 2026-09-07
[GAP] check(findings=dialogue) cannot see an off-order plugin


Avick3110:
`housecarl_check(findings=["dialogue"])` only sees plugins that are enabled and sorted in the active order. dialogue-authoring writes into a new plugin that is neither until the modder enables it, so the skill cannot validate its own output where it lands.

Wanted: the dialogue check over an off-order plugin (a file in a mod folder, or a path), the way `records` already reads one. Found by the skill eval (dialogue-authoring D5.35). After 2.0 unless promoted (DECISIONS 2026-09-07, rewrite ruling 6).

Avick3110:
Recurred on both arms of the 2026-09-09 dialogue re-gate. housecarl_check(findings=["dialogue"], seeds=["000800:HC_Gate2_dialogue.esp"]) refused with "not in the active load order — nothing to validate" on each arm. The check that would catch a mis-wired branch, a stale .seq or a blank subtype marker is exactly the one a freshly authored, not-yet-enabled patch cannot run, so every dialogue deliverable ships unvalidated by its own family. This is now the second dialogue lane with no off-order arm; the other is info_order, filed as #694.

PR #645, opened 2026-09-08
Rebuild dialogue-authoring against the 2.0 tool surface

Avick3110:
Wave 3 of the skill rewrite: `dialogue-authoring`, executed from the approved rewrite document with its verifier's corrections applied. Every tool name in the old body was a 1.x name that no longer exists, so the whole body is rebuilt rather than swept: it now drives `housecarl_create`, `housecarl_apply`, `housecarl_records`, `housecarl_check` and `housecarl_remove`, and each call was checked against the 2026-09-08 surface capture rather than from memory. The body leads with the commonest job — one line into an existing topic, two calls, because `parent` resolves an existing topic's FormID — which is what the audit's own run asked for after measuring that most of the old read did not apply to it. It drops to 251 lines and ~4,630 tokens from 413 lines and ~8,490, which puts the whole body inside the compaction re-attach window. Three facts are corrected: the first-condition-wins rule is scoped to topics whose lines are told apart by their conditions and explicitly not to generic greetings, where how the engine picks among eligible lines is stated as unknown rather than asserted; the merged INFO order is a projection on `housecarl_records project={"form":"info_order"}`, not a finding on the check; and `Speaker` has to be set in the create call or the `.fuz` path cannot be computed. The dialogue check cannot see a plugin that is not enabled yet (#615), so the body names the off-order `errors`/`scripts` sweep as the pre-enable substitute instead of claiming coverage. Tool behaviour the surface already publishes — all-or-nothing, the Creation-Kit auto-fill list, "it flags, never silently rewrites" — is no longer restated in the skill. The player-topic semantics and the three write-side recipes move to `references/` behind read conditions, the six existing reference files get the same tool-name migration, the two over 100 lines gain a contents list, and the eval set is replaced with `evals/evals.json` in the published shape: 10 positive and 10 negative trigger cases plus the audit's Belethor task as the output-quality arm, every case with a without-skill arm, run from the packaged tree.

**The document's open questions.**

1. *Player-topic semantics: body or reference?* **Decided as follows** — the body keeps the two-line kernel (`Prompt` is the player's button, `Responses` the NPC's reply, `LinkTo` the player's next options) and `references/player-topics.md` takes the elaboration and the pre-roll pattern. The ledger keeps them "in the skill", which they are; the body cannot be both under the token ceiling and a home for everything.
2. *The write grammar the surface does not publish.* **Decided as follows** — the body teaches `op:"Add"` with `compose:` (both declared members of `housecarl_apply.ops`, and measured working in the audit) and never the nested `sets:` arm, which the surface does not declare. Cloning a condition gate goes through the `bundle=`/`assignments=` copy zip instead, so nothing is computed and no undeclared member is taught. **Left for Aaron:** whether `sets` is a declared member the capture omits, and what the published op vocabulary is — today only `Remove`, `Merge`, `ReplaceAll` and `CopyFrom` appear anywhere on the surface.
3. *The `<Global=X>` / `TextDisplayGlobals` check.* **Decided as follows** — dropped. It is not among the dialogue family's published checks and no issue is filed for it, so the skill states no claim it cannot back. **Left for Aaron:** whether an issue should be filed.
4. *`@editorid`.* **Decided as follows** — dropped, in the body and in `references/dialogue-branch.md` and `references/dialogue-flow-model.md`. Both now teach the published route: the create call reports each new FormID, and a second call on the same lane (`into=`) is how one new record points at another. **Left for Aaron:** whether `@editorid` is live and undocumented, which would be the same documentation gap as question 2.
5. *A surface floor in the description.* **Decided as follows** — left out. The body drives the live surface, the alias layer refuses a 1.x name with one sentence naming its successor, and a version string nothing maintains goes stale.
6. *The `.seq` re-run cost.* **Decided as follows** — the body says nothing about it. Since the document was written the answer landed on the surface: `housecarl_write_seq.into` now publishes that a byte-identical file is left alone and reported `unchanged`, so the fact is tool behaviour and the skill does not restate it.
7. *#615 is not in 2.0.* **Decided as follows** — shipped as stated. The body tells the reader to run the off-order `errors`/`scripts` sweep before the mod is enabled, then enable and seed the dialogue check, and names the issue.
8. *Stating the ignorance about greeting selection.* **Decided as follows** — shipped as stated ignorance. The body scopes the first-match rule, says plainly that neither it nor houseCARL can tell you how the engine picks among eligible generic greetings, and points at quest priority for which topic is reached at all. Eval case E5 is written against exactly this and fails for both prior runs, so the rewrite has to beat its own previous with-skill run, not only the control.

**Verifier rows folded.** The verifier overturned seven rows; two changed what shipped. `evals/evals.json` is *not* named from the body — the packager strips the directory, and ruling 2 bars a body pointer at a stripped file, so it is out-of-band test data. And the third-person description is kept for the reason the verifier gives (a third-person "what" clause plus an imperative "Use when…" trigger clause satisfies both stated styles), not for the reason the document gave. The other five are corrections to the document's own citations and change nothing in the tree.

**Judgement calls, so a reviewer does not have to guess.** The body came in at ~17,100 characters against the document's ~15,600 budget — 10% over the estimate, still 370 tokens under the 5,000 ceiling the gate measures. The `compatibility` line is worded "a configured Mod Organizer 2 instance" to match the wave 1 and wave 2 skills, since ruling 7 asks for the same wording throughout. The six existing reference files were migrated too, not just the body: they carried the same dead tool names, and the shipped tree is what a regrade reads. The dated in-body correction at the foot of `references/dialogue-flow-model.md` is dropped for the same reason as the body's — the corrected model is now simply what the section says.

Build: package build clean (leak-check clean over 12 skills and 252 markdown files, both trees), `dotnet build` clean, 1,793 tests passed, generator `ci-all` 128/128.

## Regrade

Assembled in `dev/skill-validation/eval-2026-09/dialogue-authoring/REGRADE.md` from the five grader
files in `regrade/`. Counts are over each checklist's own criteria (D1.1–D1.42, D2.1–D2.52,
D3.1–D3.33, D4.1–D4.64, D5.1–D5.50); grader-added rows are excluded, as each grader excluded them
from its own counts.

| Angle | Conforms | Partial | Fails | n/a | Total |
|---|---|---|---|---|---|
| D1 — discovery and packaging | 37 | 1 | 0 | 4 | 42 |
| D2 — frontmatter | 40 | 0 | 0 | 12 | 52 |
| D3 — loading decision and invocation control | 26 | 1 | 0 | 6 | 33 |
| D4 — body content and structure | 37 | 3 | 1 | 23 | 64 |
| D5 — validation and standing context | 41 | 3 | 0 | 6 | 50 |
| **Total** | **181** | **8** | **1** | **51** | **241** |

**Verdict: PASS** against the regrade half of the gate in `REWRITE.md` §9a. D2 and D3 carry zero
fails. D4 carries the regrade's only fail, D4.42 (bare MCP tool names), which its grader classifies
`other` rather than factual-or-dead-call — every bare name resolves on `surface/tools.md`, and §9a
itself says this row is superseded by ruling 1 and is to be recorded, not passed. D5 carries no fail
of any kind. D1 carries no fail; its one non-conforming row is D1.28 (release tag naming), a partial
the grader records as packager-only and outside the skill's files, and ruling 8 excludes it from the
gate.

The two factual defects the graders found sit inside partials, not fails:
`references/condition-functions.md:232` still names the dead 1.x `bulk_apply` and points at a recipe
that has moved to `references/write-side-recipes.md`, and `:116` teaches a `sets:` path that
publishes nowhere on the surface (D4.43). Both D4 and D5 also flag, without faulting, that the two
primary worked examples write `"op": "Add"`, a value the surface neither publishes nor refutes —
open question 2 above, still open.


.claude/skills/dialogue-authoring/references/write-side-recipes.md
```diff
@@ -0,0 +1,77 @@
...
+   housecarl_apply(
+     into="MyPatch.esp", readback=true,
+     bundle=["Conditions"],
+     assignments=[
+       { "target": "0A12C4:MyMod.esp", "from": "0B77E0:Skyrim.esm" },
+       { "target": "0A12C5:MyMod.esp", "from": "0B77E0:Skyrim.esm" } ])
+   ```
+
+   Whether the copy merges into the target's existing entries or replaces them is the choice between
+   `op=Merge` and `op=ReplaceAll`. `target` and `from` must be the same record type — INFO to INFO.
```
Avick3110: **`op=Merge` / `op=ReplaceAll` is not a knob the copy zip has.**

`housecarl_apply` declares no top-level `op=`, and `assignments[]` declares only `{target, from, from_source}` — an undeclared member is refused by name, not ignored. `ApplyTools.ExpandZip` hard-codes every generated edit as `Op = "CopyFrom"`, and `WriteEngine.TransplantValue` rebuilds a list field from the source's elements, i.e. CopyFrom always transplants the WHOLE field. There is no merge behaviour to select.

Scenario: a reader who wants the source rows added to a target's existing gate writes `op="Merge"` on the apply call (refused: unknown parameter) or inside an assignment entry (refused by name) — and if they instead trust the sentence and run the recipe as written, the target's existing `Conditions` are silently replaced, which is the opposite of what "merges into the target's existing entries" told them to expect.

Either state plainly that the zip always replaces the named field, or route a genuine merge through `ops=` with `op:"Add"` per row.

.claude/skills/dialogue-authoring/references/write-side-recipes.md
```diff
@@ -0,0 +1,77 @@
...
+**Never hand-synthesize the operator bytes.** A `Condition` (CTDA) is a polymorphic struct: a
+`ConditionFloat` carrying a `CompareOperator`, a `ComparisonValue`, and a polymorphic `Data` holding
+the function and its parameters. Computing that encoded operator and comparison by hand is exactly
+what once wrote 26 broken conditions onto one gate. So read a known-good gate and copy it.
+
+1. **Build the gate once** — in the Creation Kit, or on one INFO you have validated — and read it
+   back: `housecarl_records(formids=["<source>"], project={"form":"fields","fields":["Conditions"],"depth":4})`.
+   Note that `depth` is a sub-parameter *inside* the form; passing it beside the form is refused, and
+   the `delta` form does not take it at all.
+2. **Read each target first and skip any that already carry `Conditions`.** There is no idempotent
```
Avick3110: **This step is stale from the old `Add`-based recipe and now makes the recipe skip the targets it exists to fix.**

"There is no idempotent verb, so the read-then-skip is yours to do" was true when step 3 replayed each source row with `op:"Add"`. Step 3 is now the `bundle=`/`assignments=` zip, which expands to `CopyFrom` — a deterministic whole-field transplant. Re-running it is already safe; nothing accumulates.

Scenario: 26 INFOs carry the broken hand-synthesized gate this recipe was written to repair. The reader follows step 2, sees every one of them already carries `Conditions`, skips all 26, and reports the clone done — with not one gate fixed.

Drop the skip, or invert it: read each target first so you know what you are about to overwrite.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -1,413 +1,250 @@
...
+
+- **Adding a line** — list only your new line. Nothing else moves. Give it a PNAM only if it must sit
+  at a particular spot rather than last.
+- **Changing an existing line** — an override of an existing INFO *is* a re-list structurally, so
+  editing one line's text moves it to the bottom unless you **carry its PNAM**. Treat that as the
+  default. Vanilla lines carry none, so there is nothing to inherit and you must supply it — which is
+  exactly what a well-behaved patch like USSEP does when it re-lists six lines and moves none.
+
+**Removing a line is not done by omitting it,** and `housecarl_remove` usually is not the tool.
+Omission is a no-op. On the default lane `housecarl_remove` drops *your patch's* override, reverting
```
Avick3110: **`housecarl_remove` has no default lane.**

`RemoveTools.cs` refuses a call with neither lane: *"no lane named. Removal never creates a new artifact — it edits one that exists: pass into=<a houseCARL patch's filename> ... or in_place=<plugin filename>"*. "Exactly one lane per call" is the tool's own wording.

Scenario: a reader takes "on the default lane" at face value and calls `housecarl_remove(formids=["<info>"])` to drop their patch's override — refused, with nothing in the skill telling them `into="<patch>.esp"` is what they needed. The sentence also never names `into=` anywhere, so the one lane the paragraph is actually describing is the one it leaves out.

Suggest: "With `into="<patch>.esp"` it drops *your patch's* override, reverting the line to the underlying winner…"

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -1,413 +1,250 @@
...
-   (`field_path="VirtualMachineAdapter.ScriptFragments"`). Both are nullable-field clears the write engine
-   allows by construction (the adapter is a nullable substruct, `ScriptFragments` a nullable polymorphic
-   field); an explicit non-nullable/required field would refuse a `Remove` instead. Read back to confirm the
-   binding is gone before enabling the patch.
+   A spoken row is a composed struct: `op:"Add"` with `compose:{ "type":"DialogResponse", "fields":{…} }`,
+   values as strings (`"ResponseNumber":"1"`, not `1`). `DialogResponse` singular is the spoken row;
+   `DialogResponses` is the INFO record.
+3. **Make it reachable.** A `Custom` topic with no entry point is byte-valid and never entered — only
+   generic subtypes are matched without one. Author a `DialogBranch` whose `StartingTopic` names the
+   topic. Each new record's FormID is reported back, and that FormID plus a second call on the same
```
Avick3110: **`@editorid` is published on the live surface, so the two-call substitution is a capability regression built on a false premise.**

The PR body's open question 4 leaves "whether `@editorid` is live and undocumented" for Aaron. It is live and it is documented — twice, on `housecarl_create` itself:

- `records[].editorid`: "…and a FormLink value can reference it as `'@<editorid>'`."
- `records[].ops[].value`: "…or `'@<editorid>'` for a same-call sibling on a FormLink field."

And `WritePatchBuilder.ResolveSiblingRefs` / `ResolveStructSiblingRefs` resolve it across `value`, `values`, `entries`, compose `fields` and nested `sets`. (Same for open question 2's `sets:` — it is a declared member of `compose` on both `housecarl_create` and `housecarl_apply`.)

Scenario: the reader authors a topic plus its `DialogBranch`, and now pays a second write call plus a FormID round-trip for a link the create call would have set in the same all-or-nothing pass. The substitution also lands in `references/dialogue-branch.md` and `references/dialogue-flow-model.md` (the PNAM chain), so the one-call route is gone from the whole tree.

Secondary: "a second call on the same lane (`into="<this patch>.esp"`)" never names *which* tool. Setting a field on a record that now exists is `housecarl_apply`, not `housecarl_create`; as written a reader can reasonably call `housecarl_create` again and get a second DLBR.

.claude/skills/dialogue-authoring/references/condition-functions.md
```diff
@@ -195,14 +205,15 @@ The frequent confusions are the `Is…` Papyrus methods vs the `Get…` conditio
...
 
 ## Decoding a condition you read back — worked example
 
-`housecarl_read_record` on an INFO with `Conditions` deep returns, per row, the arm + operator + value +
-the Data arm + its params. Read this row:
+`housecarl_records(formids=["<info>"], project={"form":"fields","fields":["Conditions"],"depth":4})`
+returns, per row, the arm + operator + value + the Data arm + its params. Read this row:
 
 ```
 ConditionFloat
```
Avick3110, line 220: **A dead 1.x tool name and a dangling pointer survive at line 232 of this file** (outside the diff hunks, so flagging it here).

> `back and replay its rows verbatim (the `bulk_apply` clone recipe in `SKILL.md`).`

Two defects in one clause, and the PR's stated goal is that every tool name it teaches exists:

1. `bulk_apply` is a 1.x name that no longer exists — the only one the migration missed in this tree (`git grep` over `.claude/skills/dialogue-authoring/` on this branch returns exactly this line).
2. The clone recipe is no longer in `SKILL.md`; this PR moved it to `references/write-side-recipes.md`. A reader who follows the pointer finds a `## Common mistakes` bullet, not the recipe.

The PR body's regrade already records this one — it should be fixed rather than shipped as a known partial, since it is precisely the class the rewrite exists to remove.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -1,413 +1,250 @@
...
+               { "field_path": "Responses", "op": "Add",
+                 "compose": { "type": "DialogResponse",
+                              "fields": { "Text": "Browse as you like.", "ResponseNumber": "1" } } } ] }])
+```
+
+Then sweep the patch before it is enabled — this lane works off-order, the dialogue one does not
+(section "Validate, then hand off"):
+
+```json
+housecarl_check(plugins=["MyGreeting.esp"], findings=["errors","scripts"])
```
Avick3110: **The fast path's second call can sweep a different file from the one the first call wrote.**

`housecarl_create`'s `patch=` is a *base* filename: "auto-suffixed if taken, so a prior patch is never overwritten." So `patch="MyGreeting"` produces `MyGreeting.esp` only when that stem is free.

Scenario: the reader ran this fast path last week, so `MyGreeting.esp` already exists. This call writes `MyGreeting-2.esp` (or whatever stem is free) — and the hardcoded `housecarl_check(plugins=["MyGreeting.esp"], ...)` then sweeps *last week's* patch, comes back clean, and the reader hands off a plugin nothing verified. A silently wrong answer from a two-call recipe, which is the failure class the cornerstones forbid.

The create call reports the filename it actually wrote; say to pass that, rather than repeating the stem.

.claude/skills/dialogue-authoring/evals/evals.json
```diff
@@ -0,0 +1,296 @@
...
+  "model_tiers": "Haiku, Sonnet and Opus. If a tier is skipped the deployment envelope is stated in results rather than left implied.",
+  "run_configurations": [
+   {
+    "name": "isolation",
+    "menu": "dialogue-authoring alone plus NONE",
+    "failure_mode_it_catches": "over-trigger -- the description claims ground it does not own (SPID/KID distribution, SkyPatcher field edits, a dark face, a schema lookup) and absorbs it because nothing else is on the menu"
+   },
+   {
+    "name": "coexistence",
+    "menu": "dialogue-authoring among the other eleven skills plus NONE",
```
Avick3110: **Off by one against the tree this file declares under test.** The coexistence menu is specified as "dialogue-authoring among the other eleven skills" — twelve entries. `.claude/skills/` on this branch holds eleven skill directories *including* `dialogue-authoring`, so there are ten others. (The changelog hunk in this same PR says "Eleven skills ship now.")

Scenario: whoever runs this suite builds a twelve-skill menu, finds only eleven skills exist, and has to guess whether a skill is missing from the tree or the number is wrong — before a single run.

.claude/skills/dialogue-authoring/evals/evals.json
```diff
@@ -0,0 +1,296 @@
...
+    "with-skill arm: the reasoning names a record's own fields as the target",
+    "without-skill arm: removing dialogue-authoring does not change where this query lands"
+   ],
+   "why_this_case": "near-miss -- a record's own fields, the third clause of the boundary"
+  },
+  {
+   "id": "T14",
+   "prompt": "my quest journal book renders as [...] instead of the text I wrote",
+   "should_trigger": false,
+   "expected_output": "dialogue-authoring is NOT loaded; the query routes to the dynamic book-entry skill.",
```
Avick3110: **T14's expected destination is not in the tree under test, so the case cannot be scored as written.**

`expected_output` says the query "routes to the dynamic book-entry skill". No such skill exists in this repo — `git ls-tree -r pr645-review | grep dynamic-book` returns nothing, and it is not among the eleven directories in `.claude/skills/`. (`dynamic-book-entries` is a personal user-level skill at `~/.claude/skills/`, and this file's own `tree_under_test` says the packaged tree is what is menued and that the installed plugin's skills are disabled for the run.)

Scenario: the runner builds the menu from the packaged tree, T14's expected destination is absent from it, and the case's positive assertion is unscorable — only the negative half ("the pick is not dialogue-authoring") can be judged. Either name a sibling that ships, or state the expectation as NONE.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -1,413 +1,250 @@
...
+## Order, and the reorder trap
+
+**Lines are not dropped by a dialogue conflict.** Every plugin that touches a topic contributes its
+lines and the game merges them, so a line you do not re-list still plays. What a conflict changes is
+**order**. "Lines vanished" almost always means a line is still there but now sits behind a broader
+one that also passes.
+
+**Where position decides, and where it does not.** In a topic whose lines are disambiguated by their
+`Conditions`, the eligible set is what matters and order settles a tie: the game walks the topic and
+plays the first line whose conditions pass. That rule does **not** extend to generic greeting
```
Avick3110, line 90: **This scoping contradicts the tool surface's own published claim, with nothing to reconcile the two.**

`housecarl_records`' `project.form='info_order'` description states flatly: *"The game walks the sequence top to bottom and plays the FIRST passing line; re-listing a line appends it to the BOTTOM unless the plugin also carries its PNAM…"* — no greeting exception. This section now says that rule "does **not** extend to generic greeting subtypes" and that how the engine picks among eligible greetings is unknowable.

Scenario: a reader runs `info_order` on `DialogGenericHello`, reads the tool's own text saying the first passing line wins, and reads the skill saying nobody knows — two rules from the same product for the same topic, and no way to tell which is the correction.

If the scoping is right, the tool description is the thing that is now overclaiming and should be filed; if the tool is right, the skill is asserting an ignorance the product does not have. Worth settling one way before this ships, since the PR's own eval E5 is written against it.

.claude/skills/dialogue-authoring/references/condition-functions.md
```diff
@@ -7,12 +7,21 @@ fills; houseCARL cannot evaluate it (only the game can), but it **can** read it
...
-field path or arm name against `mutagen-reference` before composing a write.
+field path or arm name against `housecarl:mutagen-reference` before composing a write.
+
+## Contents
+
+- **The shape of a condition (CTDA)** — the two arms, the operator, the comparison value.
+- **Run On** — Subject vs Target and the rest, and why it is the commonest silent error.
+- **The `OR` flag** — how rows group, and what it cannot express.
+- **Dialogue / quest condition functions** — the curated table.
+- **CTDA conditions vs Papyrus-only functions** — the `Is…` / `Get…` trap.
```
Avick3110, line 24: **The new contents list does not match the file's headings.** It omits `## Reading a function's parameters` (a top-level section, and the one a reader composing a condition most needs), while listing "The `OR` flag", which is a `###` nested under "The shape of a condition (CTDA)" rather than a section of its own.

Scenario: the contents list is the thing a reader scans to decide whether to read further under a token budget; a reader looking for parameter shapes concludes this file does not cover them and goes to `mutagen-reference` instead.

.claude/skills/dialogue-authoring/references/write-side-recipes.md
Avick3110: Refuted. The rewrite document settles this the other way, and the sentence is the surface's own.

REWRITE.md row 21 replaces the 1.x row-by-row `Add` replay with exactly this recipe and states its entry-level choice: "whose entry-level choice is `op=Merge` or `op=ReplaceAll` (`housecarl_copy.seed_paths`)". That citation is live on `surface/tools.md`, which is the only source this rewrite may name a tool or parameter from: `housecarl_copy.seed_paths` reads "use `housecarl_apply`'s `bundle=`/`assignments=` zip, where `op=Merge` and `op=ReplaceAll` are your choice of merging into the target's entries or replacing them."

The finding rests on `ApplyTools.ExpandZip` and `WriteEngine.TransplantValue`, i.e. on the implementation rather than the published surface. If the implementation and the surface disagree, that is a product defect to file against the tool descriptions, not a correction to make in this skill: the skill states what the surface publishes. Leaving it with Aaron on that basis.

The recipe's real defect that came out of your other comment -- step 2's read-then-skip -- is folded in 558f421b.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Primary claim refuted; the secondary point is folded.

`@editorid` appears nowhere in `surface/tools.md`, the 2026-09-08 capture this rewrite is executed against and the only source it may take a tool or parameter name from. `housecarl_create.records` there publishes the opposite route in so many words: each new FormID is "REPORTED BACK: that is how you reference the record afterwards -- to make ANOTHER record point at it, call this tool or housecarl_apply again with that FormID as the value and the same lane (into='<this patch>'...)". The only `@` spelling the capture publishes is `"@<absolute path>"` for a manifest. REWRITE.md row 24 rules on exactly this -- "NOT EXPRESSIBLE ... The published route is two calls" -- and open question 4 leaves "live and undocumented" for Aaron, which is the documentation gap you are pointing at, not a change this skill may make ahead of it. Same for `sets:` under open question 2. A skill that teaches a spelling the surface does not publish is the failure the rewrite exists to remove.

Your secondary point is real and is folded in eff74f60: the second call is now named as `housecarl_apply` in `SKILL.md`, `references/dialogue-branch.md` and `references/dialogue-flow-model.md`, since the record exists by then and this sets a field on it.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Refuted. The sentence you quote is not on `surface/tools.md`; the capture's `project` entry lists `info_order` among the forms and carries no first-passing-line claim, so there is no published contradiction for the skill to reconcile.

The scoping itself is a ruling, not a judgement call made here: REWRITE.md §7 correction 1 overturns the old body's universal first-match rule on measured evidence -- Belethor's six vanilla `Hello` lines carry `Flags.Flags = 0`, no `Random` flag, and cycle in game -- and open question 8 ships the stated ignorance. Eval case E5 is written against exactly that and fails for both prior runs, so it is the one new expectation the rewrite has to beat.

If the live `info_order` description does assert first-match without the greeting scoping, that is a tool description to file against, and worth filing -- but it is a change to the tool, not to this skill, and the capture the rewrite was graded on does not carry it.

Avick3110:
Fold pass over the ten review findings: seven folded, three refuted in replies on the comments themselves.

**Folded** (eff74f60, 558f421b, d0d2f5c6, 72b24d7f)

- The fast path swept the `patch=` stem, which is auto-suffixed when taken, so a re-run could verify a previous session's plugin. It now sweeps the filename the create call reports.
- `housecarl_remove` has no default lane; the paragraph now names `into="<your patch>.esp"`, the lane it was describing.
- The second call that points one new record at another is now named as `housecarl_apply`, in the body and in the two references that repeat the route.
- Recipe A's step 2 told the reader to skip every target that already carries `Conditions` -- every broken gate the recipe exists to repair. It now reads each target to know what the copy overwrites.
- `references/condition-functions.md` carried the tree's last 1.x tool name (`bulk_apply`) and a pointer at a recipe this PR moved; both fixed. Its contents list omitted "Reading a function's parameters" and listed a subsection as a section; both fixed.
- `evals.json`: the coexistence menu said eleven others where ten ship beside this skill, and T14 expected a skill absent from the tree under test -- the expected pick there is NONE.

**Refuted**, each with the reason on its own thread: the `op=Merge`/`op=ReplaceAll` sentence (published on `housecarl_copy.seed_paths` and chosen by REWRITE.md row 21); `@editorid` and `sets:` (absent from `surface/tools.md`, and REWRITE.md row 24 / open questions 2 and 4 leave the documentation gap with Aaron); and the `info_order` greeting scoping (the quoted first-passing-line sentence is not in the surface capture, and REWRITE.md §7 correction 1 plus open question 8 settle the scoping on measured evidence).

Package build clean (leak-check clean, 12 skills / 252 markdown files across both trees), `dotnet build` clean, 1,793 tests passed, generator `ci-all` 128/128.

.claude/skills/dialogue-authoring/references/write-side-recipes.md
```diff
@@ -0,0 +1,78 @@
...
+   housecarl_apply(
+     into="MyPatch.esp", readback=true,
+     bundle=["Conditions"],
+     assignments=[
+       { "target": "0A12C4:MyMod.esp", "from": "0B77E0:Skyrim.esm" },
+       { "target": "0A12C5:MyMod.esp", "from": "0B77E0:Skyrim.esm" } ])
+   ```
+
+   Whether the copy merges into the target's existing entries or replaces them is the choice between
+   `op=Merge` and `op=ReplaceAll`. `target` and `from` must be the same record type — INFO to INFO.
```
Avick3110: **The copy zip takes no `op` — the Merge/ReplaceAll choice does not exist.**

`housecarl_apply`'s zip arm declares only `bundle` (a string list) and `assignments[] = {target, from, from_source?}`. `ExpandZip` in `src/housecarl-mcp/ApplyTools.cs:248` sets `Op = "CopyFrom"` on every op it generates; there is nowhere to pass Merge or ReplaceAll, and `Merge` is a dict-only op that could not apply to the list field `Conditions` even on the `ops=` arm.

Two ways this bites. A reader who takes the sentence literally adds `op="Merge"` and the call is refused by name. A reader who doesn't is left believing the behaviour is selectable: step 2 above now tells them explicitly *not* to skip targets that already carry `Conditions`, so they run the zip across an INFO holding a partially-correct condition stack, and CopyFrom replaces it wholesale with nothing said. Stating that the copy replaces the named field is what step 2 needs to be safe.

```suggestion
   The copy **replaces** the target's `Conditions` with the source's — `CopyFrom` is the only verb the
   zip issues, and there is no merge variant. That is why step 2 says to read each target first.
   `target` and `from` must be the same record type — INFO to INFO.
```

.claude/skills/dialogue-authoring/references/dialogue-branch.md
```diff
@@ -56,10 +56,11 @@ other dialogue.
...
 `LinkTo` target of another reachable topic. A byte-valid `Custom` topic with neither is **never entered** in
-game — the most common "I added a topic and nothing happens" cause. When you author a new player-choice menu,
-author the DLBR in the **same** `housecarl_bulk_create` call (declared *after* the topic) with
-`StartingTopic` set to the topic's `@editorid`, `Category = Player`, and `Flags = TopLevel` for an ordinary
-side-by-side option. (See the `SKILL.md` reachability note.)
+game — the most common "I added a topic and nothing happens" cause. When you author a new player-choice
+menu, author the DLBR with `Category = Player` and `Flags = TopLevel` for an ordinary side-by-side
+option, and point its `StartingTopic` at the topic's FormID — the create call reports each new record's
+FormID, and a second `housecarl_apply` call on the same lane (`into="<this patch>.esp"`) is how one new
+record is made to point at another. (See the `SKILL.md` reachability note.)
```
Avick3110: **`@editorid` is published and live — this drops a working one-call route on a false premise.**

The PR body's open question 4 treats `@editorid` as undocumented and replaces it with a second `housecarl_apply` call. It is documented, on the create tool's own schema:

- `records[].ops[].value` — "a FormID 'XXXXXX:Plugin.esp', or '@<editorid>' for a same-call sibling on a FormLink field"
- `records[].editorid` — "a FormLink value can reference it as '@<editorid>'"

And implemented: `WriteEngine.SiblingRefSigil` / `IsSameCallSiblingRef` (`src/housecarl-core/WriteEngine.cs:3221-3234`), substituted by `WritePatchBuilder.CreateRecords` after allocation, in spec order.

The deleted text was correct, including the ordering constraint ("declared *after* the topic") that the substitution loses. The replacement is worse than verbose: authoring a topic plus its DLBR is now two writes, and if the second one is refused or forgotten the patch ships a DLBR with a null `StartingTopic` and a `Custom` topic that is never entered — the exact failure this section exists to prevent.

`SKILL.md:158-160` carries the same exclusive claim ("is how one new record is made to point at another") and needs the same fix.

```suggestion
game — the most common "I added a topic and nothing happens" cause. When you author a new player-choice
menu, author the DLBR in the **same** `housecarl_create` call, declared *after* the topic, with
`Category = Player`, `Flags = TopLevel` for an ordinary side-by-side option, and `StartingTopic` set to
the topic's `@editorid` — the create path substitutes the sibling's allocated FormID. (Across two calls
instead, the create reports each new FormID and a second `housecarl_apply` on the same lane
(`into="<this patch>.esp"`) sets the link.) (See the `SKILL.md` reachability note.)
```

.claude/skills/dialogue-authoring/references/dialogue-flow-model.md
```diff
@@ -5,9 +5,18 @@ by reading a live load order through houseCARL (a generic-goodbye topic, a DA03
...
-`mutagen-reference` before composing a write.
+Field names below are Mutagen spellings (what `housecarl_create` / `housecarl_apply` and the
+`housecarl:mutagen-reference` skill use), with the xEdit signature alongside. Confirm any exact field
+path against a real record or that skill before composing a write.
+
+## Contents
+
+- **Record hierarchy** — DLVW / DLBR / DIAL / INFO and their key fields.
+- **What drives the flow** — entry, eligibility, the topic chain, the quest tie.
+- **Authoring traps the model implies** — the four that no check can catch.
```
Avick3110, line 16: **The greeting carve-out did not reach this file, and it is the file the body points readers at.**

`SKILL.md:88-97` now scopes first-condition-wins to topics whose lines are told apart by their conditions, and says plainly that greeting selection among eligible lines is unknown. `plugin/CHANGELOG.md` ships that as a corrected fact. This reference — untouched at the relevant lines — still asserts the rule unscoped, three times:

- `:47` — "the game walks the topic top to bottom and plays the **first** eligible line"
- `:70-72` — "only the **first valid Info** in a topic plays … so everything after the first is dropped"
- `:118-119` — "order decides which line answers, because the game walks the topic top to bottom and plays the **first** line whose `Conditions` pass"

`SKILL.md:218-219` tells the reader to read this file "before authoring or auditing anything, **and before trusting any claim about which line plays**." So on the skill's own headline job — a Hello line on Belethor, eval case Q1 — the reader follows that pointer, reads `:70-72`, and concludes the bottom-appended line is starved. That is precisely what expectation **E5** fails on, and it contradicts `SKILL.md:91` ("Belethor's six vanilla `Hello` lines … demonstrably cycle in game") in the same shipped tree.

Either scope the three statements here the way the body does, or drop the greeting carve-out from the body — but the two files cannot ship saying opposite things about the one fact the changelog claims was fixed. `references/player-topics.md` gets this right by excluding itself up front ("Nothing here applies to a generic greeting or a goodbye"); this file has no such fence.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -1,413 +1,252 @@
...
-   first, then each `DialogResponses` with `parent` naming the topic's editorid (that nests the line into
-   the topic's `Responses`, so a line cannot stand alone). A worked example — a player-choice topic with two
-   lines, the second chaining on to another topic:
+| 1 | Wire topic ↔ branch ↔ quest | set `DialogTopic.Quest` / `Subtype` / `Category`; author a `DialogBranch` whose `StartingTopic` points at the topic | a `Custom` topic with no inbound branch and no inbound `LinkTo` is byte-valid and **never entered** |
+| 1b | The SNAM subtype marker | set `Subtype` on the topic and let the create path derive the marker | a new topic with a `Subtype` but a **blank** marker is a **load CTD**; set `SubtypeName` by hand only to override, or when editing a topic outside the create path |
+| 1c | CK-parity subrecords | let `housecarl_create` fill them; it reports each field it filled | an INFO with no CNAM/ENAM **crashes the Creation Kit** when its topic is opened (the game tolerates it); a bare DLVW crashes the CK's Dialogue Views editor. `Branch` (BNAM) cannot be derived — set it yourself, and `Goodbye` enders still need `Flags.Flags = Goodbye` |
```
Avick3110: **The BNAM row lost its Custom-topic scoping in the compression.**

The old body said: "`Branch` (BNAM) can't be auto-derived, so it stays yours — set it to the owning `DialogBranch`; the check **warns** a **`Custom` topic** with no BNAM." The new text is unconditional: "`Branch` (BNAM) cannot be derived — set it yourself."

That scoping is load-bearing, and two other places still carry it:

- `src/housecarl-core/DialogueValidate.cs:439-449` fires only for a `Custom` topic, and only as a Creation-Kit flowchart concern — its own text says "it plays fine in game".
- `references/dialogue-branch.md:21-23` — "generic topics have no branch … **Don't read a missing `DialogTopic.Branch` as a defect**."

Sitting in the CK-parity row beside two genuine crash items (missing CNAM/ENAM, a bare DLVW), the unscoped instruction reads as required for every topic. A reader on this body's own headline fast path — one INFO into a vanilla `Hello` topic — is told to set `Branch` on that DIAL, writing a value no generic topic carries into a record the same section otherwise warns against touching (the lean-host note at :58-64).

```suggestion
| 1c | CK-parity subrecords | let `housecarl_create` fill them; it reports each field it filled | an INFO with no CNAM/ENAM **crashes the Creation Kit** when its topic is opened (the game tolerates it); a bare DLVW crashes the CK's Dialogue Views editor. `Branch` (BNAM) cannot be derived — set it yourself on a **`Custom`** topic (generic topics carry none, and a missing one there is not a defect), and `Goodbye` enders still need `Flags.Flags = Goodbye` |
```

.claude/skills/dialogue-authoring/references/write-side-recipes.md
Avick3110: Folded in cb16a276. The zip's own expansion settles it: `ApplyTools.cs` stamps `Op = "CopyFrom"` on every op it generates from `bundle=`/`assignments=`, so there is no entry-level choice to make and `op=` anywhere on that arm is refused by name. Step 3 now reads "The copy **replaces** the target's `Conditions` with the source's — `CopyFrom` is the only verb the zip issues, and there is no merge variant. That is why step 2 says to read each target first." (`references/write-side-recipes.md:40-42`.)

The earlier refutation stood on `housecarl_copy.seed_paths`, whose text says the zip is "where `op=Merge` and `op=ReplaceAll` are your choice of merging into the target's entries or replacing them" — on the live tool as well as in the capture. That sentence is the product defect this finding exposes, and it is now the only place the spelling survives. Left for Aaron: whether to file it against `seed_paths`.

.claude/skills/dialogue-authoring/references/dialogue-branch.md
Avick3110: Folded in 6bc843c0. Verified: `CreateTools.cs:140` publishes it on `records[].editorid` ("a FormLink value can reference it as '@<editorid>'"), `:167` on `records[].ops[].value` with a dialogue example — `ops:[{field_path:'Topic', value:'@MyTopic'}, {field_path:'PreviousDialog', value:'@MyTopic_L1'}]` — and `WriteEngine.cs:3350` carries the sigil. The capture is not wrong so much as blind here: `dev/skill-validation/eval-2026-09/surface/tools.md` records top-level parameter descriptions only, so the nested `records[]` member text where `@editorid` lives never entered it. That is what the earlier refutation read as absence.

The one-call route is back in all three places, your wording in `references/dialogue-branch.md:59-64` (same call, declared after the topic, `StartingTopic` set to the topic's `@editorid`), the same in `SKILL.md:158-163`, and `references/dialogue-flow-model.md:89-92` for the PNAM chain. The two-call route stays named as the alternative in each.

.claude/skills/dialogue-authoring/references/dialogue-flow-model.md
Avick3110: Folded in 26f2e555, scoping this file the way the body does rather than dropping the carve-out. All three:

- `:44-51` — order settles it "in a topic whose lines are told apart by their conditions", then states the stop plainly: the rule does not reach the generic greeting subtypes, Belethor's six vanilla `Hello` lines carry no `Random` flag and cycle in game, how the engine picks among eligible greetings is not something this skill or houseCARL can tell you, so never call a bottom-appended `Hello` line starved.
- `:71-74` — the monologue bullet now says one Info answers a topic, never a run of them, and that in a *conditioned* topic it is the first eligible one. The authoring advice it exists for is unchanged.
- `:120-122` — same scoping, pointing back at the section above.

The file now agrees with `SKILL.md:88-97` and with what the changelog claims was fixed, and eval E5 reads one rule wherever the reader lands.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in acc074e2, your wording. Row 1c now reads "`Branch` (BNAM) cannot be derived — set it yourself on a **`Custom`** topic (generic topics carry none, and a missing one there is not a defect)". Verified against `src/housecarl-core/DialogueValidate.cs:445`, which fires only for `Subtype == Custom` and says the topic plays fine in game, and against `references/dialogue-branch.md:21-23`. The compression had dropped the scoping, which on this body's own fast path pointed the reader at setting a Branch on a vanilla `Hello` DIAL.

Avick3110:
Fold pass over Aaron's four findings: all four folded, none refuted.

- **BNAM** (acc074e2) — row 1c now scopes "set it yourself" to a `Custom` topic and says a missing Branch on a generic topic is not a defect, matching `DialogueValidate.cs:445` and `references/dialogue-branch.md:21`.
- **The copy zip** (cb16a276) — `ApplyTools.cs` stamps every generated op `CopyFrom`, so recipe A now says the copy replaces the field it names and there is no merge variant. The `op=Merge`/`op=ReplaceAll` sentence the earlier refutation stood on is `housecarl_copy.seed_paths`' own text, and is now the last place that spelling survives — left for Aaron whether to file it.
- **`@editorid`** (6bc843c0) — published on `CreateTools.cs:140` and `:167` and resolved by `WriteEngine.cs:3350`; the capture carries top-level parameter text only, so the nested `records[]` member where it lives never entered it. The same-call sibling reference is back in the body, `references/dialogue-branch.md` and the PNAM chain, with the two-call route kept as the alternative. Open question 4 is answered.
- **The greeting carve-out** (26f2e555) — `references/dialogue-flow-model.md` asserted first-eligible-wins unscoped at three places while the body scoped it; all three now carry the scoping and the greeting stop.

Package build clean (leak-check clean, 12 skills / 252 markdown files across both trees), `dotnet build` clean, 1,858 tests passed, generator `ci-all` 128/128.

Issue #617, opened 2026-09-07
[Bug] place and nif_inspect accept a raw mods path; refuse it in one sentence


Avick3110:
`housecarl_place` and `housecarl_nif_inspect` accept a raw path into the MO2 mods directory; npc-appearance-copy's prose ban is the only thing stopping a caller from writing past the VFS. The rule for errors is one sentence saying what went wrong and what to try, so this is a refusal, not a guard: a raw mods path is refused with the mod-folder address form to use instead.

Found by the skill eval (npc-appearance-copy D5.40). After 2.0 unless promoted (DECISIONS 2026-09-07, rewrite ruling 6).

PR #646, opened 2026-09-08
Rewrite the npc-appearance-copy skill: finish at the repointed mesh

Avick3110:
Wave 3 of the skill rewrite, for `npc-appearance-copy` only. This is the rebuild the approved document asks for. The flow the skill taught was three calls and finished at the placement, and that is not where the job finishes: `housecarl_place` renames the FaceGen file without touching the bytes inside it, so the placed head shape's `tex[6]` still names the *donor's* tint and the copy renders correctly for exactly as long as the donor mod stays installed. Step 4 now writes that repoint — `housecarl_nif_set` with `op="set_path"`, `texture_slot="6"`, `target=<the head shape>` — and re-reads the mesh, and the verification list gained a check that fails when it has not been done. Three factual corrections go with it. `housecarl_copy` does not refuse a copy from a `Traits`-templated donor: `seed_paths` refuses on a path's *shape*, decided on the declared type, while a seed the source leaves unset **clears** the target's, so the `Configuration.TemplateFlags` read is the only thing between the caller and a blank face reported as a success (the shipped body said the opposite). The clone lane's `Race` refusal is not unconditional — it fires on a race that would have to be stripped, not on every race. And the provider a placement names is the one whose bytes match the record you copied, decided by inspecting each candidate with `housecarl_nif_inspect mod=`, `*winner` included; for an NPC defined in `Skyrim.esm` the old rule ("carry the folder of the arm whose plugin appears in the FaceGen path") yields `Data`, the vanilla face. Around those: the frontmatter gains the `compatibility` line and a description that states the boundary against `housecarl:facegen-diagnostics` where the host can actually read it; the four Race cases and the provider test move behind one pointer at `references/race-and-provider-cases.md` (the skill shipped no reference layer); verification splits into what this session can check — including `housecarl_check` on the not-yet-enabled patch, which both audit runs reached for and neither got from the skill — and what only the caller can do; the bare `nif_inspect` / `asset_status` spellings and the positional mesh path become the published names and `mesh_paths=`; the fold-outs listed in the ledger leave (F2, F13, F27–F31, F51); and `evals/eval_set.json` becomes `evals/evals.json` on the published field names, with a without-skill arm, three model tiers, a `dist/` run and the audit's Lydia task as the output-quality arm. The body is 13,400 characters, about 30% shorter than the 19,033 it replaces.

No issue is named by the rewrite document as closing, so there is no `Closes` line. #617 (place / nif_inspect refusing a raw mods path) is named in the body as a guard that does not exist yet, per ruling 6 — it stays open.

**Decided as follows.**

- **Q1 — `housecarl_apply`'s op vocabulary.** Step 2 needs `Set` (the clone's display `Name`) and `Add` (the `Female` bit), and the captured surface publishes no op enum: the only spellings anywhere in `surface/tools.md` are `Remove`, `Merge`, `ReplaceAll` and — new in the 2026-09-08 capture — `CopyFrom`. I wrote `Set` and `Add`, verified against the server's own verb table rather than invented: `WriteVerbs.All` is `Set, Add, Remove, SetAtIndex, InsertAtIndex, ReplaceAll, Merge, CopyFrom` (`src/housecarl-core/WriteVerbs.cs:85-86`), and `Add`/`Remove` on a flags-enum leaf is the `Female` case exactly (`src/housecarl-core/CorpusRulebook.cs:781-790`). The alternative was leaving the clone lane with no instruction at all. **The surface gap is real and unfixed**: `WriteVerbs.AllRecital` exists so a description can print the vocabulary, and no parameter in the capture does. Worth an issue; I did not file one.
- **Q2 — `into=` means two things.** Said plainly in the body, at Step 4: `housecarl_nif_set.into` takes an existing houseCARL-owned **mod folder** name, while `housecarl_copy`, `housecarl_apply` and `housecarl_place` take the patch plugin's filename. The Notes bullet is scoped to those three. Whether the surface should be made consistent is yours.
- **Q3 — does the boundary clause cede the dark-after-copy case?** Written as the document drafts it, and the eval file records the tension rather than hiding it: case 3 ("cloned an npc yesterday and now she's got a pitch black face") stays `should_trigger: true` and is re-adjudicated on the next run. If it routes to `facegen-diagnostics` again, the row gets reclassified rather than the description widened.
- **Q4 — "enabled, not sorted" against a placed asset.** Decided by your own ruling of 2026-09-08 ("the two 'enabled + sorted' parameter descriptions become 'enabled' in the same PR"), so the body does not carry a sorting instruction: the caller's list says enable the mod, and nothing else. The rule itself is out of this body entirely — it is F2, in ServerInstructions.
- **Q5 — the `compatibility` wording.** Written as the majority form: "Requires the houseCARL MCP server and a configured Mod Organizer 2 instance." **This is a batch pick, not mine** — confirm it once and the other eleven PRs paste the same string.

## Judgement calls worth your eye

- **The reference file carries less than the document drafted, because FOLD-IN moved under it.** `REWRITE.md` §5 puts two passages in `references/race-and-provider-cases.md`: the four Race cases (`SKILL.md:46-53`) and the two-arm `from_source=` idiom (`:55-60`). The ledger was revised on 2026-09-08 after that document was written — **F27 widened from `:61-70` to `:55-70`**, taking F13's second arm, and F13 now "keeps the Race branches alone" (`FOLD-IN.md`:233-242, :409). The brief says not to re-fold anything routed elsewhere, so `:55-70` is deleted against what `housecarl_copy.from_source` already carries and the reference file holds the Race cases plus the corrected provider test (C3), which is a fresh correction rather than a fold. Same reason the `SKILL.md:190` pointer is simply gone: F51, added the same day, retires it.
- **Body size.** The document budgets 11,700 characters across ten sections; the body lands at 13,400 (13,565 bytes — roughly 3,400–3,600 tokens by bytes÷4 and chars÷3.7, since no tokenizer is installed here). That is 15% over the budget and 28–32% inside the 5,000-token ceiling the budget existed to serve. What is over is call blocks: six worked calls at 300–600 characters each, which the section budgets were written before. I would rather hand you a measured number than cut a call block to hit an estimate.
- **The pointer is named once.** Ruling 2 says every shipped file is named once; the outline mentions the pointer in three sections. It is written once, in the overview, carrying all three read conditions, and Steps 1 and 3 refer back to it without repeating the filename.
- **The description says "three separate calls" while the body has four steps.** That string is the approved one and I did not touch it; the three count the things that *move* (records, inline values, files), and Step 4 repoints a file already placed. If you want them to agree, the description is the cheaper edit.
- **One stale changelog clause fixed in passing.** The wave-0 deletion entry said the never-copy rule for a copied NPC "is now stated in `npc-appearance-copy`". This PR removes that sentence from the body (the rule is F3, in ServerInstructions, always on), so the clause now points at what actually carries it.
- **No regrade in this session.** The brief asks for the build, the PR and no review pass of my own, so the five-angle regrade the wave-1 PRs carry is not here.

Build and checks, all run from the worktree: `scripts/build-plugin.ps1` green — skill leak-check clean, 12 skills / 251 markdown files across both trees, so the pointer scan saw the new `references/` file and the deleted `evals/eval_set.json`; `dotnet build housecarl.sln -c Release` 0 errors; `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` 1793 passed, 0 failed; generator `ci-all` 128/128.

## Regrade

Five-angle regrade of this branch, assembled in `dev/skill-validation/eval-2026-09/npc-appearance-copy/REGRADE.md` from the grader files in `regrade/D1.md` … `D5.md`.

| Angle | Rows | Conforms | Partial | Fails | n/a | Fails: factual-or-dead-call / packager-only / other |
|---|---|---|---|---|---|---|
| D1 — discovery and packaging | 42 | 37 | 1 (D1.28) | 0 | 4 | 0 / 0 / 0 |
| D2 — frontmatter | 55 (52 + 3 added) | 42 | 1 (D2.19a) | 0 | 12 | 0 / 0 / 0 |
| D3 — loading and invocation | 33 | 26 | 1 (D3.26) | 0 | 6 | 0 / 0 / 0 |
| D4 — body, files, instruction style | 64 | 38 | 3 (D4.43, D4.50, D4.52) | 0 | 23 | 0 / 0 / 0 |
| D5 — validation and standing context | 51 | 33 | 12 | 0 | 6 | 0 / 0 / 0 |
| **Total** | **245** | **176** | **18** | **0** | **51** | **0 / 0 / 0** |

**PASS** against the regrade gate in `REWRITE.md` §9a. No grader recorded a **fails** row on any angle, so the zero-fail thresholds on D2 and D3 and the zero factual-or-dead-call thresholds on D4 and D5 are all met; D1 has no fails either, and its single partial, D1.28 release tagging, is packager-only and excluded by ruling 8. Of the eighteen partials, four are excluded by name in §9a (D1.28, D2.19a, D3.26, D5.40), eleven of D5's twelve are "authored, not yet run" rows that convert on the §9b and §9c runs rather than on another edit, and the three D4 partials are body-quality items — D4.50 (every fenced block is input-only, both readback blocks deleted by F27) is a regression this rewrite introduced.

One item is open rather than closed: `op: "Add"` and `op: "Set"` (`SKILL.md:125`, `:138`) are values `surface/tools.md` never publishes — Q1 above. If either spelling is wrong, Step 2 carries two dead calls and D5.13 turns into a factual-or-dead-call fail, which would fail the gate. The surface publishing no enum is the only reason it grades partial today.


.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,189 @@ housecarl_copy(
...
 
-**A donor in a switched-off MO2 mod carries its assets too.** Name that mod in `source_provider=` and its copy is read off disk — loose first, then that mod folder's own archives — and the result says the source was not enabled. The mod does not have to be switched on for the placement, and switching it on is not a step. Still do not compose a path into the mods folder by hand: that is a guess wearing a path, it silently places the wrong bytes when the folder name is not what you assumed, and it cannot reach a file inside the donor's archives at all.
+```
+housecarl_nif_set(
+  mesh_path    = "<the placed mesh, Data-relative>",
+  op           = "set_path",
+  texture_slot = "6",
+  target       = "<the head shape's name, as nif_inspect prints it>",
+  path         = "<the copy's own tint path>",
+  into         = "<the mod folder the placement wrote>")
```
Avick3110: **Step 4 refuses every time as written — `mod=` is missing.**

With `mod=` omitted, `nif_set` resolves `<the placed mesh>` through the VFS. `housecarl_place` wrote it into a fresh houseCARL mod folder, houseCARL never edits `modlist.txt`, and `AssetResolver` only sees `Mo2Composition.EnabledMods` — so nothing in the active order provides that path. This skill's own verification leaves enabling to the caller ("Verification — what only the caller can do", item 1), so the folder is still off-order at Step 4. `LoadOrderService.NifSet` (src/housecarl-mcp/LoadOrderService.cs:1421-1426) then returns `ABSENT — no active mod or BSA provides '<rel>', so there is no copy to edit.`

Concretely: copy an NPC, place both FaceGen files into `MyFollower`, run Step 4 exactly as printed — the repoint never happens, and the skill's headline correction ("it is finished when the placed mesh points at the copy's own tint") is unreachable.

Naming the folder is precisely what `NifPick` → `TryResolveOffOrderProvider` exists for, and it is the rule this same body states two sections earlier: "`mod=` is not optional. … `mod=` reaches a mod MO2 is not loading."

```suggestion
housecarl_nif_set(
  mesh_path    = "<the placed mesh, Data-relative>",
  op           = "set_path",
  texture_slot = "6",
  target       = "<the head shape's name, as nif_inspect prints it>",
  path         = "<the copy's own tint path>",
  mod          = "<the mod folder the placement wrote>",
  into         = "<the mod folder the placement wrote>")
```

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,189 @@ housecarl_copy(
...
+```
 
-One thing does not widen: with `source_provider=` **omitted**, resolution still sees only the mods MO2 loads. A switched-off donor's file is reachable because you named it, never by houseCARL finding it.
+Then re-read it — `housecarl_nif_inspect(mesh_paths = ["<the placed mesh>"], sections = "paths")` —
```
Avick3110: **The re-read omits `mod=` too, so verification item 4 can never pass.**

Same mechanism as the `nif_set` call above: `nif_inspect` with no `mod=` takes the VFS winner, and the folder holding the edited mesh is not enabled yet, so `LoadOrderService.NifInspect` (src/housecarl-mcp/LoadOrderService.cs:1314-1324) returns `ABSENT — no active mod or BSA provides this mesh path.` There is no reading at which "the re-read of the placed mesh shows the copy's own tint in slot 6" can be satisfied.

Worse, if the placement folder ever *is* enabled while the `nif_set` output landed somewhere else, the winner is the un-repointed placed copy — the read would confirm the donor's tint and the caller would still see a green check. This body already states the rule at "`mod=` is not optional"; the re-read is the one call that breaks it.

```suggestion
Then re-read it — `housecarl_nif_inspect(mesh_paths = ["<the placed mesh>"], sections = "paths", mod = "<the mod folder the edit landed in>")` —
```

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,189 @@ housecarl_copy(
   patch         = "MyFollower")
 ```
 
-**The seed set is these four fields** because they are the appearance fields that are a record link or a list of record links. Everything else that makes up a face is inline data and rides Step 2. The list is safe to extend when you have a reason: the shape is judged on the field's *declared* type, so a field the donor happens to carry none of is still accepted, while a path that is not a field, or whose entries are structures rather than links (`Factions`, `Perks`, `Items`), is refused by name. Those belong in Step 2's zip, where you choose between merging into the target's entries and replacing them.
+**The seed set is those four fields** because they are the appearance fields that are a record link
```
Avick3110, line 63: **Deleting the `from_source=` two-arm passage leaves the overhaul case with no instruction, and the default silently masters the overhaul.**

The body no longer mentions `from_source=` in Step 1 at all, so the printed call runs on the default `['winner']`. Per `housecarl_copy`'s own rule (src/housecarl-mcp/CopyTools.cs:37): a link is internalized only when it is defined in a plugin being copied *away from* — `from=`'s defining plugin plus every plugin named in `from_source=`, and "'winner' names no plugin, so it adds none."

Concrete failure, and it is the description's own headline use ("borrowing a face from an overhaul", eval cases 2 and 8): `from = "000A2C94:Skyrim.esm"` for a Lydia overhauled by Bijin. Only `Skyrim.esm` is being copied away from, so Bijin's head parts resolve in the active order and are **kept and mastered**. The readback then reports the source is a master and the "standalone" hard-depends on Bijin — exactly what Verification item 1 tells the caller to treat as a failure, with no remedy anywhere in the body.

The PR body routes this to `housecarl_copy.from_source`'s tool description under F27, but that description teaches the internalization *rule*, not that this flow needs `from_source=['TheOverhaul.esp','<the defining plugin>']` — and Verification item 1's "every source named" and Step 2's `from_source:` field now both reference a mechanism the body never introduces. One sentence naming the ordered two-arm form would close it.

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,189 @@ housecarl_copy(
...
+  op           = "set_path",
+  texture_slot = "6",
+  target       = "<the head shape's name, as nif_inspect prints it>",
+  path         = "<the copy's own tint path>",
+  into         = "<the mod folder the placement wrote>")
+```
 
-One thing does not widen: with `source_provider=` **omitted**, resolution still sees only the mods MO2 loads. A switched-off donor's file is reachable because you named it, never by houseCARL finding it.
+Then re-read it — `housecarl_nif_inspect(mesh_paths = ["<the placed mesh>"], sections = "paths")` —
+and check that slot 6 names the copy's own tint. Note that `into=` here takes an existing
```
Avick3110: **The `into=` divergence this note asserts does not exist — both spellings work on all four calls.**

Q2 in the PR body decides this against the code. `nif_set`'s `into=` goes through the same `ResolvePatchModFolder` → `ResolveOwnedPatchFolder(into, needEsp: false, …)` every rider uses, whose own comment reads: "a renamed houseCARL patch folder is found by the .esp it holds or by its new name, so every rider's into= behaves exactly like a record into=" (src/housecarl-mcp/LoadOrderService.cs:7938-7949). And `housecarl_place`'s `into=` takes a folder name too — "for two patches sharing a filename, pass the mod-folder name here instead" (src/housecarl-mcp/PlaceTools.cs:47).

The cost is not cosmetic: the Notes section tells the caller to pass "`into=` the same plugin filename" to keep one reviewable artifact, and this sentence tells them that will not work for Step 4 — so a reader following both ends up hunting for a folder name they were told the plugin filename cannot substitute for. Either drop the note or restate it as "either spelling works on every call".

.claude/skills/npc-appearance-copy/references/race-and-provider-cases.md
```diff
@@ -0,0 +1,55 @@
...
+## Choosing the provider a placement names
+
+The rule is **the provider whose bytes match the record you copied**, and it is decided by reading,
+not derived.
+
+- **Do not take it from the plugin in the FaceGen path.** For an NPC defined in `Skyrim.esm` that arm
+  is the game's own `Data` layer — the vanilla face, not the face you just copied. The plugin in the
+  path names which master *defines* the NPC; it says nothing about whose bytes render.
+- **Read each candidate.** `housecarl_asset_status` on the FaceGen path lists every provider that has
+  a copy. Inspect them one at a time —
```
Avick3110: **The candidate enumeration comes back empty in exactly the case this file exists for.**

`housecarl_asset_status` answers for the mods MO2 loads. The deleted SKILL.md passage said so outright ("If the donor's own mod is switched off, every path only it provides reads back **absent** there"), and SKILL.md still keeps the caveat at Step 3 — but this file, which now *owns* the provider test, drops it.

Concretely: a switched-off donor (the flow's own supported case — "a switched-off donor needs no switching on") provides the FaceGen path and nothing else does. `asset_status` lists no provider, the reader has no candidate to inspect, and the procedure dead-ends. The `nif_inspect mod=` reads below would have worked; only the enumeration step fails. Add the one-line caveat: a provider `asset_status` cannot see is still reachable by name, so the donor's own mod folder is always a candidate.

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -1,32 +1,53 @@
 ---
 name: npc-appearance-copy
 description: >-
-  Copy an NPC's face onto another NPC, or clone one as a standalone, via `housecarl_copy` — the seed field set, the Race exclusion, the tint/morph bundle, and the FaceGen mesh+tint carry. Use for a standalone follower, borrowing a face from an overhaul, or a copied NPC rendering dark. Load before the copy — the wrong seed set writes a blank face.
+  Copies one NPC's face onto another, or clones an NPC as a standalone, via housecarl_copy —
+  records, inline face values, and the FaceGen files, which are three separate calls. Use for a
```
Avick3110: **"three separate calls" is the claim this PR exists to refute.**

The body's opening correction is that three calls finish at the placement, that stopping there leaves the head reading its colour out of the donor's mod, and that the fix is a fourth call. The description is the only text a host reads at load time, and it advertises the three-call flow — including to a session that never loads the body.

The PR body defends it as "the three count the things that *move*", but the same sentence ends "which are three separate calls", which is a call count, not a count of moved things. Since the description also carries "the copy is not done when the patch is written", the cheap fix is to make the count agree with it: `records, inline face values, and the FaceGen files, plus a repoint of the placed mesh`.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Refuted — the ledger routes this passage away, and the surface carries the two-arm form verbatim.

FOLD-IN F27 was widened from `:61-70` to `:55-70` on 2026-09-08, taking F13's second arm: "one passage, one destination, and the surface carries it" (`FOLD-IN.md` F27; F13 now "keeps the Race branches alone"). REWRITE §5 deviation 1 records the resolution in the open — ruling 4 wins, what the surface already carries is not re-carried, in the body or behind a pointer. The brief for this branch says not to re-fold anything routed elsewhere, so the body cannot introduce it and neither can the reference file.

The premise is also wrong on the facts. `housecarl_copy.from_source` does not teach only the internalization rule; it publishes this exact idiom, in this exact shape: "Naming several is how you read a look from an override patch while its records come from the defining plugin: from_source=['Override.esp','Donor.esp']" (`surface/tools.md:77`). That is the ordered two-arm form, with the override first, for the overhaul case by name. A caller composing a `housecarl_copy` call reads that parameter.

What remains true is that Step 2's `from_source:` field and Verification item 1 both lean on it. They lean on a parameter the caller is reading anyway, which is what ruling 4 chose.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Refuted — the description text is fixed by the rewrite document, verbatim.

REWRITE §3 prints the frontmatter block and then the resolved string under "Final description text", ending "...records, inline face values, and the FaceGen files, which are three separate calls." It is measured there (483 characters), graded against D2.21/D3.4/D2.26/D3.3, and its clause list is designed: the boundary clause against `housecarl:facegen-diagnostics`, and the load mandate's added consequence "the copy is not done when the patch is written", which is the clause the audit asked for and the one that carries the fourth call's warning into the text a host reads at load time.

Editing the count would deviate from the approved string on a branch whose whole job is to execute it, and the D2 rows in the regrade are graded against that string. If the count should read four, that is a change to REWRITE §3 and to the batch's description pass, not a fold on this PR.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in 6fda9603. `mod = "<the mod folder the placement wrote>"` on the `nif_set` call, `into = "<that same folder>"`, and a sentence saying why: the placement's folder is not enabled yet — enabling it is the caller's last step, per the verification split — so nothing in the active order provides the placed mesh and the read refuses ABSENT (`LoadOrderService.cs:1421-1426`). The named-provider arm is answered ahead of that return, so the off-order folder is reachable by name.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in 6fda9603. The re-read now carries `mod = "<that same folder>"` too, so verification item 4 reads the copy the edit landed in rather than a VFS winner that either does not exist yet or is the un-repointed placed file.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in c61252ba. The divergence is not there: `nif_set`'s `into=` goes through `ResolvePatchModFolder` → `ResolveOwnedPatchFolder(into, needEsp: false, …)`, whose by-plugin arm finds the owned folder holding `<stem>.esp` (`LoadOrderService.cs:8547-8550`), and `housecarl_place.into` says the same in its own description. The sentence is gone and the Notes bullet now covers all four calls: pass `into=` the same patch plugin's filename and they accumulate into one artifact. This answers the PR body's Q2 against the code rather than by assertion.

.claude/skills/npc-appearance-copy/references/race-and-provider-cases.md
Avick3110: Folded, in 3a413e26. The enumeration bullet now says `asset_status` lists every provider MO2 loads, so a switched-off donor is listed by nobody and the path reads absent — and that it is still a candidate: name its mod folder and read it anyway. The `nif_inspect mod=` reads below were already right; only the step that produced the candidate list was.

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,196 @@ housecarl_copy(
...
+  mesh_path    = "<the placed mesh, Data-relative>",
+  op           = "set_path",
+  texture_slot = "6",
+  target       = "<the head shape's name, as nif_inspect prints it>",
+  path         = "<the copy's own tint path>",
+  mod          = "<the mod folder the placement wrote>",
+  into         = "<that same folder>")
+```
+
+**`mod=` is not optional here either**, and for a second reason: the placement's folder is not enabled
```
Avick3110: **The ABSENT claim only holds on the `new_editorid=` lane.**

`LoadOrderService.NifSet` returns ABSENT only when `place.Sources.Count == 0` (`src/housecarl-mcp/LoadOrderService.cs:1420-1427`). In the `target=` lane the placement destination is the *existing* target NPC's FaceGen path — `facegeom/<target's defining master>/00<target hex>.nif` — which the active order already provides (vanilla BSA at minimum), so `place.Sources` is non-empty and a `mod=`-less read resolves the **winner**, not the mesh Step 3 placed.

Concrete failure: `target=` lane, donor and target same race, so the copied head parts keep their EditorIDs (Step 1 says so at `:86`) and the target's original mesh carries shapes of the same name. An agent that trusts this sentence omits `mod=`; nif_set reads the target's original winning mesh, `target=` matches a shape, `set_path` verifies green, and the edited bytes are written to `Path.Combine(rf.OutputDir, rel)` — the same relative path inside `into=`, silently **overwriting the donor mesh Step 3 placed**. The result renders the target's own head geometry under the copy's tint, and every readback says success.

The instruction (always pass `mod=`) is right; the stated reason is what is wrong, and it is the reason that tells a reader when they may skip it. Suggest scoping it — "in the clone lane nothing in the active order provides the placed path, so a read without `mod=` refuses ABSENT; in the `target=` lane it silently returns the target's *original* winning mesh instead, which is worse".

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,196 @@ housecarl_copy(
...
+Read the donor first and split those six in two, because a donor rarely carries all of them:
 
 ```
 housecarl_apply(
   bundle      = [ ...the members the donor HAS... ],
-  assignments = [{ target: "<the new/target FormID>", from: "<donor FormID>",
+  assignments = [{ target: "<the target FormID>", from: "<donor FormID>",
                    from_source: "<the plugin the appearance came from>" }],
-  ops         = [ { formid: "<target>", field_path: "<a member the donor LACKS>", op: "Remove" }, ... ],
+  ops         = [{ formid: "<target>", field_path: "<a member the donor LACKS>", op: "Remove" }],
```
Avick3110: **`op: "Remove"` does not clear `TintLayers`, and `TintLayers` is a bundle member a donor routinely lacks.**

Of the six members, `TintLayers` is a `List<ITintLayerGetter>` (mutagen-reference `records.jsonl`, `Npc`); the other five are nullable substructs/values or non-nullable scalars. `VerbLegality` admits a keyless list `Remove` (`src/housecarl-core/CorpusRulebook.cs:389-392`) because it "legally falls back to remove-by-value" — and the apply arm then does `lt.GetMethod("Remove", ...).Invoke(list, new[] { Coerce(req.Value!, elem) })` (`src/housecarl-core/WriteEngine.cs:2785-2801`). With no `value` on the op that is `Coerce(null, ...)`, an unnamed throw; and when the donor's list is genuinely absent the earlier arm raises the named `Remove on list 'TintLayers': the collection is absent (no elements) — nothing to remove`.

Either way the op never clears anything, and because `housecarl_apply` is all-or-nothing (stated two lines below, at `:111`) the whole Step 2 write fails — so a donor with no tint layers, exactly the split this step exists for, cannot be completed by following the block. The whole-clear verb for a list is `ReplaceAll` with an empty `values`.

Related: the rewrite also dropped the old body's sentence that `CopyFrom` **refuses** an unset source, which was the only place a reader learned that naming all six members refuses rather than merely being untidy. "All-or-nothing" now carries that weight alone and does not say it.

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,196 @@ housecarl_copy(
...
-asset paths the copied records reference (this call does NOT place them — check each with
-housecarl_asset_status, then place what you keep with housecarl_place; a path only the
-mod you read FROM provides reads as absent in asset_status if MO2 does not load that mod, and is
-still placed by naming it in source_provider=):
-  - textures\actors\character\...\hair.dds
+housecarl_nif_inspect(npc = ["<donor FormID>"], sections = "paths", mod = "<the donor's mod folder>")
 ```
 
-**Decide before you place.** Run `housecarl_asset_status` on each path and read the provider chain. Carry a path only if its bytes would **vanish with the donor** — if another enabled mod still supplies it, the file already resolves and copying it just adds a redundant override. Say which paths you skipped and why; a silent skip and a deliberate one look identical afterwards.
+Paths here are **Data-relative with forward slashes** —
```
Avick3110: **"forward slashes" is wrong about what houseCARL prints, and Step 4 writes that string into the mesh verbatim.**

`FaceGenPath.cs:47-48` composes both derived paths with backslash separators, and every tool description spells the example the same way (`NifTools.cs:57`, `:65`, `:201`). The texture paths `sections="paths"` reads back come straight out of the NIF's own texture set, which is backslash-delimited by Bethesda convention. Nothing here uses forward slashes. Input tolerance ("forward or back slashes both fine") is a separate fact and is already the second sentence.

Why it matters beyond tidiness: Step 4 passes `path=` straight through to `tex.Content = op.Path` (`src/housecarl-core/NifService.cs`, the `SetPath` slot arm) — no normalization. A reader who takes this sentence as a spelling instruction bakes `textures/actors/character/facegendata/facetint/...` into slot 6 of the shipped mesh. That resolves for a loose file and fails for a BSA-packed one, so it renders correctly on the authoring machine and dark for anyone whose textures are archived — the exact class of bug Step 4 was added to close.

Suggest: "Paths here are Data-relative and printed with backslashes; both tools accept either slash on input, and neither takes a drive-rooted path. Write the path back into the mesh exactly as it was printed."

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,196 @@ housecarl_copy(
...
+entries and replacing them.
 
-**A seed the donor leaves unset clears the target's.** Deliberate, and the same rule as Step 2's partition: a copy that leaves the target's own head parts or worn armor sitting under the donor's face produces a face assembled from two records. The readback marks those fields `cleared` rather than reporting them copied.
+**A seed the donor leaves unset clears the target's.** Deliberate, and the same rule as Step 2's
+partition: a copy that leaves the target's own head parts or worn armor under the donor's face
+produces a face assembled from two records. The readback says `cleared` rather than copied.
 
-**`Race:refuse`** is there because a race is not an appearance subtree. Walking into one pulls the skeleton, the sibling races, the whole racial frame — so a race is either kept as an ordinary link or the copy stops and tells you. The exclusion only fires when the race is *inside the source universe*: defined in the donor plugin itself, or not resolving in your active load order.
+**`Race:refuse` is the default** because a race is not an appearance subtree — a walk into one pulls
```
Avick3110: **`Race:refuse` is not a default — the tool has none.**

`exclude_types` is `string[]? = null` (`src/housecarl-mcp/CopyTools.cs:42`) and `CopyTools.cs:85-108` builds the exclusion list purely from what was passed; omit the parameter and the walk runs with *zero* exclusions. `housecarl_copy`'s own description calls RACE "the standing 'refuse' **case**" — a convention the caller supplies, not a value the server applies. `references/race-and-provider-cases.md:9` repeats the same wording ("is the standing default").

Concrete failure: an agent composing its own call rather than pasting Step 1's block — a second copy for a different target, a retry after a refusal — reads "is the default" and omits `exclude_types=`. The walk enters the RACE record and internalizes the skeleton and the sibling races into the patch, or trips the runaway cap (`CopyTools.cs:219`). That is precisely the cost the same paragraph goes on to describe.

Compounding it, the diff deletes the Common-mistakes row that guarded this — old `:168`, "Seeding `Race`, or dropping the `Race:refuse` exclusion | The walk pulls the skeleton and sibling races instead of a face" — so the assertion loses the only line that contradicted it. Suggest "`Race:refuse` is the standing choice, and it is yours to pass — the tool applies no exclusion you do not name", and restoring the table row.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: **The description re-asserts the claim this PR exists to correct.**

"which are three separate calls" is contradicted twelve lines later by "**A fourth call finishes it.**" (`:25`) and by the whole of Step 4. The PR body notes this and defends the count as "the things that move" — but the description is not read alongside the body: it is the only text in context when the model decides whether to load the skill, and it is what a session that never loads the skill carries away. A model that reads "three separate calls" and then the body's own table (three rows, one per call) has been told twice that the flow ends at the placement, which is the dark-face-after-uninstall bug the rewrite was written to close.

It also puts eval case 20 at risk in the other direction: that row expects the skill *not* to be picked for a bare repoint, and it is now a step in this body while the description still advertises three calls.

Cheapest fix is the description, as the PR body says: "records, inline face values, the FaceGen files and the tint path baked inside the placed mesh — four separate calls".

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -39,152 +60,196 @@ housecarl_copy(
...
+`housecarl_asset_status` in ONE call. Carry a path only if its bytes would vanish with the donor: if
+another enabled mod supplies it, the file already resolves and a copy is a redundant override. Say
+which paths you skipped. If the donor's mod is switched off, every path only it provides reads
+**absent** — the carry case, not the skip case.
 
 ```
 housecarl_place(assets = [
   { formid: "<the NEW FormID>", kind: "mesh",
-    source: "<the DONOR's FaceGen mesh path>", source_provider: "<the donor's mod>" },
+    source: "<the donor's FaceGen mesh path>", source_provider: "<the donor's mod folder>" },
```
Avick3110: **The call block contradicts the corrected provider rule it sits above.**

The block hardcodes `source_provider: "<the donor's mod folder>"` on both members; twelve lines down (`:181-184`) the rewrite's own correction says the provider is "the one whose bytes match the record you copied — **sometimes the VFS winner**, which `source_provider` spells `*winner`", decided by inspecting each candidate.

An agent copies the block, not the paragraph. In the case the correction was written for — an NPC defined in `Skyrim.esm` whose face was read through an overhaul, where the reference file says the FaceGen path's own arm resolves to `Data` and the matching bytes are the VFS winner's — the block's placeholder names the wrong provider, and the placement lands the wrong mesh under the copy's FormID with no refusal.

Suggest making the placeholder carry the rule rather than the answer, e.g. `source_provider: "<the provider chosen below>"`, or moving the two-sentence provider test above the block.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in 442d47ce. The reason is scoped by lane now. On `new_editorid=` nothing in the active order provides the clone's FaceGen path, so a `mod=`-less read hits the `place.Sources.Count == 0` arm and refuses ABSENT (`LoadOrderService.cs:1420-1427`). On `target=` the placed path is the target's own FaceGen path, which the vanilla archives provide, so the read takes the winner — the target's original head, whose shape names still match — and the verified bytes go back to the same relative path inside `into=` (`LoadOrderService.cs:1487`), over the mesh Step 3 placed. The body says outright that the second case is the worse one. Current text at `SKILL.md:218-224`.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in 4a26d410 and cb622ce9 — with a different verb than the one suggested, because `values=[]` refuses on this field too.

`TintLayers` holds modeled elements (`TintLayer`), so a `ReplaceAll` carrying `values=` lands in the modeled-element arm: "'TintLayers' holds modeled elements (TintLayer); ReplaceAll has no build-from-parts input on this call" (`CorpusRulebook.cs:888-897`). The whole-clear that works is `ReplaceAll` with `composes: []` — legal by name at `CorpusRulebook.cs:483-484` ("the modeled twin of ReplaceAll values=[]") and carried through the wire mapper at `LoadOrderService.cs:7802-7812`. The Step 2 block now shows that op, scoped to a donor that carries no tint layers, and a new paragraph at `SKILL.md:120-123` splits the clear: `Remove` for the nullable members (`FaceMorph`, `FaceParts`, `TextureLighting`), `ReplaceAll composes: []` for `TintLayers`, and `Weight`/`Height` are not nullable so a donor always carries them.

The related half is folded too, at `SKILL.md:118-119`: the zip expands to one `CopyFrom` op per bundle path (`ApplyTools.cs:248`) and `CopyFrom` refuses an unset source rather than clearing it (`WriteEngine.cs:2058`), so naming a member the donor lacks fails the whole write. "All-or-nothing" no longer carries that alone.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in e673982f. `SKILL.md:163-166` now reads "Data-relative and printed with backslashes", shows the derived path with backslashes, keeps the input-tolerance sentence, and adds the write-back instruction: a path goes back into a mesh exactly as it was printed, because Step 4 stores the string verbatim (`NifService.cs:712`, `tex.Content = op.Path`). Verified against `FaceGenPath.cs:44-47` (both derived paths composed with backslashes) and `NifTools.cs:56-57`, `:200`.

Noting the deviation out loud: REWRITE §4's preamble says every game path is written "Data-relative with forward slashes and labelled as such", and the body no longer does. The citation behind that instruction is the slash-tolerance phrase on `mesh_paths` / `asset_paths`, which is about input and is still stated here; what the document asked for on output is not what houseCARL prints.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in c79635d6, both halves. `SKILL.md:75-78` now reads "`Race:refuse` is the standing choice, and it is yours to pass — the tool applies no exclusion you do not name, so a call that omits `exclude_types` walks with none", and the reference file's opening line carries the same correction (`references/race-and-provider-cases.md:9-10`), since it repeated "standing default". Verified: `exclude_types` is `string[]? = null` (`CopyTools.cs:42`) and the exclusion list is built only from what was passed (`CopyTools.cs:85-108`); the tool's own description calls RACE the standing 'refuse' case, a convention, not an applied value.

The common-mistakes row is restored at `SKILL.md:235`: "Seeding `Race`, or omitting `exclude_types` because you expect a `Race` default | There is none: the walk enters the race and pulls the skeleton and sibling races".

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in 8d3b59dd — your call overrides the earlier refusal of the same finding (blind-review thread on this line, answered "the string is fixed by REWRITE §3"). `SKILL.md:4-6` now reads "records, inline face values, the FaceGen files and the tint path baked inside the placed mesh, which are four separate calls", keeping the original clause frame and every other clause untouched. Resolved string is 525 characters (was 483), inside the 1,024 D2.21 measures against; the listing entry is 544 against Claude Code's 1,536.

Two things this leaves open for you, and they are why the first pass refused rather than edited:

- REWRITE §3 prints the three-call string as "Final description text" and measures it at 483. The shipped string now differs from the approved document, and the D2/D3 rows in the regrade were graded against the document's. Amend §3, or say the word and I revert the commit.
- Eval case 20 (a bare repoint, `should_trigger: false`) now faces a description that names the repoint. Its `evidence` already says to watch the row; I left the case as authored rather than reclassifying it before it has been run.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded, in a0a871bb. Both members now read `source_provider: "<the provider chosen below>"` (`SKILL.md:181`, `:183`), so the block hands the reader to the test rather than an answer the test can contradict. The paragraph under it is unchanged and still carries the rule and the `*winner` spelling (`SKILL.md:189-195`), and the reference file holds the worked test.

PR #647, opened 2026-09-08
Rebuild the bulk-record-jobs skill against the 2.0 surface

Avick3110:
Wave 3 of the skill rewrite: `bulk-record-jobs`, built from the approved rewrite document and its verifier (147 rows checked, 0 overturned, so the document is executed as written). The skill named ten tools that no longer exist and told the reader to raise `max_chars` so a big enumeration would render in one call — the opposite of what the tool does, and the exact advice that made the audited control run die on the client idle timeout with nothing to salvage. The body is rebuilt against the 2.0 read surface: one census (`project={"form":"aggregate"}` with `counts_only=true`), one spill lane with two knobs (`max_chars` caps the render, an over-ceiling result spills whole to a JSONL artifact, `to_file=` is the explicit spelling and refuses `offset=`, and `offset=` re-scans rather than seeks), the completeness check re-derived against the artifact's own manifest and one `epoch` across every window, the two scope clauses, the loop-killer map re-pointed at the 2.0 verbs, and the deliverable contract. It went from 20,842 characters to 9,964 (~2,700 tokens against the 5,000-token window), and the catalogue/link-graph recipes, the Creation-Kit crafting conventions and the patch rebuild are now three `references/` files the body names once each with the condition for reading it. The eval set becomes `evals/evals.json` with a without-skill arm on every case and the audited with/without catalogue task as the output-quality arm. No issue is named as closing.

**Checks.** `scripts/build-plugin.ps1 -PluginTreeOnly` — leak-check clean, 12 skills / 253 markdown files across both trees, `evals/` stripped and the three `references/` files shipping. `dotnet build -c Release` clean; tests 1793 passed, 0 failed (`tier!=bridge`); generator `ci-all` 128/128.

- **Q1 — does "must appear in `surface/tools.md` as written" bar a nested member or an enum value? Decided for the build, left for Aaron as a rule.** Nothing in the tree names a token with 0 hits in the capture, with one deliberate exception: the three count keys `winner` / `type` / `defined_in` are written as `group_by` values, because §4's outline asks the body to carry them and because a census section that will not spell the value it tells you to pass is not usable. `group_by` itself is rendered; its value set is only in the raw `tools-list.json`. Everywhere else the unrendered member is described rather than named — the definitions half of the `plugins` scope points at `housecarl_records.plugins` for the spelling (F32 rules it unteachable), and the identity-annotation rule states the cost and the remedy without naming the sub-parameter. If Aaron rules the other way on Q1, the count keys are the one edit.
- **Q2 — patch rebuild as a `references/` file or its own skill? Decided as follows: a reference file,** per §5, so the twelve-document count stands. It is written as the separate job D4.6 grades it — its own trigger, input and success criterion — so splitting it out later is a move, not a rewrite.
- **Q3 — how to score D4.42. Left for Aaron.** The build follows ruling 1: MCP tools bare on both hosts, siblings `housecarl:<skill>`, and the body carries the one line saying which form holds for which host. D4.42 as written cannot be met.
- **Q4 — how hard the identity-annotation rule should be. Decided as follows: soft, and stated as a cost.** The body says what the annotation costs (one winner lookup per link, 66,856 of them on the audited catalogue, and the abort that followed) and gives the two ways out — bound the set, or read identity separately and join locally. It does not forbid the annotation, because #582's answer has only partly landed.
- **Q5 — keep the subagent trigger with the fan-out contract gone? Decided as follows: kept as a pure trigger.** The clause stays in the description verbatim in substance; the body no longer claims the skill can guarantee its own presence inside a subagent, and what a subagent is handed is the deliverable schema in the body.
- **Q6 — the eval file's home. Decided as follows on the document's own inference:** `evals/evals.json` lives in the authoring tree and the runs read `dist/housecarl/skills/bulk-record-jobs/SKILL.md`. The packager strips `evals/` from every shipped copy and the leak scan re-checks for it, so the file cannot ship and no body pointer names it. The file records the packaged path it reads.
- **Q7 — rename `-jobs`? Decided as follows: kept and recorded.** Renaming costs the continuity of the eval record and the audit pair, which are this skill's strongest evidence.

## Judgement calls worth a reviewer's eye

- `evals/eval_set.json` and `eval_results_2026-07-15.md` are deleted rather than left beside the new file: `evals.json` carries their gate, their scores, the 2026-07-28 contamination audit and the recorded soft edge, and two eval files in one directory is the drift the one-convention ruling exists to stop.
- The biped-slot query stays a **negative** near-miss, as §9c lists it, even though the skill that owned it is deleted — its expected output now names the flag test on the surface instead of a sibling skill. The routing table carries no pointer to a deleted sibling (F51).
- The body measures 9,964 characters against the outline's ~9,600 estimate (+4%), well inside the gate the estimate serves.
- `CreatedObject` and the other COBJ paths in the crafting reference are record-schema field paths, not tool parameters, so they are not expected in the tool capture.

## Regrade

Assembled in `dev/skill-validation/eval-2026-09/bulk-record-jobs/REGRADE.md` from the five per-angle
regrades of this branch's tree.

| Angle | Conforms | Partial | Fails | n/a | Criteria |
|---|---|---|---|---|---|
| D1 — discovery and packaging | 36 | 0 | 1 | 5 | 42 |
| D2 — frontmatter | 39 | 1 | 0 | 12 | 52 |
| D3 — loading decision and invocation control | 25 | 2 | 0 | 6 | 33 |
| D4 — body, supporting files, instruction style | 43 | 3 | 1 | 20 | 67 |
| D5 — evals, safety, composition | 36 | 7 | 0 | 7 | 50 |
| **Total** | **179** | **13** | **2** | **50** | **244** |

**PASS** against the regrade gate. Two fails, neither factual-or-dead-call: D1.28 (release tagging)
is packager-only and is the row ruling 8 excludes from the gate; D4.42 (bare MCP tool names) is a
ruled deviation under ruling 1 — the server publishes the bare names, so the criterion as written is
unmeetable, and the host note it requires is present at `SKILL.md:22-24`. D2 and D3 have zero fails;
D4 and D5 have zero factual or dead-call fails, with every tool and parameter name in the tree
resolving on the captured surface. The 13 partials are the rows the gate lists as recorded
disagreements or unmovable, plus D4.43's residual precision defects and the three D5 rows that ask
for executed numbers the eval file so far only specifies.


.claude/skills/bulk-record-jobs/SKILL.md
```diff
@@ -1,250 +1,142 @@
...
+overrides. The number decides the plan: 40 records and 40,000 records are not the same job.
+
+Write the deliverable schema down before any extraction starts, especially before handing work out.
+
+## Getting the whole set out — one lane, not two
+
+There is one lane, with two knobs. `max_chars` is a ceiling on the **render**, never on the result:
+an over-ceiling result spills in full to a server-side JSONL artifact whose line 1 is the manifest
+(query echo, row schema, epoch), and the response names the file. So the move for a big enumeration
+is a **small** `max_chars`, not a large one.
```
Avick3110: **A small `max_chars` does not get the whole set out — it spills the `limit=` window.** The auto-spill and `to_file=` are not "the same lane": in `RecordsTools.cs` the scan's `effLimit` is `wantFile || derivedSelection ? int.MaxValue : ... (limit <= 0 ? 500 : limit)`, so only `to_file=` lifts the window. The auto-spill writes whatever `limit=` let through.

Concrete: follow this section literally on the audited NPC catalogue — `types=["NPC_"]`, `project.form="fields"`, a small `max_chars`, no `to_file=`, no `limit=` — and the artifact holds **500 rows of 66,856**. The response says so itself (`Artifacts.AppendSpillText`): `spilled: the returned WINDOW (500 rows of 66856 total matches)` … `the 66,356 matches beyond limit= are in NO file — page with offset=, raise limit=, or use to_file= for the full result.` Pinned by `RecordsArtifactTests.AWindowedAutoSpillSaysWindowAndNeverClaimsTheCompleteResult`.

The skill's own "Prove it whole" check catches the 500-row file after the fact, but the advice that produced it is what a reader acts on first. Suggest making the completeness claim conditional: `max_chars` bounds the render and spills what the window held; **`to_file=` is what captures the complete result** (it is the branch that sets the scan uncapped), or raise `limit=` alongside a small `max_chars`.

.claude/skills/bulk-record-jobs/SKILL.md
```diff
@@ -1,250 +1,142 @@
...
+written there, only the manifest rendered inline. It refuses `offset=`: the artifact is never a
+window. Re-enter it later with `formids=["@<path>"]` or `where=["formid in @<path>"]`, epoch-checked.
+
+```
+housecarl_records(types=["NPC_"], format="json",
+                  project={"form":"fields", "fields":["EditorID","Name","Race"]},
+                  to_file="C:/work/npc-winners.jsonl")
+```
 ```
+{"manifest":{"total":66856,"rendered":66856,"epoch":"7f3a1c","schema":["formid","EditorID","Name","Race"]}}
```
Avick3110: **The manifest example names three keys the artifact does not have, and gets the row schema wrong.** `ResultArtifact.Manifest.WriteTo` writes line 1 as a *flat* object: `housecarl_artifact`, `tool`, `query`, `identity`, `row_schema`, `sort`, `row_count`, `total`, `type_counts`, `epoch`, `created`, `notes`. There is no `manifest` wrapper, no `rendered`, and the schema key is `row_schema`, not `schema`.

The schema *contents* are wrong too. For a `form="fields"` scan artifact `Artifacts.WriteCrossQuery` stamps `["formid","runtime_formid","type","editorid","winner","override_depth","source","matches?","fields"]` — the requested paths are **nested under `fields`**, never top-level columns.

And the inline render shown (`wrote C:/work/... — 66,856 rows`) is not what comes back. Text gives `spilled: complete result (66856 rows) -> <path>` plus `manifest: rows=… identity=… epoch=…`; `format="json"` (as this very example passes) gives a `"spilled"` object with `path`, `reason`, `row_count`, `total`, `complete`, `identity`, `row_schema`, `sort`, `epoch`.

Failure: a reader writes the obvious follow-up — `jq '.manifest.total'` on line 1, or a script reading `row.EditorID` / `row.Race` — and gets `null` on every row, then either reports the deliverable as empty or, worse, ships rows whose columns silently never populated. The same wrong shape is repeated at line 90 in "Prove it whole", which is precisely the check meant to catch an incomplete file.

.claude/skills/bulk-record-jobs/references/patch-rebuild.md
```diff
@@ -0,0 +1,70 @@
...
+on the delta form. Naming `fields` scopes the comparison. Spill the result with `to_file=` when the
+set is large — the delta form consumes every scan match, so the scan terms are the cost bound.
+
+## 3. Forward the baseline, then edit the forwarded copy
+
+`housecarl_forward` copies a named plugin's whole record verbatim; `housecarl_apply into=` the same
+patch then edits that forwarded body rather than re-resolving the load-order winner. That pair is
+the pinned stale-winner bypass, and it is the whole reason the rebuild works while an old winner
+still sits on top. Where the contested winner is a plugin another tool regenerates, the never-copy
+rule in the server's standing instructions decides `source=`: name the authored plugin.
```
Avick3110, line 39: **"The never-copy rule in the server's standing instructions" does not exist.** `Program.cs`'s `options.ServerInstructions` is the server's only standing instruction text, and it never mentions regenerated output — no Reqtificator, DynDOLOD, Synthesis, or never-copy clause. Nor does any tool description: grepping `ForwardTools`/`ApplyTools`/`CopyTools`/`WriteSentences` for those terms returns nothing.

That rule lived in `tool-output-awareness`, which is not among the eleven skills this branch ships (`.claude/skills/` has: bulk-record-jobs, dialogue-authoring, facegen-diagnostics, kid-authoring, mutagen-reference, npc-appearance-copy, oar-authoring, papyrus-reference, skse-plugin-authoring, skypatcher-authoring, spid-authoring).

Failure: this is the one step in the rebuild that decides `source=` when the contested winner is a generated plugin. A reader hits step 3, goes looking for the named authority, finds nothing anywhere in the shipped surface, and either improvises or copies the Reqtificator's regenerated body into the authored patch — the exact outcome the sentence exists to prevent. The PR body claims this tree carries no pointer to a deleted sibling (F51); the pointer survived by being re-aimed at an authority that does not carry the rule. Either state the rule inline here, or drop the sentence.

.claude/skills/bulk-record-jobs/SKILL.md
```diff
@@ -1,250 +1,142 @@
...
+| Read records one call each | one call with `formids=` — the cost is the LIST's length, not the window's, so pass fewer ids rather than paging |
+| Read whole records just to label FormIDs | `project={"form":"summary"}` — identity off a gathered read; `form="identity"` is the dearest row on the lane, not a free one |
+| Annotate every link on a full render | bound the set, or read identity separately and join locally |
+| Parse `path = token` text back into JSON | `format="json"` — same tokens, stable document, accounting in-band |
+| Read two plugin versions and subtract by hand | `project={"form":"delta"}` with `source=` the subject and `versus=` the reference pole; either may be an on-disk, unticked plugin |
+| Walk a link chain call by call | `walk=` — the traversal is a SELECT term, and any reading form consumes what it reaches |
+| Filter matches after the read | `where=` — comparisons, flag tests, quantified list steps, `*parent`, one `->` link step, `formid in @<file>` |
+| Re-type another version's field values into ops | `housecarl_apply` with an `ops[]` entry naming `from_source=` — the field is taken from the plugin you name, not from the winner |
```
Avick3110: **`from_source=` alone does nothing — it is a `CopyFrom` member, and naming it on any other verb is silently ignored.** `ApplyOp.from_source` is documented `op='CopyFrom' only`, and the engine enforces that by construction, not by refusal: `LoadOrderService.PrepareCopyFromSources` returns immediately unless an edit has `Verb == "CopyFrom"`, and `WritePatchBuilder`'s source resolution is only reached on that verb. `ApplyTools` validates `from=` against the verb (`from= … is only valid with op='CopyFrom'`) but has **no equivalent check for `from_source=`**.

Failure: a reader follows this row and writes `{formid, field_path: "BasicStats.Damage", from_source: "OldPatch.esp", value: "14"}`. The op takes the default `Set`, `from_source` is dropped on the floor, and the patch lands with `14` — the value the caller typed, not OldPatch's version — with no refusal and no note. That is a silently wrong write into a rebuilt patch, which is the one thing the whole recipe is trying to get right. Omitting `value` too gets a missing-value refusal that names the wrong problem.

The verb is what makes the row true, so name it. Same omission in `references/patch-rebuild.md` lines 45–46.

.claude/skills/bulk-record-jobs/SKILL.md
```diff
@@ -1,250 +1,142 @@
...
+| Add list elements one op at a time | `composes` on one op — appends N elements, or rebuilds the whole list (`composes=[]` clears it) |
```
Avick3110: **`composes` is a verb-dependent input, and the row states neither verb — so `composes=[]` as written is refused, not a clear.** `WriteVerbs` binds `composes` to `Add` (append each) and to `ReplaceAll` (clear, then append each); the empty-list clear is `ReplaceAll` only. `BulkPrimitivesWave3Probe`'s P8a arm probes exactly this pair: `Verb="ReplaceAll", Composes=Array.Empty<StructInput>()` is the clear, and `Verb="Add", Composes=Array.Empty<StructInput>()` is a named refusal.

Failure: a reader takes "appends N elements" as the whole contract, writes `{formid, field_path: "Entries", op: "Add", composes: []}` to empty a leveled list, and the call refuses with nothing written; the default `Set` verb refuses too, since `Set` takes `value`/`compose`, not `composes`. Name the two verbs — `Add` appends, `ReplaceAll` rebuilds and `composes=[]` under `ReplaceAll` clears. Same wording in `references/patch-rebuild.md` lines 47–48.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The `bulk-record-jobs` skill is rebuilt against the 2.0 tool surface.** Every call it plans is now a
+  `housecarl_records`, `housecarl_apply`, `housecarl_forward` or `housecarl_check` call; the ten 1.x tool
+  names it named are gone, as are `winner_fields=`, `group_by=` as a top-level argument, `plugin_a=`/`plugin_b=`
+  and the rest of the retired spellings. The big-enumeration advice is reversed to match what the tool does:
+  `max_chars` is a ceiling on the render and an over-ceiling result spills in full to a JSONL artifact, so the
+  move is a small `max_chars` rather than a large one, `to_file=` is the explicit spelling of that same lane,
```
Avick3110: **This carries the same wrong claim as `SKILL.md:48-51` into user-facing release notes.** "An over-ceiling result spills in full to a JSONL artifact, so the move is a small `max_chars` rather than a large one" holds only when the window is not the bound. `RecordsTools` sets the scan uncapped (`int.MaxValue`) for `to_file=` and derived selections; every other path spills whatever `limit=` (default 500) let through, and the response says `spilled: the returned WINDOW (… of … total matches)` … `beyond limit= are in NO file` (`Artifacts.AppendSpillText`, pinned by `RecordsArtifactTests`).

"`to_file=` is the explicit spelling of that same lane" is the part that misleads: it is the *only* spelling that lifts the window. A user reading the changelog and reaching for a small `max_chars` on a big enumeration gets a 500-row file. Suggest: `max_chars` bounds the render and spills the window; `to_file=` captures the complete result.

.claude/skills/bulk-record-jobs/references/patch-rebuild.md
Avick3110: Refuted: the never-copy rule *is* in the server's standing instructions. `Program.cs`'s `options.ServerInstructions` carries it verbatim at the end of the string (`src/housecarl-mcp/Program.cs:165-170`):

> "NEVER COPY GENERATED OUTPUT: a tool re-derives it on its next run, so it is never copied into an authored patch and never the base for authored work — 'Requiem for the Indifferent.esp'; 'PGPatcher.esp' / 'PG_1.esp'; 'DynDOLOD.esm' / 'DynDOLOD.esp' / 'Occlusion.esp'; a mod folder holding 'NPC_Token.json' or 'ParallaxGen_Diff.json'; a Synthesis, TexGen or xLODGen output folder. A mod that only NAMES a generator ('… Resources', '… Fixes', a downloaded patch) is an INPUT it consumes, not output: patch it normally."

The comment above it in the same file says why it is there: "The three cross-skill facts the skill rewrite folds here: which runtime framework owns a job, that a write wins nothing until enabled, and the never-copy rule for generated output." That is exactly the move that let `tool-output-awareness` be deleted — the rule went into the instructions every session gets, so the reference file points at a live authority, not a deleted sibling. F51 holds: the pointer names the server's instructions, and the instructions carry the rule and the generator list.

.claude/skills/bulk-record-jobs/SKILL.md
Avick3110: Folded in 52849f9. Confirmed at `RecordsTools.cs:1300` — only `wantFile` or a derived selection lifts `effLimit` to `int.MaxValue`; every other path spills whatever `limit=` let through, and `Artifacts.AppendSpillText` says so in two different sentences depending on `m.Total == m.RowCount`.

The section now keeps the small-`max_chars` move as the render knob and hands completeness to `to_file=`: "what spills is what `limit=` let through (default 500), and the marker says which: `spilled: complete result (N rows)` against `spilled: the returned WINDOW (N rows of T total matches)`, whose matches beyond `limit=` are in no file at all", then "`to_file=` is what captures the **complete** result — it is the spelling that lifts the window". "Prove it whole" check 2 is re-pointed at the same fact: `row_count` equals `total`, short of it means the file holds a window.

.claude/skills/bulk-record-jobs/SKILL.md
Avick3110: Folded in 52849f9. Confirmed against `ResultArtifact.Manifest.WriteTo` (flat: `housecarl_artifact`, `tool`, `query`, `identity`, `row_schema`, `sort`, `row_count`, `total`, `type_counts`, `epoch`, `created`, `notes`) and `Artifacts.WriteCrossQuery`'s detail-row schema (`formid, runtime_formid, type, editorid, winner, override_depth, source, matches?, fields`).

The example now shows the real `format="json"` document — `"total"`, an empty `"matches"`, and the `spilled` object with `path` / `reason` / `complete` / `row_count` / `total` / `identity` / `epoch` / `row_schema` — followed by the sentence that kills the `row.EditorID` script: the requested paths sit under `fields`, never as top-level columns. "Prove it whole" now names the flat manifest keys and its `head -1` line shows `{"housecarl_artifact":1,"tool":"housecarl_records",…}`.

.claude/skills/bulk-record-jobs/SKILL.md
Avick3110: Folded in 52849f9 — the row now reads `op="CopyFrom"` + `from_source=`, with "the pole is CopyFrom's; any other verb refuses it by name". Same in `references/patch-rebuild.md`, plus the `CopyFrom` takes-no-`value` rule.

One correction to the stated failure, since it changes what the row has to warn about: `from_source=` on a non-CopyFrom op is **not** dropped on the floor. `LoadOrderService.MapFromPlugin` refuses it by name — `"{where}: from_source is only valid with op=CopyFrom (got op={verb})."` (`LoadOrderService.cs:7716`) — and `BulkPrimitivesWave3Probe` pins exactly the op you describe: `Verb="Set", Value="5", FromPlugin=masterName` → refused with "only valid with op=CopyFrom" (`:386-389`), with a mis-cased `copyfrom` refused the same way at the mapper. So the outcome is a refused call, not a silent write of the typed value — the reason to name the verb is that the row as written is a dead call, which is enough.

.claude/skills/bulk-record-jobs/SKILL.md
Avick3110: Folded in 52849f9. `WriteVerbs.On` binds `composes` to `Add` ("appends a new element at the END") and, on a composed-element list, to `ReplaceAll` ("clears the list, then appends each"); `Set` is absent from the list branch by construction, and the P8a arm refuses `Add` with an empty `composes`.

The row now reads: `op="Add"` appends N elements, `op="ReplaceAll"` rebuilds the whole list and `composes=[]` under it clears — with the note that the default `Set` has no list element to mean. `references/patch-rebuild.md` carries the same three clauses.

plugin/CHANGELOG.md
Avick3110: Folded in 52849f9, with the same split the body now makes: "`max_chars` is a ceiling on the render and an over-ceiling result spills to a JSONL artifact, so the move is a small `max_chars` rather than a large one; what spills is the window `limit=` allowed, and `to_file=` is what captures the complete result". The "explicit spelling of that same lane" phrasing is gone — it was the half that made the two knobs sound interchangeable.

.claude/skills/bulk-record-jobs/SKILL.md
```diff
@@ -1,250 +1,154 @@
...
+```
+"total": 66856, "matches": [],
+"spilled": {"path":"C:/work/npc-winners.jsonl", "reason":"to_file", "complete":true,
+            "row_count":66856, "total":66856, "identity":"formid", "epoch":"7f3a1c",
+            "row_schema":["formid","runtime_formid","type","editorid","winner","override_depth",
+                          "source","matches?","fields"]}
+```
+
+The row schema is the artifact's own, not the fields you asked for: the requested paths sit **under
+`fields`** on each row, never as top-level columns — `row.fields["EditorID"]`, not `row.EditorID`.
```
Avick3110: `row.fields["EditorID"]` is not the shape the artifact writes. `JsonWire.WriteFieldsArray` opens `fields` as a JSON **array** (`w.WriteStartArray("fields")`, src/housecarl-mcp/JsonWire.cs:434) whose elements are leaf objects `{path, value?, note?, display?, link?}` — not an object keyed by field path. The same writer fills the artifact rows and the inline json rows, so this holds in both.

Scenario: a reader follows this line and scripts `row.fields["EditorID"]` over `npc-winners.jsonl`. Every lookup is `undefined`, and because the row otherwise parses fine the deliverable comes out with an empty column rather than an error — exactly the silently-wrong output the deliverable contract below exists to prevent.

The correct read is to find the element whose `path` matches and take its `value`:

```suggestion
`fields`** on each row as an ARRAY of leaf objects `{path, value, …}`, never as top-level columns —
find the element whose `path` is `"EditorID"` and read its `value`, not `row.EditorID`.
```

.claude/skills/bulk-record-jobs/SKILL.md
```diff
@@ -1,250 +1,154 @@
...
+| Read whole records just to label FormIDs | `project={"form":"summary"}` — identity off a gathered read; `form="identity"` is the dearest row on the lane, not a free one |
+| Annotate every link on a full render | bound the set, or read identity separately and join locally |
+| Parse `path = token` text back into JSON | `format="json"` — same tokens, stable document, accounting in-band |
+| Read two plugin versions and subtract by hand | `project={"form":"delta"}` with `source=` the subject and `versus=` the reference pole; either may be an on-disk, unticked plugin |
+| Walk a link chain call by call | `walk=` — the traversal is a SELECT term, and any reading form consumes what it reaches |
+| Filter matches after the read | `where=` — comparisons, flag tests, quantified list steps, `*parent`, one `->` link step, `formid in @<file>` |
+| Re-type another version's field values into ops | `housecarl_apply` with an `ops[]` entry of `op="CopyFrom"` + `from_source=` — the field is taken from the plugin you name, not from the winner (the pole is CopyFrom's; any other verb refuses it by name) |
```
Avick3110, line 141: "any other verb refuses it by name" is not what the server does for `from_source=`. The only verb guard in `housecarl_apply` is on `from=` (the source RECORD): src/housecarl-mcp/ApplyTools.cs:148 refuses `from=` on a non-`CopyFrom` op. `from_source` is mapped straight through to `BulkOp.FromPlugin` at ApplyTools.cs:158 with no verb check, and `WritePatchBuilder` only ever consults `FromPlugin` inside its `CopyFrom` branch (WritePatchBuilder.cs:484, :556) — so on any other verb it is silently ignored, not refused.

Scenario: a rebuild op is written as `{formid, field_path, op: "Set", value: "12", from_source: "OldPatch.esp"}`, intending the old patch's value. The call is accepted, `from_source` is dropped, and the literal `value` is written — a wrong write that the skill has just told the reader the tool would have caught by name. (`from=` on a non-CopyFrom op *is* refused, so only the `from_source` half of the claim is wrong.)

.claude/skills/bulk-record-jobs/references/patch-rebuild.md
```diff
@@ -0,0 +1,72 @@
...
+
+## 4. Batch the re-application
+
+One `housecarl_apply` call carries the whole op list:
+
+- **Whole-field transplants**: an `ops[]` entry of `op="CopyFrom"` with `from_source="OldPatch.esp"`
+  takes that field's value from the plugin you name instead of from the winner — off-order sources
+  included. The pole belongs to that verb: on any other op it is refused by name, and `CopyFrom`
+  itself takes no `value`.
```
Avick3110, line 48: Same claim as SKILL.md:138, same problem: "on any other op it is refused by name" holds for `from=` but not for `from_source=`. src/housecarl-mcp/ApplyTools.cs:148 guards only `from=`; `from_source` is copied to `BulkOp.FromPlugin` (ApplyTools.cs:158) unconditionally, and `WritePatchBuilder` reads `FromPlugin` only in its `CopyFrom` branch (WritePatchBuilder.cs:484, :556), so a `Set`/`Add`/`ReplaceAll` op carrying `from_source` is accepted and the pole silently dropped.

Scenario: step 4's op list mixes a `CopyFrom` transplant with a `Set` that a reader also tags `from_source: "OldPatch.esp"`. The batch applies, the `Set` writes its literal `value` rather than the old patch's, and nothing in the response names the ignored pole — while this reference has told the reader the call would have refused. Since this is the rebuild recipe, that is a wrong patch shipped as a verified one.

(The `CopyFrom` takes no `value` half of the sentence is correct.)

.claude/skills/bulk-record-jobs/SKILL.md
```diff
@@ -1,250 +1,154 @@
...
+**Annotating every link is the trap that kills the job.** An identity per FormLink is one winner
+lookup per link — 66,856 of them on the catalogue above — and the client aborts on the idle timeout
+with no partial artifact to salvage. Bound the set, or read the identity catalogue separately and
+join it locally.
+
+## Prove it whole
+
+A persisted result is trustworthy only after these three checks, made **in the file**:
+
+1. Line 1 is the manifest — a flat object keyed `housecarl_artifact`, `tool`, `query`, `identity`,
```
Avick3110: The manifest is neither flat nor limited to these nine keys. `ResultArtifact.Manifest.WriteTo` (src/housecarl-core/ResultArtifact.cs:126-155) writes `query` as a nested object and `row_schema` as an array, and also writes `type_counts` (when the rows carried types), `created` always, and `notes` when the artifact owes the scoped-vs-winner or owned-child clause.

Scenario: check 1 is written as an assertion — a verification script asserts line 1 is a flat object whose key set is exactly the nine named. It fails on every real artifact (`created` is unconditional), and the reverse case is worse: `notes` is where the "these values came from the scoped body, not the winner" clause lives, so a reader told the manifest holds nine keys never looks for it and ships live-value claims off scoped-era numbers — the very thing the Scope truth section warns about.

Worth naming `created`/`notes`/`type_counts` and dropping "flat".

.claude/skills/bulk-record-jobs/SKILL.md
Avick3110: Folded in 4f1078d1. Confirmed: `JsonWire.WriteFieldsArray` opens `fields` with `w.WriteStartArray("fields")` and writes each leaf as its own object (`{path, value?, note?, …}`, plus `cells` on a folded row) — `src/housecarl-mcp/JsonWire.cs:307-335` — and the artifact rows go through that same writer, `Artifacts.WriteCrossQuery` calling `JsonWire.WriteReadRecord` per row at `src/housecarl-mcp/Artifacts.cs:179`. So the object-keyed read is empty in the file and in the inline json alike.

SKILL.md:73-76 now reads: the requested paths sit under `fields`, never as top-level columns, "and `fields` is an **array** of `{path, value}` leaves, not an object keyed by path. Take the element whose `path` is `"EditorID"` and read its `value`; `row.EditorID` and `row.fields["EditorID"]` are both empty on every row." Both wrong shapes are named, since the one this line previously taught is the one a reader would carry into the script.

.claude/skills/bulk-record-jobs/SKILL.md
Avick3110: Folded in a11c9827. Confirmed against `ResultArtifact.Manifest.WriteTo` (`src/housecarl-core/ResultArtifact.cs:126-155`): `query` is written with `WriteStartObject("query")`, `row_schema` with `WriteStartArray("row_schema")`, `created` is unconditional, and `type_counts` / `notes` are written when the rows carried types and when the result owes a clause. "Flat" and the nine-key set were both mine, not the writer's — the rewrite document only asks the check to name the manifest's query echo, row schema and `epoch` (REWRITE.md:351).

Check 1 now reads: "`housecarl_artifact`, `tool`, `query` (an object), `identity`, `row_schema` (an array), `sort`, `row_count`, `total`, `type_counts` when the rows carried types, `epoch`, `created`, and `notes` when the result owes a clause — and it is the thing you read, not the chat response's summary. Read `notes`: the scoped-vs-winner clause lands there." The last sentence is your second scenario: `Artifacts.CrossQueryNotes` puts exactly that clause in `notes` (`src/housecarl-mcp/Artifacts.cs:219-227`), so a reader told the manifest held nine keys would never have looked for it.

.claude/skills/bulk-record-jobs/SKILL.md
Avick3110: Refuted: `from_source=` on a non-`CopyFrom` op is refused by name — the guard is one layer below `ApplyTools`, in the service's op mapper.

`ApplyTools` does only the `from=` check you found (`src/housecarl-mcp/ApplyTools.cs:148-152`) and then hands the wire ops to `svc.ApplyEdits`. `ApplyEdits` maps every op through `MapEdit` before the write gate (`src/housecarl-mcp/LoadOrderService.cs:5334-5345`), and `MapEdit` calls `MapFromPlugin` for every op (`:7706`), which is:

```csharp
if (!string.Equals(verb, "CopyFrom", StringComparison.Ordinal))
{
    if (!string.IsNullOrWhiteSpace(op.FromPlugin))
        error = $"{where}: from_source is only valid with op=CopyFrom (got op={verb}).";
    return null;
}
```
(`LoadOrderService.cs:7721-7731`). A named error out of `MapEdit` goes into `problems`, and any problem refuses the whole call with `NO patch written` (`:5343-5345`) — so `WritePatchBuilder` is never reached, and the `FromPlugin`-only-in-the-`CopyFrom`-branch reading at `WritePatchBuilder.cs:484,:556` is downstream of a call that cannot get there.

Your scenario is pinned as a refusal in the probe: `BulkPrimitivesWave3Probe.cs:386-389` runs `Verb="Set", Value="5", FromPlugin=masterName` and checks the call fails with "only valid with op=CopyFrom"; `:392-394` pins a mis-cased `copyfrom` failing the same way (the compare is ordinal on purpose). The blind review made the same call earlier in the thread on comment 3954476893, and the reply there carries the same correction — the reason to name the verb in the row is that the row as written is a dead call, which is why it was folded in 52849f9 anyway.

So `SKILL.md:138`'s "(the pole is CopyFrom's; any other verb refuses it by name)" stands as written. Happy to be told otherwise if you want the sentence to say where the refusal comes from.

.claude/skills/bulk-record-jobs/references/patch-rebuild.md
Avick3110: Refuted, same as the reply on `SKILL.md:138`: "on any other op it is refused by name" holds for `from_source=` too. The guard is not in `ApplyTools` — it is in the service mapper every op passes through before the write gate. `ApplyEdits` maps each op with `MapEdit` (`src/housecarl-mcp/LoadOrderService.cs:5334-5345`), `MapEdit` calls `MapFromPlugin` (`:7706`), and `MapFromPlugin` refuses a non-`CopyFrom` verb carrying `from_source` by name: `"{where}: from_source is only valid with op=CopyFrom (got op={verb})."` (`:7725-7731`). A named error refuses the whole call — `NO patch written` at `:5343-5345` — so `WritePatchBuilder.cs:484,:556` is never reached.

Your step-4 scenario is the batch mixing a `CopyFrom` with a `Set` tagged `from_source`: that batch does not apply, it refuses with the per-op reason and nothing is written, which is what this file's own "every call is all-or-nothing" line at `references/patch-rebuild.md:55-56` says next. `BulkPrimitivesWave3Probe.cs:386-389` pins exactly that op (`Verb="Set", Value="5", FromPlugin=…` → refused, "only valid with op=CopyFrom").

The `CopyFrom` takes no `value` half is pinned in the same block (`:396-398`).

PR #649, opened 2026-09-08
Fold three of Aaron's 2026-09-08 calls

Avick3110:
Three of the six calls Aaron made on 2026-09-08 from the fold of his wave 1-3 reviews (`dev/DECISIONS.md`, the "Six calls" line). Calls 4, 5 and 6 are not in this PR.

**Call 1 — ruling 7 amended, and the two reference skills say what works offline.** Ruling 7 keeps the shared compatibility sentence on every skill; the amendment lets a skill add one second sentence saying what works offline. `papyrus-reference` and `mutagen-reference` each gain one, identical in shape: "The lookup is offline; only what reads the live load order needs them." That is Aaron's own phrasing from his review comment on PR #642, where he pointed out that the frontmatter gated a fully offline signature lookup on a configured MO2 instance, and it is true for both — papyrus-reference's availability checks call `housecarl_skse`, and mutagen-reference's schema lookup is a grep over bundled files while the record reads and writes laid against that schema are server calls. The shape follows `skse-plugin-authoring`, which already carries the shared sentence plus a second one in a folded scalar. Both skills' bodies already scoped this in prose; the frontmatter now says it on its own.

**Call 2 — the CID clause leaves the runtime routing.** No Container Item Distributor skill ships, and SkyPatcher's own container patcher does the job (`filterByContainers` with `addToContainers`, in the skill's `references/container.md`), so the clause routed a real SkyPatcher job to a skill that does not exist. It is removed from `skypatcher-authoring`'s description and from the RUNTIME DISTRIBUTION LAYERS sentence in `ServerInstructions` (FOLD-IN F1), and the routing now reads: a spell, perk, item, keyword, outfit or faction onto NPCs is SPID, best by group (faction, race, level, trait); a keyword onto item records is KID; a record's own fields, and an individual NPC, are SkyPatcher. `spid-authoring`'s description and its Scope paragraph carried the same CID line and are aligned; `skypatcher-authoring`'s Overview gains the individual-NPC half, and its description gains a container token. `kid-authoring`'s description carried no CID line, but its shipped `references/grammar-core.md` routing rows did and are aligned. The served instructions are now **3,173 characters**, up 40 from 3,133; `ServerInstructionsTests` records the measured figure and is updated with the delta.

The group/individual split is a **preference, not a prohibition** — Aaron's call is that SPID is best for groups and SkyPatcher better for NPCs individually, and SPID can still name one NPC by EditorID or FormID (`spid-authoring/references/filters.md` §1 and §2). Both skill bodies say it that way.

The three eval cases in `.claude/skills/skypatcher-authoring/evals/evals.json` that were added for this exclusion clause:

- **Case 27** ("distribute a keyword to every dagger in my load order") — **kept as should-not, reworded.** KID still owns a keyword onto item records, so the boundary it measures is unchanged; the expected output and evidence now say "item records" to match the routing's wording.
- **Case 28** ("give every bandit a spell at runtime, no esp") — **kept as should-not, rewritten to measure the new routing.** It used to grade against "forms onto NPCs is not a SkyPatcher job", which the new routing makes false — an individual NPC's edits are SkyPatcher's. It now grades on the group/individual line: "every bandit" is a group, which is SPID's half.
- **Case 29** ("put these items into a chest without an ESP") — **moved to the should-trigger arm.** With the CID clause gone this is SkyPatcher's job outright, and its own evidence already noted the problem: no CID skill ships, so the case graded on the decline alone because a routing expectation would have been unsatisfiable. On review, the description gained "putting items into a container or chest" in its use clause, because the trigger arm shows an agent descriptions only and the description carried no container token to pick on; and its second expectation moved from a reference file that arm never opens to the reasoning, matching the other should-trigger cases.

The arm counts in the suite's `cases` line move from 17/15 to 18/14; the 32 prompts are unchanged.

The same prompt appears in two sibling suites and both recorded a destination the routing no longer has: `spid-authoring` case 13 said NONE and `kid-authoring` T13 said CID. Both now say `skypatcher-authoring`, so the three suites agree.

**Call 3 — the npc-appearance-copy rewrite document matches what shipped.** `dev/skill-validation/eval-2026-09/npc-appearance-copy/REWRITE.md` §3 printed the description as three separate calls at 483 characters; the shipped skill says four, because the tint path baked inside the placed mesh is its own call. §3 now prints the description as it stands on the tip, measured at 525 characters (listing entry 544 against Claude Code's 1,536 cap, 35% of it), the D2.26a bullet quotes the four-part decomposition, and the section carries a one-line note saying it was amended 2026-09-08 on Aaron's review of PR #646. `dev/` is gitignored, so this change is not in the diff.

Package rebuilt with `scripts/build-plugin.ps1` — leak-check clean on both trees, 12 skills, 223 markdown files. `dotnet test --filter "tier!=bridge"`: 1890 passed, 0 failed. Generator `ci-all`: 128/128, `description-vocab-guard` green.

### DIFF
--- a/.claude/skills/kid-authoring/evals/evals.json
+++ b/.claude/skills/kid-authoring/evals/evals.json
@@ -183,13 +183,13 @@
    "id": "T13",
    "prompt": "add 5 gold and a healing potion to every merchant chest in my load order",
    "should_trigger": false,
-   "expected_output": "kid-authoring is NOT loaded; the query routes to skypatcher-authoring — its container patcher puts items into containers. Destination corrected 2026-09-08 on review of PR #649; the old answer named CID, which ships no skill in this tree.",
+   "expected_output": "kid-authoring is NOT loaded; the query routes to skypatcher-authoring — its container patcher puts items into containers.",
    "expectations": [
     "with-skill arm: the pick is not kid-authoring on at least 2 of 3 runs",
     "with-skill arm: the reasoning names why the target is out of KID's lane",
     "without-skill arm: removing kid-authoring does not change where this query lands"
    ],
-   "prior_evidence": "2026-07-28 run under the compressed descriptions: specificity row, routed to CID — items to containers"
+   "prior_evidence": "2026-07-28 run under the compressed descriptions: specificity row, routed to CID — items to containers. Destination corrected 2026-09-08 on review of PR #649; the recorded answer named CID, which ships no skill in this tree, and SkyPatcher's container patcher owns the job."
   },
   {
    "id": "T14",
--- a/.claude/skills/kid-authoring/evals/evals.json
+++ b/.claude/skills/kid-authoring/evals/evals.json
@@ -183,7 +183,7 @@
    "id": "T13",
    "prompt": "add 5 gold and a healing potion to every merchant chest in my load order",
    "should_trigger": false,
-   "expected_output": "kid-authoring is NOT loaded; the query routes to CID — items to containers.",
+   "expected_output": "kid-authoring is NOT loaded; the query routes to skypatcher-authoring — its container patcher puts items into containers. Destination corrected 2026-09-08 on review of PR #649; the old answer named CID, which ships no skill in this tree.",
    "expectations": [
     "with-skill arm: the pick is not kid-authoring on at least 2 of 3 runs",
     "with-skill arm: the reasoning names why the target is out of KID's lane",
--- a/.claude/skills/kid-authoring/references/grammar-core.md
+++ b/.claude/skills/kid-authoring/references/grammar-core.md
@@ -37,7 +37,8 @@ KID's target is **items** (object/base records). That's its lane in the distribu
 
 - keyword/form to an **item record** → **KID** (this skill)
 - spell/perk/item/keyword to an **NPC** → **SPID**
-- item into a **container** → **CID**
+- item into a **container** → **SkyPatcher** (its container patcher — `filterByContainers` with
+  `addToContainers`)
 - editing a record's **own fields** (damage, value, an NPC's stats) → **SkyPatcher**
 
 KID and SPID share an author and many idioms (the `~` FormID form, the `+`/`-`/`*` filter modifiers),
@@ -179,7 +180,7 @@ e.g., material or rarity tags from doubling up on the same item.
 |---|---|
 | add a keyword to **items** (weapon/armor/potion/…) | **KID** ← this skill |
 | give a spell/perk/item/keyword to **NPCs** | SPID |
-| put an item into a **container** | CID |
+| put an item into a **container** | SkyPatcher (its container patcher) |
 | change a record's **own fields** (damage, value, NPC stats) | SkyPatcher |
 
 When unsure, ask **what receives the keyword**: an item record → KID; an NPC → SPID.
--- a/.claude/skills/kid-authoring/references/grammar-core.md
+++ b/.claude/skills/kid-authoring/references/grammar-core.md
@@ -36,7 +36,8 @@ every launch, directly onto the in-memory record, so a KID mod is trace-free to 
 KID's target is **items** (object/base records). That's its lane in the distributor family:
 
 - keyword/form to an **item record** → **KID** (this skill)
-- spell/perk/item/keyword to an **NPC** → **SPID**
+- spell/perk/item/keyword to **NPCs** → **SPID**, best by group (faction, race, level, trait); for one
+  named NPC **SkyPatcher** is usually the better fit, though a single-NPC SPID line is valid
 - item into a **container** → **SkyPatcher** (its container patcher — `filterByContainers` with
   `addToContainers`)
 - editing a record's **own fields** (damage, value, an NPC's stats) → **SkyPatcher**
@@ -179,8 +180,9 @@ e.g., material or rarity tags from doubling up on the same item.
 | You want to… | Tool |
 |---|---|
 | add a keyword to **items** (weapon/armor/potion/…) | **KID** ← this skill |
-| give a spell/perk/item/keyword to **NPCs** | SPID |
+| give a spell/perk/item/keyword to **NPCs** | SPID (best by group; for one named NPC, usually SkyPatcher) |
 | put an item into a **container** | SkyPatcher (its container patcher) |
 | change a record's **own fields** (damage, value, NPC stats) | SkyPatcher |
 
-When unsure, ask **what receives the keyword**: an item record → KID; an NPC → SPID.
+When unsure, ask **what receives the keyword**: an item record → KID; NPCs → SPID, and one named NPC
+usually SkyPatcher.
--- a/.claude/skills/skypatcher-authoring/SKILL.md
+++ b/.claude/skills/skypatcher-authoring/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: skypatcher-authoring
 description: >-
-  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, or asking why a patch line isn't applying. Not for distribution — forms onto NPCs is SPID, keywords onto items is KID, items into containers is CID.
+  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, or asking why a patch line isn't applying. Not for distribution — distribution to NPCs by group (faction, race, class) is SPID, keywords onto item records is KID.
 compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
 ---
 
@@ -9,10 +9,11 @@ compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 
 
 SkyPatcher is an SKSE plugin that edits Bethesda records at runtime from plain-text INI files — no
 ESP or ESL is produced for the edit itself. This skill composes and reads those patch strings and
-places the file where SkyPatcher will read it. SkyPatcher edits a record's own fields; it does not
-distribute. Two neighbouring jobs are someone else's: what fields a record has at the Mutagen/xEdit
-level, with their types and writability, is `housecarl:mutagen-reference`; putting forms onto NPCs
-or keywords onto items is `housecarl:spid-authoring` / `housecarl:kid-authoring`. On a Codex flat
+places the file where SkyPatcher will read it. SkyPatcher edits a record's own fields, an individual
+NPC's included; it does not distribute. Two neighbouring jobs are someone else's: what fields a
+record has at the Mutagen/xEdit level, with their types and writability, is
+`housecarl:mutagen-reference`; distribution to NPCs by group — faction, race, class — or a keyword
+onto item records is `housecarl:spid-authoring` / `housecarl:kid-authoring`. On a Codex flat
 install these siblings are the bare folder names — `mutagen-reference`, `spid-authoring`,
 `kid-authoring`.
 
--- a/.claude/skills/skypatcher-authoring/SKILL.md
+++ b/.claude/skills/skypatcher-authoring/SKILL.md
@@ -12,9 +12,10 @@ ESP or ESL is produced for the edit itself. This skill composes and reads those 
 places the file where SkyPatcher will read it. SkyPatcher edits a record's own fields, an individual
 NPC's included; it does not distribute. Two neighbouring jobs are someone else's: what fields a
 record has at the Mutagen/xEdit level, with their types and writability, is
-`housecarl:mutagen-reference`; distribution to NPCs by group — faction, race, class — or a keyword
-onto item records is `housecarl:spid-authoring` / `housecarl:kid-authoring`. On a Codex flat
-install these siblings are the bare folder names — `mutagen-reference`, `spid-authoring`,
+`housecarl:mutagen-reference`; distributing a form to NPCs is `housecarl:spid-authoring`, which is
+*best* for groups — faction, race, level, trait — and can also name one NPC, though for a single NPC
+SkyPatcher is usually the *better* fit; a keyword onto item records is `housecarl:kid-authoring`. On
+a Codex flat install these siblings are the bare folder names — `mutagen-reference`, `spid-authoring`,
 `kid-authoring`.
 
 ## Route to the record type
--- a/.claude/skills/skypatcher-authoring/SKILL.md
+++ b/.claude/skills/skypatcher-authoring/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: skypatcher-authoring
 description: >-
-  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, or asking why a patch line isn't applying. Not for distribution — distribution to NPCs by group (faction, race, class) is SPID, keywords onto item records is KID.
+  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, putting items into a container or chest, or asking why a patch line isn't applying. Not for distribution — distribution to NPCs by group (faction, race, class) is SPID, keywords onto item records is KID.
 compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
 ---
 
--- a/.claude/skills/skypatcher-authoring/SKILL.md
+++ b/.claude/skills/skypatcher-authoring/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: skypatcher-authoring
 description: >-
-  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, putting items into a container or chest, or asking why a patch line isn't applying. Not for distribution — distribution to NPCs by group (faction, race, class) is SPID, keywords onto item records is KID.
+  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, putting items into a container or chest, or asking why a patch line isn't applying. Not for distribution — distribution to NPCs by group (faction, race, level, trait) is SPID, keywords onto item records is KID.
 compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
 ---
 
--- a/.claude/skills/skypatcher-authoring/evals/evals.json
+++ b/.claude/skills/skypatcher-authoring/evals/evals.json
@@ -9,7 +9,7 @@
     "output_quality": "The audit's own authoring task on the live order, run as a pair: control (no skill in the catalogue) against with-skill."
   },
   "method": {
-    "cases": "32 — 17 should-trigger, 15 should-not. The shipped 26 (14/12) survive with their queries unchanged; 6 are added for the description's changed clauses (three distribution-shaped negatives for the new exclusion clause, three positives naming the newly front-loaded terms).",
+    "cases": "32 — 18 should-trigger, 14 should-not. The shipped 26 (14/12) survive with their queries unchanged; 6 are added for the description's changed clauses (three distribution-shaped cases for the exclusion clause, three positives naming the newly front-loaded terms). Case 29 moved from should-not to should-trigger on 2026-09-08 when the 'items into containers is CID' clause was dropped from the routing; 27 and 28 stay should-not, reworded for the new routing.",
     "runs_per_query": 3,
     "split": "60/40 train/held-out",
     "without_skill_arm": "The same 32 queries against a menu that omits this skill, recording what the agent does unaided. It is the baseline the delta is taken against, and it is how a negative case is proved to be a near-miss rather than an unreachable prompt.",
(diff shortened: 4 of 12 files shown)


.claude/skills/spid-authoring/SKILL.md
```diff
@@ -22,10 +22,10 @@ from scratch each launch, so a SPID mod is trace-free to add or remove. This ski
 reads SPID lines, taking every form type, filter, modifier and value from `references/` — look a
 token up, never invent one, because a wrong token fails silently.
 
-**Scope.** SPID's target is NPCs. A keyword onto *item records* is `housecarl:kid-authoring`; a
-record's *own fields* are `housecarl:skypatcher-authoring`; items into *containers* is CID, which has
-no skill here. Siblings carry the `housecarl:` prefix on a Claude Code plugin install; a Codex flat
-install sees the bare folder name (`kid-authoring`).
+**Scope.** SPID's target is NPCs *by group* — faction, race, class. A keyword onto *item records* is
```
Avick3110: **The Scope narrowing contradicts this skill's own description and its own reference corpus.**

The frontmatter three lines up still says "targeting NPCs by faction, race, **level or trait**", and `references/filters.md:30` documents targeting an NPC's own **EditorID** (`BalgruufTheGreater`), with `references/grammar-core.md:265` stating outright that String and Form filters "can additionally target specific NPCs by EditorID/FormID". This Scope line now asserts SPID is group-only (faction, race, class) and hands *individual* NPCs to SkyPatcher.

Concrete misbehaviour: a session asked "give Balgruuf a healing spell at runtime, no ESP" reads this paragraph, declines, and routes to skypatcher-authoring — even though SPID does that job and this skill's own `filters.md` gives the exact line. Same for "give every NPC above level 30 a perk at runtime": `level` is not in the group list here, so the paragraph does not claim it either, while the description does.

Suggest keeping the group/individual split as the *routing preference* rather than a capability claim, and keeping the filter axes aligned with the description.

.claude/skills/skypatcher-authoring/evals/evals.json
```diff
@@ -241,26 +241,26 @@
...
-      "expected_output": "Route to the SPID skill — forms onto NPCs is not a SkyPatcher job.",
-      "expectations": ["does not pick skypatcher-authoring", "routes to the forms-onto-NPCs skill"],
+      "expected_output": "Route to the SPID skill — distribution to NPCs by group is not a SkyPatcher job.",
+      "expectations": ["does not pick skypatcher-authoring", "routes to the distribution-to-NPCs-by-group skill"],
       "should_trigger": false,
-      "evidence": "NEW 2026-09-08, never measured. The 'runtime, no esp' phrasing is the lure; the exclusion clause is what should decide it."
+      "evidence": "NEW 2026-09-08, never measured. The 'runtime, no esp' phrasing is the lure; the exclusion clause is what should decide it. Rewritten 2026-09-08 for the new routing: 'every bandit' is a group (faction), which is SPID's half — an individual NPC's edits are SkyPatcher's, so the case now measures the group/individual line rather than forms-onto-NPCs."
     },
     {
       "id": 29,
```
Avick3110, line 258: **Case 29 was flipped to should-trigger, but nothing in the description was changed to make it trigger.**

The trigger arm is declared as "which ONE skill a fresh-context agent loads for a query" — the agent sees descriptions only. The skypatcher-authoring description on this branch never uses the words *container*, *chest* or *loot*; its use clause is "rebalancing weapons, armor, NPCs or leveled lists without an ESP", and it now *opens* its boundary with "Not for distribution". "put these items into a chest without an ESP" reads as distribution, so the description actively pushes the router away from the skill this case now expects. Dropping the CID clause removed the reason to decline; it did not add a reason to pick.

Second problem in the same case: `"reaches the container patcher — filterByContainers with addToContainers"` is not observable in the declared trigger arm — every other should-trigger case's second expectation is about the *reasoning* ("reasoning names authoring a SkyPatcher .ini"), not about a reference file the agent never opens during a routing measurement.

Either add a container/chest token to the description so the case is reachable, or keep case 29 as a NONE-destination negative until the description covers containers.

src/housecarl-mcp/Program.cs
```diff
@@ -154,9 +154,9 @@ static void AddMcp(IServiceCollection services, bool stdio)
             "MD5. Prefer over a browser or web search; each tool's own description carries the specifics. " +
             // The three cross-skill facts the skill rewrite folds here: which runtime framework owns a job, that a
             // write wins nothing until enabled, and the never-copy rule for generated output.
-            "RUNTIME DISTRIBUTION LAYERS — which framework owns a job is decided by what RECEIVES the change: a " +
-            "spell, perk, item, keyword, outfit or faction onto NPCs is SPID; a keyword onto item records is KID; " +
-            "items into containers is CID; a record's OWN FIELDS is SkyPatcher, whose replayed layer " +
+            "RUNTIME DISTRIBUTION LAYERS — which framework owns a job is decided by what RECEIVES the change: " +
+            "distribution to NPCs BY GROUP (faction, race, class) is SPID; a keyword onto item records is KID; a " +
```
Avick3110: **The rewrite drops the SPID form-type list and narrows its filter axes to three, so two common jobs now have no home in the standing context.**

The old sentence named what SPID distributes — "a spell, perk, item, keyword, outfit or faction onto NPCs". The new one names only how SPID *selects* ("BY GROUP (faction, race, class)"). Two concrete misroutes for a session that has only the standing context and has loaded no skill:

1. "give every NPC above level 30 a free perk at runtime, no ESP" — `level` is not one of the three named groups, and it is not a record's own fields either, so the routing answers nothing. SPID's own description covers it ("faction, race, level or trait"), but the standing context no longer does.
2. "distribute this outfit to NPCs at runtime" — `outfit` was a token in the old sentence and is gone; nothing left in the standing context points at SPID.

Also note `faction` silently changed role: it used to be a *form SPID distributes*, and is now a *group SPID filters on*. Distributing a faction to NPCs is still a SPID job and is no longer named anywhere here.

Suggest keeping the form-type list alongside the new group/individual split rather than replacing it.

Avick3110:
Three more findings, on files outside the diff so they cannot be posted inline. All three are places the dropped CID clause still stands.

**1. `.claude/skills/kid-authoring/references/grammar-core.md:40` and `:182` — the shipped routing table still sends containers to CID.**

The PR body says kid-authoring "carried no CID line and needed no change". That is true of its description; it is not true of the reference a session actually reads. `grammar-core.md` §7 lists

```
- item into a **container** → **CID**
```

and repeats it in the routing table at line 182 (`| put an item into a **container** | CID |`). `_CORPUS_STATUS.md:119` says the same, though the leak-check strips that one from the package. `grammar-core.md` ships.

Concrete misbehaviour: a session asked "put 5 gold in every merchant chest, no ESP" loads kid-authoring (a plausible pick — the prompt is about items), opens grammar-core as the skill's first step, reads "container → CID", and declines to a framework that has no skill in this tree — which is exactly the failure this PR set out to remove. The standing instructions and the SkyPatcher/SPID descriptions now say containers are SkyPatcher's; this file still says otherwise, and it is the more specific of the two, so it wins.

**2. `.claude/skills/spid-authoring/evals/evals.json:167` — case 13's expectation rests on a clause that no longer exists.**

```
"expected_output": "NONE — no distributor skill covers containers, and this tree ships no CID skill",
"evidence": "Container target; the description names CID and this tree ships no CID skill."
```

The description no longer names CID — this PR removed that clause from the same file's frontmatter. `should_trigger: false` is still correct, but the recorded expected destination is now wrong: under the new routing the answer is skypatcher-authoring, not NONE. A run graded against this row's `expected_output` scores a correct SkyPatcher routing as a miss. The same suite's line 283 evidence still describes the boundary as "a target boundary against KID/CID/SkyPatcher".

**3. `.claude/skills/kid-authoring/evals/evals.json:186` — same, in KID's suite.**

```
"expected_output": "kid-authoring is NOT loaded; the query routes to CID — items to containers."
```

Prompt T13 is "add 5 gold and a healing potion to every merchant chest in my load order" — the same query as SPID's case 13 and skypatcher's case 29, now with three mutually inconsistent recorded destinations across the three suites (CID, NONE, skypatcher-authoring). skypatcher's case 29 was updated with a dated note; these two were not.

---

Verified and clean, for the record: the served-instruction length really is 3,133 → 3,131 (measured by reconstructing the concatenated literal on both revisions), the arithmetic in the `ServerInstructionsTests` doc comment adds up, the skypatcher suite really does hold 18/14 after the flip, all four changed descriptions stay well under the 1,024-character cap, the two folded-scalar `compatibility` blocks are valid YAML, and `references/container.md` exists on this branch at the path the CHANGELOG cites.

.claude/skills/skypatcher-authoring/evals/evals.json
Avick3110: Folded, by the first of your two options — 9de82835. Case 29 stays should-trigger.

The description's use clause now reads "rebalancing weapons, armor, NPCs or leveled lists without an ESP, putting items into a container or chest, or asking why a patch line isn't applying". That is the one container/chest token the trigger arm needed, and it is reachable from a description-only menu, which is what the arm measures. The description is 667 characters, under the 1,024 cap.

"Not for distribution" stays true in front of it: putting items into a container edits that container record's own item list — a record's own fields, SkyPatcher's lane — while the sentence's exclusions are still distribution to NPCs (SPID) and keywords onto item records (KID). So no rewrite of the opening was needed, and the case did not have to move back to should-not.

Your second point is folded in the same commit: the second expectation was `reaches the container patcher — filterByContainers with addToContainers`, which the routing arm never observes. It now reads `reasoning names putting items into a container without an ESP`, the reasoning shape every other should-trigger case uses. The case evidence and the suite's `method.cases` line record both amendments.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded — 647361d0.

Aaron's call behind this is a preference, not a prohibition: "for distributing stuff to NPCs SPID is best for groups like by faction, skypatcher is better for NPCs individually." You are right that the Scope line read it as a capability claim, and that the skill's own `references/filters.md` §1 (`BalgruufTheGreater` by EditorID) and `grammar-core.md:275` contradict it.

The Scope paragraph now says SPID's target is NPCs, that it is *best* for groups — faction, race, level, trait, matching the description's axes — and that it can also name one NPC by EditorID or FormID, with the reference pointer, so a single-NPC distribution is a real SPID line rather than a refusal. SkyPatcher is named as the *better* fit for one NPC, and as the owner outright of a record's own fields.

`skypatcher-authoring`'s Overview took the same wording in the same commit, since it carried the mirror of this claim.

The frontmatter's "Not this skill … an individual NPC's *edits* are SkyPatcher" is left standing: editing an NPC record's own fields really is outside SPID, and the sentence says edits, not distribution. If you read it as the wider claim, say so and it goes too.

src/housecarl-mcp/Program.cs
Avick3110: Folded — 78abd8f0. Both misroutes were real: the rewrite replaced the form list with a selection axis, so nothing in the standing context named an outfit, and `level` was outside the three groups.

The sentence now carries both — the form list first, the selection axis as a preference on it:

> RUNTIME DISTRIBUTION LAYERS — which framework owns a job is decided by what RECEIVES the change: a spell, perk, item, keyword, outfit or faction onto NPCs is SPID, best BY GROUP (faction, race, level, trait); a keyword onto ITEM RECORDS is KID; a record's OWN FIELDS, and an INDIVIDUAL NPC, are SkyPatcher, whose replayed layer housecarl_skypatcher_layer reads.

`faction` is back in its old role as a form SPID distributes, and appears as a group only inside the parenthesis; `level` and `trait` replace `class` so the axes match `spid-authoring`'s own description; and "an INDIVIDUAL NPC" is now the routing preference rather than a capability line, since SPID can name one NPC by EditorID or FormID.

Measured on the served string, not the source literal: **3,173 characters**, up 42 from 3,131. `ServerInstructionsTests`'s doc comment carries the new figure and the delta, and the PR body is updated.

Avick3110:
All three folded. The three inline findings are answered on their own threads.

**1. `kid-authoring/references/grammar-core.md` — folded, a54e0475.** You are right that the PR body's "carried no CID line" was true of the description and false of the shipped reference. §1's lane list now reads "item into a **container** → **SkyPatcher** (its container patcher — `filterByContainers` with `addToContainers`)", and §9's routing table row reads "SkyPatcher (its container patcher)". `_CORPUS_STATUS.md:119` went the same way in the same commit, even though the packager strips it, so the dev-side ledger does not disagree with the file it describes. The remaining `CID` mentions in that tree are the suffix-tilde syntax note (`grammar-core.md:166`, `value-tables.md:141`) and an `index.jsonl` alias — all still true as statements about the FormID form, and the alias now lands a searcher on a table that says SkyPatcher.

**2. `spid-authoring/evals/evals.json` case 13 — folded, 41dd3e71.** `expected_output` is now `housecarl:skypatcher-authoring`, with a second expectation "Routes to the SkyPatcher skill." and evidence recording the correction and its date. Line 283's evidence, which described the boundary as "against KID/CID/SkyPatcher", now reads "against KID/SkyPatcher".

**3. `kid-authoring/evals/evals.json` T13 — folded, same commit.** `expected_output` now routes to `skypatcher-authoring` and names its container patcher, with a dated note like the one on skypatcher's case 29. The three suites now record one destination for that prompt. T13's `prior_evidence` line still says the 2026-07-28 run routed to CID, which is left standing as a record of what that run did.

Also in this push, from the inline threads: the SPID/SkyPatcher routing is worded as a preference rather than a capability line in both skills' bodies (647361d0); the SkyPatcher description gained a container token so case 29 is reachable in a description-only arm (9de82835); the standing routing sentence has its SPID form list back (78abd8f0), served length now **3,173 characters**, up 42 from 3,131.

Package rebuilt: leak-check clean on both trees, 12 skills, 223 markdown files. `dotnet test --filter "tier!=bridge"`: 1890 passed, 0 failed. Generator `ci-all`: 128/128, `description-vocab-guard` green.

.claude/skills/kid-authoring/references/grammar-core.md
```diff
@@ -37,7 +37,8 @@ KID's target is **items** (object/base records). That's its lane in the distribu
 
 - keyword/form to an **item record** → **KID** (this skill)
 - spell/perk/item/keyword to an **NPC** → **SPID**
```
Avick3110: **The NPC rows here were not aligned with the new routing.**

The container row was corrected in this hunk, but the row above it still states the rule this PR supersedes. Line 39 says `spell/perk/item/keyword to an **NPC** → **SPID**`; the table row at line 182 and the closing "an NPC → SPID" say the same, and `references/_CORPUS_STATUS.md:118` repeats it.

Concretely: a session asked *"add a fire cloak spell to Balgruuf, no ESP"* loads `kid-authoring`, reads these rows and routes to SPID — while `src/housecarl-mcp/Program.cs:159`, `skypatcher-authoring/SKILL.md:13` and `spid-authoring/SKILL.md:25` now all say an individual NPC is SkyPatcher's. The PR body says kid-authoring's routing rows "are aligned"; only the container row was.

.claude/skills/skypatcher-authoring/SKILL.md
```diff
@@ -1,19 +1,21 @@
...
+  Authors and interprets SkyPatcher INI patches — runtime, no-ESP edits that filter Bethesda records and set, add or remove those records' own fields (SkyPatcher 6.4.1 grammar). Load before any SkyPatcher line — the `Plugin.esp|FormID` addressing, the per-type subfolder and the filename gate are non-obvious, and a wrong token fails silently. Use when writing or auditing a SkyPatcher `.ini`, rebalancing weapons, armor, NPCs or leveled lists without an ESP, putting items into a container or chest, or asking why a patch line isn't applying. Not for distribution — distribution to NPCs by group (faction, race, class) is SPID, keywords onto item records is KID.
```
Avick3110: **The group list here is `(faction, race, class)`; everywhere else it is `faction, race, level, trait`.**

`Program.cs:159`, this file's own Overview at line 14, `spid-authoring/SKILL.md:26` and `plugin/CHANGELOG.md:21` all write the four axes as faction, race, level, trait. The description drops level and trait and substitutes class.

That matters because the trigger arm shows an agent descriptions only. Take `spid-authoring` evals case 6, *"add this perk to npcs above level 20 that have decent destruction skill, no esp"* — a should-trigger for SPID. This description's exclusion covers faction, race and class and says nothing about level or skill/trait, while its use clause advertises "rebalancing … NPCs … without an ESP". Nothing in the description declines the job, so SkyPatcher can win a prompt that is squarely SPID's.

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -6,8 +6,8 @@ description: >-
   7.3.0 grammar. Load it before composing or judging any SPID line: a misread filter silently
   changes who is targeted, and an unparseable one is skipped with no error. Use when writing or
   auditing a `_DISTR.ini`, targeting NPCs by faction, race, level or trait, or asking why a line
-  isn't distributing. Not this skill: a keyword onto item records is KID, items into containers is
-  CID, a record's own fields is SkyPatcher.
+  isn't distributing. Not this skill: a keyword onto item records is KID, a record's own fields and
+  an individual NPC's edits are SkyPatcher.
```
Avick3110: **The description states the single-NPC case as an exclusion, which is the refusal the body says must not happen.**

The body (lines 25-27) and the PR description both call the group/individual split a *preference, not a prohibition* — a single-NPC distribution is a real SPID line. But `Not this skill: … an individual NPC's edits are SkyPatcher` reads flatly as a boundary, and the description is the only surface the routing arm sees. The whole distinction rests on the word "edits" versus distribution, which one clause cannot carry.

Prompt *"write a SPID line that gives Balgruuf a fire cloak spell"*: an agent reading the description declines `spid-authoring`. And the suite would not catch it — cases 1-28 in `evals/evals.json` are all group prompts or named-framework audits, with no single-NPC distribution case.

.claude/skills/kid-authoring/evals/evals.json
```diff
@@ -183,7 +183,7 @@
    "id": "T13",
    "prompt": "add 5 gold and a healing potion to every merchant chest in my load order",
    "should_trigger": false,
-   "expected_output": "kid-authoring is NOT loaded; the query routes to CID — items to containers.",
+   "expected_output": "kid-authoring is NOT loaded; the query routes to skypatcher-authoring — its container patcher puts items into containers. Destination corrected 2026-09-08 on review of PR #649; the old answer named CID, which ships no skill in this tree.",
```
Avick3110: **The revision note went into `expected_output`, the field a judge grades against.**

"Destination corrected 2026-09-08 on review of PR #649; the old answer named CID, which ships no skill in this tree" is provenance, not the expected answer. On the next run the adjudicator is handed this string as what a correct answer looks like, so an agent that correctly says "route to skypatcher-authoring" without mentioning CID's absence can be scored short.

`spid-authoring/evals/evals.json:170` makes the same correction and keeps the note in `evidence`; every other row in this file keeps history in `prior_evidence`.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The runtime-framework routing no longer names CID, and containers route to SkyPatcher.** The standing
```
Avick3110, line 22: **Call 1 ships to users but has no changelog line.**

`mutagen-reference/SKILL.md:5-7` and `papyrus-reference/SKILL.md:11-13` each gain a second `compatibility` sentence. That frontmatter ships in the plugin package and is what a user reads to decide whether the skill works on their install — the point of the amendment. `CLAUDE.md`'s PR step asks for a line under `## Unreleased` "if a user would notice the change", and only Call 2 is recorded here.

.claude/skills/kid-authoring/references/grammar-core.md
Avick3110: Folded — 355c00c6. You are right that only the container row moved; the NPC rows still stated the rule this PR supersedes, and a KID session opening `grammar-core.md` reads them before anything else.

All four are aligned now, and all four state the split as the preference Aaron's call is, not as a lane change:

- `references/grammar-core.md:39-40` (§1's lane list) — "spell/perk/item/keyword to **NPCs** → **SPID**, best by group (faction, race, level, trait); for one named NPC **SkyPatcher** is usually the better fit, though a single-NPC SPID line is valid".
- `references/grammar-core.md:183` (§9's routing table) — "SPID (best by group; for one named NPC, usually SkyPatcher)".
- `references/grammar-core.md:187-188` (§9's closing line) — "an item record → KID; NPCs → SPID, and one named NPC usually SkyPatcher."
- `references/_CORPUS_STATUS.md:118-120` — same wording, in the same commit, so the dev-side ledger does not disagree with the file it describes.

On your "add a fire cloak spell to Balgruuf, no ESP" case: the reader now lands on SkyPatcher as the better fit without being told SPID cannot do it, which is what `spid-authoring/references/filters.md` §1 and §2 would contradict.

Package rebuilt: leak-check clean on both trees, 12 skills, 223 markdown files.

.claude/skills/skypatcher-authoring/SKILL.md
Avick3110: Folded — 80d6534d. The description at `SKILL.md:4` now reads "distribution to NPCs by group (faction, race, level, trait) is SPID", so the four axes match `Program.cs:159`, this file's own Overview at line 16, and `spid-authoring/SKILL.md:8`.

The inconsistency was real and it was in the surface that matters: the trigger arm shows descriptions only, so `spid-authoring` case 6 ("add this perk to npcs above level 20 that have decent destruction skill, no esp") met a SkyPatcher description whose exclusion named class and said nothing about level or trait, while its use clause advertised rebalancing NPCs without an ESP. Nothing declined it. It does now.

Worth naming: `dev/DECISIONS.md`'s call 2 does write the parenthetical as "(faction, race, class)". The rest of the tree moved to level/trait on the earlier fold of this PR (78abd8f0), because those are `spid-authoring`'s own axes; this commit brings the last holdout into line rather than reverting three files to the ruling's illustration. Description is 668 characters, under the 1,024 cap.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Folded — a61b5f05. Both halves were real: the body said preference and the description said boundary, and the description is the only surface the routing arm reads.

`SKILL.md:8-11` now reads:

> Use when writing or auditing a `_DISTR.ini`, targeting NPCs by faction, race, level or trait, or asking why a line isn't distributing, **including a line that names one NPC**. Not this skill: a keyword onto item records is KID, and a record's own fields are SkyPatcher, **which is also usually the better fit for one named NPC — a preference, not a boundary**.

So the single-NPC case is now named on the *use* side, where the router looks for a reason to pick, and SkyPatcher keeps only what it owns outright (a record's own fields) plus a stated preference. That matches Aaron's call — SPID is best for groups, SkyPatcher better for NPCs individually, and SPID can still target one NPC — and stops resting the whole distinction on "edits". Description is 641 characters.

Your second point too: the suite really had no single-NPC distribution row. Case 29 is added, should-trigger, on your own prompt — "write a SPID line that gives Balgruuf a fire cloak spell" — with its evidence saying the decline is the failure it exists to catch and that a decline means the description is wrong, not the row. The suite is 15 should-trigger / 14 should-not; the 29 prompts are otherwise unchanged.

.claude/skills/kid-authoring/evals/evals.json
Avick3110: Folded — 4be74595. Putting the provenance in `expected_output` made it part of the answer the adjudicator grades, which would score a correct "route to skypatcher-authoring" short for not mentioning a framework that does not ship.

`evals.json:186` is now just the answer:

> "kid-authoring is NOT loaded; the query routes to skypatcher-authoring — its container patcher puts items into containers."

The note moved to `prior_evidence` (line 191), which is where every other row in this file keeps history, appended after the 2026-07-28 CID run so the record of what that run did stays intact. This file has no `evidence` key at all — 26 rows, all `prior_evidence` — so it goes there rather than mirroring `spid-authoring`'s shape.

plugin/CHANGELOG.md
Avick3110: Folded — 750044ef. Call 1's frontmatter ships in the package, and the whole point of the amendment is that a user reads it to decide whether the skill works on their install, so it is a user-visible change with no entry.

`CHANGELOG.md:16-21` now opens `## Unreleased`:

> **`mutagen-reference` and `papyrus-reference` now say in their frontmatter that their lookup works offline.** Both carried only the shared compatibility sentence — the houseCARL MCP server and a configured Mod Organizer 2 instance — which read as gating the whole skill, including a schema or signature lookup that reads files bundled with the skill and never touches the load order. Each now adds one sentence saying the lookup is offline and that only what reads the live load order needs them. Which half of each skill is which is stated in that skill's own body.

The last line points at the bound rather than restating it, per the file's own writing note: what each skill's offline half covers is scoped in its body, so the entry cannot drift from it.

Avick3110:
All five folded, none refuted. Each is answered on its own thread.

1. **`kid-authoring/references/grammar-core.md` NPC rows — 355c00c6.** §1's lane list, §9's table row and §9's closing line all said an NPC goes to SPID flat; `_CORPUS_STATUS.md:118` said the same. All four now state the group/individual split as a preference, with SkyPatcher named as the better fit for one NPC and a single-NPC SPID line still valid.
2. **`skypatcher-authoring`'s description axes — 80d6534d.** `(faction, race, class)` is now `(faction, race, level, trait)`, matching `Program.cs:159`, the file's own Overview and `spid-authoring`'s description. Noted on the thread that `DECISIONS.md`'s call 2 writes the parenthetical as "class"; the rest of the tree moved to level/trait on 78abd8f0, so this brings the last holdout into line rather than reverting three files.
3. **`spid-authoring`'s description stated the single-NPC case as an exclusion — a61b5f05.** The use clause now names "including a line that names one NPC", and the SkyPatcher clause is "usually the better fit for one named NPC — a preference, not a boundary". Case 29 is added on your prompt, should-trigger, so the suite has a single-NPC distribution row; 15 should-trigger / 14 should-not.
4. **`kid-authoring/evals/evals.json` T13 — 4be74595.** The dated note left `expected_output` for `prior_evidence`, where all 26 rows in that file keep history.
5. **Call 1 had no changelog line — 750044ef.** An entry now opens `## Unreleased`, pointing at each skill's body for what its offline half covers rather than restating it.

`ServerInstructions` is untouched by this push and measures **3,173 characters**, the figure `ServerInstructionsTests` already records.

Package rebuilt with `scripts/build-plugin.ps1`: leak-check clean on both trees, 12 skills, 223 markdown files. `dotnet build -c Release`: 0 errors. `dotnet test --filter "tier!=bridge"`: 1890 passed, 0 failed. Generator `ci-all`: 128/128, `description-vocab-guard` green.

PR #651, opened 2026-09-08
Publish required and enum on nested schema members

### LINKED ISSUE #650: [GAP] Nested parameter members publish no required or enum; both live only in description prose
The published tools/list schema carries `required`, `type` and `enum` only at the top level of each tool's parameters, where the SDK derives them from the C# signature. Inside a nested object the schema says only the type: `create.records[].editorid` is described as "REQUIRED." in prose but the schema does not require it; `apply.ops[].op`'s legal verbs (Set, Add, Remove, CopyFrom, ...) are named in the description and published as a bare string. Measured on the 2026-09-08 capture (tip 3cf3d0e3): 167 nested members across seven parameters (`apply.ops`, `apply.assignments`, `create.records`, `place.assets`, `records.plugins`, `records.project`, `records.walk`), none with `required`, none with `enum`.

Cost today: a client cannot validate a nested call before sending it, so a missing `editorid` or an `op: "Merge"` reaches the server and comes back as a refusal (one round trip a schema check would have saved); readers and reviewers reason about legal values from prose (the `op=Merge` refutation on #645 came from exactly that); and the verb list has two homes, the code's table and the description text, with nothing keeping them equal.

Wanted: the nested shapes emit `required` for members the server refuses without, and `enum` for members whose legal values are a closed set the code already holds (the write verbs first). Generated from the same tables the server validates against, never hand-written. `PublishedDescriptionBoundTests` and the schema-publishing layer (`ToolSchemas.PublishSchemas`) are where it lands; the description prose keeps its sentence but stops being the only home.

Changes the published surface, so after 2.0 unless promoted; found by the nested surface capture for the skill rewrite, 2026-09-08.

Avick3110:
The published `tools/list` schema carried `required`, `type` and `enum` only at the top level of each tool's parameters, where the SDK reads them off the C# signature. Inside a nested object it said only the type, so `create.records[].editorid` was "REQUIRED." in prose while the schema did not require it, and `apply.ops[].op`'s eight write verbs were named in a sentence and published as a bare string — a client could not check a nested call before sending it, and the verb list had two homes. A third publication pass now stamps both, generated from the shapes the server binds and the tables it validates against: `[SchemaRequired]` on the members the server refuses a call without, and `[SchemaValues]` naming the table whose values become the `enum` (`WriteVerbs.All` for apply; for create, that list minus the one verb the create surface refuses by name, derived rather than typed out again). The marks live on the member, not in a path list, because a shape is reached from several parameters — `compose` from four — and the recursion bound expands some of them twice. The pass runs after the flatten, so every expanded copy carries the same stamps its first occurrence does, and the walk is bounded by the schema rather than the type, so the open node closing a recursive chain still constrains nothing. Descriptions are untouched and nothing about what a tool accepts changed; `PublishedDescriptionBoundTests` stays green (longest published description 2,037 of 2,048).

Recaptured from the built exe over stdio with a fresh empty `HOUSECARL_DATA_DIR` (initialize, notifications/initialized, tools/list), the same way the 2026-09-08 capture was taken. Counting object members under the seven nested parameters, the surface has 118 of them; **25 now carry `required`** (0 before) and **2 carry `enum`** (0 before). The 25: `apply.ops[].formid`/`.field_path`, `apply.assignments[].target`/`.from`, `create.records[].record_type`/`.editorid`, `create.records[].ops[].field_path`, `walk.exclusions[].match`/`.severity`, and every published occurrence of `compose.type` and `compose.sets[].path` — 16 of the 25, because those two shapes recur under `compose`, `composes[]` and one expansion deeper. The 2 enums are `apply.ops[].op` (eight verbs) and `create.records[].ops[].op` (the same seven the create surface already recited).

`PublishedNestedConstraintTests` reads the served document, identifies each object node by its own member-name set, and asserts the node's `required` and `enum` equal what that C# shape declares — in both directions, plus an anti-vacuity arm that every marked shape is actually reached (21 occurrences of 7 shapes today). It reads the same tables the pass does, so a verb added to `WriteVerbs.All` moves the published enum and the expectation together and cannot drift; its walk is written independently of the pass's so it cannot agree with a stamping bug by construction. Removing the pass fails it.

**Sets left unmarked, because no table backs them.** Four closed value sets are decided by a literal pattern at the call site rather than by a collection anything else can read, so publishing an enum would have meant writing a second home rather than exposing the first: `walk.exclusions[].severity` (`stop`/`refuse`), `walk.direction` (`forward`/`reverse`), `project.form` (the ten forms, a `switch` case list), and `assets[].kind` (a `switch` over eight accepted spellings while the description names two). Each also folds case and trims before comparing, so an exact-match enum would publish narrower than the gate accepts. A fifth is worse than untabled: `compose.sets[].verb`'s description names five verbs, but the request it builds is validated by the same rulebook switch an op is, which accepts all eight — the description and the gate disagree, so there is nothing to publish that is true of both, and it is left alone. `place.assets[]` gets no `required` either: its requirement is `formid` XOR `path`, not a per-member one, and JSON Schema `required` cannot say that.

Closes #650


src/housecarl-mcp/NestedSchemaConstraints.cs
```diff
@@ -0,0 +1,139 @@
...
+            if (property.GetCustomAttribute<JsonIgnoreAttribute>() is not null) continue;
+            var wire = property.GetCustomAttribute<JsonPropertyNameAttribute>()?.Name ?? property.Name;
+            if (props[wire] is not JsonObject member) continue;
+
+            if (property.GetCustomAttribute<SchemaRequiredAttribute>() is not null) required.Add(wire);
+            if (property.GetCustomAttribute<SchemaValuesAttribute>() is { } values)
+            {
+                var legal = new JsonArray();
+                foreach (var v in Values(values.Vocabulary)) legal.Add(v);
+                member["enum"] = legal;
```
Avick3110, line 134: **The published `enum` is narrower than the gate — it refuses `"op": null`, which the server accepts.**

Both `[SchemaValues]` members are `string?`, so the generator publishes them as `type: ["string","null"]`, and this line stamps an `enum` holding only the verb names. JSON Schema applies `enum` to *every* instance, `null` included, so the two constraints on the same member now contradict each other. Captured from the built exe over stdio:

```
housecarl_apply  ops[].op            type=["string","null"]  enum=[Set,Add,…,CopyFrom]
housecarl_create records[].ops[].op  type=["string","null"]  enum=[Set,Add,…,Merge]
```

Concrete misbehaviour: a client that validates before sending — the whole reason the PR publishes these — rejects

```json
{"formid":"012345:X.esp","field_path":"Name","op":null}
```

and `{"…","op":""}` too, while the server accepts both and treats them as `Set`: `ApplyOp.Op`/`CreateFieldOp.Op` are nullable, `ListParams.Read` binds an explicit null fine, and `MapEdit`/`MapCreateEdit` do `string.IsNullOrWhiteSpace(op.Verb) ? "Set" : op.Verb`. So a call the tool answers is now refused a hop earlier, by a schema this pass wrote — and never with the tool's own sentence.

This is the same "publishing narrower than the gate accepts" the PR body uses to justify leaving `walk.exclusions[].severity`, `walk.direction`, `project.form` and `assets[].kind` unmarked; it just lands on the two members that *did* get published. Adding `null` to the emitted array when the member's published `type` admits it (or dropping `"null"` from that member's type) makes the two agree.

src/housecarl-mcp/CreateTools.cs
```diff
@@ -158,10 +158,10 @@ public sealed record CreateRecordSpec
...
     public string? FieldPath { get; init; }
 
-    [JsonPropertyName("op"), Description("Set (default) | Add | Remove | SetAtIndex | InsertAtIndex | ReplaceAll | Merge. SetAtIndex OVERWRITES the element at key=; InsertAtIndex inserts a new one AT key= and shifts the rest right (key = the list's length appends). On a [Flags] enum, Add sets one bit and Remove clears one, leaving the others untouched. There is no CopyFrom on this surface — a record that does not exist yet has no other version to copy a field from; copying a field from ANOTHER record is " + ToolNames.Apply + "'s CopyFrom.")]
+    [SchemaValues(SchemaVocabulary.CreateVerbs), JsonPropertyName("op"), Description("Set (default) | Add | Remove | SetAtIndex | InsertAtIndex | ReplaceAll | Merge. SetAtIndex OVERWRITES the element at key=; InsertAtIndex inserts a new one AT key= and shifts the rest right (key = the list's length appends). On a [Flags] enum, Add sets one bit and Remove clears one, leaving the others untouched. There is no CopyFrom on this surface — a record that does not exist yet has no other version to copy a field from; copying a field from ANOTHER record is " + ToolNames.Apply + "'s CopyFrom.")]
```
Avick3110: **The create surface's verb list still has two homes, and this PR made them diverge-able in the same field.**

The PR's stated problem is that "the verb list had two homes — the table the gate reads, and the description text". On `apply` that is now closed: the description concatenates `WriteVerbs.AllRecital`, and `DescriptionVocabularyGuardProbe`'s INV4-HOMES pins `All` == `AllRecital` == an independently written list, so a new verb cannot land in one without the other.

Here it is not. The `enum` on this member is derived (`WriteVerbs.OnCreate`), but the `[Description]` beside it still spells `"Set (default) | Add | Remove | SetAtIndex | InsertAtIndex | ReplaceAll | Merge"` as a hand-typed literal, and nothing pins the two together — `PublishedNestedConstraintTests` reads `OnCreate` on both sides of its assertion, and no probe covers this recital.

Scenario: add a ninth verb to `WriteVerbs.All`. INV4-HOMES forces `AllRecital` and the independent list to follow, `OnCreate` and this member's published `enum` pick it up for free — and this description alone still names seven. The published schema and the published description *of the same field* then disagree about what the create surface accepts, which is a sharper version of the drift the PR set out to remove. A `WriteVerbs.OnCreateRecital` const built the way `AllRecital` is (or pinning this description against `OnCreate` in the vocabulary probe) would close it.

src/housecarl-mcp/NestedSchemaConstraints.cs
```diff
@@ -0,0 +1,139 @@
...
+                changed = true;
+            }
+            changed |= Walk(property.PropertyType, member);
+        }
+
+        if (required.Count > 0) { node["required"] = required; changed = true; }
```
Avick3110: **This replaces the node's `required` rather than merging into it, and on a marked shape the loss is invisible to the new test.**

Today no nested wire member is non-nullable, so the generator emits no nested `required` and there is nothing to clobber — I confirmed that on the served document: all 21 `required` nodes are exactly the stamped ones. But the pass is written as if that will hold.

Scenario: someone gives `NestedSet` a non-nullable member (`public required string Kind { get; init; }`, or a plain non-nullable `string`). `AIJsonUtilities.CreateJsonSchema` then emits `required: ["kind"]` on every published nested-set node; this line overwrites it with `["path"]`, and the published schema silently stops requiring a member the binder still requires. `PublishedNestedConstraintTests` stays green, because it compares the published list against the *marked* list only — the generator's contribution is outside what it looks at. (The `NoPublishedNestedObjectCarriesAConstraintNoMarkedShapeAsksFor` arm catches the same drift on an *unmarked* shape, which is why the gap is specifically marked ones.)

Unioning with whatever `node["required"]` already holds, instead of assigning over it, makes the pass additive the way its own doc line ("stamps what the shape declares") reads.

src/housecarl-core/WriteVerbs.cs
```diff
@@ -99,6 +99,16 @@ public static class WriteVerbs
...
     public const string AllRecital = "Set (default) | Add | Remove | SetAtIndex | InsertAtIndex | ReplaceAll | Merge | CopyFrom";
 
+    /// <summary>The verb that copies a field from another version of a record, named once so the surfaces that
+    /// refuse it do not each spell it. <see cref="Transplant"/> is the same verb as a <see cref="VerbUse"/>.</summary>
+    public const string Transplanting = "CopyFrom";
+
+    /// <summary>The verbs a CREATE surface accepts — <see cref="All"/> minus the one it refuses by name, because a
+    /// record that does not exist yet has no other version to copy a field from. DERIVED from the two, never a third
+    /// hand-typed list, so a verb added to the vocabulary reaches this surface too.</summary>
+    public static readonly IReadOnlyList<string> OnCreate =
```
Avick3110: **`OnCreate` is derived, but its subtrahend is a hand-typed literal that nothing anchors in `All` — and when it drifts, the derivation degrades silently.**

`OnCreate` is `All` minus an ordinal match on `Transplanting = "CopyFrom"`. If the verb is ever renamed or dropped in `All` (or `Transplanting`'s spelling is edited alone), `Where` simply matches nothing and `OnCreate` becomes `All` — no error, no shorter list.

What follows from that: `create.records[].ops[].op` publishes an `enum` advertising the transplanting verb the create surface refuses, and `MapCreateEdit`'s refusal (`string.Equals(op.Verb, WriteVerbs.Transplanting, StringComparison.Ordinal)`, the line this PR moved onto the same literal) no longer fires either, so the verb reaches the rulebook instead of the create surface's own sentence. The new test cannot see it: `NestedSchemaConstraints.Values` reads `OnCreate` and `PublishedNestedConstraintTests` reads `OnCreate`, so both sides move together.

`All.Contains(Transplanting)` is the one fact the derivation rests on; asserting it where `OnCreate` is built (a throwing static initializer, or `Single(...)` instead of `Where(...)`) turns a silent widening into a loud one.

src/housecarl-mcp/NestedSchemaConstraints.cs
Avick3110: Folded in 640e32f3 — the enum now carries `null` when the member's published `type` admits it.

Verified the gate before changing anything: `ApplyOp.Op` / `CreateFieldOp.Op` are `string?`, `ListParams.Read` binds an explicit null, and `MapEdit` / `MapCreateEdit` both do `string.IsNullOrWhiteSpace(op.Verb) ? "Set" : op.Verb` (LoadOrderService.cs:7657, 7691). So a null op is Set, exactly as you say.

Which of the two fixes: **add `null` to the enum**, not drop `"null"` from the type. The brief was to make the schema say what the server does, and the server accepts null — so the nullable type is the true half and the enum was the narrow one. Dropping `"null"` would have made them agree by publishing a refusal the server does not perform.

The rule is written into the pass rather than onto the two members: an enum gets a `null` entry iff the member's published `type` admits null, read off the served document rather than off the CLR type, so a future non-nullable marked member publishes a clean enum with no entry to remove by hand. Recaptured from the built exe: `apply.ops[].op` = `[Set,…,CopyFrom,null]`, `create.records[].ops[].op` = `[Set,…,Merge,null]`, both still `type:["string","null"]`.

`{"op":""}` stays outside the published set, and deliberately. The gate is `IsNullOrWhiteSpace`, so the accepted set is the verbs plus null plus every whitespace-only string — an enum cannot state that last family, and enumerating `""` would publish an empty string as a legal verb value while still missing `" "`. So the gap narrows from "refuses a documented spelling" to "refuses a blank string nothing documents and no description offers", which I think is the right place to stop without changing what the tool accepts. Said so in the architecture note.

Left the `required`-plus-nullable half alone — that is #652's, and it is the opposite direction (the schema admitting a null the server refuses), so it wants the type changed rather than the constraint widened.

src/housecarl-mcp/CreateTools.cs
Avick3110: Folded in 7cfa9e04 — the description now concatenates a derived recital, and a probe arm pins it.

An attribute argument has to be a compile-time constant, so the description cannot read `WriteVerbs.OnCreate` directly; it gets `WriteVerbs.OnCreateRecital`, a new const built exactly the way `AllRecital` is. That alone would only move the hand-typed list one file over, so `DescriptionVocabularyGuardProbe` gains `INV4-CREATEHOMES` beside `INV4-HOMES`: it holds `OnCreate` against `OnCreateRecital` against a create vocabulary written out independently in the probe, through the same `HomesAgree` the apply arm uses (and which `RED-HOMES` already drives in both directions). No new guard file.

Your scenario now: a ninth verb in `All` reddens `INV4-HOMES` until `AllRecital` and the probe's independent list follow, and reddens `INV4-CREATEHOMES` until `OnCreateRecital` and its independent list follow. The schema and the description of `create.records[].ops[].op` cannot name different sets and stay green.

Recaptured from the built exe, the create op now reads `Set (default) | Add | Remove | SetAtIndex | InsertAtIndex | ReplaceAll | Merge.` — same text as before, different source. `PublishedDescriptionBoundTests` is green; the longest published description is unchanged.

src/housecarl-mcp/NestedSchemaConstraints.cs
Avick3110: Folded in 640e32f3 — the stamp unions with whatever `required` the node already carries instead of assigning over it.

Confirmed your reading first: no nested wire member is non-nullable today, the generator emits no nested `required`, and all 21 stamped nodes on the served document are exactly the marked ones. So this is the latent case, not a live loss — but the pass is written as if it will hold, which is the thing to fix.

The stamped names are collected as strings now and merged into the existing array, generator entries first and in their own order, deduplicated ordinally. Nothing about the served document changes.

The test moves with it, because your scenario would otherwise have reddened it for the right behaviour: `EveryPublishedOccurrenceOfAMarkedShapeCarriesItsRequiredAndEnum` asserted set equality with the marked list, which a legitimate generator contribution would now break. It asserts a superset instead — every marked name must be published — bounded on the other side by "every published name must be a member the shape declares", which is what catches a stamped name that is not one. Extras are printed with the occurrence that carries them, so a generator contribution shows up in the run rather than being quietly tolerated. `NoPublishedNestedObjectCarriesAConstraintNoMarkedShapeAsksFor` is unchanged.

src/housecarl-core/WriteVerbs.cs
Avick3110: Folded in 7cfa9e04 — `OnCreate` is now built by a method that throws when `Transplanting` is not in `All`.

```
static IReadOnlyList<string> BuildOnCreate()
{
    if (!All.Contains(Transplanting, StringComparer.Ordinal))
        throw new InvalidOperationException(...);
    return All.Where(v => !string.Equals(v, Transplanting, StringComparison.Ordinal)).ToArray();
}
```

`Single(...)` would have been the tighter spelling, but it throws with a message about a sequence rather than about the two spellings, so the explicit check states the fact it is guarding: the message names both `Transplanting` and the contents of `All` and says what would otherwise happen — the subtraction removes nothing and `OnCreate` becomes the whole vocabulary.

Exercised it rather than assuming: flipping the const to `"CopyFrm"` and rebuilding takes the whole `stdio` tier down — the server never answers `initialize`, because the type initializer throws on the first touch of `WriteVerbs` — and restoring it puts 1894/1894 back. So the degradation is loud at startup, and every existing test that drives the built server is the thing that sees it.

src/housecarl-mcp-tests/PublishedNestedConstraintTests.cs
```diff
@@ -0,0 +1,237 @@
...
+            if (missing.Length > 0)
+                problems.Add($"{node.Tool} {node.Path}: required=[{string.Join(",", node.Required)}] omits " +
+                             $"[{string.Join(",", missing)}], which {shape.Type.Name} marks");
+            foreach (var name in node.Required.Except(shape.Required, StringComparer.Ordinal))
+            {
+                if (!node.Members.Contains(name, StringComparer.Ordinal))
+                    problems.Add($"{node.Tool} {node.Path}: required names '{name}', which is not a member of " +
+                                 $"{shape.Type.Name}");
+                else extras.Add($"{node.Tool} {node.Path}: '{name}' required by the generator, not by a mark");
+            }
```
Avick3110: **Correctness (test coverage).** Relaxing this from `SequenceEqual` to a superset check drops the only assertion that would catch over-stamping on a marked shape. If the `[SchemaRequired]` lookup in `NestedSchemaConstraints.Walk` ever regresses so that every property is added, `apply.ops[].items` publishes `required=[formid,field_path,op,value,key,...]`: `missing` is empty, and each extra name *is* a member of `ApplyOp`, so it goes to `extras` — printed to test output, never asserted. `NoPublishedNestedObjectCarriesAConstraintNoMarkedShapeAsksFor` only inspects nodes whose member set is **not** a marked shape, so it passes too. Green suite, and the served schema now tells a client that `value` is mandatory on every op, so a validating client refuses `{"formid": "...", "field_path": "Name", "op": "Remove"}` — a call the server accepts.

The branch this was relaxed for is dead today: I captured `tools/list` off the built exe, and all 25 nested `required` entries come from the pass, none from the generator. So a live check was traded for a hypothetical one.

Suggest keeping the equality assert, and reinstating the superset form only once a generator-contributed entry actually appears — or asserting equality against `shape.Required` plus an explicitly named allowlist of generator entries, so the extras stay in the assertion rather than in `_out.WriteLine`.

src/housecarl-mcp-tests/PublishedNestedConstraintTests.cs
```diff
@@ -0,0 +1,237 @@
...
+            && type.EnumerateArray().Any(v => v.ValueKind == JsonValueKind.String && v.GetString() == "null");
+    }
+
+    static string Render(IEnumerable<string?> values) => string.Join(",", values.Select(v => v ?? "null"));
+
+    // ---- the assertions ----------------------------------------------------------------------------------
+
+    /// <summary>Identifying a node by its member set is only sound while no two marked shapes share one.</summary>
+    [Fact]
+    [Trait("tier", "unit")]
```
Avick3110: xUnit traits **accumulate** — a method-level `[Trait]` does not replace the class-level one, so this test carries both `tier=stdio` (from line 25) and `tier=unit`, while its constructor still takes `ServerFixture`.

Concretely: `dotnet test src/housecarl-mcp-tests --filter "tier=unit"` now matches this test, xUnit constructs the class, and the `"server"` collection fixture spawns the stdio server — so a unit-only run pays the stdio cost and fails anywhere the server can't start. `standards/TESTING.md:23` says to tag the *class* with the tier, and this method needs the fixture exactly like the other three in the file.

The reflection-only nature is worth saying in the summary, but it isn't a tier. Suggest dropping the trait.

```suggestion
    [Fact]
```

plugin/CHANGELOG.md
```diff
@@ -28,6 +28,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
+- **The published tool schema now marks nested members `required`, and names the write verbs as an `enum`, so a
+  client can check a nested call before sending it.** `required` and `enum` were emitted only at the top level of a
+  tool's parameters; inside `apply.ops`, `create.records` and the rest, a member the server refuses a call without
+  said so in its description alone, and the legal verbs on `ops[].op` were prose beside a bare `string`. Both are now
+  published, generated from the shapes the server binds and the verb tables it validates against — `WriteVerbs.All`
+  on `apply`, and the create surface's own set (that list minus the verb it refuses by name) on `create`. What the
+  tools accept, and every description, are unchanged; a member whose legal values no table holds is left as it was
+  rather than given a set invented for the schema. A verb member also accepts an explicit `null` — the server reads
+  that as "none given" and uses `Set` — so `null` is in the published set beside the names.
+
```
Avick3110: This blank line separates the new entry from the next bullet; no other entry in the file has one. Markdown reads a list with a blank line between items as *loose*, so the renderer wraps every item in this `## Unreleased` list in `<p>` and the whole section renders with extra vertical spacing — unlike every other release section. Cosmetic, but it's a shipped file. Drop the line.

src/housecarl-mcp-tests/PublishedNestedConstraintTests.cs
Avick3110: Folded in 0e20508b. Restored the equality assert against `shape.Required` (both directions), dropped the `extras` list and its `_out.WriteLine`. Confirmed live: with equality back the suite is green, so no generator-contributed nested `required` exists on this surface today. The summary now says to name a generator entry in an allowlist beside `shape.Required` when one appears, rather than relaxing back to a superset. The pass itself stays additive.

src/housecarl-mcp-tests/PublishedNestedConstraintTests.cs
Avick3110: Folded in 0e20508b. Dropped the method-level `[Trait("tier","unit")]`; the class-level `tier=stdio` at line 25 is the only tier now. Verified: `--filter "tier=unit&FullyQualifiedName~PublishedNestedConstraintTests"` matches nothing. Reflection-only moved into the summary.

plugin/CHANGELOG.md
Avick3110: Folded in 4d4464a5 — blank line deleted, the Unreleased list is tight again like every other section.

PR #657, opened 2026-09-08
Apply the four 2026-09-08 evening rulings to the shipped skills

Avick3110:
Applies the four rulings dated 2026-09-08 (evening) to the shipped skills under `.claude/skills/`. Three of them are text-only; the fourth replaces two hedges in the OAR reference with facts read off OAR's own public source, so the skill now tells you what a DAR file means instead of telling you to pick a reading and disclose it.

- **facegen-diagnostics eval case.** `st-copy-face-boundary` becomes `snt-copy-face-boundary`: a dark face right after a copy is `housecarl:npc-appearance-copy`'s own case, and the prompt asks how to do the copy, so facegen-diagnostics should not load. The expectation that the answer still checks the copied NPC's facegen files were carried stays. The coexistence `failure_mode` narrows to a diagnosis prompt with no copy in it. No case count is stated anywhere in the shipped tree, so nothing else needed updating. The description now cedes the case too, and so does §11's routing table, so both halves of the file say the same thing.
- **"Common mistakes" heading.** The eight skills that carried the bare heading now read `## Common mistakes, and the rule that replaces each`, the form facegen-diagnostics already had. `npc-appearance-copy`'s table, the one the heading was taken from, gained the third column the heading promises; the other tables are unchanged. Nothing in the shipped tree — SKILL.md bodies, `references/`, `evals/`, `plugin/codex/housecarl/SKILL.md`, `plugin/README.md`, `README.md` — pointed at the section by its old name.
- **OAR reference, the two UNVERIFIED rows.** Both settled from `ersh1/OpenAnimationReplacer`, read 2026-09-08. The reference cites the files and functions by name and records the date; no source is copied in. The conversion table's type rows are unchanged (they were already right), the DAR worked example in SKILL.md drops the "convert under one reading, say which" hedge, and the eval case that asserted the binding was unsettled now asserts the answer.
  - **AND/OR binding: `OR` binds tighter than `AND`.** A chain is an `AND` of `OR`-groups, not a flat left-to-right sequence. `Parsing.cpp`, `ConditionsTxtFile::GetConditions`: the file's conditions go into one top-level set evaluated with `EvaluateAll`; a line ending in `OR` opens a nested `ORCondition` and the recursion closes on the first line that does not end in `OR`, which is itself the group's last member. `Conditions.cpp`, `ORCondition::EvaluateImpl` evaluates that group with `EvaluateAny`. So an `OR`-group is a maximal run of lines ending in `OR` plus the single line after it.
  - **DAR type numbers convert `n → n`, including 6, 9, 10 and 11.** There is no second numbering. `OpenAnimationReplacer.cpp`, `InitFactories` registers `IsEquippedRightType` and `IsEquippedLeftType` as hidden factories building the same `IsEquippedTypeCondition` as native `IsEquippedType`, with the hand flag preset; `Conditions.cpp`, `IsEquippedTypeCondition::InitializeLegacy` parses the DAR argument straight into that condition's numeric component, labelled by `IsEquippedTypeCondition::GetEnumMap` — §5's table. So a DAR `6` is battleaxe, `9` crossbow, `10` warhammer, `11` shield. The consequence is now stated plainly: OAR is what parses the legacy folder, so the 6/10 split already applies to it before any conversion, and a faithful conversion preserves the behaviour rather than introducing the gap.
- **facegen compatibility line.** `compatibility:` was the "MO2 instance" variant; it now carries the batch string every other skill has.

`scripts/build-plugin.ps1` green, both leak scans clean (12 skills, 227 markdown files, across 2 trees; assembled-tree scan clean). `dotnet build -c Release` 0 errors, `dotnet test --filter "tier!=bridge"` 1894 passed / 0 failed, `ci-all` 128/128 passed.

### DIFF
--- a/.claude/skills/facegen-diagnostics/SKILL.md
+++ b/.claude/skills/facegen-diagnostics/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: facegen-diagnostics
-compatibility: Requires the houseCARL MCP server and a configured MO2 instance.
+compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
 description: >-
   Diagnoses and repairs the dark / grey / black-face NPC bug in Skyrim SE by diffing which mod wins the head .nif against which wins the face .dds for the same NPC, then the record winner behind them. Use for any discolored face, neck seam, "fine in xEdit but wrong in game", NPCs gone dark after an ESL-compaction or merge, or any FaceGen / facegeom / facetint mention. Not a purple or white face (a missing texture) and not player-only grey (RaceMenu/SKEE); copying a face onto another NPC is housecarl:npc-appearance-copy. Load before judging any face bug — the mesh-versus-tint pair decides the fix.
 ---
--- a/.claude/skills/facegen-diagnostics/SKILL.md
+++ b/.claude/skills/facegen-diagnostics/SKILL.md
@@ -2,7 +2,7 @@
 name: facegen-diagnostics
 compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
 description: >-
-  Diagnoses and repairs the dark / grey / black-face NPC bug in Skyrim SE by diffing which mod wins the head .nif against which wins the face .dds for the same NPC, then the record winner behind them. Use for any discolored face, neck seam, "fine in xEdit but wrong in game", NPCs gone dark after an ESL-compaction or merge, or any FaceGen / facegeom / facetint mention. Not a purple or white face (a missing texture) and not player-only grey (RaceMenu/SKEE); copying a face onto another NPC is housecarl:npc-appearance-copy. Load before judging any face bug — the mesh-versus-tint pair decides the fix.
+  Diagnoses and repairs the dark / grey / black-face NPC bug in Skyrim SE by diffing which mod wins the head .nif against which wins the face .dds for the same NPC, then the record winner behind them. Use for any discolored face, neck seam, "fine in xEdit but wrong in game", NPCs gone dark after an ESL-compaction or merge, or any FaceGen / facegeom / facetint mention. Not a purple or white face (a missing texture) and not player-only grey (RaceMenu/SKEE); copying a face onto another NPC, or a face gone dark right after a copy, is housecarl:npc-appearance-copy. Load before judging any face bug — the mesh-versus-tint pair decides the fix.
 ---
 
 # Facegen diagnostics
--- a/.claude/skills/facegen-diagnostics/SKILL.md
+++ b/.claude/skills/facegen-diagnostics/SKILL.md
@@ -260,4 +260,4 @@ nothing. That is worse than a clear non-answer.
 | Causes and fixes by letter, the symptom table, and which community tool owns a case houseCARL cannot | `references/facegen-causes-and-fixes.md` — read it to pin a specific cause or justify a fix to the user; the flow above drives most diagnoses without it |
 | The mesh-side repairs — rewriting a baked shape name, the embedded FaceTint path, or a skin slot — and what each one can and cannot prove | `references/mesh-repairs.md` — read it when §3 says the file wins but the face is still wrong, or when a mesh refuses a write |
 | An `NPC_` field path or enum spelling, before composing a `housecarl_apply` op | `housecarl:mutagen-reference` |
-| Copying a face onto a *different* NPC, or cloning one as a standalone | `housecarl:npc-appearance-copy` |
+| Copying a face onto a *different* NPC, cloning one as a standalone, or a face gone dark right after either | `housecarl:npc-appearance-copy` |
--- a/.claude/skills/facegen-diagnostics/evals/evals.json
+++ b/.claude/skills/facegen-diagnostics/evals/evals.json
@@ -47,7 +47,7 @@
     {
       "dimension": "coexistence",
       "covered_by": "the cases below plus the output-quality task, coexisting configuration",
-      "failure_mode": "housecarl:npc-appearance-copy loads instead of this skill on a diagnosis prompt, or a sweep prompt loads this skill and housecarl:bulk-record-jobs together and the two flows give contradictory instructions."
+      "failure_mode": "housecarl:npc-appearance-copy loads instead of this skill on a diagnosis prompt with no copy in it, or a sweep prompt loads this skill and housecarl:bulk-record-jobs together and the two flows give contradictory instructions."
     },
     {
       "dimension": "instruction following",
@@ -177,13 +177,13 @@
       ]
     },
     {
-      "id": "st-copy-face-boundary",
+      "id": "snt-copy-face-boundary",
       "prompt": "i copied a face from an overhaul onto my custom follower and now her head renders dark. did i copy the wrong fields, and what's the right way to do that copy?",
-      "expected_output": "facegen-diagnostics is loaded on the dark-face symptom, and the copy itself is ceded to housecarl:npc-appearance-copy.",
+      "expected_output": "housecarl:npc-appearance-copy, not this skill — a dark face right after a copy is the copy's own case, and the prompt asks how to do the copy.",
       "files": [],
       "expectations": [
-        "The facegen-diagnostics skill is loaded.",
-        "The authoring half is routed to housecarl:npc-appearance-copy rather than reinvented here.",
+        "The facegen-diagnostics skill is NOT loaded.",
+        "The copy is handled by housecarl:npc-appearance-copy rather than by a diagnosis flow.",
         "The answer still checks that the copied NPC's facegen files were carried, not only the record fields."
       ]
     },
--- a/.claude/skills/facegen-diagnostics/evals/evals.json
+++ b/.claude/skills/facegen-diagnostics/evals/evals.json
@@ -42,7 +42,7 @@
     {
       "dimension": "isolation",
       "covered_by": "the cases below plus the output-quality task, isolated configuration",
-      "failure_mode": "The flow dead-ends where the body cedes — a whole-order sweep, an NPC_ field spelling, or a face copy — because housecarl:bulk-record-jobs, housecarl:mutagen-reference or housecarl:npc-appearance-copy is absent, instead of the cede degrading to a plain instruction the agent can still follow."
+      "failure_mode": "The flow dead-ends where the body cedes — a whole-order sweep, or an NPC_ field spelling — because housecarl:bulk-record-jobs or housecarl:mutagen-reference is absent, instead of the cede degrading to a plain instruction the agent can still follow. The copy cede is not covered here: snt-copy-face-boundary grades the answer, not another skill's presence, so no case exercises it with this skill loaded."
     },
     {
       "dimension": "coexistence",
@@ -183,7 +183,7 @@
       "files": [],
       "expectations": [
         "The facegen-diagnostics skill is NOT loaded.",
-        "The copy is handled by housecarl:npc-appearance-copy rather than by a diagnosis flow.",
+        "The copy is treated as an authoring job — which fields to seed and what the copy still owes — rather than as a diagnosis flow over the load order.",
         "The answer still checks that the copied NPC's facegen files were carried, not only the record fields."
       ]
     },
--- a/.claude/skills/npc-appearance-copy/SKILL.md
+++ b/.claude/skills/npc-appearance-copy/SKILL.md
@@ -238,8 +238,8 @@ Then re-read it, naming the folder again —
 | Dropping a bundle member the donor lacks instead of clearing it | The target keeps its own morphs under the donor's head parts — a face built from two people | Copy what the donor has and clear what it lacks — `Remove`, or `ReplaceAll` with `composes: []` for `TintLayers` |
 | Stopping at the patch | The records exist, the FaceGen does not — a dark face you authored on purpose | The copy is four calls; Step 3 places the FaceGen pair |
 | Stopping at the placement | The placed mesh still points at the donor's tint: correct until the donor is uninstalled | Finish with Step 4 — repoint the tint inside the placed mesh, naming `mod=` |
-| Letting the FaceGen source default to the VFS winner | You place a replacer's face over the records of the donor you actually copied |
-| Reporting "copied" when the strip list is long | A clone with no factions, outfits, packages or scripts is not a working follower |
+| Letting the FaceGen source default to the VFS winner | You place a replacer's face over the records of the donor you actually copied | Name `source_provider` explicitly, chosen by inspecting each candidate with `housecarl_nif_inspect mod=` and keeping the copy whose baked shape names match — Step 3's "Which provider to name" |
+| Reporting "copied" when the strip list is long | A clone with no factions, outfits, packages or scripts is not a working follower | Report the strip list as work still owed, item by item, before calling the copy done |
 
 ## Verification — what this session can check
 
--- a/.claude/skills/npc-appearance-copy/SKILL.md
+++ b/.claude/skills/npc-appearance-copy/SKILL.md
@@ -229,15 +229,15 @@ Then re-read it, naming the folder again —
 
 ## Common mistakes, and the rule that replaces each
 
-| Mistake | What it costs |
-|---|---|
-| Copying from a `Traits`-templated donor | Its appearance fields are empty, so the seeds clear the target's: a blank face, reported as a success |
-| Seeding `Race`, or omitting `exclude_types` because you expect a `Race` default | There is none: the walk enters the race and pulls the skeleton and sibling races |
-| Renaming the copied head parts | The mesh's baked shape names stop matching; the engine regenerates a vanilla head |
-| Omitting `TextureLighting` | Every field reads correct and the skin renders dark |
-| Dropping a bundle member the donor lacks instead of clearing it | The target keeps its own morphs under the donor's head parts — a face built from two people |
-| Stopping at the patch | The records exist, the FaceGen does not — a dark face you authored on purpose |
-| Stopping at the placement | The placed mesh still points at the donor's tint: correct until the donor is uninstalled |
+| Mistake | What it costs | The rule that replaces it |
+|---|---|---|
+| Copying from a `Traits`-templated donor | Its appearance fields are empty, so the seeds clear the target's: a blank face, reported as a success | Read `Configuration.TemplateFlags` first and copy from the record in `Template` |
+| Seeding `Race`, or omitting `exclude_types` because you expect a `Race` default | There is none: the walk enters the race and pulls the skeleton and sibling races | Pass `Race:refuse` yourself; bundle `Race` in Step 2, and only when the two races differ |
+| Renaming the copied head parts | The mesh's baked shape names stop matching; the engine regenerates a vanilla head | Leave the EditorIDs the copy writes alone |
+| Omitting `TextureLighting` | Every field reads correct and the skin renders dark | Carry Step 2's bundle whole, `TextureLighting` included |
+| Dropping a bundle member the donor lacks instead of clearing it | The target keeps its own morphs under the donor's head parts — a face built from two people | Copy what the donor has and clear what it lacks — `Remove`, or `ReplaceAll` with `composes: []` for `TintLayers` |
+| Stopping at the patch | The records exist, the FaceGen does not — a dark face you authored on purpose | The copy is four calls; Step 3 places the FaceGen pair |
+| Stopping at the placement | The placed mesh still points at the donor's tint: correct until the donor is uninstalled | Finish with Step 4 — repoint the tint inside the placed mesh, naming `mod=` |
 | Letting the FaceGen source default to the VFS winner | You place a replacer's face over the records of the donor you actually copied |
 | Reporting "copied" when the strip list is long | A clone with no factions, outfits, packages or scripts is not a working follower |
 
--- a/.claude/skills/npc-appearance-copy/SKILL.md
+++ b/.claude/skills/npc-appearance-copy/SKILL.md
@@ -227,7 +227,7 @@ Then re-read it, naming the folder again —
 `housecarl_nif_inspect(mesh_paths = ["<the placed mesh>"], sections = "paths", mod = "<that same folder>")`
 — and check that slot 6 names the copy's own tint.
 
-## Common mistakes
+## Common mistakes, and the rule that replaces each
 
 | Mistake | What it costs |
 |---|---|
--- a/.claude/skills/open-animation-replacer/evals/evals.json
+++ b/.claude/skills/open-animation-replacer/evals/evals.json
@@ -117,7 +117,7 @@
     },
     {
       "id": 11,
-      "prompt": "this DAR file is NOT IsInCombat() AND A OR B OR C - does the combat guard apply to all three or just the first?",
+      "prompt": "my DAR _conditions.txt has one function per line - NOT IsInCombat() AND, then A OR, then B OR, then C - does the combat guard apply to all three or just the first?",
       "expected_output": "Load open-animation-replacer.",
       "expectations": ["picks open-animation-replacer", "reasoning names the DAR grammar or AND/OR binding", "the answer says the guard applies to all three, because OR binds tighter than AND and A OR B OR C is one group"],
       "should_trigger": true,
--- a/.claude/skills/open-animation-replacer/evals/evals.json
+++ b/.claude/skills/open-animation-replacer/evals/evals.json
@@ -119,9 +119,9 @@
       "id": 11,
       "prompt": "this DAR file is NOT IsInCombat() AND A OR B OR C - does the combat guard apply to all three or just the first?",
       "expected_output": "Load open-animation-replacer.",
-      "expectations": ["picks open-animation-replacer", "reasoning names the DAR grammar or AND/OR binding", "the answer states the binding is unsettled rather than asserting one reading"],
+      "expectations": ["picks open-animation-replacer", "reasoning names the DAR grammar or AND/OR binding", "the answer says the guard applies to all three, because OR binds tighter than AND and A OR B OR C is one group"],
       "should_trigger": true,
-      "evidence": "NEW with this rewrite - never measured. Added for the DAR-conversion clause and the reference's new unsettled note."
+      "evidence": "NEW with this rewrite - never measured. Added for the DAR-conversion clause and the reference's AND/OR binding note."
     },
     {
       "id": 12,
--- a/.claude/skills/open-animation-replacer/evals/evals.json
+++ b/.claude/skills/open-animation-replacer/evals/evals.json
@@ -249,7 +249,7 @@
         "The run states that OAR sorts every targeting submod by priority descending and that plugin/ESP load order is ignored, and does not inspect MO2 mod order to answer who wins.",
         "The run treats 2000030002 - the DAR folder's own name - as a priority in the same global space as native OAR integers, and compares it against them.",
         "Every child array inside AND/OR/XOR in the emitted config.json is capital-C Conditions and the submod's top-level array is lowercase conditions. Graded by parsing the emitted file, not by reading the prose: a file that parses cleanly with the wrong casing is a FAIL.",
-        "The AND/OR precedence question is named, not silently resolved. A run that emits one config without saying which binding it assumed FAILS; a run that names the ambiguity, states which reading it implemented and what the other would mean, PASSES.",
+        "The AND/OR binding is applied as the settled rule, not raised as an open question: OR binds tighter than AND, so each run of OR-terminated lines becomes one OR condition and the groups are AND-ed with the remaining terms. A run that emits a flat left-to-right chain FAILS, and so does one that presents the binding as ambiguous and picks a reading.",
         "The competing-submod scan is enumerated with housecarl_asset_status and an under= glob rather than a raw filesystem walk, and the answer names the VFS winner per config."
       ],
       "arm": "output_quality",
--- a/.claude/skills/open-animation-replacer/references/oar-config-reference.md
+++ b/.claude/skills/open-animation-replacer/references/oar-config-reference.md
@@ -7,7 +7,9 @@ legacy grammar. SKILL.md is the playbook; this is the lookup table.
 
 > **OAR version studied:** 3.0.0 (the DLL shipped in the Open Animation Replacer mod, Nexus 92109).
 > Source files cited: `src/Parsing.cpp`, `src/ReplacerMods.h`, `src/Conditions.h`,
-> `src/Conditions.cpp`, `src/BaseConditions.h`, `src/API/OpenAnimationReplacerAPI-Conditions.h`.
+> `src/Conditions.cpp`, `src/BaseConditions.h`, `src/OpenAnimationReplacer.cpp`,
+> `src/API/OpenAnimationReplacerAPI-Conditions.h`. §8's DAR facts were read from that source on
+> 2026-09-08.
 
 ## Table of contents
 1. Folder layout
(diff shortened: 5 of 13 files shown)


.claude/skills/open-animation-replacer/evals/evals.json
```diff
@@ -119,9 +119,9 @@
       "id": 11,
       "prompt": "this DAR file is NOT IsInCombat() AND A OR B OR C - does the combat guard apply to all three or just the first?",
       "expected_output": "Load open-animation-replacer.",
-      "expectations": ["picks open-animation-replacer", "reasoning names the DAR grammar or AND/OR binding", "the answer states the binding is unsettled rather than asserting one reading"],
+      "expectations": ["picks open-animation-replacer", "reasoning names the DAR grammar or AND/OR binding", "the answer says the guard applies to all three, because OR binds tighter than AND and A OR B OR C is one group"],
```
Avick3110, line 122: Eval 25 (the output-quality case, line 252 of this file) still carries the pre-ruling expectation: "The AND/OR precedence question is named, not silently resolved. A run that emits one config without saying which binding it assumed FAILS; a run that names the ambiguity, states which reading it implemented and what the other would mean, PASSES."

That is now the opposite of what the skill teaches. SKILL.md:232 says the binding is settled, and the worked example records the grouping as fact rather than as one of two readings. An agent following the updated skill emits the correct config and names no ambiguity, so it FAILS this expectation — the suite now grades the correct behaviour as a failure. Eval 11 was updated in this PR; eval 25 was missed.

.claude/skills/facegen-diagnostics/evals/evals.json
```diff
@@ -177,13 +177,13 @@
       ]
     },
     {
-      "id": "st-copy-face-boundary",
+      "id": "snt-copy-face-boundary",
```
Avick3110, line 180: This case moves to the should-not arm, but the skill's own description — unchanged by this PR — still pulls it in: "Use for any discolored face … Load before judging any face bug". Its only negative clause about copying is "copying a face onto another NPC is housecarl:npc-appearance-copy", which reads as ceding the *authoring* half, not the "her head renders dark" symptom the prompt leads with.

So a run that reads the description as written has a positive trigger ("discolored face", "renders dark", "load before judging any face bug") and no matching exclusion. The case is likely to trigger above the 0.5 per-query floor and count against `specificity_threshold`, i.e. the suite records a failure for behaviour the description still asks for. Either the description needs a negative-scope clause for "a face that went dark right after a copy" (the mirror of the sentence npc-appearance-copy already carries), or the case belongs back in the should-trigger arm.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
```diff
@@ -307,13 +309,16 @@ OAR reads **Dynamic Animation Replacer** layouts at runtime and converts each in
...
-  gap in this reference, not as a detail to resolve from memory or a web search.
+- **In a mixed chain, `OR` binds tighter than `AND`.** A chain is an `AND` of `OR`-groups, not a flat
+  left-to-right sequence. Source: `Parsing.cpp`, `ConditionsTxtFile::GetConditions` (read
+  2026-09-08). The file's conditions go into one top-level set that OAR evaluates with `EvaluateAll`
+  — every entry must pass, an implicit `AND`. When a line ends in `OR`, the parser opens a nested
+  `ORCondition` and recurses into it; the recursion takes that line and each following one, and
+  closes on the first line that does **not** end in `OR`, which is itself the last member of the
+  group. An `ORCondition` evaluates with `EvaluateAny` (`Conditions.cpp`,
+  `ORCondition::EvaluateImpl`), so the group passes if any one member does. A trailing `AND` is only
+  a separator; the parser does not otherwise act on it. So: **an `OR`-group is a maximal run of lines
```
Avick3110: "A trailing `AND` is only a separator; the parser does not otherwise act on it" contradicts the two sentences immediately above it, which are the load-bearing part of the rule: inside an OR clause it is precisely the line that does *not* end in `OR` — normally an `AND`-terminated one — that closes the group and is its last member. The `AND` is what ends the recursion, so it is not inert.

A reader who takes that sentence at face value can convert a chain wrongly: given `A OR B AND C OR D`, treating the `AND` as inert suggests one group `{A, B}` and a separate `{C, D}` with the `AND` doing nothing, or — worse — that `B` sits outside the group because its `AND` "separates" it. The correct grouping is `(A OR B) AND (C OR D)` with `B` inside the first group. Drop the clause or restate it as "a line ending in `AND` closes the open `OR` group and is its last member; at top level it only separates terms."

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -227,7 +227,7 @@ Then re-read it, naming the folder again —
 `housecarl_nif_inspect(mesh_paths = ["<the placed mesh>"], sections = "paths", mod = "<that same folder>")`
 — and check that slot 6 names the copy's own tint.
 
-## Common mistakes
+## Common mistakes, and the rule that replaces each
```
Avick3110, line 230: The heading rename lands here without the content it announces. In the other eight skills each bullet carries its corrective rule (`Count the pipes and keep blank middles`, `Take the subfolder from the table`, `Discriminate SE from AE at runtime…`), so "and the rule that replaces each" is true of them. This section is a two-column table — `Mistake | What it costs` — with no rule in any row: a reader who comes to it for "the rule that replaces each" gets the cost of the mistake and has to go back up to the step sections for the fix. Either add a third column naming the rule (or the step that carries it), or leave this one at "Common mistakes".

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **`open-animation-replacer` now states how a DAR `_conditions.txt` chain binds, and that the DAR weapon-type
+  numbers convert unchanged.** Both were marked unverified in its reference and the skill told you to convert
+  under one reading and say which. Read off OAR's own source: `OR` binds tighter than `AND`, so a chain is an
```
Avick3110: This entry breaks the writing rule at the top of this same file (lines 7-12): "point at the bound rather than restating it, so the two cannot drift apart … an unpointed restatement is the defect, because it is a second copy of a fact that will be edited once."

The entry both points ("The reference's §8 carries the source files and the date they were read") *and* restates the two rulings in full — `OR` binds tighter than `AND`, a chain is an `AND` of `OR`-groups, the two type functions are hidden factories so `n` converts to `n`. Those are now second copies of §8's own text: if the source is re-read on a version bump and either ruling is refined, the changelog keeps asserting the old one. Say what changed and point at §8 for what it now says.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
+  `AND` of `OR`-groups, and `IsEquippedRightType`/`IsEquippedLeftType` are hidden factories for the same
+  `IsEquippedType` condition, so `n` converts to `n`. The reference's §8 carries the source files and the date
+  they were read. In the same pass the nine skills' mistakes section is headed "Common mistakes, and the rule
+  that replaces each", and `facegen-diagnostics` picks up the compatibility sentence every other skill carries.
```
Avick3110: "`facegen-diagnostics` picks up the compatibility sentence every other skill carries" misstates the change on two counts, and a reader checking their install against it gets the wrong picture.

It did not pick up a compatibility sentence — it already had one; this PR only spells `MO2` out as `Mod Organizer 2` in it. And "every other skill carries" that sentence is not true either: `mutagen-reference`, `papyrus-reference` and `skse-plugin-authoring` carry a longer `compatibility:` block, which is exactly what the entry directly below this one is about. "…and `facegen-diagnostics`'s compatibility line now spells out Mod Organizer 2, matching the other skills" says what actually happened.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
```diff
@@ -307,13 +309,16 @@ OAR reads **Dynamic Animation Replacer** layouts at runtime and converts each in
...
+  ending in `OR`, plus the single line after it; the groups and the remaining lines are `AND`-ed.**
```
Avick3110: The new rule is right for a file of nothing but conditions, but `GetConditions` closes an `OR` block only on a line that actually *yields* a condition, and two common line kinds do not.

In `Parsing.cpp`, the loop skips a line that is empty after `TrimWhitespace`, and `CreateConditionFromString` returns `nullptr` for a line starting with `;` (a comment). In both cases nothing is added and the `if (!bEndsWithOR && a_bInOrBlock) return` check is never reached, so the block stays open and swallows the next real condition. (An unknown function name is *not* one of these: it comes back as an `InvalidCondition`, which does close the block.)

So for

```
A OR
B

C AND
D
```

the rule as written reads the group as `A OR B` and `C`, `D` as `AND` terms, while OAR builds `(A OR B OR C) AND D`. A blank line or a `;` comment between conditions silently widens the group by one — exactly the class of silent mis-conversion this section exists to prevent, and DAR files in the wild carry both.

Suggest tightening the closing sentence to the line that yields a condition, e.g. "an `OR`-group is a maximal run of lines ending in `OR`, plus the next line that yields a condition — blank lines and `;` comments are skipped without closing the group".

.claude/skills/facegen-diagnostics/evals/evals.json
```diff
@@ -177,13 +177,13 @@
...
       "prompt": "i copied a face from an overhaul onto my custom follower and now her head renders dark. did i copy the wrong fields, and what's the right way to do that copy?",
-      "expected_output": "facegen-diagnostics is loaded on the dark-face symptom, and the copy itself is ceded to housecarl:npc-appearance-copy.",
+      "expected_output": "housecarl:npc-appearance-copy, not this skill — a dark face right after a copy is the copy's own case, and the prompt asks how to do the copy.",
       "files": [],
       "expectations": [
-        "The facegen-diagnostics skill is loaded.",
-        "The authoring half is routed to housecarl:npc-appearance-copy rather than reinvented here.",
+        "The facegen-diagnostics skill is NOT loaded.",
+        "The copy is handled by housecarl:npc-appearance-copy rather than by a diagnosis flow.",
```
Avick3110: This expectation is a shape no other should-not case in this file uses, and it cannot pass in one of the file's own configurations.

The other twenty `snt-` cases name the skill that should have taken the prompt in `expected_output` (`snt-skypatcher-npc-face`: "housecarl:skypatcher-authoring, not this skill.") and then grade only two things: that facegen-diagnostics is NOT loaded, and how the answer reasons (`snt-player-grey-after-update`, `snt-body-neck-seam-cbbe`, `snt-mesh-winner-generic`). None of them makes another skill's presence a graded condition. This one does — and `method.configurations.isolated` is "this skill alone, from the packaged tree", where `housecarl:npc-appearance-copy` is absent by construction, so the expectation must FAIL there however well the run behaves. Grading is declared binary with no partial credit, so that is a guaranteed failure in half the configurations.

Suggest reshaping it the way the file's other should-not cases are shaped — put the routing in `expected_output` (it already is) and grade the answer's reasoning instead, e.g. "the copy is treated as an authoring job rather than a diagnosis flow".

Related, and not addressed anywhere in the diff: the `isolation` gate dimension at line 45 still lists "a face copy" as one of the three cede points these cases cover, and this was the only case with a copy in it. With it on the should-not arm, nothing in the set now exercises that cede with the skill loaded, so that dimension's `covered_by` claim no longer holds.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **`open-animation-replacer` now states how a DAR `_conditions.txt` chain binds, and that the DAR weapon-type
+  numbers convert unchanged.** Both were marked unverified in its reference and the skill told you to convert
+  under one reading and say which. Read off OAR's own source: `OR` binds tighter than `AND`, so a chain is an
+  `AND` of `OR`-groups, and `IsEquippedRightType`/`IsEquippedLeftType` are hidden factories for the same
+  `IsEquippedType` condition, so `n` converts to `n`. The reference's §8 carries the source files and the date
+  they were read. In the same pass the nine skills' mistakes section is headed "Common mistakes, and the rule
```
Avick3110: Every other bullet under `## Unreleased` is one change with one bolded lead that names it. This one leads on the OAR binding facts and then carries two more, unrelated, changes in its last sentence — the heading rename across nine skills, and facegen's compatibility string.

The facegen half is the same kind of change as the very next bullet ("`mutagen-reference` and `papyrus-reference` now say in their frontmatter …" — a frontmatter compatibility sentence), so a reader scanning bold leads for "what changed about compatibility strings" finds one bullet and misses the other. Suggest either folding the facegen sentence into that neighbouring bullet or giving it its own, and dropping the heading rename, which changes nothing a user of the plugin can observe.

Also, "the nine skills' mistakes section" pins a count that goes stale the next time a skill is added or one grows the section; the entry does not say which nine, so the count cannot be checked from the entry either.

.claude/skills/open-animation-replacer/evals/evals.json
Avick3110: Folded in 86467a90. Eval 25's fourth expectation now grades the settled rule instead of the ambiguity: `OR` binds tighter than `AND`, each run of `OR`-terminated lines becomes one `OR` condition, the groups are `AND`-ed with the remaining terms. A flat left-to-right chain fails, and so does a run that calls the binding ambiguous and picks a reading. The expectation still discriminates against the control arm, so the pass rule is unchanged.

.claude/skills/facegen-diagnostics/evals/evals.json
Avick3110: Left for Aaron: the description's cede clause is ruled text; the advisor has put the extension to him.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
Avick3110: Folded in 7951f082. The clause was true of the parser and misleading about the effect: `GetConditions` only ever tests `ends_with("OR")`, so it never acts on `AND` as a token, but a line ending in `AND` is exactly a line that does not end in `OR`, which is what closes an open group. The rule now says both halves — inside an open group an `AND`-terminated line closes it and is its last member; at top level it separates terms — with the parenthetical that the test is for the trailing `OR`, so nothing invites the reading that `B` sits outside the group.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded in 1a4108b5, the other way round: the heading rename is ruled for all nine skills, so the table gets the third column rather than the heading going back. Each row now carries the rule that replaces the mistake — read `Configuration.TemplateFlags` first and copy from the record in `Template`; pass `Race:refuse` yourself and bundle `Race` only when the races differ; and so on down to Step 4's repoint.

plugin/CHANGELOG.md
Avick3110: Folded in 523e711c. The entry now says what changed and points: both facts were marked unverified, both are now read off OAR's own source, and §8 carries the rule, the source files and the date they were read. The two rulings are no longer restated, so §8 is the only copy.

plugin/CHANGELOG.md
Avick3110: Folded in 523e711c. Both counts were wrong as you say — facegen already carried a compatibility sentence, and the change was `MO2` becoming `Mod Organizer 2`. It is now its own bullet: "**`facegen-diagnostics`'s compatibility line now spells out Mod Organizer 2.** It read `MO2`; the line now matches the string every other skill carries." No claim about what every other skill carries beyond the shared string itself.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
Avick3110: Confirmed against the source and folded in 7951f082. `Parsing.cpp` `GetConditions` guards the whole body with `if (!a_currentLine.empty())` after `TrimWhitespace`, so a blank line never reaches the close test; `Conditions.cpp` `CreateConditionFromString` returns `nullptr` on `a_line.starts_with(";")`, so a comment line fails the `if (auto newCondition = ...)` and the `!bEndsWithOR && a_bInOrBlock` return is never reached. An unrecognised function name falls through to `std::make_unique<InvalidCondition>(errorStr)`, which is non-null and does close the group, as you say.

The rule now reads "an `OR`-group is a maximal run of lines ending in `OR`, plus the next line that yields a condition; the groups and the remaining lines are `AND`-ed", with the two skipped line kinds and the `InvalidCondition` exception named above it. SKILL.md's DAR example repeated the old wording and is updated to match. The worked example itself has no blank or comment lines, so its grouping was unaffected. §8's source line now cites `Conditions.cpp` `CreateConditionFromString` alongside `GetConditions`.

.claude/skills/facegen-diagnostics/evals/evals.json
Avick3110: Folded in 8c971bba. The expectation no longer grades another skill's presence; it grades the answer the way the file's other `snt-` cases do: "The copy is treated as an authoring job — which fields to seed and what the copy still owes — rather than as a diagnosis flow over the load order." The routing stays in `expected_output`, and the third expectation about the FaceGen carry is unchanged, so the case passes or fails on the run in both configurations.

The `isolation` gate dimension is fixed in the same commit: the face copy comes out of its cede list and out of the absent-skill list, and the dimension now says outright that the copy cede is not covered here because this case grades the answer rather than another skill's presence. That is the honest state of the set rather than a `covered_by` claim nothing exercises.

plugin/CHANGELOG.md
Avick3110: Folded in 523e711c. The bullet is now one change with one bolded lead: the DAR binding and the type numbers, pointing at §8. facegen's compatibility line is its own bullet beside the frontmatter one. The heading rename is dropped from the changelog — it changes nothing a user of the plugin does differently, and the stale count went with it.

Avick3110:
Blind review folded. Eight of the nine findings folded, one left for Aaron, none refuted.

Folded:
- The DAR `OR`-group rule was wrong for two common line kinds. Verified against OAR's source: `GetConditions` skips a line that is empty after `TrimWhitespace` before the close test, and `CreateConditionFromString` returns `nullptr` on a line starting with `;`, so neither closes an open group; an unknown function name yields an `InvalidCondition`, which does. The rule now reads "a maximal run of lines ending in `OR`, plus the next line that yields a condition", with the skipped kinds and the exception named. The "trailing `AND` is only a separator" clause is replaced by what the parser does: it tests only for a trailing `OR`, so an `AND`-terminated line closes an open group as its last member and separates terms at top level. SKILL.md's DAR example repeated the old wording and is updated; the worked example has no blank or comment lines, so its grouping was unaffected. §8 now cites `Conditions.cpp` alongside `Parsing.cpp` (7951f082).
- OAR eval 25 still graded the pre-ruling ambiguity as the pass condition. It now grades the settled binding (86467a90).
- facegen's `snt-copy-face-boundary` made another skill's presence a graded condition, which cannot hold in the isolated configuration. It now grades the answer's reasoning, the way the file's other `snt-` cases do, and the `isolation` gate dimension no longer claims the copy cede is covered (8c971bba).
- npc-appearance-copy's `Mistake | What it costs` table gets a third column, so the ruled heading is true of it (1a4108b5).
- The three changelog findings: the OAR entry no longer restates the two rulings, carries one change with one bolded lead, and drops the stale nine-skill count; facegen's compatibility line is its own bullet saying what actually changed (`MO2` spelled out), and the heading rename is dropped as nothing a user does differently (523e711c).

Left for Aaron: whether facegen-diagnostics' description extends its cede clause to a face gone dark right after a copy. The description is ruled text; the advisor has put the extension to him.

Green from inside the worktree: `scripts/build-plugin.ps1` (leak-check clean, 12 skills / 227 markdown files), `dotnet build -c Release` 0 errors, 1,890 tests passed, `ci-all` 128/128.

The branch is 7 commits behind the tip, which moved today with #651 on `plugin/CHANGELOG.md`. Not rebased.

.claude/skills/facegen-diagnostics/evals/evals.json
Avick3110: Folded in 2af20ff6 on Aaron's ruling of 2026-09-08 (evening): the cede clause now reads '..., or a face gone dark right after a copy, ...'.

.claude/skills/npc-appearance-copy/SKILL.md
```diff
@@ -227,17 +227,17 @@ Then re-read it, naming the folder again —
...
+| Omitting `TextureLighting` | Every field reads correct and the skin renders dark | Carry Step 2's bundle whole, `TextureLighting` included |
+| Dropping a bundle member the donor lacks instead of clearing it | The target keeps its own morphs under the donor's head parts — a face built from two people | Copy what the donor has and clear what it lacks — `Remove`, or `ReplaceAll` with `composes: []` for `TintLayers` |
+| Stopping at the patch | The records exist, the FaceGen does not — a dark face you authored on purpose | The copy is four calls; Step 3 places the FaceGen pair |
+| Stopping at the placement | The placed mesh still points at the donor's tint: correct until the donor is uninstalled | Finish with Step 4 — repoint the tint inside the placed mesh, naming `mod=` |
 | Letting the FaceGen source default to the VFS winner | You place a replacer's face over the records of the donor you actually copied |
 | Reporting "copied" when the strip list is long | A clone with no factions, outfits, packages or scripts is not a working follower |
```
Avick3110: These two rows never got the new third column. The header is `| Mistake | What it costs | The rule that replaces it |` over a `|---|---|---|` separator and rows 1-7 supply three cells; these supply two. Markdown pads the missing cell instead of erroring, so the shipped table renders two blank rules - and they are exactly the two mistakes an agent consulting the table for a rule gets none for.

```suggestion
| Letting the FaceGen source default to the VFS winner | You place a replacer's face over the records of the donor you actually copied | Name `from_plugin=` on the FaceGen placement so the source is the donor you copied, not the VFS winner |
| Reporting "copied" when the strip list is long | A clone with no factions, outfits, packages or scripts is not a working follower | Report the strip list as work still owed, item by item, before calling the copy done |
```

Avick3110: Correction to my own suggestion above: the parameter is `source_provider`, not `from_plugin` - ignore that half of the block. The finding stands (both rows are missing the third cell); the rule for the first row should read along the lines of "Name `source_provider` explicitly, chosen by inspecting each candidate with `housecarl_nif_inspect mod=` and keeping the copy whose baked shape names match - see Step 3's 'Which provider to name'."

.claude/skills/open-animation-replacer/references/oar-config-reference.md
```diff
@@ -307,13 +309,22 @@ OAR reads **Dynamic Animation Replacer** layouts at runtime and converts each in
...
+  top-level set that OAR evaluates with `EvaluateAll` — every entry must pass, an implicit `AND`.
+  When a line ends in `OR`, the parser opens a nested `ORCondition` and recurses into it; the
+  recursion takes that line and each following one, and closes on the first line that **yields a
+  condition** without ending in `OR` — that line is itself the last member of the group. An
+  `ORCondition` evaluates with `EvaluateAny` (`Conditions.cpp`, `ORCondition::EvaluateImpl`), so the
+  group passes if any one member does. The parser tests only for a trailing `OR`, so a line ending
+  in `AND` is simply a line that does not: inside an open group it closes the group and is its last
+  member, and at top level it separates terms. Two line kinds yield no condition and are skipped
+  **without** closing an open group — a line that is empty after trimming, and a line starting with
+  `;` — so a blank line or a comment inside an `OR` run widens the group by one. An unknown function
```
Avick3110: `so a blank line or a comment inside an OR run widens the group by one` is not what the parser does, and it contradicts the rule stated two sentences later.

I read `ConditionsTxtFile::GetConditions` at `ersh1/OpenAnimationReplacer@main` to check. An empty-after-trim line fails the `if (!a_currentLine.empty())` guard outright; a `;` line reaches `CreateConditionFromString`, which returns `nullptr` at its first statement. In both cases nothing is added to the set and the `if (!bEndsWithOR && a_bInOrBlock) return` is never reached - so the skipped line is a no-op, not a widener.

Work `A OR / B OR / <blank> / C / D` through: the group is (A, B, C) and D is a top-level AND term, byte-identical to the same file with the blank line deleted. An agent applying the sentence as written pulls D in too - group (A, B, C, D), no AND term at all - which is the over-broad conversion the rest of the section exists to prevent.

The preceding clause ("skipped **without** closing an open group") is correct and is the whole of the fact; the conclusion drawn from it is the error. SKILL.md's summary of the same rule at line 231 has no such clause and is fine as it stands.

```suggestion
  `;` — a skipped line neither joins the group nor closes it, so the group is exactly what it would
  be with that line deleted. An unknown function
```

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **`open-animation-replacer` now states how a DAR `_conditions.txt` chain binds, and that the DAR weapon-type
+  numbers convert unchanged.** Both were marked unverified in its reference, and the skill told you to convert
+  under one reading and say which. Both are now read off OAR's own source; the reference's §8 carries the rule,
+  the source files and the date they were read.
```
Avick3110, line 19: `## Unreleased` covers the OAR facts and the compatibility-line wording, but not the fourth change in this PR that a user actually notices: `facegen-diagnostics`'s description now cedes a face gone dark right after a copy to `housecarl:npc-appearance-copy`. That is a routing change - the same prompt that used to load facegen-diagnostics and run the mesh-versus-tint flow now loads the copy skill instead, and `snt-copy-face-boundary` was flipped to assert it. Per CLAUDE.md that earns a line here.

Related: the PR body says "The description's cede clause is unchanged", which the diff and commit 2af20ff6 ("Cede a dark face right after a copy in the description") both contradict - worth fixing so the body matches what merged.

The mistakes tables also gained a whole rule column across the skills; your call whether that reaches the bar for its own line.

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -1,8 +1,8 @@
...
+  Diagnoses and repairs the dark / grey / black-face NPC bug in Skyrim SE by diffing which mod wins the head .nif against which wins the face .dds for the same NPC, then the record winner behind them. Use for any discolored face, neck seam, "fine in xEdit but wrong in game", NPCs gone dark after an ESL-compaction or merge, or any FaceGen / facegeom / facetint mention. Not a purple or white face (a missing texture) and not player-only grey (RaceMenu/SKEE); copying a face onto another NPC, or a face gone dark right after a copy, is housecarl:npc-appearance-copy. Load before judging any face bug — the mesh-versus-tint pair decides the fix.
```
Avick3110, line 5: The description now cedes two things - the copy, and a face gone dark right after a copy - but §11's routing table (line 263, outside this diff) still cedes only one:

    | Copying a face onto a *different* NPC, or cloning one as a standalone | `housecarl:npc-appearance-copy` |

The description is a trigger heuristic, not a gate, so the skill will still load on prompts where a copy is mentioned in passing; once loaded, the body is what the agent follows, and §11 tells it to hand off only the authoring half and diagnose the rest. The two halves of the same file then give opposite instructions on the case this ruling was about. The evals file has already moved to the new boundary (the isolation `failure_mode` was rewritten to drop the copy cede), which leaves the body as the only place carrying the old one.

Widening that row to something like "Copying a face onto a *different* NPC, cloning one as a standalone, or a face gone dark right after either" would close it.

.claude/skills/open-animation-replacer/evals/evals.json
Avick3110, line 122: Lower confidence than the others, but worth a look: the expectation now asserts one grouping unconditionally, while the prompt (unchanged) writes the chain inline - `NOT IsInCombat() AND A OR B OR C` - with no line breaks. The binding rule this PR verified is defined entirely over line boundaries: `GetConditions` tests `a_currentLine.ends_with("OR")` per line, and a chain that really is one line goes to `CreateConditionFromString` whole and yields a single condition.

So a run that answers "as four lines the guard covers all three; written as one line it is not a chain at all" is more correct than the expectation and grades FAIL. The old expectation ("states the binding is unsettled") did not depend on line structure, so this is new with the change. Splitting the prompt across lines, or saying "one function per line" in it, would make the case grade what it means to.

.claude/skills/npc-appearance-copy/SKILL.md
Avick3110: Folded in 51d943e4. Both rows now carry a third cell. Took your correction: the first reads "Name `source_provider` explicitly, chosen by inspecting each candidate with `housecarl_nif_inspect mod=` and keeping the copy whose baked shape names match — Step 3's \"Which provider to name\"". Checked the name against the tool surface (`source_provider`, and `nif_inspect` takes `mod=`) and against the skill body — the guidance is under Step 3, line 193. The second row takes your wording as written. Swept the other skills for the same defect: no other mistakes table has a row short of its header.

.claude/skills/open-animation-replacer/references/oar-config-reference.md
Avick3110: Folded in 4801470a, your wording verbatim. Confirmed against `Parsing.cpp` before taking it: the empty-after-trim line fails the `if (!a_currentLine.empty())` guard, and the `;` line reaches `CreateConditionFromString` but returns nullptr, so the `if (!bEndsWithOR && a_bInOrBlock) return` never runs — it sits inside the `if (auto newCondition = ...)` block. Neither adds a member nor closes the group, so `A OR / B OR / <blank> / C / D` is group (A, B, C) with D a top-level AND term, same as with the line deleted. SKILL.md line 231 left alone; it carries the run-plus-next-line rule with no widening claim.

plugin/CHANGELOG.md
Avick3110: Folded in cbbe96e8. Added the routing line for the facegen cede, pointing at the two places the skill routes from rather than restating the rule.

On the mistakes table: added a line, one sentence. It only reached `npc-appearance-copy` — the rule column is that one table, not all the skills (dd6da718 touches one file) — but the compatibility-line entry sets the bar, and an agent consulting the table now gets a rule where it got nothing.

PR body fixed. Dropped "The description's cede clause is unchanged" and replaced it with what 2af20ff6 and 458e684f actually did. While there: "Bullets untouched" in the heading bullet was wrong the same way, so it now names the rule column; test count updated to 1894. Everything else kept.

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded in 458e684f. The §11 row now reads "Copying a face onto a *different* NPC, cloning one as a standalone, or a face gone dark right after either" — your wording. Description and body now cede the same two cases.

.claude/skills/open-animation-replacer/evals/evals.json
Avick3110: Folded in e276bc71 (this is eval id 11; 25 is the case count). You are right — `GetConditions` reads with `std::getline` and tests `ends_with("OR")` per line, so a chain on one line goes to `CreateConditionFromString` whole and is not a chain at all. The expectation asserted a grouping the prompt did not set up.

Took the "one function per line" option rather than embedding newlines: no other prompt in any evals.json carries a `\n`, so the JSON stays the shape the harness already sees. The prompt now reads "my DAR _conditions.txt has one function per line - NOT IsInCombat() AND, then A OR, then B OR, then C - does the combat guard apply to all three or just the first?"

PR #659, opened 2026-09-08
Rebuild the Codex umbrella skill as a router

Avick3110:
Wave 4 of the skill rewrite: the Codex umbrella at `plugin/codex/housecarl`, built from the approved rewrite document and its verifier (110 rows checked, 14 overturned — each overturned row is taken as the verifier corrected it, listed below). The body catalogued the whole tool surface in ~39 lines, seventeen of those names retired with 1.x, under a sentence promising they "keep working through the 2.0 build" — a Codex session following it composed calls the server refuses by name, and the CI guard watching this file could not see it because it only checked that every tool appeared, never that every name written was real. The catalogue is deleted rather than moved (FOLD-IN F37: it restated the server's own injected instructions, which already end "each tool's own description carries the specifics", and ruling 4 bars a tool description as a destination), and so are the write-lane rules (F38) and the in-place prohibition (F39), which the server's own parameters and its `acknowledge=` consent gate already carry. What is left is what only a router can do: a row per bundled skill saying which job it owns and when to load it (eleven rows after ruling 6b, with no count written), the read-before-you-write order in four steps with one worked call and its result, the FormID form, the three lanes, an exclusion clause naming #615 and #617 for the two reaches the surface does not have, and one line for the `agents/openai.yaml` sidecar. The description goes from 1,267–1,273 characters to **611** against the 1,024 ceiling, opens on its trigger at offset 0 instead of offset ~863, drops the fourteen-name helper roster and the "even when the task looks like a single edit, load this first" clause that asserted a precedence no host enforces, and gains the batch `compatibility` line and a closing "Not for…" boundary. `evals/evals.json` is new — this skill has never had an eval set, and the 2026-09 audit ran neither arm for it, the only row in the batch with both cells "not run". Every 1.x call in the old body is migrated to the 2.0 call the document's table gives; every tool and parameter written in the tree appears verbatim in `surface/tools.md`. No issue is named as closing.

**Checks.** `scripts/build-plugin.ps1 -PluginTreeOnly` — leak-check clean, 12 skills / 227 markdown files across both trees, `evals/` stripped from the shipped Codex copy (`dist/codex/skills/housecarl/` ships `SKILL.md` + `agents/openai.yaml` and nothing else). `dotnet build -c Release` clean; tests **1894 passed, 0 failed** (`tier!=bridge`); generator `ci-all` **128/128**, the umbrella guard 13 arms green.

- **Q1 — does the router survive at all, and does the measurement come first? Left for Aaron.** No verdict was ruled ("The codex umbrella is item 27"), both audit arms are unrun, and D5.26/D5.50 make it the batch's strongest fold-or-delete candidate. This PR is the rewrite the document specifies; it does not argue the router should exist. The output-quality pair is written up in `evals/evals.json` and has not been run — if the control matches, the ruled outcome is FOLD-IN N9 (the map is dropped, not migrated), and that is a merge gate, not a rewrite of this branch.
- **Q2 — the sibling form on a Codex-only body. Decided as follows: ruled, not open** (verifier row 10.2). Ruling 1 states the form for every document in full knowledge of the two install lanes, so the body writes `housecarl:<skill>` with the one-line Codex note. Worth knowing for the review: the eleven siblings **do** install on Codex too, flat at `~/.agents/skills/<slug>` beside this one (`Program.cs`, the Codex install), so the note names a form the reader really sees there.
- **Q3 — directory submission or local install only? Left for Aaron.** Nothing in the tree assumes either. Ruling 7's packager PR already gave the Codex tree its `skills/` root, so the shape half is answered; submission would additionally make the eval case counts mandatory rather than borrowed.
- **Q4 — `policy.products` in `agents/openai.yaml`. Left for Aaron.** Unchanged in this PR; no source says it is mandatory.
- **Q5 — does the Codex bundle get `evals/evals.json` at all? Decided as follows: ruled, not open** (verifier row 10.5). Ruling 3 is one convention for all twelve. The file is authored, and what was genuinely open — that the numbers are Anthropic's — is recorded in the file itself under `numbers_are_borrowed`.
- **Q6 — the mandatory-framing enforcement route. Left for Aaron.** The "load this first" clause is deleted rather than backed; no `AGENTS.md` if/then rule is added, because the plugin tree ships none and adding one is a packaging change outside this document.
- **Q7 — do Codex sessions receive the same ServerInstructions? Left for Aaron, and it is the one dependency this PR takes on faith.** F1, F2 and F3 (the distributor routing, "nothing wins until enabled in MO2", the never-copy rule) left this body on the assumption that the injected instructions carry them. The Codex install registers the server in `~/.codex/config.toml`, so it should; if a Codex session does not receive the instructions string, three cross-skill facts are then in no file this bundle ships.

## The verifier's overturned rows, and how each was taken

- **3.12 `compatibility` wording** — corrected to the batch string ruled on 2026-09-08: "Requires the houseCARL MCP server and a configured Mod Organizer 2 instance." (76 characters). No second sentence: nothing here works offline.
- **3.7 / 3.13** — the description enumerates 15 terms, not 14, and its shape (third-person capability clause plus imperative trigger) **clears** D2.26 rather than standing as a departure. No text change; the claims were about the same string.
- **4.1 body budget** — the section table summed to 4,860 against its own stated ≤ 4,800. The stated budget wins: the body is **4,773 characters**, 59 content lines (74 including blank lines) against ≤ 70. The trims came out of prose, not out of a section.
- **6.1 / 6.2 / 6.4 / 6.18 / 7.5 site citations** — off-by-one line numbers in the *old* body, no migration target moved. Moot in the build: the old lines are deleted whole.
- **9.4 D1.15** — the clause claiming a sibling-name partial is dropped; it is graded conforms here.
- **9.6 D3.5** — recorded as **reduced and ruled**, not cleared: 620 characters ≈ 155 tokens clears both enforceable budgets and is still over the spec's "~50-100 tokens per skill" estimate.
- **9.11 the positive set** — the document's eight positives included an OAR case the description carries no term for. Decided as follows: **the positive set is re-cut against the terms the description actually carries**, and the description is left at its ruled 611 characters, because widening it is what D2.21 / D3.5 / D4.59 were failing on and both measurements feed the regrade gate. `evals/evals.json` records under `method.positives_cover` that three routing destinations — `open-animation-replacer`, `npc-appearance-copy`, `papyrus-reference` — have no positive, and why: each ships its own trigger description beside this one on both hosts.
- **10.2 / 10.5** — Q2 and Q5 recorded above as ruled rather than put to Aaron.

## Judgement calls worth a reviewer's eye

- **The CI guard changed direction, and that is the one change outside the skill.** `CodexUmbrellaCoverageProbe.cs`'s INV1 required every reflected tool name to be referenced in the router — the invariant a deleted catalogue makes unsatisfiable, and the reason `ci-all` was green through seventeen dead names (D5.51). It now asserts the reverse: every `housecarl_*` identifier the router writes resolves to a live `[McpServerTool]` name. INV2 (every bundled skill routed) is unchanged, the allow-list now suppresses skill slugs only, and the new arm has both teeth (a retired name in a synthetic router is caught) and a green-self arm (a live name is not reported dead).
- **The worked call is an `apply` with `readback=true`, not a read.** D5.35 graded the read half of the loop as the one well-built thing in the file and the write half as missing; the closing paragraph is the write half — the read-back is the written *file*, the patch name is auto-suffixed so it is reported back, and a refused call wrote nothing.
- **#617 is named as an issue but framed loosely**, as the verifier noted (row 4.10): it is a refusal-message gap rather than a job the surface cannot do. The body says the raw mods-folder path is not refused in one sentence and gives the move that works instead.
- **The eleven-row table writes no count**, per D4.43 — adding a skill is a row, never an arithmetic edit.

## Regrade

Five-checklist regrade of this branch as it stands on 2026-09-08, assembled in `dev/skill-validation/eval-2026-09/codex-umbrella/REGRADE.md` from the per-angle graders in `regrade/D1.md`–`D5.md`. This replaces the earlier regrade block, which was taken against a previous state of the worktree — the worked call has since been corrected to `012EB7:Skyrim.esm`, and D4.50 is no longer a factual fail.

| Angle | conforms | partial | fails | n/a | Total rows |
|---|---|---|---|---|---|
| D1 — discovery and packaging | 36 | 3 | 1 | 6 | 46 |
| D2 — frontmatter | 39 | 2 | 0 | 11 | 52 |
| D3 — loading and invocation | 20 | 3 | 0 | 10 | 33 |
| D4 — body, supporting files, style | 36 | 6 | 1 | 23 | 66 |
| D5 — validation and standing context | 32 | 8 | 5 | 8 | 53 |
| **Total** | **163** | **22** | **7** | **58** | **250** |

Fails by class: **0 factual-or-dead-call**, 1 packager-only (D1.28, release tagging, excluded by ruling 8), 6 other (D4.42, the ruled bare-tool-name convention; D5.19, D5.21, D5.26, D5.33, D5.52, the single absent eval run counted five times).

**Verdict: PASS** against the regrade half of the gate. D2 and D3 have zero fails. D4 and D5 have zero factual-or-dead-call fails — every one of the ten `housecarl_*` names and ten parameters the body writes is live on `surface/tools.md`, and the one file the body names ships beside it. D1 is clean except for D1.28, which ruling 8 excludes as item 30's walk. What remains is the bare-tool-name convention ruling 1 settled the other way, and five restatements of one absence: the eval set is authored and has not been run, which §9b–9c assign to the run rather than to this rewrite.

### DIFF
--- a/plugin/codex/housecarl/SKILL.md
+++ b/plugin/codex/housecarl/SKILL.md
@@ -55,15 +55,17 @@ Claude Code invocation — on Codex it is the bare folder name (`facegen-diagnos
 4. **Read the written record back**, and say what happened when it did not take.
 
 ```
-housecarl_apply(ops=[{"formid": "013BA3:Skyrim.esm", "field_path": "BasicStats.Damage", "value": 12}], patch="SwordFix", readback=true)
+housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": 12}], patch="SwordFix", readback=true)
 ```
 ```
-wrote SwordFix.esp   1 record, 1 op   epoch=7f3a1c
-  013BA3:Skyrim.esm  IronSword  BasicStats.Damage  10 -> 12
+wrote SwordFix.esp (new patch; 1284 bytes)
+full read-back — the ENTIRE record(s) as written, re-read from the patch file (NOT load-order truth):
+  Weapon 012EB7:Skyrim.esm  editorid=IronSword
+    BasicStats.Damage = 12
+    ...
 ```
 
-The read-back is the written FILE, not load-order truth. Report the patch name back — it is auto-suffixed
-when taken — and tell the user to enable it. A refused call wrote nothing: fix the path and send it again.
+Report the patch name back — it is auto-suffixed when taken. A refused call wrote nothing.
 
 ## Lanes and FormIDs
 
--- a/plugin/codex/housecarl/SKILL.md
+++ b/plugin/codex/housecarl/SKILL.md
@@ -17,9 +17,8 @@ compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 
 
 **In:** a live Mod Organizer 2 instance, read through the houseCARL MCP server. **The work:** read the
 record at its true load-order winner, check the schema, then write. **Out:** a patch plugin the user
-reviews and enables in MO2. This file does two things and defers the rest — it routes a job to the skill
-that owns its grammar, and it gives the read order every write depends on. What a tool takes is in that
-tool's own description.
+reviews and enables in MO2. This file routes a job to the skill that owns its grammar, and gives the read
+order every write depends on. What a tool takes is in that tool's own description.
 
 The user's instructions take precedence over guidelines provided in a skill.
 
@@ -40,38 +39,40 @@ The user's instructions take precedence over guidelines provided in a skill.
 | A catalogue, audit, conflict survey or link graph over many records | `housecarl:bulk-record-jobs` | before the first call |
 
 MCP tools are written bare on both hosts (`housecarl_records`); a sibling is written `housecarl:<skill>`, the
-Claude Code invocation — on Codex it is the bare folder name (`facegen-diagnostics`), installed beside this one.
+Claude Code form — on Codex it is the bare folder name (`facegen-diagnostics`) installed beside this one.
 
 ## Read before you write
 
 1. **Confirm context when it matters.** `housecarl_load_order_status` says what is active;
-   `housecarl_set_mo2_instance` when the user names a different MO2 instance folder.
+   `housecarl_set_mo2_instance` when the user names another MO2 instance folder.
 2. **Read the winner and the schema.** `housecarl:mutagen-reference` for the field path and its legal
-   values — if it has no entry for a type, say so rather than guessing; `housecarl_records` for the record
-   as the order resolves it, `project={"form":"tree"}` for every provider when the winner is contested.
+   values — no entry for a type means say so, never guess; `housecarl_records` for the record
+   as the order resolves it, `project={"form":"tree"}` for every provider when contested.
 3. **Write with one verb.** `housecarl_apply` edits fields (`ops=`), `housecarl_create` mints records
    (`records=`), `housecarl_remove` drops them (`formids=`), `housecarl_forward` carries another plugin's
-   record as an override. Every list is set-valued — one op is a set of one — there is no single/bulk pair.
-4. **Read the written record back**, and say what happened when it did not take.
+   record as an override. Every list is set-valued — one op is a set of one; no single/bulk pair.
+4. **Read the written record back**, and say what happened if it did not take.
 
 ```
 housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": 12}], patch="SwordFix", readback=true)
 ```
 ```
 wrote SwordFix.esp (new patch; 1284 bytes)
-full read-back — the ENTIRE record(s) as written, re-read from the patch file (NOT load-order truth):
+mod folder: SwordFix  — enable + sort it in MO2 to use the patch
+full read-back — … NOT load-order truth; the patch wins nothing until enabled + sorted in MO2:
   Weapon 012EB7:Skyrim.esm  editorid=IronSword
     BasicStats.Damage = 12
     ...
 ```
 
-Report the patch name back — it is auto-suffixed when taken. A refused call wrote nothing.
+Report the patch name and mod folder back — the name is auto-suffixed when taken. A refused call wrote
+nothing — except in place, where the file may already have changed.
 
 ## Lanes and FormIDs
 
 A FormID is `XXXXXX:Plugin.esp` — six hex digits, then the defining master's filename. The runtime form a log
 or the console prints is read-only: `housecarl_records` takes it, every write refuses it and names the
-`XXXXXX:Plugin.esp` form to use. SkyPatcher, SPID and KID write their own syntax.
+`XXXXXX:Plugin.esp` form. SkyPatcher, SPID and KID write their own syntax.
 
 `housecarl_apply`, `housecarl_create` and `housecarl_forward` take three lanes: `patch=` a new plugin, `into=`
 an existing houseCARL patch, `in_place=` the file it names. `housecarl_remove` edits only what exists: `into=`
@@ -79,13 +80,13 @@ or `in_place=`, no `patch=`. In place is consent-gated per plugin by `acknowledg
 
 ## Where this does not apply
 
-Another game; installing or configuring MO2; gameplay advice with no record, file or INI in it. Two reaches the
-surface does not have today, both filed: `housecarl_check` with `findings=["dialogue"]` resolves against the
-active order, so a plugin not yet enabled cannot be dialogue-checked (#615); and a hand-composed raw mods-folder
-path is not refused in one sentence by `housecarl_place` or `housecarl_nif_inspect` (#617) — pass the mod
-folder a read-back named instead.
+Another game; installing or configuring MO2; gameplay advice with no record, file or INI. Two reaches the
+surface lacks today, both filed: `housecarl_check` with `findings=["dialogue"]` resolves against the active
+order, so a plugin not yet enabled cannot be dialogue-checked (#615); and a hand-composed raw mods-folder path
+is not refused in one sentence by `housecarl_place` or `housecarl_nif_inspect` (#617) — pass the mod folder a
+read-back named.
 
 ## The sidecar
 
-Codex reads `agents/openai.yaml` beside this file for the display name and invocation policy; nothing in
-this body depends on it.
+Codex reads `agents/openai.yaml` beside this file for the display name and invocation policy; nothing here
+depends on it.
--- a/plugin/codex/housecarl/SKILL.md
+++ b/plugin/codex/housecarl/SKILL.md
@@ -54,7 +54,7 @@ Claude Code form — on Codex it is the bare folder name (`facegen-diagnostics`)
 4. **Read the written record back**, and say what happened if it did not take.
 
 ```
-housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": 12}], patch="SwordFix", readback=true)
+housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": "12"}], patch="SwordFix", readback=true)
 ```
 ```
 wrote SwordFix.esp (new patch; 1284 bytes)
--- a/plugin/codex/housecarl/SKILL.md
+++ b/plugin/codex/housecarl/SKILL.md
@@ -67,13 +67,13 @@ when taken — and tell the user to enable it. A refused call wrote nothing: fix
 
 ## Lanes and FormIDs
 
-A FormID is `XXXXXX:Plugin.esp` — six hex digits, then the filename of the master that defines the record.
-The runtime form a log or the console prints is taken too, wherever a parameter holds nothing but FormIDs.
-SkyPatcher, SPID and KID each write their own syntax; their skills say so.
+A FormID is `XXXXXX:Plugin.esp` — six hex digits, then the defining master's filename. The runtime form a log
+or the console prints is read-only: `housecarl_records` takes it, every write refuses it and names the
+`XXXXXX:Plugin.esp` form to use. SkyPatcher, SPID and KID write their own syntax.
 
-Every write tool has the same three lanes: `patch=` writes a new plugin, `into=` extends an existing houseCARL
-patch, `in_place=` overwrites the file it names. In place is consent-gated at the server, per plugin, by
-`acknowledge=` — it refuses rather than asking this skill to police it.
+`housecarl_apply`, `housecarl_create` and `housecarl_forward` take three lanes: `patch=` a new plugin, `into=`
+an existing houseCARL patch, `in_place=` the file it names. `housecarl_remove` edits only what exists: `into=`
+or `in_place=`, no `patch=`. In place is consent-gated per plugin by `acknowledge=`.
 
 ## Where this does not apply
 
--- a/plugin/codex/housecarl/SKILL.md
+++ b/plugin/codex/housecarl/SKILL.md
@@ -66,7 +66,7 @@ full read-back — … NOT load-order truth; the patch wins nothing until enable
 ```
 
 Report the patch name and mod folder back — the name is auto-suffixed when taken. A refused call wrote
-nothing — except in place, where the file may already have changed.
+nothing; a failed in-place write says so in its message.
 
 ## Lanes and FormIDs
 
--- a/plugin/codex/housecarl/SKILL.md
+++ b/plugin/codex/housecarl/SKILL.md
@@ -1,85 +1,89 @@
 ---
 name: housecarl
-description: Work with Skyrim Special Edition load-order records and assets through the houseCARL MCP server — set or switch the MO2 instance, inspect active or disabled plugins and conflict trees, read records, query across plugins, author reviewable patch ESPs, create or remove records, edit leveled lists and composed structs, diff and resolve, edit NIF meshes and facegen, author dialogue, audit the SkyPatcher and SKSE runtime layers, compact or merge plugins, drive Papyrus compile/decompile and BSA archives, and look mods up on Nexus. Also the router for the bundled Skyrim helper skills (mutagen-reference, papyrus-reference, skypatcher-authoring, spid-authoring, kid-authoring, dialogue-authoring, facegen-diagnostics, open-animation-replacer, skse-plugin-authoring, bulk-record-jobs, npc-appearance-copy). Use whenever the user mentions houseCARL, an MO2 modlist, plugins, load order, conflicts, ESP patches, overrides, a record type (ARMO/WEAP/NPC_/LVLI/MGEF/…), leveled lists, keywords, facegen or dark faces, dialogue, SKSE plugins, Papyrus, BSA archives, Nexus, or a no-ESP runtime distribution — even when the task looks like a single edit, load this first to pick the right tool and read before you write.
+description: >-
+  Use whenever a task touches a Mod Organizer 2 modlist, load order, plugins,
+  conflicts, ESP patches or overrides, a record type (ARMO, WEAP, NPC_, LVLI,
+  MGEF), leveled lists, dark faces or facegen, dialogue, NIF meshes, SkyPatcher,
+  SPID or KID lines, SKSE plugins, Papyrus, or BSA archives. Routes Skyrim
+  Special Edition data-layer work through the houseCARL MCP server — a live MO2
+  instance read at the true load-order winner, written into reviewable patch
+  plugins — and names the specialist skill that owns each grammar. Not for
+  another game, for installing MO2, or for gameplay advice with no data-layer
+  step.
+compatibility: Requires the houseCARL MCP server and a configured Mod Organizer 2 instance.
 ---
 
 # houseCARL
 
-Use this skill for data-layer Skyrim Special Edition modding through the configured houseCARL MCP server. houseCARL reads a Mod Organizer 2 instance, resolves the true load-order winner, and writes changes into reviewable patch plugins; original source mods are never edited. Beyond the record read/write core it reaches the whole data layer — assets and NIF meshes, the SkyPatcher and SKSE runtime layers, dialogue, plugin operations, Papyrus compile/decompile, BSA archives, and keyless Nexus lookups — plus the 11 focused helper skills below for the specialist grammars and workflows.
+**In:** a live Mod Organizer 2 instance, read through the houseCARL MCP server. **The work:** read the
+record at its true load-order winner, check the schema, then write. **Out:** a patch plugin the user
+reviews and enables in MO2. This file does two things and defers the rest — it routes a job to the skill
+that owns its grammar, and it gives the read order every write depends on. What a tool takes is in that
+tool's own description.
 
-## Core workflow (read before you write)
+The user's instructions take precedence over guidelines provided in a skill.
 
-1. Confirm context when it matters:
-   - `housecarl_load_order_status` for profile/plugin status, or to check whether a mod or plugin is active.
-   - `housecarl_set_mo2_instance` when the user gives a new MO2 instance folder.
-2. Read before any record write:
-   - `mutagen-reference` to verify field names, writability, enum values, and composed-struct shapes.
-   - `housecarl_read_record` or `housecarl_batch_record_detail` to inspect the current winner. Add `conflict_tree=true` for contested records or when winner provenance matters.
-   - `housecarl_cross_plugin_query` to locate records or references across the load order (page big results with `offset=`).
-3. Pick the narrowest write tool:
-   - `housecarl_set_field` for a single scalar or simple-collection edit.
-   - `housecarl_bulk_apply` for several edits in one patch, dict merges, leveled-list entries, effects, or other composed structs.
-   - `housecarl_create_record` (or `housecarl_bulk_create` for many at once) for a new top-level record — it needs an EditorID.
-   - `housecarl_remove_record` only to drop a record or override from a houseCARL-owned patch — never from a source mod.
-4. Name the patch on the **first** write that creates one with `patch=<name>` (not `housecarl_remove_record`, whose `patch=` names one that already exists) — omit it and houseCARL names it `Patch`. Either way the name is auto-suffixed if it is already taken, so read the patch name back off the response. After that, accumulate related edits into it with `into=<patch filename>`.
-5. Prefer runtime, no-ESP INI systems when they fit the user's intent — `skypatcher-authoring`, `spid-authoring`, `kid-authoring` (see the helper skills below).
+## Which skill owns the job
 
-## The full tool surface
+| The user is asking about | Load | When |
+|---|---|---|
+| A dark, grey or black NPC face; a face wrong after compacting or merging | `housecarl:facegen-diagnostics` | before judging the fix |
+| Copying one NPC's face onto another, or cloning a standalone follower | `housecarl:npc-appearance-copy` | before the copy |
+| What fields a record type has, or a legal enum value | `housecarl:mutagen-reference` | before the write |
+| A no-ESP edit to a record's own fields, or to one NPC | `housecarl:skypatcher-authoring` | before the INI line |
+| Spells, perks, items, outfits or factions onto NPCs by group | `housecarl:spid-authoring` | before the `_DISTR.ini` line |
+| Keywords onto item records | `housecarl:kid-authoring` | before the `_KID.ini` line |
+| Adding, wiring or auditing dialogue topics and lines | `housecarl:dialogue-authoring` | before the DIAL/INFO write |
+| Gating animations by weapon, keyword, perk or race | `housecarl:open-animation-replacer` | before the condition |
+| A Papyrus or SKSE function signature, or a `.psc` edit | `housecarl:papyrus-reference` | before the script edit |
+| Writing or building a native SKSE plugin DLL in C++ | `housecarl:skse-plugin-authoring` | before the first C++ |
+| A catalogue, audit, conflict survey or link graph over many records | `housecarl:bulk-record-jobs` | before the first call |
 
-Beyond the core workflow, reach for the right group. Depth for the specialist areas lives in the helper skills (next section) — load the skill before composing in that area.
+MCP tools are written bare on both hosts (`housecarl_records`); a sibling is written `housecarl:<skill>`, the
+Claude Code invocation — on Codex it is the bare folder name (`facegen-diagnostics`), installed beside this one.
 
-**Read / query / resolve**
-- `housecarl_records` — the consolidated 2.0 read surface (SELECT × SOURCE × PROJECT in one call): record lists or scans, any plugin's version wherever it lives (active or on disk), form-scoped `project=` shapes. The 1.x read tools below keep working through the 2.0 build.
-- `housecarl_read_record`, `housecarl_batch_record_detail` — read one or many records at the true winner (`conflict_tree=true` for provenance).
-- `housecarl_read_plugin_file` — read a plugin directly, even one that is disabled or not active.
-- `housecarl_cross_plugin_query` — query and filter records or references across the whole order; page with `offset=`, count with `group_by=`.
-- `housecarl_resolve` — resolve a list of FormIDs to identity; `housecarl_diff_record` — diff two plugins' versions of a record.
-- `housecarl_effect_chain` — trace a magic effect to every spell / enchantment / potion / scroll / ingredient that carries it.
-- `housecarl_load_order_status` — enabled/disabled mods & plugins; `housecarl_check_errors` — dangling refs, missing masters, broken links.
-- `housecarl_check` — the merged derived-findings sweep: `findings=` picks the family (`errors` — dangling refs, missing masters, parse failures, the default; `scripts` — unbound VMAD script properties; `dialogue` — the dialogue graph over the topics and quests `seeds=` names) or a class inside one, in one call.
+## Read before you write
 
-**Runtime layers (what xEdit can't see)**
-- `housecarl_skypatcher_read` — a record's true state after the SkyPatcher INI layer replays; `housecarl_skypatcher_layer` — the INIs, apply order, conflicts.
-- `housecarl_skse` — the SKSE layer, one family per call via `findings=`: `inventory` (SKSE-plugin DLLs, configs, provider/metadata — the default), `pairing` (native Papyrus declarations vs the DLLs implementing them), `config` (config references vs the load order).
+1. **Confirm context when it matters.** `housecarl_load_order_status` says what is active;
+   `housecarl_set_mo2_instance` when the user names a different MO2 instance folder.
+2. **Read the winner and the schema.** `housecarl:mutagen-reference` for the field path and its legal
+   values — if it has no entry for a type, say so rather than guessing; `housecarl_records` for the record
+   as the order resolves it, `project={"form":"tree"}` for every provider when the winner is contested.
+3. **Write with one verb.** `housecarl_apply` edits fields (`ops=`), `housecarl_create` mints records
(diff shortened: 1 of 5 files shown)


Avick3110:
The regrade's D4.7 / D5.13 defect is corrected in adb3fa30ae6eced26415bb25678bcd493075187f. The "Lanes and FormIDs" section now reads:

> A FormID is `XXXXXX:Plugin.esp` — six hex digits, then the defining master's filename. The runtime form a log or the console prints is read-only: `housecarl_records` takes it, every write refuses it and names the `XXXXXX:Plugin.esp` form to use. SkyPatcher, SPID and KID write their own syntax.
>
> `housecarl_apply`, `housecarl_create` and `housecarl_forward` take three lanes: `patch=` a new plugin, `into=` an existing houseCARL patch, `in_place=` the file it names. `housecarl_remove` edits only what exists: `into=` or `in_place=`, no `patch=`. In place is consent-gated per plugin by `acknowledge=`.

Both claims are now true against the captured surface: `housecarl_remove` has no `patch=` (its lanes are `into=` and `in_place=`, because a removal edits an artifact that already exists), and the runtime form is accepted by `housecarl_records` and refused by the write door every write tool parses its FormIDs through, which hands back the `XXXXXX:Plugin.esp` form.

The body stays inside the rewrite document's budgets: 4,798 characters (cap 4,800) and 58 content lines (cap 70). Nothing outside that section changed — a grep for `patch=` and `runtime` in the body found no repeat of either claim.

Green from inside the worktree: `scripts/build-plugin.ps1` (leak-check clean, 12 skills / 227 markdown files), `dotnet build` 0 errors, 1894 tests passed / 0 failed, `ci-all` 128/128.

Avick3110:
D4.50 is corrected in 7ec16ca2.

The example's FormID was wrong: `013BA3:Skyrim.esm` is the NPC Hulda, so the one call the body modelled set a weapon field on an NPC and would have been refused at pre-flight. The Iron Sword is `012EB7:Skyrim.esm` — `type=Weapon`, `editorid=IronSword` in Skyrim.esm, name "Iron Sword" — verified against the live order with `housecarl_records` on the identity projection. The printed result was also the compact default rather than the full field-by-field dump `readback=true` returns; it now shows the read-back block (banner, record line, per-field lines), elided with `...` to stay inside the body budget.

The new pair:

```
housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": 12}], patch="SwordFix", readback=true)
```
```
wrote SwordFix.esp (new patch; 1284 bytes)
full read-back — the ENTIRE record(s) as written, re-read from the patch file (NOT load-order truth):
  Weapon 012EB7:Skyrim.esm  editorid=IronSword
    BasicStats.Damage = 12
    ...
```

The body is 4,792 characters and 60 content lines, inside the rewrite's 4,800 / 70 budget, and the trim came out of the example itself: the sentence that repeated "the written FILE, not load-order truth" is now carried by the read-back banner. Nothing else in the file changed.

Checks from the worktree: `scripts/build-plugin.ps1` green with the leak scan clean (12 skills, 227 markdown files), `dotnet build` clean, 1894 tests passed, `ci-all` 128/128.

plugin/codex/housecarl/SKILL.md
```diff
@@ -1,85 +1,91 @@
...
+```
+```
+wrote SwordFix.esp (new patch; 1284 bytes)
+full read-back — the ENTIRE record(s) as written, re-read from the patch file (NOT load-order truth):
+  Weapon 012EB7:Skyrim.esm  editorid=IronSword
+    BasicStats.Damage = 12
+    ...
+```
+
+Report the patch name back — it is auto-suffixed when taken. A refused call wrote nothing.
```
Avick3110: **"A refused call wrote nothing" is false on the in-place lane.**

It holds for the patch lane (`Apply` is all-or-nothing before anything touches disk), but not for `in_place=`. `WritePatchBuilder` rewrites the target file and *then* re-opens it to verify; when that re-open fails it returns a refusal after the bytes have already changed:

```
src/housecarl-core/WritePatchBuilder.cs:1065
  return PatchOutcome.Fail($"'{fileName}' was edited in place but could not be re-opened to verify: {ex.Message}");
```

(and the same shape at `:1587` for removal, `:1897` for forward). `LoadOrderService.PersistInPlaceConsent`'s own doc comment states the case outright: "a write that lands and then fails its verify mutates the file while recording no consent, so the next call re-prompts against an already-modified file."

Scenario: a Codex session runs `housecarl_apply(..., in_place="CoolWeapons.esp", acknowledge=true)`; the rewrite lands, the verify re-open fails (file locked by MO2/xEdit, transient IO), the tool answers `error: …`. Following this line, the session tells the user nothing was written — while their original, for which houseCARL keeps no backup or undo, has already been overwritten.

Suggest scoping the sentence to the lane it is true for, e.g. "A refused call on the patch lane wrote nothing; an in-place refusal may have already rewritten the file — re-read it before reporting."

plugin/codex/housecarl/SKILL.md
```diff
@@ -1,85 +1,91 @@
...
+
+## Lanes and FormIDs
+
+A FormID is `XXXXXX:Plugin.esp` — six hex digits, then the defining master's filename. The runtime form a log
+or the console prints is read-only: `housecarl_records` takes it, every write refuses it and names the
+`XXXXXX:Plugin.esp` form to use. SkyPatcher, SPID and KID write their own syntax.
+
+`housecarl_apply`, `housecarl_create` and `housecarl_forward` take three lanes: `patch=` a new plugin, `into=`
+an existing houseCARL patch, `in_place=` the file it names. `housecarl_remove` edits only what exists: `into=`
+or `in_place=`, no `patch=`. In place is consent-gated per plugin by `acknowledge=`.
```
Avick3110, line 79: **`acknowledge=` does not carry the deleted rule (FOLD-IN F39).**

The old body's last safety note was "Never edit a source mod in place unless the user has explicitly opted into the in-place lane." The PR deletes it on the grounds that the server's consent gate already carries it. It does not: `acknowledge` is a plain `bool` parameter *the model itself sets*, and the server's first-touch prompt ends

```
src/housecarl-mcp/LoadOrderService.cs:6062
  "Re-call the SAME edit with acknowledge=true to proceed."
```

with nothing anywhere in it saying to ask the user first. `PersistInPlaceConsent` then records the acknowledgement per plugin, so every *later* in-place write to that file is unprompted.

Scenario: the user says "bump the damage on the sword in CoolWeapons.esp". The model picks `in_place="CoolWeapons.esp"`, meets the prompt, re-calls with `acknowledge=true`, and overwrites the user's original with no backup — the user never opted into the lane, and no subsequent in-place write to that plugin will prompt again. The repo's own CLAUDE.md states the intended contract ("Editing an existing plugin in place is an opt-in with per-plugin consent"), and the PR's Q7 concedes it is unverified that a Codex session even receives the ServerInstructions that state "in-place is opt-in, consent-gated" — so on Codex this rule may now live in no shipped file at all.

One sentence restoring it ("use `in_place=` only when the user has asked for it") costs ~60 of the 4,800 body characters.

plugin/codex/housecarl/SKILL.md
```diff
@@ -1,85 +1,91 @@
...
+   (`records=`), `housecarl_remove` drops them (`formids=`), `housecarl_forward` carries another plugin's
+   record as an override. Every list is set-valued — one op is a set of one — there is no single/bulk pair.
+4. **Read the written record back**, and say what happened when it did not take.
+
+```
+housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": 12}], patch="SwordFix", readback=true)
+```
+```
+wrote SwordFix.esp (new patch; 1284 bytes)
+full read-back — the ENTIRE record(s) as written, re-read from the patch file (NOT load-order truth):
```
Avick3110: **The worked transcript is presented as literal tool output but is not what the tool prints.**

The real `housecarl_apply` response for this call emits three things this block drops:

```
src/housecarl-mcp/WriteSentences.cs:113-118
  wrote {file} (new patch; {bytes} bytes)
  mod folder: {modFolder}  — enable + sort it in MO2 to use the patch
  masters: Skyrim.esm
src/housecarl-mcp/WriteTools.cs:237
  "full read-back — the ENTIRE record(s) as written, re-read from the patch file on disk "
  "(the written file's content, NOT load-order truth; the patch wins nothing until enabled + sorted in MO2):"
```

The abbreviated banner cuts exactly the "wins nothing until enabled + sorted in MO2" half, and the omitted `mod folder:` line is the actionable instruction — the one `evals/evals.json` case 10 expects the model to relay ("the answer tells the user to enable the patch in MO2"). A worked example is the strongest instruction in a router this short; a session that learns the response shape from here learns a response that does not tell the user to enable anything.

Either quote the real banner and keep the `mod folder:` line, or mark the block as elided (a leading `…` / `[trimmed]`) so it does not read as a transcript.

src/housecarl-generator/CodexUmbrellaCoverageProbe.cs
```diff
@@ -170,6 +180,13 @@ static List<string> MissingRefs(string umbrella, IEnumerable<string> required, I
         => required.Where(r => !allow.Contains(r) && !ReferencedAtBoundary(umbrella, r))
                    .OrderBy(r => r, StringComparer.Ordinal).ToList();
 
+    /// <summary>Every <c>housecarl_*</c> identifier the router writes that is NOT a reflected tool name — the
+    /// direction the old one-way check could not see. The sibling-skill form the router also writes is
+    /// <c>housecarl:&lt;skill&gt;</c>, which the underscore in this pattern excludes by construction.</summary>
+    static List<string> DeadToolNames(string umbrella, ISet<string> live)
+        => Regex.Matches(umbrella, "housecarl_[a-z0-9_]+").Select(m => m.Value).Distinct(StringComparer.Ordinal)
```
Avick3110: **INV1 has no escape hatch, and the pattern matches more than tool names.**

`DeadToolNames` reports *every* `housecarl_[a-z0-9_]+` token in the file that is not a reflected tool, and — unlike `MissingRefs` — it never consults `Allow`. The `Allow` comment was narrowed to "SKILL routing (INV2)" in this same diff, so INV1 is now unsuppressable by construction.

Scenario: the router gains one sentence about the bulk-job manifest, whose first key is `housecarl_artifact` — a JSON field name, not a tool, already written that way in a shipped sibling:

```
.claude/skills/bulk-record-jobs/SKILL.md:104
  head -1 C:/work/npc-winners.jsonl   →  {"housecarl_artifact":1,"tool":"housecarl_records",…
```

CI then goes red with `the Codex umbrella names a tool that does not exist: housecarl_artifact — fix … against the published surface`, and the only ways out are deleting correct prose or editing this guard. The same applies to any future `housecarl_*` filename, config key or mod-folder name.

Passing `Allow` through here (or restricting the match to backticked tokens) keeps the teeth and gives a recorded, one-line-reason exit the way INV2 has.

plugin/codex/housecarl/evals/evals.json
```diff
@@ -0,0 +1,168 @@
...
+  "baseline": "None. This skill has never carried an eval set, and the 2026-09 audit ran neither arm for it (audit-2026-09/SUMMARY.md:43, both cells 'not run') — the only row in the batch with no with-skill and no control log. There is nothing to carry forward; this file is the first measurement.",
+  "method": {
+    "arms": [
+      "with-skill — the packaged umbrella in the menu, the houseCARL MCP server attached and its instructions injected, as any real session has it",
+      "without-skill — the same query, the same menu minus this skill, the server still attached; a case that passes in both arms is not differentiating, and that is a finding rather than a pass"
+    ],
+    "runs_per_query": 3,
+    "trigger_rate_threshold": 0.5,
+    "split": "60/40 train / held-out; best iteration chosen by held-out score, at most 5 iterations. A fix generalises the category the failing query belongs to; it never pastes that query's own nouns into the description.",
```
Avick3110: **The held-out split has zero positive cases, which makes "best iteration chosen by held-out score" select for under-triggering.**

Ids 11–16 are the entire `held_out` set and all six are negatives (Fallout 4, installing MO2, Requiem perk build, xEdit script, Vortex, "what armour mod next"). Every positive — ids 1–10 — is in `train`.

Scenario: iteration 2 narrows the description and the router stops firing on half the routing table. Held-out score goes *up* (a skill that never triggers is perfect on an all-negative set), so the method as written picks that iteration. The failure this eval set exists to catch is the one it cannot see.

The reference set in the same batch splits both ways — `.claude/skills/bulk-record-jobs/evals/evals.json` is 5 train-positive / 6 train-negative / 4 held-out-positive / 4 held-out-negative. Two related notes on the same line: 12/6 is 67/33, not the stated 60/40; and id 17, labelled in the file itself "the hardest negative in the set", sits in `train` where iteration can overfit to it.

plugin/codex/housecarl/SKILL.md
```diff
@@ -1,85 +1,91 @@
...
+description: >-
+  Use whenever a task touches a Mod Organizer 2 modlist, load order, plugins,
```
Avick3110, line 4: **The trigger list drops Nexus entirely, and nothing owns it now.**

The old description carried "look mods up on Nexus"; the new one carries no Nexus, update-check, BSA-lookup or mod-identification term, and no sibling skill owns that grammar — the routing table's eleven rows are all record/asset/script jobs. With INV1 reversed in this same PR, no CI arm notices a capability the router no longer triggers on.

Scenario: on Codex a user asks "which of my mods have updates on Nexus?". The umbrella does not fire (nothing in the 611 characters matches), no sibling fires, and `housecarl_update_status` / `housecarl_nexus_*` are reachable only if the ServerInstructions are injected — the single dependency the PR's own Q7 flags as taken on faith.

Not necessarily a reason to widen the description (the 611 characters are load-bearing for D2.21 / D3.5), but the gap deserves the same explicit record `method.positives_cover` gives the three uncovered routing destinations, rather than passing silently.

src/housecarl-generator/CodexUmbrellaCoverageProbe.cs
```diff
@@ -81,19 +87,23 @@ from longName in required
             Check($"GUARD-SELF the boundary matcher tells apart every colliding name pair ({collisions.Count} pairs on this surface)",
                 falsePasses.Count == 0, falsePasses);
 
-            // INV1 — every tool referenced.
-            var missTools = MissingRefs(umbrella, tools, Allow);
-            Check("INV1-GREEN every MCP tool is referenced in the umbrella router", missTools.Count == 0,
-                missTools.Select(t => $"tool not routed by the Codex umbrella: {t} — add it to {UmbrellaPath.Replace('\\', '/')} (or Allow with a reason)").ToList());
+            // INV1 — every name the router writes resolves to a live tool.
+            var dead = DeadToolNames(umbrella, tools);
```
Avick3110: **The 1,024-character ceiling this file "used to breach" is still unguarded.**

The changelog line added in this PR says the description "is 611 characters against the 1,024 ceiling it used to breach". Nothing in `ci-all` measures it — `grep -rn "1024" src/housecarl-generator scripts` finds no description-length arm, and `PluginValidateProbe` checks only that `name`/`description` parse and are non-empty.

So this one file has now silently shipped two different defects that CI could not see: seventeen dead tool names (fixed here by flipping INV1) and a description over the host ceiling (fixed here by hand, guarded by nothing). The second will recur the same way the first did — the next edit that adds a clause has no red arm to stop it.

This guard already reads the umbrella and already owns "what is true of this file", so a two-line arm asserting the parsed `description` length is under the ceiling would land naturally beside INV1/INV2 — and it applies to the eleven sibling SKILL.md frontmatters too.

plugin/codex/housecarl/SKILL.md
Avick3110: Folded in 2fe31469. The sentence now reads: "A refused call wrote nothing — except in place, where the file may already have changed."

You are right about the mechanism — `WritePatchBuilder.cs:1065` returns the refusal after the rewrite has landed (same shape at :1587 and :1897), and there is no backup, so the old sentence told a Codex session to report "nothing was written" about a file that had already been overwritten. Scoped rather than expanded: the body is at 4,799 characters against the document's 4,800 ceiling, so the wording had to buy itself out of prose elsewhere.

plugin/codex/housecarl/SKILL.md
Avick3110: Refuted — the rewrite document settled this the other way, and it settled it on this exact sentence.

FOLD-IN F39 (REWRITE.md:273) names `SKILL.md:86`, "Never edit a source mod in place unless the user has explicitly opted into the in-place lane", and rules it **deleted**: "the server enforces it per-plugin; `housecarl_apply.acknowledge` is the consent gate and the refusal is the mechanism ... The body keeps only the pointer: the lane exists and the server gates it." §5's section budget says the same. The body writes exactly that pointer, so restoring the sentence is restoring the deleted line rather than adding a new one.

The observation under it is a different, real thing, and it is not this file's to fix. If `LoadOrderService.cs:6062` should say "ask the user before re-calling" instead of only "Re-call the SAME edit with acknowledge=true to proceed", that is one refusal sentence the server prints on every host, not a rule restated in one of twelve skill bodies — and restating it here would leave the Claude Code lane, where the same gate is reached from eleven other skills, still uncovered. Worth an issue against the prompt text; Aaron's call, and if he wants the line back regardless it is one sentence.

Q7 stands as written: if Codex sessions do not receive the ServerInstructions, this rule is one of several that live in no shipped file there, which is an argument about the injection, not about this paragraph.

plugin/codex/housecarl/SKILL.md
Avick3110: Folded in 2fe31469, both halves.

The block now carries the `mod folder:` line verbatim as `WriteSentences.NewOrExtendedArtifact` renders it — "mod folder: SwordFix  — enable + sort it in MO2 to use the patch" — so the actionable instruction eval case 10 expects the model to relay is in the worked example rather than only in the opening sentence. The read-back banner is now explicitly elided (`full read-back — … NOT load-order truth; the patch wins nothing until enabled + sorted in MO2:`), which keeps the half you flagged as cut and stops the line reading as a verbatim transcript.

The full banner quoted whole is ~200 characters and would have pushed the body past the document's 4,800; the ellipsis plus the mod-folder line lands at 4,799 with the trims taken out of prose, not out of a section.

src/housecarl-generator/CodexUmbrellaCoverageProbe.cs
Avick3110: Folded in e8a3c91e. `DeadToolNames` now takes the allow-list and filters on it, so INV1 has the same recorded exit INV2 has, and `Allow`'s comment covers both directions: a skill slug there is a skill deliberately not routed, a `housecarl_*` identifier there is a name deliberately not a tool (your `housecarl_artifact` manifest key, a filename, a mod folder), each with a one-line reason. The set is still empty.

The pattern is left matching every `housecarl_*` identifier rather than only backticked ones — restricting it would drop the teeth on a dead name written in prose, which is how the seventeen were written in the first place. New arm `ALLOW-1` feeds a synthetic "the manifest's first key is `housecarl_artifact`" through the real checker and asserts it is not reported dead, beside the existing red arm that a retired name still is. Guard runs 14 arms, all green.

plugin/codex/housecarl/evals/evals.json
Avick3110: Folded in a629dbfe, all three points. The split is now 11 train / 7 held out, positives and negatives on both sides: train 1, 2, 3, 5, 7, 9 positive and 11, 12, 14, 15, 18 negative; held out 4, 6, 8, 10 positive and 13, 16, 17 negative. That is 61/39, which the file states as its own counts rather than as a round claim the rows contradict. Id 17 moved to held_out for the reason you give — the hardest negative is the one iteration should not be able to fit to.

`method.split` now records why, since the trap is not obvious from the row list: iteration is scored on the held-out set, so an all-negative held-out set scores a description narrowed until the router never fires as perfect, and the failure the set exists to catch is the one it cannot see. Held-out positives cover four distinct routing destinations plus the write loop, so a narrowing iteration loses points there.

plugin/codex/housecarl/SKILL.md
Avick3110: Folded as the record you ask for, in a629dbfe: `method.trigger_gaps` in `evals/evals.json` now names Nexus as the one capability no description in the bundle carries, says the old description carried "look mods up on Nexus" and the rewrite drops it, says no sibling owns that grammar, and says the reach is left to the injected instructions — the dependency Q7 already flags. The description is left at 611 characters for the reason you name yourself: widening it is what D2.21 / D3.5 / D4.59 were failing on, and both measurements feed the regrade gate.

Whether the gap should be closed is the same question as item 27 — a router that must carry a Nexus clause to be reachable is a router earning its listing on a capability the server already advertises — so the file records it for the run rather than settling it here.

src/housecarl-generator/CodexUmbrellaCoverageProbe.cs
Avick3110: Folded in cfc22004, though in `PluginValidateProbe` rather than here — that guard already parses all twelve shipped frontmatters (both trees), so the arm lands on the eleven siblings too, which is the half of your point this file could not have covered.

What is there now: `MaxDescription = 1024` named as the loader's `MAX_DESCRIPTION_LENGTH`, a violation line on the parsed description ("description is N characters, past the 1024-character ceiling — trim it"), and a RED arm feeding a synthetic description one character past it. Measured on the parsed string, so the folded YAML block a long description is written as counts the way the loader counts it. All twelve pass today at 441–711 characters; the umbrella is 611.

The guard runs 21 arms green, and the second defect this file shipped now has an arm that goes red the next time a clause is added.

Avick3110:
Fold pass over the seven review comments: six folded, one refuted, each answered in its own thread. The branch is rebased onto the current base (the changelog entry now sits under the facegen and DAR lines) and re-verified.

**Folded.** The worked call's transcript now carries the real `mod folder:` line and marks the read-back banner as elided, keeping the "wins nothing until enabled + sorted in MO2" half it had cut; the closing sentence no longer claims a refused call never wrote, which was false on the in-place lane. `evals.json` moves positives into the held-out split (train 6+5, held out 4+3, id 17 with them) and records the Nexus trigger gap. The umbrella guard's dead-name check now consults the allow-list, with an arm proving it suppresses. The description ceiling is guarded for real: `PluginValidateProbe` measures every shipped frontmatter's parsed description against `MAX_DESCRIPTION_LENGTH`, so it covers the eleven siblings too, with a red arm at ceiling+1.

**Refuted.** Restoring "never edit a source mod in place unless the user opted in" — FOLD-IN F39 deletes that exact sentence and rules the pointer is what the body keeps. The concern under it is real and belongs to the server's own prompt text, which is one file for every host; noted in the thread for Aaron.

**Numbers the PR body states that this pass changed.** The body is **4,799** characters, not 4,773 — the mod-folder line and the scoped refusal sentence paid for themselves out of prose in five places, and it is still under the document's 4,800. The umbrella guard runs **14** arms, not 13, and `plugin-validate-guard` runs 21. Description unchanged at 611 characters; the routing table, the lanes and the exclusion clause are untouched.

**Checks after the rebase.** `scripts/build-plugin.ps1` — leak-check clean, 12 skills / 227 markdown files across both trees. `dotnet build -c Release` clean; tests **1894 passed, 0 failed** (`tier!=bridge`); generator `ci-all` **128/128**.

plugin/codex/housecarl/SKILL.md
```diff
@@ -1,85 +1,92 @@
...
+2. **Read the winner and the schema.** `housecarl:mutagen-reference` for the field path and its legal
+   values — no entry for a type means say so, never guess; `housecarl_records` for the record
+   as the order resolves it, `project={"form":"tree"}` for every provider when contested.
+3. **Write with one verb.** `housecarl_apply` edits fields (`ops=`), `housecarl_create` mints records
+   (`records=`), `housecarl_remove` drops them (`formids=`), `housecarl_forward` carries another plugin's
+   record as an override. Every list is set-valued — one op is a set of one; no single/bulk pair.
+4. **Read the written record back**, and say what happened if it did not take.
+
+```
+housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": 12}], patch="SwordFix", readback=true)
```
Avick3110: **The worked call is refused by the server.** `"value": 12` is a JSON *number*, but `ApplyOp.Value` is `string?` (`ApplyTools.cs:275`) and `ops=` deserializes through `ListParams.Strict`, which sets no `JsonNumberHandling`. STJ throws `The JSON value could not be converted to System.String. Path: $[0].value` — I ran the same options and shape to confirm — so `ReadOps` returns `ops could not be parsed … Expected a JSON ARRAY of {formid, field_path, …} elements` and nothing is written.

Every other houseCARL skill writes it quoted (`"value": "1"`, `"value": "ForceGreet"`). Suggested fix below.

Worth noting for the regrade: INV1 as rewritten matches tool **names** only, so `ci-all` is green on a body whose single executable example is a call the server refuses — the same shape of miss as the seventeen dead names, one axis over. The PR body's claim that "ten parameters the body writes is live" is hand-checked, not guarded.

```suggestion
housecarl_apply(ops=[{"formid": "012EB7:Skyrim.esm", "field_path": "BasicStats.Damage", "value": "12"}], patch="SwordFix", readback=true)
```

plugin/codex/housecarl/SKILL.md
```diff
@@ -1,85 +1,92 @@
...
+wrote SwordFix.esp (new patch; 1284 bytes)
+mod folder: SwordFix  — enable + sort it in MO2 to use the patch
+full read-back — … NOT load-order truth; the patch wins nothing until enabled + sorted in MO2:
+  Weapon 012EB7:Skyrim.esm  editorid=IronSword
+    BasicStats.Damage = 12
+    ...
+```
+
+Report the patch name and mod folder back — the name is auto-suffixed when taken. A refused call wrote
+nothing — except in place, where the file may already have changed.
```
Avick3110: **"except in place, where the file may already have changed" contradicts the in-place contract.**

Every final-swap write funnels through `AtomicFile.Commit`, and `AtomicCommitProbe` arms C and C2 assert the prior target survives byte-for-byte on a pre-swap failure *and* on a mid-swap failure with the target held by MO2/xEdit/the game. `housecarl_remove`'s own `acknowledge=` doc says it outright: "a call that is refused records nothing, so it may be needed again."

So the concrete case: a user calls `housecarl_apply(..., in_place="CoolMod.esp")` without `acknowledge=`, gets the consent prompt — a refusal — and the session, following this sentence, tells them their original may already be damaged and to go check or restore it. That is exactly the wrong answer, and the clause is new here: neither the previous body nor any sibling skill makes it.

If what it means to defend is the post-write absence/read-back verify failing *after* a successful commit, that is a failed call rather than a refused one, and the sentence should say so.

src/housecarl-generator/CodexUmbrellaCoverageProbe.cs
```diff
@@ -170,6 +188,15 @@ static List<string> MissingRefs(string umbrella, IEnumerable<string> required, I
         => required.Where(r => !allow.Contains(r) && !ReferencedAtBoundary(umbrella, r))
                    .OrderBy(r => r, StringComparer.Ordinal).ToList();
 
+    /// <summary>Every <c>housecarl_*</c> identifier the router writes that is NOT a reflected tool name and NOT
+    /// allow-listed — the direction the old one-way check could not see. The sibling-skill form the router also
+    /// writes is <c>housecarl:&lt;skill&gt;</c>, which the underscore in this pattern excludes by construction.
+    /// The pattern matches any such identifier, tool-shaped or not, so the allow-list is the recorded exit for a
+    /// <c>housecarl_*</c> name that is deliberately something else.</summary>
+    static List<string> DeadToolNames(string umbrella, ISet<string> live, ISet<string> allow)
+        => Regex.Matches(umbrella, "housecarl_[a-z0-9_]+").Select(m => m.Value).Distinct(StringComparer.Ordinal)
```
Avick3110: **The pattern misfires at both ends.**

*False positive.* `_` is inside the trailing class, so a family reference written the way `evals/evals.json` already writes one — "reaches `housecarl_update_status` and the `housecarl_nexus_*` tools" — matches as `housecarl_nexus_`. That string is in neither the reflected set nor `Allow`, so INV1-GREEN goes red with "the Codex umbrella names a tool that does not exist: housecarl_nexus_", and the only exits are deleting correct prose or allow-listing a string nobody wrote as a name — the two outcomes the `Allow` comment says it exists to avoid.

*False negative.* The class is lowercase-only and the literal prefix must be followed by at least one member of it, so `housecarl_read_Record` yields no match at all and a dead name passes INV1 silently.

Anchoring the tail and dropping case-sensitivity closes both: `Regex.Matches(umbrella, "housecarl_[a-z0-9]+(?:_[a-z0-9]+)*", RegexOptions.IgnoreCase)`.

plugin/codex/housecarl/SKILL.md
Avick3110: Folded in f9dcd963.

Confirmed before changing: `ApplyOp.Value` is `string?` (`ApplyTools.cs:275`), and `ListParams.Strict` (`ListParams.cs:104`) sets `PropertyNameCaseInsensitive`, `ReadCommentHandling`, `AllowTrailingCommas` and `UnmappedMemberHandling` only — no `JsonNumberHandling` — so the number does not bind and the one executable call in the body is refused. It is `"value": "12"` now, matching every sibling skill.

The INV1 point is noted: the probe matches tool names, so an example the server refuses for any other reason passes `ci-all`. Not adding a second guard for it — the repo's rule is fix the mistake, not add a sweep to catch the class.

plugin/codex/housecarl/SKILL.md
Avick3110: Folded in eefbee1f.

Verified: `AtomicFile.Commit` is the one final-swap primitive (`File.Replace`, falling back to `File.Move` only when the target is absent), it throws rather than half-writing, and its own contract says the prior target is byte-intact on any throw. `AtomicCommitProbe` arms C and C2 assert exactly that for a pre-swap failure and for a mid-swap failure with the target held by an external holder. So a refused call — the `acknowledge=` consent prompt included — cannot have written.

The clause is gone. The sentence now reads:

> A refused call wrote nothing; a failed in-place write says so in its message.

which is the post-commit verify case you named — `WritePatchBuilder.cs:1065`, "'<file>' was edited in place but could not be re-opened to verify" — a FAILED call, not a refused one.

Budget: the body is 4,790 characters and 61 content lines (77 with blanks), down from 4,799 / 61, so both fixes fit under 4,800 / 70 with room.

src/housecarl-generator/CodexUmbrellaCoverageProbe.cs
Avick3110: Pattern taken as given, in c5c906ef. All 14 arms green, RED arms still red. Two corrections from running both patterns against the real name set, though — the fix lands only half of what it is credited with.

**False negative — real, but not on that example.** `housecarl_read_Record` did match the old pattern: `[a-z0-9_]+` is greedy up to the capital, so it yielded `housecarl_read_`, dead, and INV1 went red under a mangled name rather than passing. The genuine miss is a capital in the FIRST segment — `housecarl_Apply` matched nothing at all under the old pattern and is caught under the new one. `IgnoreCase` closes that, and the segment anchor is what makes the reported string the whole name instead of a truncation.

**False positive — not closed.** Under the new pattern `housecarl_nexus_*` extracts `housecarl_nexus`, and there is no such tool: the five are `housecarl_nexus_search`, `_mod`, `_identify`, `_graphql`, `_check_updates` (`ToolNames.cs:33-37`). So that prose still turns INV1 red, only naming `housecarl_nexus` instead of `housecarl_nexus_` — same two exits. The guard is green today because `DeadToolNames` reads `plugin/codex/housecarl/SKILL.md` only, and the family form lives in `evals/evals.json`, which it never scans; the umbrella body itself writes no family reference. Closing it for real needs the matcher to recognise the `_*` family form, which is more than a pattern swap — not added, per "add nothing else". Your call whether it wants a follow-up.

PR #673, opened 2026-09-08
Delete bulk-record-jobs and npc-appearance-copy

Avick3110:
Deletes two shipped skills whole, per the 2026-09-08 gate dispositions: `bulk-record-jobs` and `npc-appearance-copy`, with their evals and reference files. Both TIED their controls in the behavioural gate. `dev/skill-validation/gate-2026-09/bulk-record-jobs/VERDICT.md` records the same 66,856 rows against the same epoch from both arms, the control ahead on completeness (it alone carried the template column), and the skill's bounding half already answered by the tool schema the control read at call 1. `dev/skill-validation/gate-2026-09/npc-appearance-copy/VERDICT.md` records two artifacts I read back with the tools rather than from either log: the same `000800` record field for field, the same placed mesh with `tex[6]` repointed at its own plugin's `.dds`, the same error and the same recovery — and the skill's own headline contribution falsified by the control, which derived the repoint unaided from `housecarl_place`'s and `housecarl_nif_set`'s descriptions.

They come out of the packager's ship list (`scripts/build-plugin.ps1`, now 9 skills), out of the Codex umbrella router (`plugin/codex/housecarl/SKILL.md`, two routing rows), out of both READMEs, and out of every live pointer the remaining skills had at them. `facegen-diagnostics` was the only sibling that ceded to either: its description, its §11 table, its §8 whole-order sweep and one row in `references/facegen-causes-and-fixes.md` now give the plain instruction instead of the skill name — the whole-order sweep is one `housecarl_records` call spilled with `to_file=`, and a copy is `housecarl_copy` plus `housecarl_place`. Nothing else in that skill is touched. Four eval cases that expected one of the deleted skills to load (`dialogue-authoring` T19, `spid-authoring` 16, `facegen-diagnostics` `st-dark-after-installing-mod` and `snt-copy-face-boundary`, plus the umbrella's train case 7) now expect the tools directly or NONE; the menu counts on those sets move 11 → 9, and `spid-authoring`'s deleted-skills note names all five. Neither notices file mentioned either skill, so both are unchanged. Three unreleased changelog entries about the two skills are removed with them, and the Codex leftover list in the setup-prune entry goes from three folders to five.

Two folds ride this PR, both onto the tool that does the work.

`housecarl_records`, `fields_source=` — the column the bulk skill never mentioned and the control found unaided:

> Neither pole reaches TEMPLATE inheritance: an NPC whose Configuration.TemplateFlags include Stats takes its level, and the rest of that category, from the record its Template points at, so the level rendered on its own row is not the level the game uses — read TemplateFlags and Template beside any NPC level or stat column and say which rows inherit.

`housecarl_records`, `to_file=` — the self-accounting header the skill prescribed, checked against what `ResultArtifact.WriteTo` actually writes:

> The manifest is the file's own accounting and is what you read back, not the chat summary: it carries the query echo, identity, row_schema, sort, row_count, total, type_counts where the rows carried types, epoch, created, and notes when the result owes a clause — the file holds the COMPLETE result only when row_count equals total, and windows joined across differing epochs are not one result.

`housecarl_copy`, `new_editorid=` — the fact the control reasoned out at call 12 and the skill stated only in passing:

> The clone is a whole-record duplicate, so an NPC's INLINE face values — tint layers, face morphs and parts, TextureLighting, hair colour, weight — carry by construction: no separate write of them is owed, and seeding them would only re-set what is already there.

`housecarl_copy`, tool description — the two things the with-skill run bought, on the sentence that already hands the files to `housecarl_place`:

> This tool copies RECORDS. The FILES that go with them — an NPC's FaceGen mesh and tint, say — are placed with housecarl_place; once they are placed, housecarl_check over the patch sweeps it off-order, before it is enabled, and the textures baked INSIDE a placed mesh are not copied by either tool and still resolve from the source mod, which stays a file dependency.

That last one is the tool's own description rather than a parameter's, because `housecarl_copy` has no parameter about placing files — the sentence it extends is the one that names `housecarl_place`.

Counts: 8 files deleted (1,143 lines), 15 other files changed, 9 skills ship. Build, tests and `ci-all` pass (1,894 tests, 128/128 probes), and a full `scripts/build-plugin.ps1` run packages the tree with `skills: 9` and a clean leak check.


src/housecarl-mcp/CopyTools.cs
```diff
@@ -29,7 +29,10 @@ public static class CopyTools
          "OUTPUT — patch= names a NEW plugin (folder-per-patch); into= extends an existing houseCARL patch " +
          "instead.\n\n" +
          "This tool copies RECORDS. The FILES that go with them — an NPC's FaceGen mesh and tint, say — are " +
-         "placed with " + ToolNames.Place + ". Originals are never touched, and a refusal writes nothing.")]
+         "placed with " + ToolNames.Place + "; once they are placed, " + ToolNames.Check + " over the patch " +
+         "sweeps it off-order, before it is enabled, and the textures baked INSIDE a placed mesh are not copied " +
```
Avick3110: This sentence states the baked-texture dependency but names no way out of it, and it is the one fact of the deleted skill that had an action attached. `housecarl_nif_set set_path` slot 6 is exactly the repoint — the skill's fourth call, and the PR body says the control derived it unaided from `housecarl_nif_set`'s description. `housecarl_place`'s description, which this sentence hands off to, never mentions baked textures or `nif_set` either, so nothing on the surface now closes the loop.

Concrete scenario: a caller clones an NPC into a standalone follower (the stated purpose of the tool — "the result no longer masters the plugin you copied from"), places the FaceGen pair, runs `housecarl_check` off-order as instructed, gets a clean sweep, and ships a mod that renders correctly only while the source mod stays installed. The description told them the dependency exists and implied nothing more can be done about it.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -126,7 +126,7 @@ public static string Records(
...
+        [Description("The pole field VALUES display from, when it differs from the matching pole: \"winner\" shows the live winner's values on a plugins=-scoped scan (the old winner_fields=true). Display only — where_source= governs matching. Under a plugins= scope the default renders the SCOPED plugin's OWN values — the defining plugin's era, not the number the game uses after later overrides — so a deliverable claiming live stats passes \"winner\" here. Neither pole reaches TEMPLATE inheritance: an NPC whose Configuration.TemplateFlags include Stats takes its level, and the rest of that category, from the record its Template points at, so the level rendered on its own row is not the level the game uses — read TemplateFlags and Template beside any NPC level or stat column and say which rows inherit.")]
```
Avick3110: Two problems with where this fold landed.

**It is on a parameter most NPC reads never pass.** Template inheritance falsifies any NPC level or stat column, including a plain `formids=` + `project="fields"` read with no `plugins=` scope and no `fields_source=` at all. `fields_source=` is about winner-vs-scoped *display*; a caller who never needs that distinction never sees the caveat, which is precisely the caller who renders an inherited level as if it were the record's own. The caveat belongs where every NPC stat read passes.

**It prescribes a manual method the server already implements.** "read TemplateFlags and Template beside any NPC level or stat column and say which rows inherit" is hand-work; `walk={"follow":"Template"}` with `project="chain"` already renders `Traits: INHERITED from <provider>` / `Stats: local data ACTIVE` — see `RecordsComparisonFormTests.TemplateFlagsInterpreter_InheritedCategoriesNameTheirProviderAndClearFlagsReadLocalActive`. The new text neither names that form nor mentions the interpreter, so a caller following it reimplements by hand what one call already answers with provenance.

.claude/skills/spid-authoring/evals/evals.json
```diff
@@ -188,7 +188,7 @@
     {
       "id": 16,
       "prompt": "which plugin is winning the conflict on the steel plate armor record in my lorerim order",
-      "expected_output": "NONE, or housecarl:bulk-record-jobs — this tree ships no conflict-resolution skill",
+      "expected_output": "NONE — this tree ships no conflict-resolution skill",
```
Avick3110, line 191: Stale count left behind in this same file (**line 183**, case 15, just above this hunk):

> "NONE — this tree ships no crash-diagnosis skill, and none of the **11** covers the question"

Line 5 of this file was updated to "the 9 skills this tree actually ships", and case 16 right here was rewritten, but case 15 still asserts a menu of 11. A run scored against this file now has one case whose expected output describes a menu that does not exist, which is the same drift the `method` rewrite was there to close.

```suggestion
      "expected_output": "NONE — this tree ships no conflict-resolution skill",
```
(no change needed on this line — the fix is `11` → `9` on line 183.)

plugin/codex/housecarl/evals/evals.json
```diff
@@ -22,8 +22,8 @@
...
+    "positives_cover": "one per routing destination the DESCRIPTION claims. Two of the nine destinations have no positive — open-animation-replacer and papyrus-reference are reachable from the routing table but the description carries no animation or signature-lookup term. That is deliberate: the description was not widened to chase them (its 611 characters are the measured fix for D2.21 / D3.5 / D4.59), and on both hosts those three siblings carry their own trigger descriptions beside this one. Recorded rather than papered over — verify row 9.11.",
```
Avick3110: `those three siblings` is a leftover from the pre-edit list. The sentence now names **two** destinations with no positive (open-animation-replacer and papyrus-reference) after npc-appearance-copy was dropped from it, so "those three siblings carry their own trigger descriptions" no longer resolves to anything — a reader counts two names and three siblings.

```suggestion
    "positives_cover": "one per routing destination the DESCRIPTION claims. Two of the nine destinations have no positive — open-animation-replacer and papyrus-reference are reachable from the routing table but the description carries no animation or signature-lookup term. That is deliberate: the description was not widened to chase them (its 611 characters are the measured fix for D2.21 / D3.5 / D4.59), and on both hosts those two siblings carry their own trigger descriptions beside this one. Recorded rather than papered over — verify row 9.11.",
```

.claude/skills/spid-authoring/evals/evals.json
Avick3110: Folded in 5aefa80. Line 183 now reads "none of the 9 covers the question", matching the menu on line 5.

plugin/codex/housecarl/evals/evals.json
Avick3110: Folded in 5aefa80. "those three siblings" is now "those two siblings", matching the two names the sentence lists.

src/housecarl-mcp/CopyTools.cs
Avick3110: Folded in b81d635. The sentence now carries the remedy: the baked textures "stay a file dependency until you repoint them with housecarl_nif_set set_path (texture_slot= the shape's slot — a head mesh's tint is slot 6 — and path= the copy's own texture)". Slot 6 and the two addressing forms confirmed against NifTools.cs (the texture_slot= parameter description, and BuildOp's set_path case). The whole housecarl_copy description measures 1,498 characters, under the 2,048 ceiling. housecarl_place's description is untouched.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Half refuted, half folded.

**Refuted — the placement.** `fields_source=` is where Aaron put it. DECISIONS.md, 2026-09-08 (late night): "DELETE bulk-record-jobs (fold the template-inherited-stats sentence into `fields_source` and the self-accounting header shape into `to_file`)". That is a ruling on the destination, not an accident of drafting, so moving the caveat to a parameter every NPC stat read passes is Aaron's call to reverse, not mine. Recorded here rather than acted on.

**Folded — the hand-work.** Confirmed the claim: `RecordsTools.SeedTail` renders the per-category report (`Traits: INHERITED from <provider> (<editorid>)` / `Stats: local data ACTIVE`) off `WalkSeedResult.TemplateReport`, and `walk.follow` takes `"Template"` for exactly this chain (its own description names it; the `chain` refusal at RecordsTools.cs:348 suggests `walk={"follow": "Template"}` over NPC seeds). So the sentence no longer prescribes reading TemplateFlags and Template by hand — b78a825 ends it with: "walk={\"follow\": \"Template\"} under project.form='chain' reports, per category, which rows inherit and from which record, so read the inheritance there rather than by hand."

Build clean, 1894 tests pass, ci-all 128/128.

src/housecarl-mcp/CopyTools.cs
```diff
@@ -42,7 +47,7 @@ public static string Copy(
...
-        [Description("DESTINATION: mint a CLONE of the source record with this EditorID. Every link on the clone still pointing into the source is STRIPPED and reported by name — including when clearing one takes a WHOLE property with it (a script adapter, say), which the report says out loud. A link the record model REQUIRES cannot be stripped, so that refuses loud rather than writing an invented null or silently mastering the source. Pass this OR target.")]
+        [Description("DESTINATION: mint a CLONE of the source record with this EditorID. Every link on the clone still pointing into the source is STRIPPED and reported by name — including when clearing one takes a WHOLE property with it (a script adapter, say), which the report says out loud. A link the record model REQUIRES cannot be stripped, so that refuses loud rather than writing an invented null or silently mastering the source. The clone is a whole-record duplicate, so an NPC's INLINE face values — tint layers, face morphs and parts, TextureLighting, hair colour, weight — carry by construction: no separate write of them is owed, and seeding them would only re-set what is already there. Pass this OR target.")]
```
Avick3110, line 50: **The inline-face-values fold-in only covers the clone half.**

`target=` says "the seed fields are set on it pointing at the internalized copies" — only the link fields named in `seed_paths=` get written. Tint layers, face morphs and parts, `TextureLighting`, hair colour and weight are not links, so on the `target=` path they stay the *destination* NPC's own values. That is exactly what the deleted skill carried (`plugin/CHANGELOG.md:1746`: "which inline fields (tints, morphs, `TextureLighting`, weight) ride `housecarl_apply`'s copy zip").

After this PR the only sentence about inline face values anywhere on the surface is this one, and it ends "no separate write of them is owed, and seeding them would only re-set what is already there."

Scenario: "copy that overhaul's face onto my existing follower" — the caller reads this sentence, uses `target=` (the right destination for an existing NPC), skips the `housecarl_apply` write, and ships a patch carrying the donor's `HeadParts` over the follower's own tints and morphs. Wrong face, and nothing on the tool said a second call was owed. The clone half is now covered and the `target=` half is not; before this PR both were.

src/housecarl-mcp/CopyTools.cs
```diff
@@ -29,7 +29,12 @@ public static class CopyTools
          "OUTPUT — patch= names a NEW plugin (folder-per-patch); into= extends an existing houseCARL patch " +
          "instead.\n\n" +
          "This tool copies RECORDS. The FILES that go with them — an NPC's FaceGen mesh and tint, say — are " +
-         "placed with " + ToolNames.Place + ". Originals are never touched, and a refusal writes nothing.")]
+         "placed with " + ToolNames.Place + "; once they are placed, " + ToolNames.Check + " over the patch " +
+         "sweeps it off-order, before it is enabled, and the textures baked INSIDE a placed mesh are not copied " +
```
Avick3110: **The `housecarl_check` clause hangs off the file placement, and check cannot see files.**

As written: "The FILES that go with them ... are placed with `housecarl_place`; **once they are placed**, `housecarl_check` over the patch sweeps it off-order, before it is enabled". The temporal clause attaches the sweep to the placement, so it reads as a verification of what was just placed.

`housecarl_check`'s `findings=` takes `errors`, `scripts` and `dialogue` only — no facegen or asset family. `.claude/skills/facegen-diagnostics/SKILL.md:219` spends a sentence correcting exactly this misreading: "`housecarl_check` has **no facegen finding family** ... so a call there returns nothing for this job."

Scenario: an agent copies an NPC, places the `.nif`/`.dds`, runs `housecarl_check` over the patch, gets a clean record sweep and reports the copy verified — while the mesh's baked tint path still points at the source mod's `.dds`, which is the failure the very next clause of this same description describes.

Detaching the two would fix it, e.g. "... are placed with `housecarl_place`. The patch's own records sweep off-order with `housecarl_check` before it is enabled — that sweep is records only. The textures baked INSIDE a placed mesh ..."

.claude/skills/facegen-diagnostics/SKILL.md
```diff
@@ -213,9 +213,9 @@ the baked ChangeForm. houseCARL does not edit saves.
 
 ## 8. A whole-order sweep is a bulk job
 
-"A bunch of NPCs went dark after I installed X" is an enumerate-and-dedupe job: plan it with
-`housecarl:bulk-record-jobs` — every `NPC_` the suspect plugin touches, deduped to load-order winners,
-spilled to an artifact — and bring the flagged subset back here for the pair diff and the fix.
+"A bunch of NPCs went dark after I installed X" is an enumerate-and-dedupe job: read every `NPC_` the
+suspect plugin touches in ONE `housecarl_records` call, deduped to load-order winners and spilled with
```
Avick3110: **"deduped to load-order winners" needs `fields_source="winner"`, and the call as written does not carry it.**

Under a `plugins=` scope the default renders the scoped plugin's OWN values, not the winner's — `src/housecarl-mcp/RecordsTools.cs:129` states it: "the defining plugin's era, not the number the game uses after later overrides — so a deliverable claiming live stats passes `"winner"` here."

Scenario: the reader follows this line literally, calls `housecarl_records` with `plugins=["Suspect.esp"]` and `to_file=`, and gets an artifact of the suspect plugin's own `NPC_` rows while the text told them those rows are load-order winners. They then flag or clear NPCs on values a later override has already replaced, and the flagged subset that comes back for the pair diff is the wrong subset.

The old text ceded the plan to `bulk-record-jobs`, which carried the parameter; the replacement names the call but drops it. Adding `fields_source="winner"` to the sentence closes it.

src/housecarl-mcp/CopyTools.cs
Avick3110: Folded in 0f951eb4. The sentence now lives on `target=`, where the behaviour is: only the link fields named in `seed_paths=` are written, everything else keeps the destination record's own value, and an NPC's inline face values (tint layers, face morphs and parts, `TextureLighting`, weight) are not links at all, so no seed can carry them. It then names the second call owed — `housecarl_apply`'s copy zip, `bundle=["TintLayers","FaceMorph","FaceParts","TextureLighting","HairColor","HeadTexture","Weight"]` with `assignments=` mapping the destination to the source NPC — and points back at `new_editorid=` as the half that owes none of it. The `new_editorid=` sentence is unchanged and still true for the clone. Lengths: `target=` 854 characters, `new_editorid=` 701, both under 2,048.

src/housecarl-mcp/CopyTools.cs
Avick3110: Folded in 4e08f687, detached as you suggested: "... are placed with `housecarl_place`. The patch's own records sweep off-order with `housecarl_check` before it is enabled — that sweep is RECORDS ONLY and sees no files. The textures baked INSIDE a placed mesh are not copied by either tool ..." The sweep no longer hangs off the placement and says out loud that it sees no files, so the clean-record-sweep-means-copy-verified read is closed. Published tool description is 1,528 characters (bound 2,048).

.claude/skills/facegen-diagnostics/SKILL.md
Avick3110: Folded in e3c16cb4. The sentence now carries the parameter and says why: "deduped to load-order winners with `fields_source=\"winner\"` — under a `plugins=` scope the default renders that plugin's OWN values, not the ones the game uses — and spilled with `to_file=` to an artifact". The call as written now matches what the text claims the rows are.

PR #674, opened 2026-09-08
Say when a skypatcher_layer filter matches no INI

### LINKED ISSUE #665: [Bug] housecarl_skypatcher_layer filter= that matches nothing returns the unfiltered overview
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/skypatcher-authoring/VERDICT.md`).

`housecarl_skypatcher_layer filter="weapon"` on an order with no weapon type folder returned the whole-layer overview cut at max_chars, indistinguishable from a call with no filter. A zero-match filter should say so in one sentence: the filter, its match count (0), and what folders exist. Silently degraded mode.

Avick3110:
`housecarl_skypatcher_layer filter=weapon` on an order with no weapon type folder returned the whole-layer overview, cut at max_chars — indistinguishable from a call that passed no filter, so a typo or a folder that is not there read as the layer's contents. The render now checks the filter's own match domain (type folder, providing mod, INI path/filename) before it writes anything: zero matches returns one plain statement — the filter, `0 of N INI(s) match`, the type folders that are present, a nearest-name suggestion where one clears the bar, and to omit `filter=` for the overview — plus the usual caveats, so an incomplete read still says so. An empty layer says the order has no SkyPatcher INIs at all rather than listing nothing. A filter with at least one match is unchanged. `filter=` is the tool's only selection parameter (`max_chars` is transport), so there is no second fall-through of this shape here.

Covered by three unit tests over the renderer: the zero-match answer, the empty-layer wording, and a control that a matching filter still expands the file to its lines. The first two fail on the code before this change.

Build clean; 1897 tests pass; ci-all 128/128.

Closes #665

### DIFF
--- a/src/housecarl-mcp/SkyPatcherTools.cs
+++ b/src/housecarl-mcp/SkyPatcherTools.cs
@@ -55,6 +55,9 @@ static class SkyPatcherWire
     {
         var sb = new StringBuilder();
         var folders = d.Scan.Folders;
+        // A filter matching nothing must never fall through to the unfiltered overview — that reads as the whole layer.
+        if (filter is { Length: > 0 } && !folders.Any(f => f.Files.Any(x => Matches(filter, f, x))))
+            return ZeroMatch(d, filter);
         int files = folders.Sum(f => f.Files.Count);
         int applied = folders.Sum(f => f.PatchingEnabled ? f.Files.Count(x => x.NotApplied is null) : 0);
         int lines = folders.Sum(f => f.Files.Sum(x => x.Lines.Count(l => l.Kind == SkyPatcherLineKind.Patch)));
@@ -72,8 +75,6 @@ static class SkyPatcherWire
           .Append(d.NoOps.Count).Append(" no-op write(s)\n");
         if (folders.Count == 0)
             sb.Append("\nno SkyPatcher INIs in the active order (no Data\\SKSE\\Plugins\\SkyPatcher content, or SkyPatcher itself is not installed).\n");
-
-        bool In(string? s) => filter is null || (s is not null && s.Contains(filter, StringComparison.OrdinalIgnoreCase));
 
         foreach (var f in folders)
         {
@@ -93,7 +94,7 @@ static class SkyPatcherWire
                 if (file.ShadowedProviders.Count > 0) sb.Append("  [!] shadows same-path copies from ").Append(string.Join(", ", file.ShadowedProviders));
                 sb.Append('\n');
                 // filter= match (folder, provider, or filename) expands the file to its patch lines.
-                bool expand = filter is not null && (In(f.Subfolder) || In(file.WinningProvider) || In(file.RelPath));
+                bool expand = filter is not null && Matches(filter, f, file);
                 if (expand)
                     for (int i = 0; i < file.Lines.Count; i++)
                     {
@@ -208,6 +209,37 @@ static class SkyPatcherWire
         }
         AppendCaveats(sb, d.ReadIncomplete, d.AssetWarnings);
         sb.Append("\n→ " + ToolNames.Records + " formids=['<FormID>'] source={\"overlay\": \"skypatcher\", \"state\": \"post\"} for one record's computed post-SkyPatcher state; filter='<folder/mod/file>' to expand files to their lines.");
+        return sb.ToString().TrimEnd('\n');
+    }
+
+    /// <summary>The filter's match domain: the file's type folder, its providing mod, or its path/filename.</summary>
+    static bool Matches(string filter, SkyPatcherDiscovery.FolderScan folder, SkyPatcherDiscovery.IniFile file)
+    {
+        bool In(string? s) => s is not null && s.Contains(filter, StringComparison.OrdinalIgnoreCase);
+        return In(folder.Subfolder) || In(file.WinningProvider) || In(file.RelPath);
+    }
+
+    /// <summary>What a filter matching no INI returns: the count (zero), what the filter is matched against, and the
+    /// folders that are there. Never the overview — an unfiltered dump would read as the filter's own result.</summary>
+    static string ZeroMatch(SkyPatcherLayerData d, string filter)
+    {
+        var folders = d.Scan.Folders;
+        int files = folders.Sum(f => f.Files.Count);
+        var sb = new StringBuilder();
+        sb.Append("SkyPatcher layer — filter '").Append(filter).Append("' — 0 of ").Append(files)
+          .Append(" INI(s) match [profile '").Append(d.ProfileName).Append("']\n\n");
+        if (folders.Count == 0)
+            sb.Append("nothing matched: the active order has no SkyPatcher INIs at all (no Data\\SKSE\\Plugins\\SkyPatcher content, or SkyPatcher itself is not installed).\n");
+        else
+        {
+            sb.Append("nothing matched: no type folder, providing mod, or INI filename contains '").Append(filter).Append("'.")
+              .Append(PluginNameSuggest.DidYouMean(filter, folders.Select(f => f.Subfolder)
+                  .Concat(folders.SelectMany(f => f.Files).Select(x => x.WinningProvider).Where(p => p is not null)!)
+                  .Concat(folders.SelectMany(f => f.Files).Select(x => x.SortKey))))
+              .Append(" The type folder(s) present are: ").Append(string.Join(", ", folders.Select(f => f.Subfolder)))
+              .Append(". Omit filter= for the whole-layer overview.\n");
+        }
+        AppendCaveats(sb, d.ReadIncomplete, d.AssetWarnings);
         return sb.ToString().TrimEnd('\n');
     }
 
--- a/src/housecarl-mcp/SkyPatcherTools.cs
+++ b/src/housecarl-mcp/SkyPatcherTools.cs
@@ -28,20 +28,20 @@ public static class SkyPatcherTools
          "and NO-OP WRITES (true ITM — the replay shows the SET writes the value the record already has). " +
          "Entries whose applicability " +
          "also hangs on other filters are flagged conditional rather than guessed. Pass filter= a type folder, mod, " +
-         "or filename substring to expand matching files to their patch lines. For ONE record's computed " +
+         "or filename substring to list only the matching files, expanded to their patch lines. For ONE record's computed " +
          "post-SkyPatcher state use " + ToolNames.Records + " formids=[\"<FormID>\"] source={\"overlay\": \"skypatcher\", \"state\": \"post\"} — source= is a version pole, not a selection, so the read needs formids= (or a scan scope). " +
          "Read-only.")]
     public static string SkyPatcherLayer(
         LoadOrderService svc,
         [Description("Optional. A type-folder (e.g. 'weapon'), providing-mod, or INI filename substring (case-insensitive). " +
-            "Expands the matching files to their individual patch lines. Omit for the whole-layer overview.")]
+            "Lists only the matching files, expanded to their individual patch lines. Omit for the whole-layer overview.")]
             string? filter = null,
         [Description("Optional. Max characters before lists are cut with an explicit notice. 0 = the server default (~80k).")]
             int max_chars = 0) => Guard.Tool(ToolNames.SkypatcherLayer, () =>
     {
         if (svc.ConfigPromptOrNull() is { } prompt) return prompt;
         var data = svc.SkyPatcherLayer();
-        return SkyPatcherWire.RenderLayer(data, filter?.Trim(), max_chars > 0 ? max_chars : 80_000);
+        return SkyPatcherWire.RenderLayer(data, filter, max_chars > 0 ? max_chars : 80_000);
     });
 }
 
@@ -55,9 +55,10 @@ static class SkyPatcherWire
     {
         var sb = new StringBuilder();
         var folders = d.Scan.Folders;
+        filter = string.IsNullOrWhiteSpace(filter) ? null : filter.Trim();   // a blank filter is no filter, never a match-everything
         // A filter matching nothing must never fall through to the unfiltered overview — that reads as the whole layer.
-        if (filter is { Length: > 0 } && !folders.Any(f => f.Files.Any(x => Matches(filter, f, x))))
-            return ZeroMatch(d, filter);
+        if (filter is { } zero && !folders.Any(f => f.Files.Any(x => Matches(zero, f, x))))
+            return ZeroMatch(d, zero);
         int files = folders.Sum(f => f.Files.Count);
         int applied = folders.Sum(f => f.PatchingEnabled ? f.Files.Count(x => x.NotApplied is null) : 0);
         int lines = folders.Sum(f => f.Files.Sum(x => x.Lines.Count(l => l.Kind == SkyPatcherLineKind.Patch)));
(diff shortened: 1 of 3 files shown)


src/housecarl-mcp/SkyPatcherTools.cs
```diff
@@ -211,6 +212,37 @@ public static string RenderLayer(SkyPatcherLayerData d, string? filter, int cap)
...
+        else
+        {
+            sb.Append("nothing matched: no type folder, providing mod, or INI filename contains '").Append(filter).Append("'.")
+              .Append(PluginNameSuggest.DidYouMean(filter, folders.Select(f => f.Subfolder)
+                  .Concat(folders.SelectMany(f => f.Files).Select(x => x.WinningProvider).Where(p => p is not null)!)
+                  .Concat(folders.SelectMany(f => f.Files).Select(x => x.SortKey))))
+              .Append(" The type folder(s) present are: ").Append(string.Join(", ", folders.Select(f => f.Subfolder)))
+              .Append(". Omit filter= for the whole-layer overview.\n");
+        }
+        AppendCaveats(sb, d.ReadIncomplete, d.AssetWarnings);
```
Avick3110: **ZeroMatch drops the scan notes, which are exactly what explains a zero match.**

The overview path renders `d.NoOpNotes` and `d.Scan.Notes` as `[!]` lines just above `AppendCaveats` (lines 198-209). The early return carries only `AppendCaveats`, so both note lists are silently discarded.

Those notes are the diagnostics a caller most needs on this exact path. `SkyPatcherDiscovery.Scan` adds, among others:

- `'<rel>' is shipped by N mods — only '<X>' wins the VFS; the cop(ies) from <Y> are SHADOWED and never read`
- `subfolder '<s>' is not a documented SkyPatcher record type — its N INI(s) are listed but cannot be interpreted`

Scenario: a mod ships its INIs under a misspelled subfolder (`weapons` rather than `weapon`). The scan records the undocumented-subfolder note. `filter=weapon` matches nothing under the new guard, so the answer is `nothing matched: no type folder, providing mod, or INI filename contains 'weapon'` plus the folder list — and the one note saying the layer contains an undocumented folder is suppressed. Before this change the caller at least saw it in the fall-through overview.

Same for a shadowed provider: `filter=<mod name>` where that mod's copies lose the VFS returns "nothing matched" with no mention that the mod's files are present but shadowed — the caller concludes the mod ships no INIs.

Appending the two note loops to `ZeroMatch` before `AppendCaveats` fixes it.

src/housecarl-mcp/SkyPatcherTools.cs
```diff
@@ -73,8 +76,6 @@ public static string RenderLayer(SkyPatcherLayerData d, string? filter, int cap)
         if (folders.Count == 0)
             sb.Append("\nno SkyPatcher INIs in the active order (no Data\\SKSE\\Plugins\\SkyPatcher content, or SkyPatcher itself is not installed).\n");
 
-        bool In(string? s) => filter is null || (s is not null && s.Contains(filter, StringComparison.OrdinalIgnoreCase));
-
         foreach (var f in folders)
         {
             if (sb.Length >= cap) { sb.Append("... [remaining folders omitted at max_chars — raise it or pass filter=]\n"); break; }
```
Avick3110: **The second fall-through of this shape is right here: `max_chars` can eat the match, and the notice then tells a caller who passed `filter=` to pass `filter=`.**

The PR body says "`filter=` is the tool's only selection parameter (`max_chars` is transport), so there is no second fall-through of this shape here." But `filter=` never *selects* — it only *expands*. This loop prints every folder header and every file line unconditionally; matching is additive, applied at line 97.

So on a layer whose plain inventory alone exceeds the cap, `filter=SomeMod` where that mod's folder sorts late breaks out here before the match is ever reached. The caller gets a truncated, unfiltered folder-and-file inventory with zero expanded lines — the same "indistinguishable from a call that passed no filter" failure this PR closes, just reached through the cap instead of the fall-through. And the cut notice reads `raise it or pass filter=` to someone who already did.

Concrete: ~300 INIs (a large SkyPatcher-heavy order is not unusual), default `max_chars` 80000, `filter=weapon` where `weapon` sorts after the cap is hit. Nothing about the answer says the filter's own target was never rendered.

Two cheap fixes, either sufficient: skip non-matching folders/files entirely when `filter` is set (the guard above now guarantees at least one match, so the output can never be empty), or at minimum branch the notice text on `filter is not null`.

src/housecarl-mcp/SkyPatcherTools.cs
```diff
@@ -55,6 +55,9 @@ public static string RenderLayer(SkyPatcherLayerData d, string? filter, int cap)
     {
         var sb = new StringBuilder();
         var folders = d.Scan.Folders;
+        // A filter matching nothing must never fall through to the unfiltered overview — that reads as the whole layer.
```
Avick3110, line 62: **Two different notions of "a filter is present" now live in one function, and the gap between them is the worst case.**

The guard here tests `filter is { Length: > 0 }`. The expansion at line 97 tests `filter is not null`. An empty-after-trim filter falls in the gap: `SkyPatcherLayer` passes `filter?.Trim()` (line 44), so `filter=""` or `filter="   "` arrives as `""` — non-null, length 0.

Result: the guard is skipped, then `Matches` runs `s.Contains("")`, which is `true` for every string, so `expand` is true for **every INI in the layer**. Every patch line of every file is rendered and the answer is cut at `max_chars`.

That predates this PR (the old `In` had the same `Contains("")` behaviour), but the PR is what introduces the `Length: > 0` / `is not null` asymmetry, and it is the same class of bug being fixed: a caller who thinks they filtered gets a truncated whole-layer dump instead.

Normalising once at the call site makes both tests agree and closes it:

```suggestion
        // A filter matching nothing must never fall through to the unfiltered overview — that reads as the whole layer.
        if (filter is { Length: 0 }) filter = null;                 // an empty filter is no filter, not a match-everything
        if (filter is not null && !folders.Any(f => f.Files.Any(x => Matches(filter, f, x))))
```

src/housecarl-mcp/SkyPatcherTools.cs
```diff
@@ -211,6 +212,37 @@ public static string RenderLayer(SkyPatcherLayerData d, string? filter, int cap)
...
+    }
+
+    /// <summary>What a filter matching no INI returns: the count (zero), what the filter is matched against, and the
+    /// folders that are there. Never the overview — an unfiltered dump would read as the filter's own result.</summary>
+    static string ZeroMatch(SkyPatcherLayerData d, string filter)
+    {
+        var folders = d.Scan.Folders;
+        int files = folders.Sum(f => f.Files.Count);
+        var sb = new StringBuilder();
+        sb.Append("SkyPatcher layer — filter '").Append(filter).Append("' — 0 of ").Append(files)
```
Avick3110, line 252: Minor, but this branch states an absolute the scan cannot back when the read was incomplete.

`folders.Count == 0` also happens when a BSA failed to read: `AppendCaveats` two lines below prints `a BSA failed to read this build, so an INI present only in it may be missing from this scan (Q3)`. So on that path the answer says "the active order has no SkyPatcher INIs **at all**" and then immediately says the scan may have missed some. "at all" is stronger than the overview path's wording (`no SkyPatcher INIs in the active order`) and is the half a caller quotes back.

Hedging it when `d.ReadIncomplete` is set — "none were found; the read was incomplete" — keeps the sentence true. The new test `AFilterMatchingNothingInAnEmptyLayerSaysTheLayerIsEmpty` builds its layer with `ReadIncomplete: false`, so it would not notice a change here.

src/housecarl-mcp/SkyPatcherTools.cs
Avick3110: Folded in 4751e6e. ZeroMatch now renders `d.NoOpNotes` and `d.Scan.Notes` as `[!]` lines, same shape as the overview, just above the caveats. Test: `ZeroMatchStillRendersTheScanNotes`.

src/housecarl-mcp/SkyPatcherTools.cs
Avick3110: Folded in 4751e6e, the stronger of the two fixes. `filter=` now selects as well as expands: a folder with no matching file is skipped, a non-matching file is skipped, and the expansion is the same `Matches` call that selected it, so it stays one rule. A header line says `filter 'x' — N of M INI(s) match; only those are listed below`, since the summary counts above it are the whole layer. The folder-level cut notice branches on the filter: `raise it or narrow filter=` when one was passed. Tests: `AFilterListsOnlyTheMatchingFolder`, `AMatchingFolderIsNotLostToTheCapAndTheCutNoticeDoesNotSuggestAFilter`, `AnUnfilteredCutStillSuggestsAFilter` (the last two pick their cap from the unfiltered render, so they do not depend on counting characters).

src/housecarl-mcp/SkyPatcherTools.cs
Avick3110: Folded in 4751e6e. Normalised once at the top of `RenderLayer`: `filter = string.IsNullOrWhiteSpace(filter) ? null : filter.Trim();` — whitespace included, and the `Trim()` at the tool call site is dropped so there is one place that decides. Both the guard and the expansion then test `is not null` and agree. Tests: `ABlankFilterIsNoFilterAtBothSites` for `""` and `"   "`, asserting the overview appears and no INI is expanded.

src/housecarl-mcp/SkyPatcherTools.cs
Avick3110: Folded in 4751e6e. When `d.ReadIncomplete` is set the sentence reads `nothing matched: no SkyPatcher INIs were found, and the read was incomplete (below), so the layer is not necessarily empty.`; the flat "at all" wording is kept for a complete read. New test `AnEmptyLayerWithAnIncompleteReadIsNotCalledEmpty` covers the incomplete case.

src/housecarl-mcp/SkyPatcherTools.cs
```diff
@@ -72,28 +76,39 @@ public static string RenderLayer(SkyPatcherLayerData d, string? filter, int cap)
...
+            if (filter is { } sel && !f.Files.Any(x => Matches(sel, f, x))) continue;
+            if (sb.Length >= cap)
+            {
+                sb.Append(filter is null
+                    ? "... [remaining folders omitted at max_chars — raise it or pass filter=]\n"
+                    : "... [remaining matching folders omitted at max_chars — raise it or narrow filter=]\n");
+                break;
+            }
             int fLines = f.Files.Sum(x => x.Lines.Count(l => l.Kind == SkyPatcherLineKind.Patch));
             sb.Append("\n").Append(f.Subfolder).Append(": ").Append(f.Files.Count).Append(" INI(s), ").Append(fLines).Append(" patch line(s)");
```
Avick3110: The per-folder header still counts the whole folder while only the matching files are listed beneath it.

With `filter='My Bandit Tweaks'` against an `npc` folder of 47 INIs from many mods, this prints `npc: 47 INI(s), 900 patch line(s)` and then one file. The filter line above caveats only "the counts above are the whole layer" — the layer summary — not this header, which comes after it and reads as a description of the list that follows. Before the change every file was listed, so header and listing always agreed; now they only agree when the filter matches the folder name.

Either count the matching files here, or say the header is the whole folder:

```suggestion
            int fShown = filter is { } fq ? f.Files.Count(x => Matches(fq, f, x)) : f.Files.Count;
            sb.Append("\n").Append(f.Subfolder).Append(": ").Append(f.Files.Count).Append(" INI(s)");
            if (fShown != f.Files.Count) sb.Append(" (").Append(fShown).Append(" matching, listed)");
            sb.Append(", ").Append(fLines).Append(" patch line(s)");
```

src/housecarl-mcp/SkyPatcherTools.cs
```diff
@@ -72,28 +76,39 @@ public static string RenderLayer(SkyPatcherLayerData d, string? filter, int cap)
...
             if (!f.PatchingEnabled) sb.Append("  [!] toggled OFF in SkyPatcher.ini — the DLL skips this whole folder");
             if (f.Catalog is null) sb.Append("  [!] not a documented SkyPatcher record type — content listed, not interpreted");
             sb.Append('\n');
             foreach (var file in f.Files)
             {
+                // filter= match (folder, provider, or filename) selects the file and expands it to its patch lines.
+                bool expand = filter is { } q && Matches(q, f, file);
+                if (filter is not null && !expand) continue;
```
Avick3110: Selecting instead of expanding drops the siblings that show where the matched INI sorts — which is what the skill sends the user here for.

`.claude/skills/skypatcher-authoring/SKILL.md:129` says to run this tool with `filter=` the INI filename for the file-level verdict: "whether the file is read at all, **where it sorts**, what it conflicts with". After this change a filter on `zz_MyPatch.ini` lists that one file and nothing else, so the INIs sorting before and after it — the thing that decides whether its writes survive — are gone from the answer. The apply order is also what the tool's own description leads with ("per type folder in filename apply order").

The cap problem the change is solving is real, but it is a cap problem: listing the matching folders' files and expanding only the matches would fix the late-sort cut without losing the neighbours. If selection is the intended trade, the skill text needs updating in the same PR, and the answer should say the siblings were withheld.

src/housecarl-mcp/SkyPatcherTools.cs
```diff
@@ -207,7 +222,43 @@ public static string RenderLayer(SkyPatcherLayerData d, string? filter, int cap)
...
+        {
+            sb.Append("nothing matched: no type folder, providing mod, or INI filename contains '").Append(filter).Append("'.")
+              .Append(PluginNameSuggest.DidYouMean(filter, folders.Select(f => f.Subfolder)
+                  .Concat(folders.SelectMany(f => f.Files).Select(x => x.WinningProvider).Where(p => p is not null)!)
+                  .Concat(folders.SelectMany(f => f.Files).Select(x => x.SortKey))))
+              .Append(" The type folder(s) present are: ").Append(string.Join(", ", folders.Select(f => f.Subfolder)))
+              .Append(". Omit filter= for the whole-layer overview.\n");
+        }
+        // The scan notes (shadowed copies, undocumented subfolders) are often why the filter matched nothing.
+        foreach (var note in d.NoOpNotes) sb.Append("[!] ").Append(note).Append('\n');
```
Avick3110: `ZeroMatch` never sees `cap`, so the zero-match answer ignores `max_chars`.

Neither note loop nor `AppendCaveats` is bounded here, where every other block in `RenderLayer` tests `sb.Length >= cap` before it writes. `Scan.Notes` carries one ~250-char note per same-path INI collision and per undocumented subfolder, so on a large layer a caller who passes `max_chars=2000` and mistypes the filter gets the whole note set back with no cut and no notice — where before the change the same call got the overview cut at 2000 with an explicit notice.

Pass `cap` in and break on it in the two loops, the way the same notes are handled at lines 213 and 219.

src/housecarl-mcp/SkyPatcherTools.cs
Avick3110: Folded, as the shape you describe. Selection moved up to the folder: a folder with no matching file is skipped (which is what keeps a late-sorting match inside the cap), but a folder with one lists every file in apply order and expands only the matches, so the neighbours around a match are back.

Wording follows: the header line now says "N of M INI(s) match, in K of J type folder(s); only those folders are listed below, each in full apply order with the matching files expanded to their lines"; the tool description and the `filter=` parameter description both say the filter narrows to the folders holding a match, each listed in full so the files sorting around a match stay visible; the trailing hint says the same. The skill sentence at SKILL.md:129 is true again under this shape ("where it sorts" is answered), so I left it alone.

Test: `AMatchedFilesSiblingsAreStillListedUnexpanded` — one folder, `aa_Before.ini` / `zz_MyPatch.ini` / `zzz_After.ini`, filter on the middle filename; all three listed, in apply order, only the match's lines expanded.

src/housecarl-mcp/SkyPatcherTools.cs
Avick3110: Folded, in your suggestion's shape. The folder header now counts what is expanded beneath it:

    npc: 3 INI(s) (1 matching, expanded), 3 patch line(s)

Unfiltered, and when every file of the folder matches, the parenthetical is omitted and the header reads as before. With the folder-level selection from the other thread every file is listed again, so "expanded" is the honest word rather than "listed".

Tests: `TheFolderHeaderCountsWhatIsExpandedUnderIt` asserts the exact header, `AnUnfilteredFolderHeaderCarriesNoMatchCount` asserts the parenthetical stays off when there is no filter.

src/housecarl-mcp/SkyPatcherTools.cs
Avick3110: Folded. `ZeroMatch` takes `cap` now. The two note loops are one loop over `NoOpNotes.Concat(Scan.Notes)` breaking on `sb.Length >= cap`, and a cut is announced explicitly — `... [showing 7 of 40 note(s); raise max_chars]` — in the shape the conflict and no-op blocks use.

One deliberate difference from your line: `AppendCaveats` stays unconditional, as it is at line 222 of `RenderLayer`. A cut must not be what swallows the read-incomplete warning; the class doc already says the caveats are always rendered. Say the word if you want it bounded too.

Tests: `ZeroMatchCutsItsNotesAtMaxCharsWithANotice` (40 notes, `max_chars=2000` — shorter than the unbounded render, carries the notice, drops the last note) and `ZeroMatchAlwaysRendersTheCaveatsEvenPastTheCap`.

Issue #660, opened 2026-09-08
[Bug] DIAL Subtype decodes by source record version: FormVersion-40 read prints RechargeExit where SubtypeName says HELO


Avick3110:
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/dialogue-authoring/VERDICT.md`).

Reading DIAL `02707A:Skyrim.esm` from Skyrim.esm's own copy (FormVersion 40) renders `Subtype = RechargeExit` beside `SubtypeName = HELO`; the same record at its FormVersion-44 winner renders `Subtype = Hello`. Second reproducer: `0904AC` renders `Recharge` beside `GBYE`. The enum decode appears to be applied without regard to the record's form version, so an off-order or master-copy read reports the wrong subtype while the name field is right.

Expected: the same Subtype for the same bytes regardless of which copy is read, or a note that the enum is version-dependent.

Avick3110:
Closing: the houseCARL half shipped in #675 and #676 (SNAM rendered as authoritative, a contradicting Subtype flagged). The upstream Mutagen half is tracked in #696, unmilestoned, so 2.0 does not wait on it.

PR #675, opened 2026-09-08
Flag a dialogue topic whose Subtype contradicts its SNAM marker

Avick3110:
A DIAL carries its subtype twice: the numeric DATA subtype field and the 4-char SNAM marker. SNAM is authoritative — the engine buckets topics by it — and the number is stale on any topic authored before the Dragonborn-era Creation Kit inserted six `FlyingMount*` values at index 20, so a vanilla `HELO` topic reads back as `RechargeExit`. Nothing on the record distinguishes the two numberings (form version does not: Dragonborn.esm mixes both at FormVersion 43), so this is reported, never rewritten. The dialogue check now warns when the two disagree, naming the marker as the one to believe and the renumbering as the cause; both renders label the pair (`subtype=... (stale)` / `subtype_marker=... (authoritative)` in text, `subtype_stale` in JSON); and the blank-SNAM advice, which derives its recommended marker from that same unreliable field, now says so and points at the base record's marker. The generic `project=fields` render is untouched — a per-record special case there is the hand-written coverage the cornerstones forbid.

Covered by three tests on the dialogue world, which gains a topic with the #660 shape (Subtype=RechargeExit under SNAM=HELO) and an agreeing control: the warning and the text labels, the no-false-positive control, and the JSON flag.

Part of #660. The upstream half — teaching Mutagen to derive `Subtype` from SNAM, or expose a resolved one (Mutagen-Modding/Mutagen#590) — is separate and stays open; this is the houseCARL-side mitigation and does not depend on it.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -35,8 +35,10 @@ saying it sets an expectation their install may contradict. Say what is known, a
 - **A dialogue topic whose numeric `Subtype` contradicts its SNAM marker is now flagged, and the marker is named
   as the authoritative one.** The engine buckets topics by SNAM; the numeric subtype is stale on topics authored
   before the Dragonborn-era Creation Kit inserted six `FlyingMount*` values into the enum, so a vanilla `HELO`
-  topic can read as `RechargeExit`. The dialogue check now warns on the disagreement, the text and JSON renders
-  label the pair (`subtype=… (stale)`, `subtype_marker=… (authoritative)`, `subtype_stale`), and the blank-marker
+  topic can read as `RechargeExit`. The dialogue check now warns on the disagreement — only for a topic a mod
+  defines or overrides, so a whole-quest check over vanilla topics stays quiet — and both renders label the pair
+  (`subtype=… (stale)`, `subtype_marker=… (authoritative)`, `subtype_stale`, `subtype_from_marker`) on every
+  topic. A non-blank marker houseCARL does not model is named rather than passed over, and the blank-marker
   advice says its recommended marker is derived from that unreliable field. Nothing is rewritten: no field on the
   record distinguishes the two numberings.
 
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -32,6 +32,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
   and compares that record's type against the link's own target, naming the field, the type given and the types
   allowed. A field whose link accepts any record, a null-clear, a FormID no plugin in the order defines, and
   removing a link already in a list are not refused.
+- **A dialogue topic whose numeric `Subtype` contradicts its SNAM marker is now flagged, and the marker is named
+  as the authoritative one.** The engine buckets topics by SNAM; the numeric subtype is stale on topics authored
+  before the Dragonborn-era Creation Kit inserted six `FlyingMount*` values into the enum, so a vanilla `HELO`
+  topic can read as `RechargeExit`. The dialogue check now warns on the disagreement, the text and JSON renders
+  label the pair (`subtype=… (stale)`, `subtype_marker=… (authoritative)`, `subtype_stale`), and the blank-marker
+  advice says its recommended marker is derived from that unreliable field. Nothing is rewritten: no field on the
+  record distinguishes the two numberings.
 
 - **`open-animation-replacer` now states how a DAR `_conditions.txt` chain binds, and that the DAR weapon-type
   numbers convert unchanged.** Both were marked unverified in its reference, and the skill told you to convert
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -35,12 +35,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
 - **A dialogue topic whose numeric `Subtype` contradicts its SNAM marker is now flagged, and the marker is named
   as the authoritative one.** The engine buckets topics by SNAM; the numeric subtype is stale on topics authored
   before the Dragonborn-era Creation Kit inserted six `FlyingMount*` values into the enum, so a vanilla `HELO`
-  topic can read as `RechargeExit`. The dialogue check now warns on the disagreement — only for a topic a mod
-  defines or overrides, so a whole-quest check over vanilla topics stays quiet — and both renders label the pair
-  (`subtype=… (stale)`, `subtype_marker=… (authoritative)`, `subtype_stale`, `subtype_from_marker`) on every
-  topic. A non-blank marker houseCARL does not model is named rather than passed over, and the blank-marker
-  advice says its recommended marker is derived from that unreliable field. Nothing is rewritten: no field on the
-  record distinguishes the two numberings.
+  topic can read as `RechargeExit`. The dialogue check now warns on the disagreement, and on a non-blank marker
+  houseCARL does not model. Both warnings are scoped to a pair the plugin itself states: a topic whose winner is
+  force-loaded content (the base masters, Creation Club, `_ResourcePack.esl` — the implicit group
+  `housecarl_load_order_status` lists) is passed over, and so is an override that carries the base record's pair
+  forward unchanged, so a whole-quest check over vanilla topics stays quiet. The labels are not scoped: both
+  renders show the pair (`subtype=… (stale)`, `subtype_marker=… (authoritative)`, `subtype_stale`,
+  `subtype_from_marker`) on every topic. The advice follows the numbers: a stored value exactly six below the
+  marker's modern index is the renumbering, and any other gap is two fields edited apart, where the warning says
+  the `Subtype` edit does nothing in game until SNAM is synced to it — which is what setting `Subtype` through
+  houseCARL does for you. Where a recommendation is derived from that unreliable field rather than from the base
+  record's SNAM, it says so. Nothing is rewritten: no field on the record distinguishes the two numberings.
 
 - **`open-animation-replacer` now states how a DAR `_conditions.txt` chain binds, and that the DAR weapon-type
   numbers convert unchanged.** Both were marked unverified in its reference, and the skill told you to convert
--- a/src/housecarl-core/DialogueSubtype.cs
+++ b/src/housecarl-core/DialogueSubtype.cs
@@ -181,6 +181,13 @@ public static class DialogueSubtype
     /// unmodeled (and "" for the one index Mutagen's enum omits). The honest label for a topic's subtype.</summary>
     public static string? NameForMarker(RecordType marker) => IndexForMarker(marker) is { } i ? NameAt(i) : null;
 
+    /// <summary>The best LABEL for the subtype a marker names, for a render or a message: Mutagen's enum name where
+    /// there is one, and the MARKER ITSELF for the one modeled row Mutagen's enum omits (index 3, FVDL) — never "",
+    /// which would hand a consumer nothing for a subtype this table can in fact name. null only when the marker is
+    /// blank or not one this table models, which is a different finding and says so in its own words.</summary>
+    public static string? LabelForMarker(RecordType marker) =>
+        IndexForMarker(marker) is { } i ? (NameAt(i) is { Length: > 0 } n ? n : Table[i].Marker) : null;
+
     /// <summary>True when a topic's numeric <c>Subtype</c> contradicts its SNAM marker — both modeled, and they name
     /// different subtypes. Bethesda renumbered the DATA\Subtype enum when the Dragonborn-era CK inserted six
     /// <c>FlyingMount*</c> values at index 20, so a topic authored before that stores a number six lower than the
@@ -189,6 +196,23 @@ public static class DialogueSubtype
     /// statement. Blank or unmodeled markers are NOT a disagreement — a blank one is its own finding.</summary>
     public static bool MarkerDisagreesWithSubtype(IDialogTopicGetter topic) =>
         IndexForMarker(topic.SubtypeName) is { } fromMarker && fromMarker != (int)topic.Subtype;
+
+    /// <summary>How many values the Dragonborn-era Creation Kit inserted at index 20 (the six <c>FlyingMount*</c>
+    /// rows), which is exactly how far a pre-Dragonborn DATA\Subtype sits below the modern table.</summary>
+    public const int RenumberOffset = 6;
+
+    /// <summary>The first modern index that also existed under the OLD numbering (20 + <see cref="RenumberOffset"/>).
+    /// A marker below this one names a subtype that predates the insertion or is one of the inserted rows, so it
+    /// cannot have been shifted and a disagreement there is not vintage.</summary>
+    public const int RenumberFirstShifted = 26;
+
+    /// <summary>Does a disagreeing pair carry the RENUMBERING signature — the marker's modern index sitting exactly
+    /// <see cref="RenumberOffset"/> above the stored number, at or above <see cref="RenumberFirstShifted"/>? True
+    /// means "old file, stale number, record not necessarily broken". False means the two fields were edited apart:
+    /// the numbers line up with no vintage that explains them, so the Subtype edit is an in-game no-op and the fix is
+    /// to sync SNAM. Asserting the vintage for EVERY mismatch would frame an authoring error as benign.</summary>
+    public static bool IsRenumberedVintage(int markerIndex, int subtypeValue) =>
+        markerIndex >= RenumberFirstShifted && markerIndex - subtypeValue == RenumberOffset;
 
     /// <summary>True when a topic's SNAM marker is empty/default (0000) OR whitespace-only — the malformed "no real
     /// marker" state. The single home for this test so the create path and the validator agree on what "no marker"
--- a/src/housecarl-core/DialogueSubtype.cs
+++ b/src/housecarl-core/DialogueSubtype.cs
@@ -166,6 +166,29 @@ public static class DialogueSubtype
     /// <summary>The 4-char SNAM marker for a <see cref="DialogTopic.SubtypeEnum"/>, or null if unmodeled.</summary>
     public static string? MarkerFor(DialogTopic.SubtypeEnum subtype) => MarkerFor((int)subtype);
 
+    /// <summary>Marker → subtype index, the reverse of the table. Ordinal (markers are fixed-case 4-char signatures).</summary>
+    static readonly Dictionary<string, int> ByMarker =
+        Table.Select((row, i) => (row.Marker, i)).ToDictionary(p => p.Marker, p => p.i, StringComparer.Ordinal);
+
+    /// <summary>The subtype index a 4-char SNAM marker names, or null when the marker is blank or not one this table
+    /// models. SNAM is the AUTHORITATIVE statement of a topic's subtype — the engine buckets by it — so this is the
+    /// lookup a reader should trust over <c>(int)DialogTopic.Subtype</c>.</summary>
+    public static int? IndexForMarker(RecordType marker) =>
+        !IsBlankMarker(marker) && ByMarker.TryGetValue(marker.Type, out var i) ? i : null;
+
+    /// <summary>Mutagen's SubtypeEnum name for the subtype a SNAM marker names, or null when the marker is blank or
+    /// unmodeled (and "" for the one index Mutagen's enum omits). The honest label for a topic's subtype.</summary>
+    public static string? NameForMarker(RecordType marker) => IndexForMarker(marker) is { } i ? NameAt(i) : null;
+
+    /// <summary>True when a topic's numeric <c>Subtype</c> contradicts its SNAM marker — both modeled, and they name
+    /// different subtypes. Bethesda renumbered the DATA\Subtype enum when the Dragonborn-era CK inserted six
+    /// <c>FlyingMount*</c> values at index 20, so a topic authored before that stores a number six lower than the
+    /// modern table and Mutagen labels it six entries too early. Nothing on the record distinguishes the two
+    /// numberings (form version does not: Dragonborn.esm mixes both at FormVersion 43), so SNAM is the only reliable
+    /// statement. Blank or unmodeled markers are NOT a disagreement — a blank one is its own finding.</summary>
+    public static bool MarkerDisagreesWithSubtype(IDialogTopicGetter topic) =>
+        IndexForMarker(topic.SubtypeName) is { } fromMarker && fromMarker != (int)topic.Subtype;
+
     /// <summary>True when a topic's SNAM marker is empty/default (0000) OR whitespace-only — the malformed "no real
     /// marker" state. The single home for this test so the create path and the validator agree on what "no marker"
     /// means. Whitespace counts as blank: it renders invisibly and buckets to a tag no real topic uses.</summary>
--- a/src/housecarl-core/DialogueSubtype.cs
+++ b/src/housecarl-core/DialogueSubtype.cs
@@ -14,8 +14,9 @@ namespace HousecarlCore;
 //
 // The table below is sourced from xEdit's record definition, not derived: the name↔marker mapping is not a
 // blind echo (Custom→CUST but ShootBow→FIWE), Mutagen does not model it, and it cannot be scraped from
-// vanilla because Bethesda's own DATA\Subtype numbers are unreliable (many HELO topics ship with DATA≠79) —
-// which is why xEdit ignores DATA\Subtype and treats SNAM as the required master field. Rows are the join of
+// vanilla because Bethesda's own DATA\Subtype numbers are stale under the Dragonborn-era renumbering (many HELO
+// topics ship with DATA≠79 — see MarkerDisagreesWithSubtype), which is why xEdit ignores DATA\Subtype and treats
+// SNAM as the required master field. Rows are the join of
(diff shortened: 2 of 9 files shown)


src/housecarl-core/DialogueValidate.cs
```diff
@@ -474,6 +486,24 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
                     + $"and this plugin DEFINES the topic, so a blank marker is a load CTD on load (#131); the record is malformed. {fix}"));
         }
 
+        // --- Subtype vs SNAM disagreement: the numeric DATA\Subtype says one thing and the SNAM marker another.
+        //     SNAM wins — the engine buckets by the marker, and xEdit marks DATA\Subtype cpIgnore for the same reason.
+        //     The usual cause is age, not damage: the Dragonborn-era CK inserted six FlyingMount* subtypes at index 20,
+        //     so a topic authored before that stores a number six lower than the modern enum and every reader (Mutagen,
+        //     xEdit, houseCARL) labels it six entries too early. No field distinguishes the two numberings, so this is
+        //     reported, never "fixed" — rewriting DATA\Subtype here would be a guess at what the author meant.
+        else if (DialogueSubtype.MarkerDisagreesWithSubtype(topic))
```
Avick3110, line 558: This is the blanket "marker != number" lint the codebase elsewhere says must not exist, and it will fire on almost every vanilla topic.

- `DialogueSubtype.cs` header: "Bethesda's own DATA/Subtype numbers are unreliable (many HELO topics ship with DATA != 79)".
- `WritePatchBuilder.SyncEditedTopicMarkers` doc (WritePatchBuilder.cs:734) is explicit: the sync is gated on "this call set Subtype" precisely because "a blanket 'SubtypeName != marker' lint would false-positive on those". That comment is now stale and contradicts shipped behaviour.

By the PR's own analysis the renumbering shifted every subtype index >= 20, so every pre-Dragonborn topic with a real subtype at or above `Attack` (26) stores a number six low. That is the bulk of Skyrim.esm's dialogue.

Concrete scenario: a modder runs the dialogue check over a quest that touches 30 vanilla topics. Each affected topic now emits the same ~640-character essay into `issues`, which `DialogueWire.AppendIssues` renders under a shared `max_chars` cap that cuts with `... [truncated at max_chars]`. The Warnings crowd out the Problems the user actually ran the check for (blank SNAM, unresolvable quest/branch, dead alias indices), and several topics may not render at all.

The verdict is already carried losslessly and cheaply by the two things this PR also adds: the inline `subtype=X (stale)` / `subtype_marker=Y (authoritative)` labels in the text render, and `subtype_stale` in JSON. Suggest dropping the per-topic issue (or emitting the explanation once per report, with the per-topic signal left to the labels).

src/housecarl-core/DialogueValidate.cs
```diff
@@ -474,6 +486,24 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
...
+        {
+            var fromMarker = DialogueSubtype.NameForMarker(topic.SubtypeName);
+            issues.Add(new(DialogueIssueSeverity.Warning,
+                $"DialogTopic.Subtype reads {topic.Subtype} ((int){(int)topic.Subtype}) but the SNAM marker is "
```
Avick3110: Unlike the blank-marker lint directly above it, this one does not distinguish a topic this plugin DEFINES from an override, and the field it complains about is almost never the modder's.

The blank-SNAM block computes `isOverride` and scopes both severity and wording by it, on the reasoning that a modder should not be shouted at over a record they inherited. This new warning fires identically on both. Scenario: a modder validates a small patch that overrides three vanilla `HELO` topics to change a condition. All three warn, at the same severity as a genuine authoring fault, about a stale number that came from Skyrim.esm, that the modder did not write, and that the message itself says should not be rewritten. There is no action for them to take.

At minimum the message should say the number comes from the base record when `topic.FormKey.ModKey` != `winnerPlugin`; better, reuse the existing `isOverride` computation (hoist it above the blank-marker `if`) and suppress or downgrade the override case.

src/housecarl-core/DialogueValidate.cs
```diff
@@ -461,8 +467,14 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
...
+            // before the Dragonborn-era CK renumbered the enum (see DialogueSubtype.MarkerDisagreesWithSubtype). With
+            // SNAM blank there is nothing to cross-check it against, so the advice says where it came from and points
+            // at the base record's marker rather than asserting a number-derived tag is right.
+            var derived = $" That marker is derived from the numeric Subtype, which is stale on topics authored before "
+                        + "the Dragonborn-era Creation Kit renumbered the subtype enum — check the base record's SNAM before writing it.";
             var fix = expected is not null
-                ? $"Set it to {expected} (the marker for Subtype={topic.Subtype}); houseCARL's create tools now auto-fill it, or {ToolNames.Apply} on SubtypeName with value={expected}."
+                ? $"Set it to {expected} (the marker for Subtype={topic.Subtype}); houseCARL's create tools now auto-fill it, or {ToolNames.Apply} on SubtypeName with value={expected}.{derived}"
                 : $"Set it to the correct 4-char marker for Subtype={topic.Subtype} via {ToolNames.Apply} on SubtypeName.";
```
Avick3110: The caveat is attached only to the branch that names a marker; the fallback branch is derived from the same unreliable field and says nothing.

`derived` is appended only in the `expected is not null` arm. The `else` arm still tells the user to "Set it to the correct 4-char marker for Subtype={topic.Subtype}" — i.e. go look up a marker for the very number the PR argues cannot be trusted — with no warning that the number may be stale. Scenario: a topic with a blank SNAM and a Subtype outside 0..102 (the `Unmodeled` case `SyncMarkerToSubtype` fails loudly on) gets advice that silently assumes the number is right.

Also minor: `derived`'s first segment is an interpolated string (`$"..."`) with no holes, and it is built unconditionally even when the `else` arm is taken.

src/housecarl-mcp/DialogueSweepRender.cs
```diff
@@ -295,6 +295,9 @@ static void WriteTopicRow(Utf8JsonWriter w, TopicValidation t)
         w.WriteString("category", t.Category);
         w.WriteString("subtype", t.Subtype);
         w.WriteString("subtype_marker", t.SubtypeName);
+        // The marker is authoritative when the two disagree — the engine buckets by SNAM and the numeric subtype goes
+        // stale on pre-Dragonborn topics. Emitted always so a consumer never has to infer it from the issues text.
+        w.WriteBoolean("subtype_stale", t.SubtypeDisagreesWithMarker);
```
Avick3110, line 300: The JSON transport says the value is wrong but not what the right one is, so the two renders are not equivalent.

The text render ends with "Treat this topic's subtype as Hello, not RechargeExit" — the actionable half. JSON gets `subtype` (the stale name), `subtype_marker` (the raw 4-char tag) and `subtype_stale: true`, but no marker-derived subtype name. Scenario: a bulk-catalogue job reads the JSON to group topics by subtype; it can now tell that its `subtype` column is unreliable for a row but has to re-implement the SNAM->name table itself to fix it, which is exactly the hand-written per-type mapping `DialogueSubtype` exists to be the single home for.

`DialogueSubtype.NameForMarker` already returns it. Suggest emitting it alongside, e.g. `subtype_from_marker`, when `SubtypeDisagreesWithMarker` is true.

src/housecarl-core/DialogueSubtype.cs
```diff
@@ -166,6 +166,29 @@ static readonly (string Name, string Marker)[] Table =
...
 
+    /// <summary>Marker → subtype index, the reverse of the table. Ordinal (markers are fixed-case 4-char signatures).</summary>
+    static readonly Dictionary<string, int> ByMarker =
+        Table.Select((row, i) => (row.Marker, i)).ToDictionary(p => p.Marker, p => p.i, StringComparer.Ordinal);
+
+    /// <summary>The subtype index a 4-char SNAM marker names, or null when the marker is blank or not one this table
+    /// models. SNAM is the AUTHORITATIVE statement of a topic's subtype — the engine buckets by it — so this is the
+    /// lookup a reader should trust over <c>(int)DialogTopic.Subtype</c>.</summary>
+    public static int? IndexForMarker(RecordType marker) =>
+        !IsBlankMarker(marker) && ByMarker.TryGetValue(marker.Type, out var i) ? i : null;
```
Avick3110, line 178: A non-blank marker this table does not model now falls through both checks silently.

`IndexForMarker` returns null for an unmodeled marker, so `MarkerDisagreesWithSubtype` is false; `IsBlankMarker` is also false, so the blank-marker branch does not fire either. Scenario: a mod ships a topic with SNAM set to a garbage or invented tag (`ZZZZ`, or a mis-cased `helo` — the lookup is `StringComparer.Ordinal`). The engine buckets it to a tag no dialogue handler reads, the topic never plays, and the dialogue check reports nothing at all about the marker. That is the "silently degraded mode" the error cornerstone rules out.

Since the reverse lookup now exists, an `else if (DialogueSubtype.IndexForMarker(topic.SubtypeName) is null)` arm after the disagreement check would cover it in one line: non-blank, unmodeled, worth a Warning naming the tag.

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded, with a ruling behind it. Aaron ruled tonight that the warning stays, but is emitted only for a topic the plugin under check defines or overrides — never for a base master's own record read through the order. So the flood is gone: a whole-quest check over vanilla topics is quiet, because the winning record there comes from Skyrim.esm and there is nothing the modder can act on. The gate is `!ErrorCheck.IsBaseMaster(winnerPlugin)`, hoisted beside the existing `isOverride` above the blank-marker `if`.

The verdict still rides the labels everywhere, exactly as you suggest: `subtype=X (stale)` / `subtype_marker=Y (authoritative)` in text, `subtype_stale` (plus the new `subtype_from_marker`) in JSON, on every topic including vanilla ones.

The two stale comments are updated: `WritePatchBuilder.SyncEditedTopicMarkers` now says the sync leaves those topics alone and that the validator only warns on a mod-authored record, and `DialogueSubtype`'s header says the vanilla numbers are stale under the Dragonborn renumbering rather than vaguely unreliable.

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded. The ownership computation is hoisted above the blank-marker `if` and both SNAM findings now share it. The gate is the winner's plugin, not `isOverride` alone: a base-game master's own winning record is quiet, and a record a mod defines or overrides warns — including your three-HELO-override patch, which Aaron ruled is the case that should warn (it is the modder's own record now, and the override carries the stale number forward).

New tests drive the differential on two identically shaped topics: one in a base master nothing touches (labelled, no warning) and one the last plugin overrides (warning).

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded. The caveat is now appended to the composed `fix` string, so both arms carry it, and it is a plain `const string` with no interpolation holes built once. Wording adjusted so it reads correctly after either arm: "That comes from the numeric Subtype, which is stale on topics authored before …".

src/housecarl-mcp/DialogueSweepRender.cs
Avick3110: Folded. `TopicValidation` now carries `SubtypeFromMarker` (`DialogueSubtype.NameForMarker`, "" when the marker is blank or unmodeled) and the JSON writes `subtype_from_marker` right beside `subtype_stale`. Emitted always rather than only on a disagreement, so the row shape is stable for a bulk job. Test asserts `subtype=RechargeExit` / `subtype_from_marker=Hello` on the stale topic.

src/housecarl-core/DialogueSubtype.cs
Avick3110: Folded. An `else if (modAuthored && DialogueSubtype.IndexForMarker(topic.SubtypeName) is null)` arm after the disagreement check, under the same ownership gate, warns naming the tag: it says the marker is not one houseCARL models, that the tags are fixed case (HELO, not helo), that the topic buckets where no handler reads it, and — per the coverage cornerstone — that a real marker missing from the table should be reported rather than assumed invalid. Test drives a topic with SNAM `ZZZZ`.

src/housecarl-core/DialogueValidate.cs
```diff
@@ -474,6 +494,41 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
...
+        //     reported, never "fixed" — rewriting DATA\Subtype here would be a guess at what the author meant.
+        //     Scoped to a record a mod defines or overrides (Aaron's ruling): a base master's own topic read through
+        //     the order carries Bethesda's stale number, which the modder cannot act on and which would fill a
+        //     whole-quest check with the same warning. There the verdict still rides the render's "(stale)" /
+        //     "(authoritative)" labels and the JSON's subtype_stale / subtype_from_marker fields.
+        else if (modAuthored && DialogueSubtype.MarkerDisagreesWithSubtype(topic))
+        {
+            var fromMarker = DialogueSubtype.NameForMarker(topic.SubtypeName);
+            issues.Add(new(DialogueIssueSeverity.Warning,
+                $"DialogTopic.Subtype reads {topic.Subtype} ((int){(int)topic.Subtype}) but the SNAM marker is "
```
Avick3110: **One asserted cause for every mismatch.** This message states the Dragonborn renumbering as *the* explanation whenever the two fields disagree, but the check fires on any mismatch.

Concrete: an author opens a vanilla Hello topic in xEdit, sets `DATA/Subtype` = Custom, leaves `SNAM` = HELO. That is a real authoring error — the engine still buckets it as Hello, so the subtype edit is an in-game no-op. houseCARL replies "The record is not necessarily broken; Bethesda renumbered the numeric subtype enum … a topic authored before that stores a number six lower" and "Treat this topic's subtype as Hello, not Custom": a false cause, framed as benign vintage, and it never names the fix houseCARL's own write path already applies (`SyncEditedTopicMarkers` syncs SNAM to the new Subtype).

The renumbering signature is cheap to test: the stale case is exactly `fromMarker - (int)topic.Subtype == 6` with `fromMarker >= 26`. Any other delta is two fields edited apart, and deserves the opposite advice — "your Subtype change is a no-op; sync SNAM".

src/housecarl-core/DialogueValidate.cs
```diff
@@ -474,6 +494,41 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
...
+
+        // --- A non-blank marker this table does not model: not a disagreement (there is nothing to compare) and not
+        //     blank, so it would otherwise pass both checks in silence while the engine buckets the topic to a tag no
+        //     dialogue handler reads. Same ownership gate as the disagreement above.
+        else if (modAuthored && DialogueSubtype.IndexForMarker(topic.SubtypeName) is null)
+        {
+            issues.Add(new(DialogueIssueSeverity.Warning,
+                $"DialogTopic.SubtypeName (the SNAM subtype marker) is {topic.SubtypeName.Type}, which is not a marker houseCARL "
+                + "models — the game buckets topics by this 4-char tag, so an invented or mis-cased one (the tags are fixed case: "
+                + "HELO, not helo) puts the topic in a bucket no dialogue handler reads and it never plays. Set it to the marker for "
```
Avick3110: **This advice derives the marker from the field the same diff just declared unreliable.** Twelve lines up, the blank-marker arm gained the `derived` caveat precisely because a Subtype-derived marker is stale on pre-Dragonborn topics. This arm makes the identical derivation with no caveat.

Concrete: a mod overrides a vanilla HELO topic (carrying the pre-Dragonborn number, so `Subtype` reads `RechargeExit`) and mis-types the marker as `helo`. This branch fires and advises "Set it to the marker for the intended subtype (Subtype reads RechargeExit)" — pointing the author at `RCEX` when the topic's real bucket is `HELO`, which is the exact failure the rest of the PR exists to prevent.

src/housecarl-core/DialogueValidate.cs
```diff
@@ -474,6 +494,41 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
...
+        //     SNAM wins — the engine buckets by the marker, and xEdit marks DATA\Subtype cpIgnore for the same reason.
+        //     The usual cause is age, not damage: the Dragonborn-era CK inserted six FlyingMount* subtypes at index 20,
+        //     so a topic authored before that stores a number six lower than the modern enum and every reader (Mutagen,
+        //     xEdit, houseCARL) labels it six entries too early. No field distinguishes the two numberings, so this is
+        //     reported, never "fixed" — rewriting DATA\Subtype here would be a guess at what the author meant.
+        //     Scoped to a record a mod defines or overrides (Aaron's ruling): a base master's own topic read through
+        //     the order carries Bethesda's stale number, which the modder cannot act on and which would fill a
+        //     whole-quest check with the same warning. There the verdict still rides the render's "(stale)" /
+        //     "(authoritative)" labels and the JSON's subtype_stale / subtype_from_marker fields.
+        else if (modAuthored && DialogueSubtype.MarkerDisagreesWithSubtype(topic))
```
Avick3110: **The gate excludes vanilla's stale numbers but not the overrides that carry them forward verbatim.** `modAuthored` cannot tell an override that *changed* the subtype from one that merely inherited the base record's number.

Concrete: a follower or NPC mod overrides 30 vanilla HELO topics to add lines. Every one inherits Bethesda's pre-Dragonborn number, every winner is a non-base plugin, so all 30 warn with the same text. The comment's own justification for the base-master exclusion — content "the modder cannot act on" and a whole-quest check "filled with the same warning" — applies here unchanged: the author neither wrote the number nor can fix it, since the PR states nothing is rewritten. Generic vanilla dialogue is exactly the population that is stale *and* the one mods most often override, so the noise the gate was added to prevent reappears one step out.

Comparing the winner's `(Subtype, SubtypeName)` pair against the base record's, or restricting the warning to topics the plugin **defines**, would drop the carried-forward case while keeping the one the check is for.

src/housecarl-core/DialogueValidate.cs
```diff
@@ -457,13 +468,22 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
         //         not a confirmed crash → Warning. Don't cry "guaranteed CTD" over working content.
         //     The blank test is DialogueSubtype's, shared with the create-path auto-fill so they can't drift, and the
         //     expected marker is named so the fix is a copy-paste rather than a bare "invalid".
+        // Ownership of the record the check actually reads, shared by every SNAM finding below.
+        bool isOverride = !string.Equals(topic.FormKey.ModKey.FileName.String, winnerPlugin, StringComparison.OrdinalIgnoreCase);
+        // A base-game master's own winning record is content the modder neither wrote nor can act on.
+        bool modAuthored = !ErrorCheck.IsBaseMaster(winnerPlugin);
```
Avick3110: `ErrorCheck.IsBaseMaster` matches only the five base masters, so force-loaded content the user did not author lands on the mod-authored side of this gate.

Concrete: a load order with Creation Club content. `housecarl_check` on a CC quest whose topics override vanilla HELO topics warns on each of them, because `ccBGSSSE*.esl` is not in `BaseMasterSet`. `ErrorCheck`'s own doc comment on `BaseMasterNames` calls this out — "NOT in this set: Creation Club plugins and `_ResourcePack.esl`" — and notes houseCARL's load-order status groups those *with* the base masters as implicit, which is the distinction this gate wants.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,16 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A dialogue topic whose numeric `Subtype` contradicts its SNAM marker is now flagged, and the marker is named
+  as the authoritative one.** The engine buckets topics by SNAM; the numeric subtype is stale on topics authored
+  before the Dragonborn-era Creation Kit inserted six `FlyingMount*` values into the enum, so a vanilla `HELO`
+  topic can read as `RechargeExit`. The dialogue check now warns on the disagreement — only for a topic a mod
+  defines or overrides, so a whole-quest check over vanilla topics stays quiet — and both renders label the pair
+  (`subtype=… (stale)`, `subtype_marker=… (authoritative)`, `subtype_stale`, `subtype_from_marker`) on every
+  topic. A non-blank marker houseCARL does not model is named rather than passed over, and the blank-marker
```
Avick3110: The mod-authored qualifier is stated for the disagreement warning but not for this one, and the unmodeled-marker check is gated identically (`else if (modAuthored && ...)` at `DialogueValidate.cs:522`).

A reader takes "named rather than passed over" to mean housecarl_check names a bad SNAM wherever it appears; a base master's own topic with an unmodeled marker is still passed over in silence. The sentence claims coverage the code does not have.

src/housecarl-core/DialogueSubtype.cs
```diff
@@ -166,6 +167,29 @@ static readonly (string Name, string Marker)[] Table =
...
+
+    /// <summary>Mutagen's SubtypeEnum name for the subtype a SNAM marker names, or null when the marker is blank or
+    /// unmodeled (and "" for the one index Mutagen's enum omits). The honest label for a topic's subtype.</summary>
+    public static string? NameForMarker(RecordType marker) => IndexForMarker(marker) is { } i ? NameAt(i) : null;
```
Avick3110, line 182: `NameForMarker` returns `""` for index 3 (FVDL, the row Mutagen's enum omits) and `null` for a blank or unmodeled marker, and `SubtypeFromMarker = NameForMarker(...) ?? ""` collapses the two.

Concrete: a topic with `SNAM=FVDL` and `Subtype=Custom` renders `subtype_stale=true` beside `subtype_from_marker=""`. The field's own doc promises a consumer "never has to re-implement the marker→name table" — but in the one case the table cannot name, it hands back nothing. Emitting the marker itself (or the index) instead of `""` would close it.

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded in 2d58493. The advice now splits on the signature: `DialogueSubtype.IsRenumberedVintage(markerIndex, subtypeValue)` is `markerIndex >= 26 && markerIndex - subtypeValue == 6`, and only that case keeps the "not necessarily broken" wording. Any other gap gets the opposite advice — the topic buckets as the marker, so the Subtype value is an in-game no-op, and the fix names the sync (set Subtype through houseCARL and SyncEditedTopicMarkers does it, or apply SubtypeName yourself). Tested by the Custom-beside-HELO topic in DialogueWorld.

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded in 2d58493. The unmodeled-marker arm now recommends the base record's SNAM when the topic is an override and that marker is modeled — read through the same cached base-copy seek the inheritance gate uses, and named with the defining plugin. Only with no base copy, or a base marker as unmodeled as this one, does it fall back to the numeric Subtype, and then it carries the same "that comes from the numeric Subtype, which is stale on pre-Dragonborn topics" caveat the blank arm does.

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded in 2d58493. The disagreement warning now needs the pair to be the plugin's own: it reads the topic's copy in its defining master and stays silent when the winner's (Subtype, SubtypeName) pair equals it. A topic the plugin defines has no base copy, so it warns as before. The base copy is a typed DIAL seek, cached per FormKey and asked for only by a finding about to fire, so a clean order pays nothing. Labels and the JSON fields stay ungated. Test: AnOverrideInheritingAStaleSubtypeIsNotWarnedAbout.

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded in 2d58493. The gate is now IsBaseMaster plus the force-loaded set — Mo2Composition.ImplicitPluginNames, the same grouping housecarl_load_order_status shows as implicit — passed down from LoadOrderService (once per sweep, alongside the pinned view). When the composition cannot be read it passes null rather than an empty set, so the gate degrades to the five base masters and errs loud, which the parameter doc states. Test: a ccHcTest.esl override in DialogueWorld.

plugin/CHANGELOG.md
Avick3110: Folded in 4e4fe6a. The entry now states one scope for both warnings — force-loaded content (base masters, Creation Club, _ResourcePack.esl) and an override carrying the base pair forward are passed over — says the labels are not scoped, and describes the split advice. It matches the code after the fixes for your other two comments.

src/housecarl-core/DialogueSubtype.cs
Avick3110: Folded in 2d58493. New DialogueSubtype.LabelForMarker: Mutagen's enum name where there is one, the marker itself for index 3 (FVDL), null only for a blank or unmodeled marker. subtype_from_marker and every message now use it, so FVDL renders as FVDL and "" means only "there is nothing to name". Test: TheMarkerNamesItselfWhereMutagensEnumHasNoNameForIt.

PR #676, opened 2026-09-08
Make a new line's own quest the default dialogue route

Avick3110:
The 2026-09 skill gate ran one authoring task twice — once with `dialogue-authoring` loaded and once with no skill at all — and the skill lost (`dev/skill-validation/gate-2026-09/dialogue-authoring/VERDICT.md`). The skill's fast path told the run to append its new line into the existing vanilla topic; that topic turned out to be contested ten plugins deep, and the with-skill run's own log conceded it could not say whether the line is ever selected. The unaided control instead measured the incumbent dialogue quest's priority, built its own quest above it with its own topic, wrote the `.seq`, and overrode no vanilla record — a safer result in 10 calls against the skill's 18. Aaron's ruling (`dev/DECISIONS.md`, 2026-09-08) was one bounded rewrite, then a re-gate.

So the own-quest route is now the skill's default and appending into a vanilla topic is a separate section with the case for taking it. The default section carries the three reads that settle the design — finding a real exemplar line with `references=[<npc formid>]` rather than a dead-end `Speaker` scan, taking the exemplar's `Emotion`/`EmotionValue`/`Flags`, and measuring the incumbent quest's `Priority` — then the create call, the cloned condition gate, and `housecarl_write_seq`. Both worked examples now carry conditions and the emotion fields; the append example previously wrote neither, and the gate's only field-level defect in either arm was the with-skill run shipping `EmotionValue = 0` / `Flags = 0`. The skill also now says that the activation greeting is the `HELO` subtype and there is no `GREE`, and that a topic's `SNAM` marker is authoritative for its subtype where the `Subtype` enum disagrees across form versions (issue #660; the server-side flag is a separate PR). The Q1 output-quality eval is rewritten to the new deliverable — nine expectations, with the note that the control already reached the substance of three of them unaided, so the skill's evidence has to come from the rest and from getting there in fewer calls. The trigger cases and the description are untouched.

Build clean, 1894 tests pass, 128/128 ci-all probes pass (including `plugin-validate-guard`, which measures every shipped description against the 1024-character ceiling), and the packaging leak/pointer scan is clean over 12 skills and 227 markdown files.

**After review (2026-09-09).** The default route no longer clones its gate from a vanilla exemplar. It composes the conditions in the create call as typed nested `ConditionFloat` members (`GetIsID` on the NPC plus a condition that passes only sometimes, with a `GetRandomPercent` example), which is what the gate control run did and what verified. Cloning a vanilla gate with the copy zip is the documented option, with rows that do not belong removed by key and alias-scoped rows never carried across the quest boundary. The "never compute the operator bytes" rule from 1.3.1 is retired (DECISIONS 2026-09-09). The pre-enable sweep call is repeated in the default route, the two `Priority` fields are told apart, and the #660 rule is unconditional: read `SubtypeName` on any copy.


.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,91 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
 items is `housecarl:kid-authoring`, a record's own fields with no ESP is `housecarl:skypatcher-authoring`.
 (On a Codex install those siblings appear under their bare folder names — `spid-authoring`, and so on.)
 
-## Adding one line to an existing topic — the fast path
+## Adding a line to an existing NPC — the default route: your own quest
+
+For a new line on an NPC who already has lines, **do not append into the vanilla topic.** Author your
+own start-game-enabled `QUST` at a priority above the incumbent dialogue quest's, give it its own DIAL
+topic of the same subtype, hang your INFO under that, and write the `.seq`. Nothing vanilla is
```
Avick3110: **The default route can silence the very greetings the task says not to break.** The section asserts that because nothing vanilla is touched, `nothing can be re-listed, reordered or broken`, and in the same breath that quest priority is what decides which topic the game enters. Those two claims pull against each other: a start-game-enabled quest at Priority 65 owning a HELO topic whose one INFO passes its speaker gate outranks `DialogueWhiterun` at 30, so at every activation the engine reaches *your* topic first and Belethor's six vanilla greetings stop cycling. "Nothing vanilla is touched" is true at the data layer and does not carry to in-game behaviour â€” which is exactly the distinction the "Order, and the reorder trap" section already insists on ("How the engine picks among eligible generic greetings is not something this skill or houseCARL can tell you"). Either say the preemption risk out loud and gate the new line's conditions so it does not pass on every activation, or drop the unqualified "nothing can be broken".

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,91 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+               { "field_path": "Flags",    "value": "StartGameEnabled" },
+               { "field_path": "Priority", "value": "65" } ] },
+
+    { "record_type": "DialogTopic", "editorid": "MyMod_BelethorGreetTopic",
+      "ops": [ { "field_path": "Quest",    "value": "@MyMod_BelethorGreetQuest" },
+               { "field_path": "Subtype",  "value": "Hello" },
+               { "field_path": "Priority", "value": "50" },
+               { "field_path": "Category", "value": "Misc" } ] },
+
+    { "record_type": "DialogResponses", "editorid": "MyMod_BelethorGreetInfo",
```
Avick3110, line 88: **The default route's INFO sets no `Speaker`, and the skill says elsewhere that is unrecoverable.** Lines 146-149 (the append section) state: set `Speaker` *in the create call* if you want the `.fuz` path printed, because "with no speaker the voice type comes from the quest alias at runtime and cannot be computed, and setting it afterwards does not backfill, because voice and result-script coverage are checked on create". The new default example has no `Speaker` op â€” and the quest it creates has no aliases either, so there is no runtime fallback: the run gets no computed voice path and no voice coverage on the write, with no second chance. This is also the defect the previous revision's E6 expectation existed to catch, and E6 was dropped from the eval in this same PR. Add the `Speaker` op to the default example.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,91 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+      "parent": "MyMod_BelethorGreetTopic",
+      "ops": [ { "field_path": "Responses", "op": "Add",
+                 "compose": { "type": "DialogResponse",
+                              "fields": { "Text": "Browse as you like.", "ResponseNumber": "1",
+                                          "Emotion": "Neutral", "EmotionValue": "50",
+                                          "Flags": "UseEmotionAnimation" } } } ] }
+  ])
+```
+
+Then **clone the gate** onto the new line from the vanilla exemplar you read in step 2 — never
```
Avick3110: **Cloning a vanilla exemplar's `Conditions` across quest boundaries is not the safe move it was inside the same topic.** In the append route, source and target sit under the same quest, so a copied `GetStage`/`GetIsAliasRef`/Run-On-reference-alias entry still resolves. Under the new default route the target lives in a brand-new quest with no stages and no aliases, and the copy is wholesale (`references/write-side-recipes.md`: "The copy **replaces** the target's `Conditions` with the source's â€” `CopyFrom` is the only verb the zip issues"). A vanilla dialogue-quest greeting gated on its own quest's alias index therefore lands on your INFO pointed at an alias your quest does not have, and the line is permanently silent with nothing in the data layer flagging it. The section should say to read the exemplar's gate and keep only quest-independent entries (a `GetIsID`-style speaker check), not just "copy only the entries that belong".

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,91 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+hand-synthesize the operator bytes (Recipe A in `references/write-side-recipes.md`). One call:
+
+```json
+housecarl_apply(into="<the filename the create call reported>", readback=true,
+                bundle=["Conditions"],
+                assignments=[{ "target": "<the new INFO>", "from": "02848F:Skyrim.esm" }])
+```
+
+Copy only the entries that belong: a vanilla greeting's gate is often a speaker check *plus* a cell
```
Avick3110: **"Copy only the entries that belong" is not something the shown call can do.** The example is introduced as "One call", but `bundle=["Conditions"]` + `assignments` is a wholesale `CopyFrom`: `references/write-side-recipes.md` says the copy "**replaces** the target's `Conditions` with the source's â€” ... there is no merge variant". So a reader who follows the call literally ends up with the cell check they were told to leave out, and nothing tells them a second `housecarl_apply` (a `Remove` on the unwanted `Conditions` entry) plus a readback is required. The same gap is in the append section at lines 133-136 ("drop only the entries you positively do not want"), and eval expectation E5 now grades on the run *stating which entries it kept and dropped* â€” a step the skill never shows how to perform.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,91 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+check, and a line meant to play anywhere wants the speaker check alone.
+
+Then the `.seq`, which is not optional: `housecarl_write_seq(source="<patch>.esp")`. A
```
Avick3110: **`source="<patch>.esp"` uses the stem, which this same skill says is the wrong name to use.** `patch=` is a base name, auto-suffixed when the stem is taken â€” the skill states this thirty lines down and tells the reader to sweep "the filename the create call reported back, never the stem you passed", and the `apply` call just above here correctly says `<the filename the create call reported>`. If `MyGreeting.esp` already existed, the create lands in `MyGreeting02.esp` while `housecarl_write_seq(source="MyGreeting.esp")` resolves the *older* plugin (the resolver searches every mod folder, enabled or not) and writes a `.seq` for it â€” the new quest gets none and never starts, and the write reports success. Use the reported filename here too.

.claude/skills/dialogue-authoring/evals/evals.json
```diff
@@ -288,7 +290,7 @@
...
+   "prior_evidence": "The 2026-09 gate ran this task on the live 3,801-plugin order with a 40-call cap, once with no skill and once with the old skill, and the skill LOST: the control reached the own-quest route unaided in 10 calls, while the old skill's fast path put the with-skill run into a ten-deep contested topic in 18 calls and shipped a response row at EmotionValue 0 / Flags 0. The verdict is dev/skill-validation/gate-2026-09/dialogue-authoring/VERDICT.md. E1-E7 are what the rewrite must now add over the CONTROL, not only over the previous with-skill run -- the control already passed the substance of E1, E2 and E7 with no skill at all, so the skill's evidence has to come from reaching the same design in fewer calls and from E3-E6, which the control did not all get."
   }
  ],
  "known_gap": "A dialogue check that can see an off-order plugin is issue #615 and is not in 2.0. E4 therefore tests that the run names the gap and uses the off-order errors/scripts sweep, not that the dialogue family covers the new plugin.",
```
Avick3110: **Stale expectation id: `known_gap` still points at E4.** The off-order-check expectation was E4 before this PR and is now **E8**; E4 is the emotion-fields expectation. As written the file-level `known_gap` says the #615 gap is tested by "E4 ... the run names the gap and uses the off-order errors/scripts sweep", which describes E8. A grader reading the gap note is pointed at the wrong expectation. No other case in this file has an E4 about off-order coverage.

.claude/skills/dialogue-authoring/evals/evals.json
```diff
@@ -288,7 +290,7 @@
...
     "duration_ms"
    ],
-   "prior_evidence": "The 2026-09 audit ran this task on the live 3,801-plugin order with a 40-call cap, once with no skill and once with the old skill. E1-E4 and E7 passed with the skill and E1/E4 failed without it; E5, E6 and E7 are what this rewrite adds -- E5 and E6 failed in BOTH prior runs, so the rewrite must beat its own previous with-skill run, not only the control."
+   "prior_evidence": "The 2026-09 gate ran this task on the live 3,801-plugin order with a 40-call cap, once with no skill and once with the old skill, and the skill LOST: the control reached the own-quest route unaided in 10 calls, while the old skill's fast path put the with-skill run into a ten-deep contested topic in 18 calls and shipped a response row at EmotionValue 0 / Flags 0. The verdict is dev/skill-validation/gate-2026-09/dialogue-authoring/VERDICT.md. E1-E7 are what the rewrite must now add over the CONTROL, not only over the previous with-skill run -- the control already passed the substance of E1, E2 and E7 with no skill at all, so the skill's evidence has to come from reaching the same design in fewer calls and from E3-E6, which the control did not all get."
```
Avick3110: **`prior_evidence` contradicts itself on which expectations the rewrite has to win.** It says "E1-E7 are what the rewrite must now add over the CONTROL" and then, in the same sentence, "the control already passed the substance of E1, E2 and E7 with no skill at all". Those cannot both hold; the range also silently excludes E8 and E9, which are part of this case. Say "E3-E6 are what the rewrite must add over the control; E1, E2 and E7 the control already reached unaided" â€” which is what the PR body actually claims.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,91 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+   onto your row — a response left at `EmotionValue = 0` and `Flags = 0` diverges from every vanilla
+   line around it.
+3. **Read the incumbent quest** — the `*parent.Quest` from step 1 — and take its `Priority`. Yours
+   goes comfortably above it. Vanilla `DialogueWhiterun` is `Priority = 30`, so 65 clears it. Do not
+   guess the number; measure it.
+
+**Skyrim's activation greeting is the `HELO` subtype** — `Subtype = Hello`, `SubtypeName = HELO`.
+There is no `GREE` subtype; do not go looking for one. **A DIAL's `SNAM` marker (`SubtypeName`) is
+authoritative for its subtype, and the `Subtype` enum can disagree with it on a record read from a
+different form version** — the same topic read from `Skyrim.esm` and from a later-form-version override
```
Avick3110: **The #660 sentence points the reader at the wrong half of the discrepancy.** As written, "the `Subtype` enum can disagree with it on a record read from a different form version" makes form version the cause and leaves it undecidable which of the two reads is the bad one. The measured fact (VERDICT.md, #660) is directional: `02707A` read from **Skyrim.esm's own copy** prints `Subtype = RechargeExit`, while the FormVersion-44 winner prints `Subtype = Hello`. The numeric subtype on the master's pre-Dragonborn copy is the stale one — Dragonborn inserted enum members, so old numbering decodes six positions off (`Recharge`/`GBYE` is the second reproducer).

This matters here specifically, because steps 1–2 immediately above send the reader to read a vanilla exemplar **out of Skyrim.esm**, which is exactly the read that prints the wrong name. Two lines earlier the section also asserts flatly "`Subtype = Hello`, `SubtypeName = HELO`" — which is not what that read shows.

Suggest: say the master's own copy is the one to distrust, not "a different form version".

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,91 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+      "ops": [ { "field_path": "Responses", "op": "Add",
+                 "compose": { "type": "DialogResponse",
+                              "fields": { "Text": "Browse as you like.", "ResponseNumber": "1",
+                                          "Emotion": "Neutral", "EmotionValue": "50",
+                                          "Flags": "UseEmotionAnimation" } } } ] }
+  ])
+```
+
+Then **clone the gate** onto the new line from the vanilla exemplar you read in step 2 — never
+hand-synthesize the operator bytes (Recipe A in `references/write-side-recipes.md`). One call:
```
Avick3110: **The absolute "never hand-synthesize" rule is what makes the default route cost more calls than the control it has to beat.** The gate control composed the same gate inline in the create call — `Conditions` Add compose `ConditionFloat` {ComparisonValue 1, EqualTo} with nested `Data` compose `GetIsIDConditionData` {Object `013BA1:Skyrim.esm`, RunOnType Subject} — and the judge read it back correct. VERDICT.md is explicit: "the control hand-composed the same gate and it verifies correct, so the safety rule bought no measured correctness here."

As written the route is create → apply(CopyFrom) → a third apply to drop the cell entry the clone drags in, where the control did it in one create. The skill lost this gate on 18 calls vs 10 and is being re-gated on the same axis, so mandating an extra one or two calls for a gate that verifiably composes correctly inline is spending the exact currency the re-gate measures.

Recipe A's rationale is "N lines" and a gate you cannot otherwise verify — that reasoning does not carry to a single new line with a one-condition `GetIsID` gate. Suggest scoping the rule: clone when the exemplar's gate is non-trivial or you are copying it onto several lines; a single `GetIsID` speaker check composes inline in the create and saves the round trip.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in e08cb3a8. The section now says the preemption plainly: no vanilla record is touched, so nothing in the vanilla topic is re-listed or reordered, but whenever your gate passes the engine enters your topic and the vanilla greetings do not play for that activation — and the old lines return the moment the gate fails. It closes by telling the reader to gate the line so it passes only when meant to, since a gate that passes on every activation replaces the whole greeting pool.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in e08cb3a8. The default create now carries a Speaker op (013BA1:Skyrim.esm), with a following sentence giving the reason the append section already gives: the new quest has no aliases for the voice type to come from at runtime, voice and result-script coverage are checked on create, and setting Speaker later does not backfill the .fuz path.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in e08cb3a8. The .seq call now reads housecarl_write_seq(source="<the filename the create call reported>"), matching the apply just above, with a clause saying the reported filename and never the stem, or the .seq lands on an older plugin of that name and the new quest gets none.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in e08cb3a8. The passage now says the copy to distrust is the master own: records authored before the Dragonborn CK, which inserted enum members, decode six positions off — 02707A out of Skyrim.esm prints Subtype = RechargeExit while its later winner prints Hello, for one unchanged HELO marker, with 0904AC / GBYE as the second reproducer — and it says outright that steps 1-2 send the reader into exactly that read. The flat "Subtype = Hello" two lines up is now phrased as what is written, not what the master read prints. The CHANGELOG line was corrected the same way.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Held for Aaron ruling on the condition route — no edit yet. He is deciding between the nested ConditionFloat compose in the create call as the default and the CopyFrom clone; the cross-quest resolution point lands with that ruling.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Held for Aaron ruling on the condition route — no edit yet. Whether the skill shows a second apply to drop entries, or composes the gate inline in the create instead, depends on that ruling.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Held for Aaron ruling on the condition route — no edit yet. This is the ruling itself: inline compose as the default versus the CopyFrom clone. He decides tonight or tomorrow and the fold lands after.

.claude/skills/dialogue-authoring/evals/evals.json
Avick3110: Folded in 6f197c65. known_gap now cites E8.

.claude/skills/dialogue-authoring/evals/evals.json
Avick3110: Folded in 6f197c65. prior_evidence now reads: the bar is the control, which already reached the substance of E1, E2 and E7 unaided, so E3-E6, E8 and E9 are what the rewrite must add over it, plus reaching the same design in fewer calls.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,105 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+                                          "Emotion": "Neutral", "EmotionValue": "50",
+                                          "Flags": "UseEmotionAnimation" } } } ] }
+  ])
+```
+
+`Speaker` goes in that create call, not after it: your new quest has no aliases for the voice type to
+come from at runtime, and voice and result-script coverage are checked on create, so setting it later
+does not backfill the `.fuz` path.
+
+Then **clone the gate** onto the new line from the vanilla exemplar you read in step 2 — never
```
Avick3110: **Cloned conditions that are alias-scoped die under the new quest, and nothing here says so.**

The exemplar found in step 1 belongs to `DialogueWhiterun`. A condition row with `RunOnType = QuestAlias`, or a `GetIsAliasRef`, resolves its index against the *owning* quest — and `MyMod_BelethorGreetQuest` above has no `Aliases` at all. This skill's own `references/condition-functions.md` states the rule twice: "a `QuestAlias` Run On whose index isn't a real alias of the owning quest is a dead condition", and for `GetIsAliasRef`, "Index is quest-relative — wrong index = dead".

So cloning the whole gate across a quest boundary can ship a line that never plays (or, if the dead row was the only speaker check, one that plays for everyone), with a byte-perfect readback and no check that can catch it. The append route did not have this failure mode, because the INFO stayed under the exemplar's own quest — making the own-quest route the default is what introduces it.

Worth a sentence next to "Copy only the entries that belong": a row scoped to a quest alias cannot come along, because the new quest has none.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,105 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+
+For a new line on an NPC who already has lines, **do not append into the vanilla topic.** Author your
+own start-game-enabled `QUST` at a priority above the incumbent dialogue quest's, give it its own DIAL
+topic of the same subtype, hang your INFO under that, and write the `.seq`. No vanilla record is
+touched, so nothing already in the vanilla topic is re-listed or reordered. Be clear about what that
+does not buy: the higher quest priority is what gets your topic reached at all — **which topic the
+game enters is decided across quests by priority**, before intra-topic order is even consulted — so
+whenever your line's gate passes, the engine enters *your* topic and the vanilla greetings do not
+play for that activation. That is preemption, not damage: the vanilla records are untouched and the
+old lines return the moment your gate fails. Gate the line so it passes only when you mean it to; a
```
Avick3110: **The worked example below is the case this sentence forbids.**

"Gate the line so it passes only when you mean it to; a gate that passes on every activation replaces the NPC's whole greeting pool." But the route as written produces exactly that gate: a quest at Priority 65 over `DialogueWhiterun`'s 30, a `HELO` topic, and a gate cloned from the vanilla exemplar with the cell check dropped (line 107) — i.e. a speaker check alone, which passes on every activation of Belethor. By the preemption paragraph directly above, the engine then enters the new topic every time and none of the six vanilla greetings ever play.

That is the answer to the Q1 prompt ("another greeting … **without breaking his existing greetings**") reading as the safe default. The skill gives no example, condition or heuristic for a gate that passes only *sometimes* — a `GetRandomPercent` row, a stage check, a once-only flag. Without one, "gate it so it passes only when you mean it to" has nothing behind it.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,105 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+Then **clone the gate** onto the new line from the vanilla exemplar you read in step 2 — never
+hand-synthesize the operator bytes (Recipe A in `references/write-side-recipes.md`). One call:
+
+```json
+housecarl_apply(into="<the filename the create call reported>", readback=true,
+                bundle=["Conditions"],
+                assignments=[{ "target": "<the new INFO>", "from": "02848F:Skyrim.esm" }])
+```
+
+Copy only the entries that belong: a vanilla greeting's gate is often a speaker check *plus* a cell
```
Avick3110: **"Copy only the entries that belong" has no call behind it.**

The zip above is the only mechanism shown, and `references/write-side-recipes.md` says what it does: "The copy **replaces** the target's `Conditions` with the source's — `CopyFrom` is the only verb the zip issues, and there is no merge variant." So a reader who runs it gets the speaker check *and* the shop-cell check, and nothing in SKILL.md or any reference shows the follow-up that drops a row (`op:"Remove"` with `key:"<index>"` on `Conditions`, per `housecarl_apply`'s own `key=` parameter).

The reader is then left with either the cell check shipped — a `Hello` line that only fires inside Belethor's shop, which is precisely what "a line meant to play anywhere" rules out — or hand-editing the list, which the same paragraph forbids. `evals.json` E5 scores the run on "which cloned entries it kept and which it dropped", so the skill is graded on a step it never teaches.

Same gap at line 149 in the append section.

.claude/skills/dialogue-authoring/evals/evals.json
```diff
@@ -269,15 +269,17 @@
...
+    "E3 -- the vanilla exemplar INFO is found with references=[<Belethor's FormID>], not with a where= scan on Speaker. New: the old skill said 'read a real vanilla line' without saying how, and the previous with-skill run burned a call on the Speaker dead end",
+    "E4 -- the response row carries the exemplar's Emotion / EmotionValue / Flags (Neutral / 50 / UseEmotionAnimation), not the 0 / none defaults. New: the previous with-skill run shipped 0 and none, the only field-level defect the gate judge found in either arm",
+    "E5 -- the line's Conditions are CLONED from a vanilla sibling INFO rather than hand-synthesized, and the run states which cloned entries it kept and which it dropped (the shop-cell check on a line meant to play anywhere)",
+    "E6 -- housecarl_write_seq is run against the patch, and the run says the quest and its dialogue do not exist without it",
+    "E7 -- no vanilla record is overridden: the delivered patch contains only records the run created, and the run says that is what makes 'existing greetings unbroken' true by construction rather than by measurement",
```
Avick3110: **E7 now rewards a claim the skill body says is false.**

The second commit added, to SKILL.md: "whenever your line's gate passes, the engine enters *your* topic and the vanilla greetings do not play for that activation", and "a gate that passes on every activation replaces the NPC's whole greeting pool." A run that reads and obeys that will report the greetings as **preempted** — and gets marked down here for not saying "existing greetings unbroken" is "true by construction".

The commit changed SKILL.md but not this file, so the suite now scores the skill's correct output as a failure. E7's first clause ("the delivered patch contains only records the run created") is still right; it's the "true by construction" half that needs to become something like *the run states that no vanilla record is touched **and** that the new topic preempts the vanilla greetings whenever its gate passes.*

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,105 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+   onto your row — a response left at `EmotionValue = 0` and `Flags = 0` diverges from every vanilla
+   line around it.
+3. **Read the incumbent quest** — the `*parent.Quest` from step 1 — and take its `Priority`. Yours
+   goes comfortably above it. Vanilla `DialogueWhiterun` is `Priority = 30`, so 65 clears it. Do not
+   guess the number; measure it.
+
+**Skyrim's activation greeting is the `HELO` subtype** — `SubtypeName = HELO`, written as `Subtype = "Hello"`.
+There is no `GREE` subtype; do not go looking for one. **A DIAL's `SNAM` marker (`SubtypeName`) is
+authoritative for its subtype, and the numeric `Subtype` enum can disagree with it — and the copy to
+distrust is the master's own** (issue #660). Records authored before the Dragonborn CK, which inserted
```
Avick3110: **This states a cause the repo's own authority contradicts, and #660 doesn't yet claim.**

`src/housecarl-core/DialogueSubtype.cs:16-19` gives a different explanation: the mapping "cannot be scraped from vanilla because **Bethesda's own DATA\Subtype numbers are unreliable** (many HELO topics ship with DATA≠79) — which is why xEdit ignores DATA\Subtype and treats SNAM as the required master field." Under that reading the two copies in #660 are two different physical records carrying different DATA bytes, not one record decoded two ways, and *any* plugin's copy can be noisy — not just the master's.

Issue #660 itself hedges ("The enum decode **appears to** be applied without regard to the record's form version") and asks for the cause to be established. This paragraph promotes that to a mechanism ("the Dragonborn CK, which inserted enum members", "decode six positions off") and then narrows the distrust: "the copy to distrust is the master's own." A reader who takes that literally will trust `Subtype` when reading a non-master override — e.g. a modded HELO topic whose author left DATA at its shipped value — and get the wrong subtype, which is the error this section exists to prevent.

The practical rule (read `SubtypeName`) is right either way. Suggest keeping the rule unconditional — distrust `Subtype` on *any* copy — and either dropping the causal story or marking it as the open hypothesis #660 records it as.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,105 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+                assignments=[{ "target": "<the new INFO>", "from": "02848F:Skyrim.esm" }])
+```
+
+Copy only the entries that belong: a vanilla greeting's gate is often a speaker check *plus* a cell
+check, and a line meant to play anywhere wants the speaker check alone.
+
+Then the `.seq`, which is not optional:
+`housecarl_write_seq(source="<the filename the create call reported>")` — the reported filename, never
+the stem you passed, or the `.seq` lands on an older plugin of that name and your quest gets none. A
+start-game-enabled quest with no `.seq` never starts, and its dialogue never exists. Finish with the
```
Avick3110: **The default route never shows the sweep call it's graded on.**

"Finish with the pre-enable sweep below" points into *The exception — appending to an existing vanilla topic*, whose paragraph opens "Then sweep the patch before it is enabled" as a continuation of the append example. The actual call — `housecarl_check(plugins=["<the filename the create call reported>"], findings=["errors","scripts"])` — appears only there.

`evals.json` E8 scores the run on making that call *from the skill*, before the tool's own refusal is hit. As written, the one graded call of the default route is only reachable by reading the section this page just told the reader not to take. Repeating the call in the default route (three lines) would close it.

.claude/skills/dialogue-authoring/SKILL.md
```diff
@@ -24,11 +24,105 @@ Distribution is somebody else's lane: a form onto NPCs is `housecarl:spid-author
...
+  records=[
+    { "record_type": "Quest", "editorid": "MyMod_BelethorGreetQuest",
+      "ops": [ { "field_path": "Name",     "value": "My Belethor Greeting" },
+               { "field_path": "Flags",    "value": "StartGameEnabled" },
+               { "field_path": "Priority", "value": "65" } ] },
+
+    { "record_type": "DialogTopic", "editorid": "MyMod_BelethorGreetTopic",
+      "ops": [ { "field_path": "Quest",    "value": "@MyMod_BelethorGreetQuest" },
+               { "field_path": "Subtype",  "value": "Hello" },
+               { "field_path": "Priority", "value": "50" },
```
Avick3110, line 85: Two different `Priority` fields sit side by side here with no distinction — `Quest.Priority` (a byte, 65) and `DialogTopic.Priority` (a float, 50) — immediately after step 3's "Do not guess the number; measure it."

Only the 65 is the measured one; the topic's 50 is never mentioned in prose. A reader who measures a different incumbent — say a quest already at Priority 60 — has no way to tell from the example which field carries their number, and putting it on the DIAL leaves the quest at its default and the topic unreachable: the exact failure this section is built to avoid. One clause on the topic line ("the DIAL's own priority; not the number you measured") would settle it.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in bc8b3c5.

The default route no longer clones. It composes its gate in the create call as typed `ConditionFloat` rows with a nested `Data` compose — `GetIsID(Belethor) = 1` plus `GetRandomPercent < 25` — and the section says in prose why the second row is there: a gate that passes on every activation replaces the NPC's whole greeting pool, because the higher-priority topic is reached first every time. A following paragraph gives the non-chance alternatives (a `GetStage`/`GetStageDone` check against your own quest, and the once-only shape: a result script sets the stage, a `GetStage < N` row stops the line firing again), and notes what they have in common — they read *your* quest, which exists.

Cloning is now a subsection below, "The option — cloning a vanilla gate instead of composing one", for the case your comment scopes it to: several rows, or a gate you cannot otherwise verify.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in bc8b3c5.

Both places now show the call. Under the clone option: "**The copy replaces, it does not merge.**" followed by

```json
housecarl_apply(ops=[{ "formid": "<the new INFO>", "field_path": "Conditions",
                       "op": "Remove", "key": "0" }])
```

with the note that `key` is the list index, so drop the highest index first — removing index 0 renumbers everything after it — then read back and say which rows you kept and which you dropped. The same call is repeated in the append section, and `references/write-side-recipes.md` Recipe A gains a step 4 carrying it.

E5 no longer grades a step the skill does not teach: it now accepts either composition route, and requires the kept/dropped statement plus the `op:"Remove"` drop only of a run that clones.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in bc8b3c5.

The clone option now carries it as its own caveat: "**A row scoped to a quest alias cannot come along.**" — `RunOnType = QuestAlias` resolves its index against the *owning* quest, and so does `GetIsAliasRef`, and the new quest has no aliases at all. It states the two failure shapes you name (the line never plays, or plays for everyone if the dead row was the only speaker check), with a byte-perfect readback and no check that can catch it, and says the append route does not have this failure mode because the INFO stays under the exemplar's own quest.

The route this lands on is also no longer the default: the default composes its gate inline, so the caveat scopes to the clone option where the quest boundary is actually crossed.

.claude/skills/dialogue-authoring/evals/evals.json
Avick3110: Folded in 1eed171.

E7 now reads: no vanilla record is overridden and the delivered patch contains only records the run created, AND the run states both halves — that no vanilla record is touched, so nothing in the vanilla topic is re-listed or reordered, and that the new topic PREEMPTS the vanilla greetings whenever its gate passes, the old lines returning the moment it fails. A run that reports untouched-at-the-data-layer AND preempted-in-play passes; a run that calls "existing greetings unbroken" true by construction with no preemption stated fails. `expected_output` picks up the same two halves.

On the second half of your point: E5 was also grading a step the skill did not teach. It now accepts either composition route — a `ConditionFloat` with a nested `Data` compose in the create call, or a clone off a vanilla sibling — requires the gate not to pass on every activation (speaker check plus a row that passes only sometimes), and asks for the kept/dropped statement and the `op:"Remove"` drop only from a run that clones. Both routes are shown in SKILL.md as of bc8b3c5.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in be72e6e.

The rule is now unconditional: the numeric `Subtype` can disagree with `SNAM` "on any copy — read `SubtypeName`, never `Subtype`, whatever plugin the record came from". The narrowing to the master is gone, and the paragraph now states `DialogueSubtype.cs`'s reason first — Bethesda's own numeric subtypes are unreliable, which is why xEdit ignores the number and treats `SNAM` as the required field.

The `02707A` / `0904AC` pair stays as the measured reproducer, but demoted to that: "Why the two disagree is still open there — pre-Dragonborn enum numbering is the hypothesis #660 records, not an established rule, so do not narrow the distrust to the master." The "decode six positions off" mechanism is dropped.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in 61ce13f.

The default route now carries the call itself, right after the `.seq` paragraph, rather than pointing into the append section:

```json
housecarl_check(plugins=["<the filename the create call reported>"], findings=["errors","scripts"])
```

with the off-order note, the `patch=` base-name/auto-suffix warning, and the `#615` limit to state when reporting. The append section keeps its own copy, so neither route depends on reading the other.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Folded in 181ef4f.

A paragraph directly under the create call now separates them: "**The two `Priority` fields above are different fields.** `Quest.Priority` is the byte you measured in step 3 — 65, above `DialogueWhiterun`'s 30 — and it is what gets your topic reached at all. `DialogTopic.Priority` is the DIAL's own float, carried from the exemplar topic at 50; it is not the number you measured, and putting your measured number there leaves the quest at its default and the topic unreachable."

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Held until the condition route was settled; the ruling landed 2026-09-09 and this is folded in bc8b3c5.

The ruling: the default own-quest route composes its conditions in the create call as typed nested `ConditionFloat` members — the server encodes the bytes — and cloning a vanilla gate with the copy zip is the option, not the mandate. So the cross-quest clone your comment flags is no longer the default path.

Where it does apply, the clone option now carries the caveat explicitly: "**A row scoped to a quest alias cannot come along.**" — `RunOnType = QuestAlias` and `GetIsAliasRef` resolve their index against the owning quest, the new quest has no aliases at all, and the result is a line that never plays or, if the dead row was the only speaker check, plays for everyone, with a byte-perfect readback and no check that can catch it. It says to drop those rows or compose the gate inline, and notes the append route does not have this failure mode.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Held until the condition route was settled; the ruling landed 2026-09-09 and this is folded in bc8b3c5.

The ruling scopes the rule the way you suggested. The default own-quest route composes its gate inline in the create call as typed nested `ConditionFloat` members — the shape the gate control ran and the judge read back correct — so the single-line `GetIsID` case costs no extra round trip. Cloning with the copy zip is the option, kept for the case Recipe A's rationale actually covers: a gate several rows deep, one you cannot otherwise verify, or the same gate onto many lines.

The absolute "never hand-synthesize the operator bytes" rule and its 26-broken-conditions story are removed from SKILL.md, from Recipe A's intro in `references/write-side-recipes.md`, and from the write note at the end of `references/condition-functions.md`. What is left is the accurate statement: compose typed rows and the server encodes the bytes; nothing is retyped as an encoded value either way.

.claude/skills/dialogue-authoring/SKILL.md
Avick3110: Held until the condition route was settled; the ruling landed 2026-09-09 and this is folded in bc8b3c5.

The default route no longer clones at all — it composes its gate inline in the create call — so "copy only the entries that belong" is gone from it. Where the clone option remains, the wholesale behaviour is stated ("**The copy replaces, it does not merge.**") and the follow-up call is shown:

```json
housecarl_apply(ops=[{ "formid": "<the new INFO>", "field_path": "Conditions",
                       "op": "Remove", "key": "0" }])
```

with `key` explained as the list index and the instruction to drop the highest index first, then read back and say which rows were kept and which dropped. The same call is repeated in the append section, and Recipe A in `references/write-side-recipes.md` gains a step 4 for it.

E5 is retuned to match: it accepts either composition route and asks for the kept/dropped statement and the `Remove` call only from a run that clones.

PR #678, opened 2026-09-08
Say what a no-value read was, and which half of a hybrid FormID to drop

### LINKED ISSUE #662: [Bug] Null and not-applicable fields render as "unreadable"
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/dialogue-authoring/VERDICT.md`, `dev/skill-validation/gate-2026-09/spid-authoring/VERDICT.md`).

`Speaker` renders as unreadable on 31,242 of 31,465 INFO records, and `Configuration.Level.LevelMult` on 65,892 of 66,855 NPC_ records. In both cases the field is simply null or not the active union arm. "Unreadable" claims a read failure that did not happen; an agent counting unreadable fields as errors gets the wrong picture. Render a null as null (or absent) and reserve the unreadable wording for an actual read fault.

### LINKED ISSUE #670: [GAP] The hybrid FormKey refusal does not say which half to drop
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/npc-appearance-copy/VERDICT.md`); both arms hit it.

`000A2C94:Skyrim.esm` (an 8-digit runtime-style id with a plugin suffix) is refused, but the sentence does not say that the leading two digits are the load-order index and the plugin form wants six digits (`0A2C94:Skyrim.esm`). Both arms spent a call working it out. Name the form to use in the refusal.

### LINKED ISSUE #671: [GAP] project={form:"fields"} emits fields as a list of {path,value} objects, not a map
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/kid-authoring/VERDICT.md`).

The fields form renders each record's fields as a list of `{path, value}` pairs, so a consumer indexes by scanning for the path instead of keying on it; every local join in the gate re-shaped it first. A map keyed by path is the shape every consumer built. If the list is deliberate (ordering, duplicate paths under `[*]`), say so in the parameter description.

Avick3110:
Three small render fixes found by the 2026-09-08 skill gate.

A `where=` predicate that read no value on more than half the scanned records said "had no readable value", which reads as a read fault whatever the cause: a null link (`Speaker` on 31,242 INFO records) and the other arm of a union (`Configuration.Level.LevelMult` on 65,892 NPC records) are neither faults nor errors, and an agent counting them as such gets the wrong picture of a healthy scan. The note now names each cause from the counters the scan already keeps — unset, not a field on the record read, a container, a read fault — so "unreadable" is reserved for a real parse failure; the distinction comes off those counters, not off the rendered note. A FormID that mixes the two notations (`000A2C94:Skyrim.esm` — eight runtime digits with the plugin name still on it) was refused as "Malformed FormKey string", which says nothing about which half to drop; both doors that parse a FormID now refuse it with the six-digit form to write (`0A2C94:Skyrim.esm`) and the runtime form on its own, with a light token's local id taken off the ESL window rather than the middle six digits. And `project={form:"fields"}` on json stays an ordered list of `{path, value}` — the order is the answer's and a path repeats under a quantified step, which a map could not hold — so the `format` parameter now says so and points at `dense` for the column table.

Tests: the accounting causes get a unit test per cause; the hybrid gets parser tests plus one read that must come back naming the six-digit form. Both fail before the change. Full suite 1904 passed, ci-all 128/128.

Closes #662
Closes #670
Closes #671


src/housecarl-core/FieldPredicate.cs
```diff
@@ -1316,12 +1316,27 @@ static bool TryFormKey(string s, out FormKey fk)
...
+    /// four different answers, and one word for all four ("unreadable") claims a read failure that did not happen.
+    /// The classes are counted in code as the scan runs, never re-derived from a rendered note.</summary>
+    string NoValueBreakdown(int k)
+    {
+        long unset = _noValue[k] - _noField[k] - _container[k] - _unreadable[k];
+        var parts = new List<string>();
+        if (unset > 0) parts.Add($"unset — null or absent ({unset:N0})");
+        if (_noField[k] > 0) parts.Add($"not a field on the record read ({_noField[k]:N0})");
+        if (_container[k] > 0) parts.Add($"a container/list, not a scalar ({_container[k]:N0})");
+        if (_unreadable[k] > 0) parts.Add($"a read fault — Mutagen could not parse the field ({_unreadable[k]:N0})");
```
Avick3110, line 1448: **medium** — the new "read fault" wording asserts a parse failure that a link-step scan did not actually have.

`JudgeTargets` (FieldPredicate.cs:1095) ends with:

```csharp
return (false, anyNoField ? (anyListHop ? EvalKind.ListHop : EvalKind.NoField) : EvalKind.Unreadable);
```

So a link-step predicate returns `EvalKind.Unreadable` whenever no target produced a Definite verdict and none produced a NoField — which includes the two most ordinary cases: every reached target read the field as **Unset**, and every reached target was **unresolvable** (`target is null; continue`, e.g. the link points into a plugin that is not in the order).

Concretely: `where="Items[*].Item->ObjectEffect = 0ABCDE:X.esp"` over a container scan where the linked items simply carry no enchantment now renders `a read fault — Mutagen could not parse the field (N)`. Nothing faulted. That is exactly the mislabel this PR set out to remove, just moved one hop down the path — and it is worse than the old wording, because the old note was vague where this one is definite.

Either classify the all-Unset / all-unresolvable link-step fallback as `Unset` (with an unresolvable-target counter of its own if you want it named), or have `NoValueBreakdown` word the `_unreadable` bucket so it does not claim a Mutagen parse failure it cannot distinguish from an unresolvable link target.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -1316,12 +1316,27 @@ static bool TryFormKey(string s, out FormKey fk)
             }
             else if (_noValue[k] * 2 > _scanned)
                 (notes ??= new()).Add(
-                    $"note: '{path}' had no readable value on {_noValue[k]:N0} of {_scanned:N0} scanned record(s) " +
-                    $"(absent or not a field on those types) — counted as non-matches there, not errors.");
+                    $"note: '{path}' had no value on {_noValue[k]:N0} of {_scanned:N0} scanned record(s) — " +
```
Avick3110, line 1429: **low** — the soft note now names a read fault and, in the same sentence, says it was "counted as non-matches there, not errors."

When `_unreadable[k] > 0` the rendered line reads `… a read fault — Mutagen could not parse the field (12) — counted as non-matches there, not errors.` A read fault *is* the one cause in the list where the filter genuinely could not judge the record, so calling it a non-match and not an error is the silently-degraded answer the design rules out — the loud branch at line 1303 says so explicitly ("This is a coverage/parse limit … the filter can't judge these records").

Also note the gate above (`_noValue[k] * 2 > _scanned`) means a scan with, say, 40 read faults out of 66,000 records emits no note at all — the faults are counted as non-matches with nothing said. Naming the causes makes that gap more visible, not less; consider a separate trigger on `_unreadable[k] > 0` regardless of the half-threshold.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -1316,12 +1316,27 @@ static bool TryFormKey(string s, out FormKey fk)
...
+                    NoValueBreakdown(k) + " — counted as non-matches there, not errors.");
```
Avick3110: **low** — stale comment left behind by the rename. FieldPredicate.cs:1254-1256 still reads:

> All four keep the loud marker "yielded no readable value on any" (distinct from the SOFT "had no readable value on" for a >half-but-not-all miss)

The soft marker is now `had no value on`. A future reader (or a probe author) grepping for the string the comment names will not find it.

src/housecarl-core/RuntimeFormId.cs
```diff
@@ -43,6 +43,26 @@ public static bool TryParse(string? text, out uint value)
         return uint.TryParse(s, NumberStyles.AllowHexSpecifier, CultureInfo.InvariantCulture, out value);
     }
 
+    /// <summary>The sentence for a HYBRID token — eight runtime digits AND a plugin suffix
+    /// (<c>000A2C94:Skyrim.esm</c>), which is neither notation. It names the six-digit plugin form to write
+    /// instead, so the caller is not left to work out which half to drop; null for anything that is not a hybrid.
+    /// A light token's local id comes off the ESL window, so <c>FE012800:X.esl</c> names <c>000800:X.esl</c>, not
+    /// the middle six digits.</summary>
+    public static string? HybridNote(string? text)
```
Avick3110, line 57: **medium** — `HybridNote` is wired into the two FormID doors, but a hybrid arriving as a `where=` **operand** is still swallowed silently, which is the same class of bug as #670.

`FieldPredicateSet.Parse`'s `parseFormId` callback is only reached from `ParseFormIdList`, i.e. `formid in [...]` / `not in [...]` (FieldPredicate.cs:457). Every other FormID-valued operand — `where="ObjectEffect = 000A2C94:Skyrim.esm"`, `where="Race in [000A2C94:Skyrim.esm]"` — is compared by `ValueEquals` → `TryFormKey`, which swallows the `Malformed FormKey string` in a `try/catch` and falls through to a case-insensitive **string** compare against the rendered token `0A2C94:Skyrim.esm`. Result: `0 matches`, no refusal, no accounting note (values *were* read, so the predicate reports a healthy scan).

That is a silently wrong answer for the exact token this PR is about. Running the operand through the hybrid check at parse time — or validating any operand containing `:` that fails `FormKey.Factory` — would close it.

src/housecarl-mcp/FormIdDoor.cs
```diff
@@ -42,6 +42,8 @@ internal sealed class FormIdDoor
     /// sentence on anything it cannot answer, and on a runtime FormID at a <see cref="ForWrite"/> door.</summary>
     public FormKey Parse(string? raw)
     {
+        // A runtime form with the plugin name still on it is neither notation: say which half to drop.
+        if (RuntimeFormId.HybridNote(raw) is { } hybrid) throw new FormatException(hybrid);
```
Avick3110, line 46: **medium** — `Parse` now refuses a hybrid, but `RuntimeRefusal` (line 60) does not, so `create`'s `parent=` still gets the old opaque message.

```csharp
public string? RuntimeRefusal(string? raw)
{
    if (!RuntimeFormId.TryParse(raw, out _)) return null;   // a hybrid has a ':' → TryParse false → null
```

`000A2C94:Skyrim.esm` contains a colon, so `TryParse` is false and `RuntimeRefusal` returns null before `Parse` is ever called. The only caller (LoadOrderService.cs:7403) then leaves the token "to the core's own parse", which reaches `FormKey.Factory` and throws `Malformed FormKey string` — the message this PR replaced everywhere else. Adding the same `HybridNote` check at the top of `RuntimeRefusal` makes the third door consistent with the two you fixed.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -132,7 +132,7 @@ public static string Records(
...
+        [Description("TRANSPORT: 'text' (default) | 'json' (machine-readable document; same accounting in-band) | 'dense' (scan lane: positional columnar cells 1:1 with the requested fields — the compact bulk-enumeration form; by that definition depth expansion and the 'everything' form are inexpressible in it). On 'json' a record's fields are an ORDERED LIST of {path, value} — the emission order is the answer's own, and a path REPEATS under a quantified step ('Effects[*].Data.Magnitude'), which a map keyed by path could not hold; 'dense' is the column table to join on. Every response carries the epoch stamp — the identity of the index build it was answered from — spelled epoch=<hex> on 'text' and 'dense', and as an 'epoch' member on 'json'.")]
```
Avick3110: **low** — the new sentence describes the json field shape as `{path, value}`, but a field with no value emits `note` instead of `value`.

JsonWire.cs:248-249:

```csharp
w.WriteString("path", f.Path);
if (f.HasValue) w.WriteString("value", f.Token); else w.WriteString("note", f.Note);
```

Since this description *is* the contract an agent codes against, an agent told the list is `{path, value}` will read `f.value` on an unset field and get `undefined` — silently, and precisely on the records the accounting note is telling it to be careful about. Worth one clause: `{path, value}` (or `{path, note}` where the field read no value).

src/housecarl-core/LoadOrderResolver.cs
```diff
@@ -1,4 +1,4 @@
-using System.Text;
+﻿using System.Text;
```
Avick3110: **nit** — this hunk adds a UTF-8 BOM to the file (same in `FormIdDoor.cs` and `RuntimeFormIdTests.cs`), which looks like editor collateral rather than an intended change: only 4 of the 463 tracked `.cs` files under `src/` carry a BOM today, so this triples that against the prevailing convention and puts three no-op line-1 diffs in the blame.

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded. JudgeTargets now carries the loudest no-verdict class an actually-read target produced, so the all-unset and all-unresolvable fallbacks come back as Unset and only a target that really faulted reports a read fault. Two tests in WhereAccountingCauseTests cover both (targets resolve but read unset; targets do not resolve at all).

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded. The soft note now also fires whenever _unreadable[k] > 0, whatever the ratio, and the tail is written per branch: with no read faults it still says the records counted as non-matches, and with read faults it says the filter could NOT judge those records (a coverage/parse limit) and names how many.

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded. The comment now names the current soft marker, 'had no value on'.

src/housecarl-core/RuntimeFormId.cs
Avick3110: Folded. Every where= operand now goes through the hybrid door at parse: the scalar operand, each entry of a non-formid in/not-in list, and each formid-list token (checked before the door runs, so the sentence is the same with no load order in hand). A hybrid refuses the call instead of string-comparing to a quiet zero; theory test in RuntimeFormIdTests covers all three shapes.

src/housecarl-mcp/FormIdDoor.cs
Avick3110: Folded. RuntimeRefusal checks HybridNote before TryParse, so create's parent= gives the six-digit sentence rather than 'Malformed FormKey string'. Test: CreateRefusesAHybridParentAndNamesTheSixDigitForm.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. The sentence now says a field that read no value carries {path, note} instead, saying why. The json field shape is also in the changelog now, since it is a user-visible contract.

src/housecarl-core/LoadOrderResolver.cs
Avick3110: Folded. The three BOMs are gone (verified with od -c); RecordsTools.cs keeps the one it already had.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -1076,23 +1089,28 @@ EvalKind ClassifyMiss(string note)
...
             {
                 anyVerdict = true;
                 if (sat) return (true, EvalKind.Definite);
             }
-            else if (kind is EvalKind.NoField or EvalKind.ListHop or EvalKind.NoParent) { anyNoField = true; anyListHop |= kind == EvalKind.ListHop; }
+            else if (unjudged is null || NoVerdictRank(kind) > NoVerdictRank(unjudged.Value)) unjudged = kind;
         }
         if (anyVerdict) return (false, EvalKind.Definite);
-        return (false, anyNoField ? (anyListHop ? EvalKind.ListHop : EvalKind.NoField) : EvalKind.Unreadable);
+        return (false, unjudged ?? EvalKind.Unset);
```
Avick3110, line 1172: **An unresolvable link target is not "unset" either.** Dropping the `Unreadable` fallback is right, but `Unset` claims something about the record that is false.

Scan `where="ObjectEffect->Name contains Fire"` over weapons whose enchantments live in a plugin that is disabled, or whose master is missing. `_fetchWinnerBody` returns null for every target, so `anyVerdict` stays false and `unjudged` stays null, and this line returns `Unset`. Every candidate then lands in `Matches`' `default:` bucket (`_noValue` only), so `AccountingNote` takes the `_noField == 0 && _container == 0 && _unreadable == 0` arm and prints:

> the field IS VALID; it is simply UNSET (absent/null) on every one, so the path reads fine and there are just no values in this scope. Widen the scope…

The links are present and non-null. Nothing was read because the targets do not resolve — a broken-master/disabled-plugin condition the caller now cannot see, and "widen the scope" is the wrong next move for it. `ALinkStepWhoseTargetsDoNotResolveIsNotNamedAReadFault` pins the half that is fixed; the other half is that this is a fifth cause, not the unset one. A separate class + counter (unresolved target) would let the note name it, the same way the other four are named.

src/housecarl-core/RuntimeFormId.cs
```diff
@@ -43,6 +43,26 @@ public static bool TryParse(string? text, out uint value)
...
+        var s = text?.Trim();
+        if (string.IsNullOrEmpty(s)) return null;
+        int c = s.IndexOf(':');
+        if (c <= 0 || c == s.Length - 1) return null;
+        var plugin = s[(c + 1)..];
+        if (!TryParse(s[..c], out uint v)) return null;
+        uint local = FormIdRange.LocalObjectId(v);
+        return $"'{s}' mixes the two FormID forms: eight digits is the RUNTIME form the game, the console and the " +
+               $"logs print — its leading digits are the load-order index, not part of the record's id — and a " +
+               $"plugin-qualified FormID takes SIX. Write '{local:X6}:{plugin}', or drop the plugin name and pass " +
```
Avick3110: **The two forms this sentence offers are not interchangeable, and the second one can silently read a different record.**

xEdit shows a FormID as eight digits carrying *its* session's load index. A user copies `0A012345` from xEdit and keeps the plugin name: `0A012345:MyMod.esp`. The note hands back two options —

- `012345:MyMod.esp` — order-independent, always the record meant. Correct.
- `0A012345` on its own — resolved against houseCARL's **current** order. If index `0A` is a different plugin there (xEdit was opened with some plugins unchecked, or the order changed since the paste), `RuntimeToFormKey` returns a perfectly valid FormKey for *another plugin's* record `012345` and the read succeeds with no error.

That is a silently wrong answer for the very token being asked about — the failure mode this PR is closing elsewhere. `HybridNote` is pure text with no index tables, so it cannot tell the two apart; the alternative belongs at a door that holds the order (`FormIdDoor` / `IndexView.ParseFormId`), offered only when the resolved plugin actually matches the named one. Dropping the second clause here would also be safe: the six-digit form alone answers the question.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -465,6 +465,11 @@ public static (FieldPredicateSet? Set, string? Error) Parse(IReadOnlyList<string
         if (pseudo == PseudoPath.FormId)
             return (null, $"predicate '{raw}': 'formid' takes the membership ops only — \"formid in <list>\" / \"formid not in <list>\" (a single record is \"formid in [XXXXXX:Plugin.esp]\").");
 
+        // A scalar operand that mixes the two FormID notations is refused HERE, at the same door the formid list
+        // parses through: it would otherwise fall past ValueEquals's FormKey attempt into a plain string compare
+        // and report a healthy scan with 0 matches — a silently wrong answer for the very token being asked about.
+        if (HybridRefusal(raw, operand) is { } hybrid) return (null, hybrid);
```
Avick3110, line 480: **The comment above this line describes the bare runtime form too, and that one still falls through.**

`HybridRefusal` returns null for a token with no colon, so `where="ObjectEffect = 000A2C94"` is stored as-is. In `ValueEquals` the leaf token is `0A2C94:Skyrim.esm`, and the operand `000A2C94`:

- `TryFormKey(operand)` fails — no `:Plugin` tail,
- `TryNum(operand)` fails — hex letters,
- so it lands on the case-insensitive string compare and is never equal.

Every record gets a DEFINITE false verdict, so `_valueRead == _scanned` and `_noValue == 0`, and `AccountingNote` emits **nothing at all** — a clean-looking scan reporting 0 matches for a record that does match. That is strictly quieter than the hybrid case this guard closes, and pasting a bare console/log FormID is the more likely paste of the two.

`Parse` already takes `parseFormId`, so the same door can resolve it (or refuse it by name when no order is in hand) right here.

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded in 89336bef. `UnresolvedTarget` is a fifth `EvalKind` with its own `_unresolved` counter, ranked above unset and below container. `EvalLinkStep` raises it when `_fetchWinnerBody` returns null instead of skipping the target; `Matches` counts it; `AccountingNote` has an all-unresolved arm saying the link IS SET on every one but none of its targets are in this order, sending the caller at the plugin and its masters and saying widening the scope will not help. The mixed arm and `NoValueBreakdown` name it too, and `UnsetCount` now subtracts it so the unset remainder stays honest. Test: `ALinkStepWhoseTargetsDoNotResolveIsNamedUnresolved_NotUnset` in WhereAccountingCauseTests.

src/housecarl-core/RuntimeFormId.cs
Avick3110: Folded in d33addc9. The clause is gone: `HybridNote` now offers the six-digit form only. Its doc says why, and that a door holding the order may add the runtime form once it has checked the resolved plugin is the named one. Test: `TheHybridNoteDoesNotOfferTheBareRuntimeFormAsAnAlternative`.

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded in 728630c6. A bare eight-hex operand on `=`/`!=` over a real field path is now resolved through `parseFormId` at parse, and refused by name when the call has no order to resolve it against. I did not rewrite the operand string, though: an eight-digit token is not unambiguously a FormID (`BasicStats.Value = 00001234` reads as a number), so the resolved key rides alongside as `Predicate.RuntimeKey` and the `=`/`!=` compare uses it only when the LEAF parses as a FormKey. A numeric or text leaf keeps its own vocabulary. Identity terms (`editorid`, `winner`) are excluded — they are text terms. Tests: `ABareRuntimeFormIdInAWhereOperandComparesAsAFormKey`, `...IsRefusedWithNoLoadOrder`, `AnEightDigitOperandOnANumericLeafStillComparesAsANumber`.

PR #679, opened 2026-09-08
Type-check a FormLink target at apply pre-flight

### LINKED ISSUE #663: [Bug] housecarl_apply pre-flight does not type-check a FormLink target
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/mutagen-reference/VERDICT.md`, control2 log call 9).

One dry run set keyword `06BBD2:Skyrim.esm` into `Race`, `ObjectEffect`, `TemplateArmor`, `EquipmentType`, `PickUpSound`, `PutDownSound`, `BashImpactDataSet` and `AlternateBlockMaterial` on an ARMO, and added a Race FormID to `Keywords`; all ten ops reported "passed resolve + pre-flight". The token shape is checked (call 7 shows a malformed FormID refused), the target type is not. This is a silently wrong write: the plugin is valid on disk and wrong in game.

The generated schema carries the target type on every link field (`references/records.jsonl` `target`), so the check is available by construction: refuse an op whose resolved record type is not in the link's target set, in one sentence naming the field, the type given and the types allowed.

Avick3110:
housecarl_apply's pre-flight proved a link value parsed as a FormID and said nothing about what it pointed at, so setting an armor's `Race`, `ObjectEffect` or `TemplateArmor` to a keyword passed and wrote a plugin that is valid on disk and wrong in game. The generator already stamps every formlink field with its Mutagen link target interface, so the check needed no new data: the rulebook takes an optional lookup from the write path (which holds the call's one captured view), resolves each link value to the record it points at, and refuses when that record's runtime type does not satisfy the field's target. The refusal names the field, the type given and the types allowed, and the allowed names are the corpus record types that satisfy the same interface, so nothing about them is hand-written. Where more types qualify than fit in a sentence it names the kind and the count instead. A field whose link accepts any record, a null-clear, and a FormID no plugin in the order defines are all left alone — the last because the order cannot say what it is, which is the dangling-reference check's business. So is a Remove: taking a link out never needs the target to be legal, and a list another mod wrote a bad link into is exactly what a caller needs to be able to repair.

The lookup rides on a per-call copy of the rulebook rather than a parameter at every hop, so all three value slots see it: a singular Set, a link list's elements, and a composed struct's field. All three write lanes pass it — the patch lane, in place, and create.

Resolving the targets is a prefetch, not a seek per link: each lane harvests the FormID-shaped tokens out of its own value slots, groups them by winner plugin, and collects them in one walk of each plugin, so a 200-entry leveled-list ReplaceAll into Skyrim.esm reads that plugin once. A plugin the write does not otherwise touch that cannot be opened is one sentence naming it, not a throw.

One existing probe assertion changed: apply-guard added a weapon FormID to a Keywords list on the note that "any FormID resolvable as a link value is enough", which is the bug; it now uses the fixture's own keyword, which was already there for the other link arms.

Ten tests against the built server: the two reported shapes refuse, a composed struct's own link field refuses, a create's link refuses and so does a literal FormID mixed beside a same-call `@editorid` sibling, a Remove by a wrong-typed value is not refused for its type, a link to its declared type still lands, a FormList's any-record Items is not refused, and neither an unresolvable target nor a null-clear is type-checked.

Closes #663

### DIFF
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -47,10 +47,20 @@ public sealed class CorpusRulebook
     readonly Corpus _corpus;
     /// <summary>Non-null only on a validate call the write path handed a load order to resolve link targets against.</summary>
     readonly LinkTargetLookup? _linkTargets;
-    /// <summary>Memo for the printed legal-type list, per link target type; only a refusal fills it.</summary>
+    /// <summary>Memo for the printed legal-type list, per link target type; only a refusal fills it. Real because a
+    /// write lane derives ONE link-target rulebook per call (<see cref="WithLinkTargets"/>) and validates every edit
+    /// through it, so a 200-op apply that names the same wrong target 200 times scans the corpus once. Unlocked: the
+    /// derived rulebook is a per-call object used on the calling thread, and the shared schema-only rulebook never
+    /// fills this (a refusal needs <see cref="_linkTargets"/>).</summary>
     readonly Dictionary<string, string> _linkTargetNames = new(StringComparer.Ordinal);
     CorpusRulebook(Corpus corpus, LinkTargetLookup? linkTargets = null)
         => (_corpus, _linkTargets) = (corpus, linkTargets);
+
+    /// <summary>This rulebook plus a load-order link-target resolver: the same corpus, with the FormLink TARGET TYPE
+    /// check turned on. Derived ONCE per write call, not per edit — the lookup rides on the rulebook rather than
+    /// through the recursion, so every value slot (a singular Set, a list element, a composed struct's field) sees it
+    /// without a parameter at each hop, and the memo above spans the whole call.</summary>
+    public CorpusRulebook WithLinkTargets(LinkTargetLookup linkTargets) => new(_corpus, linkTargets);
 
     /// <summary>Resolves a FormLink value (a FormID token) to the runtime type of the record it points at, or null
     /// when nothing in the load order carries it. Supplied by the write path, which holds the captured view; without
@@ -142,15 +152,11 @@ public sealed class CorpusRulebook
     /// resolves post-allocation. The set threads into composed StructSpec Fields/Sets too. Null (the override/set_field
     /// path) ⇒ an <c>@editorid</c> value is rejected loud — it has no meaning when there are no same-call
     /// creations.
-    /// <para><paramref name="linkTargets"/> (the write path, which has the load order) turns on the FormLink TARGET
-    /// TYPE check: a link whose FormID resolves to a record the field cannot point at is refused. It is carried on a
-    /// per-call copy of the rulebook rather than threaded through the recursion, so every value slot — a singular
-    /// Set, a list element, a composed struct's field — sees it without a parameter at each hop.</para></summary>
-    public string? Validate(WriteRequest req, IReadOnlyCollection<string>? siblingEditorIds = null,
-        LinkTargetLookup? linkTargets = null)
+    /// <para>The FormLink TARGET TYPE check — a link whose FormID resolves to a record the field cannot point at is
+    /// refused — runs only on a rulebook derived by <see cref="WithLinkTargets"/>, which the write path (the one with
+    /// a load order) builds once per call.</para></summary>
+    public string? Validate(WriteRequest req, IReadOnlyCollection<string>? siblingEditorIds = null)
     {
-        if (linkTargets is not null && !ReferenceEquals(linkTargets, _linkTargets))
-            return new CorpusRulebook(_corpus, linkTargets).Validate(req, siblingEditorIds);
         // (1) resolve the record, then validate rooted at it. ValidateFromType is shared with StructSpec validation
         // (a build-from-parts spec's nested writes) so the record path and the composition path can never disagree.
         var recType = Type(req.RecordType);
@@ -1145,26 +1151,23 @@ public sealed class CorpusRulebook
     /// or non-record link).</summary>
     string AllowedLinkTypes(FieldSchema leaf, Type target, string aq)
     {
-        lock (_linkTargetNames)
+        if (_linkTargetNames.TryGetValue(aq, out var cached)) return cached;
+        var names = new List<string>();
+        foreach (var ts in _corpus.Types.Values)
+            if (ts.Kind == "record" && WriteEngine.ResolveType(ts.GetterInterfaceAssemblyQualified) is { } gi
+                && target.IsAssignableFrom(gi))
+                names.Add(ts.Name);
+        names.Sort(StringComparer.Ordinal);
+        var bare = RecordNaming.StripInterfaceToConcrete(leaf.FormLinkTarget ?? target.Name);
+        const int cap = 12;
+        var phrase = names.Count switch
         {
-            if (_linkTargetNames.TryGetValue(aq, out var cached)) return cached;
-            var names = new List<string>();
-            foreach (var ts in _corpus.Types.Values)
-                if (ts.Kind == "record" && WriteEngine.ResolveType(ts.GetterInterfaceAssemblyQualified) is { } gi
-                    && target.IsAssignableFrom(gi))
-                    names.Add(ts.Name);
-            names.Sort(StringComparer.Ordinal);
-            var bare = RecordNaming.StripInterfaceToConcrete(leaf.FormLinkTarget ?? target.Name);
-            const int cap = 12;
-            var phrase = names.Count switch
-            {
-                0 => bare,
-                1 => names[0],
-                _ when names.Count <= cap => "one of: " + string.Join(", ", names),
-                _ => $"any {bare} record ({names.Count} record types qualify)",
-            };
-            return _linkTargetNames[aq] = phrase;
-        }
+            0 => bare,
+            1 => names[0],
+            _ when names.Count <= cap => "one of: " + string.Join(", ", names),
+            _ => $"any {bare} record ({names.Count} record types qualify)",
+        };
+        return _linkTargetNames[aq] = phrase;
     }
 
     /// <summary>The loud per-element rejection for a malformed FormLink collection ELEMENT — the SAME legal-set copy
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -45,7 +45,17 @@ public sealed class StructSpec
 public sealed class CorpusRulebook
 {
     readonly Corpus _corpus;
-    CorpusRulebook(Corpus corpus) => _corpus = corpus;
+    /// <summary>Non-null only on a validate call the write path handed a load order to resolve link targets against.</summary>
+    readonly LinkTargetLookup? _linkTargets;
+    /// <summary>Memo for the printed legal-type list, per link target type; only a refusal fills it.</summary>
+    readonly Dictionary<string, string> _linkTargetNames = new(StringComparer.Ordinal);
+    CorpusRulebook(Corpus corpus, LinkTargetLookup? linkTargets = null)
+        => (_corpus, _linkTargets) = (corpus, linkTargets);
+
+    /// <summary>Resolves a FormLink value (a FormID token) to the runtime type of the record it points at, or null
+    /// when nothing in the load order carries it. Supplied by the write path, which holds the captured view; without
+    /// one the schema-only rulebook cannot know what a FormID points at, so the link TYPE check does not run.</summary>
+    public delegate Type? LinkTargetLookup(string formIdToken);
 
     /// <summary>The legal shapes of a condition FormLinkOrIndex target value — shared by the nested-sets
     /// (<see cref="ValidateFromType"/>) and the flat-fields (<see cref="CheckValue"/>) rejects so the two compose entry
@@ -131,9 +141,16 @@ public sealed class CorpusRulebook
     /// quest): a FormLink value of the form <c>@editorid</c> in that set is accepted as a forward-ref the create path
     /// resolves post-allocation. The set threads into composed StructSpec Fields/Sets too. Null (the override/set_field
     /// path) ⇒ an <c>@editorid</c> value is rejected loud — it has no meaning when there are no same-call
-    /// creations.</summary>
-    public string? Validate(WriteRequest req, IReadOnlyCollection<string>? siblingEditorIds = null)
+    /// creations.
+    /// <para><paramref name="linkTargets"/> (the write path, which has the load order) turns on the FormLink TARGET
+    /// TYPE check: a link whose FormID resolves to a record the field cannot point at is refused. It is carried on a
+    /// per-call copy of the rulebook rather than threaded through the recursion, so every value slot — a singular
+    /// Set, a list element, a composed struct's field — sees it without a parameter at each hop.</para></summary>
+    public string? Validate(WriteRequest req, IReadOnlyCollection<string>? siblingEditorIds = null,
+        LinkTargetLookup? linkTargets = null)
     {
+        if (linkTargets is not null && !ReferenceEquals(linkTargets, _linkTargets))
+            return new CorpusRulebook(_corpus, linkTargets).Validate(req, siblingEditorIds);
         // (1) resolve the record, then validate rooted at it. ValidateFromType is shared with StructSpec validation
         // (a build-from-parts spec's nested writes) so the record path and the composition path can never disagree.
         var recType = Type(req.RecordType);
@@ -680,6 +697,7 @@ public sealed class CorpusRulebook
                                "declare it before the record that references it (in spec order).";
                 }
                 else if (!WriteEngine.IsValidFormLinkValue(v)) return FormLinkElementReject(v, leaf);
+                else if (LinkTypeRefusal(leaf, v, "element") is { } mixedTypeErr) return mixedTypeErr;
             }
             return null;
         }
@@ -765,7 +783,8 @@ public sealed class CorpusRulebook
                 // apply. A null-synonym clears the link; otherwise it must parse as a FormKey. The recognizer is
                 // SHARED with the engine apply path (no drift).
                 if (leaf.Cardinality == "formlink")
-                    return WriteEngine.IsValidFormLinkValue(req.Value) ? null
+                    return WriteEngine.IsValidFormLinkValue(req.Value)
+                        ? LinkTypeRefusal(leaf, req.Value, "target")
                         : $"Illegal FormLink target '{req.Value}' for '{leaf.Name}': expected a FormID " +
                           "(XXXXXX:Plugin.esp) or a null-clear ('0', '00000000', 'Null', '000000:Null').";
                 return CoercibilityReject(leaf);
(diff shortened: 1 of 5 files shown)


src/housecarl-core/CorpusRulebook.cs
```diff
@@ -825,6 +844,12 @@ static string WritabilityRejection(TypeSchema owner, FieldSchema leaf)
                 if (!WriteEngine.IsValidFormLinkValue(v)) return FormLinkElementReject(v, leaf);
             foreach (var kv in req.Entries ?? new())
                 if (!WriteEngine.IsValidFormLinkValue(kv.Value)) return FormLinkElementReject(kv.Value, leaf);
+            // …and the TYPE of every element whose shape just passed, by the same slot-faithful sweep.
+            if (LinkTypeRefusal(leaf, req.Value, "element") is { } elemTypeErr) return elemTypeErr;
```
Avick3110: The `req.Value` arm of this sweep is not verb-scoped, so it also fires on a **Remove-by-value** — and that turns the gate against the very data it exists to prevent. Scenario: an Armor's `Keywords` already carries a stray FormID pointing at a Spell (written by another mod, or by houseCARL before this PR). `housecarl_apply` with `op=Remove, field_path=Keywords, value=<that FormID>` is now refused with "Illegal FormLink element ... that record is a Spell, but 'Keywords' links to Keyword" — the one call that would repair the list is blocked, and the message reads as if the caller were adding it.

The non-formlink twin immediately below is careful about exactly this (`req.Verb is "Add" or "SetAtIndex" or "InsertAtIndex" || (Remove && Key is null)`); the type sweep should be scoped to the verbs that INSERT a value, i.e. skip `Remove` entirely — removing a link never needs the target to be legal.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -674,6 +675,24 @@ static bool TryOffOrderCopyBody(
...
+    /// type-check that link rather than refusing on a guess.</summary>
+    static CorpusRulebook.LinkTargetLookup LinkTypeLookup(LoadOrderResolver.IndexView view, LoadOrderResolver.OverlaySession session)
+    {
+        var memo = new Dictionary<string, Type?>(StringComparer.OrdinalIgnoreCase);
+        return token =>
+        {
+            if (memo.TryGetValue(token, out var hit)) return hit;
+            Type? type = null;
+            if (FormKey.TryFactory(token, out var fk) && view.ResolveWinner(fk) is { } w)
+                type = view.GetRecord(session, w.WinnerPlugin, fk)?.GetType();
```
Avick3110: `GetRecord` with no `getterType` is `SeekBody`'s **flat enumeration of the whole plugin** until the FormKey matches — the cost #251 named ("GetRecord costs one walk per record, so N of them cost N walks"). Pre-flight now pays one such walk per distinct link token, on top of the walk it already pays for the edit target.

Concrete: a `ReplaceAll` on a LeveledItem's `Entries` with 200 references into Skyrim.esm now enumerates Skyrim.esm's ~1M major records 200 times before a single byte is written; the memo only dedupes repeats of the same token. Same for a bulk apply that sets a link on many records.

Two cheaper shapes exist and are already in the resolver: group the tokens by `w.WinnerPlugin` and use `view.CollectRecords` (one walk per plugin however many keys are wanted), or pass the field's own target interface as `getterType` for the typed group seek. Either keeps the gate off the hot path.

Secondary: this opens plugins the write otherwise never touches. `SeekBody` -> `session.Overlay(idx)` on a plugin held open by another program throws, and nothing in this loop catches it — a write that names one link into a locked plugin now fails on a plugin it does not edit.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -680,6 +697,7 @@ static string WritabilityRejection(TypeSchema owner, FieldSchema leaf)
                                "declare it before the record that references it (in spec order).";
                 }
                 else if (!WriteEngine.IsValidFormLinkValue(v)) return FormLinkElementReject(v, leaf);
+                else if (LinkTypeRefusal(leaf, v, "element") is { } mixedTypeErr) return mixedTypeErr;
```
Avick3110, line 708: This call is unreachable. The branch it sits in is entered only when some element is an `@editorid` sibling ref, and it returns above when `siblingEditorIds is null` — so reaching the loop implies the CREATE lane, which is the one lane that deliberately passes no `linkTargets`. `_linkTargets` is therefore always null here and `LinkTypeRefusal` always returns null. Harmless, but it reads as coverage the create lane does not have; either drop it or say in the comment that it is there for the day create starts passing a lookup.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1079,6 +1106,65 @@ bool IsValueCoercibleElement(FieldSchema leaf)
...
+                1 => names[0],
+                _ when names.Count <= cap => "one of: " + string.Join(", ", names),
+                _ => $"one of {names.Count} types: " + string.Join(", ", names.Take(cap)) + ", …",
+            };
+            return _linkTargetNames[aq] = phrase;
+        }
+    }
+
+    /// <summary>A Mutagen getter interface name as the record name it stands for: IRaceGetter -> Race.</summary>
+    static string BareTargetName(string interfaceName)
```
Avick3110: `RecordNaming.StripGetterInterface` already does this (`INpcGetter -> Npc`), and `RecordNaming.ClassName` covers the bare-`IFoo` case this one folds in. Two copies of the Loqui naming convention can drift — the existing one strips only when BOTH the leading `I` and the trailing `Getter` are present, this one strips the `I` unconditionally, so `IItem` prints as `Item` here and as `IItem` there. Reuse the shared helper.

src/housecarl-mcp-tests/FormLinkTargetTypeTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// <summary>A FormID no plugin in the order defines is not type-checked: the order cannot say what it is, and a
+    /// link to a record that is not there is the dangling-reference check's business, not this gate's.</summary>
+    [Fact]
+    public void AnUnresolvableTargetIsNotTypeChecked()
+        => Served(Apply(Fid(W.Armor), "Race", "Set", "ABCDEF:" + W.MasterName), "Set Race");
+
+    /// <summary>A null-clear points at nothing, so it is not a wrong type.</summary>
+    [Fact]
+    public void ANullClearIsNotTypeChecked()
```
Avick3110, line 71: The PR body calls out three value slots — a singular Set, a link list's elements, and a composed struct's field (a leveled-list entry's `Reference`) — but only the first two have a test. The composed arm (CorpusRulebook.cs:1076, in `ValidateFromType`) is the one reached by a different code path than the other two and the one most likely to regress silently; add a case that composes a LeveledItem entry with a non-item `Reference` and expects the refusal.

Avick3110:
One more, lane-level rather than line-level (it has no diff line to hang on).

**The create lane still writes a wrong-typed link silently.** `WritePatchBuilder.cs:3200` (`CreateRecordsCore`) is the third `Validate` call site and it passes no `linkTargets`, so `housecarl_create` with `Race` set to a Keyword FormID is accepted exactly as `apply` was before this PR — the same silently-wrong write #663 describes, in the lane that authors brand-new records. The ingredients are already in scope: that method captures its own view at line 2856 and has the session, so it is the same one-argument wiring as the two apply lanes, and doing it would also make the `@editorid` arm's new `LinkTypeRefusal` (CorpusRulebook.cs:700) reachable instead of dead. If it is deliberately out of scope, the PR body and the CHANGELOG entry should say the gate is apply-only — as written ("`housecarl_apply` now refuses…") a reader can reasonably read create as covered too.

For completeness, two lanes I checked and think are fine: `WriteEngine.cs:161` is the CLI direct-edit path with no load order to resolve against, and the copy lane (`CopyFrom`, `housecarl_copy`) supplies no caller-authored link value — it moves a field between the same field on the same record type, so its links are legal by construction.

Also minor, on `AllowedLinkTypes`: the >12 arm prints `one of 47 types: A, B, …` — a count plus a truncated alphabetical sample, with no way to see the rest. For a wide target that is a refusal the caller cannot act on from the message alone; naming just the target interface (as the 0-name arm does) or pointing at the schema would be more use than 12 arbitrary names.

src/housecarl-core/CorpusRulebook.cs
Avick3110: Real, folded (38de52f3). The sweep is now scoped to the verbs that put a value in — `Add`/`SetAtIndex`/`InsertAtIndex` for `req.Value`, `ReplaceAll` for `req.Values`, `Merge`/`ReplaceAll` for `req.Entries` — mirroring the non-formlink twin below it. Remove is exempt at every slot, and the comment says why: the list may already carry the wrong-typed link, and refusing the Remove blocks the repair.

Test in `FormLinkTargetTypeTests.RemovingALinkByValueIsNotTypeChecked`: a `Remove` of a Spell FormID from an Armor's `Keywords` fails on the list ("nothing to remove" — the fixture armor has none) and never mentions the type.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Real on both counts, folded (89c1f080).

There is no typed FormID→type index to resolve off: `IndexSnapshot.Index` is `FormKey → (winner, count)` and nothing else, and `records` answers `types=` with `WinnerRecordsOfType`, which is itself a walk. So the type still comes from a record read — but not one per token.

The lookup is now a prefetch. Before the pre-flight loop each lane harvests the FormID-shaped tokens out of the call's own value slots (`value`, `values`, dict entry values, a compose spec's fields, ctor args and nested sets, recursively), groups them by `ResolveWinner`'s winner plugin, and calls `view.CollectRecords` once per plugin. The 200-entry leveled-list ReplaceAll is one walk of Skyrim.esm, not 200. A token the harvest missed resolves to null and is simply not type-checked, which is the pre-PR behaviour, never a wrong refusal.

The throw is caught: the `CollectRecords` call is wrapped, and a plugin that cannot be opened is one sentence naming it, how many of the write's links it defines, and that nothing was written.

src/housecarl-core/CorpusRulebook.cs
Avick3110: Right that it was dead; folded by making it live rather than dropping it. `CreateRecordsCore` now builds the same lookup and passes it to `Validate`, so a literal FormID mixed in beside the `@editorid` siblings is type-checked here. The sibling tokens are not — their records do not exist yet — and the comment now says so.

Test: `ALiteralBesideASameCallSiblingIsTypeChecked` — a create whose Armor `Keywords` ReplaceAll carries `["@HcLinkTypeKw", <a Spell FormID>]` is refused on the literal.

src/housecarl-core/CorpusRulebook.cs
Avick3110: Real, folded (38de52f3). `BareTargetName` is gone; the phrase uses `RecordNaming.StripInterfaceToConcrete`, which is the exact `IRaceGetter → Race` plus `IFoo → Foo` mapping this one was reinventing (`StripGetterInterface` alone would leave a non-Getter interface with its leading `I`).

src/housecarl-mcp-tests/FormLinkTargetTypeTests.cs
Avick3110: Real, folded (747793e0) — with a different record than you named, for a reason worth stating.

A leveled-list entry's `Reference` sits under `Data`, so `compose` reaches it through `sets: [{"path": "Data.Reference"}]`, which is a nested `WriteRequest` and lands on the singular-Set leaf, not the `ValidateFromType` Fields arm at 1076. That arm needs a struct with a link as its OWN direct field. `Effect.BaseEffect` is one, and the shared fixture already has spells with effects, so the test composes an `Effect` with a `BaseEffect` pointed at a Spell and expects "is a Spell … links to MagicEffect": `AComposedStructFieldLinkToTheWrongRecordTypeIsRefused`.

Avick3110:
Both real, both folded. Pushed 89c1f080, 38de52f3, 747793e0.

**Create lane.** `CreateRecordsCore` builds the same link lookup off the view it already captures and passes it to `Validate`, so `housecarl_create` refuses a wrong-typed link before anything is allocated. That also makes the `@editorid` arm's `LinkTypeRefusal` live: a literal FormID mixed in beside same-call siblings is checked, the siblings are not (their records do not exist yet). Two tests cover it — `ACreatedRecordsLinkToTheWrongRecordTypeIsRefused` and `ALiteralBesideASameCallSiblingIsTypeChecked`. The CHANGELOG line now names create as well as apply, and says a Remove is not refused.

Thanks for checking `WriteEngine.cs:161` and the copy lane — agreed on both, and neither is touched.

**The wide-target phrasing.** The truncated sample is gone. Twelve or fewer, they are all named; past that it reads `any Item record (47 record types qualify)` — the target's own kind and how many, which is at least true and actionable, rather than twelve arbitrary names with no way to see the rest.

Also folded from the inline comments: the sweep no longer fires on Remove (it is scoped to the verbs that put a value in), `BareTargetName` is replaced by `RecordNaming.StripInterfaceToConcrete`, the composed-struct arm has a test, and the per-token full-plugin walk is now a per-plugin prefetch with the locked-plugin throw caught and turned into a refusal.

Build clean, 1904 tests pass, ci-all 128/128.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -674,6 +678,68 @@ static bool TryOffOrderCopyBody(
...
+        }
+        var types = new Dictionary<FormKey, Type>();
+        foreach (var (plugin, keys) in byPlugin)
+        {
+            var sink = new Dictionary<FormKey, IMajorRecordGetter>();
+            try { view.CollectRecords(session, plugin, keys, null, sink); }
+            catch (Exception ex)
+            {
+                return (_ => null,
+                    $"cannot check this write's FormLink targets: '{plugin}' defines {keys.Count} of them and could not be " +
```
Avick3110: **Unreadable link-target plugin hard-fails the whole write, and the catch is broader than this message.**

This is stricter than the gate's own leniency everywhere else. An apply does `Add` on `Keywords` with `01F2A3:SomeMod.esp`; between the index build and the call the user disables/renames `SomeMod.esp`, or another tool holds it open. `session.Overlay` throws and the *entire* write is refused — a write that succeeded before this PR, and whose link the gate would happily have left unchecked had the same FormID simply not resolved (`_linkTargets(value) is not { } actual` -> `return null`). Unresolvable is skipped, unreadable is fatal; those should probably agree.

Two smaller things in the same block:

- The `catch (Exception ex)` spans the whole `CollectRecords` call, but only the OPEN raises `PluginUnreadableException`. `LoadOrderResolver.CollectRecords` says in as many words that "a fault from the walk BELOW is a different problem, and calling it a held-open file would send the user after nothing" — this message says exactly that for every fault, walk faults included.
- `keys.Count` counts every FormID-shaped token harvested from *any* value slot, not FormLink values. A call with no FormLink field at all can be refused with "'X.esp' defines 1 of them".

Suggest skipping the plugin (leaving those links unchecked, as an unresolvable FormID is) and reserving the "close whatever holds it open" wording for `PluginUnreadableException`.

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -131,9 +141,16 @@ public static Corpus LoadCorpus(string corpusJsonPath)
...
-    public string? Validate(WriteRequest req, IReadOnlyCollection<string>? siblingEditorIds = null)
+    /// creations.
+    /// <para><paramref name="linkTargets"/> (the write path, which has the load order) turns on the FormLink TARGET
+    /// TYPE check: a link whose FormID resolves to a record the field cannot point at is refused. It is carried on a
+    /// per-call copy of the rulebook rather than threaded through the recursion, so every value slot — a singular
+    /// Set, a list element, a composed struct's field — sees it without a parameter at each hop.</para></summary>
+    public string? Validate(WriteRequest req, IReadOnlyCollection<string>? siblingEditorIds = null,
+        LinkTargetLookup? linkTargets = null)
     {
+        if (linkTargets is not null && !ReferenceEquals(linkTargets, _linkTargets))
```
Avick3110: **The per-call rulebook copy makes `_linkTargetNames` dead.**

The write lanes call `rulebook.Validate(req, …, linkTypes)` on the shared schema-only rulebook, whose `_linkTargets` is null. So `ReferenceEquals` is false on *every* edit and a fresh `CorpusRulebook` — with a fresh empty dictionary — is built each time. The field's own comment calls it a memo, and the `lock` in `AllowedLinkTypes` guards a dictionary no other thread can reach.

A 200-op apply that points every `Keywords` element at a spell reports 200 problems and runs 200 full `_corpus.Types` scans with a `WriteEngine.ResolveType` per record type — exactly the work the memo was written to collapse to one. Output is correct, but the comment describes behaviour the code does not have.

Building the derived rulebook once in each write lane and passing *that* down would make the memo real (and let the lock go).

src/housecarl-mcp-tests/FormLinkTargetTypeTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        var r = Apply(Fid(W.Armor), "Keywords", "Remove", Fid(W.SpellA));
+
+        Refused(r, "nothing to remove");
+        Assert.DoesNotContain("links to Keyword", r);
+    }
+
+    /// <summary>The create lane runs the same gate: a brand-new record whose link names the wrong type is refused
+    /// before anything is allocated.</summary>
+    [Fact]
+    public void ACreatedRecordsLinkToTheWrongRecordTypeIsRefused()
```
Avick3110, line 102: **These two create tests are not dry runs, in a shared-world collection.**

The class doc says "Dry runs: the shared world must stay unwritten", but `ACreatedRecordsLinkToTheWrongRecordTypeIsRefused` and `ALiteralBesideASameCallSiblingIsTypeChecked` call `CreateRecordsBatch` for real and rely on the refusal to keep the world clean.

If the gate regresses — or a later change moves the link check after allocation — they write `HcLinkTypeCreatePatch.esp` and `HcLinkTypeSiblingPatch.esp` (with `HcLinkTypeArmor` / `HcLinkTypeKw`) into the shared `records` fixture world. Every later test in the collection then runs against a world with two extra plugins, so one broken assertion turns into a cascade of unrelated failures plus leftover files on disk.

`CreateRecordsBatch` has no `dry_run`, so the guard has to be explicit: assert no patch file was written, and clean up if one was.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Folded in 3793927. The lookup now catches only `PluginUnreadableException` from the open, skips that plugin, and carries on — those links go unchecked exactly as an unresolvable FormID's do, so a write that was legal before this gate stays legal. A walk fault is no longer caught at all and is left to the caller, per `CollectRecords`' own contract. What was a refusal is now an honesty note on the outcome (`JoinNotes` beside the master-grow note, on all three lanes' success and dry-run returns), and it quotes the exception's own sentence, so "close whatever holds it open" comes from `PluginUnreadableException` and the unscannable wording comes from `PluginUnscannableException`. On the count: the harvest is schema-blind by design, so it cannot know which tokens are FormLink-valued — the note no longer claims they are, and says "N FormID value(s) in this write resolve to 'X.esp'".

src/housecarl-core/CorpusRulebook.cs
Avick3110: Folded in 18866cd, the first way. `Validate` lost its `linkTargets` parameter and the `ReferenceEquals` rebuild; a rulebook with the lookup is now derived once per write lane (`rulebook.WithLinkTargets(linkTypes)`, in all three lanes: apply, apply-in-place, create) and every edit in the call validates through that one object. The memo is real — a 200-op apply now scans `_corpus.Types` once per link target type instead of once per edit — and the `lock` is gone, since the derived rulebook is a per-call object used on the calling thread and the shared schema-only rulebook can never fill the memo (a refusal needs `_linkTargets`). The field comment now says that.

src/housecarl-mcp-tests/FormLinkTargetTypeTests.cs
Avick3110: Folded in f5c0d55. Both create tests now end with `AssertNoPatchWritten(...)`, which enumerates `houseCARL - <stem>*` under the fixture's `ModsDir`, deletes any it finds, and then asserts the list was empty — so a regression fails its own test and leaves the shared world without the extra plugin, instead of cascading into every later test in the collection.

PR #680, opened 2026-09-08
Print every FormID token with the plugin's on-disk spelling

### LINKED ISSUE #664: [Bug] FormID tokens spell the plugin filename in inconsistent case between reads
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/bulk-record-jobs/VERDICT.md`).

The same plugin appears as `ccBGSSSE025-AdvDSGS.esm` in one read's FormID tokens and `ccbgssse025-advdsgs.esm` in another's, on opposite sides of a local join of two `housecarl_records` outputs; 34 rows joined wrong in each of two independent runs. The token should carry one canonical spelling (the filename as it exists on disk, or the load-order listing's) everywhere it is printed.

Avick3110:
A FormID token used to spell its plugin in whatever case its FormKey happened to carry. A record read straight off a plugin got the on-disk spelling, because Mutagen names the overlay from the file it opened; a link into that plugin got whatever the holding plugin's master list says, because that is where the link's ModKey comes from; and a token the caller passed in was echoed as typed. ModKey compares case-insensitively, so every lookup still landed and nothing looked wrong — but two reads of the same record printed two different strings, and a local join of two `housecarl_records` outputs silently dropped those rows. This adds one place a token is spelled, `FormIdToken` in housecarl-core: `Of(FormKey)` returns Mutagen's own `ToString()` with only the plugin half respelled from the load order's canonical names, so the local-ID format and the null form cannot drift and the token still parses back.

`LoadOrderResolver.Build` publishes its plugin filenames (each a `Path.GetFileName` of a real path) as it builds, and a publish **replaces** the table rather than extending it: the server repoints at another MO2 instance or profile without a restart (`housecarl_set_mo2_instance`, and a profile change rebuilds the resolver), so a name the new order does not carry must not keep the old order's spelling. A name the current table does not hold keeps the spelling it arrived with — an off-order plugin read straight from disk prints its on-disk name as read, and so does a token naming a plugin that is not installed. The one exception is deliberate and documented on the class: the table is keyed case-insensitively, so an off-order file matching an ACTIVE plugin's name in every way but case prints the active plugin's spelling, because the token is a load-order token. The table is process-wide because the print sites are static and deep (a reflected link value in `ReadEngine` has no load order to ask); it is a `volatile` field swapped whole, so a reader sees one order's table or the other's, never a mix.

Site inventory (296 calls through the seam: 290 `Of`, 6 `Plugin`):

- **Already canonical, unchanged in behaviour:** `LoadOrderResolver.RuntimeToFormKey`'s two `FormKey.Factory($"{…:X6}:{_names[p]}")` calls — they build a key from the canonical table itself, and are parse inputs rather than prints. Winner and toucher plugin names (`WinnerPlugin`, touching lists) likewise come from `_names`. A record's own identity read off an overlay was already canonical; it now goes through the seam anyway, so it cannot drift later.
- **Changed — routed through the seam:** core `ClosureCopy`, `ClosureWalk`, `ContainmentIndex`, `DialogueValidate`, `EffectChain`, `ErrorCheck`, `OwnedChildLifecycle`, `ReadEngine`, `RemapEngine`, `ScriptPropertyCheck`, `SkyPatcherConflicts`, `SkyPatcherOverlay`, `WriteEngine`, `WritePatchBuilder`; server `Artifacts`, `DialogueSweepRender`, `DialogueTools`, `FormIdDoor`, `JsonWire`, `LoadOrderService`, `ReadTools`, `RecordsTools`, `ReverseWalkBatch`, `WriteTools`. That covers `formid` in every row and artifact — including the write's read-back rows, the removed and forwarded rows, and the effect-chain carriers — plus link values and `NoteRef` in the fields form, walk nodes and their `pulled_by` labels and cycle strings, delta and tree rows, dangling-ref sources and targets, dialogue topic/INFO/voice/script-finding rows, the cell-shell rows, merge conflicts, the owned-child union members, the reverse walk's reached keys, and the sentences that hand a token back to paste.
- **Parse inputs take the raw key, never the seam.** `FormIdToken.Of` is for prints only, so a value fed back into a parse uses `FormKey.ToString()`: CopyFrom's settable-FormLink rebuild in `WriteEngine`, the two same-call `@editorid` references in `WritePatchBuilder`, and both arms of the SkyPatcher value token, which is handed to `WriteEngine.ApplyVerb` as a `WriteRequest.Value` (the applied-op row prints the INI's own text and the read-back leaf, which routes through `ReadEngine`). The same holds on the read side: the key lists a scan hands to the walk, delta, tree, `info_order` and body-batch lanes in `RecordsTools`, and the condition-patch target `WriteEngine.PickSameTypeTarget` picks, are all re-parsed by their callee and so take `FormKey.ToString()` too. The one place a token is both printed and fed back is the reverse carrier walk's selection, which IS the `selection` the caller reads and pastes, so it stays canonical. This keeps the write path independent of a display helper.
- **Two sites formatted a token by hand** rather than calling `ToString()` — the runtime-FormID write refusal in `FormIdDoor` and the renamed-copy note in `LoadOrderService` — and both now go through the seam; nothing is left building `"{ID:X6}:{plugin}"` by hand outside the resolver's two parse-input calls above.
- **`defined_in` grouping** (`RecordsTools` and four sites in `LoadOrderService`) is not a token but is a plugin-name join key with the same drift, so it goes through `FormIdToken.Plugin`.
- **Not routed, on purpose:** `FaceGenPath` and `VoicePath` spell `ModKey.FileName` into a path on disk, not a token — the folder is named by the Creation Kit and matched case-insensitively, so respelling it there would be a behaviour change with nothing to gain. Plugin names printed from a plugin's own master references (`ErrorCheck`'s missing-master lists) are the same class of drift on a name that is not a FormID token; that is a separate question from this issue and is left alone. A refusal that echoes the FormID string the caller typed (`ApplyTools`' assignment-pair checks) keeps the caller's own text, so they can find the line they wrote.

Cover: `FormIdTokenCaseTests` builds a world whose patch's header names its master in lowercase (written normally, then the MAST entry respelled in place — same length, so the file stays valid). It asserts that reading the master's RACE directly and through the patch's link yields the same string; that a write against a lowercase-typed FormID comes back with both the op row and the read-back row spelled the order's way; that a walk's node labels do the same; and that the text render of a read, and of a read that failed, spell the plugin the order's way rather than echoing the case typed. `FormIdTokenPublishTests` publishes two orders and asserts a name the second does not carry falls back to its own spelling.

`dotnet test` 1900 passed / 0 failed; `ci-all` 128/128. Edits are formatting only, so #678's parsing changes should merge alongside; the one overlap is the single refusal line inside `FormIdDoor.Parse`.

Closes #664

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -78,6 +78,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   value" for all five.** A null field and the other arm of a union are not read faults, and a link that is set but
   points at a disabled plugin or a missing master is not unset: its remedy is to enable the plugin, not to widen
   the scope. The counts are now named separately.
+- **A FormID token now spells its plugin the way the file is spelled on disk, in every read.** A link's plugin
+  used to be spelled from the master list of whatever plugin held the link, and a token you passed in was echoed
+  as you typed it, so the same record could come back as `Skyrim.esm` in one read and `skyrim.esm` in another and
+  a join of two outputs missed those rows. Every token, and the `defined_in` grouping key, now prints the load
+  order's own spelling; a plugin the active order does not carry keeps the spelling it arrived with.
 - **`open-animation-replacer` now states how a DAR `_conditions.txt` chain binds, and that the DAR weapon-type
   numbers convert unchanged.** Both were marked unverified in its reference, and the skill told you to convert
   under one reading and say which. Both are now read off OAR's own source; the reference's §8 carries the rule,
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -111,11 +111,6 @@ saying it sets an expectation their install may contradict. Say what is known, a
   against the 1,024 ceiling it used to breach, and opens on its trigger. The CI guard that watched this file
   changed direction to match: it used to require every tool name to appear here, which is why it stayed green
   through all seventeen dead names, and now requires every `housecarl_*` name written here to be a live tool.
-- **A FormID token now spells its plugin the way the file is spelled on disk, in every read.** A link's plugin
-  used to be spelled from the master list of whatever plugin held the link, and a token you passed in was echoed
-  as you typed it, so the same record could come back as `Skyrim.esm` in one read and `skyrim.esm` in another and
-  a join of two outputs missed those rows. Every token, and the `defined_in` grouping key, now prints the load
-  order's own spelling; a plugin the order now loaded does not carry keeps the spelling it arrived with.
 - **`mutagen-reference` and `papyrus-reference` now say in their frontmatter that their lookup works offline.**
   Both carried only the shared compatibility sentence — the houseCARL MCP server and a configured Mod Organizer 2
   instance — which read as gating the whole skill, including a schema or signature lookup that reads files bundled
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -81,8 +81,10 @@ saying it sets an expectation their install may contradict. Say what is known, a
 - **A FormID token now spells its plugin the way the file is spelled on disk, in every read.** A link's plugin
   used to be spelled from the master list of whatever plugin held the link, and a token you passed in was echoed
   as you typed it, so the same record could come back as `Skyrim.esm` in one read and `skyrim.esm` in another and
-  a join of two outputs missed those rows. Every token, and the `defined_in` grouping key, now prints the load
-  order's own spelling; a plugin the active order does not carry keeps the spelling it arrived with.
+  a join of two outputs missed those rows. Every token now prints the load order's own spelling, and a
+  `defined_in` group is labelled with that spelling instead of whichever one arrived first — the counts were
+  never split, since the grouping already ignored case, but the label a caller joins on was unstable. A plugin
+  the active order does not carry keeps the spelling it arrived with.
 - **`open-animation-replacer` now states how a DAR `_conditions.txt` chain binds, and that the DAR weapon-type
   numbers convert unchanged.** Both were marked unverified in its reference, and the skill told you to convert
   under one reading and say which. Both are now read off OAR's own source; the reference's §8 carries the rule,
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -111,6 +111,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   against the 1,024 ceiling it used to breach, and opens on its trigger. The CI guard that watched this file
   changed direction to match: it used to require every tool name to appear here, which is why it stayed green
   through all seventeen dead names, and now requires every `housecarl_*` name written here to be a live tool.
+- **A FormID token now spells its plugin the way the file is spelled on disk, in every read.** A link's plugin
+  used to be spelled from the master list of whatever plugin held the link, and a token you passed in was echoed
+  as you typed it, so the same record could come back as `Skyrim.esm` in one read and `skyrim.esm` in another and
+  a join of two outputs missed those rows. Every token, and the `defined_in` grouping key, now prints the load
+  order's own spelling; a plugin the order now loaded does not carry keeps the spelling it arrived with.
 - **`mutagen-reference` and `papyrus-reference` now say in their frontmatter that their lookup works offline.**
   Both carried only the shared compatibility sentence — the houseCARL MCP server and a configured Mod Organizer 2
   instance — which read as gating the whole skill, including a schema or signature lookup that reads files bundled
--- a/src/housecarl-core/ClosureCopy.cs
+++ b/src/housecarl-core/ClosureCopy.cs
@@ -139,7 +139,7 @@ public static class ClosureCopy
         foreach (var rec in scratch.EnumerateMajorRecords())
             if (!RemapEngine.TryAddToFlatGroup(patch, (IMajorRecord)rec))
                 return CopyResult.Fail(new CopyRefusal(CopyRefusalKind.Transplant,
-                    $"{RecordNaming.StripOverlay(rec.GetType().Name)} {rec.FormKey} could not be placed in the patch",
+                    $"{RecordNaming.StripOverlay(rec.GetType().Name)} {FormIdToken.Of(rec.FormKey)} could not be placed in the patch",
                     Key: rec.FormKey));
 
         var copied = nodes.Select(n => new CopiedRecord(
@@ -181,7 +181,7 @@ public static class ClosureCopy
                 if (singleLink.FormKeyNullable is not { } fk || fk.IsNull || !isBound(fk)) continue;
                 if (!IsNullableLink(val))
                     return new CopyRefusal(CopyRefusalKind.RequiredForeignLink,
-                        $"the record's REQUIRED field '{prop.Name}' points at {fk}, which is in the source universe " +
+                        $"the record's REQUIRED field '{prop.Name}' points at {FormIdToken.Of(fk)}, which is in the source universe " +
                         "being copied away from: it cannot be nulled without inventing data, and keeping it would " +
                         "silently master that plugin", prop.Name, fk);
                 continue;
@@ -191,7 +191,7 @@ public static class ClosureCopy
             {
                 var keys = sub.EnumerateFormLinks()
                     .Where(l => !l.FormKey.IsNull && isBound(l.FormKey))
-                    .Select(l => l.FormKey.ToString()).Distinct().ToList();
+                    .Select(l => FormIdToken.Of(l.FormKey)).Distinct().ToList();
                 if (keys.Count > 0)
                     return new CopyRefusal(CopyRefusalKind.UnclearableSubstruct,
                         $"the record's field '{prop.Name}' carries reference(s) into the source universe " +
@@ -220,10 +220,10 @@ public static class ClosureCopy
                 // Pass 1 has already proven this link is nullable; the else-branch is a backstop against the two
                 // passes ever disagreeing, not the primary refusal path.
                 if (TryNullLink(val))
-                    stripped.Add(new StripEntry(prop.Name, fk.ToString()));
+                    stripped.Add(new StripEntry(prop.Name, FormIdToken.Of(fk)));
                 else
                     return StripResult.Fail(new CopyRefusal(CopyRefusalKind.RequiredForeignLink,
-                        $"the record's REQUIRED field '{prop.Name}' points at {fk}, which is in the source universe " +
+                        $"the record's REQUIRED field '{prop.Name}' points at {FormIdToken.Of(fk)}, which is in the source universe " +
                         "being copied away from: it cannot be nulled without inventing data, and keeping it would " +
                         "silently master that plugin", prop.Name, fk));
                 continue;
@@ -244,7 +244,7 @@ public static class ClosureCopy
                     {
                         var keys = elc.EnumerateFormLinks()
                             .Where(l => !l.FormKey.IsNull && isBound(l.FormKey))
-                            .Select(l => l.FormKey.ToString()).Distinct().ToList();
+                            .Select(l => FormIdToken.Of(l.FormKey)).Distinct().ToList();
                         if (keys.Count > 0)
                         { list.RemoveAt(i); stripped.Add(new StripEntry($"{prop.Name}[{i}]", string.Join(", ", keys))); }
                     }
@@ -257,7 +257,7 @@ public static class ClosureCopy
             {
                 var keys = sub.EnumerateFormLinks()
                     .Where(l => !l.FormKey.IsNull && isBound(l.FormKey))
-                    .Select(l => l.FormKey.ToString()).Distinct().ToList();
+                    .Select(l => FormIdToken.Of(l.FormKey)).Distinct().ToList();
                 if (keys.Count == 0) continue;
                 // Nulling the property removes EVERYTHING on it, not just the bound link(s) that forced the removal.
                 // Marked so the render can say so: for VirtualMachineAdapter this is every script on the clone.
@@ -689,7 +689,7 @@ public static class ClosureCopy
                 live.Add(l.FormKey);
                 if (offender is null && l.FormKey.ModKey != patchModKey && !isOnOrder(l.FormKey.ModKey))
                     offender = new OffOrderHit(l.FormKey, rec.FormKey,
-                        $"{RecordNaming.StripOverlay(rec.GetType().Name)} '{rec.EditorID ?? "<no editorid>"}' ({rec.FormKey})");
+                        $"{RecordNaming.StripOverlay(rec.GetType().Name)} '{rec.EditorID ?? "<no editorid>"}' ({FormIdToken.Of(rec.FormKey)})");
             }
         }
         return (live, offender);
--- a/src/housecarl-core/ClosureCopy.cs
+++ b/src/housecarl-core/ClosureCopy.cs
@@ -238,7 +238,7 @@ public static class ClosureCopy
                     if (el is IFormLinkGetter elLink)
                     {
                         if (elLink.FormKeyNullable is { } lk && !lk.IsNull && isBound(lk))
-                        { list.RemoveAt(i); stripped.Add(new StripEntry($"{prop.Name}[{i}]", lk.ToString())); }
+                        { list.RemoveAt(i); stripped.Add(new StripEntry($"{prop.Name}[{i}]", FormIdToken.Of(lk))); }
                     }
                     else if (el is IFormLinkContainerGetter elc)
                     {
--- a/src/housecarl-core/ClosureWalk.cs
+++ b/src/housecarl-core/ClosureWalk.cs
@@ -433,7 +433,7 @@ public static class ClosureWalk
                         var at = path.IndexOf(next);
                         cycles.Add(new WalkCycle(
                             path.Skip(at).ToList(), next,
-                            labels.TryGetValue(key, out var lb) ? lb : key.ToString()));
+                            labels.TryGetValue(key, out var lb) ? lb : FormIdToken.Of(key)));
                     }
                     continue;
                 }
--- a/src/housecarl-core/ClosureWalk.cs
+++ b/src/housecarl-core/ClosureWalk.cs
@@ -360,7 +360,7 @@ public static class ClosureWalk
                 hit.ArmIndex, hit.Arm.Spelling,
                 ChainTo(key), pulledBy, depth));
 
-            var label = $"{typeName} {key} ({hit.Body.EditorID ?? "<no editorid>"})";
+            var label = $"{typeName} {FormIdToken.Of(key)} ({hit.Body.EditorID ?? "<no editorid>"})";
             labels[key] = label;
             var outgoing = new List<FormKey>();
             edges[key] = outgoing;
--- a/src/housecarl-core/DialogueValidate.cs
+++ b/src/housecarl-core/DialogueValidate.cs
@@ -147,7 +147,7 @@ public static class DialogueValidate
     /// every member of the family is a CK-editor / byte-parity shape the game itself tolerates. All four gap
     /// surfaces (per-INFO, quest InputIssues, DLVW, DLBR) go through this so they cannot drift apart.</summary>
     static DialogueIssue GapIssue(string noun, FormKey fk, CkParityGap gap) =>
-        new(DialogueIssueSeverity.Warning, $"{noun} {fk} is missing the {gap.Subrecord} subrecord — {gap.Detail}");
+        new(DialogueIssueSeverity.Warning, $"{noun} {FormIdToken.Of(fk)} is missing the {gap.Subrecord} subrecord — {gap.Detail}");
 
     /// <summary>Resolve <paramref name="fk"/> to its load-order winner and validate the dialogue graph: a DIAL →
     /// validate that one topic; a QUST → fan out to EVERY topic the quest owns (a whole-order DIAL winner scan,
@@ -315,12 +315,12 @@ public static class DialogueValidate
             var win = view.ResolveWinner(fk);
             if (win is null)
                 return DialogueValidationReport.ForError(fk,
-                    $"{fk} is not in the active load order — nothing to validate. Pass a dialogue topic (DIAL) FormID to validate one topic, a quest (QUST) FormID to validate all of a quest's topics, or a dialogue view (DLVW) / branch (DLBR) FormID for a record-level CK-parity check.");
+                    $"{FormIdToken.Of(fk)} is not in the active load order — nothing to validate. Pass a dialogue topic (DIAL) FormID to validate one topic, a quest (QUST) FormID to validate all of a quest's topics, or a dialogue view (DLVW) / branch (DLBR) FormID for a record-level CK-parity check.");
 
             var body = view.GetRecord(session, win.Value.WinnerPlugin, fk);
             if (body is null)
                 return DialogueValidationReport.ForError(fk,
-                    $"{fk} resolves to a winner in {win.Value.WinnerPlugin} but its body could not be fetched (the plugin may have changed since the index was built) — re-run to rebuild and try again.");
+                    $"{FormIdToken.Of(fk)} resolves to a winner in {win.Value.WinnerPlugin} but its body could not be fetched (the plugin may have changed since the index was built) — re-run to rebuild and try again.");
 
             if (body is IDialogTopicGetter topic)
             {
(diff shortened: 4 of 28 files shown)


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -852,7 +852,7 @@ List<KeyValuePair<string, string>> Echo()
                         if (carrierSeen.Add(r.Seed)) carrierSel.Add(r.Seed);
                         foreach (var row in r.Result.Rows)
                         {
-                            var key = row.Carrier.ToString();
+                            var key = FormIdToken.Of(row.Carrier);
```
Avick3110, line 855: **The effect-chain `formid` column was not routed, but the selection list built from it was.** This line canonicalises `row.Carrier` into `carrierSel`, yet the three places that actually *emit* that same carrier are untouched: `JsonWire.cs:993` (`w.WriteString("formid", row.Carrier.ToString())`), `Artifacts.cs:368` (same, and that artifact is saved with identity `"formid"`), and `ReadTools.cs:539` (the text render).

Concrete failure: `housecarl_records form=chain` over an MGEF whose carriers are reached as links from plugins that spell the defining master differently. The `carriers[].formid` column comes back with the master-list spelling, while the follow-up carrier selection this line builds comes back canonical. Feeding the artifact back in on `formid`, or joining the two outputs, drops exactly the rows the PR set out to stop dropping — and the PR body claims `formid` in every row and artifact is covered.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4102,11 +4102,11 @@ seedWin is null
                             refusal = $"the walk reached a {type} ({key}, via {pulledBy}) — a node class this call excludes with severity 'refuse'. Nothing is returned for this call.";
                             return Array.Empty<WalkSeedResult>();
                         }
-                        st.Nodes.Add(new WalkNodeRow(key.ToString(), type, body.EditorID, hop, pulledBy, "kept", $"excluded ({type}, severity stop) — recorded as a boundary, not entered"));
+                        st.Nodes.Add(new WalkNodeRow(FormIdToken.Of(key), type, body.EditorID, hop, pulledBy, "kept", $"excluded ({type}, severity stop) — recorded as a boundary, not entered"));
                         continue;
                     }
                     bool atCap = hop >= depth;
-                    st.Nodes.Add(new WalkNodeRow(key.ToString(), type, body.EditorID, hop, pulledBy,
+                    st.Nodes.Add(new WalkNodeRow(FormIdToken.Of(key), type, body.EditorID, hop, pulledBy,
```
Avick3110, line 4123: **`pulled_by` still drifts, so the walk table can no longer self-join.** The node's `formid` is canonical here (and at 4090/4105), but the label enqueued for its children at line 4117 is still built raw:

```csharp
var label = $"{type} {key} ({body.EditorID ?? "<no editorid>"})";
```

Every child row's `pulled_by` therefore names its parent with whatever spelling the link carried, while that parent's own row names it canonically. Concrete: a forward walk from an NPC through a plugin that spells `Skyrim.esm` as `skyrim.esm` in its master list yields a child row `pulled_by: "NPC_ 0013BBF:skyrim.esm (...)"` sitting next to the parent row `formid: "0013BBF:Skyrim.esm"` — matching parent to child by string, the obvious use of the column, fails.

Same class two lines up: the cycle string at 4078 (`st.Cycles.Add($"{pulledBy} -> {key} ...")`) and the refuse sentence at 4102 (`({key}, via {pulledBy})`).

src/housecarl-core/DialogueValidate.cs
```diff
@@ -548,7 +548,7 @@ internal static TopicValidation ValidateTopic(IDialogTopicGetter topic, string w
                 var lk = link.FormKey;
                 if (!lk.IsNull && BadRef(lk, "a dialogue topic (DIAL)", b => b is IDialogTopicGetter) is { } lwhy)
                     issues.Add(new(DialogueIssueSeverity.Problem,
-                        $"INFO {info.FormKey} links (LinkTo) to {lk}, which {lwhy} — the conversation chain is broken."));
+                        $"INFO {FormIdToken.Of(info.FormKey)} links (LinkTo) to {lk}, which {lwhy} — the conversation chain is broken."));
```
Avick3110: **The canonicalised half of this sentence was the half that never drifted; the drifting half was left raw.** `info.FormKey` is a record's own key read off the overlay — already the on-disk spelling by construction (the PR body says so). `lk` is a *link target*, whose ModKey comes from the master list of the plugin holding the INFO — the exact source of drift #664 is about, and it still prints `.ToString()`.

Same on lines 542 (`PNAM -> {pnam.Value}`), 757 (`Run On reference {refKey}`), 794 (`references {pk}`), 810 (`the placed reference {objFk}`). Note line 801 in this same diff *did* route `cg.ComparisonValue.FormKey`, so the file is internally inconsistent.

Concrete: validating a topic in a plugin whose header spells `Dawnguard.esm` lowercase reports `links (LinkTo) to 002B1F:dawnguard.esm`, while reading that same DIAL reports `002B1F:Dawnguard.esm` — the reader cannot paste one into the other's join.

src/housecarl-core/ClosureWalk.cs
```diff
@@ -433,7 +433,7 @@ static List<WalkCycle> FindCycles(
                         var at = path.IndexOf(next);
                         cycles.Add(new WalkCycle(
                             path.Skip(at).ToList(), next,
-                            labels.TryGetValue(key, out var lb) ? lb : key.ToString()));
+                            labels.TryGetValue(key, out var lb) ? lb : FormIdToken.Of(key)));
```
Avick3110, line 436: **This branch is unreachable, so the change is a no-op and the live label still drifts.** `labels[key]` is assigned at line 363 for every expanded node, immediately before `edges[key]` is set — and the cycle scan only ever visits keys that are in `edges`. So `TryGetValue` always succeeds and `FormIdToken.Of(key)` never runs.

The string a caller actually sees is the one built at 363:

```csharp
var label = $"{typeName} {key} ({hit.Body.EditorID ?? "<no editorid>"})";
```

which is still raw. A closure walk that closes a cycle through a record reached by a link from a plugin with a lowercase master entry reports the cycle label with the drifting spelling. The fix belongs on 363.

src/housecarl-core/ClosureCopy.cs
```diff
@@ -244,7 +244,7 @@ static StripResult ApplyStrip(IMajorRecord record, Func<FormKey, bool> isBound)
                     {
                         var keys = elc.EnumerateFormLinks()
                             .Where(l => !l.FormKey.IsNull && isBound(l.FormKey))
-                            .Select(l => l.FormKey.ToString()).Distinct().ToList();
+                            .Select(l => FormIdToken.Of(l.FormKey)).Distinct().ToList();
```
Avick3110, line 247: **Sibling branch six lines up was missed.** Line 241 handles the *list-of-direct-links* case and still writes `lk.ToString()` into the `StripEntry` value; this line handles the *list-of-link-bearing-elements* case and now writes `FormIdToken.Of(...)`. Both feed the same `stripped` list, which renders as one column of stripped references.

Concrete: copying an NPC whose `Factions[]` (link-bearing elements) and, say, `Keywords[]` (direct links) both point into the source universe produces one refusal listing `0001F1:Skyrim.esm` beside `0002A3:skyrim.esm` in the same message.

src/housecarl-core/SkyPatcherOverlay.cs
```diff
@@ -1173,7 +1173,7 @@ static void ApplySetOne(object record, RecordMap fieldMap, string opKey, OpMap m
 
         string token;
         if (v.IsNameLiteral) token = v.NameText!;                       // rename literal, wrapper stripped
-        else if (v.Address is { IsFormId: true } a && TryFormKey(a, out var fk1)) token = fk1.ToString();
+        else if (v.Address is { IsFormId: true } a && TryFormKey(a, out var fk1)) token = FormIdToken.Of(fk1);
```
Avick3110: **The other arm that assigns `token` a FormKey was left raw.** Eight lines down:

```csharp
if (ResolveFormValue(v, map.FormType, resolver) is { } rk) token = rk.ToString();
```

Same local, same `SkyPatcherAppliedOp` value column. A SkyPatcher INI that addresses a value by FormID gets the canonical spelling; one that addresses the same record by EditorID gets `rk`'s spelling. Two lines of one INI writing the same form then read back as two different tokens in the replay's `after` / applied-op rows.

src/housecarl-core/FormIdToken.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (!string.IsNullOrEmpty(n)) Canon[n] = n;
+    }
+
+    /// <summary>The canonical spelling of one plugin filename, or the name unchanged when no load order has
+    /// published it.</summary>
+    public static string Plugin(string fileName)
+        => Canon.TryGetValue(fileName, out var c) ? c : fileName;
+
+    /// <summary>This FormKey as a token. Byte-for-byte <see cref="FormKey.ToString"/> except for the plugin half,
+    /// which is respelled canonically — so the local-ID formatting and the null form can never drift from
```
Avick3110, line 56: **An off-order read is respelled to a *different file*'s name when the active order carries a same-named plugin.** The class doc directly above claims "an off-order plugin read straight from disk … keeps the spelling it arrived with" — but `Plugin()` is keyed `OrdinalIgnoreCase`, so any name that case-insensitively matches an active plugin is rewritten, regardless of which file it came from.

Concrete: two MO2 mod folders each ship a plugin, one spelled `MyMod.esp` (enabled) and one `mymod.esp` (disabled). `housecarl_records source=<path to the disabled folder>` reads the *disabled* file, and every token in that output now says `MyMod.esp` — the on-disk spelling of the other file. This codebase already treats that shape as real (LoadOrderService.cs:3315: "Two copies of one filename on opposite arms"). Either exempt the off-order lane from the seam or state the limitation honestly in the doc.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -502,7 +502,7 @@ public static string RenderBatch(IReadOnlyList<ReadOutcome> outcomes, int maxCha
                 if (ms.Length >= cap) { rowsTruncated = true; break; }
                 var o = outcomes[i];
                 string? hit = matches is { } mt && i < mt.Count ? mt[i] : null;   // multi-target references= un-merge
-                if (o.Error is not null) { w.WriteStartObject(); w.WriteString("formid", o.FormKey.ToString()); w.WriteString("error", o.Error); if (hit is not null) w.WriteString("matches", hit); w.WriteEndObject(); }
+                if (o.Error is not null) { w.WriteStartObject(); w.WriteString("formid", FormIdToken.Of(o.FormKey)); w.WriteString("error", o.Error); if (hit is not null) w.WriteString("matches", hit); w.WriteEndObject(); }
```
Avick3110, line 505: **Two more machine-readable `formid` columns were missed.** `JsonWire.cs:235` (the full-readback rows) and `JsonWire.cs:2719` both still emit `w.WriteString("formid", r.Target.ToString())`, and `WriteTools.cs:249/290/365` print `r.Target` raw in the text render.

Concrete: `housecarl_apply` against a record whose FormID the caller typed as `0013bbf:skyrim.esm` echoes the readback row as typed, while a follow-up `housecarl_records` on the same record returns `0013BBF:Skyrim.esm` — the write-then-verify join the readback exists for is exactly the one that breaks. The PR body claims "`formid` in every row and artifact" is covered.

src/housecarl-mcp/ReverseWalkBatch.cs
```diff
@@ -113,7 +113,7 @@ bool Verify(FormKey candidate, IReadOnlySet<FormKey> frontier)
         // Seeds first, then each hop in order: the selection reads in walk order, and the render says the seeds
         // are in it.
         var selection = new List<string>(seedKeys.Count);
-        foreach (var k in seedKeys) selection.Add(k.ToString());
+        foreach (var k in seedKeys) selection.Add(FormIdToken.Of(k));
```
Avick3110, line 116: Seeds go through the seam here; the reached keys appended to the *same* `selection` list on the next line do not (`hop.Reached` → `k.ToString()`). One list, two spellings. Reached keys come off the reverse index so they are usually already canonical, but a seed that reappears as a reached node — or any caller re-entering this selection as an artifact and joining it against a `records` output — meets a mixed list. Route both, or neither.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4221,26 +4221,26 @@ public IReadOnlyList<InfoOrderRow> InfoOrderBatch(IReadOnlyList<string> formids,
...
+                                          $"the winner body of {FormIdToken.Of(fk)} could not be read from '{win.Value.WinnerPlugin}'."));
                 continue;
             }
             if (body is not Mutagen.Bethesda.Skyrim.IDialogTopicGetter)
             {
                 var typeName = RecordNaming.StripOverlay(body.GetType().Name);
-                rows.Add(new InfoOrderRow(fk.ToString(), typeName, body.EditorID, win.Value.WinnerPlugin, null,
-                    $"{fk} is a {typeName}, and the info_order form renders the merged INFO sequence of a DIALOGUE TOPIC (DIAL). " +
+                rows.Add(new InfoOrderRow(FormIdToken.Of(fk), typeName, body.EditorID, win.Value.WinnerPlugin, null,
+                    $"{FormIdToken.Of(fk)} is a {typeName}, and the info_order form renders the merged INFO sequence of a DIALOGUE TOPIC (DIAL). " +
```
Avick3110, line 4252: **A token handed back for the caller to paste, left raw, one line under two that were routed.** Line 4239:

```csharp
"For a quest's topics, select them by composition: types=[\"DIAL\"] where=[\"Quest = " + fk + "\"]."
```

The remedy sentence tells the user to paste `Quest = <token>` into a `where=` clause, so this is precisely the "sentences that hand a token back to paste" class the PR body says it covers, and it disagrees with the `formid` on the very same row.

Same miss at line 4001: `$"the winner body of {seedFk} could not be read from …"`, sitting directly under the `WalkSeedResult(FormIdToken.Of(seedFk), …)` that was routed.

src/housecarl-core/WriteEngine.cs
```diff
@@ -2101,7 +2101,7 @@ static void TransplantValue(object parent, PropertyInfo prop, object srcVal)
             // a directly-assignable immutable reference (string, MemorySlice…) — safe to share while the source overlay lives
             if (pt.IsInstanceOfType(srcVal)) { prop.SetValue(parent, srcVal); return; }
             // a settable FormLink slot (rare) — build the matching concrete from the source key
-            if (srcVal is IFormLinkGetter sfl && TryFormLink(sfl.FormKey.ToString(), Nullable.GetUnderlyingType(pt) ?? pt, out var mk)
+            if (srcVal is IFormLinkGetter sfl && TryFormLink(FormIdToken.Of(sfl.FormKey), Nullable.GetUnderlyingType(pt) ?? pt, out var mk)
```
Avick3110: **A parse input routed through a print-only seam.** `FormIdToken.Of` is documented as "the ONE place a FormID token is *spelled*" and every other site is a print; here its output is fed straight back into `TryFormLink` to build a link that gets written into a plugin. Same shape at `WritePatchBuilder.cs:3488` and `3547` (same-call `@editorid` refs). The PR body itself calls out the resolver's two `FormKey.Factory` calls as "parse inputs rather than prints" and treats that as a reason they are fine — these three are the same class and were routed anyway.

It is harmless today because `ModKey` compares case-insensitively, but it makes the write path depend on a display helper: any later change to `Of` (a prefix, a decoration, a different null rendering) silently corrupts a write instead of only a message. Use `sfl.FormKey.ToString()` here.

src/housecarl-core/FormIdToken.cs
```diff
@@ -0,0 +1,55 @@
...
+/// the two spellings PRINT differently, and two reads of the same record then disagree in text, which is a silent
+/// wrong answer to anything that joins them (#664).</para>
+///
+/// <para>The canonical spelling is the load order's own: <see cref="LoadOrderResolver"/> publishes its plugin
+/// filenames (each <c>Path.GetFileName</c> of a real path) here as it builds. A name no order has published — an
+/// off-order plugin read straight from disk, a token naming a plugin that is not installed — keeps the spelling it
+/// arrived with, which for an off-order read is already the on-disk one. The table is process-wide because the
+/// print sites are static and deep (a reflected link value in <see cref="ReadEngine"/> has no load order to ask);
+/// one process serves one load order, and entries are additive and keyed case-insensitively, so a later build
+/// only ever corrects a spelling.</para>
```
Avick3110: **"One process serves one load order" is not true of this process, and the table is never cleared.** `housecarl_set_mo2_instance` (`SetupTools.cs:24` → `LoadOrderService.SetInstance`) repoints the running server at a different MO2 instance, and `LoadOrderService.cs:2089` rebuilds the resolver on a profile change, both without a restart. `Canon` is `static` and `Publish` is add-only, so every plugin name instance A published stays in the table for the life of the process.

Where that bites: a name the new instance does *not* carry keeps A's spelling forever. Switch from instance A (which ships `AwesomeMod.esp`) to instance B (which does not carry it at all), then read B's `awesomemod.esp` off-order by path — the token prints `AwesomeMod.esp`, a spelling that names no file in the load order now loaded and no file B has on disk. Same shape for a plugin renamed on disk between rebuilds: the old casing is still in the table, and if the rename dropped the plugin from the order the old spelling is what prints, silently.

The self-correcting case the doc leans on ("a later build only ever corrects a spelling") holds only while the *same* name is republished. It is exactly the names that are **not** republished that go stale, and the code has no way to notice.

Two cheap fixes, either one: clear `Canon` at the top of `Publish` so the table is exactly the current order (it is rebuilt in the same call anyway), or key the table on the resolver instance rather than the process. The first is one line and makes the doc paragraph true.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. All three sites now route: JsonWire.cs:993, Artifacts.cs:368 and the text render in ReadTools.cs. The carrier column and the selection built from it now agree, so the artifact feeds back in on formid.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The child label at 4117, the cycle string at 4078 and the refuse sentence at 4102 all route now, so pulled_by names the parent the way the parent's own row does. Same class in ClosureWalk fixed at its live label (see the reply on that thread). A test covers the walk labels.

src/housecarl-core/DialogueValidate.cs
Avick3110: Folded, and so were the other four. 542 (PNAM), 551 (LinkTo), 757 (Run On reference), 794 (condition reference) and 810 (GetIsID placed reference) all route the LINK side now. Swept the rest of the file too: the DialogTopic.Quest sentence at 431 and the dead-alias sentence at 822 were the same class and are routed.

src/housecarl-core/ClosureWalk.cs
Avick3110: Folded, on line 363 as you said - that is the label a caller sees, and it now routes. The 436 fallback stays routed as well: it is unreachable today, but leaving the two spellings different there would just plant the next bug.

src/housecarl-core/ClosureCopy.cs
Avick3110: Folded. Line 241 (list of direct links) now routes like the sibling branch, so one refusal cannot list two spellings of one universe.

src/housecarl-core/SkyPatcherOverlay.cs
Avick3110: Refuted, and the arm you were comparing against went the other way instead. `token` is not a print: it is handed to `WriteEngine.ApplyVerb` as the WriteRequest's Value, a parse input. The applied-op row records `v.Raw` (what the INI said) and `LeafToken(record, segs)` (read back through ReadEngine, which does route), never `token`. So the FormID arm at 1176 was the one out of place - it now uses `fk1.ToString()`, matching the EditorID arm's `rk.ToString()` and the parse-input rule from finding 11. Both arms raw, one contract.

src/housecarl-core/FormIdToken.cs
Avick3110: Folded together with the table-lifetime finding. The doc no longer claims what the code does not do: it now states the case-insensitive key plainly - an off-order file matching an ACTIVE plugin's name in every way but case prints the active plugin's spelling, because the token is a load-order token. The wider claim (a name no order published keeps its own spelling) is now true, because the table is replaced per build rather than accumulated.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded. JsonWire.cs:235 and 2719 route, and so do the four text-render sites in WriteTools (the readback error rows, the removed rows and the forwarded rows). A test writes through a lowercase-typed FormID and asserts both the op row and the readback row come back the order's way.

src/housecarl-mcp/ReverseWalkBatch.cs
Avick3110: Folded - `hop.Reached` routes too, so the selection list is one spelling end to end.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded, both. The remedy sentence at 4239 and the unreadable-winner sentence at 4001 route now, as do the paste-back lines in the off-order CopyFrom locator further down the file.

src/housecarl-core/WriteEngine.cs
Avick3110: Folded. All three take the raw key now - `sfl.FormKey.ToString()` here and `rec.FormKey.ToString()` at WritePatchBuilder 3488/3547. The write path no longer depends on a display helper. The same reasoning settled the SkyPatcher thread the other way (see there).

src/housecarl-core/FormIdToken.cs
Avick3110: Folded, the one-line way. `Publish` builds a fresh dictionary and swaps it in (volatile field, replaced whole so a concurrent reader sees one table or the other, never a mix), and the class doc paragraph is rewritten to say exactly that. A unit test publishes two orders and asserts a name the second does not carry falls back to its own spelling.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1360,7 +1360,7 @@ List<KeyValuePair<string, string>> Echo()
             //      rendering the chain or reading the reached set, seam-checked throughout. ----
             if (walk is not null && outcome.Error is null && outcome.Groups is null)
             {
-                var seedKeys = outcome.Keys.Select(k => k.ToString()).ToArray();
+                var seedKeys = outcome.Keys.Select(FormIdToken.Of).ToArray();
```
Avick3110: Note rather than a defect — I found no correctness bug in this diff.

The description states that parse inputs take the raw key, never the seam. Six sites here break that rule: lines 1363, 1387, 1424, 1446, 1682 and 1720 all hand `FormIdToken.Of` output straight into `WalkLane` / `DeltaBatch` / `TreeBatch` / `InfoOrderBatch` / `ResolveBatchFromPole`, each of which re-parses it.

It is safe today only because the table is keyed case-insensitively, so `Of` can only ever change case, and `FormKey.Factory` / `ModKey` compare case-insensitively. It stops being safe the day `Of` gains a fallback that does not preserve the arriving spelling. Either route these through the raw key, or correct the inventory in the description so the rule still describes the code.

src/housecarl-core/WriteEngine.cs
```diff
@@ -494,7 +494,7 @@ internal static object NavigateToConditionArm(IMajorRecord ov, string field, int
         if (linkedGetter is null) return null;
         foreach (var rec in mod.EnumerateMajorRecords())
             if (linkedGetter.IsInstanceOfType(rec) && rec.FormKey != current && !rec.FormKey.IsNull)
-                return rec.FormKey.ToString();
+                return FormIdToken.Of(rec.FormKey);
```
Avick3110: Same note as the one on `RecordsTools.cs`. `PickSameTypeTarget`'s return value becomes `WriteRequest.Value` at line 402, so this is a parse input, which the description says never goes through the seam.

Harmless while `Of` can only change case, but it is one of the eight sites that make the stated rule untrue. Either the raw key here, or a correction to the inventory.

plugin/CHANGELOG.md
```diff
@@ -78,6 +78,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   value" for all five.** A null field and the other arm of a union are not read faults, and a link that is set but
   points at a disabled plugin or a missing master is not unset: its remedy is to enable the plugin, not to widen
   the scope. The counts are now named separately.
+- **A FormID token now spells its plugin the way the file is spelled on disk, in every read.** A link's plugin
```
Avick3110, line 81: The `defined_in` half of this entry overstates what was broken. Every `defined_in` group dictionary is already `StringComparer.OrdinalIgnoreCase` — `LoadOrderService.cs:4458` and `:4958`, `RecordsTools.cs:2545` — so the counts never split by case.

What the change actually fixes is the group's displayed label, which used to be whichever spelling arrived first. That is still a real fix for a caller joining labels across two calls, but it is not a count that was wrong. Worth rewording so the entry says the label, not the join.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. All six take the raw key now (`outcome.Keys.Select(k => k.ToString())`), so the walk, delta, tree, info_order and body-batch lanes are handed the same string the caller could have typed. None of them needed canonical spelling: the callee re-parses, and every print downstream goes through the seam on its own. The inventory paragraph in the description now names them.

src/housecarl-core/WriteEngine.cs
Avick3110: Folded. `PickSameTypeTarget` returns `rec.FormKey.ToString()`, so nothing on the way to `WriteRequest.Value` touches a display helper. That is the eighth site; the only token left that is both printed and fed back is the reverse carrier walk's selection, which IS the `selection` list the caller reads and pastes, and the earlier thread asked for it canonical - the description now says so.

plugin/CHANGELOG.md
Avick3110: Folded, reworded to say the label. The entry now reads that a `defined_in` group is labelled with the order's spelling instead of whichever one arrived first, and states outright that the counts were never split because the grouping already ignored case.

PR #681, opened 2026-09-09
Harvest a write's link targets from the rulebook's own walk

Avick3110:
#679 added the FormLink target type-check, and prefetched the records it needs with a hand-written sweep of the value slots a write can carry (value, values, dict entries, a compose's fields/ctor_args/nested sets). That sweep mirrored by hand what the rulebook does when it validates, so a link slot added to the validator and not to it would resolve to null and go unchecked — a silent return to pre-#679 behaviour — and the unreadable-plugin note had to say "FormID value(s)" because the sweep could not tell a link value from any other string.

The rulebook now enumerates them itself. `WithLinkHarvest` derives a harvest rulebook whose `CollectLinkValues` runs the ordinary `Validate` walk with the link check in collect mode: at the one place the check reads a link value, the value is added to the sink instead. One walk decides both what gets resolved and what gets checked, so the two cannot name different slots, and the hand-written sweep is gone. The apply lanes need each edit's record type before they can walk it, so both now run Phase 1 in two passes — resolve and harvest per edit, then pre-flight every staged edit against the one lookup the harvest fed; problems carry their edit index so a report mixing a resolve problem with a pre-flight one still reads in edit order. The create lane's edits are already rooted at the declared type, so it harvests up front as before. Once-per-plugin resolution and the unreadable-plugin skip are unchanged, and the note now counts FormLink values truthfully.

Test: a link two composition levels down — a composed `ContainerEntry`'s nested set into `Item.Item` — pointing at a spell is refused. The existing FormLinkTargetTypeTests stay green. 1952 tests, 128/128 probes.

Relates to #679 and #663.

### DIFF
--- a/src/housecarl-core/WritePatchBuilder.cs
+++ b/src/housecarl-core/WritePatchBuilder.cs
@@ -435,26 +435,33 @@ public static class WritePatchBuilder
         //     session, never an arbitrary un-enabled plugin, so no winner-confusion hazard arises. ---
         var view = resolver.Capture();
         epoch = view.Stamp;                                               // stamped on every outcome from here down
+        // The FormLink values this call sets are harvested by the RULEBOOK's own walk (CollectLinkValues), so the set
+        // resolved below is exactly the set the check reads — no second, hand-written list of value slots to drift
+        // from it. The walk needs each edit's record TYPE, which only the resolve loop can derive, so Phase 1 runs in
+        // two passes: resolve + harvest per edit, then validate every staged edit against the lookup the harvest fed.
         var linkTokens = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-        foreach (var e in edits) HarvestLinkTokens(e.Value, e.Values, e.Entries, e.Struct, e.Structs, linkTokens);
-        var (linkTypes, linkNote) = LinkTypeLookup(view, session, linkTokens);
-        var linkRulebook = rulebook.WithLinkTargets(linkTypes);
-        var resolved = new List<(PatchEdit edit, IMajorRecordGetter? body, string? winnerPlugin, IMajorRecord? patchLocal, WriteRequest req, string label, IMajorRecordGetter? srcBody)>(edits.Count);
-        var problems = new List<string>();
+        var harvestRulebook = rulebook.WithLinkHarvest(linkTokens);
+        var staged = new List<(int order, PatchEdit edit, IMajorRecordGetter? body, string? winnerPlugin, IMajorRecord? patchLocal, WriteRequest req, string label, IMajorRecordGetter? srcBody)>(edits.Count);
+        // Ordered, so a report mixing a resolve problem with a pre-flight one still reads in the caller's edit order
+        // even though the two are decided in different passes.
+        var problems = new List<(int Order, string Message)>();
         // Records the extended patch DEFINES (FormKey in the patch's own master space — created by a prior into=
         // call), built lazily ONCE on the first load-order miss, not per miss.
         // Deliberately NOT every record the patch contains: an override the patch merely CARRIES resolves via the
         // load order like any other record, so a target whose defining plugin is disabled stays a loud refusal —
         // never a silent edit of the patch's possibly-stale override copy.
         Dictionary<FormKey, IMajorRecord>? patchDefined = null;
+        int order = -1;
+        void Problem(string message) => problems.Add((order, message));
         foreach (var e in edits)
         {
+            order++;
             IMajorRecordGetter? body = null; string? winnerPlugin = null; IMajorRecord? patchLocal = null;
             var w = view.ResolveWinner(e.Target);
             if (w is not null)
             {
                 body = view.GetRecord(session, w.Value.WinnerPlugin, e.Target);
-                if (body is null) { problems.Add($"{FormIdToken.Of(e.Target)}: winner '{w.Value.WinnerPlugin}' did not yield it on fetch (a load-order inconsistency)."); continue; }
+                if (body is null) { Problem($"{FormIdToken.Of(e.Target)}: winner '{w.Value.WinnerPlugin}' did not yield it on fetch (a load-order inconsistency)."); continue; }
                 winnerPlugin = w.Value.WinnerPlugin;
             }
             else
@@ -471,7 +478,7 @@ public static class WritePatchBuilder
                 }
                 if (patchLocal is null)
                 {
-                    problems.Add($"{FormIdToken.Of(e.Target)}: not present in the load order ({view.PluginCount} plugins)"
+                    Problem($"{FormIdToken.Of(e.Target)}: not present in the load order ({view.PluginCount} plugins)"
                         + (extend
                             ? $", and not a record '{fileName}' (the patch being extended) itself defines — a record " +
                               "the patch merely OVERRIDES resolves via the load order, so its defining plugin must be enabled."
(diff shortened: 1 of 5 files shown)


src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -924,7 +926,7 @@ static PatchOutcome ApplyInPlaceCore(
                 Key = e.Key, Value = e.Value, Values = e.Values, Entries = e.Entries, Struct = e.Struct, Structs = e.Structs,
             };
             var label = Label(req);
-            if (linkRulebook.Validate(req) is { } reject) { problems.Add($"{recType} {e.Target} [{label}]: {reject}"); continue; }
+            harvestRulebook.CollectLinkValues(req);
```
Avick3110: In-place lane: the schema pre-flight moved from here to Phase 1b, so it now runs *after* the CopyFrom block instead of before it. Two consequences for an in-place `CopyFrom` edit that has both a schema problem and a source problem — e.g. `field_path` naming a field the record type does not have, plus a `from_plugin` that is not in the load order. Before, the caller saw the field-path refusal; now the edit `continue`s out of the CopyFrom arm first and the caller only sees the source message, so the schema error stays hidden until they fix the source and re-run. Secondly, an edit that is already doomed still had its link values harvested a few lines above, so `LinkTypeLookup` may open and walk a plugin purely on behalf of an edit no pre-flight will ever validate. The patch lane does not have this asymmetry — there the harvest sits at the end of the loop, after the CopyFrom rejections. Moving `CollectLinkValues(req)` to just before `staged.Add(...)` here would make the two lanes match and drop the wasted resolve.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -717,7 +741,7 @@ static bool TryOffOrderCopyBody(
             catch (PluginUnreadableException ex)
             {
                 (skipped ??= new List<string>()).Add(
-                    $"{keys.Count} FormID value(s) in this write resolve to '{plugin}', whose link targets were NOT " +
+                    $"{keys.Count} FormLink value(s) in this write resolve to '{plugin}', whose link targets were NOT " +
```
Avick3110: The wording now says "FormLink value(s)" but the number is still `keys.Count`, which is the set of distinct **FormKeys** that resolve to this plugin — not the number of link values the write sets. Two slots pointing at the same target (a leveled list that lists one spell twice, or `Item.Item` and a second entry naming the same record), or two spellings of one FormID (`000ABC:Skyrim.esm` and `00000abc:skyrim.esm`, which the OrdinalIgnoreCase token set keeps as two tokens but `FormKey.TryFactory` collapses to one key), report as "1 FormLink value(s)". The changelog entry sells this line as counting FormLink values, so it is worth either saying "link target(s)" here or counting the tokens that mapped to these keys.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Folded in ddd1043f. `CollectLinkValues(req)` now sits just before `staged.Add(...)`, after the CopyFrom block, so the two lanes match and a doomed edit no longer makes `LinkTypeLookup` open and walk a plugin.

On the ordering: it has to change, and I could not keep schema-first without a second walk. Pre-flight needs the link-target lookup, and the lookup needs the whole harvest pass finished, so nothing can validate an edit while the loop is still on it. Validating once here for the schema-only refusals and again in Phase 1b for the link ones is the second walk. So an in-place edit with both a bad `field_path` and a bad `from_plugin` now reports the source problem and the field-path problem on the re-run — which is what the patch lane already did before this PR, so the two lanes now agree. One CHANGELOG sentence says so (2f9f835e).

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Right, the count and the word disagreed. Folded in 2f9f835e by wording the note as records, which is what `keys.Count` actually is:

    this write links to {keys.Count} distinct record(s) in '{plugin}', whose types were NOT checked: {ex.Message}

I took that over counting tokens because the token count would not be a value count either: `tokens` reaches `LinkTypeLookup` as a `HashSet<string>` (OrdinalIgnoreCase), so a leveled list naming one spell twice with the same spelling is already one token before the count is taken. Counting the write's real FormLink slots would mean carrying the harvest as a list, which is more change than the note is worth. Distinct records is true as written, and it still tells the caller how much of the write went unchecked. The CHANGELOG line now says records too, and a one-line comment at the count says why.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -529,13 +536,28 @@ static PatchOutcome ApplyCore(
                 Key = e.Key, Value = e.Value, Values = e.Values, Entries = e.Entries, Struct = e.Struct, Structs = e.Structs,
             };
             var label = Label(req);
-            if (linkRulebook.Validate(req) is { } reject) { problems.Add($"{recType} {FormIdToken.Of(e.Target)} [{label}]: {reject}"); continue; }
-            resolved.Add((e, body, winnerPlugin, patchLocal, req, label, srcBody));
+            harvestRulebook.CollectLinkValues(req);
```
Avick3110: **Pre-flight now walks every edit twice.**

`CollectLinkValues` is `Validate` itself, and Phase 1b then runs `Validate` again on the same `WriteRequest`. The old `HarvestLinkTokens` prefetch was a plain string sweep with no schema work, so this is a 2x on pre-flight across all three lanes (here, `ApplyInPlaceCore:970`, `CreateRecordsCore:2951`).

Scenario: a bulk `housecarl_apply` with 2000 ops, or a `bulk_create` whose specs carry `composes=` lists, now runs path resolution plus the reflection-backed compose recognizers (`WriteEngine.TryRecognizeCtorArgs` / `TryRecognizeInstantiable`, called per composed element in `StructSpecContents`) 4000 times instead of 2000.

The second walk is provably redundant for any edit that contributed no token. `LinkTypeRefusal` reaches its decision point under exactly the condition in which harvest mode `Add()`s (`value` non-null, not a null synonym, `FormLinkTargetAssemblyQualified` non-null), so if the sink count is unchanged across an edit's harvest, the harvest walk and the checking walk are identical and the harvest's discarded verdict is the pre-flight answer. Recording `linkTokens.Count` before and after each `CollectLinkValues`, keeping the verdict, and skipping Phase 1b's re-walk for edits that added nothing would leave the design intact (one walk still decides both sets) and pay the double cost only for edits that actually carry links.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -961,14 +963,31 @@ static PatchOutcome ApplyInPlaceCore(
                     // Deliberately not refused: copying between two records of the same file is a legitimate job.
                     if (string.Equals(srcPlugin, targetName, StringComparison.OrdinalIgnoreCase)) selfSource = true;
                 }
-                if (CrossTypeRefusal(e, srcBody, body) is { } typeErr) { problems.Add(typeErr); continue; }
+                if (CrossTypeRefusal(e, srcBody, body) is { } typeErr) { Problem(typeErr); continue; }
             }
-            resolved.Add((e, body, req, label, srcBody, selfSource));
+            // Last, as on the patch lane: an edit already rejected above is never pre-flighted, so harvesting its
+            // links would walk a plugin for a check that will not run.
```
Avick3110: **This comment misstates why the harvest sits last.**

"harvesting its links would walk a plugin" is not true of harvesting: `CollectLinkValues` only adds strings to `linkTokens`. Nothing opens a plugin until `LinkTypeLookup` runs in Phase 1b, and it walks a plugin once per plugin however many tokens name it, so a rejected edit's tokens would at most add entries to that one set.

The placement's real consequence is the one the changelog records: an edit with both a bad `field_path` and a bad `from_plugin` now reports the source problem first. A maintainer reading this line while deciding whether the harvest can move back above the `CopyFrom` block is given a cost argument that does not hold, and would restore the pre-PR message order thinking it costs a plugin walk. Say what the placement actually buys instead.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Folded in 6a2ed9b2, on all three lanes.

`CollectLinkValues` now returns the verdict its walk reached instead of discarding it. Each lane records `linkTokens.Count` before and after the harvest of an edit; if the count is unchanged the harvest's verdict stands and Phase 1b does not walk that edit again. The patch and in-place lanes carry the pair on the staged tuple; the create lane keeps a `WriteRequest`-keyed dictionary of the settled verdicts, since its harvest runs before Phase 1. A 2000-op link-free apply is back to 2000 walks; only the edits that actually set a link pay two.

One extra divergence the count had to cover: the create lane hands the harvest `allEditorIds` and the check `priorEditorIds`, so a `@editorid` forward reference is accepted by the harvest and refused by the check. That gate returns before `LinkTypeRefusal`, so it would have added nothing to the sink and been settled on the wrong answer. Harvest mode now also records a same-call sibling ref (`HarvestSibling`, at the three gates whose verdict reads the sibling SET) — it is not a FormID, so `LinkTypeLookup` parses it to nothing and skips it exactly as before, but the count says "re-walk this one".

Test: `LinkHarvestSkipTests` (corpus-only, 5 facts) pins the premise the skip rests on — a link-free write contributes nothing to the sink and gets the same verdict from the harvest walk as from `Validate` (legal and refused both, sentence for sentence), a FormLink write contributes its value, and a `@editorid` write contributes too.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Right — the cost argument was wrong: the harvest only adds strings, and `LinkTypeLookup` walks a plugin once per plugin whatever the token count. Rewritten in bd1a7ab1 to say what the placement actually buys:

    // Last, as on the patch lane, for the MESSAGE ORDER: an edit with both a bad from_plugin and a bad
    // field_path is rejected above and reports the source problem first, which is what the patch lane does.

Same wording as the CHANGELOG line, so a maintainer weighing a move back above the CopyFrom block sees the real consequence.

PR #682, opened 2026-09-09
Delete the 1.x parameter shim, keep the retired tool-name refusal

Avick3110:
This deletes the 1.x parameter shim and keeps the retired tool-name refusal, per Aaron's 2026-09-06 ruling (`dev/DECISIONS.md`, "The alias layer at 2.0.0") written into SPEC §5.4's 2026-09-06 amendment: the parameter shim goes at 2.0.0 as chartered; `AliasTable.RetiredTools` stays, because a call to a 1.x tool name is the error rule applied to an unknown name rather than old-name acceptance — nothing is accepted, redirected or executed, the call is refused with one sentence naming its 2.0 successor. Gone from `AliasTable.cs`: the `Renames` table (the §5.3 old-to-new parameter map — the plugin/plugins/plugin_name/source clique, `patch_name` to `patch`, `operations` to `ops`, and the rest), the `Dissolutions` migration hints, and the lookups feeding them. Gone from `ToolCallShim.cs`: the table half of `ResolveAliases` with its kind gate, the hint the unknown-parameter refusal attached to a retired spelling, and `LaneCorrections`. What is left of the shim is the machinery its own header always called permanent: the underscore/case bridge onto a parameter the tool actually declares (`form_id` for `formid`), shape coercion, and the named refusals for a missing required parameter, an unknown one and a kind that cannot bind. An old spelling now takes the unknown-parameter refusal, which names every parameter the tool does accept. `AliasTable.cs`'s header comment, which said the whole file was scaffolding deleted at 2.0.0, now says the retired-name rows are kept and why.

**`LaneCorrections` is shim, and it goes.** The task was to decide this rather than assume it, so the code: two of its four outcomes rewrite the call and let it execute — `in_place=true` plus a stray `target="X.esp"` became `rewritten["in_place"] = targetFile; rewritten.Remove("target"); p.Arguments = rewritten; return null;`, and a bare `in_place=false` became `rewritten.Remove("in_place"); … return null;`. That is the 1.x spelling translated into the 2.0 one and then run, which is exactly what the ruling deletes; the pass's other two outcomes are the same table teaching old callers their new spelling, in refusal form. Deleting it does not open the consent gate. Nothing rewrites `in_place` any more, so the only way into the overwrite lane is a call that spells the filename: a bare `in_place=true` is now refused by `TypeMismatches` as `in_place (expects string, received boolean)`, and a stray `target=` by `UnknownParameters` as `unknown parameter: target`, with the supported list. Both refuse by name and neither runs the call.

Tests and probes: `AliasActivationTests` (a derived cell per rename and per hint — 88 and 23 rows) and the `alias-layer-guard` probe are deleted with their subject; that probe's own retirement note said its table-driven arms retire with the table, and the mechanism arms it kept back (the bridge, no-clobber, declared-never-rewritten) are already pinned over the wire in `RecordsBindingShimTests`. The wire tests that proved a 1.x spelling bound now prove it is refused by name, and `RetiredNameRedirectTests` gains one test pinning the refusal sentence itself — refused, body did not run, the successor named. Reference sweep: `CheckTools.cs`'s "the alias rows for their names" now says retired-name rows; released CHANGELOG entries describing the alias layer are history and are not rewritten, and SPEC §5.3/§5.4's frozen text is untouched — the 2026-09-06 amendment already states the rule.

Build clean, 1,838 tests pass, ci-all 127/127 (one fewer probe).


src/housecarl-mcp/ToolCallShim.cs
```diff
@@ -37,7 +40,6 @@ internal static class ToolCallShim
                 var schema = tool.ProtocolTool.InputSchema;
                 ResolveAliases(p, schema);
                 CoerceObviousShapes(p, schema);
-                if (LaneCorrections(p, schema) is { } laneRefusal) return laneRefusal;
                 if (MissingRequired(p, schema) is { } refusal) return refusal;
```
Avick3110, line 44: With the rename table gone, a 1.x spelling that stood in for a **required** parameter never reaches `UnknownParameters`, so the caller never gets the supported list.

`MissingRequired` runs first and returns. `housecarl_compact_plugin(plugin_name="Skyrim.esm")` now answers `required parameter missing: plugin. Supplied: plugin_name.` — it never says `plugin_name` is unknown and never prints "This tool accepts only: …". The PR body's claim that "an old spelling now takes the unknown-parameter refusal, which names every parameter the tool does accept" does not hold on this shape, and the new test `TheOnePointXPluginNameSpellingIsRefusedByName_NotMappedOntoCompactPluginsOwnPlugin` pins the weaker wording rather than catching it.

The caller here is exactly the one who needs the accepted-parameter list. Either swap the two passes, or have `MissingRequired` append the supported list when the call also carries undeclared keys.

src/housecarl-mcp/ToolCallShim.cs
```diff
@@ -37,7 +40,6 @@ internal static class ToolCallShim
                 var schema = tool.ProtocolTool.InputSchema;
                 ResolveAliases(p, schema);
                 CoerceObviousShapes(p, schema);
```
Avick3110, line 43: `LaneCorrections` also normalised the **quoted** booleans, and nothing replaces that — so `in_place="false"` now selects the overwrite lane.

`in_place` is declared `string?`, so `Coerce` only converts a string to a boolean when the schema declares `boolean` (line ~160) — it does not here, and the string passes through untouched. `TypeMismatches` sees a string against a string parameter and is happy. The call reaches `ApplyTools`, where `hasInPlace = !string.IsNullOrWhiteSpace(in_place)` is **true** for `"false"`, and `ApplyEdits` goes down the `if (inPlace)` branch with `target = "false"`.

So a 1.x caller writing `housecarl_apply(ops=[…], in_place="false")` — which used to mean the default new-patch lane, and which `LaneCorrections` used to drop — now enters the opt-in overwrite lane and fails as though overwriting a plugin named `false`. Same for `in_place="true"`. It errors rather than writing, but the changelog line "nothing enters the in-place lane from a call that did not spell it" is not true of this shape, and the error names the wrong problem.

If the quoted-boolean guard is meant to go with the rest of the pass, that is a decision to state; otherwise the cheap fix is to refuse a string `in_place` that parses as a bool, in the same sentence `TypeMismatches` gives the bare `in_place=true`.

src/housecarl-mcp/AliasTable.cs
```diff
@@ -1,187 +1,13 @@
...
-    /// <paramref name="ExceptTools"/> suppresses a candidate on tools where that word means something else.</summary>
-    internal readonly record struct Rename(string Old, string[] Candidates, (string Tool, string Candidate)[]? ExceptTools = null);
-
-    /// <summary>The rename rows, in both directions: old callers must keep binding on renamed tools and new
-    /// vocabulary must already bind on not-yet-renamed ones. A parameter that became grammar rather than a new
-    /// name is not a rename — those live in <see cref="Dissolutions"/>.</summary>
-    static readonly Rename[] Renames =
-    {
-        // formid/formids is set-valued on 2.0; symmetric so singular tools accept a plural guess and vice versa.
-        new("formid",  new[] { "formids" }),
```
Avick3110, line 33: Scope check: the **symmetric** rows are not 1.x migrations, and they go out with the migration table.

`formid`↔`formids` (here) and `type`↔`types` (line 51) are listed in *both* directions. A one-way old→new map is a migration; a symmetric pair is a first-guess bridge — exactly what the surviving `ResolveAliases` header says the bridge is for ("so a first-guess miss binds instead of costing a round-trip"). `Normalize` only strips underscores and case, so it cannot carry a plural, and after this PR `housecarl_records(formid="0F1AC1:Skyrim.esm")` — the single most likely miss on a set-valued parameter that reads as singular in every example — costs a refusal and a retry.

The ruling and SPEC §5.4 delete "the §5.3 old-to-new parameter map"; these two rows are arguably not in it. Worth naming the decision explicitly either way rather than letting it ride in with the clique rows — and if they do go, the round-trip is the cost being accepted.

src/housecarl-mcp/ToolCallShim.cs
Avick3110: Folded in a13ea149. `MissingRequired` still runs first, but it now names the call's undeclared keys and the accepted list in the same sentence, off a shared `Undeclared` helper that `UnknownParameters` also uses — so an unknown name is always named, whichever pass returns. `housecarl_compact_plugin(plugin_name="Skyrim.esm")` now answers: `required parameter missing: plugin. Supplied: plugin_name. plugin_name is not a parameter of housecarl_compact_plugin (it accepts only: plugin, esl, …). Add the missing argument and retry.` The test no longer pins the weak wording — it asserts that whole sentence, plus that the accepted list carries the parameter the caller meant.

src/housecarl-mcp/ToolCallShim.cs
Avick3110: Folded in a13ea149. Correct on all of it — the quoted spelling passed the string check and reached `ApplyEdits` with `target = "false"`. A new pass, `InPlaceNamesAFile`, runs between `UnknownParameters` and `TypeMismatches`: a string `in_place` whose value parses as a bool, on a tool that declares `in_place` as a string, is refused by name — `in_place="false" names no file — in_place takes the FILENAME being overwritten (in_place="X.esp"), and "true"/"false" are not files. Omit in_place entirely for the default new-patch lane.` Dormant where `in_place` is declared boolean (compact_plugin), and a real filename is untouched. Tested both spellings, plus the filename case as the negative. The changelog line now says "nothing enters the in-place lane from a call that did not name a file", which is what the gate makes true.

src/housecarl-mcp/AliasTable.cs
Avick3110: Held, no edit. The point stands — a symmetric `formid`/`formids` or `type`/`types` row is a first-guess bridge, not an old-to-new migration, so whether it goes out with the §5.3 table is a scope call rather than a mechanical consequence of the ruling. That is Aaron's to make, and the accepted cost either way (a refusal and a retry on the most likely singular/plural miss) is stated here for him. Leaving the rows out as this PR has them until he rules; if he wants them back, they return as an explicit plural bridge rather than as migration-table residue.

src/housecarl-mcp/ToolCallShim.cs
```diff
@@ -11,10 +11,14 @@ namespace HousecarlMcp;
...
-/// number/bool); refuse a missing required parameter, an undeclared one, or a kind that cannot bind — each by
-/// name. Anything that still throws is caught here rather than above, where the SDK would genericize it.
+/// covered without per-tool wiring. In order: rename an underscore/case variant onto the declared parameter it
+/// names; coerce an unambiguous shape (a bare string for an array, a quoted number/bool); refuse a missing
+/// required parameter, an undeclared one, a quoted boolean where the in-place lane wants a filename, or a kind
+/// that cannot bind — each by name. Anything that still throws is caught here rather than above, where the SDK
+/// would genericize it.
+///
+/// No pass maps a 1.x parameter name onto a 2.0 one: that table was deleted at 2.0.0 (SPEC §5.4 amendment
+/// 2026-09-06), so an old spelling is refused by name like any other unknown parameter.
```
Avick3110, line 21: **The deletion also drops the reverse-direction rows, which are not old-name acceptance.**

`Renames` held two kinds of row. The §5.3 direction (`plugin=` → `source=`, `patch_name=` → `patch=`, `operations=` → `ops=`) is the parameter shim the ruling deletes. The other direction bridged the **2.0** vocabulary onto tools whose parameters were never renamed by a wave — and those go too:

```
new("patch",     new[] { "output", "archivename", "patchname", "pluginname", "into" }),
new("source",    new[] { "plugin", "mod" }),
new("types",     new[] { "type" }),
new("formids",   new[] { "formid" }),
new("ops",       new[] { "operations" }),
new("readback",  new[] { "fullreadback" }),
new("filter",    new[] { "lookup" }),
new("outpath",   new[] { "outputdir", "dest" }),
```

The live surface still carries the split those rows covered:

| Call | Tool declares | Now |
|---|---|---|
| `housecarl_compact_plugin(plugin="X.esp", patch="Out.esp")` | `patch_name` | `unknown parameter: patch` |
| `housecarl_create_plugin(plugin="MyTrigger")` | `plugin_name` | `unknown parameter: plugin` |
| `housecarl_remove(formids=[…], patch="P.esp")` | `into` | `unknown parameter: patch` |
| `housecarl_merge_plugins(patch=…)` | `output` / `patch_name` | `unknown parameter: patch` |
| `housecarl_bsa_extract(out_path=…)` | `output_dir` | `unknown parameter: out_path` |
| `housecarl_load_order_status(filter=…)` | `lookup` | `unknown parameter: filter` |

`patch=` is the 2.0 name for a new artifact on apply / create / copy / forward / place / write_seq, and `AliasTable`'s own kept rows teach it (`patch_name= is patch=`, and the remove row says "`patch=` names a NEW one everywhere else on the surface"). A caller who learned the surface correctly is now refused on six tools. That is not a 1.x spelling being refused — it is the current spelling being refused because six tools were never renamed.

`dev/DECISIONS.md` and the SPEC §5.4 amendment authorize deleting "the §5.3 old→new parameter table". These rows are the opposite direction, so the ruling does not obviously cover them; worth an explicit call rather than riding along.

Related: the CHANGELOG entry names only `plugin=`, `patch_name=` and `operations=`. It does not say that `formid=`, `type=` and `patch=` also stop binding — `housecarl_records(formid="0F1AC1:Skyrim.esm")` is probably the single most-typed call shape on the surface and it is now a refusal.

src/housecarl-mcp/ToolCallShim.cs
```diff
@@ -317,38 +236,61 @@ JsonValueKind.String when bool.TryParse(val.GetString(), out var b) => b,
...
+    }
+
+    /// <summary>A string <c>in_place=</c> whose value spells a boolean gets a named refusal; null to proceed. The
+    /// parameter takes the filename being overwritten, so "true"/"false" — 1.x's lane flag — is not a file, and a
+    /// string one satisfies both the schema type and the tool body's non-empty check: the call would otherwise
+    /// enter the opt-in overwrite lane and fail as though overwriting a plugin named "false".</summary>
+    static CallToolResult? InPlaceNamesAFile(CallToolRequestParams p, JsonElement schema)
+    {
+        if (p.Arguments is not { Count: > 0 } args) return null;
+        if (!args.TryGetValue("in_place", out var val) || val.ValueKind != JsonValueKind.String) return null;
```
Avick3110: **The bare boolean `in_place=true` gets a message that steers the caller into the spelling this pass refuses.**

This guard only fires on `JsonValueKind.String`, so the two spellings of the same mistake take different paths:

- `in_place="true"` → this pass: *"in_place=\"true\" names no file — in_place takes the FILENAME being overwritten (in_place=\"X.esp\") … Omit in_place entirely for the default new-patch lane."* Exactly right.
- `in_place=true` → falls through to `TypeMismatches`: *"parameter whose type could not be bound: in_place (expects string, received boolean). Fix the argument's TYPE to match the schema (… booleans take true/false) and retry."*

The second is the 1.x spelling — the likeliest arrival — and it is the one that gets the message that never mentions filenames. A caller who does what it says ("expects string") retries with `in_place="true"` and is refused again, by a different pass with different wording. Two round trips for the mistake the PR body says is now refused by name.

Widening the pass to `JsonValueKind.True`/`False` alongside `String` would let the one good sentence answer both. It is not a one-line change: line 288's `bool.TryParse(val.GetString(), …)` and the message's `val.GetString()` both assume the string kind, so the boolean arm needs `val.GetRawText()` (and skips the `TryParse` gate, since the kind already settles it).

src/housecarl-mcp-tests/ToolCallShimWirePathTests.cs
```diff
@@ -91,66 +92,66 @@ public void ApplyWithNoArgumentsBindsWithNoRequiredRefusalAndReachesTheBody()
...
-        Assert.True(array.IsError, array.Describe());
-        Assert.Contains("Supplied: plugins", array.Text, StringComparison.Ordinal);
-        Assert.DoesNotContain("could not be bound", array.Text, StringComparison.Ordinal);
-
-        var scalar = _s.Call(ToolNames.CompactPlugin, """{"plugins":"A.esp"}""");
-        Assert.False(scalar.IsError, scalar.Describe());
-        Assert.True(scalar.BodyRan, scalar.Describe());
+        var r = _s.Call(ToolNames.Apply, """{"ops":[],"in_place":"NoSuchPlugin.esp"}""");
+
+        Assert.DoesNotContain("names no file", r.Text, StringComparison.Ordinal);
```
Avick3110, line 210: **The control test does not assert what its doc comment claims.**

The summary says *"it reaches the tool body, which judges the lane"*, but the body asserts only that one substring is absent. Nothing here pins that the call got past the shim.

Drop the `bool.TryParse` guard on `ToolCallShim.cs:288` and reword the refusal, and this test still passes while `housecarl_apply(in_place="MyPatch.esp")` — the whole in-place lane — is refused for every filename. The sibling tests in this file assert `r.BodyRan`; this one should too:

```suggestion
        Assert.DoesNotContain("names no file", r.Text, StringComparison.Ordinal);
        Assert.True(r.BodyRan, r.Describe());
```

src/housecarl-mcp/ToolCallShim.cs
Avick3110: Ruled by Aaron 2026-09-09 (DECISIONS; SPEC §5.4 amended): the reverse-direction rows stay deleted and the six tools take the §5.3 vocabulary outright. The renames (compact_plugin, create_plugin, remove, merge_plugins, bsa_extract, load_order_status, plus any other holdout: patch=, source=, formids=, types=, ops=, readback=, filter=, out_path=) land in their own PR, promoted into 2.0, which merges before this one; this PR is rebased onto it. The CHANGELOG entry here will be widened in that rebase to say which spellings stop binding.

src/housecarl-mcp/ToolCallShim.cs
Avick3110: Folded in 0f301c1a. `InPlaceNamesAFile` now fires on `JsonValueKind.True`/`False` beside `String`: the bool-spelling test is the kind itself for a JSON boolean and `bool.TryParse` only for the string kind, and the message renders the value as the caller spelled it — quoted for a string, `val.GetRawText()` bare for a boolean. So `in_place=true` now takes the same sentence: `in_place=true names no file — in_place takes the FILENAME being overwritten (in_place="X.esp") … Omit in_place entirely for the default new-patch lane.` One round trip, one wording. Ordering is safe: `CoerceObviousShapes` converts number→string but never boolean→string, so the bare kind reaches the pass intact, and the pass still runs before `TypeMismatches`. The theory now runs all four spellings (`"true"`, `"false"`, `true`, `false`), asserts the refusal renders each as spelled, and adds `Assert.DoesNotContain("expects string")` so the type pass cannot quietly take the bare arm back. CHANGELOG reworded to "bare or quoted".

src/housecarl-mcp-tests/ToolCallShimWirePathTests.cs
Avick3110: Folded in 4dbbb984. Added `Assert.True(r.BodyRan, r.Describe());`, so the test now pins what its summary claims — a real filename gets past the shim and reaches the tool body. Confirmed it bites: it is the assertion that would fail if the gate widened to refuse every filename. (Renamed to `AFilenameInPlaceIsNotRefusedByTheBooleanGate`, since the gate is no longer quoted-only.)

Avick3110:
Rebased onto main after #685. Two tests #685 added asserted that `plugin=` and `plugin_name=` still renamed onto `compact_plugin`'s `source=` through the alias clique; with the table deleted they now assert the refusal by name (`required parameter missing: source. Supplied: plugin_name. plugin_name is not a parameter of housecarl_compact_plugin (it accepts only: …)`). The stray-`target=` test passes `source=` so it tests the unknown key, not a missing source. AliasTable's Renames/Dissolutions area resolved to the deletion; the retired-name rows were identical on both sides. Changelog widened: the singular-for-plural bridges (`formid=`, `type=`) go with the table, and a 2.0 word a tool does not declare is refused like any other unknown.

PR #685, opened 2026-09-09
Rename the holdout tools' parameters to the 2.0 vocabulary

Avick3110:
Aaron's 2026-09-09 ruling (dev/DECISIONS.md top line; SPEC §5.4 amendment 2026-09-09) makes the 2.0 parameter vocabulary the only spelling on every tool. The tools no earlier wave renamed still declared 1.x names, so this renames them outright — no bridge rows — and a spelling the schema does not declare is now refused by name. The renames follow the §5.3 table and touch the signatures, the parameter descriptions, every internal reference, the refusal and JSON wording that names a parameter, the probes and shim tests, and the shipped skills.

| Tool | Old | New |
|---|---|---|
| `housecarl_compact_plugin` | `plugin` | `source` |
| `housecarl_compact_plugin` | `patch_name` | `patch` |
| `housecarl_create_plugin` | `plugin_name` | `patch` |
| `housecarl_merge_plugins` | `output` | `patch` (the mod FOLDER; the merged plugin takes its name) |
| `housecarl_merge_plugins` | `patch_name` | deleted |
| `housecarl_bsa_repack` | `archive_name` | `patch` (the mod FOLDER; the `.bsa` takes its name) |
| `housecarl_bsa_repack` | `patch_name` | deleted |
| `housecarl_bsa_extract` | `dest` | `out_path` |
| `housecarl_load_order_status` | `lookup` | `filter` |
| `housecarl_compile_script` | `patch_name` | `patch` |
| `housecarl_compile_script` | `output_dir` | `out_path` |
| `housecarl_decompile_script` | `patch_name` | `patch` |
| `housecarl_nif_set` | `patch_name` | `patch` |
| `housecarl_nif_inspect` | `mod` | `source_provider` |
| `housecarl_nif_set` | `mod` | `source_provider` |
| `housecarl_write_seq` | `output_dir` | `out_path` |

`housecarl_write_seq`'s JSON response member `user_chose_output_dir` becomes `user_chose_out_path`, so the response still names the parameter it reports on.

**`housecarl_remove` is unchanged, deliberately.** The task listed `into` -> `patch` as a possibility. It is not one: §5.1 defines `patch` as the NEW artifact a write creates and `into` as an existing one, and §5.3 keeps `into` (13 tools) unchanged. A removal never creates an artifact — the tool's own doc comment and its no-lane refusal both say so — so `into=` is already the 2.0 word there. §5.3's row listing `patch` for `remove_record` is the 2026-07-29 census of the 1.x tool, which the 2.0 lane redesign has since moved to `into=`.

**The two-name tools, ruled.** `merge_plugins` and `bsa_repack` each declared two new-artifact names: the artifact itself and a separate base name for the mod FOLDER it lands in, which §5.3 could not resolve onto one `patch`. Aaron's 2026-09-09 ruling settles it: `patch=` names the output mod FOLDER on every tool that writes one, these two included, and the artifact inside takes the folder's name the way every other tool derives its plugin's name from `patch=`. `patch="MyMerge"` writes `houseCARL - MyMerge\MyMerge.esp`; `patch="MyArchive"` writes `houseCARL - MyArchive\MyArchive.bsa`, and under `into=` the archive takes the name of the folder it lands in. `patch_name=` is deleted from both, along with the `'<patch> merged'` / `'<patch> renamed'` and `houseCARL_Archive` folder defaults — a repack with no `patch=` names both the folder and the archive after the source folder.

The merge now resolves its output through the record lanes' own `ResolveOutputPath`, so the folder stem, the plugin filename, the auto-suffix and the shadow refusal are the shared ones rather than a second copy; it resolves before the remap, because the remap is keyed on the output ModKey, and a refusal past that point removes the folder again. Two things did not follow the common rule and say so: a `patch=` carrying `.esl` on a merge is still refused rather than stripped to `.esp` like any other extension, because a merge never constrains object ids to the light window and handing back a full plugin silently would be a degraded mode; and `housecarl_compact_plugin`'s output still keeps the SOURCE plugin's exact basename inside the `patch=` folder, because its masters resolve by that name — its artifact name is fixed by something other than `patch=`.

**Not renamed, and why.**
- `housecarl_check`'s `type=`. §5.1 binds `types` to set-valued everywhere; the parameter is a scalar `string?` threaded into the core sweep. Turning it into `types` as a scalar would ship the word with the wrong semantics, and making it set-valued is core plumbing, not a rename. Worth its own issue.
- `bsa_repack`'s `source_folder`; §5.3 marks `from_folder` provisional pending §6's S5 verdict.

`full_readback`, `operations`, singular `formid` and `from_plugin` are declared by no tool — they survive only in `AliasTable` and its probes.

**The nif pair's pole, ruled.** This PR first held `housecarl_nif_inspect` / `housecarl_nif_set`'s `mod=` back: §5.3 said `source`, while §5.5 ships the same provider pole on `housecarl_place` as `source_provider`. Aaron's 2026-09-09 ruling settles it — the provider mod folder is its own pole, `source_provider=`, on all three tools, and `source=` is only ever the plugin to read. Both nif tools now declare `source_provider=`, and each description says it is the same pole `housecarl_place` takes. `nif_inspect`'s tool description was 13 characters over the 2048 clients cut with the longer name, so three phrases in it are shorter.

**The shim.** Per the brief, `ToolCallShim.cs` is untouched and `AliasTable.cs` keeps its rows; only its comments moved, to say that on `merge_plugins` and `bsa_repack` the folder reading and the artifact reading of `patch=` now coincide, so the retired `output=` / `archive_name=` still reach it there while they stay refused on the tools where folder and artifact carry different names. Several rows are now dead: `pluginname -> patch`, `output -> patch`, `archivename -> patch`, `lookup -> filter`, `outputdir`/`dest` -> `outpath`, and the reverse edges that existed only to make the 2.0 word bind on a not-yet-renamed tool. The `mod -> source` row is now wrong on the two nif tools — their `mod=` is the provider pole and `source=` is the plugin — so both are excepted from it and `mod=` is refused by name there. `AliasLayerProbe`'s hand-written `createPluginToday` / `compactToday` / `writeToday` schemas are now fiction about the shipped surface (they still prove the resolver's edges, so they pass). #682 deletes the table and takes all of that with it; merge order is this PR, then #682.

Two shim tests moved with the surface: the create_plugin/compact_plugin clique tests in `ToolCallShimWirePathTests` now pin `patch=`/`source=` reaching the body and the old `plugin=` on create_plugin being refused, and `AliasActivationTests`' derived activation count moves with the surface (88: 90 after the two `patch_name=` deletions, less the two `source= -> mod=` activations the nif rename removes).

Build clean. `dotnet test --filter "tier!=bridge"`: 1964 passed, 0 failed. `ci-all`: 128/128 probes. `scripts/build-plugin.ps1 -PluginTreeOnly`: skill leak-check clean, 10 skills / 221 markdown files.


src/housecarl-mcp/WriteTools.cs
```diff
@@ -77,12 +77,12 @@ public static string CompactPlugin(
         [Description("Optional, default false. Confirms the in-place trade-off when in_place=true OR repoint_externals=true (your original file(s) get rewritten, no backup). The FIRST such call without it returns a CONFIRM prompt listing exactly what will be overwritten — re-call with acknowledge=true to proceed.")]
             bool acknowledge = false,
         [Description("Optional. Base name for the NEW mod folder (new-file lane only; auto-suffixed if taken). Ignored with in_place=true. The PLUGIN inside ALWAYS keeps the source's exact basename so external masters still resolve.")]
-            string? patch_name = null) => Guard.Tool(ToolNames.CompactPlugin, () =>
+            string? patch = null) => Guard.Tool(ToolNames.CompactPlugin, () =>
```
Avick3110, line 80: **The `plugin_name` -> `patch` alias row now misroutes on this tool (and on `compile_script` / `decompile_script` / `nif_set` / `bsa_repack`).**

`AliasTable`'s `pluginname` row is `new("pluginname", new[]{ "patch", "plugins", "plugin", "pluginnames", "source" }, ExceptTools: [(WriteSeq,"patch"), (Place,"patch")])`, and its own comment says why the exception exists: "`patch` is suppressed on the two tools where it names an output FOLDER rather than a plugin ... `plugin_name=` would silently name the folder after a plugin." This PR makes `patch` the output-folder knob on four more tools without extending that list, so the row now fires there. Confirmed against the real published schemas (the `AliasActivationTests` listing):

```
housecarl_compact_plugin:   pluginname -> patch
housecarl_compile_script:   pluginname -> patch
housecarl_decompile_script: pluginname -> patch
housecarl_nif_set:          pluginname -> patch
housecarl_bsa_repack:       pluginname -> patch
```

Concrete misbehaviour:

* `compact_plugin(plugin_name="CoolMod.esp")` — the exact call the deleted `CompactPluginResolvesTheAliasPluginNameOntoItsOwnPluginAndReachesTheBody` test pinned onto the target plugin — now binds to `patch=`, the mod-folder base name, and the call dies on "required parameter missing: source" with nothing saying `plugin_name` was consumed as a folder name.
* `compile_script(script="X.psc", plugin_name="MyMod.esp")` and the `decompile_script` / `nif_set` twins **succeed**, silently cutting `houseCARL - MyMod` as the output folder. Pre-PR these were a named unknown-parameter refusal, because none of the row's candidates were declared on those tools.

Same class, also new on `compile_script`: `output -> patch` and `archivename -> patch` now activate, so `output="C:\Mods\MyMod"` — the natural miss for `out_path=` on this tool — lands on the patch folder name instead of being refused with the supported list.

Fix is in `AliasTable.cs`, not here: add `(CompactPlugin,"patch")`, `(CompileScript,"patch")`, `(DecompileScript,"patch")`, `(NifSet,"patch")` and `(BsaRepack,"patch")` to the `pluginname` row's `ExceptTools`. If the table is genuinely frozen until #682, that is worth saying in the PR body, since the brief's "AliasTable is untouched" was written before these tools grew a `patch=`.

src/housecarl-mcp/BsaTools.cs
```diff
@@ -92,11 +92,10 @@ public static string BsaExtract(
...
 
     /// <summary>How the repack lane names its mod folder, for the into= not-found refusal (#357). It names
-    /// patch_name= and says why: this tool declares archive_name= as well, and the §5.3 candidate order routes a
-    /// bare patch= there — to the .bsa filename, not the folder — so the sibling lanes' patch= sentence would
-    /// rename the caller's archive and leave the folder defaulted.</summary>
+    /// patch_name= and says why: on this tool patch= is the .bsa filename (the artifact the caller came for),
+    /// so the sibling lanes' patch= sentence would rename the caller's archive and leave the folder defaulted.</summary>
     public static readonly LoadOrderService.RiderNaming RepackNaming = new(
-        "patch_name", "On this tool a bare patch= names the ARCHIVE (archive_name=), not the folder.");
+        "patch_name", "On this tool patch= names the ARCHIVE, not the folder.");
```
Avick3110: [nit] This doc was updated but its twin was not. `LoadOrderService.RiderNaming`'s summary (LoadOrderService.cs:7938) still reads "on `housecarl_bsa_repack` a bare `patch=` binds to `archive_name`" — `archive_name` no longer exists on this tool. Same file, `AliasTable.cs:54-62`: the candidate-order comment still says "merge_plugins declares patch_name and output (patch= must reach output), bsa_repack declares patch_name and archive_name (patch= must reach the .bsa). Hence output, then archivename, before patchname." Both tools now declare `patch` outright, so that ordering rationale is describing a surface that no longer exists — and it is the comment the next person will read before touching the row order.

src/housecarl-generator/LoadOrderStatusProbe.cs
```diff
@@ -250,10 +250,10 @@ string MakeInstance(string name, string? baseDir = null)
             {
                 // masterName is "HcLosMaster.esm" — look it up WITHOUT the extension (the bare folder/no-ext case).
                 string noExt = Path.GetFileNameWithoutExtension(masterName);
-                var hit = StatusWire.Render(svc.StatusData(), logs, svc.NamedProfileComposition(null), lookup: noExt, localized: svc.PluginLocalizedFlag(noExt), cap: 80_000);
+                var hit = StatusWire.Render(svc.StatusData(), logs, svc.NamedProfileComposition(null), filter: noExt, localized: svc.PluginLocalizedFlag(noExt), cap: 80_000);
                 Check(hit.Contains("not in the load order") && hit.Contains("Did you mean") && hit.Contains(masterName),
                       $"lookup='{noExt}' (no extension) → the plugin-miss line suggests '{masterName}'");
```
Avick3110: **low - three retired-name strings the rename did not reach.** The diff renamed the *call sites* in these files but left the surrounding prose naming the old parameter, so each is now a teaching row that teaches the retired spelling:

- **This file, line 255** (and 258): the two `StatusWire.Render` calls above were changed `lookup:` -> `filter:`, but the probe's own verdict text still prints ``lookup='{noExt}'`` and "an unrelated **lookup** renders the miss". That string is what `ci-all` prints on a failure, so the next person debugging this probe is handed the one spelling `housecarl_load_order_status` now refuses.
- `src/housecarl-generator/ExtendResolveProbe.cs:40` and `:313` - both still say `patch=` binds to `archive_name` on `bsa_repack`. `archive_name` is no longer declared anywhere; the line 313 comment is the rationale block directly above the checks this PR *did* update at 325-336, so the reasoning and the assertion now disagree about which parameter exists.
- `src/housecarl-mcp-tests/StatusLocalizedLookupTests.cs:8` - class doc opens "load_order_status' `lookup=` reports a plugin's LOCALIZED header flag", while the only call in the file (line 16) was changed to `filter:`.

Same class as the `LoadOrderService.cs:7938` / `AliasTable.cs:54-62` note; grouping them here because none of the three lines is anchorable in this diff except 255.

Checked clean alongside these: `ServerInstructions` (`Program.cs:131-171`) names no renamed parameter; the CHANGELOG entry lists all twelve rows by tool, including the `user_chose_out_path` member and the deliberate `remove` / `patch_name` carve-outs; the 2,048 tool-description ceiling is enforced for every published tool by `PublishedDescriptionBoundTests`, which passes; and on the two dual-name tools the descriptions do separate the roles (`patch=` the artifact, `patch_name=` the folder) with correct defaults for a caller who passes only `patch=` (`'<patch> merged'`, `'houseCARL_Archive'`).

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 01f04493.

`AliasTable` now carries a `FolderPatchTools` list — the tools whose `patch=` names the output mod FOLDER rather than the artifact the old spelling meant: `write_seq`, `place`, `compact_plugin`, `compile_script`, `decompile_script`, `nif_set`. The `output` and `archivename` rows take it as their `ExceptTools`, and the `pluginname` row excepts the same six plus `bsa_repack` (there `patch=` is the .bsa, so a plugin name must not reach it, while `output=`/`archive_name=` correctly do).

Result on the activation sweep: 104 -> 88. The 16 that went away are `output -> patch` and `archivename -> patch` on the six folder tools, and `pluginname -> patch` on `compile_script`, `decompile_script`, `nif_set` and `bsa_repack`. `compact_plugin: pluginname` did not vanish — it now falls through the clique to `source`, so `compact_plugin(plugin_name="CoolMod.esp")` reaches the body again instead of dying on "missing source". `merge_plugins: pluginname -> patch` is kept: there `patch=` is the merged plugin.

Test: `ToolCallShimWirePathTests.PluginNameNeverBindsToTheOutputFolderPatchOnTheRiderTools` drives both arms over the wire — `plugin_name=` on `compact_plugin` reaches the body, and on `compile_script` (which declares no plugin spelling at all) it is the named unknown with the supported list.

Changelog line added under Unreleased.

src/housecarl-mcp/BsaTools.cs
Avick3110: Folded. `AliasTable.cs`'s candidate-order comment is rewritten in 01f04493 — it now says merge_plugins and bsa_repack declare `patch` outright so the retired `output=`/`archive_name=` reach it directly, and states the new suppression on the folder-patch tools. `LoadOrderService.RiderNaming`'s summary is fixed in 8f75e622: on `bsa_repack` `patch=` names the .bsa itself, so pointing that caller at `patch=` for the folder would rename their archive.

src/housecarl-generator/LoadOrderStatusProbe.cs
Avick3110: Folded, all three, in 8f75e622:

- `LoadOrderStatusProbe.cs` 255/258: the verdict text is `filter='{noExt}'` and "an unrelated filter renders the miss".
- `ExtendResolveProbe.cs` 40 and 313: both now say `patch=` names the .bsa itself on `bsa_repack` (it declares both `patch=` and `patch_name=`, and §5.3 routes `patch=` to the artifact), so the rationale and the checks at 325-336 agree again.
- `StatusLocalizedLookupTests.cs:8`: class doc says `filter=`.

`ci-all` is 128/128 and the suite is 1958/1958 after the change.

Avick3110:
Folded the report's fourth finding (no anchorable line) in b2face21: `DialogueTools.cs` line 289 told the caller to run `housecarl_write_seq plugin=<Plugin>` in the dialogue-sweep SEQ remedy, but that tool declares `source=`. Fixed. Nothing pins that string — grepped for the sentence across tests and probes and it appears only at the emit site.

Avick3110:
Aaron's 2026-09-09 ruling landed here, so the open point in the body ("what §5.3 could not resolve") is closed rather than argued: `patch=` names the output mod FOLDER on every tool that writes one, `housecarl_merge_plugins` and `housecarl_bsa_repack` included, and the artifact inside takes that folder's name. `patch="MyMerge"` writes `houseCARL - MyMerge\MyMerge.esp`; `patch="MyArchive"` writes `houseCARL - MyArchive\MyArchive.bsa`, and with `into=` the archive takes the name of the folder it lands in.

`patch_name=` is deleted from both tools, along with the `'<patch> merged'` / `'<patch> renamed'` and `houseCARL_Archive` folder defaults — a repack with no `patch=` names both the folder and the archive after the source folder. The merge derives its output through the record lanes' own `ResolveOutputPath`, so the folder stem, the plugin filename, the auto-suffix and the shadow refusal are the shared ones rather than a second copy; it resolves before the remap because the remap is keyed on the output ModKey, and a refusal past that point removes the folder again. The rider folder now carries its stem, which is what the `.bsa` is named after. The `RiderNaming` caveat (bsa_repack was its only user, saying `patch=` named the archive) and the shadow check's "the plugin does not follow the stem" branch (merge_plugins was its only user) go with the change.

Two places the derivation deliberately does not follow the common rule, both stated in the body: a `patch=` carrying `.esl` on a merge is still refused rather than stripped to `.esp` like any other extension, since a merge never constrains object ids to the light window; and `housecarl_compact_plugin`'s output still keeps the source plugin's exact basename inside its `patch=` folder, because its masters resolve by that name.

Build clean; `dotnet test --filter "tier!=bridge"` 1966 passed, 0 failed; `ci-all` 128/128; `scripts/build-plugin.ps1 -PluginTreeOnly` clean (10 skills, 221 markdown files).

Avick3110:
Aaron's 2026-09-09 ruling landed: the provider mod folder is its own pole, so `housecarl_nif_inspect` and `housecarl_nif_set` now take `source_provider=` instead of `mod=` — the same pole `housecarl_place` already declares, with `source=` left as only ever the plugin to read. Each description says so in a sentence; the refusals were already spelling `source_provider=` through the shared source sentences.

The `mod -> source` alias row is wrong on these two tools now, so both are excepted from it: `mod=` there is refused by name rather than bound to a plugin pole. The derived rename-activation count drops 90 -> 88, losing the two `source= -> mod=` activations the rename removes.

`nif_inspect`'s tool description crossed the 2048-character client cut with the longer name, so three phrases in it are shorter.

Build clean; `tier!=bridge` 1964 passed, 0 failed; ci-all 128/128; `build-plugin.ps1 -PluginTreeOnly` clean. Pushed 14fe5c1a, 9291639a, 1e7fb27c.

src/housecarl-mcp/BsaTools.cs
```diff
@@ -127,15 +123,17 @@ public static string BsaRepack(
...
+        // A caller who spells the archive itself — patch="MyArchive.bsa" — means that name, so the extension is
+        // stripped here rather than folded into the folder name and doubled on the file.
         LoadOrderService.RiderFolder rf;
-        try { rf = svc.ResolvePatchModFolder(patch_name, into, "houseCARL_Archive", RepackNaming); }
+        var stem = patch?.Trim().Trim('"');
+        if (stem is not null && stem.EndsWith(".bsa", StringComparison.OrdinalIgnoreCase)) stem = stem[..^4];
+        try { rf = svc.ResolvePatchModFolder(stem, into, new DirectoryInfo(source_folder).Name, RepackNaming); }
         catch (InvalidOperationException ex) { return "error: " + ex.Message; }
         var folder = rf.OutputDir;
+        var name = rf.Stem + ".bsa";
```
Avick3110, line 151: **Repack `into=` now silently overwrites the folder's existing `.bsa`.**

The archive name is `rf.Stem + ".bsa"`, and on the `into=` lane `rf.Stem` is the destination folder's own name — so every repack into the same folder produces the same filename.

- `bsa_repack(source_folder="C:/work/MeshesA", into="MyPatch.esp")` writes `houseCARL - MyPatch` / `MyPatch.bsa`.
- `bsa_repack(source_folder="C:/work/TexturesB", into="MyPatch.esp")` writes `MyPatch.bsa` again, over the first one. There is no existence check before `BsaArchive.Pack`, and `PackReport` says only "packed N file(s) into MyPatch.bsa" — no "replaced", no backup. The first archive's contents are gone.

On `main` those two calls produced `MeshesA.bsa` and `TexturesB.bsa` side by side.

`patch=` is not an escape: `ResolvePatchModFolder` takes the `into=` branch before it ever reads `patchName`, and `BsaRepack` declares no `patch=`/`into=` exclusivity check, so `patch="TexturesB"` is accepted and silently ignored. The alias row maps a retired `archive_name=` onto `patch`, so an old caller who used `archive_name=` + `into=` to name a second archive in one patch folder now gets the overwrite too.

Worth either refusing `patch=` given with `into=` (the way `write_seq` refuses the pair), or refusing an existing archive at the resolved path rather than clobbering it.

src/housecarl-mcp/BsaTools.cs
```diff
@@ -127,15 +123,17 @@ public static string BsaRepack(
...
-
+        // patch= names the mod FOLDER and the .bsa inside takes that folder's name, as on every other tool that
+        // writes one — including the into= case, where the archive is named for the folder it lands in.
+        // A caller who spells the archive itself — patch="MyArchive.bsa" — means that name, so the extension is
+        // stripped here rather than folded into the folder name and doubled on the file.
         LoadOrderService.RiderFolder rf;
-        try { rf = svc.ResolvePatchModFolder(patch_name, into, "houseCARL_Archive", RepackNaming); }
+        var stem = patch?.Trim().Trim('"');
+        if (stem is not null && stem.EndsWith(".bsa", StringComparison.OrdinalIgnoreCase)) stem = stem[..^4];
+        try { rf = svc.ResolvePatchModFolder(stem, into, new DirectoryInfo(source_folder).Name, RepackNaming); }
```
Avick3110, line 148: **An auto-suffixed `patch=` stem now renames the `.bsa` itself, producing an archive the game will not load.**

Before this change the folder stem and the archive name were independent: `UniqueStem` could suffix the folder while `archive_name=` pinned the file. Now `name = rf.Stem + ".bsa"`, so the suffix lands on the archive.

With a mod folder `houseCARL - MyMod` already present (an earlier repack, say), `bsa_repack(source_folder=..., patch="MyMod")` gets `freeStem = "MyMod_001"` and writes `MyMod_001.bsa`. Skyrim SE auto-loads only `<activePluginBasename>.bsa` / `<activePluginBasename> - Textures.bsa`, or archives named in `sResourceArchiveList`, so `MyMod_001.bsa` is inert — yet `PackReport` still closes with "(a new houseCARL mod folder — enable it in MO2 to use the archive.)", which is not true of that file.

The rename also removes the only way to guarantee the archive's name, which is load-bearing here in a way a patch folder's name is not. Either say in the report when the stem was suffixed (and what that costs), or keep a way to spell the `.bsa` exactly.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7236,8 +7246,28 @@ public WritePatchBuilder.MergeOutcome MergePlugins(
                     return WritePatchBuilder.MergeOutcome.Fail(keyErr!);
                 donorKeys.Add((dName, keys));                             // a pure-override donor (0 originating keys) is a legit patch donor
             }
+            // ---- output folder and plugin: the same fresh-write resolver the record lanes use, so patch= names the
+            //      mod folder "houseCARL - <stem>" and the merged plugin inside it is "<stem>.esp". Resolved HERE,
+            //      before the remap, because the remap is keyed on the output ModKey and the stem can still be
+            //      auto-suffixed by a mod folder of that name (an active plugin of it was refused above). ----
+            string outPath;
+            bool createdFolder;
+            try { outPath = ResolveOutputPath(patchName, into: null, out _, out createdFolder); }
```
Avick3110: **A folder-only collision now silently renames the merged PLUGIN, not just its folder.**

Routing the merge through `ResolveOutputPath` means `UniqueStem` decides the plugin filename, and its first test is `IsStemFree`, which fails on a bare mod-folder collision before the shadow check ever runs.

Concretely: a leftover `houseCARL - MyMerge` folder exists (a previous merge whose plugin was reviewed and deleted, or any user folder of that name) but no `MyMerge.esp` is active. `merge_plugins(plugins=[...], patch="MyMerge")` passes the `view.ContainsPlugin("MyMerge.esp")` pre-flight, then `UniqueStem` returns `MyMerge_001`, and the merged plugin is written as `MyMerge_001.esp`. Anything keyed to the requested basename — a `_DISTR.ini`, a `_KID.ini`, a config JSON bound by plugin name, another plugin listing it as a master — misses.

Two things make this worth a second look rather than just a documented consequence:

- `patch=`'s own description says a colliding output name "is refused loud and nothing is written", including "a name that already exists somewhere your order is not loading it". The folder-collision arm does the opposite, and it also short-circuits the very shadow refusal that sentence promises.
- `create_plugin`, the other tool whose `patch=` names an artifact rather than only a folder, refuses a folder collision loudly for exactly this reason ("houseCARL won't auto-rename ... its exact basename is what makes the trigger resolve"). A merge output's basename is load-bearing in the same way, and the tool's own docs say so.

Refusing, or at least naming the rename in the outcome, would keep the two `patch=`-names-an-artifact tools consistent.

src/housecarl-mcp/BsaTools.cs
Avick3110: Folded, both halves. `patch=` given with `into=` now refuses as a pair by name, before the BSArch path is even resolved (the write_seq wording: the archive takes one folder's name and the two lanes name different folders). And an archive already at the resolved path refuses in one sentence naming the file — no clobber, no backup — so a second repack into the same folder says so instead of losing the first archive. `f24d4b1e`, tests in `PatchArtifactCollisionTests`.

src/housecarl-mcp/BsaTools.cs
Avick3110: Folded — kept as a refusal rather than a report line. A taken `patch=` on bsa_repack (a mod folder of that name, or an active plugin of that basename) is now refused by name and nothing is written, the way create_plugin refuses, so the stem never lands on the archive as a suffix. That makes the report's "enable it in MO2 to use the archive" always true of the file it names. The remedy sentence points at the three ways out: remove the folder, choose another name, or `into=`. `f24d4b1e`.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded — merge now refuses a folder collision by name, consistent with create_plugin and with what `patch=`'s description promises, so the merged plugin is never written under a suffix a `_DISTR.ini` or a master entry cannot resolve. Done as a mode on `ResolveOutputPath` (a `StemRefusal` the lane hands in, carried down to `UniqueStem`) rather than a second resolver, so the derivation stays shared and bsa_repack takes the same mode through its `RiderNaming`. The shadow refusal still runs: it is only reached when the stem is otherwise free, which is unchanged. `7749d7ad`.

PR #686, opened 2026-09-09
Make the check sweep's record-type scope a set

Avick3110:
Aaron's ruling of 2026-09-09 (dev/DECISIONS.md): `housecarl_check`'s scalar `type=` becomes the set-valued `types=` per the 2.0 spec's §5.1 rule that every selector is set-valued and one value is a set of one, core plumbing included. This does that. The parameter now takes a list of record type signatures or catalog names — the same value form and the same lookup `housecarl_records`' `types=` takes — and a sweep over N types is the sweep over each with the findings merged: the resolved types are the union of the entries, handed to the record stream exactly as the single type was, so the types you did not name still cost nothing per skipped record. A single value still works as a set of one, an unknown entry is still refused by name, and the response's narrowing line now names every type in the set. The union resolve itself is one helper (`ResolveTypeFilterSet`) that the records scan's two call sites and the sweep scope now share, so there is one grammar rather than three copies of the same loop. The parameter description, the `check_errors` alias row and the merged tool's scope lines are updated to match; no skill, eval or doc spelled `check(... type=`.

Tests: a two-type sweep reporting both types' findings, a one-type sweep still answering over that type alone, the narrowing line naming both types, and an unknown entry refusing the sweep by name. The type-filter world gains one NPC_ and one WEAP with a dangling link each, since its globals carry no links for a sweep to find.

Build, `dotnet test --filter "tier!=bridge"` (1967 passed), `ci-all` (128/128) and `scripts/build-plugin.ps1 -PluginTreeOnly` all clean.

Note: PR #685 renames other tools' parameters and does not touch check; nothing here depends on it.

### DIFF
--- a/src/housecarl-core/SweepScope.cs
+++ b/src/housecarl-core/SweepScope.cs
@@ -45,6 +45,13 @@ public sealed class SweepScope
         TypeLabel = Types is null ? null : typeLabel;
     }
 
+    /// <summary>The scope's types spelled for the response, but ONLY where the scope covers more than one — the
+    /// listing budget is one counter spent in the order these are streamed, so a multi-type scope can list the
+    /// first type and never reach the second. Null where one type (or none) is in force and that cannot happen.
+    /// <para>More than one TYPE, not more than one entry: one entry can expand to several arms (<c>Global</c> to
+    /// its int/float/short arms), and the budget is spent arm by arm there too.</para></summary>
+    public string? TypeOrder => Types is { Count: > 1 } ? TypeLabel : null;
+
     /// <summary>True when nothing is actually narrowed (every knob absent) — the caller can then pass null and keep the
     /// unscoped path byte-identical.</summary>
     public bool IsEmpty => Formids is null && EditorIdContains is null && Types is null;
--- a/src/housecarl-core/SweepScope.cs
+++ b/src/housecarl-core/SweepScope.cs
@@ -8,9 +8,8 @@ namespace HousecarlCore;
 /// The RECORD-level narrowing the sweep families (housecarl_check findings=["errors"] and ["scripts"]) share.
 /// A plugin-wide sweep of a script-heavy plugin renders past the tool-result token cap, so record-level scoping is
 /// what makes "did the unbound count for THESE few records change?" askable. The knobs deliberately reuse the sibling
-/// read tools' vocabulary: <see cref="Types"/> (cross_plugin_query's <c>type=</c>), <see cref="Formids"/>
-/// (batch_record_detail's <c>formids=</c>), and <see cref="EditorIdContains"/> (cross_plugin_query's
-/// <c>editorid_contains=</c>).
+/// read tools' vocabulary: <see cref="Types"/> (the records surface's set-valued <c>types=</c>), <see cref="Formids"/>
+/// (<c>formids=</c>), and <see cref="EditorIdContains"/> (<c>editorid_contains=</c>).
 ///
 /// <para><b>Narrowing narrows the NUMBERS.</b> A scoped sweep's totals are the totals FOR THE SCOPE, exactly as
 /// <c>plugins=</c> already behaves. The render therefore carries <see cref="Label"/> on its own line whenever
@@ -30,11 +29,11 @@ public sealed class SweepScope
     /// free pass).</summary>
     public string? EditorIdContains { get; }
 
-    /// <summary>The getter Type(s) to stream, or null for every record type. Resolved by the caller from the user's
-    /// <c>type=</c> string (the shared TypeLookup), so an unknown type fails loud before the sweep starts.</summary>
+    /// <summary>The getter Type(s) to stream, or null for every record type. Resolved by the caller as the UNION of
+    /// the user's <c>types=</c> set (the shared TypeLookup), so an unknown type fails loud before the sweep starts.</summary>
     public IReadOnlyList<Type>? Types { get; }
 
-    /// <summary>The user-facing spelling of <see cref="Types"/> (the raw <c>type=</c> string), for <see cref="Label"/>.</summary>
+    /// <summary>The user-facing spelling of <see cref="Types"/> (the raw <c>types=</c> entries), for <see cref="Label"/>.</summary>
     public string? TypeLabel { get; }
 
     public SweepScope(IReadOnlySet<FormKey>? formids, string? editorIdContains,
@@ -71,7 +70,7 @@ public sealed class SweepScope
         {
             if (IsEmpty) return null;
             var parts = new List<string>(3);
-            if (Types is not null) parts.Add($"type={TypeLabel}");
+            if (Types is not null) parts.Add($"types=[{TypeLabel}]");
             if (Formids is not null) parts.Add($"{Formids.Count} formid(s)");
             if (EditorIdContains is not null) parts.Add($"editorid_contains='{EditorIdContains}'");
             return string.Join(", ", parts);
@@ -79,10 +78,10 @@ public sealed class SweepScope
     }
 
     /// <summary>An OFF-ORDER file's record stream, type-scoped when the caller asked for one — the overlay
-    /// counterpart of <c>RecordsIn</c>'s getter-type filter, so a <c>type=</c> scope costs nothing per skipped
+    /// counterpart of <c>RecordsIn</c>'s getter-type filter, so a <c>types=</c> scope costs nothing per skipped
     /// record on that lane either. One home, because both sweep families walk an off-order file the same way.
     /// Through <see cref="RecordArms"/>, the same arm re-check the in-order lanes go through: without it an arm scope
-    /// (<c>type='GlobalShort'</c>) would sweep and count the whole GRUP here.</summary>
+    /// (<c>types=['GlobalShort']</c>) would sweep and count the whole GRUP here.</summary>
     public static IEnumerable<IMajorRecordGetter> RecordsFrom(ISkyrimModGetter ov, SweepScope? scope)
         => scope?.Types is { Count: > 0 } types
             ? RecordArms.OfTypes(ov, types)
--- a/src/housecarl-core/SweepScope.cs
+++ b/src/housecarl-core/SweepScope.cs
@@ -36,21 +36,44 @@ public sealed class SweepScope
     /// <summary>The user-facing spelling of <see cref="Types"/> (the raw <c>types=</c> entries), for <see cref="Label"/>.</summary>
     public string? TypeLabel { get; }
 
+    // The same entries with every one that EXPANDED spelling its arms ("GMST → GameSettingInt, ..."), for
+    // TypeScopeLabel. Null where the caller built no such spelling, and TypeLabel then stands in.
+    readonly string? _armLabel;
+
     public SweepScope(IReadOnlySet<FormKey>? formids, string? editorIdContains,
-                      IReadOnlyList<Type>? types, string? typeLabel)
+                      IReadOnlyList<Type>? types, string? typeLabel, string? armLabel = null)
     {
         Formids = formids is { Count: > 0 } ? formids : null;
         EditorIdContains = string.IsNullOrWhiteSpace(editorIdContains) ? null : editorIdContains.Trim();
         Types = types is { Count: > 0 } ? types : null;
         TypeLabel = Types is null ? null : typeLabel;
+        _armLabel = Types is null ? null : armLabel;
     }
 
-    /// <summary>The scope's types spelled for the response, but ONLY where the scope covers more than one — the
-    /// listing budget is one counter spent in the order these are streamed, so a multi-type scope can list the
-    /// first type and never reach the second. Null where one type (or none) is in force and that cannot happen.
-    /// <para>More than one TYPE, not more than one entry: one entry can expand to several arms (<c>Global</c> to
-    /// its int/float/short arms), and the budget is spent arm by arm there too.</para></summary>
-    public string? TypeOrder => Types is { Count: > 1 } ? TypeLabel : null;
+    /// <summary>The scope's types spelled for the short-listing rule, but ONLY where the scope covers more than one
+    /// — there is one listing for the whole scope, so with several types in it any one of them can come out short.
+    /// Null where a single type is in force and that cannot happen.
+    /// <para>More than one TYPE, not more than one entry: one entry can expand to several arms (<c>GMST</c> to its
+    /// int/float/string/bool arms), and the listing is shared across those arms too. Which is why this spells the
+    /// arms an entry expanded to: a rule about "any of those types" has to name types the reader can act on, and
+    /// <c>types=[GMST]</c> alone names none.</para></summary>
+    public string? TypeScopeLabel => Types is { Count: > 1 } ? (_armLabel ?? TypeLabel) : null;
+
+    /// <summary>One <c>types=</c> entry spelled with the arms it resolved to, where it resolved to more than one —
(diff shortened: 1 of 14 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -5242,12 +5222,14 @@ ErrorCheckResult ClassifyMissingMasters(ErrorCheckResult r)
             }
         }
 
+        // The set is the union of its entries, resolved the way records= resolves types= — one type is a set of one.
         IReadOnlyList<Type>? types = null;
-        var typeLabel = type?.Trim();
-        if (!string.IsNullOrEmpty(typeLabel))
+        string? typeLabel = null;
+        if (typeSet is { Count: > 0 })
```
Avick3110, line 5231: The scalar path treated a blank type as "no narrowing" (`if (!string.IsNullOrEmpty(typeLabel))`); the set path drops that. `typeSet is { Count: > 0 }` is true for `types=[""]` or `types=[" "]`, so `ResolveTypeFilterSet` reaches `ResolveTypeFilter("")` and the whole call is refused with `unknown record type ''. Expected a 4-char signature (e.g. 'WEAP')...`.

This is reachable from an old caller, not just a hand-typed empty string: the alias layer renames `type` onto `types` and `ToolCallShim.Coerce` wraps a bare string, so a client that sends declared-but-unset optional params as `type=""` goes from a full unscoped sweep to a refusal whose sentence names no type at all. Either skip blank entries when building the union, or refuse them with a sentence that says a `types=` entry was blank.

src/housecarl-mcp/CheckTools.cs
```diff
@@ -61,8 +61,8 @@ public static string CheckTool(
              "sweep the WHOLE active order — thorough but heavier; scope to one plugin for a fast, focused check " +
              "like the CK's per-plugin 'Check For Errors'.")]
             string[]? plugins = null,
-        [Description("Optional. Record type to sweep — a 4-char signature ('WEAP') or catalog name ('Weapon'). Applied at the record STREAM, so it is the CHEAPEST scope: skipped records cost nothing (no link walk, no .pex chain read). An unknown type is refused, naming what is expected.")]
-            string? type = null,
+        [Description("Optional. Record types to sweep — signatures ('WEAP') or catalog names ('Weapon'); one type is a set of one, and the sweep is the sweep over their UNION with the findings merged. Applied at the record STREAM, so it is the CHEAPEST scope: skipped records cost nothing (no link walk, no .pex chain read), and a two-type sweep costs the two type groups, not the order. An unknown type is refused by name, naming what is expected.")]
```
Avick3110, line 64: The union is streamed one type at a time (`RecordArms.OfTypes` = `SelectMany` over the resolved types), so findings come out grouped by the set's entry order, and `limit=` is one shared listing budget spent in that order. `types=["Npc","Weapon"] limit=20` on an order with 50 dangling NPC refs lists 20 NPC findings and zero WEAP ones, while the narrowing line says `types=[Npc, Weapon]` — reading as "weapons are clean". The totals stay honest (uncapped), so this is a listing-order confusion rather than a wrong count, but it is newly easy to hit now that the caller composes the set: worth a word in this description that limit= is spent in the order the types are named.

src/housecarl-mcp/SweepSharedInput.cs
```diff
@@ -25,9 +25,9 @@ internal static class SweepSharedInput
 
     /// <summary>The whole call's refusal where a shared input is malformed, or null where every one of them parses.
     /// Checked in the order a caller reads the parameters, so a call with two bad values names the first.</summary>
-    /// <param name="svc">the service, for the record-scope parse — <c>type=</c> resolves through the same
-    /// TypeLookup <c>cross_plugin_query</c> uses, which lives there.</param>
-    internal static string? Error(LoadOrderService svc, IReadOnlyList<string>? plugins, string? type,
+    /// <param name="svc">the service, for the record-scope parse — <c>types=</c> resolves through the same
```
Avick3110, line 28: The `<param>` doc was updated here, but the class doc two paragraphs up (line 9) still reads "The families parse `type=`, `formids=`, `plugins=` and `exclude=` themselves" — the one spelling of the parameter this PR retires that survives in the file that owns the shared-input rule.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in c8378cc5. The grammar decides it, and records was refusing it the same way the sweep now does not: `housecarl_records types=[""]` goes through the same `ResolveTypeFilterSet` -> `ResolveTypeFilter("")` and came back `unknown record type ''` — a sentence that names no type and does not say what is wrong. So the fix is one rule in the one place both surfaces resolve through: `ResolveTypeFilter` refuses a blank entry as blank ("a blank record type — pass a 4-char signature (e.g. 'WEAP') or a catalog name (e.g. 'Weapon'), or omit the parameter to leave the types unnarrowed."). An entry that names no type is refused on every surface, saying it is blank; an empty or absent SET stays the separate thing it is (no narrowing), decided by the callers. Skipping blanks silently was the other option and it is the wrong one here: `type=""` from an old caller would then silently become a full unscoped sweep, which is the same silent widening the refusal exists to stop. Two tests: the sweep refuses a blank entry saying blank and not `unknown record type`, and the records surface refuses the identical input with the identical sentence.

src/housecarl-mcp/CheckTools.cs
Avick3110: Folded in f78416d7 — the cut notice, not a shared budget. Which of the two, and why this shape:

The per-type counts are not there to name. The sweep tallies dangling refs by SOURCE plugin and by TARGET plugin and never by type, so "20 of 43 Npc listed, 0 of 7 Weapon" would mean a new by-type axis through the accounting (a second roster, its own worst-case reserve, both transports). Sharing the budget per type is larger still — it restructures the phase order the baseline split already owns. So the notice states the RULE, which is the part the reader is missing, and states no number the response does not have.

What it says, whenever a scope covering more than one type had findings the listing budget dropped: `limit= is ONE listing budget, spent in the order the scope's types are streamed (types=[Npc, Weapon]), so a type later in that order can be absent from the listing above: absent there means unlisted, not clean. Sweep that type alone, or raise limit=.` It fires on both swept families (both spend one budget over one type stream), and only where something was actually dropped — with the listing complete for every type in the scope, the warning would be about a hole that is not there. `limit=`'s own description now carries the same rule.

Gated on the scope's resolved TYPE count, not on the entry count, so `types=["Global"]` — one entry, three arms — is covered too: the budget is spent arm by arm there as well.

Five tests: a two-type sweep with room for one finding lists the NPC_, keeps the total at 2 and states the rule; an uncut two-type sweep does not state it; a cut ONE-type sweep does not state it (the rule is about the set, not the cut); and the json twin carries `limit_spent_in_type_order` under the same test, so the transports cannot disagree.

src/housecarl-mcp/SweepSharedInput.cs
Avick3110: Fixed in f93ce0c6 — the class doc now reads `types=`.

src/housecarl-mcp/ReadSentences.cs
```diff
@@ -277,6 +277,16 @@ internal static int ClauseReserve(bool mayState) =>
     [MustState("not named here")]
     internal const string SweepRosterCut = " (the {0} largest of {1}; the rest are not named here)";
 
+    /// <summary>The type-scope rule, stated wherever a MULTI-TYPE scope was in force and the listing budget dropped
+    /// something. The budget is one counter spent in the order the scope's types are streamed, so a type later in
+    /// that order can be missing from the listing entirely — and a listing that names two types while showing one
+    /// otherwise reads as "the other one is clean".</summary>
+    [MustState("limit=", "unlisted", "not clean")]
+    internal const string SweepTypeOrderRule =
```
Avick3110: **The budget is not spent in type order — it is spent plugin-major.**

This sentence tells the caller `limit=` is "spent in the order the scope's types are streamed". The sweep does not do that. `ErrorCheck.Run` calls `SweepActivePlugin` once per plugin — phase 1 over every non-base plugin (`ErrorCheck.cs:335`), phase 2 over the base masters (`ErrorCheck.cs:446`) — and each call streams `RecordsIn(new[]{plugin}, recordScope?.Types)`. Types are only ordered *within* one plugin. `ScriptPropertyCheck.cs:331` has the same shape.

Scenario: `check(types=["Npc", "Weapon"], limit=1000)` over the whole order, where an early `ModA.esp` carries 1000 WEAP dangling refs and a later `ModB.esp` carries the NPC_ ones. The budget is drained entirely on ModA's weapons and ModB's NPC_ findings are never listed — yet the response says the budget was spent in the order `types=[Npc, Weapon]`. The reader concludes Npc was drained first and is therefore complete, re-sweeps Weapon alone, and never sees the Npc findings. The json twin carries the same claim as `limit_spent_in_type_order` (`CheckAccounting.cs:550`), and the `limit=` parameter description and the CHANGELOG entry repeat it.

The honest rule for a multi-plugin sweep is about the *stream* order, not the type order: the budget is spent plugin by plugin (non-base first, base masters last) and type by type inside each, so any type can be short. Either say that, or drop the ordering claim and keep only the part that is true — a type absent from the listing is unlisted, not clean.

src/housecarl-core/SweepScope.cs
```diff
@@ -46,6 +45,13 @@ public SweepScope(IReadOnlySet<FormKey>? formids, string? editorIdContains,
         TypeLabel = Types is null ? null : typeLabel;
     }
 
+    /// <summary>The scope's types spelled for the response, but ONLY where the scope covers more than one — the
+    /// listing budget is one counter spent in the order these are streamed, so a multi-type scope can list the
+    /// first type and never reach the second. Null where one type (or none) is in force and that cannot happen.
+    /// <para>More than one TYPE, not more than one entry: one entry can expand to several arms (<c>Global</c> to
+    /// its int/float/short arms), and the budget is spent arm by arm there too.</para></summary>
+    public string? TypeOrder => Types is { Count: > 1 } ? TypeLabel : null;
```
Avick3110: **A one-entry polymorphic base fires the rule with a label naming one type.**

The gate is on the count of *resolved* getter types, and `BuildTypeLookup` (`LoadOrderService.cs:9188-9199`) maps an abstract-group base name or signature to its concrete arms — `RecordsTypeArmTests` proves `types=["GLOB"]` resolves to GlobalShort/GlobalInt/GlobalFloat, and GMST resolves to its arms the same way. The doc-comment says this is deliberate, but the sentence it renders is not written for that case.

Scenario: `check(types=["GMST"], limit=50)` on an order with more than 50 GMST findings gives `Types.Count == 4` and `TypeOrder == "GMST"`, so the response says:

> limit= is ONE listing budget, spent in the order the scope's types are streamed (types=[GMST]), so a type later in that order can be absent from the listing above: absent there means unlisted, not clean. Sweep that type alone, or raise limit=.

The caller named exactly one type and already swept it alone, the sentence names no second type, and the one question the rule exists to answer — which type got crowded out — cannot be read off it. Either gate on the entry count, or spell the resolved arms in the label so "a type later in that order" names something.

src/housecarl-mcp/CheckAccounting.cs
```diff
@@ -292,6 +299,13 @@ bool Short(Values v, SweepSubject s)
 
     int Shown(Values v, SweepSubject s) => v.Emitted.TryGetValue(s, out var e) ? e : 0;
 
+    /// <summary>Does this rendering have to state the type-order rule? Only where a multi-type scope was in force
+    /// AND this family's listing budget actually dropped findings — with nothing dropped, the listing IS the whole
+    /// answer for every type in the scope and the rule would warn about a hole that is not there. The worst case
+    /// takes it whenever the scope had one, so the reserve bounds the sentence it can write.</summary>
+    bool TypeOrderShort(Values v)
```
Avick3110: **The rule is skipped when max_chars, rather than limit=, cut the listing.**

The gate takes `v.ByBudget` and `ScriptOmittedByBudget` but not `v.ByCut`. The render cuts the same single stream in the same order the budget does, so the misreading this rule was added to prevent is still reachable through the other truncator.

Scenario: `check(types=["Npc", "Weapon"], limit=1000, max_chars=4000)` on an order with ~200 findings. `limit=` drops nothing, so `TypeOrderShort` is false, while the body render stops partway through the NPC_ entries. The response names `types=[Npc, Weapon]` on its narrowing line, lists only NPC_ findings, and states no type-order rule — WEAP's absence reads as clean. The generic `{0} did not fit this response (max_chars={1})` clause says the listing is short, but not that a whole type may be missing from it, which is the distinction the new sentence exists to draw.

Lower severity than the other two since the reader is at least told the listing is incomplete, but the gate looks like it should be `v.Worst || v.ByBudget > 0 || v.ByCut > 0 || ScriptOmittedByBudget > 0` (with the sentence naming the right knob for the cause that fired).

src/housecarl-mcp/ReadSentences.cs
Avick3110: Folded in e6b2828b. The ordering claim is gone. The sentence now reads: there is ONE listing for every type in the scope (types=[...]), filled plugin by plugin (non-base plugins first, base masters last) and type by type inside each, and <knob> cut it short — so ANY of those types can be missing from it: absent there means unlisted, not clean.

The json field followed: `limit_spent_in_type_order` is now `listing_short_across_types` (plus `listing_short_by` for the knob, see the third comment). The `limit=` description on the tool and the CHANGELOG entry say the plugin-major rule too.

Test `ACutMultiTypeListingSaysTheTypesShareOneListing` pins the new wording and asserts the text does NOT contain "in the order".

src/housecarl-core/SweepScope.cs
Avick3110: Folded in e6b2828b. The label now spells the arms an entry expanded to, so the sentence names things the reader can act on: `types=[GMST → GameSettingInt, GameSettingFloat, ...]`. Built where both halves are in hand — `ResolveTypeFilterSet` gained an overload that resolves entry by entry and spells each via `SweepScope.SpellTypeEntry`, which spells a concrete entry as itself and an expanding one with its arms (`IGameSettingIntGetter` → `GameSettingInt`).

The gate stays on the resolved type count, as the doc-comment intended: one polymorphic-base entry fires (several types share the listing), one concrete entry does not. The narrowing line still echoes the raw `types=` the caller passed — that one is about what was asked for, not about what can be short.

Tests: `AnExpandingTypeEntryIsSpelledWithItsArms` drives a real GMST sweep through the render; `AOneEntryPolymorphicBaseIsLabelledWithItsArms` pins both sides of the gate.

src/housecarl-mcp/CheckAccounting.cs
Avick3110: Folded in e6b2828b. The gate is now `_typeScope is not null && (v.Worst || ShortByBudget(v) || ShortByCut(v))`, with the two causes as named helpers: `ShortByBudget` is `v.ByBudget > 0 || ScriptOmittedByBudget > 0`, `ShortByCut` is `v.ByCut > 0 || Short(v, SweepSubject.ScriptRecords)` — the scripts family has its own render cut and it hides a type the same way.

The sentence names the knob that fired: `limit=`, `max_chars=`, or `limit= and max_chars=` where both did, via `ShortKnob(v)`. The worst case takes the both-spelling, which is the longest, so the reserve still bounds the sentence. The json twin carries `listing_short_by` beside `listing_short_across_types` under the same test, so the lanes cannot name different knobs.

Test `AMaxCharsCutFiresTheRuleAndNamesMaxCharsAsTheKnob` sweeps with `limit=1000` and searches for the cap that cuts by max_chars alone, then asserts the rule fires, names max_chars=, and that no budget cause is stated.

Issue #667, opened 2026-09-08
[Bug] housecarl_skse reports SPID 7.0.0 where the mod's meta.ini says 7.3.1.0


Avick3110:
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/spid-authoring/VERDICT.md`). Needs confirming.

`housecarl_skse` reported Spell Perk Item Distributor as 7.0.0 on the ARR 2.0 order; the mod folder's `meta.ini` records 7.3.1.0. If the version shown is the DLL's file version and the DLL really says 7.0.0, the report should say which source it read so the discrepancy is visible; if it is misread, fix the read. An agent choosing SPID grammar by version floor (7.3-only tokens) acts on this number.

Avick3110:
Checked 2026-09-08: the report is what the DLL declares. `housecarl_skse` reads the SKSEPlugin_Version manifest the loader itself reads, and po3_SpellPerkItemDistributor.dll declares 7.0.0; the mod folder's meta.ini carries the Nexus release version, 7.3.1.0. Not a misread. What remains is wording: the inventory row should say "manifest version 7.0.0" so a reader knows which version it is looking at, and the description of the version line can say the Nexus version is a different number. Stays after 2.0 as a wording change.

Avick3110:
Read SPID's source: `housecarl_skse` is not misreading anything. SPID's SKSE manifest carries the major digit only.

`SKSEPlugin_Version` in `SPID/src/main.cpp` sets `v.PluginVersion(Version::MAJOR)` — minor and patch are never written. `Version::MAJOR` is `PROJECT_VERSION_MAJOR`, off `set(VERSION 7.3.0)` in `SPID/CMakeLists.txt`. Read at commit `31e76d3` (the tip of SPID 7.3.0); unchanged at 7.3.3, current `master`. So every SPID 7.x release declares 7.0.0 through the SKSE plugin version manifest, and 7.0.0 is exactly what the DLL says about itself there.

The full number lives elsewhere in the same DLL — the Win32 version resource, `SPID/cmake/version.rc.in`:

```
FILEVERSION @PROJECT_VERSION_MAJOR@, @PROJECT_VERSION_MINOR@, @PROJECT_VERSION_PATCH@, 0
VALUE "FileVersion", "@PROJECT_VERSION@.0"
```

That is where `7.3.1.0` comes from, and why `meta.ini` agrees with it.

[...]

PR #691, opened 2026-09-09
Fold the September gate findings into four skills

Avick3110:
The 2026-09-09 behavioural re-gate ran four skills against tasks their bodies do not answer, and each verdict left a list of things the skill either got wrong or never said. This folds those lists in: `mutagen-reference` now says what `w` means, how to resolve a field's `target` (a getter interface name, so strip `I...Getter` before grepping, and a marker interface has no enumerated implementers), blesses one grep over the index keys for enumerating the modeled types, and leads its worked examples with the zero-instance lookup that nothing else can answer; `open-animation-replacer` picks up the 2026-09-08 fixes that never landed as well as this round's, so the reference now gives a legacy Form B `<Plugin.esp>/<FormID>/` folder's priority (0, `kLegacyActorBase` in OAR's `Parsing.cpp`), the skill says to beat the submod that wins the frames rather than the highest integer, verification splices the new submod in and recomputes the winner table, the locate sweep covers the base-clip layer and discards the `_1stperson` false competitors, and the worked example converts a different real folder, which also publishes the `Random` block's shape; `kid-authoring` gains the trait-to-record-field column so a census needs no cross-skill lookup, an exclusion-then-trait worked line, and one entry saying how a KID line differs from a SPID one; `spid-authoring` documents the FormLink wire form beside the link step and the one-call union census, says a union within a section is a comma list, labels its level-filter numbers as measured on one instance, and says a written-out `-0x7~Skyrim.esm` is legal but inert.

The `w` finding was a corpus problem rather than a wording one, so it is fixed in the generator: the reference emitter now reports a field marked `id` as `w:false` and counts it out of the type's `writable/total` summary, and the shards are regenerated (ARMO reads `31/32`). The read view and the write pre-flight no longer disagree about `FormKey`. The `mutagen-reference` eval that asserts the count follows the shards; no other eval's answer changed.

Four questions the bundled grammar does not settle are written into the skills as unverified, with what would settle each, rather than answered: whether KID resolves a plugin-name filter by defining plugin or by winner (the description says "defined in", which is the reading to compose against; KID's source at that point settles it); whether a `*` wildcard also tests keyword strings (the strings-only bucket is source-backed, the channel list is description-backed and lists no keywords); whether SPID's `Data/` scan descends into subfolders (the corpus says there are no per-type subfolders, which is a layout statement, not a scan statement); and which version source wins when the SPID DLL manifest and MO2's `meta.ini` disagree (issue #667 — both get quoted, and a feature gate takes the lower).

Build clean, 1876 tests pass, `ci-all` 127/127.


.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -213,45 +229,60 @@ untouched, so it can be updated without losing the override.
...
-          "Form": { "pluginName": "Woodaxeweapons.esp", "formID": "5909" }, "Left hand": false } ] } ] }
+    { "condition": "IsActorBase", "requiredVersion": "1.0.0.0",
+      "Actor base": { "pluginName": "0Kaidan.esp", "formID": "2F9A" } },
+    { "condition": "IsEquippedType", "requiredVersion": "1.0.0.0",
+      "Type": { "value": 5.0 }, "Left hand": false },
+    { "condition": "Random", "requiredVersion": "2.3.0.0",
+      "State": { "scope": "Local", "shouldResetOnLoopOrEcho": false },
+      "Minimum random value": { "value": 0.0 },
+      "Maximum random value": { "value": 1.0 },
+      "Comparison": "<", "Numeric value": { "value": 0.2 } },
```
Avick3110: **The `Random` block's parameter names are not in the bundled corpus, and the paragraph says so before publishing them anyway.**

Line 261 states "the reference does not publish its argument names, so this is the shape to copy", then lines 272-277 give a concrete key set — `"State": {"scope", "shouldResetOnLoopOrEcho"}`, `"Minimum random value"`, `"Maximum random value"`, `"Numeric value"` — plus `"requiredVersion": "2.3.0.0"`. None of those strings appear anywhere in `references/oar-config-reference.md`; §4's component table stops at Form/Keyword/Numeric/Bool/Text/Comparison/NiPoint3/Multi, and §8's roster lists `Random` with no parameter row. The `2.3.0.0` version does not appear either.

The skill already has a policy for exactly this case, one table away — the `HasPerk` / `HasSpell` row in §8 reads "the exact argument name is not published in this reference; read it off the in-game editor or an existing config that uses the condition." This block does the opposite.

Concrete failure: a reader copies the block verbatim. OAR ignores a parameter key it does not recognise, so a wrong name leaves the component at its default (min 0 / max 0, comparison unset) and the gate fires either always or never instead of 20% — and **Detected Problems will not catch it**, because that panel flags INVALID conditions and duplicate priorities, not unrecognised parameter keys. The skill's own stop condition therefore reports success on a broken conversion. Separately, if `2.3.0.0` is above the installed OAR the whole condition is flagged invalid and the submod never wins.

Either source the names (editor screenshot, a shipped config, `Conditions.cpp`) and add them to §8's table with the same provenance the `IsEquippedType` rows carry, or convert a DAR folder that uses only sourced functions and drop `Random` from the example.

.claude/skills/spid-authoring/references/filters.md
```diff
@@ -182,8 +182,14 @@ into the range and loses it when it falls out. That pass is restricted to PC-lev
...
+one heavily overhauled 3,801-plugin order, with `housecarl_records` at a stamped epoch. They are here
+to show the *shape* of the gap — hundreds-fold, not a rounding error — and nothing else. They are not
+a constant, not a baseline for another order, and not a number to quote as a finding; a later census
+of the same order, on a differently-cut group, returned 2,249 / 155 / 917. Measure your own order at
+the epoch you are working against, and quote that.
+
+On such an order the two populations are nowhere near each other — that census found 17 of
 24,782 bandit-faction NPCs on a PC level multiplier, against 4,040 matching the line's own level
 bound; that same order carries 963 PC-level-mult NPC records in total, across every faction.
 Reading the filter as "auto-levelled NPCs only" understates the reach by a factor of hundreds there.
```
Avick3110, line 195: **"that census" now points at the wrong census, and 917 vs 963 are two different order-wide totals for the same order.**

The new paragraph ends by introducing a *later* census returning `2,249 / 155 / 917`. The very next sentence then says "that census found 17 of 24,782 …" — but 17/24,782/4,040/963 are the *original* numbers. The nearest antecedent for "that census" is now the later one, so a reader attributes the old triple to the new measurement.

Worse, the triples are unlabelled. If the third slot is "PC-level-mult NPC records on that order in total" — which is what line 195 says 963 is, "across every faction" — then the same order cannot be both 917 and 963, and the reader has no way to tell whether 917 is a different quantity, a different epoch, or a contradiction. The paragraph's own instruction is "quote the number with the call that produced it"; neither triple carries one.

Name the census ("the 2026-0X-XX census" / "the re-census") and label the three slots, or drop the second triple and keep only the shape claim.

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -127,6 +141,8 @@ An enum block carries its legal values:
 
 A block's `writable` is the type's `writable/total` field count — a summary. A field's own `w` is what governs whether you can set that field.
 
+**`w` is what the write tools will accept, not merely what the library exposes a setter for.** The two differ on one class of field: an identity field, marked `id` (`FormKey`, and `ModKey` / `Master` on a header), has a setter in the library but the write pre-flight refuses it outright, because a record's identity is not free-edit content. The reference reports those as `"w":false,"id":true` and counts them out of the type's `writable/total`, so **an `id` field is never writable** and the read view and the write tools agree. Change a record's identity by creating or copying the record, never by setting the field.
```
Avick3110: **`ModKey` / `Master` are not on a header.** The one `kind:"header"` block in the shipped shards is `SkyrimModHeader`, and its fields are `Author, Deleted, Description, Flags, FormID, FormVersion, INCC, INTV, MasterReferences, OverriddenForms, Stats, TypeOffsets, Version, Version2` — no `ModKey`, no `Master`, and it is unchanged at `14/14` by this PR.

The two `id` fields outside records both live in `structs.jsonl`: `SkyrimMod.ModKey` and `MasterReference.Master`.

Concrete failure: this section is the lookup instruction for the `id` key. A reader told the non-`FormKey` identity fields sit "on a header" greps the header block, finds neither, and concludes the reference is inconsistent with the sentence — the one thing this paragraph exists to prevent.

Suggested wording: `` (`FormKey` on a record, and `ModKey` / `Master` on the `SkyrimMod` and `MasterReference` structs) ``. `plugin/CHANGELOG.md:17` carries the same "a header's `ModKey` / `Master`" phrasing and needs the same fix.

.claude/skills/kid-authoring/references/filters.md
```diff
@@ -89,6 +96,15 @@ into the requirement set (so `ArmorHeavy+ArmorGauntlet` is one requirement-pair)
...
 
+**Which channels a wildcard tests.** A `*` term is a String filter, so it is a substring test over
+§1's string channels — item name, effect archetype, actor value name, nif path — and nothing else.
+The strings-only bucket is `[source]`: a wildcard is never resolved to a form, so `*Iron` cannot test
+a keyword *record*, and `*` before a FormID, EditorID or plugin name matches nothing. The table's
+"name/keyword" wording above is the Nexus description's; whether the item's own keyword EditorIDs are
+one of the compared strings is **unverified** — §1's channel list, also from the description, does not
+include them. KID's source at the wildcard comparison would settle it. Compose as if name is the
+channel that matters, and say so when a count rests on it.
+
```
Avick3110, line 111: **This paragraph settles the keyword question and then un-settles it, three lines apart.**

Line 100 asserts the wildcard channels are "item name, effect archetype, actor value name, nif path — **and nothing else**" — which is a positive claim that keyword strings are *not* compared. Lines 103-105 then say whether keyword EditorIDs are among the compared strings is "**unverified**".

Those cannot both be acted on. A reader composing `*Iron` has to decide whether the keyword channel is ruled out (line 100) or an open question (line 104), and the section exists precisely to answer that. The last line's "compose as if name is the channel that matters" is guidance, not a resolution of the contradiction above it.

Drop "and nothing else" (the sourced fact is the strings-only *bucket*, not the channel roster — §1's list is `[desc]`), so the paragraph reads: sourced = never resolved to a form; unverified = which strings are compared.

Also in the same paragraph: "`*` before a FormID, EditorID or plugin name matches nothing" overstates the mechanism. Per §3, `*X` goes into the strings-only bucket, so `*MyMod.esp` is a substring test against the string channels — it will practically never match, but it is not the parser refusing it, and "matches nothing" invites the reader to treat it as a no-op rather than a mis-cut filter.

src/housecarl-generator/ReferenceEmitter.cs
```diff
@@ -67,7 +67,9 @@ public static void Emit(Corpus corpus, string refDir)
         Mutable = t.MutableInterface,
         AbstractBase = t.AbstractBase,
         Arms = t.Arms,
-        Writable = t.Kind == "enum" ? null : $"{t.WritableCount}/{t.FieldCount}",
+        // Identity fields carry a Mutagen setter but the write pre-flight always refuses them, so the read
+        // view reports them as not writable and the summary counts them out. corpus.json keeps the raw count.
+        Writable = t.Kind == "enum" ? null : $"{t.Fields.Count(f => f.Writable && !f.IsIdentity)}/{t.FieldCount}",
```
Avick3110: Minor: the numerator now derives from `t.Fields` while the denominator still reads the cached `t.FieldCount`. They agree today — `CorpusGenerator` sets `FieldCount = fields.Count` from the same list at construction and nothing mutates `Fields` afterwards — but before this change both halves came from one snapshot (`WritableCount`/`FieldCount`), and now a future pass that filters `Fields` would silently emit a ratio like `28/32` with no field to account for the gap, in a file whose whole contract is that it cannot disagree with `corpus.json`.

`$"{t.Fields.Count(f => f.Writable && !f.IsIdentity)}/{t.Fields.Count}"` keeps both halves on the same list for free.

.claude/skills/open-animation-replacer/SKILL.md
Avick3110: Folded, by sourcing the names rather than dropping the block. Two shipped configs in Ashe - Crystal Heart carry the block verbatim - same six keys, same 2.3.0.0: AsheConditionalAnims/Dung_In/config.json ("<" at 0.333) and AsheAttackAnims/Greatsword 02 - Power Attack 01/config.json ("<=" at 0.5), read 2026-09-09. Added a Random row plus that provenance to the DAR mapping table in the reference, and the SKILL.md sentence now points at the table and warns about the 2.3.0.0 floor instead of saying the names are unpublished.

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -213,45 +229,62 @@ untouched, so it can be updated without losing the override.
...
+§5's table — DAR's type numbers are the same numbering, `n → n`, so never translate them. `0x00002f9a`
+becomes the local hex `"2F9A"`: a DAR argument is already the local id in the plugin named beside it,
+so drop the `0x` and the leading zeros and nothing else.
 
 DAR has no parenthesis grouping, but the binding is settled: **`OR` binds tighter than `AND`**, so a
 chain is an `AND` of `OR`-groups, and an `OR`-group is a run of lines ending in `OR` plus the next
 line that yields a condition — a blank line or a `;` comment is skipped without closing the group
-(§8, from OAR's `Parsing.cpp`). Here that leaves the `NOT IsInCombat()` guard as a
-top-level term and pairs the two `IsEquippedRight` lines into one `OR`:
+(§8, from OAR's `Parsing.cpp`). Read the binding before you write anything: here all five lines end
```
Avick3110: **Two claims in this paragraph are wrong, and one of them contradicts the rule it just quoted.**

1. "here all five lines end in `AND`" — the block has five lines but only the first four carry a trailing `AND`; `NOT IsInCombat()` ends in nothing. The same sentence is baked into the JSON `description` below ("All five lines end in AND").

2. "had the two middle lines ended in `OR` they would bind into a single `OR` term" — by the rule stated two sentences earlier (and by `references/oar-config-reference.md` §8: *an OR-group is a maximal run of lines ending in OR, plus the next line that yields a condition*), an `OR` on lines 2 and 3 would open a group that also swallows line 4 (`NOT IsSneaking()`), giving a **three**-member OR, not a two-member one.

The counterfactual is the only place this section now demonstrates the binding rule (the old worked example, which had a real `OR` group and showed the nested `Conditions` shape, was replaced by a flat one). A reader converting a real mixed chain by this example drops the following term out of the group and gates the submod on the wrong situation — the exact error the section exists to prevent.

.claude/skills/kid-authoring/references/traits.md
```diff
@@ -19,95 +19,104 @@ section on them is ignored. **[source]**
...
-| Trait | Meaning |
-|---|---|
-| `S` / `-S` | teaches a spell / doesn't |
-| `AV` / `-AV` | teaches a skill (actor value) / doesn't |
-| `<av>` | a numeric actor value = the specific skill/spell-type the book is associated with (e.g. `20` = Destruction). Names→numbers in `value-tables.md`. |
+| Trait | Meaning | Record field |
+|---|---|---|
+| `S` / `-S` | teaches a spell / doesn't | `Teaches`, a polymorphic field, sitting on its `BookSpell` arm |
+| `AV` / `-AV` | teaches a skill (actor value) / doesn't | `Teaches` sitting on its `BookSkill` arm (the third arm is `BookTeachesNothing`) |
+| `<av>` | a numeric actor value = the specific skill/spell-type the book is associated with (e.g. `20` = Destruction). Names→numbers in `value-tables.md`. | `Teaches.Skill` on the `BookSkill` arm |
```
Avick3110: **`Teaches.Skill` is the `Skill` enum; KID's `<av>` is an ActorValue number. They do not line up.**

The new preamble promises this column is "the field path to hand `housecarl_records`'s `where=`". But `BookSkill.Skill` is typed `Skill` — `OneHanded, TwoHanded, Archery, Block, Smithing, HeavyArmor, LightArmor, Pickpocket, Lockpicking, Sneak, Alchemy, Speech, Alteration, Conjuration, Destruction, …` — so Destruction sits at **14**, while `value-tables.md`'s ActorValue table (the one this row cites) puts Destruction at **20**.

So the census the column advertises for `Keyword = X|Book|NONE|20` is `where=["Teaches.Skill = 20"]`, which compares against a value outside the `Skill` range and reports 0 books, silently, while the KID line tags every Destruction tome.

The SLGM row explicitly reconciles its enum offset ("with `None` at 0, so KID's 1–5 line up") and the SPEL `ST` row warns the enums differ — this row needs the same treatment: either give the name form (`Teaches.Skill = Destruction`) or flag the two numberings.

.claude/skills/kid-authoring/references/traits.md
```diff
@@ -19,95 +19,104 @@ section on them is ignored. **[source]**
...
-| Trait | Meaning |
-|---|---|
-| `T(value)` | furniture type — `value-tables.md` → Furniture Types (0–3) |
-| `BT(value)` | workbench type — `value-tables.md` → Bench Types (1–7) |
-| `US(value)` | workbench use-skill actor value — `value-tables.md` → Actor Values |
+| Trait | Meaning | Record field |
+|---|---|---|
+| `T(value)` | furniture type — `value-tables.md` → Furniture Types (0–3) | **not one scalar field** — FURN carries `Flags` (which has `IsPerch`) and a `Markers` list; resolve against the record before you census it |
+| `BT(value)` | workbench type — `value-tables.md` → Bench Types (1–7) | `WorkbenchData.BenchType` |
+| `US(value)` | workbench use-skill actor value — `value-tables.md` → Actor Values | `WorkbenchData.UsesSkill` |
```
Avick3110: Same numbering mismatch as the BOOK `<av>` row: this row labels the KID token a "workbench use-skill **actor value**" and points at `value-tables.md` → Actor Values, but `WorkbenchData.UsesSkill` is typed `Skill`, not `ActorValue` (`{"n":"UsesSkill","t":"Skill","c":"enum",…}` in `structs.jsonl`).

A census written from this column as `where=["WorkbenchData.UsesSkill = <KID number>"]` therefore selects a different skill, or none, with no error. `BT(value)` on the row above is fine — `WorkbenchData.Type` runs `None`=0 … `SmithingArmor`=7, so KID's 1–7 do line up.

.claude/skills/mutagen-reference/SKILL.md
```diff
@@ -127,6 +141,8 @@ An enum block carries its legal values:
 
 A block's `writable` is the type's `writable/total` field count — a summary. A field's own `w` is what governs whether you can set that field.
 
+**`w` is what the write tools will accept, not merely what the library exposes a setter for.** The two differ on one class of field: an identity field, marked `id` (`FormKey` on a record, and `ModKey` / `Master` on the `SkyrimMod` and `MasterReference` structs), has a setter in the library but the write pre-flight refuses it outright, because a record's identity is not free-edit content. The reference reports those as `"w":false,"id":true` and counts them out of the type's `writable/total`, so **an `id` field is never writable** and the read view and the write tools agree. Change a record's identity by creating or copying the record, never by setting the field.
```
Avick3110: "an identity field, marked `id` (`FormKey` on a record, and `ModKey` / `Master` on the `SkyrimMod` and `MasterReference` structs), **has a setter in the library** but the write pre-flight refuses it" — not true of `SkyrimMod.ModKey`.

Against `main`, `structs.jsonl` line 207 (SkyrimMod) is **unchanged** by this PR and already read `{"n":"ModKey","t":"ModKey","c":"value","w":false,"id":true}`, with the ratio still `120/130`. Only `MasterReference.Master` and the 133 record `FormKey` fields flipped `true`→`false` here — i.e. `SkyrimMod.ModKey` has no library setter at all, so it is not an instance of "the two differ".

Same claim in `plugin/CHANGELOG.md:18` ("carries a library setter"). Small, but this is the one paragraph whose job is to say exactly what `w` means.

.claude/skills/open-animation-replacer/SKILL.md
```diff
@@ -58,7 +58,12 @@ Every step below names the section it needs, so one section can be read on its o
    wildcard, so `meshes/actors/*/…` walks every facegen, body and armor mesh under `meshes/actors`
    before a row renders, and `limit=`/`offset=` do not bound that — they window the render. A DAR
    Form B `<Plugin.esp>/<FormID>/` folder carries no marker file at all, so sweep
-   `.../DynamicAnimationReplacer/**/*.hkx` too when an actor-base override is in play. The call
+   `.../DynamicAnimationReplacer/**/*.hkx` too when an actor-base override is in play. Add a third
+   selector for the base-clip layer, `meshes/actors/character/animations/<original.hkx>` — a mod that
+   replaces the original file outright competes for the same frames and neither replacer selector
+   sees it. Discard any hit under `character/_1stperson/animations/`: that is the first-person
```
Avick3110: This discard rule can never fire under the selectors this same step prescribes. All three are rooted at `meshes/actors/character/animations/` (the OAR `config.json` glob, the DAR `_conditions.txt` / `**/*.hkx` glob, and the new base-clip `<original.hkx>`), and the first-person graph is `meshes/actors/character/**_1stperson**/animations/` — a *sibling* of `animations`, not a child of it.

So either a fourth selector covering `character/_1stperson/animations/` is missing, or this sentence is dead text that reads as though the sweep covers first person when it does not. Worth saying which, since the paragraph's other additions are about making the sweep complete.

.claude/skills/open-animation-replacer/SKILL.md
Avick3110: Both fixed. The prose now says the first four lines end in AND and the last ends in nothing, and the JSON description says "No line ends in OR, so this is one flat AND set." The counterfactual was wrong as you describe, so I replaced it: there is now a second snippet with a real OR group (NOT IsInCombat AND / IsEquippedRight(A) OR / IsEquippedRight(B)) showing the nested Conditions shape, and the note about the 777000 chain now says an OR on lines 2 and 3 would give a three-member group including NOT IsSneaking, not a two-member one.

.claude/skills/kid-authoring/references/traits.md
Avick3110: Fixed. The row now says Teaches.Skill is typed Skill, not ActorValue, spells out both numberings (Skill runs OneHanded 0 ... Destruction 14 ... Enchanting 17, against the ActorValue table's 20), and gives the name form where=["Teaches.Skill = Destruction"] as the census, saying the KID number compared as a Skill silently selects a different skill or none.

.claude/skills/kid-authoring/references/traits.md
Avick3110: Fixed the same way. The US row now says WorkbenchData.UsesSkill is typed Skill, not ActorValue, points at the BOOK row for the two numberings, and gives the name form where=["WorkbenchData.UsesSkill = Smithing"]. I also spelled out on the BT row why its 1-7 do line up (WorkbenchData.BenchType runs None 0 through SmithingArmor 7).

.claude/skills/mutagen-reference/SKILL.md
Avick3110: Right, checked it: SkyrimMod line 207 is byte-identical to main and already read w:false. Reworded both places. The skill paragraph now says the pre-flight refuses every id field whether or not the library exposes a setter, and names MasterReference.Master and a record's FormKey as the two that do. The changelog entry now says those are the fields that flip true to false in this PR.

.claude/skills/open-animation-replacer/SKILL.md
Avick3110: Real, and the sweep is third-person only in fact: the re-gate arms' asset_status call used two selectors, both rooted at character/animations/, and never touched the first-person tree. So I dropped the discard sentence rather than adding a fourth selector. The step now says the three selectors cover the third-person graph, that the first-person graph is a sibling tree with a different set of frames that does not compete, and that a first-person job roots the same three selectors at that path instead.

PR #692, opened 2026-09-09
Add the facegen findings family to housecarl_check

Avick3110:
`housecarl_check findings=["facegen"]` answers the dark-face question server-side, one row per NPC: which mod wins its head `.nif`, which wins its face `.dds`, which plugin wins the `NPC_` record behind them, the mismatch class and a fix sentence. Two independent precedences decide a face — the MO2 VFS for the files, load order for the record — and the desync between them is why xEdit shows nothing; the 2026-09 gate had both arms rebuild that join by hand outside the tools, about half an hour and two dozen calls each. The sweep composes what already exists: `FaceGenPath.For` for the keyed path, `AssetResolver` for the file winners, `ReadEngine` + `FieldsDiff` for the record winner and the face-field comparison. It lives in its own core file with its own service entry and its own render, not in `LoadOrderService.cs`. Classes are `tint_absent`, `mesh_absent`, `bake_absent`, `split_bake`, `stale_bake`, `family_split` (benign — counted in the header, listed only under its own token), `foreign_index` and `inert`; the population is the union of every in-scope NPC that needs a bake plus every facegen file on disk whose key resolves to nothing. `check` also gains `to_file=`, writing every running family's findings to one JSONL artifact under the manifest convention `records` uses.

**Measured whole-order cost, and what it decided.** On the live ARR 2.0 order (3,801 plugins, 66,856 NPC records) the facegen family unscoped takes **45.0 s** with `counts_only=true` and **36.3 s** listing, against the errors family's **35.5 s** cold and **24.0 s** warm in the same process. It sits with errors, not with scripts, so **unscoped is allowed** and no `plugins=` refusal was added; the family is still kept out of the default findings set, which stays the errors family alone. Class counts on that order: `stale_bake` 578, `bake_absent` 528, `inert` 286, `family_split` 271, `split_bake` 77, `mesh_absent` 6, `tint_absent` 3, `foreign_index` 1, with 59,373 NPCs excluded (55,775 `Template`+`Traits`, 3,598 on a race carrying no `FaceGenHead` flag) and 139 clean pairs whose stale test had no comparison pole and are reported as untested rather than counted clean.

Three judgements the probe forced, each stated in the code and in the response. A pair is clean when both halves come out of one **MO2 layer** rather than one provider name — vanilla ships the head in `Skyrim - Meshes0.bsa` and the tint in `Skyrim - Textures0.bsa`, and by provider name that read as 1,950 split bakes. An NPC whose race lacks Mutagen's `Race.Flag.FaceGenHead` bakes no head at all, so creature actors are excluded and counted rather than reported as missing one (4,112 rows before that gate, 528 after). And a field list whose items are the same in a different **order** is not a stale bake, which halved that class.

`facegen-diagnostics` is deleted here, as ruled: its causes-and-fixes table becomes `docs/facegen.md`, its `references/mesh-repairs.md` folds into `housecarl_nif_set`'s `op=`, `texture_slot=` and `path=` descriptions, and the check's own description carries one fix sentence per class plus the not-this line (purple/white face, player-only grey, brown weight face, SPID-distributed appearance). Eight skills ship after this; the build script's list, the Codex umbrella's routing table, both READMEs and the naming example are updated. The private SPEC gains one dated line under its `check` section for `to_file=`, since that is a vocabulary change.

#584 is cited, not closed: this family is built beside it on `AssetResolver`, and `asset_status` still owes its own `to_file=` and formid/pair mode for the per-path lane, which is a different question. No milestone issue is closed by construction.

Changelog: two entries under `## Unreleased` in `plugin/CHANGELOG.md`, one for the family and the deleted skill, one for `to_file=`.

Build, `dotnet test --filter "tier!=bridge"` (1,901 passed) and the generator's `ci-all` (127/127) are green. The whole-order figures above were measured locally against the real ARR 2.0 instance with the server built from this branch; no test in the suite needs a real load order.


src/housecarl-core/FaceGenCheck.cs
```diff
@@ -0,0 +1,572 @@
...
+                }
+                if (judged.Contains(fk)) continue;                         // answered by the record half above
+                if (view.ResolveWinner(fk) is not { } w)
+                {
+                    Emit(FaceGenFindingClass.Inert, fk, folder, null, null, null,
+                         $@"{folder}\{name} — no plugin in this order defines this FormID, so no NPC reads this bake");
+                    continue;
+                }
+                // Resolves to something that is not an NPC at all — the PlacedNpc case both gate arms got wrong.
+                var body = view.GetRecord(session, w.WinnerPlugin, fk);
```
Avick3110, line 303: **NullReferenceException on an unreadable body, outside the try/catch.** `view.GetRecord(...)` is documented nullable and returns null when "the plugin may have changed since the index was built" — see the same call guarded exactly that way in `DialogueValidate.cs:320`. Here the result goes straight into `TypeNameOf(body)`, which does `body.GetType()`, so a null body throws NRE.

The record half of this same sweep guards the identical call (`if (view.GetRecord(...) is not { } body) continue;`); the file half does not. Worse, this block sits *after* the `try { ... } catch` that closes above it, so the throw escapes `Run` entirely: instead of the one-plain-sentence refusal the family promises, a whole-order `check findings=["facegen"]` dies with a raw NRE.

Scenario: a whole-order sweep while MO2 or xEdit rewrites a plugin; the first orphan-file key whose winner is that plugin crashes the call.

src/housecarl-core/FaceGenCheck.cs
```diff
@@ -0,0 +1,572 @@
...
+        {
+            foreach (var p in scope)
+            {
+                var name = (p ?? "").Trim();
+                if (name.Length == 0) continue;
+                if (!view.ContainsPlugin(name)) continue;              // the caller split off-order already
+                if (!targets.Contains(name, StringComparer.OrdinalIgnoreCase)) targets.Add(name);
+            }
+        }
+        if (exclude is { } ex)
```
Avick3110, line 96: **`exclude=` is silently ignored on an unscoped facegen sweep, and a typo'd exclude name is not refused.**

This block only filters `targets` and `offOrder`. But `targets` is populated *only* when `scope is { Count: > 0 }` — the whole-order path below enumerates `view.WinnerRecordsOfType(NpcTypes)` and never consults `targets`. So `housecarl_check findings=["facegen"] exclude=["implicit"]` (or `exclude=["Skyrim.esm"]`) runs `RemoveAll` over an empty list, reports every vanilla NPC anyway, and says nothing about it — a silently degraded mode.

Compare `ErrorCheck.cs:167-213`: it builds `targets` from `resolver.PluginNames` for the unscoped case *before* applying the exclusion, and additionally refuses when a caller-typed name is not in the scope the sweep would cover ("an exclusion that matches nothing would return the findings you asked to leave out"). The facegen family has neither behaviour, so a typo in `exclude=` is accepted and does nothing.

Separately, `CheckTools.cs:43` still says "the two SWEPT families share one [scope]" though facegen is now a third family taking the same axes.

src/housecarl-core/FaceGenCheck.cs
```diff
@@ -0,0 +1,572 @@
...
+        var judged = new HashSet<FormKey>();
+        var offOrderScanned = new List<string>();
+        var reported = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        string? scanError = null;
+
+        using var session = resolver.OpenSession();
+
+        void Judge(FormKey fk, IMajorRecordGetter body, string winnerPlugin, bool offOrderFile)
+        {
+            if (recordScope is not null && !recordScope.Matches(fk, body)) return;
```
Avick3110, line 140: **A record-scope narrowing turns the file half into a per-file full-plugin scan.**

`Judge` returns on `!recordScope.Matches(...)` *before* `judged.Add(fk)`, so an NPC filtered out by `types=` / `formids=` / `editorid_contains=` never enters `judged`. `wholeOrder` is still true (it tests only `scope`/`offOrder`), so every one of that NPC's facegen files falls through to `view.GetRecord(session, w.WinnerPlugin, fk)` at line 251 — called with no getter type, i.e. `SeekBody`'s flat `EnumerateMajorRecords()` walk of the entire winning plugin.

Scenario: `housecarl_check findings=["facegen"] editorid_contains="Bijin"` on the ARR order. A handful of NPCs are judged; the other ~20,000 facegen files on disk each trigger a full flat enumeration of their winner plugin (Skyrim.esm is ~1M records). The call effectively hangs, with no refusal and no note.

Recording every key `Judge` saw — matched or not — in a separate set and consulting that in the file half would fix it.

src/housecarl-mcp/CheckAccounting.cs
```diff
@@ -395,6 +416,20 @@ string Compose(Values v)
...
+        // The facegen family's own two-part shape: what this response carries against what the sweep found, then
+        // the listing budget's share of what is absent. The found total is never capped, so a short listing with no
+        // total beside it would read as the whole answer.
+        if (Has(SweepSubject.FaceGenRows))
+        {
+            sb.Append(Short(v, SweepSubject.FaceGenRows)
+                ? string.Format(ReadSentences.SweepFaceGenVisible, Shown(v, SweepSubject.FaceGenRows),
+                                Found(SweepSubject.FaceGenRows))
+                : string.Format(ReadSentences.SweepFaceGenAllVisible, Found(SweepSubject.FaceGenRows)));
+            if (_faceGenFound > Found(SweepSubject.FaceGenRows) || v.Worst)
```
Avick3110, line 430: **This sentence blames `limit=` for rows that were withheld on purpose.**

`_faceGenFound` is `r.TotalFound`, which counts `family_split` findings — and `Add` deliberately keeps those out of `r.Findings` ("the benign class is COUNTED in the header and LISTED only when the caller named it"). So `_faceGenFound > Found(SweepSubject.FaceGenRows)` is true whenever any benign row exists, regardless of the budget, and the response prints `"N of the M finding(s) this sweep found were listed: the listing budget (limit=1000) ran out."` when it did not.

Scenario: the `FaceGenWorld` fixture (5 listed findings, 1 `family_split`, default limit) renders that sentence with `limit=1000`, directly contradicting the head's own note that the family_split rows are counted but not listed. On the measured ARR order 271 benign rows guarantee it fires.

The comparison wants the total *eligible for listing* — `TotalFound` minus the withheld benign count.

src/housecarl-mcp/CheckArtifact.cs
```diff
@@ -0,0 +1,207 @@
...
+                total++;
+                writer.WriteRow((w, _) => Row(w, "facegen", f.Class, formid: f.FormId, editorid: f.EditorId,
+                                              definingMaster: f.DefiningMaster, recordWinner: f.RecordWinner,
+                                              meshWinner: f.MeshWinner, tintWinner: f.TintWinner,
+                                              owningMod: f.OwningMod, detail: f.Detail, fix: f.Fix));
+            }
+
+        // The facegen family counts findings its listing budget cut, so the file's own total says so rather than
+        // letting row_count read as the whole answer.
+        if (s.FaceGen is { Error: null } fgt) total += Math.Max(0, fgt.TotalFound - fgt.Findings.Count);
```
Avick3110: **`to_file=` silently drops the `family_split` rows while counting them in `total`.**

`fg.Findings` never contains `family_split` findings unless the caller named that class, so the artifact omits them; this line adds them back into `total`, leaving the manifest with `total > row_count` under the same "what limit= already cut is cut here too" framing the `to_file=` description uses. A consumer diffing `total` against `row_count` will conclude the listing budget ran out.

The parameter description promises "the COMPLETE findings of every family that ran"; for facegen it is complete-minus-a-class, and no manifest note says which class or why. Either write the benign rows (a file has no budget to protect) or name the omission in `notes`.

src/housecarl-mcp/ReadSentences.cs
```diff
@@ -448,6 +448,35 @@ static string ScriptPropLabel(ScriptCheckResult r)
...
+    /// which plugin wins the record — and never the render, so a clean result is not a promise that a face looks
+    /// right. The not-this line is here too: the classes this family cannot see are named rather than left to be
+    /// mistaken for a clean answer.</summary>
+    [MustState("provenance", "never the render", "purple or white face", "player-only grey", "brown weight face")]
+    internal const string SweepFaceGenBoundary =
+        "reports provenance — which mod wins each half of the bake and which plugin wins the record — and never " +
+        "the render: it cannot read a .dds's pixels, cannot bake geometry (that is the Creation Kit's Ctrl+F4), " +
+        "and a clean row is not a promise the face looks right. NOT this family: a purple or white face (a missing " +
+        "texture), player-only grey (RaceMenu/SKEE runtime state), a brown weight face (save-baked weight), or an " +
+        "appearance distributed at runtime by SPID. See docs/facegen.md for the causes behind each class.";
```
Avick3110: **This boundary sentence is rendered to end users, but `docs/facegen.md` does not ship.**

`scripts/build-plugin.ps1` copies `plugin/`, the skill folders and the corpus into the distributed plugin; `docs/` is not among them. So every facegen response — plus the `check` `findings=` description (`CheckTools.cs:168`) and `nif_set`'s `path=` description — points an installed user at a repo-only path they have no copy of.

Before this PR the same causes-and-fixes table shipped, as `.claude/skills/facegen-diagnostics/references/facegen-causes-and-fixes.md`, because skill folders are copied into the package. Deleting the skill removes the user's access to it while the runtime text still cites it. (`plugin/codex/housecarl/SKILL.md` at least qualifies it as "in the repo"; these strings do not.)

src/housecarl-generator/CheckMergeProbe.cs
```diff
@@ -1620,7 +1630,7 @@ void WriteMod(string path, string name, int npcs, int weapons)
...
         else if (!dialogueHint.Contains("info_order", StringComparison.Ordinal))
             retiredBad.Add("the housecarl_validate_dialogue row names no destination for class 8 (the effective merged INFO order), which this surface deliberately does not carry");
-        Arm($"ORCH-EVERY-ABSORBED-ANCESTOR-HAS-A-RETIRED-NAME-ROW: each of the {SweepFamilySelection.Registered.Count} families this surface registers has exactly ONE retired-name row pointing at it, reachable by the ancestor's own name, and the dialogue row names BOTH destinations — the sweep for classes 1-7 and records project=info_order for class 8. Asked off the family registry, so a family added with no row for its ancestor reddens this",
+        Arm($"ORCH-EVERY-ABSORBED-ANCESTOR-HAS-A-RETIRED-NAME-ROW: each of the 3 families this surface ABSORBED has exactly ONE retired-name row pointing at it, reachable by the ancestor's own name, and the dialogue row names BOTH destinations — the sweep for classes 1-7 and records project=info_order for class 8. Asked off the family registry, so a family added with no row for its ancestor reddens this",
```
Avick3110: **The assertion text no longer describes how it is computed.** The loop above now iterates a hardcoded three-family array, but this message still says "Asked off the family registry, so a family added with no row for its ancestor reddens this" — which the change makes false. A future absorbed-ancestor family added to `Registered` will not redden anything, while the message claims it would.

Related: the comment at line 387 still reads "TWO registered families the default does not run" and "a fourth family landing must turn this red" beside a `== 3` check.

src/housecarl-mcp/FaceGenSweepRender.cs
```diff
@@ -0,0 +1,179 @@
...
+        w.WriteString("fix", f.Fix);
+        w.WriteEndObject();
+    }
+
+    /// <summary>The row's cost, for the demand pass — measured where the unit lands, like every other family's.</summary>
+    internal static int RowCostFor(FaceGenFinding f, int depth, bool subsequent)
+        => JsonWire.MeasureUnit(depth, subsequent, w => WriteRow(w, f));
+
+    /// <summary>The columns a <c>to_file=</c> artifact's rows carry, in order.</summary>
+    internal static readonly string[] RowSchema =
```
Avick3110: **Dead schema that disagrees with the real one.** `FaceGenSweepRender.RowSchema` has no callers — the `to_file=` artifact writes `CheckArtifact.RowSchema` (16 columns including `family`, `plugin`, `target`, …), and that is what the manifest advertises. This 10-column list documents a shape nothing produces, and a later edit to the artifact columns will update one and leave the other. Delete it, or have `CheckArtifact` derive from it.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2187,6 +2194,12 @@ static int ScanErrorRowCost(RecordScriptFindings rec, int depth, bool subsequent
     /// already written by the time this axis's own rows are tested, so an axis over-reserves against itself by that
     /// much. Over-reserving costs characters; under-reserving is what this exists to stop, so the simpler
     /// arithmetic is deliberately taken in the safe direction.</para></summary>
+    /// <summary>The same two-pass axis write, keyed off the axes' own subjects so a family that already holds
+    /// <see cref="HistogramAxis"/> values does not spell its json field names a second time.</summary>
+    internal static void WriteHistogramAxes(Utf8JsonWriter w, BoundedBody? body, int rowLimit, params HistogramAxis[] axes)
```
Avick3110: **Orphaned doc comment.** This method was inserted between `WriteHistograms`' existing `<summary>` block and `WriteHistograms` itself, so `WriteHistogramAxes` now carries two stacked `<summary>` tags — the reserve/over-reserve explanation, which is about `WriteHistograms`, plus its own — and `WriteHistograms` is left undocumented. Move the new method below `WriteHistograms`, or move the old block back onto it.

src/housecarl-core/AssetResolver.cs
```diff
@@ -127,6 +127,16 @@ internal sealed class LooseSubtree
     /// import path in this same precedence.</summary>
     public IReadOnlyList<(string Name, string Dir)> LooseRoots => _looseRoots;
 
+    /// <summary>The MO2 layer each active archive's FILE lives in, keyed by the archive's filename — the same
+    /// <see cref="AssetProvider.Source"/> a BSA hit reports. A bulk caller that has to ask which MOD provides a hit
+    /// reads this once instead of paying <see cref="ResolveForPlacement"/> per path for the one field that carries
+    /// it. Empty values are dropped: an archive with no known layer answers nothing rather than "".</summary>
+    public IReadOnlyDictionary<string, string> ArchiveOwningMods
```
Avick3110: **Arbitrary owner when two active archives share a filename.** `GroupBy(Path.GetFileName).First()` picks whichever archive came first in `_archives`, but `ResolveProviders` chooses the BSA winner by plugin rank with the filename only as a tie-break — so the layer reported here can belong to a *different* archive than the one that actually won the path.

Since `Layer` decides `clean` vs `split_bake` vs `family_split` (`FaceGenCheck.Classify` compares `mw.Layer` to `tw.Layer`), a mis-attributed layer silently changes an NPC's verdict. Keying by archive *path* and looking the winner up by its own descriptor would avoid it; at minimum the ambiguity deserves naming in the doc comment, since this family's whole claim is provenance.

src/housecarl-core/FaceGenCheck.cs
Avick3110: Folded. The null body is a reported inert row now, naming the plugin and saying it changed on disk since the order was indexed, instead of reaching TypeNameOf and throwing outside the try/catch.

src/housecarl-mcp/CheckAccounting.cs
Avick3110: Folded. The sentence compares against the rows eligible for listing (found total minus the withheld family_split count) rather than the found total, so a benign row no longer makes a complete listing claim the budget ran out. Fixture test added.

src/housecarl-mcp/CheckArtifact.cs
Avick3110: Folded, the first way. The sweep keeps the withheld benign rows aside and to_file= writes them, so the artifact carries every class and total matches row_count. Confirmed on the ARR order: 1750 found, artifact 1271 rows including all 271 family_split, the remaining 479 cut by limit= alone.

src/housecarl-mcp/ReadSentences.cs
Avick3110: Folded. The three runtime strings and the codex line cite https://github.com/Avick3110/houseCARL/blob/main/docs/facegen.md now, from one ReadSentences const. No shipping mechanism added, and the one fix sentence per class stays in the row so the tool answers without the link.

src/housecarl-generator/CheckMergeProbe.cs
Avick3110: Folded. The assertion now says it is asked off the three families named in the loop, not off the registry, and the comment beside the == 3 check counts three not-run families and a fifth family landing.

src/housecarl-mcp/FaceGenSweepRender.cs
Avick3110: Folded. Deleted. CheckArtifact.RowSchema is the only one.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded. The reserve/over-reserve summary is back on WriteHistograms; WriteHistogramAxes keeps its own.

src/housecarl-core/FaceGenCheck.cs
Avick3110: Folded, both halves. exclude= now applies on every lane: the unscoped sweep builds its in-scope set from the resolver's plugin names and drops an NPC only when every plugin touching it is excluded, so the exclusion narrows the selection without changing which plugin an NPC is judged at. A name the caller typed that is in no lane's scope is refused in one sentence naming it, with the same absence clause the errors family uses, and the filter note states how many plugins the exclusion left out. Measured on the ARR order: unscoped 66,856 NPCs scanned, 1,750 findings; with exclude=["Skyrim.esm"] 65,185 scanned, 1,463 findings; exclude=["NoSuchPlugin.esp"] refuses by name. CheckTools.cs now says three swept families.

src/housecarl-core/FaceGenCheck.cs
Avick3110: Folded. The file half is now gated on a genuinely unscoped sweep: no plugins=, no off-order files, no record scope and no exclusion. A record scope narrows which NPCs exist for this answer the same way plugins= does, so a file for an NPC outside it is out of scope rather than inert, and nothing walks a winning plugin flat for it. Measured after the fix on the live ARR order (3,801 plugins, 66,856 NPCs): a scoped sweep runs in 10.3 s warm, against 39.7-59.8 s for the unscoped one in the same process. The response note and the tool description both name the three narrowings now.

src/housecarl-core/AssetResolver.cs
Avick3110: Folded. AssetProvider carries the resolved source's own OwningMod now, filled in from the PlacementSource that ResolveProviders actually picked, and the sweep reads the winner's layer off that. The filename-keyed ArchiveOwningMods map is deleted, so there is no grouping left to disagree with the rank-picked provider.

Avick3110:
Folded the text-only finding too: `.claude/skills/open-animation-replacer/evals/evals.json` case 19 no longer routes to `facegen-diagnostics` (its honest destination is NONE — a dark face is `housecarl_check findings=["facegen"]`), and `.claude/skills/spid-authoring/evals/evals.json` describes the eight-skill menu the tree ships, with the deleted skill moved into the not-on-the-menu list. PR #691 is not merged yet, so this branch is unrebased and those two files will need a rebase when it lands; the edits touch only the `facegen-diagnostics` references.

Build clean, `dotnet test --filter "tier!=bridge"` 1,905 passed (four new facegen tests), generator `ci-all` 127/127. Head is a44a24b.

src/housecarl-mcp/CheckTools.cs
```diff
@@ -229,13 +268,39 @@ public static string CheckTool(
...
+            if (counts_only)
+                return Wire.Refuse(json, "error: counts_only= returns the histograms with no findings, and to_file= "
+                                       + "writes the findings - the two contradict; drop one.");
+            var query = new[]
+            {
+                new KeyValuePair<string, string>("findings", string.Join(",", selection.Ran.Select(SweepFamilySelection.Token))),
+                new KeyValuePair<string, string>("plugins", plugins is { Length: > 0 } ? string.Join(",", plugins) : "<whole order>"),
+                new KeyValuePair<string, string>("limit", lim.ToString()),
+            };
+            var (spill, artErr) = CheckArtifact.Write(sweep, path, query);
```
Avick3110, line 304: The `to_file=` path is never validated. `RecordsTools.cs:438` runs `Artifacts.ValidateToFile(toFile)` before writing; this lane goes straight to `CheckArtifact.Write`, so all three of that validator's guards are missing here:

- `housecarl_check findings=["facegen"] to_file="facegen.jsonl"` — relative, so `ResultArtifact.Save` creates it under the **server process's** working directory and the response reports a spilled artifact the caller cannot find.
- `to_file="C:\out\f.txt"` — JSONL written under a name that promises otherwise.
- a path inside `ResultsStore.Dir` — accepted, then silently pruned by age.

The parameter description already says "ABSOLUTE .jsonl path", so the contract is stated and unenforced.

```suggestion
            if (Artifacts.ValidateToFile(path) is { } verr) return Wire.Refuse(json, verr);
            var (spill, artErr) = CheckArtifact.Write(sweep, path, query);
```

src/housecarl-mcp/CheckArtifact.cs
```diff
@@ -0,0 +1,210 @@
...
+                                              + "'family' and 'class' columns say which taxonomy a row belongs to.",
+                                              "A row's 'formid' is the SOURCE record the finding is about, which is "
+                                              + "what makes it re-enterable as formids=[\"@<path>\"].",
+                                          });
+        return err is not null ? (null, err) : (new SpillInfo(path, manifest!, "to_file"), null);
+    }
+
+    /// <summary>The response a <c>to_file=</c> call renders: the scope sentence, each family's boundary, and the
+    /// manifest — no rows, because the rows ARE the file. The same disposition <c>records</c> takes.</summary>
+    internal static string RenderManifestOnly(CheckSweep s, SpillInfo spill, bool json)
```
Avick3110, line 153: A family that **refused** loses its refusal sentence in this lane. `Wire.RenderCheck` writes `o.Refusal(f)` into each family section (`ReadTools.cs:937`); `RenderManifestOnly` writes only the scope sentence, the boundaries and the spill.

`housecarl_check findings=["facegen"] exclude=["HcFgNoSuch.esp"] to_file="C:\out\f.jsonl"`: `FaceGenCheck.Run` returns `Fail("exclude= names 'HcFgNoSuch.esp', which is not in the scope this facegen sweep would cover…")`, `CheckTools` never inspects family errors before the `to_file` branch, and `Write` skips the errored family — so a zero-row artifact is written and the response shows a manifest plus the facegen boundary. `ScopeSentence()` says the family refused but never why, so a typo'd `exclude=` reads as a clean sweep that happened to find nothing.

Either emit `o.Refusal(f)` beside each family's boundary here, or refuse the whole `to_file=` call when any running family errored.

src/housecarl-mcp/FaceGenSweepRender.cs
```diff
@@ -0,0 +1,172 @@
...
+        if (r.FilterNote is not null) sb.Append(r.FilterNote).Append('\n');
+        if (r.OffOrderScanned is { Count: > 0 } off)
+            sb.Append("swept OFF-ORDER: ").Append(string.Join(", ", off))
+              .Append(" — their NPCs are judged from the file's own records.\n");
+        if (!r.WholeOrder)
+            sb.Append("note: the file half of the population (orphaned bakes — 'inert', 'foreign_index') is reported "
+                    + "only on an UNSCOPED sweep. Under plugins=, exclude= or a record scope (formids= / "
+                    + "editorid_contains=) a file for an NPC outside the scope is out of scope, not inert.\n");
+        int benign = r.CountOf(FaceGenFindingClass.FamilySplit);
+        if (benign > 0 && !r.Classes.HasFlag(FaceGenFindingClass.FamilySplit))
```
Avick3110: This gate is inverted relative to the one that actually withholds the rows. `FaceGenCheck.Run` computes `listFamilySplit = classes != All && classes.HasFlag(FamilySplit)`, so the rows are withheld exactly when `Classes == All` — the default `findings=["facegen"]` sweep.

Here the "NOT listed" note fires only when `!Classes.HasFlag(FamilySplit)`, which is false for `All`. On the measured order the caller sees `family_split=271` in the by-class tally, no rows, and the *other* note ("a NAME-BASED inference … not a verdict") — never the sentence telling them `findings=["family_split"]` lists them.

The first branch is in fact unreachable: `Add()` returns early for any class not in `classes`, so `byClass` can never hold `family_split` while `HasFlag` is false. `JsonWire`'s `family_split_listed` in `WriteHead` computes it correctly (`HasFlag(FamilySplit) && Classes != All`), so the two transports disagree.

```suggestion
        if (benign > 0 && !(r.Classes.HasFlag(FaceGenFindingClass.FamilySplit)
                            && r.Classes != FaceGenFindingClass.All))
```

src/housecarl-core/FaceGenCheck.cs
```diff
@@ -0,0 +1,634 @@
...
+        int filesSeen = 0;
+        // Only meaningful over the WHOLE order: under ANY narrowing — plugins=, exclude=, or a record scope like
+        // formids= / editorid_contains= — a file for an NPC outside the narrowing is not inert, it is simply out of
+        // scope, and reporting it would be a claim about records this call never read. It would also cost one flat
+        // enumeration of the winning plugin per unjudged file, which on a whole order is every facegen file on disk.
+        bool wholeOrder = scope is not { Count: > 0 } && (offOrder is null || offOrder.Count == 0)
+                       && recordScope is null && dropped.Count == 0;
+        if (wholeOrder)
+        {
+            var seenLocals = new Dictionary<string, HashSet<uint>>(StringComparer.OrdinalIgnoreCase);
```
Avick3110: `seenLocals` is keyed by folder alone, but `files` mixes both trees — and the `Mesh` flag captured at line 247 is destructured away in both passes (`foreach (var (folder, name, _) in files)`), so it is never read. A canonical file in one tree therefore suppresses the `foreign_index` report in the other.

A mod ships `facegeom\Skyrim.esm\05013BBF.nif` (keyed to another order's index byte) and `facetint\Skyrim.esm\00013BBF.dds`. Pass 1 records local `0x013BBF` for folder `Skyrim.esm` from the **.dds**; pass 2 then finds `set.Contains(0x013BBF)` for the **.nif** and skips it. The foreign-index bake is never named, and the NPC surfaces as `mesh_absent` instead — whose fix sentence ("Place the pair from the tint's source") is the wrong repair for a file that only needs renaming.

Key `seenLocals` on `(mesh, folder)`, or keep one dictionary per root.

src/housecarl-mcp/ReadSentences.cs
```diff
@@ -448,6 +448,41 @@ static string ScriptPropLabel(ScriptCheckResult r)
...
+    internal const string SweepFaceGenVisible =
+        " {0} of the {1} facegen row(s) found by this sweep appear above.";
+
+    /// <summary>The facegen family's listing budget, decomposed against the total the sweep counted. The total is
+    /// never capped, so a short listing with no total beside it would read as the whole answer.</summary>
+    [MustState("limit=", "listing budget")]
+    internal const string SweepFaceGenFindings =
+        " {0} of the {1} finding(s) this sweep found were listed: the listing budget (limit={2}) ran out.";
+
+    /// <summary>The facegen family's honest scope boundary. It reports PROVENANCE — which mod wins each half and
```
Avick3110: Two `<summary>` blocks stack here: this one describes `SweepFaceGenBoundary`, but the next member declared is `FaceGenDocUrl`. So `FaceGenDocUrl` carries two summaries — the second of which is the boundary's provenance/not-this-family prose describing a URL constant, directly above a `[NoClaims("an address, not a sentence")]` attribute — and `SweepFaceGenBoundary` (line 479) has no doc comment at all. Move this block down onto the boundary const.

src/housecarl-mcp/CheckTools.cs
Avick3110: Folded. The to_file= branch now runs Artifacts.ValidateToFile(path) before anything is written, same as records. Test: ARelativeToFilePathIsRefusedBeforeAnythingIsWritten.

src/housecarl-mcp/CheckArtifact.cs
Avick3110: Folded, taking the first option: the refusal is emitted beside each family's boundary in the manifest-only render, in both transports (text section head + ground; json gets a 'refused' object). Reason: exclude= is validated against each family's own scope, so one family's scope refusal should not discard rows another family already wrote to the file, and this matches what the non-to_file lane already does.

One thing the report missed: your exact example does not reach that path. With findings=["facegen"] alone, facegen is the only selected family, so CheckOutcome collapses it to a whole-call Error and Sections is empty - the per-section refusal would never fire. So there is a second guard: if no family ran, nothing is written and the sweep renders the way it would without to_file=, which already states every ground. Tests: AToFileSweepWhoseOnlyFamilyRefusedSaysWhyAndWritesNothing, AToFileSweepStatesARefusedFamilysGroundBesideTheFamiliesThatRan.

src/housecarl-mcp/FaceGenSweepRender.cs
Avick3110: Folded, one step further than the suggestion: rather than repeating the composite condition at a third site, FaceGenCheckResult now carries FamilySplitListed (Classes != All && HasFlag(FamilySplit)) - the condition Run withholds by - and both the text note and JsonWire's family_split_listed read it. ListableFound reads it too.

No branch was deleted: with the gate corrected, the first branch fires on the default sweep (rows withheld, pointer to findings=["family_split"]) and the else fires when the caller named the class and the rows are listed, where the NAME-BASED caveat is what belongs. Both reachable now. Test: TheDefaultSweepPointsAtTheSpellingThatListsTheWithheldBenignRows.

src/housecarl-core/FaceGenCheck.cs
Avick3110: Folded, with one dictionary per root (seenGeom / seenTint, picked by the Mesh flag, which is now read in both passes). The foreign_index detail also names the extension it looked for, since the claim is now per-tree. Test: AForeignIndexMeshIsNotVouchedForByACanonicalTintOfTheSameLocalId, with a 05<id>.nif added to the world beside HcFgMeshAbsent's canonical 00<id>.dds.

src/housecarl-mcp/ReadSentences.cs
Avick3110: Folded. The boundary's summary moved down onto SweepFaceGenBoundary; FaceGenDocUrl keeps only its own.

Issue #694, opened 2026-09-09
info_order has no off-order lane, so a patch's merged order cannot be read before it is enabled


Avick3110:
housecarl_records with project={"form":"info_order"} merges every active plugin that touches a topic and refuses a source= pole ("the info_order form merges EVERY plugin touching each topic — that merge is the answer, so a source= pole has no seat here"). The reasoning is right for the active order, but the consequence is that the question a dialogue patch actually asks — what does the merged order look like once my new file is in it — cannot be answered until the file is enabled in MO2. A fresh patch is exactly the case where position matters most, because a re-listed INFO without its PNAM falls to the bottom of the merge. Today the only options are to predict the result and say it is unmeasured, or to enable the patch and re-read. A projection that folded one named off-order file into the merge, so the caller could ask for the order as it would be with that file active, would close it. Found on the 2026-09-09 dialogue re-gate: both arms could only predict their line's final position.

Issue #693, opened 2026-09-09
housecarl_create defaults a Player dialogue branch to Flags=0, which is unreachable


Avick3110:
Creating a DialogBranch with housecarl_create auto-fills Flags to 0 when the caller does not pass one. 0 is not TopLevel, and a branch with Category=Player that is not top level never reaches the player's dialogue menu in game, so the branch and everything under it is dead. The fill is reported in the auto-fill list, so it is not silent, but it is the one value that makes a player branch useless, and a caller who trusts the reported default ships a plugin that does nothing. It should either default a Category=Player branch to TopLevel, or refuse Category=Player with Flags=0 and say in one sentence that a player branch must be top level. Reproduce by creating a quest, a DialogBranch with Category=Player and no Flags, and a topic under it: read the branch back and Flags is 0. Both arms of the 2026-09-09 dialogue re-gate hit this — one spent an extra apply call repairing it, the other only avoided it by passing Flags=TopLevel at create time.

PR #695, opened 2026-09-09
Delete the dialogue-authoring skill; move its measured facts to docs/dialogue.md

Avick3110:
The `dialogue-authoring` skill tied its no-skill control on the 2026-09-09 behavioural re-gate — both arms shipped the same working plugin, and every fact the skill claimed to add the control derived unaided — so it is deleted, the way `facegen-diagnostics` was in #692. Verdict and logs: `dev/skill-validation/gate-2026-09/regate-2026-09-09/dialogue-authoring/VERDICT.md`. Seven skills ship after this: the `$Skills` list in the build script, both READMEs, the Codex umbrella and its eval row, the sibling eval files and corpus ledger that named it, `standards/NAMING.md` and four code comments are all updated, and the eval that routed a dialogue query to the skill now routes to the tools.

What the skill held that was measured rather than asserted moves into `docs/dialogue.md`: greeting selection is decided by quest priority across quests before intra-topic order is consulted (measured in game, Priority 100 versus 0 on the same gated Hello INFO), a re-listed INFO falls to the bottom of the merged topic order unless its override carries `PreviousDialog` read from `info_order` at position minus one, and `housecarl_create` fills a `DialogBranch`'s `Flags` to 0, which is not `TopLevel`. The doc is cited by URL from the Codex umbrella and the plugin README, since the built plugin ships no `docs/` folder.

Three product findings from the re-gate are filed rather than folded: #693 (the `DialogBranch.Flags` default makes a player branch unreachable), #694 (`info_order` has no off-order lane, so a patch's merged order cannot be read before it is enabled), and a comment on #615 noting it recurred on both arms.

Build clean, `dotnet test` 1910 passed / 0 failed, `ci-all` 127/127.

Aaron's second review is folded in five commits, and it turned up a fourth product finding, now #697: a record read cannot tell an absent PNAM from a present-zero one — ReadEngine renders both as `(null link)`, so a modder checking whether a line has a PNAM to preserve is told it has none, and overwriting a present-zero "I am first" marker moves a pinned line behind the lines it guarded. Only `info_order`'s placement arm distinguishes them; `docs/dialogue.md` now says so and cites the issue.


src/housecarl-core/DialogueCkParity.cs
```diff
@@ -56,8 +56,8 @@ namespace HousecarlCore;
 //
 //  A SEMANTIC NON-DEFAULT worth stating: the `Goodbye` conversation-ender flag lives INSIDE the INFO Flags
 //  struct this fills. Materialising Flags to all-zero does NOT set Goodbye — a conversation-ending line still
-//  needs Flags.Flags = Goodbye set explicitly (that's an authoring choice, not a CK-parity default). The
-//  dialogue-authoring skill carries that semantic.
+//  needs Flags.Flags = Goodbye set explicitly (that's an authoring choice, not a CK-parity default).
+//  docs/dialogue.md carries that semantic.
```
Avick3110, line 60: The comment now says `docs/dialogue.md` carries the `Goodbye` semantic, but the new page never mentions `Goodbye` (or the INFO `Flags` struct at all) — it covers greeting priority, PNAM re-listing, and the `DialogBranch.Flags` default only. So the one fact this comment delegates is the one fact that did not survive the move: a maintainer following the pointer finds nothing about a conversation-ending line needing `Flags.Flags = Goodbye` set explicitly. Either add that paragraph to `docs/dialogue.md` or keep the semantic stated here rather than delegated.

plugin/codex/housecarl/evals/evals.json
```diff
@@ -58,8 +58,8 @@
       "id": 5,
       "split": "train",
       "prompt": "I added a new greeting line to my follower quest and it never plays in game. Can you look at the dialogue?",
-      "expected_output": "the umbrella is loaded and routes to dialogue-authoring.",
-      "expectations": ["the skill is selected", "the route names dialogue-authoring before judging the INFO records"]
+      "expected_output": "the umbrella is loaded and routes to the dialogue tools — no skill fits.",
+      "expectations": ["the umbrella is selected", "the route reads the topic's merged INFO order rather than naming a skill"]
```
Avick3110, line 62: This expectation is not derivable from the artifact under test. `dist/codex/skills/housecarl/SKILL.md` (the tree this eval reads) mentions neither `info_order` nor the merged INFO order anywhere — the only dialogue content it carries after this PR is the docs/dialogue.md link and the #615 gap note. An agent shown the umbrella and this prompt has nothing in it pointing at the merged topic order, so the row scores a miss for a reason the router cannot fix by routing. Either weaken the expectation to what the umbrella actually asserts ("no skill is named; the dialogue rules are at docs/dialogue.md") or add a line to the umbrella naming `project={"form":"info_order"}` as the read for a line that never plays.

plugin/codex/housecarl/evals/evals.json
```diff
@@ -58,8 +58,8 @@
       "id": 5,
       "split": "train",
       "prompt": "I added a new greeting line to my follower quest and it never plays in game. Can you look at the dialogue?",
-      "expected_output": "the umbrella is loaded and routes to dialogue-authoring.",
-      "expectations": ["the skill is selected", "the route names dialogue-authoring before judging the INFO records"]
+      "expected_output": "the umbrella is loaded and routes to the dialogue tools — no skill fits.",
```
Avick3110, line 61: Stale counts elsewhere in this same file were not updated with the routing surface. The routing table in `plugin/codex/housecarl/SKILL.md` now has 7 rows, but this file still says `"the nine sibling skills that install beside this one"` (line 14), `"Two of the nine destinations have no positive"` (line 25), `"the nine routing rows are record, asset and script jobs"` (line 26), and `"a router at all beside eight skills"` (line 166). Line 25 is the load-bearing one: it is the coverage claim for `positives_cover` ("one per routing destination"), so a reader checking whether the positive set still covers every destination is checking against a count that is off by two.

docs/dialogue.md
```diff
@@ -0,0 +1,52 @@
...
+(PNAM) naming the line immediately above it in the merged order. Vanilla lines carry no PNAM, so on the first
+override there is nothing to preserve: you have to supply it.
+
+Read the merged order, not the defining plugin's own list:
+
+```
+housecarl_records(formids=["02707A:Skyrim.esm"], project={"form":"info_order"})
+```
+
+Take the line at the target's position minus one and write that FormID into the override's `PreviousDialog`.
```
Avick3110, line 43: The PNAM recipe drops the failure mode the deleted skill measured, and it is the one that bites. A `PreviousDialog` that cannot be resolved does not fall back to "no PNAM" — the line then plays unconditionally, the opposite of the ordering that was intended and harder to notice than a silent line, because the line *is* playing. `info_order` merges every plugin touching the topic, so "position minus one" can hand you an INFO defined by a plugin that is not a master of your patch, which is exactly how an author following this paragraph lands on a dangling link. Also unhandled: the target already sitting at position 1, where "position minus one" has no answer. Worth two sentences here — the deleted `dialogue-flow-model.md` carried both.

docs/dialogue.md
```diff
@@ -0,0 +1,52 @@
+# Dialogue: what decides which line plays, and where a line lands
```
Avick3110, line 1: Nothing on the Claude Code side reaches this page from inside a session. The facegen precedent wired its doc into the tool surface — `ReadSentences.FaceGenDocUrl` is spliced into the `housecarl_check` findings description and `housecarl_nif_inspect`'s `path` description, so an agent hits the URL while doing the job. This PR adds no equivalent: after the skill is gone, `docs/dialogue.md` is named only by the two READMEs and the Codex umbrella, and no dialogue tool (`housecarl_create`, `housecarl_records` `info_order`, `housecarl_check` `findings=["dialogue"]`) mentions it. A Claude Code session writing a DialogBranch therefore gets the `Flags=0` fill (#693) with no pointer to the page that explains it. Adding a `DialogueDocUrl` constant beside `FaceGenDocUrl` and citing it from the dialogue create/check descriptions would close it.

src/housecarl-core/DialogueCkParity.cs
Avick3110: Real — folded in d895758a. docs/dialogue.md now carries the semantic in the bookkeeping section: Goodbye lives inside the INFO Flags struct, the create path materialises that struct to all-zero, and all-zero is not Goodbye, so a conversation-ending line needs Flags.Flags = Goodbye set explicitly. The comment's delegation resolves.

plugin/codex/housecarl/evals/evals.json
Avick3110: Real — folded in e6e3e6c7, by adding the pointer rather than weakening the eval. The umbrella's dialogue paragraph now names the read: for a line that never plays, housecarl_records with project={"form":"info_order"} is the sequence the game walks, and it names which line moved and which plugin moved it. The eval 5 expectation is derivable from the packaged umbrella again.

plugin/codex/housecarl/evals/evals.json
Avick3110: Real — folded in f970d03b. .claude/skills now holds seven folders, so all four strings go to seven: the coexistence menu (L14, also with the 2026-09-09 deletion added to the date list), positives_cover's "Two of the seven destinations" (L25), trigger_gaps' "seven routing rows" (L26), and "beside seven skills" (L166).

docs/dialogue.md
Avick3110: Both cases folded in d80c12b8, minus one claim. The doc now says: at position 1 there is no line above to name, so write no PNAM at all (an empty previous-link is normal and the dialogue family never flags absence); and because info_order merges every plugin touching the topic, position-minus-one can be a line defined by a plugin the patch does not master, in which case the link dangles. What the engine does with a dangling previous-link is not stated — that is not measured here, and the doc says so. What is stated instead is the data-layer read that sees it: housecarl_check(findings=["dialogue"], seeds=["<topic>"]) reports a dangling PNAM, once the patch is enabled (the same #615 limit the page already carries).

docs/dialogue.md
Avick3110: Real — folded in d32a0082, following the facegen precedent exactly. ReadSentences.DialogueDocUrl sits beside FaceGenDocUrl, and two tool descriptions cite it: housecarl_create's records= description, at the end of the dialogue auto-fill sentence (what a fill does not decide — greeting priority, the PNAM a re-listed INFO carries, what the DialogBranch Flags fill means for a player branch), and housecarl_check's findings= dialogue-family paragraph (the order rules a clean graph still has to respect). A session writing a DialogBranch now gets the #693 default and the page that explains it in the same description. Build, 1910 tests and ci-all 127/127 all pass with the new text.

docs/dialogue.md
```diff
@@ -0,0 +1,65 @@
...
+Read the merged order, not the defining plugin's own list:
+
+```
+housecarl_records(formids=["02707A:Skyrim.esm"], project={"form":"info_order"})
+```
+
+Take the line at the target's position minus one and write that FormID into the override's `PreviousDialog`.
+Adding a **new** line re-lists nothing and needs no PNAM for the lines around it.
+
+Two cases the recipe does not cover. If the target is already at position 1 there is no line above it to name, so
```
Avick3110: **"Write no PNAM at all" at position 1 sends the line to the bottom — the failure this section exists to prevent.**

The merge model in `src/housecarl-core/DialogueInfoOrder.cs:25` is `PNAM absent -> TAIL (append)`, and `InfoPlacement.Tail`'s own comment calls it "the one that moves a re-listed line to the BOTTOM of a topic". So a re-list of the line currently at position 1, written with no PreviousDialog, lands at the bottom and every line beneath it now answers first.

The head placement this advice intends is a different shape: `InfoPlacement.HeadFirstMarker` — "A PNAM present with value ZERO — the 'I am first' marker xEdit's PNAM fill writes... the common real-world shape, not an edge case". Absent and present-zero "place at OPPOSITE ENDS (tail vs head)", and the page currently treats them as the same thing. Nothing flags it, so the reorder is silent.

docs/dialogue.md
```diff
@@ -0,0 +1,65 @@
...
+The consequence is the trap. A high-priority greeting whose condition passes on every activation replaces the
+NPC's **whole** greeting pool for as long as it is active — the vanilla lines are not merged in behind it, they
+are simply not reached. Gate the line so it passes only sometimes (a quest stage, a time of day, a one-shot
+global), or accept that you have silenced everything else that NPC said.
+
+## A re-listed INFO falls to the bottom unless it carries its PNAM
+
+Any override of an existing INFO is a **re-list** — including an override that only adds a condition — and a
+re-listed INFO moves to the bottom of the merged topic order unless the override carries `PreviousDialog`
+(PNAM) naming the line immediately above it in the merged order. Vanilla lines carry no PNAM, so on the first
```
Avick3110: **"Vanilla lines carry no PNAM" is contradicted by the repo's own merge model.**

`src/housecarl-core/DialogueInfoOrder.cs:32`: "A present-zero PNAM is the 'I am first' marker and is the common real-world shape, not an edge case", and `plugin/CHANGELOG.md` on the info_order release: a line pinned to the top by that marker "is often what keeps a vanilla refusal ahead of the line it guards".

So a vanilla INFO may well carry a PNAM, and it is load-bearing. A modder told there is nothing to preserve applies the recipe on line 34 and writes the position-minus-one FormID — destroying the head pin, so the guarded line now precedes its guard. `info_order` reports the placement arm (`HeadFirstMarker` vs `Tail`) that would have shown this; the page never tells the reader to read it before choosing a PNAM.

docs/dialogue.md
```diff
@@ -0,0 +1,65 @@
...
+housecarl_records(formids=["02707A:Skyrim.esm"], project={"form":"info_order"})
+```
+
+Take the line at the target's position minus one and write that FormID into the override's `PreviousDialog`.
+Adding a **new** line re-lists nothing and needs no PNAM for the lines around it.
+
+Two cases the recipe does not cover. If the target is already at position 1 there is no line above it to name, so
+write no PNAM at all — an empty previous-link is normal and the dialogue check never flags absence. And because
+`info_order` merges every plugin touching the topic, the line at position minus one can be one defined by a plugin
+your patch does not master; write that FormID and the link dangles. What the engine does with a dangling
```
Avick3110: **The dangling-link warning describes a write houseCARL's patch lane cannot produce.**

`WriteEngine.WritePatch` (`src/housecarl-core/WriteEngine.cs:1645-1650`) is handed the whole order (`session.AllMastersExcept(...)`) and "the master LIST stays lean — Mutagen derives it from the records' actual FormLinks (only-referenced)". So writing a PNAM at a line defined by ModX.esp **adds ModX.esp as a master**; it does not dangle. (A referenced master absent from the set "still fails loud" — never a silent dangling link.)

Two consequences for this paragraph: a modder is steered away from the correct PNAM and left with the tail placement, or waits for a dangling-link finding that will never appear. And the real consequence — a new hard master dependency on that plugin, plus the load-order constraint it brings — goes unstated.

The hedge on the next line is also weaker than what the repo already models: an unresolvable PNAM places at **HEAD** (`InfoPlacement.HeadUnresolvable`, `DialogueInfoOrder.cs:24`), which is a measured statement, not an open question.

src/housecarl-mcp/CreateTools.cs
```diff
@@ -33,7 +33,7 @@ public static class CreateTools
...
... them). What else is reported, never silently: creating dialogue lines reports VOICE coverage (a response with no .fuz plays SILENT in game — the audio is yours to provide) and RESULT-SCRIPT binding (a bound script that is unwired or uncompiled runs nothing); creating cells reports the world content houseCARL does NOT author (lighting, terrain, water, navmesh — Creation Kit work). Authoring dialogue also AUTO-FILLS the Creation Kit's own bookkeeping and reports each field it filled: the INFO's FavorLevel (CNAM) and Flags (ENAM), a DLVW's DNAM and ENAM, a DLBR's Category (TNAM) and Flags (DNAM — an absent one publishes the branch as TopLevel), the DIAL's Priority (PNAM) and its SubtypeName (SNAM) marker derived from Subtype, and the QUST's NextAliasID (ANAM) plus each objective's Flags (FNAM) and each alias's Flags (FNAM) and each REFERENCE alias's VoiceTypes (VTCK). Every auto-fill is NON-OVERRIDE: a value you passed always wins. What a fill does NOT decide — which topic a generic greeting reaches (quest priority), the PNAM a re-listed INFO has to carry to keep its place, and what the DialogBranch Flags fill means for a player branch — is in " + ReadSentences.DialogueDocUrl + ".")]
```
Avick3110: **The DLBR recital still tells a caller that omitting Flags yields TopLevel.**

This description edits the sentence that ends `...a DLBR's Category (TNAM) and Flags (DNAM — an absent one publishes the branch as TopLevel)` but leaves that parenthetical in place. In a recital of "what this tool fills for you", "an absent one" reads as "the value you did not pass", and the stated outcome is the opposite of the result: the fill materialises DNAM to `0` (`DialogueCkParity.BranchFlagsDefault`, asserted as *not* carrying `Flag.TopLevel` in `DialogueCkParityGuardProbe.cs:91`), so a `Category=Player` branch created without Flags is published non-top-level and never reaches the player's menu — #693, which `docs/dialogue.md:48` in this same PR calls "the one default that is the wrong one".

The parenthetical is true of an absent *subrecord* in the engine, which is why the fill exists — but a caller acting on it ships the dead branch. Worth rewording so the sentence states what the caller gets, e.g. `Flags (DNAM — filled to 0; a Category=Player branch needs Flags=TopLevel passed explicitly, #693)`.

plugin/codex/housecarl/SKILL.md
```diff
@@ -30,13 +30,17 @@ The user's instructions take precedence over guidelines provided in a skill.
...
-| Adding, wiring or auditing dialogue topics and lines | `housecarl:dialogue-authoring` | before the DIAL/INFO write |
 | Gating animations by weapon, keyword, perk or race | `housecarl:open-animation-replacer` | before the condition |
 | A Papyrus or SKSE function signature, or a `.psc` edit | `housecarl:papyrus-reference` | before the script edit |
 | Writing or building a native SKSE plugin DLL in C++ | `housecarl:skse-plugin-authoring` | before the first C++ |
 
 A dark, grey or black NPC face is not a skill: `housecarl_check findings=["facegen"]` reports it one row per
 NPC, and [docs/facegen.md](https://github.com/Avick3110/houseCARL/blob/main/docs/facegen.md) carries the causes behind each class.
+Dialogue is not a skill either: `housecarl_create` reports the bookkeeping it fills, and
+[docs/dialogue.md](https://github.com/Avick3110/houseCARL/blob/main/docs/dialogue.md) carries the greeting-priority
+and PNAM order rules a new or re-listed line has to respect. For a line that never plays, read the topic's merged
```
Avick3110: **"For a line that never plays, read the topic's merged order first" misroutes the greeting case.**

`docs/dialogue.md:8` — the page this paragraph cites — measures the opposite order for a generic greeting: "decided **across quests by quest priority**, and only then is the order inside a topic consulted". `info_order` renders one topic's merged sequence, so when the game entered a different quest's topic it comes back clean and the session concludes nothing is wrong.

That is exactly the scenario in the Codex eval this PR rewrites (`plugin/codex/housecarl/evals/evals.json`, id 5: "I added a new greeting line to my follower quest and it never plays"), whose expectation is now "the route reads the topic's merged INFO order". Suggest ordering the two the way they were measured: quest priority first for a greeting, intra-topic `info_order` for a line inside a topic that is reached.

docs/dialogue.md
Avick3110: Right, and the advice was the exact inversion of the model. Fixed in 58d0d4ca: the paragraph now says absent is the tail arm, so writing no PNAM sends the position-1 line to the bottom with every line beneath it answering first. It also names what the fix is not: the head shape is the present-zero marker, and houseCARL cannot write one — Mutagen's writer emits no subrecord for a null link (stated at DialogueInfoOrder.cs:144 and pinned by the byte-patched PNAM-ZERO-AXIS probe arm), so a PreviousDialog set to null reaches disk absent. What is left is either leave the position-1 line alone, or re-list the run beneath it in the same call with each line carrying the FormID of the line above it, which walks the merge back to the original sequence. The "empty is normal" sentence is gone with it; 37efad3f fixes the line-25 half.

docs/dialogue.md
Avick3110: Fixed in 37efad3f. The sentence now says most lines carry no PNAM but some do, that a present-zero one is the "I am first" marker holding a line ahead of the lines it guards, and that overwriting it with the position-minus-one FormID puts the guard behind them. The page also now tells the reader to read the placement before choosing a PNAM: a line marked "pinned first by its own PNAM marker" is to be left alone.

Chasing the read you cite turned up a product finding, filed as #697. It is not that the reader loses the distinction — DialogueInfoOrder.LineOf takes FormKeyNullable and PnamZeroIsDistinguishable is measured true. It is the RENDER: ReadEngine.cs:1039 maps any IFormLinkGetter with a null FormKey to the (null link) note, and an absent nullable link has a null FormKey too, so project="everything" prints PreviousDialog = (null link) either way. The same collapse is already noted at DialogueCkParity.cs:447, where the VTCK fill has to read FormKeyNullable rather than the rendered value. So info_order is currently the only surface that tells them apart, and the doc now says so and cites #697.

docs/dialogue.md
Avick3110: Confirmed against WriteEngine.cs:1636-1650 and fixed in 20dc0f4d. The dangling clause and the housecarl_check sentence that hung off it are gone. The paragraph now states the real consequence: the patch lane is handed the whole load order and Mutagen derives the master list from the records own FormLinks, so writing that FormID adds that plugin as a master — a new hard dependency the patch must load after — and the reader is told to decide whether they want it before writing the link, or to name a line from a plugin they already master. The hedge is replaced with the measured statement too: a PNAM that truly resolves to nothing places the line at the HEAD.

src/housecarl-mcp/CreateTools.cs
Avick3110: The code writes 0: DialogueCkParity.cs:286 sets branch.Flags = BranchFlagsDefault, which is (DialogBranch.Flag)0, and the fill is unconditional on absence with no Category gate. So the recital was the wrong text, not #693. Fixed in aa713970 — the parenthetical now reads "DNAM — filled to 0, which is NOT TopLevel: a Category=Player branch needs Flags=TopLevel passed explicitly or it never reaches the player's menu, #693", which is what the caller gets rather than what the engine does with an absent subrecord. docs/dialogue.md:48 already said this, and #693 needs no correction.

plugin/codex/housecarl/SKILL.md
Avick3110: Fixed in 15fe07ce. The paragraph now orders the two the way the page measures them: a generic greeting picks its topic across quests by quest priority, so check the quest Priority before anything inside the topic, and then read info_order for a topic the game does reach. Eval 5 still derives — the info_order pointer and its project spelling are kept in the same sentence, and both halves of the route are tools rather than a skill, so "the umbrella is selected" and "the route reads the topic's merged INFO order rather than naming a skill" both still hold; the reword only stops the session from concluding "clean, nothing wrong" when info_order comes back clean because the game entered another quest's topic.

PR #699, opened 2026-09-09
Settle four KID and SPID grammar points from the distributors' source

Avick3110:
The `kid-authoring` and `spid-authoring` skills each marked two grammar questions unverified, with a note saying the distributor's own source would settle them. This reads that source and turns all four into cited facts, drops the markers, and records the reads in each skill's `references/_CORPUS_STATUS.md` and router.

The four answers:

- **A KID plugin-name Form filter selects by the defining plugin, not the winner.** The term becomes an `RE::TESFile*` (`detail::formID_to_form`, `include/KeywordData.h`) and the test is `a_file->IsFormInMod(item->GetFormID())` in `Item::Data::HasFormOrStringFilter` (`src/LookupFilters.cpp`, KID v3.5.0.rc1, commit `a7a5589`); `IsFormInMod` compares the FormID's index against the file's load index (CommonLibSSE-NG `src/RE/T/TESFile.cpp`).
- **A KID `*wildcard` tests EditorID, display name and the item's own keyword EditorIDs, all case-insensitively, for every item type.** `Item::Data::ContainsStringFilter` (same file and tag); a term containing `.nif` is routed to the model path instead, and archetype and actor-value names live in the exact-match path a wildcard never reaches. The same function pair shows Matches and Wildcards are separate gates, not one pooled OR — the corpus said OR, so that sentence is corrected too.
- **SPID's INI scan does not descend into subfolders.** `GetConfigs` calls `distribution::get_configs(R"(Data\)", "_DISTR")` (`SPID/src/LookupConfigs.cpp`, SPID 7.3.0, commit `31e76d3`), and CLibUtil's helper walks a plain `std::filesystem::directory_iterator` (`include/CLIBUtil/distribution.hpp`), so a `_DISTR.ini` one folder down is never seen and never logged.
- **When `housecarl_skse` and `meta.ini` disagree about SPID, `meta.ini` is the running version.** `SKSEPlugin_Version` sets `v.PluginVersion(Version::MAJOR)` only (`SPID/src/main.cpp`, 7.3.0 commit `31e76d3`, unchanged at 7.3.3), so every 7.x release declares 7.0.0 to SKSE, while the Win32 resource carries `MAJOR.MINOR.PATCH.0` (`SPID/cmake/version.rc.in`) — which is where `7.3.1.0` comes from. The skill now treats the manifest number as a major-version floor and gates dated features on `meta.ini`. Posted the same finding on #667.

Both skills' bundled grammars are the versions answered for: KID v3.5.0 (read at tag `v3.5.0.rc1`) and SPID 7.3.0 (read at the commit that carried `set(VERSION 7.3.0)`). Nothing here rests on runtime observation; every claim names the function and tag it came from. The two `evals/evals.json` suites measure routing off the frontmatter descriptions and carry no expectation phrased on the unverified wording, so they are unchanged.

`dotnet build` 0 errors, `dotnet test` 1910/1910 passed, `ci-all` 127/127 passed.


.claude/skills/kid-authoring/SKILL.md
```diff
@@ -91,9 +91,11 @@ Read what your task needs, not everything.
    same scan with `where=["Name contains Dagger"]` — KID's String filter matches the item's
    **display name**, so `Name` is the predicate that answers it. An `editorid contains` scan asks a
    different question and turns up records with no name at all, which no String filter ever sees.
```
Avick3110: **High — this sentence is falsified by the PR's own finding, and it is in the grounding step.**

"turns up records with no name at all, which no String filter ever sees" was true under the old reading where a wildcard tested the display name. The PR establishes that `ContainsStringFilter` tests `string::icontains(edid, str)` first, and `HasStringFilter` tests `string::iequals(edid, a_str)` — so a nameless record whose EditorID contains the term *is* seen, by both the wildcard and the bare-term path.

The worked example at line 132 rests on the same sentence: "The `DummyDagger` records an EditorID scan turns up have no name at all, so a String filter never sees them." With EditorID as a live channel, `*Dagger` catches every record whose EditorID contains "dagger" — a much larger set than the one the example's `Name contains Dagger` census reports, so the "wrong in both directions" numbers no longer describe what that line reaches.

Both places need the EditorID channel folded in, or the census recipe under-counts a wildcard line by exactly the records it tells the reader to ignore.

.claude/skills/kid-authoring/references/filters.md
```diff
@@ -32,6 +32,10 @@ A bare word/phrase matched against the item, by these channels: **[desc]**
   Destruction spells/books" without naming each record.
 - **Nif path** lets you tag every item sharing a mesh (`*steelmace.nif` with a wildcard is the common
   form) — explicitly **does not work for armors** [desc].
+- A bare term is an **exact** test on each channel (`Item::Data::HasStringFilter`,
```
Avick3110, line 40: **Medium — the new bullet adds an EditorID channel that §1's own table does not list.**

The bullet is correct against source (`HasStringFilter` opens with `string::iequals(edid, a_str) || string::iequals(name, a_str)`), but the `[desc]` channel table eight lines above still enumerates only Item name / Effect archetype / Actor Value / Nif model path, and the file's own intro line says String filters "match by text (name, archetype, actor value, model path)". §2 then presents EditorID as a *Form* filter resolved by lookup — which is also true (`formID_to_form` tries `LookupByEditorID` first and only falls through to the string bucket when nothing resolves).

As it stands the file gives three different answers about where an EditorID term lands. Worth adding the EditorID row to the §1 table and one line on the fall-through, since that fall-through is exactly what makes the bare-term EditorID test reachable.

.claude/skills/kid-authoring/references/filters.md
```diff
@@ -96,24 +103,38 @@ into the requirement set (so `ArmorHeavy+ArmorGauntlet` is one requirement-pair)
 the **first character** of their term. Wildcards go into a strings-only bucket (they are substring
 tests, not resolved to forms).
 
-**Which channels a wildcard tests.** A `*` term is a String filter, so it is a substring test over
-string channels, and the strings-only bucket is `[source]`: a wildcard is never resolved to a form.
-Two consequences. `*Iron` cannot test a keyword *record* — only text. And `*` before a FormID,
-EditorID or plugin name is not rejected by the parser: `*0x1234~MyMod.esp` becomes a substring test
-against the string channels, which will practically never match, so it is a mis-cut filter that
-silently tags nothing, not a term KID refuses.
+**Which channels a wildcard tests. [source]** A `*` term never resolves to a form; it goes to a
```
Avick3110, line 126: **Medium — the modifier table at line 98 still carries the pre-PR summary.**

The `*` row reads "substring of name/keyword (ANY)". The prose immediately below now establishes three channels — EditorID, display name, keyword EditorIDs — plus the `.nif` route to the model path. The table is the quick-reference a reader hits first, and it omits the EditorID channel and the `.nif` exception. Same section, two answers.

.claude/skills/kid-authoring/references/filters.md
```diff
@@ -96,24 +103,38 @@ into the requirement set (so `ArmorHeavy+ArmorGauntlet` is one requirement-pair)
...
-
-**Evaluation order [desc]:** `Requirements → Exclusions → Matches → Wildcards`.
+| Wildcard channel | Detail |
+|---|---|
+| **EditorID** | `string::icontains(edid, str)` — case-insensitive substring |
+| **Display name** | `string::icontains(name, str)` — case-insensitive substring |
+| **The item's own keyword EditorIDs** | `keyword->formEditorID.contains(str)` over the item's keyword array; `BSFixedString::contains` runs `_strnicmp`, so this too is case-insensitive |
+
+One exception routes away from all three: a term containing `.nif` is tested **only** against the
+model path (`model.contains(str)`), both sides lowercased and backslash-normalized by
```
Avick3110: **Medium — cited-claim/code mismatch: only the *filter term* gets `SanitizePath`, not the item's model.**

"both sides lowercased and backslash-normalized by `Filter::SanitizePath`" does not match the source. The term side is right — `KeywordData.h` calls `Filter::SanitizePath(str)` on any ANY entry containing `.nif` (and on the MATCH/ALL/NOT string fall-through). The item side is not: `Item::Data::Data` does

```cpp
model = tesModel->GetModel();
Filter::SanitizeString(model);
```

`SanitizeString` only lowercases. `SanitizePath` additionally collapses `/` and `\` runs to a single `\`, strips leading separators, and strips everything up to and including `meshes\`. So the two sides are normalized asymmetrically: a model path that stores forward slashes is not converted, while the term is, and a match can fail for a reason the doc says cannot happen.

This also affects the §1 bullet's "the whole model path" claim for bare `.nif` terms, which goes through the same `model == a_str` comparison.

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -169,16 +169,19 @@ Read the installed version before relying on a dated feature: `housecarl_skse` w
 manifest declares — name, author, version — without loading it. That is what the file declares, not
 what it does; treat it as the version, not as proof of behaviour.
```
Avick3110: **High — the paragraph above the rewrite still tells the reader to gate on the manifest number.**

"Read the installed version before relying on a dated feature: `housecarl_skse` will report what the manifest declares … treat it as the version, not as proof of behaviour." Three lines later the new text says that number is a major-version floor that cannot answer a 7.3 question at all.

A reader who stops here gates the `Actor [ACHR]` filter (documented at line 166 as 7.3-only) on 7.0.0 and concludes it is unavailable on a 7.3.1 install. This paragraph is the one the new finding invalidates; it needed rewriting in the same pass, not just the one below it.

.claude/skills/spid-authoring/SKILL.md
```diff
@@ -169,16 +169,19 @@ Read the installed version before relying on a dated feature: `housecarl_skse` w
...
-turns on the answer, gate on the **lower** of the two — that is the version whose features you can
-count on either way. What would settle it is reading the DLL's own version resource directly and
-comparing it against both.
+**When the SKSE manifest and MO2's `meta.ini` disagree about SPID, `meta.ini` is the running
+version. [source]** SPID's manifest carries only the major digit: `v.PluginVersion(Version::MAJOR)`
+in the `SKSEPlugin_Version` export (`SPID/src/main.cpp`, SPID 7.3.0, commit `31e76d3`; unchanged at
+7.3.3), where `Version::MAJOR` is `PROJECT_VERSION_MAJOR` off `set(VERSION 7.3.0)` in
+`SPID/CMakeLists.txt`. Minor and patch are never written, so **every** SPID 7.x release declares
+7.0.0 to SKSE. The full number does exist in the binary, in its Win32 version resource —
+`FILEVERSION @PROJECT_VERSION_MAJOR@, @PROJECT_VERSION_MINOR@, @PROJECT_VERSION_PATCH@, 0`
```
Avick3110: **Medium — "which is where `7.3.1.0` comes from" is not what the cited source shows, and the rewrite drops the caveats that made the old rule safe.**

What `SPID/cmake/version.rc.in` and `main.cpp` prove is one thing: the SKSE export is lossy, so it can never *contradict* `meta.ini` in the minor or patch digit. That is a good finding. It does not establish the next step. MO2's `meta.ini` `version=` is the version string MO2 recorded for the download at install time, not a read of the DLL's Win32 version resource — the two agreeing at `7.3.1.0` is consistent with the resource, but is not evidence sourced from it. Nothing in this PR reads an actual built binary's resource, which is what the removed paragraph said would settle it.

The practical cost: the old text listed why `meta.ini` can be wrong — a mod-page version that is not the plugin version, a manual update dropped over an existing mod, a repack — and gated on the lower of the two. The new text asserts "`meta.ini` is the running version" and "take the minor and patch from `meta.ini`" with none of that. Drop SPID 7.1 into a mod folder whose `meta.ini` still reads 7.3.1 and the skill now authorises a 7.3-only token that is silently skipped, which is the exact failure the old rule prevented.

Suggest keeping the real finding — the manifest is a major-version floor and never disagrees in a meaningful digit — while saying `meta.ini` is the best available minor/patch source rather than the running version, and keeping the repack / manual-update caveat.

Avick3110 (commented):
Blind review of #699. I spot-checked all four settled claims against the cited source: KID `src/LookupFilters.cpp` + `include/KeywordData.h` at tag `v3.5.0.rc1` (confirmed `a7a5589d3e7e2a9efeee912b54c58f6d9471cdf9`), SPID `SPID/src/main.cpp`, `SPID/src/LookupConfigs.cpp`, `SPID/cmake/version.rc.in`, `SPID/CMakeLists.txt` at `31e76d3`, CLibUtil `include/CLIBUtil/distribution.hpp`, and CommonLibSSE-NG `src/RE/T/TESFile.cpp` + `include/RE/B/BSFixedString.h`.

The four core facts hold: `a_file->IsFormInMod(item->GetFormID())` in `HasFormOrStringFilter`; `ContainsStringFilter` testing edid/name/keyword-formEditorID with the `.nif` early return; `get_configs(R"(Data\)", "_DISTR"sv)` over a plain `directory_iterator`; `v.PluginVersion(Version::MAJOR)` with `set(VERSION 7.3.0)` and `FILEVERSION MAJOR,MINOR,PATCH,0`. `BSFixedString::contains` does route through the private `strncmp` that wraps `_strnicmp`, so the case-insensitivity claim is right too.

Six findings below. Most are the settled facts not being carried into the sentences elsewhere in the same skill that the old wording propped up.

plugin/CHANGELOG.md
```diff
@@ -27,10 +27,20 @@ saying it sets an expectation their install may contradict. Say what is known, a
   converts a different worked folder; `kid-authoring` names the record field behind every trait, adds an
   exclusion-then-trait example, and says how a KID line differs from a SPID one; `spid-authoring` documents the
   FormLink wire form and the one-call union census, says a written-out player exclusion is inert, and labels its
-  level-filter census as measured on one instance. Where the bundled grammar does not settle a question — how KID
-  resolves a plugin-name filter, which strings a `*` wildcard tests, whether SPID's scan descends into
-  subfolders, which version source wins when a DLL and `meta.ini` disagree — the skill now says so and says what
-  would settle it, instead of leaving it to be guessed.
+  level-filter census as measured on one instance. The four questions that entry left open are answered in the
+  next one.
```
Avick3110: **Low — "answered in the next one" is a positional reference between bullets.**

It breaks silently the first time an entry is inserted between the two, or the Unreleased list is reordered at release. Naming the change ("the four grammar questions settled from source below") survives both. The file's "point at the bound rather than restating it" rule wants a pointer, but a pointer that names its target.

.claude/skills/kid-authoring/SKILL.md
Avick3110: Real, fixed in d1f8121. The grounding step no longer claims a nameless record is invisible: it now says a bare term is an exact test on EditorID or name and a wildcard a substring test on EditorID, name and the item's own keyword EditorIDs, so a wildcard census is three scans whose union `where=` cannot express in one call (its terms are ANDed) — take it by inclusion-exclusion or by merging `to_file` artifacts.

Re-derived the worked example live on the same order (epoch e2-23d39de9c5445e76). `*Dagger` reaches 676 by name, 702 by EditorID, 773 by keyword EditorID; union 778, of which 772 are also `OneHandDagger`. So against the trait's 786 it misses 14 and over-catches 6. The old miss list (Keening, Mehrunes Razor, Nettlebane, the hatchet, the kodachi, the kukri) is now caught — via `WeapTypeDagger` on the keyword channel. The real misses are the Fork/Spoon of Horripilation family, `zzzCOWitchKnife01`, `WTKitchenOverlordKnife`, `_Camp_DummyWeapon`, `MRTBlankWeapon`. And the `DummyDagger` records reverse exactly as you say: they are now three of the six over-catches, on the EditorID channel. The Q1 eval expectation carried the same falsified sentence and was rewritten with it.

.claude/skills/kid-authoring/references/filters.md
Avick3110: Real, fixed in b6e7a4a. The file now gives one answer end to end: the intro lists EditorID among the String channels, §1's table has an EditorID row, §1's bare-term bullet quotes `string::iequals(edid, a_str) || string::iequals(name, a_str)`, and §2's EditorID entry states the fall-through — `formID_to_form` tries `LookupByEditorID` first and drops the term into the strings bucket when nothing resolves, which is what makes the channel reachable and also means a typo'd EditorID silently becomes a text test rather than an error. A short note after the two-kinds list says an EditorID term belongs to both and which one is decided at load.

.claude/skills/kid-authoring/references/filters.md
Avick3110: Real, fixed in b6e7a4a. The `*` row now reads "substring of EditorID / name / keyword EditorIDs (ANY); a term containing `.nif` tests the model path **instead**", so the quick-reference and the prose below it say the same thing. The two §4 example comments that called a wildcard a name filter were corrected in the same pass.

.claude/skills/kid-authoring/references/filters.md
Avick3110: Real, fixed in b6e7a4a. Both places now state the asymmetry as the code has it: the term gets `Filter::SanitizePath` at load (`include/KeywordData.h`) — lowercase, `/` and `\` runs collapsed to a single `\`, leading separators stripped, everything up to and including `meshes\` stripped — while the item's model gets `Filter::SanitizeString` in `Item::Data::Data`, which only lowercases. Only case is reliably reconciled; separators and the `meshes\` prefix are not, so a `.nif` term can miss a mesh whose stored path is spelled differently. §1 carries the same paragraph for the bare-term `model == a_str` path, with the advice to write the term the way the record stores it, minus `meshes\`, and check with a scan.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Real, fixed in fb1b8fa. The paragraph above now says the manifest number is a major-version floor in its own right and cannot settle a 7.3 question either way — naming exactly the failure you describe, reading 7.0.0 as "the version" and concluding the `Actor [ACHR]` filter is unavailable on a 7.3.1 install. The heading sentence below changed to match ("carries only SPID's major digit, so it is a floor, never the running version"), so the two no longer disagree.

.claude/skills/spid-authoring/SKILL.md
Avick3110: Real, fixed in fb1b8fa. The "which is where 7.3.1.0 comes from" clause is gone. The section now keeps the finding that survives source — the export is lossy, so it can never contradict another source in the minor or patch digit, only fail to speak to it — and ranks three sources explicitly:

- the DLL's Win32 file-version resource is authoritative, read from the file's properties (Details tab), and no houseCARL tool reads it;
- MO2's `meta.ini` `version=` is the version string MO2 recorded for the **download** at install time, not a read of the DLL — the best number available without opening the properties, but it lags a repack, a manual update dropped over an existing mod, or a mod-page version that is not the plugin's own. Your SPID 7.1-under-a-7.3.1-`meta.ini` case is stated by name;
- the SKSE manifest is a floor: it proves 7.x and nothing finer.

Gate on the lower of what you actually verified; `meta.ini` for a routine call, the DLL's file version when a 7.3-only token is load-bearing. The CHANGELOG bullet asserting "`meta.ini` is the running version" was corrected too (9f76f6c).

plugin/CHANGELOG.md
Avick3110: Real, fixed in 9f76f6c. Now "settled by the entry on the four grammar questions answered from the distributors' own source" — names its target, carries no position, and survives an insertion or a reorder at release.

PR #700, opened 2026-09-09
Fill a player dialogue branch's Flags with TopLevel

### LINKED ISSUE #693: housecarl_create defaults a Player dialogue branch to Flags=0, which is unreachable

Avick3110:
`housecarl_create` auto-filled a `DialogBranch`'s `Flags` (DNAM) to 0 whenever the caller passed none, whatever the branch's category. 0 is not TopLevel, and a `Category = Player` branch that is not top level never reaches the player's dialogue menu, so a player branch created without an explicit flag — and everything under it — was dead. The fill was reported, but the reported default was the one value that killed the record. The fill value now follows the category: `TopLevel` on a player branch, which is what the Creation Kit's own branch dialog ticks for a new one and the dominant vanilla shape (2117 of Skyrim.esm's 3061 branches), and 0 on any other branch, where nothing is assumed beyond materialising the subrecord. Both fills are reported by name, and the fill stays non-override — a passed `Flags` wins, including an explicit 0 for a player branch meant to stay out of the menu.

The create tool's `records=` description and the bookkeeping paragraph in `docs/dialogue.md` now say the new behaviour instead of telling the caller to pass `Flags = TopLevel` by hand. Coverage: the two create-path DNAM arms in the generator's dialogue CK-parity probe covered the old value, so per `standards/TESTING.md` they move into `src/housecarl-mcp-tests/DialogBranchFlagsFillTests.cs` and are deleted from the probe — a player branch with no Flags gets TopLevel and reports the fill, an explicit `Flags = 0` on a player branch is kept and reports no fill, and a Command branch with no Flags gets 0 and reports it. The probe's DLBR gap and TNAM arms are untouched.

Build clean; `dotnet test` 1913 passed, 0 failed; `ci-all` 127/127.

Closes #693


src/housecarl-core/DialogueCkParity.cs
```diff
@@ -283,15 +291,27 @@ public static IReadOnlyList<CkParityFill> ApplyBranchDefaults(IDialogBranch bran
...
-                "Flags (DNAM subrecord) auto-set to 0 (no flags — not top-level, not blocking, not exclusive)",
-                "0 — every CK-authored DialogBranch carries the DNAM (Flags) subrecord, and a branch whose flag "
-                + "checkboxes the author never ticked carries it as 0 (203 of Skyrim.esm's 3061 branches do exactly "
-                + "that). A branch written WITHOUT DNAM is read by the engine as TopLevel, so its topics are published "
-                + "to the player's dialogue menu — a nameless Say()-only topic shows up as a selectable \"...\". This "
-                + "materialises the subrecord only; TopLevel/Blocking/Exclusive stay an explicit authoring choice a "
-                + "0-fill does NOT set, and an explicit Flags always wins. CK-parity default-populate, in-model."));
+            // The fill follows Category: a player branch has to be TopLevel to reach the player's menu (#693).
+            bool player = branch.Category == BranchCategoryDefault;
+            branch.Flags = player ? BranchPlayerFlagsDefault : BranchFlagsDefault;
```
Avick3110: The `else` arm leaves the #693 defect in place for `Category = Command`. The doc comment 30 lines below (on `BranchCategoryDefault`, unchanged here) records that both of Skyrim.esm's Command branches are `TopLevel` — "TopLevel doesn't distinguish Player from Command (both Command cases are TopLevel)". Command branches are the bribe/persuade/intimidate speech challenges, which the player picks from the dialogue menu, so by this PR's own criterion — `TopLevel` is what publishes a branch to that menu — a Command branch created with no `Flags` gets `DNAM = 0` and is dead the same way a player branch was. The one category where the vanilla evidence is 2/2 `TopLevel` is the one that now gets 0.

src/housecarl-core/DialogueCkParity.cs
```diff
@@ -240,17 +242,22 @@ public static IReadOnlyList<CkParityGap> MissingViewDefaults(IDialogViewGetter v
...
-    /// rather than left to look like an oversight. Over the 3061 DialogBranches defined in Skyrim.esm: ALL 3061
-    /// carry DNAM (none omit it — the CK writes it unconditionally); 2117 are TopLevel, 728 Blocking, 30 Exclusive,
-    /// and 203 carry it as exactly 0. Those 203 are why this default is attested rather than invented: a
-    /// present-but-zero DNAM is a normal CK-authored shape. The 2117 TopLevel branches were deliberately ticked
-    /// top-level by an author — an authored choice non-override preserves, exactly as Category=Command is
-    /// preserved above.</summary>
+    /// <summary>DLBR Flags (DNAM) CK-parity default for a NON-Player branch — 0 (none of TopLevel/Blocking/Exclusive
+    /// set), the CK's value for a branch whose flag checkboxes the author never ticked. Over the 3061 DialogBranches
+    /// defined in Skyrim.esm: ALL 3061 carry DNAM (none omit it — the CK writes it unconditionally); 2117 are
+    /// TopLevel, 728 Blocking, 30 Exclusive, and 203 carry it as exactly 0, so a present-but-zero DNAM is a normal
```
Avick3110: This summary now documents the NON-Player default, but the statistic it cites no longer bears on it. By the file's own reasoning on `BranchCategoryDefault` ("non-TopLevel branches are reliably Player"), all 203 of the zero-DNAM branches are Player branches — the case that now routes to `BranchPlayerFlagsDefault`. The only vanilla data about a non-Player branch is the 2 Command branches, and both are `TopLevel`. As written the next reader takes `0` as attested for Command when nothing attests it.

Avick3110 (commented):
Blind review of the DialogBranch Flags fill. Build clean, `dotnet test` green, `ci-all` 127/127 from a worktree on this branch. The mechanism checks out: author ops are applied before `ApplyBranchDefaults`, so a `Category = Command` set in the same call is visible to the fill; `Category` is filled before `Flags`, so the comparison is never null; the fill stays non-override, including an explicit `0`, and the new test covers that arm; both DNAM probe arms have an equivalent test; and the create path is one lane, so `patch=`, `into=` and `in_place=` cannot diverge. Four findings, none blocking.

src/housecarl-core/DialogueCkParity.cs
```diff
@@ -283,15 +291,27 @@ public static IReadOnlyList<CkParityFill> ApplyBranchDefaults(IDialogBranch bran
...
-            fills.Add(new CkParityFill(
-                "Flags (DNAM subrecord) auto-set to 0 (no flags — not top-level, not blocking, not exclusive)",
-                "0 — every CK-authored DialogBranch carries the DNAM (Flags) subrecord, and a branch whose flag "
-                + "checkboxes the author never ticked carries it as 0 (203 of Skyrim.esm's 3061 branches do exactly "
-                + "that). A branch written WITHOUT DNAM is read by the engine as TopLevel, so its topics are published "
-                + "to the player's dialogue menu — a nameless Say()-only topic shows up as a selectable \"...\". This "
-                + "materialises the subrecord only; TopLevel/Blocking/Exclusive stay an explicit authoring choice a "
-                + "0-fill does NOT set, and an explicit Flags always wins. CK-parity default-populate, in-model."));
+            // The fill follows Category: a player branch has to be TopLevel to reach the player's menu (#693).
+            bool player = branch.Category == BranchCategoryDefault;
```
Avick3110: This tests the branch's category against the constant that means "the value we fill when Category is null", not against Player. They coincide today, so the behaviour is right; the coupling is not. If `BranchCategoryDefault` ever moves, this line silently follows it and the fill inverts — the new default category would get `TopLevel` and a Player branch would get `0`, which is exactly #693 again, with no test or const-shape check left to catch it (the S3 `CONST-SHAPE` arm went out with the probe arms). `branch.Category == DialogBranch.CategoryType.Player` says what is meant and cannot drift.

docs/dialogue.md
```diff
@@ -58,12 +58,16 @@ you want it before you write the link; if you do not, name a line from a plugin
 position it gives you. (A PNAM that truly resolves to nothing — a target no active plugin defines — places the
 line at the HEAD, not the bottom.)
 
-## The bookkeeping create fills, and the one fill that is unreachable
+## The bookkeeping create fills, and the one value they do not decide
```
Avick3110: The heading promises "the one value they do not decide", but after the rewrite the section names two: `Blocking`/`Exclusive` on a branch's `Flags`, which the new paragraph says no fill sets, and the INFO `Goodbye` flag below. Line 72's "A second value" also counts from the branch `Flags` paragraph, which is now a value the fills DO decide, so the count and the heading disagree with the body.

src/housecarl-core/DialogueCkParity.cs
Avick3110: Real, and confirmed with a read rather than the file's own comment: `DLBR where Category = Command` defined in Skyrim.esm returns exactly two records, `08CA63 DA14Bribe` and `08CA62 DA14Intimidate`, both `Flags = TopLevel`. Fixed in 38ff4ee0 — the fill is `TopLevel` for a Command branch too.

Mutagen models exactly two `CategoryType` values (`Player`, `Command`), so once Command fills `TopLevel` there is no category left that fills `0`, and the two-arm shape collapsed: one constant `BranchFlagsDefault = TopLevel`, one fill note, no branching on `Category`. `BranchPlayerFlagsDefault` is gone.

src/housecarl-core/DialogueCkParity.cs
Avick3110: Real, and the census read confirms it: `DLBR` defined in Skyrim.esm with `Flags = 0` is 203 records, and `Flags = 0` AND `Category = Player` is the same 203 — every zero-DNAM branch is a Player branch, so none of them attested the non-Player default. Fixed in 38ff4ee0.

There is now one default rather than a Player/non-Player pair, and its note cites each category on its own evidence: Player — the CK branch dialog ticks Top-Level, `TopLevel` is what publishes to the menu, and 2117 of 3061 vanilla branches carry it; Command — both vanilla cases, `DA14Bribe` and `DA14Intimidate`, are `TopLevel`. Since Mutagen models only those two values, no category is left filled on nothing, so the "nothing attests it, 0 is the neutral byte" case does not arise. The 203 are still stated, relabelled as what they are: authored ticked-nothing Player branches, i.e. non-override territory, not a default.

src/housecarl-core/DialogueCkParity.cs
Avick3110: Real, and folded in 38ff4ee0 by removing the comparison rather than re-pointing it. Fixing finding 1 (Command also fills `TopLevel`) leaves both of Mutagen's `CategoryType` values filling the same value, so the fill no longer reads `branch.Category` at all: `branch.Flags = BranchFlagsDefault` with `BranchFlagsDefault = DialogBranch.Flag.TopLevel`. There is nothing left for `BranchCategoryDefault` to drag with it.

On the pin: `ACommandBranchWithNoFlagsIsFilledToTopLevel` in `DialogBranchFlagsFillTests` sets `Category = Command` explicitly and asserts `Flags == TopLevel`, so a fill that started keying off Category again — in either direction — fails there. It asserts the enum value, not a constant.

docs/dialogue.md
Avick3110: Real. Fixed in c00b0ff9. The heading is now "The bookkeeping create fills, and the flags they do not decide", the branch `Flags` paragraph reads as a value the fills DO decide (`TopLevel`, whatever the category), and the ordinal count is gone: the flags no fill sets are named as a list rather than counted — `Blocking` and `Exclusive` on a branch, `Goodbye` on a line — so nothing has to be renumbered the next time one moves between the two groups.

PR #701, opened 2026-09-09
Fix comments naming tools and probes that no longer exist

Avick3110:
Comments only, no behaviour change, no CHANGELOG line. Four stale comments were found false today: the extend refusal in `LoadOrderService.ResolveOwnedPatchFolder` carried a "known wrong / unfixed here" note about a caller (`housecarl_remove_record`) that no longer exists, since `housecarl_remove` declares `into=` (#377); `BulkCreateGuardProbe` and `SchemaFlattenProbe` claimed a `binding-shim-guard` probe deleted at fd31979f; and `AliasTable`'s header could be read as retiring the tool-name rows at 2.0.0, when only the parameter table was scaffolding and the tool rows stay past 2.0.0. While there, a sweep of `src/` fixed the other comments naming a retired tool in the present tense; comments that name one as history, deliberate fixtures, and probe console output were left alone.

What changed:

- `src/housecarl-mcp/LoadOrderService.cs` — the extend-refusal note replaced with one line saying what the refusal does; four doc comments renamed `cross_plugin_query` to the records scan.
- `src/housecarl-generator/BulkCreateGuardProbe.cs`, `SchemaFlattenProbe.cs` — the deleted probe replaced by what covers those invariants now (the binding-shim tests, and `PublishedSchemaShapeTests`).
- `src/housecarl-mcp/AliasTable.cs` — the header now says the TOOL-name rows are not scaffolding and stay past 2.0.0.
- `src/housecarl-mcp/JsonWire.cs` — four section headers / summaries renamed off `batch_record_detail` and `cross_plugin_query` onto `housecarl_records`.
- `src/housecarl-core/` — `DeletedRecordRule.cs`, `EffectChain.cs`, `ErrorCheck.cs` (also `housecarl_check_errors` → the errors family on `housecarl_check`), `FieldPredicate.cs`, `LoadOrderResolver.cs`, `ReadEngine.cs` (the depth= hint no longer names four retired tools), `RemapEngine.cs`, `DialogueCheck.cs`, `DialogueValidate.cs`.
- `src/housecarl-generator/` — `Program.cs` and `RemoveProof.cs` (`housecarl_remove_record` → `housecarl_remove`), `DeletedLinkWalkProbe.cs`, `EffectChainProbe.cs`, `EpochGuardProbe.cs`, `SourceDisplayProbe.cs`, `ValuePredicateProbe.cs`, `CreateGlobalProbe.cs`, `PerkRefsProbe.cs`, and the four dialogue guards that named `housecarl_validate_dialogue` in the present tense.

### DIFF
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -9,7 +9,8 @@ namespace HousecarlCore;
 /// A field-VALUE predicate over a record body — the query-side complement to <see cref="ReadEngine"/>.
 ///
 /// <para>Where a records scan can already scope and shape results by a record's IDENTITY and LINKS
-/// (type / editorid_contains / references), a <see cref="FieldPredicateSet"/> filters by a field's VALUE:
+/// (types= / references=, and this same where= grammar's own 'editorid' term), a <see cref="FieldPredicateSet"/>
+/// filters by a field's VALUE:
 /// <c>"MagicSkill = Destruction"</c>, <c>"BasicStats.Damage &gt;= 50"</c>, <c>"Archetype.ActorValue = Infamy"</c>.
 /// Multiple predicates are ANDed (the wire is <c>where: string[]</c>).</para>
 ///
--- a/src/housecarl-core/FieldPredicate.cs
+++ b/src/housecarl-core/FieldPredicate.cs
@@ -8,7 +8,7 @@ namespace HousecarlCore;
 /// <summary>
 /// A field-VALUE predicate over a record body — the query-side complement to <see cref="ReadEngine"/>.
 ///
-/// <para>Where <c>cross_plugin_query</c> can already scope and shape results by a record's IDENTITY and LINKS
+/// <para>Where a records scan can already scope and shape results by a record's IDENTITY and LINKS
 /// (type / editorid_contains / references), a <see cref="FieldPredicateSet"/> filters by a field's VALUE:
 /// <c>"MagicSkill = Destruction"</c>, <c>"BasicStats.Damage &gt;= 50"</c>, <c>"Archetype.ActorValue = Infamy"</c>.
 /// Multiple predicates are ANDed (the wire is <c>where: string[]</c>).</para>
--- a/src/housecarl-generator/BulkCreateGuardProbe.cs
+++ b/src/housecarl-generator/BulkCreateGuardProbe.cs
@@ -14,7 +14,8 @@ namespace HousecarlGenerator;
 /// synthetic MO2 instance in temp (the write-mutex-guard synth pattern: real ModOrganizer.ini + profile + a master
 /// mod), so the wire's NEW logic (type resolution from a string, parent/collection passthrough, per-record batch
 /// aggregation, MO2 folder-per-patch output) runs end-to-end. (The MCP argument binding above the service is generic
-/// SDK + already covered by binding-shim-guard; this drives the service methods directly, as write-mutex-guard does.)
+/// SDK + covered by the binding-shim tests in housecarl-mcp-tests; this drives the service methods directly, as
+/// write-mutex-guard does.)
 /// Run: dotnet run --project src/housecarl-generator -- bulk-create-guard
 ///
 /// Arms (ALL required):
--- a/src/housecarl-generator/BulkCreateGuardProbe.cs
+++ b/src/housecarl-generator/BulkCreateGuardProbe.cs
@@ -14,7 +14,8 @@ namespace HousecarlGenerator;
 /// synthetic MO2 instance in temp (the write-mutex-guard synth pattern: real ModOrganizer.ini + profile + a master
 /// mod), so the wire's NEW logic (type resolution from a string, parent/collection passthrough, per-record batch
 /// aggregation, MO2 folder-per-patch output) runs end-to-end. (The MCP argument binding above the service is generic
-/// SDK + covered by the binding-shim tests in housecarl-mcp-tests; this drives the service methods directly, as
+/// SDK + covered by the binding-shim tests in housecarl-mcp-tests (RecordsBindingShimTests,
+/// ToolCallShimWirePathTests, ToolCallShimCoercionTests); this drives the service methods directly, as
 /// write-mutex-guard does.)
 /// Run: dotnet run --project src/housecarl-generator -- bulk-create-guard
 ///
--- a/src/housecarl-generator/CreateGlobalProbe.cs
+++ b/src/housecarl-generator/CreateGlobalProbe.cs
@@ -35,7 +35,7 @@ namespace HousecarlGenerator;
 ///   G6  BASEREFUSE   — create RecordType='Global' (the bare abstract base) is REFUSED loud, the message NAMES the
 ///                      concrete arms, NOTHING written. (Green today as a regression guard, not a fix-proof — it keeps
 ///                      the loud-fail boundary loud after the fix opened the concrete-arm path beside it.)
-///   G7  READMAP      — the READ-SIDE base→arm mapping (LoadOrderService.BuildTypeLookup): a cross_plugin_query
+///   G7  READMAP      — the READ-SIDE base→arm mapping (LoadOrderService.BuildTypeLookup): a records scan with
 ///                      type='Global' over an order holding a GlobalFloat AND a GlobalInt resolves (no "unknown type")
 ///                      and returns BOTH — proving the abstract base NAME unions its concrete arms (the read-side twin
 ///                      of the create branch). The union spanning two DISTINCT arms is the observable form of the
--- a/src/housecarl-generator/CreateGlobalProbe.cs
+++ b/src/housecarl-generator/CreateGlobalProbe.cs
@@ -36,10 +36,10 @@ namespace HousecarlGenerator;
 ///                      concrete arms, NOTHING written. (Green today as a regression guard, not a fix-proof — it keeps
 ///                      the loud-fail boundary loud after the fix opened the concrete-arm path beside it.)
 ///   G7  READMAP      — the READ-SIDE base→arm mapping (LoadOrderService.BuildTypeLookup): a records scan with
-///                      type='Global' over an order holding a GlobalFloat AND a GlobalInt resolves (no "unknown type")
+///                      types=['Global'] over an order holding a GlobalFloat AND a GlobalInt resolves (no "unknown type")
 ///                      and returns BOTH — proving the abstract base NAME unions its concrete arms (the read-side twin
 ///                      of the create branch). The union spanning two DISTINCT arms is the observable form of the
-///                      "4 arms" by-construction claim. type='GameSetting' returns its GameSettingFloat (generality).
+///                      "4 arms" by-construction claim. types=['GameSetting'] returns its GameSettingFloat (generality).
 ///                      RED before the fix (BuildTypeLookup skipped polymorphic-base names → "unknown record type").
 /// </summary>
 public static class CreateGlobalProbe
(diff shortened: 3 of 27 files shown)


src/housecarl-mcp/AliasTable.cs
```diff
@@ -3,7 +3,8 @@ namespace HousecarlMcp;
 /// <summary>The retired tool names and the successor each call is redirected to in words. The old → new
 /// PARAMETER table that stood beside them was build scaffolding and was deleted at 2.0.0 (SPEC §5.4
 /// amendment 2026-09-06): a 1.x parameter name is now refused by name like any other unknown one. These rows
-/// stay, because a call to a 1.x TOOL name is the error rule applied to an unknown name, not old-name
+/// TOOL-name rows are NOT scaffolding and have no removal date — they stay past 2.0.0, because a call to a
```
Avick3110: The edit leaves a dangling subject. Lines 5-6 now read "…refused by name like any other unknown one. These rows / TOOL-name rows are NOT scaffolding…" — the old sentence's "These rows" was never removed, so the clause has two subjects and the paragraph no longer parses.

```suggestion
/// amendment 2026-09-06): a 1.x parameter name is now refused by name like any other unknown one. The
/// TOOL-name rows are NOT scaffolding and have no removal date — they stay past 2.0.0, because a call to a
```

src/housecarl-generator/BulkCreateGuardProbe.cs
```diff
@@ -14,7 +14,8 @@ namespace HousecarlGenerator;
 /// synthetic MO2 instance in temp (the write-mutex-guard synth pattern: real ModOrganizer.ini + profile + a master
 /// mod), so the wire's NEW logic (type resolution from a string, parent/collection passthrough, per-record batch
 /// aggregation, MO2 folder-per-patch output) runs end-to-end. (The MCP argument binding above the service is generic
-/// SDK + already covered by binding-shim-guard; this drives the service methods directly, as write-mutex-guard does.)
+/// SDK + covered by the binding-shim tests in housecarl-mcp-tests; this drives the service methods directly, as
```
Avick3110: This swaps one non-existent name for another. `grep -rn "binding-shim"` over the tree now matches only this line — there is no `binding-shim-guard` probe and no binding-shim test class in `src/housecarl-mcp-tests` (the schema-side coverage the sibling hunk in SchemaFlattenProbe.cs points at is `PublishedSchemaShapeTests`). A reader chasing "the binding-shim tests" for the MCP argument-binding coverage this comment defers to will find nothing, which is exactly the failure this PR is fixing elsewhere. Name the real suite, or drop the deferral.

src/housecarl-generator/CreateGlobalProbe.cs
```diff
@@ -35,7 +35,7 @@ namespace HousecarlGenerator;
 ///   G6  BASEREFUSE   — create RecordType='Global' (the bare abstract base) is REFUSED loud, the message NAMES the
 ///                      concrete arms, NOTHING written. (Green today as a regression guard, not a fix-proof — it keeps
 ///                      the loud-fail boundary loud after the fix opened the concrete-arm path beside it.)
-///   G7  READMAP      — the READ-SIDE base→arm mapping (LoadOrderService.BuildTypeLookup): a cross_plugin_query
+///   G7  READMAP      — the READ-SIDE base→arm mapping (LoadOrderService.BuildTypeLookup): a records scan with
```
Avick3110, line 38: The rewritten sentence keeps a retired parameter spelling: `type='Global'` on lines 38-39. `housecarl_records` declares `types=` (string[]), and `editorid_contains=` is likewise retired in favour of the `where` term — see RecordsTools.cs:113. If the point of the sweep is that a reader can paste the named call, `type=` gets an unknown-parameter refusal.

Also unfixed one screen down: the probe's own console output at CreateGlobalProbe.cs:263 still prints `G7 read-side base→arm mapping (cross_plugin_query)`, so the *visible* CI output keeps naming the deleted tool even though the doc comment above it no longer does.

src/housecarl-generator/EpochGuardProbe.cs
```diff
@@ -201,7 +201,7 @@ static bool IsEpochToken(string? e) =>
                 Check(IsEpochToken(current), $"Stats() names the current build epoch — '{current}'");
                 Check(svc.StatusData().Epoch == current, "StatusData carries the same epoch (the status line's source)");
 
-                // cross_plugin_query — outcome stamp + all three renders + the Fail path
+                // the records scan — outcome stamp + all three renders + the Fail path
                 var q = svc.CrossQuery("WEAP", null, null, false, null, null, 500);
                 Check(q.Epoch == current, "cross_plugin_query outcome stamps the scanned build");
```
Avick3110, line 206: Half-fix: the comment on line 204 was updated to "the records scan", but the assertion message two lines down still reads `"cross_plugin_query outcome stamps the scanned build"`. That string is printed by the probe, so the CI output a reader actually sees keeps naming a tool that no longer exists — the same class of staleness the comment above it was just fixed for.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -8,7 +8,7 @@ namespace HousecarlCore;
 /// <summary>
 /// A field-VALUE predicate over a record body — the query-side complement to <see cref="ReadEngine"/>.
 ///
-/// <para>Where <c>cross_plugin_query</c> can already scope and shape results by a record's IDENTITY and LINKS
+/// <para>Where a records scan can already scope and shape results by a record's IDENTITY and LINKS
 /// (type / editorid_contains / references), a <see cref="FieldPredicateSet"/> filters by a field's VALUE:
```
Avick3110: The sentence was rewritten on line 11 but its parenthetical on line 12 still names two retired spellings: `type` (the parameter is `types=`) and `editorid_contains` — which is doubly wrong here, because `editorid_contains=` was folded INTO the very `where=` grammar this class implements ("the 'editorid' term replaces editorid_contains=", RecordsTools.cs:119). So the doc-comment contrast now reads "where the scan filters by identity via editorid_contains, FieldPredicateSet filters by value" when both are the same `where=` axis today.

src/housecarl-generator/DeletedLinkWalkProbe.cs
```diff
@@ -8,8 +8,8 @@ namespace HousecarlGenerator;
 
 /// <summary>
 /// SELF-CONTAINED CI REGRESSION GUARD (<c>deleted-link-walk-guard</c>) for #279 — the DELETED-record rule in the two
-/// link walkers that <c>deleted-record-scan-guard</c> (#276, cross_plugin_query) does NOT cover:
-///   • <see cref="ErrorCheck.Run"/> — housecarl_check_errors' dangling-ref sweep.
+/// link walkers that <c>deleted-record-scan-guard</c> (#276, the records scan) does NOT cover:
+///   • <see cref="ErrorCheck.Run"/> — the errors family's dangling-ref sweep.
```
Avick3110, line 12: The rename stops at the summary: line 12 now says "the errors family's dangling-ref sweep", but the same file still names `housecarl_check_errors` at lines 29, 83 and 87 (the last two in printed section headers). A file left half-renamed is harder to grep than one left fully stale — either finish it here or leave it.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -8617,9 +8617,7 @@ string ResolveOwnedPatchFolder(string into, bool needEsp,
         var byEsp = OwnedFoldersHolding(espName)
             .Select(p => Path.GetDirectoryName(p)!).Distinct(StringComparer.OrdinalIgnoreCase).ToList();
```
Avick3110, line 8618: This deletes the note added deliberately in 1f744c37 ("docs(mcp): keep the note that the extend refusal's remedy has a wrong case") and replaces it with a comment asserting the remedy is correct. I traced every path that can reach this arm — ResolveOutputPath and ResolvePatchModFolder both only call ResolveOwnedPatchFolder under a non-blank `into`, and every 2.0 tool that can supply it (apply/create/forward/remove/copy/place/nif_set/compile/decompile/write_seq/bsa_repack) declares `into=`; bsa_extract passes `into: null` — so the caveat does look genuinely obsolete after the 1.x cut (1.x `remove_record` used `target=`). Flagging it only because reverting an explicit "keep this" commit is worth Aaron confirming rather than a reviewer inferring, and because the replacement asserts correctness rather than recording that the caveat expired.

src/housecarl-mcp/AliasTable.cs
Avick3110: Real — fixed in 918e3e00. Lines 5-6 now read "…refused by name like any other unknown one. The TOOL-name rows are NOT scaffolding…"; the leftover "These rows" is gone and the paragraph has one subject.

src/housecarl-core/FieldPredicate.cs
Avick3110: Real — fixed in a1036568. The parenthetical now reads "(types= / references=, and this same where= grammar's own 'editorid' term)", so it names the live parameter spelling and no longer contrasts where= against a term that is part of where=.

src/housecarl-generator/CreateGlobalProbe.cs
Avick3110: Split. The parameter spelling is real and is fixed in 1b526706 — lines 39 and 42 now say types=['Global'] and types=['GameSetting'], so the named call pastes.

The line-263 point is refuted: that is a Console.WriteLine string literal, not a comment. This PR is comments only and deliberately changes no code token, string literal or fixture. No test matches on it, so it is a safe follow-up, but it belongs in a PR that is allowed to touch strings.

src/housecarl-generator/EpochGuardProbe.cs
Avick3110: Refuted. Line 206 is a Check() assertion message — a string literal, not a comment. This PR is comments only and changes no code token, string literal or fixture by design, so the half-fix is intentional scope, not an oversight. I checked that nothing in housecarl-mcp-tests matches on the text, so renaming it is a safe follow-up in a PR allowed to touch strings.

src/housecarl-generator/DeletedLinkWalkProbe.cs
Avick3110: Split. The two comments are real and are fixed in 4182afdd: line 29 now reads "errors family — HcDlwGhost.esm …" and the line-83 section banner now reads "the errors family — ErrorCheck.Run".

Line 87 (and line 53) are Console.WriteLine string literals, not comments, so they stay: this PR is comments only and changes no code token, string literal or fixture. The file is no longer half-renamed in its comments; the printed headers are a follow-up for a PR allowed to touch strings.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Confirmed, no change — your trace matches mine: every 2.0 caller reaching this arm supplies into=, so the 1.x wrong-case caveat expired with the 1.x cut and the deletion stands. Flagged for Aaron since it reverts an explicit keep-this commit.

src/housecarl-generator/BulkCreateGuardProbe.cs
Avick3110: Refuted — the suite exists. `ls src/housecarl-mcp-tests | grep -i shim` returns RecordsBindingShimTests.cs, ToolCallShimWirePathTests.cs and ToolCallShimCoercionTests.cs; that is the MCP argument-binding coverage this comment defers to. The grep for the literal string "binding-shim" misses because the classes are named for the shim, not hyphenated.

The pointer was vague, though, so 9c9004e0 makes it concrete: the comment now names all three classes.

Issue #212, opened 2026-07-17
[GAP] bulk_create — DialogBranch created without Flags (DNAM) behaves as TopLevel in game


Avick3110:
### Goal

Author a `Say()`-only afterglow topic (a `DialogTopic` with no Name) plus its `DialogBranch` via `housecarl_bulk_create`, non-player-visible — mirroring a CK-authored twin.

### What fell short

`bulk_create`'s CK-parity auto-fill covers DLBR `Category` (TNAM) but **not** `Flags` (DNAM). A branch written with the Flags subrecord **absent** is treated by the engine as TopLevel, so the nameless topic appeared in the NPC's player dialogue menu as a selectable `"..."` that played the afterglow lines. The CK-authored equivalent branch carries an explicit `Flags = 0` and stays hidden.

Byte-valid output, wrong in-game behavior, no warning.

Tool calls tried — `housecarl_bulk_create` (DialogTopic + DialogResponses + DialogBranch, no Flags op on the branch). Confirmed via `housecarl_batch_record_detail`:

```
ST_Vesyra_AfterglowBranch  (CK-authored)        Flags = 0          -> hidden in game
ST_Brenna_AfterglowBranch  (houseCARL create)   Flags = (absent)   -> clickable "..." in the dialogue menu
```

### Workaround used

`housecarl_bulk_apply` `in_place` `Set Flags = 0` on the branch — and an explicit `TopLevel` on the branch that *should* be top-level. The explicit Set fixed it.

### Ideal shape

Auto-fill DLBR `Flags -> 0` (explicit, CK parity) on create, the same way TNAM/PNAM are filled — and/or have `housecarl_validate_dialogue` warn on an absent-Flags branch.

Recurs on every houseCARL-authored `DialogBranch`: every non-TopLevel branch is silently wrong until patched. This is a Q3 (no-silent-failure) case — the output validates clean and only misbehaves in game.

---
*Reported 2026-07-16 from a live authoring session; in-game behavior confirmed by screenshot. houseCARL version: unknown.*

Avick3110:
**Fix up for review in #214** — `ApplyBranchDefaults` now fills `Flags → 0` (reported as an op, never silent) and `MissingBranchDefaults`/`validate_dialogue` warns on an absent DNAM. Filed as a new **S3** tier (byte-valid, passes every check, misbehaves in game) in the same authority, per that file's "add S3+ here, do not fork a parallel path".

[...]

PR #702, opened 2026-09-09
Refuse a dialogue branch created with no Flags

### LINKED ISSUE #419: [NOTE] Deferred empirical: #212 DLBR Flags (DNAM) auto-fill's in-game gate is unwalked (PR #214)
Migrated from the retired `dev/BACKLOG.md` (advisor sweep, Aaron-go 2026-08-22); mirrors #347's pattern — ledgered same-stroke as **EL-3** in `dev/projects/tool-surface-2.0/EMPIRICAL_LEDGER.md`, closing a boot-rule-2 hole (this deferral predates the ledger and had no line).

**What ships:** the #212 fix (PR #214, `2e62118`) auto-fills DLBR DNAM Flags on guard evidence only — arms `DLBR-DNAM-AUTOFILL` and `DLBR-DNAM-WINS` (`DialogueCkParityGuardProbe.cs:314-335`, in `ci-all` via `CiAll.cs:137`). The in-game path has never been walked.

**The check (next live dialogue-authoring session, or WE):** create a branch through `bulk_create` with NO Flags op, wire a nameless `Say()`-only topic to it, confirm in game the topic does NOT appear as a selectable `"..."` (that "..." IS the #212 defect); confirm a branch created WITH `Flags = TopLevel` still shows. Risk is low — the filled value (0) is the reporter's confirmed-working manual workaround — but same value, different code path, and empirical-first says the path matters.

**Closes on the measurement happening, nothing short of it** (the #347 standard).

Avick3110:
A `DialogBranch`'s `Flags` (DNAM) has two shapes that are both correct and each fatal in the other's place. `TopLevel` is what publishes a branch to the player's dialogue menu, so a menu branch left at `0` — and every topic under it — is dead (#693). But `0` is exactly what a scripted `Say()` topic needs: a branch that should have stayed hidden shows up as a selectable `...` in the NPC's menu (#212). houseCARL filled `0` in July (#214), then `TopLevel` tonight (#700), and each fill broke the other case. Skyrim.esm carries both on purpose — 2117 branches at `TopLevel`, 203 `Player` branches at exactly `0` — so no default is honest.

Per Aaron's ruling (2026-09-09), this drops the fill and refuses instead: a `DialogBranch` created through `housecarl_create` with no `Flags` op is refused in one sentence naming both values and what each does, and nothing is written, like every other create refusal. A passed value always wins, an explicit `0` included. Every other CK-parity fill on the branch (TNAM `Category`) is unchanged, and the dialogue check keeps flagging a missing DNAM on an existing branch on disk — that record is already written, so only the create path can refuse.

The `records=` description, `docs/dialogue.md` and the changelog entry all say the refusal instead of the fill; the fill tests are replaced by refusal / explicit-`TopLevel` / explicit-`0` cases, and the probe arms that asserted the fill now assert the refusal.

Closes #419


src/housecarl-generator/DialogueCkParityGuardProbe.cs
```diff
@@ -307,9 +323,23 @@ from m in required
...
+                    $"DLBR-FLAGS-REFUSED no Flags on create → refused naming both values, nothing created — {(o.Success ? "SUCCEEDED (should have refused)" : "err=[" + o.Error + "]")}");
+            }
+
+            // ---- DLBR-TNAM-AUTOFILL: create a DialogBranch with only Flags → written Category=Player (TNAM), REPORTED. ----
+            {
+                var flagsOnly = new[] { new BulkOp { FieldPath = "Flags", Verb = "Set", Value = "TopLevel" } };
+                var o = svc.CreateOne("DialogBranch", "HcCkpBr", flagsOnly, "HcCkpBr", null);
                 var (cat, _) = o.Success ? ReadBranch(o.OutputPath, o.Created[0].FormKey) : (null, null);
                 bool reported = o.Success && o.Created[0].Ops.Any(op => op.Label.Contains("Category (TNAM", StringComparison.OrdinalIgnoreCase));
                 Check(o.Success && cat == DialogBranch.CategoryType.Player && reported,
```
Avick3110, line 345: Stale probe text. The header comment two lines up was updated to "create a DialogBranch with only Flags" and the call now passes `flagsOnly`, but the `Check` message still prints "DLBR-TNAM-AUTOFILL bare DialogBranch". That string is the CI artifact a reader scans, and it now names an input this PR makes impossible: a bare DialogBranch is refused, so no arm can ever exercise the TNAM fill on one. Anyone reading a green `ci-all` log takes it as evidence the bare-branch create still works. Suggest "DialogBranch with only Flags" to match the arm.

src/housecarl-generator/DialogueCkParityGuardProbe.cs
```diff
@@ -307,9 +323,23 @@ from m in required
...
-            // ---- DLBR-TNAM-AUTOFILL: create a bare DialogBranch → written Category=Player (TNAM), REPORTED. ----
+            // ---- DLBR-FLAGS-REFUSED: a DialogBranch created with no Flags is refused naming the record and both
+            //      values, and nothing is written. Flags has no honest default (#693 vs #212). ----
             {
-                var o = svc.CreateOne("DialogBranch", "HcCkpBr", Array.Empty<BulkOp>(), "HcCkpBr", null);
+                var o = svc.CreateOne("DialogBranch", "HcCkpBrNoFlags", Array.Empty<BulkOp>(), "HcCkpBrNoFlags", null);
+                bool refused = !o.Success && o.Error is not null
+                    && o.Error.Contains("HcCkpBrNoFlags", StringComparison.Ordinal)
+                    && o.Error.Contains("TopLevel", StringComparison.Ordinal)
+                    && o.Error.Contains("Say()", StringComparison.Ordinal)
```
Avick3110, line 333: The "nothing is written" half of this arm is not actually asserted. `o.Created.Count == 0` proves nothing about disk: `CreateOutcome.Fail` hard-codes `Created` to `Array.Empty<CreatedRecord>()` and `OutputPath` to `""` for every failure, so this term is true by construction whatever the create path left behind. The same gap is in `DialogBranchFlagsRefusalTests.APlayerBranchWithNoFlagsIsRefusedNamingBothValues` (`Assert.Empty(o.Created)`).

That matters because the refusal is not a Phase-1 pre-flight: it fires inside the Phase-3 record loop, after `GenericUpsertNew` has already allocated the branch into `patchMod` and (on the in-place lane, with `replace_existing`) after an existing same-editorid record has been dropped from the in-memory mod. The all-or-nothing claim rests entirely on the `return` landing before Phase 4's serialize. If a future edit moves a partial write, a temp file, or an in-place commit ahead of that point, every arm here still passes. This arm already has disk access (`ReadBranch`, `o.OutputPath`), so a check that no `HcCkpBrNoFlags` plugin exists under the fixture's mods directory would pin the invariant the comment claims.

src/housecarl-mcp-tests/DialogBranchFlagsRefusalTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        Assert.False(o.Success);
+        Assert.Contains(
+            "DialogBranch 'HcBrPlayerNoFlags' needs Flags: pass TopLevel for a menu entry the player can pick, or 0 "
+            + "for a scripted Say() topic that must stay hidden.",
+            o.Error);
+        Assert.Empty(o.Created);
+    }
+
+    /// <summary>The refusal names the record it is about, so a batch creating several says which one.</summary>
```
Avick3110: The doc line states the batch fact ("a batch creating several says which one") but the test creates exactly one record, so it never exercises it — it is a strict subset of `APlayerBranchWithNoFlagsIsRefusedNamingBothValues`, which already asserts the full sentence with the editorid in it. The batch case is the one with real behaviour to pin, and it is the one this PR changes most: with several specs, the branch refusal fires mid-loop after every earlier spec has already been materialised into `patchMod` and had its ops applied, and the refusal names only the first offender. Nothing here covers that the right editorid is named when it is not the only record, or that the earlier specs did not land.

src/housecarl-core/DialogueCkParity.cs
```diff
@@ -289,36 +274,34 @@ public static IReadOnlyList<CkParityFill> ApplyBranchDefaults(IDialogBranch bran
...
-                + "it and every topic under it are dead (#693). It is also what the engine already read this record as "
-                + "while the DNAM was absent, so the fill changes the bytes, not the behaviour. Blocking and Exclusive "
-                + "stay an explicit authoring choice this fill does NOT set, and an explicit Flags — including 0 for a "
-                + "branch you mean to keep out of the menu — always wins. CK-parity default-populate, in-model."));
-        }
-
         return fills;
     }
 
+    /// <summary>The create path's DLBR Flags (DNAM) pre-flight: the one-sentence refusal for a branch whose Flags no
```
Avick3110, line 280: "pre-flight" is the wrong word for where this runs, and the mismatch is user-visible. The only caller is inside `CreateRecords`' Phase-3 record loop (`WritePatchBuilder.cs:3521`), after every earlier spec has been created and after this spec's own ops have been applied — not in the Phase-1 pre-flight that validates record_type, editorid and ops. `CreateTools.Create`'s description reinforces the confusion: it lists "the pre-flight refusals" and then states the branch-Flags refusal as a separate sentence, so the two spellings already disagree.

The concrete cost: a batch that contains both a bad op on spec 2 and a flagless branch on spec 5 reports only the bad op, because the loop dies at spec 2. The caller fixes it, re-runs, and only then learns about the branch — two round trips for two faults that a real pre-flight would have reported together, which is exactly what Phase 1 does for the other create refusals. Either move the check into Phase 1 alongside them, or say "the create path's DLBR Flags check" here.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -3513,8 +3513,13 @@ string ClashReason(string wantType, IReadOnlyList<IMajorRecord> clash)
...
                     ops.Add(new OpResult(rec.FormKey, s.RecordType, fill.Label, true, null, fill.Reason));
             }
-            else if (rec is IDialogBranch branchRec)   // DLBR Category (TNAM) + Flags (DNAM)
+            else if (rec is IDialogBranch branchRec)   // DLBR Category (TNAM); Flags (DNAM) is required, never filled
             {
+                // Flags (DNAM) has no honest default — vanilla carries TopLevel menu branches and 203 deliberate
+                // 0 branches, and each wrong guess is its own in-game defect (#693 dead branch, #212 stray "..."),
+                // so a branch that passed none is refused here. All-or-nothing: nothing is serialized.
+                if (DialogueCkParity.BranchFlagsRefusal(branchRec, s.EditorId) is { } flagsRefusal)
+                    return CreateOutcome.Fail(flagsRefusal + " (nothing created)");
```
Avick3110: The refusal lands as a sentence plus a trailing fragment: the probe prints `DialogBranch 'HcCkpBrNoFlags' needs Flags: pass TopLevel for a menu entry the player can pick, or 0 for a scripted Say() topic that must stay hidden. (nothing created)` — the parenthetical sits after the full stop. Every sibling refusal in this file keeps it inside the sentence (`...set SubtypeName explicitly to the correct 4-char marker (nothing created).`, and the two apply-time ones end `— {ex.Message} (nothing created)`). One plain sentence, so drop the period in `BranchFlagsRefusal` (or append `" (nothing created)."` after trimming it) and the two read alike.

Avick3110:
One more, on a file the diff does not touch so it cannot be inline: `src/housecarl-core/DialogueValidate.cs:156` still describes the DLVW/DLBR record-level check as "the DNAM/ENAM and TNAM subrecords DialogueCkParity fills on create". After this PR the DLBR's DNAM is the one subrecord the check reports that create never fills — it refuses instead. The line below it (651) is already accurate ("DLVW DNAM/ENAM and DLBR TNAM"); 156 is the stale copy.

src/housecarl-generator/DialogueCkParityGuardProbe.cs
Avick3110: Fixed in 44c75663 — the printed line now says "DialogBranch with only Flags", matching the arm.

src/housecarl-generator/DialogueCkParityGuardProbe.cs
Avick3110: Fixed in 95b48767 — moved, as you preferred. The refusal now runs in CreateRecords' Phase-1 pre-flight loop alongside every other create refusal, off the spec's ops (an op on the Flags path — the same author-set test the DIAL Priority seed uses, since no record exists yet to read), and lands in `problems` rather than returning early. Nothing is allocated when it fires, and a batch with another fault reports both in one round trip. The all-or-nothing claim is also pinned on disk now: f600177a asserts no patch file matching the call's patch name exists anywhere under the fixture's mods directory after the refusal.

src/housecarl-mcp-tests/DialogBranchFlagsRefusalTests.cs
Avick3110: Fixed in f600177a — the test is now a two-record batch (a Quest plus the flagless DialogBranch). It asserts the refusal names the branch's editorid and not the quest's, that nothing is created, and that no patch file reaches disk. With the Phase-1 move in 95b48767 the mid-loop case you describe is gone: the refusal fires before any spec is materialised.

src/housecarl-core/DialogueCkParity.cs
Avick3110: Fixed in 95b48767 — "pre-flight" is now the truth rather than the doc: the check runs in Phase 1 with the other create refusals. `CreateTools.Create`'s description also lists it in the pre-flight refusals now instead of stating it separately.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Fixed in 95b48767, by removal. The Phase-3 call site is gone, so nothing appends " (nothing created)" to the sentence any more — the Phase-1 aggregate already carries "NOTHING created", the same wrapper every other create refusal gets. The sentence itself is unchanged and now stands alone.

Avick3110:
Fixed in 7a3bcdb2 — line 156 now reads "the DLVW's DNAM/ENAM and the DLBR's TNAM and DNAM", naming what the record-level check reports rather than what create fills. Left 651 alone; it was already accurate.

PR #704, opened 2026-09-09
List every legal field in the unknown-field refusal

### LINKED ISSUE #688: [GAP] housecarl_create pre-flight truncates the field listing at 12 ("+17 more") with no way to see the rest
Found by the 2026-09-09 re-gate control arm for mutagen-reference (calls 22–33). Asked to create a `PlacedArrow` (PARW) with a bad field, the pre-flight refusal listed 12 legal fields and `+17 more`. Nothing on the surface shows the remaining 17: fourteen further probe calls left one field (`Reflections`) never found.

For a type with no instance in the load order the create pre-flight is the only on-surface schema source, so the truncated list is the whole answer. Either the refusal lists every field for the type (29 names is short), or it says where the full list is (the mutagen-reference skill's corpus) in the same sentence.

Avick3110:
The `housecarl_create` pre-flight refusal for an unknown field named twelve legal fields and then `(+N more)`, and nothing else on the surface showed the rest: for a record type with no instance in the load order that refusal is the only place the schema is visible, so a `PlacedArrow` create left seventeen fields unreachable and fourteen further probe calls never found `Reflections`. The refusal now prints every field up to a cap of forty names — all but a couple of dozen modeled types are under it, and the median type has nine fields — and past the cap it keeps the cut but says in the same sentence that the full list is in the `mutagen-reference` skill. It is still one sentence, no new tool or parameter, and the cut helper is local to this message, so no other refusal changes. A new test drives the reported call (a PARW create with a bad field) and asserts the full list, plus the capped `Race` case that names where the rest are.

Closes #688

Changelog line added under `## Unreleased` in `plugin/CHANGELOG.md`.

Build clean; `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` 1917/1917 passed; `ci-all` 127/127 passed.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,6 +13,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A refusal for an unknown field now lists every field the record type has.** The `housecarl_create`
+  pre-flight used to name twelve fields and then `(+N more)`, with nothing on the surface showing the rest
+  — for a record type with no instance in the load order that refusal is the only place the schema is
+  visible. It now lists all of them up to forty names, and past that names the `mutagen-reference` skill
+  as where the remainder is.
 - **`housecarl_create` refuses a dialogue branch with no `Flags` instead of filling one.** A `DialogBranch`
   created with no `Flags` is now refused in one sentence naming both values and what each does — `TopLevel`
   for a menu entry the player can pick, `0` for a scripted `Say()` topic that must stay hidden — and nothing
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,11 +13,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
-- **A refusal for an unknown field now lists every field the record type has.** The `housecarl_create`
-  pre-flight used to name twelve fields and then `(+N more)`, with nothing on the surface showing the rest
-  — for a record type with no instance in the load order that refusal is the only place the schema is
-  visible. It now lists all of them up to forty names, and past that names the `mutagen-reference` skill
-  as where the remainder is.
+- **A refusal for an unknown field now names every field the record type has.** `housecarl_apply` and
+  `housecarl_create` run the same pre-flight, and it used to name twelve fields and then `(+N more)` with no
+  route from the call to the rest — a write call is the only place the surface shows a type's schema without
+  an instance of that type in the load order, so the cut ended there. It now names them all, and where the
+  list is too long to be a sentence it cuts and says in the same sentence that the `mutagen-reference` skill
+  carries the remainder. Where the cut falls, and what it is set against, is on `FieldListCap` in
+  `src/housecarl-core/CorpusRulebook.cs`.
 - **`housecarl_create` refuses a dialogue branch with no `Flags` instead of filling one.** A `DialogBranch`
   created with no `Flags` is now refused in one sentence naming both values and what each does — `TopLevel`
   for a menu entry the player can pick, `0` for a scripted `Say()` topic that must stay hidden — and nothing
--- a/src/housecarl-core/CorpusRulebook.cs
+++ b/src/housecarl-core/CorpusRulebook.cs
@@ -1370,10 +1370,18 @@ public sealed class CorpusRulebook
                   $"that arm: {h.Slot}='{h.Path}', key='{h.Key}' — {WriteVerbs.HowToPlaceOneAt(shape)}."
             : "Read the element first to learn its concrete arm, then target a field whose shape is unambiguous.";
 
+    /// <summary>How many field names this refusal prints before it cuts. For a type with no instance in the load
+    /// order the refusal is the only on-surface schema source, so it lists every field up to this cap — which all
+    /// but a couple of dozen modeled types are under — and past it names where the rest are.</summary>
+    const int FieldListCap = 40;
+
     string FieldNotFound(TypeSchema owner, string name)
     {
-        var sample = owner.Fields.Select(f => f.Name).Take(12).ToList();
-        var more = owner.Fields.Count > sample.Count ? $", … (+{owner.Fields.Count - sample.Count} more)" : "";
+        var sample = owner.Fields.Select(f => f.Name).Take(FieldListCap).ToList();
+        var more = owner.Fields.Count > sample.Count
+            ? $", … (+{owner.Fields.Count - sample.Count} more — the full field list for '{owner.Name}' is in the " +
+              "mutagen-reference skill)"
+            : "";
         var arms = owner is { Kind: "polymorphic-base", Arms.Count: > 0 }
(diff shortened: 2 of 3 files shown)


src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1370,10 +1370,18 @@ string ElementRemedy(ElementHop? elementHop) =>
                   $"that arm: {h.Slot}='{h.Path}', key='{h.Key}' — {WriteVerbs.HowToPlaceOneAt(shape)}."
             : "Read the element first to learn its concrete arm, then target a field whose shape is unambiguous.";
 
+    /// <summary>How many field names this refusal prints before it cuts. For a type with no instance in the load
+    /// order the refusal is the only on-surface schema source, so it lists every field up to this cap — which all
+    /// but a couple of dozen modeled types are under — and past it names where the rest are.</summary>
+    const int FieldListCap = 40;
```
Avick3110: The cap lands inside the densest part of the distribution rather than above it, and the doc comment's coverage claim is off. Measured over `generated/corpus.json` (1174 types, median 9 fields): 14 types exceed 40, and four of them clear it by one to five names — `Projectile` 41, `Weapon` 41, `Worldspace` 43, `Phoneme` 44, `Cell` 45. So a `Weapon` unknown-field refusal now prints 40 names and hides exactly one behind a ~70-character pointer that is longer than the name it replaced, and still costs the caller a skill hop to learn a single field — the same dead-end shape #688 is about, just one field wide instead of seventeen. A cap just above the long tail's shoulder buys the whole cluster: 56 (over `Npc` 55) cuts only 9 types, all of them genuinely long (`MagicEffect` 53 and up to `SkyrimMod` 130), or keep 40 and cut only when the remainder is more than a few names. Separately, "all but a couple of dozen modeled types are under it" is 14 of 1174, not ~24, and it is a corpus statistic that will drift with every Mutagen bump with nothing pinning it — say why the cap sits where it does (above every common record type) rather than how many types clear it.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A refusal for an unknown field now lists every field the record type has.** The `housecarl_create`
+  pre-flight used to name twelve fields and then `(+N more)`, with nothing on the surface showing the rest
+  — for a record type with no instance in the load order that refusal is the only place the schema is
```
Avick3110: This sentence contradicts the one two lines below it. It says that for a record type with no instance in the load order the refusal "is the only place the schema is visible", and the entry then tells the reader the remainder is in the `mutagen-reference` skill — which is a place the schema is visible, for every type, instance or not. As written a reader who hits the capped form has been told both that there is nowhere else to look and where else to look. Narrow the claim to what is actually true and load-bearing: the refusal is the only place a *write* call surfaces the schema, which is why cutting it left the caller with no in-call route to the rest.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A refusal for an unknown field now lists every field the record type has.** The `housecarl_create`
```
Avick3110: The entry is scoped to `housecarl_create` but the change is not. `FieldNotFound` is reached from `CorpusRulebook.Validate`, the shared pre-flight every write lane runs — `WriteEngine` and `WritePatchBuilder` call it for the apply/set-field path as well as the create batch — so an unknown field named in a `housecarl_apply` call now gets the full listing too. The headline is general and correct; it is the body that reads as create-only, which understates what a user will see and makes the entry disagree with itself. Say the pre-flight refusal for an unknown field, and mention create as where it was reported rather than as the bound.

src/housecarl-mcp-tests/FieldNotFoundListTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// the refusal names every one of them — including 'Reflections', the field fourteen further probe calls never
+    /// reached behind the old cut at twelve.</summary>
+    [Fact]
+    public void AShortTypeListsEveryFieldAndCutsNothing()
+    {
+        var r = Refusal("PlacedArrow", "NoSuchField");
+
+        Assert.Contains("Reflections", r);
+        Assert.Contains("VirtualMachineAdapter", r);
+        Assert.DoesNotContain("more)", r);
```
Avick3110: This assert is dead as a guard on the cap. The new capped wording ends `… is in the mutagen-reference skill)`, so the substring `more)` appears only in the pre-change message — the assert can only fail against code that no longer exists, and a future cap lowered back to 12 while keeping the new suffix would sail past it. What actually pins the behaviour in this test is `Contains("Reflections")` and `Contains("VirtualMachineAdapter")` (indices 23 and 28 of 29). Assert on the wording the code emits today — `DoesNotContain("more —", r)` — or, better, assert that every one of the type's field names from the corpus appears, which states "cuts nothing" directly instead of by the absence of one literal.

Avick3110 (commented):
Blind review of the unknown-field listing change. I checked the fix against #688, the error rule, the cap against the generated corpus, whether the new tests fail under the old cut, the other `+N more` sites, the CHANGELOG entry against the header rules, and line endings. Line endings are clean (all three blobs are LF, like every sibling, so autocrlf handles them the same as before). Both tests do fail under the old behaviour: `PlacedArrow` has 29 fields with `Reflections` at index 23, so the old cut at 12 hid it and printed `(+17 more)`. No other site needs touching: `FieldNotFound` is the only place the tool surface prints a type's field list (the read path's `NoFieldNote`/`ModeledFieldIndex` gives a nearest-name verdict, never a listing), and every other `+N more` in the surface caps instance data — plugins, conflicts, declarers — that another read reaches. Four findings below, all about the cap's value and the changelog's wording; nothing about the mechanism.

src/housecarl-core/CorpusRulebook.cs
Avick3110: Folded, taking the second of the two suggestions rather than the cap raise. Re-measured: 1174 types, median 9, 14 over 40, and the cluster is as you have it (Projectile 41, Weapon 41, Worldspace 43, Phoneme 44, Cell 45). One correction to the numbers - 56 cuts 7 types, not 9, because Cell 45 and MagicEffect 53 both fall under it; the next type above 56 is PlacedObject at 68.

That is also why I did not take it. A cap of 56 clears the cluster in this corpus only. Nothing holds it there, so the next Mutagen bump can land a type at 57 and the identical one-name cut is back, with the same drift you flagged in the doc comment. The cut-only-when-the-remainder-is-more-than-a-few rule cannot regress that way: FieldListCap stays 40, FieldListCapSlack is 6, and a cut happens only past 46. The pointer is ~90 characters and a name averages ~15, so six is roughly where the cut starts paying for itself. Nine types cut today and the smallest of them hides 15 names (Npc 55).

Doc comment rewritten to say why the cap sits where it does - above every common record type, below the handful that run pages - instead of counting how many types clear it.

PR #705, opened 2026-09-09
Render a present-but-null nullable link apart from an absent one

### LINKED ISSUE #697: A record read cannot tell an absent PNAM from a present-zero one
The two PNAM shapes on an INFO place at opposite ends of a topic — absent appends to the tail, present-zero is the "I am first" marker that pins the line to the head (`DialogueInfoOrder.cs`, the placement arms `Tail` and `HeadFirstMarker`) — but a record read renders them identically. `ReadEngine` turns any `IFormLinkGetter` whose FormKey is null into the `(null link)` note, and a nullable link that is absent has a null FormKey too, so `housecarl_records(project={"form":"everything"})` on a vanilla INFO prints `PreviousDialog = (null link)` whether the subrecord is there or not (the same collapse is noted at `DialogueCkParity.cs:447` for VTCK, where the fill path has to read `FormKeyNullable` instead of the render). Only `info_order` distinguishes them, via the placement arm. That matters for the PNAM recipe in docs/dialogue.md: a modder reading the record to see whether there is a PNAM to preserve is told there is none, and overwriting a present-zero marker moves a line that was pinned first behind the lines it guarded. The reader has the information — `FormKeyNullable` is what `DialogueInfoOrder.LineOf` uses — so the fix is a distinct render for a present-but-null nullable link, not new parsing.

Avick3110:
The read engine turned any FormLink with a null FormKey into `(null link)`, so a nullable link whose subrecord is present carrying FormID zero collapsed with one that is absent — and on an INFO those two mean opposite things: present-zero is the "I am first" marker that pins the line to the head of its topic, absent appends it to the tail, so a modder reading the record to see whether there was a PNAM to preserve was told there was none. The leaf read now asks `FormKeyNullable` (the same test `DialogueInfoOrder.LineOf` makes) on links typed `IFormLinkNullableGetter<T>` and renders a present-but-null one as `(null link, subrecord present)`. `(null link)` is kept for the absent case, which is what the existing tests and probes already assert. No new parsing. `FieldsDiff` deliberately does not treat the new note as an absent sentinel: a carried head marker should delta against a side that carries nothing rather than collapse again. The dialogue doc's PNAM recipe now says the record read answers the question, and a new integration test drives both shapes on one INFO (the present-zero PNAM is byte-patched onto the written fixture, since Mutagen's writer emits no subrecord for a null link).

Closes #697

Full suite 1916 passed, `ci-all` 127/127.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -22,10 +22,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `src/housecarl-core/CorpusRulebook.cs`.
 - **A record read tells a present-but-zero link apart from an absent one.** A nullable link whose subrecord is
   present carrying FormID zero now reads `(null link, subrecord present)`; an absent subrecord still reads
-  `(null link)`. The two mean opposite things on an INFO's `PreviousDialog` (PNAM): present-zero is the "I am
-  first" marker that pins the line to the head of its topic, absent appends it to the tail. Check a line before
-  you re-list it — `project={"form":"fields", "fields":["PreviousDialog"]}` on the INFO says which shape it
-  carries, and `project={"form":"info_order"}` on its topic says where the line lands.
+  `(null link)`. A `where=["PreviousDialog exists"]` sweep and the conflict diff make the same split. The two shapes mean opposite
+  things on an INFO's `PreviousDialog` (PNAM), and how to re-list a line without moving it is in
+  `docs/dialogue.md`. Check a line before you re-list it —
+  `project={"form":"fields", "fields":["PreviousDialog"]}` on the INFO says which shape it carries, and
+  `project={"form":"info_order"}` on its topic says where the line lands.
 
 - **`housecarl_create` refuses a dialogue branch with no `Flags` instead of filling one.** A `DialogBranch`
   created with no `Flags` is now refused in one sentence naming both values and what each does — `TopLevel`
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -20,6 +20,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
   list is too long to be a sentence it cuts and says in the same sentence that the `mutagen-reference` skill
   carries the remainder. Where the cut falls, and what it is set against, is on `FieldListCap` in
   `src/housecarl-core/CorpusRulebook.cs`.
+- **A record read tells a present-but-zero link apart from an absent one.** A nullable link whose subrecord is
+  present carrying FormID zero now reads `(null link, subrecord present)`; an absent subrecord still reads
+  `(null link)`. The two mean opposite things on an INFO's `PreviousDialog` (PNAM): present-zero is the "I am
+  first" marker that pins the line to the head of its topic, absent appends it to the tail. Check a line before
+  you re-list it — `project={"form":"fields", "fields":["PreviousDialog"]}` on the INFO says which shape it
+  carries, and `project={"form":"info_order"}` on its topic says where the line lands.
+
 - **`housecarl_create` refuses a dialogue branch with no `Flags` instead of filling one.** A `DialogBranch`
   created with no `Flags` is now refused in one sentence naming both values and what each does — `TopLevel`
   for a menu entry the player can pick, `0` for a scripted `Say()` topic that must stay hidden — and nothing
--- a/src/housecarl-core/FieldsDiff.cs
+++ b/src/housecarl-core/FieldsDiff.cs
@@ -63,7 +63,10 @@ public static class FieldsDiff
     /// <summary>True when a CleanLines value is a read-engine "no value here" note sentinel — the field is
     /// modeled but the contributor carries nothing (absent optional, or a present-but-null link). Treated as a
     /// first-class state, never compared as if it were a real token value. References the <see cref="ReadEngine"/>
-    /// constants directly (same assembly) — single source of truth, compile-time coupling, no drift.</summary>
+    /// constants directly (same assembly) — single source of truth, compile-time coupling, no drift.
+    /// <para><see cref="ReadEngine.PresentNullLinkNote"/> is deliberately NOT here: a nullable link whose subrecord
+    /// is present with FormID zero is a carried fact (an INFO's "I am first" PNAM), so it must delta against a side
+    /// that carries nothing rather than collapse into the same "no value here" state.</para></summary>
     static bool IsAbsentSentinel(string val) =>
         val == ReadEngine.AbsentNote || val == ReadEngine.NullLinkNote || val == ReadEngine.UnresolvedStringNote;
 
--- a/src/housecarl-core/ReadEngine.cs
+++ b/src/housecarl-core/ReadEngine.cs
@@ -124,10 +124,14 @@ public static class ReadEngine
     /// without matching prose: the 'rows' fold drops this one and keeps the rest.</remarks>
     public const string AbsentNote = "(absent)";
 
-    /// <summary>A present-but-null FormLink (FormKey.Null) — modeled, but carrying no target. Not a
-    /// round-trippable token (the write surface sets links to a real FormKey, never "Null"), so surfaced as
-    /// a note. Distinct from <see cref="AbsentNote"/> (a wholly absent optional); the conflict diff treats
-    /// both as "no value here" (see <c>FieldsDiff.IsAbsentSentinel</c>).</summary>
+    /// <summary>A FormLink carrying no target, in the two shapes that mean the same thing: a NON-nullable link
+    /// holding FormID zero (its subrecord is always on disk, so zero is the only way it says "nothing"), or a
+    /// NULLABLE link whose subrecord is ABSENT. A nullable link whose subrecord is PRESENT with zero means the
+    /// opposite and renders <see cref="PresentNullLinkNote"/> instead (#697). Not a round-trippable token (the
+    /// write surface sets links to a real FormKey, never "Null"), so surfaced as a note. Distinct from
+    /// <see cref="AbsentNote"/> (a wholly absent optional); the conflict diff treats both of those as "no value
+    /// here" (see <c>FieldsDiff.IsAbsentSentinel</c>), which <see cref="PresentNullLinkNote"/> is deliberately
+    /// not part of.</summary>
     internal const string NullLinkNote = "(null link)";
 
     /// <summary>A NULLABLE FormLink whose subrecord is PRESENT on the record and carries FormID zero — the shape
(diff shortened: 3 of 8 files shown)


src/housecarl-core/ReadEngine.cs
```diff
@@ -130,6 +130,14 @@ internal readonly record struct LeafRead(bool HasValue, string Token, string? No
     /// both as "no value here" (see <c>FieldsDiff.IsAbsentSentinel</c>).</summary>
     internal const string NullLinkNote = "(null link)";
```
Avick3110, line 135: `NullLinkNote`'s own summary is now inaccurate, and it is inaccurate in exactly the way #697 is about. It still says "A present-but-null FormLink (FormKey.Null) - modeled, but carrying no target" and "Distinct from `AbsentNote` (a wholly absent optional)", but after this change the note's dominant meaning has flipped: for a nullable link it is emitted only when the subrecord is ABSENT, and it now means present-but-null only for a non-nullable link, whose subrecord is always on disk. A reader who lands on this constant first is told the opposite of what the new one says two lines down. Rewrite the summary to state the post-fix split - `(null link)` = a non-nullable link carrying FormID zero, or a nullable link whose subrecord is absent; `PresentNullLinkNote` = a nullable link whose subrecord is present with zero - and drop the "the conflict diff treats both as no value here" clause's implication that it covers the present-but-null case generally, since `IsAbsentSentinel` no longer does.

src/housecarl-core/ReadEngine.cs
```diff
@@ -1036,7 +1044,14 @@ internal static LeafRead EmitToken(object? val, Type declaredType, object parent
...
         // FormKey, never "Null"), so surface it as no-value — consistent with what Coerce accepts.
         if (val is IFormLinkGetter fl)
-            return fl.FormKey.IsNull ? LeafRead.None(NullLinkNote) : LeafRead.Value(FormIdToken.Of(fl.FormKey));
+        {
+            if (!fl.FormKey.IsNull) return LeafRead.Value(FormIdToken.Of(fl.FormKey));
+            // Null FormKey: on a NULLABLE link that is two different facts. FormKeyNullable is null only when the
+            // subrecord is ABSENT; a subrecord PRESENT with FormID zero keeps a (null) FormKey there. They mean
+            // opposite things (an INFO's PNAM: append to the tail vs pin to the head), so they render apart (#697).
+            bool nullable = WriteEngine.ClosedInterface(val.GetType(), typeof(IFormLinkNullableGetter<>)) is not null;
+            return LeafRead.None(nullable && fl.FormKeyNullable is not null ? PresentNullLinkNote : NullLinkNote);
```
Avick3110, line 1057: The nullability test is correct but takes the reflective route where a type test does the job: Mutagen 0.54.4 ships a non-generic `IFormLinkNullableGetter` (I checked the shipped `Mutagen.Bethesda.Core.dll` metadata), so `bool nullable = val is IFormLinkNullableGetter;` is the same answer with no `GetInterfaces()` array allocation and no dependency on the link's runtime type closing the generic form - and it also covers any implementation that carries only the non-generic interface, which the `ClosedInterface` lookup would silently miss and render as `(null link)`. This leaf runs per field per record on the bulk-read path, so the allocation is not free even though it only fires on null links. Minor, but worth taking.

src/housecarl-core/FieldsDiff.cs
```diff
@@ -63,7 +63,10 @@ public sealed record Result(IReadOnlyList<string> Deltas, bool Complete,
     /// <summary>True when a CleanLines value is a read-engine "no value here" note sentinel — the field is
     /// modeled but the contributor carries nothing (absent optional, or a present-but-null link). Treated as a
     /// first-class state, never compared as if it were a real token value. References the <see cref="ReadEngine"/>
-    /// constants directly (same assembly) — single source of truth, compile-time coupling, no drift.</summary>
+    /// constants directly (same assembly) — single source of truth, compile-time coupling, no drift.
+    /// <para><see cref="ReadEngine.PresentNullLinkNote"/> is deliberately NOT here: a nullable link whose subrecord
+    /// is present with FormID zero is a carried fact (an INFO's "I am first" PNAM), so it must delta against a side
+    /// that carries nothing rather than collapse into the same "no value here" state.</para></summary>
```
Avick3110, line 69: This paragraph settles the diff's classification of the new note but leaves the other classifier disagreeing with it, unremarked: `FieldPredicate.ClassifyPresence` (`FieldPredicate.cs:1195-1202`) falls through every note that is not `(no field` / `(unreadable` / a `[` container summary to `Presence.Absent`, so a present-zero PNAM is still Absent to `exists:`/`missing:` even though the diff now treats it as a carried fact. The result is that `missing:PreviousDialog` matches an INFO that carries the head marker and `exists:PreviousDialog` misses it - the same collapse this PR fixes on the read axis, still live on the query axis, which is the surface a modder would use to sweep a topic for lines that must be left alone. Either classify `PresentNullLinkNote` as Present there (my read of the intent: the subrecord IS present, and it is what `exists` asks about) or, if Absent is deliberate because there is no value, say so where the decision is made; at minimum that method's summary and the `// (absent) / (null link) / (unresolved localized string)` comment are now an incomplete enumeration that hides a third note.

src/housecarl-core/FieldsDiff.cs
```diff
@@ -63,7 +63,10 @@ public sealed record Result(IReadOnlyList<string> Deltas, bool Complete,
...
     static bool IsAbsentSentinel(string val) =>
         val == ReadEngine.AbsentNote || val == ReadEngine.NullLinkNote || val == ReadEngine.UnresolvedStringNote;
```
Avick3110, line 71: Narrowing `IsAbsentSentinel` falsifies a comment in the probe that pins this behaviour: `src/housecarl-generator/ConflictDiffProbe.cs:135-139` (arm I) states that "the symmetric `(null link)` sentinel (a present-but-FormKey.Null link) is handled identically by construction (IsAbsentSentinel covers both)", which is now exactly the case the `<para>` above says it deliberately does NOT cover. The probe still passes because its fixture is Mutagen-authored and so carries the absent shape, which is the reason a stale claim here is dangerous: it reads as a measured guarantee about the present-but-null case that nothing measures any more. Update that comment in the same commit to say that a nullable present-zero link is deliberately not a sentinel and where the decision lives.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A record read tells a present-but-zero link apart from an absent one.** A nullable link whose subrecord is
+  present carrying FormID zero now reads `(null link, subrecord present)`; an absent subrecord still reads
+  `(null link)`. The two mean opposite things on an INFO's `PreviousDialog` (PNAM): present-zero is the "I am
```
Avick3110: This entry restates the PNAM semantics rather than pointing at them, which is the defect the file's own writing rule names at the top ("point at the bound rather than restating it ... an unpointed restatement is the defect, because it is a second copy of a fact that will be edited once"). "Present-zero is the I-am-first marker that pins the line to the head of its topic, absent appends it to the tail" is a verbatim third copy of what `docs/dialogue.md` says and what the entry two items down already paraphrases, and this PR edits that very page - so the copies will drift on the next dialogue change. Cut the semantics to a clause and point: state what the read now renders, then "what the two shapes mean and how to re-list a line without moving it are in `docs/dialogue.md`", the way the `DialogBranch.Flags` entry below already does. The scoping to nullable links, and the two `project=` recipes, are fine as they stand.

src/housecarl-mcp-tests/PresentNullLinkRenderTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    [Fact]
+    public void APresentZeroPnamRendersApartFromAnAbsentOne()
+    {
+        var marked = ReadPnam(_w.MarkedLine);
+        var plain = ReadPnam(_w.PlainLine);
+
+        Assert.Contains("(null link, subrecord present)", marked);
+        Assert.DoesNotContain("(null link, subrecord present)", plain);
+        Assert.Contains("(null link)", plain);
+    }
```
Avick3110, line 122: The test pins the render but not the decision this PR argues hardest for: that `FieldsDiff` deliberately keeps the new note out of `IsAbsentSentinel`, so a carried head marker deltas against a side that carries nothing. Right now that lives only in a doc comment, and the next person who reads `IsAbsentSentinel` as an obvious omission and adds `PresentNullLinkNote` to it gets a green suite - the collapse is back on the conflict tree with nothing failing. The world already has both shapes on one topic; a second `[Fact]` comparing the marked INFO against the plain one through the same diff path the conflict tree uses (asserting the delta line naming the marker rather than an ABSENT/ABSENT no-delta) would cost a few lines and pin the branch that has no coverage. Worth adding while the fixture is in hand.

Avick3110 (commented):
Blind review of #705. I read the diff against `main` read-only (no checkout, no build) plus the consumers of the note constants: `FieldsDiff`, `FieldPredicate.ClassifyPresence`, `RowProjection`/`FieldFold`, `ConflictDiffProbe`, `DialogueInfoOrder.LineOf`, `DialogueCkParity`. The core change looks right: the `IFormLinkNullableGetter<>` guard is necessary (a non-nullable `FormLink<T>` reports `FormKeyNullable` non-null even when the FormKey is null, so without it every non-nullable null link would flip to the new note), `FormKeyNullable` is the same test `LineOf` already relies on, the byte-patched fixture targets a 4-byte PNAM whose payload equals the plain INFO's raw FormID and asserts a patch count of exactly 1 (the DIAL's own 4-byte PNAM priority cannot collide with it, and only the marked INFO was given a PNAM at all), keeping the new note out of `IsAbsentSentinel` is the right call and is coherent downstream (a note is never a value leaf, so `AgreedCount` is untouched, and the wAbsent arm renders the head marker against a carrying-nothing side), and every touched blob is CRLF throughout. Six findings below, the last two minor.

src/housecarl-core/ReadEngine.cs
Avick3110: Not taking this one: Mutagen 0.54.4 ships no non-generic `IFormLinkNullableGetter`, so `val is IFormLinkNullableGetter` does not compile. Reflected over the shipped `Mutagen.Bethesda.Core.dll` (0.54.4, net9.0): the only exported names matching `FormLinkNullable` are `FormLinkNullableGetter`1`, `FormLinkNullable`1`, `IFormLinkNullableGetter`1` and `IFormLinkNullable`1` — all generic. The interface list of `FormLinkNullable<IMajorRecordGetter>` carries the closed `IFormLinkNullableGetter`1` and the non-generic `IFormLinkGetter`, `ILink`, `ILinkIdentifier`, `IFormLinkIdentifier`, but no non-generic nullable form. The `ClosedInterface` lookup stays.

One thing the same reflection turned up that is worth recording: the nullability guard is load-bearing and not just a narrowing. `((IFormLinkGetter)new FormLink<IMajorRecordGetter>(FormKey.Null)).FormKeyNullable` returns `FormKey.Null`, **not** `null` — so on a non-nullable link the `FormKeyNullable is not null` half is true, and dropping the `nullable &&` would render every zero non-nullable link as `(null link, subrecord present)`.

PR #706, opened 2026-09-09
Say when a copy source sits in a disabled mod folder

### LINKED ISSUE #703: housecarl_copy does not say when the named source mod is disabled in MO2
Found running EL-1 on 2026-09-09. housecarl_copy with from_source=[EternalWomen.esp], a plugin in a mod folder MO2 is not loading, reported "source: EternalWomen.esp (MO2 mod folder ...)" and "standalone: the source is NOT a master", but never said the folder is disabled. housecarl_place and nif_inspect do say it ("a mod folder that is NOT enabled in MO2 — you named it, so houseCARL read it off disk"). The copy report should carry the same sentence so the standalone claim reads with the fact that the game is not loading the source. One sentence, no behaviour change.

Avick3110:
`housecarl_copy` reads a `from_source=` plugin straight off disk when it sits in a mod folder MO2 has switched off, but the report only named the folder and then claimed the patch is standalone — the reader was never told the game is not loading the source at all, so the standalone claim had nothing to be read against. The MO2 layer a source arm resolved from now carries whether that folder is switched on, set from the locate's own `ModDisabled` standing, and the copy readback says so in one line beside the provenance. The sentence is the one `housecarl_place` and `housecarl_nif_inspect` already print, pulled out of the place helper into a shared constant so the three cannot drift; the place text is unchanged byte for byte. Nothing about what is copied, internalized or mastered changes. Covered by a test on the existing two-disabled-donors world (it fails without the render change) plus a negative one on the enabled-donor world, and by the sentence-reach net in `write-surface-guard`.

Closes #703

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -28,6 +28,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `project={"form":"fields", "fields":["PreviousDialog"]}` on the INFO says which shape it carries, and
   `project={"form":"info_order"}` on its topic says where the line lands.
 
+- **`housecarl_copy` says when the source it read from sits in a mod folder MO2 has switched off.** A
+  `from_source=` plugin found in a disabled mod folder is reported with the same sentence
+  `housecarl_place` and `housecarl_nif_inspect` already use, beside the line naming the folder. Nothing
+  about what is copied changes: the standalone claim under it is about what the patch masters, and it now
+  reads next to the fact that the game is not loading the source.
 - **`housecarl_create` refuses a dialogue branch with no `Flags` instead of filling one.** A `DialogBranch`
   created with no `Flags` is now refused in one sentence naming both values and what each does — `TopLevel`
   for a menu entry the player can pick, `0` for a scripted `Say()` topic that must stay hidden — and nothing
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -27,12 +27,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `docs/dialogue.md`. Check a line before you re-list it —
   `project={"form":"fields", "fields":["PreviousDialog"]}` on the INFO says which shape it carries, and
   `project={"form":"info_order"}` on its topic says where the line lands.
-
-- **`housecarl_copy` says when the source it read from sits in a mod folder MO2 has switched off.** A
-  `from_source=` plugin found in a disabled mod folder is reported with the same sentence
-  `housecarl_place` and `housecarl_nif_inspect` already use, beside the line naming the folder. Nothing
-  about what is copied changes: the standalone claim under it is about what the patch masters, and it now
-  reads next to the fact that the game is not loading the source.
+- **`housecarl_copy` says which of the source folders it names the game is not loading.** Every arm of
+  `from_source=` that resolved into a mod folder the active profile does not load gets its own `note:` line
+  beside the sources it consulted — not just the arm the record was read from, since the folder a placement
+  is handed afterwards is often one of the others. A folder MO2 lists and has switched off is reported with
+  the sentence `housecarl_place` already uses; a folder MO2 has not registered gets its own, because there
+  is nothing in MO2's list to switch on for it. Nothing about what is copied changes: the standalone claim
+  under it is about what the patch masters, and it now reads next to the fact that the game is not loading
+  the source.
 - **`housecarl_create` refuses a dialogue branch with no `Flags` instead of filling one.** A `DialogBranch`
   created with no `Flags` is now refused in one sentence naming both values and what each does — `TopLevel`
   for a menu entry the player can pick, `0` for a scripted `Say()` topic that must stay hidden — and nothing
--- a/src/housecarl-core/SourceChain.cs
+++ b/src/housecarl-core/SourceChain.cs
@@ -61,13 +61,29 @@ public enum SourceLayerKind
     GameData,
 }
 
+/// <summary>Whether the active MO2 profile is loading a mod folder — the two ways it is not, kept apart because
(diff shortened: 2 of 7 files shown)


src/housecarl-mcp/CopyTools.cs
```diff
@@ -149,7 +149,12 @@ internal static string Render(ClosureCopyOutcome o)
         // Which source produced the record the caller ASKED for. Needed in both modes: the source record is not
         // among the internalized rows in either, so nothing else names where its own body came from.
         if (o.FromArm is { } fromArm)
+        {
             sb.Append(WriteSentences.CopyFromArmLead).Append(WriteSentences.CopyArm(fromArm)).Append(".\n");
+            // The source folder is switched off in MO2, which the standalone claim below is read against.
+            if (fromArm.Layer is { Kind: SourceLayerKind.ModFolder, OwnerEnabled: false } off)
```
Avick3110: The note fires only for `o.FromArm`, so in a multi-source call every OTHER disabled arm is still printed as a live folder. In the very shape the surrounding tests are built on (`TwoDisabledDonorsWorld`, both donors switched off), `from_source=["Override.esp","Donor.esp"]` gets the note for `DisabledA` and nothing for `Bijin's NPCs (SE)` — and that second folder is the one the caller actually acts on next, because `CopyAssetPathsHeader` sends them to `housecarl_place` with `source_provider=` naming the folder that ships the FaceGen, which in this shape is the arm the record did NOT come from. So the one folder whose disabled state the caller most needs is the one that stays unqualified. Either say it for every arm whose layer is a switched-off mod folder — the sources-consulted line is where a per-arm qualifier belongs — or, if from-arm-only is the deliberate scope, add a test pinning it: with both donors disabled the current test would pass equally under a note-every-arm implementation, so nothing holds the rule the probe comment documents.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **`housecarl_copy` says when the source it read from sits in a mod folder MO2 has switched off.** A
+  `from_source=` plugin found in a disabled mod folder is reported with the same sentence
+  `housecarl_place` and `housecarl_nif_inspect` already use, beside the line naming the folder. Nothing
+  about what is copied changes: the standalone claim under it is about what the patch masters, and it now
+  reads next to the fact that the game is not loading the source.
```
Avick3110: The entry says "A `from_source=` plugin found in a disabled mod folder is reported with the same sentence", which reads as any element of the list. The code says it only for the arm the record was read from (`o.FromArm` in `CopyTools.Render`), so a caller with two poles whose second is disabled will not see the line the entry promises. Per the pointing-at-bounds rule at the top of this file, this needs to name its bound rather than state the broader claim — e.g. "the source the record was read from", pointing at the provenance line the note sits beside. (If the behaviour widens instead, per the other finding, the entry is then correct as written.)

src/housecarl-core/SourceChain.cs
```diff
@@ -61,8 +61,13 @@ public enum SourceLayerKind
     GameData,
 }
 
-/// <summary>The layer a source's file sits in: which branch answered, and the name that branch produced.</summary>
-public sealed record SourceLayer(SourceLayerKind Kind, string Name);
+/// <summary>The layer a source's file sits in: which branch answered, the name that branch produced, and — for a
+/// mod folder — whether MO2 has that folder switched ON.
+/// <para><paramref name="OwnerEnabled"/> defaults to true because every other layer is always live: the game's Data
```
Avick3110: The rationale for the default is not accurate, and the inaccuracy hides a real hole. "Every other layer is always live" covers Overwrite and GameData, but the default also lands on a ModFolder layer whenever the producer does not override it — which includes `ServedStanding.ModUnregisteredLayer`, a folder modlist.txt does not mention at all. The setter in `LoadOrderService.BuildSourceChain` tests `== ServedStanding.ModDisabled` only, so a source read out of an unregistered folder travels as `OwnerEnabled: true` and the readback names it as though the game were loading it, which is the same gap #703 is about. Pick one: either narrow the field's stated meaning to "listed in modlist.txt and switched off" and say that the unregistered standing is deliberately silent here and why, or carry that standing too — noting it cannot borrow the shared `OffOrderModFolder` wording, since "NOT enabled in MO2" is false for a folder MO2 has no entry to enable, which is exactly why `WhyNotActive` keeps the two remedies apart.

Avick3110 (commented):
Blind review of #706. Three findings below. Checked and found clean: the place/nif text is byte-identical after the constant extraction (old `"...read it off " + "disk; the bytes..."` reconcatenates to exactly `OffOrderModFolder` + `"; the bytes..."`), and all three callers of `PlaceSourceOffOrder` (PlaceTools, JsonWire, the nif lane’s `SourceProvenance`) go through the one function, so nothing can drift; `OwnerEnabled` as a bool is the right carrier rather than reusing `ServedStanding`, which is an mcp-internal enum that housecarl-core cannot see, and the name matches the existing `PlacementSource.OwnerEnabled`; every `SourceLayer` construction site was checked — the only non-probe producer is `InstallLayerOfPath`, whose other caller `ActiveLayer` runs on plugins in the active order, whose folder is served and therefore on, so the default-true is not silently wrong there; the positive test genuinely fails without the render change (nothing else emits that exact string — `WhyNotActive` words it as “switched OFF in MO2”); the `WriteSurfaceGuardProbe` entry is correct, since `Make` sets `FromArm` to `srcs[0]` and the new single-arm array is the disabled one, so the render really does reach the note; copy has no JSON wire shape, so there is no second surface left silent; and every touched file is LF-only in the object store on both main and this branch, so nothing changed line endings (verified by counting 0x0d bytes out of `git cat-file blob`, not by grep).

PR #707, opened 2026-09-10
Replace the README with the 2.0 surface

Avick3110:
The top-level README still described the 1.x surface, down to the `houseCARL-1.9.0.zip` filename, so a
reader arriving at the repo got a tool list that no longer exists; this replaces it with the approved 2.0.0
text, which states what houseCARL is and is not, the four design rules, the 31 tools grouped by substrate,
the read and write axes with one worked call each, the runtime layers, the writing lanes, the seven skills,
requirements, install for both hosts, the from-source build, development paths and commands, and the
upgrade-from-1.x note. The draft was checked against the repo before landing and needed no corrections: all
ten link targets exist (`docs/facegen.md`, `docs/dialogue.md`, `CLAUDE.md`, `CONTRIBUTING.md`, `LICENSE`,
`THIRD-PARTY-NOTICES.txt`, `plugin/CHANGELOG.md`, `docs/architecture/`, `docs/decisions/`,
`scripts/build-plugin.ps1`); the 31 names in the table are exactly the 31 in `src/housecarl-core/ToolNames.cs`
with the prefix stripped; every parameter in the two example calls and the axis tables matches
`src/housecarl-mcp/RecordsTools.cs` and `src/housecarl-mcp/ApplyTools.cs` (`plugins.names`, `defined_in`,
`where`, `where_source`, `fields_source`, `source`, `versus`, `project.form`/`fields`, `format`, `limit`,
`offset`, `max_chars`, `counts_only`, `to_file`; `ops[].formid`/`field_path`/`op`/`value`, `patch`, `into`,
`in_place`, `acknowledge`, `dry_run`, `readback`), and the eight write verbs are the eight in
`src/housecarl-core/WriteVerbs.cs` in the same order; the four `check` families are the four in
`src/housecarl-mcp/CheckTools.cs`; the seven skill names are the seven folders under `.claude/skills/`; the
numbers hold — the `mutagen-reference` index carries 1,174 entries made of 133 records, 242 structs, 497
arms, 280 enums (plus 21 polymorphic bases and a header), the csproj files pin Mutagen.Bethesda.Skyrim
0.54.4, and the papyrus-reference index is 7,345 lines; the build and test commands match `CLAUDE.md`
verbatim; and the Claude Code and Codex install paths match `src/housecarl-setup/Program.cs`.
`plugin/README.md` was left alone: it is a separate, shorter marketplace-facing document that carries no 1.x
tool names and no `1.9.0` filename. `dotnet build housecarl.sln -c Release` succeeds, so nothing reads the
README at build time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,9 +92,9 @@ housecarl_apply(
 
 **FormIDs.** A record is addressed as `XXXXXX:Plugin.esp`, or in the runtime form the console, Papyrus log and crash log print (`FExxxYYY`, `XX######`), resolved against the current order. Reads accept both forms. Writes accept the plugin form only; a runtime form is refused with the plugin form to use.
 
-**Large results.** Every response carries an `epoch` stamp identifying the index build it was answered from. A result over the render limit is written in full to a JSONL file whose first line is a manifest, and the response names the file. The file re-enters a later call as `formids=["@<path>"]`, epoch-checked.
+**Large results.** Every response on the record plane carries an `epoch` stamp identifying the index build it was answered from. A result over the render limit is written in full to a JSONL file whose first line is a manifest, and the response names the file. The file re-enters a later call as `formids=["@<path>"]`, epoch-checked.
 
-**Findings.** `check` runs derived-findings families over the order: `errors` (dangling FormLinks, missing masters, unparseable records), `scripts` (script properties the VMAD does not bind), `dialogue` (a topic's graph as the game resolves it), `facegen` (per NPC: the mod winning the head `.nif`, the mod winning the face `.dds`, the plugin winning the record, the mismatch class). Each family's description states what it does not cover.
+**Findings.** `check` runs derived-findings families over the order: `errors` (dangling FormLinks, missing masters, unparseable records), `scripts` (script properties the VMAD does not bind), `dialogue` (dialogue graph validation over seeded topics and quests — this family takes seeds, it does not sweep), `facegen` (per NPC: the mod winning the head `.nif`, the mod winning the face `.dds`, the plugin winning the record, the mismatch class). Each family's description states what it does not cover.
 
 ## Coverage
 
@@ -141,7 +141,7 @@ Dialogue records have bookkeeping that a byte-valid insert does not satisfy: [do
 
 ## Skills
 
-Seven skills ship with the plugin: `/housecarl:<name>` in Claude Code, `$housecarl` in Codex. Each is a `SKILL.md` and a `references/` tree. Reference corpora are indexed by `index.jsonl` (name to file and line range) and read by grep, never loaded whole.
+Seven skills ship with the plugin: `/housecarl:<name>` in Claude Code, `$housecarl` in Codex. Each is a `SKILL.md` and a `references/` tree. Reference corpora are read by grep, never loaded whole. Four of the seven — `mutagen-reference`, `papyrus-reference`, `spid-authoring`, `kid-authoring` — carry a `references/index.jsonl` mapping a name to the file that holds it; the two generated corpora carry the line as well.
 
 | Skill | Carries |
 |---|---|
@@ -212,7 +212,7 @@ One process, MCP over stdio. The load order is read from the MO2 profile's `load
 
 Freshness is a last-write-plus-size check on every call. A rebuild is an immutable snapshot swapped in as one reference. A plugin the build cannot read is excluded whole with the reason, and every response from that build says which plugins are missing.
 
-Writes are verified by an oracle harness: each mutation is performed through the reflection engine and through a hand-written typed Mutagen setter, and the two output plugins must be byte-identical. Around 150 further probe harnesses run in CI against real plugins.
+Writes are verified by a per-kind oracle: 17 cells, each performing one mutation two ways — through the reflection engine, and through a hand-written typed Mutagen setter — with the two output plugins required to be byte-identical. The engine is blind to which record it edits, so proving a kind proves every record carrying that kind. The oracle is run by hand against a real `Skyrim.esm`. 134 probe harnesses run in CI against real plugins.
 
 The only outbound network use is the Nexus lookups. They need no account or API key. Offline, they say so, and every local tool keeps working. `update_status` reads MO2's local cache first and checks at the exact-file level.
 
@@ -242,7 +242,7 @@ Read [CLAUDE.md](CLAUDE.md) before changing anything, then the note in `docs/arc
 
 ## Upgrading from 1.x
 
-2.0.0 replaces the 1.x tool surface with 31 tools built from one grammar. There is no alias layer and no deprecation window. A 1.x tool name is refused with one sentence naming its successor. The old-to-new table is in the [changelog](plugin/CHANGELOG.md).
+2.0.0 replaces the 1.x tool surface with 31 tools built from one grammar. There is no alias layer and no deprecation window. A 1.x tool name is refused with one sentence naming its successor and the call shape to use. The rows those sentences come from are in `src/housecarl-mcp/AliasTable.cs`.
 
 ## Licence
 
--- a/README.md
+++ b/README.md
@@ -1,133 +1,183 @@
 # houseCARL
 
-**Comprehensive, data-layer access to your Skyrim Special Edition load order — in plain English, through Claude or Codex.**
+houseCARL is an MCP server that exposes a Skyrim Special Edition load order at the data layer. It reads a Mod Organizer 2 instance's profile and presents every plugin record, every Data-relative asset, the compiled Papyrus, the SKSE plugin layer and the SkyPatcher INI layer to an AI assistant as a set of tools. It is built on [Mutagen](https://github.com/Mutagen-Modding/Mutagen). The host is [Claude Code](https://claude.com/claude-code) or OpenAI Codex.
 
-houseCARL runs a local MCP server with [Mutagen](https://github.com/Mutagen-Modding/Mutagen) — the
-Bethesda-format library — kept warm in memory, giving your AI assistant direct access to every plugin
-record across your Mod Organizer 2 load order, at the **true load-order winner**, with the full conflict
-tree on request. You describe what you want in plain English; houseCARL does the mechanical work and, by
-default, writes results into a **new** plugin you review and enable in MO2 — your originals untouched. When
-you ask for it, an opt-in **in-place lane** edits an existing plugin directly instead (see below).
+A write produces a new plugin in a new MO2 mod folder. Editing an existing plugin in place is an opt-in that names the file. MO2 does not need to be running. No plugin file handle is held between calls.
 
-Coverage is **reflection-driven**: a build-time generator walks Mutagen's record interfaces and emits
-houseCARL's schema automatically, so the set of record types houseCARL understands *is* the set Mutagen
-models — by construction, not a hand-maintained subset.
+| | |
+|---|---|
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,206 +1,256 @@
...
+
+**FormIDs.** A record is addressed as `XXXXXX:Plugin.esp`, or in the runtime form the console, Papyrus log and crash log print (`FExxxYYY`, `XX######`), resolved against the current order. Reads accept both forms. Writes accept the plugin form only; a runtime form is refused with the plugin form to use.
+
+**Large results.** Every response carries an `epoch` stamp identifying the index build it was answered from. A result over the render limit is written in full to a JSONL file whose first line is a manifest, and the response names the file. The file re-enters a later call as `formids=["@<path>"]`, epoch-checked.
```
Avick3110: "Every response carries an `epoch` stamp" is true of the record plane and not of the surface as a whole, and this paragraph sits under a heading that covers all 31 tools. `BsaTools.cs`, `NexusTools.cs`, `AssetTools.cs`, `NifTools.cs`, `SkseTools.cs` and `SkyPatcherTools.cs` contain the string `epoch` zero times, so six of the tools listed in the table above return responses with no such stamp. A reader who takes the sentence literally will look for an epoch on a `bsa_list` or `nexus_mod` answer and not find one, and the whole point of the stamp is that its absence is meaningful. Scope the sentence to what carries it — "Every response on the record plane carries an `epoch` stamp" — or move the sentence into the `records` discussion above it.

README.md
```diff
@@ -1,206 +1,256 @@
...
+
+**Findings.** `check` runs derived-findings families over the order: `errors` (dangling FormLinks, missing masters, unparseable records), `scripts` (script properties the VMAD does not bind), `dialogue` (a topic's graph as the game resolves it), `facegen` (per NPC: the mod winning the head `.nif`, the mod winning the face `.dds`, the plugin winning the record, the mismatch class). Each family's description states what it does not cover.
```
Avick3110: The gloss for the `dialogue` family, "a topic's graph as the game resolves it", describes the one thing the family explicitly does not do. `CheckTools.cs` says the family is "dialogue graph validation over SEEDED topics and quests", and then, in the same description, "NOT HERE: the effective merged INFO order — the sequence the game walks, which line MOVED and which plugin moved it ... lives on housecarl_records project='info_order'". The README already lists `info_order` as a projection three lines above, so as written the two paragraphs give the same capability two homes and point a reader at the wrong call. Say what the family does instead — it validates a seeded topic's or quest's dialogue graph — and, since the sentence after this one promises that each family states its bounds, it is worth adding that this family is seeded rather than swept, which is the fact that makes it unlike the other three.

README.md
```diff
@@ -1,206 +1,256 @@
...
+
+**Readback.** Each edited field is read back from the written file. `readback=true` deep-reads every touched record. The readback reports the file's content. The patch has no effect until it is enabled and sorted in MO2.
+
+**Beyond field edits.** Create records with fresh FormIDs, including nested dialogue structures. Remove records and list entries. Forward a named plugin's version of a record as a winning override, or revert a record to vanilla. Copy an NPC's appearance closure into a standalone record with no dependency on the donor. Create an empty plugin. ESL-compact a plugin, carrying its FormID-keyed FaceGen and voice files. Merge plugins with collision-only renumbering. Unused masters are trimmed on every write.
+
+Dialogue records have bookkeeping that a byte-valid insert does not satisfy: [docs/dialogue.md](docs/dialogue.md).
+
+## Skills
+
+Seven skills ship with the plugin: `/housecarl:<name>` in Claude Code, `$housecarl` in Codex. Each is a `SKILL.md` and a `references/` tree. Reference corpora are indexed by `index.jsonl` (name to file and line range) and read by grep, never loaded whole.
```
Avick3110: "Reference corpora are indexed by `index.jsonl` (name to file and line range)" is stated of all seven skills and holds for four. `open-animation-replacer`, `skse-plugin-authoring` and `skypatcher-authoring` carry no `references/index.jsonl` at all — their references trees are plain `.md` files a reader greps directly. The parenthetical is also wrong for one of the four that do have an index: `mutagen-reference/references/index.jsonl` rows carry a single `"line"` member, not a range (`{"name":"AIData","kind":"struct","file":"references/structs.jsonl","line":1}`); only `papyrus-reference` has `line_start`/`line_end`. Either scope the sentence to the corpora that are indexed, or drop the parenthetical and keep the part that is true of all seven, which is that references are read by grep and never loaded whole.

README.md
```diff
@@ -1,206 +1,256 @@
...
-pointers.
+One process, MCP over stdio. The load order is read from the MO2 profile's `loadorder.txt`, `modlist.txt` and `plugins.txt`. No USVFS, no hook into MO2. Records parse on access from Mutagen's lazy binary overlay. A structural index (FormKey to winning plugin and override count, plus the touching-plugin list for contested records) is built by opening each plugin once and disposing it; measured at 125–185 MB on a 3,400-plugin order. A call that needs record bodies opens each plugin at most once for the call. No plugin handle is held between calls, so MO2, xEdit and Explorer operate on the plugins normally while houseCARL runs.
+
+Freshness is a last-write-plus-size check on every call. A rebuild is an immutable snapshot swapped in as one reference. A plugin the build cannot read is excluded whole with the reason, and every response from that build says which plugins are missing.
+
+Writes are verified by an oracle harness: each mutation is performed through the reflection engine and through a hand-written typed Mutagen setter, and the two output plugins must be byte-identical. Around 150 further probe harnesses run in CI against real plugins.
```
Avick3110: The oracle sentence over-states both what the harness covers and where it runs. `WriteOracle.cs` describes itself as "per-kind byte-identical cells" — it performs a fixed set of cells (`WriteProof.cs` records it as 17/17) each two ways and relies on the engine being blind to record type, "which is how full coverage is validated without hand-authoring ~9,000 per-field oracles". "Each mutation is performed ... two ways" reads as every write houseCARL makes being double-checked, which is not what happens and drops the per-kind argument that is the actual reason the proof carries. It is also not verified by CI: `oracle` is dispatched by hand from `Program.cs` line 70, carries no `[CiProbe]` attribute, does not appear in `.github/workflows`, and needs a real `Skyrim.esm` on disk — so the word "further" in the next sentence puts it in a CI run it is not in. State it as the per-kind proof it is, and say it is run by hand against a real master.

Avick3110: "Around 150 further probe harnesses run in CI" is above the real number. The CI roster is derived from the `[CiProbe]` attribute across the four houseCARL assemblies (`CiAll.cs`: "every method carrying CiProbeAttribute"), and that attribute appears 134 times — 132 of them in `src/housecarl-generator`. 150 is the count of files matching `*Probe*.cs` in the generator, which is not the same set: `WriteDiagnosticsProbe.cs` says of its own contents "None of these is a CI probe", and `Program.cs` labels several others manual or exploratory harnesses that need a live MO2 instance. Say 134, or "over 130", so the number cannot read as an inflation of a figure the repo already carries exactly.

README.md
```diff
@@ -1,206 +1,256 @@
...
+dotnet build housecarl.sln -c Release
+dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"
+dotnet src/housecarl-generator/bin/Release/net9.0/housecarl-generator.dll ci-all
+```
+
+Read [CLAUDE.md](CLAUDE.md) before changing anything, then the note in `docs/architecture/` for the subsystem. Decisions are recorded one per file in `docs/decisions/`. Contribution process: [CONTRIBUTING.md](CONTRIBUTING.md).
+
+## Upgrading from 1.x
+
+2.0.0 replaces the 1.x tool surface with 31 tools built from one grammar. There is no alias layer and no deprecation window. A 1.x tool name is refused with one sentence naming its successor. The old-to-new table is in the [changelog](plugin/CHANGELOG.md).
```
Avick3110: "The old-to-new table is in the [changelog]" points at a table that does not exist. `plugin/CHANGELOG.md` contains zero markdown table rows (no line begins with `|`), and the 1.x-to-2.0 mappings are spread across dozens of prose entries at different versions — a reader who follows this link to find, say, what replaced `housecarl_read_record` has to read the file. The mapping does exist in one place in the repo, `src/housecarl-mcp/AliasTable.cs`, which is where the refusals in the sentence before this one come from. Either point at the changelog without calling it a table, or write the table into the changelog's 2.0.0 entry and keep the link as it is.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The top-level `README.md` describes the 2.0 surface.** It names the 31 tools and the axes a call is
+  composed from, states what houseCARL is not, and carries the install, requirements and upgrade-from-1.x
+  notes. The old-to-new tool-name table stays in this file.
```
Avick3110: "The old-to-new tool-name table stays in this file" is the entry's one pointing-at-a-bound clause, and it points at a table this file does not contain: `plugin/CHANGELOG.md` has no markdown table anywhere in it (no line begins with `|`), and the 1.x-to-2.0 mappings live scattered through prose entries, with the one collected form in `src/housecarl-mcp/AliasTable.cs`. The clause is doing exactly the job the rule at the top of this file asks of it — pointing rather than restating — so it needs to point at something real: either name `AliasTable.cs` and the refusal it produces, or add the table to the 2.0.0 entry so the sentence becomes true. While the line is being edited, "names the 31 tools" is a third copy of a count whose home is `ToolNames.cs`; "names the tool surface" carries the same meaning and cannot drift.

Avick3110 (commented):
Blind review of the 2.0 README rewrite. I checked every tool name against `src/housecarl-core/ToolNames.cs` (31, exact), the read and write parameter spellings against `RecordsTools.cs` and `ApplyTools.cs` (`plugins.names`, `defined_in`, `conflicts_only`, `where`, `where_source`, `references`, `walk`, `source`, `versus`, `fields_source`, `project.form`/`fields`, `format`, `limit`, `offset`, `max_chars`, `counts_only`, `to_file`; `ops[].formid`/`field_path`/`op`/`value`, `patch`, `into`, `in_place`, `acknowledge`, `dry_run`, `readback` — all present and spelled as written), the eight write verbs against `WriteVerbs.All` (same eight, same order), the four check families against `CheckTools.cs`, the seven skill names against `.claude/skills/`, the install paths and the `[1]/[2]/[3]` menu against `src/housecarl-setup/Program.cs`, the generated index counts (index.jsonl is 1,174 lines: 133 record, 242 struct, 497 arm, 280 enum, 21 polymorphic-base, 1 header) and the papyrus index (7,345 lines), the Mutagen 0.54.4 pin in the three csproj files, the runtime-FormID write refusal in `FormIdDoor.ForWrite`, the `houseCARL - <name>` folder shape, the patch-name suffix and refusal rules, and the ASP.NET-Core-plus-base-runtime check. The unverifiable-looking figures are all corroborated in the tree: the v2.1.143 floor in `plugin/README.md`, 125-185 MB in `LoadOrderResolver.cs`, 98.80% of 10,189 in `DecompileGuardProbe.cs`, and the four grammar versions in the skills' own SKILL.md frontmatter. All ten link targets exist. None of the 23 retired names in `AliasTable.cs` survives anywhere in the new text. README.md's blob is LF-only, like the rest of the repo (0 CR bytes, 256 LF). The register holds throughout — I found no hedge, no sales verb, no "worth stating", no "buys", no "not X, it is Y". Seven findings follow, all factual.

PR #708, opened 2026-09-10
Stamp the setup exe and lead its errors with a sentence

Avick3110:
The installer gets the first of three presentation passes agreed for 2.0.0. The setup exe is now stamped with the release version the way the server already was (`build-plugin.ps1` step 9 passes `-p:Version` from `plugin.json`, the csproj carries `AssemblyTitle`/`Product` and defaults to `0.0.0-dev`), and the banner reads that version back off its own assembly so there is one version home instead of a hard-coded string.

The output that carries presentation — the banner, the rule, the step/ok lines, the notes, the refusals and the menu prompt — moved out of `Program.cs` into a new `src/housecarl-setup/Ui.cs`, a small static class with `Banner`, `Rule`, `Step`, `Ok`, `Note`, `Problem` and `Prompt`, colour set through `Console.ForegroundColor` and turned off when the stream is redirected or `NO_COLOR` is set, no new package references. `Ui` also sets `Console.OutputEncoding` to UTF-8 once, in a try/catch and only when a console is attached, so the banner rule and the em dashes render on a double-clicked run. Program.cs keeps its 31 plain `Console.*` calls — the `--help` text, the host menu, the NEXT block, the blank-line spacing and the "press any key" — because none of them are coloured or shaped, and routing them through `Ui` would only add a layer.

On top of that every refusal was rewritten to lead with one plain sentence saying what went wrong and what to try, with the paths, download links and the two-installer trap indented below it; the runtime pre-flight now names only the runtime that is actually missing instead of printing both, the missing-package refusal names which piece is absent (manifest, server exe or skills folder) and tells the reader to unzip again rather than to move the program, and the locked-file refusal names the exe that is holding things up. `TryInstall` is untouched, so `SetupSkillPruneProbe` and `SetupUpdateLockProbe` stay green without edits, and `--help` is unchanged.

PR A of three for the installer polish; the detection/plan screen and uninstall follow. No issue to close.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -18,6 +18,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   notes. The 1.x-to-2.0 tool-name mapping is the refusal
   itself: a retired name is refused with the sentence naming its successor, from `AliasTable.cs` in
   `src/housecarl-mcp/`.
+- **The setup program says which version it is installing, and leads every error with one sentence.** The
+  banner now reads `houseCARL <version>  ·  setup`, stamped from the same `plugin.json` version the server
+  reports. Each refusal — the plugin folder not being beside the program, a missing .NET runtime, a server
+  file in use — now opens with one plain sentence saying what went wrong and what to try, with the paths,
+  download links and the two-installer note indented below it; the runtime refusal names only the runtime
+  that is actually missing rather than listing both. Output is coloured unless it is redirected or
+  `NO_COLOR` is set.
+
 - **A refusal for an unknown field now names every field the record type has.** `housecarl_apply` and
   `housecarl_create` run the same pre-flight, and it used to name twelve fields and then `(+N more)` with no
   route from the call to the rest — a write call is the only place the surface shows a type's schema without
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -23,9 +23,8 @@ saying it sets an expectation their install may contradict. Say what is known, a
   reports. Each refusal — the plugin folder not being beside the program, a missing .NET runtime, a server
   file in use — now opens with one plain sentence saying what went wrong and what to try, with the paths,
   download links and the two-installer note indented below it; the runtime refusal names only the runtime
-  that is actually missing rather than listing both. Output is coloured unless it is redirected or
-  `NO_COLOR` is set.
-
+  that is actually missing rather than listing both. Output is coloured; when it is not, and why, is on `Ui`
+  in `src/housecarl-setup/Ui.cs`.
 - **A refusal for an unknown field now names every field the record type has.** `housecarl_apply` and
   `housecarl_create` run the same pre-flight, and it used to name twelve fields and then `(+N more)` with no
   route from the call to the rest — a write call is the only place the surface shows a type's schema without
--- a/src/housecarl-setup/Program.cs
+++ b/src/housecarl-setup/Program.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using System.Reflection;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 
@@ -77,8 +78,9 @@ public static class Program
 
         try
         {
-            Console.WriteLine("houseCARL setup");
-            Console.WriteLine("===============");
+            Console.WriteLine();
+            Console.WriteLine("  houseCARL " + SetupVersion() + "  ·  setup");
+            Console.WriteLine("  " + new string('─', 63));
             Console.WriteLine();
 
             // Locate the plugin shipped beside this program.
@@ -510,6 +512,18 @@ public static class Program
         Console.WriteLine("     Mod Organizer 2 folder (the one containing ModOrganizer.ini).");
         if (target is Target.Both)
             Console.WriteLine("   - (Each host runs its own server copy, so you'll set the MO2 folder once per host.)");
+    }
+
+    /// <summary>This exe's own stamped version, with any "+sha" metadata trimmed. build-plugin.ps1 passes
+    /// -p:Version from plugin.json at publish, so the banner and the shipped server report the same number;
+    /// an unstamped dev build says 0.0.0-dev rather than the 1.0.0 the SDK defaults to.</summary>
+    private static string SetupVersion()
+    {
+        string? info = typeof(Program).Assembly
+            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
+        if (string.IsNullOrWhiteSpace(info)) return "0.0.0-dev";
+        int plus = info.IndexOf('+');
+        return plus > 0 ? info[..plus] : info;
     }
 
     private static int Finish(int exitCode)
--- a/src/housecarl-setup/Program.cs
+++ b/src/housecarl-setup/Program.cs
@@ -90,9 +90,10 @@ public static class Program
             string srcSkills   = Path.Combine(pluginSrc, "skills");
             if (!File.Exists(srcManifest) || !File.Exists(srcExe) || !Directory.Exists(srcSkills))
             {
-                Console.Error.WriteLine("ERROR: couldn't find the houseCARL plugin next to this program.");
-                Console.Error.WriteLine("  Looked in: " + pluginSrc);
-                Console.Error.WriteLine("  Keep this program in the same folder as the unzipped 'housecarl' folder, then run it again.");
+                Ui.Problem(
+                    "The houseCARL plugin folder is not next to this program, so there is nothing to install — "
+                    + "keep this program beside the unzipped 'housecarl' folder and run it again.",
+                    "Looked in:  " + pluginSrc);
                 return Finish(1);
             }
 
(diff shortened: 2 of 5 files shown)


src/housecarl-setup/Program.cs
```diff
@@ -91,9 +90,10 @@ private static int Main(string[] args)
             string srcSkills   = Path.Combine(pluginSrc, "skills");
             if (!File.Exists(srcManifest) || !File.Exists(srcExe) || !Directory.Exists(srcSkills))
             {
-                Console.Error.WriteLine("ERROR: couldn't find the houseCARL plugin next to this program.");
-                Console.Error.WriteLine("  Looked in: " + pluginSrc);
-                Console.Error.WriteLine("  Keep this program in the same folder as the unzipped 'housecarl' folder, then run it again.");
+                Ui.Problem(
+                    "The houseCARL plugin folder is not next to this program, so there is nothing to install — "
+                    + "keep this program beside the unzipped 'housecarl' folder and run it again.",
+                    "Looked in:  " + pluginSrc);
```
Avick3110: This sentence asserts a cause the check did not establish. The guard above fires when *any* of `srcManifest`, `srcExe` or `srcSkills` is missing, and the comment three lines up says so explicitly — a package whose `skills` folder is absent is "a broken unzip, not a version that dropped every skill". In that case the `housecarl` folder *is* next to the program, so the sentence is flatly wrong and the advice it gives ("keep this program beside the unzipped 'housecarl' folder and run it again") tells the user to do the thing they already did, which is the silently-wrong-answer failure the cornerstone forbids. The old text was vaguer but did not make the false claim. Say what was actually not found: test the three conditions separately and lead with "The houseCARL plugin folder is not next to this program…" only when `Directory.Exists(pluginSrc)` is false, otherwise lead with something like "The houseCARL plugin folder beside this program is missing its <manifest / server / skills>, so the download did not unzip completely — unzip it again and run this setup." and name the missing piece on the detail line next to `Looked in:`.

src/housecarl-setup/Program.cs
```diff
@@ -170,13 +153,58 @@ private static int Main(string[] args)
         }
         catch (Exception ex)
         {
-            Console.Error.WriteLine();
-            Console.Error.WriteLine("ERROR: houseCARL setup did not complete.");
-            Console.Error.WriteLine("  " + ex.Message);
+            Ui.Problem(
+                "houseCARL setup stopped on an error it did not expect — the line below is what Windows "
+                + "reported, and running the setup again is the first thing to try.",
```
Avick3110: Two problems in this catch sentence. "the line below is what Windows reported" attributes the message to Windows, but this catch wraps the whole run including `JsonException` from a malformed `~/.claude.json` or `config.toml`, `UnauthorizedAccessException`, and anything the JSON DOM throws — most of what lands here is .NET, not Windows, so the sentence tells the reader to look in the wrong place. And "running the setup again is the first thing to try" is advice that will not work for the deterministic failures this catch actually sees: a corrupt host config file fails identically every time. This is the one refusal where setup genuinely does not know what to try, so say that rather than inventing a step: lead with what went wrong and the honest next move — the message below plus where to look — e.g. "houseCARL setup hit an error it does not have a fix for and stopped — the line below is what the failure reported, and the paths above it are how far it got."

Avick3110 (commented):
Blind review of PR A. Verified clean and not commented on below: the move commit is a genuine pure move — the banner (blank / two-space "houseCARL <ver>  ·  setup" / two-space rule of 63 / blank), the `[check] ` prefix, the `[host] ` + `      -> ` step pair, the `      - ` bullet and the prompt string all come out of `Ui` byte-for-byte identical to what `Program.cs` wrote before, so colour is the only added behaviour. Colour suppression is right: `NO_COLOR` is tested with `!IsNullOrEmpty` (per no-color.org, an empty value does not count) and `IsOutputRedirected` / `IsErrorRedirected` are tracked separately for the two streams, with `Console.ForegroundColor` rather than escape codes, so a captured run gets the same characters. The `Problem`-on-stderr / `Note`-on-stdout split is correct against `SetupSkillPruneProbe`: the only two callers the probe's `Console.SetOut` capture reads — `ReportNoSkillsShipped` and `ReportKeptBack` — are both `Note`, and their new text still carries the four strings T4/T5/T6 assert (`no skills folder`, `could not be deleted`, the full leftover path, and the two `removed skills this version no longer ships` step lines, unchanged via `Ui.Step`). The `InstallResult.Message` reshape is safe: `Main`'s `ServerInUse` block is the only reader, and neither probe asserts on the text; the pre-flight message still names the locked exe, and "nothing was changed" still attaches only to `RefusedBeforeAnyCopy`, so the released changelog entry for that fix stays true. `build-plugin.ps1` step 9's `-p:Version=$Version` is the same mechanism and the same `$Version` (read from `plugin.json`) as step 2, and the csproj's `<Version>0.0.0-dev</Version>` mirrors `housecarl-mcp.csproj` line 21. The attribute read survives the trimmed single-file publish: `GetCustomAttribute<AssemblyInformationalVersionAttribute>()` roots the attribute type itself, and ILLink does not strip assembly-level attribute instances without opt-in. `TryInstall`'s signature, `SetupSkillPruneProbe` and `SetupUpdateLockProbe` are untouched. `--help` is unchanged and still accurate — the four flags it lists all still work. Line endings: all five files are pure LF in the committed blobs (Ui.cs 4677 bytes / 0 CR / 111 LF; Program.cs 0 CR / 859 LF), matching main. One thing outside the findings: the PR body says "All the console output moved out of `Program.cs`", but 31 `Console.*` calls remain there — the help block, the host menu, `PrintNext`, `Finish` and the "Please type 1, 2, 3, or q." correction. That is presumably PR B's territory; worth saying so in the body rather than claiming all of it, since the file now has two ways to print.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The setup program says which version it is installing, and leads every error with one sentence.** The
+  banner now reads `houseCARL <version>  ·  setup`, stamped from the same `plugin.json` version the server
+  reports. Each refusal — the plugin folder not being beside the program, a missing .NET runtime, a server
+  file in use — now opens with one plain sentence saying what went wrong and what to try, with the paths,
+  download links and the two-installer note indented below it; the runtime refusal names only the runtime
+  that is actually missing rather than listing both. Output is coloured unless it is redirected or
+  `NO_COLOR` is set.
```
Avick3110: "Output is coloured unless it is redirected or `NO_COLOR` is set" is exactly the unpointed restatement the rule at the top of this file names as the defect: it is a second copy of a condition whose home is `Ui`'s `NoColor` / `OutColor` / `ErrColor` in `src/housecarl-setup/Ui.cs`, and the two will drift the first time someone adds a `--no-color` flag or a TERM check. Point at the bound instead — "Output is coloured; when it is not, and why, is on `Ui` in `src/housecarl-setup/Ui.cs`" — the way the neighbouring entries end on `FieldListCap` in `CorpusRulebook.cs` and on `docs/dialogue.md`. Separately, this entry is followed by a blank line while the other entries under `## Unreleased` sit flush against each other; that one blank turns the whole list loose and re-spaces every entry when it renders, so drop it to match.

src/housecarl-setup/Ui.cs
```diff
@@ -0,0 +1,111 @@
...
+        Paint(OutColor, ConsoleColor.White, () => Console.Write("houseCARL"));
+        Paint(OutColor, ConsoleColor.DarkGray, () => Console.Write(" " + version + "  ·  setup"));
+        Console.WriteLine();
+        Rule();
+        Console.WriteLine();
+    }
+
+    /// <summary>A horizontal divider the width of the banner.</summary>
+    public static void Rule()
+        => Paint(OutColor, ConsoleColor.DarkGray, () => Console.WriteLine("  " + new string('─', RuleWidth)));
```
Avick3110, line 51: Nothing sets `Console.OutputEncoding`, so these characters go out in the console's active code page, which on a stock double-clicked `houseCARL-Setup.exe` is OEM 437/850, not UTF-8. U+2500 (`─`) and U+00B7 (`·`) do exist in 437, but the em dash U+2014 does not — and this PR puts one in the middle of the lead sentence of every single refusal, where before it appeared in one `InstallResult` message. Best-fit mapping will render them as `-` or `?` on the machine that matters most here: the fresh box with no .NET installed, which is the whole reason this exe ships self-contained. Please check one run on a default `cmd.exe` before merge; if it degrades, set `Console.OutputEncoding = System.Text.Encoding.UTF8` in `Ui`'s static initialiser (guarded in a try/catch — it throws when there is no console handle) so the banner rule and the sentences render as written, or fall back to ASCII punctuation in the sentences.

Avick3110: Folded. `Ui`s static constructor now sets `Console.OutputEncoding` to UTF-8 once, in a try/catch, skipped when both streams are redirected. Aaron also ran the pre-fix build on a default cmd.exe under code page 850 and the em dash, the rule and the middot all rendered, so the fix is belt-and-braces rather than a repair, but it removes the dependence on best-fit mapping.

PR #709, opened 2026-09-10
Rewrite the plugin README for the 2.0 surface

Avick3110:
`plugin/README.md` is the README the installer copies to `~/.claude/skills/housecarl/README.md` and the marketplace shows, and it was still the whole pre-2.0 text: 1.x feature bullets, a claim that Mutagen is kept warm in memory (which the lazy-read design contradicts), no tool surface and no grammar. It is now the short installed-copy version of the top-level 2.0 README merged in #707: the opening paragraph and fact table, the 31 tools by substrate, the seven skills, requirements, both install paths as the installer lays them down, how to point it at an MO2 instance, the three write lanes and the pre-flight, licence, credits, and a pointer to the repository README for the grammar, the predicate language, coverage and the internals. Every fact is a subset of the top-level README's; the Claude Code v2.1.143 floor and its reason (earlier builds ignore `displayName`, and `--plugin-dir ./x.zip` needs v2.1.128 or newer) is kept from the old file. `packaging/START-HERE.txt` is untouched.

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -14,10 +14,9 @@ saying it sets an expectation their install may contradict. Say what is known, a
 ## Unreleased
 
 - **The plugin's own `README.md` describes the 2.0 surface.** This is the file the installer copies to
-  `~/.claude/skills/housecarl/README.md` and the marketplace shows. It carries the 31 tools by substrate,
-  the seven skills, the requirements, both install paths, how to point the server at an MO2 instance, and
-  the three write lanes. It was still the 1.x text, including a claim that Mutagen is kept warm in memory,
-  which the lazy-read design contradicts. The grammar, the predicate language, the runtime layers and the
+  `~/.claude/skills/housecarl/README.md` and the marketplace shows. It is now the installed-copy subset of
+  the top-level `README.md`, and says what it carries. It was still the 1.x text, including a claim that
+  Mutagen is kept warm in memory, which the lazy-read design contradicts. The grammar, the predicate language, the runtime layers and the
   internals stay in the top-level `README.md`, which it links to.
 - **The top-level `README.md` describes the 2.0 surface.** It names the tool surface and the axes a call is
   composed from, states what houseCARL is not, and carries the install, requirements and upgrade-from-1.x
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,6 +13,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The plugin's own `README.md` describes the 2.0 surface.** This is the file the installer copies to
+  `~/.claude/skills/housecarl/README.md` and the marketplace shows. It carries the 31 tools by substrate,
+  the seven skills, the requirements, both install paths, how to point the server at an MO2 instance, and
+  the three write lanes. It was still the 1.x text, including a claim that Mutagen is kept warm in memory,
+  which the lazy-read design contradicts. The grammar, the predicate language, the runtime layers and the
+  internals stay in the top-level `README.md`, which it links to.
 - **The top-level `README.md` describes the 2.0 surface.** It names the tool surface and the axes a call is
   composed from, states what houseCARL is not, and carries the install, requirements and upgrade-from-1.x
   notes. The 1.x-to-2.0 tool-name mapping is the refusal
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -52,7 +52,7 @@ Dialogue and facegen are not skills. Their tools carry the bookkeeping: `check f
 - Windows.
 - .NET Runtime 9.0 and ASP.NET Core Runtime 9.0, from the [.NET 9 download page](https://dotnet.microsoft.com/download/dotnet/9.0). Both are required. houseCARL ships framework-dependent, and the ASP.NET Core installer does not include the base runtime. The setup utility checks for both and names the one that is missing.
 - [Mod Organizer 2](https://www.modorganizer.org/) with a profile. MO2 does not need to be running.
-- Claude Code v2.1.143 or newer (the terminal CLI, or the Claude desktop app's Code tab), or OpenAI Codex. Earlier Claude Code builds ignore the plugin's `displayName`; loading a zipped plugin with `--plugin-dir ./x.zip` needs v2.1.128 or newer.
+- Claude Code v2.1.143 or newer (the terminal CLI, or the Claude desktop app's Code tab), or OpenAI Codex.
 
 ## Install
 
@@ -65,7 +65,7 @@ Dialogue and facegen are not skills. Their tools carry the bookkeeping: `check f
    - Codex: server under `%LOCALAPPDATA%\houseCARL\server\`, skills flat under `~/.agents/skills/` with a `$housecarl` entry point, server registered as `[mcp_servers.housecarl]` in `~/.codex/config.toml`.
 4. Restart the host.
 
-Updating: quit Claude Code and Codex first. Setup cannot replace a server a session is running; if one is, it stops and says so.
+Updating: quit Claude Code and Codex first. Setup cannot replace a server a session is running; if one is, it stops and says so. Setup overwrites only the files in the package, so the saved MO2 instance and tool paths in `houseCARL.user.json` survive a re-run.
 
 ### From source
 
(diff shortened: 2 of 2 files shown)


plugin/README.md
```diff
@@ -1,137 +1,113 @@
...
-  base .NET runtime. On Windows these are **two separate installers** — the ASP.NET Core Runtime
-  installer does **not** include the base .NET Runtime — so install both.
-- **[Mod Organizer 2](https://www.modorganizer.org/)** with a modlist. houseCARL reads the instance's
-  profile files statically — **MO2 does not need to be running.**
-- **Claude Code v2.1.143 or newer.** (Earlier builds ignore the plugin's `displayName`; loading a zipped
-  plugin with `--plugin-dir ./x.zip` needs ≥ v2.1.128.)
+- Windows.
+- .NET Runtime 9.0 and ASP.NET Core Runtime 9.0, from the [.NET 9 download page](https://dotnet.microsoft.com/download/dotnet/9.0). Both are required. houseCARL ships framework-dependent, and the ASP.NET Core installer does not include the base runtime. The setup utility checks for both and names the one that is missing.
+- [Mod Organizer 2](https://www.modorganizer.org/) with a profile. MO2 does not need to be running.
+- Claude Code v2.1.143 or newer (the terminal CLI, or the Claude desktop app's Code tab), or OpenAI Codex. Earlier Claude Code builds ignore the plugin's `displayName`; loading a zipped plugin with `--plugin-dir ./x.zip` needs v2.1.128 or newer.
```
Avick3110: The stated reason for the v2.1.143 floor does not bind the install path this file documents as primary. Both halves of it are properties of loading houseCARL as a *plugin*: `displayName` is read from `plugin/.claude-plugin/plugin.json`, and `--plugin-dir ./x.zip` is the plugin loader. But the release lane at lines 61-66 installs through `houseCARL-Setup.exe`, and `InstallForClaude` in `src/housecarl-setup/Program.cs` copies the tree to `~/.claude/skills/housecarl/` and registers the server in `~/.claude.json` — neither `plugin.json` nor `--plugin-dir` is involved, so nothing in the repo supports a 2.1.143 floor for a setup-utility install. The top-level README on `main` states the floor without a reason, which sidesteps this; here the reason is carried over from the 1.x file, when the plugin lane was the only lane. Worth either saying which path the floor applies to, or dropping the reason as the top-level file did.

plugin/README.md
```diff
@@ -1,137 +1,113 @@
...
-new MO2 instance.
+1. Download `houseCARL-<version>.zip` from the [latest release](https://github.com/Avick3110/houseCARL/releases).
+2. Extract it and run `houseCARL-Setup.exe`.
+3. Select the host: `[1] Claude Code`, `[2] Codex`, `[3] Both`.
+   - Claude Code: skills to `~/.claude/skills/housecarl/`, server registered in `~/.claude.json`. The CLI and the desktop app both read these.
+   - Codex: server under `%LOCALAPPDATA%\houseCARL\server\`, skills flat under `~/.agents/skills/` with a `$housecarl` entry point, server registered as `[mcp_servers.housecarl]` in `~/.codex/config.toml`.
+4. Restart the host.
 
-Your settings are stored under Claude Code's plugin data directory, so they survive plugin updates.
+Updating: quit Claude Code and Codex first. Setup cannot replace a server a session is running; if one is, it stops and says so.
```
Avick3110: Dropping "Your settings are stored under Claude Code's plugin data directory, so they survive plugin updates" was right as written: that location is only true on the plugin lane, where `plugin/.mcp.json` sets `HOUSECARL_DATA_DIR=${CLAUDE_PLUGIN_DATA}`. On the setup-utility lane `RegisterClaudeMcpServer` writes a `~/.claude.json` entry with `type`, `command` and `args` and no `env` at all, so `SetupHouseCarl` in `src/housecarl-mcp/Program.cs` falls back to `AppContext.BaseDirectory` and `houseCARL.user.json` lands in `~/.claude/skills/housecarl/server/`. The outcome the old sentence claimed still holds on both lanes, though: `CopyDirectory` overwrites only the files in the package, and the stale-skill prune walks `skillsDest/skills` only, so a re-run of setup never touches `houseCARL.user.json`. That is the one thing a user reading this "Updating" line wants to know, and it is now absent from the file entirely. One flat sentence naming the file and saying setup does not remove it would restore it without reintroducing the wrong location.

plugin/README.md
```diff
@@ -1,137 +1,113 @@
...
-in xEdit if you like before playing.
+The script regenerates the rulebook, publishes the server framework-dependent with trimming off, bundles the skills, builds the setup utility, and packs `release/houseCARL-<version>.zip`. Install the output with `houseCARL-Setup.exe`, with `claude --plugin-dir ./dist/housecarl`, or through the bundled local marketplace descriptor.
 
-## License
+## Pointing it at a modlist
 
-houseCARL is licensed **GPL-3.0-only** — see [LICENSE](LICENSE). This is required by Mutagen (GPL-3.0-only,
-no linking exception), which houseCARL bundles. Every third-party component and its license is listed in
-[THIRD-PARTY-NOTICES.txt](THIRD-PARTY-NOTICES.txt), along with the corresponding-source pointers.
+On first use, houseCARL asks for the MO2 instance folder, the one containing `ModOrganizer.ini`. Everything else follows from it: the mods folder, the active profile, and the load order. The instance can be changed at any time by asking.
```
Avick3110: "On first use, houseCARL asks for the MO2 instance folder" is true of the setup-utility lane, where nothing configures the instance and the tools prompt, but it is not the whole story for the lane line 80 recommends. `plugin/.claude-plugin/plugin.json` still declares `userConfig.mo2_instance_dir` as a `directory` with `"required": true`, and `plugin/.mcp.json` passes it through as `HouseCarl__Mo2InstanceDir`, so a `claude --plugin-dir ./dist/housecarl` or local-marketplace install gets the folder picker at enable time and the server boots already configured — the picker is still live, it is just no longer exercised by the release lane. The old file's folder-picker paragraph was the only place that lane was documented, and this rewrite drops it, so a reader who follows line 80 meets a prompt the README does not mention. Either the sentence should name both routes, or the `userConfig` block should go if the plugin lane is no longer supported.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **The plugin's own `README.md` describes the 2.0 surface.** This is the file the installer copies to
+  `~/.claude/skills/housecarl/README.md` and the marketplace shows. It carries the 31 tools by substrate,
```
Avick3110: The middle of this entry is the unpointed restatement the rule at the top of the file names as the defect. "It carries the 31 tools by substrate, the seven skills, the requirements, both install paths, how to point the server at an MO2 instance, and the three write lanes" is a second copy of a contents list — including two counts that also live in the fact table of both READMEs — and it is the copy nobody will edit when a tool or a skill is added. The entry's last sentence does the right thing by pointing at the split rather than restating what stayed behind; the same move works here, e.g. saying the file is now the installed-copy subset of the top-level README and letting the file itself say what it carries. The first and third sentences are fine as they stand: what the file is, and the specific false claim that came out of it.

Avick3110 (commented):
Blind review of the rewrite. What I checked and found clean: every tool name in the substrate table exists as a constant in `src/housecarl-core/ToolNames.cs`, the table sums to 31 and `ToolNames` holds exactly 31 constants; the seven skill rows match the seven directories in `.claude/skills/` and the `$Skills` list in `scripts/build-plugin.ps1`; the fact table, the requirements, the release and from-source install steps, the skills table, the writing lanes, the pre-flight list, the licence paragraph and the credits are all verbatim or faithful subsets of the top-level `README.md` on `main`, with nothing added that the top-level file does not carry; `[LICENSE](LICENSE)` and `[THIRD-PARTY-NOTICES.txt](THIRD-PARTY-NOTICES.txt)` resolve, because step 6 of `build-plugin.ps1` copies both beside `README.md` into `dist/housecarl/`, and `InstallForClaude` copies that whole tree to `~/.claude/skills/housecarl/`; the register is flat declaratives throughout, with the old file's sales opener, its bullet list and its "Talk to it" line gone and no hedging introduced; and both blobs are pure LF (0 CR bytes in 7,690 bytes of README, 0 in the CHANGELOG), matching `main` and the `.gitattributes` policy. Four findings below, none of them blocking.

PR #710, opened 2026-09-10
Show what setup found and what it will write

Avick3110:
PR B of three on the installer. Setup now reads the machine before it writes to it and shows the user what it is about to do. A detection block after the banner says which hosts are here, whether houseCARL is already installed for each and at what version (Claude Code from the copied plugin's own manifest, Codex from the installed server exe, since a Codex install copies no manifest), and whether the two .NET runtimes are there. The host menu repeats that beside each option, and where exactly one host is found a bare Enter picks it. Before any file is copied the run prints the plan: one line per destination, the exact path, labelled install or upgrade per host. The plan's paths come from the same helpers `TryInstall` copies to — `ClaudeSkillsDest`, `ClaudeJson`, `CodexServerDir`, `CodexSkillsRoot`, `CodexConfigToml` on `Program`, extracted here from the inline paths the two installers used to build — so the plan and the install cannot drift. Enter proceeds and `q` quits; the stop stands on `--claude`, `--codex` and `--both` too, and a new `--yes` skips it, as does a run whose input is redirected. `TryInstall` keeps its signature and does the same writes, and `SetupSkillPruneProbe`, `SetupUpdateLockProbe` and `DescriptionVocabularyGuardProbe` are green unedited (127/127 in `ci-all`, 1926 tests).

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -33,11 +33,19 @@ saying it sets an expectation their install may contradict. Say what is known, a
 - **The setup program says what it found, then what it is about to write, and waits for a yes.** Before it
   copies anything it prints what is on the machine — each host, whether houseCARL is installed for it and at
   what version, and the .NET runtimes — and then the plan: one line per destination, naming the exact path,
-  labelled install or upgrade per host. The plan reads the destinations from the same path helpers the
-  install writes to, on `Program` in `src/housecarl-setup/Program.cs`, so the plan and the install cannot
-  disagree. Enter proceeds and `q` quits, on `--claude`, `--codex` and `--both` runs too; `--yes` goes
-  straight on, and what else skips the stop is on `Confirmed` in that file. Where exactly one host is found,
-  Enter at the menu picks it.
+  and a per-host label from `Plan.HostAction` in `src/housecarl-setup/Plan.cs`, which compares the installed
+  version against the package's rather than only testing them for equality. The plan reads the destinations
+  from the same path helpers the install writes to, on `Program` in `src/housecarl-setup/Program.cs`, so the
+  plan and the install cannot disagree, and it says that an upgrade deletes the skill folders this version no
+  longer ships. Enter proceeds and `q` quits, on `--claude`, `--codex` and `--both` runs too; `--yes` is the
+  one way past the confirm, on `Confirm` in that file. Where exactly one host is found, Enter at the menu
+  picks it.
+- **Setup refuses a question it has nobody to ask instead of answering it.** A run whose input is redirected
+  used to take the missing keypress as a yes at the plan and as a cancel at the host menu. Both now refuse
+  with one sentence naming the flag that answers the question in advance — `--claude`, `--codex` or `--both`
+  for the host, `--yes` for the confirm — and `--help` says so. A host is also reported found only on a file
+  the host itself writes, not on a directory bearing its name; which files those are is on `Detect` in
+  `src/housecarl-setup/Detect.cs`.
 - **A refusal for an unknown field now names every field the record type has.** `housecarl_apply` and
   `housecarl_create` run the same pre-flight, and it used to name twelve fields and then `(+N more)` with no
   route from the call to the rest — a write call is the only place the surface shows a type's schema without
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -30,6 +30,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   download links and the two-installer note indented below it; the runtime refusal names only the runtime
   that is actually missing rather than listing both. Output is coloured; when it is not, and why, is on `Ui`
   in `src/housecarl-setup/Ui.cs`.
+- **The setup program says what it found, then what it is about to write, and waits for a yes.** Before it
+  copies anything it prints what is on the machine — each host, whether houseCARL is installed for it and at
+  what version, and the .NET runtimes — and then the plan: one line per destination, naming the exact path,
+  labelled install or upgrade per host. The plan reads the destinations from the same path helpers the
+  install writes to, on `Program` in `src/housecarl-setup/Program.cs`, so the plan and the install cannot
+  disagree. Enter proceeds and `q` quits, on `--claude`, `--codex` and `--both` runs too; `--yes` goes
+  straight on, and what else skips the stop is on `Confirmed` in that file. Where exactly one host is found,
+  Enter at the menu picks it.
 - **A refusal for an unknown field now names every field the record type has.** `housecarl_apply` and
   `housecarl_create` run the same pre-flight, and it used to name twelve fields and then `(+N more)` with no
   route from the call to the rest — a write call is the only place the surface shows a type's schema without
--- a/src/housecarl-setup/Detect.cs
+++ b/src/housecarl-setup/Detect.cs
@@ -1,0 +1,88 @@
+using System.Diagnostics;
+using System.Text.Json;
+
+namespace HousecarlSetup;
+
+/// <summary>
+/// What is already on this machine, read before setup writes anything: which hosts are here, whether houseCARL
+/// is installed for each, and at what version. It is its own file because it is its own job - <see cref="Program"/>
+/// stays the install, and the paths it reads come from <see cref="Program"/>'s path helpers so a destination
+/// cannot drift between what is reported and what is written.
+///
+/// Reading only. Nothing here creates a directory or a file, so a run that is cancelled at the plan leaves the
+/// machine exactly as it found it.
+/// </summary>
+public static class Detect
+{
+    /// <summary>One host as it stands before the install.</summary>
+    /// <param name="Name">The host's name, as the menu and the plan say it.</param>
+    /// <param name="Present">True when the host itself is on this machine.</param>
+    /// <param name="Installed">True when houseCARL is already installed for it.</param>
+    /// <param name="InstalledVersion">The installed version, or null when it is installed but unreadable.</param>
+    public sealed record HostState(string Name, bool Present, bool Installed, string? InstalledVersion)
+    {
+        /// <summary>The right-hand side of this host's row in the detection block and the menu.</summary>
+        public string Summary =>
+            !Present    ? "not found"
+            : !Installed ? "found  ·  houseCARL not installed"
+            : InstalledVersion is null ? "found  ·  houseCARL installed"
+            :                            "found  ·  houseCARL " + InstalledVersion + " installed";
+    }
+
+    /// <summary>Claude Code: the desktop app keeps ~/.claude.json and ~/.claude, and either one is it being here.
+    /// The installed version is the copied plugin's own manifest, which is the file that shipped it.</summary>
+    public static HostState Claude(string home)
+    {
+        bool present = File.Exists(Program.ClaudeJson(home)) || Directory.Exists(Path.Combine(home, ".claude"));
+        string dest  = Program.ClaudeSkillsDest(home);
+        bool installed = File.Exists(Program.ClaudeDestExe(home)) || Directory.Exists(dest);
+        string? version = installed
+            ? PluginVersion(Path.Combine(dest, ".claude-plugin", "plugin.json"))
+            : null;
+        return new HostState("Claude Code", present, installed, version);
+    }
+
+    /// <summary>Codex: ~/.codex (or CODEX_HOME) is it being here. A Codex install copies the server but not the
+    /// plugin manifest, so the version comes off the installed server exe, which build-plugin.ps1 stamps.</summary>
+    public static HostState Codex(string home, string? homeOverride)
+    {
+        bool present = Directory.Exists(Program.CodexConfigHome(home));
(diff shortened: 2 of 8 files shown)


src/housecarl-setup/Plan.cs
```diff
@@ -0,0 +1,69 @@
...
+            foreach (Line line in plan.Lines)
+                Ui.PlanLine(line.Path, line.What, pathWidth);
+        }
+
+        Ui.Body("Nothing else is touched. Your game, your mods and your MO2 profile are not read.");
+    }
+
+    /// <summary>What this run is to this host: a first install, or an upgrade off the version already there.</summary>
+    private static string HostAction(Detect.HostState host, string version)
+        => !host.Installed                ? "install " + version
```
Avick3110: A newer installed version is labelled an upgrade to an older one. `HostAction` only asks whether the two strings are equal, so every unequal pair falls into the last arm: installing a 2.0.0 package over a 2.5.0 install prints `Codex  .  upgrade 2.5.0 to 2.0.0`. I drove this against a scratch home and that is the literal line. Downgrading is a thing a user does deliberately (a bad release, a pinned modlist), but the one screen whose whole job is to say what is about to happen tells them the opposite of what will happen, and the plan is also the only place they could have caught it. A semver comparison of the two strings would give a fourth arm, `downgrade 2.5.0 to 2.0.0`; where either side does not parse, saying `replace 2.5.0 with 2.0.0` is at least not a claim about direction. Related: the equality on line 66 compares a string from `plugin.json` (Claude) or from a Win32 resource (Codex) against `plugin.json`, so a build whose exe resource reads `2.0.0.0` where the manifest says `2.0.0` would say `upgrade` on a reinstall of the identical package.

src/housecarl-setup/Plan.cs
```diff
@@ -0,0 +1,69 @@
...
+        foreach (HostPlan plan in plans)
+        {
+            if (!first) Console.WriteLine();
+            first = false;
+            Ui.PlanHost(plan.Host.Name, HostAction(plan.Host, version));
+            foreach (Line line in plan.Lines)
+                Ui.PlanLine(line.Path, line.What, pathWidth);
+        }
+
+        Ui.Body("Nothing else is touched. Your game, your mods and your MO2 profile are not read.");
```
Avick3110: "Nothing else is touched" is not true of the Codex install. `InstallForCodex` writes the skill record at `CodexSkillRecord`, which is `Path.GetDirectoryName(CodexServerDir(...))` + `installed-skills.txt` - i.e. `%LOCALAPPDATA%\houseCARL\installed-skills.txt`, a sibling of the `server` directory the plan names and under none of the three listed paths. A `--codex` run into a scratch home left exactly four files: the two under the listed roots, `config.toml`, and `houseCARL\installed-skills.txt`. Either add the record to the plan lines or drop the absolute claim; a sentence that promises a closed set is worth more than a plan line, and it is the sentence that goes stale first. The other half of the same problem is the verb: the heading says "This will write" / "This will replace", but the run also deletes - stale skill folders under `~/.agents/skills`, a root shared with every other agent. That is the destructive part of the install and the plan is the screen where it should be visible, even though the folders sit under a path the plan does name.

src/housecarl-setup/Program.cs
```diff
@@ -179,6 +193,39 @@ private static int Main(string[] args)
...
+    }
+
+    /// <summary>
+    /// The plan's confirm. It stands on every attended run, including one started with a host flag, because the
+    /// paths are the thing worth reading before they are written. An unattended run - <c>--yes</c>, or a run whose
+    /// input is redirected and so has nobody to press a key - goes straight on.
+    /// </summary>
+    private static bool Confirmed(string[] args)
+    {
+        if (args.Contains("--yes") || Console.IsInputRedirected) return true;
```
Avick3110: Redirected stdin is read here as consent, and the run says nothing about it. `houseCARL-Setup.exe --claude < NUL` (verified against a scratch home) prints the detection block, prints the plan, and then copies without a pause and without a line saying the confirm was skipped - the confirm prompt is never even printed, because this returns before `Ui.Prompt`. That is what `--yes` is for; making the absence of a keyboard mean the same thing gives the flag nothing to do and hands a scripted or piped run a write it never asked for. Worse, the same condition gets opposite defaults inside one run: `ResolveTarget` line 492 treats a null read as cancel ("no interactive input (redirected) - treat as cancel") while line 221 here treats it as proceed ("no interactive input after all"). One of those is wrong, and it is the one that writes. I would make `--yes` the only thing that skips the stop, and refuse a redirected run without it with the one sentence this file already writes so well - it cannot ask, so it says so.

src/housecarl-setup/Program.cs
```diff
@@ -72,6 +72,7 @@ private static int Main(string[] args)
             Console.WriteLine("    --claude   install for Claude Code only");
             Console.WriteLine("    --codex    install for Codex only");
             Console.WriteLine("    --both     install for both");
+            Console.WriteLine("    --yes      install without stopping at the plan (unattended runs)");
```
Avick3110: `--help` documents `--yes` as the way past the plan, but not the other way past it: a run whose stdin is redirected skips the stop with no flag at all (line 217). Someone piping this from a script gets a write with no confirm and no way to learn from `--help` that they were going to. The second half is the `(unattended runs)` parenthetical: `--yes` alone is not an unattended run, because without `--claude`/`--codex`/`--both` the host menu still blocks on `Console.ReadLine`. Either say what `--yes` actually skips ("skip the confirm at the plan; still asks which host unless a host flag is given") or point at the pair of flags an unattended run needs.

src/housecarl-setup/Detect.cs
```diff
@@ -0,0 +1,88 @@
...
+            :                            "found  ·  houseCARL " + InstalledVersion + " installed";
+    }
+
+    /// <summary>Claude Code: the desktop app keeps ~/.claude.json and ~/.claude, and either one is it being here.
+    /// The installed version is the copied plugin's own manifest, which is the file that shipped it.</summary>
+    public static HostState Claude(string home)
+    {
+        bool present = File.Exists(Program.ClaudeJson(home)) || Directory.Exists(Path.Combine(home, ".claude"));
+        string dest  = Program.ClaudeSkillsDest(home);
+        bool installed = File.Exists(Program.ClaudeDestExe(home)) || Directory.Exists(dest);
```
Avick3110: `Directory.Exists(dest)` on its own is enough to call houseCARL installed for Claude Code, so an empty or half-deleted `~/.claude/skills/housecarl/` reports "found . houseCARL installed" with no version, and the plan then heads that host `upgrade to 2.0.0` for what is in fact a first install. The exe test above it is the real signal and the directory test only widens it. Compare `Detect.Codex`, which keys `installed` on the exe alone and is right for it. If the directory arm is there to catch an install whose exe was deleted, the honest reading of that state is still "not installed" - there is nothing to upgrade from.

src/housecarl-setup/Program.cs
```diff
@@ -393,22 +465,34 @@ private static List<string> MissingServerRuntimes()
...
-        Console.WriteLine("  [2] Codex");
-        Console.WriteLine("  [3] Both");
+        Ui.Heading("Install houseCARL for which agent?");
+        Ui.MenuItem("1", claude.Name, claude.Summary);
+        Ui.MenuItem("2", codex.Name,  codex.Summary);
+        Ui.MenuItem("3", "Both", "");
         Console.WriteLine();
+
+        // Exactly one host on this machine leaves nothing to choose between, so a bare Enter takes it.
+        Target? onlyHost = claude.Present && !codex.Present  ? Target.Claude
```
Avick3110, line 537: The bare-Enter default keys on `Present`, which is the weakest fact detection has. For Codex, `Present` is `Directory.Exists(~/.codex)` and nothing more: a `~/.codex` left behind by an uninstalled Codex, or created by any tool that has ever touched that path, makes Codex the default on a machine that does not run it. I created an empty `~/.codex` under a scratch home with no Claude Code, piped a bare newline, and the run installed for Codex - menu line, plan and all - off a single empty directory. The mitigation in the code is that the question names the host it will pick, which is real and worth keeping. But a default should be at least as strong as the evidence: keying it on `Installed || Present` for one host while the other is neither, or simply not defaulting when the only signal is a bare directory, would leave the Enter key meaning "the host I already use" rather than "the host that left a folder here".

src/housecarl-setup/Program.cs
```diff
@@ -115,39 +116,52 @@ private static int Main(string[] args)
...
             // ASP.NET Core one does NOT include the base runtime -- a real-world install trap.
             // This exe ships self-contained precisely so it still runs on a machine with neither
             // and can say exactly what's missing, instead of the install "succeeding" into a
             // server that never starts.
-            if (!args.Contains("--skip-runtime-check"))
+            bool skipRuntimeCheck = args.Contains("--skip-runtime-check");
+            List<string> missingRuntimes = skipRuntimeCheck ? new List<string>() : MissingServerRuntimes();
+            Detect.HostState claude = Detect.Claude(home);
+            Detect.HostState codex  = Detect.Codex(home, homeOverride);
+            ReportDetected(claude, codex, missingRuntimes, skipRuntimeCheck);
```
Avick3110, line 147: This line takes over the last thing `Ui.Ok` was used for - the removed `Ui.Ok(".NET Runtime 9 + ASP.NET Core Runtime 9: found.")` - and `Ok` now has no caller anywhere in `src` (only its own definition at `Ui.cs:113` matches). It is a green `[check]` prefix that nothing prints, in the file this PR series has been narrowing to the shapes the setup actually uses. Delete it; if a later PR wants a passed-check line back it will want it in the detection block's shape, not this one.

src/housecarl-setup/Program.cs
```diff
@@ -274,13 +321,38 @@ public static InstallResult TryInstall(Target target, string pluginSrc, string h
         return new InstallResult(InstallOutcome.Installed, null);
     }
 
+    // ---- destination paths (one source of truth for pre-flight, plan and installer) ----
+
+    /// <summary>Where the Claude install puts the plugin tree.</summary>
+    internal static string ClaudeSkillsDest(string home)
```
Avick3110, line 356: The extraction does what it says for the plan and the two installers - I checked each of the five destinations and none of them is spelled out twice in this file any more. It stops at the assembly boundary, though: `internal` keeps the helpers away from `housecarl-generator`, where `SetupUpdateLockProbe.cs:58-64` still hand-builds `Path.Combine(home, ".claude", "skills", "housecarl", "server", "housecarl-mcp.exe")` and its three siblings under a comment that admits it is mirroring what `TryInstall` computes. Those are the same strings, built a second time, in the file whose job is to catch a destination going wrong - the guard would follow the installer straight off a cliff. `public`, or an `InternalsVisibleTo("housecarl-generator")` like the other projects here already use, would let the probe assert against the helper instead of against its own copy.

plugin/CHANGELOG.md
```diff
@@ -25,6 +25,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
   download links and the two-installer note indented below it; the runtime refusal names only the runtime
   that is actually missing rather than listing both. Output is coloured; when it is not, and why, is on `Ui`
   in `src/housecarl-setup/Ui.cs`.
+- **The setup program says what it found, then what it is about to write, and waits for a yes.** Before it
+  copies anything it prints what is on the machine — each host, whether houseCARL is installed for it and at
+  what version, and the .NET runtimes — and then the plan: one line per destination, naming the exact path,
+  labelled install or upgrade per host. The plan reads the destinations from the same path helpers the
```
Avick3110: "labelled install or upgrade per host" restates the label set instead of pointing at it, and is already short a label: `Plan.HostAction` also emits `reinstall <version>` when the installed version equals the package's. That is the defect the rule at the top of this file names - a second copy of a fact that will be edited once - and it has drifted before the entry has even shipped. The entry does the pointing well two sentences later ("what else skips the stop is on `Confirmed` in that file"); the same move here, "labelled per host from `Plan.HostAction` in `src/housecarl-setup/Plan.cs`", says more and cannot go stale.

Avick3110 (commented):
Blind review of the detection + plan screen. What I checked: the plan and the install both resolving every destination through the new `Program` helpers; the four install/upgrade/reinstall/downgrade label combinations, driven for real against a scratch `HOUSECARL_SETUP_HOME` (never the real home, stdin and stdout redirected, no console window); the Codex `FileVersionInfo` read against a present-but-unstamped exe and a missing one; what a redirected-stdin run actually writes; the bare-Enter default with a leftover host directory; the full file inventory a Codex install leaves behind versus what the plan promised; the `--skip-runtime-check` rows; `Ui.Ok`'s callers; the changelog against the pointing-at-bounds rule; and line endings. `dotnet build housecarl.sln -c Release` is clean (0 errors) and `ci-all` is 127/127 on the branch, with the three setup/vocabulary probes untouched in the diff. Every new and changed file is LF-only in the blob (Detect.cs 0 CR / 88 LF, Plan.cs 0 / 69, Ui.cs 0 / 170, Program.cs 0 / 957, CHANGELOG.md 0 / 3795), matching main. Nine findings inline; the two I would not merge without are the downgrade label and the file the plan does not name.

PR #712, opened 2026-09-10
Add an uninstall to the setup program, with a summary block

Avick3110:
PR C of three on the installer. It was written stacked on #710 (installer PR B, the plan screen) and is now rebased onto main, which #710 has merged into, so it targets main directly. Setup can now remove houseCARL as well as install it: `[4] Uninstall` on the menu, or `--uninstall` with `--claude`/`--codex`/`--both` and `--yes`, following the same one-rule stdin handling PR B added. It takes back exactly what the install wrote, per host — the tree at each install location, the skill folders setup recorded installing (only those, because `~/.agents/skills` is shared with every other agent's skills), the record itself, and the `housecarl` entry in `~/.claude.json` and `~/.codex/config.toml`, each backed up to a `.houseCARL.bak` and then spliced out beside the existing insert splice so every other byte in the file stays where it was. The Claude install now writes an installed-skills record too, mirroring the Codex one, so its removal is by record rather than by glob; an uninstall of an older install without one removes the `housecarl` tree it owns and says that is the route it took. The removal plan is shown before anything is deleted, with the same Enter/q confirm, and both install and uninstall now close with a summary block — what was written or removed, where, and what to do next — built from the same plan the confirm was given for so it cannot name a path the run never touched. The .NET runtime refusal moved to after the host choice, because the runtimes are the server's and refusing at the detection block left a machine missing one with no way to uninstall. `TryInstall` keeps its signature; a new `setup-uninstall` CI probe installs over two configs that already carry another MCP server, uninstalls, and asserts both files are byte-identical afterwards and no houseCARL file remains. `packaging/START-HERE.txt` and both README Install sections describe the new screens and flags.

### DIFF
--- a/src/housecarl-generator/SetupUninstallProbe.cs
+++ b/src/housecarl-generator/SetupUninstallProbe.cs
@@ -1,0 +1,215 @@
+using System.Text.Json.Nodes;
+using SetupProgram = HousecarlSetup.Program;
+using SetupUninstall = HousecarlSetup.Uninstall;
+
+namespace HousecarlGenerator;
+
+/// <summary>
+/// Setup uninstall guard (installer PR C). Until this there was no way to remove houseCARL other than deleting
+/// folders by hand and editing ~/.claude.json and ~/.codex/config.toml, and a hand edit of either is where a
+/// user loses the rest of the file. The removal is two splices beside the install's insert splices, so the
+/// property that matters is not "the entry is gone" but "the entry is gone AND every other byte is where it
+/// was". This drives the REAL <see cref="HousecarlSetup.Uninstall.TryUninstall"/> over a synthetic package, a
+/// temp home, and two host configs that already carry somebody else's MCP server:
+///
+///   T1  install for Both over the two pre-existing configs, then uninstall Both.
+///   T2  ~/.claude.json is byte-identical to before the install, and ~/.codex/config.toml likewise: the
+///       other server, the unrelated root key, the comment and the newline style all survive the round trip.
+///   T3  no houseCARL file is left: the Claude tree, the Codex server dir (the saved MO2 instance with it),
+///       the skill record and the houseCARL data dir are all gone.
+///   T4  under the SHARED ~/.agents/skills only the folders the record named went — a skill houseCARL never
+///       installed is untouched.
+///   T5  both configs were copied to a .houseCARL.bak before being edited.
+///   T6  an install too old to have left a Claude skill record still has its tree removed, and the run says
+///       that is the route it took.
+///   T7  with no Codex record, nothing under the shared skills root is removed, and the run says so.
+///   T8  a running session's server exe is pre-flighted: the uninstall refuses before deleting anything, and
+///       the installed tree is still there.
+///
+/// Self-contained: synthetic files in temp, no game data / no MO2 / no real host config. RED without the
+/// removal splices: T2 fails on both files (a whole-file rewrite reformats them, a naive line delete leaves the
+/// blank line the insert added), T3 leaves the tree, T4 takes the foreign skill with a directory diff, and T8
+/// deletes into a live install.
+///
+/// Run: dotnet run --project src/housecarl-generator setup-uninstall
+/// </summary>
+internal static class SetupUninstallProbe
+{
+    [CiProbe("setup-uninstall")]
+    public static int RunGuard(string[] args)
+    {
+        Console.WriteLine("================================================================");
+        Console.WriteLine(" setup uninstall guard — the removal takes back exactly what the install wrote");
+        Console.WriteLine("================================================================");
+        Console.WriteLine();
+        int fail = 0;
+        void Check(bool c, string label) { Console.WriteLine((c ? "  PASS  " : "  FAIL  ") + label); if (!c) fail++; }
+
+        // Keep the Codex destination hermetic (the installer honours CODEX_HOME for config.toml).
+        Environment.SetEnvironmentVariable("CODEX_HOME", null);
+
+        string root = Path.Combine(Path.GetTempPath(), "hc-setup-uninstall-" + Guid.NewGuid().ToString("N"));
+        string pkg  = Path.Combine(root, "package");
+        string src  = Path.Combine(pkg, "housecarl");
+        string home = Path.Combine(root, "home");
+
+        // Asked of the installer's own path helpers, never rebuilt here.
+        string claudeDest   = SetupProgram.ClaudeSkillsDest(home);
+        string claudeJson   = SetupProgram.ClaudeJson(home);
+        string claudeRecord = SetupProgram.ClaudeSkillRecord(home);
+        string codexServer  = SetupProgram.CodexServerDir(home, home);
+        string codexSkills  = SetupProgram.CodexSkillsRoot(home);
+        string codexRecord  = SetupProgram.CodexSkillRecord(home, home);
+        string codexToml    = SetupProgram.CodexConfigToml(home);
+        string codexDataDir = Path.GetDirectoryName(codexServer)!;
+
+        var (codexPkgPath, _) = SetupUpdateLockProbe.CodexPackagePath();
+        string umbrellaPkgDir = Path.Combine(new[] { pkg }
+            .Concat((codexPkgPath ?? "codex/skills").Split('/'))
+            .Append("housecarl").ToArray());
+
+        try
+        {
+            // ---- a synthetic package ----
+            WriteFile(Path.Combine(src, ".claude-plugin", "plugin.json"), "{}");
+            WriteFile(Path.Combine(src, "server", "housecarl-mcp.exe"), "exe");
+            foreach (string s in new[] { "skill-one", "skill-two" })
+                WriteFile(Path.Combine(src, "skills", s, "SKILL.md"), s);
+            WriteFile(Path.Combine(umbrellaPkgDir, "SKILL.md"), "umbrella");
+
+            // ---- two host configs that already carry somebody else's server ----
+            WriteFile(claudeJson, ExistingClaudeJson());
+            WriteFile(codexToml, "# my own config\n[other]\nkey = 1\n");
+            string claudeBefore = File.ReadAllText(claudeJson);
+            string tomlBefore   = File.ReadAllText(codexToml);
+
+            // A skill some other tool put in the shared ~/.agents/skills.
+            WriteFile(Path.Combine(codexSkills, "someone-elses-skill", "SKILL.md"), "not ours");
+
+            Console.WriteLine("--- T1: install for Both, then uninstall Both ---");
+            var installed = SetupProgram.TryInstall(SetupProgram.Target.Both, src, home, home);
+            Check(installed.Outcome == SetupProgram.InstallOutcome.Installed, "the install ran");
+            Check(File.Exists(claudeRecord), "the Claude install recorded which skills it wrote");
+            Check(File.ReadAllText(claudeJson).Contains("housecarl"), "the Claude config carries the entry");
+            Check(File.ReadAllText(codexToml).Contains("[mcp_servers.housecarl]"), "the Codex config carries the table");
+            // The saved MO2 instance the server writes beside itself — the removal takes it with the server dir.
+            WriteFile(Path.Combine(codexServer, "houseCARL.user.json"), "{\"Mo2InstanceDir\":\"D:\\\\MO2\"}");
+            WriteFile(Path.Combine(claudeDest, "server", "houseCARL.user.json"), "{\"Mo2InstanceDir\":\"D:\\\\MO2\"}");
+
+            SetupUninstall.Result removed = null!;
+            string said = Capture(() => removed = SetupUninstall.TryUninstall(SetupProgram.Target.Both, home, home));
+            Check(removed.What == SetupUninstall.Outcome.Removed, "the uninstall ran");
+
+            Console.WriteLine();
+            Console.WriteLine("--- T2: both host configs are byte-identical to before the install ---");
+            Check(File.ReadAllText(claudeJson) == claudeBefore,
+                  "~/.claude.json is back to its original bytes, entry and all else");
+            Check(File.ReadAllText(codexToml) == tomlBefore,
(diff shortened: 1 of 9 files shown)


src/housecarl-setup/Uninstall.cs
```diff
@@ -0,0 +1,274 @@
...
+        bool removed = false;
+        for (int i = 0; i < lines.Length; i++)
+        {
+            if (!removed && lines[i].Trim() == head)
+            {
+                int j = i + 1;
+                while (j < lines.Length)
+                {
+                    string t = lines[j].Trim();
+                    if (t.StartsWith("[") && t != head && !t.StartsWith(sub)) break;
```
Avick3110: This scan runs from the table header to the next line that starts with `[`, and everything in between is dropped — which means a comment or blank line the user wrote directly above the *following* table goes with our table. I transcribed `RemoveTomlTable` and ran it: given `# my own config` / `[other]` / `key = 1` / blank / `[mcp_servers.housecarl]` / `command = 'x'` / blank / `# my foo server` / `[mcp_servers.foo]` / `command = 'f'`, the result is `# my own config` / `[other]` / `key = 1` / blank / `[mcp_servers.foo]` / `command = 'f'` — the user's `# my foo server` comment is gone. TOML attaches a leading comment to the table below it by convention, so that comment is somebody else's text, not the tail of our table. This is the one case in the whole PR where the removal deletes something houseCARL did not write, and it is silent. The insert splice shares the shape, but there it only bites on an upgrade and there is no claim of byte preservation attached; here the doc comment two lines up and the CHANGELOG both promise every other byte stays where it was. The fix is to stop the body scan at the last non-blank, non-comment line rather than at the next `[`.

src/housecarl-setup/Uninstall.cs
```diff
@@ -0,0 +1,274 @@
...
+                removed = true;
+                continue;
+            }
+            outLines.Add(lines[i]);
+        }
+
+        if (!removed) return text;
+
+        while (outLines.Count > 0 && outLines[^1].Trim().Length == 0) outLines.RemoveAt(outLines.Count - 1);
+        return outLines.Count == 0 ? "" : string.Join(nl, outLines) + nl;
```
Avick3110: These last two lines run unconditionally, not just on the last-in-file branch, so they rewrite the tail of any file our table was removed from the middle of. Two concrete losses from the same transcription run: a config ending `[z]` / `a=1` / blank / blank keeps only `[z]` / `a=1` and one newline, and a config that ended without a trailing newline gets one added. Neither is our table. There is a third: `nl` is decided by whether the file contains any CRLF at all, and every line is then rejoined with it, so one stray CRLF in an otherwise LF file converts the whole file to CRLF (and the reverse never happens, since a pure-CRLF file stays CRLF). Separately, both `UnregisterCodexMcpServer` and `UnregisterClaudeMcpServer` read with `File.ReadAllText`, which strips a UTF-8 BOM, and write with `File.WriteAllText`, which does not write one — so a config.toml or .claude.json that had a BOM loses it. The .houseCARL.bak means none of this is unrecoverable, but the promise in the doc comment, the README and the CHANGELOG is byte preservation, and it holds only for the exact shape the probe fixture uses.

src/housecarl-setup/Uninstall.cs
```diff
@@ -0,0 +1,274 @@
...
+    ///
+    /// A table the insert APPENDED to the end of the file was written after a blank line, and one it wrote into
+    /// a file it created was written under a comment line. Those go with the table when the table is last in the
+    /// file, which is where the insert put them; a table with something after it leaves what is above it alone,
+    /// because there the blank line is the user's own separator.
+    /// </summary>
+    internal static string RemoveTomlTable(string text, string name)
+    {
+        string nl   = text.Contains("\r\n") ? "\r\n" : "\n";
+        string head = "[mcp_servers." + name + "]";
```
Avick3110: The table is matched by exact string equality against `[mcp_servers.housecarl]` on the trimmed line, but TOML permits `[mcp_servers."housecarl"]` and `[mcp_servers . housecarl]` for the same key, and Codex accepts a config written either way (a hand-edited config, or one another tool wrote). I confirmed the behaviour: with `[mcp_servers."housecarl"]` in the file, `RemoveTomlTable` returns the text unchanged, `UnregisterCodexMcpServer` returns without writing, and `TryUninstall` still returns `Outcome.Removed` — so the run tells the user houseCARL is removed from Codex while the server stays registered and Codex keeps trying to launch a deleted exe. That is the silently-wrong-answer case the second cornerstone forbids: if the exact header is not found but a `mcp_servers` key for us plausibly is, the run should say it could not find the entry and name the file, rather than report success. The same reasoning applies to the dotted-key form (`[mcp_servers]` with `housecarl = { ... }`), which neither splice handles. Note that the sibling-prefix case is fine — I checked `[mcp_servers.housecarl2]` and it is correctly left alone, both as a header and as a scan terminator.

src/housecarl-setup/Uninstall.cs
```diff
@@ -0,0 +1,274 @@
...
+    {
+        if (!File.Exists(claudeJsonPath)) return;
+
+        string text = File.ReadAllText(claudeJsonPath);
+        if (Program.FindRootMemberObject(text, "mcpServers") is not { } b) return;
+
+        string objText = text.Substring(b.start, b.end - b.start + 1);
+        JsonObject servers = JsonNode.Parse(objText) as JsonObject
+            ?? throw new InvalidDataException("mcpServers is not a JSON object.");
+        if (!servers.Remove(name)) return; // nothing of ours registered: the file is not touched
```
Avick3110: When `housecarl` is the only entry, this leaves `"mcpServers": {}` behind — and `RegisterClaudeMcpServer` has a branch (Program.cs, the `else` under `FindRootMemberObject`) that *creates* the `mcpServers` key, with its own newline and trailing comma, in a file that never had one. So for the very common case of a first-ever install on a `~/.claude.json` with no MCP servers, install-then-uninstall does not return the file to its original bytes: it leaves an empty object and a key the user never had. Either the removal should drop the whole `mcpServers` member when it empties one it can tell we created, or the round-trip claim in the CHANGELOG and both READMEs needs the bound stated. The probe cannot see this because `ExistingClaudeJson()` always seeds an `mcpServers` with another server in it — the one shape where leaving `{}` happens not to matter. Worth also noting that this `File.Copy` to `.houseCARL.bak` overwrites the backup the *install* took, so after an uninstall the user's pre-install config is no longer recoverable from the .bak; that is probably fine but it is not what a reader of "a .houseCARL.bak copy is made first" would assume.

src/housecarl-setup/Uninstall.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            List<string> recorded = Program.ReadSkillRecord(Program.ClaudeSkillRecord(home));
+            Ui.Step("Claude Code", "removing the skills, the server and the saved MO2 instance", dest);
+            foreach (string name in recorded) Ui.Bullet(name);
+            DeleteTree(dest, keptBack);
+
+            // An install from before setup started recording what it put here leaves no record. The tree is
+            // houseCARL's outright either way, so the removal is the same one - but which of the two it was is
+            // said rather than left for the reader to infer from a missing list of names.
+            if (recorded.Count == 0)
```
Avick3110, line 175: The Claude-side record is decorative: `recorded` is read on line 92 and used only to print bullets and to pick this note — the `DeleteTree(dest)` above runs identically either way, taking the whole `~/.claude/skills/housecarl` tree. So the note's contrast is false. It says the no-record route "took the whole houseCARL folder it owns", which reads as "unlike the other route", when both routes do exactly that; and the PR body and the CHANGELOG entry both say the Claude removal "is by record rather than by glob", which the code does not do. Pick one: either make the removal actually honour the record (delete the recorded folders under `dest/skills`, the server dir and the manifest, and leave anything else the user put in that tree), or drop the note and the two prose claims and say plainly that the Claude tree is houseCARL's outright and goes whole. As written a reader is told a safety property the code does not have. Related, and worth a line of comment either way: `DeleteTree` is a bare `Directory.Delete(dir, recursive: true)`, so what happens when `~/.claude/skills/housecarl` is itself a junction or a symlink the user made — link removed, or target emptied — is left to .NET's reparse-point handling rather than stated anywhere.

src/housecarl-setup/Program.cs
```diff
@@ -95,7 +99,10 @@ private static int Main(string[] args)
             // The skills folder is checked with the manifest and the exe: a package always ships one, and a
             // package missing it is a broken unzip, not a version that dropped every skill.
             string srcSkills   = Path.Combine(pluginSrc, "skills");
-            if (!File.Exists(srcManifest) || !File.Exists(srcExe) || !Directory.Exists(srcSkills))
+            // A removal reads nothing out of the package — what it takes off the machine is what the install put
+            // there — so a --uninstall run is not held to having one beside it.
+            bool needsPackage = !args.Contains("--uninstall");
```
Avick3110: `needsPackage` is keyed on the `--uninstall` *flag*, but the route both START-HERE.txt and the README document as the primary one is the menu: "Run houseCARL-Setup.exe and pick [4] Uninstall". A user on that route never reaches the menu if the package beside the exe is missing or half-unzipped — they get the broken-package refusal and no way to remove houseCARL. That is the same failure this PR fixes for the .NET runtime check four hunks down, with the same reasoning ("a machine missing a runtime is exactly a machine somebody may want to remove houseCARL from"), so it is odd to leave it standing here. Since nothing before the menu needs the package except `Detect.PluginVersion(srcManifest)` — which already falls back to `SetupVersion()` and is unused on a removal — the package check can move to sit beside the runtime refusal, after `ResolveChoice`, guarded by `mode == Mode.Install`.

src/housecarl-setup/Plan.cs
```diff
@@ -67,15 +98,71 @@ public static void Print(List<HostPlan> plans, string version)
...
+        Ui.Heading(removing ? "Removed" : "Written");
+        int pathWidth = plans.SelectMany(p => p.Lines).Select(l => l.Path.Length).DefaultIfEmpty(0).Max();
+        bool first = true;
+        foreach (HostPlan plan in plans)
+        {
+            if (!first) Console.WriteLine();
+            first = false;
+            Ui.PlanHost(plan.Host.Name, "");
+            foreach (Line line in plan.Lines)
+                Ui.PlanLine(line.Path, line.What, pathWidth);
```
Avick3110, line 185: The summary reuses the plan's `Line.What` strings verbatim, and those are present-tense statements of intent, not reports. Two consequences. First the reading: under a heading of "Removed", the line renders as `~/.claude.json    removes the server entry  (a .houseCARL.bak copy is made first)`, which is a sentence about the future in a block about the past. Second, and the reason it matters: the plan lines are unconditional, so the summary asserts work that did not happen. Remove for a host where nothing is installed — `RemovalAction` will have honestly said "houseCARL is not installed for it" on the plan screen — and the summary still prints both lines and the green "houseCARL is removed from Claude Code." The Codex case is sharper: with no record, `RemoveForCodex` deliberately removes nothing under `~/.agents/skills` and says so, yet the summary still prints `~/.agents/skills    only the skills the record below names`, and when there is no `housecarl` entry in a config the file is not written at all yet the summary claims the entry came out. The doc comment says the summary "cannot name a path the run never touched" — reusing the plan gets you paths the run *intended* to touch, which is a weaker property than the one claimed. Carrying back what each removal step actually did (or at minimum suppressing lines whose target did not exist) would close the gap.

src/housecarl-generator/SetupUninstallProbe.cs
```diff
@@ -0,0 +1,215 @@
...
+            // ---- a synthetic package ----
+            WriteFile(Path.Combine(src, ".claude-plugin", "plugin.json"), "{}");
+            WriteFile(Path.Combine(src, "server", "housecarl-mcp.exe"), "exe");
+            foreach (string s in new[] { "skill-one", "skill-two" })
+                WriteFile(Path.Combine(src, "skills", s, "SKILL.md"), s);
+            WriteFile(Path.Combine(umbrellaPkgDir, "SKILL.md"), "umbrella");
+
+            // ---- two host configs that already carry somebody else's server ----
+            WriteFile(claudeJson, ExistingClaudeJson());
+            WriteFile(codexToml, "# my own config\n[other]\nkey = 1\n");
```
Avick3110, line 92: The probe is well built — real seam, real path helpers, hermetic home, `CODEX_HOME` cleared, and T8's pre-flight assertion is exactly the right property — but T2's "byte-identical" claim rests on a single fixture shape, and every splice bug I found lives outside it. This config.toml is LF-only, ends with a trailing newline, has no BOM, and puts our table last in the file; `ExistingClaudeJson()` always has an `mcpServers` that already holds another server. So the cases that actually go wrong are all untested: our table in the *middle* with another table after it (the case where the user's comment above that next table is eaten and the trailing blank lines are trimmed), a CRLF file, a file with no trailing newline, a file with a BOM, a `mcpServers` where housecarl is the only entry, an absent `mcpServers` key, and `[mcp_servers.housecarl2]` as a sibling. The last one passes today and is worth pinning as a regression guard; the others are the findings above. Given the probe's own doc says the property is "the entry is gone AND every other byte is where it was", it should drive `RemoveTomlTable` and `UnregisterClaudeMcpServer` over a table of fixtures rather than one, which is cheap now that both are `internal`.

Avick3110 (commented):
Blind review of PR C. Eight findings, weighted toward correctness of deletion: three on the TOML removal splice, one on the JSON one, one on the Claude record being decorative, one on the package pre-flight blocking the documented uninstall route, one on the summary reusing the plan's lines, and one on probe coverage. I checked the Claude tree removal path, the record's path handling (ReadSkillRecord's bare-name filter is sound, and CodexSkillRecord sits beside the server dir rather than inside it, so the DeleteTree that precedes the read does not eat it), the pre-flight refusal ordering (T8's property holds: destExes are checked for every host in the target before the first delete), the record write ordering on install (copy first, record after, so a crash between them leaves a tree with no record, which the removal still handles), and the probe's discovery (CiAll reflects over CiProbeAttribute, so [CiProbe("setup-uninstall")] is picked up by ci-all with no roster edit). Line endings are CRLF in every new and touched file, matching the rest of the tree. I did not run the setup exe.

PR #714, opened 2026-09-10
Colour the installer's status words by what they mean

Avick3110:
The detection block and the host menu printed every status word dim, so a missing .NET runtime read the same as a host that simply is not on the machine. A status cell is now a list of status words, each carrying its kind rather than a colour: `found` and `houseCARL <version> installed` read green (the green a finished run closes with), a host that is not there and a runtime skipped with `--skip-runtime-check` stay dim, and a runtime the install needs and cannot find reads red (the red a refusal opens with). `Ui.Row` and `Ui.MenuItem` take the words and paint them, so no caller sets a colour, and the separator and the label column are untouched. Colour stays off when the stream is redirected or `NO_COLOR` is set. The characters are unchanged: a redirected scratch run before and after the change is byte-identical.

### DIFF
--- a/src/housecarl-setup/Program.cs
+++ b/src/housecarl-setup/Program.cs
@@ -218,16 +218,26 @@ public static class Program
         string aspName  = "ASP.NET Core Runtime " + ServerRuntimeMajor;
         Ui.Row(baseName, RuntimeStatus(skipped, missingRuntimes.Contains("Microsoft.NETCore.App")));
         Ui.Row(aspName,  RuntimeStatus(skipped, missingRuntimes.Contains("Microsoft.AspNetCore.App")));
-        Ui.Row(claude.Name, claude.Summary);
-        Ui.Row(codex.Name,  codex.Summary);
+        Ui.Row(claude.Name, HostStatus(claude));
+        Ui.Row(codex.Name,  HostStatus(codex));
     }
 
-    /// <summary>A runtime row's status: a runtime the install needs and cannot find is the one thing in this
-    /// block that stops it, so it reads as a problem; one that was not looked at reads plain.</summary>
-    private static Ui.StatusWord RuntimeStatus(bool skipped, bool missing)
-        => skipped ? Ui.StatusWord.Plain("not checked (--skip-runtime-check)")
-         : missing ? Ui.StatusWord.Missing("not found")
-         :           Ui.StatusWord.Good("found");
+    /// <summary>A runtime row's status. A missing runtime does not read as a problem here: this block prints
+    /// before the menu, and only an install needs the runtimes, so whether a missing one stops the run is not
+    /// known yet - the refusal after the choice is what says so.</summary>
+    private static Ui.StatusWord[] RuntimeStatus(bool skipped, bool missing)
+        => skipped ? [Ui.StatusWord.Plain("not checked (--skip-runtime-check)")]
+         : missing ? [Ui.StatusWord.Plain("not found")]
+         :           [Ui.StatusWord.Good("found")];
+
+    /// <summary>A host row's status: the host being here, and what houseCARL is to it. A host that is not here
+    /// is not a problem - either mode has the other host to work on - so it reads plain.</summary>
+    private static Ui.StatusWord[] HostStatus(Detect.HostState host)
+        => !host.Present    ? [Ui.StatusWord.Plain("not found")]
+         : !host.Installed  ? [Ui.StatusWord.Good("found"), Ui.StatusWord.Plain("houseCARL not installed")]
+         : host.InstalledVersion is null
+                            ? [Ui.StatusWord.Good("found"), Ui.StatusWord.Good("houseCARL installed")]
+         :                     [Ui.StatusWord.Good("found"), Ui.StatusWord.Good("houseCARL " + host.InstalledVersion + " installed")];
 
     /// <summary>What came back from a question setup had to ask a person.</summary>
     private enum Answer
@@ -586,8 +596,8 @@ public static class Program
         if (mode == Mode.Install)
         {
             Ui.Heading("Install houseCARL for which agent?");
-            Ui.MenuItem("1", claude.Name, claude.Summary);
-            Ui.MenuItem("2", codex.Name,  codex.Summary);
+            Ui.MenuItem("1", claude.Name, HostStatus(claude));
+            Ui.MenuItem("2", codex.Name,  HostStatus(codex));
             Ui.MenuItem("3", "Both");
             Ui.MenuItem("4", "Uninstall", Ui.StatusWord.Plain("remove houseCARL instead"));
             Console.WriteLine();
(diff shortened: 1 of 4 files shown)


Avick3110 (commented):
Blind review of the colouring change. Checked: colour lands only on the status cell (label column, `[N]` key and the padding are still written with an uncoloured `Console.Write`, the separator stays dim); `Paint` still short-circuits on `OutColor`, which is `!NO_COLOR && !Console.IsOutputRedirected`, so the new writer inherits the same rule; every `Ui.Row` / `Ui.MenuItem` caller in the tree (`git grep` finds exactly the six in `Program.cs`) type-checks under `params StatusWord[]`, and `[3] Both` / `[4] Uninstall` keep their old behaviour (`status.Length == 0` reproduces the old `note.Length == 0` early return, and `Ui.StatusWord.Plain("remove houseCARL instead")` reproduces the old note); `ForegroundColor` is restored in `Paint`'s `finally`, so a throw part-way through a multi-word cell still unwinds the colour; `SetupSkillPruneProbe`, `SetupUninstallProbe`, `SetupUpdateLockProbe` and `DescriptionVocabularyGuardProbe` are untouched by the diff and none of them reads `HostState.Summary` or asserts these strings (the vocabulary guard scans `housecarl-setup` literals for consent phrases, and splitting `"found  ·  houseCARL not installed"` into two words plus a `StatusSeparator` const introduces none); the byte-identity claim holds by inspection — `od -c` shows the old inline separator and the new `StatusSeparator` const are the same five bytes (`sp sp 0xC2 0xB7 sp sp`), and each old cell string reassembles word-for-word; all three files are still pure-LF in the blob, matching `main` and the targeted `.gitattributes` policy. Three findings below.

src/housecarl-setup/Program.cs
```diff
@@ -216,13 +216,19 @@ private static void ReportDetected(
...
+        Ui.Row(aspName,  RuntimeStatus(skipped, missingRuntimes.Contains("Microsoft.AspNetCore.App")));
         Ui.Row(claude.Name, claude.Summary);
         Ui.Row(codex.Name,  codex.Summary);
     }
 
+    /// <summary>A runtime row's status: a runtime the install needs and cannot find is the one thing in this
+    /// block that stops it, so it reads as a problem; one that was not looked at reads plain.</summary>
+    private static Ui.StatusWord RuntimeStatus(bool skipped, bool missing)
+        => skipped ? Ui.StatusWord.Plain("not checked (--skip-runtime-check)")
+         : missing ? Ui.StatusWord.Missing("not found")
```
Avick3110: The red is decided before the run knows whether it is an install. `ReportDetected` is called at line 125, and the mode only comes back from `ResolveChoice` at line 127; the runtime refusal is deliberately gated on `mode == Mode.Install` at line 138, with a comment saying why — a machine missing a runtime is exactly a machine somebody may want to remove houseCARL from, and refusing before the menu would leave no way to reach `[4] Uninstall`. `RuntimeStatus` cannot see that distinction, so on an uninstall run those two rows still read red "not found". The docstring you have written above it asserts the opposite of what the code can know: "a runtime the install needs and cannot find is the one thing in this block that stops it" — on the uninstall path it stops nothing, and the colour now says it does, which walks back part of what the earlier post-menu-refusal change bought. Note the flag-driven case is decidable at that point (`--uninstall` is in `args` before `ReportDetected` runs) even though the menu case is not, so this is either a narrower rule than the comment claims or a case for keeping the runtime rows plain in the detection block and letting the refusal carry the red, as it already does.

src/housecarl-setup/Ui.cs
```diff
@@ -62,20 +91,41 @@ public static void Heading(string text)
...
+            Paint(OutColor, StatusColor(word.Kind), () => Console.Write(word.Text));
+        }
+        Console.WriteLine();
+    }
+
+    // Green is the same green a finished run closes with, red the same red a refusal opens with.
+    private static ConsoleColor StatusColor(Status kind) => kind switch
+    {
+        Status.Good => ConsoleColor.Green,
+        Status.Missing => ConsoleColor.Red,
```
Avick3110: No `plugin/CHANGELOG.md` line under `## Unreleased`. This is a change a user sees the first time they run setup — the detection block and the host menu now print in three colours where they printed one — and the repo's rule is a line there if a user would notice the change. The existing installer entry's colour sentence ("Output is coloured; when it is not, and why, is on `Ui` in `src/housecarl-setup/Ui.cs`") still holds and does not need editing: it points at `Ui` rather than restating the rule, and the rule itself (`OutColor`) is unchanged, so it is a pointer that has not drifted. What is missing is the new fact, not a correction to the old one.

src/housecarl-setup/Ui.cs
```diff
@@ -62,20 +91,41 @@ public static void Heading(string text)
     }
 
     /// <summary>One thing that was looked for, and what was found - the rows of the detection block.</summary>
-    public static void Row(string label, string value)
+    public static void Row(string label, params StatusWord[] status)
```
Avick3110: Two smaller things about where the new type sits. `params StatusWord[]` gives `Row` an empty-argument form it has no meaning for: `Ui.Row("label")` now compiles and prints a padded label followed by trailing spaces and a newline, where the old two-argument signature would not compile. `MenuItem` genuinely wants that form (`[3] Both`), `Row` does not, and the arity is now the only thing standing between a forgotten status cell and a silently blank one. Separately, `Detect.HostState.Summary` now returns `Ui.StatusWord[]`, which puts the choice of how a detection result reads — and therefore its colour — inside the detection file; `Ui.cs`'s own header says presentation lives there so that it is one file to read and one file to change, and this is the first thing to cross that line. Both are judgement calls rather than defects, but the second is the one worth a ruling now, before a second caller follows the precedent.

src/housecarl-setup/Program.cs
Avick3110: Folded, taking the second option: the runtime rows are plain in the detection block and the post-menu refusal keeps the red. The flag-driven case being decidable there is real but would make the colour of a row depend on how the run was started, which is a worse row to read than a uniformly plain one. Text is unchanged either way - colour is a console attribute, never characters, so a redirected run was already byte-identical. Status.Missing went with it, since nothing created one any more.

src/housecarl-setup/Ui.cs
Avick3110: Folded - a line is now under ## Unreleased saying which words read green and which dim, that nothing in the block is red because a refusal says that itself, and pointing at Ui in src/housecarl-setup/Ui.cs for when output is not coloured. The existing installer entry is left as it was.

src/housecarl-setup/Ui.cs
Avick3110: Both folded. Row now takes StatusWord[] rather than params, so a row written without a status cell does not compile; MenuItem keeps params, which it has a use for. And Detect.HostState.Summary is gone: HostStatus in Program.cs builds the host row's words beside RuntimeStatus, so Detect only detects, Ui only paints, and the two callers still share one wording.

Issue #727, opened 2026-09-11
[GAP] the reverse lane costs ~50 ms per match after the index; an unbounded references= on a common target is a two-minute call


Avick3110:
Measured 2026-09-11 on a 3,801-plugin order (8C/16T, 32 GB), the installed 2.0.0 server in its own process.

The reverse lane is time-bound, not memory-bound, and the cost is **per match after the index**, not the index build:

| call | elapsed | peak WS | matches | notes |
|---|---:|---:|---:|---|
| `references=` ArmorHeavy, unbounded, first call | 54.6 s | — | 3,635 | reverse index built in 26.6 s: 14.7 M pairs, ~175 MB held |
| `references=` IronIngot, unbounded, warm | **109.6 s** | 1.8 GB | 2,251 | index already built; longer than the cold call with more matches |
| reverse `walk=` `*` from IronIngot, defaults (2,000) | 114 s | 1.6 GB | 2,000 (cut) | **57 ms per node** |
| reverse `walk=` from IronIngot, depth 100 / 1e6 | 900 s, aborted | 1.8 GB | — | |
| reverse `walk=` from PlayerRef, depth 100 / 1e6 | 600 s, aborted | 1.6 GB | — | |

The index itself is cheap (27 s once, 175 MB). What costs is the per-match body read after it — the same ~33 ms-per-body floor the tree form pays (filed separately), plus whatever the reverse lane does per hit on top. A warm unbounded `references=` at 50 ms per referencer means any common target (a vanilla ingot, keyword, or race) is a two-minute call with no estimate up front.

The memory side is fine: 1.6–1.8 GB for the unbounded arms, no growth.

Full write-up in the memory-battery note (Claude Output, 2026-09-11).

PR #731, opened 2026-09-12
Release a walked node's body when its pass ends

### LINKED ISSUE #719: [Bug] a forward walk holds ~230 KB per reached node with no upper bound on max_nodes, and runs to OOM at 56 GB
Measured 2026-09-11 on a 3,801-plugin order (8C/16T, 32 GB), the installed 2.0.0 server driven over stdio in its own process, working set sampled every 250 ms from outside.

A forward `walk=` with `follow:"*"` holds roughly **230 KB per reached node for the lifetime of the call**, and nothing on the surface bounds `walk.depth` or `walk.max_nodes` above `>= 1`. One `housecarl_records` call from a single NPC:

| max_nodes | elapsed | peak working set | peak private | result |
|---:|---:|---:|---:|---|
| 2,000 (default) | 5–9 s | 1.7 GB | 1.6 GB | cap enforced, said so |
| 10,000 | 21 s | 3.4 GB | 3.3 GB | reached, depth 5 |
| 40,000 | 60 s | 9.5 GB | 9.5 GB | reached, depth 7 |
| 1,000,000 | 471 s | 21.7 GB (RAM-capped, rest paged) | **55.9 GB** | **`OutOfMemoryException`**; 20.6 GB still held 30 s after the call returned |

Growth is linear at ~110 MB/s. It is a live set, not GC slack: workstation GC and `GCConserveMemory=5` leave the 10k-node peak at 3.4 GB. The closure from any NPC on this order is hundreds of thousands of nodes at depth 7, so any raised budget reaches it.

This is the 44 GB shape from the first real-world report — the merge path was measured clean (329 donors, 1.4 GB) and cyclic references were falsified (a constructed A↔B LVLI cycle walks in 0.06 s). What does the damage is closure breadth times per-node retention.

Two things to decide, with the numbers in hand:

1. A hard upper bound on `max_nodes`, or a memory-derived one, stated in the parameter's own description and in the refusal.
2. What the per-node ~230 KB is. A reached node should cost an identity and its links, not a materialised body held until the call ends; and the reached set should be released when the render is done, which it is not (20.6 GB held after the OOM, 5.3 GB held after the crash in the linked issue).

Related: the walk also dies raw on the order's one malformed record before it gets this far unless that record is excluded — filed separately.

Avick3110:
A forward `walk=` cached one record body per reached node for the lifetime of the call. A Mutagen record getter is a `ReadOnlyMemorySlice<byte>` over the whole GRUP it was read from, and the slice keeps that array alive, so the cache pinned one whole-group byte array per source group per plugin the walk touched — measured at ~270 KB per reached node on Aaron's ARR 2.0 order, unreleased until the call returned, and an `OutOfMemoryException` at a raised `walk.max_nodes`. The walk now works both its seed set and each hop in passes of the gather's own chunk size and releases the pass's bodies when the pass ends, so a reached node costs its row — identity, links, pull chain — and nothing is held into the render. What a node yielded is remembered as values in a per-call memo, so a key two seeds both reach is still one read per call, which is what the body cache used to buy before it was the retention. The NPC template report walks those values too, rather than keeping a getter per seed. Plugins are enumerated as often as before, because the pass size is the chunk size the gather already split on.

## Where the bytes were

Profiled with `dotnet-gcdump` against a private Release build driven over stdio, forward `*` chain from Lydia (`0A2C8E:Skyrim.esm`) at `max_nodes: 10000`, `Perk` excluded with severity `stop`:

- At peak, the **live** managed heap (gcdump induces a gen2 GC, so this is live data, not slack) was **2.67 GB** against a 3.53 GB working set. The order index accounts for 340 MB of it; the rest is `System.Byte[]` — whole-GRUP arrays, five of them averaging 35 MB each, pinned through the `Dictionary<FormKey, IMajorRecordGetter?>` the gcdump shows by name.
- Immediately after the call, the live heap is **383 MB** (the index alone) while the working set stays at 3.55 GB. So the "20.6 GB still held 30 s later" in the issue is the GC not returning freed segments to the OS, not a second leak — the retention that matters is the in-call one, and that is what this fixes.

**Per reached node: ~270 KB before** ((3,528 − 830 MB idle) / 10,000), **~4.4 KB after** ((1,694 − 830 MB) / 195,848 on the million-node arm) — and what it now buys is the node row, not a body.

## Before and after

Private `housecarl-mcp.exe` — this branch's Release build against `main`'s, both built here, neither the installed server — one cold process per arm, working set and private bytes sampled every 250 ms from outside, max inside the call's window. 3,801-plugin ARR 2.0 order, 8C/16T, 32 GB. Arms 1–3: seed Lydia, forward `*`, `project.form='chain'`, `Perk` excluded severity `stop`. Baselines at 40,000 and 1,000,000 are the issue's; 10,000 was re-run here for parity and matched (3,528 MB against the issue's 3,415 MB).

| arm | peak WS before | peak WS after | WS 35 s after the call returned | elapsed | result |
|---|---:|---:|---:|---:|---|
| `max_nodes` 10,000 | 3,528 MB | **1,430 MB** | 3,549 → **1,140 MB** | 35.9 → 34.6 s | 10,000 reached, depth 5, unchanged |
| `max_nodes` 40,000 | 9,469 MB | **1,490 MB** | — → **1,198 MB** | 60.3 → 74.4 s | 40,000 reached, depth 7, unchanged |
| `max_nodes` 1,000,000 | **OOM at 55,875 MB private** after 471 s, 20.6 GB held 30 s later | **1,694 MB** | — → **1,347 MB** | 471 s (died) → 287 s (finished) | **now completes**: 195,848 nodes, the whole closure from that seed at depth 16 |
| **multi-seed**: 2,127 NPC_ defined in `BSHeartland.esm`, depth 2, `max_nodes` 50 | 1,753 MB | **1,506 MB** | 1,786 → **1,294 MB** | 24.8 → 26.4 s | 100,402 reached, byte-identical response |

The multi-seed arm is the one the review asked for, and it earned its place: with the per-pass release alone it ran 24.8 s → **34.3 s**, because releasing bodies per pass had also dropped the per-call read-once guarantee. The memo restores it at 26.4 s, within noise of the baseline, and the memory win survives. The 40,000 arm is ~14 s slower against the issue's own baseline; the 10,000 arm came out marginally faster, so those two straddle run-to-run noise on a different build and I did not chase it further.

The 1,000,000 arm is the striking one: the call that used to consume 56 GB and die now walks the seed's entire closure to exhaustion inside 1.7 GB. It is also the arm the new bound refuses (see below) — it was run before the bound landed, so I re-ran it at the bound on the same harness to check the bound does not cost an answer:

| arm | peak WS | WS 35 s after | elapsed | result |
|---|---:|---:|---:|---|
| `max_nodes` **250,000** (the bound) | 1,808 MB | 1,679 MB | 308 s | 195,848 nodes — the same whole closure |
| `max_nodes` **250,001** | 56 MB | 58 MB | **0.3 s** | refused, naming the bound |

The refusal costs 0.3 s and 56 MB because it lands before the load order resolves at all.

## Test

Eight tests in `RecordsWalkCostTests.cs`, on two new internal counters (`LoadOrderService.WalkBodyHighWater`, `WalkBodiesHeldAtReturn`) in the same idiom as the existing cost tests' `LoadOrderResolver.BodySeeks` and `BodyPrefetch.KeysWanted` — whether a walk released a node's body or held the whole reached set is invisible in the answer and visible only in memory, which is the argument those counters were added on:

- `AWalkHoldsNoReachedBodiesPastTheGatherThatReadThem` — nothing held at return, never more than one pass at once. Fails on the old code, where held-at-return was the whole reached set.
- `AWalkSplitAcrossPassesReachesTheSameSetAndHoldsOnePass` at pass sizes 1, 7 and 64 — the answer is byte-identical to the unsplit walk, and the high water follows the pass size. `LoadOrderService.WalkPassRows` is the knob, moved by the test the way `RenderBudget.MaxRenderRows` already is; building a world with a 2,000-link hop to meet the real pass size is not a test.
- `ASeedsNodeBudgetIsSpentAcrossPassesNotPerPass` — `max_nodes` is a budget for the walk, not for each pass of it.
- `ARefusingWalkSplitAcrossPassesRefusesTheSameWayAndHoldsNothing` — the `severity: 'refuse'` sentence still names the same seed at the same hop with a hop interleaved across passes, and that return releases its pass.

- `ATemplateReportReadsNoChainNodeTheWalkAlreadyRead` — on a template walk the reached nodes *are* the chain nodes, so the report takes its facts off the bodies the walk read rather than re-reading each one with a whole-plugin seek. `AWalkGathersItsSeedBodiesOncePerPluginNotOncePerSeed` keeps its ceiling of 1 seek because of it.
- `ATransitiveReverseWalkIsNotHeldToTheForwardCeiling` — the node-budget bound belongs to the per-seed lanes, and the shared-budget lane is not caught by it.

Full suite 1,936 passed, `ci-all` 128/128.

## The ceiling — set, on Aaron's ruling

`walk.max_nodes` now has a **hard upper bound of 250,000, fixed, not memory-derived** (`RecordsTools.RecordsWalk.Ceiling`). Two measurements decided the number:

1. **The closure is finite and smaller than the budget people reach for.** The full forward closure from an NPC on this order is 195,848 nodes at depth 16. `max_nodes: 1000000` was never asking for a million nodes; it was asking for "no cap", and the cap is what turned that into an OOM rather than an answer. So the bound sits above what a single seed actually reaches: it bounds the budget, not an answer, and a walk that finishes still finishes.
2. **A reached node costs ~4.4 KB of retained row**, stably, because it is a row and no longer a body. At the bound that is ~1 GB of node rows, keeping the whole call inside ~2.5 GB on this machine including the ~830 MB order index.

Fixed rather than memory-derived because the per-node cost is now a stable row, so a fixed number is honest arithmetic rather than a guess, and because a memory-derived bound would make the same call succeed on one machine and be refused on another — the opposite of a declared cost. The downstream cost, a reading form that reads a body per rendered row, is still bounded by `RenderBudget`, which is where machine-sized questions belong.

**The bound is on the PER-SEED reading of the budget** — the forward walk and the reverse carrier walk. On the transitive reverse walk the same parameter is one budget shared across every seed and every hop: a different quantity, on the lane where this retention never existed and whose cost is a clock rather than a live set, and it is not what was ruled. That lane is left exactly as it was; `ATransitiveReverseWalkIsNotHeldToTheForwardCeiling` pins it.

What landed: the gate refuses above the bound in one sentence naming it and what to pass — `walk.max_nodes=250001 — the node budget's hard upper bound is 250000; pass that or less, which is above the whole closure a single seed reaches on a large order.` — the parameter's description carries the bound inside the clause that already says the budget is per seed on those two lanes, and `ANodeBudgetPastTheCeilingIsRefusedNamingTheBound` / `ANodeBudgetAtTheCeilingWalks` test either side of it. `walk.depth` is untouched. The tool-surface md5 moves, as expected for a description change.

**Nothing under `.claude/skills/` teaches `max_nodes`** — grepped the whole tree, and the only non-code mentions are this changelog's own history — so there was no skill or reference text to update.

**The time figure is deliberately not in the surface text.** 195,848 nodes took 287 s, about 1.5 ms a node, so above roughly 200,000 the clock refuses before the memory does. That did not fit the description's shape in one clause without becoming a second sentence about a second cost, and it is not what the bound is measuring, so the description says the bound and the refusal says what to pass. Worth its own look if a walk's declared cost ever states seconds.

Closes #719

### DIFF
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -15,14 +15,14 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 - **A forward `walk=` no longer holds every record it reached.** A record body read from a plugin is a slice
   of that record group's whole byte array and keeps it alive, so caching one body per reached node until the
-  call returned pinned one array per source group per plugin: 230 KB per reached node, and a raised
+  call returned pinned one array per source group per plugin: 270 KB per reached node, and a raised
   `walk.max_nodes` ran the process out of memory. A reached node now costs its row — its identity, its links,
   its pull chain — and a body lives only for the gather pass that read it, so the reached set is gone before
   anything renders. Measured on a 3,801-plugin order, the forward closure from one NPC: at
   `walk.max_nodes: 10000`, 3,528 MB peak working set becomes 1,480 MB; at 40,000, 9,469 MB becomes 1,516 MB;
   at 1,000,000 — which used to die with `OutOfMemoryException` at 55,875 MB private after 471 s — the walk
-  now finishes, reaching the closure's full 195,848 nodes in 289 s at 1,656 MB. What the walk reports is
-  unchanged. The bound on `walk.max_nodes` itself is unchanged, and is the open half of this: see #719.
+  now finishes, reaching the closure's full 195,848 nodes in 287 s at 1,694 MB. What the walk reports is
+  unchanged, and so is what `walk.max_nodes` accepts: the parameter still takes any budget at or above 1.
 
 ## 2.0.0 — 2026-09-11
 
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -19,16 +19,18 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `walk.max_nodes` ran the process out of memory. A reached node now costs its row — its identity, its links,
   its pull chain — and a body lives only for the gather pass that read it, so the reached set is gone before
   anything renders. Measured on a 3,801-plugin order, the forward closure from one NPC: at
-  `walk.max_nodes: 10000`, 3,528 MB peak working set becomes 1,480 MB; at 40,000, 9,469 MB becomes 1,516 MB;
-  at 1,000,000 — which used to die with `OutOfMemoryException` at 55,875 MB private after 471 s — the walk
-  now finishes, reaching the closure's full 195,848 nodes in 287 s at 1,694 MB. What the walk reports is
-  unchanged.
+  `walk.max_nodes: 10000`, 3,528 MB peak working set becomes 1,430 MB; at 40,000, 9,469 MB becomes 1,490 MB;
+  and at a budget big enough to exhaust the closure — which used to die with `OutOfMemoryException` at
+  55,875 MB private after 471 s — the walk now finishes, reaching that seed's full 195,848 nodes in 308 s at
+  1,808 MB. What the walk reports is unchanged.
 - **`walk.max_nodes` now has a hard upper bound of 250,000, and a higher budget is refused naming it.** Fixed,
   not derived from the machine: a reached node costs a row of a few KB wherever the walk runs, so the arithmetic
   is the same on every box, and a bound that moved with the machine would let one call succeed on one and be
   refused on another. It bounds the budget without bounding an answer — the whole forward closure a single seed
   reaches, measured on a 3,801-plugin order, is 195,848 nodes, so a walk that finishes still finishes. The bound
-  is on `walk.max_nodes`; `walk.depth` is unchanged.
+  is on the PER-SEED reading of the budget, which the parameter's own description separates: the forward walk and
+  the reverse carrier walk. The transitive reverse walk, where one budget is shared across every seed and every
+  hop, is unchanged, and so is `walk.depth`.
 
 ## 2.0.0 — 2026-09-11
 
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -13,6 +13,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A forward `walk=` no longer holds every record it reached.** A record body read from a plugin is a slice
+  of that record group's whole byte array and keeps it alive, so caching one body per reached node until the
+  call returned pinned one array per source group per plugin: 230 KB per reached node, and a raised
+  `walk.max_nodes` ran the process out of memory. A reached node now costs its row — its identity, its links,
+  its pull chain — and a body lives only for the gather pass that read it, so the reached set is gone before
+  anything renders. Measured on a 3,801-plugin order, the forward closure from one NPC: at
+  `walk.max_nodes: 10000`, 3,528 MB peak working set becomes 1,480 MB; at 40,000, 9,469 MB becomes 1,516 MB;
+  at 1,000,000 — which used to die with `OutOfMemoryException` at 55,875 MB private after 471 s — the walk
+  now finishes, reaching the closure's full 195,848 nodes in 289 s at 1,656 MB. What the walk reports is
+  unchanged. The bound on `walk.max_nodes` itself is unchanged, and is the open half of this: see #719.
+
 ## 2.0.0 — 2026-09-11
 
 houseCARL 2.0.0 replaces the 1.x tool surface with 31 tools. The record plane is one grammar: a read is one call composed from four axes (SELECT × SOURCE × PROJECT × TRANSPORT); a write is one call composed from an op list, a lane and a transport; one record is a set of one. Record coverage is generated from Mutagen.Bethesda.Skyrim 0.54.4 at build time, 1,174 types, and the write pre-flight and the `mutagen-reference` skill are two renderings of that one artifact. The 1.x tool and parameter names are deleted, not deprecated: a retired tool name is refused with a refusal naming its successor, from `AliasTable.cs`; a retired parameter name is refused as an unknown parameter, with the parameters the tool does take. Seven skills ship. `housecarl_check` gains the facegen family. The installer shows what it will write before writing, and uninstalls. The entries below are in the order they landed.
--- a/plugin/CHANGELOG.md
+++ b/plugin/CHANGELOG.md
@@ -22,7 +22,13 @@ saying it sets an expectation their install may contradict. Say what is known, a
   `walk.max_nodes: 10000`, 3,528 MB peak working set becomes 1,480 MB; at 40,000, 9,469 MB becomes 1,516 MB;
   at 1,000,000 — which used to die with `OutOfMemoryException` at 55,875 MB private after 471 s — the walk
   now finishes, reaching the closure's full 195,848 nodes in 287 s at 1,694 MB. What the walk reports is
-  unchanged, and so is what `walk.max_nodes` accepts: the parameter still takes any budget at or above 1.
+  unchanged.
+- **`walk.max_nodes` now has a hard upper bound of 250,000, and a higher budget is refused naming it.** Fixed,
+  not derived from the machine: a reached node costs a row of a few KB wherever the walk runs, so the arithmetic
+  is the same on every box, and a bound that moved with the machine would let one call succeed on one and be
+  refused on another. It bounds the budget without bounding an answer — the whole forward closure a single seed
+  reaches, measured on a 3,801-plugin order, is 195,848 nodes, so a walk that finishes still finishes. The bound
+  is on `walk.max_nodes`; `walk.depth` is unchanged.
 
 ## 2.0.0 — 2026-09-11
 
--- a/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
+++ b/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
@@ -284,6 +284,34 @@ public sealed class RecordsWalkCostTests
         Assert.Equal(0, LoadOrderService.WalkBodiesHeldAtReturn);
     }
 
+    // ---- the node budget's hard upper bound ---------------------------------------------------------
+
+    /// <summary>The budget has a hard upper bound, so "no cap" cannot be spelled as a huge number and walked into
+    /// the ground. The refusal names the bound and what to pass instead.</summary>
+    [Fact]
+    public void ANodeBudgetPastTheCeilingIsRefusedNamingTheBound()
+    {
+        var response = RecordsTools.Records(Svc, types: Npc, plugins: Scope(),
+                                            walk: new RecordsTools.RecordsWalk { max_nodes = RecordsTools.RecordsWalk.Ceiling + 1 },
+                                            project: Chain(), counts_only: true);
+
+        Assert.StartsWith("error:", response);
+        Assert.Contains("walk.max_nodes=" + (RecordsTools.RecordsWalk.Ceiling + 1), response);
+        Assert.Contains(RecordsTools.RecordsWalk.Ceiling.ToString(), response);
+    }
+
+    /// <summary>And the bound itself is a legal budget, not one off it — the refusal is above, not at.</summary>
+    [Fact]
+    public void ANodeBudgetAtTheCeilingWalks()
+    {
+        var response = RecordsTools.Records(Svc, types: Npc, plugins: Scope(),
+                                            walk: new RecordsTools.RecordsWalk { depth = 1, max_nodes = RecordsTools.RecordsWalk.Ceiling },
+                                            project: Chain(), counts_only: true);
+
+        Assert.DoesNotContain("error:", response);
+        Assert.Contains($"reached={WalkCostWorld.Seeds * (WalkCostWorld.ItemsPerSeed + 1) + 1}", response);
+    }
+
     /// <summary>A seed at its node budget reads no further. The gather used to take every seed's whole hop frontier
     /// before the budget was consulted on dequeue, so a walk capped at one node still read — and pinned — every
     /// link off every seed to keep one of them. The budget now rides into the gather, and the cap itself is
--- a/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
+++ b/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
@@ -164,10 +164,7 @@ public sealed class RecordsWalkCostTests
         var seeks = LoadOrderResolver.BodySeeks - before;
 
         Assert.Contains($"seeds={WalkCostWorld.Seeds + 2}", response);
-        // The gather, plus the template report's own re-read of the chain, which the walk no longer pins (#719).
-        // That re-read is the CHAIN's length, not the seed count — the reports share one cache — so the claim this
-        // test makes is unchanged: a constant, not one walk per seed.
-        Assert.True(seeks <= 3, $"{WalkCostWorld.Seeds + 2} walk seeds cost {seeks} per-record plugin walks.");
+        Assert.True(seeks <= 1, $"{WalkCostWorld.Seeds + 2} walk seeds cost {seeks} per-record plugin walks.");
     }
 
     /// <summary>Every seed advances one hop together, so a hop's reached nodes are one gather too. Before, a closure
@@ -298,6 +295,39 @@ public sealed class RecordsWalkCostTests
         Assert.StartsWith("error:", response);
         Assert.Contains("walk.max_nodes=" + (RecordsTools.RecordsWalk.Ceiling + 1), response);
         Assert.Contains(RecordsTools.RecordsWalk.Ceiling.ToString(), response);
+    }
+
+    /// <summary>The template report costs no read of its own: on a template walk the reached nodes ARE the chain
+    /// nodes, so the report takes its facts off the bodies the walk already read. It re-read every chain node from
+    /// disk once the walk stopped pinning bodies, which is the seed-count-times-whole-plugin-seek shape of
+    /// #556.</summary>
+    [Fact]
+    public void ATemplateReportReadsNoChainNodeTheWalkAlreadyRead()
+    {
+        var before = LoadOrderResolver.BodySeeks;
+        var response = RecordsTools.Records(Svc, types: Npc, plugins: Scope(), walk: TemplateWalk(),
+                                            project: Chain(), counts_only: true);
+        var seeks = LoadOrderResolver.BodySeeks - before;
+
+        Assert.DoesNotContain("error:", response);
+        Assert.True(seeks <= 1, $"the template report over {WalkCostWorld.Seeds + 2} seeds cost {seeks} per-record plugin walks.");
+    }
+
+    /// <summary>The bound is on the PER-SEED reading of the budget. On the transitive reverse walk the same
+    /// parameter is one budget shared across every seed and every hop — a different quantity, on the lane where the
+    /// retention this PR fixes never existed — so this gate leaves it alone.</summary>
+    [Fact]
+    public void ATransitiveReverseWalkIsNotHeldToTheForwardCeiling()
+    {
+        var response = RecordsTools.Records(Svc, formids: new[] { _w.RevisitSeed },
+                                            walk: new RecordsTools.RecordsWalk
+                                            {
+                                                direction = "reverse", follow = "*", depth = 1,
+                                                max_nodes = RecordsTools.RecordsWalk.Ceiling + 1,
+                                            },
+                                            project: Fields(), counts_only: true);
+
+        Assert.DoesNotContain("hard upper bound", response);
     }
 
     /// <summary>And the bound itself is a legal budget, not one off it — the refusal is above, not at.</summary>
--- a/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
+++ b/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
@@ -164,7 +164,10 @@ public sealed class RecordsWalkCostTests
         var seeks = LoadOrderResolver.BodySeeks - before;
 
         Assert.Contains($"seeds={WalkCostWorld.Seeds + 2}", response);
-        Assert.True(seeks <= 1, $"{WalkCostWorld.Seeds + 2} walk seeds cost {seeks} per-record plugin walks.");
+        // The gather, plus the template report's own re-read of the chain, which the walk no longer pins (#719).
+        // That re-read is the CHAIN's length, not the seed count — the reports share one cache — so the claim this
+        // test makes is unchanged: a constant, not one walk per seed.
+        Assert.True(seeks <= 3, $"{WalkCostWorld.Seeds + 2} walk seeds cost {seeks} per-record plugin walks.");
     }
 
     /// <summary>Every seed advances one hop together, so a hop's reached nodes are one gather too. Before, a closure
@@ -204,6 +207,24 @@ public sealed class RecordsWalkCostTests
         var ceiling = perSeedCeiling * (WalkCostWorld.Seeds + 2);
         Assert.True(allocated < ceiling,
                     $"the walk allocated {allocated / 1048576} MB over {WalkCostWorld.Seeds + 2} seeds — past the {ceiling / 1048576} MB this world's seed count allows.");
+    }
+
+    /// <summary>A reached node costs its row, not its body. A record getter is a slice of its whole GRUP's byte
+    /// array and pins it, so caching every reached node's body until the call ended pinned one array per source
+    /// GRUP per plugin — 230 KB a reached node on a real order, and an OOM at a raised budget (#719). Bodies now
+    /// live for the gather pass that read them, and the reached set is gone before anything renders.</summary>
+    [Fact]
+    public void AWalkHoldsNoReachedBodiesPastTheGatherThatReadThem()
+    {
+        var response = RecordsTools.Records(Svc, types: Npc, plugins: Scope(),
+                                            walk: new RecordsTools.RecordsWalk { depth = 1 },
+                                            project: Chain(), counts_only: true);
+
+        var reached = WalkCostWorld.Seeds * (WalkCostWorld.ItemsPerSeed + 1) + 1;
+        Assert.Contains($"reached={reached}", response);
+        Assert.Equal(0, LoadOrderService.WalkBodiesHeldAtReturn);
+        Assert.True(LoadOrderService.WalkBodyHighWater <= BodyPrefetch.ChunkRows,
+                    $"the walk held {LoadOrderService.WalkBodyHighWater} bodies at once — past the {BodyPrefetch.ChunkRows} one gather pass allows.");
     }
 
     /// <summary>A seed at its node budget reads no further. The gather used to take every seed's whole hop frontier
--- a/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
+++ b/src/housecarl-mcp-tests/RecordsWalkCostTests.cs
@@ -211,7 +211,7 @@ public sealed class RecordsWalkCostTests
 
     /// <summary>A reached node costs its row, not its body. A record getter is a slice of its whole GRUP's byte
     /// array and pins it, so caching every reached node's body until the call ended pinned one array per source
-    /// GRUP per plugin — 230 KB a reached node on a real order, and an OOM at a raised budget (#719). Bodies now
+    /// GRUP per plugin — 270 KB a reached node on a real order, and an OOM at a raised budget (#719). Bodies now
     /// live for the gather pass that read them, and the reached set is gone before anything renders.</summary>
     [Fact]
     public void AWalkHoldsNoReachedBodiesPastTheGatherThatReadThem()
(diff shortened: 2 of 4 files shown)


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4138,18 +4187,24 @@ seedWin is null
...
                 if (st.Frontier.Count == 0) { st.Visited = new(); st.Frontier = new(); }
             }
+            // The pass is over: the bodies it gathered have given up their identity and their links, so they go
+            // now rather than at the end of the call. This is the release the #719 retention was missing.
+            if (bodyCache.Count > WalkBodyHighWater) WalkBodyHighWater = bodyCache.Count;
+            bodyCache.Clear();
+            }
         }
 
+        WalkBodiesHeldAtReturn = bodyCache.Count;
```
Avick3110: **The template report puts the retention straight back, and this counter is measured before it happens.**

`WalkBodiesHeldAtReturn` is read here, and the results loop below then calls `NpcTemplateReport(Fetch, ...)`. `NpcTemplateReport` walks the chain with `fetch(nextKey)` per node, `Fetch` caches into the same `bodyCache`, and nothing clears it between seeds — the only `Clear()` is after the whole loop. So by the last seed the cache holds the union of every seed's template chain, all of them live getters pinning their whole GRUP's byte array.

Scenario: `formids=` a 2,000-NPC selection, `walk.follow='Template'`, chains 5 deep. The hops release correctly, the counter reports 0, and then the results loop rebuilds a cache of up to ~10,000 bodies (less whatever the chains share) that stays live right up to the render — the same shape as #719, on the one lane that still keeps a seed getter. Clearing `bodyCache` per seed inside the loop, or reporting the high water after the loop too, would make the claim the new test asserts actually cover this path.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4138,18 +4187,24 @@ seedWin is null
                 // and the results loop reads neither of these, so the bookkeeping goes back now rather than at return.
                 if (st.Frontier.Count == 0) { st.Visited = new(); st.Frontier = new(); }
             }
+            // The pass is over: the bodies it gathered have given up their identity and their links, so they go
+            // now rather than at the end of the call. This is the release the #719 retention was missing.
+            if (bodyCache.Count > WalkBodyHighWater) WalkBodyHighWater = bodyCache.Count;
```
Avick3110, line 4291: **The seed gather is still unbounded, so the high water is `seeds.Count + ChunkRows`, not `ChunkRows`.**

`Prefetch(seedGather)` (line 4019) puts *every* seed's body into `bodyCache` before the first hop, and the first `bodyCache.Clear()` is here — the end of hop 1's first pass. Until then all of them are live, on top of that pass's own gather.

Seeds are not a small set: `WalkLane` is fed `ids` from `Artifacts.ExpandListInput(formids, ...)`, so a walk can be seeded from a spilled scan artifact with thousands of FormIDs. A 5,000-seed forward walk therefore holds 5,000 whole-GRUP-pinning getters simultaneously before it has finished one pass — the same failure mode this PR fixes, just at the seed set instead of the reached set.

It also means the new test's `WalkBodyHighWater <= BodyPrefetch.ChunkRows` assertion holds only because `WalkCostWorld.Seeds` happens to be 300; raise the test world's seed count past ~800 and it fails for a reason that is not a bug in the release logic.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3853,6 +3853,17 @@ public IReadOnlyList<TreeRow> TreeBatch(
...
+    /// <summary>The most record bodies one forward walk held at once. Counted for the reason
+    /// <see cref="LoadOrderResolver.BodySeeks"/> is: whether the walk released a reached node's body or held the
+    /// whole reached set is invisible in the answer and only the memory differs, so this is what a test can hold
+    /// that claim to (#719). Set by <see cref="WalkForwardBatch"/>, which resets it on entry.</summary>
+    internal static int WalkBodyHighWater;
+
+    /// <summary>How many record bodies the last forward walk was STILL holding when its hops finished — the other
+    /// half of the same claim, and the one the reported bug was: the reached set has to be gone before the render,
+    /// not at the end of the call. Zero on every walk.</summary>
+    internal static int WalkBodiesHeldAtReturn;
```
Avick3110, line 3865: **"Zero on every walk" is not true on two paths.**

1. If no seed has a frontier — a seed whose record carries no form links, or `seed_paths` that resolve to nothing — the hop loop breaks at `if (!pending) break;` (line 4091) before any pass runs, so no `bodyCache.Clear()` ever executes and `WalkBodiesHeldAtReturn` is set to `seeds.Count` at line 4197. The bodies really are still held, so the counter is honest and the doc comment is wrong; either way the invariant the new test asserts does not hold for that input, and no test covers it.
2. The `severity: 'refuse'` exclusion `return Array.Empty<WalkSeedResult>()` inside the pass loop returns without clearing `bodyCache` and without setting the counter, so it keeps its entry value of 0 while a whole pass's bodies are in fact live. A test reading the counter after a refusing walk would be told 0 on a walk that held a full pass.

Either reword the guarantee to "zero once the hop loop has run a pass", or clear the cache and set the counter on the early-return and no-frontier paths too.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4138,18 +4187,24 @@ seedWin is null
                 // and the results loop reads neither of these, so the bookkeeping goes back now rather than at return.
                 if (st.Frontier.Count == 0) { st.Visited = new(); st.Frontier = new(); }
             }
+            // The pass is over: the bodies it gathered have given up their identity and their links, so they go
+            // now rather than at the end of the call. This is the release the #719 retention was missing.
+            if (bodyCache.Count > WalkBodyHighWater) WalkBodyHighWater = bodyCache.Count;
+            bodyCache.Clear();
```
Avick3110, line 4292: **Clearing per pass drops the per-call body dedup, and the benchmark cannot see it because it has one seed.**

Before, `bodyCache` lived for the whole call, so a key's body was read at most **once per call** however many seeds and hops reached it. Now the cache dies with the pass: `gatherSeen` only dedupes *within* a pass, and `Visited` is per seed, so the same key is gathered and read again in every pass that reaches it.

Two concrete amplifications, both on the multi-seed case:

- A hop frontier larger than `BodyPrefetch.ChunkRows` (2000) splits into passes. A key linked by seed 3 and seed 900 now lands in two different passes and is read twice; on a dense selection with heavy sharing (an armour keyword, a faction, a race that hundreds of NPCs point at) that is one extra whole-plugin gather slot per pass.
- A key reached at hop *d* by one seed and hop *d+1* by another is read twice, where it used to be free.

The PR's numbers are all `seed = Lydia`, one seed, so none of the arms exercise this. The lane it regresses is exactly the one #556 was about — the 2,235-NPC selection whose per-seed reads were the original OOM. A hop-scoped cache is not the answer (a hop's frontier is unbounded, which is the retention being fixed), but the trade should be measured rather than assumed: one arm at a few thousand NPC seeds would say whether the re-reads are noise or the 14 s the 40,000 arm already shows, multiplied.

src/housecarl-mcp-tests/RecordsWalkCostTests.cs
```diff
@@ -206,6 +209,24 @@ string Walk() => RecordsTools.Records(Svc, types: Npc, plugins: Scope(), walk: T
                     $"the walk allocated {allocated / 1048576} MB over {WalkCostWorld.Seeds + 2} seeds — past the {ceiling / 1048576} MB this world's seed count allows.");
     }
 
+    /// <summary>A reached node costs its row, not its body. A record getter is a slice of its whole GRUP's byte
+    /// array and pins it, so caching every reached node's body until the call ended pinned one array per source
+    /// GRUP per plugin — 230 KB a reached node on a real order, and an OOM at a raised budget (#719). Bodies now
+    /// live for the gather pass that read them, and the reached set is gone before anything renders.</summary>
+    [Fact]
+    public void AWalkHoldsNoReachedBodiesPastTheGatherThatReadThem()
```
Avick3110, line 214: **No test ever drives a second pass, so the mechanism this PR adds is untested.**

`BodyPrefetch.ChunkRows` is 2000. This test's world gives hop 1 a frontier of `Seeds * (ItemsPerSeed + 1)` = 300 * 4 = ~1,200 keys, at `depth = 1`. The other walk tests use `follow: "Template"`, a linear chain — 300 keys a hop. Every walk in the suite therefore fits in **one** pass, and the whole new construct (`while (true)`, the `atLevel` snapshot, `take[]`, `atLevel[si] -= thisPass`, the mid-hop `bodyCache.Clear()`) runs exactly one iteration.

What that leaves unverified is the interesting half: a seed whose `take[i]` comes back 0 because an earlier seed filled `frontier` to `ChunkRows` and is skipped in that pass; `room` recomputed per pass against a seed that spent its budget in pass 1; the `atLevel[si] = 0` on the `max_nodes` break; and the seed order the `severity: 'refuse'` sentence claims ("the first seed IN SEED ORDER at the SHALLOWEST hop") now that a hop is interleaved across passes. All of them look correct on a read, but the bug class the PR is guarding against is invisible in the answer — which is the argument the counters were added on.

Cheapest fix is a second `[Fact]` that temporarily lowers a pass size (or a world with > 2000 hop-1 links) so at least one walk splits, and asserts `reached` is unchanged and `WalkBodyHighWater` is bounded.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,17 @@ saying it sets an expectation their install may contradict. Say what is known, a
...
+- **A forward `walk=` no longer holds every record it reached.** A record body read from a plugin is a slice
+  of that record group's whole byte array and keeps it alive, so caching one body per reached node until the
+  call returned pinned one array per source group per plugin: 230 KB per reached node, and a raised
+  `walk.max_nodes` ran the process out of memory. A reached node now costs its row — its identity, its links,
+  its pull chain — and a body lives only for the gather pass that read it, so the reached set is gone before
+  anything renders. Measured on a 3,801-plugin order, the forward closure from one NPC: at
+  `walk.max_nodes: 10000`, 3,528 MB peak working set becomes 1,480 MB; at 40,000, 9,469 MB becomes 1,516 MB;
+  at 1,000,000 — which used to die with `OutOfMemoryException` at 55,875 MB private after 471 s — the walk
+  now finishes, reaching the closure's full 195,848 nodes in 289 s at 1,656 MB. What the walk reports is
+  unchanged. The bound on `walk.max_nodes` itself is unchanged, and is the open half of this: see #719.
```
Avick3110: Two things a user would trip over in this entry:

- "the open half of this: see #719" — but the PR body says `Closes #719`, so the moment this merges the issue the changelog sends people to is closed. If the ceiling proposal is the open half, it wants its own issue number here (or the entry should say the bound is unchanged and stop there).
- Line 18 says **230 KB** per reached node; the PR body says **~270 KB** ((3,528 − 830) / 10,000) and the test doc comment at `RecordsWalkCostTests.cs` also says 230 KB. One of the two figures is stale — worth picking one so the shipped note and the measurement agree.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. Real, and the counter was blind to it exactly as described.

`NpcTemplateReport` no longer touches a body. It takes a `WalkTemplateFact` — editorid, template link, template flags, is-NPC, is-leveled, type name — and walks facts, memoised per call in `templateFacts`, so a shared chain is still one read per call and nothing is pinned between seeds. The chain reads go through a new `FetchTransient` that resolves without writing `bodyCache`, so the results loop cannot refill it. `WalkSeedState.SeedBody` is gone with it: a template walk over 2,000 seeds used to keep 2,000 getters alive for that report alone, which was the same bug one level up from the one you found.

`WalkBodiesHeldAtReturn` is now set immediately before `return results`, after the report has run, so it covers the path you named rather than stopping short of it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The seed gather now runs in slices of `WalkPassRows`, the same size the hops use: parse the slice, gather it, build its states and first-hop frontiers, record the high water, release. The `@file` seeding you point at is exactly why — a walk can be seeded from a spilled artifact with thousands of FormIDs, and a body per seed is a record group per seed.

You were right about the test too: with `SeedBody` gone and the seed gather sliced, the assertion no longer passes by grace of `Seeds = 300`. There are now three pass sizes (1, 7, 64) driven through `WalkPassRows`, each asserting the high water follows the pass and the answer is byte-identical to the unsplit walk.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded, on both paths, and by fixing rather than rewording.

1. The no-frontier break: the seed slices each end with a `bodyCache.Clear()`, so the cache is empty by the time the hop loop runs at all. `if (!pending) break;` now returns with nothing held, and the counter says 0 honestly.
2. The `severity: 'refuse'` return now records its high water, clears the cache and sets `WalkBodiesHeldAtReturn = 0` before returning — a refusal returns no rows to release the pass with, so it releases it itself. `ARefusingWalkSplitAcrossPassesRefusesTheSameWayAndHoldsNothing` covers it, and also pins the refusal sentence against a hop that is interleaved across passes.

The doc comment keeps the guarantee rather than softening it, because the guarantee is now true: zero on every walk, the refusal and the no-frontier returns included.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded, and you were right to ask for the measurement rather than the argument — it was real and it was large.

Measured on the arm you specified: 2,127 NPC_ defined in `BSHeartland.esm` as seeds, depth 2, `max_nodes` 50, 100,402 nodes reached, `main`'s build against this one, same harness.

| | peak WS | WS after | elapsed |
|---|---:|---:|---:|
| `main` | 1,753 MB | 1,786 MB | **24.8 s** |
| per-pass release alone | 1,437 MB | 1,272 MB | **34.3 s** |
| with the memo | 1,506 MB | 1,294 MB | **26.4 s** |

So: a 38% regression, on the lane you named.

The fix is a per-call `Dictionary<FormKey, WalkNodeFact>` holding what a node yielded — resolved, type, editorid, links — and no getter, so it pins keys and strings and never a record group. A key already answered is skipped in the gather (`Memoised`) as well as in the processing. Links are filled only when the row needs them, and a node first met at the depth cap is refilled if a later seed reaches it shallower.

It is kept only when `seeds.Count > 1`. With one seed the visited set already prevents a second read of any key — a later pass of the same hop sees it in `Visited` at gather time — so the memo would never be hit and would only cost memory on exactly the arms where memory is the point. The three single-seed arms confirm it: 1,430 / 1,490 / 1,694 MB, unchanged from the release-only build.

Aaron's wording on the issue was "a reached node should cost an identity and its links", so the memo is the ruled shape, not a hedge against it.

src/housecarl-mcp-tests/RecordsWalkCostTests.cs
Avick3110: Folded. Nothing in the suite split a pass, which made the whole construct dead weight as evidence.

`LoadOrderService.WalkPassRows` is now the pass size — `BodyPrefetch.ChunkRows` by default, moved by a test through `WithPassRows`, the same way `WithBound` already moves `RenderBudget.MaxRenderRows`. On top of it:

- `AWalkSplitAcrossPassesReachesTheSameSetAndHoldsOnePass` at pass sizes 1, 7 and 64. At 1 every key is its own pass, so `take[i] == 0` for every seed after the first in each pass, and the level snapshot, the decrement and the release all run hundreds of times. It asserts the response is byte-identical to the unsplit walk, not merely that `reached` matches.
- `ASeedsNodeBudgetIsSpentAcrossPassesNotPerPass` at pass 3 with `max_nodes = 1` — `room` recomputed per pass against a seed that spent its budget in pass 1, which is the case you singled out.
- `ARefusingWalkSplitAcrossPassesRefusesTheSameWayAndHoldsNothing` at pass 5 — the refusal sentence is compared against the unsplit walk's, so the "first seed in seed order at the shallowest hop" claim is pinned across interleaving rather than trusted.

plugin/CHANGELOG.md
Avick3110: Folded, both.

- The entry no longer points at #719. It ends "so is what `walk.max_nodes` accepts: the parameter still takes any budget at or above 1" — a statement of what shipped, with nothing for a closed issue to dangle off. The ceiling proposal stays in the PR body for Aaron to rule on, and if it becomes work it gets its own issue and its own entry.
- 270 KB everywhere now: the changelog, the engine comment and the test doc comment. 230 KB was the issue's figure from the installed 2.0.0 build; 270 KB is what I measured on my own build ((3,528 − 830) / 10,000), and since the entry quotes my numbers for everything else it should quote mine for this too.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3923,15 +3960,44 @@ public IReadOnlyList<WalkSeedResult> WalkForwardBatch(
...
+        // The same read WITHOUT the cache — for the template chain, whose nodes are read once, reduced to values,
+        // and dropped. Caching them would put the retention this walk exists to remove back on that one lane.
+        IMajorRecordGetter? FetchTransient(FormKey k)
+            => bodyCache.TryGetValue(k, out var c) ? c
+             : view.ResolveWinner(k) is { } w ? view.GetRecord(session, w.WinnerPlugin, k) : null;
+
+        // One node's template facts, memoised BY VALUE: shared chains stay one read per call (the seeds of a
+        // template walk mostly share theirs) and nothing is pinned between seeds.
+        var templateFacts = new Dictionary<FormKey, WalkTemplateFact?>();
+        WalkTemplateFact? TemplateFactOf(FormKey k)
```
Avick3110, line 3985: **Template report now pays a whole-plugin seek per chain node.**

`TemplateFactOf` reads through `FetchTransient` → `view.GetRecord` → `SeekBody`, which is a flat enumeration of the whole winning plugin per key. Before this PR the report went through `Fetch` and hit the walk's body cache: on a `follow='Template'` walk the reached nodes *are* the chain nodes, so the report cost zero extra seeks. Now every distinct chain node is re-read from disk after the walk already had its body in hand and threw it away.

Scenario: `follow='Template'` over ~2,000 NPC_ seeds with mostly distinct template chains → thousands of flat scans of Skyrim.esm at report time. That is the #556 cost shape (seed count × whole-plugin seek). The loosened `seeks <= 3` in `AWalkGathersItsSeedBodiesOncePerPluginNotOncePerSeed` only holds because `WalkCostWorld`'s seeds share one chain; the comment "the chain's length, not the seed count" is a property of that world, not of the code.

Fix is small and keeps the retention win: in `FactFor`, when `templateFollow && body is not null`, do `templateFacts.TryAdd(k, FactOf(body))` (and `templateFacts[seedFk] = SeedTemplateFact` for the seed) — the fact is a value, so nothing is pinned, and the report is back to one read per node per call. The test ceiling can then go back to 1.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -312,6 +319,8 @@ IReadOnlyList<ReadOutcome> FoldRows(IReadOnlyList<ReadOutcome> read)
             if (walk.max_nodes is { } wn)
             {
                 if (wn < 1) return Wire.Refuse(json, $"error: walk.max_nodes={wn} — the node budget must be >= 1.");
+                if (wn > RecordsWalk.Ceiling)
+                    return Wire.Refuse(json, $"error: walk.max_nodes={wn} — the node budget's hard upper bound is {RecordsWalk.Ceiling}; pass that or less, which is above the whole forward closure a single seed reaches on a large order.");
```
Avick3110: **The ceiling also caps the reverse transitive lane, with a forward-only justification.**

This gate runs before lane dispatch, so it applies to `walk=reverse` (transitive) too — where `max_nodes` is *one* budget shared across every seed and every hop (per that lane's own header text at ~L787), not per seed, and where the body-retention bug this PR fixes never existed. The refusal sentence — "above the whole forward closure a single seed reaches" — does not describe that lane.

Scenario: reverse transitive from a widely-referenced record (a KYWD, a RACE) on a large order hits the shared 250k cap; the lane's own header says "raise walk.max_nodes to walk further"; the user passes 300000 and is refused with a sentence about a single seed's forward closure. Either scope the gate to the forward/carrier lanes, or make the sentence lane-aware and confirm the reverse lane's per-node cost actually supports the same number.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,23 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A forward `walk=` no longer holds every record it reached.** A record body read from a plugin is a slice
+  of that record group's whole byte array and keeps it alive, so caching one body per reached node until the
+  call returned pinned one array per source group per plugin: 270 KB per reached node, and a raised
+  `walk.max_nodes` ran the process out of memory. A reached node now costs its row — its identity, its links,
+  its pull chain — and a body lives only for the gather pass that read it, so the reached set is gone before
+  anything renders. Measured on a 3,801-plugin order, the forward closure from one NPC: at
+  `walk.max_nodes: 10000`, 3,528 MB peak working set becomes 1,480 MB; at 40,000, 9,469 MB becomes 1,516 MB;
```
Avick3110: Numbers here disagree with the PR body's table for the same arms: 10,000 → **1,480** MB here vs **1,430** MB in the PR; 40,000 → **1,516** MB here vs **1,490** MB. One of them is the recorded measurement; pick it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded, as prescribed — `fb6c813d`.

`FactFor` now does `templateFacts.TryAdd(k, FactOf(body))` whenever `templateFollow` and the body came back, and the seed's own fact goes into the same memo at setup (`templateFacts.TryAdd(seedFk, sf)`) for the seed that sits on another seed's chain. `FetchTransient` stays, but only as the fallback for a chain node the walk did not itself reach — one past this seed's `depth` or `max_nodes` cap — so it is bounded by the caps rather than by the seed count.

`AWalkGathersItsSeedBodiesOncePerPluginNotOncePerSeed` is back to `seeks <= 1`, and the loosened comment is gone with it. You were right that the old ceiling was a property of the world, not the code, so there is now a test that says the thing directly rather than tolerating it: `ATemplateReportReadsNoChainNodeTheWalkAlreadyRead`.

plugin/CHANGELOG.md
Avick3110: Folded — `cf25f388`. The changelog now carries 1,430 MB and 1,490 MB, the re-run after the review fold, which is what the PR table quotes. 1,480/1,516 were the pre-fold build's and should not have survived the re-measure.

The same entry's third figure moved for a different reason: it cited the `max_nodes: 1000000` arm, which the shipped code now refuses. It cites the 250,000 arm instead — 195,848 nodes in 308 s at 1,808 MB, the same closure — so the entry only quotes calls a user can actually make.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded by scoping, which was the cleaner of the two options you gave — `cf25f388`. `walkDirection` and `walk.follow` are both already parsed above the gate, so the condition is one line and no dispatch moves:

```csharp
bool perSeedBudget = walkDirection == "forward"
                  || string.Equals(walk.follow?.Trim(), CarrierFollow, StringComparison.OrdinalIgnoreCase);
if (perSeedBudget && wn > RecordsWalk.Ceiling) …
```

So the bound covers the two lanes that read the budget per seed — forward and the reverse carrier walk — and the transitive reverse lane is untouched. `ATransitiveReverseWalkIsNotHeldToTheForwardCeiling` passes `max_nodes = Ceiling + 1` on `direction='reverse', follow='*'` and asserts the response carries no "hard upper bound". The word "forward" also came out of the refusal sentence, since the carrier lane is per-seed but not forward: it now reads "above the whole closure a single seed reaches on a large order". The parameter's description puts the bound inside the clause that already distinguishes the two readings, rather than in the opening parenthesis where it read as global, and the changelog entry says which lanes it binds and that the shared-budget lane is unchanged.

On whether the reverse lane's per-node cost supports the same number: it does not, and not because it is worse on memory. The battery's reverse arms are time-bound, not live-set-bound — reverse `*` from IronIngot at the 2,000 default was 114 s for 2,000 nodes, about **57 ms a node**, and unbounded runs from IronIngot and PlayerRef were **aborted at 900 s and 600 s** while holding only **1.7–1.8 GB** working set. A budget of 250,000 on that lane is roughly four hours, so a node-count bound is the wrong instrument there; what that lane wants is the per-match cost, which is filed as **#727** (reverse ~50 ms per match after the index). If a reverse bound is wanted it belongs there as its own line, with a number derived from the clock rather than from the forward closure.

The 250,000 is also untouched on its own lane: the closure measurement it came from is the forward one, and I re-ran it at the bound after this change (195,848 nodes, 308 s, 1,808 MB peak).

PR #735, opened 2026-09-14
Decode Qt INI escapes in the MO2 ini readers

### LINKED ISSUE #733: Non-ASCII MO2 profile name or game path fails: Qt \xHH escapes in ModOrganizer.ini are read literally
Reported on Nexus 2026-09-11 against 2.0.0.

Any instance whose active profile name has non-ASCII characters fails every tool with `the active profile's folder is missing: '\xe5\xa4\xa7...'`. Repro: name a profile with a CJK string (reporter's: 大肥鱼整合), select it in MO2, run any tool.

Cause: MO2 stores `selected_profile` as a QByteArray and Qt writes bytes outside printable ASCII as `\xHH` escapes: `selected_profile=@ByteArray(\xe5\xa4\xa7...)`. `CleanValue` in `src/housecarl-core/Mo2Instance.cs` strips the `@ByteArray(` wrapper and unescapes `\` only, so the `\xHH` text becomes the folder name. Same path handles `gamePath`, so a non-ASCII install path breaks the same way. `Clean` in `src/housecarl-core/Mo2ModMeta.cs` has the same gap for `meta.ini` values (mod version strings are ASCII in practice, so it is latent there).

Fix: one Qt INI unescaper shared by both readers. `\xHH` runs are greedy hex decoded as UTF-8 bytes (values above 0xFF are a UTF-16 code unit, the old Qt5 form), plus the named escapes `\ \; \, \= \n \r \t \0`, and a surrounding pair of double quotes. Test with a CJK profile name and a CJK game path.

Reporter's workaround until then: junction `profiles\ASCII` to the real folder and set `selected_profile=ASCII`.

Avick3110:
MO2 stores `selected_profile` and `gamePath` as QByteArrays, and Qt writes every non-ASCII byte as a `\xHH` escape inside the `@ByteArray(...)` wrapper. `CleanValue` in `Mo2Instance.cs` unescaped a doubled backslash only, so an instance whose active profile is named 大肥鱼整合 got the literal escape text as its folder name and every tool refused with "the active profile's folder is missing"; a non-ASCII game path failed the same way, and `Clean` in `Mo2ModMeta.cs` had the same gap for meta.ini. Both readers now go through one small unescaper, `QtIniEscapes` in housecarl-core: `\xHH` runs read greedily (a run of byte-sized values is UTF-8, a value above 0xFF is a UTF-16 code unit, the old Qt5 form), the named escapes `\\ \; \, \= \n \r \t \0`, and a surrounding pair of double quotes. An escape it does not know is left exactly as written, so a value MO2 wrote with unescaped backslashes (a plain Windows path) is never mangled. No surface text changed; the missing-folder refusal now prints the decoded name because that is what was read.

Tests: a CJK profile name and a CJK game path, encoded the way Qt writes them, resolve to the right folders through `Mo2Instance.Resolve` and `ReadSelectedProfile`, plus unit cases over the grammar — a UTF-8 run, the Qt5 code-unit form, the named escapes, plain ASCII unchanged, and an unescaped `C:\Games\Skyrim` unchanged. Full suite and `ci-all` pass.

Closes #733


src/housecarl-core/QtIniEscapes.cs
```diff
@@ -0,0 +1,96 @@
...
+        var sb = new StringBuilder(value.Length);
+        List<byte>? bytes = null;
+        int i = 0;
+        while (i < value.Length)
+        {
+            char c = value[i];
+            if (c != '\\' || i + 1 >= value.Length) { Flush(); sb.Append(c); i++; continue; }
+
+            char e = value[i + 1];
+            if (e is 'x' or 'X')
```
Avick3110, line 78: A single (unescaped) backslash is no longer passed through, so this regresses the case the header comment and the PR body both promise: "a value MO2 wrote with unescaped backslashes (a plain Windows path) is never mangled". The old `v.Replace(@"\\", @"\")` really did leave such a value alone; this loop decodes `\x`, `\n`, `\r`, `\t`, `\0`, `\;`, `\,`, `\=` unconditionally.

Concretely, for an ini not written by Qt's escaper (hand-edited, or emitted by an installer that writes the file itself):

- `gamePath=@ByteArray(D:\newgame\Skyrim SE)` -> `\n` becomes LF, giving `D:` + LF + `ewgame\Skyrim SE`.
- `gamePath=@ByteArray(E:\xEdit\Stock Game)` -> `\xEd` is read greedily as hex 0xED, one byte, invalid UTF-8 on its own, so the fallback emits U+00ED: `E:íit\Stock Game`.

Either way `Derive` then fails with "the game Data folder is missing" on an instance that resolved fine before this change. The new theory case only covers `C:\Games\Skyrim`, whose characters after the backslash happen to be inert, so it does not catch this.

Worth either gating the decode on the value actually looking Qt-escaped (e.g. only decode when no lone backslash is present), or dropping the "never mangled" claim from the comment and changelog so the behaviour is at least stated accurately.

src/housecarl-core/Mo2Instance.cs
```diff
@@ -182,7 +182,7 @@ public static (bool ok, Mo2InstancePaths? paths, IReadOnlyList<string> problems)
         const string wrap = "@ByteArray(";
         if (v.StartsWith(wrap, StringComparison.Ordinal) && v.EndsWith(")", StringComparison.Ordinal))
```
Avick3110: The quote pair is stripped inside `QtIniEscapes.Unescape`, i.e. at line 185 — after this `@ByteArray(` unwrap has already been attempted. But Qt puts the quotes OUTSIDE the wrapper: `QSettingsPrivate::iniEscapedString` sets `needsQuotes` when the value contains `;`, `,`, `=` (or has a leading/trailing space) and then inserts the quote at position 0 of the whole serialized string, `@ByteArray(` prefix included. The effect is visible in a real MO2 meta.ini on this machine: `category="-1,"`.

So a profile named `Requiem, AE` is written as

```
selected_profile="@ByteArray(Requiem, AE)"
```

the `StartsWith(wrap)` test fails, `Unescape` then strips the quotes, and `ProfileName` comes back as the literal text `@ByteArray(Requiem, AE)` — every tool refuses with "the active profile's folder is missing", which is exactly the failure mode this PR is fixing for the CJK case.

Strip the surrounding quote pair here, before the `@ByteArray(` test, rather than inside `Unescape`. Same ordering issue in `Mo2ModMeta.Clean` (line 120).

src/housecarl-core/QtIniEscapes.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    if (code <= 0xFF) (bytes ??= new List<byte>()).Add((byte)code);
+                    else { Flush(); sb.Append((char)code); }
+                    i = j;
+                    continue;
+                }
+                Flush(); sb.Append(c).Append(e); i += 2; continue;   // "\x" with no usable hex: left as written
+            }
+
+            char? named = e switch
```
Avick3110: Qt's writer emits more named escapes than this switch knows: `\a`, `\b`, `\f`, `\v`, and — the one that matters — `\"` for a literal double quote (`iniEscapedString` in qsettings.cpp escapes `'"'` and `'\\'` through the same branch). Those all fall to the "escape we don't know" arm and survive as two literal characters, so a value carrying a quote reads back with a stray backslash in front of it. The class comment says the grammar covers "the named escapes", so this is a gap against its own contract rather than a deliberate omission. Adding `'"' => '"'` at minimum, and the four control ones for completeness, is a one-line change.

src/housecarl-core/Mo2ModMeta.cs
```diff
@@ -109,8 +109,8 @@ static IReadOnlyList<int> ReadInstalledFileIds(string[] lines)
         return null;
     }
 
-    /// <summary>Clean a QSettings value: strip an <c>@ByteArray(...)</c> wrapper, treat <c>@Invalid()</c> as unset,
-    /// unescape doubled backslashes, and strip a surrounding pair of double quotes. Empty ⇒ null.</summary>
+    /// <summary>Clean a QSettings value: strip an <c>@ByteArray(...)</c> wrapper, treat <c>@Invalid()</c> as unset, and
+    /// undo Qt's escaping, including a surrounding pair of double quotes (<see cref="QtIniEscapes"/>). Empty ⇒ null.</summary>
```
Avick3110: Doc drift: `Mo2Instance.cs`'s header block was updated to describe the full Qt grammar, but this file's header (lines 20-22) still reads "a value may be wrapped @ByteArray(...), @Invalid() means unset, backslashes are doubled, and string values can be surrounded by double quotes" — which is now the behaviour of the code that was just replaced. Since the header is what the next reader trusts for this file's quirks, it should point at `QtIniEscapes` the same way the xmldoc here now does.

src/housecarl-mcp-tests/Mo2IniEscapeTests.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// <summary>Write one value the way Qt does: printable ASCII as itself, a backslash doubled, everything else as
+    /// the UTF-8 bytes in <c>\xHH</c> escapes.</summary>
+    static string QtEscape(string value)
+    {
+        var sb = new StringBuilder();
+        foreach (var b in Encoding.UTF8.GetBytes(value))
+        {
+            if (b == (byte)'\\') sb.Append(@"\\");
+            else if (b >= 0x20 && b < 0x7F) sb.Append((char)b);
+            else sb.Append('\\').Append('x').Append(b.ToString("x2"));
```
Avick3110: **low/medium — the test helper does not encode the way Qt writes.** `QtEscape` is the fidelity anchor for this whole file, and it is missing Qt's `escapeNextIfDigit` rule: after emitting a `\xHH`, `QSettingsPrivate::iniEscapedString` escapes the *next* character too whenever it is a hex digit, precisely so the greedy run the decoder relies on stays unambiguous. Qt writes `C:\\游戏\\adata` as `...\x8f\x61data`; this helper writes `...\x8fadata`, which the decoder reads greedily as `8fa` (> 0xFF) and turns into a single UTF-16 code unit — a mismatch the suite can never catch, because the only non-ASCII fixtures here are followed by `\` or a space.

So the decoder's greedy rule is correct *because* Qt escapes the following digit, but no test states that. Teach the helper the rule and add a case where a non-ASCII run is immediately followed by a hex-digit character. The helper also never quotes (Qt quotes on `;` `,` `=` or a leading/trailing space) and never emits the Qt5 code-unit form, so the two theory rows for those are hand-written by the decoder's shape rather than produced by a writer that mimics Qt.

src/housecarl-core/Mo2ModMeta.cs
```diff
@@ -119,8 +119,7 @@ static IReadOnlyList<int> ReadInstalledFileIds(string[] lines)
         const string wrap = "@ByteArray(";
         if (v.StartsWith(wrap, StringComparison.Ordinal) && v.EndsWith(")", StringComparison.Ordinal))
             v = v[wrap.Length..^1];
-        if (v.Length >= 2 && v[0] == '"' && v[^1] == '"') v = v[1..^1];
-        v = v.Replace(@"\\", @"\").Trim();
+        v = QtIniEscapes.Unescape(v).Trim();
```
Avick3110: **low — the two readers are one code path but not one behaviour.** `Clean` trims *after* `Unescape`; `Mo2Instance.CleanValue` trims *before* and not after. The only reason Qt quotes a value at all is to preserve `;` `,` `=` or a leading/trailing space, so trimming the unescaped result is exactly the case the quote strip exists to protect: `installationFile=" Skyrim Mod.7z"` loses its leading space here and keeps it in the other reader. Pick one order and share it (the trim belongs before the wrapper/quote handling, on the raw line, not after decoding).

src/housecarl-core/QtIniEscapes.cs
Avick3110: Folded, and the gate you suggest is what landed. Both of your examples were real: `D:\newgame\Skyrim SE` lost its `n`, and `E:\xEdit\Stock Game` ate `\xEd`.

The decision is made once per VALUE, never per escape: `IsQtEscaped` asks whether every backslash in the value begins an escape Qt writes; if one does not, the value did not come out of Qt's writer (QSettingsPrivate::iniEscapedString doubles every backslash) and it is returned untouched, backslashes and all. `D:\newgame\Skyrim SE` fails on `\S`, `E:\xEdit\Stock Game` on `\S`, `C:\Games\Skyrim` on `\G`. New theory `HandWrittenPathsAreLeftAlone` covers all three, and `QtWrittenValuesReadBackWhole` round-trips `D:\newgame\xEdit` through the Qt writer (`D:\\newgame\\xEdit` on disk) to prove the Qt-written form still decodes.

The one case this cannot separate is a hand-written path whose every segment happens to start with a grammar letter (`D:\new\temp`): it reads as Qt's, because nothing in the file distinguishes it. That is Qt's own ambiguity, it fails loudly on the existence check, and it is stated in the class header rather than papered over.

src/housecarl-core/Mo2Instance.cs
Avick3110: Folded. The order is now trim the raw line, strip a surrounding quote pair, unwrap `@ByteArray(...)` / `@Invalid()`, then unescape — and it lives once, in `QtIniEscapes.Clean`, which both readers call, so `Mo2ModMeta` gets the same fix.

Test: `CommaProfileResolvesThroughTheQuotedWrapper` builds an instance whose profile is `Requiem, AE`, asserts the test writer emits `"@ByteArray(` (quotes outside the wrapper, as Qt does), and resolves it; `QtWrittenValuesReadBackWhole` also covers `say "hi"; done` and a value with edge spaces.

src/housecarl-mcp-tests/Mo2IniEscapeTests.cs
Avick3110: Folded. `QtWrite` now serializes the way Qt does: named escapes for `\` `"` and the control bytes, hex for anything outside printable ASCII, `escapeNextIfDigit` after every hex escape and after `\0`, and the surrounding quotes when the value holds `;` `,` `=` or an edge space — wrapper included.

The greedy-run claim is now proved by the writer rather than asserted: `QtWrittenValuesReadBackWhole` includes `大a` and `游戏7`, both a non-ASCII run followed immediately by a hex-digit character, which the writer escapes (`\xe5\xa4\xa7\x61`) and the reader therefore reads back whole. The quoted and comma cases go through the same writer too. The Qt5 code-unit form stays hand-written in `UnescapeReadsQtsGrammar`, because Qt5 with a codec is a writer this helper does not model and I would rather the row say plainly what shape it is asserting.

src/housecarl-core/QtIniEscapes.cs
Avick3110: Folded. `Named` now covers `\\ \" \; \, \= \0 \a \b \f \n \r \t \v`, and the same map is what the value-level gate consults, so a value using any of them still counts as Qt-written. `UnescapeReadsQtsGrammar` gained `\v` and `\"` to the named-escape row, and `QtWrittenValuesReadBackWhole` round-trips `say "hi"; done`.

src/housecarl-core/Mo2ModMeta.cs
Avick3110: Folded, your order. There is now one implementation — `QtIniEscapes.Clean` — and both `Mo2ModMeta.Clean` and `Mo2Instance.CleanValue` are one-line delegates to it, so they cannot drift again. The trim is on the raw line only; the decoded value is never trimmed, so `installationFile=" Skyrim Mod.7z"` keeps its leading space in both readers. Test: `QuotedEdgeSpaceSurvives`, plus the edge-space row in `QtWrittenValuesReadBackWhole`.

src/housecarl-core/Mo2ModMeta.cs
Avick3110: Folded. The header now says a value may be wrapped, quoted outside that wrapper, `@Invalid()` for unset, and Qt-escaped, and points at `QtIniEscapes.Clean` for the grammar instead of restating it — so there is one place to edit when the grammar changes. `Mo2Instance`'s header points at the same place.

PR #736, opened 2026-09-14
Publish houseCARL-Setup.exe framework-dependent

### LINKED ISSUE #734: houseCARL-Setup.exe 2.0.0 aborts in CLR startup on Ryzen 5800X3D / Win10 (CET shadow-stack assert); server itself is fine
`houseCARL-Setup.exe` (2.0.0) aborts during CLR startup on this machine and never reaches its plan prompt. The window appears and closes; nothing is written — no `installed-skills.txt`, no `*.houseCARL.uninstall.bak`, no config change.

```
CLR: Assert failure: !AreShadowStacksEnabled() || UseSpecialUserModeApc()
     coreclr\vm\threads.cpp:7954
```

**The good news, and the reason I think this is narrow: the server itself is completely fine.** Only the installer is affected.

## Environment

| | |
|---|---|
| CPU | AMD Ryzen 7 5800X3D |
| OS | Windows 10 Pro 22H2 (19045) |
| .NET runtimes | `Microsoft.NETCore.App` 5.0.13, 6.0.14, 7.0.10, 8.0.14, **9.0.1, 9.0.3, 9.0.11**, 10.0.5 |
| Package | `houseCARL-2.0.0.zip`, 19,429,980 bytes, sha256 `ba28ba9f3517a96e5e644578d114551f8368e00d4640822cf8b09984758d056a` |

## Diagnosis

The assert is the CET / hardware-shadow-stack interaction with a **self-contained** .NET apphost. I've hit the same family before on this CPU with the .NET 10 SDK's Roslyn, and it's why two other Skyrim tools here (Spriggit, the Synthesis CLI) are built from source targeting `net9.0` instead of using their shipped self-contained builds.

The distinguishing detail: **`housecarl/server/housecarl-mcp.dll` is framework-dependent** (`tfm: net9.0`, byte-identical `runtimeconfig.json` to 1.9.0's), and it starts, resolves the MO2 instance, and serves tools without any trouble:

```
$ dotnet housecarl-mcp.dll
houseCARL stdio server - reading MO2 instance 'D:\Authoria - 2.5.1' [saved user config]
   STANDALONE (MO2 need not be running); load order resolves lazily on the first tool call.
-> initialize OK in 0.4s, serverInfo {'name': 'houseCARL', 'version': '2.0.0'}
-> tools/list: 31 tools
```

So the failure is specific to the installer's own apphost, not to houseCARL.

## What I tried

`Set-ProcessMitigation -Name houseCARL-Setup.exe -Disable UserShadowStack` — every setting returned `C0000022` (access denied; it needs elevation), and the assert still fired. An elevated retry is untried, but I'd expect it to fail too: the assert is inside CLR startup, before any app code runs.

## Manual install, which worked

Recording this in case it's useful for troubleshooting docs — the package is plain enough that the installer isn't strictly required:

1. Extract `housecarl/` somewhere permanent.
2. Point the existing registration at the new `server/housecarl-mcp.dll`.
3. Copy `housecarl/skills/` into `~/.claude/skills/` (they registered with no restart).
4. Restart the host.

One thing that cost me time and might be worth a line in the README: **the Claude desktop app reads `%APPDATA%\Roaming\Claude\claude_desktop_config.json`**, not `~/.claude/claude_mcp_config.json`. I updated the latter (plus `~/.codex/config.toml`), restarted fully, and the app still spawned the 1.9.0 server — the only way I caught it was reading the spawned process's command line. Anyone hand-migrating will likely hit the same thing.

After that, everything works. For what it's worth, I ran an A/B of the new surface against answers captured from 1.9.0 before upgrading, and two of them came back as fixes: the whole-profile `housecarl_load_order_status` that used to time out on a 3,654-plugin order now returns, and `000014:Skyrim.esm` (PlayerRef) resolves instead of erroring. The retired-name refusals naming their successor and parameter mapping are genuinely nice — a stale call self-corrects instead of dead-ending.

## Suggestion

Either ship the setup as framework-dependent (it'd match the server, and .NET 9 is already required), or add a short manual-install section to the README for anyone whose machine refuses the apphost. Happy to test a build if that's useful.

Avick3110:
`houseCARL-Setup.exe` 2.0.0 shipped self-contained, single-file and trimmed, and on the reporter's Ryzen 5800X3D / Windows 10 22H2 its apphost asserts in CLR startup (`!AreShadowStacksEnabled() || UseSpecialUserModeApc()`, the CET hardware-shadow-stack interaction) before any app code runs: the window opens and closes, nothing is written. The server on the same machine is framework-dependent net9.0 and is fine. Since .NET 9 is already a requirement for the server, the embedded runtime bought nothing and cost those users the installer, so the setup now publishes framework-dependent too — `--self-contained false`, single-file kept (it bundles the managed assemblies, deps and runtimeconfig into one exe and leaves only the shared runtime outside), `PublishTrimmed` dropped because trimming is a self-contained-only option.

Size: **11,173,057 bytes before, 228,486 bytes after** (10.66 MB → 0.22 MB). The package root still holds exactly one setup file — no loose `.deps.json` or `.runtimeconfig.json` beside it — so the release zip's shape is unchanged apart from the size (12.99 MB packed).

What the .NET 9 requirement means for a user without it: the apphost prints Windows' own "You must install .NET to run this application" message with the download link and exits. That is acceptable, and no new mechanism was added for it. The setup's existing runtime preflight is still the thing that earns its keep: it catches the genuinely silent trap, where the base .NET Runtime is present (so setup starts) but the ASP.NET Core Runtime — a separate installer that the base one does not carry — is missing, which would otherwise install a server that never starts. Every place that said setup runs on a machine with no .NET at all is corrected: the publish comment in `scripts/build-plugin.ps1`, the detection comment in `Program.cs`, the `Ui.cs` header, the `START-HERE.txt` requirements block, and the requirements bullet in both READMEs. Nothing in `THIRD-PARTY-NOTICES.txt` claimed a bundled runtime — the notices generator only walks the server publish, and the server was already framework-dependent — and no test, probe or packaging step asserted the setup's publish shape.

The README also gains the reporter's two facts, in both `README.md` and the shipped `plugin/README.md`: a short by-hand install (extract, point the host's MCP entry at `server/housecarl-mcp.dll`, copy `skills/` into `~/.claude/skills/`, restart), and the note that the Claude desktop app reads its MCP servers from `%APPDATA%\Claude\claude_desktop_config.json`, not `~/.claude/claude_mcp_config.json` — editing the latter leaves the app spawning whatever server it had registered before, which is what cost the reporter time.

Verified locally on this machine: `dotnet build housecarl.sln -c Release` clean; `dotnet test --filter "tier!=bridge"` 1936 passed, 0 failed; `housecarl-generator ci-all` 128/128 passed; a full `scripts/build-plugin.ps1` run into the worktree's own `dist/` packed 293 entries with the leak-check clean; `claude plugin validate dist/housecarl --strict` passes. The new exe was then run with stdin and stdout redirected and `HOUSECARL_SETUP_HOME` / `CODEX_HOME` pointed at scratch directories — it starts, prints its banner, reports both runtimes found and neither host present, and stops with the redirected-input refusal naming the flag that would answer the question. That is CLR startup reached and app code running, which is the thing that failed before; the shadow-stack machine itself is the reporter's, so the fix is unverified there.

The reporter offered to test a build. A test build can be attached for them once you approve — I have not contacted them.

Closes #734


src/housecarl-setup/Program.cs
```diff
@@ -115,9 +115,12 @@ private static int Main(string[] args)
             // base .NET Runtime (Microsoft.NETCore.App) and the ASP.NET Core Runtime
             // (Microsoft.AspNetCore.App). On Windows those are TWO separate installers, and the
             // ASP.NET Core one does NOT include the base runtime -- a real-world install trap.
-            // This exe ships self-contained precisely so it still runs on a machine with neither
-            // and can say exactly what's missing, instead of the install "succeeding" into a
-            // server that never starts.
+            // This exe is framework-dependent too (#734: a self-contained apphost aborts in CLR
+            // startup on some CET / shadow-stack machines), so it needs the base runtime just to
+            // start; a machine without it gets Windows' own "You must install .NET" message and its
+            // download link. The check below is what catches the trap that is actually silent: the
```
Avick3110: **The base-runtime half of the preflight can now contradict the process running it.** `MissingServerRuntimes()` still treats `Microsoft.NETCore.App` as something to discover: it shells out to `dotnet --list-runtimes` and, when that fails, falls back to scanning only `%ProgramFiles%\dotnet\shared`. Neither path knows about `DOTNET_ROOT` or a runtime installed anywhere else.

Concrete scenario: .NET 9 installed to `C:\dotnet` with `DOTNET_ROOT` set (portable and locked-down enterprise layouts do exactly this), `dotnet` not on PATH so the launch throws and the `catch` at line 547 swallows it, and the folder scan finds nothing under Program Files. Setup — which the apphost has just started *on that very runtime* — prints `.NET Runtime 9  not found`, and `ReportMissingRuntimes` refuses the install with "houseCARL needs the .NET Runtime 9, which is not installed on this machine" and exits 1.

Before this PR the self-contained exe genuinely knew nothing about the machine, so that answer was at least plausible. Now the process's own existence is proof to the contrary, and the `baseMissing && aspMissing` branch ("neither is installed on this machine") is unreachable in truth while still reachable in code. The change also makes the fix cheap: seed `found` with `Microsoft.NETCore.App` unconditionally, or add `RuntimeEnvironment.GetRuntimeDirectory()`'s shared-framework root as a third source — which would fix the ASP.NET Core half in the same layouts too.

README.md
```diff
@@ -168,6 +168,17 @@ Flags, for an unattended run: `--claude` / `--codex` / `--both` pick the host, `
...
+### By hand
+
+The package is plain enough to install without the setup utility, which is the way out if the exe will not run on a machine.
+
+1. Extract `housecarl/` somewhere permanent.
+2. Register the server: point the host's MCP entry at `server/housecarl-mcp.dll`, run with `dotnet`.
+3. Copy `housecarl/skills/` into `~/.claude/skills/`.
+4. Restart the host.
+
+The Claude desktop app reads its MCP servers from `%APPDATA%\Claude\claude_desktop_config.json`, not from `~/.claude/claude_mcp_config.json`. Editing the latter changes nothing the desktop app spawns, and the app goes on starting whatever server it had registered before.
```
Avick3110: **This contradicts line 162, eighteen lines above it in the same section, and the contradiction lands on the by-hand reader.** Line 162 says the Claude install registers the server in `~/.claude.json` and that "The CLI and the desktop app both read these." This line says the desktop app reads its MCP servers from `%APPDATA%\Claude\claude_desktop_config.json`. Both cannot be true of the same host, and the by-hand step above ("point the host's MCP entry at...") never says which file it means.

Concrete scenario: a user whose exe will not run follows this section, writes `claude_desktop_config.json`, gets nothing, scrolls up, reads line 162, writes `~/.claude.json` too, and now has houseCARL registered twice with no idea which one did it — or reads line 180 first, concludes the setup utility has been writing the wrong file all along, and reports that as a bug.

The missing distinction is presumably the desktop app's *plain chat* (`claude_desktop_config.json`) versus its *Code tab* (`~/.claude.json`), which is the surface houseCARL actually targets. Worth saying which is which. Also note that `~/.claude/claude_mcp_config.json` — the path this warns readers off — appears nowhere else in the repo, while `~/.claude.json`, the path houseCARL does write, goes unmentioned here.

Same text at `plugin/README.md:82`, with the same claim at `plugin/README.md:64` above it.

README.md
```diff
@@ -168,6 +168,17 @@ Flags, for an unattended run: `--claude` / `--codex` / `--both` pick the host, `
 
 Updating: quit Claude Code and Codex first. Setup cannot replace a server a session is running; if one is, it stops and says so.
 
+### By hand
+
+The package is plain enough to install without the setup utility, which is the way out if the exe will not run on a machine.
+
+1. Extract `housecarl/` somewhere permanent.
+2. Register the server: point the host's MCP entry at `server/housecarl-mcp.dll`, run with `dotnet`.
+3. Copy `housecarl/skills/` into `~/.claude/skills/`.
```
Avick3110: **This step is ambiguous, and neither reading reproduces the layout setup actually installs.** The package ships `housecarl/skills/<skill-name>/SKILL.md`, and setup's Claude route copies the *whole plugin tree* to `ClaudeSkillsDest` = `~/.claude/skills/housecarl/` (`Program.cs:409`), so the known-good result is `~/.claude/skills/housecarl/skills/<name>/SKILL.md` alongside `.claude-plugin/plugin.json` and `server/`.

"Copy `housecarl/skills/` into `~/.claude/skills/`" reads two ways and lands on neither: taken literally it produces `~/.claude/skills/skills/<name>/SKILL.md`, which nothing discovers; taken as "copy the contents" it produces flat `~/.claude/skills/<name>/SKILL.md`, which is the *Codex* layout (`CodexSkillsRoot`, described as flat at `Program.cs:432`) and drops the `plugin.json` the plugin-tree install carries.

Concrete scenario: the reporter whose exe would not start — the exact audience this section was written for — follows step 3, restarts, and has no houseCARL skills, with no error to go on. Since this recovery path exists precisely because the supported path failed, it should spell out the layout setup produces: copy the whole `housecarl/` folder to `~/.claude/skills/housecarl/`.

Same text at `plugin/README.md:79`.

scripts/build-plugin.ps1
```diff
@@ -251,15 +251,16 @@ Copy-Item (Join-Path $PackagingSrc 'marketplace.json') (Join-Path $MpDir 'market
...
-# machine with no .NET installed at all, so it can preflight-check the two runtimes the
-# framework-dependent SERVER needs (.NET Runtime + ASP.NET Core Runtime - separate installers on
-# Windows) and say exactly which is missing. Trimming is safe HERE (setup uses only the
-# System.Text.Json DOM, no reflection serialization) - the server's trimming ban is untouched.
+# picks it up. Single-file and FRAMEWORK-DEPENDENT, matching the server: a self-contained apphost
+# aborts in CLR startup on some CET / hardware-shadow-stack machines (#734), and .NET 9 is already a
+# requirement for the server, so embedding a runtime bought nothing and cost those users the
+# installer. Setup still preflights the ASP.NET Core Runtime the server needs separately; a machine
+# missing the base .NET Runtime gets Windows' own "You must install .NET" message with the download
+# link, because the apphost never reaches app code. Trimming is self-contained only, so it is gone.
```
Avick3110: **Medium — the promised "Windows says so" message is stderr-only for a console app, so a double-click still just flashes a window.**

This comment, the `Program.cs` comment, the `START-HERE.txt` block and both README requirement bullets all rest on the same claim: without the base runtime, "Windows' own 'You must install .NET' message with the download link" appears. The apphost only shows that as a *dialog* when it was built with the Windows GUI subsystem bit — i.e. `<OutputType>WinExe</OutputType>`. `housecarl-setup.csproj` is `<OutputType>Exe</OutputType>`, so hostfxr takes the other branch and writes the missing-framework text to **stderr**.

Double-clicked from Explorer (step 1 of `START-HERE.txt`, and how every non-terminal user runs this), a console process gets a fresh console window that is destroyed the instant the process exits. The user sees a window open and close with nothing written — *the exact symptom of #734*, which is what this PR is fixing. The message is only actually readable if the exe is launched from an already-open terminal.

That makes the new doc sentences overpromise in the one situation they were written for, and it removes the clean diagnosis the self-contained build gave (it started, and `ReportMissingRuntimes` printed the two winget lines and held the window). Worth either saying so plainly in the docs ("run it from a terminal to see why"), pointing at the by-hand section from `START-HERE.txt`, or reconsidering whether the base-runtime case deserves a real dialog.

Not a blocker on the publish-flag change itself, which is right; the claim built on top of it is what is wrong.

README.md
```diff
@@ -148,7 +148,7 @@ Dialogue and facegen are not skills. Their tools carry the bookkeeping, and the
 ## Requirements
 
 - Windows.
-- .NET Runtime 9.0 and ASP.NET Core Runtime 9.0, from the [.NET 9 download page](https://dotnet.microsoft.com/download/dotnet/9.0). Both are required. The ASP.NET Core installer does not include the base runtime. The setup utility checks for both and names the one that is missing.
+- .NET Runtime 9.0 and ASP.NET Core Runtime 9.0, from the [.NET 9 download page](https://dotnet.microsoft.com/download/dotnet/9.0). Both are required. The ASP.NET Core installer does not include the base runtime. The setup utility ships framework-dependent like the server, so it needs the base runtime to start at all — without it Windows says so and links the download page — and it checks for the ASP.NET Core one and names it if that is what is missing.
```
Avick3110: **Low/medium — the architecture now matters for the setup exe too, and this bullet (and `plugin/README.md:53`) still says only "Windows."**

`packaging/START-HERE.txt:81` says "Windows (x64)"; both READMEs say "Windows." That was survivable while `houseCARL-Setup.exe` was self-contained — it carried its own x64 runtime and needed nothing installed. Now it is a `win-x64` framework-dependent apphost, so it needs an **x64** .NET 9 Runtime specifically; an arm64 install of the same version does not satisfy it. The .NET download page offers the arm64 build by default on Windows-on-ARM, so a user following exactly this bullet's link can install "the .NET Runtime 9.0" and still have the exe refuse to start — and, per the other comment, refuse silently.

Since this bullet was rewritten anyway to explain what setup needs to start, saying x64 here (matching `START-HERE.txt`) costs one word.

scripts/build-plugin.ps1
Avick3110: Folded, option (a). Confirmed: OutputType is Exe, so hostfxr writes its missing-framework text to stderr and a double-clicked console window takes it with it on exit. A dialog would need the GUI subsystem bit, which would cost this utility its entire console UI, so that is not a one-property change and is not attempted here. Every place that promised a message now names the flash-and-close itself as the diagnosis instead: both README requirement bullets and the `START-HERE.txt` requirements block say that if the window opens and closes with nothing in it, the base .NET 9 runtime is missing, install it from the linked page and run setup again. The `build-plugin.ps1` and `Program.cs` comments say the same, and why. 22575949

src/housecarl-setup/Program.cs
Avick3110: Folded. `MissingServerRuntimes` now scans two shared-framework roots rather than one: the machine-wide `%ProgramFiles%\dotnet\shared`, and the root the setup process is itself running out of, derived from `RuntimeEnvironment.GetRuntimeDirectory()` (its grandparent — verified on this machine: `C:\Program Files\dotnet\shared\Microsoft.NETCore.App\9.0.18\` gives `C:\Program Files\dotnet\shared`). That covers the ASP.NET Core half of a `DOTNET_ROOT` or portable layout too, as you noted. I scanned rather than seeding `Microsoft.NETCore.App` unconditionally because roll-forward can put this net9.0 apphost on a .NET 10 runtime, where the process's existence is not proof of a 9.x base runtime; scanning keeps the answer honest in that case. Verified only that the default layout is unchanged (both roots resolve to the same folder here) — I have no `DOTNET_ROOT` install to test against, so the new root is a superset scan, not a tested path. 5d873e11

README.md
Avick3110: Folded, in both READMEs. The truth from `Program.cs`: the Claude route writes `mcpServers.housecarl` into `~/.claude.json` as `{type: stdio, command: <dest>\server\housecarl-mcp.exe, args: []}`, and that file is Claude Code's — the terminal CLI and the desktop app's Code tab. The line above now says 'the desktop app's Code tab' rather than 'the desktop app', which is the distinction that was missing. The by-hand paragraph now names `~/.claude.json` as the file to edit, gives the exact entry, and says the desktop app's plain chat is a separate surface with its own `%APPDATA%\Claude\claude_desktop_config.json` taking the same entry — which is what the reporter hit. `~/.claude/claude_mcp_config.json` is kept only as the negative (neither surface reads it, which is why editing it changed nothing for them), no longer as the positive claim. 22575949

README.md
Avick3110: Folded, in both READMEs. Step 1 now says to move the whole `housecarl` folder to `~/.claude/skills/housecarl`, so that folder holds `.claude-plugin/`, `server/` and `skills/` — the layout `ClaudeSkillsDest` produces, stated as such — and the section says up front that it reproduces what setup does. The old wording is gone, along with its two readings. 22575949

README.md
Avick3110: Folded. Both READMEs now say 'Windows, x64' and call for the x64 builds of both runtimes, with your reason given: the download page offers arm64 by default on an ARM machine, and an arm64 runtime runs neither the win-x64 server nor the win-x64 setup apphost. `START-HERE.txt` marks both runtimes (x64) and carries the same sentence. Paired with the flash-and-close diagnosis from the other thread, the arm64 case now has something to read. 22575949

src/housecarl-setup/Program.cs
```diff
@@ -563,6 +569,25 @@ private static List<string> MissingServerRuntimes()
         return required.Where(fx => !found.Contains(fx)).ToList();
     }
 
+    /// <summary>The shared-framework roots to scan: the default machine-wide install, and the one this
+    /// process is running out of. GetRuntimeDirectory() is ...\shared\Microsoft.NETCore.App\&lt;version&gt;\,
+    /// so its grandparent is that install's shared root.</summary>
+    private static IEnumerable<string> SharedFrameworkRoots()
+    {
+        yield return Path.Combine(
```
Avick3110: **Low — the preflight is architecture-blind, so on an ARM64 machine an arm64 ASP.NET Core Runtime satisfies the check for the x64 server that cannot use it.**

Scenario: Windows on ARM64. The user installs the x64 .NET Runtime (so the win-x64 setup starts) and, taking the download page's default that the new README text warns about, the arm64 ASP.NET Core Runtime. `dotnet` on PATH is the arm64 host and `--list-runtimes` lists `Microsoft.AspNetCore.App 9.x`; if that step fails, the machine-wide scan of `%ProgramFiles%\dotnet\shared` finds the arm64 dlls as well (x64 .NET lives under `%ProgramFiles%\dotnet\x64\` on ARM64). Setup reports both runtimes present and installs; the x64 server's apphost then fails with the missing-framework error — the silent trap the README and START-HERE now say setup catches.

The own-root scan (`%ProgramFiles%\dotnet\x64\shared`) is the only architecture-correct one of the three. Either trust it alone when it resolves, or stop the `dotnet --list-runtimes` and machine-wide results counting when `RuntimeInformation.OSArchitecture != RuntimeInformation.ProcessArchitecture`.

src/housecarl-setup/Program.cs
```diff
@@ -563,6 +569,25 @@ private static List<string> MissingServerRuntimes()
...
+            Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "dotnet", "shared");
+
+        string? own = null;
+        try
+        {
+            own = Path.GetDirectoryName(
```
Avick3110: **Low — scanning the root setup itself runs from can report a framework present that the server will not find, because the server runs in the host's environment, not setup's.**

Scenario: a portable .NET 9 at `D:\dotnet` reached only through `DOTNET_ROOT` set in the user's shell profile (not the machine or user environment), `dotnet` off PATH, nothing under `%ProgramFiles%\dotnet`. Setup launched from that shell runs out of `D:\dotnet\shared\Microsoft.NETCore.App\9.x`, so `SharedFrameworkRoots()` yields `D:\dotnet\shared`, finds both frameworks, and setup installs. Claude desktop or Claude Code launched from Explorer or the Start menu does not inherit that `DOTNET_ROOT`; its spawn of `housecarl-mcp.exe` resolves through the registry and `%ProgramFiles%\dotnet` only, finds no runtime, and the server never starts. Before this change the same machine got a correct "missing" report.

At minimum, when the process-own root is what satisfied the check and it differs from the machine-wide install, the detection report should say so, so the user can tell the host may not see it.

src/housecarl-setup/Program.cs
Avick3110: Folded in fb80d70a, the second of your two options. `MissingServerRuntimes` now computes three sources separately — `dotnet --list-runtimes`, the machine-wide `%ProgramFiles%\dotnet\shared` scan, and the own-root scan — and when `RuntimeInformation.OSArchitecture != RuntimeInformation.ProcessArchitecture` the first two contribute nothing, leaving the own root (the architecture-correct one, `%ProgramFiles%\dotnet\x64\shared` in your scenario) to answer alone. I took the discount rather than 'trust the own root alone whenever it resolves' because the own root resolves on every normal machine, and dropping `dotnet --list-runtimes` there would lose the custom-location-on-PATH case the first source exists for. Untested on ARM64 — no such machine here — so this is reasoning from the layouts, not a measurement.

src/housecarl-setup/Program.cs
Avick3110: Folded in fb80d70a, at the minimum you set and no further. `MissingServerRuntimes` now returns the own root alongside the missing list, but only when it is what decided the answer — it supplied a framework no other source found — and its path differs from the machine-wide install. In that case the detection block prints one `Ui.Note` after the rows: that the runtime was found where setup is running from rather than in the machine-wide install, so the server may not start if the agent is launched without it, with the root named beneath. No gate, no refusal, no environment writing. On an ordinary machine the own root is the machine-wide one, so nothing prints — confirmed on this one.

scripts/build-plugin.ps1
Avick3110: Folded at 22575949 (now adc1771c/36ad6246 after the rebase), option one of the three you named, plus the residue your wording asked for and mine had missed. Confirmed `OutputType=Exe`, so the message is stderr-only; a dialog would need the GUI subsystem bit, which would cost this console utility its entire UI, so it is not attempted. Both README requirement bullets, the `START-HERE.txt` block and the `build-plugin.ps1` / `Program.cs` comments now name the flash-and-close as the diagnosis rather than promising a message. Added since, in 3c93755a: 'run it from an open terminal instead and you get the message the closing window takes with it' in all three docs, and `START-HERE.txt` now points at the README's By hand section for a machine where setup will not run at all.

README.md
Avick3110: Folded at 22575949 (now 36ad6246). The distinction you guessed is the one in `Program.cs`: the Claude route writes `mcpServers.housecarl` into `~/.claude.json`, which is Claude Code's file — terminal CLI and the desktop app's Code tab — so that line now says 'the desktop app's Code tab', in both READMEs. The by-hand paragraph names `~/.claude.json` as the file to edit and gives the exact entry (`type`, `command` at `server\housecarl-mcp.exe`, `args`); the desktop app's plain chat is described as the separate surface with its own `%APPDATA%\Claude\claude_desktop_config.json` taking the same entry. `~/.claude/claude_mcp_config.json` survives only as the negative — neither surface reads it, which is why editing it changed nothing for the reporter.

README.md
Avick3110: Folded at 22575949 (now 36ad6246), as the layout you named. Step 1 is now: move the whole `housecarl` folder to `~/.claude/skills/housecarl`, so that folder holds `.claude-plugin/`, `server/` and `skills/`; the section says up front it reproduces what setup does. Both readings of the old wording are gone with it. Same in `plugin/README.md`.

src/housecarl-setup/Program.cs
Avick3110: Folded at 22575949 (now adc1771c), your third-source option. `RuntimeEnvironment.GetRuntimeDirectory()`'s grandparent is scanned as a shared-framework root beside the machine-wide one, which covers the ASP.NET Core half of those layouts as you said; derivation verified on a net9.0 process here (`...\shared\Microsoft.NETCore.App\9.0.18\` gives `...\shared`). Scanned rather than seeding `Microsoft.NETCore.App` unconditionally, because roll-forward can put this net9.0 apphost on a .NET 10 runtime, where the process's existence is not proof of a 9.x base runtime. Your two later comments on this code are folded in fb80d70a.

README.md
Avick3110: Folded at 22575949 (now 36ad6246). Both READMEs say 'Windows, x64' and call for the x64 builds of both runtimes, carrying your reason: the page offers arm64 by default on Windows-on-ARM, and an arm64 runtime runs neither the win-x64 server nor the win-x64 setup apphost. `START-HERE.txt` marks both runtimes (x64) and says the same. Your ARM64 comment on the preflight is folded separately in fb80d70a, so setup no longer passes an arm64 ASP.NET Core Runtime for the x64 server either.

PR #737, opened 2026-09-14
Verify a nif_set edit against the block id the saved mesh carries

### LINKED ISSUE #732: [Bug] nif_set refuses valid BSShaderTextureSet edit on meshes with large embedded node trees
### houseCARL version
1.9.0

### Exact repro
`housecarl_nif_set` call:
```
mesh_path: meshes\haijin NPCs\Babette\femalehands_0.nif
op: set_path
target: Hands.001
texture_slot: 0
path: textures\actors\character\female\femalehands_1.dds
patch: Babette - Bijin Skin Hands Fix
```

Result (verbatim):
```
nif set — (mesh)  (profile 'Default')

the edit changed block(s) [42 BSShaderTextureSet] it should not have touched (expected only [6]). Refusing, nothing written.
  providers (1): "haijin NPCs - Babette v2.6" (loose)
```

Reproduced identically on a second, structurally similar file (`femalehands_1.nif`, same shape name `Hands.001`, same mod) — same message, same expected-vs-actual block numbers (expected `[6]`, actual `[42 BSShaderTextureSet]`).

`housecarl_nif_inspect` (`sections=paths`) run immediately before, on both files, shows a block census of exactly **one** `BSShaderTextureSet` per file, so block 42 is that mesh's *only* texture-set block — there's no ambiguity for the op to have picked the wrong one among several. The gate's "expected [6]" looks like a stale/miscomputed index rather than a real mismatch.

Both affected files are custom NPC hand meshes with an unusually large embedded node tree: 43 blocks total, 37× `NiNode` (`shapes (1): 'Hands.001'`). For comparison, this same NPC's body/feet meshes (`femalebody_0/1.nif`, `femalefeet_0/1.nif`) have 29 blocks / 23× `NiNode` and were only read, never written, so I can't say whether they'd hit the same gate — only the 37-`NiNode` hand meshes were exercised with a write. Hypothesis: whatever computes the "expected" block index for the verification gate doesn't hold up once the node tree is unusually large / the texture-set block sits far from block 6.

### Expected
The write should succeed: `target='Hands.001'`, `texture_slot=0` unambiguously resolves to the mesh's single `BSShaderTextureSet` (block 42, per `housecarl_nif_inspect`'s own census). The verification gate should accept that as correct instead of refusing on an "expected [6]" that doesn't match what the mesh actually contains.

### Environment
Skyrim SE 1.5.97, MO2-managed instance. Both files are loose, from a single-provider custom NPC replacer mod (not BSA-packed, not a merged/patched copy).

Avick3110:
`nif_set`'s first verification gate diffs the saved mesh against a normalized copy of the original, block by block, and refuses if any block other than the ones the ops declared changed. The declared id came from `BlockIndexOf`, which is the block's position in `NifFile.Blocks` as the mesh was *loaded*. That is the wrong index space: NiflySharp's `Save` defaults to `SortBlocks = true` and re-sorts the block list into its own canonical tree order, which is the order both sides of the diff are sliced in. Confirmed directly against NiflySharp 1.1.0 — a mesh saved with `SortBlocks = false` loads with its texture set at one id and, after a plain `Save`, carries it at another, while `NifFile.Blocks` and the header's block-type table agree with each other at every point. So on a mesh an exporter wrote in an order that is not nifly's — which is what both meshes in the report are — the expected id named a different block than the one the diff saw change, and a correct `set_path` was refused as collateral damage. That is the "expected only [6]" against a texture set at 42: 6 was its loaded id, 42 its saved one.

The fix keeps the block *object* through the apply loop and resolves each id only after the save, from the same post-save list the diff compares in. Nothing else moves: the gate still refuses a real collateral change, and no parameter description, refusal wording, or output shape changes.

Test: `NifSetBlockOrderTests` authors an SE mesh with the texture set and shader ahead of a run of `NiNode`s and saves it with sorting off, so its stored order is not the save order. One test pins that the fixture's two id spaces really disagree (so the other cannot go vacuous if NiflySharp stops sorting); the other runs `set_path` slot 0 on it, which refused before this change and now writes, verifies, and reports the texture set's saved id. The existing gate-1 arms in the `nif-set-guard` probe, which prove a collateral change is still caught, pass unchanged.

Build, the full xUnit suite (1,938 passed) and `ci-all` (128/128) all pass.

Closes #732


src/housecarl-core/NifService.cs
```diff
@@ -588,6 +588,13 @@ public static NifSetOutcome Set(byte[] bytes, IReadOnlyList<NifSetOp> ops)
         }
         catch (Exception ex) { return NifSetOutcome.Fail($"saving the edited mesh threw — {ex.GetType().Name}: {ex.Message}. Nothing was written."); }
 
+        // The block ids the diff compares are the ids of the SAVED file, and NiflySharp's save re-sorts the block list
+        // into its own canonical tree order — so a mesh whose on-disk order was not already that order is renumbered.
+        // Resolve each touched block's id only now, from the post-save list, or the gate compares the right block
+        // against the wrong index and refuses a correct edit.
+        var expectedBlocks = new HashSet<int>();
+        foreach (var b in touched) expectedBlocks.Add(BlockIndexOf(nif, b));
```
Avick3110: `BlockIndexOf` returns -1 when the block object is not found in the post-save list, and that -1 is added to `expectedBlocks` unchecked. Before this change the lookup ran against the list the block had just been read from, so -1 was genuinely unreachable; resolving after `Save` makes it reachable — anything nifly's save-time finalize does that replaces or drops a block object (a regenerated skin instance/partition, say) leaves the id unresolved. The failure is silent and misleading rather than a plain error: gate 1 then sees the real changed block as collateral and refuses a correct edit with `expected only [-1]`. The `object?` tuple field also drops the compile-time protection the old `int?` had, so a future op returning an index by mistake lands in the same hole. Suggest failing by name instead:

```suggestion
        var expectedBlocks = new HashSet<int>();
        foreach (var b in touched)
        {
            int id = BlockIndexOf(nif, b);
            if (id < 0) return NifSetOutcome.Fail("verification could not locate an edited block in the saved mesh — refusing to write. Nothing was written.");
            expectedBlocks.Add(id);
        }
```

Avick3110: Folded in a771ae47. An unresolved id is now a named refusal before gate 1 runs, and the tuple field is `NiflySharp.INiObject?` rather than `object?` (so is `BlockIndexOf`'s parameter), which keeps an index from landing there by mistake.

PR #738, opened 2026-09-14
Publish the tool schemas cut to HOUSECARL_MAX_SCHEMA_DEPTH

### LINKED ISSUE #730: Add HouseCarl__MaxSchemaDepth to cap published tool schema depth for providers with a nesting limit
## Summary

Follow-up to #451 (closed by #465 in 1.9.0, carried into 2.0.0). The recursive `$ref` is gone — `binding-shim-guard` enforces zero `$ref` members across all 31 published schemas, and the "Recursive JSON schemas are not currently supported" error no longer fires. But flattening the recursion inlined the compose chain to its full depth, and two tools now publish schemas that **exceed the 10-level nesting cap** some strict providers enforce. The failure shape is identical to #451 from the user's seat: the whole server is refused at `tools/list`, naming no tool, and the only workaround is toggling houseCARL off for that model.

## Exact error

```
Error from provider (Console Go): Upstream request failed: [invalid_request_error] JSON schema exceeds the maximum nesting depth of 10 levels
```

- Provider: `opencode-go` (Console Go relay), model `muse-spark-1.3-contributor`
- Trigger: houseCARL MCP 2.0.0 enabled → the provider validates every tool's `inputSchema` at `tools/list` and fails the **entire session** (not just the offending tool)
- Other models on the same provider (deepseek-flash, glm-5.x, qwen3.7-plus, kimi-k2.x ...) accept the same schemas fine — only strictly-validating providers reject depth > 10

## Measured on the installed 2.0.0 server

All 31 tool `inputSchema`s, max nesting depth (object/array member recursion):

| depth | tool | over 10? |
|---:|---|:---:|
| 24 | `housecarl_create` | yes |
| 21 | `housecarl_apply` | yes |
| 9 | `housecarl_place`, `housecarl_records` | no |
| 5 | `housecarl_copy`, `housecarl_nif_inspect`, `housecarl_remove`, `housecarl_forward`, `housecarl_check`, `housecarl_asset_status` | no |
| ≤4 | the other 21 tools | no |

Only `housecarl_create` and `housecarl_apply` exceed 10. Both are the write surface — `ops[].compose.sets[].compose...` and `records[].ops[].compose.sets[].compose...` are where the depth lives, the same recursive DTO chain #465 inlined. The other 29 tools (all reads, NIF, BSA, Nexus, merge, compile, etc.) are at 9 or below and load fine on the same provider.

Server self-report: `2.0.0+5e3c744862bc17c85ebe850830603a9a3e6f0779`.

## Repro

1. houseCARL 2.0.0 enabled in opencode (local stdio, `housecarl-mcp.exe`)
2. Select model `muse-spark-1.3-contributor` via `opencode-go`
3. Send any message → the above error, every time
4. Disable houseCARL → the model works; re-enable → error returns

## Why it matters

Same user-facing shape as #451: a growing set of smaller / relay providers validate `tools/list` strictly and refuse the whole server on a schema feature they don't support, with an error naming neither houseCARL nor the tool. ADR 0002 already rules the posture for the `$ref` case — "a schema feature that some validators reject is a liability regardless of who is technically correct" — and the same posture applies to unbounded nesting depth. The two affected tools are also the entire write surface, so a per-tool blacklist (the workaround today) costs the contract's write half.

## Suggested fix

Analogous to the `FlatToolSchemas` switch #451's reporter floated, but for depth: a bounded-depth cap on the published schemas, configurable so the unbounded form stays available for providers that accept it. Concretely:

- A `HouseCarl__MaxSchemaDepth` env (default: unlimited, or a generous cap like 32), and when set `ToolSchemas.PublishSchemas` truncates any branch past the cap with a permissive open node — the same shape the inliner already emits for forms it does not handle (`{"type":["object","null"]}`). `tools/call` parameter handling is unchanged; only the published schema is bounded.
- Or, default to a conservative cap (e.g. 10) at the publication layer, since a deeply-nested compose is a structural fact the caller rarely needs at the schema level — the call-time validation already enforces the real shape, and the description carries the verbs and member names a client needs to compose a call.

Either keeps the write surface usable on strict providers without narrowing the contract on providers that accept the depth. A depth cap is also the cheapest reachable fix for the "no `$ref` members" invariant the guard already holds: the truncated node is a leaf, so it cannot introduce a `$ref` the inliner would have to chase.

## Environment

- houseCARL 2.0.0 (local stdio via `housecarl-mcp.exe`, opencode desktop)
- Windows 11, .NET 9
- Affected provider: opencode-go / Console Go, model `muse-spark-1.3-contributor`
- Measured against the installed 2.0.0 server, `server: 2.0.0+5e3c7448`

## Workaround in use today

A per-tool blacklist (`housecarl_housecarl_create` / `housecarl_housecarl_apply` = false in the MCP client config) loads the server on the strict provider, at the cost of the two write tools — the rest of the surface (29 tools, all reads + NIF + BSA + Nexus + merge + compile) remains usable. Filed so the cap can replace the blacklist.

Avick3110:
A `HOUSECARL_MAX_SCHEMA_DEPTH` environment variable that publishes the tool schemas cut to a nesting depth, for a provider that refuses the whole server at `tools/list` over depth. Flattening the recursion in #465 inlined the compose chain to its full depth, so `housecarl_create` and `housecarl_apply` publish at 24 and 21 levels and a provider capped at 10 takes neither — and it refuses the entire server, naming no tool. Set the variable to that provider's cap and a fourth publication pass (`SchemaDepthCap.cs`, run last over the finished document) cuts every schema there and closes the cut with the SAME node the recursion bound already emits: the node's own `type` and description plus the clause saying nesting continues below with the same shape and is accepted. `tools/call` is untouched at every accepted cap — the binder never consults the schema, and the floor of 3 keeps the shim's top-level `properties` intact — so a call nested deeper than the cut binds and is answered exactly as before; the schema is less descriptive below the cut, never wrong. Measured on the real server at 10: every published schema fits, `apply` lands at 10 and `create` at 9.

Depth is raw JSON container nesting — the schema object is level 1, every object or array below it one more — which is the measure the issue reported with; it reproduces the issue's table exactly (24 / 21 / 9 / 9 / 5 / ≤4) on today's surface. The reporter's trap is handled and tested: `properties`, `patternProperties` and `$defs` are name-to-schema dictionaries, so the cut recurses into their values and never replaces the container (`additionalProperties` is a schema and is cut as one).

Unset — the default — the pass does nothing. That is asserted rather than argued: the served schemas of a plain server are re-run through the uncapped cut, which must report no change and leave every document byte-identical, and a dump of `tools/list` before and after this branch is byte-for-byte the same 202,118 characters. A value that is not a whole number of 4 or more refuses the server's start with one plain sentence on stderr naming the variable, rather than booting and publishing the schemas the caller set it to avoid. The floor is 4 because `ToolCallShim` reads exactly two members of a published schema — the top-level `properties` and each parameter's `type` — and root=1, `properties`=2, parameter=3, that parameter's `type` list=4: at 1 or 2 the cut closes the root, at 3 it closes each parameter with a node too shallow for its `["array","null"]` list, and either way coercion, the named missing-parameter refusal, the typed-mismatch refusal, the undeclared-key refusal and the in-place filename refusal quietly stop happening. A cut may say less about a nested shape; it may not change what a call gets back. Two more refusals have the same shape: a member at a cut point the pass has no rule for, and a finished document still measuring over the cap (`Cut` re-measures rather than trusting the induction over its branches).

**Spelling.** The issue proposed `HouseCarl__MaxSchemaDepth`; houseCARL's existing environment variables are `HOUSECARL_DATA_DIR` and `HOUSECARL_SETUP_HOME`, so this one is `HOUSECARL_MAX_SCHEMA_DEPTH` to match. (`HouseCarl__` is the .NET configuration-binder spelling and would have bound through `IConfiguration`; this value is read before either host is built, because an unusable value must stop the start.)

Thirty-nine tests on the served surface (`PublishedSchemaDepthTests`), a capped server of their own so the shared fixture is untouched: the cap fits every schema, no tool leaves the surface, the zero-`$ref` invariant still holds, every cut document is still structurally a schema (the same walk run over the uncapped surface, so a green result is not the check passing anything), a deeply nested `apply` call still reaches the tool body, each invalid value is refused at startup in one line, at the shallowest accepted cap every tool still publishes its own parameters and still refuses a missing required argument by name, every node the cut closed carries the cut's clause and none carries the bound's, every parameter keeps the `type` it published uncut at the floor and a bare-string `formids` is still coerced there, and the served surface carries no schema member outside the cut's keyword sets. Documented in `docs/architecture/tool-schema-publication.md` as pass 4 and in README beside the install notes. Full suite 1,995 green, `ci-all` 128/128.

Closes #730


src/housecarl-mcp/SchemaDepthCap.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// more. Refused rather than ignored: a caller who set it did so because a provider refuses the uncut schema,
+    /// and silently publishing uncut would fail that provider with an error naming neither houseCARL nor this
+    /// variable.</exception>
+    internal static int? Configured() => Read(Environment.GetEnvironmentVariable(Variable));
+
+    /// <summary>Parse one value of the variable. Separate from <see cref="Configured"/> so a test can drive it.</summary>
+    internal static int? Read(string? raw)
+    {
+        if (string.IsNullOrWhiteSpace(raw)) return null;
+        if (int.TryParse(raw.Trim(), NumberStyles.Integer, CultureInfo.InvariantCulture, out var depth) && depth >= 1)
```
Avick3110: `Read` accepts any depth of 1 or more, but a cap of 1 or 2 silently guts the call path the PR says it leaves alone.

At cap 2, `Shrink(root, 2)` hits the `properties` key, sees `budget < 3`, and returns `Terminator(root, 2)` — so every tool publishes `{"type":"object","description":"Nesting continues…"}` with **no `properties` and no `required`**. At cap 1 the `budget <= 1` branch does the same thing one step earlier.

That is not just "less descriptive": every pass in `ToolCallShim` bails on a schema with no top-level `properties` (`ToolCallShim.cs` lines 80, 121, 203, 262, 289, 312 all return when `TryGetProperty("properties", …)` fails). So with `HOUSECARL_MAX_SCHEMA_DEPTH=2` the server boots happily and quietly loses argument coercion, the named missing-required refusal, and the undeclared-key rejection — a caller who omits a required parameter gets the SDK's generic binding error instead of houseCARL's sentence, which is the silently-degraded mode the cornerstones forbid. The PR text ("`tools/call` is untouched — the binder never consulted the schema") and the doc paragraph are both wrong in this range, since the shim *does* consult the top-level `properties`.

The cut only behaves as advertised from cap 3 up, where `root(1)/properties(2)/param(3)` still leaves each parameter as a terminator carrying its `type`. Simplest fix: refuse below 3 in `Read`, and say so in the message.

src/housecarl-mcp/SchemaDepthCap.cs
```diff
@@ -0,0 +1,143 @@
...
+            else if (Depth(value) > budget - 1) node.Remove(key);
+        }
+        return node;
+    }
+
+    /// <summary>The node the cut closes a branch with — the SAME one the recursion bound emits, trimmed to fit:
+    /// a <c>type</c> spelled as a list needs a level of its own, which the shallowest budget does not have.</summary>
+    static JsonObject Terminator(JsonObject node, int budget)
+    {
+        var open = ToolSchemas.Terminator(node, node);
```
Avick3110: Reusing `ToolSchemas.Terminator` here also reuses its sentence, and that sentence is not true of a depth cut. It reads "Nesting continues below this level with the same shape **shown above**…" — accurate for the recursion bound, where the shape genuinely was spelled out earlier in the same document, but for a cut the shape appears nowhere at all: the branch was truncated because the document ran out of budget, not because it repeated. A model reading a capped `housecarl_apply` is told to look above for a `Condition` / `FunctionConditionData` shape that is not in the schema it was handed, and will either invent one or ask.

The cut's terminator should say what is true of a cut (nesting continues below this level and is accepted, but is not spelled out in this document), even if that means a second constant beside the recursion one.

src/housecarl-mcp/SchemaDepthCap.cs
```diff
@@ -0,0 +1,143 @@
...
+        throw new ArgumentException(
+            $"{Variable} is \"{raw}\" — set it to a whole number of 1 or more (the JSON nesting depth to publish " +
+            "the tool schemas at), or leave it unset to publish them in full.");
+    }
+
+    /// <summary>Cut one tool schema to <paramref name="maxDepth"/>. Returns false — leaving the document
+    /// untouched — when no cap is configured or the document already fits.</summary>
+    internal static bool Cut(JsonObject root, int? maxDepth)
+    {
+        if (maxDepth is not { } cap || Depth(root) <= cap) return false;
```
Avick3110, line 97: `Cut` reports success purely from the pre-cut measurement — it returns `true` as soon as `Depth(root) > cap` and never re-measures the result. The fit is only an induction over `Shrink`, and that induction has holes: the `Dictionaries` branch shrinks a member only `if (dictionary[name] is JsonObject member)`, and the `SchemaLists` branch only `if (arms[i] is JsonObject arm)`. Any non-object value in those positions is skipped entirely, keeps its full depth, and leaves the node it sits in over budget.

The failure is silent and is exactly the failure the feature exists to prevent: the operator sets the cap their provider demands, the server publishes a schema still above it, and the provider refuses the whole server at `tools/list` naming no tool — with nothing on stderr pointing at houseCARL. A single `Depth(root) <= cap` check at the end of `Cut` (throwing, or one sentence on stderr) turns that into a visible failure for the cost of one more walk at startup.

src/housecarl-mcp/SchemaDepthCap.cs
```diff
@@ -0,0 +1,143 @@
...
+            else if (SchemaLists.Contains(key) && value is JsonArray arms)
+            {
+                if (budget < 3) return Terminator(node, budget);
+                for (var i = 0; i < arms.Count; i++)
+                    if (arms[i] is JsonObject arm && Shrink(arm, budget - 2) is var cut
+                        && !ReferenceEquals(cut, arm)) arms[i] = cut;
+            }
+            // Anything else is an annotation or a scalar list (type, enum, required, default). It constrains
+            // nothing structurally, so one too deep to keep is dropped rather than cut into a shape of its own.
+            else if (Depth(value) > budget - 1) node.Remove(key);
```
Avick3110: The final `else` deletes any member the three hand-maintained sets do not name once it is too deep, and more than annotations reach it. Two schema-bearing shapes land here:

* `items` in its tuple spelling (`"items": [ … ]`) fails `SubSchemas.Contains(key) && value is JsonObject sub`, is not in `SchemaLists`, and so is dropped whole rather than cut arm by arm.
* Any keyword outside the three sets whose value is a schema — `dependentSchemas`, `contentSchema`, `additionalItems` — is dropped the same way.

Dropping only widens, so nothing invalid is published, but a constraint vanishes from the surface with no signal, and these sets are exactly the hand-written per-shape coverage the design says not to carry. At minimum, assert in a test that the served surface carries no member name outside the three sets plus the known annotations, so a new SDK keyword fails loudly instead of being deleted.

src/housecarl-mcp-tests/PublishedSchemaDepthTests.cs
```diff
@@ -0,0 +1,308 @@
...
+{
+    const int Cap = 10;
+
+    /// <summary>The shared server — the plain environment, which is what "unset" means.</summary>
+    readonly ServerFixture _plain;
+
+    /// <summary>This class's own server, started with the cap. Never the shared one: the variable it sets is the
+    /// subject, and every other stdio test reads the uncapped surface.</summary>
+    readonly ServerFixture _capped =
+        new(new Dictionary<string, string> { [SchemaDepthCap.Variable] = Cap.ToString() });
```
Avick3110: `_capped` is a field initializer, so it runs on every instance of this class — and xUnit builds a new instance per test. That is a real `housecarl-mcp.exe` process, a fresh temp data dir, an `initialize` and a `tools/list`, twelve times (seven `[Fact]`s plus five `[InlineData]` rows), all inside `[Collection("server")]` so they run one after another. Five of those twelve — the `AnInvalidDepthIsRefusedAtStartupInOneSentence` rows — never touch `_capped` at all; they start and tear down a whole capped server to assert on a process `Boot` starts itself.

Moving the capped server into its own `IClassFixture` (or behind a `Lazy<ServerFixture>` created on first use) starts it once and leaves the invalid-value rows paying nothing.

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Folded in 9889e1f1. `Read` now refuses below 3 and says why: a schema root is level 1, `properties` level 2, each parameter level 3, so 3 is the shallowest cap that leaves every root carrying its own parameters. Two tests on a server at that cap: every tool still publishes a non-empty top-level `properties`, and every tool with a required parameter still refuses `{}` with houseCARL's own named sentence rather than the SDK generic error (population taken off the pre-flatten surface, same as the shim tests). The PR text and the doc paragraph are corrected to say the shim reads the top-level `properties` and that the floor exists to protect it.

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Folded in 9889e1f1, as a wording ADDITION on the cut path only — flagging it because Aaron ruled the terminator identical to the recursion one. The node still reuses `ToolSchemas.Terminator`, so the claim is unchanged ("nesting continues below this level with the same shape shown above; it is accepted but not spelled out again here"); the cut appends one sentence after it: "Nesting was cut here at depth N by HOUSECARL_MAX_SCHEMA_DEPTH: the tool accepts the full shape, and this parameter's description carries its members." Nothing the node accepts or claims moves, and the recursion bound's own terminator is untouched — a test asserts the clause appears on every node the capped surface closed, names the configured depth, and appears nowhere on the uncapped surface. Aaron's call at his gate whether the addition stays.

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Folded in 9889e1f1. `Cut` re-measures the finished document and throws when it is still over the cap; `PublishSchemas` catches that, writes the one sentence (naming the variable, the tool and the depth reached) to stderr and exits 1, so it is a refusal at startup rather than a host stack trace. The two holes you named are also closed at their own site: a non-object value in a dictionary or a schema list that is too deep to keep is now a named refusal instead of being skipped.

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Folded in 9889e1f1, and further than suggested: the silent-delete branch is gone. `items` in the tuple spelling is now read off the VALUE, not the name, and cut arm by arm; `dependentSchemas` joins the dictionaries; `additionalItems` and `contentSchema` join the sub-schemas. Anything else too deep to keep is a startup refusal naming the member and the tool, not a drop — these schemas are generated from a closed set of shapes, so a member reaching there is drift to report rather than a case to widen at a user's server start. That makes the test you asked for dynamic and total: the class's minimum-cap server boots at 3, which cuts every branch of all 31 schemas, so an unhandled member anywhere on the surface fails the fixture instead of vanishing.

src/housecarl-mcp-tests/PublishedSchemaDepthTests.cs
Avick3110: Folded in 9889e1f1. The capped servers moved into a `CappedSchemaServers` class fixture (the one at 10 and the one at the minimum, 3), started once for the class; the invalid-value theory rows touch neither and pay only for the process `Boot` starts itself. The whole class, now 35 tests including the new minimum-cap rows, runs in under a second.

src/housecarl-mcp/SchemaDepthCap.cs
```diff
@@ -0,0 +1,200 @@
...
+    /// HOUSECARL_SETUP_HOME).</summary>
+    internal const string Variable = "HOUSECARL_MAX_SCHEMA_DEPTH";
+
+    /// <summary>The shallowest cap that still publishes a tool's own parameters. A schema's root is level 1, its
+    /// <c>properties</c> dictionary level 2 and each parameter level 3, so below 3 the root itself is the node
+    /// that gets closed — and <see cref="ToolCallShim"/> READS a schema's top-level <c>properties</c>, so a root
+    /// without one silently drops argument coercion, the named missing-parameter refusal and the undeclared-key
+    /// refusal. A cut is allowed to say less about a nested shape; it is not allowed to change what a call
+    /// gets back.</summary>
+    internal const int Minimum = 3;
```
Avick3110: **Cap 3 strips every nullable parameter's `type`, and the shim reads it.**

At cap 3 each parameter node gets a budget of 1. Every parameter whose `type` is the exporter's list form (`["array","null"]`, `["string","null"]` — most of them) is depth 2, so `Shrink` returns a Terminator, and `Terminator` (line 188) then removes `type` because the list needs a level it does not have. But `ToolCallShim.DeclaredTypes` reads exactly that member, so at cap 3 argument coercion, the typed-mismatch refusal and the `in_place`-names-a-file refusal all silently stop — the degraded mode the floor exists to rule out, and the PR's "tools/call is untouched at every accepted cap" does not hold.

Concrete: `HOUSECARL_MAX_SCHEMA_DEPTH=3`, call `housecarl_forward` with `formids: "013989:Skyrim.esm"` (bare string). Uncut, `Coerce` wraps it into an array and the call runs; cut, `formids` publishes as `{description}` only, `DeclaredTypes` is empty, nothing wraps and the SDK binder fails. Same for `in_place: true`, which is no longer refused by name.

The shallowest-cap tests check only that `properties` exists and that `required` still refuses — neither reads a parameter's `type`, which is why they stay green.

Fix: `Minimum = 4` — a parameter's `type` list sits at level 4, and at budget 2 every parameter keeps it (`items`/`properties`/`anyOf` below it are cut, `type` and `enum` fit). Update the two prose sites (README, CHANGELOG) and add a shallowest-cap test asserting every parameter that carried a `type` uncapped still carries one. Alternatively keep 3 and have `Terminator` collapse the list to its first non-null scalar rather than dropping it, but that changes what `DeclaredTypes` reports for `null`.

Avick3110: Folded in c7f5a7d7, as `Minimum = 4` — I did not take the collapse-the-list alternative, for the reason you gave: it changes what `DeclaredTypes` reports for null, so the cut would be buying depth with call behaviour, which is the thing the floor exists to forbid. At 4 a parameter node has budget 2, its `type` list fits, and only the shapes below a parameter are cut; verified on the wire at cap 4, zero of the 31 tools loses or changes a parameter `type` against the uncapped surface. Two tests at the floor: every parameter that carried a `type` uncut still carries the identical one, and `housecarl_forward` with `formids: "013989:Skyrim.esm"` as a bare string is still wrapped and reaches the body. `3` joins the refused rows of the invalid-value theory. README, CHANGELOG, the architecture note and the PR body now say 4 and say which two members the shim reads.

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Residue folded in c7f5a7d7. My earlier fold kept your wording concern half-open: the node still carried the bound's "the same shape shown above" with a sentence appended. It is now a second constant beside the recursion one, exactly as you allowed — `ToolSchemas.RecursionContinues` for the bound, `SchemaDepthCap.CutContinues(cap)` for the cut, same node and same claim, and the cut's reads: "Nesting continues below this level and is accepted, but is not spelled out in this document: it was cut at depth N by HOUSECARL_MAX_SCHEMA_DEPTH. The tool accepts the full shape, and the members are named in the description of this parameter." The test now asserts the bound's wording appears nowhere on the capped surface and the cut's nowhere on the uncapped one.

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Residue folded in c7f5a7d7: the test you asked for at minimum now exists as a static one. It walks the served surface at schema positions only (never a dictionary's keys, which are parameter and definition names) across the uncapped, cap-10 and floor servers, collects every member name, and fails naming anything outside the cut's keyword sets plus the annotations that hold no schema. So a keyword a future SDK starts emitting reddens here, at a depth today's cut happens not to reach, instead of reaching a user as a startup refusal. The startup refusal stays as the runtime backstop.

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Re-checked against your wording after the rebase: folded at 9889e1f1, and the floor moved again in c7f5a7d7 — your 14:07 comment showed 3 was still one level short, since `DeclaredTypes` reads a parameter's `type` and at 3 that list has no room. `Minimum = 4` now, covered by four tests at the floor (properties present, missing-required refused by name, every parameter type preserved, bare-string coercion still wrapping).

src/housecarl-mcp/SchemaDepthCap.cs
Avick3110: Re-checked against your wording: folded at 9889e1f1, unchanged since. `Cut` re-measures the finished document and refuses over the cap, in one sentence on stderr naming the variable, the tool and the depth reached; the two skipped positions you named (a non-object in a dictionary or a schema list) are named refusals at their own site rather than skips.

src/housecarl-mcp-tests/PublishedSchemaDepthTests.cs
Avick3110: Re-checked against your wording: folded at 9889e1f1, unchanged since. `CappedSchemaServers` is the class fixture, holding the cap-10 server and the floor server; the invalid-value rows touch neither. The class is 39 tests now and runs in about 0.8s.

Issue #721, opened 2026-09-11
[GAP] tree and delta cost 0.24 s per row over a REFR scan and limit= does not cut the work; the 20-minute read reproduced


Avick3110:
Measured 2026-09-11 on a 3,801-plugin order (8C/16T, 32 GB), the installed 2.0.0 server in its own process.

The reporter's 20-minute `records types=["REFR"] project={form:"tree"}` read is reproduced, and it is **time, not memory**: peak working set 1.3 GB throughout. Each row of the comparison forms costs a fixed serial amount, and `limit=` does not reduce it.

| call over `Nature of the Wild Lands.esp` REFRs | rows | elapsed | per row | peak WS |
|---|---:|---:|---:|---:|
| summary (flat scan), to_file | 17,727 | **0.14 s** | — | 742 MB |
| **tree, to_file** | 17,727 | **900 s, aborted** (~70 min predicted) | — | 1.3 GB |
| tree, 100 ids | 100 | 31.4 s | 0.31 s | 1.2 GB |
| tree, 1,000 ids | 1,000 | 235 s | 0.24 s | 1.2 GB |
| tree, 1,000 ids, `limit=10` | 1,000 | 230 s | 0.23 s | 1.2 GB — limit does not cut the work |
| delta vs previous_provider, 1,000 ids | 1,000 | 190 s | 0.19 s | 1.5 GB |
| fields (Base, Position, *parent.EditorID), 1,000 ids | 1,000 | 33.6 s | **33 ms per body read** | 1.3 GB |
| everything, 1,000 ids | 1,000 | 32.4 s | 33 ms | 1.1 GB |
| tree over 194 REFRs on a profile with missing masters | 194 | 87 s | 0.45 s | 1.3 GB |

CPU sat at ~1.2 cores for the whole 15 minutes. Per tree row the server opens a fresh overlay session, resolves every provider of the record, and reads every provider's body to diff depth (`LoadOrderService.ResolveTreePinned`); at ~33 ms per body on this order and several providers per contested REFR, that is the 0.24–0.31 s.

Two facts to act on:

1. **33 ms per single body read** is the floor every per-row form pays. The flat scan over the same 17,727 records streams from the index in 0.14 s, so the cost is the per-row session/overlay open, not the record. This is the same pathology #672 observed on point reads (42–111 s there, unreproduced; here it is a steady 33 ms).
[...]

Issue #672, opened 2026-09-08
[NOTE] Single-record point reads observed at 42 to 111 seconds where a 66,856-row scan took 5; did not reproduce


Avick3110:
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/bulk-record-jobs/VERDICT.md`, `dev/skill-validation/gate-2026-09/mutagen-reference/VERDICT.md`).

Four arms saw a single-FormID `housecarl_records` read take 42,529 ms (on-order) and 74,000 to 135,000 ms (off-order) while a whole-order 66,856-row scan took 4,839 ms. The re-judge could not reproduce it (106 ms and 84 ms later in the session), so it looks like a cold index build charged to the first small read. Filed as a note: if the first read after startup pays an index build, the response should say so ("built the record index, N ms") rather than look like a slow read.

Avick3110:
Related measurement 2026-09-11: on a 3,801-plugin order every per-row body read costs a steady ~33 ms (fields and everything forms over 1,000 REFR ids: 32–34 s), while a flat scan of the same 17,727 records streams in 0.14 s. Not the 42–111 s seen here, but the same lane; the timed investigation this issue asks for has a baseline now. Details on #721.

Avick3110:
Folded into #721 per the backlog ruling of 2026-09-11. PR #764 measured the per-row cost: the suspect (a fresh overlay session per row) is not the cause; the 33 ms floor is the per-provider body seek, now tracked as #765.

PR #739, opened 2026-09-14
Release a parent predicate's body when its candidate ends

### LINKED ISSUE #720: [Bug] a where= predicate leading with *parent costs ~0.8 MB per candidate: 15 GB in 11 s over Skyrim.esm's REFRs
Measured 2026-09-11 on a 3,801-plugin order (8C/16T, 32 GB), the installed 2.0.0 server in its own process, working set sampled every 250 ms.

A `where=` predicate that leads with the containment step `*parent` costs roughly **0.8 MB per candidate evaluated**, and the memory is held for the call. Every other predicate shape over the same scopes is flat.

| where= | candidates | elapsed | peak working set | result |
|---|---:|---:|---:|---|
| `*parent.EditorID = X` | 194 REFRs | 1.4 s | +150 MB | 0 matches |
| `*parent.EditorID = X` | 17,727 REFRs (one plugin) | **21 s, killed** | **12.0 GB** | — |
| `*parent.EditorID = X` | 693,333 REFRs (Skyrim.esm) | **10.6 s, killed** | **15.3 GB** (~1.4 GB/s) | — |
| `editorid contains X` | 693,333 REFRs | 0.83 s | 968 MB | 383 |
| `Base->editorid contains X` | 17,727 REFRs | 2.0 s | 1.1 GB | 687 |
| `Temporary[*count] > 200` | all CELL | 2.4 s | 1.2 GB | 230 |

All `counts_only`, so no render is involved; this is the predicate itself. The same `*parent` step under the `fields` form costs ~17 ms per row per hop and no memory, and a `*parent` *walk* over 1,000 REFR seeds is index-only (0.12 s, no rise) — so the containment index answers the question cheaply and the predicate path is doing something else per candidate. It looks like each evaluation materialises the parent (or the parent chain) and keeps it.

Any `*parent` where= over a REFR-sized scope will take a 32 GB machine down in under half a minute, which is the second reproduction of the 44 GB shape (the first is the unbounded forward walk, filed separately).

Avick3110:
A `where=` predicate that leads with the containment step `*parent` shared the `->` link step's target cache, so every containing record it hopped to stayed held for the whole call. A Mutagen record getter is a `ReadOnlyMemorySlice<byte>` over the whole GRUP it was read from and keeps that array alive — the same mechanism as #719 — and `types=` bounds the CHILD population, not the parent one, so `types=["PlacedObject"] where=["*parent.EditorID = X"]` retained one pinned whole-GRUP array per distinct CELL, measured at ~0.8 MB per parent on the battery order and enough to take a 32 GB machine down over a REFR-sized scope. The hop now climbs the containment map in KEYS — `ParentOf` answers from a key, so a record in the middle of a `*parent.*parent` chain is never read at all — fetches a body only for the record the terms are actually read on, through the same bound view the `->` step resolves through, and releases that body when the candidate's evaluation returns. What carries across candidates is the VERDICT, not the body: every child under one containing record decides the same way for the same predicate, so one memo entry buys what the cached getter used to buy, at a bool's cost. The memo is keyed by the predicate and by which side hopped, because a link predicate can hop on its left path and its right one and those are different questions about the same containing record.

**One deliberate behaviour change.** An INTERMEDIATE record on a chain is no longer fetched, so it can no longer report `Unreadable` for a body that would not fetch: the climb only needs its key, and the hop now reports `Unreadable` for the record the terms run on. A record that nothing contains is still the same named `NoParent` no-verdict, naming the child-bearing property set; where a chain dead-ends ABOVE the first hop the type name for that sentence is fetched lazily, on the miss alone, and where even that will not fetch the sentence says "these records" rather than naming a type. No surface text changed — no parameter description, no refusal wording, no output shape.

## Before and after

Private `housecarl-mcp.exe` built from this branch and from `main` (d7795f38), neither the installed server, one cold process per arm, `HOUSECARL_DATA_DIR` pointed at a scratch config so the installed server's own config was untouched. Working set and private bytes sampled every 250 ms from outside, max inside the call's window, 35 s of idle sampling after the call returned. Abort valves: 10 GB working set, 900 s. Order: `E:\Authoria - Requiem Reforged`, profile `houseCARL Test Bench Profile`, 3,254 plugins (3,244 checked), 2,544,804 containment edges, epoch `e2-9a563d2a2b725b4f`, on 8C/16T and 32 GB.

This is NOT the battery's order — 3,254 plugins against 3,801, and the battery's "3 orphaned NOTWL patches" scope does not exist here — so the issue's numbers are a third column for shape, not a row-for-row comparison, and `main` was re-run here for every arm. Every arm is `counts_only`, so no render is involved. X is `RiverwoodSleepingGiantInn`, a real Skyrim.esm cell.

| arm (`types=["PlacedObject"]`, `counts_only`) | candidates | `main`, here | this branch, here | battery order (#720) |
|---|---:|---|---|---|
| `*parent.EditorID = X`, the 27 NOTWL patch plugins | 541 | 2,359 MB peak, **2,358 MB still held 35 s later**, 20.2 s, 0 matches | **950 MB peak, 814 MB after**, 21.5 s, 0 matches | 194 candidates: +150 MB, 1.4 s |
| `*parent.EditorID = X`, `Nature of the Wild Lands.esp` | 17,727 | **killed at the 10 GB valve**, 10,248 MB in 40.6 s | **1,264 MB peak, 994 MB after**, 47.5 s, 0 matches, completes | 12,038 MB in 21 s, killed |
| `*parent.EditorID = X`, `Skyrim.esm` | 693,333 | **killed at the 10 GB valve**, 10,315 MB in 22.5 s | **flat at 1,561 MB**; killed at the 900 s valve, still running | 15,333 MB in 10.6 s, killed |
| control `editorid contains …`, `Skyrim.esm` | 693,333 | 901 MB, 13.3 s, 0 matches | 876 MB, 13.8 s, 0 matches | 968 MB, 0.83 s |
| parity `*parent.EditorID = X`, `…Ivy Riverwood Overhaul.esp` | — | 905 MB, 12.0 s, **1 match** | 844 MB, 12.0 s, **1 match** | — |

The retention is gone on every arm: the 541-candidate arm drops from 2,358 MB still resident 35 s after the call to 814 MB, and the two arms that used to eat the machine no longer rise — the 17,727 one now finishes inside 1.3 GB. The control moved by 25 MB and 0.5 s, inside run-to-run noise, so nothing else changed. The parity arm returns the same single match on both builds.

**What this does not fix, and is not pretending to.** Arm 3 is now bounded by the clock rather than by memory: the hop reads a containing record's body with an UNTYPED whole-plugin seek (`LoadOrderResolver.SeekBody` with no getter type), one per distinct parent, and Skyrim.esm's REFRs sit under tens of thousands of distinct cells. That is the same per-body-read cost as #721/#672 and it belongs there, not here — this change is the live set. The memo is what keeps it to one seek per parent per call rather than one per candidate.

**The clock figures above are not all seeks** — the review caught a second contributor this branch had added itself: the held-bodies counter walked the whole fetched set on every memo miss, quadratic in distinct parents and running in production rather than only under test. It is a plain depth now. Re-running arm 2 on the folded build, the same call, same order: **47.5 s becomes 41.1 s**, peak 1,264 MB becomes 1,081 MB and what is left 35 s later drops from 994 MB to 742 MB. So roughly 6 s of that arm was instrumentation and the remaining 41 s is the seeks. Arm 3's own share is bigger — that quadratic grows with the parent population, and Skyrim.esm's is the largest here — so its 900 s is a ceiling on the seek cost, not a measurement of it.

## Test

`src/housecarl-mcp-tests/WhereContainmentCostTests.cs`, five tests on three new internal counters on the predicate set — `ParentBodiesHeld`, `ParentBodyHighWater`, `ParentBodyFetches` — in the idiom the walk's cost tests and `LoadOrderResolver.BodySeeks` already use: whether the hop released a containing record's body or kept one per distinct parent is invisible in the answer and visible only in memory, which is the argument those counters are added on. `ParentBodiesHeld` counts parent bodies actually reachable from the set (the one in flight, plus any that ended up in the target cache), so it rises again if a future change starts caching them rather than asserting a constant.

- `AParentPredicateHoldsNoParentBodyPastTheCandidateThatReadIt` — 200 candidates under 200 distinct parents: nothing held at return, never more than one at once, 200 bodies read.
- `AParentPredicateReadsOneBodyPerContainingRecordNotOnePerCandidate` — 200 candidates under 4 parents: 4 reads, not 200, so the memo keeps the read-once-per-call guarantee the body cache used to buy.
- `AParentChainReadsOnlyTheRecordItsTermsRunOn` — `*parent.*parent` over 50 placed references: one body read, the worldspace, and not one cell body on the way up.
- `TwoClausesHoppingToOneParentKeepTheirOwnVerdicts` and `AHopOnEachSideOfALinkStepKeepsItsOwnVerdict` — the two ways one memo entry could answer the wrong question.

Verified against the old behaviour by reinstating the one cache line: the first three fail (held 200 / 4 / 1 instead of 0). Full suite 2,000 passed, `ci-all` ALL PASS, both after the rebase onto `d7795f38`.

Closes #720


src/housecarl-core/FieldPredicate.cs
```diff
@@ -1083,35 +1122,67 @@ EvalKind ClassifyMiss(string note)
...
-            var pk = _parentOf(body.FormKey);
-            if (pk is null) { _lastNoParent = RecordNaming.StripOverlay(body.GetType().Name); return (null, EvalKind.NoParent); }
-            if (!_targetCache.TryGetValue(pk.Value, out var parent))
-                _targetCache[pk.Value] = parent = _fetchWinnerBody(pk.Value);
-            if (parent is null) return (null, EvalKind.Unreadable);   // the parent is indexed but its body would not fetch
-            body = parent;
+            var pk = _parentOf(at);
+            if (pk is null)
+            {
+                var of = i == 0 ? body : _fetchWinnerBody(at);
```
Avick3110: The chain's `NoParent`-above-the-first-hop path re-fetches a body **per candidate**, not per parent — the one place the memo does not cover, and the work is thrown away after the first one.

`ClimbToParentKey` returns before `EvalAtParent` ever reaches `_parentVerdicts`, so this `_fetchWinnerBody(at)` runs for every single candidate that dead-ends here. The value it produces is consumed only by `_noParentWhat[k] ??= _lastNoParent` — a `??=`, so candidate #2 onward pay an untyped whole-plugin seek for a string that is immediately discarded.

Concrete: `types=["PlacedObject"] where=["*parent.*parent.EditorID = X"]` over a scope holding interior REFRs. An interior CELL has no containing worldspace, so every interior candidate hits `i == 1`, `pk is null`, and seeks its cell's body. Twenty thousand interior REFRs under four thousand cells = 20,000 seeks where `main` did 4,000 (it cached the parent in `_targetCache`). The PR's stated guarantee — "a containing record is still read once per call however many children sit under it" — holds on the match path and is lost on this one; the exact `#721`/`#672` per-body-read cost the description says the memo keeps to one seek per parent.

Cheapest fix is to memo the miss too, e.g. cache `_lastNoParent` per dead-end key (or skip the fetch entirely once `_noParentWhat[k]` is non-null for the predicate in hand).

src/housecarl-core/FieldPredicate.cs
```diff
@@ -1083,35 +1122,67 @@ EvalKind ClassifyMiss(string note)
...
+        (bool Satisfied, EvalKind Kind) verdict;
+        if (parent is null)
+            verdict = (false, EvalKind.Unreadable);              // the parent is indexed but its body would not fetch
+        else
         {
-            var (hopped, miss) = HopToParent(body, p.LinkParentHops);
-            if (miss is { } m) return (false, m);
-            body = hopped!;
+            _parentInFlight = parent;
+            if (ParentBodiesHeld > ParentBodyHighWater) ParentBodyHighWater = ParentBodiesHeld;
```
Avick3110: `ParentBodiesHeld` is O(`_parentsFetched.Count`) — it walks the whole fetched set on every call — and this line calls it once per memo miss, i.e. once per distinct parent, in **production**, not only under test. Total work is O(P^2) in the number of distinct containing records.

Arm 3 of the table is exactly the shape that pays: Skyrim.esm's REFRs sit under tens of thousands of distinct cells. At P = 50,000 that is ~1.25e9 `HashSet` enumerations plus `_targetCache.ContainsKey` probes, added to a call the table already reports as clock-bound and killed at the 900 s valve. The "what this does not fix" paragraph attributes that entirely to the untyped seeks; this quadratic is a second contributor introduced by the same commit and is pure instrumentation overhead — no answer depends on it.

The loop is also only ever counting parents that ended up in `_targetCache`, which the hop no longer writes to. Suggest maintaining the high-water incrementally (a counter incremented/decremented alongside `_parentInFlight`, or an `_targetCache`-overlap count updated when the link step inserts) so the hot path stays O(1) per parent.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -1083,35 +1122,67 @@ EvalKind ClassifyMiss(string note)
...
+            try { verdict = below(parent); }
+            finally { _parentInFlight = null; }
```
Avick3110: `EvalAtParent` nests, and this `finally` clears `_parentInFlight` unconditionally — so the inner hop erases the outer hop's record of a body that is still alive on the stack.

`EvalAtParent(linkSide: true)` runs `EvalLinkPath(p, parent, 0)`, which resolves each link target and calls `EvalCore(p, target)` (line 1234); for a predicate with `ParentHops > 0` that re-enters `EvalAtParent(linkSide: false)`. So on `*parent.Quest->*parent.EditorID = X` — the grammar `AHopOnEachSideOfALinkStepKeepsItsOwnVerdict` exercises — two parent GRUP-backed getters are genuinely live at once, and the counters say one:

- the inner `finally` sets `_parentInFlight = null` while the outer parent is still held, so `ParentBodiesHeld` reports 0 for the rest of the outer evaluation;
- `ParentBodyHighWater` never rises above 1, because the inner call computes `held = 1` (the outer parent is tracked nowhere else — `_parentsFetched` only counts it if it is in `_targetCache`, which the hop no longer populates).

These counters are the whole evidence the PR rests on ("invisible in the answer and visible only in memory"), and on the one shape where the real high-water is 2 they under-report. No existing test asserts counters on a two-sided hop, so this passes silently.

Fix wants both halves: save and restore the previous `_parentInFlight` in the `finally` rather than nulling it, and make `ParentBodiesHeld` count the in-flight chain (a depth counter or a small stack) instead of a single slot.

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded in e0c0374. The dead-end type name is now remembered per key (`_noParentTypes`, read through `NoParentTypeOf`), so a chain that dead-ends above its first hop reads each such record once per call and every path obeys the same read-once rule. The first hop still needs no read at all — the candidate body is in hand.

New test `AChainThatDeadEndsAboveItsFirstHopReadsEachDeadEndOnce`: 200 placed references under 50 parentless cells, `*parent.*parent.EditorID = X`, asserting 50 reads and that the rollup sentence still names the type. It fails on the pre-fold code with 200 (verified by disabling the memo lookup). The fetch counter it asserts on is now every read the hop pays rather than a distinct-parent count, so a read the memo should have saved cannot hide behind it — which is also why `TwoClausesHoppingToOneParentKeepTheirOwnVerdicts` now expects 2 (one per predicate over one parent, not one per candidate).

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded in e0c0374, and you were right that the note was wrong about it. `ParentBodiesHeld` is now a plain `int` maintained as the hops nest, so the hot path is O(1) per parent and nothing walks a set; the dead `_targetCache` overlap loop is gone with it, since the hop no longer writes there.

Measured rather than argued, on arm 2 (17,727 REFRs, the one that completes), same order and harness: **47.5 s → 41.1 s**, peak 1,264 MB → 1,081 MB, and what is still resident 35 s later 994 MB → 742 MB. So about 6 s of that arm was the counter and the remaining 41 s is the untyped seeks. Arm 3 could not be re-timed — it is still killed at the 900 s valve — but its parent population is the largest here, so its share of the quadratic was larger too; the PR body now says the 900 s is a ceiling on the seek cost rather than a measurement of it, and carries the arm 2 split.

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded in e0c0374. In-flight is a depth now, not a slot: `++_parentsInFlight` before the nested evaluation and `--` in the `finally`, so an inner hop can no longer erase the outer one, and the high water follows the deepest hop chain instead of stopping at one.

`AHopOnEachSideOfALinkStepKeepsItsOwnVerdict` now asserts the counters on exactly that grammar — `ParentBodyHighWater == 2` on `*parent.Quest->*parent.EditorID`, and `ParentBodiesHeld == 0` at return, so both bodies are released with the candidate. Verified it fails on the pre-fold single-slot version (high water 1). The two bodies are held by the evaluation stack itself, which is what the depth counts.

PR #741, opened 2026-09-14
Report the cycles a forward walk actually found

### LINKED ISSUE #726: [Bug] the chain form reports cycles: 0 on a real A<->B cycle
Found 2026-09-11 while testing whether cyclic references drive unbounded traversal (they do not).

A throwaway plugin on a scratch profile carried two leveled items: `A` whose entries point at `A` and at `B`, and `B` whose entries point at `A`. A forward `walk=` from `A` under `project.form="chain"` with `depth:1000, max_nodes:100000` terminates correctly in 0.06 s reaching 2 nodes — and reports **`cycles: 0`**.

The chain form's description says it renders "the walk's own paths, endpoints and cycles". A self-reference and a two-node loop are the simplest cycles there are, and neither is surfaced. Either the cycle detection is not wired to the report, or it only counts something narrower than a revisited node — in which case the description over-promises.

Not a memory or time issue: the walk is bounded by its visited set and stops. It is a wrong answer, quietly.

Avick3110:
A forward `walk=` under `project.form="chain"` reported `cycles: 0` on a plugin that is nothing but cycles. The cause was not a broken detector: the forward walk only ever recorded a cycle when `walk.follow` named a single chain, on the reasoning that such a walk is a linear chain so any revisit is a loop. A closure walk (no `follow=`, the default) dropped every revisit on its visited set and said nothing at all, which is why a leveled item pointing at itself and a pair pointing at each other both came back with no cycles.

The fix is the one the copy walk already uses. The forward walk now records the edges it crosses — every link off a node it entered, whether or not the target turns out to be new — and the cycles are found from that edge set once it is done, by a depth-first colouring where an edge into an on-stack node is a back edge. That is also what tells a real loop apart from two paths meeting on one record, which a count of revisits cannot do: a diamond is a revisit and is not a cycle. The finder moved out of `ClosureWalk` into `GraphCycles` in `housecarl-core`, so both walks answer the question the same way and `ClosureWalk` keeps only the labelling. A seed's edge set is released where its visited set and frontier already go, so a batch holds one seed's edges at a time rather than every seed's until the last one finishes.

The chain form now renders each cycle under its seed as the loop itself — `A (…) -> B (…) -> A (…)`, naming the record the loop closes on at both ends — and carries a `cycles` count beside `seeds` and `reached`, so `counts_only` and the JSON envelope state it too.

The cycle search is capped at 200 loops per seed. A strongly connected region of n entered nodes carries up to n-squared back edges, each holding a path of up to n keys, so an uncapped search is quadratic in the region rather than linear in `walk.max_nodes` — an OOM on a read. A seed that hit the cap says so, in the text tail, in a `cycles_capped` json flag and in a header line counting them, so a stopped search never passes as a finished one. A reading form never looks at the cycles, so the walk now records no edges and settles none unless `form="chain"` will render them. Both renders hold the cycle list to `max_chars` and name how many they held back, and a json row that elides inside itself now sets the response's `truncated` flag so the ceiling spill writes the complete walk to an artifact instead of returning past `max_chars` with nothing said.

The count carries its bounds and they are stated wherever it is. It is found over the nodes the walk actually ENTERED, so one closing past `walk.depth` or `walk.max_nodes` is not yet visible. And it is one per closing link, which makes it a lower bound on how many distinct loops there are — `A -> B`, `A -> C`, `B -> A`, `C -> B` carries two loops and reports the one back edge. Over a walk that FINISHED, none reported does mean none, since a graph with a loop always has a back edge; over one cut at `walk.depth` or `walk.max_nodes` it means only that none was found in what was read, and the response says so beside the count — on `counts_only` too, where no seed row carries the cap note and zero is exactly the answer that reads as proof. Enumerating every simple cycle is exponential in the graph, which is not a cost a read may pay. The cycle list is also the one part of a seed's tail that nothing bounds, so it is laid to its own room in the render and says how many it held back, rather than being reserved whole beside every node line and pricing the seed's own nodes out of the response.

Closes #726

**Verified.** A new fixture world builds exactly the issue's plugin — a leveled item A whose entries point at A and at B, and B whose entries point back at A — and asserts the rendered cycle paths (`A -> A` and `A -> B -> A`, two lines and no more) and the count (`cycles=2`). A third test walks an acyclic seed whose two branches meet on one shared record and asserts `cycles=0` with no cycle line, so a diamond cannot pass for a loop. A fourth walks a seed fanning over 12 form lists that each link all 12, at `max_chars=4000`, and asserts the node lines are there, the held-back line is there, and the seed is not rolled back out whole. A fifth asserts a `counts_only` response carries `cycles=2` and does not tell the reader to look under seeds it never prints. A sixth drives the same dense-fan call through the json lane and asserts `truncated` is set and the held-back entry is written; a seventh asserts a seed that hit the 200-cycle search cap says so; an eighth walks the dense fan at `depth=1` under `counts_only` and asserts `cycles=0` arrives with the cut sentence beside it. All fail before the change they cover.

Full suite 2009/2009 and `ci-all` 128/128 from the worktree, rebased on `origin/main`. One later full run failed `RecordsOwnedChildTests.Json_ATextRowsDeclarersBlockAloneCanTripMaxChars_AndTheResponseIsMarkedTruncated`, a spill-to-disk assertion in the tree form that nothing in this change touches; it passes on rerun and in its own run, and this machine had several build lanes testing in parallel. The existing walk-cost world also carries a dense cyclic fan (40 form lists each pointing at every one of them, itself included) walked through the chain form, and its cost and allocation gates still pass with the edge recording and the post-walk pass in place.

**Not run on a real install.** The plugin in the report was a throwaway on a scratch profile and is gone; the fixture reproduces it exactly, and no part of this change depends on load-order shape.


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3921,10 +3922,41 @@ sealed class WalkSeedState
         public string Type = "";
         public string Label = "";
         public List<WalkNodeRow> Nodes = new();
-        public List<string> Cycles = new();
+        /// <summary>The walked graph as edges, parent to target, per node this seed entered.</summary>
+        public Dictionary<FormKey, List<FormKey>> Edges = new();
```
Avick3110, line 3936: `Edges` is held for the whole batch call, unlike `Visited`/`Frontier`.

A few lines down (around 4319) a finished seed releases its bookkeeping — `if (st.Frontier.Count == 0) { st.Visited = new(); st.Frontier = new(); }` — which is the #719 retention discipline. `Edges` is not released there: it is only read at the very end, by `CyclesFound()` in the results loop, so every seed's edge set stays live until the last seed finishes.

And `Edges` is the biggest of the three. It is one `FormKey` per link CROSSED, not per node entered, with no dedup (`Edge` appends unconditionally), so a leveled list with 40 duplicate entries contributes 40 entries and a node with a 500-link FormList contributes 500. `max_nodes` bounds the node count; nothing bounds the fanout.

Concrete scenario: a `formids=` seed set spilled from an artifact — thousands of seeds, the case the seed loop is already sliced into `WalkPassRows` passes for — each walking a few hundred nodes with a normal fanout. Before this change the per-seed peak was released as each seed finished; now the sum of every seed's edges is resident at return. Same shape as the retention this walk was rewritten to remove.

The fix fits the existing release point: compute the seed's cycles where `Visited`/`Frontier` are dropped (the seed is finished and `Nodes` — all `CyclesFound` needs for labels — is complete), store the resulting strings, and null out `Edges` there.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -943,11 +943,16 @@ string RenderRev(SpillState? sp, out bool trunc) => json
                 Arm("winner (the walk expands the winner link graph)");
                 int reached = rows.Where(r => r.Error is null).Sum(r => r.Nodes.Count(n => !n.Status.StartsWith("no links")));
                 int errs = rows.Count(r => r.Error is not null);
+                // A cycle is a record the walk reached again from itself, counted over every seed and listed under
+                // the seed that reached it. It is a fact about the walked graph, so counts_only states it too.
+                int cycles = rows.Sum(r => r.Cycles.Count);
```
Avick3110, line 949: The cycle list is now unbounded, and `SeedTail` reserves it in full — a densely cyclic seed can render as nothing at all.

`SeedTail` (line ~2361) appends one `cycle: ` line per reported cycle, and `RenderRecordsChain` reserves that whole tail beside every node line:

```
if (sb.Length + line.Length + nodesCut.Length + tail.Length > budget)
{
    said = Said(sb, nodesCut, budget - tail.Length);
    mute = !said;
    break;
}
```

Before this change `tail` was always small: cycles only came from a named-follow walk, plus a one-line truncation note and the template report. Now the cycle count is O(back edges in the walked graph) and each line is the full loop, every record spelled `0x… (EditorID)`.

Concrete scenario: the dense cyclic fan the walk-cost world already carries — 40 form lists each linking all 40, itself included — walked through `project.form="chain"`. That graph yields on the order of several hundred back edges, each printing a path up to 40 records long, so `tail` alone runs to hundreds of KB. At default `max_chars` the first node line fails the check, `Said` is called with a negative budget and returns false, `mute` is set, and `Crossed(..., force: mute)` rolls the seed back out whole. The response is the header plus `... [rendered 0 of 1 seeds at max_chars=…]` — no nodes, no cycles, and not even the `nodesCut` notice that would name the remedy. Before this change the same call rendered its node list and cut it with a notice.

The tail needs its own bound the way the node loop has one: print cycles up to the room available and say how many were held back, rather than letting them evict the seed.

src/housecarl-core/GraphCycles.cs
```diff
@@ -0,0 +1,73 @@
...
+// record) from a genuine cycle (a record that reaches itself). Neither can the traversal TREE: a mutual
+// reference between two SIBLINGS is a real cycle whose nodes are never each other's tree ancestors. So
+// the walk records its edges and asks this afterwards, once, over the whole graph.
+//
+// Shared by every walk that keeps its edges, so one answer to "is this a cycle" serves all of them.
+
+/// <summary>Cycle finding over a walked link graph.</summary>
+public static class GraphCycles
+{
+    /// <summary>Every cycle in the recorded edge set. Each result is the loop itself, keys in order: it starts at the
```
Avick3110: "Every cycle in the recorded edge set" overstates what this returns, and the new user-facing `cycles=N` inherits the overstatement.

A colouring DFS finds every BACK EDGE, not every simple cycle. An edge into a node that is already `Finished` is skipped as "an ordinary diamond", but a finished node can still be on a cycle the DFS never enumerates.

Concrete graph: `A -> B`, `A -> C`, `B -> A`, `C -> B`. There are two simple cycles: `A -> B -> A` and `A -> C -> B -> A`. Rooting at A, the DFS goes A, B, takes the back edge `B -> A` and reports `A -> B -> A`, finishes B, then walks `A -> C`; `C -> B` lands on a FINISHED node and is skipped. One cycle is reported and the response says `cycles=1` and "1 cycle(s)". In Skyrim terms that is a leveled list A holding B and C where B holds A and C holds B — an ordinary mutually-recursive list pair with a shortcut entry.

No node on a loop is lost (each one appears in at least one reported cycle), so the answer is not wrong about WHETHER there is a cycle — but the count is a lower bound and the doc says otherwise. Given the errors-cornerstone, either say what it is ("one cycle per back edge — a lower bound on the number of distinct loops") in this summary, in `WalkSeedResult`, and in the `project.form` description, or enumerate properly.

The `project.form` text already bounds the answer by the nodes the walk entered; this second bound is not stated anywhere.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -943,11 +943,16 @@ string RenderRev(SpillState? sp, out bool trunc) => json
...
                 envelope.Add(new("walk", $"forward{(walk.follow is { } f2 ? $" follow={f2}" : " (closure)")} depth={walkDepth}"));
+                if (cycles > 0)
+                    headerLine += $"\n{cycles} cycle(s) — a record the walk reached again from itself; each one is listed under its seed.";
```
Avick3110: This sentence is appended to `headerLine` before the `counts_only` branch, so a `counts_only=true` call prints "…each one is listed under its seed" above a response that lists no seeds at all — the only other line is `seeds=… reached=… errors=… cycles=…`. The reader is told to look for something that is not there and has no way to know the remedy is to drop `counts_only`.

Either move the `headerLine +=` below the `counts_only` return, or word the counts_only case for what it is.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. `WalkSeedState.Settle()` now finds the seed's cycles and drops `Edges` at the point where `Visited` and `Frontier` are released — the seed is finished there and `Nodes` is complete, which is all the labels need. A seed that never ran a pass (no links off it at all) has not settled by then, so the results loop calls `Settle()` too; it is idempotent. The per-seed peak is released as each seed finishes again, as it was before.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. `SeedTail` takes a `cycleRoom` and lays the cycle block to it, replacing what does not fit with `... [N more cycle(s) held back at max_chars=… — raise max_chars, or to_file= for the complete walk]`. The room is half the budget, so the tail can no longer price the node list out; the cap note and the template report are bounded by their own shape and stay unconditional. New test: a seed fanning over 12 form lists that each link all 12, rendered at max_chars=4000 — its node lines are there, it says how many cycles were held back, and it is not rolled back out.

src/housecarl-core/GraphCycles.cs
Avick3110: Folded — and your graph is worse than you said: C is on a loop and appears in no reported cycle, so `no node on a loop is lost` does not hold either. `GraphCycles.Find` now says one loop per BACK EDGE, that a count is a lower bound on the distinct loops, that the loops returned are not a complete list of the records lying on one, and that NONE means the graph is acyclic — which is the claim that does hold, since a graph with a loop always has a back edge. Not enumerated properly because enumerating every simple cycle is exponential in the graph, which is not a cost a read may pay. The same bound is now in `WalkSeedResult`, in the `project.form` description, in the header sentence, and in the changelog.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. The `headerLine +=` moved below the `counts_only` return, so that sentence is said only where the seeds are listed; a counts_only response carries its `cycles=N` and nothing pointing at seeds it does not print. New test asserts the counts_only response contains `cycles=2` and not `listed under its seed`.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -956,7 +963,7 @@ string RenderRev(SpillState? sp, out bool trunc) => json
                     if (aerr is not null) return json ? JsonWire.RenderError(aerr, wEpoch) : "error: " + aerr;
                     spill = SpillState.Spilled(s!, manifestOnly: true);
                 }
-                var chainCounts = new[] { KvI("seeds", rows.Count), KvI("reached", reached), KvI("errors", errs) };
+                var chainCounts = new[] { KvI("seeds", rows.Count), KvI("reached", reached), KvI("errors", errs), KvI("cycles", cycles) };
                 string Render(SpillState? sp, out bool trunc) => json
                     ? JsonWire.RenderChain(winRows, max_chars, wEpoch, envelope, chainCounts, sp, out trunc)
```
Avick3110, line 978: The JSON chain render has no bound on the new cycle list. `JsonWire.WriteChainRow` checks `ms.Length >= cap` inside the `nodes` loop, but writes `cycles` whole (`if (row.Cycles.Count > 0) WriteStringArray(w, "cycles", row.Cycles)`, JsonWire.cs:904) with no check and no "held back" note. The text render got exactly this guard here (`SeedTail(..., budget/2, cap)`); the json lane did not.

Before this PR that array was only ever populated by a `follow=` chain walk (one short line per revisit), so it was effectively bounded. Now a closure walk over a densely cyclic graph emits one line per back edge, each the full loop path. The fixture's own 12-hub fan already produces a five-figure `cycles` array; at the default `walk.max_nodes=2000` a cyclic subgraph produces a multi-megabyte row.

Worse, `RenderChain` only sets `rowsTruncated` from the pre-row `ms.Length >= cap` check, so a single oversized row leaves `truncated == false` — the ceiling-spill fallback below (`if (spill is null && truncated)`) never fires, and the caller gets a response many times over `max_chars` with no artifact and no note that anything was elided. `ADenselyCyclicSeedStillRendersItsNodesAndSaysWhatItHeldBack` covers the text lane only; the same call with `json: true` is untested.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3921,10 +3923,55 @@ sealed class WalkSeedState
...
+                [FormIdToken.Of(Key)] = EditorId,
+            };
+            foreach (var n in Nodes) editorIds[n.Key] = n.EditorId;
+            string Label(FormKey k)
+            {
+                var token = FormIdToken.Of(k);
+                var editorId = editorIds.TryGetValue(token, out var e) ? e : null;
+                return $"{token} ({editorId ?? "<no editorid>"})";
+            }
+            return found.Select(path => string.Join(" -> ", path.Select(Label)) + " -> " + Label(path[0])).ToList();
```
Avick3110, line 3987: `CyclesFound` materialises every cycle, whole, before anything bounds it: `GraphCycles.Find` returns one `path.Skip(at).ToList()` per back edge, and this line then turns each into a full joined string. Nothing caps either list — the render-side `cycleRoom` and the JSON writer both run on the finished collection.

The cost is quadratic-to-cubic in the size of a cyclic region, not linear in the node cap. `GraphCycles.Find` reports one cycle per distinct `(from, to)` back-edge pair, so a strongly connected region of *n* entered nodes yields up to O(n²) back edges, each carrying a path slice of up to *n* keys and then a label string of ~40 chars per key. The existing walk-cost world's 40-node dense fan is ~1600 cycles of ~20 keys, which is fine; at the default `walk.max_nodes=2000` the same shape is millions of paths of a thousand keys each — an OOM on a read, which is the class of retention #719 was about, and the one release discipline `Settle()` is careful about for `Edges`.

Since the count is already documented as a lower bound and the render only ever shows the first few, a cap on how many cycles `Find` collects (and a flag saying it stopped) would keep the claim intact without the allocation. At minimum `WalkForwardBatch` should not pay this on the non-chain lanes, which record edges and settle cycles that no consumer ever reads.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -943,11 +943,18 @@ string RenderRev(SpillState? sp, out bool trunc) => json
...
+                // A cycle is a record the walk reached again from itself, counted over every seed and listed under
+                // the seed that reached it. It is a fact about the walked graph, so counts_only states it too.
+                int cycles = rows.Sum(r => r.Cycles.Count);
                 envelope.Add(new("walk", $"forward{(walk.follow is { } f2 ? $" follow={f2}" : " (closure)")} depth={walkDepth}"));
                 if (counts_only)
                     return json
-                        ? JsonWire.RenderNamedCounts(envelope, new[] { KvI("seeds", rows.Count), KvI("reached", reached), KvI("errors", errs) }, wEpoch)
-                        : Census($"{headerLine}\nseeds={rows.Count} reached={reached} errors={errs}" + Wire.EpochLine(wEpoch));
+                        ? JsonWire.RenderNamedCounts(envelope, new[] { KvI("seeds", rows.Count), KvI("reached", reached), KvI("errors", errs), KvI("cycles", cycles) }, wEpoch)
+                        : Census($"{headerLine}\nseeds={rows.Count} reached={reached} errors={errs} cycles={cycles}" + Wire.EpochLine(wEpoch));
```
Avick3110, line 963: `counts_only` prints `cycles=0` with nothing beside it when the walk was cut, and the tool description this PR rewrites tells the reader that means there are none: "no cycles reported means there are none".

Concretely: walk a seed whose closure exceeds `walk.max_nodes` (default 2000) and whose loop closes past the cut, or past `walk.depth` (default 16). Every seed carries `TruncationNote` explaining the cap, but this Census line renders only `seeds/reached/errors/cycles` — `row.TruncationNote` is never surfaced on the counts_only path, and the extra header sentence added just below is deliberately skipped here. So the response is `cycles=0` and, per the documented contract, acyclic — for a graph the walk simply did not finish reading. The `if (cycles > 0)` header does not help, because the wrong answer is precisely the zero.

The other lanes in this file already fold their cap into the header (see the `walk.max_nodes budget ... was reached` lines at 795/887). Either state the cut here (`cycles=0 (walk truncated — a loop closing past the cap is not visible)` when any row has a `TruncationNote`), or drop "none means none" from the counts_only wording.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. `WriteChainRow` now lays the cycles array to the same cap the nodes loop uses, closing with `[N more cycle(s) held back at max_chars — raise max_chars, or to_file= for the complete walk]`, and returns whether anything in the row was elided — the nodes loop's own silent elision included, which had the same hole. `RenderChain` ORs that into `rowsTruncated`, so an oversized row now trips the ceiling spill instead of returning past max_chars with no artifact. The artifact writer passes `int.MaxValue`, where both guards stay inert. New test drives the same dense-fan call with `format: \"json\"` and asserts `truncated == true` and the held-back entry.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded, both halves. `GraphCycles.Find` takes a limit and an `out bool capped`, and stops the whole search — not just the collecting — at `LoadOrderService.WalkCycleCap` (200), so the quadratic walk is never paid past what anything can render. The cap is reported: `WalkSeedResult.CyclesCapped`, a `cycles_capped` flag in json, a line in the seed's tail, and a header sentence counting the seeds that hit it — a stopped search never passes as a finished one. And `WalkForwardBatch` takes `wantCycles`, which `RecordsTools` sets only for `form=\"chain\"`: a reading form now records no edges at all and settles none, so it pays neither the allocation nor the retention. `ClosureWalk` is left uncapped on purpose (a one-line comment says why): the copy walk refuses at its node cap rather than truncating, so a partial cycle list would be the one degraded answer on a path that has none.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Folded. The chain lane now counts seeds carrying a `TruncationNote` and folds it into the header before the `counts_only` return, the way 795/887 do: `[!] N seed(s) stopped at walk.depth (D) or walk.max_nodes (M) — a loop closing past the cut is not visible, so this cycle count is not proof of none; raise the cap to finish the walk.` The json counts carry `truncated_seeds` beside `cycles`. The `project.form` text no longer says none means none flatly: it says so over a walk that FINISHED, and says that over a cut one zero means only that none was found in what was read. New test walks the dense fan at `depth=1` with `counts_only` and asserts both `cycles=0` and the cut sentence.

PR #742, opened 2026-09-14
Refuse a merge donor's missing master from the headers

### LINKED ISSUE #729: [GAP] merge refuses a donor with a missing master only after the 25 s identify pass; the header check takes 1.6 s
Measured 2026-09-11 on a throwaway MO2 instance with 224 plugins carrying missing masters (three masters unticked on a copy of a 3,801-plugin profile).

`housecarl_merge_plugins` over donors whose master is missing refuses plainly and writes nothing — the right answer — but only **after the ~25 s identify pass** has walked the whole order:

| call | elapsed | peak WS | result |
|---|---:|---:|---|
| merge 8 Lux Via patches, `Lux Via.esp` unticked | 26.2 s | 1.2 GB | refused, nothing written |
| merge 20 intact + the 8 orphaned | 28.2 s | 1.1 GB | same refusal, same wait |
| `housecarl_check` `missing_masters` only, counts | **1.6 s** | 736 MB | 224 missing masters |

The check that answers the question takes 1.6 s and reads the headers. The merge already knows its donor set before step 3; reading each donor's master list against the active order is the same header read and belongs before the identify pass, so the refusal arrives in two seconds rather than thirty.

Same family as #715 part 3 (the pre-flight should catch what the write later throws), and adjacent to #718.

Memory is unaffected by missing masters on every arm measured (index, scan, tree, merge).

Avick3110:
`housecarl_merge_plugins` refused a donor whose master is not active in the load order only at write time, after the identify pass had walked the whole order — the refusal that answers the question from the plugin headers arrived after tens of seconds. The merge already knows its donor set long before that pass, so the master union (a read of each donor's header, the same read `housecarl_check` `missing_masters` makes) now runs straight after the donors are validated and sorted, and each master is checked against the active order there. The refusal is the sentence the write already gave, naming the master to enable, and nothing is written. `MergeBuild` keeps its own check, because a master can go away between the pre-flight and the serialize.

The test fixture has a master installed but not activated and a donor that references it. It asserts the refusal names the master, and that the identify pass did not run — `RemapEngine.IdentifyPasses` is a counter on that pass, in the style of the resolver's existing `CollectPasses`/`BodySeeks` seams — with a control arm showing an intact donor still runs it. The counter arm fails on `main` and passes here.

Measured locally, since this needs a real order: a throwaway MO2 instance in the session scratchpad (a copy of `ModOrganizer.ini` naming a new profile, a junction to the live `mods` folder, and a copy of the live profile with `Lux Via.esp` unticked — the live instance was not modified), merging three Lux Via patches whose master is that unticked plugin. Before: 64.2 s to the refusal. After: 18.0 s and 18.8 s on two runs, same refusal, nothing written. The remaining time is the load-order index build over 3,245 plugins, not the identify pass; the issue's own before-number was 26.2 s on a quieter machine, and other lanes were building and testing on this one throughout, so read the pair, not the absolute figures.

`dotnet build`, `dotnet test --filter "tier!=bridge"` (2,004 passed) and `ci-all` (128/128) all green.

Closes #729


src/housecarl-mcp-tests/MergeMissingMasterPreflightTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        var w = new Weapon(new FormKey(key, 0x800), SkyrimRelease.SkyrimSE) { EditorID = key.Name + "Own" };
+        w.Template.SetTo(new FormKey(masterKey, 0x800));
+        m.Weapons.Add(w);
+        m.BeginWrite.ToPath(Path.Combine(dir, key.FileName.String))
+            .WithLoadOrder(new ISkyrimModGetter[] { master }).Write();
+    }
+
+    public void Dispose()
+    {
+        try { Directory.Delete(Root, recursive: true); } catch { /* best-effort */ }
```
Avick3110: The world never disposes `Svc`. Every other MO2-instance world that writes through the service (`AbsentMasterWorld`, `CheckErrorsWorld`, `RecordsWorld`) calls `Svc.Dispose()` before deleting `Root`, and this one runs a real merge that writes a plugin into the instance's `mods` folder. With the service still alive and holding its resolver/asset caches, the `Directory.Delete(Root, recursive: true)` can fail on Windows with a sharing violation — swallowed by the empty `catch`, so the run silently litters a temp instance per fixture construction and leaks the service for the rest of the process.

```suggestion
    public void Dispose()
    {
        Svc.Dispose();
        try { Directory.Delete(Root, recursive: true); } catch { /* best-effort */ }
    }
```

Avick3110: Folded in bb508394 — the world now disposes Svc before deleting Root, as MergeLightCarryWorld and the other MO2-instance worlds do. Test class and the full suite green after the change.

Issue #131, opened 2026-06-30
bulk_create/create_record omits DialogTopic SNAM subtype marker → guaranteed load CTD


matashina:
Hi, first of all, thank you @Avick3110 for houseCARL. These tools have made me so much happier and more productive while modding Skyrim. I wanted to report an issue I encountered a few times while creating some dialogue mods. I've encountered this behavior on versions 1.2.3 and 1.4 of houseCARL.

### Summary
When `bulk_create` / `create_record` makes a `DialogTopic` (DIAL) with a `Subtype` (e.g. `Hello`), it writes the `DATA` subtype enum but leaves the `SNAM` subtype **marker** as `00000000`. The engine buckets topics by that 4-character marker, so a null marker is a guaranteed access-violation CTD on load. This is one of the Creation-Kit bookkeeping jobs a byte-valid insert skips.

### Crash signature
EXCEPTION_ACCESS_VIOLATION walking the quest's topic list:

EXCEPTION_ACCESS_VIOLATION ... mov rdx, [rbx+0x20]   (rbx = garbage pointer)
RBP:    (TESQuest*)  "<your quest>.esp" {Stage=0, Active=false}
RSP+20: (TESTopic*)  "<nameless Hello topic>" "<your plugin>.esp"

The engine iterates the topic's `TESTopicInfo` array (`add rbx, 0x08`) into an invalid node. The garbage `GetFullName` is just a nameless Hello topic (normal) — a red herring.

### Minimal repro
1. `bulk_create` a `DialogTopic` with `Subtype=Hello` + one child INFO.
2. Read the topic back — `SNAM` is `00000000`.
3. Load in-game → access violation walking the topic's INFO array.

### Diagnosis (three independent signals)
- `validate_dialogue` → `category=Misc  subtype=Hello  subtype_marker=<none>`
- Raw bytes → `SNAM (4) 00000000`
- 10 vanilla Hello topics → all `SubtypeName = HELO`

### Local fix (confirms the cause)
set_field formid=<topic> field_path=SubtypeName value=HELO into=<same plugin>
4-byte fixed-length field → no offset shift. Byte-diff: only those 4 bytes changed
(`0,0,0,0` → `H,E,L,O`); all INFOs and their `GetIsID` conditions intact;
`validate_dialogue` then reports `subtype_marker=HELO`, graph OK.

### Requests
1. **`create_record`/`bulk_create` should auto-write `SNAM` from the `Subtype` enum.**
[...]

PR #743, opened 2026-09-14
Report each applied edit from the written file, not from memory

### LINKED ISSUE #683: [Bug] housecarl_bulk_apply: Add verb on a LeveledItem's Entries reports success and a grown patch, but writes nothing to disk
### houseCARL version

1.9.0

### Exact repro

```text
Bug: An Add operation targeting a LeveledItem's Entries list returns a success response — including an incremented element count and an "existing patch grown" line — but the entry is never written. The patch file is byte-identical afterward. Failure is silent; only an independent re-read reveals it.

Repro: Confirm the starting state — housecarl_batch_record_detail on 09ABAB:Unslaad.esm, depth: 3 → Entries = [list: 10 item(s)].
Run housecarl_bulk_apply with into: "Blades Katana Leveled List Patch.esp" and:
{"formid":"09ABAB:Unslaad.esm","field_path":"Entries","verb":"Add",
 "compose":{"type":"LeveledItemEntry","Data.Level":"1",
            "Data.Reference":"03AEB9:Skyrim.esm","Data.Count":"1"}}
Observe the response reports success:
extended Blades Katana Leveled List Patch.esp (existing patch grown; 1692 bytes)
1 edit:
  LeveledItem 09ABAB:Unslaad.esm  Add Entries  -> [list: 11 item(s)]
Re-read the record → still 10 entries.
Re-run step 2 with full_readback: true. The disk-level readback prints Entries = [list: 10 item(s)], directly contradicting the same response's own -> [list: 11 item(s)] line.

Additional observations narrowing the cause:

Reported patch size stayed 1692 bytes on every attempt. Genuine edits in the same session did change size (2378 → 2274, 11044 → 10960).
Batching made no difference: a single Add and a batch of 12 both failed identically (the batch reported ascending counts 11→22; disk remained 10).
Set, SetAtIndex, and Remove all persisted correctly in the same session across 8 WEAP and 24 COBJ records (verified by re-reading), including on list elements via bracketed mid-paths (Items[0].Item.Count) and SetAtIndex on Keywords. The defect appears specific to the Add verb.
full_readback: true appears trustworthy; the standard edit-line readback does not reflect what was written.

Edit 9/9/2026: Verified after repairing an unrelated missing-master condition on the target patch. check_errors clean, all masters present, and the Add still silently no-ops identically. Not a masters/resolution issue.
```

### Expected

Add should either (a) append the composed LeveledItemEntry and persist it, so a subsequent read returns 11 entries and the patch file grows — or (b) fail loudly with an error and write nothing. The current behavior, reporting a successful mutation that did not occur, is the worst outcome: a caller that trusts the response will believe a balance change landed when it did not.

### Environment (where relevant)

MO2 instance C:\Authoria, profile Authoria - Requiem Reforged - Main Profile.
Target record: 09ABAB:Unslaad.esm (zzzCrbLitemWeapBlades1H), a LeveledItem defined in Unslaad.esm, already overridden at depth 2 by Blades Katana Leveled List Patch.esp (mod folder [NoDelete] 562 Blades Katana Leveled List Patch). Starting state: 10 entries.

Avick3110:
A write response may only report what it re-read from the written file, and must say not-checked only for what genuinely cannot be judged. `housecarl_apply`'s patch lane re-opened the file it had just written only to read its master header, so every per-edit line printed the applied edit's own in-memory reading. A record can exist in memory and serialize to nothing, and when that happened the response reported the edit as applied against a file that never changed — #683's Add on a LeveledItem's `Entries` answering `-> [list: 11 item(s)]` over a file still holding 10, while `full_readback: true`, which did read the file, contradicted it in the same response.

The patch lane now runs the same per-op verify the in-place lane already ran, off the one re-open it was already doing for the master header, and the per-edit line prints what that re-read says. Three states are distinguished rather than collapsed. An edit whose record the re-opened file does not contain is the one reading that says the edit is not in the file — #683's own failure mode — so it is said outright on the line and counted above the edit list, outside the `max_chars` budget, where a cut cannot drop it; that verdict is only drawn when the walk of the file COMPLETED, since a walk that threw answers about nothing. A leaf the file could not read is genuinely unjudgeable, and says `not-checked` with no value — where an opaque field that reads back unreadable lands. An op a later op in the same call superseded takes the leaf's FINAL file reading, marked as such: the file answers for the leaf even where it cannot answer for the op, so 50 Adds into one list read as 50 values rather than 49 blanks and a count. `LandedOnDisk`, which is COMPARED, stays null there, so the existing superseded handling is untouched. The patch lane's re-open also moves to the strings-aware `LoadOrderResolver.OpenOverlay`, the factory the in-place lane already used, so a comparison is never made against a localized plugin read empty.

`WriteTools.Render` is shared, so `in_place=` responses change here too: only their compact read-back clause came from the file before, while their edit lines came from memory exactly like the patch lane's. The SNAM topic-marker sync appends an op whose `After` is a sentence about what the write did rather than a field reading, and it is now marked as one so the explanation is still printed.

The record-absent verdict and the records it names are hoisted out of the row budget on both transports — the text header names them, json carries `ops_record_absent` and `record_absent_formids` beside `verify_ran` — so a `max_chars` cut cannot leave a response asserting that every edit applied; the truncation clause is gated on the same count. In json a superseded op's `after_on_disk` carries `after_on_disk_is_final_leaf`, the marker the text line got, since that value is the field's final state and not that op's own result.

`housecarl_create` is NOT converted here and its response now says so: its per-field values are the applied edit's own reading, and after this change a create line and a verified edit line are identical in shape, so the create response states which it is and points at `readback=true`. Its re-open is swapped to the strings-aware factory at the same time. The conversion itself is #763 — it needs a nullable request on the verify's pair list (the CK-parity fill ops have no request behind them), a flatten-and-re-split across created records so the file is still walked once, and its own hoisted absent count on both transports; that is a second lane's worth of the same work rather than a line of it.

**Which verbs' edit lines changed source: all of them, on both write lanes.** The change is at the render, not per verb — `Set`, `Add`, `Remove`, `SetAtIndex`, `InsertAtIndex`, `ReplaceAll`, `Merge` and `CopyFrom` all had their line fed from `OpResult.After` (in memory) and now take `OpResult.AfterOnDisk` (the written file). The in-place lane's per-op "what landed" clause in the read-back already came from the file and is unchanged; its edit lines did not, and change with everything else. The dry-run render is deliberately untouched: it writes nothing, says so in its header, and its "would become" values are in-memory by definition. In `format: "json"` the file's reading is the new `after_on_disk` key, and `landed_source` gains `record_absent` beside the words it already had; `after` stays the in-memory reading.

**Does the write itself still no-op?** No. A fixture reproducing #683's exact shape — a LeveledItem defined in a master, overridden by an active patch that is the load-order winner, extended with a composed `Add` on `Entries` through `into=` — grows the file from 10 entries to 11 and grows its byte length. The `Add` lands; what was wrong was the report. The two remaining variables from the issue's install (that plugin, that instance) could not be checked: see below.

**Timing on a bulk apply.** 500 `Set` ops across 500 records, into a fresh patch, on a synthetic 500-record order; five runs per arm, best of five, other lanes measuring on the same machine throughout. Before: 175 ms (four sampling rounds gave 175 / 190 / 210 / 213). After: 226 ms (226 / 227 / 245, with one 560 ms round under load). So roughly +50 ms per 500 ops, about 0.1 ms per op, for one reflective read per op. The file itself is opened and walked once — the verify reuses the overlay the lane already opened and looks each edited record up in a single pass — so the per-op cost is the leaf read alone and nothing scales with the patch's size a second time.

**Checked against the live order.** This lane's own Release build was driven over stdio with `HOUSECARL_DATA_DIR` pointing at a scratchpad config naming the instance `E:\Authoria - Requiem Reforged` (3254 active plugins); the installed server's config was not touched, and no profile list was edited. `000EDE:Skyrim.esm` (`LItemEnchWeaponBowBlacksmith`) is a vanilla LeveledItem whose winner is `Requiem.esp` at override depth 2 — the same overridden-vanilla-list shape #683 reports. One composed `Add` of an iron dagger entry onto its `Entries`, into a fresh throwaway patch on the default lane, answered:

```
wrote HcRb683Check.esp (new patch; 568 bytes)
1 edit:
  LeveledItem 000EDE:Skyrim.esm  Add Entries  -> [list: 19 item(s)]
```

An independent fresh read of that written file — `housecarl_records source="HcRb683Check.esp"`, which resolves it off disk out of the load order — reports `Entries[*count] = 19`. The line and the file agree, and the line is now the file's own reading rather than the record the call held. (19, not 11: the override copies Requiem's winning 18-entry list, and the Add makes 19.) The throwaway mod folder was deleted afterwards and left no trace in `modlist.txt` or `plugins.txt`.

#683's exact record could not be read: `09ABAB:Unslaad.esm` and `Blades Katana Leveled List Patch.esp` belong to the reporter's own instance (`C:\Authoria`), which is not this machine.

Tests: a fixture in #683's shape asserting the file grows and that the per-edit line matches a fresh read of the file; an integration test that many Adds into one list print the file's reading on every line; one pinning the marker-sync sentence; and render tests driving the seam directly — an outcome whose in-memory value and file value deliberately disagree — asserting the response prints the file's and never the in-memory one, that a record missing from the file is said outright while a failed walk is not, and that a superseded op takes the leaf's final file reading.

Closes #683


src/housecarl-mcp/WriteTools.cs
```diff
@@ -305,6 +305,17 @@ static void AppendCompactReadback(StringBuilder sb, IReadOnlyList<WritePatchBuil
...
+    /// re-read after the serialize. Never the in-memory reading — a record that exists in memory and serializes to
+    /// nothing rendered as an applied value under a response that had not looked at the file (#683). Where the file
+    /// cannot answer for this op the line says so and prints no value, because the only value available there is the
+    /// one the file does not vouch for.</summary>
+    static string EditLineValue(WritePatchBuilder.OpResult op) =>
+        op.AfterOnDisk is { } disk ? "  -> " + disk
+        : op.SupersededInCall ? "  -> not-checked [a later op in this call wrote the same field; that op's line carries the file's reading]"
+        : op.VerifyAttempted ? "  -> not-checked [the re-opened file did not answer for this op]"
+        : "  -> not-checked [no file check ran for this op]";
+
```
Avick3110, line 332: **The SNAM topic-marker sync op's whole message is now dropped from the text response.** `SyncEditedTopicMarkers` appends an `OpResult` whose `After` carries the entire explanation — `"Hello — you set Subtype=…; the game buckets by the SNAM marker, so it was synced to match (#131 — otherwise the Subtype change is a silent no-op)"` — and it has no `Landed`, so the per-edit line was the only place it ever appeared (the compact read-back filters ops whose `Landed`/`LandedOnDisk` are null). That op is appended past `resolved`, so `VerifyAttempted` stays false and it now falls to this last arm: `-> not-checked [no file check ran for this op]`. A user who sets `Subtype` on a DialogTopic no longer learns the marker was synced or why. It is also the one op the file *could* answer for — its target is in the written file — so either include it in `perOp` for the verify, or keep printing its `After` for ops the verify never asked about. Nothing in the test suite pins that sentence, so this ships silently.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -305,6 +305,17 @@ static void AppendCompactReadback(StringBuilder sb, IReadOnlyList<WritePatchBuil
...
 
+    /// <summary>The value clause on a completed write's per-edit line: what the WRITTEN FILE holds at that op's leaf,
+    /// re-read after the serialize. Never the in-memory reading — a record that exists in memory and serializes to
+    /// nothing rendered as an applied value under a response that had not looked at the file (#683). Where the file
+    /// cannot answer for this op the line says so and prints no value, because the only value available there is the
+    /// one the file does not vouch for.</summary>
+    static string EditLineValue(WritePatchBuilder.OpResult op) =>
+        op.AfterOnDisk is { } disk ? "  -> " + disk
+        : op.SupersededInCall ? "  -> not-checked [a later op in this call wrote the same field; that op's line carries the file's reading]"
+        : op.VerifyAttempted ? "  -> not-checked [the re-opened file did not answer for this op]"
```
Avick3110, line 330: **A bulk write into one leaf now renders every line but the last as value-free — and this is new for the in-place lane too.** Adding 50 entries to one LeveledItem's `Entries` in a single call (the documented bulk pattern, and #683's own shape) makes ops 1..49 `SupersededInCall`, so the response is 49 lines of `-> not-checked [a later op in this call wrote the same field…]` and one line with a count. Before this PR every one of those lines printed its own reading. The superseded state itself predates the PR, but it only fed `landed_source` and the compact read-back; routing the *edit line* through it is what makes the bulk response unreadable. Worth considering a middle ground for this arm — e.g. printing the file's final reading for the leaf once, or marking the run rather than each member — since the file genuinely does answer for the leaf, just not per op.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -3822,6 +3842,9 @@ internal static IReadOnlyList<OpResult> VerifyLandedAgainstFile(   // internal:
                 // non-null — so presenting it would stamp the op "verified" and print a read failure as the file's
                 // content. Null routes it to the state that already exists for this: attempted, no answer.
                 LandedOnDisk = diskReadable ? landedDisk : null,
+                // The leaf reading travels with it, for the same reason and on the same condition: the per-edit line
+                // prints this one, and a null there is what makes it say not-checked instead of the memory value.
+                AfterOnDisk = diskReadable ? afterDisk : null,
```
Avick3110, line 3930: **The one case that is a verdict is collapsed into the one that isn't.** `VerifyLandedAgainstFile` distinguishes *the written file does not contain this record at all* (`found.TryGetValue` misses → `op with { VerifyAttempted = true }`, both disk fields null) from *the record is there but the leaf read failed* (`diskReadable` false). Both reach the renderer as the same state and print `-> not-checked [the re-opened file did not answer for this op]`, under a header that still says `wrote … N edits:`. But a target record missing from the file the call just wrote is not a representation ambiguity — it is exactly #683's failure mode, and the only signal that the write did not land. On the patch lane `full_readback` is off by default, so nothing else in the response catches it. Suggest carrying the record-not-found case as its own state and wording it loudly ("the written file does not contain this record"), rather than as an unremarkable not-checked.

plugin/CHANGELOG.md
```diff
@@ -13,6 +13,20 @@ saying it sets an expectation their install may contradict. Say what is known, a
 
 ## Unreleased
 
+- **A write's per-edit line now reports what the written file holds, not what the call held in memory.**
+  `housecarl_apply` re-opened the patch it had just written only to read its master header, so each edit's
+  `-> value` came from the record in memory. A record can exist in memory and serialize to nothing, and when it
+  did the response reported the edit as applied — an Add on a leveled list answering `-> [list: 11 item(s)]`
+  over a file still holding 10 — while `full_readback: true`, which did read the file, contradicted it in the
+  same response. Every edit line is now re-read off the written file, for every verb, out of the re-open the
+  lane was already doing; the in-place lane, which already re-read, is unchanged. Where the file cannot answer
```
Avick3110: **"the in-place lane, which already re-read, is unchanged" is not accurate.** `WriteTools.Render` is shared by both lanes, and before this PR its per-edit line was `op.After is not null ? "  -> " + op.After : "  -> applied"` on in-place too — only the *compact read-back* clause came from the file there. So the in-place lane's edit lines change source in this PR as well, and in-place responses lose values in the same two cases (a superseded sibling, and the appended SNAM-sync op, which on that lane now says `no file check ran for this op` even though the verify did run). Same wording in the PR body. Worth correcting so a user reading the changelog doesn't attribute an in-place output change to something else.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded. The op now carries `AfterIsNote`, saying its `After` is a sentence about what the write did rather than a reading of a field, and the per-edit line prints it as it stands. Kept the message rather than wiring the op into `perOp`: there is no `WriteRequest` behind it, and inventing one to re-read `SubtypeName` would put a synthetic op in the pass whose whole job is to answer for real ones. `TheTopicMarkerSyncStillExplainsItself` pins the sentence.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded, taking the middle ground you name: a superseded op now carries the leaf's FINAL file reading in `AfterOnDisk`, and the line prints it with a clause saying a later op in this call wrote it too. `LandedOnDisk` stays null there, so the comparison that the superseded arm exists to avoid is untouched and `landed_source` still says `superseded`. Three Adds into one list now print three values instead of two blanks and a count — `EveryLineOfAManyOpRunIntoOneListCarriesTheFilesReading`.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Folded. `RecordAbsentFromFile` is now its own state on `OpResult`, set only when the walk of the written file COMPLETED — a walk that threw answers about no record, and claiming absence from it would invent a verdict, which `AFailedWalkIsNotCheckedRatherThanAVerdict` pins. The line reads `DID NOT LAND — the written file was re-opened and does not contain this record…`, and the count is repeated above the edit list, outside the ops budget, so a `max_chars` cut cannot be what removes it. `landed_source` gains `record_absent`.

plugin/CHANGELOG.md
Avick3110: Correct, and folded. Both the changelog entry and the PR body now say the in-place lane's edit lines change source too, and that what already came from the file there was the read-back clause alone. Its appended marker-sync op keeps its message rather than reporting no file check, per the fix on your first finding.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -132,6 +132,12 @@ internal static string Render(WritePatchBuilder.PatchOutcome o, int maxChars = 0
         else
             sb.Append(WriteSentences.NewOrExtendedArtifact(o.Extended, file, o.Bytes, modFolder));
         sb.Append(WriteSentences.Masters(o.Masters));
+        // Said ABOVE the ops and outside their budget: a record missing from the file the call just wrote is the one
+        // reading that contradicts the header, and an op-list cut must not be what removes it.
+        int absent = o.Ops.Count(op => op.RecordAbsentFromFile);
+        if (absent > 0)
+            sb.Append("! ").Append(absent).Append(absent == 1 ? " edit did NOT land: " : " edits did NOT land: ")
+              .Append(WriteSentences.RecordAbsentFromWrittenFile).Append(". Each is marked below.\n");
```
Avick3110: The new "did NOT land" header is hoisted above the op list, but the op list below it is still cut at `opCap`, and the truncation bracket it is cut with says the opposite thing.

Concrete case: a 400-op bulk apply under a small `max_chars` where 3 targets are missing from the written file. The response opens with `! 3 edits did NOT land: ... Each is marked below.` and then, once `sb.Length >= opCap`, closes the list with `... [truncated: N of 400 edit(s) listed ...; every one WAS applied — ...]`. So one response asserts both "3 edits did NOT land" and "every one WAS applied", and if the three absent ops sit past the cut, "Each is marked below" is simply false and the caller has no way to learn *which* three did not land.

Two things would fix it: make `RowsCutOperationIntact`'s "every one WAS applied" conditional on `absent == 0` (or reword it to "every one was attempted"), and list the absent ops' FormIDs in the hoisted header rather than deferring to the budgeted list.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2443,7 +2443,7 @@ public static string RenderPatchOutcome(WritePatchBuilder.PatchOutcome o, int ma
             foreach (var m in o.Masters) w.WriteStringValue(m);
             w.WriteEndArray();
 
-            // Did the per-op file check RUN at all? A patch-lane or dry-run document has no per-op file readings in
+            // Did the per-op file check RUN at all? A dry-run document has no per-op file readings in
             // it, and a consumer needs to tell that from "it ran and everything came off the file". Outside the ops
             // budget, because that is the one fact a max_chars cut must not remove.
             w.WriteBoolean("verify_ran", o.Ops.Any(op => op.VerifyAttempted));
```
Avick3110, line 2476: The json lane hoists `verify_ran` out of the ops array on the stated grounds that it is "the one fact a max_chars cut must not remove" — but the new record-absent verdict, which is a strictly stronger fact, gets no such treatment. It exists only as `landed_source: "record_absent"` on each op object, inside the array that `Over(w, ms, cap)` truncates.

Concrete case: a 400-op apply with `format: "json"` and a `max_chars` that cuts the array at op 120, where the three records missing from the written file are ops 200, 300 and 390. The document then carries `ok: true`, `verify_ran: true`, no `record_absent` anywhere, and a `truncated_note` reading `... every one WAS applied — ...`. A json consumer is told the write succeeded and every edit applied, with the only contradicting evidence silently dropped. The text lane hoists the count for exactly this reason; the two lanes now disagree about the same call.

`after_on_disk` being added to every op object also makes each op bigger, so the cut lands earlier than before for the same `max_chars`.

Suggest a hoisted counterpart beside `verify_ran`, e.g. `w.WriteNumber("ops_record_absent", o.Ops.Count(op => op.RecordAbsentFromFile));`, and gating `RowsCutOperationIntact`'s "every one WAS applied" on that count being zero.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -2462,6 +2462,10 @@ public static string RenderPatchOutcome(WritePatchBuilder.PatchOutcome o, int ma
                 w.WriteBoolean("applied", op.Applied);
                 WriteNullable(w, "error", op.Error);
                 WriteNullable(w, "after", op.After);
+                // The leaf as the WRITTEN FILE holds it — what the text render's per-edit line prints. Null when the
+                // file could not answer for this op, which `landed_source` names; `after` beside it stays the applied
+                // edit's own in-memory reading, so the two are never confused for one another.
+                WriteNullable(w, "after_on_disk", op.AfterOnDisk);
```
Avick3110, line 2516: This comment is now inaccurate for the superseded arm, and the inaccuracy is load-bearing for json consumers. `VerifyLandedAgainstFile` populates `AfterOnDisk` on a superseded op with `DescribeApplied(rec, perOp[i].Req).After` — the leaf's FINAL state after every op in the call — while `landed_source` for that op is `"superseded"`, i.e. "the file ... cannot speak for this one". So `after_on_disk` is non-null in precisely the case the doc says it is null, and it holds a value that is not this op's result.

Concrete case: the 3-Add-into-one-list shape the new integration test covers. In json, ops 0 and 1 come back with `after_on_disk: "[list: 13 item(s)]"` — the same value as op 2 — with nothing in the op object but the `landed_source` word to say that 13 is not what op 0 produced. The text render marks it inline (`[the leaf as the file now holds it; a later op in this call wrote it too]`); json has no equivalent marker, so a consumer reading `after_on_disk` as "what this op put on disk" misreports N-1 of N ops in any bulk run into one field.

Either restate the contract here (non-null on superseded, where it is the leaf's final reading rather than this op's), or give json the same distinction the text lane got — e.g. an `after_on_disk_is_final_leaf` boolean, or a separate key.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -305,6 +311,21 @@ static void AppendCompactReadback(StringBuilder sb, IReadOnlyList<WritePatchBuil
     /// say afterwards (a duplicate Add). Empty when there is nothing to say.</summary>
     static string ApplyNote(WritePatchBuilder.OpResult op) => op.ApplyNote is { } n ? "  [" + n + "]" : "";
 
+    /// <summary>The value clause on a completed write's per-edit line: what the WRITTEN FILE holds at that op's leaf,
+    /// re-read after the serialize. Never the in-memory reading — a record that exists in memory and serializes to
+    /// nothing rendered as an applied value under a response that had not looked at the file (#683). Where the file
+    /// cannot answer for this op the line says so and prints no value, because the only value available there is the
+    /// one the file does not vouch for — but a record the file does not CONTAIN is an answer, and a bad one, so it is
+    /// said outright rather than as one more unchecked line.</summary>
+    static string EditLineValue(WritePatchBuilder.OpResult op) =>
```
Avick3110, line 324: The create lane's per-edit line was left on the in-memory reading, and after this PR the two are indistinguishable in the output.

`RenderCreate` still renders each created record's ops as `op.Label + (op.After is not null ? "  -> " + op.After : "  -> applied")` (same file, ~line 943) — `OpResult.After`, the value read off the in-memory record the instant the verb applied. `housecarl_create` re-opens the file it wrote (`WritePatchBuilder` ~line 3611) but only for `masters`/`bytes` and, when `full_readback: true`, `ReadBackInFull`; it never runs `VerifyLandedAgainstFile`, and its compact read-back is called with `Array.Empty<OpResult>()`, so on the default path nothing in a create response has looked at the written file at all.

That is the same class as #683 — a record that exists in memory and serializes to nothing, or a field that does — and it is now harder for a caller to notice, not easier: an apply response's `Type 000EDE:Skyrim.esm  Set Name  -> "Foo"` is the file's reading and a create response's `Name  -> "Foo"` is memory's, and the two lines are byte-identical in shape. The PR body's "all of them, on both write lanes" is true of the apply verbs but reads, in a changelog, as a property of write responses generally.

Two things would be worth deciding explicitly rather than leaving implicit: whether create is in scope for the same treatment (the re-open is already there, and `VerifyLandedAgainstFile` takes `(Target, Req)` pairs which the create lane has), and — if it is not — whether the create lane's lines should be distinguishable from the now-verified ones. A follow-up issue is a fine answer; silently divergent provenance behind identical syntax is the part worth avoiding.

Separately and pre-existing, that create re-open at ~3611 is the bare `SkyrimMod.CreateFromBinaryOverlay`, not `LoadOrderResolver.OpenOverlay` — the exact swap this PR made on the patch lane, for the exact reason stated there (a localized plugin read bare returns every `TranslatedString` empty). Today that only reaches `full_readback`, but it would be the first thing to fix if create ever does start reporting from the file.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded, both halves. `RowsCutOperationIntact` takes a `someDidNotLand` flag and swaps to "every one was attempted, and the ones that did NOT land are named outside this cut" when the count is non-zero; the other three verbs pass false and their wording is unchanged. The hoisted header now names the absent records itself, distinct and bounded at 10 with the remainder counted, so the list cannot be what a cut drops. `ACutOpListStillNamesTheRecordsThatDidNotLand` renders 41 ops with the absent one last under a 900-char cap and asserts its row is past the cut while its FormID is still in the header; `ACutOpListDoesNotClaimEveryEditAppliedWhenOneDidNot` pins both wordings.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded as suggested. `ops_record_absent` sits beside `verify_ran`, with `record_absent_formids` (bounded at the same 10) and `record_absent_formids_total` so a consumer can always tell the array was capped, and `truncated_note` is gated on the same count. `TheJsonHoistsTheAbsentVerdictOutOfTheOpsArray` drives the same 41-op cut and asserts the document still carries the verdict and no longer claims every edit applied.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded, both parts. The comment now states the real contract — non-null on a superseded op, where it is the leaf's final state rather than that op's result — and json gets the marker the text lane has: `after_on_disk_is_final_leaf: true` on exactly those ops. A boolean rather than a separate key, so `after_on_disk` stays one thing to read and the caveat is beside it. Pinned by `TheJsonMarksASupersededOpsValueAsTheLeafsFinalState`.

src/housecarl-mcp/WriteTools.cs
Avick3110: Right on both counts. Folded the two cheap halves here: the create lane's re-open is now `LoadOrderResolver.OpenOverlay`, and a create response whose records carry field ops states that those values are the applied edits' own readings, not re-read from the written file, pointing at readback=true — so the identical-looking lines are no longer silently divergent, which is the part you flagged as worth avoiding.

The conversion itself is #763 rather than this PR. It is genuinely a second lane's worth: `VerifyLandedAgainstFile` takes positionally-aligned (Target, Req) pairs, and the create lane's op lists interleave CK-parity fill ops (SNAM marker, DIAL Priority, INFO/View/Branch/Quest defaults) that have no request behind them, so the pair list needs a nullable request or the fills need separating; the ops are nested per created record and the verify walks the file once, so it has to be flattened and re-split rather than called per record; and the create render and its json twin would each need the hoisted absent count and the gated truncation clause this PR's own review just showed are required, or the create lane reproduces exactly that defect. The issue writes all of that down.

PR #744, opened 2026-09-14
Stop the verify calling an unparsed blob clean, and annotate it with the FormVersion

### LINKED ISSUE #529: [Bug] the in-place verify says re-read clean for an opaque-bytes field it cannot judge: a FormVersion-mismatched MODT passes and CTDs the game
From a Discord report (2026-08-26, houseCARL 1.9). After 2.0 unless the probe below promotes it.

A `Model.Data` (MODT) blob copied verbatim from a FormVersion-39 source onto a new record stamped FormVersion 44 is written with no conversion and no refusal; the forced in-place verify in `WriteTools.cs` ("re-read clean") is a Mutagen re-read that can never fail on a malformed MODT. The plugin CTDs on load.

Scope of this issue is the honest half only:
- the verify must not say "clean" about a structure it never checked (opaque `bytes` fields);
- a read of a layout-versioned blob should not render as unannotated hex;
- a MODT-vs-FormVersion note in the Mutagen-lag lane.

Out of scope: a hand-written MODT decoder or normaliser. `Model.Data` is opaque bytes to Mutagen and per-type hand-coding is what the coverage rule forbids.

Probe before scheduling: does Mutagen's `Duplicate(newKey)` preserve the source FormVersion or re-stamp 44? If it re-stamps, `housecarl_copy new_editorid=` off a FormVersion-39 STAT reproduces the CTD and this promotes to W5 item 19.

Avick3110:
The forced in-place verify is a Mutagen re-read, and Mutagen models a `bytes` field as raw bytes it never decodes — `Model.Data` (MODT) is the case that bites. A blob whose internal layout belongs to a different FormVersion re-reads byte-identical, so the verify said "re-read clean" about a structure it had never looked at and the plugin crashed the game on load. This makes the two honest halves true: the verify names what it could not judge, and a read shows the FormVersion the blob's layout belongs to. No MODT decoder, no per-record-type anything — the byte marker comes off the leaf's own type, so every `bytes` field on every record type is covered by construction.

The verify line now reads `✓ Static <formid> — re-read clean (N field(s), except Model.Data — re-read as 12 opaque byte(s) only, structure NOT checked)`, with the exclusion inside the "clean" parenthetical so it cannot be read as covering the blob. A read renders `Model.Data = 020000000000000000000000   (opaque bytes, 12 byte(s) — layout follows this record's FormVersion 40; not parsed)`. The annotation rides `FieldValue.Display`, so it is display-only: the hex token still round-trips to a write, and the read-proof oracle, the write surface and the conflict diff never see it. The `readback=true` dump and the json read-back carry it too (json also gets an `opaque_bytes` length, so a consumer need not parse prose to know a value went unjudged). The schema skill gains the rule this leaves the modder with: a blob is only safe to copy between records at the same FormVersion.

**The probe the issue asked for: Mutagen's `Duplicate(newKey)` PRESERVES the source record's FormVersion — it does not re-stamp 44.** Two tests assert it, one on the duplicate in memory and one on the duplicate written to a plugin and read straight back, because the writer is the half that decides what the game sees. So the promotion condition in the issue does not fire: `housecarl_copy new_editorid=` off a FormVersion-39 STAT carries the 39 stamp along with the MODT, and the two stay consistent — that path does not reproduce the CTD, and I have filed nothing for it. The path that does is a FIELD-level copy of a blob between records at different FormVersions (`op: "CopyFrom"` on `Model.Data`, or a hand-pasted hex value), which nothing refuses and nothing can check without decoding MODT; that is what the new verify clause and the read annotation exist to make visible, and it stays out of scope here per the issue's own bound.

Verified: `dotnet build housecarl.sln -c Release`, `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` (2008 passed), and `ci-all` (128/128). Six new tests cover the verify clause, the read annotation, the structural byte marker, the untouched round-trip token, and the two Duplicate probes. One suite run hit a single unrelated intermittent failure that did not reproduce on two further full runs — this machine is running ten lanes in parallel over shared temp. Real-install check: drove this worktree's Release server over stdio against the live MO2 instance at `E:\Authoria - Requiem Reforged` and read `Model.Data` off three Skyrim.esm STATs; all three annotated as 12 opaque bytes at FormVersion 40 — real vanilla records sitting four versions below the 44 a new record is stamped at, which is the mismatch the annotation exists to show.

Closes #529


src/housecarl-core/ReadEngine.cs
```diff
@@ -262,9 +272,32 @@ public static RecordFields ReadFields(IMajorRecordGetter record, IReadOnlyList<s
                 EmitWithDepth(on, string.Join(".", tail), d, fields, ref budget, p);
             }
         }
+        AnnotateOpaqueBytes(fields, record.FormVersion);
```
Avick3110: A `*parent`-hopped leaf gets stamped with the WRONG record's FormVersion. `HopToParent` rebinds `on` to the containing record and `ReadLeaf` reads the leaf off that record, but `AnnotateOpaqueBytes` is handed `record.FormVersion` — the record the read started from. Concrete case: `housecarl_records formids=<a REFR> fields=["*parent.OcclusionData"]` (Cell carries `LNAM`, `OcclusionData`, `XWCN`, `XWCS` as `bytes`, and Worldspace carries `OffsetData`) renders the CELL's blob annotated "layout follows this record's FormVersion 44" using the PlacedObject's stamp. Since the whole point of the annotation is to let a modder spot a FormVersion mismatch before copying a blob, a confidently wrong version is worse than none: a CELL at 43 under a REFR at 44 reads as a match and a real mismatch reads as a difference. The fix is to annotate off the record the leaf was actually read on — carry the hop's `on.FormVersion` per field (or annotate inside each branch after the hop) rather than once at the end with the entry record's.

src/housecarl-mcp/JsonWire.cs
```diff
@@ -247,6 +247,10 @@ static void WriteReadbackBlock(Utf8JsonWriter w, MemoryStream ms, int cap,
                     w.WriteStartObject();
                     w.WriteString("path", f.Path);
                     if (f.HasValue) w.WriteString("value", f.Token); else w.WriteString("note", f.Note);
+                    // An opaque blob's annotation rides here too: a json consumer reading a hex value must be able to
+                    // see it was re-read as bytes only, under which FormVersion, without matching prose.
+                    if (f.Display is not null) w.WriteString("display", f.Display);
+                    if (f.Bytes is { } n) w.WriteNumber("opaque_bytes", n);
```
Avick3110: `opaque_bytes` is emitted only on the write read-back lane; the ORDINARY read json lane (`WriteLeaf`, ~line 354) writes `display` but no `opaque_bytes`. So a consumer of `housecarl_records format=json` — the lane most likely to be machine-read, and the one a field-level `CopyFrom` of a blob would be driven from — still has to regex the prose to learn a value went unjudged, which is exactly what the new `FieldValue.Bytes` doc says a consumer must not have to do. Adding the same `if (f.Bytes is { } n) w.WriteNumber("opaque_bytes", n);` to `WriteLeaf` closes it (and covers folded `cells` rows for free).

src/housecarl-core/ReadEngine.cs
Avick3110: Folded in 247931ba. You are right and the concrete case is real: Cell carries LNAM/OcclusionData/XWCN/XWCS as bytes and Worldspace carries OffsetData, so a `*parent`-hopped read of one of those was stamped with the entry record's FormVersion. `AnnotateOpaqueBytes` now takes a start index and is called PER PATH with `on.FormVersion` — the record the hop rebound to — in both the depth-1 and the deep branch, so the stamp is the one of the record the leaf was actually read on. A test builds an interior cell at FormVersion 43 carrying a blob with a PlacedObject inside it and reads `*parent.OcclusionData` off the reference: it asserts the annotation says 43 and does not say 44. I checked it fails against the old one-call-at-the-end version before passing against the fix.

src/housecarl-mcp/JsonWire.cs
Avick3110: Folded in fb674ad5 — exactly as you wrote it: `WriteLeaf` now emits `opaque_bytes` beside `display`, so the ordinary `format=json` read lane carries the length as a number and folded `cells` rows get it for free. A test reads `Model.Data` through `housecarl_records format=json` and asserts the number off the field object rather than matching prose.

Avick3110:
Review folded — all three findings were real and all three are fixed; nothing refuted.

- **`*parent` FormVersion (High)** — 247931ba. `AnnotateOpaqueBytes` is now called per path with the record the hop rebound to, in both read branches. New test: an interior cell at FormVersion 43 carrying a blob, a PlacedObject inside it, `*parent.OcclusionData` read off the reference. It fails against the old code.
- **`opaque_bytes` on the ordinary json read lane (Medium)** — fb674ad5, plus a test that reads the number out of the json document.
- **Stale comment in `LoadOrderService.cs` (Low, unposted)** — 1565ae6f; the comment now names both Display producers.

Rebuilt and retested on origin/main: full suite 2010 passed, `ci-all` 128/128.

src/housecarl-core/ReadEngine.cs
Avick3110: **The annotation can state the wrong FormVersion — exactly the mismatch it exists to reveal.** `ReadFields` runs `HopToParent` per path, so a `*parent...` path reads its leaves off the *containing* record, but `AnnotateOpaqueBytes` is handed `record.FormVersion` — the FormVersion of the record the caller named, not of the record the blob was actually read off. Scenario: a read that reaches a bytes leaf through a `*parent` hop renders `layout follows this record's FormVersion 44` while the blob came off a FormVersion-39 parent (and a read mixing a plain path with a `*parent` path annotates both from the child). Since the whole point of the line is to make a FormVersion mismatch visible, a silently wrong version here is worse than no annotation. The FormVersion needs capturing per leaf at the record it was read from (carry it on `LeafRead`/`FieldValue` beside `ByteLength`, or annotate inside the loop from `on.FormVersion`), not once from the top-level record.

src/housecarl-mcp/JsonWire.cs
Avick3110: **`opaque_bytes` is emitted on the write read-back lane only — the ordinary json read never carries it.** `WriteLeaf` (JsonWire.cs ~L347-355), which is what `housecarl_records form=fields` json goes through, writes `display` but no `opaque_bytes`. So a json consumer reading `Model.Data` through the normal read lane gets only the prose — the exact thing `FieldValue.Bytes`' own doc comment says a consumer "must not have to" parse ("deciding whether a value was actually JUDGED must not have to match a hex-looking token or parse the display annotation"). The CHANGELOG also claims "every read (and the `readback=true` dump, and the json `display` / `opaque_bytes` fields)", which is not true of the read lane today. Add the same `if (f.Bytes is { } n) w.WriteNumber("opaque_bytes", n);` to `WriteLeaf`.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -301,6 +306,21 @@ static void AppendCompactReadback(StringBuilder sb, IReadOnlyList<WritePatchBuil
...
+    /// the known case) as raw bytes it never parses — so a blob whose layout does not suit the record's FormVersion
+    /// re-reads without complaint and the game crashes on it. The verify must not claim a structure it never looked
+    /// at, so it names those fields in the SAME sentence and says they were re-read as bytes only (#529). Empty when
+    /// the record carries no such field, which is the common case.</summary>
+    static string OpaqueBytesCaveat(RecordFields rec)
+    {
+        var opaque = rec.Fields.Where(f => f.Bytes is not null).Select(f => f.Path).ToList();
+        if (opaque.Count == 0) return "";
+        int bytes = rec.Fields.Where(f => f.Bytes is not null).Sum(f => f.Bytes!.Value);
+        return ", except " + string.Join(", ", opaque) + " — re-read as " + bytes
```
Avick3110: **The caveat is unbounded and can blow past the compact render's char cap.** `rec.Fields` here is the depth-16 full read-back, and two modeled record types carry a *list* of blobs — `DialogView.TNAMs` and `MaterialObject.DNAMs` are `List<ReadOnlyMemorySlice<byte>>`. At depth 16 every element expands to its own leaf with `Bytes` set, so an in-place edit to such a record produces `..., except TNAMs[0], TNAMs[1], … TNAMs[N] — re-read as …` on a *single* line. `AppendCompactReadback` only checks `sb.Length >= cap` *before* appending a record, so one record's caveat overshoots the cap arbitrarily — in the lane whose whole purpose is "report it compactly so it cannot overflow the host token cap". Cap the path list (e.g. first few plus "and N more").

Secondary, same line: with several opaque fields the sentence reads `except A, B — re-read as 36 opaque byte(s) only`, where 36 is the *sum*; a reader naturally attaches it to the last named field. Either say "36 byte(s) across 2 field(s)" or put the count beside each path.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -256,7 +256,11 @@ static void AppendFullReadback(StringBuilder sb, IReadOnlyList<WritePatchBuilder
                       .Append(hint).Append("]\n");
                     return;
                 }
-                sb.Append("    ").Append(f.Path).Append(" = ").Append(f.HasValue ? f.Token : f.Note).Append('\n');
+                sb.Append("    ").Append(f.Path).Append(" = ").Append(f.HasValue ? f.Token : f.Note);
+                // The display-only annotation, exactly as the read lane renders it: an opaque blob must not appear as
+                // unannotated hex here either, since this dump IS the read of the record just written.
+                if (f.Display is not null) sb.Append("   (").Append(f.Display).Append(')');
```
Avick3110: **This widens the read-back dump for every annotated field, not just blobs.** `FieldValue.Display` is already populated by `FlagDisplay` (the biped-slot decode and the `FlagBitsDisplay` unknown-bits decode). Gating on `f.Display is not null` therefore adds a trailing `   (Body | Hands)` / `   (… (+unknown bits 0x…))` to every flags line in the `readback=true` dump, which previously rendered as the bare token — a change to the write-lane output for records that carry no blob at all. Same for the new `w.WriteString("display", …)` in the json read-back block. It may well be the render you want (the CLI and `ReadTools` already do it), but it is not what the comment says ("an opaque blob must not appear as unannotated hex here either") and it is not mentioned in the CHANGELOG, so anyone parsing read-back lines gets a surprise. Either narrow the gate to `f.Bytes is not null` or state the broader change.

src/housecarl-core/ReadEngine.cs
```diff
@@ -262,9 +272,32 @@ public static RecordFields ReadFields(IMajorRecordGetter record, IReadOnlyList<s
...
+    }
+
+    /// <summary>The DISPLAY-ONLY annotation on an opaque blob leaf: how many bytes, and the FormVersion of the record
+    /// they were read off. A blob like <c>Model.Data</c> (MODT) is laid out per the record's FormVersion, and Mutagen
+    /// models it as raw bytes it never parses — so unannotated hex hides the one thing that decides whether the bytes
+    /// suit the record carrying them. Naming the FormVersion beside the blob makes a mismatch visible to the reader:
+    /// bytes lifted off a FormVersion-39 record and rendered here under FormVersion 44 are a mismatch nothing else
+    /// reports (#529). It says "not parsed" because houseCARL does not parse it — there is no hand-written decoder
+    /// and no per-type layout table.</summary>
+    public static string BytesDisplay(int length, ushort? formVersion) =>
```
Avick3110, line 316: **The prose lands inline in dense scan cells too.** `RowProjection.Cell` does `val += $" ({f.Display})"`, so every projected blob cell now carries ~90 characters of prose after the hex (`0200…   (opaque bytes, 12 byte(s) — layout follows this record's FormVersion 40; not parsed)`) in the DENSE row render, where cells are joined with a separator and rows are width/cap bounded. A scan over many STATs projecting `Model.Data` at depth 2 will truncate far earlier than before, and the per-cell text now contains `(`, `)` and `—` in a lane that is positional. Consider a short form for the cell (`[opaque 12B @FV40]`) and keeping the full sentence for the single-record read and the json `display`.

src/housecarl-core/ReadEngine.cs
Avick3110: This was real and is already fixed — in 247931ba, from the first review round; this read a stale diff. `AnnotateOpaqueBytes` now takes a start index and is called PER PATH with `on.FormVersion` (ReadEngine.cs lines 259 and 280), so the stamp is the record the hop rebound to, in both the depth-1 and the deep branch. `FieldValue.BytesFormVersion` now carries it structurally too, as you suggested. Covered by `AParentHoppedBlobCarriesTheContainingRecordsFormVersion`: an interior cell at FormVersion 43 carrying a blob, a PlacedObject inside it, `*parent.OcclusionData` read off the reference — asserts 43 and asserts NOT 44. I checked it fails against the pre-fix version.

src/housecarl-mcp/JsonWire.cs
Avick3110: Also already fixed, in fb674ad5 from the first round — same stale diff. `WriteLeaf` now has `if (f.Bytes is { } opaque) w.WriteNumber("opaque_bytes", opaque);` at JsonWire.cs line 357, so the ordinary `format=json` read carries it and folded `cells` rows get it for free. Covered by `TheJsonReadLaneCarriesTheByteLengthAsANumber`, which reads the number off the field object. The changelog sentence you flagged is true as of that commit, and I have since rewritten it anyway for the other findings here.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 2a7265a9, taking the narrow option: both gates are now `f.Bytes is not null`, in `AppendFullReadback` and in the json read-back block, so a flags line in `readback=true` and in the json read-back renders exactly as it did before this PR. Widening the write lane was not intended and was not in the changelog, so narrowing is the right call rather than documenting an accident. The json read-back now also carries `opaque_form_version` beside `opaque_bytes`, for the same reason `opaque_bytes` exists.

src/housecarl-core/ReadEngine.cs
Avick3110: Folded in eb27d702, using your suggested shape. `FieldValue` now carries `BytesFormVersion` structurally beside `Bytes`, and `RowProjection.Cell` builds a short form from the two — `[opaque 4B @FV44]` — instead of appending the sentence; every other render keeps the prose. New test `TheFoldedRowRenderUsesTheShortAnnotation` asserts the short form is present in a folded row render and that the sentence is not.

src/housecarl-mcp/WriteTools.cs
Avick3110: Both halves folded in e888b0e4. `MaterialObject.DNAMs` and `DialogView.TNAMs` are indeed lists of `ReadOnlyMemorySlice<byte>` (plus `PackageBranch.Unknown`), so the list really is unbounded at depth 16. The caveat now names the first three paths and counts the rest, and each named path carries its OWN byte count instead of one sum — so it reads `except DNAMs[0] (4 byte(s)), DNAMs[1] (4 byte(s)), DNAMs[2] (4 byte(s)) and 5 more — re-read as bytes only, structure NOT checked`. New test `ARecordWithManyBlobsNamesAFewAndCountsTheRest` edits a fixture MaterialObject carrying eight DNAMs in place and asserts the first is named with its own count, that the line says "and 5 more", and that the last element is NOT named.

Avick3110:
Second review folded. Three of the five were new and real and are fixed; two were already fixed in the first round and the reviewer read a stale diff — replies on each thread point at the commit and the test.

- **Unbounded caveat list, and a byte sum read as one field's (Medium)** — e888b0e4. `MaterialObject.DNAMs` / `DialogView.TNAMs` / `PackageBranch.Unknown` really are lists of blobs, so at read-back depth the list was unbounded on a line whose whole job is to stay under the cap. It now names three paths with per-field counts and counts the rest. New test drives a fixture record carrying eight blobs.
- **The `Display` gate widened every flags line in the write lane (Low)** — 2a7265a9. Both gates narrowed to `f.Bytes is not null`, so a flags line in `readback=true` and in the json read-back renders exactly as it did before this PR. The json read-back also gained `opaque_form_version`.
- **Prose in dense/folded cells (Low)** — eb27d702. `FieldValue.BytesFormVersion` carries the version structurally, and the folded cell renders `[opaque 4B @FV44]` instead of the ~90-character sentence; every other render keeps the prose. New test.
- Findings 1 and 2 (`*parent` FormVersion; `opaque_bytes` on the read lane) were fixed in 247931ba and fb674ad5 before this review ran.

Changelog rewritten to match. Rebuilt and retested on origin/main: full suite 2012 passed, `ci-all` 128/128.

PR #745, opened 2026-09-14
Read a tree's providers one at a time instead of all at once

### LINKED ISSUE #722: [GAP] tree over two worldspace records peaks at 4–7 GB and holds it after the call

`project.form="tree"` over two worldspace records (Tamriel and WhiterunWorld, `formids=`) peaks at **3.8–4.7 GB** in 30 s, 7.4 GB under 16 GC heaps, and the memory is held after the call returns. The same two records under `everything` at depth 4 cost 0.9 s and no rise; a `chain` walk from Tamriel reaches 189 nodes in 2.3 s.

| call | records | elapsed | peak WS | peak private |
|---|---:|---:|---:|---:|
| tree, Tamriel + WhiterunWorld | 2 | 30.4 s | **3,846 MB** | **4,097 MB** |
| everything depth 4, same | 2 | 0.9 s | — | — |
| tree, Dragonsreach + Jarl's quarters (CELL) | 2 | 4.3 s | 1,018 MB | 957 MB |
| tree, Tamriel + WhiterunWorld, `GCHeapCount=16` | 2 | 30.6 s | 7,445 MB | — |

A worldspace on this order has hundreds of providers, and the tree reads every provider's whole body to diff depth and keeps them all until the render ends. Two records should not cost 4 GB; the diff needs one reference body plus one provider body at a time.

Related: the per-row cost of the same form over a REFR scan is filed separately (time, not memory).

Avick3110:
A `project.form="tree"` row read every provider of the record and kept each one's fields alive until the whole render was assembled, so two worldspace records — Tamriel has 772 providers on this order, WhiterunWorld 78 — cost gigabytes and the process still held them after the call returned. The fill now reads the providers WINNER FIRST and hands them out one at a time: the reference pole is the winner unless `versus=` named another, so reading it first is what lets every other provider be diffed and released before the next is fetched. A record getter is a slice of its whole GRUP's byte array and pins it, and the fields read off it are the bigger half again; the diff needs one reference plus one provider, which is now all that is alive. The resolver gained a streamed sibling of `ResolveTree` (`StreamTree`, with a `winnerFirst` walk), and `ResolveTreePinned` became a fold — `FoldTreePinned` — that the render drives with a visitor diffing each provider as it arrives; the whole-tree materialisation stays as a one-line wrapper over the fold for the one caller that genuinely needs the providers side by side (the conflict-diff probe). A `versus=` pole that refuses now stops the walk after the winner instead of reading every provider for a row that is discarded.

No response change.

## Measured

Private `housecarl-mcp` per arm over stdio against the live order (`E:\Authoria - Requiem Reforged`, 3,571 modlist lines), index warmed with `load_order_status` first, working set and private bytes sampled every 250 ms from outside, max inside the call's window. "held" is the reading one second after the call returned.

| arm | before | after |
|---|---|---|
| tree, Tamriel + WhiterunWorld (`formids=`) | 52.3 s · peak WS **2,971 MB** · peak priv 3,022 MB · held 2,600 MB | 53.1 s · peak WS **1,878 MB** · peak priv 1,865 MB · held 1,818 MB |
| tree, Dragonsreach + Jarl's quarters (CELL, the control) | 5.3 s · peak WS 1,023 MB · peak priv 955 MB · held 901 MB | 4.6 s · peak WS 1,073 MB · peak priv 989 MB · held 970 MB |

The control moves inside its own noise (its two runs started from index baselines 38 MB apart). Elapsed is unchanged on both arms: this was memory, not time.

What is left on the worldspace arm is the answer itself. A diagnostic arm on the pre-change build, the same two records with `project.fields=["EditorID"]` — which reads the same 850 provider bodies but renders almost nothing per node — peaked at 1,099 MB, i.e. the index plus ~300 MB. So the remaining ~800 MB is the deltas of 850 providers plus the garbage from reading their fields, and the lever for that is the number of records asked for, not the fill.

## Verified

- **Byte-identity on the real order.** Both arms were run with the result spilling to a JSONL artifact and the artifacts compared before and after: every data row is identical (`md5` of lines 2..n equal on both arms — 67.8 MB of worldspace rows, and the CELL pair). The manifest line differs only in its `created` timestamp.
- `dotnet build housecarl.sln -c Release`, `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` (2,002 passed), `ci-all` (128/128).
- New test `TheTreeRenderOverThreeProvidersIsCharacterForCharacterWhatItWas` pins the whole tree render over a three-provider record character for character (epoch normalised). It is a golden: it cannot fail before the change, because the change does not move the render — it is what catches the render drifting later while the reading tests stay green. The before/after proof is the artifact comparison above.

Closes #722


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3797,56 +3837,54 @@ public IReadOnlyList<TreeRow> TreeBatch(
...
+                             ? refPole.LabelVersus(refPlugin) : refPlugin;
+                }
+                bool isRef = refReader is null ? isWinner
+                           : refIsActiveProvider && string.Equals(plugin, refPlugin, StringComparison.OrdinalIgnoreCase);
+                if (isRef) { nodes.Add(new TreeNodeDelta(plugin, isWinner, true, Array.Empty<string>(), 0, true, null)); return true; }
+                var d = FieldsDiff.Compare(read, refFields!, referenceLabel: refLabel);
+                nodes.Add(new TreeNodeDelta(plugin, isWinner, false, d.Deltas, d.AgreedCount, d.Complete, null));
+                return true;
+            });
+            if (fill is null || nodes.Count == 0)
```
Avick3110: **A `versus=` pole that refuses now reports the wrong error, and loses Type/EditorId.**

The winner callback returns `false` *before* it adds a node, so on a `versus:` error `nodes` is always empty. This `if (fill is null || nodes.Count == 0)` therefore fires first and `continue`s, and the `versusError` row six lines below is unreachable.

Concrete case: `project.form="tree"`, `versus=SomePlugin.esp`, on a record that plugin does not touch. `MakePoleReader`'s named-arm reader returns `"'SomePlugin.esp' does not define or override 0x... — it has no version of this record. Touched by (active order, winner last): ..."`. Before this change the row carried `versus: <that message>` plus the winner's `Type` and `EditorId`; now it carries `"the provider bodies of 0x... could not be read."` with both null — a message that names the wrong cause (the bodies read fine) and gives the user no way to see what actually went wrong. Same for an excluded pole plugin and an off-order file that lacks the record.

That old `nodes.Count == 0` branch was effectively dead before (a key with `touchers.Count > 0` always resolved a tree), so nothing catches this; the suite covers this refusal only on the delta form (`RecordsComparisonFormTests.DeltaP4_...`), not on tree.

Swapping the two checks fixes it.

```suggestion
            if (versusError is not null)
            {
                rows.Add(new TreeRow(FormIdToken.Of(fk), fill?.Type, fill?.EditorId,
                                     touchers, null, Array.Empty<TreeNodeDelta>(), versusError,
                                     Array.Empty<ChildDeclarers>()));
                continue;
            }
            if (fill is null || nodes.Count == 0)
```

Avick3110: Folded — the versus check now runs before the empty-nodes one, and a test on the tree form covers the refusal (RecordsComparisonFormTests.ATreeVersusAPluginThatDoesNotTouchTheRecordRefusesNamingThePoleAndKeepsTheIdentity); it fails before the swap with "could not be read".

One correction to the finding's second half: Type and EditorId are carried on the row but neither render prints them on an error row — JsonWire.WriteTreeRow writes formid, error and touchers and returns, and the text lane prints formid + error. So nothing user-visible was lost there before or after; the visible half was the wrong cause, which is what the fix and the test are about. The fill still passes fill?.Type/fill?.EditorId so the row stays correct if a render ever shows them.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7426,6 +7466,36 @@ public WritePatchBuilder.MergeOutcome MergePlugins(
             }
             donorInfos.Sort((a, b) => a.Order.CompareTo(b.Order));
             var donorNames = donorInfos.Select(d => d.Name).ToList();
+            var transformSet = new HashSet<string>(donorNames, StringComparer.OrdinalIgnoreCase);
+
+            // ---- 2. masters = union(donor declared masters) − donors, load-order sorted (each donor's own header order
```
Avick3110, line 7696: **Scope / reviewability (low).** This hunk — hoisting the donor-master union and adding an up-front `view.PluginPath(mfn) is null` refusal (`#729`) — is unrelated to the tree-memory work this PR is about (`Closes #722`). It rides in commit `6e77e8b4` alongside the `FoldTreePinned` rewrite, is not mentioned in the PR body, and gets no `plugin/CHANGELOG.md` line even though it is user-visible: a merge whose donor declares an inactive master now fails at this point instead of after the record walk, so the refusal the caller sees for a merge that ALSO has unresolvable donor `.STRINGS` changes (master error now wins; previously the strings refusal did). The logic itself checks out — `resolveMasterPath` in `MergeBuild` is `view.PluginPath`, the same predicate, so no merge that used to succeed now refuses, and `MergeOutcome.Fail` (not `FailAfterFolder`) is correct here because `ResolveOutputPath` runs later. But per `CLAUDE.md` ("one change per commit") this belongs in its own commit/PR so it can be reviewed and changelogged on its own.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7426,6 +7466,36 @@ public WritePatchBuilder.MergeOutcome MergePlugins(
...
+            }
+            masterSet.Sort((a, b) =>
+                (orderIndex.TryGetValue(a, out var ia) ? ia : int.MaxValue).CompareTo(orderIndex.TryGetValue(b, out var ib) ? ib : int.MaxValue));
+            // The same predicate the serialize applies (MergeBuild resolves each master to an overlay and refuses when
+            // one is absent), asked from the headers instead of after the walk. MergeBuild still asks it: a master can
+            // go away between here and the write.
+            foreach (var mfn in masterSet)
+                if (view.PluginPath(mfn) is null)
+                    return WritePatchBuilder.MergeOutcome.Fail(
+                        $"cannot merge: donor master '{mfn}' is not active in the load order, so the references into it can't " +
```
Avick3110, line 7722: **Duplicated refusal text (low).** This sentence is now byte-for-byte identical to the one `WritePatchBuilder.MergeBuild` emits when `resolveMasterPath` returns null (`src/housecarl-core/WritePatchBuilder.cs`, the master-resolve loop). Two copies of the same user-facing string in two assemblies will drift the first time either is reworded, and the user then sees two different sentences for the same cause depending on whether the master went away before or after the record walk. Worth hoisting to one shared helper (or having this pre-check call into the same message builder) rather than keeping a hand-synced copy.

plugin/CHANGELOG.md
```diff
@@ -31,7 +31,6 @@ saying it sets an expectation their install may contradict. Say what is known, a
   apply peaked at 3.3 GB, took 73 s, and left the server sitting at 3.3 GB. The bodies are now gathered a plugin at
   a time, one walk each: the same apply takes 3.2 s, peaks at 0.9 GB and leaves the server back at its idle size.
   The written patch is unchanged — byte-identical on the same inputs.
-
 - **A write's per-edit line now reports what the written file holds, not what the call held in memory.**
```
Avick3110, line 60: **Stray whitespace edit (low).** The blank line that used to separate this entry from the one above it ("A bulk `housecarl_apply` ... byte-identical on the same inputs.") was deleted in commit `23f4e3a3`, which is otherwise about the versus-pole refusal. Every other entry in this list is separated by a blank line, so this one item is now tight against its predecessor and renders with different spacing from the rest of the list. Unrelated to the change; restore the blank line.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Removed — the whole #729 hoist is out of this branch (commit 7dfcf837); it arrived here through a git stash collision between parallel worktrees, not deliberately. #742 lands it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Moot here: the hunk carrying that second copy is gone from this branch (7dfcf837). The point stands for #742, which owns the change — worth raising there so the sentence has one home.

plugin/CHANGELOG.md
Avick3110: Restored in 7dfcf837 — the blank line was lost when I moved the entry to the end of Unreleased.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Correction to my earlier reply: #742 merged to main while this branch was being rebased, so main now carries the hoist legitimately. My removal commit would have deleted main's own code — it broke MergeMissingMasterPreflightTests on the rebase — so it is gone from the branch entirely (reset, not reverted). The branch now differs from main only in the tree fill, the resolver's StreamTree, the tests and the changelog; the merge pre-flight is exactly main's.

PR #746, opened 2026-09-14
Warn when a write forks a record another plugin already overrides

### LINKED ISSUE #565: Silent record-forking across patches, no proactive warning (actually caused a live bug)
### What you were trying to do

Over several sessions, built three separate, independently-created weapon patches (Jade Dragon Katana, Flowering Katana, Akaviri Warblade), each adding its own new weapon into the same shared vendor leveled list (RVT_LI_WeaponExotic50), intending for that list to accumulate a growing pool of reward weapons over time.

### What you tried

```text
for each patch, called bulk_apply/set_field on that shared list with into=<ThatPatch's own name>.esp — routing each edit into its own self-contained patch, since each weapon felt like an independent addition. Every individual call succeeded cleanly, and housecarl_check_errors reported all three patches as fully clean, both individually and swept across the whole load order.

What actually happened: because each patch carries its own independent override of the same record, each one is really a full fork of that record as it existed at the moment that patch was written — containing only that patch's own weapon, not the other two's. Skyrim's load order doesn't merge competing overrides of the same record; only the last-loaded plugin's version wins outright. So only one of the three patches' forks was ever actually live in-game; the other two patches' weapon additions silently never appeared, despite every edit reporting success. This was only caught later, by chance, when asked to reweight the list and explicitly reading that record's conflict_tree.
```

### What would have worked

houseCARL already has all the data needed to catch this automatically — it's the same conflict-tree information read_record conflict_tree=true / cross_plugin_query conflicts_only=true already expose on request. The gap is that nothing checks this proactively at write time. A bulk_apply/set_field call with into=<PatchX> targeting a record that's already overridden by a different plugin belonging to the same project could simply surface a warning — "this record is already overridden by <PatchY>.esp; continuing forks it, and only whichever loads last will apply — did you mean into=<PatchY>.esp instead?" — rather than succeeding silently. It shouldn't need to block the edit (deliberate forking is sometimes legitimate), just surface the ambiguity before it becomes a days-later in-game mystery instead of after.

Avick3110:
Skyrim never merges competing overrides of one record: only the last-loaded copy applies. So a patch that overrides a record a different patch already overrides splits it in two, and one half never reaches the game — which is what happened in #565, where three weapon patches each carried their own override of one shared leveled list, only the last-loaded fork was live, and every write reported success.

Every lane that lands an override — `housecarl_apply`, `housecarl_forward`, and the nested lane of `housecarl_create` (which overrides the child's parent in), each of them on a new patch, on `into=` an existing one, and on `in_place=` — now asks one more question before answering, and adds one `warning:` line to the response when the answer is yes. It never blocks the write: deliberate forking is legitimate.

**Two arms, because the hazard is not the same on both.**

- **The patch has a position** — `into=` an existing patch, or in place. Any plugin overriding the record at or below that position out-loads what is being written, so the edit can be the dead copy. Every such plugin is named, whoever wrote it, and the remedy names the lane that can edit it: `into=` for a houseCARL patch, `in_place=` for anything else.
- **The patch is new**, so it has no position yet, and it has just copied the winner's body in. The hazard #565 describes is a *sibling patch* — "a different plugin belonging to the same project" — so the fresh arm fires only for a plugin houseCARL made, judged by the `meta.ini` ownership marker `into=` itself is gated on; no name heuristic. That check moved into `HousecarlOwnerMeta.MarksOwned` so the extend lane's gate and this reader share one implementation. The sentence makes no claim about where the new patch will sort; the write renders' existing "a patch wins nothing until enabled in MO2" line speaks to that.

**Two bounds, stated rather than papered over.**

- It reads the **active** load order. A sibling patch that is installed but unticked in `plugins.txt` is not named — its records are not in the index, and finding them would mean opening every installed plugin on every write. The line says nothing about an unticked plugin either way.
- When one call's records are forked by **different** plugins there is no single `into=` that fixes them all, so the line names them and says to route each record into whichever of them already overrides it, rather than naming one and sending the caller to re-fork the rest. A long list of names is cut after three with a count, like the other set-valued blocks in these responses.

It costs no scan. The provider list is `TouchingPlugins`, the O(1) index lookup the read path's conflict tree is already built from, read off the same captured view the write already resolved its winners against — no body is fetched. The ownership marker is one `meta.ini` read per candidate *plugin* (a record has a handful), off the path the view already holds, memoised per call, and caught there (never inside the marker, where the extend gate needs the fault loud — `extend-resolve-guard` holds it to that). The new `ForkWarning` lives in its own file; the only resolver addition is `IndexView.OrderIndexOf`.

**Verified:** `dotnet build housecarl.sln -c Release`, `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` (2019 passed), `dotnet …/housecarl-generator.dll ci-all` (128/128), all after a rebase on `origin/main`.

`ForkWarningTests` (10) builds an order of a master, a houseCARL-owned patch A overriding one leveled list, and a third-party plugin overriding another. It covers: the fresh arm naming A; the json `warning` key; a dry run predicting it; a forward into a fresh patch; an in-place edit naming the plugin below its target; two records forked by different plugins getting no single `into=`; a fresh patch over a record only the third-party plugin overrides warning about nothing; extending A over that same record warning and offering `in_place=`; a write `into=` A itself; and a fresh patch over a record nobody else overrides.

**Run locally against the real install** (`E:\Authoria - Requiem Reforged`, ~2000 plugins), driving this branch's Release server over stdio with `HOUSECARL_DATA_DIR` pointed at a scratch config, every arm `dry_run=true` so nothing was written:

- `05E9D5:BecomeABard.esp` (a MagicEffect overridden by `Authoria - Master Patch - Bards Reborn Tweaks.esp`, a folder carrying `generated=true`), forwarded into a fresh patch → `warning: this record is already overridden by Authoria - Master Patch - Bards Reborn Tweaks.esp — only the last-loaded copy of a record applies, so this write forks it rather than adding to that copy; pass into="Authoria - Master Patch - Bards Reborn Tweaks.esp" to build on that copy instead.`
- `015C6E:Skyrim.esm` (ClothesFarm, overridden by `Update.esm`) into a fresh patch → no warning. That is the narrowing doing its job, and the reason for it: unnarrowed, this fired on nearly every write on this order.
- `10FAD9:Skyrim.esm` (DeathItemThalmor, override_depth 1) into a fresh patch → no warning.
- `000030:Aetherium Forge Destroys Items.esp`, overridden by a folder whose `meta.ini` carries `editedInPlace=` rather than `generated=true` → no warning, the fail-safe reading of the marker.

The first blind review's six findings are answered inline: five folded (the in-place lanes, the unreadable `meta.ini`, the name-list bound, the per-record remedy, and the active-only bound stated), one refuted (suppressing third-party providers on the fresh arm is the deliberate scoping, with the reason above).

Closes #565

---

**Second review folded** (`fde25a27`), all five findings:

- The bounded name list keeps the **last-loaded** three and counts the earlier ones — the last-loaded copy is the one that beats the write, so eliding it hid the only plugin that mattered.
- The per-record remedy carries the same `into=`/`in_place=` split the single-plugin one does.
- A single record forked by two plugins now takes the single-plugin remedy naming the last-loaded of them: per record the answer is its last-loaded forker, so uniformity is judged on that pick rather than on "exactly one forker".
- The middle clause is per arm — fresh says "forks it", positioned says "out-loaded by that copy rather than applying", because an in-place edit of a record the plugin already carries makes no second copy.
- `ForkWarning.For` now expands its own input to each target's container chain (`IndexView.ParentOf`, capped at 8 hops), so overriding a placed reference or an INFO asks the question about the cell, worldspace or topic it drags in too. That was the sharpest gap: nothing in the caller's op list mentions the container.

Tests are now 13 in `ForkWarningTests` (2054 in the suite), ci-all 128/128, rebased on `origin/main`.


src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -565,6 +573,10 @@ static PatchOutcome ApplyCore(
                 + string.Join("\n  - ", problems.Select(p => p.Message)));
         }
 
+        // Is another patch already overriding one of these records? Read off the same captured view, so it costs the
+        // index lookup and no scan. A warning, never a block — a deliberate fork is legitimate.
+        var forkWarning = ForkWarning.For(view, resolved.Select(r => r.edit.Target), fileName);
```
Avick3110, line 615: **The in-place lanes never get the warning, but the CHANGELOG says they do.**

`ForkWarning.For` is wired into `ApplyCore`, `ForwardRecordsCore` and `CreateRecordsCore` only. `ApplyInPlaceCore` (the `in_place=` lane) returns its `PatchOutcome` at ~L1142 (dry run) and ~L1213 with `Warning` never set, and `ForwardRecordsInPlaceCore` (~L1915) likewise. `CreateRecordsInPlace` is fine because it delegates to `CreateRecords`.

But `ForkWarning`'s own header comment and the CHANGELOG entry both promise it: *"The patch has a POSITION (into= an existing patch, **or in place**)"* / *"extending a patch (**or writing in place**) counts every plugin at or below it"*.

Scenario: `housecarl_apply in_place="SomeMod.esp" ...` on a record that a plugin loading after `SomeMod.esp` also overrides. The edit is the dead copy — exactly the #565 hazard, on the lane where the user cannot simply re-target — and the response says nothing. This is also the only lane whose remedy sentence (`in_place="X"`) the positioned arm was written for, so it is the lane the wording most obviously anticipates.

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,99 @@
...
+            var definer = fk.ModKey.FileName.String;
+            bool anyHere = false;
+            foreach (var p in touching)
+            {
+                if (string.Equals(p, patchFileName, StringComparison.OrdinalIgnoreCase)) continue;
+                if (string.Equals(p, definer, StringComparison.OrdinalIgnoreCase)) continue;
+                int idx = view.OrderIndexOf(p);
+                if (fresh)
+                {
+                    if (!IsOwnedPatch(view, p, ownedCache)) continue;   // its content came along with the winner
```
Avick3110: **The sibling patch this arm exists to catch is exactly the one it cannot see.**

The stated target is "another houseCARL patch, **possibly not enabled yet and therefore not the winner**, whose content this write did not carry". But the candidate set is `view.TouchingPlugins(fk)`, and the resolver index is built from `_paths` — "every **active** plugin's path" (`LoadOrderResolver.cs:118`); `ApplyInPlaceCore` even refuses with "is not an active plugin in the load order" off the same index. A patch that is installed but unticked in `plugins.txt` is not in the index, so it never appears in `TouchingPlugins` and can never be named here.

So the fresh arm only ever fires for an **active** sibling patch — and for an active sibling the write *did* carry its content if it was the winner, which is the reason the arm gives for excluding third-party plugins. Net effect: the case the narrowing was designed to keep (not-yet-enabled sibling) is undetectable, and the case it fires on is the one its own justification argues is benign. The `ForkWarningTests` fixture has `HcForkA.esp` active in both `loadorder.txt` and `plugins.txt`, so it exercises only the second.

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,99 @@
...
+/// whose copy can beat the one being written.</summary>
+public static class ForkWarning
+{
+    /// <summary>The warning for the records <paramref name="targets"/> names, or null when none of them is forked.
+    /// A provider never counts when it is the patch being written (<paramref name="patchFileName"/>) or the record's
+    /// own defining plugin; past that the two arms above decide.</summary>
+    public static string? For(LoadOrderResolver.IndexView view, IEnumerable<FormKey> targets, string patchFileName)
+    {
+        int patchIdx = view.OrderIndexOf(patchFileName);       // -1 = a new patch: no position yet
+        bool fresh = patchIdx < 0;
```
Avick3110, line 64: **"A new patch sorts to the top" is an assumption the pipeline does not enforce, and the fresh arm is silent when it is wrong.**

The fresh arm suppresses every non-houseCARL provider on the grounds that the patch "will sort to the top and has just copied the winner's body in". houseCARL writes the plugin into a new mod folder; where it actually lands in `loadorder.txt` is decided afterwards by MO2/LOOT or by the user. LOOT routinely sorts a fresh ESP above unrelated third-party mods.

Scenario: a fresh patch overrides a leveled list that `SomeOverhaul.esp` also overrides. The write reports success with no warning. The user then runs LOOT, which places the new patch before `SomeOverhaul.esp` — `SomeOverhaul.esp`'s copy wins and the edit never applies. That is #565 verbatim, with a third-party plugin in the sibling's place, and this is the one arm that will not say so.

At minimum worth saying out loud in the warning's absence (or gating the suppression on the provider actually being the winner whose body was copied, rather than on ownership).

src/housecarl-core/HousecarlOwnerMeta.cs
```diff
@@ -18,4 +18,25 @@ public static class HousecarlOwnerMeta
...
+    /// <c>into=</c> can extend? Fail-safe: a missing, unreadable or stripped marker reads as NOT owned.
+    /// <para>The one implementation, so the extend lane's ownership gate and any other reader of the same fact
+    /// cannot drift.</para></summary>
+    public static bool MarksOwned(string? folder)
+    {
+        if (string.IsNullOrEmpty(folder)) return false;
+        var meta = Path.Combine(folder, "meta.ini");
+        if (!File.Exists(meta)) return false;
+        bool inMarker = false;
+        foreach (var raw in File.ReadLines(meta))
```
Avick3110, line 37: **An advisory warning can now abort an otherwise-valid write.**

`File.ReadLines` throws on a `meta.ini` that exists but cannot be read — MO2 holding it open with a share-denying handle, an ACL the process lacks, a folder on a disconnected network path. In its original home (`LoadOrderService.IsHouseCarlOwned`) that fault happened on a gate whose job was to refuse, so throwing was tolerable. `ForkWarning.IsOwnedPatch` now calls it from inside the write pipeline, on a *best-effort advisory*, and nothing between `ForkWarning.For` and the tool boundary catches it.

Scenario: MO2 is open with a houseCARL patch's `meta.ini` locked; a `housecarl_apply` into a fresh patch over a record that patch also overrides throws `IOException` out of `ApplyCore` at `WritePatchBuilder.cs:578` and the whole write is reported as an engine error, when the only thing that failed was deciding whether to print one sentence. (Nothing is corrupted — the call sites all sit before the serialize — but a good write is lost.)

Wrap the `MarksOwned` call in `IsOwnedPatch` in a try/catch returning `false`, matching the class's own "a missing, unreadable or stripped marker reads as NOT owned" contract — `File.Exists` covers *missing*, not *unreadable*.

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,99 @@
...
+                }
+                else if (idx < patchIdx) continue;                      // it loads above the patch, so the patch wins outright
+                forkers[p] = idx;
+                anyHere = true;
+            }
+            if (anyHere) forkedRecords++;
+        }
+        if (forkers.Count == 0) return null;
+
+        var names = forkers.OrderBy(kv => kv.Value).Select(kv => kv.Key).ToList();
```
Avick3110, line 104: **The plugin list is unbounded, and every other set-valued block in these renders is budgeted.**

`forkers` accumulates across every target of the call and `Names(names)` joins all of them into one line. `WriteTools.Render` appends it *after* the op loop, outside `WriteSentences.Cap(maxChars)`, and `JsonWire` writes it as a plain `WriteNullable` outside the `truncated` budget — so unlike `ops`, `created` and the contested-parent block (which is explicitly bounded by `Wire.ContestedHostsShown` for exactly this reason), nothing caps it.

Scenario: a bulk `housecarl_apply` over a few hundred conflicted records in a ~2000-plugin order — the case the tool exists for — produces a single `warning:` line naming dozens or hundreds of distinct plugins, several KB of one sentence, and in the text render it can crowd out the response the host is already cutting. Bound it the way the contested-parent block is bounded ("… and N further plugin(s)").

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,99 @@
...
+        var last = names[^1];                                   // the last-loaded of them: the copy that currently applies
+        // The remedy names the lane that can actually edit that plugin. On the fresh arm every name is a houseCARL
+        // patch by construction, so into= is the whole answer; on the positioned arm the name may be anyone's.
+        var remedy = fresh
+            ? $"pass into=\"{last}\" to build on that copy instead."
```
Avick3110: **The remedy names one plugin for a set of records that may be forked by different plugins.**

`last` is the globally last-loaded forker across *all* targets of the call, but the forked records are not necessarily all forked by it.

Scenario: one `housecarl_apply` edits records X and Y into a fresh patch. X is already overridden only by `PatchA.esp` (order 40); Y only by `PatchB.esp` (order 90). The sentence reads "2 of the records written here are already overridden by PatchA.esp and PatchB.esp … pass into=\"PatchB.esp\" to build on that copy instead." Following it moves both edits into `PatchB.esp`, which still forks X against `PatchA.esp` — the caller does the remediation and the bug survives, silently, and re-running the call from `PatchB.esp` then warns about `PatchA.esp` in turn.

Either restrict the remedy to the single-forker case, or say which plugin forks which record.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Real, folded in 81cdea02.

`ApplyInPlaceCore` and `ForwardRecordsInPlaceCore` now compute the warning off their own captured view, right after resolve, and set it on both their dry-run and their real outcome. The target is an active plugin there, so it takes the positioned arm by construction: every plugin overriding those records at or below it is named, and the remedy sentence that arm already carried (`in_place="X"`) is the one it was written for.

`ForkWarningTests.AnInPlaceEditIsWarnedAboutThePluginsBelowIt` covers it: an in-place edit of a record a plugin below the target also overrides now names that plugin and offers `in_place=` for it.

src/housecarl-core/ForkWarning.cs
Avick3110: Real, and stated rather than fixed.

You are right that `TouchingPlugins` is the active index, so an unticked sibling patch can never be named. Closing that would mean opening every installed-but-inactive plugin on every write to see which records it carries — a whole-install scan on the write path, for an advisory sentence. That is a worse trade than the gap, so the bound is now said out loud instead of implied: the class comment states it, and the CHANGELOG entry and the PR body both say the warning reads the active load order and says nothing about an unticked plugin either way.

One correction to the second half. An active sibling patch is not always benign: the fresh patch copies the **winner**, and an active sibling that is *not* the winner (something below it wins) has content that did not come along — and even when it is the winner, the new patch can still sort above it and lose to it. So the arm does fire on real cases; the ones it cannot see are the unticked ones, which is now written down.

The header comment's "possibly not enabled yet" clause is gone, since that was the claim your finding shows the code cannot make.

src/housecarl-core/ForkWarning.cs
Avick3110: Refuting this one, with the sentence tightened.

The suppression is deliberate, not an oversight. #565 scopes the hazard to a project's own patches — "a different plugin belonging to the same project" — and on a ~2000-plugin order the unnarrowed rule fired on nearly every write: measured against the live install, a fresh patch over `015C6E:Skyrim.esm` warned about `Update.esm`, which is noise, not a finding. A warning that fires on almost every call is one nobody reads, and it would have buried the case the issue is about.

Your LOOT scenario is real but it is a different question — where a new patch will sort — and it applies to every record any third-party plugin overrides, which is most of them. That belongs in a sort/position check, not in this sentence.

What I did take from it: the warning must never *claim* the patch will win. It does not — it says only that the last-loaded copy applies, and the existing "a patch wins nothing until enabled in MO2, and a write into an existing mod keeps that mod's priority and may still need sorting above the current winner" line in the write renders is what speaks to position. The class comment now says the suppression is a scoping choice and that this sentence makes no claim about where the new patch will sort.

src/housecarl-core/HousecarlOwnerMeta.cs
Avick3110: Real, folded in 81cdea02 — and your instinct about *where* was right.

I first put the catch inside `MarksOwned` and `extend-resolve-guard` failed three checks: that probe deliberately holds the extend gate to "an unreadable folder mid-scan yields NO candidates, not the half the scan got to". Swallowing there turns "several patches carry this name" into a refusal naming one spelling as unambiguous. So the marker still throws, and the catch went exactly where you put it — `ForkWarning.IsOwnedPatch`, on `IOException` and `UnauthorizedAccessException`, reading as not-owned. `HousecarlOwnerMeta`'s doc now spells out that the two callers want opposite things from that fault and why.

src/housecarl-core/ForkWarning.cs
Avick3110: Real, folded in 81cdea02. The list is cut after three names with a count ("A.esp, B.esp, C.esp and 7 more"), the same shape as the contested-parent block. The bound is stated in the CHANGELOG entry too.

src/housecarl-core/ForkWarning.cs
Avick3110: Real, folded in 81cdea02. Your X/Y scenario is exactly right — following the sentence would have moved both edits into `PatchB.esp` and left X forked.

The single-plugin remedy is now given only when every forked record is forked by the same one plugin. Otherwise the sentence says: "they are not all forked by the same plugin, so route each record into whichever of those already overrides it — read the conflict tree (project.form='tree') to see which." `ForkWarningTests.RecordsForkedByDifferentPluginsGetNoSingleIntoRemedy` writes two records forked by two different plugins and asserts the response names both and offers no `into=`.

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+            if (here.Count == 0) continue;
+            forkedRecords++;
+            if (here.Count > 1) uniform = false;
+            else if (sole is null) sole = here[0];
+            else if (!string.Equals(sole, here[0], StringComparison.OrdinalIgnoreCase)) uniform = false;
+        }
+        if (forkers.Count == 0) return null;
+
+        var names = forkers.OrderBy(kv => kv.Value).Select(kv => kv.Key).ToList();
```
Avick3110, line 104: The name list is sorted **ascending** by order index and then cut after the first three, so the three plugins that load *earliest* survive and the ones that load *latest* are elided into "and N more" — but on the positioned arm the last-loaded forker is precisely the one whose copy beats the write, and therefore the one the caller most needs named.

Concrete: `into=` a patch at index 100 over a leveled list also overridden by `A.esp` (200), `B.esp` (300), `C.esp` (400) and `D.esp` (500). The sentence reads "...already overridden by A.esp, B.esp, C.esp and 1 more", hiding `D.esp` — the only one that actually wins in game. The remedy also falls to the generic branch here, so nothing else names it either.

Ordering the shown names by *descending* index (or keeping the last three) puts the winner in the sentence.

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,136 @@
...
+        // The remedy names the lane that can actually edit that plugin. On the fresh arm every name is a houseCARL
+        // patch by construction, so into= is the whole answer; on the positioned arm the name may be anyone's.
+        string remedy;
+        if (uniform && sole is not null)
+            remedy = fresh
+                ? $"pass into=\"{sole}\" to build on that copy instead."
+                : $"to build on that copy instead, pass into=\"{sole}\" if it is a houseCARL patch, else in_place=\"{sole}\".";
+        else
+            remedy = "they are not all forked by the same plugin, so route each record into whichever of those " +
```
Avick3110: The non-uniform remedy tells the caller to "route each record **into** whichever of those already overrides it", but on the positioned arm the named plugins may be anyone's — and `into=` is ownership-gated on the `meta.ini` marker, so that instruction is refused for a third-party plugin. The uniform branch two lines above is careful about exactly this (`into=... if it is a houseCARL patch, else in_place=...`); this branch loses it.

`ForkWarningTests.RecordsForkedByDifferentPluginsGetNoSingleIntoRemedy` is this case: one record forked by `HcForkA.esp` (owned) and the other by `HcForkForeign.esp` (not owned). Following the sentence for the second record means `into="HcForkForeign.esp"`, which the extend gate rejects, and the caller has to re-derive `in_place=` themselves.

Suggest mirroring the uniform branch's wording — "route each record into whichever of those already overrides it (`into=` if it is a houseCARL patch, else `in_place=`)".

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,136 @@
...
+                {
+                    if (!IsOwnedPatch(view, p, ownedCache)) continue;   // not the same-project hazard this arm names
+                }
+                else if (idx < patchIdx) continue;                      // it loads above the patch, so the patch wins outright
+                forkers[p] = idx;
+                here.Add(p);
+            }
+            if (here.Count == 0) continue;
+            forkedRecords++;
+            if (here.Count > 1) uniform = false;
```
Avick3110: `uniform` is set false by `here.Count > 1` (line 84), i.e. a **single** record forked by two or more plugins, so a one-record write can land in the generic branch. The composed sentence then reads:

> this record is already overridden by A.esp and B.esp — ... ; they are not all forked by the same plugin, so route each record into whichever of those already overrides it

There is only one record, so "they are not all forked by the same plugin" and "route each record" are both false on their face, and the caller is given no way to choose between A and B (the answer being the last-loaded one). Worth a third phrasing for the one-record-many-forkers case, or at least keeping the singular subject consistent.

src/housecarl-core/ForkWarning.cs
```diff
@@ -0,0 +1,136 @@
...
+        string remedy;
+        if (uniform && sole is not null)
+            remedy = fresh
+                ? $"pass into=\"{sole}\" to build on that copy instead."
+                : $"to build on that copy instead, pass into=\"{sole}\" if it is a houseCARL patch, else in_place=\"{sole}\".";
+        else
+            remedy = "they are not all forked by the same plugin, so route each record into whichever of those " +
+                     "already overrides it — read the conflict tree (project.form='tree') to see which.";
+        return $"{(forkedRecords == 1 ? "this record is" : $"{forkedRecords} of the records written here are")} already " +
+               $"overridden by {Names(names)} — only the last-loaded copy of a record applies, so this write forks it " +
```
Avick3110: The sentence body is fixed ("so this write **forks** it rather than adding to that copy"), but on the `in_place=` lane it is usually false. An in-place edit of a record the target plugin already carries creates no second copy at all — the hazard is that the existing copy is out-loaded by a plugin below, so the edit is dead, which is a different statement.

The PR's own `AnInPlaceEditIsWarnedAboutThePluginsBelowIt` is exactly this: the edit goes into `HcForkMaster.esm`, the record's *definer*, and the response still claims the write forks it. A reader who acts on "forks" will go looking for a duplicate record that does not exist.

The positioned arm probably wants its own middle clause — "...so this edit is out-loaded by that copy rather than applying" — while the fresh arm keeps "forks".

src/housecarl-core/WritePatchBuilder.cs
Avick3110, line 615: The apply/forward lanes ask the fork question only about `r.edit.Target` / `r.spec.Target`, but Phase 3 below overrides *more* than the target when the record is nested: `GenericGetOrAddAsOverride` on a placed reference, an INFO, a navmesh or a landscape drags the containing CELL (and WRLD) or DIAL into the patch as an override too. Those parent overrides fork the parent exactly as the target does — which is why the create lane's twin check (`parentPlans` at the bottom of `CreateRecords`) exists.

Concrete: `housecarl_apply` editing one REFR into a fresh patch, where a sibling houseCARL patch already overrides that REFR's CELL, writes a second CELL override and warns about nothing; only one of the two CELL copies will apply, which is the #565 failure on a record type where it bites harder than a leveled list.

The create lane already resolves its parent set; the apply/forward lanes would need the same parent keys fed into `ForkWarning.For` alongside the targets.

src/housecarl-core/ForkWarning.cs
Avick3110: Real, folded in fde25a27. The cut now keeps the LAST three and counts the earlier ones: your example reads "already overridden by 1 earlier plugin(s), then B.esp, C.esp and D.esp", so `D.esp` — the copy that actually wins — is in the sentence.

src/housecarl-core/ForkWarning.cs
Avick3110: Real, folded in fde25a27. The per-record branch now carries the same lane split: "route each record into whichever of those loads last over it (into= if it is a houseCARL patch, else in_place=)". The fresh arm omits that clause, where every name is owned by construction. `RecordsForkedByDifferentPluginsGetNoSingleIntoRemedy` asserts the clause is present.

src/housecarl-core/ForkWarning.cs
Avick3110: Real, and the fix goes deeper than the wording — folded in fde25a27.

Per record the answer is its LAST-LOADED forker: building on that one fixes the whole of that record's stack, because everything else named for it loads before it. So uniformity is now judged on that per-record pick rather than on "exactly one forker", and your case — one record, `A.esp` and `B.esp` — lands in the single-plugin branch naming `B.esp`, with the singular subject intact. The per-record branch is now reached only when two records genuinely answer to different plugins, where "route each record" is true.

`OneRecordForkedByTwoPluginsIsPointedAtTheLastLoadedOne` covers it, and asserts the remedy names the last-loaded of the two rather than the first.

src/housecarl-core/ForkWarning.cs
Avick3110: Real, folded in fde25a27. The middle clause is per arm now. Fresh keeps "so this write forks it rather than adding to that copy"; positioned says "so what this write puts in is out-loaded by that copy rather than applying", which is true whether or not a second copy exists. `AnInPlaceEditIsNotDescribedAsForkingTheRecord` asserts the in-place response says "out-loaded" and never "this write forks".

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Real, folded in fde25a27, and thank you — this is the one that would have bitten hardest, because nothing in the caller's own op list mentions the container.

Rather than feeding parent keys in per lane, `ForkWarning.For` now expands its own input: each target, then its container chain via `IndexView.ParentOf` (the O(1) containment edge), capped at 8 hops. That covers the REFR -> CELL -> WRLD and INFO -> DIAL cases for apply, forward and both in-place lanes at once, and the create lane's parent keys get the same climb for free.

`ForwardingANestedRecordWarnsAboutItsContainer` is your scenario in miniature: the master defines a topic with one line, `HcForkA.esp` overrides the TOPIC but not the line, and forwarding the line into a fresh patch now names `HcForkA.esp`. It fails before this change.

PR #747, opened 2026-09-14
Gather write-lane record bodies a plugin at a time

### LINKED ISSUE #723: [GAP] bulk apply and forward hold ~2 MB per operation, resident after the call: 6.2 GB for a 1.4 MB patch

Bulk `housecarl_apply` and `housecarl_forward` hold roughly **2 MB per operation**, resident after the call returns.

| call | ops | elapsed | peak WS | written |
|---|---:|---:|---:|---:|
| apply, 500 ops, dry run | 500 | 6.6 s | 1.7 GB (+0.8 GB) | — |
| apply, 500 ops → patch | 500 | 6.9 s | 1.9 GB | 293 KB |
| **apply, 2,484 ops → patch** | 2,484 | 39.7 s | **6.2 GB** | 1.4 MB |
| apply, 2,484 ops, cold process | 2,484 | 49.3 s | 6.1 GB | same shape |
| forward, 500 records from Skyrim.esm | 500 | 15.7 s | (6.2 GB carried from the previous call) | 282 KB |
| **forward, 2,484 records** | 2,484 | 68.4 s | **5.1 GB** | 1.4 MB |
| compact off-order, 11,190 originating records (real renumber) | — | 27.0 s | no rise | 3.9 MB |

A 1.4 MB patch should not cost 6 GB to build. The dry run costs the same as the write, so it is the per-op materialisation (a full record copy plus its overlay, probably) and not the serialize. The merge, by contrast, handles 25,063 originating records at 1.4 GB, and compact renumbers 11k records with no rise, so the write engine's per-record cost is not inherent — it is the apply/forward path.

The resident-after-return part matters as much as the peak: a session that runs one big apply then a merge starts the merge from 6 GB, which is how the earlier measurement's merge50 peak went from 1.2 GB to 3.2 GB.

Avick3110:
A bulk `housecarl_apply` or `housecarl_forward` fetched each record's body one at a time, and a fetch (`IndexView.GetRecord` → `SeekBody`) finds a record by enumerating its plugin from the top — Mutagen builds an overlay wrapper for every record it steps over, so the cost of one op is the winner plugin's whole record count, not the one record it wanted. That is where the reported ~2 MB per op came from, and why the dry run cost the same as the write: the fetch is Phase 1, before anything is serialized. It is churn rather than a leak — nothing is held on the service — but on a 2,484-op call it is gigabytes of it, and Server GC does not hand the heap back, which is the "resident after the call returns" part.

The lanes now declare every (plugin, record) the call will need and gather them a plugin at a time through the resolver's existing `CollectRecords`, which answers many keys in one walk. `BodyGather` is the bookkeeping for that: `Want` / `Gather` / `Body`, with `Body` falling back to the single fetch for a pair that was never gathered or a plugin the walk could not read, so no answer and no exception can differ from before. Declared under the same guards each loop refuses on, so a refusal that used to cost nothing still costs nothing. The fallback is answer-equivalent but costs the old load, so an out-of-memory fault in the walk is rethrown rather than caught — paying N whole-plugin walks because memory ran out makes the failure worse — and the plugins whose walk did fault are readable on `Faulted`, so a caller can tell the slow path from the fast one.

Applied to the three lanes this issue names — `Apply`, `ApplyInPlace` and `ResolveForwardSources` (both forward lanes) — for winner bodies and for in-order `CopyFrom` sources. **Not** every per-record fetch in the file: the create lane's parent-body read (`ParentBodyFrom`) still walks a plugin per distinct parent behind its memo, which is #757; folding the three sibling gathers in the tree into one primitive is #756.

The comment in `ForwardRecordsCore` that described this exact fix as "deferred until measured" is replaced; it has now been measured.

## Measured

Local run, not CI. MO2 instance `E:\Authoria - Requiem Reforged`, 3,254 active plugins, 8C/16T. Each arm is its own cold `housecarl-mcp` process driven over stdio; the order index is built by a warm-up call outside the measured window, and the child process's working set is sampled every 250 ms from outside — the peak is the max inside the call's window, and "resident after" is the working set 6 s after the call returned. Arms: 500 and 2,484 ARMO records of `Skyrim.esm` (`Set Value`, and `forward source=Skyrim.esm`). Idle after the warm-up is 700–894 MB (the order index itself), so a resident-after figure at that level means the call left nothing behind.

| arm | before: elapsed | peak WS | resident after | after: elapsed | peak WS | resident after |
|---|---:|---:|---:|---:|---:|---:|
| apply 500 ops, dry run | 14.0 s | 1,279 MB | 1,284 MB | **1.3 s** | **846 MB** | **793 MB** |
| apply 2,484 ops → patch | 72.9 s | 3,293 MB | 3,294 MB | **3.2 s** | **881 MB** | **735 MB** |
| forward 500 from Skyrim.esm | 15.0 s | 1,267 MB | 1,273 MB | **1.9 s** | **876 MB** | **761 MB** |
| forward 2,484 from Skyrim.esm | 74.5 s | 3,495 MB | 3,496 MB | **2.3 s** | **919 MB** | **920 MB** |

The three arms that write produced **byte-identical** patches on the two builds (SHA-256 of `ZZapplymemBaseB/C/D.esp` vs `ZZapplymemNewB/C/D.esp`: 906,854 / 185,170 / 879,531 bytes, equal hashes). The throwaway mod folders were deleted afterwards; no profile list was touched.

The per-op allocation was also confirmed to scale with the winner plugin's record count rather than the record, on a synthetic single-plugin order: 200 ops cost 225 / 595 / 1,407 KB per op against a 500 / 2,000 / 8,000-record plugin before, and 46 / 48 / 52 KB per op after.

## Tests

`BulkWriteCostTests` is the one that fails if the gather stops working: a bulk apply and a bulk forward over a single-plugin fixture must grow the resolver's `CollectPasses` by exactly 1 and its `BodySeeks` by 0, and the total must not change between a 2-op call and a 300-op one. Checked by commenting out the three `gather.Gather()` calls: these three tests fail and the other four still pass, which is exactly the hole in the first round of tests.

`BodyGatherEquivalenceTests` asserts the gather's own claim without depending on the file format: for every (plugin, record) a lane asks for, `BodyGather.Body` answers what `IndexView.GetRecord` answers, including the same null for a record the plugin does not hold, and an undeclared pair still answers through the fallback.

`BulkWriteBytesTests` writes a bulk apply and a bulk forward over a 300-record fixture and asserts the SHA-256 of each patch against the hash the pre-change build produced. Those digests pin the serializer as well as the gather, so the class says how to tell the two apart — if the equivalence test passes and this one fails, the bytes moved downstream of the gather — and the failure message prints the digest to re-record.

Existing apply and forward tests are unchanged: 2,016 passed, `ci-all` all pass.

Closes #723


src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -1754,6 +1783,13 @@ static RemovalOutcome RemoveRecordsInPlaceCore(
         // answer. Memoized per CALL, not per resolver: the explainer reads the profile fresh by design, and a cache
         // living longer than one refusal batch would reintroduce the staleness it exists to avoid.
         var absenceMemo = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        // One walk per SOURCE plugin for the whole call, not one per record (#723): the per-record fetch enumerated
+        // from_plugin from the top each time, so 2,484 records out of one plugin cost 2,484 whole-plugin walks.
+        var gather = new BodyGather(view, session);
+        foreach (var s in specs)
```
Avick3110, line 1794: The want-pass here declares a body for every spec, but the loop below refuses several classes of spec *before* it ever calls `gather.Body` — a duplicate `Target` (the `seen` check), `sourceIsSelf`, and an origin master that is not active. For those specs the gather still walks `s.FromPlugin`, and because the wanted key is normally *not* present in that plugin in the origin-inactive case, `CollectRecords` never reaches `got == want.Count` and walks the plugin **to the end**.

Concrete case: `housecarl_forward` with `source=` naming the output patch itself, on an order where that patch is active. The old code refused on the `sourceIsSelf` guard without opening or enumerating anything; now the gather opens that plugin's overlay and enumerates it fully first, and only then is the call refused with nothing written. Same for a spec whose record originates in a disabled plugin — a named refusal that used to be free now pays a whole-plugin walk, which is exactly the cost #723 is about.

Mirroring the loop's three cheap guards in the want-pass (skip duplicates, skip `FromPlugin == fileName`, skip an inactive `originMaster`) keeps the gather's set equal to the set the loop actually fetches, which is the invariant the Apply lane's comment already claims for itself.

src/housecarl-core/BodyGather.cs
```diff
@@ -0,0 +1,63 @@
...
+    public void Gather()
+    {
+        foreach (var (plugin, keys) in _declared)
+        {
+            if (_walked.Contains(plugin)) continue;
+            if (!_held.TryGetValue(plugin, out var sink)) _held[plugin] = sink = new Dictionary<FormKey, IMajorRecordGetter>();
+            // A plugin that opened at index time but cannot be opened now: leave it to Body's per-record fallback,
+            // which raises the same fault the caller saw before this existed.
+            try { _view.CollectRecords(_session, plugin, keys, null, sink); }
+            catch (PluginUnreadableException) { continue; }
```
Avick3110: Only the *open* fault is caught here. `CollectRecords` deliberately leaves a fault from the **walk** unwrapped (its doc says so), and `WinnerBodies.For` — the sibling that does the same job — catches that second case explicitly (`catch (Exception ex) { unreadable[plugin] = new PluginUnscannableException(plugin, ex); }`), as does `BodyPrefetch.Chunk.Body`.

Because `Gather()` now runs as an up-front pass, a walk-phase fault (a record Mutagen cannot parse in a source plugin) escapes before a single edit has been pre-flighted, so the caller gets a raw exception instead of the lane's all-or-nothing `refused — N of M ... NOTHING written` report that names the record. Previously the same fault surfaced from `GetRecord` inside the loop, after the earlier edits' problems had been collected.

Catching the walk fault and falling through to `Body`'s per-record fetch (leaving `_walked` unset, as the unreadable arm already does) keeps the stated contract — "the exception a caller used to see is the one it still sees" — for both fault kinds rather than just the open.

src/housecarl-core/BodyGather.cs
```diff
@@ -0,0 +1,63 @@
...
+/// a time therefore allocates megabytes PER OP: a 2,484-op apply on a real order churned about 2 MB an op, which the
+/// GC heap then holds after the call (#723). <see cref="LoadOrderResolver.CollectRecords"/> already answers many keys
+/// from one plugin in one pass; this is the bookkeeping that lets a caller who discovers its wants one at a time use
+/// it — declare every (plugin, record) the call will need, gather once, then read them back.</para>
+///
+/// <para>Answers cannot differ from the one-at-a-time path: <see cref="Body"/> falls back to the single fetch for a
+/// pair that was never declared, and for a plugin <see cref="Gather"/> could not open — so the exception a caller
+/// used to see is the one it still sees.</para>
+/// </summary>
+public sealed class BodyGather
```
Avick3110, line 28: This is the third implementation of "gather many bodies, one walk per plugin": `HousecarlCore.WinnerBodies` (#251, keyed by FormKey, winner plugin derived internally) and `HousecarlMcp.BodyPrefetch.Chunk` (#582, keyed by (plugin, FormKey) with a lazy per-plugin walk and a per-record fallback) already exist, and `BodyPrefetch.Chunk` is very nearly this class with the walk deferred to first use instead of an explicit `Gather()`.

Three copies means three different answers to the same questions — which exceptions are swallowed, whether an ungathered pair falls back, whether a gathered-and-absent pair is a real null — and they already disagree (see the catch above). Worth folding these into one primitive (or at least saying in the comment why `BodyPrefetch.Chunk` could not be lifted into core and reused), rather than leaving the write lanes and the read lanes on separate bookkeeping.

src/housecarl-mcp-tests/BulkWriteBytesTests.cs
```diff
@@ -0,0 +1,96 @@
...
+/// (<see cref="BodyGather"/>, #723) replaced the per-record fetch. The hashes below were taken from the build
+/// before that change, so a difference here means the gather changed the product and not just its cost.
+/// </summary>
+[Trait("tier", "integration")]
+public sealed class BulkWriteBytesTests : IDisposable
+{
+    const string ApplySha = "BD730E59DEBF0009D30EE216D138283EC599B8ECA7CA2E2EBE487B72CD9B3A1A";
+    const string ForwardSha = "0B4805A31399C3DFA2FC4DC328FD0398C7D850C5CBED6EA3BEFA072A7018AA57";
+
+    const int Records = 300;
```
Avick3110, line 29: These hardcoded digests pin the exact bytes Mutagen's serializer emits, not anything this change owns. A Mutagen upgrade, or any unrelated change to how the write lanes stamp a patch header (masters ordering, `NextFormID` floor, ONAM), turns both tests red with the message "bulk apply wrote different bytes than before the body gather" — pointing the next reader at `BodyGather`, which will not be the cause. There is also no way to regenerate the expected value from the test itself, so whoever hits it has to check out the pre-change build to find out whether the difference is real.

A cheaper equivalent that survives the library: write the same patch twice in one test — once through the gather and once through a path forced onto the per-record fallback (e.g. build the edits so nothing is declared, or expose a "no gather" switch) — and assert the two files hash equal. That asserts the actual claim ("the gather did not change the product") without freezing Mutagen's output format.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Folded in 965d0056. The want-pass now mirrors the loop's three cheap guards — a repeated target (its own `wantSeen` set, so it does not consume the loop's `seen`), a `FromPlugin` equal to the file being written, and an origin master that is neither the output nor active — so the gather's set is the set the loop actually fetches and a free refusal stays free.

The origin-inactive case you describe is the worse of the two, for the reason you give: the key is normally absent from that plugin, so `CollectRecords` never short-circuits and walks to the end.

src/housecarl-core/BodyGather.cs
Avick3110: Folded in d39a9a09. `Gather` now catches any fault, not just the open, and leaves the plugin unwalked so `Body` falls back to the per-record fetch — which raises the same fault from inside the caller's own loop, where the edit that owns it can be named. The stated contract held only for the open fault; it holds for both now.

Deliberately not the `WinnerBodies` shape (record the cause and report it): a write is all-or-nothing and has no place to report a partially-read plugin, and the per-record re-raise gets the refusal that names the record, which is what the lane's report is for.

src/housecarl-core/BodyGather.cs
Avick3110: Half folded, half deferred.

Folded: the disagreement you point at is gone — `Gather` now handles a walk fault the way `WinnerBodies.For` does, by not pretending the plugin was read (d39a9a09). And the class comment now names both siblings and says what each guarantees and why: `WinnerBodies` reports an unreadable plugin because a scan has to say which plugins it could not read; `BodyPrefetch.Chunk` walks lazily and answers null for anything ungathered because a render stops mid-chunk and must not pay for rows nobody sees; a write reads every body it declared, all-or-nothing, and must raise each fault from the edit that owns it, which is why this one is eager with a per-record fallback.

Deferred: `BodyPrefetch.Chunk` could not be lifted as-is — it is `internal` to the server assembly and core cannot see it, and its null-for-ungathered contract would turn a pair a write lane failed to declare into a spurious refusal, which is the failure mode the fallback exists to prevent. Folding the three into one primitive with lazy/eager and report/fallback as options is a real change and touches the read lanes' render budget and cancellation, so it is #756 rather than part of a write-lane memory fix.

src/housecarl-mcp-tests/BulkWriteBytesTests.cs
Avick3110: Folded the substance, kept the digests. 1f35068b.

You are right that the digests pin the serializer and would blame the gather for a Mutagen bump, so the claim now has its own test that does not depend on the file format: `BodyGatherEquivalenceTests` asserts that for every (plugin, record) a lane asks for, `BodyGather.Body` answers what `IndexView.GetRecord` answers — same record, same fields, same null for one the plugin does not hold — over a fixture where a second plugin holds only some of the wanted keys, plus that an undeclared pair still answers through the fallback. That is the "the gather did not change the product" assertion, and it survives the library.

The digests stay because the issue asked in so many words for a test that the written patch is byte-identical to the pre-change output, and they are the only thing that can carry a comparison against a build that no longer exists. What changed is that a failure no longer misdirects: the class says that if the equivalence test passes and this one fails the bytes moved downstream of the gather, and the failure message prints the digest this build produced, which is the value to re-record.

I did not take the write-it-twice form: the second path would need a "no gather" switch in production code that exists only for the test, and the fallback it would exercise is the same `IndexView.GetRecord` the equivalence test already compares against directly.

Separately, the real-order run in the PR body is the same assertion at full scale: the three writing arms produced byte-identical patches on the pre- and post-change builds (906,854 / 185,170 / 879,531 bytes, equal SHA-256).

Avick3110:
Two findings that came in outside the inline comments:

**bulk_create's `ParentBodyFrom` still fetches per record.** Correct, and the PR body overstated "all three lanes". Fixed the wording: this PR covers the three lanes #723 names, and the create lane's parent-body read is filed as #757 with the shape the fix would take (declare each parented spec's `(definer, parentFk)` under the same guards the loop refuses on, then let the winner fallback stay a single fetch). Not done here because this PR measured apply and forward only, and that lane's parent resolution has in-place, patch-carried and sibling branches that deserve their own measurement rather than riding on someone else's numbers.

**Stale doc comment on `ApplyInPlace`.** Folded in 965d0056 — the content-source bullet said `view.GetRecord(session, target, fk)`, which is no longer how the body is read; it now names the gather and says it walks the target once for the whole call.

Rebased on `origin/main`, rebuilt, re-run: 2,005 tests passed, `ci-all` 128/128.

src/housecarl-mcp-tests/BodyGatherEquivalenceTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        var gather = new BodyGather(view, session);
+        foreach (var k in _keys) { gather.Want(_masterName, k); gather.Want(_replName, k); }
+        gather.Gather();
+
+        using var plain = _resolver.OpenSession();
+        var plainView = _resolver.Capture();
+        foreach (var k in _keys)
+            foreach (var plugin in new[] { _masterName, _replName })
+                Assert.Equal(Describe(plainView.GetRecord(plain, plugin, k)), Describe(gather.Body(plugin, k)));
```
Avick3110, line 78: Nothing in this PR can fail if the gather stops working.

`BodyGather.Body` falls back to `IndexView.GetRecord` for any pair it does not hold, and `Gather()` swallows every fault, so a build where `Gather()` throws on the first plugin — or where a declaration loop drifts and stops declaring the right (plugin, record) pairs — answers *every* call through the old one-at-a-time fetch. This test still passes, because the thing it compares against is exactly that fallback. `BulkWriteBytesTests` also still passes, for the same reason: the bytes are identical either way. So the whole point of #723 can regress silently and CI stays green.

The resolver already counts what would catch it: `LoadOrderResolver.CollectPasses` and `LoadOrderResolver.BodySeeks`, used this way in `RecordsWalkCostTests` and `RecordsRenderCostTests`. A test that runs a bulk apply over a single-plugin fixture and asserts `CollectPasses` grew by 1 while `BodySeeks` stayed flat would pin the claim the PR is actually making.

src/housecarl-core/BodyGather.cs
```diff
@@ -0,0 +1,72 @@
...
+        foreach (var (plugin, keys) in _declared)
+        {
+            if (_walked.Contains(plugin)) continue;
+            if (!_held.TryGetValue(plugin, out var sink)) _held[plugin] = sink = new Dictionary<FormKey, IMajorRecordGetter>();
+            // Any fault — a plugin that opened at index time but cannot be opened now, or a record the walk cannot
+            // parse — leaves this plugin unwalked, so Body falls back to the per-record fetch and the caller sees the
+            // same fault, from the same place in its own loop, that it saw before this existed. Swallowing it here
+            // would move a named per-record refusal to an up-front throw that names no record.
+            try { _view.CollectRecords(_session, plugin, keys, null, sink); }
+            catch (Exception) { continue; }
```
Avick3110: The catch-all silently restores the exact pathology this PR removes, and nothing anywhere says so.

Concrete case: on a large winner plugin the gather walk throws `OutOfMemoryException` (or any transient IO fault). The plugin is left unwalked, so every one of the 2,484 declared records now falls back to `IndexView.GetRecord` — 2,484 full-plugin enumerations, i.e. the 3.3 GB / 73 s behaviour from #723, entered *because* memory was already tight, and with no counter, note or message distinguishing it from the fast path. The cheaper faults have the same shape: a `PluginUnreadableException` out of `session.Overlay` costs the call N walks instead of 1 and the user sees only a slow call.

The per-record fallback for answer-equivalence is right; what is missing is any signal that it happened. At minimum, not catching `OutOfMemoryException` would stop the degraded path from making a memory fault worse, and recording the faulted plugin so the outcome can carry a note would make the degradation observable.

src/housecarl-mcp-tests/BulkWriteBytesTests.cs
```diff
@@ -0,0 +1,103 @@
...
+/// and every later decision about how a patch's header is stamped, not only the gather. Read a difference as "the
+/// bytes moved", then find out which: <see cref="BodyGatherEquivalenceTests"/> asserts the gather's own claim (a
+/// gathered body is the body the single fetch returns) without depending on the format, so if IT passes and this
+/// fails, the cause is downstream of the gather — a library bump, a header change — and the digest below is what
+/// needs re-recording. The failure message prints the digest this build produced, which is the new value.</para>
+/// </summary>
+[Trait("tier", "integration")]
+public sealed class BulkWriteBytesTests : IDisposable
+{
+    const string ApplySha = "BD730E59DEBF0009D30EE216D138283EC599B8ECA7CA2E2EBE487B72CD9B3A1A";
```
Avick3110, line 26: These two digests are the whole serialized plugin, so they are a tripwire on Mutagen's serializer, not on the gather — the class comment says as much. In practice that means the next Mutagen bump (or any change to how a patch header is stamped) fails two tests that have nothing to do with the change that broke them, on a branch whose author has no reason to look at #723.

And the equivalence they are meant to protect is already unfalsifiable here: `BodyGather.Body` falls back to `GetRecord`, so a build where the gather does nothing writes the same bytes and this test passes too (see the comment on `BodyGatherEquivalenceTests`).

If the golden is kept, consider hashing the record bytes rather than the whole file, so a header-format change does not fire it.

src/housecarl-mcp-tests/BodyGatherEquivalenceTests.cs
Avick3110: Folded, and you were exactly right about what would happen. `BulkWriteCostTests` runs a bulk apply and a bulk forward over a single-plugin fixture and asserts `CollectPasses` grew by 1 and `BodySeeks` by 0, plus that the two counters' total is the same for a 2-op call and a 300-op one — the property the per-record fetch broke.

Checked that it can fail, by the route you describe: comment out the three `gather.Gather()` calls, and the three cost tests fail ("200 edits out of one plugin walked it 0 times") while `BodyGatherEquivalenceTests` and `BulkWriteBytesTests` all four still pass. That is the hole, reproduced and now closed.

src/housecarl-core/BodyGather.cs
Avick3110: Folded both halves.

`OutOfMemoryException` is rethrown rather than caught, for the reason you give: the fallback costs one whole-plugin walk per declared record, so entering it because memory already ran out turns a fault into the 3.3 GB behaviour this PR removes. The comment says that in as many words.

The degradation is now observable: a plugin whose walk faulted goes into `Faulted`, which the caller can read. I stopped short of carrying it into the response — the issue's scope is explicitly no response change — so today it is a readable fact rather than a rendered note; wiring it to an outcome note is a small follow-up if that reads better.

I did not narrow the catch to a list of record-parse exception types. `CollectRecords`'s contract is that a walk fault comes through unwrapped and Mutagen does not promise a closed set of types for a malformed subrecord, so a type list would be a guess that silently stops catching when Mutagen adds one — the exact drift the review above flags between the three gathers. Both siblings (`WinnerBodies.For`, `BodyPrefetch.Chunk.Body`) catch broadly for the same reason. Excluding OOM removes the case where catching broadly is actively harmful; if there is a second such case, name it and it gets the same treatment.

src/housecarl-mcp-tests/BulkWriteBytesTests.cs
Avick3110: The same point was raised in the first round and answered there (#discussion_r4010321393); the reasoning has not changed, but your second half has been fixed.

The unfalsifiability half is real and is now closed, though not in these tests: `BulkWriteCostTests` asserts the resolver's `CollectPasses` / `BodySeeks` counters, so a build where the gather does nothing fails — verified by commenting out the three `gather.Gather()` calls, which fails those three and leaves these four passing.

The digests stay. The issue asked in so many words for a test that the written patch is byte-identical to the pre-change output, and a digest recorded from that build is the only thing that can compare against a build no longer in the tree. The misdirection is handled rather than ignored: the class says a failure means "the bytes moved" and how to tell which — if `BodyGatherEquivalenceTests` passes and this fails, the cause is downstream of the gather — and the message prints the digest to re-record.

Not switching to hashing the record bytes only: extracting them means a test reimplementing part of the layout it is supposed to be checking, and it would discard the recorded pre-change value, which is of the whole file. The full-scale version of the same assertion is in the PR body — the three real-order arms wrote byte-identical patches on the two builds.

PR #748, opened 2026-09-14
Keep an unparseable record as a walk boundary

### LINKED ISSUE #725: [Bug] a forward walk dies with a raw MalformedDataException on the order's one unscannable record; the scan lanes exclude it
Measured 2026-09-11 on a 3,801-plugin order, the installed 2.0.0 server in its own process.

A forward `walk=` with `follow:"*"` and a budget large enough to reach `Requiem - Special Feats.esp 00080E` dies with a raw Mutagen exception after 30 s and 5–6 GB:

```
MalformedDataException: PerkEntryPointModifyActorValue did not have expected parameter type flag: Float
```

The scan lanes exclude that record cleanly and report it as unscannable in the accounting (every `check` and every merge names it, once). The walk does not: it enters the record, throws, and the whole call fails through the generic internal-failure wrapper with no sentence about which record, what to do, or that `walk.exclusions` with `severity:"stop"` on `Perk` gets past it (it does — that was the workaround used to reach the OOM measurement).

| call | budget | elapsed | peak WS | result |
|---|---|---:|---:|---|
| forward `*`, chain, from an NPC | depth 1000 / 1e6 | 30.3 s | 5.4 GB | raw `MalformedDataException`; 5.3 GB still held after |
| same, summary counts_only | same | 30.0 s | 6.4 GB | same crash — so the walk, not the render |
| same, `exclusions:[{match:"Perk",severity:"stop"}]` | 10,000 | 21 s | 3.4 GB | runs |

Two parts:

1. The walk should treat an unreadable node the way the scan does: record it as a boundary in the response (record, exception name) and continue. Cornerstone: a tool never fails silently, and it never fails raw either.
2. The Mutagen coverage gap on that record (a `PerkEntryPointModifyActorValue` whose parameter type flag is not Float) is upstream and worth its own line to Mutagen if it is not already there; every arm of the battery met it.

Avick3110:
A forward `walk=` that reached a record Mutagen cannot parse entered it, the lazy parse of its content threw, and the whole call came back through the generic internal-failure wrapper — no record named, no remedy, nothing about the exclusion that got past it. The scan lanes already answer this shape: they isolate the fault per record, keep the record out, and account it as unscannable. The walk now does the same. Reading a node's links is wrapped per record, and a node whose content will not parse is recorded as a boundary — its FormID, type and EditorID, `status=kept`, and one note carrying the scan lanes' own "could not be scanned (Mutagen could not parse its content)" wording plus the exception type and message — after which the walk carries on with every other chain. The seed's own first hop is wrapped the same way, so a walk seeded on such a record names it instead of raising. No new parameter: the memo that lets two seeds share one read remembers the fault the way it remembers a node's links, so the record is read once.

Closes #725

## Test

`src/housecarl-mcp-tests/RecordsWalkUnscannableTests.cs` builds its own MO2 instance holding an NPC that points at a race, a sound perk, and a perk whose one entry-point effect has its EPFT parameter-type flag byte corrupted on disk — `ProbeBytes.CorruptEpftBytes`, the same fixture the scan-lane guards use, which produces exactly the reported `MalformedDataException`. The world asserts up front that the fixture still throws from `EnumerateFormLinks`, so a green is not vacuous. Reverted against the fix, both tests fail with the issue's own sentence (`error: housecarl_records failed unexpectedly — MalformedDataException: PerkEntryPointModifyActorValue did not have expected parameter type flag: 99`). Full suite: 2003 passed; `ci-all` 128/128.

## The real arm, on the live order

Run locally against `E:\Authoria - Requiem Reforged` (3,571 modlist lines) with this branch's Release `housecarl-mcp` driven over stdio in a private process, working set sampled every 250 ms from outside, the whole call bracketed. Warm-up read first, so the arm excludes the order-index build.

| call | budget | elapsed | peak WS | peak private | result |
|---|---|---:|---:|---:|---|
| forward `*`, chain, from `0A2C94:Skyrim.esm` | depth 1000 / max_nodes 250000 | 1,074 s | 1,869 MB | 1,779 MB | completed — 249,027 nodes reached, 0 errors, one unscannable boundary |

The issue's arm asked for `max_nodes` 1,000,000; that is now refused at the 250,000 hard ceiling added since (#727), so the arm was run at the ceiling. It is above this seed's whole closure — 249,027 nodes, under the cap — so the walk finished rather than being cut.

The boundary is the record from the issue, named once and not entered:

```
00080E:Requiem - Special Feats.esp  Perk  Feat_Perk_Skill_HeavyArmor_DevastatingTackle  [kept]
  could not be scanned (Mutagen could not parse its content) and was not entered —
  MalformedDataException: PerkEntryPointModifyActorValue did not have expected parameter type flag: Float
```

The issue's comment says the excluded walk then ran to the #719 OOM. That is no longer what happens: with the record excluded as a boundary rather than by hand, the same unbounded walk ran to completion and the process held 1.87 GB at its peak and 1.64 GB after the call — no runaway, no OOM. That is #731's release of a pass's bodies doing its half of the job; this change is what lets the walk reach it. An earlier identical run of the same arm took 809 s; the 1,074 s above is with other measuring lanes busy on the same machine, so the elapsed figure is a ceiling, not a clock.

## Part 2 (upstream, not this PR)

A Mutagen issue for the parse gap exists: Mutagen-Modding/Mutagen#456, "Remove Perk EPFT malform record throws", open since 2023 and labelled good-for-first-contribution. Nothing filed.

## Review fold (2026-09-15)

Six findings from the blind review, five folded whole and one in part; each has a reply on its own thread. The guard now also covers the template read (`FactOf` carries the fault instead of throwing, so the template lane and its report name the record too), lets cancellation and out-of-memory through rather than recording them as parse faults, remembers a seed's fault in the shared memo, records it once per record and holds the row to the node budget, and keeps the depth-cap note on a memoised boundary at the cap. Tests: the reading-form test renders rows instead of counting, and two new tests seed on the unparseable record (with and without `seed_paths`); a fourth covers the template lane on an NPC whose ACBS is cut short. Rebased on main (the cycle-report work landed meanwhile), rebuilt, 2014 tests pass and `ci-all` is 128/128. The measured arm was not re-run: it is a closure walk, and nothing on that path changed behaviour.

## Second review fold (2026-09-15)

Four more findings, all on the seed path, all folded. The frontier is no longer cleared on a seed fault, so chains an earlier `seed_paths` entry proved are kept; every path is evaluated and answers with its own note, so a mistyped path behind a faulting one still fails loudly; a seed whose own content will not parse is reported on the seed's line ("Nothing to walk from") rather than as a node it reached, so nothing is counted as reached; and the record-fault filter is now a parse-family test (Mutagen's own exceptions plus the argument/format family its lazy reads raise, with cancellation and out-of-memory refused anywhere in the chain) instead of a catch-all, so an IO, disposed-session or internal fault propagates instead of being mislabelled. The `*parent` hop's fetch also got its own guard, so a fault reading the containing record is named as that record. Four new tests cover the reached count, the kept chain, the mistyped path behind a fault, and the seed-path case. Rebased again on main, 2024 tests pass, `ci-all` 128/128.


src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4075,7 +4078,11 @@ WalkNodeFact FactFor(FormKey k, bool atCap)
             // in hand here. Without this it re-read every chain node from disk — a whole-plugin seek each — after
             // the walk had already held that body and let it go.
             if (templateFollow && body is not null) templateFacts.TryAdd(k, FactOf(body));
-            if (body is not null && !atCap && !Excluded(fact.Type)) fact.Links = LinksOf(body, followSegs, out _);
+            // PER-RECORD FAULT ISOLATION, the twin of the scan lanes': reading a node's links parses its content
+            // lazily, so one record Mutagen cannot parse is recorded as a boundary and the walk goes on.
+            if (body is not null && !atCap && !Excluded(fact.Type))
+                try { fact.Links = LinksOf(body, followSegs, out _); }
```
Avick3110, line 4194: The guard covers `LinksOf`, but not the other lazy parse in the same function. On a `follow="Template"` walk, `FactOf(body)` runs one line above this try (line 4080), and `FactOf` reads `npc.Template` and `npc.Configuration.TemplateFlags` — both lazily parsed subrecords. The same two unguarded calls sit at line 4138 (`SeedTemplateFact = ... FactOf(seedBody)`) and at line 4330, where the results loop calls `NpcTemplateReport(TemplateFactOf, ...)` and `TemplateFactOf` → `FactOf` outside any try.

Scenario: a template walk (`follow="Template"`) that reaches an NPC_ whose record content Mutagen cannot parse. The throw comes from `FactOf`, not `LinksOf`, so it escapes `RecordsWalk` exactly as before this PR — the generic internal-failure wrapper, no record named. The issue's shape is fixed for closure and named-follow walks but not for the template lane the same function serves.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4075,7 +4078,11 @@ WalkNodeFact FactFor(FormKey k, bool atCap)
...
+                catch (Exception ex) { fact.Unscannable = $"{ex.GetType().Name}: {ex.Message}"; }
```
Avick3110: `catch (Exception ex)` here swallows every failure of the link read, not only a Mutagen parse fault, and reports all of them with the sentence "Mutagen could not parse its content". Two concrete cases:

1. `LinksOf` with a `*parent` path calls `Fetch(pk.Value)` for the containing record — a fault reading *that* record is attributed to this node, naming the wrong FormID in the boundary line.
2. An `OutOfMemoryException` on a large walk (the #719 shape this branch's own arm exercises) is caught, mislabelled as an unparseable record, and the walk keeps going — the next node then very likely throws again. A resource failure should end the call, not be recorded as a data fault.

The scan lane this is modelled on rethrows the exceptions that are not record faults (`catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` in the cross-query scan). Consider narrowing, or at least rethrowing the critical ones.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4137,22 +4144,36 @@ seedWin is null
...
+            // boundary row like any other unscannable node rather than raised out of the call.
+            List<FormKey> SeedLinks(string[]? segs, out string? note)
+            {
+                try { return LinksOf(seedBody, segs, out note); }
+                catch (Exception ex)
+                {
+                    note = null;
+                    st.Nodes.Add(new WalkNodeRow(FormIdToken.Of(seedFk), seedType, seedBody.EditorID, 0, st.Label, "kept",
+                                                 $"could not be scanned (Mutagen could not parse its content) and was not entered — {ex.GetType().Name}: {ex.Message}"));
+                    return new List<FormKey>();
```
Avick3110, line 4274: `SeedLinks` records the fault in `st.Nodes` but never writes it into `nodeFacts`, so the PR body's claim — "the memo that lets two seeds share one read remembers the fault the way it remembers a node's links, so the record is read once" — does not hold for a seed. Scenario: a multi-seed walk where seed A is itself unparseable and seed B's chain reaches A as a node. `FactFor(A, ...)` finds no memo entry, re-fetches the body, re-parses, throws again, and the record is read and thrown from twice. Same for a seed named twice in `formids=`. Correctness of the output is unaffected; the "read once" property is not.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4137,22 +4144,36 @@ seedWin is null
...
+
             // First hop: seed_paths (each path's links) or every link on the seed.
             if (seedPaths is { Count: > 0 })
             {
                 foreach (var p in seedPaths)
                 {
                     var segs = p.Trim().Split('.', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
                     if (segs.Length == 0) continue;
-                    var links = LinksOf(seedBody, segs, out var note);
+                    var links = SeedLinks(segs, out var note);
```
Avick3110, line 4285: `SeedLinks` is called once per entry in `seed_paths`, and it adds a boundary row on every throw. Scenario: `walk.seed_paths=["Perks","Race","Items"]` on a seed record Mutagen cannot parse — `ReadEngine.CollectLinksAt` throws for each of the three paths, so the seed's node list gets three identical "could not be scanned … and was not entered" rows for the same FormID. The node-side path records the fact once per record (it is memoised / short-circuited); the seed side should too — record the row on the first throw and reuse the empty result for the remaining paths. The row is also added without checking `maxNodes`, unlike every other `st.Nodes.Add` in the hop loop.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4270,6 +4291,14 @@ seedWin is null
                         st.Nodes.Add(new WalkNodeRow(FormIdToken.Of(key), type, fact.EditorId, hop, pulledBy, "kept", $"excluded ({type}, severity stop) — recorded as a boundary, not entered"));
                         continue;
                     }
+                    // A node Mutagen could not parse: named, kept as a boundary, and the walk continues — the same
+                    // answer the scan lanes give, never a raw exception out of the whole call.
+                    if (fact.Unscannable is { } unscannable)
```
Avick3110, line 4423: This branch `continue`s without setting `st.Truncation`, and it is checked *before* the `atCap` row below, so a node that is both at the depth cap and memoised-unscannable loses the depth-cap notice.

Scenario (multi-seed only, i.e. `nodeFacts` non-null): seed A reaches record K at hop 2 of a depth-4 walk, `LinksOf` throws, `Unscannable` is memoised. Seed B reaches the same K at hop 4 — `atCap` is true, so `FactFor` returns the memo untouched, this branch fires, and seed B's result carries no `walk reached its depth cap` truncation line even though that chain was in fact cut at the cap. Run the same seed B alone and it reports the depth cap instead, so the answer depends on how many seeds were in the call. Either set `st.Truncation ??= …` here when `atCap`, or order the cap check first.

src/housecarl-mcp-tests/RecordsWalkUnscannableTests.cs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// <summary>The reading forms consume the same reached set, so they complete too — the crash was in the walk,
+    /// not in the render.</summary>
+    [Fact]
+    public void AReadingFormOverTheSameWalkCompletesToo()
+    {
+        var r = RecordsTools.Records(
+            _w.Svc, formids: new[] { _w.SeedFid },
+            walk: new RecordsTools.RecordsWalk { depth = 4 },
+            project: new RecordsTools.RecordsProject { form = "summary" }, counts_only: true);
```
Avick3110: `counts_only: true` means the reading form never renders a record body, so this test does not exercise what its own summary claims ("the reading forms consume the same reached set… the crash was in the render"). Drop `counts_only` (or assert on the rendered rows) so a regression in the render path over an unparseable node actually fails here.

Related gap: neither test seeds the walk *on* the unparseable record, so the new `SeedLinks` wrapper — including the `seed_paths` branch — has no coverage at all. A third test with `formids: new[] { _w.BadPerkFid }` would cover it, and would also show the duplicate-row behaviour when several `seed_paths` are given.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. FactOf now carries the same per-record guard the link read does: it computes the type off the header, reads the template link, the template flags and the EditorID inside a try, and on a record fault returns a WalkTemplateFact carrying Unscannable instead of throwing. That covers all three call sites at once — FactFor (where the fault is promoted to the node's own Unscannable, so the node is a boundary row), the seed, and TemplateFactOf on the report path. NpcTemplateReport now tells that case apart from a broken chain: a template target that resolves but will not parse gets its own note rather than 'unresolved'. Covered by ATemplateWalkNamesAChainNodeThatWillNotParse, whose fixture is an NPC with its ACBS cut short, so the throw comes from the template flags and not from the link read.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in part. Cancellation and out-of-memory now go on up: every guard on this path is 'catch (Exception ex) when (IsWalkRecordFault(ex))', and IsWalkRecordFault excludes OperationCanceledException and OutOfMemoryException — a resource or cancellation failure ends the call rather than being recorded as a data fault and walked past. I did not narrow further: the scan lane this is modelled on catches Exception per record (ErrorCheck.Run's per-record guard is a bare catch), so anything tighter here would make the walk and the scan disagree about what an unscannable record is. On the *parent case: a fault fetching the containing record does land on the node whose links were asked for, which is the node the walk was reading links FOR — the same attribution LinksOf already makes when that fetch returns null ('the containing record X would not fetch'), and the note carries the exception text. Worth its own issue if that attribution should change; it would change both lanes, not just this guard.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The seed's fault is written into nodeFacts as a WalkNodeFact with Resolved=true and Unscannable set, so another seed reaching that record — or the same seed named twice — answers off the memo and never re-reads it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. SeedLinks now records the fault once per seed (the first throw sets seedFault and every later path short-circuits on it), the boundary row is added once after the first hop rather than inside the per-path loop, and it is held to maxNodes like every other row, setting the same truncation sentence when the budget is spent. Worth noting the scenario you describe does not actually reach SeedLinks's catch: on a NAMED path ReadEngine.CollectLinksAt handles the fault itself and returns it as that path's '(unreadable: ...)' note, one row per path asked for, which is pre-existing behaviour. The throw reaches SeedLinks on the closure seed (no seed_paths). Both are covered now — AnUnscannableSeedIsNamedAsItsOwnBoundary and AnUnscannableSeedUnderSeedPathsStillAnswers.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The unscannable branch now sets st.Truncation with the depth-cap sentence when atCap, so a memoised boundary reached at the cap still says the chain was cut there and the answer no longer depends on how many seeds were in the call.

src/housecarl-mcp-tests/RecordsWalkUnscannableTests.cs
Avick3110: Folded. counts_only is gone from that test and it now asserts on the rendered rows (the reached race and perk, and the unparseable record's FormID). Two tests seed ON the unparseable record: one with no seed_paths, asserting exactly one boundary row for the record, and one with repeated seed_paths, asserting the call still answers and the record is never named twice.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4200,22 +4241,49 @@ seedWin is null
...
+                    st.Nodes.Add(new WalkNodeRow(FormIdToken.Of(seedFk), seedType, st.EditorId, 0, st.Label, "kept",
+                                                 WalkUnscannableNote(fault)));
+                else
+                    st.Truncation ??= $"walk truncated: the {maxNodes}-node cap was reached — what is listed IS reached and proved; raise walk.max_nodes to walk further.";
+                if (nodeFacts is not null)
+                    nodeFacts[seedFk] = new WalkNodeFact
+                    {
+                        Resolved = true, Type = seedType, EditorId = st.EditorId, Unscannable = fault,
+                    };
+                st.Frontier.Clear();
```
Avick3110: A fault on a *later* seed path silently discards the chains an *earlier* one already produced. `SeedLinks` sets `seedFault` from whichever path throws, but the frontier is cleared after the whole `seedPaths` loop. Concrete case: `seed_paths=["Conditions","Effects"]` on the corrupt PERK — `Conditions` parses, its links are enqueued and edged, then `Effects` throws, and this `Clear()` throws away every chain the first path proved. The result names the seed as a boundary but says nothing about the chains that were dropped, and no `TruncationNote` is set, so the walk reads as complete when it is not. If the intent is that one unparseable path condemns the whole seed, the row should say the other paths' chains were dropped; otherwise the frontier gathered before the fault should be kept.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4200,22 +4241,49 @@ seedWin is null
             // The seed's facts go in the shared memo too, for the seed that sits on another seed's chain.
             if (st.SeedTemplateFact is { } sf) templateFacts.TryAdd(seedFk, sf);
 
+            // Reading the seed's own links parses its content, so a seed Mutagen cannot parse is recorded as a
+            // boundary rather than raised out of the call — ONCE for the record, however many seed paths ask for
+            // its links, the way the node side records a fault once per record.
+            List<FormKey> SeedLinks(string[]? segs, out string? note)
```
Avick3110, line 4265: Once `seedFault` is set, every remaining seed path short-circuits to an empty list with `note = null`, so the caller's `if (links.Count == 0 && note is not null)` never fires and those paths produce no row at all. A mistyped path spelled after the faulting one — `seed_paths=["Effects", "Efects"]` — loses the "(seed path 'Efects') no links … no such field" row that the comment on that line calls failing loudly; the user sees only the boundary note for the record and no sign their second path was never evaluated. Cheapest fix: keep answering each path's own note (for example return the fault as the note for the remaining paths, or skip the short-circuit and let each path record its own outcome).

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4200,22 +4241,49 @@ seedWin is null
...
             {
-                foreach (var l in LinksOf(seedBody, null, out _)) { st.Edge(seedFk, l); st.Frontier.Enqueue((l, 1, st.Label)); }
+                foreach (var l in SeedLinks(null, out _)) { st.Edge(seedFk, l); st.Frontier.Enqueue((l, 1, st.Label)); }
+            }
+            // The seed itself would not parse: one boundary row for the record, held to the same node budget every
+            // other row is, and the fault goes in the shared memo so another seed reaching it reads it no further.
+            if (seedFault is { } fault)
+            {
+                if (st.Nodes.Count < maxNodes)
+                    st.Nodes.Add(new WalkNodeRow(FormIdToken.Of(seedFk), seedType, st.EditorId, 0, st.Label, "kept",
```
Avick3110: This boundary row is the *seed itself*, at depth 0, but it goes into `st.Nodes`, which is what the chain header counts as reached: `RecordsTools.Records` computes `reached = rows.Sum(r => r.Nodes.Count(n => !n.Status.StartsWith("no links")))`. Walking with a single unparseable seed therefore prints `1 seed(s), 1 node(s) reached` when nothing was reached — everywhere else in this surface a seed is not a node (the consumption lane adds `r.Seed` separately from `r.Nodes`). Its `PulledBy` is also `st.Label`, i.e. the record says it was pulled by itself. Worth either excluding this row from the reached count (a distinct status would do it) or stating the seed's fault on the seed line rather than as a node.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -3878,14 +3878,30 @@ sealed class WalkNodeFact
...
+        /// and message, the same fact the scan lanes account as an unscannable record.</summary>
+        public string? Unscannable;
     }
 
     static readonly List<FormKey> EmptyKeys = new();
 
+    /// <summary>Whether a fault reading one record is the RECORD's — a body Mutagen cannot parse, which a walk
+    /// records as a boundary and steps over. A cancellation and an out-of-memory failure belong to the call, not to
+    /// the record, so they go on up rather than being recorded as a parse fault and walked past.</summary>
+    static bool IsWalkRecordFault(Exception ex)
```
Avick3110, line 3894: Everything that is not a cancellation or an OOM is now reported to the user as `could not be scanned (Mutagen could not parse its content)`, but the guarded region is wider than a content parse. `LinksOf` also resolves `*parent` hops (`view.ParentOf`, `Fetch(pk.Value)`) and calls `ContainmentIndex.SplitHops` / `ReadEngine.CollectLinksAt`, so an `IOException` from a plugin file that moved or a locked handle mid-walk, an `ObjectDisposedException` from the session, or an NRE in houseCARL's own read path all come back as a named record Mutagen "could not parse" — a wrong diagnosis the modder will take upstream to Mutagen, and the walk keeps going past a real failure instead of stopping. Narrowing to the parse family (MalformedDataException / SubrecordException plus a `RecordException` unwrap, say) keeps the intended isolation without mislabelling the rest.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The frontier is no longer cleared: chains an earlier seed path proved are kept and walked, and only the faulting path contributes nothing. That path says so on its own row, so the result shows both what was proved and what could not be read.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. SeedLinks no longer short-circuits: every path is evaluated and answers with its own note, so a mistyped path after a faulting one still gets its loud row. The record's fault is remembered once (seedFault ??=) for the memo and the seed's own line, which is what keeps the boundary from being repeated per path. Worth recording what the fixture shows: on a NAMED path the parse fault does not reach this catch at all — ReadEngine.CollectLinksAt handles it and returns it as that path's '(unreadable: …)' note — so each path was already answering for itself there; the catch is what the closure seed (no seed_paths) hits. Covered by AMistypedPathAfterAFaultingOneStillFailsLoudly and AFaultOnOneSeedPathKeepsWhatAnotherPathProved.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The seed's fault is no longer a node row. When the seed produced no chains at all it is the SEED's own error line — '<formid> could not be scanned (Mutagen could not parse its content) and was not entered — <exception>. Nothing to walk from.' — the same shape an unreadable winner body already gets, so nothing is counted as reached and nothing claims to have pulled itself. Under seed_paths each path has already said it on its own row, so no seed-level line is added there. AnUnscannableSeedIsNotCountedAsAReachedNode asserts '0 node(s) reached'.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. IsWalkRecordFault is now a parse-family test rather than a catch-all: it walks the exception chain, refuses it outright if a cancellation or an out-of-memory is anywhere in it, and otherwise accepts only exceptions from the Mutagen.Bethesda namespace (which covers RecordException and its wrapped cause, since the chain is read) plus the ArgumentException/FormatException/IndexOutOfRangeException/OverflowException/InvalidCastException family its lazy span reads raise. An IOException, an ObjectDisposedException or an NRE in our own read path now propagates and ends the call instead of being reported as a record Mutagen could not parse. I also narrowed the attribution you raised in round one: the *parent hop's Fetch has its own guard, so a fault reading the CONTAINING record is named as that record on the path's note rather than blamed on this node.

PR #749, opened 2026-09-14
Carry injected records through a merge, and fix the advice

### LINKED ISSUE #715: [Bug] merge's description recommends merging a mod with its own patches — bad advice, and that shape fails
The merge tool's description says "merging a mod WITH its patches is the intended use". That is bad advice, and the shape it recommends also happens to fail.

Real practice is the opposite: `Lux.esp` is left alone, and a user merges their 300 Lux patches into one mega patch that sits above it. Merging patches into the mod they patch is something people do only in narrow cases, and it is not the shape to point a modder at.

So there are two things here, and the first is the one that matters.

## 1. The description recommends the wrong shape

`housecarl_merge_plugins` should describe the normal job — merge a family of patches into one plugin, leave the base mod alone — and should not present merging a mod together with its own patches as the intended use. Anyone following the current sentence is being steered into worse load-order hygiene.

## 2. That shape also fails, with a bad error

Measured on a 3,801-plugin load order. 100 donors, Lux patches plus the three base plugins `Lux Via.esp`, `Lux Orbis.esp` and `Lux - Resources.esp`:

```
error: writing the merged plugin failed (serialize or commit; nothing partial left):
MissingModException: A referenced mod was not present on the load order being sorted against:
Lux Via.esp.  This mod was referenced by MajorRecord: 1134C9:Lux Via.esp.
```

It fails at the write, after ~55 seconds of work. Nothing partial is left.

The same 100 donors minus those three base plugins merged fine (7,694 originating records, 447 cross-donor conflicts, 1.8 MB output), as did the full 329-patch set (25,063 originating records, 3,548 conflicts, 10.7 MB output). The patches-only shape — the one people actually use — is healthy.

A reference survived pointing at `Lux Via.esp`, which was itself a donor, so by write time it named a plugin that is neither a master nor the output. `1134C9:Lux Via.esp` has the shape of an injected record, which would explain it: a record not enumerated among `Lux Via.esp`'s own originating records never enters the remap dict. That is a guess from the error, not a proven cause; the check is whether `1134C9:Lux Via.esp` is defined by `Lux Via.esp` or injected into it by something else.

Worth fixing whichever way it goes, because the same gap would bite any legitimate merge that includes an injected record.

## 3. The error shape is wrong regardless

A raw Mutagen exception type through the generic internal-failure wrapper, after a minute of work, telling the user nothing to do next. The pre-flight already walks every donor's records to build the remap and already knows the donor set, so a donor reference that cannot be remapped can be found there and refused in one sentence before any work starts.

Avick3110:
Three things on `housecarl_merge_plugins`, all from #715.

The tool description said that merging a mod together with its own patches is the intended use. It now leads with the normal job — merge a family of patches into one plugin and leave the mods they patch alone and active — and says the mod-plus-its-patches shape is a narrow case, because it moves the base mod's records to a new plugin identity and everything else that patches it then has to be merged or repointed. No skill text repeated the old advice (`.claude/skills/` carries no merge guidance at all), so the description is the only copy. The published description grew past the 2,048-character client cut, so the light/master-header paragraph was tightened to the claim plus a pointer at the report, which the bound test pins.

The injected-record gap: the remap is built from each donor's ORIGINATING keys, the records whose FormID master index names that donor. A record injected into donor A's space by donor B is originating to neither, so it never entered the dict, was copied into the merged plugin at its old identity, and the write then named a plugin that is neither a master nor the output. The donor scan now reads, in ONE enumeration per donor, the records the donor defines, the donor-space records it carries without defining, every record key it holds, and its links into donor space; an injected key is renumbered with the first donor carrying it, so the remap treats it like any other record that donor brings. It makes no claim about which plugin DEFINES the record — an override of an injected record declares the same master the injector does and can sort anywhere after it, so the load order cannot settle that. A plugin outside the merge that carries the record too is a dependent like any other: the key is in the dict, so the identify pass finds it and the external-overrider warning names it, which is the merge's standing posture.

The third part: a donor reference the remap cannot carry is refused from the same scan, before the identify pass and before anything is written, instead of surviving the renumber into the serialize. It asks only about the links that reach the output — a merge keeps the load-order winner's body for a record two donors both carry, so a stale reference a later donor already fixed does not refuse the merge — and the refusal offers adding the plugin that holds the target, since an active plugin outside the merge injecting it is the common cause. The scan follows `DeletedRecordRule` like the three sibling link walkers (that file's own list now names it as the fourth), so a deleted record's stale link neither throws on lazy parse nor refuses a merge.

Fixture arms, all of which fail on the parent branch: an injected record carried by a donor (the merged plugin declares no donor as a master and every record in it carries the output's ModKey), the same record carried by a plugin outside the merge (merges, and the warning names that plugin), a donor reference nothing holds (refused, no output folder left behind), that same reference already fixed by a later donor (merges), and that same reference on a DELETED record (merges).

The tools-list snapshot in `src/housecarl-mcp-tests/data` turned out not to need updating: it captures tool NAMES only, as the anchor test's own summary says, so a description change does not touch it. Called out because the brief expected otherwise.

`dotnet build`, `dotnet test --filter "tier!=bridge"` (2,024 passed) and `ci-all` (128/128) all green.

Stacked on #742 (#729); retarget once that merges.

Closes #715


src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2622,6 +2622,51 @@ public static CreatePluginOutcome CreatePlugin(string outPath, bool esl, string?
...
+            ov = SkyrimMod.CreateFromBinaryOverlay(srcPath, SkyrimRelease.SkyrimSE);
+            var originating = new List<FormKey>();
+            var carried = new List<FormKey>();
+            var links = new List<(FormKey, FormKey)>();
+            var seenLink = new HashSet<FormKey>();
+            foreach (var rec in ov.EnumerateMajorRecords())
+            {
+                if (rec.FormKey.ModKey == modKey) originating.Add(rec.FormKey);
+                else if (donorKeys.Contains(rec.FormKey.ModKey)) carried.Add(rec.FormKey);
+                foreach (var link in rec.EnumerateFormLinks())
```
Avick3110: This is a fourth walker over `EnumerateFormLinks`, and it is the only one that does not follow `DeletedRecordRule`. That file's doc names the three (`CrossQuery`'s references= arm, `ErrorCheck`'s dangling sweep, `RemapEngine.IdentifyExternalReferencers`) and says all of them "must treat a deleted record the same way", for two reasons that both bite here:

1. **Crash.** An engine-authored deleted record can leave a content-free-but-not-clean body behind (the doc cites deleted PACKs in a follower mod); Mutagen's lazy parse throws when the walk reaches for its links. The other three isolate that per record. Here the only `catch` is around the whole enumeration, so one such record in any donor turns the merge into `cannot parse 'X.esp' to renumber it (…)` and refuses the entire job. Before this PR the pre-flight was `TryReadOriginatingKeys`, which touched headers only and never reached a body — so a donor that merged fine yesterday now refuses.

2. **Semantics.** A deleted record links to nothing. Its stale link into donor space is now collected into `DonorLinks` and, if no donor defines the target, `MergeInjection.UnremappableLink` refuses the merge over a reference the game never reads. (`MergeBuild`'s existing post-remap check walks links without the guard too, but it walks the *Mutagen-authored* in-memory copy, which the doc says is clean by construction — this walk is over the source overlay, exactly the unsafe case.)

Both go away with the same two lines the other walkers use: `if (DeletedRecordRule.HasNoLiveBody(rec)) continue;` before the link walk, plus a per-record `try/catch` so one unscannable record is accounted rather than aborting the merge. Note the identity halves (`originating` / `carried`) read the header and must stay *above* the guard, as `IdentifyExternalReferencers` does with its overrider test.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7412,14 +7412,23 @@ public WritePatchBuilder.MergeOutcome MergePlugins(
...
+                if (!WritePatchBuilder.TryScanMergeDonor(dPath, dKey, donorModKeys, out var scan, out var keyErr))
                     return WritePatchBuilder.MergeOutcome.Fail(keyErr!);
-                donorKeys.Add((dName, keys));                             // a pure-override donor (0 originating keys) is a legit patch donor
-            }
+                scans.Add((dName, scan.Originating, scan.Carried));       // a pure-override donor (0 originating keys) is a legit patch donor
+                donorLinks.Add((dName, scan.DonorLinks));
+            }
+            // An injected record belongs to the donor that DEFINES it, which is the first plugin touching that FormID in
+            // the active order; one no donor defines is refused here rather than at the write (#715).
+            var injection = MergeInjection.Classify(scans, fk => view.TouchingPlugins(fk) is { Count: > 0 } t ? t[0] : null);
```
Avick3110: `definerOf` is "first plugin touching the FormID in the active order", which is not the same thing as "the plugin that defines it" for exactly the record class this code is about. An override of an injected record only needs the HOST plugin as a master, never the injector — so a plugin that merely overrides the injected record can legally load between the host and the real definer.

Concretely: order is `Host.esp`, `Override.esp`, `Injector.esp`; `Injector` defines `000901:Host.esp`, `Override` overrides it. Merge `[Host.esp, Injector.esp]`. `TouchingPlugins(000901:Host)[0]` is `Override.esp`, which is not a donor, so `Classify` refuses with `…is defined by 'Override.esp', which is not in the merge` — a false refusal naming the wrong plugin, and the suggested remedy (`add 'Override.esp' to plugins=`) is not the fix. The true definer is in the merge and the record was renumberable.

If first-toucher is deliberate (it is xEdit's own "master of the record" convention), the refusal wording is still wrong — it asserts "is defined by" about a plugin that may only be overriding. Either say "the first plugin in your order carrying it" in the message, or resolve the definer by preferring a toucher that does not declare the host-space owner... worth a decision either way rather than leaving the sentence making a claim the lookup can't support.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -100,13 +104,12 @@ public static string CompactPlugin(
...
...rs renumber ids already taken, and ANY donor's ids below the 0x800 floor renumber too (all records necessarily move to the new plugin's identity). Cross-donor conflicts on the SAME record resolve to the LOAD-ORDER WINNER and are each REPORTED; a losing donor's nested children the winner doesn't re-list (a base mod's dialogue lines under a patched topic; placed refs under a patched cell) are GRAFTED into the winner's copy, which is what lets a patch family merge cleanly even where one patch relists a topic another one fills. THE SAFETY (Q3): plugins OUTSIDE the merge that reference or override donor records are WARNED and NAMED, never refused — the donors stay active until you swap in MO2, so nothing breaks at write time; the remedy is to include those patches in the merge set or re-point them before disabling the donors. Refuses loud + writes nothing on: a donor unparseable / not on disk; a dangling donor-internal reference (a donor referencing a FormID no donor defines); a donor-space record INJECTED by a plugin outside the merge (add that plugin, or drop the donor it injects into); a declared master not active. Those are all found before any work starts, from the donors alone.")]
```
Avick3110: The closing claim — "Those are all found before any work starts, from the donors alone" — is not true of either new refusal, and this is a description a caller plans against:

- The injected-record refusal is decided by `view.TouchingPlugins(fk)` in `LoadOrderService.MergePlugins`, i.e. against the whole ACTIVE load order, not the donors alone. That matters to the reader: the same donor set can merge on one profile and refuse on another, and the refusal names a plugin the user never passed.
- The dangling-reference refusal (`MergeInjection.UnremappableLink`) runs *after* `ResolveOutputPath` has created the output mod folder and after `BuildMergeRemap`; it only reads as "before any work starts" because `FailAfterFolder` removes the folder again. "Nothing was written" is accurate; "before any work starts" is not.

Suggest dropping the last sentence or narrowing it to "all refused before anything is written".

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Folded in 82bbae94. TryScanMergeDonor now skips a deleted record's links (DeletedRecordRule.HasNoLiveBody) with the identity halves staying above the guard, as IdentifyExternalReferencers does, and wraps the per-record link walk so one unparseable record costs that record's links rather than the merge. The skipped links are not accounted in the report: the identity halves are header reads and so unaffected, and MergeBuild's post-remap walk over the clean in-memory copy still refuses a donor reference that survived the renumber — that is stated in the comment. A fixture arm now carries the dangling-arm's exact stale reference on a DELETED record and asserts the merge goes through. On the four walkers being one: agreed in principle, but unifying them is a change to three existing call sites with different accounting, so I left it out of this PR rather than widening it.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 82bbae94, by dropping the definer concept rather than refining it. You are right that first-toucher cannot support the sentence: an override of an injected record declares the same master the injector does and may sort anywhere after it, so the order does not settle who defines it. The rule is now simply that an injected key is renumbered with the first donor carrying it, and there is no injected-record refusal at all. A plugin outside the merge that carries the record too is then a dependent like any other: the key is in the dict, so the identify pass sees it and the external-overrider warning names it — which is the merge's standing posture (warn, do not refuse; nothing breaks until the user swaps). Your scenario now merges, and the test that asserted the refusal was replaced by one asserting the warning names the outsider.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 82bbae94. The injected-record half of that sentence is gone with the refusal itself; what is left reads 'Each of those is refused before anything is written', plus a sentence saying an injected record is merged like any other and a plugin outside the merge carrying it too is warned about by name.

src/housecarl-core/MergeInjection.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// throws a raw missing-mod fault (#715) — so it is asked here, before anything is built. Null when every donor
+    /// link is covered.</summary>
+    public static string? UnremappableLink(
+        IReadOnlyDictionary<FormKey, FormKey> dict,
+        IReadOnlyList<(string Donor, IReadOnlyList<(FormKey Source, FormKey Target)> Links)> donorLinks)
+    {
+        foreach (var (donor, links) in donorLinks)
+            foreach (var (source, target) in links)
+            {
+                if (dict.ContainsKey(target)) continue;
```
Avick3110, line 73: **Pre-flight refuses on dangling links the merge would have thrown away.**

The scan collects links from *every* record of *every* donor, but `MergeModsInto` walks winner-first and does **not** copy a cross-donor conflict loser's body (only its un-relisted children are grafted). So a dangling donor-space reference that exists only in a losing donor's copy of a record never reaches `M`, and `MergeBuild`'s post-remap check (WritePatchBuilder.cs step 2) never fired on it.

Concrete scenario, and it is the shape this PR now advertises as the normal job: base mod `A.esp` defines `X` with a stale `Template`/link pointing at `A.esp|0x9FF`, which `A` no longer defines; patch `B.esp` (later in load order) overrides `X` and fixes that reference. Merging `[A, B]` used to succeed — `B`'s corrected `X` is the winner and is what lands in the output. After this change the pre-flight sees the link in `A`'s discarded body and refuses the whole merge with "a FormID in donor 'A.esp's space that no donor holds", telling the user to fix a reference that the merge output would not have contained.

Worth restricting the pre-flight to the links that actually survive to the output (e.g. skip a source key that a later donor also carries), or keeping the pre-flight advisory and leaving the authoritative refusal to the post-build walk.

src/housecarl-core/MergeInjection.cs
```diff
@@ -0,0 +1,67 @@
...
+                return $"cannot merge: {FormIdToken.Of(source)} in donor '{donor}' references {FormIdToken.Of(target)}, a FormID in " +
+                       $"donor '{target.ModKey.FileName}'s space that no donor holds, so after the merge it would point at a plugin " +
```
Avick3110, line 75: **The refusal's remedy misses the common cause.** "a FormID in donor 'X's space that no donor holds" is true of the donor set, but the target very often *is* held — by an active plugin outside the merge that injected it into X's space. That is exactly the world this PR adds support for: `HcInjOutsider.esp` injects `HcInjSpace.esp|0x901`; merge `[Space, SomeDonorReferencingIt]` without `Outsider` and the user is told to "fix that reference in xEdit, or drop that donor", when the fix that actually works is to add the plugin carrying the target to the merge set (as the `Space + Overrider` arm shows). Suggest naming that remedy alongside the other two.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7412,14 +7412,23 @@ public WritePatchBuilder.MergeOutcome MergePlugins(
...
-                donorKeys.Add((dName, keys));                             // a pure-override donor (0 originating keys) is a legit patch donor
-            }
+                scans.Add((dName, scan.Originating, scan.Carried));       // a pure-override donor (0 originating keys) is a legit patch donor
+                donorLinks.Add((dName, scan.DonorLinks));
+            }
+            // An injected record — one whose FormID names a donor while another plugin carries it — is renumbered with
+            // the donor carrying it, instead of being copied at an identity the merge is about to remove (#715). A
+            // plugin outside the merge that carries it too is warned about by the identify pass below, like any other
+            // external overrider, because the key is now in the dict.
+            var donorKeys = MergeInjection.Renumberable(scans);
```
Avick3110, line 7718: **The PR description describes a different design from the one implemented here.** The body says "an injected record is given to the donor that DEFINES it — the first plugin touching that FormID in the active order" and "When the definer is not in the merge ... the merge refuses in one sentence naming both." Neither is in the code: `MergeInjection.Renumberable` gives the key to the first donor *carrying* it, its own XML doc says which plugin defines an injected record is not decidable from the order and makes no such claim, and `APluginOutsideTheMergeCarryingTheSameInjectedRecordIsWarnedAboutByName` pins warn-and-proceed rather than a refusal. The code's posture looks like the deliberate one; the description is stale and will mislead whoever reviews or merges on it (and the fixture-arm paragraph describes a refusal arm that does not exist). Please update the PR body.

src/housecarl-core/WritePatchBuilder.cs
```diff
@@ -2622,6 +2622,63 @@ public static CreatePluginOutcome CreatePlugin(string outPath, bool esl, string?
...
+            var seenLink = new HashSet<FormKey>();
+            foreach (var rec in ov.EnumerateMajorRecords())
+            {
+                // Identity first: it is read from the record HEADER, so it is safe on the records the link walk below
+                // cannot touch, and every record the merge carries has to be classified.
+                if (rec.FormKey.ModKey == modKey) originating.Add(rec.FormKey);
+                else if (donorKeys.Contains(rec.FormKey.ModKey)) carried.Add(rec.FormKey);
+                // A deleted record's links are not live, and reaching for them throws on the engine-authored bodies
+                // DeletedRecordRule describes — the rule every other link walker here follows.
+                if (DeletedRecordRule.HasNoLiveBody(rec)) continue;
```
Avick3110, line 2739: **`DeletedRecordRule`'s own registry is now stale.** Not re-raising the unification ask that was already answered on the thread above — this is a one-line consequence the fold left behind.

`DeletedRecordRule.cs` opens with "The ONE place the rule lives, so the scans that read a record's content can't drift apart on it", then enumerates its consumers: "Three walk Mutagen's `EnumerateFormLinks` over every record in an order, and all three must treat a deleted record the same way", followed by three named bullets (`CrossQuery`'s references= arm, `ErrorCheck`'s dangling sweep, `IdentifyExternalReferencers`). As of this line there are four, and the fourth is not named. The file also spells out the CONSEQUENCE clause per walker ("not returned by references=, not reported as dangling by check_errors, not listed as an external referencer") — the merge pre-flight's consequence, that a deleted record's stale donor-space link no longer refuses the merge, is the newest and least obvious of them and is documented only here in `WritePatchBuilder`, not where the rule lives.

The next person auditing "who follows this rule" reads the rule file, counts three, and does not find this walker. `DeletedRecordRule.cs` is not in the diff, so nothing flags it. A bullet and a consequence line there closes it.

src/housecarl-core/MergeInjection.cs
Avick3110: Folded in e1a8966e — real, and the fixture reproduced it exactly as described once the space plugin was in the donor set. The scan now also returns every record key a donor holds, and UnremappableLink resolves the same winner MergeModsInto does (last donor in load order holding the record) and skips the links of a losing body the merge discards. A new arm merges [Space, Dangler, Fixer], where Fixer overrides the dangling record cleanly: it refuses on the parent commit with this exact sentence and merges here. The per-donor link dedupe also moved from the target to the (source, target) pair, since which record a link came from is now part of the fact.

src/housecarl-core/MergeInjection.cs
Avick3110: Folded in e1a8966e. The refusal now offers three remedies, leading with yours: add the plugin that holds that record to plugins= (a plugin outside the merge may be injecting it), fix the reference in xEdit, or drop that donor.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded — the PR body is rewritten. The code's posture is the deliberate one; the body was left describing the pre-fold design. It now says the key goes to the first donor carrying it, that no claim is made about which plugin defines an injected record and why, that an outside carrier is warned about rather than refused, and it lists the five fixture arms that actually exist.

src/housecarl-core/WritePatchBuilder.cs
Avick3110: Folded in e1a8966e. DeletedRecordRule.cs now reads 'Four walk ... over records read from disk', names WritePatchBuilder.TryScanMergeDonor's merge pre-flight as the fourth (noting it walks the donors rather than the whole order), and the CONSEQUENCE clause gains its consequence: a deleted record's stale link does not refuse a merge whose donor it sits in.

PR #750, opened 2026-09-14
Read a [*count] column off the list, not through it

### LINKED ISSUE #724: [GAP] the fields form with a [*count] step over CELL bodies costs ~17 MB per row
Measured 2026-09-11 on a 3,801-plugin order, the installed 2.0.0 server in its own process.

The `fields` form with a `[*count]` step over CELL bodies costs about **17 MB per row**:

| call | rows | elapsed | peak WS | note |
|---|---:|---:|---:|---|
| `fields:["Temporary[*count]"]` over 60 CELL records | 60 | 7.4 s | **1.96 GB** (+1 GB) | |
| `where:["Temporary[*count] > 200"]` over every CELL, counts_only | 230 matches | 2.4 s | 1.2 GB | the predicate form of the same step is cheap |

Counting a cell's placed children under `fields` apparently materialises every child; the `where=` form of the same quantifier over every CELL in the order costs less than one sixth of that. The count is available from the containment index (2.7 M children indexed at startup) without reading a body at all.

Small on its own; filed because a user asking "which cells are dense" over a few hundred cells would hit several GB through the natural spelling.

Avick3110:
`project.fields=["Temporary[*count]"]` rendered a number and paid for the whole list to get it. Two costs. The container summary counted a collection by enumerating it, which builds every element of a Mutagen overlay list; and, on a field that holds CHILD RECORDS, the read assembled the additive union, which opens a body per plugin touching the record — a sentence the count cell then dropped on the floor, so nothing in the response carried it either.

Three changes:

- **The count comes off the collection's own count** (`ReadEngine.CountOf`, cached per runtime type, falling back to one enumeration where a type has no `Count`). A Mutagen overlay list knows its length from the record locations it parsed and builds an element only when one is indexed. The predicate side takes the same helper, so `where=["Temporary[*count] > 200"]` and the `fields` column are one number from one place.
- **A count column reads at depth 1** whatever depth the caller asked for: nothing under the list is rendered, so expanding it spends the shared expansion budget on lines the fold discards.
- **A field read only for a count takes the child union's index-only tier.** The count is the read body's own list, which is not what the game assembles, so the number must say so — it now carries `N other plugin(s) touch this record; their declarations for this child record field were not read`, and the response states that tier's clause over the count's own spelling. That tier costs one index lookup instead of a body per touching plugin. Naming the list without the token is what asks for the assembled union, and a count column beside a plain one does not cost the plain one its union: the tier is per field.

Closes #724.

## Measured

Private `housecarl-mcp` per arm over stdio (`HOUSECARL_DATA_DIR` at a scratch `houseCARL.user.json`), against the live MO2 instance `E:\Authoria - Requiem Reforged` — 3,245 plugin lines, 1 plugin excluded for a load failure. The index build is a separate warm call; the figure is the max working set inside the measured call's own window, sampled every 250 ms from outside. Two runs per arm where the arm moved.

The issue's own selection (60 CELL records with more than 200 placed children) is cheap on this order because those cells sit at override depth 1 — the union has one body to open. The arm that reproduces the issue's shape here is the same read over 60 **contested** dense cells (override depth 2–10, 325 provider bodies across the 60).

| arm | before | after |
|---|---|---|
| `fields:["Temporary[*count]"]`, 60 contested dense CELLs | 2.08 s / 1,185 MB · 2.23 s / 1,129 MB (idle 780 / 702 MB) | 0.19 s / 752 MB · 0.18 s / 806 MB (idle 759 / 800 MB — no rise) |
| `fields:["Temporary[*count]"]`, 60 CELLs at override depth 1 (the issue's own selection) | 0.82 s / 907 MB (idle 898 MB) | 0.82 s / 820 MB (idle 859 MB) |
| `where:["Temporary[*count] > 200"]` over every CELL, counts_only (control) | 4.26 s / 1,080 MB · 4.61 s / 1,008 MB | 3.28 s / 924 MB · 4.34 s / 983 MB |

The control moves within its own run-to-run spread on a machine running other lanes. The rendered numbers are identical: all 60 `Temporary[*count]` lines match between the before and after responses on both fields arms, and the where= scan reports the same 240 matches either side. The after responses additionally carry the index-only note on every count line and the clause once — neither of which the before responses stated at all.

## Tests

- `ACellsPlacedChildCountIsOneNumberUnderFieldsAndUnderWhere` — a cell with 4 placed children reads 4 under `project.fields`, matches `where=["Temporary[*count] = 4"]` and does not match `= 5`.
- `ACountColumnStatesTheIndexOnlyTierOfTheChildUnion` — the count line carries the index-only sentence and not the assembled union, the response states that tier's clause over `Temporary[*count]`, naming the list still gets the union, and a count column beside a named list does not cost the list its union.
- `AnOrdinaryListsCountIsOneNumberUnderFieldsAndUnderWhere` — the same agreement on a list that is not a containment collection (a form list's entries).
- `ACountColumnReadsTheSameNumberAtEveryDepth`.
- `EachTiersClauseNamesOnlyItsOwnFields` — a record annotated in both tiers at once states one clause per tier, each over its own fields.
- `AHoppedCountColumnTakesTheIndexOnlyTierToo` — a `*parent` count column takes the cheap tier too, matched on the column's whole path.

Full suite (`tier!=bridge`) 2,074 passed; `ci-all` 128/128, both after the rebase onto `origin/main`.


src/housecarl-mcp/FieldFold.cs
```diff
@@ -10,6 +10,12 @@ sealed record FieldFold(string Requested, string Root, string[] Tail, PathFold F
     /// bracketed index inside a step is another, because <c>ReadEngine.Expand</c> spends one level on each — so
     /// 'Conditions[0].Data' is three, not two, and counting segments alone stops the read a level short.</summary>
     internal int TailLevels => Tail.Sum(s => 1 + s.Count(c => c == '['));
+
+    /// <summary>The path the READ runs. A <c>[*count]</c> column reads the list under its own whole spelling: the
+    /// line it gets back is the same list line, and reading it there keeps the count clear of the annotations the
+    /// bare list path earns — the child-union note above all, which opens a body per touching plugin to say
+    /// something a number does not show.</summary>
+    internal string ReadPath => Fold == PathFold.Count ? Requested : Root;
```
Avick3110: Reading the count under its own spelling does not only drop a note that was already being discarded — it drops the owned-child clause from two lanes that DID state it.

`AnnotateOwnedChildContent` matches `owning.ContainsKey(below)` on the emitted line's path. With the read path now `Temporary[*count]`, nothing matches, so `OwnedChildFields` comes back null. Two consumers keyed off the fold's `Root` rather than the line path and therefore used to fire:

- `JsonWire.cs:1366` — `var owner = fold.Folds[c]?.Root; if (o.OwnedChildFields?.ContainsKey(owner) == true) childFields.Add(owner);` — the dense scan's response-level `owned_child_note`.
- `Artifacts.cs:164` — `foreach (var annotatedPath in af.Keys) annotated.Add(annotatedPath);` — the artifact manifest's clause.

On the scan lane that annotation cost nothing extra: with `memo` null, `AnnotateOwnedChildContent` takes the index-only tier (`view.TouchingPlugins(onKey)`) and opens no bodies. So the saving there is zero and the warning is gone.

Concrete: this PR's own test `ACountColumnDoesNotPayForTheChildUnion` asserts `Temporary[*count] = 0` on CellA, while the test right above it asserts the list line says `3 child record(s) across 1 plugin(s) … this body's own list carries 0`. A `records` scan with `fields:["Temporary[*count]"]`, or an artifact of one, now answers `0` for a cell the game fills with three placed references, with nothing anywhere in the response saying the number is the winner body's own list. Same for `where:["Temporary[*count] > 200"]`, which silently skips contested cells whose children are spread across plugins. That is the silently-degraded mode the cornerstone forbids.

If the union really must not be paid for on a count, the count cell should at least carry the cheap index-only sentence (`ReadSentences.NotReadNote(others)`), which needs only `view.TouchingPlugins`.

src/housecarl-mcp/FieldFold.cs
```diff
@@ -37,7 +43,9 @@ sealed record FoldPlan(IReadOnlyList<string> Requested, string[] Paths, FieldFol
         var depths = new List<int>(Paths.Length);
         for (int i = 0; i < Paths.Length; i++)
         {
-            int d = Folds[i] is { Fold: PathFold.Set } ? Depth : CallerDepth;
+            // A [*count] column renders the list's own count line and nothing under it, so it reads at depth 1
+            // whatever the caller asked for: expanding it would build every element for lines the fold discards.
+            int d = Folds[i] switch { { Fold: PathFold.Set } => Depth, { Fold: PathFold.Count } => 1, _ => CallerDepth };
```
Avick3110, line 42: Giving a count column its own read path makes the read emit a line whose path is `Temporary[*count]`, and both of the fold's path-matching helpers treat a `[…]` suffix as a step UNDER the list. Mixing a count column with an ordinary or `[*]` column on the same list now leaks a phantom line.

1. `fields:["Temporary","Temporary[*count]"]`, `depth:2`. `Read()` no longer de-duplicates (the paths differ), so `rec.Fields` holds `Temporary`, `Temporary[0]`… and `Temporary[*count]`. Column 0 has no fold and goes through `Lines(fields, "Temporary", CallerDepth)`, whose test is `RowProjection.IsUnder(f.Path, path) && Levels(f.Path, path) < depth`. `IsUnder("Temporary[*count]", "Temporary")` is true (`path[9] == '['`) and `Levels` is 1 < 2, so the plain column renders an extra row `Temporary[*count] = [list: N item(s)]` beside the real count cell.

2. `fields:["Temporary[*]","Temporary[*count]"]`. `RowProjection.RowKey("Temporary[*count]", ["Temporary"])` returns `Temporary[*count]` (it takes the first bracket segment after the root), so `Elements` yields it as an element row: one extra, bogus element in the `[*]` column carrying the list summary instead of an element. In the dense json lane `FoldPlan.ElementKey` keys that row `*count`, which also defeats the `keys.All(int.TryParse)` numeric sort, so the real element rows stop coming out in index order.

Neither case existed before, because both paths deduplicated to `Root`. A fix is to have `Lines`/`RowKey`/`Elements` skip lines whose trailing bracket segment is a quantifier token, or to keep the read path as `Root` and suppress the annotation some other way.

src/housecarl-core/ReadEngine.cs
```diff
@@ -265,6 +265,20 @@ public static RecordFields ReadFields(IMajorRecordGetter record, IReadOnlyList<s
         return new RecordFields(typeName, FormIdToken.Of(record.FormKey), record.EditorID, fields);
     }
 
+    /// <summary>The steps a read path NAVIGATES. A trailing <c>[*count]</c> is the fold's own step rather than a
+    /// field — the list's own line already carries how many elements it holds — so it is stripped here and the line
+    /// is still emitted under the caller's whole spelling. Reading the count under its own path is also what keeps a
+    /// count off the annotations the LIST line earns: a child-union note the count never renders would open a body
+    /// per touching plugin for nothing.</summary>
+    static string[] Segments(string path)
```
Avick3110: `Segments` strips the token for every `ReadFields` caller, but only the read lane's own record is folded afterwards — the conflict tree is not.

`LoadOrderService.cs:2751` builds each `ConflictNodeView` with `ReadEngine.ReadFields(n.Record, fields, ConflictDiffDepth, …)` over the same `fields` array, which is now `["Temporary[*count]"]`. `FoldPlan.Apply` only rewrites `o.Record`, so the tree nodes keep the raw lines. Result: `housecarl_records` with `conflict_tree` and `fields:["Temporary[*count]"]` renders per-plugin rows labelled `Temporary[*count] = [list: 4 item(s)]` — a path spelled as a count showing a summary, not a number. Before this change those rows read `Temporary = [list: 4 item(s)]`, which at least matched what they showed.

Same shape reaches `WriteFieldsArray` / any other consumer that matches a FieldValue path against an unfolded field list.

src/housecarl-core/FieldPredicate.cs
```diff
@@ -955,26 +959,45 @@ static int FirstFold(Fold[]? folds, int from, int len)
...
-        foreach (var e in (System.Collections.IEnumerable)val) if (e is not null) list.Add(e);
-        return (list, parent, null);
+        if (coll is null) return list;
+        foreach (var e in (System.Collections.IEnumerable)coll) if (e is not null) list.Add(e);
+        return list;
     }
 
+    /// <summary>How many elements a navigated collection holds — an absent one holds none. The same number the
+    /// project.fields column reads, from the same engine helper.</summary>
+    static int Count(object? coll)
```
Avick3110, line 1003: `Count` and `Materialise` no longer count the same things. `Materialise` drops nulls (`if (e is not null) list.Add(e)`); `CountOf` returns the collection's `Count` property, which includes them.

So on a list that carries a null element, `where:["X[*count] = 0"]` is false while `where:["X[*any] …"]` and `[*none]` fold over an empty element set, and `[*all]` is vacuously true. Two predicates on the same list now disagree about whether it is empty.

Aligning `[*count]` with the `project.fields` column (which never filtered nulls) is the right call, so the fix is probably the other direction: have `Materialise` keep nulls too, or at least state which of the two definitions a list's "elements" have. Low severity — depends on whether any Mutagen getter list actually yields nulls — but the two helpers sitting next to each other with different element sets is a trap.

src/housecarl-mcp/FieldFold.cs
Avick3110: Folded, and you are right that this was the real defect rather than a cost question. The count is the read body's own list, and the thing that says so was exactly what got dropped.

The read path is back to `Root`, so `AnnotateOwnedChildContent` matches again and `OwnedChildFields` is populated — the dense scan clause (JsonWire:1362) and the artifact manifest clause (Artifacts:164) fire as before. What is skipped is only the assembled union, and only for a field whose every column is a `[*count]`: `FoldPlan.CountOnlyPaths` names those, they are threaded to `ResolveRead` as `countFields`, and `AnnotateOwnedChildContent` leaves them out of `wanted` so that field takes the INDEX-ONLY tier — one `TouchingPlugins` lookup, no bodies. The count cell then carries the head line's `Display`, so the number renders as `Temporary[*count] = 207   (2 other plugin(s) touch this record; their declarations for this child record field were not read)`, and `FoldPlan.Apply` keys the annotation under the quantified spelling as well as the root so the response-level clause is earned on the text lane too (it was not stated there at all before this PR). A count column beside a named list does not cost the list its union — the tier is per field.

The scan lane is unchanged: `memo` is null there, it was already index-only, and nothing was saved or lost.

Test `ACountColumnStatesTheIndexOnlyTierOfTheChildUnion` pins all of it. I also widened the `NotReadFraming` remedy, which said "for the union, use formids=" — true from a scan, wrong on a formids= call that took this tier for a count; it now says to name the field without a quantifier.

src/housecarl-mcp/FieldFold.cs
Avick3110: Folded — by reverting the cause. `ReadPath` is gone and both columns deduplicate to `Root` again, so neither the `Lines`/`IsUnder` leak nor the `RowKey`/`ElementKey` phantom element row can arise. The annotation is suppressed through an explicit `countFields` set threaded to the read instead, which touches no path matching at all.

src/housecarl-core/ReadEngine.cs
Avick3110: Folded — `ReadEngine.Segments` is gone with the tokened read path, so `ReadFields` sees the same paths it always did and the conflict-tree nodes read `Temporary` again.

src/housecarl-core/FieldPredicate.cs
Avick3110: Folded as documentation rather than behaviour, and here is the reason. Aligning `[*count]` with the project.fields column is the call you name, so `Count` stays the collection's own count. Changing `Materialise` to keep nulls is the part I did not do: `FoldOver`'s per-element lambda calls `e.GetType()` and `EmitToken` on each element, so a null would have to be given a verdict class, and there is no fixture anywhere with a null-yielding getter list to pin what that class should be — I would be inventing a semantics blind. Both helpers now state their element set in their summaries, and each points at the other, so the difference is declared where someone reading either one will see it. If a real null-bearing list turns up, that is the fixture the fold semantics should be decided on.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2584,7 +2595,8 @@ static RecordFields AnnotateOwnedChildContent(RecordFields rf, IMajorRecordGette
             map ??= new Dictionary<string, ChildUnion?>(StringComparer.Ordinal);
             foreach (var (i, field) in hits)
             {
-                var u = unions?[field];
+                var u = unions is not null && unions.TryGetValue(field, out var found) ? found : null;
```
Avick3110: **The response-level clause can now state the UNION tier over a line annotated index-only.**

`ChildNotes.Emitted` ORs the tier flag (`_unioned |= unioned` in `ReadTools.cs`, and `childUnioned |= o.OwnedChildUnioned` in `JsonWire.cs`), and `OwnedChildUnioned` is per-outcome (`m.Values.Any(v => v is not null)`). Before this PR a single outcome's annotation map was uniform — the union lane unioned every annotated field — so the mixture could not arise. Making `wanted` exclude count-only fields makes it reachable for the first time.

Concrete case, which this PR's own test `ACountColumnStatesTheIndexOnlyTierOfTheChildUnion` already constructs but does not assert on:

```
project.fields = ["Temporary", "Persistent[*count]"]   on a CELL several plugins touch
```

`Temporary` gets `UnionNote`, `Persistent[*count]` gets `NotReadNote` ("…their declarations for this child record field were not read"). `_unioned` becomes true, so the one clause emitted is `UnionFraming`, naming *both* fields, and it asserts "the additive union beside it is the whole set the game assembles" — there is no union beside the count line. That is exactly what `OwnedChildClause`'s own doc forbids: "One door, so a lane cannot state a union clause over index-only notes or the reverse."

The tier is now per field, so the clause has to be too — either state both clauses over their own field lists, or exclude index-only fields from a union clause's field list.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2561,19 +2564,27 @@ static RecordFields AnnotateOwnedChildContent(RecordFields rf, IMajorRecordGette
...
             // Narrowed to the fields this read emitted: the union opens a body per touching plugin, and assembling a
             // worldspace's cells for a read that asked for EditorID would be a cost nobody asked for.
+            // A field read only for a [*count] takes the INDEX-ONLY tier: the count renders one number and no line
+            // under it, so the assembled union — a body per touching plugin — would be paid for a sentence that
+            // number does not carry. Both tiers state that the value is this body's own list; naming the list
+            // without the token is what asks for the assembled one.
             var wanted = new Dictionary<string, OwnedChildShape>(hits.Count, StringComparer.Ordinal);
-            foreach (var (_, field) in hits) wanted[field] = owning[field];
+            foreach (var (_, field) in hits)
+                if (countFields?.Contains(field) != true) wanted[field] = owning[field];
```
Avick3110: **`countFields` keys and `field` are two different spellings, so a hopped count column misses this test.**

`FoldPlan.CountOnlyPaths` yields `Paths[i]`, which `FieldFolds.Parse` sets to `folds[i].Root` — the WHOLE path including any `*parent.` hops (`string.Join(".", segs[..s].Append(bare))`, and `IsParentStep` segments are `continue`d past, not stripped). `field` here is the path BELOW the hops (`rf.Fields[i].Path.Split(...)[hops..]`).

So for `project.fields=["*parent.Temporary[*count]"]` on a placed reference, `countFields` holds `"*parent.Temporary"`, `field` is `"Temporary"`, the `Contains` misses, and the field stays in `wanted` — the read assembles the full additive union, opening a body per plugin touching the cell. That is precisely the cost this PR exists to remove, and it is the densest case (a worldspace cell). The count line then renders `UnionNote`, which also contradicts the sentence this PR added to `NotReadFraming` telling callers that "a '[*count]' column renders one number and takes this tier too".

The inverse is the dangerous half: if a child-bearing field name ever exists on both a record and its parent, a `[*count]` on the record's own field would silently downgrade the hopped, *unquantified* column of the same name to the index-only tier. Today's set (Cell / DialogTopic / Worldspace) has no such collision, but the matching is name-only and nothing enforces that.

Match on the same spelling on both sides — strip the hops from the `CountOnlyPaths` entries (or key them by hop count + below-hops path) rather than comparing a full read path against a below-hops one.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -2561,19 +2564,27 @@ static RecordFields AnnotateOwnedChildContent(RecordFields rf, IMajorRecordGette
...
-            if (memo is not null) unions = memo.Union(onKey, () => OwnedChildUnion.Compute(view, session, onKey, onSource, on, wanted));
-            else if (view.TouchingPlugins(onKey) is not { Count: > 1 }) continue;   // sole toucher: its own body IS the whole story
-            if (memo is not null && unions is null) continue;                       // same, on the union lane
-            var others = unions is null ? view.TouchingPlugins(onKey)!.Count - 1 : 0;
+            if (memo is not null && wanted.Count > 0)
+                unions = memo.Union(onKey, () => OwnedChildUnion.Compute(view, session, onKey, onSource, on, wanted));
+            var touchers = view.TouchingPlugins(onKey);
+            // Sole toucher: its own body IS the whole story, and the index-only tier has nothing to say about
+            // plugins that are not there. A union the lane did assemble still states its own negative.
+            if (unions is null && touchers is not { Count: > 1 }) continue;
```
Avick3110, line 2625: Minor, on the two guards here.

1. `if (unions is null && touchers is not { Count: > 1 }) continue;` — `OwnedChildUnion.Compute` returns non-null only when `touching.Count > 1`, so reaching the loop always implies `touchers.Count > 1`, hence `others > 0`. The `if (u is null && others == 0) continue;` inside the loop is therefore dead. Not wrong, but it reads as a live case and will mislead the next reader.

2. `unions?[field]` becoming `unions.TryGetValue(field, …)` turns a missing key from a loud `KeyNotFoundException` into a silent index-only note. That is the intended path for count-only fields, but it also swallows the one case that would genuinely be a bug: `ChildUnionMemo` keys purely on `FormKey`, never on the `wanted` set it was computed for, so a union memoised for a narrower `wanted` now downgrades a later, differently-spelled field instead of failing. Today every row in a batch shares one field list so it is not reachable, but the crash was the thing keeping it that way. Worth keying the memo on the wanted set, or asserting membership for fields that were in `wanted`.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The tier is per field, so the clause is now per tier everywhere: `ChildNotes` keeps one set per tier instead of a set plus an OR'd bool, the json lanes and the artifact manifest carry `field -> unioned` rather than a set plus a per-outcome flag, and `ReadSentences.OwnedChildClauses` states one clause per non-empty tier over that tier's own fields. `ClauseReserve` now takes how many clauses may still be said (1, or 2 where a record can annotate in both) so a two-clause response still answers inside max_chars.

Your exact case is the new test `EachTiersClauseNamesOnlyItsOwnFields`: `fields:["Temporary","Persistent[*count]"]` on CellA renders the union on `Temporary`, the index-only note on `Persistent[*count]`, and two clauses — the union one naming `CHILD RECORDS (Temporary)`, the index-only one naming `CHILD RECORDS (Persistent[*count])`. (Both framings open with the same sentence, so the test finds each by the half that is its own.)

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. The test now matches on the row's WHOLE read path — `rf.Fields[row].Path`, which is the spelling `CountOnlyPaths` is keyed by, hops included — instead of the below-hops field name. That fixes the miss you found and closes the inverse at the same time: nothing is compared by bare field name any more, so a name shared by a record and its parent cannot cross-match.

New test `AHoppedCountColumnTakesTheIndexOnlyTierToo` drives `project.fields=["*parent.Temporary[*count]"]` off the fixture's placed NPC and asserts the line carries the index-only sentence and not the union; it fails on the previous commit.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Both folded.

1. The dead `if (u is null && others == 0) continue;` is gone, and `others` is now plainly `touchers!.Count - 1` with the reason on the guard above it: `Compute` returns null below two touchers, so past that line every annotated field has at least one other plugin to name.
2. `TryGetValue` is gone. The lookup is now `unions is not null && wanted.ContainsKey(field) ? unions[field] : null` — a count-only field is null because it was never in `wanted`, and a field that WAS in `wanted` indexes the dictionary and throws exactly as before if the memo answered for a different field set. The loud failure you wanted kept is kept, without it swallowing the intended path.

PR #751, opened 2026-09-14
Read a DLL's file version and name the source of the version shown

### LINKED ISSUE #667: [Bug] housecarl_skse reports SPID 7.0.0 where the mod's meta.ini says 7.3.1.0

Avick3110:
`housecarl_skse` showed Spell Perk Item Distributor as 7.0.0 while the mod folder said 7.3.x, with nothing on the row to say where that number came from. Confirmed on the live order (MO2 instance `E:\Authoria - Requiem Reforged`, profile "Authoria - Requiem Reforged"): the DLL's `SKSEPlugin_Version` blob really does declare `7.0.0` — the name and author decode correctly from the same blob, so the offsets are right — while the DLL's Win32 version resource says `7.3.3.0` (PowerShell `(Get-Item ...).VersionInfo` agrees: FileVersion and ProductVersion are both 7.3.3.0) and the mod's `meta.ini` records `version=7.3.3.0`. So the read was not wrong; the report was silent about which of three numbers it was showing. This is not a SPID quirk: sweeping every DLL under `SKSE\Plugins` on that order, 51 of 313 declare a manifest version that disagrees with their own file version (EngineFixes 7.0.0 vs 7.0.19.0, JContainers 0.0.0.4 vs 4.2.9.0, Paraglider 1.0.0 vs 1.11.0.0, and so on).

The reader now also reads the image's version resource — the three-level `.rsrc` tree to `RT_VERSION`, then the `VS_VERSIONINFO` block's `VS_FIXEDFILEINFO`, whose signature is checked before any field is believed — riding the PE open the manifest read already pays for, and answering null rather than a guessed number when there is no resource or the block is not the struct it claims to be. Every rendered version is now labelled with its source, and the DLL's file version and the winning mod's `meta.ini` version are printed on the same line wherever they disagree with the declaration; versions that agree are not repeated. `format='json'` carries `file_version` and `mod_version` alongside `plugin_version`. Generic for every SKSE plugin, including the legacy and non-plugin DLLs that have no manifest at all — for those the file version is now the version the detail view shows.

What was run locally, driving this branch's Release build over stdio against the live order (the installed server's config points at a path that no longer exists, so a private `HOUSECARL_DATA_DIR` was used; the installed config was not touched):

```
po3_SpellPerkItemDistributor.dll  <- Spell Perk Item Distributor (loose)
  "Spell Perk Item Distributor" by powerofthree
  version 7.0.0 (SKSE manifest) - DLL file version 7.3.3.0
```

and in the whole-layer roster, `EngineFixes.dll  "EngineFixes" v7.0.0 (SKSE manifest) - DLL file version 7.0.19.0, meta.ini 7.0.19.0-AIO`. A sweep of all 313 DLLs compared the new read against Windows' own `FileVersionInfo`: 252 agree, 61 carry no version resource at either reader, 0 disagree.

Tests: a synthetic PE carrying a real resource tree and version block (read; no resource gives null; wrong signature gives null), plus renders pinning that a disagreeing row shows both numbers and names the source, that a meta.ini agreeing with neither is shown too, and that agreeing versions stay one number. `dotnet test` (2007 passed) and `ci-all` both pass.

Closes #667


src/housecarl-mcp/SkseTools.cs
```diff
@@ -830,6 +836,33 @@ internal static string Showing(int shown, int total, string noun = "", string hi
...
+        if (Differs(declared, mod) && Differs(file, mod)) others.Add($"meta.ini {mod}");
+        return others.Count == 0
+            ? $"{declared} (SKSE manifest)"
+            : $"{declared} (SKSE manifest) — {string.Join(", ", others)}";
+    }
+
+    /// <summary>Two version strings that are both present and NOT numerically equal ("7.0.0" vs "7.0.0.0" agree). A
+    /// missing one is not a disagreement — an unread version says nothing about the one that was read.</summary>
+    static bool Differs(string? a, string? b) =>
+        a is { Length: > 0 } && b is { Length: > 0 } && !SksePluginReader.VersionsEqual(a, b);
```
Avick3110: `Differs` calls `VersionsEqual`, which returns **not equal** for any segment that is not a plain integer. MO2 meta.ini versions routinely carry a non-numeric suffix, so every such mod is reported as a disagreement even when it is the same version. This PR's own measured output shows it: `EngineFixes.dll "EngineFixes" v7.0.0 (SKSE manifest) - DLL file version 7.0.19.0, meta.ini 7.0.19.0-AIO` — `7.0.19.0-AIO` and `7.0.19.0` are the same version, and the `meta.ini` clause is pure noise. Same for the common `5.2SE`, `1.0b`, `v1.2` forms; a mod whose meta.ini reads `2.1.0-beta` will always print a spurious third number. The changelog claim "agreeing versions are not repeated" does not hold for the meta.ini column.

A non-numeric segment is genuinely *unknown*, not *different* — consider normalising (strip a leading `v`, truncate at the first non-`[0-9.]` char) before comparing, or treating an unparseable pair as "no disagreement to report" the same way a missing one is.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -830,6 +836,33 @@ internal static string Showing(int shown, int total, string noun = "", string hi
...
+    /// file version and the mod's meta.ini version ride the same line wherever they differ, and an agreeing one stays
+    /// silent rather than tripling the width of every row.</summary>
+    internal static string VersionText(SksePluginReader.SksePluginInfo? p, string? modVersion)
+    {
+        string declared = p?.Version?.PluginVersion ?? "";
+        string file = p?.FileVersion ?? "";
+        string mod = modVersion ?? "";
+        // No manifest (a legacy or non-SKSE DLL): the file version IS the answer, labelled for what it is.
+        if (declared.Length == 0)
+            return file.Length == 0 ? "" : $"{file} (DLL file version)";
```
Avick3110: The `declared.Length == 0` early return drops `modVersion` entirely, so for a manifest-less DLL — `LegacyQuery`, `NotSkse`, `Unreadable`, exactly the kinds routed through the new detail line at SkseTools.cs:626 — the meta.ini version is never shown.

Two concrete misses:
- A legacy SE/VR plugin with **no** version resource but a meta.ini `version=2.1`: `VersionText` returns `""`, the `is { Length: > 0 }` guard at line 626 fails, and the detail view prints no version at all — even though houseCARL read one. This is the case where meta.ini is the *only* number available, so it is the one most worth printing.
- A non-SKSE bundled DLL whose file version is `1.0.0.0` while the shipping mod's meta.ini says `3.4`: the row shows `1.0.0.0 (DLL file version)` and silently hides the disagreement, which is the very thing the PR exists to stop.

The tool description added in this PR says the meta.ini version is printed "wherever they disagree" without excepting manifest-less DLLs, and the changelog says the feature is generic; the code is not.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -1735,7 +1771,7 @@ static string DllLine(NativePairedDll dll, string? runtime, bool withVersion)
         var sb = new StringBuilder();
         sb.Append(fate switch { DllFate.Dead => "[DEAD] ", DllFate.Verify => "[VERIFY] ", _ => "[LOADS] " })
           .Append(dll.Group.Length > 0 ? dll.Group + "\\" : "").Append(dll.FileName);
-        if (withVersion && dll.Info?.Version is { } v) sb.Append("  \"").Append(v.Name).Append("\" v").Append(v.PluginVersion);
+        if (withVersion && dll.Info?.Version is { } v) sb.Append("  \"").Append(v.Name).Append("\" v").Append(SkseInventoryWire.VersionText(dll.Info, null));
```
Avick3110, line 1799: The doc comment three lines up pins this row's format as `"[FATE] groupile ("name" vX)? — detail"`, with ` — ` as the single separator before the verdict. `VersionText` can now itself contain ` — ` (`7.0.0 (SKSE manifest) — DLL file version 7.3.3.0`), so a version-disagreeing DLL renders as:

```
[DEAD] po3_SpellPerkItemDistributor.dll  "Spell Perk Item Distributor" v7.0.0 (SKSE manifest) — DLL file version 7.3.3.0 — no class pairs to this DLL
```

The reason is no longer the text after the em dash, and on the 51-of-313 DLLs measured in the PR body this is the normal case, not the edge. Either use a different joiner inside `VersionText` (a comma, or parentheses) or pass only the declared version here — the native-pairing audit's job is the load verdict, not version provenance, and the JSON side already carries `file_version` separately.

src/housecarl-core/Mo2ModMeta.cs
```diff
@@ -61,6 +61,20 @@ public static class Mo2ModMeta
...
+    /// Stock Game / overwrite / hand-installed provider carries no meta.ini). The mod root is the file's path with its
+    /// Data-relative tail removed, which is where MO2 keeps meta.ini — no mods-folder scan, and no guess when the path
+    /// does not end in the tail it was resolved for.</summary>
+    public static string? VersionForLooseFile(string looseFilePath, string dataRelativePath)
+    {
+        var tail = dataRelativePath.Replace('/', Path.DirectorySeparatorChar);
+        var full = looseFilePath.Replace('/', Path.DirectorySeparatorChar);
+        if (!full.EndsWith(tail, StringComparison.OrdinalIgnoreCase)) return null;
+        var root = full[..(full.Length - tail.Length)].TrimEnd(Path.DirectorySeparatorChar);
+        if (root.Length == 0) return null;
```
Avick3110: `Read` does a full `File.ReadAllLines` + line scan of meta.ini, and this is called once per loose DLL with no memo. On the order this PR was measured against that is ~313 meta.ini reads per `housecarl_skse` call, and nothing caches across calls — a mod shipping several DLLs (EngineFixes AIO) re-reads the same file once per DLL, and the `ModMetaIni` it builds is discarded apart from one field.

A small `ConcurrentDictionary<string, string?>` keyed on the derived mod root (the same shape as `_resolvableMemo` in `SksePluginReader`) would collapse this to one read per mod.

Minor, separate: `root` is only rejected when empty, so a DLL resolved directly under a drive root leaves `root` as `C:` and `Path.Combine` yields the drive-relative `C:meta.ini`, resolved against the process's current directory on that drive rather than the drive root.

src/housecarl-core/SksePluginReader.cs
```diff
@@ -438,6 +448,87 @@ bool Walk(int dirRva, int dirSize, int stride, int nameOff, bool delay)
...
+    /// <c>VS_FIXEDFILEINFO</c>, whose 0xFEEF04BD signature is checked before any field is believed.</summary>
+    static string? ReadFileVersionResource(PEReader pe)
+    {
+        try
+        {
+            var dir = pe.PEHeaders.PEHeader!.ResourceTableDirectory;
+            if (dir.RelativeVirtualAddress == 0) return null;             // no resources at all — the common case for a lean DLL
+            var block = pe.GetSectionData(dir.RelativeVirtualAddress);
+            if (block.Length == 0) return null;
+            var res = block.GetContent();                                 // offsets inside the tree are relative to this base
```
Avick3110: `block.GetContent()` copies the **entire remainder of the .rsrc section** into a fresh `ImmutableArray<byte>`, and `ReadFileVersionResource` now runs on every DLL read — the whole `SKSE\Plugins` sweep (313 DLLs on the measured order), plus every BSA-packed DLL that goes through `ReadBytes`. The walk only ever touches the first three directory headers (a few dozen bytes at the start of the block); the version data itself is fetched separately by RVA at line 484, so the big copy buys nothing.

For a DLL that embeds a large payload in .rsrc (icon sets, bundled binaries) this is a multi-MB allocation per file, repeated for every call of the tool. A bounded `block.GetContent(0, Math.Min(block.Length, N))` covers the tree in practice, though it is not guaranteed by the format — if you take that route, treat a hop whose offset lands past the truncation as unreadable (null) rather than as a wrong answer, which the existing bounds checks already do.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded. Versions are now compared on their numeric prefix (leading 'v' dropped, trailing tag cut), and a pair with no numeric prefix has nothing to disagree about, so '7.0.19.0-AIO' vs '7.0.19.0' is no longer a disagreement. VersionsEqual itself is unchanged - it gates the runtime lock, where a non-numeric segment must stay 'not equal'. The live EngineFixes row now reads: EngineFixes.dll "EngineFixes" v7.0.0 (SKSE manifest; DLL file version 7.0.19.0). Test: ATaggedVersionAgreeingOnTheNumbersIsNotADisagreement.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded. With no manifest the text now falls back to whatever version was read - the file version, or the mod's meta.ini when that is the only number - and prints meta.ini beside the file version when the two disagree: '1.0.0.0 (DLL file version; meta.ini 3.4)'. Tests: AManifestLessDllFallsBackToTheModVersion, AManifestLessDllStillShowsAModVersionThatDisagrees.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded. The extra versions are now parenthesised instead of joined with an em dash: '7.0.0 (SKSE manifest; DLL file version 7.3.3.0)', so the fate line keeps one separator. Test: TheVersionTextDoesNotUseTheRowSeparator.

src/housecarl-core/Mo2ModMeta.cs
Avick3110: Folded. The root derivation moved to Mo2ModMeta.ModRootForLooseFile and the inventory loop memoises mod root to version for the call, so each mod's meta.ini is read once regardless of how many DLLs it ships. Per-call rather than a static memo, since meta.ini changes while the server runs. A bare drive root ('C:') is now refused, so no drive-relative 'C:meta.ini'.

src/housecarl-core/SksePluginReader.cs
Avick3110: Folded. The tree walk now copies a bounded head of the section (64 KiB) rather than the whole .rsrc remainder; the version data is still fetched by its own RVA. A hop landing past the bound fails the existing length checks and yields null, so a truncated tree is unknown, never a wrong number.

src/housecarl-mcp/SkseTools.cs
```diff
@@ -830,6 +836,56 @@ internal static string Showing(int shown, int total, string noun = "", string hi
...
+        // No manifest (a legacy, non-SKSE or unreadable DLL): whatever version WAS read is the answer, labelled for
+        // what it is — the file version, or meta.ini when that is the only number in sight.
+        if (declared.Length == 0)
+        {
+            if (file.Length == 0) return mod.Length == 0 ? "" : $"{mod} (mod meta.ini)";
+            return Differs(file, mod) ? $"{file} (DLL file version; meta.ini {mod})" : $"{file} (DLL file version)";
+        }
+        var others = new List<string>();
+        if (Differs(declared, file)) others.Add($"DLL file version {file}");
+        if (Differs(declared, mod) && Differs(file, mod)) others.Add($"meta.ini {mod}");
```
Avick3110: **A meta.ini disagreement is dropped whenever the DLL has no version resource.**

`others.Add($"meta.ini {mod}")` is gated on `Differs(file, mod)`, but `Differs` returns **false** when either side is empty. So when `file` is empty — the DLL carries no Win32 version resource, which by this PR's own sweep is 61 of 313 DLLs on the live order — the second clause is false and the meta.ini number is suppressed even though it disagrees with the manifest.

Concrete: a resource-less DLL whose SKSE manifest declares `7.0.0` while MO2's meta.ini records `7.3.3` renders `v7.0.0 (SKSE manifest)` with no hint that another number exists — exactly the silent stale-version the PR exists to surface. The intent of the clause is only "don't repeat meta.ini when it already equals the file version", so it should suppress only when `file` is present *and* equal:

```suggestion
        if (Differs(declared, mod) && !(file.Length > 0 && !Differs(file, mod))) others.Add($"meta.ini {mod}");
```

(Or equivalently: `if (Differs(declared, mod) && (file.Length == 0 || Differs(file, mod)))`.) Worth a test alongside `AModVersionAgreeingWithNeitherIsShownToo` with `fileVersion: null`.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -470,14 +471,24 @@ public SkseInventoryData SkseInventory(string? peekFilter = null)
...
                 string? note = null;
                 if (winner is { Kind: AssetKind.Loose, LooseFilePath: { } path })
+                {
                     info = SksePluginReader.Read(path);           // the winning loose copy
+                    // What MO2 recorded for the mod that ships it, read once per mod: a mod shipping several DLLs
+                    // (an AIO) would otherwise re-read the same meta.ini for each of them.
+                    if (Mo2ModMeta.ModRootForLooseFile(path, rel) is { } modRoot)
+                    {
+                        if (!modVersions.TryGetValue(modRoot, out modVersion))
+                            modVersions[modRoot] = modVersion = Mo2ModMeta.Read(Path.Combine(modRoot, "meta.ini"))?.Version;
```
Avick3110, line 487: **The mod's meta.ini version is attached to every DLL the mod ships, including bundled third-party dependencies — which turns an AIO into a row of false disagreements.**

`ModVersion` is the winning provider's folder version, so every DLL under that mod's `SKSE/Plugins` gets the same number. For a bundled dependency (`Kind == NotSkse`, so `declared` is empty) `VersionText` then takes the manifest-less branch and prints `{file} (DLL file version; meta.ini {mod})` whenever the two differ — which for a bundled DLL they essentially always do.

Concrete: Engine Fixes AIO (meta.ini `7.0.19.0-AIO`) ships `tbb.dll` alongside `EngineFixes.dll`. `tbb.dll` renders as e.g. `version 2020.3.0.0 (DLL file version; meta.ini 7.0.19.0)`, which reads as "this DLL is out of sync with what was installed" when nothing is wrong — the meta.ini number simply describes a different piece of software. Same for every SKSE-framework AIO and every mod that vendors a redistributable.

A cheap narrowing: only carry `ModVersion` onto entries whose `Kind` is `Modern`/`LegacyQuery` (a DLL the mod is plausibly *about*), or skip it for `NotSkse` in `VersionText`. Otherwise the third number is noise on exactly the DLLs it says least about.

src/housecarl-mcp/SkseTools.cs
Avick3110: Folded. meta.ini is now held back only when a file version is present and equal: `Differs(declared, mod) && (file.Length == 0 || Differs(file, mod))`. Added AModVersionDisagreeingWithTheManifestIsShownWithNoFileVersion. On the live order this immediately surfaces resource-less DLLs, e.g. ActorLimitFix.dll "Actor Limit Fix" v0.0.0.9 (SKSE manifest; meta.ini 9.0.0.0).

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. ModVersion is now carried only for a DLL whose kind is Modern or LegacyQuery, so a bundled redistributable never gets the shipping mod's number. On the live order the one non-plugin DLL (CrashLogger's msdia140.dll) now carries no mod version at all.

PR #752, opened 2026-09-14
Read and write plugin text as UTF-8 with a Windows-1252 fallback

### LINKED ISSUE #740: [Bug] UTF-8 translated names read as Windows-1252 mojibake and write back as ?: English-language tables and inline FULL fields carrying UTF-8
Reported by a Japanese user on 2.0.0 (Codex host), 2026-09-14.

`housecarl_records` returns garbled Japanese for translated names, in text and JSON output alike. `10FC28:Skyrim.esm` should read `エルフの防御術の盾(優)` and comes back as mojibake. The setup is the common Japanese-community one: `sLanguage=ENGLISH` with the `*_English.STRINGS` tables replaced by UTF-8 Japanese translations, and inline `FULL` fields in non-localized ESPs also carrying UTF-8. The source bytes are valid UTF-8.

## Cause

Every overlay is opened with Mutagen's default encoding (`LoadOrderResolver.OpenOverlay`, both the strings-redirected branch and the plain `CreateFromBinaryOverlay` beside it; the mutable imports and readbacks the same way). Mutagen picks the encoding from the target language, and English is Windows-1252, so UTF-8 bytes decode as 1252 mojibake on read. On write the 1252 encoder turns any character outside 1252 into `?`, so an in-place edit or a copy of such a record destroys the text silently. That second half is a silent-wrongness write, not a display nit.

## Fix

Mutagen 0.54.4 already carries the encoding the report asks for: `MutagenEncoding._utf8_1252` (strict UTF-8 with a Windows-1252 fallback), beside `_utf8_932` and siblings, and the read and write parameters expose `NonTranslatedEncodingOverride` and `NonLocalizedEncodingOverride` (`WithUtf8Encoding` helpers). Set both overrides to the UTF-8-with-1252-fallback encoding at the existing strings-correct choke points (overlay open, mutable import for in-place and copy, readback) and on the write parameters, leaving language selection and the localized-table language untouched. A Windows-1252 string that happens to be valid UTF-8 is the one theoretical miss; state it, do not guard it.

Tests, fixtures only (no MO2 instance): a UTF-8 inline `FULL` reads back as the same string; a UTF-8 `_English.STRINGS` entry reads back as the same string; a Windows-1252 plugin with accented characters reads exactly as before; an in-place edit and a copy of a UTF-8 name round-trip byte-identical; JSON and text projections carry the characters unescaped.

## Open question

Whether 2.0.0 regressed this. 1.9.0 shipped Mutagen 0.53.1 with the same language-driven default, so 1.9 should have shown the same mojibake unless Mutagen's Skyrim SE English default moved between 0.53.1 and 0.54.4. Settle it in the PR by running the 1.9.0 server against the UTF-8 `FULL` fixture and say which.

The reporter has a local patch on this approach that passed the full suite (1,933 tests) and the 128 probes.

Avick3110:
Mutagen picks a plugin's text encoding from the target language, and Skyrim SE + English is the one pairing whose answer (`MutagenEncoding._1252`) carries no UTF-8 lane at all. That is exactly the setup the Japanese community ships — `sLanguage=ENGLISH` with the `*_English.STRINGS` tables replaced by UTF-8 translations, and UTF-8 in the inline `FULL` fields of non-localized ESPs — so every translated name read as Windows-1252 mojibake, and a Windows-1252 *encoder* has no spelling for a character outside 1252, so an in-place edit or a copy of such a record wrote `?` over the text. The second half is a silent-wrongness write, not a display nit.

`src/housecarl-core/PluginTextEncoding.cs` is the one home for the encoding: strict UTF-8 first, and whatever Mutagen would have chosen for the target language as the fallback beneath it. It is applied at every place an overlay is opened, a mutable is imported (in place and copy), a readback is taken — 41 call sites in `housecarl-core`, `housecarl-mcp` has none — and on the write parameters at both `BeginWrite` choke points in `WriteEngine`. Language selection and the localized table's language are untouched: only the encoding a language's answer carries is widened. The one theoretical miss is stated in the file's own comment and not guarded — a Windows-1252 string whose bytes also happen to be valid UTF-8 reads as UTF-8, because the bytes are genuinely ambiguous and a guard would have to guess.

**Why not Mutagen's own `_utf8_1252`, which the issue named.** Its UTF-8 half is `Encoding.UTF8`, which replaces invalid bytes with U+FFFD rather than throwing, so the 1252 fallback under it is never reached: a real Windows-1252 name comes back as replacement characters. Measured — `Épée d'acier` read as `�p�e d'acier` through `_utf8_1252`. Mutagen's `MutagenEncodingFallbackWrapper` is used as-is; only the UTF-8 half is rebuilt with `throwOnInvalidBytes`, which is what makes the fallback fire. The 1252 no-regression arm in the tests is the one that pins this.

**Three lanes, not one.** The inline `FULL` lane is fixed by `NonTranslatedEncodingOverride`; the `.STRINGS` table lane is not — Mutagen reads a localized table with the encoding its provider returns for the target language, which the overrides never reach. So the read parameters also carry an `IMutagenEncodingProvider`. `NonLocalizedEncodingOverride` is deliberately left unset so the provider supplies it per language: pinning it would fix the fallback codepage to 1252 regardless of which language was selected, which is the one thing the issue asked to leave alone. The three lanes and the reasoning are in `PluginTextEncoding.cs`.

**Did 2.0.0 regress this? No.** No 1.9.0 build is on this machine — both installed server copies (`~/.claude/skills/housecarl/server` and `%LOCALAPPDATA%\houseCARL\server`) report 2.0.0, and `dist/` holds the 2.0.0 build — so the fixture was not run against one. The question is settled from source instead, which is stronger than a run: `git grep` over `v1.9.0` finds no `EncodingOverride`, `MutagenEncoding`, `WithUtf8Encoding` or `EncodingBundle` anywhere, so 1.9.0 took Mutagen's default; and Mutagen 0.53.1's `MutagenEncodingProvider.GetSkyrimSeEncoding` returns `_1252` for `Language.English`, character for character the same as 0.54.4's. 1.9.0 showed the identical mojibake. This is a bug of long standing, not a 2.0.0 regression.

**One thing left out, on purpose.** The issue's test list asked for the json projection to carry the characters unescaped. It does not, and the change that would have done it was reverted: `Utf8JsonWriter`'s default encoder escapes non-ASCII to `\uXXXX`, which is ASCII, so a json response's byte length and its character length coincide — and the check family's cap accounting measures UTF-8 bytes (`MemoryStream.Length`) against a cap the caller states in characters. Letting non-ASCII ride unescaped breaks that identity, and `check-guard`'s `MATRIX-REMEDY-CLEARS-IN-ONE-STEP` caught it immediately: the overrun notice reported its own response as 3325 chars when it was 3319. The cap invariants stayed safe (measured ≥ actual), but a notice that misstates its own length is a silently wrong answer, and fixing the byte/character split across that accounting is its own change, not this one. `\uXXXX` is json's own spelling for these characters and parses back to the identical string, so nothing is lost; the test asserts the parsed value. Worth its own issue.

Tests are fixtures only, no MO2 instance: `src/housecarl-mcp-tests/Utf8NameTests.cs` builds three plugins — a UTF-8 inline `FULL`, a localized plugin whose `_English.STRINGS` is genuinely UTF-8 (written through a `StringsWriter` with a UTF-8 provider, because the table encoding does not come from the embedded bundle), and a real Windows-1252 accented name. Six arms: each of the three reads back as itself, the json projection carries the same characters, an in-place edit leaves the untouched UTF-8 name intact in the file's bytes and on the next read, and a copy into a new patch carries it verbatim. Verified to bite: with the whole change stashed, four of the six fail; with only the write half removed, the two write arms fail — so neither half passes on the other's work.

Run: `dotnet build housecarl.sln -c Release` clean, `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` 2007/2007, `ci-all` 128/128.

Closes #740


src/housecarl-core/PluginTextEncoding.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// not valid UTF-8.</summary>
+    public static readonly IMutagenEncodingProvider Provider = new LanguageProvider();
+
+    /// <summary>The read parameters for a plain overlay open or mutable import.</summary>
+    public static readonly BinaryReadParameters Read =
+        BinaryReadParameters.Default with { StringsParam = Strings(null) };
+
+    /// <summary>The encodings a write embeds in the plugin — the same encoding as the read side, so a record read and
+    /// written straight back round-trips.</summary>
+    public static readonly EncodingBundle Write = new(Utf8Then1252, Utf8Then1252);
```
Avick3110: **The write side now re-encodes every Windows-1252 plugin to UTF-8, which is the same silent-wrongness in the other direction.**

`Utf8Then1252` is `MutagenEncodingFallbackWrapper(StrictUtf8, _1252)`, and on the ENCODE path the strict UTF-8 half never throws for a normal .NET string — so, as the doc comment says, "Encoding always emits UTF-8". That is applied to both halves of the bundle (`NonTranslated` and `NonLocalized`) at both `BeginWrite` choke points.

Concrete failure: a French/German mod whose inline `FULL` is genuinely Windows-1252 (`Épée d'acier` = `C9 70 E9 65`, or the extremely common curly apostrophe `0x92`). houseCARL reads it correctly through the 1252 fallback, then an in-place edit — which re-serializes the WHOLE plugin, not just the edited record — writes every one of those names back as UTF-8 bytes. The game with `sLanguage=ENGLISH` decodes 1252, so a name that rendered correctly before the edit now renders as `Ã‰pÃ©e d'acier`. Before this PR that round-trip was byte-identical. The copy lane has the same shape: a record copied from a 1252 source into a new patch lands as UTF-8 in the patch.

`NonTranslated` makes it worse than a display nit: that half covers plain string subrecords including model/texture paths. An accented asset path (routine in French mods) is rewritten to UTF-8 bytes, and the game then looks for a path that does not exist — a missing mesh where one loaded before.

The read side is genuinely ambiguity-tolerant (strict UTF-8, fall back to 1252). The write side does not have to guess in the same way: an encoder can be 1252-first and fall back to UTF-8 only for characters 1252 has no spelling for, which keeps Western text byte-identical AND still writes Japanese correctly. That needs a 1252 encoder that THROWS rather than substituting `?`:

```suggestion
    /// <summary>Windows-1252 that THROWS on an unmappable character, so the UTF-8 fallback under it actually fires
    /// instead of silently writing `?`.</summary>
    static readonly IMutagenEncoding Strict1252 = new MutagenEncodingWrapper(
        Encoding.GetEncoding(1252, EncoderFallback.ExceptionFallback, DecoderFallback.ReplacementFallback));

    /// <summary>The encodings a write embeds. 1252 first so Western text round-trips byte-identical, UTF-8 for any
    /// character 1252 cannot spell — so a Japanese name is never `?`, and a French name is never re-encoded.</summary>
    public static readonly EncodingBundle Write =
        new(new MutagenEncodingFallbackWrapper(Strict1252, StrictUtf8),
            new MutagenEncodingFallbackWrapper(Strict1252, StrictUtf8));
```
(`CodePagesEncodingProvider.Instance.GetEncoding(1252, ...)` if the provider is not registered.) Whatever the chosen shape, the current one converts the encoding of text it was never asked to touch.

src/housecarl-mcp-tests/Utf8NameTests.cs
```diff
@@ -0,0 +1,240 @@
...
+        Assert.Contains(LocalizedName, r);
+    }
+
+    /// <summary>The no-regression arm: a real Windows-1252 name still reads as itself, which is what the strict
+    /// UTF-8 half buys — a lenient UTF-8 decoder would hand back replacement characters instead of falling back.</summary>
+    [Fact]
+    public void AWindows1252NameStillReadsAsItself()
+    {
+        var r = ReadName(_latinWeapon);
+        Assert.DoesNotContain("error:", r);
```
Avick3110, line 229: **The 1252 no-regression arm only covers reads, so it cannot catch the write-side re-encode.**

`AWindows1252NameStillReadsAsItself` asserts through `ReadName`, which goes through the new UTF-8-first reader. That assertion passes whether the on-disk bytes are 1252 or UTF-8 — the reader accepts both — so nothing in this file pins that a 1252 plugin's bytes SURVIVE a write. The two write arms both use the Japanese plugin, where any output is an improvement.

With the current `Write` bundle, an in-place edit of `HcUtf8Latin.esp` rewrites `Épée d'acier` from `C9 70 E9 65` to UTF-8 (see the comment on `PluginTextEncoding.cs`), and every test here still goes green.

The missing arm is the mirror of `AnInPlaceEditLeavesAUtf8NameIntact`: apply an in-place edit to `LatinPluginName`, then assert the file still holds the 1252 bytes — e.g. a `FileHoldsBytes(path, Encoding.Latin1.GetBytes(LatinName))` next to the existing `FileHolds`, plus `Assert.False(FileHolds(path, LatinName))` so a UTF-8 rewrite fails loudly. As written the suite would ratify the regression.

src/housecarl-core/PluginTextEncoding.cs
```diff
@@ -0,0 +1,76 @@
...
+/// <para><b>The one theoretical miss.</b> A Windows-1252 string whose bytes also happen to be valid UTF-8 decodes as
+/// UTF-8 and reads differently than it did. Nothing guards it: the bytes are genuinely ambiguous, and any guard
+/// would have to guess.</para>
+/// </summary>
+public static class PluginTextEncoding
+{
+    /// <summary>UTF-8 that THROWS on invalid bytes — the strictness is the whole mechanism, because the fallback
+    /// below only fires on an exception.</summary>
+    static readonly IMutagenEncoding StrictUtf8 =
+        new MutagenEncodingWrapper(new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true));
```
Avick3110: Minor, but worth knowing before it shows up as a slow sweep: the fallback is driven by a thrown exception. `MutagenEncodingFallbackWrapper.GetString` is `try { primary } catch { secondary }`, so with strict UTF-8 as the primary EVERY non-ASCII Windows-1252 string in the load order costs a `DecoderFallbackException` throw + catch (roughly 5–20 µs each), not a branch.

For a Western modlist that is not a rare path — a full-order sweep or a bulk `records` read touching hundreds of thousands of `FULL` fields in French/German/Spanish plugins (or anything using the 1252 curly apostrophe `0x92`) pays it once per string, which can add seconds to a job that was previously decode-only. Nothing is wrong; just that "strict UTF-8 first" makes the non-UTF-8 corpus the expensive case rather than the cheap one. A cheap pre-check (scan the span for a byte ≥ 0x80 and, if present, validate UTF-8 with `Utf8.IsValid` before deciding) would keep the same answer without the throw, if this ever measures.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: Folded — this was the real one, and the worst of the three. Fixed in `17fbcdb2`.

You are right that the two sides are not the same problem. Reading, the bytes are ambiguous and UTF-8 has to go first; writing, the value is already a string and nothing needs guessing. `Write` is now `MutagenEncodingFallbackWrapper(Strict1252, StrictUtf8)` on both halves, with the 1252 encoder built `EncoderFallback.ExceptionFallback` — via `CodePagesEncodingProvider.Instance.GetEncoding(1252, ...)`, since the provider is not globally registered — so it throws instead of substituting `?` and the UTF-8 fallback beneath it actually fires. A plugin that was already 1252 now comes back byte-identical from an in-place edit, accented asset paths in `NonTranslated` included, and a Japanese name is still written as itself.

The mirror miss you named — a Latin-accented string that came out of a UTF-8 file being written back as 1252 — is now stated beside the read-side one in the same paragraph of the class doc, unguarded, with the asymmetry spelled out in its own paragraph above it.

src/housecarl-mcp-tests/Utf8NameTests.cs
Avick3110: Folded, and the diagnosis was exact: with the old UTF-8-first bundle all seven arms went green, so the suite would have ratified the regression.

`AnInPlaceEditLeavesAWindows1252NameAsWindows1252Bytes` (`17fbcdb2`) is the mirror of the Japanese arm — an in-place edit of `HcUtf8Latin.esp`, then `FileHoldsBytes(path, Cp1252.GetBytes(LatinName))` true and `FileHolds(path, LatinName)` (the UTF-8 spelling) false, so a re-encode fails loudly rather than passing quietly. `FileHolds` is now a thin wrapper over a `FileHoldsBytes` that takes raw bytes, and the Latin fixture gained a `BasicStats` so there is something to edit on it. The Japanese in-place arm also gained `Assert.False(FileHoldsBytes(path, ...))` on the question-mark spelling, so both directions are pinned by bytes rather than by the lenient reader.

Verified to bite: putting the UTF-8-first bundle back fails exactly this new arm and nothing else. `7e38462e` makes the two read arms point at each other so the next reader sees which one covers bytes and which covers decoding.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: Folded, and measured — it cost more than "minor", so it is fixed rather than noted.

The read side no longer uses Mutagen's `MutagenEncodingFallbackWrapper`. A small `Utf8First` decides with `Utf8.IsValid(bytes)` and decodes as UTF-8 or as the language's fallback: same answer, no throw. The write side still uses Mutagen's wrapper unchanged, because there a throw happens only for a string Windows-1252 cannot spell, which is the rare case by construction.

A 20,000-weapon plugin (2.3 MB) whose every `FULL` is Windows-1252 and not valid UTF-8, swept end to end through `CreateFromBinaryOverlay` touching every `Name`. This box is shared with other lanes measuring in parallel, so the absolute numbers are noisy; the ordering held on every run.

| read encoding | sweep |
|---|---|
| Mutagen default (1252, no fallback) | 83 / 87 / 94 / 102 ms |
| throw-driven `FallbackWrapper(strict utf8, 1252)` | 272 / 296 / 383 / 525 ms |
| branch-driven `Utf8.IsValid` then 1252 (shipped) | 23 / 21 / 21 / 150 ms |

So the throw-driven shape ran 3–5x the unfixed read on a Western order — seconds on a full-order sweep, as predicted — and the branch-driven one lands at or under the unfixed read. The 150 ms sample is another lane's arm landing on the same machine.

The benchmark lives in this session's scratchpad rather than the repo: it needs a 2.3 MB generated fixture and it measures a Mutagen behaviour, not a houseCARL one.

src/housecarl-core/PluginTextEncoding.cs
```diff
@@ -0,0 +1,113 @@
...
+    public static readonly BinaryReadParameters Read =
+        BinaryReadParameters.Default with { StringsParam = Strings(null) };
+
+    /// <summary>The encodings a write embeds in the plugin: Windows-1252 first, so a plugin that was already 1252
+    /// comes back byte-identical, and UTF-8 only for a string 1252 cannot spell, so a Japanese name is written as
+    /// itself instead of as <c>?</c>. 1252 is Mutagen's own write default and what the game reads at the language
+    /// houseCARL targets.</summary>
+    public static readonly EncodingBundle Write = new(
+        new MutagenEncodingFallbackWrapper(Strict1252, StrictUtf8),
+        new MutagenEncodingFallbackWrapper(Strict1252, StrictUtf8));
```
Avick3110: **The 1252-first write regresses a round trip that used to be byte-exact: a UTF-8 plugin's Latin-accented strings.**

Before this PR the read and the write were both Windows-1252, so a UTF-8 plugin whose `FULL` held `é` (bytes `C3 A9`) decoded to the two chars `Ã` + `©` and re-encoded to the same `C3 A9`. Mojibake on screen, but the file came back byte-identical from an in-place edit.

After this PR the read is UTF-8-first, so those bytes correctly decode to `é` — and then the 1252-first write emits a single `E9`. The file has changed, and a UTF-8 consumer (xTranslator, or the game at a UTF-8-reading language) now hits an invalid byte where a valid character used to be. The concrete case is not exotic: a French/German/Spanish/Polish UTF-8 translation ESP, where *every* accented character flips, after any `apply ... in_place=` or a `merge_plugins`/`compact_plugin` re-serialize.

Worse, on a Japanese ESP that also carries one accented Latin name or asset path, the rewritten file is now **mixed**: CJK as UTF-8, Latin-1 as 1252. No single-encoding reader gets both lanes right.

The fixture cannot see this — the UTF-8 plugin holds only CJK and the 1252 plugin only Latin, so no arm exercises an accented character that *came out of* a UTF-8 file. A fourth fixture plugin (UTF-8 bytes, accented Latin name) edited in place would fail today.

The symmetric fix is per-file: the reader already knows which lane it used, so a plugin any of whose strings decoded through the UTF-8 lane could be written UTF-8-first, and everything else 1252-first. That keeps the Western-plugin argument in the comment above intact (a genuinely 1252 plugin never decodes through the UTF-8 lane) without breaking the UTF-8 plugins this PR exists to serve. At minimum the changelog's "written back as Windows-1252" line should say that this is a *change of bytes that previously survived*, not merely a miss.

src/housecarl-core/PluginTextEncoding.cs
```diff
@@ -0,0 +1,113 @@
...
+    static readonly Encoding StrictUtf8Encoding =
+        new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
+
+    static readonly IMutagenEncoding StrictUtf8 = new MutagenEncodingWrapper(StrictUtf8Encoding);
+
+    /// <summary>Windows-1252 that THROWS on a character it cannot spell rather than substituting <c>?</c> — the
+    /// throw is what lets the UTF-8 fallback beneath it fire.</summary>
+    static readonly IMutagenEncoding Strict1252 = new MutagenEncodingWrapper(
+        CodePagesEncodingProvider.Instance.GetEncoding(
+            1252, EncoderFallback.ExceptionFallback, DecoderFallback.ReplacementFallback)!);
```
Avick3110, line 86: **The write side pays exactly the per-string throw the read side was rewritten to avoid.**

`Utf8First` (line 96) explains that Mutagen's `MutagenEncodingFallbackWrapper` is a try/catch, and that a strict primary would make "every non-ASCII Windows-1252 string in a sweep pay a throw" — so the read side decides with `Utf8.IsValid` instead. The write bundle then does the very thing that comment rejects, in the other direction: `Strict1252` is built with `EncoderFallback.ExceptionFallback`, wrapped in that same try/catch wrapper, so every character 1252 cannot spell costs a thrown `EncoderFallbackException`.

And it is two throws per string, not one: the wrapper falls back on `GetByteCount` and again on `GetBytes`. Re-serializing a fully-Japanese plugin — an in-place edit, or `merge_plugins`/`compact_plugin` over a translated modlist, which rewrites every record — means one throw per name, per description, per topic text, twice over. On a large translated ESP that is six figures of exceptions inside the serialize.

The same branch-not-throw trick applies here: ask whether the string is 1252-spellable (a scan for any char the codepage maps to `?`, or a `1252.GetByteCount` against an `EncoderReplacementFallback` sentinel) and pick the encoder, rather than throwing to find out.

src/housecarl-core/PluginTextEncoding.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        readonly IMutagenEncoding _fallback;
+        internal Utf8First(IMutagenEncoding fallback) => _fallback = fallback;
+
+        public string GetString(ReadOnlySpan<byte> bytes)
+            => Utf8.IsValid(bytes) ? StrictUtf8Encoding.GetString(bytes) : _fallback.GetString(bytes);
+
+        public int GetByteCount(ReadOnlySpan<char> str) => StrictUtf8Encoding.GetByteCount(str);
+
+        public int GetBytes(ReadOnlySpan<char> chars, Span<byte> bytes) => StrictUtf8Encoding.GetBytes(chars, bytes);
```
Avick3110: **`Utf8First`'s encode direction silently contradicts the write policy, guarded only by a comment.**

The class comment says "a read encoding never encodes" — but the type implements the encode half anyway, as plain UTF-8, and `Utf8Then1252` is `public` and is installed as `NonTranslatedEncodingOverride` on the read parameters. If any path ever encodes through it (a future reuse of the public field, or a Mutagen path that consults the read parameters' encoding when sizing or emitting a plain string subrecord), plain string subrecords go out UTF-8-first while `Write` says 1252-first, and the two halves of the "one home" disagree — silently, since UTF-8 output is always valid and nothing fails.

Nothing tests this, and the file's own premise is that the read and write sides must not be able to disagree. Make the contradiction loud instead of latent: `GetByteCount`/`GetBytes` here should `throw new NotSupportedException("read encoding; writes use PluginTextEncoding.Write")`. If the assumption holds nothing changes; if it is ever wrong, it fails by name rather than by writing the wrong bytes.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1857,7 +1858,7 @@ public static LocalizedFlagRead PluginIsLocalized(string path)
         ISkyrimModGetter? ov = null;
         try
         {
-            ov = SkyrimMod.CreateFromBinaryOverlay(path, SkyrimRelease.SkyrimSE);
+            ov = SkyrimMod.CreateFromBinaryOverlay(path, SkyrimRelease.SkyrimSE, PluginTextEncoding.Read);
```
Avick3110: The doc above this method (line 1848) still says it reads "deliberately with the BARE overlay", which is now false — it takes `PluginTextEncoding.Read`. Harmless for the header flag this actually reads, but the comment is the stated reason the call is shaped this way, so it now misleads the next reader into thinking the parameters were left off on purpose here.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: Folded, and you are right about more than the regression: the per-string rule was wrong in principle. A file has one encoding. Rewritten in `a048f298`.

The read now records, per plugin, which lane its own bytes resolved to — UTF-8 if any non-ASCII string in it was valid UTF-8, legacy if non-ASCII was read and none of it was, ASCII-only otherwise (a pure-ASCII string says nothing and records nothing). The write then uses ONE encoding for the whole file:

- **in place** — the lane the read of that same file resolved to, so a UTF-8 translation ESP comes back UTF-8 and a genuinely 1252 plugin comes back 1252, both byte-identical where the write did not change text;
- **a new file** (patch, merge, compacted copy) — UTF-8 if any plugin that contributed a record to it resolved as UTF-8, else the language default.

One correction to the obvious shape of that second rule: the declared master list is not enough, because Mutagen computes it *during* the write, so before the write it can still be empty — the patch arm failed on exactly that. The answer that always holds is the plugin each record came from, which is its FormKey's ModKey; the file's own name (the extend lane) and the master list are checked first because they are cheaper. The bound is stated in the comment: a lane that remaps records into the output's own ModKey — a merge — has no provenance left in the FormKeys and answers from the donors' master lists and the output name alone.

Two misses, both stated in the class comment and unguarded: reading, a 1252 string whose bytes are also valid UTF-8; writing, a value typed by hand into a patch whose contributing plugins were all ASCII-only, which still goes out in the language default. That second one is a `?` written silently, which I am not comfortable with — I have followed the steer to state rather than guard it, and flagged to Aaron that a loud refusal may be the better end state.

New fixture arms: a fourth plugin (UTF-8 bytes, accented Latin only) round-trips byte-identical through an in-place edit; the Japanese plugin now also carries an accented Latin name and both stay UTF-8 after an in-place edit, with the 1252 spelling asserted absent; the 1252 plugin still comes back 1252. Nine arms, all green; the changelog says the rule and both misses.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: Folded. The write side no longer throws at all, because it no longer decides anything per string.

Once the encoding is chosen per file (see the reply on the `Write` bundle), the bundle is a single plain encoding for the whole serialize — `StrictUtf8` for a UTF-8 file, Mutagen's own language default for everything else — so there is no `MutagenEncodingFallbackWrapper` on the write path, no `ExceptionFallback`, and nothing to throw once per `GetByteCount` and again per `GetBytes`. Your "six figures of exceptions inside the serialize" was the right thing to be worried about; the fix removed the fallback rather than making it cheaper.

Re-measured after the rewrite, since the read path gained the lane recording. A 20,000-weapon plugin (2.3 MB), every `FULL` Windows-1252 and not valid UTF-8, swept through `CreateFromBinaryOverlay` touching every `Name`, three runs on a box shared with other lanes:

| read encoding | sweep |
|---|---|
| Mutagen default (1252, no fallback) | 76 / 60 / 133 ms |
| throw-driven `FallbackWrapper(strict utf8, 1252)` | 246 / 242 / 435 ms |
| shipped (ASCII probe + `Utf8.IsValid` + lane record) | 32 / 62 / 38 ms |

So the lane recording is free at this scale and the read still lands at or under the unfixed default.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: Folded exactly as suggested. `Utf8First.GetByteCount` and `GetBytes` now throw `NotSupportedException("read encoding; a write picks its encoding per file through PluginTextEncoding.WriteInPlace/WriteNew")`.

That turned out to matter more after the per-file rewrite than before: the write encoding is now chosen per file, so a plain string subrecord encoded through the read encoding would not merely disagree with `Write` — it would put one string in the file in a different encoding from every other one, which is the single thing the new design exists to prevent. Making it fail by name is the right shape. The full suite and the 128 probes pass with the throw in place, so nothing reaches it today.

The type is also no longer exposed: `Utf8Then1252` as a public field is gone, and the read parameters are built per plugin by `ReadFor` / `ReadWith`, so there is no public handle to reuse it through by accident.

src/housecarl-core/WriteEngine.cs
Avick3110: Fixed. The paragraph now says what is actually true of the call: it reads the header only and deliberately WITHOUT the resolver's game-Data strings redirect — the localized flag is in the header, so unlike reading the strings themselves it needs no fallback and cannot be wrong about the flag because a strings source is missing — and it notes that it still opens under the ordinary read parameters, so the encoding is the one every other read uses.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: The remaining miss is closed rather than stated, in `0f125ee1`.

The legacy lane now writes through a **strict** encoder (`EncoderFallback.ExceptionFallback`), so "can this file hold this text" is the answer of the thing that would have written it, at the moment it writes — no second table to drift. The two sides then differ because their situations differ:

- **A new file** has no bytes to preserve, so it does not refuse. The first pass goes out in the provenance-resolved lane; if the strict encoder says a value has no spelling there, the staged temp is discarded and the whole file is written again as UTF-8. The resolution is therefore "UTF-8 if any contributing plugin resolved as UTF-8, **or if any value needs it**" — never half a file, never a substituted character. Only the unspellable case pays the second pass.
- **An in-place write** cannot do that: changing the file's encoding would convert every other name in it, which is the mixed/converted failure this design exists to prevent. So it refuses, and nothing is written — the staged temp is discarded and the original is byte-intact.

The refusal names the value, the character with its codepoint, the encoding, and the fix: *houseCARL did not write 'X.esp' — the file is unchanged and nothing was staged. Its text is in Windows-1252 (that is what reading it resolved to), and the value "…" contains 'エ' (U+30A8), which Windows-1252 has no spelling for — writing it here would either store a '?' in place of the character or convert every other name in the file to UTF-8. Write this value into a new patch instead (drop in_place=), which is free to be a UTF-8 file.*

One deviation from the steer, deliberate: it names the **value**, not the field path. The check is the encoder's own answer, and the encoder has no field context — `EncoderFallbackException` carries a character and an index into a string the caller never sees. Recovering a field path would mean a second walk over the record beside the encoder, which is the "second table" the same instruction rules out, so the wrapper captures the whole string being encoded instead and the sentence quotes it. The value is what the caller typed, so it is recognisable; if the field path is wanted it has to come from the apply layer, above the write, as its own change.

Two arms, both green first run: a Japanese value typed into a patch off an ASCII-only plugin makes the patch UTF-8 (and is asserted NOT to be question marks); the same value into an in-place edit of the Windows-1252 plugin refuses, names the character and the codepoint and the remedy, and the file's bytes are `Assert.Equal` to what they were before the call.

`2098/2098` and `ci-all 128/128` after rebasing onto `7a1f03b5`.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1833,6 +1849,13 @@ public static void WriteInPlace(SkyrimMod targetMod, IReadOnlyList<ISkyrimModGet
         // temp is already discarded on any throw (nothing on disk), original byte-intact.
         string staged;
         try { staged = WriteInPlaceStaged(targetMod, ordered, outputPath); }
+        // A value this file's own encoding has no spelling for. In place there is no second pass — rewriting the file
+        // as UTF-8 would convert every other string in it — so this is a refusal, with the original byte-intact.
+        catch (Exception ex) when (PluginTextEncoding.RootUnspellable(ex) is { } bad)
+        {
+            throw new UnspellableTextException(
+                PluginTextEncoding.UnspellableRefusal(bad, Path.GetFileName(outputPath)), bad);
+        }
```
Avick3110, line 1872: **The refusal is emitted double-led, with the exception type name in it.** `UnspellableTextException` derives from `InvalidOperationException` and no caller special-cases it, so every in-place lane catches it in its generic arm: `ApplyInPlaceCore` (WritePatchBuilder.cs:1226), `RemoveRecordsInPlaceCore`, `ForwardRecordsInPlaceCore`, `CreateRecordsCore`'s in-place branch and `RemapEngine.RepointInPlace` all fall through to `WritePatchBuilder.SerializeFailure`, which returns `lead + WriteEngine.Describe(ex)`. `Describe` prefixes the type name, so the user sees:

`error: writing 'X.esp' in place failed (serialize or commit; the existing file is untouched): UnspellableTextException: houseCARL did not write 'X.esp' — the file is unchanged and nothing was staged. Its text is in Windows-1252 …`

Two leads for one refusal, plus a leaked type name — and the lead attributes it to "serialize or commit", a step that was aborted. This is the exact class `SerializeFailure` already special-cases at WritePatchBuilder.cs:2165-2169 for `UnopenableBaselineMasterException` and `LocalizedTargetUnsupportedException` ("the exception's own message is the whole sentence"). `Utf8NameTests.AJapaneseValueIntoAWindows1252FileRefusesAndWritesNothing` only asserts `Contains`, so it does not catch this. Add `if (b is UnspellableTextException ut) return ut.Message;` to the same walk in `SerializeFailure`.

src/housecarl-core/WriteEngine.cs
```diff
@@ -1749,14 +1749,30 @@ static string WritePatchStaged(SkyrimMod patchMod, ISkyrimModGetter[] ordered, M
...
+
+        try
+        {
+            if (PluginTextEncoding.NewFileIsUtf8(patchMod)) Serialize(PluginTextEncoding.Utf8Bundle);
+            else
+                // The strict encoder is the CHECK: a value the language default cannot spell stops the write rather
+                // than landing as '?'. A new file has no bytes to preserve, so the answer is simply to write the whole
+                // thing again as UTF-8 — never half of it, and never a substituted character.
+                try { Serialize(PluginTextEncoding.LegacyBundle); }
+                catch (Exception ex) when (PluginTextEncoding.RootUnspellable(ex) is not null)
```
Avick3110, line 1784: **A LOCALIZED output still writes `?` into its `.STRINGS` tables, and the strict-encoder retry cannot fire for it.** `WithEmbeddedEncodings` covers only text embedded in the plugin file. A `TranslatedString` in a mod with `UsingLocalization` is written by Mutagen's `StringsWriter`, which takes its encoding from its own `IMutagenEncodingProvider` (default: the language answer = 1252 for SkyrimSE+English), not from this bundle — the fixture at `Utf8NameTests.WriteLocalized` demonstrates exactly that by having to pass `WithStringsWriter(new StringsWriter(..., new AlwaysUtf8(), ...))` to produce a UTF-8 table.

Scenario: `housecarl_apply ... into=SomeLocalized.esp` (the extend lane, `WritePatchBuilder.ApplyCore`) sets a `Name` to a Japanese string. `patchMod` was loaded from a localized plugin so `UsingLocalization` is true; the FULL goes to the strings table through the default 1252 writer and lands as `?`. `NewFileIsUtf8` may well be false, but the legacy pass never throws — `SpellCheck` is only wired to the embedded encodings, so it never sees that value — and the UTF-8 retry at line 1770 never runs. Nothing in the patch lane refuses a localized output the way `WriteInPlace` does (`UsingLocalization` appears only at WriteEngine.cs:1815/1824). So the silent-`?` half of #740 survives on this path. Either pass a `StringsWriter` built from the resolved lane's provider, or refuse a localized patch output the way the in-place lane already does.

src/housecarl-core/PluginTextEncoding.cs
```diff
@@ -0,0 +1,320 @@
...
+    ///
+    /// <para><c>NonTranslated</c> — plain string subrecords — defaults to Windows-1252 unconditionally, never from
+    /// the language, so it takes an explicit override. <c>NonLocalized</c> — an inline <c>FULL</c> in a plugin that
+    /// is not flagged localized — is read from the provider for the target language, so it is deliberately left to
+    /// the provider rather than pinned by an override: that is what keeps language selection untouched.</para></summary>
+    public static BinaryReadParameters ReadWith(string pluginPath, StringsReadParameters? baseline)
+    {
+        // A fresh sink per open, so a file replaced on disk is re-resolved rather than keeping the old answer.
+        var sink = new LaneSink();
+        Lanes[Path.GetFileName(pluginPath)] = sink;
```
Avick3110: **A lazy overlay opened over the same filename silently discards a lane the eager read had resolved, and the in-place write has no retry to recover.** Every `ReadFor`/`ReadWith` installs a *fresh, empty* `LaneSink` under the plugin's filename, and overlays parse on access — so a header-only open permanently resets the lane to `AsciiOnly` while never re-populating it. `WriteEngine.PluginIsLocalized` (WriteEngine.cs:1884) is exactly that: it opens an overlay, reads only `ModHeader.Flags`, and disposes. Same for any `LoadOrderResolver.OpenOverlay` of a plugin whose strings are never touched.

On the NEW-file lane this self-heals (the strict encoder retries as UTF-8). On the IN-PLACE lane it does not: `WriteInPlace(targetPath)` reads `LaneOf` and, on `AsciiOnly`, picks `LegacyBundle` — and there is no second pass. A UTF-8 plugin whose non-ASCII is all 1252-representable (a French/German translation ESP — the case `AUtf8LatinPluginRoundTripsByteIdenticalThroughAnInPlaceEdit` pins) is then silently rewritten with every `é` flipped from `C3 A9` to `E9`, no throw, no notice — the byte-identity guarantee the PR body states is gone. Today's in-place callers happen to do an eager `CreateFromBinary(target, ReadFor(target))` immediately before the write, so this is latent rather than live, but nothing enforces that ordering and the type gives a caller no way to see it was lost. Consider making the sink additive per filename (merge into an existing sink rather than replace), or having `WriteInPlace` take the lane from the mod it is about to write rather than from a process-global keyed by name.

src/housecarl-core/PluginTextEncoding.cs
```diff
@@ -0,0 +1,320 @@
...
+        => $"houseCARL did not write '{pluginFileName}' — the file is unchanged and nothing was staged. Its text is in "
+         + $"Windows-1252 (that is what reading it resolved to), and the value \"{ex.Text}\" contains "
+         + $"'{ex.Character}' (U+{(int)ex.Character:X4}), which Windows-1252 has no spelling for — writing it here "
+         + "would either store a '?' in place of the character or convert every other name in the file to UTF-8. "
+         + "Write this value into a new patch instead (drop in_place=), which is free to be a UTF-8 file.";
+
+    /// <summary>Unwrap a serialize-boundary throw to the unspellable-value failure inside it, or null. Mutagen runs
+    /// record writes through a PARALLEL path, so the same throw surfaces bare or wrapped in one or more
+    /// <see cref="AggregateException"/>s — the same shape <c>WriteEngine.RootNullArm</c> normalizes.</summary>
+    public static UnspellableTextException? RootUnspellable(Exception ex)
```
Avick3110, line 216: **`RootUnspellable` masks an unrelated concurrent serialize failure, unlike its sibling `RootNullArm`.** The comment two methods down (`RootNullArm`, WriteEngine.cs:1721) is explicit that it re-stamps *only when EVERY leaf's root* is an NRE, precisely so "a genuine other error keeps its own type + message; never mask an unrelated throw." This one returns the **first** leaf whose chain holds an `UnspellableTextException` and ignores the rest.

Mutagen's record writer is parallel, so an `AggregateException` can carry one unspellable-value leaf alongside, say, a `MissingModException` leaf. On the in-place lane (WriteEngine.cs:1854) that means the user is told the write was refused because Windows-1252 cannot spell a character and to use a new patch instead — when the write would also have failed for an unresolvable master, which the "drop in_place=" remedy will not fix. Mirror `RootNullArm`: return the unspellable only when every flattened leaf roots in one.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: Folded in `a29236ca`. The lane record is no longer reset by an open that decodes nothing.

`ReadFor`/`ReadWith` now take the sink from `SinkFor(path)`, which keys it by filename **and the file's version** — last-write time plus length, the same cheap stamp the resolver's freshness check uses. Every open of the same file version shares one sink, so a header-only overlay (`PluginIsLocalized`, or any `OpenOverlay` whose strings are never touched) adds nothing and takes nothing away; a file actually changed on disk gets a fresh one, which is the reason a plain accumulate-forever sink would have been wrong in the other direction. A stat that fails keeps whatever is recorded — losing a resolved lane is the outcome with teeth, so that is the direction it fails in.

You were right that this was latent rather than live only by accident of caller ordering, and that the type gave a caller no way to see the lane had been lost. It no longer depends on that ordering.

`AHeaderOnlyOpenDoesNotCostAFileTheLaneItsFullReadResolved` pins it in the order you described: full read of the UTF-8 Latin plugin, then `WriteEngine.PluginIsLocalized` on it (the header-only open), then `LaneOf` still `Utf8`, then an in-place edit that leaves the accented name as UTF-8 bytes with the 1252 spelling asserted absent.

src/housecarl-core/WriteEngine.cs
Avick3110: Folded — refused rather than routed, in `a29236ca`.

Routing was the other option and it does not survive contact with the staging: the strings writer puts the tables beside the plugin, which during stage 1 is the `.housecarl-tmp` directory, and the atomic swap then moves only the plugin file. The tables would be left behind in the temp and the committed plugin would reference string IDs whose tables never arrived — broken independently of encoding. So `WriteEngine.WritePatch` now refuses a `UsingLocalization` output at the same point the in-place lane refuses one: before the staging directory exists, so nothing is written, with a sentence saying the text lives in separate `.STRINGS` files houseCARL cannot write as one set with the plugin and to write the edits into a fresh patch instead.

That closes the silent-`?` you found and also a pre-existing orphaned-tables failure on the same path.

`ALocalizedPatchOutputIsRefusedRatherThanWritingQuestionMarksToItsTables` drives it through `apply into=` with a Japanese value. Reaching the write at all took making the fixture's localized mod houseCARL-owned (`[houseCARL]` / `generated=true` in its `meta.ini`), because the extend lane's ownership gate otherwise stops the call before the write — worth knowing if anyone else tries to exercise this path.

src/housecarl-core/PluginTextEncoding.cs
Avick3110: Folded in `a29236ca` — `RootUnspellable` now mirrors `RootNullArm` exactly: it flattens the aggregate, returns null the moment any leaf roots in something else, and re-stamps only when every leaf is this failure.

Your example is the one that matters: an unspellable value beside an unresolvable master would have been reported as an encoding refusal whose "write it into a new patch instead" remedy fixes neither half. The non-aggregate path is unchanged.

src/housecarl-core/WriteEngine.cs
Avick3110: Folded in `a29236ca`, exactly where you pointed: `SerializeFailure`'s walk now carries `if (b is UnspellableTextException ut) return ut.Message;` beside the baseline-master and localized-target arms, for the same reason those two are there — the write refused before it staged anything, so the lead attributes it to a phase that never ran, and the message is already the whole sentence.

You were also right that the existing arm could not catch it: `AJapaneseValueIntoAWindows1252FileRefusesAndWritesNothing` only asserted `Contains`. It now also asserts `DoesNotContain("Exception")` and `DoesNotContain("serialize or commit")`, so a leaked type name or a doubled lead fails the arm rather than riding under it. The new localized-patch arm asserts the same.

PR #753, opened 2026-09-14
Say what the apply zip, a nested set's verb, and a required member actually take

### LINKED ISSUE #648: [Bug] copy.seed_paths description names op=Merge / op=ReplaceAll on the apply zip, which apply refuses
`housecarl_copy.seed_paths`'s description tells the reader the apply zip (`assignments[]`) takes `op=Merge` / `op=ReplaceAll`. `housecarl_apply` has no top-level `op=`, `assignments[]` declares only `{target, from, from_source}` and refuses undeclared members by name, and `ApplyTools.ExpandZip` hard-codes `Op="CopyFrom"`, so the zip always transplants the whole field. The spelling survives only in that parameter text now that dialogue-authoring's recipe (PR #645) was corrected.

Fix: the description says what the zip does (a whole-field CopyFrom per target) and drops the two verbs. One sentence, one test on the published description if one exists for this parameter. Found folding Aaron's review on #645, 2026-09-08.

### LINKED ISSUE #652: [Bug] compose.sets[].verb: description names five verbs, the gate accepts eight; required nested members still publish a null type
`compose.sets[].verb`'s published description names five verbs; the request it builds is validated by the same rulebook switch an `ops[].op` is, which accepts all eight write verbs. Description and gate disagree, so #650's pass (PR #651) could not publish an `enum` for it: neither set is true of both.

Decide which is right (does a composed set legitimately take Remove, CopyFrom, and the index verbs, or should the gate refuse them there?), make the description and the gate say the same thing, then let the nested-schema pass mark it from the table.

Related, same PR: a member marked `required` still publishes `type: ["string","null"]`, so a validating client can satisfy the schema with an explicit null the server then refuses; the nullable type on required nested members should go with it.

Found building #650, 2026-09-08.

Avick3110:
Three published shapes said something the server does not do, and this makes each say what it does.

**#648 — `housecarl_copy.seed_paths`.** The description routed a caller to `housecarl_apply`'s `bundle=`/`assignments=` zip "where op=Merge and op=ReplaceAll are your choice". That zip has no `op=` at all: `assignments[]` declares only `{target, from, from_source}` and refuses an undeclared member by name, and `ApplyTools.ExpandZip` hard-codes `Op="CopyFrom"`, so it transplants each named field whole onto the target — one CopyFrom per named field per target (`ExpandZip` emits N targets x M bundle paths). Both the parameter text and the refusal that routes there (`WriteSentences.CopySeedShapeRoute`, which carried the same advice — the issue thought the parameter text was the last copy) now say that in one sentence. `CopyParserProbe`'s arm on that refusal moves with it.

**#652 — `compose.sets[].verb`: the answer is five, and the gate now refuses the other three.** The description named five while the gate took all eight, so the two disagreed and #650 could publish no enum.

My first reading of this was **seven**, and it was wrong — recorded here because the PR shipped it once. The reasoning was that `WriteEngine.BuildStruct` replays each nested set through `ApplyVerb` itself, so every verb `ApplyVerb` implements at a leaf works there, `ReplaceAll` and `Merge` included. What that missed is the INPUT side: `NestedSet` is `{path, verb, value, key, compose}` — there is no `values`, no `entries`, no `from`/`from_source`. `MapStruct` therefore builds every nested `WriteRequest` with `Values = null` and `Entries = null`, so `ApplyListVerb`'s `ReplaceAll` iterates an empty `Values` and `ApplyDictVerb`'s `Merge` iterates an empty `Entries`. Both then report a write that did not happen — the silently degraded mode the third cornerstone forbids — and `CopyFrom` reaches apply as a verb the leaf does not take. A verb whose input the shape cannot carry is not a verb the shape takes, whatever the engine would do with it.

So the vocabulary is the five: `Set`, `Add`, `Remove`, `SetAtIndex`, `InsertAtIndex`. All three slotless verbs are refused at pre-flight in one sentence each, naming the slot they read and the fact a nested set has no member for it. Giving a nested set a `values=` slot would make some of them expressible, but that is a change to the grammar's shape and a SPEC question, not a reading of it — it is not in this PR.

`WriteVerbs.InCompose` derives the five as `All` minus a named `SlotBearing` list, with the same asserted-subtraction throw `OnCreate` uses, and `INV4-COMPOSEHOMES` holds it against `InComposeRecital` and an independently written list, the way `INV4-CREATEHOMES` does for create. The published description is lane-neutral: `NestedSet` is shared verbatim by `housecarl_create`, which has no `CopyFrom` op, so the description points at `housecarl_apply`'s op rather than at a `from=` a create caller cannot put anywhere. The runtime refusal, which CAN tell the lanes apart, splits its remedy on the same `siblingEditorIds` signal the `@editorid` gate beside it already reads.

**Same PR — required nested members, and the other direction.** A member marked `SchemaRequired` still published `type: ["string","null"]`, so a validating client could satisfy the schema with an explicit null the server refuses; the publication pass now drops the null arm, reading both spellings the generator uses (a `type` array and an `anyOf` union). The reverse also had to hold: `NestedSet.Verb` is now `string?` so its enum carries the null the gate DEFAULTS on (`MapStruct` reads a null verb as `Set`), matching `ApplyOp.op` and `CreateFieldOp.op`.

**What a reviewer should check.** Published descriptions and schemas changed, so the description-bound assertions moved deliberately: `CopyParserProbe`'s seed-shape-route arm, and three new arms in `PublishedNestedConstraintTests` / `ComposeNestedVerbTests`. The tools-list snapshot in `src/housecarl-mcp-tests/data` did NOT need updating — it captures tool NAMES only; the nested `required`/`enum` expectations there are derived from the C# shapes and moved on their own. `.claude/skills/` carries no copy of the old `op=Merge`/`op=ReplaceAll` advice (checked; the dialogue-authoring recipe was already corrected in #645).

The two lane arms go to `CorpusRulebook.Validate` directly rather than through `housecarl_create`: create has no `dry_run`, so a call that stopped being refused — the one regression those arms exist to catch — would allocate a patch in the shared fixture's ModsDir. `Validate` writes nothing whatever it answers.

Verified in the worktree, rebased on origin/main: `dotnet build -c Release` clean, `dotnet test --filter "tier!=bridge"` 2049/2049 pass, `ci-all` 128/128 pass. Two arms were sabotage-checked rather than assumed: reverting `Verb`'s nullability reddens the new enum-nullability arm alone, and commenting out the null drop reddens the required-member arm alone.

Closes #648
Closes #652


src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1133,13 +1133,22 @@ bool IsValueCoercibleElement(FieldSchema leaf)
...
             return buildErr;
         foreach (var s in spec.Sets ?? new())
+        {
+            // The one verb a nested set cannot take. The rest run through the verb engine itself (BuildStruct
+            // replays them with ApplyVerb), but CopyFrom reads a SOURCE RECORD the nested shape has no slot to
+            // name, so it would pass the leaf gate and then throw as an unknown verb at apply.
+            if (string.Equals(s.Verb, WriteVerbs.Transplanting, StringComparison.Ordinal))
+                return $"'{WriteVerbs.Transplanting}' is not a verb a compose's nested sets take — it copies a field " +
+                       "from ANOTHER record, which only an op can name (from= / from_source=), so make it its own op " +
+                       $"on the field itself. Legal here: {string.Join(", ", WriteVerbs.InCompose)}.";
```
Avick3110: This refusal is a dead end on the CREATE surface. `StructSpecContents` is shared by both lanes (that is what `siblingEditorIds` threading is about), so `housecarl_create` with `ops:[{field_path:'Entries', op:'Add', compose:{type:'LeveledItemEntry', sets:[{path:'Data.Reference', verb:'CopyFrom'}]}}]` now gets told to "make it its own op on the field itself (from= / from_source=)". On `housecarl_create` that op does not exist: `CreateFieldOp` publishes `SchemaValues(CreateVerbs)` — `WriteVerbs.OnCreate`, i.e. no `CopyFrom` — and carries no `from` / `from_source` members at all, so the strict reader refuses both by name. The caller follows the remedy and hits a second refusal saying the opposite ("There is no CopyFrom on this surface").

The remedy needs to split by lane the way the surrounding code already does — on the create lane the honest route is: create the record, then `housecarl_apply` with `op='CopyFrom'` into the same patch.

src/housecarl-mcp/WriteSentences.cs
```diff
@@ -299,14 +299,13 @@ internal static string CopySourceFault(string what, SourceArmRef? arm, string ca
...
-    /// already does — and there the caller picks replace-vs-merge with the grammar that lane already has, rather
-    /// than this one inventing a second answer.</summary>
+    /// already does — as a whole-field transplant per target, which is the only thing that zip does.</summary>
     [MustState("seed_paths takes a record link or a list of record links", ToolNames.Apply)]
     internal const string CopySeedShapeRoute =
         " — seed_paths takes a record link or a list of record links, and nothing else. Copying a field whose " +
         "entries carry links INSIDE them is a field-bundle copy: use " + ToolNames.Apply + "'s bundle=/assignments= zip, " +
-        "where op=Merge and op=ReplaceAll are your choice between merging into the target's entries and replacing " +
-        "them. Nothing was written.";
+        "which transplants each named field WHOLE onto the target — one CopyFrom per target, replacing what the " +
```
Avick3110: "one CopyFrom per target" undercounts whenever `bundle=` names more than one field. `ApplyTools.ExpandZip` emits one `CopyFrom` op **per bundle path per assignment** — its own summary says so: "for each assignment, ONE CopyFrom op per bundle path… N targets x M paths is N*M ops", and the origin string it stamps is `assignments[i] x bundle[j]`. A caller reading this sentence with `bundle=['HeadParts','Factions']` and two assignments expects 2 ops and gets 4, and a refusal naming `assignments[0] x bundle[1]` will not match the shape the sentence just described.

"one CopyFrom per named field, per target" says the same thing accurately. The same phrase is in `CopyTools.seed_paths` (CopyTools.cs:44) and in the CHANGELOG entry.

src/housecarl-mcp/NestedSchemaConstraints.cs
```diff
@@ -157,6 +168,20 @@ static bool Walk(Type type, JsonObject node)
         return changed;
     }
 
+    /// <summary>Drop <c>"null"</c> from a member's published type. Generic: it reads the DOCUMENT, so every member
+    /// marked required loses the null arm without anything here naming one. A type that is null and nothing else is
+    /// left alone — narrowing that publishes a member no value can satisfy.</summary>
+    static bool DropNull(JsonObject member)
+    {
```
Avick3110, line 176: `DropNull` only handles a member whose schema spells `type` — `if (member["type"] is not JsonArray types) return false;`. A member the generator emits as a union (`anyOf: [{type:"string"}, {type:"null"}]`) keeps its null arm, and `Walk` itself already handles `anyOf` nodes elsewhere, so the shape is one this pass expects to meet.

The new test cannot catch that case either: `PublishedNestedConstraintTests.AdmitsNull` reads only `type` as well, so such a member is simply absent from `NullableMembers` and `NoRequiredMemberPublishesANullableType` reports clean on a member that still admits an explicit null. No required member is published that way today, so this is latent — but the pass and the test agree by having the same blind spot, which is the one thing the file's doc comment says it is trying to avoid ("written independently of the publication pass's"). Reading the arms in both places closes it.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -1223,7 +1223,7 @@ public sealed record NestedSet
     [SchemaRequired, JsonPropertyName("path"), Description("Dotted path within the struct, e.g. 'Data.Level'.")]
     public string? Path { get; init; }
 
-    [JsonPropertyName("verb"), Description("Set (default) | Add | Remove | SetAtIndex | InsertAtIndex.")]
+    [SchemaValues(SchemaVocabulary.ComposeVerbs), JsonPropertyName("verb"), Description(WriteVerbs.InComposeRecital + ". The nested write runs through the same verb engine an op does, so the verb is chosen by the nested target's own cardinality. There is no CopyFrom here — it reads a SOURCE RECORD and a nested set has no slot to name one; make it its own op on the field itself.")]
```
Avick3110: **medium** — this enum publishes narrower than the gate accepts, which is the same defect this PR fixes on the other half.

`Verb` is a non-nullable `string` with a default, and the SDK's generator types it accordingly. I dumped `tools/list` off a built server to check rather than infer it:

```
housecarl_apply/.../compose/properties/sets/items
    path {'type': ['string', 'null']}
    verb {'type': 'string'}
    ...
```

So `NestedSchemaConstraints.AdmitsNull(member)` is false here and the stamped `enum` is the seven names with no null entry. But `LoadOrderService.MapStruct` (LoadOrderService.cs:7965) reads a null or blank verb as `Set`:

```csharp
Verb = string.IsNullOrWhiteSpace(ns.Verb) ? "Set" : ns.Verb,
```

A schema-validating client will now refuse `{"path":"Data.Level","verb":null,"value":"1"}` a hop before the server, which answers it. That is exactly what the class doc-comment on `NestedSchemaConstraints` promises does not happen — "an `enum` on a member the generator typed as nullable carries null, so nothing it stamps publishes narrower than the gate accepts" — and the escape hatch it names does not fire, because this member is the one marked member the generator does **not** type as nullable. `ApplyOp.Op` and `CreateFieldOp.Op` are both `string?`, so they pick up the null arm and this one silently does not.

Two ways out, either fine: declare `public string? Verb { get; init; }` and let the existing `AdmitsNull` branch add the null (`MapStruct` already defaults it, so nothing downstream changes), or add the null arm at the stamp for a member whose gate defaults rather than refuses.

Worth a test arm either way — nothing today asserts that a member whose gate DEFAULTS publishes an enum admitting the values it defaults on, so this passed the new `NoRequiredMemberPublishesANullableType` arm and `EveryPublishedOccurrenceOfAMarkedShapeCarriesItsRequiredAndEnum` untouched (the latter only checks the published enum against the published type, so it agrees with the narrowing by construction).

Separately and pre-existing, not for this PR: a blank `"verb": ""` is normalized to `Set` too, and no enum on this surface admits it.

Avick3110: Folded, and you were right that this is the same defect on the other half. Taken the first of your two routes: NestedSet.Verb is now string?, so the generator types it [string,null], the existing AdmitsNull branch adds the null entry, and MapStruct keeps defaulting it. Added the test arm you asked for: EveryMemberWithAClosedValueSetPublishesATypeAdmittingNull — every member carrying a closed value set on this surface is a verb slot the server defaults rather than refuses, so its published type must admit null. Sabotage-checked: reverting Verb to non-nullable turns that arm red and nothing else. The blank verb stays out of scope, as you said.

src/housecarl-core/CorpusRulebook.cs
Avick3110: Folded. The remedy now splits by lane on the same signal the @editorid gate below it already reads (siblingEditorIds null = edit lane, non-null = create). On the create lane it says to create the record here and copy the field in with a second housecarl_apply call. Added a create-lane arm to ComposeNestedVerbTests that asserts the refusal names housecarl_apply and does NOT name from_source=.

src/housecarl-mcp/WriteSentences.cs
Avick3110: Folded — undercounted, agreed. Now one CopyFrom per named field per target, in all four places: the refusal, CopyTools.seed_paths, CopyParserProbe's arm, and the changelog entry.

src/housecarl-mcp/NestedSchemaConstraints.cs
Avick3110: Folded. DropNull now reads an anyOf union's arms as well as a type array, dropping the null arm and recursing into what is left; the test's own AdmitsNull reads anyOf too, so the two are no longer blind in the same place. Still latent today, as you say - no required member is published as a union.

Avick3110:
Folded all four findings in 184dcdb6 — all real, nothing refuted.

- **sets[].verb published narrower than the gate** — `NestedSet.Verb` is now `string?`, matching `ApplyOp.op` and `CreateFieldOp.op`, so the generator types it `["string","null"]` and the enum stamp carries the null the gate defaults on. New arm `EveryMemberWithAClosedValueSetPublishesATypeAdmittingNull` asserts it; sabotage-checked (reverting the nullability turns that arm red and nothing else).
- **create-lane remedy was a dead end** — the refusal splits by lane on the same `siblingEditorIds` signal the @editorid gate beside it reads, and on create sends the caller to a second `housecarl_apply` call instead of a `from=` that surface does not have. Covered by a create-lane test arm.
- **"one CopyFrom per target" undercounts** — now "one CopyFrom per named field per target", in the refusal, `seed_paths`, the probe arm and the changelog.
- **DropNull blind to anyOf** — both `DropNull` and the test's own `AdmitsNull` now read a union's arms.

Re-verified after rebase on origin/main: build clean, `ci-all` 128/128, tests 2006/2006. One caveat worth recording: across four full test runs one run reported a single failure that the re-runs did not reproduce and whose name I did not capture (only a bare reflection stack reached the tail). Three subsequent full runs are clean, so I believe it is a flake on a machine several lanes are building on, not this change — but I could not settle it, so I am not claiming it was.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -1223,8 +1223,10 @@ public sealed record NestedSet
     [SchemaRequired, JsonPropertyName("path"), Description("Dotted path within the struct, e.g. 'Data.Level'.")]
     public string? Path { get; init; }
 
-    [JsonPropertyName("verb"), Description("Set (default) | Add | Remove | SetAtIndex | InsertAtIndex.")]
-    public string Verb { get; init; } = "Set";
+    [SchemaValues(SchemaVocabulary.ComposeVerbs), JsonPropertyName("verb"), Description(WriteVerbs.InComposeRecital + ". The nested write runs through the same verb engine an op does, so the verb is chosen by the nested target's own cardinality. There is no CopyFrom here — it reads a SOURCE RECORD and a nested set has no slot to name one; make it its own op on the field itself.")]
```
Avick3110: Two of the seven now published here have no input slot on `NestedSet`, and both fail silently rather than refusing.

`NestedSet` is `{path, verb, value, key, compose}` — no `values`, no `entries`, no `composes`. `MapStruct` (LoadOrderService.cs:8023) therefore builds every nested `WriteRequest` with `Values = null` and `Entries = null`. At apply:

- `ApplyListVerb` `case "ReplaceAll"`: `Clear()`, then `foreach (var v in req.Values ?? Array.Empty<string>())` — always empty. The collection is **wiped** and whatever the caller put in `value=` is ignored.
- `ApplyDictVerb` `case "Merge"`: `foreach (var kv in req.Entries ?? new())` — always empty. An unconditional **no-op**, reported as a successful write.

Nothing refuses it first. `CorpusRulebook`'s element-value presence gate is verb-scoped to `Add`/`SetAtIndex`/`InsertAtIndex`; the `ReplaceAll`/`Merge` blocks iterate the null `Values`/`Entries` slots and pass; `VerbLegality` accepts `ReplaceAll` on any list/dict and `Merge` on any dict. The modeled-element branch refuses them by name, but a **coercible-element** or **formlink** collection reached from a compose does not.

Concrete: `compose={"type":"ScriptStringListProperty","sets":[{"path":"Data","verb":"ReplaceAll","value":"Foo"}]}` clears `Data` and reports the write as landed — a silently wrong answer of exactly the kind the cornerstones forbid.

Before this PR the description named five verbs, so nothing routed a caller into those two. Publishing them as a machine-checkable `enum` advertises them as supported. Either make `InCompose` the five verbs the nested shape can actually feed, or refuse `ReplaceAll`/`Merge` in a nested set at pre-flight the way `CopyFrom` now is (same loop, same reason: no slot to name the input).

src/housecarl-core/CorpusRulebook.cs
```diff
@@ -1133,13 +1133,28 @@ bool IsValueCoercibleElement(FieldSchema leaf)
...
+            // name, so it would pass the leaf gate and then throw as an unknown verb at apply.
+            // The remedy splits by LANE on the same signal the @editorid gate below reads: the create surface has no
+            // CopyFrom op to send the caller to, so naming one there would route them into a second refusal.
+            if (string.Equals(s.Verb, WriteVerbs.Transplanting, StringComparison.Ordinal))
+                return $"'{WriteVerbs.Transplanting}' is not a verb a compose's nested sets take — it copies a field " +
+                       "from ANOTHER record, which only an op naming the source can do, so "
+                     + (siblingEditorIds is null
+                           ? "make it its own op on the field itself (from= / from_source=)."
+                           : $"create the record here and copy the field in with a second {ToolNames.Apply} call "
+                             + "(op='CopyFrom') into the same patch.")
```
Avick3110: The runtime refusal splits its remedy by lane here, correctly — but the published description on `NestedSet.verb` (WriteTools.cs:1226) cannot, and it hands every lane the apply-lane remedy: "make it its own op on the field itself."

`NestedSet` is shared verbatim by `housecarl_create`'s `ops[].compose.sets[]`. On that surface there is no `CopyFrom` op to make it — `CreateFieldOp.op` publishes `CreateVerbs` (the vocabulary minus `CopyFrom`) and its own description says "There is no CopyFrom on this surface." So a create caller who reads the nested `verb` description is told to do the one thing that lane refuses, which is the second-refusal loop the `siblingEditorIds` split here exists to avoid.

Worth wording the description so it survives both lanes (e.g. "copying a field from another record is `housecarl_apply`'s `CopyFrom` op"), rather than naming `from=` / `from_source=` a create caller has no op to put them on.

src/housecarl-mcp-tests/ComposeNestedVerbTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        => Served(ComposeWithNestedSet(@"{""path"":""ActorValue"",""value"":""Destruction""}"),
+                  "Set Conditions[0].Data");
+
+    /// <summary>The CREATE lane shares this gate, and its remedy has to be one that lane can follow: there is no
+    /// CopyFrom op on the create surface, so pointing at from= / from_source= there would land the caller on a
+    /// second refusal saying the opposite.</summary>
+    [Fact]
+    public void TheCreateLaneGetsARemedyThatLaneCanFollow()
+    {
+        var r = CreateTools.Create(Svc, records: Je(
```
Avick3110: **low** — this arm writes into the shared world when it fails, and the class doc above it says it does not.

`CreateTools.Create` has no `dry_run` parameter (the two arms above use `ApplyTools.Apply(..., dry_run: true)`; create's lane parameters are `patch` / `into` / `in_place` only). This call passes no lane, so on the default `patch` lane a create that is NOT refused allocates a fresh `houseCARL - Patch` mod folder and an .esp in the `records` collection fixture's ModsDir — the fixture every other test in that collection reads.

Today it is safe only because the gate under test refuses first, which is circular: the one regression this test exists to catch is exactly the one that makes it write. `standards/TESTING.md` states the rule directly: "A test that mutates the world (rewrites a plugin, touches an mtime) builds its own instance so it cannot poison the shared one."

The class comment at line 12 ("Dry runs: the shared world must stay unwritten") is now false for a third of the class, which is how this would be missed on a later read.

Either give this arm its own instance, or assert the refusal against a lane that cannot land (an `into=` naming a patch that does not exist would refuse on its own, so not that) — simplest is a class-scoped fixture of its own for the create arm.

src/housecarl-mcp-tests/PublishedNestedConstraintTests.cs
```diff
@@ -124,13 +127,19 @@ static void Walk(string tool, string path, JsonElement node, List<Node> found)
...
             : Array.Empty<string>();
 
-        found.Add(new Node(tool, path, members.OrderBy(n => n, StringComparer.Ordinal).ToArray(), required, enums));
+        found.Add(new Node(tool, path, members.OrderBy(n => n, StringComparer.Ordinal).ToArray(), required, enums,
+                           nullable.OrderBy(n => n, StringComparer.Ordinal).ToArray()));
     }
 
     /// <summary>Does a published member's <c>type</c> accept a JSON null? Written here off the served document, not
     /// read from the pass, so the two are still independent statements about the same member.</summary>
+    /// <para>Both spellings the generator uses: a <c>type</c> (string or array) and an <c>anyOf</c> union whose arms
```
Avick3110, line 136: **low** — the new paragraph sits OUTSIDE the doc comment it belongs to: line 135 closes `</summary>`, and this `<para>` is then a top-level element with no enclosing tag. It is not part of the summary and will not render with it; every other multi-paragraph doc comment in this file and in `NestedSchemaConstraints.cs` keeps the `<para>` inside `<summary>`.

Move the `</summary>` to the end of the new paragraph.

src/housecarl-mcp/WriteTools.cs
Avick3110: Correction to my own comment above, on the IMPACT only — the finding stands but "wiped" overstates it.

A nested set never runs against a live collection: `BuildStruct` instantiates the struct fresh and only then replays `spec.Sets` through `ApplyVerb`, and its three callers (`ApplyScalarVerb`'s `Set`+`Struct`, the list `Add`/`SetAtIndex`/`InsertAtIndex` compose, and `composes[]`) all hand it a newly built instance. So the `Clear()` clears a collection that is already empty.

What is left is still a silently degraded mode rather than data loss: `ReplaceAll` on a nested set discards whatever the caller put in `value=` and reports the write as landed, and `Merge` is an unconditional no-op reported the same way. Both are unreachable inputs advertised by the new `enum`. The remedy is unchanged — either narrow `InCompose`, or refuse the two at pre-flight in the same loop `CopyFrom` now is, for the same reason (no slot to name the input).

Avick3110: Folded, and this changes the answer — thank you, this is the finding of the two rounds. You are right that I reasoned from the apply side only (BuildStruct replays through ApplyVerb, so every verb the engine implements at a leaf works) and never checked the INPUT side: NestedSet has no values, entries or from members, so MapStruct builds every nested request with Values and Entries null and those two verbs consume nothing. InCompose is now the five the shape can feed, derived as All minus a named SlotBearing list with the same asserted-subtraction throw OnCreate uses, and all three slotless verbs are refused at pre-flight in one sentence each naming the slot they read. INV4-COMPOSEHOMES holds the list against the recital and an independently written one, as INV4-CREATEHOMES does for create. Giving a nested set a values= slot would make some of them expressible, but that is a change to the grammar's shape and a SPEC question, so it is not in this PR. The PR body now records that seven was wrong and why.

Avick3110: Noted, and the correction is folded into the wording rather than the fix: the changelog and the refusals say each consumed nothing and reported a write that landed, not that anything was wiped. The remedy you named is the one taken - refused at pre-flight in the same loop CopyFrom is, for the same reason.

src/housecarl-core/CorpusRulebook.cs
Avick3110: Folded. The published description is now lane-neutral - it names housecarl_apply's CopyFrom op rather than from= / from_source=, and it says the same for ReplaceAll's values= and Merge's entries=. The runtime refusal keeps the lane split, since it can tell them apart and the description cannot.

src/housecarl-mcp-tests/ComposeNestedVerbTests.cs
Avick3110: Folded, and you are right that it was circular - the one regression the arm exists to catch is the one that makes it write. Both lane arms now go to CorpusRulebook.Validate directly instead of through housecarl_create: Validate writes nothing whatever it answers, so the class comment is true again for every arm. That also let me assert both lanes' remedies rather than just create's.

src/housecarl-mcp-tests/PublishedNestedConstraintTests.cs
Avick3110: Folded - the para now sits inside the summary it belongs to.

PR #755, opened 2026-09-14
Refuse three inputs that were silently accepted

### LINKED ISSUE #617: [Bug] place and nif_inspect accept a raw mods path; refuse it in one sentence

### LINKED ISSUE #630: [Bug] apply accepts from_source= on a non-CopyFrom op and ignores it
`from_source=` is honoured only by `op='CopyFrom'`, but an op that names it on any other verb is accepted and written, with the pole silently dropped.

Where it goes: `ApplyTools.cs:148` gates `from=` on `CopyFrom` and refuses it by name on anything else. `from_source` gets no such gate — it is a declared member of `ApplyOp`, so strict binding accepts it, and it maps straight to `BulkOp.FromPlugin` at `ApplyTools.cs:157`. Every consumer downstream requires the verb: `PrepareCopyFromSources` early-returns when no op has `Verb == "CopyFrom"` (`LoadOrderService.cs:5416`), and both `WritePatchBuilder.IsOffOrderCopySource` (`:688`) and `ResolveCopyPole` (`:697`) test the same verb.

So this writes, reports success, and bases the edit on the load-order winner rather than the plugin the caller named:

```
{formid: '0012AB:Mod.esp', field_path: 'BasicStats.Damage', op: 'Set', value: '12', from_source: 'Authored.esp'}
```

That is a silently degraded mode, which the error rule does not allow. Fix: refuse `from_source` on a non-`CopyFrom` verb the way `from=` is already refused, in one sentence naming the op index and the verb it got.

Found while folding review on #623, which scoped the `ops=` description to `op='CopyFrom'`; the description now says the right thing, but the pre-flight still does not enforce it.

### LINKED ISSUE #666: [Bug] housecarl_records plugins= fails the whole call when one named plugin is not in the load order
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/facegen-diagnostics/VERDICT.md`).

A `plugins={...}` read naming several plugins was refused whole because one of them (`Kaidan_WNAM_Fix.esp`, which owns an orphan facegen folder but is not loaded) is not in the order. The refusal is correct for that plugin; failing the other named plugins' reads with it is not. Either answer the loaded plugins and name the missing one beside the result, or refuse in one sentence that names it and tells the caller to drop it.

Avick3110:
Three refusal-ergonomics bugs, one per commit. Each is the same rule: a tool never accepts an input it cannot honour, and when it cannot do the thing it says so in one sentence naming what went wrong and what to try.

`housecarl_apply` accepted `from_source=` on any verb. Only `op='CopyFrom'` reads it — `PrepareCopyFromSources`, `IsOffOrderCopySource` and `ResolveCopyPole` all test the verb — so on another verb the pole was dropped, the edit was based on the load-order winner rather than the plugin the caller named, and success was reported. It is now gated exactly the way `from=` already is, in the same pre-flight loop, naming the op index and the verb it got; nothing is written.

`housecarl_place` and `housecarl_nif_inspect` accepted a raw path into MO2's mods directory, which reads and writes past the virtual file system — only the retired npc-appearance-copy skill's prose stood against it. A new `ModsPathAddress` splits such a path into the mod folder and the Data-relative remainder and builds one sentence handing back the address form every source pole already takes: `source_provider='<mod folder>'` with the Data-relative path. It fires on `place`'s `path=` and `source=` (a `<bsa>|<entry>` source is judged on the archive's own path) and on each `nif_inspect` `mesh_paths=` entry, where it stays a per-path failure so the rest of the batch still reads. A path anywhere else on disk is unchanged — naming one exact copy outside the mods tree is still how you place it. One existing test (`PlaceJsonServedLaneTests`) sourced from a mods path to exercise the withheld pole and now sources from outside the tree instead.

`housecarl_records` with `plugins=` failed the whole call when one named plugin was not in the order — `LoadOrderResolver.ScopeIndices` throws on the first name it cannot find. `CrossQuery` now partitions the scope against the captured view before the scan: the loaded plugins are scanned and the missing ones are named beside the result on the scan note, which both the text and json lanes already render. Answering was chosen over refusing (the issue offers either) because the other named plugins' reads are valid. A scope whose every name is missing is still refused, naming them and carrying the existing absence clause — there is nothing left to scan. `CrossQuery` is the single lane every `housecarl_records` `plugins=` call reaches, so every form (scan, delta, tree, chain, aggregate, walk) gets the same shape. An excluded plugin still refuses the call as before: that is a different fact with its own stated reason.

Review fold (all five findings were real, one commit):

- The place source refusal keeps a `'<archive.bsa>|<entry>'` source's entry in the remedy — a bare archive path would have placed the whole .bsa — and a both-slots member (`formid=` with no `kind=`) is told to name the provider or pick a slot, because no single `source=` names two files and the Data-relative form it used to recommend is itself refused on the retry.
- The source check is a per-member failure now, in `PlaceOne` beside the other source errors, so a 40-destination batch with one pasted mods path still places the other 39. The destination check stays with the malformed members, where the published `assets=` description already puts it.
- The `plugins=` split guards a null entry by its index instead of reaching the resolver as a null key, which came back worded as an internal failure rather than bad input.
- Each missing name carries its own `AbsenceClause` — the untick, the did-you-mean — on the served note and on the refusal, which previously attached one only when exactly one name was missing.
- `CrossQuery` was not the only `plugins=` lane: `OffOrderQuery` hard-failed on the first unknown name, so the same scope behaved two ways depending on `source=`. Both lanes now share one `ScopeSplit`, which owns the split and its three sentences.

One thing the fold turned up that the review did not: the body forms (`fields`/`rows`/`everything`) built their own header and never carried `ScanNote`, in-order and off-order alike, so the missing-plugin line — and the existing unscannable-record and unreadable-plugin notes — were stated on a scan render and dropped on a body render of the same scan. Both header lines carry it now.

Second review fold (three findings, all real, one commit):

- `Path.GetFullPath` resolves a relative path against the server's working directory, so the mods-path check fired on ordinary Data-relative paths whenever that directory sat under the mods tree — the normal address form would stop working for a whole session. `Split` returns early now unless the path is fully qualified.
- The `ScanNote` fold reached only the text header; the json arm of the same body lane takes no header, so json still dropped it. Both body lanes put `scan_note` on the envelope. (The `fields` form in the report renders through a different path whose json already carried it under `notes`; the lane that dropped it is the body-lane forms — `everything`, `rows`, a quantified fields path — and my earlier text test had the same blind spot, so both tests use `form='everything'` now.)
- The both-slots remedy fired only on a fully-qualified `.bsa` under the mods tree, which is precisely the one `source=` that serves two slots — so it denied a documented shape and offered a non-equivalent replacement. It keeps that archive now: `source_provider='<archive filename>'` with no `source=`, a BSA filename being a provider name, with each slot deriving its own entry.

No published tool description changed, so the tools-list snapshot is untouched.

Verified: `dotnet build housecarl.sln -c Release` clean; `dotnet test src/housecarl-mcp-tests -c Release --no-build --filter "tier!=bridge"` 2085/2085 pass; `housecarl-generator ci-all` 128/128 pass. New fixture tests: `ApplyFromSourceVerbGateTests` (the refusal names `ops[0]`, `op='CopyFrom'` and the verb it got, including when `op` is omitted and defaults to `Set`), `RawModsPathRefusalTests` (place's source and destination, and nif_inspect's per-path refusal with the rest of the batch still answered), `RecordsScopeMissingPluginTests` (the loaded plugin's rows plus the missing-plugin line, and the all-missing refusal). Also run against the real install — this worktree's Release `housecarl-mcp` driven over stdio with `HOUSECARL_DATA_DIR` pointed at a scratch config naming the `E:\Authoria - Requiem Reforged` instance (the installed config was not touched), five arms:

- `nif_inspect mesh_paths=["E:\Authoria - Requiem Reforged\mods\360 Ward\meshes\magic\wardbodyfx.nif", "meshes\magic\wardbodyfx.nif"]` — the raw path is refused on its own row with "…is a raw path into MO2's mods folder, which reads past the virtual file system — a mod folder is named, never pathed. Address that copy instead with `source_provider='360 Ward'` and `mesh_paths='meshes\magic\wardbodyfx.nif'`", and the second path is still read in full (94 blocks, 5 shapes, providers `"360 Ward" (loose) > "Skyrim - Meshes1.bsa" (BSA)`). The mod folder the sentence names is the one that really provides the file.
- `place assets=[{path: "meshes\hcprobe\dest.nif", source: "<that same raw path>"}]` — refused pre-flight, "refused — 1 malformed destination(s); nothing placed", the sentence naming `source_provider='360 Ward'` and `source='meshes\magic\wardbodyfx.nif'`. Nothing was written.
- `place assets=[{path: "<that same raw path>"}]` — same refusal on the destination axis, naming `path='meshes\magic\wardbodyfx.nif'`. Nothing was written.
- `records plugins=["Skyrim.esm", "Dawnguard.esm", "Kaidan_WNAM_Fix.esp"] types=["AMMO"]` — served: 49 matches, three AMMO rows with their real winners, and the line "note: plugins= named Kaidan_WNAM_Fix.esp, which the load order does not carry — this answer covers the 2 named plugin(s) that are loaded, and nothing from the missing one(s)." That is the exact call shape from #666, with the plugin the issue names.
- `records plugins=["Kaidan_WNAM_Fix.esp"]` — still refused, naming it: nothing left to scan.

Re-run against the same instance after the review fold:

- `place` with two destinations, one carrying a raw mods source — "placed 1 of 2 asset(s) (1 failed)": the raw-path row fails on its own and the other destination places. Its probe mod folder was deleted from the instance afterwards.
- `place` with a `'<mods path>\Nope.bsa|meshes\magic\wardbodyfx.nif'` source — the remedy keeps the entry: `source='Nope.bsa|meshes\magic\wardbodyfx.nif'`.
- `place assets=[{formid: "013480:Skyrim.esm", source: "<mods path>\Nope.bsa"}]` (both FaceGen slots) — both rows say "Placing BOTH FaceGen slots reads two files, which no single source= names — pass source_provider='360 Ward' with no source=, or set kind= mesh or tint…".
- `records plugins={"names": ["Skyrim.esm", null]}` — "plugins.names[1] is empty — every entry is a plugin FILENAME…", not the internal-failure wording.
- `records plugins=["Skyrim.esm", "Requiem.esp.esp"]` — served, with the cause carried: "…and nothing from the 1 it does not: 'Requiem.esp.esp'. Did you mean `Requiem.esp`?"
- `records plugins=["Skyrim.esm", "Kaidan_WNAM_Fix.esp"] types=["NPC_"] source="EternalWomen.esp"` (a disabled mod's plugin, so the off-order lane) with `project.form='fields'` — 258 matches served with the note beside them, where before the fold the whole call was refused on the scope name.

And after the second fold, with the server's working directory set to a mod folder:

- `nif_inspect mesh_paths=['meshes\magic\wardbodyfx.nif']` from inside `<mods>\360 Ward` — reads the mesh normally; the Data-relative form is unaffected.
- `place assets=[{formid: "013480:Skyrim.esm", source: "<mods>\360 Ward\Nope.bsa"}]` — "Address that archive instead with source_provider='Nope.bsa' and no source=…".
- `records plugins=["Skyrim.esm","Kaidan_WNAM_Fix.esp"] project.form='everything' format='json'` — the document carries `scan_note` naming the missing plugin.

Closes #630
Closes #617
Closes #666


src/housecarl-mcp/PlaceTools.cs
```diff
@@ -124,6 +129,19 @@ static string PlaceTargets(LoadOrderService svc, PlaceTarget[] assets, string? s
...
+    /// source is split at the pipe first, so an archive inside a mod folder is judged on the archive's own path.</summary>
+    static string? RawModsPathProblem(string where, string param, string? value, string? modsRoot)
+    {
+        var v = NullIfBlank(value)?.Trim('"');
+        if (v is null) return null;
+        int pipe = v.IndexOf('|');
+        var probe = pipe >= 0 ? v.Substring(0, pipe) : v;
+        return ModsPathAddress.Split(probe, modsRoot) is not { } hit
+            ? null
+            : ModsPathAddress.Refusal(where, v, hit.ModFolder, hit.RelPath, param, "source_provider");
```
Avick3110: The `<archive.bsa>|<entry>` split probes the archive path but the suggested replacement drops the entry, so the sentence hands back a call that places the wrong bytes.

`v` is the full value (archive path + pipe + entry); `hit.RelPath` is the Data-relative path of the **archive only**. For `source = "E:/MO2/mods/360 Ward/x.bsa|meshes/magic/wardbodyfx.nif"` the refusal reads `source_provider='360 Ward'` and `source='x.bsa'` — following it literally places the whole BSA at the destination instead of the entry. The suggestion should keep the entry, e.g. `hit.RelPath + v.Substring(pipe)`.

Same path has a second dead end: for a `formid=` member with no `kind=` (both slots), `MapTarget`'s `srcOkForBoth` requires `Path.IsPathFullyQualified(srcProbe)`, so the Data-relative `source='x.bsa'` this refusal recommends is itself refused on the retry. That case needs a different sentence (name `source_provider=` and drop `source=`, or set `kind=`).

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -104,9 +104,14 @@ static string PlaceTargets(LoadOrderService svc, PlaceTarget[] assets, string? s
...
         var poleWithheld = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         var door = svc.OpenWriteFormIdDoor();
+        var modsRoot = svc.ModsRootOrNull;
         for (int i = 0; i < assets.Length; i++)
         {
             var a = assets[i];
+            // A raw path into the mods tree goes around the VFS on either axis — the destination it writes at, or the
+            // copy it reads. Refused with the address form rather than honoured, on the member that carries it.
+            if (RawModsPathProblem($"assets[{i}]: ", "path", a.Path, modsRoot) is { } destErr) { problems.Add(destErr); continue; }
+            if (RawModsPathProblem($"assets[{i}]: ", "source", a.Source, modsRoot) is { } srcErr) { problems.Add(srcErr); continue; }
```
Avick3110: Putting the `source=` check into the `problems` list escalates a source-axis problem to an all-or-nothing refusal, which contradicts the published contract for that axis.

The tool description on `assets=` says a malformed member (bad FormID, bad kind, neither/both of formid and path, a both-slots member whose source is not a full .bsa path) refuses the WHOLE call, but "a source that is ambiguous, absent or unreadable is a PER-MEMBER error and the rest still place". A raw mods path is a source problem, not one of the enumerated malformed shapes. Scenario: a 40-destination batch where one member's `source=` was pasted from an MO2 mod folder now places nothing at all; before this change the other 39 placed and that one row carried its own error.

The destination check on line 113 is fine where it is — destinations are already all-or-nothing. The source check wants to be a per-member failure instead, or the published description needs updating to say so (the PR body states no published description changed).

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4464,6 +4489,31 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
+                (view.ContainsPlugin(p.Trim()) ? present : missing).Add(p.Trim());
+            if (missing.Count > 0)
+            {
+                if (present.Count == 0)
+                    return CrossQueryOutcome.Fail(
+                        $"plugins= names {(missing.Count == 1 ? "a plugin" : "plugins")} the load order does not carry and nothing else to scan: "
+                        + string.Join(", ", missing) + "."
+                        + (missing.Count == 1 ? view.AbsenceClause(missing[0]) : "")
+                        + " Drop the name(s), or scope to plugins that are loaded.");
+                scopeMissingNote =
```
Avick3110: The served note drops `AbsenceClause`, so the most common real cause of a missing name is withheld exactly where the caller no longer gets a refusal that would have carried it.

`IndexView.AbsenceClause`'s own doc says "Every ContainsPlugin-false refusal should carry it — a bare not-found makes the reader re-derive a fact the tool already had". The clause is what distinguishes "installed, but UNTICKED in plugins.txt" from a typo, and it supplies the did-you-mean. Scenario: `records plugins=["Skyrim.esm", "Requiem.esp"]` where Requiem.esp is installed but unticked — before this change the caller got the refusal naming the untick; now they get an answer plus "which the load order does not carry" and no way to tell a typo from a tick they need to flip.

The all-missing refusal just above has the mirror gap: `missing.Count == 1 ? view.AbsenceClause(...) : ""` means a two-name all-missing scope carries no cause for either name. `AbsenceClause` is expensive (a profile parse plus an install sweep), so the served path may want a cap — but the single-missing case should carry it, and the refusal should carry one per name.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4464,6 +4489,31 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
+        // plugins' reads are valid, and failing them with it would throw away a whole answer over one bad name. The
+        // missing names ride the scan note so the result says what was left out; only an ALL-missing scope is refused,
+        // because then there is nothing to scan.
+        string? scopeMissingNote = null;
+        if (plugins is { Count: > 0 })
+        {
+            var present = new List<string>(plugins.Count);
+            var missing = new List<string>();
+            foreach (var p in plugins)
+                (view.ContainsPlugin(p.Trim()) ? present : missing).Add(p.Trim());
```
Avick3110: `p.Trim()` throws on a null element, and the throw now escapes to `Guard.Tool`, which reports it as an internal failure rather than bad input.

`RecordsScope.names` is `string[]?`, so `plugins={"names":["Skyrim.esm", null]}` binds a null element. Previously that reached `LoadOrderResolver.ScopeIndices`, whose `_nameToIdx.TryGetValue(null)` raises `ArgumentNullException` inside the scan's `catch (ArgumentException)` (~line 4937) and came back as a named, recoverable error. This partition runs before that try, so the NRE lands in `Guard.Named`, which tells the caller "This is an internal houseCARL failure (the arguments bound fine), not bad input" — the opposite of the truth.

A null/blank-guarded partition (skip or classify as missing) fixes it; `foreach (var p in plugins)` with `(p ?? "").Trim()` is enough.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4464,6 +4489,31 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
         // The caller's own build when its FormID door already captured one, so the tokens it parsed and the
         // records this scan matches come from ONE build; otherwise one build for the scan and every fill it makes.
         var view = pinnedView ?? resolver.Capture();
+        // A plugins= scope naming a plugin the order does not carry answers for the ones it does: the other named
+        // plugins' reads are valid, and failing them with it would throw away a whole answer over one bad name. The
+        // missing names ride the scan note so the result says what was left out; only an ALL-missing scope is refused,
+        // because then there is nothing to scan.
+        string? scopeMissingNote = null;
```
Avick3110, line 4794: The PR body says "`CrossQuery` is the single lane every `housecarl_records` `plugins=` call reaches", but it is not: `RecordsTools.cs:1673` routes a call with an out-of-load-order `source=` to `svc.OffOrderQuery(pole, types, refFks, null, plugins?.names, ...)`, and `OffOrderQuery` still hard-fails the whole call on the first scope name it cannot find ("plugins= scope 'X' is not in the active load order ...").

Scenario: `records plugins=["Skyrim.esm", "NotInstalled.esp"] source={file: "OldPatch.esp"}` still throws the whole answer away, while the same scope without `source=` now answers. Same input, two contradictory behaviours depending on an unrelated axis. Either the partition belongs somewhere both lanes see it, or #666's fix should be stated as scan-lane-only.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Both real, both folded. The remedy for a `<archive.bsa>|<entry>` source now keeps the entry (`hit.RelPath + v.Substring(pipe)`), so it reads `source='Nope.bsa|meshes\magic\wardbodyfx.nif'` rather than naming the archive alone. The both-slots dead end gets its own sentence: the member is told to pass `source_provider=` with no `source=`, or to set `kind=` and pass that slot's path — never a form its own retry would refuse. Confirmed against a real order; the outputs are in the PR body.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Real — folded. The source check moved into `PlaceOne`, beside the other per-member source failures, so a batch with one pasted mods path now reports that row and places the rest; the live re-run shows "placed 1 of 2 asset(s) (1 failed)". The destination check stays with the malformed members, where the published `assets=` description already puts it, so no description changed.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real — folded. `ScopeSplit.Of` walks by index with `(names[i] ?? "").Trim()` and refuses an empty entry by its index before any lookup, so the caller gets "plugins.names[1] is empty — every entry is a plugin FILENAME…" instead of the internal-failure wording. Covered by a fixture test and confirmed on a real order.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real — folded. Both sentences now carry `AbsenceClause` per missing name, so a near-miss gets its did-you-mean and an installed-but-unticked plugin gets the untick. It is paid once per MISSING name only, never on a scope that is wholly present. Left uncapped for now: the list is the caller's own, and capping it would be a rule to remember rather than a fix — say the word if you want a cap and I will add one.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real — folded, and the PR body no longer makes that claim. Both lanes share one `ScopeSplit` now, so `OffOrderQuery` answers a partly-absent scope the same way and adds only the clause that says what the scope MEANS over a file. Verified on a real order: `plugins=["Skyrim.esm","Kaidan_WNAM_Fix.esp"] source="EternalWomen.esp"` served 258 matches with the note, where it used to refuse outright.

Your finding also turned up one you did not have the diff for: the body forms (`fields`/`rows`/`everything`) build their own header and never carried `ScanNote`, in-order and off-order alike — so this note, and the existing unscannable-record and unreadable-plugin notes, were stated on a scan render and dropped on a body render of the same scan. Both header lines carry it now.

src/housecarl-mcp/ModsPathAddress.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// folder's tree IS the Data tree, so the remainder is already what the VFS calls the file, normalized the way
+    /// every other asset path is. The remainder is null when the path names the mod folder itself and nothing
+    /// inside it.</summary>
+    internal static (string ModFolder, string? RelPath)? Split(string? rawPath, string? modsRoot)
+    {
+        if (string.IsNullOrWhiteSpace(rawPath) || string.IsNullOrWhiteSpace(modsRoot)) return null;
+        string full, root;
+        try
+        {
+            full = Path.GetFullPath(rawPath.Trim().Trim('"'));
```
Avick3110: `Path.GetFullPath` resolves a RELATIVE path against the server process's current directory, so this check does not only fire on raw mods paths — it fires on ordinary Data-relative ones whenever the process CWD happens to sit under the mods tree. A modder who starts Claude Code from inside a mod folder (e.g. `E:\Authoria - Requiem Reforged\mods\MyMod`) then gets `nif_inspect mesh_paths=['meshes\magic\wardbodyfx.nif']` refused as "a raw path into MO2's mods folder", and every `place` `path=` and `source=` in that session too — the Data-relative address form, which is the normal one, stops working. The check is only meaningful for a path the caller actually rooted, so gate it on that: `if (!Path.IsPathFullyQualified(rawPath.Trim().Trim('"'))) return null;` before the GetFullPath.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -1553,6 +1553,10 @@ List<KeyValuePair<string, string>> Echo()
                 var bodyEpoch = bodyEpochs.FirstOrDefault() ?? outcome.Stamp;
                 envelope.Add(new("total", outcome.Total.ToString()));
                 headerLine += $"\n{outcome.Total} match(es); bodies for the {keys.Count}-row window below";
+                // What the SCAN owes about its own coverage — a plugins= name the order does not carry, records it
+                // could not parse, a plugin it could not open. The scan lane prints it; a body form reads the same
+                // scan, so withholding it here would answer the same question two ways.
+                if (outcome.ScanNote is not null) headerLine += "\n" + outcome.ScanNote;
```
Avick3110: `headerLine` only reaches the TEXT renderer — the json arm two lines down calls `JsonWire.RenderBatch(bodies, max_chars, sp, out trunc, envelope, evLevers, …)`, which takes no header and no note parameter. So `housecarl_records plugins=["Skyrim.esm","NotInstalled.esp"] project.form='fields' format='json'` still drops the missing-plugin line entirely, along with the unscannable-record and unreadable-plugin notes, while the same call in text mode carries it. That is the same "one question answered two ways" the PR description sets out to close, just moved from text-vs-scan to json-vs-text. The fix is one line on the envelope beside `total`/`not_touched`, e.g. `if (outcome.ScanNote is not null) envelope.Add(new("scan_note", outcome.ScanNote));` — here and at the off-order body lane (line 1785). `RecordsScopeMissingPluginTests.ABodyFormCarriesTheScopeNoteToo` only exercises the text lane, so nothing catches this.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1580,6 +1605,21 @@ public PlaceOutcome PlaceAssets(IReadOnlyList<PlaceRequest> requests, string? pa
...
 
+    /// <summary>The refusal for a <c>source=</c> that reaches into MO2's mods tree, or null when it does not. A
+    /// <c>'&lt;archive.bsa&gt;|&lt;entry&gt;'</c> pair is judged on the archive's path and keeps its entry in the
+    /// remedy; a both-slots member is sent to the provider pole, the only form that serves two files.</summary>
+    string? RawModsSourceRefusal(string source, bool bothSlots)
+    {
+        var v = source.Trim().Trim('"');
+        int pipe = v.IndexOf('|');
+        if (ModsPathAddress.Split(pipe >= 0 ? v.Substring(0, pipe) : v, ModsRootOrNull) is not { } hit) return null;
+        var remedy = bothSlots
```
Avick3110, line 1648: The `bothSlots` remedy states something the tool contradicts, and the case it fires on is the one form that worked. `MapTarget` (PlaceTools.cs ~178) only ever sets `BothSlots` with a non-null source when the source is a FULLY-QUALIFIED `.bsa` path with no `|` — every other shape is already refused there. So this branch fires exclusively on `<mods root>\<Mod>\Something.bsa`, and tells the caller "Placing BOTH FaceGen slots reads two files, which no single source= names", which is false: a full `.bsa` path is precisely the single source= that names both, with each slot's entry derived. Since in an MO2 install every non-vanilla BSA lives under the mods tree, this removes the documented way to place both FaceGen slots from a specific archive, and the offered replacement (`source_provider='<mod>'` with no `source=`) is not equivalent — it auto-resolves the VFS winner for each slot, which can be a loose file from a different mod, without saying so. Either let a fully-qualified `.bsa` under the mods tree through on the both-slots path, or give a remedy that still names that archive.

src/housecarl-mcp/ModsPathAddress.cs
Avick3110: Real — folded. `Split` now returns early unless the path is fully qualified, so the check reads only a path the caller actually rooted. Verified by running the server with its working directory set to `E:\Authoria - Requiem Reforged\mods\360 Ward`: `nif_inspect mesh_paths=['meshes\magic\wardbodyfx.nif']` reads the mesh normally there. A fixture test drives the same case by moving the process CWD into a mod folder.

src/housecarl-mcp/RecordsTools.cs
Avick3110: Real — folded. Both body lanes add `scan_note` to the envelope beside `total`, so json carries what the text header does. Worth recording for whoever reads this later: the shape in your example (`project.form='fields'`) does not take that lane — it renders through `RenderCrossQuery`, whose json already carries the note under `notes`. The lane that reaches `RenderBatch` is the body-lane forms (`everything`, `rows`, a quantified fields path), and that one did drop it on both transports. My earlier text test had the same blind spot and was hitting the scan render, not the header I had changed; it and the new json test both use `form='everything'` now.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Real — folded, and thank you for chasing the reachability: you are right that this branch only ever fires on a fully-qualified `.bsa` under the mods tree, which is exactly the source that serves both slots. The remedy now keeps that archive rather than sending the caller away from it — `source_provider='<archive filename>'` with no `source=`, since a BSA filename is a provider name and each slot derives its own entry from it. That preserves the documented both-slots shape and pins the archive, so it does not fall back to per-slot VFS auto-resolve the way naming the mod folder would. Live: `place assets=[{formid: "013480:Skyrim.esm", source: "<mods>\360 Ward\Nope.bsa"}]` now answers "Address that archive instead with source_provider='Nope.bsa' and no source=…".

PR #758, opened 2026-09-14
Pick place's enable instruction on the lane, not just contention

### LINKED ISSUE #632: [Bug] place tells the fresh-folder lane to sort above a winner it already outranks
`PlaceTools.EnableAndSort` (`src/housecarl-mcp/PlaceTools.cs:298`) picks its sentence on CONTENTION only — whether any placed row had a current winner — and never on the LANE. So a fresh-folder placement over an existing provider prints:

> Enable the mod '<folder>' in MO2 and SORT it (left pane) ABOVE the current winner(s) listed above. Only then does the placed copy win.

On the default lane the mod folder is new: `ResolveOutputPath` creates `<ModsDir>\houseCARL - <name>`, houseCARL never writes `modlist.txt`, and MO2 registers a folder it has not seen at the highest priority. The new mod therefore outranks the previous winner the moment it is ticked, and the sorting step is work the user does not need to do. The no-contention arm of the same method already says the right thing ("once enabled the placed copy wins").

The sort instruction is correct on `into=`, where the placement lands in an existing patch mod whose priority is already fixed.

Fix: branch on the lane as well as on contention — the sort clause on `into=`, the enable-only clause on the fresh-folder lane, with the "sort it above any mod you later add" tail kept. `LoadOrderService.PlaceOne`'s summary (`LoadOrderService.cs:1573`) and `PlaceResult.CurrentWinner` (`:9766`) carry the same unlaned wording and should move with it.

Noted while folding review on #623, whose standing-instructions sentence now distinguishes the two lanes; this is the place response saying the opposite.

Avick3110:
`PlaceTools.EnableAndSort` picked its closing sentence on contention alone — whether any placed row had a current winner — and never on the lane. A default-lane placement creates a mod folder MO2 has not seen, which MO2 registers at the highest priority, so the placed copy out-ranks the previous winner the moment it is ticked; telling the caller to sort it above that winner was work they did not need to do. Both `housecarl_place` and `housecarl_nif_set` now branch on the lane as well: the fresh folder gets the enable-only clause (keeping the "sort it above any mod you later add" tail), and `into=`, whose folder already sits somewhere in the priority order, keeps the sort clause. `PlaceOutcome` and `NifSetResult` carry the lane (`RiderFolder.CreatedFresh`) so the render can see it, the per-row winner line and its json twin (`winner_note`) move with the closing sentence, and `PlaceOne`'s summary and `PlaceResult.CurrentWinner`'s doc comment say the laned thing.

Riding with it, from the backlog's W6 list: the "enable + sort" strings the 2026-09-07 ruling 4 had not reached. A patch houseCARL just created is new to MO2, so enabling it is the whole job — `WriteSentences.NewOrExtendedArtifact`'s new-patch arm, `create_plugin`, `merge_plugins`, the read-back caveat and the not-in-the-load-order refusal all now say "enable it in MO2". The forward row keeps the sort for an `into=` or in-place patch, which already has a place in the order. The published descriptions of `housecarl_place`, `housecarl_nif_set` and `housecarl_create_plugin` changed with them; the checked-in `tools-list-2.0.json` capture holds tool NAMES only, so no snapshot needed updating.

Verified: `dotnet build -c Release` clean, `dotnet test --filter "tier!=bridge"` 2006 passed / 0 failed, `housecarl-generator ci-all` 128/128 pass. Five new fixture cases in `PlaceRenderTests`: a fresh-folder placement over an existing provider asserts the enable-only sentence and the absence of the sort (text and json, closing sentence and per-row note); `into=` over an existing provider asserts the sort sentence; the no-contention arm is asserted unchanged on both lanes.

Closes #632


src/housecarl-mcp/PlaceTools.cs
```diff
@@ -335,7 +346,9 @@ void Withheld()
         // Name the destination folder rather than saying "the mod": the off-order line above can put a SECOND mod in
         // scope, and it ends by saying enabling THAT one is not required.
         sb.Append(r.CurrentWinner is not null
-            ? $"        currently wins the VFS: {r.CurrentWinner} — sort the new mod ABOVE it\n"
+            ? freshFolder
+                ? $"        currently wins the VFS: {r.CurrentWinner} — a folder MO2 has not seen registers at the highest priority, so '{modFolder ?? "(the new folder)"}' out-ranks it once enabled\n"
```
Avick3110: **A fresh folder does not out-rank MO2's `overwrite` layer.** `AssetResolver.BuildLooseRoots` orders the loose roots `overwrite > mods (priority) > Data`, and `CurrentWinner` is `DescribeSource(res.Sources[0])`, so it can legitimately be `overwrite (loose)`. A newly created mod folder registers at the highest *mod* priority, which is still below overwrite.

Scenario: a texture was previously written into MO2's overwrite folder (an xLODGen/DynDOLOD/BodySlide run, a recompiled script, an earlier in-place edit). `housecarl_place` on the default lane now prints `currently wins the VFS: overwrite (loose) — a folder MO2 has not seen registers at the highest priority, so 'houseCARL - MyFixes' out-ranks it once enabled`, and the trailer (line 316) says the copy wins `with no sorting`. Both are false: the overwrite copy keeps winning, and there is no sort that fixes it — the user has to clear or move the overwrite file. The pre-change sentence ("sort the new mod ABOVE it") at least prompted an action; this one tells the caller the job is done.

Same claim, same problem, at `JsonWire.cs:3070` (`winner_note`) and `NifTools.cs:871`. Suggest gating the fresh-lane arm on the winner not being `AssetResolver.OverwriteLayerName` — the constant already exists for exactly this distinction.

src/housecarl-mcp/NifTools.cs
```diff
@@ -864,6 +866,11 @@ public static string Render(NifSetResult d)
                 sb.Append("  TO MAKE IT WIN: nothing else provides this path — once '")
                   .Append(Path.GetFileName(d.OutputModFolder) is { Length: > 0 } f ? f : "the new folder")
                   .Append("' is enabled in MO2, the edited copy wins. 'Wrote it' is not 'it wins' until you do.\n");
+            else if (d.FreshFolder)
+                // A folder MO2 has not seen registers at the highest priority, so ticking it is the whole job.
+                sb.Append("  TO MAKE IT WIN: enable this folder in MO2 — MO2 registers a folder it has not seen at the highest priority, so it out-ranks ")
```
Avick3110, line 888: Same overwrite-layer hole as `housecarl_place` (see the comment on `PlaceTools.cs:350`). `d.CurrentWinner` here is `providers[0].Text` from the same resolver, whose loose precedence is `overwrite > mods > Data`. When the current winner is MO2's overwrite layer, "enable this folder in MO2 ... so it out-ranks X with no sorting" is wrong and unactionable — the edited mesh will still lose, and no left-pane sort can change that. The parenthetical "(loose beats BSA; among loose, the later mod wins)" is carried over from the sort arm and does not cover the overwrite case either.

Secondary, in the same block: the default-lane header two lines above still reads `wrote the verified mesh into a new mod folder:` unconditionally. Now that the lane is branched right below it, an `into=` call prints "a new mod folder" for a folder it did not create.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -335,7 +346,9 @@ void Withheld()
...
+                : $"        currently wins the VFS: {r.CurrentWinner} — sort '{modFolder ?? "(the patch folder)"}' ABOVE it\n"
```
Avick3110: **The `into=` arm can now name the winner as its own sort target.** Naming `modFolder` here (it used to say "the new mod") makes a self-referential instruction reachable: place into an *enabled* houseCARL patch folder that already provides the path. `PlaceOne` resolves `CurrentWinner` from the asset view before the write, so `res.Sources[0]` is that same folder and `DescribeSource` renders `houseCARL - MyFixes (loose)`, while `modFolder` is `houseCARL - MyFixes`. The row then reads `currently wins the VFS: houseCARL - MyFixes (loose) — sort 'houseCARL - MyFixes' ABOVE it`, and the trailer (line 319) adds `SORT it (left pane) ABOVE the current winner(s) listed above. Only then does the placed copy win.`

That is an impossible instruction for a case where the placement already wins on write — a re-place into an existing patch is the normal way to arrive here. Worth treating "the winner IS this folder" as a third arm alongside null/contended.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -233,7 +233,7 @@ static void AppendFullReadback(StringBuilder sb, IReadOnlyList<WritePatchBuilder
         sb.Append(dryRun
             ? "full preview — the ENTIRE record(s) as they WOULD be written, read from the in-memory would-be content (nothing is on disk):\n"
             : "full read-back — the ENTIRE record(s) as written, re-read from the patch file on disk " +
-              "(the written file's content, NOT load-order truth; the patch wins nothing until enabled + sorted in MO2):\n");
+              "(the written file's content, NOT load-order truth; the patch wins nothing until it is enabled in MO2):\n");
```
Avick3110: **This caveat is lane-agnostic, so de-laning it drops the sort on the one lane the PR says keeps it.** `AppendFullReadback` is called from four places (163, 216, 479, 952) and never learns the lane — it serves the fresh-patch lane, the `into=` extend, and the in-place lane alike.

Concrete: `housecarl_forward into="MyPatch.esp" readback=true` now renders, in one response, the per-record line `(out-ranks the current winner X once this patch is enabled and sorted above it)` (your new line 459) and then this header saying `the patch wins nothing until it is enabled in MO2`. The extend lane's folder already sits at a fixed priority, which is the whole premise of the change — so the header now understates the requirement on exactly that lane, and contradicts the line printed a few rows above it. Either pass the lane down here or leave this sentence saying enable + sort.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -32,7 +32,7 @@ public static class WriteTools
          "or a houseCARL folder of that name already exists, it REFUSES loud rather than rename or overwrite (Q3). Pass " +
          "esl=true for the lightest trigger (a header-only light plugin consumes no consequential load-order slot; with " +
          "zero records the ESL FormID-range rule is trivially satisfied). author/description are optional TES4 header " +
-         "text. Returns the plugin path + mod folder — enable + sort it in MO2 to use it. To author actual records, use " +
+         "text. Returns the plugin path + mod folder — enable it in MO2 to use it (a NEW folder registers at MO2's highest priority). To author actual records, use " +
```
Avick3110: The parenthetical imports the *asset* mechanism into a *plugin* claim, and they are different systems. Whether a created plugin's records win is decided by the plugin load order (MO2's right pane / `plugins.txt` + `loadorder.txt`), not by the mod folder's left-pane priority — the left-pane priority only decides which loose file wins the VFS. A new folder registering at the highest mod priority says nothing about where the .esp lands in the load order.

The conclusion ("enabling is usually the whole job", because MO2 appends a newly activated plugin at the end) may still hold, but the reason given is the wrong one, and it is the reason a reader will generalise from — e.g. to a `compact_plugin` output that keeps the source's basename, or to an ESL-flagged plugin. Same substitution at `WriteSentences.cs:117`, `WriteTools.cs:501` and `WriteTools.cs:685`. Suggest citing the load order rather than folder priority.

src/housecarl-mcp/WriteSentences.cs
```diff
@@ -114,7 +114,7 @@ internal static string NewOrExtendedArtifact(bool extended, string file, long by
         (extended ? $"extended {file} (existing patch grown; {bytes} bytes)\n"
                   : $"wrote {file} (new patch; {bytes} bytes)\n")
       + (extended ? $"mod folder: {modFolder}\n"
-                  : $"mod folder: {modFolder}  — enable + sort it in MO2 to use the patch\n");
+                  : $"mod folder: {modFolder}  — enable it in MO2 to use the patch\n");
```
Avick3110: Two shipped docs quote this exact string and were not updated, so the plugin now ships a transcript that no build produces:

- `plugin/codex/housecarl/SKILL.md:67` — `mod folder: SwordFix  — enable + sort it in MO2 to use the patch`
- `plugin/codex/housecarl/SKILL.md:68` — `... the patch wins nothing until enabled + sorted in MO2:` (the string changed at `WriteTools.cs:236`)

Also `README.md:126`: "The patch has no effect until it is enabled and sorted in MO2."

Avick3110:
All six findings were real and are folded in `13c9d689`, on top of a rebase onto current main. Build clean, 2017 tests pass, `ci-all` 128/128.

**1 + 2 — the overwrite layer.** Right, and the pre-change sentence really was the safer of the two. `AssetResolver.BuildLooseRoots` puts `overwrite` above every mod, so no fresh folder and no left-pane sort reaches a path it wins. `PlaceResult` now carries `WinnerIsOverwrite`, set in `PlaceOne` from `res.Sources[0]` against `AssetResolver.OverwriteLayerName` (kind loose), and `NifSetResult` carries the same flag from the same resolver. Both renders answer it apart from the lane: the row says "MO2's overwrite folder is ABOVE every mod, so no enable and no sort out-ranks it; move or delete the overwrite copy of this path", and `place`'s trailer appends the same fact when any placed row is won from overwrite. Overwrite rows are excluded from the contention count that picks the lane arm, so a batch whose only contention is overwrite no longer falls through to either "with no sorting" or "Nothing else provided these path(s)". `nif_set`'s default-lane header is branched too — an `into=` call now says "into the mod folder", not "into a new mod folder".

**3 — the self-referential sort.** Folded. `PlaceResult.WinnerIsDestination` compares the winning provider's name with the destination folder, and a row it is true of gets its own sentence ("'X' already provided this path — the placed copy replaces its own earlier copy and keeps winning") and is not counted as contention, so the trailer stops asking for a sort as well. The four arms now live in one `PlaceWire.WinnerLine`, which the json `winner_note` calls instead of keeping its own copy; `winner_is_overwrite` and `winner_is_destination` ride the json row as typed fields.

**4 — `AppendFullReadback` is lane-agnostic.** Folded rather than reverted: it takes `freshPatch` from each of its four callers (`!o.Extended && !o.InPlace`) and says "enabled" or "enabled and sorted" accordingly, so the header and the per-record lines above it now agree on every lane.

**5 — folder priority cited for a plugin claim.** Agreed, that was the wrong mechanism. The forward row no longer names a mechanism at all and is lane-neutral again: "(out-ranks the current winner X once this patch is enabled and loaded after it)". Where the reason is worth stating it now cites the load order — `NewOrExtendedArtifact`'s new-patch arm and the merge line say "MO2 adds a newly activated plugin at the END of the load order". `create_plugin`'s description drops the priority parenthetical for the fact that actually applies: it carries no records, so where it sits does not matter. The asset-lane descriptions (`place`, `nif_set`) still cite folder priority, which is the right mechanism there.

**6 — shipped docs.** Updated: both lines in `plugin/codex/housecarl/SKILL.md` and the readback sentence in `README.md:126`.

**Test helper default.** Fixed — `Outcome(...)` now defaults `fresh: true`, so the accounting and cap cases exercise the default lane. Three cases added for the new arms: an overwrite winner on each lane, and a re-place into a folder that already wins.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -1501,7 +1501,11 @@ public NifSetResult NifSet(string relPath, IReadOnlyList<NifSetOp> ops, string?
             }
 
             string? winner = providers.Count > 0 ? providers[0].Text : null;
-            return NifSetResult.OkNewFolder(rel, chosenProv, providers, place.Ambiguous, report, rf.ModFolder, winner, MergeWarnings(report.Warnings, warnings, null), profileName);
+            // MO2's overwrite folder is the TOP loose root, so no mod folder out-ranks it and no left-pane sort reaches it.
+            bool winnerIsOverwrite = place.Sources.Count > 0 && place.Sources[0].Kind == AssetKind.Loose
+                && string.Equals(place.Sources[0].ProviderName, AssetResolver.OverwriteLayerName, StringComparison.OrdinalIgnoreCase);
```
Avick3110: **nif_set never got the self-winner arm place did, so an `into=` re-edit is told to sort a folder above itself.**

`PlaceOne` learned `WinnerIsDestination` (LoadOrderService.cs ~1603) so a re-place into an already-winning houseCARL folder is not answered with a sort. `NifSet` computes only `winnerIsOverwrite` here, so the iterate loop `nif_set(rel, into="houseCARL - MyFixes")` -> tweak -> `nif_set(rel, into="houseCARL - MyFixes")` resolves `providers[0]` to that same enabled folder, `FreshFolder` is false, and `NifSetWire` falls to the last arm and prints `TO MAKE IT WIN: enable this folder in MO2 and sort it ABOVE houseCARL - MyFixes (loose)` — the folder the file was just written into. That is exactly the uninterpretable instruction this PR set out to remove, in the lane where re-editing the same mesh is the normal workflow.

The check is the same one `PlaceOne` already writes: compare `place.Sources[0].ProviderName` to `Path.GetFileName(rf.ModFolder)`, carry it on `NifSetResult`, and give `NifSetWire` the matching arm.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -334,9 +357,27 @@ void Withheld()
...
+        if (r.CurrentWinner is null)
+            return $"nothing else provides this path — once '{folder}' is enabled, the placed copy wins";
+        if (r.WinnerIsDestination)
+            return $"'{folder}' already provided this path — the placed copy replaces its own earlier copy and keeps winning";
+        if (r.WinnerIsOverwrite)
+            return $"currently wins the VFS: {r.CurrentWinner} — MO2's overwrite folder is ABOVE every mod, so no enable "
+                 + $"and no sort out-ranks it; move or delete the overwrite copy of this path, then '{folder}' wins";
+        return freshFolder
+            ? $"currently wins the VFS: {r.CurrentWinner} — a folder MO2 has not seen registers at the highest priority, so '{folder}' out-ranks it once enabled"
+            : $"currently wins the VFS: {r.CurrentWinner} — sort '{folder}' ABOVE it";
```
Avick3110, line 401: **The `into=` arm still asks for a sort against two winners no left-pane sort can reach: a BSA and the game's own `Data` layer.**

`AssetResolver.ResolveProviders` returns loose first, BSA after, with the loose roots ordered `overwrite > mods > Data`. So the winner is a BSA **only** when no loose copy exists, and is `Data (loose)` **only** when no mod provides the path. In both cases the placed loose copy beats it the moment the `into=` folder is enabled, whatever its priority — loose beats BSA, and every enabled mod beats `Data`. The render still prints `sort 'houseCARL - MyFixes' ABOVE Skyrim - Meshes0.bsa` / `ABOVE Data (loose)`, neither of which is a row in MO2's left pane.

This is the headline case for the tool: re-asserting a vanilla or BSA-packed asset as a loose override. The overwrite layer got its own arm here; its two siblings at the other end of the root list need the same treatment — for a BSA winner or `AssetResolver.DataLayerName`, the true sentence is the no-contention one ("once enabled, the placed copy wins"). The same fall-through exists in `NifSetWire`'s final arm (NifTools.cs ~882).

src/housecarl-mcp/WriteTools.cs
```diff
@@ -32,7 +32,7 @@ public static class WriteTools
          "or a houseCARL folder of that name already exists, it REFUSES loud rather than rename or overwrite (Q3). Pass " +
          "esl=true for the lightest trigger (a header-only light plugin consumes no consequential load-order slot; with " +
          "zero records the ESL FormID-range rule is trivially satisfied). author/description are optional TES4 header " +
-         "text. Returns the plugin path + mod folder — enable + sort it in MO2 to use it. To author actual records, use " +
+         "text. Returns the plugin path + mod folder — enable it in MO2 to use it (it carries no records, so where it sits in the load order does not matter). To author actual records, use " +
```
Avick3110: **"it carries no records, so where it sits in the load order does not matter" is false for the main reason `create_plugin` exists.**

A trigger plugin's whole job (the render two hundred lines down says so) is to make a basename present so a basename-bound `<stem>.bsa` loads or a FormID range is reserved. A BSA's precedence is its **owning plugin's** rank — this repo's own resolver encodes that: `AssetResolver.DedupeArchives` keeps the max `PluginRank` and `ResolveProviders` orders BSA sources `OrderByDescending(b => b.rank)`. So a trigger plugin sitting low in the order makes its archive lose to another mod's archive, and where it sits decides which assets the user gets. Same for the FormID range: the load index is the range.

The old "enable + sort" was over-strong, but the replacement claims the opposite absolute. Something like "enabling it is enough for the name to exist; if it is there to load a BSA, its position still decides that archive's precedence" keeps both halves true.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -289,32 +291,53 @@ internal static string LeftoverNote(string leftoverFolder)
...
+            ? (o.FreshFolder
+                ? ". MO2 registers a folder it has not seen at the highest priority, so once enabled the placed copy " +
+                  "out-ranks the current winner(s) with no sorting (sort it above any mod you later add that also provides these path(s))."
+                : shownContended
+                    ? " and SORT it (left pane) ABOVE the current winner(s) listed above. Only then does the placed copy win."
+                    : " and SORT it (left pane) ABOVE the current winner(s) — max_chars cut the row(s) naming them from " +
+                      "this render, so raise max_chars and re-read to see which. Only then does the placed copy win.")
+            : anyOverwrite
+                ? "."
                 : ". Nothing else provided these path(s), so once enabled the placed copy wins (sort it above any mod you later add that also provides them).";
```
Avick3110: **A batch whose every placed row is won by the destination folder closes on the no-contention sentence, which contradicts the rows above it.**

`WinnerIsDestination` rows `continue` without setting either flag, so a re-place into an enabled houseCARL patch (every row `WinnerIsDestination`) leaves `anyContended` and `anyOverwrite` false and the trailer reads `Enable the mod 'houseCARL - MyFixes' in MO2. Nothing else provided these path(s), so once enabled the placed copy wins.` — one line under rows that each just said `'houseCARL - MyFixes' already provided this path`. The folder is already enabled (that is how it is winning the VFS), so the one action the closing sentence names is the one action there is nothing left to do.

The skipped case wants its own arm, the way overwrite got one: the placement replaced the folder's own earlier copy and already wins, nothing to enable or sort.

src/housecarl-mcp/PlaceTools.cs
```diff
@@ -289,32 +291,53 @@ internal static string LeftoverNote(string leftoverFolder)
...
+        if (anyOverwrite)
+            sort += " MO2's overwrite folder sits ABOVE every mod in the VFS, so for the path(s) it currently wins " +
+                    "neither enabling nor sorting is enough — move or delete the overwrite copy of each path whose " +
```
Avick3110: **The overwrite remedy points at rows max_chars may have cut.**

"each path whose winner reads 'overwrite (loose)' above" only works if those rows reached the page. The contended arm two branches up handles exactly this hazard — it swaps "listed above" for "max_chars cut the row(s) naming them ... raise max_chars and re-read" when `!shownContended` — but `anyOverwrite` is collected without any `i < rendered` test, so a 200-destination batch whose overwrite-won rows sit past the cut tells the caller to go move copies of paths this response never names, with no hint that raising max_chars would reveal them. Track a `shownOverwrite` alongside `shownContended` and give the cut case the same re-read remedy.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -225,15 +225,19 @@ static string RenderDryRun(WritePatchBuilder.PatchOutcome o, int maxChars, bool
     /// nothing until enabled in MO2). Char-budget-bounded with an explicit notice, at the lower
     /// <see cref="Wire.ReadbackMaxChars"/> default so the cut-off output stays under the host token ceiling and the
     /// truncation note reaches the caller.</summary>
+    /// <summary><paramref name="freshPatch"/> is the LANE: a patch this call created is new to the load order, so
```
Avick3110: Second `<summary>` on the same member: `AppendFullReadback` now carries two summary elements, so one of them is dropped from the generated XML / IntelliSense and which one survives is not this file's choice. It reads as a param note — make it `<param name="freshPatch">...</param>` (or fold it into the existing summary).

Avick3110: Fixed - it is a `<param name=freshPatch>` now, and the surviving `<summary>` points at the parameter for the lane rather than restating the caveat.

src/housecarl-mcp/WriteTools.cs
Avick3110: Agreed - I over-corrected into the opposite absolute. The description now says: enable it in MO2 to make the basename exist; if it is there to load a `<stem>.bsa` or to reserve a FormID range, its POSITION in the load order still decides that archive's precedence and that range. Thanks for the `DedupeArchives` / `ResolveProviders` pointer - the rank ordering is exactly the thing that makes the claim false.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in `0977b320`. `NifSet` now computes `winnerIsDestination` the same way `PlaceOne` does — `place.Sources[0].ProviderName` against `Path.GetFileName(rf.ModFolder)` — and carries it on `NifSetResult`. `NifSetWire` gets the matching arm ahead of the others: "ALREADY WINS: 'X' already provided this mesh and is enabled, so the edited copy replaces its own earlier copy and keeps winning — nothing to enable or sort." You are right that this is the lane where re-editing the same mesh is the normal loop.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in `0977b320`, and this was the worst of the six — it is the tool's headline case. `PlaceResult.WinnerLosesOnEnable` is set when the winning source is a BSA or `AssetResolver.DataLayerName`; `NifSetResult` carries the same flag. Both renders answer it before the lane branch, so neither lane asks for a sort: the row says "a loose file in an enabled mod beats it at any priority, so once 'X' is enabled the placed copy wins", and the closer says "Once enabled the placed copy wins — a loose file beats a BSA, and any enabled mod beats the game's Data folder". Deliberately not the literal no-contention sentence: "Nothing else provided these path(s)" would be false when a BSA does provide them. A theory case covers both lanes.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in `0977b320`. The trailer counts placed rows and destination-won rows, and when every placed row is destination-won it returns its own sentence instead of the enable opener: "the placed file(s) went into 'X', which already provided these path(s) and already wins the VFS for them — it is enabled, so there is nothing to enable or sort". The existing self-winner test now also asserts the closer carries neither the enable opener nor the no-contention sentence.

src/housecarl-mcp/PlaceTools.cs
Avick3110: Folded in `0977b320`. `shownOverwrite` is tracked beside `shownContended` with the same `i < rendered` test, and the remedy takes the cut wording when it is false: "move or delete the overwrite copy of the path(s) it wins — max_chars cut the row(s) naming them from this render, so raise max_chars and re-read to see which." Covered by a fixture that puts the overwrite row last in a 40-row batch and caps the render below it.

PR #759, opened 2026-09-14
Bound the list-lane aggregate and name empty requested types

### LINKED ISSUE #631: [Bug] records aggregate on the formids= lane ignores max_chars
`housecarl_records` with `project={form:'aggregate', group_by:…}` renders every group row on the `formids=` list lane, whatever `max_chars=` says, and says nothing about it.

The scan lane is fine: `RenderCrossQueryGroups` (`ReadTools.cs:435`) budgets its group rows against the cap and emits a truncation notice. The list lane is not: `RecordsTools.cs:682` returns `RenderListAggregate(...)`, and that method (`RecordsTools.cs:2316`) takes no cap parameter at all — it appends a row per group and returns without `RenderCap.Settle`, unlike every sibling render on the same lane.

So `formids=[…500 ids…]` with `group_by='defined_in'` and `max_chars=2000` can return hundreds of rows well past the ceiling, with no notice. Every other render on the lane holds the ceiling, so a caller sizing a response has no reason to expect this one not to.

Fix: pass the cap into `RenderListAggregate` and settle it the way the scan lane does, so the render is bounded and a truncation is stated.

Found while folding review on #623: the `project=` description now names the lane the bound holds on, which is true but is a description standing in for a render that should hold it.

### LINKED ISSUE #689: [GAP] records aggregate group_by=type emits no row for a requested type with zero records, so absence has to be inferred from a missing group
Found by both 2026-09-09 re-gate mutagen-reference arms and re-verified by the judge. `housecarl_records` with `types=[PARW, PBAR, PBEA, PCON, PFLA, PMIS, ...]` and `project={form: aggregate, group_by: type}` returns rows only for types that have records; a requested type with none is simply absent from the table. The caller has to diff the request against the response to learn that PARW is empty.

When a type is named in `types=`, the aggregate should carry a `0` row for it. Mitigated only by an unknown type name being refused by name, which is a different question from "known type, no records".

Avick3110:
Two fixes to the `housecarl_records` aggregate count table.

The `formids=` list lane's render appended a row per group and returned, taking no cap and never settling against `max_chars`, unlike every sibling render on that lane — so `formids=[…500 ids…]` with `group_by='defined_in'` and `max_chars=2000` could return hundreds of rows past the ceiling with no notice. It now charges its cut notice and its accounting line before the first group row, stops whole rows at the budget, names how many of its groups it laid, and settles against the cap the way the scan lane's `RenderCrossQueryGroups` does. The counts are unchanged — only the rendering is cut, so the table stays exact. The json twin had the same hole and gains `rendered` and `truncated` beside `groups`, which is the shape the scan lane's json count table already has.

Second, `group_by=type` dropped a type named in `types=` out of the table entirely when it matched nothing, so "PARW is empty" had to be read by diffing the request against the response. Every requested type is now seated at zero before a match is counted, on the in-order scan, the off-order scan, and the list lane (whose one shape that takes `types=` is the typed MGEF carrier walk, where `types=` narrows the carrier types). An unknown type name is still refused by name, which is a different answer from a known type with no records.

No tool description changed, so the tools-list snapshot is untouched.

Tests: a list-lane aggregate over one id per record type, at a cap derived from its own uncapped length, asserts fewer rows and the truncation notice; a `types=` request naming an empty type asserts the `0` row on the scan lane (`AMMO`) and on the list lane (`SCRL` carriers). Verified in the worktree: `dotnet build -c Release` clean, `dotnet test --filter "tier!=bridge"` 2004/2004 passed, `ci-all` 128/128 ALL PASS.

Closes #631
Closes #689


src/housecarl-mcp/JsonWire.cs
```diff
@@ -639,14 +643,20 @@ public static string RenderListAggregate(string groupBy, IReadOnlyList<KeyValueP
...
+                if (ms.Length >= cap) { truncated = true; break; }
                 w.WriteStartObject();
                 w.WriteString("key", key);
                 w.WriteNumber("count", n);
                 w.WriteEndObject();
+                rendered++;
             }
             w.WriteEndArray();
+            w.WriteNumber("rendered", rendered);
+            w.WriteBoolean("truncated", truncated);
```
Avick3110, line 686: The json twin cuts group rows but never states how many groups there were, so `truncated: true` is the whole story a json caller gets. The text twin says "rendered R of N groups", which is what lets a caller size the retry; here `count` is the record count, not the group count, so with 40 groups and `max_chars=300` the document carries `"rendered":6, "truncated":true` and the 40 appears nowhere — the caller can only guess how much to raise `max_chars` by. Writing the group total beside `rendered` (e.g. `w.WriteNumber("groups_total", rows.Count);`) would close the gap the text notice already covers.

Avick3110: Folded in 81c8eb0f. Both json count tables now write `groups_total` just before the `groups` array — the list lane's off `rows.Count`, the scan lane's off `q.Groups.Count` — so a cut document says what it was cut from and a json caller can size the retry the way the text twin's "rendered N of M groups" lets it. The scan lane's `total` is the MATCH count, not the group count, so it was short of this too; folding only the list lane would have left the two disagreeing. New test `ListLaneAggregateInJson_SaysHowManyGroupsItWasCutFrom` asserts rendered == groups_total and truncated false uncapped, then truncated true with groups_total unchanged and rendered short of it at half the cap. Rebased on origin/main, 2013/2013 tests and ci-all 128/128 pass.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -2607,18 +2620,33 @@ static string RenderListAggregate(IReadOnlyList<ReadOutcome> outcomes, string gr
         }
         var rows = groups.OrderByDescending(g => g.Value).ThenBy(g => g.Key, StringComparer.Ordinal).ToList();
```
Avick3110: The two halves of this PR fight each other here: a seeded type has count 0, and the sort is `OrderByDescending(count)`, so every newly seated empty type lands at the bottom of `rows` — exactly the rows the new max_chars cut takes first. `types=[...30 types...] project={form:'aggregate', group_by:'type'}` at the default cap renders the types that HAVE records and cuts the `0` rows, so the fact this change exists to state is the first thing dropped. The notice does say N of M were rendered, so it is not silent — but the caller who asked "which of these types are empty?" gets a table that answers everything except that. Same on the scan lane (ReadTools.RenderCrossQueryGroups) and in the json twin, whose group array is cut in the same order. Worth considering seating the zero rows ahead of the cut (or naming how many zero rows were held back in the notice) so the empty-type answer survives a capped table.

Avick3110: Folded in 187b887a — real, and the two halves did fight. A zero can only come from seeding, so that is the discriminator: the counted table now carries groups with a count above zero, and the requested-but-empty types are stated apart from it. Text: one `no records: A, B, C` line after the counted rows, charged with the truncation notice and the accounting line BEFORE the first counted row, so the cap takes a counted row before it takes the empty answer; the line has its own bound (half the cap, floor 120) and past it says ", and N more" rather than growing without limit. Json: an `empty_groups` array written ahead of the `groups` array, which the per-row cap check cannot reach. All four renders — scan text, list text, and both json twins. `groups_total` is now the counted groups, which is what `rendered` is short of. Three tests: the empty type survives a cut at half the uncapped length on both lanes, and the json `groups` array carries no zero-count entry. Rebased on origin/main, 2023/2023 tests and ci-all 128/128 pass.

PR #760, opened 2026-09-14
Say where the merged plugin has to load

### LINKED ISSUE #718: [GAP] merge never says where the merged plugin has to sit, though it already knows
A merge tells the modder to "enable + sort it in MO2" and stops. It never says where the new plugin needs to sit, even though the merge itself has just computed everything needed to say so.

## What a real user did

He asked his agent to merge nine Mostly Treeless Tundra patches to reclaim ESL slots, and in the same breath asked for "the optimal placement for load order since they're spread out among the load order".

The merge finished. Then, with no cheap way to answer the placement half, the agent went looking for it the only way the surface allows:

```
housecarl_records
  plugins: 10 plugins
  types:   ["REFR"]
  project: {"form":"tree","fields":["FormVersion"]}
  to_file: ...\tree.jsonl
```

Every placed reference in ten plugins with its full conflict stack. He interrupted it at twenty minutes.

## What the merge already knows

By the time it writes, the merge has:

- each donor's index in the load order (it sorts donors by load order before merging — `orderIndex`),
- the cross-donor conflicts and which donor won each,
- every external plugin that references a donor record, overrides a donor record, or declares a donor as a master (it already reports all three as warnings),
- the merged plugin's master list.

That is enough to state the constraint: the merged plugin must load after its masters, and to preserve the conflict outcomes the merge just baked in, it must sit at or after the last donor's position — with the plugins between the first and last donor that touch the same records named, because those are the ones whose relative position changes meaning.

Roughly:

> placement: the donors sat at positions 412 through 2,180. Load this after `<last master>`; anything below position 2,180 that also touches these records will now win over content that used to beat it — 3 plugins do: A.esp, B.esp, C.esp.

Derived from data already in hand, at no extra scan cost.

## Why it matters

Merging scattered patches is a normal job, and "where does the result go" is not an afterthought — get it wrong and the merge quietly changes which mod wins. Right now the tool performs the dangerous half and leaves the modder to reason out the safe half by hand, or watch an agent spend twenty minutes brute-forcing it.

## Related

- #716 — the tree projection that agent reached for commits to unbounded work with no estimate. Fixing this issue removes the reason it reached for it at all.
- #694 — `info_order` has no off-order lane, so a patch's merged position cannot be read before it is enabled. Same question from the other end.

Avick3110:
A merge told the user to enable and sort the output in MO2 and stopped there, though by that point it had computed each donor's index in the load order and the output's master list. The report now carries a placement paragraph derived from exactly that — no extra scan, no new parameter.

It gives the load-order positions the donors sat at, tells the caller to load the merged plugin after its last master (named, with its position) and AT the last donor's position, and says why both directions are wrong: the merge resolved the donors' conflicts as they stood there, so an earlier slot lets content the donors used to beat win over the merge, and a later one puts the merge over plugins that used to beat them. It then says what that position governs — the donors' OVERRIDES, kept at their masters' FormIDs — and what it does not: the donors' own records are renumbered into the output's FormID space, where nothing outside the merge shares an id with them, so a plugin that overrode one is orphaned by the swap (the report's existing warning names it) rather than outranked by where the output sits. A master the order carries BELOW the last donor is said plainly, because the output cannot then be both after it and where the donors were. A single-donor rename gets the same paragraph with its one position.

It names no other plugins. The first cut listed the plugins between the first and last donor that the identify pass had found touching a donor record, but review showed that claim is false for exactly those plugins: the pass targets the records the donors originate, and every one of them is renumbered away. The records whose winner the position really decides are the donors' overrides at their masters' FormIDs, which a merge never enumerates — so the constraint is stated and the roster is not guessed at.

Fixture arms: donors at positions 2 and 5 of a five-plugin order give the range and the slot; the paragraph does not name the plugin between them that overrides a donor record, while the overrider warning still does; the last master and its position are named; and a rename claims nothing about an interval. Two unit arms on Derive pin the 0-based-in/1-based-out contract and the master-below-the-last-donor flag.

`dotnet build`, `dotnet test --filter "tier!=bridge"` (2,030 passed) and `ci-all` (128/128) all green.

Stacked on #749 (#715), which is stacked on #742 (#729); retarget once those merge.

Closes #718


src/housecarl-mcp/WriteTools.cs
```diff
@@ -587,6 +587,36 @@ static void AppendUnscannablePlugins(StringBuilder sb, IReadOnlyList<RemapEngine
...
+        sb.Append("placement: ");
+        if (p.FirstPosition == p.LastPosition)
+            sb.Append("the donor sat at load-order position ").Append(p.LastPosition).Append(". ");
+        else
+            sb.Append("the donors sat at load-order positions ").Append(p.FirstPosition).Append('–').Append(p.LastPosition)
+              .Append(" (").Append(p.FirstDonor).Append(" … ").Append(p.LastDonor).Append("). ");
+        sb.Append("Load ").Append(o.OutputName);
+        if (p.LastMaster is not null)
+            sb.Append(" after its last master ").Append(p.LastMaster).Append(" (position ").Append(p.LastMasterPosition).Append(')').Append(", and");
+        sb.Append(" at or after position ").Append(p.LastPosition)
```
Avick3110: "at or after position N" is unsound on the *after* half, and the paragraph deliberately omits the plugins that make it unsound.

Only `at` position N preserves the resolution the merge baked in. Placing the output later flips conflicts against every plugin after the last donor that overrides a donor record: donor B at position 30, plugin Y at 40 overrides one of B's records and beat B before the merge; drop the merged plugin at position 45 on this advice and the merge now beats Y — a change in winner the paragraph just told the user is safe.

`Derive` filters `Intervening` to `first.Order < p < last.Order` (MergeLoadPosition.cs:43), so the plugins that break this half of the advice are precisely the ones never named here. Either say "at position N" (the position that reproduces the old outcome), or keep "at or after" and name the after-the-last-donor touchers too.

src/housecarl-core/MergeLoadPosition.cs
```diff
@@ -0,0 +1,51 @@
...
+    string? LastMaster, int LastMasterPosition, IReadOnlyList<string> Intervening);
+
+/// <summary>The merged plugin's placement constraint. It must load after its masters, and at or after the LAST donor's
+/// position: the merge baked the donors' conflict outcomes in at that position, and anything that used to lose to a
+/// donor there would win over the merge if the merge sat earlier. The plugins BETWEEN the first and last donor that
+/// also touch donor records are named, because those are the ones whose position relative to the merge changes which
+/// version wins.</summary>
+public static class MergeLoadPosition
+{
+    /// <summary>Positions are 1-based places in the ACTIVE load order. <paramref name="touching"/> is every plugin
```
Avick3110: The doc says the positions are 1-based, but the parameters are 0-based: the sole caller passes `d.Order` and `orderIndex[...]`, both raw indices into `resolver.PluginNames`, and `Derive` converts with `+ 1` on the way out (`first.Order + 1`, `lastMasterPos + 1`).

So this sentence describes the *return* values while sitting on the *parameter* contract. A second caller that reads it literally and hands in already-1-based positions gets every printed position one too high, and silently — the comparisons against `first.Order` / `last.Order` stay self-consistent, so nothing throws and no test catches it. Worth splitting: inputs are 0-based indices into the active order, `MergeSiting`'s fields are 1-based.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -587,6 +587,36 @@ static void AppendUnscannablePlugins(StringBuilder sb, IReadOnlyList<RemapEngine
...
     // never start. Reported, never silent. inPlace is always false for merge — it has no in-place lane.
 
+    /// <summary>Where the merged plugin has to load, from the positions, masters and dependents the merge already
+    /// computed — the question a caller otherwise answers by scanning the order by hand (#718).</summary>
+    static void AppendPlacement(StringBuilder sb, WritePatchBuilder.MergeOutcome o)
+    {
+        if (o.Placement is not { } p) return;
+        sb.Append("placement: ");
+        if (p.FirstPosition == p.LastPosition)
+            sb.Append("the donor sat at load-order position ").Append(p.LastPosition).Append(". ");
```
Avick3110, line 661: The single-donor arm takes the `FirstPosition == p.LastPosition` branch here and then still falls through to the shared "no plugin between the first and last donor also touches these records" line — which is vacuously true (there is no interval) but reads as a finding.

For a rename the reader is told, in the same breath as "the donor sat at position 5", that nothing touches these records, while plugins at positions 4 and 6 may well override them; the range filter in `Derive` (`p > first.Order && p < last.Order`, empty when first == last) guarantees the sentence is printed no matter what the order holds. `ASingleDonorGetsItsOwnPositionAndNoInterveningList` asserts the current wording, so it locks the misleading sentence in. Suggest the single-donor arm say nothing about intervening plugins, or say "a rename has no interval — see the external-overrider warnings above".

src/housecarl-mcp/WriteTools.cs
```diff
@@ -587,6 +587,36 @@ static void AppendUnscannablePlugins(StringBuilder sb, IReadOnlyList<RemapEngine
...
+            sb.Append("the donors sat at load-order positions ").Append(p.FirstPosition).Append('–').Append(p.LastPosition)
+              .Append(" (").Append(p.FirstDonor).Append(" … ").Append(p.LastDonor).Append("). ");
+        sb.Append("Load ").Append(o.OutputName);
+        if (p.LastMaster is not null)
+            sb.Append(" after its last master ").Append(p.LastMaster).Append(" (position ").Append(p.LastMasterPosition).Append(')').Append(", and");
+        sb.Append(" at or after position ").Append(p.LastPosition)
+          .Append(" — the merge resolved the donors' conflicts as they stood there, so anything earlier lets content the ")
+          .Append("donors used to beat win over the merge instead.\n");
+        if (p.Intervening.Count == 0)
+            sb.Append("  no plugin between the first and last donor also touches these records.\n");
```
Avick3110: The intervening set is scoped to records **originating** in the donors, not the records the merge carries, so this negative sentence can be flatly false in the most common merge shape.

`Intervening` comes from `id.ExternalOverriders` / `id.ExternalPlugins`, and that identify pass runs with `targets = plan.Dict.Keys` (LoadOrderService.cs:7472). `BuildMergeRemap` enters only each donor's **originating** FormKeys into the dict (RemapEngine.cs:648-659) — override records the donors carry at their *masters'* keys (a donor overriding a `Skyrim.esm` NPC, say) are never in `targets`.

Concrete case, and it is the normal one for a patch family: donors A (pos 10) and B (pos 30) both override `Skyrim.esm` NPC records; plugin X at pos 20 overrides the same NPC records and originates nothing of A's or B's. X is not an external referencer or overrider of any *donor-originating* key, so it never reaches `Derive`, and the report prints "no plugin between the first and last donor also touches these records" — while X is exactly the plugin whose position relative to the merge decides the winner. The paragraph is confidently wrong about the case it exists to answer.

Either widen the scope (the carried-override keys are known to `MergeBuild`) or narrow the sentence so it claims only what was checked ("records the donors originate").

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -7541,12 +7541,19 @@ WritePatchBuilder.MergeOutcome FailAfterFolder(string msg)
                   "auto-starting before; they will now."
                 : null;
 
+            // Where the output has to sit, off the positions, masters and dependents this call already computed.
+            var placement = MergeLoadPosition.Derive(
+                donorInfos.Select(d => (d.Name, d.Order)).ToList(), build.Masters,
+                id.ExternalOverriders.Concat(id.ExternalPlugins),
```
Avick3110: `id.ExternalOverriders` / `id.ExternalPlugins` are a *scanned-through* set, not a complete one, so the "no plugin between the first and last donor also touches these records" sentence asserts a negative the pass never established.

`IdentifyExternalReferencers` skips a plugin outright when `view.ExcludedPlugins.ContainsKey(plugin)` (RemapEngine.cs:151), and a plugin whose header or record enumeration faults lands in `UnscannablePlugins` having contributed to neither list. Those plugins still occupy positions in `resolver.PluginNames`, so one can sit squarely between the donors.

Scenario: donors at positions 10 and 30; plugin X at 20 was excluded at index build (unparseable) and overrides a donor record. X is invisible to `Derive`, `Intervening` comes back empty, and the report says no plugin in the range touches these records — a silently wrong answer about a range the merge could not see into. `id.UnscannablePlugins` and the excluded set are both to hand here; the sentence should degrade to "none found, and N plugin(s) in that range could not be read" rather than claim none exist.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 970676db, by narrowing the claim rather than widening the scan. You are right that the identify pass targets only the donors' originating keys, so a plugin contending over an override a donor carries at its master's FormID never reaches Derive. Widening would need a second whole-order walk with a different target set (the existing one has to stay scoped to the records that actually move, since it feeds the referencer/overrider warnings), and the issue's premise is no extra scan. So the sentences now say what was checked: '... references or overrides a record the donors ORIGINATE', and the negative adds 'overrides the donors carry at their masters' FormIDs are outside what this pass looked for'. A test pins both halves.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded in 970676db. MergeSiting now carries UnreadBetween, counted from the excluded set and the identify pass's UnscannablePlugins, and the negative degrades to '... could not be read, so they were not checked' when any of them sit in the range; the positive list carries the same count as a trailing note. A unit test on Derive pins the count and that a plugin past the last donor is not in the range.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 970676db — the paragraph now says 'at position N, where the last donor sat' and gives both failure directions: an earlier slot lets content the donors used to beat win over the merge, a later one puts the merge over plugins that used to beat the donors, pointing at the external-overrider warning that names those. I did not add the after-the-last-donor touchers to the list, because the list is about the interval the merge spans; the constraint is now a single position, so they are a reason, not a roster.

src/housecarl-core/MergeLoadPosition.cs
Avick3110: Folded in 970676db. Derive's doc now says the inputs are 0-based indices into the active order and every position on MergeSiting is 1-based, and the new unit test passes 0-based positions and asserts the 1-based output.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 970676db. The rename arm no longer prints the interval sentence at all: it says a rename has no interval and points at the referencer and overrider warnings below, and the test now asserts the vacuous line is absent.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -587,6 +587,49 @@ static void AppendUnscannablePlugins(StringBuilder sb, IReadOnlyList<RemapEngine
...
+              .Append("below name what contends over these records.\n");
+        else if (p.Intervening.Count == 0)
+            sb.Append("  no plugin between the first and last donor references or overrides a record the donors ORIGINATE")
+              .Append(p.UnreadBetween > 0 ? " — though " + p.UnreadBetween + " plugin(s) in that range could not be read, so they were not checked" : "")
+              .Append("; overrides the donors carry at their masters' FormIDs are outside what this pass looked for.\n");
+        else
+        {
+            sb.Append("  ").Append(p.Intervening.Count)
+              .Append(p.Intervening.Count == 1 ? " plugin between the first and last donor also references or overrides" : " plugins between the first and last donor also reference or override")
+              .Append(" a record the donors ORIGINATE, so where the merge sits relative to each one decides which version wins: ")
```
Avick3110: The populated arm claims the wrong thing about the plugins it names, and drops the caveat the empty arm carries.

`Intervening` is built from `id.ExternalOverriders` + `id.ExternalPlugins`, whose `targets` are `plan.Dict.Keys` — the donor-space keys. Every one of those keys is renumbered into the OUTPUT's ModKey by the merge, so the merged copy of a donor-originated record is a brand-new FormID that nothing outside the merge can conflict with. Concretely, with the fixture's own shape: HcPlMiddle.esp overrides HcPlEarly.esp's weapon at `800:HcPlEarly.esp`; after the merge that record lives in the output at a fresh id in output space, and HcPlMiddle.esp still overrides the donor's copy. Moving the output above or below HcPlMiddle.esp changes no winner at all — before the swap the donor+Middle pair resolve exactly as they did, and after the swap Middle's override is orphaned (which is what the existing external-overrider WARNING already says), not outranked.

The records whose winner placement actually decides are the overrides the donors carry at their MASTERS' FormIDs — the ones the report counts on the "override kept at its master FormID" line. The empty arm says exactly that ("overrides the donors carry at their masters' FormIDs are outside what this pass looked for"), but this arm asserts "where the merge sits relative to each one decides which version wins" with no caveat, so the one path that names plugins is the one that states something untrue about them.

src/housecarl-mcp/WriteTools.cs
```diff
@@ -587,6 +587,49 @@ static void AppendUnscannablePlugins(StringBuilder sb, IReadOnlyList<RemapEngine
...
+        sb.Append("Load ").Append(o.OutputName);
+        if (p.LastMaster is not null)
+            sb.Append(" after its last master ").Append(p.LastMaster).Append(" (position ").Append(p.LastMasterPosition).Append("), and");
+        // AT that position, not "at or after": the merge resolved the donors' conflicts as they stood there, and moving
+        // the output either way changes a winner — earlier over the donors' own losers, later over the plugins past the
+        // last donor that override the same records, which this paragraph does not name.
+        sb.Append(" at position ").Append(p.LastPosition).Append(", where the last donor sat")
+          .Append(" — the merge resolved the donors' conflicts as they stood there, so an earlier slot lets content the ")
+          .Append("donors used to beat win over the merge, and a later one puts the merge over plugins that used to beat ")
+          .Append("the donors (the external-overrider warning below names those).\n");
```
Avick3110: "(the external-overrider warning below names those)" points at a block that is often not there. The overrider section in RenderMerge is emitted only under `if (o.ExternalOverriders.Count > 0)` — unlike the referencer section it has no `else` "none" line. So on the common clean merge (no external overriders) the placement paragraph, which is appended unconditionally, sends the reader to a warning that does not exist in the report. Either gate the parenthetical on `o.ExternalOverriders.Count > 0` or give the overrider section a "none" line the way referencers have one.

src/housecarl-core/MergeLoadPosition.cs
```diff
@@ -0,0 +1,71 @@
...
+        between.Sort((a, b) => a.Order.CompareTo(b.Order));
+
+        // A plugin in the range the pass could not read is not evidence of absence, so it is counted and said.
+        int unread = 0;
+        var seenUnread = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        foreach (var u in unreadable)
+        {
+            if (donorNames.Contains(u) || !seenUnread.Add(u)) continue;
+            var p = positionOf(u);
+            if (p is not null && p.Value > first.Order && p.Value < last.Order) unread++;
```
Avick3110: A plugin can land in BOTH `Intervening` and `UnreadBetween`, so the report names it and then counts it again as unseen. In `IdentifyExternalReferencers`, the header-read `catch (Exception)` arm sets `headerFaulted = true`, adds an `UnscannablePlugin(EnumerationFault)` entry, and deliberately FALLS THROUGH to the record walk ("the record stream is a different read that may well succeed"). If that walk then finds a ref or an override, the same plugin is in `ExternalPlugins`/`ExternalOverriders` and in `UnscannablePlugins`. For a plugin between the donors the render prints "1 plugin between the first and last donor also references or overrides … : X.esp (1 more plugin(s) in that range could not be read)" — where the "1 more" IS X.esp. Subtract the names already in `between` from the unread tally.

src/housecarl-core/MergeLoadPosition.cs
```diff
@@ -0,0 +1,71 @@
...
+        var first = donorsByLoadOrder[0];
+        var last = donorsByLoadOrder[^1];
+
+        string? lastMaster = null;
+        int lastMasterPos = -1;
+        foreach (var m in masters)
+        {
+            var p = positionOf(m);
+            if (p is null) continue;                                       // a master the order does not carry is the pre-flight's refusal, not this sentence's business
+            if (lastMaster is null || p.Value > lastMasterPos) { lastMaster = m; lastMasterPos = p.Value; }
```
Avick3110, line 42: Nothing checks that the last master sits before the last donor, and when it does not the paragraph gives two instructions that cannot both be followed. `build.Masters` is read back from the written header, so it can include an esp-as-master (a plugin listed as a master without the ESM flag) — MO2 forces flagged masters to the top of the order but not those, so a donor can declare one that sits AFTER the last donor in the user's actual order. Derive then reports `LastMasterPosition > LastPosition` and the render prints "Load M.esp after its last master Foo.esp (position 12), and at position 5, where the last donor sat". Worth either clamping the advised slot to `max(lastMasterPos, lastPosition)` or saying plainly that the order already has the master out of place.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 9d1ea934, by dropping the roster rather than caveating it. You are right that every key the identify pass targets is renumbered into the output's own ModKey, so those plugins are orphaned by the swap — which the existing warning says — not outranked by where the output sits; naming them under a position claim was the one untrue sentence in the paragraph. The intervening list and its unread counter are gone. The paragraph now states the constraint and says what the position decides (the donors' overrides at their masters' FormIDs) and what it does not (their own records, renumbered out of reach), with no roster, because a merge never enumerates the records that are position-sensitive. A test asserts the paragraph does not name the fixture's overriding plugin while the warning still does.

src/housecarl-mcp/WriteTools.cs
Avick3110: Folded in 9d1ea934. The pointer at the warning block is now emitted only when there is something to point at (ExternalOverriders or ExternalPlugins non-empty); on a clean merge the sentence ends at the claim.

src/housecarl-core/MergeLoadPosition.cs
Avick3110: Folded in 9d1ea934 — the double count is gone with the counter itself, since the intervening list it qualified no longer exists. The fall-through you describe is real and would have produced exactly that line.

src/housecarl-core/MergeLoadPosition.cs
Avick3110: Folded in 9d1ea934. MergeSiting carries MasterAfterLastDonor, and when the last master sits below the last donor the paragraph says so and stops there — 'after its last master M.esp, which already sits at position 12, AFTER the last donor — so it cannot also sit where the donors did, and the donors' conflict outcomes will not all survive the move' — instead of printing two instructions that cannot both be followed. I said it rather than clamping silently, since the order itself is the thing to look at. A unit arm pins it.

PR #761, opened 2026-09-14
Gather the reverse lane's bodies by plugin, not per match

### LINKED ISSUE #727: [GAP] the reverse lane costs ~50 ms per match after the index; an unbounded references= on a common target is a two-minute call

Avick3110:
The reverse lane's cost was never the index — it was what each match cost after it. Both spellings (an unbounded `references=` and a reverse `walk=`) take the candidates the reverse-reference index names and read each candidate's winner body to judge it, and that read went through `GetRecord`, which enumerates the winner plugin from the top to find one record. On a real order the candidates nearly all win in a handful of large masters, so a common target paid thousands of whole-plugin walks and a warm call ran longer than the cold one that built the index.

Both now gather the candidates they are about to judge a block at a time, one enumeration per winner plugin, through the `WinnerBodies` gather the `where_source=winner` scan and the forward walk already use. The scan's formid-set branch buffers a chunk of the universe, gathers it, and then scans that chunk in the universe's own order, so `offset=`/`limit=` still tile exactly as before. The walk hands each block of candidates to the gather just before it verifies them — only the block it is about to judge, so a spent `walk.max_nodes` budget still stops the body reads with the reach, and a candidate judged at an earlier hop is remembered and not read again. `ResolveTreePinned` is untouched.

No response change: the same matches, the same order, the same per-cause drop counts, the same index accounting.

Closes #727

## Measured

Live order `E:\Authoria - Requiem Reforged`, 3,244 active plugins, driven over stdio as a private server per arm, working set sampled every 250 ms from the driver and reported as the max inside each call's window. Other lanes were measuring on the same machine throughout, so the wall clocks carry that contention — the index build alone ranged 42–63 s across runs, and it is not what changed. Each arm is `counts_only=true`; the order index was built before the first timed arm.

| arm | before | after | peak WS before → after |
|---|---:|---:|---|
| `references=` ArmorHeavy, unbounded, cold (includes the index build) | 100.0 s | 64.1 s | 1,758 → 1,456 MB |
| … the same arm minus the index build the response reports | 58.2 s (build 41.8 s) | 7.1 s (build 57.0 s) | |
| `references=` IronIngot, unbounded, warm | 223.3 s | 6.4 s | 1,669 → 1,255 MB |
| reverse `walk=` `*` from IronIngot, defaults (depth 16 / 2,000) | 275.0 s | 17.8 s | 1,537 → 1,337 MB |

Every arm returned the same answer before and after, character for character apart from the clocks: 3,314 matches for ArmorHeavy, 1,988 for IronIngot, and the walk `hop 1: 1988, hop 2: 12 (cut by walk.max_nodes) — selection = 2001`, with the same 293 index candidates dropped for the same cause.

## Tests

`RecordsReverseIndexTests.TheReverseLaneAnswersTheSameWithoutAPluginWalkPerMatch` drives both spellings over the fixture world and asserts the whole rendered answer — header, index accounting, drop line and every row — against goldens captured by running the same two calls on the pre-change build, with the epoch, the freshness key, whether this call paid the build, and the clocks blanked. It also asserts neither call costs a per-record plugin walk. On the pre-change code it fails on exactly that cost claim with the rows already matching, which is the claim: same answer, fewer walks. Existing reverse-lane tests are unchanged.

Full suite 2,002 passed (`tier!=bridge`), `ci-all` ALL PASS.


src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -366,12 +374,23 @@ public static IReadOnlyList<Hop> Transitive(ReverseReferenceIndex index, IReadOn
...
+            for (int i = 0; i < candidates.Count; i++)
             {
+                var k = candidates[i];
                 if (visited.Contains(k)) continue;
                 // The budget is spent before the candidate is verified, so a spent budget stops the body reads
                 // too — verification has no ordering effect, so a raised budget on a retry still sees this graph.
                 if (reached >= maxNodes) { capped = true; cut = true; break; }
+                if (prepare is not null && i >= prepared)
+                {
+                    int end = Math.Min(candidates.Count, i + PrepareBlock);
```
Avick3110: The gather runs a whole block ahead of the node budget, so `walk.max_nodes` no longer bounds the body reads — which is what both doc paragraphs on this method still claim ("a spent budget bounds the body reads as well as the reach", and the new "It is called only for candidates the walk is about to verify, so a spent budget stops the gather with the reads").

Concretely: a reverse `walk=` with `max_nodes=1` on a seed whose hop-1 candidate list is 50,000 long. Before, `Verify` read one body, `reached` hit 1, and the hop cut. Now the first non-visited candidate at `i=0` triggers `prepare` with a block of `PrepareBlock` = 2,000 candidates, so `ReverseWalkBatch.Gather` calls `WinnerBodies.For` on 2,000 keys and walks every winner plugin among them before the first `Verify` even runs — then the budget cuts after one. A deliberately small budget (the documented way to probe a hub cheaply, and the documented retry path after a cap) now costs up to 2,000 winner-body reads per hop instead of ~`max_nodes`.

Clamping `end` by the budget still left (`i + Math.Min(PrepareBlock, ...)`) is not exact, because drops mean the remaining budget is not an upper bound on candidates consumed — but it is a bound on candidates *reached*, so some clamp, or an honest rewording of both doc claims, would keep the invariant the comment asserts.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4629,67 +4629,95 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
...
+                    setPending.Add(fk);
+                    if (setPending.Count == SetGatherChunk && !DrainSet()) { setStopped = true; break; }
+                }
+                if (!setStopped && setPending.Count > 0) DrainSet();
+
+                // Gather one chunk's winner bodies — one enumeration per winner plugin — and scan its keys in the
+                // universe's own order. Returns false when the scan must stop.
+                bool DrainSet()
+                {
+                    var bodies = WinnerBodies.For(view, sess, setPending, null, out var faults);
```
Avick3110: `faults` is consumed only for the up-to-3 `unscannableSamples` strings and then dropped — nothing is added to `unreadablePlugins`. The sibling gather in the `where_source=winner` lane does the opposite (`foreach (var (plugin, fault) in faults) if (faultedWinners!.Add(plugin)) unreadablePlugins.Add(fault);`), and `unreadablePlugins` is what produces the response's `coverage gap: N plugin(s) could not be read ...` line and the `UnreadPlugins` list.

Scenario: MO2 (or xEdit) is holding `Skyrim.esm` open while an unbounded `references=` runs, and 1,800 of the 1,988 candidates win there. `CollectRecords` throws on the open, `WinnerBodies.For` records one `PluginUnreadableException` for the plugin, and all 1,800 records land in the miss branch. The caller sees `unscannable=1800` with three sampled rows and an otherwise clean-reading scan — no coverage-gap sentence, empty `UnreadPlugins` — even though the cause is now in hand and is exactly the whole-plugin gap the other lane names once. Since the two lanes answer the same shaped question, folding `faults` into `unreadablePlugins` here (guarded by a seen-set, as above) would keep them consistent.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4629,67 +4629,95 @@ public CrossQueryOutcome CrossQuery(IReadOnlyList<string>? typeSet, IReadOnlyLis
                         : null,
                     predicate.NeedsContainment ? fk => view.ParentOf(fk) : null);
                 var seenSet = new HashSet<FormKey>();
+                // The universe's bodies are gathered a CHUNK at a time, one enumeration per winner plugin in the
+                // chunk, instead of the whole-overlay seek per record GetRecord costs (#251, the shape the
+                // where_source=winner lane above already uses). An unbounded references= is thousands of keys whose
+                // winners sit in a handful of large masters, so per record that seek was the whole call. The chunk is
+                // the memory bound, and the rows are still scanned in the universe's own order, so offset= and
+                // limit= tile exactly as before.
+                const int SetGatherChunk = 10_000;
```
Avick3110, line 4700: The chunk is now the cancellation granularity, and `WinnerBodies.For` takes no `ct`. The retained comment two lines down still says "a client that aborted stops the scan inside one record", which was true when each record cost one `GetRecord`; it is no longer.

Worst case is the negated-only unbounded `references=` (the orphan sweep), whose universe is spread over the whole order rather than concentrated in a few masters: one 10,000-key chunk can name hundreds or thousands of distinct winner plugins, and `WinnerBodies.For` walks each one to completion with no cancellation check between them. A client that aborts is not observed until the whole chunk is gathered. `ReverseWalkBatch.Gather` has the same shape with no `ct` at all.

A `ct` parameter on `WinnerBodies.For`, checked once per plugin in its `foreach (var (plugin, wanted) in byPlugin)` loop, would restore per-plugin responsiveness without touching the gather shape.

src/housecarl-core/ReverseReferenceIndex.cs
Avick3110: Folded. The block span is now `Math.Max(1, Math.Min(PrepareBlock, maxNodes - reached))`, so `max_nodes=1` on a 50,000-candidate hop gathers one body, not 2,000. Both doc claims are reworded to what the clamp actually buys: the block is never larger than the nodes the budget could still record, so a spent budget stops the gather and a small one keeps the reads near its own size — and because drops do not spend the budget, a drop-heavy hop asks for another block rather than reading past it. The default-budget arms are unchanged by the clamp (re-measured after: cold 62.2 s, warm 6.4→7.2 s, reverse walk 16.9 s, same answers).

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded. `DrainSet` now does what the winner-source lane does: `foreach (var (plugin, fault) in faults) if (faultedSetWinners.Add(plugin)) unreadablePlugins.Add(fault);` against a per-call seen-set, so a held-open winner plugin is named once in the coverage-gap line and in `UnreadPlugins` instead of surfacing only as three sampled unscannable rows. Worth noting this is a gain over the pre-change lane too, not just a restoration: the per-record `GetRecord` threw per key and the catch only ever counted it, so the whole-plugin cause was never in hand there.

src/housecarl-mcp/LoadOrderService.cs
Avick3110: Folded, both halves. `WinnerBodies.For` takes an optional `CancellationToken` and checks it once per plugin in the `byPlugin` loop, and the scan passes its `ct`. The reverse walk had no token at all, so `ReverseWalkBatch.Run` now takes one and `RecordsTools` passes the tool call's `ct` into it, which the gather then honours the same way. The retained comment is corrected to say what is now true — the scan stops between records and the gather stops between plugin walks.

Avick3110:
Folded all three review findings in 56dd648e, rebased on origin/main (the CHANGELOG conflict with the cycle-report entry resolved by keeping both), rebuilt, retested: 2,010 passed (`tier!=bridge`), `ci-all` ALL PASS.

Re-measured after the fold on the same live order, so the clamp is on the record as costing nothing at the default budget:

| arm | before | after (folded) |
|---|---:|---:|
| `references=` ArmorHeavy cold (incl. index build) | 100.0 s | 62.2 s (build 54.8 s) |
| `references=` IronIngot warm | 223.3 s | 7.2 s |
| reverse `walk=` from IronIngot, defaults | 275.0 s | 16.9 s |

Same answers as before the change: 3,314 and 1,988 matches, `hop 1: 1988, hop 2: 12 (cut by walk.max_nodes) — selection = 2001`, 293 candidates dropped for the same cause.

One thing I did not add: a test for the budget clamp. The records fixture reaches one candidate per hop, so a `max_nodes=1` walk asks for a one-key block either way and the assertion would pass on the unclamped code too — it would need a world with a wide hop, which is a bigger fixture than this change warrants. The claim is in the clamp and in the two doc paragraphs; say the word if you want the world built.

src/housecarl-core/ReverseReferenceIndex.cs
```diff
@@ -366,12 +377,28 @@ public static IReadOnlyList<Hop> Transitive(ReverseReferenceIndex index, IReadOn
...
                 // The budget is spent before the candidate is verified, so a spent budget stops the body reads
                 // too — verification has no ordering effect, so a raised budget on a retry still sees this graph.
                 if (reached >= maxNodes) { capped = true; cut = true; break; }
+                if (prepare is not null && i >= prepared)
+                {
+                    // The block never runs further ahead than the budget could still record, so a deliberately small
+                    // max_nodes reads few bodies: a spent budget breaks above before a block is ever asked for, and
+                    // a nearly-spent one asks for a block that size. Candidates the check drops do not spend the
+                    // budget, so a drop-heavy hop asks again rather than reading past it.
+                    int span = Math.Max(1, Math.Min(PrepareBlock, maxNodes - reached));
```
Avick3110: Clamping the gather block to the remaining node budget can fold the fix back to one whole-plugin walk per candidate on a drop-heavy tail.

`span` is `min(PrepareBlock, maxNodes - reached)`, and a candidate the verifier drops never advances `reached` — so once the budget is nearly spent the block size stays pinned at the tiny remainder while the walk keeps chewing through candidates. Concrete case, on the default budget: hop 1 reaches 1,988 of 2,000 (the PR's own IronIngot arm). Hop 2's `ReferencersOf` over a 1,988-record frontier is tens of thousands of candidates. If the first several thousand of them are new (not in `linksOf`) but are dropped by the winner-body check — winner dropped the link, deleted, no live body — then every 12 candidates triggers its own `prepare`, i.e. its own `WinnerBodies.For` over the distinct winner plugins of those 12 keys. That is thousands of full plugin enumerations, which is exactly the cost #727 is about. `walk.max_nodes=1` is the same shape at the extreme, and is reachable by a user probing the walk.

The reach bound is `reached >= maxNodes` above and is unaffected by how wide the gather is — the budget clamp here only limits how many bodies get *read* past the cut, at most one block's worth. Gathering `PrepareBlock` regardless (and letting the budget check above stop the reach) trades ≤ 2,000 possibly-unused body reads for a hard cap on the plugin walks. If the deliberate choice is the other way, it would help to say in the comment that a drop-heavy hop under a nearly-spent budget still pays the old per-candidate cost, since the PR body reads as though that case is closed.

src/housecarl-mcp/ReverseWalkBatch.cs
```diff
@@ -70,6 +71,20 @@ public static Result Run(LoadOrderService svc, IReadOnlyList<string> seeds, int
...
+        // nearly all win in a handful of large masters, and per record that seek was the walk. Only the block the
+        // walk is about to judge is gathered, so a spent node budget stops the gather too.
+        Dictionary<FormKey, IMajorRecordGetter> gathered = new();
+        var gatheredKeys = new HashSet<FormKey>();
+        void Gather(IReadOnlyList<FormKey> block)
+        {
+            // A candidate judged at an earlier hop is remembered, so its body is not read again and not gathered.
+            var need = new List<FormKey>(block.Count);
+            foreach (var k in block) if (!linksOf.ContainsKey(k)) need.Add(k);
+            gathered = WinnerBodies.For(view, session, need, null, out _, ct);
```
Avick3110: `out _` throws away the one piece of information the gather added over `GetRecord`.

`WinnerBodies.For` now separates the two causes — `PluginUnreadableException` for a plugin that will not open (a file another program is holding open) versus `PluginUnscannableException` for a fault in the walk after a good open (the plugin changed under the index) — and both scan lanes in `LoadOrderService` name the plugin once in the response's coverage-gap line. This lane discards the dictionary, so every candidate whose winner plugin could not be read lands in `DropCensus.Unreadable` as a bare count with no plugin named.

Scenario: a user has one master open in xEdit and runs a reverse `walk=`. Every candidate winning in that master is silently folded into "N unreadable", the same undifferentiated number they would get from a genuine index inconsistency, and the response never says which file to close. The census record's own doc comment says an unreadable winner is "a coverage gap, not a verdict" — a gap the caller can only report if the cause travels with it. Collecting the faults into a set and surfacing the plugin names beside the drop line would keep the behaviour parity claim and add the disclosure the rest of the tree makes.

src/housecarl-core/WinnerBodies.cs
```diff
@@ -44,6 +46,7 @@ public static Dictionary<FormKey, IMajorRecordGetter> For(
         }
         foreach (var (plugin, wanted) in byPlugin)
         {
+            ct.ThrowIfCancellationRequested();   // a client that aborted stops the gather between plugin walks
```
Avick3110, line 49: Not from this diff, but this PR adds two new callers that make it reachable: the `catch (Exception ex)` a few lines below swallows `OutOfMemoryException` and records it as "plugin X could not be read".

`BodyGather` deliberately rethrows OOM for exactly this reason — its comment says an OOM "is NOT a fault of this plugin", and treating it as one makes the failure worse rather than recovering from it. Here it is worse still: the caller carries on to the next plugin in `byPlugin` and keeps allocating, and the formid-set lane added in this PR then reports every candidate of that plugin as unscannable and moves to the next 10,000-key chunk. The user gets a response that names a perfectly readable master as a coverage gap, with no hint that the machine ran out of memory.

An unbounded `references=` on a very common target with an orphan-sweep-sized universe is the arm most likely to hit it. `catch (OutOfMemoryException) { throw; }` before the general catch, matching `BodyGather`, would keep the two causes honest.

src/housecarl-core/ReverseReferenceIndex.cs
Avick3110: Folded, and you are right that the first clamp traded one degeneration for another. The block is now `Math.Clamp(maxNodes - reached, MinPrepareBlock, PrepareBlock)` with a floor of 256: the budget still shrinks the block, so a small `max_nodes` does not gather a whole hop ahead (review 1's finding), but it can never shrink below the floor, so the 12-left tail you describe gathers 256 candidates at a time instead of 12 and cannot fall back to a walk per candidate. The overshoot past the cut is bounded by the floor, and both doc paragraphs now say that rather than claiming the budget bounds the reads exactly. The IronIngot arm — whose hop 2 is exactly the drop-heavy tail at 12 of budget left — went from 16.9 s to 13.0 s with the floor in, on the same order and same answer.

src/housecarl-mcp/ReverseWalkBatch.cs
Avick3110: Folded. `Gather` now collects the fault dictionary's plugin names into a seen-set-guarded list, `ReverseWalkBatch.Result` carries them as `UnreadableWinners`, and the walk lane renders them in parentheses after the unreadable cause: `N whose winning plugin could not be read — a coverage gap, not a verdict (Skyrim.esm)`. So the xEdit-holding-a-master case now names the file to close instead of handing back the same bare count a genuine index inconsistency would give.

src/housecarl-core/WinnerBodies.cs
Avick3110: Folded: `catch (OutOfMemoryException) { throw; }` ahead of the two fault catches in `WinnerBodies.For`, with a comment pointing at the same reason BodyGather gives — an OOM is not a fault of this plugin, naming a readable master as a coverage gap misreports it, and carrying on to the next plugin keeps allocating into the failure.

Avick3110:
Review 2 folded in 8eba2047, rebased on origin/main (CHANGELOG conflict with the bulk-apply gather entry resolved by keeping both), rebuilt, retested: 2,017 passed (`tier!=bridge`), `ci-all` ALL PASS.

Re-measured on the live order after the fold — the block floor helps the very hop the finding named (the IronIngot walk's hop 2 runs with 12 of budget left):

| arm | before | review 1 | review 2 |
|---|---:|---:|---:|
| `references=` ArmorHeavy cold (incl. build) | 100.0 s | 62.2 s (build 54.8 s) | 48.5 s (build 44.0 s) |
| `references=` IronIngot warm | 223.3 s | 7.2 s | 4.6 s |
| reverse `walk=` from IronIngot, defaults | 275.0 s | 16.9 s | 13.0 s |

Same answers throughout: 3,314 and 1,988 matches, `hop 1: 1988, hop 2: 12 (cut by walk.max_nodes) — selection = 2001`, 293 dropped for the same cause.

PR #762, opened 2026-09-14
Negate a where= string operator, and name the near miss on an exact editorid=

### LINKED ISSUE #668: [GAP] where= has no negated string operator; "not contains" is refused with a pointer to "not in"
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/kid-authoring/VERDICT.md`).

`where=["Name not contains Dagger"]` refuses with "'not' must be followed by 'in'", and `not in` takes a value list, so a complement over a substring cannot be expressed; the agent spilled 786 rows and filtered locally. One value on the operator axis (`not contains`, or a general `not` prefix over any string operator) closes it.

### LINKED ISSUE #669: [GAP] A legal where= that matches nothing gives no near-miss hint
Found by the skill gate of 2026-09-08 (`dev/skill-validation/gate-2026-09/mutagen-reference/VERDICT.md`).

`editorid = ArmorIronCuirass` matched zero records because the winner (Requiem for the Indifferent.esp) renames it `REQ_Heavy_Iron_Body`; both arms fell back to a FormID from memory. An empty result on an exact EditorID match could name the nearest candidates: the same EditorID in a losing copy (`defined_in: Skyrim.esm` has it), or the winner's EditorID for that FormID. One sentence, no inventory.

Avick3110:
`where=` gains a negated string operator, and a zero-row exact `editorid =` now says when the name it was given is real but carried by a losing copy.

**#668 — the negated string operator.** The grammar already read the operator as one word, with `not` as a special case that consumed a second word and accepted only `in`. Making `not` a general prefix over the word table is therefore ONE rule in the place that rule already existed, where three new operator values (`not contains`, `not startswith`, and a third if a string operator is ever added) would be three. So `not` now leads a string operator as well as the membership complement: `Name not contains Dagger`, `editorid not startswith REQ_`. There is no `ends_with` in the grammar, so the string ops it covers are `contains` and `startswith`; a fourth would inherit the complement for free. The prefix is deliberately NOT general over every operator, because every other one already has its own complement and two spellings for one test is the drift the one-grammar rule exists to avoid — `not` in front of `=`, `exists`, `has` or `in`-the-wrong-way is refused by name and pointed at `!=`, `missing`, `has_none`, `not in`. Negation is carried as a flag on the parsed predicate and applied at the leaf comparison, not at the outside of the predicate, so a quantifier keeps its own meaning: `Effects[*any].Name not contains X` is "some effect whose name lacks X", and `[*none]` is still how you ask the other question. It flips only a DEFINITE verdict, so on a FIELD path a record the path reads no value on stays a no-verdict and the existing accounting note still fires — a mistyped path under `not contains` cannot match everything. The identity term `editorid` is the exception, and is documented as one: it always gives a verdict, so a record with no EditorID is on the negated side, exactly where `!=` and `not in` already put it.

This changes the surface's vocabulary, so the `where=` parameter description carries the new form, its bound and that exception, and `kid-authoring` — the skill whose gate found the gap, where a `-exclusion` term is grounded — now names `not contains` as the census for a `-*Dagger` line. The tools-list snapshot in `src/housecarl-mcp-tests/data` records published NAMES only, with no descriptions, so no name changed and it is unchanged; the served description is still held to the client's 2,048-character cut by `PublishedDescriptionBoundTests` (`where=` is a parameter description, which the client delivers whole).

**#669 — the near-miss sentence.** The winner lane filters on the load-order winner's body, so an EditorID that only a losing copy carries reads as a clean zero. There is no EditorID index for a lookup to consult, so the hint is a bounded second look, and the gate is the shape the sentence can EXPLAIN rather than merely the shape that reaches the line: a zero-row, non-aggregate, winner-lane scan bounded by `types=`, with no `plugins=`, `formids=` set (including the universe an unbounded `references=` installs), `references=`, `references_none=`, `editorid_contains=` or `conflicts_only=`, and whose `where=` is nothing but the exact un-negated `editorid =` term. Anything else gives the zero another possible cause, and the sentence asserts one cause. The walk reads the EDID header and nothing else, goes one plugin at a time so a plugin that indexed but will not open now skips instead of ending the stream, skips a candidate whose winner body will not fetch, and stops at the first losing copy carrying the name or at `EditorIdNearMiss.Budget`. No candidate, no sentence. It lives in its own file rather than in `LoadOrderService.cs`.

Measured on the live order (`E:\Authoria - Requiem Reforged`, warm index, this branch's Release build driven over stdio): a zero-row `types=["ARMO"]` scan is 1,126 ms; the same scan as `editorid = <a name no plugin carries>`, which walks the whole type scope and gives up, is 2,278 ms; `editorid = ArmorIronCuirass`, which exits at Skyrim.esm, is 1,201 ms. The shapes the gate excludes pay nothing: `formids=["012E49:Skyrim.esm"] where=["editorid = ArmorIronCuirass"]` is 29 ms with no sentence, and `types=["ARMO"] where=["editorid = ArmorIronCuirass", "Value > 9000"]` is 1,237 ms with no sentence. The issue's own case now reads: `near miss: Skyrim.esm carries 'ArmorIronCuirass' at 012E49:Skyrim.esm, but Requiem for the Indifferent.esp wins that record and names it 'REQ_Heavy_Iron_Body' — editorid= is matched against the winner, so ask for that name, or scope the scan with plugins=["Skyrim.esm"]`. A winner that DROPPED the EditorID gets its own clause, pointing at the FormID and the `plugins=` scope rather than at a name that does not exist.

The negation was checked on the same order as a partition: over `types=["WEAP"] editorid startswith REQ_` (5,501), `Name contains Dagger` is 387, `Name not contains Dagger` is 5,109 and `Name missing` is 5 — 387 + 5,109 + 5 = 5,501, the complement over value-bearing records with the no-value records in neither, which is what the description and the changelog say for a field path.

Fixture tests cover the negated operator's positive and negative rows on both the `editorid` term and a real leaf, a negated term over a valid-but-unset leaf matching nothing, the three malformed-`not` refusals, the negated operator end to end through a scan, and the near-miss sentence on a renamed record — with the no-candidate, matching-scan, non-exact-operator, second-predicate, `formids=`-bounded and `plugins=`-scoped cases all getting no sentence. `dotnet build`, `dotnet test --filter "tier!=bridge"` (2,024 passed) and `housecarl-generator ci-all` (128/128) are green.

Closes #668
Closes #669


src/housecarl-mcp/RecordsTools.cs
```diff
@@ -125,7 +125,7 @@ public static string Records(
...
...ma or bracket is not expressible in a list — test it with '='. ONE '->' LINK STEP: 'Perks->editorid startswith REQ_NULL_'. PROVENANCE: 'winner = X.esp' — which records does X WIN; that term forces winner resolution over the scanned scope, the same declared cost as any winner scan. UNION-ARM tip: when a field is one of several shapes (an NPC's Configuration.Level is a fixed level OR a PC-level multiplier), a scalar predicate on one arm's sub-field doubles as an ARM-PRESENCE test: where=[\"Configuration.Level.LevelMult >= 0\"] returns exactly the NPCs on a multiplier. A body scan — must be combined with types= or plugins= to bound the work. A wrong path is reported loud, never a silent '0 matches'. Paths are scalar leaves: step into a list element with BRACKETS ('Effects[0].Data.Magnitude'), never a dotted hop; a WILDCARD over a list ('Effects[*].Magnitude') is a known future capability and is not built. FormLink equality takes a wire token ('WorkbenchKeyword = 088108:Skyrim.esm'), so a link filter is one predicate rather than a post-filter. A flag test uses has, never '=': '= 4194304' matches only records whose flags are EXACTLY that bit and skips every combination containing it.")]
```
Avick3110: The new rule stated here — "A negated term matches only records the path READS A VALUE on, so it is the complement over value-bearing records, not over every record" — is false for the `editorid` term, and the PR's own test asserts the opposite: `NotContains_OnEditoridKeepsEveryRecordTheContainsMatchDropped` expects the no-EditorID weapon to MATCH `editorid not contains recw1` (FieldPredicate.cs applies `if (p.Negate) ok = !ok;` after a switch whose default for a null EditorID is `false`, and that whole branch returns `EvalKind.Definite`).

Concrete misbehaviour: the census arithmetic this description (and the kid-authoring skill, and the changelog) tells a caller to do breaks on `editorid`. On a scope of N records, `editorid contains X` + `editorid not contains X` + `editorid missing` sums to MORE than N, because the records with no EditorID land in both the negated count and the `missing` count. On a real leaf (`Name`) the three do partition, which is what the PR body measured — so the rule as written reads as universal and is only true for one of the two path kinds.

Either exempt the identity term in the sentence (`editorid`/`winner` always have a definite verdict, so a negated `editorid` term is the complement over EVERY record, exactly as `!=` and `not in` already are there) or drop the universal phrasing. Same sentence needs the same fix in `plugin/CHANGELOG.md` and `.claude/skills/kid-authoring/SKILL.md`.

src/housecarl-core/EditorIdNearMiss.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                if (ct.IsCancellationRequested) return null;
+                if (++seen > Budget) return null;
+                if (depth <= 1) continue;                       // only one plugin provides it, so its copy IS the winner the scan already read
+                if (!string.Equals(body.EditorID, wanted, StringComparison.OrdinalIgnoreCase)) continue;
+                if (view.ResolveWinner(fk) is not { } w) continue;
+                if (string.Equals(w.WinnerPlugin, source, StringComparison.OrdinalIgnoreCase)) continue;   // this IS the winner — the scan saw this name and judged it
+                using var session = resolver.OpenSession();
+                var winnerBody = view.GetRecord(session, w.WinnerPlugin, fk);
+                if (winnerBody is null) return null;
```
Avick3110: `return null` here abandons the whole search on the first candidate whose winner body will not fetch, instead of moving on to the next copy.

Scenario: the order carries two records whose losing copies both spell `wanted`. The first one the walk reaches (load order) has a winner in a plugin that opens at index-build time but faults on this fetch — `GetRecord` hands back null — so the walk returns with no sentence even though the SECOND record is the genuine rename the user is asking about. The same `return null` on a transient fetch fault turns the feature off for that call with no trace.

`continue` is the conservative behaviour here and costs nothing: the budget still bounds the walk, and a candidate you cannot judge is exactly the one to skip rather than the one to stop on.

src/housecarl-core/EditorIdNearMiss.cs
```diff
@@ -0,0 +1,61 @@
...
+                var winnerEid = winnerBody.EditorID;
+                if (string.Equals(winnerEid, wanted, StringComparison.OrdinalIgnoreCase)) return null;     // same name at the top — the miss is not a rename
+                return $"near miss: {source} defines '{wanted}' at {FormIdToken.Of(fk)}, but {w.WinnerPlugin} wins that record and names it "
```
Avick3110: `source` is the FIRST plugin in load order whose copy carries `wanted`, which is not necessarily the plugin that DEFINES the record — but the sentence says "defines" and then prints `FormIdToken.Of(fk)`, whose plugin half is the definer.

Concrete case: Skyrim.esm defines the record as `ArmorIronCuirass`, a mid-order patch `Retexture.esp` overrides it and keeps that EditorID, Requiem renames it. If the type scope means Skyrim.esm's copy is skipped for any reason (it is not — but the general shape holds whenever the base copy carries a DIFFERENT name and a mid-order override introduces the name being asked for), the sentence reads `Retexture.esp defines 'X' at 012E49:Skyrim.esm` — it names one plugin as the definer and prints another in the FormID token in the same clause.

The `plugins=["{source}"]` advice is still correct, so the fix is only the verb: say the plugin CARRIES the name rather than defines it.

src/housecarl-core/EditorIdNearMiss.cs
```diff
@@ -0,0 +1,61 @@
...
+                     + $"'{winnerEid ?? "(no EditorID)"}' — editorid= is matched against the winner, so ask for that name, or scope the scan with plugins=[\"{source}\"].";
```
Avick3110: When the winner's copy carries no EDID at all, `winnerEid` is null and the sentence renders as `… wins that record and names it '(no EditorID)' — editorid= is matched against the winner, so ask for that name`. "Ask for that name" is unusable advice when there is no name, and the framing ("renames it") is wrong for this case: the winner DROPPED the EditorID, it did not rename the record. The caller is told to re-ask with a value that does not exist, and the only remedy left — the `plugins=[…]` scope — is buried behind advice that cannot work.

Worth splitting the null case into its own clause ("the winner carries no EditorID, so no `editorid=` value can reach this record — scope the scan with plugins=[…] or ask by FormID"), or dropping straight to the `plugins=`/FormID remedy.

src/housecarl-core/EditorIdNearMiss.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+        // A hint is never worth failing the answer it rides on: a plugin that will not open here is already
+        // reported by the scan itself, and anything else leaves the zero-row result exactly as it was.
+        catch (Exception) { return null; }
```
Avick3110: `RecordsIn` throws `PluginUnreadableException` on the first scoped plugin that opened at index-build time but will not open now, and that throw ENDS THE WHOLE STREAM (see its doc comment). The scan lane this hint rides on does not behave that way — `WinnerRecordsOfType` collects the unreadable plugin into `unreadablePlugins` and carries on.

So on an order where any plugin has become unreadable since the index build, the walk aborts at that plugin, this catch swallows it, and the near-miss sentence never fires for that order — for every query, silently, including the ones where the candidate sits AFTER the bad plugin. The scan itself reports the coverage gap, so the comment's "already reported by the scan itself" is true for the plugin, but not for the hint quietly turning itself off.

Catching `PluginUnreadableException` around the per-plugin step (or driving the walk off the same tolerant stream the scan uses) keeps the hint working on the rest of the order.

src/housecarl-mcp/LoadOrderService.cs
```diff
@@ -4935,6 +4935,12 @@ bool ScanRow(FormKey fk, int depth, IMajorRecordGetter filterBody, string? sourc
                        + string.Join("; ", unreadablePlugins.Select(u => u.Message));
             scanNote = scanNote is null ? gap : scanNote + " " + gap;
         }
+        // A legal editorid= that matched nothing on the WINNER lane: the name may be real and simply carried by a
+        // losing copy the winner renames, which a bare "0 matches" cannot say. One sentence when there is such a
+        // candidate, nothing when there is not (EditorIdNearMiss owns the rule and the budget).
+        if (total == 0 && groups is null && !hasPlugins && predicate?.ExactEditorId is { } wantedEid
```
Avick3110: The gate is `!hasPlugins`, but `plugins=` is not the only bound a `where=` scan can take -- `formids=` is a third one (line 4487: the refusal fires only when `!hasType && !hasPlugins && !hasFormidSet`), and an unbounded `references=` becomes one by setting `formidSet = universe; hasFormidSet = true` (line ~4502). All three branches fall through to this same line, so the near miss fires on scans whose scope it was never measured against, and it is handed `types`, which is **null** when the caller gave no `types=`.

Concretely, each of these is a legal zero-row call that reaches here with `!hasPlugins` true and `types == null`:

- `formids=[a few keys] where=["editorid = X"]` -- the set-alone branch. The scan read a handful of winner bodies; the hint then enumerates **every record of every plugin in the order** (`RecordsIn(Array.Empty<string>(), null)` -> `ScopeIndices` empty = whole order, `getterTypes` null = `EnumerateMajorRecords()`), opening every overlay until `Budget` = 400,000 headers.
- `references=[X] where=["editorid = Y"]` -- same, via the reverse-index universe.
- `types=["ARMO"] formids=[three keys] where=["editorid = X"]` -- bounded, but the hint walks all ARMO in the order, not the three records the caller asked about.

Two problems, and the second is the one that matters:

**Cost.** `EditorIdNearMiss`'s own doc comment justifies `Budget` with "The scope is already bounded (a where= scan must name types= or plugins=), so this is the backstop for an unusually broad one, not the normal stop" (`EditorIdNearMiss.cs:23`). That premise is false for the three shapes above -- there the budget *is* the stop, and the PR's measured 1.5 s worst case (a scoped `types=["ARMO"]` walk) is not the number that applies. A three-FormID lookup silently becomes a whole-order header sweep.

**A wrong explanation.** The sentence asserts a cause: "editorid= is matched against the winner, so ask for that name, or scope the scan with `plugins=[...]`". On a `formids=`/`references=` scan the zero row has nothing to do with the winner lane -- none of the *requested* records carries that EditorID -- and the hint names a record that was never in the caller's selection, with a remediation (`plugins=[source]`) that returns a record they did not ask for. Same shape for the terms the gate does not consult at all: `conflicts_only=`, `editorid_contains=`, `references_none=`. That is the "never a silently wrong answer" cornerstone: the hint is only true when the `editorid =` term is the *sole* reason for the zero, and the gate does not establish that.

Suggest tightening the gate to the shape the hint was actually reasoned about and measured on -- a type- or plugin-scoped winner-lane scan whose SELECT carries nothing but the `editorid =` term -- i.e. also require `types is not null`, `!hasFormidSet`, no `references`/`references_none`/`editorid_contains`, and `!conflictsOnly`. Anything looser and either the budget or the sentence is writing a cheque the gate cannot cover.

Avick3110:
Review folded, in `Hold the near-miss hint to the shape it can explain` (rebased on main, rebuilt, retested, live arms re-run).

**Gate (LoadOrderService).** Real, and the sharper half is the wrong explanation rather than the cost. The gate is now the shape the sentence can explain: `total == 0`, no `group_by=`, `types` non-empty, `!hasPlugins`, `!hasFormidSet`, `!conflictsOnly`, no `references=`, no `references_none=`, no `editorid_contains=` — and, one step past the finding, the `where=` set must be *nothing but* the `editorid =` term. `FieldPredicateSet.ExactEditorId` now returns null unless it is the sole predicate, because a second ANDed term is the same defect in miniature: the zero has another candidate cause and a record found by name may fail that other term anyway. `EditorIdNearMiss`'s budget comment no longer claims a premise the caller does not guarantee — it says the caller's gate bounds the walk to the scan's own `types=` scope.

Measured on the live order (warm index, this branch's build over stdio): `formids=["012E49:Skyrim.esm"] where=["editorid = ArmorIronCuirass"]` is **29 ms with no sentence** — before, that was the whole-order header sweep you describe. `types=["ARMO"] where=["editorid = ArmorIronCuirass", "Value > 9000"]` is 1,237 ms with no sentence. The shape that does fire is unchanged: baseline zero-row ARMO scan 1,126 ms, the answering case 1,201 ms, the no-candidate-anywhere walk 2,278 ms.

**Unreadable plugin ending the stream.** Real. The walk is now one plugin at a time over a new `IndexView.ScannablePluginNames` (load order, minus this build's excluded set), with the catch around the per-plugin step, so a file that indexed but will not open now skips and the rest of the order still answers.

**`return null` on an unfetchable winner.** Real, now `continue`. Kept distinct from the next one: the body fetch failing and the winner having no EditorID were both `null` through a `?.` and are now two separate reads.

**"defines" vs "carries".** Real — `source` is the first plugin in load order carrying the name, which need not be the definer, and `FormIdToken.Of(fk)` prints the definer in the same clause. The verb is now "carries".

**The winner with no EditorID.** Real. That case gets its own sentence: `… wins that record and gives it NO EditorID, so no editorid= value reaches it — ask by FormID, or scope the scan with plugins=[…]`. No "ask for that name" when there is no name.

**The `editorid` exception to the negation rule — folded as the wording fix, behaviour kept.** You are right that the sentence was universal and is not, and that the census arithmetic breaks on `editorid`; the description, `plugin/CHANGELOG.md` and `kid-authoring/SKILL.md` all now say that a negated term is the complement over value-bearing records **on a field path**, and that the identity term `editorid` always gives a verdict, so a record with no EditorID is on the negated side and overlaps `editorid missing`.

I did not take the other branch (make `editorid not contains` give no verdict on a null EditorID), because it would break a rule that predates this PR: on the `editorid` term, `!=` and `not in` already match a record with no EditorID, deliberately — `FieldPredicate.cs` carries the note that a blanket non-match there "would silently drop every such record from '!='". `not contains` is the same complement, so making it alone behave the other way would put the three complements on that one term into disagreement, which is a worse inconsistency than the one between the field path and the identity path — and that one is the grammar's own, already true of `!=` before this change. If you would rather the identity term lost its polarity everywhere, that is a change to `!=`/`not in` as well and belongs in its own issue.

`dotnet build`, `dotnet test --filter "tier!=bridge"` (2,024 passed, three new gate tests) and `ci-all` (128/128) are green on the rebase.

src/housecarl-core/EditorIdNearMiss.cs
```diff
@@ -0,0 +1,75 @@
...
+                             + $"so no editorid= value reaches it — ask by FormID, or scope the scan with plugins=[\"{source}\"].";
+                    if (string.Equals(winnerEid, wanted, StringComparison.OrdinalIgnoreCase)) continue;        // same name at the top — not the rename this is for
+                    return $"near miss: {source} carries '{wanted}' at {FormIdToken.Of(fk)}, but {w.WinnerPlugin} wins that record and names it "
+                         + $"'{winnerEid}' — editorid= is matched against the winner, so ask for that name, or scope the scan with plugins=[\"{source}\"].";
+                }
+            }
+            // A plugin that indexed but will not open now is already named in the scan's own coverage gap; the rest
+            // of the order still answers. Any other fault on one plugin is skipped on the same reasoning: a hint is
+            // never worth failing the answer it rides on.
+            catch (Exception) { continue; }
```
Avick3110, line 81: The per-plugin catch-all swallows `OutOfMemoryException`, which the same fault-isolation pattern in `BodyGather.CollectMany` was just taught to rethrow (73fd1aee: "OutOfMemoryException is NOT a fault of this plugin"). Concrete scenario: a plugin whose record walk exhausts memory on a large order — the OOM is caught here, the loop moves straight on to the next plugin and starts another whole-plugin enumeration, and the walk grinds through the rest of the order under memory pressure before returning a plain `null`. The caller never learns the machine ran out of memory; it just gets "no near miss". The same catch also hides a cancellation thrown from inside the enumerator: `ct` is only polled per record, so an `OperationCanceledException` surfacing from `RecordsIn` is absorbed and the walk continues over every remaining plugin, where the rest of the scan lane deliberately rethrows it (`catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` in `LoadOrderService.CrossQuery`).

src/housecarl-core/FieldPredicate.cs
```diff
@@ -337,9 +356,10 @@ public static (FieldPredicateSet? Set, string? Error) Parse(IReadOnlyList<string
         if (i >= text.Length)
             return (null, $"predicate '{raw}': no operator. Use one of = != > >= < <= contains startswith has has_any has_none exists missing in 'not in', e.g. \"{path} = <value>\" or \"{path} exists\".");
```
Avick3110: Two of the three operator-vocabulary sentences were left on the old grammar. This "no operator" message (and the symbolic one at line 371, `unrecognized operator at '...'`) still end at `in 'not in'`, while the word-branch refusal below was updated to name `'not' before contains/startswith`. Scenario: a caller writes `where=["Name"]` or `where=["Name !contains Dagger"]`, is handed the full operator list as the teaching, and that list does not contain the negated form the PR just added — so the one place a user goes looking for the vocabulary teaches the pre-#668 grammar.

src/housecarl-core/EditorIdNearMiss.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    if (ct.IsCancellationRequested) return null;
+                    if (++seen > Budget) return null;
+                    if (depth <= 1) continue;                      // only one plugin provides it, so its copy IS the winner the scan already read
+                    if (!string.Equals(body.EditorID, wanted, StringComparison.OrdinalIgnoreCase)) continue;
+                    if (view.ResolveWinner(fk) is not { } w) continue;
+                    if (string.Equals(w.WinnerPlugin, source, StringComparison.OrdinalIgnoreCase)) continue;   // this IS the winner — the scan saw this name and judged it
+                    // A winner body that will not fetch makes THIS candidate unjudgeable, not the walk: the next
+                    // copy carrying the name may be the rename the caller is asking about.
+                    var winnerBody = view.GetRecord(session, w.WinnerPlugin, fk);
```
Avick3110: This fetch passes no `getterType`, so `LoadOrderResolver.SeekBody` takes the flat branch and enumerates EVERY major record in the winner plugin to find one FormKey — the cost #354 added the typed group seek to avoid — even though the walk has the type scope in hand (`getterTypes`, and the candidate's own getter type off `body`). Scenario: the near miss lands on a record whose winner is a large patch (Requiem for the Indifferent.esp, a merged patch), and that one candidate pays a full flat overlay walk of that plugin. It compounds when several losing copies carry `wanted` and their winners keep the same name: each pays its own flat walk before the same-name check at the next line drops it with a `continue`. Passing the single scan type when `getterTypes` has exactly one entry (or the candidate's registered getter type) keeps the typed seek, and `SeekBody` still falls back to the flat walk on a miss.

src/housecarl-mcp-tests/WhereNearMissTests.cs
```diff
@@ -0,0 +1,76 @@
...
+                                     where: new[] { $"editorid contains {RecordsWorld.RenamedArmorOldEid}" });
+        Assert.DoesNotContain("near miss", r);
+    }
+
+    // ---- the gate: only where the sentence's one cause is the only cause ---------------------------
+
+    [Fact]
+    public void ASecondPredicateGetsNoSentence_TheZeroHasAnotherCandidateCause()
+    {
+        var r = RecordsTools.Records(Svc, types: new[] { "ARMO" },
```
Avick3110, line 68: The two pieces of behaviour the first review's fold ADDED are the only ones this file does not cover.

**The dropped-EditorID clause.** `EditorIdNearMiss` now has two sentences, and the second one — winner carries no EditorID, so the advice becomes "ask by FormID" rather than "ask for that name" — has no fixture. `RecordsWorld` gained `RenamedArmor` for the rename case; the null case needs one more override (a second armor whose winning override sets `EditorID = null`) to assert that the FormID/`plugins=` remedy is what comes back, and that the word "names it" does NOT. That clause is a user-facing sentence asserting a cause, on the surface whose cornerstone is "never a silently wrong answer" — it is exactly the kind of text a later edit can break without a single test going red.

**The gate terms.** The gate was widened from `!hasPlugins` to seven conditions (`groups is null`, `!hasFormidSet`, `!conflictsOnly`, `refSet is null`, `refNone is null`, `editoridContains` empty, `types is { Count: > 0 }`). The file tests the two that existed before the fold (`formids=`, `plugins=`) and none of the five the fold introduced. `conflicts_only=true types=["ARMO"] where=["editorid = <old name>"]` and `editorid_contains=... types=["ARMO"] where=["editorid = <old name>"]` are both one-line additions in the shape already used here, and they are the assertions that stop the gate silently loosening back to where the first review found it.

src/housecarl-mcp/RecordsTools.cs
```diff
@@ -125,7 +125,7 @@ public static string Records(
...
```
Avick3110, line 128: The new form is now named in four places — this parameter description, `plugin/CHANGELOG.md`, `kid-authoring/SKILL.md`, and the refusal messages — and missed in the one a user reads first.

`README.md:81` is the grammar's published vocabulary sentence: "`where=` accepts comparisons ..., EditorID tests ..., flag tests over bit fields (`has`, `has_any`, `has_none`), presence ..., membership from a file ..., one link step ..., a provenance term ..., a containment step ..., and quantified steps over a list." It enumerates the operator set deliberately and exhaustively — `has_any`/`has_none` were added to it when they shipped — and it still has no negated string operator in it. CLAUDE.md points users at README.md as the entry document, so a reader who takes that sentence as the operator list has no way to learn `not contains` exists except by trying it.

One clause in that sentence, beside the EditorID tests.

Avick3110:
All five folded, in `Fold the second review on the near-miss hint` (rebased on main, rebuilt, retested).

**OOM and cancellation swallowed by the per-plugin catch.** Real. `OutOfMemoryException` and `OperationCanceledException` now leave by their own `catch`/`throw` ahead of the general one, matching `BodyGather.CollectMany` (73fd1aee) and the scan lane's own rethrow. The per-record poll went with them: it was `if (ct.IsCancellationRequested) return null`, which would have contradicted the rethrow one line away, and is now `ct.ThrowIfCancellationRequested()`.

**Flat winner fetch.** Real. The fetch now passes the candidate's own getter type, picked out of the scan's type scope (`getterTypes?.FirstOrDefault(t => t.IsInstanceOfType(body))`), so `SeekBody` takes the typed group seek and keeps its flat fallback. That also fixes the compounding case: several losing copies whose winners keep the same name each paid a flat walk before the same-name `continue`.

**The two stale operator lists.** Real. Both the "no operator" and the symbolic "unrecognized operator at" refusals now end `… missing in, 'not in', or 'not' before contains/startswith`, the same tail the word-branch refusal carries.

**README.** Real. The grammar sentence now names the negated string operators beside the EditorID tests.

**Tests.** Real, and both added. `RecordsWorld` gained a second armor whose winning override sets `EditorID = null`, and `AWinnerThatDroppedTheEditorIdGetsTheFormIdRemedy_NotAskForThatName` asserts the FormID/`plugins=` remedy comes back and that "names it" does not. The gate terms the first fold introduced are now covered one test each: `conflicts_only=`, `references=`, the negated `references=` entry (which is how `references_none` is spelled on this tool), `group_by=` aggregate, and the second-`where=`-predicate case, beside the `formids=` and `plugins=` tests that were already there.

One substitution worth naming: there is no `editorid_contains=` parameter on `housecarl_records` — 2.0 replaced it with the `editorid` term — so that gate condition cannot be driven from the tool surface. It still guards the other `CrossQuery` callers that pass it, and the test in its place is the negated `references=` entry.

`dotnet build`, `dotnet test --filter "tier!=bridge"` (2,047 passed) and `ci-all` (128/128) are green on the rebase.

